PR #2502, opened 2021-08-25
Update Kata to use qemu-6.1

dgibson:
qemu-6.1 is now released.  This includes a bunch of things we've been carrying around patches for, and also ACPI hotplug support for q35 which we need to fix some future problems.

Update our packaging to use the new qemu version.

PR #2645, opened 2021-09-15
runtime/qemu: Move from query-cpus to query-cpus-fast

### LINKED ISSUE #2643: CI failure: QMP command query-cpus not found
> Warning  FailedCreatePodSandBox  11s (x6 over 88s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = CreateContainer failed: failed to get thread ids from hypervisor: QMP command failed: The command query-cpus has not been found: unknown

Looks to be related to https://github.com/kata-containers/kata-containers/pull/2502 ?

/cc @dgibson Does qemu v6.1 drop some QMP commands?

dgibson:
We recently updated to using qemu-6.1 (from qemu 5.2).  Unfortunately one
breaking change in qemu 6.0 wasn't caught by the CI.

The query-cpus QMP command has been removed, replaced by query-cpus-fast
(which has been available since qemu 2.12).  govmm already had support for
query-cpus-fast, we just weren't using it, so the change is quite easy.

fixes #2643

### DIFF
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -2194,7 +2194,7 @@ func (q *qemu) getThreadIDs(ctx context.Context) (vcpuThreadIDs, error) {
 		return tid, err
 	}
 
-	cpuInfos, err := q.qmpMonitorCh.qmp.ExecQueryCpus(q.qmpMonitorCh.ctx)
+	cpuInfos, err := q.qmpMonitorCh.qmp.ExecQueryCpusFast(q.qmpMonitorCh.ctx)
 	if err != nil {
 		q.Logger().WithError(err).Error("failed to query cpu infos")
 		return tid, err
@@ -2203,7 +2203,7 @@ func (q *qemu) getThreadIDs(ctx context.Context) (vcpuThreadIDs, error) {
 	tid.vcpus = make(map[int]int, len(cpuInfos))
 	for _, i := range cpuInfos {
 		if i.ThreadID > 0 {
-			tid.vcpus[i.CPU] = i.ThreadID
+			tid.vcpus[i.CPUIndex] = i.ThreadID
 		}
 	}
 	return tid, nil


amshinde (approved):

amshinde:
@dgibson Can we add a test that exercises this code?

dgibson:
> @dgibson Can we add a test that exercises this code?

We can and we should.  *I* am entirely unfamiliar with this part of the code, so I'm not sure I'm the right person to do it.

Also, I'm a bit confused by the situation: in #2643, @bergwolf notes it as a CI failure, but it didn't show up in the CI runs that happened before #2502 was merged (or it wouldn't have been merged).  I'm hoping @bergwolf can clarify what circumstances the problem is showing up in.

fidencio:
> > @dgibson Can we add a test that exercises this code?
> 
> We can and we should. _I_ am entirely unfamiliar with this part of the code, so I'm not sure I'm the right person to do it.
> 
> Also, I'm a bit confused by the situation: in #2643, @bergwolf notes it as a CI failure, but it didn't show up in the CI runs that happened before #2502 was merged (or it wouldn't have been merged). I'm hoping @bergwolf can clarify what circumstances the problem is showing up in.

See: https://github.com/kata-containers/kata-containers/issues/2643#issuecomment-919769680

bergwolf (approved):

src/runtime/virtcontainers/qemu.go
```diff
@@ -2203,7 +2203,7 @@ func (q *qemu) getThreadIDs(ctx context.Context) (vcpuThreadIDs, error) {
 	tid.vcpus = make(map[int]int, len(cpuInfos))
 	for _, i := range cpuInfos {
 		if i.ThreadID > 0 {
-			tid.vcpus[i.CPU] = i.ThreadID
+			tid.vcpus[i.CPUIndex] = i.ThreadID
```
fidencio, line 2206: So, the CPU field got renamed to CPUIndex, is it?

fidencio (approved):
lgtm, thanks @dgibson.

src/runtime/virtcontainers/qemu.go
dgibson: Not exactly.  It's actually a entirely different structure, though it has similar contents.  The only change that mattered to us was that cpu index is how in this field rather than the other one.

dgibson:
The metrics and vfio CI jobs seemed to be stuck (triggered, but no details showing).
Since this is a pretty urgent fix, and those are unlikely to exercise things that the other jobs don't, I'm going ahead with the merge.

dgibson:
It seems like the metrics and vfio jobs are stuck.  I've triggered them, but no results are showing up :(.

src/runtime/virtcontainers/qemu.go
fidencio: Ack.

fidencio:
> It seems like the metrics and vfio jobs are stuck. I've triggered them, but no results are showing up :(.

Yep, machine seems to be off. /cc @chavafg @GabyCT, please, could you bring this machine up again?

devimc (approved):

PR #2655, opened 2021-09-16
runtime: Remove outdated TestStoreContainer

Jakob-Naucke:
Due to #2332 being merged after running tests for #2604, and the latter
being merged now, a test for the now removed `storeContainer` was added.
Remove it.

Fixes: #2652
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

PR #2662, opened 2021-09-16
runtime: tracing: Use root context to stop tracing

### LINKED ISSUE #2661: Use rootCtx instead of shim ctx when stopping tracing
Call `StopTracing` with `s.rootCtx`, which is the root context for tracing, instead of `s.ctx`, which is parent to a subset of trace spans.

cmaf:
Call StopTracing with s.rootCtx, which is the root context for tracing,
instead of s.ctx, which is parent to a subset of trace spans.

Fixes #2661


cmaf:
Static check failure should be fixed after https://github.com/kata-containers/kata-containers/pull/2655

fidencio (approved):
lgtm, thanks @cmaf!

devimc (approved):

PR #2664, opened 2021-09-16
stable-2.2 | runtime: tracing: Use root context to stop tracing

### LINKED ISSUE #2663: 2.2 Backport: Use rootCtx instead of shim ctx when stopping tracing
Backport of #2661 

Call StopTracing with s.rootCtx, which is the root context for tracing, instead of s.ctx, which is parent to a subset of trace spans.

cmaf:
Call StopTracing with s.rootCtx, which is the root context for tracing,
instead of s.ctx, which is parent to a subset of trace spans.

Backport of #2662

Fixes #2663


fidencio (approved):
lgtm, thanks @cmaf!

devimc (approved):

PR #2666, opened 2021-09-16
runtime: tracing: Fix logger passed in newContainer

### LINKED ISSUE #2665: Trace call in newContainer uses incorrect logger
In calls to `Trace`, if a subsystem's logger is available it will be used to generate a log message for nil context errors. If a logger isn't available it will use the `kataTraceLogger`. 

A call to `Trace` in container.go uses the sandbox logger.  The container logger has not been created yet, but we can use `nil` to use the `kataTraceLogger` so that the log message doesn't erroneously attribute a call from the container subsystem to the sandbox subsystem instead.

cmaf:
Change logger in Trace call in newContainer from sandbox.Logger() to
nil. Passing nil will cause an error to be logged by kataTraceLogger
instead of the sandbox logger, which will avoid having the log message
report it as part of the sandbox subsystem when it is part of the
container subsystem.

The kataTraceLogger will not log it as related to the container
subsystem, but since the container logger has not been created at this
point, and we already use the kataTraceLogger in other instances where a
subsystem's logger has not been created yet, this PR makes the call
consistent with other code.

Fixes #2665

### DIFF
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -738,7 +738,7 @@ func (c *Container) initConfigResourcesMemory() {
 
 // newContainer creates a Container structure from a sandbox and a container configuration.
 func newContainer(ctx context.Context, sandbox *Sandbox, contConfig *ContainerConfig) (*Container, error) {
-	span, ctx := katatrace.Trace(ctx, sandbox.Logger(), "newContainer", containerTracingTags, map[string]string{"container_id": contConfig.ID, "sandbox_id": sandbox.id})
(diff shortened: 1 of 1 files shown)


cmaf:
static checks failure is related to https://github.com/kata-containers/kata-containers/pull/2655

fengwang666 (approved):

Jakob-Naucke (approved):

fidencio (approved):
lgtm, thanks @cmaf!

~@jodh-intel, may I bother you for a review on this  one?~ That was a c&p error. :-)

PR #2668, opened 2021-09-16
stable-2.2 | runtime: tracing: Fix logger passed in newContainer

### LINKED ISSUE #2667: backport: Trace call in newContainer uses incorrect logger
Backport of #2665 

In calls to Trace, if a subsystem's logger is available it will be used to generate a log message for nil context errors. If a logger isn't available it will use the kataTraceLogger.

A call to Trace in container.go uses the sandbox logger. The container logger has not been created yet, but we can use nil to use the kataTraceLogger so that the log message doesn't erroneously attribute a call from the container subsystem to the sandbox subsystem instead.

cmaf:
Change logger in Trace call in newContainer from sandbox.Logger() to
nil. Passing nil will cause an error to be logged by kataTraceLogger
instead of the sandbox logger, which will avoid having the log message
report it as part of the sandbox subsystem when it is part of the
container subsystem.

The kataTraceLogger will not log it as related to the container
subsystem, but since the container logger has not been created at this
point, and we already use the kataTraceLogger in other instances where a
subsystem's logger has not been created yet, this PR makes the call
consistent with other code.

Backport of #2666
Fixes #2667

### DIFF
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -753,7 +753,7 @@ func (c *Container) initConfigResourcesMemory() {
 
 // newContainer creates a Container structure from a sandbox and a container configuration.
 func newContainer(ctx context.Context, sandbox *Sandbox, contConfig *ContainerConfig) (*Container, error) {
-	span, ctx := katatrace.Trace(ctx, sandbox.Logger(), "newContainer", sandbox.tracingTags())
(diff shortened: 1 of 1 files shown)


fidencio:
@cmaf, sorry for using this PR to do a test, but it's the most hand one to check whether /test-fc works.

fidencio (approved):
lgtm, thanks @cmaf!

devimc (approved):

PR #2654, opened 2021-09-16
shim: Add PullImage support

ashleyrobertson:
Add a new PullImage endpoint to the shim API.
Add new PullImage functions to the virtcontainers files, which allows 
the PullImage endpoint on the agent to be called.
Update the containerd vendor files to support new PullImage API changes.

Fixes #2651

Signed-off-by: ashleyrobertson ashleyro@uk.ibm.com

PR #2556, opened 2021-09-01
CCv0 | agent: Add PullImage endpoint

stevenhorsman:
This commit adds the PullImge endpoint to the agent
and the agent-ctl command to test it.

Fixes: #2509

Signed-off-by: Georgina Kinge <georgina.kinge@ibm.com>

PR #2669, opened 2021-09-17
[discussion] agent: for Confidential Computing poc

wllenyj:
* add ImageService in agent, use script to pull image and unpack.

```
IMAGE=$1
BUNDLE=$2

WD=/run/images

mkdir -p $WD
cd $WD

OCI_IMAGE=$(echo $IMAGE | awk -F\/ '{print $NF}')
IMAGE_NAME=$(echo $OCI_IMAGE | awk -F: '{print $1}')

skopeo copy docker://$IMAGE oci:$OCI_IMAGE > /dev/null || exit 1

umoci unpack --image $OCI_IMAGE --rootless ${IMAGE_NAME}_bundle > /dev/null

echo -n "$WD/${IMAGE_NAME}_bundle"
```

* CreateContainer need merge image spec.

Signed-off-by: wllenyj <wllenyj@linux.alibaba.com>

### DIFF
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -1,0 +1,127 @@
+use std::process::Stdio;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use protocols::image;
+use ttrpc;
+
+use crate::sandbox::Sandbox;
+
+pub struct ImageService {
+    sandbox: Arc<Mutex<Sandbox>>,
+}
+
+impl ImageService {
+    pub fn new(sandbox: Arc<Mutex<Sandbox>>) -> Self {
+        Self { sandbox }
+    }
+
+    async fn pull_image(&self, image: &str) -> Result<String> {
+        let shell = std::path::PathBuf::from("/opt/pull_image.sh");
+        let child = tokio::process::Command::new(shell)
+            .arg(image)
+            .kill_on_drop(true)
+            .stdin(Stdio::null())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()
+            .expect("pull image command failed to start");
+
+        let join_handler = tokio::spawn(async move {
+            match child.wait_with_output().await {
+                Ok(output) => {
+                    if output.status.success() {
+                        //println!(
+                        //    "child success {}",
+                        //    String::from_utf8(output.stdout).unwrap()
+                        //);
+                        return Ok(String::from_utf8(output.stdout).unwrap());
+                    } else {
+                        //println!("child fail {}", String::from_utf8(output.stderr).unwrap());
+                        return Err(anyhow!(
+                            "child fail {}",
+                            String::from_utf8(output.stderr).unwrap()
+                        ));
+                    }
+                }
+                Err(e) => {
+                    //println!("child exec err. {:?}", e);
+                    return Err(e.into());
+                }
+            }
+        });
+
+        let resp =
+            tokio::time::timeout(std::time::Duration::from_secs(100), join_handler).await???;
+        let mut sandbox = self.sandbox.lock().await;
+        sandbox.images.insert(String::from(image), resp.clone());
+        info!(
+            slog_scope::logger(),
+            "agent pull image success: {:?}", sandbox.images
+        );
+        return Ok(String::from(image));
+    }
+}
+
+#[async_trait]
+impl protocols::image_ttrpc::ImageService for ImageService {
+    async fn list_images(
+        &self,
+        _ctx: &ttrpc::r#async::TtrpcContext,
+        _req: image::ListImagesRequest,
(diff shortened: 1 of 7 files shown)


sameo:
@wllenyj Thanks for the PR. Please look at the [CCv0 branch](https://github.com/kata-containers/kata-containers/tree/CCv0) as there is a lot of duplication between what you're proposing and the work in that branch.

I think it would make sense to have a dedicated image service for the kata-agent, but:

1. That definition should first hit the CCv0 branch. There is no justification for pushing that upstream when we don't even know yet what the shim API for image management is going to look like.
2. It seems that the image service you're adding here is the entire (or close enough) CRI image service definition. I would prefer for the kata-agent service to be more limited, because the scope of that agent is much more limited than a container runtime (CRIO or containerd). For example attaching an entire sandbox context to a pull image request is not necessary as the kata-agent will only handle one unique pod. So this could be vastly simplified.

wllenyj:
@sameo Thanks for your commented. 
1. I'm just offering some ideas for discussion in my poc :) no need to merge.
2. Yes, it is CRI ImageService. Is it possible to need ListImage, ImageFsInfo ？

sameo:
> @sameo Thanks for your commented.
> 
>     1. I'm just offering some ideas for discussion in my poc :) no need to merge.

Ah, sorry for the misunderstanding. Thanks a lot for sharing!
 
>     2. Yes, it is CRI ImageService. Is it possible to need ListImage, ImageFsInfo ？

We may need an image listing endpoint, yes. What we know for sure at the moment is that we need an image pulling endpoint. Maybe we should start small and define an `image.proto` with only that endpoint for now, based on the CRI spec but reduced to take into account that the agent only manages one pod.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+#[async_trait]
+impl protocols::image_ttrpc::ImageService for ImageService {
+    async fn list_images(
+        &self,
+        _ctx: &ttrpc::r#async::TtrpcContext,
+        _req: image::ListImagesRequest,
+    ) -> ttrpc::Result<image::ListImagesResponse> {
+        Err(ttrpc::Error::RpcStatus(ttrpc::get_status(
```
liubin, line 75: Suggestions:
1 Use `UNIMPLEMENTED` instead of `NOT_FOUND`
2 Use a macro to simplify code `unimplemented_service("ListImage")`

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        });
+
+        let resp =
+            tokio::time::timeout(std::time::Duration::from_secs(100), join_handler).await???;
+        let mut sandbox = self.sandbox.lock().await;
+        sandbox.images.insert(String::from(image), resp.clone());
+        info!(
+            slog_scope::logger(),
+            "agent pull image success: {:?}", sandbox.images
```
liubin, line 62: This log will print all images in a sandbox, but not the just pulled one.

liubin:
Besides comments above, here also some questions:

- Does this conflict with https://github.com/kata-containers/kata-containers/pull/2556 ?
- Does this work with https://github.com/kata-containers/kata-containers/pull/2654 ?

wllenyj:
> Besides comments above, here also some questions:
> 
> * Does this conflict with [CCv0 | agent: Add PullImage endpoint #2556](https://github.com/kata-containers/kata-containers/pull/2556) ?
> * Does this work with [CCv0 | shim: Add PullImage support #2654](https://github.com/kata-containers/kata-containers/pull/2654) ?

Thanks for review, as i said above：
> I'm just offering some ideas for discussion in my poc :) no need to merge.
so should I change my title to avoid misunderstanding?

stevenhorsman:
@wllenyj - I'd echo Samuel's comments above:
- I link the creation of the separate image service and if we can start with the smallest change we require, so just PullImage for now that would be great.
- We've made similar changes in the [`CCv0` branch](https://github.com/kata-containers/kata-containers/tree/CCv0), but not split it out into a separate image service, so if you could re-work this, or create a new PR based on those changes that would be great.
Give me a shout on here, or slack if you want to discuss it further. Thanks!

wllenyj:
> @wllenyj - I'd echo Samuel's comments above:
> 
> * I link the creation of the separate image service and if we can start with the smallest change we require, so just PullImage for now that would be great.
> * We've made similar changes in the [`CCv0` branch](https://github.com/kata-containers/kata-containers/tree/CCv0), but not split it out into a separate image service, so if you could re-work this, or create a new PR based on those changes that would be great.
>   Give me a shout on here, or slack if you want to discuss it further. Thanks!

Hi, 
Thanks for your reply. 

I make it re-work now, the pr is: 
runtime : https://github.com/kata-containers/kata-containers/pull/2733
agent: https://github.com/kata-containers/kata-containers/pull/2734

Jakob-Naucke:
@wllenyj with #2733 open, can this be closed?

Issue #2592, opened 2021-09-07
virtcontainers: fc: parse vcpuID correctly


ananos:
In `getThreadIDs()`, the `cpuID` variable is derived from a string that already contains a whitespace. As a result, `strings.SplitAfter` returns the cpuID with a leading space. This makes any go variant of string to int fail (`strconv.ParseInt()` in our case).

This has been caused by a change in the naming scheme of vcpu threads for Firecracker after v0.19.1.

PR #2678, opened 2021-09-18
stable-2.2 | virtcontainers: fc: parse vcpuID correctly

ananos:
In getThreadIDs(), the cpuID variable is derived from a string that
already contains a whitespace. As a result, strings.SplitAfter returns
the cpuID with a leading space. This makes any go variant of string to int
fail (strconv.ParseInt() in our case). This patch makes sure that the
leading space character is removed so the string passed to
strconv.ParseInt() is "CPUID" and not " CPUID".

This has been caused by a change in the naming scheme of vcpu threads
for Firecracker after v0.19.1.

Fixes: #2592

Signed-off-by: Anastassios Nanos <ananos@nubificus.co.uk>


fidencio (approved):
thanks, @ananos!

cmaf (approved):

Issue #2691, opened 2021-09-22
The TestTemplateFactory  unit test case of the template_test.go executed failed


YchauWang:
# Description of problem

The TestTemplateFactory  unit test case of the template_test.go executed failed. We got a os.PathError, because the `/tmp/vc/mockfs` path was removed. the case, `expect tt.statePath not exist, if exist, it means this case failed.`, had been executed correctly, but the assert code failed. We should asserted a error.
```
    template_test.go:132: 
        	Error Trace:	template_test.go:132
        	Error:      	Expected nil, but got: &os.PathError{Op:"stat", Path:"/tmp/vc/mockfs", Err:0x2}
        	Test:       	TestTemplateFactory
```

# Expected result

the unit test case executed correctly.

YchauWang:
It istn't a bug, my env error.

PR #2692, opened 2021-09-22
runtime: fixed the factory template unit test case

YchauWang:
The '/tmp/vc/mockfs' path was removed. We should get a PathError by os.Stat method.
Changed the assert code of this test case.

Fixes: #2691


YchauWang:
Mistake commit!

PR #2702, opened 2021-09-22
packaging: Backport QEMU's GitLab switch to 5.1.x

Jakob-Naucke:
This brings #2699 to 5.1.x for ARM. Add a `no_patches` for 5.1.0 which
was missing apparently.

Fixes: #2701
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

/cc @gkurz @jongwu
I hope the attribution is correct this way, PTAL @wainersm

Issue #2701, opened 2021-09-22
Backport QEMU's move to GitLab to 5.1.x


Jakob-Naucke:
ouch -- ARM still failed because the patch didn't apply. Let me TAL.

_Originally posted by @Jakob-Naucke in https://github.com/kata-containers/kata-containers/issues/2699#issuecomment-924979395, see also rest of thread_

Jakob-Naucke:
that link didn't work somehow

PR #2706, opened 2021-09-23
qemu: use GitLab repos instead of qemu.org

yuanzhe-liu0:
arm using qemu 5.1.0, thus is affected by the wired submodules
link.

Fixes: #2705
Signed-off-by: Yuanzhe Liu <yuanzheliu09@gmail.com>


yuanzhe-liu0:
@jongwu

jongwu:
thanks @yuanzhe-liu0 
/test

jongwu:
help review this @fidencio @Jakob-Naucke

Jakob-Naucke:
Hmm how does this stack up against #2701? (@jongwu I told you I was working on it already...) Also, there's two redundant patch files in the project root.

yuanzhe-liu0:
hello @Jakob-Naucke -, this PR is for main branch

Jakob-Naucke:
@yuanzhe-liu0 ah, of course, sorry. Regardless, can you remove the patches in the project root and move qboot to GitLab as discussed in https://github.com/kata-containers/kata-containers/pull/2702#discussion_r714243271?

yuanzhe-liu0:
thanks @Jakob-Naucke -, I bring the patch in stalbe-2.2. here.

gkurz (approved):
It feels a bit strange that this change will get merged in main **after** it was merged in stable-2.2. Shouldn't this happen the other way around ?

Jakob-Naucke:
@gkurz this happened because a similar change affecting all architectures (5.2.x) only mattered for 2.2, and since that patch didn't work for ARM, I also created the patch for ARM 2.2 5.1.x (not least because I wasn't sure whether @jongwu was planning to go ahead with 6.1 right away). We could do a formal forward-port, but I'd skip it since this is already open.

Jakob-Naucke:
Just now I saw that @yuanzhe-liu0 added a commit referencing QEMU 6.1, but it... doesn't seem to do that?

gkurz (changes requested):
I don't understand these changes:
- adding a file which is a kata diff (patch 0001)
- adding a QEMU 6.1 diff at root (patch 0002)

It looks like patch 0001 should be dropped and patch 0002 be moved to tools/packaging/qemu/patches/6.1.x/ . Also I don't see any trace of patch 0002 in QEMU upstream, nor in the qemu-devel mailing list archives... is this the way to go ? Shouldn't the patch go through the QEMU community first ? Especially if it is supposed to fix a bug.

yuanzhe-liu0:
Oh! I made a mistake, thanks @gkurz @Jakob-Naucke -, I correct it now.

Jakob-Naucke (changes requested):
@yuanzhe-liu0 this is starting to look good.
- With patches present, you should remove `tools/packaging/qemu/patches/6.1.x/no_patches.txt`.
- Do you want to upgrade ARM to QEMU 6.1 by removing this? https://github.com/kata-containers/kata-containers/blob/3276f3b5b6c2f73f97405acd646581f0ae76c246/versions.yaml#L101-L106
- If yes, do we need the 5.1.x patch any more? We can keep it if you'd rather keep it for compatibility.

yuanzhe-liu0:
> @yuanzhe-liu0 this is starting to look good.
> 
> * With patches present, you should remove `tools/packaging/qemu/patches/6.1.x/no_patches.txt`.
> * Do you want to upgrade ARM to QEMU 6.1 by removing this? https://github.com/kata-containers/kata-containers/blob/3276f3b5b6c2f73f97405acd646581f0ae76c246/versions.yaml#L101-L106
> * If yes, do we need the 5.1.x patch any more? We can keep it if you'd rather keep it for compatibility.

The qemu installation for arm is using these lines so we can't remote them for now. I just add qemu patch for arm here.

gkurz (changes requested):
I'm still concerned that these patches don't come from upstream QEMU. I'm opening an issue so that we can discuss this matter before going forward with this PR.

Jakob-Naucke (commented):
It LGTM now, but let's settle #2721.

jongwu:
Hello @gkurz -, yeah, these two patches are not from upstream qemu. For the first one - cold plug nvdimm -. this patch is from the early stage of the "enable nvdimm hotplug" patch set. the code plug parts is removed later but it's import for nvdimm hotplug without ACPI, that's why we pick it up.
the second one, it's a bug fix for hotplug memory. there is an alignment error when memory hotplug after nvdimm plug.
these 2 patches works well in qemu 5.1 so we port it to 6.1

gkurz:
Hi @jongwu !

> Hello @gkurz -, yeah, these two patches are not from upstream qemu. For the first one - cold plug nvdimm -. this patch is from the early stage of the "enable nvdimm hotplug" patch set. the code plug parts is removed later but it's import for nvdimm hotplug without ACPI, that's why we pick it up.

Ok so this is the justification for the need. My concern is more : why this isn't getting pushed to upstream QEMU as well ?
What will happen if some user decides to use QEMU 6.1 from upstream or from distro without this patch instead of ours ?
I'm afraid this could break and be blamed on kata, even though the missing piece should be in QEMU.

> the second one, it's a bug fix for hotplug memory. there is an alignment error when memory hotplug after nvdimm plug.
> these 2 patches works well in qemu 5.1 so we port it to 6.1

This issue looks serious enough that we should also push it to upstream QEMU. We'd get valuable feedback from the QEMU community on the way.

Jakob-Naucke:
see https://github.com/kata-containers/tests/issues/4013 for tracking of CLH failure

Jakob-Naucke (approved):
Okay, let's do this first.

yuanzhe-liu0:
Hello @Jakob-Naucke @gkurz -, I have removed the qemu patch, please review.

gkurz (approved):
It wouldn't hurt to mention the QEMU upstream commit this was backported from, but anyway the patch is fine.

tools/packaging/qemu/patches/5.1.x/0014-gitmodules-use-GitLab-repos-instead-of-qemu.org.patch
```diff
@@ -0,0 +1,111 @@
+From 104e56711f131d80de82caed8759947509576e3b Mon Sep 17 00:00:00 2001
```
Jakob-Naucke, line 1: > It wouldn't hurt to mention the QEMU upstream commit this was backported from, but anyway the patch is fine.

@gkurz isn't this a mention

gkurz: Hmm I don't know where this hex blob is coming from but it isn't a valid SHA1 in the upstream QEMU git repo.

Original commit is :

https://gitlab.com/qemu-project/qemu/-/commit/9911ca0d1bca846f159ebdb48b9d8a784c959589

Jakob-Naucke: Ah yes. Of course it's different because the patch is different.

Issue #2638, opened 2021-09-14
Optional configmap and secret volumeMounts that are not present cause fatal error instead of creating folders


skaegi:
This is another case we found in our CI/CD system when moving from Kata1 to Kata2 where consumers of our SaaS are referencing "optional" secrets and configmaps. These pods fail with a message like... `Error: failed to create containerd task: No such file or directory (os error 2): unknown`

Here's a sample pod to illustrate
```
apiVersion: v1
kind: Pod
metadata:
  name: rc-alpine
spec:
  runtimeClassName: kata-qemu
  containers:
    - name: alpine
      image: alpine
      imagePullPolicy: IfNotPresent
      command:
      - sleep
      - "50000000"
      workingDir: /does-not-exist
      volumeMounts:
      - mountPath: /config/not-provided-secret
        name: secret
      - mountPath: /config/not-provided-config
        name: configmap
  volumes:
    - name: configmap
      configMap:
        name: not-provided-config
        optional: true
    - name: secret
      secret:
        optional: true
        secretName: not-provided-secret
```

This results in ...
```
Events:
  Type     Reason     Age              From               Message
  ----     ------     ----             ----               -------
  Normal   Scheduled  11s              default-scheduler  Successfully assigned default/rc-alpine to 10.94.4.156
  Normal   Pulled     6s (x2 over 6s)  kubelet            Container image "alpine" already present on machine
  Normal   Created    6s (x2 over 6s)  kubelet            Created container alpine
  Warning  Failed     6s (x2 over 6s)  kubelet            Error: failed to create containerd task: No such file or directory (os error 2): unknown
  Warning  BackOff    3s (x3 over 5s)  kubelet            Back-off restarting failed container
```

What should happen is that empty folders should be created...
e.g. with vanilla runc or kata 1 and the same pod definition I get
```
$ k apply -f ~/kube/rc-alpine.yaml 
pod/rc-alpine created
$ k exec rc-alpine -- ls -al /config
total 12
drwxr-xr-x    4 root     root          4096 Sep 14 20:04 .
drwxr-xr-x    1 root     root          4096 Sep 14 20:04 ..
drwxrwxrwx    3 root     root          4096 Sep 14 20:04 not-provided-config
drwxrwxrwt    3 root     root            80 Sep 14 20:04 not-provided-secret
```

fidencio:
Let me also cc @lifupan

egernst:
[...]

PR #2714, opened 2021-09-23
watcher: ensure we create target mount point for storage

egernst:
We would only create the target when updating files. We need to make
sure that we create the target if the source is a directory. Without
this, we'll fail to start a container that utilizes an empty configmap,
for example.

Add unit tests for this.

Fixes: #2638

### DIFF
--- a/src/agent/src/watcher.rs
+++ b/src/agent/src/watcher.rs
@@ -269,6 +269,19 @@ impl SandboxStorages {
             let entry = Storage::new(storage)
                 .await
                 .with_context(|| "Failed to add storage")?;
+
+            // If the storage source is a directory, let's create the target mount point:
+            if entry.source_mount_point.as_path().is_dir() {
+                fs::create_dir_all(&entry.target_mount_point)
+                    .await
+                    .with_context(|| {
+                        format!(
+                            "Unable to mkdir all for {}",
+                            entry.target_mount_point.display()
+                        )
+                    })?;
+            }
+
             self.0.push(entry);
         }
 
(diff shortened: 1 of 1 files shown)


src/agent/src/watcher.rs
```diff
@@ -475,6 +488,85 @@ mod tests {
         Ok((storage, src_path))
     }
 
+    #[tokio::test]
+    async fn test_empty_sourcedir_check() {
+        //skip_if_not_root!();
```
bergwolf, line 493: can be dropped.

src/agent/src/watcher.rs
```diff
@@ -475,6 +488,85 @@ mod tests {
...
+        assert_eq!(entries.0.len(), 1);
+
+        assert_eq!(std::fs::read_dir(src_path).unwrap().count(), 0);
+        assert_eq!(std::fs::read_dir(dest_path).unwrap().count(), 0);
+        assert_eq!(std::fs::read_dir(dir.path()).unwrap().count(), 2);
+    }
+
+    #[tokio::test]
+    async fn test_single_file_check() {
+        //skip_if_not_root!();
```
bergwolf, line 530: same here.

bergwolf (approved):
lgtm!

fidencio (approved):
lgtm, thanks @egernst!

@bergwolf, as this is a backport, I'll merge it as it's and hopefully Eric can address your comments in the main branch, carrying it down here to the stable-2.2 after it's approved there.

Thanks both of you!

Issue #2715, opened 2021-09-24
agent-ctl: Implement Linux OCI spec handling in the oci_to_grpc code


liudalibj:
**Which feature do you think can be improved?**

There is no `Linux OCI spec handling` in the oci_to_grpc code.
https://github.com/kata-containers/kata-containers/blob/main/tools/agent-ctl/src/utils.rs#L346

When we use agent-ctl to run containers, without linux namesapce support, the started container process is under the root context on sandbox VM.
To let the started container is running under it's own namespace, we will need  implement Linux OCI spec handling in the oci_to_grpc code

There are FIXME marks in the code https://github.com/kata-containers/kata-containers/blob/main/tools/agent-ctl/src/utils.rs#L346

**How can it be improved?**

Implement Linux OCI spec handling in the oci_to_grpc code

**Additional Information**

Anything else to add?

N/A

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?
Yes

PR #2716, opened 2021-09-24
Implement Linux OCI spec handling

liudalibj:
convert linux field from oci spec to grpc spec
Fixes: #2715

Signed-off-by: Da Li Liu <liudali@cn.ibm.com>


liudalibj:
@bergwolf @jodh-intel would like help to run the pipeline and review this PR?
With this PR we can use agent-ctl to start a kata-container with correct Linux Namespaces.

tools/agent-ctl/src/utils.rs
```diff
@@ -328,6 +337,338 @@ fn process_oci_to_grpc(p: &ociProcess) -> grpcProcess {
     }
 }
 
+fn idmaps_oci_to_grpc(res: &[oci::LinuxIdMapping]) -> protobuf::RepeatedField<grpcLinuxIDMapping> {
```
bergwolf, line 340: The protocol between kata runtime and agent is in ttrpc. So it is a bit wrong in the function naming. Ditto the following functions.

I noted that the original names are wrong as well. Please do not follow them ;) Let's fix them instead.

bergwolf (commented):
Thanks for fix the FIXME! Please note that `agent-ctl` is just for agent functionality validation.

tools/agent-ctl/src/utils.rs
liudalibj: I fix these names, please help to check again. :-)

liudalibj:
> Thanks for fix the FIXME! Please note that `agent-ctl` is just for agent functionality validation.

I am playing some codes with agent, and found that part of my codes can fix the container linux namespace issue which is be marked as FIXME! in agent-ctl.
So I submit this PR, maybe after I play with others codes, I can contribute my codes to the agent core functions.  ^_^

Issue #2725, opened 2021-09-24
workflows: Fix tags attribution in the release.yaml


fidencio:
While releasing kata-containers 2.3.0-alpha1 we've hit some issues as
the tags attribution is done incorrectly.  We want an array of tags to
iterate over, but the currently code is just lost is the parenthesis.

This issue was introduced in a156288c1f8251bdd3231ec765d324f95a3da33e.

PR #2727, opened 2021-09-24
workflows: Fix tag attribution

fidencio:
While releasing kata-containers 2.3.0-alpha1 we've hit some issues as
the tags attribution is done incorrectly.  We want an array of tags to
iterate over, but the currently code is just lost is the parenthesis.

This issue was introduced in a156288c1f8251bdd3231ec765d324f95a3da33e.

Fixes: #2725

Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -100,8 +100,8 @@ jobs:
         run: |
           # tag the container image we created and push to DockerHub
           tag=$(echo $GITHUB_REF | cut -d/ -f3-)
-          tags=$(tag)
-          tags+=$([[ "$tag" =~ "alpha"|"rc" ]] && echo "latest" || echo "stable")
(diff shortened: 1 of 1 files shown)


devimc (approved):
thanks @fidencio  - good catch!

cmaf (approved):

wainersm (approved):
LGTM and works:
```
# tag=rc
# tags=($tag)
# tag="alpha"
# tags+=($([[ "$tag" =~ "alpha"|"rc" ]] && echo "latest" || echo "stable"))
# for t in ${tags[@]}; do echo $t; done
rc
latest
```

fidencio:
Let me merge this one even without the clh test finish as it'll save us some time and the patch is not related to that at all.

Thanks for the reviews.

PR #2732, opened 2021-09-26
qemu: prepare to upgrade qemu version to 6.1.0 for arm

jongwu:
there are 2 issue to upgrade qemu version to 6.1.0 for arm:

1. using the current qemu config will lead to compile error: 

       qemu/build/../tests/plugin/bb.c:134: undefined reference to `qemu_plugin_register_vcpu_tb_trans_cb'
       qemu/build/../tests/plugin/bb.c:135: undefined reference to `qemu_plugin_register_atexit_cb'
       qemu/build/../tests/plugin/bb.c:131: undefined reference to `qemu_plugin_register_vcpu_idle_cb'
       collect2: error: ld returned 1 exit status
 that's why I disable plugin in the first patch.

2. when memory hotplug, kernel will check the alignment size of start hotplug size which is at the end of nvidmm memory size. The current nvdimm size align with 128M, however, in kernel it's 1G which will lead to error.
that's why I port the upstreamed patch to reduce the memory alignment size to 128M.

### DIFF
--- a/tools/packaging/kernel/patches/5.10.x/0001-arm64-sparsemem-reduce-SECTION_SIZE_BITS.patch
+++ b/tools/packaging/kernel/patches/5.10.x/0001-arm64-sparsemem-reduce-SECTION_SIZE_BITS.patch
@@ -1,0 +1,94 @@
+From 5745513881520f48401ac40306ac8cd6ef35beaf Mon Sep 17 00:00:00 2001
+From: Sudarshan Rajagopalan <sudaraja@codeaurora.org>
+Date: Sun, 26 Sep 2021 13:10:41 +0800
+Subject: [PATCH] arm64/sparsemem: reduce SECTION_SIZE_BITS
+
+memory_block_size_bytes() determines the memory hotplug granularity i.e the
+amount of memory which can be hot added or hot removed from the kernel. The
+generic value here being MIN_MEMORY_BLOCK_SIZE (1UL << SECTION_SIZE_BITS)
+for memory_block_size_bytes() on platforms like arm64 that does not override.
+
+Current SECTION_SIZE_BITS is 30 i.e 1GB which is large and a reduction here
+increases memory hotplug granularity, thus improving its agility. A reduced
+section size also reduces memory wastage in vmemmmap mapping for sections
+with large memory holes. So we try to set the least section size as possible.
+
+A section size bits selection must follow:
+(MAX_ORDER - 1 + PAGE_SHIFT) <= SECTION_SIZE_BITS
+
+CONFIG_FORCE_MAX_ZONEORDER is always defined on arm64 and so just following it
+would help achieve the smallest section size.
+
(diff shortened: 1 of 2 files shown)


bergwolf (approved):

jongwu:
thanks @bergwolf . Help review @fidencio @devimc

tools/packaging/kernel/patches/5.10.x/0001-arm64-sparsemem-reduce-SECTION_SIZE_BITS.patch
```diff
@@ -0,0 +1,94 @@
+From 5745513881520f48401ac40306ac8cd6ef35beaf Mon Sep 17 00:00:00 2001
```
Jakob-Naucke, line 1: @jongwu shouldn't the patch be for `tools/packaging/kernel/patches/6.1.x` rather than `tools/packaging/kernel/patches/5.10.x`?

tools/packaging/scripts/configure-hypervisor.sh
```diff
@@ -437,6 +437,8 @@ generate_qemu_options() {
 	# aarch64 need to explictly set --enable-pie
 	if [ -z "${static}" ] && [ "${arch}" = "aarch64" ]; then
 		qemu_options+=(arch:"--enable-pie")
+		# pie is conflict with plugins build for qemu 6.1.0
+		[ "${qemu_version}" == "6.1.0" ] && qemu_options+=(arch:"--disable-plugins")
```
devimc, line 441: this is 6.1.0 specific or for versions >= 6.1.0 ?

tools/packaging/kernel/patches/5.10.x/0001-arm64-sparsemem-reduce-SECTION_SIZE_BITS.patch
jongwu: It's a fixup in kernel part and I think it is more elegant than a fix in qemu side.

Jakob-Naucke: oops, yes. Sorry.

sameo (approved):

gkurz (approved):
So this is upstream linux commit [f0b13ee23241846f6f6cd0d119a8ac8059416ec4](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f0b13ee23241846f6f6cd0d119a8ac8059416ec4).

Definitely better than forcing alignment in QEMU.

fidencio (approved):
lgtm, on behalf of @gkurz! :-)
lgtm, on my own behalf.  Thanks @jongwu!

tools/packaging/scripts/configure-hypervisor.sh
gkurz: I can reproduce this with upstream QEMU. I'll investigate some more for a fix but in the meantime it seems reasonable to disable plugins.

Issue #1693, opened 2021-04-15
Add kubernetes CRI Port-Forward support


gnawux:
In CRI spec, there is a Port-Forward method, which forwards a user request stream to a port of the container. (ref: [the original CRI streaming design document](https://docs.google.com/document/d/1OE_QoInPlVCK9rMAx9aybRmgFiVjHpJCHI9LrfdNM_s/edit#heading=h.4yfjiw58o8d3))

The implementation of this API in runC is `nsenter` + `socat` in brief. For hyper runV, we once did this by `exec` a `socat` in the sandbox but not in any user containers. (as follows)

```
                         +-------------------------------------------+
                         | sandbox                                   |
                         |              +-----------+ +-----------+  |
                         |              | container | | container |  |
                         |              +-----+-----+ +-----------+  |
                         |                    ^                      |
          exec           |  +-------+         |                      |
stream +------------------> | socat +---------+                      |
                         |  +-------+                                |
                         +-------------------------------------------+

```

For kata containers, we have at least two options to implement this

#### Add a new agent API to support this behavior

The agent could just do the stream termination and forward traffic to the port of the sandbox network.

```
                         +-------------------------------------------+
                         | sandbox                                   |
                         |              +-----------+ +-----------+  |
                         |              | container | | container |  |
                         |              +-----+-----+ +-----------+  |
                         |                    ^                      |
          PortForward()  |  +-------+         |                      |
stream +------------------> | agent +---------+                      |
                         |  +-------+                                |
                         +-------------------------------------------+
```

#### Intercept the network traffic of the containers from outside of the sandbox

I think this may lead to different ways for different CNI implementations

Any ideas, guys?

egernst:
Draft implementation: https://github.com/kata-containers/kata-containers/pull/5979

mkulke:
A while ago I wrote [kube-relay](https://github.com/mkulke/kube-relay), which automates deploying a container with socat and creating a port-forwarding to it. this addresses cases in which you want to connect to cluster-accessible destinations from a local development machine. @bpradipt mentioned that this might be helpful as temporary mitigation for this issue, and we could consider adding the tool to the kata repo. If ppl find this useful, I'm open to do that.

burgerdev:
I wrote a thing that should help iterating on a solution for this, with more rapid feedback than redeploying Kata. 

https://github.com/burgerdev/kata-network-simulator

[...]

PR #2733, opened 2021-09-26
CCv0: Use a separate ImageService to work with containerd

wllenyj:
* use an ImageService to implement PullImage
* support CreateContainer, when there is no rootfs in host. 

There are still some problems:

- [ ] 1. The patch `shim-plugins`(https://github.com/containerd/containerd/pull/5817) of containerd has been meraged. But it was found that ImageService could not communicate with TaskService, when I implemented ImageService using shim-plugin. So I commit a pr for this problems (https://github.com/containerd/containerd/pull/6156), it is reviewing.
- [x] 2. [Fixed] The `shim-plugins` will published in the `1.6.0` version. But `1.6.0` version of containerd used `otel` mod is `1.0.x` version. It is very different from the current version, It is used in `pkg/katautils/katatrace/tracing.go` files, and need to be fixed. the pr is WIP (https://github.com/wllenyj/kata-containers/commit/a020882f5b01e4ef3ab80504089df294027d7b1e),  Currently just work around.


src/runtime/virtcontainers/agent.go
```diff
@@ -190,4 +191,6 @@ type agent interface {
 
 	// getAgentMetrics get metrics of agent and guest through agent
 	getAgentMetrics(context.Context, *grpc.GetMetricsRequest) (*grpc.Metrics, error)
+
+	task.ImageService
```
bergwolf, line 195: I would prefer to introducing kata's own image API between runtime and agent instead of relying on containerd's definition. Even if they end up the same, it would allow us to maintain the agent API internally, since it is internal API afterall.

src/runtime/pkg/containerd-shim-v2/image.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func initImageService(ic *plugin.InitContext) (interface{}, error) {
+	i, err := ic.GetByID(plugin.TTRPCPlugin, "task")
+	if err != nil {
+		return nil, errors.Errorf("get task plugin error. %v")
+	}
+	task := i.(*shim.TaskService)
+	s := task.Local.(*service)
+	is := &ImageService{s: s}
```
bergwolf, line 38: Any reason to choose containerd image service instead of cri image service here? Did you discuss it with the containerd maintainers on the shim interface extension?

src/runtime/pkg/containerd-shim-v2/create.go
```diff
@@ -243,6 +243,11 @@ func loadRuntimeConfig(s *service, r *taskAPI.CreateTaskRequest, anno map[string
 }
 
 func checkAndMount(s *service, r *taskAPI.CreateTaskRequest) (bool, error) {
+	// In the confidential computing scenario, there is no Image information on the host,
+	// so there is no Rootfs.
+	if len(r.Rootfs) == 0 {
```
bpradipt, line 248: Should there be a check to verify if `image.ServiceOffload` is set to true ?

src/agent/protocols/protos/image.proto
```diff
@@ -0,0 +1,29 @@
...
+// Image defines the public APIs for managing images.
+service Image {
+    // PullImage pulls an image with authentication config.
+    rpc PullImage(PullImageRequest) returns (PullImageResponse) {}
+}
+
+message PullImageRequest {
+    // Name of the image.
+	string image = 1;
+	string container_id = 2;
```
bpradipt, line 21: Can you please explain why container_id is needed for pulling image ?

src/runtime/virtcontainers/container.go
```diff
@@ -1256,6 +1256,11 @@ func (c *Container) hotplugDrive(ctx context.Context) error {
 	// Check to see if the rootfs is an umounted block device (source) or if the
 	// mount (target) is backed by a block device:
 	if !c.rootFs.Mounted {
+		// In the confidential computing scenario, there is no Image information on the host,
+		// so there is no Rootfs.Source.
+		if c.rootFs.Source == "" {
```
bpradipt, line 1261: Same comment as https://github.com/kata-containers/kata-containers/pull/2733/files#r740123183

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1237,6 +1238,11 @@ func (k *kataAgent) rollbackFailingContainerCreation(ctx context.Context, c *Con
 }
 
 func (k *kataAgent) buildContainerRootfs(ctx context.Context, sandbox *Sandbox, c *Container, rootPathParent string) (*grpc.Storage, error) {
+	// In the confidential computing scenario, there is no Image information on the host,
+	// so there is no Rootfs.Target.
+	if c.rootFs.Target == "" {
```
bpradipt, line 1243: Same comment as https://github.com/kata-containers/kata-containers/pull/2733/files#r740123183

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+        let target_path_oci = format!("oci://{}", oci_path);
+
+        fs::create_dir_all(&manifest_path)?;
+        fs::create_dir_all(&oci_path)?;
+
+        info!(sl!(), "Attempting to pull image {}...", &source_image);
+
+        let mut pull_command = Command::new(SKOPEO_PATH);
+        pull_command
+            .arg("--insecure-policy")
```
bpradipt, line 56: This might be ok for our next demo, but we should think about a proper policy to be used if planning to go with skopeo as the final solution for pulling images.

src/agent/protocols/protos/image.proto
wllenyj: This field was moved from CCv0 to here, which I feel is not used.

src/agent/src/image_rpc.rs
wllenyj: I feel that this will be re-implemented through `ImageManager` in the future.

src/runtime/pkg/containerd-shim-v2/create.go
wllenyj: I got it, it commited by https://github.com/kata-containers/kata-containers/commit/539afba03daa341474a6223a852bed09eb835bc4. 
I will fix.

wllenyj: I have a question, does a host machine need to support starting different instances? 
In other words, do CC instances and Kata instances need to be mixed on the same machine?

sameo: > I have a question, does a host machine need to support starting different instances? In other words, do CC instances and Kata instances need to be mixed on the same machine?

Not for now.

src/agent/protocols/protos/image.proto
sameo: When running from k8s/crictl, we should not need a this one.

src/agent/src/image_rpc.rs
sameo: I think it's fine for now, as @bpradipt said.

sameo:
@wllenyj Could we please start rebasing that PR on top of the CCv0 branch?

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) 2019 alibaba
```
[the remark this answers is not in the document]
wllenyj: fixed

src/runtime/virtcontainers/kata_agent.go
wllenyj: fixed

src/runtime/virtcontainers/container.go
wllenyj: fixed

src/runtime/pkg/containerd-shim-v2/create.go
wllenyj: fixed

wllenyj:
> @wllenyj Could we please start rebasing that PR on top of the CCv0 branch?

Done

sameo:
FWIW, @wllenyj is extracting the agent Image service part of this PR into a separate one at https://github.com/kata-containers/kata-containers/pull/2956.

This way we can separate the part that we fully control (The runtime <-> agent side) from the one that we don't (The containerd <-> runtime side). The former could eventually get merged into kata's main branch while we'll depend on containerd's upstream for the former and we'll have to keep it in the CCv0 branch for probably longer.

jodh-intel:
> FWIW, @wllenyj is extracting the agent Image service part of this PR into a separate one at #2956.

@wllenyj - please can you update this PR to remove the image service? It's **very** confusing imho having multiple open PRs containing the same features in different commits imho.

wllenyj:
> @wllenyj - please can you update this PR to remove the image service? It's **very** confusing imho having multiple open PRs containing the same features in different commits imho.

@jodh-intel @bpradipt @bergwolf @liubin This pr is ready to review, thanks for your comments.

jodh-intel:
@wllenyj - Thanks for the notification, but the branch is conflicted currently.

src/agent/src/rpc.rs
```diff
@@ -124,6 +125,34 @@ pub fn verify_cid(id: &str) -> Result<()> {
...
+            root.path = String::from(rp.to_str().unwrap());
+        }
+    }
+    if let Some(proc) = ori.process.as_mut() {
+        if let Some(image_proc) = image.process.as_ref() {
+            merge_process(proc, image_proc);
+        }
+    }
+}
+
```
sameo, line 155: I find this piece of code a little confusing, mostly because of naming. 
Here is a suggested improvement:

```rust
// Partially merge an OCI process specification into another one.
fn merge_oci_process(target: &mut oci::Process, source: &oci::Process) {
    if target.args.is_empty() && !source.args.is_empty() {
        target.args.append(&mut source.args.clone());
    }

    if target.cwd.is_empty() && !source.cwd.is_empty(){
        target.cwd = String::from(&source.cwd);
    }

    target.env.append(&mut source.env.clone());
}

// When being passed an image name through a container annotation, merge its
// corresponding bundle OCI specification into the passed container creation one.
async fn merge_bundle_oci(&self, container_oci: &oci::Spec) -> Result<()> {
        if let Some(image_name) = container_oci.annotations.get(&ANNO_K8S_IMAGE_NAME.to_string()) {
                if let Some(container_id) = self.sandbox.clone().lock().await.images.get(image_name) {
                     let image_oci_config_path = PathBuf::from(CONTAINER_BASE);
                     image_oci_config_path.push(&container_id);
                     image_oci_config_path.push(&String::from("config.json"));
                     debug!(sl!(), "Image bundle config path: {}", image_oci_config_path);
                
                     let image_oci = oci::Spec::load(&image_oci_config_path).context("load image bundle")?;
 
                     if let Some(container_root) = container_oci.root.as_mut() {
                          if let Some(image_root) = image_oci.root.as_ref() {
                              let mut root_path = PathBuf::from(CONTAINER_BASE);
                              root_path.push(container_id);
                              root_path.push(image_root.path.clone());
                              container_root.path = String::from(root_path.to_str().unwrap());
                          }
                     }

                     if let Some(container_process) = container_oci.process.as_mut() {
                          if let Some(image_process) = image_oci.process.as_ref() {
                              merge_oci_process(container_process, image_process);
                          }
                     }
                }
            }

        Ok(())
}
```

src/agent/src/rpc.rs
```diff
@@ -150,6 +179,16 @@ impl AgentService {
...
+        // Merge OCI Spec for Image
+        if let Some(image_name) = oci.annotations.get(&ANNO_K8S_IMAGE_NAME.to_string()) {
+            if let Some(cid) = self.sandbox.clone().lock().await.images.get(image_name) {
+                let config_path = format!("{}/{}/config.json", CONTAINER_BASE, cid);
+                debug!(sl!(), "merge image bundle path: {}", config_path);
+                let image_oci = oci::Spec::load(&config_path).context("load image bundle")?;
+                merge_oci(cid, &mut oci, &image_oci);
+            }
+        }
+
```
sameo, line 191: With the suggested improvement above, this becomes:

```rust
     self.merge_bundle_oci(&oci)?;
```

src/agent/src/rpc.rs
```diff
@@ -79,6 +79,7 @@ use std::path::PathBuf;
 
 pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
+const ANNO_K8S_IMAGE_NAME: &str = "io.kubernetes.cri.image-name";
```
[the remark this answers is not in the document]
sameo: Update after a private conversation with @wllenyj:

The `io.kubernetes.cri.image-name` annotation is set by containerd and although it's a hackish way to pass down the image name information from containerd to the shim, it already exists, it's upstream and most importantly: It's completely transparent to the end user (no one has to manually add annotations to their pod yaml for that).

So my comments on points 1 and 2 are irrelevant and should not be addressed. Point 3 is still relevant I think: We should not rely on an annotation to retrieve a piece of information that was part of the initial CRI request. That should not be addressed through that PR, but as a longer term point.

src/agent/src/rpc.rs
sameo: I added that implementation to your branch @wllenyj .

Please let me know if it's ok with you.

wllenyj: LGTM

sameo:
@jodh-intel @bergwolf @bpradipt @stevenhorsman All green, CI passed. Could you PTAL?

sameo (approved):

src/agent/src/rpc.rs
```diff
@@ -537,6 +554,54 @@ impl AgentService {
...
+    // corresponding bundle OCI specification into the passed container creation one.
+    async fn merge_bundle_oci(&self, container_oci: &mut oci::Spec) -> Result<()> {
+        if let Some(image_name) = container_oci
+            .annotations
+            .get(&ANNO_K8S_IMAGE_NAME.to_string())
+        {
+            if let Some(container_id) = self.sandbox.clone().lock().await.images.get(image_name) {
+                let mut image_oci_config_path = PathBuf::from(CONTAINER_BASE);
+                image_oci_config_path.push(&container_id);
+                image_oci_config_path.push(&String::from("config.json"));
```
jodh-intel, line 568: Nit: This well-known filename string should really be a `const` (as it's also used in another place in this file).

src/agent/src/rpc.rs
```diff
@@ -124,6 +125,19 @@ pub fn verify_cid(id: &str) -> Result<()> {
...
+fn merge_oci_process(target: &mut oci::Process, source: &oci::Process) {
+    if target.args.is_empty() && !source.args.is_empty() {
+        target.args.append(&mut source.args.clone());
+    }
+
+    if target.cwd.is_empty() && !source.cwd.is_empty() {
+        target.cwd = String::from(&source.cwd);
+    }
+
+    target.env.append(&mut source.env.clone());
```
jodh-intel, line 138: OOI, why no check here as is done for `args` and `cwd`?

sameo:
Testing against the `ali-CCv0` branch.

src/agent/src/rpc.rs
```diff
@@ -537,6 +554,54 @@ impl AgentService {
...
+                            image_oci_config_path
+                        )
+                    })?)
+                    .context("load image bundle")?;
+
+                if let Some(container_root) = container_oci.root.as_mut() {
+                    if let Some(image_root) = image_oci.root.as_ref() {
+                        let mut root_path = PathBuf::from(CONTAINER_BASE);
+                        root_path.push(container_id);
+                        root_path.push(image_root.path.clone());
```
jodh-intel, line 587: You can remove the `mut` if you call `.join()` instead of `.push()`.

jodh-intel (commented):
Thanks @wllenyj - a few comments.

src/agent/src/rpc.rs
wllenyj: `args` and `cwd` are the command to be executed by the container, there can be only one, and it should be mainly input by the user, and only use the one in Image if the user didn't input it.

src/agent/src/rpc.rs
sameo: Fixed, thanks.

src/agent/src/rpc.rs
sameo: Ah, good call :) Fixed as well.

stevenhorsman:
Sorry for the delay in looking, I've been a bit unwell recently. One thing that sprung out to me. Are we ready to update https://github.com/kata-containers/kata-containers/blob/d22706c060e57d2d62dbe5274152939d968624b6/versions.yaml#L198 to point to the new containerd branch during with this change as it looks like we are re-vendoring to use it in the code? Once this is done I'm happy to give it a manual test.

Jakob-Naucke:
We have different failures that I want to see /retest 'ed.
- Fedora: CNI isn't coming up, this looks like disk pressure (but the instance should be freshly provisioned?)
- Ubuntu: CNI is failing (different from Fedora!) due to an apparent version mismatch (maybe tests/CCv0 needs a rebase?)
- Firecracker: Pull from a local registry is failing. This is possibly legit, although @wllenyj has told me this test shouldn't use the guest pull. If this persists, we should revert https://github.com/kata-containers/tests/pull/3679 for CCv0.

stevenhorsman:
Sorry for the delayed response, my availability this week is bad. As a data point I've run the testing of this manually with the `CCv0.sh` script and the existing behaviour seems to work okay (except the bug I raised in https://github.com/kata-containers/kata-containers/issues/3020). Is there any new behaviour that we should test/document for this enabling more e2e `crtictl`/Kubernetes support?

stevenhorsman:
> We have different failures that I want to see /retest 'ed.
> 
> * Fedora: CNI isn't coming up, this looks like disk pressure (but the instance should be freshly provisioned?)
> * Ubuntu: CNI is failing (different from Fedora!) due to an apparent version mismatch (maybe tests/CCv0 needs a rebase?)
> * Firecracker: Pull from a local registry is failing. This is possibly legit, although @wllenyj has told me this test shouldn't use the guest pull. If this persists, we should revert [integration/k8s: Build custom stress image tests#3679](https://github.com/kata-containers/tests/pull/3679) for CCv0.

Jakob, this probably isn't helpful, but when I originally created a `CCv0` fork of containerd a few months ago I had to based it on the `1.5` release branch as I couldn't get anything from `main` (ie 1.6) working with the Kata CI, so I'm not sure if there is anything in `1.6` that is incompatible with Kata's current implementation?

Jakob-Naucke:
> Jakob, this probably isn't helpful, but when I originally created a `CCv0` fork of containerd a few months ago I had to based it on the `1.5` release branch as I couldn't get anything from `main` (ie 1.6) working with the Kata CI, so I'm not sure if there is anything in `1.6` that is incompatible with Kata's current implementation?

@stevenhorsman entirely possible. (Unless someone else feels inclined) I can try and test that today independently, but would you by any chance remember which errors you saw then?

stevenhorsman:
> @stevenhorsman entirely possible. (Unless someone else feels inclined) I can try and test that today independently, but would you by any chance remember which errors you saw then?

Sorry it's a few months ago now, so I can't recall :(

Jakob-Naucke:
@wllenyj I know you're quite busy with the ongoing changes, but I would want to try this. Did you say `crictl run --with-pull` was enough? Because that still tries to decrypt an encrypted image on the host. I am using the ali-CCv0 branch from confidential-containers/containerd and this PR.

wllenyj:
@Jakob-Naucke @stevenhorsman 
Thanks for your help. 

I seem to have found the cause of the problem. When registry.mirrors use localhost, containerd will panic. 
fix pr is: https://github.com/confidential-containers/containerd/commit/776bd31b634e3b2af1d333558f6f5fc59815cc46

Now see if this fix works. /test-fc

In addition, I simply wrote a document. @Jakob-Naucke 
https://github.com/wllenyj/kata-containers/commit/81692efa3bf72b8dfbcb699681f961bfe715f095

thank you all so much.

Jakob-Naucke:
@wllenyj alright, let's see how it goes. In the meantime, I do want to know what happens on #3053.

> In addition, I simply wrote a document.

Oops, missed that. Thank you, will TAL.

wllenyj:
> @wllenyj alright, let's see how it goes. In the meantime, I do want to know what happens on #3053.
> 
> > In addition, I simply wrote a document.
> 
> Oops, missed that. Thank you, will TAL.

When containerd use this config:
```
[plugins."io.containerd.grpc.v1.cri".registry.mirrors]                   
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
      endpoint = ["http://localhost:5000"]                               
```
https://github.com/containerd/containerd/blob/main/pkg/cri/server/image_pull.go#L365-L380
Due to the newTransport function does not initialize the TLSClientConfig field.

stevenhorsman:
@wllenyj -  sorry if I've missed something, but should https://github.com/wllenyj/kata-containers/commit/81692efa3bf72b8dfbcb699681f961bfe715f095 be part of this PR?

Jakob-Naucke:
- @wllenyj fixed the issue we saw on Firecracker (and likely would have seen later in the other jobs). Awesome!
- The issue on Ubuntu (`failed to setup network for sandbox "…": plugin type="bridge" failed (add): incompatible CNI versions; config is "1.0.0", plugin supports ["0.1.0" "0.2.0" "0.3.0" "0.3.1" "0.4.0"]`) is **reproducible** just by upgrading containerd. I shall try also upgrading cni-plugins, if that is it, we should do the same on this PR. It does surprise me we didn't hit this issue on the Firecracker job though (it uses containerd just like the Ubuntu job?)
- The Fedora job is still failing and I think it's due to disk pressure. @snir911 may I bother you for a look? How much disk space would these instances be provisioned with?

Btw, I've made some progress in trying this out, still failing to resolve names (IPs seem to work though)

wllenyj:
@stevenhorsman The documentation is a bit simple, I'll think about whether it's better to open a another pr. Because with my English level, there may be a lot of revisions:)

@Jakob-Naucke There may be other problems, wait for the results of this test and take a look.

Thanks all.

stevenhorsman:
> @stevenhorsman The documentation is a bit simple, I'll think about whether it's better to open a another pr. Because with my English level, there may be a lot of revisions:)
I understand. I'm from England and my English is sloppy, so I'm very impressed by yours! The next couple of days I'm very busy, but if there is anything I can do to help with the doc then please let me know and I'm very happy to work together with you (and it benefits me as I get to learn how to use the new features better!)

wllenyj:
@stevenhorsman That's very great, we can communicate on slack, haha.

Jakob-Naucke:
@wllenyj were you able to resolve domain names with crictl? I can only achieve domain name resolution by manually editing the `/etc/resolv.conf` _after_ boot, which is obviously not practicable. I'll look into k8s otherwise.

bpradipt:
Using this PR I was able to successfully spin up a k8s POD running `containerd` from `ali-CCv0` branch. . The image got downloaded inside the VM.

However using the agent config file `agent.config_file=/etc/kata-config.toml` results in kata-agent crashing.

wllenyj:
@Jakob-Naucke I am very confused, is there any DNS problem with other ci ? about `failed to resolve image "k8s.gcr.io/pause:3.5": not found`

In the CI_JOB=CRI_CONTAINERD_K8S, the ci of cri_containerd is executed, the host pulls the Image.

sameo:
> Using this PR I was able to successfully spin up a k8s POD running `containerd` from `ali-CCv0` branch. . The image got downloaded inside the VM.
> 
> However using the agent config file `agent.config_file=/etc/kata-config.toml` results in kata-agent crashing.

@bpradipt Did the agent crash because `agent.config_file=/etc/kata-config.toml` was part of the kernel cmdline but the file did not exist?

bpradipt:
> > Using this PR I was able to successfully spin up a k8s POD running `containerd` from `ali-CCv0` branch. . The image got downloaded inside the VM.
> > However using the agent config file `agent.config_file=/etc/kata-config.toml` results in kata-agent crashing.
> 
> @bpradipt Did the agent crash because `agent.config_file=/etc/kata-config.toml` was part of the kernel cmdline but the file did not exist?

True, the file did not exist in the rootfs whereas the kernel command line had the stanza.

Jakob-Naucke:
I'm still having issues with crictl (I can't resolve any names even without CC in a crictl Kata container TBH) but I can at least get this to work with k8s. One challenge remaining for me is getting https://github.com/confidential-containers/documentation/pull/7 to work, having trouble forwarding ports due to #1693.

jodh-intel (approved):
Thanks @wllenyj.

lgtm

Jakob-Naucke:
alright, https://github.com/kata-containers/tests/pull/4197 is in. /test (let's do all just to be sure)

PR #2734, opened 2021-09-26
[WIP] cc: agent support for cc

wllenyj:
* add ImageService in agent, use script to pull image and unpack.

```
IMAGE=$1
BUNDLE=$2

WD=/run/images

mkdir -p $WD
cd $WD

OCI_IMAGE=$(echo $IMAGE | awk -F\/ '{print $NF}')
IMAGE_NAME=$(echo $OCI_IMAGE | awk -F: '{print $1}')

skopeo copy docker://$IMAGE oci:$OCI_IMAGE > /dev/null || exit 1

umoci unpack --image $OCI_IMAGE --rootless ${IMAGE_NAME}_bundle > /dev/null

echo -n "$WD/${IMAGE_NAME}_bundle"
```

* CreateContainer need merge image spec.

Signed-off-by: wlleny <wllenyj@linux.alibaba.com>

### DIFF
--- a/src/agent/protocols/build.rs
+++ b/src/agent/protocols/build.rs
@@ -13,6 +13,7 @@ fn main() {
         "protos/health.proto",
         "protos/google/protobuf/empty.proto",
         "protos/oci.proto",
+        "protos/image.proto",
     ];
 
     Codegen::new()
--- a/src/agent/protocols/protos/image.proto
+++ b/src/agent/protocols/protos/image.proto
@@ -1,0 +1,74 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// To regenerate api.pb.go run hack/update-generated-runtime.sh
+syntax = "proto3";
+
+package containerd.task.v2;
+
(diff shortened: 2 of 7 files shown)


bergwolf:
Is this just for discussion or an implementation proposal? It duplicates with `PullImage` API in `agent/protocols/protos/agent.proto`. Also, could you merge your two PRs so that the changes can be reviewed together? Since it is easier to review them and they cannot be merged separately anyway.

stevenhorsman:
+1 to bergwolf's comments.
There is also something weird going on caused by my weekly rebase of the `CCv0` branch on `main`, so sorry about that. It looks like as I can see changes that I delivered coming as part of your PR?
Please take a look at the existing code that is added for this functionality and reconcile the duplications. You can see if fairly easily by [comparing the `CCv0` and `main` branches](https://github.com/kata-containers/kata-containers/compare/main...CCv0#).
I like the new image service and I'm sure that some of your rust code is much better than ours, but as well as removing duplications I think there are some aspects that should be kept from the existing PullImage. For example in your PR you do the pull and unpack by calling a hardcoded script name `/opt/pull_image.sh` which doesn't seem to be checked in? and I'm not sure if a particularly nice way to do it by calling out to a shell script. I'd prefer to see the behaviour done in the rust code like in https://github.com/kata-containers/kata-containers/blob/162e577caba14a5643c0afd36a0f60b5ebdd3bdf/src/agent/src/rpc.rs#L1765-L1820
Thanks!

wllenyj:
Hi, @bergwolf  @stevenhorsman 
Thanks for review.
I will  merge  this pr and re-commit with https://github.com/kata-containers/kata-containers/pull/2733 together. And I will modify osbuilder to support Attestation Agent. 

Now, From the previous comments, the different points of view are:
1. Should the PullImage interface be used as an independent ImageService? or added directly to the shimV2 interface.
2. If PullImage interface be used as an independent ImageService, Should it reuse CRI's ImageService?
3. ImageManager has not been implemented yet, so PullImage in the Agent relies on `skpoeo/umoci` and will be completely replaced by ImageManager in the future. So how to make future replacements smoother? Is it more convenient to coordinate with Attestation Agent ?

Issue #2747, opened 2021-09-28
Optimize func noNeedForOutput and add test cases


liubin:
Optimize func [`noNeedForOutput`](https://github.com/kata-containers/kata-containers/blob/fff82b4ef564c4a16ece8d9779bd43951b49e767/src/runtime/pkg/containerd-shim-v2/utils.go#L114) and add test cases for this func.

PR #2748, opened 2021-09-28
runtime: Optimize func noNeedForOutput and add test cases

liubin:
Optimize func noNeedForOutput and add test cases for this func.

Fixes: #2747


liubin:
@lifupan PTAL

cmaf (approved):

PR #2756, opened 2021-09-28
packaging: Configure QEMU with --enable-pie

gkurz:
We explicitely set the Postion Independant Executlable (PIE) options
in the extra CFLAGS and LDFLAGS that are passed to the QEMU configure
script for all archs. This means that these options are used pretty
much everywhere, including when building the sample plugins under the
test directory. These cannot be linked with -pie and break the build,
as experienced recently on ARM (see PR #2732).

This only broke on ARM because other archs are configured with
--disable-tcg : this disables plugins which are built by default
otherwise.

The --enable-pie option is all that is needed. The QEMU build system
knows which binaries should be created as PIE, e.g. the important
bits like QEMU and virtiofsd, and which ones should not, e.g. the
sample plugins that aren't used in production.

Rely on --enable-pie only, for all archs. This allows to drop the
workaround that was put in place in PR #2732.

Fixes: #2757

### DIFF
--- a/tools/packaging/scripts/configure-hypervisor.sh
+++ b/tools/packaging/scripts/configure-hypervisor.sh
@@ -434,12 +434,10 @@ generate_qemu_options() {
 		qemu_options+=(arch:"--target-list=${arch}-softmmu")
 	fi
 
-	# aarch64 need to explictly set --enable-pie
-	if [ -z "${static}" ] && [ "${arch}" = "aarch64" ]; then
-		qemu_options+=(arch:"--enable-pie")
-		# pie is conflict with plugins build for qemu 6.1.0
-		[ "${qemu_version}" == "6.1.0" ] && qemu_options+=(arch:"--disable-plugins")
-	fi
+	# SECURITY: Create binary as a Position Independant Executable,
+	# and take advantage of ASLR, making ROP attacks much harder to perform.
+	# (https://wiki.debian.org/Hardening)
+	[ -z "${static}" ] && qemu_options+=(arch:"--enable-pie")
 
 	_qemu_cflags=""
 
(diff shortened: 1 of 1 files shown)


tools/packaging/scripts/configure-hypervisor.sh
```diff
@@ -434,12 +434,15 @@ generate_qemu_options() {
...
-		[ "${qemu_version}" == "6.1.0" ] && qemu_options+=(arch:"--disable-plugins")
-	fi
+	# SECURITY: Create binary as a Position Independant Executable,
+	# and take advantage of ASLR, making ROP attacks much harder to perform.
+	# (https://wiki.debian.org/Hardening)
+	case "$arch" in
+	aarch64) [ -z "${static}" ] && qemu_options+=(arch:"--enable-pie") ;;
+	x86_64) [ -z "${static}" ] && qemu_options+=(arch:"--enable-pie") ;;
+	ppc64le) [ -z "${static}" ] && qemu_options+=(arch:"--enable-pie") ;;
+	s390x) [ -z "${static}" ] && qemu_options+=(arch:"--enable-pie") ;;
```
fidencio, line 444: As we're doing this for all supported arches, do we still need the switch?  If so, why?

dgibson: Seconded.  Apart from that LGTM.

devimc (approved):
thanks @gkurz

dgibson (approved):

gkurz:
The failing ARM and s390x CIs are caused by https://github.com/kata-containers/tests/issues/4044 IIUC. This should be fixed with https://github.com/kata-containers/tests/pull/4048.

snir911:
> The failing ARM and s390x CIs are caused by [kata-containers/tests#4044](https://github.com/kata-containers/tests/issues/4044) IIUC. This should be fixed with [kata-containers/tests#4048](https://github.com/kata-containers/tests/pull/4048).

Merged, let's see
/retest-arm
/retest-s390x

gkurz:
> > The failing ARM and s390x CIs are caused by [kata-containers/tests#4044](https://github.com/kata-containers/tests/issues/4044) IIUC. This should be fixed with [kata-containers/tests#4048](https://github.com/kata-containers/tests/pull/4048).
> 
> Merged, let's see /retest-arm /retest-s390x

Still failing but now it seems to be issue https://github.com/kata-containers/kata-containers/issues/2779 (PR https://github.com/kata-containers/kata-containers/pull/2780)

amshinde (approved):

Issue #2768, opened 2021-09-29
Start Sandbox failure when adding IPv6 interface routes


sameo:
An interface route is one that only contains a device (the interface), but no destination, source or gateway:

```shell
ip route add default dev eth0
```

This is a perfectly valid route that some CNIs define and use when setting a k8s pod networking namespace.
When a CNI sets an IPv6 interface route, we're seeing StartSandbox failures due to the agent trying to set up some conflicting routes in the guest:

```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim: Failed to update routes: Failed to add new routes
```

This is due to how currently the kata-agent selects if a route from UpdateRoute is an IPv6 or an IPv4 one:

```rust
let is_v6: bool = is_ipv6(route.get_gateway()) || is_ipv6(route.get_dest());
```

which set `is_v6` to false with any interface route (gateway and destination are empty), including IPv6 ones.

So the kata-agent ends up setting an IPv4 interface route while an IPv6 one is set on the host. That can easily create conflicts with existing IPv4 routes, and make the kata-agent `update_routes` routine fail, which will cascade into a Sandbox start failure.

PR #2769, opened 2021-09-29
Pass the host route IP family to the guest 

sameo:
So that we can properly select the same IP family when setting and updating routes in the guest.

Fixes: #2768


fidencio:
Although this sounds intrusive for a backport, I think it's needed & worth it.  Could someone from @kata-containers/redhat review this one?

sameo:
> Although this sounds intrusive for a backport, I think it's needed & worth it. Could someone from @kata-containers/redhat review this one?

Right, I will backport it once it's approved and CI green.

src/agent/src/netlink.rs
```diff
@@ -312,7 +312,11 @@ impl Handle {
 
         for route in list {
             let link = self.find_link(LinkFilter::Name(&route.device)).await?;
-            let is_v6 = is_ipv6(route.get_gateway()) || is_ipv6(route.get_dest());
+            let is_v6 = if route.get_gateway().is_empty() && route.get_dest().is_empty() {
+                route.get_family() == IPFamily::v6
```
egernst, line 316: Is there any reason we wouldn't want to use get_family directly, regardless if the gateway and destination are not provided?  Are we populating the family as part of the gRPC regardless?

sameo: This is for backward compatibility purposes, unless we're ok with saying the new agent will only work with the corresponding  runtime changes.

src/runtime/cmd/netmon/netmon.go
```diff
@@ -170,7 +170,7 @@ func newNetmon(params netmonParams) (*netmon, error) {
 
 func (n *netmon) cleanup() {
 	os.RemoveAll(n.storagePath)
-	n.netHandler.Delete()
```
c3d, line 173: Is this a change in the interface? If so, could you announce it in the commit message to explain the change?

c3d (approved):
Looks good to me.

Do you have a test case? (I did not see it)

src/runtime/cmd/netmon/netmon.go
sameo: Good point. I added a note on the commit message now.

src/agent/src/netlink.rs
sameo: @egernst I changed my mind and went the cleaner way: We only rely on the Route family field now for switching between v4 and v6. Running this new agent means you'll have to run the fixed runtime as well, but I believe this is acceptable.

Jakob-Naucke (approved):

jodh-intel (approved):
Thanks @sameo.

lgtm. Since @egernst has commented, maybe we should hold off merging until he's acked this too.

src/runtime/go.sum
```diff
@@ -532,8 +532,9 @@ github.com/urfave/cli v1.22.1/go.mod h1:Gos4lmkARVdJ6EkW0WaNv/tZAAMe9V7XWyB60NtX
 github.com/urfave/cli v1.22.2 h1:gsqYFH8bb9ekPA12kRo0hfjngWQjkJPlN9R0N78BoUo=
 github.com/urfave/cli v1.22.2/go.mod h1:Gos4lmkARVdJ6EkW0WaNv/tZAAMe9V7XWyB60NtXRu0=
 github.com/vishvananda/netlink v1.1.0/go.mod h1:cTgwzPIzzgDAYoQrMm0EdrjRUBkTqKYppBueQtXaqoE=
-github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852 h1:cPXZWzzG0NllBLdjWoD1nDfaqu98YMv+OneaKc8sPOA=
 github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852/go.mod h1:twkDnbuQxJYemMlGd4JFIcuhgX83tXhKS2B/PRMpOho=
```
gkurz, line 535: Shouldn't the go.mod line be removed all the same ?

sameo: FWIW, this was autogenerated for me by `go mod`. I'm not sure I want to manually modify that file...

gkurz: Indeed ! And if you remove it, `go mod tidy` consistently puts it back. Thanks, that answers my question.

PR #2771, opened 2021-09-29
runtime: generate different rand sequences

### LINKED ISSUE #2770: use seed() to generate different random sequences
When we create a random user for rootless hypervisor, [the same random sequence](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/create.go#L359) is generated across multiple runs, creating unstable behavior when retrying on transient failures. 

Adding a seed() before solves the problem.

fengwang666:
When we create a random user for rootless hypervisor, the same
random sequence is generated across multiple runs,
creating unstable behavior when retrying on transient failures.

Fixes #2770

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/create.go
+++ b/src/runtime/pkg/containerd-shim-v2/create.go
@@ -10,8 +10,14 @@ package containerdshim
 import (
 	"context"
 	"fmt"
+	containerd_types "github.com/containerd/containerd/api/types"
+	"github.com/containerd/containerd/mount"
+	taskAPI "github.com/containerd/containerd/runtime/v2/task"
+	"github.com/containerd/typeurl"
 	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/rootless"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
 	"math/rand"
 	"os"
 	"os/user"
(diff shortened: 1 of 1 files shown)


src/runtime/pkg/containerd-shim-v2/create.go
```diff
@@ -10,22 +10,22 @@ package containerdshim
...
 	"context"
 	"fmt"
+	containerd_types "github.com/containerd/containerd/api/types"
+	"github.com/containerd/containerd/mount"
+	taskAPI "github.com/containerd/containerd/runtime/v2/task"
+	"github.com/containerd/typeurl"
 	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/rootless"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
```
bergwolf, line 20: could you seperate golang deps, external project deps and kata internal deps, with a blank line between them? Just so that they look more clear.
E.g., 
```
"context"
"fmt"

"github.com/containerd/containerd/mount"
"github.com/pkg/errors"

"github.com/kata-containers/kata-containers/src/runtime/pkg/utils"
```

src/runtime/pkg/containerd-shim-v2/create.go
```diff
@@ -355,6 +357,7 @@ func createVmmUser() (string, error) {
 	// Add retries to mitigate temporary errors and race conditions. For example, the user already exists
 	// or another instance of the runtime is also creating a user.
 	maxAttempt := 5
+	rand.Seed(time.Now().UnixNano())
 	for i := 0; i < maxAttempt; i++ {
 		userName = fmt.Sprintf("kata-%v", rand.Intn(100000))
 		_, err = utils.RunCommand([]string{useraddPath, "-M", "-s", nologinPath, userName, "-c", "\"Kata Containers temporary hypervisor user\""})
```
bergwolf, line 363: `useradd` doesn't really handle races. How is race handled here? It seems possible that a user might still be used for different vmms?

fengwang666: I believe `useradd` locks the `/etc/passwd` and `/etc/group` files. Once a user is created and given to a VM, it will not be reused.

fengwang666:
@devimc Please take a look

devimc (approved):

egernst:
I don't love leaving the path on the host and cleaning up "only as needed."

I don't have a suggestion at the moment, but hoping some other folks can take a look too?

@sameo @bergwolf @amshinde can you PTAL?

PR #2775, opened 2021-09-30
kata-monitor: add index page

### LINKED ISSUE #2292: Add index page for kata monitor
```bash
$ curl localhost:8090
404 page not found
```

Add an index page showing all the sub-commands(pages) will be helpful:

```
$ curl localhost:8090
/metircs: show sandbox metrics
/sandboxes: list all sandboxes
...
```

fgiudici:
Porting of @liubin reference implementation shared in #2292.

Fixes #2292.


src/runtime/cmd/kata-monitor/main.go
```diff
@@ -119,6 +172,14 @@ func main() {
 	logrus.Fatal(svr.ListenAndServe())
 }
 
+func indexPage(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte("Available HTTP endpoints:\n"))
+
+	for i := range endpoints {
+		w.Write([]byte(fmt.Sprintf("%-25s: %s\n", endpoints[i].path, endpoints[i].desc)))
```
liubin, line 179: The parameter `%-25s` depends on the actual output of the index page(the `path` property's length).
And I suggest adding a comment that when a new endpoint is added,  the index page should be checked that whether the output has corrupted.

liubin (approved):

src/runtime/cmd/kata-monitor/main.go
fgiudici: Right! Well, just changed to retrieve the max length dynamically now (🤔  too much for something we will not update that frequently? well, the overhead would be ridiculous btw).
Thanks @liubin  :-)

Jakob-Naucke (approved):
Personally, I'd add a test function for `indexPage`, but maybe this is too mundane.

Issue #2777, opened 2021-09-30
Increase timeout for check in cloud-hypervisor


jcvenegas:
`virtcontainers.clh` `check` may cause some race conditions , as API calls handle by cloud-hypervisor is single thread and at kata start having multple API calls can delay the API response.

PR #2778, opened 2021-09-30
clh: Fix race condition that prevent start pods 

jcvenegas:
The thread monitor will check if the agent and the VMM are alive every
second in a blocking thread. The Cloud hypervisor API server is
single-threaded, if the monitor does a `check()`, while a slow request
is still in progress, the monitor check() method will timeout. The
monitor thread will stop all the shim-v2 execution.

This commit modifies the monitor thread to make it check the status of
the hypervisor after 5 seconds. Additionally, the `check()` method from
cloud-hypervisor will use the method `clh.isClhRunning(timeout)` with a
10 seconds timeout. The monitor function does no timeout, so even if
`hypervisor.check()` takes more 10 seconds, the isClhRunning method
handles errors doing a VmmPing and retry in case of errors until the
timeout is reached.

Reduce the time to the next check to 5 should not affect any functionality,
but it will reduce the overhead polling the hypervisor.

Fixes: #2777

Signed-off-by: Carlos Venegas <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -761,12 +761,18 @@ func (clh *cloudHypervisor) Load(s persistapi.HypervisorState) {
 	clh.state.apiSocket = s.APISocket
 }
 
-func (clh *cloudHypervisor) Check() error {
-	cl := clh.client()
-	ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
-	defer cancel()
+// Check is the implementation of Check from the Hypervisor interface.
+// Check if the VMM API is working.
 
-	_, _, err := cl.VmmPingGet(ctx)
+func (clh *cloudHypervisor) Check() error {
+	// Use a long timeout to check if the VMM is running:
+	// Check is used by the monitor thread(a background thread). If the
+	// monitor thread calls Check() during the Container boot, it will take
+	// longer than usual specially if there is a hot-plug request in progress.
+	running, err := clh.isClhRunning(10)
+	if !running {
+		return fmt.Errorf("clh is not running: %s", err)
+	}
 	return err
 }
 
@@ -1034,8 +1040,6 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 
 	pid := clh.state.PID
 
-	// Check if clh process is running, in case it is not, let's
-	// return from here.
 	if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
 		return false, nil
 	}
@@ -1048,6 +1052,8 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 		_, _, err := cl.VmmPingGet(ctx)
 		if err == nil {
 			return true, nil
+		} else {
+			clh.Logger().WithError(err).Warning("clh.VmmPingGet API call failed")
 		}
 
 		if time.Since(timeStart).Seconds() > float64(timeout) {
--- a/src/runtime/virtcontainers/monitor.go
+++ b/src/runtime/virtcontainers/monitor.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	defaultCheckInterval = 1 * time.Second
+	defaultCheckInterval = 5 * time.Second
 	watcherChannelSize   = 128
 )
 


src/runtime/virtcontainers/sandbox.go
```diff
@@ -1589,7 +1588,8 @@ func (s *Sandbox) ResumeContainer(ctx context.Context, containerID string) error
 
 // createContainers registers all containers, create the
 // containers in the guest and starts one shim per container.
-func (s *Sandbox) createContainers(ctx context.Context) error {
+func (s *Sandbox) createContainers(ctx context.Context) (err error) {
```
jcvenegas, line 1591: katautil.go may be a good place for this function.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2259,3 +2267,11 @@ func (k *kataAgent) getAgentMetrics(ctx context.Context, req *grpc.GetMetricsReq
 
 	return resp.(*grpc.Metrics), nil
 }
+
+//defer func(){ err = errorContext("Error from x", err)}()
+func errorContext(con string, err error) error {
```
liubin, line 2272: How about use [`pkg/errors`](https://github.com/pkg/errors#adding-context-to-an-error) instead of `errorContext` ?

```golang
_, err := ioutil.ReadAll(r)
if err != nil {
        return errors.Wrap(err, "read failed")
}
```

src/runtime/pkg/containerd-shim-v2/wait.go
```diff
@@ -151,7 +151,7 @@ func watchOOMEvents(ctx context.Context, s *service) {
 				// If the GetOOMEvent call is not implemented, then the agent is most likely an older version,
 				// stop attempting to get OOM events.
 				// for rust agent, the response code is not found
-				if isGRPCErrorCode(codes.NotFound, err) || err.Error() == "Dead agent" {
+				if isGRPCErrorCode(codes.NotFound, err) || err.Error() == "DeadAgent" {
```
liubin, line 154: `err.Error() == "DeadAgent"` will not work, since `kata_agent.go` has changed the error to DeadAgent(xxx):
```golang
 err = fmt.Errorf("DeadAgent(%s)", k.deadErr)
```

How about add a util func `isDeadAgentError`:

```golang
func isDeadAgentError(err error) bool {
    return err != nil && strings.HasPrefix(err.Error(), "DeadAgent")
}
```

src/runtime/virtcontainers/clh.go
```diff
@@ -730,12 +732,21 @@ func (clh *cloudHypervisor) load(s persistapi.HypervisorState) {
...
 
-	_, _, err := cl.VmmPingGet(ctx)
+	if err = syscall.Kill(clh.state.PID, syscall.Signal(0)); err != nil {
+		return err
+	}
+
+	_, _, err = cl.VmmPingGet(ctx)
+	if err != nil {
+		err = fmt.Errorf("Pid(%d(%s))", clh.state.PID, err)
+	}
```
likebreath, line 749: Looks like we can reuse `isClhRunning()`, right?

src/runtime/virtcontainers/clh.go
```diff
@@ -736,7 +736,7 @@ func (clh *cloudHypervisor) check() (err error) {
 	defer func() { err = errorContext("clh.check", err) }()
 
 	cl := clh.client()
-	ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
likebreath, line 739: Given your comments that the monitor thread would `check` every seconds every second, while we are giving each check a timeout of `10*1` second. Is it an issue?

likebreath (commented):
@jcvenegas Thank you for the hard work and debugging. Left few questions and comments.

cmaf:
Ping @jcvenegas

src/runtime/virtcontainers/clh.go
```diff
@@ -1034,8 +1040,6 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 
 	pid := clh.state.PID
 
-	// Check if clh process is running, in case it is not, let's
-	// return from here.
```
likebreath, line 1038: Not sure whether you intentionally remove the comments.

likebreath (approved):
@jcvenegas Thank you for the detailed commit message and comments. The PR looks good to me. Only one question about the comment you removed.

Issue #2779, opened 2021-10-01
static check fails on kata-check_*_test.go


Amulyam24:
```
kata-check_ppc64le_test.go:59:16: fieldalignment: struct with 24 pointer bytes could be 16 (govet)
        type testData struct {
                      ^
```

```
[static-checks.sh:360] INFO: Running golangci-lint on /home/katauser/go/src/github.com/kata-containers/kata-containers/src/runtime/cmd/kata-runtime
kata-check_generic_test.go:42:2: SA4017: setCPUtype is a pure function but its return value is ignored (staticcheck)
        setCPUtype(config.HypervisorType)
        ^
```
This came up while testing on ppc64le. I'm surprised on how we didn't see it occur on x86_64 or s390x/arm64(where static checks were enabled recently)
/cc @Jakob-Naucke

PR #2780, opened 2021-10-01
cmd: get return value for setCPUtype

Amulyam24:
Accept and assert the return value in testSetCPUTypeGeneric.

Fixes: #2779

Signed-off-by: Amulya Meka <amulmek1@in.ibm.com>

### DIFF
--- a/src/runtime/cmd/kata-runtime/kata-check_generic_test.go
+++ b/src/runtime/cmd/kata-runtime/kata-check_generic_test.go
@@ -39,7 +39,8 @@ func testSetCPUTypeGeneric(t *testing.T) {
 	_, config, err := makeRuntimeConfig(tmpdir)
 	assert.NoError(err)
 
-	setCPUtype(config.HypervisorType)
+	err = setCPUtype(config.HypervisorType)
+	assert.NoError(err)
 
 	assert.Equal(archRequiredCPUFlags, savedArchRequiredCPUFlags)
 	assert.Equal(archRequiredCPUAttribs, savedArchRequiredCPUAttribs)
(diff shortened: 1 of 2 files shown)


Amulyam24:
> Of course, this is a fairly obviously correct change, but I wonder @Amulyam24 -- this was the only thing failing in the static checks? And you saw this only on ppc64le?

No, I see a couple of other failing tests too now on ppc64le(fixes coming up), such as the field alignment failure. And yes, this error suprisingly came up when testing static-checks on ppc64le. Since the static checks are already being run on x86 and now enabled on s390x and arm as well, I wonder how come we didn't see it there.

Jakob-Naucke:
Yeah. I mean, the static checks are not yet succeeding on ARM/s390x, and the field alignment was in arch-specific code, so it obviously wasn't caught, but what you have got here almost makes me think the GHA checks aren't quite running the same things.

Jakob-Naucke:
> fixes coming up

If they're reasonably small and close to each other, I think it's fine to roll them up in one PR (maybe with separated commits)

Amulyam24:
> > fixes coming up
> 
> If they're reasonably small and close to each other, I think it's fine to roll them up in one PR (maybe with separated commits)

Okay, I'll add another fix to this PR. About another, I'm still investigating.

cmaf (approved):

gkurz:
Now that https://github.com/kata-containers/tests/pull/4048 got merged.

/retest-arm

/retest-s390x

Amulyam24:
The s390x CI failure would be fixed by #2774 
@jongwu, can you PTAL at the static check failures on the ARM CI?

cmaf:
@Amulyam24 can we merge this or are you waiting for #2774?

Amulyam24:
@cmaf , this can be merged. Please go ahead.

Issue #2782, opened 2021-10-01
runtime: clh: Enable `seccomp` from Cloud Hypervisor


likebreath:
Cloud Hypervisor (CLH) provides fine-grained support of seccomp filters for each of its child threads. This feature from CLH was disabled in Kata (back to 1.x) due to two reasons: 1) periodic failures from Kata's CI jobs; 2) noticeable memory foot-print increase. 

Considering the security benefits from using the sccomp feature, let's revisit the issue and try to bring it back to Kata when using CLH.

PR #2783, opened 2021-10-01
virtcontainers: clh: Enable the `seccomp` feature

likebreath:
This patch enables the `seccomp` feature from Cloud Hypervisor which
provides fine-grained allowed syscalls for each of its worker
threads. It brings important security benefits, while would increase
memory footprint.

Fixes: #2782


likebreath:
Run the CI for the second time

/test

src/runtime/virtcontainers/clh.go
```diff
@@ -947,12 +947,6 @@ func (clh *cloudHypervisor) launchClh() (int, error) {
 		args = append(args, "-v")
 	}
 
-	// Disable the 'seccomp' option in clh for now.
-	// In this way, we can separate the periodic failures caused
-	// by incomplete `seccomp` filters from other failures.
-	// We will bring it back after completing the `seccomp` filter.
-	args = append(args, "--seccomp", "false")
```
amshinde, line 954: I am in very much in favor of enabling seccomp for cloud-hypervisor. However we did run into issues last time this feature was enabled for certain workloads to an overly restrictive list of system calls being restricted. 
In favor of adopting this feature for workloads, I think it may be worthwhile to have this as a config option initially set to true by default to easily check for any issues. Eventually, we can remove the config option. 
I'd like to see what others think as well.

likebreath: I like this idea and will update the PR in few days unless there are objections from the community.

likebreath:
Running the metrics CI for the second time (after the memory-footprint of CLH was added).

/test-ubuntu-metrics

likebreath:
@amshinde @jcvenegas @jodh-intel @GabyCT The PR is ready for review and landed (assuming the second/third run of the metrics CI is fine).

Issue #2791, opened 2021-10-06
Remove cri-containerd related things.


liubin:
Through #2754, we can confirm that there are some `cri-containerd` related things that remain in the docs/code, and sometimes may lead to users being confused.

PR #2792, opened 2021-10-06
runtime: use containerd package instead of cri-containerd

liubin:
cri-containerd project has been merged into containerd repo, and
we should not reference it any more in code and docs.

This commit will use containerd package instead of cri-containerd
package.

Fixes: #2791


versions.yaml
```diff
@@ -189,6 +189,16 @@ externals:
     # version older than them.
     version: "v1.5.2"
 
+  containerd:
```
amshinde, line 192: @lifupan Do we need the `cri-containerd` version on line 182. I reckon we can take that out as well as part of this PR.

jodh-intel:
@liubin - branch conflicted.

PR #2802, opened 2021-10-06
WIP | vendor: Unpin dependencies that don't have to be pinned

fidencio:
Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>


jodh-intel:
Hi @fidencio - if this is still needed, please can you add a fixes issue to explain why we are doing this / how we noticed that we could unpin? Also, branch conflicted.

Issue #2805, opened 2021-10-07
Rootful TestCreateSandboxSuccess leaves /run/vc/sbs/777-77-77777777 behind


Jakob-Naucke:
https://github.com/kata-containers/kata-containers/blob/10ec4b133c55396d3ddfa93063350a06d45a7089/src/runtime/pkg/containerd-shim-v2/create_test.go#L33

```
kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2 # rm -rf /run/vc/sbs && go test && ls /run/vc/sbs
...
777-77-77777777
```

this breaks the `tests/integration/stability/soak_parallel_rm.sh` test which checks `/run/vc/sbs`.

/cc @Amulyam24 @fidencio

liubin:
Hi @Jakob-Naucke ,  I think I have found the root cause:

https://github.com/kata-containers/kata-containers/blob/988eb95621b0cea53c78136e1c1b759b64316202/src/runtime/pkg/containerd-shim-v2/create.go#L134

The management server created a socket file and prevented the directory to be cleaned.

I will create a PR to fix this.

PR #2808, opened 2021-10-08
runtime: don't start shim management server in tests

liubin:
Shim management server is running in a go routine, in test mode
this will cause the directory where the listen socket
file(/run/vc/sbs/777-77-77777777/shim-monitor.sock) in leak
after the tests finished.

Fixes: #2805


Jakob-Naucke:
thanks @liubin, /test

Issue #2820, opened 2021-10-12
Refine the usage of guest console and kernel parameters with Cloud Hypervisor


likebreath:
Following the work from @jodh-intel on https://github.com/kata-containers/kata-containers/pull/2751, we want to revisit the guest console and kernel parameter usage in Kata with Cloud Hypervisor, so that we have the best performance in production mode while have the full debugging log in debug mode.

PR #2821, opened 2021-10-12
clh: Refine the usage of guest console and kernel parameters with Cloud Hypervisor 

likebreath:
### virtcontainers: clh: Turn-off serial and virtio-console by default

We will need to have console output from the guest only for debugging
purposes. As a result, we can turn-off both the serial and
virtio-console devices by default for better boot time.

### virtcontainers: clh: Use 'quiet' as the default kernel parameter

The 'quiet' kernel parameter can avoid guest kernel logs while booting,
which can reduce boot time.

Fix: #2820


fidencio:
@likebreath, could you open a backport of this one to the stable-2.2 branch?

fidencio:
2821-

> @likebreath, could you open a backport of this one to the stable-2.2 branch?

I did as https://github.com/kata-containers/kata-containers/pull/2839

Issue #2854, opened 2021-10-15
CCv0 | Add a Makefile target for building & uploading a kata-deploy image for the v0 milestone


fidencio:
As the confidential-containers operator is using kata-deploy, let's provide an image that contains the CCv0 bits on it.

PR #2855, opened 2021-10-15
CCv0 | kata-deploy: Add a basic script to build & upload a CCv0 image

fidencio:
This is a dirty hack, that we should expand later so we can pass one or
n number of repos where we'll upload our images, and use it as part of
the release scripts.

For now, however, let's just do this quick & dirty hack so we can
present the CCv0 demo using the operator, even knowing that the
kubernetes part of the work is not done yet and that the demo itself
will be done connecting to a node and doing all the shenanigans
manually.

Fixes: #2854

### DIFF
--- a/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
+++ b/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
@@ -18,7 +18,7 @@ spec:
           katacontainers.io/kata-runtime: cleanup
       containers:
       - name: kube-kata-cleanup
-        image: quay.io/kata-containers/kata-deploy:latest
+        image: quay.io/kata-containers/kata-deploy-cc:v0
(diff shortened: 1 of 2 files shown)


fidencio:
Please, see: https://quay.io/repository/kata-containers/kata-deploy-cc?tab=tags

fidencio:
When testing this PR, do:
```sh
$ cd ${GOPATH}/src/github.com/kata-containers/kata-containers/
$ docker login quay.io
$ make image
```

That should be enough to build the image and update it to `quay.io/kata-containers/kata-deploy-cc:v0`.  @stevenhorsman has access to the `kata-deploy-cc` repo, and rights to add whoever he thinks that should be there.

Ah, this PR depends on https://github.com/kata-containers/kata-containers/pull/2846 .. meaning, please, rebase the CCv0 content atop of `main` when that PR gets merged.

PR #2874, opened 2021-10-21
CCv0|osbuilder: support to build Centos8.4 rootfs

### LINKED ISSUE #2873: CCv0 | rootfs - Need to add CentOS 8.4 into supported OS list
# Description of problem
- As CentOS, currently rootfs build script only support CentOS 7. But we need to build a guest rootfs image that can be used inside TDX environment. And currently TDX's offical MVP package only supports CentOS 8.4. And we also don't sure whether TDX will support CentOS 7 in future.

# Proposal
- In order to let CCv0 also workable inside TDX, it needs to extend the rootfs build script supported OS list to include CentOS 8.4 also.

wllenyj:
Fixes #2873

Signed-off-by: wanglei01 <wllenyj@linux.alibaba.com>

### DIFF
--- a/tools/osbuilder/rootfs-builder/centos8/Dockerfile.in
+++ b/tools/osbuilder/rootfs-builder/centos8/Dockerfile.in
@@ -1,0 +1,42 @@
+#
+# Copyright (c) 2021 Alibaba Cloud
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM quay.io/centos/centos:stream8
+
+@SET_PROXY@
+
+RUN yum -y update \
+    && yum install -y 'dnf-command(config-manager)' \
+    && yum config-manager --set-enabled powertools \
+    && yum install -y \
+    autoconf \
+    automake \
+    binutils \
+    chrony \
+    coreutils-single \
+    curl \
+    gcc \
+    gcc-c++ \
+    git \
+    glibc-common \
+    glibc-devel \
+    glibc-headers \
+    glibc-static \
+    glibc-utils \
+    libseccomp \
+    libseccomp-devel \
+    libstdc++-devel \
+    libstdc++-static \
+    m4 \
+    make \
+    sed \
+    tar \
+    vim \
+    which \
+    && yum clean all && rm -rf /var/cache/yum
+
+# This will install the proper packages to build Kata components
+@INSTALL_MUSL@
+@INSTALL_RUST@
--- a/tools/osbuilder/rootfs-builder/centos8/config.sh
+++ b/tools/osbuilder/rootfs-builder/centos8/config.sh
@@ -1,0 +1,38 @@
+#
+# Copyright (c) 2021 Alibaba Cloud
+#
+# SPDX-License-Identifier: Apache-2.0
+
+OS_NAME="Centos"
+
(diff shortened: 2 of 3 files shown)


fidencio:
> > But also, since [CentOS Linux goes EOL in 3 months](https://www.centos.org/centos-linux) should we be adding a config for CentOS Stream or a CentOS variant (such as AlmaLinux or Rocky Linux) instead?
> 
> Sorry, I haven't build other CentOS variant, I think it should be possible if their package management tools are the same and there is a basic Docker Image.

You can find the container images at https://quay.io/repository/centos/centos?tab=tags and I think the recommended way would be to follow one of the paths:
* Use CentOS Stream 8
* Use one of the RHEL forks (Alma, Rocky, for instance)
* Use RHEL for developers and its UBI image (which may be for free depending on the use case).

As James mentioned we'll need to evaluate what to do with the CentOS based CIs sooner than later, and I'll take care of opening an issue for that by the beginning of the next week.

jodh-intel:
Any update on this @wllenyj?

Jakob-Naucke:
@jodh-intel yes, the "community" version of RHEL for s390x used to be ClefOS maintained up to version 7, but CentOS 9 appears to be coming (currently in Stream).

jodh-intel:
@Jakob-Naucke - ack. Does this mean we can't move to CentOS Stream currently then? I've just checked Alma and Rocky and it looks like both are only currently available for x86_64 and aarch64.

jodh-intel:
@Jakob-Naucke - Stream 8 seems like the best option generally (since presumably it's the closest to CentOS 8) even though it is architecture restricted currently. I guess this doesn't change the situation much for s390x (see https://github.com/kata-containers/kata-containers/tree/main/tools/osbuilder#platform-distro-compatibility-matrix), but you can indicate your views via an ack/nack on this PR of course.

Jakob-Naucke:
General ack -- I think we should try to move to CentOS 8 Stream overall then. As this is upon `CCv0`, it has some time to settle anyway.

fidencio:
> > You can find the container images at https://quay.io/repository/centos/centos?tab=tags and I think the recommended way would be to follow one of the paths:
> > 
> > * Use CentOS Stream 8
> > * Use one of the RHEL forks (Alma, Rocky, for instance)
> > * Use RHEL for developers and its UBI image (which may be for free depending on the use case).
> > 
> > As James mentioned we'll need to evaluate what to do with the CentOS based CIs sooner than later, and I'll take care of opening an issue for that by the beginning of the next week.
> > EDITED: CentOS Stream 8 used `dnf` as CentOS 8 does.
> 
> like this that you mean ? `FROM quay.io/centos/centos:stream8`

Exactly!

Jakob-Naucke:
so @wllenyj can we move this to stream then?

jodh-intel:
Any update @wllenyj ?

Jakob-Naucke:
might be superseded by #3328/#3341

wllenyj:
@jodh-intel @Jakob-Naucke Merry Christmas.

About #2873, Now it is compatible with ubuntu. So Centos is not necessary.

jodh-intel:
Thanks @wllenyj ! Closing...

PR #2658, opened 2021-09-16
[CCv0] Agent configuration file and API restriction

sameo:
This is a port of upstream PR #2517

PR #2517, opened 2021-08-26
Agent configuration file and API restriction

sameo:
This PR adds 2 new features to the kata agent:

1. The ability to configure it through a configuration file from the root filesystem, referenced from the kernel command line (`agent.config_file`)
2. The ability to block part of the grpc API, through the configuration file.

Fixes #1837

Issue #2386, opened 2021-08-04
initrd does not work with systemd unless built with dracut


Jakob-Naucke:
Yes. rootfs+agent init works, but a Kata container with initrd+systemd fails to start (haven't dug deeper than that yet, but it was not just on s390x). Apparently, initrd+systemd does work when the initrd is built with dracut, but not with the "mainstream" workflow.

_Originally posted by @Jakob-Naucke in https://github.com/kata-containers/kata-containers/issues/2268#issuecomment-882440680_

When launching a (distro-built) initrd with systemd, the logs go: 
```
{"msg":"handling signal","level":"INFO","ts":"2021-08-03T14:34:12.419852193+00:00","subsystem":"signals","source":"agent","name":"kata-agent","pid":"56","version":"0.1.0","signal":"SIGCHLD"}
{"msg":"wait_status","level":"INFO","ts":"2021-08-03T14:34:12.421947101+00:00","subsystem":"signals","pid":"56","version":"0.1.0","source":"agent","name":"kata-agent","wait_status result":"Exited(Pid(61), 0)"}
{"msg":"notify term to close","level":"INFO","ts":"2021-08-03T14:34:12.422046846+00:00","subsystem":"signals","child-pid":"61","source":"agent","name":"kata-agent","version":"0.1.0","pid":"56"}
{"msg":"failed to handle signal","level":"ERRO","ts":"2021-08-03T14:34:12.422136737+00:00","subsystem":"signals","source":"agent","version":"0.1.0","name":"kata-agent","pid":"56","error":"waitpid reaper failed

Caused by: 
    ECHILD: No child processes"}
{"msg":"wait process","level":"INFO","ts":"2021-08-03T14:34:12.425450255+00:00","name":"kata-agent","source":"agent","pid":"56","subsystem":"rpc","version":"0.1.0","exec-id":"foo","container-id":"foo"}
{"msg":"cid foo eid foo waiting for exit signal","level":"INFO","ts":"2021-08-03T14:34:12.425562283+00:00","pid":"56","subsystem":"rpc","version":"0.1.0","name":"kata-agent","source":"agent"}
{"msg":"cid foo eid foo received exit signal","level":"INFO","ts":"2021-08-03T14:34:12.425644027+00:00","subsystem":"rpc","source":"agent","name":"kata-agent","pid":"56","version":"0.1.0"}
[...]

PR #2911, opened 2021-10-27
CCv0: Attestation Agent Integration

fitzthum:
Use the Attestation Agent as a keyprovider for Skopeo/ocicrypt when we pull the container from inside the guest. This requires starting the Attestation Agent, creating an ocicrypt config file, and amending the arguments of the skopeo copy command.

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -83,6 +83,7 @@ pub struct AgentConfig {
     pub endpoints: AgentEndpoints,
     pub supports_seccomp: bool,
     pub container_policy_path: String,
+    pub aa_kbc_params: String,
 }
 
 #[derive(Debug, Deserialize)]
@@ -99,6 +100,7 @@ pub struct AgentConfigBuilder {
     pub tracing: Option<bool>,
     pub endpoints: Option<EndpointsConfig>,
     pub container_policy_path: Option<String>,
+    pub aa_kbc_params: Option<String>,
 }
 
 macro_rules! config_override {
@@ -161,6 +163,7 @@ impl Default for AgentConfig {
             endpoints: Default::default(),
             supports_seccomp: rpc::have_seccomp(),
             container_policy_path: String::from(""),
+            aa_kbc_params: String::from(""),
         }
     }
 }
@@ -190,6 +193,7 @@ impl FromStr for AgentConfig {
         config_override!(agent_config_builder, agent_config, unified_cgroup_hierarchy);
         config_override!(agent_config_builder, agent_config, tracing);
         config_override!(agent_config_builder, agent_config, container_policy_path);
+        config_override!(agent_config_builder, agent_config, aa_kbc_params);
 
         // Populate the allowed endpoints hash set, if we got any from the config file.
         if let Some(endpoints) = agent_config_builder.endpoints {
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -4,8 +4,10 @@
 //
 
 use std::fs;
+use std::io::Write;
 use std::path::Path;
 use std::process::{Command, ExitStatus};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use anyhow::{anyhow, ensure, Result};
@@ -21,6 +23,9 @@ use crate::AGENT_CONFIG;
 const SKOPEO_PATH: &str = "/usr/bin/skopeo";
 const UMOCI_PATH: &str = "/usr/local/bin/umoci";
 const IMAGE_OCI: &str = "image_oci:latest";
+const AA_PATH: &str = "/usr/local/bin/attestation-agent";
+const AA_PORT: &str = "127.0.0.1:50000";
+const OCICRYPT_CONFIG_PATH: &str = "/tmp/ocicrypt_config.json";
 
 // Convenience macro to obtain the scope logger
 macro_rules! sl {
@@ -31,11 +36,15 @@ macro_rules! sl {
 
 pub struct ImageService {
     sandbox: Arc<Mutex<Sandbox>>,
+    attestation_agent_started: AtomicBool,
 }
 
 impl ImageService {
     pub fn new(sandbox: Arc<Mutex<Sandbox>>) -> Self {
-        Self { sandbox }
+        Self {
+            sandbox,
+            attestation_agent_started: AtomicBool::new(false),
+        }
     }
 
     fn pull_image_from_registry(
@@ -43,6 +52,7 @@ impl ImageService {
         cid: &str,
         source_creds: &Option<&str>,
         policy_path: &Option<&String>,
+        aa_kbc_params: &str,
     ) -> Result<()> {
         let source_image = format!("{}{}", "docker://", image);
 
@@ -78,6 +88,15 @@ impl ImageService {
         }
 
         debug!(sl!(), "skopeo command: {:?}", &pull_command);
+        if !aa_kbc_params.is_empty() {
+            // Skopeo will copy an unencrypted image even if the decryption key argument is provided.
+            // Thus, this does not guarantee that the image was encrypted.
+            pull_command
+                .arg("--decryption-key")
+                .arg(format!("provider:attestation-agent:{}", aa_kbc_params))
+                .env("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
+        }
+
         let status: ExitStatus = pull_command.status()?;
 
         if !status.success() {
@@ -118,9 +137,39 @@ impl ImageService {
         Ok(())
     }
 
+    // If we fail to start the AA, Skopeo/ocicrypt won't be able to unwrap keys
+    // and container decryption will fail.
+    fn init_attestation_agent() {
+        let config_path = OCICRYPT_CONFIG_PATH;
+
+        // The image will need to be encrypted using a keyprovider
+        // that has the same name (at least according to the config).
+        let ocicrypt_config = serde_json::json!({
+            "key-providers": {
+                "attestation-agent":{
+                    "grpc":AA_PORT
+                }
+            }
+        });
+
+        let mut config_file = fs::File::create(config_path).unwrap();
+        config_file
+            .write_all(ocicrypt_config.to_string().as_bytes())
+            .unwrap();
+
+        // The Attestation Agent will run for the duration of the guest.
+        Command::new(AA_PATH)
+            .arg("--grpc_sock")
+            .arg(AA_PORT)
+            .spawn()
+            .unwrap();
+    }
+
     async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
         let image = req.get_image();
         let mut cid = req.get_container_id();
+
+        let aa_kbc_params = &AGENT_CONFIG.read().await.aa_kbc_params;
 
         if cid.is_empty() {
             let v: Vec<&str> = image.rsplit('/').collect();
@@ -133,6 +182,18 @@ impl ImageService {
             verify_cid(cid)?;
         }
 
+        if !aa_kbc_params.is_empty() {
+            match self.attestation_agent_started.compare_exchange_weak(
+                false,
+                true,
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            ) {
+                Ok(_) => Self::init_attestation_agent(),
+                Err(_) => info!(sl!(), "Attestation Agent already running"),
+            }
+        }
+
         let source_creds = (!req.get_source_creds().is_empty()).then(|| req.get_source_creds());
 
         // Read the policy path from the agent config
@@ -140,7 +201,7 @@ impl ImageService {
         let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
         info!(sl!(), "Using container policy_path: {:?}...", &policy_path);
 
-        Self::pull_image_from_registry(image, cid, &source_creds, &policy_path)?;
+        Self::pull_image_from_registry(image, cid, &source_creds, &policy_path, aa_kbc_params)?;
         Self::unpack_image(cid)?;
 
         let mut sandbox = self.sandbox.lock().await;


fitzthum:
What exactly is the aa_kbc_kbs parameter? I thought the `kbc_name` and `kbs_uri` were both part of the payload sent by ocicrypt.

jialez0:
@fitzthum This KBC_KBS information is what we call KBC name and KBS URI. Their current format is like `xxx_kbc::<KBS_URI>`. Do you remember? The result of our previous discussion is that these two information may be placed in the configuration file of kata agent, passed by kata agent to ocicrypt-rs, and then passed to AA through keyprovider protocol. in the current ccv0 scheme, this parameter is actually passed directly through the provider parameter in the `skopeo copy` command.

fitzthum:
@jialez0 Are you sure about that? As I recall we decided to move the `kbc_name` and `kbs_uri` from the annotation, but @voidC1326 and @liangzhou121 wanted it to be part of the keyprovider_protocol rather than have it be a startup parameter of the AA. This is in the discussion here https://github.com/containers/attestation-agent/pull/5

Furthermore, if you look at the source, it seems like the `kbc_name` and `kbs_uri` come from the `un_wrap_key` function in `src/grpc/mod.rs` where they are derived from the gRPC request. Am I missing something?

jialez0:
@fitzthum Yeah, that's exactly what I mean. `kbc_name` and `kbs_uri` here are extra parameters from the provider of skopeo, then ocicrypt will add them to keyprovider request. 
PS: My previous GitHub ID was voidC1326, but now it's jialez0 :)

fitzthum:
@jialez0 Good question. We could put `kbc_name` and `kbs_uri` in the agent config. That would mean that the measurement would change each time either of those parameters changes. You would basically need to provide a new initrd each time you move the KBS and two tenants couldn't share the same config. I think that's why we were talking about putting it into the pod annotation. I don't think either of those parameters needs to be measured (a malicious KBS shouldn't be able to do much) and having them be set by the pod annotation might make orchestration easier.

I'm not totally sure what that would entail from an engineering perspective, but I will look into it.

fitzthum:
Btw I will rebase this with the various changes and post a cleaner version in the nearish future.

jialez0:
@fitzthum Okey, my little worry is the engineering feasibility of putting it in the pod annotation. As far as I know, the agent currently reads the pod annotation when the `StartContainer` is called. If we put KBC name and KBS URI in the pod annotation, we may need to advance this reading task to `PullImage`?

fitzthum:
Yeah setting up the parameters via the annotation is definitely more complicated. I guess we can skip that for CCv0 and just use the config file, but let's keep it in mind for the future.

src/agent/src/rpc.rs
```diff
@@ -1737,6 +1774,137 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
...
+    use_aa: bool,
+    agent_started: bool,
+) -> Result<()> {
+    let source_image = format!("{}{}", "docker://", image);
+
+    let manifest_path = format!("/tmp/{}/image_manifest", cid);
+    let target_path_manifest = format!("dir://{}", manifest_path);
+
+    // Define the target transport and path for the OCI image, without signature
+    let oci_path = format!("/tmp/{}/image_oci:latest", cid);
```
jodh-intel, line 1790: - Can you make `image_oci:latest` a `const` as it's hard-coded multiple times?
- Would there ever be a need to *not* use latest? We often see problems with the latest versions so it might be prudent to allow the specification of an alternative image somehow.

src/agent/src/rpc.rs
```diff
@@ -1737,6 +1774,137 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
...
+    let target_path_oci = format!("oci://{}", oci_path);
+
+    fs::create_dir_all(&manifest_path)?;
+    fs::create_dir_all(&oci_path)?;
+
+    info!(sl!(), "Attempting to pull image {}...", &source_image);
+
+    let mut pull_command = Command::new(SKOPEO_PATH);
+    pull_command
+        .arg("--insecure-policy")
```
jodh-intel, line 1800: Please add a comment explaining why this option is needed.

src/agent/src/rpc.rs
```diff
@@ -1737,6 +1774,137 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
...
+        }
+
+        // not 100% sure we actually need this. the arg gets ignored anyway
+        pull_command
+            .arg("--decryption-key")
+            .arg("provider:attestation-agent:offline_fs_kbc::null")
+            .env("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
+
+        //TODO: check if image is really encrypted since skopeo doesn't
+        //      seem to care
```
jodh-intel, line 1821: Needs a GitHub issue URL for tracking this remaining item.

src/agent/src/rpc.rs
```diff
@@ -1737,6 +1774,137 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
...
+
+    if use_aa {
+        if !agent_started {
+            init_attestation_agent();
+        }
+
+        // not 100% sure we actually need this. the arg gets ignored anyway
+        pull_command
+            .arg("--decryption-key")
+            .arg("provider:attestation-agent:offline_fs_kbc::null")
```
jodh-intel, line 1817: ? Maybe a comment would be helpful here too?

src/agent/src/rpc.rs
```diff
@@ -1737,6 +1774,137 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
...
+    let mut config_file = File::create(config_path).unwrap();
+    config_file
+        .write_all(ocicrypt_config.to_string().as_bytes())
+        .unwrap();
+
+    // start AA
+    // should we use tokio for this?
+    Command::new(AA_PATH)
+        .arg("--grpc_sock")
+        .arg(format!("127.0.0.1:{}", AA_PORT))
```
jodh-intel, line 1901: Can you create a `const for `127.0.0.1` which is also hard-coded multiple times?

src/agent/src/rpc.rs
```diff
@@ -1737,6 +1774,137 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
...
+        .spawn()
+        .unwrap();
+
+    //TODO: we should probably clean up the AA on shutdown?
```
jodh-intel, line 1905: - How do we determine the answer to this?
- We need to ensure the shutdown is not blocked indefinitely for the agent tracing scenario.

jodh-intel (commented):
Thanks @fitzthum.

Some comments (some the same as the various other implementations floating around ;)

jodh-intel:
Any update on this @fitzthum? Please can you rebase to resolve the conflicts too.

src/agent/src/rpc.rs
fitzthum: Sure, this is not part of my patch tho. I guess the rebase has kind of confounded this PR.

src/agent/src/rpc.rs
fitzthum: Again, not part of my patch.

src/agent/src/rpc.rs
fitzthum: I've been thinking about this a little bit, and it probably isn't worth it to add anything to Kata here. Instead, I will open an issue on the new `image-rs` project to make sure that copying an image will fail if you provide a keyprovider but the image is not encrypted.

src/agent/src/rpc.rs
jodh-intel: Ack. As mentioned on another #2658, imho we should try to get these CC PRs landed into the `CCv0` branch as soon as we can, close outstanding stale PRs and potentially break the PRs into smaller units to speed up the process. I understand how we've got to the current situation, but it's starting to get extremely confusing, atleast to me.

fitzthum:
Yeah I rebase and try to get this PR back to just the commits that it started as.

jodh-intel:
Ok, thanks @fitzthum.

src/agent/src/rpc.rs
Jakob-Naucke: @fitzthum do we have any cleanup functionality/tasks whatsoever?

src/agent/src/image_rpc.rs
```diff
@@ -121,10 +140,44 @@ impl ImageService {
...
+            .unwrap();
+
+        // The Attestation Agent will run for the duration of the guest.
+        Command::new(AA_PATH)
+            .arg("--grpc_sock")
+            .arg(AA_PORT)
+            .spawn()
+            .unwrap();
+
+        //TODO: Do we need to clean up the AA on shutdown?
```
Jakob-Naucke, line 170: @fitzthum do you want to keep this a TODO for this PR?

wllenyj:
There is a config `image.provision` about Container image decryption keys provisioning in the configuration-qemu.toml.in. 
It is added by this patch https://github.com/kata-containers/kata-containers/commit/539afba03daa341474a6223a852bed09eb835bc4. 
Does it have anything to do with skopeo's decryption-key params ?

src/agent/src/image_rpc.rs
```diff
@@ -121,10 +140,44 @@ impl ImageService {
...
+    }
+
     async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
         let image = req.get_image();
         let mut cid = req.get_container_id();
 
+        let agent_config = AGENT_CONFIG.read().await;
+        let aa_kbc_params = &agent_config.aa_kbc_params;
+        let aa_started = agent_config.confidential_setup_complete;
```
wllenyj, line 179: `let aa_started = !agent_config.confidential_setup_complete;` ?
Or change the naming?

src/agent/src/image_rpc.rs
```diff
@@ -136,9 +189,15 @@ impl ImageService {
             verify_cid(cid)?;
         }
 
+        if aa_kbc_params != "" && aa_started {
```
wllenyj, line 192: `! aa_started` ?  the same as above
and `aa_kbc_params.is_empty()`. Otherwise static-check will fail.

Jakob-Naucke:
> There is a config `image.provision` about Container image decryption keys provisioning in the configuration-qemu.toml.in. It is added by this patch [539afba](https://github.com/kata-containers/kata-containers/commit/539afba03daa341474a6223a852bed09eb835bc4). Does it have anything to do with skopeo's decryption-key params ?

@wllenyj yeah, just recently I was reminded about these options again. We discussed them in #1332 when much about how things would work wasn't really clear.

- `data:///local/key/file`: I think this was added for some vague idea of what would later become the [offline file system key broker client](https://github.com/confidential-containers/attestation-agent/tree/main/src/kbc_modules/offline_fs_kbc) (read from a file, which is secure when the initrd is encrypted, e.g. Secure Execution, PEF IIUC). However, the offline FS KBC is an Attestation Agent module, so this isn't very useful.
- `file:///path/to/bin/fetcher/in/guest`: This _seems_ what the Attestation Agent is now _at first glance_, but I think it was more intended for firmware calls, which is somewhat related to SEV{,-ES}'s pre-attestation. This is currently under PR as an Attestation Agent module, https://github.com/confidential-containers/attestation-agent/pull/19.
- `https://my-key-broker.foo/tenant/<tenant-id>`: This was intended for remote key provision, e.g. through runtime attestation with EAA ([KBC module](https://github.com/confidential-containers/attestation-agent/tree/main/src/kbc_modules/eaa_kbc)), TDX, SEV-SNP. (One might also think about providing keys remotely through static TLS certificates in encrypted images, which would be a bit of a hybrid with the first way).

In short, these configurations are superseded by the Attestation Agent, and in some cases, it might actually be important to have these things measured, i.e. they shouldn't be host-configurable in the first place like this configuration skeleton suggests.
They are generally superseded by the `aa_kbc_params` variable suggested here (which is guest, not host configuration) and the architecture of the respective AA KBC modules.

@sameo am I overlooking something here?

src/agent/src/image_rpc.rs
fitzthum: I can remove it given that it is not really significant for CCv0. I think there is some discussion to be had about the best way to close persistent connections that the AA has started. If we just use the AA for getting container secrets all at once and at the beginning of execution, then we can probably just have the server timeout and not worry about cleaning anything up for the AA.

Jakob-Naucke:
> Is it possible for AA to be started by Systemd, same as `agent` ?

@wllenyj Could be. However, I think we should not cease supporting what is suggested in this PR as long as #2386 is not resolved, as some architectures depend on initrd to run in a TEE.

src/agent/src/image_rpc.rs
fitzthum: If the setup is complete the agent has been started, so I think this name is fine. I changed the `!` below.

fitzthum:
> > Is it possible for AA to be started by Systemd, same as `agent` ?
> 
> @wllenyj Could be. However, I think we should not cease supporting what is suggested in this PR as long as #2386 is not resolved, as some architectures depend on initrd to run in a TEE.

We haven't talked much about guest initrd support for CC, but probably we want to support guests that run the agent as init and don't have systemd.

src/agent/src/image_rpc.rs
wllenyj: Oh, yes,  my mistake,  naming is ok.

src/agent/src/config.rs
```diff
@@ -180,6 +185,7 @@ impl FromStr for AgentConfig {
         config_override!(agent_config_builder, agent_config, server_addr);
         config_override!(agent_config_builder, agent_config, unified_cgroup_hierarchy);
         config_override!(agent_config_builder, agent_config, tracing);
+        config_override!(agent_config_builder, agent_config, aa_kbc_params);
```
wllenyj, line 188: Do we need support to configure this parameter separately? It seems that it can only be configured via config_file now.

Jakob-Naucke: @wllenyj that is what this macro does, it sets the variable to what was set in the config file. See #2517 for the introduction of this.

Jakob-Naucke:
@liangzhou121 did you use this PR for last Thursday's demo? When I try it, the log messages stop, no pull is happening, and later it looks as if the attestation agent is being started with another instance of it already running.

fitzthum:
I think they are using this PR, but only to get the parameters from the config file. I think the demo was using the rust agent patch and skipping over any skopeo-specific encryption stuff.

sameo:
@fitzthum Could we rebase that PR? Or do you want to wait for PR #3035 to be merged first?

src/agent/src/image_rpc.rs
```diff
@@ -136,9 +188,15 @@ impl ImageService {
             verify_cid(cid)?;
         }
 
+        if aa_kbc_params != "" && !aa_started {
+            Self::init_attestation_agent();
+            let mut agent_config = AGENT_CONFIG.write().await;
```
Jakob-Naucke, line 193: @sameo we see this lock failing to be acquired with shim pull, would you have an idea why that happens?

fitzthum: New version sidesteps the agent config by adding an AtomicBool as a field in the image service.

Jakob-Naucke:
I can get this to work with Secure Execution, https://github.com/confidential-containers/documentation/pull/7 works. Execs are blocked etc.

liangzhou121:
> 

@fitzthum @Jakob-Naucke Yes, we only applied this PR partially (only agent config file load changes) in order to pass EAA KBC parameters in our demo. So we didn't encounter the strange error you mentioned.

fitzthum:
> I can get this to work with Secure Execution, [confidential-containers/documentation#7](https://github.com/confidential-containers/documentation/pull/7) works. Execs are blocked etc.

Works with SEV as well.

Jakob-Naucke:
@sameo anything you think needs to be addressed before Thanksgiving?

sameo (approved):

sameo:
> @sameo anything you think needs to be addressed before Thanksgiving?

It LGTM. The only failing check is the snap one and that's because the snap servers seem to be having a bad day. I'm going to merge that one regardless.

PR #2918, opened 2021-10-27
update container type handling

egernst:
Today we assume that if the CRI/upper layer doesn't provide a container
type annotation, it should be treated as a sandbox. Up to this point, a
sandbox with a pause container in CRI context and a single container
(ala ctr run) are treated the same.

For VM sizing and container constraining, it'll be useful to know if
this is a sandbox or if this is a single container.

In updating this, we cleanup the type handling tests and we update the
containerd annotations vendoring.

Fixes: #2926


src/runtime/virtcontainers/pkg/oci/utils_test.go
```diff
@@ -16,8 +16,9 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/cri-o/cri-o/pkg/annotations"
+	ctrAnnotations "github.com/containerd/containerd/pkg/cri/annotations"
 	crioAnnotations "github.com/cri-o/cri-o/pkg/annotations"
+	dockerAnnotations "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/annotations/dockershim"
```
amshinde, line 21: I don't this file existing in out repo yet. Did you forget to include it? Also, should'nt this be an upstream change to dockershim itself? Is that something that you are working on?

liubin (approved):

src/runtime/pkg/containerd-shim-v2/create.go
```diff
@@ -70,7 +70,7 @@ func create(ctx context.Context, s *service, r *taskAPI.CreateTaskRequest) (*con
 	rootfs := filepath.Join(r.Bundle, "rootfs")
 
 	switch containerType {
-	case vc.PodSandbox:
+	case vc.PodSandbox, vc.SingleContainer:
```
jodh-intel, line 73: I'm wondering if a `Debug()` call could be useful somewhere around here to show the `containerType`.

jodh-intel (approved):
Thanks @egernst.

lgtm

sameo (approved):

PR #2927, opened 2021-10-29
Update k8s SR-IOV plugin environment variables to work properly with Kata

dgibson:
The Kubernetes SR-IOV plugin communicates which virtual functions to use to container payloads via environment variables giving PCI addresses.  This works fine with traditional containers, however, with Kata the devices will probably have different PCI addresses within the VM.

Therefore we need to update the environment variables to match the guest rather than the host toplogy.

While we're working on this, make a number of cleanups to the adjustments we need to make to the OCI specification for host vs. guest differences.

### DIFF
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -845,7 +845,7 @@ mod tests {
 
         spec.linux = Some(Linux::default());
 
-        // linux.devices is empty
+        // linux.devices doesn't contain the updated device
         let res = update_spec_devices(
             &mut spec,
             HashMap::from_iter(vec![(
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -500,81 +500,86 @@ impl<T: Into<DevUpdate>> From<T> for SpecUpdate {
 //     container_path: the path to the device in the original OCI spec
 //     update: information on changes to make to the device
 #[instrument]
-fn update_spec_devices(spec: &mut Spec, updates: HashMap<&str, DevUpdate>) -> Result<()> {
+fn update_spec_devices(spec: &mut Spec, mut updates: HashMap<&str, DevUpdate>) -> Result<()> {
     let linux = spec
         .linux
         .as_mut()
         .ok_or_else(|| anyhow!("Spec didn't contain linux field"))?;
+    let mut res_updates = HashMap::<(&str, i64, i64), DevNumUpdate>::with_capacity(updates.len());
 
-    let mut devidx = HashMap::<String, (usize, Vec<usize>)>::new();
-
-    for (i, d) in linux.devices.iter().enumerate() {
-        let mut residx = Vec::new();
-
-        if let Some(linuxres) = linux.resources.as_ref() {
-            for (j, r) in linuxres.devices.iter().enumerate() {
-                if r.r#type == d.r#type && r.major == Some(d.major) && r.minor == Some(d.minor) {
-                    residx.push(j);
-                }
-            }
-        }
-        devidx.insert(d.path.clone(), (i, residx));
-    }
-
-    for (container_path, update) in updates {
-        info!(
-            sl!(),
-            "update_spec_devices() considering device";
-            "container_path" => container_path,
-            "guest_major" => update.num.guest_major,
-            "guest_minor" => update.num.guest_minor,
-        );
-
-        if let Some(idxdata) = devidx.get(container_path) {
-            let dev = &mut linux.devices[idxdata.0];
-            let host_major = dev.major;
-            let host_minor = dev.minor;
-
-            dev.major = update.num.guest_major;
-            dev.minor = update.num.guest_minor;
-            if let Some(final_path) = update.final_path {
-                dev.path = final_path;
-            }
+    for specdev in &mut linux.devices {
+        if let Some(update) = updates.remove(specdev.path.as_str()) {
+            let host_major = specdev.major;
+            let host_minor = specdev.minor;
 
             info!(
                 sl!(),
-                "update_spec_devices(): updating device";
-                "container_path" => container_path,
+                "update_spec_devices() updating device";
+                "container_path" => &specdev.path,
+                "type" => &specdev.r#type,
                 "host_major" => host_major,
                 "host_minor" => host_minor,
-                "updated_path" => &dev.path,
-                "guest_major" => dev.major,
-                "guest_minor" => dev.minor,
+                "guest_major" => update.num.guest_major,
+                "guest_minor" => update.num.guest_minor,
+                "final_path" => update.final_path.as_ref(),
             );
 
-            // Resources must be updated since they are used to identify
-            // the device in the devices cgroup.
-            for ridx in &idxdata.1 {
-                // unwrap is safe, because residx would be empty if there
-                // were no resources
-                let res = &mut linux.resources.as_mut().unwrap().devices[*ridx];
-                res.major = Some(update.num.guest_major);
-                res.minor = Some(update.num.guest_minor);
-
-                info!(
-                    sl!(),
-                    "set resources for resource";
-                    "guest_major" => update.num.guest_major,
-                    "guest_minor" => update.num.guest_minor,
-                );
+            specdev.major = update.num.guest_major;
+            specdev.minor = update.num.guest_minor;
+            if let Some(final_path) = update.final_path {
+                specdev.path = final_path;
             }
-        } else {
-            return Err(anyhow!(
-                "Should have found a device {} in the spec",
-                container_path
-            ));
+
+            if res_updates
+                .insert(
+                    (specdev.r#type.as_str(), host_major, host_minor),
+                    update.num,
+                )
+                .is_some()
+            {
+                return Err(anyhow!(
+                    "Conflicting resource updates for host_major={} host_minor={}",
+                    host_major,
+                    host_minor
+                ));
+            }
         }
     }
+
+    // Make sure we applied all of our updates
+    if !updates.is_empty() {
+        return Err(anyhow!(
+            "Missing devices in OCI spec: {:?}",
+            updates
+                .keys()
+                .map(|d| format!("{:?}", d))
+                .collect::<Vec<_>>()
+                .join(" ")
+        ));
+    }
+
+    if let Some(resources) = linux.resources.as_mut() {
+        for r in &mut resources.devices {
+            if let (Some(host_major), Some(host_minor)) = (r.major, r.minor) {
+                if let Some(update) = res_updates.get(&(r.r#type.as_str(), host_major, host_minor))
+                {
+                    info!(
+                        sl!(),
+                        "update_spec_devices() updating resource";
+                        "type" => &r.r#type,
+                        "host_major" => host_major,
+                        "host_minor" => host_minor,
+                        "guest_major" => update.guest_major,
+                        "guest_minor" => update.guest_minor,
+                    );
+
+                    r.major = Some(update.guest_major);
+                    r.minor = Some(update.guest_minor);
+                }
+            }
+        }
+    }
+
     Ok(())
 }
 
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -511,8 +511,8 @@ fn update_spec_device(
         Ok(())
     } else {
         Err(anyhow!(
-            "Should have found a matching device {} in the spec",
-            vm_path
+            "Should have found a device {} in the spec",
+            host_path
         ))
     }
 }
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -745,7 +745,15 @@ pub async fn add_devices(
     for device in devices.iter() {
         let update = add_device(device, sandbox).await?;
         if let Some(dev_update) = update.dev {
-            dev_updates.insert(&device.container_path, dev_update);
+            if dev_updates
+                .insert(&device.container_path, dev_update)
+                .is_some()
+            {
+                return Err(anyhow!(
+                    "Conflicting device updates for {}",
+                    &device.container_path
+                ));
+            }
         }
     }
 
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -503,82 +503,83 @@ impl<T: Into<DevUpdate>> From<T> for SpecUpdate {
     }
 }
 
-// update_spec_device updates the device list in the OCI spec to make
+// update_spec_devices updates the device list in the OCI spec to make
 // it include details appropriate for the VM, instead of the host.  It
-// is given:
+// is given a map of (container_path => update) where:
 //     container_path: the path to the device in the original OCI spec
 //     update: information on changes to make to the device
 #[instrument]
-fn update_spec_device(
+fn update_spec_devices(
     spec: &mut Spec,
     devidx: &DevIndex,
-    container_path: &str,
-    update: DevUpdate,
+    updates: HashMap<&str, DevUpdate>,
 ) -> Result<()> {
-    // If no container_path is provided, we won't be able to match and
-    // update the device in the OCI spec device list. This is an error.
-    if container_path.is_empty() {
-        return Err(anyhow!("Container path cannot be empty for device"));
-    }
-
-    let linux = spec
-        .linux
-        .as_mut()
-        .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
-
-    info!(
-        sl!(),
-        "update_spec_device() considering device";
-        "container_path" => container_path,
-        "guest_major" => update.num.guest_major,
-        "guest_minor" => update.num.guest_minor,
-    );
-
-    if let Some(idxdata) = devidx.0.get(container_path) {
-        let dev = &mut linux.devices[idxdata.idx];
-        let host_major = dev.major;
-        let host_minor = dev.minor;
-
-        dev.major = update.num.guest_major;
-        dev.minor = update.num.guest_minor;
-        if let Some(final_path) = update.final_path {
-            dev.path = final_path;
+    for (container_path, update) in updates {
+        // If no container_path is provided, we won't be able to match and
+        // update the device in the OCI spec device list. This is an error.
+        if container_path.is_empty() {
+            return Err(anyhow!("Container path cannot be empty for device"));
         }
+
+        let linux = spec
+            .linux
+            .as_mut()
+            .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
 
         info!(
             sl!(),
-            "update_spec_device(): updating device";
+            "update_spec_devices() considering device";
             "container_path" => container_path,
-            "host_major" => host_major,
-            "host_minor" => host_minor,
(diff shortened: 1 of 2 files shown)


src/agent/src/device.rs
```diff
@@ -438,204 +459,174 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
-    let linux = spec
-        .linux
-        .as_mut()
-        .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
+    let devidx = DevIndex::new(spec);
 
-    if !Path::new(vm_path).exists() {
-        return Err(anyhow!("vm_path:{} doesn't exist", vm_path));
-    }
+    for fixup in fixups {
```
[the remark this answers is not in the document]
dgibson: I plan to make some significant rearrangements of this function in subsequent patches, so there's not much point to that at this stage.

src/agent/src/device.rs
```diff
@@ -53,6 +53,27 @@ pub const DRIVER_VFIO_GK_TYPE: &str = "vfio-gk";
 // container as a VFIO device node
 pub const DRIVER_VFIO_TYPE: &str = "vfio";
 
+#[derive(Debug)]
+struct DevSpecFixup {
```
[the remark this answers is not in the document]
dgibson: I don't think so.  That puts the code applying the fixup in the position of explicitly handling a "vfio" and "nonvfio" case, when in fact it doesn't care about the type of device at all, so it's a (minor) layering violation.

I could potentially make the `new` variable an `Option`, or I could certainly make a "short form" constructor for DevSpecFixup that fills in `old` and `new` with the same values.

src/agent/src/device.rs
```diff
@@ -438,204 +459,174 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
         info!(
             sl!(),
-            "change the device from path: {} major: {} minor: {} to vm device path: {} major: {} minor: {}",
-            host_path,
-            host_major,
-            host_minor,
-            dev.path,
-            dev.major,
-            dev.minor,
+            "update_spec_device(): vm_path={}, major: {}, minor: {}\n", fixup.vm, major_id, minor_id
```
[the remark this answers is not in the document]
dgibson: Actually that was basically just a debug message, I'll take it out entirely.

src/agent/src/device.rs
```diff
@@ -438,204 +459,174 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+            let old_minor = dev.minor;
+
+            dev.major = major_id as i64;
+            dev.minor = minor_id as i64;
+            dev.path = fixup.new.to_string();
 
             info!(
                 sl!(),
-                "set resources for device major: {} minor: {}\n", major_id, minor_id
+                "change the device from path: {} major: {} minor: {} to vm device path: {} major: {} minor: {}",
```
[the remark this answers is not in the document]
dgibson: I've removed the other one.

src/agent/src/device.rs
```diff
@@ -438,204 +459,174 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+        }
 
-    info!(
-        sl!(),
-        "update_spec_device(): vm_path={}, major: {}, minor: {}\n", vm_path, major_id, minor_id
-    );
+        let linux = spec
+            .linux
+            .as_mut()
+            .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
```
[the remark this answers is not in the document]
dgibson: It could be, but it's out of scope for the patches so far.  The intention is that the current patch just re-indent the existing update_device_spec() code with a new outer block, without rearranging, for ease of review.  I'll be making further rearrangements in later patches.

src/agent/src/device.rs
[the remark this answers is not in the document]
dgibson: Wait... no I haven't.  That's not in scope for the patches I've written so far, since it's basically just indenting the existing messages in an additional block.

src/agent/src/device.rs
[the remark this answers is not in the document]
dgibson: No, wait, this was an existing message, it's just been re-indented.  Fixing it is not in scope for the patches so far, but I'll look at it as I write the remaining patches.

src/agent/src/device.rs
```diff
@@ -53,14 +53,42 @@ pub const DRIVER_VFIO_GK_TYPE: &str = "vfio-gk";
...
 pub const DRIVER_VFIO_TYPE: &str = "vfio";
 
-#[derive(Debug)]
-struct DevIndexEntry {
-    idx: usize,
-    residx: Vec<usize>,
+#[derive(Debug, Clone)]
+struct DevSpecFixup {
+    // the path to the device in the original OCI spec
+    container: String,
```
jodh-intel, line 59: I'm still slightly confused by these names tbh. How about:

- `container` => `oci_spec_path`
- `vm` => `vm_path` (or `container_path`)
- `update` => `update_path`

jodh-intel (approved):
Thanks for updating @dgibson.

src/agent/src/device.rs
```diff
@@ -437,205 +465,189 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
         info!(
             sl!(),
-            "change the device from path: {} major: {} minor: {} to vm device path: {} major: {} minor: {}",
-            host_path,
-            host_major,
-            host_minor,
-            dev.path,
-            dev.major,
-            dev.minor,
+            "update_spec_device(): vm_path={}, major: {}, minor: {}\n",
```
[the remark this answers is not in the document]
dgibson: Again, this is just a code motion / re-indent, changing the content of the code doesn't belong in the same patch.

src/agent/src/device.rs
dgibson: TBH, I'm really struggling to find good names for these amongst other things.  (`container_path` for the thing that's currently `vm` is definitely wrong though; that path might not appear within the container in any form).

jodh-intel: @dgibson - haha, yes names are always surprisingly hard I find. Anyone else have thoughts on the above?

dgibson: Well, I've done a substantial rework for the next spin.  I think it makes these clearer, but let's see what you think.

jodh-intel (approved):
Thanks @dgibson.

lgtm

dgibson:
@jodh-intel I'm still not done here.. so far it's just cleanups before actually implementing the thing I want to implement.  Which is why it's still `do-not-merge` tagged, and marked as a Draft.  So acking it might be a bit premature here.

jodh-intel:
@dgibson - yeah, I did spot that but acking as a general indication that I'm happy with the progress ;)

dgibson:
The ARM and s390x failures looked to be because those only have Rust 1.47.0 and I'm using some newer features.  Should I rewrite to avoid those, or do we just update to a newer rustc?  Looks like we'd need 1.52.0 for `str::split_once` and 1.53.0 for the implementation of `IntoIterator` on arrays, which is the cause of the errors on `HashMap::from_iter`.

src/agent/src/device.rs
```diff
@@ -543,6 +564,32 @@ fn update_spec_devices(spec: &mut Spec, mut updates: HashMap<&str, SpecUpdate>)
...
     }
 
+    // Correct PCI addresses in the environment
+    if let Some(process) = spec.process.as_mut() {
+        for envvar in process.env.iter_mut() {
+            let (name, val) = envvar
+                .split_once('=')
+                .ok_or_else(|| anyhow!("Malformed OCI env entry {:?}", envvar))?;
+
+            if !name.starts_with("PCIDEVICE_") {
```
mythi, line 574: `PCIDEVICE_` is specific to that one plugin but there are also other env vars for the same purpose (e.g., QAT plugin sets `$QAT<id>`). I guess I could move to `PCIDEVICE_` as well but what's the "spec" for that?

dgibson: Alas, there is no standardized way of conveying this information, so the best we can do is fix up those properties we know about.  That said, I believe there was a proposal to standardize `PCIDEVICE_*` across a number of things.  I've lost track of the document about that, I'll see if I can find it.

With the infrastructure that's been added here, it should be relatively simple to add support for extra formats like QAT, as we discover them, but I don't think that's in scope for this initial series.

mythi: > That said, I believe there was a proposal to standardize `PCIDEVICE_*` across a number of things. 
> I've lost track of the document about that, I'll see if I can find it.

Thanks! I'd be happy to help with the doc and also update my plugin(s) to follow.

zvonkok: Is this a public document? If yes can you please share? Thanks!

dgibson: Ok, managed to dig up the document I was thinking of.  It is public, but it doesn't cover quite what I thought it does.  It does attempt to start standardizing device information more than currently, but it's not  really based on `PCIDEVICE_*` environment variables.  It also appears to be at least somewhat specific to network devices.

https://github.com/k8snetworkplumbingwg/device-info-spec/

zvonkok: @dgibson Thanks for that, yeah network-specific, we (RH, NV, IBM, Intel, ...) started some time ago a project that my address this, https://github.com/container-orchestrated-devices/container-device-interface

marcel-apf (approved):
Weak ack: I agree with the approach, however I don't have the necessary rust skills.

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+    fn from_vm_path<T: AsRef<Path>>(vm_path: T) -> Result<Self> {
+        let vm_path = vm_path.as_ref();
 
-    // If no container_path is provided, we won't be able to match and
-    // update the device in the OCI spec device list. This is an error.
-    if host_path.is_empty() {
-        return Err(anyhow!("Host path cannot empty for device"));
+        if !vm_path.exists() {
+            return Err(anyhow!(
+                "VM device path {} doesn't exist",
```
jodh-intel, line 448: Nit: Might be worth making the format `{:?}` so it gets quoted (which would make it clearer if empty for example).

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+            ));
+        }
+
+        let devid = fs::metadata(vm_path)?.rdev();
+        let guest_major = stat::major(devid);
+        let guest_minor = stat::minor(devid);
+
+        Ok(SpecUpdate {
+            guest_major: guest_major as i64,
+            guest_minor: guest_minor as i64,
```
jodh-intel, line 459: You could simplify this:

```suggestion
        let guest_major = stat::major(devid) as i64;
        let guest_minor = stat::minor(devid) as i64;

        Ok(SpecUpdate {
            guest_major,
            guest_minor,
```

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+            final_path: None,
+            pci: Vec::new(),
```
jodh-intel, line 461: You could auto-generate these if you add `#[derive(Default)]` and change these lines to `..Default::default()`.

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+        })
+    }
+    fn with_final_and_pci<T: AsRef<Path>>(
```
jodh-intel, line 464: Nit: Missing blank line between functions.

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+//     container_path: the path to the device in the original OCI spec
+//     update: information on changes to make to the device
+#[instrument]
+fn update_spec_devices(spec: &mut Spec, mut updates: HashMap<&str, SpecUpdate>) -> Result<()> {
     let linux = spec
         .linux
         .as_mut()
-        .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
+        .ok_or_else(|| anyhow!("Spec didn't contain linux field"))?;
+    let mut resupdates = HashMap::<(&str, i64, i64), SpecUpdate>::with_capacity(updates.len());
```
jodh-intel, line 488: This would be slightly clearer as `res_updates`.

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+                        other_guest
+                    ));
+                }
+            }
+
+            if resupdates
+                .insert((&specdev.r#type, host_major, host_minor), update)
+                .is_some()
+            {
+                return Err(anyhow!(
```
jodh-intel, line 529: Worth adding `specdev` to the returned error?

src/agent/src/device.rs
jodh-intel: @dgibson - could you add a comment in the code with a link to that doc for future reference (to make less likely someone will come along and decide they'd like to change it to `PCI_DEVICE_` or similar ;)

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+fn update_spec_devices(spec: &mut Spec, mut updates: HashMap<&str, SpecUpdate>) -> Result<()> {
     let linux = spec
         .linux
         .as_mut()
-        .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
+        .ok_or_else(|| anyhow!("Spec didn't contain linux field"))?;
+    let mut resupdates = HashMap::<(&str, i64, i64), SpecUpdate>::with_capacity(updates.len());
+    let mut pcimap = HashMap::<pci::Address, pci::Address>::new();
+
+    for specdev in &mut linux.devices {
```
jodh-intel, line 491: This would be more understandable if you could call `update_device()` here (and make it easier to test).

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
+                        "guest_major" => update.guest_major,
+                        "guest_minor" => update.guest_minor,
+                    );
+
+                    r.major = Some(update.guest_major);
+                    r.minor = Some(update.guest_minor);
+                }
+            }
+        }
+    }
```
jodh-intel, line 565: This isn't a huge block but could it be replaced by a call to something like `apply_resource_updates()?` ?

src/agent/src/device.rs
```diff
@@ -436,217 +426,242 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
-        // Resources must be updated since they are used to identify
-        // the device in the devices cgroup.
-        for ridx in &idxdata.residx {
-            // unwrap is safe, because residx would be empty if there
-            // were no resources
-            let res = &mut linux.resources.as_mut().unwrap().devices[*ridx];
-            res.major = Some(major_id as i64);
-            res.minor = Some(minor_id as i64);
+    // Correct PCI addresses in the environment
+    if let Some(process) = spec.process.as_mut() {
```
jodh-intel, line 568: Again, this would shrink if you could move this block into a `set_pci_envvars()` function or similar.

jodh-intel (commented):
Thanks @dgibson - a bunch of comments.

src/agent/src/device.rs
dgibson: First, I'm not sure which of the documents you mean.  But second, neither one actually talks about `PCIDEVICE_` environment variables.  They're both talking about possible new ways of describing devices, which is a great goal, but not useful to interoperating with the SR-IOV plugin we have right now.

src/agent/src/device.rs
dgibson: Fair enough, done.

src/agent/src/device.rs
dgibson: Updated, thanks.

src/agent/src/device.rs
dgibson: I don't believe I can `#[derive(Default)]` since the `guest_major` and `guest_minor` fields have types which don't implement `Default`.

dgibson: No, I was mistaken, `i64` does implement `Default`.  Nonetheless, I don't want to implement `Default` for `SpecUpdate` becuase it would mean picking a default value for `guest_major` and `guest_minor` which doesn't make sense - these must be specifically selected every time.

src/agent/src/device.rs
dgibson: Fixed.

src/agent/src/device.rs
dgibson: Sure, changed.

src/agent/src/device.rs
dgibson: Eh, sure, updated.

dgibson: Oof, wait, I take that back.  At least when combined with splitting out the single device update into a helper function, attempting to print `specdev` here results in a horrible lifetime tangle (the fact that the resource map key points into the oci spec for the `type` field keeps the mutable reference to specdev alive overlapping with the immutable reference we'd need to print it here).

I tried a few ways to work around it, but didn't succeed.  I'm sure it's possible, but not worth the effort for just making an unlikely error message marginally nicer.

src/agent/src/device.rs
dgibson: Ok, I've split most of this block out into a helper, specifically a `SpecUpdate::apply_to_device` method.

dgibson: No... wait... again.  It seems silly to split this out if we don't also split out the resource updates.  But that causes different problems, see comment below.

src/agent/src/device.rs
dgibson: So, it turns out that splitting off the resource updates causes a real lifetime mess.  With the update moved to a separate function, the compiler is not longer able to see that we don't change `specdev.r#type`, which means the immutable borrow of `specdev.r#type` we need for the hash lookup conflicts with the mutable borrow of `r` itself.  I don't think shortening this not-huge function is worth the trouble if disentangling this.

dgibson:
@jodh-intel new spin should address some of your comments.

In addition it makes the handling of PCI corrections more flexible, so they don't have to accompany a device node correction.  This may be important for `vfio_mode=guest-kernel`.

dgibson:
Urgh, dealing with Rust versions turned out to have a lot more complications than I anticipated (see #3042, #3061, https://github.com/kata-containers/tests/issues/4188 and https://github.com/kata-containers/tests/issues/4193).

So for the time being, I'm just modifying this series to not use the newer Rust (or technically, `std` crate) features.

Issue #2868, opened 2021-10-20
Regression for systemd cgroups following last vendoring


c3d:
As very briefly documented in [a comment attached to PR #2332](https://github.com/kata-containers/kata-containers/pull/2332#issuecomment-947463493) and in a [question asked on containerd/cgroups](https://github.com/containerd/cgroups/issues/209#issuecomment-947466225), it looks like there was a regression introduced with the last vendored version of `containerd/cgroups` regarding the `systemd` case.

@fgiudici is presently investigating more in depth, as it appears to have broken metrics at least in the OpenShift case.

This is distantly related to #2539 insofar as it seems to be impacted by the same vendored code.

c3d:
From @fgiudici:

The cgroups creation complying with the systemd cgroups management has been dropped .
See [kata code](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/pkg/cgroups/cgroups.go#L125)

What we need seems to be to create the cgroups as explained [in the containerd/cgroups README](https://github.com/containerd/cgroups#create-with-systemd-slice-support)

c3d:
From private exchanges between @fgiudici and myself, it looks like a regression introduced by revendoring:

The cgroups creation (through `containerd/cgroup` code) seems now cgroupfs compliant only (coming to this statement following the links in the above comment)).

Yep, it is a regression. @sameo  changed the cgroups library we used this is why not trivial to spot.
[Previous library had a boolean flag](https://github.com/kata-containers/kata-containers/blob/stable-2.1/src/runtime/virtcontainers/pkg/cgroups/manager.go#L128) to control the creation of the cgroup in cgroupfs vs systemd way, managing all of this transparently using an interface and different back-ends implementing the interface.
Now, with the containerd cgroup code, we have to pass the right hierarchy type (i.e. pass `cgroups.Systemd`) and optionally pre-parse the cgroup path (to comply with the systemd slice format) based on the cgroup driver in use (cgroupfs vs systemd) as showed in the links in the above comment. Samuel just set it statically the cgroupfs way.

egernst:
Francesco, Christophe -- is there an e2e / integration test in place that should've caught this?  (I expect there wasn't?) Can you share a test for reproducing the failure?

PR #2933, opened 2021-10-29
WIP: Fix systemd cgroup

gkurz:
This is an early tentative fix for the recent regression with systemd cgroups.

Changelogs should be written. First two patches should be submitted to upstream containerd, but it is usable and brings the metrics back.

Great thanks to @snir911 for providing a lot of valuable infos and pieces of code.

Fixes: #2868


snir911 (commented):
Thanks Greg, I'd send the prepend patch in separate as and also mention 9949daf4dcba2d39cd0de6b4dc425570b2648cc1 which added it before

additionally this change also requires[ doc update ](https://github.com/kata-containers/kata-containers/blob/main/docs/design/host-cgroups.md)

fidencio:
@sameo, may I bother you to take a look at this one?

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -179,7 +191,14 @@ func (c *Cgroup) Update(resources *specs.LinuxResources) error {
...
 func (c *Cgroup) MoveTo(path string) error {
-	newCgroup, err := cgroups.Load(cgroups.V1, cgroups.StaticPath(path))
+	var err error
+	var newCgroup cgroups.Cgroup
+
+	if !IsSystemdCgroup(path) {
+		newCgroup, err = cgroups.Load(cgroups.V1, cgroups.StaticPath(path))
+	} else {
+		newCgroup, err = cgroups.Load(cgroups.Systemd, cgroups.Slice(getSliceName(path), getUnitName(path)))
+	}
```
sameo, line 201: This is:

```go
newCgroup, err := Load(path)
```

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -116,13 +116,18 @@ func sandboxDevices() []specs.LinuxDeviceCgroup {
...
 	cgroupPath, err := ValidCgroupPath(path, IsSystemdCgroup(path))
 	if err != nil {
 		return nil, err
 	}
 
-	cgroup, err := cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupPath), resources)
+	if !IsSystemdCgroup(path) {
+		cgroup, err = cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupPath), resources)
+	} else {
+		cgroup, err = cgroups.New(cgroups.Systemd, cgroups.Slice(getSliceName(cgroupPath), getUnitName(cgroupPath)), resources)
```
sameo, line 129: It would be cleaner to factorize the `Hierarchy` and `path` building part into a common routine. Something from `pkg/cgroups/utils.go` like:

```go
func cgroupHierarchy(path string) (cgroups.Hierarchy, cgroups.Path) {
         if !IsSystemdCgroup(path) {
                 return cgroups.V1, cgroups.StaticPath(path)
         } else {
                 return cgroups.Systemd, cgroups.Slice(getSliceName(path), getUnitName(path))
         }
}
```

and then the above code would become simpler to read and maintain:

```go
     cgroup, err := cgroups.New(cgroupHierarchy(cgroupPath))
```

src/runtime/vendor/github.com/containerd/cgroups/systemd.go
```diff
@@ -105,13 +106,22 @@ func (s *SystemdController) Create(path string, _ *specs.LinuxResources) error {
...
+	if strings.HasSuffix(name, ".slice") {
+		properties = append(properties,
+			systemdDbus.PropWants(slice),
+		)
+	} else if strings.HasSuffix(name, ".scope") {
+		properties = append(properties,
+			systemdDbus.PropSlice(slice),
+			systemdDbus.PropPids(uint32(os.Getpid())),
+		)
+	}
```
sameo, line 128: Putting this comment here but it's not directly to the above piece of code...

1. Have we sent a patch upstream to fix that?
2. I'd prefer if we'd create a temporary `containerd/cgroups` fork that carries those 2 patches and use that for as our vendored package (through the `replace` function in `runtime/go.mod`). Once/if those patches get accepted upstream we can switch back to the upstream version. If we agree with that approach we can quickly create a `cgroups` fork under the kata-containers organization.

sameo (commented):
Thanks for the patch @gkurz . I think with some minor changes, we can quickly get that one in, see my comments.

snir911:
Closing, continuing @gkurz work in https://github.com/kata-containers/kata-containers/pull/2959
@sameo comments were resolved there

Issue #2936, opened 2021-10-29
    virtcontainers: clh: Guest kernel uses the "serial" as the default console by default 


likebreath:
We observed that the guest Linux kernel picks its own default console, e.g. `ttyS0` (the serial console), when we don't specify the console parameter expicitly. For this reason, the guest kernel will try to read and write to the serial port that is not emulated by the VMM. It brings several unnecessary PIO/ VMM exits that slow-down the boot time. Also, we don't have control over the guest that may read/write to the console that can further introduce more PIO and slow-down the container. As a trade-off, using the virtio-console `Null` mode along with setting the guest kernel console explicitly as `hvc0` is the best way for steady performance.

likebreath:
After additional debugging, we confirmed that the handful of PIOs (always 3 write and 3 read) from serial port while booting the guest is a normal behavior for Linux. It convinced us to keep the current usage of console with Kata+clh, e.g. turn-off both serial and virtio-console in non-debug node. Closing the issue.

PR #2751, opened 2021-09-28
clh: Enable guest userland output

jodh-intel:
Fix guest userland logging for Cloud Hypervisor (CLH) by:

- Disabling CLH's serial console.
- Enabling CLH's console.
- Booting the kernel specifying a para-virtualised console device.

Previously, CLH's serial console was used, resulting in only
kernel boot output and CLH's own debug output being logged.

Switching to using the CLH console enables userland logging: systemd and
it's jobs, such as the Kata agent.

Fixes: #2726.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #2937, opened 2021-10-29
virtcontainers: clh: Use virtio-console `Null` mode by default

likebreath:
We observed that the guest Linux kernel picks its own default console,
e.g. `ttyS0` (the serial console), when we don't specify the console
parameter expicitly. For this reason, the guest kernel will try to read
and write to the serial port that is not emulated by the VMM. It brings
several unnecessary PIO/ VMM exits that slow-down the boot time. Also,
we don't have control over the guest that may read/write to the console
that can further introduce more PIO and slow-down the container. As a
trade-off, using the virtio-console `Null` mode along with setting the
guest kernel console explicitly as `hvc0` is the best way for steady
performance.

Fixes: #2936

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -269,6 +269,9 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, networkNS N
 	} else {
 		// start the guest kernel with 'quiet' in non-debug mode
 		params = append(params, Param{"quiet", ""})
+		// Explicitly set console to 'hvc0' in non-debug mode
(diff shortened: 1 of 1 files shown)


devimc (approved):

jodh-intel (changes requested):
Hi @likebreath - this fails if you enable hypervisor debug:

```
ctr: failed to create shim: error: 500  reason: VmBoot(VmBoot(ConfigValidation(DoubleTtyMode))): unknown
```

There is some more detail about that on https://github.com/kata-containers/kata-containers/pull/2751.

likebreath:
See comment: https://github.com/kata-containers/kata-containers/issues/2936#issuecomment-965633218

PR #2908, opened 2021-10-26
osbuilder: Build attestation-agent and friends

Jakob-Naucke:
When the environment variable $AA_KBC is set, skopeo, umoci, and
attestation-agent (with that KBC) are built inside the guest build
container and installed to the guest rootfs. The respective build- and
runtime dependencies are added. This respects the (existing) $LIBC
variable (gnu/musl) and avoids issues with glibc mismatches. (musl
depends on https://github.com/containers/attestation-agent/pull/20)
At this time, only Ubuntu is supported, due to issues with building
Golang applications on Fedora/CentOS.

Fixes: #2907
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

/cc @sameo @fitzthum @davidhay1969 @bpradipt

Issue #2682, opened 2021-09-20
CCv0 | agent: PoC for adding verify signed image support


stevenhorsman:
### Problem

As part of the [CCv0 plan](https://docs.google.com/spreadsheets/d/1M_MijAutym4hMg8KtIye1jIDAUMUWsFCri9nq4dqGvA/edit#gid=0) we've implemented pulling and image into the Kata guest. As part of feature 1b) we want to support verifying the signature of the image, if it has been signed. This helps to validate where the container image came from and ensure that it hasn't been tampered with.

### Initial Proposal

There are a few different options we can take with the implementation of container verification, and I've listed what I can come up with in in [comment #1](https://github.com/kata-containers/kata-containers/issues/2682#issuecomment-925652415). Whilst trying to remember that this is a PoC for a limited demo, I've suggested that we split the approach into a short-term one in this issue and later follow up with a solution that integrates better with the attestation agent approach.

# Tasks
- [x] Create some test images and push them to `quay.io/kata-containers/confidential-containers`
  - [x] A busybox image that is signed and we keep the gpg key and signature for
  - [x] A busybox image that is not signed
  - [x] An nginx image that is signed, but with a different gpg key that we don't keep and signature that we do keep
- [x] Upload the extracted signature and gpg key into the kata containers git repo
- [x] Modify the build to add the signature to the default sigstore, the public gpg key and  a custom `policy.json` file that cross references the gpg key and the quay repo into the image
- [x] Modify the skopeo pull command in `rpc.rs` to use the custom policy.json with the `--policy` flag
  - [x] Basic hard-coded implementation working
  - [x] Investigate if we can move the 2-stage copy during pull image after we have built in verification
  - [x] Investigate (optionally) passing in the policy file through the agent configuration
  - [x] Check the TODOs to see if there is anything else we can tidy
- [x] Test the 4 images produce the correct results: see acceptance criteria 
- [x] Rebase the changes on the upstream kata-containers/CCv0 branch
- [x] Update the [CCv0 doc](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/how-to-build-and-test-ccv0.md) with how to use/demo this capability
- [x] Test the changes still work
- [x] Create the PR and get the community CI tests running
- [x] Merge the changes

## Acceptance Criteria

**Scenario:** Create an image bundle from a non-signed image, on an 'unprotected registry' works
[...]

PR #2956, opened 2021-11-03
CCv0|agent: Separate the ImageService from the AgentService

wllenyj:
Use a separate ImageService to support PullImage.

Signed-off-by: wllenyj <wllenyj@linux.alibaba.com>


liubin:
Hi @wllenyj , is there any plan that add more messages in the `image.proto` in the future?

src/agent/protocols/protos/image.proto
```diff
@@ -0,0 +1,29 @@
...
+service Image {
+    // PullImage pulls an image with authentication config.
+    rpc PullImage(PullImageRequest) returns (PullImageResponse) {}
+}
+
+message PullImageRequest {
+    // Name of the image.
+	string image = 1;
+	string container_id = 2;
+	string source_creds = 3;
```
jodh-intel, line 22: What is the format of this field?

src/agent/protocols/protos/image.proto
```diff
@@ -0,0 +1,29 @@
...
+
+// Image defines the public APIs for managing images.
+service Image {
+    // PullImage pulls an image with authentication config.
+    rpc PullImage(PullImageRequest) returns (PullImageResponse) {}
+}
+
+message PullImageRequest {
+    // Name of the image.
+	string image = 1;
```
jodh-intel, line 20: Is this just a name or does it also include a version number? If not, how do you know what version you are pulling?

src/agent/protocols/protos/image.proto
```diff
@@ -0,0 +1,29 @@
...
+	string container_id = 2;
```
jodh-intel, line 21: The name of an *existing* (or about-to-be-created) container ID?

src/agent/protocols/protos/image.proto
```diff
@@ -0,0 +1,29 @@
...
+	string source_creds = 3;
+}
+
+message PullImageResponse {
+    // Reference to the image in use. For most runtimes, this should be an
+    // image ID or digest.
```
jodh-intel, line 27: What is the difference between an "image ID" and a "digest"? How will the runtime know the difference if I call my "image ID" `7745bf77-8be3-4164-a503-18d411c666b1` for example?

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+    sandbox: Arc<Mutex<Sandbox>>,
+}
+
+impl ImageService {
+    pub fn new(sandbox: Arc<Mutex<Sandbox>>) -> Self {
+        Self { sandbox }
+    }
+
+    fn pull_image_from_registry(image: &str, cid: &str, source_creds: &Option<&str>) -> Result<()> {
+        let source_image = format!("{}{}", "docker://", image);
```
jodh-intel, line 40: Why are we hard-coding `docker` here?

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+        let target_path_oci = format!("oci://{}", oci_path);
+
+        fs::create_dir_all(&manifest_path)?;
+        fs::create_dir_all(&oci_path)?;
+
+        info!(sl!(), "Attempting to pull image {}...", &source_image);
+
+        let mut pull_command = Command::new(SKOPEO_PATH);
+        pull_command
+            .arg("--insecure-policy")
```
jodh-intel, line 56: This needs a comment explaining why we are using an insecure option for Confidential Computing imho ;)

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+            .arg("copy")
+            .arg(source_image)
+            .arg(&target_path_manifest);
+
+        if let Some(source_creds) = source_creds {
+            pull_command.arg("--src-creds").arg(source_creds);
+        }
```
jodh-intel, line 63: Why are source credentials not mandatory? This needs documenting in the proto file.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        // To save space delete the manifest.
+        // TODO LATER - when verify image is added, this will need moving the end of that, if required
+        fs::remove_dir_all(&manifest_path)?;
+        Ok(())
+    }
+
+    fn unpack_image(cid: &str) -> Result<()> {
+        let source_path_oci = format!("/tmp/{}/image_oci:latest", cid);
```
jodh-intel, line 93: `image_oci:latest` is also used on line 46, so please can you create a `const` for this?

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+        let target_path_bundle = format!("{}{}{}", CONTAINER_BASE, "/", cid);
```
jodh-intel, line 94: We should really be using `Path` to build up paths.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+#[async_trait]
+impl protocols::image_ttrpc::Image for ImageService {
+    async fn pull_image(
+        &self,
+        _ctx: &ttrpc::r#async::TtrpcContext,
+        req: image::PullImageRequest,
+    ) -> ttrpc::Result<image::PullImageResponse> {
+        //trace_rpc_call!(ctx, "pull_image", req);
```
jodh-intel, line 148: Commented out code.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+            pull_command.arg("--src-creds").arg(source_creds);
+        }
+
+        let status: ExitStatus = pull_command.status()?;
+
+        if !status.success() {
+            return Err(anyhow!(format!("failed to pull image: {:?}", status)));
+        }
+
+        // Copy image from one local file-system to another
```
jodh-intel, line 71: Can you explain why we are having to do this copy?

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+        fs::remove_dir_all(&manifest_path)?;
+        Ok(())
+    }
+
+    fn unpack_image(cid: &str) -> Result<()> {
+        let source_path_oci = format!("/tmp/{}/image_oci:latest", cid);
+        let target_path_bundle = format!("{}{}{}", CONTAINER_BASE, "/", cid);
+
+        info!(sl!(), "cid is {:?}", cid);
+        info!(sl!(), "target_path_bundle is {:?}", target_path_bundle);
```
jodh-intel, line 97: It's best to have a single log call where possible. You could also consider using structured logging, something like:

```rust
info!(sl!(), "doing something"; "cid" => cid, "target-bundle-path" => target_path_bundle);
```

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
+        let image = req.get_image();
+        let mut cid = req.get_container_id();
+
+        if cid.len() == 0 {
+            let v: Vec<&str> = image.rsplit('/').collect();
+            if v[0].len() != 0 {
+                cid = v[0]
+            }
```
jodh-intel, line 127: It looks like `cid` could still be blank, so I think we need a check after this block so we can return an error in that case.

jodh-intel (commented):
Thanks @wllenyj - a few comments.

Also, it would be great if you could add some unit tests (which might require a bit of refactoring, but would allow some of the failure scenarios to be exercised).

sameo:
> Hi @wllenyj , is there any plan that add more messages in the `image.proto` in the future?
> 
> And this PR modified files that in #2733 , will that to be closed?

@liubin No, #2733 will only contain the integration part with the containerd image API, but it wont be closed.

liubin:
> > Hi @wllenyj , is there any plan that add more messages in the `image.proto` in the future?
> > And this PR modified files that in #2733 , will that to be closed?
> 
> @liubin No, #2733 will only contain the integration part with the containerd image API, but it wont be closed.

Thank you for your comment @sameo , is it right that this files will be "removed" from #2733 ?

sameo:
> > @liubin No, #2733 will only contain the integration part with the containerd image API, but it wont be closed.
> 
> Thank you for your comment @sameo , is it right that these files will be "removed" from #2733 ?

Correct. Once we merge that one here, @wllenyj would rebase #2733 on top of it and the files will be removed from there.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        let source_creds = (!req.get_source_creds().is_empty()).then(|| req.get_source_creds());
+
+        Self::pull_image_from_registry(image, cid, &source_creds)?;
+        Self::unpack_image(cid)?;
+
+        let mut sandbox = self.sandbox.lock().await;
+        sandbox.images.insert(String::from(image), cid.to_string());
```
liubin, line 136: `images` is written but not read, is it for future use?

src/agent/protocols/protos/image.proto
wllenyj: This field is moved from the previous version of the agent.proto file. For compatibility, the field name has not been modified. I have added a comment explaining this field.

src/agent/protocols/protos/image.proto
wllenyj: This field is just to uniquely represent an image, it will not be processed at runtime, it will be returned to k8s as is, k8s will use this value to call CreateContainer afterwards, then the value will be passed to agent again, this time we can use the value to hash index the Image.
PullImage:
https://github.com/kubernetes/kubernetes/blob/a2960dc2775848f3b54fb288650e02b4f691743b/pkg/kubelet/images/image_manager.go#L89
CreateContainer
https://github.com/kubernetes/kubernetes/blob/a2960dc2775848f3b54fb288650e02b4f691743b/pkg/kubelet/kuberuntime/kuberuntime_container.go#L177

src/agent/src/image_rpc.rs
wllenyj: This should be the way scpo is used, as I understand it. Here is the help message it prints.
> Supported transports:
> containers-storage, dir, docker, docker-archive, docker-daemon, oci, oci-archive, ostree, tarball

src/agent/protocols/protos/image.proto
wllenyj: Added comments

src/agent/protocols/protos/image.proto
wllenyj: Added comments

src/agent/src/image_rpc.rs
wllenyj: I am not very clear about this. In my test, it can work without this option, and this option may be removed in the future.

src/agent/src/image_rpc.rs
wllenyj: added

src/agent/src/image_rpc.rs
wllenyj: removed

src/agent/src/image_rpc.rs
wllenyj: In my test, this step can actually be omitted. In one of the earlier versions(https://github.com/kata-containers/kata-containers/pull/2734) I submitted, it can directly:
```
skopeo copy docker://$IMAGE oci:$OCI_IMAGE
```
It can omit the manifest step.

src/agent/src/image_rpc.rs
wllenyj: It's easier to use String for the arguments you want afterwards. As far as I know, Path conversion to str may introduce a few more errors to handle.

src/agent/src/image_rpc.rs
wllenyj: fixed

src/agent/src/image_rpc.rs
jodh-intel: Having that option looks a bit alarming to the reader tbh :fearful:

src/agent/src/image_rpc.rs
jodh-intel: I appreciate it's easier, but consider what would happen if `cid` was "`../../../../sbin/init`" or something crazy like that? In fact [that isn't allowed](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L113) but you get the idea ;)

src/agent/src/image_rpc.rs
wllenyj: > This field is just to uniquely represent an image, it will not be processed at runtime, it will be returned to k8s as is, k8s will use this value to call CreateContainer afterwards, then the value will be passed to agent again, this time we can use the value to hash index the Image. PullImage: https://github.com/kubernetes/kubernetes/blob/a2960dc2775848f3b54fb288650e02b4f691743b/pkg/kubelet/images/image_manager.go#L89 CreateContainer https://github.com/kubernetes/kubernetes/blob/a2960dc2775848f3b54fb288650e02b4f691743b/pkg/kubelet/kuberuntime/kuberuntime_container.go#L177

As above,
We need to use `ImageRef` to find the Image bundle directory when you CreateContainer later. Currently, ImageName is used as ImageRef. I think a format like `docker.io/library/busybox:latest` is enough to ensure the uniqueness of Image.
The current HashMap should be considered the simplest ImageManager.

src/agent/src/image_rpc.rs
wllenyj: I was a bit lazy here and wanted to rely on the skopeo return error later.

src/agent/protocols/protos/image.proto
wllenyj: > Also, it would be great if you could add some unit tests (which might require a bit of refactoring, but would allow some of the failure scenarios to be exercised).

Thanks for your review.
These codes will definitely be replaced by ImageManager in the future, so... err....

wllenyj:
> Also, it would be great if you could add some unit tests (which might require a bit of refactoring, but would allow some of the failure scenarios to be exercised).

Thanks for your review.
These codes will definitely be replaced by ImageManager in the future, so... err....

src/agent/src/image_rpc.rs
bpradipt: Yes, skopeo uses the `docker:` prefix to indicate a container registry which could be anything. Few examples here will make it clearer -  https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md

src/agent/src/image_rpc.rs
bpradipt: @jodh-intel you are right. It definitely appears alarming, especially in the context of confidential computing ;) . However, if you are ok, for next demo we can leave it as-is with a comment. 
Ideally if we stick to skopeo for image pull in the long run, then we need to create a proper policy file and use the same.  This was discussed briefly here as well - https://github.com/kata-containers/kata-containers/pull/2733/files#r740133328

Reference for policy: https://github.com/containers/skopeo/blob/main/docs/skopeo.1.md

src/agent/src/image_rpc.rs
wllenyj: Your are right, `verify_cid` is added at the begin of PullImage.

src/agent/protocols/protos/agent.proto
```diff
@@ -67,7 +67,6 @@ service AgentService {
 	rpc CopyFile(CopyFileRequest) returns (google.protobuf.Empty);
 	rpc GetOOMEvent(GetOOMEventRequest) returns (OOMEvent);
 	rpc AddSwap(AddSwapRequest) returns (google.protobuf.Empty);
-	rpc PullImage(PullImageRequest) returns (google.protobuf.Empty);
```
sameo, line 70: I think we will also need to update the virtcontainers agent code to have it use the new Image service as well?

src/agent/src/image_rpc.rs
wllenyj: Thanks.
Yes, k8s reference the image by `ImageRef` that is returned by `PullImage`. So we save `ImageRef` as key to HashMap. The value of HashMap is `rootfs` (bundle path).

src/agent/src/image_rpc.rs
wllenyj: OCI format is needed. 
Eventually the spec and rootfs in the OCI format will be used, but manifest is not needed, it is only an intermediate format and was eventually removed at https://github.com/kata-containers/kata-containers/pull/2956/files#diff-a14bc6af50efd1f06bfe256b3d5d9a0b3be97955c120b9994e7a02b0996d2ff3R88L88.

wllenyj: Two copies are currently used:
```
skopeo --insecure-policy copy docker://docker.io/library/busybox:latest dir:///tmp/<ID>/image_manifest
skopeo --insecure-policy copy dir:///tmp/<ID>/image_manifest oci:///tmp/<ID>image_oci:latest --remove-signatures
```
When I tested it, it seemed that once is fine:
```
skopeo copy  docker://docker.io/library/busybox:latest  oci:///tmp/<ID>image_oci:latest
```

src/agent/protocols/protos/image.proto
```diff
@@ -0,0 +1,31 @@
...
+    rpc PullImage(PullImageRequest) returns (PullImageResponse) {}
+}
+
+message PullImageRequest {
+    // Image name (e.g. docker.io/library/busybox:latest).
+    string image = 1;
+    // Unique image identifier, used to avoid duplication when unpacking the image layers.
+    string container_id = 2;
+    // Use USERNAME[:PASSWORD] for accessing the registry
+    string source_creds = 3;
```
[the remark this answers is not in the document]
wllenyj: removed

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    fn unpack_image(cid: &str) -> Result<()> {
+        let source_path_oci = format!("/tmp/{}/{}", cid, IMAGE_OCI);
+        let target_path_bundle = format!("{}{}{}", CONTAINER_BASE, "/", cid);
+
+        info!(sl!(), "unpack image"; "cid" => cid, "target_bundle_path" => &target_path_bundle);
+
+        // Unpack image
+
```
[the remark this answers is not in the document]
wllenyj: fiexed

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,162 @@
...
+            .arg("--verbose")
+            .arg("unpack")
+            .arg("--image")
+            .arg(&source_path_oci)
+            .arg(&target_path_bundle)
+            .status()?;
+
+        if !status.success() {
+            return Err(anyhow!(format!("failed to unpack image: {:?}", status)));
+        }
```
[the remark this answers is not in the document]
wllenyj: fixed

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,162 @@
...
+    fn unpack_image(cid: &str) -> Result<()> {
+        let source_path_oci = format!("/tmp/{}/{}", cid, IMAGE_OCI);
+        let target_path_bundle = format!("{}{}{}", CONTAINER_BASE, "/", cid);
+
+        info!(sl!(), "unpack image"; "cid" => cid, "target_bundle_path" => &target_path_bundle);
+
+        // Unpack image
+
+        let status: ExitStatus = Command::new(UMOCI_PATH)
+            .arg("--verbose")
```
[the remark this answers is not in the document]
wllenyj: Is this supposed to be for debugging or demo?
removed

src/agent/protocols/protos/agent.proto
wllenyj: You said is this: https://github.com/kata-containers/kata-containers/pull/2956/files#diff-f87c1c212bba3abd0b07bafe69d5b01ad00dfe2c01bed585e2a75fcbb6442b89L681

src/agent/src/image_rpc.rs
wllenyj: fixed

src/agent/src/image_rpc.rs
sameo: Ok, so we should keep it for now, thanks.

src/agent/protocols/protos/agent.proto
sameo: No, I mean that virtcontainers won't be able to reach the new agent Image service API. I think you need to add something like https://github.com/kata-containers/kata-containers/pull/2733/files#diff-6fb7f551035031ed29ad100c61f1a7e92b0fa15d2f67e4c7ea7ebf3b715583f4 to this PR.

src/agent/src/image_rpc.rs
sameo: If things work with one single copy, then yes we need to do that.
@stevenhorsman @davidhay1969 Any idea why we're going through those 2 separate steps?

src/agent/protocols/protos/agent.proto
wllenyj: oh, I got, coming soon.

src/agent/src/image_rpc.rs
sameo: @wllenyj I agree with @jodh-intel here. Could we use `Path` to more safely build those paths and convert them to strings when needed? It's slightly less easy (you'll need an additional `ok_or_else()`) but I think it's worth it.

src/agent/src/image_rpc.rs
stevenhorsman: Dave can probably confirm, but I think it it was 2 steps as the middle form was needed to run `skopeo standalone-verify` against to check the signature, but that shouldn't be required anymore with the plan in https://github.com/kata-containers/kata-containers/issues/2682

davidhay1969: From recollection, `umoci` cannot "unpack" an OCI image that contains a signature, hence we used `skopeo copy` on the second occasion to copy **from** the local manifest-formatted image **to** the OCI-formatted image, and remove the signatures

wllenyj: @davidhay1969 @stevenhorsman 
Thanks for your comments.

src/agent/src/image_rpc.rs
wllenyj: fixed

src/agent/protocols/protos/image.proto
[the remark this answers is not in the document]
stevenhorsman: @sameo - what is the plan for getting source credentials for containers registries back in? Obviously we want to get them via the AA, but I'm not sure that is available yet. We agreed that this image service switch over wouldn't regress currently working function and without being about to pull images from authenticated registries that limits some of our ability to test.

src/agent/protocols/protos/agent.proto
sameo: Thanks.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,166 @@
...
+use async_trait::async_trait;
+use protocols::image;
+use tokio::sync::Mutex;
+use ttrpc::{self, error::get_rpc_status as ttrpc_error};
+
+use crate::rpc::{verify_cid, CONTAINER_BASE};
+use crate::sandbox::Sandbox;
+
+const SKOPEO_PATH: &str = "/usr/bin/skopeo";
+const UMOCI_PATH: &str = "/usr/local/bin/umoci";
```
jodh-intel, line 21: - Can you add a comment explaining why these binaries live in different directories?
- How big are these golang binaries? (I think I asked that on a previous PR or an earlier version but cannot find where ;)
- Are there any plans to rewrite these tools in rust, ideally as crates to avoid having to fork processes?

src/agent/src/image_rpc.rs
jodh-intel: A comment is fine for now imho.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,166 @@
...
+        pull_command
+            .arg("--insecure-policy")
+            .arg("copy")
+            .arg(source_image)
+            .arg(&target_path_manifest);
+
+        let status: ExitStatus = pull_command.status()?;
+        ensure!(
+            status.success(),
+            format!("failed to copy image manifest: {:?}", status)
```
jodh-intel, line 74: You don't need this `format!()`:

```suggestion
            "failed to copy image manifest: {:?}", status)
```

Same comment for the other `ensure!()` calls.

src/agent/src/image_rpc.rs
```diff
@@ -0,0 +1,166 @@
...
+            .arg("--remove-signatures")
+            .status()?;
+
+        ensure!(
+            status.success(),
+            format!("failed to copy image oci: {:?}", status)
+        );
+
+        // To save space delete the manifest.
+        // TODO LATER - when verify image is added, this will need moving the end of that, if required
```
jodh-intel, line 94: Please raise a GitHub issue and add the URL here.

src/agent/src/rpc.rs
```diff
@@ -1412,15 +1394,21 @@ pub fn start(s: Arc<Mutex<Sandbox>>, server_address: &str) -> TtrpcServer {
...
     let health_worker = Arc::new(health_service);
 
+    let image_service = Box::new(image_rpc::ImageService::new(s))
+        as Box<dyn protocols::image_ttrpc::Image + Send + Sync>;
+
     let aservice = protocols::agent_ttrpc::create_agent_service(agent_worker);
 
     let hservice = protocols::health_ttrpc::create_health(health_worker);
 
+    let iservice = protocols::image_ttrpc::create_image(Arc::new(image_service));
```
jodh-intel, line 1404: Nit: I know this is consistent with the code that was already there, but I'd consider making this clearer by creating an `agent_service`, `health_service` and `image_service`.

jodh-intel (commented):
Thanks @wllenyj - a few more comments.

Also, it would be great if you could create a few unit tests if possible.

src/agent/src/image_rpc.rs
stevenhorsman: I think the PR you are referring to is https://github.com/kata-containers/kata-containers/pull/2908#discussion_r736632575. We were using a fedora base which installed the 2 processes differently and when we switched to ubuntu they were kept for compatibility, but once the image refactor changes from Jakob's PR are merged we can consider normalising the paths.

The sizes of the binaries:
```
ls -al /usr/bin/skopeo
-rwxr-xr-x 1 root root 33296024 Nov  8 11:12 /usr/bin/skopeo
``
```
ls -al /usr/local/bin/umoci
-rwxr--r-- 1 root root 7499776 Nov  8 11:12 /usr/local/bin/umoci
```

https://github.com/confidential-containers/image-rs is the repo for the crate that will replace these binaries, but as you can see it isn't ready yet!

Comment (jodh-intel) [reply src/agent/src/image_rpc.rs:21]: Thanks @stevenhorsman. There's not small are they? :)

Comment (stevenhorsman) [reply src/agent/src/image_rpc.rs:21]: Nope - we always planned to replace them with rustified versions, post PoC, and that code has been worked on elsewhere, but isn't ready to go in before we get to the CNCF proposal.

File: src/agent/src/image_rpc.rs (hunk 6 above)
Comment (wllenyj) [reply src/agent/src/image_rpc.rs:67]: added

File: src/agent/src/image_rpc.rs (hunk 21 above)
Comment (wllenyj) [reply src/agent/src/image_rpc.rs]: fixed

File: src/agent/src/rpc.rs (hunk 23 above)
Comment (sameo) [reply src/agent/src/rpc.rs:1404]: > Your means like this?
> 
> ```
> let agent_service = protocols::agent_ttrpc::create_agent_service(agent_worker);
> let health_service = protocols::health_ttrpc::create_health(health_worker);
> let image_service = protocols::image_ttrpc::create_image(Arc::new(image_service));
> ```

Yes, I believe this is what @jodh-intel means.

File: src/agent/src/image_rpc.rs (hunk 20 above)
Comment (sameo) [reply src/agent/src/image_rpc.rs:21]: I confirm using `skopeo` and `umoci` is not the long term plan, they will be replaced by container image management crates.

File: src/agent/src/rpc.rs (hunk 23 above)
Comment (jodh-intel) [reply src/agent/src/rpc.rs:1404]: Yes, exactly that. Not essential, but I think it help the clarity a little.

File: src/agent/src/image_rpc.rs (hunk 21 above)
Comment (jodh-intel) [reply src/agent/src/image_rpc.rs]: Thanks.

**Comment by sameo** [comment]
@wllenyj Could you please add a `Fixes #2984` line at the end of your commits, so that the static checks can pass?

File: src/agent/src/rpc.rs (hunk 23 above)
Comment (wllenyj) [reply src/agent/src/rpc.rs]: fixed

**Review by sameo — APPROVED** [review]

**Comment by sameo** [comment]
@jodh-intel @bpradipt @stevenhorsman Any further comment?

**Comment by stevenhorsman** [comment]
@sameo - I'd still like to hear your response to https://github.com/kata-containers/kata-containers/pull/2956#discussion_r743697631 - about regressing the function of being able to use an authenticated registry

**Review by jiangliu — APPROVED** [review]

**Review by jodh-intel — APPROVED** [review]
Thanks @wllenyj.

Let's just get this landed :smile:

lgtm

**Comment by bpradipt** [comment]
> @jodh-intel @bpradipt @stevenhorsman Any further comment?

No further comment.. Looks good

**Review by bpradipt — APPROVED** [review]
/lgtm

File: src/agent/protocols/protos/image.proto (hunk 16 above)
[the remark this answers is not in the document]
Comment (sameo) [reply src/agent/protocols/protos/image.proto]: @stevenhorsman I believe there are at least a couple of opens related to that:

1. Is the registry authentication part of the CC trust boundary? The current way to have k8s pull an image from authenticated registries is to pass a secret through the pod description. Is that breaking the CC threat model? The host will be able to pull the CC container image because it knows the registry credentials, but the image will be encrypted (or signed). Having the registry credentials won't let the host tamper with or read the container image content, imho. So I wonder if it really makes sense to get those credentials only via attestation.
2.  **If** we want to support a model where one could retrieve registry authentication credentials through attestation, the current KBC and KBS solution for that are still TBD. Discussions are starting in the attestation-agent repo, but we don't have any clear answer for that atm.

Unless there's a strong disagreement about my first point, I think for now we could only focus on getting those credentials through the current k8s path. Given that our `source_creds` field of the `PullImage` payload would be unused in a k8s e2e context (the shim wont fill it, the AA won't neither), I proposed that we remove it for now and add authenticated registries back as its own standalone PR, after merging that one and #2733. This PR would add support for authenticated registries on top of a working k8s e2e flow for CC by:

1. Having containerd forward the `AuthConfig` information that it gets from CRI to the kata shim  
```grpc
// AuthConfig contains authorization information for connecting to a registry.
message AuthConfig {
    string username = 1;
    string password = 2;
    string auth = 3;
    string server_address = 4;
    // IdentityToken is used to authenticate the user and get
    // an access token for the registry.
    string identity_token = 5;
    // RegistryToken is a bearer token to be sent to a registry
    string registry_token = 6;
}
```
2. Defining our own `AuthConfig` subset and add it as one additional field to our agent `PullImage` API
3. Having the shim convert CRI's `AuthConfig` into our own and passing it as part of the agent `PullImage` request. 
4. Using that information to pass the credentials to `skopeo` (which is essentially the part that we have today).

With those 4 points we'd have something that could work e2e with k8s, for authenticated registries.

Let me know if that's an acceptable approach/plan or not. If it is, I'll create an issue to track that work so that we can push it as soon as #2956 and #2733 are in.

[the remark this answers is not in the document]
jiangliu: Agree, let go along the first proposal until there's security threaten found.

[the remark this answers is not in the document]
bpradipt: This makes sense.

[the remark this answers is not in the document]
stevenhorsman: So I'm definitely not an expert, but my initial thoughts are:
- I don't know if passing keys through as secrets breaks the CC threat model, but I'm not sure it is great practice for what we are after. Unless we have an encryption scheme for those secrets then can't a CSP admin just spin up a different pod that will get list out the secrets and thereby get the credentials? If we ensure encryption, or verification for all images that it shouldn't effect the workload, but I wonder if there are things like DoS attacks (certain registries like docker hub have limits for number of images a user can pull), or other security implications of exposing the credentials?
- My expectation was that we wanted to get authentication credentials from the attestation agent's KBS, but as you say that isn't ready yet.
- I agree that the current solution is not the end one, but it wasn't intended to be. It was a temporary solution so we could test pulling a variety of images from authenticated registries in the kata agent, hence why I didn't extend that into the kata shim when adding ctr pull-image support. In the absence of generic key support from the attestation agent (which I think might cause other problems) I don't have an issue with your proposed implementation of building the model on the containerd AuthConfig either, but what I still don't understand is why we want to remove something that current works and we are using for testing without having a replacement. What was the gain from doing extra work to delete this code that I missed?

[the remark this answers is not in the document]
sameo: >     * I don't know if passing keys through as secrets breaks the CC threat model, but I'm not sure it is great practice for what we are after. Unless we have an encryption scheme for those secrets then can't a CSP admin just spin up a different pod that will get list out the secrets and thereby get the credentials? If we ensure encryption, or verification for all images that it shouldn't effect the workload, but I wonder if there are things like DoS attacks (certain registries like docker hub have limits for number of images a user can pull), or other security implications of exposing the credentials?
> 

I agree, but I think those are all problems that exist today outside of the CC context. And what I'm saying is that the CC threat model is not covering this. Using authenticated registries is definitely a good practice, but I don't think it protects workload data in-use.

And to be clear, even with an in-guest image management model, we do want to eventually support authenticated registries.


>     * I agree that the current solution is not the end one, but it wasn't intended to be. It was a temporary solution so we could test pulling a variety of images from authenticated registries in the kata agent, hence why I didn't extend that into the kata shim when adding ctr pull-image support. 

Yes, that was clear to me as well.

> In the absence of generic key support from the attestation agent (which I think might cause other problems) I don't have an issue with your proposed implementation of building the model on the containerd AuthConfig either

Thanks.

> but what I still don't understand is why we want to remove something that current works and we are using for testing without having a replacement. What was the gain from doing extra work to delete this code that I missed?

You did not miss anything. It's just that with the next set of milestones this code would not be exercised except through `agent-ctl`. And it will be modified into something slightly more structured when we actually will support it end-to-end. I felt it was clearer to not continue supporting this when we're saying that our next goals are really focused on the k8s e2e pipeline.

If that's really problematic and slows our progress down, I can quickly add it back as a separate commit on top of that PR.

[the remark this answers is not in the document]
stevenhorsman: Ok, thanks for the explanation. I don't want to hold up this PR, so I think the pragmatic path is to get this in as-is, but if the containerd k8s e2e solution isn't added soon then I can re-create the PR to temporary add agent-ctl source credentials support back in for temporary testing purposes.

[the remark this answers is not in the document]
sameo: Hey @stevenhorsman I actually removed the commit that deleted `source_creds` and kept it as is for now.

Your request is fair, sorry for the back and forth. Do you think you could help with actually improving the current state and using a more structured agent API payload for registry authentication credentials, something that would be a subset of `AuthConfig` for now like e.g.

```
message AuthConfig {
    string username = 1;
    string password = 2;
}
```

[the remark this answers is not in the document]
stevenhorsman: Sure, I think we should create an issue and get a bit of design done. I'm not sure it is as simple as username and password as some (most of the container registries we use recommend using API keys and some of these have a pre-prepended userid and other don't. That's why we originally kept it as a single combined string to push that complication on the user, but understanding what the kubernetes secret flows down into the containerd `AuthConfig` would be really useful. I think as get closer to moving to the Alibaba containerd fork we might learn a bit more.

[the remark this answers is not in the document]
sameo: Right. With a simple username+password `AuthConfig` we know we're not overengineering it at least, and we'd get the same level of functionality as we have today. And we can expand it with what we need as we learn more, without breaking much.

Issue #2957, opened 2021-11-03
The agent fails if an unknown seccomp system call is set to a container


ManaSugi:
# Description of problem

Currently, Kata agent cannot continue to run if an unknown seccomp system call that is not supported by the kernel is set to a container.

For example, we set `RuntimeDefault` seccomp profiles using `crictl`, the agent may fail in case the profile includes unknown system calls in the kernel.

```bash
$ sudo crictl run -r kata container.json sandbox.json
```

<details><summary>container.json</summary>

```json
{
    "metadata": {
        "name": "ubuntu"
    },
    "image": {
        "image": "docker.io/ubuntu"
    },
    "command": [
        "/bin/sh"
    ],
    "working_dir": "/",
    "envs": [
        {
            "key": "PATH",
            "value": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
        },
        {
            "key": "TERM",
            "value": "xterm"
        }
    ],
    "stdin": true,
    "stdin_once": true,
    "tty": true,
    "linux": {
        "security_context": {
            "seccomp_profile_path": "runtime/default"
        }

    }
}
```

</details>

<details><summary>sandbox.json</summary>

```json
{
    "metadata": {
        "name": "ubuntu-pod",
        "uid": "ubuntu-pod",
        "namespace": "test.kata"
    },
    "hostname": "ubuntu_host",
    "dns_config": {
    },
    "port_mappings": [],
    "resources": {
    },
    "linux": {
    }
}
```

</details>

Related issue: https://github.com/kata-containers/tests/pull/4132

# Expected result

The agent should launch a container with restricted system calls by seccomp.

# Actual result

```
DEBU[0000] get runtime connection
DEBU[0000] connect using endpoint 'unix:///var/run/containerd/containerd.sock' with '10s' timeout
DEBU[0000] connected successfully using endpoint: unix:///var/run/containerd/containerd.sock
DEBU[0000] get image connection
DEBU[0000] connect using endpoint 'unix:///var/run/containerd/containerd.sock' with '10s' timeout
DEBU[0000] connected successfully using endpoint: unix:///var/run/containerd/containerd.sock
DEBU[0000] RunPodSandboxRequest: &RunPodSandboxRequest{Config:&PodSandboxConfig{Metadata:&PodSandboxMetadata{Name:ubuntu-pod,Uid:ubu
[...]

PR #2958, opened 2021-11-03
agent: Ignore unknown seccomp system calls

ManaSugi:
If Kata agent cannot resolve the system calls given by seccomp profiles,
the agent ignores the system calls and continues to run without an error.

Fixes: #2957

### DIFF
--- a/src/agent/rustjail/src/seccomp.rs
+++ b/src/agent/rustjail/src/seccomp.rs
@@ -68,7 +68,14 @@ pub fn init_seccomp(scmp: &LinuxSeccomp) -> Result<()> {
         }
 
         for name in &syscall.names {
-            let syscall_num = get_syscall_from_name(name, None)?;
+            let syscall_num = match get_syscall_from_name(name, None) {
+                Ok(num) => num,
+                Err(_) => {
+                    // If we cannot resolve the given system call, we assume it is not supported
+                    // by the kernel. Hence, we skip it without generating an error.
+                    continue;
+                }
+            };
 
             if syscall.args.is_empty() {
                 filter.add_rule(action, syscall_num, None)?;
--- a/src/agent/rustjail/src/seccomp.rs
+++ b/src/agent/rustjail/src/seccomp.rs
@@ -39,6 +39,24 @@ fn get_rule_conditions(args: &[LinuxSeccompArg]) -> Result<Vec<ScmpArgCompare>> 
     Ok(conditions)
 }
 
+pub fn get_unknown_syscalls(scmp: &LinuxSeccomp) -> Option<Vec<String>> {
+    let mut unknown_syscalls: Vec<String> = Vec::new();
+
+    for syscall in &scmp.syscalls {
+        for name in &syscall.names {
+            if get_syscall_from_name(name, None).is_err() {
+                unknown_syscalls.push(name.to_string());
+            }
+        }
+    }
+
+    if unknown_syscalls.is_empty() {
+        None
+    } else {
+        Some(unknown_syscalls)
+    }
+}
+
 // init_seccomp creates a seccomp filter and loads it for the current process
 // including all the child processes.
 pub fn init_seccomp(scmp: &LinuxSeccomp) -> Result<()> {
@@ -116,6 +134,72 @@ mod tests {
         };
     }
 
+    const TEST_DATA: &str = r#"{
+          "defaultAction": "SCMP_ACT_ALLOW",
+          "architectures": [
+          ],
+          "flags": [
+              "SECCOMP_FILTER_FLAG_LOG"
+          ],
+          "syscalls": [
+              {
+                 "names": [
+                      "dup3",
+                      "invalid_syscall1",
+                      "invalid_syscall2"
+                  ],
+                  "action": "SCMP_ACT_ERRNO"
+              },
+              {
+                 "names": [
+                      "process_vm_readv"
+                  ],
+                  "action": "SCMP_ACT_ERRNO",
+                  "errnoRet": 111,
+                  "args": [
+                      {
+                          "index": 0,
+                          "value": 10,
+                          "op": "SCMP_CMP_EQ"
+                      }
+                  ]
+              },
+              {
+                 "names": [
+                      "process_vm_readv"
+                  ],
+                  "action": "SCMP_ACT_ERRNO",
+                  "errnoRet": 111,
+                  "args": [
+                      {
+                          "index": 0,
+                          "value": 20,
+                          "op": "SCMP_CMP_EQ"
+                      }
+                  ]
+              },
+              {
+                 "names": [
+                      "process_vm_readv"
+                  ],
+                  "action": "SCMP_ACT_ERRNO",
+                  "errnoRet": 222,
+                  "args": [
+                      {
+                          "index": 0,
+                          "value": 30,
+                          "op": "SCMP_CMP_EQ"
+                      },
+                      {
+                          "index": 2,
+                          "value": 40,
+                          "op": "SCMP_CMP_EQ"
+                      }
+                  ]
+              }
+          ]
+      }"#;
+
     #[test]
     fn test_get_filter_attr_from_flag() {
         skip_if_not_root!();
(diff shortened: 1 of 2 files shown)


ManaSugi:
We are supposed to record a log about the unknown system call using `log_child`.
However, `cfd_log` fd is already closed before calling the second `init_seccomp`, so we cannot do it.

src/agent/rustjail/src/seccomp.rs
```diff
@@ -68,7 +68,14 @@ pub fn init_seccomp(scmp: &LinuxSeccomp) -> Result<()> {
         }
 
         for name in &syscall.names {
-            let syscall_num = get_syscall_from_name(name, None)?;
+            let syscall_num = match get_syscall_from_name(name, None) {
+                Ok(num) => num,
+                Err(_) => {
+                    // If we cannot resolve the given system call, we assume it is not supported
+                    // by the kernel. Hence, we skip it without generating an error.
+                    continue;
```
liubin, line 76: In this case, `libseccomp` returns `Could not resolve syscall name` error without the name couldn't be resolved, so if we can log it will make users easy for debugging.

ManaSugi: Thank you for the comment. You're right. I need to fix `libseccomp` crate.

liubin (approved):

ManaSugi:
https://github.com/kata-containers/tests/pull/4132, all tests were passed successfully.

snir911:
Hi @ManaSugi , LGTM, thanks! are you planning to add log as discussed above?

ManaSugi:
@snir911 I cannot use `cfd_log` in the second `init_seccomp`, so I think there is no way to log it. However, if the agent can delay the closing `cfd_log`, it is possible to log it. Anyway, I don't know if I can record the kata log without modifying the current code.
I modified `libseccomp` crate to be able to generate an error with unresolved syscall name, however, the agent can not log the error message as it is now.

src/agent/rustjail/src/seccomp.rs
jodh-intel: Might be good to log the error though?

jodh-intel (approved):
Thanks @ManaSugi.

src/agent/rustjail/src/seccomp.rs
ManaSugi: Thank you for the comment. Yes, I want to log the error, but the agent cannot do it without `cfd_log` file descriptor.

GabyCT (approved):

src/agent/rustjail/src/seccomp.rs
jodh-intel: In which case, I wonder if it makes sense to split  that syscall checking code out into a new `check_seccomp() -> Result<()>` function and call that before the fork. That would allow the parent to log the errors. However, that might not be a good idea if the loop takes a long time to run. If it is slow, we could always create a channel that the parent can read asynchronously. If the child finds any erorrs, it could send the details back to the parent through the channel, to allow the parent to log them.

ManaSugi: Thank you for the suggestion. it makes sence, but I need to make `do_init_child` and `init_child` async function.
I'll try to create it.

ManaSugi:
@jodh-intel 
I added the asynchronous logging for unknown seccomp system calls.
But, I'm not sure whether making `check_unknown_syscall` asynchronous is valuable or not as of now because the tokio runtime itself requires additional overhead. If many system calls is given by seccomp profiles, `I guess get_unknown_syscall` may be CPU bound workloads but I may need to measure the performance.
In addition, `do_init_child` become an asynchronous function by this commit, so the agent could be better to use `async` functions such as `read_async` instead of `read_sync` to remove blocking functions.

ManaSugi:
I found mistakes.

ManaSugi:
I'm sorry for the late.
## sync vs async for logging unknown seccomp syscalls
### Cases

I created three cases of ways to log unknown seccomp system calls and evaluated each case.

1. **sync (this branch: ignore-unknown-systemcall)**

2. **async with `block_in_place` (branch: [ignore-unknown-systemcall-blockinplace](https://github.com/ManaSugi/kata-containers/commit/e9d57ff65484725a55ed6f02dcede0ee18a1d6fd))**

   * `get_unknown_syscalls` may become a blocking function if many seccomp system calls are given, so I use `block_in_place` .

   * `check_unknown_syscalls` runs as another thread.

3. **async without `block_in_place` (branch: [ignore-unknown-systemcall-async](https://github.com/ManaSugi/kata-containers/commit/d81c6a1fb0c6ae7774f71468d1369333915a3c27))**

   * `check_unknown_syscalls` runs as another thread.

     

### Evaluation

**Environment:**

| | |
|---|---|
| CPU    | 3.8GHz 12-Core |
| Memory | 32GB           |
| crictl | v1.21.0        |

**`crictl` configuration files:**

Use `runtime/default` profiles as seccomp profiles because they already have many system calls (about 340 system calls)
Ref: https://github.com/containerd/containerd/blob/main/contrib/seccomp/seccomp_default.go

<details><summary>container.json</summary>

```json
{
    "metadata": {
        "name": "ubuntu"
    },
    "image": {
        "image": "docker.io/ubuntu"
    },
    "command": [
        "/bin/sh"
    ],
    "working_dir": "/",
    "envs": [
        {
            "key": "PATH",
            "value": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
        },
        {
            "key": "TERM",
            "value": "xterm"
        }
    ],
    "stdin": true,
    "stdin_once": true,
    "tty": true,
    "log_path": "/var/run/kata.log",
    "linux": {
        "security_context": {
            "seccomp_profile_path": "runtime/default"
        }

    }
}
```
</details>

<details><summary>sandbox.json</summary>

```json
{
    "metadata": {
        "name": "ubuntu-pod",
        "uid": "ubuntu-pod",
        "namespace": "test.kata"
    },
    "hostname": "ubuntu_host",
    "dns_config": {
    },
    "port_mappings": [],
    "resources": {
    },
    "linux": {
    }
}
```
</details>

**Measuring method:**

The elapsed time is the execution time from the begging of `do_init_child` to the point of closing `cfd_log`.
The reason for the measurement by the point of closing `cfd_log` is because the elapsed time is recorded through the logging.

```rust
// rustjail/src/container.rs
fn do_init_child(cwfd: RawFd) -> Result<()> {
    // Start
    let start_time = std::time::Instant:now();
    
    <...>
    // notify parent that the child's ready to start           
	write_sync(cwfd, SYNC_SUCCESS, "")?;                       
	log_child!(cfd_log, "ready to run exec"); 
    
    // End
	let end_time = start_time.elapsed();                                 
	log_child!(                                                
    	cfd_log,                                               
   	 	"do_init_child elapsed time: {}",                                
    	end_time.as_secs() as f64 + end_time.subsec_nanos() as f64 * 1e-9
	); 
    
    let _ = unistd::close(cfd_log);   
    <...>
}
```

**Command:**

```bash
$ sudo crictl run -r kata container.json sandbox.json
```

### Result

This table shows the average of the elapsed time for running 100 Kata containers using `crictl` in each case.
As a result, all three ways have almost the same elapsed time considering the standard deviation.
I think that we do not need to use `tokio async` for traversing system calls, but rather `sync` is simpler and cleans the code.

|                           | sync | async with block_in_place | async without block_in_place |
| ------------------------- | --------- | ------------------------- | ---------------------------- |
| elapsed time [msec]       | 18.8085   | 18.7854                   | 19.1449                      |
| standard deviation [msec] | 1.89755   | 1.70503                   | 2.01963                      |


The log for unknown seccomp system calls is recorded correctly as follows.

```
Nov 23 23:09:34 manasugi kata[2286067]: time="2021-11-23T23:09:34.934939341+09:00" level=debug msg="reading guest console" console-protocol=unix console-url=/run/vc/vm/174c114d5b30f44a6709de7999de64549afd3b30f280f1391a84ff8f5fad3d9c/console.sock name=containerd-shim-v2 pid=2286067 sandbox=174c114d5b30f44a6709de7999de64549afd3b30f280f1391a84ff8f5fad3d9c source=virtcontainers subsystem=sandbox vmconsole="{\"msg\":\"unknown seccomp system calls: close_range epoll_pwait2\",\"level\":\"INFO\",\"ts\":\"2021-11-23T14:09:34.931020676+00:00\",\"version\":\"0.1.0\",\"name\":\"kata-agent\",\"pid\":\"55\",\"source\":\"agent\",\"subsystem\":\"container\",\"eid\":\"6ca733ca1f0ed867c78d8a4fbbf825fa8e3a829f1a666271846be45a7a806a92\",\"action\":\"child process log\",\"cid\":\"6ca733ca1f0ed867c78d8a4fbbf825fa8e3a829f1a666271846be45a7a806a92\",\"module\":\"rustjail\"}"
```

### Conclusion

I think It is better to log unknown seccomp system calls without using `tokio async` and another thread.
@jodh-intel  Could you tell me your thoughts?

ManaSugi:
@liubin @jodh-intel @GabyCT I would appreciate it if you could re-review.

src/agent/rustjail/src/seccomp.rs
```diff
@@ -39,6 +39,20 @@ fn get_rule_conditions(args: &[LinuxSeccompArg]) -> Result<Vec<ScmpArgCompare>>
     Ok(conditions)
 }
 
+pub fn check_unknown_syscalls(scmp: &LinuxSeccomp) -> String {
+    let mut unknown_syscalls = String::new();
```
jodh-intel, line 43: This might be better as a `Vec<String>` (or `Vec<&str>`).

src/agent/rustjail/src/seccomp.rs
```diff
@@ -39,6 +39,20 @@ fn get_rule_conditions(args: &[LinuxSeccompArg]) -> Result<Vec<ScmpArgCompare>>
...
+
+    for syscall in &scmp.syscalls {
+        for name in &syscall.names {
+            if get_syscall_from_name(name, None).is_err() {
+                unknown_syscalls += format!("{} ", name).as_str();
```
jodh-intel, line 48: If you want to keep the list as a string you can simplify to something like:

```suggestion
                unknown_syscalls += name.to_string();
```

src/agent/rustjail/src/seccomp.rs
```diff
@@ -39,6 +39,20 @@ fn get_rule_conditions(args: &[LinuxSeccompArg]) -> Result<Vec<ScmpArgCompare>>
     Ok(conditions)
 }
 
+pub fn check_unknown_syscalls(scmp: &LinuxSeccomp) -> String {
```
jodh-intel, line 42: It would be more idiomatic to make this `-> Result<Vec<String>>` or similar.

jodh-intel (commented):
Thanks @ManaSugi

src/agent/rustjail/src/seccomp.rs
ManaSugi: I fixed it. Thank you.

src/agent/rustjail/src/seccomp.rs
ManaSugi: I fixed it. Thank you.

src/agent/rustjail/src/seccomp.rs
ManaSugi: I use `Option<Vec<String>` instead of `Result<Vec<String>>`.

PR #2959, opened 2021-11-03
cgroups: Fix systemd cgroup support

snir911:
This fixes the cgroups v1 support when systemd cgroups driver is used
As github.com/containerd/cgroups doesn't support scope units which are essential in some cases, lets create the cgroups 
manually and load it trough the cgroups api
It continues @gkurz [work](https://github.com/kata-containers/kata-containers/pull/2933)

A follow up TODOs:
* https://github.com/kata-containers/kata-containers/issues/2964
* https://github.com/kata-containers/kata-containers/issues/2963

Fixes: #2868


snir911:
@sameo do you mind to have a look at this (continuing #2933 and address your suggestions from there)
@c3d , @fgiudici also :)

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -116,13 +118,33 @@ func sandboxDevices() []specs.LinuxDeviceCgroup {
...
 	}
 
-	cgroup, err := cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupPath), resources)
+	if IsSystemdCgroup(path) {
+		slice, unit, err := getSliceAndUnit(path)
+		if err != nil {
+			return nil, err
+		}
+		// github.com/containerd/cgroups doesn't support creating a scope unit with
+		// v1 cgroups against systemd, the following intreacts directly with systemd
```
jodh-intel, line 134: Nit: s/intreacts/interacts/

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -116,13 +118,33 @@ func sandboxDevices() []specs.LinuxDeviceCgroup {
...
+		// to create the cgroup and then load it using containerd's api
+		err = createCgroupsSystemd(slice, unit, uint32(os.Getpid())) // adding runtime process, it makes calling setupCgroups redundant
+		// TODO: make pid as input variable?
```
jodh-intel, line 137: Is this comment redundant due to the `Getpid()` call above?

src/runtime/virtcontainers/pkg/cgroups/utils.go
```diff
@@ -75,6 +79,67 @@ func IsSystemdCgroup(cgroupPath string) bool {
...
+func cgroupHierarchy(path string) (cgroups.Hierarchy, cgroups.Path) {
+	if !IsSystemdCgroup(path) {
+		return cgroups.V1, cgroups.StaticPath(path)
+	} else {
+		slice, unit, _ := getSliceAndUnit(path)
+		cgroupSlicePath, _ := systemd.ExpandSlice(slice)
+		return cgroups.Systemd, cgroups.Slice(cgroupSlicePath, unit)
+	}
+}
+func createCgroupsSystemd(slice string, unit string, pid uint32) error {
```
jodh-intel, line 98: Nit: Missing blank line between functions.

src/runtime/virtcontainers/pkg/cgroups/utils.go
```diff
@@ -75,6 +79,67 @@ func IsSystemdCgroup(cgroupPath string) bool {
...
+		Value: dbus.MakeVariant(units),
+	}
+}
+
+func cgroupHierarchy(path string) (cgroups.Hierarchy, cgroups.Path) {
+	if !IsSystemdCgroup(path) {
+		return cgroups.V1, cgroups.StaticPath(path)
+	} else {
+		slice, unit, _ := getSliceAndUnit(path)
+		cgroupSlicePath, _ := systemd.ExpandSlice(slice)
```
jodh-intel, line 94: These calls suggest `cgroupHierarchy()` should be returning an `error`.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -669,6 +671,7 @@ func (s *Sandbox) createCgroups() error {
 		// We're creating an overhead cgroup, with no constraints. Everything but
 		// the vCPU threads will eventually make it there.
 		overheadCgroup, err := cgroups.NewCgroup(fmt.Sprintf("/%s/%s", cgroupKataOverheadPath, s.id), &specs.LinuxResources{})
+		// TODO: support systemd cgroups overhead cgroup
```
jodh-intel, line 674: Could you raise a GitHub issue for all TODO's and FIXME's and reference their URLs in the comments?

jodh-intel (commented):
Thanks @snir911. A few comments.

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
snir911: Well, I meant through the function call, will remove though

src/runtime/virtcontainers/pkg/cgroups/utils.go
snir911: Right but i wouldn't be able to use `groupHierarchy()` as input to Load() :P, i think it should always pass as the string format is checked in the `IsSystemdCgroup()`

jodh-intel: You can do something like this I think:

```diff
diff --git a/src/runtime/virtcontainers/pkg/cgroups/cgroups.go b/src/runtime/virtcontainers/pkg/cgroups/cgroups.go
index a0dfdab5..c9a34183 100644
--- a/src/runtime/virtcontainers/pkg/cgroups/cgroups.go
+++ b/src/runtime/virtcontainers/pkg/cgroups/cgroups.go
@@ -137,7 +137,13 @@ func NewCgroup(path string, resources *specs.LinuxResources) (*Cgroup, error) {
                if err != nil {
                        return nil, err
                }
-               if cgroup, err = cgroups.Load(cgroupHierarchy(path)); err == nil {
+
+               cgHierarchy, cgPath, err := cgroupHierarchy(path)
+               if err != nil {
+                       return nil, err
+               }
+
+               if cgroup, err = cgroups.Load(cgHierarchy, cgPath); err == nil {
                        err = cgroup.Update(resources)
                }
        } else {
```

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -117,13 +118,32 @@ func sandboxDevices() []specs.LinuxDeviceCgroup {
...
-	cgroup, err := cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupPath), resources)
+	if IsSystemdCgroup(path) {
+		slice, unit, err := getSliceAndUnit(path)
+		if err != nil {
+			return nil, err
+		}
+		// github.com/containerd/cgroups doesn't support creating a scope unit with
+		// v1 cgroups against systemd, the following interacts directly with systemd
+		// to create the cgroup and then load it using containerd's api
+		err = createCgroupsSystemd(slice, unit, uint32(os.Getpid())) // adding runtime process, it makes calling setupCgroups redundant
```
fgiudici, line 136: To my understanding we create the scope as per the container engine request and we will not create other cgroups there.
If we add the pid of the shim here, we will not honor sandbox_cgroup_only=false, right?
Maybe we could put a comment here and deal with this later.

src/runtime/virtcontainers/pkg/cgroups/utils.go
```diff
@@ -75,6 +79,68 @@ func IsSystemdCgroup(cgroupPath string) bool {
...
+	}
+
+	// https://github.com/opencontainers/runc/blob/master/docs/systemd.md
+	if strings.HasSuffix(unit, ".scope") {
+		// Otherwise it's a scope, which we put into a Slice=.
+		properties = append(properties, newProperty("Delegate", true))
+		properties = append(properties, systemdDbus.PropSlice(slice))
+		properties = append(properties, systemdDbus.PropPids(pid))
+	} else {
+		// If we create a slice, the parent is defined via a Wants=.
```
fgiudici, line 121: Since we will detect systemd cgroups driver by the format of the cgroups path passed down by the container engine ([slice]:[prefix]:[name]) where a scope is expected: is there any case where we will need to create a slice and not a scope?

fgiudici (approved):
Thanks @snir911 , lgtm.
While this PR may not be perfect (e.g., the sandbox_cgroup_only=false support) it should fix the systemd cgroups driver support which is currently broken.

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
snir911: Yes, I left a TODO to address the  sandbox_cgroup_only=false case later, however, IIUC according to: https://github.com/kata-containers/kata-containers/blob/a7a47bd7d48b49c9cb207b93473f99fd7965c4e8/src/runtime/virtcontainers/sandbox.go#L2185-L2189
we add this process either way

src/runtime/virtcontainers/pkg/cgroups/utils.go
snir911: You're right, I left as i though it may be used somehow later, I'll remove it for now

src/runtime/virtcontainers/pkg/cgroups/utils.go
snir911: Sure i can, i was tempted by the one-liner :) done

jodh-intel (approved):
Thanks @snir911.

lgtm

jodh-intel:
Looks like the static checker is unhappy with the vendoring.

fidencio:
> Looks like the static checker is unhappy with the vendoring.

@snir, `make -C src/runtime handle_vendor` should do the trick to get the vendor check happy.
While here, I'd like to ask to change `white list the ...` for simply `allow the ...`.  That has the same meaning, is shorter, and goes together with the concious language effort that's happening all over the place.

sameo (approved):

sameo:
Thanks for the fix @snir911

src/runtime/go.sum
```diff
@@ -97,6 +97,7 @@ github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5P
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/cilium/ebpf v0.2.0/go.mod h1:To2CFviqOWL/M0gIMsvSMlqe7em/l1ALkX1PyjrX2Qs=
 github.com/cilium/ebpf v0.4.0/go.mod h1:4tRaxcgiL706VnOzHOdBlY8IEAIdxINsQBcU4xJJXRs=
+github.com/cilium/ebpf v0.6.2 h1:iHsfF/t4aW4heW2YKfeHrVPGdtYTL4C4KocpM8KTSnI=
```
c3d, line 100: Unclear to me what in your changes added this, but I'm not an expert at Go modules.

There seem to be many additional files for `ebff` that do not seem justified based on usage, so I wonder if a `go mod tidy` would not be in order.

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -53,9 +53,10 @@ func sandboxDevices() []specs.LinuxDeviceCgroup {
 	// In order to run Virtual Machines and create virtqueues, hypervisors
 	// need access to certain character devices in the host, like kvm and vhost-net.
 	hypervisorDevices := []string{
-		"/dev/kvm",       // To run virtual machines
-		"/dev/vhost-net", // To create virtqueues
-		"/dev/vfio/vfio", // To access VFIO devices
+		"/dev/kvm",         // To run virtual machines
+		"/dev/vhost-net",   // To create virtqueues
+		"/dev/vfio/vfio",   // To access VFIO devices
+		"/dev/vhost-vsock", // To interact with vsock if
```
c3d, line 60: Ah, that was it. So I was right ;-)

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -116,14 +117,42 @@ func sandboxDevices() []specs.LinuxDeviceCgroup {
...
+		}
+		// github.com/containerd/cgroups doesn't support creating a scope unit with
+		// v1 cgroups against systemd, the following interacts directly with systemd
+		// to create the cgroup and then load it using containerd's api
+		err = createCgroupsSystemd(slice, unit, uint32(os.Getpid())) // adding runtime process, it makes calling setupCgroups redundant
+		if err != nil {
+			return nil, err
+		}
+
+		cgHierarchy, cgPath, err := cgroupHierarchy(path)
```
c3d, line 140: Much tidier than the first version I saw. Looks good.

src/runtime/virtcontainers/pkg/cgroups/utils.go
```diff
@@ -75,6 +80,76 @@ func IsSystemdCgroup(cgroupPath string) bool {
...
+		return err
+	}
+	defer conn.Close()
+
+	properties := []systemdDbus.Property{
+		systemdDbus.PropDescription("cgroup " + unit),
+		newProperty("DefaultDependencies", false),
+		newProperty("MemoryAccounting", true),
+		newProperty("CPUAccounting", true),
+		newProperty("BlockIOAccounting", true),
```
c3d, line 121: You may want to use `IOAccounting` now (see https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html). I think `BlockIOAccounting` is deprecated.

Do we want to enable `IPAccounting` as well?

c3d (approved):
Approved. However, I would recommend:

* Replacing `BlockIOAccounting` (deprecated) with `IOAccounting`, and possibly enable `IPAccounting` which I suspect will be necessary for network metrics.
* Checking if a `go mod tidy` would clean up the various added files in the `ebpf` module

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
snir911: Actually I was wondering why it didn't make issues until now, did it?

src/runtime/go.sum
snir911: Me neither I think it's because of "github.com/opencontainers/runc/libcontainer/cgroups/systemd"

snir911:
> cgroups.go:151:3: ineffectual assignment to err (ineffassign)
		err = cgroup.Update(resources)

I couldn't figure why linter was complaining, changed a bit the conditions structure 

/test

src/runtime/virtcontainers/pkg/cgroups/utils.go
snir911: I'm not sure, in my environment network metrics are currently being collected (from another source i guess), if someone uses it i can enable it

c3d:
> > cgroups.go:151:3: ineffectual assignment to err (ineffassign)
> > err = cgroup.Update(resources)
> 
> I couldn't figure why linter was complaining, changed a bit the conditions structure

I don't see any assignment on line 151 in the current version. What was the assignment for?

snir911:
@c3d it's in line 147 now, because `Load` does not apply the resources as `Create` does i do it using `Update` after loading the cgroups

jodh-intel:
@snir911 - can we remove the dnm label to allow this to land yet?

snir911:
Not yet @jodh-intel, i added this label as i noticed the cenos8 is failing, this is happening due to the fact the  sandbox_cgroup_only=false and i missed to leave the behavior in this case as it was (and as i stated), it should be a simple fix I'll do it by the end of the week and remove the tag.
Also I opened an issue suggesting to change the default sandbox_cgroup_only value:  https://github.com/kata-containers/kata-containers/issues/2996

snir911:
Changed mostly NewSandboxCgroup and NewCgroup so that it will act as before when sandbox_cgroup_only=false, I'd appreciate if you can re-visit

fidencio:
> Changed mostly **NewSandboxCgroup()** and **NewCgroup()** so that it will act as before when `sandbox_cgroup_only=false`, I'd appreciate if you could re-visit

Based on this, would it make sense to add a `do-not-merge` label till @sameo reviews it?

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -136,15 +137,58 @@ func NewCgroup(path string, resources *specs.LinuxResources) (*Cgroup, error) {
...
 
-func NewSandboxCgroup(path string, resources *specs.LinuxResources) (*Cgroup, error) {
+func NewSandboxCgroup(path string, resources *specs.LinuxResources, sandboxCgroupOnly bool) (*Cgroup, error) {
+	var cgroup cgroups.Cgroup
 	sandboxResources := *resources
 	sandboxResources.Devices = append(sandboxResources.Devices, sandboxDevices()...)
 
-	return NewCgroup(path, &sandboxResources)
+	// Currently we know to handle systemd cgroup path only when it's the only cgroup (no overhead group), hence,
+	// if sandboxCgroupOnly is not true we treat it as cgroupfs path as it used to be, although it may incorrect
```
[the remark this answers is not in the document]
snir911: Thanks, it just work as current code , it should be addressed in #2963

sameo (approved):

Issue #2966, opened 2021-11-04
Enable debug logs for the agent


jodh-intel:
We have a local logging crate that is used by the rust agent, the `trace-forwarder` and the `agent-ctl` tool:

- https://github.com/kata-containers/kata-containers/tree/main/pkg/logging

That crate is built on the [`slog`](https://docs.rs/slog/) logging framework which strips out extraneous logging calls at _compile time_. The strip behaviour is controlled by the `slog` `*_max_level*` features:

- https://github.com/kata-containers/kata-containers/blob/main/pkg/logging/Cargo.toml#L15

Enabling the max level feature allows the log level to be modified at runtime (which all three of our tool support). However, there is a bug: we've set the maximum release log level to "info" meaning "strip out all logging calls 'below' this log level", which includes all debug calls.

We need to set `release_max_level_debug` to allow the tools to emit debug messages when their debug options are enabled.

PR #2967, opened 2021-11-04
logging: Enable agent debug output for release builds

jodh-intel:
Raise the `slog` maximum log level feature for release code from `info`
to `debug` by changing the `slog` maximum level features in the shared
`logging` crate. This allows the consumers of the `logging` crate (the
agent, the `trace-forwarder` and the `agent-ctl` tool) to produce debug
output when their debug options are enabled. Currently, those options
will essentially be a NOP (unless using a debug version of the code).

Testing showed that setting the `slog` maximum level features in the
rust manifest files for the consumers of the `logging` crate has no
impact: those values are ignored, so they have been removed and replaced
with a comment stating the levels are set in the `logging` crate.

Fixes: #2966.

### DIFF
--- a/src/agent/vsock-exporter/Cargo.toml
+++ b/src/agent/vsock-exporter/Cargo.toml
@@ -15,6 +15,6 @@ serde = { version = "1.0.126", features = ["derive"] }
 tokio-vsock = "0.3.1"
 bincode = "1.3.3"
 byteorder = "1.4.3"
-slog = { version = "2.5.2", features = ["dynamic-keys", "max_level_trace", "release_max_level_info"] }
+slog = { version = "2.5.2", features = ["dynamic-keys", "max_level_trace", "release_max_level_debug"] }
 async-trait = "0.1.50"
 tokio = "1.2.0"
--- a/pkg/logging/Cargo.toml
+++ b/pkg/logging/Cargo.toml
@@ -12,7 +12,7 @@ serde_json = "1.0.39"
 # - Dynamic keys required to allow HashMap keys to be slog::Serialized.
 # - The 'max_*' features allow changing the log level at runtime
 #   (by stopping the compiler from removing log calls).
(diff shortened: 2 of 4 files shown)


liubin (approved):

jodh-intel:
I've updated the branch. After testing, it appears that although the three components both consume the `logging` create and depend on the `slog` crate explicitly, setting the `slog` max level values in each components `Cargo.toml` is effectively a "NOP": it isn't required. The only place where it matters is in the `logging` crates `Cargo.toml`.

This took a while to test because tweaking `./logging/Cargo.toml` did _not_ rebuild the agent for example. That surprised me since the `logging` crate is an explcit dependency of the agent and `logging`'s `Cargo.toml` is part of that dependency.

Therefore, to test these changes fully, you need to do the following:

```bash
$ cd $GOPATH/src/github.com/kata-containers/kata-containers/

# Force a full rebuild
$ (cd pkg/logging && cargo clean --release)

# Build the agent
$ cd src/agent
$ make    # or "cargo build --release"
```

/cc @liubin

src/agent/vsock-exporter/Cargo.toml
```diff
@@ -15,6 +15,6 @@ serde = { version = "1.0.126", features = ["derive"] }
 tokio-vsock = "0.3.1"
 bincode = "1.3.3"
 byteorder = "1.4.3"
-slog = { version = "2.5.2", features = ["dynamic-keys", "max_level_trace", "release_max_level_info"] }
+slog = { version = "2.5.2", features = ["dynamic-keys", "max_level_trace", "release_max_level_debug"] }
```
jodh-intel, line 18: This is required because the vsock exporter doesn't consume the `logging` crate.

devimc (approved):

GabyCT (approved):

Issue #2969, opened 2021-11-04
Always run logging crate tests


jodh-intel:
As a follow-up to #2966, ensure that the tests for the `logging` crate are run before building any of its consumers: the tests are currently not run and are partially broken.

Additionally, add new tests for the debug log level issue identified in #2966.

PR #2970, opened 2021-11-04
logging: Always run crate tests

jodh-intel:
Ensure the tests in the local `logging` crate are run for all consumers
of it.

Additionally, add a new test which checks that output is generated by a
range of different log level `slog` macros. This is designed to ensure
debug level output is always available for the consumers of the
`logging` crate.

Fixes: #2969.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,12 @@ TOOLS += agent-ctl
 
 STANDARD_TARGETS = build check clean install test vendor
 
-include utils.mk
-include ./tools/packaging/kata-deploy/local-build/Makefile
+default: all
 
 all: build
+
+include utils.mk
+include ./tools/packaging/kata-deploy/local-build/Makefile
 
 # Create the rules
 $(eval $(call create_all_rules,$(COMPONENTS),$(TOOLS),$(STANDARD_TARGETS)))
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,10 @@ STANDARD_TARGETS = build check clean install test vendor
 
 default: all
 
-all: build
+all: logging-crate-tests build
+
+logging-crate-tests:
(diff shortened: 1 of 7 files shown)


jodh-intel:
@liubin - could you tal?

Makefile
```diff
@@ -17,11 +17,16 @@ TOOLS += agent-ctl
 
 STANDARD_TARGETS = build check clean install test vendor
 
+default: all
+
+all: logging-crate-tests build
```
liubin, line 22: Does `logging-crate-tests` will run twice? `build` target in sub-directories will also trigger `logging-crate-tests` too.

jodh-intel: @liubin - it might yes. The problem is we need to ensure the logging crate tests run however the build is invoked - either running `make` from the top-level, or running `make -C src/agent` for example. I'm not sure how to avoid running it multiple times whilst also ensuring that it _does_ run. Any ideas?

ftr, running the logging crate tests twice takes < 300ms on a laptop system so it's potentially unecessary to run multiple times, but it isn't going to increase the build time much.

Ideally, we want `cargo test` and `cargo test --release` to run when any of the other components run `cargo build`, but I don't think that's possible. I did think of one way, but it's horrid: create `pkg/logging/build.rs` which runs `make`! ;-)

jodh-intel: It's a hack, but I could then remove all the changes to the other makefiles.

liubin (approved):

jodh-intel:
CI green, so just need 1 more ack here.

fidencio:
Going through this one now, James.

fidencio (approved):
@jodh-intel, lgtm, but take this with a grain of salt considering my current rust skills!
Thanks!

Issue #2882, opened 2021-10-22
hypervisor: remove sandbox references


egernst:
remove any sandbox references from the hypervisor implementation

PR #2974, opened 2021-11-04
hypervisor: move sandbox specific hypervisor validation to sandbox

egernst:
Today we have a check for ensuring a kernel and initrd or image are
specified in the HypervisorConfig. Since it is possible that a
hypervisor could boot with other configurations, it would make more
sense to put this check at the sandbox level. We'll still do agnostic
hypervisor checks in hypervisor.go.

This is done in preparation for making hypervisor its own package which
is agnostic to the sandbox/virtcontainers logic.

Fixes: #2882 


Jakob-Naucke (approved):
thank you @egernst -- nothing really controversial I guess. There are some static failures still if you can TAL.

PR #2982, opened 2021-11-07
agent: Remove mounts only once when removing container

### LINKED ISSUE #2965: agent: remove_container umounts container_mounts twice
hi, all,
Recently I found something strange that container mountpoints umount twice when agent removes container.
1. https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L261 
2. https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L271

Mount point that umounted in the first place will be umounted again in the second place. So the second umount will return an error when traverse the mount points, leaving the other mount point's refCount not zero. When the container was created again, its storage will not be mounted(refCont is greater than 0) but actually it has been umounted in the first place

Burning1020:
Container mounts should be removed only once when removing container. If not, there will be an error while removing sandbox storage, making its loop finished in advance and leaving the other storage unhandled.

Fixes #2965 

Signed-off-by: Burning <burning9699@gmail.com>

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -258,8 +258,6 @@ impl AgentService {
             // Find the sandbox storage used by this container
             let mounts = sandbox.container_mounts.get(&cid);
             if let Some(mounts) = mounts {
-                remove_mounts(mounts)?;
-
                 for m in mounts.iter() {
                     if sandbox.storages.get(m).is_some() {
                         cmounts.push(m.to_string());


src/agent/src/rpc.rs
```diff
@@ -258,8 +258,6 @@ impl AgentService {
             // Find the sandbox storage used by this container
             let mounts = sandbox.container_mounts.get(&cid);
             if let Some(mounts) = mounts {
-                remove_mounts(mounts)?;
-
                 for m in mounts.iter() {
```
jodh-intel, line 261: @Burning1020 - I'm not sure this is correct - see https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L159..L165

/cc @sameo, @yyyeerbo since they have touched this part of the code.

tbh, it would be great if we could add some [unit tests](https://github.com/kata-containers/tests/blob/main/Unit-Test-Advice.md#rust-table-driven-tests) for this functionality to make the behaviour totally clear.

Burning1020: plz see that the contaier mounts were removed in the func `unset_and_remove_sandbox_storage` again, and all mounts in `sandbox.container_mounts` must be contained in `sandbox.storages` as it all comes from `add_storages`.
/cc @sameo, @yyyeerbo

egernst: Having a unit test here would definitely help clarify, though I admit, it'll be a larger test with a lot of setup.

ie, we'd want to:
```
create_sandbox
do_create_container
do_remove_container
 ** check mounts **
destory_sandbox
 ** check all mounts are gone **
 ```
 
 In this particular case, though, I agree -- we are checking and of applicable removing the mount(s0 as part of the unset_and_remove_storage() a few lines down. A unit test will help clarify. @Burning1020 I appreciate your help; is this something you'd be able to help push?

Comment (Burning1020) [reply src/agent/src/rpc.rs:261]: @egernst OK, i will try some ut. Well, you know, hard to work with RUST for me. Anyway, plz review it more carefully after my new push.

Comment (jodh-intel) [reply src/agent/src/rpc.rs:261]: Thanks @Burning1020 - please could you ensure that test only runs as `root` as we run the tests as `root` *and* non-`root`.

File: src/agent/src/rpc.rs [hunk 2]
```diff
@@ -258,8 +258,6 @@ impl AgentService {
...
                     if sandbox.storages.get(m).is_some() {
```
ManaSugi, line 262: @Burning1020 After you check that removing the mounts is a correct behavior, it would be better to change this code as follows.

```rust
if sandbox.storages.get(m).is_some() {
    sandbox.unset_and_remove_sandbox_storage(m)?;
}
```

I think `cmounts` var is no longer needed and you can also remove 268-270 lines.

Burning1020:
@egernst @jodh-intel @ManaSugi Finally I've found that the variables in `sandbox.storages` only comes from `set_sandbox_storage` which was called when storage type is either "ephemeral" or "local". And in this cases, this handlers do not return their moint point to `container.mounts`.
+ sandbox.storages:            only ephemeral and local
+ container.mounts:            others, like 9p, virtio-fs, blk...

In the container remove process, "others" were successfully unmounted in the first `remove_mounts` while "ephemeral" and "local" were unhandled because `container.mounts` doesn't contain their moint point.

Sorry, I have no idea how to fix it...

Issue #2994, opened 2021-11-09
agent: Got wrong created time of container after an exec


Tim-Zhang:
This is the bug:
https://github.com/kata-containers/kata-containers/blob/b192d388c1e2b0b612df14a3476df36d08682475/src/agent/rustjail/src/container.rs#L998

The code would run both in create_container and exec, we should only set the created time in create_container

PR #2995, opened 2021-11-09
rustjail: Fix created time of container

Tim-Zhang:
Got wrong created time of container after an exec
this commit will fix this problem.

Fixes: #2994

Signed-off-by: Tim Zhang <tim@hyper.sh>

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -995,8 +995,6 @@ impl BaseContainer for LinuxContainer {
 
         info!(logger, "entered namespaces!");
 
-        self.created = SystemTime::now();
-
         if p.init {
             let spec = self.config.spec.as_mut().unwrap();
             update_namespaces(&self.logger, spec, p.pid)?;


src/agent/rustjail/src/container.rs
```diff
@@ -995,9 +995,9 @@ impl BaseContainer for LinuxContainer {
 
         info!(logger, "entered namespaces!");
 
-        self.created = SystemTime::now();
-
         if p.init {
+            self.created = SystemTime::now();
+
```
jodh-intel, line 1000: Hi @Tim-Zhang - I'm not sure this is correct due to:

https://github.com/kata-containers/kata-containers/blob/main/src/agent/rustjail/src/container.rs#L1425

Which looks like for the non-init case we need to unset `self.created`. If that's true, we should change `self.created` to `Option<SystemTime>` though and set it to `None` initially.

Tim-Zhang: yes, it is indeed a problem, let me have a think

jodh-intel: This code is quite subtle, so when we have a fix, it would be great to add some unit tests if possible.

Tim-Zhang: @jodh-intel I have deleted this line, now we use the time when the LinuxContainer object was created as the `created` time. I think is good enough because the init status is `ContainerState::Created`, so we use set the created-time in the real-time makes sense.

And since we don't change the `created` time, the unit test can be omitted.

lifupan (approved):

jodh-intel (approved):
Thanks @Tim-Zhang.

lgtm

It would be even better if you could create a unit test that created a container, saved the `created` time and then "called" `exec` to ensure that `created` doesn't change, but I haven't looked to see how much work that would be.

fidencio:
@Tim-Zhang, could you please backport this one to the `stable-2.3` branch?

Issue #2950, opened 2021-11-02
Kata not creating symbolic links on mounted ConfigMap


hleal18:
## Kubelet version

v1.20.11+k3s1 - deployed on k3s.

v1.20.10-gke.2100 - gke cluster.

## Operating System

Ubuntu 20.04.3 LTS

## Kata version

kata-runtime: 2.2.0-alpha1 → 2.3.0-alpha2

# Description of problem

When mounting a ConfigMap the default behavior is for Kubernetes to create hidden folders and some symbolic links between them and the file on the top level. We've found that when using `kata-clh` runtime class this is not the case. Rather than symbolic links being created, the original files are used instead.

This affects the behavior of commands that scan for scripts on the mounted folder, as using the following command `find -type f` will show the same file three times: 

- The mounted file.
- The file inside the `..{DATE}` folder
- The file inside the `..data` folder

This interferes with our operation of certain helm charts as the initialization scripts loaded from a ConfigMap will mistakenly get executed three times which is not desired.

We have attached more information about the expected outcome and the behavior we're experiencing.

# Sample Manifests

The following are the Kubernetes resources used to reproduce the issue. Once deployed, exec into the container and go to the `/scripts` folder:

```bash
kubectl exec {nginx-pod-name} -i -t -- /bin/bash
cd /scripts
```

### Deployment

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 1
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: docker.io/library/nginx:latest
        ports:
        - containerPort: 80
        volumeMounts:
        - name: scripts
          mountPath: /scripts
      volumes:
      - name: scripts
        configMap:
          name: nginx-configmap
      runtimeClassName: kata-clh
```

### ConfigMap

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-configmap
data:
  init-script.sh: |
    #!/bin/bash
    echo "Hello there"
```

# Expected result

Using the previous manifests without a runtime class, inspecting the mounted folder on the `/scripts` path, which contains one file called `init-script.sh` , we can observe the expected results after running `ls -la`:

```jsx
/scripts$ ls -la
total 13
drwxrwxrwx 3 root root 4096 Oct 29 03:10 .
[...]

Issue #2946, opened 2021-11-02
watcher::tests::test_watch_entries fails occasionally


Jakob-Naucke:
across [x86_64 and s390x], I can also observe
```
---- watcher::tests::test_watch_entries stdout ----
thread 'watcher::tests::test_watch_entries' panicked at 'assertion failed: `(left == right)`
  left: `"original"`,
 right: `"updated"`', src/watcher.rs:739:9
```

_Originally posted by @Jakob-Naucke in https://github.com/kata-containers/kata-containers/issues/2809#issuecomment-940788225_

This has not been fixed by #2844.

Jakob-Naucke:
@fidencio I _suspect_ this doesn't happen on x86's GHA because its resources are single-threaded. We've seen other problems in the Rust unit tests that were due to multithreading, like #2833.

Jakob-Naucke:
Suspicions around glibc or multithreading could not be confirmed, also I cannot reproduce it on x86 any more. Possibly truly an s390x issue.

Jakob-Naucke:
here's an MWE:
```rust
async fn test_watch_entries() {
    skip_if_not_root!();
    let dir = tempfile::tempdir().expect("failed to create tempdir");
    let (storage0, _) = create_test_storage(dir.path(), "1").await.expect("failed to create storage");
    let (storage1, src1_path) = create_test_storage(dir.path(), "2").await.expect("failed to create storage");
    fs::write(src1_path.join("foo.txt"), "original").unwrap();
    let logger = slog::Logger::root(slog::Discard, o!());
    let mut entries = SandboxStorages { ..Default::default() };
    entries.add(std::iter::once(storage0), &logger).await.unwrap();
    entries.add(std::iter::once(storage1), &logger).await.unwrap();
    fs::write(src1_path.join("foo.txt"), "updated").unwrap();
    assert!(entries.check(&logger).await.is_ok());
    assert_eq!(fs::read_to_string(&entries.0[1].target_mount_point.as_path().join("foo.txt")).unwrap(), "updated");
}
```

when it doesn't work, the logs **don't** have this one line: `Copy from /tmp/.tmpHF7DnG/src2/foo.txt to /tmp/.tmpHF7DnG/dest2/foo.txt`

Jakob-Naucke:
[...]

PR #2998, opened 2021-11-09
watchers: don't dereference symlinks when copying files

egernst:
The current implementation just copies the file, dereferencing any
  simlinks in the process. This results in symlinks no being preserved,
  and a change in layout relative to the mount that we are making
  watchable.
    
 What we want is something like "cp -d"
    
This isn't available in a crate, so let's go ahead and introduce a copy
function which will create a symlink with same relative path if the
source file is a symlink. Regular files are handled with the standard
fs::copy.
   
 Introduce a unit test to verify symlinks are now handled 

Fixes: #2950

### DIFF
--- a/src/agent/src/watcher.rs
+++ b/src/agent/src/watcher.rs
@@ -49,7 +49,7 @@ struct Storage {
     /// the source becomes too large, either in number of files (>16) or total size (>1MB).
     watch: bool,
 
-    /// The list of files to watch from the source mount point and updated in the target one.
+    /// The list of files, directories, symlinks to watch from the source mount point and updated in the target one.
     watched_files: HashMap<PathBuf, SystemTime>,
 }
 
@@ -80,9 +80,13 @@ impl Drop for Storage {
 }
 
 async fn copy(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<()> {
-    // if source is a symlink, just create new symlink with same link source
     if fs::symlink_metadata(&from).await?.file_type().is_symlink() {
-        fs::symlink(fs::read_link(&from).await?, to).await?;
+        // if source is a symlink, create new symlink with same link source. If
+        // the symlink exists, remove and create new one:
+        if fs::symlink_metadata(&to).await.is_ok() {
+            fs::remove_file(&to).await?;
+        }
+        fs::symlink(fs::read_link(&from).await?, &to).await?;
     } else {
         fs::copy(from, to).await?;
     }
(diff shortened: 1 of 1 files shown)


src/agent/src/watcher.rs
```diff
@@ -110,21 +120,7 @@ impl Storage {
...
-        fs::copy(&source_file_path, &dest_file_path)
-            .await
-            .with_context(|| {
-                format!(
-                    "Copy from {} to {} failed",
-                    source_file_path.display(),
-                    dest_file_path.display()
-                )
-            })?;
+        copy(&source_file_path, &dest_file_path).await?;
```
sameo, line 123: Could we keep the error message?

src/agent/src/watcher.rs
```diff
@@ -843,6 +839,95 @@ mod tests {
...
+        tokio::fs::symlink(&src_file, &src_symlink_file)
+            .await
+            .unwrap();
+        copy(src_symlink_file, &dst_symlink_file).await.unwrap();
+        // verify destination:
+        assert!(fs::symlink_metadata(&dst_symlink_file)
+            .unwrap()
+            .file_type()
+            .is_symlink());
+        assert_eq!(fs::read_link(dst_symlink_file).unwrap(), src_file);
```
sameo, line 871: You want to take a reference on `dst_symlink_file` here and not move it.

sameo (approved):

mxpv (approved):

egernst:
We should also make sure we treat directory symlinks similar to a file. This'll require a bit more refactoring.

Moved back to draft.

fidencio:
```
thread 'watcher::tests::watch_directory_with_symlinks' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', src/watcher.rs:1005:9
test watcher::tests::watch_directory_with_symlinks ... FAILED
test watcher::tests::watch_file ... ok
test watcher::tests::watch_directory_too_large ... ok
test util::tests::test_interruptable_io_copier_reader ... ok
test watcher::tests::verify_container_cleanup_watching ... ok

failures:

failures:
    watcher::tests::watch_directory_with_symlinks

test result: FAILED. 92 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '-p kata-agent --bin kata-agent'
```
@Jakob-Naucke already mentioned that such failures may be happening due to the tests running in parallel on BM.

Jakob-Naucke:
yeeeeah it's possible this PR introduces a sibling to #2946, which I'm not super happy about. We'll have to fix it in the same way I guess.

fidencio:
Same issue.  @egernst, mind to hold this for one day?  It will give time to @Jakob-Naucke either debug it, or let you know that this test would be good candidate to be skipped on s390x.  By doing this we avoid breaking s390x CI entirely.

egernst:
debugging s390x failure:
```
thread 'watcher::tests::watch_directory_with_symlinks' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', src/watcher.rs:1005:9
 ```
 
 This seems same condition brought up in https://github.com/kata-containers/kata-containers/issues/2946

Issue #3002, opened 2021-11-10
snap build fails at qemu on ppc64le


Amulyam24:
```
/usr/bin/ld: cannot find -ludev                                                                                                              
collect2: error: ld returned 1 exit status  
ninja: build stopped: subcommand failed.
make[1]: *** [Makefile:156: run-ninja] Error 1
make: *** [GNUmakefile:11: all] Error 2
```

gkurz:
> ```
> /usr/bin/ld: cannot find -ludev                                                                                                              
> collect2: error: ld returned 1 exit status  
> ninja: build stopped: subcommand failed.
> make[1]: *** [Makefile:156: run-ninja] Error 1
> make: *** [GNUmakefile:11: all] Error 2
> ```

Hi !

Please provide the full log for this build. A variety of reasons could lead to /usr/bin/ld not finding libudev.

Amulyam24:
Hi @gkurz , the steps to reproduce this on ppc64le would be:
```
cd $GOPATH/go/src/github.com/kata-containers/kata-containers/
sudo snapcraft -d snap --destructive-mode
```

Full log of qemu build - [qemu.log](https://github.com/kata-containers/kata-containers/files/7553565/qemu.log)

gkurz:
> Hi @gkurz , the steps to reproduce this on ppc64le would be:
> 
> ```
> cd $GOPATH/go/src/github.com/kata-containers/kata-containers/
> sudo snapcraft -d snap --destructive-mode
> ```
> 
> Full log of qemu build - [qemu.log](https://github.com/kata-containers/kata-containers/files/7553565/qemu.log)

So this log has :

```
WARNING: Static library 'udev' not found for dependency 'libudev', may not be statically linked
Run-time dependency libudev found: YES 245
```

which indicates that libudev-dev is installed but as said in https://github.com/kata-containers/kata-containers/pull/3003#issuecomment-970473200 the static library isn't shipped. This causes the linking error.

I reproduce exactly the same with a freshly installed ubuntu 20.04.3 on x86_64 if I install libudev-dev.

So this isn't a question of arch, it is just that you cannot build a static QEMU with libudev under ubuntu,
and most other distros as well, since they don't have the libudev.a. The correct fix is to disable libudev
when building static as suggested in https://github.com/kata-containers/kata-containers/pull/3003#pullrequestreview-805086454 .

PR #3003, opened 2021-11-10
qemu: fix snap build on ppc64le

Amulyam24:
While building snap, static qemu is considered. Disable libudev
as it doesn't have static libraries on ppc64le.

Fixes: #3002

### DIFF
--- a/tools/packaging/scripts/configure-hypervisor.sh
+++ b/tools/packaging/scripts/configure-hypervisor.sh
@@ -310,6 +310,11 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-qom-cast-debug)
 	qemu_options+=(size:--disable-tcmalloc)
 
+	# Disable libudev for static build
+	if gt_eq "${qemu_version}" "5.2.0" ; then
+		[ "${static}" == "true" ] && qemu_options+=(size:--disable-libudev)
+	fi
+
 	# Disallow network downloads
 	qemu_options+=(security:--disable-curl)
 


jodh-intel (approved):
Thanks @Amulyam24.

lgtm

ooi, why doesn't ppc64le provide a static `libudev` library?

tools/packaging/scripts/configure-hypervisor.sh
```diff
@@ -310,6 +310,9 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-qom-cast-debug)
 	qemu_options+=(size:--disable-tcmalloc)
 
+	# Disable libudev for static build on ppc64le
+	[ "$arch" == "ppc64le" ] && qemu_options+=(size:--disable-libudev)
```
devimc, line 314: is this needed only for static build? maybe you want to check if  `"${static}" == "true"` first ... ?

Amulyam24: yes, that would be better. Thanks @devimc, updated.

gkurz (changes requested):
Does this only affect ppc64le ? Fedora 34 on x86_64 doesn't have a libudev.a for example. It seems that you should only check if a static build is requested.

Amulyam24:
> Does this only affect ppc64le ? 

Considering the static qemu build for snap, yes, this was observed on ppc64le.

> Fedora 34 on x86_64 doesn't have a libudev.a for example. It seems that you should only check if a static build is requested.

@gkurz , are you suggesting we disbale libudev on all archs and distros when building static qemu? other distros and archs might have static libraries available for it.

gkurz:
> > Does this only affect ppc64le ?
> 
> Considering the static qemu build for snap, yes, this was observed on ppc64le.
> 

Did you try to perform the same static qemu build for snap on some other arch ?

> > Fedora 34 on x86_64 doesn't have a libudev.a for example. It seems that you should only check if a static build is requested.
> 
> @gkurz , are you suggesting we disbale libudev on all archs and distros when building static qemu? other distros and archs might have static libraries available for it.

Basically yes based on the fact that:
- I'm not sure that the issue you're seeing is ppc64le specific since some distros don't ship libudev.a at all no matter the arch
- the default behavior of the script should be that it always works, i.e. doesn't try to build an unsupported configuration

Alternatively you might try to check the availibility of libudev.a .

Amulyam24:
> > > Does this only affect ppc64le ?
> > 
> > 
> > Considering the static qemu build for snap, yes, this was observed on ppc64le.
> 
> Did you try to perform the same static qemu build for snap on some other arch ?
> 
The snap test which builds static qemu is run as a part of the CI on all archs. s390x and ARM are running the build(both on Ubuntu 20.04) in their jenkins job and x86 runs it as a github action.
> > > Fedora 34 on x86_64 doesn't have a libudev.a for example. It seems that you should only check if a static build is requested.
> > 
> > 
> > @gkurz , are you suggesting we disbale libudev on all archs and distros when building static qemu? other distros and archs might have static libraries available for it.
> 
> Basically yes based on the fact that:
> 
> * I'm not sure that the issue you're seeing is ppc64le specific since some distros don't ship libudev.a at all no matter the arch

Based on the above, this occurred only on ppc64le as of now considering it being tested on Ubuntu 20.04 on all archs.
> * the default behavior of the script should be that it always works, i.e. doesn't try to build an unsupported configuration
> 
> Alternatively you might try to check the availibility of libudev.a .

Let me check on it.

clnperez:
Could it be that libudev.a is coming from somewhere we're not pulling in for Power? I don't see that provided in focal for any arch: https://packages.ubuntu.com/search?searchon=contents&keywords=libudev&mode=filename&suite=focal&arch=any

Amulyam24:
@gkurz , Thank you for confirming [here](https://github.com/kata-containers/kata-containers/issues/3002#issuecomment-971815274). Changes applied, PTAL.

gkurz (approved):
LGTM. Thanks @Amulyam24 !

Amulyam24:
@gkurz , regarding the ARM failure, looks like the option is not available
`ERROR: unknown option --disable-libudev `

gkurz:
> @gkurz , regarding the ARM failure, looks like the option is not available `ERROR: unknown option --disable-libudev `

Yeah I saw that. It seems that ARM is using QEMU 5.1.0 but `--disable-libudev` was introduced in QEMU 5.2.0.

Until ARM bumps its QEMU version, we might need to also check `gt_eq "${qemu_version}" "5.2.0" `

gkurz:
> 
> Until ARM bumps its QEMU version, we might need to also check `gt_eq "${qemu_version}" "5.2.0" `

@jongwu any chance we can just bump the QEMU version used by ARM ?

Amulyam24:
@gkurz , I applied the changes, if @jongwu can bump the QEMU version, shall revert it.

amshinde:
I havent looked too much into this, but going by this: https://patchew.org/QEMU/20201015150638.1396659-1-pbonzini@redhat.com/, it looks like we don't really need libudev for dynamic builds either. Would be worth disabling this be default. 
cc @devimc

jongwu:
Hello @gkurz @Amulyam24 -, you can try to bump qemu version to 6.1.0 to fix Arm ci.

gkurz:
> I havent looked too much into this, but going by this: https://patchew.org/QEMU/20201015150638.1396659-1-pbonzini@redhat.com/, it looks like we don't really need libudev for dynamic builds either. Would be worth disabling this be default. cc @devimc

Fair point. We already disable USB and we don't seem to be using `qemu-pr-helper`, so it seems we can safely disable it for all builds, not just static ones.

jongwu:
OK, the all the CI passed, neglact

> @gkurz , I applied the changes, if @jongwu can bump the QEMU version, shall revert it.

OK, I'll bump the qemu version for arm later.

gkurz (approved):
Let's go forward with this. We can drop the check on static build later on when we're sure we don't need libudev at all.

Issue #2999, opened 2021-11-10
kata doesn't work with older containerd like 1.3.9 and 1.4.6


bergwolf:
Those are using cri-containerd for runtime options. We should revert commit 76f16fd1a7eb5bfb04fdac0ac60f7cb5bfcb028e to keep parsing old cri-containerd runtime options.
```
    runtime: use containerd package instead of cri-containerd

    cri-containerd project has been merged into containerd repo, and
    we should not reference it any more in code and docs.

    This commit will use containerd package instead of cri-containerd
    package.
```

PR #3008, opened 2021-11-10
runtime: Revert "runtime: use containerd package instead of cri-conta…

fidencio:
…inerd"

This reverts commit 76f16fd1a7eb5bfb04fdac0ac60f7cb5bfcb028e to bring
back cri-containerd crioptions parsing so that kata works with older
containerd versions like v1.3.9 and v1.4.6.

Fixes: #2999
Signed-off-by: Peng Tao <bergwolf@hyper.sh>
(cherry picked from commit eacfcdec19a0fca5d50812b32117f722bf438b49)


bergwolf (approved):
Thanks @fidencio !

fidencio:
@jongwu, could you take a look on what's going on with the ARM CI?
It seems to be failing on some spell checker test, which I am not convinced that it should be running on an arch-dependent way.

fidencio:
> @jongwu, could you take a look on what's going on with the ARM CI? It seems to be failing on some spell checker test, which I am not convinced that it should be running on an arch-dependent way.

@Jakob-Naucke, same happening with s390x.

Do we only run the spell checker test on baremetal? Oo

fidencio:
Cherry-picking b09dd7a8833908453d326febb21ae4898d265d74 should solve the issue.

gkurz (approved):
LGTM. Thanks @fidencio !

Jakob-Naucke:
> > ARM CI seems to be failing on some spell checker test, which I am not convinced that it should be running on an arch-dependent way.
> 
> Do we only run the spell checker test on baremetal? Oo

@fidencio Ah yes, those don't run en-US or en-GB, they run en-ARM and en-IBMZ :grin: jk.
There is an `INFO: Spell checking file 'docs/design/kata-2-0-metrics.md'` in the GHA log, so we are running _something_. It seems the changed version was not actually used; this might be related to https://github.com/kata-containers/tests/issues/632?

Issue #3011, opened 2021-11-10
agent: Remove some unwraps and expects


jodh-intel:
Remove some `unwrap()` and `expect()` calls, replacing them with code to return the error to the caller.

jodh-intel:
Related: #821.

PR #3012, opened 2021-11-10
agent: Remove some unwrap and expect calls

jodh-intel:
Replace some `unwrap()` and `expect()` calls with code to return the
error to the caller.

Fixes: #3011.

### DIFF
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -11,7 +11,7 @@ use std::fmt;
 use std::fs;
 use std::os::unix::ffi::OsStrExt;
 use std::os::unix::fs::MetadataExt;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use tokio::sync::Mutex;
@@ -157,20 +157,22 @@ pub fn pcipath_to_sysfs(root_bus_sysfs: &str, pcipath: &pci::Path) -> Result<Str
         let bridgebuspath = format!("{}{}/pci_bus", root_bus_sysfs, relpath);
         let mut files: Vec<_> = fs::read_dir(&bridgebuspath)?.collect();
 
-        if files.len() != 1 {
-            return Err(anyhow!(
-                "Expected exactly one PCI bus in {}, got {} instead",
-                bridgebuspath,
-                files.len()
-            ));
-        }
-
-        // unwrap is safe, because of the length test above
-        let busfile = files.pop().unwrap()?;
-        bus = busfile
-            .file_name()
-            .into_string()
-            .map_err(|e| anyhow!("Bad filename under {}: {:?}", &bridgebuspath, e))?;
+        match files.pop() {
+            Some(busfile) if files.is_empty() => {
+                bus = busfile?
+                    .file_name()
+                    .into_string()
+                    .map_err(|e| anyhow!("Bad filename under {}: {:?}", &bridgebuspath, e))?;
+            }
+            _ => {
+                return Err(anyhow!(
+                    "Expected exactly one PCI bus in {}, got {} instead",
+                    bridgebuspath,
+                    // Adjust to original value as we've already popped
+                    files.len() + 1
+                ));
+            }
+        };
     }
 
     Ok(relpath)
@@ -218,8 +220,9 @@ impl VirtioBlkPciMatcher {
     fn new(relpath: &str) -> VirtioBlkPciMatcher {
         let root_bus = create_pci_root_bus_path();
         let re = format!(r"^{}{}/virtio[0-9]+/block/", root_bus, relpath);
+
         VirtioBlkPciMatcher {
-            rex: Regex::new(&re).unwrap(),
+            rex: Regex::new(&re).expect("BUG: failed to compile VirtioBlkPciMatcher regex"),
         }
     }
 }
@@ -257,7 +260,7 @@ impl VirtioBlkCCWMatcher {
             root_bus_path, device
         );
         VirtioBlkCCWMatcher {
-            rex: Regex::new(&re).unwrap(),
+            rex: Regex::new(&re).expect("BUG: failed to compile VirtioBlkCCWMatcher regex"),
         }
     }
 }
@@ -413,12 +416,15 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
 
     for entry in fs::read_dir(SYSFS_SCSI_HOST_PATH)? {
         let host = entry?.file_name();
-        let scan_path = format!(
-            "{}/{}/{}",
-            SYSFS_SCSI_HOST_PATH,
-            host.to_str().unwrap(),
-            "scan"
-        );
+
+        let host_str = host.to_str().ok_or_else(|| {
+            anyhow!(
+                "failed to convert directory entry to unicode for file {:?}",
+                host
+            )
+        })?;
+
+        let scan_path = PathBuf::from(&format!("{}/{}/{}", SYSFS_SCSI_HOST_PATH, host_str, "scan"));
 
         fs::write(scan_path, &scan_data)?;
     }
@@ -722,25 +728,17 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) -> 
         if vfio_in_guest {
             pci_driver_override(SYSFS_BUS_PCI_PATH, guestdev, "vfio-pci")?;
 
-            let devgroup = pci_iommu_group(SYSFS_BUS_PCI_PATH, guestdev)?;
-            if devgroup.is_none() {
-                // Devices must have an IOMMU group to be usable via VFIO
-                return Err(anyhow!("{} has no IOMMU group", guestdev));
+            // Devices must have an IOMMU group to be usable via VFIO
+            let devgroup = pci_iommu_group(SYSFS_BUS_PCI_PATH, guestdev)?
+                .ok_or_else(|| anyhow!("{} has no IOMMU group", guestdev))?;
+
+            if let Some(g) = group {
+                if g != devgroup {
+                    return Err(anyhow!("{} is not in guest IOMMU group {}", guestdev, g));
+                }
             }
 
-            if group.is_some() && group != devgroup {
-                // If PCI devices associated with the same VFIO device
-                // (and therefore group) in the host don't end up in
-                // the same group in the guest, something has gone
-                // horribly wrong
-                return Err(anyhow!(
-                    "{} is not in guest IOMMU group {}",
-                    guestdev,
-                    group.unwrap()
-                ));
-            }
-
-            group = devgroup;
+            group = Some(devgroup);
 
             pci_fixups.push((host, guestdev));
         }
@@ -748,7 +746,8 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) -> 
 
     let dev_update = if vfio_in_guest {
         // If there are any devices at all, logic above ensures that group is not None
-        let group = group.unwrap();
+        let group = group.ok_or_else(|| anyhow!("failed to get VFIO group: {:?}"))?;
+
         let vm_path = get_vfio_device_name(sandbox, group).await?;
 
         Some(DevUpdate::from_vm_path(&vm_path, vm_path.clone())?)
@@ -844,11 +843,8 @@ pub fn update_device_cgroup(spec: &mut Spec) -> Result<()> {
         .as_mut()
         .ok_or_else(|| anyhow!("Spec didn't container linux field"))?;
 
-    if linux.resources.is_none() {
-        linux.resources = Some(LinuxResources::default());
-    }
+    let resources = linux.resources.get_or_insert(LinuxResources::default());
 
-    let resources = linux.resources.as_mut().unwrap();
     resources.devices.push(LinuxDeviceCgroup {
         allow: false,
         major: Some(major),
--- a/src/agent/src/main.rs
+++ b/src/agent/src/main.rs
@@ -113,10 +113,10 @@ async fn create_logger_task(rfd: RawFd, vsock_port: u32, shutdown: Receiver<bool
         )?;
 
         let addr = SockAddr::new_vsock(libc::VMADDR_CID_ANY, vsock_port);
-        socket::bind(listenfd, &addr).unwrap();
-        socket::listen(listenfd, 1).unwrap();
+        socket::bind(listenfd, &addr)?;
+        socket::listen(listenfd, 1)?;
 
-        writer = Box::new(util::get_vsock_stream(listenfd).await.unwrap());
+        writer = Box::new(util::get_vsock_stream(listenfd).await?);
     } else {
         writer = Box::new(tokio::io::stdout());
     }
@@ -326,7 +326,7 @@ async fn start_sandbox(
     sandbox.lock().await.sender = Some(tx);
 
     // vsock:///dev/vsock, port
-    let mut server = rpc::start(sandbox.clone(), config.server_addr.as_str());
+    let mut server = rpc::start(sandbox.clone(), config.server_addr.as_str())?;
     server.start().await?;
 
     rx.await?;
--- a/src/agent/src/metrics.rs
+++ b/src/agent/src/metrics.rs
@@ -8,6 +8,7 @@ extern crate procfs;
 use prometheus::{Encoder, Gauge, GaugeVec, IntCounter, TextEncoder};
 
 use anyhow::Result;
+use slog::warn;
 use tracing::instrument;
 
 const NAMESPACE_KATA_AGENT: &str = "kata_agent";
@@ -74,7 +75,7 @@ pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
     AGENT_SCRAPE_COUNT.inc();
 
     // update agent process metrics
-    update_agent_metrics();
+    update_agent_metrics()?;
 
     // update guest os metrics
     update_guest_metrics();
@@ -84,23 +85,26 @@ pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
 
     let mut buffer = Vec::new();
     let encoder = TextEncoder::new();
-    encoder.encode(&metric_families, &mut buffer).unwrap();
+    encoder.encode(&metric_families, &mut buffer)?;
 
-    Ok(String::from_utf8(buffer).unwrap())
+    Ok(String::from_utf8(buffer)?)
 }
 
 #[instrument]
-fn update_agent_metrics() {
+fn update_agent_metrics() -> Result<()> {
     let me = procfs::process::Process::myself();
 
-    if let Err(err) = me {
-        error!(sl!(), "failed to create process instance: {:?}", err);
-        return;
-    }
+    let me = match me {
+        Ok(p) => p,
+        Err(e) => {
+            // FIXME: return Ok for all errors?
+            warn!(sl!(), "failed to create process instance: {:?}", e);
 
-    let me = me.unwrap();
+            return Ok(());
+        }
+    };
 
-    let tps = procfs::ticks_per_second().unwrap();
+    let tps = procfs::ticks_per_second()?;
 
     // process total time
     AGENT_TOTAL_TIME.set((me.stat.utime + me.stat.stime) as f64 / (tps as f64));
@@ -109,7 +113,7 @@ fn update_agent_metrics() {
     AGENT_TOTAL_VM.set(me.stat.vsize as f64);
 
     // Total resident set
-    let page_size = procfs::page_size().unwrap() as f64;
+    let page_size = procfs::page_size()? as f64;
     AGENT_TOTAL_RSS.set(me.stat.rss as f64 * page_size);
 
     // io
@@ -132,11 +136,11 @@ fn update_agent_metrics() {
     }
 
     match me.status() {
-        Err(err) => {
-            info!(sl!(), "failed to get process status: {:?}", err);
-        }
+        Err(err) => error!(sl!(), "failed to get process status: {:?}", err),
         Ok(status) => set_gauge_vec_proc_status(&AGENT_PROC_STATUS, &status),
     }
+
+    return Ok(());
 }
 
 #[instrument]
--- a/src/agent/src/metrics.rs
+++ b/src/agent/src/metrics.rs
@@ -140,7 +140,7 @@ fn update_agent_metrics() -> Result<()> {
         Ok(status) => set_gauge_vec_proc_status(&AGENT_PROC_STATUS, &status),
     }
 
-    return Ok(());
+    Ok(())
 }
 
 #[instrument]
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -139,8 +139,8 @@ pub const STORAGE_HANDLER_LIST: &[&str] = &[
 
 #[instrument]
 pub fn baremount(
-    source: &str,
-    destination: &str,
+    source: &Path,
+    destination: &Path,
     fs_type: &str,
     flags: MsFlags,
     options: &str,
@@ -148,11 +148,11 @@ pub fn baremount(
 ) -> Result<()> {
     let logger = logger.new(o!("subsystem" => "baremount"));
 
-    if source.is_empty() {
+    if source.as_os_str().is_empty() {
         return Err(anyhow!("need mount source"));
     }
 
-    if destination.is_empty() {
+    if destination.as_os_str().is_empty() {
         return Err(anyhow!("need mount destination"));
     }
 
@@ -444,16 +444,19 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
     let options_vec = options_vec.iter().map(String::as_str).collect();
     let (flags, options) = parse_mount_flags_and_options(options_vec);
 
+    let source = Path::new(&storage.source);
+    let mount_point = Path::new(&storage.mount_point);
+
     info!(logger, "mounting storage";
-    "mount-source:" => storage.source.as_str(),
-    "mount-destination" => storage.mount_point.as_str(),
+    "mount-source" => source.display(),
+    "mount-destination" => mount_point.display(),
     "mount-fstype"  => storage.fstype.as_str(),
     "mount-options" => options.as_str(),
     );
 
     baremount(
-        storage.source.as_str(),
-        storage.mount_point.as_str(),
+        source,
+        mount_point,
         storage.fstype.as_str(),
         flags,
         options.as_str(),
@@ -579,7 +582,10 @@ fn mount_to_rootfs(logger: &Logger, m: &InitMount) -> Result<()> {
 
     fs::create_dir_all(Path::new(m.dest)).context("could not create directory")?;
 
-    baremount(m.src, m.dest, m.fstype, flags, &options, logger).or_else(|e| {
+    let source = Path::new(m.src);
+    let dest = Path::new(m.dest);
+
+    baremount(source, dest, m.fstype, flags, &options, logger).or_else(|e| {
         if m.src != "dev" {
             return Err(e);
         }
@@ -937,14 +943,10 @@ mod tests {
                 std::fs::create_dir_all(d).expect("failed to created directory");
             }
 
-            let result = baremount(
-                &src_filename,
-                &dest_filename,
-                d.fs_type,
-                d.flags,
-                d.options,
-                &logger,
-            );
+            let src = Path::new(&src_filename);
+            let dest = Path::new(&dest_filename);
+
+            let result = baremount(src, dest, d.fs_type, d.flags, d.options, &logger);
 
             let msg = format!("{}: result: {:?}", msg, result);
 
@@ -1021,15 +1023,11 @@ mod tests {
                 .unwrap_or_else(|_| panic!("failed to create directory {}", d));
         }
 
+        let src = Path::new(mnt_src_filename);
+        let dest = Path::new(mnt_dest_filename);
+
         // Create an actual mount
-        let result = baremount(
-            mnt_src_filename,
-            mnt_dest_filename,
-            "bind",
-            MsFlags::MS_BIND,
-            "",
-            &logger,
-        );
+        let result = baremount(src, dest, "bind", MsFlags::MS_BIND, "", &logger);
         assert!(result.is_ok(), "mount for test setup failed");
 
         let tests = &[
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -628,8 +628,7 @@ pub fn get_mount_fs_type_from_file(mount_file: &str, mount_point: &str) -> Resul
     let file = File::open(mount_file)?;
     let reader = BufReader::new(file);
 
-    let re = Regex::new(format!("device .+ mounted on {} with fstype (.+)", mount_point).as_str())
-        .unwrap();
+    let re = Regex::new(format!("device .+ mounted on {} with fstype (.+)", mount_point).as_str())?;
 
     // Read the file line by line using the lines() iterator from std::io::BufRead.
     for (_index, line) in reader.lines().enumerate() {
@@ -707,20 +706,21 @@ pub fn get_cgroup_mounts(
             }
         }
 
-        if fields[0].is_empty() {
+        let subsystem_name = fields[0];
+
+        if subsystem_name.is_empty() {
             continue;
         }
 
-        if fields[0] == "devices" {
+        if subsystem_name == "devices" {
             has_device_cgroup = true;
         }
 
-        if let Some(value) = CGROUPS.get(&fields[0]) {
-            let key = CGROUPS.keys().find(|&&f| f == fields[0]).unwrap();
+        if let Some((key, value)) = CGROUPS.get_key_value(subsystem_name) {
             cg_mounts.push(InitMount {
                 fstype: "cgroup",
                 src: "cgroup",
-                dest: *value,
+                dest: value,
                 options: vec!["nosuid", "nodev", "noexec", "relatime", key],
             });
         }
@@ -773,10 +773,9 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
         return Err(anyhow!("{:?} exists but is not a regular file", path));
     }
 
-    // The only way parent() can return None is if the path is /,
-    // which always exists, so the test above will already have caught
-    // it, thus the unwrap() is safe
-    let dir = path.parent().unwrap();
+    let dir = path
+        .parent()
+        .ok_or_else(|| anyhow!("failed to find parent path for {:?}", path))?;
 
     fs::create_dir_all(dir).context(format!("create_dir_all {:?}", dir))?;
 
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -183,7 +183,7 @@ impl AgentService {
         update_device_cgroup(&mut oci)?;
 
         // Append guest hooks
-        append_guest_hooks(&s, &mut oci);
+        append_guest_hooks(&s, &mut oci)?;
 
         // write spec to bundle path, hooks might
         // read ocispec
@@ -205,21 +205,14 @@ impl AgentService {
             LinuxContainer::new(cid.as_str(), CONTAINER_BASE, opts, &sl!())?;
 
         let pipe_size = AGENT_CONFIG.read().await.container_pipe_size;
-        let p = if oci.process.is_some() {
-            Process::new(
-                &sl!(),
-                oci.process.as_ref().unwrap(),
-                cid.as_str(),
-                true,
-                pipe_size,
-            )?
+
+        let p = if let Some(p) = oci.process {
+            Process::new(&sl!(), &p, cid.as_str(), true, pipe_size)?
         } else {
             info!(sl!(), "no process configurations!");
             return Err(anyhow!(nix::Error::from_errno(nix::errno::Errno::EINVAL)));
         };
-
         ctr.start(p).await?;
-
         s.update_shared_pidns(&ctr)?;
         s.add_container(ctr);
         info!(sl!(), "created container!");
@@ -241,11 +234,17 @@ impl AgentService {
 
         ctr.exec()?;
 
+        if sid == cid {
+            return Ok(());
+        }
+
         // start oom event loop
-        if sid != cid && ctr.cgroup_manager.is_some() {
-            let cg_path = ctr.cgroup_manager.as_ref().unwrap().get_cg_path("memory");
-            if cg_path.is_some() {
-                let rx = notifier::notify_oom(cid.as_str(), cg_path.unwrap()).await?;
+        if let Some(ref ctr) = ctr.cgroup_manager {
+            let cg_path = ctr.get_cg_path("memory");
+
+            if let Some(cg_path) = cg_path {
+                let rx = notifier::notify_oom(cid.as_str(), cg_path.to_string()).await?;
+
                 s.run_oom_event_monitor(rx, cid.clone()).await;
             }
         }
@@ -345,14 +344,13 @@ impl AgentService {
         let s = self.sandbox.clone();
         let mut sandbox = s.lock().await;
 
-        let process = if req.process.is_some() {
-            req.process.as_ref().unwrap()
-        } else {
-            return Err(anyhow!(nix::Error::from_errno(nix::errno::Errno::EINVAL)));
-        };
+        let process = req
+            .process
+            .into_option()
+            .ok_or_else(|| anyhow!(nix::Error::from_errno(nix::errno::Errno::EINVAL)))?;
 
         let pipe_size = AGENT_CONFIG.read().await.container_pipe_size;
-        let ocip = rustjail::process_grpc_to_oci(process);
+        let ocip = rustjail::process_grpc_to_oci(&process);
         let p = Process::new(&sl!(), &ocip, exec_id.as_str(), false, pipe_size)?;
 
         let ctr = sandbox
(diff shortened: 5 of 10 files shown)


src/agent/src/device.rs
```diff
@@ -167,20 +167,18 @@ pub fn pcipath_to_sysfs(root_bus_sysfs: &str, pcipath: &pci::Path) -> Result<Str
         let bridgebuspath = format!("{}{}/pci_bus", root_bus_sysfs, relpath);
         let mut files: Vec<_> = fs::read_dir(&bridgebuspath)?.collect();
 
-        if files.len() != 1 {
```
liubin, line 170: Is it safe to delete the check of `files.len() != 1`?

src/agent/src/device.rs
```diff
@@ -671,31 +675,25 @@ async fn vfio_device_handler(
...
             pci_driver_override(SYSFS_BUS_PCI_PATH, guestdev, "vfio-pci")?;
 
-            let devgroup = pci_iommu_group(SYSFS_BUS_PCI_PATH, guestdev)?;
-            if devgroup.is_none() {
-                // Devices must have an IOMMU group to be usable via VFIO
-                return Err(anyhow!("{} has no IOMMU group", guestdev));
-            }
+            let devgroup = pci_iommu_group(SYSFS_BUS_PCI_PATH, guestdev)?
+                .ok_or(format!("{} has no IOMMU group", guestdev))
+                .map_err(|e| anyhow!(e))?;
```
dgibson, line 680: I think you can simplify this to just
```
.ok_or_else(|| anyhow!("{} has no IOMMU group", guestdev))?;
```

src/agent/src/device.rs
```diff
@@ -671,31 +675,25 @@ async fn vfio_device_handler(
...
+            group = Some(devgroup);
         }
     }
 
     if vfio_in_guest {
         // If there are any devices at all, logic above ensures that group is not None
-        let group = group.unwrap();
+        let group = group
+            .ok_or("failed to get VFIO group")
+            .map_err(|e| anyhow!(e))?;
```
dgibson, line 696: Again, I believe you can just do `.ok_or_else(|| anyhow!(...))?`

src/agent/src/device.rs
```diff
@@ -799,7 +797,12 @@ pub fn update_device_cgroup(spec: &mut Spec) -> Result<()> {
         linux.resources = Some(LinuxResources::default());
     }
 
-    let resources = linux.resources.as_mut().unwrap();
+    let resources = linux
+        .resources
+        .as_mut()
+        .ok_or("failed to get Linux resources")
+        .map_err(|e| anyhow!(e))?;
+
```
dgibson, line 805: `unwrap()` is appropriate here, since we literally just handled the case where `linux.resources` was `None` by populating it with a default.  It wouldn't surprise me if the compiler already optimzed the `unwrap()` to a no-op..

src/agent/src/main.rs
```diff
@@ -113,10 +113,10 @@ async fn create_logger_task(rfd: RawFd, vsock_port: u32, shutdown: Receiver<bool
         )?;
 
         let addr = SockAddr::new_vsock(libc::VMADDR_CID_ANY, vsock_port);
-        socket::bind(listenfd, &addr).unwrap();
-        socket::listen(listenfd, 1).unwrap();
+        socket::bind(listenfd, &addr)?;
+        socket::listen(listenfd, 1)?;
 
-        writer = Box::new(util::get_vsock_stream(listenfd).await.unwrap());
+        writer = Box::new(util::get_vsock_stream(listenfd).await?);
```
dgibson, line 119: These are all good ideas.

src/agent/src/mount.rs
```diff
@@ -622,8 +622,7 @@ pub fn get_mount_fs_type_from_file(mount_file: &str, mount_point: &str) -> Resul
     let file = File::open(mount_file)?;
     let reader = BufReader::new(file);
 
-    let re = Regex::new(format!("device .+ mounted on {} with fstype (.+)", mount_point).as_str())
-        .unwrap();
+    let re = Regex::new(format!("device .+ mounted on {} with fstype (.+)", mount_point).as_str())?;
```
dgibson, line 625: As with the other use of Regex, a failure here indicates a bug in the code, not a problem in the inputs, which makes `unwrap()` appropriate.

src/agent/src/mount.rs
```diff
@@ -767,10 +771,10 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
...
     }
 
-    // The only way parent() can return None is if the path is /,
-    // which always exists, so the test above will already have caught
-    // it, thus the unwrap() is safe
-    let dir = path.parent().unwrap();
+    let dir = path
+        .parent()
+        .ok_or(format!("failed to find parent path for {:?}", path))
+        .map_err(|e| anyhow!(e))?;
```
dgibson, line 777: As noted in the comment, we've already tested for the failure case, so `unwrap()` is reasonable here.

src/agent/src/rpc.rs
```diff
@@ -111,15 +111,26 @@ pub struct AgentService {
...
+        .enumerate()
+        .filter(|(i, c)| match *i {
+            // Check 1st char
+            0 => c.is_alphanumeric(),
+            // Check remaining chars
+            _ => (c.is_alphanumeric() || ['.', '-', '_'].contains(&c)),
+        })
+        // Remove the index
+        .map(|(_i, c)| c)
+        .collect::<String>();
```
dgibson, line 126: This seems a pretty complex way to do this.  How about
```
let chars = id.chars();
let valid = if let Some(first) = chars.next() {
    first.is_alpha_numeric() && chars().all(...)
} else {
    false
}
```

src/agent/src/rpc.rs
```diff
@@ -201,7 +212,9 @@ impl AgentService {
         let p = if oci.process.is_some() {
             Process::new(
                 &sl!(),
-                oci.process.as_ref().unwrap(),
+                oci.process
+                    .as_ref()
+                    .ok_or_else(|| anyhow!("cannot get OCI process"))?,
```
dgibson, line 217: We've checked just above that `oci.process.is_some()`.  Instead we should use an `if let` to avoid the `unwrap()`.

src/agent/src/rpc.rs
```diff
@@ -236,9 +249,18 @@ impl AgentService {
...
+                .ok_or_else(|| anyhow!("cannot get cgroup manager"))?
+                .get_cg_path("memory");
+
             if cg_path.is_some() {
-                let rx = notifier::notify_oom(cid.as_str(), cg_path.unwrap()).await?;
+                let rx = notifier::notify_oom(
+                    cid.as_str(),
+                    cg_path.ok_or_else(|| anyhow!("cannot get cgroup path"))?,
+                )
+                .await?;
```
dgibson, line 263: Same for both these cases, we should use an `if let` instead of `if XXX.is_some() { XXX.unwrap()`.

src/agent/src/rpc.rs
```diff
@@ -378,7 +403,7 @@ impl AgentService {
 
         let p = find_process(&mut sandbox, cid.as_str(), eid.as_str(), init)?;
 
-        let mut signal = Signal::try_from(req.signal as i32).unwrap();
+        let mut signal = Signal::try_from(req.signal as i32)?;
```
dgibson, line 406: The error message direct from `try_from` probably won't be very useful, so you probably want to use a `.context()` or similar here.

src/agent/src/rpc.rs
```diff
@@ -639,8 +669,8 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
 
         let resp = Empty::new();
 
-        if res.is_some() {
-            let oci_res = rustjail::resources_grpc_to_oci(&res.unwrap());
+        if let Some(res) = res.as_ref() {
```
dgibson, line 672: That's the way!  There are a bunch of places elsewhere in this patch (and elsewhere in the code in general) where this construct is what you want.

src/agent/src/rpc.rs
```diff
@@ -1020,12 +1049,21 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+            .map_err(|e| ttrpc_error(ttrpc::Code::INTERNAL, e.to_string()))?;
         // Close get_oom_event connection,
         // otherwise it will block the shutdown of ttrpc.
         sandbox.event_tx.take();
 
-        sandbox.sender.take().unwrap().send(1).unwrap();
+        sandbox
+            .sender
+            .take()
+            .ok_or("failed to get sandbox")
```
dgibson, line 1063: Not an accurate message: we've got the sandbox just fine, it just didn't have a populated `sender` field.

src/agent/src/rpc.rs
```diff
@@ -1020,12 +1049,21 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+            .map_err(|e| ttrpc_error(ttrpc::Code::INTERNAL, e.to_string()))?
```
dgibson, line 1064: Also, you can collapse the `.ok_or(...).map_err(...)` chain into just an `.ok_or_else()`.

src/agent/src/rpc.rs
```diff
@@ -1432,23 +1472,39 @@ fn update_container_namespaces(
...
-        pid_ns.path = String::from(sandbox.sandbox_pidns.as_ref().unwrap().path.as_str());
+        pid_ns.path = String::from(
+            sandbox
+                .sandbox_pidns
+                .as_ref()
+                .ok_or("failed to get sandbox ref")
+                .map_err(|e| anyhow!(e))?
+                .path
+                .as_str(),
+        );
```
dgibson, line 1483: You have an `is_some()` check above, so again you want `if let` rather than returning an error.

src/agent/src/rpc.rs
```diff
@@ -1432,23 +1472,39 @@ fn update_container_namespaces(
...
+        return Ok(());
     }
-    let guest_hooks = s.hooks.as_ref().unwrap();
+
+    let guest_hooks = s
+        .hooks
+        .as_ref()
+        .ok_or("failed to get guest hook reference")
+        .map_err(|e| anyhow!(e))?;
+
```
dgibson, line 1500: Again the `is_none()` case is already dealt with so you want an `if let` (or `match`).

src/agent/src/rpc.rs
```diff
@@ -1538,7 +1594,12 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
         PathBuf::from("/")
     };
 
-    fs::create_dir_all(dir.to_str().unwrap()).or_else(|e| {
+    fs::create_dir_all(
+        dir.to_str()
+            .ok_or("failed to get directory to create")
+            .map_err(|e| anyhow!(e))?,
+    )
+    .or_else(|e| {
```
dgibson, line 1602: So, this code is taking a `Path`, converting to a string (which might fail) then passing it to `std::fs` function which actually wants a `Path`.  You should be able to just remove the `to_str().unwrap()` instead.

src/agent/src/rpc.rs
```diff
@@ -1547,21 +1608,28 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
     })?;
 
     std::fs::set_permissions(
-        dir.to_str().unwrap(),
+        dir.to_str()
```
dgibson, line 1611: Same here.

src/agent/src/rpc.rs
```diff
@@ -1547,21 +1608,28 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
...
+        .to_str()
+        .ok_or("failed to get tmpfile string")
+        .map_err(|e| anyhow!(e))?;
+
     let file = OpenOptions::new()
         .write(true)
         .create(true)
         .truncate(false)
-        .open(tmpfile.to_str().unwrap())?;
+        .open(tmpfile_name)?;
```
dgibson, line 1629: And here.

src/agent/src/rpc.rs
```diff
@@ -1610,30 +1678,50 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
     if spec.root.is_none() {
         return Err(nix::Error::Sys(Errno::EINVAL).into());
     }
-    let spec_root = spec.root.as_ref().unwrap();
+
+    let spec_root = spec
+        .root
+        .as_ref()
+        .ok_or("failed to get spec root as string")
+        .map_err(|e| anyhow!(e))?;
```
dgibson, line 1686: `if let` or `match` again.

src/agent/src/rpc.rs
```diff
@@ -1610,30 +1678,50 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
...
+    let rootfs_path_name = rootfs_path
+        .to_str()
+        .ok_or("failed to get rootfs path name")
+        .map_err(|e| anyhow!(e))?;
+
     fs::create_dir_all(&rootfs_path)?;
     baremount(
         &spec_root.path,
-        rootfs_path.to_str().unwrap(),
+        rootfs_path_name,
```
dgibson, line 1700: The better fix would be to make `baremount` take `Path`s rather than bare `&str` making the switches back and forth unnecessary.

src/agent/src/rpc.rs
```diff
@@ -1666,8 +1754,8 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
 
     match status.code() {
         Some(code) => {
-            let std_out: String = String::from_utf8(output.stdout).unwrap();
-            let std_err: String = String::from_utf8(output.stderr).unwrap();
+            let std_out: String = String::from_utf8(output.stdout)?;
+            let std_err: String = String::from_utf8(output.stderr)?;
```
dgibson, line 1758: The only thing these values are used for is debug output, so you can use `String::from_utf8_lossy()` instead.

src/agent/src/util.rs
```diff
@@ -104,7 +108,9 @@ mod tests {
 
             let mut vec_locked = vec_ref.lock();
 
-            let mut v = vec_locked.as_deref_mut().unwrap();
+            let mut v = vec_locked
+                .as_deref_mut()
+                .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e.to_string()))?;
```
dgibson, line 113: Since this is only used in the unit tests, `unwrap()` is fine - it's just a specialized `assert`.

dgibson (changes requested):
So, I like the idea of getting rid of `unwrap()`s.  However it need to be done with care and attention to each case, not just blanket changing things to return an error.

src/agent/src/device.rs
```diff
@@ -225,12 +223,13 @@ struct VirtioBlkPciMatcher {
...
+    fn new(relpath: &str) -> Result<VirtioBlkPciMatcher> {
         let root_bus = create_pci_root_bus_path();
         let re = format!(r"^{}{}/virtio[0-9]+/block/", root_bus, relpath);
-        VirtioBlkPciMatcher {
-            rex: Regex::new(&re).unwrap(),
-        }
+
+        Ok(VirtioBlkPciMatcher {
+            rex: Regex::new(&re)?,
+        })
```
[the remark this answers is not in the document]
jodh-intel: I understand exactly what you're saying, but:

- API: If there was no point in checking the value from `Regex::new()`, it wouldn't return an error.
- Maintenance: Although the code is *currently* correct, who knows what changes that code will under go in the future, so I'd always vote for the most cautious / paranoid / defensive strategy.
- `?` is shorter and less visually distracting than `.unwrap()` :smile:

src/agent/src/device.rs
```diff
@@ -503,9 +503,13 @@ fn update_spec_device(
...
-            // unwrap is safe, because residx would be empty if there
-            // were no resources
-            let res = &mut linux.resources.as_mut().unwrap().devices[*ridx];
+            let res = &mut linux
+                .resources
+                .as_mut()
+                .ok_or("failed to get Linux resources")
+                .map_err(|e| anyhow!(e))?
+                .devices[*ridx];
+
```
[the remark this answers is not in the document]
jodh-intel: And I read that comment but I'd rather codify that expectation in code rather than a comment. Again, I'm thinking of future code changes here.

src/agent/src/device.rs
```diff
@@ -671,31 +675,25 @@ async fn vfio_device_handler(
...
-                // the same group in the guest, something has gone
-                // horribly wrong
-                return Err(anyhow!(
-                    "{} is not in guest IOMMU group {}",
-                    guestdev,
-                    group.unwrap()
-                ));
+            if let Some(g) = group {
+                if g != devgroup {
+                    return Err(anyhow!("{} is not in guest IOMMU group {}", guestdev, g));
```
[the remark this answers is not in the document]
jodh-intel: Re-added.

jodh-intel (commented):
Thanks for your very thorough review @dgibson! I've had this branch lying around for a while but it keeps getting conflicted, so rather than "solving the world" my current strategy is to attack this (large) problem in small parts over a number of PRs (and I'll probably need help! :)

I think I've applied all your suggestions but it's a big patch with a lot of comments.

Note that I have also reworked `baremount()` as you suggested. I can split that into a separate commit next week if required.

jodh-intel:
Spotted a comment I haven't yet applied so adding dnm...

src/agent/src/device.rs
```diff
@@ -175,12 +175,12 @@ pub fn pcipath_to_sysfs(root_bus_sysfs: &str, pcipath: &pci::Path) -> Result<Str
             ));
```
dgibson, line 175: The first paragraph of your commit message is no terribly accurate.

Splitting the changes to baremount() into a preliminary patch would probably make this easier to review (and to rebase)

src/agent/src/device.rs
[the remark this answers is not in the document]
dgibson: Well, I don't feel strongly enough about this to insist on this point, but I do disagree here:

> I understand exactly what you're saying, but:
> 
>     * API: If there was no point in checking the value from `Regex::new()`, it wouldn't return an error.

The API makes sense for `Regex::new` because it could be getting user data, so needs to validate and report it.  But in our case the regex is a literal, so a failure here is a bug in our code not a situational problem.

>     * Maintenance: Although the code is _currently_ correct, who knows what changes that code will under go in the future, so I'd always vote for the most cautious / paranoid / defensive strategy.

Theoretically, yes, but this is not fragile against likely code changes.  In any situation I can imagine, it would be a code bug for us to generate an invalid regex string here.

>     * `?` is shorter and less visually distracting than `.unwrap()` smile

Well, sure.

My counterpoint is that as a rule problems that are necessarily bugs in our code should be reported as asserts or panics, not as error codes.  It clearly distinguishes a "this should not happen, ever" case from "maybe my configuration is bad", and it puts the crash as close to the bug as we can, which helps with debugging.

Admittedly this is offset in the specific case of Kata, since the way we package it means that agent panics aren't properly reported up to somewhere useful without jumping through a lot of hoops :(.

src/agent/src/device.rs
```diff
@@ -423,12 +424,13 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
-        let scan_path = format!(
-            "{}/{}/{}",
-            SYSFS_SCSI_HOST_PATH,
-            host.to_str().unwrap(),
-            "scan"
-        );
+
+        let host_str = host
+            .to_str()
+            .ok_or(format!("failed to get host string for file {:?}", host))
```
dgibson, line 430: Not really accurate.  We have the string (you're debug printing it right there), it just didn't convert to unicode properly.  Again this would be a lot smoother if we used `Path` instead of `str` throughout here.

src/agent/src/device.rs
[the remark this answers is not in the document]
dgibson: Please leave it.  Again, I'm completely replacing this code block in #2927, so it's pretty much moot.  The new version doesn't have an equivalent case.

src/agent/src/device.rs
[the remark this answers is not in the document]
dgibson: Uh.. I meant the `// Devices must have an IOMMU group to be usable via VFIO` which still appears to be gone.

src/agent/src/device.rs
```diff
@@ -671,31 +677,27 @@ async fn vfio_device_handler(
...
-            group = devgroup;
+            group = Some(devgroup);
         }
     }
 
     if vfio_in_guest {
         // If there are any devices at all, logic above ensures that group is not None
-        let group = group.unwrap();
+        let group = group.ok_or_else(|| anyhow!("failed to get VFIO group: {:?}"))?;
+
```
dgibson, line 700: Again, we've ensured this is the case in the code enough, so reporting an error is misleading.  A failure here is definitely a code bug, hence an assert (which unwrap() is a specialized form of) is appropriate.

src/agent/src/metrics.rs
```diff
@@ -84,23 +84,23 @@ pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
...
 
 #[instrument]
-fn update_agent_metrics() {
+fn update_agent_metrics() -> Result<()> {
     let me = procfs::process::Process::myself();
 
     if let Err(err) = me {
         error!(sl!(), "failed to create process instance: {:?}", err);
-        return;
+        return Ok(());
```
dgibson, line 98: Given we just printed an error, should we really be returning `Ok(())`?

src/agent/src/metrics.rs
```diff
@@ -84,23 +84,23 @@ pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
...
     }
 
-    let me = me.unwrap();
+    let me = me?;
```
dgibson, line 101: We just disposed of the error case above.  We can avoid both `unwrap()` and `?` if you combine the two cases into a `match`.
```
let me = match me {
    Ok(v) => v,
    Err(e) => {
        error!(sl!(), "failed to create process instance: {:?}", err);
        return Ok(());
}
```

src/agent/src/mount.rs
```diff
@@ -444,16 +444,19 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
     let options_vec = options_vec.iter().map(String::as_str).collect();
     let (flags, options) = parse_mount_flags_and_options(options_vec);
 
+    let source = Path::new(&storage.source);
+    let mount_point = Path::new(&storage.mount_point);
+
```
dgibson, line 449: `Path::new` returns an `&Path` not a bare `Path` (which is unsized)....

src/agent/src/mount.rs
```diff
@@ -444,16 +444,19 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+    "mount-destination" => mount_point.display(),
     "mount-fstype"  => storage.fstype.as_str(),
     "mount-options" => options.as_str(),
     );
 
     baremount(
-        storage.source.as_str(),
-        storage.mount_point.as_str(),
+        &source,
+        &mount_point,
```
dgibson, line 459: ..which means you don't need an extra `&` here.

src/agent/src/util.rs
```diff
@@ -159,6 +165,7 @@ mod tests {
     }
 
     impl ToString for BufWriter {
+        // No way to return an error here, hence the unwrap calls ;(
```
dgibson, line 168: We're inside test code, so asserts are standard practice, and `unwrap()` is just a specialized assert.  I don't think this needs commenting on.

tools/agent-ctl/src/main.rs
```diff
@@ -156,8 +156,10 @@ fn connect(name: &str, global_args: clap::ArgMatches) -> Result<()> {
             .collect();
```
dgibson, line 156: Having two commits immediately after each other with identical one-line summaries is probably not a great idea.

tools/agent-ctl/src/main.rs
```diff
@@ -169,10 +171,11 @@ fn connect(name: &str, global_args: clap::ArgMatches) -> Result<()> {
         None => 0,
     };
 
-    let hybrid_vsock_port: u64 = args
+    let hybrid_vsock_port = args
         .value_of("hybrid-vsock-port")
         .ok_or("Need Hybrid VSOCK port number")
-        .map(|p| p.parse::<u64>().unwrap())
+        .map_err(|e| anyhow!(e))?
```
dgibson, line 177: Again you can collapse `.ok_or().map_err()` into just `.ok_or_else()` (or, theoretically to just `.ok_or()`, but `.ok_or_else()` is preferred, since it avoids constructing the error object on the happy path).

src/agent/src/device.rs
jodh-intel: @dgibson - Yep - currently doing that...

src/agent/src/device.rs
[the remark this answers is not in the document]
jodh-intel: Ok. I've removed the `Result` return value from `VirtioBlkPciMatcher` and changed the original `unwrap()` to an `expect`()` for `VirtioBlkPciMatcher` and `VirtioBlkCCWMatcher` so we get a slightly better message that shows this is an internal bug.

The remaining `Regex::new()` shouldn't use `unwrap()` though since the string passed to that call is coming via a parameter (aka untrusted source).

src/agent/src/device.rs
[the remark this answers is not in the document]
jodh-intel: Ok - I've reverted this change.

src/agent/src/device.rs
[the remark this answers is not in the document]
jodh-intel: Fixed.

jodh-intel:
@dgibson - I've reworked these changes further so please can you tal?

@liubin - could you also tal?

src/agent/src/rpc.rs
```diff
@@ -238,13 +240,23 @@ impl AgentService {
...
+
+            if let Some(cg_path) = cg_path {
+                let rx = notifier::notify_oom(cid.as_str(), cg_path.to_string()).await?;
+
                 s.run_oom_event_monitor(rx, cid.clone()).await;
+            } else {
+                return Err(anyhow!("cannot get cgroup path"));
             }
+        } else {
+            return Err(anyhow!("cannot get cgroup manager"));
```
dgibson, line 259: This is changing behaviour, not just removing the `unwrap`, though I can't immediately tell which behaviour is correct: AFAICT the old version would just carry on in the `!ctr.cgroup_manager.is_some()` and `!cg_path.is_some()` cases, rather than returning an error.

src/agent/src/rpc.rs
```diff
@@ -343,7 +355,10 @@ impl AgentService {
         let mut sandbox = s.lock().await;
 
         let process = if req.process.is_some() {
-            req.process.as_ref().unwrap()
+            req.process
+                .as_ref()
+                .ok_or("Unable to reference process")
+                .map_err(|e| anyhow!(e))?
```
dgibson, line 361: You can use `if let` instead here, again.

src/agent/src/rpc.rs
```diff
@@ -441,7 +461,12 @@ impl AgentService {
             Some(p) => p,
             None => {
                 // Lost race, pick up exit code from channel
-                resp.status = exit_recv.recv().await.unwrap();
+                resp.status = exit_recv
+                    .recv()
+                    .await
+                    .ok_or("Failed to receive exit code")
+                    .map_err(|e| anyhow!(e))?;
```
dgibson, line 468: Unnecessary `.ok_or().map_err()` chain again.

src/agent/src/rpc.rs
```diff
@@ -1291,7 +1328,10 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
                 info!(
                     sl!(),
                     "hotplug memory error: {}",
-                    e.as_errno().unwrap().desc()
+                    e.as_errno()
+                        .ok_or("Unable to extract errno")
+                        .map_err(|e| anyhow!(e))?
+                        .desc()
```
dgibson, line 1334: Hmm... using `as_errno()` seems bogus in the first place.  We're just trying to display the error in the log message, so why not just... display the error in the log message rather than (fallibly) converting to an errno and back again.

src/agent/src/rpc.rs
```diff
@@ -1542,18 +1585,15 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
         PathBuf::from("/")
     };
 
-    fs::create_dir_all(dir.to_str().unwrap()).or_else(|e| {
+    fs::create_dir_all(dir.clone()).or_else(|e| {
```
dgibson, line 1588: You shouldn't need to `clone()` here. No idea what the `to_str()` was about in the first place, either.  AFAICT just passing `&dir` should be fine.

src/agent/src/rpc.rs
```diff
@@ -1562,10 +1602,10 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
         .write(true)
         .create(true)
         .truncate(false)
-        .open(tmpfile.to_str().unwrap())?;
+        .open(tmpfile.clone())?;
 
     file.write_all_at(req.data.as_slice(), req.offset as u64)?;
-    let st = stat::stat(tmpfile.to_str().unwrap())?;
+    let st = stat::stat(&tmpfile.clone())?;
```
dgibson, line 1608: And here, shouldn't need the `clone()` as well.

src/agent/src/rpc.rs
```diff
@@ -1574,7 +1614,7 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
     file.set_permissions(std::fs::Permissions::from_mode(req.file_mode))?;
 
     unistd::chown(
-        tmpfile.to_str().unwrap(),
+        &tmpfile.clone(),
```
dgibson, line 1617: And again.

src/agent/src/rpc.rs
```diff
@@ -1632,15 +1672,32 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
         "",
         &sl!(),
     )?;
+
+    let rootfs_path_name = rootfs_path
+        .to_str()
+        .ok_or("failed to get rootfs path name")
```
dgibson, line 1678: Misleading message: this just means that the path wasn't valid Unicode.

src/agent/src/rpc.rs
```diff
@@ -1632,15 +1672,32 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
...
+        .map_err(|e| anyhow!(e))?
```
dgibson, line 1679: Unnecessary `.ok_or().map_err()` chain.

src/agent/src/rpc.rs
```diff
@@ -1632,15 +1672,32 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
...
+        path: rootfs_path_name,
         readonly: spec_root.readonly,
     });
 
-    let _ = spec.save(config_path.to_str().unwrap());
+    let _ = spec.save(
+        config_path
+            .to_str()
+            .ok_or("cannot get config path")
+            .map_err(|e| anyhow!(e))?,
```
dgibson, line 1691: Again, a misleading message and unnecessary method chain.

src/agent/src/rpc.rs
```diff
@@ -1632,15 +1672,32 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
...
+    );
 
     let olddir = unistd::getcwd().context("cannot getcwd")?;
-    unistd::chdir(bundle_path.to_str().unwrap())?;
+    unistd::chdir(
+        bundle_path
+            .to_str()
+            .ok_or("unable to get bundle path")
+            .map_err(|e| anyhow!(e))?,
```
dgibson, line 1699: Same again.

src/agent/src/rpc.rs
```diff
@@ -1673,8 +1730,8 @@ fn load_kernel_module(module: &protocols::agent::KernelModule) -> Result<()> {
 
     match status.code() {
         Some(code) => {
-            let std_out: String = String::from_utf8(output.stdout).unwrap();
-            let std_err: String = String::from_utf8(output.stderr).unwrap();
+            let std_out = String::from_utf8_lossy(&output.stdout).to_string();
+            let std_err = String::from_utf8_lossy(&output.stderr).to_string();
```
dgibson, line 1734: You shouldn't need the `to_string()` calls here.

dgibson:
> @dgibson - I've reworked these changes further so please can you tal?

I'm not sure what tal means, but I've done another round of review.

gkurz:
> > @dgibson - I've reworked these changes further so please can you tal?
> 
> I'm not sure what tal means, but I've done another round of review.

tal means "take a look" which you did apparently :smiley:

src/agent/src/mount.rs
```diff
@@ -139,20 +139,20 @@ pub const STORAGE_HANDLER_LIST: &[&str] = &[
```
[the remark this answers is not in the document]
jodh-intel: Fixed.

src/agent/src/mount.rs
jodh-intel: I've now resolved all those `Path` issues afaik.

jodh-intel (commented):
Thanks for reviewing again @dgibson. Branch updated.

src/agent/src/rpc.rs
```diff
@@ -345,14 +345,14 @@ impl AgentService {
         let s = self.sandbox.clone();
         let mut sandbox = s.lock().await;
 
-        let process = if req.process.is_some() {
-            req.process.as_ref().unwrap()
+        let process = if let Some(p) = req.process.into_option() {
+            p
         } else {
             return Err(anyhow!(nix::Error::from_errno(nix::errno::Errno::EINVAL)));
         };
```
dgibson, line 352:
```suggestion
let process = req
    .process
    .into_option()
    .ok_or_else(|| anyhow!(nix::Error::from_errno(nix::errono::Errno::EINVAL)))?;
```

src/agent/src/rpc.rs
```diff
@@ -807,25 +816,24 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+                };
+
+                let err = libc::ioctl(fd, TIOCSWINSZ, &win);
+                Errno::result(err).map(drop).map_err(|e| {
+                    ttrpc_error(ttrpc::Code::INTERNAL, format!("ioctl error: {:?}", e))
+                })?;
+            }
+        } else {
             return Err(ttrpc_error(ttrpc::Code::UNAVAILABLE, "no tty".to_string()));
         }
```
dgibson, line 835:
```suggestion
let term_master = p
    .term_master
    .ok_or_else(||  Err(ttrpc_error(ttrpc::Code::UNAVAILABLE, "no tty".to_string())))?;
```

followed by the existing code but with `term_master` replacing `p.term_master.unwrap()`.

src/agent/src/rpc.rs
dgibson: You still have `clone()`.  Just `fs::create_dir_all(&dir)` should do the trick.

src/agent/src/rpc.rs
```diff
@@ -1565,10 +1582,10 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
         .write(true)
         .create(true)
         .truncate(false)
-        .open(tmpfile.to_str().unwrap())?;
+        .open(tmpfile.clone())?;
```
dgibson, line 1585: Similarly `&tmpfile` should work here instead of cloning.

src/agent/src/rpc.rs
```diff
@@ -1607,19 +1607,22 @@ async fn do_add_swap(sandbox: &Arc<Mutex<Sandbox>>, req: &AddSwapRequest) -> Res
...
     let bundle_path = Path::new(CONTAINER_BASE).join(cid);
     let config_path = bundle_path.join("config.json");
     let rootfs_path = bundle_path.join("rootfs");
 
     fs::create_dir_all(&rootfs_path)?;
     baremount(
-        &spec_root.path,
-        rootfs_path.to_str().unwrap(),
+        &spec_root_path,
+        &rootfs_path,
```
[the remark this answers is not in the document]
dgibson: Oh... but they're removed in a later patch in the series.  Best to fold this hunk into the earlier patch.

src/agent/src/rpc.rs
```diff
@@ -1628,22 +1645,36 @@ fn setup_bundle(cid: &str, spec: &mut Spec) -> Result<PathBuf> {
...
+            .ok_or_else(|| anyhow!("cannot convert path to unicode"))?,
+    );
 
     let olddir = unistd::getcwd().context("cannot getcwd")?;
-    unistd::chdir(bundle_path.to_str().unwrap())?;
+    unistd::chdir(
+        bundle_path
+            .to_str()
+            .ok_or_else(|| anyhow!("cannot convert bundle path to unicode"))?,
+    )?;
```
dgibson, line 1677: A bunch more of these can probably be fixed more nicely by moving stuff to use `Path` throughout, but that's probably a challenge for another day.

src/agent/src/metrics.rs
liubin: @dgibson @jodh-intel IMO, we need to avoid returning errors, that the `update_guest_metrics` after `update_agent_metrics` will get a chance to run, that we can get parts of metrics.

Some I think dgibson's suggestion can be moved to line 77:

```rust
match update_agent_metrics() {
    Ok(_) => {},
    Err(e) => { error!(sl!(), "failed to collect agent process metrics: {:?}", err);},
}
```

How about this?

tools/agent-ctl/src/main.rs
```diff
@@ -218,7 +216,7 @@ fn real_main() -> Result<()> {
                 .long("log-level")
                 .short("l")
                 .help("specific log level")
-                .default_value(logging::slog_level_to_level_name(DEFAULT_LOG_LEVEL).unwrap())
+                .default_value(logging::slog_level_to_level_name(DEFAULT_LOG_LEVEL).map_err(|e| anyhow!(e))?)
```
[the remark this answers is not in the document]
jodh-intel: It's required because that function returns a `&'static str`. There might be a way to improve the code, but this is a trivial fix to a test tool so imho we don't need to labour this.

src/agent/src/rpc.rs
jodh-intel: Applied.

src/agent/src/rpc.rs
jodh-intel: I don't follow. Maybe you could clarify or raise a PR after this one lands to resolve? My fix works so to avoid holding this PR from landing any longer, you might want to help with the effort and raise a PR or two?

src/agent/src/rpc.rs
jodh-intel: Fixed.

src/agent/src/metrics.rs
```diff
@@ -132,10 +127,11 @@ fn update_agent_metrics() {
     }
 
     match me.status() {
-        Err(err) => {
-            info!(sl!(), "failed to get process status: {:?}", err);
+        Err(err) => Err(anyhow!("failed to get process status: {:?}", err)),
```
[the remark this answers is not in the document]
jodh-intel: @dgibson - for now I've retained the behaviour of "log on error". If @liubin doesn't get a chance to comment, let's just land it as-is and I can raise a follow-on PR if he agrees we should return errors here.

jodh-intel (commented):
Thanks for the review @dgibson. I believe I've now addressed your comments so let's try to get this landed before the code goes stale underneath this PR and it requires yet another rework.

There is the question of whether `update_agent_metrics()` should return errors or not but I've rearranged the changes on that function to retain the existing behaviour. Once @liubin get's a chance to comment, I'll raise a follow-on PR if necessary.

src/agent/src/device.rs
```diff
@@ -413,12 +416,16 @@ fn scan_scsi_bus(scsi_addr: &str) -> Result<()> {
...
-            "scan"
-        );
+
+        let host_str = host
+            .to_str()
+            .ok_or(format!(
+                "failed to convert directory entry to unicode for file {:?}",
+                host
+            ))
+            .map_err(|e| anyhow!(e))?;
```
dgibson, line 426: One more redundant `.ok_or().map_err()` chain here.

src/agent/src/device.rs
dgibson: Looks good.

src/agent/src/mount.rs
```diff
@@ -94,21 +94,21 @@ pub struct InitMount {
...
+        m.insert("cpuset".into(), "/sys/fs/cgroup/cpuset".into());
+        m.insert("memory".into(), "/sys/fs/cgroup/memory".into());
+        m.insert("devices".into(), "/sys/fs/cgroup/devices".into());
+        m.insert("freezer".into(), "/sys/fs/cgroup/freezer".into());
+        m.insert("net_cls".into(), "/sys/fs/cgroup/net_cls".into());
+        m.insert("perf_event".into(), "/sys/fs/cgroup/perf_event".into());
+        m.insert("net_prio".into(), "/sys/fs/cgroup/net_prio".into());
+        m.insert("hugetlb".into(), "/sys/fs/cgroup/hugetlb".into());
+        m.insert("pids".into(), "/sys/fs/cgroup/pids".into());
+        m.insert("rdma".into(), "/sys/fs/cgroup/rdma".into());
```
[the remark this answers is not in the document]
dgibson: Hmmm... I'm still not convinced this is necessary.  You should be able to do a `.as_str()` on the lookup, and a `.to_owned()` when you want to copy something out of the table.

src/agent/src/mount.rs
```diff
@@ -773,10 +773,9 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
         return Err(anyhow!("{:?} exists but is not a regular file", path));
     }
 
-    // The only way parent() can return None is if the path is /,
-    // which always exists, so the test above will already have caught
-    // it, thus the unwrap() is safe
-    let dir = path.parent().unwrap();
+    let dir = path
+        .parent()
+        .ok_or_else(|| anyhow!("failed to find parent path for {:?}", path))?;
```
dgibson, line 778: I still think an assert/unwrap is more appropriate here, but I don't care enough to insist.

src/agent/src/rpc.rs
dgibson: Eh, fair enough.

dgibson (commented):
I commented one remaining `,ok_or().map_err()` chain, otherwise this now LGTM.

src/agent/src/device.rs
jodh-intel: Fixed.

jodh-intel:
@dgibson - Regarding the `CGROUPS` hashmap... you're right! Fixed! :smile:

liubin (approved):

lifupan (approved):

Issue #3014, opened 2021-11-11
release: upload-libseccomp-tarball fails because of using a system version of yq


ManaSugi:
# Description of problem

`upload-libseccomp-tarball` action gets the version information of libseccomp from `versions.yaml` using `yq` command.
Currently, the action fails because a system version of yq that cannot understand `read` argument is used.
We need to explicitly call `${GOPATH}/bin/yq` that is installed by `ci/install_yq.sh`

https://github.com/kata-containers/kata-containers/runs/4171853039?check_suite_focus=true

# Expected result

We can get correct versions from `versions.yaml`.

# Actual result

```
Run pushd $GITHUB_WORKSPACE
~/work/kata-containers/kata-containers ~/work/kata-containers/kata-containers
Error: unknown command "read" for "yq"
Run 'yq --help' for usage.
Error: Process completed with exit code 1.
```

PR #3015, opened 2021-11-11
release: Use ${GOPATH}/bin/yq for upload-libseccomp-tarball action

ManaSugi:
We need to explicitly call `${GOPATH}/bin/yq` that is installed by `ci/install_yq.sh`.

Fixes: #3014

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -158,13 +158,14 @@ jobs:
       - name: download-and-upload-tarball
         env:
           GITHUB_TOKEN: ${{ secrets.GIT_UPLOAD_TOKEN }}
+          GOPATH: ${HOME}/go
         run: |
           pushd $GITHUB_WORKSPACE
           ./ci/install_yq.sh
           tag=$(echo $GITHUB_REF | cut -d/ -f3-)
           versions_yaml="versions.yaml"
-          version=$(yq read ${versions_yaml} "externals.libseccomp.version")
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -158,13 +158,14 @@ jobs:
       - name: download-and-upload-tarball
         env:
           GITHUB_TOKEN: ${{ secrets.GIT_UPLOAD_TOKEN }}
+          GOPATH: ${HOME}/go
```
bergwolf, line 161: Is this where `GO` is installed in a github action environment?

ManaSugi: No, I think Go is not installed in a github action environment but `install_yq.sh` creates the `GO` directory at the following line. Hence, we can use it.
https://github.com/kata-containers/kata-containers/blob/main/ci/install_yq.sh#L24

bergwolf (approved):

.github/workflows/release.yaml
fidencio: This will work, but this is error-prone (**not** your fault @ManaSugi).

I'm pretty sure we'll change that script at some point and the change will byte us quite hard. :-/

In any case, this is good enough for now.

fidencio (approved):
lgtm, thanks @ManaSugi!

Could you also raise a backport to the stable-2.3 branch?

PR #3018, opened 2021-11-11
stable 2.3: cgroups systemd fix

snir911:
bcf181b7 cgroups: Fix systemd cgroup support
b34ed403 cgroups: pass vhost-vsock device to cgroup
7362e1e8 runtime: remove prefix when cgroups are managed by systemd


snir911:
not main sorry

Issue #3031, opened 2021-11-12
Runtime: merge virtcontainers/pkg/types into virtcontainers/types


liubin:
There are two `types` packages under virtcontainers, and the `virtcontainers/pkg/types` has a few codes, merging them into one can make it easy for outstanding and using `types` package.

PR #3032, opened 2021-11-12
runtime: merge virtcontainers/pkg/types into virtcontainers/types

liubin:
There are two types packages under virtcontainers, and the
virtcontainers/pkg/types has a few codes, merging them into
one can make it easy for outstanding and using types package.

Fixes: #3031


Jakob-Naucke (approved):
LGTM, this looks pretty uncontroversial and I think we can depend on the tests for something like this :slightly_smiling_face: Thanks @liubin!

egernst (approved):

Issue #3033, opened 2021-11-12
Remove actions that are specific to kata-containers 1.x


fidencio:
When we the kata-containers repo was created we ended up inheriting a bunch of content from the 1.x repos, including GitHub actions.
Some of those actions were and still are used by us, while some are totally tied to kata-containers 1.x and could be simply removed.

PR #3034, opened 2021-11-12
workflows: Remove non-used main.yaml

fidencio:
The main.yaml workflow was created and used only on 1.x.  We inherited
it, but we didn't remove it after deprecating the 1.x repos.

Fixes: #3033


devimc (approved):

gkurz (approved):
It took less than a microsecond for @fidencio to update the documentation when I asked on slack ! :1st_place_medal: 

LGTM now. Thanks @fidencio !

PR #3035, opened 2021-11-12
CCv0: Rustify image download

arronwy:
This PR try to use rust crate to download container image and decrypt with ocicrypt-rs crate.

We keep current skopeo solution with use_skopeo flag, but if upstream is agree, we can delete skopeo related code.

BTW this PR also depends on https://github.com/kata-containers/kata-containers/pull/2911 which will pass the `kbc_name` and `kbs_uri` to ocicrypt decrypt config

Private registry is supported with current rustified image download implementation, but we have one limitation is don't support container image manifiest lists and we will focuse on implementing new image-rs crate to address all gaps.

Fixes: #3041 

@sameo @fitzthum @stevenhorsman

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -79,7 +79,7 @@ jobs:
     - name: Check vendored code
       if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
       run: |
-        cd ${GOPATH}/src/github.com/${{ github.repository }} && make vendor
+        cd ${GOPATH}/src/github.com/${{ github.repository }} && CARGO_HOME="/tmp/.cargo" make vendor
     - name: Static Checks
       if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
       run: |
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -61,6 +61,13 @@ vsock-exporter = { path = "vsock-exporter" }
 serde = { version = "1.0.129", features = ["derive"] }
 toml = "0.5.8"
 
+# Image pull/decrypt
+oci-distribution = "0.7.0"
+ocicrypt-rs = { git = "https://github.com/containers/ocicrypt-rs" }
+sha2 = "0.9.8"
+# "vendored" feature for openssl is required by musl build
+openssl = { version = "0.10.38", features = ["vendored"] }
+
 [dev-dependencies]
 tempfile = "3.1.0"
 
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -22,7 +22,7 @@ use crate::AGENT_CONFIG;
 
 const SKOPEO_PATH: &str = "/usr/bin/skopeo";
 const UMOCI_PATH: &str = "/usr/local/bin/umoci";
-const IMAGE_OCI: &str = "image_oci:latest";
+const IMAGE_OCI: &str = "image_oci";
 const AA_PATH: &str = "/usr/local/bin/attestation-agent";
 const AA_PORT: &str = "127.0.0.1:50000";
 const OCICRYPT_CONFIG_PATH: &str = "/tmp/ocicrypt_config.json";
@@ -58,7 +58,7 @@ impl ImageService {
 
         let tmp_cid_path = Path::new("/tmp/").join(cid);
         let oci_path = tmp_cid_path.join(IMAGE_OCI);
-        let target_path_oci = format!("oci://{}", oci_path.to_string_lossy());
+        let target_path_oci = format!("oci://{}:latest", oci_path.to_string_lossy());
 
         fs::create_dir_all(&oci_path)?;
 
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -3,9 +3,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use std::env;
 use std::fs;
-use std::io::Write;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::{Command, ExitStatus};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
@@ -13,6 +13,8 @@ use std::sync::Arc;
 use anyhow::{anyhow, ensure, Result};
 use async_trait::async_trait;
 use protocols::image;
+use std::convert::TryFrom;
+use std::fs::File;
 use tokio::sync::Mutex;
 use ttrpc::{self, error::get_rpc_status as ttrpc_error};
 
@@ -20,18 +22,46 @@ use crate::rpc::{verify_cid, CONTAINER_BASE};
 use crate::sandbox::Sandbox;
 use crate::AGENT_CONFIG;
 
+use oci_distribution::client::{ImageData, ImageLayer};
+use oci_distribution::manifest::{OciDescriptor, OciManifest};
+use oci_distribution::{manifest, secrets::RegistryAuth, Client, Reference};
+use ocicrypt_rs::config::CryptoConfig;
+use ocicrypt_rs::encryption::decrypt_layer;
+use ocicrypt_rs::helpers::create_decrypt_config;
+use ocicrypt_rs::spec::{
+    MEDIA_TYPE_LAYER_ENC, MEDIA_TYPE_LAYER_GZIP_ENC, MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_ENC,
+    MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_GZIP_ENC,
+};
+use serde::Serialize;
+use sha2::{Digest, Sha256};
+use std::collections::HashMap;
+use std::io::{Read, Write};
+
 const SKOPEO_PATH: &str = "/usr/bin/skopeo";
 const UMOCI_PATH: &str = "/usr/local/bin/umoci";
 const IMAGE_OCI: &str = "image_oci";
 const AA_PATH: &str = "/usr/local/bin/attestation-agent";
 const AA_PORT: &str = "127.0.0.1:50000";
 const OCICRYPT_CONFIG_PATH: &str = "/tmp/ocicrypt_config.json";
+const OCI_ANNOTATION_REF_NAME: &str = "org.opencontainers.image.ref.name";
+const OCI_IMAGE_MANIFEST_NAME: &str = "application/vnd.oci.image.manifest.v1+json";
+const OCI_LAYOUT: &str = r#"{"imageLayoutVersion": "1.0.0"}"#;
+const IMAGE_DOCKER_LAYER_FOREIGN_GZIP_MEDIA_TYPE: &str =
+    "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip";
+const DIGEST_SHA256: &str = "sha256";
 
 // Convenience macro to obtain the scope logger
 macro_rules! sl {
     () => {
         slog_scope::logger()
     };
+}
+
+#[derive(Serialize, Debug, Default, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct IndexDescriptor {
+    pub schema_version: u8,
+    pub manifests: Vec<OciDescriptor>,
 }
 
 pub struct ImageService {
@@ -45,6 +75,214 @@ impl ImageService {
             sandbox,
             attestation_agent_started: AtomicBool::new(false),
         }
+    }
+
+    fn build_oci_path(cid: &str) -> PathBuf {
+        let mut oci_path = PathBuf::from("/tmp");
+        oci_path.push(cid);
+        oci_path.push(IMAGE_OCI);
+        oci_path
+    }
+
+    fn decrypt_layer_data(
+        layer: &ImageLayer,
+        layer_digest: &str,
+        image_manifest: &mut OciManifest,
+        crypto_config: &CryptoConfig,
+        oci_blob_path: &Path,
+    ) -> Result<()> {
+        if let Some(decrypt_config) = &crypto_config.decrypt_config {
+            for layer_desc in image_manifest.layers.iter_mut() {
+                if layer_desc.digest.as_str() == layer_digest {
+                    let (layer_decryptor, _dec_digest) =
+                        decrypt_layer(decrypt_config, layer.data.as_slice(), layer_desc, false)?;
+                    let mut plaintxt_data: Vec<u8> = Vec::new();
+                    let mut decryptor =
+                        layer_decryptor.ok_or_else(|| anyhow!("Missing layer decryptor"))?;
+
+                    decryptor.read_to_end(&mut plaintxt_data)?;
+                    let layer_name = format!("{:x}", Sha256::digest(&plaintxt_data));
+                    let mut out_file = File::create(oci_blob_path.join(&layer_name))?;
+                    out_file.write_all(&plaintxt_data)?;
+                    layer_desc.media_type = manifest::IMAGE_LAYER_GZIP_MEDIA_TYPE.to_string();
+
+                    layer_desc.digest = format!("{}:{}", DIGEST_SHA256, layer_name);
+                }
+            }
+        } else {
+            return Err(anyhow!("No decrypt config available"));
+        }
+
+        Ok(())
+    }
+
+    fn handle_layer_data(
+        image_data: &ImageData,
+        image_manifest: &mut OciManifest,
+        crypto_config: &CryptoConfig,
+        oci_blob_path: &Path,
+    ) -> Result<()> {
+        for layer in image_data.layers.iter() {
+            let layer_digest = layer.clone().sha256_digest();
+
+            if layer.media_type == MEDIA_TYPE_LAYER_GZIP_ENC
+                || layer.media_type == MEDIA_TYPE_LAYER_ENC
+            {
+                Self::decrypt_layer_data(
+                    layer,
+                    &layer_digest,
+                    image_manifest,
+                    crypto_config,
+                    oci_blob_path,
+                )?;
+            } else if let Some(layer_name) =
+                layer_digest.strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+            {
+                let mut out_file = File::create(oci_blob_path.join(&layer_name))?;
+                out_file.write_all(&layer.data)?;
+            } else {
+                error!(
+                    sl!(),
+                    "layer digest algo not supported:: {:?}", layer_digest
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    #[tokio::main]
+    async fn download_image(
+        image: &str,
+        auth: &RegistryAuth,
+    ) -> anyhow::Result<(OciManifest, String, ImageData)> {
+        let reference = Reference::try_from(image)?;
+        let mut client = Client::default();
+        let (image_manifest, _image_digest, image_config) =
+            client.pull_manifest_and_config(&reference, auth).await?;
+
+        // TODO: Get the value from config
+        let max_attempt = 2;
+        let attempt_interval = 1;
+        for i in 1..max_attempt {
+            match client
+                .pull(
+                    &reference,
+                    auth,
+                    vec![
+                        manifest::IMAGE_DOCKER_LAYER_GZIP_MEDIA_TYPE,
+                        manifest::IMAGE_LAYER_GZIP_MEDIA_TYPE,
+                        MEDIA_TYPE_LAYER_GZIP_ENC,
+                        MEDIA_TYPE_LAYER_ENC,
+                        MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_ENC,
+                        MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_GZIP_ENC,
+                    ],
+                )
+                .await
+            {
+                Ok(data) => return Ok((image_manifest, image_config, data)),
+                Err(e) => {
+                    info!(
+                        sl!(),
+                        "Got error on pull call attempt #{}. Will retry in {}s: {:?}",
+                        attempt_interval,
+                        i,
+                        e
+                    );
+                    tokio::time::sleep(tokio::time::Duration::from_secs(attempt_interval)).await;
+                }
+            }
+        }
+
+        Err(anyhow!("Failed to download image data"))
+    }
+
+    fn pull_image_with_oci_distribution(
+        image: &str,
+        cid: &str,
+        source_creds: &Option<String>,
+        aa_kbc_params: &str,
+    ) -> Result<()> {
+        let oci_path = Self::build_oci_path(cid);
+        fs::create_dir_all(&oci_path)?;
+
+        let mut auth = RegistryAuth::Anonymous;
+        if let Some(source_creds) = source_creds {
+            if let Some((username, password)) = source_creds.split_once(':') {
+                auth = RegistryAuth::Basic(username.to_string(), password.to_string());
+            } else {
+                return Err(anyhow!("Invalid authentication info ({:?})", source_creds));
+            }
+        }
+
+        let (mut image_manifest, image_config, image_data) = Self::download_image(image, &auth)?;
+
+        // Prepare OCI layout storage for umoci
+        image_manifest.config.media_type = manifest::IMAGE_CONFIG_MEDIA_TYPE.to_string();
+        // TODO: support other digest algo like sha512
+        let oci_blob_path = oci_path.join(format!("blobs/{}", DIGEST_SHA256));
+        fs::create_dir_all(&oci_blob_path)?;
+
+        if let Some(config_name) = &image_manifest
+            .config
+            .digest
+            .strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+        {
+            let mut out_file = File::create(oci_blob_path.join(config_name))?;
+            out_file.write_all(image_config.as_bytes())?;
+        }
+
+        let mut cc = CryptoConfig::default();
+
+        if !aa_kbc_params.is_empty() {
+            let decrypt_config = format!("provider:attestation-agent:{}", aa_kbc_params);
+            cc = create_decrypt_config(vec![decrypt_config], vec![])?;
+        }
+
+        // Covert docker layer media type to OCI type
+        for layer_desc in image_manifest.layers.iter_mut() {
+            if layer_desc.media_type == manifest::IMAGE_DOCKER_LAYER_GZIP_MEDIA_TYPE
+                || layer_desc.media_type == IMAGE_DOCKER_LAYER_FOREIGN_GZIP_MEDIA_TYPE
+            {
+                layer_desc.media_type = manifest::IMAGE_LAYER_GZIP_MEDIA_TYPE.to_string();
+            }
+        }
+
+        Self::handle_layer_data(&image_data, &mut image_manifest, &cc, &oci_blob_path)?;
+
+        let manifest_json = serde_json::to_string(&image_manifest)?;
+        let manifest_digest = format!("{:x}", Sha256::digest(manifest_json.as_bytes()));
+
+        let mut out_file = File::create(oci_blob_path.join(manifest_digest))?;
+        out_file.write_all(manifest_json.as_bytes())?;
+
+        let mut annotations = HashMap::new();
+        annotations.insert(OCI_ANNOTATION_REF_NAME.to_string(), "latest".to_string());
+
+        let manifest_descriptor = OciDescriptor {
+            media_type: OCI_IMAGE_MANIFEST_NAME.to_string(),
+            digest: format!(
+                "{}:{:x}",
+                DIGEST_SHA256,
+                Sha256::digest(manifest_json.as_bytes())
+            ),
+            size: manifest_json.len() as i64,
+            annotations: Some(annotations),
+            ..Default::default()
+        };
+
+        let index_descriptor = IndexDescriptor {
+            schema_version: image_manifest.schema_version,
+            manifests: vec![manifest_descriptor],
+        };
+
+        let mut out_file = File::create(format!("{}/index.json", oci_path.to_string_lossy()))?;
+        out_file.write_all(serde_json::to_string(&index_descriptor)?.as_bytes())?;
+
+        let mut out_file = File::create(format!("{}/oci-layout", oci_path.to_string_lossy()))?;
+        out_file.write_all(OCI_LAYOUT.as_bytes())?;
+
+        Ok(())
     }
 
     fn pull_image_from_registry(
@@ -61,8 +299,6 @@ impl ImageService {
         let target_path_oci = format!("oci://{}:latest", oci_path.to_string_lossy());
 
         fs::create_dir_all(&oci_path)?;
-
-        info!(sl!(), "Attempting to pull image {}...", &source_image);
 
         let mut pull_command = Command::new(SKOPEO_PATH);
         pull_command
@@ -166,6 +402,8 @@ impl ImageService {
     }
 
     async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
+        env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
+
         let image = req.get_image();
         let mut cid = req.get_container_id();
 
@@ -196,12 +434,30 @@ impl ImageService {
 
         let source_creds = (!req.get_source_creds().is_empty()).then(|| req.get_source_creds());
 
-        // Read the policy path from the agent config
-        let config_policy_path = &AGENT_CONFIG.read().await.container_policy_path;
-        let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
-        info!(sl!(), "Using container policy_path: {:?}...", &policy_path);
+        if Path::new(SKOPEO_PATH).exists() {
+            // Read the policy path from the agent config
+            let config_policy_path = &AGENT_CONFIG.read().await.container_policy_path;
+            let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
 
-        Self::pull_image_from_registry(image, cid, &source_creds, &policy_path, aa_kbc_params)?;
+            Self::pull_image_from_registry(image, cid, &source_creds, &policy_path, aa_kbc_params)?;
+        } else {
+            let image = image.to_string();
+            let cid = cid.to_string();
+            let source_creds =
+                (!req.get_source_creds().is_empty()).then(|| req.get_source_creds().to_string());
+            let aa_kbc_params = aa_kbc_params.to_string();
+
+            // ocicrypt-rs keyprovider module will create a new runtime to talk with
+            // attestation agent, to avoid startup a runtime within a runtime, we
+            // spawn a new thread here.
+            tokio::task::spawn_blocking(move || {
+                Self::pull_image_with_oci_distribution(&image, &cid, &source_creds, &aa_kbc_params)
+                    .map_err(|err| warn!(sl!(), "pull image failed: {:?}", err))
+                    .ok();
+            })
+            .await?;
+        }
+
         Self::unpack_image(cid)?;
 
         let mut sandbox = self.sandbox.lock().await;
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -405,19 +405,20 @@ impl ImageService {
         env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
 
         let image = req.get_image();
-        let mut cid = req.get_container_id();
+        let mut cid = req.get_container_id().to_string();
 
         let aa_kbc_params = &AGENT_CONFIG.read().await.aa_kbc_params;
 
         if cid.is_empty() {
             let v: Vec<&str> = image.rsplit('/').collect();
             if !v[0].is_empty() {
-                cid = v[0]
+                // ':' have special meaning for umoci during upack
+                cid = v[0].replace(":", "_");
             } else {
                 return Err(anyhow!("Invalid image name. {}", image));
             }
         } else {
-            verify_cid(cid)?;
+            verify_cid(&cid)?;
         }
 
         if !aa_kbc_params.is_empty() {
@@ -439,7 +440,13 @@ impl ImageService {
             let config_policy_path = &AGENT_CONFIG.read().await.container_policy_path;
             let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
 
-            Self::pull_image_from_registry(image, cid, &source_creds, &policy_path, aa_kbc_params)?;
+            Self::pull_image_from_registry(
+                image,
+                &cid,
+                &source_creds,
+                &policy_path,
+                aa_kbc_params,
+            )?;
         } else {
             let image = image.to_string();
             let cid = cid.to_string();
@@ -458,7 +465,7 @@ impl ImageService {
             .await?;
         }
 
-        Self::unpack_image(cid)?;
+        Self::unpack_image(&cid)?;
 
         let mut sandbox = self.sandbox.lock().await;
         sandbox.images.insert(String::from(image), cid.to_string());
--- a/src/agent/src/uevent.rs
+++ b/src/agent/src/uevent.rs
@@ -240,7 +240,6 @@ pub(crate) fn spawn_test_watcher(sandbox: Arc<Mutex<Sandbox>>, uev: Uevent) {
                     if matcher.is_match(&uev) {
                         let (_, sender) = watch.take().unwrap();
                         let _ = sender.send(uev.clone());
-                        return;
                     }
                 }
             });
--- a/versions.yaml
+++ b/versions.yaml
@@ -281,7 +281,7 @@ languages:
       description: |
         'newest-version' is the latest version known to work when
         building Kata
-      newest-version: "1.54.0"
+      newest-version: "1.56.0"
 
 specs:
   description: "Details of important specifications"
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -24,6 +24,42 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
+version = "0.7.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
+dependencies = [
+ "cfg-if 1.0.0",
+ "cipher",
+ "cpufeatures",
+ "ctr",
+ "opaque-debug",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df5f85a83a7d8b0442b6aa7b504b8212c1733da07b98aae43d4bc21b2cb3cdf6"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -65,6 +101,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "171374e7e3b2504e0e5236e3b59260560f9fe94bfe9ac39ba5e4e929c5590625"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "648ed8c8d2ce5409ccd57453d9d1b214b342a0d69376a6feda1fd6cae3299308"
+dependencies = [
+ "proc-macro2 1.0.26",
+ "quote 1.0.9",
+ "syn 1.0.72",
+]
+
+[[package]]
 name = "async-trait"
 version = "0.1.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -97,6 +154,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b25d992356d2eb0ed82172f5248873db5560c4721f564b13cb5193bda5e668e"
+dependencies = [
+ "byteorder",
+]
+
+[[package]]
+name = "base64"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
+
+[[package]]
+name = "base64-serde"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e964e3e0a930303c7c0bdb28ebf691dd98d9eee4b8b68019d2c995710b58a18"
+dependencies = [
+ "base64 0.13.0",
+ "serde",
+]
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -110,6 +192,15 @@ name = "bitflags"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
+
+[[package]]
+name = "block-buffer"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4152116fd6e9dadb291ae18fc1ec3575ed6d84c29642d97890f4b4a3417297e4"
+dependencies = [
+ "generic-array",
+]
 
 [[package]]
 name = "bumpalo"
@@ -208,8 +299,42 @@ dependencies = [
  "libc",
  "num-integer",
  "num-traits",
- "time",
+ "time 0.1.43",
  "winapi",
+]
+
+[[package]]
+name = "cipher"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7ee52072ec15386f770805afd189a01c8841be8696bed250fa2f13c4c0d6dfb7"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
+name = "core-foundation"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6888e10551bb93e424d8df1d07f1a8b4fceb0001a3a4b048bfc47554946f47b3"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation-sys"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
+
+[[package]]
+name = "cpufeatures"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "95059428f66df56b63431fdb4e1947ed2190586af5c5a8a8b71122bdf5a7f469"
+dependencies = [
+ "libc",
 ]
 
 [[package]]
@@ -243,6 +368,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "crypto-mac"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1d1a86f49236c215f271d40892d5fc950490551400b02ef360692c29815c714"
+dependencies = [
+ "generic-array",
+ "subtle",
+]
+
+[[package]]
+name = "ctr"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "049bb91fb4aaf0e3c7efa6cd5ef877dbbbd15b39dad06d9948de4ec8a75761ea"
+dependencies = [
+ "cipher",
+]
+
+[[package]]
 name = "derive-new"
 version = "0.5.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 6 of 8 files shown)


src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+            oci_blob_path,
+            Sha256::digest(manifest_json.as_bytes())
+        );
+
+        let mut out_file = File::create(&image_manifest_file)?;
+        out_file.write_all(manifest_json.as_bytes())?;
+
+        let mut annotations = HashMap::new();
+        annotations.insert(
+            "org.opencontainers.image.ref.name".to_string(),
```
jodh-intel, line 224: Please make this a `const` with an explanatory comment.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+            "latest".to_string(),
+        );
+
+        let manifest_descriptor = OciDescriptor {
+            media_type: "application/vnd.oci.image.manifest.v1+json".to_string(),
```
jodh-intel, line 229: Please make the media type a `const` with an explanatory comment.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+        let index_descriptor = IndexDescriptor {
+            schema_version: image_manifest.schema_version,
+            manifests: vec![manifest_descriptor],
+        };
+
+        let mut out_file = File::create(format!("{}/index.json", oci_path.to_string_lossy()))?;
+        out_file.write_all(serde_json::to_string(&index_descriptor).unwrap().as_bytes())?;
+
+        let mut out_file = File::create(format!("{}/oci-layout", oci_path.to_string_lossy()))?;
+        let oci_layout = r#"{"imageLayoutVersion": "1.0.0"}"#;
```
jodh-intel, line 245: This could also be a `const`.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+            ..Default::default()
+        };
+
+        let index_descriptor = IndexDescriptor {
+            schema_version: image_manifest.schema_version,
+            manifests: vec![manifest_descriptor],
+        };
+
+        let mut out_file = File::create(format!("{}/index.json", oci_path.to_string_lossy()))?;
+        out_file.write_all(serde_json::to_string(&index_descriptor).unwrap().as_bytes())?;
```
jodh-intel, line 242: Please don't `unwrap()` but instead pass the error back to the caller.

Jakob-Naucke:
@arronwy you appear to keep pushing the same commits (?) At least one of your commits needs to link to an issue, e.g. `Fixes: #IssueNumberHere`.

src/agent/protocols/protos/image.proto
```diff
@@ -22,6 +22,8 @@ message PullImageRequest {
     string container_id = 2;
     // Use USERNAME[:PASSWORD] for accessing the registry
     string source_creds = 3;
+    // Use skopeo or rust crate to download container image
+    bool use_skopeo = 4;
```
jodh-intel, line 26: I wonder if we should invert this and make it something like:

```suggestion
    // If set use an external helper application (such as skopeo) to download container image, otherwise use the built-in capability.
    bool use_external_download_helper = 4;
```

jodh-intel (commented):
Thanks for raising @arronwy !

First pass comments.

It would be great if you could add a few unit tests too, particularly to test a few failure scenarios:

- https://github.com/kata-containers/tests/blob/main/Unit-Test-Advice.md#rust-table-driven-tests

Jakob-Naucke:
Note to (probably) self: If we merge this, the code from #2908 should be updated maybe; Skopeo is pretty big and has tighter restrictions on Golang versions and libs.

arronwy:
> @arronwy you appear to keep pushing the same commits (?) At least one of your commits needs to link to an issue, e.g. `Fixes: #IssueNumberHere`.
> 
> e: maybe that was a GitHub glitch, I saw the "added 3 commits" many times earlier, but always with the same hashes.

Thanks for the reminder, added now :-)

src/agent/Cargo.toml
```diff
@@ -61,6 +61,12 @@ vsock-exporter = { path = "vsock-exporter" }
 serde = { version = "1.0.129", features = ["derive"] }
 toml = "0.5.8"
 
+# Image pull/decrypt
+oci-distribution = "0.7"
+ocicrypt-rs = { git = "https://github.com/containers/ocicrypt-rs" }
+sha2 = ">=0.9"
+openssl = { version = "0.10", features = ["vendored"] }
```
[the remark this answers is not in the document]
arronwy: ok, it is required by musl build, I'll add this comment.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+                    ],
+                )
+                .await
+            {
+                Ok(data) => {
+                    image_data = data;
+                    break;
+                }
+                Err(e) => {
+                    println!(
```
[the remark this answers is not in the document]
arronwy: thanks, I'll fix it.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+            "Attempting to pull image with rust crate {}...", image
+        );
+
+        let mut auth = RegistryAuth::Anonymous;
+        if let Some(source_creds) = source_creds {
+            let auth_info: Vec<&str> = source_creds.split(':').collect();
+            if auth_info.len() == 2 {
+                auth = RegistryAuth::Basic(auth_info[0].to_string(), auth_info[1].to_string());
+            } else {
+                warn!(sl!(), "invalid source_creds format \n");
```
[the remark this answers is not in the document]
arronwy: ok, I'll fix it.

src/agent/src/image_rpc.rs
arronwy: yes, it should be const, which not defined  by current oci-distribution crate, will fix it.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+        let (image_manifest, _image_digest, image_config) =
+            client.pull_manifest_and_config(&reference, auth).await?;
+
+        let mut last_error = None;
+        let mut image_data = ImageData {
+            layers: Vec::with_capacity(0),
+            digest: None,
+        };
+
+        for i in 1..2 {
```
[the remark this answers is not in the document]
arronwy: Yes, I'll use a variable, in the long-term，a config option is the right way, current we can use a default value.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
         oci_path
     }
 
+    #[tokio::main(flavor = "current_thread")]
```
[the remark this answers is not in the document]
arronwy: Yes, we don't need it anymore, we can remove it now.

src/agent/protocols/protos/image.proto
arronwy: ok, I'll use with this way.

src/agent/src/image_rpc.rs
arronwy: thanks, fixed now.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+                Ok(data) => {
+                    image_data = data;
+                    break;
+                }
+                Err(e) => {
+                    println!(
+                        "Got error on pull call attempt {}. Will retry in 1s: {:?}",
+                        i, e
+                    );
+                    last_error.replace(e);
```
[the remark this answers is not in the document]
arronwy: Yes, we don't need this, I'll remove it.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+                    tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
```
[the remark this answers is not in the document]
arronwy: Fixed now.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+        image: &str,
+        cid: &str,
+        source_creds: &Option<String>,
+    ) -> Result<()> {
+        let oci_path = Self::build_oci_path(cid);
+        fs::create_dir_all(&oci_path)?;
+
+        info!(
+            sl!(),
+            "Attempting to pull image with rust crate {}...", image
```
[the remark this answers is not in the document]
arronwy: Updated now.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+        // aa_kbc_params will get from PR: https://github.com/kata-containers/kata-containers/pull/2911
+        let aa_kbc_params = "kbc:ip:port".to_string();
+        if aa_kbc_params != "" {
+            let decrypt_config = format!("provider:attestation-agent:{}", aa_kbc_params);
+            cc = create_decrypt_config(vec![decrypt_config], vec![])?;
+        }
+
+        // Covert docker layer media type to OCI type
+        for layer_desc in image_manifest.layers.iter_mut() {
+            if layer_desc.media_type == manifest::IMAGE_DOCKER_LAYER_GZIP_MEDIA_TYPE {
```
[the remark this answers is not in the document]
arronwy: https://docs.docker.com/registry/spec/manifest-v2-2/

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+        }
+
+        let (mut image_manifest, image_config, image_data) = Self::download_image(image, &auth)?;
+
+        // Prepare OCI layout storage for umoci
+        image_manifest.config.media_type = manifest::IMAGE_CONFIG_MEDIA_TYPE.to_string();
+        let oci_blob_path = format!("{}/blobs/sha256/", oci_path.to_string_lossy());
+        fs::create_dir_all(Path::new(&oci_blob_path))?;
+
+        if let Some(config_name) = &image_manifest.config.digest.strip_prefix("sha256:") {
```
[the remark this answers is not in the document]
arronwy: yes, current we support sha256 digest algo which is the most common type, I'll add comment on that

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+            } else {
+                warn!(sl!(), "invalid source_creds format \n");
+            }
+        }
+
+        let (mut image_manifest, image_config, image_data) = Self::download_image(image, &auth)?;
+
+        // Prepare OCI layout storage for umoci
+        image_manifest.config.media_type = manifest::IMAGE_CONFIG_MEDIA_TYPE.to_string();
+        let oci_blob_path = format!("{}/blobs/sha256/", oci_path.to_string_lossy());
```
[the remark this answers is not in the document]
arronwy: currently, we only support sha256. For to_string_lossy(), since oci_blob_patch is formated from oci_path which is a PathBuf struct. we may need keep oci_blob_patch as str since it will be used to format the filename to store file.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +70,184 @@ impl ImageService {
...
+                    last_error.replace(e);
+                    tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+                }
+            }
+        }
+
+        Ok((image_manifest, image_config, image_data))
+    }
+
+    fn pull_image_with_oci_distribution(
```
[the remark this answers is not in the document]
arronwy: Yes, thanks, I'll add handler_layer_data function to simplify this function.

stevenhorsman:
@arronwy - thanks for sharing the code. I have a few high level questions:
- I thought the plan was to implement the rust image management support in the [image-rs](https://github.com/confidential-containers/image-rs) crate based on a proper design and set of tests and then switch over to it post CNCF proposal demo? If we aren't removing skopeo (which is definitely bigger than we'd like) then I'm not sure of the benefit of adding both implementations in right now?
- From my initial skim it doesn't look like this implementation supports verification of signed containers, is that correct?
- It look like you are changing the default pull method to the new code and skopeo support needs enabling through setting `use_external_download_helper=true` on the agent request. I can see this should be straight forward using `agent-ctl`, but do you have suggestions of how we can do this for the e2e scenario we are targeting the demo at, or at least `ctr shim pull-image` which we currently have working?
Thanks!

arronwy:
> @arronwy - thanks for sharing the code. I have a few high level questions:
> 
> * I thought the plan was to implement the rust image management support in the [image-rs](https://github.com/confidential-containers/image-rs) crate based on a proper design and set of tests and then switch over to it post CNCF proposal demo? If we aren't removing skopeo (which is definitely bigger than we'd like) then I'm not sure of the benefit of adding both implementations in right now?

`image-rs` will not be ready in short time. This PR works for current CCv0 requirement, since we can do it with incremental and find the potential GAP as early as possible.

> * From my initial skim it doesn't look like this implementation supports verification of signed containers, is that correct?

Yes, image signing is not supported, we also need add this feature in ocicrypt and attestation agent.

> * It look like you are changing the default pull method to the new code and skopeo support needs enabling through setting `use_external_download_helper=true` on the agent request. I can see this should be straight forward using `agent-ctl`, but do you have suggestions of how we can do this for the e2e scenario we are targeting the demo at, or at least `ctr shim pull-image` which we currently have working?

Good point, current we tried with your ccv0.sh script which works well, if we don't want change the pull method API, we can get `use_external_download_helper` from kata config file.

arronwy:
Swithed to Kata Agent Config interface for `use_external_download_helper` flag.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +76,197 @@ impl ImageService {
...
+
+                        layer_desc.digest = format!("sha256:{}", layer_name);
+                    }
+                }
+            } else if let Some(layer_name) = layer_digest.strip_prefix("sha256:") {
+                let mut out_file = File::create(format!("{}/{}", oci_blob_path, layer_name))?;
+                info!(sl!(), "Saving layer file {}...", layer_name);
+                out_file.write_all(&layer.data)?;
+            }
+        }
```
sameo, line 122: So what happens for layers that are not `GZIP_ENC` or `ENC` and for which the digest is not a SHA256? I think we should error out in that case because we don't support it.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +76,197 @@ impl ImageService {
...
+        crypto_config: &CryptoConfig,
+        oci_blob_path: &str,
+    ) -> Result<()> {
+        for layer in image_data.layers.iter() {
+            let layer_digest = layer.clone().sha256_digest();
+
+            if layer.media_type == MEDIA_TYPE_LAYEE_GZIP_ENC
+                || layer.media_type == MEDIA_TYPE_LAYER_ENC
+            {
+                if crypto_config.decrypt_config.is_none() {
```
sameo, line 91: Can we use the

```rust
if let Some(ref decrypt_config: &{unknown}) = crypto_config.decrypt_config {
}
```

pattern instead, and avoid unwrapping later?

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +76,197 @@ impl ImageService {
...
+                .await
+            {
+                Ok(data) => {
+                    image_data = data;
+                    break;
+                }
+                Err(e) => {
+                    info!(
+                        sl!(),
+                        "Got error on pull call attempt {}. Will retry in 1s: {:?}", i, e
```
sameo, line 167: Nit: this should be:

```rust
"Got error on pull call attempt #{}. Will retry in {}s: {:?}", attempt_interval, i, e
```

src/agent/src/image_rpc.rs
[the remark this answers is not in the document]
sameo: Then I think we want `oci_blob_path` to be a `PathBuf` as well, built on top of `oci_path`.

src/agent/src/config.rs
```diff
@@ -91,6 +93,7 @@ pub struct AgentConfigBuilder {
     pub unified_cgroup_hierarchy: Option<bool>,
     pub tracing: Option<bool>,
     pub endpoints: Option<EndpointsConfig>,
+    pub use_external_download_helper: Option<bool>,
```
sameo, line 96: I'd prefer if we don't have any setting at all and decide if we're going to use skopeo or not by checking if `SKOPEO_PATH` is installed or not. If it is installed we use it to download images. If it's not we're falling back to the rust implementation.

src/agent/src/image_rpc.rs
```diff
@@ -3,23 +3,48 @@
...
+use crate::AGENT_CONFIG;
+
+use oci_distribution::client::ImageData;
+use oci_distribution::manifest::{OciDescriptor, OciManifest};
+use oci_distribution::{manifest, secrets::RegistryAuth, Client, Reference};
+use ocicrypt_rs::config::CryptoConfig;
+use ocicrypt_rs::encryption::decrypt_layer;
+use ocicrypt_rs::helpers::create_decrypt_config;
+use ocicrypt_rs::spec::{
+    MEDIA_TYPE_LAYEE_GZIP_ENC, MEDIA_TYPE_LAYER_ENC, MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_ENC,
```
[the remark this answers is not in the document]
arronwy: Fixed now.

src/agent/src/config.rs
arronwy: Yes, it will be much simpler for test and runtime selection, I switch to this one

src/agent/src/image_rpc.rs
arronwy: Thanks, I added the error info for non supported digest algo now.

src/agent/src/image_rpc.rs
arronwy: Yes, thanks, fixed now.

src/agent/src/image_rpc.rs
arronwy: Yes, fixed now

src/agent/src/image_rpc.rs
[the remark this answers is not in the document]
arronwy: And switched to use DIGEST_SHA256 now.

src/agent/src/image_rpc.rs
```diff
@@ -44,6 +76,208 @@ impl ImageService {
...
+            }
+        }
+
+        Ok((image_manifest, image_config, image_data))
+    }
+
+    fn pull_image_with_oci_distribution(
+        image: &str,
+        cid: &str,
+        source_creds: &Option<String>,
```
sameo, line 189: `source_creds` could be a `&Option<&str>` to map the `pull_image_from_registry` signature.

src/agent/src/image_rpc.rs
```diff
@@ -138,7 +377,21 @@ impl ImageService {
...
 
-        Self::pull_image_from_registry(image, cid, &source_creds)?;
+        if Path::new(SKOPEO_PATH).exists() {
+            Self::pull_image_from_registry(image, cid, &source_creds)?;
+        } else {
+            let image = image.to_string();
+            let cid = cid.to_string();
+            let source_creds =
+                (!req.get_source_creds().is_empty()).then(|| req.get_source_creds().to_string());
+            tokio::task::spawn_blocking(move || {
```
sameo, line 387: Any reason why we're spawning a thread for that? Why not calling `Self::pull_image_with_oci_distribution` directly? This way you can also reuse image, cid and source_creds directly.

src/agent/src/image_rpc.rs
arronwy: `source_creds` will be moved to `pull_image_with_oci_distribution`

src/agent/src/image_rpc.rs
arronwy: The reason we spawn a new thread is current `pull_image` is a async function, in `pull_image_with_oci_distribution`, we talk with attestation agent and download image will both do in async way, this will have "thread 'main' panicked at 'Cannot start a runtime from within a runtime." error.

src/agent/src/image_rpc.rs
sameo: Yes, because we have to use a separate thread for it...

src/agent/src/image_rpc.rs
sameo: That's quite a shame. Could you please add a comment on that? It was not obvious to me at all.

arronwy: My bad, I should add the comment for that, added now :-)

sameo (approved):

Jakob-Naucke:
@arronwy I get `unsupported media type: application/vnd.docker.distribution.manifest.list.v2+json"` with a normal `docker.io/library/nginx:latest`. Is that expected?

arronwy:
> @arronwy I get `unsupported media type: application/vnd.docker.distribution.manifest.list.v2+json"` with a normal `docker.io/library/nginx:latest`. Is that expected?

Yes, manifest lists feature can be used to support different platfrom, oci-distribution v0.7 don't support yet, they latest commits seems fix this issue but not release yet, currently we can use address with arch info like: 'docker.io/amd64/nginx:latest'

stevenhorsman:
Hi folks, sorry if I'm late to the party and this has already been discussed (I can see Jakob alluded to it earlier): 
Now that the code is not using configuration to switch whether to use `skopeo` or not, but based on whether the binary is available or not I was wondering if we should try and make it easier for developers to switch between these scenarios?
I was wondering if as part of this PR we want to update the code in [`rootfs.sh`](https://github.com/kata-containers/kata-containers/blob/27c0dc260c1864beaf52ddc28f2fa0c5b5babfb2/tools/osbuilder/rootfs-builder/rootfs.sh#L637-L654) that  decides whether to add `skopeo` and `umoci` to the image so that each can be updated independently (as we need umoci in both 'rustified' and 'skopeo' pull. If it is helpful I can have a crack at created this commit before getting Jakob to check it? With this in we can just add a env_var of something to `ccv0.sh` to either add `skopeo` or not for easy testing/switching.

Alternatively if I've missed something and there is a currently an easy way to switch between both scenarios (I just modified the `SKOPEO_PATH` value in `image_rpc.rs`, but appreciate that isn't super user friendly) that I'm missing and I can help doc then let me know that as well.

**TL;DR:** is there anything I can do to help with this PR?

Jakob-Naucke:
@arronwy with Skopeo and the current state of this PR, I get `⨯ tag is not found: latest/image_oci` from umoci (have to print its stderr to see that)
without Skopeo and `docker.io/amd64/nginx:latest`, I get `⨯ open CAS: validate: read oci-layout: invalid image detected` from umoci
without Skopeo, #2911 applied and an encrypted image, I get `Unable to read ocicrypt config file: No such file or directory (os error 2)`, doesn't that env have to be set from here? We set that env in #2911, but only for Skopeo.

Can you reproduce any of these?

@stevenhorsman I have not yet looked into splitting Skopeo and umoci installation up if you want to have a shot.

stevenhorsman:
> @arronwy with Skopeo and the current state of this PR, I get `⨯ tag is not found: latest/image_oci` from umoci (have to print its stderr to see that) 

IIRC this happens if there source path to umoci is missing the tag. That's why we had `const IMAGE_OCI: &str = "image_oci:latest"` to include the tag, but now that has been updated to remove the `:latest` the unpack of umoci seems to be using "image_oci", not the tagged version?

arronwy:
> @arronwy with Skopeo and the current state of this PR, I get `⨯ tag is not found: latest/image_oci` from umoci (have to print its stderr to see that) without Skopeo and `docker.io/amd64/nginx:latest`, I get `⨯ open CAS: validate: read oci-layout: invalid image detected` from umoci without Skopeo, #2911 applied and an encrypted image, I get `Unable to read ocicrypt config file: No such file or directory (os error 2)`, doesn't that env have to be set from here? We set that env in #2911, but only for Skopeo.
> 
> Can you reproduce any of these?
> 
> @stevenhorsman I have not yet looked into splitting Skopeo and umoci installation up if you want to have a shot.

The root casue for this issue is in PullImage, we will pass cid as the storage path:
let mut cid = req.get_container_id();
When upper layer don't pass cid, it will be random like: `25f087bad77b4d3d89c0e53d44915768d57c1937618615feadab2209941b41c8`
if we pass cid="", cid will be the image name with tag like: `nginx:latest`
but `:` have special meaning for `umoci` during unpack the image, we use a workaround to parse `:` to `_`
```
let mut cid = req.get_container_id().to_string();
cid = v[0].replace(":", "_");
```

arronwy:
Current PullImage API works for:
```
PullImage image=docker.io/library/helloworld:latest cid=0123456789
PullImage image=docker.io/library/helloworld:latest  
```
but failed for : 
```
PullImage image=docker.io/library/helloworld:latest  cid=""
```
One fix can be:
```
diff --git a/src/agent/src/image_rpc.rs b/src/agent/src/image_rpc.rs
index d7c1cdc8..2151958b 100644
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -366,17 +366,17 @@ impl ImageService {
         env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", KEYPROVIDER_PATH);

         let image = req.get_image();
-        let mut cid = req.get_container_id();
+        let mut cid = req.get_container_id().to_string();

         if cid.is_empty() {
             let v: Vec<&str> = image.rsplit('/').collect();
             if !v[0].is_empty() {
-                cid = v[0]
+                cid = v[0].replace(":", "_");
             } else {
                 return Err(anyhow!("Invalid image name. {}", image));
             }
         } else {
-            verify_cid(cid)?;
+            verify_cid(&cid)?;
         }

         let source_creds = (!req.get_source_creds().is_empty()).then(|| req.get_source_creds());
@@ -387,7 +387,7 @@ impl ImageService {
             let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
             info!(sl!(), "Using container policy_path: {:?}...", &policy_path);

-            Self::pull_image_from_registry(image, cid, &source_creds, &policy_path)?;
+            Self::pull_image_from_registry(image, &cid, &source_creds, &policy_path)?;
         } else {
             let image = image.to_string();
             let cid = cid.to_string();
@@ -404,7 +404,7 @@ impl ImageService {
             .await?;
         }

-        Self::unpack_image(cid)?;
+        Self::unpack_image(&cid)?;

         let mut sandbox = self.sandbox.lock().await;
         sandbox.images.insert(String::from(image), cid.to_string());
```

arronwy:
> @arronwy with Skopeo and the current state of this PR, I get `⨯ tag is not found: latest/image_oci` from umoci (have to print its stderr to see that) without Skopeo and `docker.io/amd64/nginx:latest`, I get `⨯ open CAS: validate: read oci-layout: invalid image detected` from umoci without Skopeo, #2911 applied and an encrypted image, I get `Unable to read ocicrypt config file: No such file or directory (os error 2)`, doesn't that env have to be set from here? We set that env in #2911, but only for Skopeo.

For ocicrypt config file, this PR use env var:
```
const KEYPROVIDER_PATH: &str = "/etc/ocicrypt_keyprovider.json";
env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", KEYPROVIDER_PATH);
```
I not sure this config path is same with the config in your rootfs

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+        Ok((image_manifest, image_config, image_data))
+    }
+
+    fn pull_image_with_oci_distribution(
+        image: &str,
+        cid: &str,
+        source_creds: &Option<String>,
+    ) -> Result<()> {
+        let oci_path = Self::build_oci_path(cid);
+        fs::create_dir_all(&oci_path)?;
```
wllenyj, line 193: little tip, `tokio::fs` can be used.
Although my previous code didn't exactly use async either :)

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+        // TODO: support other digest algo like sha512
+        let oci_blob_path = oci_path.join(format!("blobs/{}", DIGEST_SHA256));
+        fs::create_dir_all(&oci_blob_path)?;
+
+        if let Some(config_name) = &image_manifest
+            .config
+            .digest
+            .strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+        {
+            let mut out_file = File::create(oci_blob_path.join(config_name))?;
```
wllenyj, line 223: `tokio::fs::File` can be used.

Jakob-Naucke:
Thank you @arronwy for your hints -- I had missed the `/etc/ocicrypt_keyprovider.json`. Do you have a use case for setting this file with custom contents or do you think we can move forward with what #2911 does [here](https://github.com/kata-containers/kata-containers/pull/2911/files#diff-a14bc6af50efd1f06bfe256b3d5d9a0b3be97955c120b9994e7a02b0996d2ff3R143-R158)?
With your replace patch, I was able to get this working with and without Skopeo :+1: Any reason you haven't applied it yet?

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+                    sl!(),
+                    "layer digest algo not supported:: {:?}", layer_digest
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    #[tokio::main]
```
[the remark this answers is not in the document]
arronwy: We spawn a new task and async function need add this macro.

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+        if let Some(source_creds) = source_creds {
+            let auth_info: Vec<&str> = source_creds.split(':').collect();
+            if auth_info.len() == 2 {
+                auth = RegistryAuth::Basic(auth_info[0].to_string(), auth_info[1].to_string());
+            } else {
+                warn!(sl!(), "invalid auth_info format");
+            }
+        }
+
+        let (mut image_manifest, image_config, image_data) = Self::download_image(image, &auth)?;
```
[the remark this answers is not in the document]
arronwy: Yes, we can do it, but the reason we spawn a task is for handle_layer_data which for decrypt the layer data, if we spawn there we may need move more data.

src/agent/src/image_rpc.rs
arronwy: Yes, we shoud use this one, thanks.

arronwy:
> Thank you @arronwy for your hints -- I had missed the `/etc/ocicrypt_keyprovider.json`. Do you have a use case for setting this file with custom contents or do you think we can move forward with what #2911 does [here](https://github.com/kata-containers/kata-containers/pull/2911/files#diff-a14bc6af50efd1f06bfe256b3d5d9a0b3be97955c120b9994e7a02b0996d2ff3R143-R158)? 

Yes, in fact we also local apply that PR to pass kbc_params
> With your replace patch, I was able to get this working with and without Skopeo 👍 Any reason you haven't applied it yet?

Thanks, I just added that patch

.github/workflows/snap-release.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
 
       - name: Build snap
         run: |
-          sudo apt-get install -y git git-extras
+          sudo apt-get install -y git git-extras pkg-config libssl-dev
```
jodh-intel, line 22: Nit: Packages not in alphebetical order ;) Same comment for below.

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+            "Attempting to pull image with native implementation {}...", image
+        );
+
+        let mut auth = RegistryAuth::Anonymous;
+        if let Some(source_creds) = source_creds {
+            let auth_info: Vec<&str> = source_creds.split(':').collect();
+            if auth_info.len() == 2 {
+                auth = RegistryAuth::Basic(auth_info[0].to_string(), auth_info[1].to_string());
+            } else {
+                warn!(sl!(), "invalid auth_info format");
```
jodh-intel, line 206: Should this return `Err()` instead? If not, can you change the message to make the behaviour clearer? Something like:

```suggestion
                warn!(sl!(), "ignoring invalid authentication info ({:?})", source_creds);
```

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+        fs::create_dir_all(&oci_path)?;
+
+        info!(
+            sl!(),
+            "Attempting to pull image with native implementation {}...", image
+        );
+
+        let mut auth = RegistryAuth::Anonymous;
+        if let Some(source_creds) = source_creds {
+            let auth_info: Vec<&str> = source_creds.split(':').collect();
```
jodh-intel, line 202: It's unclear from `image.proto` what format the password is in (a more detailed comment would help), but if it could contain a colon we should use `.split_once()`.

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+            } else {
+                warn!(sl!(), "invalid auth_info format");
+            }
+        }
+
+        let (mut image_manifest, image_config, image_data) = Self::download_image(image, &auth)?;
+
+        // Prepare OCI layout storage for umoci
+        image_manifest.config.media_type = manifest::IMAGE_CONFIG_MEDIA_TYPE.to_string();
+        // TODO: support other digest algo like sha512
```
jodh-intel, line 214: Please raise a GitHub issue and add the URL here so there is more visibility for this work item.

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+            .digest
+            .strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+        {
+            let mut out_file = File::create(oci_blob_path.join(config_name))?;
+            out_file.write_all(image_config.as_bytes())?;
+        }
+
+        let mut cc = CryptoConfig::default();
+
+        // aa_kbc_params will get from PR: https://github.com/kata-containers/kata-containers/pull/2911
```
jodh-intel, line 229: Do we need this comment now that PR has landed?

src/agent/src/image_rpc.rs
```diff
@@ -119,29 +363,48 @@ impl ImageService {
...
+
         let image = req.get_image();
-        let mut cid = req.get_container_id();
+        let mut cid = req.get_container_id().to_string();
 
         if cid.is_empty() {
             let v: Vec<&str> = image.rsplit('/').collect();
             if !v[0].is_empty() {
-                cid = v[0]
+                cid = v[0].replace(":", "_");
```
jodh-intel, line 374: I don't understand why this is needed.

src/agent/src/image_rpc.rs
```diff
@@ -119,29 +363,48 @@ impl ImageService {
...
 
-        // Read the policy path from the agent config
-        let config_policy_path = &AGENT_CONFIG.read().await.container_policy_path;
-        let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
-        info!(sl!(), "Using container policy_path: {:?}...", &policy_path);
+        if Path::new(SKOPEO_PATH).exists() {
+            // Read the policy path from the agent config
+            let config_policy_path = &AGENT_CONFIG.read().await.container_policy_path;
+            let policy_path = (!config_policy_path.is_empty()).then(|| config_policy_path);
+            info!(sl!(), "Using container policy_path: {:?}...", &policy_path);
```
jodh-intel, line 388: I'd remove the ellipsis as it suggests something else will follow (on that log line)... but it doesn't ;)

src/agent/src/image_rpc.rs
```diff
@@ -119,29 +363,48 @@ impl ImageService {
...
+            let image = image.to_string();
+            let cid = cid.to_string();
+            let source_creds =
+                (!req.get_source_creds().is_empty()).then(|| req.get_source_creds().to_string());
+            // ocicrypt-rs keyprovider module will create a new runtime to talk with
+            // attestation agent, to avoid startup a runtime within a runtime, we
+            // spawn a new thread here.
+            tokio::task::spawn_blocking(move || {
+                Self::pull_image_with_oci_distribution(&image, &cid, &source_creds)
+                    .map_err(|err| println!("{:?}", err))
```
jodh-intel, line 401: Why the call to `println!()` and not `info!()` or similar?

versions.yaml
```diff
@@ -281,7 +281,7 @@ languages:
       description: |
         'newest-version' is the latest version known to work when
         building Kata
-      newest-version: "1.54.0"
+      newest-version: "1.56.1"
```
jodh-intel, line 284: OOI, why are we updating rust here? I see the " Update to Rust 2021 edition" commit but it doesn't explain *why*.

We've had some problems changing that version recently, but that was going backwards ;) For refernce though: https://github.com/kata-containers/tests/pull/4192

/cc @fidencio

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+        for layer in image_data.layers.iter() {
+            let layer_digest = layer.clone().sha256_digest();
+
+            if layer.media_type == MEDIA_TYPE_LAYER_GZIP_ENC
+                || layer.media_type == MEDIA_TYPE_LAYER_ENC
+            {
+                if let Some(decrypt_config) = &crypto_config.decrypt_config {
+                    for layer_desc in image_manifest.layers.iter_mut() {
+                        if layer_desc.digest == layer_digest {
+                            let (layer_decryptor, _dec_digest) = decrypt_layer(
```
jodh-intel, line 95: Six levels of indentation is too much imho. Could you create a `handle_layer()` function or similar to reduce this a bit?

Please also consider adding some tests for this file as it has zero atm :sob:

- https://github.com/kata-containers/kata-containers/blob/main/docs/code-pr-advice.md#unit-tests

src/agent/src/image_rpc.rs
```diff
@@ -33,11 +58,227 @@ pub struct ImageService {
...
+                layer_digest.strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+            {
+                let mut out_file = File::create(oci_blob_path.join(&layer_name))?;
+                info!(sl!(), "Saving layer file {}...", layer_name);
+                out_file.write_all(&layer.data)?;
+            } else {
+                error!(
+                    sl!(),
+                    "layer digest algo not supported:: {:?}", layer_digest
+                );
```
jodh-intel, line 126: I haven't checked, but this formatting looks odd: it's too spaced out. Are you using `rustfmt`?

jodh-intel (commented):
Thanks @arronwy. A few comments.

src/agent/src/image_rpc.rs
jodh-intel: ftr, this is the sort of code that really suits some table-based unit tests as shown in the example in:

- https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md#rust-table-driven-tests

versions.yaml
fidencio: Bumping the newest version is super fine, I would like to not bump the minimum one though, not unless we make a decision on what's the minimum supported one and how we can ensure our code builds with minimum, maximum, and everything in between those.  However, that's more a requirement for the moment this code reaches `main`.  For `ccv0` I'd say we're good to go with the bump.

.github/workflows/snap-release.yaml
arronwy: Thanks, I'll fix it

src/agent/src/image_rpc.rs
arronwy: thanks, this warning will be more clear

src/agent/src/image_rpc.rs
arronwy: Thanks, we should use `split_once()`, `source_creds` is a parameter for skopeo which colon to seperate username and password.

src/agent/src/image_rpc.rs
arronwy: I'll rebase to latest head and remove this comment

src/agent/src/image_rpc.rs
arronwy: when cid is empty, since we use the image info as part of container image store path like:
/tmp/busybox:latest/image_oci/
when umoci upack the container image, it will also parase colon and be missleading the tag info

src/agent/src/image_rpc.rs
arronwy: ok, I'll remove this one

src/agent/src/image_rpc.rs
arronwy: thanks, will fix it

src/agent/src/image_rpc.rs
arronwy: I'll move this part in decrypt_layer_data function to reduce the level of indentation.

src/agent/src/image_rpc.rs
arronwy: Yes, cargo fmt generated this format

versions.yaml
arronwy: The reason we update rust is due to we use tonic to build GRPC client to talk with attestation agent, and tonic require newer version of rust. I'll add this info to the commit message and change the version to `1.56.0`

.github/workflows/kata-deploy-push.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
 
       - name: Build ${{ matrix.asset }}
         run: |
+          sudo apt-get install -y git git-extras libipc-run-perl libssl-dev pkg-config protobuf-compiler
```
jodh-intel, line 28: ooi, why is `git-extras` needed here and below?

.github/workflows/static-checks.yaml
```diff
@@ -79,7 +79,7 @@ jobs:
     - name: Check vendored code
       if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
       run: |
-        cd ${GOPATH}/src/github.com/${{ github.repository }} && make vendor
+        cd ${GOPATH}/src/github.com/${{ github.repository }} && CARGO_HOME="/tmp/.cargo" make vendor
```
jodh-intel, line 82: Nit: You could simplify this to `CARGO_HOME="..." make -C $dir vendor`.

src/agent/Cargo.toml
```diff
@@ -61,6 +61,13 @@ vsock-exporter = { path = "vsock-exporter" }
 serde = { version = "1.0.129", features = ["derive"] }
 toml = "0.5.8"
 
+# Image pull/decrypt
+oci-distribution = "0.7"
```
jodh-intel, line 65: - I understand why this is being specified as a partial semver, but all other versions are specified in full semver format, so for consistency this should be too, or a comment added to explain why [*].
- Can you update to the latest version is `"0.8.0"`?

[*] - I wonder if we need to discuss this topic at the AC tbh. However, once we get all rust code to pass `cargo audit` we will introduce a GitHub action to ensure we stay current so full specification shouldn't be an issue.

src/agent/Cargo.toml
```diff
@@ -61,6 +61,13 @@ vsock-exporter = { path = "vsock-exporter" }
...
+ocicrypt-rs = { git = "https://github.com/containers/ocicrypt-rs" }
+sha2 = ">=0.9"
```
jodh-intel, line 67: Same comment as above regarding partial semver.

src/agent/Cargo.toml
```diff
@@ -61,6 +61,13 @@ vsock-exporter = { path = "vsock-exporter" }
...
+# "vendored" feature for openssl is required by musl build
+openssl = { version = "0.10", features = ["vendored"] }
```
jodh-intel, line 69: And again ;)

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,229 @@ impl ImageService {
...
+                    &layer_digest,
+                    image_manifest,
+                    crypto_config,
+                    oci_blob_path,
+                )?;
+            } else if let Some(layer_name) =
+                layer_digest.strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+            {
+                let mut out_file = File::create(oci_blob_path.join(&layer_name))?;
+                info!(sl!(), "Saving layer file {}...", layer_name);
```
jodh-intel, line 140: Please can you remove the ellipsis here (as there is no final "Done" / "saved" message).

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,229 @@ impl ImageService {
...
+    ) -> anyhow::Result<(OciManifest, String, ImageData)> {
+        let reference = Reference::try_from(image)?;
+        let mut client = Client::default();
+        let (image_manifest, _image_digest, image_config) =
+            client.pull_manifest_and_config(&reference, auth).await?;
+
+        let mut image_data = ImageData {
+            layers: Vec::with_capacity(0),
+            digest: None,
+        };
```
jodh-intel, line 166: - It's a shame the `oci-distrubution` crate doesn't derive `Default` here :sob:.
- Nit: `Vec::new()` would be shorter and have the same effect.
- But how about changing this to something like:
  ```suggestion
        let mut image_data: Option<ImageData> = None;
  ```

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,229 @@ impl ImageService {
...
+                        MEDIA_TYPE_LAYER_GZIP_ENC,
+                        MEDIA_TYPE_LAYER_ENC,
+                        MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_ENC,
+                        MEDIA_TYPE_LAYER_NON_DISTRIBUTABLE_GZIP_ENC,
+                    ],
+                )
+                .await
+            {
+                Ok(data) => {
+                    image_data = data;
```
jodh-intel, line 188: This then becomes:

```suggestion
                    image_data = Some(data);
```

src/agent/src/image_rpc.rs
jodh-intel: But doesn't that imply a bug in umoci? Either way, please add a comment in the code as it's not obvious why this rename is required.

jodh-intel (commented):
Thanks for the updates @arronwy. A few more comments and a reminder that some unit tests on this PR would be awesome :smile:

- https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md#rust-table-driven-tests

.github/workflows/kata-deploy-push.yaml
arronwy: I can remove `git-extras` in kata-deploy-push.yaml which is new added, but in snap-release.yaml which already have it before my pr

.github/workflows/static-checks.yaml
arronwy: Fixed now.

src/agent/Cargo.toml
arronwy: Yes, binary shoud use full semver, I changed the version to `0.7.0`, since ocicrypt-rs also have depdency for `oci-distribution` and we not test with `0.8.0` yet, we will update them at the same time

src/agent/src/image_rpc.rs
arronwy: thanks, removed now.

src/agent/src/image_rpc.rs
arronwy: Yes, it's a umoci bug which shoud parse the tag from the righ, added the comment now.

.github/workflows/kata-deploy-push.yaml
arronwy: Samuel found we need add the dependency in docker file, this change will be dropped

src/agent/src/uevent.rs
```diff
@@ -240,7 +240,6 @@ pub(crate) fn spawn_test_watcher(sandbox: Arc<Mutex<Sandbox>>, uev: Uevent) {
                     if matcher.is_match(&uev) {
                         let (_, sender) = watch.take().unwrap();
                         let _ = sender.send(uev.clone());
-                        return;
```
jodh-intel, line 243: This clippy fix should be on a separate commit really.

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,227 @@ impl ImageService {
...
+    }
+
+    fn decrypt_layer_data(
+        layer: &ImageLayer,
+        layer_digest: &str,
+        image_manifest: &mut OciManifest,
+        crypto_config: &CryptoConfig,
+        oci_blob_path: &Path,
+    ) -> Result<()> {
+        if let Some(decrypt_config) = &crypto_config.decrypt_config {
```
jodh-intel, line 94: This function is only called for encrypted layers. But if the function returns `Ok(())` if there is no decrypt config, what happens next? Shouldn't this function error out if there is no decrypt config since the layer will surely be unusable unless you can decrypt it...?

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,227 @@ impl ImageService {
...
+        cid: &str,
+        source_creds: &Option<String>,
+        aa_kbc_params: &str,
+    ) -> Result<()> {
+        let oci_path = Self::build_oci_path(cid);
+        fs::create_dir_all(&oci_path)?;
+
+        info!(
+            sl!(),
+            "Attempting to pull image with native implementation {}...", image
```
jodh-intel, line 216: Nit: Another elllipsis in the log message.

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,227 @@ impl ImageService {
...
+
+        let mut auth = RegistryAuth::Anonymous;
+        if let Some(source_creds) = source_creds {
+            if let Some((username, password)) = source_creds.split_once(':') {
+                auth = RegistryAuth::Basic(username.to_string(), password.to_string());
+            } else {
+                warn!(
+                    sl!(),
+                    "ignoring invalid authentication info ({:?})", source_creds
+                );
```
jodh-intel, line 227: I'm looking at this now wondering whether this should return an `Err()` at this point?

src/agent/src/image_rpc.rs
```diff
@@ -58,11 +309,14 @@ impl ImageService {
...
         let oci_path = tmp_cid_path.join(IMAGE_OCI);
-        let target_path_oci = format!("oci://{}", oci_path.to_string_lossy());
+        let target_path_oci = format!("oci://{}:latest", oci_path.to_string_lossy());
 
         fs::create_dir_all(&oci_path)?;
 
-        info!(sl!(), "Attempting to pull image {}...", &source_image);
+        info!(
+            sl!(),
+            "Attempting to pull image with helper binary {}...", &source_image
```
jodh-intel, line 318: Nit: Unecessary ellipsis.

jodh-intel (commented):
Thanks @arronwy. A few more comments...

src/agent/src/uevent.rs
arronwy: With seperated commit now :)

src/agent/src/image_rpc.rs
arronwy: Return `Err` when no decrypt config avaiable

src/agent/src/image_rpc.rs
arronwy: Thaks, Good suggestion,  when we success get the `image_data`  I return it directly which can avoid this declare.

src/agent/src/image_rpc.rs
arronwy: Remove now

src/agent/src/image_rpc.rs
arronwy: Thanks, use `Err()` now.

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,213 @@ impl ImageService {
         }
     }
 
+    fn build_oci_path(cid: &str) -> PathBuf {
```
jodh-intel, line 80: I wonder if this should return `Result<PathBuf>` for the case where `cid == ""`?

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,213 @@ impl ImageService {
...
+                    let layer_name = format!("{:x}", Sha256::digest(&plaintxt_data));
+                    let mut out_file = File::create(oci_blob_path.join(&layer_name))?;
+                    out_file.write_all(&plaintxt_data)?;
+                    layer_desc.media_type = manifest::IMAGE_LAYER_GZIP_MEDIA_TYPE.to_string();
+
+                    layer_desc.digest = format!("{}:{}", DIGEST_SHA256, layer_name);
+                }
+            }
+        } else {
+            return Err(anyhow!("No decrypt config available"));
```
jodh-intel, line 112: This would be simple to unit test.

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,213 @@ impl ImageService {
...
+        crypto_config: &CryptoConfig,
+        oci_blob_path: &Path,
+    ) -> Result<()> {
+        if let Some(decrypt_config) = &crypto_config.decrypt_config {
+            for layer_desc in image_manifest.layers.iter_mut() {
+                if layer_desc.digest.as_str() == layer_digest {
+                    let (layer_decryptor, _dec_digest) =
+                        decrypt_layer(decrypt_config, layer.data.as_slice(), layer_desc, false)?;
+                    let mut plaintxt_data: Vec<u8> = Vec::new();
+                    let mut decryptor = layer_decryptor.unwrap();
```
jodh-intel, line 100: Please don't use `unwrap()` or `expect()`, except in the scenarios shown in the table in this document:

- https://github.com/kata-containers/kata-containers/blob/main/docs/code-pr-advice.md#rust-language-specifics

Instead, this line should be:

```suggestion
                    let mut decryptor = layer_decryptor?;
```

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,213 @@ impl ImageService {
...
+                    crypto_config,
+                    oci_blob_path,
+                )?;
+            } else if let Some(layer_name) =
+                layer_digest.strip_prefix(format!("{}:", DIGEST_SHA256).as_str())
+            {
+                let mut out_file = File::create(oci_blob_path.join(&layer_name))?;
+                out_file.write_all(&layer.data)?;
+            } else {
+                error!(
```
jodh-intel, line 143: This is also relatively easy to unit test: wrong media type + no layers == `Ok(())` + error message.

Random thought: Rather than returning `Ok(())`, you _could_ maybe return `Result<u64>` where the numeric is the number of layers in the image that have been handled? That way, you could check for `Ok(0)` in the test mentioned above.

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,213 @@ impl ImageService {
...
+                        attempt_interval,
+                        i,
+                        e
+                    );
+                    tokio::time::sleep(tokio::time::Duration::from_secs(attempt_interval)).await;
+                }
+            }
+        }
+
+        Err(anyhow!("Failed to download image data"))
```
jodh-intel, line 196: This should be easy to force in a unit test.

src/agent/src/image_rpc.rs
```diff
@@ -3,29 +3,52 @@
...
 const UMOCI_PATH: &str = "/usr/local/bin/umoci";
-const IMAGE_OCI: &str = "image_oci:latest";
+const IMAGE_OCI: &str = "image_oci";
 const AA_PATH: &str = "/usr/local/bin/attestation-agent";
 const AA_PORT: &str = "127.0.0.1:50000";
 const OCICRYPT_CONFIG_PATH: &str = "/tmp/ocicrypt_config.json";
+const OCI_ANNOTATION_REF_NAME: &str = "org.opencontainers.image.ref.name";
+const OCI_IMAGE_MANIFEST_NAME: &str = "application/vnd.oci.image.manifest.v1+json";
+const OCI_LAYOUT: &str = r#"{"imageLayoutVersion": "1.0.0"}"#;
+const IMAGE_DOCKER_LAYER_FOREIGN_GZIP_MEDIA_TYPE: &str =
```
jodh-intel, line 49: What is a foreign rootfs diff?

src/agent/src/image_rpc.rs
```diff
@@ -47,6 +77,213 @@ impl ImageService {
...
+                    );
+                    tokio::time::sleep(tokio::time::Duration::from_secs(attempt_interval)).await;
+                }
+            }
+        }
+
+        Err(anyhow!("Failed to download image data"))
+    }
+
+    fn pull_image_with_oci_distribution(
```
jodh-intel, line 199: I still think this function is too complex.

jodh-intel (commented):
Thanks @arronwy. A few more comments.

sameo:
@jodh-intel Are you ok with the latest changes for this PR?

jodh-intel:
@sameo - there are a few remaining nit. My main worry is total lack of unit tests, however maybe that could be handled by a follow-up PR?

Given the schedule, I guess this can land, but I'd love to see those unit tests at some point ;)

stevenhorsman:
I've created a couple of commits (https://github.com/stevenhorsman/kata-containers/commit/9a8e695564168e4d5180015ef68f5ed1be40ce93 and https://github.com/stevenhorsman/kata-containers/commit/7eeeda3998a35ec450d346c36b344dcbc4a3176b) to stop skopeo being built and added to the rootfs by default in CCv0, so we can choose to turn it on if we want verify support, but otherwise use the rust implementation.
There is a corresponding test commit https://github.com/stevenhorsman/tests/commit/6f52ff3583bc4b8a57dc03b5b420acabdb317c46 as well for this.

If you want you can pull them into this PR and test them, but as it's pretty close now I think it is probably better to make another PR after this has merged with the changes?

src/agent/src/image_rpc.rs
sameo: I agree we should not unwrap, however the fix should be:

```rust
let mut decryptor = layer_decryptor.ok_or_else(|| anyhow!("Missing layer decryptor"))?;                                                                                                                                                                                             
```

arronwy: Thanks, fixed now.

PR #3039, opened 2021-11-14
stable-2.3:remove non used actions

snir911:
This is backport of #3034, as link is not available anymore this is required in stable otherwise documentation checks will fail

Generally speaking, should we make sure all linked in documentation are pointing to the same branch? it is possible to have a variable to expand the branch name in such way within documentation links?

Fixes: #3033


sameo (approved):

gkurz (approved):
LGTM. Thanks @snir911 !

fidencio:
@snir911, thanks for backporting this one, but it shows something is wrong with the check.
The file was removed, indeed, from **main**.  Checking against main and reporting an error on `stable-2.x` is bad.

Would you mind to open an issue about this?
/cc @jodh-intel in case he happens to have some input.

snir911:
@fidencio indeed, this is what i meant in the PR msg, usually when we point to something in our on repo we don't bother to use the link with the branch path so it points to main, I'll open the issue

Issue #2498, opened 2021-08-24
How do I specify a configuration file in cri-o?


jclab-joseph:
For containerd you can find this at "https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/containerd-kata.md"
But in the case of cri-o, there is no example.

How can I specify a config file in cri-o?

Also, request to improve the documentation.

ariel-adam:
@fgiudici any idea?

c3d:
@jclab-joseph The procedure for `cri-o` should be very similar to what you do for `containerd`. I agree the documentation could be improved.

fgiudici:
CRI-O should be configured to let it know that kata runtime is available.
It could be done adding a configuration snippet like this:
`[crio.runtime.runtimes.kata-shim2]`
`  runtime_path = "/usr/local/bin/containerd-shim-kata-v2"`
`  runtime_type = "vm"`
`  runtime_root = "/run/vc"`
`  privileged_without_host_devices = true`

jclab-joseph:
@fgiudici Yes it can be solved with a shell script containing environment variable. thank you :)

https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/run-kata-with-k8s.md
This documentation says to use the oci runtime.
But oci doesn't work (#2497), vm runtime works fine and fgiudici's advice also works.

fidencio:
While going through the old issues I noticed this one, and took the freedom to just open a PR updating the docs.

PR #2115, opened 2021-06-25
cargo: Use latest nix crate for all Rust code bases

sameo:
Our dependencies already bring several versions of nix, we should avoid
adding even more fragementation.

Fixes #2114

Signed-off-by: Samuel Ortiz <samuel.e.ortiz@protonmail.com>

PR #3046, opened 2021-11-15
Update CRI-O documentation

fidencio:
CRI-O deployment documentation was quite outdated, giving info from the
`1.x` era.  Let's update this to reflect what we currently have.

Fixes: #2498

### DIFF
--- a/docs/how-to/run-kata-with-k8s.md
+++ b/docs/how-to/run-kata-with-k8s.md
@@ -20,7 +20,7 @@ required to spawn pods and containers, and this is the preferred way to run Kata
 An equivalent shim implementation for CRI-O is planned.
 
 ### CRI-O
-For CRI-O installation instructions, refer to the [CRI-O Tutorial](https://github.com/kubernetes-incubator/cri-o/blob/master/tutorial.md) page.
+For CRI-O installation instructions, refer to the [CRI-O Tutorial](https://github.com/cri-o/cri-o/blob/main/tutorial.md) page.
 
 The following sections show how to set up the CRI-O configuration file (default path: `/etc/crio/crio.conf`) for Kata.
 
@@ -30,7 +30,7 @@ Unless otherwise stated, all the following settings are specific to the `crio.ru
 # runtime used and options for how to set up and manage the OCI runtime.
 [crio.runtime]
 ```
-A comprehensive documentation of the configuration file can be found [here](https://github.com/cri-o/cri-o/blob/master/docs/crio.conf.5.md).
+A comprehensive documentation of the configuration file can be found [here](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md).
 
 > **Note**: After any change to this file, the CRI-O daemon have to be restarted with:
 >````
--- a/docs/how-to/run-kata-with-k8s.md
+++ b/docs/how-to/run-kata-with-k8s.md
@@ -59,57 +59,6 @@ You can also add multiple entries to specify alternatives hypervisors, e.g.:
   runtime_type = "oci"
 ```
 
-#### Untrusted annotation (until CRI-O v1.12)
-The untrusted annotation is used to specify a runtime for __untrusted__ workloads, i.e.
-a runtime to be used when the workload cannot be trusted and a higher level of security
-is required. An additional flag can be used to let CRI-O know if a workload
-should be considered _trusted_ or _untrusted_ by default.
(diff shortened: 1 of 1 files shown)


fidencio:
Oooi?
```
M src/agent/Cargo.lock
index 0b320adb..43cf60f1 100644
--- a/src/agent/Cargo.lock
make[1]: *** [Makefile:675: vendor] Error 1
+++ b/src/agent/Cargo.lock
@@ -1,7 +1,5 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
-
make: *** [Makefile:31: vendor-runtime] Error 2
 [[package]]
 name = "addr2line"
 version = "0.16.0"
make[1]: Leaving directory '/home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime'
```

Not sure what's triggering this here :-/

fidencio:
cc @jodh-intel, any idea on what could be triggering this weirdness in the `cargo vendor` check?

jodh-intel:
@fidencio - Looks like you might have inadvertently modified `src/agent/Cargo.lock` on this (doc only) PR?

fidencio:
> @fidencio - Looks like you might have inadvertently modified `src/agent/Cargo.lock` on this (doc only) PR?

I don't think so.  The error happened, and then I ran `cargo update` to see whether it'd magically fix it (it didn't).
Now I just removed the commit `cargo update` commit, it will break again on the same point and you can see the diff doesn't touch anything related to the agent.

docs/how-to/run-kata-with-k8s.md
```diff
@@ -40,74 +40,13 @@ A comprehensive documentation of the configuration file can be found [here](http
...
-#### Network namespace management
-To enable networking for the workloads run by Kata, CRI-O needs to be configured to
-manage network namespaces, by setting the following key to `true`.
-
-In CRI-O v1.16:
-```toml
-manage_network_ns_lifecycle = true
-```
-In CRI-O v1.17+:
+To use this feature, Kata must added as a runtime handler using, and that can be done by dropping a `50-kata` snippet file into `/etc/crio/crio.conf.d`, with the content shown below:
```
fgiudici, line 43: I think something got deleted after "using".  Maybe we can just leave how we drop the config...
```suggestion
To use this feature, Kata must added as a runtime handler. This can be done by dropping a `50-kata` snippet file into `/etc/crio/crio.conf.d`, with the content shown below:
```

fgiudici (approved):
Clearly the cargo error is not related to this PR.
LGTM, thanks @fidencio !

jodh-intel:
@fidencio - Taking a closer look at this, I think it's the recent downgrade of rust:

- https://github.com/kata-containers/tests/pull/4189

We were using rust v1.54.0 for the CI, but now that PR has landed, we've reverted to v1.45.0.

Looking at the rust release notes for v1.53.0:

- https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1530-2021-06-17

> Cargo now supports git repositories where the default HEAD branch is not "master". This also includes a switch to the version 3 Cargo.lock format which can handle default branches correctly.

That seems to align with the error you're seeing:

```diff
diff --git a/src/agent/Cargo.lock b/src/agent/Cargo.lock
index 0b320adb..43cf60f1 100644
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -1,7 +1,5 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
-
 [[package]]
 name = "addr2line"
 version = "0.16.0"
```

That `version = 3` line got added on https://github.com/kata-containers/kata-containers/pull/2115, presumably from running a version of rust localy >= v1.53.0. Maybe @sameo can confirm? But the version of rust in the CI is complaining about that line.

Hence, I think the fix is to remove that line. The safest way would be to install rust <  v1.53.0 locally, regenerate the lockfile, commit and push to a PR branch.

/cc @dgibson.

docs/how-to/run-kata-with-k8s.md
fidencio: @fgiudici, done!

sameo (approved):

jodh-intel (approved):
Thanks @fidencio.

lgtm

Jakob-Naucke:
static checks are complaining -- this does need a backport

PR #3066, opened 2021-11-18
PoC/RFC: support passing devices by annotations

liubin:
K8s can't support specify `device` properties by its definition YAML, this may make some limitations for users.

We can simply pass devices info to Kata Containers by annotations.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: mypod
  annotations:
    io.katacontainers.container.devices: "/var/run/kata-containers/vhost-user/block/devices/vhostblk0:/dev/vdb"
spec:
  runtimeClassName: kata
  containers:
  - name: mypod
    image: docker.io/containerstack/alpine-stress
    command:
      - top
```

And run Pods:

```bash
# kubectl create -f mypod.yaml 
# kubectl get pods -w
# kubectl exec -it mypod sh
/ # ls -tl /dev/vdb 
brw-r--r--    1 root     root      254,   0 Nov 18 08:01 /dev/vdb
/ # dd if=/dev/vdb of=/tmp/ddtest bs=4k count=20
20+0 records in
20+0 records out
/ # 
```


liubin:
https://github.com/kata-containers/govmm/pull/206 will block this PR

src/runtime/pkg/oci/utils.go
```diff
@@ -362,6 +363,91 @@ func SandboxID(spec specs.Spec) (string, error) {
 	return "", fmt.Errorf("Could not find sandbox ID")
 }
 
+func addContainerAnnotations(ocispec specs.Spec, _config *vc.ContainerConfig) error {
```
shuochen0311, line 366: This is add container device from annotation right? Not add container annotation

liubin: Yes, this function is for updating the container's spec or config, now it only does such work for devices, but for future usage, I chose this name, according to the same naming style for updating sandbox config.

amshinde:
@liubin Devices are passed in k8s using device plugins or CSI in case of storage devices. Here we are essentially basically digressing from the standard techniques and instead passing devices using annotations. 

I would like to open this for discussion first if we want to support this at all.
cc @jodh-intel @fidencio @kata-containers/architecture-committee

liubin:
@amshinde Yes, this is not the most standard and ideal way, but it can be used in the following scenarios:

- Don't use k8s(ctr or crictl)
- Don't want to write a CSI driver or device plugin

shuochen0311:
The following deployment yaml file does not work with your pr:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment1
  annotations:
    io.katacontainers.container.devices: "/var/run/kata-containers/vhost-user/block/devices/vhostblk1:/dev/vdb"
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 1 # tells deployment to run 2 pods matching the template
  template:
    metadata:
      labels:
        app: nginx
    spec:
      runtimeClassName: kata-qemu
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
        securityContext:
          privileged: true
```
Looks like deployment annotation can not be reached from your function call branch?

liubin:
@shuochen0311 You are right. This patch can't work with Deployment.

The annotations are passed from Deployment to ReplicaSet, but not Pod, the Pod has no annotations copied from ReplicaSet.

This will be a fatal limitation.

src/runtime/pkg/oci/utils.go
```diff
@@ -362,6 +363,91 @@ func SandboxID(spec specs.Spec) (string, error) {
...
+// Valid mode is a composition of r (read), w (write), and m (mknod).
+func validDeviceMode(mode string) bool {
+	var legalDeviceMode = map[rune]bool{
+		'r': true,
+		'w': true,
+		'm': true,
+	}
+	if mode == "" {
+		return false
+	}
```
jodh-intel, line 441: Nit: This could be checked before defining `legalDeviceMode`.

src/runtime/pkg/oci/utils_test.go
```diff
@@ -1014,3 +1014,68 @@ func TestParseAnnotationBoolConfiguration(t *testing.T) {
...
+		annotaionValue string
+		err            bool
+		result         parseResult
+	}{
+		{
+			annotaionValue: "",
+			err:            true,
+			result:         parseResult{},
+		},
+		{
```
jodh-intel, line 1037: Can you add some tests with things like an:

- `/this/path/is/invalid`
- `../../../../`
- `.`
- `..`
- `/`
- unreadable path? (I'm thinking rootless / non-priv)

jodh-intel:
Thanks for raising @liubin. A few questions:

- If we add this feature, will it be a temporary one? Leading on to...

- > K8s can't support specify device properties by its definition YAML
  Do you know if this limitation is being worked on upstream? If so, is there an issue and an ETA?

- Can anyone see any potential security implications of this change?

- Is there any value in making `io.katacontainers.container.devices` a JSON fragment?

/cc @dgibson for thoughts on all this.

dgibson:
> Thanks for raising @liubin. A few questions:
> 
>     * If we add this feature, will it be a temporary one? Leading on to...
> 
>     * > K8s can't support specify device properties by its definition YAML
>       > Do you know if this limitation is being worked on upstream? If so, is there an issue and an ETA?
> 
>     * Can anyone see any potential security implications of this change?

I don't know off hand if k8s restricts what annotations a user is allowed to set on pods.  If a user *is* able to set this annotation, then this is not so much a "security implication" as a complete removal of security, since it would allow a pod full access to an arbitrary host device.

>     * Is there any value in making `io.katacontainers.container.devices` a JSON fragment?
> 
> 
> /cc @dgibson for thoughts on all this.

jodh-intel:
Any update on this @liubin ?

Issue #3001, opened 2021-11-10
test_kata_deploy should validate the testing PR too


bergwolf:
Right now it uses v2.0.0 release to make sure kata-deploy works. But when tagging a new release, we really want to know if it works for the current code as well. So ideally it should validate the current PR as well. 

/cc @fidencio @egernst

PR #3071, opened 2021-11-18
kata-deploy: Ensure we test HEAD with `/test_kata_deploy`

fidencio:
Is the past few releases we ended up hitting issues that could be easily
avoided if `/test_kata_deploy` would use HEAD instead of a specific
tarball.

By the end of the day, we want to ensure kata-deploy works, but before
we cut a release we also want to ensure that the binaries used in that
release are in a good shape.  If we don't do that we end up either
having to roll a release back, or to cut a second release in a really
short time (and that's time consuming).

Note: there's code duplication here that could and should be avoided,b
but I sincerely would prefer treating it in a different PR.

Fixes: #3001

### DIFF
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -5,46 +5,91 @@ on:
 name: test-kata-deploy
 
 jobs:
-  create-and-test-container:
-    if: |
-      github.event.issue.pull_request
-      && github.event_name == 'issue_comment'
-      && github.event.action == 'created'
-      && startsWith(github.event.comment.body, '/test_kata_deploy')
+  build-asset:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        asset:
+          - cloud-hypervisor
+          - firecracker
+          - kernel
+          - qemu
+          - rootfs-image
+          - rootfs-initrd
+          - shim-v2
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install docker
+        run: |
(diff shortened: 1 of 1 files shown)


fidencio:
> /test_kata_deploy

Oh, yeah, I can't test it here.  I need to run this from a PR where this code is already part of the repo.  Let me test it on my own branch, update this PR and report the results accordingly.

.github/workflows/kata-deploy-test.yaml
```diff
@@ -5,47 +5,91 @@ on:
...
-            echo "reference for PR: " ${ref}
-            echo "##[set-output name=pr-ref;]${ref}"
+          curl -fsSL https://test.docker.com -o test-docker.sh
+          sh test-docker.sh
 
-      - name: check out
-        uses: actions/checkout@v2
+      - name: Build ${{ matrix.asset }}
+        run: |
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh --build="${KATA_ASSET}"
```
jcvenegas, line 29: Now should be possible use the toplevel makefile with
```
make binary-tarball
```

fidencio: Changed!

fidencio:
https://github.com/fidencio/kata-containers/runs/4258246865?check_suite_focus=true shows that this works well, but I don't have neither quay.io nor aks credentials set up on my side.  But, clearly, the tarball generated from HEAD was being used.

bergwolf (approved):

jcvenegas (approved):

Issue #3086, opened 2021-11-19
 `./static-build/kernel/build.sh` fails to build the kernel


wainersm:
# Description of problem

The script `./static-build/kernel/build.sh` fails to build the kernel because `yq` is not installed on the build root container.

# Expected result

The kernel is built inside a container and then copied to the host.

# Further information

Build log:
```
[root@xxxxxx packaging]# ./static-build/kernel/build.sh 
STEP 1: FROM ubuntu
STEP 2: ENV DEBIAN_FRONTEND=noninteractive
--> 1dc7931c604
STEP 3: RUN apt update
Get:1 http://archive.ubuntu.com/ubuntu focal InRelease [265 kB]          
Get:2 http://security.ubuntu.com/ubuntu focal-security InRelease [114 kB]
Get:3 http://security.ubuntu.com/ubuntu focal-security/restricted amd64 Packages [682 kB]
Get:4 http://archive.ubuntu.com/ubuntu focal-updates InRelease [114 kB]
Get:5 http://archive.ubuntu.com/ubuntu focal-backports InRelease [101 kB]
Get:6 http://archive.ubuntu.com/ubuntu focal/universe amd64 Packages [11.3 MB]
Get:7 http://security.ubuntu.com/ubuntu focal-security/multiverse amd64 Packages [30.1 kB]
Get:8 http://security.ubuntu.com/ubuntu focal-security/main amd64 Packages [1232 kB]
Get:9 http://security.ubuntu.com/ubuntu focal-security/universe amd64 Packages [809 kB]
Get:10 http://archive.ubuntu.com/ubuntu focal/restricted amd64 Packages [33.4 kB]
Get:11 http://archive.ubuntu.com/ubuntu focal/multiverse amd64 Packages [177 kB]
Get:12 http://archive.ubuntu.com/ubuntu focal/main amd64 Packages [1275 kB]
Get:13 http://archive.ubuntu.com/ubuntu focal-updates/universe amd64 Packages [1098 kB]
Get:14 http://archive.ubuntu.com/ubuntu focal-updates/restricted amd64 Packages [738 kB]
Get:15 http://archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages [1689 kB]
Get:16 http://archive.ubuntu.com/ubuntu focal-updates/multiverse amd64 Packages [33.3 kB]
Get:17 http://archive.ubuntu.com/ubuntu focal-backports/main amd64 Packages [2668 B]
[...]

PR #3090, opened 2021-11-22
tools/packaging: use clone_tests_repo() from ci/lib.sh

Bevisy:
This changed the tools/packaging/scripts/lib.sh to source from
ci/lib.sh so that clone_tests_repo() is reused. Its implementation
from clone_tests_repo() is more convenient because doesn't use
go to clone the tests repository (some scripts run on containers
where golang is not installed).
    
Fixes: #3086

### DIFF
--- a/tools/packaging/scripts/lib.sh
+++ b/tools/packaging/scripts/lib.sh
@@ -6,24 +6,13 @@
 #
 
 export GOPATH=${GOPATH:-${HOME}/go}
-export tests_repo="${tests_repo:-github.com/kata-containers/tests}"
-export tests_repo_dir="$GOPATH/src/$tests_repo"
 
 this_script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${this_script_dir}/../../../ci/lib.sh"
 
 short_commit_length=10
 
 hub_bin="hub-bin"
-
-clone_tests_repo() {
-	# KATA_CI_NO_NETWORK is (has to be) ignored if there is
-	# no existing clone.
-	if [ -d "${tests_repo_dir}" ] && [ -n "${KATA_CI_NO_NETWORK:-}" ]; then
-		return
-	fi
-
-	go get -d -u "$tests_repo" || true
-}
 
 install_yq() {
 	clone_tests_repo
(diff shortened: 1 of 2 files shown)


Bevisy:
Check the yq image in dockehub, found that version [3.4.1](https://hub.docker.com/r/mikefarah/yq/tags?page=1&name=3.4.1) image only supports amd64 architecture, maybe need to recreate the yq multi-architecture image.

tools/packaging/static-build/kernel/Dockerfile
```diff
@@ -2,9 +2,13 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+FROM mikefarah/yq:3.4.1 as build
```
fidencio, line 5: @Bevisy, I'd like to suggest a different approach here.

Instead of using a multi-stage build and end up depending on an image that may or may not be updated, or that may or may not be supported in different architectures, why not doing something like what's done on https://github.com/kata-containers/tests/blob/main/.ci/install_yq.sh?

fidencio (changes requested):
@Bevisy, thanks for the contribution, please, take a look at the suggestion made "inline".

tools/packaging/static-build/kernel/Dockerfile
Bevisy: Reasonable. A simpler way to support multiple architectures makes more sense. The lack of multi-architecture image makes the use of multi-stage builds seem stupid, although I was planning to maintain a multi-architecture image of 3.4.1 yq.

wainersm: I actually like the a idea of using the multi-stage build but the lack of images for non-x86_64 architectures makes the solution complicated.

On https://github.com/kata-containers/kata-containers/issues/3086#issuecomment-975754430 I pasted a diff with changes that could be made to install yq with `install_yq.sh` script as suggested by @fidencio

Bevisy: @wainersm @fidencio By looking at the normal github-action ci/kata-deploy, it is executing docker inside a docker environment with the yq tool prepared. So directly executing static-build will fail, but I don't feel the need to fix this failure because we are no longer using a context with the ci environment when the problem is found. In fact, what we did was to prepare yq in the image, so if it is your own local build, you can still consider using a multi-stage build, you just need to simply add two lines of dockerfile, but for ci execution, I think we can keep the status quo.

Bevisy:
Closed this pr for [comments](https://github.com/kata-containers/kata-containers/pull/3090#discussion_r754898118). Thx all.

Issue #3048, opened 2021-11-16
agent: musl target build is broken


bergwolf:
It seems that `LIBC=xxx` is ignored now?

```
$make LIBC=musl
<snip...>
$ldd target/x86_64-unknown-linux-musl/release/kata-agent
        linux-vdso.so.1 (0x00007fffff9dd000)
        libseccomp.so.2 => /lib64/libseccomp.so.2 (0x00007f2c4de30000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f2c4da6a000)
        /lib/ld64.so.1 => /lib64/ld-linux-x86-64.so.2 (0x00007f2c4e07e000)
```

bergwolf:
Ok, turns out that I have to disable seccomp support as well.

fidencio:
> Ok, turns out that I have to disable seccomp support as well.

@bergwolf, just thinking out-loud here, but wouldn't be reasonable to keep this issue open and work on our Makefile to disable seccomp when using `LIBC=musl`, unless its usage is explicitly required?

ManaSugi:
@bergwolf @fidencio 
Thank you for the issue.
If `LIBSECCOMP_LINK_TYPE=static`  is not set, the Kata agent tries to link the `libseccomp` dynamically that already exists in the system, so the `kata-agent` binary is generated as above.
I may need to fix the `Makefile` to build the agent with seccomp capability only if `LIBSECCOMP_LINK_TYPE=static` and `LIBSECCOMP_LIB_PATH` environment variables are set as follows.
https://github.com/kata-containers/kata-containers/blob/main/tools/osbuilder/rootfs-builder/rootfs.sh#L579

If the variables are not set, the agent is built with `SECCOMP=no` forcibly.
Could you give me your thoughts? Thank you.

Issue #3088, opened 2021-11-22
agent: `make test` failed


YchauWang:
# Description of problem

Running the `make test` under the [agent](https://github.com/kata-containers/kata-containers/tree/main/src/agent) failed, it reported "error: test failed, to rerun pass '-p rustjail --lib'".

# Expected result

Execute `make test` for the agent project correctly.

# Further information

```
root@david-vm:/workspaces/go/src/github.com/kata-containers/kata-containers/src/agent# make test
...
running 55 tests
test cgroups::fs::tests::test_line_to_vec ... ok
test cgroups::fs::tests::test_lines_to_map ... ok
test container::tests::test_linuxcontainer_config ... ok
test container::tests::test_linuxcontainer_destroy ... ok
test container::tests::test_linuxcontainer_do_init_child ... ok
test container::tests::test_linuxcontainer_exec ... ok
test container::tests::test_linuxcontainer_get_process ... ok
test container::tests::test_execute_hook ... ok
test container::tests::test_linuxcontainer_get_process_not_found ... ok
test container::tests::test_linuxcontainer_oci_state_no_root_parent ... ok
test container::tests::test_linuxcontainer_oci_state ... ok
test container::tests::test_linuxcontainer_pause ... ok
test container::tests::test_linuxcontainer_pause_bad_status ... ok
test container::tests::test_linuxcontainer_pause_cgroupmgr_is_none ... ok
test container::tests::test_linuxcontainer_processes ... ok
test container::tests::test_linuxcontainer_resume ... ok
test container::tests::test_linuxcontainer_resume_bad_status ... ok
test container::tests::test_linuxcontainer_resume_cgroupmgr_is_none ... ok
test container::tests::test_linuxcontainer_run ... ok
test container::tests::test_linuxcontainer_set ... ok
test container::tests::test_linuxcontainer_signal ... ok
test container::tests::test_linuxcontainer_state ... ok
test container::tests::test_linuxcontainer_start ... ok
test container::tests::test_namespaces ... ok
test container::tests::test_set_stdio_permissions ... ok
test container::tests::test_linuxcontainer_stats ... ok
test container::tests::test_status_transtition ... ok
test container::tests::test_typetoname ... ok
test container::tests::test_valid_env ... ok
test mount::tests::test_check_proc_mount ... ok
test mount::tests::test_dev_rel_path ... ok
old cwd: /workspaces/go/src/github.com/kata-containers/kata-containers/src/agent/rustjail
[...]

PR #3094, opened 2021-11-22
docs: Improve the documentation about how to build the agent with libseccomp

ManaSugi:
Add descriptions about how to build the agent with the libseccomp library
to the developer guide. In order to build the agent linked with the libseccomp
statically by hand, not using the osbuilder script, the user will need to build
the libseccomp from the sources using our `install_libseccomp.sh` script
and set the environment variables for the libseccomp crate. If the user wants
to build the agent linked with the libseccomp dynamically, the user needs to
specify `LIBC=gnu` to use `glibc`.

Fixes: #3048, #3088

### DIFF
--- a/docs/Developer-Guide.md
+++ b/docs/Developer-Guide.md
@@ -210,7 +210,8 @@ $ sudo systemctl restart systemd-journald
 
 > **Note:**
 >
-> - You should only do this step if you are testing with the latest version of the agent.
+> - You should only do this step if you are testing with the latest version of the agent or you want to build
+>   a custom agent without seccomp feature.
 
 The agent is built with a statically linked `musl.` The default `libc` used is `musl`, but on `ppc64le` and `s390x`, `gnu` should be used. To configure this:
 
@@ -228,16 +229,70 @@ $ go get -d -u github.com/kata-containers/kata-containers
 $ cd $GOPATH/src/github.com/kata-containers/kata-containers/src/agent && make
 ```
 
-The agent is built with seccomp capability by default.
-If you want to build the agent without the seccomp capability, you need to run `make` with `SECCOMP=no` as follows.
+### Build the agent with seccomp
 
+The agent is built with seccomp feature by default.
+You **MUST** install the `libseccomp` library in your build environment and need to set the environment
+variables for the [`libseccomp` crate](https://github.com/libseccomp-rs/libseccomp-rs) as necessary to
+link the agent binary with a static `libseccomp` library.
+
+#### Link the `libseccomp` library statically against `musl`
+
+Install a static `libseccomp` library for `musl`:
+
+```sh
+$ sudo $GOPATH/src/github.com/kata-containers/kata-containers/ci/install_libseccomp.sh /usr/local/kata-libseccomp /usr/local/kata-gperf
 ```
+
+> **Note:**
+>
+> - If you want to build the agent linked with the `libseccomp` library statically against `musl`, you'll
+>   need to compile the `libseccomp` library using the above script to generate the static library for `musl`
+>   or use a `musl`-based distribution (e.g. `Alpine`) that provides a package for the static library for `musl`.
+
+Set the environment variables for the `libseccomp` crate:
+
+```sh
+$ export LIBSECCOMP_LINK_TYPE=static
+$ export LIBSECCOMP_LIB_PATH=/usr/local/kata-libseccomp/lib
+```
+
+> **Note:**
+>
+> - If you set `LIBSECCOMP_LINK_TYPE` to `static`, `LIBSECCOMP_LIB_PATH` **MUST** be set, or the build fails.
+
+Build the agent:
+
+```sh
+$ make -C $GOPATH/src/github.com/kata-containers/kata-containers/src/agent
+```
+
+#### Link the `libseccomp` library dynamically
+
+Install a shared `libseccomp` library:
+
+> e.g. `libseccomp-dev` for Ubuntu, or `libseccomp-devel` for CentOS
+
+If you set neither the environment variables for the `libseccomp` crate, the agent will assume the `libseccomp` library
+is available dynamically.
+
+Build the agent with `LIBC=gnu` for a `gnu`-based rootfs:
+
+```sh
+$ make -C $GOPATH/src/github.com/kata-containers/kata-containers/src/agent LIBC=gnu
+```
+
+### Build the agent without seccomp feature
+
+If you want to build the agent without seccomp feature, you need to run `make` with `SECCOMP=no` as follows.
+
+```sh
 $ make -C $GOPATH/src/github.com/kata-containers/kata-containers/src/agent SECCOMP=no
 ```
 
 > **Note:**
 >
-> - If you enable seccomp in the main configuration file but build the agent without seccomp capability,
+> - If you enable seccomp in the main configuration file but build the agent without seccomp feature,
 >   the runtime exits conservatively with an error message.
 
 ## Get the osbuilder
@@ -268,10 +323,10 @@ You can get a supported distributions list in the Kata Containers by running the
 $ ./rootfs.sh -l
 ```
 
-If you want to build the agent without seccomp capability, you need to run the `rootfs.sh` script with `SECCOMP=no` as follows.
+If you want to build the agent without seccomp feature, you need to run the `rootfs.sh` script with `SECCOMP=no` as follows.
 
 ```
-$ script -fec 'sudo -E GOPATH=$GOPATH AGENT_INIT=yes USE_DOCKER=true SECCOMP=no ./rootfs.sh ${distro}'
+$ script -fec 'sudo -E GOPATH=$GOPATH USE_DOCKER=true SECCOMP=no ./rootfs.sh ${distro}'
 ```
 
 > **Note:**
@@ -285,12 +340,24 @@ $ script -fec 'sudo -E GOPATH=$GOPATH AGENT_INIT=yes USE_DOCKER=true SECCOMP=no 
 
 > **Note:**
 >
-> - You should only do this step if you are testing with the latest version of the agent.
+> - You should only do this step if you are testing with the latest version of the agent or you want to add
+>   a custom agent for seccomp to a rootfs.
 
 ```
 $ sudo install -o root -g root -m 0550 -t ${ROOTFS_DIR}/usr/bin ../../../src/agent/target/x86_64-unknown-linux-musl/release/kata-agent
 $ sudo install -o root -g root -m 0440 ../../../src/agent/kata-agent.service ${ROOTFS_DIR}/usr/lib/systemd/system/
 $ sudo install -o root -g root -m 0440 ../../../src/agent/kata-containers.target ${ROOTFS_DIR}/usr/lib/systemd/system/
+```
+
+The agent binary is linked with a static `libseccomp` library (`libseccomp.a`) by default.
+If you want to create a rootfs including the agent binary linked with a shared `libseccomp` library
+(`libseccomp.so`), see the [Link the `libseccomp` library dynamically](#link-the-libseccomp-library-dynamically)
+section to build a custom agent according to the `$distro` you chose and add the agent to the image as follows.
+
+```sh
+$ export DISTRO=${distro}
+$ [ ${DISTRO} = "alpine" ] && export LIBC=musl || export LIBC=gnu
+$ sudo install -o root -g root -m 0550 -t ${ROOTFS_DIR}/usr/bin ../../../src/agent/target/x86_64-unknown-linux-${LIBC}/release/kata-agent
 ```
 
 ### Build a rootfs image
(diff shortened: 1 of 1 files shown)


liubin (dismissed):

fidencio:
@ManaSugi, first of all, thanks for improving the documentation and the Makefile for using the seccomp feature.

One thing that I don't fully understand though, is how we differentiate the static & dynamic linkage with the current state of our Makefile.  For instance, we should be able to use the seccomp feature by dynamically linking against `libseccomp.so.2`, when building using GNU libc instead of musl libc, right?  My understanding (which is limited and may be wrong) is that our Makefile is not prepared for that.  Is this the case or am I missing something?

The main reason I'm asking is because some architectures (such as s390x and ppc64le) and some companies (such as Red Hat) either won't use musl due to the lack of support or will prefer to using the GNU libc, and we should consider also providing the `seccomp` support on those cases.

ManaSugi:
@fidencio Thank you for your comment. 

If you want to use the seccomp feature by dynamically linking,
you can set `LIBSECCOMP_LINK_TYPE=dylib` (default) according to the [rust link flag](https://doc.rust-lang.org/rustc/command-line-arguments.html#option-l-link-lib).
Also, if you do not set `LIBSECCOMP_LINK_TYPE`, `libseccomp-rs` tries to link the libseccomp dynamically by default.
https://github.com/ManaSugi/libseccomp-rs/blob/main/libseccomp-sys/build.rs#L13

Therefore, we can build an agent with dynamically linked the libseccomp library. However, I think we'll have to set `rpath` additionally for `libseccomp.so.2` when building the agent in order to run the agent inside the VM because the build environment and the execution environment are different.

Regarding `s390` and `ppc64le`, the current `ci/install_libseccomp.sh` does not set `cflags` for `musl`,
so the agent is built with statically linked libseccomp using `GNU libc` currently.
https://github.com/kata-containers/kata-containers/blob/main/ci/install_libseccomp.sh#L50

I hope that it answers your questions.

fidencio:
> @fidencio Thank you for your comment.
> 
> If you want to use the seccomp feature by dynamically linking, you can set `LIBSECCOMP_LINK_TYPE=dylib` (default) according to the [rust link flag](https://doc.rust-lang.org/rustc/command-line-arguments.html#option-l-link-lib). Also, if you do not set `LIBSECCOMP_LINK_TYPE`, `libseccomp-rs` tries to link the libseccomp dynamically by default. https://github.com/ManaSugi/libseccomp-rs/blob/main/libseccomp-sys/build.rs#L13
> 
> Therefore, we can build an agent with dynamically linked the libseccomp library. However, I think we'll have to set `rpath` additionally for `libseccomp.so.2` when building the agent in order to run the agent inside the VM because the build environment and the execution environment are different.
> 
> Regarding `s390` and `ppc64le`, the current `ci/install_libseccomp.sh` does not set `cflags` for `musl`, so the agent is built with statically linked libseccomp using `GNU libc` currently. https://github.com/kata-containers/kata-containers/blob/main/ci/install_libseccomp.sh#L50
> 
> I hope that it answers your questions.

I would like to see someone from @kata-containers/redhat to chime on this and ensure this works for them as well.
Till we get someone from their side to review, I'm putting this on hold (and, please, reviewer, remove the "do-not-merge" label).

c3d:
I can't say that I really like the approach of building without seccomp by default, and require active changes in all downstream consumers to re-enable it. This is likely to trigger errors downstream.

@kata-containers/redhat has been notified thanks to @fidencio, and @gkurz has been following up internally. We can obviously change our own builds. However, we are presumably not the only consumers, and the fact that the the build silently changes from the current default to a less safe one is really annoying.

@ManaSugi Looking at the problem in #3088, I can see two other ways to address the problem.

1. Explicitly pass `SECCOMP=no` in the environments where the test fails (i.e. fix the test build, not all builds)
2. Deal more gracefully with `init_seccomp()` failing (i.e. not summarily `unwrap()`ing it. Possibly, we could even have an agent that gracefully reports the runtime issue to suggest that `libseccomp` should be added to the image (and where) with a nice log message.

docs/Developer-Guide.md
```diff
@@ -219,15 +219,36 @@ rustup target add x86_64-unknown-linux-musl
 sudo ln -s /usr/bin/g++ /bin/musl-g++
 ```
 
+The agent is built with seccomp capability by default.
```
c3d, line 222: `seccomp`

c3d: This line becomes false with the changes you submit next, since the default is not to have the two required variables set.

docs/Developer-Guide.md
```diff
@@ -219,15 +219,36 @@ rustup target add x86_64-unknown-linux-musl
...
+
+To install `libseccomp` library:
+
+```
+$ $GOPATH/src/github.com/kata-containers/kata-containers/install_libseccomp.sh /tmp/kata-libseccomp /tmp/kata-gperf
+```
+
+> **Note**
+>
+> - You should build `libseccomp` library from sources using the above script to generate the static library for `musl`.
```
c3d, line 234: Could you elaborate a bit why this is needed?

(I'm interested because, while we do not use `musl` ourselves, any bad interaction with the system `libseccomp` is interesting for me to know).

docs/Developer-Guide.md
```diff
@@ -219,15 +219,36 @@ rustup target add x86_64-unknown-linux-musl
...
+
+To set environment variables for static link:
```
c3d, line 236: Could you also document the dynamic link case and explain how to build an image with `libseccomp` in it?

docs/Developer-Guide.md
```diff
@@ -219,15 +219,36 @@ rustup target add x86_64-unknown-linux-musl
...
 To build the agent:
 
 ```
 $ go get -d -u github.com/kata-containers/kata-containers
 $ cd $GOPATH/src/github.com/kata-containers/kata-containers/src/agent && make
 ```
 
-The agent is built with seccomp capability by default.
 If you want to build the agent without the seccomp capability, you need to run `make` with `SECCOMP=no` as follows.
+In addition, if you do not set `LIBSECCOMP_LINK_TYPE` or `LIBSECCOMP_LIB_PATH`, the agent is built with `SECCOMP=no` forcibly.
```
c3d, line 251: This contradicts the "by default" above, and is in my opinion an undesirable behavior.

What about changing it to (and implementing as):

- If you set `LIBSECCOMP_LINK_TYPE` to `static`, `LIBSECCOMP_LIB_PATH` must be set or the build with fail.
- If you set `LIBSECCOMP_LIB_PATH` explicitly, then it must point to a `libseccomp` library built in your build environment, and `LIBSECCOMP_LINK_TYPE` will default to `static` if not set.
- If you set neither variables and you have not set `SECCOMP=no`, then the agent will assume that `libseccomp` is available dynamically, and report an error otherwise.

src/agent/Makefile
```diff
@@ -32,6 +32,16 @@ EXTRA_RUSTFEATURES :=
 ##VAR SECCOMP=yes|no define if agent enables seccomp feature
 SECCOMP := yes
 
+# If LIBSECCOMP_LINK_TYPE or LIBSECCOMP_LIB_PATH is not set,
+# we assume that libseccomp is not installed in the environment.
+# Therefore, the agent disables seccomp feature.
```
c3d, line 37: If you agree with my previous comment, then that logic would obviously have to be changed ;-)

c3d (changes requested):
For the moment, I don't like the changes as is.

I think that at the minimum, you should add another commit with a better handling if `init_seccomp()` returning an error than just `unwrap()`. It's better if we have a clean error in the log than a panic.

jodh-intel:
Any update on this @ManaSugi?

ManaSugi:
@jodh-intel I apologize for the late update. I'll update this PR next week.

jodh-intel:
Any update on this @ManaSugi?

ManaSugi:
> I can't say that I really like the approach of building without seccomp by default, and require active changes in all downstream consumers to re-enable it. This is likely to trigger errors downstream.
> 
> @kata-containers/redhat has been notified thanks to @fidencio, and @gkurz has been following up internally. We can obviously change our own builds. However, we are presumably not the only consumers, and the fact that the the build silently changes from the current default to a less safe one is really annoying.
> 
> @ManaSugi Looking at the problem in #3088, I can see two other ways to address the problem.
> 
> 1. Explicitly pass `SECCOMP=no` in the environments where the test fails (i.e. fix the test build, not all builds)
> 2. Deal more gracefully with `init_seccomp()` failing (i.e. not summarily `unwrap()`ing it. Possibly, we could even have an agent that gracefully reports the runtime issue to suggest that `libseccomp` should be added to the image (and where) with a nice log message.

@c3d Thank you for the review, and I do aplogize for the bad fixes and the late reply. 
I updated it and added the comments.

docs/Developer-Guide.md
ManaSugi: I removed the changes from `Makefile`.

docs/Developer-Guide.md
ManaSugi: I added a description about this to `Developers-Guide` as follows.

> If you want to build the agent linked the `libseccomp` library statically against `musl`, you need to compile the
> `libseccomp` library using the above script to generate the static library for `musl` or use a `musl`-based distribution
> (e.g. `Alpine`) that provides a package for the static library that was compiled against `musl`.

Most `libseccomp` distribution packages are compiled with `FORTIFY_SOURCE=2` by defaut that provides lightweight support for detecting buffer overflows in various functions, but `musl`-libc does not have the functions are provided by `FORTIFY_SOURCE`. Hence, we need to compile `libseccomp` with `FORTIFY_SOURCE=1`, or a `libseccomp` package that `musl`-based distribution (e.g. `Alpine`) provides, otherwise the compliation fails.

docs/Developer-Guide.md
ManaSugi: I added the descriptions

src/agent/Makefile
ManaSugi: I removed these changes.

docs/Developer-Guide.md
ManaSugi: Thank you for the suggestions.

> If you set LIBSECCOMP_LINK_TYPE to static, LIBSECCOMP_LIB_PATH must be set or the build with fail.

I added this to the docs.

> If you set LIBSECCOMP_LIB_PATH explicitly, then it must point to a libseccomp library built in your build environment, and LIBSECCOMP_LINK_TYPE will default to static if not set.

I think this seems to have an impact on the downstream because the [`libseccomp` crate](https://crates.io/crates/libseccomp) sets `LIBSECCOMP_LINK_TYPE` to `dylib` by default if users do not set it explicitly.

> If you set neither variables and you have not set SECCOMP=no, then the agent will assume that libseccomp is available dynamically, and report an error otherwise.

I added the description to the docs.

fidencio:

jodh-intel:
Any update on this @ManaSugi ?

arronwy:

ManaSugi:
@c3d @fidencio @jodh-intel  @liubin 

I do appologize for the late update. I finished this task. Could you review this again please?

Thank you.

docs/Developer-Guide.md
```diff
@@ -210,7 +210,8 @@ $ sudo systemctl restart systemd-journald
 
 > **Note:**
 >
-> - You should only do this step if you are testing with the latest version of the agent.
+> - You should only do this step if you are testing with the latest version of the agent or you want to build
+>   a custom agent for seccomp.
```
liubin, line 214: Is this line needed? SECCOMP is yes by default.

liubin: Or change `for seccomp` to `without seccomp`?

docs/Developer-Guide.md
```diff
@@ -228,16 +229,70 @@ $ go get -d -u github.com/kata-containers/kata-containers
...
+You **MUST** install the `libseccomp` library in your build environment and need to set the environment
+variables for the [`libseccomp` crate](https://github.com/libseccomp-rs/libseccomp-rs) as necessary to
+link the agent binary with a static `libseccomp` library.
+
+#### Link the `libseccomp` library statically against `musl`
+
+Install a static `libseccomp` library for `musl`:
+
+```sh
+$ sudo $GOPATH/src/github.com/kata-containers/kata-containers/ci/install_libseccomp.sh /tmp/kata-libseccomp /tmp/kata-gperf
```
liubin, line 244: Suggest changing the installation path to a directory that is not under `/tmp`, they will disappear after reboot.

liubin (approved):

liubin:
This question is not directly related to this PR:

Since in some environments, `LIBSECCOMP_LINK_TYPE=static` and `LIBSECCOMP_LIB_PATH` are needed, should the Makefile be updated to check them when the `SECCOMP=yes` to avoid the build error?

docs/Developer-Guide.md
ManaSugi: I adopted `a custom agent without seccomp feature`. Thank you.

docs/Developer-Guide.md
ManaSugi: I use `/usr/local/kata-libseccomp` and `/usr/local/kata-gperf` instead of `/tmp`.

ManaSugi:
@liubin Thank you for your review.

> Since in some environments, `LIBSECCOMP_LINK_TYPE=static` and `LIBSECCOMP_LIB_PATH` are needed, should the Makefile be updated to check them when the `SECCOMP=yes` to avoid the build error?

Yes, I also think that It would be better to check it in `Makefile`.
Example: 
If the user tries to build the agent with `LIBC=musl` and `SECCOMP=yes` but `LIBSECCOMP_LINK_TYPE` is not set to `static` and `LIBSECCOMP_LIB_PATH` is empty, the build will fail  with a kind error message.

Issue #3093, opened 2021-11-22
Agent `make clippy` failed


YchauWang:
# Description of problem

When I executed the `make clippy` under the agent project, the command ran failed. And showing the error:
```
root@david-vm:/workspaces/go/src/github.com/kata-containers/kata-containers/src/agent# make clippy
cargo clippy --all-targets --all-features --release \
	-- \
	-Aclippy::redundant_allocation \
	-D warnings
    Checking kata-agent v0.1.0 (/workspaces/go/src/github.com/kata-containers/kata-containers/src/agent)
error: unneeded `return` statement
   --> src/uevent.rs:243:25
    |
243 |                         return
    |                         ^^^^^^ help: remove `return`
    |
    = note: `-D clippy::needless-return` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return

error: could not compile `kata-agent` due to previous error
make: *** [Makefile:121: clippy] Error 101
```

# Expected result

Want to execute the `make` command correctly.

YchauWang:
This issue is I made a mistake. It was using rust 1.56.0 version produce the error msg. When I changed to 1.54.0, there is no error. The command executes correctly.

Cancel and close this issue now.

YchauWang:
For updating the rust version, Open this PR and issue now.

YchauWang:
https://github.com/kata-containers/kata-containers/pull/3125

PR #3095, opened 2021-11-22
agent: allow the clippy needless_return for the for_each block

YchauWang:
the clippy needless_return would break the `make clippy` command,
when the for_each block using the `return;`. This PR fixed this bug
by using `#[allow(clippy::needless_return)]`. Keep the `make clippy`
command executing correctly.

Fixes: #3093

### DIFF
--- a/src/agent/src/uevent.rs
+++ b/src/agent/src/uevent.rs
@@ -230,6 +230,7 @@ pub async fn watch_uevents(
 
 // Used in the device module unit tests
 #[cfg(test)]
+#[allow(clippy::needless_return)]
 pub(crate) fn spawn_test_watcher(sandbox: Arc<Mutex<Sandbox>>, uev: Uevent) {
     tokio::spawn(async move {
         loop {


src/agent/src/uevent.rs
```diff
@@ -230,6 +230,7 @@ pub async fn watch_uevents(
 
 // Used in the device module unit tests
 #[cfg(test)]
+#[allow(clippy::needless_return)]
```
bergwolf, line 233: Why not actually fix the needless return instead?

YchauWang: > 

I tried to remove the return of this test function. But could not keep the original code logic.

liubin: How about replacing `return` with `break`?

YchauWang: Thanks, but for_each block couldn't use break. I have to add `#[allow(clippy::needless_return)]`

liubin: @YchauWang What's your rust version? I can't get the same error as you.
And the usage of `for_each` is wrong here, we can use `filter` function to get the matched watcher to avoid using `for_each`.

YchauWang:
```
root@david-vm:~/devspaces/k3s-build-images# rustc -V
rustc 1.54.0 (a178d0322 2021-07-26)
```

YchauWang:
This issue is I made a mistake. It was using rust 1.56.0 version produce the error msg. When I changed to 1.54.0, there is no error. The command executes correctly.

liubin:
Hi @YchauWang this will be an issue when rust version is updated, so you can improve by using something replace `for_each`, for example:

```diff
--- a/src/agent/src/uevent.rs
+++ b/src/agent/src/uevent.rs
@@ -235,12 +235,11 @@ pub(crate) fn spawn_test_watcher(sandbox: Arc<Mutex<Sandbox>>, uev: Uevent) {
         loop {
             let mut sb = sandbox.lock().await;
             let uev = uev.clone();
-            sb.uevent_watchers.iter_mut().for_each(move |watch| {
+            let _ = sb.uevent_watchers.iter_mut().map(move |watch| {
                 if let Some((matcher, _)) = watch {
                     if matcher.is_match(&uev) {
                         let (_, sender) = watch.take().unwrap();
                         let _ = sender.send(uev.clone());
-                        return;
                     }
                 }
             });
```

I think you can reopen the issue and fix it although it has not happened yet.

So if you want you can reopen this.

YchauWang:
> 

Thanks @liubin !
For updating the rust version, Open this PR and issue now.

Jakob-Naucke:
@YchauWang can you apply the fix with eliminating the needless returns?

jodh-intel:
@YchauWang - any update on this?

YchauWang:
> @YchauWang can you apply the fix with eliminating the needless returns?

I’m very sorry. My recent work has focused on other projects, which resulted in no timely response.

YchauWang:
The issue has been solved by [PR #3125](https://github.com/kata-containers/kata-containers/pull/3125)

Issue #3059, opened 2021-11-17
Defer functions never run with os.Exit(0)


cmaf:
We use a `defer` function in `Shutdown` [here](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L930) and [here,](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L926) but they never run because of [os.Exit(0)](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L956). See https://pkg.go.dev/os#Exit. This could be fixed by moving the code to just before `os.Exit(0)`, and also before the [`return`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L942) in the `if statement`.

Bevisy:
> We use a `defer` function in `Shutdown` [here](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L930) and [here,](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L926) but they never run because of [os.Exit(0)](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L956). See https://pkg.go.dev/os#Exit. This could be fixed by moving the code to just before `os.Exit(0)`, and also before the [`return`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L942) in the `if statement`.

We can keep the original `defer func()` to take effect in the `if statement` and add the functions that would otherwise be executed in the `defer func()` before os.exit().

PR #3097, opened 2021-11-22
Fix defer funtions never run with os.Exit()

Bevisy:
os.Exit() will terminate program immediately, the defer functions
won't be executed, so we add defer functions again before os.Exit().
Refer to https://pkg.go.dev/os#Exit

FIxes: #3059


Bevisy:
It looks like a github bug was triggered and two identical pr's were created at the same time, closing this one.

Issue #3099, opened 2021-11-22
Create document giving code PR advice


jodh-intel:
We have a document that gives advice on [making documentation changes](https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md), so we should create one for making code changes.

PR #3100, opened 2021-11-22
docs: Add a code PR advice document

jodh-intel:
Add a document giving advice to code PR authors.

Fixes: #3099.

### DIFF
--- a/docs/Unit-Test-Advice.md
+++ b/docs/Unit-Test-Advice.md
@@ -129,7 +129,7 @@ pub type Result<T> = std::result::Result<T, &'static str>;
 // Accepts a string and an integer and returns the
 // result of sticking them together separated by a dash as a string.
 fn join_params_with_dash(str: &str, num: i32) -> Result<String> {
-    if str == "" {
+    if str.is_empty() {
         return Err("string cannot be blank");
     }
 
--- a/docs/Unit-Test-Advice.md
+++ b/docs/Unit-Test-Advice.md
@@ -1,19 +1,5 @@
 # Unit Test Advice
 
-* [Overview](#overview)
-* [Assertions](#assertions)
-    * [golang assertions](#golang-assertions)
-    * [rust assertions](#rust-assertions)
-* [Table driven tests](#table-driven-tests)
-    * [golang table driven tests](#golang-table-driven-tests)
-    * [rust table driven tests](#rust-table-driven-tests)
-* [Temporary files](#temporary-files)
-    * [golang temporary files](#golang-temporary-files)
-    * [rust temporary files](#rust-temporary-files)
-* [User running the test](#user-running-the-test)
-    * [running golang tests as different users](#running-golang-tests-as-different-users)
-    * [running rust tests as different users](#running-rust-tests-as-different-users)
-
 ## Overview
 
 This document offers advice on writing a Unit Test (UT) in
--- a/docs/Unit-Test-Advice.md
+++ b/docs/Unit-Test-Advice.md
@@ -1,0 +1,333 @@
+# Unit Test Advice
+
+* [Overview](#overview)
+* [Assertions](#assertions)
+    * [golang assertions](#golang-assertions)
+    * [rust assertions](#rust-assertions)
+* [Table driven tests](#table-driven-tests)
+    * [golang table driven tests](#golang-table-driven-tests)
+    * [rust table driven tests](#rust-table-driven-tests)
+* [Temporary files](#temporary-files)
+    * [golang temporary files](#golang-temporary-files)
+    * [rust temporary files](#rust-temporary-files)
+* [User running the test](#user-running-the-test)
+    * [running golang tests as different users](#running-golang-tests-as-different-users)
+    * [running rust tests as different users](#running-rust-tests-as-different-users)
+
+## Overview
+
+This document offers advice on writing a Unit Test (UT) in
+[`golang`](https://golang.org) and [`rust`](https://www.rust-lang.org).
+
+## Assertions
+
+### golang assertions
+
+Use the `testify` assertions package to create a new assertion object as this
+keeps the test code free from distracting `if` tests:
+
+```go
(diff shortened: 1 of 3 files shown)


fidencio:
@jodh-intel, I appreciate the document and you have my `lgtm`!
I am **not** approving it yet as I do want to let people go through this, rather than having it merged ASAP.

So, thumbs-up and virtual approval from my side.

jodh-intel:
@fidencio - Thanks - I agree with that approach. We don't want to "Robocop" this PR [1] so let's give folks a few days to digest and comment...

Conversely, silence is compliance... :smile:

---

[1] - *"You have 20 seconds to comply!"* :robot: :gun: :gun: :smile:

jodh-intel:
Note to reviewers: this doc is *not* a coding standard - it's a set of suggestions only.

We could do with some golang advice. Note to self for tomorrow: add golang suggestion to always create a constructor function: we've seen lots of bugs in golang code where code create a struct and "fills in **some** the blanks" leaving other in an invalid state (oops!) Rust doesn't have that problem thankfully.

cmaf (commented):
@jodh-intel, all of the unit test advice is helpful but I'm wondering if it is better suited for https://github.com/kata-containers/tests/blob/main/Unit-Test-Advice.md. I think we should definitely emphasize how important it is to write unit tests in this document, but it might be helpful to keep most of the unit test advice in one location.

jodh-intel:
@cmaf - Agreed. In fact I've moved the UT advice document across to this repo since, strictly, unit tests belong "here" whereas the `tests` repo is for other types of tests.

Once this PR lands, I'll remove https://github.com/kata-containers/tests/blob/main/Unit-Test-Advice.md.

jodh-intel:
Ping y'all - please could any passing developers tal at this... :smile:

jodh-intel:
@wainersm - this new document focuses on go+rust. wdyt about shell advice? I could add that into this document, but it *may* be better to have a separate document for that? Or is it even required? There is a lot that could be said about shell and it's an important language so I almost feel it deserves a separate doc if we want to provide shell scripting advice but open to opinions.

wainersm:
@jodh-intel I am going to review it still today :)

docs/Unit-Test-Advice.md
```diff
@@ -273,13 +273,13 @@ Some repositories already provide utility functions to skip a test:
 - if running as `root`
 - if not running as `root`
 
-### running golang tests as different users
+### running Golang tests as a different user
```
wainersm: Should read 'Running' (lead caps) instead of 'running'? Likewise below.

wainersm: Disregard this comment, you changed that line later.

cmaf (approved):
@jodh-intel lgtm

docs/code-pr-advice.md
```diff
@@ -0,0 +1,246 @@
...
+## Code
+
+### Architectures
+
+Do not write architecture-specific code if it is possible to write the
+code generically.
+
+### General advice
+
+- Do not write code to impress: instead write code that is easy to ready and understand.
```
bpradipt, line 33: nit: ready -> read

bpradipt:
Really a nice doc @jodh-intel and thanks for putting it together.

bpradipt (approved):
lgtm

docs/code-pr-advice.md
```diff
@@ -0,0 +1,246 @@
...
+by using more meaningful variable names).
+
+### Constants
+
+Don't embed magic numbers and strings in functions, particularly if
+they are used repeatedly.
+
+Create constants at the top of the file instead.
+
+### Copyright and license
```
wainersm, line 51: Can we have an example of copyright/license statement here?

docs/code-pr-advice.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Names
+
+Give functions, macros and variables clear and meaningful names.
+
+### Structures
+
+#### Golang structures
+
+Unlike Rust, Go does not enforce that all structure members be set.
```
wainersm, line 93: Do we prefer calling it Go or Golang?

docs/code-pr-advice.md
jodh-intel: Good catch @bpradipt! :smile: Updated.

docs/code-pr-advice.md
jodh-intel: Golang. Even though this is not the standard name, it's a common alias - see the commit message for c1111a1d2d7720cbf0087b384eeb29d5c3fee31c for more detail.

wainersm:
Hi @jodh-intel ,

> @wainersm - this new document focuses on go+rust. wdyt about shell advice? I could add that into this document, but it _may_ be better to have a separate document for that? Or is it even required? There is a lot that could be said about shell and it's an important language so I almost feel it deserves a separate doc if we want to provide shell scripting advice but open to opinions.

I agree advises for shell scripts are important as we have many on kata-containers (6.7%) and tests (60.1%) repositories. I like the idea of having a separate doc.

Things that could be covered:
 -  Quoting variables
 -  Set errexit / nounset /pipefail / errtrace
 -  Use die() and warn() from lib.sh
 - Send error messages to stderr
 - Structure the code around the pattern:
   ```
   var="bar"
   foo() {
   }
   main() {
     foo
   }
   main $@
   ```
 - Use `trap remove_tmp_files EXIT` pattern to clean up resources
 - Default values to variables subject to overwritten (`FOO="${FOO:-foo}"`)
 - Use `local` for local variables
 - Avoid complex logical operations (`foo || bar && foo1 && foo2 || echo "just to print hello World"`)
 - Recommended way to figure out the script's directory

There should be more 10^n (n > 1) things to advice :)

jodh-intel:
@wainersm - do you want to tal at adding such a doc? alternatively, we could create a wiki page, hack for a few days and then create a doc from that?

Issue #3124, opened 2021-11-25
update rust crate versions to the latest possible


jodh-intel:
Update the crate versions for all rust components to the latest available API compatible versions.

PR #3125, opened 2021-11-25
Update rust crate versions

jodh-intel:
Run `cargo update` to update to the latest crate dependency versions.

The agent is an application so this includes expanding the partially
specified semvers to full semver values for the following crates,
which makes those crates consistent with the other agent dependencies:

- `futures`
- `regex`
- `scan_fmt`
- `tokio`

Also includes:

- `cargo update` lockfiles updates for the `agent-ctl` tool and the `trace-forwarder`.
- An update to the latest `nix` crate version.

Fixes: #3124.

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -1388,13 +1388,13 @@ impl LinuxContainer {
         .context(format!("cannot change onwer of container {} root", id))?;
 
         if config.spec.is_none() {
-            return Err(nix::Error::EINVAL.into());
+            return Err(anyhow!(nix::Error::EINVAL));
         }
 
         let spec = config.spec.as_ref().unwrap();
 
         if spec.linux.is_none() {
-            return Err(nix::Error::EINVAL.into());
+            return Err(anyhow!(nix::Error::EINVAL));
         }
 
         let linux = spec.linux.as_ref().unwrap();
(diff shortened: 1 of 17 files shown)


jodh-intel:
Just a sec! More updates to come...

src/agent/rustjail/src/container.rs
```diff
@@ -1390,13 +1388,13 @@ impl LinuxContainer {
...
         if config.spec.is_none() {
-            return Err(nix::Error::Sys(Errno::EINVAL).into());
+            return Err(nix::Error::EINVAL.into());
         }
 
         let spec = config.spec.as_ref().unwrap();
 
         if spec.linux.is_none() {
-            return Err(nix::Error::Sys(Errno::EINVAL).into());
+            return Err(nix::Error::EINVAL.into());
```
liubin, line 1397: Great, the new version of `nix` also eliminates the problem of mixing `nix::Error::Sys()` and `nix::Error::from_errno()`.

@jodh-intel , do you think should we also unify how to return an `err`, now we have at least two styles(after your patch):

- `return Err(anyhow!(nix::Error::EINVAL));`
- `return Err(nix::Error::EINVAL.into());`

jodh-intel: Agreed. Which do you prefer?

liubin: If have to choose one, I prefer the  `return Err(anyhow!(nix::Error::EINVAL));`.

jodh-intel: @liubin - branch updated.

liubin (approved):

lifupan (approved):

PR #2150, opened 2021-06-29
osbuilder: Upgrade Ubuntu guest to 20.04

Jakob-Naucke:
- no need to create `/usr/lib/systemd/systemd` link any more
- install `chrony` as extra package and install extra packages in chroot
  rather than `debootstrap`, because `chrony` provides `time-daemon`,
  which under 20.04 is provided by `systemd-timesyncd`, which is
  required by `systemd`, and `debootstrap`'s conflict resolvement can't
  handle this, but `apt`'s can.

Fixes: #2147
> Ubuntu 18.04 didn't work with Podman, but Ubuntu 20.04 seems to work. Try that to see if Fedora CI can set up Ubuntu 20.04 through Podman instead.

Depends-on: https://github.com/kata-containers/tests/pull/3636
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

/cc @fidencio

PR #3127, opened 2021-11-25
ccv0: Support kata-deploy

Jakob-Naucke:
This is still a draft, but it is supposed to shine a little light on how we might deploy CCv0 along with the [operator payload image config](https://github.com/confidential-containers/operator/tree/ccv0-demo/demo/payload-image).
The image would be built with something like `SKOPEO_UMOCI=yes AA_KBC=offline_fs_kbc INCLUDE_ETC="$(realpath aa-offline_fs_kbc-keys.json) $(realpath kata-config.toml)" make` (files must be in repo so they're included in the build container).

@sameo do you think this is the way to go at all?
/cc @bpradipt

### DIFF
--- a/tools/packaging/guest-image/build_image.sh
+++ b/tools/packaging/guest-image/build_image.sh
@@ -26,13 +26,17 @@ build_initrd() {
 	info "Build initrd"
 	info "initrd os: $initrd_distro"
 	info "initrd os version: $initrd_os_version"
-	sudo -E PATH="$PATH" make initrd \
-		DISTRO="$initrd_distro" \
-		DEBUG="${DEBUG:-}" \
-		OS_VERSION="${initrd_os_version}" \
-		ROOTFS_BUILD_DEST="${builddir}/initrd-image" \
-		USE_DOCKER=1 \
-		AGENT_INIT="yes"
+	local rootfs_build_dest="${builddir}/initrd-image"
+	export DISTRO="$initrd_distro"
+	export OS_VERSION="${initrd_os_version}"
+	export USE_DOCKER=1
+	export AGENT_INIT="yes"
+	# ROOTFS_BUILD_DEST is a Make variable
+	sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}"
+	if [ -n "${INCLUDE_ROOTFS:-}" ]; then
+		sudo cp -RL --preserve=mode "${INCLUDE_ROOTFS}/." "${rootfs_build_dest}/${initrd_distro}_rootfs/"
+	fi
+	sudo -E PATH="$PATH" make initrd ROOTFS_BUILD_DEST="${rootfs_build_dest}"
 	mv "kata-containers-initrd.img" "${install_dir}/${initrd_name}"
 	(
 		cd "${install_dir}"
@@ -44,6 +48,9 @@ build_image() {
 	info "Build image"
 	info "image os: $img_distro"
 	info "image os version: $img_os_version"
+	# CCv0 on image is currently unsupported, do not pass
+	unset SKOPEO_UMOCI
+	unset AA_KBC
 	sudo -E PATH="${PATH}" make image \
 		DISTRO="${img_distro}" \
 		DEBUG="${DEBUG:-}" \
--- a/versions.yaml
+++ b/versions.yaml
@@ -138,8 +138,8 @@ assets:
     url: "https://github.com/kata-containers/kata-containers/tools/osbuilder"
     architecture:
       aarch64:
-        name: &default-initrd-name "alpine"
-        version: &default-initrd-version "3.13.5"
+        name: &default-initrd-name "ubuntu"
+        version: &default-initrd-version "20.04"
       ppc64le:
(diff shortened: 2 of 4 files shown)


Jakob-Naucke:
Removing workflow for image -- it's read-only and CCv0 won't work.

tools/packaging/guest-image/build_image.sh
```diff
@@ -26,13 +26,23 @@ build_initrd() {
 	info "Build initrd"
 	info "initrd os: $initrd_distro"
 	info "initrd os version: $initrd_os_version"
-	sudo -E PATH="$PATH" make initrd \
+	rootfs_build_dest="${builddir}/initrd-image"
+	makeopts='
```
jodh-intel, line 30: Nit: These should really be `local`.

tools/packaging/guest-image/build_image.sh
```diff
@@ -26,13 +26,23 @@ build_initrd() {
...
 		USE_DOCKER=1 \
-		AGENT_INIT="yes"
+		AGENT_INIT="yes" \
+		SKOPEO_UMOCI="${SKOPEO_UMOCI:-}" \
+		AA_KBC="${AA_KBC:-}"
+
+	'
+	eval sudo -E PATH="$PATH" make rootfs $makeopts
+	if [ -n "${INCLUDE_ETC:-}" ]; then
+		sudo cp ${INCLUDE_ETC} "${rootfs_build_dest}/${initrd_distro}_rootfs/etc/"
```
jodh-intel, line 43: Are perms and ownership important (aka do we need to use `install(1)` here)?

versions.yaml
```diff
@@ -126,7 +126,7 @@ assets:
         name: "ubuntu"
         version: "latest"
       x86_64:
-        name: &default-image-name "clearlinux"
+        name: &default-image-name "ubuntu"
```
jodh-intel, line 129: I looked at https://github.com/kata-containers/kata-containers/pull/2908 but didn't see mention of clearlinux there? The rationale to switch to Ubuntu seems to be to make building golang apps easier but clearlinux tends to be updated daily (sometimes more so) so I'd be surprised if it didn't come with the latest possible version of golang ;) It's also smaller than Ubuntu.

jodh-intel (commented):
Thanks @Jakob-Naucke - a few comments.

fidencio (commented):
@Jakob-Naucke lgtm, thanks!  I'll approve after you reply to @jodh-intel comments.

tools/packaging/guest-image/build_image.sh
Jakob-Naucke: Ah, yes. Updated.

tools/packaging/guest-image/build_image.sh
Jakob-Naucke: Good call; yes, we do. I can't say I'm very happy with this solution (you can specify multiple space-separated files), which is the main reason this is a draft. Would you know of a more elegant (low-code :)) solution? Maybe even just using a folder is nicer.

versions.yaml
Jakob-Naucke: Ah. I've removed it again because this was only for image (not initrd, and only initrd will work ATM) anyway.
To answer the question still: We use Fedora, not Clear Linux, as a build container for Clear Linux. (On architectures that cannot use musl with Rust, this could lead to glibc mismatches, but Clear Linux is not supported on those anyway.) I excluded Clear Linux from #2908 because it would require an update to the Fedora build container and historically, we experienced too much pain on those upgrades (cf. #2128, #2150) to make it worthwile for a solution that is temporary anyhow for we will move to fully rustified image pulls (cf. #3035).

tools/packaging/guest-image/build_image.sh
jodh-intel: `cp -aL` might be sufficient then?

Jakob-Naucke: Hm, on second thought I think we should preserve mode (no use cases known to me right now, but who knows), but ownership should be `root` (as everything in that rootfs), i.e. `sudo cp -LR --preserve=mode ${INCLUDE_ETC}/* …`. Limiting to `/etc` might be too inflexible, I'm growing more fond of having a "merge rootfs" instead, so you could have a folder with `etc`, `var`, `root`, whatever and that would just get merged.

Jakob-Naucke:
and now it would be `SKOPEO_UMOCI=yes AA_KBC=offline_fs_kbc INCLUDE_ROOTFSC="$(realpath include_rootfs)" make` with sth. like
```
include_rootfs
└── etc
    ├── aa-offline_fs_kbc-keys.json
    └── kata-config.toml
```

not tested yet though

sameo:
> @sameo do you think this is the way to go at all? /cc @bpradipt

I think so, yes. Thanks a lot @Jakob-Naucke !

sameo (approved):

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

Jakob-Naucke:
Thank you for your reviews! It is tested now. Any additional comments? /test

Issue #3122, opened 2021-11-25
Some agent tests can run by normal users


liubin:
Some tests in `sandbox.rs` need root user to run, actually this is a limit that shouldn't be there, like `add_and_get_container()`:

https://github.com/kata-containers/kata-containers/blob/86b5bb58014a49d387ac730d9195d71ec1e7e70c/src/agent/src/sandbox.rs#L723-L731

This is because the test container is using a system directory `/run/agent`, and the codes will try to crate the directory:

https://github.com/kata-containers/kata-containers/blob/86b5bb58014a49d387ac730d9195d71ec1e7e70c/src/agent/src/sandbox.rs#L687-L695

By using a temporary directory as the root of a container, like what the rustjail is doing, these tests can run undernormal users.

PR #3129, opened 2021-11-26
agent: use temp directory for test containers

liubin:
Some tests in sandbox.rs need root user to run, because they need create
directories under /run/agent directories, actually this is a limit
that shouldn't be there. By using a temp directory for test containers
will not need run tests as root user.

Fixes: #3122


jodh-intel (approved):
Nice! Thanks @liubin !

fidencio (approved):
lgtm, thanks @liubin!

Issue #3130, opened 2021-11-26
actions: Limit who can call `/test_kata_deploy`


fidencio:
`/test_kata_deploy` is an action we have that generates the kata-deploy binaries, generates a new quay.io image with them, and then test them out on an AKS cluster.

This is a time consuming and slightly expensive operation that, right now, can be triggered by anyone regardless of being part of the organization or not.

Although we never had any issues with outisders abusing this, better be safe than sorry and limit this to org members.

> **Note:** This was noticed by @jcvenegas, kudos to him!

PR #3134, opened 2021-11-26
workflows: only allow org members to run `/test_kata_deploy`

fidencio:
Let's take advantage of the "is-organization-member" action and only
allow members who are part of the `kata-containers` organization to
trigger `/test_kata_deploy`.

One caveat with this approach is that for the user to be considered as
part of an organization, they **must** have their "Organization
Visibility" configured as Public (and I think the default is Private).

This was found out and suggested by @jcvenegas!

Fixes: #3130

### DIFF
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -5,13 +5,38 @@ on:
 name: test-kata-deploy
 
 jobs:
-  build-asset:
+  check-comment-and-membership:
+    runs-on: ubuntu-latest
     if: |
       github.event.issue.pull_request
       && github.event_name == 'issue_comment'
       && github.event.action == 'created'
       && startsWith(github.event.comment.body, '/test_kata_deploy')
+    steps:
+      - name: Check membership
+        uses: kata-containers/is-organization-member@1.0.1
+        id: is_organization_member
+        with:
+          organization: kata-containers
+          username: ${{ github.event.comment.user.login }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Fail if not member
+        run: |
+          result=${{ steps.is_organization_member.outputs.result }}
+          if [ $result == false ]; then
+              user=${{ github.event.comment.user.login }}
+              echo Either ${user} is not part of the kata-containers organization
+              echo or ${user} has its Organization Visibility set to Private at
+              echo https://github.com/orgs/kata-containers/people?query=${user}
+              echo 
+              echo Ensure you change your Organization Visibility to Public and
+              echo trigger the test again.
+              exit 1
+          fi
+
+  build-asset:
     runs-on: ubuntu-latest
+    needs: check-comment-and-membership
     strategy:
       matrix:
         asset:


fidencio:
Typos apart I have tested this on my local repo, with the help of @elianerpeira.
Here's the two runs that matter for the context of this PR:
* from a member of the organization:
  * https://github.com/fidencio/kata-containers/runs/4332979940?check_suite_focus=true
* form a non-member of the organization:
  * https://github.com/fidencio/kata-containers/runs/4333004002?check_suite_focus=true

fidencio:
Last but not least, the reason I do the check on `run` instead of using `if` is because I couldn't get it working using `if`, and I'd prefer not spend much time on this, considering this solution may be "good enough".

.github/workflows/kata-deploy-test.yaml
```diff
@@ -5,13 +5,36 @@ on:
...
+          organization: kata-containers
+          username: ${{ github.event.comment.user.login }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Fail if not member
+        run: |
+          result=${{ steps.is_organization_member.outputs.result }}
+          if [ $result == false ]; then
+              user=${{ github.event.comment.user.login }}
+              echo ${user} is not part of the kata-containers organization!
+              echo If the ${user} should be part of the organiation, please,
```
jcvenegas, line 29: typo: s/organiation/organization/

.github/workflows/kata-deploy-test.yaml
```diff
@@ -5,13 +5,36 @@ on:
...
+  check-comment-and-membership:
+    runs-on: ubuntu-latest
     if: |
       github.event.issue.pull_request
       && github.event_name == 'issue_comment'
       && github.event.action == 'created'
       && startsWith(github.event.comment.body, '/test_kata_deploy')
+    steps:
+      - name: Check membership
+        uses: JamesSingleton/is-organization-member@1.0.1
```
jcvenegas, line 17: Sorry about the distrust for third party actions, but I have already reviewed the action and it looks good to me but as it interacts with the github token. I prefer to fork that action on our org and upgrade the branch if needed in the future.

fidencio: @jcvenegas, I forked the repo onto kata-containers org and now we're using it from there.
This makes sense, mainly considering how non-restrictive is our actions configurations, which is something we should re-evaluate at some point soon.

.github/workflows/kata-deploy-test.yaml
fidencio: I reworded this bit.  Please, give it a second try when time allows.

jcvenegas (approved):

.github/workflows/kata-deploy-test.yaml
jodh-intel: Thanks @jcvenegas. Talking in general terms but if we've reviewed the code, in theory we could just specify the git commit for the version we want to use since that cannot change (without us noticing), whereas the numeric (tag) version _could_ change. But this coment applies to all GitHub actions versions so isn't specific to this PR.

jodh-intel (approved):
Thanks @fidencio.

wainersm:
> Typos apart I have tested this on my local repo, with the help of @elianerpeira. Here's the two runs that matter for the context of this PR:
> 
>     * from a member of the organization:
>       
>       * https://github.com/fidencio/kata-containers/runs/4332979940?check_suite_focus=true
> 
>     * form a non-member of the organization:
>       
>       * https://github.com/fidencio/kata-containers/runs/4333004002?check_suite_focus=true

Did you test making any comment other than "/test_kata_deploy"? (Just to check it doesn't break "normal" comments). In theory it should work out because "check-comment-and-membership" won't run and "build-assets" needs the later; but because I don't know much about github actions...don't know if there a catch here...better ask :)

wainersm (approved):
LGTM. Thanks @fidencio !

fidencio:
> > Typos apart I have tested this on my local repo, with the help of @elianerpeira. Here's the two runs that matter for the context of this PR:
> > ```
> > * from a member of the organization:
> >   
> >   * https://github.com/fidencio/kata-containers/runs/4332979940?check_suite_focus=true
> > 
> > * form a non-member of the organization:
> >   
> >   * https://github.com/fidencio/kata-containers/runs/4333004002?check_suite_focus=true
> > ```
> 
> Did you test making any comment other than "/test_kata_deploy"? (Just to check it doesn't break "normal" comments). In theory it should work out because "check-comment-and-membership" won't run and "build-assets" needs the later; but because I don't know much about github actions...don't know if there a catch here...better ask :)

I did test running other comments and that's not triggered.  That specific check was not changed. :-)

Issue #3101, opened 2021-11-22
kata-deploy: Add back the checks for running `/test_kata_deploy`


fidencio:
Commit 3c9ae7fb4b93a33a8c2c9b39a99dac541bc859ee made `/test_kata_deploy` run
against HEAD, but it also mistakenly removed all the checks that ensure
`/test_kata_deploy` only runs when explicitly called.

Mea culpa on this, and let's add the tests back.

PR #3141, opened 2021-11-29
stable-2.3 | backport kata-deploy fixes / improvements

fidencio:
Let's bring to the stable-2.3 branches the following fixes:
* kata-deploy: Ensure we test HEAD with `/test_kata_deploy (#3001)
  * workflows: Add back the checks for running test-kata-deploy (#3101)
* workflows: only allow org members to run `/test_kata_deploy` (#3130)

### DIFF
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -6,6 +6,11 @@ name: test-kata-deploy
 
 jobs:
   build-asset:
+    if: |
+      github.event.issue.pull_request
+      && github.event_name == 'issue_comment'
+      && github.event.action == 'created'
+      && startsWith(github.event.comment.body, '/test_kata_deploy')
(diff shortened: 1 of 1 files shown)


bergwolf (approved):

c3d (approved):
Approved.

I like the "organization only" test filtering. Did you run this by people who did not make their org affiliation public and make sure they don't have some good reason? (If so, we may need another method. I suspect that's not a problem).

gkurz (approved):

fidencio:
> 

I didn't, but that's an excellent point.  We should check that with the folks who are part of the release rota.

fidencio:
> I like the "organization only" test filtering. Did you run this by people who did not make their org affiliation public and make sure they don't have some good reason? (If so, we may need another method. I suspect that's not a problem).

I didn't, but that's an excellent point and we should check with the folks who are in the release rota for the coming releases.

PR #3143, opened 2021-11-29
stable-2.3 | osbuilder: add coreutils to guest rootfs

bergwolf:
So that the debug console is more useful. In the meantime, remove
iptables as it is not used by kata-agent any more.


fidencio (approved):
lgtm, thanks @bergwolf!

gkurz (approved):
I'd appreciate seeing the `cherry picked from commit XXX` string for easier review.

Anyway, LGTM.

Issue #3140, opened 2021-11-29
Nested data volumes using configmap, downwardsapi, secrets, emptydir are broken in Kata 2


skaegi:
# Description of problem

If a pod declares mountpoints that are nested or overlap that are using a mixture of configmaps, secrets, downward api, and emptydir we are seeing unexpected behaviour that result in CrashLoopBackOffs, RunContainerError, and/or very slow resolution. The same pod definitions will work correctly with a standard runc based runtime Kubernetes as well as Kata v1.

Here's a gist that includes testcases we're using -- https://gist.github.com/skaegi/fa67f00ed0fe34d91a6b2a699b0f9ba4

The above gist includes test-cases for vanilla k8s, kata-qemu runtimeclass, and the old untrusted annotation style. To try the runtime class variant you can use...
```
kubectl apply -f https://gist.githubusercontent.com/skaegi/fa67f00ed0fe34d91a6b2a699b0f9ba4/raw/870491963ca7a0c75605b0979b6e81cea8da8903/testcase-nested-kata-qemu.yaml
```

The suite above is more complete but here's a sample test...
```
apiVersion: v1
data:
  config: my config
kind: ConfigMap
metadata:
  name: config
---
apiVersion: v1
data:
  secret: c2VjcmV0Cg==
kind: Secret
metadata:
  name: moresecrets
type: Opaque
---
apiVersion: v1
kind: Pod
metadata:
  name: nested-config-secret
spec:
  containers:
  - args:
    - -c
    - ls /config/secrets
    command:
    - /bin/sh
    image: gcr.io/google_containers/busybox:latest
    name: busybox
    volumeMounts:
    - mountPath: /config
      name: config
    - mountPath: /config/secrets
      name: secrets
  restartPolicy: OnFailure
  volumes:
  - name: secrets
    secret:
      secretName: secrets
  - configMap:
      name: config
    name: config
 ```
 
 These tests all worked correctly in Kata v1 as well as vanilla K8s but do not work in Kata 2.2.1 or 2.2.3. 

# Expected result

These pods should resolve and the containers complete successfully quickly.
```
$ k get po
NAME                            READY   STATUS      RESTARTS   AGE
nested-config-config            0/1     Completed   0          19m
nested-config-config-revmount   0/1     Completed   0          19m
nested-config-down              0/1     Completed   0          19m
nested-config-empty             0/1     Completed   0          19m
nested-config-secret            0/1     Completed   0          19m
[...]

PR #3152, opened 2021-11-30
agent: copy empty directories for watchable-bind mounts

liubin:
In function `update_target`, if the updated source is a directory,
    we should create the corresponding directory.

Fixes: #3140

### DIFF
--- a/src/agent/src/watcher.rs
+++ b/src/agent/src/watcher.rs
@@ -109,11 +109,12 @@ impl Storage {
 
         // if we are creating a directory: just create it, nothing more to do
         if source_file_path.symlink_metadata()?.file_type().is_dir() {
-            fs::create_dir_all(source_file_path)
+            let dest_file_path = self.make_target_path(&source_file_path)?;
+
+            fs::create_dir_all(&dest_file_path)
                 .await
-                .with_context(|| {
-                    format!("Unable to mkdir all for {}", source_file_path.display())
-                })?
+                .with_context(|| format!("Unable to mkdir all for {}", dest_file_path.display()))?;
+            return Ok(());
         }
 
         // Assume we are dealing with either a file or a symlink now:
@@ -922,7 +923,7 @@ mod tests {
             .file_type()
             .is_symlink());
         assert_eq!(fs::read_link(&dst_symlink_file).unwrap(), src_file);
-        assert_eq!(fs::read_to_string(&dst_symlink_file).unwrap(), "foo")
+        assert_eq!(fs::read_to_string(&dst_symlink_file).unwrap(), "foo");
     }
 
     #[tokio::test]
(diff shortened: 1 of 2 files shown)


src/agent/src/watcher.rs
```diff
@@ -87,6 +87,8 @@ async fn copy(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<()> {
             fs::remove_file(&to).await?;
         }
         fs::symlink(fs::read_link(&from).await?, &to).await?;
+    } else if from.as_ref().is_dir() {
+        fs::create_dir_all(to).await?;
```
egernst, line 91: Can we add/augment a unit test to exercise this?

egernst:
Nice catch @liubin! I think we should add a unit test to exercise this path, otherwise looks good.

/test

jodh-intel:
> I think we should add a unit test

@egernst - dude you read my mind! ;)

egernst:
related: https://github.com/kata-containers/tests/issues/4239

src/agent/src/watcher.rs
liubin: Yes, I'add.

egernst: Great thanks. Any reason to not (also?) add to test_copy?  https://github.com/kata-containers/kata-containers/pull/3152/files#diff-3f11f988ef87c954cf00b1d8a6977cf2cd3cc6aac9514d3496b3bafd37258dfcR898

egernst (approved):

src/agent/src/watcher.rs
liubin: @egernst There is no reason to do not to test `test_copy`, only because I missed it. Now added.

egernst (approved):

bergwolf (approved):
good catch!

liubin:
Hi, @egernst I found that maybe the `copy` is used for copying files only, the creating directories should reside in the `update_target` function. Currently, the creation operation creates the source directory, not the target directory. I think this is wrong.

I tried fix it by the second commit, please review it, and I will squash these two commits later.

egernst (approved):

skaegi:
Thanks again @liubin -- is there a way I can track a backport to 2.3.x as that's likely what we will integrate with.

liubin:
> Thanks again @liubin -- is there a way I can track a backport to 2.3.x as that's likely what we will integrate with.

@fidencio @bergwolf Can you help to answer this question? Are there will be bulk backport PRs?

fidencio:
@liubin, nops.

Could you raise a backport yourself of this PR?  Then the person who's cutting the release will just cut the release.

liubin:
@skaegi @fidencio I have created https://github.com/kata-containers/kata-containers/pull/3184 .

Issue #3185, opened 2021-12-03
Update node.k8s.io/v1beta1 to node.k8s.io/v1


liubin:
Related: https://github.com/kata-containers/tests/pull/4244

```bash
$ grep -rn "node.k8s.io/v1beta1" .
./tools/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml:3:apiVersion: node.k8s.io/v1beta1
./tools/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml:16:apiVersion: node.k8s.io/v1beta1
./tools/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml:29:apiVersion: node.k8s.io/v1beta1
./docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md:157:apiVersion: node.k8s.io/v1beta1
```

fidencio:
Please, whoever will work on this, pretty please, mind: https://kubernetes.io/releases/.

If something is deprecated in v1.21+, we can only drop it when v1.20 reaches its EOL (which should happen in 2022-02-08).
If something is deprecated in v1.22+, we can only drop it when v1.21 reaches its EOL (which should happen in 2022-06-28).

fidencio:
Is my understanding https://github.com/kata-containers/kata-containers/pull/3188#issuecomment-985498653 correct?
If so, what's the way to go about this?  I would not like to have something that can still work with a maintained version of k8s to stop working just because we want to use a new API (without providing a backwards compatible solution, if needed).

fidencio:
Okay, for RuntimeClass, from https://kubernetes.io/docs/concepts/containers/runtime-class/
```
Runtime Class
FEATURE STATE: Kubernetes v1.20 [stable]
```

PR #3188, opened 2021-12-03
update apiVersion

weiyuanke:
kata-deploy: Update API Version of RuntimeClass to v1
API Version of node.k8s.io/v1beta1 is deprecated in
v1.22+, unavailable in v1.25+

Fixes: #3185

### DIFF
--- a/docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
+++ b/docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
@@ -154,7 +154,7 @@ From Kubernetes v1.12, users can use [`RuntimeClass`](https://kubernetes.io/docs
 
 ```bash
 $ cat > runtime.yaml <<EOF
-apiVersion: node.k8s.io/v1beta1
+apiVersion: node.k8s.io/v1
 kind: RuntimeClass
 metadata:
   name: kata
--- a/tools/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml
+++ b/tools/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml
@@ -1,6 +1,6 @@
 ---
 kind: RuntimeClass
-apiVersion: node.k8s.io/v1beta1
+apiVersion: node.k8s.io/v1
 metadata:
     name: kata-qemu
 handler: kata-qemu
@@ -13,7 +13,7 @@ scheduling:
     katacontainers.io/kata-runtime: "true"
 ---
 kind: RuntimeClass
-apiVersion: node.k8s.io/v1beta1
+apiVersion: node.k8s.io/v1
 metadata:
     name: kata-clh
 handler: kata-clh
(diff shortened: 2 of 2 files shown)


liubin (approved):

lifupan (approved):

fidencio:
Wait, wait, let's take a step back here.
Would this work without issues a kubernetes v1.20 cluster?  I **think** it will not, but I may be totally wrong here.

IMHO, we should only do the switched after ensuring the oldest supported k8s can support "node.k8s.io/v1".  If that's the case, then "yay!", let's go ahead and merge this.  Otherwise I'd be slightly more careful about those changes and do then later in the cycle.

docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
```diff
@@ -19,7 +19,7 @@ The Kubernetes cluster will use the
 
 ## Install and configure containerd
 
-First, follow the [How to use Kata Containers and Containerd](containerd-kata.md) to install and configure containerd. 
+First, follow the [How to use Kata Containers and Containerd](containerd-kata.md) to install and configure containerd.
```
fidencio, line 22: This change is not related to the problem you're fixing, please, drop it.

docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
```diff
@@ -44,7 +44,7 @@ In order to allow Kubelet to use containerd (using the CRI interface), configure
   ```bash
   $ sudo mkdir -p  /etc/systemd/system/kubelet.service.d/
   $ cat << EOF | sudo tee  /etc/systemd/system/kubelet.service.d/0-containerd.conf
-  [Service]                                                 
+  [Service]
```
fidencio, line 47: This change is not related to the problem you're fixing, please, drop it as well.

docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
```diff
@@ -182,7 +182,7 @@ If a pod has the `runtimeClassName` set to `kata`, the CRI plugin runs the pod w
     containers:
     - name: nginx
       image: nginx
-      
+
```
fidencio, line 185: Same here about the additional line.  Please, drop it.

fidencio (changes requested):
In general LGTM.  Please, just follow the comments made inline and we can have it merged.

I also checked that RuntimeClass became stable on 1.20 release, so the change is good to go.

weiyuanke:
@fidencio PTAL

liubin:
@fidencio Can you take a look again?

jodh-intel (approved):
Thanks @weiyuanke.

lgtm

@fidencio - ptal.

jodh-intel:
Ping @fidencio.

fidencio:
Folks, please, feel free to send me a message on Slack about those PRs that get blocked and end up lost on my email.
Huge sorry from my side, let me take a look at it now.

fidencio (approved):
lgtm, thanks @weiyuanke!

PR #3189, opened 2021-12-03
kata-deploy: Update RuntimeClass' API Version to v1

Bevisy:
RuntimeClass' API Version of node.k8s.io/v1beta1 is
deprecated in v1.22+, unavailable in v1.25+

Fixes: #3185

### DIFF
--- a/docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
+++ b/docs/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
@@ -153,15 +153,15 @@ By default, all pods are created with the default runtime configured in CRI cont
 From Kubernetes v1.12, users can use [`RuntimeClass`](https://kubernetes.io/docs/concepts/containers/runtime-class/#runtime-class) to specify a different runtime for Pods.
 
 ```bash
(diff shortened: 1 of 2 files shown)


Bevisy:
It looks like there are different PRs working on the same issue fix, feel free to close this PR if needed. https://github.com/kata-containers/kata-containers/pull/3188

fidencio:
Same comment as in https://github.com/kata-containers/kata-containers/pull/3188#issuecomment-985498653
```
Wait, wait, let's take a step back here.
Would this work without issues a kubernetes v1.20 cluster?  I **think** it will not, but I may be totally wrong here.

IMHO, we should only do the switched after ensuring the oldest supported k8s can support "node.k8s.io/v1".  If that's the case, then "yay!", let's go ahead and merge this.  Otherwise I'd be slightly more careful about those changes and do then later in the cycle.
```

fidencio:
@Bevisy, although your patch is ready to be merged, I'll proceed reviewing https://github.com/kata-containers/kata-containers/pull/3188 as that one was opened first.

fidencio:
Ah, of course, thanks a lot for the contribution and for pointing us out to #3188.

Issue #3190, opened 2021-12-03
Update the architecture document


jodh-intel:
The [architecture doc](https://github.com/kata-containers/kata-containers/blob/main/docs/design/architecture.md) needs a general review and refresh to remove talk of the "2.0 release" and provide more details.

- Phase 1: do ^
- Phase 2: Get someone (else :) with good diagramming skills to update the diagrams
  (or rework them in ASCII art if no volunteers :smile:)

PR #3195, opened 2021-12-03
docs: Update architecture document

jodh-intel:
Refresh the content and formatting of the architecture document.

Out of scope of these changes:

- Diagram updates.
- Updates to the Networking section.

Fixes: #3190.

### DIFF
--- a/docs/design/architecture.md
+++ b/docs/design/architecture.md
@@ -2,127 +2,639 @@
 
 ## Overview
 
-This is an architectural overview of Kata Containers, based on the 2.0 release.
+Kata Containers is an open source community working to build a secure
+container [runtime](#runtime) with lightweight virtual machines (VM's)
+that feel and perform like standard Linux containers, but provide
+stronger [workload](#workload) isolation using hardware
+[virtualization](#virtualization) technology as a second layer of
+defence.
 
-The primary deliverable of the Kata Containers project is a CRI friendly shim. There is also a CRI friendly library API behind them.
+Kata Containers runs on [multiple architectures](../../src/runtime/README.md#platform-support)
+and supports [multiple hypervisors](../hypervisors.md).
 
-The [Kata Containers runtime](../../src/runtime)
-is compatible with the [OCI](https://github.com/opencontainers) [runtime specification](https://github.com/opencontainers/runtime-spec)
-and therefore works seamlessly with the [Kubernetes\* Container Runtime Interface (CRI)](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/container-runtime-interface.md)
-through the [CRI-O\*](https://github.com/kubernetes-incubator/cri-o) and
-[Containerd\*](https://github.com/containerd/containerd) implementation.
-
-Kata Containers creates a QEMU\*/KVM virtual machine for pod that `kubelet` (Kubernetes) creates respectively.
-
-The [`containerd-shim-kata-v2` (shown as `shimv2` from this point onwards)](../../src/runtime/cmd/containerd-shim-kata-v2/)
-is the Kata Containers entrypoint, which
-implements the [Containerd Runtime V2 (Shim API)](https://github.com/containerd/containerd/tree/master/runtime/v2) for Kata.
-
-Before `shimv2` (as done in [Kata Containers 1.x releases](https://github.com/kata-containers/runtime/releases)), we need to create a `containerd-shim` and a [`kata-shim`](https://github.com/kata-containers/shim) for each container and the Pod sandbox itself, plus an optional [`kata-proxy`](https://github.com/kata-containers/proxy) when VSOCK is not available. With `shimv2`, Kubernetes can launch Pod and OCI compatible containers with one shim (the `shimv2`) per Pod instead of `2N+1` shims, and no standalone `kata-proxy` process even if no VSOCK is available.
-
-![Kubernetes integration with shimv2](arch-images/shimv2.svg)
-
-The container process is then spawned by
-[`kata-agent`](../../src/agent), an agent process running
-as a daemon inside the virtual machine. `kata-agent` runs a [`ttRPC`](https://github.com/containerd/ttrpc-rust) server in
-the guest using a VIRTIO serial or VSOCK interface which QEMU exposes as a socket
-file on the host. `shimv2` uses a `ttRPC` protocol to communicate with
-the agent. This protocol allows the runtime to send container management
-commands to the agent. The protocol is also used to carry the I/O streams (stdout,
-stderr, stdin) between the containers and the manage engines (e.g. CRI-O or containerd).
-
-For any given container, both the init process and all potentially executed
-commands within that container, together with their related I/O streams, need
-to go through the VSOCK interface exported by QEMU.
-
-The container workload, that is, the actual OCI bundle rootfs, is exported from the
-host to the virtual machine.  In the case where a block-based graph driver is
-configured, `virtio-scsi` will be used. In all other cases a `virtio-fs` VIRTIO mount point
-will be used. `kata-agent` uses this mount point as the root filesystem for the
-container processes.
+This document is a summary of the Kata Containers architecture.
 
 ## Virtualization
 
-How Kata Containers maps container concepts to virtual machine technologies, and how this is realized in the multiple
-hypervisors and VMMs that Kata supports is described within the [virtualization documentation](./virtualization.md)
+For details on how Kata Containers maps container concepts to VM
+technologies, and how this is realized in the multiple hypervisors and
+VMMs that Kata supports see the
+[virtualization documentation](./virtualization.md).
+
+## Compatibility
+
+The [Kata Containers runtime](../../src/runtime) is compatible with
+the [OCI](https://github.com/opencontainers)
+[runtime specification](https://github.com/opencontainers/runtime-spec)
+and therefore works seamlessly with the
+[Kubernetes Container Runtime Interface (CRI)](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/container-runtime-interface.md)
+through the [CRI-O](https://github.com/kubernetes-incubator/cri-o)
+and [containerd](https://github.com/containerd/containerd)
+implementations.
+
+Kata Containers provides a ["shimv2"](#shim-v2-architecture) compatible runtime.
+
+## Shim v2 architecture
+
+The Kata Containers runtime is shim v2 ("shimv2") compatible. This
+section explains what this means.
+
+### History
+
+In the old [Kata 1.x architecture](https://github.com/kata-containers/documentation/blob/master/design/architecture.md),
+the Kata [runtime](#runtime) was an executable called `kata-runtime`.
+The container manager called this executable multiple times when
+creating each container. Each time the runtime was called a different
+OCI command-line verb was provided. This architecture was simple, but
+not well suited to creating VM based containers due to the issue of
+handling state between calls. Additionally, the architecture suffered
+from performance issues related to continually having to spawn new
+instances of the runtime binary, and
+[Kata shim](https://github.com/kata-containers/shim) and
+[Kata proxy](https://github.com/kata-containers/proxy) processes for systems
+that did not provide VSOCK.
+
+### An improved architecture
+
+The
+[containerd runtime shimv2 architecture](https://github.com/containerd/containerd/tree/main/runtime/v2)
+or _shim API_ architecture resolves the issues with the old
+architecture by defining a set of shimv2 APIs that a compatible
+runtime implementation must supply. Rather than calling the runtime
+binary multiple times for each new container, the shimv2 architecture
+runs a single instance of the runtime binary (for any number of
+containers). This improves performance and resolves the state handling
+issue.
+
+The shimv2 API is similar to the
+[OCI runtime](https://github.com/opencontainers/runtime-spec)
+API in terms of the way the container lifecycle is split into
+different verbs. Rather than calling the runtime multiple times, the
+container manager creates a socket and passes it to the shimv2
+runtime. The socket is a bi-directional communication channel that
+uses a gRPC based protocol to allow the container manager to send API
+calls to the runtime, which returns the result to the container
+manager using the same channel.
+
+The shimv2 architecture allows running several containers per VM to
+support container engines that require multiple containers running
+inside a pod.
+
+With the new architecture [Kubernetes](#kubernetes-support) can
+launch both Pod and OCI compatible containers with a single
+[runtime](#runtime) shim per Pod, rather than `2N+1` shims. No stand
+alone `kata-proxy` process is required, even if VSOCK is not
+available.
+
+### Architectural comparison
+
+| Kata version | Kata Runtime process calls | Kata shim processes | Kata proxy processes (if no VSOCK) |
+|-|-|-|-|
+| 1.x | multiple per container | 1 per container connection | 1 |
+| 2.x | 1 per VM (hosting any number of containers) | 0 | 0 |
+
+> **Notes:**
+>
+> - A single VM can host one or more containers.
+>
+> - The "Kata shim processes" column refers to the old
+>   [Kata shim](https://github.com/kata-containers/shim) (`kata-shim` binary),
+>   *not* the new shimv2 runtime instance (`containerd-shim-kata-v2` binary).
+
+The diagram below shows how the original architecture was simplified
+with the advent of shimv2.
+
+![Kubernetes integration with shimv2](arch-images/shimv2.svg)
+
+## Root filesystem
+
+This document uses the term _rootfs_ to refer to a root filesystem
+which is mounted as the top-level directory ("`/`") and often referred
+to as _slash_.
+
+It is important to understand this term since the overall system uses
+multiple different rootfs's (as explained in the
+[Environments](#environments) section.
+
+## Example command
+
+The following containerd command creates a container. It is referred
+to throughout this document to help explain various points:
+
+```bash
+$ sudo ctr run --runtime "io.containerd.kata.v2" --rm -t "quay.io/libpod/ubuntu:latest" foo sh
+```
+
+This command requests that containerd:
+
+- Create a container (`ctr run`).
+- Use the Kata [shimv2](#shim-v2-architecture) runtime (`--runtime "io.containerd.kata.v2"`).
+- Delete the container when it [exits](#workload-exit) (`--rm`).
+- Attach the container to the user's terminal (`-t`).
+- Use the Ubuntu Linux [container image](#container-image)
+  to create the container [rootfs](#root-filesystem) that will become
+  the [container environment](#environments)
+  (`quay.io/libpod/ubuntu:latest`).
+- Create the container with the name "`foo`".
+- Run the `sh(1)` command in the Ubuntu rootfs based container
+  environment.
+
+  The command specified here is referred to as the [workload](#workload).
+
+> **Note:**
+>
+> For the purposes of this document and to keep explanations
+> simpler, we assume the user is running this command in the
+> [host environment](#environments).
+
+## Container image
+
+In the [example command](#example-command) the user has specified the
+type of container they wish to run via the container image name:
+`ubuntu`. This image name corresponds to a _container image_ that can
+be used to create a container with an Ubuntu Linux environment. Hence,
+in our [example](#example-command), the `sh(1)` command will be run
+inside a container which has an Ubuntu rootfs.
+
+> **Note:**
+>
+> The term _container image_ is confusing since the image in question
+> is **not** a container: it is simply a set of files (_an image_)
+> that can be used to _create_ a container. The term _container
+> template_ would be more accurate but the term _container image_ is
+> commonly used so this document uses the standard term.
+
+For the purposes of this document, the most important part of the
+[example command line](#example-command) is the container image the
+user has requested. Normally, the container manager will _pull_
+(download) a container image from a remote site and store a copy
+locally. This local container image is used by the container manager
+to create an [OCI bundle](#oci-bundle) which will form the environment
+the container will run in. After creating the OCI bundle, the
+container manager launches a [runtime](#runtime) which will create the
+container using the provided OCI bundle.
+
+## OCI bundle
+
+To understand what follows, it is important to know at a high level
+how an OCI ([Open Containers Initiative](https://opencontainers.org)) compatible container is created.
+
+An OCI compatible container is created by taking a
+[container image](#container-image) and converting the embedded rootfs
+into an
+[OCI rootfs bundle](https://github.com/opencontainers/runtime-spec/blob/main/bundle.md),
+or more simply, an _OCI bundle_.
+
+An OCI bundle is a `tar(1)` archive normally created by a container
+manager which is passed to an OCI [runtime](#runtime) which converts
+it into a full container rootfs. The bundle contains two assets:
+
+- A container image [rootfs](#root-filesystem)
+
+  This is simply a directory of files that will be used to represent
+  the rootfs for the container.
+
+  For the [example command](#example-command), the directory will
+  contain the files necessary to create a minimal Ubuntu root
+  filesystem.
+
+- An [OCI configuration file](https://github.com/opencontainers/runtime-spec/blob/main/config.md)
+
+  This is a JSON file called `config.json`.
+
+  The container manager will create this file so that:
+
+  - The `root.path` value is set to the full path of the specified
+    container rootfs.
+
+    In [the example](#example-command) this value will be `ubuntu`.
+
+  - The `process.args` array specifies the list of commands the user
+    wishes to run. This is known as the [workload](#workload).
+
+    In [the example](#example-command) the workload is `sh(1)`.
+
+## Workload
+
+The workload is the command the user requested to run in the
+container and is specified in the [OCI bundle](#oci-bundle)'s
+configuration file.
+
+In our [example](#example-command), the workload is the `sh(1)` command.
+
+### Workload root filesystem
+
+For details of how the [runtime](#runtime) makes the
+[container image](#container-image) chosen by the user available to
+the workload process, see the
+[Container creation](#container-creation) and [storage](#storage) sections.
+
+Note that the workload is isolated from the [guest VM](#environments) environment by its
+surrounding [container environment](#environments). The guest VM
+environment where the container runs in is also isolated from the _outer_
+[host environment](#environments) where the container manager runs.
+
+## System overview
+
+### Environments
+
+The following terminology is used to describe the different or
+environments (or contexts) various processes run in. It is necessary
+to study this table closely to make sense of what follows:
+
+| Type | Name | Virtualized | Containerized | rootfs | Rootfs device type | Mount type | Description |
+|-|-|-|-|-|-|-|-|
+| Host | Host | no `[1]` | no | Host specific | Host specific | Host specific | The environment provided by a standard, physical non virtualized system. |
+| VM root | Guest VM | yes | no | rootfs inside the [guest image](#guest-image) | Hypervisor specific `[2]` | `ext4` | The first (or top) level VM environment created on a host system. |
+| VM container root | Container | yes | yes | rootfs type requested by user ([`ubuntu` in the example](#example-command)) | `kataShared` | [virtio FS](#virtio-fs) | The first (or top) level container environment created inside the VM. Based on the [OCI bundle](#oci-bundle). |
+
+**Key:**
+
+- `[1]`: For simplicity, this document assumes the host environment
+  runs on physical hardware.
+
+- `[2]`: See the [DAX](#dax) section.
+
+> **Notes:**
+>
+> - The word "root" is used to mean _top level_ here in a similar
+>   manner to the term [rootfs](#root-filesystem).
+>
+> - The term "first level" prefix used above is important since it implies
+>   that it is possible to create multi level systems. However, they do
+>   not form part of a standard Kata Containers environment so will not
+>   be considered in this document.
+
+The reasons for containerizing the [workload](#workload) inside the VM
+are:
+
+- Isolates the workload entirely from the VM environment.
+- Provides better isolation between containers in a [pod](#kubernetes-support).
+- Allows the workload to be managed and monitored through its cgroup
+  confinement.
+
+### Container creation
+
+The steps below show at a high level how a Kata Containers container is
+created using the containerd container manager:
+
+1. The user requests the creation of a container by running a command
+   like the [example command](#example-command).
+1. The container manager daemon runs a single instance of the Kata
+   [runtime](#runtime).
+1. The Kata runtime loads its [configuration file](#configuration).
+1. The container manager calls a set of shimv2 API functions on the runtime.
+1. The Kata runtime launches the configured [hypervisor](#hypervisor).
+1. The hypervisor creates and starts (_boots_) a VM using the
+   [guest assets](#guest-assets):
+
+   - The hypervisor [DAX](#dax) shares the [guest image](#guest-image)
+     into the VM to become the VM [rootfs](#root-filesystem) (mounted on a `/dev/pmem*` device),
+     which is known as the [VM root environment](#environments).
+   - The hypervisor mounts the [OCI bundle](#oci-bundle), using [virtio FS](#virtio-fs),
+     into a container specific directory inside the VM's rootfs.
+
+     This container specific directory will become the
+     [container rootfs](#environments), known as the
+     [container environment](#environments).
+
+1. The [agent](#agent) is started as part of the VM boot.
+
+1. The runtime calls the agent's `CreateSandbox` API to request the
+   agent create a container:
+
+   1. The agent creates a [container environment](#environments)
+      in the container specific directory that contains the [container rootfs](#environments).
+
+      The container environment hosts the [workload](#workload) in the
+      [container rootfs](#environments) directory.
+
+   1. The agent spawns the workload inside the container environment.
+
+   > **Notes:**
+   >
+   > - The container environment created by the agent is equivalent to
+   >   a container environment created by the
+   >   [`runc`](https://github.com/opencontainers/runc) OCI runtime;
+   >   Linux cgroups and namespaces are created inside the VM by the
+   >   [guest kernel](#guest-kernel) to isolate the workload from the
+   >   VM environment the container is created in. See the
+   >   [Environments](#environments) section for an explanation of why
+   >   this is done.
+   >
+   > - See the [guest image](#guest-image) section for details of
+   >   exactly how the agent is started.
+
+1. The container manager returns control of the container to the
+   user running the `ctr` command.
+
+> **Note:**
+>
+> At this point, the container is running and:
+>
+> - The [workload](#workload) process ([`sh(1)` in the example](#example-command))
+>   is running in the [container environment](#environments).
+> - The user is now able to interact with the workload
+>   (using the [`ctr` command in the example](#example-command)).
+> - The [agent](#agent), running inside the VM is monitoring the
+>   [workload](#workload) process.
+> - The [runtime](#runtime) is waiting for the agent's `WaitProcess` API
+>   call to complete.
+
+Further details of these steps are provided in the sections below.
+
+### Container shutdown
+
+There are two possible ways for the container environment to be
+terminated:
+
+- When the [workload](#workload) exits.
+
+  This is the standard, or _graceful_ shutdown method.
+
+- When the container manager forces the container to be deleted.
+
+#### Workload exit
+
+The [agent](#agent) will detect when the [workload](#workload) process
+exits, capture its exit status (see `wait(2)`) and return that value
+to the [runtime](#runtime) by specifying it as the response to the
+`WaitProcess` agent API call made by the [runtime](#runtime).
+
+The runtime then passes the value back to the container manager by the
+`Wait` [shimv2 API](#shim-v2-architecture) call.
+
+Once the workload has fully exited, the VM is no longer needed and the
+runtime cleans up the environment (which includes terminating the
+[hypervisor](#hypervisor) process).
+
+> **Note:**
+>
+> When [agent tracing is enabled](../tracing.md#agent-shutdown-behaviour),
+> the shutdown behaviour is different.
+
+#### Container manager requested shutdown
+
+If the container manager requests the container be deleted, the
+[runtime](#runtime) will signal the agent by sending it a
+`DestroySandbox` [ttRPC API](../../src/agent/protocols/protos/agent.proto) request.
 
 ## Guest assets
 
-The hypervisor will launch a virtual machine which includes a minimal guest kernel
-and a guest image.
+Kata Containers creates a VM in which to run one or more containers. It
+does this by launching a [hypervisor](#hypervisor) to create the VM.
+The hypervisor needs two assets for this task: a Linux kernel and a
+small root filesystem image to boot the VM.
 
 ### Guest kernel
 
-The guest kernel is passed to the hypervisor and used to boot the virtual
-machine. The default kernel provided in Kata Containers is highly optimized for
-kernel boot time and minimal memory footprint, providing only those services
-required by a container workload. This is based on a very current upstream Linux
-kernel.
+The [guest kernel](../../tools/packaging/kernel)
+is passed to the hypervisor and used to boot the VM.
+The default kernel provided in Kata Containers is highly optimized for
+kernel boot time and minimal memory footprint, providing only those
+services required by a container workload. It is based on the latest
+Linux LTS (Long Term Support) [kernel](https://www.kernel.org).
 
 ### Guest image
 
-Kata Containers supports both an `initrd` and `rootfs` based minimal guest image.
+The hypervisor uses an image file which provides a minimal root
+filesystem used by the guest kernel to boot the VM and host the Kata
+Container. Kata Containers supports both initrd and rootfs based
+minimal guest images. The [default packages](../install/) provide both
+an image and an initrd, both of which are created using the
+[`osbuilder`](../../tools/osbuilder) tool.
+
+> **Notes:**
+>
+> - Although initrd and rootfs based images are supported, not all
+>   [hypervisors](#hypervisor) support both types of image.
+>
+> - The guest image is *unrelated* to the image used in a container
+>   workload.
+>
+>   For example, if a user creates a container that runs a shell in a
+>   BusyBox image, they will run that shell in a BusyBox environment.
+>   However, the guest image running inside the VM that is used to
+>   *host* that BusyBox image could be running Clear Linux, Ubuntu,
+>   Fedora or any other distribution potentially.
+>
+>   The `osbuilder` tool provides
+>   [configurations for various common Linux distributions](../../tools/osbuilder/rootfs-builder)
+>   which can be built into either initrd or rootfs guest images.
+>
+> - If you are using a [packaged version of Kata
+>   Containers](../install), you can see image details by running the
+>   [`kata-collect-data.sh`](../../src/runtime/data/kata-collect-data.sh.in)
+>   script as `root` and looking at the "Image details" section of the
+>   output.
 
 #### Root filesystem image
 
-The default packaged root filesystem image, sometimes referred to as the "mini O/S", is a
-highly optimized container bootstrap system based on [Clear Linux](https://clearlinux.org/). It provides an extremely minimal environment and
-has a highly optimized boot path.
+The default packaged rootfs image, sometimes referred to as the _mini
+O/S_, is a highly optimized container bootstrap system.
 
-The only services running in the context of the mini O/S are the init daemon
-(`systemd`) and the [Agent](#agent). The real workload the user wishes to run
-is created using libcontainer, creating a container in the same manner that is done
-by `runc`.
+If this image type is [configured](#configuration), when the user runs
+the [example command](#example-command):
 
-For example, when `ctr run -ti ubuntu date` is run:
+- The [runtime](#runtime) will launch the configured [hypervisor](#hypervisor).
+- The hypervisor will boot the mini-OS image using the [guest kernel](#guest-kernel).
+- The kernel will start the init daemon as PID 1 (`systemd`) inside the VM root environment.
+- `systemd`, running inside the mini-OS context, will launch the [agent](#agent)
+  in the root context of the VM.
+- The agent will create a new container environment, setting its root
+  filesystem to that requested by the user (Ubuntu in [the example](#example-command)).
+- The agent will then execute the command (`sh(1)` in [the example](#example-command))
+  inside the new container.
 
-- The hypervisor will boot the mini-OS image using the guest kernel.
-- `systemd`, running inside the mini-OS context, will launch the `kata-agent` in
-  the same context.
-- The agent will create a new confined context to run the specified command in
-  (`date` in this example).
-- The agent will then execute the command (`date` in this example) inside this
-  new context, first setting the root filesystem to the expected Ubuntu\* root
-  filesystem.
+The table below summarises the default mini O/S showing the
+environments that are created, the services running in those
+environments (for all platforms) and the root filesystem used by
+each service:
+
+| Process | Environment | systemd service? | rootfs | User accessible | Notes |
+|-|-|-|-|-|-|
+| systemd | VM root | n/a | [VM guest image](#guest-image)| [debug console][debug-console] | The init daemon, running as PID 1 |
+| [Agent](#agent) | VM root | yes | [VM guest image](#guest-image)| [debug console][debug-console] | Runs as a systemd service |
+| `chronyd` | VM root | yes | [VM guest image](#guest-image)| [debug console][debug-console] | Used to synchronise the time with the host |
+| container workload (`sh(1)` in [the example](#example-command)) | VM container | no | User specified (Ubuntu in [the example](#example-command)) | [exec command](#exec-command) | Managed by the agent |
+
+See also the [process overview](#process-overview).
+
+> **Notes:**
+>
+> - The "User accessible" column shows how an administrator can access
+>   the environment.
+>
+> - The container workload is running inside a full container
+>   environment which itself is running within a VM environment.
+>
+> - See the [configuration files for the `osbuilder` tool](../../tools/osbuilder/rootfs-builder)
+>   for details of the default distribution for platforms other than
+>   Intel x86_64.
 
 #### Initrd image
 
-A compressed `cpio(1)` archive, created from a rootfs which is loaded into memory and used as part of the Linux startup process. During startup, the kernel unpacks it into a special instance of a `tmpfs` that becomes the initial root filesystem.
+The initrd image is a compressed `cpio(1)` archive, created from a
+rootfs which is loaded into memory and used as part of the Linux
+startup process. During startup, the kernel unpacks it into a special
+instance of a `tmpfs` mount that becomes the initial root filesystem.
 
-The only service running in the context of the initrd is the [Agent](#agent) as the init daemon. The real workload the user wishes to run is created using libcontainer, creating a container in the same manner that is done by `runc`.
+If this image type is [configured](#configuration), when the user runs
+the [example command](#example-command):
+
+- The [runtime](#runtime) will launch the configured [hypervisor](#hypervisor).
+- The hypervisor will boot the mini-OS image using the [guest kernel](#guest-kernel).
+- The kernel will start the init daemon as PID 1 (the [agent](#agent))
+  inside the VM root environment.
+- The [agent](#agent) will create a new container environment, setting its root
+  filesystem to that requested by the user (`ubuntu` in
+  [the example](#example-command)).
+- The agent will then execute the command (`sh(1)` in [the example](#example-command))
+  inside the new container.
+
+The table below summarises the default mini O/S showing the environments that are created,
+the processes running in those environments (for all platforms) and
+the root filesystem used by each service:
+
+| Process | Environment | rootfs | User accessible | Notes |
+|-|-|-|-|-|
+| [Agent](#agent) | VM root | [VM guest image](#guest-image) | [debug console][debug-console] | Runs as the init daemon (PID 1) |
+| container workload | VM container | User specified (Ubuntu in this example) | [exec command](#exec-command) | Managed by the agent |
+
+> **Notes:**
+>
+> - The "User accessible" column shows how an administrator can access
+>   the environment.
+>
+> - It is possible to use a standard init daemon such as systemd with
+>   an initrd image if this is desirable.
+
+See also the [process overview](#process-overview).
+
+#### Image summary
+
+| Image type | Default distro | Init daemon | Reason | Notes |
+|-|-|-|-|-|
+| [image](#root-filesystem-image) | [Clear Linux](https://clearlinux.org) (for x86_64 systems)| systemd | Minimal and highly optimized | systemd offers flexibility |
+| [initrd](#initrd-image) | [Alpine Linux](https://alpinelinux.org) | Kata [agent](#agent) (as no systemd support) | Security hardened and tiny C library |
+
+See also:
+
+- The [osbuilder](../../tools/osbuilder) tool
+
+  This is used to build all default image types.
+
+- The [versions database](../../versions.yaml)
+
+  The `default-image-name` and `default-initrd-name` options specify
+  the default distributions for each image type.
+
+## Hypervisor
+
+The [hypervisor](../hypervisors.md) specified in the
+[configuration file](#configuration) creates a VM to host the
+[agent](#agent) and the [workload](#workload) inside the
+[container environment](#environments).
+
+> **Note:**
+>
+> The hypervisor process runs inside an environment slightly different
+> to the host environment:
+>
+> - It is run in a different cgroup environment to the host.
+> - It is given a separate network namespace from the host.
+> - If the [OCI configuration specifies a SELinux label](https://github.com/opencontainers/runtime-spec/blob/main/config.md#linux-process),
+>   the hypervisor process will run with that label (*not* the workload running inside the hypervisor's VM).
 
 ## Agent
 
-[`kata-agent`](../../src/agent) is a process running in the guest as a supervisor for managing containers and processes running within those containers.
+The Kata Containers agent ([`kata-agent`](../../src/agent)), written
+in the [Rust programming language](https://www.rust-lang.org), is a
+long running process that runs inside the VM. It acts as the
+supervisor for managing the containers and the [workload](#workload)
+running within those containers. Only a single agent process is run
+for each VM created.
 
-For the 2.0 release, the `kata-agent` is rewritten in the [RUST programming language](https://www.rust-lang.org/) so that we can minimize its memory footprint while keeping the memory safety of the original GO version of [`kata-agent` used in Kata Container 1.x](https://github.com/kata-containers/agent). This memory footprint reduction is pretty impressive, from tens of megabytes down to less than 100 kilobytes, enabling Kata Containers in more use cases like functional computing and edge computing.
+### Agent communications protocol
 
-The `kata-agent` execution unit is the sandbox. A `kata-agent` sandbox is a container sandbox defined by a set of namespaces (NS, UTS, IPC and PID). `shimv2` can
-run several containers per VM to support container engines that require multiple
-containers running inside a pod.
+The agent communicates with the other Kata components (primarily the
+[runtime](#runtime)) using a
+[`ttRPC`](https://github.com/containerd/ttrpc-rust) based
+[protocol](../../src/agent/protocols/protos).
 
-`kata-agent` communicates with the other Kata components over `ttRPC`.
+> **Note:**
+>
+> If you wish to learn more about this protocol, a practical way to do
+> so is to experiment with the
+> [agent control tool](#agent-control-tool) on a test system.
+> This tool is for test and development purposes only and can send
+> arbitrary ttRPC agent API commands to the [agent](#agent).
 
 ## Runtime
 
-`containerd-shim-kata-v2` is a [containerd runtime shimv2](https://github.com/containerd/containerd/blob/v1.4.1/runtime/v2/README.md) implementation and is responsible for handling the `runtime v2 shim APIs`, which is similar to [the OCI runtime specification](https://github.com/opencontainers/runtime-spec) but simplifies the architecture by loading the runtime once and making RPC calls to handle the various container lifecycle commands. This refinement is an improvement on the OCI specification which requires the container manager call the runtime binary multiple times, at least once for each lifecycle command.
+The Kata Containers runtime (the [`containerd-shim-kata-v2`](../../src/runtime/cmd/containerd-shim-kata-v2
+) binary) is a [shimv2](#shim-v2-architecture) compatible runtime.
 
-`containerd-shim-kata-v2` heavily utilizes the
-[virtcontainers package](../../src/runtime/virtcontainers/), which provides a generic, runtime-specification agnostic, hardware-virtualized containers library.
+> **Note:**
+>
+> The Kata Containers runtime is sometimes referred to as the Kata
+> _shim_. Both terms are correct since the `containerd-shim-kata-v2`
+> is a container runtime, and that runtime implements the containerd
+> shim v2 API.
+
+The runtime makes heavy use of the [`virtcontainers`
+package](../../src/runtime/virtcontainers), which provides a generic,
+runtime-specification agnostic, hardware-virtualized containers
+library.
+
+The runtime is responsible for starting the [hypervisor](#hypervisor)
+and it's VM, and communicating with the [agent](#agent) using a
+[ttRPC based protocol](#agent-communications-protocol) over a VSOCK
+socket that provides a communications link between the VM and the
+host. 
+
+This protocol allows the runtime to send container management commands
+to the agent. The protocol is also used to carry the standard I/O
+streams (`stdout`, `stderr`, `stdin`) between the containers and
+container managers (such as CRI-O or containerd).
+
+## Utility program
+
+The `kata-runtime` binary is a utility program that provides
+administrative commands to manipulate and query a Kata Containers
+installation.
+
+> **Note:**
+>
+> In Kata 1.x, this program also acted as the main
+> [runtime](#runtime), but this is no longer required due to the
+> improved shimv2 architecture.
+
+### exec command
+
+The `exec` command allows an administrator or developer to enter the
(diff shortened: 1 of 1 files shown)


jodh-intel:
@fidencio, @sameo, @egernst, @gnawux, @bergwolf, @kata-containers/architecture-committee *et al* - ptal! :smile:

fidencio:
I will go through this later Today / Monday in the worst case scenario.

devimc (commented):
thanks @jodh-intel - I've left some questions and comments

docs/design/architecture.md
```diff
@@ -2,128 +2,354 @@
...
+architecture by defining as set of shimv2 APIs that a compatible
+runtime implementation must supply. Rather than calling the runtime
+binary multiple times for each new container, the shimv2 architecture
+runs a single instance of the runtime binary (for any number of
+containers). This improves performance and resolves the state handling
+issue.
+
+The shimv2 API is similar to the
+[OCI runtime](https://github.com/opencontainers/runtime-spec)
+
```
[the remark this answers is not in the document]
jodh-intel: I tend to put MD links in column 1 if they are very long, but this isn't. Fixed.

docs/design/architecture.md
```diff
@@ -2,128 +2,354 @@
...
-
-The [`containerd-shim-kata-v2` (shown as `shimv2` from this point onwards)](../../src/runtime/cmd/containerd-shim-kata-v2/)
-is the Kata Containers entrypoint, which
-implements the [Containerd Runtime V2 (Shim API)](https://github.com/containerd/containerd/tree/master/runtime/v2) for Kata.
+## Virtualization
 
-Before `shimv2` (as done in [Kata Containers 1.x releases](https://github.com/kata-containers/runtime/releases)), we need to create a `containerd-shim` and a [`kata-shim`](https://github.com/kata-containers/shim) for each container and the Pod sandbox itself, plus an optional [`kata-proxy`](https://github.com/kata-containers/proxy) when VSOCK is not available. With `shimv2`, Kubernetes can launch Pod and OCI compatible containers with one shim (the `shimv2`) per Pod instead of `2N+1` shims, and no standalone `kata-proxy` process even if no VSOCK is available.
+For details on how Kata Containers maps container concepts to VM
+technologies, and how this is realized in the multiple hypervisors and
+VMMs that Kata supports see the
```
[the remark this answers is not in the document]
jodh-intel: Yep - that was already there, but you're right. Fixed.

devimc (approved):

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
 
 ## Overview
 
-This is an architectural overview of Kata Containers, based on the 2.0 release.
+This document is a summary of the Kata Containers architecture.
```
eadamsintel, line 5: Consider combining these three sentences into one paragraph.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+
+In the old [Kata 1.x architecture](https://github.com/kata-containers/documentation/blob/master/design/architecture.md),
+the Kata [runtime](#runtime) was an executable called `kata-runtime`.
+The container manager called this executable multiple times when
+creating each container. Each time the runtime was called a different
+OCI command-line verb was provided. This architecture was simple, but
+not well suited to creating VM based containers due to the issue of
+handling state between calls. Additionally, the architecture suffered
+from performance issues related to continually having to spawn new
+instances of the runtime binary and the [Kata shim](https://github.com/kata-containers/shim)) and Kata
```
eadamsintel, line 49: I know this part wasn't in your review but this rendered incorrectly on GitHub. It showed as  "Kata shim) and Kata proxy](https://github.com/kata-containers/proxy) processes"

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+proxy](https://github.com/kata-containers/proxy) processes for systems
+that did not provide VSOCK.
+
+### An improved architecture
+
+The
+[containerd runtime shimv2 architecture](https://github.com/containerd/containerd/tree/master/runtime/v2)
+or "shim API" architecture resolves the issues with the old
+architecture by defining as set of shimv2 APIs that a compatible
```
eadamsintel, line 58: "as" should be "a"

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+runtime implementation must supply. Rather than calling the runtime
+binary multiple times for each new container, the shimv2 architecture
+runs a single instance of the runtime binary (for any number of
```
eadamsintel, line 61: Someone might wonder what happens if the long running shim binary crashes. If possible I would address that here.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+runtime. The socket is a bi-directional communication channel that
+uses a gRPC based protocol to allow the container manager to send API
+calls to the runtime, which returns the result to the container
+manager using the same channel.
+
+The shimv2 architecture allows running several containers per VM to
+support container engines that require multiple containers running
+inside a pod.
+
+With the new architecture [Kubernetes](#kubernetes-support)) can
```
eadamsintel, line 79: This rendered incorrectly on GitHub.  "With the new architecture Kubernetes) can launch "

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+launch Pod and OCI compatible containers with a single runtime shim
+(`containerd-shim-kata-v2`) per Pod instead of `2N+1` shims. No stand alone
+`kata-proxy` process is required, even if VSOCK is not available.
+
+### Architectural comparison
+
+| Kata version | Kata Runtime process calls | Kata shim processes | Kata proxy processes (if no VSOCK) |
+|-|-|-|-|
+| 1.x | multiple / container | 1 / container connection | 1 |
+| 2.x | 1 / VM | 0 | 0 |
```
eadamsintel, line 89: This is kind of a nit, but it looks the numbers are being divided.   I'd consider saying  "multiple per container"   "1 per VM"  Could you make the comparison "Multiple per VM" and "One per VM" instead?  My mind starts to wonder what the difference is between container and VM in this scenario.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+created using the containerd container manager:
+
+1. The user requests the creation of a container by running `ctr run`.
+1. The container manager daemon runs a single instance of the Kata [runtime](#runtime).
+1. The Kata runtime loads its [configuration file](#configuration).
+1. The container manager calls a set of shimv2 API functions on the runtime.
+1. The Kata runtime launches the configured [hypervisor](#hypervisor).
+1. The hypervisor creates a VM using the [guest assets](#guest-assets).
+1. The hypervisor boots (starts) the VM.
+1. The [agent](#agent) is launched inside the VM.
```
eadamsintel, line 116: Who launches the Agent?  Consider saying  "The agent is started as part of the VM boot"

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+1. The agent creates a container environment to host the [workload](#workload).
+1. The agent spawns the workload inside the container environment.
+1. Control is returned to the user running `ctr`.
```
eadamsintel, line 119: Which binary returns the control?  Since this is an arch document I think going into a little more detail at this point is good.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+
+   If the user requested it, the standard I/O streams of the workload
+   running inside the container in the VM are connected to the user's
+   terminal.
```
eadamsintel, line 123: Putting this into a single picture with numbers would make this much easier to understand. It might be a little challenging, but it could be similar to the Network hotplug picture further below.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+Further details of these steps are provided in the sections below.
+
+### Container shutdown
 
-How Kata Containers maps container concepts to virtual machine technologies, and how this is realized in the multiple
-hypervisors and VMMs that Kata supports is described within the [virtualization documentation](./virtualization.md)
+There are two possible ways for the container environment to be
+destroyed:
+
+- When the [workload](#workload) exits.
```
eadamsintel, line 132: It might be worth calling out which one is a graceful exit and which one isn't. It is kind of implied but when you state that there are two ways for it to be destroyed I can think of other ways like the computer rebooting.   Also, destroyed sounds kind of harsh. Perhaps say terminated instead.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+- When the container manager forces the container to be deleted.
+
+#### Workload exit
+
+The [agent](#agent) will detect when the [workload](#workload) process
+exits, capture its exit code and return it to the runtime, which
```
eadamsintel, line 138: I am a little confused on what exactly is returned to the runtime. Is the exit code returned?  I think so but what "it" refers to could refer to the workload process which doesn't make much sense.   Here is an alternative suggestion "The agent will detect when the workload process gracefully terminates, capture its exit code, and return that exit code to the runtime"

Is it even exit code that is returned or is it exit status?  Or is it actually the state information from the finished workload returned?  I don't quite understand this part if I really think about it.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+> **Note:**
+>
+> When [agent tracing is enabled](../tracing.md#agent-shutdown-behaviour),
+> the shutdown behaviour is different.
+
+#### Container manager requested shutdown
+
+If the container manager requests the container be deleted, the
+[runtime](#runtime) will signal the agent by sending it a
+[`DestroySandbox` ttRPC API](../../src/agent/protocols/protos/agent.proto) request.
```
eadamsintel, line 153: Since the link goes to the agent.proto file and not directly to the DestroySandbox line in it I would wrap the URL around "ttRPC API" only and not have the DestroySandbox text linkable.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
-The guest kernel is passed to the hypervisor and used to boot the virtual
-machine. The default kernel provided in Kata Containers is highly optimized for
-kernel boot time and minimal memory footprint, providing only those services
-required by a container workload. This is based on a very current upstream Linux
-kernel.
+The Linux [guest kernel](../../tools/packaging/kernel)
+is passed to the hypervisor and used to boot the VM.
+The default kernel provided in Kata Containers is highly optimized for
+kernel boot time and minimal memory footprint, providing only those
+services required by a container workload. This is based on a very
```
eadamsintel, line 167: Would it be accurate to say that Kata Containers is based on the latest LTS kernels? If so I think calling that out here makes sense instead of just saying  "very current upstream kernel"

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
 
 ### Guest image
 
-Kata Containers supports both an `initrd` and `rootfs` based minimal guest image.
+The hypervisor uses an image file which provides a minimal root
+filesystem used by the guest kernel to boot the VM and host the Kata Container.
+
+Kata Containers supports both initrd and rootfs based minimal guest images.
+
+The default images are created using the [`osbuilder`](../../tools/osbuilder) tool.
```
eadamsintel, line 177: I would consider combining these three separate sentences into a paragraph.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+boot path.
+
+The only services running in the context of the mini O/S are:
+
+| Process | Environment | systemd service? | Notes |
+|-|-|-|-|
+| systemd | VM root | n/a | The init daemon, running as PID 1 |
+| [Agent](#agent) | VM root | yes | Runs as a systemd service |
+| `chronyd` | VM root | yes | Used to synchronise the time with the host |
+| container workload | VM container | no | Managed by the agent |
```
eadamsintel, line 209: I would consider another column   "User accessible"  The Agent isn't directly accessible by the user but the container workload is.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
-(`systemd`) and the [Agent](#agent). The real workload the user wishes to run
-is created using libcontainer, creating a container in the same manner that is done
-by `runc`.
+> **Note:** The container workload is running inside a full container
+> environment which itself is running within a VM environment.
 
-For example, when `ctr run -ti ubuntu date` is run:
+For example, when the user runs `ctr run -ti ubuntu date`:
 
 - The hypervisor will boot the mini-OS image using the guest kernel.
```
eadamsintel, line 218: I know this wasn't part of your review but I think adding a step above that says "Kata will launch the configured hypervisor" before the hypervisor does the boot helps connect the 12 steps listed above with these condensed steps.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+> which can be built into either initrd or rootfs guest images.
 
 #### Root filesystem image
 
-The default packaged root filesystem image, sometimes referred to as the "mini O/S", is a
-highly optimized container bootstrap system based on [Clear Linux](https://clearlinux.org/). It provides an extremely minimal environment and
-has a highly optimized boot path.
+The default packaged root filesystem image ("rootfs"), sometimes
+referred to as the "mini O/S", is a highly optimized container
+bootstrap system based on [Clear Linux](https://clearlinux.org). It
```
eadamsintel, line 198: Is Clear Linux used as the default across all architectures?

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+For example, when the user runs `ctr run -ti ubuntu date`:
 
 - The hypervisor will boot the mini-OS image using the guest kernel.
 - `systemd`, running inside the mini-OS context, will launch the `kata-agent` in
   the same context.
 - The agent will create a new confined context to run the specified command in
   (`date` in this example).
 - The agent will then execute the command (`date` in this example) inside this
-  new context, first setting the root filesystem to the expected Ubuntu\* root
+  new context, first setting the root filesystem to the expected Ubuntu root
```
eadamsintel, line 224: This line left me confused.  It sounds like Ubuntu is the rootfs but isn't Ubuntu the container image?  It seems that rootfs is being used interchangeably for the VM rootfs and the containers file system which I don't know if that is actually a rootfs.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+rootfs which is loaded into memory and used as part of the Linux
+startup process. During startup, the kernel unpacks it into a special
+instance of a `tmpfs` mount that becomes the initial root filesystem.
+
+The only service running in the context of the initrd is the
+[Agent](#agent) as the init daemon. The real workload the user wishes
+to run is created using libcontainer, creating a container in the same
+manner that is done by `runc`.
+
+| Process | Environment | Notes |
```
eadamsintel, line 239: I don't really understand what environment means here.  The terms VM root and VM container don't make sense to me.  I think an extra column "User Accessible" or some other name to indicate what the user can interact with and what they can't would help.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+|-|-|-|
+| [Agent](#agent) | VM root | Runs as the init daemon (PID 1) |
+| container workload | VM container | Managed by the agent |
+
+See also the [process overview](#process-overview).
+
+## Hypervisor
 
-The only service running in the context of the initrd is the [Agent](#agent) as the init daemon. The real workload the user wishes to run is created using libcontainer, creating a container in the same manner that is done by `runc`.
+The [configured](#configuration) [hypervisor](../hypervisors.md) creates
```
eadamsintel, line 248: These two links render like it is one link. It isn't obvious each word links to something different.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+a VM to host the [agent](#agent) and the [workload](#workload).
+
+> **Note:**
+>
+> The hypervisor process run inside a separate "partial container"
```
eadamsintel, line 253: Change run to runs.

Also I really don't know what this is. Does this have something to do with the init container or is this something that runs on the host and uses container concepts like cgroups and namespaces?

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
-[`kata-agent`](../../src/agent) is a process running in the guest as a supervisor for managing containers and processes running within those containers.
+The Kata Containers agent ([`kata-agent`](../../src/agent)), written
+in the [Rust programming language](https://www.rust-lang.org), is a
+long running process that runs inside the VM. It acts as the
+supervisor for managing the containers and the [workload](#workload)
+running within those containers.
+
+> **Note:**
+>
+> Only a single agent process is run for each VM created.
```
eadamsintel, line 267: Does this really need to be a note that takes up a lot of space?  I would just add this statement to the paragraph above.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+
+### Agent communications protocol
+
+The agent communicates with the other Kata components using a
+[`ttRPC`](https://github.com/containerd/ttrpc-rust) based
+[protocol](../../src/agent/protocols/protos).
+
+See also the [agent control tool](#agent-control-tool).
```
eadamsintel, line 275: This tells users to go see the agent control tool but doesn't state what this is. I think you could combine this with the sentence above and add another sentence explaining what an agent control tool is exactly and then link to i.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+The workload is the OCI bundle rootfs. It is exported from the host
+into the VM.
 
-For the 2.0 release, the `kata-agent` is rewritten in the [RUST programming language](https://www.rust-lang.org/) so that we can minimize its memory footprint while keeping the memory safety of the original GO version of [`kata-agent` used in Kata Container 1.x](https://github.com/kata-containers/agent). This memory footprint reduction is pretty impressive, from tens of megabytes down to less than 100 kilobytes, enabling Kata Containers in more use cases like functional computing and edge computing.
+### Workload root filesystem
 
-The `kata-agent` execution unit is the sandbox. A `kata-agent` sandbox is a container sandbox defined by a set of namespaces (NS, UTS, IPC and PID). `shimv2` can
-run several containers per VM to support container engines that require multiple
-containers running inside a pod.
+For details of how the [runtime](#runtime) make the workload image
```
eadamsintel, line 284: "makes"

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+
+The agent communicates with the other Kata components using a
+[`ttRPC`](https://github.com/containerd/ttrpc-rust) based
+[protocol](../../src/agent/protocols/protos).
+
+See also the [agent control tool](#agent-control-tool).
+
+## Workload
+
+The workload is the OCI bundle rootfs. It is exported from the host
```
eadamsintel, line 279: I keep getting confused by how rootfs is used for the VM image and the workload image.  I think we need another name throughout the document to help readers keep it sorted.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
-run several containers per VM to support container engines that require multiple
-containers running inside a pod.
+For details of how the [runtime](#runtime) make the workload image
+chosen by the user (such as `busybox:latest`) available to the
+workload process, see the [storage](#storage) section.
 
-`kata-agent` communicates with the other Kata components over `ttRPC`.
+The container workload process defined by the OCI bundle's
+configuration file (`config.json`) is created using the Kata
+Containers [agent](#agent). The agent creates a container environment
```
eadamsintel, line 290: It would help to remind people this is in a VM.   "The agent inside the VM creates ..."

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+to run the workload in, in a similar manner to that done by the
+[`runc`](https://github.com/opencontainers/runc) OCI runtime, using
+Linux cgroups and namespaces.
```
eadamsintel, line 293: Does it create cgroups and namespaces on the host, in the VM, or both? I think explicitly calling that out would help.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
-`containerd-shim-kata-v2` is a [containerd runtime shimv2](https://github.com/containerd/containerd/blob/v1.4.1/runtime/v2/README.md) implementation and is responsible for handling the `runtime v2 shim APIs`, which is similar to [the OCI runtime specification](https://github.com/opencontainers/runtime-spec) but simplifies the architecture by loading the runtime once and making RPC calls to handle the various container lifecycle commands. This refinement is an improvement on the OCI specification which requires the container manager call the runtime binary multiple times, at least once for each lifecycle command.
+The Kata Containers runtime (or shim), is a [shimv2](#shim-v2-architecture) compatible runtime.
 
-`containerd-shim-kata-v2` heavily utilizes the
-[virtcontainers package](../../src/runtime/virtcontainers/), which provides a generic, runtime-specification agnostic, hardware-virtualized containers library.
+The name of the Kata Containers shim binary is
+[`containerd-shim-kata-v2`](../../src/runtime/cmd/containerd-shim-kata-v2).
 
-### Configuration
+The runtime makes heavy use of the [`virtcontainers`
```
eadamsintel, line 307: This is just a nit but there are 4 paragraphs in a row of just a single sentence.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+runtime-specification agnostic, hardware-virtualized containers
+library.
+
+The runtime is responsible for starting the [hypervisor](#hypervisor)
+and it's VM, and communicating with the [agent](#agent) using a
+[ttRPC based protocol](#agent-communications-protocol) over a VSOCK
+socket that provides a communications link between the VM and the
+host. 
+
+This protocol allows the runtime to send container management commands
```
eadamsintel, line 318: I think another sentence explaining how this communication into the VM is secure would help reinforce Kata as a security isolation product.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
 
-The runtime uses a TOML format configuration file called `configuration.toml`. By default this file is installed in the `/usr/share/defaults/kata-containers` directory and contains various settings such as the paths to the hypervisor, the guest kernel and the mini-OS image.
+## Utility program
 
-The actual configuration file paths can be determined by running:
-```
-$ kata-runtime --show-default-config-paths
-```
-Most users will not need to modify the configuration file.
+The `kata-runtime` binary is a utility program that provides
```
eadamsintel, line 325: One thing that would help is to explain how the kata-runtime is used in the shimv2 case and how it is used in the crio case.  I think that the kata-runtime binary is only for the SHIMv1 case with CRIO.  I think clearly separating out these two cases throughout the document would help a lot.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+### Configuration
+
+See the [configuration file details](../../src/runtime/README.md#configuration).
 
 The configuration file is also used to enable runtime [debug output](../Developer-Guide.md#enable-full-debug).
 
+## Process overview
+
+The table below shows an example of the main processes running when a
+Kata Container is created with containerd:
```
eadamsintel, line 344: What about CRIO?  I think the CRIO use case still needs a more robust explanation if it will be used for awhile still.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+
+| Description | Host | VM root context | VM container context |
```
eadamsintel, line 346: I mentioned this above but I think someone would be confused by the VM root context and VM container context. I think adding in what the user has access to and what they don't would help as well.

docs/design/architecture.md
```diff
@@ -135,8 +360,10 @@ In order to do so, container engines will usually add one end of a virtual
 ethernet (`veth`) pair into the container networking namespace. The other end of
```
eadamsintel, line 360: For consistency I would remove the parenthesis.

docs/design/architecture.md
```diff
@@ -156,7 +383,7 @@ on par with MACVTAP.
```
eadamsintel, line 383: I tried to leave a comment on the picture but I noticed it is referencing QEMU and not the other hypervisors. Maybe just leave it as KVM if you do decide to change the pic?  Also, didn't kata-shim and kata-proxy go away? This looks like an older diagram that is v1.x specific.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
 ![Network Hotplug](arch-images/kata-containers-network-hotplug.png)
 
 ## Storage
-Container workloads are shared with the virtualized environment through [virtio-fs](https://virtio-fs.gitlab.io/).
 
-The [devicemapper `snapshotter`](https://github.com/containerd/containerd/tree/master/snapshots/devmapper) is a special case. The `snapshotter` uses dedicated block devices rather than formatted filesystems, and operates at the block level rather than the file level. This knowledge is used to directly use the underlying block device instead of the overlay file system for the container root file system. The block device maps to the top read-write layer for the overlay. This approach gives much better I/O performance compared to using `virtio-fs` to share the container file system.
+### virtio SCSI
+
+If a block-based graph driver is configured,
+`virtio-scsi` is used to "share" the workload image (such as
```
eadamsintel, line 404: Does it share the workload, copy it, or give the VM exclusive access to block device range?  I think you could go really specific here if you wanted since it is an arch document.  I think going into detail here is good because people will wonder if this is secure or not. If it is shared it doesn't sound too secure.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
+
+### virtio FS
+
+In all other cases a [`virtio-fs`](https://virtio-fs.gitlab.io)
+(`VIRTIO`) mount point is used instead. The [agent](#agent) uses this
+mount point as the root filesystem for the container processes.
+
+For `virtio-fs`, the [runtime](#runtime) starts one `virtiofsd` daemon
+(that runs in the host context) for each VM created.
+
```
eadamsintel, line 415: What about 9pfs? Is it no longer supported at all?  If it is still supported I would mention it. If it isn't supported I would add a note it has been deprecated since version (not sure when it happened) of Kata.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
-Container workloads are shared with the virtualized environment through [virtio-fs](https://virtio-fs.gitlab.io/).
 
-The [devicemapper `snapshotter`](https://github.com/containerd/containerd/tree/master/snapshots/devmapper) is a special case. The `snapshotter` uses dedicated block devices rather than formatted filesystems, and operates at the block level rather than the file level. This knowledge is used to directly use the underlying block device instead of the overlay file system for the container root file system. The block device maps to the top read-write layer for the overlay. This approach gives much better I/O performance compared to using `virtio-fs` to share the container file system.
+### virtio SCSI
+
+If a block-based graph driver is configured,
+`virtio-scsi` is used to "share" the workload image (such as
+`busybox:latest`) into the container's environment inside the VM.
+
+### virtio FS
```
eadamsintel, line 407: Although inferred I would explicitly state this is an overlay filesystem or perhaps create a subsection for storage for overlay filesystems and devicemapper.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
 
-Kata Containers has the ability to hotplug and remove block devices, which makes it possible to use block devices for containers started after the VM has been launched.
+The
+[devicemapper `snapshotter`](https://github.com/containerd/containerd/tree/master/snapshots/devmapper)
+is a special case. The `snapshotter` uses dedicated block devices
+rather than formatted filesystems, and operates at the block level
+rather than the file level. This knowledge is used to directly use the
+underlying block device instead of the overlay file system for the
+container root file system. The block device maps to the top
+read-write layer for the overlay. This approach gives much better I/O
```
eadamsintel, line 425: Does it really give much better I/O performance? That sounds like we are recommending everyone to try devicemapper first.  Listing some of the pros and cons of each approach would help. That might be out of scope for an arch document though.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
 ![Network Hotplug](arch-images/kata-containers-network-hotplug.png)
 
 ## Storage
```
eadamsintel, line 399: One thing about storage with Kata that always confused me was which type of storage is being discussed.  This whole section seems to be about just the container image filesystem and how it is mounted, but what if you pass in a VolumeMount. Does this section apply to both the container image and any mounts?

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
+Kata Containers has the ability to hot plug add and hot plug remove
+block devices. This makes it possible to use block devices for
+containers started after the VM has been launched.
+
+Users can check to see if the container uses the `devicemapper` block
+device as its rootfs by calling `mount(8)` within the container.  If
+the `devicemapper` block device is used, the root filesystem (`/`)
+will be mounted from `/dev/vda`.  Users can disable direct mounting of
+the underlying block device through the runtime
+[configuration](#configuration).
```
eadamsintel, line 440: This goes along with my comment at the start. I am confused on how storage works between the container image that is started in the VM and any mounts that are done to the VM. This section seems to blend between the two.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
 
 ## Kubernetes support
 
-[Kubernetes\*](https://github.com/kubernetes/kubernetes/) is a popular open source
+[Kubernetes](https://github.com/kubernetes/kubernetes/), or K8s is a popular open source
```
eadamsintel, line 444: Need a comma after k8s

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
 only needs to run a container runtime and a container agent (called a
 [Kubelet](https://kubernetes.io/docs/admin/kubelet/)).
 
-A Kubernetes cluster runs a control plane where a scheduler (typically running on a
-dedicated master node) calls into a compute Kubelet. This Kubelet instance is
-responsible for managing the lifecycle of pods within the nodes and eventually relies
-on a container runtime to handle execution. The Kubelet architecture decouples
-lifecycle management from container execution through the dedicated
-`gRPC` based [Container Runtime Interface (CRI)](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/container-runtime-interface-v1.md).
+Kata Containers represents a Kubelet pod as a VM.
```
eadamsintel, line 453: Small nit but these are three individual sentences in a row each as their own paragraph.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
 ## Kubernetes support
 
-[Kubernetes\*](https://github.com/kubernetes/kubernetes/) is a popular open source
+[Kubernetes](https://github.com/kubernetes/kubernetes/), or K8s is a popular open source
 container orchestration engine. In Kubernetes, a set of containers sharing resources
 such as networking, storage, mount, PID, etc. is called a
-[Pod](https://kubernetes.io/docs/user-guide/pods/).
+[pod](https://kubernetes.io/docs/user-guide/pods/).
+
 A node can have multiple pods, but at a minimum, a node within a Kubernetes cluster
```
eadamsintel, line 449: It might be worth identifying what a node could be. In the case of a VM as a node mention that to use Kata you need nesting enabled.

docs/design/architecture.md
```diff
@@ -170,47 +397,89 @@ The following diagram illustrates the Kata Containers network hotplug workflow.
...
-lifecycle management from container execution through the dedicated
-`gRPC` based [Container Runtime Interface (CRI)](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/container-runtime-interface-v1.md).
+Kata Containers represents a Kubelet pod as a VM.
+
+A Kubernetes cluster runs a control plane where a scheduler (typically
+running on a dedicated master node) calls into a compute Kubelet. This
+Kubelet instance is responsible for managing the lifecycle of pods
+within the nodes and eventually relies on a container runtime to
+handle execution. The Kubelet architecture decouples lifecycle
+management from container execution through the dedicated gRPC based
```
eadamsintel, line 460: It sounds a little better to me to say "execution through a dedicated gRPC"

docs/design/architecture.md
```diff
@@ -221,48 +490,35 @@ or `container`. Kata Containers will then use this annotation to decide if it ne
...
+| `sandbox` | yes | yes (inside new VM) |
+| `container`| no | yes (in existing VM) |
 
 #### Mixing VM based and namespace based runtimes
 
 > **Note:** Since Kubernetes 1.12, the [`Kubernetes RuntimeClass`](https://kubernetes.io/docs/concepts/containers/runtime-class/)
 > has been supported and the user can specify runtime without the non-standardized annotations.
 
-With `RuntimeClass`, users can define Kata Containers as a `RuntimeClass` and then explicitly specify that a pod being created as a Kata Containers pod. For details, please refer to [How to use Kata Containers and Containerd](../../docs/how-to/containerd-kata.md).
+With `RuntimeClass`, users can define Kata Containers as a
```
eadamsintel, line 503: Does RuntimeClass work with CRI-O?  I would try to more clearly separate out how crio and containerd work in this section. They seem to be quite different.

docs/design/architecture.md
```diff
@@ -221,48 +490,35 @@ or `container`. Kata Containers will then use this annotation to decide if it ne
...
 # Appendices
 
 ## DAX
 
 Kata Containers utilizes the Linux kernel DAX [(Direct Access filesystem)](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.rst?h=v5.14)
 feature to efficiently map some host-side files into the guest VM space.
-In particular, Kata Containers uses the QEMU NVDIMM feature to provide a
-memory-mapped virtual device that can be used to DAX map the virtual machine's
+
+When configured to use the QEMU hypervisor Kata Containers uses QEMU's NVDIMM feature to provide a
```
eadamsintel, line 520: I think a comma in here somewhere to separate out the thought would make this a little easier to read. Perhaps put it after "QEMU hypervisor"

docs/design/architecture.md
```diff
@@ -221,48 +490,35 @@ or `container`. Kata Containers will then use this annotation to decide if it ne
...
+Kata Containers and Containerd](../../docs/how-to/containerd-kata.md).
 
+## Tracing
+
+The [tracing document](../tracing.md) provides details on the tracing
+architecture.
 
 # Appendices
 
 ## DAX
```
eadamsintel, line 515: Does this use pmem?  I often get really confused with this feature because it says nvdimm below but I think that is a construct vs an actual pmem device.  If this does not use hardware persistent memory and a linux construct I would explain that.

docs/design/architecture.md
eadamsintel: Also, what about pmem?  How is that case covered?  I believe it works differently and uses map_sync.  That would be worth explaining.

docs/design/architecture.md
```diff
@@ -279,12 +535,24 @@ file and device mapping mechanisms:
...
 
 ![DAX](arch-images/DAX.png)
 
-Information on the use of NVDIMM via QEMU is available in the [QEMU source code](http://git.qemu-project.org/?p=qemu.git;a=blob;f=docs/nvdimm.txt;hb=HEAD)
+Information on the use of NVDIMM via QEMU is available in the [QEMU
+source code](http://git.qemu-project.org/?p=qemu.git;a=blob;f=docs/nvdimm.txt;hb=HEAD)
+
+## Agent control tool
+
+The [agent control tool](../../tools/agent-ctl) is a test and
```
eadamsintel, line 552: Would people use this tool to validate their setup as well?

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+uses a gRPC based protocol to allow the container manager to send API
+calls to the runtime, which returns the result to the container
+manager using the same channel.
+
+The shimv2 architecture allows running several containers per VM to
+support container engines that require multiple containers running
+inside a pod.
+
+With the new architecture [Kubernetes](#kubernetes-support)) can
+launch Pod and OCI compatible containers with a single runtime shim
```
fidencio, line 80: As Eric didn't complain, it may be a language thing ... but "can launch Pod and OCi compatible containers" feels like missing something, at least for a non-native speaker.

Would "can launch both Pod and OCI compatible containers" be an acceptable suggestion, James?

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+
+If the container manager requests the container be deleted, the
+[runtime](#runtime) will signal the agent by sending it a
+[`DestroySandbox` ttRPC API](../../src/agent/protocols/protos/agent.proto) request.
 
 ## Guest assets
 
-The hypervisor will launch a virtual machine which includes a minimal guest kernel
-and a guest image.
+To create a VM to host a container, Kata Containers launches a
```
fidencio, line 157: "To create a VM to host a container" is not that natural to read, but I'm struggling to suggest something that could improve this.

Maybe "To create the VM where the container will be running"?  Not sure. :-/

docs/design/architecture.md
fidencio: If not, we need to update that kernel more frequently. :-)

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+services required by a container workload. This is based on a very
+current upstream Linux kernel.
 
 ### Guest image
 
-Kata Containers supports both an `initrd` and `rootfs` based minimal guest image.
+The hypervisor uses an image file which provides a minimal root
+filesystem used by the guest kernel to boot the VM and host the Kata Container.
+
+Kata Containers supports both initrd and rootfs based minimal guest images.
```
fidencio, line 175: I'd make a huge note there saying "depending on the hypervisor used", as I don't think we can use initrd with all the VMMs, neither I think we maintain tests with that (although Red Hat uses initrd for its product).

docs/design/architecture.md
fidencio: What are the default images? Where could the reader find those?

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
 #### Initrd image
 
-A compressed `cpio(1)` archive, created from a rootfs which is loaded into memory and used as part of the Linux startup process. During startup, the kernel unpacks it into a special instance of a `tmpfs` that becomes the initial root filesystem.
+The initrd image is a compressed `cpio(1)` archive, created from a
+rootfs which is loaded into memory and used as part of the Linux
+startup process. During startup, the kernel unpacks it into a special
+instance of a `tmpfs` mount that becomes the initial root filesystem.
+
+The only service running in the context of the initrd is the
+[Agent](#agent) as the init daemon. The real workload the user wishes
```
fidencio, line 235: We still support systemd as init, as that's what Red Hat is using on their product.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
 
-The only service running in the context of the initrd is the [Agent](#agent) as the init daemon. The real workload the user wishes to run is created using libcontainer, creating a container in the same manner that is done by `runc`.
+The [configured](#configuration) [hypervisor](../hypervisors.md) creates
+a VM to host the [agent](#agent) and the [workload](#workload).
+
+> **Note:**
+>
+> The hypervisor process run inside a separate "partial container"
+> environment comprised of cgroups and a separate network namespace
+> from the host.
```
fidencio, line 255: I'd also mention that the hypervisor can be SELinux confined, running under the container_kvm_t type (see: https://www.redhat.com/sysadmin/selinux-kata-containers)

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
-`kata-agent` communicates with the other Kata components over `ttRPC`.
+The container workload process defined by the OCI bundle's
+configuration file (`config.json`) is created using the Kata
+Containers [agent](#agent). The agent creates a container environment
+to run the workload in, in a similar manner to that done by the
+[`runc`](https://github.com/opencontainers/runc) OCI runtime, using
+Linux cgroups and namespaces.
+
+Note that the workload is isolated from the VM root environment by its
+surrounding container environment. The VM environment the workload
```
fidencio, line 296: "The VM environment the workload container runs in ..." that's hard to read.

"The VM environment where the workload container runs in" would make it easier.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+container runs in is also isolated from the "outer" host environment
+where the container manager runs.
 
 ## Runtime
 
-`containerd-shim-kata-v2` is a [containerd runtime shimv2](https://github.com/containerd/containerd/blob/v1.4.1/runtime/v2/README.md) implementation and is responsible for handling the `runtime v2 shim APIs`, which is similar to [the OCI runtime specification](https://github.com/opencontainers/runtime-spec) but simplifies the architecture by loading the runtime once and making RPC calls to handle the various container lifecycle commands. This refinement is an improvement on the OCI specification which requires the container manager call the runtime binary multiple times, at least once for each lifecycle command.
+The Kata Containers runtime (or shim), is a [shimv2](#shim-v2-architecture) compatible runtime.
```
fidencio, line 302: I think kata runtime, runtime, shim, are being used as the same thing for way too long to just get an explanation about it now.

Would it make sense to, maybe, clarify that those are all the same earlier in the document and, maybe, refer to it always as "shim"?

docs/design/architecture.md
fidencio: And Eric's comment server as a point of how kata-runtime / runtime / shim can be confusing

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
...
+The `kata-runtime` binary is a utility program that provides
+administrative commands to manipulate and query a Kata Containers
+installation.
 
-The file is well commented and provides a few "knobs" that can be used to modify the behavior of the runtime and your chosen hypervisor.
+> **Note:**
+>
+> In Kata 1.x, this program also acted as the main
+> [runtime](#runtime), but this is no longer required due to the
+> improved shimv2 architecture.
```
fidencio, line 333: I understand why we mention 1.x, but I'd rather not even mention that and avoid getting people confused by what happened in the past and is not supported anymore.

docs/design/architecture.md
fidencio: No, CRI-O uses shimv2 binary, exactly as containerd does.  So, no, nothing specific to CRI-O.

docs/design/architecture.md
fidencio: I'm not that comfortable with this part.
Users can check whether a container is using device mapper or not, but usually only admins should change the configuration (and that's also an argument for why we should disable annotations, but that's a different discussion).

docs/design/architecture.md
fidencio: They should be the same, Eric!

docs/design/architecture.md
fidencio, line 515: DAX is not enabled by default on whatever we distribute :-/

docs/design/architecture.md
```diff
@@ -279,12 +535,24 @@ file and device mapping mechanisms:
...
-  feature enabled, allowing direct page mapping and access, thus bypassing the
-  guest page cache.
+   feature enabled, allowing direct page mapping and access, thus bypassing the
+   guest page cache.
 
 ![DAX](arch-images/DAX.png)
 
-Information on the use of NVDIMM via QEMU is available in the [QEMU source code](http://git.qemu-project.org/?p=qemu.git;a=blob;f=docs/nvdimm.txt;hb=HEAD)
+Information on the use of NVDIMM via QEMU is available in the [QEMU
+source code](http://git.qemu-project.org/?p=qemu.git;a=blob;f=docs/nvdimm.txt;hb=HEAD)
```
fidencio, line 548: Better point to QEMU's github page.

docs/design/architecture.md
```diff
@@ -2,128 +2,353 @@
 
 ## Overview
 
-This is an architectural overview of Kata Containers, based on the 2.0 release.
+This document is a summary of the Kata Containers architecture.
 
-The primary deliverable of the Kata Containers project is a CRI friendly shim. There is also a CRI friendly library API behind them.
+Kata Containers provides a container runtime that creates containers
+inside lightweight, hardware isolated and secure virtual machines
```
egernst, line 8: security is a spectrum.  I would rather see something like what's on katacontainers.io:
```
Kata Containers is an open source community working to build a secure container runtime with lightweight virtual machines that feel and perform like containers, but provide stronger workload isolation using hardware virtualization technology as a second layer of defense.
```

egernst:
Big PR with lots of feedback so far - thanks for driving this (big) improvement James! Just one comment so far from me!

docs/design/architecture.md
eadamsintel: I meant to say there are 4 sentences in a row.

docs/design/architecture.md
eadamsintel: Oh I did not know that.  I thought CRIO used the `kata-runtime` binary and containerd used the `containerd-shim-kata-v2` binary. I did not realize that changed.

docs/design/architecture.md
eadamsintel: What I was trying to say is that containerd is mentioned in this section but no mention of CRIO.  There are other parts of this guide where only containerd is mentioned leading the reader to think this particular section is only containerd. I think mentioning CRIO here would help a lot.

docs/design/architecture.md
jodh-intel: I've merged two of them. They are quite detail heavy so even though it might look a little odd, I do think it might be best to keep them in separate paragraphs, but I could be convinced otherwise.

docs/design/architecture.md
jodh-intel: This doc only applies to Kata 2.x. Again, we could add extra detail, but to give this PR a chance of ever landing, let's defer that one for now.

docs/design/architecture.md
jodh-intel: That's a valid point. But where does this information belong in that case? It certainly doesn't belong in the 1.x arch doc. A high level comparison could be useful for some folk (just today we heard from a company who are using Kata 1.x heavily for example but were talking about 2.x).

Random thoughts:

- put it in an appendix.
- put it in a separate document.
- put it in the wiki?

Thoughts?

docs/design/architecture.md
jodh-intel: Added lots more detail about this in the "Environments" section.

docs/design/architecture.md
jodh-intel: Section has only been reformatted so let's attack this in a future PR.

docs/design/architecture.md
jodh-intel: I've reworded this.

docs/design/architecture.md
jodh-intel: I've pointed to the official site as that seems safer since the GH page is a clone (albeit an official one).

jodh-intel (commented):
Thanks for a great review @egernst. I've made a *lot* of changes to the doc, but realistically I cannot update the entire document in a single PR.

Networking, images and probably Storage (and CRIO?) will need future PRs to improve those areas.

jodh-intel (commented):
Thanks for a great review @eadamsintel . I've made a *lot* of changes to the doc, but realistically I cannot update the entire document in a single PR.

Networking, images and probably Storage (and CRIO?) will need future PRs to improve those areas.

On a related note, I've discovered that pending comments survive a browser restart, but not a force push ;( I've addressed a lot of your coments but my responses are somewhere in the ether ;(

jodh-intel:
And after some weird browser/GH wtf-iness, my comments seem to be back ;)

docs/design/architecture.md
jodh-intel: Yes, I'm not happy with the current diagrams. See #3245.

docs/design/architecture.md
jodh-intel: @egernst - agreed. Updated with light modifications to your suggested wording.

docs/design/architecture.md
eadamsintel: It has better flow.

docs/design/architecture.md
eadamsintel: I think a separate document makes a lot of sense.

docs/design/architecture.md
eadamsintel: Looks better.

docs/design/architecture.md
eadamsintel: It reads better to me.

docs/design/architecture.md
eadamsintel: I think how you wrote it distinguishes it better.

docs/design/architecture.md
eadamsintel: I am still confused :)   For the container workload are you referring a container image like Nginx that has the Nginx application inside of a containerized image based on Ubuntu or another OS?  That is what I think of when I think container workload. However, under user accessible you put "exec command" with a link that when you click it says  "The exec command allows an administrator or developer to enter the VM environment which is inaccessible to the container workload."   I didn't think the container workload would have any access to the VM.

docs/design/architecture.md
eadamsintel: This is hard to not confuse.   Would it make sense to have a glossary of terms at the beginning of the document or this section to state what it is?  I'll have to think about this one a bit more.  You could expand the table out to include the binaries that run on the host as well but im not sure all the columns make sense and isn't necessarily relevant to this section. 

One thing that still confuses me is the `sh(1)` as the container workload.  When I think of workload I think of envoy or nginx.

docs/design/architecture.md
eadamsintel: It is much more readable.

docs/design/architecture.md
eadamsintel: It is hard to see my previous review but it looks like in step 8 there should be i.  ii.  and iii.  for the 3 sentences.  I only see the first two.

docs/design/architecture.md
eadamsintel: They are long sentences so it is fine.

docs/design/architecture.md
eadamsintel: I just thought about this some more.  Does that mean crio uses the `containerd-shim-kata-v2` binary with the containerd name in it?  If so shouldn't it be called  `kata-shim-v2 binary` instead or something like that?

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+> commonly used so this document uses the standard term.
+
+For the purposes of this document, the most important part of the
+command above is the container image the user has requested. Normally,
+a container image will be "pulled" (downloaded) from a remote site and
+a copy stored locally. This local container image is used by the
+container manager to create an [OCI bundle](#oci-bundle) which will
+form the environment the container will run in.
+
+containerd itself does not create containers, it simply manages them.
```
eadamsintel, line 173: Containerd should be capitalized. Al,so the comma should be a period since it is two separate sentences.

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+> The term "container image" is confusing since the image in question
+> is **not** a container: it is simply a set of files ("an image")
+> that can be used to _create_ a container. The term "container
+> template" would be more accurate but the term "container image" is
+> commonly used so this document uses the standard term.
+
+For the purposes of this document, the most important part of the
+command above is the container image the user has requested. Normally,
+a container image will be "pulled" (downloaded) from a remote site and
+a copy stored locally. This local container image is used by the
```
eadamsintel, line 169: I would say  "and store a copy locally"

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+container manager to create an [OCI bundle](#oci-bundle) which will
+form the environment the container will run in.
+
+containerd itself does not create containers, it simply manages them.
+However, before it launches a [runtime](#runtime) to create the
+container, containerd will create an [OCI rootfs bundle](https://github.com/opencontainers/runtime-spec/blob/main/bundle.md).
```
eadamsintel, line 175: Right above this you mention "OCI bundle" and here you mention "OCI rootfs bundle"  Are these two different things?  It seems like they are the same but named slightly differently.

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+
+## OCI bundle
+
+To understand what follows, it is important to know at a high level
+how an OCI compatible container is created.
```
eadamsintel, line 180: I don't see "OCI" defined anywhere in the document but other acronyms are explicitly called out.

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+  This is a JSON file called `config.json`.
+
+  The container manager will create this file so that:
+
+  - The `root.path` value is set to the full path of the specified
+    container rootfs.
+
+    In [our example](#example-command) this value will be `ubuntu`.
+
+  - The `process.args` array to the list of commands the user wishes
```
eadamsintel, line 212: Did you mean to say " the process.args array contains the list of commands the user wishes..."

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+In our [example](#example-command), the workload is the `sh(1)` command.
+
+### Workload root filesystem
+
+For details of how the [runtime](#runtime) makes the
+[container image](#container-image) chosen by the user available to
+the workload process, see the
+[Container creation](#container-creation) and [storage](#storage) sections.
+
+See the [container creation](#container-creation) section for further
```
eadamsintel, line 232: This seems redundant with the previous sentence.

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
+The [agent](#agent) will detect when the [workload](#workload) process
+exits, capture its exit status (see `wait(2)`) and return that value
+to the [runtime](#runtime) by specifying it as the response to the
+`WaitProcess` agent API call made by the [runtime](#runtime).
+
+The runtime then passes the value back to the container manager by the
+`Wait` [shimv2 API](#shim-v2-architecture) call.
+
+Once the workload has fully exited, the VM is no longer needed and the
+runtime cleans up the environment (which includes killing the
```
eadamsintel, line 359: I would consider using the word  "terminating"

docs/design/architecture.md
```diff
@@ -2,128 +2,617 @@
...
-`containerd-shim-kata-v2` is a [containerd runtime shimv2](https://github.com/containerd/containerd/blob/v1.4.1/runtime/v2/README.md) implementation and is responsible for handling the `runtime v2 shim APIs`, which is similar to [the OCI runtime specification](https://github.com/opencontainers/runtime-spec) but simplifies the architecture by loading the runtime once and making RPC calls to handle the various container lifecycle commands. This refinement is an improvement on the OCI specification which requires the container manager call the runtime binary multiple times, at least once for each lifecycle command.
+The Kata Containers runtime (the [`containerd-shim-kata-v2`](../../src/runtime/cmd/containerd-shim-kata-v2
+) binary) is a [shimv2](#shim-v2-architecture) compatible runtime.
 
-`containerd-shim-kata-v2` heavily utilizes the
-[virtcontainers package](../../src/runtime/virtcontainers/), which provides a generic, runtime-specification agnostic, hardware-virtualized containers library.
+> **Note:**
+>
+> The Kata Containers runtime is sometimes referred to as the Kata
+> "shim". Both terms are correct since the `containerd-shim-kata-v2`
```
eadamsintel, line 559: I always try to avoid ending sentences with a quotation because technically you are supposed to put the period inside the quotation to be correct. It just looks wrong when I do it that way.

docs/design/architecture.md
eadamsintel: sounds good to me.

eadamsintel (commented):
I added some additional comments on the new sections.

docs/design/architecture.md
jodh-intel: Actually, containerd stylises itself with a lower-case initial letter 'c' so doesn't need to change afaik:

- https://github.com/containerd/containerd
- https://containerd.io/

That's why we've specified it as such in our dictionary:

- https://github.com/kata-containers/tests/blob/main/cmd/check-spelling/data/projects.txt#L12

jodh-intel (commented):
Hi @eadamsintel - Branch updated. I've addressed most of your comments but some can be deferred to #3246 I think. Once this PR lands, I'll work on splitting the arch doc into pieces for #3246 so that we can attach the different sections (files) in parallel.

jodh-intel:
@egernst - could you tal again please? Once this lands, I'll split the doc into pieces on #3246 so we can polish individual sections further (focussing on storage and networking).

docs/design/architecture.md
eadamsintel: I was really referring to the fact that a sentence is starting with a lower case letter.  If you want to keep it lower case that is fine by me but it looks wrong when it is the first word of a sentence.   You could say  "Containers are not actually created by containerd."  or "The containerd binary..."

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
+1. The [agent](#agent) is started as part of the VM boot.
+
+1. The runtime calls the agent's `CreateSandbox` API to request the
+   agent create a container:
+
+   1. The agent creates a [container environment](#environments)
+      in the container specific directory that contains the
+      [container rootfs](#environments).
+
+      The container environment
```
eadamsintel, line 299: The way this renders it looks like it is the second step of a 3 step process and the ii. is missing in front of it.  It renders very similar to this. 

```
i. The agent creates a container environment in the container specific directory that contains the container rootfs.

The container environment hosts the workload in the container rootfs directory.

ii. The agent spawns the workload inside the container environment.
```

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
+   1. The agent spawns the workload inside the container environment.
+
+   > **Notes:**
+   >
+   > - The container environment created by the agent is equivalent to
+   >   a container environment created by the
+   >   [`runc`](https://github.com/opencontainers/runc) OCI runtime;
+   >   Linux cgroups and namespaces are created inside the VM by the
+   >   [guest kernel](#guest-kernel) to isolate the workload from the
+   >   VM environment the container is created in.
```
eadamsintel, line 312: Someone might wonder why cgroups are enforced inside the VM since the VM is isolated from other VM's.  I don't know if you want to include justifications as why things are they way they are in the arch doc but isn't part of the reason because it provides better isolation between containers in a pod that are in a single VM? It is just a thought to consider.

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
+>   script as `root` and looking at the "Image details" section of the
+>   output.
 
 #### Root filesystem image
 
-The default packaged root filesystem image, sometimes referred to as the "mini O/S", is a
-highly optimized container bootstrap system based on [Clear Linux](https://clearlinux.org/). It provides an extremely minimal environment and
+The default packaged rootfs image, sometimes referred to as the _mini
+O/S_, is a highly optimized container bootstrap system. On Intel
+x86_64 platforms, it is based on [Clear Linux](https://clearlinux.org)
```
eadamsintel, line 424: As I look at this again Clear Linux is recommended for rootfs but Alpine is recommended for initrd. Nothing really explains why not use the same for both.  I think adding a little clarification that Clear Linux is highly optimized and supports systemd init would help. I would clarify on the Alpine image below that it doesn't support systemd init which is why agent as the init works in that scenario. I thought on this repo somewhere there was a guide stating which OS images work for systemd rootfs and which ones work with agent as the init but I can't find it. If that guide does exist I would add a link to it here somewhere.

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
+[container environment](#environments).
+
+> **Note:**
+>
+> The hypervisor process runs inside an environment slightly different
+> to the host environment:
+>
+> - It is run in a different cgroup environment to the host.
+> - It is given a separate network namespace from the host.
+> - If the [OCI configuration specifies an SELinux label](https://github.com/opencontainers/runtime-spec/blob/main/config.md#linux-process),
```
eadamsintel, line 522: I believe  "an" should be "a" since label is singular.

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
+running within those containers. Only a single agent process is run
+for each VM created.
 
-For the 2.0 release, the `kata-agent` is rewritten in the [RUST programming language](https://www.rust-lang.org/) so that we can minimize its memory footprint while keeping the memory safety of the original GO version of [`kata-agent` used in Kata Container 1.x](https://github.com/kata-containers/agent). This memory footprint reduction is pretty impressive, from tens of megabytes down to less than 100 kilobytes, enabling Kata Containers in more use cases like functional computing and edge computing.
+### Agent communications protocol
 
-The `kata-agent` execution unit is the sandbox. A `kata-agent` sandbox is a container sandbox defined by a set of namespaces (NS, UTS, IPC and PID). `shimv2` can
-run several containers per VM to support container engines that require multiple
-containers running inside a pod.
+The agent communicates with the other Kata components (primary the
```
eadamsintel, line 536: primary = primarily

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
+The Kata Containers runtime (the [`containerd-shim-kata-v2`](../../src/runtime/cmd/containerd-shim-kata-v2
+) binary) is a [shimv2](#shim-v2-architecture) compatible runtime.
 
-`containerd-shim-kata-v2` heavily utilizes the
-[virtcontainers package](../../src/runtime/virtcontainers/), which provides a generic, runtime-specification agnostic, hardware-virtualized containers library.
+> **Note:**
+>
+> The Kata Containers runtime is sometimes referred to as the Kata
+> _shim_. Both terms are correct since the `containerd-shim-kata-v2`
+> is a container runtime. And that runtime implements the containerd
```
eadamsintel, line 558: "And the runtime implements.." is a sentence fragment. You could just use a comma to join the sentences.

docs/design/architecture.md
eadamsintel: Sure lets defer to [3246](https://github.com/kata-containers/kata-containers/issues/3246)

docs/design/architecture.md
eadamsintel: Lets defer then.

docs/design/architecture.md
eadamsintel: Agree

eadamsintel (commented):
Overall it is looking really good. Most of what I found were relatively minor things and new suggestions once I reread the entire thing in context.

docs/design/architecture.md
```diff
@@ -221,70 +754,84 @@ or `container`. Kata Containers will then use this annotation to decide if it ne
...
+[guest VM environment](#environments) to become the VM's
+[rootfs](#root-filesystem).
+
+If the [configured](#configuration) [hypervisor](#hypervisor) is set
+to either QEMU or Cloud Hypervisor, DAX is used by utilising the
+features as shown in the table below:
+
+| Hypervisor | Feature used |
+|-|-|
+| Cloud Hypervisor | `dax` `FsConfig` configuration option |
```
likebreath, line 794: The statement that CLH is utilizing the DAX feature for the rootfs is correct, but it is actually using pmem device (instead of virtio-fs device) for the rootfs image. 

Also, technically the DAX option for virtio-fs now is disabled from the configuration-clh.toml [1] due to some instability issues related to virtio-fs. @jcvenegas Please correct me or add more comments if needed.

[1] https://github.com/kata-containers/kata-containers/blob/main/src/runtime/Makefile#L169-L170

likebreath (approved):
@jodh-intel Great work. Only one comment related to CLH. Thank you.

docs/design/architecture.md
jamesodhunt: I agree with you @eadamsintel, but someone seems to have decided that project is exempt from the normal rules ;)

docs/design/architecture.md
jodh-intel: This is all explained in the "Environments" section with concrete examples of the terms. I've clarified the exec command text a little but the key word in that sentence is "**in**accessible". Did you possibly mis-read that as "accessible"? I've reworded this to now say:

> The exec command allows an administrator or developer to enter the VM root environment which is not accessible by the container workload.

docs/design/architecture.md
jodh-intel: Hi @likebreath - Thanks for reviewing! I'm not quite clear what needs to change for your first point. Are you referring to the table in the "Environments" section? If so do you mean that the container rootfs needs to show DAX for CH?

I think we need another table like the following, but I'm wondering if that should be added to https://github.com/kata-containers/kata-containers/blob/main/docs/hypervisors.md rather than this doc:

| Hypervisor | VM rootfs mount type | container rootfs mount type |
|-|-|-|
| ACRN | ... | ... |
| CH | DAX | DAX |
| FC | ... | ... | 
| QEMU | DAX | virtio-fs |

jodh-intel: /cc @jcvenegas.

docs/design/architecture.md
eadamsintel: No I read it as the "kubectl" exec command allows a developer into the VM environment but the container workload doesn't have access to that environment.  I was thinking of exec as in  "kubectl exec -it ..."  and maybe that is why I am confused.  I did not think anyone with "kubectl exec" access would have access to the VM environment. I think exec in this example is being used in a different way and isn't kubernetes related.

docs/design/architecture.md
```diff
@@ -2,128 +2,615 @@
...
-will be used. `kata-agent` uses this mount point as the root filesystem for the
-container processes.
+## Root filesystem
 
-## Virtualization
+This document uses the term _rootfs_ to refer to a root filesystem
+which is mounted as the top-level directory ("`/`") and often referred
+to as _slash_.
+
+It is important to understand this term since the overall system uses
```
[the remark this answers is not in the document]
jodh-intel: I think I missed this comment first time round. Getting the ordering right is difficult as we do one of:

- Defining a huge bunch of stuff up front

  > "You'll need to understand these concepts for later, so absorb a huge bunch of stuff now"

- Or we go through a slightly less scary set of steps where we refer to concepts not yet defined

  > "What the heck is this `$thing` I keep seeing mentioned?"

The current PR does a bit of both. We could try moving some of the initial sections around maybe or is the current revision clear enough do you think?

jodh-intel (commented):
Thanks @eadamsintel. More updates to your comments.

docs/design/architecture.md
egernst: The guest OS rootfs utilizes DAX.  This is separate from virtiofs

egernst (approved):
big improvement.  Let's keep it going!

Thanks James.

Issue #3196, opened 2021-12-03
Missing commit information for kata-runtime in the static-kata package release


Bevisy:
Download kata-static-*.tar.xz and extract. Then using `./kata-runtime --help`, we will find commit message missing as below:
```sh
NAME:
   kata-runtime - kata-runtime runtime

kata-runtime is a command line program for running applications packaged
according to the Open Container Initiative (OCI).

USAGE:
   kata-runtime [global options] command [command options] [arguments...]

VERSION:
   kata-runtime  : 2.4.0-alpha0
   commit   : <<unknown>>           <<---- missing
   OCI specs: 1.0.2-dev
```

Bevisy:
I think it maybe easily fixed just by adding `git` to the shim-v2 build image.

PR #3197, opened 2021-12-03
packaging: Fix missing commit message in building kata-runtime

Bevisy:
add `git` package to the shim-v2 build image

Fixes: #3196

### DIFF
--- a/tools/packaging/static-build/shim-v2/Dockerfile
+++ b/tools/packaging/static-build/shim-v2/Dockerfile
@@ -5,7 +5,13 @@
 FROM ubuntu
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y make curl sudo gcc
+RUN apt-get update && \
+    apt-get install -y \
+        curl \
+        gcc \
+        git \
+        make \
+        sudo
 
 ADD install_go.sh /usr/bin/install_go.sh
 ARG GO_VERSION


tools/packaging/static-build/shim-v2/Dockerfile
```diff
@@ -2,10 +2,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM ubuntu
+FROM ubuntu:20.04
```
fidencio, line 5: This is a valid change, but I'd like to ask you to split this into a different commit.  Also, please, check if the other Dockerfiles need the same change.

tools/packaging/static-build/shim-v2/Dockerfile
```diff
@@ -2,10 +2,16 @@
...
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y make curl sudo gcc
+RUN apt-get update && \
+    apt-get install -y \
+        curl \
+        gcc \
+        git \
+        make \
+        sudo
```
fidencio, line 14: This part is good to go! :-)

fidencio (changes requested):
@Bevisy, thanks for the PR!
I've added one comment inline, and once that's fixed we can have this one merged.

tools/packaging/static-build/shim-v2/Dockerfile
Bevisy: Ok, I will remove this change. There is a nother pr #3109 which is used to delint dockerfile.

Bevisy:
```sh
❯ ./kata-runtime --help
NAME:
   kata-runtime - kata-runtime runtime

kata-runtime is a command line program for running applications packaged
according to the Open Container Initiative (OCI).

USAGE:
   kata-runtime [global options] command [command options] [arguments...]

VERSION:
   kata-runtime  : 2.4.0-alpha0
   commit   : 31547cfbcb190bd20d228aea346fd7e996aac27d       <--- fixed
   OCI specs: 1.0.2-dev
```

fidencio (approved):
lgtm, thanks @Bevisy!

fidencio:
@Bevisy, and about the other change you proposed (fixing a version of ubuntu in the dockerfiles), please, open a new PR for that as that's worth having fixed as well.

Bevisy:
> 

By reviewing #3109, I found that this problem happened to be solved by adding `git`, and it also included the modification of the specific version of the image. So we may not open a new PR for the time being to fix the image version problem, but we still need to create a new backport PR to ensure that this modification can appear in release v2.3.

liubin (approved):

Issue #3217, opened 2021-12-07
Trace span name not correct


liubin:
`StopVM` should be `stopSandbox`:

https://github.com/kata-containers/kata-containers/blob/01b6ffc0a4be5071d3e7ac9938ad5103c232d764/src/runtime/virtcontainers/kata_agent.go#L910-L912

PR #3218, opened 2021-12-07
runtime: correct span name for stopSandbox function

liubin:
Normally the span name should be the same as function
name, so chagne `StopVM` to `stopSandbox`.

Fixes: #3217


snir911 (approved):
/retest-ubuntu
LGTM!

cmaf (approved):

Issue #3193, opened 2021-12-03
"hypervisor config does not match" error in template mode


liubin:
Using `main` branch, when trying to init a factory VM, then run a pod will get this warning:

[...]

PR #3028, opened 2021-11-12
Hypervisor cleanup, refactoring

egernst:
With a goal of creating a standalone hypervisor pkg, there are a few challenges that we're looking to overcome in an effort to make the code base more maintainable, including appropriate levels of abstraction. 

This PR addresses a few of them:
  1) hypervisor and its implementors should not rely on any sandbox-level functionality or details.
  2) hypervisor shouldn't rely on persist API. Instead, we can define the data types that are relevant to hypervisor persistence and implement functions for loading/storing this structure. The persist package itself can then consume these data types, functions.
 3) hypervisor logging shouldn't rely on a virtcontainers logger. Instead, implement `SetLogger`, which virtcontainers can call.

In doing this, we acknowledge that parts of ACRN will be broken. The current hypervisor interface implementation is too tightly coupled with persist API and with "sandbox." This needs to be fixed. Rather than removing the interface implementation entirely, we are just removing the problematic parts that impact common hypervisor code.

Outstanding work that isn't addressed in this PR, but will be in future, includes:
 - [ ] consume network device types rather than endpoints: we only care about the VM-side of the net pair; no need to consume endpoint information in hypervisor part of code
 - [ ] clean up dependencies on tracing package
 - [ ] consolidate hypervisor utils/types into hypervisors pkg
 - [ ] move hypervisor code to the hypervisors pkg

PR #3226, opened 2021-12-07
runtime/template: Handling new attributes for hypervisor config

liubin:
Some new attributes are added to hypervisor config:
- VMStorePath
- RunStorePath
- SharedPath

These attributes should be handled in two places:

- reset when check the new hypervisor's config is suitable
  to the base config.
- copy from new hypervisor's config when create new VM

Fixes: #3193


fidencio (approved):
lgtm, thanks @liubin!

snir911 (approved):
Thanks

snir911:
@liubin does backporting needed here? https://github.com/kata-containers/kata-containers/pull/3028 doesn't exist on stable AFAICT

liubin:
> 

This depends on #3028, they should be a pair if backporting is needed.

Issue #3219, opened 2021-12-07
q.stopVirtiofsd should only be called when virtio-fs is used for file sharing


liubin:
Setting `shared_fs = "virtio-9p"` will get this warning:

> Dec  7 15:05:06 kant-server containerd[12860]: time="2021-12-07T15:05:06.306571213+08:00" level=warning msg="The virtiofsd had stopped" name=containerd-shim-v2 pid=10778 sandbox=39a992835d1277873aacfa3c78287eadca9727ab19e91214e64d959e0cbcb73b source=virtcontainers/hypervisor subsystem=qemu

PR #3220, opened 2021-12-07
runtime: only call stopVirtiofsd when shared_fs is virtio-fs

liubin:
If shared_fs is set to virtio-9p, the virtiofsd is not started,
so there is no need to stop it.

Fixes: #3219

Signed-off-by: bin <bin@hyper.sh>

PR #3230, opened 2021-12-08
[backport]runtime: only call stopVirtiofsd when shared_fs is virtio-fs

liubin:
If shared_fs is set to virtio-9p, the virtiofsd is not started,
so there is no need to stop it.

Fixes: #3219

Backport: #3220


fidencio (approved):
lgtm, thanks @liubin!

devimc (approved):

Issue #3236, opened 2021-12-08
RemoteEventsPublisher rpc client should be closed properly


Bevisy:
There is no call to publisher.Close() here 
```go
// src/runtime/pkg/containerd-shim-v2/service.go L:259
func (s *service) forward(ctx context.Context, publisher events.Publisher) {
	for e := range s.events {
		ctx, cancel := context.WithTimeout(ctx, timeOut)
		err := publisher.Publish(ctx, getTopic(e), e)
		cancel()
		if err != nil {
			shimLog.WithError(err).Error("post event")
		}
	}
}
```

Bevisy:
Closed by https://github.com/kata-containers/kata-containers/pull/3237#discussion_r765423947

PR #3237, opened 2021-12-08
shim/service: Close events publisher properly 

Bevisy:
Fixes: #3236

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/service.go
+++ b/src/runtime/pkg/containerd-shim-v2/service.go
@@ -17,7 +17,6 @@ import (
 	eventstypes "github.com/containerd/containerd/api/events"
 	"github.com/containerd/containerd/api/types/task"
 	"github.com/containerd/containerd/errdefs"
-	"github.com/containerd/containerd/events"
 	"github.com/containerd/containerd/namespaces"
 	cdruntime "github.com/containerd/containerd/runtime"
 	cdshim "github.com/containerd/containerd/runtime/v2/shim"
@@ -256,15 +255,14 @@ func (s *service) StartShim(ctx context.Context, opts cdshim.StartOpts) (_ strin
 	return address, nil
 }
 
-func (s *service) forward(ctx context.Context, publisher events.Publisher) {
+func (s *service) forward(ctx context.Context, publisher cdshim.Publisher) {
 	for e := range s.events {
-		ctx, cancel := context.WithTimeout(ctx, timeOut)
 		err := publisher.Publish(ctx, getTopic(e), e)
-		cancel()
 		if err != nil {
 			shimLog.WithError(err).Error("post event")
 		}
 	}
+	publisher.Close()
 }
 
 func (s *service) send(evt interface{}) {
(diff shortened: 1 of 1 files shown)


Bevisy:
static checks  error as below:
```sh
curl: (22) The requested URL returned error: 503 
[install_yq.sh:32] ERROR: curl -o /home/runner/work/kata-containers/kata-containers/bin/yq -LSsf https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
```
Wait until tomorrow to retrigger :)

src/runtime/pkg/containerd-shim-v2/service.go
```diff
@@ -256,15 +255,14 @@ func (s *service) StartShim(ctx context.Context, opts cdshim.StartOpts) (_ strin
...
+func (s *service) forward(ctx context.Context, publisher cdshim.Publisher) {
 	for e := range s.events {
-		ctx, cancel := context.WithTimeout(ctx, timeOut)
 		err := publisher.Publish(ctx, getTopic(e), e)
-		cancel()
 		if err != nil {
 			shimLog.WithError(err).Error("post event")
 		}
 	}
+	publisher.Close()
```
liubin, line 265: I think the containerd is taking the charge of [closing](https://github.com/containerd/containerd/blob/e1ad7791077916aac9c1f4981ad350f0e3fce719/runtime/v2/shim/shim.go#L203-L207
) the publisher:


```golang
	ttrpcAddress := os.Getenv(ttrpcAddressEnv)
	publisher, err := NewPublisher(ttrpcAddress)
	if err != nil {
		return err
	}
	defer publisher.Close()

	ctx := namespaces.WithNamespace(context.Background(), namespaceFlag)
	ctx = context.WithValue(ctx, OptsKey{}, Opts{BundlePath: bundlePath, Debug: debugFlag})
	ctx = log.WithLogger(ctx, log.G(ctx).WithField("runtime", id))
	ctx, cancel := context.WithCancel(ctx)
	service, err := initFunc(ctx, idFlag, publisher, cancel)
	if err != nil {
		return err
	}
```

Bevisy: I agree with you 😄 
By the way, do we need ctx with timeout in this function? It looks like containerd also cover it.
```go
func (l *RemoteEventsPublisher) forwardRequest(ctx context.Context, req *v1.ForwardRequest) error {
	service, err := l.client.EventsService()
	if err == nil {
		fCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
		_, err = service.Forward(fCtx, req)
		cancel()
		if err == nil {
			return nil
		}
	}
...
	// try again with a fresh context, otherwise we may get a context timeout unexpectedly.
	fCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
	_, err = service.Forward(fCtx, req)
	cancel()
	if err != nil {
		return err
	}

	return nil
}
```

liubin: @Bevisy without clear evidence, we can keep it.

Issue #3242, opened 2021-12-08
Automate creation of Secure Execution image


Jakob-Naucke:
Creating a protected image for IBM Secure Execution for Linux requires an initrd as per #3127, a kernel, and a guest command line to be combined and encrypted. This should be automated.

PR #3243, opened 2021-12-08
CCv0 | kata-deploy: Support Secure Execution

Jakob-Naucke:
Add script and Makefile target for Secure Execution. This way, a full
Kata image can be deployed just by preparing offline keys, agent config,
and the host key document.

Fixes: #3242

I have tested this end-to-end. The command is something like `UMOCI=yes AA_KBC=offline_fs_kbc INCLUDE_ROOTFS=include_rootfs/ HKD=/path/to/HKD-8561-0123456.crt make secure-execution`, which creates an `/opt/kata` tarball.
/cc @fitzthum @magowan @sameo @stevenhorsman

### DIFF
--- a/src/runtime/pkg/katautils/config.go
+++ b/src/runtime/pkg/katautils/config.go
@@ -465,11 +465,13 @@ func (h hypervisor) getInitrdAndImage() (initrd string, image string, err error)
 
 	image, errImage := h.image()
 
-	if image != "" && initrd != "" {
+	if h.ConfidentialGuest && h.MachineType == vc.QemuCCWVirtio {
+		if image != "" || initrd != "" {
+			return "", "", errors.New("Neither the image or initrd path may be set for Secure Execution")
+		}
+	} else if image != "" && initrd != "" {
 		return "", "", errors.New("having both an image and an initrd defined in the configuration file is not supported")
-	}
-
-	if errInitrd != nil && errImage != nil {
+	} else if errInitrd != nil && errImage != nil {
 		return "", "", fmt.Errorf("Either initrd or image must be set to a valid path (initrd: %v) (image: %v)", errInitrd, errImage)
 	}
 
@@ -599,16 +601,6 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 	pflashes, err := h.PFlash()
 	if err != nil {
 		return vc.HypervisorConfig{}, err
-	}
-
-	if image != "" && initrd != "" {
-		return vc.HypervisorConfig{},
-			errors.New("having both an image and an initrd defined in the configuration file is not supported")
-	}
-
-	if image == "" && initrd == "" {
-		return vc.HypervisorConfig{},
-			errors.New("either image or initrd must be defined in the configuration file")
 	}
 
 	firmware, err := h.firmware()
--- a/src/runtime/virtcontainers/hypervisor.go
+++ b/src/runtime/virtcontainers/hypervisor.go
@@ -544,11 +544,13 @@ func (conf *HypervisorConfig) Valid() error {
 		return fmt.Errorf("Missing kernel path")
 	}
 
-	if conf.ImagePath == "" && conf.InitrdPath == "" {
+	if conf.ConfidentialGuest && conf.HypervisorMachineType == QemuCCWVirtio {
+		if conf.ImagePath != "" || conf.InitrdPath != "" {
+			return fmt.Errorf("Neither the image or initrd path may be set for Secure Execution")
+		}
+	} else if conf.ImagePath == "" && conf.InitrdPath == "" {
 		return fmt.Errorf("Missing image and initrd path")
-	}
-
-	if conf.ImagePath != "" && conf.InitrdPath != "" {
+	} else if conf.ImagePath != "" && conf.InitrdPath != "" {
 		return fmt.Errorf("Image and initrd path cannot be both set")
 	}
 
@@ -570,7 +572,7 @@ func (conf *HypervisorConfig) Valid() error {
 
 	if conf.BlockDeviceDriver == "" {
 		conf.BlockDeviceDriver = defaultBlockDriver
-	} else if conf.BlockDeviceDriver == config.VirtioBlock && conf.HypervisorMachineType == "s390-ccw-virtio" {
+	} else if conf.BlockDeviceDriver == config.VirtioBlock && conf.HypervisorMachineType == QemuCCWVirtio {
 		conf.BlockDeviceDriver = config.VirtioBlockCCW
 	}
 
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -15,8 +15,8 @@ endef
 
 kata-tarball: | all-parallel merge-builds
 
-all-parallel:
-	${MAKE} -f $(MK_PATH) all -j$$(( $$(nproc) - 1  )) NO_TTY="true" V=
+%-parallel:
+	${MAKE} -f $(MK_PATH) $(subst -parallel,,$@) -j$$(( $$(nproc) - 1  )) NO_TTY="true" V=
 
 all: cloud-hypervisor-tarball \
 	firecracker-tarball \
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -62,3 +62,8 @@ install-tarball:
 
 image: kata-tarball
 	$(MK_DIR)kata-deploy-build-and-upload-image.sh $(CURDIR)/kata-static.tar.xz
+
+secure-execution-tarballs: kernel-tarball qemu-tarball rootfs-initrd-tarball shim-v2-tarball
+
+secure-execution: secure-execution-tarballs-parallel
+	$(MK_DIR)/kata-deploy-secure-execution.sh
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
@@ -1,0 +1,87 @@
+#!/bin/bash -eu
+# Copyright (c) 2021 IBM Corp.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[ -n "${DEBUG:-}" ] && set -x
+[ "$(uname -m)" = s390x ] || (echo >&2 "ERROR: Creation of Secure Execution images is currently only supported on s390x." && exit 1)
+
+kata_build_dir=build
+cmdline="loglevel=0 panic= scsi_mod.scan=none swiotlb=262144 agent.config_file=/etc/kata-containers/agent.toml"
+
+usage() {
+	cat >&2 << EOF
+Usage:
+  $0 [options]
+
+Options:
+  -b <build dir>     : Kata build directory, containing kata-static tarballs
+                       Default: "$kata_build_dir"
+  -c <kernel cmdline>: Guest kernel command line
+                       Default: "$cmdline"
+  -h                 : Show this help
+
+Environment variables:
+  HKD (required): Secure Execution host key document, generally specific to your machine. See
+                  https://www.ibm.com/docs/en/linux-on-systems?topic=tasks-verify-host-key-document
+                  for information on how to retrieve and verify this document.
+  DEBUG         : If set, display debug information.
+EOF
+	exit "${1:-0}"
+}
+
+while getopts "b:c:h" opt; do
+	case $opt in
+		b) kata_build_dir="$OPTARG";;
+		c) cmdline="$OPTARG";;
+		h) usage;;
+		*) usage 1;;
+	esac
+done
+shift $(( OPTIND - 1 ))
+
+[ -n "${HKD:-}" ] || (echo >&2 "No host key document specified." && usage 1)
+command -v genprotimg || (echo >&2 "genprotimg is not installed. Install s390-tools." && usage 1)
+
+declare hkd_options
+eval "for hkd in $HKD; do
+	hkd_options+=\"--host-key-document=\\\"\$hkd\\\" \"
+done"
+
+tar_path="kata-static.tar.xz"
+
+tarball_content_dir="$PWD/kata-tarball-content"
+protimg_content_dir="$PWD/kata-protimg-content"
+rm -rf "$tarball_content_dir" "$protimg_content_dir"
+mkdir "$tarball_content_dir" "$protimg_content_dir"
+
+pushd "$kata_build_dir"
+for c in qemu shim-v2; do
+	tar xvf kata-static-$c.tar.xz -C "$tarball_content_dir"
+done
+
+for c in kernel rootfs-initrd; do
+	tar xvf kata-static-$c.tar.xz -C "$protimg_content_dir"
+done
+popd
+
+parmfile="$(mktemp --suffix=-cmdline)"
+trap 'rm -f "$parmfile"' EXIT
+chmod 600 "$parmfile"
+echo "$cmdline" > "$parmfile"
+
+protimg_kata_dir="$protimg_content_dir/opt/kata/share/kata-containers"
+tarball_kata_dir="$tarball_content_dir/opt/kata/share/kata-containers"
+mkdir "$tarball_kata_dir"
+
+eval genprotimg \
+	"$hkd_options" \
+	--output="$tarball_kata_dir/kata-containers-secure.img" \
+	--image="$protimg_kata_dir/vmlinuz.container" \
+	--ramdisk="$protimg_kata_dir/kata-containers-initrd.img" \
+	--parmfile="$parmfile" \
+	--no-verify # TODO add support once available
+
+pushd "$tarball_content_dir"
+tar cJvf "$tar_path" .
+popd
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -12,6 +12,8 @@ crio_drop_in_conf_dir="/etc/crio/crio.conf.d/"
 crio_drop_in_conf_file="${crio_drop_in_conf_dir}/99-kata-deploy"
 containerd_conf_file="/etc/containerd/config.toml"
 containerd_conf_file_backup="${containerd_conf_file}.bak"
+kata_share_dir="/opt/kata/share/kata-containers"
+kata_defaults_dir="/opt/kata/share/defaults/kata-containers"
 
 shims=(
 	"fc"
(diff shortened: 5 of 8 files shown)


tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
```diff
@@ -0,0 +1,85 @@
...
+		b) kata_build_dir="$OPTARG";;
+		c) cmdline="$OPTARG";;
+		h) usage;;
+		*) usage 1;;
+	esac
+done
+shift $(( OPTIND - 1 ))
+
+[ -n "$HKD" ] || (echo >&2 "No host key document specified." && usage 1)
+command -v genprotimg || (echo >&2 "genprotimg is not installed. Install s390-tools." && usage 1)
```
jodh-intel, line 43: Oh - is this PR arch-specific? If so, please add a check on `uname -m` and die if the value is unexpected.

jodh-intel (commented):
Thanks @Jakob-Naucke. A few comments...

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
```diff
@@ -0,0 +1,85 @@
...
+	--host-key-document="$HKD" \
+	--output="$tarball_kata_dir/kata-containers-secure.img" \
+	--image="$protimg_kata_dir/vmlinuz.container" \
+	--ramdisk="$protimg_kata_dir/kata-containers-initrd.img" \
+	--parmfile="$parmfile" \
+	--no-verify # TODO add support once available
+
+# Alibi initrd
+# TODO just remove the check instead?
+echo 0 > "$tarball_content_dir/opt/kata/share/kata-containers/kata-containers-dummy-initrd.img"
```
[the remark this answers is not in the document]
Jakob-Naucke: In Secure Execution, only a `-kernel` is specified (with the image generated above, contains kernel, initrd, and kernel cmdline in encrypted form), `-initrd` is ignored. However, we still need Kata to do an initrd-style launch (affects block devices etc.), which is detected by having an initrd set (Kata will also complain if you set neither image nor initrd). https://github.com/kata-containers/kata-containers/blob/97e18cf2d0d1a83b139c5117299c1766b6365a22/src/runtime/pkg/katautils/config.go#L458-L460

Kata checks whether the initrd has contents: https://github.com/kata-containers/kata-containers/blob/97e18cf2d0d1a83b139c5117299c1766b6365a22/src/runtime/pkg/katautils/config.go#L1298-L1300
Therefore, I can't just `touch` that file, but this makes the test pass (I just found out that `echo > …` would even be enough).

Is that a bit of a horrible workaround? Yeah.
We could reduce this to a `touch` by checking `HypervisorConfig.ConfidentialGuest` (a field I hadn't realised earlier was available) && s390x (or removing the check altogether, as suggested in the comment).
What would be even nicer in theory is to not specify an initrd at all, but that's some deeper code base modification.
We could also simply set initrd to a file we know exists (e.g. the kernel image, which doesn't really seem to have a performance impact).
@jodh-intel what would be your favourite approach?

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -62,3 +62,8 @@ install-tarball:
 
 image: kata-tarball
 	$(MK_DIR)kata-deploy-build-and-upload-image.sh $(CURDIR)/kata-static.tar.xz
+
+secure-execution-tarballs: kernel-tarball qemu-tarball rootfs-initrd-tarball shim-v2-tarball
```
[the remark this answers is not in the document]
Jakob-Naucke: This whole workflow might deserve some documentation, but I had wanted to put out some things before the holidays and was also still involved in #3346 et al. :) Let me get back to you.

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
Jakob-Naucke: Secure Execution is the TEE on s390x, yes. _Right now_, this PR is architecture-specific, but not because you can't get `s390-tools` on other platforms (you can), but because the builds of initrd etc. will be native, so while all architectures could run this, only a result produced by s390x would work. I hope to enable cross builds one day, but that's more to come. 

For now, it is probably better to add a catch, yes. Will update.

Jakob-Naucke:
@jodh-intel I have addressed some comments. There is also another update that I still want to provide, which is supporting several HKDs.

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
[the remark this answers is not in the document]
jodh-intel: Thanks for the explanation. Short term I'd just add a summary of your comments here as a comment in the code explaining what that `echo 0` is for as it looks rather odd. The file name does give a clue but maybe you could make it even clearer by doing something like:

```bash
echo this-is-a-fake-initrd > kata-containers-dummy-initrd.img
```

I'd rather we kept that size check as we used to get a lot of issues raised way back by users who had built their own images but hadn't successfully installed them, so the check provided a hint to them as to what had gone wrong when they tried to start a container.

But yes, a better solution would be to check `HypervisorConfig.ConfidentialGuest` as you say.

wdyt @devimc?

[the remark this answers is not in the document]
devimc: I think we should check `HypervisorConfig.ConfidentialGuest` instead of messing up the environment with fake files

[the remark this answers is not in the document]
Jakob-Naucke: @devimc how unhappy would you be with setting initrd to the kernel file (not in the config, in the code) when we're on s390x & confidential guest? That way, we wouldn't need fake files, but wouldn't have to rewrite all bits of initrd logic to also trigger for SE.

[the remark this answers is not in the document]
devimc: @Jakob-Naucke  do you think it may confuse developers when debugging SE ? if not, then go ahead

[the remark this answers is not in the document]
Jakob-Naucke: to be peeeerfectly honest I think it might. Let me TAL at how deep the changes would have to be.

Jakob-Naucke:
I've updated some things from the discussions (docs likely in confidential-containers namespace). @jiangliu @Alex-Carter01 while this is mostly an s390x-specific patch, is there something you would like to see from the x86 side?

src/runtime/pkg/katautils/config.go
```diff
@@ -451,11 +451,13 @@ func (h hypervisor) getInitrdAndImage() (initrd string, image string, err error)
 
 	image, errImage := h.image()
 
-	if image != "" && initrd != "" {
+	if h.ConfidentialGuest && h.MachineType == "s390-ccw-virtio" {
```
devimc, line 454: can you use `QemuCCWVirtio` in this package ?

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -211,12 +211,23 @@ function configure_containerd() {
...
-			"/opt/kata/share/defaults/kata-containers/configuration-cc.toml"
+			"$cc_config"
+		if [ "$(uname -m)" == "s390x" ]; then
+			sed -Ei \
+				-e 's#^kernel = .+#kernel = "/opt/kata/share/kata-containers/kata-containers-secure.img"#' \
+				-e '/^image = .+/d' \
+				-e 's/^(# )?confidential_guest = .+/confidential_guest = true/' \
+				-e 's#^shared_fs = .+#shared_fs = "virtio-9p"#' \
+				"$cc_config"
+				# TODO remove 9p workaround once https://bugzilla.linux.ibm.com/show_bug.cgi?id=194746 is resolved
```
devimc, line 227: I guess this is an internal bug...  :confused: :question:

devimc (commented):
thanks @Jakob-Naucke - I've left some comments

src/runtime/pkg/katautils/config.go
Jakob-Naucke: ah yes, we already import `virtcontainers` anyway

tools/packaging/kata-deploy/scripts/kata-deploy.sh
Jakob-Naucke: There are some comments there that are currently set to internal, yes.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -211,12 +211,23 @@ function configure_containerd() {
...
+		else
+			sed -Ei 's#^image = .+#initrd = "/opt/kata/share/kata-containers/kata-containers-initrd.img"#' "$cc_config"
```
jodh-intel, line 229: It might be good to rework slightly so that this path be specified as `"$protimg_kata_dir/kata-containers-initrd.img"`.

jodh-intel: Same comment for the previous `sed` call.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -211,12 +213,23 @@ function configure_containerd() {
...
-			"/opt/kata/share/defaults/kata-containers/configuration-cc.toml"
+			-e 's|^(kernel_params = .+)"|\1 agent.config_file=/etc/kata-containers/agent.toml"|' \
+			-e 's|^[# ]*(service_offload = ).+|\1true|' \
+			"$kata_defaults_dir/configuration-qemu.toml" > \
+			"$cc_config"
+		if [ "$(uname -m)" == "s390x" ]; then
+			sed -Ei \
+				-e "s|^(kernel = ).+|\\1\"$kata_share_dir/kata-containers-secure.img\"|" \
+				-e '/^image = .+/d' \
+				-e 's|^[# ]*(confidential_guest = ).+|\1true|' \
```
Alex-Carter01, line 226: Thanks @Jakob-Naucke. My only question from an x86 perspective is if this confidential_guest flag is s390x specific, or if it could apply for the entire CONFIGURE_CC case?

Jakob-Naucke: @Alex-Carter01 it isn't, but your question got me thinking. When enabled, the use of the platform-specific confidential computing technology is attempted. I hadn't enabled this in general, because on x86, if you're not capable of TDX/SEV, there will be errors, and this script is also supposed to work for a generic try-out setup without a real TEE. Always enabling it on s390x is a slight oversimplification (although TBF no one really has _just_ s390x available); ideally, we'd run a Secure Execution check here. However, we're already checking this [in the runtime](https://github.com/kata-containers/kata-containers/blob/d370604fa5cdc78ecc438ee3c65e2e805c54c160/src/runtime/virtcontainers/hypervisor_s390x.go) _plus_ there's https://github.com/confidential-containers/operator/issues/24 which also seems to want to deal with this.

@bpradipt DYT it is unnecessary to check capabilities in operator, kata-deploy, _and_ runtime?

Alex-Carter01: I see, thank you for the clarification. It makes sense to me then why it is right to leave it here, and then have an equivalent for x86 behind a Secure Execution check if that ends up being necessary.

bpradipt: @Jakob-Naucke checking of capabilities in operator is required as we'll be supporting confidential computing technologies and it's important to have the visibility at the operator layer for life-cycle management of the runtimes. A simple use-case being installing the kata-deploy-cc artefacts for SEV only on capable systems in the cluster.
It's also good to have the capabilities check in the runtime since there will be use-cases where the runtime is executed on single system (without requiring k8s+operator infra).
As for kata-deploy, I'm not sure if having specific CC capability checks will help. For CC case, I would prefer keeping kata-deploy as pure payload, and having the capability check in the operator and the runtime.
For example kata-deploy image for CC on s390x having all required configs for SE  (quay.io/confidential-containers/runtime-payload-s390x). Likewise for SEV, TDX etc.

bpradipt: @Jakob-Naucke following is the high level workflow of the operator:
1. Deploy operator
2. Create CC config. This will detect the eligibility of the nodes in the cluster w.r.to specific TEE. For example if it's an s390x cluster, then it will detect which nodes in the cluster are eligible to run confidential containers using SE. Likewise for SEV and TDX _(note that this functionality is not yet implemented)_
3. Once the operator knows the eligible nodes, it'll perform the required installation tasks on the specific nodes using the kata-deploy payload. Current implementation sets the CONFIGURE_CC env variable and execute the installer script.

Having the CC settings under a single conditional (`CONFIGURE_CC`) is better.  You can as well add CC capability check in kata-deploy in case CONFIGURE_CC is unset, but honestly I'm not sure if it's really necessary to add additional CC checks to kata-deploy. 

Whatever other combinations are needed to determine a true value for `CONFIGURE_CC` can be handled by the operator.

Jakob-Naucke: Thanks @bpradipt. To summarise:
- _For the time being_, on x86_64, `kata-deploy.sh` will install the non-TEE "CC" setup. (In the longer term, we might want to drop that overall maybe? We should avoid anyone running real workloads with the non-TEE demo thinking they had enabled TEE.)
- On s390x, `kata-deploy.sh` will already assume it was invoked on a capable machine.

bpradipt: @Jakob-Naucke yes. Once we have SEV and TDX pieces we'll have a real TEE "CC" setup and can rename the non-TEE "CC" setup accordingly (test/demo etc..)

fidencio:

larrydewey (commented):
Nothing crucial, just grammar comments.

src/runtime/virtcontainers/hypervisor.go
```diff
@@ -549,11 +549,13 @@ func (conf *HypervisorConfig) Valid() error {
 		return fmt.Errorf("Missing kernel path")
 	}
 
-	if conf.ImagePath == "" && conf.InitrdPath == "" {
+	if conf.ConfidentialGuest && conf.HypervisorMachineType == QemuCCWVirtio {
+		if conf.ImagePath != "" || conf.InitrdPath != "" {
+			return fmt.Errorf("Neither of image and initrd path can be set for Secure Execution")
```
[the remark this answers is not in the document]
Jakob-Naucke: Thanks @larrydewey - not a native speaker :sweat_smile:

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash -eu
+# Copyright (c) 2021 IBM Corp.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[ -n "${DEBUG:-}" ] && set -x
+[ "$(uname -m)" = s390x ] || (echo >&2 "Creation of Secure Execution images is currently only supported on s390x." && exit 1)
```
jodh-intel, line 7: Nit: I'd consider prefixing the message with `"ERROR:`".

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
```diff
@@ -0,0 +1,87 @@
...
+tar_path="kata-static.tar.xz"
+
+tarball_content_dir="$PWD/kata-tarball-content"
+protimg_content_dir="$PWD/kata-protimg-content"
+rm -rf "$tarball_content_dir" "$protimg_content_dir"
+mkdir "$tarball_content_dir" "$protimg_content_dir"
+
+pushd "$kata_build_dir"
+for c in qemu shim-v2; do
+	tar xvf kata-static-$c.tar.xz -C "$tarball_content_dir"
```
jodh-intel, line 60: Just checking that there is no architecture in the filenames?

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
```diff
@@ -0,0 +1,87 @@
...
+done
+
+for c in kernel rootfs-initrd; do
+	tar xvf kata-static-$c.tar.xz -C "$protimg_content_dir"
+done
+popd
+
+parmfile="$(mktemp /tmp/cmdline.XXXXXXXXXX)"
```
jodh-intel, line 68: Nit: Hard-coded tempdir. It might be better to say:

```suggestion
parmfile="$(mktemp --suffix=-cmdline)
```

tools/packaging/kata-deploy/scripts/kata-deploy.sh
jodh-intel: Rather than the arch test, you could use the following (which needs an update for `s390x` I think?):

```bash
$ kata-runtime env --json | jq '.Host.AvailableGuestProtections'
```

tools/packaging/kata-deploy/local-build/kata-deploy-secure-execution.sh
jodh-intel: No, I'm just checking that the name of those tar files is correct (and doesn't contain an arch name, as we do for `kata-static-2.3.2-x86_64.tar.xz` for example).

Jakob-Naucke: These are not the ones that would end up in `$kata_build_dir`, that stuff happens here: https://github.com/kata-containers/kata-containers/blob/8f80dffead8066421deaa9c94a2de2e1c5c8fb13/.github/workflows/release.yaml#L129

Jakob-Naucke:
@jodh-intel GitHub won't load the reply field for this particular comment across two browsers, I'll reply here:

> Rather than the arch test, you could use the following (which needs an update for s390x I think?):
> ``` 
> $ kata-runtime env --json | jq '.Host.AvailableGuestProtections'
> ```

This does work with Secure Execution, the only trouble is that this wants `vmlinux.container` and `kata-containers.img`. I'll look into it.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
Jakob-Naucke: Ah -- I could not reply further down because you can only reply in the original thread, which is here.
My original comment (which I've deleted) was:

> > Rather than the arch test, you could use the following (which needs an update for s390x I think?):
> > ```
> > $ kata-runtime env --json | jq '.Host.AvailableGuestProtections'
> > ```
> 
> This does work with Secure Execution, the only trouble is that this wants `vmlinux.container` and `kata-containers.img`. I'll look into it.

and to give an update on that: it does work when setting using the CC config, but that introduces a self-reference because that config is only generated here. So I'd keep it.

Jakob-Naucke:
Ah yes, the grand merge. I'll rebase.

Jakob-Naucke:
plain rebase for static checks

fidencio:
I think this PR is a good example of what can and should be done as when we start working on having a kata-deploy build specific for the CC case.  Although I'm not against having it merged, I'd prefer having a structure discussion on how we'll approach the hiow `kata-deploy`subject and then get pieces & bits of this (if not this entirely) merged into `main` when the time comes.

@Jakob-Naucke, would that be a big problem for you?

jodh-intel:
Ping @Jakob-Naucke.

Jakob-Naucke:
ah yes, had forgotten to reply here -- @jodh-intel just now I've begun branching it out :) Yes, this sounds reasonable. I'm going to bring changes that aren't really related to kata-deploy into separate PRs.

Jakob-Naucke:
Please see #3921, #3923 for the related PRs. Closing.

Issue #3216, opened 2021-12-07
Alpine rootfs is much bigger in Kata 2 -- 113 MB vs 14 MB


skaegi:
We build our own rootfs based on alpine and previously had noted the size of the kata-containers-initrd.img...
Kata 1.12 -- 14 MB
Kata 2.3.0 -- 113 MB

It looks like most of this is coming from /usr ...
```
/go/src/github.com/kata-containers/kata-containers/tools/osbuilder/rootfs-builder/rootfs # du -hs *
2.1M	bin
4.0K	dev
1.3M	etc
4.0K	home
0	init
4.7M	lib
4.0K	media
4.0K	mnt
4.0K	opt
4.0K	proc
4.0K	root
4.0K	run
16M	sbin
4.0K	srv
4.0K	sys
4.0K	tmp
294M	usr
52K	var

/go/src/github.com/kata-containers/kata-containers/tools/osbuilder/rootfs-builder/rootfs # du -hs usr/*
68M	usr/bin
20M	usr/include
91M	usr/lib
76M	usr/libexec
36K	usr/local
20K	usr/sbin
39M	usr/share
1.6M	usr/x86_64-alpine-linux-musl
```

Not sure of all this stuff is really needed or if the agent is statically compiled

liubin:
This commit https://github.com/kata-containers/kata-containers/commit/eb5dd76e9da1331aecce409d956a2520a6e08a78 installs many packages.

fidencio:
> This commit [eb5dd76](https://github.com/kata-containers/kata-containers/commit/eb5dd76e9da1331aecce409d956a2520a6e08a78) installs many packages.

Hmm.  Let me cc @Jakob-Naucke.

We should be able to only install the packages to build the agent, without increasing the final size of the rootfs image.

Jakob-Naucke:
Hi @skaegi! I overlooked that in Alpine specifically, we don't seem to bootstrap an extra rootfs, but we bootstrap the rootfs from the build container itself. I guess for musl agents (the only ones we (can) use in Alpine), building in a container is not strictly necessary (unlike for glibc agents, where the glibc version must match), but an external bootstrap (similar to `debootstrap` or `dnf --installroot`) would be nicer if possible.

Jakob-Naucke:
[...]

PR #3222, opened 2021-12-07
osbuilder: Revert to using apk.static for Alpine

Jakob-Naucke:
#2399 partially reverted #418, missing on returning to bootstrapping a
rootfs with `apk.static` instead of copying the entire root, which can
result in drastically larger (more than 10x) images.

Fixes: #3216
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

PR #3247, opened 2021-12-09
Backport osbuilder: Revert to using apk.static for Alpine

Jakob-Naucke:
Backport of #3222. For this, we're using the existing Alpine 3.13 rather than the upgrade to 3.15.
/cc @fidencio @skaegi

### DIFF
--- a/tools/osbuilder/rootfs-builder/alpine/Dockerfile.in
+++ b/tools/osbuilder/rootfs-builder/alpine/Dockerfile.in
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 ARG IMAGE_REGISTRY=docker.io
-FROM ${IMAGE_REGISTRY}/alpine:3.13.5
+FROM ${IMAGE_REGISTRY}/alpine:3.13
 
 RUN apk update && apk add \
     apk-tools-static \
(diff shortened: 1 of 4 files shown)


cmaf (approved):
@Jakob-Naucke lgtm, but what is the reasoning for using 3.13 vs 3.15 in stable-2.3?

Jakob-Naucke:
@cmaf What I am backporting is the use of `apk.static` rather than copying parts of the container rootfs (this fixes #3216). To make that stable, it is better to pin a version (and a minor version rather than patch level). However, upgrading to a newer version of Alpine is potentially a breaking change (we have seen some problems with Docker versions that we don't officially support in #3222), therefore, it doesn't belong in the backport IMO.

Jakob-Naucke:
E2E:
```
[init] Using Kubernetes version: v1.22.0
[preflight] Running pre-flight checks
error execution phase preflight: [preflight] Some fatal errors occurred:
	[ERROR Port-6443]: Port 6443 is in use
	[ERROR Port-10259]: Port 10259 is in use
	[ERROR Port-10257]: Port 10257 is in use
```
/retest-fedora-crio-k8s-e2e-minimal

Firecracker:
```
20.04: Pulling from library/ubuntu
received unexpected HTTP status: 503 Service Unavailable
```

/retest-fc

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

Issue #1468, opened 2021-02-26
Pattern for directly assigning block backed filesystem volumes


egernst:

jodh-intel:
Hi @egernst - could you add a few words here please? :smile:

PR #3248, opened 2021-12-09
docs: repropose direct-assigned volume

fengwang666:
Fixes: #1468

### DIFF
--- a/docs/design/direct-blk-device-assignment.md
+++ b/docs/design/direct-blk-device-assignment.md
@@ -1,0 +1,253 @@
+# Motivation
+Today, there exist a few gaps between Container Storage Interface (CSI) and virtual machine (VM) based runtimes such as Kata Containers 
+that prevent them from working together smoothly.
+
+First, it’s cumbersome to use a persistent volume (PV) with Kata Containers. Today, for a PV with Filesystem volume mode, Virtio-fs
+is the only way to surface it inside a Kata Container guest VM. But often mounting the filesystem (FS) within the guest operating system (OS) is 
+desired due to performance benefits, availability of native FS features and security benefits over the Virtio-fs mechanism.
+
+Second, it’s difficult if not impossible to resize a PV online with Kata Containers. While a PV can be expanded on the host OS, 
+the updated metadata needs to be propagated to the guest OS in order for the application container to use the expanded volume. 
+Currently, there is not a way to propagate the PV metadata from the host OS to the guest OS without restarting the Pod sandbox.
+
+# Proposed Solution
+
+Because of the OS boundary, these features cannot be implemented in the CSI node driver plugin running on the host OS 
+as is normally done in the runc container. Instead, they can be done by the Kata Containers agent inside the guest OS, 
+but it requires the CSI driver to pass the relevant information to the Kata Containers runtime. 
+An ideal long term solution would be to have the `kubelet` coordinating the communication between the CSI driver and 
+the container runtime, as described in [KEP-2857](https://github.com/kubernetes/enhancements/pull/2893/files). 
+However, as the KEP is still under review, we would like to propose a short/medium term solution to unblock our use case.
+
+The proposed solution is built on top of a previous [proposal](https://github.com/egernst/kata-containers/blob/da-proposal/docs/design/direct-assign-volume.md) 
+described by Eric Ernst. The previous proposal has two gaps:
+
+1. Writing a `csiPlugin.json` file to the volume root path introduced a security risk. A malicious user can gain unauthorized 
+access to a block device by writing their own `csiPlugin.json` to the above location through an ephemeral CSI plugin.  
+
+2. The proposal didn't describe how to establish a mapping between a volume and a kata sandbox, which is needed for 
+implementing CSI volume resize and volume stat collection APIs.
+
+This document particularly focuses on how to address these two gaps.
+
+## Assumptions and Limitations
+1. The proposal assumes that a block device volume will only be used by one Pod on a node at a time, which we believe 
+is the most common pattern in Kata Containers use cases. It’s also unsafe to have the same block device attached to more than 
+one Kata pod. In the context of Kubernetes, the `PersistentVolumeClaim` (PVC) needs to have the `accessMode` as `ReadWriteOncePod`. 
+2. More advanced Kubernetes volume features such as, `fsGroup`, `fsGroupChangePolicy`, and `subPath` are not supported. 
+
+## End User Interface
+
+1. The user specifies a PV as a direct-assigned volume. How a PV is specified as a direct-assigned volume is left for each CSI implementation to decide.
+There are a few options for reference:
+   1. A storage class parameter specifies whether it's a direct-assigned volume. This avoids any lookups of PVC 
+   or Pod information from the CSI plugin (as external provisioner takes care of these). However, all PVs in the storage class with the parameter set 
+   will have host mounts skipped.
+   2. Use a PVC annotation. This approach requires the CSI plugins have `--extra-create-metadata` [set](https://kubernetes-csi.github.io/docs/external-provisioner.html#persistentvolumeclaim-and-persistentvolume-parameters)
+   to be able to perform a lookup of the PVC annotations from the API server. Pro: API server lookup of annotations only required during creation of PV. 
+   Con: The CSI plugin will always skip host mounting of the PV.
+   3. The CSI plugin can also lookup pod `runtimeclass` during `NodePublish`. This approach can be found in the [ALIBABA CSI plugin](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/master/pkg/disk/nodeserver.go#L248).
+2. The CSI node driver delegates the direct assigned volume to the Kata Containers runtime. The CSI node driver APIs need to 
+   be modified to pass the volume mount information and collect volume information to/from the Kata Containers runtime by invoking `kata-runtime` command line commands.
+   * **NodePublishVolume** -- It invokes `kata-runtime direct-volume add --volume-path [volumePath] --mount-info [mountInfo]` 
+   to propagate the volume mount information to the Kata Containers runtime for it to carry out the filesystem mount operation.
+   The `volumePath` is the [target_path](https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1364) in the CSI `NodePublishVolumeRequest`.
+   The `mountInfo` is a serialized JSON string. 
+   * **NodeGetVolumeStats** -- It invokes `kata-runtime direct-volume stats --volume-path [volumePath]` to retrieve the filesystem stats of direct-assigned volume.
+   * **NodeExpandVolume** -- It invokes `kata-runtime direct-volume resize --volume-path [volumePath] --size [size]` to send a resize request to the Kata Containers runtime to
+   resize the direct-assigned volume.
+   * **NodeStageVolume/NodeUnStageVolume** -- It invokes `kata-runtime direct-volume remove --volume-path [volumePath]` to remove the persisted metadata of a direct-assigned volume.
+
+The `mountInfo` object is defined as follows:
+```Golang
+type MountInfo struct {
+    // The type of the volume (ie. block)
+    VolumeType string `json:"volume-type"`
+    // The device backing the volume.
+    Device string `json:"device"`
+    // The filesystem type to be mounted on the volume.
+    FsType string `json:"fstype"`
+    // Additional metadata to pass to the agent regarding this volume.
+    Metadata map[string]string `json:"metadata,omitempty"`
+    // Additional mount options.
+    Options []string `json:"options,omitempty"`
+}
+```
+Notes: given that the `mountInfo` is persisted to the disk by the Kata runtime, it shouldn't container any secrets (such as SMB mount password).
+
+## Implementation Details
+
+### Kata runtime
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal, here we propose that the CSI node driver passes the mount information to 
+the Kata Containers runtime through a new `kata-runtime` commandline command. The `kata-runtime` then writes the mount 
+information to a `mount-info.json` file in a predefined location (`/run/kata-containers/shared/direct-volumes/[volume_path]/`).
+
+When the Kata Containers runtime starts a container, it verifies whether a volume mount is a direct-assigned volume by checking 
+whether there is a `mountInfo` file under the computed Kata `direct-volumes` directory. If it is, the runtime parses the `mountInfo` file, 
+updates the mount spec with the data in `mountInfo`. The updated mount spec is then passed to the Kata agent in the guest VM together
+with other mounts. The Kata Containers runtime also creates a file named by the sandbox id under the `direct-volumes/[volume_path]/` 
+directory. The reason for adding a sandbox id file is to establish a mapping between the volume and the sandbox using it. 
+Later, when the Kata Containers runtime handles the `get-stats` and `resize` commands, it uses the sandbox id to identify 
+the endpoint of the corresponding `containerd-shim-kata-v2`.
+
+### containerd-shim-kata-v2 changes
+`containerd-shim-kata-v2` provides an API for sandbox management through a Unix domain socket. Two new handlers are proposed: `/direct-volume/stats` and `/direct-volume/resize`:
+
+Example:
+
+```bash
+$ curl --unix-socket "$shim_socket_path" -I -X GET 'http://localhost/direct-volume/stats/[urlSafeVolumePath]'
+$ curl --unix-socket "$shim_socket_path" -I -X POST 'http://localhost/direct-volume/resize' -d '{ "volumePath"": [volumePath], "Size": "123123" }'
+```
+
+The shim then forwards the corresponding request to the `kata-agent` to carry out the operations inside the guest VM. For `resize` operation, 
+the Kata runtime also needs to notify the hypervisor to resize the block device (e.g. call `block_resize` in QEMU). 
+
+### Kata agent changes
+
+The mount spec of a direct-assigned volume is passed to `kata-agent` through the existing `Storage` GRPC object. 
+Two new APIs and three new GRPC objects are added to GRPC protocol between the shim and agent for resizing and getting volume stats:
+```protobuf
+
+rpc GetVolumeStats(VolumeStatsRequest) returns (VolumeStatsResponse);
+rpc ResizeVolume(ResizeVolumeRequest) returns (google.protobuf.Empty);
+
+message VolumeStatsRequest {
+// The volume path on the guest outside the container
+    string volume_guest_path = 1;
+}
+
+message ResizeVolumeRequest {
+// Full VM guest path of the volume (outside the container)
+    string volume_guest_path = 1;
+    uint64 size = 2;
+}
+
+// This should be kept in sync with CSI NodeGetVolumeStatsResponse (https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto)
+message VolumeStatsResponse {
+   // This field is OPTIONAL.
+   repeated VolumeUsage usage = 1;
+   // Information about the current condition of the volume.
+   // This field is OPTIONAL.
+   // This field MUST be specified if the VOLUME_CONDITION node
+   // capability is supported.
+   VolumeCondition volume_condition = 2;
+}
+message VolumeUsage {
+   enum Unit {
+      UNKNOWN = 0;
+      BYTES = 1;
+      INODES = 2;
+   }
+   // The available capacity in specified Unit. This field is OPTIONAL.
+   // The value of this field MUST NOT be negative.
+   uint64 available = 1;
+
+   // The total capacity in specified Unit. This field is REQUIRED.
+   // The value of this field MUST NOT be negative.
+   uint64 total = 2;
+
+   // The used capacity in specified Unit. This field is OPTIONAL.
+   // The value of this field MUST NOT be negative.
+   uint64 used = 3;
+
+   // Units by which values are measured. This field is REQUIRED.
+   Unit unit = 4;
+}
+
+// VolumeCondition represents the current condition of a volume.
+message VolumeCondition {
+
+   // Normal volumes are available for use and operating optimally.
+   // An abnormal volume does not meet these criteria.
+   // This field is REQUIRED.
+   bool abnormal = 1;
+
+   // The message describing the condition of the volume.
+   // This field is REQUIRED.
+   string message = 2;
+}
+
+```
+
+### Step by step walk-through
+
+Given the following definition:
+```YAML
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: app
+spec:
+  runtime-class: kata-qemu
+  containers:
+  - name: app
+    image: centos
+    command: ["/bin/sh"]
+    args: ["-c", "while true; do echo $(date -u) >> /data/out.txt; sleep 5; done"]
+    volumeMounts:
+    - name: persistent-storage
+      mountPath: /data
+  volumes:
+  - name: persistent-storage
+    persistentVolumeClaim:
+      claimName: ebs-claim
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  annotations:
+    skip-hostmount: "true"
+  name: ebs-claim
+spec:
+  accessModes:
+    - ReadWriteOncePod
+  volumeMode: Filesystem
+  storageClassName: ebs-sc
+  resources:
+    requests:
+      storage: 4Gi
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: ebs-sc
+provisioner: ebs.csi.aws.com
+volumeBindingMode: WaitForFirstConsumer
+parameters:
+  csi.storage.k8s.io/fstype: ext4
+
+```
+Let’s assume that changes have been made in the `aws-ebs-csi-driver` node driver.
+
+**Node publish volume**
+1. In the node CSI driver, the `NodePublishVolume` API invokes: `kata-runtime direct-volume add --volume-path "/kubelet/a/b/c/d/sdf" --mount-info "{\"Device\": \"/dev/sdf\", \"fstype\": \"ext4\"}"`.
+2. The `Kata-runtime` writes the mount-info JSON to a file called `mountInfo.json` under `/run/kata-containers/shared/direct-volumes/kubelet/a/b/c/d/sdf`.
+
+**Node unstage volume**
+1. In the node CSI driver, the `NodeUnstageVolume` API invokes: `kata-runtime direct-volume remove --volume-path "/kubelet/a/b/c/d/sdf"`.
+2. Kata-runtime deletes the directory `/run/kata-containers/shared/direct-volumes/kubelet/a/b/c/d/sdf`.
+
+**Use the volume in sandbox**
+1. Upon the request to start a container, the `containerd-shim-kata-v2` examines the container spec,
+and iterates through the mounts. For each mount, if there is a `mountInfo.json` file under `/run/kata-containers/shared/direct-volumes/[mount source path]`,
+it generates a `storage` GRPC object after overwriting the mount spec with the information in `mountInfo.json`.
+2. The shim sends the storage objects to kata-agent through TTRPC.
+3. The shim writes a file with the sandbox id as the name under `/run/kata-containers/shared/direct-volumes/[mount source path]`.
+4. The kata-agent mounts the storage objects for the container.
+
+**Node expand volume**
+1. In the node CSI driver, the `NodeExpandVolume` API invokes: `kata-runtime direct-volume resize –-volume-path "/kubelet/a/b/c/d/sdf" –-size 8Gi`.
+2. The Kata runtime checks whether there is a sandbox id file under the directory `/run/kata-containers/shared/direct-volumes/kubelet/a/b/c/d/sdf`.
+3. The Kata runtime identifies the shim instance through the sandbox id, and sends a GRPC request to resize the volume.
+4. The shim handles the request, asks the hypervisor to resize the block device and sends a GRPC request to Kata agent to resize the filesystem.
+5. Kata agent receives the request and resizes the filesystem.
+
+**Node get volume stats**
+1. In the node CSI driver, the `NodeGetVolumeStats` API invokes: `kata-runtime direct-volume stats –-volume-path "/kubelet/a/b/c/d/sdf"`.
+2. The Kata runtime checks whether there is a sandbox id file under the directory `/run/kata-containers/shared/direct-volumes/kubelet/a/b/c/d/sdf`.
+3. The Kata runtime identifies the shim instance through the sandbox id, and sends a GRPC request to get the volume stats.
+4. The shim handles the request and forwards it to the Kata agent.
+5. Kata agent receives the request and returns the filesystem stats.
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -11,6 +11,7 @@ Kata Containers design documents:
 - [`Inotify` support](inotify.md)
 - [Metrics(Kata 2.0)](kata-2-0-metrics.md)
 - [Design for Kata Containers `Lazyload` ability with `nydus`](kata-nydus-design.md)
+- [Design for direct-assigned volume](direct-blk-device-assignment.md)
 
 ---
 


fengwang666:
Rendered link: https://github.com/fengwang666/kata-containers/blob/62a8f6f1dd814df2c8e40ae260de1f76bbca2a8c/docs/design/proposals/direct-blk-device-assignment.md

Thanks @bergwolf, @egernst, Deep Debroy and Shuo Chen for the valuable input!

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
+# Background
+Today, there exists a few gaps between Container Storage Interface (CSI) and virtual-machine(VM) based runtimes like Kata Container
```
liubin, line 2: `Kata Container` -> `Kata Containers`

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+
+## End User Interface
+If a PV is intended to be consumed by a VM-based runtime, the user can specify this by adding an annotation called 
+`skip-hostmount` to the persistent volume claim (PVC). Example:
+```
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  annotations:
+    skip-hostmount: "true"
```
liubin, line 49: Annotations used by Kata Containers should have `io.katacontainers.` prefix: https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/pkg/annotations/annotations.go#L9

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Assumptions
+The proposal assumes that a block device volume will only be used by one Pod on a node at a time, which we believe 
+is the more common pattern in the Kata container. It’s rare and unsafe to have the same block device attached to more than 
+one Kata Pods. In the context of Kubernetes, the PersistentVolumeClaim (PVC) needs to have accessMode as `ReadWriteOncePod`. 
+
+## End User Interface
+If a PV is intended to be consumed by a VM-based runtime, the user can specify this by adding an annotation called 
+`skip-hostmount` to the persistent volume claim (PVC). Example:
+```
```
liubin, line 44: It's better to add `YAML`  after the triple backquotes to set the file extension to help a syntax highlight.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+containerd-shim-kata-v2 provides an API for sandbox management through a unix domain socket. Two new handlers are proposed:
+```
+New handlers:
+/fs-stats
+/fs-resize
+
+Example:
+
+curl —unix-socket <shim socket path> -I \
+-X PUT ‘http://localhost/fs-stats' \
```
liubin, line 131: Suggest changing `PUT` to `GET` for read-only request.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+-d ‘{ “block-device”: “/dev/sda” }’ 
+
+curl —unix-socket <shim socket path> -I \
+-X PUT ‘http://localhost/fs-stats' \
```
liubin, line 135: `fs-stats` should be `fs-resize`?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+**Use the volume in sandbox**
+1. Upon the request to start a sandbox, the containerd-shim-kata-v2(shim for abbreviation) examines the container spec 
+and iterates through the mounts, for each mount, if the device name maps to a directory under `/run/kata-containers/shared/direct-block-devices`, 
+it then generates a `storage` grpc object combining the mount info file and the mount attributes in the container spec.
+2. The shim sends the storage objects to kata-agent through ttrpc.
+3. The shim writes a file with the sandbox id as the name under `/run/kata-containers/shared/direct-block-devices/[device_name]`.
+4. The kata-agent mounts the storage objects into the containers.
+
+**Node expand volume**
+1. In the node csi driver, the NodePublishVolume API invokes: `kata-runtime fs resize-volume –-device-path /dev/sdf –-size 8Gi`
```
liubin, line 232: NodePublishVolume  -> NodeExpandVolume ?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+2. The kata-runtime finds if the directory `/run/kata-containers/shared/direct-block-devices/sdf` exists. 
+If yes, it then goes into the directory, finds the sandbox id file and reads the file name.
+3. The kata-runtime identifies the shim instance through the sandbox id, and sends a grpc request to resize the volume.
+4. The shim handles the request, asks the hypervisor to resize the block device and send a grpc request to kata agent to resize the filesystem.
+5. Kata agent receives the request and resizes the filesystem.
+
+**Node get volume stats**
+1. In the node csi driver, the NodePublishVolume API invokes: `kata-runtime fs stats –-device-path /dev/sdf`
```
liubin, line 240: NodePublishVolume  -> NodeGetVolumeStats ?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal. Here we propose that the CSI node driver passes the mount information to 
+the kata-runtime through a new kata-runtime command called `add-blk-mount-info`. The kata-runtime then write the mount 
+information to a `“`mount-info.json`”` file in a hardcoded location (e.g. `/run/kata-containers/shared/direct-block-devices/[block_dev_name]/`).
+
+When a Kata Pod sandbox is started, the containerd-kata-shim-v2 checks whether there is any mount identified 
+by its source path in the container specification that has a mount info file in the `direct-block-devices` directory. 
+If it exists, the containerd-kata-shim-v2 parses the mount info file and propagates it to the kata agent to carry out the mount. 
+Lastly,the containerd-kata-shim-v2 creates a file named by the sandbox id under the `direct-block-devices/[block_dev_name]/`
```
liubin, line 116: Question: is the file(`a file named by the sandbox id`) empty and does only the file name has its meaning?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+from CSI to the Kata container runtime. 
+
+### Custom CSI node driver changes
+In the `skip-hostmount` mode, CSI node driver defers the filesystem mount operation to the container runtime. 
+The implementation of a few node driver APIs needs to be updated.
+
+* **NodePublishVolume** -- It propagates the mount information to the container runtime for it to carry out filesystem mount operation.
+* **NodeGetVolumeStats** -- Since the container runtime mounts the filesystem, it needs to forward the request to the container runtime.
+* **NodeExpandVolume** -- The CSI cannot implement this API alone. It needs to propagate the new size to the container runtime so that the filesystem metadata can be refreshed accordingly.
+* **NodeStageVolume/NodeUnStageVolume** -- Invoke a kata-runtime handler to clean up the stale mount information.
```
ddebroy, line 73: Typical block device based CSI plugins (like EBS, Azure, GCE PD, Ceph, etc) will mount the PV to a global mount path in the host during NodeStage. It should be stated here that this global mount of the PV will need to be skipped during NodeStage. If the CSI plugin knows host mount should be skipped for the PV using PVC annotation or storage class parameters passed as volume context, this becomes trivial. If the CSI plugin decides to lookup pod runtimeclass to determine whether to skip host mount, then the CSI plugin should wait until NodePublishVolume to decide whether to perform/skip the host mount.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
...
+
+```
+kata-runtime fs —help
+NAME:
+kata-runtime fs [subcommand] [options]— interact with mounted filesystem volume
+
+SUBCOMMANDS:
+ - add-blk-mount-info <mountinfo> -- add the mount info of a volume to the Kata runtime.
+ - reset-blk-mount-info <host-blk-device-name> –- clean up the mount info directory under kata runtime
+ - get-stats <host-blk-device-name> — show status and usage details for the mounted filesystem backed by the given block device, <host-blk-device-name>.
```
ddebroy, line 88: Can this also specify a return json to map the fields associated with CSI get stats (Condition, Size, Inode details, etc)?

ddebroy (commented):
Looks great! I added a few suggestions to clarify certain CSI details.

docs/design/proposals/direct-blk-device-assignment.md
fengwang666: Thanks!

docs/design/proposals/direct-blk-device-assignment.md
fengwang666: Exactly.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,246 @@
+# Background
+Today, there exists a few gaps between Container Storage Interface (CSI) and virtual-machine(VM) based runtimes like Kata Container 
+that prevent them from working together smoothly.
+
+First, it’s cumbersome to use a persistent volume (PV) with Kata Containers. Unlike in runc containers where the CSI plugin 
+provisions the PV and sets up the filesystem, due to the OS boundary, a PV can only work with the Kata containers with 
+Block volume mode. Individual workload applications need to perform the filesystem mount inside the guest OS, which is burdensome.
```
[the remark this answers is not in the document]
fengwang666: Reworded accordingly. Thanks!

docs/design/proposals/direct-blk-device-assignment.md
fengwang666: AFAIK, it's just the [interface](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/5d782870e3ee2e56e36202362d7b65d54ce8f866/vendor/github.com/container-storage-interface/spec/lib/go/csi/csi.pb.go#L4115-L4126) specified by the CSI, right?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+The volume stas is defined as below following the [csi spec](https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go).
+```Golang
+type NodeGetVolumeStats struct {
+	// This field is OPTIONAL.
+	Usage []*VolumeUsage `json:"usage,omitempty"`
+	// Information about the current condition of the volume.
+	// This field is OPTIONAL.
+	// This field MUST be specified if the VOLUME_CONDITION node
+	// capability is supported.
+	VolumeCondition      *VolumeCondition `json:"volume_condition,omitempty"`
```
bergwolf, line 120: Missing definition of `VolumeUsage` and `VolumeCondition`

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+parameters:
+  csi.storage.k8s.io/fstype: ext4
+
+```
+Let’s assume that changes have been made in aws-ebs-csi-driver node driver.
+
+**Node publish volume**
+
+1. In the node csi driver, the NodePublishVolume API invokes: `kata-runtime fs add-blk-mount-info “{device: /dev/sdf,fstype: ext4}”`.
+2. Kata-runtime writes the json to a file called `mount-info.json` with the above json under `/run/kata-containers/shared/direct-block-devices/sdf/`.
```
bergwolf, line 234: How is the `sdf` part of `/run/kata-containers/shared/direct-block-devices/sdf/` is chosen? IWO, we need to make sure the volume name does not conflict between different volumes. Are you mapping from `/dev/sdf` to `/direct-block-devices/sdf/`?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+**Use the volume in sandbox**
+1. Upon the request to start a sandbox, the containerd-shim-kata-v2(shim for abbreviation) examines the container spec 
+and iterates through the mounts, for each mount, if the device name maps to a directory under `/run/kata-containers/shared/direct-block-devices`, 
+it then generates a `storage` grpc object combining the mount info file and the mount attributes in the container spec.
+2. The shim sends the storage objects to kata-agent through ttrpc.
+3. The shim writes a file with the sandbox id as the name under `/run/kata-containers/shared/direct-block-devices/[device_name]`.
+4. The kata-agent mounts the storage objects into the containers.
+
+**Node expand volume**
+1. In the node csi driver, the NodeExpandVolume API invokes: `kata-runtime fs resize-volume –-device-path /dev/sdf –-size 8Gi`
```
bergwolf, line 250: All other commands are using `/dev/sdf` as volume identifier. The `--device-path` parameter does not seem necessary.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
+# Background
+Today, there exists a few gaps between Container Storage Interface (CSI) and virtual-machine(VM) based runtimes like Kata Containers 
+that prevent them from working together smoothly.
+
+First, it’s cumbersome to use a persistent volume (PV) with Kata Containers. Today, for a PV with Filesystem volume mode, `virtiofs`
+is the only way to surface it insider a Kata guest VM. But often mounting the filesystem (FS) within the guest environment is
```
jodh-intel, line 6: s/insider/inside/

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
+# Background
+Today, there exists a few gaps between Container Storage Interface (CSI) and virtual-machine(VM) based runtimes like Kata Containers
```
jodh-intel, line 2: s/virtual-machine(VM)/virtual machine (VM)/

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
+# Background
+Today, there exists a few gaps between Container Storage Interface (CSI) and virtual-machine(VM) based runtimes like Kata Containers 
+that prevent them from working together smoothly.
+
+First, it’s cumbersome to use a persistent volume (PV) with Kata Containers. Today, for a PV with Filesystem volume mode, `virtiofs`
```
jodh-intel, line 5: You can say "Virtio-fs" as the spell check dictionary knows about that (https://github.com/kata-containers/tests/blob/main/cmd/check-spelling/data/acronyms.txt).

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+Second, it’s difficult if not impossible to resize a PV online with Kata Containers. While a PV can be expanded on the host OS, 
+the updated metadata needs to be propagated to the guest OS in order for the application container to use the expanded volume. 
+Currently, there does not exist a way to propagate the PV metadata from the host OS to the guest OS without restarting the sandbox.
+
+Our use cases at Databricks need two features in the Kata Containers: support PV Filesystem volume mode with native FS types and expand a PV online.
+
+# Proposed Solution
+
+Because of the OS boundary, these features cannot be implemented in the CSI driver plugin running on the host OS 
+as is normally done in the runc container. Instead, they can be done by the kata agent inside the guest OS,
```
jodh-intel, line 18: s/kata/Kata/

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+the updated metadata needs to be propagated to the guest OS in order for the application container to use the expanded volume. 
+Currently, there does not exist a way to propagate the PV metadata from the host OS to the guest OS without restarting the sandbox.
+
+Our use cases at Databricks need two features in the Kata Containers: support PV Filesystem volume mode with native FS types and expand a PV online.
+
+# Proposed Solution
+
+Because of the OS boundary, these features cannot be implemented in the CSI driver plugin running on the host OS 
+as is normally done in the runc container. Instead, they can be done by the kata agent inside the guest OS, 
+but it requires the CSI driver to pass the relevant information to the Kata container runtime.
```
jodh-intel, line 19: s/Kata container/Kata Containers/

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+The implementation of a few node driver APIs needs to be updated.
+
+* **NodePublishVolume** -- It propagates the mount information to the container runtime for it to carry out filesystem mount operation.
+* **NodeGetVolumeStats** -- Since the container runtime mounts the filesystem, it needs to forward the request to the container runtime.
+* **NodeExpandVolume** -- The CSI cannot implement this API alone. It needs to propagate the new size to the container runtime so that the filesystem metadata can be refreshed accordingly.
+* **NodeStageVolume/NodeUnStageVolume** -- Invoke a kata-runtime handler to clean up the stale mount information.
+Typical block device based CSI plugins (like EBS, Azure, GCE PD, Ceph, etc) 
+mounts the PV to a global mount path in the host during NodeStageVolume, which can be skipped in this case.
+
+### Kata-runtime changes
```
jodh-intel, line 81: s/Kata-runtime/Kata Containers runtime/ (or just "Runtime changes").

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+The host agent (CSI node driver, as an example) will call into kata-runtime. Kata-runtime behaves as the API for
```
jodh-intel, line 82: s/kata-runtime/the Kata Containers runtime/

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+    Device string json:"device,omitempty"
+    // FsType: filesystem that needs to be used to mount the storage inside the VM
+    FsType string json:"fsType,omitempty"
+    // metadata field for easy extension
+    Metadata map[string]string json:"metadata,omitempty"
+    // Options: additional options that might be needed to mount the storage filesystem
+    Options []string json:"options,omitempty"
+}
+```
+The volume stas is defined as below following the [csi spec](https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go).
```
jodh-intel, line 111: "stas" => "state" or "struct"?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+	VolumeCondition      *VolumeCondition `json:"volume_condition,omitempty"`
+}
+```
+
+#### Quick notes on implementation
+
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal. Here we propose that the CSI node driver passes the mount information to 
+the kata-runtime through a new kata-runtime command called `add-blk-mount-info`. The kata-runtime then write the mount 
+information to a `“`mount-info.json`”` file in a hardcoded location (e.g. `/run/kata-containers/shared/direct-block-devices/[block_dev_name]/`).
```
jodh-intel, line 129: Please don't use "unicode quotes".

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+	// This field MUST be specified if the VOLUME_CONDITION node
+	// capability is supported.
+	VolumeCondition      *VolumeCondition `json:"volume_condition,omitempty"`
+}
+```
+
+#### Quick notes on implementation
+
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal. Here we propose that the CSI node driver passes the mount information to
```
jodh-intel, line 127: The first two sentences here should be linked into a single sentence I think?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+#### Quick notes on implementation
+
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal. Here we propose that the CSI node driver passes the mount information to 
+the kata-runtime through a new kata-runtime command called `add-blk-mount-info`. The kata-runtime then write the mount 
+information to a `“`mount-info.json`”` file in a hardcoded location (e.g. `/run/kata-containers/shared/direct-block-devices/[block_dev_name]/`).
+
+When a Kata Pod sandbox is started, the containerd-kata-shim-v2 checks whether there is any mount identified
```
jodh-intel, line 131: `containerd-kata-shim-v2` or "Kata Containers shimv2 runtime".

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+containerd-shim-kata-v2 provides an API for sandbox management through a unix domain socket. Two new handlers are proposed:
+```
+New handlers:
+/fs-stats
+/fs-resize
+
+Example:
+
+curl —unix-socket <shim socket path> -I \
+-X GET ‘http://localhost/fs-stats' \
```
jodh-intel, line 149: Please don't use unicode single quotes.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+### containerd-shim-kata-v2 changes
+containerd-shim-kata-v2 provides an API for sandbox management through a unix domain socket. Two new handlers are proposed:
+```
+New handlers:
+/fs-stats
+/fs-resize
+
+Example:
+
+curl —unix-socket <shim socket path> -I \
```
jodh-intel, line 148: Please don't fold lines - let GitHub's markdown renderer handle that.

Also, please put a `$ ` shell prompt at the start of every command and wrap those commands inside a ` ```bash ` block.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+-X GET ‘http://localhost/fs-stats' \
+-d ‘{ “block-device”: “/dev/sda” }’ 
+
+curl —unix-socket <shim socket path> -I \
```
jodh-intel, line 152: Unicode dashes as well I think?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+-X _PUT_ ‘http://localhost/fs-resize' \
+-d ‘{ “block-device”: “/dev/sda”,
+“volume-size”: “123123”
```
jodh-intel, line 155: Abusing this command could cause real problems for the container I suspect.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Kata-agent changes
+
+For mounting a filesystem on the block device volume, the information can be passed to Kata-agent through the existing 
+`Storage` grpc object. Two new grpc request objects are added to the Kata-agent grpc interface for 
+resizing and getting volume stats:
+```Golang
+type ResizeFileSystemRequest struct {
+  mountPoint string
+  size       int64
```
jodh-intel, line 171: In the example above you've specified an actual (new) volume side so this isn't a delta value. In other words, if a volume was _reduced_ in size, the `size`/`volume-size` would still be positive right? In which case, why is this not `uint64`?

Also, please specify the units for this struct member.

jodh-intel (commented):
Thanks for raising @fengwang666. A few comments.

I think this needs a lot more eyes on it so,

/cc @amshinde, @mcastelino, @egernst, @dgibson.

jodh-intel:
Any update on this @fengwang666?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal. Here we propose that the CSI node driver passes the mount information to 
+the kata-runtime through a new kata-runtime command called `add-blk-mount-info`. The kata-runtime then write the mount 
+information to a `“`mount-info.json`”` file in a hardcoded location (e.g. `/run/kata-containers/shared/direct-block-devices/[block_dev_name]/`).
+
+When a Kata Pod sandbox is started, the containerd-kata-shim-v2 checks whether there is any mount identified 
+by its source path in the container specification that has a mount info file in the `direct-block-devices` directory. 
+If it exists, the containerd-kata-shim-v2 parses the mount info file and propagates it to the kata agent to carry out the mount. 
+Lastly,the containerd-kata-shim-v2 creates a file named by the sandbox id under the `direct-block-devices/[block_dev_name]/`
```
yibozhuang, line 134: s/Lastly,the/Lastly, the/

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+volumeBindingMode: WaitForFirstConsumer
+parameters:
+  csi.storage.k8s.io/fstype: ext4
+
+```
+Let’s assume that changes have been made in aws-ebs-csi-driver node driver.
+
+**Node publish volume**
+
+1. In the node csi driver, the NodePublishVolume API invokes: `kata-runtime fs add-blk-mount-info “{device: /dev/sdf,fstype: ext4}”`.
```
yibozhuang, line 233: This approach would require the CSI driver to exec and invoke the `kara-runtime` binary and also need to mount the `/run/vc/sbs` sandbox path inside of the CSI driver container correct? I guess if we were to have the CSI driver directly talk to the shim via UDS at `shim-monitor.sock` then it would need to know the sandbox ID? Do we have an approach where we can still directly invoke the shim APIs without needing to do exec and call `kata-runtime`?

fengwang666: CSI driver needs to invoke kata-runtime binary, but it doesn't need to mount `/run/vc/sbs` path.

yibozhuang: `kata-runtime` binary calls the shimclient correct according to: https://github.com/kata-containers/kata-containers/pull/3406/files#diff-48ea539a2381b43645d5e2ae42a582d935385617661359fbabc88affd6e76746R21

so if we deploy the CSI driver as a DaemonSet in K8s, wouldn't the CSI driver need to have something like
```
- hostPath:
     path: /run/vc/sbs
     type: Directory
   name: runtime-monitor
```

in the pod spec in order to connect to the `shim-monitor.sock`?

fengwang666: @yibozhuang That's correct. The socket needs to be visible inside the csi node driver container.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+
+2. The proposal doesn’t describe how to establish a mapping between a volume and a kata sandbox, which is needed for 
+implementing CSI volume resize and volume stat collection APIs.
+
+This document particularly focuses on how to address these two gaps.
+
+## Assumptions
+The proposal assumes that a block device volume will only be used by one Pod on a node at a time, which we believe 
+is the more common pattern in the Kata container. It’s rare and unsafe to have the same block device attached to more than 
+one Kata Pods. In the context of Kubernetes, the PersistentVolumeClaim (PVC) needs to have accessMode as `ReadWriteOncePod`.
```
ddebroy, line 38: Can we also call out under Assumptions that Kubernetes pod specs that specify [1] fsGroup, [2] fsGroupChangePolicy and [3] subPaths as part of their container mounts won't be supported in the context of this proposal?

docs/design/proposals/direct-blk-device-assignment.md
ddebroy: ok so the assumption here is that [1] this will be identical to the CSI get stats fields and [2] If CSI stats evolves with other fields, the structs/output of this will change correspondingly as well, correct?

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+```Golang
+type MountInfo struct {
+    // Device: source device path for this volume
+    Device string json:"device,omitempty"
+    // FsType: filesystem that needs to be used to mount the storage inside the VM
+    FsType string json:"fsType,omitempty"
+    // metadata field for easy extension
+    Metadata map[string]string json:"metadata,omitempty"
+    // Options: additional options that might be needed to mount the storage filesystem
+    Options []string json:"options,omitempty"
```
ddebroy, line 108: Given this information gets persisted on disk as part of the implementation, how about a specific note here indicating `options` should not include any secrets (that may have been passed through CSI). For example, SMB mount password.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+Let’s assume that changes have been made in aws-ebs-csi-driver node driver.
+
+**Node publish volume**
+
+1. In the node csi driver, the NodePublishVolume API invokes: `kata-runtime fs add-blk-mount-info “{device: /dev/sdf,fstype: ext4}”`.
+2. Kata-runtime writes the json to a file called `mount-info.json` with the above json under `/run/kata-containers/shared/direct-block-devices/sdf/`.
+
+**Node unstage volume**
+
+1. In the node csi driver, the NodeUnstageVolume API invokes: `kata-runtime fs reset-blk-mount-info "/dev/sdf"`.
```
ddebroy, line 238: This should happen during Node stage volume as well (to clean up any stale mountinfo persisted earlier but that may not have been cleaned up if the host crashed/rebooted ungracefully).

fidencio:

c3d:
@fengwang666 I assume you are aware of https://github.com/kubernetes/enhancements/pull/2893?

jodh-intel:
Any update on this @fengwang666?

fengwang666:
> Any update on this @fengwang666?

@jodh-intel The implementation is in PR. Once that's merged, I'll update the doc to the latest.

bergwolf:
Let's put `wip` label on this and wait for https://github.com/kata-containers/kata-containers/pull/3406 to be merged first.

liubin:
@fengwang666 #3406 has been merged, so can you update this documentation?

docs/design/proposals/direct-blk-device-assignment.md
liubin: As @jodh-intel said, there are some full-width double/single quotes, you'd better replace the whole doc.

@jodh-intel maybe feng is using a Chinese input soft(IME), to switch between English/Chinese, usually, one touch of SHIFT key is used, to input a double quote, SHIFT is needed too, therefore, it is easy to mix Chinese characters when entering English.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,264 @@
...
+2. Kata-runtime deletes the directory `/run/kata-containers/shared/direct-block-devices/sdf/`.
+
+**Use the volume in sandbox**
+1. Upon the request to start a sandbox, the containerd-shim-kata-v2(shim for abbreviation) examines the container spec
```
liubin, line 242: `start a sandbox` or `star a container`?

fengwang666:
@liubin @jodh-intel @bergwolf The doc has been updated. PTAL.

liubin:
Hi  @fengwang666 

> [static-checks.sh:709] ERROR: Document docs/design/proposals/direct-blk-device-assignment.md is not referenced

You can add a link in for example `docs/design/README.md`.

docs/design/proposals/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,253 @@
...
+```
+Notes: given that the `mountInfo` is persisted to the disk by the Kata runtime, it shouldn't container any secrets (such as SMB mount password).
+
+## Implementation Details
+
+### Kata runtime
+Instead of the CSI node driver writing the mount info into a `csiPlugin.json` file under the volume root, 
+as described in the original proposal, here we propose that the CSI node driver passes the mount information to 
+the Kata Containers runtime through a new `kata-runtime` commandline command. The `kata-runtime` then writes the mount 
+information to a `mount-info.json` file in a predefined location (`/run/kata-containers/shared/direct-volumes/[volume_path]/`).
```
liubin, line 84: `mount-info.json` -> `mountInfo.json`

arronwy:

c3d (approved):
@fengwang666 Now that the corresponding changes have been merged, I believe that the doc should be updated as well.

docs/design/proposals/direct-blk-device-assignment.md
c3d: @fengwang666 Could you please answer @ddebroy's question above? Thanks ;-)

fengwang666: @ddebroy Sorry for missing this comment. Yes. The assumptions are correct. @c3d Thanks for reminding me!

fengwang666:
@liubin Please take another look. Thanks!

docs/design/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,253 @@
...
+the endpoint of the corresponding `containerd-shim-kata-v2`.
+
+### containerd-shim-kata-v2 changes
+`containerd-shim-kata-v2` provides an API for sandbox management through a Unix domain socket. Two new handlers are proposed: `/direct-volume/stats` and `/direct-volume/resize`:
+
+```bash
+Example:
+
+$ curl —unix-socket <shim socket path> -I -X GET 'http://localhost/direct-volume/stats/[urlSafeVolumePath]'
+$ curl —unix-socket <shim socket path> -I -X POST 'http://localhost/direct-volume/resize' -d '{ "volumePath"": [volumePath], "Size": "123123" }'
```
jodh-intel, line 101:
```suggestion
$ curl --unix-socket $shim_socket_path" -I -X GET 'http://localhost/direct-volume/stats/[urlSafeVolumePath]'
$ curl --unix-socket "$shim_socket_path" -I -X POST 'http://localhost/direct-volume/resize' -d '{ "volumePath"": [volumePath], "Size": "123123" }'
```

docs/design/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+The shim then forwards the corresponding request to the `kata-agent` to carry out the operations inside the guest VM. For `resize` operation, 
+the Kata runtime also needs to notify the hypervisor to resize the block device (e.g. call `block_resize` in QEMU). 
+
+### Kata-agent changes
```
jodh-intel, line 107:
```suggestion
### Kata agent changes
```

docs/design/direct-blk-device-assignment.md
```diff
@@ -0,0 +1,253 @@
...
+directory. The reason for adding a sandbox id file is to establish a mapping between the volume and the sandbox using it. 
+Later, when the Kata Containers runtime handles the `get-stats` and `resize` commands, it uses the sandbox id to identify 
+the endpoint of the corresponding `containerd-shim-kata-v2`.
+
+### containerd-shim-kata-v2 changes
+`containerd-shim-kata-v2` provides an API for sandbox management through a Unix domain socket. Two new handlers are proposed: `/direct-volume/stats` and `/direct-volume/resize`:
+
+```bash
+Example:
+
```
jodh-intel, line 99:
```suggestion
Example:

```bash
```

jodh-intel (approved):
Thanks @fengwang666 - a few doc "nits" but overall

lgtm

chestack:
@fengwang666  could you help explain what's the difference between this direct-assigned volume and [Raw Block Volume feature ](https://kubernetes.io/blog/2019/03/07/raw-block-volume-support-to-beta/) in Kubernetes pvc

ddebroy:
With raw-block, you need to mount the file system inside the workload pod. This requires special capabilities to execute the mount syscall which the pod spec may not specify. Further, the containers in a pod usually expect the file system to be already mounted on the block device during pod initialization. Thus, performing the file system mount within the pod requires changes in the containers (assuming they have the necessary capabilities in the first place).

chestack:
@ddebroy thanks for your quick response.  I understand that direct-assigned volume does more things than raw block PVC such as: automatically mount fs,  native fs feature and PVC resize.

for performance(IO path), is there any difference between direct-assigned volume and block PVC.

lnattrass:
Just read through the design here, IIUC the stop gap until KEP-2857 is in place, is that we will require the CSI provider to support direct-assigned volumes for Kata specifically?

If this is the case, are there any CSI providers which support this already, or any that intend to support it that we're aware of?

Issue #3254, opened 2021-12-12
kata-manager.sh fails with "Cannot determine Kata Containers release file" error on x86_64 platform


radeksm:
Trying to install Kata with _**kata-manager.sh**_ on Rocky Linux release 8.5, so no kata-collect-data.sh to collect the basic info.

Using latest _**kata-manager.sh**_ from official repo and run it like this:

> bash -x kata-manager.sh

getting below error because release URL does not include the architecture name.

> ++ local version=2.3.0
++ '[' -z 2.3.0 ']'
+++ github_download_release https://api.github.com/repos/kata-containers/kata-containers/releases 2.3.0
+++ local url=https://api.github.com/repos/kata-containers/kata-containers/releases
+++ local version=2.3.0
+++ pushd /tmp/tmp.yaB5l9SFTB
++++ github_get_release_file_url https://api.github.com/repos/kata-containers/kata-containers/releases 2.3.0
++++ local url=https://api.github.com/repos/kata-containers/kata-containers/releases
++++ local version=2.3.0
+++++ curl -sL https://api.github.com/repos/kata-containers/kata-containers/releases
+++++ jq --arg version 2.3.0 -r '.[] | select(.tag_name == $version) | .assets[0].browser_download_url'
++++ download_url=https://github.com/kata-containers/kata-containers/releases/download/2.3.0/kata-containers-2.3.0-vendor.tar.gz
++++ '[' https://github.com/kata-containers/kata-containers/releases/download/2.3.0/kata-containers-2.3.0-vendor.tar.gz = null ']'
++++ '[' -z https://github.com/kata-containers/kata-containers/releases/download/2.3.0/kata-containers-2.3.0-vendor.tar.gz ']'
+++++ uname -m
++++ local arch=x86_64
++++ '[' x86_64 = x86_64 ']'
++++ arch='(x86_64|amd64)'
++++ echo https://github.com/kata-containers/kata-containers/releases/download/2.3.0/kata-containers-2.3.0-vendor.tar.gz
++++ egrep -q '(x86_64|amd64)'
++++ die 'No release for '\''(x86_64|amd64) architecture (https://api.github.com/repos/kata-containers/kata-containers/releases)'
++++ echo -e 'ERROR: No release for '\''(x86_64|amd64) architecture (https://api.github.com/repos/kata-containers/kata-containers/releases)'
ERROR: No release for '(x86_64|amd64) architecture (https://api.github.com/repos/kata-containers/kata-containers/releases)
[...]

PR #3255, opened 2021-12-12
utils/kata-manager.sh: Remove target machine arch check

radeksm:
Download url does not contain the arch name, let's remove it and don't
fail installation with kata-manager.sh.

Fixes: #3254


Jakob-Naucke:
@radeksm the bug is valid, but I think we're actually looking for another fix. See #3271 for rationale.

radeksm:
@Jakob-Naucke  sure, I understand. I am starting with Kata and it was rather naive attempt to fix or to remove something what was broken to me :) Didn't know how to extract the architecture and build the URL.

PR #3283, opened 2021-12-16
Convert src/libs into a Cargo workspace

jiangliu:
Convert libs into a Cargo workspace, so all libraries could share the build infrastructure.


Jakob-Naucke:
@jiangliu can this be closed in favour of #3293 or do you want separate PRs?

PR #3288, opened 2021-12-16
qemu: Disable libudev for QEMU 5.2 and newer

### LINKED ISSUE #3078: qemu: disable libudev by default
#3003 disabled libudev for static builds.

> I havent looked too much into this, but going by this: https://patchew.org/QEMU/20201015150638.1396659-1-pbonzini@redhat.com/, it looks like we don't really need libudev for dynamic builds either. Would be worth disabling this be default. cc @devimc

Fair point. We already disable USB and we don't seem to be using `qemu-pr-helper`, so it seems we can safely disable it for all builds, not just static ones.

_Originally posted by @gkurz in https://github.com/kata-containers/kata-containers/issues/3003#issuecomment-972936660_

gkurz:
Commit 112ea25859d6 disabled libudev for static builds because it was
breaking snap. It turns out that the only users of libudev in QEMU are
qemu-pr-helper and USB. Kata already disables USB and doesn't use
qemu-pr-helper. Disable libudev for all builds if QEMU supports it, i.e.
version 5.2 or newer.

Fixes #3078

### DIFF
--- a/tools/packaging/scripts/configure-hypervisor.sh
+++ b/tools/packaging/scripts/configure-hypervisor.sh
@@ -310,9 +310,10 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-qom-cast-debug)
 	qemu_options+=(size:--disable-tcmalloc)
 
-	# Disable libudev for static build
+	# Disable libudev since it is only needed for qemu-pr-helper and USB,
+	# none of which are used with Kata
 	if gt_eq "${qemu_version}" "5.2.0" ; then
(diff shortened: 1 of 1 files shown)


devimc (approved):
thanks @gkurz

wainersm (approved):
LGTM! Thanks @gkurz !

wainersm:
Since this isn't a fix I assume it won't need a backport, so I will set the label properly.

amshinde (approved):

fidencio:
@gkurz, I think this could be backported to the stable-2.3 branch according to your own discretion.

PR #3351, opened 2021-12-24
agent: Fix execute_hook() args error

Bevisy:
The hook.args[0] is the binary name which shouldn't be included
in the Command.args.

refer [posix-platform-hooks](https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks)

Fixes: #2610

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -1482,7 +1482,12 @@ async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()> {
         return Err(anyhow!(nix::Error::EINVAL));
     }
 
-    let args = h.args.clone();
+    let mut args = h.args.clone();
+    // the hook.args[0] is the hook binary name which shouldn't be included
+    // in the Command.args
+    if args.len() > 1 {
+        args.remove(0);
+    }
     let env: HashMap<String, String> = h
         .env
         .iter()
@@ -1529,7 +1534,7 @@ async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()> {
         // Close stdin so that hook program could receive EOF
         child.stdin.take();
 
-        // read something from stdout for debug
+        // read something from stdout and stderr for debug
         let mut out = String::new();
         child
             .stdout
@@ -1540,6 +1545,16 @@ async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()> {
             .unwrap();
         info!(logger, "child stdout: {}", out.as_str());
 
+        let mut err = String::new();
+        child
+            .stderr
+            .as_mut()
+            .unwrap()
+            .read_to_string(&mut err)
+            .await
+            .unwrap();
+        info!(logger, "child stderr: {}", err.as_str());
+
         match child.wait().await {
             Ok(exit) => {
                 let code = exit
@@ -1547,7 +1562,10 @@ async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()> {
                     .ok_or_else(|| anyhow!("hook exit status has no status code"))?;
 
                 if code != 0 {
-                    error!(logger, "hook {} exit status is {}", &path, code);
+                    error!(
+                        logger,
+                        "hook {} exit status is {}, error message is {}", &path, code, err
+                    );
                     return Err(anyhow!(nix::Error::UnknownErrno));
                 }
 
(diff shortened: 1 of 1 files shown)


Bevisy:
find_hooks() https://github.com/kata-containers/kata-containers/blob/0fe20854e7e34199a6acec68f24e2ea91597eff5/src/agent/src/sandbox.rs#L333

src/agent/rustjail/src/container.rs
```diff
@@ -1655,7 +1660,7 @@ mod tests {
             &slog_scope::logger(),
             &Hook {
                 path: sleep,
-                args: vec!["2".to_string()],
+                args: vec!["sleep".to_string(), "2".to_string()],
```
jodh-intel, line 1663: Looking at the existing unit tests, I don't see one that correctly runs a hook and checks the hook output (the `xargs` test doesn't really count imho). That probably explains why this feature has been broken too! ;)

Therefore, to prove the fix on this PR works please can you add a new test that does the equivalent of something like this:

```bash
tmpdir=$(mktemp -d)
touch ${tmpdir}/a ${tmpdir}/b ${tmpdir}/c
```

The test should then check to ensure all three files are created.

jodh-intel (commented):
Thanks @Bevisy! I've made a suggestion for a new test that would help here I think.

src/agent/rustjail/src/container.rs
Bevisy: Thanks for your advice, I will do some fixes about tests when I figured out other problems I found related to using hook in guest vm.

liubin:
@Bevisy any update?

Bevisy:
PTAL @jodh-intel @liubin 
The failure of ci-s390x does not seem to be related to this modification.

src/agent/rustjail/src/container.rs
```diff
@@ -1540,14 +1545,26 @@ async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()> {
             .unwrap();
         info!(logger, "child stdout: {}", out.as_str());
 
+        let mut err = String::new();
```
liubin, line 1548: @Bevisy you added reading stderr code here, so can you also add a test case that runs invalid hooks and generate stderr?

Bevisy: Thank you for your suggestion, I will deal with it later

Bevisy: I have add the test which can redirect logs to stderr, but this test looks awkward. Because it looks like doing nothing. Can you give any advice?

Bevisy:
PTAL @liubin @jodh-intel

src/agent/rustjail/src/container.rs
```diff
@@ -1644,7 +1663,10 @@ mod tests {
             },
         )
         .await
-        .unwrap()
+        .unwrap();
+        defer!(fs::remove_file(temp_file).unwrap(););
```
liubin, line 1667: It's better and safe to move this `defer` before `execute_hook`.

src/agent/rustjail/src/container.rs
```diff
@@ -1677,6 +1699,31 @@ mod tests {
...
+                version: "1.2.3".to_string(),
+                id: "321".to_string(),
+                status: ContainerState::Running,
+                pid: 2,
+                bundle: "".to_string(),
+                annotations: Default::default(),
+            },
+        )
+        .await
+        .unwrap()
```
liubin, line 1724: `execute_hook` is calld but result is not checked, `Ok` or `Err`.

Bevisy:
Hold on until the issue https://github.com/kata-containers/tests/issues/4388 fixed, or ci can't pass.

Bevisy:
@Tim-Zhang Do you have any idea about the ci failure?
```sh
thread 'container::tests::test_execute_hook' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }', rustjail/src/container.rs:1532:14
```

src/agent/rustjail/src/container.rs
```diff
@@ -1540,14 +1545,27 @@ async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()> {
...
 
+        let mut err = String::new();
+        child
+            .stderr
+            .as_mut()
+            .unwrap()
+            .read_to_string(&mut err)
+            .await
+            .unwrap();
+        error!(logger, "child stderr: {}", err.as_str());
```
liubin, line 1556: Since below will also log an error message, here `error!` can be replaced with `info!`.

liubin (dismissed):

src/agent/rustjail/src/container.rs
Bevisy: Ok

src/agent/rustjail/src/container.rs
```diff
@@ -1624,15 +1642,19 @@ mod tests {
 
     #[tokio::test]
     async fn test_execute_hook() {
-        let xargs = which("xargs").await;
+        let temp_file = "/tmp/test_execute_hook";
+
+        let touch = which("touch").await;
```
Jakob-Naucke, line 1647: Can you `let` the command once? (Also for the ones below.) LGTM otherwise.

Bevisy: Sorry, I don't understand what once means here, can you say it more clearly or give an example?

fidencio:

src/agent/rustjail/src/container.rs
Jakob-Naucke: Sorry, I was very brief. I meant something like
```rust
let command = "touch";
let touch = which(command).await;
…
    args: vec![command.to_string(), …
```

etc.

bergwolf (approved):
lgtm!

fidencio (approved):
lgtm, thanks @Bevisy!

Issue #3353, opened 2021-12-24
Return more details if error to shim


liubin:
For calls from shim to agent, the return error will be processed like this:

```rust
         match self.do_start_container(req).await {
             Err(e) => Err(ttrpc_error(ttrpc::Code::INTERNAL, e.to_string())),
             Ok(_) => Ok(Empty::new()),
         }
```

The `e.to_string()` return only a part of the error(for example set by `context()`), this may lead lack of information.

More detailed error message is needed.

PR #3354, opened 2021-12-24
agent: return detail error message for RPC calls from shim

liubin:
For calls from shim to agent, the return error will be processed like this:

```rust
match self.do_start_container(req).await {
    Err(e) => Err(ttrpc_error(ttrpc::Code::INTERNAL, e.to_string())),
    Ok(_) => Ok(Empty::new()),
}
```

The `e.to_string()` return only a part of the error(for example set by `context()` on a `nix` Error),
this may lead lack of information.

The `format!("{:?}", err)` will return more info.

Logs before and after this patch:

Before:
> "rpc error: code = Internal desc = failed to umount \"/run/kata-containers/sandbox/ephemeral/hugepage\""

After:
> "rpc error: code = Internal desc = failed to umount \"/run/kata-containers/sandbox/ephemeral/hugepage\"\n\nCaused by:\n    EINVAL: Invalid argument"

Fixes: #3353

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -14,7 +14,7 @@ use std::path::Path;
 use std::sync::Arc;
 use ttrpc::{
     self,
-    error::get_rpc_status as ttrpc_error,
+    error::get_rpc_status,
     r#async::{Server as TtrpcServer, TtrpcContext},
 };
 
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @liubin.

lgtm

Bevisy (approved):
lgtm

bergwolf (approved):

Issue #3369, opened 2021-12-30
agent's tokio runtime thread's namespace was changed


lifupan:
Kata agent's tokio  runtime thread's ipc and uts namespace were changed, which isn't expected.

```
bash-5.0# ps -ef | grep agent
root        58     1  0 05:26 ?        00:00:00 /usr/bin/kata-agent
root        78    76  0 05:27 pts/1    00:00:00 grep agent
bash-5.0# ls -al /proc/58/task
total 0
dr-xr-xr-x 5 root root 0 Dec 30 05:26 .
dr-xr-xr-x 9 root root 0 Dec 30 05:26 ..
dr-xr-xr-x 7 root root 0 Dec 30 05:27 58
dr-xr-xr-x 7 root root 0 Dec 30 05:26 60
dr-xr-xr-x 7 root root 0 Dec 30 05:27 61
bash-5.0# ls -al /proc/58/ns
total 0
dr-x--x--x 2 root root 0 Dec 30 05:27 .
dr-xr-xr-x 9 root root 0 Dec 30 05:26 ..
lrwxrwxrwx 1 root root 0 Dec 30 05:27 cgroup -> 'cgroup:[4026531835]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 ipc -> 'ipc:[4026531839]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 mnt -> 'mnt:[4026531840]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 net -> 'net:[4026531992]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 pid -> 'pid:[4026531836]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 pid_for_children -> 'pid:[4026531836]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 user -> 'user:[4026531837]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 uts -> 'uts:[4026531838]'
bash-5.0# ls -al /proc/60/ns 
total 0
dr-x--x--x 2 root root 0 Dec 30 05:27 .
dr-xr-xr-x 9 root root 0 Dec 30 05:27 ..
lrwxrwxrwx 1 root root 0 Dec 30 05:27 cgroup -> 'cgroup:[4026531835]'
lrwxrwxrwx 1 root root 0 Dec 30 05:27 ipc -> 'ipc:[4026532264]'
[...]

PR #3370, opened 2021-12-30
agent: fix the issue of creating new namespaces for agent

lifupan:
The tokio's spawn will only create an future async task
instead of a new real thread, thus executing unshare to
create a new namespace in tokio's async task would make
the agent process to join in the new created namespace,
which isn't expected.

Thus, we'd better to to the unshare in a real thread to
prevent moving the agent process into a new namespace.

Fixes: #3369

### DIFF
--- a/src/agent/src/namespace.rs
+++ b/src/agent/src/namespace.rs
@@ -100,7 +100,7 @@ impl Namespace {
         self.path = new_ns_path.clone().into_os_string().into_string().unwrap();
         let hostname = self.hostname.clone();
 
-        let new_thread = tokio::spawn(async move {
+        let new_thread = std::thread::spawn(move || {
             if let Err(err) = || -> Result<()> {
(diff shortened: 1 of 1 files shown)


egernst (commented):
Change itself looks good. Is there an integration or unit test that we can add to exercise this? Thanks Fupan!

jodh-intel (commented):
Thanks @lifupan and good catch!

As @egernst says, please could you add a unit test that asserts the expected namespace behaviour.

lifupan:
> Change itself looks good. Is there an integration or unit test that we can add to exercise this? Thanks Fupan!

Yes, I added one https://github.com/kata-containers/tests/pull/4341

jodh-intel (approved):
Thanks @lifupan.

lgtm

amshinde (approved):

Tim-Zhang (approved):
LGTM, thanks @lifupan

Issue #1864, opened 2021-05-15
NewTestConstraint in runtime/katatestutils panic on Debian testing/sid


zhsj:
# Description of problem

As Debian testing and sid are rolling update OS, they don't have version.

So when `getDistroDetails` returns `errUnknownDistroName`, `NewTestConstraint` panic.

# Expected result

Can run test on Debian testing and sid.

# Actual result

Panic

# Further information

```
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

zhsj:
I'd like to fix it. However I'd like to see your opinions. I don't see any use of  `NeedDistroVersionXX`, can we just remove them? It seems complex to implement these fucntions well for rolling update OS.

zhsj:
CC @jodh-intel , as these functions are implemented in https://github.com/kata-containers/runtime/pull/1587/files

c3d:
Any PR would be welcome. I don't see your work as overlapping with what @jodh-intel was proposing.

PR #3389, opened 2022-01-05
katatestutils: remove distro constraints

zhsj:
The distro constraint parses os release files, which may not contain
distro version(VERSION_ID field), for example rolling release distributions
like Debian testing, archlinux.

These distro constraints are not used anyway, so removing them instead
of fixing the complex version detection.

Fixes: #1864

Signed-off-by: Shengjing Zhu <zhsj@debian.org>

### DIFF
--- a/src/runtime/pkg/katatestutils/README.md
+++ b/src/runtime/pkg/katatestutils/README.md
@@ -102,36 +102,6 @@ func TestOnlyRunWhenNotRoot(t *testing.T) {
 }
 ```
 
-#### Skip tests based on distro
-
-Use the `NeedDistro()` constraint to skip a test unless running on a
-particular Linux distribution:
-
-```go
-func TestOnlyRunOnUbuntu(t *testing.T) {
-
-    if tc.NotValid(ktu.NeedDistro("ubuntu")) {
-        t.Skip("skipping test as not running on ubuntu")
-    }
-
-    // Test code to run on Ubuntu only ...
-}
-```
-
-Use the `NeedDistroNotEquals()` constraint to skip a test unless running
-on a Linux distribution other than the one specified:
-
-```go
-func TestDontRunOnFedora(t *testing.T) {
-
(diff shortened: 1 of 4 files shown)


jodh-intel:
@zhsj - Thanks for raising. I'm a little confused here: although `VERSION_ID` is optional according to [`os-release(5)`](https://www.freedesktop.org/software/systemd/man/os-release.html), alphabetics are supported:

> VERSION_ID=
> A lower-case string (mostly numeric, no spaces or other characters outside of 0–9, a–z, ".", "_" and "-") identifying the operating system version...

It's admittedly written in a rather tortuous manner, but this implies to me that Debian testing could set:

```
VERSION_ID=debian-testing
```

... since there appears to be no requirement that the `VERSION_ID` be a value that can be compared numerically to another (unlike https://semver.org).

Aside: The `os-release` file format standard appears to be, shall we say, "overly flexible" (aka not very helpful in pathological cases): _all_ the fields that most folk care about (name and version) are optional [1]. This seems odd frankly because unless I need more coffee, _an entirely empty_ `/etc/os-release` _file **is valid**_ !!!

---

[1] - Apart from `IMAGE_VERSION=` which isn't used for "normal" distros anyway).

zhsj:
I don't think we are at the place that shall ask all rolling release distributions to set `VERSION_ID` in os-release file. It's not only Debian testing, I also checked archlinux.

The overall complex that makes me think removing the unused code is more convenient.

Jakob-Naucke:
@jodh-intel if we aren't using it anyhow, do we care a lot about whether we could use `VERSION_ID`?

fidencio:

jodh-intel (approved):
Let's remove it and re-add later if required...

Jakob-Naucke (approved):

bergwolf (approved):
Looks Reasonable to me.

Issue #3364, opened 2021-12-26
Please change network checks in `kata-runtime check` to opt-in


zhsj:
**Which feature do you think can be improved?**

kata-runtime check command.

**How can it be improved?**

Currently the network checks are opt-out. When users first run this command, without looking `--help` or other documents, they are risked to do unexpected network access.

**Additional Information**

Anything else to add?

No

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

Yes.

PR #735, opened 2020-09-16
runtime: make kata-check check for newer release

jodh-intel:
Update `kata-check` to see if there is a newer version available for download. Useful for users installing static packages (without a package manager).

Fixes: #734.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #3390, opened 2022-01-05
runtime: change network checks in kata check command to opt-in

zhsj:
Fixes: #3364

Signed-off-by: Shengjing Zhu <zhsj@debian.org>

### DIFF
--- a/src/runtime/cmd/kata-runtime/kata-check.go
+++ b/src/runtime/cmd/kata-runtime/kata-check.go
@@ -70,9 +70,6 @@ const (
 	genericCPUFlagsTag    = "flags"      // nolint: varcheck, unused, deadcode
 	genericCPUVendorField = "vendor_id"  // nolint: varcheck, unused, deadcode
 	genericCPUModelField  = "model name" // nolint: varcheck, unused, deadcode
-
-	// If set, do not perform any network checks
-	noNetworkEnvVar = "KATA_CHECK_NO_NETWORK"
 )
 
 // variables rather than consts to allow tests to modify them
@@ -324,10 +321,6 @@ var kataCheckCLICommand = cli.Command{
 			Usage: "Don't filter out pre-release release versions",
 		},
 		cli.BoolFlag{
-			Name:  "no-network-checks, n",
-			Usage: "Do not run any checks using the network",
-		},
-		cli.BoolFlag{
 			Name:  "only-list-releases",
 			Usage: "Only list newer available releases (non-root only)",
 		},
@@ -340,21 +333,13 @@ var kataCheckCLICommand = cli.Command{
 			Usage: "display the list of checks performed",
 		},
 	},
-	Description: fmt.Sprintf(`tests if system can run %s and version is current.
-
-ENVIRONMENT VARIABLES:
-
-- %s: If set to any value, act as if "--no-network-checks" was specified.
+	Description: fmt.Sprintf(`Tests if system can run %s and version is current (need network access).
 
 EXAMPLES:
 
 - Perform basic checks:
 
   $ %s check
-
-- Local basic checks only:
-
-  $ %s check --no-network-checks
 
 - Perform further checks:
 
@@ -373,8 +358,6 @@ EXAMPLES:
   $ %s check --only-list-releases --include-all-releases
 `,
 		katautils.PROJECT,
-		noNetworkEnvVar,
-		katautils.NAME,
 		katautils.NAME,
 		katautils.NAME,
 		katautils.NAME,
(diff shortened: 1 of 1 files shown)


jodh-intel:
Thanks for raising @zhsj, but I'm not sure this is a good idea. Please can you explain a little more about why this should be done?

The check command has evolved over time and is quite careful about how it works: network checks are disabled when running as `root` but when run as a non-`root` user we _want_ users/admins to be notified of new releases in case they are using one of the (many) instlalation types that do not offer automatic updates:

- https://github.com/kata-containers/kata-containers/tree/main/docs/install#packaged-installation-methods

That might not work for all use-cases, but we want to make new releases as visible as possible even for static installations so that users are warned that they should upgrade.

We added the ability to disable network checks to allow the command to be run in a CI / build environment where there was no network access. I'm assuming you wish to run `kata-runtime check` in some sort of build environment that disallows network access so why not just set the env var or specify the option to disable network checks?

zhsj:
@jodh-intel This is change doesn't remove network check, but only changing it to opt-in, instead of opt-out. As I said in #3364, the opt-out policy makes user risky to do accident network access, and may leak user privacy.

zhsj:
> I'm assuming you wish to run kata-runtime check in some sort of build environment that disallows network access so why not just set the env var or specify the option to disable network checks

No, that's not my wish. I want users who run `check` command only access internet when they explicitly allow such behavior. This is not about CI or any build environment.

jcvenegas:
@zhsj so If I understand correctly. The main problem you see is this:
# Problem:
If users use the `check` subcommand it will do some network access to verify the version by default. And this is something that you consider a privacy issue. 

## What is the potential privacy concern?
The privacy issue is that by doing network access to a remote  endpoint their IPs may be collected?

Solution:
Your proposed solution, make explicit that the command will do network usage and remove the option that prevents it?

---
If this is correct I would suggest just add an extra comment in the `check` description saying that it does network access by default.  It will do the same as you propose now, but we keep the logic of check for limited network environments that is the main goal of the flag you want to remote  .

zhsj:
@jcvenegas The point I raised in #3364 is

> When users first run this command, without looking --help or other documents, they are risked to do unexpected network access.

So I don't think adding extra comment in description helps.

zhsj:
> What is the potential privacy concern?
> The privacy issue is that by doing network access to a remote endpoint their IPs may be collected?

Yes. leaking users IP, and their usage of kata-containers.

jodh-intel:
It's worth stating that the `check` command is never run by Kata itself. It's a utility command that the user/admin _may_ run if they want to, but it's not needed to run Kata at all.

If the docs for `--help` are made clearer I would have thought that would be adequate: if you really care about privacy, presumably you don't run random commands, hoping they won't do things you don't expect: you'd look at the docs before running the command (and obviously hope those docs are correct, or check the source code ;-)

If we did land this, we'd improve privacy, but potentially reduce security as users wouldn't be aware of critically important updates. That clearly isn't ideal for everyone.

Also, even if users don't want `check` to hit the network:

- Normally, they would have to hit the network at some point to download a container image (unless those are all generated locally and hosted in a local registry)
- How do they download Kata in the first place?

@zhsj - The other point is that this PR switches the current behaviour, so we would break users who have come to expect this behaviour.

Let's see what others feel as we've got a mix of security, usability, discoverability and privacy arguments here, and I suspect we'll get a number of differnet views on what folk see as the most important.

zhsj:
> Also, even if users don't want check to hit the network:
> Normally, they would have to hit the network at some point to download a container image (unless those are all 
> generated locally and hosted in a local registry)
>How do they download Kata in the first place?

My point is not "check" shouldn't nit network. But should with explicit permission. This is different with the cases you give. That's the whole point of this PR, "opt-in" vs "opt-out".

zhsj:
Let me make my point clear: any action that has potential privacy breach should be **opt-in**. It's not any action that has potential privacy breach should be removed.

jodh-intel:
Can we get more input from the community on this topic please?

jodh-intel:
For reference, the `kata-runtime check` feature that checks for newer Kata Containers releases was added on https://github.com/kata-containers/kata-containers/pull/735 in September 2020.

amshinde:
@zhsj @jodh-intel I am not in favor of this change. Like @jodh-intel mentioned, `kata-runtime check` is a user utility meant to be run separately. If the --help does not mention that we access the internet to check for the latest versions, we can do so. I do not see making  a curl call on a public url to get the latest version is a security breach or compromises his privacy as long as this is documented.

Jakob-Naucke:
@zhsj Let me make a comparison. I know you're coming from Debian, and e.g. `apt` doesn't do this, but the `dnf` package manager will update repositories if they're out of date when doing a `dnf search`, even though the unitiated user might think it would just search the offline index and not create any connection. I'm not aware of such complaints there. Therefore, I think it is not a problem to make such connections, and anyone who is really conscious about not leaking their IP should probably not run commands they don't know well or firewall on an allow-list basis entirely (or heck, air gap).

fidencio:

fidencio:
@zhsj, thanks for your input on this, but at this moment several of our contributors, including myself, think the current approach is reasonable.

If that's a limitation for Debian, I'd suggest keeping a downstream patch on Debian side to avoid such connections.

I'm closing this PR now, thanks a lot for your contribution and your understanding!

PR #3406, opened 2022-01-07
Implement direct-assigned volume

fengwang666:
The implementation is based on a previous [design](https://github.com/kata-containers/kata-containers/pull/3248) discussed in AC meeting. It also references a previous [PR](https://github.com/kata-containers/kata-containers/pull/2825) by @egernst. 

The PR is divided into 8 commits. The commit message describe what's in each commit. There are still a couple pieces missing, which will be done in separate PRs:
1. Implement stat and resize operations inside the agent.
2. Add resize operation on the hypervisor (e.g. QEMU).


src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,55 @@ message GetMetricsRequest {}
...
+	// An abnormal volume does not meet these criteria.
+	// This field is REQUIRED.
+	bool abnormal = 1;
+
+	// The message describing the condition of the volume.
+	// This field is REQUIRED.
+	string message = 2;
+}
+message ResizeVolumeRequest {
+	string path = 1;
```
liubin, line 559: How about changing `path` to `device-path`, since kata-volume.go is using `device-path`.

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,55 @@ message GetMetricsRequest {}
...
+	// The value of this field MUST NOT be negative.
+	int64 available = 1;
+
+	// The total capacity in specified Unit. This field is REQUIRED.
+	// The value of this field MUST NOT be negative.
+	int64 total = 2;
+
+	// The used capacity in specified Unit. This field is OPTIONAL.
+	// The value of this field MUST NOT be negative.
+	int64 used = 3;
```
jodh-intel, line 540: Shouldn't these all be `uint64`'s?

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,55 @@ message GetMetricsRequest {}
...
+	Unit unit = 4;
+}
+
+// VolumeCondition represents the current condition of a volume.
+message VolumeCondition {
+
+	// Normal volumes are available for use and operating optimally.
+	// An abnormal volume does not meet these criteria.
+	// This field is REQUIRED.
+	bool abnormal = 1;
```
jodh-intel, line 552: This isn't very granular. Is there any way to make this an enum showing the actual state?

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,55 @@ message GetMetricsRequest {}
...
+	repeated VolumeUsage usage = 1;
+	// Information about the current condition of the volume.
+	// This field is OPTIONAL.
+	// This field MUST be specified if the VOLUME_CONDITION node
+	// capability is supported.
+	VolumeCondition volume_condition = 2;
+}
+message VolumeUsage {
+	enum Unit {
+		UNKNOWN = 0;
```
jodh-intel, line 526: Why is this needed?

jodh-intel (commented):
Thanks @fengwang666. A few comments.

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,55 @@ message GetMetricsRequest {}
...
 	string metrics = 1;
 }
+
+message VolumeStatsRequest {
+	string path = 1;
+}
+
+message VolumeStatsResponse {
+	// This field is OPTIONAL.
+	repeated VolumeUsage usage = 1;
```
bergwolf, line 517: why is it `repeated`?

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,55 @@ message GetMetricsRequest {}
 message Metrics {
 	string metrics = 1;
 }
+
+message VolumeStatsRequest {
+	string path = 1;
```
bergwolf, line 512: Is it a `guest path` or a `container path`?

jodh-intel: +1.

Not specific to this PR but I really wish _all_ protocol buffer messages were fully documented.

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -148,6 +198,8 @@ func (s *service) startManagementServer(ctx context.Context, ociSpec *specs.Spec
 	m := http.NewServeMux()
 	m.Handle("/metrics", http.HandlerFunc(s.serveMetrics))
 	m.Handle("/agent-url", http.HandlerFunc(s.agentURL))
+	m.Handle("/direct-volume/stats", http.HandlerFunc(s.serveVolumeStats))
+	m.Handle(DirectVolumeResizeUrl, http.HandlerFunc(s.serveVolumeResize))
```
egernst, line 202: Why hardcoded for one and variable for other?

egernst (commented):
Is it possible to break this up in individual commits? It'll make reviewing a bit easier for me.  In my initial PR, I broke it out per interface/layer change: see https://github.com/kata-containers/kata-containers/pull/2825/commits

If it isn't too much work, can you do something similar (so I can compare the implementation!)?

Thanks for driving this.

src/libs/protocols/protos/agent.proto
fengwang666: Updated

src/libs/protocols/protos/agent.proto
fengwang666: updated

src/libs/protocols/protos/agent.proto
fengwang666: I'm copying this from the [csi spec](https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1470)

src/libs/protocols/protos/agent.proto
fengwang666: I'm copying this from the [csi spec](https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1449). Happy to remove it.

src/libs/protocols/protos/agent.proto
fengwang666: This is how it's defined in the [CSI spec ](https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1439). `VolumeUsage` has two Units: `BYTES` and `INODES`.

src/runtime/pkg/containerd-shim-v2/shim_management.go
fengwang666: updated

src/libs/protocols/protos/agent.proto
fengwang666: It's the guest path. I have updated and added comments.

snir911:
Hi @fengwang666 , please see static checks:
```
Error: utils_test.go:64:13: undeclared name: `getSandboxIdForVolume` (typecheck)
	id, err := getSandboxIdForVolume(volumePath)
	           ^
```

src/runtime/pkg/direct-volume/utils_test.go
```diff
@@ -61,7 +61,7 @@ func TestRecordSandboxId(t *testing.T) {
 	err = RecordSandboxId(sandboxId, volumePath)
 	assert.Nil(t, err)
 
-	id, err := getSandboxIdForVolume(volumePath)
+	id, err := GetSandboxIdForVolume(volumePath)
```
egernst, line 64: This was made public in a different commit, right? Should squash?

src/runtime/pkg/utils/shimclient/shim_management_client.go
```diff
@@ -0,0 +1,74 @@
+package shimclient
```
egernst, line 1: A bit of a nit; we should make this package introduction (or migration from kata-monitor) either its own commit, or update the commit to include these details.

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,95 @@
...
+func RecordSandboxId(sandboxId string, volumePath string) error {
+	mountInfoFilePath := filepath.Join(kataDirectVolumeRootPath, volumePath, mountInfoFileName)
+	if _, err := os.Stat(mountInfoFilePath); err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filepath.Join(kataDirectVolumeRootPath, volumePath, sandboxId), []byte(""), 0600)
+}
+
+func getSandboxIdForVolume(volumePath string) (string, error) {
```
egernst: I think you update later in the commit series -- we should make sure this is public when we introduce this pkg, right?

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+)
+
+const (
+	mountInfoFileName        = "mountInfo.json"
+	defaultTimeout = 3 * time.Second
+)
+var kataDirectVolumeRootPath = "/run/kata-containers/shared/direct-volumes"
+
+// MountInfo contains the information for the Kata agent to mount a filesystem on the volume inside the guest VM.
```
egernst, line 18: Some info is consumed by the shim, and other info is eventually passed to the guest agent. What do you think about rewording, like:
```
MountInfo contains the information needed by Kata to consume a host block device and mount it as a filesystem inside the guest VM.
```

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,95 @@
...
+type MountInfo struct {
+	// The host path of the block device backing the volume.
+	DevicePath string `json:"device_path"`
```
egernst, line 21: I think Device would be better than DevicePath. It needn't necessarily be a block device path on the host. It could, for example, be a URL?

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+const (
+	mountInfoFileName        = "mountInfo.json"
+	defaultTimeout = 3 * time.Second
+)
+var kataDirectVolumeRootPath = "/run/kata-containers/shared/direct-volumes"
+
+// MountInfo contains the information for the Kata agent to mount a filesystem on the volume inside the guest VM.
+type MountInfo struct {
```
egernst, line 19: Is there a scenario where we'd want to handle different Volume types? For example, block device, network attached, etc. ? May be good to have a VolumeType field?

@bergwolf WDYT -- I recall this feedback from initial PRs.

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,95 @@
+package volume
```
egernst, line 1: In this design, the CSI plugin will interact with kata-runtime CLI online, but will still likely vendor this package for the MountInfo structure, correct?

It isn't expected that the CSI plugin will need or call these public functions?

egernst (commented):
Still working through it -- I like the improvements - thanks Feng.

src/runtime/cmd/kata-runtime/kata-volume.go
```diff
@@ -35,16 +34,16 @@ var kataVolumeCommand = cli.Command{
 }
 
 var addCommand = cli.Command{
-	Name: "add",
+	Name:  "add",
```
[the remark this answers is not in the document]
fengwang666: The static check complains the fmt error. I should have added this to the previous commit.

src/runtime/virtcontainers/container.go
```diff
@@ -695,6 +696,28 @@ func (c *Container) createBlockDevices(ctx context.Context) error {
 			continue
```
shuochen0311, line 696: will the directly assigned volume get correctly handled or get omitted?

src/agent/src/rpc.rs
```diff
@@ -1336,6 +1364,23 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
     Ok((size, plug))
 }
 
+fn get_volume_stats(path: &str) -> Result<(VolumeUsage, VolumeCondition)> {
```
yibozhuang, line 1367: This only handles disk usage. What about `inodes` stats if the direct-assigned volume is mounted in the guest? That would be a separate PR?

src/runtime/pkg/direct-volume/utils.go
fengwang666: That's correct. CSI plugin shouldn't need to call these public functions. It does need to access the MountInfo structure.

src/runtime/pkg/utils/shimclient/shim_management_client.go
fengwang666: updated the commit message to reflect the change.

src/runtime/pkg/direct-volume/utils.go
bergwolf: Yes, we should have a volume type there, even if only block types are handled at first, so that we won't break APIs when adding other volume types in the future.

fengwang666: added `volumeType` field

src/agent/src/rpc.rs
```diff
@@ -1336,6 +1381,39 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
...
 
+fn get_volume_capacity_stats(path: &str) -> Result<VolumeUsage> {
+    let mut usage = VolumeUsage::new();
+
+    let s = System::new();
+    for disk in s.disks() {
+        if disk.name().to_str().unwrap().eq(path) {
+            usage.available = disk.available_space();
+            usage.total = disk.total_space();
+            usage.used = usage.available - usage.total;
```
ddebroy, line 1392: Should this be `usage.total - usage.available`

src/agent/src/rpc.rs
```diff
@@ -1249,6 +1253,47 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+                usage_vec.push(u);
+            }
+            Err(e) => {
+                info!(sl!(), "fail to get volume inode!");
+                return Err(ttrpc_error!(ttrpc::Code::INTERNAL, e));
+            }
+        }
+
+        resp.usage = RepeatedField::from_vec(usage_vec);
+        resp.volume_condition = SingularPtrField::some(condition);
```
ddebroy, line 1293: Is the condition - abnormal or not populated somewhere? How about just doing a stat on the root of the volume and report if the FS is mounted and functional or not?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1589,16 +1556,46 @@ func (k *kataAgent) handleVhostUserBlkVolume(c *Container, m Mount, device api.D
...
+	case config.VhostUserBlk:
+		vol, err = k.handleVhostUserBlkVolume(c, m, device)
+	default:
+		return nil, fmt.Errorf("Unknown device type")
+	}
+
+	return vol, err
+}
+
+// handleBlkOCIMounts will create a unique destination mountpoint in the guest for each volume in the
```
ddebroy, line 1591: Should we reconsider the name here if we want the logic to also be able to handle shared fs types (NFS/SMB) where the volumetype is not block?

fidencio:

src/runtime/virtcontainers/kata_agent.go
fengwang666: I don't have a strong preference either way. Given that we don't have other volume types right now, I'm inclined to leave the name as is and refactor it later when a new volume type is added.

src/agent/src/rpc.rs
fengwang666: updated

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1609,29 +1606,33 @@ func (k *kataAgent) handleBlockVolumes(c *Container) ([]*grpc.Storage, error) {
...
-			vol, err = k.handleVhostUserBlkVolume(c, m, device)
-		default:
-			k.Logger().Error("Unknown device type")
-			continue
-		}
+		// The device will be mounted at a unique location within the VM. Mounting
+		// to the container specific location is handled within the OCI spec. Let's ensure that
+		// the storage mount point is unique, and that this is utilized as the source in the OCI
+		// spec.
+		filename := fmt.Sprintf("%s-%s", uuid.Generate().String(), filepath.Base(vol.MountPoint))
```
ddebroy, line 1619: If two containers within a pod wish to mount the same block device based volume, does this work out? Should the generation of the unique filename (where the file system will be mounted to in the guest) be done during sandbox creation rather than container creation?

ddebroy (commented):
Thanks for addressing the previous set of comments. Overall looking good although I am not very familiar with the codebase. Some overall comments (besides some nits):
1. Are you planning to add agent-ctl [commands](https://github.com/kata-containers/kata-containers/blob/66510b977de358d91ee2ca660286a3a2ff121477/src/tools/agent-ctl/README.md) corresponding to the volume stats and volume resize RPCs? 
2. The agent inside the guest needs to handle the FS resize based on the FS type. Is the implementation of that somewhere already in this PR?
3.  A concern around whether the approach in `handleBlkOCIMounts` can properly handle multiple containers within the pod trying to mount/share the same volume (backed by a direct assigned block device).

jodh-intel:
@fengwang666 - please can you update to remove the conflict?

src/agent/src/rpc.rs
```diff
@@ -1249,6 +1253,59 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
         Err(ttrpc_error!(ttrpc::Code::INTERNAL, ""))
     }
 
+    async fn get_volume_stats(
```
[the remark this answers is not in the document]
fengwang666: Not yet. I intended to address this in a separate PR if that's okay.

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,100 @@
...
+	return &mountInfo, nil
+}
+
+// RecordSandboxId associates a sandbox id with a direct volume.
+func RecordSandboxId(sandboxId string, volumePath string) error {
+	mountInfoFilePath := filepath.Join(kataDirectVolumeRootPath, volumePath, mountInfoFileName)
+	if _, err := os.Stat(mountInfoFilePath); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
fengwang666: @ddebroy This is something I'm not sure. In a previous offline discussion with the Kata AC, my impression is we don't want to restrict a volume to a single sandbox, even though the behavior might be undefined. @egernst, can you comment on this?

src/runtime/virtcontainers/kata_agent.go
fengwang666: @ddebroy Good catch! But I think it might be safe to mount the same filesystem onto the device twice. I tried to mount a loopback device to two different mount points, it appears both mount points have the same content. 
@bergwolf  @egernst wdyt?

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,58 @@ message GetMetricsRequest {}
 message Metrics {
 	string metrics = 1;
 }
+
+message VolumeStatsRequest {
+	// The volume path on the guest outside the container
```
jodh-intel, line 512: Nit: This wording is slightly clearer to me:

```suggestion
	// Full VM guest path of the volume (outside the container)
```

src/libs/protocols/protos/agent.proto
```diff
@@ -505,3 +507,58 @@ message GetMetricsRequest {}
...
+	string volume_guest_path = 1;
+}
+
+// This should be kept in sync with CSI NodeGetVolumeStatsResponse (https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1437)
```
jodh-intel, line 516: Thanks for adding this explanatory comment but this is a little worrying. A few questions / thoughts:

- If you've copied the code, are there any copyright/license issues? Which version of the upstream file have you copied (commit hash)?
- Can you put the copied code into a separate csi.proto and `import` it into `agent.proto` to keep it separate?
- What's the plan for keeping this file in sync?
- How are we going to notice it changed and who is going to be responsible for this?
- The top of that file shows, `// Code generated by make; DO NOT EDIT.`
- Ideally, we've vendor in just the definitions we needed but that might require upstream changes to move these types into a separate (and smaller) `.proto` file.
- Which of our [dependencies](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml), if any, will be impacted if the upstream `csi.proto` changes?

src/libs/protocols/protos/agent.proto
jodh-intel: ack, but see my comments above about the csi spec...

src/libs/protocols/protos/agent.proto
jodh-intel: ack

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -126,6 +138,47 @@ func decodeAgentMetrics(body string) []*dto.MetricFamily {
 	return list
 }
 
+func (s *service) serveVolumeStats(w http.ResponseWriter, r *http.Request) {
+	volumePath, err := url.PathUnescape(strings.TrimPrefix(r.URL.Path, DirectVolumeStatUrl))
+	if err != nil {
+		shimMgtLog.WithError(err).Error("failed to unescape the volume stat url path")
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte(err.Error()))
+	}
```
jodh-intel, line 147: Aren't all these error blocks missing a `return`?

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -126,6 +138,47 @@ func decodeAgentMetrics(body string) []*dto.MetricFamily {
...
+
+	buf, err := s.sandbox.GuestVolumeStats(context.Background(), volumePath)
```
jodh-intel, line 149: `volumePath` is undefined I think? See comment above.

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,100 @@
...
+	Metadata map[string]string `json:"metadata,omitempty"`
+	// Additional mount options.
+	Options []string `json:"options,omitempty"`
+}
+
+// Add writes the mount info of a direct volume into a filesystem path known to Kata Container.
+func Add(volumePath string, mountInfo string) error {
+	volumeDir := filepath.Join(kataDirectVolumeRootPath, volumePath)
+	stat, err := os.Stat(volumeDir)
+	if stat != nil && !stat.IsDir() {
```
jodh-intel, line 42: Is this safe? Why can't the code check the `err` first to be assured that `stat` is  non-`nil` and actually valid?

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,100 @@
...
+	if errors.Is(err, os.ErrNotExist) {
+		os.MkdirAll(volumeDir, 0700)
+	}
+
+	return ioutil.WriteFile(filepath.Join(volumeDir, mountInfoFileName), []byte(mountInfo), 0600)
+}
+
+// Remove deletes the direct volume path including all the files inside it.
+func Remove(volumePath string) error {
+
```
jodh-intel, line 54: Nit: Blank line not needed.

src/runtime/pkg/utils/shimclient/shim_management_client.go
```diff
@@ -0,0 +1,79 @@
...
+	return client, nil
+}
+
+func DoGet(sandboxID string, timeoutInSeconds time.Duration, urlPath string) ([]byte, error) {
+	client, err := BuildShimClient(sandboxID, timeoutInSeconds)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := client.Get(fmt.Sprintf("http://shim/%s", urlPath))
```
jodh-intel, line 51: No check that `urlPath` is valid. Same comment for function below.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1701,29 +1698,33 @@ func (k *kataAgent) handleBlockVolumes(c *Container) ([]*grpc.Storage, error) {
...
+		path := filepath.Join(kataGuestSandboxStorageDir(), filename)
 
-		if vol == nil || err != nil {
-			return nil, err
+		// Update applicable OCI mount source
+		for idx, ociMount := range spec.Mounts {
+			if ociMount.Destination != vol.MountPoint {
+				continue
+			}
+			k.Logger().Debugf("Replacing OCI mount source (%s) with %s", ociMount.Source, path)
```
jodh-intel, line 1719: Nit: Might be worth using `%q` to quote the values. But you could also use structure logging. Something like:

```go
k.Logger().WithFields(logrus.Fields{
  "original-source": ociMount.Source,
  "new-source": path,
}).Debug("Replacing OCI mount source")

File: src/runtime/virtcontainers/sandbox.go [hunk 32]
```diff
@@ -2217,6 +2217,46 @@ func (s *Sandbox) GetAgentURL() (string, error) {
...
+	}
+	return s.agent.getGuestVolumeStats(ctx, guestMountPath)
+}
+
+// ResizeGuestVolume resizes a volume in the guest.
+func (s *Sandbox) ResizeGuestVolume(ctx context.Context, volumePath string, size uint64) error {
+	// TODO: invoke the resize block device feature provided by the hypervisor.
+	// if err := s.hypervisor.ResizeBlockDevice(ctx, devicePath, size); err != nil {
+	// 	   return err
+	//}
```
jodh-intel, line 2234: Please remove commented out code and add a TODO comment with a link to a GitHub issue if required.

src/tools/agent-ctl/src/client.rs
```diff
@@ -172,6 +172,11 @@ static AGENT_CMDS: &'static [AgentCmd] = &[
         st: ServiceType::Agent,
         fp: agent_cmd_sandbox_get_oom_event,
     },
+    AgentCmd {
+        name: "GetVolumeStats",
```
jodh-intel, line 176: Thanks for updating this tool! :smile:

jodh-intel (commented):
Thanks @fengwang666. A few comments...

src/runtime/pkg/utils/shimclient/shim_management_client.go
fengwang666: @jodh-intel Can you be more specific on what validation you have in mind? AFAIK, this method is only used internally. If `urlPath` is invalid (an empty string for example), it would be a programmatic bug.

src/runtime/pkg/containerd-shim-v2/shim_management.go
fengwang666: @jodh-intel Thanks for catching this! I don't know how I would miss it :(

src/libs/protocols/protos/agent.proto
fengwang666: Thanks! @jodh-intel 
1. The csi spec is under Apache 2.0 license. I assume we don't have any copyright/license issue to copy the code.
2. I have separated the copied ode to its own `csi.proto` following your suggestion. 
3. In term of keeping the file in sync, I think it can be ad-hoc. 
   * Scenario 1: the upstream made a backward-compatible change to the objects we copied (e.g. adding a new optional field), the existing code should still work. We'll need to update the code only if we want to use the new field.
   * Scenario 2: the upstream made a back-incompatible change to the objects (which is unlikely), the code will break at the csi when it returns the volume response. However this failure case always exists.

src/agent/src/rpc.rs
```diff
@@ -1341,6 +1398,39 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
...
+    let mut usage = VolumeUsage::new();
+
+    let s = System::new();
+    for disk in s.disks() {
+        if disk.name().to_str().unwrap().eq(path) {
+            usage.available = disk.available_space();
+            usage.total = disk.total_space();
+            usage.used = usage.total - usage.available;
+            usage.unit = VolumeUsage_Unit::BYTES; // bytes
+        }
```
liubin, line 1411: Lost a `break` to skip remaining disks?

src/agent/src/rpc.rs
```diff
@@ -1341,6 +1398,39 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
...
+    let mut usage = VolumeUsage::new();
+
+    let s = System::new();
+    for disk in s.disks() {
+        if disk.name().to_str().unwrap().eq(path) {
+            let mount_point = disk.mount_point();
+            let meta = fs::metadata(mount_point.to_str().unwrap())?;
+            let inode = meta.ino();
+            usage.used = inode;
+            usage.unit = VolumeUsage_Unit::INODES;
```
liubin, line 1427: The same as above

src/runtime/cmd/kata-runtime/kata-volume.go
```diff
@@ -0,0 +1,143 @@
...
+	Usage:       "directly assign a volume to Kata Container to manage",
+	Subcommands: volumeSubCmds,
+	Action: func(context *cli.Context) {
+		cli.ShowSubcommandHelp(context)
+	},
+}
+
+var addCommand = cli.Command{
+	Name:  "add",
+	Usage: "add a direct assigned block volume device to the Kata Container runtime",
```
liubin, line 43: Nit: `Kata Container` -> `Kata Containers`

src/runtime/cmd/kata-runtime/kata-volume.go
```diff
@@ -0,0 +1,143 @@
...
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "volume-path",
+			Usage:       "the target volume path the volume is published to",
+			Destination: &volumePath,
+		},
+		cli.StringFlag{
+			Name:        "mount-info",
+			Destination: &mountInfo,
```
liubin, line 52: Nit: volume path has a Usage field, but mount-info has not.

src/runtime/cmd/kata-runtime/kata-volume.go
```diff
@@ -0,0 +1,143 @@
...
+	Usage: "resize a direct assigned block volume",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "volume-path",
+			Usage:       "the target volume path the volume is published to",
+			Destination: &volumePath,
+		},
+		cli.Uint64Flag{
+			Name:        "size",
+			Destination: &size,
```
liubin, line 106: Nit: Add Usage field.

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -126,6 +138,52 @@ func decodeAgentMetrics(body string) []*dto.MetricFamily {
...
+		w.Write([]byte(err.Error()))
+		return
+	}
+	w.Write(buf)
+}
+
+func (s *service) serveVolumeResize(w http.ResponseWriter, r *http.Request) {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		shimMgtLog.WithError(err).Error("failed to read request body")
```
liubin, line 163: Nit: This block occurs 3 times, considering using an anonymous function.

```
shimMgtLog.WithError(err).Error("...")
w.WriteHeader(http.StatusInternalServerError)
w.Write([]byte(err.Error()))
```

src/tools/agent-ctl/src/client.rs
```diff
@@ -1644,6 +1649,29 @@ fn agent_cmd_sandbox_get_oom_event(
...
+) -> Result<()> {
+    let req: VolumeStatsRequest = utils::make_request(args)?;
+
+    let ctx = clone_context(ctx);
+
+    debug!(sl!(), "sending request"; "request" => format!("{:?}", req));
+
+    let reply = client
+        .get_volume_stats(ctx, &req)
+        .map_err(|e| anyhow!("{:?}", e).context(ERR_API_FAILED))?;
```
liubin, line 1667: Nit: `anyhow!("{:?}", e)`  -> `anyhow!(e)`

src/runtime/virtcontainers/container.go
```diff
@@ -695,6 +696,28 @@ func (c *Container) createBlockDevices(ctx context.Context) error {
...
+		if mntInfo != nil {
+			// Write out sandbox info file on the mount source to allow CSI to communicate with the runtime
+			if err := volume.RecordSandboxId(c.sandboxID, m.Source); err != nil {
+				c.Logger().WithError(err).Error("error writing sandbox info")
+			}
+
+			c.mounts[i].Source = mntInfo.Device
+			c.mounts[i].Type = mntInfo.FsType
+			c.mounts[i].Options = mntInfo.Options
+			m.Source = mntInfo.Device
```
liubin, line 716: Nit: what's the difference between `c.mounts[i]` and `m`? if they are the same, can be unified into one style.

Kvasscn:
A side question, is it possible to support attaching qemu virtual disk image (qcow2 or raw format) to  kata container?  @fengwang666  @ddebroy  
If not supported currently,   what additional work needs to be done?  Thanks.

src/runtime/virtcontainers/container.go
fengwang666: m is a copy of the mount object that is mutated and used to create device. But here we also want to mutate the original mount object. That's why you see this "duplicate" code. 
I can change it to something like this if it's more readable:
```
			m.Source = mntInfo.Device
			m.Type = mntInfo.FsType
			m.Options = mntInfo.Options
			c.mounts[i].Source = m.Source
			c.mounts[i].Type = m.Type
			c.mounts[i].Options = m.Options
```

src/runtime/virtcontainers/kata_agent.go
egernst: Deep, I think your premise is good here, thought we don't know about the volumes at sandbox creation time, only at sandbox.

IMO, the sandbox should be in charge of adding the volume, and at container creation, it should call into sandbox to say "please make this volume available to the guest."

The sandbox then should have a refcount for each volume _source_. If this volume _source_ is new, we should be handling appropriately (either hot plug a new blk device, or bind mount into the virtiofs shared directory), along with directions on where to make available in guest (either by mounting, or based on path from virtiofs shared directory).  If the volume already exists, increment refcount. In either case, finish by returning the in-VM location of the volume to create Container so it can then update the in-guest OCI spec to reflect the in-guest mount source.

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -0,0 +1,104 @@
...
+	if stat != nil && !stat.IsDir() {
+		return fmt.Errorf("%s should be a directory", volumeDir)
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		if err := os.MkdirAll(volumeDir, 0700); err != nil {
+			return err
+		}
+	}
+
+	return ioutil.WriteFile(filepath.Join(volumeDir, mountInfoFileName), []byte(mountInfo), 0600)
```
egernst, line 54: With the current implementation, we don't do any validation that mountInfo is a json string, or if it matches our expectations for being serialized data for a mount info struct?

Before blindly writing to file, would it be better to attempt to unmarshal this, return errors if any, and then write it to our mountInfo.json?

src/runtime/virtcontainers/qemu.go
```diff
@@ -1266,6 +1266,10 @@ func (q *qemu) qmpShutdown() {
 
 func (q *qemu) hotplugAddBlockDevice(ctx context.Context, drive *config.BlockDrive, op Operation, devID string) (err error) {
 	// drive can be a pmem device, in which case it's used as backing file for a nvdimm device
+	q.Logger().WithFields(logrus.Fields{
```
egernst, line 1269: perhaps as separate commit?

src/runtime/virtcontainers/kata_agent.go
fengwang666: @egernst @ddebroy @bergwolf and I had an offline discussion on this. The conclusion is, the CSI driver should be able to indicate appropriate mount permission (and the permission for device passing to hypervisor). The volume should be refcounted and the filesystem should be only mounted once inside the guest. The changes have been made in the latest commit of this PR.

src/runtime/pkg/direct-volume/utils_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"errors"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/uuid"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAdd(t *testing.T) {
```
egernst, line 19: Saw you added the check in the Add function -- can we add a test case to exercise this?

assert.Err(t, Add(volumePath, "{this actually isn't the json you wanted"}). ?

src/agent/src/rpc.rs
```diff
@@ -1254,6 +1258,59 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+        match get_volume_capacity_stats(&req.volume_guest_path) {
+            Ok(u) => {
+                usage_vec.push(u);
+            }
+            Err(e) => {
+                info!(sl!(), "failed to get volume capacity stats");
+                return Err(ttrpc_error!(ttrpc::Code::INTERNAL, e));
+            }
+        }
+
```
sameo, line 1295: This is equivalent to
```rust
get_volume_capacity_stats(&req.volume_guest_path)
            .map(|u| usage_vec.push(u))                                                                                                                                                                                                                                                                       kata_agent::rpc fn get_volume_capacity_stats(path: &str) -> Result<VolumeUsage>  get_volume_capacity_stats
            .map_err(|e| ttrpc_error!(ttrpc::Code::INTERNAL, e))?;
```

src/agent/src/rpc.rs
```diff
@@ -1254,6 +1258,59 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+        match get_volume_inode_stats(&req.volume_guest_path) {
+            Ok(u) => {
+                usage_vec.push(u);
+            }
+            Err(e) => {
+                info!(sl!(), "failed to get volume inode stats");
+                return Err(ttrpc_error!(ttrpc::Code::INTERNAL, e));
+            }
+        }
+
```
sameo, line 1306: Same as above.

src/agent/src/rpc.rs
```diff
@@ -1254,6 +1258,59 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+        let mut condition = VolumeCondition::new();
+
+        match File::open(&req.volume_guest_path) {
+            Ok(_) => {
+                condition.abnormal = false;
+                condition.message = String::from("OK");
+            }
+            Err(err) => {
+                condition.abnormal = true;
+                condition.message = err.to_string();
```
sameo, line 1279: Shouldn't we return a grpc response here if opening the file did not even work?

src/agent/src/rpc.rs
```diff
@@ -1341,6 +1398,41 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
     Ok((size, plug))
 }
 
+fn get_volume_capacity_stats(path: &str) -> Result<VolumeUsage> {
+    let mut usage = VolumeUsage::new();
+
+    let s = System::new();
+    for disk in s.disks() {
+        if disk.name().to_str().unwrap().eq(path) {
```
sameo, line 1404: I think we should return an error if `to_str()` fails.

src/agent/src/rpc.rs
```diff
@@ -1341,6 +1398,41 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
...
+
+    Ok(usage)
+}
+
+fn get_volume_inode_stats(path: &str) -> Result<VolumeUsage> {
+    let mut usage = VolumeUsage::new();
+
+    let s = System::new();
+    for disk in s.disks() {
+        if disk.name().to_str().unwrap().eq(path) {
```
sameo, line 1421: Ditto.

src/runtime/cmd/kata-runtime/kata-volume.go
```diff
@@ -1,3 +1,8 @@
+// Copyright (c) 2022 Databricks Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
```
sameo, line 5: Not a big deal, but that should not be part of this commit.

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -155,7 +155,6 @@ func (s *service) serveVolumeStats(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Write(buf)
-	return
```
sameo: Same here.

src/runtime/pkg/utils/shimclient/shim_management_client.go
```diff
@@ -1,3 +1,8 @@
+// Copyright (c) 2022 Databricks Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
```
sameo, line 5: Again, this is unrelated.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1559,7 +1559,7 @@ func (k *kataAgent) handleBlkOCIMounts(c *Container, spec *specs.Spec) ([]*grpc.
 			}
 			k.Logger().WithFields(logrus.Fields{
 				"original-source": ociMount.Source,
-				"new-source": path,
+				"new-source":      path,
```
sameo, line 1566: That's a gofmt fix for an earlier commit?

src/libs/protocols/protos/agent.proto
```diff
@@ -65,6 +66,8 @@ service AgentService {
 	rpc CopyFile(CopyFileRequest) returns (google.protobuf.Empty);
 	rpc GetOOMEvent(GetOOMEventRequest) returns (OOMEvent);
 	rpc AddSwap(AddSwapRequest) returns (google.protobuf.Empty);
+	rpc GetVolumeStats(VolumeStatsRequest) returns (VolumeStatsResponse);
+	rpc ResizeVolume(ResizeVolumeRequest) returns (google.protobuf.Empty);
```
sameo, line 70: Could we add those 2 new endpoints to the `samples/configuration-all-endpoints.toml` file, please?

src/libs/protocols/protos/csi.proto
```diff
@@ -0,0 +1,60 @@
...
+
+  // Normal volumes are available for use and operating optimally.
+  // An abnormal volume does not meet these criteria.
+  // This field is REQUIRED.
+  bool abnormal = 1;
+
+  // The message describing the condition of the volume.
+  // This field is REQUIRED.
+  string message = 2;
+}
```
yibozhuang, line 60: nit: new line

src/agent/src/rpc.rs
```diff
@@ -1341,6 +1396,41 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
...
+            let mount_point = disk.mount_point();
+            let meta = fs::metadata(mount_point.to_str().unwrap())?;
```
yibozhuang, line 1423: nit: would be good to use `ok_or()` instead of `unwrap()` to convert `Option` to `Result` with descriptive error message.

yibozhuang: or even simpler
```rust
let meta = fs::metadata(disk.mount_point())?;
```

src/agent/src/rpc.rs
fengwang666: Updated. But I am very very new to Rust. Let me know if there is a better way to write this.

fengwang666:
The direct-assigned volume behavior has been tested and verified manually with a Kubernetes cluster and standalone ctr commands by both @egernst and me. Eric is also adding a bash integration test (create a follow up [issue](https://github.com/kata-containers/kata-containers/issues/3829)). cc @fidencio @bergwolf @sameo @ddebroy

egernst (approved):
Thanks Feng.

src/runtime/virtcontainers/container.go
```diff
@@ -638,15 +638,11 @@ func (c *Container) createBlockDevices(ctx context.Context) error {
 			c.mounts[i].Type = mntInfo.FsType
 			c.mounts[i].Options = mntInfo.Options
 			c.mounts[i].ReadOnly = readonly
-			m.Source = mntInfo.Device
-			m.Type = mntInfo.FsType
-			m.Options = mntInfo.Options
-			m.ReadOnly = readonly
```
sameo: Nice, thanks @fengwang666 !

sameo (approved):
LGTM, thanks @fengwang666 !

chestack:
@fengwang666,  volume resize handler has not been implemented yet in kata agent (but only gets volume stats), right?

fengwang666:
@chestack Correct. Feel free to help if you can :)

Issue #3373, opened 2022-01-03
Host Memory Exhaustion Attack from Inside Kata Containers


WatchDoug:
This is Yutian Yang from Zhejiang University. Our team have discovered
a new attack from inside Kata containers, leading to host memory
exhaustion. 

The root cause lies in the Linux kernel, even in the latest version.
Briefly speaking, the kernel memcg does not charge posix locks
allocated by user processes. The bug report and patches can be found
at
https://lore.kernel.org/linux-mm/20210902215519.AWcuVc3li%25akpm@linux-foundation.org/

Unfortunately, we find that even virtualized containers like Kata are
also affected by this bug. With "-o posix_lock" option enabled, Kata
runtime forwards posix lock allocation requests to virtiofsd on the
host. The virtiofsd then allocates posix locks on behalf. Although
virtiofsd processes are limited by memcg, memory consumption of posix
locks in kernel are not properly charged. Attackers inside containers
can thus allocate a huge number of posix locks to run out of all
memory on the node. Note that the number of posix locks are not
limited by rlimit/sysctl by default.

The PoC is contained in the attached file and you could follow these
steps to reproduce the attack.

0. The environment is Ubuntu 20.04 + Kata 2.3.0, but not limited to them.
1. Compile fcntl-kata.c and build the container image.
2. Check that "-o posix_lock" is enabled in Kata configs.
3. Run a kata container with the image and attach to bash in the container.
4. Decompress the garbage.tar.gz in /. The PoC program will allocate
locks on the 1,000 garbage files.
5. Run /root/fcntl-kata.

You should observe that around 2G memory is consumed by the virtiofsd
after the PoC finishes, while memcg charges much less. The uncharged
memory could be much higher and cause DoS if you raise the number of
allocated locks in the PoC.

We also want to discuss whether there is a way to mitigate such problems.
A quick mitigation could be disabling the "-o posix_lock" option. However, 
how can we enable the functionality without triggering kernel bugs before 
they are patched in the kernel code?

[kata-PoC.zip](https://github.com/kata-containers/kata-containers/files/7801853/kata-PoC.zip)

fidencio:
@WatchDoug, thanks for raising the issue.  I see this as a possible CVE, with the issue not being on the Kata Containers itself.
For now, I'd like to ask you to treating this as a CVE, and thus following our CVE process (please, see: https://github.com/kata-containers/community/blob/main/VMT/VMT.md).

[...]

PR #3410, opened 2022-01-09
runtime: Enable locking on virtio-fs by default

### LINKED ISSUE #3408: flock Doesn't Lock With Default Configuration
With default configuration, kata doesn't support locking files using the `flock` syscall. There's not even an error. Consequently, the file is seemingly locked, but it isn't. The file can be seemingly locked in multiple containers at once.

# Expected result

`flock` should work as in runc, in order to avoid a surprise (and potential data corruption) when replacing runc by Kata in, e.g., a Kubernetes Deployment.

# Workaround

Apply the following patch to Kata's `configuration.toml`:

```diff
@@ -154,7 +154,7 @@
 #   ["-o", "arg1=xxx,arg2", "-o", "hello world", "--arg3=yyy"]
 #
 # see `virtiofsd -h` for possible options.
-virtio_fs_extra_args = ["--thread-pool-size=1"]
+virtio_fs_extra_args = ["--thread-pool-size=1", "-o", "flock"]
 
 # Cache mode:
 #
```

# Actual result

When deploying the following example on Kubernetes, you can see in the logs of the two Pods that both seemlingly hold the lock simultaneously.

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: lock
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 128Mi

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  replicas: 2
  selector:
    matchLabels:
      component: test
  template:
    metadata:
      labels:
        component: test
    spec:
      runtimeClassName: kata
      containers:
        - name: debian
          image: debian
          command:
            - bash
            - -c
            - |
              die() {
                  echo "*** $*" 1>&2
                  exit 1
              }

              set -e

              echo "Acquiring lock..."
              exec 42>/lock/lockfile || die "Failed to access lock"
              flock -w 120 42 || die "Timeout waiting for lock" # does NOT work with kata

              echo "Starting sleep ..."
              for i in {1..20}; do sleep 1; echo $i; done
              echo "Done."

              # Lock released by process exit
          volumeMounts:
            - name: lock
              mountPath: /lock
      volumes:
        - name: lock
          persistentVolumeClaim:
            claimName: lock
```

haslersn:
When using the virtiofsd defaults `no_flock` and `no_posix_lock`, the corresponding syscalls `flock` and `fcntl` still succeed, so the caller thinks that the file is locked while it's not. This can lead to data corruption/loss. Thus, I believe `flock` and `posix_lock` are important to enable by default.

Fixes #3408

### DIFF
--- a/src/runtime/Makefile
+++ b/src/runtime/Makefile
@@ -174,7 +174,7 @@ DEFVIRTIOFSCACHE ?= auto
 #
 # see `virtiofsd -h` for possible options.
 # Make sure you quote args.
-DEFVIRTIOFSEXTRAARGS ?= [\"--thread-pool-size=1\"]
+DEFVIRTIOFSEXTRAARGS ?= [\"--thread-pool-size=1\", \"-o\", \"flock\"]
 DEFENABLEIOTHREADS := false
 DEFENABLEVHOSTUSERSTORE := false
 DEFVHOSTUSERSTOREPATH := $(PKGRUNDIR)/vhost-user


fidencio:
This is a big no for me.
Using "-o posix_lock" could lead to https://github.com/kata-containers/kata-containers/issues/3373.

I think you should open an issue to virtiofs in order to have that fixed there, rather than papering over on the kata-containers side: https://gitlab.com/virtio-fs

haslersn:
I removed `-o posix_lock` from the commit. It still fixes #3408, which was only about flock. But the same issue can of course be opened about fcntl.

snir911:
@haslersn following your last comment i'm closing this issue, I'd suggest to create a new one once more progress on this is made

PR #3418, opened 2022-01-11
runtime: it should rollback when failed in Sandbox AddInterface

yangfeiyu20102011:
When Sandbox AddInterface() is called, it may fail after endpoint.HotAttach,
we'd better rollback and call save() in the end.

Fixes: #3419 

Signed-off-by: yangfeiyu <yangfeiyu20102011@163.com>

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -869,15 +869,28 @@ func (s *Sandbox) AddInterface(ctx context.Context, inf *pbTypes.Interface) (*pb
 		return nil, err
 	}
 
-	// Update the sandbox storage
-	s.networkNS.Endpoints = append(s.networkNS.Endpoints, endpoint)
-	if err := s.Save(); err != nil {
-		return nil, err
-	}
+	defer func() {
+		if err != nil {
+			if errDetach := endpoint.HotDetach(ctx, s.hypervisor, s.networkNS.NetNsCreated, s.networkNS.NetNsPath); errDetach != nil {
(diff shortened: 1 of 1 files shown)


yangfeiyu20102011:
@liubin PTAL
Thanks.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -869,15 +869,28 @@ func (s *Sandbox) AddInterface(ctx context.Context, inf *pbTypes.Interface) (*pb
 		return nil, err
 	}
 
+	defer func() {
+		if err != nil {
+			if errDetach := endpoint.HotDetach(ctx, s.hypervisor, s.networkNS.NetNsCreated, s.networkNS.NetNsPath); errDetach != nil {
+				s.Logger().WithField("endpoint-type", endpoint.Type()).Error("rollback hot attaching endpoint failed")
```
jodh-intel, line 875: How about also logging the new error as that might give a clue about what went wrong:

```suggestion
				s.Logger().WithField("endpoint-type", endpoint.Type()).WithError(errDetach).Error("rollback hot attaching endpoint failed")
```

jodh-intel (approved):
Thanks for raising @yangfeiyu20102011 !

It would be great if you could also add a unit test for this change. However, there doesn't appear to be an existing test for this method and it looks like that could be rather awkward, so...

lgtm

src/runtime/virtcontainers/sandbox.go
yangfeiyu20102011: OK，I have fixed it.

jodh-intel (approved):
Thanks @yangfeiyu20102011.

lgtm

devimc (approved):

Issue #3424, opened 2022-01-11
Not correctly closed trace spans in unmountHostMounts function


liubin:
In function  unmountHostMounts, some spans may be not correctly closed for the early return:

https://github.com/kata-containers/kata-containers/blob/076dbe6cea359d890e416e158e8da9ace48e4428/src/runtime/virtcontainers/container.go#L611-L636

PR #3425, opened 2022-01-11
runtime: close span before return from function in case of error

liubin:
Return before closing span will cause invalid spans, so span should
be closed before function return.

Fixes: #3424


jodh-intel (approved):
Thanks @liubin.

lgtm

jcvenegas (approved):

Issue #3429, opened 2022-01-11
test-kata-deploy is failing to cleanup on aks


snir911:
# Description of problem

When trying to remove deployed kata that was deployed with kata-deploy cleaning up process is failing
This is also causing test-kata-deploy to fail

# Expected result

kata and all kata-deploy related resources are removed and cleaned successfully

# Actual result
test-kata-deploy [fails](https://github.com/kata-containers/kata-containers/runs/4760012461?check_suite_focus=true) with :
```
daemonset.apps/kubelet-kata-cleanup created
+ timeout=120
+ waitForLabelRemoval 120
+ wait_time=120
+ sleep_time=5
+ echo 'waiting for kata-runtime label to be removed'
+ [[ 120 -gt 0 ]]
waiting for kata-runtime label to be removed
++ kubectl get nodes --selector katacontainers.io/kata-runtime
++ grep NAME
+ [[ -z NAME                       STATUS   ROLES   AGE     VERSION ]]
+ sleep 5
+ wait_time=115
.
.
.
+ echo 'failed to cleanup'
633
+ return 1
634
+ destroy_aks
```
(on 2.3.1)
when running locally i get on the kubelet-kata-cleanup pod:
`0/1     Error `
`Back-off restarting failed container`
and the container logs:
`/opt/kata-artifacts/scripts/kata-deploy.sh: line 272: /etc/containerd/config.toml: No such file or directory`

# Further information

I think the issue is caused by the fact 8457150684a8bf5920da8de62c6fd9f1035334ce is trying to create the containerd's conf file also on the [clean up pod](https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml) which is missing this path mount and i think runtime bin is also missing

* I encountered this when i sent PR for 2.3.1, however, test-kata-deploy has passed when 2.4.0-alpha1 released although i'd expect it to fail there as well

snir911:
More info:
Although test-kata-deploy [passed](https://github.com/kata-containers/kata-containers/runs/4724384188?check_suite_focus=true)  when 2.4.0-alpha1 PR sent, running the very same kata-deploy-ci image used there locally is also failing at clean up.
Hence i believe kata-deploy is currently failing both in main and stable

snir911:
[...]

PR #3433, opened 2022-01-12
kata-deploy: validate conf file can be created

snir911:
As containerd doesn't exist at cleanup

Fixes: #3429

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -268,7 +268,8 @@ function main() {
 		containerd_conf_file_backup="${containerd_conf_file}.bak"
 	else
 		# runtime == containerd
-		if [ ! -f "$containerd_conf_file" ]; then
+		if [ ! -f "$containerd_conf_file" ] && [ -d $(dirname "$containerd_conf_file") ] && \
+			[ -x $(command -v containerd) ]; then
 			containerd config default > "$containerd_conf_file"
 		fi
 	fi


jodh-intel (approved):

snir911:
The kata-deploy-ci image doesn't contain the changes made by this PR ( in kata-deploy.sh), hence, the test it keeps failing with the same error.
@fidencio @egernst @wainersm is this a known issue? can i somehow workaround just for testing?

fidencio:
> The kata-deploy-ci image doesn't contain the changes made by this PR ( in kata-deploy.sh), hence, the test it keeps failing with the same error. @fidencio @egernst @wainersm is this a known issue? can i somehow workaround just for testing?

The way would be to manually test it, I'm afraid.

snir911:
locally it seems to fix the issue

wainersm:
> The kata-deploy-ci image doesn't contain the changes made by this PR ( in kata-deploy.sh), hence, the test it keeps failing with the same error. @fidencio @egernst @wainersm is this a known issue? can i somehow workaround just for testing?

In that kind of situation I do:
1. build the image and push to my quay.io
1. add a commit on top of the original PR changing the image url to the new location (create a issue to remember that you need to revert that, add a `#TODO`)
1. once the PR is merged, I re-build the image and push to the old location
1. Then I revert the patch of step 2.

Don't know this is the best optimized way possible :)

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -268,7 +268,8 @@ function main() {
 		containerd_conf_file_backup="${containerd_conf_file}.bak"
 	else
 		# runtime == containerd
-		if [ ! -f "$containerd_conf_file" ]; then
```
wainersm, line 271: I'm not expert on kata-deploy so caution with my suggestion: I think this whole if-then-else block needs a refactoring because it does some initialization that are specific to given actions and the actions are called just later.

For example, if `runtime` is `k3s` and action is `cleanup` then copying `containerd_conf_file` to `containerd_conf_tmpl_file` still make sense?

In other words, you fixed one case but others might be broken because the initialization is wrong there.

snir911: right, although in that case it's less urgent as it's not harmful.
Thing is, the difficulty of testing changes on kata-deploy makes it discouraging to touch.
I'd suggest to have currently it as is and backport to stable to unblock the release

amshinde (approved):

PR #3438, opened 2022-01-12
scripts: Use shebang /usr/bin/env bash

haslersn:
Not all distros have `/bin/bash`, e.g. NixOS.

### DIFF
--- a/ci/go-test.sh
+++ b/ci/go-test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright (c) 2020 Intel Corporation
 #
--- a/ci/install_go.sh
+++ b/ci/install_go.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright (c) 2019 Intel Corporation
 #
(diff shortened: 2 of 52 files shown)


dgibson:
Chances are many of these scripts don't actually use bash specific extensions, they just have `#! /bin/bash` from habit.  So a more elegant approach would probably be to change that to `#! /bin/sh` for those scripts that don't specifically require bash.

jcvenegas:
yay for NixOS, @haslersn  could you plase open an issue and add a 
```
Fixes: #<issue_number>
```
as part of the commit boddy.

I hope the next docs help you about the commit format:  https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#github-best-practices

haslersn:
@jcvenegas Done.

jodh-intel (approved):
Thanks @haslersn.

lgtm, although if you want to fix this issue permanents, you should also consider adding a static check in the script below to avoid any new scripts being added that hard-code the shell path:

- https://github.com/kata-containers/tests/blob/main/.ci/static-checks.sh#L1055

jcvenegas (approved):

jcvenegas:
@haslersn  It would be great,  if you are interesting on contribute with an static check for scripts.

Issue #3445, opened 2022-01-13
Provide /proc/config.gz in guest kernels


jodh-intel:
Enable the guest kernel configuration options to provide a `/proc/config.gz` file in the guest kernel to allow the guest environment to determine which kernel configuration options were set when building the guest kernel.

This is useful for debugging, problem determination, and potentially to help us optimise the guest kernels further if desired (see #3399).

Issue originally raised as: https://github.com/kata-containers/packaging/issues/222.

jodh-intel:
Having the ability to query the guest kernel config inside the guest could be useful:

- By the agent/runtime to determine if a particular workload could be run.
- To determine why a particular custom kernel is not working as expected.
- To help compare guest kernels.
- For general debug / problem determination

It also provides some sort of parity with the existing `osbuilder.yaml` file embedded in the images. They help to identify the image whilst this feature helps to identify the kernel (`uname -r` is useful but doesn't really provide enough detail).

Issue #3399, opened 2022-01-06
Switch to full fat kernels with all features enabled


jodh-intel:
# History

In the very early days we used to use a very aggressively optimised kernel that only provided a minimal set of features. Over time, users requested additional features.

# Now

We currently create "mimimal-ish" kernels using the [kernel fragments](https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/kernel). These minimal-ish kernels support lots of useful functionality, but not all.

# Problem

The guest kernels are no longer minimal, but neither do they offer all available features. This makes life difficult for users wanting to try out advanced features such as GPU and QAT.

# Plan

Switch the default architecture kernels to "full fat" by which we mean, "enable everything" (all features), just like a distro kernel does. 

# Impact

There may be boot speed and memory density impacts, but *iff* we do this, there is also the issue of what happens to the current kernel fragments. One answer might be that we again reduce the functionality to an absolute minimum to make it easier for users to compare the default ("full fat") kernel with a minimal one. This needs further discussion.

# Pros and cons

See the presentation I've put together to present to the [Architecture Committee meeting soon](https://etherpad.opendev.org/p/Kata_Containers_2022_Architecture_Committee_Mtgs) here:

- https://gist.github.com/jodh-intel/b9e0d241b6bb1700bc0074dadc3c99b8
  (see the comment at the end if you want to view it as an actual presentation!)

Related: #3146.

jodh-intel:
This was discussed at the [AC meeting today](https://etherpad.opendev.org/p/Kata_Containers_2022_Architecture_Committee_Mtgs).

# Summary

## Existing guest kernel config

The community does care about the [current kernel config](https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/kernel), even though we know this isn't as optimal as it could be. However, as @egernst points out, it's "more optimal than distro kernels".

## A standardised "full" kernel config

@sameo suggested we could identify a trusted distro and copy their kernel config for the "full fat" guest kernel.

## Use-cases

The guest kernel use-cases folk seem to care about are:

- "Just works" (™)
  Make life easy for new users who want to try out various advanced features such as [GPU](https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/GPU-passthrough-and-Kata.md).
- Speed
- Size
- Security

## Alternative approach

@bergwolf has suggested we could:

- Enable loadable modules for the remaining features that we haven't enabled.
[...]

PR #3446, opened 2022-01-13
packaging: Make kernel config accessible to guest

jodh-intel:
Provide the `/proc/config.gz` file in guest kernels that allow the guest
to determine the kernel configuration used to build the running kernel.

Note that since `gunzip` expects to rename the gzip'ed file it operates
on, to use this feature you need to run something like the following in
the container environment:

```bash
# cat /proc/config.gz|gunzip -c
```

Fixes: #3445.

### DIFF
--- a/tools/packaging/kernel/configs/fragments/common/debug.conf
+++ b/tools/packaging/kernel/configs/fragments/common/debug.conf
@@ -1,0 +1,6 @@
+# Debug related settings
+
+# Enable /proc/config.gz to allow the guest to query the kernel
+# configuration used for the running kernel.
+CONFIG_IKCONFIG=y
+CONFIG_IKCONFIG_PROC=y


jodh-intel:
Related: #3399 (the "how to using the existing kernel fragments if we switch to a 'full fat' kernel" part)

And yes, I appreciate the irony of enabling a few more kernel config options to allow us to try to minimise our kernel config :smile:

amshinde (approved):

devimc:
thanks @jodh-intel - could you please fix your commit message? I think it's incomplete 


>the container environment:
>
>\```bash
>\```

jodh-intel:
@devimc - good catch. This is pretty funny: what I'd originally written was:

<pre>
```bash
# cat /proc/config.gz|gunzip -c
```
</pre>

... but `git(1)` treated that root prompt as a comment, so removed it from my commit! :smile:

tools/packaging/kernel/configs/fragments/common/base.conf
```diff
@@ -55,3 +55,7 @@ CONFIG_PERF_EVENTS=y
 # CONFIG_BLK_DEV_LOOP needs
 CONFIG_BLOCK=y
 CONFIG_BLK_DEV=y
+
+# Enable /proc/config.gz
```
jcvenegas, line 59: Hey the addition looks goot to me about the place,  I dont know if having it in a different fragment file `debug.conf` ? may help to semantically.  But this is not a hard requirement.

jcvenegas (commented):
Hello James aside my comment that is not required, I am not sure, but as far as I remember the CI should for a bump in the number from https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kernel/kata_config_version  at least to make invalid the kernel cache ci. I dont know why the CI did not catch it.  But just in case please increase the value in that file.

egernst:
Just curious; why is this helpful/needed?

jodh-intel:
@egernst - I've added some more verbiage to #3445 with thoughts on this.

tools/packaging/kernel/configs/fragments/common/base.conf
jodh-intel: @jcvenegas - Agreed. Branch updated.

snir911:
Removing dnm label as the comment has been resolved @jodh-intel FYI
LGTM, I'm fine with merging.. thanks!

Issue #3443, opened 2022-01-13
kata-deploy action is used against wrong branch


snir911:
# Description of problem

Running **test-kata-deploy** on **stable** branch is actually running against **main**, generates image and tarball containing **main** branch artifacts

# Expected result

**test-kata-deploy** running on **stable** should run against **stable** and generate **stable** artifacts

# Actual result

**test-kata-deploy** running on **stable** generates **main** based artifices (and IIUC also runs against **main** branch)

# Further information
To validated
1. take the kata-deploy-ci image hash used for a test-kata-deploy action running on stable branch
2. run container using this image and inspect the artifacts (under /opt/) e.g
`docker run -it quay.io/kata-containers/kata-deploy-ci:97e18cf2d0d1a83b139c5117299c1766b6365a22 bash`
(taken from [2.3.1 action](https://github.com/kata-containers/kata-containers/runs/4760012461?check_suite_focus=true))

* It means any kata-deploy changes currently done on stable won't affect (changes on main will)
* Also see, https://github.com/fidencio/kata-containers/commit/edabc1131f45b0c151b0118fd16efd894486e2c8

fidencio:
This is a quite interesting one.

The action used for [testing kata-deploy](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/kata-deploy-test.yaml) is based on the action used to [build the kata-deploy content](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/kata-deploy-push.yaml).  So, both should work in a similar way, right?

Well, they do, apart from this part here:

* [From the test-kata-deploy](https://github.com/kata-containers/kata-containers/blob/7120c789469ed81f2c5cbe4e05960d78c2ca77b5/.github/workflows/kata-deploy-test.yaml#L1-L2):
```
on:
  issue_comment:
    types: [created, edited]
```

* [From the action used to build kata-deploy](https://github.com/kata-containers/kata-containers/blob/7120c789469ed81f2c5cbe4e05960d78c2ca77b5/.github/workflows/kata-deploy-push.yaml#L3):
```
on: [push, pull_request]
```

When taking a look at the [GitHub Actions documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment), I have the amazing suprise that the GITHUB_REF passed is the "default branch".  So, the action is behaving as it's coded to behave.

[...]

PR #3463, opened 2022-01-17
workflows: Use the correct branch ref on test kata-deploy

fidencio:
The action used for testing kata-deploy is entirely based on the action
used to build the kata-deploy tarball, but while the latter is able to
use the correct branch, the former always uses `main`.

This happens as the `issue_comment`, from GitHub actions, passed the
"default branch" as the GITHUB_REF.

As we're not the first ones to face such a issue, I've decided to take
one of the approaches suggested at one of the checkout's issues,
https://github.com/actions/checkout/issues/331, and take advantage of a
new action provided by the community, which will get the PR where the
comment was made, give us that ref, and that then can be used with the
checkout action, resulting on what we originally wanted.

Fixes: #3443

### DIFF
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -48,7 +48,18 @@ jobs:
           - rootfs-initrd
           - shim-v2
     steps:
+      # As Github action event `issue_comment` does not provide the right ref
+      # (commit/branch) to be tested, let's use this third part action to work
+      # this limitation around.
+      - name: resolve pr refs
+        id: refs
+        uses: kata-containers/resolve-pr-refs@v0.0.3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ steps.refs.outputs.head_ref }}
       - name: Install docker
         run: |
           curl -fsSL https://test.docker.com -o test-docker.sh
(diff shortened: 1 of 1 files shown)


fidencio:
Here you can see the action in action: https://github.com/fidencio/kata-containers/runs/4843220516?check_suite_focus=true#step:3:459 :-)

fidencio:
I know, in the past we had a convo about pulling community actions into our org, I can do that if that's the preference (and we should document that somehow / somewhere).

.github/workflows/kata-deploy-test.yaml
```diff
@@ -48,12 +48,19 @@ jobs:
           - rootfs-initrd
           - shim-v2
     steps:
+      - name: resolve pr refs
```
jcvenegas, line 51: Could you add a comment that quickly explains why it is done?  Some little help for us in the future  to help us to understand why we add it and when we could remove it.

Something like:
```sh
# The Github action event `issue_comment` does not provide the right ref (commit/branch) from the PR to be tested.
# It is needed to use this third party action until this is issue is fixed
# URL_TO_ISSUE
```

jcvenegas (commented):
Hey @fidencio thanks for the changes, I am OK if it is not forked, but always is what I prefer. Agree, we need a guidelines as community on how much attention  requires running third party actions on how we should handle them.

.github/workflows/kata-deploy-test.yaml
```diff
@@ -48,12 +48,19 @@ jobs:
...
+        id: refs
+        uses: eficode/resolve-pr-refs@main
```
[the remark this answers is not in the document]
fidencio: Fixed. thanks!

.github/workflows/kata-deploy-test.yaml
fidencio: Partially done, thanks!
There's no issue to link, as this seems to be intented behaviour of the `issue_comment` event.

fidencio:
> Hey @fidencio thanks for the changes, I am OK if it is not forked, but always is what I prefer. Agree, we need a guidelines as community on how much attention requires running third party actions on how we should handle them.

Thanks @jcvenegas, both comments were addressed.

fidencio:
Adding the `do-not-merge` label till I test it on my own fork.

jcvenegas (approved):
LGTM thanks!

fidencio:
> Adding the `do-not-merge` label till I test it on my own fork.

Removing as it works as expected: https://github.com/fidencio/kata-containers/runs/4846622070?check_suite_focus=true#step:3:467

bergwolf (approved):
Thanks!

Issue #3489, opened 2022-01-19
vendor: Do not pin a specific version for runc


fidencio:
There's absolutely no reason for us to pin a specific version of runc.
We don't actually use it in our code and it's only vendored here as a
dependency of our dependencies, and we better let them free to use the
version that makes more sense for them.

PR #3490, opened 2022-01-19
vendor: Unpin runc

fidencio:
There's absolutely no reason for us to pin a specific version of runc.
We don't actually use it in our code and it's only vendored here as a
dependency of our dependencies, and we better let them free to use the
version that makes more sense for them.

Fixes: #3489


bergwolf:
@fidencio IIRC there are CVE issues with runc older than `v1.0.3`. That probably makes it worthwhile to pin it to `v1.0.3` instead.

fidencio:
> @fidencio IIRC there are CVE issues with runc older than `v1.0.3`. That probably makes it worthwhile to pin it to `v1.0.3` instead.

The CVE comes from one place where we use it.  We update that place (runc), and we're good to go.  But yeah, better be safe than sorry.  I am closing this one in favour of yours.

PR #3492, opened 2022-01-19
stable-2.3: Update some go dependencies

fidencio:
Let's update containerd to its 1.5.x latest release and together with that get rid of dependencies that were unnecessarily pinned.  While here, let's also add a comment on the only dependency that seems strictly necessary to be pinned.

Fixes: #3489 
This is some sort of backport of: #3490


amshinde (approved):

fidencio:
Closing this one in favour of Tao's PR (and backport).

Issue #3508, opened 2022-01-20
osbuilder: Remove libseccomp package from Dockerfile


ManaSugi:
**Which feature do you think can be improved?**

Currently, the `libseccomp` package is installed in docker container on `alpine` and `clearlinux` for building the kata-agent although the `libseccomp` library is installed by [`ci/install_libseccomp.sh`](https://github.com/kata-containers/kata-containers/blob/main/ci/install_libseccomp.sh) script.

**How can it be improved?**

Remove the `libseccomp` package from Dockerfile of `alpine` and `clearlinux`.

PR #3509, opened 2022-01-20
osbuilder: Remove libseccomp from Dockerfile

ManaSugi:
Remove the libseccomp package from Dockerfile of `alpine` and `clearlinux`
because the libseccomp library is installed by the `ci/install_libseccomp.sh`
script when building the kata-agent.

Fixes: #3508


dborquez (approved):
You need to add the required porting labels. Besides that LGTM, thanks @ManaSugi

egernst (approved):

Issue #3521, opened 2022-01-21
runtime: runtime binaries fail to execute on ppc64le


Amulyam24:
Failure caught on baseline CI
```
06:28:23 kata-runtime: error while loading shared libraries: R_PPC64_ADDR16_HA reloc at 0x00000b1018ca5800 for symbol `' out of range
06:28:23 [install_kata.sh:56] ERROR: kata-runtime kata-env
```

PR #3522, opened 2022-01-21
runtime: rectify passing empty options to -ldflags

Amulyam24:
When no options are passed to `-ldflags`, it considers
the next string(in this case, `$BUILDFLAGS`) as its value.
Fix passing empty values by passing `$KATA_LDFLAGS`
in quotes.

Fixes: #3521

Signed-off-by: Amulya Meka <amulmek1@in.ibm.com>


fidencio:
@Amulyam24, thanks for the contribution.  I'm setting the labels according to my own judgement here but, please, feel free to let me know if you think it doesn't have to be backported.

fidencio (approved):
lgtm, thanks @Amulyam24!

Jakob-Naucke (approved):
@Amulyam24 my bad. FWIW it's not backported yet, so I'll squash that in there.

Issue #3525, opened 2022-01-21
Static checks ignore arguments


Jakob-Naucke:
We run this: https://github.com/kata-containers/tests/blob/cdc0a0a10cd458a0087d7bd724a35b36da81f74f/.ci/jenkins_job_build.sh#L141 but contrary to an easy-to-make assumption, in the Kata repo, this does not run `tests/.ci/static-checks.sh`, it runs `kata-containers/ci/static-checks.sh`, which does this: https://github.com/kata-containers/kata-containers/blob/365e358115824beb055a64767dc23c46dfcd8813/ci/static-checks.sh#L12 rendering the arguments useless. They should be passed.

PR #3330, opened 2021-12-21
stable-2.3 | kata-manager: Retrieve static tarball

Jakob-Naucke:
In `utils/kata-manager.sh`, we download the first asset listed for the
release, which used to be the static x86_64 tarball. If that happened to
not match the system architecture, we would abort. Besides that logic
being invalid for !x86_64 (despite not distributing other tarballs at
the moment), the first asset listed is also not the static tarball any
more, it is the vendored source tarball. Retrieve all _static_ tarballs
and select the appropriate one depending on architecture.

Fixes: #3254
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

Backport of #3271

PR #3526, opened 2022-01-21
stable-2.3 | ci: Pass function arguments in static-checks.sh

Jakob-Naucke:
e.g. when called from the tests repo

Fixes: #3525

Creating on stable first because it is affected right now.

### DIFF
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -36,7 +36,7 @@ run_static_checks()
 	# Make sure we have the targeting branch
 	git remote set-branches --add origin "${branch}"
 	git fetch -a
-	bash "$tests_repo_dir/.ci/static-checks.sh" "github.com/kata-containers/kata-containers"
+	bash "$tests_repo_dir/.ci/static-checks.sh" "$@"
 }
 
 run_go_test()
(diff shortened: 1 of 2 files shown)


ci/lib.sh
```diff
@@ -36,7 +36,7 @@ run_static_checks()
 	# Make sure we have the targeting branch
 	git remote set-branches --add origin "${branch}"
 	git fetch -a
-	bash "$tests_repo_dir/.ci/static-checks.sh" "github.com/kata-containers/kata-containers"
+	bash "$tests_repo_dir/.ci/static-checks.sh" $@
```
fidencio, line 39:
```suggestion
	bash "$tests_repo_dir/.ci/static-checks.sh" "$@"
```
Could you double-quote `$@`?

fidencio:
I was about to start working on this for a different reason, thanks for being faster. :-)

ci/lib.sh
Jakob-Naucke: Huh. I was convinced that that would squash all arguments into one, but shell quoting never ceases to ~confuse the heck out of~ amaze me. Updated.

devimc (approved):

Jakob-Naucke:
we're seeing k8s bringup failures again on ARM, I'll have one more shot /retest-arm

Jakob-Naucke:
seems more stable now, /retest-arm

Jakob-Naucke:
will forward port once I see successes on #3330/#3528

PR #3306, opened 2021-12-19
runtime-rs: Introduce kata-sys-util library crate

jiangliu:
With the runtime-rs related work, there will be code shared by multiple Kata Containers components, or even some crates may be published and consumed by other projects. So introduce kata-types and kata-sys-util to crates to host shared data structures and common code to access system services.

PR #3537, opened 2022-01-22
Refine build/makefile framework for src/libs

jiangliu:
Current src/libs are built as a sub-target of agent, tools/agent-ctl and tools/trace-forwarder. That's not well formed. Let's make src/libs a dedicated main target to simplify the dependency chain.
Also include the kata-type crate, which has been reviewed in #3306, so we could open following-on PRs based on it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@
 # List of available components
 COMPONENTS =
 
+COMPONENTS += libs
 COMPONENTS += agent
 COMPONENTS += runtime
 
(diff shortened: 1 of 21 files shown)


Tim-Zhang (approved):
LGTM, thanks @jiangliu

bergwolf:
This also adds a new crate `kata-types` to `src/libs`. Is it intentional? It is a major change that is not mentioned in the PR title/message.

bergwolf (approved):
lgtm!

Issue #3551, opened 2022-01-25
[kata-monitor] Link kata metrics to kubernetes PODs


fgiudici:
kata-monitor exposes a /metrics endpoint for scraping by Prometheus.
The only label that kata-monitor adds to the raw metrics, is the sandbox_id, which is the pod id of the Container Manager.
When looking at the metrics in Prometheus it becomes hard to link them to the actual kubernetes PODs, as those are identified by the Kubernetes POD UIDs and/or the Kubernetes POD Name and Namespace (both the UID and the tuple {resource_type, Name, Namespace} are enforced to be unique in a kubernetes cluster).
Add the kubernetes POD metadata (UID, Name and Namespace) as additional labels to the metrics we collect.

Issue #3550, opened 2022-01-25
[kata-monitor] kata may miss collecting metrics for the last started kata pod(s)


fgiudici:
# Description of problem

Kubernetes cluster with metrics 2.0 in place (kata-monitor + Prometheus scraping configured).
Sometimes the last started kata pod is not reported in the Prometheus metrics.

# Expected result

All the kata pods up and running in the kubernetes cluster are reported by Prometheus.

# Actual result

The last started kata pod(s) sometimes is not included in the metrics scraping.

# Further information

Checking the responsible kata-monitor instance, the issue is related to how the kata-monitor cache sandbox is maintained.
When a new kata pod is started:
1. kata-monitor detects a fs event on the vc/sbs dir:
    a. it retrieves the sandbox id and adds it to the sandbox cache (**sandbox is in cache!**)
    b. it triggers a sync with the Container Manager (via the CRI) with a delay of `podCacheRefreshDelaySeconds = 5` seconds.
2. the Container Manager sync happens - the sandboxes sent by the Container Manager become the new sandbox cache.
    a. if the sandbox is not ready (yet) for the Container Manager, the sandbox will not be in the list and we will miss it (**sandbox removed from cache!**)

The sandbox will be added back the next time we sync with the Container Manager: this will happens only when we get a fs event on the vc/sbs dir, which will happen only if/when a kata sandbox is stopped or started.

PR #3553, opened 2022-01-25
kata-monitor: simplify sandbox cache management and attach kubernetes POD metadata to metrics

fgiudici:
Currently, kata-monitor detects started and terminated kata pods by monitoring the vc/sbs fs (this makes sense since we will have to access that path to access the sockets there to get the metrics from the shim).
While kata-monitor updates its sandbox cache based on the sbs fs events, it will schedule also a sync with the container manager via the CRI in order to sync the list of sandboxes there.
The container manager will be the ultimate source of truth, so we will stick with the response from the container manager, removing the sandboxes not reported from the container manager.

May happen anyway that when we check the container manager, the new kata pod is not reported yet, and we will remove it from the kata-monitor pod cache. If we don't get any new kata pod added or removed, we will not check with the container manager again, missing reporting metrics about that kata pod (see #3550).

### Proposal
Let's stick with the vc/sbs fs as the source of truth: we will update the cache just following what happens on the vc/sbs fs.
Let's use the container manager connection to get the kube pod metadata associated to the discovered sandboxes, i.e., attach the kube UID, Name and Namespace associated with the sandbox.
With kubernetes pod metadata we will be able to add them as labels to map explicitly the metrics to the kubernetes workloads  (see #3551).
Every time we get a new sandbox from the sbs fs we will try to retrieve the kubernetes pod metadata associated with it. If we fail, we will keep trying (should we give up at some point?): this will address #3550.

### DIFF
--- a/src/runtime/pkg/kata-monitor/cri.go
+++ b/src/runtime/pkg/kata-monitor/cri.go
@@ -113,15 +113,15 @@ func parseEndpoint(endpoint string) (string, string, error) {
 	}
 }
 
-// getSandboxes gets ready sandboxes from the container engine and returns an updated sandboxMap
-func (km *KataMonitor) getSandboxes(sandboxMap map[string]bool) (map[string]bool, error) {
-	newMap := make(map[string]bool)
+// syncSandboxes gets pods metadata from the container manager and updates the sandbox cache.
+func (km *KataMonitor) syncSandboxes(sandboxList []string) ([]string, error) {
 	runtimeClient, runtimeConn, err := getRuntimeClient(km.runtimeEndpoint)
 	if err != nil {
-		return newMap, err
+		return sandboxList, err
 	}
 	defer closeConnection(runtimeConn)
 
+	// TODO: if len(sandboxList) is 1, better we just runtimeClient.PodSandboxStatus(...) targeting the single sandbox
 	filter := &pb.PodSandboxFilter{
 		State: &pb.PodSandboxStateValue{
 			State: pb.PodSandboxState_SANDBOX_READY,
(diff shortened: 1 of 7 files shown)


fgiudici:
I would say we don't want to backport this, open to discuss anyway (so not putting any label for now).

fidencio:

src/runtime/pkg/kata-monitor/monitor.go
```diff
@@ -112,28 +143,27 @@ func (km *KataMonitor) startPodCacheUpdater() {
...
+			sandboxList, err = km.syncSandboxes(sandboxList)
 			if err != nil {
-				monitorLog.WithError(err).Error("failed to get sandboxes")
+				monitorLog.WithError(err).Error("failed to get sandboxes metadata")
 				continue
 			}
-			monitorLog.WithField("count", len(sandboxes)).Info("synced sandbox cache with the container engine")
-			monitorLog.WithField("sandboxes", sandboxes).Debug("dump sandbox cache")
-			km.sandboxCache.set(sandboxes)
+			if len(sandboxList) > 0 {
```
fgiudici, line 159: Here we keep checking with the Container Manager (CRI interface) if we are not able to retrieve the sandbox kubernetes metadata (name and UID).
Should we give up at some point, maybe changing the time we wait using an exponential backoff value? While this would look reasonable to me, would made the code a bit more complex as to do it properly we should track per sandbox failures. Opinions welcome! :-)

src/runtime/pkg/kata-monitor/cri.go
```diff
@@ -113,15 +113,15 @@ func parseEndpoint(endpoint string) (string, string, error) {
...
+// syncSandboxes gets pods metadata from the container manager and updates the sandbox cache.
+func (km *KataMonitor) syncSandboxes(sandboxList []string) ([]string, error) {
 	runtimeClient, runtimeConn, err := getRuntimeClient(km.runtimeEndpoint)
 	if err != nil {
-		return newMap, err
+		return sandboxList, err
 	}
 	defer closeConnection(runtimeConn)
 
+	// TODO: if len(sandboxList) is 1, better we just runtimeClient.PodSandboxStatus(...) targeting the single sandbox
```
fgiudici, line 124: Would this be worth? Should we start querying for the single sandbox only (we already will have the ID now), or just trying to being smart (get the list if we have more than X sandboxes needing metadata), or stick with getting the full list and do the parsing? Opinions welcome :-)

jodh-intel:
@liubin, @egernst - could you tal?

src/runtime/pkg/kata-monitor/cri.go
fgiudici: Not critical, I think could be maybe an improvement in a following patch.

src/runtime/pkg/kata-monitor/monitor.go
fgiudici: This should be improved: we want to give up at some point (see #3705).

src/runtime/pkg/kata-monitor/cri.go
```diff
@@ -132,29 +131,35 @@ func (km *KataMonitor) getSandboxes(sandboxMap map[string]bool) (map[string]bool
...
+					namespace: pod.Metadata.Namespace,
+				})
+
+				sandboxList = removeFromSandboxList(sandboxList, sandbox)
+
+				monitorLog.WithFields(logrus.Fields{
+					"Pod Name":      pod.Metadata.Name,
+					"Pod Namespace": pod.Metadata.Namespace,
+					"Pod UID":       pod.Metadata.Uid,
+				}).Debugf("Synced KATA POD %s", pod.Id)
```
jodh-intel, line 156:
```suggestion
				monitorLog.WithFields(logrus.Fields{
					"Pod Name":      pod.Metadata.Name,
					"Pod Namespace": pod.Metadata.Namespace,
					"Pod UID":       pod.Metadata.Uid,
					"Pod ID":         pod.Id,
				}).Debug("Synced pod")
```

But also, it's much easier for consumers of these logs if we avoid spaces in fields (and use dashes instead).

fgiudici: Yep, I will fix this in a follow-up PR. Regarding #3705, while here we are still enforcing CRI , now that we basically do detection based on FS events only, we can make CRI optional. Something I will take care in a follow-up PR soon.

PR #3554, opened 2022-01-25
stable-2.3 | kata-monitor: increase delay before syncing with the container manager

fgiudici:
This is meant to address #3550 on the stable branch.

### Proposal
We are stable, so we probably don't want to backport #3553 here.
Since we use the sync with the container manager basically to just remove pods from the cache, we can wait some more time before syncing (and so reduce the chance to miss a kata pod just because it was not ready yet) without missing reporting any metric. The sync will just clean-up the sandbox cache from stale entries.

So, let's just raise the waiting time before starting the sync timer to an higher value: this will not fix the issue (if a pod has not completely started when the sync timer fires, we will miss its metrics till a new kata pod starts / is removed) but will make the issue to be much more unlikely to happen.

### DIFF
--- a/src/runtime/pkg/kata-monitor/monitor.go
+++ b/src/runtime/pkg/kata-monitor/monitor.go
@@ -24,7 +24,7 @@ const (
 	RuntimeContainerd           = "containerd"
 	RuntimeCRIO                 = "cri-o"
 	fsMonitorRetryDelaySeconds  = 60
-	podCacheRefreshDelaySeconds = 5
+	podCacheRefreshDelaySeconds = 60
 )
 
 // SetLogger sets the logger for katamonitor package.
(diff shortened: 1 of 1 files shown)


c3d (approved):
Looks good to me.

Wonder if we could make that a configurable option in a later iteration?

fidencio:

fgiudici:
Thanks for reviewing!

> Wonder if we could make that a configurable option in a later iteration?

Yeah, I had that thought too.
I think that the proper fix is to stick with the sbs fs to populate the sandbox cache. In this way we could even completely drop the sync with the container manager: looked a bit too intrusive for a stable branch ;-) (in #3553 we just use the container manager to get the kubernetes metadata to attach to the metrics).
Here my idea is to just do the bare minimum, i.e., put a delay high enough to be confident that the pod will be running once we check the container manager (*if* no issues on startup happen).
I would just leave this static and move to the #3553 solution (adding a new parameter here shouldn't be that hard anyway).

fidencio (approved):
lgtm as a workaround and I do agree it's not worth pulling the full fix down to the stable branch.
Thanks, @fgiudici!

Issue #3063, opened 2021-11-18
container process send kill -15 0 will cause kata-agent terminated


lifupan:
# Get your issue reviewed faster

Since the container without tty console would share the same process group with kata-agent, which would
cause the kata-agent terminated when kill the container process's group.

PR #3563, opened 2022-01-28
[backport from main] agent: fix the issue of missing create a new session for container

lifupan:
When the container didn't had a tty console, it would be in a same
process group with the kata-agent, which wasn't expected. Thus,
create a new session for the container process.

Fixes: #3063


bergwolf (approved):

amshinde (approved):

jodh-intel (approved):
Thansk @lifupan.

lgtm

A few thoughts though:

- ooi, how did you find this?
- Ideally, we'd have a test to assert the correct behaviour somehow.
- Have you checked the session behaviour for the guest hooks?

Issue #3549, opened 2022-01-25
CCv0 | doc: Improve `ccv0.sh` and CCv0 how-to document 


stevenhorsman:
Originally, over 6 months ago, when I created [`ccv0.sh`](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/ccv0.sh), it was just intended for me local, based on my scripting of the dev guide and was one of the first things I did was Kata. Over time it has been added to in order to build, configure and test more facets of Confidential Containers and has collected some technical debt, or the debt that was always in it has expanded and people like James Hunt have kindly spotted things that could be done better. As it is still part of the `CCv0` branch I'd like to address some of them before it grows further. Some examples include:
- Change `cd`s for `pushd` and `popd`s, so that running the script doesn't leave you in a strange directory.
- Remove hard coded architectures, eg `rustup target add x86_64-unknown-linux-musl`
- Tighten up the security where we `chmod 777`
- Add support for not running as source
- Updates so it doesn't do `ctr pull` if the image is on the local system already
- Doc and Test running as non-root user (covered by #2879)
- Some of the doc is out of date as development changes have gone in that changed functionality without doing associated doc changes (notably the native rust implementation of image pull)

stevenhorsman:
I've made the changes, but just created the [PR in my fork](https://github.com/stevenhorsman/kata-containers/pull/3) for now as the changes depend on https://github.com/kata-containers/kata-containers/pull/3517, so once that is merged I can create the 'real' PR

stevenhorsman:
closed with #3564 merging

Issue #2879, opened 2021-10-22
CCv0 | doc Update CCv0 how-to document to fix root prompts


stevenhorsman:
Following some feedback from @jodh-intel about the [CCv0 doc](https://github.com/kata-containers/kata-containers/blob/CCv0/docs/how-to/how-to-build-and-test-ccv0.md)

> We've built up a set of documentation best practises that are themselves documented here:
> https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md
> In summary, we almost never use a root prompt, use sudo in documents and scripts everywhere possible and try to avoid showing command output (since it tends to change frequently and result in extra maintenance work).
>
> If that script should be run as root, for consistency with the Kata guidelines, can you show the calls as:
>
> $ sudo ~/ccv0.sh shim_pull_image ...
> (or sudo ~/ccv0.sh shim_pull_image ... if referred to inline).
>
> However, if would be good if the script did not need to run entirely as root if possible. If that's the case, always specify in the docs without sudo, but add sudo calls inside the script where essential.

He is completely correct - I've only tested it with a root user, but been sloppy with how I've documented the commands, so the prompts are ambiguous. In order to comply with the standards I need to test it on a not root user and add the sudo's properly as and when required.

stevenhorsman:
Closed with #3564 merging

PR #3564, opened 2022-01-28
CCv0 | doc: Improve ccv0.sh and CCv0 how-to document 

stevenhorsman:
General doc enhancements including:
- Change `cd`s for `pushd` and `popd`s
- Remove hard coded architectures
- Tighten up the security where we `chmod 777`
- Add support for not running as source
- Updates so it doesn't do `ctr pull` if the image is on the
 local system already
- Doc and Test running as non-root user (covered by #2879)
- Update doc to match image_rpc changes

Fixes: #3549
Fixes: #2879

### DIFF
--- a/docs/how-to/ccv0.sh
+++ b/docs/how-to/ccv0.sh
@@ -265,7 +265,7 @@ create_a_local_rootfs() {
     pushd ${katacontainers_repo_dir}/tools/osbuilder/rootfs-builder
     export distro="ubuntu"
     [[ -z "${USE_PODMAN:-}" ]] && use_docker="${use_docker:-1}"
-    sudo -E OS_VERSION="${OS_VERSION:-}" GOPATH=$GOPATH DEBUG="${DEBUG}" USE_DOCKER="${use_docker:-}" SKOPEO=${SKOPEO:-} UMOCI=yes SECCOMP=yes ./rootfs.sh -r ${ROOTFS_DIR} ${distro}
+    sudo -E OS_VERSION="${OS_VERSION:-}" GOPATH=$GOPATH EXTRA_PKGS="vim iputils-ping net-tools" DEBUG="${DEBUG}" USE_DOCKER="${use_docker:-}" SKOPEO=${SKOPEO:-} UMOCI=yes SECCOMP=yes ./rootfs.sh -r ${ROOTFS_DIR} ${distro}
 
      # Install_rust.sh during rootfs.sh switches us to the main branch of the tests repo, so switch back now
     pushd "${tests_repo_dir}"
--- a/docs/how-to/ccv0.sh
+++ b/docs/how-to/ccv0.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 #
-# Copyright (c) 2021 IBM Corporation
+# Copyright (c) 2021, 2022 IBM Corporation
 #
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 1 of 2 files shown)


wainersm:
Hi @stevenhorsman , first of all, very nice and helpful script!

I downloaded the `ccv0.sh` script from that PR, ran and it failed on the `.ci/install_rust.sh` script because in my environment (a fresh Ubuntu 20.04 VM) there isn't `curl`. I was not luck because I wget'ed `ccv0.sh`; If I had tried `curl` I would had it installed :)

I am going to install curl and continue the testing...

stevenhorsman:
> I downloaded the `ccv0.sh` script from that PR, ran and it failed on the `.ci/install_rust.sh` script because in my environment (a fresh Ubuntu 20.04 VM) there isn't `curl`. I was not luck because I wget'ed `ccv0.sh`; If I had tried `curl` I would had it installed :)
> 
> I am going to install curl and continue the testing...

Sorry about that, I obviously have a slightly different version of Ubuntu provisioned as a fresh machine does have `curl` for me.
I've squashed a commit adding it to the small list of bootstrap packages, so hopefully no-one will fall into this trap again!

stevenhorsman:
Now that PR 3517, which this was based on has merged into `CCv0` I've rebased this, so it needs to be /retest ed

wainersm:
I ran the new ccv0.sh from this PR. Apparently it built and installed everything correctly (`kata-check` passed). I could create the pod as explained in the docs but then it failed to create a container.

The error message:
```
wainersm@vm-ubuntu2004:~$ ~/ccv0.sh crictl_create_cc_container
FATA[0005] creating container: rpc error: code = Internal desc = failed to unpack image: ExitStatus(ExitStatus(256)) 
```

docs/how-to/ccv0.sh
```diff
@@ -421,7 +466,7 @@ open_kata_console() {
 
 open_kata_shell() {
     get_ids
-    sudo kata-runtime exec ${sandbox_id}
+    sudo -E env "PATH=$PATH" kata-runtime exec ${sandbox_id}
```
wainersm, line 469: Just curious why it needs `env` here.

stevenhorsman: That's a really good question and one that I don't have an answer to, so I assume I made an error typing in that file rather than the doc or something. I'll remove it now. Thanks for the catch!

stevenhorsman:
> 
```
wainersm@vm-ubuntu2004:~$ ~/ccv0.sh crictl_create_cc_container
FATA[0005] creating container: rpc error: code = Internal desc = failed to unpack image: ExitStatus(ExitStatus(256)) 
```

So there is bug/not-nicety in the `image_rpc.rs` native rust pull image implementation that means if the pull fails then apart from it getting logged in the agent console (which is difficult to access) the problem is ignored, so shows up as a failed to unpack (and the image is probably not there to unpack).

I'm guessing there is a network issue or something that means the pull image is failing, but I think the easier course of action is to get on a video conference and debug it together if that is okay with you @wainersm ?

sameo (approved):

stevenhorsman:
Thanks for the review Samuel, I'll just give @wainersm a day to comment as he didn't finish his review yet, before merging.

wainersm:
> > 
> 
> ```
> wainersm@vm-ubuntu2004:~$ ~/ccv0.sh crictl_create_cc_container
> FATA[0005] creating container: rpc error: code = Internal desc = failed to unpack image: ExitStatus(ExitStatus(256)) 
> ```
> 
> So there is bug/not-nicety in the `image_rpc.rs` native rust pull image implementation that means if the pull fails then apart from it getting logged in the agent console (which is difficult to access) the problem is ignored, so shows up as a failed to unpack (and the image is probably not there to unpack).
> 
> I'm guessing there is a network issue or something that means the pull image is failing, but I think the easier course of action is to get on a video conference and debug it together if that is okay with you @wainersm ?

DNS resolution is what causing that issue and I still couldn't find the root cause. But the following operations on the guest side (connected to the guest via shell) make it work:
```
mount -o remount,rw /etc/resolv.conf
echo 'nameserver 8.8.8.8' > /etc/resolv.conf
```
Let' s have this PR merged, I will test it again in another machine as my VM might be the problem.

wainersm (approved):

PR #3587, opened 2022-02-01
kata-deploy: for testing, make sure we use the PR branch

egernst:
## Changes

This pull request makes two changes to the kata-deploy worfklows.

1. For kata-deploy-test, let's make sure we are exercising changes that are introduced by the PR itself, rather than our base. With this, you can expect changes to our build scripts and/or kata-deploy scripts to be exercised. This doesn't change behavior, however, if changes to the kata-deploy-test workflow itself. Developers will need to continue using a fork to verify these (which is how this PR is tested - details below).
2. For kata-deploy-push:
 *  let's make sure that we only run the workflow if the PR includes changes to ./tools/**. This way, we don't waste cycles for every single PR, and only exercise for use case that we care about (ensuring that ./tools/packaging and ./tools/osbuilder are still running correctly).  This'll reduce our load on GitHub Actions, and should reduce our average queue time for jobs.

## Testing

To verify the change in behavior for (1): 
- In this repo, verify existing behavior, showing that we don't exercise changes, as described in (1). See https://github.com/kata-containers/kata-containers/actions/runs/1781160146, which shows the pipeline succeeding, despite my test PR which breaks the build. 
- In my fork, which is enabled to run the tests, I set *my* main to match this PR. Running same PR on top of this, we can see that the builds are failing, and that the GitHub action is catching this failure: See https://github.com/egernst/kata-containers/runs/5029448370?check_suite_focus=true

### ** Additional test **
To ensure this _also_ works on non-main branch, I tested on my fork, instead with a PR to stable-foobar:
 - action run: https://github.com/egernst/kata-containers/runs/5039331801?check_suite_focus=true#step:5:13
 - PR: https://github.com/egernst/kata-containers/pull/14

To verify (2):
- I have shown following PR which will just run on main without checks, for non ./tools paths:  https://github.com/kata-containers/kata-containers/pull/3588
- With PR in place, we can see that workflow changes are not exercised if path doesn't include ./tools/** in [1] , and that the job will run if tools is touched in [2].
[1] - https://github.com/egernst/kata-containers/pull/12
[2] - https://github.com/egernst/kata-containers/pull/13

### DIFF
--- a/.github/workflows/kata-deploy-push.yaml
+++ b/.github/workflows/kata-deploy-push.yaml
@@ -9,7 +9,9 @@ on:
       - synchronize
       - labeled
       - unlabeled
-  push:
+    paths:
+      - tools/**
+      - versions.yaml
 
 jobs:
   build-asset:
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -48,18 +48,16 @@ jobs:
           - rootfs-initrd
           - shim-v2
     steps:
-      # As Github action event `issue_comment` does not provide the right ref
-      # (commit/branch) to be tested, let's use this third part action to work
-      # this limitation around.
-      - name: resolve pr refs
-        id: refs
-        uses: kata-containers/resolve-pr-refs@v0.0.3
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-
+      - name: get-PR-ref
+        id: get-PR-ref
+        run: |
+            ref=$(cat $GITHUB_EVENT_PATH | jq -r '.issue.pull_request.url' | sed  's#^.*\/pulls#refs\/pull#' | sed 's#$#\/merge#')
+            echo "reference for PR: " ${ref}
+            echo "##[set-output name=pr-ref;]${ref}"
       - uses: actions/checkout@v2
         with:
-          ref: ${{ steps.refs.outputs.base_ref }}
+          ref: ${{ steps.get-PR-ref.outputs.pr-ref }}
+
       - name: Install docker
         run: |
           curl -fsSL https://test.docker.com -o test-docker.sh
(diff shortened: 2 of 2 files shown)


fidencio:
> /test_kata_deploy

This won't have any effect as the action tested is the one that's part of the repo.
In order to test this change, I'd like to ask you to:
* push this to your `main`.
* let someone else open a PR against your main, that has this yaml file committed.

For now, let me add a `do-not-merge` label to ensure this gets properly tested.

.github/workflows/kata-deploy-push.yaml
```diff
@@ -9,7 +9,8 @@ on:
       - synchronize
       - labeled
       - unlabeled
-  push:
+    paths:
+      - tools/**
```
fidencio, line 13: @egernst, I'd prefer having this is a different PR, so we can have more discussions on whether or not to keep this action.

snir911:
/test-fedora
It seems to me it's working as expected, commenting on a PR on a non-main branch made the workflow to pull the actual PR.
However!, i was not able to test it properly so i tested a degraded version of the workflow, if someone will be able to run it through the complete workflow it would be better

fidencio:
@snir911, where you able to test (even with the dummy workflow) on a branch different than the `main` one?

snir911:
Yes, I was using the PR link git presents when i pushed a temporary branch to my own repo and just set it to be a PR on the non-main branch

egernst:
I have tests linked above as well @snir911 - you should be able to see the test against master. As well as duplicated pipeline in my org. 

@fidencio I don’t think the tools path filter is too controversial. If we also want it removed, we can follow up?

fidencio:
@snir911, @egernst, can you please paste the link of a run of `/test_kata_deploy` running against a branch that's different than the `main` one?  Similarly to what's been done the kata-deploy-push tests.

That would simplify immensely the life of other reviewers.

egernst:
@fidencio done - it behaves same as main, since its just checking out the PR.  Updated in PR description under testing. Link here, too: https://github.com/egernst/kata-containers/runs/5039331801?check_suite_focus=true#step:5:13

fidencio (approved):
lgtm, thanks @egernst!

snir911 (dismissed):
Thanks!

.github/workflows/kata-deploy-push.yaml
wainersm: Does kata-deploy read values from `versions.yaml`? If yes, then I think that file should be included in the list of paths.

egernst: The build scripts do reference -- good suggestion.

.github/workflows/kata-deploy-test.yaml
```diff
@@ -48,18 +48,16 @@ jobs:
           - rootfs-initrd
           - shim-v2
     steps:
-      # As Github action event `issue_comment` does not provide the right ref
```
wainersm, line 51: I suggest to rephrase that comment and keep it as it helps a lot to understand why the heck it needs to get the ref.

wainersm (approved):
Hi @egernst ,
I don't have any further comments. Thanks for addressing these issues!

fidencio (approved):
lgtm, thanks @egernst!

PR #3591, opened 2022-02-01
cleaning up the workflow

ege8200:


devimc:
please file a bug and add more details in your commit message

PR #3597, opened 2022-02-02
Make the network API non Linux specific

### LINKED ISSUE #3005: Kata host networking implementation is Linux specific
As part of #3004, we would like to have Kata's sandbox host networking support non-Linux UNIXes.

At the moment, the `Network` Kata implementation is mostly based on Linux networking namespaces and a Kata sandbox network contains `Endpoint` that, for most of them, are Linux specific as well.

Instead we'd like a Kata sandbox `Network` to be defined as an interface and have separate implementations for each operating system that we'd like to see Kata run on top of.

sameo:
This fixes #3005 by removing all Linux specific bits from our Sandbox networking implementation.
In particular, it removes the dependencies to the Linux netns packages and convert the `Network` structure into an interface. That will allow us to define per host OSes implementation for managing host networking.


src/runtime/virtcontainers/sandbox.go
```diff
@@ -803,6 +809,12 @@ func (s *Sandbox) createNetwork(ctx context.Context) error {
 	span, ctx := katatrace.Trace(ctx, s.Logger(), "createNetwork", sandboxTracingTags, map[string]string{"sandbox_id": s.id})
 	defer span.End()
 
+	network, err := NewNetwork(&s.config.NetworkConfig)
```
egernst, line 808: Remind me; we are setting the sandbox's network in both newSandbox and createNetwork -- are these executed in differing flows, or is this duplicated?

src/runtime/virtcontainers/network.go
```diff
@@ -236,6 +236,208 @@ func closeSpan(span otelTrace.Span, err error) {
...
+	}
+
+	n.Endpoints = append(n.Endpoints, endpoint)
+
+	return endpoint, nil
+}
+
+func (n *Network) detachEndpoint(ctx context.Context, s *Sandbox, idx int, hotplug bool) error {
+	if idx > len(n.Endpoints)-1 {
+		return fmt.Errorf("Enpoint index overflow")
```
egernst: nit: s/Enpoint/Endpoint

src/runtime/virtcontainers/network.go
```diff
@@ -196,7 +196,17 @@ type Network struct {
 	NetmonPID int
 }
 
-func NewNetwork(config *NetworkConfig) (*Network, error) {
+func NewNetwork(configs ...*NetworkConfig) (*Network, error) {
+	if len(configs) > 1 {
```
egernst: Can you clarify this part?

src/runtime/virtcontainers/network.go
```diff
@@ -220,6 +221,40 @@ func NewNetwork(configs ...*NetworkConfig) (*Network, error) {
 	}, nil
 }
 
+func LoadNetwork(netInfo persistapi.NetworkInfo) *Network {
```
egernst: Not for changing in this PR per se, but I think it'd be better if we had our own defined type for NetworkInfo in this package, and had persistapi use this (instead of the opposite and relying on that otherwise unrelated package). Would you have any concerns/input on this, if I were to change this in a follow on PR?

egernst (commented):
13 commits of great cleanup. Left a couple of nits/questions. Generally looks good to me.

egernst:
Saw this lint: 
```
Error: network.go:97:18: fieldalignment: struct with 480 pointer bytes could be 464 (govet)
795
``` 

I hate field alignment error reports with the passion of a thousand burning suns.  I am not sure these 26 Bytes of savings are worth it, particularly if we don't have tools for fixing automatically, and because often it sacrifices readability/organization of the struct. Opened https://github.com/kata-containers/kata-containers/issues/3604 to discuss.

src/runtime/virtcontainers/network.go
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) 2016 Intel Corporation
+// Copyright (c) 2022 pple Inc.
```
egernst, line 2: nit: typo: should be Apple

src/runtime/virtcontainers/network.go
```diff
@@ -9,19 +10,12 @@ import (
...
 	"net"
 	"os"
-	"os/exec"
-	"runtime"
-	"sort"
-	"time"
 
-	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
```
egernst, line 17: It seems that you removed *some* of the netlink utilization here -- we still have netlink included for some of the structure definitions which still sit in this file?

src/runtime/virtcontainers/network.go
```diff
@@ -37,6 +31,78 @@ var networkTracingTags = map[string]string{
...
+// NetworkInterfacePair defines a pair between VM and virtual network interfaces.
+type NetworkInterfacePair struct {
+	TapInterface
+	VirtIface NetworkInterface
+	NetInterworkingModel
+}
+
+// NetlinkIface describes fully a network interface.
+type NetlinkIface struct {
+	netlink.LinkAttrs
```
egernst, line 91: Still dependent on netlink?

src/runtime/virtcontainers/sandbox.go
sameo: It's a dup, carried over from the switch between netns to the `Network` structure and then interface.
I pushed an additional commit to simplify `createNetwork()`.

src/runtime/virtcontainers/network.go
sameo: Fixed.

src/runtime/virtcontainers/network.go
sameo: Fixed.

src/runtime/virtcontainers/network.go
sameo: I added a comment in the `network_linux.go` file for clarification:

```
// NewNetwork creates a new Linux Network from a NetworkConfig.
// The constructor is overloaded as it can be called with 0 or 1
// argument. The former is used to create empty networks, mostly
// for unit testing. Passing more than one NetworkConfig pointer
// will make the constructor fail.
```

sameo:
> Saw this lint:
> 
> ```
> Error: network.go:97:18: fieldalignment: struct with 480 pointer bytes could be 464 (govet)
> 795
> ```
> 
> I hate field alignment error reports with the passion of a thousand burning suns. I am not sure these 26 Bytes of savings are worth it, particularly if we don't have tools for fixing automatically, and because often it sacrifices readability/organization of the struct. Opened #3604 to discuss.

FWIW, all linter warnings should be fixed now.

src/runtime/virtcontainers/network.go
sameo: oops. Fixed as well.

egernst (dismissed):

fidencio:
Let me add a do-not-merge label as I really want to go through this (and other big PRs that were opened recently).

jodh-intel:
I haven't reviewed this yet (hopefully Monday!) but for changes like this, we need to keep in mind the potential need to update the docs (and diagrams), in particular API docs and possibly the arch doc (https://github.com/kata-containers/kata-containers/blob/main/docs/design/architecture/networking.md).

bergwolf (approved):
Indeed nice cleanup! Thanks @sameo!

fidencio:
@sameo, first of all, thanks for breaking up the patches in a way those are easy to review, very much appreciated.

I second James comments about updating the documentation and I'll leave it up to your discretion to either do this now, as part of this PR (which I think that would be the best possible case), or as a coming up PR.

Apart from this, lgtm!  I'd like to have this merged sooner than later to avoid conflicts that may come from smaller patches touching this area (I have at least one series that can wait till this one gets merged).

fidencio (dismissed):

fidencio:
(I've added back the `do-not-merge label` so @sameo can remove it depending on his decision to update the docs now or later).

fgiudici:
Just for information: I'm a bit late, but I would like and will take a look at this by EoD (don't want to slow down merging, I will take a look also if it gets merged in the meanwhile).

sameo:
> Just for information: I'm a bit late, but I would like and will take a look at this by EoD (don't want to slow down merging, I will take a look also if it gets merged in the meanwhile).

Thanks for the heads up @fgiudici . I'll wait for your review, take your time.

sameo:
@jodh-intel @fidencio Thanks for the review and the reminder about the documentation update.

This PR does not carry any external API networking changes, but does change `NetworkConfig` (To make it not Linux netns bound) and so I now updated the virtcontainers `api.md` file.

sameo:
> (I've added back the `do-not-merge` label so @sameo can remove it depending on his decision to update the docs now or later).

I updated the docs, but I'll wait for @fgiudici review before removing the label.

fgiudici (approved):
Many thanks @sameo  for waiting on me :-)
I was not that familiar with the code here, but this rework looks quite good to me!
Thanks

egernst (approved):

Issue #3601, opened 2022-02-02
Kata host resources control is Linux specific


sameo:
This is a sub issue for #3005 

Currently, Kata assumes it will and can control host resources through a Linux API (`cgroups`). 

Other OSes may have vastly different ways of controlling how processes can consume host resources, and in order to build and run Kata on non Linux hosts, we should abstract the current cgroup structure into an interface. Each supported host OS could then provide its own implementation.

PR #3602, opened 2022-02-02
Define a resource controller interface

sameo:
This is an implementation for #3601 .

We create a more generic `ResourceController` interface and the Linux implementation is obviously based on `cgroups`. That will allow other host OSes to define their own host resource control implementations.


src/runtime/virtcontainers/pkg/cgroups/cgroups_linux.go
```diff
@@ -0,0 +1,310 @@
+// Copyright (c) 2021 Apple Inc.
```
egernst: HAPPY NEW YEAR!

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2109,7 +2113,7 @@ func (s *Sandbox) cgroupsDelete() error {
 		return nil
 	}
 
-	sandboxCgroup, err := cgroups.Load(s.state.SandboxCgroupPath)
+	sandboxCgroup, err := cgroups.LoadCgroup(s.state.SandboxCgroupPath)
```
egernst: That call is Linux specific, though?

src/runtime/virtcontainers/pkg/cgroups/utils.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2020 Intel Corporation
```
egernst, line 1: Not entirely sure -- does this need updating?

src/runtime/virtcontainers/pkg/cgroups/cgroups_linux.go
sameo: Year of the Tiger!

Fixed now, thanks :-)

src/runtime/virtcontainers/sandbox.go
sameo: This is replaced with

```golang
sandboxController, err := resCtrl.LoadResourceController(s.state.SandboxCgroupPath)
```

which is arguably less Linux specific, hopefully.

src/runtime/virtcontainers/pkg/cgroups/utils.go
sameo: I'll let Intel decide about it?

src/runtime/virtcontainers/pkg/cgroups/cgroups.go
```diff
@@ -6,24 +6,22 @@
...
-type Cgroup struct {
-	cgroup  cgroups.Cgroup
-	path    string
-	cpusets *specs.LinuxCPU
-	devices []specs.LinuxDeviceCgroup
+// prepend a kata specific string to oci cgroup path to
+// form a different cgroup path, thus cAdvisor couldn't
+// find kata containers cgroup path on host to prevent it
+// from grabbing the stats data.
+const CgroupKataPrefix = "kata"
```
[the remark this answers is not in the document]
sameo: Since it's only ever used by `cgroups.go` I thought it would be a good place for it.
I moved the `const DefaultResourceControllerID = "/vc"` definition to `utils_linux.go` though, as it's used outside of the package.

src/runtime/virtcontainers/pkg/resourcecontrol/controller.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2021 Apple Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package cgroups
```
egernst, line 6: should the package name be updated as well?

src/runtime/virtcontainers/sandbox.go
```diff
@@ -37,9 +37,9 @@ import (
 	pbTypes "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/annotations"
-	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/cgroups"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/compatoci"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/cpuset"
+	resCtrl "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/resourcecontrol"
```
egernst, line 42: No change needed, but curious -- any reason you think this should live here at ./virtcontainers/pkg/resourcecontrol instead of at ./pkg/resourcecontrol?

src/runtime/virtcontainers/sandbox.go
```diff
@@ -670,30 +671,30 @@ func (s *Sandbox) createCgroups() error {
...
+	s.state.SandboxCgroupPath = s.sandboxController.ID()
 	s.state.OverheadCgroupPath = ""
 
 	if s.config.SandboxCgroupOnly {
-		s.overheadCgroup = nil
+		s.overheadController = nil
 	} else {
 		// The shim configuration is requesting that we do not put all threads
 		// into the sandbox cgroup.
 		// We're creating an overhead cgroup, with no constraints. Everything but
```
egernst, line 688: while we're updated from cgroups to a more generic `resource-controller` interface, I do note that most of the comments (and some naming) at the sandbox level are about cgroups.  Since we don't have any other implementations of resource controller introduced to Kata at this point, perhaps we can address this part at a later time, but there'll be a disconnect here?

fidencio:
Adding a `do-not-merge` label to ensure folks have the needed time to carefully go through this one.

src/runtime/virtcontainers/sandbox.go
sameo: No specific reasons, besides, as you write, the fact that it was there in the first place.
I think the main reason for moving it to `runtime/pkg` is to have one common places for all runtime packages, and that's a good enough reason. I added a top commit for that purpose, if we decide we don't want to move it there, I can pop it up.

src/runtime/virtcontainers/sandbox.go
sameo: All those are good points.
So I added a `Type` method to the `ResourceController` interface, and tried to make a lot of the `sandbox.go` comments and error strings less cgroups specific. Let me know if it looks better to you.

src/runtime/virtcontainers/pkg/resourcecontrol/controller.go
sameo: Yes :) Fixed as well.

src/runtime/pkg/resourcecontrol/controller.go
```diff
@@ -0,0 +1,82 @@
...
+package resourcecontrol
+
+import (
+	v1 "github.com/containerd/cgroups/stats/v1"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	controllerLogger = logrus.WithField("source", "virtcontainers/pkg/cgroups")
```
jodh-intel, line 15: Should this be instead?:

```suggestion
	controllerLogger = logrus.WithField("source", "virtcontainers/pkg/resourcecontrol")
```

jodh-intel (approved):
Thanks @sameo! lgtm

Random bits of bike-shedding:

- We could call this "Resource Manager" rather than "Resource Controller".
- Once the term and it's acronym have been defined, I don't see why we couldn't shorten some of the names to `ResMan` / `ResCtl` (or even just `RM` or `RC`) depending on the first bullet.

src/runtime/pkg/resourcecontrol/controller.go
sameo: You're right, fixed now. Thanks.

sameo:
> Thanks @sameo! lgtm
> 
> Random bits of bike-shedding:
> 
>     * We could call this "Resource Manager" rather than "Resource Controller".

I gave the naming of this package too many of my few remaining brain cells :)
I wanted to have something that is more generic than cgroup, but still reflects the fact that we'd be controlling resources. Hence `ResourceController` as I feel it's closer to a controller than a manager.

> 
>     * Once the term and it's acronym have been defined, I don't see why we couldn't shorten some of the names to `ResMan` / `ResCtl` (or even just `RM` or `RC`) depending on the first bullet.

That's very much arguable but when reading a new piece of code (from kata or otherwise) I feel like name shortening add to the cognitive effort of already having to understand a new code base. I know it's chatty, but I personally feel using full names is helping people trying to get familiar with (and e.g. review) the code base. It's quite a personal opinion though.

jodh-intel (approved):
Thanks @sameo.

lgtm

jodh-intel:
@sameo - Yeah, it was just a random comment and I'm not disagreeing. Clarity has to win out and I can live with Resource Controller :smile:

egernst (approved):

fidencio (approved):
lgtm, thanks @sameo!

PR #3606, opened 2022-02-02
openshift-ci: switch to CentOS Stream

### LINKED ISSUE #3605: openshift-ci: CentOS 8 based build root stopped working
The transition from CentOS 8 to Stream 8 has broken our CI jobs on OpenShift CI as the current CentOS 8 image cannot get some yum repositories resolved. Therefore, let[ s migrate to Stream 8.

wainersm:
The build root container is switched from CentOS 8 to Stream 8 as
the former reached EOL.

Fixes #3605

### DIFF
--- a/ci/openshift-ci/images/Dockerfile.buildroot
+++ b/ci/openshift-ci/images/Dockerfile.buildroot
@@ -4,7 +4,7 @@
 #
 # This is the build root image for Kata Containers on OpenShift CI.
 #
-FROM registry.centos.org/centos:8
+FROM quay.io/centos/centos:stream8
 
 RUN yum -y update && \
     yum -y install \


wainersm:
Tested on OpenShift CI. Logs: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_release/25938/rehearse-25938-pull-ci-kata-containers-kata-containers-main-images/1489003173925359616/build-log.txt

wainersm:
/test-fedora

^ @fidencio let's see if the Fedora 35 jobs are getting triggered and reporting back correctly.

snir911:
@wainersm wouldn't it be better to use ubi in this case? is it possible? does it make sense?

wainersm:
> @wainersm wouldn't it be better to use ubi in this case? is it possible? does it make sense?

good point @snir911 , I didn't consider UBI and let me check if it provides a suitable environment. Adding /do-not-merge

wainersm:
> > @wainersm wouldn't it be better to use ubi in this case? is it possible? does it make sense?
> 
> good point @snir911 , I didn't consider UBI and let me check if it provides a suitable environment. Adding /do-not-merge

@snir911 UBI is a no-go for this build root because it doesn't provide the kernel packages; and we are adding some kernel modules in the guest image. There might be a workaround but right now I'd like to have the CI jobs back running...so stick with Stream 8 :)

snir911 (approved):

wainersm:
/test-firecracker

As it aborted due to timeout.

fidencio (approved):
lgtm, thanks @wainersm!

PR #3615, opened 2022-02-03
Make the hypervisor framework not Linux specific

### LINKED ISSUE #3614: Kata hypervisor framework is Linux specific
This is a sub issue for #3004 .

A lot of the hypervisor code and the VMM implementations (QEMU, FC, etc) is Linux specific. It assumes Kata will always build and run on top of Linux, even when supporting different hardware architectures.

We should split this code into Linux and OS agnostic code in order to allow for at least supporting other UNIXes.

sameo:
* Move all the Linux specific bits into their own `*_linux.go` files, per architecture.
* Build all existing VMMs only on Linux
* Keep the OS agnostic bits in non GOOS files.

This fixes #3614


fidencio:
Adding a `do-not-merge` label to ensure folks have the needed time to carefully go through this one.

src/runtime/pkg/katautils/config.go
```diff
@@ -329,7 +329,7 @@ func (h hypervisor) defaultVCPUs() uint32 {
 
 func (h hypervisor) defaultMaxVCPUs() uint32 {
 	numcpus := uint32(goruntime.NumCPU())
-	maxvcpus := vc.MaxQemuVCPUs()
+	maxvcpus := vc.MaxVCPUs()
```
fidencio, line 332: @sameo, I'm adding the comment here, but it's rather related to 9d8fe3e0e454e6e17a27d5789ee1d69c2f615649 than to this specific part.

I like you're making things that are QEMU specific less QEMU specific, which is a good thinng.  But I wonder whether the values should be coming from GoVMM (that currently lives inside our repo), and I think this refactor is a nice opportunity to also touch this.

WDYT?

fidencio (commented):
@sameo, in general it looks good to me.
While you already have your hands dirty, I've left a comment about re-working 9d8fe3e0e454e6e17a27d5789ee1d69c2f615649 and possibly moving QEMU pecific bits to "GoVMM" (as, differently than what its name suggests, it's QEMU specific).

I **think** documents will need to be reworked as well, but again, I am fine about having those re-worked when the Darwin specific bits start to land.

src/runtime/pkg/katautils/config.go
sameo: I agree. It should look better now, hopefully.
I added those functions to the govmm package and they're all gone from virtcontainers.

Let me know if that was not what you were looking for.

jodh-intel (dismissed):
Nice! Thanks @sameo!

lgtm

egernst (approved):

amshinde (dismissed):

src/runtime/virtcontainers/hypervisor.go
```diff
@@ -894,17 +880,55 @@ func GetHypervisorPid(h Hypervisor) int {
...
+
+	// IBM POWER 9 Protected Execution Facility
+	// https://www.kernel.org/doc/html/latest/powerpc/ultravisor.html
+	// Exclude from lint checking for it won't be used on arm64 code
+	pefProtection
+
+	// IBM Secure Execution (IBM Z & LinuxONE)
+	// https://www.kernel.org/doc/html/latest/virt/kvm/s390-pv.html
+	// Exclude from lint checking for it won't be used on arm64 code
+	seProtection
```
jodh-intel, line 907: Nit: If possible (?) might be worth sorting alphabetically but I see this is simply a verbatim move from another file.

src/runtime/virtcontainers/hypervisor_linux.go
```diff
@@ -0,0 +1,45 @@
...
+	case QemuHypervisor:
+		return &qemu{}, nil
+	case FirecrackerHypervisor:
+		return &firecracker{}, nil
+	case AcrnHypervisor:
+		return &Acrn{}, nil
+	case ClhHypervisor:
+		return &cloudHypervisor{}, nil
+	case MockHypervisor:
+		return &mockHypervisor{}, nil
```
jodh-intel, line 41: Nit: Unsorted list.

jodh-intel (approved):
Thanks @sameo.

egernst:
@fidencio are you able to take a pass at this? Have had a few LGTMs at this point.

fidencio (approved):
lgtm, thanks @sameo!

This, IMHO, is good to go as it's, but I'll leave it up to you to fix the nits pointed by James, or not.
If you happen to do, poke me on Slack for a re-ack on this.

Right now I restarted the failing CIs directly on the Jenkins side.  Please, monitor those.

sameo:
Thanks @fidencio 
@jodh-intel Are you ok merging that one? I'll send a follow up PR right away for addressing your 2 comments.

egernst:
Eek - clicked the green button.  James' items were nits (that existed in prior code), so I don't feel _horrible_, I just feel bad.

PR #3619, opened 2022-02-03
runtime: support QEMU SGX

### LINKED ISSUE #3436: qemu: sgx: add support for SGX
latest version of QEMU supports SGX, see https://gitlab.com/qemu-project/qemu/-/blob/master/docs/system/i386/sgx.rst

devimc:
Enable SGX in QEMU when `sgx.intel.com/epc` annotation is defined

fixes #3436

Signed-off-by: Julio Montes <julio.montes@intel.com>

### DIFF
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -24,6 +24,8 @@ type qemuAmd64 struct {
 	vmFactory bool
 
 	devLoadersCount uint32
+
+	sgxEPCSize int64
 }
 
 const (
@@ -140,6 +142,17 @@ func newQemuArch(config HypervisorConfig) (qemuArch, error) {
 		}
 	}
 
+	if config.SGXEPCSize != 0 {
+		q.sgxEPCSize = config.SGXEPCSize
+		if q.qemuMachine.Options != "" {
+			q.qemuMachine.Options += ","
+		}
+		// qemu sandboxes will only support one EPC per sandbox
+		// this is because there is only one annotation (sgx.intel.com/epc)
+		// to specify the size of the EPC.
+		q.qemuMachine.Options += "sgx-epc.0.memdev=epc0,sgx-epc.0.node=0"
+	}
+
 	q.handleImagePath(config)
 
 	return q, nil
@@ -233,6 +246,16 @@ func (q *qemuAmd64) enableProtection() error {
 
 // append protection device
 func (q *qemuAmd64) appendProtectionDevice(devices []govmmQemu.Device, firmware, firmwareVolume string) ([]govmmQemu.Device, string, error) {
+	if q.sgxEPCSize != 0 {
+		devices = append(devices,
+			govmmQemu.Object{
+				Type:     govmmQemu.MemoryBackendEPC,
+				ID:       "epc0",
+				Prealloc: true,
+				Size:     uint64(q.sgxEPCSize),
+			})
+	}
+
 	switch q.protection {
 	case tdxProtection:
 		id := q.devLoadersCount


devimc:
cc @mythi @eadamsintel

src/runtime/virtcontainers/qemu_amd64.go
```diff
@@ -173,6 +185,10 @@ func (q *qemuAmd64) cpuModel() string {
 		cpuModel += ",vmx=off"
 	}
 
+	if q.sgxEPCSize != 0 {
```
egernst, line 188: Just from reading nearby code -- is there any limitation with respect to migration and SGX?

src/runtime/virtcontainers/qemu_amd64.go
```diff
@@ -140,6 +140,18 @@ func newQemuArch(config HypervisorConfig) (qemuArch, error) {
...
 
+	if config.SGXEPCSize != 0 {
+		q.sgxEPCSize = config.SGXEPCSize
+		if q.qemuMachine.Options != "" {
+			q.qemuMachine.Options += ","
+		}
+		// the same way we did with cloud-hypervisor,
+		// qemu sandboxes will only support one EPC per sandbox
+		// this is because there is only one annotation (sgx.intel.com/epc)
+		// to specify the size of the EPC.
```
egernst, line 151: Can you help clarify -- we could have multiple EPC in a VM, but we aren't supporting it because of how we defined the sgx annotations?

I assume we receive this annotation in a container spec? or, can you please clarify?  Who defines this/these annotations to set this limitation?

Thanks -- just want to make sure I understand!

mythi: The annotation comes from the SGX device plugin (or it can be specified as the pod annotation manually) and it defines the sum of EPC resource requests of all containers in the pod.

We could have multiple sections in a VM but is there a need for it? The kernel EPC allocator is NUMA aware and checks the EPC section topology but the Kata VM is created without numa topology so only one section sounds justified.

src/runtime/virtcontainers/qemu_amd64.go
mythi: https://www.qemu.org/docs/master/system/i386/sgx.html#migration

devimc: thanks @mythi , live migration is supported but guest software stack should support recreating enclaves when it suffers sudden lose of EPC

egernst: Got it. Either way, if an operator / user is configuring SGX, trying to use any of they templating logic, they will have their work cut out for them!

src/runtime/virtcontainers/qemu_amd64.go
egernst: Yeah, makes sense -- just wanted to understand. Thanks Mikko!

egernst: cool. One nit (sorry) - I wouldn't mention CLH here:

```suggestion
		// qemu sandboxes will only support one EPC per sandbox
		// this is because there is only one annotation (sgx.intel.com/epc)
		// to specify the size of the EPC.
```

devimc:
thanks @egernst - changes applied

fidencio (approved):
lgtm, thanks @devimc!

src/runtime/virtcontainers/qemu_amd64.go
```diff
@@ -173,6 +186,10 @@ func (q *qemuAmd64) cpuModel() string {
 		cpuModel += ",vmx=off"
 	}
 
+	if q.sgxEPCSize != 0 && !q.vmFactory {
+		cpuModel += ",+sgx"
```
mythi, line 190: is this necessary with `defaultCPUModel`?

devimc: you are right, thanks :+1:

devimc:
@mythi changes applied, thanks

mythi (approved):
@devimc thanks!

GabyCT (approved):

jodh-intel (approved):
Thanks @devimc.

lgtm

PR #3623, opened 2022-02-04
Container host files sharing abstraction

### LINKED ISSUE #3622: Host filesystem sharing assumes Linux as a host
Kata shares container files (Mounts) and bundle root filesystems with the guest.

Today, this is done through a set of black magic 🪄 sequence of bind mounts into the `shared` path.
Unfortunately, bind mounts are not supported by all UNIXes (Hello DarwinOS...) and make that implementation almost exclusive to Linux.

We should provide a cleaner abstraction (a.k.a. an Interface) for sharing container files from the host with the guest, and allow for all host OSes to provide their own implementation.

sameo:
This PR abstracts how Kata shares its container files (mounts and rootfs) between the host and the guest.
The goal is to make it not Linux specific and allow for other UNIXes to provide their own implementation for it.

First we define a `FilesystemSharer` interface that tries to abstract the way Kata shares container files with the guest into a small set of methods.

Then we implement the Linux support for it, based on a combination of bind mounts and either block based devices and virtiofs. This implementation is an aggregation and hopefully a simplification of different pieces of code spread across `sandbox.go`, `container.go` and `kata_agent.go`

Finally we replace all direct calls to bind mounts and other Linuxisms for sharing files between a host and a guest, with calls to the `FilesystemSharer` methods. 

This PR fixes #3622


src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -0,0 +1,368 @@
...
+	defer func() {
+		if err != nil {
+			if umountErr := unmountNoFollow(sharePath); umountErr != nil {
+				f.Logger().WithError(umountErr).Errorf("failed to unmount vm share path %s", sharePath)
+			}
+		}
+	}()
+
+	// Setup sandbox bindmounts, if specified.
+	if err = f.prepareBindMounts(ctx); err != nil {
```
egernst, line 152: I understand that this is an artifact of how it is written today in `kata_agent`, but I'm not sure how I feel about the sandbox bind mounts being part of the shared filesystem setup in `Preparation`.
 
So, I suppose the alternative would be to make it explicit as part of the fs_share interface, like ShareSandboxFile, which... may be worse than just having it part of Prepare/Cleanup.

(ie, this one is unique from ShareFile, since it isn't dealing with a mount tied to a container).

Curious if you have thoughts on this.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -492,6 +498,10 @@ func createSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Fac
 		return s, nil
 	}
 
+	if err := s.fsShare.Prepare(ctx); err != nil {
+		return nil, err
+	}
+
 	// Below code path is called only during create, because of earlier Check.
```
egernst, line 505: should this comment move above the prepare (and could be clarified?). 

Either way, afaiu, we  only want to  setup the shared filesystem if we are creating a new sandbox, instead of loading the sandbox state, correct?

src/runtime/virtcontainers/fs_share.go
```diff
@@ -0,0 +1,55 @@
...
+	Prepare(context.Context) error
+
+	// Cleanup cleans the host filesystem up from the initial
+	// setup created by Prepare.
+	Cleanup(context.Context) error
+
+	// ShareFile shares a file (a regular file or a directory)
+	// from the host filesystem with a container running in the
+	// guest. The host file to be shared is described by the
+	// Mount argument.
```
egernst, line 40: Should we expand the description a bit to describe the ShareFile return value?

I appreciate setting SharedFile to nil as a way to indicate that we're "ignoring" the given file.

src/runtime/virtcontainers/fs_share_linux.go
sameo: > I understand that this is an artifact of how it is written today in `kata_agent`, but I'm not sure how I feel about the sandbox bind mounts being part of the shared filesystem setup in `Preparation`.
> 
> So, I suppose the alternative would be to make it explicit as part of the fs_share interface, like ShareSandboxFile, which... may be worse than just having it part of Prepare/Cleanup.
> 
> (ie, this one is unique from ShareFile, since it isn't dealing with a mount tied to a container).
> 
> Curious if you have thoughts on this.

I took this approach and decided not to explicitly expose a Sandbox file sharing API because we do it implicitly in today's current code. So having an interface API for that  would mean we'd unconditionally call it for every sandbox creation. Basically I was trying to keep the interface as short as I could, we have many examples in the current code base where we've split our interfaces into methods that we call once, unconditionally, that imho should not be part of the API itself.

But that's not a super strong opinion, almost a matter of taste I guess.

src/runtime/virtcontainers/container.go
```diff
@@ -540,20 +473,18 @@ func (c *Container) mountSharedDirMounts(ctx context.Context, sharedDirMounts, i
...
-		var guestDest string
-		guestDest, ignore, err = c.shareFiles(ctx, m, idx)
+		sharedFile, err := c.sandbox.fsShare.ShareFile(ctx, c, &m)
 		if err != nil {
 			return storages, err
 		}
 
 		// Expand the list of mounts to ignore.
-		if ignore {
+		if sharedFile == nil {
```
egernst, line 482: I like this improvement, thank you.

src/runtime/virtcontainers/container.go
```diff
@@ -580,11 +511,11 @@ func (c *Container) mountSharedDirMounts(ctx context.Context, sharedDirMounts, i
 				return storages, fmt.Errorf("unable to create watchable path: %s: %v", watchableHostPath, err)
 			}
 
-			watchableGuestMount := filepath.Join(kataGuestSharedDir(), "watchable", filepath.Base(guestDest))
+			watchableGuestMount := filepath.Join(kataGuestSharedDir(), "watchable", filepath.Base(sharedFile.guestPath))
```
egernst, line 514: Does watchableGuestMount fit with these reworks?

While it doesn't carry out any additional sharing between the host/guest, it does have some assumptions with respect to paths used on the host/guest, which is otherwise nicely abstracted (I think) into fs_share.

egernst (commented):
Abstracting linux "stuff" aside, I really like having the shared filesystem code in one spot, behind an interface. Couple more questions - thanks!

src/runtime/virtcontainers/sandbox.go
sameo: I clarified and moved the comment, thanks.

Also, I changed the `Prepare` implementation to make it idempotent, and it is now safe to call it several times in a row.

src/runtime/virtcontainers/fs_share.go
sameo: Yes, good point. I added some additional comments/documentation to the interface.

src/runtime/virtcontainers/container.go
sameo: I think it does fit, and the guest paths assumptions are now carried through the `SharedFile` return value, which indicates where in the guest the mount is going to live.
Do you think there's anything wrong with the implementation?

src/runtime/virtcontainers/fs_share.go
```diff
@@ -0,0 +1,66 @@
...
+	guestPath string
+}
+
+type FilesystemSharer interface {
+	// Prepare will set the host filesystem up, making it ready
+	// to share container files and directories with the guest.
+	// Implementation of this method must be idempotent and be
+	// ready to potentially be called several times in a row,
+	// without symetric calls to Cleanup in between.
+	Prepare(context.Context) error
```
c3d, line 34: Could you give examples of what this could do. I _guess_ it would include unmounting host filesystems and things like that, but I'm not entirely sure that's the intent.

Also, it might be interesting to watch https://github.com/kubernetes/enhancements/pull/2893 while you are working in that space.

src/runtime/virtcontainers/fs_share.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// ShareFile shares a file (a regular file or a directory)
+	// from the host filesystem with a container running in the
+	// guest. The host file to be shared is described by the
+	// Mount argument.
+	// The returned SharedFile pointer describes how the file
+	// should be shared between the host and the guest. If it
+	// is nil, then the shared filed described by the Mount
+	// argument will be ignored by the guest, i.e. it will NOT
+	// be shared with the guest.
```
c3d, line 51: Two questions:
1. Is it OK to call it multiple times with different `Mount`s, and if so, what is the result
2. Since this returns a `SharedFile`, this means the implementations decide the strategy on what can be shared. Doesn't that imply that the description should be "i.e. it _can_ NOT been shared with the guest" instead of "it _will_ not"?

src/runtime/virtcontainers/fs_share.go
```diff
@@ -0,0 +1,66 @@
...
+	ShareFile(context.Context, *Container, *Mount) (*SharedFile, error)
+
+	// UnshareFile stops sharing a container file, described by
+	// the Mount argument.
```
c3d, line 55: Shouldn't you pass the `SharedFile` result in case an implementation needs it for some cleanup?

src/runtime/virtcontainers/fs_share.go
```diff
@@ -0,0 +1,66 @@
...
+	UnshareFile(context.Context, *Container, *Mount) error
+
+	// ShareRootFilesystem shares a container bundle rootfs with
+	// the Kata guest, allowing the kata agent to eventually start
+	// the container from that shared rootfs.
+	ShareRootFilesystem(context.Context, *Container) (*SharedFile, error)
+
+	// UnshareRootFilesystem stops sharing a container bundle
+	// rootfs.
+	UnshareRootFilesystem(context.Context, *Container) error
```
c3d, line 65: Same comment as for `UnshareFile`, may be safer to pass a `SharedFile` argument even if some implementations don't use it.

jiangliu:
Soon nydus will get a fuse backend for MacOS so nydus will worker on MacOS too:)
Please help to review https://github.com/cloud-hypervisor/fuse-backend-rs/pull/51

src/runtime/virtcontainers/fs_share.go
sameo: > Could you give examples of what this could do. I _guess_ it would include unmounting host filesystems and things like that, but I'm not entirely sure that's the intent.
> 

A further commit implements that interface for Linux, so I guess it's a good example :)
I added more comments to briefly describer the Linux case.

src/runtime/virtcontainers/fs_share.go
sameo: > Two questions:
> 
>     1. Is it OK to call it multiple times with different `Mount`s, and if so, what is the result

It's more than ok, it's what implementation would do. The description states that this method is for sharing one file (regular or directory), and that the file is described through the `Mount` argument:

```
        // ShareFile shares a file (a regular file or a directory)
        // from the host filesystem with a container running in the
        // guest. The host file to be shared is described by the
        // Mount argument.
```

I guess this is not clear enough, I extended it to be more explicit.

>     2. Since this returns a `SharedFile`, this means the implementations decide the strategy on what can be shared. Doesn't that imply that the description should be "i.e. it _can_ NOT been shared with the guest" instead of "it _will_ not"?
 
Both. It can not and thus will not. I tried to make that clearer.

src/runtime/virtcontainers/fs_share.go
sameo: > Shouldn't you pass the `SharedFile` result in case an implementation needs it for some cleanup?

`SharedFile` is not a handler on the object that's shared, but it's a description of how `Mount` is going to be shared. So I would not pass it back for now.

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -0,0 +1,394 @@
...
+
+	f.Lock()
+	defer f.Unlock()
+
+	// Prepare is idempotent, i.e. can be called multiple times in a row, without failing
+	// and without modifying the filesystem state after the first call.
+	if f.prepared {
+		return nil
+	} else {
+		f.prepared = true
```
jodh-intel, line 139: Shouldn't this be set right at the end of the function strictly (maybe with a `defer`)? Currently, it's set early but there are lots of error scenarios below, which will return the error but still keep `f.prepared == true`.

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+func (f *FilesystemShare) Cleanup(ctx context.Context) error {
+	f.Lock()
+	defer f.Unlock()
+
+	// Cleanup is idempotent, i.e. can be called multiple times in a row, without failing
+	// and without modifying the filesystem state after the first call.
+	if !f.prepared {
+		return nil
```
jodh-intel, line 181: How about a debug log call here so the admin can see how many times this gets called "erroneously"?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -0,0 +1,394 @@
...
+func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount) (*SharedFile, error) {
+	randBytes, err := utils.GenerateRandomBytes(8)
+	if err != nil {
+		return nil, err
+	}
+
+	filename := fmt.Sprintf("%s-%s-%s", c.id, hex.EncodeToString(randBytes), filepath.Base(m.Destination))
+	guestPath := filepath.Join(kataGuestSharedDir(), filename)
+
+	// copy file to contaier's rootfs if filesystem sharing is not supported, otherwise
```
jodh-intel, line 222:
```suggestion
	// copy file to container's rootfs if filesystem sharing is not supported, otherwise
```

jodh-intel (commented):
Thanks @sameo. A few comments...

src/runtime/virtcontainers/fs_share.go
```diff
@@ -0,0 +1,73 @@
...
+type FilesystemSharer interface {
+	// Prepare will set the host filesystem up, making it ready
+	// to share container files and directories with the guest.
+	// For example, the Linux implementation would create and
+	// prepare the host shared folders, and also make all
+	// sandbox mounts ready to be shared.
+	//
+	// Implementation of this method must be idempotent and be
+	// ready to potentially be called several times in a row,
+	// without symetric calls to Cleanup in between.
```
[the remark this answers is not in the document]
sameo: Added, thanks.

src/runtime/virtcontainers/fs_share_linux.go
sameo: Added.

src/runtime/virtcontainers/fs_share_linux.go
sameo: Fixed, thanks.

egernst:
Looks like firecracker is failing to launch a pod. Perhaps an issue with block device rootfs handling?

sameo:
> Looks like firecracker is failing to launch a pod. Perhaps an issue with block device rootfs handling?

Indeed. It should be fixed now, thanks for the debug help @egernst

sameo:
All green! @egernst @jodh-intel @fidencio would you mind taking a final look?

bergwolf (approved):
lgtm!

egernst (approved):

Issue #3627, opened 2022-02-07
upgrade qemu version to 6.1.0 for snap build on arm


jongwu:
qemu version for snap build on arm is 5.1 and should keep pace with common build.

PR #3628, opened 2022-02-07
snap: update qemu version to 6.1.0 for arm

jongwu:
Update qemu version of snap for arm to 6.1.0 thus the arch specific qemu
version for arm needs clean up.

Fixes: #3627


fidencio (approved):
@jongwu, nice!
lgtm, thanks!

devimc (approved):

Issue #3629, opened 2022-02-07
Add lack annotations for OCI state 


liubin:
Thanks for taking a look @liubin 

>  I will try to add this hook and check that whether it can resolve this issue.

In addition to `createRuntime`, kata seems to lack support for propagating OCI annotations to the State object: https://github.com/opencontainers/runtime-spec/blob/v1.0.2/runtime.md#state

_Originally posted by @AkihiroSuda in https://github.com/kata-containers/kata-containers/issues/1935#issuecomment-1031221039_

PR #3630, opened 2022-02-07
[PoC]hooks: support nerdctl to handle CNI network

liubin:
Nerdctl uses internal hooks(added by `nerdctl run` to OCI spec and passed to lower runtimes ) to create CNI network, this commit will:

- add more properties passing to hooks that OCI spec required.
- add createRuntime hook support, preStart hook is deprecated.
- run hooks in network namespace only if the network namespace is created
outside, that is, run hooks in host network namespace if the container
namespace is created by Kata Containers.

Fixes: #1935, #3629

### DIFF
--- a/src/runtime/pkg/katautils/create.go
+++ b/src/runtime/pkg/katautils/create.go
@@ -155,10 +155,26 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
 		}
 	}()
 
-	// Run pre-start OCI hooks.
-	err = EnterNetNS(sandboxConfig.NetworkConfig.NetNSPath, func() error {
+	// FIXME workaround
+	if sandboxConfig.NetworkConfig.NetNsCreated {
+		annotations := ociSpec.Annotations
+		annotations["kata-netns-path"] = sandboxConfig.NetworkConfig.NetNSPath
+	}
+
+	hooksFunc := func() error {
+		if err := CreateRuntimeHooks(ctx, ociSpec, containerID, bundlePath); err != nil {
+			return err
+		}
 		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
-	})
+	}
+
+	// Run pre-start/createRuntime OCI hooks.
+	// if sandboxConfig.NetworkConfig.NetNsCreated {
+	// network namespace created by kata, hooks run in host namespace.
+	err = hooksFunc()
+	// } else {
+	// err = EnterNetNS(sandboxConfig.NetworkConfig.NetNSPath, hooksFunc)
+	// }
 	if err != nil {
 		return nil, vc.Process{}, err
 	}
@@ -247,10 +263,22 @@ func CreateContainer(ctx context.Context, sandbox vc.VCSandbox, ociSpec specs.Sp
 		return vc.Process{}, err
 	}
 
-	// Run pre-start OCI hooks.
-	err = EnterNetNS(sandbox.GetNetNs(), func() error {
+	hooksFunc := func() error {
+		if err := CreateRuntimeHooks(ctx, ociSpec, containerID, bundlePath); err != nil {
+			return err
+		}
 		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
-	})
+	}
+
+	netNs := sandbox.GetNetworkNamespace()
+	// Run pre-start/createRuntime OCI hooks.
+	if netNs.NetNsCreated {
+		// network namespace created by kata, hooks run in host namespace.
+		err = hooksFunc()
+	} else {
+		err = EnterNetNS(netNs.NetNsPath, hooksFunc)
+	}
+
 	if err != nil {
 		return vc.Process{}, err
 	}
--- a/src/runtime/pkg/katautils/create.go
+++ b/src/runtime/pkg/katautils/create.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 
@@ -155,27 +156,43 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
 		}
 	}()
 
-	// FIXME workaround
-	if sandboxConfig.NetworkConfig.NetNsCreated {
-		annotations := ociSpec.Annotations
-		annotations["kata-netns-path"] = sandboxConfig.NetworkConfig.NetNSPath
+	// this pid will bound to the network namespace.
+	var pid int = -1
+
+	var cmd *exec.Cmd
+	tmpProcess := func() error {
+		cmd = exec.Command("sleep", "3600")
+		if err := cmd.Start(); err != nil {
+			kataUtilsLogger.WithError(err).Error("failed to start temp process for running hooks")
+			return err
+		}
+		pid = cmd.Process.Pid
+		return nil
 	}
 
-	hooksFunc := func() error {
-		if err := CreateRuntimeHooks(ctx, ociSpec, containerID, bundlePath); err != nil {
+	defer func() {
+		if err := cmd.Process.Kill(); err != nil {
+			kataUtilsLogger.WithError(err).Error("failed to kill temp process")
+		}
+		if err := cmd.Wait(); err != nil {
+			kataUtilsLogger.WithError(err).Error("failed to kill temp process")
+		}
+	}()
+
+	// run tmp process in the network namespace and pass the pid of this
+	// process to hooks to let hooks get the correct network namespace
+	if err = EnterNetNS(sandboxConfig.NetworkConfig.NetNSPath, tmpProcess); err != nil {
+		kataUtilsLogger.WithError(err).Error("failed to run temp process in network namespace")
+		return nil, vc.Process{}, err
+	}
+
+	if err = func() error {
+		if err := CreateRuntimeHooks(ctx, ociSpec, containerID, bundlePath, pid); err != nil {
 			return err
 		}
 		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
-	}
-
-	// Run pre-start/createRuntime OCI hooks.
-	// if sandboxConfig.NetworkConfig.NetNsCreated {
-	// network namespace created by kata, hooks run in host namespace.
-	err = hooksFunc()
-	// } else {
-	// err = EnterNetNS(sandboxConfig.NetworkConfig.NetNSPath, hooksFunc)
-	// }
-	if err != nil {
+	}(); err != nil {
+		kataUtilsLogger.WithError(err).Error("failed to run hooks")
 		return nil, vc.Process{}, err
 	}
 
@@ -264,9 +281,6 @@ func CreateContainer(ctx context.Context, sandbox vc.VCSandbox, ociSpec specs.Sp
 	}
 
 	hooksFunc := func() error {
-		if err := CreateRuntimeHooks(ctx, ociSpec, containerID, bundlePath); err != nil {
-			return err
-		}
 		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
 	}
 
--- a/src/runtime/pkg/containerd-shim-v2/start.go
+++ b/src/runtime/pkg/containerd-shim-v2/start.go
@@ -53,10 +53,20 @@ func startContainer(ctx context.Context, s *service, c *container) (retErr error
 		}
 	}
 
-	// Run post-start OCI hooks.
-	err := katautils.EnterNetNS(s.sandbox.GetNetNs(), func() error {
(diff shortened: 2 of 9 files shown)


liubin:
This PR has a **big assumption**: hooks will run in the host network namespace if the container network namespace is created by Kata Containers, that is, hooks will run in the container network namespace if the namespace is created outside and passed to Kata Containers.

/cc @lifupan @bergwolf @egernst @jodh-intel @fidencio @AkihiroSuda

AkihiroSuda:
Thanks, but this doesn't seem to work for me

First run:
```console
# nerdctl run -it --rm --runtime=io.containerd.kata.v2 alpine
/ # ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
```

Second run:
```console
# nerdctl run -it --rm --runtime=io.containerd.kata.v2 alpine
FATA[0000] failed to create shim task: exit status 1: stdout: , stderr: time="2022-02-08T17:13:29+09:00" level=fatal msg="failed to call cni.Setup: plugin type=\"bridge\" failed (add): container veth name provided (eth0) already exists"
Failed to write to log, write /var/lib/nerdctl/1935db59/containers/default/6542351436f68d9e278a71f9edc55b3b48eb2a17c3bfc8b0f8c290eb3d5197b8/oci-hook.createRuntime.log: file already closed
: unknown 
```

Version:
```
# /opt/kata/bin/kata-runtime -v
kata-runtime  : 2.4.0-alpha2
   commit   : 3c083a834a8eadaa00da7a6483cc5f905dc11080
   OCI specs: 1.0.2-dev

# nerdctl version
Client:
 Version:       v0.16.1
 Git commit:    c4bd56b3aa220db037cc6c0a4e0c8cc062f2cc4c

Server:
 containerd:
  Version:      v1.6.0-rc.1
  GitCommit:    ad771115b82a70cfd8018d72ae489c707e63de16
```

liubin:
@AkihiroSuda Thank you for the test, I found the root cause may be that nerdctl OCI hook can't get the correct namespace the kata created.

I updated the code and you can check it out from https://github.com/liubin/kata-containers/tree/fix/1935-add-createRuntime-hook, the main change is to pass network namespace to OCI hooks by annotation.

And changes to nerdctl are also needed(`pkg/ocihook/ocihook.go`):

```diff
 func getNetNSPath(state *specs.State) (string, error) {
+       if state.Annotations != nil {
+               if v, ok := state.Annotations["kata-netns-path"]; ok && v != nil {
+                       return v, nil
+               }
+       }
+
        if state.Pid == 0 {
                return "", errors.New("state.Pid is unset")
        }
```

This is my test:

```bash
# nerdctl run --runtime='io.containerd.kata.v2' -it --network=mynet containerstack/alpine-stress:latest /bin/bash
bash-4.3# ip addr show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP qlen 1000
    link/ether 9a:29:0c:28:d5:b1 brd ff:ff:ff:ff:ff:ff
    inet 172.19.0.60/24 brd 172.19.0.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::9829:cff:fe28:d5b1/64 scope link 
       valid_lft forever preferred_lft forever


# nerdctl run --runtime='io.containerd.kata.v2' -it  containerstack/alpine-stress:latest /bin/bash
bash-4.3# ip addr show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP qlen 1000
    link/ether 06:61:8d:53:c1:49 brd ff:ff:ff:ff:ff:ff
    inet 10.4.0.35/24 brd 10.4.0.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::461:8dff:fe53:c149/64 scope link 
       valid_lft forever preferred_lft forever
```

AkihiroSuda:
Thanks! 

>  state.Annotations["kata-netns-path"];

Do we need this annotation?
Can we just use `.linux.namespaces` object of OCI `config.json`?
https://github.com/opencontainers/runtime-spec/blob/v1.0.2/config-linux.md#namespaces

liubin:
> Thanks!
> 
> > state.Annotations["kata-netns-path"];
> 
> Do we need this annotation? Can we just use `.linux.namespaces` object of OCI `config.json`? https://github.com/opencontainers/runtime-spec/blob/v1.0.2/config-linux.md#namespaces

No, this(using an annotation) is not the final solution.

And `.linux.namespaces`  contains a network namespace, but doesn't specify a `path`, thus Kata Containers creates the network path.

The ideal solution is passing a Pid(as runc did) to hooks and letting hooks get the correct namespace path.

liubin:
I added a new https://github.com/kata-containers/kata-containers/pull/3630/commits/6398ec3a58bccde63d9485eb9157a61b1398c9ae , this commit doesn't depend on annotations to pass network namespace, instead, it creates a new temporality and bind the network namespace created by kata to this process, and pass this process' PID to hooks that let the hooks get the correct network namespace.

src/runtime/pkg/katautils/create.go
```diff
@@ -155,10 +155,26 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
...
+			return err
+		}
 		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
-	})
+	}
+
+	// Run pre-start/createRuntime OCI hooks.
+	// if sandboxConfig.NetworkConfig.NetNsCreated {
+	// network namespace created by kata, hooks run in host namespace.
+	err = hooksFunc()
```
sameo: That is correct, imo.

src/runtime/pkg/katautils/create.go
```diff
@@ -247,10 +263,22 @@ func CreateContainer(ctx context.Context, sandbox vc.VCSandbox, ociSpec specs.Sp
...
+	}
+
+	netNs := sandbox.GetNetworkNamespace()
+	// Run pre-start/createRuntime OCI hooks.
+	if netNs.NetNsCreated {
+		// network namespace created by kata, hooks run in host namespace.
+		err = hooksFunc()
+	} else {
+		err = EnterNetNS(netNs.NetNsPath, hooksFunc)
+	}
```
sameo, line 294: But that is not correct. Both `createRuntime` and the deprecated `preStart` should always run in the runtime netns. Never in the sandbox one.

src/runtime/pkg/katautils/create.go
```diff
@@ -155,27 +156,43 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
 		}
 	}()
 
-	// FIXME workaround
-	if sandboxConfig.NetworkConfig.NetNsCreated {
-		annotations := ociSpec.Annotations
-		annotations["kata-netns-path"] = sandboxConfig.NetworkConfig.NetNSPath
-	}
+	// this pid will bound to the network namespace.
+	var pid int = -1
```
sameo, line 160: Could you please elaborate as to why we need that?

sameo:
> I added a new [6398ec3](https://github.com/kata-containers/kata-containers/commit/6398ec3a58bccde63d9485eb9157a61b1398c9ae) , this commit doesn't depend on annotations to pass network namespace, instead, it creates a new temporality and bind the network namespace created by kata to this process, and pass this process' PID to hooks that let the hooks get the correct network namespace.

Do you do that because the hook resolves the netns path to use from `/proc/<pid>` where `pid` is coming from the `spec.State`?

sameo:
> > I added a new [6398ec3](https://github.com/kata-containers/kata-containers/commit/6398ec3a58bccde63d9485eb9157a61b1398c9ae) , this commit doesn't depend on annotations to pass network namespace, instead, it creates a new temporality and bind the network namespace created by kata to this process, and pass this process' PID to hooks that let the hooks get the correct network namespace.
> 
> Do you do that because the hook resolves the netns path to use from `/proc/<pid>` where `pid` is coming from the `spec.State`?

I guess it is:
https://github.com/containerd/nerdctl/blob/master/pkg/ocihook/ocihook.go#L240

sameo:
> > Thanks!
> > > state.Annotations["kata-netns-path"];
> > 
> > 
> > Do we need this annotation? Can we just use `.linux.namespaces` object of OCI `config.json`? https://github.com/opencontainers/runtime-spec/blob/v1.0.2/config-linux.md#namespaces
> 
> No, this(using an annotation) is not the final solution.
> 
> And `.linux.namespaces` contains a network namespace, but doesn't specify a `path`, thus Kata Containers creates the network path.
> 
> The ideal solution is passing a Pid(as runc did) to hooks and letting hooks get the correct namespace path.

Creating a temporary, idle process is a little hackish imho. Either using an annotation and using that from nerdctl, or adding the created namespace path to the bundle `config.json` would be cleaner.

liubin:
> > > I added a new [6398ec3](https://github.com/kata-containers/kata-containers/commit/6398ec3a58bccde63d9485eb9157a61b1398c9ae) , this commit doesn't depend on annotations to pass network namespace, instead, it creates a new temporality and bind the network namespace created by kata to this process, and pass this process' PID to hooks that let the hooks get the correct network namespace.
> > 
> > 
> > Do you do that because the hook resolves the netns path to use from `/proc/<pid>` where `pid` is coming from the `spec.State`?
> 
> I guess it is: https://github.com/containerd/nerdctl/blob/master/pkg/ocihook/ocihook.go#L240

Yes.

In this PR I tried two implementations.

First commit uses annotation pass the netns to nerdctl, if nerdctl can get the netns from annotation, it will give up get netns from `/proc/pid`.

The second commit runs a temp process in the created netns and pass its' pid to nerdctl hook and then nerdctl can get the correct netns from `/proc/pid`.

liubin:
> Creating a temporary, idle process is a little hackish imho. Either using an annotation and using that from nerdctl, or adding the created namespace path to the bundle `config.json` would be cleaner.

From [runtime spec](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/config.md#posix-platform-hooks):

> The [state](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/runtime.md#state) of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.

Hooks can get input only from the container's state object, reading `config.json` is not required by the spec.

sameo:
> > Creating a temporary, idle process is a little hackish imho. Either using an annotation and using that from nerdctl, or adding the created namespace path to the bundle `config.json` would be cleaner.
> 
> From [runtime spec](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/config.md#posix-platform-hooks):
> 
> > The [state](https://github.com/opencontainers/runtime-spec/blob/v1.0.2/runtime.md#state) of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.
> 
> Hooks can get input only from the container's state object, reading `config.json` is not required by the spec.

I understand that, but I see that @AkihiroSuda added some workaround for reading `config.json` back:

https://github.com/containerd/nerdctl/pull/788/files#diff-e0596cfbff08dfa3c5d049b51af0bf64607ac3fcd0fd48087b17be54c6632ef6R55

Honestly, I dont think this is the cleanest approach and I believe we should define a label in `nerdctl` [labels package](https://github.com/containerd/nerdctl/blob/master/pkg/labels/labels.go) and use that to pass our Kata defined netns through a `nerdctl` namespaced annotation.

Ideally, `nerdctl` would create the netns for us (like the CRI flow), but it seems that would be quite an intrusive change?

sameo:
@liubin Please see #3670. I think it's a simpler solution, provided the [nerdctl PR](https://github.com/containerd/nerdctl/pull/817) gets accepted.

With #3670 and the nerdctl PR, I have networking working as expected with Kata.

liubin:
Close due to #3670

Issue #3649, opened 2022-02-11
agent test test_execute_hook_with_error sometime fails


liubin:
Some static-checks CI fails with this error.

The hook command is `ls`, maybe the hook process has exited, and then try to write stdin of it.

> running 57 tests
test cgroups::fs::tests::test_lines_to_map ... ok
test cgroups::fs::tests::test_line_to_vec ... ok
test container::tests::test_execute_hook ... ok
thread 'container::tests::test_execute_hook_with_error' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }', rustjail/src/container.rs:1532:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::result::unwrap_failed
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/result.rs:1613:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/result.rs:1295:23
   4: rustjail::container::execute_hook::{{closure}}::{{closure}}
             at ./src/container.rs:1526:9
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/future/mod.rs:80:19
   6: tokio::runtime::task::core::CoreStage<T>::poll::{{closure}}
             at /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.15.0/src/runtime/task/core.rs:161:17
   7: tokio::loom::std::unsafe_cell::UnsafeCell<T>::with_mut
             at /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.15.0/src/loom/std/unsafe_cell.rs:14:9
   8: tokio::runtime::task::core::CoreStage<T>::poll
             at /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.15.0/src/runtime/task/core.rs:151:13
[...]

PR #3652, opened 2022-02-12
agent: use sleep as hooks test to prevent hooks quit too fast

liubin:
When running agent hooks, tokio will spawn a new process
and write the container's state to hooks through stdin.
Tests case for hooks use ls and touch commands, that
will not block on reading stdin and maybe exit immediately
after the process started, this may lead to test cases
sometimes failing due to broken pipe error.

Changing the test command of hooks to `sleep 1` ensures
that there will be enough time to write state to
stdin of the hook processes and avoid broken pipe errors.

Fixes: #3649

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -1642,17 +1642,13 @@ mod tests {
 
     #[tokio::test]
     async fn test_execute_hook() {
-        let temp_file = "/tmp/test_execute_hook";
+        let sh = which("sh").await;
 
-        let touch = which("touch").await;
-
-        defer!(fs::remove_file(temp_file).unwrap(););
-
-        execute_hook(
+        let res = execute_hook(
             &slog_scope::logger(),
             &Hook {
-                path: touch,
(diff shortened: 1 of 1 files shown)


src/agent/rustjail/src/container.rs
```diff
@@ -1665,21 +1661,24 @@ mod tests {
...
         let res = execute_hook(
             &slog_scope::logger(),
             &Hook {
-                path: ls,
-                args: vec!["ls".to_string(), "/tmp/not-exist".to_string()],
+                path: sh,
+                args: vec![
+                    "sh".to_string(),
+                    "-c".to_string(),
+                    "sleep 1 && exit 1".to_string(),
```
jodh-intel, line 1680: Is there any other way to deal with this? Sleeps are fragile and this is bound to fail on some (slow) h/w one day.

liubin: @jodh-intel  I'm going to close this one and fix this in https://github.com/kata-containers/kata-containers/pull/3668

PR #3670, opened 2022-02-15
Support nerdctl OCI hooks

sameo:
The kata runtime does not properly work with `nerdctl` [1], and this is partly due to the fact that it does not support the `createRuntime` hooks and also because it can not tell `nerdctl` which networking namespace to use when calling the CNI plugins.

This PR:

1. Add support for the `createRuntime` hooks
2. Run both the (deprecated) `preStart` and `createRuntime` hooks in the runtime networking namespace
3. Pass the spec annotations back to the called hooks
4. Add a `nerdctl` specific annotation for indicating which networking namespace to use when calling the CNI plugins.

For a fully working `nerdctl`, this PR depends on https://github.com/containerd/nerdctl/pull/817

[1] https://github.com/containerd/nerdctl/issues/787

### DIFF
--- a/src/runtime/pkg/katautils/create.go
+++ b/src/runtime/pkg/katautils/create.go
@@ -160,6 +160,11 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
 		return nil, vc.Process{}, err
 	}
 
+	// Run create runtime OCI hooks, in the runtime namespace.
+	if err := CreateRuntimeHooks(ctx, ociSpec, containerID, bundlePath); err != nil {
+		return nil, vc.Process{}, err
+	}
+
 	sandbox, err := vci.CreateSandbox(ctx, sandboxConfig)
 	if err != nil {
 		return nil, vc.Process{}, err
--- a/src/runtime/pkg/katautils/create.go
+++ b/src/runtime/pkg/katautils/create.go
@@ -155,11 +155,8 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
 		}
 	}()
 
-	// Run pre-start OCI hooks.
-	err = EnterNetNS(sandboxConfig.NetworkConfig.NetworkID, func() error {
-		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
-	})
-	if err != nil {
+	// Run pre-start OCI hooks, in the runtime namespace.
+	if err := PreStartHooks(ctx, ociSpec, containerID, bundlePath); err != nil {
 		return nil, vc.Process{}, err
 	}
 
--- a/src/runtime/pkg/katautils/create.go
+++ b/src/runtime/pkg/katautils/create.go
@@ -155,6 +155,12 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
 		}
 	}()
 
+	if ociSpec.Annotations == nil {
+		ociSpec.Annotations = make(map[string]string)
+	}
+	ociSpec.Annotations["nerdctl/network-namespace"] = sandboxConfig.NetworkConfig.NetworkID
+	sandboxConfig.Annotations["nerdctl/network-namespace"] = ociSpec.Annotations["nerdctl/network-namespace"]
+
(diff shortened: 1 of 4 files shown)


src/runtime/pkg/katautils/create.go
```diff
@@ -155,11 +155,18 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec specs.Spec, runtimeCo
...
 
-	// Run pre-start OCI hooks.
-	err = EnterNetNS(sandboxConfig.NetworkConfig.NetworkID, func() error {
-		return PreStartHooks(ctx, ociSpec, containerID, bundlePath)
-	})
-	if err != nil {
+	if ociSpec.Annotations == nil {
+		ociSpec.Annotations = make(map[string]string)
+	}
+	ociSpec.Annotations["nerdctl/network-namespace"] = sandboxConfig.NetworkConfig.NetworkID
```
jodh-intel, line 161: @bergwolf, @liubin, @jiangliu - ooi, do you have thoughts on how we're going to manage adding new features like this into the `runtime-rs` version?

jodh-intel (approved):
Thanks @sameo.

Any chance you could add a new unit test to check the annotation?

lgtm

src/runtime/pkg/katautils/create.go
jiangliu: My first thought would be  "rely on the CI system":)
It would be challenging to keep sync with 2.x branch, so we plan to have runnable 3.0 branch and let the CI system to    disclose missed features.

jodh-intel: Agreed. The CI is going to be crucially important, but relying on it assumes that _every_ new feature has some sort of test with it, so we're all going to have to be extremely careful to ensure that all PRs do have tests either unit tests or an integration test in the `tests` repo.

sameo:
> Any chance you could add a new unit test to check the annotation?

Good suggestion, I just added one now.

jodh-intel (approved):
Thanks @sameo.

lgtm

liubin:
I have only one question, run hooks in host ns may break Docker usage, though Docker can't use Kata Containers now.
Docker's hook run in new netns, so it can get netns path from `/proc/pid/ns/net` and pass it to `dockerd` through `/var/run/docker/libnetwork/xxx.sock`. The real network operations may be run in `dockerd` daemon. That is the difference between Docker and nerdctl.

sameo:
> I have only one question, run hooks in host ns may break Docker usage, though Docker can't use Kata Containers now. Docker's hook run in new netns, so it can get netns path from `/proc/pid/ns/net` and pass it to `dockerd` through `/var/run/docker/libnetwork/xxx.sock`. The real network operations may be run in `dockerd` daemon. That is the difference between Docker and nerdctl.

Hey @liubin . Docker expects *all* OCI hooks to run in the container netns?

liubin:
> > I have only one question, run hooks in host ns may break Docker usage, though Docker can't use Kata Containers now. Docker's hook run in new netns, so it can get netns path from `/proc/pid/ns/net` and pass it to `dockerd` through `/var/run/docker/libnetwork/xxx.sock`. The real network operations may be run in `dockerd` daemon. That is the difference between Docker and nerdctl.
> 
> Hey @liubin . Docker expects _all_ OCI hooks to run in the container netns?

I'm not very sure about that, I'm only curious why hooks run in container netns for a long time, if Docker worked before, I's don't sure it will still work after this PR.

jodh-intel:
The OCI spec (now) provides some detail about which ns the hooks are supposed to run in:

- https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks

egernst:
@liubin @sameo  -- WDYT for next steps here?

sameo:
Both createRuntime and preStart `MUST be executed in the [runtime namespace](https://github.com/opencontainers/runtime-spec/blob/main/glossary.md#runtime-namespace).` So I think this PR is actually fixing a bug.

egernst (approved):

egernst:
Agreed. LGTM. James' approval was reset when you updated -- can someone TAL and ACK?

fidencio (approved):
lgtm, thanks @sameo!

Issue #3672, opened 2022-02-15
packaging: Allow building cloud-hypervisor from source, even when there's a released binary


fidencio:
A released binary doesn't guarantee it's been built with all the needed features for this or that feature worked on from the kata-containers side.

So, as every now and then we'll need a binary from a specific release, but built with different flags, we should relax https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh to allow that case.

Issue #3671, opened 2022-02-15
packaging: Add support for passing down a flag to cargo when building cloud-hypervisor


fidencio:
https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh is the script used to either gather the released binary or to build cloud-hypervisor from sources.  Ideally, this script should support flags being passed down to the `./scripts/dev_cli.sh` script from the cloud-hypervisor side.

Right new the script is called as (see: https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh):
```
    ./scripts/dev_cli.sh build --release --libc musl
```

But we could ideally do:
```
    ./scripts/dev_cli.sh build --release --libc musl -- --features foo
```

PR #3673, opened 2022-02-15
tools: clh: Allow to set when to build from sources and the build flags passed down to cargo

fidencio:
This short series adds the possibility to force building Cloud Hypervisor from sources as not always the released binary, even if it matches the version we're trying to use / ship, has been built with the needed features enabled.  And talking about building something with the needed features, support for doing this is also added as part of this series.

Fixes: #3672 
Fixes: #3671

### DIFF
--- a/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh
+++ b/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh
@@ -15,6 +15,7 @@ ARCH=$(uname -m)
 
 script_dir=$(dirname $(readlink -f "$0"))
 kata_version="${kata_version:-}"
+force_build_from_source="${force_build_from_source:-false}"
 
 source "${script_dir}/../../scripts/lib.sh"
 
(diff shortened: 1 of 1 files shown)


tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh
```diff
@@ -49,13 +54,32 @@ build_clh_from_source() {
...
-    ./scripts/dev_cli.sh build --release --libc musl
+    if [ -n "${extra_build_args}" ]; then
+        info "Build cloud-hypervisor with extra args: ${extra_build_args}"
+        ./scripts/dev_cli.sh build --release --libc musl -- ${extra_build_args}
+    else
+        ./scripts/dev_cli.sh build --release --libc musl
+    fi
     rm -f cloud-hypervisor
-    cp build/cargo_target/$(uname -m)-unknown-linux-musl/release/cloud-hypervisor .
+    cp build/cargo_target/$(uname -m)-unknown-linux-musl/release/cloud-hypervisor "${binary_name}"
```
jodh-intel, line 64: Nit: I can see this is the minimal fix but I do wonder if it would be better to say the following (as it avoids hard-coding libc and OS names and ensures that if cargo changes "under the covers" the script won't break):

```bash
cargo install --path .
```

The message can then tell the user to find the binary at `~/.cargo/bin/${binary_name}`.

jodh-intel (dismissed):
Thanks @fidencio.

tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh
fidencio: @jodh-intel, actually, I'll end up dropping this commit from the PR as, after a conversation with Julio, I think we don't need to set a different binary name (as we may not introduce `clh-experimental`).

Your point is still valid, but I'd rather leave it for another series than touch not exactly related pieces as part of this PR.

devimc (approved):

amshinde (approved):

Issue #3674, opened 2022-02-15
kata-manager script fails to install Kata or containerd


jodh-intel:
The [`kata-manager.sh`](https://github.com/kata-containers/kata-containers/blob/main/utils/README.md) script is currently broken:

```bash
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh)"
ERROR: File '/tmp/tmp.ledbdnAF1w/kata-static-2.3.2-x86_64.tar.xz' contains unexpected paths: './'
```

Working around that (manually) results in another error:

```bash
INFO: Running pre-checks
INFO: Checking dependencies
INFO: Downloading Kata Containers release (latest version)
INFO: Installing Kata Containers release 2.3.2 from /tmp/tmp.TB2Awhos6C/kata-static-2.3.2-x86_64.tar.xz
INFO: Kata Containers installed

kata-runtime  : 2.3.2
   commit   : 1af292c9e693e9bc8e8324a9eb860dad45306fb5
   OCI specs: 1.0.2-dev
INFO: Downloading containerd release (latest version)
ERROR: Cannot determine download URL for version v1.5.9 (https://api.github.com/repos/containerd/containerd/releases)
ERROR: Cannot determine download URL for version v1.5.9
ERROR: Cannot determine containerd release file
```

This also shows we need a GitHub action to test https://github.com/kata-containers/kata-containers/blob/main/utils/README.md.

jodh-intel:
That'll have to be a Jenkins job I think as GH actions doesn't support kvm I think.

PR #3675, opened 2022-02-15
Kata manager fix install

jodh-intel:
Fix installation of both Kata and containerd for the `kata-manager.sh` script and update the docs.

Fixes: #3674.

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -560,9 +560,16 @@ handle_installation()
 
 handle_args()
 {
-	case "${1:-}" in
-		-h|--help|help) usage; exit 0;;
-	esac
+	local opt
+
+	while getopts "h" opt "$@"
+	do
+		case "$opt" in
+			h) usage; exit 0 ;;
+		esac
+	done
+
+	shift $[$OPTIND-1]
 
 	local kata_version="${1:-}"
 	local containerd_version="${2:-}"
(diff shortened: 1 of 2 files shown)


utils/kata-manager.sh
```diff
@@ -543,31 +608,72 @@ test_installation()
...
+			o) only_kata="true" ;;
+			r) cleanup="false" ;;
+		esac
+	done
+
+	shift $[$OPTIND-1]
 
-	local kata_version="${1:-}"
-	local containerd_version="${2:-}"
+	[ -z "$kata_version" ] && kata_version="${1:-}"
```
devimc, line 670: Q. can this fail with `set -e` ?  for this kind of checks I think the best  option is `-n` and `||` , isn't it? :thinking:

devimc (approved):
thanks @jodh-intel - just one question

utils/kata-manager.sh
jodh-intel: I think it's correct, but given the shell's obscure rules I've added `|| true` "just in case" ;)

fidencio (approved):
lgtm, with the caveat that I am very much **not** familiar with kata manager.

jodh-intel:
@devimc - ptal.

jodh-intel:
Could we get 1 more ack on this folks? We really need this to land asap.

devimc (approved):
thanks @jodh-intel

Issue #3678, opened 2022-02-15
Agent VFIO handlers assume PCI


Jakob-Naucke:
Besides VFIO-PCI, there is e.g. VFIO-AP. Passing VFIO-AP devices was initially supported in #578, but I had only written agent support for the v1 Go agent at that time. Since #1682, not handling VFIO-AP devices is punished: `Bad host PCI address in VFIO option {:?}` (never mind the missing format).

PR #1190, opened 2020-12-11
Clean up PCI path handling

dgibson:
Consistent and safer handling, forward port of https://github.com/kata-containers/agent/pull/855 and https://github.com/kata-containers/runtime/pull/3003.

PR #578, opened 2020-08-26
runtime: Add support for VFIO-AP pass-through

Jakob-Naucke:
Recognise when a device to be hot-plugged is an IBM Adjunct Processor (AP) device and execute VFIO AP hot-plug accordingly. 
This enables running cryptographic workloads on an IBM CryptoCard inside a Kata container, since these cards use the AP bus on s390x (IBM Z mainframe).
I am momentarily calling this a WIP because I am not entirely sure about documentation.

To use this, one has to make an adapter and domain available for VFIO and assign them to a mediated device as per the [kernel documentation](https://www.kernel.org/doc/html/latest/s390/vfio-ap.html), determine the IOMMU group for this mediated device similar to [Kata's guide on doing this with an Intel i915 GPU](https://github.com/kata-containers/kata-containers/blob/2.0-dev/docs/use-cases/Intel-GPU-passthrough-and-Kata.md), and run something like

```sh
$ ctr run --runtime io.containerd.run.kata.v2 -t --rm --device /dev/vfio/0 docker.io/library/ubuntu:latest foo bash
$ lszcrypt # requires s390-tools to be installed
CARD.DOMAIN TYPE  MODE        STATUS  REQUESTS
----------------------------------------------
09          CEX6C CCA-Coproc  online         1
09.003d     CEX6C CCA-Coproc  online         1
```

Should this be a full guide when it's so similar to the iGPU method? WDYT?

/cc @alicefr

Fixes: #491

PR #3679, opened 2022-02-15
agent: Add VFIO-AP device handling

Jakob-Naucke:
Initial VFIO-AP support (#578) was simple, but somewhat hacky; a
different code path would be chosen for performing the hotplug, and
agent-side device handling was bound to knowing the assigned queue
numbers (APQNs) through some other means; plus the code for awaiting
them was written for the Go agent and never released. This code also
artificially increased the hotplug timeout to wait for the (relatively
expensive, thus limited to 5 seconds at the quickest) AP rescan, which
is impractical for e.g. common k8s timeouts.

Since then, the general handling logic was improved (#1190), but it
assumed PCI in several places.

In the runtime, introduce and parse AP devices. Annotate them as such
when passing to the agent, and information about the enabled APQNs is
included.

The agent awaits the passed APQNs through uevents and triggers a
rescan directly.

Fixes: #3678

### DIFF
--- a/src/agent/src/ap.rs
+++ b/src/agent/src/ap.rs
@@ -1,0 +1,79 @@
+// Copyright (c) IBM Corp. 2022
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use std::fmt;
+use std::str::FromStr;
+
+use anyhow::{anyhow, Context};
+
+// IBM Adjunct Processor (AP) is the bus used by IBM Crypto Express hardware security modules on
+// IBM Z & LinuxONE (s390x)
+// AP bus ID follow the format <xx>.<xxxx> [1, p. 476], where
+//   - <xx> is the adapter ID, i.e. the card and
+//   - <xxxx> is the adapter domain.
+// [1] https://www.ibm.com/docs/en/linuxonibm/pdf/lku5dd05.pdf
+
+#[derive(Debug)]
+pub struct Address {
+    pub adapter_id: u8,
+    pub adapter_domain: u16,
+}
+
+impl Address {
+    pub fn new(adapter_id: u8, adapter_domain: u16) -> Address {
+        Address {
+            adapter_id,
+            adapter_domain,
+        }
+    }
+}
+
+impl FromStr for Address {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        let split: Vec<&str> = s.split('.').collect();
+        if split.len() != 2 {
+            return Err(anyhow!(
+                "Wrong AP bus format. It needs to be in the form <xx>.<xxxx>, got {:?}",
+                s
+            ));
+        }
+
+        let adapter_id = u8::from_str_radix(split[0], 16).context(format!(
+            "Wrong AP bus format. AP ID needs to be in the form <xx>, got {:?}",
+            split[0]
+        ))?;
+        let adapter_domain = u16::from_str_radix(split[1], 16).context(format!(
+            "Wrong AP bus format. AP domain needs to be in the form <xxxx>, got {:?}",
+            split[1]
+        ))?;
+
+        Ok(Address::new(adapter_id, adapter_domain))
+    }
+}
+
+impl fmt::Display for Address {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f, "{:02x}.{:04x}", self.adapter_id, self.adapter_domain)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_from_str() {
+        let device = Address::from_str("a.1").unwrap();
+        assert_eq!(format!("{}", device), "0a.0001");
+
+        assert!(Address::from_str("").is_err());
+        assert!(Address::from_str(".").is_err());
+        assert!(Address::from_str("0.0.0").is_err());
+        assert!(Address::from_str("0g.0000").is_err());
+        assert!(Address::from_str("0a.10000").is_err());
+    }
+}
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -45,11 +45,11 @@ pub const DRIVER_NVDIMM_TYPE: &str = "nvdimm";
 pub const DRIVER_EPHEMERAL_TYPE: &str = "ephemeral";
 pub const DRIVER_LOCAL_TYPE: &str = "local";
 pub const DRIVER_WATCHABLE_BIND_TYPE: &str = "watchable-bind";
-// VFIO device to be bound to a guest kernel driver
-pub const DRIVER_VFIO_GK_TYPE: &str = "vfio-gk";
-// VFIO device to be bound to vfio-pci and made available inside the
+// VFIO PCI device to be bound to a guest kernel driver
+pub const DRIVER_VFIO_PCI_GK_TYPE: &str = "vfio-pci-gk";
+// VFIO PCI device to be bound to vfio-pci and made available inside the
 // container as a VFIO device node
-pub const DRIVER_VFIO_TYPE: &str = "vfio";
+pub const DRIVER_VFIO_PCI_TYPE: &str = "vfio-pci";
 pub const DRIVER_OVERLAYFS_TYPE: &str = "overlayfs";
 pub const FS_TYPE_HUGETLB: &str = "hugetlbfs";
 
@@ -704,7 +704,7 @@ async fn virtio_nvdimm_device_handler(
     Ok(DevNumUpdate::from_vm_path(&device.vm_path)?.into())
 }
 
-fn split_vfio_option(opt: &str) -> Option<(&str, &str)> {
+fn split_vfio_pci_option(opt: &str) -> Option<(&str, &str)> {
     let mut tokens = opt.split('=');
     let hostbdf = tokens.next()?;
     let path = tokens.next()?;
@@ -719,14 +719,18 @@ fn split_vfio_option(opt: &str) -> Option<(&str, &str)> {
 // Each option should have the form "DDDD:BB:DD.F=<pcipath>"
 //     DDDD:BB:DD.F is the device's PCI address in the host
 //     <pcipath> is a PCI path to the device in the guest (see pci.rs)
-async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) -> Result<SpecUpdate> {
-    let vfio_in_guest = device.field_type != DRIVER_VFIO_GK_TYPE;
+#[instrument]
+async fn vfio_pci_device_handler(
+    device: &Device,
+    sandbox: &Arc<Mutex<Sandbox>>,
+) -> Result<SpecUpdate> {
+    let vfio_in_guest = device.field_type != DRIVER_VFIO_PCI_GK_TYPE;
     let mut pci_fixups = Vec::<(pci::Address, pci::Address)>::new();
     let mut group = None;
 
     for opt in device.options.iter() {
         let (host, pcipath) =
-            split_vfio_option(opt).ok_or_else(|| anyhow!("Malformed VFIO option {:?}", opt))?;
+            split_vfio_pci_option(opt).ok_or_else(|| anyhow!("Malformed VFIO option {:?}", opt))?;
         let host =
             pci::Address::from_str(host).context("Bad host PCI address in VFIO option {:?}")?;
         let pcipath = pci::Path::from_str(pcipath)?;
@@ -833,7 +837,9 @@ async fn add_device(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) -> Result<Sp
         DRIVER_MMIO_BLK_TYPE => virtiommio_blk_device_handler(device, sandbox).await,
         DRIVER_NVDIMM_TYPE => virtio_nvdimm_device_handler(device, sandbox).await,
         DRIVER_SCSI_TYPE => virtio_scsi_device_handler(device, sandbox).await,
-        DRIVER_VFIO_GK_TYPE | DRIVER_VFIO_TYPE => vfio_device_handler(device, sandbox).await,
+        DRIVER_VFIO_PCI_GK_TYPE | DRIVER_VFIO_PCI_TYPE => {
+            vfio_pci_device_handler(device, sandbox).await
+        }
         _ => Err(anyhow!("Unknown device type {}", device.field_type)),
     }
 }
@@ -1492,13 +1498,13 @@ mod tests {
     }
 
     #[test]
-    fn test_split_vfio_option() {
+    fn test_split_vfio_pci_option() {
         assert_eq!(
-            split_vfio_option("0000:01:00.0=02/01"),
+            split_vfio_pci_option("0000:01:00.0=02/01"),
             Some(("0000:01:00.0", "02/01"))
         );
-        assert_eq!(split_vfio_option("0000:01:00.0=02/01=rubbish"), None);
-        assert_eq!(split_vfio_option("0000:01:00.0"), None);
+        assert_eq!(split_vfio_pci_option("0000:01:00.0=02/01=rubbish"), None);
+        assert_eq!(split_vfio_pci_option("0000:01:00.0"), None);
     }
 
     #[test]
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -280,7 +280,7 @@ pub async fn get_virtio_blk_ccw_device_name(
     sandbox: &Arc<Mutex<Sandbox>>,
     device: &ccw::Device,
 ) -> Result<String> {
-    let matcher = VirtioBlkCCWMatcher::new(&create_ccw_root_bus_path(), device);
+    let matcher = VirtioBlkCCWMatcher::new(CCW_ROOT_BUS_PATH, device);
     let uev = wait_for_uevent(sandbox, matcher).await?;
     let devname = uev.devname;
     return match Path::new(SYSTEM_DEV_PATH).join(&devname).to_str() {
@@ -1378,7 +1378,7 @@ mod tests {
     #[cfg(target_arch = "s390x")]
     #[tokio::test]
     async fn test_virtio_blk_ccw_matcher() {
-        let root_bus = create_ccw_root_bus_path();
+        let root_bus = CCW_ROOT_BUS_PATH;
         let subsystem = "block";
         let devname = "vda";
         let relpath = "0.0.0002";
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -50,11 +50,13 @@ pub const DRIVER_VFIO_PCI_GK_TYPE: &str = "vfio-pci-gk";
 // VFIO PCI device to be bound to vfio-pci and made available inside the
 // container as a VFIO device node
 pub const DRIVER_VFIO_PCI_TYPE: &str = "vfio-pci";
+pub const DRIVER_VFIO_AP_TYPE: &str = "vfio-ap";
 pub const DRIVER_OVERLAYFS_TYPE: &str = "overlayfs";
 pub const FS_TYPE_HUGETLB: &str = "hugetlbfs";
 
 cfg_if! {
     if #[cfg(target_arch = "s390x")] {
+        use crate::ap;
         use crate::ccw;
     }
 }
@@ -404,6 +406,39 @@ async fn get_vfio_device_name(sandbox: &Arc<Mutex<Sandbox>>, grp: IommuGroup) ->
 
     let uev = wait_for_uevent(sandbox, matcher).await?;
     Ok(format!("{}/{}", SYSTEM_DEV_PATH, &uev.devname))
+}
+
+#[cfg(target_arch = "s390x")]
+#[derive(Debug)]
+struct ApMatcher {
+    syspath: String,
+}
+
+#[cfg(target_arch = "s390x")]
+impl ApMatcher {
+    fn new(address: ap::Address) -> ApMatcher {
+        ApMatcher {
+            syspath: format!(
+                "{}/card{:02x}/{}",
+                AP_ROOT_BUS_PATH, address.adapter_id, address
+            ),
+        }
+    }
+}
+
+#[cfg(target_arch = "s390x")]
+impl UeventMatcher for ApMatcher {
+    fn is_match(&self, uev: &Uevent) -> bool {
+        uev.action == "add" && uev.devpath == self.syspath
+    }
+}
+
+#[cfg(target_arch = "s390x")]
+#[instrument]
+async fn wait_for_ap_device(sandbox: &Arc<Mutex<Sandbox>>, address: ap::Address) -> Result<()> {
+    let matcher = ApMatcher::new(address);
+    wait_for_uevent(sandbox, matcher).await?;
+    Ok(())
 }
 
 /// Scan SCSI bus for the given SCSI address(SCSI-Id and LUN)
@@ -772,6 +807,28 @@ async fn vfio_pci_device_handler(
     })
 }
 
+// The VFIO AP (Adjunct Processor) device handler takes all the APQNs provided as device options
+// and awaits them. It sets the minimum AP rescan time of 5 seconds and temporarily adds that
+// amoutn to the hotplug timeout.
+#[cfg(target_arch = "s390x")]
+#[instrument]
+async fn vfio_ap_device_handler(
+    device: &Device,
+    sandbox: &Arc<Mutex<Sandbox>>,
+) -> Result<SpecUpdate> {
+    // Force AP bus rescan
+    fs::write(AP_SCANS_PATH, "1")?;
+    for apqn in device.options.iter() {
+        wait_for_ap_device(sandbox, ap::Address::from_str(apqn)?).await?;
+    }
+    Ok(Default::default())
+}
+
+#[cfg(not(target_arch = "s390x"))]
+async fn vfio_ap_device_handler(_: &Device, _: &Arc<Mutex<Sandbox>>) -> Result<SpecUpdate> {
+    Err(anyhow!("AP is only supported on s390x"))
+}
+
 #[instrument]
 pub async fn add_devices(
     devices: &[Device],
@@ -840,6 +897,7 @@ async fn add_device(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) -> Result<Sp
         DRIVER_VFIO_PCI_GK_TYPE | DRIVER_VFIO_PCI_TYPE => {
             vfio_pci_device_handler(device, sandbox).await
         }
+        DRIVER_VFIO_AP_TYPE => vfio_ap_device_handler(device, sandbox).await,
         _ => Err(anyhow!("Unknown device type {}", device.field_type)),
     }
 }
@@ -1582,5 +1640,36 @@ mod tests {
 
         // Test dev2
         assert!(pci_iommu_group(&syspci, dev2).is_err());
+    }
+
+    #[cfg(target_arch = "s390x")]
+    #[tokio::test]
+    async fn test_vfio_ap_matcher() {
+        let subsystem = "ap";
+        let card = "0a";
+        let relpath = format!("{}.0001", card);
+
+        let mut uev = Uevent::default();
+        uev.action = U_EVENT_ACTION_ADD.to_string();
+        uev.subsystem = subsystem.to_string();
+        uev.devpath = format!("{}/card{}/{}", AP_ROOT_BUS_PATH, card, relpath);
+
+        let ap_address = ap::Address::from_str(&relpath).unwrap();
+        let matcher = ApMatcher::new(ap_address);
+
+        assert!(matcher.is_match(&uev));
+
+        let mut uev_remove = uev.clone();
+        uev_remove.action = U_EVENT_ACTION_REMOVE.to_string();
+        assert!(!matcher.is_match(&uev_remove));
+
+        let mut uev_other_device = uev.clone();
+        uev_other_device.devpath = format!(
+            "{}/card{}/{}",
+            AP_ROOT_BUS_PATH,
+            card,
+            format!("{}.0002", card)
+        );
+        assert!(!matcher.is_match(&uev_other_device));
     }
 }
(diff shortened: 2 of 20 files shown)


c3d:
@Jakob-Naucke 
> I have decided to prepend the agent device options with pci: or ap: respectively. If you strongly disagre, we can find another way.

Does not sound like a bad plan, but isn't it the case that the host pretty much determines the kind of device you have? In other words, are there platforms with both PCI and AP?

src/agent/src/ap.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+impl FromStr for Device {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        let split: Vec<&str> = s.split('.').collect();
+        if split.len() != 2 {
+            return Err(anyhow!(
+                "Wrong AP bus format. It needs to be in the form <xx>.<xxxx>, got {:?}",
```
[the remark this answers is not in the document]
Jakob-Naucke: Hmm. Elsewhere, we are doing things like: https://github.com/kata-containers/kata-containers/blob/a3b3274121f4bfb0a9d0ba0759d4d2e2a778134b/src/agent/src/pci.rs#L125 or https://github.com/kata-containers/kata-containers/blob/a3b3274121f4bfb0a9d0ba0759d4d2e2a778134b/src/agent/src/ccw.rs#L54 (that one was me, to be fair). What I had wanted to express with this format is make it clear that we're expecting hex numbers here. We can also put semantic names in the error message though. WDYT?

src/agent/src/ap.rs
```diff
@@ -0,0 +1,77 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_from_str() {
+        let device = Device::from_str("a.1").unwrap();
+        assert_eq!(format!("{}", device), "0a.0001");
+
+        assert!(Device::from_str("0.0.0").is_err());
```
jodh-intel, line 73: A couple of ideas for additional tests:

- Empty string.
- "`.`"
- Value that doesn't contain any periods.
- Value that is too big for the type.
- Non-hex numeric value.

Jakob-Naucke:
> > I have decided to prepend the agent device options with pci: or ap: respectively. If you strongly disagre, we can find another way.
> 
> Does not sound like a bad plan, but isn't it the case that the host pretty much determines the kind of device you have? In other words, are there platforms with both PCI and AP?

AP is for s390x, which (with some specialities) also has PCI support, so yes, but those would not be within one device, so I wouldn't really see the issue.

src/agent/src/device.rs
```diff
@@ -762,6 +836,45 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) ->
...
+            anyhow!(
+                "VFIO AP device option {} unexpectedly missed `ap:` indication",
+                ap_opt
+            )
+        })?;
+
+        let mut config = AGENT_CONFIG.write().await;
+        let hotplug_timeout = config.hotplug_timeout;
+        warn!(sl!(), "Temporarily adding {} seconds to hotplug timeout for AP devices due to rescan intervals", AP_CONFIG_TIME_SECONDS);
+        config.hotplug_timeout += time::Duration::from_secs(AP_CONFIG_TIME_SECONDS as u64 + 5);
```
jodh-intel, line 859: You're doing `+=` so why the `+ 5` at the end? I'm guessing that was the original value before you added `AP_CONFIG_TIME_SECONDS`?

src/agent/src/device.rs
```diff
@@ -762,6 +836,45 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) ->
...
+        // Drop write lock on config so that the uevent matcher can read from it
+        drop(config);
```
jodh-intel, line 861: I'm confused. You changed the config on the previous line, but are now dropping it? Should that `drop()` actually happen after the `write()`?

src/agent/src/device.rs
```diff
@@ -762,6 +836,45 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) ->
...
+        fs::write(AP_CONFIG_TIME_PATH, AP_CONFIG_TIME_SECONDS.to_string())?;
```
jodh-intel, line 862: Should this actually be?:

```suggestion
        fs::write(AP_CONFIG_TIME_PATH, config.hotplug_timeout.to_string())?;
```

src/agent/src/device.rs
```diff
@@ -762,6 +836,45 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) ->
...
+
+        for apqn in opt.split(',') {
+            wait_for_ap_device(sandbox, ap::Device::from_str(apqn)?).await?;
+        }
+
+        AGENT_CONFIG.write().await.hotplug_timeout = hotplug_timeout;
```
jodh-intel, line 868: Should this be `config.hotplug_timout` after the increment but before the `drop()`?

src/runtime/virtcontainers/device/config/config.go
```diff
@@ -263,9 +263,15 @@ type VFIODev struct {
...
 	GuestPciPath vcTypes.PciPath
 
+	// APDevices are the Adjunct Processor devices in the mdev. Mutually exclusive with PCI.
+	APDevices []string
+
 	// Type of VFIO device
 	Type VFIODeviceType
 
+	// IsAP specifies if device is AP or PCI
+	IsAP bool
```
jodh-intel, line 273: If this is `true`, `IsPCIe` will also be `true` right? Is this being `true` but `IsPCIe` false a legal state (denoting AP + PCIe?) I'm wondering if these values would be better as an `enum`?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1214,8 +1215,14 @@ func (k *kataAgent) appendVfioDevice(dev ContainerDevice, device api.Device, c *
...
 
-	for i, pciDev := range devList {
-		kataDevice.Options[i] = fmt.Sprintf("0000:%s=%s", pciDev.BDF, pciDev.GuestPciPath)
+	// The agent must handle VFIO-PCI and VFIO-AP devices differently.
+	// Prepend the options with `ap:` or `pci:` accordingly.
+	for i, device := range devList {
+		if device.IsAP {
+			kataDevice.Options[i] = fmt.Sprintf("ap:%s", strings.Join(device.APDevices, ","))
+		} else {
+			kataDevice.Options[i] = fmt.Sprintf("pci:0000:%s=%s", device.BDF, device.GuestPciPath)
```
jodh-intel, line 1224: This suggests "AP" trumps "PCI" right?

src/runtime/virtcontainers/device/config/config.go
```diff
@@ -263,9 +263,15 @@ type VFIODev struct {
 	// Guest PCI path of device
 	GuestPciPath vcTypes.PciPath
 
+	// APDevices are the Adjunct Processor devices in the mdev. Mutually exclusive with PCI.
+	APDevices []string
+
 	// Type of VFIO device
 	Type VFIODeviceType
 
+	// IsAP specifies if device is AP or PCI
```
jodh-intel, line 272: Is this comment correct? It seems a bit odd that there are two `bool`'s to denote a PCI device.

jodh-intel (commented):
Thanks @Jakob-Naucke - a few comments...

src/agent/src/device.rs
Jakob-Naucke: Oops, that is not necessary and was added in development in confusion over what turned out to be another bug. Will remove.

src/agent/src/device.rs
Jakob-Naucke: No :) Let me explain:
- Hotplugging, generally, can experience some delays. This is the normal `hotplug_timeout`, which defaults to 3 seconds.
- The uevent for AP devices will come in upon a rescan, which happens periodically.
- This period is configurable, down to a minimum of 5 seconds, and is configured in this line.

The `hotplug_timeout` that I want to use for AP is the "general" grace period of 3 seconds plus the up-to-5 seconds until rescan.

src/agent/src/device.rs
Jakob-Naucke: No, this is just the `hotplug_timeout` as it was before we updated it. This is a regular backup-and-restore.

src/runtime/virtcontainers/device/config/config.go
Jakob-Naucke: AP + PCIe would, indeed, not be legal. Let me look into it.

src/agent/src/device.rs
jodh-intel: Ah - it's the `lazy_static` global - ignore that comment as I'd read it as a local variable so it looked very odd to set some state, then dtrop it ;)

src/runtime/virtcontainers/kata_agent.go
Jakob-Naucke: Because of the order? I had just used that order to have
```
if cond {
  do_cond_stuff()
} else {
  do_non_cond_stuff()
}
```

instead of
```
if !cond {
  do_non_cond_stuff()
} else {
  do_cond_stuff()
}
```

but this would change with the enum anyway

src/runtime/virtcontainers/device/config/config.go
Jakob-Naucke: related to AP/PCI/PCIe discussion above

src/agent/src/ap.rs
```diff
@@ -0,0 +1,77 @@
...
+
+// IBM Adjunct Processor (AP) is the bus used by IBM Crypto Express hardware security modules on
+// IBM Z & LinuxONE (s390x)
+// AP bus ID follow the format <xx>.<xxxx> [1, p. 476], where
+//   - <xx> is the adapter ID, i.e. the card and
+//   - <xxxx> is the adapter domain.
+// [1] https://www.ibm.com/docs/en/linuxonibm/pdf/lku5dd05.pdf
+
+#[derive(Debug)]
+pub struct Device {
```
dgibson, line 19: AFAICT this is just encoding a device address, so I wonder if `Address` might be a better name, especially with the analogy to `crate::pci::Address`.

src/agent/src/ap.rs
dgibson: Seconded.  The unit testss in the `crate::pci` module may give some inspiration.

src/agent/src/device.rs
```diff
@@ -408,6 +416,39 @@ async fn get_vfio_pci_device_name(
     Ok(format!("{}/{}", SYSTEM_DEV_PATH, &uev.devname))
 }
 
+#[cfg(target_arch = "s390x")]
+#[derive(Debug)]
+struct VfioApMatcher {
```
dgibson, line 421: This is analogous to `PciMatcher` *not* to `VfioMatcher` which you renamed to `VfioPciMatcher`.

src/agent/src/device.rs
```diff
@@ -408,6 +416,39 @@ async fn get_vfio_pci_device_name(
...
+                AP_ROOT_BUS_PATH, device.adapter_id, device
+            ),
+        }
+    }
+}
+
+#[cfg(target_arch = "s390x")]
+impl UeventMatcher for VfioApMatcher {
+    fn is_match(&self, uev: &Uevent) -> bool {
+        uev.action == "add" && uev.devpath == self.syspath
```
dgibson, line 440: I do wonder if we should make a common matcher which just checks syspath, for all the cases where we can determine the exact sysfs path ahead of time.  Then we can just have multiple constructors or helper functions to generate that matcher based on whatever info we have.

That can be a clean up for another time though.

src/agent/src/device.rs
```diff
@@ -717,20 +758,48 @@ fn split_vfio_pci_option(opt: &str) -> Option<(&str, &str)> {
     }
 }
 
+// VFIO device could be PCI or AP, which requires different handling. The first option is checked
+// for the prefix, `pci:` or `ap:` (both within one device are not possible).
+#[instrument]
```
dgibson, line 763: Demultiplexing here seems silly to me.  Why not just create new device type strings for AP devices (and rename the existing PCI ones if we want)?

src/agent/src/device.rs
```diff
@@ -717,20 +758,48 @@ fn split_vfio_pci_option(opt: &str) -> Option<(&str, &str)> {
...
+async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) -> Result<SpecUpdate> {
+    let first_option = device
+        .options
+        .get(0)
+        .ok_or_else(|| anyhow!("VFIO device {} had no options", device.container_path))?;
+    match first_option.split_once(':') {
```
dgibson, line 769: Note that using `split_once` introduces a new Rust version dependency.

src/agent/src/device.rs
```diff
@@ -717,20 +758,48 @@ fn split_vfio_pci_option(opt: &str) -> Option<(&str, &str)> {
...
+            anyhow!(
+                "VFIO PCI device option {} unexpectedly missed `pci:` indication",
+                pci_opt
+            )
+        })?;
         let (host, pcipath) =
-            split_vfio_option(opt).ok_or_else(|| anyhow!("Malformed VFIO option {:?}", opt))?;
-        let host =
-            pci::Address::from_str(host).context("Bad host PCI address in VFIO option {:?}")?;
+            split_vfio_pci_option(opt).ok_or_else(|| anyhow!("Malformed VFIO option {:?}", opt))?;
```
dgibson, line 800: Uh... doesn't the fact that you're just changing `split_vfio_option` here mean you broke compile wtih the earlier patch that renamed the function?

src/agent/src/device.rs
dgibson: Yikes.  Making the global config writable just for this seems like a terrible idea to me.  That seems really likely to break people's assumptions that it's essentially a write-once global.

src/agent/src/device.rs
dgibson: I suspect this will not work quite the way you want, or at least, not reliably.  Part of the reason for 3s timeout is that there's often a ~5s timeout imposed by the layers above (e.g. CRI-O and/or k8s).  Making sure we time out first means we can give a useful error, rather than getting just a generic "container start up timed out" error.

But the point is, that increasing this timeout won't save you: if the hotplug takes too long, you'll still hit the upper levels timeout, just with a less meaningful error.  I've hit this in practice when I attempted to increase the timeout to deal with SHPC PCI hotplug (which also has a 5s delay).

Basically if yout hotplug takes this long, it's kind of fundamentally incompatible with container usage :(.  You may need to invent a different (VM only) hotplug mechanism that will complete more quickly.

src/runtime/virtcontainers/device/config/config.go
```diff
@@ -263,9 +263,15 @@ type VFIODev struct {
 	// Guest PCI path of device
 	GuestPciPath vcTypes.PciPath
 
+	// APDevices are the Adjunct Processor devices in the mdev. Mutually exclusive with PCI.
```
dgibson, line 266: What "in the mdev" means is unclear to me.  It's also unclear why they'd be mutually exclusive with PCI - s390x can have a PCI bus, IIUC, even if it's not particularly common.

src/runtime/virtcontainers/device/config/config.go
```diff
@@ -263,9 +263,15 @@ type VFIODev struct {
...
+	APDevices []string
```
dgibson, line 267: Why multiple devices here?  For the PCI, `VFIODev` represents a single PCI device, with multiple devices in the VFIO group being represented by multiple `VFIODev`s in a single `VFIODevice` (yes, the names are confusing).  Shouldn't the same model apply here?

Or, better yet `VFIODevice` could list either a bunch of `PCIDev`s or a bunch of `APDev`s, improving the naming and reducing redundancy in the `VFIODev` structure.

src/runtime/virtcontainers/kata_agent.go
dgibson: As I noted in the agent code, I think using a different `kataDevice.Type` value would make more sense than putting this into all the options.  You're also making things unnecessarily different from the PCI case by folding all the AP devices into a single option, whereas PCI devices are placed one per option.

src/runtime/virtcontainers/qemu.go
```diff
@@ -1570,7 +1571,7 @@ func (q *qemu) hotplugVFIODevice(ctx context.Context, device *config.VFIODev, op
 			case config.VFIODeviceNormalType:
 				err = q.qmpMonitorCh.qmp.ExecuteVFIODeviceAdd(q.qmpMonitorCh.ctx, devID, device.BDF, device.Bus, romFile)
 			case config.VFIODeviceMediatedType:
```
dgibson, line 1573: This is the main consumer of the type value.  So it seems like it might make more sense to make AP device a separate option at that level, rather than testing for mediated type *then* for `isAP`.

dgibson (changes requested):

src/agent/src/ap.rs
```diff
@@ -0,0 +1,77 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use std::fmt;
+use std::str::FromStr;
+
+use anyhow::{anyhow, Context};
+
+// IBM Adjunct Processor (AP) is the bus used by IBM Crypto Express hardware security modules on
+// IBM Z & LinuxONE (s390x)
```
jodh-intel, line 12: fyi, I harvested this for the [CC glossary](https://github.com/confidential-containers/documentation/wiki/Glossary#adjunct-processor) :smile:

src/agent/src/device.rs
Jakob-Naucke: @dgibson Thank you for your review. You made some important comments, but let me look into this before anything else because it's a potential showstopper.

Jakob-Naucke:
I'll put a comment at the end for readability: Because of the concerns raised in https://github.com/kata-containers/kata-containers/pull/3679#discussion_r807525412, this PR is on hold until we find a better solution than needing to wait 5 seconds for the related uevent. This seems possible and we are looking into this.

src/agent/src/device.rs
Jakob-Naucke: I have a kernel patch that resolves this issue from a colleague specialised in this at this point, but let's conclude :arrow_down: first

src/runtime/virtcontainers/qemu.go
Jakob-Naucke: You make some valid concerns about API cleanliness, but there is a catch I'd like to hear your opinion on. 
@dgibson, you suggest breaking this down by `config.VFIODev.Type`, which, until now, can be a `VFIODeviceNormalType` or a `VFIODeviceMediatedType`, i.e. we'd have e.g. a `VFIOAPDeviceMediatedType` (is always mediated) (and perhaps rename the others).
Meanwhile, IIUC, both you and @jodh-intel suggest breaking down `config.VFIODev` (which also assumes PCI through and through) into two devices. (James, you said `enum` -- I guess you meant an `interface`?)
However, implementing both of these would introduce redundancy and possible illegal states (e.g. a `config.VFIOAPDev` whose `Type` is `VfioPCIDeviceNormalType` or such).
And we cannot break up the VFIO devices "from the beginning" (i.e. `DeviceManager.NewDevice`), because only after checking IOMMU groups can you know whether a device is PCI or AP. 
WDYT?

(Note that this is not the distinction of `kataDevice.Type`, which currently has `vfio` or `vfio-gk` (where `gk` is guest kernel, i.e. not for the container, which changes the agent-side handling, which I shouldn't think we'd need for AP). I guess adding an option there is not a huge conflict though.)

Jakob-Naucke (commented):
@dgibson @jodh-intel please refer mostly to the longest comment :)

src/agent/src/device.rs
```diff
@@ -53,6 +52,12 @@ pub const DRIVER_VFIO_GK_TYPE: &str = "vfio-gk";
 pub const DRIVER_VFIO_TYPE: &str = "vfio";
 pub const DRIVER_OVERLAYFS_TYPE: &str = "overlayfs";
 
+cfg_if! {
+    if #[cfg(target_arch = "s390x")] {
+        use crate::ccw;
+    }
+}
+
```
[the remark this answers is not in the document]
dgibson: Right, I was reading the commits in order, so it wasn't obvious when I wrote this.

src/agent/src/ap.rs
dgibson: Um.. I don't really understand the question, sorry.

src/runtime/virtcontainers/qemu.go
dgibson: The Go code, and the agent-protocol is already full of places which have representable invalid states.  One more doesn't make much difference.

I doubt you need a `vfio-gk` equivalent for AP.  `vfio` passes through devices in the way that a non-Kata container would do so: a VFIO device in the host becomes a VFIO device in the guest.  `vfio-gk` is a effectively poorly-defined Kata extension, which has the VFIO device in the host become.. something.. in the guest depending what driver is available in the guest kernel.

Jakob-Naucke (commented):
@jodh-intel @dgibson sorry for the wait. I _think_ this is technically what you asked for, but I can't say I'm very happy with it. Please tell me what you think.

dgibson (commented):
I like this revision much more than the earlier one.

I'm now pretty much happy with the agent side code, apart from some minor things.  I do still have some thoughts on the organization of the runtime side code.

Jakob-Naucke (commented):
And here's to more updates @jodh-intel @dgibson. I tried to split them out as well as I could.

Jakob-Naucke:
Needed a `govet` update -- rebased

Jakob-Naucke:
needed test fix -- rebased

Jakob-Naucke:
@dgibson ping for another look if you can find the time :)

src/runtime/virtcontainers/device/drivers/vfio.go
```diff
@@ -182,10 +202,10 @@ func (device *VFIODevice) Save() persistapi.DeviceState {
...
 		if dev != nil {
 			ds.VFIODevs = append(ds.VFIODevs, &persistapi.VFIODev{
-				ID:       dev.ID,
-				Type:     uint32(dev.Type),
-				BDF:      dev.BDF,
-				SysfsDev: dev.SysfsDev,
+				ID:       *(*dev).GetID(),
+				Type:     uint32(*(*dev).GetType()),
+				BDF:      *(*dev).GetBDF(),
+				SysfsDev: *(*dev).GetSysfsDev(),
```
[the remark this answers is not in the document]
dgibson: New version is better.  Ugly, but that's a result of dealing with the ugly existing persist ingerface.

src/agent/src/ap.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) IBM Corp. 2022
```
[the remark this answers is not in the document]
dgibson: This doesn't seem to have been addressed - I still find the commit message pretty hard to read.

src/agent/src/device.rs
```diff
@@ -768,6 +807,28 @@ async fn vfio_device_handler(device: &Device, sandbox: &Arc<Mutex<Sandbox>>) ->
...
+        wait_for_ap_device(sandbox, ap::Address::from_str(apqn)?).await?;
+    }
+    Ok(Default::default())
+}
+
+#[cfg(not(target_arch = "s390x"))]
+async fn vfio_ap_device_handler(_: &Device, _: &Arc<Mutex<Sandbox>>) -> Result<SpecUpdate> {
+    Err(anyhow!("AP is only supported on s390x"))
+}
+
```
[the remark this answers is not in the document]
dgibson: Pretty dire: in some cases it could put a PCI device in an unusable state that required a *host* reboot to fix :(.  AP is very different of course, so it wouldn't suffer the same problems technically, but like PCI this is probably a rarely tested path.

src/runtime/virtcontainers/device/config/config.go
```diff
@@ -270,6 +288,91 @@ type VFIODev struct {
...
+	ID string
+
+	// sysfsdev of VFIO mediated device
+	SysfsDev string
+
+	// Type of VFIO device
+	Type VFIODeviceType
+
+	// APDevices are the Adjunct Processor devices assigned to the mdev
+	APDevices []string
```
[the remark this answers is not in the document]
dgibson: Right, but AFAICT the mediated VFIO device should map to the `VFIODevice` which can already list multiple `VFIODev`s, each oone can be a single APQN.  This doesn't seem different to me from the PCI case where a single `/dev/vfio/NNN` (a single `VFIODevice`) corresponds to multiple hardware devices in the VFIO (each a `VFIODev`).  This can happen, for example, if you have a bunch of devices beneath a PCIe to PCI bridge.

dgibson (changes requested):

src/runtime/virtcontainers/device/config/config.go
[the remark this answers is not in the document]
Jakob-Naucke: We go through `/sys/kernel/iommu_groups/NNN/devices`, but with VFIO-AP, that is just going to be one UUID, potentially with several APQNs behind it (see also https://docs.kernel.org/s390/vfio-ap.html). I think what you're describing applies to PCI normal, but not mediated.

Jakob-Naucke (commented):
@dgibson thanks for the iteration -- please see the comments below.

src/runtime/virtcontainers/clh.go
```diff
@@ -741,7 +748,7 @@ func (clh *cloudHypervisor) HotplugRemoveDevice(ctx context.Context, devInfo int
 	case BlockDev:
 		deviceID = clhDriveIndexToID(devInfo.(*config.BlockDrive).Index)
 	case VfioDev:
-		deviceID = devInfo.(*config.VFIODev).ID
+		deviceID = *(*(devInfo.(*config.VFIOPCIDev))).GetID()
```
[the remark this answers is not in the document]
Jakob-Naucke: Oops, that one wasn't in the update previously.

tools/packaging/kernel/patches/5.15.x/0001-s390-ap-enable-sysfs-attribute-scans-to-force-AP-bus.patch
```diff
@@ -0,0 +1,44 @@
+From 837cd1059a985ee610e4b72dcdddb287e7b97de0 Mon Sep 17 00:00:00 2001
+From: Harald Freudenberger <freude@linux.ibm.com>
+Date: Wed, 16 Feb 2022 14:26:48 +0100
+Subject: [PATCH] s390/ap: enable sysfs attribute scans to force AP bus rescan
```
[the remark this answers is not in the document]
hbrueckner: > Hm... ok. I don't love merging this in Kata before the change is merged in a released kernel upstream. Red Hat, for one, deploys Kata with a packaged guest kernel rather than something custom built with these patches.

This patch has been merged into 5.17-rc3 (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/s390/crypto/ap_bus.c?id=8944d05f9bbf910c8b241e29a3de114900e31e42)

src/runtime/virtcontainers/device/drivers/vfio.go
```diff
@@ -210,7 +210,11 @@ func (device *VFIODevice) Load(ds persistapi.DeviceState) {
 // It should implement GetAttachCount() and DeviceID() as api.Device implementation
 // here it shares function from *GenericDevice so we don't need duplicate codes
 func getVFIODetails(deviceFileName, iommuDevicesPath string) (deviceBDF, deviceSysfsDev string, vfioDeviceType config.VFIODeviceType, err error) {
-	vfioDeviceType = GetVFIODeviceType(deviceFileName)
+	sysfsDevStr := filepath.Join(iommuDevicesPath, deviceFileName)
+	vfioDeviceType, err = GetVFIODeviceType(sysfsDevStr)
```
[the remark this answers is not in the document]
dgibson: Right, using `filepath.Base` within the function is a fine approach.  I do suggest renaming the parameter to match as well, though.

src/runtime/virtcontainers/device/drivers/utils.go
```diff
@@ -90,18 +90,18 @@ func readPCIProperty(propertyPath string) (string, error) {
 	return strings.Split(string(buf), "\n")[0], nil
 }
 
-func GetVFIODeviceType(deviceFileName string) config.VFIODeviceType {
+func GetVFIODeviceType(deviceFileName string) (vfioDeviceType config.VFIODeviceType, err error) {
```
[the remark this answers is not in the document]
dgibson: Oof... ok.  Folding the error into the normal return type seems horrible; very against normal Go conventions.

src/runtime/virtcontainers/device/drivers/vfio.go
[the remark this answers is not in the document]
Jakob-Naucke: Done.

tools/packaging/kernel/patches/5.15.x/0001-s390-ap-enable-sysfs-attribute-scans-to-force-AP-bus.patch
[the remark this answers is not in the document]
Jakob-Naucke: Thanks @hbrueckner -- the included patch is now as released on kernel.org.

arronwy:

BbolroC:
Hi, I was wondering why this PR has been stuck quite a while. I would be grateful if someone from the reviewers could mention a reason and what is needed to be done in order to get this going forward. @Jakob-Naucke if you know why, please leave any meaningful feedback. Thanks.

Jakob-Naucke:
@BbolroC hmm, there were lots of good comments made by @dgibson who, like me, isn't really active here any more. Some things weren't settled 100% -- this would need further reviewer input.

I could resolve the conflicts, but I couldn't really test the updated version. I could sit down with you briefly to assist if you have trouble and when both of us find the time.

src/agent/src/ap.rs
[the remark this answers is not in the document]
c3d: Can you express the "hex" with an example?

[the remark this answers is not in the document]
BbolroC: You mean "e.g. 0a.0033" after "<xx>.<xxxx>", right? I could update this after this gets merged.

BbolroC:
handed over to @BbolroC

PR #3665, opened 2022-02-14
CCv0: Merge main into CCv0 branch

stevenhorsman:
Merge remote-tracking branch 'upstream/main' into CCv0

Re-attempting the merge from 14th Feb `main` after confidential-containers' containerd was updated.
    
Fixes: #3573
Signed-off-by: stevenhorsman <steven@uk.ibm.com>

PR #3680, opened 2022-02-15
C cv0 rebase main 14 feb

stevenhorsman:
The same code as https://github.com/kata-containers/kata-containers/pull/3665, but rebased on `main` to hopefully avoid some of the ci problems.

This should not be merged as it would impact the commit history of `CCv0`, but it just for test purposes


stevenhorsman:
It looks like with the conflicts we're not going to get good results from these tests, so closing this PR :(

Issue #3681, opened 2022-02-15
Support cross-building guest image


Jakob-Naucke:
Especially in the context of, but not limited to Confidential Containers, it would be useful to e.g. not require already trusted s390x hardware to be able to run on untrusted s390x hardware. Thus, as a start, we should be able to build a guest rootfs image from a foreign architecture.

PR #3682, opened 2022-02-15
Multistrap Ubuntu & enable cross-building guest

Jakob-Naucke:
osbuilder: Multistrap Ubuntu

Use `multistrap` for building Ubuntu rootfs. Adds support for building
for foreign architectures using the `ARCH` environment variable.
In the process, the Ubuntu rootfs workflow is vastly simplified.

---
packaging: Enable cross-building agent

Requires setting ARCH and CC. 

- Add CC linker option for building agent.
- Set host for building libseccomp.

Fixes: #3681

---
osbuilder: Remove musl installations

Remove a lot of cruft of musl installations -- we needed those for the 
Go agent, but Rustup just takes care of everything.

---
osbuilder: Source .cargo/env before checking Rust

We install Rust in the build containers, but we also install Rust in
`rootfs.sh` if it is missing. It makes sense to install Rust in the build
containers so it does not have to be installed every time, but for that check
to work on non-login shells, we should source `.cargo/env` before running it. 

---
/cc @stevenhorsman -- I have also tested this with CC, where some additional steps are required. Let's discuss this one first.


Jakob-Naucke:
Static checks need https://github.com/kata-containers/tests/issues/4465 to be fixed

jodh-intel (approved):
Thanks @Jakob-Naucke.

lgtm

Jakob-Naucke:
sorry -- also needs https://github.com/kata-containers/tests/pull/4505

Jakob-Naucke:
this is not going to work with ARM as is, let me post updates later

Jakob-Naucke:
A couple of notes:
- Foreign builds will not work with musl. I can only find pre-built binaries from an [unofficial HTTP website](http://musl.cc) and building the toolchains isn't worth it IMO.
- While x86_64 didn't need any system musl tooling, aarch64 actually does. I have added `musl-tools` to the Debian & Ubuntu workflows (CI uses Ubuntu), but I'm less sure about Alpine. @jongwu can I ask you to take a shot with this, Alpine guest on native aarch64? (Feel free to also try a cross-built GNU agent and Ubuntu, e.g. `make ARCH=aarch64 DISTRO=ubuntu LIBC=gnu USE_DOCKER=yes image` on x86_64, ditto @Amulyam24 for ppc64le).
If we can't make this work, we won't be able to skip all the musl installations (which I would have loved to lose, it's quite some LOC) and I'll have to rework this a little.

Regardless, I'd like to see /test now...

Jakob-Naucke:
@jongwu scratch that, it doesn't even work with `musl-tools`. I'll have to say goodbye to the big musl installation removal I guess :/ Let me update.

Jakob-Naucke:
also, OOI, let me highjack this for a /debug-s390x …

Jakob-Naucke:
Similar linking step required for Alpine -- /test

Jakob-Naucke:
That triggered a lint of something that was already there -- /test

fidencio:
@Jakob-Naucke, I actually would like to suggest a different approach, using https://github.com/tonistiigi/xx instead, as done on contained side (see: https://github.com/containerd/containerd/blob/main/.github/workflows/release/Dockerfile)

With that we could:
* For now, stick to using ubuntu / debian / alpine, as that's the supported package managers
* It supports all the targets we currently care about (see: https://github.com/tonistiigi/xx#supported-targets)
* Someone from Red Hat / IBM could take the bullet to add yum / dnf support

Jakob-Naucke:
@fidencio I must admit I don't fully understand where you'd want to hook `xx` in.

The workflow as is is:
- Build a container with some build tooling in it
- Use its build tooling to bootstrap a guest rootfs (`debootstrap`, `dnf --installroot …`, `apk --root …`)
- Use its build tooling to build the agent and install it in said rootfs
- Use another container to generate an image or an initrd from this rootfs

The workflow changes as currently suggested in my PR are:
- For Ubuntu, install cross toolchains if required
- Use `multistrap` instead (there was an article from Nathaniel McCallum somewhere about a foreign-arch bootstrap being made possible with `dnf` too, but I saw some issues initially)
- Use said cross toolchains to build the agent and install it
- image/initrd unchanged

I don't see where `xx` comes in.
I don't reckon `xx` would be very useful for doing the cross build of the agent -- Rust/LLVM work fine here. Likewise, `xx` appears to support installing packages of foreign architectures _for the container itself_ (which has its merits, don't get me wrong), but I don't see its use for bootstrapping a self-contained rootfs like we need. You could install a foreign architecture's `rustc` or `debootstrap`, but you still couldn't run them.

fidencio:
@Jakob-Naucke, I had building the runtime in my mind using xx, buit yeah, this PR is about the building the guest / rootfs.

fidencio (approved):
lgtm, thanks @Jakob-Naucke1

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -40,7 +39,11 @@ handle_error() {
 trap 'handle_error $LINENO' ERR
 
 # Default architecture
-ARCH=$(uname -m)
+export ARCH=${ARCH:-$(uname -m)}
+if [ "$ARCH" == "ppc64le" ] || [ "$ARCH" == "s390x" ]; then
+	LIBC=gnu
+	echo "WARNING: Forcing LIBC=gnu because $ARCH has no musl Rust target"
```
jodh-intel, line 45: Nit: Redirect to `stderr`?

tools/osbuilder/scripts/lib.sh
```diff
@@ -204,107 +203,27 @@ generate_dockerfile()
...
-			-e "s|@INSTALL_RUST@|${install_rust//$'\n'/\\n}|g" \
-			-e "s|@SET_PROXY@|${set_proxy:-}|g" \
-			"${dockerfile_template}" > Dockerfile
-	fi
+
+	sed \
+		-e "s#@OS_VERSION@#${OS_VERSION:-}#g" \
+		-e "s#@ARCH@#$ARCH#g" \
+		-e "s#@INSTALL_RUST@#${install_rust//$'\n'/\\n}#g" \
+		-e "s#@SET_PROXY@#${set_proxy:-}#g" \
```
jodh-intel, line 225: Nit: The hash symbols make this rather difficult to read. Commas might be clearer as they are not as tall :)

jodh-intel (approved):
Thanks @Jakob-Naucke.

Issue #3576, opened 2022-01-31
kernel: Upgrade to new kernel LTS 5.15.x


jcvenegas:
**Which feature do you think can be improved?**

Kata follows the latest LTS  kernel.  The version should be upgraded.

PR #3684, opened 2022-02-16
versions: Linux 5.15.x

fidencio:
Upgrade to new Linux kernel LTS version.

Fixes: #3576

Signed-off-by: Carlos Venegas <jose.carlos.venegas.munoz@intel.com>


fidencio:
@bergwolf, in the original PR you commented about the presence of `CONFIG_ARM64_UAO` in the `whitelist.conf` file and asked whether this should go under some ARM specific bit.  The issue with that is that, right now, we don't support `whitelist.conf` fragments.

bergwolf:
@fidencio What does `whitelist.conf` do? Can some file in `tools/packaging/kernel/configs/fragments/arm64` handle `CONFIG_ARM64_UAO`. IWO, if we leave `CONFIG_ARM64_UAO` in `whitelist.conf`, will it impact non-arm kernels?

fidencio:
> @fidencio What does `whitelist.conf` do? Can some file in `tools/packaging/kernel/configs/fragments/arm64` handle `CONFIG_ARM64_UAO`. IWO, if we leave `CONFIG_ARM64_UAO` in `whitelist.conf`, will it impact non-arm kernels?

From a comment in the file:
```
# configuration options which may dropped in newer kernels
# without generating an error in fragment merging
```

So, leaving `CONFIG_ARM64_UAO` in the `whitelist` will not impact other architectures, as the other architectures do not have that as part of their fragments.  And the impact on the arm side is that the config will be filtered-out if present in the fragments.

jodh-intel (approved):
Thanks @fidencio.

lgtm

Jakob-Naucke (approved):

bergwolf (approved):
Thanks @fidencio !

Issue #3342, opened 2021-12-23
hugepages support


liubin:
https://github.com/kata-containers/agent/pull/872

https://github.com/kata-containers/runtime/pull/3109

lilongjiang1:
@liubin  thanks for your job. Do we need to mount /sys/fs/cgroup/hugetlb, and write hugepage limits to hugetlb cgroup for the container?

liubin:
@lilongjiang1 Yes, the `resources` claimed by containers should be limited by cgroup in the guest OS.

PR #3344, opened 2021-12-23
feature: hugepages support

liubin:
Porting of:

- kata-containers/agent#872
- kata-containers/runtime#3109

Fixes: #3342

PR #3686, opened 2022-02-16
kata-agent hugepage support

shippomx:
I have read @liubin 's [commit](https://github.com/kata-containers/kata-containers/issues/3342), which have some useless codes. 

In main branch, kata-agent has supported the hugepages feature, but kata-runtime do not.

So I pass the hugepages params in kata-runtime in my commits. It works.

The compressed file [hugepages.cfg.tar.gz](https://github.com/kata-containers/kata-containers/files/8078590/hugepages.cfg.tar.gz) contains my configurations on my environment.
`config.toml` is `/etc/containerd/config.toml`
`configuration.toml` is `/etc/kata-containers/configuration.toml`
`ctnr.hugepages.toml` is container configuration
`pod.hugepages.toml` is pod configuration

### Step 1:
create the pod and container with these configuration which I mentioned upon.
```bash
crictl run --no-pull ctnr.hugepage.json pod.hugepage.json
```
It returns the container id, and pod id we can get with ctr command
- pod id: 7b9dc401b3574c8f7027bcec69dbcc1f361a9435c6f12d07b0c25b0fe768bd6c
- container id: 65188749240715da63bebe94909406b1f5b17c9c043e0afd457afde9b0bdfc4c

### Step 2: 
check the hugepage feature:
```bash
$ kata-runtime exec 7b9dc401b3574c8f7027bcec69dbcc1f361a9435c6f12d07b0c25b0fe768bd6c
bash-4.4# cat/sys/fs/cgroup/hugetlb/k8s.io/65188749240715da63bebe94909406b1f5b17c9c043e0afd457afde9b0bdfc4chugetlb.1GB.limit_in_bytes
1073741824

bash-4.4# cat/sys/fs/cgroup/hugetlb/k8s.io/65188749240715da63bebe94909406b1f5b17c9c043e0afd457afde9b0bdfc4chugetlb.2MB.limit_in_bytes
16777216

bash-4.4# cat /sys/kernel/mm/hugepages/hugepages-1048576kB/*
1
1
0
0
0

bash-4.4# cat /sys/kernel/mm/hugepages/hugepages-2048kB/*   
1024
1024
0
0
0

```
It  works well.

### DIFF
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -1315,7 +1315,7 @@ func (c *Container) hotplugDrive(ctx context.Context) error {
 			c.rootfsSuffix = ""
 		}
 		// If device mapper device, then fetch the full path of the device
-		devicePath, fsType, err = utils.GetDevicePathAndFsType(dev.mountPoint)
+		devicePath, fsType, _, err = utils.GetDevicePathAndFsTypeOptions(dev.mountPoint)
 		if err != nil {
 			return err
 		}
--- a/src/runtime/virtcontainers/device/config/pmem.go
+++ b/src/runtime/virtcontainers/device/config/pmem.go
@@ -75,7 +75,7 @@ func PmemDeviceInfo(source, destination string) (*DeviceInfo, error) {
 		return nil, fmt.Errorf("backing file %v has not PFN signature", device.HostPath)
 	}
 
-	_, fstype, err := utils.GetDevicePathAndFsType(source)
+	_, fstype, _, err := utils.GetDevicePathAndFsTypeOptions(source)
 	if err != nil {
 		pmemLog.WithError(err).WithField("mount-point", source).Warn("failed to get fstype: using ext4")
 		fstype = "ext4"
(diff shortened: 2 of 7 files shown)


jodh-intel:
@shippomx - Thanks for raising. Just to clarify, is this PR an alternative hugepages implementation to #3344? I don't see any comments from you on either that PR or #3342.

liubin:
@shippomx Do you mean that to support hugepages, only the runtime side needs to be implemented?

BTW `kata-runtime exec` will run a shell in the guest OS, but not the container. You should use `crictl exec` to check if container.

shippomx:
> @shippomx Do you mean that to support hugepages, only the runtime side needs to be implemented?
> 
> BTW `kata-runtime exec` will run a shell in the guest OS, but not the container. You should use `crictl exec` to check in the container.

Yes, your commit still pass the hugepage configuration with nil to the agent.
`src/runtime/virtcontainers/kata_agent.go`
```go
 func (k *kataAgent) constrainGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool, stripVfio bool) {
       ...
	grpcSpec.Linux.Resources.HugepageLimits = nil
      ...
```
As a result, guest OS have no hugepages limits on containers.

liubin:
@shippomx Thank you, but can you create an issue and raise a new PR to fix it if there is a problem?

And here is a basic guide about [contributing](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md).

shippomx:
ok, thanks for your advice

Issue #3688, opened 2022-02-16
clh: Ensure we build the binary with `--features tdx`


fidencio:
Right now TDx support on Cloud Hypervisor is gated behind a "--features
tdx" flag.  However, having TDx support enabled should not and does not
impact on the general usability of cloud-hypervisor.

As sooner than later we'll need kata-deploy binaries to be tested on a
CI that's TDx capable, for the confidential containers effort, let's
take the bullet and already enable it by default.

As we have different actions using this we need to ensure this gets propagated to:
* kata-deploy-push
* kata-deploy-test
* release

A counter part of this very same issue must be opened on the tests repo.

PR #3689, opened 2022-02-16
tools: Build cloud-hypervisor with "--features tdx"

fidencio:
Right now TDx support on Cloud Hypervisor is gated behind a "--features
tdx" flag.  However, having TDx support enabled should not and does not
impact on the general usability of cloud-hypervisor.

As sooner than later we'll need kata-deploy binaries to be tested on a
CI that's TDx capable, for the confidential containers effort, let's
take the bullet and already enable it by default.

By the way, touching kata-deploy-binaries.sh as it's ensure the change
will be used in the following workflows:
* kata-deploy-push
* kata-deploy-test
* release

Fixes: #3688

Together with this one, in a separate commit, there's also a small cleanup.

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -125,11 +125,6 @@ install_firecracker() {
 
 # Install static cloud-hypervisor asset
 install_clh() {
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @fidencio.

lgtm

devimc (approved):

PR #3697, opened 2022-02-17
Support Peer Pod VMs

yoheiueda:
This PR contains necessary changes in Kata shim and agent to support Peer Pod VMs. Peer Pod VMs are sandbox VMs that are not managed by local hypervisor, but by remote hypervisor. Typically, peer pod VMs are VM instances of IaaS Cloud.

The detailed description of Peer Pod VMs will be posted at https://github.com/confidential-containers/cloud-api-adaptor. 

The main changes are as follows.

* A new hypervisor type `remote` is introduced to support a out-of-tree hypervisor driver, which is running as a external process and communicates with the shim using TTRPC via a Unix domain socket.
* A new agent socket type is also introduced to support Unix domain socket in addition to VSock.
* Recursive file copying of shared files is implemented to support Kubernetes volumes

Note that this PR also includes some work around fixes for issues we encountered to run our prototype.

A minimum kata shim configuration (`/etc/kata-containers/configuration.toml`) looks like this.
```
[runtime]
internetworking_model = "none"
disable_new_netns = true
enable_pprof = true
enable_debug = true
[hypervisor.remote]
remote_hypervisor_socket = "/run/peerpod/hypervisor.sock"
[agent.kata]
```
A kata shim and external hypervisor driver process communicates using Unix Domain Socket `/run/peerpod/hypervisor.sock`.

### Architecture
Architecture


fengwang666:
@yoheiueda This is interesting! I would love to know more about this project. A few questions, 
1. How do you deal with the slowness of cloud VM provisioning? Would it make Kata container really slow?
2. What are the other potential feature and performance tradeoffs?
3. Is there a detail security analysis regarding this model?

yoheiueda:
@fengwang666

> 1. How do you deal with the slowness of cloud VM provisioning? Would it make Kata container really slow?

I did experiments using IBM Cloud VPC, and VM provisioning takes about 1 minute.  At least with this range of provisioning time, Kata containers and Kubernetes work fine without timeout errors. 

If we need faster provisioning, we need to implement some kind of optimization such as preallocation of VMs. VM preallocation are possibly not suitable with confidential computing scenarios, and we may need some ideas to implement it.

> 2. What are the other potential feature and performance tradeoffs?

We need a network tunnel between a worker node and pod VMs. All communication from/to a pod VM goes to its worker node VM. This may degrade network performance. We can implement some optimization to alleviate this network performance issue.

> Is there a detail security analysis regarding this model?

We haven't done yet, but I agree that we need security analysis.

The point here is that a pod workload does not trust worker nodes and network at all. We provide a network tunnel between worker node and pod VMs, but a pod workload must not trust the network. A worker node provides volume contents of ConfigMap and Secret, but a pod workload must not trust the integrity of the volume contents. Pod workloads must checks data integrity by themselves using security measures such as TLS and digital signatures.

fitzthum:
> If we need faster provisioning, we need to implement some kind of optimization such as preallocation of VMs. VM preallocation are possibly not suitable with confidential computing scenarios, and we may need some ideas to implement it.

I've thought about pre-alllocation for CC a little bit.

I think the potential for pre-allocation varies by platform. In general CC is relatively well-suited for pre-allocating VMs because the initrd is usually generic (even between different tenants). Secure Execution uses an encrypted image, which is less generic, but for other platforms you can potentially boot the VM before you know what workloads are going to run or even who will own the workloads. That said, for SEV(-ES) you would have to pause the guest before the secret injection phase, which means that you won't be able to load the kernel. You would at least get to save the time it takes the PSP to pre-encrypt the firmware (a serialized operation that can bog things down if you start a bunch of VMs at once). Of course, with peer pods a CSP would have to allow you to request VMs in this very specific manner and it's not clear how you would get secret injection to work via a CSP anyway. For TDX and SNP I think things might be easier.

arronwy:

yoheiueda:
@bpradipt @stevenhorsman I've rebased the PR to the latest CCv0 branch.

yoheiueda:
A workaround patch  https://github.com/kata-containers/kata-containers/pull/3697/commits/9a13b1adad35ca7d3a369a4eeacda35c4daf9c76 is still required for the "pause" container issue https://github.com/kata-containers/kata-containers/issues/4477 of CCv0 containerd.

stevenhorsman:
cc @kata-containers/architecture-committee - we are looking to get this PR for remote-hypervisor support reviewed soon and wanted to cover the topic of Peer pods in the AC meeting as part of that process

src/runtime/pkg/katautils/config.go
```diff
@@ -96,6 +97,7 @@ type hypervisor struct {
 	GuestHookPath                  string   `toml:"guest_hook_path"`
 	GuestMemoryDumpPath            string   `toml:"guest_memory_dump_path"`
 	SeccompSandbox                 string   `toml:"seccompsandbox"`
+	RemoteHypervisorSocket         string   `toml:"remote_hypervisor"`
```
pmores, line 100: Wouldn't it be clearer to call the TOML key simply `remote_hypervisor_socket`?

yoheiueda: I agree. I'll change it to `remote_hypervisor_socket`

yoheiueda: I renamed the option name to `remote_hypervisor_socket`, and also rebased the branch to the latest `CCv0` branch.

src/runtime/pkg/katautils/config.go
```diff
@@ -986,6 +989,18 @@ func newClhHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
...
 }
 
+func newRemoteHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
+
+	return vc.HypervisorConfig{
+		RemoteHypervisorSocket: h.RemoteHypervisorSocket,
+		// TODO: The values below are dummy ones to bypass the checks in validateHypervisorConfig
+		KernelPath:            "dummy",
+		ConfidentialGuest:     true,
+		HypervisorMachineType: virtcontainers.QemuCCWVirtio,
```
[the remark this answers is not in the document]
yoheiueda: Remote hypervisor is architecture agnostic. These values are dummy values to bypass the checks in `validateHypervisorConfig` of `virtcontainers/hypervisor_config_linux.go`.

This is just a work around, so I will introduce a flag in `HypervisorConfig` to indicate whether remote hypervisor is enable or not, and then skip the checks in `validateHypervisorConfig` when the flag is enabled.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -75,7 +76,7 @@ const (
 
 var (
 	checkRequestTimeout          = 30 * time.Second
-	defaultRequestTimeout        = 60 * time.Second
+	defaultRequestTimeout        = 300 * time.Second
```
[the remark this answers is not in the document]
yoheiueda: OK, I see. I'll change the code to use a lager timeout value only when remote hypervisor is enabled.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1262,6 +1272,31 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
...
+		// TODO: remove this code when containerd supports the new Sandbox API
+
+		image, err := getImageName(ociSpec.Annotations)
+		if err != nil {
+			k.Logger().Infof("no image spec available: %v", err)
+			return nil, err
+		}
+
+		//Send PullImageRequest only for "pause" image
+		if image == "k8s.gcr.io/pause:3.2" {
```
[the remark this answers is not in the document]
bpradipt: @yoheiueda for crio we'll need to pull all the images including pause. This ensures no changes to the crio side. Can we enable pullImageRequest for all the images in case of remote hypervisor ?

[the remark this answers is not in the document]
yoheiueda: I am going to implement image pulling in the cloud-api-adaptor side. https://github.com/confidential-containers/cloud-api-adaptor/issues/88

If cri-o provides image name in annotations, cloud-api-adaptor will be able to send pull image requests for all container images including pause.

The current code to get container image name in kata shim is this one.
https://github.com/yoheiueda/kata-containers/blob/8f176024bcd7d78b81439df75824e068b31eaddd/src/runtime/virtcontainers/kata_agent.go#L1316-L1338

For some reason, containerd v1.6.6-15 does not provide annotation for image name, so we still need the custom conatinerd.

@bpradipt do you know whether cri-o provide image name as annotation `io.kubernetes.cri-o.ImageName`?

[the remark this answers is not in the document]
bpradipt: @yoheiueda yes, the annotation is provided. 
This is from a running system
```
        "io.kubernetes.cri-o.ImageName": "quay.io/karmab/origin-coredns:latest",
        "io.kubernetes.cri-o.Name": "k8s_coredns_coredns-test-worker-0.ocp411.com_openshift-infra_dbbb729d580e7461e98ebf6dba93ab7c_1",
        "io.kubernetes.cri-o.ContainerType": "container",
        "io.kubernetes.cri-o.Stdin": "false",
```

src/runtime/pkg/katautils/config.go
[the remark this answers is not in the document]
yoheiueda: I changed `validateHypervisorConfig` to check whether `RemoteHypervisorSocket` is specified or not. If it is set, `validateHypervisorConfig` skips the remaining checks.

```
func checkHypervisorConfig(config vc.HypervisorConfig) error {
	if config.RemoteHypervisorSocket != "" {
		return nil
	}
```

src/runtime/virtcontainers/kata_agent.go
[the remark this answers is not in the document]
yoheiueda: I changed timeout value only for the remote hypervisor type.

src/runtime/virtcontainers/kata_agent.go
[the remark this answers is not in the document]
yoheiueda: I implemented https://github.com/confidential-containers/cloud-api-adaptor/issues/88, and remove the work around from kata shim.

I also added the cri-o support.

yoheiueda:
I rebased the code to the latest CCv0 branch.

I also added the `remote_hypervisor_timeout` parameter to specify a custom StartVM timeout value in seconds.

https://github.com/kata-containers/kata-containers/blob/6656c06924cc3b137f2086342538edaaeb851a96/src/runtime/pkg/katautils/config.go#L133

src/runtime/cmd/kata-runtime/kata-check_s390x.go
```diff
@@ -55,7 +55,14 @@ func kvmIsUsable() error {
 }
 
 func archHostCanCreateVMContainer(hypervisorType vc.HypervisorType) error {
-	return kvmIsUsable()
+	switch hypervisorType {
+	case "qemu":
+		return kvmIsUsable()
+	case "remote":
+		return nil
+	default:
```
wainersm, line 63: Aren't Firecracker and Cloud Hypervisor supported on s390x?

src/runtime/pkg/katautils/config.go
```diff
@@ -1564,6 +1578,11 @@ func checkFactoryConfig(config oci.RuntimeConfig) error {
 // checkHypervisorConfig performs basic "sanity checks" on the hypervisor
 // config.
 func checkHypervisorConfig(config vc.HypervisorConfig) error {
+
+	if config.RemoteHypervisorSocket != "" {
```
wainersm, line 1585: Would make sense to have some silly checks like "timeout is unsigned int greater than zero"?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -722,30 +729,38 @@ func (k *kataAgent) startSandbox(ctx context.Context, sandbox *Sandbox) error {
...
+			return err
+		}
+		if err = k.addARPNeighbors(ctx, neighs); err != nil {
+			return err
+		}
+
+		kmodules = setupKernelModules(k.kmodules)
 	}
 
 	storages := setupStorages(ctx, sandbox)
```
wainersm, line 763: Sorry to ask: does the storage setup keeps the same for remote hypervisor?

src/runtime/virtcontainers/pkg/agent/protocols/client/client.go
```diff
@@ -237,6 +238,11 @@ func parse(sock string) (string, *url.URL, error) {
 		}
 		hybridVSockPort = uint32(port)
 		grpcAddr = HybridVSockScheme + ":" + hvsocket[0]
+	case RemoteSockScheme:
+		if addr.Host != "" {
+			return "", nil, grpcStatus.Errorf(codes.InvalidArgument, "Invalid remote sock scheme: %s", sock)
```
wainersm, line 243: I suggest to explain it is invalid because the host address should be nil, if I understood correctly :)

src/runtime/virtcontainers/remote.go
```diff
@@ -0,0 +1,285 @@
...
+	if tracer, ok := err.(interface{ StackTrace() errors.StackTrace }); ok {
+		for _, f := range tracer.StackTrace() {
+			logrus.Errorf("%+s:%d\n", f, f)
+		}
+	}
+
+	return err
+}
+
+func (rh *remoteHypervisor) PauseVM(ctx context.Context) error {
```
wainersm, line 179: do you plan to implement any of the following not implemented functions still on this PR?

src/runtime/cmd/kata-runtime/kata-check_s390x.go
yoheiueda: They will possibly be supported on s390x. I'll fix the logic.

src/runtime/pkg/katautils/config.go
yoheiueda: OK. I'll add such checks.

src/runtime/virtcontainers/kata_agent.go
yoheiueda: [`setupStorage`](https://github.com/kata-containers/kata-containers/blob/6656c06924cc3b137f2086342538edaaeb851a96/src/runtime/virtcontainers/kata_agent.go#L804-L872) does nothing for remote hypervisor, since it does not support file system sharing.

I will change the code to call `setupStorage` unless remote hypervisor is used.

src/runtime/virtcontainers/pkg/agent/protocols/client/client.go
yoheiueda: OK. I'll add such explanation in the error message.

src/runtime/virtcontainers/remote.go
yoheiueda: I am thinking of a small start. I have implemented a minimum set of functions to run our demo of peer pod VMs.

If some important functions are missing in my PR, I am desired to implement such functions.

src/runtime/virtcontainers/remote.go
```diff
@@ -0,0 +1,287 @@
...
+func (rh *remoteHypervisor) setConfig(config *HypervisorConfig) error {
+	// Create a Validator specific for remote hypervisor
+	rh.config = *config
+
+	return nil
+}
+
+func (rh *remoteHypervisor) GetPids() []int {
+	// TODO
+	return []int{99999}
```
snir911, line 253: This pid is used by crio to fetch startTime, could you do please do something like:
```
diff --git a/src/runtime/virtcontainers/remote.go b/src/runtime/virtcontainers/remote.go
index 818da5167..ea1b3718f 100644
--- a/src/runtime/virtcontainers/remote.go
+++ b/src/runtime/virtcontainers/remote.go
@@ -8,6 +8,7 @@ import (
        "fmt"
        "net"
        "time"
+       "os"
 
        cri "github.com/containerd/containerd/pkg/cri/annotations"
        "github.com/containerd/ttrpc"
@@ -249,8 +250,8 @@ func (rh *remoteHypervisor) setConfig(config *HypervisorConfig) error {
 }
 
 func (rh *remoteHypervisor) GetPids() []int {
-       // TODO
-       return []int{99999}
+       // meanwhile let's use shim pid as it used by crio to fetch start time
+       return []int{os.Getpid()}
 }
 
 func (rh *remoteHypervisor) GetVirtioFsPid() *int {
 ```

File: src/runtime/pkg/katautils/config.go (hunk 6 above)
Comment (yoheiueda) [reply src/runtime/pkg/katautils/config.go:1659]: RemoteHypervisorSocket is an unsigned variable, and zero means no timoeout, so we don't need to non-negative check here.

File: src/runtime/cmd/kata-runtime/kata-check_s390x.go (hunk 5 above)
Comment (yoheiueda) [reply src/runtime/cmd/kata-runtime/kata-check_s390x.go]: done

File: src/runtime/virtcontainers/kata_agent.go (hunk 7 above)
Comment (yoheiueda) [reply src/runtime/virtcontainers/kata_agent.go:757]: I investigated the code, and found that `setupStorage` must be called even for remote hypervisor, so I keep the old code.

The following code needs to be called to prepare shm device on pod vm.
https://github.com/kata-containers/kata-containers/blob/f30931cda97fbbf6253eac97ed2858327ae9493e/src/runtime/virtcontainers/kata_agent.go#L851-L862

File: src/runtime/virtcontainers/pkg/agent/protocols/client/client.go (hunk 8 above)
Comment (yoheiueda) [reply src/runtime/virtcontainers/pkg/agent/protocols/client/client.go]: done

File: src/runtime/virtcontainers/remote.go (hunk 10 above)
Comment (yoheiueda) [reply src/runtime/virtcontainers/remote.go]: I've included this change.

**Comment by yoheiueda** [comment]
I've rebased the code to the latest CCv0 branch, and also included new changes based on review comments.

File: src/runtime/pkg/katautils/config.go (hunk 6 above)
Comment (wainersm) [reply src/runtime/pkg/katautils/config.go:1659]: Ok, makes sense. Thanks for addressing my comment!

**Comment by wainersm** [comment]
@yoheiueda thanks for addressing all my comments. As I mentioned before somewhere I am not an expert on this area, but my review didn't find any obvious problem. While it looks good to me, I prefer others with more knowledge to approve it for merging.

File: src/runtime/virtcontainers/fs_share_linux.go [hunk 11]
```diff
@@ -224,6 +224,52 @@ func (f *FilesystemShare) Cleanup(ctx context.Context) error {
...
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	return file.Readdirnames(0)
+}
+
+func (f *FilesystemShare) recursiveCopy(ctx context.Context, src, dst string) (bool, error) {
```
bergwolf, line 238: Is it possible to use https://pkg.go.dev/path/filepath#WalkDir instead? It should be able to save some lines.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2099,40 +2120,56 @@ func (k *kataAgent) setGuestDateTime(ctx context.Context, tv time.Time) error {
...
+		Gid:      int32(st.Gid),
 	}
 
-	fileSize := int64(len(b))
+	var b []byte
+
+	switch sflag := st.Mode & unix.S_IFMT; sflag {
+	case unix.S_IFREG:
+		var err error
+		b, err = ioutil.ReadFile(src)
```
bergwolf, line 2141: We should not read the file in one shot but it is not this PR's fault...

bergwolf (commented):
Generally lgtm. Just one minor comment.

src/runtime/virtcontainers/kata_agent.go
yoheiueda: I agree. I'll add a TODO comment here for now. We can fix the issue in another PR. 

I think we need to update the protobuf definition to support file range in CopyFileRequest.

src/runtime/virtcontainers/fs_share_linux.go
yoheiueda: done

src/runtime/virtcontainers/kata_agent.go
yoheiueda: done

yoheiueda:
@bergwolf I've rebased the code to the latest CCv0 branch, and also included new changes based on your review comments.

bergwolf (approved):
lgtm! Thanks!

yoheiueda:
@bergwolf Thank you very much!

egernst:
Sorry I haven't had much time to review. I reviewed first couple of patches and didn't see any major issues. Since this is for CCv0 branch, I don't have any concerns with putting in, and I will continue to review the hypervisor interface portion next.  Thanks for driving this ya'll!

src/runtime/pkg/katautils/config.go
pmores: @yoheiueda Thanks!

pmores (approved):
Look good to me (to the admittedly limited extent I can judge), and thanks for addressing my comment!

stevenhorsman:
> Sorry I haven't had much time to review. I reviewed first couple of patches and didn't see any major issues. Since this is for CCv0 branch, I don't have any concerns with putting in, and I will continue to review the hypervisor interface portion next. Thanks for driving this ya'll!

Thanks Eric, we'll get this merged in now, but as always can fix any problems in later PRs. Thanks for everyone's support!

PR #3721, opened 2022-02-18
kernel: remove SYS_SUPPORTS_HUGETLBFS from powerpc fragments

Amulyam24:
The name of kernel config `SYS_SUPPORTS_HUGETLBFS` has been changed to
`ARCH_SUPPORTS_HUGETLBFS` which is being selected on default
by another kernel config.

Fixes: #3720


jodh-intel (approved):
Thanks @Amulyam24.

lgtm

amshinde (approved):

jongwu:
> Hi @Amulyam24 - could you provide details in the commit of when that option was removed (which kernel versions are affected)?
> 
> Also, it looks like ARM might be affected by this @jongwu.

Yeah, it's the really the same with arm. As it does not affect the current default kernel version 5.10, we can hold the old one until the next kernel upgrade.

Issue #3722, opened 2022-02-18
kata-deploy containers don't support s390x


Jakob-Naucke:
use CentOS 7, which has no s390x support

Issue #3723, opened 2022-02-18
kata-deploy containers will not work with more recent systemd


Jakob-Naucke:
# Description of problem

kata-deploy is run as a "payload" container (we upload the images to [quay.io/kata-containers/kata-deploy](https://quay.io/repository/kata-containers/kata-deploy)). Its base image is CentOS 7: https://github.com/kata-containers/kata-containers/blob/12e83a99edd3a44119c88a113c7556b8043b26b2/tools/packaging/kata-deploy/Dockerfile#L5 which will be supported until 2024. That is some time to go, but in my attempt to fix s390x (which is not supported by CentOS < Stream 9), I discovered a problem that will arise at least then (see [TBA] for a "tactical" fix).

The simplest diff to reproduce is something like:

```diff
diff --git a/tools/packaging/kata-deploy/Dockerfile b/tools/packaging/kata-deploy/Dockerfile
index e89d24292..71990fd39 100644
--- a/tools/packaging/kata-deploy/Dockerfile
+++ b/tools/packaging/kata-deploy/Dockerfile
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM registry.centos.org/centos:7 AS base
+FROM fedora:28 AS base
 
 ENV container docker
 
@@ -29,8 +29,7 @@ COPY ${KATA_ARTIFACTS} ${WORKDIR}
 
 RUN \
 yum -y update && \
-yum install -y epel-release && \
-yum install -y bzip2 jq && \
+yum install -y xz systemd && \
 yum clean all && \
 mkdir -p ${DESTINATION} && \
 tar xvf ${KATA_ARTIFACTS} -C ${DESTINATION}/ && \
```

> Hints for testing: Run a local registry (`docker run -d -p 5000:5000 --name registry registry`). Build a local image and push it (`docker build -t <ip>:5000/kata-deploy:latest . && docker push <ip>:5000/kata-deploy:latest`). Use as image in `kata-deploy.yaml`. Might need config à la https://stackoverflow.com/a/67310470.)

# Expected result

kata-deploy still works as usual when using this image.

# Actual result

kata-deploy will start, but crash at `systemctl daemon-reload`.

> Hint for debugging: Use `set -x` and redirect STDERR.

# Further information

Running this pod with the same mounts interactively and executing `systemctl` yields
```
Failed to connect to bus: No data available
```

although `/run/dbus` is available. `strace` output looked unsuspicious to me :/

[...]

PR #3724, opened 2022-02-18
kata-deploy: Simplify Dockerfile and support s390x

Jakob-Naucke:
The kata-deploy Dockerfile is based on CentOS 7, which has no s390x
support. Add an `IMAGE` argument to specify the registry, which still
defaults to CentOS, but e.g. ClefOS can be selected instead.

Other x86_64 assumptions are also removed. Other general simplicifations
are made.

This does not address the more general issue of #3723 -- what we're
doing here does not seem to be working with systemd >= something between
235-237.

Fixes: #3722

/cc @bpradipt since you were asking -- this is a co-requirement for s390x CC operator support.

### DIFF
--- a/tools/packaging/kata-deploy/Dockerfile
+++ b/tools/packaging/kata-deploy/Dockerfile
@@ -1,27 +1,10 @@
-# Copyright Intel Corporation.
+# Copyright Intel Corporation, 2022 IBM Corp.
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM registry.centos.org/centos:7 AS base
-
-ENV container docker
-
-RUN (cd /lib/systemd/system/sysinit.target.wants/ && for i in *; do [ "$i" = systemd-tmpfiles-setup.service ] || rm -f "$i"; done); \
-rm -f /lib/systemd/system/multi-user.target.wants/*; \
-rm -f /etc/systemd/system/*.wants/*; \
-rm -f /lib/systemd/system/local-fs.target.wants/*; \
-rm -f /lib/systemd/system/sockets.target.wants/*udev*; \
-rm -f /lib/systemd/system/sockets.target.wants/*initctl*; \
-rm -f /lib/systemd/system/basic.target.wants/*; \
-rm -f /lib/systemd/system/anaconda.target.wants/*;
-
-VOLUME [ "/sys/fs/cgroup" ]
-
-CMD ["/usr/sbin/init"]
-
-FROM base
-
-ARG KUBE_ARCH=amd64
+# Specify alternative base image, e.g. clefos for s390x
+ARG IMAGE
+FROM ${IMAGE:-registry.centos.org/centos}:7
 ARG KATA_ARTIFACTS=./kata-static.tar.xz
 ARG DESTINATION=/opt/kata-artifacts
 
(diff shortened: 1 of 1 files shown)


tools/packaging/kata-deploy/Dockerfile
```diff
@@ -1,45 +1,23 @@
...
-ENV container docker
-
-RUN (cd /lib/systemd/system/sysinit.target.wants/ && for i in *; do [ "$i" = systemd-tmpfiles-setup.service ] || rm -f "$i"; done); \
-rm -f /lib/systemd/system/multi-user.target.wants/*; \
-rm -f /etc/systemd/system/*.wants/*; \
-rm -f /lib/systemd/system/local-fs.target.wants/*; \
-rm -f /lib/systemd/system/sockets.target.wants/*udev*; \
-rm -f /lib/systemd/system/sockets.target.wants/*initctl*; \
-rm -f /lib/systemd/system/basic.target.wants/*; \
-rm -f /lib/systemd/system/anaconda.target.wants/*;
```
fidencio, line 16: You cannot remove this bits, @Jakob-Naucke, as those are needed to have an image that's systemd capable.
Why do we need that, you may ask? Because of operations like: https://github.com/kata-containers/kata-containers/blob/7db9bef72ce2f99a2924fa6c7fd4f9706ded5305/tools/packaging/kata-deploy/scripts/kata-deploy.sh#L67-L68

I'd be fine switching any other distro, preferably smaller than CentOS, preferably with multi arch support, as long as we don't lose functionality.

If you're solving this in a different way and I missed it, please, just let me know.

fidencio (changes requested):
Please, take a look at my comment, but I think the changes will break the kata-deploy image.

tools/packaging/kata-deploy/Dockerfile
Jakob-Naucke: The image must be systemd-capable, but it is not systemd capable because we remove some systemd targets, but because of mounts. kata-deploy works without these removals.

Jakob-Naucke: > I'd be fine switching any other distro, preferably smaller than CentOS, preferably with multi arch support, as long as we don't lose functionality.

We're currently stuck with systemd <= ~237 (see #3723), and I think CentOS/ClefOS 7 are the only non-commercial distros that still support this :/

fidencio: > The image must be systemd-capable, but it is not systemd capable because we remove some systemd targets, but because of mounts. kata-deploy works without these removals.

Yep, I should have been more explicit marking every part of the file that worried me, such as the volume removal as well.

Anyways, test-kata-deploy seems to be happy enough about this, so I am fine with the changes.

fidencio:
kata-deploy seems happy enough this this: https://github.com/kata-containers/kata-containers/actions/runs/1874827427

fidencio:
static-checks are failind due to:
```
 INFO: Checking Dockerfile 'tools/packaging/kata-deploy/Dockerfile'
tools/packaging/kata-deploy/Dockerfile:21 DL3032 warning: `yum clean all` missing after yum command.
[static-checks.sh:1256] ERROR: failed to check Dockerfile 'tools/packaging/kata-deploy/Dockerfile'
```

fidencio (approved):
lgtm, thanks @Jakob-Naucke!

GabyCT (approved):

Jakob-Naucke:
I have decided not to backport. It's not reeeally a bugfix, and it isn't a lot of fun without publicly built deploy images anyway. I did this mainly for CC operator support.

Issue #3690, opened 2022-02-16
apply_patches.sh does not apply patches


Jakob-Naucke:
Discovered in applying kernel patches for #3679.
`tools/packaging/scripts/apply_patches.sh` runs `git apply $patch`, but this won't really apply to subdirectories:
```
~/go/src/github.com/kata-containers/kata-containers/tools/packaging/kernel/kata-linux-5.10.25-88$ git apply -v ../patches/5.10.x/0001-arm64-mmu-compared-with-linear-start-physical-addres.patch
Skipped patch 'arch/arm64/mm/mmu.c'.
```

PR #3691, opened 2022-02-16
packaging: Use `patch` for applying patches

Jakob-Naucke:
`tools/packaging/scripts/apply_patches.sh` uses `git apply $patch`, but
this will not apply to subdirectories. If one wanted to apply with
`git apply`, they'd have to run it with `--directory=...`
_relative to the Git tree's root_ (absolute will not work!). I suggest
we just use `patch`, which will do what we expected `git apply` would
do.

Fixes: #3690
Signed-off-by: Jakob Naucke <jakob.naucke@ibm.com>

/cc @wainersm

PR #3725, opened 2022-02-18
stable-2.3 | packaging: Use `patch` for applying patches

Jakob-Naucke:
`tools/packaging/scripts/apply_patches.sh` uses `git apply $patch`, but
this will not apply to subdirectories. If one wanted to apply with
`git apply`, they'd have to run it with `--directory=...`
_relative to the Git tree's root_ (absolute will not work!). I suggest
we just use `patch`, which will do what we expected `git apply` would
do.

`patch` is also added to build containers that require it.

Fixes: #3690

Backport of #3691


fidencio (approved):
lgtm, thanks @Jakob-Naucke!

Issue #3693, opened 2022-02-16
snap: cloud-hypervisor build for snap cannot handle the "build from sources" case


fidencio:
https://github.com/kata-containers/kata-containers/blob/b4a1150638a37efa21c07e80a4f1b1fca3009014/snap/snapcraft.yaml#L326-L340

We need to add support for the cases where building for source is required.

PR #3728, opened 2022-02-20
snap: clh: Re-use kata-deploy script here

fidencio:
The current snap build for clh is broken as it's not aware of how to
build the binary from sources.

Instead of fixing it here, let's take advantage of the kata-deploy
script, which is capable of building from sources, and re-use it here.

Fixes: #3693

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -320,17 +320,20 @@ parts:
     plugin: nil
     after: [godeps]
     override-build: |
-      export GOPATH=${SNAPCRAFT_STAGE}/gopath
-      yq=${SNAPCRAFT_STAGE}/yq
-      kata_dir=${GOPATH}/src/github.com/${SNAPCRAFT_PROJECT_NAME}/${SNAPCRAFT_PROJECT_NAME}
-      versions_file="${kata_dir}/versions.yaml"
-      version="$(${yq} r ${versions_file} assets.hypervisor.cloud_hypervisor.version)"
-      url="https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/${version}"
-      curl -L ${url}/cloud-hypervisor-static -o cloud-hypervisor
-      curl -LO ${url}/clh-remote
+      sudo apt-get -y update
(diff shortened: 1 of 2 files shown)


fidencio:
Okay, this PR depends on https://github.com/cloud-hypervisor/cloud-hypervisor/pull/3777, and in an update of Cloud Hypervisor then.

fidencio:
ARM CI: http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-ARM-PR/3424/console
This means we finally can use the very same script for building cloud-hypervisor in every single place where we build it. \o/

/test

fgiudici (approved):
LGTM, thanks @fidencio !

snap/snapcraft.yaml
```diff
@@ -320,17 +320,20 @@ parts:
     plugin: nil
     after: [godeps]
     override-build: |
+      sudo apt-get -y update
+      sudo apt-get -y install ca-certificates curl gnupg lsb-release
+      curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --batch --yes --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+      echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+      sudo apt-get -y update
+      sudo apt-get -y install docker-ce docker-ce-cli containerd.io
```
Jakob-Naucke, line 328: Absolutely not a blocker because architectures currently supported with CLH are mutex with those, but: This isn't packaged for all architectures, while the in-repo `docker.io` is. Are there specific reasons to use `download.docker.com`?

Jakob-Naucke (approved):

snap/snapcraft.yaml
fidencio: > Are there specific reasons to use download.docker.com?

Just following what's suggested in https://docs.docker.com/engine/install/ubuntu/

Issue #3228, opened 2021-12-07
kata-deploy creates no containerd-shim-kata-v2 binary


Jakob-Naucke:
Let me preface by saying I'm not sure whether this is intentional, but I find it a little inconsistent.
When applying kata-deploy, we create
```
$ ls -l /opt/kata/share/defaults/kata-containers/
total 64
-rw-r--r-- 1 root root  9384 Nov 29 19:42 configuration-acrn.toml
-rw-r--r-- 1 root root 11507 Nov 29 19:42 configuration-clh.toml
-rw-r--r-- 1 root root 14807 Nov 29 19:43 configuration-fc.toml
-rw-r--r-- 1 root root 24138 Nov 29 19:43 configuration-qemu.toml
lrwxrwxrwx 1 root root    23 Nov 29 19:43 configuration.toml -> configuration-qemu.toml
```

and in `/etc/containerd/config.toml`:
```toml
[plugins.cri.containerd.runtimes.kata]
  runtime_type = "io.containerd.kata.v2"
  …
  [plugins.cri.containerd.runtimes.kata.options]
    ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration.toml"
[plugins.cri.containerd.runtimes.kata-fc]
  runtime_type = "io.containerd.kata-fc.v2"
  …
  [plugins.cri.containerd.runtimes.kata-fc.options]
    ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration-fc.toml"
```

but only:
```
$ ls -l /usr/local/bin/*kata*
-rwxr-xr-x 1 root root 134 Dec  7 19:25 /usr/local/bin/containerd-shim-kata-clh-v2
-rwxr-xr-x 1 root root 133 Dec  7 19:25 /usr/local/bin/containerd-shim-kata-fc-v2
-rwxr-xr-x 1 root root 135 Dec  7 19:25 /usr/local/bin/containerd-shim-kata-qemu-v2
```

i.e. we create a `kata` runtime class that will use the QEMU config by default (:heart: for cross-platform compatibility) but no containerd shim to match? This means the `kata` runtime isn't usable as is. I think we should either:

1. remain neutral and remove the link and `kata` runtime class, or
2. accept QEMU as default and also create a `containerd-shim-kata-v2`.

Issue #3734, opened 2022-02-21
io.containerd.kata.v2" binary not installed "containerd-shim-kata-v2" while running a pod with Katacontainer through Kata-deploy


oshi36:
I am trying to set up Kata-Container with kata-deploy in a Kubernetes cluster with Containerd as CRI.

I am installing a stable version of kata-deploy with a containerd version as 1.5.5

Kata-deploy is been running also in the cluster `kubectl get pods -n kube-system`
```
NAME                             READY   STATUS    RESTARTS   AGE
coredns-64897985d-2xpfp          1/1     Running   0          6m26s
coredns-64897985d-bvghs          1/1     Running   0          6m26s
etcd-master                      1/1     Running   0          6m32s
kata-deploy-x2g5m                1/1     Running   0          4m36s
kube-apiserver-master            1/1     Running   0          6m32s
kube-controller-manager-master   1/1     Running   0          6m31s
kube-flannel-ds-zs6hg            1/1     Running   0          6m5s
kube-proxy-djbbz                 1/1     Running   0          6m27s
kube-scheduler-master            1/1     Running   0          6m32s
```

The logs of **kata-deploy-x2g5m**  pod are 
```
copying kata artifacts onto host
#!/bin/bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-fc.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
#!/bin/bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-qemu.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
#!/bin/bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-clh.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
Add Kata Containers as a supported runtime for containerd
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata]
  runtime_type = "io.containerd.kata.v2"
  privileged_without_host_devices = true
  pod_annotations = ["io.katacontainers.*"]
  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata.options]
    ConfigPath = "/opt/kata/share/defaults/kata-containers/configuration.toml"
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata-fc]
  runtime_type = "io.containerd.kata-fc.v2"
  privileged_without_host_devices = true
[...]

PR #3735, opened 2022-02-21
kata-deploy: Use (kata with) qemu as the default shim-v2 binary

fidencio:
When using kata-deploy, no `containerd-shim-kata-v2` binary is deployed,
but we do deploy a `kata` runtime class, which seems very much
incosistent.

As the default configuration for kata-containers points to QEMU, let's
also use kata with QEMU as the default shim-v2 binary.

Fixes: #3228, #3734

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -19,6 +19,8 @@ shims=(
 	"clh"
 )
 
+default_shim="qemu"
+
 # If we fail for any reason a message will be displayed
 die() {
         msg="$*"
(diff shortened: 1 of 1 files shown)


tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -112,6 +119,8 @@ function cleanup_different_shims_base() {
 			mv "$shim_backup" "$shim_file"
 		fi
 	done
+
+	rm /usr/local/bin/containerd-shim-kata-v2
```
Jakob-Naucke, line 123: TIL you don't need `-f` in non-interactive shells...

Jakob-Naucke (approved):
Thanks @fidencio!

snir911 (approved):
LGTM, thanks
FTR- test-kata-deploy passed (and seems to actually pull this change)

fidencio:
> LGTM, thanks FTR- test-kata-deploy passed (and seems to actually pull this change)

https://github.com/kata-containers/kata-containers/runs/5271806137?check_suite_focus=true

PR #3745, opened 2022-02-22
stable-2.3 | kata-deploy: Use (kata with) qemu as the default shim-v2 binary

fidencio:
When using kata-deploy, no `containerd-shim-kata-v2` binary is deployed,
but we do deploy a `kata` runtime class, which seems very much
incosistent.

As the default configuration for kata-containers points to QEMU, let's
also use kata with QEMU as the default shim-v2 binary.

Fixes: #3228, #3734
Backports: #3735

(cherry picked from commit 11220f052f3a7e7c2efd9db188d5aaf26b152ab7)

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -19,6 +19,8 @@ shims=(
 	"clh"
 )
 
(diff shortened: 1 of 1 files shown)


snir911 (approved):

c3d (approved):
Thanks @fidencio

Issue #3746, opened 2022-02-22
rust agent-protocol support async


quanweiZhou:
We plan to support runtime-rs and reuse the agent-protocols. Currently, agent-protocols does not support async-client, and the ttrpc and protobuf crate are older than ![shim-proto](https://github.com/containerd/rust-extensions/blob/main/crates/shim-protos/Cargo.toml), needs to be upgraded.

PR #3747, opened 2022-02-22
agent-protocol support async

quanweiZhou:
agent: agent-protocol support async

1. support async.
2. update ttrpc and protobuf
    update ttrpc to 0.6.0
    update protobuf to 2.23.0
3. support trans from oci

Fixes: #3746
Signed-off-by: Quanwei Zhou <quanweiZhou@linux.alibaba.com>


src/agent/protocols/src/trans.rs
```diff
@@ -0,0 +1,1079 @@
+use std::collections::HashMap;
```
lifupan, line 1: The license header missed.

quanweiZhou: have been added

lifupan (approved):

jiangliu:
The protocol crate has been moved to libs/protocol, it would be better to first merge main branch into runtime-rs. Otherwise it may cause trouble when merging runtime-rs into main branch later.

egernst:
(I’m still working on reviewing this)

src/agent/protocols/Cargo.toml
```diff
@@ -9,11 +9,12 @@ default = []
 with-serde = [ "serde", "serde_json" ]
 
 [dependencies]
-ttrpc = { version = "0.5.0", features = ["async"] }
+ttrpc = { version = "0.6.0" }
```
bergwolf, line 12: Does `0.6.0` enable `async` by default now?

src/agent/protocols/build.rs
```diff
@@ -107,7 +107,7 @@ fn real_main() -> Result<(), std::io::Error> {
         .for_each(|p| println!("cargo:rerun-if-changed={}", &p));
 
     let ttrpc_options = Customize {
-        async_server: true,
+        async_all: true,
```
bergwolf, line 110: There are still some concerns about adding async support to the runtime. Please either make `async_client` optional or get a consensus about async runtime before making such a change.

quanweiZhou: Add #[cfg(feature = "async")] to control whether to use sync or async

src/agent/protocols/Cargo.toml
quanweiZhou: ttrpc added when enabling async feature

bergwolf (approved):

Issue #3759, opened 2022-02-24
packaging: kernel: Add `CONFIG_IA32_EMULATION` to the kernel configuration


fidencio:
It's been reported on #kata-dev that we don't support 32bits workloads due to a kernel limitation on our guest VM, which comes with `CONFIG_IA32_EMULATION` not enabled.

Let's enable it as the first step and also double-check what may be needed on the rootfs side, later on, to fully support such workloads.

Greatsamps:
We have deployed this kernel, but there still appears to be something missing. We can get a simple application such as notepad.exe to run, but when trying to run something more complex, the program crashes.

The application we use uses sockets, and interestingly using this kernel we are also unable to attach to the supervisord socket via supervisorctl

Please see details of a test docker image that can be used.

Once running, you can http to the container on port 8080 and get into a vnc session. to re-produce the fault please run

wine /wine-prefix-32/drive_c/App/app.exe

it will crash after a short period.

the app should load a desktop application if working correctly.

container: greatsamps/dev:app

![Screenshot 2022-02-25 at 08 53 15](https://user-images.githubusercontent.com/19320680/155685704-f492cb42-cf7a-4d19-9dc3-a1bb073d8bbd.png)

fidencio:
I've took a stab and got the following container image working with my own kernel and image built using https://github.com/fidencio/kata-containers/tree/wip/32-bits-compat-with-ubuntu-18.04.

This use case has been briefly discussed in the Architecture Committee of March 8th, 2022, and this was considered a corner case, and whoever is dealing with this corner case should build their own image / kernel for that.

For now, I'm closing this PR and we should consider, maybe, writing a blog post so folks are aware on what to do in case this use-case becomes needed in the future.

I've synced with @Greatsamps privately and I think we're good on this front (but, please, feel free to re-open it in the future)/

PR #3760, opened 2022-02-24
kernel: Enable 32bits emulation support

fidencio:
Without this we cannot easily support 32bits workloads inside a Kata
Containers.

Fixes: #3759


bergwolf (approved):

devimc (approved):

fidencio:
As discussed in the Architecture Committee of March 8th, 2022, this is considered a corner case an someone dealing with this corner case should build their own image / kernel for that.

The last 3 commits of the following branch is what's needed for providing a kernel and an image for such use-cases: https://github.com/fidencio/kata-containers/tree/wip/32-bits-compat-with-ubuntu-18.04

For now, I'm closing this PR and we should consider, maybe, writing a blog post so folks are aware on what to do in case this use-case becomes needed in the future.

stevefan1999-personal:
@fidencio Is it possible to reconsider this PR? I have a bunch of game servers that needs Steam access, but it currently lacks 32-bit compatibility due to Steam native library is still 32-bit based, rendering Kata hard to use in this case.

Issue #3004, opened 2021-11-10
Build and run Kata Containers on other UNIXes host operating systems


sameo:
Kata Containers currently only runs on top of Linux, and both its design and implementation assume that Linux is and will be the only supported host operating system.

We'd like to extend Kata Containers host OS support by having it run on other UNIXes, like e.g. [Darwin](https://en.wikipedia.org/wiki/Darwin_(operating_system)). 

Non UNIX host operating systems (e.g. Windows) is out of this issue's scope.
And to be very clear: Non Linux **guest** operating systems is out of scope as well.

At the moment, Kata does not even build on anything else than Linux as it has many dependencies on Linux specific packages.
A solution to this problem would be to extract host OS agnostic pieces of code from the Linux specific ones, create interfaces out of the former while implementing them from the latter. The Linux dependencies would then live in different parts of the code base that we could build conditionally (Through e.g. `+build` tags or through file suffixes like e.g. `foo_linux.go`).

This is also an opportunity for cleaning parts of the code base where the Linux only assumptions allowed us to take architectural shortcuts. Making Kata build and run on non Linux host OSes is a forcing function for decoupling the OS agnostic parts from the Linux specific ones and creating cleaner interfaces and hopefully generating standalone packages

I will create sub issues for addressing all those dependencies one by one. At the moment there are 2 main areas where we have strong Linux dependencies and I will be initially working on those: 

* [ ] **Host networking** a.k.a. networking namespaces. The current `Network` implementation depends on Linux kernel APIs and Linux only packages.
* [ ] **Host resources control** a.k.a. host cgroups. This also is completely Linux specific and depend on a set of Linux specific container packages.

Jakob-Naucke:
I have no strong opinion on this, but are BSDs in scope for this?

fidencio:
@sameo, apart from the production environment, do you think we could get some CI runners to ensure this newly introduced support will keep working as expected?

sameo:
> I have no strong opinion on this, but are BSDs in scope for this?

The goal is for any BSD to be able to e.g. add its own `Network` implementation. It's possible that some BSDs will share some code with Linux and we'll have to play with build tags a bit further.

That being said, my first goal is to get things running on Darwin.

sameo:
> @sameo, apart from the production environment, do you think we could get some CI runners to ensure this newly introduced support will keep working as expected?

I'll work on that, I think this is definitely something that should be added to our CI.

[...]

PR #3774, opened 2022-02-25
cleanup runtime pkgs for Darwin build, add basic Darwin build/unit test

egernst:
A few of the existing Kata runtime packages are Linux specific only. To aid in being able to compile on alternative OS's, we can make some changes to have linux specific implementations were needed, in linux specific files.

With this PR, we make some adjustments to:
-  resource_control: allow it to build on Darwin, add non-Linux helpers to non-linux file.
- katautils: add a Darwin stub so katautils can (almost be) built on Linux and Darwin
- virtcontainers: first pass easy fixes for types compatibility

This addresses part of https://github.com/kata-containers/kata-containers/issues/3004.

After these changes, we still have a few fixes for virtcontainers package itself: 
 - add stub or actual hypervisor implementation for Darwin
 - address mount implementation in virtcontainers

Once those are fixed, we'll be able to add MacOS to the static checks GitHub action.  In the meantime, this PR also introduces a simple GitHub action to test build of the packages which we *do* expect to build. This action was verified and tested on my fork of the repository - see [test pr](https://github.com/egernst/kata-containers/pull/16):
image


egernst:
This is an interesting static-check failure, as this is required for building on Darwin, but results in a n'ack on Linux side:
```
14:41:12 utils.go:39:34: unnecessary conversion (unconvert)
14:41:13 	major := int64(unix.Major(uint64(st.Rdev)))
14:41:13 	                                ^
14:41:13 utils.go:40:34: unnecessary conversion (unconvert)
14:41:13 	minor := int64(unix.Minor(uint64(st.Rdev)))
```

egernst:
(retesting on top of https://github.com/kata-containers/tests/pull/4516)

bergwolf (approved):
lgtm

src/runtime/virtcontainers/utils/utils_darwin.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2022 Apple Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package utils
+
+func GetDevicePathAndFsTypeOptions(mountPoint string) (devicePath, fsType string, fsOptions []string, err error) {
+	return
```
Jakob-Naucke, line 9: I get that this is a stub to be expanded later, but this doesn't even build, does it? Wouldn't it be better to return empty, but language-correct values?

egernst: It builds -- it'll have default empty values, since the return values are named in the function signature.

Jakob-Naucke: ah, thanks -- I tested this on my side with a `bool`, but it does work with those types

Jakob-Naucke (approved):

egernst:
Looks like VFIO job had a flake -- restarted 🤞🏻

GabyCT (approved):

Issue #3783, opened 2022-02-28
nydusd process become zombie process if start failed


liubin:
For example with wrong command line options, nydusd start will fail, but the process will not exit and become `defunct`.

PR #3784, opened 2022-02-28
runtime: use Cmd.StdoutPipe instead of self-created pipe

liubin:
Nydusd uses a bufio.Scanner to check if nydusd process has
existed, but stderr/stdout passed to Cmd is self-created pipe,
this pipe will not be closed if the process start failing.

Use standard Cmd.StdoutPipe can close the stdout and kata shim
will detect the existence of the nydusd process, then call cmd.Wait to
reap the process' resources.

Fixes: #3783


Jakob-Naucke (approved):

fidencio (approved):
lgtm, thanks @liubin!

Issue #3631, opened 2022-02-07
selinux build is unconditional


egernst:
Andrea made changes into the Kata makefile to make SELinux build configurable, utilizing the _then_ supported pattern of setting selinux on build tag to indicate selinux should be used: https://github.com/kata-containers/kata-containers/commit/d78ffd653d051948218e8b8ba21a551ad15c2b45

This pattern is no longer supported in the selinux package: https://github.com/opencontainers/selinux/pull/132

We should either:
(1) update the Makefile to drop the tags utilization (always build for selinux), or
(2) update the selinux handling in Kata to again be build-time configurable (introduce our own build tags handling), or
(3) make selinux runtime configurable.

I think (2) makes most sense?

fidencio:
I'd go for either (2) or (3), as I can see use-cases where users may want to explicitly disable SELinux support on Kata Containers.
One of the cases that crosses my mind is, for instance, someone running a cluster with a mix of vanilla containers and kata-containers, with SELinux enabled on the CRI Engine side, on a system that doesn't have `container_kvm_t` label support (ie, old enough container-selinux package).

That's a valid situation where I can see the admin building kata-containers without SELinux support, as regardless of what the CRI Engine passes down, Kata Containers wouldn't be able to properly behave in unless SELinux support is disabled.

I don´t have a strong opinion on either of those, and I'd leave this up for the person implementing the solution, as long as we do **not** expose this via annotations.

One thing to keep in mind is that going for (3) instead of (2) is a change on the behaviour and it must be very well documented, advertised, and sync'ed with the consumers of the project.

egernst:
Yeah. I think (3) is more "maintainable" for the user (easier to edit a toml file than scp a binary). I was paranoid that someone could change the setting on the node, but that would require root access already to modify the file, at which point... we're in a pretty bad state. (this option, disable_vmm_selinux ?, would be disabled by default).

Any other feedback/suggestions?

tanweernoor:
acknowledged!! I will take a look. thanks

egernst:
Thanks Tanweer.  I think we'll want to add a config option (probably add it to the sandbox config), and then predicate setting process labels based on the flag value.

ie, if ! disable_vmm_selinux around this: https://github.com/kata-containers/kata-containers/blob/88a70d32ba5d3833ae84c8f65dd89fa17e3d0b1b/src/runtime/virtcontainers/qemu.go#L811-L818

[...]

PR #3794, opened 2022-03-01
back port:: runtime: make selinux configurable

egernst:
removes --tags selinux handling in the makefile (part of it introduced here: d78ffd6)
and makes selinux configurable via configuration.toml

Fixes: #3631
Signed-off-by: Tanweer Noor <tnoor@apple.com>


bergwolf (approved):

fidencio (approved):
lgtm, thanks @egernst!

Issue #3795, opened 2022-03-01
kata-monitor: Listen to localhost by default


jodh-intel:
Kata monitor should only listen to `127.0.0.1` by default.

PR #3796, opened 2022-03-01
Fix monitor listen address

jodh-intel:
monitor: Listen to localhost only by default

Change `kata-monitor` to listen to port `8090` on the local interface
only by default.

> **Note:**
>
> This is a breaking change as previously it listened on all interfaces.

Fixes: #3795.

### DIFF
--- a/src/runtime/cmd/kata-monitor/main.go
+++ b/src/runtime/cmd/kata-monitor/main.go
@@ -18,7 +18,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var monitorListenAddr = flag.String("listen-address", ":8090", "The address to listen on for HTTP requests.")
+const defaultListenAddress = "127.0.0.1:8090"
+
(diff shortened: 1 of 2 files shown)


liubin (approved):

src/runtime/cmd/kata-monitor/main.go
```diff
@@ -18,7 +18,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var monitorListenAddr = flag.String("listen-address", ":8090", "The address to listen on for HTTP requests.")
+const defaultListenAddress = "127.0.0.1:8090"
```
fgiudici, line 21: The idea to leave it running on all the addresses was because the main usage we (or just I? :-P ) had in mind was to have kata-monitor running as a DaemonSet on a Kubernetes cluster, exposing the endpoint to Prometheus.
This anyway could be used on the host directly, so makes sense to be on the safe side and just pass the address if/when running as a DaemonSet.

fgiudici (approved):
LGTM! Thanks @jodh-intel

Issue #3804, opened 2022-03-02
Update CODEOWNERS file


jodh-intel:
See https://github.com/kata-containers/community/issues/253.

PR #3805, opened 2022-03-02
codeowners: Expand scope

jodh-intel:
Improve the `CODEOWNERS` file by specifying more groups. The hope is
this will help reduce the PR backlog.

See: https://github.com/kata-containers/community/issues/253

Fixes: #3804.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,4 +9,55 @@
 # Order in this file is important. Only the last match will be
 # used. See https://help.github.com/articles/about-code-owners/
 
-*.md    @kata-containers/documentation
+VERSION				@kata-containers/release
+
+# The versions database needs careful handling
+versions.yaml			@kata-containers/ci @kata-containers/release @kata-containers/tests
+
+# Documentation related files could also appear anywhere
+# else in the repo.
+*.drawio			@kata-containers/documentation
+*.jpg				@kata-containers/documentation
+*.md				@kata-containers/documentation
+*.png				@kata-containers/documentation
+*.svg				@kata-containers/documentation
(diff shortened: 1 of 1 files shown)


jodh-intel:
As shown, the `CODEOWNERS` file contains errors relating to non existant teams. If folk agree, I'll create those...

fidencio:
@jodh-intel, I like the idea / approach, but I'd like to have a clear idea about each one of the teams and their members, in order to avoid two problems:
1. overload the same people who already are overloaded;
2. ensure folks there are responsive, avoiding an issue we currently have.

So, with this in mind, could you expand on who is part of those groups?

Also, the idea is to have more folks taking a look at this and that, but unresponsiveness is a real deal and what I can see happening is we approving a PR as long as the fastest 2 out of n groups marked reviewed a PR.  Would it be the need to have a policy for that?

fidencio:
Partially answering my own question here.

https://github.com/orgs/kata-containers/teams is the groups we have, and some of those groups have 20, 30 people.  Of those 20, 30, we're lucky to get 20% of the group actively reviewing code.  This will be a related but different thing to solve, re-working on who's part of each group, based on having actually being able to review when needed.

jodh-intel:
I'm not seeing quite as many comments as I was hoping for on this PR and https://github.com/kata-containers/tests/pull/4534 :disappointed: 

The idea of these files was to help speed up PRs landing. If there are no comments, I guess we can jsut close these PRs and maintain the _status quo_ :guitar: :guitar: :guitar: :guitar: :drum:

arronwy:

Issue #3810, opened 2022-03-02
clh: tdx: Confidential Guests should be using virtio-block rather than virtio-fs.


fidencio:
https://github.com/kata-containers/kata-containers/pull/3771 added TDX support for Cloud Hypervisor, but two big things got overlooked while doing that.
1. virtio-fs, as of now, **cannot** be part of the trust boundary, so the Confidential Guest will **not** be using it.
2. virtio-block hotplug should be enabled in order to use virtio-block for the rootfs (used with the `devmapper` plugin).

When trying to use cloud-hypervisor with TDX using virtio-fs, we're facing the following error on the guest kernel:
```
 "virtiofs virtio2: device must provide VIRTIO_F_ACCESS_PLATFORM"
```
After checking and double-checking with virtiofs and cloud-hypervisor developers, it happens as confidential containers might put some limitations on the device, so it can't access all of the guests' memory and that's where this restriction seems to be coming from.  Vivek mentioned that virtiofsd do not support VIRTIO_F_ACCESS_PLATFORM (aka VIRTIO_F_IOMMU_PLATFORM) yet, and that for ecrypted guests virtiofs may not be the best solution at the moment.

@sboeuf put this in a very nice way: "if the virtio-fs driver doesn't support VIRTIO_F_ACCESS_PLATFORM, then the pages corresponding to the virtqueues and the buffers won't be marked as SHARED, meaning the VMM won't have access to it".

Interestingly enough, it works with QEMU, and it may be due to some change done on the patched QEMU that @devimc is packaging, but we won't take the path to figure out what was the change and patch cloud-hypervisor on the same way, because of 1.

PR #3811, opened 2022-03-02
clh: tdx: Don't use sharedFS with Confidential Guests

fidencio:
clh: Don't use sharedFS with Confidential Guests

kata-containers/pulls#3771 added TDX support for Cloud Hypervisor, but
two big things got overlooked while doing that.

1. virtio-fs, as of now, cannot be part of the trust boundary, so the
   Confidential Guest will not be using it.

2. virtio-block hotplug should be enabled in order to use virtio-block
   for the rootfs (used with the devmapper plugin).

When trying to use cloud-hypervisor with TDX using virtio-fs, we're
facing the following error on the guest kernel:
```
virtiofs virtio2: device must provide VIRTIO_F_ACCESS_PLATFORM
```

After checking and double-checking with virtiofs and cloud-hypervisor
developers, it happens as confidential containers might put some
limitations on the device, so it can't access all of the guests' memory
and that's where this restriction seems to be coming from. Vivek
mentioned that virtiofsd do not support VIRTIO_F_ACCESS_PLATFORM (aka
VIRTIO_F_IOMMU_PLATFORM) yet, and that for ecrypted guests virtiofs may
not be the best solution at the moment.

@sboeuf put this in a very nice way: "if the virtio-fs driver doesn't
support VIRTIO_F_ACCESS_PLATFORM, then the pages corresponding to the
virtqueues and the buffers won't be marked as SHARED, meaning the VMM
won't have access to it".

Interestingly enough, it works with QEMU, and it may be due to some
change done on the patched QEMU that @devimc is packaging, but we won't
take the path to figure out what was the change and patch
cloud-hypervisor on the same way, because of 1.

Fixes: #3810

---

Revert "hypervisors: Confidential Guests do not support Device hotplug"

This reverts commit df8ffecde0b4190c6c8ce8ee21e871ceafe132f9, as device
hotplug *is* supported and, more than that, is very much needed when
using virtio-blk instead of virtio-fs.

---

And here is what I finally am able to get as a result of spawning a simple pod using the proposed patches:
```
# ctr run --snapshotter devmapper --rm --runtime io.containerd.run.kata-clh-tdshim.v2 -t --rm docker.io/library/busybox:latest fidencio sh
/ # dmesg | grep -i tdx
[    0.000000] tdx: Force enabling TDX Guest feature
[    0.000000] TDX: Disabled TDX guest filter support
[    0.000000] tdx: Setting event notification interrupt failed
[    0.000000] tdx: Guest initialized

```

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -638,10 +638,6 @@ func (clh *cloudHypervisor) HotplugAddDevice(ctx context.Context, devInfo interf
 	span, _ := katatrace.Trace(ctx, clh.Logger(), "HotplugAddDevice", clhTracingTags, map[string]string{"sandbox_id": clh.id})
 	defer span.End()
 
-	if clh.config.ConfidentialGuest {
-		return nil, errors.New("Device hotplug addition is not supported in confidential mode")
-	}
-
 	switch devType {
 	case BlockDev:
 		drive := devInfo.(*config.BlockDrive)
@@ -658,10 +654,6 @@ func (clh *cloudHypervisor) HotplugAddDevice(ctx context.Context, devInfo interf
 func (clh *cloudHypervisor) HotplugRemoveDevice(ctx context.Context, devInfo interface{}, devType DeviceType) (interface{}, error) {
 	span, _ := katatrace.Trace(ctx, clh.Logger(), "HotplugRemoveDevice", clhTracingTags, map[string]string{"sandbox_id": clh.id})
 	defer span.End()
-
-	if clh.config.ConfidentialGuest {
-		return nil, errors.New("Device hotplug removal is not supported in confidential mode")
-	}
 
 	var deviceID string
 
@@ -917,9 +909,7 @@ func (clh *cloudHypervisor) Capabilities(ctx context.Context) types.Capabilities
 	clh.Logger().WithField("function", "Capabilities").Info("get Capabilities")
 	var caps types.Capabilities
 	caps.SetFsSharingSupport()
-	if !clh.config.ConfidentialGuest {
-		caps.SetBlockDeviceHotplugSupport()
-	}
+	caps.SetBlockDeviceHotplugSupport()
 	return caps
 }
 
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -257,6 +257,22 @@ func (clh *cloudHypervisor) setupVirtiofsDaemon(ctx context.Context) error {
 	return nil
 }
 
+func (clh *cloudHypervisor) stopVirtiofsDaemon(ctx context.Context) (err error) {
+	if clh.state.VirtiofsDaemonPid == 0 {
+		clh.Logger().Warn("The virtiofsd had stopped")
+		return nil
+	}
+
+	err = clh.virtiofsDaemon.Stop(ctx)
+	if err != nil {
+		return err
+	}
+
+	clh.state.VirtiofsDaemonPid = 0
+
+	return nil
+}
+
 func (clh *cloudHypervisor) nydusdAPISocketPath(id string) (string, error) {
 	return utils.BuildSocketPath(clh.config.VMStorePath, id, nydusdAPISock)
 }
@@ -516,7 +532,7 @@ func (clh *cloudHypervisor) StartVM(ctx context.Context, timeout int) error {
 
 	pid, err := clh.launchClh()
 	if err != nil {
-		if shutdownErr := clh.virtiofsDaemon.Stop(ctx); shutdownErr != nil {
+		if shutdownErr := clh.stopVirtiofsDaemon(ctx); shutdownErr != nil {
 			clh.Logger().WithError(shutdownErr).Warn("error shutting down VirtiofsDaemon")
 		}
 		return fmt.Errorf("failed to launch cloud-hypervisor: %q", err)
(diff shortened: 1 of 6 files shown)


fidencio:
> Do we at this point have any design doc around Kata + TDX, and/or more specifically Kata + CLH + TDX?

We're working on that as well, Eric: https://github.com/kata-containers/kata-containers/pull/3713.

Out of curiosity, I wonder if there's any specific reason to ask that for Kata + TDX but there wasn't such requirement for the other protections added in the past (SEV, SE, for instance).

fidencio:
VFIO CI was interrupted.
/retest-vfio

devimc (approved):
thanks @fidencio - QEMU + tdx works well

fidencio:
https://github.com/kata-containers/kata-containers/pull/3812 is a fix for the PPC CI.

src/runtime/virtcontainers/clh.go
```diff
@@ -638,10 +638,6 @@ func (clh *cloudHypervisor) HotplugAddDevice(ctx context.Context, devInfo interf
 	span, _ := katatrace.Trace(ctx, clh.Logger(), "HotplugAddDevice", clhTracingTags, map[string]string{"sandbox_id": clh.id})
 	defer span.End()
 
-	if clh.config.ConfidentialGuest {
-		return nil, errors.New("Device hotplug addition is not supported in confidential mode")
-	}
-
```
sameo, line 644: My fault for mentioning that device hot plug is not supported with TDX. Even though virtio device hot plug is supported, I don't think passthrough is. As a matter of fact, I don't think you can do passthrough at all with neither SEV nor TDX.

src/runtime/virtcontainers/clh.go
```diff
@@ -187,6 +187,12 @@ var clhDebugKernelParams = []Param{
 //
 //###########################################################
 
+func (clh *cloudHypervisor) supportsSharedFS() bool {
+	caps := clh.Capabilities(clh.ctx)
+
+	return caps.IsFsSharingSupported()
```
sameo, line 193: nit: `return clh.Capabilities(clh.ctx).IsFsSharingSupported()`

src/runtime/virtcontainers/clh.go
```diff
@@ -255,12 +261,16 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
...
-			debug:      clh.config.Debug,
-			socketPath: virtiofsdSocketPath,
+
+		if clh.supportsSharedFS() {
+			clh.virtiofsDaemon = &virtiofsd{
+				PID:        clh.state.VirtiofsDaemonPid,
+				sourcePath: hypervisorConfig.SharedPath,
+				debug:      clh.config.Debug,
+				socketPath: virtiofsdSocketPath,
+			}
```
sameo, line 271: I feel this adds a bit of complexity to the code and makes it slightly less easy to read/review. It could be worth defining a set of `init/start/stop` routines for managing the virtiofsd lifecycle and include the `supportsSharedFS()` call in there. This would make for a more readable code imho.
But that's not a show stopper for that PR.

src/runtime/virtcontainers/clh.go
```diff
@@ -908,7 +930,9 @@ func (clh *cloudHypervisor) Capabilities(ctx context.Context) types.Capabilities
 
 	clh.Logger().WithField("function", "Capabilities").Info("get Capabilities")
 	var caps types.Capabilities
-	caps.SetFsSharingSupport()
+	if !clh.config.ConfidentialGuest {
+		caps.SetFsSharingSupport()
+	}
```
sameo, line 935: Call me pedantic, but I would add this blob and the one where you define `supportsSharedFS()` in a separate commit. Just to make things easier to follow.

egernst:
> > Do we at this point have any design doc around Kata + TDX, and/or more specifically Kata + CLH + TDX?
> 
> We're working on that as well, Eric: #3713.
> 
> Out of curiosity, I wonder if there's any specific reason to ask that for Kata + TDX but there wasn't such requirement for the other protections added in the past (SEV, SE, for instance).

If there aren't docs there, then that's a gap too.  I thought to ask here since there was initial confusion on what is and what is not supported, etc. With a document we can list out the expectations, feature details, limitations, etc.

fidencio:
@sameo, I've updated the series covering pretty much all the points you raised, apart from the fact that VFIO is not supported for Confidential Guests, as I'd like to address that one in a different PR (which will then touch QEMU as well).

Does this sound reasonable? Please, let me know if there's still something pending from your side.

Last but not least, I've used the same names of the virtio-fs daemon helpers as the ones used on the QEMU side, for the sake of consistency.

src/runtime/virtcontainers/clh.go
fidencio: Addressed!

src/runtime/virtcontainers/clh.go
fidencio: Addressed!

src/runtime/virtcontainers/clh.go
fidencio: caps is a pointer and the compiler complains if we do that.  The cleaner way I found is this.

src/runtime/virtcontainers/clh.go
fidencio: I'd like to cover this in a different PR, if you don't mind.

fidencio:
Re-pushed the series foxing `go fmt` issues, my bad.

sameo (approved):

PR #3812, opened 2022-03-02
snap: Don't build cloud-hypevisor on ppc64le

fidencio:
snapcraft build is failing due to:
 ```
utils.mk:130: "WARNING: powerpc64le-unknown-linux-musl target is unavailable"
```

It seems to happen as powerpc64-unknown-linux-musl is a target that
although there's support for it, it's not exactly built or
automatically tested, at least according to:
https://doc.rust-lang.org/rustc/platform-support.html

Fixes: #3803

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -320,20 +320,23 @@ parts:
     plugin: nil
     after: [godeps]
     override-build: |
-      sudo apt-get -y update
-      sudo apt-get -y install ca-certificates curl gnupg lsb-release
-      curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --batch --yes --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
-      echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
-      sudo apt-get -y update
-      sudo apt-get -y install docker-ce docker-ce-cli containerd.io
-      sudo systemctl start docker.socket
+      arch=$(uname -m)
+      if [ "{$arch}" == "aarch64" ] || [ "${arch}" == "x64_64" ]; then 
+          sudo apt-get -y update
+          sudo apt-get -y install ca-certificates curl gnupg lsb-release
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by fidencio** [comment]
A green Power CI: http://jenkins.katacontainers.io/job/kata-containers-2.0-ppc64le-containerd-k8s-ubuntu-20-04-PR/127/

**Review by Jakob-Naucke — APPROVED** [review]
@fidencio no blockers, but a surprise: According to the linked issue, this was introduced by #3728, but I can't see why. Can you?
Also I'm afraid that CLH support isn't just dictated by a musl Rust target, but that's another can of worms...

**Comment by fidencio** [comment]
> @fidencio no blockers, but a surprise: According to the linked issue, this was introduced by #3728, but I can't see why. Can you? Also I'm afraid that CLH support isn't just dictated by a musl Rust target, but that's another can of worms...

Previously to that commit we'd simply do:
```
$ curl -L https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/${version}/cloud-hypervisor -o cloud-hypervisor
```

That would work well if the binary existed, if the release was there.  But consider version = `batata` (which means potato, in pt_BR), then we'd do:
```
[fidencio@pessoa ~]$ curl -L https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/batata/cloud-hypervisor-static -o cloud-hypervisor
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0     25      0 --:--:-- --:--:-- --:--:--    25
[fidencio@pessoa ~]$ echo $?
0
[fidencio@pessoa ~]$ cat cloud-hypervisor 
Not Found
```

So, it never worked for anything that didn't have a release out.

---

About actually building cloud-hypervisor for power, if that's desired, taking a look at https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/scripts/dev_cli.sh may be the first step.  By the way, I don't think it supports ppc64le, at all.

Do you want me to adjust the commit message and update the PR?

Jakob-Naucke:
Ah, that makes sense. No strong opinion about the commit message -- it's not false per se, more like an understatement :)
About adding support -- cc @Amulyam24 I guess but supporting a new architecture in a hypervisor is probably a non-trivial undertaking (I can almost promise it is for s390x because of PCI vs. CCW).

devimc (approved):

Issue #3813, opened 2022-03-02
Update the comments about disable_block_device_use


fidencio:
`disable_block_device_use` has outdated comments, which leads to the reader to assume that virtio-9p is the default option for file sharing.

Together with this, the option is exposed on `firecracker`, which **must** use virtio-block.

PR #3814, opened 2022-03-02
Minor fixes for the `disable_block_device_use` comments

fidencio:
config: fc: Don't expose disable_block_device_use

Relying on virtio-block is the *only* way to use Firecracker with Kata
Containers, as shared FS (virtio-{fs,fs-nydus,9p}) is not supported by
Firecracker.

As configuration doesn't make sense to be exposed, we hardcode the
`false` value in the Firecracker configuration structure.

---

config: qemu: Fix disable_block_device_use comments

virtio-fs, instead of virtio-9p, is the default shared file system type
in case virtio-blk is not used.

---

Fixes: #3813

### DIFF
--- a/src/runtime/config/configuration-fc.toml.in
+++ b/src/runtime/config/configuration-fc.toml.in
@@ -103,14 +103,6 @@ default_memory = @DEFMEMSZ@
 # Default 0
 #memory_offset = 0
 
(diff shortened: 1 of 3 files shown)


cmaf (approved):
thanks @fidencio

egernst (approved):

Issue #3819, opened 2022-03-03
clh: virtiofsd: Process is not stopped if Cloud Hypervisor fails to boot up the VM


fidencio:
While working on an unrelated refactor on Cloud Hypervisor (see: https://github.com/kata-containers/kata-containers/pull/3811), I've noticed that the process will not be stopped if Cloud Hypervisor is launched but, for some reason, fails to boot the VM up.

The faulty code can be seen here:
```
	pid, err := clh.launchClh()
	if err != nil {
		if shutdownErr := clh.virtiofsDaemon.Stop(ctx); shutdownErr != nil {
			clh.Logger().WithError(shutdownErr).Warn("error shutting down VirtiofsDaemon")
		}
		return fmt.Errorf("failed to launch cloud-hypervisor: %q", err)
	}
	clh.state.PID = pid

	if err := clh.bootVM(ctx); err != nil {
		return err
	}

	clh.state.state = clhReady
	return nil
```
 https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/clh.go#L493-L507

PR #3823, opened 2022-03-03
clh: stop virtofsd if clh fails to boot up the vm

fidencio:
If, for some reason, we're able to launch cloud hypervisor but not able
to boot the VM up, the virtiofsd process would be left behind.

Let's ensure, via defer, that we stop virtiofsd in case of errors.

Fixes: #3819


fidencio:
@egernst, sorry for linking you as a reviewer without previous notice, but I vaguely remember you solving a very similar issue on the QEMU side.

Jakob-Naucke (approved):

devimc (approved):

PR #3836, opened 2022-03-05
Enhancement: fix comments/logs and delete not used function

liubin:
- rustjail: delete function signal in BaseContainer
    
> Function signal in BaseContainer is not used anymore.

- agent: delete meaningless FIXME comment
    
> The test has passed, the FIX comment should
be deleted.

- runtime: fix invalid comments for pkg/resourcecontrol
    
> Some comments are copied and not adjusted to the
pkg/resourcecontrol package.


egernst (approved):

fidencio (approved):
lgtm, thanks @liubin!

PR #3838, opened 2022-03-06
agent: add tests for get_memory_info function

### LINKED ISSUE #3837: agent tests: Add tests for get_memory_info function in src/rpc.rs
Add tests for get_memory_info function in the `src/rpc.rs` file in the agent. Related to #292.

bradenrayhorn:
Add test coverage for get_memory_info function in src/rpc.rs

Fixes #3837

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -85,6 +85,8 @@ use std::path::PathBuf;
 const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
 
+const ERR_INVALID_BLOCK_SIZE: &str = "Invalid block size";
+
 // Convenience macro to obtain the scope logger
 macro_rules! sl {
     () => {
@@ -1154,7 +1156,12 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
         info!(sl!(), "get guest details!");
         let mut resp = GuestDetailsResponse::new();
         // to get memory block size
-        match get_memory_info(req.mem_block_size, req.mem_hotplug_probe) {
+        match get_memory_info(
+            req.mem_block_size,
+            req.mem_hotplug_probe,
+            SYSFS_MEMORY_BLOCK_SIZE_PATH,
+            SYSFS_MEMORY_HOTPLUG_PROBE_PATH,
+        ) {
             Ok((u, v)) => {
                 resp.mem_block_size_bytes = u;
                 resp.support_mem_hotplug_probe = v;
@@ -1343,24 +1350,29 @@ impl protocols::health_ttrpc::Health for HealthService {
     }
 }
 
-fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
+fn get_memory_info(
+    block_size: bool,
+    hotplug: bool,
+    block_size_path: &str,
+    hotplug_probe_path: &str,
+) -> Result<(u64, bool)> {
     let mut size: u64 = 0;
     let mut plug: bool = false;
     if block_size {
-        match fs::read_to_string(SYSFS_MEMORY_BLOCK_SIZE_PATH) {
+        match fs::read_to_string(block_size_path) {
             Ok(v) => {
                 if v.is_empty() {
-                    info!(sl!(), "string in empty???");
-                    return Err(anyhow!("Invalid block size"));
+                    warn!(sl!(), "file {} is empty", block_size_path);
+                    return Err(anyhow!(ERR_INVALID_BLOCK_SIZE));
                 }
 
                 size = u64::from_str_radix(v.trim(), 16).map_err(|_| {
                     warn!(sl!(), "failed to parse the str {} to hex", size);
-                    anyhow!("Invalid block size")
+                    anyhow!(ERR_INVALID_BLOCK_SIZE)
                 })?;
             }
             Err(e) => {
-                info!(sl!(), "memory block size error: {:?}", e.kind());
+                warn!(sl!(), "memory block size error: {:?}", e.kind());
                 if e.kind() != std::io::ErrorKind::NotFound {
                     return Err(anyhow!(e));
                 }
@@ -1369,10 +1381,10 @@ fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
     }
 
     if hotplug {
-        match stat::stat(SYSFS_MEMORY_HOTPLUG_PROBE_PATH) {
+        match stat::stat(hotplug_probe_path) {
             Ok(_) => plug = true,
             Err(e) => {
-                info!(sl!(), "hotplug memory error: {:?}", e);
+                warn!(sl!(), "hotplug memory error: {:?}", e);
                 match e {
                     nix::Error::ENOENT => plug = false,
                     _ => return Err(anyhow!(e)),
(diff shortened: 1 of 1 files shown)


src/agent/src/rpc.rs
```diff
@@ -1343,11 +1348,16 @@ impl protocols::health_ttrpc::Health for HealthService {
     }
 }
 
-fn get_memory_info(block_size: bool, hotplug: bool) -> Result<(u64, bool)> {
+fn get_memory_info(
+    block_size: bool,
+    hotplug: bool,
+    block_size_path: &str,
+    hotplug_probe_path: &str,
```
fidencio, line 1355: Those were introduced in order to help testing this function down below.

src/agent/src/rpc.rs
```diff
@@ -1906,6 +1917,91 @@ mod tests {
...
+                result: (0, false),
+                error_contains: "Invalid block size",
+            },
+            TestData {
+                block_size_data: Some("    "),
+                hotplug_probe_data: None,
+                result: (0, false),
+                error_contains: "Invalid block size",
+            },
+        ];
```
fidencio, line 1969: It seems we're not exercising the `Err(e)` path of the function.

Maybe creating or not the block_size_path / hotplug_probe_path as part of the TestData would be an way to easily be able to check the rainy path as well.

Also, are we fully exercising the hotplug path? It seems we're focusing more on the block_size part of the function.

fidencio (commented):
@bradenrayhorn,

Thanks for the PR!
I've raised a few questions / comments in the code.

src/agent/src/rpc.rs
```diff
@@ -1906,6 +1917,91 @@ mod tests {
...
+
+            let dir = tempdir().expect("failed to make tempdir");
+            let block_size_path = dir.path().join("block_size_bytes");
+            let hotplug_probe_path = dir.path().join("probe");
+
+            if let Some(block_size_data) = d.block_size_data {
+                fs::write(&block_size_path, block_size_data).unwrap();
+            }
+            if let Some(hotplug_probe_data) = d.hotplug_probe_data {
+                fs::write(&hotplug_probe_path, hotplug_probe_data).unwrap();
```
liubin, line 1982: Nit: `hotplug_probe_path` is only checked for existence, so `hotplug_probe_data` can be a bool type, and if true then touch an empty file.

src/agent/src/rpc.rs
```diff
@@ -1906,6 +1917,91 @@ mod tests {
...
+            }
+
+            let result = get_memory_info(
+                true,
```
liubin, line 1986: Should add some cases for `false` of the first two parameters?

src/agent/src/rpc.rs
```diff
@@ -1906,6 +1917,91 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_get_memory_info() {
+        #[derive(Debug)]
+        struct TestData<'a> {
+            // if None is provided, no files will be generated, else the data in the Option will populate the file
+            block_size_data: Option<&'a str>,
+            hotplug_probe_data: Option<&'a str>,
+
+            result: (u64, bool),
+            error_contains: &'a str,
```
jodh-intel, line 1929: You could combine these into a `Result` as that would match the type returned by the function being tested.

```suggestion
            result: Result<(u64, bool)>,
```

That would allow you to create tests with values like:

```rust
            result: Ok((1234, true)),
            result: Ok((0, false)),
            result: Err(anyhow!("foo bar")),
```

For a few other examples simiilar to this, see:

- https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/config.rs#L1228
- https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/config.rs#L970

src/agent/src/rpc.rs
```diff
@@ -1906,6 +1917,91 @@ mod tests {
...
+                error_contains: "",
+            },
+            TestData {
+                block_size_data: None,
+                hotplug_probe_data: None,
+                result: (0, false),
+                error_contains: "",
+            },
+            TestData {
+                block_size_data: Some(""),
```
jodh-intel, line 1952: Could you also add a `TestData` for non numeric data? Something like `Some("foo bar")` to ensure the error path is exercised in that scenario? Yes, it should be "impossible" for scenarios like this to happen, but it never hurts to be paranoid and these sorts of tests are quick to add once you're using the `TestData` approach ;)

src/agent/src/rpc.rs
```diff
@@ -1906,6 +1917,91 @@ mod tests {
...
+                hotplug_probe_data: None,
+                result: (0, false),
+                error_contains: "Invalid block size",
+            },
+            TestData {
+                block_size_data: Some("-1"),
+                hotplug_probe_data: None,
+                result: (0, false),
+                error_contains: "Invalid block size",
```
jodh-intel, line 1961: Looking at `get_memory_info()`, I see that this string is hard-coded twice, so I'd consider creating a `const` for it. You could then check for:

```rust
const ERR_INVALID_BLOCK_SIZE: &str = "Invalid block size";
```

Your test could then check for:

```rust
result: Err(anyhow!(ERR_INVALID_BLOCK_SIZE)),
```

src/agent/src/rpc.rs
```diff
@@ -1343,11 +1348,16 @@ impl protocols::health_ttrpc::Health for HealthService {
...
+    hotplug_probe_path: &str,
+) -> Result<(u64, bool)> {
     let mut size: u64 = 0;
     let mut plug: bool = false;
     if block_size {
-        match fs::read_to_string(SYSFS_MEMORY_BLOCK_SIZE_PATH) {
+        match fs::read_to_string(block_size_path) {
             Ok(v) => {
                 if v.is_empty() {
                     info!(sl!(), "string in empty???");
```
jodh-intel, line 1363: This looks like a typo so you could consider creating an initial commit which fixes this to something like the following (we don't need punctuation in error messages so periods, exclamation marks and question marks can -- and ideally should -- be removed imho):

```suggestion
                    info!(sl!(), "string is empty");
```

jodh-intel: One other point about the existing function: along with returning the errors, it also makes a number of log calls. However, two different log levels are being used `info!()` and `warn!()`. For consistency, they should all be `warn!()` log calls or possibly all `debug!()` calls since the scenarios are handled by returning errors.

jodh-intel (commented):
Thanks for raising @bradenrayhorn ! A few comments...

src/agent/src/rpc.rs
bradenrayhorn: > Maybe creating or not the block_size_path / hotplug_probe_path as part of the TestData would be an way to easily be able to check the rainy path as well.

I might be misunderstanding, but currently both `block_size_data` and `hotplug_probe_data` are of the type `Option`. In the `TestData` struct, if they are set to `None`, the file is not created. Is that what you were looking for?

> Also, are we fully exercising the hotplug path? It seems we're focusing more on the block_size part of the function.

The hotplug path "file not existing" path is being handled, but the path not currently being handled in both the block size and hotplug paths is if an error other than "does not exist" occurs when reading the file. The best idea I had to test that was changing permissions of the file to deny reading, but I wasn't sure if that was a good idea.

src/agent/src/rpc.rs
bradenrayhorn: One question with this, I like the `assert_result` macro, but it is currently defined in `config.rs`. I have copied it to `rpc.rs`, but is there a place for functions/macros shared between tests in different files? I looked around a bit but didn't see any.

liubin (approved):

src/agent/src/rpc.rs
```diff
@@ -1803,8 +1815,35 @@ mod tests {
...
     use oci::{Hook, Hooks};
+    use tempfile::tempdir;
     use ttrpc::{r#async::TtrpcContext, MessageHeader};
 
+    // Parameters:
+    //
+    // 1: expected Result
+    // 2: actual Result
+    // 3: string used to identify the test on error
+    macro_rules! assert_result {
```
jodh-intel, line 1826: Ideally we'd create a local crate to avoid the duplication (redefining this macro), but a quick look shows that might be harder than it appears, so I think this is ok for now, particularly given this is in the test code.

jodh-intel (approved):
Thanks @bradenrayhorn - great piece of work!

lgtm

src/agent/src/rpc.rs
```diff
@@ -1343,24 +1350,29 @@ impl protocols::health_ttrpc::Health for HealthService {
...
     let mut size: u64 = 0;
     let mut plug: bool = false;
     if block_size {
-        match fs::read_to_string(SYSFS_MEMORY_BLOCK_SIZE_PATH) {
+        match fs::read_to_string(block_size_path) {
             Ok(v) => {
                 if v.is_empty() {
-                    info!(sl!(), "string in empty???");
-                    return Err(anyhow!("Invalid block size"));
+                    warn!(sl!(), "string is empty");
```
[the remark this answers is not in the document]
bradenrayhorn: Updated, good suggestion.

fidencio (approved):

Issue #3847, opened 2022-03-07
tools: release: Do not consider release candidates as stable releases


fidencio:
During the release of 2.4.0-rc0 @egernst noticed an incositency in the
way we handle release tags, as release candidates are being taken as
"stable" releases, while both the kata-deploy tests and the release
action consider this as "latest".

Ideally we should have our own tag for "release candidate", but that's
something that could and should be discussed more extensively outside of
the scope of this quick fix.

For now, let's align the code generating the PR for bumping the release
with what we already do as part of the release action and kata-deploy
test, and tag "-rc"  as latest, regardless of which branch it's coming
from.

PR #3848, opened 2022-03-07
tools: release: Do not consider release candidates as stable releases

fidencio:
During the release of 2.4.0-rc0 @egernst noticed an incositency in the
way we handle release tags, as release candidates are being taken as
"stable" releases, while both the kata-deploy tests and the release
action consider this as "latest".

Ideally we should have our own tag for "release candidate", but that's
something that could and should be discussed more extensively outside of
the scope of this quick fix.

For now, let's align the code generating the PR for bumping the release
with what we already do as part of the release action and kata-deploy
test, and tag "-rc"  as latest, regardless of which branch it's coming
from.

Fixes: #3847

### DIFF
--- a/tools/packaging/release/update-repository-version.sh
+++ b/tools/packaging/release/update-repository-version.sh
@@ -68,7 +68,6 @@ generate_kata_deploy_commit() {
 kata-deploy files must be adapted to a new release.  The cases where it
 happens are when the release goes from -> to:
 * main -> stable:
-  * kata-deploy / kata-cleanup: change from \"latest\" to \"rc0\"
   * kata-deploy-stable / kata-cleanup-stable: are removed
 
 * stable -> stable:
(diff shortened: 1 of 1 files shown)


egernst (approved):

tools/packaging/release/update-repository-version.sh
```diff
@@ -183,21 +182,30 @@ bump_repo() {
...
+				## to "rcX+1", which should still be tagged as latest.
+				version_to_replace="latest"
+				replacement="latest"
+			fi
 		fi
 
-		if [ "${version_to_replace}" != "${replacement}" ]; then
-			## this covers case 2) and 3), as on both of them we have changes on kata-deploy / kata-cleanup  files
+		if [ "${version_to_replace}" != "${replacement}" ] || [ "${removed_files}" == "true" ]; then
+			## this covers case 3), as it has changes on kata-deploy / kata-cleanup  files
```
egernst, line 208: the sed below needs fixing?

egernst (changes requested):

tools/packaging/release/update-repository-version.sh
fidencio: Fixed!

jodh-intel (approved):
Thanks @fidencio.

lgtm

egernst (approved):

Issue #3874, opened 2022-03-11
Shim and QEMU leak due to ESRCH error


fengwang666:
During the process of stopping a Pod sandbox, the CRI sends the containerd-shim a [KillRequest(https://github.com/containerd/containerd/blob/8b9498909dba6dda44fda26d95ef7dbcd1765995/pkg/cri/server/container_stop.go#L174) for each container. In the case of Kata, this eventually translates into a [signal::kill](https://github.com/kata-containers/kata-containers/blob/c088a3f3adcc4b8ff1134bed9335aca31a3d131e/src/agent/src/rpc.rs#L419) call inside the kata-agent to stop the container processes. However, [kill](https://pubs.opengroup.org/onlinepubs/9699919799/functions/kill.html) call can return `ESRCH` in certain situations. In the case runc v2, the `ESRCH` is [translated](https://github.com/containerd/containerd/blob/77d53d2d230c3bcd3f02e6f493019a72905c875b/pkg/process/utils.go#L135-L139) to ErrNotFound and gracefully [handled](https://github.com/containerd/containerd/blob/77d53d2d230c3bcd3f02e6f493019a72905c875b/pkg/cri/server/container_stop.go#L174) by the cri server. But Kata doesn't translate this error right now and can cause shim to leak (failure to stop).

egernst:
Couple of questions:
 - How often do you see this failure?
 - Is there a particular workload that increases the failure rate?
 
 I'd like to be able to reproduce this to help verify.

fengwang666:
Discussed with @egernst offline and we think the more correct fix is to have the containerd cri server handles the ESRCH error directly. But meanwhile, we also should unblock ourselves. Opened an issue in containerd: https://github.com/containerd/containerd/issues/6665

fengwang666:
> Couple of questions:
> 
> * How often do you see this failure?
> * Is there a particular workload that increases the failure rate?

We saw it quite often in our pre-prod environment. But we haven't been able to figure out what exactly triggers this ESRCH error.

PR #3875, opened 2022-03-11
runtime: properly handle ESRCH error when signaling container

fengwang666:
Currently kata shim v2 doesn't translate ESRCH signal, causing container
fail to stop and shim leak.

Fixes: #3874

Signed-off-by: <feng.wang@databricks.com>

### DIFF
--- a/src/runtime/virtcontainers/utils/utils.go
+++ b/src/runtime/virtcontainers/utils/utils.go
@@ -321,6 +321,7 @@ func WaitLocalProcess(pid int, timeoutSecs uint, initialSignal syscall.Signal, l
 	if initialSignal != syscall.Signal(0) {
 		if err = syscall.Kill(pid, initialSignal); err != nil {
 			if err == syscall.ESRCH {
+				logger.WithField("pid", pid).Warnf("kill encounters ESRCH, process already finished")
 				return nil
 			}
 
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -1060,7 +1061,18 @@ func (c *Container) signalProcess(ctx context.Context, processID string, signal 
 		return fmt.Errorf("Container not ready, running or paused, impossible to signal the container")
 	}
 
-	return c.sandbox.agent.signalProcess(ctx, c, processID, signal, all)
+	// kill(2) method can return ESRCH in certain cases, which is not handled by containerd cri server in container_stop.go.
+	// CRIO server also doesn't handle ESRCH. So kata runtime will swallow it here.
+	var err error
+	if err = c.sandbox.agent.signalProcess(ctx, c, processID, signal, all); err != nil &&
+		strings.Contains(err.Error(), "ESRCH: No such process") {
+		c.Logger().WithFields(logrus.Fields{
+			"container":  c.id,
+			"process-id": processID,
+		}).Warn("signal encounters ESRCH, process already finished")
+		return nil
+	}
+	return err
 }
 
 func (c *Container) winsizeProcess(ctx context.Context, processID string, height, width uint32) error {


src/runtime/pkg/containerd-shim-v2/service.go
```diff
@@ -794,7 +795,18 @@ func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.E
 		return empty, nil
 	}
 
-	return empty, s.sandbox.SignalProcess(spanCtx, c.id, processID, signum, r.All)
+	// kill(2) method can return ESRCH in certain cases, which is not handled by containerd cri server in container_stop.go.
+	// runc v2 translates ESRCH to ErrNotFound, so we'll handle similarly in Kata.
+	if err = s.sandbox.SignalProcess(spanCtx, c.id, processID, signum, r.All); err != nil &&
+		strings.Contains(err.Error(), "ESRCH: No such process") {
```
egernst, line 801: Should this just be:
```
err == unix.ESRCH 
```

egernst (commented):
I think we'll def need a patch to fix this up here while we look at perhaps addressing at CRI-containerd.

I think we can simplify the err check? I'm curious how this is handled in the CRIO case.

src/runtime/pkg/containerd-shim-v2/service.go
fengwang666: @egernst I'm not sure if the err type is preserved across the grpc call between shim and kata agent. If that's the case, I'm happy to update it.

src/runtime/pkg/containerd-shim-v2/service.go
```diff
@@ -794,7 +795,18 @@ func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.E
...
+		shimLog.WithFields(logrus.Fields{
+			"sandbox":   s.sandbox.ID(),
+			"container": c.id,
+			"exec-id":   r.ExecID,
+		}).Warn("signal encounters ESRCH, process already finished")
+		return empty, errors.Wrapf(errdefs.ErrNotFound, "process already finished")
```
egernst, line 807: So, AFAICT CRIO doesn't know what to do with ErrNotFound, so while this'll make containerd happy, it'll break CRIO.

Perhaps the workaround at this point is to instead "swallow" the error here. Just print a warning (what you have there), and return empty, nil?

egernst (approved):
Thanks -- LGTM.

src/runtime/pkg/containerd-shim-v2/service.go
fidencio: Something similar to this has already been done on `WaitLocalProcess()`:
https://github.com/kata-containers/kata-containers/blob/358081c4ae1c9c8f03fb6520411f41ba3eda5f1f/src/runtime/virtcontainers/utils/utils.go#L323-L325

So, based on that, I think this approach is the way to go.
However, I wouldn't even warn here (or I would also warn there).  If this is a no-error, this is a no-error and users shouldn't be notified about this.

Apart from this, I do second Eric's opinion to at least check if we can use unix.ESRCH / syscall.ESRCH, we should be able to do that rather than matching strings.

fidencio (changes requested):
@fengwang666, please, see the comments inline.

fidencio:
> I think we can simplify the err check? I'm curious how this is handled in the CRIO case.

Right now it's not handled there, and I don't think we have hit this before (or at least I don't recall it being reported there before). 

I've pointed out to another place where we deal with ESRCH in a way that makes me think we're working that around to passing that up to the container engine.  Not sure if this is special enough to VM runtime types that doesn't make sense to be treated in the upper layer.

src/runtime/pkg/containerd-shim-v2/service.go
bergwolf: @fidencio `WaitLocalProcess` is to handle process exits on the host side, whereas this is trying to fix the race window that a container/process exits before a signal is processed by the agent.

bergwolf: @fengwang666 While the race window is real, could you fold the change in `virtcontainers/container.go:signalProcess()` so that it works for all agent-handled signals?

fengwang666: @bergwolf I like your suggestion. Updated.
@fidencio I personally think it's beneficial to log a warning on `ESRCH` error. It might indicate some other issues if it's happening more often than it should. So I added a log in the other place as well. Regarding checking the error with `unix.ESRCH / syscall.ESRCH`, I don't think ttrpc translates the rust ESRCH to golang ESRCH. Otherwise the error message we see in the log for [ESRCH](https://github.com/golang/go/blob/master/src/syscall/syscall_unix.go#L124) should be `errno 3`.

bergwolf (approved):
Thanks!

src/runtime/pkg/containerd-shim-v2/service.go
fidencio: > @fidencio WaitLocalProcess is to handle process exits on the host side, whereas this is trying to fix the race window that a container/process exits before a signal is processed by the agent.

Yet another reason to avoid pushing for reviews over the weekend. :-)

fidencio: > @fidencio I personally think it's beneficial to log a warning on ESRCH error. It might indicate some other issues if it's happening more often than it should.

The part that bothers me is we're working something around that should "just work" and then raising a red herring to folks debugging some problem that may or may not be related to this.
I sincerely don't think we should log a warn there, but you already have 2 approvals so I won't block it.

> Regarding checking the error with unix.[ESRCH](https://github.com/golang/go/blob/master/src/syscall/syscall_unix.go#L124) / syscall.ESRCH, I don't think ttrpc translates the rust ESRCH to golang ESRCH. Otherwise the error message we see in the log for ESRCH should be errno 3.

Fair. :-)

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -321,6 +321,7 @@ func WaitLocalProcess(pid int, timeoutSecs uint, initialSignal syscall.Signal, l
 	if initialSignal != syscall.Signal(0) {
 		if err = syscall.Kill(pid, initialSignal); err != nil {
 			if err == syscall.ESRCH {
+				logger.WithField("pid", pid).Warnf("kill encounters ESRCH, process probably already finished")
```
fidencio, line 324:
```suggestion
				logger.WithField("pid", pid).Warnf("kill encounters ESRCH, process already finished")
```

fidencio: @fengwang666, please, just change the warn message, removing the "probably" from there.

egernst (commented):
Looks good modulo the message cleanup suggested by Fabiano.

Thanks again Feng.

src/runtime/virtcontainers/utils/utils.go
fengwang666: done!

fidencio (approved):
lgtm, thanks @fengwang666!

fengwang666:
backport PR: https://github.com/kata-containers/kata-containers/pull/3892

fengwang666:
backport PR: https://github.com/kata-containers/kata-containers/pull/3893

Issue #3888, opened 2022-03-14
Incorrect statfs types for s390x


Jakob-Naucke:
```
Compiling kata-sys-util v0.1.0 (/home/ubuntu/go/src/github.com/kata-containers/kata-containers/src/libs/kata-sys-util)
error[E0308]: mismatched types
  --> /home/ubuntu/go/src/github.com/kata-containers/kata-containers/src/libs/kata-sys-util/src/fs.rs:46:38
   |
46 |         if st.filesystem_type().0 == FUSE_SUPER_MAGIC {
   |                                      ^^^^^^^^^^^^^^^^ expected `u32`, found `i64`
```

cf. https://github.com/rust-lang/libc/pull/1999

Jakob-Naucke:
linkage!

PR #3786, opened 2022-02-28
runtime-rs: shim part

quanweiZhou:
initial commit for runtime of rust version
shim crate implements containerd runtime v2 API

Fixes: #3785

PR #3889, opened 2022-03-14
runtime-rs: uint FsType for s390x

Jakob-Naucke:
statfs type on s390x should be c_uint, not __fsword_t

Fixes: #3888

Fix from #3786 -- sorry, I thought I had tested it…


bergwolf (approved):
Thanks @Jakob-Naucke ! I wonder is there any way we can (at least) build test `runtime-rs` for s390x? How is `s390` tested for the `main` branch?

liubin (approved):

Jakob-Naucke:
@bergwolf thanks for the review. The `main`/`stable` s390x CI is hopefully about to be re-enabled. It is available over Jenkins, so that would be some jobs to create. Also, I guess if we are talking about s390x, we should also talk about ARM (could be used via GHA AFAIU?) and ppc64le (Jenkins too). (I cannot guarantee I'll be personally able to look into this due to my pending leave of absence.)

bergwolf:
@Jakob-Naucke Thanks for the info! Yes, indeed, we should look at different architectures for the runtime-rs branch. Maybe at a later point when the code base is more available.

Issue #3898, opened 2022-03-15
CCv0: Refactor ccv0.sh to use integration test lib.sh


Megan-Wright:
Update ccv0.sh to remove duplicated code also found in lib.sh

stevenhorsman:
Closed with #3899 merging

PR #3899, opened 2022-03-15
CCv0: Refactor CCv0.sh to use integration test lib.sh

Megan-Wright:
Updates ccv0.sh to remove duplicated code
Fixes: #3898
Signed-off-by: Megan Wright <megan.wright@ibm.com>
Signed-off-by: Georgina Kinge  <georgina.kinge@ibm.com>

### DIFF
--- a/docs/how-to/ccv0.sh
+++ b/docs/how-to/ccv0.sh
@@ -51,6 +51,12 @@ export CONTAINER_ID="${CONTAINER_ID:-0123456789}"
 source /etc/os-release || source /usr/lib/os-release
 grep -Eq "\<fedora\>" /etc/os-release 2> /dev/null && export USE_PODMAN=true
 
+export BATS_TEST_DIRNAME="${tests_repo_dir}/integration/containerd/confidential"
+# If we've already checked out the test repo then source the confidential scripts
+[ -d "${BATS_TEST_DIRNAME}" ] && source "${BATS_TEST_DIRNAME}/lib.sh"
+
+export RUNTIME_CONFIG_PATH=/etc/kata-containers/configuration.toml
+
 debug_output() {
     if [ -n "${DEBUG}" ]
     then
@@ -70,7 +76,7 @@ debug_function() {
 
 usage() {
     exit_code="$1"
-    cat <<EOT
+    cat <<EOF
 Overview:
     Build and test kata containers from source
     Optionally set kata-containers and tests repo and branch as exported variables before running
@@ -110,7 +116,7 @@ Commands:
 Options:
     -d: Enable debug
     -h: Display this help
-EOT
+EOF
     # if script sourced don't exit as this will exit the main shell, just return instead
     [[ $_ != $0 ]] && return "$exit_code" || exit "$exit_code"
 }
@@ -175,6 +181,9 @@ check_out_repos() {
         git checkout ${tests_branch}
     fi
     git reset --hard origin/${tests_branch}
+
+    source "${BATS_TEST_DIRNAME}/lib.sh"
+
     popd
 
     echo "Creating repo: ${katacontainers_repo} and branch ${katacontainers_branch} into ${katacontainers_repo_dir}..."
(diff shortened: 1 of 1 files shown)


docs/how-to/ccv0.sh
```diff
@@ -51,6 +51,10 @@ export CONTAINER_ID="${CONTAINER_ID:-0123456789}"
 source /etc/os-release || source /usr/lib/os-release
 grep -Eq "\<fedora\>" /etc/os-release 2> /dev/null && export USE_PODMAN=true
 
+if [ -d ${tests_repo_dir} ]; then
```
wainersm, line 54: I'm not sure this checking is needed. Before some variables that holds file/directory path aren't checked. Is there anything special with `BATS_TEST_DIRNAME` that requires this checking?

docs/how-to/ccv0.sh
```diff
@@ -51,6 +51,10 @@ export CONTAINER_ID="${CONTAINER_ID:-0123456789}"
...
+    export BATS_TEST_DIRNAME="${tests_repo_dir}/integration/containerd/confidential/"
```
wainersm, line 55: The name `BATS_TESTS_DIRNAME` is a special variable of bats (see [here](https://bats-core.readthedocs.io/en/stable/writing-tests.html#special-variables)). It might silently introduce a bug If in the future `ccv0.sh` calls bats due to clash on the variable names.

docs/how-to/ccv0.sh
```diff
@@ -199,26 +206,33 @@ build_and_install_kata_runtime() {
...
     # Add cni directory to containerd config
+    sudo grep -qF "conf_dir = \"/etc/cni/net.d\"" ${containerd_config} || \
     echo "    [plugins.cri.cni]
       # conf_dir is the directory in which the admin places a CNI conf.
-      conf_dir = \"/etc/cni/net.d\"" | sudo tee -a /etc/containerd/config.toml
+      conf_dir = \"/etc/cni/net.d\"" | sudo tee -a ${containerd_config}
     
     # Switch image offload to true in kata config
-    sudo sed -i -e 's/^# *\(service_offload\).*=.*$/\1 = true/g' /etc/kata-containers/configuration.toml
+    # sudo sed -i -e 's/^# *\(service_offload\).*=.*$/\1 = true/g' /etc/kata-containers/configuration.toml
```
wainersm, line 233: Really want this line commented?

wainersm: ah, yes, later you call `switch_image_service_offload` so maybe you just need to delete that line.

docs/how-to/ccv0.sh
```diff
@@ -435,45 +449,27 @@ crictl_create_cc_pod() {
     sudo iptables -P FORWARD ACCEPT
     
     # Create crictl pod config
-cat << EOF > ${HOME}/pod-config.yaml
-metadata:
-  name: ${crictl_sandbox_name}
-EOF
+    source ${tests_repo_dir}/integration/containerd/confidential/lib.sh
```
wainersm, line 452: Likewise.

docs/how-to/ccv0.sh
```diff
@@ -435,45 +449,27 @@ crictl_create_cc_pod() {
...
 }
 
 crictl_delete_cc() {
-    local pod_id=$(sudo crictl pods --name ${crictl_sandbox_name} -q)
-    local container_id=$(sudo crictl ps --pod ${pod_id} -q)
-    if [ -n "${container_id}" ]; then
-        sudo crictl stop ${container_id} && sudo crictl rm ${container_id}
-    fi
-    sudo crictl stopp ${pod_id} && sudo crictl rmp ${pod_id}
+    source ${tests_repo_dir}/integration/containerd/confidential/lib.sh
```
wainersm, line 471: and here.

wainersm:
> @wainersm Thanks so much! I've updated almost everything you mentioned and retested to ensure the script still works as expected... the reason we set `$BATS_TEST_DIRNAME` is because it is required to already be set by `lib.sh` in the tests repo when it is called (the code where this variable gets set in testing isn't called/used by us here).

Oh, true, I forgot `lib.sh` was dependent of a `$BATS_TEST_DIRNAME`. Because I source the file from inside a bats test, the variable will be already set by the framework.

wainersm:
@Megan-Wright hello, you will need to change your commit to fix the failure on the static check.

docs/how-to/ccv0.sh
```diff
@@ -199,26 +204,32 @@ build_and_install_kata_runtime() {
...
     # Add cni directory to containerd config
+    sudo grep -qF "conf_dir = \"/etc/cni/net.d\"" ${containerd_config} || \
     echo "    [plugins.cri.cni]
       # conf_dir is the directory in which the admin places a CNI conf.
-      conf_dir = \"/etc/cni/net.d\"" | sudo tee -a /etc/containerd/config.toml
+      conf_dir = \"/etc/cni/net.d\"" | sudo tee -a ${containerd_config}
     
     # Switch image offload to true in kata config
-    sudo sed -i -e 's/^# *\(service_offload\).*=.*$/\1 = true/g' /etc/kata-containers/configuration.toml
+    source ${BATS_TEST_DIRNAME}/lib.sh
```
jodh-intel, line 231: Can you replace the 4 function-local `source` calls with a single top-level `source` maybe around line 51 where we source `/etc/os-release`?

Megan-Wright:
Hi @jodh-intel thanks so much for the review! I've addressed all your comments as far as I can - we still need to do the `source ${BATS_TEST_DIRNAME}/lib.sh` in two places as we can't source the code until it's been checked out in `check_out_repos` but I've removed the remaining source commands.

docs/how-to/ccv0.sh
```diff
@@ -175,6 +183,9 @@ check_out_repos() {
         git checkout ${tests_branch}
     fi
     git reset --hard origin/${tests_branch}
+
+    source "${BATS_TEST_DIRNAME}/lib.sh"
```
jodh-intel, line 187: I think this line can be removed?

docs/how-to/ccv0.sh
```diff
@@ -199,47 +210,38 @@ build_and_install_kata_runtime() {
...
     debug_function enable_full_debug
     # Temp PoC verify code: Inject policy path config parameter
-    sudo sed -i -e 's%^kernel_params = "\(.*\)"%kernel_params = "\1 agent.container_policy_file=/etc/containers/quay_verification/quay_policy.json"%g' /etc/kata-containers/configuration.toml
-
+    add_kernel_params
+    
     # If using AA then need to add the agent_config
     if [ "${AA_KBC}" == "offline_fs_kbc" ]; then
-        sudo sed -i -e 's%^kernel_params = "\(.*\)"%kernel_params = "\1 agent.config_file=/etc/agent-config.toml"%g' /etc/kata-containers/configuration.toml
+        add_kernel_params
```
jodh-intel, line 222: Are these `add_kernel_params()` calls missing an arg?

jodh-intel (commented):
Thanks @Megan-Wright. A few more comments...

docs/how-to/ccv0.sh
Megan-Wright: The reason we need to source `lib.sh` in this function as well as globally is because the repo is checked out in this function and the code in `lib.sh` is then referenced before exiting the function :)

docs/how-to/ccv0.sh
Megan-Wright: ah yep good spot - I completely missed that... I've updated now

docs/how-to/ccv0.sh
```diff
@@ -51,6 +51,14 @@ export CONTAINER_ID="${CONTAINER_ID:-0123456789}"
 source /etc/os-release || source /usr/lib/os-release
 grep -Eq "\<fedora\>" /etc/os-release 2> /dev/null && export USE_PODMAN=true
 
+export BATS_TEST_DIRNAME="${tests_repo_dir}/integration/containerd/confidential"
+# If we've already checked out the test repo then source the confidential scripts
+if [ -d "${BATS_TEST_DIRNAME}" ]; then
+    source "${BATS_TEST_DIRNAME}/lib.sh"
+fi
```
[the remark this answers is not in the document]
Megan-Wright: done :)

docs/how-to/ccv0.sh
stevenhorsman: Hey James, thanks so much for the reviews. As Megan says I think we still need both of the `source`s as they run under different circumstances, depending on whether the code has already been checked out or not:
- The first one: `[ -d "${BATS_TEST_DIRNAME}" ] && source "${BATS_TEST_DIRNAME}/lib.sh"` will only run if the code has already been checked out (otherwise it isn't there to source!)
- This one runs in `check_out_repos()`, which will run for a fresh environment when the script hasn't previously run, but not be run most of the time when the tests repo has already been checked out
Together they should provide coverage of the two different states. I hope this makes sense. If you have a better suggestion to solve this then that would be great though!

stevenhorsman:
It's worth noting that in order to work with `ccv0.sh` these changes require https://github.com/kata-containers/tests/pull/4591 to be merged (basically allowing a non-root user to run the integration script methods that this PR is now using), so I've added the `do-not-merge` label.

stevenhorsman:
https://github.com/kata-containers/tests/pull/4591 has merged now, so I've removed the `do-not-merge` label.

docs/how-to/ccv0.sh
```diff
@@ -430,50 +419,29 @@ kubernetes_delete_ssh_demo_pod() {
...
-
-    local pod_id=$(sudo crictl runp -r kata ${HOME}/pod-config.yaml)
-    sudo crictl pods ${pod_id}
+    crictl_delete_cc_pod_if_exists "${crictl_sandbox_name}"
+    crictl_create_cc_pod "${pod_config}"
+    sudo crictl pods
 }
 
-crictl_create_cc_container() {
+unshadow_crictl_create_cc_container() {
```
c3d, line 434: The `unshadow` prefix is confusing. I think that `unshadow` here refers to the same name in `lib.sh`, but you are not really unshadowing it, more like wrapping it to make it easier to use. It's also a bit inconsistent with the `crictl_delete_cc` case. What about a `wrap` or `call` prefix?

Megan-Wright:
@c3d Thanks for the review :) You're right it is confusing, I just couldn't think of a better way to put it! I've updated the prefixes to `call_` now

stevenhorsman:
@jodh-intel, @wainersm, @c3d - thanks for all the review comments. I think Megan has address them all now, so is there anything else you'd like to see considered/changed before we get this merged?

jodh-intel (approved):
Thanks @Megan-Wright. I personally still find the "double `source`" is a bit confusing, but other than that...

lgtm

wainersm (approved):

Issue #3904, opened 2022-03-17
Improve file installation


jodh-intel:
Improve the makefiles to:

- make them more consistent
- remove duplication
- ensure that all files are installed with explicit owner, group and permissions.

PR #3905, opened 2022-03-17
Build improve installs

jodh-intel:
build: Improve file installation

Move the `make(1)` code used to install files to `utils.mk`, removing
all duplication in the process.

> **Notes:**
>
> - This change tightens up the permissions so that all files are now
>   installed explicitly as user `root` and group `adm`. The "other"
>   permissions are zero meaning you need elevated privileges to run
>   and view the installed files.
>
> - This change also corrects a few buglets where some of the installation
>   targets were specified as a file. This isn't necessary as it's simpler
>   to allow `install(1)` to derive that from the specified source.

Summary of permission and ownership changes:

| File type    | Old metadata             | New metadata            |
|--------------|--------------------------|-------------------------|
| binaries     | `-rwxr-xr-x 1 root root` | `-rwxr-x--- 1 root adm` |
| config files | `-rw-r--r-- 1 root root` | `-rw-r----- 1 root adm` |

Fixes: #3904.


jodh-intel:
@amshinde, @fidencio, @devimc, @wainersm, @liubin, @bergwolf - ptal. This PR is a delicate one as it will affect packaging. It will potentially also affect users given the changes to the group and perms.

utils.mk
```diff
@@ -3,6 +3,27 @@
...
 #
 
+# Owner for installed files
+INSTALL_OWNER     ?= root
+
+# Group for installed files
+INSTALL_GROUP     ?= adm
+
+# Permissions for installed executable files
+INSTALL_EXE_PERMS ?= 0750
```
liubin, line 13: CI fails with:

```
20:43:23 .ci/install_kata.sh: line 56: /usr/local/bin/kata-runtime: Permission denied
```

The mod 750 means others can't execute these binaries.

utils.mk
```diff
@@ -3,6 +3,27 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+# Owner for installed files
+INSTALL_OWNER     ?= root
+
+# Group for installed files
+INSTALL_GROUP     ?= adm
```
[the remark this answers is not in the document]
jodh-intel: https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/usernames.html

It's strictly optional but afaik it is provided by the main distros. But if not, those distros could run `make INSTALL_GROUP=root` if they wished.

utils.mk
devimc: and this may clash with the rootless use case .. :thinking:  :question: 

cc @fengwang666

jodh-intel: That's true. But I think that's what we should be encouraging. The reason is that the configuration file could contain sensitive information such as [Jaeger credentials](https://github.com/kata-containers/kata-containers/commit/340650270658d2e6ae0a3e52c1865c0bb4b81acf). That implies that "other" should not be able to read the file. But when you run `kata-runtime`, the code will read the configuration file. Overall, that means that admins should not allow non-priv / "other" users to run `kata-runtime`. We could keep the group ownership as `root`, but by using `adm` admins can give that group membership to some users without giving them full `root` privs. In summary, the behaviour of this PR is that you would need to be in the `INSTALL_GROUP` group to be able to run any Kata commands.

This changes makes the install "safe by default". Again if packagers / distro wish to change that behaviour, they can run `make INSTALL_EXE_PERMS="..."` as they wish.



That's right, and I think that's actually how users should be running it.

fidencio:
> /test_kata_deploy

Let me see how kata-deploy behaves with this: https://github.com/kata-containers/kata-containers/actions/runs/2000458019

utils.mk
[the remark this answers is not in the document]
fidencio: I think the `adm` group is present in the majority of the distros, @liubin.

Not sure about darwin, though. /cc @sameo @egernst

fidencio:
> > /test_kata_deploy
> 
> Let me see how kata-deploy behaves with this: https://github.com/kata-containers/kata-containers/actions/runs/2000458019

kata-deploy seems to be very happy with these changes.

utils.mk
c3d: This means we would at least add some supplemental steps for validation, e.g. creating a `kata` group? Is that the intent?

jodh-intel: Hi @c3d - I'm not sure I follow? The intent of this PR is to improve the installation of files and make them as secure as possible by default. A `kata` group _could_ be used but I'm suggesting we use the `adm` group as the default as that should already exist. Having a `kata` group would be better imho in the longer term as that would give admins more scope and flexibility. And for some context lthough containerd doesn't provide it's own group, docker does.

An alternative to `adm` group ownership and clearing all the "other" perms and allowing non-priv users to run `kata-runtime` as they can currently could be to rework the config handling so that we store sensitive config details in a separate (more secure) file, but that's a much bigger change than this, and I'm personally not convinced about that idea currently.

c3d: Sorry @jodh-intel I had not refreshed my browser and your comment did not show up. I was replying directly to @devimc

devimc: @c3d that's a good question, we may need to add `kata-$RANDOM` user to the `adm` group, but the most important thing here is that the job that was testing rootless has been removed/disabled, **the CI is not testing rootless** :exclamation::exclamation:
 
:speaker: @fengwang666

jodh-intel:
Any thoughts on this @bergwolf, @liubin?

fidencio:
@jodh-intel, for the sake of consistency, shall we ensure that when building & shipping qemu / firecracker / firecracker's jailer / cloud-hypervisor those also follow the very same pattern for the permissions? (no, this is not a suggestion, just a question).

Currently we have, as part of the kata-static:
```
-rwxr--r--. 1 root root 7.0M Mar 21 10:22 cloud-hypervisor
-rwxr-x---. 1 root adm   48M Mar 21 10:21 containerd-shim-kata-v2
-rwxr--r--. 1 root root 3.2M Mar 21 10:20 firecracker
-rwxr--r--. 1 root root 2.5M Mar 21 10:20 jailer
-rwxr-x---. 1 root adm   17K Mar 21 10:21 kata-collect-data.sh
-rwxr-x---. 1 root adm   39M Mar 21 10:21 kata-monitor
-rwxr-x---. 1 root adm   49M Mar 21 10:21 kata-runtime
-rwxr-xr-x. 1 root root  18M Mar 21 10:27 qemu-system-x86_64
```

With this in mind, I think we'd also need to change the permissions of the `containerd-shim-kata-${hypevisor}-v2` files created by kata-deploy, and I think this change should be part of this series (this is where we could make sure the owner / group of the files are the expected ones.: https://github.com/kata-containers/kata-containers/blob/3edf25b6c998e32400d579c2e833f3b5cd6b7234/tools/packaging/kata-deploy/scripts/kata-deploy.sh#L102).

wainersm:
Hi @jodh-intel , I hope not too late...I will have time to review this today! \o/

utils.mk
fengwang666: @devimc As long as the qemu executable permission is not changed, the the rootless hypervisor feature shouldn't be impacted. The [rootless](https://github.com/kata-containers/tests/blob/main/.ci/run.sh#L68) CI seems still running?

jodh-intel:
@fidencio - I've pushed another commit for the follow-on `kata-deploy` changes (which are currently untested).

It would be great if folk could help out testing this PR as I don't think we have full CI coverage for `kata-deploy`.

Also note that I've tightened the perms on binaries further (and updated the table at the top of this PR) since binary files don't need to be owner writable.

src/agent/Makefile
```diff
@@ -146,7 +146,7 @@ run:
 install-services: $(GENERATED_FILES)
 ifeq ($(INIT),no)
 	@echo "Installing systemd unit files..."
-	$(foreach f,$(UNIT_FILES),$(call INSTALL_FILE,$f,$(UNIT_DIR)))
+	$(foreach f,$(UNIT_FILES),$(call INSTALL_FILE,$f,$(DESTDIR)/$(UNIT_DIR)))
```
wainersm: In src/agent/Makefile:
`DESTDIR :=`
In src/runtime/Makefile:
`DESTDIR ?= /`

I didn't check other Makefiles nor variables but I think we should make the default values consistent across makefiles as well. In a follow up PR?

utils.mk
```diff
@@ -160,3 +180,62 @@ standard_rust_check:
...
+        --group $(KATA_INSTALL_GROUP) \
+        -D $1 $2/$(notdir $1) || exit 1;
+endef
+
+# Install an executable file.
+#
+# params:
+# $1 : file to install
+# $2 : directory path where file will be installed
+define INSTALL_EXEC
```
wainersm, line 209: `INSTALL_EXEC` and `INSTALL_CONFIG` used to honor `DESTDIR` so that the caller didn't need to explicitly prefix in the target path parameter. I'm wondering what's the rationale behind that change of behavior.

Implicitly prefixing `DESTDIR` makes it less flexible, however more robust as in the future a new caller might forget to consider `DESTDIR` .

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -34,6 +34,10 @@ workdir="${WORKDIR:-$PWD}"
 
 destdir="${workdir}/kata-static"
 
+KATA_INSTALL_OWNER="${KATA_INSTALL_OWNER:-root}"
```
wainersm, line 37: Makes me sad see that block of initialization repeated in many places. is it a bad idea to move them all into a `common.bash` file in the root of the project?

wainersm:
Hi @jodh-intel , I reviewed the PR fully and left some comments. I will be testing it doesn't break the scripts I have for OpenShift CI.

utils.mk
devimc: @fengwang666 

> The [rootless](https://github.com/kata-containers/tests/blob/main/.ci/run.sh#L68) CI seems still running?

no, there is no job exporting `CRIO_K8S` variable, that chunk of code is useless https://github.com/kata-containers/tests/blob/main/.ci/run.sh#L64-L70

wainersm: Hi @devimc ,

It should be running on the baseline jobs (run daily). Example:
http://jenkins.katacontainers.io/view/Daily%20baseline/job/kata-containers-main-fedora-35-x86_64-CRIO_K8S-baseline/

And in jobs triggered by pull requests. Example:
http://jenkins.katacontainers.io/view/PR%20Triggered/job/kata-containers-fedora-35-x86_64-CRIO_K8S-PR/

wainersm:
with this PR:
```
[wmoschet@wainer-laptop kata-containers]$ make kata-tarball
tools/packaging/kata-deploy/local-build/Makefile:30: ../../../../utils.mk: No such file or directory
make: *** No rule to make target '../../../../utils.mk'.  Stop.

```

wainersm:
Great, this PR doesn't break the workflow that I have for OpenShift CI :)

jodh-intel:
Thanks for testing @wainersm. I plan to update the PR further based on feedback tomorrow...

jodh-intel:
@wainersm - thanks for reviewing. I've updated the branch based on your feedback and found a cleaner way of passing the installation variables to `kata-deploy`.

@wainersm, @fidencio - Could you re-review and try re-testing this branch?

jodh-intel:
@wainersm and @fidencio  - do you have any cycles to help re-test the kata-deploy changes? If this is going to be a problem, I'll splint this into 2 PRs: my originally focussed changes and a separate one for the kata-deploy specific bits @fidencio asked for.

fidencio:
> /test_kata_deploy

Not passing: https://github.com/kata-containers/kata-containers/runs/5978347167?check_suite_focus=true

fidencio:
> @wainersm and @fidencio - do you have any cycles to help re-test the kata-deploy changes? If this is going to be a problem, I'll splint this into 2 PRs: my originally focussed changes and a separate one for the kata-deploy specific bits @fidencio asked for.

Please, split this into 2 PRs @jodh-intel.  Let's get the first bit in, considering it passes all the tests, and spend time on the binaries that are not directly coming from our GitHub on a second pass.

By the way, sorry for not reviewing this one earlier, but I thought you were still working on it as the "kata deploy build" tests were not passing.

jodh-intel:
@fidencio, @wainersm - ok, I've now reduced the scope of this PR back to the main components only.

Once this lands, I'll raise a follow-on for the kata-deploy changes...

fidencio:
static-check failures:
```
    rpc::tests::test_update_interface
    sandbox::tests::test_find_container_process
```

And snap CI is also failing.

arronwy:

jodh-intel:
@fidencio, @wainersm - It would be geat to get this landed, so please could you tal?

fidencio:
> @fidencio, @wainersm - It would be geat to get this landed, so please could you tal?

You have my LGTM, considering the tests are green.

jodh-intel:
Hmm, it seems that snap is failing because it's very / overly opinionated:

```
container.go:177: in snap "kata-containers": "usr/bin/containerd-shim-kata-v2" should be world-readable and executable, and isn't: -r-xr-x---
container.go:177: in snap "kata-containers": "usr/bin/kata-collect-data.sh" should be world-readable and executable, and isn't: -r-xr-x---
container.go:177: in snap "kata-containers": "usr/bin/kata-runtime" should be world-readable and executable, and isn't: -r-xr-x---
error: cannot pack "/home/runner/work/kata-containers/kata-containers/prime": snap is unusable due to bad permissions
```

I've tweaked the binary perms to set them for "other" (but not for config files) to see if that makes snap any happier...

c3d:
I ran with my local CI script, and I got
```
ERROR: Found unexpected /usr/bin/qemu-system-x86_64 present
```

I don't think this is related to this commit. Need to run again on `main`.

c3d:
> I ran with my local CI script, and I got
> 
> ```
> ERROR: Found unexpected /usr/bin/qemu-system-x86_64 present
> ```
> 
> I don't think this is related to this commit. Need to run again on `main`.

OK; seen only with Alma Linux (twice). Works with Fedora. Let's ignore that for the moment.

utils.mk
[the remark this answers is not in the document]
amshinde: @why not just keep the group as root? Whats the benefit of having admin here?

utils.mk
amshinde: Just to make it clear, there are two scenarios for rootless - rootless hypervisor and rootless runtime itself when the upper layers such as containerd/podman itself call kata as a non-priveleged user.
I am afraid that this PR will break the later case, as we restrict kata-runtime to be called by restricted user group. 
We do not have control over what user calls kata-runtime in that case.

My suggestion would be to have kata-runtime as world-executable and the kata conf file as world-readable. I believe any sensitive information should not be part of config files itself.  I know this exposes any sensitive information present in the kata-configuration file, but I don't think they should have been there in any place. We could have another file to hold any sensitive data and that could be restricted to be read on by admin/root group.

cc @jodh-intel

src/agent/Makefile
```diff
@@ -45,12 +45,11 @@ ifneq ($(EXTRA_RUSTFEATURES),)
     override EXTRA_RUSTFEATURES := --features "$(EXTRA_RUSTFEATURES)"
 endif
 
+##VAR DESTDIR=<path> is a directory prepended to each installed target file
```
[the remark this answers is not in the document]
jodh-intel: @amshinde - fixed.

utils.mk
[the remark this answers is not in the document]
jodh-intel: We could make it `root` but I chose `adm` to give admins a change to add select users to that group (rather than the `root` group) in recognition that this PR _may_ cause issues for some as currently any user can run the binaries, for instance.

utils.mk
jodh-intel: @amshinde - I'm a little confused. This PR is trying to resolve a permissions issue to ensure that if the admin enters sensitive information in the configuration file the effect is limited since only privileged users would be able to access that configuration file. Are you proposing that we just close this PR and come up with an alternative mechanism for storing the sensitive details? If so, _maybe_ https://github.com/kata-containers/kata-containers/pull/4109 could be used/abused for this purpose since you could then store just the sensitive info in a `secrets.toml` or similar. However, if a non-priv user runs `kata-runtime` currently, the (entire) configuration needs to be available to that user. This PR resolves that issue by requiring the user running the command to be privileged, but if I'm understanding what you're saying, we'd need to make the fragments PR "tolerate" unreadable files by just ignoring them, to handle the scenario where a system has sensitive information stored in a config fragment.

src/agent/Makefile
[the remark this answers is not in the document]
jodh-intel: Fixed.

utils.mk
jodh-intel: > The question is whether the resulting semantics aren't a bit too subtle and whether it wouldn't invite unexpected problems 

Agreed. I really don't think that random thought is a good idea, it was more to try to understand @amshinde's comments.

> handling ordinary and sensitive data seems just fundamentally different. 

Again, I agree. I think it was a mistake to allow sensitive data to be stored in our config file. However, we do allow it so this PR was the most immediate solution to the problem using simple primitives. But that doesn't mean we cannot resolve this in other more secure ways in future given additional engineering effort.

@amshinde - if you wish to look into an alternative scheme, that's perfectly fine by me. However, it is worth pointing out that _by default_ neither this PR nor a more elaborate scheme to store secrets is necessary since by default no secrets are stored in the configuration file. It's also worth stressing that if we allow non-priv users to continue to run the `kata-runtime` binary, we're going to hit complications if we store the secrets in a different location or use a different method since (atleast currently) the entire configuration needs to be read by the runtime at startup. I guess we could consider having `kata-runtime` and `containerd-shim-kata-v2` special-case reading secrets based on the uid of the caller ("If running non-priv don't load secrets, else do"). However special casing like this isn't ideal imho and would need careful testing.

jodh-intel:
Let's spin the wheels again...

/test

jodh-intel:
The CI is green so what do we think about the approach folks?

utils.mk
```diff
@@ -163,3 +169,7 @@ standard_rust_check:
 	cargo clippy --all-targets --all-features --release \
 		-- \
 		-D warnings
+
+# Callers should check if this variable is set to avoid including this
+# file multiple times.
+INCLUDED_UTILS_MK = 1
```
pmores, line 274: I assume there's no way to move this check inside `utils.mk`, is that correct? (I'm not aware of any, just making sure we've considered it.)

jodh-intel: Yes, I considered that. We could do it, nominally we'd end up with `utils.mk` looking like this:

```make
ifeq ($(INCLUDED_UTILS_MK),)

     # << Entire contents of "utils.mk" goes here >>

    INCLUDED_UTILS_MK = 1
endif
```

That would be simpler from the callers perspective (just call `include utils.mk` everywhere you want), but awkward to update `utils.mk` itself.

However, I've just found an alternative which keeps the simpler caller behaviour and avoids the potentially confusing / easy to break `utils.mk` as show above. But there is a catch: we'd need two files, not one:

1. Rename the actual `utils.mk` to something like `private_utils.mk`.
1. Create `utils.mk` as:
    ```make
    ifeq ($(INCLUDED_UTILS_MK),)
        $(eval include private_utils.mk)
    endif
    ```
1. Change all calls to:
    ```make
    include .../utils.mk
    ```

That's a bit more complex, but we could obviously add comments to the `utils.mk` and `private_utils.mk` to explain what's happening.

What do folk think?

pmores: That's interesting! In my view, given the multiple sites where `utils.mk` is included, I think a solution that doesn't require callers to think of putting the `include` statement within guards would be considerably more robust.  From this perspective, both of the solutions you mention would seem acceptable to me (maybe I'm overlooking something but even wrapping the whole contents of the file in guards doesn't seem too bad to me).

fidencio (approved):
lgtm, thanks @jodh-intel!

jodh-intel:
@amshinde - please can you comment on https://github.com/kata-containers/kata-containers/pull/3905#discussion_r873583062.

jodh-intel:
Branch updated to resolve the conflict.

@pmores - I've also gone with the guard in `utils.mk` (aka no extra makefile) and simplified the code in the callers to remove all the checks as that check is now handled by `utils.mk` itself.

/test

pmores (approved):
@jodh-intel Great, thanks! I'm approving this PR with the caveat that I don't think I can really properly review the final commit of the series.

jodh-intel:
Ping @amshinde.

/retest-arm

fidencio:
@jodh, we're closing this PR as it's been opened and inactive for more than one year.

Issue #3913, opened 2022-03-19
Kata shim leak due to container process unclean exit inside guest 


fengwang666:
Recently we observed a failure scenario where kata shim leaks because the container fails to exit. 

### *A normal container exit sequence*:

In the [stopContainer()](https://github.com/containerd/containerd/blob/main/pkg/cri/server/container_stop.go#L60) method inside cri server, it sends signal ([SIGTERM](https://github.com/containerd/containerd/blob/main/pkg/cri/server/container_stop.go#L151) and [SIGKILL](https://github.com/containerd/containerd/blob/main/pkg/cri/server/container_stop.go#L174)) to stop the task and then [wait](https://github.com/containerd/containerd/blob/main/pkg/cri/server/container_stop.go#L179) for a [TaskExit](https://github.com/containerd/containerd/blob/0c701654a1f496d73cb414c7b30b03ee4d115b18/pkg/cri/server/events.go#L365) event to confirm the container has exited successfully. In the case of Kata Container, both `task.Kill()` and `waitContainerStop()` methods eventually translate into grpc requests to the kata agent. 

When a container is started, the shim also starts a [wait](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/start.go#L91) goroutine to reap the container when it exits. The wait goroutine normally is blocked by the [exitIOch](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L33) channel. The shim keeps [three goroutines](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/stream.go#L89) to copy the stdin, stdout and stderr between the container process inside the guest and shim. The `exitIOch` channel is closed when the ioCopy is stopped, which in turn unblocks the wait method and allows it to call [waitProcess](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L46) to reap the process inside the guest VM. 

### *Failure Scenario*:

Previously we have seen ESRCH error not handled in the kill method, which is fixed [here](https://github.com/kata-containers/kata-containers/commit/aa5ae6b17c688a3fa22c28d1ec3a71a6cabb9c57). 
However, after adressing ESRCH, now the container exit is stuck at waiting for the taskExit event. 
[...]

PR #3914, opened 2022-03-19
agent: Close process stream if the process exits already

fengwang666:
This is to address the scenario when there is a bug that prevents a process
from exiting cleanly. The following signal should be able to clean the io stream
and thus unblock the shim side to progress the exit sequence.

Fixes: #3913


fengwang666:
@liubin @Tim-Zhang @bergwolf PTAL

PR #3927, opened 2022-03-22
stable-2.4: Bump QEMU to 6.2 (bringing then SGX support in)

fidencio:
This is a controversial one as we passed "-rc0" already, but would be very nice if we could bump QEMU to 6.2.0 (currently using 6.1.0) as part of 2.4.0 release.

SGX code has already been on Kata Containers for a while and it's supported when using Cloud Hypervisor and QEMU as VMMs. The Cloud Hypervisor version used already supports SGX, and so does the Kernel.  The only missing piece is QEMU, which slipped off the "-rc0".

I'd like to ask the @kata-containers/architecture-committee, @Amulyam24 (as the power maintainer), @Jakob-Naucke (as the still s390x maintainer), and @jongwu (as the ARM maintainer) on whether this would be okay to bring in, and what are your concerns from the multi arch point of view.

### DIFF
--- a/docs/use-cases/using-Intel-SGX-and-kata.md
+++ b/docs/use-cases/using-Intel-SGX-and-kata.md
@@ -21,20 +21,7 @@ CONFIG_X86_SGX_KVM=y
    * [Intel SGX Kubernetes device plugin](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/sgx_plugin#deploying-with-pre-built-images)
 
 > Note: Kata Containers supports creating VM sandboxes with Intel® SGX enabled
-> using [cloud-hypervisor](https://github.com/cloud-hypervisor/cloud-hypervisor/) VMM only. QEMU support is waiting to get the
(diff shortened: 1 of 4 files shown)


fidencio:
ARM CI failure is related to https://github.com/kata-containers/tests/pull/4604, which also requires a backport (and is not related to this PR at all).

devimc (approved):
thanks @fidencio

Jakob-Naucke (approved):
sorry s390x CI is still under maintenance, but I've tested this -- seems to not cause any problems, so fine with me from the arch side, and also LGTM in general. Thanks @fidencio!

jongwu:
LGTM! thanks @fidencio

jongwu (approved):

Amulyam24 (approved):
LGTM, Thanks @fidencio !

fidencio:
Cool, thanks everyone for taking a look at this.

With the green light from the person who worked the original patches, from members of the A/C, and from the other architecture maintainers, I am going ahead and merging this one.

PR #3940, opened 2022-03-23
agent: add tests for is_signal_handled function

### LINKED ISSUE #3939: agent tests: Add tests for is_signal_handled function in src/rpc.rs
Add tests for `is_signal_handled` function in the `src/rpc.rs` file in the agent. Related to #292.

bradenrayhorn:
Add test coverage for is_signal_handled function in rpc.rs. Includes
refactors to make the function testable and handle additional cases.

Fixes #3939

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -407,7 +407,8 @@ impl AgentService {
             // For container initProcess, if it hasn't installed handler for "SIGTERM" signal,
             // it will ignore the "SIGTERM" signal sent to it, thus send it "SIGKILL" signal
             // instead of "SIGTERM" to terminate it.
-            if p.init && sig == libc::SIGTERM && !is_signal_handled(p.pid, sig as u32) {
+            let proc_status_file = format!("/proc/{}/status", p.pid);
+            if p.init && sig == libc::SIGTERM && !is_signal_handled(&proc_status_file, sig as u32) {
                 sig = libc::SIGKILL;
             }
             p.signal(sig)?;
@@ -1635,21 +1636,33 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
     Ok(())
 }
 
-// Check is the container process installed the
+// Check if the container process installed the
 // handler for specific signal.
-fn is_signal_handled(pid: pid_t, signum: u32) -> bool {
-    let sig_mask: u64 = 1u64 << (signum - 1);
-    let file_name = format!("/proc/{}/status", pid);
+fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
+    let shift_count: u64 = if signum == 0 {
+        // signum 0 is used to check for process liveness.
+        // Since that signal is not part of the mask in the file, we only need
+        // to know if the file (and therefore) process exists to handle
+        // that signal.
+        return fs::metadata(proc_status_file).is_ok();
+    } else if signum > 64 {
+        // Ensure invalid signum won't break bit shift logic
+        warn!(sl!(), "received invalid signum {}", signum);
+        return false;
+    } else {
+        (signum - 1).into()
+    };
 
     // Open the file in read-only mode (ignoring errors).
-    let file = match File::open(&file_name) {
+    let file = match File::open(proc_status_file) {
         Ok(f) => f,
         Err(_) => {
-            warn!(sl!(), "failed to open file {}\n", file_name);
+            warn!(sl!(), "failed to open file {}", proc_status_file);
             return false;
         }
     };
 
+    let sig_mask: u64 = 1 << shift_count;
     let reader = BufReader::new(file);
 
     // Read the file line by line using the lines() iterator from std::io::BufRead.
@@ -1657,21 +1670,21 @@ fn is_signal_handled(pid: pid_t, signum: u32) -> bool {
         let line = match line {
             Ok(l) => l,
             Err(_) => {
-                warn!(sl!(), "failed to read file {}\n", file_name);
+                warn!(sl!(), "failed to read file {}", proc_status_file);
                 return false;
             }
         };
         if line.starts_with("SigCgt:") {
             let mask_vec: Vec<&str> = line.split(':').collect();
             if mask_vec.len() != 2 {
-                warn!(sl!(), "parse the SigCgt field failed\n");
+                warn!(sl!(), "parse the SigCgt field failed");
                 return false;
             }
             let sig_cgt_str = mask_vec[1];
             let sig_cgt_mask = match u64::from_str_radix(sig_cgt_str, 16) {
                 Ok(h) => h,
                 Err(_) => {
-                    warn!(sl!(), "failed to parse the str {} to hex\n", sig_cgt_str);
+                    warn!(sl!(), "failed to parse the str {} to hex", sig_cgt_str);
                     return false;
                 }
             };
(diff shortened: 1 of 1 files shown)


src/agent/src/rpc.rs
```diff
@@ -1571,15 +1573,19 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
 
 // Check is the container process installed the
```
bergwolf, line 1574: please fix the comment typo btw, thanks!

`check is` -> `check if`

src/agent/src/rpc.rs
```diff
@@ -1571,15 +1573,19 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
...
 // handler for specific signal.
-fn is_signal_handled(pid: pid_t, signum: u32) -> bool {
+fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
+    if signum < 1 {
```
bergwolf, line 1577: signum 0 is a special number used to check a process liveness. Do we want to keep it being handled?

signum 1 is SIGHUP. we should handle it too IMO.

bradenrayhorn: @bergwolf 
> signum 1 is SIGHUP. we should handle it too IMO.

I think signum 1 is still be handled properly. If it is `1`, the `sig_mask` variable is set to `1`, so it will be able to be compared with the process's `SigCgt` field. Please let me know if I'm misunderstanding.

> signum 0 is a special number used to check a process liveness. Do we want to keep it being handled?

I'm not sure how we would check that using the `SigCgt` field. Would it make sense to just return true if the process status file exists and skip the comparison with the `SigCgt` field in the case `signum == 0`?

src/agent/src/rpc.rs
```diff
@@ -1569,17 +1571,21 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
...
+// Check if the container process installed the
 // handler for specific signal.
-fn is_signal_handled(pid: pid_t, signum: u32) -> bool {
+fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
+    if signum < 1 {
+        warn!(sl!(), "received invalid signum {}\n", signum);
+        return false;
+    }
+
     let sig_mask: u64 = 1u64 << (signum - 1);
```
jodh-intel, line 1582: Not part of the change, but this line is a potential crasher bug: if `signum > 63`, this will fail.

We can't currently hit that scenario since the (only current) caller first calls `Signal::try_from(req.signal)`, but I do wonder if we should add a check here too. It's debatable. We could leave as-is (potentially dangerous for future changes), call  `Signal::try_from(req.signal)`. I'd consider adding a simpler solution which constrains the value by it's absolute maximum, not by thet maximum signal number. Not perfect, but fast and avoids any possibility of a crash here:

```rust
if signum > 63 {
    // Handle impossible scenarios
    return false;
}

File: src/agent/src/rpc.rs (hunk 2 above)
Comment (jodh-intel) [reply src/agent/src/rpc.rs:1577]: Yep - I think this is fine as we're not stopping signal zero being delivered, we're just checking the signal mask here.

**Review by jodh-intel — COMMENTED** [review]
Thanks @bradenrayhorn. Looks good, but see my comment on adding an extra check in `is_signal_handled()`...

**Comment by bradenrayhorn** [comment]
Updated to resolve merge conflict.

File: src/agent/src/rpc.rs [hunk 4]
```diff
@@ -1634,29 +1635,37 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
...
 }
 
-// Check is the container process installed the
+// Check if the container process installed the
 // handler for specific signal.
-fn is_signal_handled(pid: pid_t, signum: u32) -> bool {
-    let sig_mask: u64 = 1u64 << (signum - 1);
-    let file_name = format!("/proc/{}/status", pid);
+fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
+    if signum > 64 {
```
jodh-intel, line 1641: Can you add a comment here explaining what this test is doing? Nominally, it appears to be just checking against some arbitrary maximum value for a signal number. But that's not the case - this line is "bound" with the bit shift done on line 1660.

Alternatively, you could do something like this (not: no `\n` required on the log calll):

```rust
// Ensure invalid signal number doesn't break the shift logic
let shift_count: u64 = if signum > 64 {
    warn!(sl!(), "received invalid signum {}", signum);
    return false;
} else {
    signum - 1
};

// ...

let sig_mask: u64 = 1u64 << shift_count;
```

src/agent/src/rpc.rs
```diff
@@ -1634,29 +1635,37 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
...
+        return false;
+    }
 
     // Open the file in read-only mode (ignoring errors).
-    let file = match File::open(&file_name) {
+    let file = match File::open(proc_status_file) {
         Ok(f) => f,
         Err(_) => {
-            warn!(sl!(), "failed to open file {}\n", file_name);
+            warn!(sl!(), "failed to open file {}\n", proc_status_file);
```
jodh-intel, line 1650: As mentioned above, none of the log calls should have a newline.

jodh-intel (commented):
Thanks @bradenrayhorn. A few comments...

src/agent/src/rpc.rs
bradenrayhorn: Thanks @jodh-intel, good suggestion. I left the `signum - 1` where it was, because moving it above the process file check and `signum == 0` check led to trying to put a negative number in an unsigned integer, as we are allowing `0` as a valid signum.

I guess I could have moved the `signum > 64` check below the process file check and `signum == 0` check, but I think having the validation at the top of function makes the most sense.

bradenrayhorn:
Resolved merge conflict

src/agent/src/rpc.rs
```diff
@@ -2305,6 +2315,102 @@ mod tests {
...
+                status_file_data: None,
+                signum: 0,
+                result: false,
+            },
+        ];
+
+        for (i, d) in tests.iter().enumerate() {
+            let msg = format!("test[{}]: {:?}", i, d);
+
+            let dir = tempdir().expect("failed to make tempdir");
```
liubin, line 2399: Nit: Moving creating of tmpdir (line 2399 and 2400) out of `for` loop may improve a bit performance.

liubin (approved):

src/agent/src/rpc.rs
jodh-intel: +1 for handling the checks at the start of the function. The problem is having multiple checks on the same variable spread out through that function. You could handle all those checks first with something like this:

```diff
diff --git a/src/agent/src/rpc.rs b/src/agent/src/rpc.rs
index 488e5edf2..dc85db043 100644
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -1639,11 +1639,21 @@ fn append_guest_hooks(s: &Sandbox, oci: &mut Spec) -> Result<()> {
 // Check if the container process installed the
 // handler for specific signal.
 fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
-    // Ensure invalid signum won't break bit shift logic
-    if signum > 64 {
+    let shift_count: u64 = if signum == 0 {
+        // signum 0 is used to check for process liveness.
+        // Since that signal is not part of the mask in the file, we only need
+        // to know if the file (and therefore) process exists to handle
+        // that signal.
+        return fs::metadata(proc_status_file).is_ok();
+    } else if signum > 64 {
+        // Ensure invalid signal number doesn't break the shift logic.
         warn!(sl!(), "received invalid signum {}", signum);
         return false;
-    }
+    } else {
+        (signum - 1).into()
+    };
+
+    let sig_mask: u64 = 1u64 << shift_count;

     // Open the file in read-only mode (ignoring errors).
     let file = match File::open(proc_status_file) {
@@ -1654,12 +1664,6 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
         }
     };

-    // signum 0 is used to check for process liveness
-    if signum == 0 {
-        return true;
-    }
-
-    let sig_mask: u64 = 1u64 << (signum - 1);
     let reader = BufReader::new(file);

     // Read the file line by line using the lines() iterator from std::io::BufRead.
```

bradenrayhorn: @jodh-intel Thanks, that all makes sense. I'll keep it in mind for the future.

src/agent/src/rpc.rs
bradenrayhorn: I think I'll leave it as is, moving the tempdir creation outside of the for loop would necessitate some more logic in the loop to delete the file if it was created by one of the previous test cases.

jodh-intel (approved):
Thanks @bradenrayhorn.

lgtm

arronwy:

Issue #3941, opened 2022-03-23
Hook unit tests should rely on globally installed test hook


dgibson:
# Description of problem

A handful of the tests in `src/runtime/pkg/katautils/hook_test.go` require a specific mock test hook, which is built from code in `src/runtime/virtcontainers/hook/mock` and installed to `/usr/bin/virtcontainers/bin/test/hook`.

Unit tests shouldn't rely on non-standard components installed system-wide, and they *definitely* shouldn't install things system wide, which the current Make target to run the tests does.

PR #3942, opened 2022-03-23
Don't use a globally installed mock hook for hook tests

dgibson:
Address #3941.  Make some additional cleanups in vicinity while we're at it.


fidencio (changes requested):
I can't review the commit message, but ...
```
virtcontainers: Run mock hook from build tree rather than system bin dir

Running unit tests should generally have minimal dependencies on things
outside the build tree.  It *definitely* should modify system wide things
outside the build tree.  Currently the runtime "make test" target does so,
though.
...
```

By saying "It *definitely* should modify system wide things
outside the build tree. ", did you mean "It *definitely* should NOT modify system wide things
outside the build tree. "?

```
virtcontainers: Remove unused install/uninstall targets
We've now removed the need to install the mock hook binary for unit tests.
However, it turns out that managing that was the *only* thing that the
install and uninstall targets in the virtcontainers Makefile handlded.
```

Shouldn't "handlded" be "handled"?

Requesting changes till the commit messages are fixed.

Changes are good otherwise.

I'd let CI finish this run before resubmitting the series, so we can have an idea whether it's happy or not with those.

c3d:
@fidencio Could you edit your comment above to replace "NOT should" by "should NOT" ? ;-)

fidencio:
> @fidencio Could you edit your comment above to replace "NOT should" by "should NOT" ? ;-)

Done.

dgibson:
> I can't review the commit message, but ...
> 
> ```
> virtcontainers: Run mock hook from build tree rather than system bin dir
> 
> Running unit tests should generally have minimal dependencies on things
> outside the build tree.  It *definitely* should modify system wide things
> outside the build tree.  Currently the runtime "make test" target does so,
> though.
> ...
> ```
> 
> By saying "It _definitely_ should modify system wide things outside the build tree. ", did you mean "It _definitely_ should NOT modify system wide things outside the build tree. "?

Oops, yes, corrected.

> ```
> virtcontainers: Remove unused install/uninstall targets
> We've now removed the need to install the mock hook binary for unit tests.
> However, it turns out that managing that was the *only* thing that the
> install and uninstall targets in the virtcontainers Makefile handlded.
> ```
> 
> Shouldn't "handlded" be "handled"?

Oops, again, fixed.

liubin (approved):
Thank you @dgibson, installing the hook to the system is really a bad idea.

fidencio (approved):
lgtm, thanks @dgibson!

Issue #3928, opened 2022-03-22
doc | CCv0: Update the documentation after the logging changes


stevenhorsman:
Following https://github.com/kata-containers/tests/issues/4609, we'll need to update the CCv0 documentation and/or scripts to indicate that we get the agent logs from the journal rather than the console socket.

This should also be done after https://github.com/kata-containers/kata-containers/pull/3899 is merged to avoid conflicts.

stevenhorsman:
Closed with #3945 merging

PR #3945, opened 2022-03-23
docs: Update to reflect logging changes

stevenhorsman:
- We've updated the CC logging scripts to log to the journal
rather than a socket, so remove socat scripts and instructions
to reflect this

Depends-on: github.com/kata-containers/tests#4610
Fixes: #3928

### DIFF
--- a/docs/how-to/ccv0.sh
+++ b/docs/how-to/ccv0.sh
@@ -106,7 +106,6 @@ Commands:
 - kubernetes_create_ssh_demo_pod:   Create a Kata CC runtime pod based on the ssh demo
 - kubernetes_delete_cc_pod:         Delete the Kata CC runtime busybox-based pod in Kubernetes
 - kubernetes_delete_ssh_demo_pod:   Delete the Kata CC runtime pod based on the ssh demo
-- open_kata_console:                Stream the kata runtime's console
 - open_kata_shell:                  Open a shell into the kata runtime
 - rebuild_and_install_kata:         Rebuild the kata runtime and agent and build and install the image
 - shim_pull_image:                  Run PullImage command against the shim with ctr
@@ -146,7 +145,7 @@ rebuild_and_install_kata() {
 # Based on the jenkins_job_build.sh script in kata-containers/tests/.ci - checks out source code and installs dependencies
 initialize() {
     # We need git to checkout and bootstrap the ci scripts and some other packages used in testing
-    sudo apt-get update && sudo apt-get install -y curl git socat qemu-utils
+    sudo apt-get update && sudo apt-get install -y curl git qemu-utils
     
     grep -qxF "export GOPATH=\${HOME}/go" "${PROFILE}" || echo "export GOPATH=\${HOME}/go" >> "${PROFILE}"
     grep -qxF "export GOROOT=/usr/local/go" "${PROFILE}" || echo "export GOROOT=/usr/local/go" >> "${PROFILE}"
@@ -208,8 +207,9 @@ build_and_install_kata_runtime() {
 }
 
 configure() {
-    debug_function configure_kata_to_use_rootfs
-    debug_function enable_full_debug
+    configure_kata_to_use_rootfs
+    enable_full_debug
+    enable_agent_console
 
     # Switch image offload to true in kata config
     switch_image_service_offload "on"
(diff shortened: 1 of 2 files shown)


wainersm (approved):
LGTM.

GabyCT (approved):

c3d (commented):
I somewhat understand the rationale, but I wonder if it would not be useful to keep the "console" option for developers.

Just looking at how the messages look before/after:

```
FATA[0001] Source image rejected: Invalid GPG signature...
```

becomes

```
vmconsole="\x1b[31mFATA\x1b[0m[0002] Source image rejected: Signature for identity quay.io/kata-containers/confidential-containers:signed is not accepted "
```

BTW, should we try to strip the color escapes?

stevenhorsman:
Hi Christophe, thanks for the comments. I think they are fair, but I'll add my thoughts and see what you think:
> I somewhat understand the rationale, but I wonder if it would not be useful to keep the "console" option for developers.

I was originally keen to do this (as I like using the console), but my main concern is that @fidencio found that the console streaming approach didn't work for cloud hypervisor, so rather than add separate docs for different hypervisors and possibly confuse people (well confuse me at least!) I'd try and work with the lowest common denominator.
I'd love to just make both options available, so users could pick and choose, but logging to the journal, or console seem to be mutually exclusive in Kata. When `enable_debug=true` is set in the hypervisor section of the Kata configuration then the logs go out to the journal, but stop showing on the console.
Happy to take another look if anyone has any suggestions that I've missed though!

> Just looking at how the messages look before/after:
> 
> ```
> FATA[0001] Source image rejected: Invalid GPG signature...
> ```
> 
> becomes
> 
> ```
> vmconsole="\x1b[31mFATA\x1b[0m[0002] Source image rejected: Signature for identity quay.io/kata-containers/confidential-containers:signed is not accepted "
> ```
> 
> BTW, should we try to strip the color escapes?

I think this would be a good idea, but I'm not sure how to do it. I assume that `logrus` or something in the containers libraries are applying this and relying on `tty` to convert it, but I can't spot a configuration parameter and it seems the Kata logging system isn't aware to do the conversion.
The only good news is that we are 🤞 only a very short timeframe away from replacing skopeo with the image-rs native function, so the logging there will be different and should be customisable, so I think it's probably not worth investing too much time in.

jodh-intel (approved):
Thanks @stevenhorsman.

lgtm

c3d:
> Hi Christophe, thanks for the comments. I think they are fair, but I'll add my thoughts and see what you think:
> 
> I was originally keen to do this (as I like using the console), but my main concern is that @fidencio found that the console streaming approach didn't work for cloud hypervisor, so rather than add separate docs for different hypervisors and possibly confuse people (well confuse me at least!) I'd try and work with the lowest common denominator.

It's a bit annoying to make things more inconvenient for developers because one particular hypervisor does not support the "easy way". IMHO, it would be better to go the other way, i.e. make it easy for everyone not using cloud hypervisor, and have a special option (or improve cloud hypervisor) for the inconvenient case.

> I'd love to just make both options available, so users could pick and choose, but logging to the journal, or console seem to be mutually exclusive in Kata. When `enable_debug=true` is set in the hypervisor section of the Kata configuration then the logs go out to the journal, but stop showing on the console. Happy to take another look if anyone has any suggestions that I've missed though!

Hmmm… I have not checked in the code, but what you describe seems backwards. When we enable debug, isn't that when we should get more direct access to the logs?

> >   
> > BTW, should we try to strip the color escapes?
> 
> I think this would be a good idea, but I'm not sure how to do it. I assume that `logrus` or something in the containers libraries are applying this and relying on `tty` to convert it, but I can't spot a configuration parameter and it seems the Kata logging system isn't aware to do the conversion.

No, I don't think it exists. But while we rework these parts, it could be nice to remove stuff that really gets in the way of log readability (notably, seeing this in CI logs would make search more difficult).

> The only good news is that we are 🤞 only a very short timeframe away from replacing skopeo with the image-rs native function, so the logging there will be different and should be customisable, so I think it's probably not worth investing too much time in.

Agreed. Also, another good news is that @littlejawa  is presently looking at "deinterlacing" the agent's output, reworking `kata-log-parser`. So maybe that will be sufficient to clean things up after the fact.

c3d:
@stevenhorsman I think this is an improvement, so merging. We can address my comments later, notably as part of the work @littlejawa is doing on log parsing.

stevenhorsman:
> @stevenhorsman I think this is an improvement, so merging. We can address my comments later, notably as part of the work @littlejawa is doing on log parsing.

@c3d - thanks Christophe - I appreciate that. I have on my todo list to investigate the debug setting affecting the console socket, so will raise an issue if I find anything interesting!

PR #3955, opened 2022-03-24
agent: Signal the whole process group

fengwang666:
This PR addresses #3913.

As described in #3913, the current container termination is insufficient for use cases where a pid namespace is shared among multiple containers. I updated the kata agent container termination logic to follow the [runc](https://github.com/opencontainers/runc/blob/485e6c84e7ef783c126c392af072a974c9fb88ce/libcontainer/init_linux.go#L528) implementation that kills all the processes in the same cgroup, as it's safe to assume each container has its own cgroup. The other change is to set `KillRequest.All` as `true` when the `ExecID` is empty. This separates the container termination versus container exec signaling. 

It's also worth calling out, I didn't change the kata shim v2 service to follow the [runc v2 logic](https://github.com/fengwang666/containerd/blob/e48bbe83949a43dedd3e2727452259f99dd81635/runtime/v2/runc/v2/service.go#L766-L773) where they terminate the child processes after the init process exits. The main reason is because the container termination sequences in runc and kata is a bit different. In runc, the runtime service gets notified immediately after the container process exits by [subscribing](https://github.com/containerd/containerd/blob/e48bbe83949a43dedd3e2727452259f99dd81635/runtime/v2/runc/v2/service.go#L98) to the process reaper, but in kata, the shim needs to poll the kata agent through a [wait](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go) goroutine that blocks on the io copy streams until the container has exited cleanly. It would be quite complicated trying to fit the runc approach to the current kata container termination sequence. And I believe it's sufficient to kill the container processes all at once.

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -19,6 +19,7 @@ use ttrpc::{
 };
 
 use anyhow::{anyhow, Context, Result};
+use cgroups::freezer::FreezerState;
 use oci::{LinuxNamespace, Root, Spec};
 use protobuf::{Message, RepeatedField, SingularPtrField};
 use protocols::agent::{
@@ -40,8 +41,9 @@ use rustjail::specconv::CreateOpts;
 use nix::errno::Errno;
 use nix::mount::MsFlags;
 use nix::sys::signal::Signal;
-use nix::sys::stat;
+use nix::sys::{signal, stat};
 use nix::unistd::{self, Pid};
+use rustjail::cgroups::Manager;
 use rustjail::process::ProcessOperations;
 
 use sysinfo::{DiskExt, System, SystemExt};
@@ -391,7 +393,6 @@ impl AgentService {
         let cid = req.container_id.clone();
         let eid = req.exec_id.clone();
         let s = self.sandbox.clone();
-        let mut sandbox = s.lock().await;
 
         info!(
             sl!(),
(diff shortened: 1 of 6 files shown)


fengwang666:
Pending verification

src/agent/rustjail/src/process.rs
```diff
@@ -93,7 +93,7 @@ impl ProcessOperations for Process {
     }
 
     fn signal(&self, sig: Signal) -> Result<()> {
-        signal::kill(self.pid(), Some(sig))
+        signal::kill(Pid::from_raw(-self.pid), Some(sig))
```
fengwang666, line 96: probably need to separate the signal and only do this when it's SIGTERM/SIGKILL. Will update tomorrow.

src/agent/src/rpc.rs
```diff
@@ -400,24 +401,45 @@ impl AgentService {
...
             ))
         })?;
 
+        if eid.is_empty() {
+            // eid is empty, signal all the processes in the container cgroup
+            let ctr = sandbox.get_container(&cid).ok_or_else(|| anyhow!("Invalid container id"))?;
+            match ctr.cgroup_manager {
+                None => Err(anyhow!("container cgroup not exist")),
+                Some(cm) => {
+                    let pids = cm.get_pids();
```
liubin, line 417: CgroupManager.gett_pids calls cgroup-rs' `tasks()` function, which returns values of tasks that may contain threads that are not in the container's process(`sandbox.find_process(pid)`) list. So considering doesn't return an error at line 422?

src/agent/src/rpc.rs
```diff
@@ -400,24 +401,46 @@ impl AgentService {
...
+            let ctr = sandbox
+                .get_container(&cid)
+                .ok_or_else(|| anyhow!("Invalid container id"))?;
+            let cm = if ctr.cgroup_manager.is_some() {
+                ctr.cgroup_manager.as_ref().unwrap()
+            } else {
+                return Err(anyhow!("container cgroup not exist"));
+            };
+
+            let pids = cm.get_pids()?;
```
bergwolf, line 422: Do we need to freeze the cgroup manager, like runc shimv2 dose? Otherwise it looks like there can be a race window that some process forked another process after `cm.get_pids()` and then we'll fail to kill it and leak it.

Another option might be that we send `kill(-pid, signal)` to all processes returned by the cgroup manager.

fengwang666: @bergwolf Thanks. I have updated the change following your suggestion.

fengwang666:
@kata-containers/architecture-committee Updated the description on the rationale of the change. I have tested and verified the issue has been fixed. PTAL.

bergwolf (approved):
lgtm! Thanks @fengwang666 !

liubin (approved):

PR #3957, opened 2022-03-24
CCv0: agent: Support https_proxy config for image download in guest

### LINKED ISSUE #3956: CCv0: Support https_proxy config for kata-agent
Containerd can support set a proxy when downloading images with a environment variable.
For CC stack, image download is offload to the Kata agent, we also need support similar feature.

arronwy:
Containerd can support set a proxy when downloading images with a environment variable.
For CC stack, image download is offload to the kata agent, we need support similar feature.

Fixes #3956

Signed-off-by: Arron Wang <arron.wang@intel.com>

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -11,6 +11,7 @@ use std::fs;
 use std::str::FromStr;
 use std::time;
 use tracing::instrument;
+use url::Url;
 
 const DEBUG_CONSOLE_FLAG: &str = "agent.debug_console";
 const DEV_MODE_FLAG: &str = "agent.devmode";
@@ -24,6 +25,8 @@ const CONTAINER_PIPE_SIZE_OPTION: &str = "agent.container_pipe_size";
 const UNIFIED_CGROUP_HIERARCHY_OPTION: &str = "agent.unified_cgroup_hierarchy";
 const CONFIG_FILE: &str = "agent.config_file";
 const CONTAINER_POLICY_FILE: &str = "agent.container_policy_file";
+const HTTPS_PROXY: &str = "agent.https_proxy";
+const NO_PROXY: &str = "agent.no_proxy";
 
 const DEFAULT_LOG_LEVEL: slog::Level = slog::Level::Info;
 const DEFAULT_HOTPLUG_TIMEOUT: time::Duration = time::Duration::from_secs(3);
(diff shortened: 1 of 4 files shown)


fidencio (commented):
Wow, I clearly can see where it comes from.

Is `https_proxy` the only one that has to be exported?  Wouldn't we have the same issue if we don't export `http_proxy` as well?

arronwy:
> Wow, I clearly can see where it comes from.
> 
> Is `https_proxy` the only one that has to be exported? Wouldn't we have the same issue if we don't export `http_proxy` as well?

Yes, the reason we only export `https_proxy` is a plain http registry is insecure which may suffer man-in-the-middle attack and most of the production ready registries are https based, we can add `http_proxy` support if we really need it.

src/agent/src/config.rs
```diff
@@ -24,6 +24,7 @@ const CONTAINER_PIPE_SIZE_OPTION: &str = "agent.container_pipe_size";
 const UNIFIED_CGROUP_HIERARCHY_OPTION: &str = "agent.unified_cgroup_hierarchy";
 const CONFIG_FILE: &str = "agent.config_file";
 const CONTAINER_POLICY_FILE: &str = "agent.container_policy_file";
+const HTTPS_PROXY: &str = "agent.https_proxy";
```
devimc, line 27: can we need no_proxy to pull images from internal registries?

arronwy: Yes, if we pull image from both internal and external registries, we need no_proxy, fixed now.

fidencio (approved):
lgtm, thanks @arronwy!

src/agent/src/config.rs
```diff
@@ -84,6 +86,8 @@ pub struct AgentConfig {
     pub supports_seccomp: bool,
     pub container_policy_path: String,
     pub aa_kbc_params: String,
+    pub https_proxy: String,
+    pub no_proxy: String,
```
jodh-intel, line 90: I think these should be `Url`'s so we can use the `url` crate so we can call `url::Url::parse(value)?` on them.

jodh-intel (changes requested):
Thanks for raising @arronwy.

src/agent/src/config.rs
arronwy: > I think these should be `Url`'s so we can use the `url` crate so we can call `url::Url::parse(value)?` on them.

Yes, it make sense to do more checks on these inputs.
`url::Url::parse()` will require a `scheme` like `http` seperated by `:`, it will be ok for `https_proxy`. we can add a `get_url_value()` based on current `get_string_value()` with url parse check.
But for `no_proxy` it will be a list of IP with CIDR/asterisks or domain names, `url::Url::parse()` can not check them correctly.

jodh-intel (approved):
Thanks @arronwy.

lgtm

PR #3964, opened 2022-03-25
runtime-rs: Make macros in agent crate more readable

c3d:
There are two useful macros in the agent crate to implement boilerplate
code. However, the syntax that was chosen for the macros make them a bit hard to
read. Use a syntax that is a bit closer to normal Rust syntax.

Signed-off-by: Christophe de Dinechin <dinechin@redhat.com>


src/runtime-rs/crates/agent/src/kata/agent.rs
```diff
@@ -58,19 +58,18 @@ macro_rules! impl_health_service {
...
 
 macro_rules! impl_agent {
-    ($($name: tt | $req: ty | $resp: ty | $new_timeout: expr),*) => {
+    ($(impl $name:ident ($req:ident, $resp:ident) await $new_timeout:expr),*) => {
         #[async_trait]
         impl Agent for KataAgent {
-            $(async fn $name(&self, req: $req) -> Result<$resp> {
-                let r = req.into();
+            $(async fn $name(&self, req: crate::$req) -> Result<crate::$resp> {
+                let r = reeq.into();
```
jodh-intel, line 70: Syntax error? `reeq` -> `req`?

src/runtime-rs/crates/agent/src/kata/agent.rs
```diff
@@ -58,19 +58,18 @@ macro_rules! impl_health_service {
...
 impl_health_service!(
-    check | crate::CheckRequest | crate::HealthCheckResponse,
-    version | crate::CheckRequest | crate::VersionCheckResponse
+    impl check (CheckRequest, HealthCheckResponse),
+    impl version (CheckRequest, VersionCheckResponse)
 );
 
 macro_rules! impl_agent {
-    ($($name: tt | $req: ty | $resp: ty | $new_timeout: expr),*) => {
+    ($(impl $name:ident ($req:ident, $resp:ident) await $new_timeout:expr),*) => {
```
jodh-intel, line 66: I think you can remove `await` from here and from all the `impl`'s below?

jodh-intel (commented):
Thanks @c3d.

/cc @liubin.

src/runtime-rs/crates/agent/src/kata/agent.rs
c3d: I could, but I thought that using the "await" keyword to prefix the timeout made the intent clearer.

liubin: It's great to make macro more function likely for better reading.

And I think since almost all functions in agent are async, `await` can be omitted.

Nit: Only two functions need the `timeout` parameter, so can it be an optional parameter of the macro?
For example change `$new_timeout:expr` to `$($new_timeout:expr)?` and impl_agent will be :

```rust
    impl signal_process (SignalProcessRequest, Empty),
    impl wait_process (WaitProcessRequest, WaitProcessResponse) Some(0),
```

bergwolf: `await` is a rust keyword. I would prefer not to use it for other purposes. The `new_timeout` parameter already makes it clear that it is a timeout. The caller shouldn't care if it is used for await or other things for abstraction purposes. If it must be improved, how about `timeout_option` instead, to make it clear that it is an `Option`?

arronwy:

jodh-intel:
Hi @c3d - any update on this?

c3d:
Hi @jodh-intel, I was on PTO last week. I need to do a respin, and also need to understand what happened with the typo about `reeq`, which I fixed on my side, so I don't understand why my force-push did not bring up the change.

Unfortunately, I am presently fighting a broken laptop, which forces me to work in non-ideal conditions. I'll try to quickly post a respin, but that may take a few more hours

src/runtime-rs/crates/agent/src/kata/agent.rs
c3d: I understand your argument. I used `await` precisely because I thought using a keyword made the macro parsing more robust (keywords being reserved). In the notation @liubin proposed, we don't know what the `Some(0)` represents.

If I create variants, what about replacing `await` with `timeout` ? In other words, I'd prefer `timeout 0` than `Some(0)`.

jodh-intel:
Hi @c3d - Sorry to hear about the lappy! No panic, I just wondering as it would be good to land this.

src/runtime-rs/crates/agent/src/kata/agent.rs
bergwolf: @c3d Sorry for the late response! I didn't realise that it was blocked by me til now.

IIUC, you meant to change `$new_timeout:expr` into a value, dropping the `Option` property, right?

In the original form, there are three cases:
1. None: use the default client timeout
2. Some(0): no timeout
3. Some(v): set timeout to v

We cannot express the first case without accepting an `Option` value.

bergwolf:
@c3d Any updates on this one?

PR #3972, opened 2022-03-26
agent: use ms as unit of cputime instead of ticks

wfly1998:
For the library `procfs`, the unit of values in `CpuTime` is ticks, and we do not know how many ticks per second from metrics because the `tps` in `CpuTime` is private\[1\].

But there are some implements in `CpuTime` for getting these values\[2\], e.g., `user_ms()` for `user`, and `nice_ms()` for `nice`.  With these values, accurate time can be obtained.

\[1\]: https://github.com/eminence/procfs/blob/cc431de72dd4178e057d19b280654df52c6ae4be/src/lib.rs#L799
\[2\]: https://github.com/eminence/procfs/blob/cc431de72dd4178e057d19b280654df52c6ae4be/src/lib.rs#L741-L748


liubin:
@wfly1998 You should create an issue describing what you want to fix, and write the `Fixes:` in your commit message.

You can get commit message format [here](http://liubin.org/kata-dev-book/src/your-first-pr.html#%E6%8F%90%E4%BA%A4%E4%BB%A3%E7%A0%81).

wfly1998:
@liubin My issue #3979 is opened and the commit with `Fixes` has been force pushed, please review it, thanks.

liubin (approved):

amshinde (approved):

wfly1998:
@liubin Excuse me and I was reminded of a problem just now.

This commit fixed the problem that the unit of cputime was ticks after upgrading `procfs` and I changed it to milliseconds, but the previous unit of cputime is seconds.

So do I need to submit another commit for changing milliseconds to seconds?

Issue #3973, opened 2022-03-26
Add **all** rust vendored code as part of the vendored tarball distributed as part of the release


fidencio:
@etrunko raised an issue that, when doing a downstream build of kata-containers, some of the vendored code was not present.
And that is indeed the case.  Right now we're only shipping the agent vendored code, and we could and should actually ship everything we needed for a full build of kata-containers.

PR #3974, opened 2022-03-26
release: Include all the rust vendored code into the vendored tarball

fidencio:
workflows,release: Ship *all* the rust vendored code

Instead of only vendoring the code needed by the agent, let's ensure we
vendor all the needed rust code, and let's do it using the newly
introduced enerate_vendor.sh script.

Fixes: #3973

---

tools: Add a generate_vendor.sh script

This script is responsible for generating a tarball with all the rust
vendored code that is needed for fully building kata-containers on a
disconnected environment.

---

How to verify:
* From the top dir of the project, do:
```
bash -c "tools/packaging/release/generate_vendor.sh kata-containers-vendor.tar.gz"
```
* Do a downstream build using the vendored source, instead of the one currently used.

@etrunko, please, verify if it works on Fedora and let us know ASAP.

### DIFF
--- a/tools/packaging/release/generate_vendor.sh
+++ b/tools/packaging/release/generate_vendor.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+script_name="$(basename "${BASH_SOURCE[0]}")"
+
+# This is very much error prone in case we re-structure our
+# repos again, but it's also used in a few other places :-/
+repo_dir="${script_dir}/../../.."
+
+function usage() {
+
+	cat <<EOF
+Usage: ${script_name} tarball-name
+This script creates a tarball with all the cargo vendored code
+that a distro would need to do a full build of the project in
+a disconnected environment, generating a "tarball-name" file.
+
+EOF
+
+}
+
+create_vendor_tarball() {
+	vendor_dir_list=""
+	pushd ${repo_dir}
+		for i in $(find . -name 'Cargo.lock'); do
+			dir="$(dirname $i)"
+			pushd "${dir}"
+				[ -d .cargo ] || mkdir .cargo
+				cargo vendor >> .cargo/config
+				vendor_dir_list+=" $dir/vendor $dir/.cargo/config"
+				echo "${vendor_dir_list}"
+			popd
+		done
+	popd
+	
+	tar -cvzf ${1} ${vendor_dir_list}
+}
+
+main () {
+	[ $# -ne 1 ] && usage && exit 0
+	create_vendor_tarball ${1}
+}
+
+main "$@"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -141,13 +141,10 @@ jobs:
       - uses: actions/checkout@v2
       - name: generate-and-upload-tarball
         run: |
-          pushd $GITHUB_WORKSPACE/src/agent
-          cargo vendor >> .cargo/config
-          popd
           tag=$(echo $GITHUB_REF | cut -d/ -f3-)
           tarball="kata-containers-$tag-vendor.tar.gz"
           pushd $GITHUB_WORKSPACE
-          tar -cvzf "${tarball}" src/agent/.cargo/config src/agent/vendor
+          bash -c "tools/packaging/release/generate_vendor.sh ${tarball}"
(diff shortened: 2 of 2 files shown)


fidencio:
@jongwu, mind to take a look at the ARM failure?

snir911 (approved):
I thought cargo vendor should do that :P 
It doesn't work for me from the agent dir, only from the top dir (if it's not intentional)
I'd wait to Eduardo to verify but LGTM

etrunko:
This script is kind of broken, but I don't know exactly the reason.
What I did:

1) Download and extract the -rc0.tar.gz
2) Copy the generate_vendor.sh scritpt to the tree and adjust the repo_dir variable
3) run ./generate_vendor.sh kata-containers-2.4.0-rc0-vendor.tar.gz. Then I get:

```
~/projects/rhel/kata-containers/kata-containers-2.4.0-rc0/src/agent/rustjail ~/projects/rhel/kata-containers/kata-containers-2.4.0-rc0 ~/projects/rhel/kata-containers/kata-containers-2.4.0-rc0
error: config values not loaded yet
```

etrunko:
I changed the script to ignore the src/agent/rustjail/Cargo.toml and src/agent/vsock-exporter/Cargo.toml and it now generates the tarball correctly. I will try it with the offline builder now.

fidencio:
> I changed the script to ignore the `src/agent/rustjail/Cargo.toml` and `src/agent/vsock-exporter/Cargo.toml` and it now generates the tarball correctly. I will try it with the offline builder now.

One way to work this around would be searching for "Cargo.lock" rather than "Cargo.toml".
I would not like to hardcode all the paths we need to look at as that will just break by the moment a new path exists.

@liubin, @bergwolf, @jodh-intel, @dgibson, any suggestion on this one?

dgibson:
I think looking for `Cargo.lock` might be the way to go here.  I'm not an expert on cargo stuff, so I'm not certain though.

tools/packaging/release/generate_vendor.sh
```diff
@@ -33,7 +33,7 @@ create_vendor_tarball() {
 	# This is very much error prone in case we re-structure our
 	# repos again, but it's also used in a few other places :-/
 	pushd ${repo_dir}
-		for i in `find . | grep Cargo.toml`; do
+		for i in `find . | grep Cargo.lock`; do
```
c3d, line 36: Why is this not

```
for in in $(find . -name 'Cargo.lock'); do
```

??? Why use two processes when one would do.

Also, can you update the comment above to explain what you are looking for. My understanding is that you only have the `Cargo.lock` for top-level components (as opposed to downloaded crates). Is that the intent?

c3d (approved):
Overall, /lgtm, but I think a comment explaining the `Cargo.lock` trick better than "it's done elsewhere" (_what_ is done???) would be welcome.

tools/packaging/release/generate_vendor.sh
fidencio: I don't understand enough to give you an answer.
I actually asked for something quite similar a few comments above, see https://github.com/kata-containers/kata-containers/pull/3974#issuecomment-1081175702

fidencio: Oh, the comment shouldn't be there, the comment actually applies to the "repo_dir=" var assignment.

fidencio:
> Overall, /lgtm, but I think a comment explaining the `Cargo.lock` trick better than "it's done elsewhere" (_what_ is done???) would be welcome (I'm referring to the comment just above that line).

The comment shouldn't even be there in the first place, that commend was about the `repo_dir` being set.
I've adopted and adapted your suggestion and ended up using:
```
for i in $(find . -name 'Cargo.lock'); do
`` 
(Yours was `for in in ...`, which is wrong ;-)).

I didn't add a comment saying we're looking for `Cargo.lock` as that seems easy to guess from the command.

Apart from those, did you check that a downstream build work with this?  That's a blocker for the release right now and we'd expect that to be done ASAP.

c3d:
> > Overall, /lgtm, but I think a comment explaining the `Cargo.lock` trick better than "it's done elsewhere" (_what_ is done???) would be welcome (I'm referring to the comment just above that line).
> 
> The comment shouldn't even be there in the first place, that comment was about the `repo_dir` being set. I've adopted and adapted your suggestion and ended up using:
> 
> ```
> for i in $(find . -name 'Cargo.lock'); do
> ```
> 
> (Yours was `for in in ...`, which is wrong ;-)).

Just checking if you were awake ;-)

> 
> I didn't add a comment saying we're looking for `Cargo.lock` as that seems easy to guess from the command.
> 
> Apart from those, did you check that a downstream build work with this? That's a blocker for the release right now and we'd expect that to be done ASAP.

fidencio:
Merging as @etrunko confirmed it worked on Fedora.

PR #3977, opened 2022-03-28
stable-2.4 | tools: release: Do not consider release candidates as stable releases

fidencio:
During the release of 2.4.0-rc0 @egernst noticed an incositency in the
way we handle release tags, as release candidates are being taken as
"stable" releases, while both the kata-deploy tests and the release
action consider this as "latest".

Ideally we should have our own tag for "release candidate", but that's
something that could and should be discussed more extensively outside of
the scope of this quick fix.

For now, let's align the code generating the PR for bumping the release
with what we already do as part of the release action and kata-deploy
test, and tag "-rc"  as latest, regardless of which branch it's coming
from.

Fixes: #3847

(cherry picked from commit 4adf93ef2c7ebf75ad6c662a98726c286dd3f87e)

### DIFF
--- a/tools/packaging/release/update-repository-version.sh
+++ b/tools/packaging/release/update-repository-version.sh
@@ -68,7 +68,6 @@ generate_kata_deploy_commit() {
 kata-deploy files must be adapted to a new release.  The cases where it
 happens are when the release goes from -> to:
 * main -> stable:
-  * kata-deploy / kata-cleanup: change from \"latest\" to \"rc0\"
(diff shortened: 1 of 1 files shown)


snir911:
If if followed correctly it fine with me, however, i wonder if the "stable" tag is ever used
also minor typo in the commit msg s/incositency/inconsistency/

snir911 (approved):
Thanks, I missed that's a backport /o\

Issue #3976, opened 2022-03-28
Proposal: add a webhook component for Kata Containers pods


liubin:
As well known, we have some issues about the limitation of using Kata Containers with Kubernetes, for example, hugepages support, emptydir with sizeLimit options, that all because of the existence of VM isolation layer, too many operations of host resources like storage, are handled by kubelet and only the processed info are passed to container runtimes.

And also, Kata Containers support almost all configuration items by annotations, which let users configure pods individually, so we can pass special resources configurations to Kata Containers through annotations. But writing them to pods’ spec manually is hard for use, so a new component acting as a webhook that dynamically updates annotations will help users do implicit conversions.

**Note**: this is a non-core, reference implementation component. 

![image](https://user-images.githubusercontent.com/1212008/160363355-bcec29c7-3cca-474b-94b1-8ae32d0b7e2a.png)

With this webhook, we can add some pre-process for known Kubernetes issues, for example:

- sizeLimit emptyDir: can parse the size option and pass it to Kata Containers.
- hugepages: can calculate total hugepages memory and use the `enable_static_resource_management` to set up hugepages at sandbox starts.

## Related issues:

- #3839
- #3765

c3d:
Hey @liubin where are we on that front? I just reopened #3765 because AFAIU it's still a problem. But what you did, if I understand correctly, is another approach that would also solve a problem I was addressing in a separate thread about pod overhead. So I wanted to check if this is still relevant, alive, or if I am missing something.

Also, am I correct that the problem is essentially the same for Go and Rust runtimes.

PR #3978, opened 2022-03-28
add a webhook component for Kata Containers pods

liubin:
See https://github.com/kata-containers/kata-containers/issues/3976


snir911:
nice! would it make sense to name it "annotations-webhook" or similar? as we also have the kata-webhook under the tests repo, ATM it's just a bit confusing but if we'll merge the repos it would be more confusing :p

liubin:
@snir911 actually this webhook is based on the webhook in the `tests` repo, it seems that the webhook only adds the kata runtime class to Pod. Of course, we could combine the two into one, although I think that one is mainly for testing purposes

src/tools/kata-webhook/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+
+	annoationValue, err := eds.String()
+	if err != nil {
+		return nil, err
+	}
+	// pod.Annotations["io.katacontainers.sandbox.volumes.emptydir"] = annoationValue
```
amshinde, line 84: You may want to get rid of this commented code.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1341,7 +1341,7 @@ func (k *kataAgent) handleHugepages(mounts []specs.Mount, hugepageLimits []specs
 
 // handleEphemeralStorage handles ephemeral storages by
 // creating a Storage from corresponding source of the mount point
-func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) ([]*grpc.Storage, error) {
+func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount, annoations map[string]string) ([]*grpc.Storage, error) {
```
amshinde, line 1344: s/annoations/annotations/

amshinde:
@liubin Approach looks good to me as it does solve the problem of respecting a sizeLimit provided by kubernetes, something that is missing today. 
But I do feel we should address this inherent issue in the long run at the CRI level by passing the sizeLimit information down the stack. I think there was some work being  done around this couple of years back, but I cant seem to find it. The only issue related to this that I had opened a while back about this : https://github.com/kata-containers/runtime/issues/2631

Would like to hear from others as well.

egernst:
I have similar thoughts to what @amshinde already shared.

Ideally, we wouldn't need annotations at all, and if we did need/consume them, it'd be provided by the the CRI implementor via Kubelet when generating the config/spec, rather than via a web hook.

Do we have a list of the current limitations / gaps in support for features / parameters that aren't available to us from Kubelet/CRI? I feel like we're ad hoc working on each one to date (ie, see fsgroup support, direct volume support).

arronwy:

src/tools/kata-webhook/main.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	var port string
+
+	if cfg.TLSEnabled() {
+		port = ":443"
+		logger.Infof("Listening on %s", port)
+		err = http.ListenAndServeTLS(port, cfg.certFile, cfg.keyFile, whHandler)
+	} else {
+		port = ":8080"
```
jodh-intel, line 148: [security]: Do we really want to listen on all interfaces?

fidencio:
@liubin, we're closing this PR as it's been opened and inactive for more than one year.

Issue #3982, opened 2022-03-28
Automate SEV deployment


Alex-Carter01:
Using CCv0 for AMD SEV requires many steps of configuration that can be automated via kata-deploy.

Details of the requirements can be found at:
https://github.com/confidential-containers/documentation/tree/main/demos/sev-demo

In summary: 

- [ ] OVMF - AmdSev firmware build

- [ ] Kernel - efi_secret patch

- [ ] Initrd - efi_secret module, AA built with offline_sev_kbc, and an agent config file with the right KBC

- [ ] Shim - PR for the shim and a number of configuration file changes

- [ ] KBS - You need to provision keys.json and keysets.json and install the gop-client.

- [ ] Image - You need to encrypt and image using the sample KBS

Alex-Carter01:
Proposed labels:

- **area/kata-deploy**

- **area/confidential-containers**

stevenhorsman:
I'm guessing this was long resolved, but SEV is removed now, so closing this

PR #3983, opened 2022-03-28
kata-deploy: add support for building sev firmware

Alex-Carter01:
### SEV with secret injection requires the AmdSev firmware build, adding this build to kata-deploy.

This PR is intended to be one part of the several needed for kata-deploy to handle setting up a node to use AMD SEV.

Fixes: #3982

Details of the required components can be found at:

https://github.com/confidential-containers/documentation/tree/main/demos/sev-demo

AMD SEV support in kata-deploy roughly breaks down into the six areas below. I am aware that @fidencio is in favor of having kata-deploy patches be unified, so my intention would be to update this PR with each additional component. 

- [x]  OVMF

- [ ] Kernel

- [ ] Initrd

- [ ] Shim

- [ ] KBS

- [ ] Image

An initial question: In `kata-deploy.sh` proper, in **configure_kata()**, there currently is a CONFIGURE_CC variable, I would propose introducing a similar option for SEV (perhaps CONFIGURE_SEV ?). I am currently checking for if the cpu has an sev flag, but this does not necessarily mean the user wants to enable SEV. Would this be in line with the expectation for how a user would choose to enable SEV?

### DIFF
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -54,6 +54,9 @@ rootfs-initrd-tarball:
 shim-v2-tarball:
 	${MAKE} $@-build
 
+ovmf-tarball:
+ 	${MAKE} $@-build
+
 merge-builds:
 	$(MK_DIR)/kata-deploy-merge-builds.sh build
 
@@ -62,3 +65,7 @@ install-tarball:
 
 image: kata-tarball
 	$(MK_DIR)kata-deploy-build-and-upload-image.sh $(CURDIR)/kata-static.tar.xz
+
+sev-tarballs: kernel-tarball qemu-tarball rootfs-initrd-tarball shim-v2-tarball ovmf-tarball
+
+sev: sev-tarballs-parallel merge-builds
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -15,8 +15,11 @@ endef
 
 kata-tarball: | all-parallel merge-builds
 
-all-parallel:
-	${MAKE} -f $(MK_PATH) all -j$$(( $$(nproc) - 1  )) NO_TTY="true" V=
+$(MK_DIR)/dockerbuild/install_yq.sh:
+	$(MK_DIR)/kata-deploy-copy-yq-installer.sh
+
+%-parallel: $(MK_DIR)/dockerbuild/install_yq.sh
+	${MAKE} -f $(MK_PATH) $(subst -parallel,,$@) -j$$(( $$(nproc) - 1  ))  V=
 
 all: cloud-hypervisor-tarball \
 	firecracker-tarball \
@@ -55,7 +58,7 @@ shim-v2-tarball:
 	${MAKE} $@-build
 
 ovmf-tarball:
- 	${MAKE} $@-build
+	${MAKE} $@-build
 
 merge-builds:
 	$(MK_DIR)/kata-deploy-merge-builds.sh build
--- a/tools/packaging/kata-deploy/local-build/dockerbuild/Dockerfile
+++ b/tools/packaging/kata-deploy/local-build/dockerbuild/Dockerfile
@@ -32,6 +32,10 @@ RUN apt-get update && \
   apt-get install --no-install-recommends -y \
   build-essential \
   cpio \
+  uuid-dev \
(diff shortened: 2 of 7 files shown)


fidencio:
@Alex-Carter01, I think those patches could and should be split in a better way so we can, already, have things on `main`.
I'd like to see, for instance:
* Add support for building and installing OVMF.
* Hold a little bit the patch for the tarball / kata-deploy.

The reason for that is because we need to have a conversation and agree on how we approach the kata-tarball and part of the release process.  We can have that as part of a meeting, or even in an issue upstream, but we need to have that conversation.

Does this sound reasonable?

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -62,3 +65,7 @@ install-tarball:
 
 image: kata-tarball
 	$(MK_DIR)kata-deploy-build-and-upload-image.sh $(CURDIR)/kata-static.tar.xz
+
+sev-tarballs: kernel-tarball qemu-tarball rootfs-initrd-tarball shim-v2-tarball ovmf-tarball
+
+sev: sev-tarballs-parallel merge-builds
```
fidencio, line 71: This looks weird to me.
Do we already scripts for building sev?  Do we need a specific target for building sev?

Alex-Carter01:
@fidencio Sure, I see no issue with splitting this into separate parts. It would be straightforward to compartmentalize everything needed to build/install ovmf starting from an 'ovmf-tarball' make target. In my mind this is the most significant part of this PR.

 I understand how the target for the sev tarball/bundle could be a separate matter, so it would be fine to separate that. 

Similarly, I would definitely appreciate some input on the best way to handle the change to kata-deploy.

I am always happy to discuss what changes would be best. My intention of submitting this as a draft PR was to get some early feedback and show the general direction we are thinking about for contributions in this area. It sounds like you have talked with Tobin to set up separate time to meet given the full agendas for the Thursday meetings, so I will be ready to talk.

fidencio:
> Similarly, I would definitely appreciate some input on the best way to handle the change to kata-deploy.
> 
> I am always happy to discuss what changes would be best. My intention of submitting this as a draft PR was to get some early feedback and show the general direction we are thinking about for contributions in this area. It sounds like you have talked with Tobin to set up separate time to meet given the full agendas for the Thursday meetings, so I will be ready to talk.

I created an issue for that (https://github.com/kata-containers/kata-containers/issues/3992) and the first steps can be discussed there, no need to have a formal meeting.  Once we agree on the steps and start splitting tasks, we can have a meeting with the small group of people working there on that, and report back in the confidential containers meetings.

Alex-Carter01:
Sounds good, I will keep following that issue and be ready to discuss tasks.

tools/packaging/kata-deploy/local-build/Makefile
Alex-Carter01: I added a new target for SEV since I didn't think other use cases would need OVMF. I am open to adding ovmf to other images, but just was trying to avoid potentially unnecessary dependencies. I definitely see how this fits into the discussion on #3992, so I am curious how that pans-out.

arronwy:

jodh-intel:
Hi @Alex-Carter01 - please could you rebase to resolve the conflict?

Also, looking at the "[check for sev flag in cpuinfo](https://github.com/kata-containers/kata-containers/pull/3983/commits/59bc1f35b0e18aaa0c1542882fcbfb2fab1e480b)" commit, I'm not seeing any code to _check for the sev flag in cpuinfo_...? :smile:

Alex-Carter01:
> Hi @Alex-Carter01 - please could you rebase to resolve the conflict?
> 
> Also, looking at the "[check for sev flag in cpuinfo](https://github.com/kata-containers/kata-containers/pull/3983/commits/59bc1f35b0e18aaa0c1542882fcbfb2fab1e480b)" commit, I'm not seeing any code to _check for the sev flag in cpuinfo_...? 😄

Sure, I can go resolve the conflict.

Whoops! Good catch, the name of that commit is a holdover from how I was originally going to look for sev capability, but am now planning to have this change enabled with a CONFIGURE_SEV flag, but I failed to update the commit title during my rebase. I can fix the commit title to something more correct.

wainersm:
@Alex-Carter01 I'm kind of still  lost regarding sev kernel/firmware/qemu build. Is this PR still needed or we can close it?

PR #3985, opened 2022-03-28
agent: add tests for do_write_stream function

### LINKED ISSUE #3984: agent tests: Add tests for do_write_stream function in src/rpc.rs
Add tests for `AgentService::do_write_stream` function in the `src/rpc.rs` file in the agent. Related to #292.

bradenrayhorn:
Add test coverage for do_write_stream function of AgentService
in src/rpc.rs. Includes minor refactoring to make function more
easily testable.

Fixes #3984

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -85,6 +85,7 @@ use std::path::PathBuf;
 const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
 
+const ERR_CANNOT_GET_WRITER: &str = "Cannot get writer";
 const ERR_INVALID_BLOCK_SIZE: &str = "Invalid block size";
 
 // Convenience macro to obtain the scope logger
@@ -577,7 +578,7 @@ impl AgentService {
             }
         };
 
-        let writer = writer.ok_or_else(|| anyhow!("cannot get writer"))?;
+        let writer = writer.ok_or_else(|| anyhow!(ERR_CANNOT_GET_WRITER))?;
         writer.lock().await.write_all(req.data.as_slice()).await?;
 
         let mut resp = WriteStreamResponse::new();
@@ -1880,7 +1881,7 @@ mod tests {
     use super::*;
     use crate::protocols::agent_ttrpc::AgentService as _;
     use oci::{Hook, Hooks};
-    use tempfile::tempdir;
+    use tempfile::{tempdir, TempDir};
     use ttrpc::{r#async::TtrpcContext, MessageHeader};
 
     // Parameters:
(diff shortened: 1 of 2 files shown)


src/agent/src/rpc.rs
```diff
@@ -2016,6 +2055,140 @@ mod tests {
...
+                data: vec![],
+                result: Ok(WriteStreamResponse {
+                    len: 0,
+                    ..WriteStreamResponse::default()
+                }),
+                ..TestData::default()
+            },
+            TestData {
+                break_pipe: true,
+                result: Err(anyhow!("Broken pipe (os error 32)")),
```
bradenrayhorn, line 2114: One point I'd like feedback on is whether we should be asserting error messages that come from the operating system. It's something that might have the potential to change outside of the control of this repository.

src/agent/src/rpc.rs
```diff
@@ -2016,6 +2055,140 @@ mod tests {
...
+                    &oci::Process::default(),
+                    &exec_process_id.to_string(),
+                    false,
+                    1,
+                )
+                .unwrap();
+
+                let fd = {
+                    if d.has_fd {
+                        let (rfd, wfd) = unistd::pipe().unwrap();
```
bradenrayhorn, line 2151: Also, is it worth adding code to close these file descriptors within the test? Or is it okay to let the OS handle these when the test runner finishes?

src/agent/src/rpc.rs
fidencio: I am not that comfortable doing this, but if we rely on that for something else already, then we better ensure a change on the message won't also break the "something else" relying on this.

src/agent/src/rpc.rs
fidencio: Please, close those within the test. :-).

src/agent/src/rpc.rs
bradenrayhorn: I agree, I'm not too comfortable with it either. I've removed it for now. I'm open to ideas for how to test the error case of [this line](https://github.com/bradenrayhorn/kata-containers/blob/test-do-write-stream/src/agent/src/rpc.rs#L588). The ideas I've got seem like they'd complicate the test more than the value it'd provide.

jodh-intel: The best solution is to convert the `std::io::Error` to an O/S error (numeric `errno`) value in the test. You could do something like this in the test code (no error checking below but just shows the concept):

```rust
let result: std::io::Error = some_function();

// Convert the Error to a numeric value
let errno_value = e.raw_os_error().unwrap();

assert_eq!(errno_value, libc::EPIPE, "{}", msg);
```

bradenrayhorn: Thanks for the info @jodh-intel, I've updated the test to make its assertions in a similar method.

jodh-intel (commented):
Thanks @bradenrayhorn.

I'm seeing `test_do_write_stream()` fail locally though:

```
$  export test=rpc::tests::test_do_write_stream; cargo test "$test" -vv -- --exact --nocapture
running 1 test
thread 'rpc::tests::test_do_write_stream' panicked at 'called `Result::unwrap()` on an `Err` value: EINVAL: Invalid argument', src/rpc.rs:1938:14
test rpc::tests::test_do_write_stream ... FAILED

failures:

failures:
    rpc::tests::test_do_write_stream
```

bradenrayhorn:
Thanks @jodh-intel, I'll be honest I'm a bit confused by the error you are seeing. In the commit you reviewed, [line 1938](https://github.com/kata-containers/kata-containers/commit/65ecfa342597cff9296d658b34c36afadebc34a2#diff-f87c1c212bba3abd0b07bafe69d5b01ad00dfe2c01bed585e2a75fcbb6442b89R1938) is not unwrapping anything, so I'm not sure where exactly this error is coming from.

jodh-intel (approved):
Thanks @bradenrayhorn.

lgtm

btw - The error I was seeing was caused by a bad install of rust fwics.

liubin (approved):

bradenrayhorn:
I'm not sure why the metrics CI is failing. @jodh-intel or @liubin, if you have time, do you have any idea? Do we just need to retest again?

liubin:
> I'm not sure why the metrics CI is failing. @jodh-intel or @liubin, if you have time, do you have any idea? Do we just need to retest again?

Don't worry about it, eventually, it will pass, we can retry it.

Issue #3986, opened 2022-03-29
release: kata-deploy commit generation is broken


bergwolf:
When bumping from `-rc` to a stable release, `update-repository-version.sh` tries to create a standalone commit but its replacement sed is broken because it tries to bump from `current_version` tag and non-stable releases do not have kata-deploy tags.

```
$./update-repository-version.sh -p "$NEW_VERSION" "$BRANCH"
INFO: Update kata-containers to version 2.4.0
INFO: remote: https://github.com/kata-containers/kata-containers.git
From https://github.com/kata-containers/kata-containers
 * branch              stable-2.4 -> FETCH_HEAD
Branch '2.4.0-branch-bump' set up to track remote branch 'stable-2.4' from 'origin'.
Switched to a new branch '2.4.0-branch-bump'
INFO: Updating VERSION file
diff --git a/VERSION b/VERSION
index cbc70e35..197c4d5c 100644
--- a/VERSION
+++ b/VERSION
@@ -1 +1 @@
-2.4.0-rc0
+2.4.0
INFO: Updating kata-deploy / kata-cleanup image tags
diff --git a/VERSION b/VERSION
index cbc70e35..197c4d5c 100644
--- a/VERSION
+++ b/VERSION
@@ -1 +1 @@
-2.4.0-rc0
+2.4.0
INFO: Creating the commit with the kata-deploy changes
On branch 2.4.0-branch-bump
Your branch is up to date with 'origin/stable-2.4'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   VERSION

no changes added to commit (use "git add" and/or "git commit -a")
```

PR #3987, opened 2022-03-29
kata-deploy: fix version bump from -rc to stable

bergwolf:
In such a case, we should bump from the "latest" tag rather than from
current_version.

Fixes: #3986


egernst (dismissed):

fidencio (approved):
lgtm, thanks @bergwolf1

snir911 (approved):
LGTM, thanks, FWIW AFAIU it's not tested by /test_kata_deploy

bergwolf:
@snir911 
> FWIW AFAIU it's not tested by /test_kata_deploy

I did trigger `/test_kata_deploy`. Do you mean the comment command does not work?

snir911:
@bergwolf it should be with underscore IIRC but i think the update-repository-version.sh script will not be tested by it

PR #3988, opened 2022-03-29
kata-deploy: fix version bump from -rc to stable

bergwolf:
In such a case, we should bump from the "latest" tag rather than from
current_version.

backporting #3987


bergwolf:
tagging `wip` to block on https://github.com/kata-containers/kata-containers/pull/3987

fidencio (approved):
lgtm, thanks @bergwolf!

bergwolf:
As discussed in slack, s390x CI is disabled and can be ignored for now.

teawater (approved):

Issue #3998, opened 2022-03-30
kata-monitor: help shows duplicated default for the "runtime-endpoint" parameter


fgiudici:
_(default: "/run/containerd/containerd.sock")_ is duplicated in the kata-monitor usage:
```
[root@kubernetes ~]# kata-monitor --help
Usage of kata-monitor:
  -listen-address string
        The address to listen on for HTTP requests. (default ":8090")
  -log-level string
        Log level of logrus(trace/debug/info/warn/error/fatal/panic). (default "info")
  -runtime-endpoint string
        Endpoint of CRI container runtime service. (default: "/run/containerd/containerd.sock") (default "/run/containerd/containerd.sock")

```

PR #3999, opened 2022-03-30
kata-monitor: fix duplicated output when printing usage

fgiudici:
(default: "/run/containerd/containerd.sock") is duplicated when
printing kata-monitor usage:

[root@kubernetes ~]# kata-monitor --help
Usage of kata-monitor:
  -listen-address string
        The address to listen on for HTTP requests. (default ":8090")
  -log-level string
        Log level of logrus(trace/debug/info/warn/error/fatal/panic). (default "info")
  -runtime-endpoint string
        Endpoint of CRI container runtime service. (default: "/run/containerd/containerd.sock") (default "/run/containerd/containerd.sock")

the golang flag package takes care of adding the defaults when printing
usage. Remove the explicit print of the value so that it would not be
printed on screen twice.

Fixes: #3998

### DIFF
--- a/src/runtime/cmd/kata-monitor/main.go
+++ b/src/runtime/cmd/kata-monitor/main.go
@@ -21,7 +21,7 @@ import (
 const defaultListenAddress = "127.0.0.1:8090"
 
 var monitorListenAddr = flag.String("listen-address", defaultListenAddress, "The address to listen on for HTTP requests.")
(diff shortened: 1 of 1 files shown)


fgiudici:
While the fix is not critical at all, is a bug, so I think would be nice to backport on 2.4 (applied label accordingly).

fidencio (approved):
lgtm, thanks @fgiudici!

fidencio:
/test

And please, raise the backport already

snir911 (approved):
LGTM thanks!

PR #4002, opened 2022-03-30
tools/packaging: Expose docker config to build-kata-deploy container

### LINKED ISSUE #4001: make kata-tarball hits the docker rate limit when building firecracker
# Description of problem

1. authenticate to `docker.io` with `docker login`
2. build kata artifacts with `make kata-tarball`.

# Expected result

Build succeeds.

# Actual result

It might fail in the firecracker build. And when it does, this can last for an indefinite period of time.

```
[Firecracker devtool] About to pull docker image fcuvm/dev:v17
v17: Pulling from fcuvm/dev
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
[Firecracker devtool] Error pulling docker image. Aborting.
```

This seriously damages the developer experience, hence labelling this as a bug.

# Further information

The failing `docker` comes from the `firecracker` build tool, i.e. it is running nested and doesn't have access to the user's docker auth tokens. A straightforward solution is to expose `$HOME/.docker` in the `build-kata-deploy` container.

gkurz:
`make kata-tarball` may fail consistently because it hits the
docker rate limit when building the firecracker asset. This is
especially painful when used behind the firewall of a large
organization because this aggregates all docker requests as
coming from a few IP addresses, and `make kata-tarball` may
stay broken for an indefinite period of time.

The usual mitigation for this is to have an account at docker.io
and to `docker login`. Unfortunately this doesn't work here
because the failing docker is running nested and doesn't see the
user's auth tokens.

Expose `$HOME/.docker` to the `build-kata-deploy` container so
that nested docker invocations can access the auth tokens.

Fixes #4001

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -28,6 +28,7 @@ docker build -q -t build-kata-deploy \
 
 docker run \
 	-v /var/run/docker.sock:/var/run/docker.sock \
+	-v $HOME/.docker:$HOME/.docker \
 	--user ${uid}:${gid} \
 	--env USER=${USER} -v "${kata_dir}:${kata_dir}" \
 	--rm \


fidencio (approved):
lgtm, thanks @gkurz!

gkurz:
@snir911 pointed out that passing credentials to a container isn't generally a safe thing to do, as it requires 100% trust in the container image and the commands that will run in the container. Passing the docker daemon socket as we already do is also quite unsafe BTW.

I then realized that, starting with `v0.23.4` , the firecracker `devtool` script pulls from AWS instead of docker.io. This fixes the rate limit issue without having to expose the user credentials, so I'd rather close this PR and open a new one to bump the firecracker version. WDYT @fidencio ?

fidencio:
> 

+1.

Issue #4015, opened 2022-03-31
agent: Document the configuration options


wainersm:
The `kata-agent` have a couple of configuration options and it can be configured on many ways (e.g. environment variables, argument, kernel parameters...). However that is documented nowhere; currently we only have snippets in various documents, for example, https://github.com/kata-containers/kata-containers/blob/main/docs/Developer-Guide.md#enable-agent-debug-console and https://github.com/kata-containers/kata-containers/blob/main/docs/tracing.md#enable-agent-tracing

We should have the agent configuration options proper documented.

PR #4016, opened 2022-03-31
agent: Document the configuration options

wainersm:
The information about the configuration options of the agent is spread in documents and is incomplete. Let's have it proper documented.

While here I am sending a small change to the config.rs file.

Fixes: #4015

### DIFF
--- a/src/agent/README.md
+++ b/src/agent/README.md
@@ -84,6 +84,39 @@ To install the `protoc` command on a Fedora/CentOS/RHEL system:
 $ sudo dnf install -y protobuf-compiler
 ```
 
+## Configuring the agent
+
+The agent can be configured in a couple of ways.
+
+If running under the supervision of the Kata runtime, the agent will be automatically configured according to some properties in the runtime's `configuration.toml` file (see [runtime configuration](../runtime/README.md#configuration)). For example, the `enable_debug` option under the `[agent.kata]` section can be used to set debug-level logs. However, not all the accepted configurations are mapped into `configuration.toml`, so any additional options should be passed directly to the agent via kernel parameters.
+
+Any kernel parameter is be specified in the `kernel_params` property in the runtime's `configuration.toml`. As an example, to enable the Agent support for unified cgroup hierarchy and set debug logs, write the following line in `configuration.toml`:
+
+```
+kernel_params = "agent.unified_cgroup_hierarchy=true agent.log=debug"
+```
+
+The agent options passed via kernel parameters have the `agent.OPTION` format, where `OPTION` can be:
+
+| Option | Description | Example |
+|-|-|-|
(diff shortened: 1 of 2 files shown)


src/agent/README.md
```diff
@@ -84,6 +84,33 @@ To install the `protoc` command on a Fedora/CentOS/RHEL system:
 $ sudo dnf install -y protobuf-compiler
 ```
 
+## Configuring the agent
+
+The agent can be configured in a couple of ways.
+
+If running under the supervision of the Kata runtime, the agent will be automatically configured according to some properties in the runtime's `configuration.toml` file. For example, the `enable_debug` option under the `[agent.kata]` section can be used to set debug-level logs. However, not all the accepted configurations are mapped into `configuration.toml`, so any additional options should be passed directly to the agent via kernel parameters.
```
amshinde, line 91: Would be good to point out how kernel parameters are passed.

liubin: Agree with @amshinde , it's better to give an example like `kernel_params = "agent.log=debug agent.debug_console agent.debug_console_vport=1026 agent.hotplug_timeout=10"`.

wainersm: Thanks for your comments @amshinde and @liubin ! I agree with your suggestion, I will send a v2 of this PR soon.

wainersm:
Just update this PR to address the suggestions of @amshinde and @liubin

liubin (approved):

jodh-intel:
I'm in two minds about this. Docs are good, but I'm not totally convinced we want to document these options. The official way to configure Kata is using the `configuration.toml` config file. The `kernel_params=` option is a necessary evil but users should not need to tweak most of the "behind the scenes" agent settings. If they did, we'd add a proper "knob" in the config file.

The other problem with documenting them is that we create a potential maintenance problem: it's harder to modify the behaviour of the agent settings (or even remove them).

An alternative to this PR would be to add them to the developer guide (to minimise the chance that real users (non developers) try to set these options and get into difficulties as a result).

If we **really** feel these need to be documented, I'd suggest a very clear warning message along the lines of:

> **Warning:**
>
> The official way to modify the Kata Containers agent behaviour is using the settings provided in the Kata Containers configuration file.
>
> The agent options listed here may change their behaviour between releases or they may even be removed (meaning setting them would silently fail).

The other problem with documenting the options of course is that the doc is not tied to the code so they are bound to deviate at some point.

I can see a possible compromise solution (not perfect, but maybe workable): define all the agent options in a YAML file that is parsed by the agent **at build time** and which is of course still human readable (we can add descriptions for all the options but the agnet wouldn't consume those).

fidencio:
@wainersm, we're closing this PR as it's been opened and inactive for more than one year.

Issue #4019, opened 2022-04-01
Failed to hotplug SPDK vhost-user-blk on Arm


Jaylyn-Ren:
# Description of problem
### Use Spdk vhost-user-blk in Kata on Arm:

1. Based on [run spdk vhost-user with kata](https://github.com/kata-containers/documentation/blob/master/use-cases/using-SPDK-vhostuser-and-kata.md) create vhost-user-blk and set parameters needed in the configuration. toml
    > [hypervisor.qemu]
    > enable_hugepages = true
    > enable_vhost_user_store = true
    > vhost_user_store_path = "/var/run/kata-containers/vhost-user"
    > pflashes = ["/usr/share/kata-containers/kata-flash0.img", "/usr/share/kata-containers/kata-flash1.img"]

3. The version of kata is 2.4.0. Use ctr to run kata:
    `sudo ctr run -d --runtime io.containerd.run.kata.v2 --config config.json spdk_container`
    > config.json (not complete):
    > `"root": {
      "path": "/home/linux/jaylyn/bundle/rootfs",
      "readonly": false}`
    > `"devices": [{
          "path": "/dev/vda",
          "type": "b",
          "major": 241,
          "minor": 0,
          "fileMode": 440,
          "uid": 0,
          "gid": 0
      }]`

Spdk vhost-user-blk can be hotplugged successfully for kata containers on X86, but it is failed on Arm. 

# Expected result

Kata containers can run with spdk vhost-user-blk successfully on Arm. 

# Actual result

Error occurs:  
ctr: failed to create shim: Timeout after 3s waiting for uevent VirtioBlkPciMatcher
{rex:^/devices/pci0000:00/0000:00:02.0/0000:01:01.0/virtio[0-9]+/block/ }: unknown

PR #4031, opened 2022-04-02
Virtcontainers: Enable hot plugging vhost-user-blk device on ARM

Jaylyn-Ren:
The vhost-user-blk can be hotplugged on the PCI bridge succesfullly on
X86, but failed on Arm. However, hotplugging it on Root Port as a PCIe
device can work well on ARM.
Open the "pcie_root_port" in configuration.toml is needed.

Fixes: #4019

Signed-off-by: Jaylyn Ren <jaylyn.ren@arm.com>

### DIFF
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -38,6 +38,7 @@ import (
 	pkgUtils "github.com/kata-containers/kata-containers/src/runtime/pkg/utils"
 	"github.com/kata-containers/kata-containers/src/runtime/pkg/uuid"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/device/config"
+	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/device/drivers"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/types"
 	vcTypes "github.com/kata-containers/kata-containers/src/runtime/virtcontainers/types"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/utils"
@@ -129,6 +130,8 @@ const (
 	fallbackFileBackedMemDir = "/dev/shm"
 
 	qemuStopSandboxTimeoutSecs = 15
+
+	qomPathPrefix = "/machine/peripheral/"
 )
 
 // agnostic list of kernel parameters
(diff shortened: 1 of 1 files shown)


amshinde (approved):

jongwu:
TAL @fidencio @jodh-intel

jongwu (approved):

dgibson:
@Jaylyn-Ren what hotplug mechanism are you using on ARM at the low-level?  Is it PCIe native hotplug?  ACPI hotplug?  Something else ARM specific?

If it's PCIe native there are probably still problems here - on x86 we can have ugly races between the guest kernel's routine PCI scan and the hotplug which can cause a device to appear momentarily before being removed again.  I'd expect the same problems to be present on ARM.

jongwu:
Thanks @dgibson -, I think that it's pcie native hotplug. 
Can you show me how to reproduce the issue you refer to? And is there any way to bypass that?

arronwy:

dgibson:
> Thanks @dgibson -, I think that it's pcie native hotplug. Can you show me how to reproduce the issue you refer to?

Alas, not easily.  This used to reproduce pretty reliably on x86, but I remembered that was in the days when the agent still forced a PCI rescan when it started.  That's been removed which probably makes this a bunch less likely to happen.  I suspect it's still possible but would rely on the guest's normal PCI scan during boot colliding with the hotplug being issued by the runtime.  Those things are happening essentially independently controlled by different entities (runtime vs. guest kernel), so that's basically just unhappy chance.

That said... with the agent forced rescans gone, we might be ok.  If the only guest rescan is happening during boot - specifically *before* userspace starts - then we should be ok as long as the runtime waits for a response for the agent before issuing the hotplug.  I'm not certain if that's the case or not.

Fwiw, I can describe the rough mechanics of the race.  PCIe hotplug was desgined originally for physical, operator-present hotplugging, rather than for VMs.  The mechanism as it appears to the guest is, roughly, that some registers in the PCIe bridge port change value (in hardware representing the physical presence detect) then an interrupt is issued representing the operator pressing the "attention" button once the device is latched in to trigger the actual hotplug.

If this occurs while the guest is scanning the PCI bus that attention interrupt is delayed.  Meanwhile the scan can pick up the new device along with everything else.  With the scan complete, the guest gets the attention interrupt, but since it already sees the device, it interprets this as a hot unplug request, and powers down the new device.

This is quite similar to the race with SHPC, but the window in PCIe is much, much tighter.  There's also an ugly symptom where sometimes the device can get into a physical state where configuration space doesn't respond which requires a *host* reboot to fix.  I've never figured out the exact mechanism of that though.

> And is there any way to bypass that?

Uh.. maybe.  If you can delay hotplugging until scans are done you'll be ok, but how to figure that out safely is kind of complicated (and probably not 100% safe, since anything with root in the guest could trigger a rescan).  It might be possible to fix this in qemu: while doing the rescan the guest sets a LED to blinking mode by writing a special value to a register, qemu could fail or defer hotplugs while the LED control register is in that state.  Deferring introduces some complexities  in the qemu code, and if it outright failed, then the Kata side would have to have it's own retry logic of some sort.  This whole thing is arguably a (guest) kernel bug, so there's maybe a way to fix that there - I don't think the PCI maintainers are particularly interested in looking at this though.

So.. possibly it can be fixed, but it's kind of a pain.  Which is why for x86 we decided to just require ACPI hotplug instead.  I don't know if that's an option for ARM.

jongwu:
Hello @dgibson -, thanks for your elaborate explanation. I really learn a lot.
As there is no ACPI hotplug for pci device support on QEMU/virt, we have no choice but use the PCIE native hotplug. Also @Jaylyn-Ren and I do lots of tests. It shows that the Vhost user hotplug using PCIE native is stable enough.  So, what about keep it this way first and improve it later when there are related features involved in, eg, ACPI device hotplug for QEMU/virt.

jodh-intel (approved):
Thanks @Jaylyn-Ren.

lgtm

jongwu:
Ping @dgibson ...

src/runtime/virtcontainers/qemu.go
```diff
@@ -1396,31 +1399,68 @@ func (q *qemu) hotplugAddVhostUserBlkDevice(ctx context.Context, vAttr *config.V
...
+		if err != nil {
+			return err
+		}
+
+		if err = q.qmpMonitorCh.qmp.ExecutePCIVhostUserDevAdd(q.qmpMonitorCh.ctx, driver, devID, vAttr.DevID, addr, bridgeId); err != nil {
+			return err
+		}
+
+	default:
+		addr, bridge, err := q.arch.addDeviceToBridge(ctx, vAttr.DevID, types.PCI)
```
dgibson, line 1439: Hrm.  So arguably this puts the switch on machine type in the wrong place.  The reason you need this is that despite the name, `addDeviceToBridge` is not a useful cross-machine abstraction, and in fact its interfaces are misdesigned so that it can't really be tweaked into one. A better approach long term would be a proper abstraction for placing PCI devices.  I've looked at doing this, but haven't had the time to attempt it.

I recognize that that's a much bigger refactor than you probably want to take on just to get ARM working.  So, I won't request it be changed just on that basis.  That said, a major reason that the code in this area is such a godawful mess is that people keep adding things without refactoring the base to support it cleanly, so there's that ://

dgibson (approved):

src/runtime/virtcontainers/qemu.go
jongwu: Thanks @dgibson -, it's pretty not good to do this here. However, to refactor this is beyond this PR. So, thanks for the "precious" approval!

Issue #4038, opened 2022-04-04
CCv0 | Add Cloud Hypervisor support to the `ccv0.sh` script


fidencio:
`ccv0.sh` is a really nice way to do some quick hacks, do some simple tests, or even have an environment setup for playing (as in, developing and manually testing) with the CCv0 effort.

However, the script is totally tied to QEMU as its VMM, and we should make it VMM agnostic (at least for the VMMs that are aligning themselves with the Confidential Container effort).

fidencio:
Closing as the PR got merged but, for some reason, it was not linked here.

PR #4039, opened 2022-04-04
CCv0 | ccv0.sh: Expand to also using Cloud Hypervisor

fidencio:
ccv0: Don't use the QEMU process to get the sandbox ID

Instead, rely on the conntainerd-shim-kata-v2 process, as this makes
this script VMM agnostic.

---

ccv0.sh: Expand to also using Cloud Hypervisor

Right now the script only support QEMU, but there's not a reason to do
that, mainly considering we already have the tests parity in the CIs
between QEMU and Clouud Hypervisor.

With this in mind, let's expand this script to also using Cloud
Hypervisor.

Whether this script should use QEMU or Cloud Hypervisor is defined
according to the KATA_HYPERVISOR environment variable.

Fixes: #4038

---

@stevenhorsman, please, take a look at the changes and let me know if you want something done differently.

### DIFF
--- a/docs/how-to/ccv0.sh
+++ b/docs/how-to/ccv0.sh
@@ -91,6 +91,7 @@ Commands:
 - build_and_install_all:            Build and install everything
 - build_and_install_rootfs:         Builds and installs the rootfs image
 - build_kata_runtime:               Build and install the kata runtime
+- build_cloud_hypervisor            Checkout, patch, build and install Cloud Hypervisor
 - build_qemu:                       Checkout, patch, build and install QEMU
 - configure:                        Configure Kata to use rootfs and enable debug
 - connect_to_ssh_demo_pod:          Ssh into the ssh demo pod, showing that the decryption succeeded
@@ -127,7 +128,18 @@ build_and_install_all() {
     create_a_local_rootfs
     build_and_install_rootfs
     install_guest_kernel_image
-    build_qemu
+    case "$KATA_HYPERVISOR" in
+        "qemu") 
+            build_qemu
+            ;;
+        "cloud-hypervisor") 
+            build_cloud_hypervisor
+            ;;
+        *)
+            echo "Invalid option: $KATA_HYPERVISOR is not supported." >&2
+            ;;
+    esac
+
     check_kata_runtime
     if [ "${KUBERNETES}" == "yes" ]; then
         init_kubernetes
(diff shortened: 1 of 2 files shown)


docs/how-to/ccv0.sh
```diff
@@ -467,7 +467,7 @@ run_kata_and_capture_logs() {
 
 get_ids() {
     guest_cid=$(sudo ss -H --vsock | awk '{print $6}' | cut -d: -f1)
-    sandbox_id=$(ps -ef | grep qemu | egrep -o "sandbox-[^,][^,]*" | sed 's/sandbox-//g' | awk '{print $1}')
+    sandbox_id=$(ps -ef | grep containerd-shim-kata-v2 | egrep -o "id [^,][^,].* " | awk '{print $2}')
```
wainersm, line 490: This looks correct if you want to make it hypervisor-agnostic. At least I don't know if there is a better way.

Keep in mind that assume you have only one shim process running...which might not be true. To be fair, the old version of the code had the same assumption.

wainersm:
@fidencio this PR looks good to me

docs/how-to/ccv0.sh
fidencio: Exactly.  The old code was assuming only one QEMU process running.
And if you happen to know a better way to do so, I'm here to learn. :-)

dborquez (approved):
LGTM, thanks @fidencio

fidencio:
Please, let's hold this one as I had a private convo with @stevenhorsman and he spot one issue and made at least one suggestion.  I will re-work this so he can review and put his stamp Tomorrow.

fidencio:
@stevenhorsman, please, remove the `do-not-merge` label after your review. :-)

stevenhorsman (commented):
Thanks for the updates and fixing some of my bad English! I was thinking it might be useful to add some CLH instructions at the beginning e.g.
```
diff --git a/docs/how-to/how-to-build-and-test-ccv0.md b/docs/how-to/how-to-build-and-test-ccv0.md
index 1f5ea11e4..c6976ca22 100644
--- a/docs/how-to/how-to-build-and-test-ccv0.md
+++ b/docs/how-to/how-to-build-and-test-ccv0.md
@@ -44,6 +44,11 @@ $ chmod u+x ccv0.sh
       ```
       `skopeo` is
       required for passing source credentials and verifying container image signatures using the kata agent.
+    - By default the build and configuration using QEMU as the hypervisor. In order to use Cloud Hypervisor instead set:
+       ```
+       $ export KATA_HYPERVISOR="cloud-hypervisor"
+       ```
+       before running the build.
 
 - At this point you can provision a Kata confidential containers pod and container with either
   [`crictl`](#using-crictl-for-end-to-end-provisioning-of-a-kata-confidential-containers-pod-with-an-unencrypted-image),
``` 
I'm not sure what you think of that idea?

fidencio:
@stevenhorsman, done, thanks for the suggestion!

stevenhorsman (approved):
LGTM and I think it includes all the patches I tested with yesterday. Thanks!

stevenhorsman:
I think this can be merged now, but it's worth noting that until https://github.com/kata-containers/tests/pull/4601/commits/8edc78074b69c5e52f7189a206a84ff7482673a5 is merged into `CCv0` then the configuration will need modifying to point to `/opt/kata/bin/cloud-hypervisor` rather than `/usr/bin/cloud-hypervisor`

fidencio:
Just re-pushed to make sure the check for know-words is happy (it seems to not be happy about QEMU).

PR #4042, opened 2022-04-05
fsGroup support for direct-assigned volume

### LINKED ISSUE #4018: fsGroup support for direct-assigned volume

direct-assigned volume

**How can it be improved?**

When using Kata with K8s with direct-assigned volume, there isn't a good way to honor the `fsGroup` specified by pod security context since the `fsGroup` change is [done by Kubelet today](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_mounter.go#L283) after CSI `NodePublishVolume` is called.

Although there is an option to defer `fsGroup` to the CSI driver from this [KEP](https://github.com/gnufied/enhancements/blob/master/keps/sig-storage/2317-fsgroup-on-mount/README.md) but its use case is mostly intended for plugins (like SMB) that does not support fsGroup but does similar with mount options.

Thus, right now the CSI client call ctx for CSI APIs are very short at [2 minutes](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_plugin.go#L53)
so if we use the defer `fsGroup` to CSI driver, the recursive group change can take quite a bit of time for large number of files and can cause the client to timeout.

I propose we enhance the direct-assign volume feature to add additional options to specify `fsGroup` and `fsGroupChangePolicy` as part of the `Metadata` info in `MountInfo` struct [here](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/direct-volume/utils.go#L33)
and have those fields made available to the kata agent. The agent will do the optional recursive group owner change after the filesystem has been mounted.

**Additional Information**

I think we can add these two fields to the current `agent.proto` under [Storage](https://github.com/kata-containers/kata-containers/blob/main/src/libs/protocols/protos/agent.proto#L404)

And then in the kata-agent code, in `common_storage_handler` handler, after performing `mount_storage()`, we can add the additional logic to perform the group owner change.

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)? Yes

yibozhuang:
This change adds support for allowing CSI plugins to defer
setting fsGroup to inside the guest by kata-agent for
direct-assigned volume.

The fsGroup will be specified by the fsGroup key in the direct-assign
MountInfo metadata field.

On the agent protobuf, adding two fields `FSGroup` and `FSGroupChangePolicy` 
to allow the shim to provide the group id as well as the policy to perform
the recursive directory ownership change to the agent.

The agent will perform the recursive fsGroup change after the mount
has been completed and also update the permission to have the
SetGid bit set for directories.

Fixes #4018 


src/runtime/virtcontainers/container.go
```diff
@@ -639,6 +639,22 @@ func (c *Container) createBlockDevices(ctx context.Context) error {
...
+					gid, err := strconv.Atoi(value)
+					if err != nil {
+						c.Logger().WithError(err).Error(fmt.Sprintf("invalid group id provided for key %s", volume.FSGroupMetadataKey))
+						continue
+					}
+					c.mounts[i].FSGroup = uint32(gid)
+				case volume.FSGroupChangePolicyMetadataKey:
+					c.mounts[i].FSGroupChangePolicy = volume.FSGroupChangePolicy(value)
+				default:
+					c.Logger().Debugf("Ignoring unsupported direct-assignd volume metadata key: %s, value: %s", key, value)
```
liubin, line 655: Nit: Log level should be at least warning

src/runtime/virtcontainers/container.go
```diff
@@ -639,6 +639,22 @@ func (c *Container) createBlockDevices(ctx context.Context) error {
 			c.mounts[i].Type = mntInfo.FsType
 			c.mounts[i].Options = mntInfo.Options
 			c.mounts[i].ReadOnly = readonly
+
+			for key, value := range mntInfo.Metadata {
+				switch key {
+				case volume.FSGroupMetadataKey:
+					gid, err := strconv.Atoi(value)
+					if err != nil {
+						c.Logger().WithError(err).Error(fmt.Sprintf("invalid group id provided for key %s", volume.FSGroupMetadataKey))
```
liubin, line 648: Please use `Errorf()` instead of `Error(fmt.Sprinf())`.

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
     )
 }
 
+#[instrument]
+pub fn set_ownership(logger: &Logger, storage: &Storage) -> Result<()> {
+    let logger = logger.new(o!("subsystem" => "change_ownership"));
```
liubin, line 586: Here may be some miss-use of `subsystem` if log fields in this file(mount.rs). Suggest keep subsystem to `mount`  and set func to it's name by using `o!("subsystem" => "mount", "fn" => "set_ownership")`.

```rust
src/agent/src/mount.rs:150:    let logger = logger.new(o!("subsystem" => "baremount"));
src/agent/src/mount.rs:531:    let logger = logger.new(o!("subsystem" => "mount"));
src/agent/src/mount.rs:728:    let logger = logger.new(o!("subsystem" => "mount"));
src/agent/src/mount.rs:868:    let logger = logger.new(o!("subsystem" => "mount"));
```

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+    if opts_vec.contains(&String::from("ro")) {
+        read_only = true;
+    }
+
+    let mount_path = Path::new(&storage.mount_point);
+
+    let mut skip_ownership_change = false;
+    if storage.fs_group_change_policy == FSGroupChangePolicy::OnRootMismatch {
+        if mount_path.metadata()?.gid() == storage.fs_group {
+            let mut mask = RW_MASK;
```
liubin, line 604: Nit: Here can use Rust `let x = if (y){ x1} else {x2}`, but since `clippy` doesn't complain it, don't matter.

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+    let opts_vec: Vec<String> = storage.options.to_vec();
+    if opts_vec.contains(&String::from("ro")) {
+        read_only = true;
+    }
+
+    let mount_path = Path::new(&storage.mount_point);
+
+    let mut skip_ownership_change = false;
+    if storage.fs_group_change_policy == FSGroupChangePolicy::OnRootMismatch {
+        if mount_path.metadata()?.gid() == storage.fs_group {
```
liubin, line 603: Here `mount_path.metadata()?.gid()` and line 614 `mount_path.metadata()?.permissions().mode()` both call `mount_path.metadata()?`, suggest saving `mount_path.metadata()` to a local variable.

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+            // gid of the mount path root directory matches the desired gid
+            // and the current permission of mount path root directory is correct,
+            // then ownership change will be skipped.
+            let current_mode = mount_path.metadata()?.permissions().mode();
+            if (mask & current_mode == mask) && (current_mode & MODE_SETGID != 0) {
+                info!(logger, "skipping ownership change for volume";
+                "mount-destination" => mount_path.to_str(),
+                "fs-group" => storage.fs_group.to_string(),
+            );
+                skip_ownership_change = true
```
liubin, line 620: Here directly `return Ok(());` can save a `skip_ownership_change` variable?

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+        )?;
+    }
+
+    Ok(())
+}
+
+#[instrument]
+pub fn recursive_ownership_change(
+    path: &Path,
+    uid: Option<Uid>,
```
liubin, line 643: It seems the `uid` is not used, is it only a placeholder?

liubin:
Thank you @yibozhuang, LGTM.

This PR adds two additional fields to the `Storage` struct, which will make manipulating fsgroup clearly than putting them in the `options` field.

Maybe shift the old mechanism to this one if possible for this workaround-like code:

https://github.com/kata-containers/kata-containers/blob/b39caf43f1486ab503c645f1597c72bcf96d1a02/src/agent/src/mount.rs#L204-L211


/cc @fidencio @jodh-intel @lifupan @bergwolf

src/agent/src/mount.rs
ddebroy, line 603: How about logging an error if the metadata reads fail for some reason?

src/runtime/pkg/direct-volume/utils.go
```diff
@@ -17,6 +17,19 @@ import (
 
 const (
 	mountInfoFileName = "mountInfo.json"
+
+	FSGroupMetadataKey             = "fsGroup"
+	FSGroupChangePolicyMetadataKey = "fsGroupChangePolicy"
+)
+
+type FSGroupChangePolicy string
```
ddebroy, line 25: Should there be a comment here indicating that `FSGroupChangePolicy` and its values below roughly tracks https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/api/core/v1/types.go#L3244 and it is up to the CSI plugin to either directly map to the corresponding fields in a pod or specify the most optimal setting?

ddebroy (commented):
The logic around passing and handling of  `fsGroup` and `fsGroupChangePolicy` LGTM. A couple of minor comments around comments and logs.

src/runtime/virtcontainers/container.go
```diff
@@ -639,6 +639,22 @@ func (c *Container) createBlockDevices(ctx context.Context) error {
...
+						continue
+					}
+					c.mounts[i].FSGroup = uint32(gid)
+				case volume.FSGroupChangePolicyMetadataKey:
```
fengwang666, line 652: Should we check the error case when `FSGroupMetadataKey` is missing and `FSGroupChangePolicyMetadataKey` exists?

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+            // With fsGroup change policy to OnRootMismatch, if the current
+            // gid of the mount path root directory matches the desired gid
+            // and the current permission of mount path root directory is correct,
+            // then ownership change will be skipped.
+            let current_mode = mount_path.metadata()?.permissions().mode();
+            if (mask & current_mode == mask) && (current_mode & MODE_SETGID != 0) {
+                info!(logger, "skipping ownership change for volume";
+                "mount-destination" => mount_path.to_str(),
+                "fs-group" => storage.fs_group.to_string(),
+            );
```
fengwang666, line 619: nit: indentation looks off

src/agent/src/mount.rs
fengwang666, line 620: Why not just return here?

src/agent/src/mount.rs
```diff
@@ -573,6 +581,92 @@ fn mount_storage(logger: &Logger, storage: &Storage) -> Result<()> {
...
+    }
+
+    Ok(())
+}
+
+#[instrument]
+pub fn recursive_ownership_change(
+    path: &Path,
+    uid: Option<Uid>,
+    gid: Option<Gid>,
```
fengwang666, line 644: What's the reason to use an option for `gid`?

fengwang666 (commented):
looks good to me. A few minor comments.

src/runtime/virtcontainers/container.go
yibozhuang: Thanks, fixed.

src/runtime/virtcontainers/container.go
yibozhuang: Thanks, done.

src/agent/src/mount.rs
yibozhuang: thanks, changed to `logger.new(o!("subsystem" => "mount", "fn" => "set_ownership"))`

src/agent/src/mount.rs
yibozhuang: Thanks, changed.

src/agent/src/mount.rs
yibozhuang: Thanks, fixed.

src/agent/src/mount.rs
yibozhuang: Sure, added `map_err` to add the additional logging

```rust
    let metadata = mount_path.metadata().map_err(|err| {
        error!(logger, "failed to obtain metadata for mount path";
            "mount-path" => mount_path.to_str(),
            "error" => err.to_string(),
        );
        err
    })?;
```

src/agent/src/mount.rs
yibozhuang: Thanks, fixed.

src/runtime/pkg/direct-volume/utils.go
yibozhuang: Thanks, added comment indicating `FSGroupChangePolicy` and values track the `PodFSGroupChangePolicy` defined in:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1469,6 +1479,10 @@ func (k *kataAgent) handleDeviceBlockVolume(c *Container, m Mount, device api.De
 	if len(vol.Options) == 0 {
 		vol.Options = m.Options
 	}
+	if m.FSGroup > 0 {
+		vol.FsGroup = m.FSGroup
+	}
+	vol.FsGroupChangePolicy = getFSGroupChangePolicy(m.FSGroupChangePolicy)
```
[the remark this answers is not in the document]
yibozhuang: Thanks, fixed.

src/runtime/virtcontainers/container.go
yibozhuang: Thanks, added additional check when `FSGroupChangePolicyMetadataKey` exists but `FSGroupMetadataKey` is missing to log error and continue

src/agent/src/mount.rs
yibozhuang: thanks, fixed.

src/agent/src/mount.rs
yibozhuang: Thanks, fixed.

src/agent/src/mount.rs
yibozhuang: Was hoping to make `recursive_ownership_change` generic that it can also be used for changing `uid` so basically the input parameter to this function matches with [nix](https://docs.rs/nix/0.9.0/nix/index.html)::[unistd](https://docs.rs/nix/0.9.0/nix/unistd/index.html)::[chown](https://docs.rs/nix/0.9.0/nix/unistd/fn.chown.html) input parameter.

src/agent/src/mount.rs
yibozhuang: Was hoping to make the input parameter match with [nix](https://docs.rs/nix/0.9.0/nix/index.html)::[unistd](https://docs.rs/nix/0.9.0/nix/unistd/index.html)::[chown](https://docs.rs/nix/0.9.0/nix/unistd/fn.chown.html) api input parameter so callers using this function can provide the same input parameter as if calling `nix::unistd::chown` but it does the recursive walk.

Using `Option` here allows caller to provide `None` for `uid` or `gid` when calling this function to indicate that the `uid` or `gid` should not be modified.

yibozhuang:
> Maybe shift the old mechanism to this one if possible for this workaround-like code:
> 
> https://github.com/kata-containers/kata-containers/blob/b39caf43f1486ab503c645f1597c72bcf96d1a02/src/agent/src/mount.rs#L204-L211
> 

Yes, I was hoping that moving the `ephemeral_storage_handler` and `local_storage_handler` to leverage the new `fs_group` rather than parsing it out from `options` field can be a separate PR?

Will file an issue to work on once this change goes in.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1469,6 +1479,10 @@ func (k *kataAgent) handleDeviceBlockVolume(c *Container, m Mount, device api.De
 	if len(vol.Options) == 0 {
 		vol.Options = m.Options
 	}
+	if m.FSGroup > 0 {
```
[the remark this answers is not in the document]
yibozhuang: That's a great point!
I've updated the `agent.proto` to create a `FSGroup` message

```proto
// FSGroup consists of the group id and group ownership change policy
// that a volume should have its ownership changed to.
message FSGroup {
	// GroupID is the ID that the group ownership of the
	// files in the mounted volume will need to be changed to.
	uint32 group_id = 2;
	// GroupChangePolicy specifies the policy for applying group id
	// ownership change on a mounted volume.
	types.FSGroupChangePolicy group_change_policy = 3;
}
```

and have Storage use this for the `fs_group` field. This will allow the field to be optional
and in `mount.rs`, we check if this field is set and will perform the chown on the group id.

This should cover the case of user setting fsGroup to some value and then setting it back to 0 (root) after.

Thanks for spotting this.

liubin:
> 
> Yes, I was hoping that moving the `ephemeral_storage_handler` and `local_storage_handler` to leverage the new `fs_group` rather than parsing it out from `options` field can be a separate PR?
> 
> Will file an issue to work on once this change goes in.

I think it's better to issue another new PR to do that, make this one simple, and go on.

arronwy:

fengwang666 (approved):
LGTM. Thanks! @yibozhuang

liubin (approved):

PR #4051, opened 2022-04-07
enable vmx for vm factory

bergwolf:


liubin (approved):

src/runtime/virtcontainers/qemu_amd64.go
```diff
@@ -38,6 +38,8 @@ const (
 
 	defaultQemuMachineOptions = "accel=kvm,kernel_irqchip=on"
 
+	splitIrqChipMachineOptions = "accel=kvm,kernel_irqchip=split"
+
```
gkurz, line 42: This change doesn't seem to belong here. Please move it to another patch.

gkurz (commented):
FYI, migration of L1 with running L2s is supposed to _work_ with recent kernels according to https://bugzilla.kernel.org/show_bug.cgi?id=53851#c3 .

src/runtime/virtcontainers/qemu_ppc64le.go
```diff
@@ -115,10 +115,6 @@ func (q *qemuPPC64le) bridges(number uint32) {
 	q.Bridges = genericBridges(number, q.qemuMachine.Type)
 }
 
-func (q *qemuPPC64le) cpuModel() string {
-	return defaultCPUModel
-}
-
```
gkurz, line 121: This change makes sense but it isn't really related to this PR either.

arronwy:

jodh-intel:
Ping @bergwolf.

src/runtime/virtcontainers/qemu_amd64.go
bergwolf: It's a simple cleanup. I've split it into a separate commit.

src/runtime/virtcontainers/qemu_ppc64le.go
bergwolf: Same here. It's simple enough to go through in the same PR but in a separate commit.

bergwolf:
rebased on the latest main.

bergwolf:
@gkurz The intention is not to migrate from L1 to L2, but from L1 to L1 with vmx capability enabled.

teawater (approved):

Issue #4052, opened 2022-04-07
agent: The kata-agent isn't rebuilt when Cargo "features" is changed


ManaSugi:
# Description of problem

Currently, `Makeifle` for Kata Agent does not rebuild the agent sources when Cargo "features" is changed because the `Makefile` specifies the sources for prerequisites as follows.

```Makefile
SOURCES := \
    $(shell find . 2>&1 | grep -E '.*\.rs$$') \
    Cargo.toml

$(TARGET_PATH): $(SOURCES) | show-summary
    @RUSTFLAGS="$(EXTRA_RUSTFLAGS) --deny warnings" cargo build --target $(TRIPLE) --$(BUILD_TYPE) $(EXTRA_RUSTFEATURES)
```

This is a troublesome problem because users need to run `make clean` once when they want to rebuild the agent.

# Expected result

If we rebuild the agent with `SECCOMP=no` after building the agent with `SECCOMP=yes` (default), the agent will become a static binary as folows because the binary does not have any dependencies of libseccomp.

```sh
$ cd src/agent
$ make
$ ldd target/x86_64-unknown-linux-musl/release/kata-agent
        linux-vdso.so.1 (0x00007ffc7e1d5000)
        libseccomp.so.2 => /usr/local/lib/libseccomp.so.2 (0x00007ff380ee5000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff380cf3000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff38185e000)

$ make SECCOMP=no
   Compiling rustjail v0.1.0 (/home/sugimoto/go/src/github.com/kata-containers/kata-containers/src/agent/rustjail)
   Compiling kata-agent v0.1.0 (/home/sugimoto/go/src/github.com/kata-containers/kata-containers/src/agent)
     Finished release [optimized] target(s) in 1m 30s
$ ldd target/x86_64-unknown-linux-musl/release/kata-agent
        statically linked
```

# Actual result

The agent isn't rebuilt because the sources aren't modified.

```sh
$ cd src/agent
$ make
$ ldd target/x86_64-unknown-linux-musl/release/kata-agent
        linux-vdso.so.1 (0x00007ffc7e1d5000)
        libseccomp.so.2 => /usr/local/lib/libseccomp.so.2 (0x00007ff380ee5000)
[...]

PR #4053, opened 2022-04-07
agent: Allow the agent to be rebuilt with the change of Cargo features

ManaSugi:
This allows the kata-agent to be rebuilt when Cargo "features" is
changed. The Makefile for the agent do not need to specify the
sources for prerequisites by having Cargo check for the sources
changes.

Fixes: #4052


liubin (approved):

bergwolf (approved):
Thanks!

Issue #4054, opened 2022-04-07
kata-monitor: links in the profile index page (/debug/pprof/?sandbox=$ID) are broken


fgiudici:
One of the kata-monitor functionality is to act as a proxy to retrieve data profiles from the kata shims running in the same node.
In order to do this, kata-monitor exposes all the available pprof endpoints for each shim: to identify the right shim, it requires a query string containing the sandbox id that should be targeted (i.e., `?sandbox=$SANDBOX_ID`).
In particular, the /debug/pprof/ endpoint is a convenient way  to access the profiling data from a web browser: it will present a simple html index page with links to all the available profiling data endpoints. It could be accessed with:
`http://$NODEIP:8090/debug/pprof/?sandbox=$ID`
This index page is generated by the kata shim associated with the sandbox id specified in the query string. The request to kata-monitor is in fact forwarded to the kata shim. The reply is then forwarded by kata-monitor as is.
The issue is that the profiling data index page will not include the query string present in the original request: so, the links will be broken when accessed through kata-monitor as will miss the reference to the sandbox id, which will point kata-monitor to the right kata shim.

Example of a generated link:
`<a href='allocs?debug=1'>allocs</a>`
Example of  the link required in order to be reachable via kata-monitor:
`<a href='allocs?sandbox=$SANDBOX_ID&debug=1'>allocs</a>`

PR #4055, opened 2022-04-07
kata-monitor: update the hrefs in the debug/pprof index page

fgiudici:
kata-monitor allows to get data profiles from the kata shim
instances running on the same node by acting as a proxy
(e.g., http://$NODE_ADDRESS:8090/debug/pprof/?sandbox=$MYSANDBOXID).
In order to proxy the requests and the responses to the right shim,
kata-monitor requires to pass the sandbox id via a query string in the
url.

The profiling index page proxied by kata-monitor contains the link to all
the data profiles available. All the links anyway do not contain the
sandbox id included in the request: the links result then broken when
accessed through kata-monitor.
This happens because the profiling index page comes from the kata shim,
which will not include the query string provided in the http request.

Let's add on-the-fly the sandbox id in each one of the href tags returned
by the kata shim index page before providing the proxied page.

Fixes: #4054

### DIFF
--- a/src/runtime/pkg/kata-monitor/pprof.go
+++ b/src/runtime/pkg/kata-monitor/pprof.go
@@ -10,6 +10,8 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"regexp"
+	"strings"
 
 	cdshim "github.com/containerd/containerd/runtime/v2/shim"
 
@@ -33,7 +35,13 @@ func (km *KataMonitor) composeSocketAddress(r *http.Request) (string, error) {
 	return shim.SocketAddress(sandbox), nil
 }
 
-func (km *KataMonitor) proxyRequest(w http.ResponseWriter, r *http.Request) {
+func (km *KataMonitor) proxyRequest(w http.ResponseWriter, r *http.Request,
+	proxyResponse func(req *http.Request, w io.Writer, r io.Reader) error) {
+
+	if proxyResponse == nil {
+		proxyResponse = copyResponse
+	}
+
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 
 	socketAddress, err := km.composeSocketAddress(r)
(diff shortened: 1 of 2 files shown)


fgiudici:
I would backport this to 2.4 branch, so adding porting label accordingly.
If instead for any reason it may be better to not backport, just change the label please, I swear I will not complain (too much ;-D ).

src/runtime/pkg/kata-monitor/pprof.go
```diff
@@ -73,38 +82,97 @@ func (km *KataMonitor) proxyRequest(w http.ResponseWriter, r *http.Request) {
...
 	w.Header().Set("Content-Disposition", `attachment; filename="trace"`)
-	km.proxyRequest(w, r)
+	km.proxyRequest(w, r, nil)
+}
+
+func copyResponse(req *http.Request, w io.Writer, r io.Reader) {
+	io.Copy(w, r)
+}
+
+func copyResponseAddingSandboxIdToHref(req *http.Request, w io.Writer, r io.Reader) {
```
liubin, line 129: Since the pprof index page is in a fixed format, and hard to consider change frequently, it is recommended to use regular expressions to  append the `sandbox` parameter like [this](https://go.dev/play/p/VuRljpIiDPa):

```golang
re := regexp.MustCompile(`<a href=['"](\w+)\?(debug=[0-9])['"]>`)
sandbox := "123456"
newHtml := re.ReplaceAllString(oldHtml, fmt.Sprintf("<a href='$1?sandbox=%s&$2'>", sandbox))
```

liubin, line 129: Since the pprof index page is in a fixed format, and hard to consider change frequently, it is recommended to use regular expressions to  append the `sandbox` parameter like [this](https://go.dev/play/p/VuRljpIiDPa):

```golang
re := regexp.MustCompile(`<a href=['"](\w+)\?(debug=[0-9])['"]>`)
sandbox := "123456"
newHtml := re.ReplaceAllString(oldHtml, fmt.Sprintf("<a href='$1?sandbox=%s&$2'>", sandbox))
```

fgiudici: Hey thanks @liubin :-) 
Changed the code to use regexp. Ok to drop support to hrefs with no query string (?debug=1).
Note: I tweaked a little bit the regular expression to match any kind of query string, not just "?debug".

liubin: Since pprof index page is a static page but not a stream, so I think we can process the entire body as a string after reading completed from the request, how about this?

src/runtime/pkg/kata-monitor/pprof.go
```diff
@@ -73,38 +82,64 @@ func (km *KataMonitor) proxyRequest(w http.ResponseWriter, r *http.Request) {
...
+	sb, err := getSandboxIDFromReq(req)
+	if err != nil {
+		monitorLog.WithError(err).Warning("missing sandbox query in pprof url")
+		copyResponse(req, w, r)
+		return
+	}
+	buf, err := io.ReadAll(r)
+	if err != nil {
+		monitorLog.WithError(err).Fatal("cannot read pprof index page")
+		os.Exit(1)
```
liubin, line 139: Don't `os.Exit`, use `serveError() && return` instead, `os.Eixt` will terminate kata-monitor process.

fgiudici: mmm, yeah, maybe too drastic to terminate kata-monitor :-)
Reworked to return an error (opted for an "InternalServer" error as my guess would be something weird happened with the connection to the runtime), how that looks?

lifupan (approved):

arronwy:

Issue #4028, opened 2022-04-01
Failed to get D-Bus connection: Operation not permitted


liyimeng:
# Get your issue reviewed faster

To help us understand the problem more quickly, please do the following:

1. Run the `kata-collect-data.sh` script, which is installed as part of Kata Containers
   or `kata-containers.collect-data`, which is installed as part of the Kata Containers
   snapcraft package.
   ```sh
   $ sudo kata-collect-data.sh > /tmp/kata.log
   ```
   or
   ```sh
   $ sudo kata-containers.collect-data > /tmp/kata.log
   ```
1. Review the output file (`/tmp/kata.log`) to ensure it doesn't
   contain any private / sensitive information
1. Paste the *entire* contents of the file into this issue as a comment
   (the script generates markdown format output).

The information provided will help us to understand the problem more quickly
so saves time for both of us! :smile:

# Description of problem

I follow the kata-deploy instruction to deploy kata on a k3s node, the pod failed to run through, logs says: 

```
kubectl -n kube-system logs -f kata-deploy-7vnd9
Failed to get D-Bus connection: Operation not permitted
copying kata artifacts onto host
warning: containerd-shim-kata-fc-v2 already exists
#!/usr/bin/env bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-fc.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
warning: containerd-shim-kata-qemu-v2 already exists
#!/usr/bin/env bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-qemu.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
Creating the default shim-v2 binary
warning: containerd-shim-kata-clh-v2 already exists
#!/usr/bin/env bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-clh.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
Add Kata Containers as a supported runtime for containerd
Configuration exists for plugins.cri.containerd.runtimes.kata, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata.options, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-fc, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-fc.options, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-qemu, overwriting
[...]

PR #4060, opened 2022-04-08
WIP: support both openrc and systemd

liyimeng:
Update deployment script to support both openrc and systemd host system.
To fix ssues #4028


liyimeng:
#needs-backport

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -66,8 +66,13 @@ function configure_cri_runtime() {
...
 	esac
-	systemctl daemon-reload
-	systemctl restart "$1"
+	
+	if [ -x /sbin/openrc-run ]; then
+		service k3s-service restart
+	else
+		systemctl daemon-reload
+		systemctl restart "$1"
+	fi
```
fidencio, line 75: I'd create a new function, something like:
```sh
function restart_service() {
    if [ -x /sbin/openrc-run ]; then
        service "$1" restart
    else
        systemctl daemon-reload
        systemctl restart "$1"
    fi
}
```

And then simply call:
```sh
    restart_service "$1"
```

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -66,8 +66,13 @@ function configure_cri_runtime() {
 		configure_containerd
 		;;
 	esac
-	systemctl daemon-reload
-	systemctl restart "$1"
+	
+	if [ -x /sbin/openrc-run ]; then
+		service k3s-service restart
```
fidencio, line 71: We don't want to restart k3s-service, we want to restart "$1", which may be crio, containerd, k3s, k3s-agent.  See the suggestion below, please.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -248,10 +253,17 @@ function cleanup_containerd() {
...
 function reset_runtime() {
 	kubectl label node "$NODE_NAME" katacontainers.io/kata-runtime-
-	systemctl daemon-reload
-	systemctl restart "$1"
-	if [ "$1" == "crio" ] || [ "$1" == "containerd" ]; then
-		systemctl restart kubelet
+
+	if [ -x /sbin/openrc-run ]; then
+		if [ "$1" == "crio" ] || [ "$1" == "containerd" ]; then
+			service restart kubelet
```
fidencio, line 259: Here you don't want to restart only kubelet, but also the containerd engine ...

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -248,10 +253,17 @@ function cleanup_containerd() {
...
+		fi
+	else
+		systemctl daemon-reload
+		systemctl restart "$1"
+		if [ "$1" == "crio" ] || [ "$1" == "containerd" ]; then
+			systemctl restart kubelet
+		fi
```
fidencio, line 266: Here we can also take advantage of the function suggested above, and do something like:
```sh
    restart_service "$1"
    if [ "$1" == "crio" ] || [ "$1" == "containerd" ]; then
        restart_service kubelet
    fi
```

WDYT?

fidencio (commented):
I have left a few comments in the series.
Also, please, take a look at https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#examples, there you'll see how to format your commit and commit message in a way that our CI is happy about that.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: Good point

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: I check k3s install script, it named the service as k3s-service, might be that need to taken into account?

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: Yes!

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: Good points!

liyimeng:
@fidencio is it a bit better looking now?

tools/packaging/kata-deploy/scripts/kata-deploy.sh
fidencio: It might be taken into account, and you should also check the name on different distros supporting OpenRC.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -62,12 +74,12 @@ function configure_cri_runtime() {
 	crio)
 		configure_crio
 		;;
-	containerd | k3s | k3s-agent)
+	containerd | k3s*)
```
fidencio, line 77: This change is good, but it's unrelated to what we're trying to achieve, so I'd have it in a different commit (still part of the same PR, but in a different commit).

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -62,12 +74,12 @@ function configure_cri_runtime() {
...
 		configure_containerd
 		;;
 	esac
-	systemctl daemon-reload
-	systemctl restart "$1"
+	
+	restart_service $1
```
fidencio, line 82:
```suggestion
	restart_service "$1"
```

Sorry, I'm paranoid about not quoting those.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -39,14 +39,26 @@ function get_container_runtime() {
...
+		echo $(k3s_service_name)
+	else
+		echo "$runtime" | awk -F '[:]' '{print $1}'
+	fi
+}
+
+function k3s_service_name() {
+	local k3s_service=""
+	if [ -x /sbin/openrc-run ]; then
+		k3s_service=$(ls /etc/init.d/ | grep k3s)
```
fidencio, line 51: This feel error prone, in case something else k3s gets added there.
Do the services have a specific name that happens to be distro agnostic with OpenRC?

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -39,14 +39,26 @@ function get_container_runtime() {
...
 		else
-			echo "k3s"
+			k3s_service="k3s"
 		fi
-	else
-		echo "$runtime" | awk -F '[:]' '{print $1}'
 	fi
+
+	if [ -z $k3s_service ]; then die "can't find k3s service"; fi
+	echo $k3s_service
```
fidencio, line 61:
```suggestion
	echo "$k3s_service"
```

fidencio (commented):
It's looking very good!
I've left a few questions and suggestions.  Thanks @liyimeng!

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: I actually need this to get it worked in my env. which I install k3s with install.sh script. The script names k3s service as k3s_service.  I have not verified how k3s look like if it is installed in another way, like using package manger, so I do it in this way to avoid breaking early implementation, which should be tested and works OK.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: you actually have good point here

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: agree, I wasn't sure here as well. In my setup, it is named as k3s-service. If checking with k3s official site, it dose name like this everywhere. But you never know if someone re-pack and re-distribute k3s and pick up another name.

liyimeng:
@fidencio Update with addressing your comment.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
fidencio: Could you add this as a previous commit, explaining this in the commit message?

This is really good info and deserves its own commit.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
fidencio: @jodh-intel, @devimc, may I borrow your brain here a little bit?  Do you have any suggestion on this?

fidencio (commented):
Thanks, we're getting there!
I've looped in two colleagues, as I'd like to borrow their brain (usually work better than mine ;-)) and get their suggestion.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: Sorry I don't have resource for doing so.  Providing many distribution running systemd, OpenRC only used in some lightweight systems, I think install.sh is likely the main way to get k3s install, hence get cover with this change.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: see https://github.com/kata-containers/kata-containers/pull/4065

fidencio: Thanks for splitting the commit.  I have closed #4065 as having the commit split but part of this very same PR is preferred.

fidencio:
@liyimeng, could you give a try on the following image? quay.io/kata-containers/kata-deploy-ci:1d39dd163e03b88303008afecc4c0b5b1a3facba

For testing purposes, you can replace the image used https://github.com/kata-containers/kata-containers/blob/698e45f403a84c089f9fcd2623f84cb2926dec39/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml#L19 by the one I pointed above.

The same has to be done for kata-cleanup: https://github.com/kata-containers/kata-containers/blob/698e45f403a84c089f9fcd2623f84cb2926dec39/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml#L21

This PR doesn't break the current workflow with systemd, as shown here: https://github.com/kata-containers/kata-containers/runs/5955626170?check_suite_focus=true

Also, @liyimeng, are you on the Kata Containers slack? :-)

liyimeng:
@fidencio Not, I am not. I just drop by. :)  kata is something pretty new to me. I just get hands on to see what is it about .

liyimeng:
@fidencio This change seem not helpful enough.  I still get Dbus error as before 

```
$ kubectl -n kube-system get daemonset  kata-deploy -o yaml | grep image:
        image: quay.io/kata-containers/kata-deploy-ci:1d39dd163e03b88303008afecc4c0b5b1a3facba
$ kubectl -n kube-system logs -f kata-deploy-4d9tt
Failed to get D-Bus connection: Operation not permitted
warning: containerd-shim-kata-fc-v2 already exists
#!/usr/bin/env bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-fc.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
warning: containerd-shim-kata-qemu-v2 already exists
#!/usr/bin/env bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-qemu.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
Creating the default shim-v2 binary
warning: containerd-shim-kata-clh-v2 already exists
#!/usr/bin/env bash
KATA_CONF_FILE=/opt/kata/share/defaults/kata-containers/configuration-clh.toml /opt/kata/bin/containerd-shim-kata-v2 "$@"
Add Kata Containers as a supported runtime for containerd
Configuration exists for plugins.cri.containerd.runtimes.kata, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata.options, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-fc, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-fc.options, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-qemu, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-qemu.options, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-clh, overwriting
Configuration exists for plugins.cri.containerd.runtimes.kata-clh.options, overwriting
Failed to get D-Bus connection: Operation not permitted

```

liyimeng:
@fidencio Is the script running inside container or host?  I find my fix make little sense. The scripts assume host running systemd and possible to communicate with it using dbus.  While my fix check the existence of /sbin/openrc-run, this for sure never be true since openrc is never installed inside container. 

if we mount host rootfs and run the script via chroot, it might make some sense. Otherwise, this is a no-go.

fidencio:
> @fidencio Is the script running inside container or host?

That's running inside the container, but we expose some mounts there.
Please, take a look at:https://github.com/kata-containers/kata-containers/blob/698e45f403a84c089f9fcd2623f84cb2926dec39/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml#L33-L65

With this in mind, we need to expose as little as possible to be able to figure out whether we're running on OpenRC or not.

> I find my fix make little sense. The scripts assume host running systemd and possible to communicate with it using dbus. While my fix check the existence of /sbin/openrc-run, this for sure never be true since openrc is never installed inside container.

See the answer above. :-)

> 
> if we mount host rootfs and run the script via chroot, it might make some sense. Otherwise, this is a no-go.

And bingo! :-)

Your fixes are on the right path, but we need to update our mountings.

fidencio:
> @fidencio No, I am not. I just drop by. :) kata is something pretty new to me. I just get hands on to see what is it about .

Please, join us on https://bit.ly/kataslack

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -39,13 +39,28 @@ function get_container_runtime() {
...
+		echo $(k3s_service_name)
+	else
+		echo "$runtime" | awk -F '[:]' '{print $1}'
+	fi
+}
+
+function k3s_service_name() {
+	local k3s_service=""
+	if [ -x /sbin/openrc-run ]; then
+		k3s_service=$(ls /etc/init.d/ | grep k3s)
```
jodh-intel, line 51: That's an awkward one. I think the best you could do is make this slightly more reliable by anchoring "k3b". Something like this:

```suggestion
		k3s_service=$(ls /etc/init.d/ | grep -E "\<k3s\>")
```

... which will still match for a service name like `k3s-service`.

liyimeng:
> > @fidencio No, I am not. I just drop by. :) kata is something pretty new to me. I just get hands on to see what is it about .
> 
> Please, join us on https://bit.ly/kataslack

Slack tell me "This link is no longer active"

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: Totally agree, a better solution need a big re-fact, like chroot based solution mentioned above, I guess.

liyimeng:
@fidencio I think a less hacky solution is to mount the host rootfs into container, the perform a chroot before executing the the script. This allow a full inspection on the host setup. I am not sure if kata like this idea since it might "less" secure, or it is ok since host already allow to change its runtime, giving access to the rest is fine? 
 
I am pretty new here, just wanna give a taste on kata, so far, I have got it yet :)

fidencio:
> @fidencio I think a less hacky solution is to mount the host rootfs into container, the perform a chroot before executing the the script. This allow a full inspection on the host setup. I am not sure if kata like this idea since it might "less" secure, or it is ok since host already allow to change its runtime, giving access to the rest is fine?
> 
> I am pretty new here, just wanna give a taste on kata, so far, I have got it yet :)

I start to think we should have two base images, instead, and have the init system as an argument to the kata-deploy script.

Meanwhile, please, use the following link to join us, and let's have a brainstorm session there: https://join.slack.com/t/katacontainers/shared_invite/zt-16w1u6usn-sK871qbMxVN8KsCP5Gr56A

SunnyCai1:
> > > @fidencio No, I am not. I just drop by. :) kata is something pretty new to me. I just get hands on to see what is it about .
> > 
> > 
> > Please, join us on https://bit.ly/kataslack
> 
> Slack tell me "This link is no longer active"



> > > @fidencio No, I am not. I just drop by. :) kata is something pretty new to me. I just get hands on to see what is it about .
> > 
> > 
> > Please, join us on https://bit.ly/kataslack
> 
> Slack tell me "This link is no longer active"

Thank you for flagging. The link is fixed on https://katacontainers.io/community/

arronwy:

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -216,6 +231,15 @@ function configure_containerd() {
 	done
 }
 
+function restart_service() {
+    if [ -x /sbin/openrc-run ]; then
+        service "$1" restart
+    else
```
jodh-intel, line 237: It might be worth adding a check for systemd here so we can have an `else die "unknown init daemon"` to "handle the impossible" :smile:

jodh-intel (commented):
@liyimeng - I think we should be able to land this soon. My concern though is how it's maintained as we don't have any non-systemd CI systems afaik.

Are you happy to ensure `kata-deploy` continues to work with openrc? Theoretically we could add a non-systemd CI but, for better or worse, all our infrastructure assumes systemd, so it would potentially be a difficult and pain-staking task.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
liyimeng: I am actually giving up on this. I  now understand why you guys stick to systemd for some reason.  Please see this https://github.com/OpenRC/openrc/issues/518. 

Providing the complexity it brings, I think manually installing kata on each host is more feasible way to go. 

Instead of providing a solution, this PR may fallback to document a limit instead. 

For OpenRC fans, if this (https://github.com/kata-containers/kata-containers/issues/4110) could be landed, Go with a dirty hands is probably to the way to go.   I have hacked the Makefile to make kata statically linked with muslc to run it on Alpine Linux.  It built, but not run.  Seem more work to do. Not a simply static link.

Issue #4063, opened 2022-04-08
no need to write error of virtiofsd to kata log


tiezhuoyu:
The scanner read nothing from viriofsd stderr pipe, so there is no need to write `scanner.Text()` to kata log.

The arg `--syslog` redirects stderr of virtiofsd to syslog.
``` golang
func (v *virtiofsd) args(FdSocketNumber uint) ([]string, error) {

	args := []string{
		// Send logs to syslog
		"--syslog",
        ...
```

Therefore, the scanner created in func `Start` actually reads nothing from stderr pipe.
``` golang
// Start the virtiofsd daemon
func (v *virtiofsd) Start(ctx context.Context, onQuit onQuitFunc) (int, error) {
        ...
	stderr, err := cmd.StderrPipe()
	if err != nil {
		return pid, err
	}

	if err = utils.StartCmd(cmd); err != nil {
		return pid, err
	}

	// Monitor virtiofsd's stderr and stop sandbox if virtiofsd quits
	go func() {
		scanner := bufio.NewScanner(stderr)
		for scanner.Scan() {
			v.Logger().WithField("source", "virtiofsd").Info(scanner.Text())
		}
		v.Logger().Info("virtiofsd quits")
		// Wait to release resources of virtiofsd process
		cmd.Process.Wait()
		if onQuit != nil {
			onQuit()
		}
	}()
```

To prove this, I have an example as follow:
1. Run a container:
```
[root@cloud-1 kata-runtime]# ctr run --rm --runtime io.containerd.run.kata.v2 -t docker.io/library/busybox:latest test-kata sh
/ # 
```
2. Get pid of virtiofsd in a new terminal, 1006604 & 1006612:
```
[root@cloud-1 runtime]# ps -ef | grep virtiofsd
root     1006604 1006593  0 16:12 ?        00:00:00 /opt/kata/libexec/kata-qemu/virtiofsd --syslog -o cache=auto -o no_posix_lock -o source=/run/kata-containers/shared/sandboxes/test-kata/shared --fd=3 -f --thread-pool-size=1 -o announce_submounts
root     1006612 1006604  0 16:12 ?        00:00:00 /opt/kata/libexec/kata-qemu/virtiofsd --syslog -o cache=auto -o no_posix_lock -o source=/run/kata-containers/shared/sandboxes/test-kata/shared --fd=3 -f --thread-pool-size=1 -o announce_submounts
[...]

PR #4064, opened 2022-04-08
runtime: no need to write virtiofsd error to log

tiezhuoyu:
The scanner readis nothing from viriofsd stderr pipe, because param
'--syslog' rediercts stderr to syslog. So there is no need to write
scanner.Text() to kata log.

Fixes: #4063

Signed-off-by: tiezhuoyu <tiezhuoyu@outlook.com>

### DIFF
--- a/src/runtime/virtcontainers/virtiofsd.go
+++ b/src/runtime/virtcontainers/virtiofsd.go
@@ -6,7 +6,6 @@
 package virtcontainers
 
 import (
-	"bufio"
 	"context"
 	"fmt"
 	"net"
@@ -134,24 +133,14 @@ func (v *virtiofsd) Start(ctx context.Context, onQuit onQuitFunc) (int, error) {
 
 	v.Logger().WithField("path", v.path).Info()
 	v.Logger().WithField("args", strings.Join(args, " ")).Info()
-	stderr, err := cmd.StderrPipe()
-	if err != nil {
-		return pid, err
-	}
 
 	if err = utils.StartCmd(cmd); err != nil {
 		return pid, err
 	}
 
-	// Monitor virtiofsd's stderr and stop sandbox if virtiofsd quits
 	go func() {
-		scanner := bufio.NewScanner(stderr)
-		for scanner.Scan() {
-			v.Logger().WithField("source", "virtiofsd").Info(scanner.Text())
-		}
-		v.Logger().Info("virtiofsd quits")
-		// Wait to release resources of virtiofsd process
 		cmd.Process.Wait()
+		v.Logger().Info("virtiofsd quits")
 		if onQuit != nil {
 			onQuit()
 		}


src/runtime/virtcontainers/virtiofsd.go
```diff
@@ -147,7 +147,7 @@ func (v *virtiofsd) Start(ctx context.Context, onQuit onQuitFunc) (int, error) {
 	go func() {
 		scanner := bufio.NewScanner(stderr)
 		for scanner.Scan() {
-			v.Logger().WithField("source", "virtiofsd").Info(scanner.Text())
```
liubin, line 150: @tiezhuoyu Can you explain if `scanner.Text()` will be called if what you said `scanner readis nothing from viriofsd stderr` is right.

tiezhuoyu: `scanner.Text()` is not excuted, because `scanner.Scan()` blocks to read from buffer. I think it is enough with only `scanner.Scan()`.

gkurz: Naive question : if the scanners basically sits reading nothing from the pipe, why is the scanner needed at all in the first place ?

tiezhuoyu: Yes, I miss this... Both `scanner.Scan` and `cmd.Process.Wait()` do the same thing, to block untile virtiofsd exits. Since read nothing from stderr, it is not necessary to create a `scanner` and call `scanner.Scan()`.

gkurz: > Yes, I miss this... Both `scanner.Scan` and `cmd.Process.Wait()` do the same thing, to block until virtiofsd exits. Since read nothing from stderr, it is not necessary to create a `scanner` and call `scanner.Scan()`.

That was exactly my point. :smile:

src/runtime/virtcontainers/virtiofsd.go
```diff
@@ -134,24 +133,16 @@ func (v *virtiofsd) Start(ctx context.Context, onQuit onQuitFunc) (int, error) {
...
 	}
 
 	// Monitor virtiofsd's stderr and stop sandbox if virtiofsd quits
 	go func() {
-		scanner := bufio.NewScanner(stderr)
-		for scanner.Scan() {
-			v.Logger().WithField("source", "virtiofsd").Info(scanner.Text())
-		}
-		v.Logger().Info("virtiofsd quits")
 		// Wait to release resources of virtiofsd process
```
liubin, line 143: I'd suggest moving `v.Logger().Info("virtiofsd quits")` to after `cmd.Process.Wait()`.

gkurz, line 143: It would be fine to drop this comment as well.

src/runtime/virtcontainers/virtiofsd.go
```diff
@@ -134,24 +133,16 @@ func (v *virtiofsd) Start(ctx context.Context, onQuit onQuitFunc) (int, error) {
...
-	stderr, err := cmd.StderrPipe()
-	if err != nil {
-		return pid, err
-	}
 
 	if err = utils.StartCmd(cmd); err != nil {
 		return pid, err
 	}
 
 	// Monitor virtiofsd's stderr and stop sandbox if virtiofsd quits
```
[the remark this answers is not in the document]
tiezhuoyu: got it.

src/runtime/virtcontainers/virtiofsd.go
tiezhuoyu: ok.

tiezhuoyu:
Sorry for my carelessness, modified.

gkurz:
The 'ubuntu-clh-crio' CI is failing with a really weird crash during the build:

```
BUILD    /tmp/jenkins/workspace/kata-containers-2.0-ubuntu-PR-x86_64-clh-crio-kata-repo/go/src/github.com/kata-containers/kata-containers/src/runtime/kata-runtime
# github.com/BurntSushi/toml
unexpected fault address 0xa5b900
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0xa5b900 pc=0xa5b900]
```

@GabyCT any idea ?

arronwy:

liubin (approved):

Issue #4068, opened 2022-04-11
agent tests: Add tests for mount_storage function in src/mount.rs


bradenrayhorn:
Add tests for `mount_storage` function in `src/mount.rs` file in the agent.

PR #4069, opened 2022-04-11
agent: add tests for mount_storage

bradenrayhorn:
Add test coverage for mount_storage function in src/mount.rs.

Fixes: #4068

### DIFF
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -1498,6 +1498,130 @@ mod tests {
     }
 
     #[test]
+    fn test_mount_storage() {
+        #[derive(Debug)]
+        struct TestData<'a> {
+            test_user: TestUserType,
+            storage: Storage,
+            error_contains: &'a str,
+
+            make_source_dir: bool,
+            make_mount_dir: bool,
+            deny_mount_permission: bool,
+        }
+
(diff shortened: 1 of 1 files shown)


src/agent/src/mount.rs
```diff
@@ -1497,6 +1497,130 @@ mod tests {
...
+                .unwrap();
+            }
+
+            if result.is_ok() {
+                nix::mount::umount(&mount_point).unwrap();
+            }
+
+            let msg = format!("{}: result: {:?}", msg, result);
+            if d.error_contains.is_empty() {
+                assert!(result.is_ok(), "{}", msg);
```
liubin, line 1615: Can here be replaced by `assert!(result.is_ok(), msg);` ? And line 1617

bradenrayhorn: @liubin That would work for now, but in Rust 2021 that would become an error, so I think we need to keep it as a string literal. 

https://doc.rust-lang.org/nightly/edition-guide/rust-2021/panic-macro-consistency.html

liubin (approved):

jodh-intel (approved):
Thanks @bradenrayhorn !

lgtm

Issue #4082, opened 2022-04-13
rustjail: Add tests for parse_mount_table function


bradenrayhorn:
Add tests for the `parse_mount_table` function in `rustjail/src/mount.rs` Relates to #284.

PR #4083, opened 2022-04-13
rustjail: add tests for parse_mount_table

bradenrayhorn:
Add tests for parse_mount_table function in rustjail/src/mount.rs.
Includes some minor refactoring improve the testability of the
function and improve its error values.

Fixes: #4082


src/agent/rustjail/src/mount.rs
```diff
@@ -32,16 +32,19 @@ use crate::log_child;
...
     mount_point: String,
     optional: String,
     fstype: String,
 }
 
 const MOUNTINFOFORMAT: &str = "{d} {d} {d}:{d} {} {} {} {}";
+const MOUNTINFO_PATH: &str = "/proc/self/mountinfo";
 const PROC_PATH: &str = "/proc";
 
+const ERR_FAILED_PARSE_MOUNTINFO: &str = "failed to parse mount info file";
```
liubin, line 46: Nit: `mount info` -> `mountinfo`?

src/agent/rustjail/src/mount.rs
```diff
@@ -1050,10 +1055,37 @@ mod tests {
...
 
+    // Parameters:
+    //
+    // 1: expected Result
+    // 2: actual Result
+    // 3: string used to identify the test on error
+    macro_rules! assert_result {
+        ($expected_result:expr, $actual_result:expr, $msg:expr) => {
+            if $expected_result.is_ok() {
+                let expected_level = $expected_result.as_ref().unwrap();
```
liubin, line 1071: What's the `_level` mean?

src/agent/rustjail/src/mount.rs
```diff
@@ -1050,10 +1055,37 @@ mod tests {
...
+                let actual_level = $actual_result.unwrap();
+                assert!(*expected_level == actual_level, "{}", $msg);
+            } else {
+                let expected_error = $expected_result.as_ref().unwrap_err();
+                let expected_error_msg = format!("{:?}", expected_error);
```
liubin, line 1076: Duplicated with line 1079

src/agent/rustjail/src/mount.rs
```diff
@@ -1050,10 +1055,37 @@ mod tests {
...
+
+                if let Err(actual_error) = $actual_result {
+                    let actual_error_msg = format!("{:?}", actual_error);
+
+                    assert!(expected_error_msg == actual_error_msg, "{}", $msg);
+                } else {
+                    assert!(expected_error_msg == "expected error, got OK", "{}", $msg);
```
liubin, line 1083: This line is something tricky.  It's better to use `assert!(result.is_err())` to check error result. For example, how about this:

```rust
assert!($actual_result.is_err());

let actual_error_msg = format!("{:?}", $actual_result.unwrap_err());
assert!(expected_error_msg == actual_error_msg, "{}", $msg);
```

bradenrayhorn: This `assert_result` macro is interesting, it's been useful so it is now copied in a few different places. See [this comment](https://github.com/kata-containers/kata-containers/pull/3838/files#r822590397) from James.

I'll incorporate your feedback in this PR though, and probably in all places the `assert_result` macro is defined so they stay in sync.

liubin: @bradenrayhorn you can create an issue to track this, and put them in `agent/src/test_utils.rs` if possible.

arronwy:

src/agent/rustjail/src/mount.rs
bradenrayhorn: Thank you @liubin, I did not realize this file existed. I created #4093. Did you want this change in this PR? Or should I leave `assert_result` as-is in this PR, and open a future PR to address #4093?

liubin: @bradenrayhorn I think both are fine.

src/agent/rustjail/src/mount.rs
bradenrayhorn: Renamed to `_value`

src/agent/rustjail/src/mount.rs
bradenrayhorn: @liubin I've moved the `assert_result` macro to where the other `rustjail` common test macros are. Because `agent/src/test_utils.rs` is in a different crate, I don't think we can have rustjail importing in, but please correct me if I am wrong.

I will open a separate PR to add `assert_result` to the `agent/src/test_utils.rs` file for use in the agent crate tests.

liubin (approved):

src/agent/rustjail/src/mount.rs
```diff
@@ -32,16 +32,19 @@ use crate::log_child;
...
-#[derive(std::fmt::Debug)]
+#[derive(std::fmt::Debug, PartialEq)]
 pub struct Info {
     mount_point: String,
     optional: String,
     fstype: String,
 }
 
-const MOUNTINFOFORMAT: &str = "{d} {d} {d}:{d} {} {} {} {}";
+const MOUNTINFO_FORMAT: &str = "{d} {d} {d}:{d} {} {} {} {}";
```
jodh-intel, line 42: Thanks! That single byte changes makes the identifier much easier to read ;)

src/agent/rustjail/src/mount.rs
```diff
@@ -578,12 +581,14 @@ fn parse_mount_table() -> Result<Vec<Info>> {
...
-        )?;
+        )
+        .map_err(|_| anyhow!(ERR_FAILED_PARSE_MOUNTINFO))?;
 
         let fields: Vec<&str> = line.split(" - ").collect();
         if fields.len() == 2 {
             let (fstype, _source, _vfs_opts) =
-                scan_fmt!(fields[1], "{} {} {}", String, String, String)?;
+                scan_fmt!(fields[1], "{} {} {}", String, String, String)
+                    .map_err(|_| anyhow!(ERR_FAILED_PARSE_MOUNTINFO))?;
```
jodh-intel, line 591: It might be worth making this error message slightly "richer": the error shows the problem area, but it could state that the specific problem is that the optional mountinfo args are invalid, which would help debug any future issues in this area.

src/agent/rustjail/src/mount.rs
```diff
@@ -569,7 +572,7 @@ fn parse_mount_table() -> Result<Vec<Info>> {
 
         let (_id, _parent, _major, _minor, _root, mount_point, _opts, optional) = scan_fmt!(
             &line,
-            MOUNTINFOFORMAT,
+            MOUNTINFO_FORMAT,
             i32,
             i32,
             i32,
```
jodh-intel, line 578: Looking at this unchanged line, I don't think it's correct. `proc(5)` shows that the first two fields are mount ID and parent ID. The next two fields are major and minor values. But looking at:

https://github.com/torvalds/linux/blob/master/fs/proc_namespace.c#L143

... shows that the 3rd and 4th fields are unsigned. We should probably fix that but on a separate commit.

/cc @liubin.@yyyeerbo for thoughts on this.

src/agent/rustjail/src/mount.rs
```diff
@@ -1401,6 +1408,91 @@ mod tests {
...
 
+    #[test]
+    fn test_parse_mount_table() {
+        #[derive(Debug)]
+        struct TestData<'a> {
+            mountinfo_data: Option<&'a str>,
+            result: Result<Vec<Info>>,
+        }
+
+        let tests = &[
```
jodh-intel, line 1419: Could you add a few more tests to this list. Specfiically:

- A test where the first fields are valid, but which has too many optional fields in it.
- Tests where `mountinfo_data` is `"-"`, `"--"`, `"- -"` and `" - "`.

jodh-intel (commented):
Thanks @bradenrayhorn - a few comments...

src/agent/rustjail/src/lib.rs
```diff
@@ -523,6 +523,31 @@ mod tests {
         };
     }
 
+    // Parameters:
+    //
+    // 1: expected Result
+    // 2: actual Result
+    // 3: string used to identify the test on error
+    #[macro_export]
+    macro_rules! assert_result {
```
jodh-intel, line 532: Note to self: we need to see if we can create a local crate that exports this macro so we can use the single definition everywhere.

src/agent/rustjail/src/mount.rs
bradenrayhorn: What would too many optional fields look like? I was looking at [this documentation](https://www.kernel.org/doc/Documentation/filesystems/proc.txt), which says "zero or more optional fields."

src/agent/rustjail/src/mount.rs
bradenrayhorn: Similar to my question below about optional fields. Would this indicate a problem with optional fields? Or are the last 3 items (9, 10, 11) as shown in this example optional, and I'm getting mixed up between those and optional fields (7)?

```
36 35 98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 /dev/root rw,errors=continue
(1)(2)(3)   (4)   (5)      (6)      (7)   (8) (9)   (10)         (11)
```

src/agent/rustjail/src/lib.rs
bradenrayhorn: While you were out, we did get the `assert_result` macro into a shared place for the agent, but unfortunately like you said this isn't usable in rustjail. But, it was a slight improvement.

https://github.com/kata-containers/kata-containers/pull/4104

src/agent/rustjail/src/mount.rs
jodh-intel: @bradenrayhorn - sorry, bad wording. I didn't mean optional fields, but the final fields. Currently, the code will fail here with the error "failed to parse mountinfo file", but we could be more specific and say something like, "failed to parse final fields in mountinfo file".

src/agent/rustjail/src/mount.rs
jodh-intel: Again, I meant "final fields" here, not "optional fields". There should be 3 fields after the optional field terminator (`-`).

jodh-intel: One other thought: it might be worth adding a test for a mount that contains the pattern `" - "`. This isn't documented in `proc(5)`, but the spaces get encoded as shown below:

```bash
$ mkdir -p /tmp/"foo - bar"
$ sudo mount -obind /tmp/"foo - bar" /mnt
$ sudo grep foo /proc/self/mountinfo
1537 94 0:39 /foo\040-\040bar /mnt rw,nosuid,nodev shared:58 - tmpfs tmpfs rw,seclabel,size=16324552k,nr_inodes=1048576,inode64
```

jodh-intel (approved):
Thanks @bradenrayhorn.

lgtm

jodh-intel:
Restarted the FC CI which had timed out.

PR #4084, opened 2022-04-13
agent: add test coverage for find_process function

Shensd:
Add test coverage for the function find_process in src/sandbox.rs.

Signed-off-by: Jack Hance <jack.hance@ndsu.edu>

### DIFF
--- a/src/agent/src/sandbox.rs
+++ b/src/agent/src/sandbox.rs
@@ -851,4 +851,50 @@ mod tests {
         let p = s.find_container_process("not-exist-cid", "");
         assert!(p.is_err(), "Expecting Error, Got {:?}", p);
     }
+
+    #[tokio::test]
+    async fn test_find_process() {
+        let logger = slog::Logger::root(slog::Discard, o!());
+        let mut s = Sandbox::new(&logger).unwrap();
+        let (mut linux_container, _root) = create_linuxcontainer();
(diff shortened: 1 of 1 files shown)


src/agent/src/sandbox.rs
```diff
@@ -851,4 +851,59 @@ mod tests {
         let p = s.find_container_process("not-exist-cid", "");
         assert!(p.is_err(), "Expecting Error, Got {:?}", p);
     }
+
+    #[tokio::test]
+    async fn find_process_exists() {
```
liubin, line 856: Normally test function starts with `test_`.

src/agent/src/sandbox.rs
```diff
@@ -851,4 +851,59 @@ mod tests {
...
+
+        // then test if it founds the correct process
+        assert_eq!(
+            found_process.pid, test_pid,
+            "Should be able to find correct process"
+        );
+    }
+
+    #[tokio::test]
+    async fn find_process_doesnt_exist() {
```
liubin, line 891: These two cases is not very big, so they can be put in one function.

arronwy:

Shensd:
I made the edits recommended by @liubin

Shensd:
Changes are now being discussed in #4137.

Issue #4087, opened 2022-04-13
Support using annotation to adjust the number of queues for NIC


Gezx:
In xConnectVMNetwork()(src/runtime/virtcontainers/network_linux.go), the NIC multi-queue only support to use the numbers of vCPU as the numbers of queue.

`func xConnectVMNetwork(ctx context.Context, endpoint Endpoint, h Hypervisor) error {
        var err error

        span, ctx := networkTrace(ctx, "xConnectVMNetwork", endpoint)
        defer closeSpan(span, err)

        netPair := endpoint.NetworkPair()

        queues := 0
        caps := h.Capabilities(ctx)
        if caps.IsMultiQueueSupported() {
                queues = int(h.HypervisorConfig().NumVCPUs)
        }
        ...
}`

But sometimes it is not necessary to use all the vCPUs, so we add the function to adjust the numbers of NIC queue by annotation.

PR #4089, opened 2022-04-13
runtime: use annotation to config the numbers of NIC queue

Gezx:
The NIC multi-queue only support to use the numbers of vCPU as the numbers of queue. But sometimes it is not necessary to use all the vCPUs, so we add the function to adjust the numbers of NIC queue by annotation.

Fixes: #4087

Signed-off-by: gezhixin <2771485026@qq.com>

### DIFF
--- a/src/runtime/pkg/oci/utils.go
+++ b/src/runtime/pkg/oci/utils.go
@@ -764,6 +764,12 @@ func addHypervisporNetworkOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConf
 		return err
 	}
 
+        if err := newAnnotationConfiguration(ocispec, vcAnnotations.NumQueues).setUint(func(numQueues uint64) {
+                sbConfig.HypervisorConfig.NumQueues = uint32(numQueues)
+        }); err != nil {
+                return err
+        }
+
 	if err := newAnnotationConfiguration(ocispec, vcAnnotations.RxRateLimiterMaxRate).setUint(func(rxRateLimiterMaxRate uint64) {
 		sbConfig.HypervisorConfig.RxRateLimiterMaxRate = rxRateLimiterMaxRate
 	}); err != nil {
--- a/src/runtime/virtcontainers/hypervisor.go
+++ b/src/runtime/virtcontainers/hypervisor.go
@@ -501,6 +501,9 @@ type HypervisorConfig struct {
 
 	// Disable selinux from the hypervisor process
 	DisableSeLinux bool
+
+	// NumQueues is the number of queues for NIC
+	NumQueues uint32
 }
 
 // vcpu mapping from vcpu number to thread number
@@ -577,6 +580,13 @@ func (conf *HypervisorConfig) Valid() error {
 	if conf.Msize9p == 0 && conf.SharedFS != config.VirtioFS {
 		conf.Msize9p = defaultMsize9p
 	}
+
+	if conf.NumQueues == 0 {
+		conf.NumQueues = conf.NumVCPUs
+	} else if conf.NumQueues > conf.NumVCPUs {
+		hvLogger.Warnf("The NumQueues should not be larger than NumVCPUs, Setting NumQueues to NumVCPUs (%d)", conf.NumVCPUs)
+		conf.NumQueues = conf.NumVCPUs
+	}	
 
 	return nil
 }
--- a/src/runtime/virtcontainers/network_linux.go
+++ b/src/runtime/virtcontainers/network_linux.go
@@ -516,7 +516,7 @@ func xConnectVMNetwork(ctx context.Context, endpoint Endpoint, h Hypervisor) err
 	queues := 0
 	caps := h.Capabilities(ctx)
 	if caps.IsMultiQueueSupported() {
-		queues = int(h.HypervisorConfig().NumVCPUs)
+		queues = int(h.HypervisorConfig().NumQueues)
 	}
 
 	disableVhostNet := h.HypervisorConfig().DisableVhostNet
--- a/src/runtime/virtcontainers/documentation/api/1.0/api.md
+++ b/src/runtime/virtcontainers/documentation/api/1.0/api.md
@@ -348,6 +348,9 @@ type HypervisorConfig struct {
 	// Enable or disable different hardware features, ranging
 	// from memory encryption to both memory and CPU-state encryption and integrity.
 	ConfidentialGuest bool
+
+	// NumQueues is the number of queues for NIC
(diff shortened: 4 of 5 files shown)


arronwy:

src/runtime/pkg/oci/utils.go
```diff
@@ -770,6 +770,12 @@ func addHypervisporNetworkOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConf
 		return err
 	}
 
+	if err := newAnnotationConfiguration(ocispec, vcAnnotations.QueueNums).setUint(func(queueNums uint64) {
+       		sbConfig.HypervisorConfig.QueueNums = uint32(queueNums)
+     	}); err != nil {
+       		return err
+    	}
+
```
gkurz, line 778: Maybe move this a few lines up so that `RxRateLimiterRate` and `TxRateLimiterMaxRate` stay side by side ?

Gezx: Thanks for your suggestions, I will move it.

src/runtime/virtcontainers/network_linux.go
```diff
@@ -516,7 +516,13 @@ func xConnectVMNetwork(ctx context.Context, endpoint Endpoint, h Hypervisor) err
 	queues := 0
 	caps := h.Capabilities(ctx)
 	if caps.IsMultiQueueSupported() {
-		queues = int(h.HypervisorConfig().NumVCPUs)
+		if h.HypervisorConfig().QueueNums == 0 {
+            		queues = int(h.HypervisorConfig().NumVCPUs)
+       		} else if int(h.HypervisorConfig().NumVCPUs) > int(h.HypervisorConfig().QueueNums) {
```
gkurz, line 521: Both `NumVCPUs` abd `QueueNums` are `uint32` : no need to convert to `int`.

src/runtime/virtcontainers/network_linux.go
```diff
@@ -516,7 +516,13 @@ func xConnectVMNetwork(ctx context.Context, endpoint Endpoint, h Hypervisor) err
...
+            		queues = int(h.HypervisorConfig().QueueNums)
+       		} else{
+            		queues = int(h.HypervisorConfig().NumVCPUs)
+       		}
```
gkurz, line 525: The logic looks a bit fuzzy to me, especially when `QueueNums == 0`... how is this happening ? Because the user passed the annotations with a value of 0 ? Because the user didn't pass the annotation and it supposedly defaults to 0 ?

I think that `QueueNums` should belong to [1..NumVCPUs]. Any value outside this range should cause an error or at least a visible warning. The default value should be NumVCPUs to match the existing behavior.

It seems all of this would be better located in the [`Valid()` function of the `hypervisor` module](https://github.com/kata-containers/kata-containers/blob/4443bb68a45f3d1eb169b6c1cb223bfbfe9d09cb/src/runtime/virtcontainers/hypervisor.go#L529) to keep `xConnectVMNetwork()` simple.

Also, annotations are API that must be documented in `src/runtime/virtcontainers/documentation/api/1.0/api.md`.

Gezx: You are right, I use `QueueNums == 0`  that means the user didn't pass the annotation, but as you can see this may be not suitable. I will use a default value and achieve it in `func (conf *HypervisorConfig) Valid() error`.

src/runtime/virtcontainers/hypervisor.go
```diff
@@ -502,8 +502,8 @@ type HypervisorConfig struct {
 	// Disable selinux from the hypervisor process
 	DisableSeLinux bool
 
-	// NIC Multi Queue
-    	QueueNums uint32
+	// NumQueues is the number of queues for NIC
+    	NumQueues uint32
```
gkurz, line 506: space indent

src/runtime/virtcontainers/network_linux.go
gkurz: Actually you should only have `queues = int(h.HypervisorConfig().NumQueues)` in this function. All the checking should go to `Valid()`. Also, even if it is generally a bad idea to have more queues than processing power, we shouldn't silently change a user requested setting. Personally I would just print a warning.

Gezx: I follow your suggestions that have move the checking go to `Valid()`.
But I think it better to change the number of queues when more queues than processing power, if I set `NumQueues = 3` and `NumVCPUs = 2`, it can be seen as follows:
`Channel parameters for tap0:
Pre-set maximums:
RX:             0
TX:             0
Other:          0
Combined:       3
Current hardware settings:
RX:             0
TX:             0
Other:          0
Combined:       2
`
Actually, the max number of queues should be 2 not 3.

gkurz: Fine with me. So now, just fold everything into a single patch. Ensure you don't have spaces in the indention. Fix the From: and Signed-off-by:, force-push to GH and we should be good to go @Gezx . :smile:

c3d: @Gezx I tend to agree with @gkurz here. Notably, if there are multiple containers with multiple annotations, and the second container adds CPUs, then you may end up in a reasonable configuration.

You could also recognize a special value, e.g. `-1`, to mean "use number of VCPUs".

c3d:
Nit: the title of the PR is misspelled (annonation -> annotation)

c3d:
@Gezx Thanks a lot for the contribution.

The three commits use the same description. It would be better to either squash them together, or to explain what is the difference between them.

bigclouds:
It seems a useful feature, please also update  docs/how-to/how-to-set-sandbox-config-kata.md

jodh-intel (approved):
Thanks @Gezx!

lgtm

/cc @amshinde for possible networking and annotations input.

gkurz (commented):
Hi @Gezx !

This PR is currently blocked because the changelog text is sitting in a single and thus very long line:

```
ERROR: commit 666a44573ba1342e3fbfc92c957e2b26c56771b8: body line 1 too long (max 150, got 217): "The NIC multi-queue only support to use the numbers of vCPU as the number of queues. But sometimes it is not necessary to use all the vCPUs, so we add the function to adjust the number of queues for NIC by annotation."
```
Please break it down info smaller lines, ideally less than 80 characters, for a pleasant `git show` experience from a terminal.

While here, please also replace "gezhixin" in your From and Signed-off-by with a valid "FirstName LastName".

gkurz:
Hi @Gezx ! Any updates ? You'll need to rebase.

PR #4114, opened 2022-04-20
agent: modify the type of swappiness to u64

yangfeiyu20102011:
The type of MemorySwappiness in runtime is uint64, and the type of swappiness in agent is int64,
    if we set max uint64 in runtime and pass it to agent, the value will be equal to -1. We should
    modify the type of swappiness to u64

Fixes: https://github.com/kata-containers/kata-containers/issues/4123

Signed-off-by: yangfeiyu <yangfeiyu20102011@163.com>

### DIFF
--- a/src/agent/rustjail/src/cgroups/fs/mod.rs
+++ b/src/agent/rustjail/src/cgroups/fs/mod.rs
@@ -391,7 +391,7 @@ fn set_memory_resources(cg: &cgroups::Cgroup, memory: &LinuxMemory, update: bool
 
     if let Some(swappiness) = memory.swappiness {
         if (0..=100).contains(&swappiness) {
-            mem_controller.set_swappiness(swappiness as u64)?;
+            mem_controller.set_swappiness(swappiness)?;
         } else {
             return Err(anyhow!(
                 "invalid value:{}. valid memory swappiness range is 0-100",
--- a/src/agent/rustjail/src/lib.rs
+++ b/src/agent/rustjail/src/lib.rs
@@ -265,7 +265,7 @@ pub fn resources_grpc_to_oci(res: &grpc::LinuxResources) -> oci::LinuxResources 
             swap: Some(mem.Swap),
             kernel: Some(mem.Kernel),
             kernel_tcp: Some(mem.KernelTCP),
-            swappiness: Some(mem.Swappiness as i64),
+            swappiness: Some(mem.Swappiness),
             disable_oom_killer: Some(mem.DisableOOMKiller),
         })
     } else {
--- a/src/libs/oci/src/lib.rs
+++ b/src/libs/oci/src/lib.rs
@@ -381,7 +381,7 @@ pub struct LinuxMemory {
     #[serde(default, skip_serializing_if = "Option::is_none", rename = "kernelTCP")]
     pub kernel_tcp: Option<i64>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub swappiness: Option<i64>,
+    pub swappiness: Option<u64>,
     #[serde(
         default,
         skip_serializing_if = "Option::is_none",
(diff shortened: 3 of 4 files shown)


liubin:
Thank you @yangfeiyu20102011 ,  which runtime are you using and pass `-1` to lower container runtime?

yangfeiyu20102011:
> Thank you @yangfeiyu20102011 , which runtime are you using and pass `-1` to lower container runtime?

I use container engine iSula.

teawater:
Could you tell me how to setup a container with MemorySwappiness -1?
I think it should not happen.
And if kata runtime send a "MemorySwappiness -1" to agent, it should be blocked in kata runtime side but agent side.

Best,
Hui

yangfeiyu20102011:
> Could you tell me how to setup a container with MemorySwappiness -1? I think it should not happen. And if kata runtime send a "MemorySwappiness -1" to agent, it should be blocked in kata runtime side but agent side.
> 
> Best, Hui

I use container engine iSula, it set default value -1 in oci spec
oci_spec->linux->resources->memory->swappiness = (uint64_t)(-1)

Then the value will be passed to agent.

amshinde:
@yangfeiyu20102011 What does MemorySwappiness=-1 mean? Think this is not a really a correct value being passed down the stack, and if its the case should be fixed in the container engine.

yangfeiyu20102011:
> @yangfeiyu20102011 What does MemorySwappiness=-1 mean? Think this is not a really a correct value being passed down the stack, and if its the case should be fixed in the container engine.

The type of MemorySwappiness in runtime is uint64, and the type of swappiness in agent is int64,
if we set max uint64 in runtime and pass it to agent, the value will be equal to -1.

bergwolf:
@yangfeiyu20102011 Good point! IMO, we should fix https://github.com/kata-containers/kata-containers/blob/main/src/libs/oci/src/lib.rs#L384 instead though. OCI spec `swappiness` is `u64` and let's just make it be so. We already define it as `u64` in the agent protocol. We just need to fix the place that actually uses it.

teawater:
> > Could you tell me how to setup a container with MemorySwappiness -1? I think it should not happen. And if kata runtime send a "MemorySwappiness -1" to agent, it should be blocked in kata runtime side but agent side.
> > Best, Hui
> 
> I use container engine iSula, it set default value -1(max uint64) in oci spec oci_spec->linux->resources->memory->swappiness = (uint64_t)(-1)
> 
> In runtime, MemorySwappiness is uint64 and will not be blocked, then the value will be passed to agent. In agent, swappiness is i64 and equals to -1.

I think iSula want swappiness not to be set, I think runtime should set MemorySwappiness to none is better.

btw, after Linux kernel c843966c556d7370bb32e7319a6d164cb8c70ae2
the scope of swappiness is updated from 0-100 to 0-200.

yangfeiyu20102011:
> @yangfeiyu20102011 Good point! IMO, we should fix https://github.com/kata-containers/kata-containers/blob/main/src/libs/oci/src/lib.rs#L384 instead though. OCI spec `swappiness` is `u64` and let's just make it be so. We already define it as `u64` in the agent protocol. We just need to fix the place that actually uses it.

I have modified the type to u64.

yangfeiyu20102011:
> > > Could you tell me how to setup a container with MemorySwappiness -1? I think it should not happen. And if kata runtime send a "MemorySwappiness -1" to agent, it should be blocked in kata runtime side but agent side.
> > > Best, Hui
> > 
> > 
> > I use container engine iSula, it set default value -1(max uint64) in oci spec oci_spec->linux->resources->memory->swappiness = (uint64_t)(-1)
> > In runtime, MemorySwappiness is uint64 and will not be blocked, then the value will be passed to agent. In agent, swappiness is i64 and equals to -1.
> 
> I think iSula want swappiness not to be set. What about let runtime set MemorySwappiness to none?
> 
> btw, after Linux kernel c843966c556d7370bb32e7319a6d164cb8c70ae2 the scope of swappiness is updated from 0-100 to 0-200.

Yeah, I will make a PR to iSula and set the value to none.

liubin (approved):

bergwolf (approved):
Thanks @yangfeiyu20102011 !

PR #4135, opened 2022-04-21
Implement network and disk rate limiter for Cloud Hypervisor

fidencio:
In a similar way to what's already exposed as RxRateLimiterMaxRate and TxRateLimiterMaxRate for other hypervisors, let's add support for configuring a Bandwidth and Operations rate limiter for Cloud Hypervisor.

There are some not so subtle differences here compared to the other Hypervisors as Cloud Hypervisor doesn't offer a way to configure a different Rate Limiter for the Tx and Rx queues, it exposes the very same configuration which is used by both. Another difference here is that apart from the Rate Limiter based on bandwidth (`bits / sec`), Cloud Hypervisor also offers a Rate Limiter based on operations (`ops / sec`), which can be combined with the former.

Apart from those, we're also exposing an "OneTimeBurst" configuration for both bandwidth (`bits / sec`) and operations (`ops / sec`) rate limiters, and it's important to make clear that it's only used as an *initial* max rate, atop of the one defined by the admin, and it doesn't replenish.

Please note that I've decided to not fully expose all the knobs provided by Cloud Hypervisor in order to avoid complexity, thus I've adopted the very same approach already adopted by Firecracker, that's using a constant refill timer of 1000ms (or 1s). 

Last but not least, I'm using `bits/sec` for the bandwidth for the sake of compatibility with what's already present for the other VMMs.

Please, @skaegi, @egernst, @likebreath, take a look at the series and let me know if there's something you'd like to see improved and / or done in a different way.


fidencio:
I am adding the `do-not-merge` label as I want to give it a solid week so folks have enough time to review and experiment with the series. but it should be considered ready.

fidencio:
`s/BW/Bw/g` everywhere so the static checks are happy with the series.

likebreath (approved):
LGTM. Very clear commit messages and comments. Thanks @fidencio.

likebreath:
Please also keep us posted on how the rate limiter performs if any of you got time to play with it. Also, if there are improvements may need from CH to better support your usage cases. Please feel free to share or open a CH issue.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -25,6 +25,8 @@ const cpBinaryName = "cp"
 
 const fileMode0755 = os.FileMode(0755)
 
+const DefaultRateLimiterRefillTime = 1000 //miliseconds
```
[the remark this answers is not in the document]
fidencio: Done!

fidencio:
I've update this series exposing *also* the I/O throttling options for the disk.

likebreath (approved):
Still LGTM.

fidencio:
When starting a pod with the following configuration ...
```
# Bandwidth rate limiter options
# 
# net_rate_limiter_bw_max_rate controls network I/O bandwidth (size in bits/sec
# for SB/VM).
# The same value is used for inbound and outbound bandwidth.
# Default 0-sized value means unlimited rate.
net_rate_limiter_bw_max_rate = 1024
#
# net_rate_limiter_bw_one_time_burst increases the initial max rate and this
# initial extra credit does *NOT* and can be used for an *initial* burst of
# data.
# This is *optional* and only takes effect if net_rate_limiter_bw_max_rate is
# set.
net_rate_limiter_bw_one_time_burst = 10240
#
# Operation rate limiter options
#
# net_rate_limiter_ops_max_rate controls network I/O bandwidth (size in ops/sec
# for SB/VM).
# The same value is used for inbound and outbound bandwidth.
# Default 0-sized value means unlimited rate.
net_rate_limiter_ops_max_rate = 1000
#
# net_rate_limiter_ops_one_time_burst increases the initial max rate and this
# initial extra credit does *NOT* and can be used for an *initial* burst of
# data.
# This is *optional* and only takes effect if net_rate_limiter_bw_max_rate is
# set.
net_rate_limiter_ops_one_time_burst = 10000
#
# These options are related to disk rate limiter at the VMM level, and are
# based on the Cloud Hypervisor I/O throttling.  Those are disabled by default
# and we strongly advertise users to refer the Cloud Hypervisor official
# documentation for a better understanding of its internals:
# https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/io_throttling.md
# 
# Bandwidth rate limiter options
# 
# disk_rate_limiter_bw_max_rate controls disk I/O bandwidth (size in bits/sec
# for SB/VM).
# The same value is used for inbound and outbound bandwidth.
# Default 0-sized value means unlimited rate.
disk_rate_limiter_bw_max_rate = 1048576
#
# disk_rate_limiter_bw_one_time_burst increases the initial max rate and this
# initial extra credit does *NOT* and can be used for an *initial* burst of
# data.
# This is *optional* and only takes effect if disk_rate_limiter_bw_max_rate is
# set.
disk_rate_limiter_bw_one_time_burst = 10240
#
# Operation rate limiter options
#
# disk_rate_limiter_ops_max_rate controls disk I/O bandwidth (size in ops/sec
# for SB/VM).
# The same value is used for inbound and outbound bandwidth.
# Default 0-sized value means unlimited rate.
disk_rate_limiter_ops_max_rate = 1000000
#
# disk_rate_limiter_ops_one_time_burst increases the initial max rate and this
# initial extra credit does *NOT* and can be used for an *initial* burst of
# data.
# This is *optional* and only takes effect if disk_rate_limiter_bw_max_rate is
# set.
disk_rate_limiter_ops_one_time_burst = 10000
```

I get the following on the cloud-hypervisor side (please, search for "rate_limiter_config"):
```
$ ch-remote --api-socket /run/vc/vm/f5d0477fd25ee5cb19a6202820e32ec9bc7951f3fe72419a2e9a61c772fccaa0/clh-api.sock info | jq
{
  "config": {
    "cpus": {
      "boot_vcpus": 1,
      "max_vcpus": 4,
      "topology": {
        "threads_per_core": 1,
        "cores_per_die": 4,
        "dies_per_package": 1,
        "packages": 1
      },
      "kvm_hyperv": false,
      "max_phys_bits": 46,
      "affinity": null,
      "features": {}
    },
    "memory": {
      "size": 2147483648,
      "mergeable": false,
      "hotplug_method": "Acpi",
      "hotplug_size": 8348377088,
      "hotplugged_size": null,
      "shared": true,
      "hugepages": false,
      "hugepage_size": null,
      "prefault": false,
      "zones": null
    },
    "kernel": {
      "path": "/usr/share/kata-containers/vmlinux-5.15.26-90"
    },
    "initramfs": null,
    "cmdline": {
      "args": "root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro ro rootfstype=ext4 panic=1 no_timer_check noreplace-smp console=ttyS0,115200n8 systemd.log_target=console systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket agent.log=debug agent.debug_console agent.debug_console_vport=1026"
    },
    "disks": [
      {
        "path": "/dev/dm-30",
        "readonly": false,
        "direct": false,
        "iommu": false,
        "num_queues": 1,
        "queue_size": 128,
        "vhost_user": false,
        "vhost_socket": null,
        "poll_queue": true,
        "rate_limiter_config": {
          "bandwidth": {
            "size": 134216,
            "one_time_burst": 1304,
            "refill_time": 1000
          },
          "ops": {
            "size": 1000000,
            "one_time_burst": 10000,
            "refill_time": 1000
          }
        },
        "id": "clh_drive_0",
        "disable_io_uring": false,
        "pci_segment": 0
      },
      {
        "path": "/dev/dm-38",
        "readonly": false,
        "direct": false,
        "iommu": false,
        "num_queues": 1,
        "queue_size": 128,
        "vhost_user": false,
        "vhost_socket": null,
        "poll_queue": true,
        "rate_limiter_config": {
          "bandwidth": {
            "size": 134216,
            "one_time_burst": 1304,
            "refill_time": 1000
          },
          "ops": {
            "size": 1000000,
            "one_time_burst": 10000,
            "refill_time": 1000
          }
        },
        "id": "clh_drive_1",
        "disable_io_uring": false,
        "pci_segment": 0
      }
    ],
    "net": [
      {
        "tap": "tap0_kata",
        "ip": "192.168.249.1",
        "mask": "255.255.255.0",
        "mac": "72:f6:c4:c8:b4:4e",
        "host_mac": "8e:fd:3e:81:9e:49",
        "iommu": false,
        "num_queues": 2,
        "queue_size": 256,
        "vhost_user": false,
        "vhost_socket": null,
        "vhost_mode": "Client",
        "id": "_net0",
        "fds": null,
        "rate_limiter_config": {
          "bandwidth": {
            "size": 128,
            "one_time_burst": 1304,
            "refill_time": 1000
          },
          "ops": {
            "size": 1000,
            "one_time_burst": 10000,
            "refill_time": 1000
          }
        },
        "pci_segment": 0
      }
    ],
    "rng": {
      "src": "/dev/urandom",
      "iommu": false
    },
    "balloon": null,
    "fs": [
      {
        "tag": "kataShared",
        "socket": "/run/vc/vm/f5d0477fd25ee5cb19a6202820e32ec9bc7951f3fe72419a2e9a61c772fccaa0/virtiofsd.sock",
        "num_queues": 1,
        "queue_size": 1024,
        "dax": false,
        "cache_size": 0,
        "id": "_fs1",
        "pci_segment": 0
      }
    ],
    "pmem": [
      {
        "file": "/usr/share/kata-containers/kata-containers.img",
        "size": null,
        "iommu": false,
        "mergeable": false,
        "discard_writes": true,
        "id": "_pmem2",
        "pci_segment": 0
      }
    ],
    "serial": {
      "file": null,
      "mode": "Tty",
      "iommu": false
    },
    "console": {
      "file": null,
      "mode": "Off",
      "iommu": false
    },
    "devices": null,
    "user_devices": null,
    "vdpa": null,
    "vsock": {
      "cid": 3,
      "socket": "/run/vc/vm/f5d0477fd25ee5cb19a6202820e32ec9bc7951f3fe72419a2e9a61c772fccaa0/clh.sock",
      "iommu": false,
      "id": "_vsock3",
      "pci_segment": 0
    },
    "iommu": false,
    "sgx_epc": null,
    "numa": null,
    "watchdog": false,
    "tdx": null,
    "platform": null
  },
  "state": "Running",
  "memory_actual_size": 2147483648,
  "device_tree": {
    "_pmem2": {
      "id": "_pmem2",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 70368366690304,
            "size": 134217728
          }
        }
      ],
      "parent": "_virtio-pci-_pmem2",
      "children": [],
      "pci_bdf": null
    },
    "clh_drive_0": {
      "id": "clh_drive_0",
      "resources": [],
      "parent": "_virtio-pci-clh_drive_0",
      "children": [],
      "pci_bdf": null
    },
    "_net0": {
      "id": "_net0",
      "resources": [],
      "parent": "_virtio-pci-_net0",
      "children": [],
      "pci_bdf": null
    },
    "_virtio-pci-_fs1": {
      "id": "_virtio-pci-_fs1",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 70368500908032,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "_fs1"
      ],
      "pci_bdf": "0000:00:03.0"
    },
    "_ioapic": {
      "id": "_ioapic",
      "resources": [],
      "parent": null,
      "children": [],
      "pci_bdf": null
    },
    "_fs1": {
      "id": "_fs1",
      "resources": [],
      "parent": "_virtio-pci-_fs1",
      "children": [],
      "pci_bdf": null
    },
    "_vsock3": {
      "id": "_vsock3",
      "resources": [],
      "parent": "_virtio-pci-_vsock3",
      "children": [],
      "pci_bdf": null
    },
    "_virtio-pci-_vsock3": {
      "id": "_virtio-pci-_vsock3",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 70368365641728,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "_vsock3"
      ],
      "pci_bdf": "0000:00:05.0"
    },
    "_virtio-pci-_pmem2": {
      "id": "_virtio-pci-_pmem2",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 70368366166016,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "_pmem2"
      ],
      "pci_bdf": "0000:00:04.0"
    },
    "_serial": {
      "id": "_serial",
      "resources": [],
      "parent": null,
      "children": [],
      "pci_bdf": null
    },
    "_virtio-pci-clh_drive_0": {
      "id": "_virtio-pci-clh_drive_0",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 3221225472,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "clh_drive_0"
      ],
      "pci_bdf": "0000:00:06.0"
    },
    "clh_drive_1": {
      "id": "clh_drive_1",
      "resources": [],
      "parent": "_virtio-pci-clh_drive_1",
      "children": [],
      "pci_bdf": null
    },
    "_rng": {
      "id": "_rng",
      "resources": [],
      "parent": "_virtio-pci-_rng",
      "children": [],
      "pci_bdf": null
    },
    "_virtio-pci-_net0": {
      "id": "_virtio-pci-_net0",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 70368501956608,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "_net0"
      ],
      "pci_bdf": "0000:00:01.0"
    },
    "_virtio-pci-_rng": {
      "id": "_virtio-pci-_rng",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 70368501432320,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "_rng"
      ],
      "pci_bdf": "0000:00:02.0"
    },
    "_virtio-pci-clh_drive_1": {
      "id": "_virtio-pci-clh_drive_1",
      "resources": [
        {
          "MmioAddressRange": {
            "base": 3221749760,
            "size": 524288
          }
        }
      ],
      "parent": null,
      "children": [
        "clh_drive_1"
      ],
      "pci_bdf": "0000:00:07.0"
    }
  }
}
```

fidencio:
I guess this is obvious, but it's still important to note, **if the VM rate_limit is too low the VM won't even start**, but that's expected.  Maybe would be worth adding this to the documentation?

pmores:
@fidencio Thanks for the very clean and easy-to-review PR! I'll add some comments to specific chunks shortly but overall, I'm just wondering if it wouldn't have been possible to reuse the existing fields after all and just document that for hypervisors that don't support asymmetric rate limiting, just, say, Tx value is used for both directions and Rx is ignored?

Perhaps that way we'd have fewer fields/config knobs and all hypervisor would be equally supported.  With the approach of this PR, I suspect that hypervisors that do support asymmetric rate limiting are now less supported (I don't know if this might be an actual problem though).

fidencio:
> @fidencio Thanks for the very clean and easy-to-review PR! I'll add some comments to specific chunks shortly but overall, I'm just wondering if it wouldn't have been possible to reuse the existing fields after all and just document that for hypervisors that don't support asymmetric rate limiting, just, say, Tx value is used for both directions and Rx is ignored?

I sincerely think that would be *super* confusing.
Plus, when exposing the operations limiter we'd be starting with something that's wrong, copying Rx / Tx there, or differ between bandwidth / operations, which is even worse.

> Perhaps that way we'd have fewer fields/config knobs and all hypervisor would be equally supported. With the approach of this PR, I suspect that hypervisors that do support asymmetric rate limiting are now less supported (I don't know if this might be an actual problem though).

So, the changes added are *specific* to Cloud Hypervisor and *not* exposed for other VMMs.  The other VMMs are as supported as they were before, nothing changes for them.

src/runtime/pkg/katautils/config.go
```diff
@@ -490,6 +494,34 @@ func (h hypervisor) getTxRateLimiterCfg() uint64 {
...
+func (h hypervisor) getNetRateLimiterBwMaxRate() uint64 {
+	return h.NetRateLimiterBwMaxRate
+}
+
+func (h hypervisor) getNetRateLimiterBwOneTimeBurst() uint64 {
+	if h.NetRateLimiterBwOneTimeBurst != 0 && h.getNetRateLimiterBwMaxRate() == 0 {
+		kataUtilsLogger.Warn("The NetRateLimiterBwOneTimeBurst is set but NetRateLimiterBwMaxRate is not set, this option will be ignored.")
+
+		h.NetRateLimiterBwOneTimeBurst = 0
+	}
```
pmores, line 538: I'd consider it cleaner if this check (and other similar in this PR) were performed in a one-off sanity checking/consistency fixing section after a config is loaded instead of on each get.

OTOH I suspect that these gets aren't called very often so it might not make a huge difference in practice.  I still thought that I'd mention it anyway for sake of conceptual cleanliness.

src/runtime/virtcontainers/hypervisor.go
```diff
@@ -386,6 +386,24 @@ type HypervisorConfig struct {
...
+	// NetRateLimiterBwRate is used to control network I/O bandwidth on VM level.
+	// The same value, defined in bits per second, is used for inbound and outbound bandwidth.
+	NetRateLimiterBwMaxRate uint64
+
+	// NetRateLimiterBwOneTimeBurst is used to control network I/O bandwidth on VM level.
+	// This increases the initial max rate and this initial extra credit does *NOT* replenish
+	// and can be used for an *initial* burst of data.
+	NetRateLimiterBwOneTimeBurst uint64
+
+	// NetRateLimiterOpsRate is used to control network I/O bandwidth on VM level.
```
pmores, line 416: s/bandwidth/operations per second/ ?

If so then the same applies to `NetRateLimiterOpsOneTimeBurst` below and their equivalents for disk I/O.

Given that "bandwidth" and "operations per second" are clearly distinct concepts in this PR I think it's worth the trouble to avoid any risk of mixing them up.

fidencio: Very nice catch!

src/runtime/virtcontainers/clh_test.go
```diff
@@ -191,6 +195,193 @@ func TestCloudHypervisorAddNetCheckEnpointTypes(t *testing.T) {
...
+	}
+
+	//nolint: govet
+	tests := []struct {
+		name                  string
+		args                  args
+		expectsRateLimiter    bool
+		expectsBwBucketToken  bool
+		expectsOpsBucketToken bool
+		wantErr               bool
```
pmores, line 221: IIUC this field is not actually used by any tests.  If so, is that intentional?

src/runtime/pkg/katautils/config.go
```diff
@@ -74,74 +74,78 @@ type factory struct {
...
+	EnableIOThreads                bool     `toml:"enable_iothreads"`
+	DisableImageNvdimm             bool     `toml:"disable_image_nvdimm"`
+	HotplugVFIOOnRootBus           bool     `toml:"hotplug_vfio_on_root_bus"`
+	DisableVhostNet                bool     `toml:"disable_vhost_net"`
+	GuestMemoryDumpPaging          bool     `toml:"guest_memory_dump_paging"`
+	ConfidentialGuest              bool     `toml:"confidential_guest"`
+	GuestSwap                      bool     `toml:"enable_guest_swap"`
+	Rootless                       bool     `toml:"rootless"`
+	DisableSeccomp                 bool     `toml:"disable_seccomp"`
+	DisableSeLinux                 bool     `toml:"disable_selinux"`
```
pmores, line 148: On a personal note, these huge chunks are part of the reason I've learnt to dislike spurious formatting in source code.  For an uninitiated, it takes a bit of staring into this to realise that the whole thing is (probably) just about adding a couple of lines.  Anyway, just saying. ;-)

pmores:
> > @fidencio Thanks for the very clean and easy-to-review PR! I'll add some comments to specific chunks shortly but overall, I'm just wondering if it wouldn't have been possible to reuse the existing fields after all and just document that for hypervisors that don't support asymmetric rate limiting, just, say, Tx value is used for both directions and Rx is ignored?
> 
> I sincerely think that would be **super** confusing. Plus, when exposing the operations limiter we'd be starting with something that's wrong, copying Rx / Tx there; or differ between bandwidth / operations, which is even worse.

I'm not sure that I'm following you here but that's fine - if you thought about this aspect and arrived to the conclusion that this is the best way then that's what I wanted to know.
 
> > Perhaps that way we'd have fewer fields/config knobs and all hypervisor would be equally supported. With the approach of this PR, I suspect that hypervisors that do support asymmetric rate limiting are now less supported (I don't know if this might be an actual problem though).
> 
> So, the changes added are **specific** to Cloud Hypervisor and **not** exposed to the other VMMs. The other VMMs are as supported as they were before, nothing changes for them.

Alright, my point here is that the `configuration.toml` syntax and the corresponding `hypervisor` struct are common to all hypervisors, right?  And right now, if a hypervisor supports asymmetric rate limiting (now or in the future) and we want to expose it the way this PR exposes rate limiting for CLH, we don't have a way to do that without adding yet another set of similar fields, just this time separate for tx and rx.  Does that make sense, or am I completely off here?

fidencio:
> > > @fidencio Thanks for the very clean and easy-to-review PR! I'll add some comments to specific chunks shortly but overall, I'm just wondering if it wouldn't have been possible to reuse the existing fields after all and just document that for hypervisors that don't support asymmetric rate limiting, just, say, Tx value is used for both directions and Rx is ignored?
> > 
> > 
> > I sincerely think that would be **super** confusing. Plus, when exposing the operations limiter we'd be starting with something that's wrong, copying Rx / Tx there; or differ between bandwidth / operations, which is even worse.
> 
> I'm not sure that I'm following you here but that's fine - if you thought about this aspect and arrived to the conclusion that this is the best way then that's what I wanted to know.

Okay.  Basically what I was trying to say is that we have two options:
* Keep and document the Rx / Tx queues, as we already do for bandwidth for FC and QEMU, but only use one of those for Cloud Hypervisor.
  * This, IMHO, looks confusing.
  * When adding the bandwidth rate limiter, we'd already add new fields, and those would, for consistency sake, be tied to Rx / Tx queues, with only one being used, which is also confusing.
* Add the new fields as I did.

I must say that I was not particularly happy with the difference, but it's specific to the hypervisor and I decided to go to the way that would be clear for the user / admin on what they're setting.

So, yeah, I put some thought into this before starting implementing it.  I guess I took what I considered the less worse approach for the user / admin.

> 
> > > Perhaps that way we'd have fewer fields/config knobs and all hypervisor would be equally supported. With the approach of this PR, I suspect that hypervisors that do support asymmetric rate limiting are now less supported (I don't know if this might be an actual problem though).
> > 
> > 
> > So, the changes added are **specific** to Cloud Hypervisor and **not** exposed to the other VMMs. The other VMMs are as supported as they were before, nothing changes for them.
> 
> Alright, my point here is that the `configuration.toml` syntax and the corresponding `hypervisor` struct are common to all hypervisors, right? And right now, if a hypervisor supports asymmetric rate limiting (now or in the future) and we want to expose it the way this PR exposes rate limiting for CLH, we don't have a way to do that without adding yet another set of similar fields, just this time separate for tx and rx. Does that make sense, or am I completely off here?

For asymmetric rate limiting folks should use the **already** provided Rx / Tx, rather than exposing new fields.
If QEMU or FC decides to expose "operations", though, they most likely will add two new fields, for Rx / Tx.  I know, two more fields, but I still think it's the most clean way for the admin / user, considering the difference exists between the VMMs themselves.

jodh-intel (commented):

pmores:
> > > > @fidencio Thanks for the very clean and easy-to-review PR! I'll add some comments to specific chunks shortly but overall, I'm just wondering if it wouldn't have been possible to reuse the existing fields after all and just document that for hypervisors that don't support asymmetric rate limiting, just, say, Tx value is used for both directions and Rx is ignored?
> > > 
> > > 
> > > I sincerely think that would be **super** confusing. Plus, when exposing the operations limiter we'd be starting with something that's wrong, copying Rx / Tx there; or differ between bandwidth / operations, which is even worse.
> > 
> > 
> > I'm not sure that I'm following you here but that's fine - if you thought about this aspect and arrived to the conclusion that this is the best way then that's what I wanted to know.
> 
> Okay. Basically what I was trying to say is that we have two options:
> 
>     * Keep and document the Rx / Tx queues, as we already do for bandwidth for FC and QEMU, but only use one of those for Cloud Hypervisor.
>       
>       * This, IMHO, looks confusing.
>       * When adding the bandwidth rate limiter, we'd already add new fields, and those would, for consistency sake, be tied to Rx / Tx queues, with only one being used, which is also confusing.
> 
>     * Add the new fields as I did.
> 
> 
> I must say that I was not particularly happy with the difference, but it's specific to the hypervisor and I decided to go to the way that would be clear for the user / admin on what they're setting.
> 
> So, yeah, I put some thought into this before starting implementing it. I guess I took what I considered the less worse approach for the user / admin.

Fair enough, there is no clean solution here - we're basically replicating here a textbook problem of object-oriented design where you have a `class Rectangle` with `width()` and `height()` and now you want to use it to represent a square as well. ;-)

So I guess it's a matter of taste, and I guess my idea was, if we can only have solutions that are more or less confusing, let's go with the one that implies fewer fields.  But I'm not arguing, let's consider this settled.  I'll shut up now. :-)

> > > > Perhaps that way we'd have fewer fields/config knobs and all hypervisor would be equally supported. With the approach of this PR, I suspect that hypervisors that do support asymmetric rate limiting are now less supported (I don't know if this might be an actual problem though).
> > > 
> > > 
> > > So, the changes added are **specific** to Cloud Hypervisor and **not** exposed to the other VMMs. The other VMMs are as supported as they were before, nothing changes for them.
> > 
> > 
> > Alright, my point here is that the `configuration.toml` syntax and the corresponding `hypervisor` struct are common to all hypervisors, right? And right now, if a hypervisor supports asymmetric rate limiting (now or in the future) and we want to expose it the way this PR exposes rate limiting for CLH, we don't have a way to do that without adding yet another set of similar fields, just this time separate for tx and rx. Does that make sense, or am I completely off here?
> 
> For asymmetric rate limiting folks should use the **already** provided Rx / Tx, rather than exposing new fields. If QEMU or FC decides to expose "operations", though, they most likely will add two new fields, for Rx / Tx. I know, two more fields, but I still think it's the most clean way for the admin / user, considering the difference exists between the VMMs themselves.

And the same goes for burst values - all in all, quite a bunch of fields.  But as I said above, let's consider this settled.

src/runtime/virtcontainers/clh.go
```diff
@@ -442,6 +442,16 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
 			disk := chclient.NewDiskConfig(imagePath)
 			disk.SetReadonly(true)
 
+			diskRateLimiterConfig := clh.getRateLimiterConfig(
+				int64(utils.RevertBytes(clh.config.DiskRateLimiterBwMaxRate/8)),
+				int64(utils.RevertBytes(clh.config.DiskRateLimiterBwOneTimeBurst/8)),
+				int64(clh.config.DiskRateLimiterOpsMaxRate),
+				int64(clh.config.DiskRateLimiterOpsOneTimeBurst))
```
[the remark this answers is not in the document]
fidencio: This requires a change on Cloud Hypervisor (see: https://github.com/cloud-hypervisor/cloud-hypervisor/pull/4019) and I will update my series pulling in the version that containers the fix, as soon as the fix is approved.

fidencio:
@pmores, @jodh-intel, thanks a whole lot for the reviews!
I will address the comments shortly after I get the PR merged on the cloud-hypervisor side.

src/runtime/virtcontainers/clh.go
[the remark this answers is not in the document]
fidencio: Okay, it seems openAPI doesn't support uint64 as it simply interprets that as int32.  Yes, uint64 -> int32.
I'm going to propose to revert the change on Cloud Hypervisor side and leave this as it's for now.

A next step will be evaluate whether keep using openAPI makes sense, as swagger is what's used for Firecracker, and maybe we should align on that.  But that will be a whole lot of work that may, or may not, be worth it, mainly considering that soon enough we hopefully will be using the rust runtime.

I will address all the other comments.

src/runtime/pkg/katautils/config.go
fidencio: I've left this as it's as we have a few other getters doing the same thing already.

src/runtime/virtcontainers/clh_test.go
```diff
@@ -191,6 +195,181 @@ func TestCloudHypervisorAddNetCheckEnpointTypes(t *testing.T) {
...
+			clhConfig.NetRateLimiterOpsMaxRate = tt.args.opsMaxRate
+			clhConfig.NetRateLimiterOpsOneTimeBurst = tt.args.opsOneTimeBurst
+
+			clh := &cloudHypervisor{}
+			clh.config = clhConfig
+			clh.APIClient = &clhClientMock{}
+
+			if err := clh.addNet(validVeth); err != nil {
+				t.Errorf("cloudHypervisor.addNet() error = %v", err)
+			} else if err == nil {
```
pmores, line 347: I believe this `else if`  could now become a simple `else` as any non-nil `err` value has already been handled by the previous branch, right?

src/runtime/virtcontainers/clh_test.go
```diff
@@ -191,6 +195,181 @@ func TestCloudHypervisorAddNetCheckEnpointTypes(t *testing.T) {
...
+						opsBucketToken := rateLimiterConfig.GetOps()
+						assert.Equal(opsBucketToken.GetSize(), int64(tt.args.opsMaxRate))
+						assert.Equal(opsBucketToken.GetOneTimeBurst(), int64(tt.args.opsOneTimeBurst))
+					}
+				}
+			}
+		})
+	}
+}
+
```
pmores, line 372: Also, shouldn't there be an analogous `TestCloudHypervisorDiskRateLimiter()` to test the disk part of the PR?

src/runtime/virtcontainers/clh_test.go
fidencio: Oh yeah, this slipped out when reworking the PR, thanks for noticiing!

src/runtime/virtcontainers/clh_test.go
fidencio: This is really not easy to implement, @pmores.

Right now the main path we change this is when we hotplug a virtio-blk device and mocking that path has proved to be not exactly trivial.

src/runtime/virtcontainers/clh.go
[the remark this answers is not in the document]
jodh-intel: > openAPI doesn't support uint64 as it simply interprets that as int32. Yes, uint64 -> int32.

omg ouch^W^W ack :smile:

jodh-intel (approved):
Thanks @fidencio.

src/runtime/virtcontainers/clh_test.go
fidencio: Anyways, let me take a second stab at this with fresh eyes.

pmores: Right, I suspected something like that would be the case. :-) You could probably still write tests targeting the layer right below the one you're mentioning - but I guess those tests would be close to trivial so I'd think this is fair enough.

Thanks for the explanation!

fidencio: Yeah, really not easy and I'd have to expand a whole lot on the test suit to have something that's meaningful.
I'm not saying it's not something worth doing, but I must admit I don't think it's worth doing as part of this PR.

Is that okay for you?

src/runtime/virtcontainers/clh_test.go
fidencio: Fixed this one in the last push.

fidencio:
Last push was fixing `gofmt` issues.

src/runtime/virtcontainers/clh_test.go
pmores: Sure, thanks!

pmores (approved):
Thanks @fidencio !

Issue #4140, opened 2022-04-22
Go unit tests should not leave stray files and/or mountpoints in /tmp


dgibson:
# Description of problem

Unit tests, being unit tests, should be self-contained and not leave material on the system that could interfere with future runs of the same tests (or anything else).  However, the current Go unit tests leave behind a number of files in `/tmp`.  Most are in uniquely named temporary directories, which are relatively harmless but still messy.  However some are in subdirectories of `/tmp` with fixed names which can interfere with future test runs (or concurrent test runs).  When run as root some test cases also leave behind mount points under `/tmp`, which require explicit removal, and so are also more problematic.

[Note, some unit tests also leave behind files and mountpoints outside of `/tmp`, including notably network namespaces.  That's really a worse problem that the stuff under `/tmp`, but this issue is only covering the things under `/tmp` - have to start somewhere].

PR #4141, opened 2022-04-22
Fix Go unit tests to clean up /tmp after themselves

dgibson:
A number of the Go unit tests leave things behind in `/tmp`.  As described in #4140, this isn't ideal.  Fix them to clean up after themselves.

### DIFF
--- a/src/runtime/virtcontainers/persist/fs/mockfs_test.go
+++ b/src/runtime/virtcontainers/persist/fs/mockfs_test.go
@@ -13,19 +13,19 @@ import (
 
 func TestMockAutoInit(t *testing.T) {
 	assert := assert.New(t)
-	orgMockTesting := mockTesting
+	orgMockRootPath := mockRootPath
 	defer func() {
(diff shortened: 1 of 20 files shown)


src/runtime/virtcontainers/persist/fs/mockfs_test.go
```diff
@@ -0,0 +1,34 @@
+// Copyright Red Hat.
```
liubin, line 1: Should add `2022` here?

liubin (approved):
@dgibson The only change need to consider is the license header in file `src/runtime/virtcontainers/persist/fs/mockfs_test.go`

src/runtime/virtcontainers/persist/fs/mockfs_test.go
dgibson: Not according to the guidelines from Red Hat legal.

bergwolf (approved):
lgtm!

Issue #366, opened 2020-06-30
shimv2: Improve debuggability


fidencio:
**Which feature do you think can be improved?**

Currently `containerd-shim-kata-v2` binary has as an option `-debug`. On CRI-O side, and possibly on contanerd as well, `-debug` is passed down when calling the binary whenever CRI-O log level is set to `debug`.

**How can it be improved?**

What's described above is horrible as we either don't have enough debug info, or we have too much debug info.
CRI-O supports the following debug levels: fatal, panic, error, warn, info, debug and trace. Kata supports at least warn, info, and debug. Knowing that, would be really good to have a way to make the CRI-O (or containerd) log level match with the Kata one.

**Additional Information**

This can be achieved in different ways, such as:
- Adding `-info`, `-warn`, command line options;
- Deprecated `-debug` command line option, and adding a `--log-level` option;
- Using shimv2 CreateTaskRequest API, passing this value as part of the `Options` field;

I'd like to have the discussion here on what would be the preferred way to proceed with this, and also learn about possible limitations that I'm not considering.

liubin:
How about add an argument in `configuration.toml`, this will not need to modify caller(e.g., `containerd`).

marcel-apf:
Hi! I want to "handle" this task as a way to start to contributing to the Kata project. Thanks @fidencio that pointed me to it.

bergwolf:
Please note that modifying `CreateTaskRequest` API may introduce backward incompatiblity.

fidencio:
@bergwolf, sure, we're aiming for the getting @liubin's suggestion implemented.

ariel-adam:
@marcel-apf Hi Marcel, how is it going with this issue :-) ?

marcel-apf:
> @marcel-apf Hi Marcel, how is it going with this issue :-) ?

The issue is blocked by a prerequisite bug: https://github.com/kata-containers/runtime/issues/3039 that has an outstanding PR: https://github.com/kata-containers/runtime/pull/3040.

In the meantime v1 is not that relevant, so I am considering "porting" the PR to v2, I will contact @fidencio  for some assistance.

cmaf:
@marcel-apf are there still plans for a PR, or has this been resolved possibly in PRs not linked to this particular issue? We can close it if it is no longer relevant.

PR #4148, opened 2022-04-25
runtime: add runtime loglevel config in the configuration.toml

flyflypeng:
reason: currently containerd-shim-kata-v2 binary has as an option
` -debug` to control print runtime log into the journal log.
However, runtime default log level is `warn`, which is too little
for debugging usage.
This patch adds a new config `loglevel` under runtime section  in
the configuration.toml to specifies the log level, which makes user
to select the log level simply.

Fixes: #366

Signed-off-by: flyflypeng <flyflyflypeng@gmail.com>

### DIFF
--- a/src/runtime/pkg/katautils/config.go
+++ b/src/runtime/pkg/katautils/config.go
@@ -149,6 +149,7 @@ type runtime struct {
 	SandboxBindMounts         []string `toml:"sandbox_bind_mounts"`
 	Experimental              []string `toml:"experimental"`
 	Debug                     bool     `toml:"enable_debug"`
+	LogLevel                  string   `toml:"loglevel"`
 	Tracing                   bool     `toml:"enable_tracing"`
 	DisableNewNetNs           bool     `toml:"disable_new_netns"`
 	DisableGuestSeccomp       bool     `toml:"disable_guest_seccomp"`
@@ -1104,7 +1105,14 @@ func LoadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
 	}
 
 	config.Debug = tomlConf.Runtime.Debug
-	if !tomlConf.Runtime.Debug {
+	config.LogLevel, err = logrus.ParseLevel(tomlConf.Runtime.LogLevel)
+	if err != nil {
+		return "", oci.RuntimeConfig{}, err
+	}
+
+	if tomlConf.Runtime.Debug {
+		kataUtilsLogger.Logger.Level = config.LogLevel
+	} else {
 		// If debug is not required, switch back to the original
 		// default log priority, otherwise continue in debug mode.
 		kataUtilsLogger.Logger.Level = originalLoggerLevel
--- a/src/runtime/config/configuration-acrn.toml.in
+++ b/src/runtime/config/configuration-acrn.toml.in
@@ -154,6 +154,12 @@ block_device_driver = "@DEFBLOCKSTORAGEDRIVER_ACRN@"
 # (default: disabled)
 #enable_debug = true
 #
+# Specifies the runtime debug log level
+# loglevel is valid only with enable_debug config is true
+# loglevel optional values: panic, fatal, error, warn/warning, info, debug, trace
+# (default: "warn")
+#loglevel = "warn"
+#
 # Internetworking model
 # Determines how the VM should be connected to the
 # the container network interface
--- a/src/runtime/config/configuration-clh.toml.in
+++ b/src/runtime/config/configuration-clh.toml.in
@@ -216,6 +216,12 @@ block_device_driver = "virtio-blk"
 # (default: disabled)
 #enable_debug = true
 #
+# Specifies the runtime debug log level
+# loglevel is valid when enable_debug config is true
+# loglevel optional values: panic, fatal, error, warn/warning, info, debug, trace
+# (default: "warn")
+#loglevel = "warn"
+#
 # Internetworking model
 # Determines how the VM should be connected to the
 # the container network interface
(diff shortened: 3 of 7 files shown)


fidencio:
@flyflypeng, the way Kata Containers works nowadays is by setting the `enable_debug` flag for the components you need, considering that `-debug` was passed to the `containerd-shim-kata-v2` binary, which has to be set in the CRI engine (CRI-O or containerd).

I, myself, am not against this change but from a quick look at the code it doesn't reflect into the agent code.  Or did I miss it?


I would really like to hear inputs from @amshinde, @bergwolf, @egernst, and @sameo

src/runtime/pkg/katautils/config.go
```diff
@@ -1104,7 +1105,14 @@ func LoadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
...
 
 	config.Debug = tomlConf.Runtime.Debug
-	if !tomlConf.Runtime.Debug {
+	config.LogLevel, err = logrus.ParseLevel(tomlConf.Runtime.LogLevel)
+	if err != nil {
+		return "", oci.RuntimeConfig{}, err
+	}
+
+	if tomlConf.Runtime.Debug {
+		kataUtilsLogger.Logger.Level = config.LogLevel
```
jodh-intel, line 1114: This is neat as it doesn't break existing users but it's slightly confusing because you need to enable debug to change the level to something other than debug.

I wonder if instead we should just deprecate `enable_debug`. Something like:

```go
if tomlConf.Runtime.LogLevel != "" {
    kataUtilsLogger.Logger.Level = config.LogLevel
} else if tomlConf.Runtime.Debug {
     kataUtilsLogger.Logger.Level = Level.DebugLevel
}
```

If we do something like that shown above, we should also display a warning message similar to:

```
'enable_debug=' is deprecated - use 'loglevel=' instead"
```

... but note that actually outputting that message needs care since:

- This code is setting up the log levels and we always want that message to be displayed.
- Displaying the message "too early" will upset the container manager, which expects certain early output from the runtime and will fail if it sees something it doesn't expect.

We'd also need to think carefully about actually _removing_ `enable_debug=` since it would need changes to atleast:

- Docs.
- A bunch of scripts in the tests repo.

wdyt @amshinde, @bergwolf, @fidencio, @egernst?

jodh-intel (commented):
Hi @flyflypeng - Thanks for raising. This is something we've been meaning to add for a long time now ;) I've added a comment but I agree with @fidencio that it would probably be best if we update the code for the runtime and the agent at the same time for consistency.

The code for the agent is here:

- https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/kata_agent.go#L312..L314

Basically, we need to pass `"agent.log=$level"` rather than hard-coding `"debug"`.

src/runtime/pkg/katautils/config.go
amshinde: @jodh-intel  +1 from me on removing enable_debug altogether although we may need to make quite a few changes in our logs and testing scripts. Having the ability to set the log level as added by this PR is of value to the user but having the enable_debug option is quite confusing and may conflict with the log level set by the option introduced in this PR.

flyflypeng: @jodh-intel @fidencio I agree with replacing `enable_debug` config with `loglevel` config. If all committers reach an agreement with this change, I will modify the patch to support this change.

fidencio: @jodh-intel, worries me quite a bit we remove an option and add a totally different one in between **minor** releases.

This may break compatibility for users who have their custom config / custom scripts to change those knobs in their nodes.

I sincerely think something as huge as this should be targeting a **major** release instead, being material for 3.0 rather than 2.x.

fidencio:
I've added this PR for discussion in the next Architecture Committee meeting: https://etherpad.opendev.org/p/Kata_Containers_2022_Architecture_Committee_Mtgs#L23

fidencio:
We've discussed this change as part of the architecture committee meeting and we think we should go ahead with this PR.
We should **not** break the current approach, but having the possibility to set the log_level right now is something desired.

jodh-intel:
Looking forward to seeing this land @flyflypeng ! ;)

flyflypeng:
> We've discussed this change as part of the architecture committee meeting and we think we should go ahead with this PR. We should **not** break the current approach, but having the possibility to set the log_level right now is something desired.

According to your reply,  I am still confused about how to implement this PR, keep the `enable_debug` config option and add a new `log_level` option, or others?

flyflypeng:
> Looking forward to seeing this land @flyflypeng ! ;)

This may take some time， I am learning the Rust now which is a little hard for me.

liubin:
@flyflypeng CI has some errors:

> tainers/src/github.com/kata-containers/kata-containers/src/runtime/pkg/katatestutils
INFO: Running golangci-lint on /home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime/pkg/katautils
Error: config-settings.go:76:7: `defaultLogLevel` is unused (deadcode)
const defaultLogLevel string = "warn"
      ^
Error: config.go:143:14: fieldalignment: struct of size 160 could be 152 (govet)
type runtime struct {
             ^

fengwang666:
There is not any update for a few months. @flyflypeng Are you still able to finish it?

fengwang666:
fyi, since @flyflypeng hasn't replied yet, I volunteer to take over the PR and finish it.

yumingqiao:
Hi Kata community, I am working closely with @fengwang666. After coordinating with @fengwang666 , I volunteer to take over this change.

fidencio:
Closing this one in favour of https://github.com/kata-containers/kata-containers/pull/6122

Issue #4151, opened 2022-04-26
Broken fallback path in `build-static-clh.sh`


gkurz:
Since PR #3673 the clh tarball can be force-built from source. As part of this series, commit 948a2b099c6f introduced a typo that causes `build-static-clh.sh` to misbehave when called directly :

```
$ bash -x tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh

[..skip..]

INFO: Download cloud-hypervisor version: v23.0
+ cloud_hypervisor_binary=https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v23.0/cloud-hypervisor-static
+ curl --fail -L https://github.com/cloud-hypervisor/cloud-hypervisor/releases/download/v23.0/cloud-hypervisor-static -o cloud-hypervisor-static
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   662  100   662    0     0   4684      0 --:--:-- --:--:-- --:--:--  4695
100 5002k  100 5002k    0     0  11.3M      0 --:--:-- --:--:-- --:--:-- 11.3M
+ mkdir -p cloud-hypervisor

Let's create the dir

+ mv -f cloud-hypervisor-static cloud-hypervisor/cloud-hypervisor
+ chmod +x cloud_hypervisor/cloud-hypervisor
chmod: cannot access 'cloud_hypervisor/cloud-hypervisor': No such file or directory

Name should be 'cloud-hypervisor//cloud-hypervisor'

+ info 'Failed to pull cloud-hypervisor released binary, trying to build from source'
+ echo 'INFO: Failed to pull cloud-hypervisor released binary, trying to build from source'
INFO: Failed to pull cloud-hypervisor released binary, trying to build from source
+ build_clh_from_source
+ info 'Build https://github.com/cloud-hypervisor/cloud-hypervisor.git version: v23.0'
+ echo 'INFO: Build https://github.com/cloud-hypervisor/cloud-hypervisor.git version: v23.0'
INFO: Build https://github.com/cloud-hypervisor/cloud-hypervisor.git version: v23.0
++ basename https://github.com/cloud-hypervisor/cloud-hypervisor.git
+ repo_dir=cloud-hypervisor.git
+ repo_dir=cloud-hypervisor
+ '[' -d cloud-hypervisor ']'
+ pushd cloud-hypervisor

Reusing the previously created directory instead of cloning clh

[...]

PR #4152, opened 2022-04-26
packaging: Fix broken path in `build-static-clh.sh`

gkurz:
A recent change introduced a typo that could cause `pull_clh_released_binary()` to fail. This in turn revealed a flaw in `build_clh_from_source()` that would simply terminate the build.

This never did visible harm because kata-deploy force-builds clh from source for TDX. Fix these two anyway.

Fixes: #4151


c3d (approved):
Looks good to me. Tanks @gkurz

fidencio (approved):
lgtm, thanks @gkurz!

Issue #4080, opened 2022-04-12
Add support for pod specific IPtables


egernst:
We have a use case where we want to inject pod-specific ip table rules. In the case of runc this is as simple as creating the rules within the applicable network namespace.

For Kata this won't work, since the rules are managed managed at L3, while we use tcfilter at L2. Since we're doing a direct connection, no rules are exercised.

What I'd like to see is a mechanism to reflect the host side IPTables rules inside the guest, similar to what we do today with routes. In this way, we'd be able to effectively do "iptables-save" in the pod's network namespace on the host, and then have the agent do an iptables-restore inside the guest OS.

Further, there are use cases where the IPtables will be dynamic over time (Kube-proxy scenario), and we'd like to have a way to receive notification that a pods network namespace/iptables have been updated. This'll allow for updates to the IPTables rules when the pod is running, and will also allow us to not (re)introduce something like network-monitor.

christopherhein:
Thank you @egernst for filing this issue!

bergwolf:
@egernst We had a `netmon` to do similar work in the old 1.X times (https://github.com/kata-containers/runtime/blob/master/netmon/netmon.go#L671). While we certainly do not want to bring `netmon` back, we can use a goroutine thread from the shimv2 process to enter the pod netns on the host and do the same ip/route monitoring and copying work. Have you considered that approach?

egernst:
I had considered looking at net link (in the pod namespace, not host).  For our use case I didn’t need  to complicate things, add event listening. I think this can still be added on top of this implementation - it would just call directly from the shim’s go routine.

PR #4159, opened 2022-04-27
feature: add ability to interact with IPTables within the guest

egernst:
Introduce support for getting and setting IPTables within the guest. This pull request includes changes to get/set IPTables using iptables-save and iptables-restore within the guest.  

Note ** this requires that the guest has the ip(6)tables-save and ip(6)tables-restore binaries available in the guest rootfs **

To facilitate this, we augment the Agent API to include new requests for get and set IPTables, which takes a boolean to identify if the IPTables rules are ipv6 or not.

This API is made available in the sandbox interface. Today this can be exercised using the shim's shim-management HTTP endpoint, via ip6tables and iptables endpoint, which handles GET and PUT requests. This is likely the best interface for any host level service that would want to make modifications to the pods' IPTables.

This http API can also be exercised via kata-runtime. As an example, for a sandbox/container with ID "test-ctr"

 Get IPTables:
```
sudo kata-runtime iptables get --sandbox-id test
# Generated by iptables-save v1.4.21 on Thu Apr 28 21:24:29 2022
*security
:INPUT ACCEPT [80:6240]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [80:6240]
COMMIT
# Completed on Thu Apr 28 21:24:29 2022
# Generated by iptables-save v1.4.21 on Thu Apr 28 21:24:29 2022
*raw
...
```

To set IPTables which are set within the file `iptable_data_file`:

```
kata-runtime iptables set --sandbox-id test ./iptable_data_file
```

This functionality is introduced based on use case described here: https://github.com/kata-containers/kata-containers/issues/4080

```
We have a use case where we want to inject pod-specific ip table rules. In the case of runc this is as simple as creating the rules within the applicable network namespace.

For Kata this won't work, since the rules are managed managed at L3, while we use tcfilter at L2. Since we're doing a direct connection, no rules are exercised.

What I'd like to see is a mechanism to reflect the host side IPTables rules inside the guest, similar to what we do today with routes. In this way, we'd be able to effectively do "iptables-save" in the pod's network namespace on the host, and then have the agent do an iptables-restore inside the guest OS.

Further, there are use cases where the IPtables will be dynamic over time (Kube-proxy scenario), and we'd like to have a way to receive notification that a pods network namespace/iptables have been updated. This'll allow for updates to the IPTables rules when the pod is running, and will also allow us to not (re)introduce something like network-monitor.
```

Good reference on how we'd like to have an enhanced kube proxy use case feasible: https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/doc/vc-icdcs.pdf

Issue for adding documentation: https://github.com/kata-containers/kata-containers/issues/4173


mxpv (approved):

src/agent/src/rpc.rs
```diff
@@ -997,6 +1002,91 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+        &self,
+        ctx: &TtrpcContext,
+        req: GetIPTablesRequest,
+    ) -> ttrpc::Result<GetIPTablesResponse> {
+        trace_rpc_call!(ctx, "get_iptables", req);
+        is_allowed!(req);
+
+        info!(sl!(), "get_ip_tables: request received");
+
+        let mut cmd = "/sbin/iptables-save";
```
liubin, line 1015: Recommend:
```rust
let cmd = if req.is_ipv6 {
    "/sbin/ip6tables-save"
} else {
    "/sbin/iptables-save"
}
```

src/agent/src/rpc.rs
```diff
@@ -2749,4 +2840,123 @@ OtherField:other
...
+COMMIT
+
+"#;
+        let result = agent_service
+            .set_ip_tables(
+                &ctx,
+                SetIPTablesRequest {
+                    is_ipv6: false,
+                    data: valid_rules.as_bytes().to_vec(),
+                    //data: buf,
```
liubin, line 2937: Remove this line?

src/libs/protocols/protos/agent.proto
```diff
@@ -51,6 +51,8 @@ service AgentService {
 	rpc ListInterfaces(ListInterfacesRequest) returns(Interfaces);
 	rpc ListRoutes(ListRoutesRequest) returns (Routes);
 	rpc AddARPNeighbors(AddARPNeighborsRequest) returns (google.protobuf.Empty);
+	rpc GetIPTables(GetIPTablesRequest) returns (GetIPTablesResponse);
+	rpc SetIPTables(SetIPTablesRequest) returns (SetIPTablesResponse);
```
jodh-intel, line 55: Please can you update the [`agent-ctl` tool](https://github.com/kata-containers/kata-containers/blob/main/src/tools/agent-ctl/src/client.rs) for the new API calls so we have a way to test the agent directly.

jodh-intel (commented):
Thanks @egernst - a few comments...

src/agent/src/rpc.rs
```diff
@@ -90,6 +93,8 @@ const ERR_INVALID_BLOCK_SIZE: &str = "Invalid block size";
 const ERR_NO_LINUX_FIELD: &str = "Spec does not contain linux field";
 const ERR_NO_SANDBOX_PIDNS: &str = "Sandbox does not have sandbox_pidns";
 
+const IPTABLES_RESTORE_WAIT_SEC: &str = "5";
```
yibozhuang, line 96: How did we arrive at the 5s wait time? Could we either add a comment describing the choice of 5 or make this a configurable parameter?

src/agent/src/rpc.rs
```diff
@@ -997,6 +1002,91 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+        let mut file = NamedTempFile::new().unwrap();
+        let _ = match file.write(&*req.data) {
+            Ok(size) => size,
+            Err(e) => {
+                warn!(sl!(), "failed to write iptables to temp file");
+                return Err(ttrpc_error!(ttrpc::Code::INTERNAL, e));
+            }
+        };
+
+        let mut cmd = "/sbin/iptables-restore";
```
yibozhuang, line 1054: nit:
```rust
let cmd = if req.is_ipv6 {
    "/sbin/ip6tables-restore"
} else {
    "/sbin/iptables-restore"
}
```

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2195,6 +2203,26 @@ func (k *kataAgent) getAgentMetrics(ctx context.Context, req *grpc.GetMetricsReq
...
+func (k *kataAgent) getIPTables(ctx context.Context, isIPv6 bool) ([]byte, error) {
+	resp, err := k.sendReq(ctx, &grpc.GetIPTablesRequest{IsIpv6: isIPv6})
+	if err != nil {
+		return nil, err
+	}
+	return resp.(*grpc.GetIPTablesResponse).Data, nil
+}
+
+func (k *kataAgent) setIPTables(ctx context.Context, isIPv6 bool, data []byte) error {
+	_, err := k.sendReq(ctx, &grpc.SetIPTablesRequest{
```
yibozhuang, line 2215: nit, prefer
```go
	if _, err := k.sendReq(ctx, &grpc.SetIPTablesRequest{
		IsIpv6: isIPv6,
		Data:   data,
	}); err != nil {
		k.Logger().WithError(err).Errorf("setIPTables request to agent failed")
                return err
	}
	return nil
```

src/runtime/cmd/kata-runtime/kata-iptables.go
```diff
@@ -0,0 +1,118 @@
...
+	Usage: "get iptables from the Kata Containers guest",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for getting the iptables",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.BoolFlag{
+			Name:        "isIPv6",
```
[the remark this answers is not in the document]
egernst: I love kebab.

Will fix

src/agent/src/rpc.rs
mcastelino: Instead of using --wait, consider taking an explicit flock that you control. That way you know that nothing else is active (or you know that something else is active). relying on iptables's lock is opaque.

src/agent/src/rpc.rs
```diff
@@ -997,6 +1006,92 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+            }
+        };
+
+        let cmd = if req.is_ipv6 {
+            "/sbin/ip6tables-restore";
+        } else {
+            "/sbin/iptables-restore";
+        };
+
+        let output = match Command::new(cmd)
```
mcastelino, line 1064: You may want to first run the command with `-t`. i.e. test to ensure that it can apply. And if you get a failure return an error. That ensure that you catch any incompatibilities between host and guest versions of iptables.

It also ensure the guest remains consistent.

That also addresses the issue of ensuring that host and guest side iptables binaries are compatible.

liubin:
Beyond this case, can we add a common API to run commands in the guest OS?

For example:

```proto
rpc ExecCommand(ExecCommandRequest) returns (ExecCommandResponse);

message ExecCommandRequest {
        string command = 1;
        repeated string dns = 2;
        bytes stdin = 3;
        int32 timeout= 4;
}

message ExecCommandResponse{
        int32 exit_code= 1;
        bytes stdout = 2;
        bytes stderr = 3;
}
```

And from  command line we can just:

```bash
kata-runtime exec-commmand -- --sandbox abc123 /sbin/iptables-save
kata-runtime exec-commmand -- --sandbox abc123 /sbin/iptables-restore /tmp/abc.rules
```

For commands that need files as parameters, like `iptables-restore`, we can add `kata-runtime copy /host/path /guest/path` to copy files between host and guest.

With this common interface, we can avoid adding new APIs if we want to run other commands in the guest.

src/agent/src/rpc.rs
egernst: This is interesting. I think that a lock I own assumes the other potential writer would also be using this lock (not realistic).  I think we can utilize a sandbox lock on host side to ensure that we don't have multiple instances of get/set/get running at a time, but still rely on the xtable lock in the guest? WDYT?

src/agent/src/rpc.rs
egernst: From my testing, if iptables-restore will fail, the existing iptables are not modified.  In other words, I'll just return an error response over ttrpc if the -t failed, or if the actual attempt failed. End result should be the same?

src/agent/src/rpc.rs
mcastelino: in that case do not use the wait and fail right away.

src/agent/src/rpc.rs
mcastelino: ack. This was just a paranoid check which is lockless

src/libs/protocols/protos/agent.proto
egernst: Done!

egernst:
> Beyond this case, can we add a common API based on this PR to run commands in the guest OS?
> 

I'm pretty mixed on this -- while I know the primary threat model is to protect host from guest today, I am unsure if having a totally open option for executing a command is needed, pragmatic. I'm happy to hear what others think?

src/agent/src/rpc.rs
```diff
@@ -2749,4 +2886,127 @@ OtherField:other
             }
         }
     }
+
+    #[tokio::test]
+    async fn test_ip_tables() {
+        if !nix::unistd::Uid::effective().is_root() {
```
yibozhuang, line 2892: nit: can be replaced with `skip_if_not_root!();`

src/agent/src/rpc.rs
```diff
@@ -2749,4 +2886,127 @@ OtherField:other
...
+        let result = agent_service
+            .get_ip_tables(
+                &ctx,
+                GetIPTablesRequest {
+                    is_ipv6: true,
+                    ..Default::default()
+                },
+            )
+            .await;
+        assert!(result.is_ok(), "get ip tables should succeed");
```
yibozhuang, line 2935: nit: maybe something like `get ip6 ip tables should succeed`
just so its clear if the test fails that we can distinguish between ip and ip6

src/agent/src/rpc.rs
```diff
@@ -2749,4 +2886,127 @@ OtherField:other
...
+                    is_ipv6: false,
+                    data: garbage_rules.as_bytes().to_vec(),
+                    ..Default::default()
+                },
+            )
+            .await;
+        assert!(result.is_err(), "set iptables with garbage should fail");
+
+        // Verify setup of valid iptables:Setup  valid set of iptables:
+        let valid_rules = r#"
```
yibozhuang, line 2975: do we want to add a similar valid test case for ip6? right now there is no test for valid ip6 tables set/get

src/runtime/cmd/kata-runtime/kata-iptables.go
```diff
@@ -0,0 +1,118 @@
...
+	containerdshim "github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/katautils"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils/shimclient"
+	"github.com/urfave/cli"
+)
+
+var (
+	sandboxID    string
+	iptablesFile string
+	isIPv6       bool = false
```
yibozhuang, line 16:
```suggestion
	isIPv6 bool
```

src/runtime/cmd/kata-runtime/kata-iptables.go
```diff
@@ -0,0 +1,118 @@
...
+
+	containerdshim "github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/katautils"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils/shimclient"
+	"github.com/urfave/cli"
+)
+
+var (
+	sandboxID    string
+	iptablesFile string
```
yibozhuang, line 15: delete this, `iptablesFile` is not being used.

yibozhuang (approved):
mostly small nits, lgtm

src/runtime/cmd/kata-runtime/kata-iptables.go
egernst: Actually, we just use it for testing as part of the kata-runtime CLI

egernst:
@bergwolf -- in our use case, the IPTables rules are unique to the pod. In our case, there's an already existing agent that calls directly to the shim *instead* of writing an IPTables on the host (either in host's netns or the pods netns).

If we have use cases that require monitoring the pod netns for IPTables changes, then I think we can augment this exactly as you describe (go routine as part of shim, which then would make the corresponding call to virtcontainers -> kata-agent --> guest agent). Is it okay to add this if/when it is needed?

bergwolf:
@egernst Yes, we can add it later. I think it would help kata to work properly with kube-proxy.

src/agent/src/rpc.rs
```diff
@@ -994,6 +1001,140 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+    ) -> ttrpc::Result<GetIPTablesResponse> {
+        trace_rpc_call!(ctx, "get_iptables", req);
+        is_allowed!(req);
+
+        info!(sl!(), "get_ip_tables: request received");
+
+        let cmd = if req.is_ipv6 {
+            "/sbin/ip6tables-save"
+        } else {
+            "/sbin/iptables-save"
```
bergwolf, line 1017: Please constify the two binaries paths.

src/agent/src/rpc.rs
```diff
@@ -994,6 +1001,140 @@ impl protocols::agent_ttrpc::AgentService for AgentService {
...
+    ) -> ttrpc::Result<SetIPTablesResponse> {
+        trace_rpc_call!(ctx, "set_iptables", req);
+        is_allowed!(req);
+
+        info!(sl!(), "set_ip_tables request received");
+
+        let cmd = if req.is_ipv6 {
+            "/sbin/ip6tables-restore"
+        } else {
+            "/sbin/iptables-restore"
```
bergwolf, line 1046: Same here.

src/runtime/cmd/kata-runtime/kata-iptables.go
```diff
@@ -0,0 +1,122 @@
...
+	Usage: "set iptables in a specifc Kata Containers guest based on file",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for setting the iptables",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.BoolFlag{
+			Name:        "isIPv6",
```
bergwolf, line 83: How about just naming it `v6`. `isIPv6` looks like a function parameter.

bergwolf (commented):
Generally LGTM! Thanks @egernst ! A few minor comments inline. Also, we have dropped `iptables` from the osbuilder rootfs package lists. So we need to add it back in order to allow kata-agent to work properly with this. Are you planning to do it in the same PR or in a following up one?

egernst:
@bergwolf -- I want to add the pkg as a separate PR.  See https://github.com/kata-containers/kata-containers/pull/4357

bergwolf (approved):
lgtm! Thanks a lot @egernst !

Issue #4160, opened 2022-04-27
Add RKE2 support to kata-deploy


fidencio:
kata-deploy currently supports plain kubernetes (https://kubernetes.io/), k3s (https://k3s.io/), but would simply bail when trying to install on a RKE2 (https://docs.rke2.io/) cluster.

RKE2 is the Rancher's Next Generation Kubernetes Distribution, and it uses `k3s` behind the scenes.  However, it does it with some modifications, such as:
* The location where the containerd configuration is stored:
  * k3s: /var/lib/rancher/k3s/agent/etc/containerd/
  * rke2: /var/lib/rancher/rke2/agent/etc/containerd/
* The name of the systemd services used:
  * k3s: k3s.service or k3s-agent.service
  * rke2: rke2-server.service or rke2-agent.service

Knowing this, simple adjustments can be made to kata-deploy / kata-cleanup overlays and in the kata-deploy.sh script itself, so we can successfully install and run Kata Containers on RKE2.

PR #4161, opened 2022-04-27
kata-deploy: Add support to RKE2

fidencio:
"RKE2 - Rancher's Next Generation Kuberentes Distribution" can easily be
supported by kata-deploy with some simple adjustments to what we've been
relying on for "k3s".

The main differences between k3s and RKE2 are, basically:
1. The location where the containerd configuration is stored
   - k3s: /var/lib/rancher/k3s/agent/etc/containerd/
   - rke2: /var/lib/rancher/rke2/agent/etc/containerd/
2. The name of the systemd services used:
   - k3s: k3s.service or k3s-agent.service
   - rke2: rke2-server.service or rke2-agent.service

Knowing this, let's add a new overlay for RKE2, adapt the kata-deploy
and the kata-cleanup scripts, and that's it.

Fixes: #4160

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -39,7 +39,11 @@ function get_container_runtime() {
                 die "invalid node name"
 	fi
 	if echo "$runtime" | grep -qE 'containerd.*-k3s'; then
-		if systemctl is-active --quiet k3s-agent; then
+		if systemctl is-active --quiet rke2-agent; then
+			echo "rke2-agent"
+		elif systemctl is-active --quiet rke2-server; then
+			echo "rke2-server"
+		elif systemctl is-active --quiet k3s-agent; then
 			echo "k3s-agent"
 		else
 			echo "k3s"
@@ -62,7 +66,7 @@ function configure_cri_runtime() {
 	crio)
 		configure_crio
 		;;
-	containerd | k3s | k3s-agent)
+	containerd | k3s | k3s-agent | rke2-agent | rke2-server)
 		configure_containerd
 		;;
 	esac
@@ -228,7 +232,7 @@ function cleanup_cri_runtime() {
 	crio)
 		cleanup_crio
 		;;
-	containerd | k3s | k3s-agent)
+	containerd | k3s | k3s-agent | rke2-agent | rke2-server)
 		cleanup_containerd
 		;;
 	esac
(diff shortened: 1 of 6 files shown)


fidencio:
/test_kata_deploy

Let's we ensure we didn't break the only case we have a test for.

fidencio:
By the way, I've tested this locally with an RKE2 single "server" node.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -39,7 +39,11 @@ function get_container_runtime() {
                 die "invalid node name"
 	fi
 	if echo "$runtime" | grep -qE 'containerd.*-k3s'; then
-		if systemctl is-active --quiet k3s-agent; then
+		if systemctl is-active --quiet rke2-agent; then
+			echo "rke2-agent"
+		elif systemctl is-active --quiet rke2-server; then
```
jodh-intel, line 44: Is there any possibility of having both agent and server on the same node (for testing?)

jodh-intel (commented):

fidencio:
By the way, test kata-deploy passed: https://github.com/kata-containers/kata-containers/actions/runs/2234039278

fidencio:
I, personally, would like to see this backported as it does **not** touch any kata-containers features and increases the places where the project can be deployed.

However, I'd understand if the other members of the A/C don't think it's a good candidate for a backport.

egernst:
I’m usually more conservative on backports. This feels like more of a feature’ish enhancement to me. If someone wants to use in RKE2, they can use the 2.5 alpha release that this’ll be available in right?  This is enabling new user/usage more than fixing an existing?

amshinde (approved):

amshinde:
Agree with @egernst. This looks like a feature add and a good candidate for the the next release.

fidencio:
> I’m usually more conservative on backports. This feels like more of a feature’ish enhancement to me. If someone wants to use in RKE2, they can use the 2.5 alpha release that this’ll be available in right? This is enabling new user/usage more than fixing an existing?

Partially.  Right now when users try to install on RKE2 the k3s path is taken, as that's what's detected by kata-deploy, and the installation bails.

But yeah, I understand the conservative approach, I just think it's sad to say "you want to use stable on RKE2? come back in 2 months ...", mainly for a fix that does **not** touch any of the functional code of the project.

jodh-intel (approved):
Thanks @fidencio.

lgtm

Issue #4164, opened 2022-04-28
kata-agent-ctl does not connect to agent running in same env


cmaf:
@jodh-intel, possibly @liubin 

# Description of problem
The kata-agent-ctl tool does not connect to an agent running in the same environment using QEMU (per these instructions: https://github.com/kata-containers/kata-containers/tree/main/src/tools/agent-ctl#run-the-tool-and-the-agent-in-the-same-environment). This breaks the `local` option for the agent shutdown test. 

# Expected result
Several lines of output that look somewhat like this with no final error message:
```
{"msg":"Command DestroySandbox (1 of 1) returned (Ok(()), true)","level":"INFO","ts":"2022-04-27T23:17:33.225401455Z","pid":"155954","subsystem":"rpc","version":"0.1.0","source":"kata-agent-ctl","name":"kata-agent-ctl"}
```

# Actual result
Output ending with this:
```
ERROR: Failed to connect to Unix Domain abstract socket
```

# Further information
I'm using the agent shutdown test which creates an OCI bundle, uses the appropriate flag for it in the kata-agent-ctl tool, and configures debug settings in the kata configuration file. 

I have tried changing the server address from abstract as well as the server address when initially running the agent. I also tried the --hybrid-vsock flag (not relevant for QEMU). 

I haven't seen this problem running a kata container (which is used in the agent shutdown test `vm` option). 

# Kata-collect data
<details>
<summary>Show <tt>kata-collect-data.sh</tt> details</summary>
<p>

# Meta details

Running `kata-collect-data.sh` version `2.5.0-alpha0 (commit 83979ece18b95bbf0f53517a44b8bd2e95f29637)` at `2022-04-27.23:31:07.048381536+0000`.

---

<details>
<summary>Runtime</summary>
<p>

Runtime is `/usr/local/bin/kata-runtime`.

# `kata-env`

<details>
<summary><tt>/usr/local/bin/kata-runtime kata-env</tt></summary>
<p>

```toml
[Kernel]
  Path = "/usr/share/kata-containers/vmlinux-5.15.26-90"
  Parameters = "systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket scsi_mod.scan=none"

[Meta]
  Version = "1.0.26"

[Image]
  Path = "/usr/share/kata-containers/kata-containers.img"

[Initrd]
  Path = ""

[Hypervisor]
  MachineType = "q35"
[...]

PR #4166, opened 2022-04-28
agent-ctl: Fix abstract socket connections

jodh-intel:
Unbreak the `agent-ctl` tool connecting to the agent with a Unix domain
socket.

It appears that [1] changed the behaviour of connecting to the agent
using a local Unix socket (which is not used by Kata under normal
operation).

The change can be seen by reverting to commit
72b8144b569f1d0fef3001642f244bf79b17ddc6 (the one before [1]) and
running the agent manually as:

```bash
$ sudo KATA_AGENT_SERVER_ADDR=unix:///tmp/foo.socket target/x86_64-unknown-linux-musl/release/kata-agent
```

Before [1], in another terminal we see this:

```bash
$ sudo lsof -U 2>/dev/null |grep foo|awk '{print $9}'
@/tmp/foo.socket@
```

But now, we see the following:

```bash
$ sudo lsof -U 2>/dev/null |grep foo|awk '{print $9}'
@/tmp/foo.socket
```

Note the last byte which represents a nul (`\0`) value.

The `agent-ctl` tool used to add that trailing nul but now it seems to not
be needed, so this change removes it, restoring functionality. No
external changes are necessary so the `agent-ctl` tool can connect to
the agent as below like this:

```bash
$ cargo run -- -l debug connect --server-address "unix://@/tmp/foo.socket" --bundle-dir "$bundle_dir" -c Check -c GetGuestDetails
```

[1] - https://github.com/kata-containers/kata-containers/issues/3124

Fixes: #4164.

### DIFF
--- a/src/tools/agent-ctl/README.md
+++ b/src/tools/agent-ctl/README.md
@@ -211,9 +211,11 @@ $ sudo install -o root -g root -m 0755 ~/.cargo/bin/kata-agent-ctl /usr/local/bi
 
 > **Warnings:**
 >
-> - This method is **only** for testing and development!
+> - These methods are **only** for testing and development!
 > - Only continue if you are using a non-critical system
 >   (such as a freshly installed VM environment).
+
+#### Use a Unix abstract domain socket
 
 1. Start the agent, specifying a local socket for it to communicate on:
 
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @jodh-intel!

fidencio:
@jodh-intel, do you think this one deserves a backport?

jodh-intel:
Updated to add doc details of how to use a local VSOCK socket as that's related to this indirectly (local VSOCK is an alternative to a Unix socket).

jodh-intel:
@fidencio - It's not a tool we provide to users, so I'm uncertain tbh.

cmaf (commented):
@jodh-intel The VSOCK loopback socket workaround didn't work for me, but I'm going to test the agent-ctl change locally

cmaf (approved):
thanks @jodh-intel, the change to client.rs fixed it

Issue #4167, opened 2022-04-28
Chronyd.service fails on boot causing clock drift over time


Champ-Goblem:
In kata version 2.4.0 and 2.3.3 we have been experiencing some clock drift over time, for example a new pod started in kata has the correct time with respect to NTP standard, but over time this slowly drifts out, with some 6 day old pods being around 5-6 seconds out of sync. 

We found https://github.com/kata-containers/runtime/issues/1279 issue from the old repository that speaks about using kvm_ptp along with chronyd to keep the time in sync. On the current kata versions I can confirm that `/dev/ptp0` exists so the kvm_ptp module is not disabled, but checking dmesg showed that chronyd.service was failing on boot with status `266/NAMESPACE` and no logs in Journald. 

After some investigation and manual editing of the chrony service, the problem seems to stem from the line `ReadWritePaths=/run /var/lib/chrony -/var/log`, editing the middle path to `/var/run` seems to allow chronyd.service to start, although I havent tested if this resolves the drift issue yet. I straced the chronyd -d process (output below) and I can see that it really only accesses /var/run/chrony but setting the middle ReadWritePaths to /var/run/chrony breaks it again. My resolution may or may not be optimal so it would be nice to receive some feedback on the situation.

### `chronyd -d` strace:
<details>
<summary>strace output</summary>

```
strace chronyd -d
execve("/usr/bin/chronyd", ["chronyd", "-d"], 0x7ffcb324e368 /* 35 vars */) = 0
brk(NULL)                               = 0x56154f85e000
arch_prctl(0x3001 /* ARCH_??? */, 0x7fffc9ba24a0) = -1 EINVAL (Invalid argument)
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f4054a30000
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/var/cache/ldconfig/ld.so.cache", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/glibc-hwcaps/x86-64-v4/libm.so.6", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
[...]

PR #4168, opened 2022-04-28
rootfs: Fix chronyd.service failing on boot

Champ-Goblem:
In at least kata versions 2.3.3 and 2.4.0 it was noticed that the guest operating system's clock would drift out of sync slowly over time whilst the pod was running.

This had previously been raised and fixed in the old reposity via [1] . In essence kvm_ptp and chrony were paired together in order to keep the system clock up to date with the host.

In the recent versions of kata metioned above, the chronyd.service fails upon boot with status `266/NAMESPACE` which seems to be due to the fact that the `/var/lib/chrony` directory no longer exists.

This change sets the `/var/lib/chrony` directory for the `ReadWritePaths` to be ignored when the directory does not exist, as per [2].

[1] https://github.com/kata-containers/runtime/issues/1279
[2] https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ReadWritePaths=

Fixes: https://github.com/kata-containers/kata-containers/issues/4167

Signed-off-by: Cameron McDermott <cameron_mcdermott@yahoo.co.uk>

### DIFF
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -540,8 +540,13 @@ EOF
 
 	if [ -f "$chrony_systemd_service" ]; then
 		# Remove user option, user could not exist in the rootfs
+		# Set the /var/lib/chrony for ReadWritePaths to be ignored if
+		# its nonexistent, this broke the service on boot previously
+		# due to the directory not being present "(code=exited, status=226/NAMESPACE)"
 		sed -i -e 's/^\(ExecStart=.*\)-u [[:alnum:]]*/\1/g' \
-		       -e '/^\[Unit\]/a ConditionPathExists=\/dev\/ptp0' ${chrony_systemd_service}
+		       -e '/^\[Unit\]/a ConditionPathExists=\/dev\/ptp0' \
+		       -e 's/^ReadWritePaths=\(.\+\) \/var\/lib\/chrony \(.\+\)$/ReadWritePaths=\1 -\/var\/lib\/chrony \2/m' \
+		       ${chrony_systemd_service}
 	fi
 
 	AGENT_DIR="${ROOTFS_DIR}/usr/bin"


tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -540,8 +540,12 @@ EOF
 
 	if [ -f "$chrony_systemd_service" ]; then
 		# Remove user option, user could not exist in the rootfs
+		# Set the /var/lib/chrony for ReadWritePaths to be ignored if
+		# its nonexistent, this broke the service on boot previously
+		# due to the directory not being present "(code=exited, status=226/NAMESPACE)"
 		sed -i -e 's/^\(ExecStart=.*\)-u [[:alnum:]]*/\1/g' \
-		       -e '/^\[Unit\]/a ConditionPathExists=\/dev\/ptp0' ${chrony_systemd_service}
+		       -e '/^\[Unit\]/a ConditionPathExists=\/dev\/ptp0' \
+					 -e '/^ReadWritePaths=(.+) \/var\/lib\/chrony (.+)$/ReadWritePaths=\1 -\/var\/lib\/chrony \2/' ${chrony_systemd_service}
```
fidencio, line 548: It seems the indentation is not aligned (two tabs + some spaces vs only tabs on your patch).

Apart from this, this looks good to me.
If the issue is happening with the default clearlinux images, I wonder how we could report this back inside Intel.

bryteise: I think because the Clear Linux base image doesn't have `/var` writable (correct me if I'm wrong) this configuration is needed. It isn't something we'd try to account for by default in the distro though so making the change here seems best.

amshinde: This used to work in the past, so I am wondering what exactly has caused this breakage. Has there been a change on the Clearlinux side or the chrony service?

Champ-Goblem: Looks like the indentation is fixed now

fidencio:
Still on this, I'm checking a container that I have here running with the default image, and what I am facing is:
```
root@rate-limitting:/home/ubuntu# kata-runtime exec c43d86ba7155832e02f920404244bca3fcbc356024d6b7001a6c04c35b4ad0b7
root@localhost:/# systemctl status chrony
● chrony.service - chrony, an NTP client/server
     Loaded: loaded (/lib/systemd/system/chrony.service; enabled; vendor preset:
 enabled)
     Active: inactive (dead)
  Condition: start condition failed at Thu 2022-04-28 09:42:28 
UTC; 23h ago
             └─ ConditionPathExists=/dev/ptp0 was not met
       Docs: man:chronyd(8)
             man:chronyc(1)
             man:chrony.conf(5)
```

So, there may be more into this.

@amshinde, would you have some cycles to figure out where / when we regressed?

Last but not least, I will open an issue in the `tests` repo, as this is something we should be able to test and avoid regressions in the future.

fidencio:
https://github.com/kata-containers/tests/issues/4746 is the issue for adding a test for this case.

Champ-Goblem:
I did not check whether the issue exists in Centos at first, but I have just built the image from the 2.4.0 tag and I can see that it seems to be fine in this case. A quick check of the service manifest for chronyd shows that the line `ReadWritePaths` does not exist.
Mounting the image again and updating the service to contain the same line as seen in the Clear Linux version causes the service to fail again on boot, presumably because the `/var/lib/chrony` directory does not exist again and /var is read only. I can temporarily fix the service by remounting the rootfs as read-write and creating the /var/lib/chrony directory manually when the ReadWritePaths line is set.

Champ-Goblem:
Looking at the different versions of kata:
2.1.0 -> Chronyd working and no ReadWritePaths present
2.2.0 -> Same as before
2.3.0 -> Same as before
2.3.2 -> Same as before
2.3.3 -> Failing here as above
I think also aligns with the fact that I have not noticed the drift on our clusters till very recently

Hope this information helps

haslersn:
Hi! What's the status of this fix? The time drift recently caused issues with our OpenID connect provider.

fidencio:
@amshinde, @jodh-intel, can we go ahead and have this one merged as it is?
I'm afraid that expecting for someone to have enough time to find the possible root cause of the issue may take way too long, and this is something that should be fixed ASAP.

fidencio (approved):
lgtm, thanks @Champ-Goblem!

jodh-intel (approved):
Thanks @Champ-Goblem.

lgtm

haslersn:
Will this fix be included in 2.5.0?

Issue #4169, opened 2022-04-28
Firmware support for SEV: AmdSev build of OVMF


Alex-Carter01:
As part of #3982, SEV will need this specific firmware build.

PR #4170, opened 2022-04-28
Add support AmdSev build of OVMF

Alex-Carter01:
Fixes: #4169
Signed-off-by: Alex Carter <alex.carter@ibm.com>

Introduces a build script for AmdSev build of OVMF, which will be needed for SEV measured direct boot.

### DIFF
--- a/tools/packaging/static-build/ovmf/Dockerfile
+++ b/tools/packaging/static-build/ovmf/Dockerfile
@@ -1,0 +1,21 @@
+# Copyright (c) 2022 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ubuntu:20.04
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+        g++ \
+        gcc \
+        git \
+        iasl  \
+        make \
+        nasm \
+        python \
+        python3 \
+        uuid-dev && \
+    apt-get clean && rm -rf /var/lib/lists/
--- a/tools/packaging/static-build/ovmf/build-ovmf.sh
+++ b/tools/packaging/static-build/ovmf/build-ovmf.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${script_dir}/../../scripts/lib.sh"
+
+# disabling set -u because scripts attempt to expand undefined variables
+set +u
+ovmf_repo="${ovmf_repo:-}"
+ovmf_dir="edk2"
+ovmf_version="${ovmf_version:-}"
+ovmf_package="${ovmf_package:-}"
+package_output_dir="${package_output_dir:-}"
+DESTDIR=${DESTDIR:-${PWD}}
+PREFIX="${PREFIX:-/opt/kata}"
+architecture="${architecture:-X64}"
+toolchain="${toolchain:-GCC5}"
+build_target="${build_target:-RELEASE}"
+
+[ -n "$ovmf_repo" ] || die "failed to get ovmf repo"
+[ -n "$ovmf_version" ] || die "failed to get ovmf version or commit"
+[ -n "$ovmf_package" ] || die "failed to get ovmf package or commit"
+[ -n "$package_output_dir" ] || die "failed to get ovmf package or commit"
+
+info "Build ${ovmf_repo} version: ${ovmf_version}"
+
+build_root=$(mktemp -d)
+pushd $build_root
+git clone "${ovmf_repo}"
+cd "${ovmf_dir}"
+git checkout "${ovmf_version}"
+git submodule init
+git submodule update
+
+info "Using BaseTools make target"
+make -C BaseTools/
+
+info "Calling edksetup script"
+source edksetup.sh
+
+info "Building ovmf"
+build -b "${build_target}" -t "${toolchain}" -a "${architecture}" -p "${ovmf_package}"
+
+info "Done Building"
+
+build_path="Build/${package_output_dir}/${build_target}_${toolchain}/FV/OVMF.fd"
+stat "${build_path}"
+
+#need to leave tmp dir
+popd
+
+info "Install fd to destdir"
+mkdir -p "$DESTDIR/$PREFIX/share/ovmf"
+cp $build_root/$ovmf_dir/"${build_path}" "$DESTDIR/$PREFIX/share/ovmf"
--- a/tools/packaging/static-build/ovmf/build-ovmf.sh
+++ b/tools/packaging/static-build/ovmf/build-ovmf.sh
@@ -13,6 +13,7 @@ source "${script_dir}/../../scripts/lib.sh"
 
 # disabling set -u because scripts attempt to expand undefined variables
 set +u
+ovmf_build="${ovmf_build:-x86_64}"
 ovmf_repo="${ovmf_repo:-}"
 ovmf_dir="edk2"
 ovmf_version="${ovmf_version:-}"
@@ -44,6 +45,12 @@ make -C BaseTools/
 
 info "Calling edksetup script"
 source edksetup.sh
+
+if [ "${ovmf_build}" == "sev" ]; then
+       info "Creating dummy grub file"
+       #required for building AmdSev package without grub
+       touch OvmfPkg/AmdSev/Grub/grub.efi
+fi
 
 info "Building ovmf"
 build -b "${build_target}" -t "${toolchain}" -a "${architecture}" -p "${ovmf_package}"
--- a/tools/packaging/static-build/ovmf/build.sh
+++ b/tools/packaging/static-build/ovmf/build.sh
@@ -34,6 +34,10 @@ if [ "${ovmf_build}" == "x86_64" ]; then
        [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.x86_64.version" "${kata_version}")
        [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.x86_64.package" "${kata_version}")
        [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.x86_64.package_output_dir" "${kata_version}")
+elif [ "${ovmf_build}" == "sev" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.sev.version" "${kata_version}")
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.sev.package" "${kata_version}")
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.sev.package_output_dir" "${kata_version}")
 fi
 
 [ -n "$ovmf_version" ] || die "failed to get ovmf version or commit"
@@ -45,6 +49,7 @@ sudo docker build -t "${container_image}" "${script_dir}"
 sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
 	-w "${PWD}" \
 	--env DESTDIR="${DESTDIR}" --env PREFIX="${PREFIX}" \
+	--env ovmf_build="${ovmf_build}" \
 	--env ovmf_repo="${ovmf_repo}" \
 	--env ovmf_version="${ovmf_version}" \
 	--env ovmf_package="${ovmf_package}" \
--- a/tools/packaging/static-build/ovmf/build.sh
+++ b/tools/packaging/static-build/ovmf/build.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2022 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+readonly repo_root_dir="$(cd "${script_dir}/../../../.." && pwd)"
+readonly ovmf_builder="${script_dir}/build-ovmf.sh"
+
+source "${script_dir}/../../scripts/lib.sh"
+
+DESTDIR=${DESTDIR:-${PWD}}
+PREFIX=${PREFIX:-/opt/kata}
+container_image="kata-ovmf-builder"
+ovmf_build="${ovmf_build:-x86_64}"
+kata_version="${kata_version:-}"
+ovmf_repo="${ovmf_repo:-}"
+ovmf_version="${ovmf_version:-}"
+ovmf_package="${ovmf_package:-}"
+package_output_dir="${package_output_dir:-}"
+
+if [ -z "$ovmf_repo" ]; then
+       ovmf_repo=$(get_from_kata_deps "externals.ovmf.url" "${kata_version}")
+fi
+
+[ -n "$ovmf_repo" ] || die "failed to get ovmf repo"
+
+if [ "${ovmf_build}" == "x86_64" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.x86_64.version" "${kata_version}")
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.x86_64.package" "${kata_version}")
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.x86_64.package_output_dir" "${kata_version}")
+fi
+
+[ -n "$ovmf_version" ] || die "failed to get ovmf version or commit"
+[ -n "$ovmf_package" ] || die "failed to get ovmf package or commit"
+[ -n "$package_output_dir" ] || die "failed to get ovmf package or commit"
+
+sudo docker build -t "${container_image}" "${script_dir}"
+
+sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
+	-w "${PWD}" \
+	--env DESTDIR="${DESTDIR}" --env PREFIX="${PREFIX}" \
+	--env ovmf_repo="${ovmf_repo}" \
+	--env ovmf_version="${ovmf_version}" \
+	--env ovmf_package="${ovmf_package}" \
+	--env package_output_dir="${package_output_dir}" \
+	"${container_image}" \
+	bash -c "${ovmf_builder}"
--- a/versions.yaml
+++ b/versions.yaml
@@ -251,7 +251,12 @@ externals:
       version: "edk2-stable202202"
       package: "OvmfPkg/OvmfPkgX64.dsc"
       package_output_dir: "OvmfX64"
-      
+    sev:
+      description: "AmdSev build needed for SEV measured direct boot."
+      version: "edk2-stable202202"
+      package: "OvmfPkg/AmdSev/AmdSevX64.dsc"
+      package_output_dir: "AmdSev"
+
   virtiofsd:
     description: "vhost-user virtio-fs device backend written in Rust"
     url: "https://gitlab.com/virtio-fs/virtiofsd"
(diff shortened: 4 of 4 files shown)


versions.yaml
```diff
@@ -243,6 +243,12 @@ externals:
     url: "https://github.com/containerd/nydus-snapshotter"
     version: "v0.1.0"
 
+  ovmf:
```
fidencio, line 246: I'd like to ask you to add an subsection called sev under the newly introduced OVMF.  QEMU with TDX also uses OVMF and we can leverage this very same entry later on.

An example of how can be done is: https://github.com/kata-containers/kata-containers/blob/3f069c7acbfa3bd789255ca72f5b56414f4123e3/versions.yaml#L153-L160

Alex-Carter01: Sounds good, I have added a new SEV subsection. Glad to know this can be used for tdx in the future.

Alex-Carter01:
> Also, I'd suggest the commit message to be changed to something like:
> 
> ```
> static-build: Add AmdSev build of OMVF
> 
> Introduce a build script for AmdSev build of OMVF, which is needed for
> SEV measured direct boot.
> 
> Fixes: #4168
> Signed-off-by: Alex Carter <alex.carter@ibm.com>
> ```

Updated the commit message to be more informative. Thank you for all the feedback!

fidencio:
@Alex-Carter01, the commit check is failing and the author's email of the committer and the "signed-off-by" do not match.
Please, update that.

@amshinde, mind to also take a look at this as QEMU also uses OVMF for TDX?

Alex-Carter01:
I have added my IBM email to this account, and amended the commit with that email is the author so it matches the signature, I hope that resolves things.

fidencio:
LGTM, but I'd like to hear from folks working on the QEMU side at Intel, so I'll wait for @amshinde to give it a review as well.

Alex-Carter01:
resolved small new conflict in versions file

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,64 @@
...
+set +u
+source edksetup.sh
+set -u
+
+info "Creating dummy grub file"
+#required for building AmdSev package without grub
+touch OvmfPkg/AmdSev/Grub/grub.efi
+
+info "Building ovmf"
+build -t GCC5 -a X64 -p OvmfPkg/AmdSev/AmdSevX64.dsc
```
dubek, line 56: Suggestion: extract the platform file name `OvmfPkg/AmdSev/AmdSevX64.dsc` to a configuration entry as well, and then `build-ovmf.sh` can be used for other OVMF targets (TDX?). However, I see that further down we look for `Build/AmdSev/DEBUG_GCC5/FV/OVMF.fd`, which is AmdSev-specific, so maybe some more config entries are needed? Not sure how generic this should be.

Alex-Carter01: Sure, I can look into what I can do to make this more generic now that I know it will likely also be used for TDX.

Alex-Carter01: Sure, I can look into what I can do to make this more generic now that I know it will likely also be used for TDX.

amshinde (approved):
lgtm

Alex-Carter01:
Made some changes based on the suggestions from @dubek to make this script more generic for other builds of OVMF. 

The way I would imagine say TDX using this script would be to add another entry in the versions file, now with the desired package, and set a environment variable "ovmf_build" to 'tdx'. I have currently set the default value for this to sev. Then there would need to be a much simpler PR against this to add cases for tdx to change where to pull versions information from, but the rest of the build process should then be flexible.

Thank you @amshinde for the review! Letting you know this push to make it more generic is relatively large compared to past tweaks.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+
+#todo: check that this is even valid shell
+if [ "${ovmf_build}" == "sev" ]; then
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.sev.package" "${kata_version}")
+fi
+[ -n "$ovmf_package" ] || die "failed to get ovmf package or commit"
+info "Building ovmf"
+build -b ${build_target} -t ${toolchain} -a ${architecture} -p ${ovmf_package}
```
larrydewey, line 70: Same as on line 42, quoting to prevent interpreter problems from occurring.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,85 @@
...
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.sev.package_output_dir" "${kata_version}")
+fi
+[ -n "$package_output_dir" ] || die "failed to get ovmf package or commit"
+
+build_path="Build/${package_output_dir}/${build_target}_${toolchain}/FV/OVMF.fd"
+stat ${build_path} 
+
+info "Install fd to destdir"
+mkdir -p ../../destdir/opt/kata/share/ovmf
+cp ${build_path} ../../destdir/opt/kata/share/ovmf
```
larrydewey, line 85: Same as 42.

dubek (approved):

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,85 @@
...
+
+set +u
+ovmf_build="${ovmf_build:-sev}"
+architecture="X64"
+toolchain="GCC5"
+build_target="RELEASE"
+
+if [ -z "$ovmf_repo" ]; then
+       info "Get ovmf information from runtime versions.yaml"
+       ovmf_url=$(get_from_kata_deps "externals.ovmf.url" "${kata_version}")
```
wainersm, line 28: `git clone https://github.com/tianocore/edk2` is enough to clone the repo, you don't need to append `.git` in the url. So this block can be simplified.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,85 @@
...
+[ -n "$ovmf_repo" ] || die "failed to get ovmf repo"
+
+if [ "${ovmf_build}" == "sev" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.sev.version" "${kata_version}")
+fi
+[ -n "$ovmf_version" ] || die "failed to get ovmf version or commit"
+
+info "Build ${ovmf_repo} version: ${ovmf_version}"
+
+[ -d "${ovmf_dir}" ] || git clone "${ovmf_repo}"
```
wainersm, line 42: Even if `${ovmf_dir}` exists it is not guarantee to be the same sources as from `${ovmf_repo}`. My suggestion is to delete and then clone. Actually, I think it would be better to use a temporary directory for the build (and properly clean up it afterwards), for example:

```
....
build_root=$(mktemp -d)
cleanup() {
    if [ "${DEBUG:-}" == true ]; then
       info "Not deleted build root directory: ${build_root}"
       return
    fi
    rm -rf "${build_root}"
}
....
trap cleanup EXIT
```

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,85 @@
...
+if [ "${ovmf_build}" == "sev" ]; then
+       info "Creating dummy grub file"
+       #required for building AmdSev package without grub
+       touch OvmfPkg/AmdSev/Grub/grub.efi
+fi
+
+
+#todo: check that this is even valid shell
+if [ "${ovmf_build}" == "sev" ]; then
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.sev.package" "${kata_version}")
```
wainersm, line 66: Move this (and the other initializations below) up at the begin of the script so that if some information for the build is missing then it exists right away (fail fast).

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,85 @@
...
+cd "${ovmf_dir}"
+git checkout "${ovmf_version}"
+git submodule init
+git submodule update
+
+info "Using BaseTools make target"
+make -C BaseTools/
+set -u
+
+info "Calling edksetup scipt"
```
wainersm, line 52: s/scipt/script/

wainersm:
Hi @Alex-Carter01 , thanks for working on this. I finished my review, left just a couple of comments.

Alex-Carter01:
This change accounts for 3/5 of @wainersm 's suggestions. I will be coming back to add a fix to accommodate DESTDIR/PREFIX values and a temporary build directory

wainersm:
> This change accounts for 3/5 of @wainersm 's suggestions. I will be coming back to add a fix to accommodate DESTDIR/PREFIX values and a temporary build directory

Great. thanks @Alex-Carter01

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+
+source "${script_dir}/../../scripts/lib.sh"
+
+# disabling set -u because scripts attempt to expand undefined variables
+set +u
+ovmf_repo="${ovmf_repo:-}"
+ovmf_dir="edk2"
+ovmf_version="${ovmf_version:-}"
+kata_version="${kata_version:-}"
+DESTDIR="${DESTDIR:-../../destdir}"
```
fidencio, line 21: The other scripts are using `${PWD}` as the destdir, any specific reason to use `../../destir`?

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+PREFIX="${PREFIX:-/opt/kata}"
+ovmf_build="${ovmf_build:-sev}"
```
fidencio, line 23: The default build should be the "neutral" one, the `x64` one.  Then in the kata-deploy scripts we pass whether we'll build OVMF for SEV or TDVF for TDX.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+
+if [ "${ovmf_build}" == "sev" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.sev.version" "${kata_version}")
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.sev.package" "${kata_version}")
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.sev.package_output_dir" "${kata_version}")
+elif [ "${ovmf_build}" == "x64" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.x64.version" "${kata_version}")
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.x64.package" "${kata_version}")
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.x64.package_output_dir" "${kata_version}")
+
```
fidencio, line 42: Extra empty line.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+fi
```
fidencio, line 43: I'd prefer to have those being introduced in two different commits.
The first one adding support for building OVMF, and a second one adding support for building OVMF for SEV.  That would be cleaner.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+[ -n "$ovmf_repo" ] || die "failed to get ovmf repo"
+
+if [ "${ovmf_build}" == "sev" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.sev.version" "${kata_version}")
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.sev.package" "${kata_version}")
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.sev.package_output_dir" "${kata_version}")
+elif [ "${ovmf_build}" == "x64" ]; then
+       [ -n "$ovmf_version" ] || ovmf_version=$(get_from_kata_deps "externals.ovmf.x64.version" "${kata_version}")
+       [ -n "$ovmf_package" ] || ovmf_package=$(get_from_kata_deps "externals.ovmf.x64.package" "${kata_version}")
+       [ -n "$package_output_dir" ] || package_output_dir=$(get_from_kata_deps "externals.ovmf.x64.package_output_dir" "${kata_version}")
```
fidencio, line 41: I know this is how it's called on OVMF side, but for consistency sake I'd prefer having `x64` renamed to `x86_64` both here and in the versions file.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+
+cleanup() {
+    if [ "${DEBUG:-}" == true ]; then
+       info "Not deleted build root directory: ${build_root}"
+       return
+    fi
+    rm -rf "${build_root}"
+}
+
+trap cleanup EXIT
```
fidencio, line 94: I wouldn't do the cleanup.  Leave the directory there, always, as already done for different build scripts.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+       info "Creating dummy grub file"
+       #required for building AmdSev package without grub
+       touch OvmfPkg/AmdSev/Grub/grub.efi
+fi
+
+info "Building ovmf"
+build -b "${build_target}" -t "${toolchain}" -a "${architecture}" -p "${ovmf_package}"
+
+info "Done Building"
+pwd
```
fidencio, line 74: Is this a leftover to show the directory you're at?  I think this can be removed.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+
+build_path="Build/${package_output_dir}/${build_target}_${toolchain}/FV/OVMF.fd"
+stat "${build_path}"
+
+#need to leave tmp dir
+popd
+
+info "Install fd to destdir"
+mkdir -p "$DESTDIR/$PREFIX/share/ovmf"
+cp $build_root/$ovmf_dir/"${build_path}" "$DESTDIR/$PREFIX/share/ovmf"
```
fidencio, line 84: What are the files you're coping?
No need to copy `OVMF_{CODE,VARS}.fd`? If not, why not? :-)

fidencio (commented):
@Alex-Carter01, This one looks good, and I'm making several suggestions to what I think that would make it cleaner.

Please, let me know your thoughts on those.

fidencio:
One last bit, it'd be good if we can have it already being built in a container.  So, I'd suggest having a dockerfile introduced as part of the first commit, in a similar way we already have it for QEMU / Kernel.

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -0,0 +1,94 @@
...
+ovmf_repo="${ovmf_repo:-}"
+ovmf_dir="edk2"
+ovmf_version="${ovmf_version:-}"
+kata_version="${kata_version:-}"
+DESTDIR="${DESTDIR:-../../destdir}"
+PREFIX="${PREFIX:-/opt/kata}"
+ovmf_build="${ovmf_build:-sev}"
+architecture="X64"
+toolchain="GCC5"
+build_target="RELEASE"
```
fidencio, line 26: @Alex-Carter01, what about not hardcoding this ones?
We could default to those values, but let the person doing the build specify env-vars for those, as done with the ones above.

Alex-Carter01:
These changes are for most of the feedback from @fidencio regarding having the first commit be for generic (X86_64) OVMF script. This commit still needs an accompanying dockerfile.

The second commit will reinstate the AmdSev build options

tools/packaging/static-build/ovmf/build-ovmf.sh
Alex-Carter01: Variables now allow specifying different values

tools/packaging/static-build/ovmf/build-ovmf.sh
Alex-Carter01: The content of that directory should just be the OVMF.fd file

tools/packaging/static-build/ovmf/build-ovmf.sh
Alex-Carter01: Renamed, and made the default option. It is the only inclusion in the first commit.

fidencio:
@Alex-Carter01, the first commit seems good from a quick look.
I'd like to have the dockerfile as part of this first commit, and then the SEV part as part of a second commit, in this very same PR.

Makes sense?

Alex-Carter01:
The first commit now has the dockerfile. The new build script makes the docker container and then runs the ovmf script inside of it. My next step is adding the second commit with the AmdSev additions.

Alex-Carter01:
Added the second commit with the extra option for the AmdSev build

tools/packaging/static-build/ovmf/build.sh
```diff
@@ -0,0 +1,25 @@
...
+
+DESTDIR=${DESTDIR:-${PWD}}
+PREFIX=${PREFIX:-/opt/kata}
+container_image="kata-ovmf-builder"
+
+sudo docker build -t "${container_image}" "${script_dir}"
+
+sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
+	-w "${PWD}" \
+	--env DESTDIR="${DESTDIR}" --env PREFIX="${PREFIX}" \
```
fidencio, line 23: As mentioned above, I'd add more environment variables to be passed / respected here, so `yq` can be removed as a hard dependency.

fidencio (changes requested):
@Alex-Carter01, thanks for the work on this, I've added a few suggestions.

tools/packaging/static-build/ovmf/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+	    git \
+        make \
+        g++ \
+        gcc \
+        uuid-dev \
+        python \
+        python3 \
```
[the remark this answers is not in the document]
Alex-Carter01: I found it interesting, but yes the build fails without both python and python3

tools/packaging/static-build/ovmf/build.sh
Alex-Carter01: I will cleanup the ordering and use variables outside so yq isn't needed in the docker container.

Alex-Carter01:
This change passes variables into the docker environment so yq is not needed inside the container.

tools/packaging/static-build/ovmf/Dockerfile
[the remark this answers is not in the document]
fidencio: Yeah, weird. :-)

tools/packaging/static-build/ovmf/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2022 IBM
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ubuntu:20.04
+ENV DEBIAN_FRONTEND=noninteractive
+ENV INSTALL_IN_GOPATH=false
```
fidencio, line 7: As `yq` is not installed anymore, this bit should be dropped.

And once it's dropped, we're good to have this one merged.  Thanks @Alex-Carter01!

Alex-Carter01:
This change removes the holdover go path variable now that yq is no longer needed inside the docker container from the first commit.

Alex-Carter01:
Fixing a static check I saw failing for trailing whitespace in the versions file.

wainersm:
Hi @Alex-Carter01 @fidencio ,  I cloned this on my local Fedora 34 x86_64 machine, ran the build script and it built OVMF just fine. I didn't check the built firmware works as I don't have a SEV machine. Anyway, I think it is ready to land in main...so we keep the work to enable SEV CI.

fidencio (approved):
lgtm, thanks @Alex-Carter01!

fidencio:
> Hi @Alex-Carter01 @fidencio , I cloned this on my local Fedora 34 x86_64 machine, ran the build script and it built OVMF just fine. I didn't check the built firmware works as I don't have a SEV machine. Anyway, I think it is ready to land in main...so we keep the work to enable SEV CI.

Glad your comment was positive, as it showed up just after I merged the PR. :-)

Issue #4179, opened 2022-04-29
kernel building: efi_secret module


Alex-Carter01:
As part of #3982, SEV needs to build a fork of the kernel with efi_secret, and needs to build the efi_secret kernel module.

PR #4180, opened 2022-04-29
kernel building: efi_secret module

Alex-Carter01:
Add kernel fork for sev to kernel builder with efi_secret. Additionally, install efi_secret module for sev.

Fixes: #4179
Signed-off-by: Alex Carter <alex.carter@ibm.com>

Misc note: This feature will likely be upstreamed in the next kernel release. When this happens, there will no longer need to be a separate entry to the versions file or a get_sev_kernel function, but there will still need to be a check to build the kernel module, and the changes to the kernel config.


fidencio:
@Alex-Carter01, as you have the author and the signed-off-by emails not matching, you're going to face the same issue here that you faced with the other PR.

Apart from this, the PR lgtm!

@amshinde, mind to also take a look at this one?

Alex-Carter01:
Updated the commit's author email to match the sign-off email.

fidencio:
@Alex-Carter01, the commit message check is failing due to unknown subsystem.  Please, just use `kernel: build efi_secret module for SEV`.  That will fix the problem.

fidencio:
In general, LGTM.  Thanks @Alex-Carter01 !

Alex-Carter01:
Changed the title of the commit message.

Alex-Carter01:
Fixing static checks for tabbing and line length in the versions file.

@fidencio, who would you recommend as a good person to review this?

fidencio:
> Fixing static checks for tabbing and line length in the versions file.
> 
> @fidencio, who would you recommend as a good person to review this?

@amshinde would be a good fit for reviewing this PR.

tools/packaging/kernel/build-kernel.sh
```diff
@@ -132,6 +132,23 @@ get_tdx_kernel() {
...
+
+	kernel_url=$(get_from_kata_deps "assets.kernel.sev.url")
+	kernel_tarball="${version}.tar.gz"
+
+	if [ ! -f "${kernel_tarball}" ]; then
+	   curl --fail -OL "${kernel_url}${kernel_tarball}"
+	fi
+	
+	mkdir -p ${kernel_path}
+	tar --strip-components=1 -xf ${kernel_tarball} -C ${kernel_path}
```
jodh-intel, line 149: I suspect this is based on `get_tdx_kernel()` which has the same limitation but I'd prefer that all downloads are checksumed, as we do in the default `get_kernel()`.

tools/packaging/kernel/build-kernel.sh
```diff
@@ -399,6 +419,9 @@ build_kernel() {
 	arch_target=$(arch_to_kernel "${arch_target}")
 	pushd "${kernel_path}" >>/dev/null
 	make -j $(nproc) ARCH="${arch_target}"
+	if [ "${conf_guest}" == "sev" ]; then
+		make -j $(nproc) INSTALL_MOD_STRIP=1 INSTALL_MOD_PATH=${kernel_path} modules_install
```
jodh-intel, line 423: Nit: This is a potential DoS scenarios when run on a dev box. I'd suggest something like:

```suggestion
		make -j $(nproc --ignore=1) INSTALL_MOD_STRIP=1 INSTALL_MOD_PATH=${kernel_path} modules_install
```

... but that would require an extra commit to tweak `build_kernel()` which has the same issue.

jodh-intel (commented):
Thanks @Alex-Carter01.

tools/packaging/kernel/build-kernel.sh
dubek: `efi_secret` should be part of the mainline kernel when 5.19 is released in about 8-9 weeks. So maybe for now it's good enough without the checksum, and after 5.19 is released we can just download an official checksummed file from kernel.org.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
```diff
@@ -4,3 +4,9 @@ CONFIG_AMD_MEM_ENCRYPT=y
 CONFIG_AMD_MEM_ENCRYPT_ACTIVE_BY_DEFAULT=y
 CONFIG_CRYPTO_DEV_SP_PSP=y
 CONFIG_CRYPTO_DEV_CCP=y
+CONFIG_SECURITYFS=y
+CONFIG_VIRT_DRIVERS=y
+CONFIG_EFI=y
+CONFIG_EFI_SECRET=y
+CONFIG_MODULE_SIG=y
+CONFIG_MODULES=y
```
dubek, line 12: @Alex-Carter01 please remind me why we set `CONFIG_EFI_SECRET=y` instead of `=m` ? Is it because modules are not copied into the initrd? If we're okay with it set to `=y` (built-in), then maybe there's no need to enable `CONFIG_MODULES` and `CONFIG_MODULE_SIG` ?

tools/packaging/kernel/build-kernel.sh
Alex-Carter01: Sure, I can tweak this PR to exclude one processor like you suggested and also look at opening an issue for the default make to avoid the same issue.

Alex-Carter01:
This change adds @jodh-intel 's suggestion for avoiding potentially using all processors in the make. I will update regarding opening a separate commit for a similar fix to the previous line (the broader kernel make command), since that would be a somewhat different scope.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
Alex-Carter01: I think this was a mistake on my end that `m` would be the correct option here. I am looking into this to make sure, but will probably switch to `m` soon. Good catch since I remember this coming up in a prior conversation but not sure where I regressed.

larrydewey (approved):
This looks good to me.

Alex-Carter01:
Fixed `CONFIG_EFI_SECRET` to m, thank you for the good catch @dubek.

Kernel version 5.19-rc1 unfortunately does not seem to be available at https://cdn.kernel.org/pub/linux/kernel/v5.x/, meaning this will still require a different function to pull the kernel regardless of where we get it from.

amshinde (approved):

Issue #4182, opened 2022-04-30
runk: support --all option for runk kill command


liubin:
crictl stop container will issue `kill --all`, with runk this will lead to container can't be stopped and deleted.

PR #4184, opened 2022-04-30
runk: use custom Kill command to support --all option

liubin:
runk uses liboci-cli crate to parse command line options,
but liboci-cli does not support --all option for kill command,
though this is the runtime spec behavior.

But crictl will issue kill --all command when stopping containers,
as a workaround, we use a custom kill command instead of the one
provided by liboci-cli.

Fixes: #4182

PTAL @ManaSugi

### DIFF
--- a/src/tools/runk/src/commands/kill.rs
+++ b/src/tools/runk/src/commands/kill.rs
@@ -3,9 +3,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use crate::Kill;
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @liubin!

ManaSugi:
@liubin LGTM. Thank you for fixing it. I forgot to re-implement this when I switched from own cli to liboci-cli crate.

lifupan (approved):

Issue #4185, opened 2022-04-30
Add a macro to skip a loop easier


liubin:
Add a macro to skip a loop easier:

```diff
-            if d.test_user == TestUserType::RootOnly {
-                skip_loop_if_not_root!(msg);
-            } else if d.test_user == TestUserType::NonRootOnly {
-                skip_loop_if_root!(msg);
-            }
+            skip_loop_by_user!(msg, d.test_user);

```

PR #4186, opened 2022-04-30
agent: Add a macro to skip a loop easier

liubin:
Add a macro to skip a loop easier without using a
if {} else {} condition check.

Fixes: #4185

### DIFF
--- a/src/agent/src/test_utils.rs
+++ b/src/agent/src/test_utils.rs
@@ -85,4 +85,15 @@ pub mod test_utils {
             }
         };
     }
+
+    #[macro_export]
+    macro_rules! skip_loop_by_user {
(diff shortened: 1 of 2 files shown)


src/agent/src/test_utils.rs
```diff
@@ -85,4 +85,15 @@ pub mod test_utils {
             }
         };
     }
+
+    #[macro_export]
+    macro_rules! skip_loop_by_user {
```
fidencio, line 90: Just saying, I am not a big fan of how the macro was names, but I don't have a much better suggestion.

fidencio (approved):
lgtm, thanks @liubin!

GabyCT (approved):

Issue #4188, opened 2022-05-03
agent watchers: emptyDir permission and ownership does not match with source


yibozhuang:
Today in agent `watcher.rs`, when updating the target, we check if the source is a directory and if so, call `fs::create_dir_all()` [here](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/watcher.rs#L114)

There is no additional set permission here on the destination directory performed here so this can lead to the destination directory permission being different than source directory.
The same applies for `copy` [here](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/watcher.rs#L91) where if the source path had a different `uid` and `gid` than `root`, then the destination file `uid/gid` will not match the source file ownership.

Both of these cases can lead to permission issues.

Example:

I have a pod that uses `emptyDir`
with spec

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: foo
spec:
  containers:
  - name: pod1
    image: busybox
    command: ["sh"]
    args: ["-c", "sleep infinity"]
    volumeMounts:
    - mountPath: /cache
      name: test-cache
  volumes:
  - name: test-cache
    emptyDir: {}
```

Here once I apply the pod with runtime `runc`, and then exec into the pod, I see the `/cache` directory has permission
```
drwxrwxrwx    2 root     root             6 May  2 23:43 cache
```

Where as when I use pod with Kata, this is the permission of the `/cache` directory
```
drwxr-xr-x    2 root     root             6 May  2 23:47 cache
```

Ideally the directory permissions should be identical when using Kata vs. `runc`

PR #4189, opened 2022-05-03
agent watchers: ensure uid/gid is preserved on copy/mkdir

yibozhuang:
Today in agent watchers, when we copy files/symlinks
or create directories, the ownership of the source path
is not preserved which can lead to permission issues.

In copy, ensure that we do a chown of the source path
uid/gid to the destination file/symlink after copy to
ensure that ownership matches the source ownership.
fs::copy() takes care of setting the permissions.

For directory creation, ensure that we set the permissions
of the created directory to the source directory permissions
and also perform a chown of the source path uid/gid to
ensure directory ownership and permissions matches to the source.

Fixes: #4188

### DIFF
--- a/src/agent/src/watcher.rs
+++ b/src/agent/src/watcher.rs
@@ -6,6 +6,7 @@
 #![allow(unknown_lints)]
 
 use std::collections::HashMap;
+use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::SystemTime;
@@ -13,6 +14,7 @@ use std::time::SystemTime;
 use anyhow::{ensure, Context, Result};
 use async_recursion::async_recursion;
 use nix::mount::{umount, MsFlags};
+use nix::unistd::{Gid, Uid};
 use slog::{debug, error, info, warn, Logger};
 use thiserror::Error;
 use tokio::fs;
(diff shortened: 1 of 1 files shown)


src/agent/src/watcher.rs
```diff
@@ -80,16 +82,30 @@ impl Drop for Storage {
 }
 
 async fn copy(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<()> {
-    if fs::symlink_metadata(&from).await?.file_type().is_symlink() {
+    let mut metadata = fs::symlink_metadata(&from).await?;
+    let mut gid = metadata.gid();
```
mxpv, line 86: Nit: to keep things symmetric, you can do:

```rust
let uid;

if condition {
    uid = 1;
} else {
    uid = 2;
}

```

mxpv (approved):

jodh-intel (approved):
Thanks @yibozhuang.

lgtm

egernst:
thanks Yibo!  nit from our static check: 
```
ERROR: Commit 328b67cb8069f81f984aa636a5f6853026dd63f7: Failed to find subsystem in subject: "agent watchers: ensure uid/gid is preserved on copy/mkdir"
```

Seems it'd be better if it was `agent: watcher: ...`

BbolroC:
Sorry for the test failure on s390x. There was an issue due to lack of memory. I doubled it. This time the test will pass.

src/agent/src/watcher.rs
yibozhuang: Thanks Maksym, I've actually further simplified this without needing the if else since calling `symlink_metadata` on a non-symlink file gets the same result as calling `metadata` so we don't need the if else check and can simply just use the `metadat.uid()` and `metadata.gid()`.

egernst (approved):

egernst:
seems powerpc test is having a bad day.  Restarted again...

fidencio:
> seems powerpc test is having a bad day. Restarted again...

Yeah, I've spoken with @Amulyam24 and we're making it not required.

Issue #3963, opened 2022-03-25
Create a GitHub action to skip the CI for doc-only changes


jodh-intel:
## Background

We now [run the static checks in a GitHub action](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/static-checks.yaml).

## Idea

This means that we could create a GH action that:

- Checks if a PR only changes doc files.
- If it does, add the `force-skip-ci` label to the PR.

Then, when reviewers run the CI (using `./test`), the CI will become a NOP.

## Advantages

- Faster to land doc changes

  Doc changes are arguably more important than code changes.

- Saves real money

 The [real CI](http://ci.katacontainers.io) costs the project real money to run).

## Example PR

For a recent example of a PR that qualifies, see https://github.com/kata-containers/kata-containers/pull/3930.

## Plan

The new GHA could be similar to the [PR sizing action](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/add-pr-sizing-label.yaml) (which checks a PR and adds a label based on a condition).

PR #4195, opened 2022-05-03
Create docs-only workflow action

glyphek:
#3963


.github/workflows/check-docs-only.yaml
```diff
@@ -0,0 +1,40 @@
...
+  check_and_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      
+      - name: Remove label
+        uses: actions-ecosystem/action-remove-labels@v1
+        with:
+          labels: force-skip-ci
```
jodh-intel, line 26: I'm not clear why this step is required since even if a PR is not "doc only", it could still be valid for it to be labelled with `force-skip-ci`.

.github/workflows/check-docs-only.yaml
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+name: docs-only-check
```
jodh-intel, line 6: Could you make this a bit more verbose. Maybe something like:

```suggestion
name: Skip the CI if the PR only changes document files.
```

jodh-intel (commented):
Thanks @glyphek - a few comments...

Please could you also squash the two commits into one. You'll also need to tweak the format of your commit to match for format specified in the doc below:

- https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format
  (See the output of the failing "Commit Message Check" GitHub action)

cmaf:
Hi @glyphek, any update to this? It would be really useful to get merged!

liubin:
First, this PR, currently created a new workflow job that will skip the other GitHub actions jobs by adding a `force-skip-ci` label, but should this job run in parallel with other jobs that depends on this `force-skip-ci` label, is it possible that other jobs have run before this job adding the label.

For `.github/workflows/static-checks.yaml`, I think we may do other more work to reduce the GitHub actions time.

Current the job `Static Checks`'s source is:

```yaml
    - name: Static Checks
      if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
      run: |
        cd ${GOPATH}/src/github.com/${{ github.repository }} && make static-checks
```

The static-check target in Makefile of src/runtime is:

```
static-checks: build
	bash ci/static-checks.sh
```

It depends on `build`, that is meaning it will first build runtime/agent, this is s time-consuming operation.

So my suggestion is to split a new doc-check job, like this:

```yaml
    - name: Doc Only Static Checks
      if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
      run: |
        cd ${GOPATH}/src/github.com/${{ github.repository }} && make docs-url-alive-check

```

Thankful that we have already a `docs-url-alive-check` ready to check docs, maybe we should rename it to more common name, like `docs-check`, since doesn't only check the URL connectivity.

And then, if this is a docs only PR, we can skip remain job steps.

Last, we should delete the docs check from the old static check step.

PR #4197, opened 2022-05-03
clh: Rely on Cloud Hypervisor for generating the device ID

fidencio:
We're currently hitting a race condition on the Cloud Hypervisor's
driver code when quickly removing and adding a block device.

This happens because the device removal is an asynchronous operation,
and we currently do *not* monitor events coming from Cloud Hypervisor to
know when the device was actually removed.  Together with this, the
sandbox code doesn't know about that and when a new device is attached
it'll quickly assign what may be the very same ID to the new device,
leading to the Cloud Hypervisor's driver trying to hotplug a device with
the very same ID of the device that was not yet removed.

This is, in a nutshell, why the tests with Cloud Hypervisor and
devmapper have been failing every now and then.

The workaround taken to solve the issue is basically *not* passing down
the device ID to Cloud Hypervisor and simply letting Cloud Hypervisor
itself generate those, as Cloud Hypervisor does it in a manner that
avoids such conflicts.  With this addition we have then to keep a map of
the device ID and the Cloud Hypervisor's generated ID, so we can
properly remove the device.

This workaround will probably stay for a while, at least till someone
has enough cycles to implement a way to watch the device removal event
and then properly act on that.  Spoiler alert, this will be a complex
change that may not even be worth it considering the race can be avoided
with this commit.

Fixes: #4176

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -165,6 +165,7 @@ type cloudHypervisor struct {
 	APIClient      clhClient
 	ctx            context.Context
 	id             string
+	devicesIds     map[string]string
 	vmconfig       chclient.VmConfig
 	state          CloudHypervisorState
 	config         HypervisorConfig
@@ -358,6 +359,7 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
 
 	clh.id = id
 	clh.state.state = clhNotReady
+	clh.devicesIds = make(map[string]string)
 
 	clh.Logger().WithField("function", "CreateVM").Info("creating Sandbox")
 
@@ -670,7 +672,6 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
 	clhDisk := *chclient.NewDiskConfig(drive.File)
 	clhDisk.Readonly = &drive.ReadOnly
 	clhDisk.VhostUser = func(b bool) *bool { return &b }(false)
-	clhDisk.Id = &driveID
 
 	diskRateLimiterConfig := clh.getDiskRateLimiterConfig()
 	if diskRateLimiterConfig != nil {
@@ -683,6 +684,7 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
 		return fmt.Errorf("failed to hotplug block device %+v %s", drive, openAPIClientError(err))
 	}
 
+	clh.devicesIds[driveID] = pciInfo.GetId()
 	drive.PCIPath, err = clhPciInfoToPath(pciInfo)
 
 	return err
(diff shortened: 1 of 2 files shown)


fidencio:
I'm adding the `do-not-merge` label till I hear back from the Cloud Hypervisor team.

fidencio:
The tests failing are due to field alignment, which I will fix Tomorrow morning:
```
clh.go:162:22: fieldalignment: struct with 1232 pointer bytes could be 1056 (govet)
```

Ah, How I wish we'd have a tool that could properly take care of that, maybe a compiler? :-)

fidencio:
http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-1804-PR-cloud-hypervisor-k8s-containerd-devmapper/90/console is another one with 20 consecutive passes.

fidencio:
http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-1804-PR-cloud-hypervisor-k8s-containerd-devmapper/92/console is another one with 20 consecutive passes.

fidencio:
@Amulyam24, any known issues with the power CI?

snir911:
Thanks! nicely explained 
nit :P 
`This is, in a nutshell, why the tests with Cloud Hypervisor and
devmapper have been <failing?> every now and then.`

fidencio:
> Thanks! nicely explained nit :P `This is, in a nutshell, why the tests with Cloud Hypervisor and devmapper have been <failing?> every now and then.`

Thanks!  I've updated the commit message.

Amulyam24:
> @Amulyam24, any known issues with the power CI?

Hi @fidencio, the Power CI is failing at running the stability test. Will need to debug further. In the meantime would you recommend marking it as non-required?

fidencio:
> > @Amulyam24, any known issues with the power CI?
> 
> Hi @fidencio, the Power CI is failing at running the stability test. Will need to debug further. In the meantime would you recommend marking it as non-required?

Yeah, I will mark it as non-required then in order to cut a release later Today.
Would you mind to open an issue about this in the kata-containers/ci repo so we can keep track of those changes?

src/runtime/virtcontainers/clh.go
```diff
@@ -670,7 +672,6 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
 	clhDisk := *chclient.NewDiskConfig(drive.File)
 	clhDisk.Readonly = &drive.ReadOnly
 	clhDisk.VhostUser = func(b bool) *bool { return &b }(false)
-	clhDisk.Id = &driveID
```
pmores, line 675: Just out of curiosity (and my apologies if the following questions are stupid as I'm totally not familiar with this code):

- we're now leaving `clhDisk.Id` uninitialised, is that OK?  From my cursory check the surrounding code seems prepared in general to handle an empty `clhDisk.Id`, just asking if there might be a code path that relies on a non-empty value for case of a hot-plugged drive.
- would it be possible to store the new device ID right in `clhDisk.Id` instead of a newly added map?

fidencio: @pmores, great question.

The clhDisk.Id is only used for raising a request to Cloud Hypervisor, and then, **if** filled, Cloud Hypervisor will use the id passed to manage the device internally using that Id.  If we do **not** pass anything, Cloud Hypervisor will generate it automatically, and that's the thing we're taking advantage here.

Storing it clhDisk is **not** the place to store it, as it's ephemeral, it's just a variable created to raise a request for Cloud Hypervisor to add a disk on its side.  Once this function is over, that variable is out of scope and doesn't exist anymore.  That's the reason we store the Cloud Hypervisor's generated id in a map, and then use that to match the drive / device ID coming from the sandbox layer.

The question then is ... can we change the `drive.ID`?  And the answer is not without a really good investigation and a deep dive into what's done inside other VMMs, and that's something I'm not willing to risk while solving a bug that close to the release.

Did I answer all your questions, @pmores?

sboeuf (approved):
Looking good!

src/runtime/virtcontainers/clh.go
pmores: @fidencio That makes perfect sense, thanks a lot for the explanation. I have no further questions - would you like me also to approve the PR formally?

fidencio: If you're comfortable doing so, I'd appreciate that. :-)

pmores (approved):

Issue #4219, opened 2022-05-09
Don't run "Docs URL Alive Check" workflow on forked repos


liubin:
Sometimes I get error emails from GitHub that say "Docs URL Alive Check failed" on my forked repo.

[This workflow](https://github.com/kata-containers/kata-containers/blob/aa6886f1edf96f3afeb72f76f9baa88b255ec6cc/.github/workflows/docs-url-alive-check.yaml) should only run https://github.com/kata-containers/kata-containers but not the forked ones.

PR #4220, opened 2022-05-09
ci: Don't run Docs URL Alive Check workflow on forks

liubin:
This workflow is a scheduled job that runs at 23:00
every Sunday, it should only run the main repo
but not the forked ones.

Fixes: #4219

**NOTE:** This is review-able but I added a DNM label for verification.

/cc @Bevisy @jodh-intel

### DIFF
--- a/.github/workflows/docs-url-alive-check.yaml
+++ b/.github/workflows/docs-url-alive-check.yaml
@@ -14,31 +14,31 @@ jobs:
       target_branch: ${{ github.base_ref }}
     steps:
     - name: Install Go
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @liubin and good catch! :smile:

lgtm

Bevisy (approved):
I checked my forked repo and it doesn't have the Actions tab. So there has not been this problem, it should be related to repository role permissions. This change looks good to me.

lifupan (approved):

Issue #4250, opened 2022-05-13
Agent: is_signal_handled does not remove space before the hex string causing conversion to u64 to fail


Champ-Goblem:
We were experiencing issues previously with pods that were terminating which has now been fixed and release by https://github.com/kata-containers/kata-containers/pull/4072. When we were initially debugging this we noticed that the signal handling in Kata seemed to differ from the steps that runC performed. It was noticed that a SIGTERM was not being observed for terminating workloads and instead these were being immediately killed through SIGKILL. This proved to be problematic for stateful workloads, especially databases, which hook into the TERM signal in order to shutdown the database properly. We recently rolled out Kata 2.4.1 to our clusters and we still noticed that our databases were reporting unclean shutdown.

I have managed to pin this to a bug in the `is_signal_handled` code when fetching the `SigCgt` field from `/proc/<pid>/status`. The current function splits this line by a `:` which separates the two values, unfortunately there is no handling for the space/tab that is present after the colon, before the hex string (eg `SigCgt:	0000000000000000`). This causes the function to return false even in the case of a signal handler being installed, and the agent logs to the vmconsole `failed to parse the str \\t0000000000000000 to hex`.

PR #4252, opened 2022-05-13
agent: Fix is_signal_handled failing parsing str to u64

Champ-Goblem:
In the is_signal_handled function, when parsing the hex string returned
from `/proc/<pid>/status` the space/tab character after the colon
is not removed.

This patch trims the result of SigCgt so that
all whitespace characters are removed. It also extends the existing
test cases to check for this scenario.

Fixes: #4250
Signed-off-by: Champ-Goblem <cameron@northflank.com>

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -1682,7 +1682,7 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
                 warn!(sl!(), "parse the SigCgt field failed");
                 return false;
             }
-            let sig_cgt_str = mask_vec[1];
+            let sig_cgt_str = mask_vec[1].trim();
             let sig_cgt_mask = match u64::from_str_radix(sig_cgt_str, 16) {
                 Ok(h) => h,
                 Err(_) => {
@@ -2450,6 +2450,26 @@ OtherField:other
             },
             TestData {
                 status_file_data: Some("SigCgt:000000004b813efb"),
+                signum: 4,
+                result: true,
+            },
+            TestData {
+                status_file_data: Some("SigCgt:\t000000004b813efb"),
+                signum: 4,
+                result: true,
+            },
+            TestData {
+                status_file_data: Some("SigCgt: 000000004b813efb"),
+                signum: 4,
+                result: true,
+            },
+            TestData {
+                status_file_data: Some("SigCgt:000000004b813efb "),
+                signum: 4,
+                result: true,
+            },
+            TestData {
+                status_file_data: Some("SigCgt:\t000000004b813efb "),
                 signum: 4,
                 result: true,
             },


egernst:
Good find @Champ-Goblem -- is it feasible to add some unit tests to exercise this?

src/agent/src/rpc.rs
```diff
@@ -1682,7 +1682,10 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
                 warn!(sl!(), "parse the SigCgt field failed");
                 return false;
             }
-            let sig_cgt_str = mask_vec[1];
+            let mut sig_cgt_str = mask_vec[1];
```
egernst, line 1685: I wonder if we should introduce a function here for parsing the field (maybe starting with [1679](https://github.com/kata-containers/kata-containers/pull/4252/files#diff-f87c1c212bba3abd0b07bafe69d5b01ad00dfe2c01bed585e2a75fcbb6442b89L1679)?), which would allow us to be able to add same unit tests?

src/agent/src/rpc.rs
```diff
@@ -1682,7 +1682,10 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
...
+            if !sig_cgt_str.chars().nth(0).unwrap().is_numeric() {
```
fidencio, line 1686: Out of curiosity, can't use simply  trim()` the string?

If that works, that would be cleaner, IMHO.

Note: https://doc.rust-lang.org/std/primitive.str.html#method.trim

src/agent/src/rpc.rs
```diff
@@ -1682,7 +1682,10 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
...
+                sig_cgt_str = &sig_cgt_str[1..sig_cgt_str.len()-1]
```
cmaf, line 1687: @Champ-Goblem if you end up sticking with this approach could you please update the format? Our static check failed on `cargo fmt -- --check`
```
-                sig_cgt_str = &sig_cgt_str[1..sig_cgt_str.len()-1]
+                sig_cgt_str = &sig_cgt_str[1..sig_cgt_str.len() - 1]
```

src/agent/src/rpc.rs
Champ-Goblem: Nice spot, trim works perfectly, I have updated the code to use this

src/agent/src/rpc.rs
Champ-Goblem: I ended up using .trim in the end but I ran this against the new commit and there should be no issues now with formatting

Champ-Goblem:
@egernst I see there already exists some tests for this function `test_is_signal_handled`, it doesnt look like it has any cases to cover a whitespace so I have extended this to include some

src/agent/src/rpc.rs
Champ-Goblem: This should be covered by `test_is_signal_handled` already, let me know if you would prefer to split the function down and change the related tests

fidencio (approved):
lgtm, thanks @Champ-Goblem!

cmaf (approved):

fidencio:
Hmm.  CIs are failing to find the C version of virtiofsd and I **think** it would be easily solved with a rebase.

fidencio:
Okay, I was able to rebase it myself.
Let me try again ...

/test

fidencio:
```
time="2022-05-17T07:47:28-04:00" level=error msg="unable to delete test" error="write vsock host(2):1562314465->vm(256215773):1024: cannot allocate memory: unknown"
ctr: write vsock host(2):1562314465->vm(256215773):1024: cannot allocate memory: unknown
Ending hypervisor stability test
[hypervisor_stability_kill_test.sh:233] INFO: Wait until the containers gets removed
ctr: write vsock host(2):1562314465->vm(256215773):1024: cannot allocate memory: unknown
make: *** [Makefile:73: stability] Error 1
```

Power CI failure is unrelated.

PR #2810, opened 2021-10-08
docs: use-cases: Update Intel SGX use case

mythi:
The upstream kernel SGX support has changed drastically since
the initial version of the Intel SGX use case doc was written.

The updated use case documents how to easily setup SGX with
Kata Containers running in a Kubernetes cluster.

Fixes: #2811 

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

PR #4267, opened 2022-05-16
packaging: Add kernel config option for SGX in Gramine

### LINKED ISSUE #4266: Enable kernel config option for SGX support with Gramine
@cmaf, while you're working on this, I was taking a look at https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/using-Intel-SGX-and-kata.md#notes and the last bullet point mentions:
"""
Projects like [Gramine Shielded Containers (GSC)](https://gramine-gsc.readthedocs.io/en/latest/) are also known to work. For GSC specifically, the Kata guest kernel needs to have the CONFIG_NUMA=y enabled and at least one CPU online when running the GSC container.
"""

Although orthogonal to this PR, do you think it'd be worth to open a new PR enabling the CONFIG_NUMA so projects like Gramine would work out-of-the-box?

_Originally posted by @fidencio in https://github.com/kata-containers/kata-containers/issues/4255#issuecomment-1127297360_

cmaf:
For the Gramine Shielded Containers guest kernel, CONFIG_NUMA must be
enabled.

Fixes  #4266


fidencio:
cc @mythi

fidencio (commented):
@cmaf, you also need to bumo the kernel version: https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kernel/kata_config_version

fidencio:
I was wondering whether the docs should also be updated as well, but I think we should wait till a release is out, then we update the docs reflecting the state of what users can have their hands on without the need to build the project themselves.  WDYT?

fidencio (approved):
lgtm, thaks @cmaf!

GabyCT (approved):

cmaf:
> I was wondering whether the docs should also be updated as well, but I think we should wait till a release is out, then we update the docs reflecting the state of what users can have their hands on without the need to build the project themselves. WDYT?

@fidencio I have plans to update the SGX docs, next release is a good target

cmaf:
Note: failures depend on https://github.com/kata-containers/tests/pull/4792 being fixed, retest once merged

tools/packaging/kernel/configs/fragments/x86_64/sgx.conf
```diff
@@ -1 +1,4 @@
+# Intel Software Guard Extensions
+
 CONFIG_X86_SGX=y
+CONFIG_NUMA=y
```
amshinde, line 4: @cmaf Do you know why this option is required? Also can you add a comment here that it is specifically for Gramine.

cmaf: @amshinde it is specified in our own documentation in a [note](https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/using-Intel-SGX-and-kata.md#notes) for using kata containers with Gramine:

https://github.com/kata-containers/kata-containers/blob/6ffdebd202223407f0306ea346ba777836923465/docs/use-cases/using-Intel-SGX-and-kata.md?plain=1#L100-L102

I can't find a reference in it in the Gramine docs and don't know offhand why it is required, so maybe @mythi or @jodh-intel could shed some light on whether or not it is still a requirement. (Referring to changes in 8c8bcb7 https://github.com/kata-containers/kata-containers/pull/2810)

mythi: Gramine wants to verify the CPU/NUMA topology etc. Without this config option, `/sys/devices/system/node/*` etc are not available for containers to read and gramine fails because of that.

mythi:
Thanks!

PR #4271, opened 2022-05-17
runtime: do not check for EOF error in console watcher

### LINKED ISSUE #4079: func (s *Scanner) Err() will not return EOF error
Through https://pkg.go.dev/bufio#Scanner.Err :

> Err returns the first non-EOF error that was encountered by the Scanner.

But line 1040 checks the error:

https://github.com/kata-containers/kata-containers/blob/1e301482e7c5cb8ae5cea4ea89ead37e6ba03660/src/runtime/virtcontainers/sandbox.go#L1039-L1049

r4f4:
The documentation of the bufio package explicitly says

"Err returns the first non-EOF error that was encountered by the
Scanner."

When io.EOF happens, `Err()` will return `nil` and `Scan()` will return
`false`.

Fixes #4079

Signed-off-by: Rafael Fonseca <r4f4rfs@gmail.com>

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -1037,15 +1037,13 @@ func (cw *consoleWatcher) start(s *Sandbox) (err error) {
 		}
 
 		if err := scanner.Err(); err != nil {
-			if err == io.EOF {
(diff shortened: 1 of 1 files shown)


cmaf (approved):
thanks @r4f4

PR #4274, opened 2022-05-17
stable-2.4: backport agent fixes

egernst:
Backport signal handling fix, as well as agent test improvements.

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -432,38 +432,14 @@ fn get_container_pipe_size(param: &str) -> Result<i32> {
 
 #[cfg(test)]
 mod tests {
+    use crate::assert_result;
+
     use super::*;
     use anyhow::anyhow;
     use std::fs::File;
     use std::io::Write;
     use std::time;
     use tempfile::tempdir;
-
-    // Parameters:
-    //
-    // 1: expected Result
-    // 2: actual Result
-    // 3: string used to identify the test on error
-    macro_rules! assert_result {
-        ($expected_result:expr, $actual_result:expr, $msg:expr) => {
-            if $expected_result.is_ok() {
-                let expected_level = $expected_result.as_ref().unwrap();
-                let actual_level = $actual_result.unwrap();
-                assert!(*expected_level == actual_level, "{}", $msg);
-            } else {
-                let expected_error = $expected_result.as_ref().unwrap_err();
-                let expected_error_msg = format!("{:?}", expected_error);
-
-                if let Err(actual_error) = $actual_result {
-                    let actual_error_msg = format!("{:?}", actual_error);
-
-                    assert!(expected_error_msg == actual_error_msg, "{}", $msg);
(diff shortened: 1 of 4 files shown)


fidencio:
I am not sure if I understand the errors I'm seeing here.  It seems to be using `main` tests branch against a `stable-2.4` kata-containers branch.

fidencio:
I am not sure if I understand the errors I'm seeing here.  It seems to be using `main` tests branch against a `stable-2.4` kata-containers branch.

fidencio:
> I am not sure if I understand the errors I'm seeing here. It seems to be using `main` tests branch against a `stable-2.4` kata-containers branch.

Okay, no, that's **not** the case.
What's actually happening is that the cached QEMU version used is the same, regardless to which branch is being used.  However, we stopped building virtiofsd as part of QEMU on `main`, so the cached version doesn't have it ... and here we are trying to install it.

@wainersm, any idea on how to easily solve this?

wainersm:
> 

> > I am not sure if I understand the errors I'm seeing here. It seems to be using `main` tests branch against a `stable-2.4` kata-containers branch.
> 
> Okay, no, that's **not** the case. What's actually happening is that the cached QEMU version used is the same, regardless to which branch is being used. However, we stopped building virtiofsd as part of QEMU on `main`, so the cached version doesn't have it ... and here we are trying to install it.
> 
> @wainersm, any idea on how to easily solve this?

easily... hmm... maybe change the value of `qemu_latest_build_url` so use an invalid job, it will give "cache miss" and then try to build locally:

```
diff --git a/.ci/install_qemu.sh b/.ci/install_qemu.sh
index f20f6889..3137f23a 100755
--- a/.ci/install_qemu.sh
+++ b/.ci/install_qemu.sh
@@ -220,7 +220,7 @@ main() {
                vanilla)
                        CURRENT_QEMU_VERSION=$(get_version "assets.hypervisor.qemu.version")
                        QEMU_REPO_URL=$(get_version "assets.hypervisor.qemu.url")
-                       qemu_latest_build_url="${jenkins_url}/job/kata-containers-2.0-qemu-$(uname -m)/${cached_artifacts_path}"
+                       qemu_latest_build_url="${jenkins_url}/job/kata-containers-stable-2.4-qemu-$(uname -m)/${cached_artifacts_path}"
                        ;;
                *)
                        info "qemu type '${qemu_type}' not supported"

```
Ideally we should have cache jobs for the branches.

egernst:
@wainersm -- can you clarify -- we should send a PR on the stable branches to update QEMU caching, then all should be good -- is that right?

egernst:
(rebased)

snir911 (approved):
Thanks! would be nice also to refer to the original PRs/commit in each of the commits but LGTM anyhow

fidencio (approved):
lgtm, thanks @egernst!

jodh-intel (approved):

PR #4276, opened 2022-05-18
virtiofsd: static build virtiofsd from rust code for non-x86

jongwu:
enable build rust version of virtiofsd from source for non-x86_64.
Only test on Arm.


tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
```diff
@@ -40,4 +40,43 @@ pull_virtiofsd_released_binary() {
...
+   git clone ${virtiofsd_url} virtiofsd
+   pushd virtiofsd
+
+   RUSTFLAGS='-C target-feature=+crt-static -C link-self-contained=yes'
+   LIBSECCOMP_LINK_TYPE=static
+   LIBSECCOMP_LIB_PATH=/usr/lib/${ARCH_LIBC}
+   LIBCAPNG_LINK_TYPE=static
+   LIBCAPNG_LIB_PATH=/usr/lib/${ARCH_LIBC}
+   
+   cargo build --release --target $(uname -m)-unknown-linux-${LIBC}
```
Amulyam24, line 73: Hi @jongwu, firstly, Thank you for the PR! 

Here, using `$uname -m` won't work for `ppc64le` since the rust arch is `powerpc64le` causing the build to fail since the target wouldn't exist.
Can we use something from here - https://github.com/kata-containers/tests/blob/main/.ci/kata-arch.sh#L48?

PS: Tested the build on `ppc64le`, it works fine.

gkurz: > Hi @jongwu, firstly, Thank you for the PR!
> 
> Here, using `$uname -m` won't work for `ppc64le` since the rust arch is `powerpc64le`, thereby causing the build to fail as the target wouldn't exist. e: We already have `ARCH` declared above, it could be used instead of `uname -m` and probably a condition can be added for `ppc64le` for changing the arch?
> 

What about getting the ARCH from `rustc` instead of `uname -m` ? Something like the following (or nicer :smiley:) :

```
ARCH=$(rustc -Vv | grep ^host: | cut -d' ' -f2 | cut -d- -f1)
```



> PS: Tested the build on `ppc64le`, it works fine.

fidencio: We already have an `arch_to_golang` function in https://github.com/kata-containers/kata-containers/blob/c88a48be216db18d40ebc6c396c75da10d3857f3/tools/packaging/scripts/lib.sh#L82-L93

We could expand lib.sh and add something like `arch_to_rust`.

tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
```diff
@@ -40,4 +40,43 @@ pull_virtiofsd_released_binary() {
...
+    ARCH_LIBC=${ARCH}-linux-${LIBC}
+}
+	   
+build_virtiofsd_from_source() {
+   echo "build viriofsd from source"
+   init_env
+
+   virtiofsd_url=$(get_from_kata_deps "externals.virtiofsd.url")
+
+   git clone ${virtiofsd_url} virtiofsd
```
fidencio, line 64:
```suggestion
   git clone --depth 1 --branch ${virtiofsd_version} ${virtiofsd_url} virtiofsd
```

This way we clone the version we want to rely on.

tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
```diff
@@ -40,4 +40,43 @@ pull_virtiofsd_released_binary() {
     popd
 }
 
-pull_virtiofsd_released_binary
+init_env() {
+   case ${ARCH} in
```
fidencio, line 44: Please, add `x86_64` as part of the witch as well.  And, as aarch64, it should be using musl.

fidencio (commented):
@jongwu, nice work!

I've left a few comments / suggestion, but it looks good overall.

tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
jongwu: Thanks @fidencio , I will fix it based on all above comments.

fidencio:
@jongwu, I think this is failing for two reasons:
* We have multiple depends-on, please, leave the `Depends-on` just as part of one commit
* We have a cyclic depends-on.  I'd recommend leave that either on the tests repo or here, but not on both.

tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
Amulyam24: >we can just modify "ARCH" in "init_env" in this PR for ppc64le. Am I right? @Amulyam24

right, thanks @jongwu!

fidencio:
Power failure is not related.
/retest-power

fidencio (approved):
lgtm, thanks @jongwu!

I'm assuming there'll be a next patch / next PR removing the ifdef from https://github.com/kata-containers/kata-containers/blob/6d00701ec9f204d8c10358649b4e825431839829/src/runtime/Makefile#L179-L181, right?

Amulyam24 (approved):
Thanks @jongwu!

Jakob-Naucke:
@jongwu I'm on leave of absence, please page @BbolroC for s390x matters (I'm only taking a look occasionally plus I don't have hardware to test)

Issue #4293, opened 2022-05-20
kata-runtime direct-volume stats Action signature is incorrect


yibozhuang:
# Description of problem

Looks like running `kata-runtime direct-volume stats` is returning

```
ERROR invalid Action type. Must be `func(*Context`)` or `func(*Context) error).  This is an error in the application.  Please contact the distributor of this application if this is not you.See https://github.com/urfave/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature
```

I think this is because at
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-volume.go#L92

the `Action` signature is incorrect: it needs to return just `error` and print out the volume stats rather than returning.

# Expected result

Expect `kata-runtime direct-volume stats` command to successfully show the stats of a direct-assigned volume.

# Actual result

```
ERROR invalid Action type. Must be `func(*Context`)` or `func(*Context) error).  This is an error in the application.  Please contact the distributor of this application if this is not you.See https://github.com/urfave/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature
```

# Further information

(replace this text with any extra information you think might be useful)

yibozhuang:
I'll take a stab at fixing this.

PR #4294, opened 2022-05-20
runtime: fix incorrect Action function for direct-volume stats

yibozhuang:
The action function expects a function that returns error
but the current direct-volume stats Action returns
(string, error) which is invalid.

This change fixes the format and print out the stats from
the command instead.

Fixes: #4293


fengwang666 (approved):
Thanks!

egernst (approved):

egernst:
Looks good -- please also back port to stable-2.4!

egernst:
(restarting static checks, prior run had a flake from ARP. test)

PR #4296, opened 2022-05-21
enable compilation of runtime on Darwin hosts

egernst:
We have a goal of being able to run Kata Containers on Darwin based hosts (macOS). Like prior commits before, this PR seeks to be able to refactor parts of the runtime in order to facilitate being able to build. This pull request won't allow for running at this point, as it  is mostly skeleton code. With this PR we are able to now build the runtime, however, which will help ensure that we can continue progressing further.

After this PR, I intend to:
- improve existing unit tests, ensuring that running tests on all runtime packages succeeds on Darwin as well as Linux (today this just Linux)
- Introduce implementations for networking, shared fs, resource-control and hypervisor on Darwin.

Back to this PR....  

This introduces skeleton implementation for:
 - networking
 - shared fs
 - resource control
 - hypervisor + virtualization-framework
 -  kata-check

On top of this, there were changes needed in order to be able to build:
- There were a couple of changes required for tooling (makefile cleanup so it can succeed on 
- Some minor refactoring required within virtcontainers and cmd/kata-runtime in order to isolate more Linux specific code (whether it be unique signals, hypervisors only available on Linux, or assumptions around namespaces) into... linux specific files.
- Revert the QEMU specific code added to verify S390x QEMU CoCo config from arch/host-os agnostic part of code.


egernst:
(I'm still working through to see if I can cleanup the S390x code and config validation so we don't have to revert)

egernst:
eek.  marking WIP.

Issue #4297, opened 2022-05-21
runtime: direct-volume stats does not work today


yibozhuang:
While testing `kata-runtime direct-volume stats`, I noticed that this feature doesn't work today due to 3 issues:

1. on the client side we put the volume path in the url
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-volume.go#L131
while at the shim side, the handler is just for `/direct-volume/stats`
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/shim_management.go#L209
The go default http mux AFAIK doesn’t support pattern routing so right now I am getting 404 on volume stats

The proposed solution is instead of appending the volume path as a route under `/direct-volume/stats`, we should just use the GET query parameter instead.

2. once first issue above is fixed, I found once above is fixed is that I was seeing

```
mount foo not found in sandbox
```
message and this is because we pass in the volume publish path when invoking `kata-runtime direct-volume stats` but in shim `sandbox.go` https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/sandbox.go#L2286 it expects the mount `Source` which for direct-assigned volume, it is the `Device` field under:
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/direct-volume/utils.go#L48
so we need to do a look up of the `MountInfo` metadata from the volume path and then send the `Device` to the shim

3. lastly, in the agent code, [get_volume_capacity_stats](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L1468) and [get_volume_inode_stats](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L1489) it is trying to match the path to the device name which will never match since the path in the pb request to the agent is the mount path in the guest and not the device name, so here we should simplify and just get the fs stats of the mount path directly instead of needing to walk all the disks.

yibozhuang:
I will submit a PR to fix this.

PR #1792, opened 2021-05-04
WIP: RFC: direct assignment

egernst:
POC implementation based on direct-assigned design RFC: https://github.com/kata-containers/kata-containers/pull/1568

PR #4298, opened 2022-05-21
Fix issues with direct-volume stats feature

yibozhuang:
This PR contains the changes needed to make direct-assigned volume stats work as expected.

NOTE: The agent code is derived from @egernst's original PR #1792

1. direct-volume stats update to use GET parameter.
The go default http mux AFAIK doesn’t support pattern
routing so right now client is padding the url
for direct-volume stats with a subpath of the volume
path and this will always result in 404 not found returned
by the shim. This change will update the shim to take the volume
path as a GET query parameter instead of a subpath.
If the parameter is missing or empty, then return
400 BadRequest to the client.

2. direct-volume stats use correct name.
Today the shim does a translation when doing
direct-volume stats where it takes the source and
returns the mount path within the guest.
The source for a direct-assigned volume is actually
the device path on the host and not the publish
volume path. This change will perform a lookup of the mount info
during direct-volume stats to ensure that the
device path is provided to the shim for querying
the volume stats.

3. The current implementation in agent of walking the
disks to match with the requested volume path
in agent doesn't work because the volume path
provided by the shim to the agent is the mount
path within the guest and not the device name.
The current logic is trying to match the
device name to the volume path which will never
match. This change will simplify the
`get_volume_capacity_stats` and
`get_volume_inode_stats` to just call statfs and
get the bytes and inodes usage of the volume
path directly.

Fixes: #4297 

### DIFF
--- a/src/runtime/cmd/kata-runtime/kata-volume.go
+++ b/src/runtime/cmd/kata-runtime/kata-volume.go
@@ -7,10 +7,11 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/url"
 
 	containerdshim "github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2"
-	"github.com/kata-containers/kata-containers/src/runtime/pkg/direct-volume"
+	volume "github.com/kata-containers/kata-containers/src/runtime/pkg/direct-volume"
 	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils/shimclient"
 
 	"github.com/urfave/cli"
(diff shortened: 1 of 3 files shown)


src/runtime/cmd/kata-runtime/kata-volume.go
```diff
@@ -127,8 +128,14 @@ func Stats(volumePath string) ([]byte, error) {
...
 		return nil, err
 	}
-	urlSafeDevicePath := url.PathEscape(volumePath)
-	body, err := shimclient.DoGet(sandboxId, defaultTimeout, containerdshim.DirectVolumeStatUrl+"/"+urlSafeDevicePath)
+	volumeMountInfo, err := volume.VolumeMountInfo(volumePath)
+	if err != nil {
+		return nil, err
+	}
+
+	urlSafeDevicePath := url.PathEscape(volumeMountInfo.Device)
```
fengwang666, line 136: Thanks for fixing this!

fengwang666 (approved):
@yibozhuang Thanks so much for fixing these issues.

egernst (approved):

Issue #4309, opened 2022-05-24
add support for core scheduling


egernst:
In linux 5.14 and hopefully some backports, core scheduling allows processes to
be co scheduled within the same domain on SMT enabled systems.

We should support!

kernel docs: https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/core-scheduling.html

PR #4310, opened 2022-05-24
shim: add support for core scheduling

egernst:
In linux 5.14 and hopefully some backports, core scheduling allows processes to
be co scheduled within the same domain on SMT enabled systems.

Containerd impl sets the core sched domain when launching a shim. This
allows a clean way for each shim(container/pod) to be in its own domain and any
additional containers, (v2 pods) be be launched with the same domain as well as
any exec'd process added to the container.

kernel docs: https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/core-scheduling.html

For Kata specifically, we will look for CORE_SCHED environment variable
to be set to indicate we shuold create a new schedule core domain.

Fixes: #4309

Signed-off-by: Michael Crosby <michael@thepasture.io>


fengwang666 (commented):
Do you how it would play out with cgroup cpuset? For example if the cpu manager static policy is enabled and the pod has guaranteed QOS, the pod will have an exclusive cpu set. Would the cpuset conflict with the share core group?

fengwang666 (approved):

src/runtime/pkg/utils/schedcore.go
```diff
@@ -0,0 +1,38 @@
...
+	// ThreadGroup affects all threads in the group
+	ThreadGroup PidType = pidtypeTgid
+	// ProcessGroup affects all processes in the group
+	ProcessGroup PidType = pidtypePgid
+)
+
+const (
+	pidtypePid  = 0
+	pidtypeTgid = 1
+	pidtypePgid = 2
```
jodh-intel, line 27: - Could you define these values before they are used for easier reading?
- Could you expand these two names? Pid is clear enough but `Tg` and `Pg` are not very clear. Why not just use `pidTypeThreadGroup` and `pidTypeProcessGroup`.

jodh-intel (commented):
Thanks @egernst. A few comments...

Also, a couple of other considerations:

- Should we add an annotation as an alternative to this "magic variable".
- We need to document this variable somewhere.
- What are the plans for Kata 3.x / runtime-rs

/cc @bergwolf, @amshinde.

src/runtime/pkg/utils/schedcore.go
```diff
@@ -0,0 +1,36 @@
...
+	ProcessGroup PidType = pidTypeProcessGroupId
+)
+
+// Create a new sched core domain
+func Create(t PidType) error {
+	return unix.Prctl(unix.PR_SCHED_CORE, unix.PR_SCHED_CORE_CREATE, 0, uintptr(t), 0)
+}
+
+// ShareFrom shares the sched core domain from the provided pid
+func ShareFrom(pid uint64, t PidType) error {
```
bergwolf, line 34: Unused?

docs/design/core-scheduling.md
```diff
@@ -0,0 +1,12 @@
...
+
+Core scheduling is a Linux kernel feature that allows only trusted tasks to run concurrently on
+CPUs sharing compute resources (for example, hyper-threads on a core).
+
+Containerd versions >= 1.6.4 leverage this to treat all of the processes associated with a
+given pod or container to be a single group of trusted tasks. To indicate this should be carried
+out, containerd sets the `SCHED_CORE` environment variable for each shim it spawns. When this is
+set, the Kata Containers shim implementation uses the `prctl` syscall to create a new core scheduling
+domain for the shim process itself as well as future VMM processes it will start.
+
```
bergwolf, line 11: It is still unclear to me how to enable the feature with containerd. Is there a new config option to enable it?

bergwolf:
For runtime-rs equivalency, I created https://github.com/kata-containers/kata-containers/issues/4429

Issue #4324, opened 2022-05-27
runk: Occurs a slient error when tty is used without console socket


ManaSugi:
# Description of problem

Currently, `runk` fails with a slient error when the user tries to launch a container without a console socket (`--console-socket`) whereas enabling terminal in `config.json`.
In the case of `runc`, `runc` exits properly with the following error message when the user tries to do the above operation.

```
# sudo runc create test
ERRO[0000] cannot allocate tty if runc will detach without setting console socket

# runc run --detach test
ERRO[0000] cannot allocate tty if runc will detach without setting console socket
```

`runk` always launches a container with [detached mode](https://github.com/opencontainers/runc/blob/master/docs/terminals.md#detached) when the user uses `run` command, so `runk` should output
an error message when a container is launched in both cases `run` and `create` without a console socket whereas enabling terminal.

# Expected result

`runk` should exit properly with an error message like `runc`.

# Actual result

`runk` fails silently.

```
# jq .process config.json
  "terminal": true,
  "user": {
    "uid": 0,
    "gid": 0
  },
  "args": [
    "echo",
    "Hello runk"
  ],
...

# runk create test                                                             
# runk state test
{
  "ociVersion": "1.0.2-dev",
  "id": "test",
  "pid": 259462,
  "status": "stopped",
  "bundle": "/tmp/runk/bundle",
  "created": "2022-05-27T06:24:57.267258172Z"
}
# runk delete test

# runk run test
// hang up

# (another terminal) ps aux | grep runk
root      259712  0.0  0.0  55240  2544 pts/3    Sl+  15:26   0:00 runk run test
root      259737  0.0  0.0      0     0 pts/3    Z+   15:26   0:00 [runk] <defunct>
```

PR #4325, opened 2022-05-27
runk: Return error when tty is used without console socket

ManaSugi:
runk always launches containers with detached mode,
so users have to use a console socket with `run` and
`create` operations when a terminal is used.
If users set `terminal` to `true` in `config.json` and
try to launch a container without specifying a console
socket, runk returns an error with a message early.

Fixes: #4324

### DIFF
--- a/src/tools/runk/libcontainer/src/builder.rs
+++ b/src/tools/runk/libcontainer/src/builder.rs
@@ -48,6 +48,16 @@ impl Container {
             }
         }
 
+        if let Some(process) = spec.process.as_ref() {
+            // runk always launches containers with detached mode, so users have to
+            // use a console socket with run or create operation when a terminal is used.
+            if process.terminal && self.console_socket.is_none() {
+                return Err(anyhow!(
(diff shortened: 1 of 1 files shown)


src/tools/runk/libcontainer/src/builder.rs
```diff
@@ -48,6 +48,16 @@ impl Container {
             }
         }
 
+        if let Some(process) = spec.process.as_ref() {
+            // runk always launches containers with detached mode, so users have to
+            // use a console socket with run or create operation when a terminal is used.
+            if process.terminal && self.console_socket.is_none() {
+                return Err(anyhow!(
+                    "cannot allocate a pseudo-TTY without setting a console socket"
```
jodh-intel, line 56: Could you add a unit test for this scenario?

ManaSugi: @jodh-intel Thank you for the review. I added the unit test.

liubin (approved):

PR #902, opened 2020-10-09
Validate runtime annotations

c3d:
This fixes #901 and https://bugs.launchpad.net/katacontainers.io/+bug/1878234.

Need to add a link to code reviews once this is merged.

PR #4331, opened 2022-05-28
Allow io.katacontainers.config.hypervisor.enable_iommu annotation by …

### LINKED ISSUE #4330: Enable io.katacontainers.config.hypervisor.enable_iommu annotation by default
# Description of problem

Since #902 the `io.katacontainers.config.hypervisor` pod annotations have only been permitted if explicitly allowed in the global configuration.  The default global configuration allows no such annotations.  That's important because several of those annotations would cause Kata to execute arbitrary binaries, and so were wildly unsafe.

However, this is inconvenient for the `io.katacontainers.config.hypervisor.enable_iommu` annotation specifically, which controls whether the sandbox VM includes a vIOMMU.  A guest side vIOMMU is necessary to implement VFIO passthrough devices with `vfio_mode = vfio`, so enabling that mode of operation requires a global configuration change.

Unlike some of the other hypervisor annotations, the `enable_iommu` annotation is quite safe.  By default the vIOMMU is not present, so allowing a user to override it for a pod only improves their facilities for isolation.  Even if the global default were changed to enable the vIOMMU, that doesn't compel the guest kernel to use it, so allowing a user to disable the vIOMMU doesn't materially affect isolation either.

dgibson:
…default

Since #902 the `io.katacontainers.config.hypervisor` pod annotations
have only been permitted if explicitly allowed in the global
configuration.  The default global configuration allows no such
annotations.  That's important because several of those annotations
would cause Kata to execute arbitrary binaries, and so were wildly
unsafe.

However, this is inconvenient for the
`io.katacontainers.config.hypervisor.enable_iommu` annotation
specifically, which controls whether the sandbox VM includes a vIOMMU.
A guest side vIOMMU is necessary to implement VFIO passthrough devices
with `vfio_mode = vfio`, so enabling that mode of operation currently
requires a global configuration change, and can't just be enabled
per-pod.

Unlike some of the other hypervisor annotations, the `enable_iommu`
annotation is quite safe.  By default the vIOMMU is not present, so
allowing a user to override it for a pod only improves their
facilities for isolation.  Even if the global default were changed to
enable the vIOMMU, that doesn't compel the guest kernel to use it, so
allowing a user to disable the vIOMMU doesn't materially affect
isolation either.

Therefore, allow the io.katacontainers.config.hypervisor.enable_iommu
annotation to work in the default configurations.

fixes #4330

### DIFF
--- a/src/runtime/Makefile
+++ b/src/runtime/Makefile
@@ -160,7 +160,7 @@ DEFMEMSZ := 2048
 DEFMEMSLOTS := 10
 #Default number of bridges
 DEFBRIDGES := 1
-DEFENABLEANNOTATIONS := []
+DEFENABLEANNOTATIONS := [\"enable_iommu\"]
 DEFDISABLEGUESTSECCOMP := true
 DEFDISABLEGUESTEMPTYDIR := false
 #Default experimental features enabled


dgibson:
Making the case for a backport to 2.4.x, on the grounds that that introduced the working `vfio_mode=vfio` feature, which is difficult to use without this change.

snir911:
Is this supported by all kata's supported hypervisors? according to what you mentioned why not changing the global default instead of the annotation?

dgibson:
> Is enable_iommu available for all of kata's supported hypervisors? according to what you mentioned why not changing the global default instead of the annotation?

I'm also looking at changing the defaults in #4332. It's pretty plausible there are cases where we want to override this on a per-pod basis (in one direction or another), though, and since it's fairly safe to do so I think we should allow it.

In addition, we can't reasonably make such a visible behavior change within 2.4.x, but I think this change allowing the per-pod override could be backported to 2.4.x.

dgibson:
ppc64le failure appears to be a missing `virtiofsd` which I don't think is related to this change.

fidencio:
Not exactly related to this patch, but this is a bigger discussion that we should have.

What are the annotation values we want to expose by default, if any?  My understanding is that each downstream will have their own preference on this ...

Maybe this is something to be discussed at the next A/C meeting?

sboeuf:
> /cc @amshinde, @sboeuf, @rbradford.

See my comment https://github.com/kata-containers/kata-containers/pull/4333#issuecomment-1141293808

dgibson:
> > /cc @amshinde, @sboeuf, @rbradford.
> 
> See my comment [#4333 (comment)](https://github.com/kata-containers/kata-containers/pull/4333#issuecomment-1141293808)

I think your comments in #4333 are only tangentially relevant here.  Even if we don't want to change the defaults, I think that's *more* reason to allow overriding on a per-pod basis.

sboeuf:
> > > /cc @amshinde, @sboeuf, @rbradford.
> > 
> > 
> > See my comment [#4333 (comment)](https://github.com/kata-containers/kata-containers/pull/4333#issuecomment-1141293808)
> 
> I think your comments in #4333 are only tangentially relevant here. Even if we don't want to change the defaults, I think that's _more_ reason to allow overriding on a per-pod basis.

Oh yes sorry I missed the point of this PR. It's actually nice to allow overriding the settingon a per-pod basis.

dgibson:
> > > > /cc @amshinde, @sboeuf, @rbradford.
> > > 
> > > 
> > > See my comment [#4333 (comment)](https://github.com/kata-containers/kata-containers/pull/4333#issuecomment-1141293808)
> > 
> > 
> > I think your comments in #4333 are only tangentially relevant here. Even if we don't want to change the defaults, I think that's _more_ reason to allow overriding on a per-pod basis.
> 
> Oh yes sorry I missed the point of this PR. It's actually nice to allow overriding the setting on a per-pod basis.

Mind giving an ack then :)

sboeuf (approved):

amshinde (approved):

PR #4213, opened 2022-05-06
runtime: Adding the correct detection of mediated PCIe devices

zvonkok:
Fixes #4212

Signed-off-by: Zvonko Kaiser <zkaiser@nvidia.com>

PR #4337, opened 2022-05-30
stable-2.4: runtime: Adding the correct detection of mediated PCIe devices

### LINKED ISSUE #4212: runtime: Allow MDEV VFIO devices to be attached to a PCIe root port
The current implementation in Kata attaches mediated devices to a PCIe bus, for hotplugging we need to add a PCIe mediated device to a PCIe root port.

The issue is that the runtime will only attach a PCI device to a root port if it identifies it as a PCIe device. Currently, it does it by reading the size of the config space of a specific BDF. "If size > 256 then PCIe device else PCI device" and it is attached to the PCI bus and not a root port. 

The check is done by using the BDF of a PCIe device and since MDEV devices usually "only" have a sysfsdev entry (which is in most cases sufficient)  the check fails and does not recognize a MDEV device as PCIe device and hence it is attached to the PCI bus and not to a PCIe root port. 

The sysfsdev entry looks like the following: 
```
 /sys/devices/pci0000:xY/BDF0/BDF1/.../MDEVBDF/UUID
```
This is the complete hierarchy of PCIe busses, ports and devices attached along the path.

snir911:
Fixes #4212

Backport-of: https://github.com/kata-containers/kata-containers/pull/4213
Signed-off-by: Zvonko Kaiser <zkaiser@nvidia.com>

### DIFF
--- a/src/runtime/virtcontainers/device/drivers/utils.go
+++ b/src/runtime/virtcontainers/device/drivers/utils.go
@@ -45,9 +45,8 @@ func deviceLogger() *logrus.Entry {
 	return api.DeviceLogger()
 }
 
-// Identify PCIe device by /sys/bus/pci/slots/xx/max_bus_speed, sample content "8.0 GT/s PCIe"
-// The /sys/bus/pci/slots/xx/address contains bdf, sample content "0000:04:00"
-// bdf format: bus:slot.function
+// Identify PCIe device by reading the size of the PCI config space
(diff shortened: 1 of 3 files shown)


fidencio (approved):
lgtm, thanks @snir911!

BbolroC:
Please ignore the result for s390x, it is the cpu check error which passes on a IBM Cloud machine.

GabyCT (approved):

Issue #4349, opened 2022-05-31
test-kata-deploy workflow is running from main branch on stable branch


snir911:
# Description of problem

Commenting `/test_kata_deploy` on stable branch triggers the `test-kata-deploy` workflow which exists on the main branch instead of the workflow that exist on the tested branch (stable).

# Further information

The reason this is happening seems to be because of the environment variables that are passed when [issue_comment](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment) event is triggered

PR #4350, opened 2022-05-31
workflows: add workflow_dispatch triggering to test-kata-deploy

snir911:
This will allow to trigger the test-kata-deploy workflow manually from
any branch instead of using always the one that is defined on main

While here, let's also update the doc how to use it and with some github_token instructions
See: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

Depends-on: github.com/kata-containers/tests#4823 (to fix static checks)
Fixes: #4349

### DIFF
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -1,4 +1,5 @@
 on:
+  workflow_dispatch: # this is used to trigger the workflow on non-main branches
   issue_comment:
     types: [created, edited]
 
--- a/docs/Release-Process.md
+++ b/docs/Release-Process.md
@@ -48,7 +48,7 @@
 ### Merge all bump version Pull requests
 
   - The above step will create a GitHub pull request in the Kata projects. Trigger the CI using `/test` command on each bump Pull request.
-  - Trigger the test-kata-deploy workflow on the kata-containers repository bump Pull request using `/test_kata_deploy` (monitor under the "action" tab).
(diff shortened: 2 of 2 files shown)


egernst:
It seems this is for manually triggering (GUI) the workflow -- if you run via the comment trigger, we'll still have same issue, right?

snir911:
@egernst yes, I mean, it will pass if main's workflow is valid also for the stable branch but it's indeed incorrect.
I didn't remove it as it still works correctly for main, i planned to remove after the release making sure everything is fine (although i tested on private repo and it worked as expected)
/retest-metrics
/retest-arm

snir911:
added dependeson to fix static checks let's see if it helps

snir911:
It's still failing as we don't fetch the depends-on in this case :(

wainersm:
> @wainersm I'll add the comment I find this options a proper solution, AFAIU it's not possible to refer to a different branch other than the default branch with this event type ([issue_comment](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment)) (maybe we can workaround it by adding manually the branch name in the comment or something, but i don't find it better)

hmm... I glanced at the documentation and I didn't think a way to read the branch either. Let's live with that :)

egernst (approved):
suboptimal, but this gives us a workaround at least.   Thanks - sgtm

snir911:
This change was not enough, it requires to fix also the `check-comment-and-membership` job which i shamelessly missed, meanwhile i blocked (dnm) the release PRs
FYI @egernst @wainersm @fidencio

Issue #4111, opened 2022-04-19
CCv0: Refactor signature verification PoC changes to move into a test fixture


stevenhorsman:
### Description

In the `CCv0` branch, in order to test the image signature verification PoC, I added some temporary PoC files that we wanted to get into the guest image.
The artifacts are in https://github.com/kata-containers/kata-containers/tree/CCv0/tools/osbuilder/rootfs-builder/signed-container-artifacts
and the code that added them is:
https://github.com/kata-containers/kata-containers/blob/e4bae434dea76700a4adce144657fe060045b33b/tools/osbuilder/rootfs-builder/rootfs.sh#L643-L658

In https://github.com/kata-containers/tests/pull/4677/files Wainer has added a `cp_to_guest_img` method. I think this is what we need to add in test content without polluting the code base.

### Task list

- [ ] Move the image signature verification required artefacts from https://github.com/kata-containers/kata-containers/tree/CCv0/tools/osbuilder/rootfs-builder/signed-container-artifacts into a tests repo fixtures directory (either in https://github.com/kata-containers/tests/tree/CCv0/integration/containerd/confidential/fixtures, or even better in a shared confidential directory that is used by both Kubernetes and crictl)
- [ ] In the image signature verification test add in the logic that calls `cp_to_guest_img` to copy the artifacts into the correct place in the guest image, ensuring that all the files created in `rootfs.sh` temp PoC code are put into the correct place.
  - [ ] So we can manually do dev and test on this we should create a new function in `ccv0.sh` that does the same copy operation, either driven by an environment variable, or a new command and document this in the how-to markdown file.
- [ ] Ensure that all of the code in the dev repo relating to injecting artefacts (as linked in the description) is removed.
- [ ] Make sure that all the manual and automated tests are still working

PR #4354, opened 2022-05-31
 CCv0: Refactor signature verification PoC changes

GeorginaKin:
Removes signature verification files and related code

Fixes: #4111
Depends-on: github.com/kata-containers/tests#4821

Signed-off-by: Georgina Kinge <Georgina.Kinge@ibm.com>
Co-authored-by: Megan Wright <megan.wright@ibm.com>


GeorginaKin:
> We need a small change to update/doc the `ccv0.sh` user flow. We also need to remove the references to the now deleted README.md file in https://github.com/GeorginaKin/kata-containers/blob/CCv0/tools/osbuilder/rootfs-builder/README.md#confidential-containers-support

Hey @stevenhorsman I've added some changes in the [build and test ccv0](https://github.com/GeorginaKin/kata-containers/blob/CCv0/docs/how-to/how-to-build-and-test-ccv0.md#verifying-signed-images) document to cover running the `copy_files_to_guest` function so that manual signature testing still works. 
I've also removed the last reference to the README we deleted.

Hopefully this is what you were looking for? 😄

docs/how-to/how-to-build-and-test-ccv0.md
```diff
@@ -333,6 +333,9 @@ to create containers from these images using `crictl`:
 the guest agent code has been implemented, the Kata confidential-containers build needs to be run with 
 `export SKOPEO=yes` as documented [above](#basic-script-set-up-and-optional-environment-variables) in order to use the
 image signature verification function.
+
+- The policy file, along with the public GPG key and signatures file, will need to be copied over before the pod is created by using the `copy_files_to_guest` function in ccv0.sh. To do this, you should run `~/ccv0.sh copy_files_to_guest` before creating the sandbox pod in the steps below.
```
stevenhorsman, line 337: The preferred style for documentation is to keep the lines <=120 chars long in order to make it easier to see diff edits, so if you could add some line breaks in that would be great

tools/osbuilder/rootfs-builder/README.md
```diff
@@ -192,5 +192,4 @@ needed. Changes affect the files included in the final guest image.
 
 When building the rootfs for confidential containers if `SKOPEO=yes` is set then the `skopeo`
 package is built and added into the rootfs. It also adds the signature verification proof of concept files.
```
stevenhorsman, line 194: You've left in 'It also adds the signature verification proof of concept files.' and `rootfs.sh` doesn't do that anymore ?

stevenhorsman (changes requested):

tools/osbuilder/rootfs-builder/README.md
GeorginaKin: Ah yeah good point - have removed that now, but left the info about skopeo and umoci

docs/how-to/how-to-build-and-test-ccv0.md
```diff
@@ -333,6 +333,11 @@ to create containers from these images using `crictl`:
 the guest agent code has been implemented, the Kata confidential-containers build needs to be run with 
 `export SKOPEO=yes` as documented [above](#basic-script-set-up-and-optional-environment-variables) in order to use the
 image signature verification function.
+
+- The policy file, along with the public GPG key and signatures file, will need to be copied over before the pod is created 
+by using the `copy_files_to_guest` function in ccv0.sh. 
+To do this, you should run `~/ccv0.sh copy_files_to_guest` before creating the sandbox pod in the steps below.
```
stevenhorsman, line 339: Given that the next instruction is 'If you don't already have a Kata sandbox pod created with `crictl`, then follow the'..., do we need to update it to get the user to create the pod even if they had one previously, if the `copy_files_to_guest` command it needed before the pod is created?

GeorginaKin: I hadn't thought of that, thank you! I've added details on what to do if you **do** have a pod and what to if you **don't**.

stevenhorsman: So I did a manual test of running `~/ccv0.sh copy_files_to_guest` when I already had a running pod and then doing `~/ccv0.sh crictl_create_cc_pod` after and that worked fine, so I think we can simplify the instructions to something like:
- Leave the old instruction about creating a crictl pod linking above
- Add a new instruction to say the user needs to use copy_files_to_guest to enable the guest image and then re-run `~/ccv0.sh crictl_create_cc_pod` to create a pod from that image
How does that sound?

GeorginaKin: That's a good point actually because the create pod function deletes the existing one so we don't need to run the manual delete.
I've updated the instructions to make them simpler as suggested 🙂

stevenhorsman:
`do-not-merge` added to this PR as it wants to be merged with/after https://github.com/kata-containers/tests/pull/4821

docs/how-to/how-to-build-and-test-ccv0.md
```diff
@@ -333,9 +333,14 @@ to create containers from these images using `crictl`:
 the guest agent code has been implemented, the Kata confidential-containers build needs to be run with 
 `export SKOPEO=yes` as documented [above](#basic-script-set-up-and-optional-environment-variables) in order to use the
 image signature verification function.
+
 - If you don't already have a Kata sandbox pod created with `crictl`, then follow the 
 [instructions above](#using-crictl-for-end-to-end-provisioning-of-a-kata-confidential-containers-pod-with-an-unencrypted-image)
 up to, and including, the `~/ccv0.sh crictl_create_cc_pod` command.
+
+- In order to enable the guest image, you will need to copy over the signature files needed by running
```
wainersm, line 341: You might want to mention here that the `quay_policy.json` file is also copied, as in a early phrase it is mentioned a policy file is passed to the agent.

stevenhorsman (approved):

Issue #4359, opened 2022-06-01
Another shim leak and questions on the exit error handing 


fengwang666:
### Bug Description ### 

When a container process exits, its wait goroutine invokes [sandbox.StopContainer](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L81), which then calls the [stop()](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/container.go#L966) method in `container.go`. In the error case we hit, the `stop()` method exits early due to an error returned from the kata agent when calling [stopContainer()](https://github.com/databricks/kata-containers/blob/9049409bc3ef9a594c2a3a750550ad58e58ef826/src/runtime/virtcontainers/container.go#L1006-L1008) method, which results in a shim leak. 

Because of the early exit, the `container state` is [not set](https://github.com/databricks/kata-containers/blob/9049409bc3ef9a594c2a3a750550ad58e58ef826/src/runtime/virtcontainers/container.go#L1035-L1037) to `StateStopped`. The `wait goroutine` [logs a warning](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L82) on the error from `StopContainer` and continues its execution to its end, which also [set the container task status](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L85) to `StatusStopped`. In this case the [KillRequest](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L740) from containerd cri server is fulfilled completely. 

The failure happens in the [DeleteRequest](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L469). The `Delete()` method fails on [deleteContainer](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L490) because the `container state` is [neither StateReady nor StateStopped](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/container.go#L897-L900). 
The relevant error log we see:
```
[...]

PR #4360, opened 2022-06-01
runtime: ignore ESRCH error from stop container

fengwang666:
Kata agent can return ESRCH to the agent.stopContainer call,
which causes shim leak.

Fixes: #4359

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -1063,7 +1063,19 @@ impl BaseContainer for LinuxContainer {
         let st = self.oci_state()?;
 
         for pid in self.processes.keys() {
-            signal::kill(Pid::from_raw(*pid), Some(Signal::SIGKILL))?;
+            match signal::kill(Pid::from_raw(*pid), Some(Signal::SIGKILL)) {
+                Err(Errno::ESRCH) => {
+                    info!(
+                        self.logger,
+                        "kill encounters ESRCH, pid: {}, container: {}",
+                        pid,
+                        self.id.clone()
+                    );
+                    continue;
+                }
+                Err(err) => return Err(anyhow!(err)),
+                Ok(_) => continue,
+            }
         }
 
         if spec.hooks.is_some() {
--- a/src/runtime/pkg/containerd-shim-v2/wait.go
+++ b/src/runtime/pkg/containerd-shim-v2/wait.go
@@ -78,7 +78,7 @@ func wait(ctx context.Context, s *service, c *container, execID string) (int32, 
 				shimLog.WithField("sandbox", s.sandbox.ID()).Error("failed to delete sandbox")
 			}
 		} else {
-			if _, err = s.sandbox.StopContainer(ctx, c.id, false); err != nil {
+			if _, err = s.sandbox.StopContainer(ctx, c.id, true); err != nil {
 				shimLog.WithError(err).WithField("container", c.id).Warn("stop container failed")
 			}
 		}


bergwolf (approved):
lgtm! Besides these changes, could you also change `wait.go` that when a container process exits, let's [force stop the container](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L81). It should help avoiding the inconsistency between the two should-otherwise-be-merged container status in shimv2 and virtcontainers.

fengwang666:
> lgtm! Besides these changes, could you also change wait.go that when a container process exits, let's [force stop the container](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L81). It should help avoiding the inconsistency between the two should-otherwise-be-merged container status in shimv2 and virtcontainers.

@bergwolf My current implementation will first gracefully stop the container in [the wait.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L81), only [force stop](https://github.com/kata-containers/kata-containers/blob/d6b9080995b9d739b2939cf1d46891be53a4a4a4/src/runtime/pkg/containerd-shim-v2/delete.go#L25-L27) the container in the delete.go if the container was not stopped cleanly. If we force stop the container in the wait.go, then any error in the [cleanup steps](https://github.com/kata-containers/kata-containers/blob/ce2e521a0f5f08698429daeacdc852b6a12f1679/src/runtime/virtcontainers/container.go#L1005-L1029) is ignored. Is that desirable?

egernst:
I guess the question really is if there's a scenario where we don't want the container to (forcibly) be stopped in after we see completion in wait.go? Are there any paths that wouldn't just lead to the force stop being called in container delete?

Thanks for doing such a great job digging into this @fengwang666

fengwang666:
> I guess the question really is if there's a scenario where we don't want the container to (forcibly) be stopped in after we see completion in wait.go? Are there any paths that wouldn't just lead to the force stop being called in container delete?

@egernst Based on my code tracing in containerd, a TaskExit event always triggers `DeleteRequest`, which go through the container delete code path. I'm not sure whether that's the case for crio. But since [deleteContainer](https://github.com/kata-containers/kata-containers/blob/4f7cc186224fa9e724dc503f98dbebbb151bd2c8/src/runtime/pkg/containerd-shim-v2/delete.go#L17) is only called by [service.Delete()](https://github.com/kata-containers/kata-containers/blob/71ce6f537fd6565c07a469cc9f34a97ee0859919/src/runtime/pkg/containerd-shim-v2/service.go#L490), it seems safe to [force stop the container](https://github.com/fengwang666/kata-containers/blob/1df891924bcf897056b0a75898bc246264ac4497/src/runtime/pkg/containerd-shim-v2/delete.go#L25-L27) during the delete container operation if the [previous stop container](https://github.com/fengwang666/kata-containers/blob/1df891924bcf897056b0a75898bc246264ac4497/src/runtime/pkg/containerd-shim-v2/wait.go#L81-L83) in the wait.go failed.

bergwolf:
> My current implementation will first gracefully stop the container in [the wait.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L81), only [force stop](https://github.com/kata-containers/kata-containers/blob/d6b9080995b9d739b2939cf1d46891be53a4a4a4/src/runtime/pkg/containerd-shim-v2/delete.go#L25-L27) the container in the delete.go if the container was not stopped cleanly. If we force stop the container in the wait.go, then any error in the [cleanup steps](https://github.com/kata-containers/kata-containers/blob/ce2e521a0f5f08698429daeacdc852b6a12f1679/src/runtime/virtcontainers/container.go#L1005-L1029) is ignored. Is that desirable?

@fengwang666 My concern is that w/o force stop in wait.go, we are leaving vc container status and shimv2 container status in an inconsistent status. Force stop is a way for us to handle stop when we cannot report stop failures to upper layers. wait.go seems to fall into that category IMO.

egernst:
I agree with @bergwolf - I think keeping it consistent would be good, and it seems there shouldn't be downside with forcing at that point, and not relying on the containerd side.  If we should remove the container, let's ensure we do it ?

fengwang666:
@bergwolf @egernst I have updated the commit to force stop the container in the wait.go. I also revert the change in the delete.go as that's not necessary any more. Also worth noting, after this change, I don't see there is any place that actually pass `false` to the force lag any more (except for the[ delete.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/delete.go#L20), which is a dead codepath). We probably can clean up the code to remove the flag. wdyt?

egernst (approved):

PR #4395, opened 2022-06-06
agent: rollback mounted storage if err met

windniw:


snir911:
Thanks! LGTM, currently it fails due to rust format check:
```
cargo fmt -- --check
Diff in /home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/agent/src/mount.rs at line 15:
 use tokio::sync::Mutex;
-use nix::mount::{MsFlags, umount};
+use nix::mount::{umount, MsFlags};
 use nix::unistd::{Gid, Uid};
 use regex::Regex;
Diff in /home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/agent/src/mount.rs at line 7[88](https://github.com/kata-containers/kata-containers/runs/6771289058?check_suite_focus=true#step:12:89):
                 for m in mount_list {
                     let _ = umount(m.as_str());
                 }
-                return Err(anyhow!("handle storage {} failed case {:?}", handler_name, e))
+                    "handle storage {} failed case {:?}",
+                    handler_name,
+                    e
+                ));
             }
         }
     }
```
While fixing that please also create an issue and follow the [patch-format](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format) requirements to make sure it will pass all the static checks.

src/agent/src/mount.rs
```diff
@@ -15,7 +15,7 @@ use std::sync::Arc;
 
 use tokio::sync::Mutex;
```
c3d, line 16: @windniw Could you expand on the commit text, make sure there is at least one line of description 😄 OK to repeat the issue's description.

src/agent/src/mount.rs
```diff
@@ -778,11 +778,18 @@ pub async fn add_storages(
...
-            mount_list.push(mount_point);
+        match res {
+            Ok(mount_point) => {
+                if !mount_point.is_empty() {
+                    mount_list.push(mount_point);
+                }
+            }
+            Err(e) => {
+                for m in mount_list {
+                    let _ = umount(m.as_str());
```
c3d, line 789: This may error out, probably a good idea to log if that happens.

src/agent/src/mount.rs
```diff
@@ -778,11 +778,18 @@ pub async fn add_storages(
...
-        if !mount_point.is_empty() {
-            mount_list.push(mount_point);
+        match res {
+            Ok(mount_point) => {
+                if !mount_point.is_empty() {
+                    mount_list.push(mount_point);
+                }
+            }
+            Err(e) => {
+                for m in mount_list {
```
gkurz, line 788: umount() may fail if mount points overlap, resulting in a silent mount point leak. Rollback should always undo things in reverse order to avoid this.

src/agent/src/mount.rs
gkurz: Please also fix your Signed-off-by: to display a real name instead of the `windniw` pseudonym, as required by the DCO process followed by kata (go to the `Developer's Certificate of Origin 1.1` section in https://www.kernel.org/doc/Documentation/process/submitting-patches.rst).

src/agent/src/mount.rs
gkurz, line 789: It seems fine to go on with the rollback if umount() fails but the failure should be logged anyway.

gkurz:
ping @windniw . Please address the several points that were raised:
- create an issue
- provide details and fix Signed-off-by: in the commit changelog
- reverse the umount loop
- print an warning/error if umount fails
- fix format before pushing to GH

Thanks !

fidencio:
@windniw, we're closing this PR as it's been opened and inactive for more than one year.

Issue #4376, opened 2022-06-01
When a Kata sandbox first starts I always get one or two "failed to get OOM event from sandbox"


skaegi:
In my Kubernetes installation every time a new kata micro-vm  starts I see one or two "failed to get OOM event from sandbox" messages in my syslog. This seems to sort itself out after 10 seconds or so but in my clusters where I run millions of containers a month I see a ton of these messages in my syslog over time.

I recognize this is by no means fatal but is there something we can do here?

liubin:
Usually, this log will be printed when a container is to be deleted, when you start a  new kata pod, is there an old one to be deleted?

There is a goroutine that watches the OOM event for containers, when the container exits, the goroutine will exit too, so this log is a warning log but not fatal errors.

If there will be only 1 or 2 lines of this for one pod/container, that will not be an issue.

For now, this is a bit not so easy to check the `getOOMEvent` all (from runtime to agent inside guest OS) failed because all containers stopped or an unexpected issue.

skaegi:
Thanks @liubin. No we are seeing this for brand new Pod/Containers when first created with no pre-existing container to delete. The pods do use a number of short-lived init containers though so perhaps that's related as the containers complete.

Everything else you say makes sense and it is indeed a warning... just at our scale the messages add up.

liubin:
@skaegi I think this log(GetOOMEvent call failing) is not a critical log, so I changed it from `warn` to `info` in #4425 , will this help you to reduce the noise logs?

Issue #2478, opened 2021-08-21
Failed to get OOM event from sandbox is a horribly chatty red herring, and should be info, not warn


egernst:
AFAICT, we will hit this, [1], for most every sandbox when it is shutting down. If you run a lot of Kata pods, you'll see a log of these messages when there isn't an OOM event to be received.

We should probably downgrade this from a warning to info, and should consider adding smarter error handling here to identify when the agent is not responding? 

[1]- https://github.com/kata-containers/kata-containers/blob/3e8137399c6dc1ad2af7a236c281a2ee60a817a7/src/runtime/containerd-shim-v2/wait.go#L150

egernst:
Looks like this was added to help address: https://github.com/kata-containers/kata-containers/commit/cb0e6094ff93a2720ced71bf569c6b170b81e16c

But, really we shouldn't be outputting this as a warning, then, since shutdown is a very common scenario shared by all.

@liubin  - WDYT?

liubin:
I have ever considered to check the sandbox's state, for example if in the `shuttingdown` state then break the for loop.
Or using a chan to receive sandbox shutdown event to break the for loop.

sameo:
I think we should only print out an error if the agent is not shutting down. When it is shutting down and destroying the sandbox, the OOM event receiver will be closed and we will hit that path: https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L1197. This is where we could add a custom ttrpc error string/code and handle that nicely from the runtime.
But meanwhile, I believe we should at least downgrade to info.

ariel-adam:
@snir911 also relevant to your work

liubin:
closed via #4425

PR #4425, opened 2022-06-08
shim: change the log level for GetOOMEvent call failures

liubin:
GetOOMEvent is a blocking call that will fail if
the container exit, in this case, it's not an error or warning.

Changing the log level for logs in case of GetOOMEvent call fails
will reduce log noise in a large cluster that has pods
creating/deleting frequently.

Fixes: #4376


egernst (approved):

cmaf (approved):
thanks @liubin 

Also related but w.r.t. shutting down: https://github.com/kata-containers/kata-containers/issues/2478

amshinde (approved):

fidencio (approved):
Thanks @liubin!

Issue #4420, opened 2022-06-08
Support shim v2 logging plugin


liubin:
Now kata shim v2 only accepts stdout/stderr as pipe file from upper container engine like containerd, it will not work if the path is `binary:///home/kant/nerdctl?_NERDCTL_INTERNAL_LOGGING=%2Fvar%2Flib%2Fnerdctl%2F1935db59`.

reference: 
- https://github.com/containerd/nerdctl/pull/959/
- https://github.com/containerd/containerd/pull/3085
- #4391

PR #4432, opened 2022-06-09
shim: support shim v2 logging plugin

liubin:
Now kata shim only supports stdout/stderr of fifo from
containerd/CRI-O, but shim v2 supports logging plugins,
and nerdctl default will use the binary schema for logs.

This commit will add the others type of log plugins:

- file
- binary

In case of binary, kata shim will receive a stdout/stderr like:

binary:///nerdctl?_NERDCTL_INTERNAL_LOGGING=/var/lib/nerdctl/1935db59

That means the nerdctl process will handle the logs(stdout/stderr)

Fixes: #4420

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/service.go
+++ b/src/runtime/pkg/containerd-shim-v2/service.go
@@ -7,6 +7,7 @@ package containerdshim
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	sysexec "os/exec"
@@ -83,6 +84,11 @@ func New(ctx context.Context, id string, publisher cdshim.Publisher, shutdown fu
 	vci.SetLogger(ctx, shimLog)
 	katautils.SetLogger(ctx, shimLog, shimLog.Logger.Level)
 
+	ns, found := namespaces.Namespace(ctx)
+	if !found {
+		return nil, fmt.Errorf("shim namespace cannot be empty")
+	}
+
 	s := &service{
 		id:         id,
 		pid:        uint32(os.Getpid()),
@@ -91,6 +97,7 @@ func New(ctx context.Context, id string, publisher cdshim.Publisher, shutdown fu
 		events:     make(chan interface{}, chSize),
 		ec:         make(chan exit, bufferSize),
 		cancel:     shutdown,
+		namespace:  ns,
 	}
 
 	go s.processExits()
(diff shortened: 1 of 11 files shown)


src/runtime/pkg/containerd-shim-v2/service.go
```diff
@@ -129,6 +136,9 @@ type service struct {
 
 	id string
 
+	// Namespace from upper container engine
+	namespace string
```
amshinde, line 140: @liubin Can you add comments here what this namespace is and why its required.

src/runtime/pkg/containerd-shim-v2/stream.go
```diff
@@ -27,107 +33,110 @@ var (
...
-	cf(tty.Stdout)
-	cf(tty.Stderr)
+	tty.io.Close()
 }
 
-func newTtyIO(ctx context.Context, stdin, stdout, stderr string, console bool) (*ttyIO, error) {
-	var in io.ReadCloser
-	var outw io.Writer
-	var errw io.Writer
+func newTtyIO(ctx context.Context, ns, id, stdin, stdout, stderr string, console bool) (*ttyIO, error) {
```
amshinde, line 58: Its not super clear what the new introduced id here is meant to be or used for. Can you name this so that it is clear and add some comments?

liubin: @amshinde I'll add some comments to describe why ns and id parameters are required.

src/runtime/pkg/containerd-shim-v2/shim_io_binary.go
```diff
@@ -0,0 +1,215 @@
...
+	closers = append(closers, serr.Close)
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		return nil, err
+	}
+	closers = append(closers, r.Close, w.Close)
+
+	cmd := newBinaryCmd(uri, id, ns)
+	cmd.ExtraFiles = append(cmd.ExtraFiles, out.r, serr.r, w)
```
lifupan, line 68: Can you give a detail comments on how the binary used those pipe files, it seemed it's a standard protocol and it's better document or commented it here.

lifupan:
It's LGTM, but if we can add some detail comments on how does the binary use stdout and stderr pipe files, it would be much more clear for us.

Tim-Zhang (approved):

PR #4436, opened 2022-06-10
runtime: fix error when trying to parse sandbox sizing annotations

### LINKED ISSUE #4435: memory annotation parse error
# Description of problem

kata tries to parse the sizing information provided as annotations by the container runtime. If the memory annotation value is bigger than 2GiB parsing fails and the container is created with default memory size.

alex-matei:
Changed bitsize for parsing functions to 64-bit in order to avoid
parsing errors.

Fixes #4435


amshinde (commented):
LGTM. Thanks @alex-matei! Can you verify this by adding to the test case here: https://github.com/kata-containers/kata-containers/blob/fa61bd43ee4709fea08bbf04335b829be3bfef28/src/runtime/pkg/oci/utils_test.go#L1156

alex-matei:
@amshinde Thanks for reply. I added a test case.

amshinde (approved):

liubin (approved):

PR #4467, opened 2022-06-16
device package cleanup/refactor

egernst:
Today, device package imports persist API, simply so we can have the persist API structures associated with the device's Load/Save.

I don't think device package should rely on persist; rather I'd like it reversed so we can deduplicate fields, have a cleaner device package.

With this PR, I:
 - move device package from ./virtcontainers/device to ./pkg/device
 - remove dependency on ./virtcontainers/persist from ./pkg/device
 - deduplicate device structures from persist and what's in devices, and simplify the device's Load/Save functions


src/runtime/pkg/device/drivers/block.go
```diff
@@ -159,13 +158,13 @@ func (device *BlockDevice) GetDeviceInfo() interface{} {
...
 // Save converts Device to DeviceState
-func (device *BlockDevice) Save() persistapi.DeviceState {
+func (device *BlockDevice) Save() config.DeviceState {
 	ds := device.GenericDevice.Save()
 	ds.Type = string(device.DeviceType())
 
 	drive := device.BlockDrive
 	if drive != nil {
-		ds.BlockDrive = &persistapi.BlockDrive{
+		ds.BlockDrive = &config.BlockDriveState{
```
mcastelino, line 167: So we have unified the Configuration and State to be one? Did we always have `BlockDriveState`

src/runtime/pkg/device/drivers/vfio.go
```diff
@@ -174,26 +173,21 @@ func (device *VFIODevice) GetDeviceInfo() interface{} {
...
 	devs := device.VfioDevs
 	for _, dev := range devs {
 		if dev != nil {
-			ds.VFIODevs = append(ds.VFIODevs, &persistapi.VFIODev{
-				ID:       dev.ID,
-				Type:     uint32(dev.Type),
-				BDF:      dev.BDF,
-				SysfsDev: dev.SysfsDev,
-			})
+			ds.VFIODevs = append(ds.VFIODevs, dev)
```
mcastelino, line 183: So there is no more persist configuration and actual running configuration. This is nice.

src/runtime/pkg/device/config/config.go
```diff
@@ -441,3 +441,45 @@ func getVhostUserDevName(dirname string, majorNum, minorNum uint32) (string, err
 	return "", fmt.Errorf("Required device node (%d:%d) doesn't exist under directory %s",
 		majorNum, minorNum, dirname)
 }
+
+// DeviceState is a structure which represents host devices
+// plugged to a hypervisor, one Device can be shared among containers in POD
+// Refs: virtcontainers/device/drivers/generic.go:GenericDevice
+type DeviceState struct {
```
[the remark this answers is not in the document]
egernst: b/c it was exactly that name when this was all in persist :shrug:  

Give me a better name and I can sed  -- but, for now, was just seeking to keep changes as minimal as possible.

src/runtime/pkg/device/drivers/block.go
egernst: It was called BlockDrive when it was unique to the persist API package.  I needed to rename to avoid conflict when pulling it into device/drivers package. The third commit ends up reconciling this.

mcastelino (approved):

liubin (approved):

liubin:
@egernst You need format your code and push again.

PR #4469, opened 2022-06-16
Config validation refactor

egernst:


src/runtime/virtcontainers/vm.go
```diff
@@ -42,9 +43,36 @@ type VMConfig struct {
 	HypervisorConfig HypervisorConfig
 }
 
-// Valid Check VMConfig validity.
 func (c *VMConfig) Valid() error {
-	return c.HypervisorConfig.Valid()
+
```
mcastelino, line 47: Why do we have two?

```
	if err := validateHypervisorConfig(&sandboxConfig.HypervisorConfig); err != nil {
 		return nil, err
 	}
```
And this?

mcastelino (approved):

zvonkok:
Indeed, this bugged me as well, when I was adding a new config variable for testing.

egernst:
I think the ppc executor is offline. If someone else on AC agrees, I'd be happy if we disabled the "required" option for this job and merge the PR.

Issue #4475, opened 2022-06-17
vCPUs pinning support for kata containers, related to issue 3726 


LitFlwr0:
**Is your feature request related to a problem? Please describe.**

Provided vCPUs pinning support for Kata Containers. We can pin each vCPU thread to one CPU when vCPU threads number are equal to that of CPUs required in CPUSet, to reduce extra cost for vCPU threads scheduling. 

**Describe the solution you'd like**

Use Linux's setting threads affinity api. Can acquire vCPU threads and CPUSet info from already-built Kata apis.

**Additional context**

Related to issue [https://github.com/kata-containers/kata-containers/issues/3726](url). Continued work from [https://github.com/fengwang666/kata-containers/commit/d373e2a775b7ea0c9cefe98e825112d9f55afd43](url).

PR #4476, opened 2022-06-17
vCPUs pinning support for Kata Containers

LitFlwr0:
VCPUs pinning support for Kata Containers.
Used Linux's setting thread scheduling affinity api.  Sandbox level configuration. Can use this feature through toml configuration file or annotations. 
Fixes issue 4775: [https://github.com/kata-containers/kata-containers/issues/4475](url)


src/runtime/pkg/oci/utils.go
```diff
@@ -635,6 +635,12 @@ func addHypervisorCPUOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig) e
 		return err
 	}
 
+	if err := newAnnotationConfiguration(ocispec, vcAnnotations.EnableVCPUsPinning).setBool(func(EnableVCPUsPinning bool) {
```
liubin, line 638: Use a camel style name:

`EnableVCPUsPinning` -> `enableVCPUsPinning`

egernst:
Thanks for the pull request @LitFlwr0 , and glad to have you involved with `soc`! 

This is a big change in behavior, AFAIU.  Can you add better description of the change itself, and how vCPU are treated, versus how VMM/IO-threads/contaienrd-shim-kata-v2 are handled, with respect to CPUSets cgroups?

fengwang666:
@LitFlwr0 Thanks for picking this up! Just to echo what @egernst commented, this is a hairy topic and there are various edge cases we need to consider. For example, we need to consider the case when there are multiple Kata Containers Pods on the same node.

LitFlwr0:
> Thanks for the pull request @LitFlwr0 , and glad to have you involved with `soc`!
> 
> This is a big change in behavior, AFAIU. Can you add better description of the change itself, and how vCPU are treated, versus how VMM/IO-threads/contaienrd-shim-kata-v2 are handled, with respect to CPUSets cgroups?

Hi, I applied linux sched_setAffinity api to vCPU threads. If the number of vCPU threads acquired by s.hypervisor.GetThreadIDs(), and the number of CPUs acquired by method s.getSandboxCPUSet() are equal, each vCPU thread is set to be scheduled to only one CPU with that api, to reduce cost of original random scheduling. 

This happens when adding, starting, updating and deleting a container, because under these conditions the CPU resources would change. Also, when the two numbers are not equal, vCPU threads are set to be scheduled to any CPU in CPUSet.

Other threads are not touched.

bergwolf:
@LitFlwr0 Thanks for working on it! I tend to agree with @egernst and @fengwang666 that this is a bit complex change in vCPU handling bevihavors. It would make sense to start with a design doc, get everyone on the same page, and then move on to implementation. WDYT? /cc @liubin

LitFlwr0:
Hi all, my design doc is here "https://docs.google.com/document/d/1_Zy9jRR3n6p408oDT1deVQzrjptf6HmU4S3YyEGlG-w/edit?usp=sharing". If you have any question or design suggestion just comment.

liubin:
@LitFlwr0 Design doc should be committed into the repo under `docs/design` and any image should also provide the source.

LitFlwr0:
Hi, pushed my code again with design doc under docs/design

sboeuf:
Just a note that for Cloud Hypervisor, the vCPU pinning should be very straightforward since the feature is directly baked in. Please refer to the documentation https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/cpu.md#affinity

liubin:
@LitFlwr0 you need to rebase this PR.
@lifupan PTAL

liubin:
Releated to https://github.com/kata-containers/kata-containers/pull/4797

lifupan:
Hi  @egernst  @fengwang666  @bergwolf 

Just as this patch said, it mainly focus on the case which the users would like pining the kata pod on a special cpu sets, thus it's better to pin a vcpu on one cpu one by one to reduce the scheduling noise.  

It seemed this commit doesn't involve the match between vcpu and cpu in guest, since not all of the hypervisor support it, and 
for the case of multi kata pods shared cpus on the node, I think it's the pod scheduler's job to make it not happen,  and the runtime should only follow the instructions from it. 

Hi @LitFlwr0 

Can you rebase your commit to fix the conflicting?

docs/design/vcpu-threads-pinning.md
```diff
@@ -0,0 +1,37 @@
...
+
+| Possible scenes                   | Related Code                               |
+|-----------------------------------|--------------------------------------------|
+| when creating a container         | File Sandox.go, in method CreateContainer  |
+| when starting a container         | File Sandox.go, in method StartContainer   |
+| when deleting a container         | File Sandox.go, in method DeleteContainer  |
+| when updating a container         | File Sandox.go, in method UpdateContainer  |
+| when creating multiple containers | File Sandox.go, in method createContainers |
+
+### Core Pinning Logics
```
bergwolf, line 34: Please describe the core logic here, explicitly I'm missing:
1. How are the default vCPUs handled?
2. cpuset only comes from a container spec. How do we handle multiple containers with different cpusets?
3. How does this work from Kubernetes e2e point of view?

LitFlwr0: Hi,

LitFlwr0: Hi,

LitFlwr0:
Hi, please do not delete this PR, I'm reopen it later

docs/design/vcpu-threads-pinning.md
LitFlwr0: Hi @bergwolf 
1. I didn't take default vCPU logics into consideration in this case. I only checks if the number of vCPU threads and the number of CPU requested are equal, and I need the number of default vCPU threads in this case, by using method "s.hypervisor.GetThreadIDs(ctx)". I think this method has taken default vCPU logics into consideration.
2. CPUSets are acquired by using method "s.getSandboxCPUSet()". In this method the sandbox goes through CPUSet of each container and returns a final overall CPUSet.
3. Two ways are provided to use this feature. By default the switch of this feature is passed by line "enable_vcpus_pinning = ture" in file "configuration-qemu.toml.in". Also this switch can be passed by annotation configuration, in file "oci/utils.go" "vcAnnotations.EnableVCPUsPinning".

LitFlwr0: Hi if you have any other questions just let me know and I'll fix them and make some changes to the code.

liubin:
@LitFlwr0 please fix your commit message format and delete the merge commit.

LitFlwr0:
> @LitFlwr0 please fix your commit message format and delete the merge commit.

Hi fixed the commit message

LitFlwr0:
> @LitFlwr0 please fix your commit message format and delete the merge commit.

Hi, adjusted commit messages. If there're other problems just let me know and I'll fix them.

liubin (approved):

Issue #4478, opened 2022-06-17
get handled signal include SigBlk, SigIgn and SigCgt


quanweiZhou:
At present, to stop the container will send the SIGTERM and then the SIGKILL. Currently, for the scenario where SIGTERM is not processed by the container, it will be converted to SIGKILL. The `is_signal_handled` function determine which signal the container process has been handled. But currently `is_signal_handled` only handles catching (SigCgt). 
When the container process is ignoring (SigIgn), blocking (SigBlk) also should not be converted to SIGKILL, should use SIGTERM.

gkurz:
> The `is_signal_handled` function in the agent/src/rpc.rs should include handled signal SigBlk SigIgn and SigCgt.

Hi @quanweiZhou !

The semantics of this function as described in the comment header are:
```
// Check if the container process installed the
// handler for specific signal.
```
which maps exactly to the semantics of `SigCgt`.

'SigIgn' and 'SigBlk' have really different semantics. It is thus hard to see what you're trying to do here.

You'd need to put more context. Why is this needed ? Does it fix an actual bug ?

quanweiZhou:
> > The `is_signal_handled` function in the agent/src/rpc.rs should include handled signal SigBlk SigIgn and SigCgt.
> 
> Hi @quanweiZhou !
> 
> The semantics of this function as described in the comment header are:
> 
> ```
> // Check if the container process installed the
> // handler for specific signal.
> ```
> 
> which maps exactly to the semantics of `SigCgt`.
> 
> `SigIgn` and `SigBlk` have really different semantics. It is thus hard to see what you're trying to do here.
> 
> You'd need to put more context. Why is this needed ? Does it fix an actual bug ?

At present, to stop the container will send the SIGTERM and then the SIGKILL. Currently, for the scenario where SIGTERM is not processed by the container, it will be converted to SIGKILL. The `is_signal_handled` function determine which signal the container process has been handled. But currently `is_signal_handled` only handles catching (SigCgt).
When the container process is ignoring (SigIgn) or blocking (SigBlk) also should not be converted to SIGKILL, should use SIGTERM. For example, when use terminationGracePeriodSeconds the k8s will send SIGTERM first and then send SIGKILL , in this case the container ignore the SIGTERM, so we should send the SIGTERM not the SIGKILL to container.

PR #4479, opened 2022-06-17
agent: enhance get handled signal

quanweiZhou:
Get handled signal include SigBlk, SigIgn and SigCgt.

Fixes: #4478
Signed-off-by: quanweiZhou <quanweiZhou@linux.alibaba.com>

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -1793,34 +1793,25 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
     let sig_mask: u64 = 1 << shift_count;
     let reader = BufReader::new(file);
 
-    // Read the file line by line using the lines() iterator from std::io::BufRead.
-    for (_index, line) in reader.lines().enumerate() {
-        let line = match line {
-            Ok(l) => l,
-            Err(_) => {
-                warn!(sl!(), "failed to read file {}", proc_status_file);
-                return false;
-            }
-        };
-        if line.starts_with("SigCgt:") {
+    // read lines start with SigBlk/SigIgn/SigCgt and check any match the signal mask
+    reader
+        .lines()
+        .flatten()
+        .filter(|line| {
+            line.starts_with("SigBlk:")
+                || line.starts_with("SigIgn:")
+                || line.starts_with("SigCgt:")
+        })
+        .any(|line| {
             let mask_vec: Vec<&str> = line.split(':').collect();
-            if mask_vec.len() != 2 {
-                warn!(sl!(), "parse the SigCgt field failed");
-                return false;
-            }
-            let sig_cgt_str = mask_vec[1].trim();
-            let sig_cgt_mask = match u64::from_str_radix(sig_cgt_str, 16) {
-                Ok(h) => h,
-                Err(_) => {
-                    warn!(sl!(), "failed to parse the str {} to hex", sig_cgt_str);
-                    return false;
+            if mask_vec.len() == 2 {
+                let sig_str = mask_vec[1].trim();
+                if let Ok(sig) = u64::from_str_radix(sig_str, 16) {
+                    return sig & sig_mask == sig_mask;
                 }
-            };
-
-            return (sig_cgt_mask & sig_mask) == sig_mask;
-        }
-    }
-    false
+            }
+            false
+        })
 }
 
 fn do_mem_hotplug_by_probe(addrs: &[u64]) -> Result<()> {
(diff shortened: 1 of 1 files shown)


quanweiZhou:
> > Get handled signal include SigBlk, SigIgn and SigCgt.
> > Fixes: #4478
> > Signed-off-by: quanweiZhou [quanweiZhou@linux.alibaba.com](mailto:quanweiZhou@linux.alibaba.com)
> 
> @quanweiZhou Please provide details in the issue on what you're trying to achieve with this patch.

Thanks @gkurz update the comment

c3d:
> Thanks @gkurz, the comment has been update.

@quanweiZhou I do not see the updated comment in the new commit message.

Essentially, I think that the commit message needs to explain, at lesat to the level described in issue #4478 : 
* Why is it a problem to turn `SIGTERM` into `SIGKILL` (I think, based on [this comment](https://github.com/kata-containers/kata-containers/issues/4478#issuecomment-1165313855) that your rationale is that we should leave that policy to Kubernetes, which does it on its own)
* Why we should send a signal to a process that ignores it. Blocking is often temporary, so it could make sense to not immediately convert to a `SIGKILL`, if if the process is ignored, you need to make it clear that you send the process to be ignored only because you expect Kubernetes to later request a `SIGKILL`.
* What about the non-Kubernetes cases? To what extent do we rely on a policy of the orchestration layer that is not necessarily guaranteed?

All this makes me wonder if the simplest would not be to simply pass the signals we receive as is. Does it make any sense to turn `SIGTERM` into `SIGKILL`? If so, why does it make sense for Kata and not for `runc` or `crun`?

src/agent/src/rpc.rs
```diff
@@ -1802,7 +1802,10 @@ fn is_signal_handled(proc_status_file: &str, signum: u32) -> bool {
...
             }
         };
-        if line.starts_with("SigCgt:") {
+
+        // check SigBlk, SigIgn, SigCgt
+        if line.starts_with("SigBlk:") || line.starts_with("SigIgn:") || line.starts_with("SigCgt:")
+        {
             let mask_vec: Vec<&str> = line.split(':').collect();
             if mask_vec.len() != 2 {
                 warn!(sl!(), "parse the SigCgt field failed");
```
Tim-Zhang, line 1811: It seems that the loop should continue to try the next sigmask

quanweiZhou:
> > Thanks @gkurz, the comment has been update.
> 
> @quanweiZhou I do not see the updated comment in the new commit message.
> 
> Essentially, I think that the commit message needs to explain, at lesat to the level described in issue #4478 :
> 
> * Why is it a problem to turn `SIGTERM` into `SIGKILL` (I think, based on [this comment](https://github.com/kata-containers/kata-containers/issues/4478#issuecomment-1165313855) that your rationale is that we should leave that policy to Kubernetes, which does it on its own)
> * Why we should send a signal to a process that ignores it. Blocking is often temporary, so it could make sense to not immediately convert to a `SIGKILL`, if if the process is ignored, you need to make it clear that you send the process to be ignored only because you expect Kubernetes to later request a `SIGKILL`.
> * What about the non-Kubernetes cases? To what extent do we rely on a policy of the orchestration layer that is not necessarily guaranteed?
> 
> All this makes me wonder if the simplest would not be to simply pass the signals we receive as is. Does it make any sense to turn `SIGTERM` into `SIGKILL`? If so, why does it make sense for Kata and not for `runc` or `crun`?

Thanks @c3d update the comment
```
For runC, send the signal to the init process directly.
For kata, we try to send `SIGKILL` instead of `SIGTERM` when the process has not installed the handler for `SIGTERM`.
The `is_signal_handled` function determine which signal the container process has been handled. But currently `is_signal_handled` is only catching (SigCgt). While the container process is ignoring (SigIgn) or blocking (SigBlk) also should not be converted from the `SIGTERM` to `SIGKILL`. For example, when using terminationGracePeriodSeconds the k8s will send SIGTERM first and then send `SIGKILL`, in this case, the container ignores the `SIGTERM`, so we should send the `SIGTERM` not the `SIGKILL` to the container.
```

src/agent/src/rpc.rs
quanweiZhou: Thanks @Tim-Zhang, change to filter to check signal mask.

Tim-Zhang (approved):
LGTM thanks @quanweiZhou

bergwolf:
@c3d 
> All this makes me wonder if the simplest would not be to simply pass the signals we receive as is. Does it make any sense to turn SIGTERM into SIGKILL? If so, why does it make sense for Kata and not for runc or crun?

To further explain this one, for some init containers, they may not install the SIGTERM handler. If we do not turn SIGTERM into SIGKILL in such case, containerd would still send SIGKILL later on, after a long delay, which is not what we want when stopping the pod. See the comments in the `do_signal_process` function.

bergwolf (approved):

PR #4503, opened 2022-06-21
block: Leverage multiqueue for virtio-block

### LINKED ISSUE #4502: Add multiple queues for virtio-block devices
Similar to network, we can use multiple queues for virtio-block devices. This can help improve storage performance.
   
Today the default number of cpus a VM starts with is 1. Hence the queues used will be 1. The change described in this issue will help improve performance when the default cold-plugged cpus is greater than one by changing this in the config file. This may also help when we use the sandboxing feature with k8s that passes down the sum of the resources required down to Kata.

amshinde:
Similar to network, we can use multiple queues for virtio-block
devices. This can help improve storage performance.
This commit changes the number of queues for block devices to
the number of cpus for cloud-hypervisor and qemu.

Today the default number of cpus a VM starts with is 1.
Hence the queues used will be 1. This change will help
improve performance when the default cold-plugged cpus is greater
than one by changing this in the config file. This may also help
when we use the sandboxing feature with k8s that passes down
the sum of the resources required down to Kata.

Fixes #4502

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -753,6 +753,11 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
 	clhDisk.Readonly = &drive.ReadOnly
 	clhDisk.VhostUser = func(b bool) *bool { return &b }(false)
 
+	queues := int32(clh.config.NumVCPUs)
+	queueSize := int32(1024)
+	clhDisk.NumQueues = &queues
+	clhDisk.QueueSize = &queueSize
+
 	diskRateLimiterConfig := clh.getDiskRateLimiterConfig()
 	if diskRateLimiterConfig != nil {
 		clhDisk.SetRateLimiterConfig(*diskRateLimiterConfig)
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -1341,7 +1341,9 @@ func (q *qemu) hotplugAddBlockDevice(ctx context.Context, drive *config.BlockDri
 			return err
 		}
 
-		if err = q.qmpMonitorCh.qmp.ExecutePCIDeviceAdd(q.qmpMonitorCh.ctx, drive.ID, devID, driver, addr, bridge.ID, romFile, 0, true, defaultDisableModern); err != nil {
+		queues := int(q.config.NumVCPUs)
+
+		if err = q.qmpMonitorCh.qmp.ExecutePCIDeviceAdd(q.qmpMonitorCh.ctx, drive.ID, devID, driver, addr, bridge.ID, romFile, queues, true, defaultDisableModern); err != nil {
 			return err
 		}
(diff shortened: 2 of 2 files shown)


src/runtime/virtcontainers/qemu.go
```diff
@@ -1341,7 +1341,9 @@ func (q *qemu) hotplugAddBlockDevice(ctx context.Context, drive *config.BlockDri
 			return err
 		}
 
-		if err = q.qmpMonitorCh.qmp.ExecutePCIDeviceAdd(q.qmpMonitorCh.ctx, drive.ID, devID, driver, addr, bridge.ID, romFile, 0, true, defaultDisableModern); err != nil {
+		queues := int(q.config.NumVCPUs)
+
+		if err = q.qmpMonitorCh.qmp.ExecutePCIDeviceAdd(q.qmpMonitorCh.ctx, drive.ID, devID, driver, addr, bridge.ID, romFile, queues, true, defaultDisableModern); err != nil {
```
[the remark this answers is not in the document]
amshinde: @fengwang666 Yes, I am planning to add this to scsi as well, maybe in a follow-up PR. I do want to do some performance testing before that, like figuring out if we need to cap the number of queues. Beyond a point, increasing the queues may not give any benefits.

fengwang666 (approved):
LGTM

src/runtime/virtcontainers/clh.go
```diff
@@ -753,6 +753,11 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
 	clhDisk.Readonly = &drive.ReadOnly
 	clhDisk.VhostUser = func(b bool) *bool { return &b }(false)
 
+	queues := int32(clh.config.NumVCPUs)
+	queueSize := int32(1024)
```
liubin, line 757: @amshinde Can you add a comment to describe why here use 1024 as a queue size?

bergwolf:
Nice work @amshinde ! Is it the only missing piece to have multi-queue support throughout the block IO stack or do we need to change more things inside the guest to more performance out of it (e.g., enable blk-mq in the guest virtio-block driver etc.)?

amshinde:
@bergwolf This is the kernel config required `CONFIG_BLK_MQ_VIRTIO`. We have to have that explicitly enabled before we moved to fragments structure for building kernel. We dont explicitly enable it, but its default value is "yes
" when VIRTIO is turned on : https://github.com/torvalds/linux/blob/master/block/Kconfig#L215
I verified this with a kernel I built using our kernel setup script and saw that `CONFIG_BLK_MQ_VIRTIO` was turned on by default.

src/runtime/virtcontainers/clh.go
sboeuf: Yep I agree with @liubin it'd be good to understand why you picked 1024. Having a deeper queue might bring some benefits or drawbacks depending on the use case.

sboeuf (approved):
LGTM

PR #4506, opened 2022-06-22
runk: Support `exec` sub-command

### LINKED ISSUE #4363: runk: support exec sub-command
add `runk exec` support

cyyzero:
`exec` will execute a command inside a container which exists and is not
frozon or stopped. *Inside* means that the new process share namespaces
and cgroup with the container init process. Command can be specified by
`--process` parameter to read from a file, or from other parameters such
as arg, env, etc.

Fixes #4363

Signed-off-by: Chen Yiyang <cyyzero@qq.com>

### DIFF
--- a/src/tools/runk/libcontainer/src/builder.rs
+++ b/src/tools/runk/libcontainer/src/builder.rs
@@ -3,22 +3,53 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use crate::container::{get_config_path, ContainerContext};
+use crate::container::{get_config_path, ContainerLauncher};
+use crate::{
+    status::{get_current_container_state, Status},
+    utils::validate_process_spec,
+};
 use anyhow::{anyhow, Result};
 use derive_builder::Builder;
-use oci::Spec;
+use oci::{ContainerState, Process as OCIProcess, Spec};
+use rustjail::container::update_namespaces;
+use rustjail::{container::LinuxContainer, specconv::CreateOpts};
+use slog::{debug, Logger};
+use std::fs::File;
 use std::path::{Path, PathBuf};
 
-#[derive(Default, Builder, Debug)]
-pub struct Container {
+/// Used for create and run commands. It will prepare the options used for creating a new container.
+#[derive(Default, Builder, Debug, Clone)]
+#[builder(build_fn(validate = "Self::validate"))]
+pub struct InitContainer {
     id: String,
     bundle: PathBuf,
     root: PathBuf,
     console_socket: Option<PathBuf>,
+    pid_file: Option<PathBuf>,
 }
 
-impl Container {
-    pub fn create_ctx(self) -> Result<ContainerContext> {
+impl InitContainerBuilder {
+    /// Pre-validate before building InitContainer
+    fn validate(&self) -> Result<(), String> {
+        // ensure container hasn't already been created
+        let id = self.id.as_ref().unwrap();
+        let root = self.root.as_ref().unwrap();
+        let path = root.join(id);
+        if path.as_path().exists() {
+            return Err(format!(
+                "container {} already exists at path {:?}",
+                id, root
+            ));
+        }
+        Ok(())
+    }
+}
+
+impl InitContainer {
+    /// Create ContainerLauncher that can be used to launch a new container.
+    /// It will read the spec under bundle path.
+    pub fn create_launcher(self, logger: &Logger) -> Result<ContainerLauncher> {
+        debug!(logger, "enter InitContainer::create_launcher {:?}", self);
         let bundle_canon = self.bundle.canonicalize()?;
         let config_path = get_config_path(&bundle_canon);
         let mut spec = Spec::load(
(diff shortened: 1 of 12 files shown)


cyyzero:
@liubin @ManaSugi  Please review code

liubin:
@cyyzero Thank you.

Now Kata Containers is using Rust 1.58.1:

https://github.com/kata-containers/kata-containers/blob/78e27de6c3677040a803dcd71e97f3ccf5a5a46e/versions.yaml#L280-L288

Before committing, you can run `cargo fmt` and `cargo clippy` to prevent CI errors.

cyyzero:
@liubin  I have fixed static check warnings that generated by rust 1.58.1.

src/tools/runk/libcontainer/src/container.rs
```diff
@@ -127,6 +140,11 @@ pub fn get_fifo_path(status: &Status) -> PathBuf {
     status.root.join(&status.id).join(EXEC_FIFO_FILENAME)
 }
 
+fn create_pid_file<P: AsRef<Path>>(pid_file: P, pid: Pid) -> Result<()> {
+    fs::write(pid_file.as_ref(), format!("{}", pid))?;
```
liubin, line 144: These two lines can be simplified by:

```rust
fs::write(pid_file.as_ref(), format!("{}", pid))
```

src/tools/runk/src/commands/exec.rs
```diff
@@ -0,0 +1,220 @@
...
+use oci::{ContainerState, Process as OCIProcess, Spec};
+use rustjail::container::update_namespaces;
+use slog::{debug, info, Logger};
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+pub async fn run(opts: Exec, root: &Path, logger: &Logger) -> Result<()> {
+    let status = Status::load(root, &opts.container_id)?;
+    let state = get_current_container_state(&status)?;
+    if let ContainerState::Paused | ContainerState::Stopped = state {
```
liubin, line 22: Can here use a white list instead of a black list to check which state can run exec? For example, only goon when `state == running`.

If a container with a `Created` status will pass this check.

cyyzero:
@liubin Thanks for review. Branch updated.

src/tools/runk/src/commands/exec.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    let ctx = ContainerBuilder::default()
+        .id(opts.container_id.clone())
+        .bundle(status.bundle)
+        .root(root.to_path_buf())
+        .console_socket(opts.console_socket.clone())
+        .pid_file(opts.pid_file.clone())
+        .spec(adapt_exec_spec(
+            status.config.spec,
```
ManaSugi, line 36: I think `status.config.spec` is not definitely `None` here, so it would be better to fix like below. 
What do you think?

``` rust
.spec(adapt_exec_spec(                                                        
    status                                                                    
        .config                                                               
        .spec                                                                 
        .ok_or_else(|| anyhow!("cannot load spec from a running container"))?,
    status.pid,                                                               
    opts,                                                                     
    logger,                                                                   
)?)
```

src/tools/runk/src/commands/exec.rs
```diff
@@ -0,0 +1,223 @@
...
+        } else {
+            return Err(anyhow!("process is empty in spec"));
+        };
+        update_namespaces(logger, spec, pid)?;
+        debug!(&logger, "spec is {:?}", spec);
+        Ok(oci_spec)
+    } else {
+        Ok(None)
+    }
+}
```
ManaSugi, line 71: If you apply my above comment, you can fix here like below. 

```rust
pub fn adapt_exec_spec(                                      
    mut oci_spec: Spec,                                      
    pid: i32,                                                
    opts: Exec,                                              
    logger: &Logger,                                         
) -> Result<Option<Spec>> {                                  
    // if with --process, load process from file             
    if let Some(process_path) = opts.process {               
        oci_spec.process = Some(get_process(process_path)?); 
    } else if let Some(process) = oci_spec.process.as_mut() {
        update_process(process, opts)?;                      
    } else {                                                 
        return Err(anyhow!("process is empty in spec"));     
    };                                                       
    update_namespaces(logger, &mut oci_spec, pid)?;          
    debug!(&logger, "spec is {:?}", oci_spec);               
    Ok(Some(oci_spec))                                       
}
```

ManaSugi (commented):
@cyyzero Thank you for adding the feature!

src/tools/runk/src/commands/exec.rs
cyyzero: Thanks. I think you are right.

cyyzero:
@ManaSugi Thanks for review. I have applied your comment.

liubin:
@cyyzero please add a command line options check for checking if the process is specified:

https://github.com/opencontainers/runc/blob/214c16fd9a3bd9fe56a2476933e31749a7c0576b/exec.go#L158-L160


```bash
# runk exec test 
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /go/src/github.com/kata-containers/kata-containers/src/agent/rustjail/src/container.rs:676:32
```

cyyzero:
@liubin I have checked args to avoid being empty. Just like runc, I also add a validation check for spec.process.

cc @ManaSugi

liubin (approved):

ManaSugi (approved):

lifupan (approved):

src/tools/runk/libcontainer/src/container.rs
```diff
@@ -34,11 +36,14 @@ pub struct ContainerContext {
     pub spec: Spec,
     pub no_pivot_root: bool,
     pub console_socket: Option<PathBuf>,
+    pub pid_file: Option<PathBuf>,
 }
 
 impl ContainerContext {
-    pub async fn launch(&self, action: ContainerAction, logger: &Logger) -> Result<Pid> {
-        Status::create_dir(&self.state_root, &self.id)?;
+    pub async fn launch(&self, action: ContainerAction, logger: &Logger) -> Result<()> {
```
Tim-Zhang, line 43: I think the method `launch` should be refactored in this pr or a new pr, because the conditional branch of action is too many here, it's hard to read now and will be unmaintainable in the future.

src/tools/runk/src/commands/create.rs
```diff
@@ -6,32 +6,24 @@
...
+        .spec(None)
         .build()?
         .create_ctx()?;
 
-    let pid = ctx.launch(ContainerAction::Create, logger).await?;
-
-    if let Some(ref pid_file) = opts.pid_file {
-        create_pid_file(pid_file, pid)?;
-    }
+    info!(&logger, "create ctx successfully");
```
Tim-Zhang, line 23: Is this log necessary?

src/tools/runk/src/commands/exec.rs
```diff
@@ -0,0 +1,210 @@
+// Copyright 2021-2022 Kata Contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::anyhow;
+use anyhow::Result;
```
Tim-Zhang, line 7: We can merge these imports sentences like  
```
use std::path::{Path, PathBuf, fs::File}; 

use libcontainer::{status::Status, builder::ContainerBuilder, container::ContainerAction, status::get_current_container_state,};
```

And I think it could be better if the std crate imported first then the third party ones

src/tools/runk/src/commands/exec.rs
```diff
@@ -0,0 +1,210 @@
...
+use slog::{debug, info, Logger};
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+pub async fn run(opts: Exec, root: &Path, logger: &Logger) -> Result<()> {
+    let status = Status::load(root, &opts.container_id)?;
+    let state = get_current_container_state(&status)?;
+
+    // if state is Created or Running, we can execute the process. Otherwise, return an error.
+    match state {
```
Tim-Zhang, line 24: It seems that `if` is more clear than match

src/tools/runk/src/commands/exec.rs
```diff
@@ -0,0 +1,210 @@
...
+                .config
+                .spec
+                .ok_or_else(|| anyhow!("cannot load spec from a running container"))?,
+            status.pid,
+            opts,
+            logger,
+        )?))
+        .build()?
+        .create_ctx()?;
+    info!(&logger, "create ctx successfully");
```
Tim-Zhang, line 46: Should we keep this log?

src/tools/runk/libcontainer/src/builder.rs
```diff
@@ -15,39 +18,46 @@ pub struct Container {
...
+            if spec.root.is_some() {
+                let mut spec_root = spec
+                    .root
+                    .as_mut()
+                    .ok_or_else(|| anyhow!("root config was not present in the spec file"))?;
+                let rootfs_path = Path::new(&spec_root.path);
+
+                // If the rootfs path in the spec file is a relative path,
+                // convert it into a canonical path to pass validation of rootfs in the agent.
+                if !&rootfs_path.is_absolute() {
```
Tim-Zhang, line 47: It seems that `&` is useless here

src/tools/runk/src/commands/exec.rs
cyyzero: Ok, I will merge them. The sequence is default sorted by `cargo fmt` right now. Maybe cargo can automatically group them by adding a fmt configure file.

src/tools/runk/src/commands/exec.rs
cyyzero: I use it to debug .I will remove it and add some more specific debug output at some right place.

cyyzero:
In order to be compatible with `create`/`run` commands, I refactor libcontainer. 

`Container` in builder.rs is divided into `InitContainer` and `ActivatedContainer`. 

- `InitContainer` is used for `create`/`run` command. It will load spec from given bundle path and works as before.
- `ActivatedContainer` is used by `exec` command. It will read the container's status file, which stores the spec and `CreateOpt` for creating the rustjail::LinuxContainer. Then adapt the spec by replacing the process with given options and updating the namesapces with first process's namespace paths to join the container. 

I also rename the `ContainerContext` as `ContainerLauncher`, which is only used to spawn process now. It uses the `LinuxContaier` in rustjail as the runner. For `create`/`run`, the `launch` method will create a new container and run the first process. For `exec`, the `launch` method will spawn a process which joins a container.

cc @liubin @ManaSugi  @Tim-Zhang @lifupan

liubin:
The refactor made a big change, so I add a DNM label to let's review it again.

Issue #4517, opened 2022-06-23
CCv0: Pin a specific version of image-rs


fidencio:
Instead of relying on `main` repo, which would end up being update on any run of `cargo update`, let's pin a specific version of the `image-rs` crate and ensure we don't fall into bugs due to the non-intended update, and also take a step forward related to reproducible builds.

PR #4518, opened 2022-06-23
CCv0 | agent: Pin a specific version of image-rs

fidencio:
Let's pin a specific version of image-rs, one that pins a specific
version of ocicrypt-rs on their side, and ensure we don't fall into
issues by consuming the content from `main` on those repos, and also
helping to ensure reproducible builds from our side.

Fixes: #4517 

---

Note: with this we also update the vendored dependencies.

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -1487,7 +1487,7 @@ dependencies = [
 [[package]]
 name = "image-rs"
 version = "0.1.0"
-source = "git+https://github.com/confidential-containers/image-rs#ec5f7c53f1860e64c883828ec4c84b03d9b7d123"
+source = "git+https://github.com/confidential-containers/image-rs?rev=ec5f7c53f1860e64c883828ec4c84b03d9b7d123#ec5f7c53f1860e64c883828ec4c84b03d9b7d123"
 dependencies = [
  "anyhow",
  "flate2",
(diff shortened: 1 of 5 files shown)


stevenhorsman (approved):
LGTM - One question though and I’m not suggesting that you aren’t on top of it, but it there release guide or anything being created, so we remember to do things like update the dependency versions (both inside kata-containers code and the ‘external’ ones too)?

fidencio:
> LGTM - One question though and I’m not suggesting that you aren’t on top of it, but it there release guide or anything being created, so we remember to do things like update the dependency versions (both inside kata-containers code and the ‘external’ ones too)?

Yep, documentation will be a full part of the effort, and @bpradipt and I discussed this during this morning.

Right now this PR cannot even be considered part of the "release process", but rather "best practices" to pin a version of a component we rely on, so we avoid surprises if something unexpected changes there.

stevenhorsman:
> > LGTM - One question though and I’m not suggesting that you aren’t on top of it, but it there release guide or anything being created, so we remember to do things like update the dependency versions (both inside kata-containers code and the ‘external’ ones too)?
> 
> Yep, documentation will be a full part of the effort, and @bpradipt and I discussed this during this morning.
> 
> Right now this PR cannot even be considered part of the "release process", but rather "best practices" to pin a version of a component we rely on, so we avoid surprises if something unexpected changes there.

Thanks - I'll try and keep it in mind that we might have to update it if & when new enhancements go into `image-rs` :)

fidencio:
I'm trying to re-trigger the TDX CI but for some reason I'm failing to do so.
I'm adding a `do-not-merge` label as I want this PR to pass that CI before merging it.

arronwy:
> I'm trying to re-trigger the TDX CI but for some reason I'm failing to do so. I'm adding a `do-not-merge` label as I want this PR to pass that CI before merging it.

I triggered the rebuild for TDX CI and passed now :)

fidencio:
Thanks @arronwy!

PR #4533, opened 2022-06-24
tools/snap: simplify nproc

### LINKED ISSUE #3967: Simplify nproc calls
In various places, we run build jobs in parallel (`make -j`).

However, since that can DoS your system if running locally, we tend to reserve a single CPU to avoid DoS-ing our local systems by specifying the cpu count less one. However, the syntax we tend to use can get a bit... hairy. This table summarises the current situation and what we should be doing:

| Description | | Shell | Makefile | Notes |
|-|-|-|-|-|
| good | :smile: | `make -j $(nproc --ignore=1)` | `$(shell nproc --ignore=1)` | What we should be doing |
| bad | :disappointed: | `make -j $(nproc)` | `$$((nproc))` | What we should not do |
| ugly | :sob: | `make -j $(($(nproc)-1))` | `$$(( $$(nproc) - 1  ))` | What we are currently doing |

Extra points if we conditionally set `--ignore=1` if we're running in a CI environment (where we want all the CPU power we can get:

```bash
nproc_args=
[ -n "$CI" ] && nproc_args="--ignore 1"
```

bookinabox:
Replaces calls of nproc	with nproc with

nproc ${CI:+--ignore 1}

to run nproc with one less processing unit than the maximum to prevent
DOS-ing the local machine.

If process is being run in a container (determined via whether $CI is
null), all processing units avaliable will be used.

Fixes #3967

### DIFF
--- a/docs/Developer-Guide.md
+++ b/docs/Developer-Guide.md
@@ -425,7 +425,7 @@ To build utilizing the same options as Kata, you should make use of the `configu
 $ cd $your_qemu_directory
 $ $packaging_dir/scripts/configure-hypervisor.sh kata-qemu > kata.cfg
 $ eval ./configure "$(cat kata.cfg)"
-$ make -j $(nproc)
+$ make -j $(nproc --ignore=1)
 $ sudo -E make install
 ```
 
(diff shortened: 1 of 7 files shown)


bergwolf (approved):
Sounds reasonable. Thanks!

liubin (approved):

BbolroC:
Please do not care about a result for s390x (performing some experiments atm). Thanks.

BbolroC:
I found that this change breaks a CI job for s390x, which has the connection to the slave instance lost during the test. I tried the test 4 times over 3 different machines, but no success and the same failures. I have just confirmed that this does not happen with the main branch. Please see the log: http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-PR/1963/

bookinabox:
Not too familiar with how most things work still, but I did notice problem in my code after looking at it again. I left an extra $() in the snapcraft.yaml. I fixed that, so hopefully it'll work.

I don't think I'm able to re-run the jenkins job. Is there a better way for me to test by myself?

PR #4543, opened 2022-06-28
runtime-rs: Dragonball sandbox - add Vcpu::configure() function for aarch64

xuejun-xj:
This pull request adds Vcpu::configure() function for aarch64.

Fixes: #4445

PR #4515, opened 2022-06-22
runtime-rs: built-in Dragonball sandbox part III - virtio-blk, virtio-fs, virtio-net and VMM API support

studychao:
In this PR, we introduce the abstractions for the virtio-blk, virtio-fs, virtio-net manager, and VMM API support for the runtime to communicate with VMM in one process.

PR #4544, opened 2022-06-28
runtime-rs: Dragonball-sandbox - add virtio device feature support for aarch64

xuejun-xj:
In this PR, we introduce two virtio device features for aarch64.

First, we implement generate_virtio_device_info() and get_virtio_mmio_device_info() funtions to support the mmio_device_info member, which is used by FDT.

Second, we add the implementation of legacy device on aarch64 platform.

### DIFF
--- a/src/dragonball/src/device_manager/legacy.rs
+++ b/src/dragonball/src/device_manager/legacy.rs
@@ -156,18 +156,26 @@ pub(crate) mod aarch64 {
 
     type Result<T> = ::std::result::Result<T, Error>;
 
+    /// LegacyDeviceType: com1
+    pub const COM1: &str = "com1";
+    /// LegacyDeviceType: com2
+    pub const COM2: &str = "com2";
+    /// LegacyDeviceType: rtc
+    pub const RTC: &str = "rtc";
+
     impl LegacyDeviceManager {
+        /// Create a LegacyDeviceManager instance handling legacy devices.
         pub fn create_manager(
             bus: &mut IoManager,
             vm_fd: Option<Arc<VmFd>>,
             resources: &HashMap<String, DeviceResources>,
         ) -> Result<Self> {
             let (com1_device, com1_eventfd) =
-                Self::create_com_device(bus, vm_fd.as_ref(), resources.get("com1").unwrap())?;
+                Self::create_com_device(bus, vm_fd.as_ref(), resources.get(COM1).unwrap())?;
             let (com2_device, com2_eventfd) =
-                Self::create_com_device(bus, vm_fd.as_ref(), resources.get("com2").unwrap())?;
+                Self::create_com_device(bus, vm_fd.as_ref(), resources.get(COM2).unwrap())?;
             let (rtc_device, rtc_eventfd) =
-                Self::create_rtc_device(bus, vm_fd.as_ref(), resources.get("rtc").unwrap())?;
+                Self::create_rtc_device(bus, vm_fd.as_ref(), resources.get(RTC).unwrap())?;
 
             Ok(LegacyDeviceManager {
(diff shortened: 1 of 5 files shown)


jodh-intel:
Hi @xuejun-xj - Similar to #4543, this PR claims to only be adding a feature for aarch64, but there are 25 commits which seem to be doing a lot more ;) Does this PR need a rebase or something?

Given the number and size of these runtime-rs PRs, it would be extremely helpful to keep them as small and simple as possible. That will ensure they can be reviewed quickly and hopefully landed quickly too ;)

xuejun-xj:
> Hi @xuejun-xj - Similar to #4543, this PR claims to only be adding a feature for aarch64, but there are 25 commits which seem to be doing a lot more ;) Does this PR need a rebase or something?
> 
> Given the number and size of these runtime-rs PRs, it would be extremely helpful to keep them as small and simple as possible. That will ensure they can be reviewed quickly and hopefully landed quickly too ;)

Thanks, I will fix it.

xuejun-xj:
@jodh-intel Thanks for your recommendations. It is the same problem with issue #4543 . This PR is based on mainline Part III (#4515 ). With the same reason, we will also rebase this PR simultaneously after mainline Part III is merged into runtime-rs. In the future, we will also avoid such PRs and disassemble independent modules as much as possible.

xuejun-xj:
Hi, @jodh-intel !
I have already rebased my commits. Please review them. Thank you! :)

src/dragonball/src/device_manager/legacy.rs
```diff
@@ -140,6 +146,83 @@ pub(crate) mod x86_64 {
...
+            bus: &mut IoManager,
+            vm_fd: Option<Arc<VmFd>>,
+            resources: &HashMap<String, DeviceResources>,
+        ) -> Result<Self> {
+            let (com1_device, com1_eventfd) =
+                Self::create_com_device(bus, vm_fd.as_ref(), resources.get("com1").unwrap())?;
+            let (com2_device, com2_eventfd) =
+                Self::create_com_device(bus, vm_fd.as_ref(), resources.get("com2").unwrap())?;
+            let (rtc_device, rtc_eventfd) =
+                Self::create_rtc_device(bus, vm_fd.as_ref(), resources.get("rtc").unwrap())?;
```
jodh-intel, line 170: Nit: It might be worth defining `com1`, `com2` and `rtc` as `public const` values as these strings are duplicated in two files in this PR.

jodh-intel (approved):
Thanks @xuejun-xj.

lgtm

src/dragonball/src/device_manager/legacy.rs
xuejun-xj: Thanks, @jodh-intel ! I have addressed that.

src/dragonball/src/device_manager/legacy.rs
```diff
@@ -42,6 +44,10 @@ pub enum Error {
 pub struct LegacyDeviceManager {
     #[cfg(target_arch = "x86_64")]
     i8042_reset_eventfd: EventFd,
+    #[cfg(target_arch = "aarch64")]
+    pub(crate) _rtc_device: Arc<Mutex<RTCDevice>>,
```
liubin, line 48: Does it have any special meaning that adding an underscore as a prefix of the `_rtc_device`?

xuejun-xj: Thanks for your question! @liubin In rust, if a variable is not in use, it should add an underscore as a prefix. We use `_rtc_device` here as a placeholder for extending it in future.

liubin (approved):

Issue #4419, opened 2022-06-07
When QEMU process is killed by oom-killer Pod Status is Succeeded instead of Error


skaegi:
We've noticed when there is memory pressure on the nodes where we run kata that invariably the VMM processes are the top of the list to be killed by the oom-killer as they generally have several Gigs of RAM in use. e.g. `/opt/kata/bin/qemu-system-x86_64`

Unfortunately when the oom-killer sends a sigkill to the vmm process to reclaim memory the resulting Pod state is...
```
Status:       Succeeded
```
With all containers similarly looking like...
```
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Tue, 07 Jun 2022 11:38:21 -0700
      Finished:     Tue, 07 Jun 2022 11:40:18 -0700
 ```
 
 If running Jobs or TektonTasks this state is a real problem for us because this appears that things completed correctly when in reality things were summarily halted half way through the job/task.
 
 Instead the pod State should be something  like...
 ```
Status: Failed (or possibly OOMKilled but this is harder to determine)
```
With any "running" containers looking like...
```
    State:          Terminated
      Reason:       Error
      Exit Code:    255
      Started:      Tue, 07 Jun 2022 15:17:56 -0700
      Finished:     Tue, 07 Jun 2022 15:18:25 -0700
 ```
 
 Would it be possible for the shim / runtime to manage an abruptly terminated VMM process it's managing and report the error.
 
 --
 
To replicate this problem you can simulate what the oom-killer is doing by starting any pod that stays running under kata onto the pod's node and `kill -9` the associated `/opt/kata/bin/qemu-system-x86_64` process or whichever VMM process. Then exit the node and `kubectl describe` the pod

To replicate the "correct" error state again start any pod that stays running under kata and then nsenter onto the pod's node and `kill -9` the `/opt/kata/bin/containerd-shim-kata-v2` process. Again exit the node and `kubectl describe` the pod

skaegi:
Can someone give me a hand here. I'd like to contribute a fix but am having a hard time figuring out where to start. Thanks.

fengwang666:
@skaegi It probably requires some code tracing. You can start with kubelet to see how the PodStatus is updated, and then trace down to containerd, and kata runtime to see which information is passed up to transit the PodStatus. 
Also, have you reproduced this on runc container?
Can you also upload the kubelet and containerd logs?

fengwang666:
[...]

PR #4549, opened 2022-06-29
shim: set a non-zero return code if the wait process call failed.

liubin:
Return code is an int32 type, so if an error occurred, the default value
may be zero, this value will be created as a normal exit code.

Set return code to 255 will let the caller(for example Kubernetes) know
that there are some problems with the pod/container.

Fixes: #4419

Signed-off-by: liubin <liubin0329@gmail.com>


fidencio (approved):
lgtm, thanks @liubin!

fengwang666 (approved):

GabyCT (approved):

Issue #4551, opened 2022-06-29
CC | Use ubuntu as the base for the VM image


fidencio:
Our CI has been running tests using ubuntu as the VM image, but `kata-deploy` will generate a `clearlinux` based image.  This is a problem reported upstream (see: https://github.com/kata-containers/tests/issues/4850), and it also should be solved there.

While we don't have a solution there, let's ensure we switch the default image in the CCv0 branch to using an ubuntu based image as that's the one we've been using in our tests, and that's the one we can, for sure, install the extra packages needed.

The decision on doing the switch to `ubuntu` for the default image, rather than ensuring everything works with `clearlinux` is to make sure we don't spend too much time and end up derailing the progress of this effort with something that's not that much important for the stage of this right now.  Of course, once we drop the extra external dependencies and have a decision on what to do with the main branch, we'll follow it in the CCv0 branch as well.

PR #4552, opened 2022-06-29
CC | versions: Always use ubuntu as rootfs for CC

fidencio:
As we're still depending on components that are only being tested on
Ubuntu, let's make sure the VM image distributed is exactly the same
we've been testing.

Fixes: #4551

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -124,7 +124,7 @@ assets:
         name: "ubuntu"
         version: "latest"
       x86_64:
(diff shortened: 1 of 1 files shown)


GabyCT (approved):

stevenhorsman (commented):
LGTM - Getting the deploy code using ubuntu which we test against makes a lot of sense. Thanks!

Issue #4564, opened 2022-06-30
CC | Provide a way to build Kata Containers with the needed changes for Confidential Containers


fidencio:
Currently there's no single command that allow us to build Kata Containers in a way it can be used in the Confidential Containers effort.
What's currently missing is a build target in the scripts used to generate the tarball that could build Kata Containers:
* ensuring the correct install path for it to be used with Confidential Containers;
* ensuring the correct path for the binaries used;
* enabling the image's service_offload configuration;
* removing the non-needed VMM configuration for the CC effort

PR #4565, opened 2022-06-30
CC | Build the shim-v2 with the needed configurations for Confidential Containers usage

fidencio:
Let's add a new build target for our local-build scripts, cc-shim-v2,
and use it to build Kata Containers properly configured for the CC
use-case.

Fixes: #4564

Together with this work, we've:
* Exposed to the user the ability to configure the image's service_offload configuration during build time
* The ability to remove non-needed VMM configurations (for the CC use case)


fidencio:
This PR has a dependency on https://github.com/kata-containers/tests/pull/4895

fidencio:
> LGTM I wonder if it is worth updating the ACRN and firecracker config as well, even though we aren't using them (and they could be hard coded to false), just so we don't have two different approaches (some comment in and others set to true/false)?

I wouldn't bother for now, @stevenhorsman.
IMHO having the configuration option there gives the hopes that the VMMs will, at some point, work with that configuration.  Also, it's easier to add a configuration whenever it's needed than removing one, as there's always a chance to break user scripts removing options.

GabyCT (approved):

Issue #4596, opened 2022-07-06
commit message line limit is way to small


bergwolf:
Right now we have a `72` bytes limit, which is way too small. Even kernel nerds have extended to 100 bytes (from the years old 80!). Let's extend it to `150` bytes (was 149 before 72) since very few people still use such a small terminal nowadays.

PR #4597, opened 2022-07-06
action: revert commit message limit to 150 bytes

bergwolf:
So that we can add move info there and few people use such small
terminals nowadays.

Fixes: #4596


fidencio (approved):
lgtm, thanks @bergwolf!

liubin (approved):

Issue #4599, opened 2022-07-06
Cannot disable SELinux label on VMM process by configuration


ManaSugi:
# Description of problem

Currently, we cannot disable SELinux label on VMM process by the runtime configuration (`disable_selinux`) because Kata runtime does not handle the config properly.

# Expected result

Can disable SELinux on VMM process (qemu and virtiofsd) by setting `disable_selinux = true` in the config.

# Actual result

Applied `svirt_lxc_net_t` (alias of `container_t`) label as type to VMM process even if `disable_selinux = true`.
(My environment does not have `container_kvm_t`, so the runtime applies `container_t` instead.)

```sh
$ ps auxZ | grep qemu
[...]

PR #4600, opened 2022-07-06
runtime: Fix DisableSelinux config

ManaSugi:
Enable Kata runtime to handle `disable_selinux` flag properly in order
to be able to change the status by the runtime configuration whether the
runtime applies the SELinux label to VMM process.

Fixes: #4599


liubin:
@ManaSugi does it mean that this config was never affected? Since it was under the `runtime` section.

ManaSugi:
@liubin  Yes, the config was under runtime section, so the `HypervisorConfig` could not take it.
In the first place, the hypervisors except for `clh` could not take it even if the config was under hypervisor section because the config wan't handled in  `new{Qemu,Acrn,Firecracker}HypervisorConfig()` function.

fidencio (approved):
lgtm, thanks @ManaSugi!

cc @egernst

PR #4621, opened 2022-07-11
Pass vhost-user-blk device to container

caiwenzh:
As kubernetes doesn't take vhost-user-blk device into consider, when
process block volume it expects the device CSI driver provided is a
real block device. But after setup vhost-user block device, there is
only a socket and a block device file (no real block device is attached
to it). If pass the vhost-user-blk device directly to kubernetes it
will fail at device mapping stage.

To avoid this problem:
  In CSI driver:
    1. create a small image file <vhost_user_store_path>/<devname>.img
    2. losetup -f <vhost_user_store_path>/<devname>.img
    3. pass the loopback device attached to the image file to kubernetes.

  In kata, check the devices when create a new container, if the device
is a loopback device try to retieve the vhost-user-blk information and
replace the loopback device with vhost-user-blk device.

### DIFF
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -9,7 +9,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -59,6 +61,8 @@ var cdromMajors = map[int64]string{
 // https://github.com/torvalds/linux/blob/master/include/uapi/linux/major.h
 // #define FLOPPY_MAJOR		2
 const floppyMajor = int64(2)
+
+const loopMajor = int64(7)
 
 // Process gathers data related to a container process.
 type Process struct {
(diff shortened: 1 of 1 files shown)


liubin:
@caiwenzh thank you for your PR.

Is the CSI driver a standard way of handling vhost-user-blk? Or just a workaround?

Or can you get some hints from https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/using-SPDK-vhostuser-and-kata.md ?

caiwenzh:
@liubin Thanks for your comment. It's not a standard way. This proposal is similar to direct-assigned volume (https://github.com/kata-containers/kata-containers/pull/3248). It requires to propose a solution in both CSI driver and Kata side. Direct-assigned volume fixes the gap of mounted volume. This PR is trying to fix the gap of block volume. 
yes, I did refer to the link you mentioned. it bypasses the problem as it uses config.json directly.

fidencio:
@caiwenzh, we're closing this PR as it's been opened and inactive for around one year.

Issue #4627, opened 2022-07-11
packaging: Avoid duplication in the code responsible for building a TEE kernel


fidencio:
The code responsible for building a TEE specific kernel can be simplified to avoid duplication and ease its maintenance in the future.

PR #4628, opened 2022-07-11
kernel: Deduplicate code used for building TEE kernels

fidencio:
There's no need to have the entire function for building SEV / TDX
duplicated.

Let's remove those functions and create a `get_tee_kernel()` which takes
the TEE as the argument.

Fixes: #4627


fidencio:
I've also added a fix for #4629 as part of this PR, as the fix depends on the first commit here.

stevenhorsman (approved):
LGTM - Thanks!

Issue #4638, opened 2022-07-12
packaging: Rework how ${BUILD_SUFFIX} is used with the QEMU builder scripts


fidencio:
Currently the `${BUILD_SUFFIX}` environment variable is part of the QEMU builder scripts but it's not actually used with the qemu binary, as if it's set the binary will always be names as `-experimental`.

Also, just for clarity sake, let's append a `-` to the `${BUILD_SUFFIX}` rather than passing a `-${BUILD_SUFFIX}` variable (which would then lead to a qemu binary named `--${BUILD_SUFFIX}`.

PR #4639, opened 2022-07-12
packaging: Rework how ${BUILD_SUFFIX} is used with the QEMU builder scripts

fidencio:
Currently the `${BUILD_SUFFIX}` environment variable is part of the QEMU builder scripts but it's not actually used with the qemu binary, as if it's set the binary will always be names as `-experimental`.

Also, just for clarity sake, let's append a `-` to the `${BUILD_SUFFIX}` rather than passing a `-${BUILD_SUFFIX}` variable (which would then lead to a qemu binary named `--${BUILD_SUFFIX}`.

For more details, please, see each one of the commit messages.

Fixes: #4638


tools/packaging/static-build/qemu/Dockerfile
```diff
@@ -72,7 +72,8 @@ RUN git clone --depth=1 "${QEMU_REPO}" qemu && \
     git fetch --depth=1 origin "${QEMU_VERSION}" && git checkout FETCH_HEAD && \
     scripts/git-submodule.sh update meson capstone && \
     /root/patch_qemu.sh "${QEMU_VERSION}" "/root/kata_qemu/patches" && \
-    (PREFIX="${PREFIX}" /root/configure-hypervisor.sh -s "kata-qemu${BUILD_SUFFIX}" | xargs ./configure \
+    [ -n "${BUILD_SUFFIX}" ] && HYPERVISOR_NAME="kata-qemu-${BUILD_SUFFIX}" || HYPERVISOR_NAME="kata-qemu" && \
+    (PREFIX="${PREFIX}" /root/configure-hypervisor.sh -s "${HYPERVISOR_NAME}" | xargs ./configure \
 	--with-pkgversion="kata-static${BUILD_SUFFIX}") && \
```
stevenhorsman, line 77: It looks like `kata-static${BUILD_SUFFIX}` might need hypen to separate the suffix like the HYPERVISOR_NAME needs?

fidencio: Nice catch, thanks @stevenhorsman!

Issue #4645, opened 2022-07-13
qemu: Add liburing to qemu build


amshinde:
io_uring is a Linux API for asynchronous I/O introduced in qemu 5.0.

https://wiki.qemu.org/Features/IOUring

It is designed to better performance than older `aio` API.
We could leverage this in order to get better storage performance.

PR #4646, opened 2022-07-13
qemu: Add liburing to qemu build

amshinde:
io_uring is a Linux API for asynchronous I/O introduced in qemu 5.0.
It is designed to better performance than older `aio` API.
We could leverage this in order to get better storage performance.

Add liburing-dev to qemu build so that we can make use of it.

Fixes: #4645

### DIFF
--- a/tools/packaging/static-build/qemu/Dockerfile
+++ b/tools/packaging/static-build/qemu/Dockerfile
@@ -11,6 +11,7 @@ WORKDIR /root/qemu
 # This is required to keep build dependencies with security fixes.
 ARG CACHE_TIMEOUT
 RUN echo "$CACHE_TIMEOUT"
+ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get upgrade -y && \
     apt-get --no-install-recommends install -y \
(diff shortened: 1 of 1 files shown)


bergwolf (approved):

tools/packaging/static-build/qemu/Dockerfile
```diff
@@ -59,14 +60,19 @@ ARG BUILD_SUFFIX
 ARG QEMU_DESTDIR
 ARG QEMU_TARBALL
 
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN git clone  https://github.com/axboe/liburing/ ~/liburing && \
+    cd ~/liburing && \
+    git checkout tags/liburing-2.1 && \
+    make && make install && ldconfig
```
egernst, line 67: What happens when the host kernel doesn't have iouring support?

amshinde: yes, will raise a follow-up PR to update the qemu usage. iouring was added in 5.1, so its should be mostly available unless you are using a kernel older than that. We can check for its availability, and fallback to another I/O mechanism.

amshinde:
@GabyCT @cmaf Can you take a look?

GabyCT (approved):

Issue #4636, opened 2022-07-12
Documentation is needed to use DAX


liubin:
Some related issues:

- https://github.com/kata-containers/kata-containers/issues/1132
- https://github.com/kata-containers/kata-containers/issues/4632

assume that the DAX(`virtio_fs_cache_size`) can work out of the box, but indeed it does not, extra work is needed.

Documentation is needed to show how to build experimental parts to use DAX.

PR #4648, opened 2022-07-13
docs: add a how-to to help users to run DAX

liubin:
Kata supports DAX but not out of the box.
Currently, users need to build an experimental QEMU to use
DAX with virtiofs, but the configuration file makes users confused.

This commit adds comments in the configuration file about the
limitation of DAX support and adds a how-to to help user to
build an experimental QEMU.

Fixes: #4636

Signed-off-by: liubin <liubin0329@gmail.com>

### DIFF
--- a/docs/how-to/how-to-build-experimental-qemu.md
+++ b/docs/how-to/how-to-build-experimental-qemu.md
@@ -1,0 +1,59 @@
+# How to build and use experimental QEMU
+
+This document describes how to build an experimental QEMU to use features that the
+main branch of QEMU does not support yet, for example, DAX with `Nydus`.
+
+This how-to will show how to build an experimental QEMU and use it with DAX.
+
+## Pre-requisites
+
+You must have cloned the Kata Containers repo first.
+
+## Build experimental QEMU
+
+```bash
+$ cd $GOPATH/github.com/kata-containers/kata-containers/tools/packaging/static-build/qemu
+$ make build-experimental
+```
+
+If the build is finished successfully, you will get the build artifact like this:
+
+```bash
+$ ls -tl
+total 16760
+-rw-r--r-- 1 vagrant vagrant 17139024 Jul 13 06:40 kata-static-qemu-experimental.tar.gz
+```
+
+## Use the experimental QEMU
+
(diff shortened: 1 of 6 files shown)


docs/how-to/how-to-build-experimental-qemu.md
```diff
@@ -0,0 +1,55 @@
...
+Edit your `configuration.toml` and ensure these configuration items:
+
+```
+# cache mod should be always or auto to use DAX
+virtio_fs_cache = "auto"
+shared_fs = "virtio-fs"
+virtio_fs_cache_size = 256 
+```
+
+After updating the `configuration.toml`, you can create new containers with DAX enabled.
```
ManaSugi, line 55: I think it would be better to add such a following note to the last line or each configuration file because this behavior is for Kata only.

> **Note:**
>
> DAX is disabled if the `virtio_fs_cache_size` is set to 0.

liubin: Added to both this how-to and the two configuration files.

ManaSugi: Thank you!

fidencio:
@liubin, it's more complicated than that.  The rust version of virtiofsd doesn't support DAX yet, and we entirely dropped building the C version.

Maybe what could be done would be adding back the C version when building QEMU experimental?  Still, not sure if we should be advertising users to get something that's already deprecated with support that never made upstream. :-/

bergwolf:
@liubin @fidencio Well, if necessary, users can use `nydusd` for DAX support, which is enabled by default plus the newly introduced per-file DAX feature.

fidencio:
> @liubin @fidencio Well, if necessary, users can use `nydusd` for DAX support, which is enabled by default plus the newly introduced per-file DAX feature.

That's a good point, @bergwolf.
I'd rather have instructions to use nydusd if DAX is required.

We should also be shipping nydusd as part of the release tarball, but that's for another issue.

fidencio:
> We should also be shipping nydusd as part of the release tarball, but that's for another issue.

I've opened https://github.com/kata-containers/kata-containers/issues/4665 for that.

Issue #4659, opened 2022-07-14
CC | packaging: Add a new cc-tdx-cloud-hypervisor-tarball target


fidencio:
For consistency sake, we need to add a new target for building a TDX capable Cloud Hypervisor
for CC..  As the current version of Cloud Hypervisor is already built with TDX support, we can simply rely on calling the same `install_cc_clh()` function, as done for the non-tee `cc` target.

PR #4660, opened 2022-07-14
CC | packaging: Build and ship Cloud Hypervisor

fidencio:
We're adding a new target for building a TDX capable Cloud Hypervisor
for CC.

As the current version of Cloud Hypervisor is already built with TDX
support, we just rely on calling the same `install_cc_clh()` function,
as done for the non-tee `cc` target.

Fixes: #4659


stevenhorsman (approved):
LGTM - thanks!

Issue #4663, opened 2022-07-14
Some daemon processes left after container terminated when enabled shared pidns


lifupan:
When a container terminated, we should make sure there's no processes
left after destroying the container.

kata-agent depended on the kernel's pidns
to destroy all of the process in a container after the 1 process
exit in a container. This is true for those container using a
separated pidns, but for the case of shared pidns within the
sandbox, the container exit wouldn't trigger the pidns terminated,
and there would be some daemon process left in this container, this
wasn't expected.

PR #4664, opened 2022-07-14
container: kill all of the processes in a container when it terminated

lifupan:
When a container terminated, we should make sure there's no processes
left after destroying the container.

Before this commit, kata-agent depended on the kernel's pidns
to destroy all of the process in a container after the 1 process
exit in a container. This is true for those container using a
separated pidns, but for the case of shared pidns within the
sandbox, the container exit wouldn't trigger the pidns terminated,
and there would be some daemon process left in this container, this
wasn't expected.

Fixes: #4663

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -1092,6 +1092,16 @@ impl BaseContainer for LinuxContainer {
         fs::remove_dir_all(&self.root)?;
 
         if let Some(cgm) = self.cgroup_manager.as_mut() {
+            // Kill all of the processes created in this container to prevent
+            // the leak of some daemon process when this container shared pidns
+            // with the sandbox.
+            let pids = cgm.get_pids().context("get cgroup pids")?;
+            for i in pids {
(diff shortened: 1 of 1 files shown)


fidencio:
@lifupan, looks good to me, would be possible to have a test added?
Also, this is material to be backported at least to the stable-2.5 branch.

lifupan:
> @lifupan, looks good to me, would be possible to have a test added? Also, this is material to be backported at least to the stable-2.5 branch.

NP， I'll add a test case for it later.

fidencio (approved):
lgtm, thanks @lifupan!

cmaf (approved):
thanks @lifupan

amshinde:
@lifupan Can you add an integration test for this before this is merged.

liubin:
> @lifupan Can you add an integration test for this before this is merged.

I added a DNM label for this.

lifupan:
> @lifupan Can you add an integration test for this before this is merged.

integrated  test:  https://github.com/kata-containers/tests/pull/4937

PR #4676, opened 2022-07-18
runtime-rs: fix some bugs to make runtime-rs on aarch64

xuejun-xj:
This PR fixes some bugs when use "make" in runtime-rs directory. 

1. fix a vcpu_manager warning in dragonball
2. fix a module use warning in runtime-rs
3. add compile option file for aarch64
4. update dbs-xxx dependencies in runtime-rs/Cargo.toml

### DIFF
--- a/src/dragonball/src/vm/aarch64.rs
+++ b/src/dragonball/src/vm/aarch64.rs
@@ -98,7 +98,8 @@ impl Vm {
         );
         self.vcpu_manager()
             .map_err(StartMicroVmError::Vcpu)?
-            .set_reset_event_fd(reset_eventfd);
+            .set_reset_event_fd(reset_eventfd)
+            .map_err(StartMicroVmError::Vcpu)?;
 
         // On aarch64, the vCPUs need to be created (i.e call KVM_CREATE_VCPU) and configured before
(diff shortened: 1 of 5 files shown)


src/dragonball/src/vm/aarch64.rs
```diff
@@ -98,7 +98,8 @@ impl Vm {
         );
         self.vcpu_manager()
             .map_err(StartMicroVmError::Vcpu)?
-            .set_reset_event_fd(reset_eventfd);
+            .set_reset_event_fd(reset_eventfd)
+            .map_err(StartMicroVmError::Vcpu)?;
```
Tim-Zhang, line 102: It's weird to have two same map_err here. How about using Into/From?

xuejun-xj: Hi, @Tim-Zhang !
Thanks for your recommendation. I have raised an issue for this problem, which explains why we use two map_err here. And in the future, we will work on it. 

issue: #4679

jodh-intel (approved):
Thanks @xuejun-xj. I can't test it on aarch64, but

lgtm

liubin (approved):

Issue #4677, opened 2022-07-18
Runtime-rs: remove the value of hypervisor path in  Dragonball configuration.toml


Tim-0731-Hzt:
As a built in VMM, Path, jailer path, ctlpath  are not needed for Dragonball. So we don't generate those value in Makefile.

PR #4678, opened 2022-07-18
runtime-rs: remove the value of hypervisor path in DB config

Tim-0731-Hzt:
As a built in VMM, Path, jailer path, ctlpath are not needed for
Dragonball. So we don't generate those value in Makefile.

Fixes: #4677


src/runtime-rs/Makefile
```diff
@@ -49,9 +48,7 @@ DEFAULT_HYPERVISOR ?= $(HYPERVISOR_DB)
 
 HYPERVISORS := $(HYPERVISOR_DB)
 
-DBPATH := $(DBBINDIR)/$(QEMUCMD)
-DBVALIDHYPERVISORPATHS := [\"$(DBPATH)\"]
-
+DBVALIDHYPERVISORPATHS := []
```
liubin, line 51: Is `DBVALIDHYPERVISORPATHS` itself still needed?

Tim-0731-Hzt: yes, because runtime will check the value of valid_hypervisors, for Dragonball case, we need it to be empty.

jodh-intel (approved):
Thanks @Tim-0731-Hzt.

lgtm

liubin (approved):

Issue #4680, opened 2022-07-18
Runtime-rs: Set share sandbox pid namespace


Tim-0731-Hzt:
Cannot see the pause container after the pod contianer start

PR #4681, opened 2022-07-18
runtime-rs: fix set share sandbox pid namespace

Tim-0731-Hzt:
Set the share sandbox pid namepsace from spec

Fixes:#4680

### DIFF
--- a/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
@@ -80,8 +80,8 @@ impl Container {
         let mut inner = self.inner.write().await;
         let toml_config = self.resource_manager.config().await;
         let config = &self.config;
-        amend_spec(&mut spec, toml_config.runtime.disable_guest_seccomp).context("load spec")?;
-
+        let sandbox_pidns = amend_spec(&mut spec, toml_config.runtime.disable_guest_seccomp)
+            .context("load spec")?;
         // handler rootfs
         let rootfs = self
             .resource_manager
@@ -143,7 +143,7 @@ impl Container {
             storages,
             oci: Some(spec),
             guest_hooks: None,
-            sandbox_pidns: false,
+            sandbox_pidns,
             rootfs_mounts: vec![],
         };
 
@@ -373,7 +373,7 @@ impl Container {
     }
 }
 
-fn amend_spec(spec: &mut oci::Spec, disable_guest_seccomp: bool) -> Result<()> {
+fn amend_spec(spec: &mut oci::Spec, disable_guest_seccomp: bool) -> Result<bool> {
     // hook should be done on host
     spec.hooks = None;
 
@@ -390,6 +390,8 @@ fn amend_spec(spec: &mut oci::Spec, disable_guest_seccomp: bool) -> Result<()> {
             resource.network = None;
         }
 
+        // Host pidns path does not make sense in kata. Let's just align it with
+        // sandbox namespace whenever it is set.
         let mut ns: Vec<oci::LinuxNamespace> = Vec::new();
         for n in linux.namespaces.iter() {
             match n.r#type.as_str() {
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -400,3 +403,17 @@ fn amend_spec(spec: &mut oci::Spec) -> Result<()> {
 
     Ok(())
 }
+
+fn handle_pid_namespace(namespaces: &[oci::LinuxNamespace]) -> bool {
```
jodh-intel, line 407: Could you add a unit test for this new function?

jodh-intel (commented):
Thanks @Tim-0731-Hzt.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
Tim-0731-Hzt: fixed

bergwolf: I believe the idea of the patch is to copy `handlePidNamespace` from go runtime, in which case additional changes are required:
1. please remove the pidns setting from container spec
2. please keep the code comments of the go runtime as it helps others understand why it is implemented that way

bergwolf: @Tim-0731-Hzt Thanks for updating. To be specific, I think the following line of comment is necessary:
```
                // host pidns path does not make sense in kata. Let's just align it with
                // sandbox namespace whenever it is set.
```

Also any explanation why we do not remove the pidns setting from the container spec in runtime-rs?

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -387,6 +388,9 @@ fn amend_spec(spec: &mut oci::Spec) -> Result<()> {
             resource.network = None;
         }
 
+        *sandbox_pidns = handle_pid_namespace(&linux.namespaces);
```
jodh-intel, line 391: This function needs a unit test, but see https://github.com/kata-containers/kata-containers/pull/4692#issuecomment-1189955702

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -79,7 +79,8 @@ impl Container {
         // process oci spec
         let mut inner = self.inner.write().await;
         let config = &self.config;
-        amend_spec(&mut spec).context("load spec")?;
+        let mut sandbox_pidns = false;
+        amend_spec(&mut spec, &mut sandbox_pidns).context("load spec")?;
```
quanweiZhou, line 83: Suggest to return the `sandbox_pidns` by amend_spec not pass a `mut sandbox_pidns`.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -400,3 +404,54 @@ fn amend_spec(spec: &mut oci::Spec) -> Result<()> {
...
+                    return true;
+                }
+            }
+            _ => continue,
+        }
+    }
+    false
+}
+
+#[test]
```
quanweiZhou, line 424: Suggest adding tests in mod
```
#[cfg(test)]
mod tests {
  #[test]
  fn test_***() {}
}
```

quanweiZhou (approved):
Thanks @Tim-0731-Hzt, LGTM.

liubin (approved):

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
bergwolf: @Tim-0731-Hzt ping, any comment about this question?

> Also any explanation why we do not remove the pidns setting from the container spec in runtime-rs?

bergwolf: Ah, I see that it is handled above in line 395. Thanks @Tim-0731-Hzt !

bergwolf (approved):

PR #4700, opened 2022-07-20
runtime: do not mark agent dead in the first connection

wxx213:
runtime: do not mark agent dead when dial fail with kata agent

Sometimes the guest VM boots slowly and may be the runtime will
retry several times to dial the kata agent in VM when start sandbox.

Fixes: https://github.com/kata-containers/kata-containers/issues/4722

Signed-off-by: wxx213 <stellarwxx@163.com>


liubin:
@wxx213 I have a question, you said "the runtime will retry several times", where do you mean?

cmaf (commented):
@wxx213 what VMs are you using?

I'm not sure we should make this change since in this case the agent won't be marked as dead even if it should be.

ManaSugi:
@wxx213 
If this problem occurs due the delay of VM startup, I think you can change the timeout by setting `dial_timeout` in the runtime configuration. Could you try `dial_timeout` whether the issue is solved?

cmaf:
@wxx213 did changing the dial_timeout help?

Issue #4712, opened 2022-07-21
Failed to adjust default vcpu when it is greater than default max vcpu


Tim-0731-Hzt:
when the default vcpu is greater than the default max vcpu, the default vcpu number is not adjusted.

PR #4713, opened 2022-07-21
runtime-rs: handle default_vcpus greator than default_maxvcpu

Tim-0731-Hzt:
when the default_vcpus is greater than the default_maxvcpus, the default
vcpu number should be set equal to the default_maxvcpus.

Fixes: #4712

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/dragonball.rs
+++ b/src/libs/kata-types/src/config/hypervisor/dragonball.rs
@@ -65,6 +65,10 @@ impl ConfigPlugin for DragonballConfig {
                 db.cpu_info.default_maxvcpus = default::MAX_DRAGONBALL_VCPUS;
             }
 
+            if db.cpu_info.default_vcpus as u32 > db.cpu_info.default_maxvcpus {
+                db.cpu_info.default_vcpus = db.cpu_info.default_maxvcpus as i32;
+            }
+
             if db.machine_info.entropy_source.is_empty() {
                 db.machine_info.entropy_source =
                     default::DEFAULT_DRAGONBALL_ENTROPY_SOURCE.to_string();


liubin (approved):

studychao (approved):

src/libs/kata-types/src/config/hypervisor/dragonball.rs
```diff
@@ -65,6 +65,10 @@ impl ConfigPlugin for DragonballConfig {
                 db.cpu_info.default_maxvcpus = default::MAX_DRAGONBALL_VCPUS;
             }
 
+            if db.cpu_info.default_vcpus as u32 > db.cpu_info.default_maxvcpus {
+                db.cpu_info.default_vcpus = db.cpu_info.default_maxvcpus as i32;
```
bergwolf, line 69: Any reason we are not defining `default_vcpus` as `u32`?

Tim-0731-Hzt: 1. When `default_vcpus` < 0, will be set to the actual number of physical cores.
2. When 0 < `default_vcpus` <= number of physical cores, will be set to the specified number

bergwolf: Then what about changing the first condition to:

when `default_vcpus` == 0, set the vcpu number to the actual number of physical cores.

Then `default_vcpus` can be `u32`.

Tim-0731-Hzt: @bergwolf I just checked our design. There are three cases for `default_vcpus`.

1. when `default_vcpus unspecified or 0. it will be set to `default::DEFAULT_GUEST_VCPUS`
2. when `default_vcpus < 0, will be set to the actual number of physical cores.
3. when 0 < `default_vcpus` <= number of physical cores, will be set to the specified number

So, `default_vcpus` has to be `i32`

bergwolf: @Tim-0731-Hzt I see, this is inherited from go runtime. We can optimize but let's not block the PR here. Thanks for explanation.

bergwolf (approved):

Issue #4698, opened 2022-07-20
[agent]: Restarting container causes rootfs to be mounted again on each restart


Champ-Goblem:
I have recently been debugging some memory issues and noticed that some pods stuck in a constant CrashLoopBackOff phase are using a lot of memory as reported by the host machine. After executing into the VM via the debug console, a quick looked showed that the buffers/cache usage of the guest was quite high. This usage seemed to correspond to a lot of mounts of the restarting container's rootfs eg (`/run/kata-containers/shared/containers/<cid>/rootfs`) which seems to happen on each pod restart (each mount had a different container id, which corresponded to the container id history of the restarting container). Unmounting a few of these quickly showed the buffers and cache usage drop by a couple of MBs each time.

fe:
```
> cat /proc/mounts | grep rootfs | wc -l
105
```

As we run `sandbox_cgroup_only=false` these containers will never be restarted when breaching the host's cgroup limits and thus grows until it reaches the size of the hotplug memory allocated to the VM. I implemented a quick and dirty fix for the kata agent to unmount the rootfs on container exit, the patch for which can be seen below:

```
---
 src/agent/src/rpc.rs | 5 ++++-
 1 file changed, 4 insertions(+), 1 deletion(-)

diff --git a/src/agent/src/rpc.rs b/src/agent/src/rpc.rs
index 4ae0f9d19..36d1a6f9f 100644
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -50,7 +50,7 @@ use crate::device::{
 };
 use crate::linux_abi::*;
 use crate::metrics::get_metrics;
-use crate::mount::{add_storages, baremount, STORAGE_HANDLER_LIST};
+use crate::mount::{add_storages, baremount, STORAGE_HANDLER_LIST, remove_mounts};
 use crate::namespace::{NSTYPEIPC, NSTYPEPID, NSTYPEUTS};
 use crate::network::setup_guest_dns;
 use crate::pci;
@@ -553,6 +553,9 @@ impl AgentService {
             let _ = s.send(p.exit_code).await;
         }
 
+        let rootfs_path = format!("/run/kata-containers/shared/containers/{}/rootfs", &cid);
+        remove_mounts(&[rootfs_path])?;
+
         ctr.processes.remove(&pid);
 
         Ok(resp)
-- 
2.37.0
```

This seems to solve the problem and thus the overall VM memory size doesnt grow any more than when the pod was first created.
[...]

PR #4716, opened 2022-07-21
agent: unmount rootfs on container restart

Champ-Goblem:
The current mounting system for the root filesystem was not correctly
handling the unmount steps during a container restart. After restarting
consistently, the agent would have a number of left over mounts from
previous container instances which could be seen with
`cat /proc/mounts`. Each of these mounts would use 1-2MB of guest
buffer/cache memory and therefore would slowly eat host memory over
time.

This patch resolves this by explicitly unmounting the rootfs at
path `/run/kata-containers/shared/containers/<cid>/rootfs` on each
restart.

Fixes: #4698
Signed-off-by: Champ-Goblem <cameron@northflank.com>

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -40,7 +40,7 @@ use rustjail::process::Process;
 use rustjail::specconv::CreateOpts;
 
 use nix::errno::Errno;
-use nix::mount::MsFlags;
+use nix::mount::{umount, MsFlags};
 use nix::sys::{stat, statfs};
 use nix::unistd::{self, Pid};
 use rustjail::cgroups::Manager;
@@ -552,6 +552,15 @@ impl AgentService {
                 return Ok(resp);
             }
         };
+
+        let rootfs_path = Path::new(CONTAINER_BASE)
+            .join("shared/containers")
+            .join(&cid)
+            .join("rootfs");
+        umount(rootfs_path.to_str().unwrap()).context(format!(
+            "failed to unmount rootfs on container exit {:?}",
+            rootfs_path
+        ))?;
 
         // need to close all fd
         // ignore errors for some fd might be closed by stream


src/agent/src/rpc.rs
```diff
@@ -553,6 +553,15 @@ impl AgentService {
...
         };
 
+        let rootfs_path = Path::new(CONTAINER_BASE)
+            .join("shared/containers")
+            .join(&cid)
+            .join("rootfs");
+        umount(rootfs_path.to_str().unwrap()).context(format!(
+            "failed to unmount rootfs on container exit {:?}",
+            rootfs_path
+        ))?;
```
bergwolf, line 563: No, this is the wrong place to clean up container rootfs. `wait_process` is called for both containers and exec processes. In the latter case, we should NEVER umount the container rootfs.

Instead, a container is destroyed and recreated upon restart, and rustjail's `container.destroy()` is supposed to cleanup the container rootfs. Do you see any issue there?

bergwolf (changes requested):

src/agent/src/rpc.rs
liubin: Don't know what's the pod definition of @Champ-Goblem 's , for test purposes, I specified a non-exist command as the container's entry point, and the agent will only get `create_container` be called, the sequence is:

|Shim|             agent|
|---|---|
|->Start ||
||->create_container|
|->Delete ||
|->Shutdown ||

So maybe we need to add a rollback in `create_container` in the agent to do some clear work.

My test yaml file:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cpu-test
spec:
  terminationGracePeriodSeconds: 0
  runtimeClassName: kata
  containers:
  - name: first-cpu-container
    image: busybox:latest 
    command:
        - sleepp
        - "300"
    resources:
      limits:
        cpu: "1"
      requests:
        cpu: "500m"

```

Champ-Goblem: @bergwolf makes sense that the current definition is not in the right place. Funnily enough when I looked at implementing this PR I did see the `destroy` method and also noticed the unmount it performs (https://github.com/kata-containers/kata-containers/blob/main/src/agent/rustjail/src/container.rs#L1088). I believe that the `spec.root` might point to the wrong location, if I remember correctly it was pointing at something like `/run/kata-containers/<cid>/rootfs` instead of `/run/kata-containers/containers/shared/<cid>/rootfs`. Perhaps this is where the bug lies? although I don't have much context if `/run/kata-containers/<cid>/rootfs` is also a valid path.

@liubin my container spec seems pretty similar:
```
- image: ubuntu:latest
  name: crashcontainer
  command:
  - bash
  - -c
  - "exit 1"
  resources:
    requests:
      cpu: 50m
      memory: 64Mi
    limits:
      cpu: 500m
      memory: 512Mi
```

liubin: @Champ-Goblem do you mean this code? It mount `/run/kata-containers/shared/containers/<CID>/rootfs` to `/run/kata-containers/<CID>/rootfs`.

https://github.com/kata-containers/kata-containers/blob/175d4968c5301b8662e8168b6e6a1e296836d48c/src/agent/src/rpc.rs#L1931-L1963

Champ-Goblem: Ah okay then yes the original unmount in `destroy` method looks correct.

bergwolf: @liubin Yes, looks like we need some rollback logic in `do_create_container` to clean up things if it fails.

In @Champ-Goblem's case, the container exits very quickly. Then do you observe that the container rootfs is left mounted? Is there a `do_remove_container` call in your test workflow?

Champ-Goblem: Looks like `do_remove_container` is being called on each restart, the only mounts that are being leftover are the ones at `/run/kata-containers/shared/containers/<cid>/rootfs`.

Champ-Goblem: @liubin @bergwolf do we agree with moving this PR's logic from `do_wait_process` to `do_remove_container` so that its called when a container is remove?

bergwolf: @Champ-Goblem After reviewing the code, `do_wait_process` does call `container.destroy()` and clean up the rootfs mount properly. @liubin and I both tried your yaml locally and cannot reproduce the problem. Can you show how you determined that a container's rootfs mount is left after it is stopped and removed? FYI, there is always a sandbox container (the pause container) running inside the sandbox. I'm not sure if you mistook it as the removed container.

Champ-Goblem: My pod definition:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: crashcontainer
spec:
  selector:
    matchLabels:
      app: crashcontainer
  template:
    metadata:
      annotations:
        sidecar.istio.io/inject: "false"
      labels:
        app: crashcontainer
    spec:
      containers:
      - image: ubuntu:latest
        name: sleep
        command:
        - sleep
        - "500000"
      - image: ubuntu:latest
        name: crashcontainer
        command:
        - bash
        - -c
        - "exit 1"
        resources:
          requests:
            cpu: 50m
            memory: 64Mi
          limits:
            cpu: 500m
            memory: 512Mi
      runtimeClassName: kata-clh
```
(I dont think having the second container is necessary but I might be wrong). 

Then via the debug console of the VM:
```
> cat /proc/mounts
none /run/kata-containers/shared/containers/8d8fbe45f1ff3344bbd4b408d0e8ce69b21878ee5a9ce649457dd1502d5b37a3/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/3212f460b4746ee439342648afcbb8ead93d1fca42057eabcf03051acedfa82f/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/83f5df38761d493b54e711fa3e610f7144eab9f61db328cc2db2d6f488942def/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/8f1517cbbcb997562b9bef15cd0168d8767c737a5b907f7380767f6daf31bafd/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/6fbfb0baf0e920b8232ba1c7860ea659d4e88b519723ce7ddd6a91932894bdf7/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/fcdf5cb799a33a0668d3ccc74153894c46f5e555de8354fe37b0d8eb21394370/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/a7a481ee7508a1735e49d5499fdc996f051c959b910898485ea20f40ec4630a1/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/cc8be23706daadbc9d050bc662504af79bbf3a0d4877540b8fb71b9e2c6dd306/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/f652aafabbd4a35fdeff0de20ad3ca00c082c0ad5c5bdf8e60596d3e7375bd77/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/8385f3a63cc7ad8f274681c49a8e6a06cd1b0102fcdb45f97eb822574a145b8d/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/c68f0c8d884ab930c597b143f0f571044380a4842cd1d84930e616580ff460df/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/10779e293173eaecc90e114ac8a5b5cdb6bd200f6e0c43e90d44809c7c1c6284/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/4724262a24c634e1e806223eeb2c6b815d6fe2c7abe58abe3ff1d85146531dd7/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/c0ce913a3500cd1d3d18450913db5d606f7b2210dbef1808fb9eb76aa424290d/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/d499f0d5e6753cd61c782bdca66bd9102c905cbe00395af6f681535ba1f7d385/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/f16b176011c04c78285253e713d05dee41713738d0fa362c8110bba13e1ecd66/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/73c3fb44c25835a17172f19137d2302b80777062309a09a4e7c9a0436840a7e5/rootfs virtiofs rw,relatime 0 0
none /run/kata-containers/shared/containers/0af690365126272f8d8e0708528c5e84f013a3ad52ad4cc83c248599d179af1b/rootfs virtiofs rw,relatime 0 0
```
Each of the container IDs in each of the above mounts corresponds to the container ID of a previous restart, around 16 restarts had happened when this was taken (18 mounts = 16 restarts + 2 (1 original mount per container))

PR #4719, opened 2022-07-21
github-actions: Add cargo-deny

### LINKED ISSUE #3359: Add a "cargo deny" GitHub action
See:

- https://crates.io/crates/cargo-deny
- https://github.com/confidential-containers/td-shim/pull/20

/cc @fidencio

bookinabox:
Adds cargo-deny to scan for vulnerabilities and license issues regarding
rust crates.

Some modifications were required for the repo to pass the tests:

 ~~Updates ttrpc to avoid using nix 0.16.0
   https://rustsec.org/advisories/RUSTSEC-2021-0119~~ (This got fixed in a different PR while this was pending)

 ~~Updates slog-json to avoid MLP license (copyleft)~~(I ended up allowing copy-left licenses)

 Updates crossbeam-channel because 0.52.0 was a yanked package

 Ignores https://rustsec.org/advisories/RUSTSEC-2020-0071
   because chrono is dependent on that version of time.
   chronotope/chrono#578

 Allow multiple versions of the same package
  (package dependencies require this)

 Adds "oci" to src/libs workplace

 Adds Apache-2.0 license to workplace modules that did not have them
 because cargo-deny complains about them not having licenses.

Notes
GitHub Actions does not have an obvious way to loop over each of the
Cargo.toml files. To avoid hardcoding it, I worked around the problem
using a composite action that first generates the cargo-deny action by
finding all Cargo.toml files before calling this new generated action in
the master workflow.

Fixes #3359

### DIFF
--- a/.github/cargo-deny-composite-action/cargo-deny-generator.sh
+++ b/.github/cargo-deny-composite-action/cargo-deny-generator.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+script_dir=$(dirname "$(readlink -f "$0")")
+parent_dir=$(realpath "${script_dir}/../..")
+cidir="${parent_dir}/ci"
+source "${cidir}/lib.sh"
+
+cargo_deny_file="${script_dir}/action.yaml"
+
+cat cargo-deny-skeleton.yaml.in > "${cargo_deny_file}"
+
+changed_files_status=$(run_get_pr_changed_file_details)
+changed_files_status=$(echo "$changed_files_status" | grep "Cargo\.toml$" || true)
+changed_files=$(echo "$changed_files_status" | awk '{print $NF}' || true)
+
+if [ -z "$changed_files" ]; then
+  cat >> "${cargo_deny_file}" << EOF
+    - run: echo "No Cargo.toml files to check"
+      shell: bash
+EOF
+fi
+
+for path in $changed_files
+do
+    cat >> "${cargo_deny_file}" << EOF
+
+    - name: ${path}
+      continue-on-error: true
+      shell: bash
+      run: |
+        pushd $(dirname ${path})
+        cargo deny check
+        popd
+EOF
+done
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,33 @@
+targets = [
+    { triple = "x86_64-apple-darwin" },
+    { triple = "x86_64-unknown-linux-gnu" },
+    { triple = "x86_64-unknown-linux-musl" },
+]
+
+[advisories]
+vulnerability = "deny"
+unsound = "deny"
+unmaintained = "deny"
+ignore = ["RUSTSEC-2020-0071"]
+
+[bans]
+multiple-versions = "allow"
+deny = [
+    { name = "cmake" },
+    { name = "openssl-sys" },
+]
+
+[licenses]
+unlicensed = "deny"
+allow-osi-fsf-free = "neither"
+copyleft = "allow"
+# We want really high confidence when inferring licenses from text
+confidence-threshold = 0.93
+allow = ["0BSD", "Apache-2.0", "BSD-2-Clause", "BSD-3-Clause", "CC0-1.0", "ISC", "MIT", "MPL-2.0"]
+private = { ignore = true}
+
+exceptions = []
+
+[sources]
+unknown-registry = "allow"
+unknown-git = "allow"
--- a/.github/cargo-deny-composite-action/cargo-deny-skeleton.yaml.in
+++ b/.github/cargo-deny-composite-action/cargo-deny-skeleton.yaml.in
@@ -1,0 +1,30 @@
+#
+# Copyright (c) 2022 Red Hat
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+name: 'Cargo Crates Check'
+description: 'Checks every Cargo.toml file using cargo-deny'
+
+env:
+  CARGO_TERM_COLOR: always
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: nightly 
+        override: true
+
+    - name: Cache
+      uses: Swatinem/rust-cache@v2
+
+    - name: Install Cargo deny
+      shell: bash
+      run: |
+        which cargo
+        cargo install --locked cargo-deny || true
--- a/.github/workflows/cargo-deny-runner.yaml
+++ b/.github/workflows/cargo-deny-runner.yaml
@@ -1,0 +1,19 @@
+name: Cargo Crates Check Runner
+on: [pull_request]
+jobs:
+  cargo-deny-runner:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
+        uses: actions/checkout@v3
+      - name: Generate Action
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
+        run: bash cargo-deny-generator.sh
+        working-directory: ./.github/cargo-deny-composite-action/
+        env:
+          GOPATH: ${{ runner.workspace }}/kata-containers
+      - name: Run Action
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
+        uses: ./.github/cargo-deny-composite-action
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -54,3 +54,13 @@ run_docs_url_alive_check()
 	git fetch -a
 	bash "$tests_repo_dir/.ci/static-checks.sh" --docs --all "github.com/kata-containers/kata-containers"
 }
+
+run_get_pr_changed_file_details()
+{
+	clone_tests_repo
+	# Make sure we have the targeting branch
+	git remote set-branches --add origin "${branch}"
+	git fetch -a
+	source "$tests_repo_dir/.ci/lib.sh"
+	get_pr_changed_file_details
+}
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -18,6 +18,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "ansi_term"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -180,14 +189,16 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.19"
+version = "0.4.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
+checksum = "bfd4d1b31faaa3a89d7934dbded3111da0d2ef28e3ebccdb4f0179f5929d1ef1"
 dependencies = [
- "libc",
+ "iana-time-zone",
+ "js-sys",
  "num-integer",
  "num-traits",
  "time 0.1.44",
+ "wasm-bindgen",
  "winapi",
 ]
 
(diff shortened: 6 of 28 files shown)


bookinabox:
Force push was just to get rid of the action.yaml since it gets generated on run anyways.

bookinabox:
I guess one more issue I thought of last night is that we may want to consider is that cargo-deny does take a little while to run, especially with so many different Cargo.toml files. I'm not sure if we want to run this for every pull request as to not waste resources.

I can try to make it only run if the cargo files are updated if we think this is a concern.

wainersm:
Hi @bookinabox ,

First of all, I found the way you solve the lack of "for" in github actions very creative.

I never tried to generate and run an action on-the-fly so it will take a few more cycles for me to completely review it. So far I got two comments:
 * As you can see the "Cargo Crates Check Runner" actions failed. It seems they miss a configuration file somewhere;
 * I liked that you fixed the errors found by cargo deny alongside adding the action. But it is all in one commit, this is not so good. I'd like that you break on multiple commits, one for each type of error fixed.

bookinabox:
I'll look into the failed tests and break it up. Thanks!

bookinabox:
Looks like the checkout is having an issue in post. Probably due to having multiple checkout actions.

Not sure why the advisories is taking so long. Not sure why the run time is so long.

bookinabox:
That _should_ fix the issues regarding checkout. Additionally, cargo-deny found another vulnerability, so I had to modify that as well. Fixed the licenses, so it should pass the static checks now, fingers-crossed.

deny.toml
```diff
@@ -0,0 +1,30 @@
+targets = [
+    { triple = "x86_64-unknown-linux-gnu" },
+    { triple = "x86_64-unknown-linux-musl" },
+    { triple = "x86_64-apple-darwin" },
+    { triple = "x86_64-pc-windows-msvc" },
```
wainersm, line 5: afaik we don't support windows.

wainersm:
@bookinabox I ran the action a couple of times and I didn't see the timeout happening again.

wainersm:
@kata-containers/agent there is a couple of changes on Cargo.[toml|lock] for the agent on this pull request. Would you mind to review?

deny.toml
bookinabox: Okay. I based this off the default cargo-deny deny.toml. I'll remove this.

wainersm (approved):
LGTM. Thanks @bookinabox !

.github/cargo-deny-composite-action/cargo-deny-generator.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+script_dir=$(dirname "$(readlink -f "$0")")
+parent_dir=$(realpath "${script_dir}/../..")
+cidir="${parent_dir}/ci"
+source "${cidir}/lib.sh"
+
+temp_checkout_dir="./cargo-deny-action-copy"
```
jodh-intel, line 13: Nit: unused?

deny.toml
```diff
@@ -0,0 +1,29 @@
+targets = [
+    { triple = "x86_64-unknown-linux-gnu" },
+    { triple = "x86_64-unknown-linux-musl" },
+    { triple = "x86_64-apple-darwin" },
```
jodh-intel, line 4: Nit: unsorted list.

deny.toml
```diff
@@ -0,0 +1,29 @@
...
+vulnerability = "deny"
+unsound = "deny"
+unmaintained = "deny"
+ignore = ["RUSTSEC-2020-0071"]
+
+[bans]
+multiple-versions = "allow"
+deny = [
+    { name = "openssl-sys" },
+    { name = "cmake" },
```
jodh-intel, line 17: Nit: unsorted list.

jodh-intel (approved):
Thanks @bookinabox. A few nits but overall...

lgtm

jodh-intel:
@bookinabox - please can you rebase to resolve the conflicts on this PR?

bookinabox:
I rebased this quickly on Friday and didn't look too closely. I might have messed up the rebasing, so I will try to quickly fix this.

bookinabox:
> hi @bookinabox , as you rebased this PR more issues were found with recent changes on rust code. It will need to fix those issues then :( Hopefully it seems mere upgrades to crate versions will work out.

Oops, should have caught that. Will look into right now.

bookinabox:
@wainersm @jodh-intel _Should_ be done now. Sorry about the mess above. I was previously using nekto/act to test stuff locally, but that suddenly stopped working correctly and I could not figure out why.

a summary of what I ended up changing:

- Added a bunch of acceptable licenses to the list
- Allowed copyleft licenses so MLP-2.0 is not caught (I'm not sure why the default deny.toml has this disabled, but I did not see anything wrong with enabling it and could not find a way to easily work around it)
- Added if: {{ always() }} to each step in the generated action to replace the continue-on-error that I unknowingly broke after I made this a composite action. (Doing this earlier would have saved a lot of extra commit changes because it was not checking all the Cargo.tomls after error :( )
- Added Apache-2.0 license to a bunch of Cargo.tomls

Some things I changed, but I'm afraid might affect other things
- Bumped sha2 from =0.9.3 to ^0.9.3 to fix an error. I couldn't figure out a reason it _needed_ to be 0.9.3 and cargo-deny was complaining it was using the deprecated repository cpuid-bool. If this is a problem, reverting it shouldn't cause any issues as cpuid-bool was mainly deprecated due to a name change to cpufeatures instead (There was not security error or anything, the action was just complaining about it because it noticed cpuid-bool was marked as deprecated)
- Bumped nix in a couple places due to https://rustsec.org/advisories/RUSTSEC-2021-0119. To do this I did change libseccomp from 0.1.3 to 0.2.3 (though it was marked as optional). Looking at the release [logs](https://github.com/libseccomp-rs/libseccomp-rs/releases/tag/v0.2.0), I don't think any of the incompatible changes in the version just were in the project.

It seems to pass in my local fork's PR, so hopefully everything is smooth from here. Unfortunately, with all the new Cargo.toml files, it does take quite a long time to run now. Again, this shouldn't be an issue for future commits that aren't modifying close to all of them as it _should_ only be checking changed Cargo.toml files. Sorry about the wall of text, I hope its readable.

bookinabox:
Hmm, that's annoying. It was working on [my fork yesterday](https://github.com/bookinabox/kata-containers/runs/7869927938?check_suite_focus=true). I'll re-run it again, but since the reported error https://rustsec.org/advisories/RUSTSEC-2022-0004 is not new, I wonder why it was not caught initially. Will have to look into it.

wainersm:
> Hmm, that's annoying. It was working on [my fork yesterday](https://github.com/bookinabox/kata-containers/runs/7869927938?check_suite_focus=true). I'll re-run it again, but since the reported errors
> 
>     * https://rustsec.org/advisories/RUSTSEC-2022-0004
> 
>     * https://rustsec.org/advisories/RUSTSEC-2020-0016
> 
>     * https://rustsec.org/advisories/RUSTSEC-2022-0019
> 
>     * https://rustsec.org/advisories/RUSTSEC-2022-0041
>       ,  are not new, I wonder why it was not caught initially. Will have to look into it
> 
> 
> Edit: Oh, it's because main keeps getting updated and the changed files are the same Cargo.tomls :(, which means more rebasing I guess . Will have to think about this for a bit. In theory, if cargo-deny runs for everyone, this should not be an issue though.

You should manually run cargo-deny against all Cargo.toml then clean up all the caught errors. It is tough I know.

As the security database is kept updated, running cargo-deny for only the changed cargo.toml is not so effective. Likely to have situations where a file is not touched for a long time and, when it does, errors are caught; and worse, cargo deny will block a pull request due to changes not related. I can see two solutions:

1. run cargo-deny for every cargo.toml regardless if it changed or not. But again, it can block a PR with unrelated changes.
2. run cargo-deny as daily job. Someone will need to monitor that job and/or publish the report to Slack in case of errors and hope the community will fix it.

wainersm:
@kata-containers/architecture-committee we appreciate your comments on this PR, specially on the cargo deny profile (https://github.com/kata-containers/kata-containers/pull/4719/commits/2a6358db5174b21bcfa909e74b51ce09ec8af609#diff-1040309c64844eb1b6b63d8fd67938adbf9461f1b3c61f12cf738f064a02d3de) where it has a list of licenses that are/aren't accepted for the dependent crates.

bookinabox:
>  As the security database is kept updated, running cargo-deny for only the changed cargo.toml is not so effective. Likely to have situations where a file is not touched for a long time and, when it does, errors are caught; and worse, cargo deny will block a pull request due to changes not related. I can see two solutions:

What do you mean by security database?

I don't see why once its merged in it would not work. In theory, if starting from this PR everything has cargo-deny enforced, there should rarely be any side effects from other PRs bleeding into main. I guess a potential issue could be someone branching off an old version and the merge conflict leads them to change crate versions.

bookinabox:
@wainersm and I walked a bit today.

Here are some points that we might want to discuss:
- It looks like cargo-deny could potentially be annoying as we can see above. The frequency in which weird conflicts and advisories appear should be reduced significantly after everyone is using cargo-deny, but there still is the possibility that new security bugs are reported, which could affect old commits.
- Additionally, cargo-deny could potentially take a while to run, which with the number of PRs might waste a lot of server-time/ people's time.
- Not all advisories are really that impactful. I've been trying to fix every single one for this PR, but realistically, many of them are edge cases which should not cause any issues. Currently, I have it fail on any failed advisories, but it may be better to strategically ignore some.

Things we've considered:
1. Instead of running for every PR, we run this on a schedule and compile a list of errors that can be addressed. This would then be posted somewhere (Slack?)
2. We run this only when making a new version. One issue brought up for this was whether new bugs introduced by resolving these conflicts will be caught if we only run it at the end.
3. We could leave it for every PR and make advisories non-mandatory. This might lead to people just ignoring all of them though.

Any input to this would be appreciated.

bookinabox:
Regarding why cargo-deny is taking so long on kata-containers vs dragonflyoss/image-service:

First, I did go a roundabout way of getting all the Cargo.toml files scanned. Normally when cargo-deny runs, it works by scanning either the root directory or a provided directory for a cargo.toml file. However, the kata project does not have a single "root" cargo file. Instead, it has a bunch of components, each with their own files, each file potential having its own workspace. I made a composite action to find all the Cargo.toml files (that were modified)  and then run the action on each one individually. In contrast, the image-service does have a "root" Cargo file that lists all other cargo.toml files in its workspace.  Rust also does not support nested workspaces as far as I can tell, which prevents me from easily just merging all our exisitng files under one large global workspace. There is overhead here being incurred as well as likely some redundancy (files belonging to a workspace may be double counted, but I cannot think of a better way to do this).

Secondly, there are currently 30 Cargo.tomls in kata-containers, 20 of which are being caught by cargo-deny in my PR right now. dragonflyoss/image-service has 9. This probably has some effect as well.

Thirdly, I did split the advisories and license checks into two separate jobs. I did this when I started because it was the recommended workflow on the cargo-deny-action page, but is unnecessary if we are just keeping everything as non-blocking anyways.

I can change the last one and see how it affects performance. The first issue I struggled with for a while and did not see any other obvious solution, so if anyone has any suggestions that would be appreciated. The
 second point is basically unavoidable.

bookinabox:
Made it non-blocking. Was not able to get it to be any faster though.

Also, I've stopped trying to update dependencies. We may even want to consider not merging in the dependency changes I already made because there were quite a lot and the only testing I did was through cargo test.

Other potential points of consideration might be the rules in the deny.toml. For example, I added licenses that were being used that seemed reasonable, but if someone else wants to take a look that would be appreciated. 

As mentioned in the meeting, a previous cargo audit ticket was blocked by a chrono issue. In the deny.toml I just have it ignore the issue.

I'm sure there's a more efficient way to find and test all the cargo files that I'm not seeing, but I think this works for now.

bookinabox:
After trying to speed it up some more, I've come to the breakthrough that running cargo-deny by itself rather than the cargo-deny-action is significantly faster. There must be some large overhead involved with running the action (perhaps its re-downloading the action each time?). I could not figure out a way to "cache" the action, so I'm going to push my change to run cargo-deny manually in a few mins, because 20 mins really is unreasonable.

bookinabox:
There were some weird issues regarding dbs-arch not compiling, which is odd given that the hash/version was the same as versions that were compiling on successful runs. I had it update the lock file. Doubt it will make a difference, but not sure how else to approach debugging this.

wainersm:
> There were some weird issues regarding dbs-arch not compiling, which is odd given that the hash/version was the same as versions that were compiling on successful runs. I had it update the lock file. Doubt it will make a difference, but not sure how else to approach debugging this.

Still failing. Could you please rebase your branch with main?

bookinabox:
@wainersm The static checks were failing due to some dependency being changed. Instead of pinpointing each one, I re-did the dependency change commit and only changed things that were simple to change leaving the rest because this is no longer a blocking action. I ran cargo test on each and they seem to work.

PR #4743, opened 2022-07-26
mount: support checking multiple kinds of block device driver

### LINKED ISSUE #4714: Enhance supporting multiple kinds of host block device
Kata Containers supports using block device as container's rootfs, and it is the only way for firecracker which needs to setup a devmapper snapshotter as [guided](https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/how-to-use-kata-containers-with-firecracker.md).
As far as I see, device mapper is the only supported storage driver so far, which seems limiting. From [container.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/container.go#L1235) and [mount.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/mount.go#L209), it only check if the driver is device mapper and return if not. 
Actually, Kata Containers can work well with raw image on loop device or [overlaybd](https://github.com/containerd/overlaybd#overlaybd) on tcmu device. It is necessary to enhance supporting of multiple kinds of host block device.

yuchen0cc:
Device mapper is the only supported block device driver so far, which seems limiting. Kata Containers can work well with other block devices. It is necessary to enhance supporting of multiple kinds of host block device.

Fixes #4714

Signed-off-by: yuchen.cc <yuchen.cc@alibaba-inc.com>

### DIFF
--- a/src/runtime/virtcontainers/mount.go
+++ b/src/runtime/virtcontainers/mount.go
@@ -194,14 +194,14 @@ func getDeviceForPath(path string) (device, error) {
 	return dev, nil
 }
 
-var blockFormatTemplate = "/sys/dev/block/%d:%d/dm"
+var blockFormatTemplate = "/sys/dev/block/%d:%d/"
 
-var checkStorageDriver = isDeviceMapper
+var checkStorageDriver = isBlockDevice
 
-// isDeviceMapper checks if the device with the major and minor numbers is a devicemapper block device
-func isDeviceMapper(major, minor int) (bool, error) {
+// isBlockDevice checks if the device with the major and minor numbers is a block device
+func isBlockDevice(major, minor int) (bool, error) {
 
-	//Check if /sys/dev/block/${major}-${minor}/dm exists
+	//Check if /sys/dev/block/${major}-${minor}/ exists
 	sysPath := fmt.Sprintf(blockFormatTemplate, major, minor)
 
 	_, err := os.Stat(sysPath)
@@ -209,9 +209,9 @@ func isDeviceMapper(major, minor int) (bool, error) {
 		return true, nil
 	} else if os.IsNotExist(err) {
 		return false, nil
+	} else {
+		return false, err
 	}
-
-	return false, err
 }
 
 const mountPerm = os.FileMode(0755)
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -1319,12 +1319,12 @@ func (c *Container) hotplugDrive(ctx context.Context) error {
 		"mount-point":  dev.mountPoint,
 	}).Info("device details")
 
-	isDM, err := checkStorageDriver(dev.major, dev.minor)
+	isBD, err := checkStorageDriver(dev.major, dev.minor)
 	if err != nil {
 		return err
 	}
 
-	if !isDM {
+	if !isBD {
 		return nil
 	}
 
--- a/src/runtime/virtcontainers/mount_linux_test.go
+++ b/src/runtime/virtcontainers/mount_linux_test.go
@@ -305,20 +305,31 @@ func TestGetDeviceForPathValidMount(t *testing.T) {
 	assert.Equal(dev.mountPoint, expected)
 }
 
-func TestIsDeviceMapper(t *testing.T) {
+func TestIsBlockDevice(t *testing.T) {
 	assert := assert.New(t)
 
 	// known major, minor for /dev/tty
 	major := 5
 	minor := 0
 
-	isDM, err := isDeviceMapper(major, minor)
+	isBD, err := isBlockDevice(major, minor)
 	assert.NoError(err)
-	assert.False(isDM)
+	assert.False(isBD)
 
 	// fake the block device format
+	blockFormatTemplateOld := blockFormatTemplate
+	defer func() {
+		blockFormatTemplate = blockFormatTemplateOld
+	}()
+
 	blockFormatTemplate = "/sys/dev/char/%d:%d"
-	isDM, err = isDeviceMapper(major, minor)
+	isBD, err = isBlockDevice(major, minor)
 	assert.NoError(err)
-	assert.True(isDM)
+	assert.True(isBD)
+
+	// invalid template
+	blockFormatTemplate = "\000/sys/dev/char/%d:%d"
+	isBD, err = isBlockDevice(major, minor)
+	assert.Error(err)
+	assert.False(isBD)
 }


yuchen0cc:
Please tell me the porting labels to be applied.

src/runtime/virtcontainers/mount.go
```diff
@@ -208,7 +208,7 @@ func getDeviceForPath(path string) (device, error) {
 
 var blockFormatTemplate = "/sys/dev/block/%d:%d/dm"
 
-var checkStorageDriver = isDeviceMapper
+var checkStorageDriver = checkBlockDevice
```
bergwolf, line 211: `isDeviceMapper` is left over but it's a piece of dead code now? And the new code is not unit-tested as the code pieces.

bergwolf (commented):
A more general question: the DM device check is a workaround for the device mapper snapshotter. Please explain a bit why we need to check for the extra block device types and how they are supposed to work with Kata Containers. Thanks!

src/runtime/virtcontainers/mount.go
```diff
@@ -226,6 +226,40 @@ func isDeviceMapper(major, minor int) (bool, error) {
...
+		return true, nil
+	} else if os.IsNotExist(err) {
+		return false, nil
+	}
+
+	return false, err
+}
+
+func checkBlockDevice(major, minor int) (bool, error) {
+	var lastErr error = nil
```
shuaichang, line 249: `= nil` seems to be unnecessary

src/runtime/virtcontainers/mount.go
```diff
@@ -226,6 +226,40 @@ func isDeviceMapper(major, minor int) (bool, error) {
...
+	for dev, template := range blockFormatTemplates {
+		isBD, err := isBlockDevice(major, minor, template)
+		if isBD {
+			mountLogger().Infof("Found storage driver for %s", dev)
```
shuaichang, line 253: This log seems to be a little off from the context, can we log "device % is identified as a block device"?

src/runtime/virtcontainers/mount.go
```diff
@@ -226,6 +226,40 @@ func isDeviceMapper(major, minor int) (bool, error) {
...
+	"loop":  "/sys/dev/block/%d:%d/loop",
+}
+
+func isBlockDevice(major, minor int, formatTemplate string) (bool, error) {
+	sysPath := fmt.Sprintf(formatTemplate, major, minor)
+
+	_, err := os.Stat(sysPath)
+	if err == nil {
+		return true, nil
+	} else if os.IsNotExist(err) {
```
shuaichang, line 241: I think we can remove the `else if`, or to add the last return into an `else` block for consistency

```
if err == nil {
    return true, nil
}

if os.IsNotExist(err) {
    return false, nil
}

return false, err
```

src/runtime/virtcontainers/mount.go
shuaichang: +1 for removing the isDeviceMapper code if not used

src/runtime/virtcontainers/mount.go
```diff
@@ -226,6 +226,40 @@ func isDeviceMapper(major, minor int) (bool, error) {
...
+
+func checkBlockDevice(major, minor int) (bool, error) {
+	var lastErr error = nil
+	for dev, template := range blockFormatTemplates {
+		isBD, err := isBlockDevice(major, minor, template)
+		if isBD {
+			mountLogger().Infof("Found storage driver for %s", dev)
+			return isBD, err
+		} else if err != nil {
+			lastErr = err
```
shuaichang, line 256: Can we make this logic easier to understand like the following? I think if any error occurred while checking can just throw the error without continuing.

```
isBD, err := isBlockDevice(major, minor, template)
if err != nil {
    return 
}
```

shuaichang (changes requested):
Thanks for making the change, very useful to extend block device support outside of device mapper!

yuchen0cc:
@bergwolf Thanks for commenting! 
Kata Containers has already supported using block device as container's rootfs, and it is the only way for firecracker.
Device mapper is the only supported block device driver so far, which seems limiting. Actually, Kata Containers can work well with raw image on loop device or [overlaybd](https://github.com/containerd/overlaybd#overlaybd) on tcmu device. It is necessary to enhance supporting of multiple kinds of host block device driver.

I will add unit test and remove unused code.

yuchen0cc:
What does "add-pr-size-label" failed mean ?

src/runtime/virtcontainers/mount.go
```diff
@@ -226,6 +226,40 @@ func isDeviceMapper(major, minor int) (bool, error) {
 	return false, err
 }
 
+var blockFormatTemplates = map[string]string{
```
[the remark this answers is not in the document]
yuchen0cc: blockFormatTemplates is consist with old code.

src/runtime/virtcontainers/mount.go
yuchen0cc: done

src/runtime/virtcontainers/mount_test.go
```diff
@@ -205,22 +205,24 @@ func TestGetDeviceForPathBindMount(t *testing.T) {
...
-	isDM, err := isDeviceMapper(major, minor)
+	isBD, err := checkBlockDevice(major, minor)
 	assert.NoError(err)
-	assert.False(isDM)
+	assert.False(isBD)
 
 	// fake the block device format
-	blockFormatTemplate = "/sys/dev/char/%d:%d"
-	isDM, err = isDeviceMapper(major, minor)
+	blockFormatTemplates = map[string]string{
```
shuaichang, line 220: Better to reset `blockFormatTemplates` so in case other case will not be impacted, I think the old code probably missed this part

```
	blockFormatTemplatesOld := blockFormatTemplates
	defer func() {
		blockFormatTemplates = blockFormatTemplatesOld
	}()
	blockFormatTemplates = map[string]string{
		"/dev/tty": "/sys/dev/char/%d:%d",
	}
```

src/runtime/virtcontainers/mount.go
```diff
@@ -206,24 +206,40 @@ func getDeviceForPath(path string) (device, error) {
...
+	} else {
+		return false, err
+	}
+}
+
+func checkBlockDevice(major, minor int) (bool, error) {
+	for dev, template := range blockFormatTemplates {
+		isBD, err := isBlockDevice(major, minor, template)
+		if err != nil {
+			return false, err
```
shuaichang, line 234: Can we add UTs to cover all the branches for the newly added code?

src/runtime/virtcontainers/mount_test.go
yuchen0cc: It's a good way, I have done so.

src/runtime/virtcontainers/mount.go
yuchen0cc: The unchecked branch is `err != nil` which only happens when `isBlockDevice` return error. That happens when `os.Stat` fails and not because of `os.IsNotExist`. 
I really don't know when it happens. Do you have any idea?

sazzy4o: I think the easiest way to accomplish this here is to use a path with an invalid character (the null character `\000` will work unix and windows):
```
fmt.Println("Example: Invalid character")
_, err := os.Stat("\000path")
fmt.Println(err)
```

You can override the `blockFormatTemplates` like you did the tests below to get a path with `\000` in it

(Permissions errors should also work, but might be a little hard to setup test for)

Also, if you need help getting this PR to the finish line, please let me know

yuchen0cc: @sazzy4o I have supplemented the test as you suggested.
It's really grateful that you can help finishing this PR.

sazzy4o:
@shuaichang Could you please have another look at reviewing this PR?

shuaichang (approved):
LGTM

sazzy4o:
@yuchen0cc Could you please merge `kata-containers:main` into your branch, I think one of the PR checks is failing because the branch is not up to date?

yuchen0cc:
> @yuchen0cc Could you please merge `kata-containers:main` into your branch, I think one of the PR checks is failing because the branch is not up to date?
OK, I've rebase it by mainstream.

sazzy4o:
@bergwolf Could you please review this PR?

sazzy4o:
@yuchen0cc It looks like one of the checks is failing because [a749dd1](https://github.com/kata-containers/kata-containers/commit/a749dd1ada39aa8a088142969881eff7211db6ad) does not have a commit message body

I think updating it to something like this should work:
```
mount_test: supplement test for checking block device

Adds test for handling os.Stat error

Signed-off-by: yuchen.cc <yuchen.cc@alibaba-inc.com>
```

See [here](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#supplementary-patch) for more info on expected commit message format

I hope this is the last change. If you want me to take over changes to this PR, please let me know (I just need permission to push to your fork)

yuchen0cc:
@sazzy4o sorry to forget this checking, I've fixed it.

src/runtime/virtcontainers/mount.go
```diff
@@ -194,24 +194,40 @@ func getDeviceForPath(path string) (device, error) {
...
+var blockFormatTemplates = map[string]string{
+	"dm":    "/sys/dev/block/%d:%d/dm",
+	"block": "/sys/dev/block/%d:%d/device/block",
+	"loop":  "/sys/dev/block/%d:%d/loop",
+}
 
-	//Check if /sys/dev/block/${major}-${minor}/dm exists
-	sysPath := fmt.Sprintf(blockFormatTemplate, major, minor)
+func isBlockDevice(major, minor int, formatTemplate string) (bool, error) {
+	sysPath := fmt.Sprintf(formatTemplate, major, minor)
```
bergwolf, line 206: It does not seem to work for nvme devices. Why not just check for `/sys/dev/block/<major:minor>/` and forget about the specific block device types?

bergwolf (commented):
It looks useful to enable overlaybd for Kata. Could you add a howto document in the `docs/how-to/` directory?

src/runtime/virtcontainers/mount.go
yuchen0cc: I have actually tested the above 3 block types with kata, and I'm not sure kata is compatible with all block types.
Just to be cautious, I make a minimal supportation, and allow for extensions.

bergwolf: All block devices are equal in terms of virtio-block backend. So it doesn't matter where it comes from.

yuchen0cc: OK, it seems reasonable.

yuchen0cc:
As suggested by @bergwolf , I think there is no longer need to check multiple templates (using map and loop). 
So I squash the previous 2 commit and make it a minor change.
@shuaichang @sazzy4o @bergwolf Could you please take a look at this.

bergwolf (approved):
lgtm!

yuchen0cc:
TestIsDeviceMapper has  been moved from [mount_test.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/mount_test.go) to [mount_linux_test](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/mount_linux_test.go)

bergwolf:
Please test the code locally before sending it for review.
```
# github.com/kata-containers/kata-containers/src/runtime/virtcontainers [github.com/kata-containers/kata-containers/src/runtime/virtcontainers.test]
virtcontainers/mount_linux_test.go:315:15: undefined: isDeviceMapper
virtcontainers/mount_linux_test.go:[32](https://github.com/kata-containers/kata-containers/actions/runs/7047880519/job/19205798649?pr=4743#step:9:33)1:14: undefined: isDeviceMapper
```

sazzy4o:
I think that code looks good, but I'll try running locally (I think I should have some time to test locally tomorrow)

sazzy4o:
Tested locally with `overlaybd` and it works as expected:
```
$ sudo nerdctl run --cgroup-manager=cgroupfs --runtime io.containerd.run.kata.v2 --snapshotter=overlaybd --rm -it registry.hub.docker.com/overlaybd/redis:6.2.1_obd /bin/sh
# df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda         63G  107M   60G   1% /
tmpfs            64M     0   64M   0% /dev
none            322G  179G  127G  59% /data
shm             996M     0  996M   0% /dev/shm
tmpfs           996M     0  996M   0% /sys/fs/cgroup
kataShared      322G  179G  127G  59% /etc/hosts
# 
```
Notes:
- Tested with cloud-hypervisor
- overlaybd installed via instructions [here](https://github.com/containerd/accelerated-container-image/blob/main/docs/QUICKSTART.md)
- **Important**: `sandbox_cgroup_only=true` must be set in your `configuration.toml`

sazzy4o (approved):
Code looks good and works when testing locally

jepio (approved):
looks reasonable

Issue #4747, opened 2022-07-26
exec a backgroud process without tty would hang


lifupan:
crictl -i <container> sh -c 'nohup tail -f /dev/null &'  
this command would hang there.

PR #4748, opened 2022-07-26
agent: fix the issue of exec hang with a backgroud process

lifupan:
When run a exec process in backgroud without tty, the
exec will hang and didn't terminated.

For example:

crictl -i <container id> sh -c 'nohup tail -f /dev/null &'

Fixes: #4747


liubin:
We need an integration test to ensure that no regression occurs in the future, or has there been one already exist?

Tim-Zhang (approved):
LGTM, thanks @lifupan

liubin (approved):

cmaf:
@lifupan this could probably use a backport if you are able to add one

Issue #4752, opened 2022-07-26
Bug: stream fds were closed twice 


Tim-Zhang:
https://github.com/kata-containers/kata-containers/blob/e764a726abbbec88b9e1ccecdae78aa7a045f245/src/agent/rustjail/src/process.rs#L30-L31

This would cause EBADF error or it would close other fds by mistake.

PR #4753, opened 2022-07-26
agent: Fix stream fd's double close

Tim-Zhang:
The fd would be closed on Pipestream's dropping and we should
not close it agian.

Fixes: #4752


liubin (approved):

bergwolf (approved):
lgtm!

Issue #4749, opened 2022-07-26
Rollback if case of do_create_container failed


liubin:
Don't know what's the pod definition of @Champ-Goblem 's , for test purposes, I specified a non-exist command as the container's entry point, and the agent will only get `create_container` be called, the sequence is:

|Shim|             agent|
|---|---|
|->Start ||
||->create_container|
|->Delete ||
|->Shutdown ||

So maybe we need to add a rollback in `create_container` in the agent to do some clear work, for example:

https://github.com/kata-containers/kata-containers/blob/ac91fb7a126238b595afa59c17fbb82a12c4f5ad/src/agent/src/rpc.rs#L226-L229

Add some error-recover code here.

My test yaml file:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cpu-test
spec:
  terminationGracePeriodSeconds: 0
  runtimeClassName: kata
  containers:
  - name: first-cpu-container
    image: busybox:latest 
    command:
        - sleepp
        - "300"
    resources:
      limits:
        cpu: "1"
      requests:
        cpu: "500m"

```

After minutes, I will get the same result as you:

```bash
# in host
$ kubectl get pod -w
NAME       READY   STATUS             RESTARTS        AGE
cpu-test   0/1     CrashLoopBackOff    7 (12s ago)     11m

# in guest
# mount | grep rootfs | wc -l
9
```

_Originally posted by @liubin in https://github.com/kata-containers/kata-containers/pull/4716#discussion_r927480850_

Champ-Goblem:
@liubin just an FYI, this doesn't seem to have fixed #4716, still seeing the memory usage increase for failing containers and the mounts are still persisting across container restarts

PR #4754, opened 2022-07-26
agent: do some rollback works if case of do_create_container failed

liubin:
In some cases do_create_container may return an error, mostly due to
`container.start(process)` call. This commit will do some rollback
works if this function failed.

Fixes: #4749


fidencio (approved):
lgtm, thanks @liubin!

liubin:
PTAL @jodh-intel @lifupan @bergwolf

jodh-intel (approved):
Thanks @liubin.

lgtm

Issue #4761, opened 2022-07-28
CC | Support cryptsetup required kernel config and userspace command line tools


arronwy:
CoCo runtime will need support trust storage and measured boot image:
https://github.com/confidential-containers/documentation/issues/39
https://github.com/confidential-containers/documentation/issues/40 

Both require the kernel support dm-crypt/dm-integrity/dm-verity features and related crypto config.
Also it requires the usespace cryptsetup command line tools.

PR #4762, opened 2022-07-28
CC | Add cryptsetup support in Guest kernel and rootfs

arronwy:
Add required kernel config for dm-crypt/dm-integrity/dm-verity
and related crypto config.

Add userspace command line tools for disk encryption support
and ext4 file system utilities.

Fixes: #4761

Signed-off-by: Arron Wang <arron.wang@intel.com>


fidencio:
@arronwy, please, also bump the kernel configuration version: https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kernel/kata_config_version

arronwy:
> @arronwy, please, also bump the kernel configuration version: https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kernel/kata_config_version

Thanks @fidencio , updated now :-)

tools/osbuilder/rootfs-builder/ubuntu/config.sh
```diff
@@ -5,7 +5,7 @@
 OS_NAME=ubuntu
 # This should be Ubuntu's code name, e.g. "focal" (Focal Fossa) for 20.04
 OS_VERSION=${OS_VERSION:-focal}
-PACKAGES="chrony iptables"
+PACKAGES="chrony iptables cryptsetup-bin e2fsprogs"
```
wainersm, line 8: @arronwy @fidencio just thinking aloud...when we attempt to merge CCv0 -> main it will be a nightmare as we will need to search for all CC specific snippets of code. For example here, I don't believe we will include disk crypto utils for the non-CC image, won't we?

So I was wondering if we could start "marking" those places, or use `if` guards. While ago, when @stevenhorsman , @fidencio and I started the work to merge the branches, I introduced the `KATA_BUILD_CC` variable (see https://github.com/kata-containers/tests/pull/4675) for the purpose of `if KATA_BUILD_CC=yes then do_some_CC_specific_stuff` logic. Would make sense to start using that variable right way?

wdyt?

tools/packaging/kernel/configs/fragments/common/cryptsetup.conf
```diff
@@ -0,0 +1,43 @@
+CONFIG_MD=y
+# CONFIG_BLK_DEV_MD is not set
+# CONFIG_BCACHE is not set
+CONFIG_BLK_DEV_DM_BUILTIN=y
+CONFIG_BLK_DEV_DM=y
+# CONFIG_DM_DEBUG is not set
```
wainersm, line 6: Is there a reason to keep those commented options?

tools/packaging/kernel/configs/fragments/common/cryptsetup.conf
```diff
@@ -0,0 +1,43 @@
+CONFIG_MD=y
```
wainersm, line 1: I am unsure the `common` dir is best place for the `cryptsetup.conf` file as it will include dm stuffs for all kernels. Ignore the following idea if you think it is okay that.

I'm not so used to the kernel build scripts of kata, but looking at `build-kernel.sh` it seems the kernel for confidential guest is built with passing `-x tdx|sev` option and their specific configs live under `configs/fragments/x86_64`. There is also the `configs/fragments/build-type` dir which contain configs for when passing `-b` (build type), `-e` (experimental) or `-E` (arch-experimental). I'm wondering if putting our configs under `configs/fragments/build-type/*confidential_containers*` then setting the `build_type` variable when `-x tdx|sev` wouldn't keep it more organized.

tools/osbuilder/rootfs-builder/ubuntu/config.sh
arronwy: Yes, if we finally need merge CCv0 branch with main branch, we can do it as early as possible

tools/packaging/kernel/configs/fragments/common/cryptsetup.conf
arronwy: yes, this is auto generated, can be remove.

tools/packaging/kernel/configs/fragments/common/cryptsetup.conf
arronwy: I changed to use `KATA_BUILD_CC` to control the build for this feature, since we may need test these features without HW TEE. `cryptsetup.conf` is generic for different HW, then I put it under `common/confidential_containers` directory.

fidencio (approved):
lgtm, thanks @arronwy!

stevenhorsman:
I'll let @wainersm have the final say in the review here, but a concern I have here is in testing whether `KATA_BUILD_CC` variable flows through okay, or if one of the times that we run as `sudo` might lose the value (I might be wrong, but I seem to remember this happening). In the latest test run it just picks up the cached version, rather than rebuilding, so I can see the evidence either way.

tools/packaging/kernel/build-kernel.sh
```diff
@@ -247,6 +248,12 @@ get_kernel_frag_path() {
 		all_configs="${all_configs} ${conf_configs}"
 	fi
 
+	if [[ "${KATA_BUILD_CC}" == "yes" ]];then
+		info "Enabling config for confidential guest trust storage protection"
+		local cryptsetup_configs="$(ls ${common_path}/confidential_containers/cryptsetup.conf)"
+		all_configs="${all_configs} ${cryptsetup_configs}"
+	fi
```
fidencio, line 255: @arronwy, I would not take this path.

As we want this to be common to both TDX and SEV, I'd not put this under KATA_BUILD_CC, but rather in the `if` above, where we check for `conf_guest`.

By doing this we know for sure we have it enabled in both TDX and SEV cases, without the need to pass yet another environment variable.

```sh
diff --git a/tools/packaging/kernel/build-kernel.sh b/tools/packaging/kernel/build-kernel.sh
index 7090e28e5..6caddc5c1 100755
--- a/tools/packaging/kernel/build-kernel.sh
+++ b/tools/packaging/kernel/build-kernel.sh
@@ -243,15 +243,13 @@ get_kernel_frag_path() {
        fi
 
        if [[ "${conf_guest}" != "" ]];then
-               info "Enabling config for '${conf_guest}' confidential guest protection"
-               local conf_configs="$(ls ${arch_path}/${conf_guest}/*.conf)"
-               all_configs="${all_configs} ${conf_configs}"
-       fi
-
-       if [[ "${KATA_BUILD_CC}" == "yes" ]];then
                info "Enabling config for confidential guest trust storage protection"
                local cryptsetup_configs="$(ls ${common_path}/confidential_containers/cryptsetup.conf)"
                all_configs="${all_configs} ${cryptsetup_configs}"
+
+               info "Enabling config for '${conf_guest}' confidential guest protection"
+               local conf_configs="$(ls ${arch_path}/${conf_guest}/*.conf)"
+               all_configs="${all_configs} ${conf_configs}"
        fi
 
        info "Constructing config from fragments: ${config_path}"

``

**Review by fidencio — CHANGES_REQUESTED** [review]
@arronwy, I've approved too soon, sorry.

Please, take a look at the suggestion made below about where to add the common confidential configurations for the kernel build, and about passing down to the scripts the KATA_BUILD_CC variable.

File: tools/osbuilder/rootfs-builder/ubuntu/config.sh [hunk 5]
```diff
@@ -7,6 +7,7 @@ OS_NAME=ubuntu
 OS_VERSION=${OS_VERSION:-focal}
 PACKAGES="chrony iptables"
 [ "$AGENT_INIT" = no ] && PACKAGES+=" init"
+[ "$KATA_BUILD_CC" = yes ] && PACKAGES+=" cryptsetup-bin e2fsprogs"
```
fidencio, line 10: While adding the `KATA_BUILD_CC` here is fine, I'd like to see this propagated to the scripts we use to generate the image.

So, you'd need to adapt the following scripts to ensure we have `KATA_BUILD_CC` set.
* https://github.com/kata-containers/kata-containers/blob/81e065f7bdb741ef4bd2106c19067f795bbbdb06/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh#L47
* https://github.com/kata-containers/kata-containers/blob/81e065f7bdb741ef4bd2106c19067f795bbbdb06/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L118

tools/packaging/kernel/build-kernel.sh
wainersm: That's a good idea @fidencio. If we could avoid yet another envvar to control the script flow, the better.

wainersm:
> @arronwy, I've approved too soon, sorry.
> 

Fat finger? :)

tools/packaging/kernel/build-kernel.sh
arronwy: Thanks @fidencio fixed now :-)

tools/osbuilder/rootfs-builder/ubuntu/config.sh
arronwy: Thanks for the reminder, fixed now!

fidencio (approved):
lgtm, thanks @arronwy!

tools/packaging/kernel/configs/fragments/x86_64/crypto.conf
```diff
@@ -0,0 +1,3 @@
+# x86 cryptographic instructions to improve AES encryption and SHA256 hashing.
```
hbrueckner, line 1: @fidencio @arronwy How do we handle config options for non-x86 architectures? Add them here or separate commit in this PR or additional PR?

cc: @stevenhorsman @BbolroC

hbrueckner: @arronwy thanks. Looking on the `s390x` the are a couple of config options already in-place. I will check for those which has been added in the last 2y. So let me know whether to follow-up here or separately?

arronwy: Yes, like `arm` also already have some crypto accelerated config, I think we can also open an PR for `s390x` if needed

stevenhorsman:
It looks like the hugepages test has failed for the second time in the fedora test:
```
not ok 1 Hugepages
# (in test file k8s-hugepages.bats, line 52)
# INFO: k8s configured to use runtimeclass
# pod/hugepage-pod created
# error: timed out waiting for the condition on pods/hugepage-pod
```
There isn't anything that I can immediately spot in this PR that would effect this, unless the propagation of `KATA_BUILD_CC` has enabled something extra? I'll kick the tests off again and see if we get third time lucky
/retest-fedora

wainersm (approved):

PR #4772, opened 2022-07-29
Drop-in cfg files support in runtime-rs

pmores:
This PR adds handling of drop-in config file fragments to runtime-rs to match the same feature already implemented in golang runtime.  Behaviour of this Rust port should be the same as the Golang version so no additional documentation should be necessary.

The change is fairly non-intrusive, mostly adding new code, but the final commit does involve some refactoring of the existing code.  Admittedly there were multiple ways to achieve the goal of creating a single point that drop-in handling needs to be added to.  Being not entirely clear about the intended usage of `load_raw_from_file()` (a public function, however it doesn't seem to be called from anywhere currently) and `load()` (public as well but only called internally AFAICT) I just went for what seemed to be the simplest and least intrusive option.

### DIFF
--- a/src/libs/kata-types/src/config/drop_in.rs
+++ b/src/libs/kata-types/src/config/drop_in.rs
@@ -212,3 +212,171 @@ mod toml_tree_ops {
         }
     }
 }
+
+mod drop_in_directory_handling {
+    use crate::config::TomlConfig;
+    use std::fs;
+    use std::io::{self, Result};
+    use std::path::{Path, PathBuf};
+
+    fn get_dropin_dir_path(base_cfg_file_path: &Path) -> Result<PathBuf> {
+        let mut dropin_dir = base_cfg_file_path.to_path_buf();
+        if !dropin_dir.pop() {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "base cfg file path too short",
+            ));
+        }
+        dropin_dir.push("config.d");
+        Ok(dropin_dir)
+    }
+
+    fn update_from_dropin(base_config: &mut toml::Value, dropin_file: &fs::DirEntry) -> Result<()> {
+        if !dropin_file.file_type()?.is_file() {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                "drop-in cfg file can only be a regular file or a symlink",
+            ));
+        }
+        let dropin_contents = fs::read_to_string(&dropin_file.path())?;
+        let dropin_config: toml::Value = toml::from_str(&dropin_contents)?;
+        super::toml_tree_ops::merge(base_config, dropin_config);
+        Ok(())
+    }
+
+    fn update_from_dropins(base_config: &mut toml::Value, dropin_dir: &Path) -> Result<()> {
+        let dropin_files_iter = match fs::read_dir(dropin_dir) {
+            Ok(iter) => iter,
+            Err(err) => {
+                if err.kind() == io::ErrorKind::NotFound {
+                    return Ok(());
+                } else {
+                    return Err(err);
+                }
+            }
+        };
+
+        let mut dropin_files = dropin_files_iter.collect::<Result<Vec<_>>>()?;
+        dropin_files.sort_by_key(|direntry| direntry.file_name());
+        for dropin_file in &dropin_files {
+            update_from_dropin(base_config, dropin_file)?;
+        }
+        Ok(())
+    }
+
+    pub fn load(base_cfg_file_path: &Path) -> Result<TomlConfig> {
+        let base_toml_str = fs::read_to_string(&base_cfg_file_path)?;
+        let mut base_config: toml::Value = toml::from_str(&base_toml_str)?;
+        let dropin_dir = get_dropin_dir_path(base_cfg_file_path)?;
+
+        update_from_dropins(&mut base_config, &dropin_dir)?;
+
+        let config: TomlConfig = base_config.try_into()?;
+        Ok(config)
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+        use std::io::Write;
+
+        const BASE_CONFIG_DATA: &str = r#"
+            [hypervisor.qemu]
+            path = "/usr/bin/qemu-kvm"
+            default_bridges = 3
+            [runtime]
+            enable_debug = true
+            internetworking_model="tcfilter"
+        "#;
+
+        fn check_base_config(config: &TomlConfig) {
+            assert_eq!(
+                config.hypervisor["qemu"].path,
+                "/usr/bin/qemu-kvm".to_string()
+            );
+            assert_eq!(config.hypervisor["qemu"].device_info.default_bridges, 3);
+            assert!(config.runtime.debug);
+            assert_eq!(config.runtime.internetworking_model, "tcfilter".to_string());
+        }
+
+        fn create_file(path: &Path, contents: &[u8]) -> Result<()> {
+            fs::File::create(path)?.write_all(contents)
+        }
+
+        #[test]
+        fn test_no_dropins_dir() {
+            let tmpdir = tempfile::tempdir().unwrap();
+
+            let config_path = tmpdir.path().join("runtime.toml");
+            create_file(&config_path, BASE_CONFIG_DATA.as_bytes()).unwrap();
+
+            let config = load(&config_path).unwrap();
+            check_base_config(&config);
+        }
+
+        #[test]
+        fn test_no_dropins() {
+            let tmpdir = tempfile::tempdir().unwrap();
+
+            let config_path = tmpdir.path().join("runtime.toml");
+            create_file(&config_path, BASE_CONFIG_DATA.as_bytes()).unwrap();
+
+            let dropin_dir = tmpdir.path().join("config.d");
+            fs::create_dir(&dropin_dir).unwrap();
+
+            let config = load(&config_path).unwrap();
+            check_base_config(&config);
+        }
+
+        #[test]
+        fn test_dropins() {
+            let tmpdir = tempfile::tempdir().unwrap();
+
+            let dropin_data = r#"
+                [hypervisor.qemu]
+                default_vcpus = 2
+                default_bridges = 4
+                shared_fs = "virtio-fs"
+                [runtime]
+                sandbox_cgroup_only=true
+                internetworking_model="macvtap"
+                vfio_mode="guest-kernel"
+            "#;
+
+            let dropin_override_data = r#"
+                [hypervisor.qemu]
+                shared_fs = "virtio-9p"
+                [runtime]
+                vfio_mode="vfio"
+            "#;
+
+            let config_path = tmpdir.path().join("runtime.toml");
+            create_file(&config_path, BASE_CONFIG_DATA.as_bytes()).unwrap();
+
+            let dropin_dir = tmpdir.path().join("config.d");
+            fs::create_dir(&dropin_dir).unwrap();
+
+            let dropin_path = dropin_dir.join("10-base");
+            create_file(&dropin_path, dropin_data.as_bytes()).unwrap();
+
+            let dropin_override_path = dropin_dir.join("20-override");
+            create_file(&dropin_override_path, dropin_override_data.as_bytes()).unwrap();
+
+            let config = load(&config_path).unwrap();
+            assert_eq!(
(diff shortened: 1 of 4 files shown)


Tim-0731-Hzt:
> a public function, however it doesn't seem to be called from anywhere currently) and `load()` (public as well but only called internally

Yes, I think so. Both `load_raw_from_file()` and `load()` could be made privately.

> This PR adds handling of drop-in config file fragments to runtime-rs to match the same feature already implemented in golang runtime. Behaviour of this Rust port should be the same as the Golang version so no additional documentation should be necessary.
> 
> The change is fairly non-intrusive, mostly adding new code, but the final commit does involve some refactoring of the existing code. Admittedly there were multiple ways to achieve the goal of creating a single point that drop-in handling needs to be added to. Being not entirely clear about the intended usage of `load_raw_from_file()` (a public function, however it doesn't seem to be called from anywhere currently) and `load()` (public as well but only called internally AFAICT) I just went for what seemed to be the simplest and least intrusive option.

Yes, I think so. Both `load_raw_from_file()` and `load` could be made privately.

src/libs/kata-types/src/config/mod.rs
```diff
@@ -94,21 +94,15 @@ impl TomlConfig {
...
-        info!(
-            sl!(),
-            "load configuration from: {}",
-            file_path.to_string_lossy()
-        );
-        let content = fs::read_to_string(&file_path)?;
-        let config = Self::load(&content)?;
+        let mut result = Self::load_raw_from_file(config_file);
+        if let Ok((ref mut config, _)) = result {
+            Hypervisor::adjust_config(config)?;
```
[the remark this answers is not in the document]
pmores: I believe it should be, fundamentally.  However, `load()` is used to initialise `KATA_DEFAULT_CONFIG` and `KATA_ACTIVE_CONFIG` and I wasn't sure what the ramifications would be if I messed with `load()`'s implementation.  

Also, if `load()` were to handle drop-ins as well its interface would need to change - it would have to take a _path_ to the base config file, not just its _contents_ as it's impossible to figure out where to look for drop-ins given just the base config file contents.

That said I'm very open to suggestions how best to integrate drop-in handling into existing config loading code.  I'm not at all sure the path I've taken is the best one.

src/libs/kata-types/src/config/mod.rs
```diff
@@ -164,6 +160,267 @@ impl TomlConfig {
     }
 }
 
+mod drop_in {
```
Tim-0731-Hzt, line 163: I think it's better to put the `drop_in` mod and the unit test into a new file `drop_in.rs` under the `config` folder, and declare it  in this file.

src/libs/kata-types/src/config/mod.rs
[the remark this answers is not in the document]
Tim-0731-Hzt: The implementation looks good so far

src/libs/kata-types/src/config/mod.rs
pmores: I agree and will do!

pmores:
@Tim-0731-Hzt I pushed a new version that moves the drop-in handling implementation out to `drop_in.rs` as suggested.  This is just a pure code reorganisation, none of the actual algorithms are changed.

I also took this opportunity to split the private implementation into two private submodules to make it explicit that it has two distinct layers.  The lower one deals exclusively with TOML parse trees merging and is used by the higher one that handles the filesystem-related aspect of drop-in cfg file fragments.  The public reexport of `load()` is the only public interface of the `drop_in` module.

pmores:
@quanweiZhou I'm looking into the dragonball test failure.

pmores:
@quanweiZhou Wow, there was an actual omission in `config.d` handling - good catch, CI, and shame on me! :-) I fixed that and added a couple of related tests.

liubin (approved):

liubin:
Since `runtime-rs` has been merged, please change your target branch to `main`.

pmores:
@liubin Thanks for reminding me, done!

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -0,0 +1,215 @@
...
+            merge(&mut base, dropin);
+
+            assert_eq!(
+                base.try_into(),
+                Ok(Config {
+                    number: 42,
+                    string: "foo".into(),
+                    sub: SubConfig {
+                        ..Default::default()
+                    },
```
dgibson, line 98: Can just use `sub: Default::default()`

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+mod drop_in_directory_handling {
+    use crate::config::TomlConfig;
+    use std::fs;
+    use std::io::{self, Result};
+    use std::path::{Path, PathBuf};
+
+    fn get_dropin_dir_path(base_cfg_file_path: &Path) -> Result<PathBuf> {
+        let mut dropin_dir = base_cfg_file_path.to_path_buf();
+        if !dropin_dir.pop() {
+            return Err(io::Error::new(
```
dgibson, line 228: Using `io::Error` here is maybe a bit odd, since the error isn't really related to IO.  Consider using `anyhow::Error` instead?

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+                io::ErrorKind::InvalidInput,
+                "base cfg file path too short",
+            ));
+        }
+        dropin_dir.push("config.d");
+        Ok(dropin_dir)
+    }
+
+    fn update_from_dropin(base_config: &mut toml::Value, dropin_file: &fs::DirEntry) -> Result<()> {
+        if !dropin_file.file_type()?.is_file() && !dropin_file.file_type()?.is_symlink() {
```
dgibson, line 238: Double check if you need `is_symlink`; in many languages a symlink to a regular file will pass both "is file" and "is symlink" checks.

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+                    return Ok(());
+                } else {
+                    return Err(err);
+                }
+            }
+        };
+
+        let mut dropin_files = dropin_files_iter.collect::<Result<Vec<_>>>()?;
+        dropin_files.sort_by_key(|direntry| direntry.file_name());
+        for dropin_file in dropin_files.iter() {
```
dgibson, line 264: Can be just `for dropin_file in &dropin_files`

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+                [hypervisor.qemu]
+                path = "/usr/bin/qemu-kvm"
+                default_bridges = 3
+                [runtime]
+                enable_debug = true
+                internetworking_model="tcfilter"
+            "#;
+
+            let mut config_path = PathBuf::from(tmpdir.path());
+            config_path.push("runtime.toml");
```
dgibson, line 304: Slightly clearer to use `tmpdir.path().join("runtime.toml")`

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+            let result = create_file(&config_path, runtime_config_file_data.as_bytes());
+            assert!(result.is_ok());
```
dgibson, line 306: Could just `unwrap()` in the previous line - unwrap is a special form of assertion, so is reasonable in tests.

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+
+            let config = load(&config_path).unwrap();
+            assert_eq!(
+                config.hypervisor["qemu"].path,
+                "/usr/bin/qemu-kvm".to_string()
+            );
+            assert_eq!(config.hypervisor["qemu"].device_info.default_bridges, 3);
+            assert!(config.runtime.debug);
+            assert_eq!(config.runtime.internetworking_model, "tcfilter".to_string());
```
dgibson, line 348: Could use a common helper function for the last piece of the test in both `test_no_dropins_dir` and `test_no_dropins`.

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+            let tmpdir = tempfile::tempdir().unwrap();
+
+            let runtime_config_file_data = r#"
+                [hypervisor.qemu]
+                path = "/usr/bin/qemu-kvm"
+                default_bridges = 3
+                [runtime]
+                enable_debug = true
+                internetworking_model="tcfilter"
+            "#;
```
dgibson, line 329: Could use a common const definition for this in the test module.

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -213,3 +213,205 @@ mod toml_tree_ops {
...
+            assert_eq!(
+                config.hypervisor["qemu"].path,
+                "/usr/bin/qemu-kvm".to_string()
+            );
+            assert_eq!(config.hypervisor["qemu"].cpu_info.default_vcpus, 2);
+            assert_eq!(config.hypervisor["qemu"].device_info.default_bridges, 4);
+            assert_eq!(
+                config.hypervisor["qemu"].shared_fs.shared_fs,
+                Some("virtio-9p".to_string())
+            );
```
dgibson, line 412: This does an allocation of a new copy of the "virtio-9p" string, which is a little bit ugly, though it doesn't really matter for a test case.  If you wanted to, the way to do this comparison without the allocation is:
```
assert_eq!(config.hypervisor["qemu"].shared_fs.shared_fs.as_ref(), Some("virtio-9p"));
```

dgibson (approved):
I have commented with a number of minor nits that could be polished, but overall this looks good to me.

pmores:
Pushed a new version to integrate @dgibson's proposals.  Thank you very much for taking the time to do such a detailed review.

src/libs/kata-types/src/config/drop_in.rs
pmores: Yes, you're right and this is an actual bug - good catch, thanks for that! A symlink to a directory would be handled inappropriately by this code.

src/libs/kata-types/src/config/drop_in.rs
dgibson: Ok, fair enough.

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -212,3 +212,171 @@ mod toml_tree_ops {
...
+                vfio_mode="vfio"
+            "#;
+
+            let config_path = tmpdir.path().join("runtime.toml");
+            create_file(&config_path, BASE_CONFIG_DATA.as_bytes()).unwrap();
+
+            let dropin_dir = tmpdir.path().join("config.d");
+            fs::create_dir(&dropin_dir).unwrap();
+
+            let dropin_path = dropin_dir.as_path().join("10-base");
```
dgibson, line 361: I don't believe you need the `.as_path()` calls - Deref coercion should mean you can can call `join` directly on the `PathBuf`.

src/libs/kata-types/src/config/drop_in.rs
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) 2022 Red Hat
```
dgibson, line 1: Not really an upstream consideration, but fwiw, I believe Red Hat Legal recommends using simply `Copyright Red Hat` rather than putting years or `(c)` in such notices (I was surprised too).

dgibson (approved):
Thanks for the polish changes.  I noted a couple of remaining trivial nits, but LGTM.

jodh-intel (approved):
Thanks @pmores.

lgtm, although we will need some doc to support this change. Could you tal at that in a follow-up PR?

pmores:
Pushed the remaining fixes as proposed by @dgibson.

@jodh-intel Thanks! Obviously I'm happy to adjust docs but I'm afraid I'll need a hint as to what's necessary. I thought originally that existing documentation would do since this change is meant to work precisely the same as the golang version but I'm probably overlooking something.

Issue #4779, opened 2022-08-01
CC | Build TD-shim to be used with Cloud Hypervisor


fidencio:
TD-shim is the firmware used to boot up a TDVM with Cloud Hypervisor, therefore we should track and build it.

PR #4783, opened 2022-08-01
versions: Track and add support for building TD-shim

fidencio:
TD-shim is a simplified TDX virtual firmware, used by Cloud Hypervisor,
in order to create a TDX capable VM.

TD-shim is heavily under development, and is hosted as part of the
Confidential Containers project:
https://github.com/confidential-containers/td-shim

The version chosen for this commit, is a version that's being tested
inside Intel, but we, most likely, will need to change it before we have
it officially packaged as part of an official release.

Fixes: #4779

### DIFF
--- a/tools/packaging/static-build/td-shim/Dockerfile
+++ b/tools/packaging/static-build/td-shim/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright (c) 2022 Intel
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ubuntu:20.04
+ENV DEBIAN_FRONTEND=noninteractive
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+ARG RUST_TOOLCHAIN
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        clang \
+        curl \
+        gcc \
+        git \
+        llvm \
+        nasm && \
+    apt-get clean && rm -rf /var/lib/lists/ && \
+    curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain ${RUST_TOOLCHAIN} && \
+    source "$HOME/.cargo/env" && \
+    rustup component add rust-src && \
+    cargo install cargo-xbuild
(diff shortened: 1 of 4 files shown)


tools/packaging/static-build/td-shim/build-td-shim.sh
```diff
@@ -0,0 +1,41 @@
...
+[ -n "${tdshim_repo}" ] || die "Failed to get TD-shim repo"
+[ -n "${tdshim_version}" ] || die "Failed to get TD-shim version or commit"
+
+info "Build ${tdshim_repo} version: ${tdshim_version}"
+
+source ${HOME}/.cargo/env
+
+build_root=$(mktemp -d)
+pushd ${build_root}
+git clone --single-branch "${tdshim_repo}"
```
fidencio, line 27: @wainersm, the reason I'm not using `--depth 1` here is because if I do, I can't simply checkout to the tdshim_version a few lines below.

wainersm: ok @fidencio . thanks for the heads up.

tools/packaging/static-build/td-shim/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        clang \
+        curl \
+        gcc \
+        git \
+        llvm \
+        nasm && \
+    apt-get clean && rm -rf /var/lib/lists/ && \
+    curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2022-04-07 && \
```
wainersm, line 19: I read on td-shim's readme why they are using the nightly build of rust. That's ok. However I noticed that since we are pinning on commit "5f62a0e367b1845a54e534d103ed4a697a599ac3", wouldn't be possible to pin the rust version too?

My worry is about the build getting broke on our face from time to time. Apart from that, I suggest to put that toolchain version on `versions.yml` as well, under the td-shim section.

fidencio: Done!

wainersm:
Tested on my local Fedora 34 x86_64 machine.

wainersm (approved):

amshinde (approved):

Issue #4808, opened 2022-08-03
OVMF: Adjust final tarball location


fidencio:
Let's create the OVMF tarball in the directory where the script was
called from, instead of doing it in the $DESTDIR.

This aligns with the logic being used for creating / extracting the
tarball content, which is already in use by the kata-deploy local build
scripts.

PR #4810, opened 2022-08-03
OVMF / td-shim: Adjust final tarball location

fidencio:
Let's create the OVMF / td-shim tarball in the directory where the script was
called from, instead of doing it in the $DESTDIR.

This aligns with the logic being used for creating / extracting the
tarball content, which is already in use by the kata-deploy local build
scripts.

Fixes: #4808, #4809


wainersm (approved):

stevenhorsman (approved):
LGTM. Thanks!

fidencio:
Power CI seems to be failing in a bunch of different PRs.  As this one doesn't touch any part that's specific to it, I'm merging the PR.

Issue #1378, opened 2021-02-09
agent should use OsString or std::fs::Path in many places it uses String


dgibson:
# Description of problem

Rust `String` and `str` types represent valid UTF-8 strings.  OS paths aren't necessarily the same thing: on Linux they're a string of non-zero bytes, which need not be valid UTF-8, on Windows they're UCS-2 or UTF-16 and depending on version may not even be properly validated UTF-16.  Rust has specific types for representing these, `OsString`, `OsStr` and `std::path::Path`.

We should use those types wherever possible if we're manipulating OS relevant string, but in most cases we use `str` and `String` instead.  That both results in a number of unnecessary conversions, and may mean we get hard to debug failures if we encounter an OS filename which is *not* UTF-8, rare, but possible.

fidencio:
@dgibson, thanks for the bug report.  I've removed the "needs-review" label and added the "area/agent" one instead.
Would you be interested to work on this one?

dgibson:
@fidencio I hope to work on this at some point, but it will be fitted around other things I'm doing, so no promises of any quick results.

bookinabox:
I will look at this, but I expect to be slow because I'm not too familiar with Rust.

PR #4815, opened 2022-08-04
logging: Replace nix::Error::EINVAL with more descriptive msgs

### LINKED ISSUE #954: Return meaningful error message for Errs
In agent there are many `Error`s returned by:

```rust
return Err(anyhow!(nix::Error::from_errno(Errno::EINVAL)));
```

With this we might only get an error message of "Invalid argument".

These errors are lack of detail information about the error, hard to understand, and not user friendly.

`Error`s should contains more meaningful information.

bookinabox:
Replaces instances of anyhow!(nix::Error::EINVAL) with other messages to
make it easier to debug.

Fixes #954

### DIFF
--- a/src/agent/rustjail/src/cgroups/fs/mod.rs
+++ b/src/agent/rustjail/src/cgroups/fs/mod.rs
@@ -174,7 +174,7 @@ impl CgroupManager for Manager {
                 freezer_controller.freeze()?;
             }
             _ => {
-                return Err(anyhow!(nix::Error::EINVAL));
+                return Err(anyhow!("Invalid FreezerState"));
             }
         }
 
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -106,6 +106,11 @@ impl Default for ContainerStatus {
     }
 }
 
+// We might want to change this to thiserror in the future
+const MissingCGroupManager: &str = "failed to get container's cgroup Manager";
+const MissingLinux: &str = "no linux config";
+const InvalidNamespace: &str = "invalid namespace type";
+
 pub type Config = CreateOpts;
 type NamespaceType = String;
 
@@ -292,7 +297,7 @@ impl Container for LinuxContainer {
             self.status.transition(ContainerState::Paused);
             return Ok(());
         }
-        Err(anyhow!("failed to get container's cgroup manager"))
+        Err(anyhow!(MissingCGroupManager))
     }
 
     fn resume(&mut self) -> Result<()> {
@@ -310,7 +315,7 @@ impl Container for LinuxContainer {
             self.status.transition(ContainerState::Running);
             return Ok(());
         }
-        Err(anyhow!("failed to get container's cgroup manager"))
+        Err(anyhow!(MissingCGroupManager))
     }
 }
 
@@ -397,7 +402,7 @@ fn do_init_child(cwfd: RawFd) -> Result<()> {
     };
 
     if spec.linux.is_none() {
-        return Err(anyhow!("no linux config"));
+        return Err(anyhow!(MissingLinux));
     }
     let linux = spec.linux.as_ref().unwrap();
 
@@ -411,7 +416,7 @@ fn do_init_child(cwfd: RawFd) -> Result<()> {
     for ns in &nses {
         let s = NAMESPACES.get(&ns.r#type.as_str());
         if s.is_none() {
-            return Err(anyhow!("invalid ns type"));
+            return Err(anyhow!(InvalidNamespace));
         }
         let s = s.unwrap();
 
@@ -1437,17 +1442,9 @@ impl LinuxContainer {
             Some(unistd::getuid()),
             Some(unistd::getgid()),
         )
-        .context(format!("cannot change onwer of container {} root", id))?;
-
-        if config.spec.is_none() {
-            return Err(anyhow!(nix::Error::EINVAL));
-        }
+        .context(format!("Cannot change owner of container {} root", id))?;
 
         let spec = config.spec.as_ref().unwrap();
-
-        if spec.linux.is_none() {
-            return Err(anyhow!(nix::Error::EINVAL));
-        }
 
         let linux = spec.linux.as_ref().unwrap();
 
@@ -1525,7 +1522,7 @@ pub async fn execute_hook(logger: &Logger, h: &Hook, st: &OCIState) -> Result<()
     let binary = PathBuf::from(h.path.as_str());
     let path = binary.canonicalize()?;
     if !path.exists() {
-        return Err(anyhow!(nix::Error::EINVAL));
+        return Err(anyhow!("Path {:?} does not exist", path));
     }
 
     let mut args = h.args.clone();
--- a/src/agent/rustjail/src/mount.rs
+++ b/src/agent/rustjail/src/mount.rs
@@ -1020,9 +1020,7 @@ pub fn finish_rootfs(cfd_log: RawFd, spec: &Spec, process: &Process) -> Result<(
 }
 
 fn mask_path(path: &str) -> Result<()> {
-    if !path.starts_with('/') || path.contains("..") {
-        return Err(anyhow!(nix::Error::EINVAL));
-    }
+    check_paths(path)?;
 
     match mount(
         Some("/dev/null"),
@@ -1040,9 +1038,7 @@ fn mask_path(path: &str) -> Result<()> {
 }
 
 fn readonly_path(path: &str) -> Result<()> {
-    if !path.starts_with('/') || path.contains("..") {
-        return Err(anyhow!(nix::Error::EINVAL));
-    }
+    check_paths(path)?;
 
     if let Err(e) = mount(
         Some(&path[1..]),
@@ -1065,6 +1061,16 @@ fn readonly_path(path: &str) -> Result<()> {
         None::<&str>,
     )?;
 
+    Ok(())
+}
+
+fn check_paths(path: &str) -> Result<()> {
+    if !path.starts_with('/') || path.contains("..") {
+        return Err(anyhow!(
+            "Cannot mount {} (path does not start with '/' or contains '..').",
+            path
+        ));
+    }
     Ok(())
 }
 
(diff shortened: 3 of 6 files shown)


bookinabox:
I'm not too familiar with the internals of the agent or rust, so please let me know if there are any inaccuracies or bad practices.

src/agent/rustjail/src/mount.rs
```diff
@@ -1041,7 +1044,10 @@ fn mask_path(path: &str) -> Result<()> {
 
 fn readonly_path(path: &str) -> Result<()> {
     if !path.starts_with('/') || path.contains("..") {
-        return Err(anyhow!(nix::Error::EINVAL));
+        return Err(anyhow!(
+            "Cannot mount {} (path does not start with '/' or contains ...",
```
ManaSugi, line 1048: It might be a good:

```
"Cannot mount {} (path does not start with '/' or contains '..')."
```

src/agent/rustjail/src/mount.rs
```diff
@@ -1021,7 +1021,10 @@ pub fn finish_rootfs(cfd_log: RawFd, spec: &Spec, process: &Process) -> Result<(
 
 fn mask_path(path: &str) -> Result<()> {
     if !path.starts_with('/') || path.contains("..") {
-        return Err(anyhow!(nix::Error::EINVAL));
+        return Err(anyhow!(
+            "Cannot mount {} (path does not start with '/' or contains ...)",
```
ManaSugi, line 1025: It might be a good:

```
"Cannot mount {} (path does not start with '/' or contains '..')."
```

ManaSugi (commented):
Thank you for improving the error messages! I have a few comments.

bookinabox:
Thanks for the feedback!

ManaSugi:
> @bookinabox Please check the format by running `cargo fmt -- --check`. https://github.com/kata-containers/kata-containers/runs/7667490131?check_suite_focus=true#step:12:81

@bookinabox  Could you check it as well?

bookinabox:
> > @bookinabox Please check the format by running `cargo fmt -- --check`. https://github.com/kata-containers/kata-containers/runs/7667490131?check_suite_focus=true#step:12:81
> 
> @bookinabox Could you check it as well?
Hmm, I did run it. Let me double check all my versions are the same and the specific file its failing at. Sorry about that.

ManaSugi:
@bookinabox Thank you for the update

> well, I changed my versions to the ones on the runner. Not sure if that's what fixed it, but fmt did change stuff now.

JFYI: Kata maintains the version information in the [versions.yaml](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml) and the current Rust version Kata uses is bellow.
https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L306

bookinabox:
> @bookinabox Thank you for the update
> 
> > well, I changed my versions to the ones on the runner. Not sure if that's what fixed it, but fmt did change stuff now.
> 
> JFYI: Kata maintains the version information in the [versions.yaml](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml) and the current Rust version Kata uses is bellow. https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L306

Thanks!

ManaSugi:
@bookinabox if the PR are ready, please remove the draft status.

ManaSugi (approved):

fidencio (approved):
lgtm, thanks @bookinabox!

src/agent/rustjail/src/mount.rs
```diff
@@ -1041,7 +1044,10 @@ fn mask_path(path: &str) -> Result<()> {
 
 fn readonly_path(path: &str) -> Result<()> {
     if !path.starts_with('/') || path.contains("..") {
-        return Err(anyhow!(nix::Error::EINVAL));
+        return Err(anyhow!(
```
jodh-intel, line 1047: It might be worth creating a common check function for this and `mask_path()` to avoid the duplicated code so both functions could start with something like:

```rust
check_path(path)?;

// ...
```

src/agent/rustjail/src/validator.rs
```diff
@@ -4,17 +4,15 @@
...
 
-fn einval() -> Error {
-    anyhow!(nix::Error::EINVAL)
-}
-
 fn get_linux(oci: &Spec) -> Result<&Linux> {
-    oci.linux.as_ref().ok_or_else(einval)
+    oci.linux
+        .as_ref()
+        .ok_or_else(|| anyhow!("Unable to get Linux from Spec"))
```
jodh-intel, line 15: Nit: This makes sense to developers but might be clearer for end users as something like:

```suggestion
        .ok_or_else(|| anyhow!("Unable to get Linux section from Spec"))
```

But, either way, this message is inconsistent with the (redundant) same check on the Linux spec in `src/agent/rustjail/src/container.rs`.

src/agent/rustjail/src/container.rs
```diff
@@ -1437,16 +1437,18 @@ impl LinuxContainer {
             Some(unistd::getuid()),
             Some(unistd::getgid()),
         )
-        .context(format!("cannot change onwer of container {} root", id))?;
+        .context(format!("Cannot change owner of container {} root", id))?;
 
         if config.spec.is_none() {
-            return Err(anyhow!(nix::Error::EINVAL));
+            return Err(anyhow!("Cannot create the container: the spec is none"));
```
jodh-intel, line 1443: - Nit: You could make this shorter and not lose detail:

  ```suggestion
            return Err(anyhow!("Cannot create the container: no spec"));
  ```
- For strings with no format specifiers like this, I'd be tempted to define them as top-level `const` values. That also means the unit tests can reference those variables to assert the expected message when they provoke errors.

src/agent/rustjail/src/cgroups/fs/mod.rs
```diff
@@ -174,7 +174,9 @@ impl CgroupManager for Manager {
                 freezer_controller.freeze()?;
             }
             _ => {
-                return Err(anyhow!(nix::Error::EINVAL));
+                return Err(anyhow!(
+                    "Invalid FreezerState. Must be FreezerState::Frozen or FreezerState::Thawed"
```
jodh-intel, line 178: I'd consider inverting this as although the message is superficially helpful, it doesn't result in the actual state being logged, so doesn't help developers. I'm being arguably overly harsh/pedantic here as there is only one other state, but what if there are >1 other states in future?

I'd just log something like:

```suggestion
                    "Invalid FreezerState: {:?}", state
```

Of course, that won't currently work as FreezerState` doesn't support `Display`/`Debug`, but I think it should. If you're keen, you might want to raise a PR on https://github.com/kata-containers/cgroups-rs to add atleast `#[derive(Debug)]` to the `enum`.

src/agent/rustjail/src/container.rs
```diff
@@ -1437,16 +1437,18 @@ impl LinuxContainer {
...
         }
 
         let spec = config.spec.as_ref().unwrap();
 
         if spec.linux.is_none() {
-            return Err(anyhow!(nix::Error::EINVAL));
+            return Err(anyhow!(
+                "Cannot create the container: the linux spec is none"
```
jodh-intel, line 1450: Interestingly, both this check and the one above are technically redundant because the first thing the function does is call `validate()` which already performs these checks.

src/agent/rustjail/src/validator.rs
```diff
@@ -31,7 +29,10 @@ fn rootfs(root: &str) -> Result<()> {
     let path = PathBuf::from(root);
     // not absolute path or not exists
     if !path.exists() || !path.is_absolute() {
-        return Err(einval());
+        return Err(anyhow!(
+            "Path from {} does not exist or is not absolute",
```
jodh-intel, line 33: I'd use `{:?}` for all paths to ensure they are quoted (so you can see if they are blank).

src/agent/rustjail/src/validator.rs
```diff
@@ -58,10 +59,10 @@ fn rootfs(root: &str) -> Result<()> {
         cleaned.push(e);
     }
 
-    let canon = path.canonicalize().context("canonicalize")?;
+    let canon = path.canonicalize().context("failed to canonicalize path")?;
     if cleaned != canon {
         // There is symbolic in path
-        return Err(einval());
+        return Err(anyhow!("Cleaned and canonicalized path do not match. There may be illegal symbols in the path name"));
```
jodh-intel, line 65: - The invalid path isn't being logged in the message so this is reliant on the caller logging it. Could you check/adjust as necessary?
- This canonicalisation check is another one that could be added to a new `check_path()` function possibly.

jodh-intel (commented):
Thanks @bookinabox - a few comments...

src/agent/rustjail/src/mount.rs
jodh-intel: ... if you did do this (possibly on anther PR), you could consider writing a set of table-based tests:

- https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md#rust-table-driven-tests

There are lots of examples of these in the repo - search for `TestData` to find a bunch.

ManaSugi:
@bookinabox 
Could you also replace remaining `EINVAL` with detailed messages?

```
src/rpc.rs
180:                return Err(anyhow!(nix::Error::EINVAL));
250:            return Err(anyhow!(nix::Error::EINVAL));
1930:        return Err(anyhow!(nix::Error::EINVAL));

src/netlink.rs
615:        let v = u8::from_str_radix(split.next().ok_or_else(|| anyhow!(nix::Error::EINVAL))?, 16)?;
```

src/agent/rustjail/src/container.rs
bookinabox: If we want to take it a step further I could try to convert the errors to use thiserror instead of anyhow, though that might be a little overkill. I do see thiserror being used a couple other places.

src/agent/rustjail/src/validator.rs
bookinabox: I was working on a different [issue](https://github.com/kata-containers/kata-containers/issues/1378) converting String to OsString/Paths. While I haven't look at this too closely, it looks like this is making sure that paths are valid Strings. Perhaps we might not need this check in the future if we are able to properly replace String instances?

bookinabox:
^ didn't run clippy, so I had an error not caught by fmt

bookinabox:
Should have learned my lesson. It should pass now. I remembered to run test, clippy, and fmt.

Issue #4816, opened 2022-08-04
runtime-rs: build failed when the arch is s390x


Tim-0731-Hzt:
`kata-containers-2.0-ubuntu-s390x-unit` will compile the `runtime-rs`. However, `s390x` platform is not supported currently for `runtime-rs.`

studychao:
PS: rust runtime could work with s390x but our built-in VMM could not work on s390x platform.

After we implement the support for QEMU, we could have support for s390x through QEMU.

PR #4817, opened 2022-08-04
runtime-rs:skip the build process when the arch is s390x

Tim-0731-Hzt:
Depends-on: github.com/kata-containers/tests#4986. To avoid returning an error when running the ci, we just skip the build
process if the arch is s390x

Fixes: #4816

### DIFF
--- a/src/runtime-rs/Makefile
+++ b/src/runtime-rs/Makefile
@@ -21,14 +21,19 @@ ARCH_DIR = arch
 ARCH_FILE_SUFFIX = -options.mk
 ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
 
-
 ifeq ($(ARCH), s390x)
-##TARGET default: build code
 default:
 	@echo "s390x not support currently"
 	exit 0
+test:
+	@echo "s390x not support currently"
+	exit 0
 else
+##TARGET default: build code
 default: runtime show-header
+#TARGET test: run cargo tests
+test:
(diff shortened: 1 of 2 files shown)


src/runtime-rs/Makefile
```diff
@@ -21,8 +21,18 @@ ARCH_DIR = arch
...
+default:
+	@echo "s390x not support currently"
+	exit 0
+else
+default: runtime show-header
+endif
+
 ifeq (,$(realpath $(ARCH_FILE)))
-    $(error "ERROR: invalid architecture: '$(ARCH)'")
+	$(error "ERROR: invalid architecture: '$(ARCH)'")
```
liubin, line 35: Please keep the original indent chars as it was (tabs or spaces).

Tim-0731-Hzt: fixed

studychao (approved):

src/runtime-rs/Makefile
```diff
@@ -21,6 +21,21 @@ ARCH_DIR = arch
...
 
+ifeq ($(ARCH), s390x)
+default:
+	@echo "s390x not support currently"
+	exit 0
+test:
+	@echo "s390x not support currently"
+	exit 0
+else
+##TARGET default: build code
```
studychao, line 32: Why is there two `#` here? Could you keep the comment in the same style?

liubin (approved):

amshinde (approved):

PR #4839, opened 2022-08-08
CC | Build and ship QEMU TDX (and all the needed artefacts)

fidencio:
Let's add a new configuration for using a TDX capable QEMU with Kata Containers, including all the artefacts needed, and the changes that are needed as part of the Makefile and deployment scripts.


stevenhorsman (approved):
LGTM - just a minor point/question.

src/runtime/Makefile
```diff
@@ -254,12 +261,28 @@ ifneq (,$(QEMUCMD))
...
     # qemu-specific options (all should be suffixed by "_QEMU")
     DEFBLOCKSTORAGEDRIVER_QEMU := virtio-scsi
     DEFNETWORKMODEL_QEMU := tcfilter
     KERNELTYPE = uncompressed
     KERNELNAME = $(call MAKE_KERNEL_NAME,$(KERNELTYPE))
     KERNELPATH = $(KERNELDIR)/$(KERNELNAME)
+
+    KERNELTDXTYPE = compressed
+    KERNELTDXNAME = $(call MAKE_KERNEL_TDX_NAME,$(KERNELTDXTYPE))
+    KERNELTDXPATH = $(KERNELDIR)/$(KERNELTDXNAME)
```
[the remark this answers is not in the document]
fidencio: > minor nit, but the other 'non-standard' hypervisor type variables seem to have the format:
KERNELTYPE_<hypervisor>, KERNEL_NAME_<hypervisor>, KERNELPATH_<hypervisor>, so I wonder if KERNELTYPE_TDX, KERNEL_NAME_TDX, KERNELPATH_TDX would be better for consistency?

I thought about that, @stevenhorsman, and I noticed that KERNELTYPE_<hypervisor> is not followed with QEMU, as QEMU is the default hypervisor.  Thus I've decided to not add it for TDX as it'd be used with QEMU (at least for now).

I can change it, no problem.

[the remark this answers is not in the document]
fidencio: > Also just to check if there is a valid scenario that the user might want to override something like KERNELTYPE_TDX/KERNELTDXTYPE to make the vmlinuz version for TDX and therefore require it in the USER_VARS list?

Hmm.  I wne tback and forth on that as well, but as we're providing the needed kernel, and it's very niche at the moment, I decided to only add it if someone requests for it.

Makes sense?  I can add it to the user vars if you prefer.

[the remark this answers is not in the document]
stevenhorsman: I'm not unhappy with the current implementation and the fact that you've thought about it and decided on a better approach is fine, so no need to change it :)

fidencio:
This `make kata-tarball` test constantly failing in different repos is not good. :-/

src/runtime/Makefile
```diff
@@ -112,6 +113,9 @@ KERNELDIR := $(PKGDATADIR)
 IMAGEPATH := $(PKGDATADIR)/$(IMAGENAME)
 FIRMWAREPATH :=
 FIRMWAREVOLUMEPATH :=
+TDVFFIRMWAREPATH := $(PREFIXDEPS)/share/tdvf/OVMF.fd
+TDVFFIRMWAREVOLUMEPATH := $(PREFIXDEPS)/share/tdvf/OVMF_VARS.fd
+TDXKERNELPARAMS := force_tdx_guest tdx_disable_filter
```
wainersm, line 118: Shouldn't the kata machinery append the kernel parameters "auto-magically" in case of  ` confidential_guest = true`?

fidencio: Not really, @wainersm.  For now these parameters are needed, but we're not sure if that will be the case with newer versions of the kernel, neither what's the situation with AMD SEV.

fidencio: There's also another thing, @wainersm, we're currently hardcoding the `confidential_guest` value depending on the configuration file used, so we can generate everything as part of the very same build.  Adding logic to do "smarter" things at this point of the project could actually lead us to more complexity without much gain.

We should, though,revisit this discussion when merging things into the `main` branch.

wainersm: got it @fidencio and I agree.

wainersm (approved):
@fidencio sorry for taking too long to review it.

Issue #4835, opened 2022-08-05
Skip CI for document PRs


cmaf:
We currently are required to run the CI for documents. We could skip to save resources. Also consider for test repo.

cc @amshinde

bookinabox:
You already commented on it, but if anyone else is looking at this #4195 partially implemented this for #3963, but has not been work on in a while.

PR #4845, opened 2022-08-08
[DEBUG]workflows: skip static checks for docs only PRs

liubin:
If a PR only has docs/ content changed, then skip
other CI checks.

Fixes: #4835


liubin:
replaced by #4946

Issue #4852, opened 2022-08-09
Use systemd cgroups when sandbox_cgroup_only = false


yaoyinnan:
When using systemd to manage cgroups, the scope unit is automatically unmounted when all processes are removed from the cgroup. When it is used again, the scope unit of the original cgroup path cannot be found, and an error is thrown.

Originally, due to this issue, when sandbox_cgroup_only = false, whether the systemd cgroup path was correct or not, it was managed as a cgroupfs cgroup path.

https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/resourcecontrol/cgroups.go#L172
```
// Currently we know to handle systemd cgroup path only when it's the only cgroup (no overhead group), hence,
// if sandboxCgroupOnly is not true we treat it as cgroupfs path as it used to be, although it may be incorrect.
if !IsSystemdCgroup(path) || !sandboxCgroupOnly {
return NewResourceController(path, &sandboxResources)
}
```

Therefore, the fix plan is to add a temporary process to the scope unit to prevent the scope unit from being automatically destroyed when all processes are removed.

PR #4853, opened 2022-08-09
fix: Use systemd cgroups when sandbox_cgroup_only = false

yaoyinnan:
When using systemd to manage cgroups, the scope unit is automatically unmounted when all processes are removed from the cgroup. When it is used again, the scope unit of the original cgroup path cannot be found, and an error is thrown.

Originally, due to this issue, when sandbox_cgroup_only = false, whether the systemd cgroup path was correct or not, it was managed as a cgroupfs cgroup path.

https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/resourcecontrol/cgroups.go#L172

```
// Currently we know to handle systemd cgroup path only when it's the only cgroup (no overhead group), hence,
// if sandboxCgroupOnly is not true we treat it as cgroupfs path as it used to be, although it may be incorrect.
if !IsSystemdCgroup(path) || !sandboxCgroupOnly {
return NewResourceController(path, &sandboxResources)
}
```

Therefore, the fix plan is to add a temporary process to the scope unit to prevent the scope unit from being automatically destroyed when all processes are removed.

Fixes: #4852

### DIFF
--- a/src/runtime/pkg/resourcecontrol/cgroups.go
+++ b/src/runtime/pkg/resourcecontrol/cgroups.go
@@ -10,7 +10,6 @@ package resourcecontrol
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


ManaSugi:
@yaoyinnan Thank you.
I think it would be better to control this using  `RemainAfterExit=yes|no` that is a service unit setting.

> RemainAfterExit=
   Takes a boolean value that specifies whether the service shall be considered active even when all its processes exited. Defaults to no.

In addition, If possible, it would be nice if users could change systemd settings via the runtime configuration.
Also, Is it correct behavior that the scope unit is always not removed even if the all processes terminate?

yaoyinnan:
> RemainAfterExit

Thanks a lot for your advice and help.

I found in my tests that RemainAfterExit seems to only work on the service unit, not the scope unit.

And, during the testing process, I also found other issues, which I sorted out in this https://github.com/kata-containers/kata-containers/issues/4886.

PR #4849, opened 2022-08-09
kernel: remove depricated CONFIG_ND_BLK

Alex-Carter01:
Fragment included config depricated as of this commit:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=f8669f1d6a86a6b17104ceca9340ded280307ac1

Fixes: kata-containers#4848
Signed-off-by: Alex Carter <alex.carter@ibm.com>

Issue #4787, opened 2022-08-01
CI kata-deploy failing to make kata tarball due to shared agent directory conflicts


ryansavino:
# Description of problem

CI kata-deploy test for 'make-kata-tarball' fails for branch `main`.

Testing command used to reproduce (directory: kata-containers):
```
make kata-tarball
```

# Expected result

kata tarball should be created.

# Actual result

The kata-agent fails to build during the tarball creation.
tools/packaging/kata-deploy/local-build/Makefile
Target: rootfs-initrd-tarball-build

```
   Compiling slog v2.7.0
   Compiling hashbrown v0.12.1
error: could not write output to /kata-containers/src/agent/target/release/build/slog-e9179efe9a84fbd0/build_script_build-e9179efe9a84fbd0.build_script_build.ca7294ca-cgu.13.rcgu.o: No such file or directory

error: error writing dependencies to `/kata-containers/src/agent/target/release/deps/hashbrown-d69ab10a53de4603.d`: No such file or directory (os error 2)

error: could not compile `slog` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `hashbrown` due to previous error
make: *** [Makefile:116: target/x86_64-unknown-linux-musl/release/kata-agent] Error 101
Failed at 585: make LIBC=${LIBC} INIT=${AGENT_INIT} SECCOMP=${SECCOMP}
make: *** [Makefile:93: /home/runner/work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/rootfs-initrd/builddir/initrd-image/.alpine_rootfs.done] Error 2
make[2]: *** [tools/packaging/kata-deploy/local-build/Makefile:34: rootfs-initrd-tarball-build] Error 1
make[2]: Leaving directory '/home/runner/work/kata-containers/kata-containers'
make[1]: *** [/home/runner/work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/Makefile:56: rootfs-initrd-tarball] Error 2
make[1]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/home/runner/work/kata-containers/kata-containers'
make[1]: Leaving directory '/home/runner/work/kata-containers/kata-containers'
make: *** [tools/packaging/kata-deploy/local-build/Makefile:22: all-parallel] Error 2
Error: Process completed with exit code 2.
```

ryansavino:
[...]

Issue #4786, opened 2022-08-01
CI kata-deploy failing to make kata tarball due to libseccomp linking issue


ryansavino:
# Description of problem

CI kata-deploy test for `make-kata-tarball` fails for branch `main`.

Testing command used to reproduce (directory: kata-containers):
```
make kata-tarball
```

# Expected result

kata tarball should be created.

# Actual result

The kata-agent fails to build during the tarball creation.
tools/packaging/kata-deploy/local-build/Makefile
Target: rootfs-initrd-tarball-build

```
error: linking with `cc` failed: exit status: 1
  |
[...]

Issue #4390, opened 2022-06-03
agent: "unable to write to a control group file" on kernel>=5.16


dubek:
(This is a preliminary report, I'll be able to add more details and kata-collect-data output next week. I wondered if anyone noticed this as well.)

# Description of problem

Agent doesn't start if the guest VM has kernel 5.16 or later.

# Expected result

Agent starts with newer kernels (>= 5.16)

# Actual result

```
{"msg":"cgroup manager set cpu","level":"INFO","ts":"2022-05-31T19:57:24.957559869+00:00","pid":"1","version":"0.1.0","source":"agent","subsystem":"cgroups","name":"kata-agent"}
{"msg":"cgroup manager set memory","level":"INFO","ts":"2022-05-31T19:57:24.957577181+00:00","version":"0.1.0","pid":"1","source":"agent","subsystem":"cgroups","name":"kata-agent"}
{"msg":"create container process error unable to write to a control group file caused by: Os { code: 95, kind: Uncategorized, message: \"Not supported\" }","level":"ERRO","ts":"2022-05-31T19:57:24.957629159+00:00","subsystem":"container","module":"rustjail","eid":"e22a347dcc0d7ccbbfc6f8684b47c8bca0f5da69296378d7727d7f038cef0a75","source":"agent","name":"kata-agent","cid":"e22a347dcc0d7ccbbfc6f8684b47c8bca0f5da69296378d7727d7f038cef0a75","version":"0.1.0","pid":"1"}
{"msg":"child exited unexpectedly","level":"INFO","ts":"2022-05-31T19:57:24.957651851+00:00","pid":"1","version":"0.1.0","child-pid":"66","name":"kata-agent","source":"agent","subsystem":"signals"}
{"msg":"read child process log end","level":"INFO","ts":"2022-05-31T19:57:24.957763361+00:00","eid":"e22a347dcc0d7ccbbfc6f8684b47c8bca0f5da69296378d7727d7f038cef0a75","version":"0.1.0","module":"rustjail","subsystem":"container","name":"kata-agent","pid":"1","action":"child process log","source":"agent","cid":"e22a347dcc0d7ccbbfc6f8684b47c8bca0f5da69296378d7727d7f038cef0a75"}
```

# Further information

Specifically the call that fails is `set_kmem_limit()` call in `src/agent/rustjail/src/cgroups/fs/mod.rs`.

[...]

PR #4861, opened 2022-08-09
kernel: upgrade guest kernel support to 5.19

### LINKED ISSUE #4860: Upgrade guest kernel support to 5.19
Guest kernel used can be upgraded to newly released 5.19.

**How can it be improved?**
Upgrading the guest kernel provides fixes for known issues and support for new features.

ryansavino:
Fixes #4860

kernel: upgrade guest kernel support to 5.19


ryansavino:
This is blocked by #4849. That change is needed to build the 5.19 .config without any issues.

tools/packaging/kernel/build-kernel.sh
```diff
@@ -144,7 +144,7 @@ get_kernel() {
 	[ -n "${kernel_path}" ] || die "kernel_path not provided"
 	[ ! -d "${kernel_path}" ] || die "kernel_path already exist"
 
-	if [ "${conf_guest}" != "" ]; then
+	if [[ "${conf_guest}" != "" && "${conf_guest}" != "sev" ]]; then
```
fidencio, line 147: I'm not sure if I understand this part here.  Why we would like to **not** get the kernel version from the `tee` entry in case of SEV?

tools/packaging/kernel/configs/fragments/x86_64/mmu.conf
```diff
@@ -1,4 +1,5 @@
 # x86 specific mmu/memory related items
 
 # Remove the kernel mapping from the user space - security improvement.
+CONFIG_SPECULATION_MITIGATIONS=y
```
fidencio, line 4: You're already setting this as part part of the base.conf

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
```diff
@@ -1,12 +1,9 @@
...
 CONFIG_AMD_MEM_ENCRYPT=y
-CONFIG_AMD_MEM_ENCRYPT_ACTIVE_BY_DEFAULT=y
-CONFIG_CRYPTO_DEV_SP_PSP=y
-CONFIG_CRYPTO_DEV_CCP=y
-CONFIG_SECURITYFS=y
-CONFIG_VIRT_DRIVERS=y
 CONFIG_EFI=y
 CONFIG_EFI_SECRET=m
+CONFIG_EFI_STUB=y
+CONFIG_MODULES=y
```
fidencio, line 7: Please, keep this sorted alphabetically.

versions.yaml
```diff
@@ -153,16 +153,15 @@ assets:
   kernel:
     description: "Linux kernel optimised for virtual machines"
     url: "https://cdn.kernel.org/pub/linux/kernel/v5.x/"
-    version: "v5.15.48"
+    version: "v5.19"
```
fidencio, line 156: I'd prefer one commit bumping the kernel version, and another commit bumping the version for SEV, it'd be cleaner which change is related to each one of the targets.

fidencio (changes requested):
I've left some commets, @ryansavino.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
jongwu: Just out of curiosity. If kata always use direct kernel boot, what does efi stub kernel use for?

tools/packaging/kernel/build-kernel.sh
ryansavino: SEV can use the same base 5.19 kernel as general kata as all changes required for the specific kernel fork have been upstreamed. `get_tee_kernel` has some specifics regarding retrieving the version and associating it with the name of the tarball and the link to retrieve it. As kernel for SEV is the same as general kata now, I thought it best to have the script treat it the same. As an alternative, I can add handling in `get_tee_kernel` to fix the download link and the tarball name for SEV specifically.

tools/packaging/kernel/configs/fragments/x86_64/mmu.conf
ryansavino: Errors were being thrown during the .config file generation due to dependency config options not being present:
`CONFIG_PAGE_TABLE_ISOLATION=y` in mmu.conf has a dependency on `CONFIG_SPECULATION_MITIGATIONS=y`.
`CONFIG_RETPOLINE=y` in base.conf has a dependency on `CONFIG_SPECULATION_MITIGATIONS=y`.

I thought it best to add the same dependency in both respective files. This would avoid any .config file generation issues if one of these files was not included in the .config generation for some reason. It has no effect on the end .config as all options are merged in and redundancies are removed.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
ryansavino: https://www.kernelconfig.io/CONFIG_EFI_STUB?q=CONFIG_EFI_STUB&kernelversion=5.19

This kernel feature allows a bzImage to be loaded directly
by EFI firmware without the use of a bootloader.

Perhaps it should be added to a different conf?

ryansavino: @fidencio, pushed the alphabetization.

versions.yaml
ryansavino: done

ryansavino:
It's worth mentioning here that the 5.15.x patches have several arm64 changes. Was wondering if someone from that side can help port these patches to the 5.19.x folder? @jongwu, got any bandwidth to help out with this? I'm suspecting certain CI tests will probably fail otherwise for arm64 testing.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
jongwu: AFAIK, there is no kernel inside kata image. Am I right? @fidencio . So, where does firmware load the EFI stub kernel?

jongwu:
> It's worth mentioning here that the 5.15.x patches have several arm64 changes. Was wondering if someone from that side can help port these patches to the 5.19.x folder? @jongwu, got any bandwidth to help out with this? I'm suspecting certain CI tests will probably fail for arm64 testing.

Ok, I will take a look at this.

jongwu:
@ryansavino -, the patches under 5.15/arm-experimental are for experimental use on Arm. It will not impact the main kernel upgrade.

tools/packaging/kernel/build-kernel.sh
fidencio: If SEV can be used from the upstream kernel, I'd recommend removing the SEV entry from the versions.yaml file.  The only reason it was added was because the upstream kernel didn't have all the needed patches.

fidencio: Hmm. Thinking a little bit more about it I'm not sure if we should have SEV support enabled by default on the guest kernel, thus having the separate path would still make sense.  I need to think a little bit more about this.

tools/packaging/kernel/configs/fragments/common/dax.conf
```diff
@@ -12,7 +12,6 @@ CONFIG_SPARSEMEM_VMEMMAP=y
 # Without these the pmem_should_map_pages() call in the kernel fails with new
 # Related to the ARCH_HAS_HMM set in the arch files.
 CONFIG_ZONE_DEVICE=y
-CONFIG_DEV_PAGEMAP_OPS=y
```
fidencio, line 15: Why are you removing this one?

tools/packaging/kernel/configs/fragments/whitelist.conf
```diff
@@ -14,4 +14,4 @@ CONFIG_HAVE_NET_DSA
 CONFIG_NF_LOG_COMMON
 CONFIG_MANDATORY_FILE_LOCKING
 CONFIG_ARM64_UAO
-CONFIG_VFIO_MDEV_DEVICE
+CONFIG_VFIO_MDEV_DEVICE
```
fidencio, line 17: This change is not needed / shouldn't be part of the PR.

tools/packaging/kernel/kata_config_version
```diff
@@ -1 +1 @@
-94
+96
```
fidencio, line 1: 95, but I'd do the bump in the next commit.

fidencio (changes requested):
For the "kernel: upgrade guest kernel support to 5.19", I'd suggest the following commit message:
```
kernel: Upgrade guest kernel support to 5.19

Let's update to the latest 5.19.x released kernel, and while doing this,
let's also enable the `CONFIG_SPECULATION_MITIGATIONS`, which is
requiired to enable `RETPOLINE and `PAGE_TABLE_ISOLATION`.

Fixes #4860
```

tools/packaging/kernel/build-kernel.sh
fidencio: So, no.  The kernel is the same, but the options for `SEV` are only being applied in case of `conf_guest` is set to `sev`, and I think we should keep it that way for the time being.

So, please, remove this part of the commit.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
```diff
@@ -1,12 +1,9 @@
 # AMD Secure Encrypted Virtualization (SEV)
- 
+
```
fidencio, line 2: Not related to the PR.

fidencio (changes requested):
I'd do the version bump as part of the "kernel: SEV guest kernel upgrade to 5.19" commit.

And I suggest the following commit message:
```
kernel: Update SEV guest kernel to 5.19.1

Kernel 5.19.1 has all the needed patches for running SEV, thus let's update it
and stop using the version coming from confidential-containers.
```

Also, I think this one also deserves an issue as that will help folks tracking the confidential-containers effort to easily pinpoint this change when needed.

versions.yaml
fidencio: Also, please, update to v5.19.1

versions.yaml
```diff
@@ -159,10 +159,9 @@ assets:
       url: "https://github.com/intel/linux-kernel-dcp/archive/refs/tags"
       tag: "SPR-BKC-PC-v9.6"
     sev:
-      description: "Linux kernel with efi_secret support"
-      url: "https://github.com/confidential-containers-demo/\
-      linux/archive/refs/tags/"
-      tag: "efi-secret-v5.17-rc6"
+      description: "Linux kernel with SEV support"
+      url: "https://cdn.kernel.org/pub/linux/kernel/v5.x/"
+      tag: "v5.19"
```
fidencio, line 164: Also update this one to v5.19.1

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
dubek: My understanding is:

In direct boot (with `-kernel`), QEMU passes the content of the kernel (and initrd, and cmdline) to the guest via the fw_cfg interface. OVMF (firmware) reads the kernel from that interface, places it in memory, and jumps to the kernel's EFI stub.

Since SEV (and TDX) must have OVMF (or TDVF), the kernels for these guests must have `CONFIG_EFI_STUB=y`.

versions.yaml
dubek: I think it should be `version:` instead of `tag:`, because you're downloading from the CDN (and not from github).

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
```diff
@@ -1,12 +1,9 @@
...
 CONFIG_AMD_MEM_ENCRYPT=y
-CONFIG_AMD_MEM_ENCRYPT_ACTIVE_BY_DEFAULT=y
-CONFIG_CRYPTO_DEV_SP_PSP=y
-CONFIG_CRYPTO_DEV_CCP=y
-CONFIG_SECURITYFS=y
```
dubek, line 7: I understand the CRYPTO settings are for the host and not needed in the guest; good.

What about CONFIG_SECURITYFS? Our efi_secret module (`CONFIG_EFI_SECRET=m`) uses it. Is it enabled in some other place?

tools/packaging/kernel/kata_config_version
ryansavino: done

versions.yaml
ryansavino: done

tools/packaging/kernel/configs/fragments/common/dax.conf
ryansavino: CONFIG_DEV_PAGEMAP_OPS is no longer configurable: https://www.kernelconfig.io/config_dev_pagemap_ops

Will add to commit message.

tools/packaging/kernel/configs/fragments/whitelist.conf
ryansavino: Added back the newline. Done.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
ryansavino: Re-added the space on the line.

ryansavino: Also, updated the commit message.

versions.yaml
ryansavino: Version updated. @dubek you mind if we address the version/tag item in a separate issue?

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
ryansavino: @dubek I was attempting to cleanup the sev.conf by removing all unnecessary CONFIG options. I confirmed that the list I currently provide in this PR is the minimal set, and I successfully booted an SEV kata container using a 5.19 guest. I'm using the command `./build-kernel.sh -x sev setup` to generate the .config. CONFIG_SECURITYFS is making it's way into the final .config. I'm not sure if something else is adding it, or it's there by default now? I'm happy to undo it's removal in this PR if you feel it necessary.

tools/packaging/kernel/build-kernel.sh
ryansavino: I've removed this `conf_guest` check for `sev` and added a check in `get_tee_kernel` to see if `kernel_url` contains the string `cdn.kernel.org`. If it finds the generic kernel URL, it prepends `linux-` at the front or the tarball name. Tried to do this in the most generic way possible. Let me know if you'd like me to modify it differently.

tools/packaging/kernel/configs/fragments/x86_64/sev/sev.conf
ryansavino:
```
root@rysavino-ubuntu-02:~/kata-containers/tools/packaging/kernel/configs# grep -R "CONFIG_SECURITYFS" .
./arm64_kata_kvm_4.14.x:# CONFIG_SECURITYFS is not set
./arm64_kata_kvm_virtio-fs-v0.3.x:# CONFIG_SECURITYFS is not set
./powerpc_kata_kvm_4.19.x:# CONFIG_SECURITYFS is not set
./arm64_kata_kvm_4.19.x:# CONFIG_SECURITYFS is not set
./s390_kata_kvm_5.4.x:CONFIG_SECURITYFS=y
./fragments/x86_64/.config:CONFIG_SECURITYFS=y
./arm64_kata_kvm_5.4.x:# CONFIG_SECURITYFS is not set
./powerpc_kata_kvm_5.4.x:# CONFIG_SECURITYFS is not set
./x86_64_kata_kvm_4.14.x:# CONFIG_SECURITYFS is not set
./s390_kata_kvm_4.19.x:# CONFIG_SECURITYFS is not set
./powerpc_kata_kvm_4.14.x:# CONFIG_SECURITYFS is not set
```

Maybe default option? From above search it's making its way into fragments/x86_64/.config, but isn't being set in any conf that would affect x86_64.

dubek: If it's in the final `.config` I think it's fine for now (maybe CONFIG_EFI_SECRET selects it, I need to look it up).

ryansavino:
`kata deploy build / make-kata-tarball (pull_request)` is failing due to #4786 and #4787

tools/packaging/kernel/configs/fragments/x86_64/mmu.conf
fidencio: Please. drop this one from here.
The `base` config will **always** be used, not making much sense to have this one duplicated here.

fidencio: And also from the commit message.

fidencio:
@ryansavino, just a minor comment about dropping the duplicated config, as it's not needed.

Please, update that, including the commit message, and this one will be ready to be merged.

fidencio (approved):
lgtm, thanks @ryansavino!

fidencio:
cc @wainersm for a look on this as well.

fidencio:
@ryansavino, the kernel upgrade is surely causing some issues with the CI.  Do you mind debugging it?

dubek:
Hmmm, I found this error in the log of jenkins-ci-ubuntu-20.04 :

```
Running nydus integration tests
Image is up to date for sha256:dcac4621b8636ac822ab7737648a272fcf378b8df6d315816036f00d5536d1e2
Pod 6bc88318657fef6d6855c02dd1fdf5a2a094731ce210e406059ad9d3f7cd4ce8 created
Container d44191eddc024a568c3c944a8c2d57f18308a8550a6a30b51389441710749041 created
time="2022-08-18T07:21:50Z" level=fatal msg="starting the container \"d44191eddc024a568c3c944a8c2d57f18308a8550a6a30b51389441710749041\": rpc error: code = Unknown desc = failed to create containerd task: failed to create shim: unable to write to a control group file caused by: Os { code: 95, kind: Uncategorized, message: \"Not supported\" }: unknown"
```

Maybe related to #4390 (affects linux >= 5.16)? I already submitted the fix and it was merged in a new cgroups-rs release, but kata-agent wasn't updated yet.

ryansavino:
> @ryansavino, the kernel upgrade is surely causing some issues with the CI. Do you mind debugging it?

Sure, let me look.

ryansavino:
Confirmed that I can create test-lifecycle-events.yaml with no errors after upgrading agent cgroup-rs crate to 0.2.10. Added commit and pushed. Thanks for the tip @dubek.

fidencio:
@BbolroC, any chance to help us to understand the failures in the s390x CI?

wainersm:
> @BbolroC, any chance to help us to understand the failures in the s390x CI?

This PR deleted [tools/packaging/kernel/configs/fragments/s390/network.conf](https://github.com/kata-containers/kata-containers/pull/4861/commits/32960d340905b57cf08c16e1be866e1a2f8589bf#diff-3d519fef6459ab438b480a7d6a049ee2b25a252a9d1f1cdcd5d03dcbc9066af4)

By mistake @ryansavino ?

wainersm:
> > @BbolroC, any chance to help us to understand the failures in the s390x CI?
> 
> This PR deleted [tools/packaging/kernel/configs/fragments/s390/network.conf](https://github.com/kata-containers/kata-containers/pull/4861/commits/32960d340905b57cf08c16e1be866e1a2f8589bf#diff-3d519fef6459ab438b480a7d6a049ee2b25a252a9d1f1cdcd5d03dcbc9066af4)
> 
> By mistake @ryansavino ?

It seems `CONFIG_PACK_STACK ` is still supported by s390. Where did you find it is not?

tools/packaging/kernel/build-kernel.sh
wainersm: @ryansavino I've a suggestion to address the tarball name issue. I will comment in line.

tools/packaging/kernel/build-kernel.sh
```diff
@@ -127,7 +127,12 @@ get_tee_kernel() {
 	mkdir -p ${kernel_path}
 
 	[ -z "${kernel_url}" ] && kernel_url=$(get_from_kata_deps "assets.kernel.${tee}.url")
-	kernel_tarball="${version}.tar.gz"
+
```
wainersm, line 130: My suggestion is to assume `kernel_tarball="linux-${version}.tar.gz"` is expected, otherwise add the 'tarball' property in the `versions.yml`. Something like:

```
kernel_tarball="linux-${version}.tar.gz"
tarball_name=$(get_from_kata_deps "assets.kernel.${tee}.tarball")
[ -z "$tarball_name" ] || kernel_tarball="$tarball_name"
```
Then change the TDX properties in `versions.yaml`:
```
diff --git a/versions.yaml b/versions.yaml
index eb6753293..b76f6e104 100644
--- a/versions.yaml
+++ b/versions.yaml
@@ -158,6 +158,7 @@ assets:
       description: "Linux kernel that supports TDX"
       url: "https://github.com/intel/linux-kernel-dcp/archive/refs/tags"
       tag: "SPR-BKC-PC-v9.6"
+      tarball: "SPR-BKC-PC-v9.6.tar.gz"
     sev:
       description: "Linux kernel with efi_secret support"
       url: "https://github.com/confidential-containers-demo/\
```

tools/packaging/kernel/build-kernel.sh
```diff
@@ -553,7 +558,8 @@ main() {
 			esac
 		elif [[ "${conf_guest}" != "" ]]; then
 			#If specifying a tag for kernel_version, must be formatted version-like to avoid unintended parsing issues
-			kernel_version=$(get_from_kata_deps "assets.kernel.${conf_guest}.tag")
+			kernel_version=$(get_from_kata_deps "assets.kernel.${conf_guest}.version" 2>/dev/null || true)
+			[ -z "${kernel_version}" ] && kernel_version=$(get_from_kata_deps "assets.kernel.${conf_guest}.tag")
```
wainersm, line 562: Avoid the use of `A && B` because that expression can return non-zero status. Instead use `! A || B` which always return zero (ah, unless `B` return fails.. which is not the case here).

wainersm:
@ryansavino I don't have more to say than what @fidencio and @dubek already said.

BbolroC:
@fidencio Hi, unfortunately due to the internal situation, I am currently supporting other projects urgently. I could only fix any broken CI jobs caused by infra problems for the time being (maybe for a month at least). Hopefully I could come back as fast as possible.

fidencio:
> > > @BbolroC, any chance to help us to understand the failures in the s390x CI?
> > 
> > 
> > This PR deleted [tools/packaging/kernel/configs/fragments/s390/network.conf](https://github.com/kata-containers/kata-containers/pull/4861/commits/32960d340905b57cf08c16e1be866e1a2f8589bf#diff-3d519fef6459ab438b480a7d6a049ee2b25a252a9d1f1cdcd5d03dcbc9066af4)
> > By mistake @ryansavino ?
> 
> It seems `CONFIG_PACK_STACK ` is still supported by s390. Where did you find it is not?

Nice catch, I missed this.

tools/packaging/kernel/build-kernel.sh
ryansavino: done

tools/packaging/kernel/build-kernel.sh
ryansavino: done

wainersm:
> > > > @BbolroC, any chance to help us to understand the failures in the s390x CI?
> > > 
> > > 
> > > This PR deleted [tools/packaging/kernel/configs/fragments/s390/network.conf](https://github.com/kata-containers/kata-containers/pull/4861/commits/32960d340905b57cf08c16e1be866e1a2f8589bf#diff-3d519fef6459ab438b480a7d6a049ee2b25a252a9d1f1cdcd5d03dcbc9066af4)
> > > By mistake @ryansavino ?
> > 
> > 
> > It seems `CONFIG_PACK_STACK ` is still supported by s390. Where did you find it is not?
> 
> Nice catch, I missed this.

I was chatting with @ryansavino on private about it. On https://www.kernelconfig.io/CONFIG_PACK_STACK?q=CONFIG_PACK_STACK&kernelversion=5.19.2 it is not clear whether that config is still supported or not. If that config is include then the `.config` is not generated as you can see in http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-main-PR/2210/consoleFull

Omitting that config, it builds the kernel but fails to copy the `vmlinux` file at  
https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kernel/build-kernel.sh#L458 . That script expects the vmlinux on `arch/s390/boot/compressed/vmlinux`...

ryansavino:
I'll need access to an s390 system in order to debug further. Looks like the `jenkins-ci-s390x-unit-ubuntu-20-04` job is failing because it can't find the C++ compiler. The `jenkins-ci-s390x-main-ubuntu-20-04` and the `jenkins-ci-s390x-snap-ubuntu-20-04` jobs are failing because 'arch/s390/boot/compressed/vmlinux' isn't present. @wainersm any way I can get access to the s390 CI node directly?

BbolroC:
@fidencio @wainersm I am supporting Ryan atm for debugging. Hopefully, we could make the best out of it. :wink:

BbolroC:
@ryansavino https://github.com/kata-containers/kata-containers/blob/main/snap/snapcraft.yaml#L209 should be changed as well?

ryansavino:
Opened a tests issue (https://github.com/kata-containers/tests/issues/5047) regarding failing `jenkins-ci-s390x-unit-ubuntu-20-04`.

fidencio:
Once this one lands, we will need to raise a backport for the 5.19 bump, the SEV work should **not** be backported though.

wainersm (approved):

wainersm:
> Once this one lands, we will need to raise a backport for the 5.19 bump, the SEV work should **not** be backported though.

hi @fidencio , I didn´t get why we need to backport the kernel update. Is it fixing bugs?

fidencio:
> > Once this one lands, we will need to raise a backport for the 5.19 bump, the SEV work should **not** be backported though.
> 
> hi @fidencio , I didn´t get why we need to backport the kernel update. Is it fixing bugs?

It brings in the `CONFIG_SPECULATION_MITIGATIONS`, which would be welcome in the stable branch.

dubek:
> It brings in the `CONFIG_SPECULATION_MITIGATIONS`, which would be welcome in the stable branch.

I think we can update the stable branch to latest in 5.15.y -- it should have the RETbleed mitigations (5.15 is the most recent LTS kernel). (but 5.15 doesn't have efi_secret kernel module which is needed for SEV support)


On another subject: @fidencio  How do we get the changes from this PR in the CCv0 branch? Do we regularly rebase CCv0 on top of main?

dubek:
I see that `main` was merged into branch `CCv0`, so we now have 5.19.2 in CCv0 as well. (Thanks @Megan-Wright )

Issue #4847, opened 2022-08-08
memory hot plug may fail if not enough guest memory


egernst:
A certain amount of memory is required in order to be able to online hot plugged memory. 

```
vmconsole="[ 3.277869] ------------[ cut here ]------------"
vmconsole="[ 3.278033] WARNING: CPU: 1 PID: 7 at arch/x86/mm/init_64.c:852 add_pages+0x5c/0x60"
vmconsole="[ 3.278122] Modules linked in:"
vmconsole="[ 3.278178] CPU: 1 PID: 7 Comm: kworker/u48:0 Not tainted kata-kernel #1"
vmconsole="[ 3.278267] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS [rel-1.14.0-0-g155821a1990b-prebuilt.qemu.org](http://rel-1.14.0-0-g155821a1990b-prebuilt.qemu.org/) (http://rel-1.14.0-0-g155821a1990b-prebuilt.qemu.org/) 04/01/2014“
vmconsole="[ 3.278409] Workqueue: kacpi_hotplug acpi_hotplug_work_fn"
vmconsole="[ 3.278473] RIP: 0010:add_pages+0x5c/0x60"
vmconsole="[ 3.278538] Code: 26 48 89 15 b6 35 90 01 48 b9 00 00 00 00 80 88 ff ff 48 89 15 b5 35 90 01 48 c1 e2 0c 48 01 ca 48 89 15 57 34 90 01 5b 5d c3 <0f> 0b eb b6 0f 1f 44 00 00 41 55 49 89 cd 41 54 41 89 fc 55 48 89"
vmconsole="[ 3.278771] RSP: 0018:ffffc90000047d10 EFLAGS: 00010282"
vmconsole="[ 3.278867] RAX: 00000000fffffff4 RBX: 0000000000120000 RCX: 0000000000000000"
vmconsole="[ 3.278982] RDX: 0000000000000000 RSI: 0000000000000000 RDI: ffffffffb510001e"
vmconsole="[ 3.279061] RBP: 0000000003700000 R08: 0000000000000000 R09: ffffea00002b3f48"
vmconsole="[ 3.279203] R10: 0000000000000000 R11: 0000000000000000 R12: 0000003700000000"
vmconsole="[ 3.279281] R13: 0000000000000000 R14: 0000000000000000 R15: ffffffffb5545080"
[...]

PR #4862, opened 2022-08-10
Address Memory hotplug limitation

egernst:
If we're using ACPI hotplug for memory, there's a limitation on the
amount of memory which can be hotplugged at a single time.

During hotplug, we'll allocate memory for the memmap for each page,
resulting in a 64 byte per 4KiB page allocation. As an example, hotplugging 12GiB
of memory requires ~192 MiB of *free* memory, which is about the limit
we should expect for an idle 256 MiB guest (conservative heuristic of 75%
of provided memory).

From experimentation, at pod creation time we can reliably add 48 times
what is provided to the guest. (a factor of 48 results in using 75% of
provided memory for hotplug). Using prior example of a guest with 256Mi
RAM, 256 Mi * 48 = 12 Gi; 12GiB is upper end of what we should expect
can be hotplugged successfully into the guest.

Note: It isn't expected that we'll need to hotplug large amounts of RAM
after workloads have already started -- container additions are expected
to occur first in pod lifecycle. Based on this, we expect that provided
memory should be freely available for hotplug.

If virtio-mem is being utilized, there isn't such a limitation - we can
hotplug the max allowed memory at a single time.

Fixes: https://github.com/kata-containers/kata-containers/issues/4847

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -77,6 +77,14 @@ const (
 
 	// Restricted permission for shared directory managed by virtiofs
 	sharedDirMode = os.FileMode(0700) | os.ModeDir
+
+	// hotplug factor indicates how much memory can be hotplugged relative to the amount of
+	// RAM provided to the guest. This is a conservative heuristic based on needing 64 bytes per
+	// 4KiB page of hotplugged memory.
+	//
+	// As an example: 12 GiB hotplugged -> 3 Mi pages -> 192 MiBytes overhead (3Mi x 64B).
+	// This is approximately what should be free in a relatively unloaded 256 MiB guest (75% of available memory). So, 256 Mi x 48 => 12 Gi
+	acpiMemoryHotplugFactor = 48
 )
 
 var (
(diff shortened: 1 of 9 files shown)


src/runtime/virtcontainers/sandbox.go
```diff
@@ -2012,9 +2012,60 @@ func (s *Sandbox) updateResources(ctx context.Context) error {
...
 
-	// Update Memory
-	s.Logger().WithField("memory-sandbox-size-byte", sandboxMemoryByte).Debugf("Request to hypervisor to update memory")
+	// Update Memory --
+	// If we're using ACPI hotplug for memory, there's a limitation on the amount of memory which can be hotplugged at a single time.
+	// We must have enough free memory in the guest kernel to cover 64bytes per (4KiB) page of memory added for mem_map.
+	// See https://github.com/kata-containers/kata-containers/issues/4847 for more details.
+	// For typical pod lifecycle, we expect that each container is added then we start the workloads. based on this, we'll "assume" that majority
+	// of the guest memory is readily available. From experimentation, we see that we can add approximately 48 times what is already provided to
+	// the guest workload. For example, a 256 MiB guest should be able to accomodate hotplugging 12 GiB  of memory.
```
mcastelino, line 2021: Can this be impacted by amount of memory currently in use within the guest. In our case we do the hot plug really early, so it should not be an issue. But if we hot plug later we may need to take that into consideration.

egernst: Yes, for sure. If you later try to add a very large container, and the others are using near their limit, there's a good chance the hot plug will fail.  Given expected pod lifecycle, I don't expect this scenario.  This PR should address the existing common scenario, but we're using a heuristic here.  Ideally, when triggering the hot plug event we'd have a better signal to see if we were able to actually add the memory or not.  This is disjoint today.

egernst:
I noticed that the updateResources unit tests are very much lacking, and I think broken. keeping this PR WIP as I verify the tests  behavior. Works will locally e2e :)

cmaf (commented):
@egernst if I am understanding it correctly, using an`acpiMemoryHotplugFactor` with a higher number (the ones from the table in the issue vs the example in the comments) would result in a higher failure rate, vs the conservatively computed one which is less likely to fail at the cost of it potentially requiring an additional hotplug?

Is there a reason we shouldn't just compute the maximum hotpluggable memory based on the known memory overhead and the current size (or expected free memory) rather than a heuristic based on a 256 MB guest? (Asking based on the factors you calculated from different guest memory, seen in the issue)

egernst:
Hey Chelsea -- thanks for the review help. Indeed, 48 lined up closely to the 256 MiB case I described in the issue, and aligned with the expected overhead based on insights from David Hildenbrand.

Realistically, if an end user is adding a (very) large container, they'll need to break it up into 2 individual hot plugs either way. I'd rather rely on 75% of the existing RAM rather than pushing for a more aggressive number to: (1) keep things simple, (2) be more conservative.

This'll be more foolproof when we can rely on CRI passing the total sizing information so we can avoid the initial hot plug in the first place, but for now, I want to have a workaround in place for folks who aren't on >=1.23 version of Kubernetes.

cmaf:
@egernst makes sense, lgtm

bergwolf:
@egernst IIUC, it also depends on if the guest has hugepage enabled or not. So it is not always the same linear calculation.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -77,6 +77,14 @@ const (
 
 	// Restricted permission for shared directory managed by virtiofs
 	sharedDirMode = os.FileMode(0700) | os.ModeDir
+
+	// hotplug factor indicates how much memory can be hotplugged relative to the amount of
+	// RAM provided to the guest. This is a conservative heuristic based on neededing 64 bytes per
```
littlejawa, line 82: nit: s/neededing/needing/

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2012,9 +2020,60 @@ func (s *Sandbox) updateResources(ctx context.Context) error {
 	}
 	s.Logger().Debugf("Sandbox CPUs: %d", newCPUs)
 
-	// Update Memory
-	s.Logger().WithField("memory-sandbox-size-byte", sandboxMemoryByte).Debugf("Request to hypervisor to update memory")
+	// Update Memory --
+	// If we're using ACPI hotplug for memory, there's a limitation on the amount of memory which can be hotplugged at a single time.
+	// We must have enough free memory in the guest kernel to cover 64bytes per (4KiB) page of memory added for mem_map.
+	// See https://github.com/kata-containers/kata-containers/issues/4847 for more details.
+	// For typical pod lifecycle, we expect that each container is added then we start the workloads. based on this, we'll "assume" that majority
```
littlejawa, line 2027: nit:
- s/then/when/
- s/based/Based/

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2012,9 +2020,60 @@ func (s *Sandbox) updateResources(ctx context.Context) error {
...
+	// of the guest memory is readily available. From experimentation, we see that we can add approximately 48 times what is already provided to
+	// the guest workload. For example, a 256 MiB guest should be able to accommodate hotplugging 12 GiB  of memory.
```
littlejawa, line 2029: nit: additional space between "GiB" and "of"

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2012,9 +2020,60 @@ func (s *Sandbox) updateResources(ctx context.Context) error {
...
+
+	hconfig := s.hypervisor.HypervisorConfig()
+
+	for {
+		currentMemoryMB := s.hypervisor.GetTotalMemoryMB(ctx)
+
+		maxhotPluggableMemoryMB := currentMemoryMB * acpiMemoryHotplugFactor
+
+		// In the case of virtio-mem, we don't have a restriction on how much can be hotplugged at
+		// a single time. As a result, the max hotpluggable is only limited my the maximum memory size
```
littlejawa, line 2044: nit: s/my/by/

littlejawa (approved):
Found some typos in comments, but no issue in the code.
LGTM

egernst:
I'm seeing some issues where the guest becomes unresponsive when carrying out the on-lining which may be related to the failure that's seen in azure for CI.  Debugging now....

egernst:
Seeing:
```
not ok 2 Running within memory constraints
# (in test file k8s-memory.bats, line 49)
```

egernst:
🤦🏻  - I identified original patchset had issue when the added memory wasn't 128MiB aligned, the delta between what we have and what we'd want is negative, but we comparing unsigned.  Fixed it up and re-pushed.

/test

egernst:
> @egernst IIUC, it also depends on if the guest has hugepage enabled or not. So it is not always the same linear calculation.

Indeed, if you're using 2MB pages, assuming its still 64B per page, you'd have a much different number. Page size can vary with huge pages -- I wonder if for a first pass its easiest to just use the most conservative number (4KiB) for pages? I can foresee this getting complicated if we want to find the "best" amount allowed for hot plug for each potential page size?

fengwang666 (approved):
LGTM! Thanks!

bergwolf:
> Indeed, if you're using 2MB pages, assuming its still 64B per page, you'd have a much different number. Page size can vary with huge pages -- I wonder if for a first pass its easiest to just use the most conservative number (4KiB) for pages? I can foresee this getting complicated if we want to find the "best" amount allowed for hot plug for each potential page size?

Yeah, I can live with it as a first pass. In the longer term, we should return the hugepage information in `GuestDetailsResponse`

bergwolf (approved):
lgtm, thanks!

Issue #4792, opened 2022-08-02
surpport hugepage_size with clh


liangxianlong:
**Which feature do you think can be improved?**

clh support hugepage，but it use default hugepage_size from system only.

**How can it be improved?**

Provides a configuration item and annotation to specify the huge page type.
annotations:
```toml
"io.katacontainers.config.hypervisor.hugepage_size": "1G"
```
configuration:
```toml
enable_annotations = ["hugepage_size"]

# hugepage size for VM RAM. The size is arch related
# For amd64: ["2M", "1G"]
# For aarch64: 4KB PageSize ["2M", "1G"], 64KB PageSize ["512M"]
# If no huge page size is supplied the system's default huge page size is used
hugepage_size = "2M"
```

PR #4874, opened 2022-08-11
runtime: surpport hugepage_size with clh

liangxianlong:
clh support hugepage，but it use default hugepage_size from system only.Provides a configuration item and annotation to specify the huge page type.

Fixes: #4792

Signed-off-by: liangxianlong <liang.xianlong@zte.com.cn>


src/runtime/virtcontainers/hypervisor.go
```diff
@@ -894,6 +909,16 @@ func genericAvailableGuestProtections() (protections []string) {
...
+func GetHugepageTypes() []string {
+	arch := runtime.GOARCH
+	if arch == "amd64" {
+		return []string{HugepageSize2M, HugepageSize1G}
+	} else if arch == "arm64" {
+		return []string{HugepageSize2M, HugepageSize512M, HugepageSize1G}
+	}
+	return nil
+}
+
```
pmores, line 921: All `GetHugepageType()` invocations so far follow this pattern:

```
	hugepageTypes := GetHugepageTypes()
	for _, t := range hugepageTypes {
		if t == valueBeingChecked { do something }
	}
```
To reduce the amount of lowlevel iteration and branching code, how about adding something like
```
func isHugepageSizeValid(sz string) bool {
	hugepageTypes := GetHugepageTypes()
	for _, t := range hugepageTypes {
		if t == sz { return true }
	}
	return false
}
```
(Please don't mind the formatting, it's kind of tedious to write code in a comment. :-))

src/runtime/pkg/oci/utils.go
```diff
@@ -611,6 +611,16 @@ func addHypervisorMemoryOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig
...
 	}
 
+	// HugepageSize valid only HugePages is true
+	if value, ok := ocispec.Annotations[vcAnnotations.HugepageSize]; sbConfig.HypervisorConfig.HugePages && ok {
+		hugepageTypes := vc.GetHugepageTypes()
+		for _, t := range hugepageTypes {
+			if value == t {
+				sbConfig.HypervisorConfig.HugepageSize = value
+			}
+		}
```
pmores, line 621: With the `isHugepageSizeValid()` function proposed below, this would reduce to

```
if isHugepageSizeValid(value) {
	sbConfig.HypervisorConfig.HugepageSize = value
}
```
right?

src/runtime/virtcontainers/clh.go
```diff
@@ -479,6 +479,19 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
...
+	if *clh.vmconfig.Memory.Hugepages {
+		hugepageTypes := GetHugepageTypes()
+		for _, t := range hugepageTypes {
+			if t == clh.config.HugepageSize {
+				if v, ok := HugePageMap[t]; ok {
+					clh.vmconfig.Memory.HugepageSize = &v
+					break
+				}
+			}
+		}
```
pmores, line 492: With the isHugepageSizeValid() function proposed below, this would reduce to
```
if isHugepageSizeValid(clh.config.HugepageSize) {
	if v, ok := HugePageMap[clh.config.HugepageSize]; ok {
		clh.vmconfig.Memory.HugepageSize = &v
	}
}
```
if I'm not mistaken. (And that could likely be further refactored for more clarity.)

Issue #4877, opened 2022-08-11
Use its own VERSION for runtime-rs


liubin:
Now runtime-rs has a VERSION linked to root path of this repo, that is a 2.x series version.
For 3.0, it should use its own version file.

PR #4878, opened 2022-08-11
runtime-rs: use its own VERSION file

liubin:
For runtime-rs, it should have a different VERSION
file other than that for 2.x.

Fixes: #4877


GabyCT (approved):

studychao:
```
3.0.0-rc0
```
Should this be
```
3.0.0-alpha0
```
?

src/runtime-rs/Makefile
```diff
@@ -374,7 +374,7 @@ vendor:
 ##TARGET check: run test
 check: $(GENERATED_FILES) standard_rust_check
 
-##TARGET run: build and run agent
+##TARGET run: build and run kata shim
```
jodh-intel, line 377: Presumably this PR is essentially a temporary change (since at some point we'll "switch" to 3.x), so I'm wondering if it makes sense to add a comment to the new `VERSION` file explaining that, and then doing something like:

```diff
diff --git a/src/runtime-rs/Makefile b/src/runtime-rs/Makefile
index 795399a4a..4112e4734 100644
--- a/src/runtime-rs/Makefile
+++ b/src/runtime-rs/Makefile
@@ -133,7 +133,7 @@ MONITOR = kata-monitor
 MONITOR_OUTPUT = $(CURDIR)/$(MONITOR)
 MONITOR_DIR = $(CLI_DIR)/kata-monitor
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
-VERSION := ${shell cat ./VERSION}
+VERSION := ${shell grep -v ^\# ./VERSION}

 # List of configuration files to build and install
 CONFIGS =
```

That way, we're less likely to forget to update it ;)

jodh-intel (approved):
Thanks @liubin - a random suggestion but overall...

lgtm

src/runtime-rs/Makefile
liubin: Yes, I'd like to fix this next week, so add a DNM label, thank you @jodh-intel

liubin:
Close as described as https://github.com/kata-containers/kata-containers/pull/4911#issuecomment-1216117075

Issue #4882, opened 2022-08-11
CC | Implement trusted storage for CoCo


arronwy:
After these two PR: Add cryptsetup support in Guest kernel and rootfs: https://github.com/kata-containers/kata-containers/pull/4762 and Integrate pause image inside rootfs https://github.com/kata-containers/kata-containers/pull/4769

We can implement the trusted storage for CoCo as describe in: https://github.com/confidential-containers/documentation/issues/39
Frist we will guest provided `pause` image and use `pause` sidecar container to host the block volume
Next we will use an shell script with `cryptsetup` tools to initialize the trust stroage.
Since data integrity protection will have many performance issues, we will enable it by default but introduce an config to allow the user to disable it if they have performance concerns.

PR #4769, opened 2022-07-29
CC | Integrate pause image inside rootfs

arronwy:
For CoCo stack, the pause image is managed by host side,
then it may configure a malicious pause image, we need package
a pause image inside the rootfs and don't the pause image from host.

The reasons we don't pull the pause image inside the guest, one is performance, another is the trusted storage for CoCo is not ready when launch the pause container https://github.com/confidential-containers/documentation/issues/39

Fixes: #4768

Signed-off-by: Wang, Arron <arron.wang@intel.com>

PR #4885, opened 2022-08-11
CC | Implement trusted storage for CoCo

arronwy:
After these two PR: Add cryptsetup support in Guest kernel and rootfs: https://github.com/kata-containers/kata-containers/pull/4762 and Integrate pause image inside rootfs https://github.com/kata-containers/kata-containers/pull/4769

We can implement the trusted storage for CoCo as describe in: https://github.com/confidential-containers/documentation/issues/39
Frist we will guest provided pause image and use pause sidecar container to host the block volume
Next we will use an shell script with cryptsetup tools to initialize the trust storage.
Since data integrity protection will have huge performance drop, we will enable it by default but introduce an config to allow the user to disable it if they have performance concerns.

Fixed https://github.com/kata-containers/kata-containers/issues/4882

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -27,6 +27,7 @@ const CONFIG_FILE: &str = "agent.config_file";
 const CONTAINER_POLICY_FILE: &str = "agent.container_policy_file";
 const HTTPS_PROXY: &str = "agent.https_proxy";
 const NO_PROXY: &str = "agent.no_proxy";
+const ENABLE_DATA_INTEGRITY: &str = "agent.data_integrity";
 
 const DEFAULT_LOG_LEVEL: slog::Level = slog::Level::Info;
 const DEFAULT_HOTPLUG_TIMEOUT: time::Duration = time::Duration::from_secs(3);
@@ -89,6 +90,7 @@ pub struct AgentConfig {
     pub aa_kbc_params: String,
     pub https_proxy: String,
     pub no_proxy: String,
+    pub data_integrity: bool,
 }
 
 #[derive(Debug, Deserialize)]
@@ -108,6 +110,7 @@ pub struct AgentConfigBuilder {
     pub aa_kbc_params: Option<String>,
     pub https_proxy: Option<String>,
     pub no_proxy: Option<String>,
+    pub data_integrity: Option<bool>,
 }
 
 macro_rules! config_override {
@@ -173,6 +176,7 @@ impl Default for AgentConfig {
             aa_kbc_params: String::from(""),
             https_proxy: String::from(""),
             no_proxy: String::from(""),
+            data_integrity: false,
         }
     }
 }
@@ -205,6 +209,7 @@ impl FromStr for AgentConfig {
         config_override!(agent_config_builder, agent_config, aa_kbc_params);
         config_override!(agent_config_builder, agent_config, https_proxy);
         config_override!(agent_config_builder, agent_config, no_proxy);
+        config_override!(agent_config_builder, agent_config, data_integrity);
 
         // Populate the allowed endpoints hash set, if we got any from the config file.
         if let Some(endpoints) = agent_config_builder.endpoints {
@@ -312,6 +317,12 @@ impl AgentConfig {
 
             parse_cmdline_param!(param, HTTPS_PROXY, config.https_proxy, get_url_value);
             parse_cmdline_param!(param, NO_PROXY, config.no_proxy, get_string_value);
+            parse_cmdline_param!(
+                param,
+                ENABLE_DATA_INTEGRITY,
+                config.data_integrity,
+                get_bool_value
+            );
         }
 
         if let Ok(addr) = env::var(SERVER_ADDR_ENV_VAR) {
@@ -534,6 +545,7 @@ mod tests {
             container_policy_path: &'a str,
             https_proxy: &'a str,
             no_proxy: &'a str,
+            data_integrity: bool,
         }
 
         impl Default for TestData<'_> {
@@ -552,6 +564,7 @@ mod tests {
                     container_policy_path: "",
                     https_proxy: "",
                     no_proxy: "",
+                    data_integrity: false,
                 }
             }
         }
@@ -946,6 +959,31 @@ mod tests {
                 no_proxy: "192.168.1.0/24,172.16.0.0/12",
                 ..Default::default()
             },
+            TestData {
+                contents: "",
+                data_integrity: false,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.data_integrity=true",
+                data_integrity: true,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.data_integrity=false",
+                data_integrity: false,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.data_integrity=1",
+                data_integrity: true,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.data_integrity=0",
+                data_integrity: false,
+                ..Default::default()
+            },
         ];
 
         let dir = tempdir().expect("failed to create tmpdir");
@@ -1000,6 +1038,7 @@ mod tests {
             );
             assert_eq!(d.https_proxy, config.https_proxy, "{}", msg);
             assert_eq!(d.no_proxy, config.no_proxy, "{}", msg);
+            assert_eq!(d.data_integrity, config.data_integrity, "{}", msg);
 
             for v in vars_to_unset {
                 env::remove_var(v);
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -33,6 +33,8 @@ const AA_GETRESOURCE_PORT: &str = "127.0.0.1:50001";
 const OCICRYPT_CONFIG_PATH: &str = "/tmp/ocicrypt_config.json";
 // kata rootfs is readonly, use tmpfs before CC storage is implemented.
 const KATA_CC_IMAGE_WORK_DIR: &str = "/run/image/";
+const KATA_CC_PAUSE_BUNDLE: &str = "/pause_bundle";
+const CONFIG_JSON: &str = "config.json";
 
 // Convenience macro to obtain the scope logger
 macro_rules! sl {
@@ -146,6 +148,32 @@ impl ImageService {
         Ok(())
     }
 
+    // pause image is packaged in rootfs for CC
+    fn unpack_pause_image(cid: &str) -> Result<()> {
+        let cc_pause_bundle = Path::new(KATA_CC_PAUSE_BUNDLE);
+        if !cc_pause_bundle.exists() {
+            return Err(anyhow!("Pause image not present in rootfs"));
+        }
+
+        info!(sl!(), "use guest pause image cid {:?}", cid);
+        let pause_bundle = Path::new(CONTAINER_BASE).join(&cid);
+        let pause_rootfs = pause_bundle.join("rootfs");
+        let pause_config = pause_bundle.join(CONFIG_JSON);
+        let pause_binary = pause_rootfs.join("pause");
+        fs::create_dir_all(&pause_rootfs)?;
+        if !pause_config.exists() {
+            fs::copy(
+                cc_pause_bundle.join(CONFIG_JSON),
+                pause_bundle.join(CONFIG_JSON),
+            )?;
+        }
+        if !pause_binary.exists() {
+            fs::copy(cc_pause_bundle.join("rootfs").join("pause"), pause_binary)?;
+        }
+
+        Ok(())
+    }
+
     // If we fail to start the AA, Skopeo/ocicrypt won't be able to unwrap keys
     // and container decryption will fail.
     fn init_attestation_agent() {
@@ -204,6 +232,16 @@ impl ImageService {
             }
         } else {
             verify_cid(&cid)?;
+        }
+
+        // Can switch to use cid directly when we remove umoci
+        let v: Vec<&str> = image.rsplit('/').collect();
+        if !v[0].is_empty() && v[0].starts_with("pause:") {
+            Self::unpack_pause_image(&cid)?;
+
+            let mut sandbox = self.sandbox.lock().await;
+            sandbox.images.insert(String::from(image), cid.to_string());
+            return Ok(image.to_owned());
         }
 
         if !aa_kbc_params.is_empty() {
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -89,6 +89,8 @@ pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
 const ANNO_K8S_IMAGE_NAME: &str = "io.kubernetes.cri.image-name";
 const CONFIG_JSON: &str = "config.json";
+const INIT_TRUSTED_STORAGE: &str = "/usr/bin/kata-init-trusted-storage";
+const TRUSTED_STORAGE_DEVICE: &str = "/dev/trusted_store";
 
 const IPTABLES_SAVE: &str = "/sbin/iptables-save";
 const IPTABLES_RESTORE: &str = "/sbin/iptables-restore";
@@ -216,6 +218,30 @@ impl AgentService {
         // match real devices inside the VM. This step is necessary since we
         // cannot predict everything from the caller.
         add_devices(&req.devices.to_vec(), &mut oci, &self.sandbox).await?;
+
+        let linux = oci
+            .linux
+            .as_mut()
+            .ok_or_else(|| anyhow!("Spec didn't contain linux field"))?;
+
+        for specdev in &mut linux.devices {
+            let dev_major_minor = format!("{}:{}", specdev.major, specdev.minor);
+
+            if specdev.path == TRUSTED_STORAGE_DEVICE {
+                let data_integrity = AGENT_CONFIG.read().await.data_integrity;
+                info!(
+                    sl!(),
+                    "trusted_store device major:min {}, enable data integrity {}",
+                    dev_major_minor,
+                    data_integrity.to_string()
+                );
+
+                Command::new(INIT_TRUSTED_STORAGE)
+                    .args(&[&dev_major_minor, &data_integrity.to_string()])
+                    .output()
+                    .expect("Failed to initialize confidential storage");
+            }
+        }
 
         // Both rootfs and volumes (invoked with --volume for instance) will
         // be processed the same way. The idea is to always mount any provided
--- a/tools/osbuilder/rootfs-builder/init_trusted_storage.sh
+++ b/tools/osbuilder/rootfs-builder/init_trusted_storage.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o errtrace
+
+[ -n "${DEBUG:-}" ] && set -o xtrace
+
+handle_error() {
+        local exit_code="${?}"
+        local line_number="${1:-}"
+        echo "error:"
+        echo "Failed at $line_number: ${BASH_COMMAND}"
+        exit "${exit_code}"
+}
+trap 'handle_error $LINENO' ERR
+
+die()
+{
+	local msg="$*"
+	echo >&2 "ERROR: $msg"
+	exit 1
+}
+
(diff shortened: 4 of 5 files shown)


fidencio (approved):
lgtm, @arronwy!

I'd like to bother some other people for reviewing this.
/cc @bpradipt @jodh-intel @stevenhorsman @sameo

fidencio:
A nit, @arronwy, in the commit message you mention:
"Since data integrity protection will have huge performance drop, we will enable it by default but introduce an config to allow the user to disable it if they have performance concerns."

It's worth rewriting it to make sure we will **NOT** enable it by default.

fidencio:
I'm adding the do-not-merge label just so we have the other PRs related to this also reviewed before this one gets merged.

src/agent/src/rpc.rs
```diff
@@ -85,6 +85,8 @@ pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
 const ANNO_K8S_IMAGE_NAME: &str = "io.kubernetes.cri.image-name";
 const CONFIG_JSON: &str = "config.json";
+const INIT_TRUSTED_STORAGE: &str = "init_trusted_storage.sh";
```
c3d, line 88: Don't we want to have an absolute path here?

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
c3d, line 1: Should we use bash here, or allow for a more restricted shell?

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+if [ -n "${2-}" ]; then
+        enable_data_integrity="$2"
+fi
+
+device_name=$(cat /sys/dev/block/${device_num}/uevent|grep DEVNAME|cut -d "=" -f2)
+device_path="/dev/$device_name"
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+	dd if=/dev/urandom of=/run/cc_storage.key bs=1 count=4096
+
+	if [ "$enable_data_integrity" == "false" ]; then
```
c3d, line 31: The tests above use values `0` and `1`, and I wonder what is the "specification" for that argument. Value `0` here would enable integrity, as I read it, but is it what is expected?

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+
+device_num=${1:-}
+if [ -z "$device_num" ]; then
+        echo "ERROR: invalid arguments" >&2
+        exit 1
+fi
+
+enable_data_integrity="true"
+if [ -n "${2-}" ]; then
+        enable_data_integrity="$2"
```
c3d, line 23: I would add some code here to:
1. Check that we only get `true`, `false`, `0` or `1`
2. Convert `0` to `false` and `1` to true
3. Convert any other value to `true` and issue a logged warning.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+	rm /run/cc_storage.key
+	mkfs.ext4 /dev/mapper/encrypted_disk
+
+	if [ ! -d "/run/image" ]; then
+		mkdir /run/image
+	fi
+
+	mount /dev/mapper/encrypted_disk /run/image
+else
+        echo "ERROR: invalid device" >&2
```
c3d, line 49: Output the invalid device `$device_path` here for debugging.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+        enable_data_integrity="$2"
+fi
+
+device_name=$(cat /sys/dev/block/${device_num}/uevent|grep DEVNAME|cut -d "=" -f2)
+device_path="/dev/$device_name"
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+	dd if=/dev/urandom of=/run/cc_storage.key bs=1 count=4096
+
+	if [ "$enable_data_integrity" == "false" ]; then
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
```
c3d, line 32: Absolute path for cryptsetup?

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+[ -n "${DEBUG:-}" ] && set -o xtrace
+
+device_num=${1:-}
+if [ -z "$device_num" ]; then
+        echo "ERROR: invalid arguments" >&2
+        exit 1
+fi
+
+enable_data_integrity="true"
+if [ -n "${2-}" ]; then
```
c3d, line 22: What is the purpose of `"${2-}"` here (not sure why you are not using `${2:-}` or even just `$2`)?

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+device_name=$(cat /sys/dev/block/${device_num}/uevent|grep DEVNAME|cut -d "=" -f2)
+device_path="/dev/$device_name"
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+	dd if=/dev/urandom of=/run/cc_storage.key bs=1 count=4096
+
+	if [ "$enable_data_integrity" == "false" ]; then
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
+			--cipher aes-xts-plain64 /run/cc_storage.key
```
c3d, line 33: I would put the key parameters to `cryptsetup` as variables, and possibly later add options to pass them around from configuration data.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[ -n "${BASH_VERSION:-}" ] && set -o errtrace
```
c3d, line 12: What is the `BASH_VERSION` test for?

Rationale: I'm trying to minimize the possible path variations in the behavior of a critical script. IMO, we want to either have tests checking the correct behavior in both alternatives, or else, not have the test.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+        echo "ERROR: invalid arguments" >&2
+        exit 1
+fi
+
+enable_data_integrity="true"
+if [ -n "${2-}" ]; then
+        enable_data_integrity="$2"
+fi
+
+device_name=$(cat /sys/dev/block/${device_num}/uevent|grep DEVNAME|cut -d "=" -f2)
```
c3d, line 26: Consider `sed -e 's/DEVNAME=//g;t;d' /sys/dev/block/${device_num}/uevent` to avoid launching three processes just to get the device name 😄

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+
+	if [ "$enable_data_integrity" == "false" ]; then
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
+			--cipher aes-xts-plain64 /run/cc_storage.key
+	else
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
+			 --cipher aes-xts-plain64 --integrity hmac-sha256 /run/cc_storage.key
+	fi
+
+	cryptsetup luksOpen -d /run/cc_storage.key "$device_path" encrypted_disk
```
c3d, line 39: In case of failure of cryptsetup, we get no message. The script will just exit, but it will not be very helpful.

Consider adding an error handler to display an error similar to the other ones?

```
error_handler() {
    echo "ERROR: $0:$2: Unexpected error $1 from cryptsetup, exiting"
    exit $1
}
trap 'error_handler $? $LINENO' ERR
```

c3d (approved):
Overall, it looks like good work, but there are a few minor details I'd rather fix in the script before committing (or that should be fixed as a follow-up).

c3d:
Do we have actual data on integrity performance drop? Found things [like that](https://unix.stackexchange.com/questions/591540/cryptsetup-with-dm-integrity-weird-benchmark-results), but it looks like this really depends on the exact cipher options chosen (another reason for making _that_ configurable).

c3d:
Note for later: for read-only volumes, we might want to use verity instead of integrity.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+
+enable_data_integrity="true"
+if [ -n "${2-}" ]; then
+        enable_data_integrity="$2"
+fi
+
+device_name=$(cat /sys/dev/block/${device_num}/uevent|grep DEVNAME|cut -d "=" -f2)
+device_path="/dev/$device_name"
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+	dd if=/dev/urandom of=/run/cc_storage.key bs=1 count=4096
```
c3d, line 29: Just curious if setting up a random key each time is intended to be the long-term solution, or just a temporary kludge until we get actual keys from the KBS? If the latter, probably need to add another option that distinguishes between ephemeral storage (where you can generate the key that way) and persistent storage (where you would need to get the key from the agent).

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
huoqifeng, line 39: should we give `encrypted_disk` a more specific name like `image_encrypted_disk` or `ephemeral_image_encrypted_disk` so that it does not need be changed when support persistent encrypted volume later?

src/agent/src/config.rs
```diff
@@ -27,6 +27,7 @@ const CONFIG_FILE: &str = "agent.config_file";
 const CONTAINER_POLICY_FILE: &str = "agent.container_policy_file";
 const HTTPS_PROXY: &str = "agent.https_proxy";
 const NO_PROXY: &str = "agent.no_proxy";
+const ENABLE_DATA_INTEGRITY: &str = "agent.enable_data_integrity";
```
jodh-intel, line 30: This is fine, but thinking to the future, do you think there is any chance we might want to pass additional values here? If so, maybe something like `agent.data_integrity=[true|false]` would be better for now (so we could enhance that "language" later ( `agent.data_integrity=[true|false|foo=x,bar=y,baz=(1,2,"hello,world")]`).

src/agent/src/image_rpc.rs
```diff
@@ -206,6 +208,36 @@ impl ImageService {
             verify_cid(&cid)?;
         }
 
+        // Can switch to use cid directly when we remove umoci
+        let v: Vec<&str> = image.rsplit('/').collect();
+        if !v[0].is_empty() && v[0].starts_with("pause:") {
+            let cc_pause_bundle = Path::new(KATA_CC_PAUSE_BUNDLE);
+            if !cc_pause_bundle.exists() {
+                return Err(anyhow!("Pause image not exists in rootfs"));
```
jodh-intel, line 216:
```suggestion
                return Err(anyhow!("Pause image not present in rootfs"));
```

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
jodh-intel: I think this can be simplified to `set -o errtrace` here. I tend to check `BASH_VERSION` to allow a bash script to be sourced by zsh for example, but that doesn't apply here.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
jodh-intel: That's to stop the shell immediately exiting of `$2` is *not* set.

src/agent/src/rpc.rs
jodh-intel: +1 to @c3d's comment. Maybe `/usr/bin/kata-init-trusted-storage` or similar? Also, I'd consider removing the `.sh` suffix to allow users to create their own version of this file (in their custom rootfs) but writing it in rust, say.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+			 --cipher aes-xts-plain64 --integrity hmac-sha256 /run/cc_storage.key
+	fi
+
+	cryptsetup luksOpen -d /run/cc_storage.key "$device_path" encrypted_disk
+	rm /run/cc_storage.key
+	mkfs.ext4 /dev/mapper/encrypted_disk
+
+	if [ ! -d "/run/image" ]; then
+		mkdir /run/image
+	fi
```
jodh-intel, line 45:
```suggestion
	[ ! -d "/run/image" ] && mkdir /run/image
```

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+device_path="/dev/$device_name"
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+	dd if=/dev/urandom of=/run/cc_storage.key bs=1 count=4096
+
+	if [ "$enable_data_integrity" == "false" ]; then
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
+			--cipher aes-xts-plain64 /run/cc_storage.key
+	else
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
+			 --cipher aes-xts-plain64 --integrity hmac-sha256 /run/cc_storage.key
```
jodh-intel, line 36: Please can you use more variables to avoid all the duplication (`/run/cc_storage.key`, _etc_).

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
...
+device_name=$(cat /sys/dev/block/${device_num}/uevent|grep DEVNAME|cut -d "=" -f2)
+device_path="/dev/$device_name"
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+	dd if=/dev/urandom of=/run/cc_storage.key bs=1 count=4096
+
+	if [ "$enable_data_integrity" == "false" ]; then
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
+			--cipher aes-xts-plain64 /run/cc_storage.key
+	else
+		echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
```
jodh-intel, line 35: What if the `cryptsetup` command isn't available (in the potentially custom image)? The script should probably loop over the commands it needs to run and do something like:

```bash
die()
{
    local msg="$*"

    echo >&2 "ERROR: $msg"

    exit 1
}

setup()
{
    local cmds=()

    cmds+=("cryptsetup")

    local cmd
    for cmd in "${cmds[@]}"
    do
        command -v "$cmd" &>/dev/null || die "need command: '$cmd'"
    done
}
```

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -692,6 +693,10 @@ EOF
 		popd
 	fi
 
+	if [ "${KATA_BUILD_CC}" == "yes" ]; then
+		install -o root -g root -m 0755 init_trusted_storage.sh "${ROOTFS_DIR}/usr/local/bin/"
```
jodh-intel, line 697: - Random thought: what about a context (`--context=`) if the image supports SELinux?
- Can't the perms be `0750` (or even `0500`)?

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+#
```
jodh-intel, line 2: Some random thoughts:

- If we were _really_ paranoid, this script could check to ensure it was being called by the agent process and `die()` (see below) if not.
- This script isn't idempotent. That _shouldn't_ be a problem, but with my paranoia hat on again it might be best if it is!

jodh-intel (commented):
Thanks @arronwy - a few comments...

arronwy:
> A nit, @arronwy, in the commit message you mention: "Since data integrity protection will have huge performance drop, we will enable it by default but introduce an config to allow the user to disable it if they have performance concerns."
> 
> It's worth rewriting it to make sure we will **NOT** enable it by default.

Yes, depends on the banlance between performance and more strict security, we can disable it by default.

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
arronwy: Great inputs, thanks a lot!

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
arronwy: During setup will check the availability of required commands.

arronwy:
> Do we have actual data on integrity performance drop? Found things [like that](https://unix.stackexchange.com/questions/591540/cryptsetup-with-dm-integrity-weird-benchmark-results), but it looks like this really depends on the exact cipher options chosen (another reason for making _that_ configurable).

Yes, I only test the initialize time which is three times slower, the read write performance is from: https://archive.fosdem.org/2018/schedule/event/cryptsetup/

I tried with `cryptsetup benchmark` , current cipher `aes-xts` have best performance

tools/osbuilder/rootfs-builder/init_trusted_storage.sh
arronwy: Yes, current is for init empheral storage, when we start support persistent storage, we will need get key from the KBS.

src/agent/src/image_rpc.rs
```diff
@@ -206,6 +208,36 @@ impl ImageService {
...
+                )
+                .await?;
+            }
+            if !pause_binary.exists() {
+                tokio::fs::copy(cc_pause_bundle.join("rootfs").join("pause"), pause_binary).await?;
+            }
+
+            let mut sandbox = self.sandbox.lock().await;
+            sandbox.images.insert(String::from(image), cid.to_string());
+            return Ok(image.to_owned());
```
sameo, line 238: Could we have this implemented into a separate `ImageService.unpack_pause_image()` method?

src/agent/src/image_rpc.rs
```diff
@@ -206,6 +208,36 @@ impl ImageService {
...
+        }
+
         if !aa_kbc_params.is_empty() {
```
sameo, line 241: I suppose it's intentional to not start the AA when we unpack the pause image?

src/agent/src/image_rpc.rs
arronwy: Yes, implement as a seperate `unpack_pause_image` will be much clean, will fix.

src/agent/src/image_rpc.rs
arronwy: Yes, pull `pause` image don't need start AA and `PullImageRequest` will pull one image each time, AA will be started when pull other images if needed.

huoqifeng (approved):
LGTM

sameo (approved):

fidencio (approved):
lgtm, thanks @arronwy!

dubek:
Hi @arronwy , I know this is already merged, but I have a question:

We generate a random key and then use it for luksFormat and luksOpen. However, since cryptsetup is usually used with keyboard-entered passphrases, it runs PBKDF to consume CPU+memory to prevent brute-force attempts or dictionary scans. This is not needed because the key we generate is long enough and has high entropy.

[This post](https://blastrock.github.io/fde-tpm-sb.html) suggests adding

```
--pbkdf-force-iterations=4 --pbkdf-parallel=1  --pbkdf-memory=32
```

to the cryptsetup command-line (I guess both for luksFormat and luksOpen) to reduce the time for these operations.

Ideally we could measure duration of these operations before and after such suggested change.

WDYT?

PR #4892, opened 2022-08-12
runtime-rs: support loading kernel modules in guest vm

YushuoEdge:
Users can specify the kernel module to be loaded through the agent
configuration in kata configuration file or in pod anotation file.

And information of those modules will be sent to kata agent when
sandbox is created.

Fixes: #4894 

### DIFF
--- a/src/libs/kata-types/src/config/agent.rs
+++ b/src/libs/kata-types/src/config/agent.rs
@@ -9,8 +9,10 @@ use crate::config::{ConfigOps, TomlConfig};
 
 pub use vendor::AgentVendor;
 
+use super::default::{DEFAULT_AGENT_LOG_PORT, DEFAULT_AGENT_VSOCK_PORT};
+
 /// Kata agent configuration information.
-#[derive(Debug, Default, Deserialize, Serialize)]
+#[derive(Debug, Default, Deserialize, Serialize, Clone)]
 pub struct Agent {
     /// If enabled, the agent will log additional debug messages to the system log.
     #[serde(default, rename = "enable_debug")]
@@ -34,11 +36,11 @@ pub struct Agent {
     pub debug_console_enabled: bool,
 
     /// Agent server port
-    #[serde(default)]
+    #[serde(default = "default_server_port")]
     pub server_port: u32,
 
     /// Agent log port
-    #[serde(default)]
+    #[serde(default = "default_log_port")]
     pub log_port: u32,
 
     /// Agent connection dialing timeout value in millisecond
@@ -75,23 +77,31 @@ pub struct Agent {
     pub container_pipe_size: u32,
 }
 
+fn default_server_port() -> u32 {
+    DEFAULT_AGENT_VSOCK_PORT
+}
+
+fn default_log_port() -> u32 {
+    DEFAULT_AGENT_LOG_PORT
+}
+
 fn default_dial_timeout() -> u32 {
-    // 10ms
+    // ms
     10
 }
 
 fn default_reconnect_timeout() -> u32 {
-    // 3s
+    // ms
     3_000
 }
 
 fn default_request_timeout() -> u32 {
-    // 30s
+    // ms
     30_000
 }
 
 fn default_health_check_timeout() -> u32 {
-    // 90s
+    // ms
     90_000
 }
 
--- a/src/libs/kata-types/src/config/default.rs
+++ b/src/libs/kata-types/src/config/default.rs
@@ -16,6 +16,8 @@ lazy_static! {
     ];
 }
 pub const DEFAULT_AGENT_NAME: &str = "kata-agent";
+pub const DEFAULT_AGENT_VSOCK_PORT: u32 = 1024;
+pub const DEFAULT_AGENT_LOG_PORT: u32 = 1025;
 
 pub const DEFAULT_INTERNETWORKING_MODEL: &str = "tcfilter";
 
--- a/src/runtime-rs/crates/agent/src/types.rs
+++ b/src/runtime-rs/crates/agent/src/types.rs
@@ -4,6 +4,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use anyhow::{anyhow, Result};
+use std::convert::TryFrom;
+
 use serde::Deserialize;
 
 #[derive(PartialEq, Clone, Default)]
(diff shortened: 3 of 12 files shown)


src/libs/kata-types/src/config/agent.rs
```diff
@@ -75,6 +75,14 @@ pub struct Agent {
     pub container_pipe_size: u32,
 }
 
+fn default_server_port() -> u32 {
+    1024
+}
+
+fn default_log_port() -> u32 {
+    1025
```
studychao, line 83: Why this is the default value?

studychao:
Could you add some UTs for this PR?

src/libs/kata-types/src/config/agent.rs
YushuoEdge: That is how it was written, even code included in the go version. By convention, the kata agent server is connected in port 1024.  And we add a log port at 1025.

YushuoEdge:
> Could you add some UTs for this PR?

Ok, I will add some UTs later.

src/libs/kata-types/src/config/agent.rs
jodh-intel: Can we define these values once then? For example `default_server_port()` and the agent should both uses a common `const` (aka move [`VSOCK_PORT`](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/config.rs#L31) to somewhere under `src/libs/`).

src/libs/kata-types/src/config/agent.rs
```diff
@@ -75,6 +75,14 @@ pub struct Agent {
...
+}
+
 fn default_dial_timeout() -> u32 {
     // 10ms
```
jodh-intel, line 87: We don't need to comment the value, just the units:

```suggestion
    // ms
```

jodh-intel (commented):
Thanks @YushuoEdge  - a couple of comments. Also, please add some unit tests to support this feature.

src/libs/kata-types/src/config/agent.rs
YushuoEdge: Thanks for your suggestions!

src/libs/kata-types/src/config/default.rs
```diff
@@ -16,6 +16,8 @@ lazy_static! {
     ];
 }
 pub const DEFAULT_AGENT_NAME: &str = "kata-agent";
+pub const DEFAULT_AGENT_VSOCK_PORT: u32 = 1024;
```
YushuoEdge, line 19: `vsock_port` and `log_port` are defined here(`/src/libs/kata-types/src/config/default`).  Is it ok to do this? @jodh-intel

src/runtime-rs/crates/agent/src/types.rs
```diff
@@ -390,6 +390,32 @@ pub struct KernelModule {
     pub parameters: Vec<String>,
 }
 
+impl KernelModule {
+    // input string: "ModuleName Param1 Param2 ..."
+    pub fn new(str: String) -> Self {
+        let split: Vec<&str> = str.split(' ').collect();
```
jodh-intel, line 396: Although the current module handling doesn't support it, it's worth pointing out that this won't work for all modules. From https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/kernel-parameters.rst:

> Double-quotes can be used to protect spaces in values, e.g.:
>
> `param="spaces in here"`

src/runtime-rs/crates/agent/src/types.rs
```diff
@@ -390,6 +390,32 @@ pub struct KernelModule {
     pub parameters: Vec<String>,
 }
 
+impl KernelModule {
+    // input string: "ModuleName Param1 Param2 ..."
+    pub fn new(str: String) -> Self {
```
jodh-intel, line 395: What if `str == ""`? Is that a problem? If so, maybe we could `impl TryFrom<...> for KernelModule`?

jodh-intel (commented):
Thanks @YushuoEdge. A few comments...

src/runtime-rs/crates/agent/src/types.rs
YushuoEdge: Thanks for reminding. This case is handled in set_kernel_modules.

src/runtime-rs/crates/agent/src/types.rs
YushuoEdge: Do we need to add this part of the implementation to the existing module handling as well? 🤔

jodh-intel (approved):
Thanks @YushuoEdge.

lgtm

src/runtime-rs/crates/agent/src/types.rs
YushuoEdge: Yes, this is indeed a corner case.  I can think of a scenario where this would be required. For example, If we implement a module with a message to be printed, and this message is passed as parameter, then we should consider the "space". 🙃

src/runtime-rs/crates/agent/src/types.rs
```diff
@@ -390,6 +393,57 @@ pub struct KernelModule {
...
+        let split: Vec<&str> = str.split(' ').collect();
+        let mut name = String::new();
+        let mut parameters = Vec::new();
+
+        let mut flag = false;
+        for (index, info) in split.iter().enumerate() {
+            if index == 0 {
+                name = info.to_string();
+            } else {
+                match flag {
```
liubin, line 424: `flag` is a `bool` type, why not just `if flag` but use a `match`?

src/runtime-rs/crates/agent/src/types.rs
```diff
@@ -486,3 +540,42 @@ pub struct VersionCheckResponse {
...
+        let kernel_module_str1 = "ModuleName Param1 Param2";
+        let kernel_module1 = KernelModule::try_from(kernel_module_str1.to_string()).unwrap();
+        assert!(kernel_module1.name == "ModuleName");
+        assert!(kernel_module1.parameters[0] == "Param1");
+        assert!(kernel_module1.parameters[1] == "Param2");
+
+        let kernel_module_str2 = "ModuleName Param1=\"spaces in here\"";
+        let kernel_module2 = KernelModule::try_from(kernel_module_str2.to_string()).unwrap();
+        assert!(kernel_module2.name == "ModuleName");
+        assert!(kernel_module2.parameters[0] == "Param1=\"spaces in here\"");
```
liubin, line 561: Add a test case for exceptions.

For example, input string is `mod1 "aaab c`, the result should be an `Err`.

liubin:
@YushuoEdge unfortunately this PR has a conflicts issue, you should do a rebase btw.

src/runtime-rs/crates/agent/src/types.rs
YushuoEdge: Thanks for reminding! This pr is update.

liubin (approved):

liubin:
Vendor check failed, please ensure the `Cargo.lock` has been committed.

```
all modules verified
./hack/tree_status.sh
tree is dirty, please commit all changes

 M src/agent/Cargo.lock
index 987f16ffa..de5606107 100644
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -625,6 +625,7 @@ dependencies = [
  "futures",
  "ipnetwork",
  "kata-sys-util",
+ "kata-types",
  "lazy_static",
make[1]: *** [Makefile:672: vendor] Error 1
  "libc",
  "log",
make[1]: Leaving directory '/home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime'
make: *** [Makefile:30: vendor-runtime] Error 2
```

studychao:
retest

Issue #4896, opened 2022-08-12
runk: seccomp support is not enabled by default


ManaSugi:
# Description of problem

Currently, seccomp support is not enabled in `runk` by default because the compilation does not enable `seccomp` feature in `rustjail`.
Let's enable it by default.

# Expected result

If we use `libseccomp` as dynamic linking.

```bash
$ ldd target/x86_64-unknown-linux-musl/release/runk
        linux-vdso.so.1 (0x00007ffd41fb0000)
        libseccomp.so.2 => /usr/local/lib/libseccomp.so.2 (0x00007f2255524000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2255332000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2255adc000)
```

# Actual result

```bash
$  ldd target/x86_64-unknown-linux-musl/release/runk
        statically linked
```

PR #4897, opened 2022-08-12
runk: Enable seccomp support by default

ManaSugi:
Enable seccomp support in `runk` by default.
The user can disable seccomp by `make SECCOMP=no`

Fixes: #4896

### DIFF
--- a/src/tools/runk/Cargo.lock
+++ b/src/tools/runk/Cargo.lock
@@ -581,6 +581,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "libseccomp"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "49bda1fbf25c42ac8942ff7df1eb6172a3bc36299e84be0dba8c888a7db68c80"
+dependencies = [
+ "libc",
+ "libseccomp-sys",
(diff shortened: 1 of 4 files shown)


bergwolf (approved):
Thanks!

liubin (approved):

ManaSugi:
@bergwolf @liubin 
Sorry, I updated the commit.
From this commit, `runk` is built with `glibc` by default because the building `runk` with statically linked the libseccomp library against `musl` requires additional configurations manually for the [libseccomp crate](https://github.com/libseccomp-rs/libseccomp-rs). (Previously, `runk` has been built with `musl` by default like `kata-agent`)
In addition, general container runtimes are built as dynamically linked binaries by default, so I think it would be better to build `runk` with `glibc`. Of course, the user can build `runk` with `musl` by setting `LIBC=musl`.
Could you review this again? Thank you.

liubin (approved):

ManaSugi:
@bergwolf 
Could you review this again? Thank you.
From this commit, `runk` is built with `gnu libc` by default.

liubin:
@bergwolf @lifupan @Tim-Zhang  PATL.

@ManaSugi there is a git conflict that needs to resolve.

PR #4903, opened 2022-08-12
runtime: tracing: End root span at end of trace

### LINKED ISSUE #4902: Root span should exist for the duration of the trace
The root span in the tracing output should, but does not, exist the duration of the trace. Since spans are derived from a root context saved in the service struct it does not affect the rest of the spans. It can be fixed by deferring the ending of the span to the end of the trace instead of ending it in the function it was created.

cmaf:
The root span should exist the duration of the trace. Defer ending span
until the end of the trace instead of end of function. Add the span to
the service struct to do so.

Fixes #4902

The output of the rest of the trace is not affected since the rest of the spans are ultimately derived from a context saved in the service struct, but this is technically a bug and should be backported.

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/create.go
+++ b/src/runtime/pkg/containerd-shim-v2/create.go
@@ -97,9 +97,10 @@ func create(ctx context.Context, s *service, r *taskAPI.CreateTaskRequest) (*con
 		}
 
 		// create root span
+		// rootSpan will be ended when the entire trace is ended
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @cmaf!

cmaf:
cc @jodh-intel

jodh-intel (approved):
Thanks @cmaf!

lgtm

PR #4909, opened 2022-08-15
runtime: surpport hugepage_size with clh

liangxianlong:
clh support hugepage，but it use default hugepage_size from system only.Provides a configuration item and annotation to specify the huge page type.

Fixes: #4792

Signed-off-by: liangxianlong <liang.xianlong@zte.com.cn>


amshinde:
@liangxianlong Why is this required, clh should dynamically detect the configured huge page size on the system.

PR #4917, opened 2022-08-15
agent/wip: Replace String with OsString or Path

### LINKED ISSUE #1378: agent should use OsString or std::fs::Path in many places it uses String

bookinabox:
To cover the edge case where paths may not be UTF-8 encoded, paths
should be type OsString or Path rather than String.

This replaces most instances of String with Path or OsString.

Fixes #1378

### DIFF
--- a/src/agent/rustjail/src/cgroups/fs/mod.rs
+++ b/src/agent/rustjail/src/cgroups/fs/mod.rs
@@ -34,7 +34,7 @@ use protocols::agent::{
 };
 use std::collections::HashMap;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 const GUEST_CPUS_PATH: &str = "/sys/devices/system/cpu/online";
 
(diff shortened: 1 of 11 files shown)


bookinabox:
I've been slowly replacing instances I can find. I am worried I am breaking parts of the code I don't understand because of my inexperience with rust and the codebase. I have been running the cargo test unittests, but I'm not sure if that is enough.

There are some structs in oci.rs that are auto-generated by proto3, which as far I know, cannot have the regenerate with anything but String.

Any input would be appreciated!

bookinabox:
I may close this one. I was recommended it and started it thinking it would be a simple find and replace, but I think without a good understanding of the structure of the code, I may have introduced more complexities than it may be worth. I still have not figured out a way of replacing instances of string that are auto-generated.

Issue #4920, opened 2022-08-16
Delete build dependencies vergen


liubin:
[`vergen`](https://docs.rs/vergen/latest/vergen/) is a build dependency, but it is not being used.

PR #4921, opened 2022-08-16
runtime-rs: delete vergen dependency

liubin:
vergen is a build dependency, but it is not being used.
we are processing ver/commit hash by make command, but not by vergen.

Deleting this dependency can save 40+ crates and 1 minute to build.

Fixes: #4920


studychao (approved):
Thanks, Bin:)

Tim-0731-Hzt (approved):

jodh-intel (approved):
Good catch @liubin !

lgtm

cmaf (approved):

Issue #4926, opened 2022-08-16
CC | TDVF is wrongly configured when in the QEMU TDX configuration file


fidencio:
TDVF can be configured in one of the following ways:
```
firmware = "/path/to/OVMF.fd"
```
Or
```
firmware = "/path/to/OVMF_CODE.fd"
firmware_volume = "/path/to/OVMF_VARS.fd"
```

But, mistakenly, we're currently doing:
```
firmware = "/path/to/OVMF.fd"
firmware_volume = "/path/to/OVMF_VARS.fd"
```

And this mistake leads to the container not starting, as shown below:
```
ctr: failed to create shim task: failed to launch qemu: exit status 1, error messages from qemu log: qemu-system-x86_64-tdx: TDVF metadata doesn't specify TD_HOB location. : unknown
```

PR #4927, opened 2022-08-16
CC | runtime: Fix TDVF configuration with QEMU TDX

fidencio:
Instead of setting:
```
firmware = "/path/to/OVMF.fd"
firmware_volume = "/path/to/OVMF_VARS.fd"
```

We should either be setting:
```
firmware = "/path/to/OVMF.fd"
```
Or:

```
firmware = "/path/to/OVMF_CODE.fd"
firmware_volume = "/path/to/OVMF_VARS.fd"
```

I'm taking the approach to setting up the latter, as that's what's been
tested as part of our TDX CI.

Fixes: #4926


stevenhorsman (approved):
LGTM. Thanks!

PR #4933, opened 2022-08-16
Upgrade to Cloud Hypervisor v26.0

likebreath:
Highlights from the Cloud Hypervisor release v26.0:

**SMBIOS Improvements via `--platform`**
`--platform` and the appropriate API structure has gained support for supplying
OEM strings (primarily used to communicate metadata to systemd in the guest)

**Unified Binary MSHV and KVM Support**
Support for both the MSHV and KVM hypervisors can be compiled into the same
binary with the detection of the hypervisor to use made at runtime.

**Notable Bug Fixes**
* The prefetchable flag is preserved on BARs for VFIO devices
* PCI Express capabilties for functionality we do not support are now filtered
out
* GDB breakpoint support is more reliable
* SIGINT and SIGTERM signals are now handled before the VM has booted
* Multiple API event loop handling bug fixes
* Incorrect assumptions in virtio queue numbering were addressed, allowing
thevirtio-fs driver in OVMF to be used
* VHDX file format header fix
* The same VFIO device cannot be added twice
* SMBIOS tables were being incorrectly generated

**Deprecations**
Deprecated features will be removed in a subsequent release and users should
plan to use alternatives.

The top-level `kernel` and `initramfs` members on the `VmConfig` have been
moved inside a `PayloadConfig` as the `payload` member. The OpenAPI document
has been updated to reflect the change and the old API members continue to
function and are mapped to the new version. The expectation is that these old
versions will be removed in the v28.0 release.

**Removals**
The following functionality has been removed:

The unused poll_queue parameter has been removed from --disk and
equivalent. This was residual from the removal of the vhost-user-block
spawning feature.

Details can be found: https://github.com/cloud-hypervisor/cloud-hypervisor/releases/tag/v26.0

Fixes: https://github.com/kata-containers/kata-containers/issues/4952


src/runtime/virtcontainers/clh.go
```diff
@@ -459,13 +459,15 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
 	// to fetch if this is the first time the hypervisor is created.
 	clh.Logger().WithField("function", "CreateVM").Info("Sandbox not found creating")
 
+	// Create the VM config via the constaructor to ensure default values are properly assigned
```
fidencio, line 462:
```suggestion
	// Create the VM config via the constructor to ensure default values are properly assigned
```

versions.yaml
```diff
@@ -75,7 +75,7 @@ assets:
       url: "https://github.com/cloud-hypervisor/cloud-hypervisor"
       uscan-url: >-
         https://github.com/cloud-hypervisor/cloud-hypervisor/tags.*/v?(\d\S+)\.tar\.gz
-      version: "v25.0"
+      version: "8f3462b3e356f56fee33ae4ba833a1fef012c1f0"
```
fidencio, line 78: You're tagging the commit as the release was not yet made, right?

fidencio (commented):
A few comments for when you remove the draft status.

versions.yaml
likebreath: Right. It was a dry-run to make sure is working.

fidencio (approved):
lgtm, thanks @likebreath !

GabyCT (approved):

Issue #4939, opened 2022-08-17
ci: Update libseccomp version in CI


ManaSugi:
# Description of problem

A few days ago, #4858 is merged, but the libseccomp version specified in `versions.yaml` is used in only [Release page](https://github.com/kata-containers/kata-containers/releases),

The actual version statically linked with the agent is the version specified in [this script](https://github.com/kata-containers/kata-containers/blob/main/ci/install_libseccomp.sh).

PR #4940, opened 2022-08-17
ci: Update libseccomp version

ManaSugi:
Updates the libseccomp version that is being used in the Kata CI.

Fixes: #4858, #4939


ManaSugi:
We need to modify the libseccomp notice section manually in the release page about 3.0.0-alpha0 because the page mentions the libseccomp version as 2.5.4.
In addition, currently, we specify the libseccomp version directly in the script  without using `versions.yaml`
because the current Snap workflow is incomplete. I'll investigate it to solve the issue.

fidencio (approved):
lgtm, thanks @ManaSugi!

liubin (approved):

PR #4945, opened 2022-08-17
Revert "kata-deploy: Rely on the configured config path"

fidencio:
This reverts commit f7ccf92dc82e9264fafd0ccf565f6dd014786e3e.

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -98,7 +98,10 @@ function configure_different_shims_base() {
 			fi
 		fi
 
-		ln -sf /opt/kata/bin/containerd-shim-kata-v2 "${shim_file}"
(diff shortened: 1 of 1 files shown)


egernst:
Happy if it works in your testing, but I'm curious to understand what changed / why this is needed.

fidencio:
> Happy if it works in your testing, but I'm curious to understand what changed / why this is needed.

Right now we're not even sure if this is actually needed., but this is the only change done recently that could cause the issues faced by @bergwolf when cutting the 3.0.0-alpha0 release.

It also makes me think that, **maybe**, the script used to by kata-deploy to do the needed changes in the host may not actually be used as part of the CI, but this needs proper investigation.

fidencio:
I'm adding a `do-not-merge` label for now till we find some time to properly debug the issue.

fidencio:
Closing this one as kata-deploy does work on my local environment without this change.

PR #4957, opened 2022-08-19
virtcontainers: Fix ServiceOffload usage with devmapper

fidencio:
When using device mapper we shouldn't return early, as we still need the
block device mounted on the guest.

Fixes: #0000


BbolroC:
@fidencio there should be a space between the ending double quote and the opening brace?

```
fs_share_linux.go:378: File is not `gofmt`-ed with `-s` (gofmt)
	if f.sandbox.config.ServiceOffload && c.rootFs.Target == "" && c.state.BlockDeviceID == ""{
```

Issue #4966, opened 2022-08-23
CC | image-builder: generate rootfs hash data for verified/measured boot


arronwy:
As described in the measured boot image proposal for CoCo https://github.com/confidential-containers/documentation/issues/40
We need generate rootfs hash data during creating the kata rootfs, current kata image only have one partition, we need add another partition as hash device to hold the hash data of rootfs data blocks.

PR #4967, opened 2022-08-23
CC | image-build: generate root hash as an separate partition for rootfs

arronwy:
One task for issue: https://github.com/confidential-containers/documentation/issues/40

Generate rootfs hash data during creating the kata rootfs,
current kata image only have one partition, we add another
partition as hash device to save hash data of rootfs data blocks.

Fixes: #4966


fidencio (approved):
lgtm, thanks @arronwy!

amshinde (approved):

Issue #4981, opened 2022-08-24
qemu: Drop unnecessary `tdx_guest` kernel parameter


fidencio:
With the current TDX kernel used with Kata Containers, `tdx_guest` is
not needed, as TDX_GUEST is now a kernel configuration.

With this in mind, let's just drop the kernel parameter.

Issue #4980, opened 2022-08-24
clh: Use HVC console with TDX


fidencio:
As right now the TDX guest kernel doesn't support "serial" console,
let's switch to using HVC in this case.

Issue #4979, opened 2022-08-24
clh: Avoid crashing when memory hotplug is not allowed


fidencio:
The runtime will crash when trying to resize memory when memory hotplug
is not allowed.

This happens because we cannot simply set the hotplug amount to zero,
leading is to not set memory hotplug at all, and later then trying to
access the value of a nil pointer.

Issue #4978, opened 2022-08-24
clh: Increase API and SandboxStop timeouts for TDX


fidencio:
While doing tests using `ctr`, I've noticed that I've been hitting those
timeouts more frequently than expected.

Till we find the root cause of the issue (which is *not* in the Kata
Containers), let's increase the timeouts when dealing with a
Confidential Guest.

PR #4982, opened 2022-08-24
TDX: Get TDX working again with Cloud Hypervisor + a minor change on QEMU's code

fidencio:
clh: Lift the sharedFS restriction used with TDX

When booting the TDX kernel with `tdx_disable_filter`, as it's been done
for QEMU, VirtioFS can work without any issues.

Whether this will be part of the upstream kernel or not is a different
story, but it easily could make it there as Cloud Hypervisor relies on
the VIRTIO_F_IOMMU_PLATFORM feature, which forces the guest to use the
DMA API, making these devices compatible with TDX.

See Sebastien Boeuf's explanation of this in the
3c973fa7ce208e7113f69424b7574b83f584885d commit:
"""
By using DMA API, the guest triggers the TDX codepath to share some of
the guest memory, in particular the virtqueues and associated buffers so
that the VMM and vhost-user backends/processes can access this memory.
"""

Fixes: #4977

---

clh: Increase API and SandboxStop timeouts for TDX

While doing tests using `ctr`, I've noticed that I've been hitting those
timeouts more frequently than expected.

Till we find the root cause of the issue (which is *not* in the Kata
Containers), let's increase the timeouts when dealing with a
Confidential Guest.

Fixes: #4978

---

clh: Avoid crashing when memory hotplug is not allowed

The runtime will crash when trying to resize memory when memory hotplug
is not allowed.

This happens because we cannot simply set the hotplug amount to zero,
leading is to not set memory hotplug at all, and later then trying to
access the value of a nil pointer.

Fixes: #4979

---

clh: Use HVC console with TDX

As right now the TDX guest kernel doesn't support "serial" console,
let's switch to using HVC in this case.

Fixes: #4980

---

qemu: Drop unnecessary `tdx_guest` kernel parameter

With the current TDX kernel used with Kata Containers, `tdx_guest` is
not needed, as TDX_GUEST is now a kernel configuration.

With this in mind, let's just drop the kernel parameter.

Fixes: #4981

### DIFF
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -210,7 +210,6 @@ func (q *qemuAmd64) enableProtection() error {
 			q.qemuMachine.Options += ","
 		}
 		q.qemuMachine.Options += "kvm-type=tdx,confidential-guest-support=tdx"
-		q.kernelParams = append(q.kernelParams, Param{"tdx_guest", ""})
 		logger.Info("Enabling TDX guest protection")
 		return nil
 	case sevProtection:
--- a/src/runtime/config/configuration-clh.toml.in
+++ b/src/runtime/config/configuration-clh.toml.in
@@ -28,10 +28,6 @@ image = "@IMAGEPATH@"
 #   - CPU Hotplug 
 #   - Memory Hotplug
 #   - NVDIMM devices
-#   - SharedFS, such as virtio-fs and virtio-fs-nydus
-#
-# Requirements:
-# * virtio-block used as rootfs, thus the usage of devmapper snapshotter.
 #
 # Supported TEEs:
 # * Intel TDX
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -877,7 +877,13 @@ func (clh *cloudHypervisor) ResizeMemory(ctx context.Context, reqMemMB uint32, m
 		return 0, MemoryDevice{}, err
 	}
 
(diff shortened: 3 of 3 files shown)


sboeuf (approved):

fidencio:
Force pushed to fix the `gofmt` issues.

src/runtime/virtcontainers/qemu_amd64.go
```diff
@@ -210,7 +210,6 @@ func (q *qemuAmd64) enableProtection() error {
 			q.qemuMachine.Options += ","
 		}
 		q.qemuMachine.Options += "kvm-type=tdx,confidential-guest-support=tdx"
-		q.kernelParams = append(q.kernelParams, Param{"tdx_guest", ""})
```
wainersm, line 213: is this `TDX_GUEST` kernel configuration already upstream? More precisely, any risk of that change in near future so that we will need to revert this change?

likebreath (approved):

src/runtime/virtcontainers/qemu_amd64.go
fidencio: AFAIR it's not upstreamed yet, but the `tdx_guest` parameter was from an older review.  So, if it changes, hardly it'll change back to that format. :-)

fidencio:
> Force pushed to fix the `gofmt` issues.

Second round of `gofmt` fixes.

fidencio:
Power CI has been failing in different PRs and it's very unlikely that the issue is related to this PR.
/cc @Amulyam24

Issue #4367, opened 2022-06-01
runk: support init sub-command


liubin:
add `runk init` support

PR #4983, opened 2022-08-24
runk: Add cli message for init command

ManaSugi:
Add cli message for init command to tell the user
not to run this command directly.

Fixes: #4367

### DIFF
--- a/src/tools/runk/src/main.rs
+++ b/src/tools/runk/src/main.rs
@@ -28,6 +28,7 @@ enum SubCommand {
     Common(CommonCmd),
     #[clap(flatten)]
     Custom(CustomCmd),
+    /// Launch an init process (do not call it outside of runk)
     Init {},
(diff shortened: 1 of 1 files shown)


ManaSugi:
This message is printed like below:

```bash
$ runk --help | grep init
    init           Launch an init process (do not call it outside of runk)
```

fidencio (approved):
lgtm, thanks @ManaSugi!

bergwolf:
@ManaSugi Did you forget to add some new files in the commit?

ManaSugi:
@bergwolf 
This commit is just adding the help message for the existing `init` command that is printed in CLI, so there are no additional new files.

```bash
$ runk --help | grep init
    init           Launch an init process (do not call it outside of runk)
```

amshinde (approved):

PR #4984, opened 2022-08-24
CC | Build and ship CLH TDX (and all the needed artefacts)

fidencio:
Let's add a new configuration for using a TDX capable CLH with Kata Containers, including all the artefacts needed, and the changes that are needed as part of the Makefile and deployment scripts.


fengwang666 (approved):
LGTM

PR #5004, opened 2022-08-25
Updating Reduced Physical Bits

larrydewey:
This updates the ReducedPhysBits to use the value of 1, as `cpuid` provides information which is relevant to the host, and does not match what a guest would expect.


fidencio:
@larrydewey, something went awry when you opened the PR.
I'd recommend you rebase your work atop of the current `CCv0` branch and force-push it here, so we have a clean diff, of only one single patch.

larrydewey:
Accidentally branched off of main. Closing for another PR.

PR #5005, opened 2022-08-25
Update ReducedPhysBit

larrydewey:
Updating this field, as `cpuid` provides host level data, which is not
what a guest would expect for Reduced Phsycial Bits. In almost all
cases, we should be using `1` for the value here.

Signed-off-by: Larry Dewey <larry.dewey@amd.com>


fidencio:
@larrydewey, is there any documentation that you could refer to about this change?  Just thinking here on our future selves trying to figure where the `1` value came from. :-)

larrydewey:
Good question, @fidencio. Here are a few reference points:

* [AMD64 Architecture Programmer’s Manual, Volume 2: System Programming - 7.10.5 - I/O Accesses](https://www.amd.com/system/files/TechDocs/24593.pdf#page=290) - Explains the how that reduced physical address spaces are expected, and mentions the c-bit in particular. As of right now, this should be the only reduced bit the guest cares about.
* [Guest Hypervisor Communication Block (GHCB) Standardization - 4.1.5 - #NPF/MMIO Access](https://developer.amd.com/wp-content/resources/56421.pdf#page=43) - Explains the page fault behavior and expected calculation of offset.
* [Qemu Upstream](https://github.com/qemu/qemu/blob/master/target/i386/sev.c#L611) - Explanation of how qemu expects the `reduced_phys_bits` to be used.

Let me know if additional sources would be beneficial.

fidencio:
Adding the `do-not-merge` label till we get the PR for the `main` branch merged.
This was privately discussed with @larrydewey!

fidencio:
@larrydewey, the first link is worth adding to the commit message.

Issue #5006, opened 2022-08-25
Fix ReducedPhysBits field


larrydewey:
# Description of problem

There is an issue with the ReducedPhysBits field pulling information from `cpuid`, which is not appropriate for what a guest would expect. On an SEV machine, this value should almost always be `1`. This will hard-code this value to prevent unexpected behavior.

# Actual Result
Use correct reduced physcial bits to prevent unexpected behaviors within the vm space.

Fix Pending PR

PR #5007, opened 2022-08-25
SEV: Update ReducedPhysBits

larrydewey:
Updating this field, as `cpuid` provides host level data, which is not
what a guest would expect for Reduced Phsycial Bits. In almost all
cases, we should be using `1` for the value here.

Fixes: #5006

Signed-off-by: Larry Dewey <larry.dewey@amd.com>

### DIFF
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -290,7 +290,7 @@ func (q *qemuAmd64) appendProtectionDevice(devices []govmmQemu.Device, firmware,
 				Debug:           false,
 				File:            firmware,
 				CBitPos:         cpuid.AMDMemEncrypt.CBitPosition,
-				ReducedPhysBits: cpuid.AMDMemEncrypt.PhysAddrReduction,
+				ReducedPhysBits: 1,
 			}), "", nil
 	case snpProtection:
 		return append(devices,
--- a/src/runtime/virtcontainers/qemu_amd64_test.go
+++ b/src/runtime/virtcontainers/qemu_amd64_test.go
@@ -286,7 +286,7 @@ func TestQemuAmd64AppendProtectionDevice(t *testing.T) {
 			Debug:           false,
 			File:            firmware,
 			CBitPos:         cpuid.AMDMemEncrypt.CBitPosition,
-			ReducedPhysBits: cpuid.AMDMemEncrypt.PhysAddrReduction,
+			ReducedPhysBits: 1,
 		},
 	}
 


fidencio:
Adding the same comment here.

@larrydewey, is there any documentation that you could refer to about this change? Just thinking here on our future selves trying to figure where the 1 value came from. :-)

fidencio:
Also, how fast do you need this one backported to the `CCv0` branch?  If you can wait till the rebase happens (every Wednesday) you don't need https://github.com/kata-containers/kata-containers/pull/5005

larrydewey:
> Adding the same comment here.
> 
> @larrydewey, is there any documentation that you could refer to about this change? Just thinking here on our future selves trying to figure where the 1 value came from. :-)

Good question, @fidencio. Here are a few reference points:

[AMD64 Architecture Programmer’s Manual, Volume 2: System Programming - 7.10.5 - I/O Accesses](https://www.amd.com/system/files/TechDocs/24593.pdf#page=290) - Explains the how that reduced physical address spaces are expected, and mentions the c-bit in particular. As of right now, this should be the only reduced bit the guest cares about.
[Guest Hypervisor Communication Block (GHCB) Standardization - 4.1.5 - #NPF/MMIO Access](https://developer.amd.com/wp-content/resources/56421.pdf#page=43) - Explains the page fault behavior and expected calculation of offset.
[Qemu Upstream](https://github.com/qemu/qemu/blob/master/target/i386/sev.c#L611) - Explanation of how qemu expects the reduced_phys_bits to be used.
Let me know if additional sources would be beneficial.

> > Also, how fast do you need this one backported to the `CCv0` branch? If you can wait till the rebase happens (every Wednesday) you don't need #5005

Nothing is critically broken (that we can tell), but we want to prevent potential issues. Between the two branches the file seems significantly different. I am happy to wait for the backport on Wednesday, though.

I know we spoke on Slack, but just for history.

fitzthum (approved):
Nice. Original behavior was a bug imo.

fidencio:
lgtm, thanks @larrydewey!

c3d:
@larrydewey I am not entirely. The `govmmQemu.Object` we are building leads to a `qemu` command-line that looks like:     `-object sev-guest,id=sev0,cbitpos=47,reduced-phys-bits=1` (see [qemu documentation](https://www.qemu.org/docs/master/system/confidential-guest-support.html)). This is what qemu needs to properly allocate _host_ memory, so it is correct to pass host information. If a particular implementation of SEV loses 5 bits of physical address space when SEV is enabled, then `qemu` has to assume that the guest physical addresses are reduced by 5 bits, not 1.

The documentation you pointed to indicates that the C bit may be in physical addresses, but that may not be the only reason for stealing guest physical address. Another reason could be to leave room for firmware or some VMPL0 stuff (e.g. you could steal one bit just to make sure that guest physical addresses using that bit are invalid). The [qemu source code](https://github.com/qemu/qemu/blob/master/qapi/misc-target.json#L152) shows an example where running SEV, for whatever reason, steals 5 bits:

```
 <- { "return": { "pdh": "8CCDD8DDD", "cert-chain": "888CCCDDDEE",
                  "cpu0-id": "2lvmGwo+...61iEinw==",
                  "cbitpos": 47, "reduced-phys-bits": 5}}

```

So for the moment, I'm going to suggest we delay this until you explain to me why you think we should pass a different value to the host `qemu` process.

c3d (changes requested):
For now, blocking this since I believe this is incorrect. I think that we should indeed pass the host value to `qemu`, since that is what `qemu` needs to properly allocate guest physical memory.

c3d:
@larrydewey Also, I'm curious if this led to an actual issue? if there was a test that you think presumably failed because of this, we need that info in #5006 .

c3d:
> [AMD APM Vol 2](https://www.amd.com/system/files/TechDocs/24593.pdf) section 15.34.6 Page Table Support says:
> 
> > Note that because guest physical addresses are always translated through the nested page tables, the
> > size of the guest physical address space is not impacted by any physical address space reduction
> > indicated in CPUID 8000_001F[EBX].

IMO, that only means that the translation process itself can target host physical addresses that are not reduced (i.e. the nested page table can address the entire host physical address space. The examples they give make that relatively clear, I think. I do not interpret that as meaning that the guest physical address space reduction info can be messed with.

> If the C-bit is a physical address bit however, the guest
> > physical address space is effectively reduced by 1 bit

That reinforces what I wrote above: the C bit takes space in the guest-visible PTEs, so that's where the impact is.

> 
> So it looks like this PR is fine (hard-coding to 1).
> 
> Some more data points:
> 
> * QEMU always returns reduced-phys-bits=1 when responding to `query-sev-capabilities` QMP command (see [source](https://gitlab.com/qemu-project/qemu/-/blob/master/target/i386/sev.c#L614)). I guess we should fix QEMU's documentation because that `"reduced-phys-bits":5` response is impossible.

Let me ask in the qemu community. I think that it _currently_ always returns 1, but it _architecturally_ could return other values.

> * QEMU doesn't use the value of `reduced-phys-bits` (except for checking that it's >= 1).

That is incorrect. It is used [in the emulation of the `cpuid` instruction ](https://github.com/qemu/qemu/blob/277ad7d58611b455662f2e3f7bd24ce5bfeb2fdc/target/i386/cpu.c#L5851), which makes it possible to pass the data to the guest.

For a KVM SEV guest, this is then used to generate the [physical addresses bitmask](https://github.com/torvalds/linux/blob/master/arch/x86/kvm/svm/svm.c#L4193), although I must admit that it's possible that the value passed by qemu is explicitly [zeroed by the kernel](https://github.com/torvalds/linux/blob/277ad7d58611b455662f2e3f7bd24ce5bfeb2fdc/arch/x86/kvm/cpuid.c#L1172) (see the `GENMASK` macro), which might make the value we pass irrelevant today.

Irrespective of whether the kernel actually uses it today, I'm pretty sure that we should not lie to qemu or to the guest kernel about it. They can decide to ignore it, we should not.

One question I have, though, is why we need to pass the value at all. I don't see why qemu is not using the host value itself, instead of relying on a command-line option.

dubek:
> That is incorrect. It is used [in the emulation of the `cpuid` instruction ](https://github.com/qemu/qemu/blob/277ad7d58611b455662f2e3f7bd24ce5bfeb2fdc/target/i386/cpu.c#L5851), which makes it possible to pass the data to the guest.

You're right, I missed that.

> One question I have, though, is why we need to pass the value at all. I don't see why qemu is not using the host value itself, instead of relying on a command-line option.

I remember something related to QEMU machine definition and migration. For example, you have to define the property `cbitpos=47` (even though QEMU can get this value from the host), and this way it'll prevent (or fail) migration of the guest to another host in which `cbitpos != 47`.

For `cbitpos` the guest's value must be equal to the host's value (in current SEV generations); currently QEMU enforces this.

For `reduced-phys-bits`, I think this is not the case (host value can be 5 and guest value can be 1).

tlendacky:
> For `cbitpos` the guest's value must be equal to the host's value (in current SEV generations); currently QEMU enforces this.
> 
> For `reduced-phys-bits`, I think this is not the case (host value can be 5 and guest value can be 1).

That is correct. As specified in the APM, if the C-bit is a physical address bit, then the guest physical address space is reduced by 1 and therefore the reduced-phys-bits is 1 and never more than that. The guest reduced physical addressing reduction is not the same as the host physical addressing reduction.

larrydewey:
> c3d:
>
> \@larrydewey I am not entirely. The govmmQemu.Object we are building leads to a qemu command-line that looks like: -object sev-guest,id=sev0,cbitpos=47,reduced-phys-bits=1 (see [qemu documentation](https://www.qemu.org/docs/master/system/confidential-guest-support.html)). This is what qemu needs to properly allocate host memory, so it is correct to pass host information. If a particular implementation of SEV loses 5 bits of physical address space when SEV is enabled, then qemu has to assume that the guest physical addresses are reduced by 5 bits, not 1.
>
> The documentation you pointed to indicates that the C bit may be in physical addresses, but that may not be the only reason for stealing guest physical address. Another reason could be to leave room for firmware or some VMPL0 stuff (e.g. you could steal one bit just to make sure that guest physical addresses using that bit are invalid). The [qemu source code](https://github.com/qemu/qemu/blob/master/qapi/misc-target.json#L152) shows an example where running SEV, for whatever reason, steals 5 bits:

@c3d you are correct in stating there are other reasons one might reduce the physical bit space. Due to the x86_64 architecture, the traditional 48 bit address (bits 0-47) is mapped by the processor. The host does play a roll in configuring that position, this tells qemu that the most significant bit (47) will be representing the c-bit, and thus it needs to be reduced by one (as the c-bit should not be considered part of a valid address). However, from my understanding, the examples of firmware or VMPLs are not implemented using `reducedPhysBits` (@tlendacky you may correct me here if I am wrong), but are tracked through a different means. You also cited the documentation in qemu, which from everything I have been able to find is also incorrect, and should be changed to more accurately describe the qemu object in question.

> tlendacky:
>
> That is correct. As specified in the APM, if the C-bit is a physical address bit, then the guest physical address space is reduced by 1 and therefore the reduced-phys-bits is 1 and never more than that. The guest reduced physical addressing reduction is not the same as the host physical addressing reduction.

Like @tlendacky mentioned above, the guest reduced physical addressing will not be the same as the host reduced physical addressing. This provides the answer as to why qemu is not pulling the value directly from the host.

This PR is looking to rectify the previously incorrect behavior, and should be unblocked and merged.

c3d:
Responding after being ping'ed on Slack. FWIW, I have only received one reply to my email asking about this to the relevant Red Hat experts, and that one answer appears to concur with my interpretation (including the fact that qemu should probably just read the CPUID instead of relying on a command-line option)

> @c3d you are correct in stating there are other reasons one might reduce the physical bit space, but this is not one of these examples are not the case. Due to the x86_64 architecture, the traditional 48 bit address (bits 0-47) is mapped by the processor. The host does play a roll in configuring that position, this tells qemu that the most significant bit (47) will be representing the c-bit, and thus it needs to be reduced by one (as the c-bit should not be considered part of a valid address).

I understand that. And we correctly pass this information today, don't we? We get it from  `cpuid.AMDMemEncrypt.PhysAddrReduction`, which is designed precisely to tell us about this. You did not answer my earlier question about why you believe that we should not trust that value? Have you observed cases where you think the value given there was wrong?

I suspect the root cause of our different interpretation is that you interpret that CPUID `reducedPhysBits` value as being to be read _from the guest_, and I read section 15.34.1 as indicating that it is to be read _from the host_ to tell you _about the SEV guests_.

Let me quote 15.34.1 (emphasis mine):

_When memory encryption features are present, CPUID 8000_001F[EBX] and 8000_001F[ECX] supply additional information regarding the use of memory encryption, such as the number of keys supported simultaneously and which page table bit is used to mark pages as encrypted. **Additionally, in some implementations, the physical address size of the processor may be reduced when memory encryption features are enabled, for example from 48 to 43 bits. In this example, physical address bits 47:43 would be treated as reserved except where otherwise indicated**. When memory encryption is supported in an implementation, CPUID 8000_001F[EBX] reports any physical address size reduction present. Bits reserved in this mode are treated the same as other page table reserved bits, and will generate a page fault if found to be non-zero when used for address translation._

Given the wording above, I do not see how it would be correct to use any other value than what we get from CPUID. Hard-coding 1 there definitely seems wrong, as it ignores the very reason for that CPUID value to even exist. Getting that wrong would lead to guest crashes, with "mysterious" page faults whenever the guest attempts to use these bits.

>However, from my understanding, the examples of firmware or VMPLs are not implemented using `reducedPhysBits` (@tlendacky you may correct me here if I am wrong), but are tracked through different means.

This is not what I was saying. My point was that even on systems where the hardware CPUID reduced phys bits value is 1, some VMPL0 wrapper might need to reserve part of the address space, and a way to do that would be by stealing some bits from the address space. For example, they might reserve any page table entry with bit 46 set to VMPL0 "kernel" code, and tell the guest about this by making more bits reserved. There is no better way to tell the VMPL1 software about such an address space reduction that I am aware of. I'm not saying that such a scenario exists today, not even that it is likely, but that this is an example of why even if all the systems we have today were returning 1 in that CPUID field, it would _still_ be a good idea to take the CPUID value into account.

But that point is moot, since the documentation, as quoted above, makes it quite clear that some implementations might reserve bits when running confidential VMs for any other reason. I do not know if such implementations exist today, or if they are on the drawing board. It does not matter either way: the `reducedPhysBits` value is there for a reason.

> You also cited the documentation in qemu, which from everything I have been able to find is also incorrect, and should be changed to more accurately describe the qemu object in question.

I am sorry, but you are stating that qemu is incorrect without having answered my earlier questions about what makes you believe the CPUID value is not the right source for that information, or why qemu is incorrect. In other words, I feel like you are just asserting that without having given me a good reason for such an assertion.

So let me rephrase my earlier question: have you observed incorrect values there either on some hardware, or with some particular firmware? If so, could you share that information? If not, I stick by my earlier interpretation of what the documentation states, which is also what qemu implements. If you have observed bad values, then we would need a workaround to override a bad `reducedPhysBits` value, but we should treat it as a workaround, not as the default behaviour. In that case, I would recommend an annotation like `overrideReducedPhysBits` that someone in that situation could use to override the value from CPUID.

> 
> > tlendacky:
> > That is correct. As specified in the APM, if the C-bit is a physical address bit, then the guest physical address space is reduced by 1 and therefore the reduced-phys-bits is 1 and never more than that. The guest reduced physical addressing reduction is not the same as the host physical addressing reduction.
> 
> Like @tlendacky mentioned above, the guest reduced physical addressing will not be the same as the host reduced physical addressing. This provides the answer as to why qemu is not pulling the value directly from the host.
> 
> This PR is looking to rectify the previously incorrect behavior, and should be unblocked and merged.

Again, as long as you do not provide any rationale why you think that the CPUID `reducedPhysBit` value is incorrect, I do not understand why I should change my analysis.

fitzthum:
Well I am hesitant to get sucked into this, but the way I read the manual is that CPUID gives us the reduced physical bits of the host address space. Enabling confidential computing might reduce the host address space, which is maybe a bit counterintuitive because confidential computing seems more like a guest feature. Nonetheless, I don't think the guest address space changes (except that the guest will also have a cbit in addition to the host reductions) and it seems like the manual more or less spells it out that the guest value should be 1 in almost every situation.

There is currently a bit of an inconsistency in QEMU. The example return value of the SevCapability @c3d first pointed out has five as the `reduced-phys-bits`, but as @dubek points out, QEMU hard codes this value to one. So the documentation is incorrect in the sense that it does not reflect what QEMU does. When QEMU reports the `reduced-phys-bits` via the SEV capability, it will always return one. QEMU also takes `reduced-phys-bits` as a parameter. QEMU checks that the value is greater than zero. QEMU does not get `reduced-phys-bits` from CPUID. On most Milan machines CPUID reports 5 as the host reduced physical bits value. The manual seems to make it clear in section 15.34.6 (see @dubek's response above) that reduced physical bits should be 1.

Keep in mind that this PR is to fix the Kata shim. QEMU does not use CPUID to set `reduced-phys-bits`, although I think they might have in the past, hence the mistaken documentation. I don't think the kata shim should either. Now we haven't actually identified this as the cause of any problems, but I still think it is a mistake in the kata shim. This PR does not contradict the QEMU implementation, in fact it brings us more in line with it.

jpecholt:
I originally wrote to @larrydewey because I was confused what value was to be used for the `reduced-phys-bits` field and what its impact was exactly. I had seen that `1` was the default in the majority of cases (AMD's guides, most of QEMU...) but noticed that Kata-Containers was using the host's physical address space reduction, namely `5`. Looking a bit further into the documentation, I couldn't quite figure out what was the correct value and what impact it had on what component exactly.
@fitzthum summed it up pretty well, that was the latest understanding I had myself.

The reason why this was an issue for me: I had varying success depending on the hardware I was on, though I am blaming this on outdated firmware and software components. I have since updated all components wherever newer releases where available and no longer have these issues.

*(Tiny in-depth analysis below that can be skipped for this discussion)*

 Generally, I observed the following for my most up-to-date setup:

SEV-VMs allow a range from `reduced-phys-bits=1-(2^32-1)`
SNP-VMs allow a range from `reduced-phys-bits=1-319`

Inside the guest, the value is always the same as the specified value up until `63`, after that it stays cropped at `63`, probably because of some bit mask for the 6-bit field somewhere. I concluded that it is not used inside the guest for now since it has no apparent impact. The field is set in the current [QEMU code](https://github.com/AMDESE/qemu/blob/snp-v3/target/i386/cpu.c#L5981) as follows:

```bash
    case 0x8000001F:
        *eax = sev_enabled() ? 0x2 : 0;
        *eax |= sev_es_enabled() ? 0x8 : 0;
        *eax |= sev_snp_enabled() ? 0x10 : 0;
        *ebx = sev_get_cbit_position();
        *ebx |= sev_get_reduced_phys_bits() << 6;
        *ecx = 0;
        *edx = 0;
        break;
```

This is the CPUID definition in [AMD's manual Vol. 3, section E.4.17](https://www.amd.com/system/files/TechDocs/24594.pdf) :

> ####  CPUID Fn8000_001F_EBX Secure Encryption
> |Bits |Field Name Description                            |
> |-----|--------------------------------------------------|
> |31:16| — Reserved.                                      |
> |15:12| NumVMPL Number of VM Permission Levels supported |
> |11:6 | PhysAddrReduction Physical Address bit reduction |
> |5:0  | CbitPosition C-bit location in page table        |

To me, it looks like the SNP-specific QEMU version should probably be cleaned up to 
1) contain up-to-date information on the field and its impact 
2) not use both hard-coded values and user-specified values or at least explicitly state why this is done in each case (if the value is always `1` for now, why even provide a field to modify it?)
3) have some sort of (unified) verification of the input value since right now I can start an SEV-VM with `reduced-phys-bits=4294967295` but "only" boot SNP-VMS with up to `reduced-phys-bits=319`, (which might have to do with 4.) and 
4) probably use a bit mask when setting the `ebx` field for the reduced bits in the above code since right now this overwrites all higher fields if I am not mistaken which could cause issues if these fields are used

This is getting very specific to QEMU though, so I'm keeping it at that.

**tl;dr**:
Value `reduced-phys-bits=1` should be correct for the guest and the latest QEMU version for SNP should probably be updated to reflect this in both code and documentation? I am happy to stand corrected, though!

tlendacky:
> @larrydewey Responding after being ping'ed on Slack. Sorry for the delay, KVM Forum + waiting for more replies to my email. FWIW, I have only received one reply to my email asking about this to the relevant Red Hat experts, and that one answer appears to concur with my interpretation (including the fact that qemu should probably just read the CPUID instead of relying on a command-line option)

The CPUID information that Qemu reads in the HOST will be the HOST physical address space reduction, which is NOT the same as the guest physical address reduction.

> 
> > @c3d you are correct in stating there are other reasons one might reduce the physical bit space, but this is not one of these examples are not the case. Due to the x86_64 architecture, the traditional 48 bit address (bits 0-47) is mapped by the processor. The host does play a roll in configuring that position, this tells qemu that the most significant bit (47) will be representing the c-bit, and thus it needs to be reduced by one (as the c-bit should not be considered part of a valid address).
> 
> I understand that. And we correctly pass this information today, don't we? We get it from `cpuid.AMDMemEncrypt.PhysAddrReduction`, which is designed precisely to tell us about this. You did not answer my earlier question about why you believe that we should not trust that value? Have you observed cases where you think the value given there was wrong?

Again, no, cpuid.AMDMemEncrypt.PhysAddrReduction in the HOST is the HOST physical address space reduction as a result of enabling the hardware memory encryption features, which does not equal the GUEST address space reduction. The GUEST address space reduction needs to be programmed into the CPUID information presented to the guest - which should be 1 as documented in the last paragraph of section 15.34.6 of APM Vol 2.

This physical address space reduction happens for bare-metal, regardless of whether an SEV guest will ever be run. This allows SME (HOST memory encryption) to be used in the HOST.

> 
> I suspect the root cause of our different interpretation is that you interpret that CPUID `reducedPhysBits` value as being to be read _from the guest_, and I read section 15.34.1 as indicating that it is to be read _from the host_ to tell you _about the SEV guests_.

This information, when read from the HOST, tells the effects that enabling memory encryption features has on the HOST, not the GUEST. If the GUEST will encounter a physical address space reduction, it should be presented with a virtualized CPUID value that the GUEST will then read.

Please see section 15.34.6 of APM Vol 2.

> 
> Let me quote 15.34.1 (emphasis mine):
> 
> _When memory encryption features are present, CPUID 8000_001F[EBX] and 8000_001F[ECX] supply additional information regarding the use of memory encryption, such as the number of keys supported simultaneously and which page table bit is used to mark pages as encrypted. **Additionally, in some implementations, the physical address size of the processor may be reduced when memory encryption features are enabled, for example from 48 to 43 bits. In this example, physical address bits 47:43 would be treated as reserved except where otherwise indicated**. When memory encryption is supported in an implementation, CPUID 8000_001F[EBX] reports any physical address size reduction present. Bits reserved in this mode are treated the same as other page table reserved bits, and will generate a page fault if found to be non-zero when used for address translation._

The sentences that you added emphasis to are HOST statements.

> 
> Given the wording above, I do not see how it would be correct to use any other value than what we get from CPUID. Hard-coding 1 there definitely seems wrong, as it ignores the very reason for that CPUID value to even exist. Getting that wrong would lead to guest crashes, with "mysterious" page faults whenever the guest attempts to use these bits.

Again, the CPUID value on the HOST is how the memory encryption support effects the HOST.

The reduced physical address reduction then needs to be supplied to the guest through CPUID presented to the GUEST only if the last paragraph of 15.34.6 of APM Vol 2 applies and then the value would be exactly 1.

I believe rather than checking the HOST full physical addressing range against the GUEST physical address range, it was just decided to always specify 1. And the specification of this on the Qemu command line was also a safe guard for live migration.

> 
> > However, from my understanding, the examples of firmware or VMPLs are not implemented using `reducedPhysBits` (@tlendacky you may correct me here if I am wrong), but are tracked through different means.
> 
> This is not what I was saying. My point was that even on systems where the hardware CPUID reduced phys bits value is 1, some VMPL0 wrapper might need to reserve part of the address space, and a way to do that would be by stealing some bits from the address space. For example, they might reserve any page table entry with bit 46 set to VMPL0 "kernel" code, and tell the guest about this by making more bits reserved. There is no better way to tell the VMPL1 software about such an address space reduction that I am aware of. I'm not saying that such a scenario exists today, not even that it is likely, but that this is an example of why even if all the systems we have today were returning 1 in that CPUID field, it would _still_ be a good idea to take the CPUID value into account.

That isn't currently how it is done. The guest is told how much memory it has available (and KVM will fail access to a PFN outside of that range). The SVSM will protect its memory through VMPL permissions, so even if the guest tried to access memory in the SVSM memslot (which it doesn't really know about, but could try to access) it's access would fail based on VMPL permissions.

It may be possible that we could further update the CPUID information that the SVSM presents to the GUEST and reduce what the GUEST thinks it has available, but that would be through CPUID 0x80000008_EAX, primarily.

> 
> But that point is moot, since the documentation, as quoted above, makes it quite clear that some implementations might reserve bits when running confidential VMs for any other reason. I do not know if such implementations exist today, or if they are on the drawing board. It does not matter either way: the `reducedPhysBits` value is there for a reason.
> 
> > You also cited the documentation in qemu, which from everything I have been able to find is also incorrect, and should be changed to more accurately describe the qemu object in question.
> 
> I am sorry, but you are stating that qemu is incorrect without having answered my earlier questions about what makes you believe the CPUID value is not the right source for that information, or why qemu is incorrect. In other words, I feel like you are just asserting that without having given me a good reason for such an assertion.
> 
> So let me rephrase my earlier question: have you observed incorrect values there either on some hardware, or with some particular firmware? If so, could you share that information? If not, I stick by my earlier interpretation of what the documentation states, which is also what qemu implements. If you have observed bad values, then we would need a workaround to override a bad `reducedPhysBits` value, but we should treat it as a workaround, not as the default behaviour. In that case, I would recommend an annotation like `overrideReducedPhysBits` that someone in that situation could use to override the value from CPUID.
> 
> > > tlendacky:
> > > That is correct. As specified in the APM, if the C-bit is a physical address bit, then the guest physical address space is reduced by 1 and therefore the reduced-phys-bits is 1 and never more than that. The guest reduced physical addressing reduction is not the same as the host physical addressing reduction.
> > 
> > 
> > Like @tlendacky mentioned above, the guest reduced physical addressing will not be the same as the host reduced physical addressing. This provides the answer as to why qemu is not pulling the value directly from the host.
> > This PR is looking to rectify the previously incorrect behavior, and should be unblocked and merged.
> 
> Again, as long as you do not provide any rationale why you think that the CPUID `reducedPhysBit` value is incorrect, I do not understand why I should change my analysis.

The last paragraph of 15.36.6 clearly indicates that a guest loses at most 1 bit of physical addressing. That is lost if the encryption bit is part of the guest physical address space.

So, for example, if Qemu only provides the guest with 40-bits of physical addressing (through CPUID 0x80000008) and the encryption bit is bit 47, then the guest actually does not lose any of the 40-bits of physical addressing available to it. However, if Qemu was to provide the guest with 48-bits of physical addressing with the encryption bit being bit 47, then the guest would lose 1 bit of physical addressing.

Thanks,
Tom

c3d:
@tlendacky Thank you for your response. I think you seem to imply that a guest can use PTE / TLB bits that are marked reserved in the host. I was not aware this was allowed, I thought this would fault.

tlendacky:
> @tlendacky Thank you for your response. I think you seem to imply that a guest can use PTE / TLB bits that are marked reserved in the host. I was not aware this was allowed, I thought this would fault.

That's correct, the PTE bits that are now reserved in the host as a result of the reduction in physical addressing after enabling memory encryption on the platform (identified by CPUID 0x8000001F_EBX[11:6]) are available to be used by the guest.

tlendacky:
On 9/20/22 08:37, Christophe de Dinechin wrote:
> Responding after being ping'ed on Slack. FWIW, I have only received one reply to my email asking about this to the relevant Red Hat experts, and that one answer appears to concur with my interpretation (including the fact that qemu should probably just read the CPUID instead of relying on a command-line option)

The CPUID information that Qemu reads in the HOST will be the HOST 
physical address space reduction, which is NOT the same as the guest 
physical address reduction.

> 
>> @c3d you are correct in stating there are other reasons one might reduce the physical bit space, but this is not one of these examples are not the case. Due to the x86_64 architecture, the traditional 48 bit address (bits 0-47) is mapped by the processor. The host does play a roll in configuring that position, this tells qemu that the most significant bit (47) will be representing the c-bit, and thus it needs to be reduced by one (as the c-bit should not be considered part of a valid address).
> 
> I understand that. And we correctly pass this information today, don't we? We get it from  `cpuid.AMDMemEncrypt.PhysAddrReduction`, which is designed precisely to tell us about this. You did not answer my earlier question about why you believe that we should not trust that value? Have you observed cases where you think the value given there was wrong?

Again, no, cpuid.AMDMemEncrypt.PhysAddrReduction in the HOST is the HOST 
physical address space reduction as a result of enabling the hardware 
memory encryption features, which does not equal the GUEST address space 
reduction. The GUEST address space reduction needs to be programmed into 
the CPUID information presented to the guest - which should be 1 as 
documented in the last paragraph of section 15.34.6 of APM Vol 2.

This physical address space reduction happens for bare-metal, regardless 
of whether an SEV guest will ever be run. This allows SME (HOST memory 
encryption) to be used in the HOST.

> 
> I suspect the root cause of our different interpretation is that you interpret that CPUID `reducedPhysBits` value as being to be read _from the guest_, and I read section 15.34.1 as indicating that it is to be read _from the host_ to tell you _about the SEV guests_.

This information, when read from the HOST, tells the effects that enabling 
memory encryption features has on the HOST, not the GUEST. If the GUEST 
will encounter a physical address space reduction, it should be presented 
with a virtualized CPUID value that the GUEST will then read.

Please see section 15.34.6 of APM Vol 2.

> 
> Let me quote 15.34.1 (emphasis mine):
> 
> _When memory encryption features are present, CPUID 8000_001F[EBX] and 8000_001F[ECX] supply additional information regarding the use of memory encryption, such as the number of keys supported simultaneously and which page table bit is used to mark pages as encrypted. **Additionally, in some implementations, the physical address size of the processor may be reduced when memory encryption features are enabled, for example from 48 to 43 bits. In this example, physical address bits 47:43 would be treated as reserved except where otherwise indicated**. When memory encryption is supported in an implementation, CPUID 8000_001F[EBX] reports any physical address size reduction present. Bits reserved in this mode are treated the same as other page table reserved bits, and will generate a page fault if found to be non-zero when used for address translation._

The sentences that you added emphasis to are HOST statements.

> 
> Given the wording above, I do not see how it would be correct to use any other value than what we get from CPUID. Hard-coding 1 there definitely seems wrong, as it ignores the very reason for that CPUID value to even exist. Getting that wrong would lead to guest crashes, with "mysterious" page faults whenever the guest attempts to use these bits.

Again, the CPUID value on the HOST is how the memory encryption support 
effects the HOST.

The reduced physical address reduction then needs to be supplied to the 
guest through CPUID presented to the GUEST only if the last paragraph of 
15.34.6 of APM Vol 2 applies and then the value would be exactly 1.

I believe rather than checking the HOST full physical addressing range 
against the GUEST physical address range, it was just decided to always 
specify 1. And the specification of this on the Qemu command line was also 
a safe guard for live migration.

> 
>> However, from my understanding, the examples of firmware or VMPLs are not implemented using `reducedPhysBits` ***@***.*** you may correct me here if I am wrong), but are tracked through different means.
> 
> This is not what I was saying. My point was that even on systems where the hardware CPUID reduced phys bits value is 1, some VMPL0 wrapper might need to reserve part of the address space, and a way to do that would be by stealing some bits from the address space. For example, they might reserve any page table entry with bit 46 set to VMPL0 "kernel" code, and tell the guest about this by making more bits reserved. There is no better way to tell the VMPL1 software about such an address space reduction that I am aware of. I'm not saying that such a scenario exists today, not even that it is likely, but that this is an example of why even if all the systems we have today were returning 1 in that CPUID field, it would _still_ be a good idea to take the CPUID value into account.

That isn't currently how it is done. The guest is told how much memory it 
has available (and KVM will fail access to a PFN outside of that range). 
The SVSM will protect its memory through VMPL permissions, so even if the 
guest tried to access memory in the SVSM memslot (which it doesn't really 
know about, but could try to access) it's access would fail based on VMPL 
permissions.

It may be possible that we could further update the CPUID information that 
the SVSM presents to the GUEST and reduce what the GUEST thinks it has 
available, but that would be through CPUID 0x80000008_EAX, primarily.

> 
> But that point is moot, since the documentation, as quoted above, makes it quite clear that some implementations might reserve bits when running confidential VMs for any other reason. I do not know if such implementations exist today, or if they are on the drawing board. It does not matter either way: the `reducedPhysBits` value is there for a reason.
> 
>> You also cited the documentation in qemu, which from everything I have been able to find is also incorrect, and should be changed to more accurately describe the qemu object in question.
> 
> I am sorry, but you are stating that qemu is incorrect without having answered my earlier questions about what makes you believe the CPUID value is not the right source for that information, or why qemu is incorrect. In other words, I feel like you are just asserting that without having given me a good reason for such an assertion.
> 
> So let me rephrase my earlier question: have you observed incorrect values there either on some hardware, or with some particular firmware? If so, could you share that information? If not, I stick by my earlier interpretation of what the documentation states, which is also what qemu implements. If you have observed bad values, then we would need a workaround to override a bad `reducedPhysBits` value, but we should treat it as a workaround, not as the default behaviour. In that case, I would recommend an annotation like `overrideReducedPhysBits` that someone in that situation could use to override the value from CPUID.
> 
>>
>>> tlendacky:
>>> That is correct. As specified in the APM, if the C-bit is a physical address bit, then the guest physical address space is reduced by 1 and therefore the reduced-phys-bits is 1 and never more than that. The guest reduced physical addressing reduction is not the same as the host physical addressing reduction.
>>
>> Like @tlendacky mentioned above, the guest reduced physical addressing will not be the same as the host reduced physical addressing. This provides the answer as to why qemu is not pulling the value directly from the host.
>>
>> This PR is looking to rectify the previously incorrect behavior, and should be unblocked and merged.
> 
> Again, as long as you do not provide any rationale why you think that the CPUID `reducedPhysBit` value is incorrect, I do not understand why I should change my analysis.

The last paragraph of 15.36.6 clearly indicates that a guest loses at most 
1 bit of physical addressing. That is lost if the encryption bit is part 
of the guest physical address space.

So, for example, if Qemu only provides the guest with 40-bits of physical 
addressing (through CPUID 0x80000008) and the encryption bit is bit 47, 
then the guest actually does not lose any of the 40-bits of physical 
addressing available to it. However, if Qemu was to provide the guest with 
48-bits of physical addressing with the encryption bit being bit 47, then 
the guest would lose 1 bit of physical addressing.

Thanks,
Tom

> 
> 
> 
> 
>

c3d (approved):
After re-reading the documentation, and notably the last paragraph of 15.34.6, it seems actually true that the guest can lose less bits than the host. Approving.

fitzthum:
We should merge this sometime.

larrydewey:
Re-basing to merge

larrydewey:
@kata-containers/kata-containers-admin I am pretty sure this is good to go, but has been stale for a REALLY long time. I fixed the re-base issue I was having. Any way we can get this merged?

amshinde (approved):

PR #5009, opened 2022-08-25
kata-check: Add cri-o/containerd version check

### LINKED ISSUE #2793: Add CRI implementation check for kata-runtime check
`kata-runtime check` run checks for KVM/Kernel modules.
Add check of CRI(CRI-O containerd)'s version checks sometimes may be helpful to troubleshooting.

bookinabox:
Adds checks to cri-o and containerd version in kata-check.go

As only the cri-o branch is specified, cri-o is only compared on
MAJOR.MINOR rather than MAJOR.MINOR.patch.

Fixes #2793

### DIFF
--- a/src/runtime/cmd/kata-runtime/kata-check.go
+++ b/src/runtime/cmd/kata-runtime/kata-check.go
@@ -32,6 +32,11 @@ import (
 	"github.com/urfave/cli"
 )
 
+var (
+	containerdRecVersion string
+	crioRecVersion       string
+)
+
 type kernelModule struct {
 	// maps parameter names to values
 	parameters map[string]string
@@ -440,8 +445,8 @@ EXAMPLES:
 
 			fmt.Println(successMessageCreate)
 		}
-
-		return nil
+		err = checkVersions()
+		return err
 	},
 }
 
@@ -537,4 +542,41 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
 	}
 
 	return results, nil
+}
+
+func checkVersions() error {
+	// Extract version of Crio
+	crioVerCmd := exec.Command("crio", "--version")
+	crioVerArr, err := crioVerCmd.Output()
+
+	if err != nil {
+		fmt.Printf("Unable to determine cri-o version (you may not be using cri-o)%s\n", err)
+	} else {
+		crioVer := string(crioVerArr)
+		versionStart, versionEnd := "Version:          ", "\nGoVersion:"
+		versionIndex, endIndex := strings.Index(crioVer, versionStart), strings.Index(crioVer, versionEnd)
+		crioVer = crioVer[versionIndex+len(versionStart) : endIndex]
+
+		crioRelease := strings.Join(strings.Split(crioVer, ".")[0:2], ".")
+
+		if crioRelease != crioRecVersion {
+			fmt.Printf("You are using cri-o %s, but %s is recommended\n", crioRelease, crioRecVersion)
+		}
+
+	}
+
+	// Extract version of containerd
+	containerdVerCmd := exec.Command("containerd", "--version")
+	containerdVerArr, err := containerdVerCmd.Output()
+	if err != nil {
+		fmt.Printf("Unable to determine containerd version (you may not be using containerd) %s\n", err)
+	} else {
+		containerdVer := "v" + strings.Split(string(containerdVerArr[:]), " ")[2]
+
+		if "v"+containerdVer != containerdRecVersion {
+			fmt.Printf("You are using containerd %s, but %s is recommended\n", containerdVer, containerdRecVersion)
+		}
+
+	}
+	return nil
 }
--- a/src/runtime/Makefile
+++ b/src/runtime/Makefile
@@ -228,6 +228,10 @@ MONITOR_DIR = $(CLI_DIR)/kata-monitor
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
 VERSION := ${shell cat ./VERSION}
 
+VERSIONSFILEPATH = "${CURDIR}/../../versions.yaml"
+CONTAINERDRECVERSION = $(shell yq r $(VERSIONSFILEPATH) 'externals.containerd.version')
+CRIORECVERSION = $(shell yq r $(VERSIONSFILEPATH) 'externals.crio.branch' | cut -d "-" -f2)
+
 # List of configuration files to build and install
 CONFIGS =
 CONFIG_PATHS =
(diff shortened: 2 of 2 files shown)


bookinabox:
@liubin, could you look at this? I'm not too familiar with Go libraries, so I'm sure there is a more efficient way to parse yaml files/do string/version manipulations than what I did.

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -87,6 +89,108 @@ var (
 	kvmDevice = "/dev/kvm"
 )
 
+// Generated from https://zhwt.github.io/yaml-to-go/
+type ThirdPartyVersions struct {
+	Externals struct {
```
liubin, line 94: Should only containerd and CRI-O be kept and remove others?

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -441,6 +545,12 @@ EXAMPLES:
 			fmt.Println(successMessageCreate)
 		}
 
+		err = checkVersions()
+
```
liubin, line 549: No blank line between `err = ` and `if err != nil`, also here you can only `return err` regardless of what the value of `err`.

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -538,3 +648,62 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
 
 	return results, nil
 }
+
+func getVersionsYaml() (ThirdPartyVersions, error) {
+
+	versionFile, err := ioutil.ReadFile("../../versions.yaml")
```
liubin, line 654: For me the body of this function can be 

```golang
	var data ThirdPartyVersions
	versionFile, err := ioutil.ReadFile("../../versions.yaml")
	if err != nil {
		return data, err
	}

	err = yaml.Unmarshal(versionFile, &data)

	return data, err
```

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -538,3 +648,62 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+		return data, err2
+	}
+
+	return data, nil
+}
+
+func checkVersions() (err error) {
+
+	versionYaml, err := getVersionsYaml()
+
```
liubin, line 676: No blank lines

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -538,3 +648,62 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+		return err
+	}
+
+	// Crio is specified by branch/release rather than version in versions.yaml
+	crioVerRec := strings.Join(strings.SplitN(versionYaml.Externals.Crio.Branch, "release-", 2), "")
+	containerdVerRec := versionYaml.Externals.Containerd.Version
+
+	// Extract version of Crio
+	crioVerCmd := exec.Command("crio", "--version")
+	crioVerArr, err := crioVerCmd.Output()
```
liubin, line 687: this `err` should be checked in case of that CRI-O is not installed.

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -538,3 +648,62 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+	crioVer := string(crioVerArr[:])
```
liubin, line 688: `crioVer := string(crioVerArr[:])` may be the same as `crioVer := string(crioVerArr)`

src/runtime/cmd/kata-runtime/kata-check.go
bookinabox: Makes sense

src/runtime/cmd/kata-runtime/kata-check.go
bookinabox: Thanks!

bookinabox:
Failing because I used a relative path. Will look for a way to avoid that

bookinabox:
I"m thinking of including the recommended version using ldflags during the Go build process instead of having it look for the versions.yaml file during runtime. This should save the headache of locating where the versions.yaml file is.

bookinabox:
@wainersm @liubin I changed it to use go build -ldflags to populate the containerd and crio versions. This pushes the responsibility of parsing versions.yaml into the Makefile instead of kata-checks.go, but let's us not care about finding where versions.yaml is during runtime.

bookinabox:
Did not realize I had to run go mod tidy to fix the go.mod file. Should work now, fingers crossed.

bookinabox:
@wainersm Sorry to keep bugging you about my failed static checks. Not used to the checks I should run for Go. There ended up being unused code from my old approach I did not remove that I did not catch that was caught by the linter.

bookinabox:
Failed due to `crio --version` not finding crio in $PATH. I guess I could just catch the error and print something else instead.

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -538,3 +543,40 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+
+		crioRelease := strings.Join(strings.Split(crioVer, ".")[0:2], ".")
+
+		if crioRelease != crioRecVersion {
+			fmt.Printf("You are using cri-o %s, but %s is recommended\n", crioRelease, crioRecVersion)
+		}
+
+	}
+
+	//Extract version of containerd
```
liubin, line 568: Add a space after `//`

src/runtime/cmd/kata-runtime/kata-check.go
```diff
@@ -538,3 +543,40 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
 	return results, nil
 }
+
+func checkVersions() error {
+	// Extract version of Crio
+	crioVerCmd := exec.Command("crio", "--version")
+	crioVerArr, err := crioVerCmd.Output()
+
+	if err != nil {
+		fmt.Printf("Unable to determine crio version %s\n", err)
```
liubin, line 553: How about adding some message like `maybe you are not using CRI-O` in the output to let the user know that this is not a critical error?

liubin:
> Failed due to `crio --version` not finding crio in $PATH. I guess I could just catch the error and print something instead of erroring out. I'm assuming its perfectly reasonable to be using only one, but to check for both and even have both installed.

Yes, CRI-O and containerd maybe only have one be installed, so printing the message and not returning an error is the right choice.

bookinabox:
Sorry for the delay. My internship is over and school is starting soon, so I may be slower with any responses.

Issue #5008, opened 2022-08-25
`kata-runtime kata-env` throws error when kernel built for SEV


ryansavino:
# Description of problem

`kata-runtime kata-env` throws an error when the kernel is built for SEV.

# Expected result

`kata-runtime kata-env` should display kata environment details.

# Actual result

`kata-runtime kata-env` returns an error:

```
jenkins@cc-node-ubuntu-20-04:~/ryansavino/tests$ kata-runtime kata-env
/usr/share/defaults/kata-containers/configuration.toml: file /usr/share/kata-containers/vmlinux.container does not exist
```

ryansavino:
When `build-kernel.sh install -x sev` is called, a suffix `-sev` is added to the installed files. I think the best solution here is to add a symlink for the basename file to the suffix named file. I will create a PR.

stevenhorsman:
I'm guessing this was long resolved, but SEV is removed now, so closing this

PR #5010, opened 2022-08-25
kernel: add basename kernel symlinks when suffix is used

ryansavino:
'kata-runtime kata-env' throws an error when the kernel is built for a specific TEE like SEV.
Basename kernel symlinks added to prevent error.

Fixes: #5008

### DIFF
--- a/tools/packaging/kernel/build-kernel.sh
+++ b/tools/packaging/kernel/build-kernel.sh
@@ -470,6 +470,13 @@ install_kata() {
 	ln -sf "${vmlinux}" "${install_path}/vmlinux${suffix}.container"
 	ls -la "${install_path}/vmlinux${suffix}.container"
 	ls -la "${install_path}/vmlinuz${suffix}.container"
+
+	if [ -n "$suffix}" ]; then
+		ln -sf "vmlinux${suffix}.container" "${install_path}/vmlinux.container"
+		ln -sf "vmlinuz${suffix}.container" "${install_path}/vmlinuz.container"
+		ls -la "${install_path}/vmlinux.container"
+		ls -la "${install_path}/vmlinuz.container"
+	fi
 	popd >>/dev/null
 }
 


fidencio (changes requested):
No, that's not the way to go.

We rather fix `kata-runtime kata-env`, otherwise we'll be simply overwriting the link to the non-TEE kernel, which is a valid use case for when we ship, for instance, normal Kata Containers and CC Kata Containers as part of the same payload.

fidencio:
Also, this is what I see as part of the TDX CI:
```
01:10:15 [Kernel]
01:10:15   Path = "/usr/share/kata-containers/vmlinuz-SPR-BKC-PC-v9.6-96-tdx"
01:10:15   Parameters = "systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket scsi_mod.scan=none tdx_disable_filter"
01:10:15 
```

What are you getting as part of the SEV CI?

fidencio:
> 

Just checked the issue.  The error you're facing should be fixed on the configuration that you're using so, there, you point to the vmlinux-sev.container (or similar ;-)).  That's what we do with TDX, and with any other kernel that's not the vanilla one.

ryansavino:
Thanks for the walkthrough @fidencio. I've created PR #5013 to address this. Closing this PR.

ryansavino:
> Also, this is what I see as part of the TDX CI:
> 
> ```
> 01:10:15 [Kernel]
> 01:10:15   Path = "/usr/share/kata-containers/vmlinuz-SPR-BKC-PC-v9.6-96-tdx"
> 01:10:15   Parameters = "systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket scsi_mod.scan=none tdx_disable_filter"
> 01:10:15 
> ```
> 
> What are you getting as part of the SEV CI?

I can give you an update on what the SEV CI flow shows once #5013 is merged.

PR #5014, opened 2022-08-26
agent : add  OFFLINE parameter

eive001:
Some times it is necessary to compile in isolation from the network, add the OFFLINE parameter, in OFFLINE=yes, the --offline parameter is added when using cargo build


amshinde:
@eive001 Please update the commit message following these guidelines:https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

PR #5024, opened 2022-08-29
CC | rootfs builder: add agent config file to rootfs for offline_sev_kbc

Alex-Carter01:
Adds default config file.
Adds case in rootfs.sh to copy config.

Fixes kata-containers#5023

Signed-Off-By: Alex Carter <alex.carter@ibm.com>

### DIFF
--- a/tools/osbuilder/rootfs-builder/agent-config.toml.in
+++ b/tools/osbuilder/rootfs-builder/agent-config.toml.in
@@ -1,0 +1,44 @@
+# Copyright (c) 2022 IBM Corp.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+aa_kbc_params = "$AA_KBC_PARAMS"
+
+[endpoints]
+allowed = [
+"AddARPNeighborsRequest",
+"AddSwapRequest",
+"CloseStdinRequest",
+"CopyFileRequest",
+"CreateContainerRequest",
+"CreateSandboxRequest",
+"DestroySandboxRequest",
+# "ExecProcessRequest",
+"GetMetricsRequest",
+"GetOOMEventRequest",
+"GuestDetailsRequest",
+"ListInterfacesRequest",
+"ListRoutesRequest",
+"MemHotplugByProbeRequest",
+"OnlineCPUMemRequest",
+"PauseContainerRequest",
+"PullImageRequest",
+"ReadStreamRequest",
+"RemoveContainerRequest",
+# "ReseedRandomDevRequest",
+"ResumeContainerRequest",
+"SetGuestDateTimeRequest",
+"SignalProcessRequest",
+"StartContainerRequest",
+"StartTracingRequest",
+"StatsContainerRequest",
+"StopTracingRequest",
+"TtyWinResizeRequest",
+"UpdateContainerRequest",
+"UpdateInterfaceRequest",
+"UpdateRoutesRequest",
+"WaitProcessRequest",
+"WriteStreamRequest"
+]
+
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -654,7 +654,10 @@ EOF
           UMOCI="yes"
           warning "UMOCI wasn't set, but is required for attestation, so overridden"
         fi
-
+		if [ "${AA_KBC}" == "offline_sev_kbc" ]; then
+			info "Adding agent config for ${AA_KBC}"
+			AA_KBC_PARAMS="offline_sev_kbc::null" envsubst < "${script_dir}/agent-config.toml.in" | tee "${ROOTFS_DIR}/etc/agent-config.toml"
+		fi
 		attestation_agent_url="$(get_package_version_from_kata_yaml externals.attestation-agent.url)"
 		attestation_agent_branch="$(get_package_version_from_kata_yaml externals.attestation-agent.branch)"
 		info "Install attestation-agent with KBC ${AA_KBC}"
--- a/tools/osbuilder/rootfs-builder/ubuntu/Dockerfile.in
+++ b/tools/osbuilder/rootfs-builder/ubuntu/Dockerfile.in
@@ -39,6 +39,7 @@ RUN apt-get update && \
     musl-tools \
     pkg-config \
     protobuf-compiler \
+    gettext-base \
     umoci
 
 # aarch64 requires this name -- link for all


tools/osbuilder/rootfs-builder/agent.cfg
```diff
@@ -0,0 +1,47 @@
+aa_kbc_params = "offline_sev_kbc::100.64.0.203:44444"
```
fitzthum, line 1: it might be best i the second parameter (after the `::`) was just `null` (as in the string `null` not actually null)

tools/osbuilder/rootfs-builder/agent.cfg
```diff
@@ -0,0 +1,47 @@
...
+debug_console = true
```
fitzthum, line 2: This should be `false`.

tools/osbuilder/rootfs-builder/agent.cfg
```diff
@@ -0,0 +1,47 @@
...
+[endpoints]
+allowed = [
+"AddARPNeighborsRequest",
+"AddSwapRequest",
+"CloseStdinRequest",
+"CopyFileRequest",
+"CreateContainerRequest",
+"CreateSandboxRequest",
+"DestroySandboxRequest",
+"ExecProcessRequest",
```
fitzthum, line 21: Might be good to comment this out.

fitzthum (approved):
A few minor comments but seems good.

Alex-Carter01:
Changes replace address with null, remove the debug console, and comment out some endpoints.

Alex-Carter01:
renaming to agent-config.toml for consistency with CI and the makefile-generated kata configs.

wainersm:
@Alex-Carter01 please remove `CC | ` of the commit subject to make the commit checker happy.

wainersm:
@fitzthum do we need the config file for CI sake only? or do we expect final users to rely on it?

Alex-Carter01:
Changed commit message title to hopefully appease commit message checker.

Also uncommented a couple of endpoints, my bad, I had a bit of a miscomm with Tobin about which endpoints should be allowed.

fitzthum:
> do we need the config file for CI sake only? or do we expect final users to rely on it?

The artifacts in the bundle will be the artifacts that users actually use. We don't really want to require a user to rebuild the initrd to add their configuration so I think that this configuration is going to be the "real" one that people use when they run in an SEV environment. 

There are some issues with this. For instance, there are some fields that a user might want to change. For the first release I think this is fine, but there are some open discussions about how to handle this best in the future (adding new config endpoints so we don't have to update the config). This is more of a problem when we use an online kbs where the KBS_URI parameter (highly dynamic) is set in the config file. For now we are using the `offline_sev_kbc` (for SEV, at least) so we don't need to worry about it too much. Generally, though, this is a limitation with the first release.

I do wonder if we should be more standardized about the config file across the project. It might be good to just have one default config that is pulled in wherever the initrd it built.

Alex-Carter01:
Changed agent config based on 

https://github.com/kata-containers/tests/pull/4977/files#diff-3c1f075cc76dbd146bc3a5a3271b90559d5290508ba74680d1043c2782df14b7 

removed some debug options

stevenhorsman:
@Alex-Carter01 - I think `osbuilder` is the usual subsystem for roofs changes.

tools/osbuilder/rootfs-builder/agent-config.toml
```diff
@@ -0,0 +1,44 @@
+# Copyright 2022 Advanced Micro Devices, Inc.
```
stevenhorsman, line 1: Could you either re-use, or refactor `docs/how-to/data/confidential-agent-config.toml.in` as it's pretty similar code?

Alex-Carter01:
changed subsystem to osbuilder

tools/osbuilder/rootfs-builder/agent-config.toml
fitzthum: I think we should try to standardize the config at some point, but I'm not sure the one in the docs is up-to-date anymore (although it's not too different). It makes me a bit nervous to start changing the config right before the release. I think we might have some unnecessary entries in the new one.

ryansavino: The agent config file should probably go under `kata-containers/tools/osbuilder/rootfs-builder`. This is where Alex has it currently. However, as Tobin mentioned, it's not standardized and it is tailored for the SEV use case:
`aa_kbc_params = "offline_sev_kbc::null"`

We could implement a variable and set it during the build. @stevenhorsman would that suffice?

 The docs config is being utilized by `.github/workflows/deploy-ccv0-demo.yaml`, and we can update it in this PR or wait for further documentation revisions.

We discussed the name of the agent config and decided to go with `/etc/agent-config.toml` as it's referenced in the tests repo:
`tests/integration/kubernetes/confidential/lib.sh`:    `local rootfs_agent_config="/etc/agent-config.toml"`

We could always update the name of the config in the tests repo if that is preferred.

@fitzthum Alex and I both had the same entries for the agent config in this PR and tests 4977. However, some may not be necessary. I'm pretty sure `docs/how-to/data/confidential-agent-config.toml.in` is out of date and missing required items. I could do some testing to determine what's actually required, but 

We should work on documentation updates before/during the freeze period. I can help.

stevenhorsman: Hey, thanks for the discussion. I think we want to consolidate both agent-config.toml for the reason that if we have multiple it will be harder to keep them all update and in sync (I say this in shame as I'm adding another one to the tests repo at the moment, but I'll try and reconcile this), but if you want this to become the new 'main' version then I can see that it makes sense to have it in the `rootfs-builder` and the name is fine `agent-config.toml.in` makes sense to me. In the short term passing in the aa_kbc_params via and environment variable similar to how we do in https://github.com/kata-containers/tests/blob/926c91af076af51690df067e0d821cbfed82a571/integration/kubernetes/confidential/lib.sh#L156 sounds sensible and we might need to slightly expand the scope to include things like the http proxy.
If you want to spin out a separate issue for trying to update the doc reference to move to this one that's totally fine too, I just wanted to try and avoid us duplicating and diverging too much.
PS: I hope this rambling make sense, otherwise give me a shout!

Alex-Carter01: switched to a model with an agent-config.toml.in and using envsubst to template the kbc name

Alex-Carter01:
Added kbc name templating, required envsubst dependency in docker environment

tools/osbuilder/rootfs-builder/agent-config.toml
stevenhorsman: Cool, thanks for this Alex. Post release we can have a proper discussion about what the officially 'sensitive' endpoints are and work to reconcile the agent config to remove duplication, but we are likely to be pretty close :)

wainersm (approved):

PR #5026, opened 2022-08-29
CC| packaging: Add AmdSev OVMF target for use in payload images

Alex-Carter01:
Currently leaving the cc-sev-ovmf-tarball target out of the cc payload.
I was not sure where discussion had landed on the number of payload bundles.
e.g. could be included in a cc bundle along with tdx support or create an SEV bundle.

Fixes: kata-containers#5025

### DIFF
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -88,7 +88,8 @@ cc: cc-cloud-hypervisor-tarball \
 	cc-tdx-kernel-tarball \
 	cc-tdx-qemu-tarball \
 	cc-tdx-td-shim-tarball \
-	cc-tdx-tdvf-tarball
+	cc-tdx-tdvf-tarball \
+	cc-sev-ovmf-tarball
 
 cc-cloud-hypervisor-tarball:
 	${MAKE} $@-build
@@ -121,4 +122,7 @@ cc-tdx-td-shim-tarball:
 	${MAKE} $@-build
 
 cc-tdx-tdvf-tarball:
+	${MAKE} $@-build
+
+cc-sev-ovmf-tarball:
 	${MAKE} $@-build
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -92,6 +92,7 @@ options:
 	cc-rootfs-image
 	cc-shimv2
 	cc-virtiofsd
+	cc-sev-ovmf
 EOF
 
 	exit "${return_code}"
@@ -205,6 +206,10 @@ install_cc_tee_ovmf() {
 
 install_cc_tdx_tdvf() {
 	install_cc_tee_ovmf "tdx" "edk2-staging-tdx.tar.gz"
+}
+
+install_cc_sev_ovmf(){
+ 	install_cc_tee_ovmf "sev" "edk2-sev.tar.gz"
 }
 
 #Install guest image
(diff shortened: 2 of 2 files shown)


fidencio:
@Alex-Carter01, quick question here, what's the full path where it'd be installed?

The reason I'm asking is because, when using TDVF, we install in `$PREFIX/usr/share/tdvf/`.  I'm not sure whether SEV / SEV-ES / SEV-SNP will be using the very same OVMF, but if it's not using the very same one, you need to ensure one won't override the other when creating the final tarball.

fitzthum:
>> And are we going to use exactly the same OVMF for SEV / SEV-ES / SEV-SNP?
If not, it'd require a change in the script to not override the OVMF.

I think if we build both the `AMDSEV` and `OVMF` firmware packages, we should be covered for basically any version of SEV.

fitzthum:
> I think both AMDSEV and OVMF are installing into $PREFIX/share/ovmfwe'd be overwriting one of those.
For now, this PR adds support to adding AMDSEV to the script that will generate the payload image. Will OVMF also be required?

For the first release we really only need the `AMDSEV` package. Providing the `OVMF` binary in the bundle will make things a bit more flexible, though, and should cover all the bases with SNP.

Alex-Carter01:
I think this proposed change would be best in a separate commit/PR, but in order to not have the script overwrite its past versions, is it fine to just give each version a distinct install path?
 
I am thinking we could replace `install_dir="${DESTDIR}/${PREFIX}/share/ovmf"` with `install_dir="${DESTDIR}/${PREFIX}/share/${new_variable_from_version_file}"`
And in each version entry provide a directory as a more broad version of how tdx currently switched to /tdvf/

fidencio:
> I think this proposed change would be best in a separate commit/PR, but in order to not have the script overwrite its past versions, is it fine to just give each version a distinct install path?
> 
> I am thinking we could replace `install_dir="${DESTDIR}/${PREFIX}/share/ovmf"` with `install_dir="${DESTDIR}/${PREFIX}/share/${new_variable_from_version_file}"` And in each version entry provide a directory as a more broad version of how tdx currently switched to /tdvf/

If we're only shipping one version of OVMF, I'd say nothing is needed.  Just, please, open an issue to track the possible overwrite issue for our future selves.

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -117,3 +117,6 @@ cc-tdx-td-shim-tarball:
 
 cc-tdx-tdvf-tarball:
 	${MAKE} $@-build
+
+cc-sev-ovmf-tarball:
+	${MAKE} $@-build
```
fidencio, line 122: @Alex-Carter01, could you add this as part of the cc-tarball target?

fidencio: Ah, one important thing, by the moment we add this to the tarball, we should also handle cleaning up whatever was added as part of this: https://github.com/kata-containers/kata-containers/blob/b9c0f7fb09a9f89fe370be2382b8cc911ba4a9b7/tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh#L186-L201

Please, double-check that everything you generate is cleaned up there and, if not, add there what has to be cleaned up.

Alex-Carter01:
Added the sev-ovmf target to the cc bundle.

Also added `/opt/confidential-containers/share/ovmf/` to the removed artifacts list.

wainersm (approved):
@Alex-Carter01 LGTM

stevenhorsman:
@Alex-Carter01 - I think the commit message check is stumbling over the commit heading. Can you try updating it to leave out the `CC|` bit please and just leave the subsystem as `packaging` and I think that will then pass

Alex-Carter01:
Reworded commit message title to help commit checker.

Alex-Carter01:
Change commit message to add clearer "Fixes"

PR #5036, opened 2022-08-30
kernel: Whitelist cleanup

### LINKED ISSUE #5035: CC | kernel: Cleanup whitelist.conf
Some options inside the `whitelist.conf` file are not needed (any longer) and can be removed as they are not set for any kernel.

jpecholt:
This removes two options that are not needed (any longer). These
are not set for any kernel so they do not need to be ignored either.

Fixes #5035

Signed-off-by: Joana Pecholt <joana.pecholt@aisec.fraunhofer.de>


fidencio (approved):
lgtm, thanks @jpecholt!

gkurz (approved):
LGTM. Thanks @jpecholt !

Issue #5037, opened 2022-08-30
CC | packaging: Build SEV capable kernel + efi_secret module


Alex-Carter01:
We need  an SEV capable guest kernel which can be distributed as part payloads for the Operator. This also includes building the efi_secret kernel module

stevenhorsman:
SEV is not supported now

PR #5038, opened 2022-08-30
CC | packaging: Build SEV capable kernel + efi_secret module

Alex-Carter01:
Adds a new make target for an sev kernel which can be built and put into payload bundles for the operator.

Currently not including this sev kernel target in the cc payload bundle.

Unfortunately having to breakflow from using the generic cc_tee_kernel functions in either the kata-deploy-binaries or build-kernel.
Largely based on using an upstreamed kernel release, meaning the url is the defaul cdn, and e.g. we use version rather than tag.
The upside of this is that we can use the sha sum checking functionality from the generic get_kernel function.

Fixes: #5037

### DIFF
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -82,6 +82,7 @@ cc: cc-cloud-hypervisor-tarball \
 	cc-shim-v2-tarball \
 	cc-virtiofsd-tarball \
 	cc-tdx-kernel-tarball \
+	cc-sev-kernel-tarball \
 	cc-tdx-qemu-tarball \
 	cc-tdx-tdvf-tarball
 
@@ -107,6 +108,9 @@ cc-tdx-cloud-hypervisor-tarball:
 	${MAKE} $@-build
 
 cc-tdx-kernel-tarball:
+	${MAKE} $@-build
+
+cc-sev-kernel-tarball:
 	${MAKE} $@-build
 
 cc-tdx-qemu-tarball:
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -87,6 +87,7 @@ options:
 	cc-cloud-hypervisor
 	cc-kernel
 	cc-tdx-kernel
+	cc-sev-kernel
 	cc-qemu
 	cc-tdx-qemu
 	cc-rootfs-image
@@ -161,17 +162,23 @@ install_tdx_cc_clh() {
 #Install CC kernel assert, with TEE support
 install_cc_tee_kernel() {
 	tee="${1}"
+	kernel_version="${2}"
 
-	[ "${tee}" != "tdx" ] && die "Non supported TEE"
+	[[ "${tee}" != "tdx" && "${tee}" != "sev" ]] && die "Non supported TEE"
 
-	export kernel_version="$(yq r $versions_yaml assets.kernel.${tee}.tag)"
-	export kernel_url="$(yq r $versions_yaml assets.kernel.${tee}.url)"
+	kernel_url="$(yq r $versions_yaml assets.kernel.${tee}.url)"
 	DESTDIR="${destdir}" PREFIX="${cc_prefix}" "${kernel_builder}" -x "${tee}" -v "${kernel_version}" -u "${kernel_url}"
 }
 
 #Install CC kernel assert for Intel TDX
 install_cc_tdx_kernel() {
-	install_cc_tee_kernel "tdx"
+	kernel_version="$(yq r $versions_yaml assets.kernel.tdx.tag)"
+	install_cc_tee_kernel "tdx" "${kernel_version}"
+}
+
+install_cc_sev_kernel() {
+	kernel_version="$(yq r $versions_yaml assets.kernel.sev.version)"
+	install_cc_tee_kernel "sev" "${kernel_version}"
 }
 
 install_cc_tee_qemu() {
@@ -328,6 +335,8 @@ handle_build() {
 	cc-tdx-cloud-hypervisor) install_tdx_cc_clh ;;
 
 	cc-tdx-kernel) install_cc_tdx_kernel ;;
+
+	cc-sev-kernel) install_cc_sev_kernel ;;
 
 	cc-tdx-qemu) install_cc_tdx_qemu ;;
 
--- a/tools/packaging/static-build/kernel/Dockerfile
+++ b/tools/packaging/static-build/kernel/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update && \
 	    git \
 	    iptables \
 	    libelf-dev \
+	    libssl-dev \
 	    patch && \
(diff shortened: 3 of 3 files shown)


tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -174,6 +175,12 @@ install_cc_tdx_kernel() {
 	install_cc_tee_kernel "tdx"
 }
 
+install_cc_sev_kernel() {
+	export tee="sev"
+	export kernel_version="$(yq r $versions_yaml assets.kernel.${tee}.version)"
```
gkurz, line 180: Why exporting those to the environment ?

tools/packaging/kernel/build-kernel.sh
```diff
@@ -147,7 +147,7 @@ get_kernel() {
 	[ -n "${kernel_path}" ] || die "kernel_path not provided"
 	[ ! -d "${kernel_path}" ] || die "kernel_path already exist"
 
-	if [ "${conf_guest}" != "" ]; then
+	if [ "${conf_guest}" == "tdx" ]; then
```
fitzthum, line 150: It seems like `get_tee_kernel` is supposed to be generic. Why don't we want to use it for SEV?

fidencio, line 150: Hmm.  No, please, take a look at: https://github.com/kata-containers/kata-containers/pull/4861#discussion_r942103605

Also https://github.com/kata-containers/kata-containers/blob/1b92a946d6c416b8fd1b2cb201bbe4c8669531c9/versions.yaml#L162-L165, so we can and should be using this as it's.

tools/packaging/static-build/kernel/Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN apt-get update && \
 	    git \
 	    iptables \
 	    libelf-dev \
+		libssl-dev \
```
fidencio, line 20: Wrong identation here

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
fidencio: I guess we don't need to export, indeed.  And I think it came due to a c&p from the TDX work that may also be unnecessarily exporting those (and I will fix that in a different PR).

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -109,6 +109,9 @@ cc-tdx-cloud-hypervisor-tarball:
 cc-tdx-kernel-tarball:
 	${MAKE} $@-build
 
+cc-sev-kernel-tarball:
+	${MAKE} $@-build
+
```
fidencio, line 114: Could you also add the cc-sev-kernal-tarball to the cc target?

fidencio (changes requested):
I've left a few comments, @Alex-Carter01!
Thanks for working on this.

gkurz:
> Unfortunately having to breakflow from using the generic cc_tee_kernel functions in either the kata-deploy-binaries or build-kernel.

Duplicating scriptery is generally suboptimal in the long term. What about moving kernel_version and kernel_url to the callers ?
Something like:

```
@@ -161,17 +162,24 @@ install_tdx_cc_clh() {
 #Install CC kernel assert, with TEE support
 install_cc_tee_kernel() {
        tee="${1}"
+       kernel_version="${2}"
+       kernel_url="${3}"
 
-       [ "${tee}" != "tdx" ] && die "Non supported TEE"
-
-       export kernel_version="$(yq r $versions_yaml assets.kernel.${tee}.tag)"
-       export kernel_url="$(yq r $versions_yaml assets.kernel.${tee}.url)"
-       DESTDIR="${destdir}" PREFIX="${cc_prefix}" "${kernel_builder}" -x "${tee}" -v "${kernel_version}" -u "${kernel_url}"
+       DESTDIR="${destdir}" PREFIX="${cc_prefix}" "${kernel_builder}" -x "${tee}" -v "${kernel_version}" ${kernel_url:+-u "$kernel_url"}
 }
 
 #Install CC kernel assert for Intel TDX
 install_cc_tdx_kernel() {
-       install_cc_tee_kernel "tdx"
+       kernel_version="$(yq r $versions_yaml assets.kernel.${tee}.tag)"
+       kernel_url="$(yq r $versions_yaml assets.kernel.${tee}.url)"
+
+       install_cc_tee_kernel "tdx" "$kernel_version" "$kernel_url"
+}
+
+install_cc_sev_kernel() {
+       kernel_version="$(yq r $versions_yaml assets.kernel.${tee}.version)"
+
+       install_cc_tee_kernel "sev" "$kernel_version" "$kernel_url"
 }
```

tools/packaging/kata-deploy/local-build/Makefile
fidencio: Ah, you did! :-)

Alex-Carter01:
Target added to the cc target and tabbing fixed. Redoing to make sure I am using the tee_kernel function.

Alex-Carter01:
Redone to use tee_kernel()

fitzthum (approved):

gkurz (approved):
LGTM. Thanks @Alex-Carter01 !

gkurz:
@Alex-Carter01 it seems that the `CC | ` prefix in the commit title is making the commit message check unhappy. Please just drop it.

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -82,6 +82,7 @@ cc: cc-cloud-hypervisor-tarball \
 	cc-shim-v2-tarball \
 	cc-virtiofsd-tarball \
 	cc-tdx-kernel-tarball \
+	cc-sev-kernel-tarball \
```
fidencio, line 85: Ah, one important thing, by the moment we add this to the tarball, we should also handle cleaning up whatever was added as part of this: https://github.com/kata-containers/kata-containers/blob/b9c0f7fb09a9f89fe370be2382b8cc911ba4a9b7/tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh#L186-L201

Please, double-check that everything you generate is cleaned up there and, if not, add there what has to be cleaned up.

Alex-Carter01: Everything ends up under /opt/confidential-containers/share/kata-containers/, so it looks like it taken care of.

tools/packaging/static-build/kernel/Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN apt-get update && \
 	    git \
 	    iptables \
 	    libelf-dev \
+	    libssl-dev \
```
wainersm, line 20: Needed to build the kernel?

wainersm (approved):

tools/packaging/static-build/kernel/Dockerfile
Alex-Carter01: Yes / between the kernel and the module

Alex-Carter01:
double brackets and spacing

wainersm:
@kata-containers/intel please, someone with access to TDX jobs could check this PR was the cause for the jobs failure?

Issue #5043, opened 2022-08-30
CC | Regression: `make cc-tdx-kernel-tarball` is failing due to needing `yq`


fidencio:
00aadfe20abdcc81b39e7b284efde58f087ae547 introduced a regression on `make cc-tdx-kernel-tarball` as we stopped passing all the needed information to the `build-kernel.sh` script, leading to requiring `yq` installed in the container used to build the kernel.

PR #5046, opened 2022-08-30
kernel: Re-work get_tee_kernel()

fidencio:
00aadfe20abdcc81b39e7b284efde58f087ae547 introduced a regression on
`make cc-tdx-kernel-tarball` as we stopped passing all the needed
information to the `build-kernel.sh` script, leading to requiring `yq`
installed in the container used to build the kernel.

This commit partially reverts the faulty one, rewritting it in a way the
old behaviour is brought back, without changing the behaviour that was
added by the faulty commit.

Fixes: #5043

### DIFF
--- a/tools/packaging/kernel/build-kernel.sh
+++ b/tools/packaging/kernel/build-kernel.sh
@@ -128,13 +128,12 @@ get_tee_kernel() {
 
 	[ -z "${kernel_url}" ] && kernel_url=$(get_from_kata_deps "assets.kernel.${tee}.url")
 
-	kernel_tarball="linux-${version}.tar.gz"
-	tarball_name=$(get_from_kata_deps "assets.kernel.${tee}.tarball")
-	[ -z "$tarball_name" ] || kernel_tarball="$tarball_name"
+	local kernel_tarball="${version}.tar.gz"
 
-	if [ ! -f "${kernel_tarball}" ]; then
-	   curl --fail -OL "${kernel_url}/${kernel_tarball}"
-	fi
+	# Depending on where we're getting the terball from it may have a
+	# different name, such as linux-${version}.tar.gz or simply
+	# ${version}.tar.gz.  Let's try both before failing.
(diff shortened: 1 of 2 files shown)


wainersm:
@fidencio the other day I was experimenting with kata-deploy build container and faced a similar issue due to the lack of yq. How hard would be to install it in the image?

I reviewed the commit that introduced the error; I was trying to avoid any TEE specific branching in the kernel build script... i.e. avoid "if sev then do this else do that".

fidencio:
@wainersm, I've updated the PR to try to download linux-${tarball_name} and ${tarball_name}.  This helps us to avoid depending on something else from the tests repo, and also to not add a new parameter to the already too confusing `build_kernel.sh` script.

Please, let me know what you think about this approach.

wainersm (approved):

tools/packaging/kernel/build-kernel.sh
```diff
@@ -128,13 +128,12 @@ get_tee_kernel() {
...
-	kernel_tarball="linux-${version}.tar.gz"
-	tarball_name=$(get_from_kata_deps "assets.kernel.${tee}.tarball")
-	[ -z "$tarball_name" ] || kernel_tarball="$tarball_name"
+	local kernel_tarball="${version}.tar.gz"
 
-	if [ ! -f "${kernel_tarball}" ]; then
-	   curl --fail -OL "${kernel_url}/${kernel_tarball}"
-	fi
+	# Depending on where we're getting the terball from it may have a
+	# different name, such as linuux-${version}.tar.gz or simply
```
stevenhorsman, line 134: Minor typo in the comment: `linuux`

stevenhorsman (approved):
LGTM. Thanks

tools/packaging/kernel/build-kernel.sh
fidencio: Fixes, thanks!

Issue #4895, opened 2022-08-12
Random fails due to neighbors in dragonball test


liubin:
@studychao @Tim-0731-Hzt PTAL

[Here](http://jenkins.katacontainers.io/job/kata-containers-dragonball-ubuntu-20.04-PR/19/consoleFull)

> time="2022-08-10T00:49:06Z" level=info msg="Using the following image list: {Alpine:docker.io/library/alpine:latest BusyBox:docker.io/library/busybox:latest Pause:k8s.gcr.io/pause:3.5 ResourceConsumer:k8s.gcr.io/e2e-test-images/resource-consumer:1.9 VolumeCopyUp:gcr.io/k8s-cri-containerd/volume-copy-up:2.0 VolumeOwnership:gcr.io/k8s-cri-containerd/volume-ownership:2.0}"
=== RUN   TestContainerRestart
    container_restart_test.go:17: Create a pod config and run sandbox container
E0810 00:49:09.462276   75505 remote_runtime.go:130] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim: Others("failed to handler message try init runtime instance\n\nCaused by:\n    0: init runtime handler\n    1: start sandbox\n    2: setup device after start vm\n    3: handle neighbors\n    4: update neighbors\n    5: rpc status: code: INTERNAL message: \"\\\"Failed to add ARP neighbours: Failed to add ARP neighbor ff02::1:ffd0:c735: Received a netlink error message File exists (os error 17)\\\"\""): unknown
    container_restart_test.go:20: 
        	Error Trace:	container_restart_test.go:20
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unknown desc = failed to create containerd task: failed to create shim: Others("failed to handler message try init runtime instance\n\nCaused by:\n    0: init runtime handler\n    1: start sandbox\n    2: setup device after start vm\n    3: handle neighbors\n    4: update neighbors\n    5: rpc status: code: INTERNAL message: \"\\\"Failed to add ARP neighbours: Failed to add ARP neighbor ff02::1:ffd0:c735: Received a netlink error message File exists (os error 17)\\\"\""): unknown
        	Test:       	TestContainerRestart
--- FAIL: TestContainerRestart (3.28s)
FAIL

And [here](http://jenkins.katacontainers.io/job/kata-containers-dragonball-ubuntu-20.04-PR/24/consoleFull):

> + cri-integration
[...]

PR #5074, opened 2022-09-01
runtime-rs: debug update neighbours

Tim-0731-Hzt:
This commit is just for debuging. Sometimes updating neighbors will fail with a netlink error File exists.
we need to check if there is any duplicate neighbour ip before pass that
to the agent.

Fixes: #4895


jodh-intel (approved):
Thanks @Tim-0731-Hzt.

lgtm

Issue #5078, opened 2022-09-02
`make cc-tdx-qemu-tarball` target output to incorrect directory


ryansavino:
# Description of problem

Branch `CCv0`

Directory:
`kata-containers` OR
`kata-containers/tools/packaging/kata-deploy/local-build`

Command Run:
`make cc-tdx-qemu-tarball`

When building the tdx qemu tarball, output directory is missing suffix. 

# Expected result

qemu directories should be `kata-qemu-tdx`.

# Actual result

Instead qemu is built in directory `kata-qemu`.

```
drwxr-xr-x root/root         0 2022-09-01 21:40 ./
drwxr-xr-x root/root         0 2022-09-01 20:36 ./opt/
drwxr-xr-x root/root         0 2022-09-01 20:36 ./opt/confidential-containers/
drwxr-xr-x root/root         0 2022-09-01 20:36 ./opt/confidential-containers/bin/
-rwxr-xr-x root/root  18368760 2022-09-01 20:36 ./opt/confidential-containers/bin/qemu-system-x86_64-tdx
drwxr-xr-x root/root         0 2022-09-01 20:36 ./opt/confidential-containers/share/
drwxr-xr-x root/root         0 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/
drwxr-xr-x root/root         0 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/
-rw-r--r-- root/root      1536 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/pvh.bin
-rw-r--r-- root/root    540672 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/edk2-i386-vars.fd
-rw-r--r-- root/root     79688 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/s390-netboot.img
-rw-r--r-- root/root   3653632 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/edk2-x86_64-code.fd
-rw-r--r-- root/root   3653632 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/edk2-x86_64-secure-code.fd
-rw-r--r-- root/root   3653632 2022-09-01 20:36 ./opt/confidential-containers/share/kata-qemu/qemu/edk2-i386-code.fd
[...]

PR #5079, opened 2022-09-02
qemu: fix tdx qemu tarball directories

ryansavino:
Dockerfile cannot decipher multiple conditional statements in the main RUN call.
Cannot segregate statements in Dockerfile with '{}' braces without wrapping entire statement in 'bash -c' statement.
Dockerfile does not support setting variables by bash command.
Must set HYPERVISOR_NAME and PKGVERSION from parent script: build-base-qemu.sh

Fixes: #5078


fidencio (approved):
lgtm, thanks @ryansavino!

stevenhorsman (approved):
LGTM. Thanks!

Issue #5080, opened 2022-09-02
arm64: kata base on cloud hypervisor fail to start in hypervisor/debug mode


jongwu:
When hypervisor debug mode is enabled, kata bases on Cloud Hypervisor will not start.

PR #4956, opened 2022-08-19
kernel: Add CONFIG_CGROUP_HUGETLB=y as part of the cgroup fragments

shippomx:
fix #4953
@liangxianlong

Signed-off-by: Miao Xia <xia.miao1@zte.com.cn>

PR #5081, opened 2022-09-02
clh:arm64: use arm AMBA uart for hypervisor debug

jongwu:
cloud hypervisor on arm64 only support arm AMBA UART(pl011) as
tty. So, the console should be set to "ttyAMA0" instead of "ttyS0"
when enable hypervisor debug mode.
Also, Arm kernel should enable pl011 support.

Fixes: #5080


tools/packaging/kernel/configs/fragments/arm64/serial.conf
```diff
@@ -1,3 +1,7 @@
 # This option is used for all 8250 compatible serial ports
 # that are probed through device tree.
 CONFIG_SERIAL_OF_PLATFORM=y
+
+# Enable Arm AMBA UART
+CONFIG_SERIAL_AMBA_PL011=y
+CONFIG_SERIAL_AMBA_PL011_CONSOLE=y
```
fidencio, line 7: Please, also bump the kernel version with this change.

sboeuf (commented):
Looks good to me

tools/packaging/kernel/configs/fragments/arm64/serial.conf
jongwu: Done, thanks.

jongwu:
I'm confused why this patch affect clh test on x86. Any idea? @fidencio

fidencio:
@jongwu, I must that i don't fully understand why it'd fail.

snir911 (approved):
LGTM thanks!

jongwu:
After lots of tests, I find that bumping config version will lead to the failure of clh test on x86, but I don't know why. 

> 16:53:20 INFO: Run tests
16:53:20 1..1
16:53:20 not ok 1 Running with postStart and preStop handlers

jongwu:
I raise a issue [#5196 ](https://github.com/kata-containers/kata-containers/issues/5196) to track this problem.

jongwu:
Hello @fidencio -, as I can't fix the clh CI test caused by bumping kata_config_version for now, can I cancel the bump op?

jongwu:
Hello @jpecholt @shippomx, it seems both of your change, [cmommit1](https://github.com/kata-containers/kata-containers/pull/4956) and [commit2](https://github.com/kata-containers/kata-containers/pull/5036) cause the clh test failure. Could you please take a look at this? Thanks!

jpecholt:
Hi @jongwu, if I understand you correctly, you are saying that my whitelist cleanup is causing issues or having an impact on your code? I am slightly confused as to why this should be the case, the commit cleaned up kernel config options that weren't used by any kernel to begin with and thus needn't be whitelisted for compatibility reasons either. The only kernel that would be remotely impacted is the AMD SEV one. I checked out your PR and verified: the kernel configs really shouldn't have any impact as they are in the whitelist and one extra comment only.

All I can see from the tests that are failing right now (not all have run yet) is missing backport labels and that the commit message of one of your commits requires a reference to this PR... I'd have a look again once the other tests ran through

jongwu:
Hello @jpecholt -, sorry for having not made it clear.

For the current code, once kernel_config_version is bumped, the clh test will fail. After lots of tests, I find that the recent two changes for kernel config, that's the 2 commits I mentioned in my last thread, cause the failure as the clh test passes after the changes are reverted. Please check the second commit in this PR. So the 2 changes don't impact on my code but on the clh test. For now, I can just give the test result but no reason for that. That's why I need your help to find out why.

PR #5082, opened 2022-09-02
dragonball: Fix problem that stdio console cannot connect to stdout

QiliangFan:
- [x] Fix bug: the stdio console cannot connect to stdout stream

Fixes: #5083
Signed-off-by: qiliangfan <fanqiliang@mail.nankai.edu.cn>

### DIFF
--- a/src/dragonball/src/device_manager/mod.rs
+++ b/src/dragonball/src/device_manager/mod.rs
@@ -591,14 +591,19 @@ impl DeviceManager {
             .map_err(|_| StartMicroVmError::EventFd)?;
 
         info!(self.logger, "init console path: {:?}", com1_sock_path);
-        if let Some(path) = com1_sock_path {
-            if let Some(legacy_manager) = self.legacy_manager.as_ref() {
-                let com1 = legacy_manager.get_com1_serial();
-                self.con_manager
-                    .create_socket_console(com1, path)
-                    .map_err(StartMicroVmError::DeviceManager)?;
+
+        if let Some(legacy_manager) = self.legacy_manager.as_ref() {
+            if let Some(path) = com1_sock_path {
+                // Currently, the `com1_sock_path` "stdio" is only reserved for creating the stdio console
+                if path != "stdio" {
+                    let com1 = legacy_manager.get_com1_serial();
+                    self.con_manager
+                        .create_socket_console(com1, path)
+                        .map_err(StartMicroVmError::DeviceManager)?;
+                    return Ok(());
+                }
             }
-        } else if let Some(legacy_manager) = self.legacy_manager.as_ref() {
+
             let com1 = legacy_manager.get_com1_serial();
             self.con_manager
                 .create_stdio_console(com1)
--- a/src/dragonball/src/device_manager/console_manager.rs
+++ b/src/dragonball/src/device_manager/console_manager.rs
@@ -74,11 +74,20 @@ impl ConsoleManager {
 
     /// Create a console backend device by using stdio streams.
     pub fn create_stdio_console(&mut self, device: Arc<Mutex<SerialDevice>>) -> Result<()> {
+        device
+            .lock()
+            .unwrap()
+            .set_output_stream(Some(Box::new(std::io::stdout())));
         let stdin_handle = std::io::stdin();
         stdin_handle
             .lock()
             .set_raw_mode()
             .map_err(|e| DeviceMgrError::ConsoleManager(ConsoleManagerError::StdinHandle(e)))?;
+        stdin_handle
+            .lock()
+            .set_non_block(true)
+            .map_err(ConsoleManagerError::StdinHandle)
+            .map_err(DeviceMgrError::ConsoleManager)?;
 
         let handler = ConsoleEpollHandler::new(device, Some(stdin_handle), None, &self.logger);
         self.subscriber_id = Some(self.epoll_mgr.add_subscriber(Box::new(handler)));


src/dragonball/src/device_manager/mod.rs
```diff
@@ -591,14 +591,24 @@ impl DeviceManager {
...
+                if path == "None" {
+                    info!(
+                        _ctx.logger(),
+                        "This token is used to create stdio console. (means not to create sock console)"
+                    );
+                } else if let Some(legacy_manager) = self.legacy_manager.as_ref() {
+                    let com1 = legacy_manager.get_com1_serial();
+                    self.con_manager
+                        .create_socket_console(com1, path)
+                        .map_err(StartMicroVmError::DeviceManager)?;
```
wllenyj, line 605: Is `Return` required here?

QiliangFan: It is ok to create the sock console and stdio console simultaneously, but I will add the `return` statement to avoid unnecessary confusion.

liubin:
@QiliangFan your commit message subject is too long:

>   1a8b1287644c6b1c4d7ff0c225ac1f6075ada864    Subject too long (max 75)

QiliangFan:
> @QiliangFan your commit message subject is too long:
> 
> > [1a8b128](https://github.com/kata-containers/kata-containers/commit/1a8b1287644c6b1c4d7ff0c225ac1f6075ada864)    Subject too long (max 75)

:smile:  Thank you. Commit subgject is modified.

src/dragonball/src/lib.rs
```diff
@@ -32,7 +32,7 @@ pub mod vcpu;
 /// Virtual machine manager for virtual machines.
 pub mod vm;
 
-mod event_manager;
+pub mod event_manager;
```
studychao, line 35: Why does this switch to `pub`?
Don't know if I miss something, but I couldn't find where you use event_manager in thie pull request.

src/dragonball/src/device_manager/mod.rs
```diff
@@ -591,14 +591,25 @@ impl DeviceManager {
...
+                    return Ok(());
+                }
             }
-        } else if let Some(legacy_manager) = self.legacy_manager.as_ref() {
+            None => {}
```
studychao, line 609: If you do nothing when the `com1_sock_path` is `None`, why not use `if let`?
e.g.
```
if let Some(path) = com1_sock_path {
     if path == "None" {
    .......
} else if ........ {
   .....
}

}
```

src/dragonball/src/device_manager/mod.rs
```diff
@@ -591,14 +591,25 @@ impl DeviceManager {
...
         info!(self.logger, "init console path: {:?}", com1_sock_path);
-        if let Some(path) = com1_sock_path {
-            if let Some(legacy_manager) = self.legacy_manager.as_ref() {
-                let com1 = legacy_manager.get_com1_serial();
-                self.con_manager
-                    .create_socket_console(com1, path)
-                    .map_err(StartMicroVmError::DeviceManager)?;
+        match com1_sock_path {
+            Some(path) => {
+                if path == "None" {
```
studychao, line 596: Why do you need to check if this is `"None"`?
Why can't this case be covered by `match com1_sock_path` which `com1_sock_path` is `None`?

src/dragonball/src/device_manager/mod.rs
```diff
@@ -591,14 +591,25 @@ impl DeviceManager {
...
+                    info!(
+                        _ctx.logger(),
```
studychao, line 598: why `_ctx.logger()` is used here?
BTW, `_` in Rust means this parameter is not used in the function, but you use `ctx` here.

studychao (commented):
Thanks. I left few comments.

src/dragonball/src/lib.rs
QiliangFan: `event_manager` is previously used in a CLI parser. After removing unused components, I found this modification unnecessary. So I will reset the access policy.

src/dragonball/src/device_manager/mod.rs
QiliangFan: In `src/api/v1/vmm_action.rs`:

```rust
        let serial_path = match machine_config.serial_path {
            Some(value) => value,
            None => {
                if config.serial_path.is_none() {
                    String::from("/run/dragonball/") + &vm_id + "_com1"
                } else {
                    // Safe to unwrap() because we have checked it has a value.
                    config.serial_path.as_ref().unwrap().clone()
                }
            }
        };
```

This code snippet always assigns a default value to serial_path, namely the `com1_sock_path`. I couldn't figure out whether the default value is required in other places. So in the CLI parser, I set a special string value "None" to make a distinction.

src/dragonball/src/device_manager/mod.rs
QiliangFan: This case can be seen in PR: [Create VM from command line arguments](https://github.com/openanolis/dbs-cli/pull/1). As described in the last comment, I must set a unique value to make a distinction.

QiliangFan:
```rust
            // as in crate `dragonball` serial_path will be assigned with a default value,
            // we need a special token to enable the stdio console.
            serial_path: Some("None".to_owned()),
```

src/dragonball/src/device_manager/mod.rs
QiliangFan: :smile:  Is ok to modify the `_ctx` to `ctx`?

src/dragonball/src/device_manager/mod.rs
QiliangFan: I am some kind of confused with the `if let` statement. Since the `com1_sock_path` can not be `None`, and we only need to consider the special string value `"None"`. If I am supposed to do this?

```rust 
if let some("None") == com1_sock_path {
                    info!(
                        ctx.logger(),
                        "This token is used to create stdio console. (means not to create sock console)"
                    );
                }
```

QiliangFan:
```rust
        if let some(path) = com1_sock_path {
            // The token "None" is used to create stdio console. (means not to create sock console)"
            if path != "None" {
                let com1 = legacy_manager.get_com1_serial();
                self.con_manager
                    .create_socket_console(com1, path)
                    .map_err(StartMicroVmError::DeviceManager)?;
                return Ok(());
            }
        }
```

src/dragonball/src/device_manager/mod.rs
studychao: That's strange. Could you just set it to `None` instead of string `"None"`?

src/dragonball/src/device_manager/mod.rs
studychao: I think there is no need here to print the log.
Also, path == "None" looks strange in Rust, please help to see if this could be better

src/dragonball/src/device_manager/mod.rs
studychao: Yes! You use `if let` correctly, enjoy Rust:)

Again, please help to see string `"None"` could be optimized.

studychao (commented):
Hi Qiliang, a few more comments.

src/dragonball/src/device_manager/mod.rs
QiliangFan: > ```rust
> if let some(path) = com1_sock_path {
>     // The token "None" is used to create stdio console. (means not to create sock console)"
>     if path != "None" {
>         let com1 = legacy_manager.get_com1_serial();
>         self.con_manager
>             .create_socket_console(com1, path)
>             .map_err(StartMicroVmError::DeviceManager)?;
>         return Ok(());
>     }
> }
> ```

If `None` is given, we would always get a default value, which is done by this code snippet. It ok to set it to `None`, but the default value of serial_path is unexpected.

src/dragonball/src/device_manager/mod.rs
QiliangFan: Should I remove the default value of `serial_path` when `serial_path` is `None`? Or I just change another intuitive string value for it?

QiliangFan:
> Hi Qiliang, a few more comments.

Thank you :smile:. But we must need a special string value. I wonder if the Dragonball is designed to use the sock console at the very beginning, so `serial_path` always has a string value. But right now, when it comes to the stdio console, there is no need to create a sock console. However, the mechanism that we assign a default value to the serial path when it is `None` make it not so easy to make the process concise.

src/dragonball/src/device_manager/mod.rs
studychao: Why default value is unexpected?
Do you mean the default value could not work to create a socket console?

QiliangFan: When it is time to create a stdio console, we don't want to create a socket console. Similarly, it is not supposed to create a stdio console when we need a socket console. The default value makes these two separate operations bound together, therefore we must set a special value to distinguish them. To make it clear, I change the `"None"` to `"stdio"`.

```rust
if let Some(legacy_manager) = self.legacy_manager.as_ref() {
    if let Some(path) = com1_sock_path {
        if "stdio" != path {
            let com1 = legacy_manager.get_com1_serial();
            self.con_manager
                .create_socket_console(com1, path)
                .map_err(StartMicroVmError::DeviceManager)?;
            return Ok(());
        }
    }

    let com1 = legacy_manager.get_com1_serial();
    self.con_manager
        .create_stdio_console(com1)
        .map_err(StartMicroVmError::DeviceManager)?;
}
```

QiliangFan: By the way,  it seems that we cannot remove the functionality, setting a default value to serial path.

QiliangFan: > Why default value is unexpected? Do you mean the default value could not work to create a socket console?

With the default value, `device_manager` will create two consoles (socket console and stdio console) simultaneously, which is unexpected.

QiliangFan:
@studychao   :thinking:  Thanks for your previous review. Since the default value of `com1_sock_path` will disturb the branch of creating an interactive console, setting a special string value is the simplest way to ensure the correctness of creating a console without the modification to API. Is there any more elegant way?

src/dragonball/src/device_manager/mod.rs
```diff
@@ -591,14 +592,19 @@ impl DeviceManager {
...
-            if let Some(legacy_manager) = self.legacy_manager.as_ref() {
-                let com1 = legacy_manager.get_com1_serial();
-                self.con_manager
-                    .create_socket_console(com1, path)
-                    .map_err(StartMicroVmError::DeviceManager)?;
+
+        if let Some(legacy_manager) = self.legacy_manager.as_ref() {
+            if let Some(path) = com1_sock_path {
+                // Currently, the `com1_sock_path` "stdio" is only reserved for creating the stdio console
+                if "stdio" != path {
```
Tim-Zhang, line 599: `if path != "stdio"` would be better.

QiliangFan: 😄 Thank you. Modified already.

studychao (commented):
LGTM

studychao:
Got that, thanks @QiliangFan

QiliangFan:
:cry:  Is there anything I need to do?

wllenyj (approved):

Tim-Zhang (approved):

Issue #5087, opened 2022-09-02
Shim process leak in case of startup failure


liubin:
How to reproduce:

1. Have a configuration file with error, for example, invalid `runtime.hypervisor_name`.

2. run `ctr run -t --rm --runtime io.containerd.kata.v2  docker.io/containerstack/alpine-stress:latest c1 sh`

Result:

```shell
$ ps -ef | grep kata

root     26837     1  0 16:32 ?        00:00:00 /home/kant/containerd-shim-kata-v2 -id 813d023657fbf5aa993eb79a5225738cbc42a6666cf3d5ca3b8a55bf92cf54c6 -namespace default -address /run/containerd/containerd.sock -publish-binary /home/kant/containerd -debug

```

There are no resources remain in containerd, the shim process is not exit.

Some log:

> Sep  2 16:08:20 kant-server containerd[902]: {"msg":"====> task service ShutdownContainer(ShutdownRequest { container_id: \"77e97c7b2d4c8a4ddab219e93760d2a3a8e939ca9feb63a5a890add282124347\", is_now: false })","level":"DEBG","ts":"2022-09-02T08:08:20.035844202Z","steam id":"5","subsystem":"service","name":"kata-runtime","pid":"26254","version":"0.1.0","source":"77e97c7b2d4c8a4ddab219e93760d2a3a8e939ca9feb63a5a890add282124347"}
Sep  2 16:08:20 kant-server containerd[902]: time="2022-09-02T16:08:20.035986279+08:00" level=info msg="shim disconnected" id=77e97c7b2d4c8a4ddab219e93760d2a3a8e939ca9feb63a5a890add282124347
Sep  2 16:08:20 kant-server containerd[902]: time="2022-09-02T16:08:20.036026367+08:00" level=warning msg="cleaning up after shim disconnected" id=77e97c7b2d4c8a4ddab219e93760d2a3a8e939ca9feb63a5a890add282124347 namespace=default
Sep  2 16:08:20 kant-server containerd[902]: time="2022-09-02T16:08:20.036034699+08:00" level=info msg="cleaning up dead shim"
Sep  2 16:08:20 kant-server containerd[902]: time="2022-09-02T16:08:20.040925586+08:00" level=debug msg="remove snapshot" key=77e97c7b2d4c8a4ddab219e93760d2a3a8e939ca9feb63a5a890add282124347
[...]

PR #5088, opened 2022-09-02
runtime-rs: force shutdown shim process in it can't exit

liubin:
In some case the call of cleanup from shim to service manager will fail,
and the shim process will continue to running, that will make process leak.

This commit will force shutdown the shim process in case of any errors in
service crate.

Fixes: #5087

### DIFF
--- a/src/runtime-rs/crates/shim/src/shim_delete.rs
+++ b/src/runtime-rs/crates/shim/src/shim_delete.rs
@@ -6,6 +6,9 @@
 
 use anyhow::{Context, Result};
 use containerd_shim_protos::api;
+use kata_sys_util::spec::{get_bundle_path, get_contaier_type, load_oci_spec};
+use kata_types::container::ContainerType;
+use nix::{sys::signal::kill, sys::signal::SIGKILL, unistd::Pid};
 use protobuf::Message;
 use std::{fs, path::Path};
 
@@ -14,7 +17,7 @@ use crate::{shim::ShimExecutor, Error};
 impl ShimExecutor {
     pub async fn delete(&mut self) -> Result<()> {
         self.args.validate(true).context("validate")?;
-        let rsp = self.do_cleanup().await.context("do cleanup")?;
+        let rsp = self.do_cleanup().await.context("shim do cleanup")?;
         rsp.write_to_writer(&mut std::io::stdout())
             .context(Error::FileWrite(format!("write {:?} to stdout", rsp)))?;
         Ok(())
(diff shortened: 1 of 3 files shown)


cmaf (approved):
thanks @liubin

src/runtime-rs/crates/shim/src/shim_delete.rs
```diff
@@ -41,9 +44,28 @@ impl ShimExecutor {
...
+
+        if let Err(e) = service::ServiceManager::cleanup(&self.args.id).await {
+            error!(
+                sl!(),
+                "failed to cleanup in service manager: {:?}. force shutdown shim process", e
+            );
+
+            let bundle_path = get_bundle_path().context("get bundle path")?;
+            if let Ok(spec) = load_oci_spec() {
+                if let Ok(ContainerType::PodSandbox) = get_contaier_type(&spec) {
```
jodh-intel, line 56: Typo? `get_container_type()`?

src/runtime-rs/crates/shim/src/shim_delete.rs
```diff
@@ -41,9 +44,28 @@ impl ShimExecutor {
...
-            .await
-            .context("cleanup")?;
+
+        if let Err(e) = service::ServiceManager::cleanup(&self.args.id).await {
+            error!(
+                sl!(),
+                "failed to cleanup in service manager: {:?}. force shutdown shim process", e
+            );
+
+            let bundle_path = get_bundle_path().context("get bundle path")?;
```
jodh-intel, line 54: tbh, I'd prefer to have a `self.force_shutdown()` method.

src/runtime-rs/crates/shim/src/shim_delete.rs
```diff
@@ -41,9 +44,28 @@ impl ShimExecutor {
...
+            if let Ok(spec) = load_oci_spec() {
+                if let Ok(ContainerType::PodSandbox) = get_contaier_type(&spec) {
+                    // only force shutdown for sandbox container
+                    if let Ok(shim_pid) = self.read_pid_file(&bundle_path) {
```
jodh-intel, line 58: Is it necessary to invalidate `self` after the kill (something like `self.shutdown = true`? If so, that would make it easier to unit test (particularly if the method was specified as:

```rust
fn force_shutdown(&self, signal: nix::signal::Signal) { ... }

jodh-intel (commented):
Thanks @liubin - a few comments...

PR #5090, opened 2022-09-02
qemu: Keep passing BUILD_SUFFIX

fidencio:
In the commit 54d6d01754b689a32da852cfd85a69d2c4730baa we ended up
removing the BUILD_SUFFIX argument passed to QEMU as it only seemed to
be used to generate the HYPERVISOR_NAME and PKGVERSION, which were added
as arguments to the dockerfile.

However, it turns out BUILD_SUFFIX is used by the `qemu-build-post.sh`
script, so it can rename the QEMU binary accordingly.

Let's just bring it back.

Fixes: #5078


stevenhorsman (approved):
Thanks for the fix :)

wainersm (approved):

Issue #5094, opened 2022-09-02
SEV kernel build failing


ryansavino:
# Description of problem

SEV kernel build is failing.

Directory:
`kata-containers/tools/packaging/kernel`

Command Run:
`build-kernel.sh -v 5.19.2 -x sev setup`

# Expected result

Built kernel without any errors.

# Actual result

```
11:23:44 + /home/jenkins/workspace/wainer-CCv0-ubuntu-20.04-x86_64-CC_SEV_CRI_CONTAINERD_K8S/go/src/github.com/kata-containers/kata-containers/tools/packaging/kernel/build-kernel.sh -v 5.19.2 -x sev setup
11:23:44 INFO: Config version: 96
11:23:44 INFO: Kernel version: 5.19.2
11:23:44 INFO: kernel path does not exist, will download kernel
11:23:44   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
11:23:44                                  Dload  Upload   Total   Spent    Left  Speed
11:23:44 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0  198M    0 15888    0     0   7544      0  7:40:13  0:00:02  7:40:11  7540
100  198M  100  198M    0     0  80.6M      0  0:00:02  0:00:02 --:--:-- 80.5M
11:23:46 tar: 5.19.2.tar.gz: Cannot open: No such file or directory
11:23:46 tar: Error is not recoverable: exiting now
11:23:46 ++ handle_error 55
11:23:46 ++ local exit_code=2
11:23:46 ++ local line_number=55
11:23:46 +++ basename .ci/install_kata_kernel.sh
```

ryansavino:
Tarball name issue. Making a PR.

PR #5095, opened 2022-09-02
kernel: fix kernel tarball name for SEV

ryansavino:
'linux-' prefix needed for tarball name in SEV case.

Fixes: #5094

Signed-Off-By: Ryan Savino <ryan.savino@amd.com>

PR #5096, opened 2022-09-03
kernel: fix kernel tarball name for SEV for CCv0

ryansavino:
Commit cherry picked from #5095. This is for CCv0.

'linux-' prefix needed for tarball name in SEV case.

Fixes: #5094

Signed-Off-By: Ryan Savino [ryan.savino@amd.com](mailto:ryan.savino@amd.com)

### DIFF
--- a/tools/packaging/kernel/build-kernel.sh
+++ b/tools/packaging/kernel/build-kernel.sh
@@ -133,7 +133,7 @@ get_tee_kernel() {
 	# Depending on where we're getting the terball from it may have a
 	# different name, such as linux-${version}.tar.gz or simply
 	# ${version}.tar.gz.  Let's try both before failing.
-	curl --fail -OL "${kernel_url}/linux-${kernel_tarball}" || curl --fail -OL "${kernel_url}/${kernel_tarball}"
+	curl --fail -L "${kernel_url}/linux-${kernel_tarball}" -o ${kernel_tarball} || curl --fail -OL "${kernel_url}/${kernel_tarball}"
 	
 	mkdir -p ${kernel_path}
 	tar --strip-components=1 -xf ${kernel_tarball} -C ${kernel_path}


tools/packaging/kernel/build-kernel.sh
```diff
@@ -136,6 +136,7 @@ get_tee_kernel() {
 	curl --fail -OL "${kernel_url}/linux-${kernel_tarball}" || curl --fail -OL "${kernel_url}/${kernel_tarball}"
```
c3d, line 136:
```suggestion
curl --fail -L "${kernel_url}/linux-${kernel_tarball}" -o ${kernel_tarball} || curl --fail -OL "${kernel_url}/${kernel_tarball}"
```

tools/packaging/kernel/build-kernel.sh
```diff
@@ -136,6 +136,7 @@ get_tee_kernel() {
...
 	
 	mkdir -p ${kernel_path}
+	tar --strip-components=1 -xf linux-${kernel_tarball} -C ${kernel_path} || \
```
c3d, line 139: With the above, this would no longer be needed.

c3d (commented):
Looks good, but I suggested an alternative, since there is already an `||` in the code we can use.

tools/packaging/kernel/build-kernel.sh
ryansavino: done

tools/packaging/kernel/build-kernel.sh
ryansavino: done

ryansavino:
Not sure why this closed. Might have something to do with a reset and a push I did.

Made the changes @c3d suggested. Would be great if this could get in today as it's holding up the SEV CI.

fitzthum (approved):

Issue #5068, opened 2022-09-01
runtime-rs: debug console support in runtime


Ji-Xinyou:
we need read debug console configuration and pass to kernel parameters for agent to read. Currently, go runtime uses http + kata-monitor to retreive socket path and scheme, if we want to use this directly, we rather hard-code this socket address by `hvsock:///run/kata/ + sandboxid + /root/kata/hvsock:1026'.

I think since currently runtime-rs + dbs only uses hvsock on host side when commnunicating with guest, we can add an separate CLI binary in Rust **or** we can write a kata-monitor in Rust runtime, but it should be in another PR.

PR #5097, opened 2022-09-05
runtime-rs: debug console support in runtime

Ji-Xinyou:
Read debug console configuration in kernel params. The debug console is opened when the configuration is set, and to be consistent with kata 2.x, the port is default to be 1026 in runtime. A console can log into the pod's shell by connecting to vport 1026.

Fixes: #5068
Signed-Off-By: Ji-Xinyou <jerryji0414@outlook.com>

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/mod.rs
+++ b/src/libs/kata-types/src/config/hypervisor/mod.rs
@@ -50,6 +50,8 @@ const VIRTIO_FS: &str = "virtio-fs";
 const VIRTIO_FS_INLINE: &str = "inline-virtio-fs";
 const MAX_BRIDGE_SIZE: u32 = 5;
 
+const KERNEL_PARAM_DELIMITER: &str = " ";
+
 lazy_static! {
     static ref HYPERVISOR_PLUGINS: Mutex<HashMap<String, Arc<dyn ConfigPlugin>>> =
         Mutex::new(HashMap::new());
@@ -235,6 +237,14 @@ impl BootInfo {
             return Err(eother!("Can not configure both initrd and image for boot"));
         }
         Ok(())
+    }
+
+    /// Add kernel parameters to bootinfo. It is always added before the original
+    /// to let the original one takes priority
+    pub fn add_kparams(&mut self, params: Vec<String>) {
+        let mut p = params;
+        p.push(self.kernel_params.clone()); // [new_params0, new_params1, ..., original_params]
+        self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
     }
 
     /// Validate guest kernel image annotaion
--- a/src/libs/kata-types/src/config/hypervisor/mod.rs
+++ b/src/libs/kata-types/src/config/hypervisor/mod.rs
@@ -243,8 +243,12 @@ impl BootInfo {
     /// to let the original one takes priority
     pub fn add_kparams(&mut self, params: Vec<String>) {
         let mut p = params;
-        p.push(self.kernel_params.clone()); // [new_params0, new_params1, ..., original_params]
-        self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
+        if self.kernel_params.is_empty() {
+            self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
+        } else {
+            p.push(self.kernel_params.clone()); // [new_params0, new_params1, ..., original_params]
+            self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
+        }
     }
 
     /// Validate guest kernel image annotaion
@@ -1076,5 +1080,32 @@ mod tests {
 
         assert!(get_hypervisor_plugin("dragonball").is_some());
         assert!(get_hypervisor_plugin("dragonball2").is_none());
+    }
+
+    #[test]
+    fn test_add_kparams() {
+        let mut boot_info = BootInfo {
+            ..Default::default()
+        };
+        let params = vec![
+            String::from("foo"),
+            String::from("bar"),
+            String::from("baz=faz"),
+        ];
+        boot_info.add_kparams(params);
+
+        assert_eq!(boot_info.kernel_params, String::from("foo bar baz=faz"));
+
+        let new_params = vec![
+            String::from("boo=far"),
+            String::from("a"),
+            String::from("b=c"),
+        ];
+        boot_info.add_kparams(new_params);
+
+        assert_eq!(
+            boot_info.kernel_params,
+            String::from("boo=far a b=c foo bar baz=faz")
+        );
     }
 }
--- a/src/libs/kata-types/src/config/hypervisor/mod.rs
+++ b/src/libs/kata-types/src/config/hypervisor/mod.rs
@@ -241,14 +241,12 @@ impl BootInfo {
 
     /// Add kernel parameters to bootinfo. It is always added before the original
     /// to let the original one takes priority
-    pub fn add_kparams(&mut self, params: Vec<String>) {
+    pub fn add_kernel_params(&mut self, params: Vec<String>) {
         let mut p = params;
-        if self.kernel_params.is_empty() {
-            self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
-        } else {
+        if !self.kernel_params.is_empty() {
             p.push(self.kernel_params.clone()); // [new_params0, new_params1, ..., original_params]
-            self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
         }
+        self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
     }
 
     /// Validate guest kernel image annotaion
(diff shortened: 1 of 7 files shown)


src/runtime-rs/crates/runtimes/virt_container/src/lib.rs
```diff
@@ -125,6 +137,38 @@ fn new_agent(toml_config: &TomlConfig) -> Result<Arc<KataAgent>> {
     }
 }
 
+// agent-specific kernel parameters
+fn agent_kparams(toml_config: &TomlConfig) -> Result<Vec<String>> {
```
Tim-0731-Hzt, line 141: I think it's better to do the operation of handling config file when loading the toml config. Try to put this inside `kata-types/src/config`

Ji-Xinyou: Thanks for the review, @Tim-0731-Hzt 

The way you suggest will bring unwanted dependencies to libs, but I will try to do this in another way.
- 1. From toml_config, get a kv of agent kernel parameters (this can be done in libs using `HashMap<String, String>`)
- 2. From the kernel parameters kv, get the agent parameters and update this to hypervisor config (this can be done in load_config)

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -237,6 +239,14 @@ impl BootInfo {
         Ok(())
     }
 
+    /// Add kernel parameters to bootinfo. It is always added before the original
+    /// to let the original one takes priority
+    pub fn add_kparams(&mut self, params: Vec<String>) {
+        let mut p = params;
+        p.push(self.kernel_params.clone()); // [new_params0, new_params1, ..., original_params]
+        self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
+    }
```
cmaf, line 248: It would be helpful to check for errors somewhere in here.

src/libs/kata-types/src/config/mod.rs
```diff
@@ -149,7 +149,29 @@ impl TomlConfig {
...
+                kv.insert("agent.trace".to_string(), "true".to_string());
+            }
+            if cfg.container_pipe_size > 0 {
+                let container_pipe_size = cfg.container_pipe_size.to_string();
+                kv.insert("agent.container_pipe_size".to_string(), container_pipe_size);
+            }
+            if cfg.debug_console_enabled {
+                kv.insert("agent.debug_console".to_string(), "".to_string());
+                kv.insert("agent.debug_console_vport".to_string(), "1026".to_string());
+            }
```
cmaf, line 169: Should we have the strings here (`agent.trace`, `1026`, etc) as `const` variables somewhere?

cmaf (commented):
Hi @Ji-Xinyou, a few questions/comments below. Also, I think `add_kparams` and `get_agent_kparams` should have unit tests.

src/libs/kata-types/src/config/hypervisor/mod.rs
Ji-Xinyou: Sorry for the late reply. I will add some unit tests for `add_kparams` and `get_agent_kparams`, and add some error checking in `add_kparams`.

The reason why I didn't add error check here is that function here now only receives `params` from `KernelParam.to_string()`, where there is some error check.
But I think it would be safer to add an extra error check here.

src/libs/kata-types/src/config/mod.rs
Ji-Xinyou: I am totally agree with this. I hardcoded "1026" here is because this is the behavior in go runtime, but I think this should be a **constant or a optional config** somewhere. I will fixed this first by changing this to a constant.

src/libs/kata-types/src/config/hypervisor/mod.rs
Ji-Xinyou: Fixed

src/libs/kata-types/src/config/mod.rs
Ji-Xinyou: Fixed

Ji-Xinyou:
@cmaf Thanks for the review. The problem you mentioned is solved in the second commit. 

After changing the raw strings to constant, we may want to let `src/agent` to use the constants in `src/libs` for consistency reason. But I think that should be a separate PR :)

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -237,6 +239,18 @@ impl BootInfo {
         Ok(())
     }
 
+    /// Add kernel parameters to bootinfo. It is always added before the original
+    /// to let the original one takes priority
+    pub fn add_kparams(&mut self, params: Vec<String>) {
```
liubin, line 244: Suggestion changing `add_kparams` to `add_kernel_params`, `k` is too short as an abbreviation.

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -237,6 +239,18 @@ impl BootInfo {
...
+        let mut p = params;
+        if self.kernel_params.is_empty() {
```
liubin, line 246: Is this ok?

```rust
        if !self.kernel_params.is_empty() {
            p.push(self.kernel_params.clone()); // [new_params0, new_params1, ..., original_params]
        }
        self.kernel_params = p.join(KERNEL_PARAM_DELIMITER);
```

src/libs/kata-types/src/config/mod.rs
```diff
@@ -151,7 +170,32 @@ impl TomlConfig {
         Ok(())
     }
 
-    ///  Probe configuration file according to the default configuration file list.
+    /// Get agent-specfic kernel parameters for further Hypervisor config revision
+    pub fn get_agent_kparams(&self) -> Result<HashMap<String, String>> {
```
liubin, line 174: `get_agent_kparams` -> `get_agent_kernel_params`

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
```diff
@@ -28,6 +29,18 @@ impl Param {
             value: value.to_owned(),
         }
     }
+
+    pub fn to_string(&self) -> Result<String> {
+        if self.key.is_empty() && self.value.is_empty() {
+            Err(anyhow!("Empty key and value"))
+        } else if self.key.is_empty() {
+            Err(anyhow!("Empty key"))
+        } else if self.value.is_empty() {
```
liubin, line 38: Is an empty value key-value valid? For example `a=b ccc`, `ccc` is an option without an value.

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -332,3 +334,20 @@ fn load_config(spec: &oci::Spec) -> Result<TomlConfig> {
     info!(sl!(), "get config content {:?}", &toml_config);
     Ok(toml_config)
 }
+
+// this update the agent-specfic kernel parameters into hypervisor's bootinfo
+// the agent inside the VM will read from file cmdline to get the params and function
+fn update_agent_kparams(config: &mut TomlConfig) -> Result<()> {
```
liubin, line 340: `update_agent_kparams` -> `update_agent_kernel_params`

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
Ji-Xinyou: Yes, an value w/o a key is valid.
e.g. `agent.debug_console` is a standalone flag.

src/libs/kata-types/src/config/hypervisor/mod.rs
Ji-Xinyou: Fixed

src/libs/kata-types/src/config/hypervisor/mod.rs
Ji-Xinyou: Thanks @liubin , fixed

src/libs/kata-types/src/config/mod.rs
Ji-Xinyou: Fixed

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: Fixed

Ji-Xinyou: Fixed

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
Ji-Xinyou: A key w/o a value is valid. It works as a standalone flag passed to kernel parameters. e.g. `agent.debug_console` is one of them.

liubin (approved):

Tim-0731-Hzt (approved):

Issue #5100, opened 2022-09-05
libs/kata-types: mistake of get_sandbox_cpu_period and get_sandbox_cpu_quota


liubin:
In libs/kata-types/src/annotations/mod.rs, the data type of period and quota are exchanged.

https://github.com/kata-containers/kata-containers/blob/ba013c5d0f31d0e585bce15cfa898003c0270af3/src/libs/kata-types/src/annotations/mod.rs#L382-L395

The right types are:

- period: u64
- quota: i64

PR #5101, opened 2022-09-05
kata-types: change return type of getting CPU period/quota function

liubin:
period should have a type of u64, and quota should be i64, the
function of getting CPU period and quota from annotations should
use the same data type as function return type.

Fixes: #5100

### DIFF
--- a/src/runtime-rs/crates/runtimes/src/static_resource.rs
+++ b/src/runtime-rs/crates/runtimes/src/static_resource.rs
@@ -128,8 +128,8 @@ fn get_sizing_info(annotation: Annotation) -> Result<(u64, i64, i64)> {
     // since we are *adding* our result to the config, a value of 0 will cause no change
     // and if the annotation is not assigned (but static resource management is), we will
     // log a *warning* to fill that with zero value
-    let period = annotation.get_sandbox_cpu_quota();
-    let quota = annotation.get_sandbox_cpu_period();
+    let period = annotation.get_sandbox_cpu_period();
+    let quota = annotation.get_sandbox_cpu_quota();
     let memory = annotation.get_sandbox_mem();
     Ok((period, quota, memory))
 }
--- a/src/libs/kata-types/src/annotations/mod.rs
+++ b/src/libs/kata-types/src/annotations/mod.rs
@@ -379,17 +379,17 @@ impl Annotation {
     }
 
     /// Get the annotation of cpu quota for sandbox
-    pub fn get_sandbox_cpu_quota(&self) -> u64 {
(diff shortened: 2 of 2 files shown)


fidencio (approved):
lgtm, thanks @liubin!

src/runtime-rs/crates/runtimes/src/static_resource.rs
```diff
@@ -128,40 +128,186 @@ fn get_sizing_info(annotation: Annotation) -> Result<(u64, i64, i64)> {
...
-        let cpu = oci::LinuxCpu {
-            period: Some(period),
-            quota: Some(quota),
-            ..Default::default()
-        };
-        if let Ok(cpu_resource) = LinuxContainerCpuResources::try_from(&cpu) {
-            if let Some(v) = cpu_resource.get_vcpus() {
-                assert_eq!(v, 3);
-            }
+    fn test_static_resource_mgmt_sandbox() {
```
c3d, line 144: These tests are what had you discovered the #5100 bug, I suppose.

They should go in different commits, as it's not directly related to #5100. This would allow a later git blame to correctly report what the code is about.

src/runtime-rs/crates/runtimes/src/static_resource.rs
```diff
@@ -128,40 +128,186 @@ fn get_sizing_info(annotation: Annotation) -> Result<(u64, i64, i64)> {
...
+        struct InputData<'a> {
```
c3d, line 145: It looks like `InputData` and `TestData` could be shared between the sandbox and container tests. using generic adjustments on the type used to pass data. Not sure it's worth it.

c3d (changes requested):
I would prefer to split the new tests from the bug fix.

The new tests look fine and useful.

src/runtime-rs/crates/runtimes/src/static_resource.rs
liubin: Yes, I'd like to split them.

src/runtime-rs/crates/runtimes/src/static_resource.rs
liubin: I'd like to give a try, but if there are seldom use-cases, that would not save any code I think.

liubin:
@c3d The CI is for PR unit, it's not easy to find broken between many commits so that I will split the original PR into two:

- one is this, correct the data type
- and a next PR will add the tests later

cmaf (approved):

amshinde (approved):

liubin:
@c3d PTAL

Issue #5104, opened 2022-09-05
"inline-virtio-fs" should not check "virtio_fs_daemon"


liubin:
If `shared_fs` is set to "inline-virtio-fs", the `virtio_fs_daemon` should be ignored, but not to check if it is empty or exist.

PR #5105, opened 2022-09-05
kata-types: don't check virtio_fs_daemon for inline-virtio-fs

liubin:
If the shared_fs is set to "inline-virtio-fs", the "virtio_fs_daemon"
should be ignored.

Fixes: #5104


fidencio (approved):
lgtm, thanks @liubin!

studychao (approved):
Thanks!

GabyCT (approved):

PR #5106, opened 2022-09-05
microvm: Remove kernel_irqchip=on option

### LINKED ISSUE #1984: microvm: legacy mode for virtio-mmio doesn't work.
# Description of problem

One of the devices seems to not come up and the kata agent can't initialize properly.
Turning off legacy mode (`-global virtio-mmio.force-legacy=false`) fixes this.

norbjd:
Closes #1984 and #4386

With microvm machine types, `kernel_irqchip=split` option is necessary (see related issues for more debug info).

### DIFF
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -38,7 +38,7 @@ const (
 
 	defaultQemuMachineType = QemuQ35
 
-	defaultQemuMachineOptions = "accel=kvm,kernel_irqchip=on"
+	defaultQemuMachineOptions = "accel=kvm"
 
 	splitIrqChipMachineOptions = "accel=kvm,kernel_irqchip=split"
 


fidencio (approved):
lgtm, thanks @norbjd!

wainersm:
It would be nice to have a tests and/or CI job to test microvm so to avoid regressions.

The change looks good to me.
FYI @gkurz

amshinde (approved):
lgtm. Would be good to have a test/CI to verify this.

gkurz:
> It would be nice to have a tests and/or CI job to test microvm so to avoid regressions.
> 
> The change looks good to me. FYI @gkurz

@wainersm, thanks for summoning me !

Split irqchip is already the default behaviour for the Microvm machine in QEMU. It is thus quite unfortunate to hardcode this in kata as well.

This seems to be the consequence of `kernel_irqchip=on` being in `defaultQemuMachineOptions`. Why is it the case ? If there are no valid reasons, I'd rather drop that instead of adding yet another dependency on QEMU internals.

@norbjd ? @fidencio ?

fidencio:
@gkurz, this is the commit that introduce the `kernel_irqchip=on`:

```
commit 316509566966e4c9b3fd9ba3521554b384fdbf88
Refs: 2.2.0-alpha1-133-g316509566
Author:     David Gibson <david@gibson.dropbear.id.au>
AuthorDate: Wed Jul 28 13:44:02 2021 +1000
Commit:     David Gibson <david@gibson.dropbear.id.au>
CommitDate: Wed Aug 4 14:34:11 2021 +1000

    runtime/qemu: Use explicit "on" for kernel_irqchip parameter

    Kata uses the 'kernel_irqchip' machine option to qemu.  By default it
    uses it in what qemu calls the "short-form boolean" with no parameter.
    That style was deprecated by qemu between 5.2 and 6.0 (commit
    ccd3b3b8112b) and effectively removed entirely between 6.0 and 6.1
    (commit d8fb7d0969d5).

    Update ourselves for newer qemus by using an explicit
    "kernel_irqchip=on".

---
 src/runtime/virtcontainers/qemu_amd64.go | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/runtime/virtcontainers/qemu_amd64.go b/src/runtime/virtcontainers/qemu_amd64.go
index d19bab5c5..1129d4d34 100644
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -31,7 +31,7 @@ const (

        defaultQemuMachineType = QemuQ35

-       defaultQemuMachineOptions = "accel=kvm,kernel_irqchip"
+       defaultQemuMachineOptions = "accel=kvm,kernel_irqchip=on"

        qmpMigrationWaitTimeout = 5 * time.Second

```

fidencio:
> @gkurz, this is the commit that introduce the `kernel_irqchip=on`:
> 
> ```
> commit 316509566966e4c9b3fd9ba3521554b384fdbf88
> Refs: 2.2.0-alpha1-133-g316509566
> Author:     David Gibson <david@gibson.dropbear.id.au>
> AuthorDate: Wed Jul 28 13:44:02 2021 +1000
> Commit:     David Gibson <david@gibson.dropbear.id.au>
> CommitDate: Wed Aug 4 14:34:11 2021 +1000
> 
>     runtime/qemu: Use explicit "on" for kernel_irqchip parameter
> 
>     Kata uses the 'kernel_irqchip' machine option to qemu.  By default it
>     uses it in what qemu calls the "short-form boolean" with no parameter.
>     That style was deprecated by qemu between 5.2 and 6.0 (commit
>     ccd3b3b8112b) and effectively removed entirely between 6.0 and 6.1
>     (commit d8fb7d0969d5).
> 
>     Update ourselves for newer qemus by using an explicit
>     "kernel_irqchip=on".
> 
>     Signed-off-by: David Gibson <david@gibson.dropbear.id.au>
> ---
>  src/runtime/virtcontainers/qemu_amd64.go | 2 +-
>  1 file changed, 1 insertion(+), 1 deletion(-)
> 
> diff --git a/src/runtime/virtcontainers/qemu_amd64.go b/src/runtime/virtcontainers/qemu_amd64.go
> index d19bab5c5..1129d4d34 100644
> --- a/src/runtime/virtcontainers/qemu_amd64.go
> +++ b/src/runtime/virtcontainers/qemu_amd64.go
> @@ -31,7 +31,7 @@ const (
> 
>         defaultQemuMachineType = QemuQ35
> 
> -       defaultQemuMachineOptions = "accel=kvm,kernel_irqchip"
> +       defaultQemuMachineOptions = "accel=kvm,kernel_irqchip=on"
> 
>         qmpMigrationWaitTimeout = 5 * time.Second
> ```

That's actually just adjusting the format.  I couldn't find why it's been added in the first place, and I'd recommend removing it and running our tests.  Maybe we could drop it entirely.

gkurz:
> 
> That's actually just adjusting the format. I couldn't find why it's been added in the first place,

Yeah I've blamed this up to the initial commit:

```
commit 24eff72d82dfc444ea0dc15977cfa8c6b9c67959
Author: Samuel Ortiz <sameo@linux.intel.com>
Date:   Fri Feb 23 10:35:08 2018 +0100

    virtcontainers: Initial import
```

>and I'd recommend removing it and running our tests. Maybe we could drop it entirely.

Makes sense as `kernel-irqchip=on` is the default behavior since QEMU 5.0 (commit [11bc4a13d1f4b](https://github.com/qemu/qemu/blob/11bc4a13d1f4b/qemu-options.hx#L131).) In the meantime, I'll try to dig some more in the virtcontainers history.

gkurz:
> In the meantime, I'll try to dig some more in the virtcontainers history.

Blamed up to the initial [commit](https://github.com/containers/virtcontainers/commit/8fbd5de595dd8e9eaa9c0f93f771a5ede06e485d) as well without finding any clue.

@norbjd can you please give a try at @fidencio's suggestion ?

norbjd:
@gkurz Sorry about the late answer, I'll try to remove the option and see if it still works :smile:

norbjd:
It indeed works without `kernel_irqchip=on` on `q35` and `microvm` machines.

But I don't know if this create regression, since I don't understand why `kernel_irqchip=on` was set on the first place :thinking:

If removing this option create issues, I hope that the tests can catch them.

fidencio:
I've fixed the commit message, squashed both commits, and force-pushed into @norbjd's repo in order to avoid bothering the contributor to do so.

I'm rerunning the tests now.

fidencio:
/retest-arm

The failure is unrelated to this PR.

gkurz:
> It indeed works without `kernel_irqchip=on` on `q35` and `microvm` machines.
> 
> But I don't know if this create regression, since I don't understand why `kernel_irqchip=on` was set on the first place thinking
> 

Hi @norbjd ! Sorry for the delay, I overlooked the message from GH...

As I've mentioned earlier, this setting has been around since the beginning without any documentation. I believe that using the in-kernel irqchip wasn't the default behavior with older versions of QEMU and users had to explicitly ask for it. Since this is now the default since QEMU 5.0 which is the older version that we support (correct me if I'm wrong @fidencio), it is safe to simply get rid of it.


> If removing this option create issues, I hope that the tests can catch them.

gkurz (approved):
That's what we had agreed upon. Glad to see that this flies with both `q35` and `microvm`.

Thanks @norbjd and @fidencio !

Issue #5110, opened 2022-09-06
"default_maxvcpus = 0" and "default_vcpus = 1" will not work.


liubin:
`configuration.toml` with this setting:

```
default_maxvcpus = 0
default_vcpus = 1
```

Will lead to error:

> ctr: failed to create shim task: Others("failed to handler message try init runtime instance\n\nCaused by:\n    0: init runtime handler\n    1: start sandbox\n    2: start vm\n    3: set vm base config\n    4: set vm configuration\n    5: Failed to set vm configuration VmConfigInfo { vcpu_count: 0, max_vcpu_count: 0, cpu_pm: \"on\", cpu_topology: CpuTopology { threads_per_core: 1, cores_per_die: 1, dies_per_socket: 1, sockets: 1 }, vpmu_feature: 0, mem_type: \"shmem\", mem_file_path: \"\", mem_size_mib: 1024, serial_path: Some(\"/run/kata/c1/console.sock\") }\n    6: vmm action error: MachineConfig(InvalidVcpuCount(0))"): unknown

The log:

> cpu_info: CpuInfo { cpu_features: \"pmu=off\", default_vcpus: 0, default_maxvcpus: 0 }

The reason:

https://github.com/kata-containers/kata-containers/blob/cb7f9524be3247ca08cdbbce0ada2d493afcc7ef/src/libs/kata-types/src/config/hypervisor/dragonball.rs#L64-L70

When `default_maxvcpus` == 0, the `default_vcpus` is greater than 0, and will be set to 0 as `default_maxvcpus`

PR #5111, opened 2022-09-06
libs/kata-types: adjust default_vcpus correctly

liubin:
With default_maxvcpus = 0 and default_vcpus = 1 settings, the
default_vcpus will be set to 0 and leads to starting failure.

The default_maxvcpus is not set correctly when it is set to 0,
and the default_vcpus will be set to 0.

The correct action is setting default_maxvcpus to the max number
of CPUs or MAX_DRAGONBALL_VCPUS, and the default_vcpus should be
set to the desired value if the valuse is between 0 and
default_maxvcpus.

Fixes: #5110

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/mod.rs
+++ b/src/libs/kata-types/src/config/hypervisor/mod.rs
@@ -301,34 +301,38 @@ impl CpuInfo {
     pub fn adjust_config(&mut self) -> Result<()> {
         let features: Vec<&str> = self.cpu_features.split(',').map(|v| v.trim()).collect();
         self.cpu_features = features.join(",");
+
+        let cpus = num_cpus::get() as u32;
+
+        // adjust default_maxvcpus
+        if self.default_maxvcpus == 0 || self.default_maxvcpus > cpus {
+            self.default_maxvcpus = cpus;
+        }
+
+        // adjust default_vcpus
+        if self.default_vcpus < 0 || self.default_vcpus as u32 > cpus {
+            self.default_vcpus = cpus as i32;
+        } else if self.default_vcpus == 0 {
+            self.default_vcpus = default::DEFAULT_GUEST_VCPUS as i32;
+        }
+
+        if self.default_vcpus > self.default_maxvcpus as i32 {
+            self.default_vcpus = self.default_maxvcpus as i32;
+        }
+
         Ok(())
     }
 
     /// Validate the configuration information.
     pub fn validate(&self) -> Result<()> {
+        if self.default_vcpus > self.default_maxvcpus as i32 {
+            return Err(eother!(
+                "The default_vcpus({}) is greater than default_maxvcpus({})",
+                self.default_vcpus,
+                self.default_maxvcpus
+            ));
+        }
         Ok(())
-    }
-
-    /// Get default number of guest vCPUs.
-    pub fn get_default_vcpus(&self) -> u32 {
-        let cpus = num_cpus::get() as u32;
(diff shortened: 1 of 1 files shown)


liubin:
@studychao @quanweiZhou @Tim-0731-Hzt PATL.
Thanks.

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -289,34 +289,38 @@ impl CpuInfo {
...
-    pub fn validate(&self) -> Result<()> {
-        Ok(())
-    }
 
-    /// Get default number of guest vCPUs.
-    pub fn get_default_vcpus(&self) -> u32 {
         let cpus = num_cpus::get() as u32;
+
+        // adjust default_maxvcpus
+        if self.default_maxvcpus == 0 || self.default_maxvcpus > cpus {
```
studychao, line 296: What if the cpus we get from the host is larger than the vcpu number that the VMM can support (e.g Dragonball could only support up to 254 cpus and some AMD MILAN chip could have 256 cpus)

liubin: This will handle your case I think:

https://github.com/kata-containers/kata-containers/blob/cb7f9524be3247ca08cdbbce0ada2d493afcc7ef/src/libs/kata-types/src/config/hypervisor/dragonball.rs#L64-L66

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -289,34 +289,38 @@ impl CpuInfo {
...
-    /// Get default maximal number of guest vCPUs.
-    pub fn get_default_max_vcpus(&self) -> u32 {
-        let cpus = num_cpus::get() as u32;
-        if self.default_maxvcpus == 0 || self.default_maxvcpus > cpus {
-            cpus
-        } else {
-            self.default_maxvcpus
+    /// Validate the configuration information.
+    pub fn validate(&self) -> Result<()> {
+        if self.default_vcpus > self.default_maxvcpus as i32 {
```
quanweiZhou, line 316: What if `self.default_vcpus == 0` or `self.default_maxvcpus ==0`.

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -289,34 +289,38 @@ impl CpuInfo {
...
 
-    /// Get default maximal number of guest vCPUs.
-    pub fn get_default_max_vcpus(&self) -> u32 {
-        let cpus = num_cpus::get() as u32;
-        if self.default_maxvcpus == 0 || self.default_maxvcpus > cpus {
-            cpus
-        } else {
-            self.default_maxvcpus
+    /// Validate the configuration information.
+    pub fn validate(&self) -> Result<()> {
```
quanweiZhou, line 315: I think need to add `validate()` in UT.

quanweiZhou (commented):
Thanks, @liubin, few comments.

src/libs/kata-types/src/config/hypervisor/mod.rs
liubin: If the call sequence is correct, there will be no such case, neither of them will be zero. Even the validate() should be reverted, because `default_maxvcpus` is ensured to be greater or equal to `default_vcpus`.

liubin:
@quanweiZhou @studychao @Tim-0731-Hzt ptal

quanweiZhou (approved):
Thanks, @liubin LGTM.

Tim-0731-Hzt (approved):
Thanks! @liubin

Issue #5112, opened 2022-09-06
TomlConfig's validate function is not called


liubin:
`TomlConfig` has a function `validate()`, but it has never been called.

It should be called after loaded from file and adjusted by annotations.

PR #5113, opened 2022-09-06
runtime-rs: call TomlConfig's validate function after load

liubin:
Call TomlConfig's validate function after it is loaded and
adjusted by annotations.

Fixes: #5112


cmaf (approved):
thanks @liubin

Tim-0731-Hzt (approved):
Thanks! @liubin

PR #5120, opened 2022-09-06
CC, packaging: add sev-initrd target, and add efi_secret module to sev's initrd 

Alex-Carter01:
Adds a make target for an sev initrd.

Changes the kata-deploy-binaries functions to be more generic.

Adds efi_secret kernel module to the sev initrd. 

Moved kernel config version checking function from kernel builder to lib script.

First question: should this get added to the cc target?

### DIFF
--- a/tools/packaging/guest-image/build_image.sh
+++ b/tools/packaging/guest-image/build_image.sh
@@ -17,6 +17,9 @@ readonly packaging_root_dir="$(cd "${script_dir}/../" && pwd)"
 readonly repo_root_dir="$(cd "${script_dir}/../../../" && pwd)"
 readonly osbuilder_dir="$(cd "${repo_root_dir}/tools/osbuilder" && pwd)"
 
+patches_path=""
+readonly default_patches_dir="${packaging_root_dir}/kernel/patches"
+
 export GOPATH=${GOPATH:-${HOME}/go}
 source "${packaging_root_dir}/scripts/lib.sh"
 
@@ -32,7 +35,17 @@ build_initrd() {
 	export USE_DOCKER=1
 	export AGENT_INIT="yes"
 	# ROOTFS_BUILD_DEST is a Make variable
-	sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}"
+
+	if [ -z "${AA_KBC}" == "offline_sev_kbc" ]; then
+		config_version=$(get_config_version)
+		kernel_version="$(get_from_kata_deps "assets.kernel.sev.version")"
+		kernel_version=${kernel_version#v}
+		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-sev-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
+		sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}" KERNEL_MODULES_DIR="${module_dir}"
+	else
+		sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}"
+	fi
+
 	if [ -n "${INCLUDE_ROOTFS:-}" ]; then
 		sudo cp -RL --preserve=mode "${INCLUDE_ROOTFS}/." "${rootfs_build_dest}/${initrd_distro}_rootfs/"
 	fi
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -89,7 +89,8 @@ cc: cc-cloud-hypervisor-tarball \
 	cc-tdx-qemu-tarball \
 	cc-tdx-td-shim-tarball \
 	cc-tdx-tdvf-tarball \
-	cc-sev-ovmf-tarball
+	cc-sev-ovmf-tarball \
+	cc-sev-rootfs-initrd-tarball
 
 cc-cloud-hypervisor-tarball:
 	${MAKE} $@-build
@@ -101,6 +102,9 @@ cc-qemu-tarball:
 	${MAKE} $@-build
 
 cc-rootfs-image-tarball:
+	${MAKE} $@-build
+
+cc-sev-rootfs-initrd-tarball: cc-sev-kernel-tarball
 	${MAKE} $@-build
 
 cc-shim-v2-tarball:
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -90,6 +90,7 @@ options:
 	cc-qemu
 	cc-tdx-qemu
 	cc-rootfs-image
+	cc-sev-initrd-image
 	cc-shimv2
 	cc-virtiofsd
 	cc-sev-ovmf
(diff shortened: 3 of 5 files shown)


fidencio:
> First question: should this get added to the cc target?

Yes, it should! :-)

tools/packaging/guest-image/build_image.sh
```diff
@@ -32,7 +35,17 @@ build_initrd() {
 	export USE_DOCKER=1
 	export AGENT_INIT="yes"
 	# ROOTFS_BUILD_DEST is a Make variable
-	sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}"
+
+	if [ -z "${AA_KBC}" == "offline_sev_kbc" ]; then
```
fitzthum, line 39: At some point soonish we will want to add support for `online_sev_kbc`. Not for the first release, tho.

fitzthum (approved):
LGTM. 

Currently the only thing that delineates the SEV rootfs is that it is an initrd rather than an image? I wonder if we should label it more explicitly? Probably fine for now.

fidencio:
One thing that I commented on your other PRs and apply here as well, is that we want to make sure that we cleanup all the artefacts we install, here: https://github.com/kata-containers/kata-containers/blob/3c7fe939975ad95843979f4fa4019cf2191cfe43/tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh#L198

Alex-Carter01:
> One thing that I commented on your other PRs and apply here as well, is that we want to make sure that we cleanup all the artefacts we install, here:
> 
> https://github.com/kata-containers/kata-containers/blob/3c7fe939975ad95843979f4fa4019cf2191cfe43/tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh#L198

Ope! I had meant to comment that all artifacts end up in `/opt/confidential-containers/share/kata-containers/`, a location that is covered.

tools/packaging/guest-image/build_image.sh
Alex-Carter01: I'll open an issue regarding the online kbc to make sure I keep track of it.

Alex-Carter01:
> LGTM.
> 
> Currently the only thing that delineates the SEV rootfs is that it is an initrd rather than an image? I wonder if we should label it more explicitly? Probably fine for now.

vs the generic cc image, the fact that it is an initrd, plus passing the image builder the right AA KBC. But yes, very high similarity.

c3d (approved):
Looks good to me.

wainersm:
@Alex-Carter01 please remove 'CC |' on subject of the commits so to silence the commit checker.

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -97,6 +97,9 @@ cc-qemu-tarball:
 cc-rootfs-image-tarball:
 	${MAKE} $@-build
 
+cc-sev-initrd-image-tarball: cc-sev-kernel-tarball
```
wainersm, line 100: Sorry I am being picky here. That target name is not following the "standard". We got:

rootfs-image-tarball
rootfs-initrd-tarball
cc-rootfs-image-tarball

Then I would expect this to be `cc-rootfs-initrd-sev-tarball`

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -112,13 +113,26 @@ install_cc_clh() {
 
 #Install cc capable guest image
 install_cc_image() {
+	export AA_KBC="${1}"
```
wainersm, line 116: Suggestion. Fallback to defaults if $1 and/ $2 aren't passed, this way you don't need to create the `install_cc_generic_image`.

```
export AA_KBC="${1:-offline_fs_kbc}"
image_type="${2:-image}"
```

tools/packaging/guest-image/build_image.sh
wainersm, line 39: @Alex-Carter01 most (if not all) script we set `set -o nounset`. Here if I run the script without exporting `AA_KBC` then it will fail. You must initialize the variable:

`if [ -z "${AA_KBC:-}" == "offline_sev_kbc" ]; then`

tools/packaging/guest-image/build_image.sh
```diff
@@ -32,7 +35,17 @@ build_initrd() {
...
+		config_version=$(get_config_version)
+		kernel_version="$(get_from_kata_deps "assets.kernel.sev.version")"
+		kernel_version=${kernel_version#v}
+		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-sev-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
```
wainersm, line 43: If you use `packaging_root_dir` variable you will save some chars.

tools/packaging/guest-image/build_image.sh
wainersm: Actually I've a question about this change... the `build_image.sh` is more like a wrapper for `make rootfs` which in the end is a call to `tools/osbuilder/rootfs-builder/rootfs.sh`. Is there any reason to introduce this platform-specific code here and not in `tools/osbuilder/rootfs-builder/rootfs.sh` (or somewhere on `tools/osbuilder`)?

Here you could export AA_KBC on the make call `sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest} ${AA_KBC:+AA_KBC=$AA_KBC}"`

ps: `${AA_KBC:+AA_KBC=$AA_KBC}` is a little bit crazy, there must be a better way.

tools/packaging/scripts/lib.sh
```diff
@@ -97,3 +97,19 @@ get_kata_hash() {
 	ref=$2
 	git ls-remote --heads --tags "https://github.com/${project}/${repo}.git" | grep "${ref}" | awk '{print $1}'
 }
+
+get_config_and_patches() {
```
wainersm, line 101: Those functions in `lib.sh` will be shared with other scripts? If yes, we need to document because here they loose context. Would make sense to pass `default_patches_dir` as parameter rather than assume it was exported?

Alex-Carter01:
Fixed commit messages, hopefully passes commit checker.

Made changes to the first commit, have not moved to second yet.

Changed naming of make target

Added make target to the cc bundle

removed generic function and replaced with default variables.

tools/packaging/guest-image/build_image.sh
Alex-Carter01: This is definitely a good discussion to have, so I want to follow up here before making the tweaks to build_image.

I had originally gone back and further between having the logic for locating the efi_secret module in one of three places: kata-deploy-binaries.sh, build_image.sh, and the rootfs.sh.

My thinking had been to keep any platform specific logic out of the rootfs.sh script, and use its KERNEL_MODULES_DIR parameter. Currently the only change I am making to the rootfs.sh script for sev is generic.

The tricky piece of logic in getting the location of the kernel module is using getting the kernel config version. It ends up using the functions which were previously in the kernel builder, but are ultimately reading the kernel config version file. I viewed this as analogous to reading the broader kata version file, which is already being done in the build_image script e.g. to get the distro and os_version.

I am definitely willing to move things around if this belongs better elsewhere.

stevenhorsman:
@Alex-Carter01 - the header commit message for `074eb44` is causing some problems (length and lack of subsystem), so can you take a look at that please.

Alex-Carter01:
Fixing commit message title

wainersm:
@kata-containers/intel hi, there is this jenkins-ci-tdx-x86_64-cc_cri_containerd_k8s-cloud-hypervisor  job that runs behind the firewall and we don't have access to the logs. Does someone in Intel could check if this PR has broken the job or possible it is an unrelated failure?

wainersm:
@Alex-Carter01 one more conflict to fix. I just merged another of your PRs so maybe that's what caused that conflict.

PR #4770, opened 2022-07-29
agent: Use rtnetlink's neighbours API to add neighbors

justxuewei:
Bump rtnetlink version from 0.8.0 to 0.11.0. Use rtnetlinks's API to
add neighbors and fix issues to adapt new verson of rtnetlink.

Fixes: #4607

Signed-off-by: Xuewei Niu <justxuewei@apache.org>

PR #5124, opened 2022-09-06
Revert arp neighbour api

amshinde:
This reverts PR: https://github.com/kata-containers/kata-containers/pull/4770

Using  rtnetlink's neighbours API to add neighbors is causing issues in AKS while adding neighbours. Hence reverting the change for now till we resolve the issue.

### DIFF
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -35,7 +35,7 @@ tokio = { version = "1.14.0", features = ["full"] }
 tokio-vsock = "0.3.1"
 
 netlink-sys = { version = "0.7.0", features = ["tokio_socket",]}
-rtnetlink = "0.11.0"
+rtnetlink = "0.8.0"
 netlink-packet-utils = "0.4.1"
 ipnetwork = "0.17.0"
 
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -654,8 +654,8 @@ dependencies = [
  "libc",
  "log",
  "logging",
- "netlink-packet-utils 0.4.1",
- "netlink-sys 0.7.0",
+ "netlink-packet-utils",
+ "netlink-sys",
  "nix 0.24.2",
  "oci",
  "opentelemetry",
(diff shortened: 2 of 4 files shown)


src/agent/Cargo.toml
```diff
@@ -35,7 +35,7 @@ tokio = { version = "1.14.0", features = ["full"] }
 tokio-vsock = "0.3.1"
 
 netlink-sys = { version = "0.7.0", features = ["tokio_socket",]}
-rtnetlink = "0.11.0"
+rtnetlink = "0.8.0"
```
fidencio, line 38: Do we have to revert this?

fidencio (commented):
lgtm, apart from reverting rtnetlink to an older version.

And I see that `test_kata_deploy` successfully passed with this PR.

src/agent/Cargo.toml
amshinde: The commit that I reverted is this: https://github.com/kata-containers/kata-containers/pull/4770/commits/845c1c03cfbe8435ac383964ef243da252726351

It upgraded rtnetlink to 0.11.0 and used the new API for adding neighbours. Hence the revert includes going back to previous version. We can keep the newer version while still using the older API or upgrade to newer version of rtnetlink when use of the new API is added back again. I am ok with either. Let me know.

fidencio: If the newer version works fine with the older API, I'd go for that.

amshinde: @fidencio Doesnt seem to work.
```
error[E0599]: no method named `set_name_filter` found for struct `LinkGetRequest` in the current scope
   --> src/netlink.rs:167:47
    |
167 |             LinkFilter::Name(name) => request.set_name_filter(name.to_owned()),
    |                                               ^^^^^^^^^^^^^^^ method not found in `LinkGetRequest
```

cmaf (approved):
thanks @amshinde

dborquez (approved):
lgtm, thanks @amshinde

Issue #5135, opened 2022-09-07
CC | Generate initramfs for CoCo measured boot


arronwy:
For [Measured Boot Image proposal for CoCo](https://github.com/confidential-containers/documentation/issues/40#top), we need an initramfs to measure the rootfs with different verity scheme like `dm-verity`.

We need static build `veritysetup` and init.sh to setup and mount the root device before use.

PR #5136, opened 2022-09-07
initramfs: Add build script to generate initramfs

arronwy:
For https://github.com/confidential-containers/documentation/issues/40#top, we need an initramfs to measure the rootfs with different verity scheme like `dm-verity`.

The init.sh in initramfs will parse the verity scheme, roothash, root device and setup the root device accordingly.
All cryptsetup related binaries are build staticly.

Fixes: #5135

### DIFF
--- a/tools/packaging/static-build/initramfs/Dockerfile
+++ b/tools/packaging/static-build/initramfs/Dockerfile
@@ -1,0 +1,38 @@
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+from ubuntu:20.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+ENV TZ=UTC
+RUN apt-get update &&\
+    apt-get --no-install-recommends install -y software-properties-common &&\
+    add-apt-repository ppa:git-core/ppa -y &&\
+    apt-get update && apt-get upgrade -y && \
+    apt-get --no-install-recommends install -y \
+	    apt-utils \
+	    asciidoctor \
+	    autoconf \
+            autopoint \
+	    automake \
+	    busybox-static \
+	    ca-certificates \
+	    curl \
+	    gcc \
+	    gettext \
+	    git \
+	    libaio-dev \
+	    libblkid-dev \
+	    libselinux1-dev \
+	    libtool \
+            libpopt-dev \
+            libjson-c-dev \
+	    libssl-dev \
+	    make \
+	    ninja-build \
+	    pkg-config \
+	    uuid-dev \
+	    libseccomp-dev \
+	    libseccomp2 \
+	    zlib1g-dev &&\
+    apt-get clean && rm -rf /var/lib/apt/lists/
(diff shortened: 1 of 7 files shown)


tools/packaging/static-build/initramfs/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+from ubuntu:20.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+ENV TZ=America/New_York
```
fidencio, line 7: Out of curiosity, why do we need this?

tools/packaging/static-build/initramfs/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN apt-get update &&\
+    apt-get --no-install-recommends install -y software-properties-common &&\
+    add-apt-repository ppa:git-core/ppa -y &&\
```
fidencio, line 10: You shouldn't need a ppa to install git.  Or do you need it for something else?

fidencio (commented):
lgtm, but I left a few comments, just out of curiosity.

And not related to this patch, but I start to think that we should start having the build containers stored in quay.io, with all the machinery needed inside those, and we just pull them and do the artefact generation, rather than building the container every single time.  But that's material for a different issue / different PR (cc @stevenhorsman and @wainersm about this).

tools/packaging/static-build/initramfs/Dockerfile
arronwy: The git from default base ubuntu 20.04 repo don't support sparse checkout feature, during install `gen_init_cpio` tool, we will clone the kernel tree.

tools/packaging/static-build/initramfs/Dockerfile
arronwy: During the installation of some required packages, Ubuntu usually installs the tzdata package, it will hangs and waits for user input to continue the installation: https://stackoverflow.com/questions/67452096/docker-build-hangs-based-on-order-of-install

fidencio: Let's set it to UTC / GMT then.
Another possible way to go is simply remove tzdata as already done in the Kata Containers CI.

arronwy: I updated to use UTC, tzdata is defined in docker file but included in the package dependencies.

fidencio (approved):
lgtm, thanks @arronwy!

fidencio:
@arronwy, could you fix:
```
ERROR: Required SPDX license header check ('SPDX-License-Identifier: Apache-2.0') failed for the following files:

tools/packaging/static-build/initramfs/initramfs.list

make: *** [Makefile:39: static-checks] Error 1
```

That's the only missing part in order to get this one merged.

arronwy:
> @arronwy, could you fix:
> 
> ```
> ERROR: Required SPDX license header check ('SPDX-License-Identifier: Apache-2.0') failed for the following files:
> 
> tools/packaging/static-build/initramfs/initramfs.list
> 
> make: *** [Makefile:39: static-checks] Error 1
> ```
> 
> That's the only missing part in order to get this one merged.

Thanks @fidencio , fixed now :-)

Issue #5139, opened 2022-09-07
CC tee kernel failing based on name mismatch untiring


Alex-Carter01:
get_tee_kernel() checks for two naming standards for curl, but does not check when trying to unpack the tarball

	# Depending on where we're getting the terball from it may have a
        # different name, such as linux-${version}.tar.gz or simply
 	# ${version}.tar.gz.  Let's try both before failing.
 	curl --fail -OL "${kernel_url}/linux-${kernel_tarball}" || curl --fail -OL "${kernel_url}/${kernel_tarball}"

vs

       tar --strip-components=1 -xf ${kernel_tarball} -C ${kernel_path}

proposal: add same naming system for the untarring

PR #5140, opened 2022-09-07
CC add tarball naming option when untarring in get_tee_kernel

Alex-Carter01:
In the kernel builder for tee tarballs there are currently two ptions checked when trying to pull a tarball.

Also trying these two options (with or without 'linux-') when untarring.

Fixes: #5139


fidencio:
This is a dup of https://github.com/kata-containers/kata-containers/pull/5096

Issue #5148, opened 2022-09-09
CC | Integrate initramfs into Guest kernel for CoCo measured boot


arronwy:
For [Measured Boot Image proposal for CoCo](https://github.com/confidential-containers/documentation/issues/40#top), after we generate an initramfs with rootfs measurement capability, we need integrate this initramfs into the guest kernel as one binary, since guest kernel can be measured by the firmware.

PR #5149, opened 2022-09-09
kernel: Integrate initramfs into Guest kernel

arronwy:
Implement subtask of https://github.com/confidential-containers/documentation/issues/40

Integrate initramfs into guest kernel as one binary, which will be measured by the firmware together.

Fixes: #5148


fidencio (approved):
lgtm, thanks @arronwy!

Issue #4886, opened 2022-08-11
vmm process was incorrectly added to kata_sandbox cgroup in cgroup v2


yaoyinnan:
After https://github.com/kata-containers/kata-containers/pull/4397 was merged, Kata can run on systems using cgroup v2. When the system uses cgroup v2, the vmm process will be incorrectly added to `kata_sandbox`.

```
root@k8s-worker01:/opt/shell# ps -A
  34962 ?        00:00:00 containerd-shim
  34971 ?        00:00:00 virtiofsd
  34977 ?        00:00:03 qemu-system-x86
  34979 ?        00:00:00 virtiofsd
  34982 ?        00:00:00 vhost-34977
  34985 ?        00:00:00 kvm-pit/34977
  34986 ?        00:00:00 vhost-34977
  34997 pts/0    00:00:00 ps

root@k8s-worker01:/opt/shell# cat /proc/{34962,34977,34971,34979}/cgroup
0::/kata_overhead/84f329f0d35ce26872da8296b441f0474721f083e73e9c82decd971a6ce5f3da
0::/kubepods/besteffort/podf5199815-6dd7-4250-8f04-69b89527bad7/kata_84f329f0d35ce26872da8296b441f0474721f083e73e9c82decd971a6ce5f3da
0::/kata_overhead/84f329f0d35ce26872da8296b441f0474721f083e73e9c82decd971a6ce5f3da
0::/kata_overhead/84f329f0d35ce26872da8296b441f0474721f083e73e9c82decd971a6ce5f3da

root@k8s-worker01:/opt/shell# ps -eT | grep 34977
  34977   34977 ?        00:00:00 qemu-system-x86
  34977   34978 ?        00:00:00 qemu-system-x86
  34977   34983 ?        00:00:00 qemu-system-x86
  34977   34984 ?        00:00:27 qemu-system-x86
  34982   34982 ?        00:00:00 vhost-34977
  34985   34985 ?        00:00:00 kvm-pit/34977
  34986   34986 ?        00:00:00 vhost-34977

root@k8s-worker01:/opt/shell# cat /proc/{34977,34978,34983,34984}/cgroup
0::/kubepods/besteffort/podf5199815-6dd7-4250-8f04-69b89527bad7/kata_84f329f0d35ce26872da8296b441f0474721f083e73e9c82decd971a6ce5f3da
[...]

PR #5153, opened 2022-09-09
runtime: using threaded mode to manage cgroupv2

yaoyinnan:
In the current cgroupv2, because the process mode cannot separate processes and threads into different cgroups, the vmm process is incorrectly added to the sandbox cgroup. The sandbox and overhead are placed in the same cgroup through thread mode to realize that the vCPU thread is placed in the sandbox and other processes are placed in the overhead.

image

Fixes: #4886

### DIFF
--- a/src/runtime/pkg/resourcecontrol/cgroups.go
+++ b/src/runtime/pkg/resourcecontrol/cgroups.go
@@ -276,12 +276,7 @@ func (c *LinuxCgroup) Delete() error {
 	case cgroups.Cgroup:
 		return cg.Delete()
 	case *cgroupsv2.Manager:
-		if IsSystemdCgroup(c.ID()) {
-			if err := cg.DeleteSystemd(); err != nil {
-				return err
-			}
-		}
-		return cg.Delete()
+		return deleteCgroup(cg, filepath.Join(UnifiedMountpoint, c.path))
 	default:
 		return ErrCgroupMode
 	}
@@ -339,10 +334,10 @@ func (c *LinuxCgroup) Update(resources *specs.LinuxResources) error {
 	}
 }
 
-func (c *LinuxCgroup) MoveTo(path string) error {
+func (c *LinuxCgroup) MoveTo(path string, sandboxCgroupOnly bool) error {
 	switch cg := c.cgroup.(type) {
 	case cgroups.Cgroup:
-		cgHierarchy, cgPath, err := cgroupHierarchy(path)
+		cgHierarchy, cgPath, err := cgroupHierarchy(path, sandboxCgroupOnly)
 		if err != nil {
 			return err
 		}
@@ -352,11 +347,25 @@ func (c *LinuxCgroup) MoveTo(path string) error {
 		}
 		return cg.MoveTo(newCgroup)
 	case *cgroupsv2.Manager:
-		newCgroup, err := cgroupsv2.LoadManager(unifiedMountpoint, path)
-		if err != nil {
-			return err
+		var newCgroup *cgroupsv2.Manager
+		if IsSystemdCgroup(path) && sandboxCgroupOnly {
+			slice, unit, err := getSliceAndUnit(path)
+			if err != nil {
+				return err
+			}
+			cg, err := cgroupsv2.LoadSystemd(slice, unit)
+			if err != nil {
+				return err
+			}
+			newCgroup = cg
+		} else {
+			cg, err := cgroupsv2.LoadManager(UnifiedMountpoint, path)
+			if err != nil {
+				return err
+			}
+			newCgroup = cg
 		}
-		return cg.MoveTo(newCgroup)
+		return moveTo(cg, newCgroup)
 	default:
 		return ErrCgroupMode
 	}
--- a/src/runtime/pkg/resourcecontrol/cgroups.go
+++ b/src/runtime/pkg/resourcecontrol/cgroups.go
@@ -345,7 +345,10 @@ func (c *LinuxCgroup) MoveTo(path string, sandboxCgroupOnly bool) error {
 		if err != nil {
 			return err
 		}
-		return cg.MoveTo(newCgroup)
+		err = RetryOperation(func() error { return cg.MoveTo(newCgroup) })
+		if err != nil {
+			return err
+		}
 	case *cgroupsv2.Manager:
 		var newCgroup *cgroupsv2.Manager
 		if IsSystemdCgroup(path) && sandboxCgroupOnly {
@@ -365,10 +368,15 @@ func (c *LinuxCgroup) MoveTo(path string, sandboxCgroupOnly bool) error {
 			}
 			newCgroup = cg
 		}
-		return moveTo(cg, newCgroup)
+		err := RetryOperation(func() error { return moveTo(cg, newCgroup) })
+		if err != nil {
+			return err
+		}
 	default:
 		return ErrCgroupMode
 	}
+
+	return nil
 }
 
 func (c *LinuxCgroup) AddDevice(deviceHostPath string) error {
--- a/src/runtime/pkg/resourcecontrol/cgroups.go
+++ b/src/runtime/pkg/resourcecontrol/cgroups.go
@@ -25,9 +25,6 @@ const (
 	// find kata containers cgroup path on host to prevent it
 	// from grabbing the stats data.
 	CgroupKataPrefix = "kata"
-
-	// cgroup v2 mount point
-	unifiedMountpoint = "/sys/fs/cgroup"
 )
 
 func RenameCgroupPath(path string) (string, error) {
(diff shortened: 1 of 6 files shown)


liubin:
It seems that containerd/cgroups doesn't support macOS very well.
/test

src/runtime/pkg/resourcecontrol/utils.go
```diff
@@ -80,3 +82,33 @@ func IsSystemdCgroup(cgroupPath string) bool {
...
+func SandboxAndOverheadPath(cgroupPath string, overheadPath string, sandboxCgroupOnly bool) (string, string, error) {
+	isCgroupV1, err := IsCgroupV1()
+	if !isCgroupV1 && !IsSystemdCgroup(cgroupPath) && !sandboxCgroupOnly {
+		overheadPath = filepath.Join(cgroupPath, "overhead")
+		cgroupPath = filepath.Join(cgroupPath, "sandbox")
+	} else if !isCgroupV1 && IsSystemdCgroup(cgroupPath) && !sandboxCgroupOnly {
+		overheadPath = filepath.Join("/", cgroupPath, "overhead")
+		cgroupPath = filepath.Join("/", cgroupPath, "sandbox")
+	}
+	if err != nil {
```
liubin, line 109: `err` should be first and return early if error.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -719,9 +729,7 @@ func (s *Sandbox) createResourceController() error {
 		// into the sandbox resource controller.
 		// We're creating an overhead controller, with no constraints. Everything but
 		// the vCPU threads will eventually make it there.
-		overheadController, err := resCtrl.NewResourceController(fmt.Sprintf("%s%s", resCtrlKataOverheadID, s.id), &specs.LinuxResources{})
-		// TODO: support systemd cgroups overhead cgroup
-		// https://github.com/kata-containers/kata-containers/issues/2963
```
liubin, line 724: Does this PR close 2963?

src/runtime/pkg/resourcecontrol/utils.go
```diff
@@ -80,3 +82,33 @@ func IsSystemdCgroup(cgroupPath string) bool {
...
+	} else {
+		return false, ErrCgroupMode
+	}
+}
+
+// SandboxAndOverheadPath gets the cgroup path in thread mode in cgroup v2.
+// The sandbox and overhead cgroup in threaded mode are placed under the same cgroup in domain threaded mode.
+// In this way, vCPU threads and VMM processes can be separated into two cgroups.
+// For details, please refer to https://github.com/kata-containers/kata-containers/issues/4886.
+func SandboxAndOverheadPath(cgroupPath string, overheadPath string, sandboxCgroupOnly bool) (string, string, error) {
```
liubin, line 100: Input parameter `overheadPath` is not used, so it can be deleted.

liubin: Tough `cgroupPath` is used, but changing the input parameter and returning it is not a good design pattern.

src/runtime/virtcontainers/sandbox.go
yaoyinnan: I deleted this issue record by mistake and it has been restored.

src/runtime/pkg/resourcecontrol/utils.go
yaoyinnan: I modified how the parameters are used and added comments for two cases.

yaoyinnan:
> It seems that containerd/cgroups doesn't support macOS very well. /test

Yes, it looks like this repo unsupported macOS. The https://github.com/containerd/cgroups/issues/55 also mentioned this situation.

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -309,12 +315,14 @@ func (c *LinuxCgroup) AddProcess(pid int, subsystems ...string) error {
...
+func (c *LinuxCgroup) AddThread(tid int, subsystems ...string) error {
 	switch cg := c.cgroup.(type) {
 	case cgroups.Cgroup:
-		return cg.AddTask(cgroups.Process{Pid: pid})
+		return cg.AddTask(cgroups.Process{Pid: tid})
 	case *cgroupsv2.Manager:
-		return cg.AddProc(uint64(pid))
+		// Currently containerd/cgroups does not provide the AddThread() method.
+		// To be replaced by the method in manager after https://github.com/containerd/cgroups/pull/243 is merged.
+		return ioutil.WriteFile(filepath.Join(unifiedMountpoint, c.ID(), "cgroup.threads"), []byte(strconv.FormatUint(uint64(tid), 10)), os.FileMode(0))
```
liubin, line 325: `ioutil.WriteFile` is deprecated, please use `os.WriteFile` instead.

yaoyinnan: Already used `os.WriteFile` instead. `runc_cgroups.WriteFile` only allows the parameter type to be `string` and cannot write to cgroup.threads.

yaoyinnan:
@liubin Please help me to run workflows. thanks.

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -148,10 +145,13 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 		if err != nil {
 			return nil, err
 		}
-		cgroup, err = cgroupsv2.NewManager(unifiedMountpoint, cgroupPath, cgroupsv2.ToResources(resources))
+		cgroup, err = cgroupsv2.NewManager(UnifiedMountpoint, cgroupPath, cgroupsv2.ToResources(resources))
 		if err != nil {
 			return nil, err
 		}
+		if !sandboxCgroupOnly {
+			SetThreadedMode(cgroupPath)
```
bergwolf, line 153: Does it break systemd cgroup manager?

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -275,12 +275,7 @@ func (c *LinuxCgroup) Delete() error {
 	case cgroups.Cgroup:
 		return cg.Delete()
 	case *cgroupsv2.Manager:
-		if IsSystemdCgroup(c.ID()) {
-			if err := cg.DeleteSystemd(); err != nil {
-				return err
-			}
-		}
-		return cg.Delete()
+		return deleteCgroup(cg, filepath.Join(UnifiedMountpoint, c.path))
```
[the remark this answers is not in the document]
yaoyinnan: After testing, it is found that it is not necessary to use DeleteSystemd to send a job to delete cgroups to Systemd to delete systemd cgroups.

src/runtime/pkg/resourcecontrol/cgroups.go
bergwolf: In the case of systemd managed cgroups, we'd get something like `cgroupPath = /slice:scopeprefix:name`. It doesn't seem that `SetThreadedMode` is capable of handling such kind of path.

yaoyinnan: > In the case of systemd managed cgroups, we'd get something like `cgroupPath = /slice:scopeprefix:name`. It doesn't seem that `SetThreadedMode` is capable of handling such kind of path.

When `cgroup driver = systemd` and `sandbox_cgroup_only = false`, the following cgroup hierarchy will be generated.

```
root@k8s-worker01:~# ls /sys/fs/cgroup/kubepods-besteffort-podcb76d0a3_83c7_4ad2_922c_095eeea7e6fe.slice\:cri-containerd\:6595af1eab666f43d9bd8ca4e12347a40e3a2bac81393dc9fa3318e15050ece0/
cgroup.controllers      cgroup.stat             cpuset.cpus            cpu.uclamp.max   io.stat              memory.low        overhead      sandbox
cgroup.events           cgroup.subtree_control  cpuset.cpus.effective  cpu.uclamp.min   io.weight            memory.max        pids.current
cgroup.freeze           cgroup.threads          cpuset.cpus.partition  cpu.weight       memory.current       memory.min        pids.events
cgroup.max.depth        cgroup.type             cpuset.mems            cpu.weight.nice  memory.events        memory.oom.group  pids.max
cgroup.max.descendants  cpu.max                 cpuset.mems.effective  io.max           memory.events.local  memory.pressure   rdma.current
cgroup.procs            cpu.pressure            cpu.stat               io.pressure      memory.high          memory.stat       rdma.max
```

To be honest, this is not a reasonable way to manage cgroups, because cgroups generated using cgroupfs cannot be managed with systemd. And it does not follow the cgroup hierarchy of kubernetes, although in fact the source of statistical metrics is the cgroup in the agent, and the host cgroup does not affect the metrics.

If this approach is acceptable, I consider could using the `kata` as  parent directory to all cgroups created in the form of `cgroupPath = /slice:scopeprefix:name`, for example: `cgroupPath = /kata/slice:scopeprefix:name/sandbox` and `overheadPath = /kata/slice:scopeprefix:name/overhead`.

If this approach is unacceptable, it may be possible need to report an error when the user configures `cgroup driver = systemd` and `sandbox_cgroup_only = false` telling the user: Kata does not support using systemd to manage cgroups when `sandbox_cgroup_only = false`.

liubin (approved):

yaoyinnan:
@bergwolf @jodh-intel @fidencio @snir911  This PR has basically been completed. Can you review and push to approve this PR for me?

Issue #4336, opened 2022-05-30
Support systemd cgroup for kata agent


liubin:
Support systemd cgroup for kata agent.

Yuan-Zhuo:
I want to do this.

PR #5154, opened 2022-09-11
agent: support systemd cgroup for kata agent.

Yuan-Zhuo:
1. Implemented a rust module for operating cgroups through systemd with the help of zbus (src/agent/rustjail/src/cgroups/systemd).
2. Add support for optional cgroup configuration through fs and systemd at agent (src/agent/rustjail/src/container.rs).
3. Described the usage and supported properties of the agent systemd cgroup (docs/design/agent-systemd-cgroup.md).

Fixes: #4336

Signed-off-by: Yuan-Zhuo <yuanzhuo0118@outlook.com>


liubin:
@Yuan-Zhuo you need correct your commit message according to this [format](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format).

docs/design/agent-systemd-cgroup.md
```diff
@@ -0,0 +1,71 @@
...
+
+## usage
+
+For systemd, kata agent configures cgroups according to the following `linux.cgroupsPath` format standard provided by runc (`[slice]:[prefix]:[name]`). If you don't provide a valid `linux.cgroupsPath`, kata agent will treat it as `"system.slice:kata_agent:<container-id>"`. 
+
+> Here slice is a systemd slice under which the container is placed. If empty, it defaults to system.slice, except when cgroup v2 is used and rootless container is created, in which case it defaults to user.slice.
+>
+> Note that slice can contain dashes to denote a sub-slice (e.g. user-1000.slice is a correct notation, meaning a subslice of user.slice), but it must not contain slashes (e.g. user.slice/user-1000.slice is invalid).
+>
+> A slice of - represents a root slice.
```
liubin, line 13: "A slice of - represents a root slice." -> "A slice of `-` represents a root slice."

docs/design/agent-systemd-cgroup.md
```diff
@@ -0,0 +1,71 @@
...
+>
+> Next, prefix and name are used to compose the unit name, which is `<prefix>-<name>.scope`, unless name has .slice suffix, in which case prefix is ignored and the name is used as is.
```
liubin, line 15: ".slice" -> "`.slice`"

src/agent/rustjail/src/cgroups/mod.rs
```diff
@@ -38,4 +40,20 @@ pub trait Manager {
     fn set(&self, _container: &LinuxResources, _update: bool) -> Result<()> {
         Err(anyhow!("not supported!"))
     }
+
+    fn update_cpuset_path(&self, _: &str, _: &str) -> Result<()> {
+        Ok(())
```
liubin, line 45: Why not returning `Err` like other functions?

src/agent/rustjail/src/cgroups/systemd/cgroups_path.rs
```diff
@@ -0,0 +1,89 @@
+use anyhow::Result;
```
liubin, line 1: All new files should add license header, like this:

```
// Copyright 2021-2022 Kata Contributors
//
//


```

src/agent/rustjail/src/cgroups/systemd/cgroups_path.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use super::common::{DEFAULT_SLICE, SCOPE_SUFFIX, SLICE_SUFFIX};
+use std::string::String;
+
+macro_rules! sl {
+    () => {
+        slog_scope::logger()
```
liubin, line 8: Custom by tag, for example:

```rust
        slog_scope::logger().new(o!("subsystem" => "cgroups/systemd"))
```

src/agent/rustjail/src/cgroups/systemd/cgroups_path.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CgroupsPath {
+    pub slice: String,
+    pub prefix: String,
+    pub name: String,
+}
+
+impl CgroupsPath {
+    pub fn new(cgroups_path_str: &str) -> Result<Self> {
+        let path_vec: Vec<&str> = cgroups_path_str.split(':').collect();
```
liubin, line 21: It's better to check the path format, at least it must have 3 parts, to avoid panic of out of index error.

src/agent/rustjail/src/cgroups/systemd/cgroups_path.rs
```diff
@@ -0,0 +1,89 @@
...
+        return Ok((
+            parse_parent(self.slice.to_owned())?,
+            get_unit_name(self.prefix.to_owned(), self.name.to_owned()),
+        ));
+    }
+}
+
+fn parse_parent(slice: String) -> Result<String> {
+    if !slice.ends_with(SLICE_SUFFIX) || slice.contains('/') {
+        info!(sl!(), "invalid slice name: {}", slice);
```
liubin, line 45: return an `Err`?

src/agent/rustjail/src/cgroups/systemd/cgroups_path.rs
```diff
@@ -0,0 +1,89 @@
...
+    } else if slice == "-.slice" {
+        return Ok(String::from("/"));
+    }
+
+    let mut slice_path = String::new();
+    let mut prefix = String::new();
+    for subslice in slice.trim_end_matches(SLICE_SUFFIX).split('-') {
+        if subslice.is_empty() {
+            info!(sl!(), "invalid slice name: {}", slice);
```
liubin, line 54: return Err?

src/agent/rustjail/src/cgroups/systemd/dbus_client.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    fn add_process(&self, pid: i32, unit_name: &str) -> Result<()> {
+        let proxy = self
+            .build_proxy()
+            .with_context(|| format!("could not build proxy"))?;
+
+        proxy
+            .attach_processes_to_unit(unit_name, "/", &vec![pid as u32])
+            .with_context(|| format!("failed to stop unit {}", unit_name))?;
```
liubin, line 129: `"failed to stop unit {}"` -> `"failed to add process {}"`

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,127 @@
...
+        CpuSet::apply(r, &mut properties, &self.cg_hierarchy, systemd_version_str)?;
+
+        self.dbus_client
+            .set_properties(self.unit_name.as_str(), &properties)?;
+
+        Ok(())
+    }
+
+    fn get_stats(&self) -> Result<CgroupStats> {
+        return self.fs_manager.get_stats();
```
liubin, line 71: remove `return` and last `;`.

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    fn freeze(&self, state: FreezerState) -> Result<()> {
+        return self.fs_manager.freeze(state);
```
liubin, line 75: remove `return` and last `;`.

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    fn destroy(&mut self) -> Result<()> {
+        self.dbus_client.stop_unit(self.unit_name.as_str())?;
+        self.fs_manager.destroy()?;
+
+        Ok(())
+    }
+
+    fn get_pids(&self) -> Result<Vec<pid_t>> {
+        return self.fs_manager.get_pids();
```
liubin, line 86: remove `return` and last `;`.

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    fn update_cpuset_path(&self, guest_cpuset: &str, container_cpuset: &str) -> Result<()> {
+        return self
```
liubin, line 90: remove `return` and last `;`.

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,127 @@
...
+            .fs_manager
+            .update_cpuset_path(guest_cpuset, container_cpuset);
+    }
+
+    fn get_cg_path(&self, cg: &str) -> Option<String> {
+        return self.fs_manager.get_cg_path(cg);
```
liubin, line 96: remove `return` and last `;`.

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,127 @@
...
+        return self.fs_manager.get_stats();
+    }
+
+    fn freeze(&self, state: FreezerState) -> Result<()> {
+        return self.fs_manager.freeze(state);
+    }
+
+    fn destroy(&mut self) -> Result<()> {
+        self.dbus_client.stop_unit(self.unit_name.as_str())?;
+        self.fs_manager.destroy()?;
```
liubin, line 80: remove last `;` and `Ok(())`.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn unified_apply(
+        cpu_resources: &LinuxCpu,
+        properties: &mut Properties,
+        systemd_version: &str,
+    ) -> Result<()> {
+        if let Some(shares) = cpu_resources.shares {
+            let unified_shares = get_unified_cpushares(shares);
+            properties.push(("CPUShares", Value::U64(unified_shares)));
+        }
+        let quota = cpu_resources.quota.unwrap_or(i64::MAX);
```
liubin, line 59: For an invalid value, it's better to return an `Err` instead of using a Max value or 0, IMO.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
```diff
@@ -0,0 +1,110 @@
...
+        if quota != 0 || period != 0 {
+            let cpu_quota_per_sec_usec = get_unified_cpuquota(quota, period);
+            properties.push(("CPUQuotaPerSecUSec", Value::U64(cpu_quota_per_sec_usec)));
+        }
+        Ok(())
+    }
+}
+
+// [2-262144] to [1-10000]
+fn get_unified_cpushares(shares: u64) -> u64 {
```
liubin, line 73: Adding a link to describe why using this calculation is better.

liubin:
@Yuan-Zhuo are these two files:

- src/agent/rustjail/src/cgroups/systemd/interface/session.rs
- src/agent/rustjail/src/cgroups/systemd/interface/system.rs

generated by some tools?

If yes, there need more things to do with it:

- write a doc to describe how to generate it;
- skip the license header check by updating [this](https://github.com/kata-containers/tests/blob/b0ec458e3b77c2c52290948905a90b46bc600009/.ci/static-checks.sh#L424-L466).

Yuan-Zhuo:
@liubin Thank you very much for your careful review!

 I've done most of the fixes as you suggested, but there are still two issues that confuse me.

1. Lines 9 to 15 in agent-cgroups.md are quoted from `docs/design/agent-systemd-cgroup.md`. It is mentioned in section `references`. Is there any additional explanation needed?

2. I don't know the exact specification to describe generated code. I wrote the following draft to describe, and I plan to place this document in the same directory as `system.rs` and `session.rs`. Is this Ok?

   > # Systemd Interface
   >
   > `session.rs` and `system.rs` in the current directory are automatically generated by [`zbus-xmlgen`](https://gitlab.freedesktop.org/dbus/zbus/-/tree/main/zbus_xmlgen), which is is an accompanying tool provided by zbus to generate Rust code from D-Bus XML interface descriptions. The specific commands to generate these two files are as follows: 
   >
   > ```shell
   > // system.rs
   > zbus-xmlgen --system org.freedesktop.systemd1 /org/freedesktop/systemd1
   > // session.rs
   > zbus-xmlgen --session org.freedesktop.systemd1 /org/freedesktop/systemd1
   > ```
   >
   > The current implementation of `cgroups/systemd` only uses `system.rs` while `session.rs` could be used to build rootless containers.

src/agent/rustjail/src/cgroups/systemd/mod.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2021-2022 Kata Contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
```
bergwolf, line 5: Can this be moved to `src/libs`? I assume that runtime-rs will need to use it as well when it supports systemd cgroup on the host.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl Cpu {
+    // v1:
+    // cpu.shares <-> CPUShares
+    fn legacy_apply(cpu_resources: &LinuxCpu, properties: &mut Properties) -> Result<()> {
```
bergwolf, line 43: Let's just call it `apply_v1` and `apply_v2`.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpuset.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::convert::TryInto;
+
+    use crate::cgroups::systemd::subsystem::cpuset::BitMask;
+
+    #[test]
+    fn test_bitmask_conversion() {
+        let cpus = "2-4";
```
bergwolf, line 115: Please add more cases to validate BitMask calculation.

src/agent/rustjail/src/cgroups/systemd/common.rs
```diff
@@ -0,0 +1,17 @@
...
+pub const SLICE_SUFFIX: &str = ".slice";
+pub const SCOPE_SUFFIX: &str = ".scope";
+pub const UNIT_MODE: &str = "replace";
+
+pub type Properties<'a> = Vec<(&'a str, zbus::zvariant::Value<'a>)>;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum CgroupHierarchy {
+    Legacy,
+    Unified,
```
bergwolf, line 16: How about just V1 and V2 to be future proof? `Unified` can be legacy in future as well.

src/agent/rustjail/Cargo.toml
```diff
@@ -34,6 +34,9 @@ async-trait = "0.1.31"
 inotify = "0.9.2"
 libseccomp = { version = "0.2.3", optional = true }
 
+zbus = "2.3.0"
+bit-vec = "0.6.3"
```
bergwolf, line 38: please keep them in the same code block as the above dependencies.

src/agent/rustjail/src/cgroups/fs/mod.rs
```diff
@@ -193,6 +194,79 @@ impl CgroupManager for Manager {
...
+                &container_path,
+                container_cpuset
+            );
+            container_cpuset_controller.set_cpus(container_cpuset)?;
+        }
+
+        Ok(())
+    }
+
+    fn get_cg_path(&self, cg: &str) -> Result<String> {
```
bergwolf, line 257: please rename it to `get_cgroup_path`.

src/agent/rustjail/src/cgroups/systemd/manager.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        Ok(Manager {
+            paths: fs_manager.paths.clone(),
+            mounts: fs_manager.mounts.clone(),
+            cgroups_path,
+            cpath: cpath.to_owned(),
+            unit_name: unit_name.to_owned(),
+            dbus_client: DBusClient {},
+            fs_manager,
+            cg_hierarchy: if cgroups::hierarchies::is_cgroup2_unified_mode() {
```
bergwolf, line 122: One property of cgroups v2 is that it can be mixed with v1. It seems that we do not support such a setup. If so, please document it clearly in the design doc.

src/agent/rustjail/src/container.rs
```diff
@@ -1448,18 +1475,40 @@ impl LinuxContainer {
 
         let linux = spec.linux.as_ref().unwrap();
 
-        let cpath = if linux.cgroups_path.is_empty() {
-            format!("/{}", id.as_str())
+        // determine which cgroup driver to take and then assign to config.use_systemd_cgroup
+        // systemd: "[slice]:[prefix]:[name]"
+        // fs: "/path_a/path_b"
+        let cpath = if SYSTEMD_CGROUP_PATH_FORMAT.is_match(linux.cgroups_path.as_str()) {
```
bergwolf, line 1481: This should be determined when the agent starts (and save as a sandbox property) instead of when creating each container.

bergwolf (commented):
Mostly lgtm. Some comments inline. Thanks!

src/agent/rustjail/src/cgroups/mock.rs
```diff
@@ -53,6 +54,18 @@ impl CgroupManager for Manager {
     fn get_pids(&self) -> Result<Vec<pid_t>> {
         Ok(Vec::new())
     }
+
+    fn update_cpuset_path(&self, _: &str, _: &str) -> Result<()> {
+        Ok(())
+    }
+
+    fn get_cg_path(&self, _: &str) -> Result<String> {
```
fidencio, line 62: Same comment here about naming it `get_cgroup_path`.

src/agent/rustjail/src/cgroups/mod.rs
```diff
@@ -38,4 +40,22 @@ pub trait Manager {
     fn set(&self, _container: &LinuxResources, _update: bool) -> Result<()> {
         Err(anyhow!("not supported!"))
     }
+
+    fn update_cpuset_path(&self, _: &str, _: &str) -> Result<()> {
+        Err(anyhow!("not supported!"))
+    }
+
+    fn get_cg_path(&self, _: &str) -> Result<String> {
```
fidencio, line 48: Same comment here about renaming to `get_cgroup_path`.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
fidencio: If we're changing it, please, let's keep it consistent and change `legacy` -> `v1`, and `unified` -> `v2` everywhere.

fidencio (commented):
Work in general looks good.  I've left a few minor comments.

Once this PR gets updated addressing @bergwolf's comments, I'd like to ask for a few hours to run some tests with this.

docs/design/README.md
```diff
@@ -8,6 +8,7 @@ Kata Containers design documents:
 - [VSocks](VSocks.md)
 - [VCPU handling](vcpu-handling.md)
 - [Host cgroups](host-cgroups.md)
+- [Agent systemd cgroup](agent-systemd-cgroup.md)
```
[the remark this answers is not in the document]
Yuan-Zhuo: It seems that the rest of the doc doesn't follow either.

src/agent/rustjail/src/container.rs
Yuan-Zhuo: There is currently no explicit flag to indicate whether the agent should use systemd cgroup, but instead decides based on the `linux.cgroups_path` in the container's `config.json` when creating the container. Should I modify the logic so that the sandbox decides whether the entire sandbox uses systemd when the first container is created?

src/agent/rustjail/src/cgroups/systemd/common.rs
Yuan-Zhuo: `legacy` and `unified` seem to be widely used labels to distinguish cgroups, e.g. cgroup v2 can be enabled by append `systemd.unified_cgroup_hierarchy=1` to boot options.
At present, `FsManager` of kata-agent uses legacy and unified to identify v1 and v2, and the same is true for runtime.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpuset.rs
Yuan-Zhuo: Fixed

src/agent/rustjail/src/cgroups/systemd/manager.rs
Yuan-Zhuo: Fixed

liubin:
@Yuan-Zhuo please let this PR depend on https://github.com/kata-containers/tests/pull/5123 to pass the license header check.

liubin:
ping @kata-containers/architecture-committee, PTAL.

liubin:
@bergwolf @fidencio Do you have a time to review this?

src/agent/rustjail/src/container.rs
```diff
@@ -239,7 +243,7 @@ pub struct LinuxContainer {
     pub id: String,
     pub root: String,
     pub config: Config,
-    pub cgroup_manager: Option<FsManager>,
+    pub cgroup_manager: Option<Box<dyn Manager + Send + Sync>>,
```
bergwolf, line 246: Unrelated to this PR, but I wonder why wouldn't we always have a cgroup manager for a container. IOW, can we drop the `Option` here? It will simplify the code a lot by not having to check for NONE everywhere cgroup_manager is used.

src/agent/rustjail/src/container.rs
bergwolf: In the current implementation, it is required that the guest have to follow the same cgroup manager choice as the host. But what if the host asks for systemd cgroup but the guest only supports cgroup fs, or vice versa? So correct me if I'm wrong, but I think a better choise would be

bergwolf: In the current implementation, it is required that the guest have to follow the same cgroup manager choice as the host. But what if the host asks for systemd cgroup but the guest only supports cgroup fs? IIUC, systemd cgroup manager is actually a guest capability that may or may not exist (depending on the guest rootfs image setup). So a better choice might be to determine it when creating the guest VM, let the runtime know it and let the runtime decide which cgroup manager to use. So that we can still run kata-containers w/o guest systemd in a systemd host environment.

I admit that it is too much for the current PR so it is fine to merge it as it, adding a requirement that anyone who wants to run systemd cgroup manager must enable it both on the host and in the guest. And we can improve it incrementally in the future.

bergwolf (approved):
I'm giving it +1 and we can improve the systemd cgroup manager support incrementally in future PRs. Thanks @Yuan-Zhuo !

lifupan (approved):

Issue #5168, opened 2022-09-15
CC | Add root hash value and measure config to Kata configuration file


arronwy:
For https://github.com/confidential-containers/documentation/issues/40#top, after we have a guest kernel with builtin initramfs which provide the rootfs measurement capability and Kata rootfs image with hash device, we need set related root hash value and measure config to the kernel params in kata configuration file.

PR #5169, opened 2022-09-15
config: Add root hash value and measure config to kernel params

arronwy:
Implement subtask of https://github.com/confidential-containers/documentation/issues/40

After we have a guest kernel with builtin initramfs which provide the rootfs measurement capability and Kata rootfs image with hash device, we need set related root hash value and measure config to the kernel params in kata configuration file.

Fixes: #5168


fidencio (approved):
lgtm, thanks @arronwy!

I'd also like to hear from @stevenhorsman @ryansavino and @fitzthum

stevenhorsman (approved):
This looks fine to me, but I'll admit that I'm far from an expert on this stuff!

Issue #5170, opened 2022-09-15
Use macro to reduce duplicated code


liubin:
Use macro to reduce duplicated code in file `src/runtime-rs/crates/resource/src/network/utils/link/manager.rs`

PR #5171, opened 2022-09-15
runtime-rs/resource: use macro to reduce duplicated code

liubin:
Some device types have the same definition, they can be implemented by macro to reduce code.

And this commit also deleted the `peer_name` field of the structs that is never been used.

Fixes: #5170

### DIFF
--- a/src/runtime-rs/crates/resource/src/network/utils/link/manager.rs
+++ b/src/runtime-rs/crates/resource/src/network/utils/link/manager.rs
@@ -181,119 +181,40 @@ fn parse_bridge(mut ibs: Vec<InfoBridge>) -> Bridge {
     }
     bridge
 }
-#[derive(Debug, PartialEq, Eq, Clone, Default)]
-pub struct Device {
-    attrs: Option<LinkAttrs>,
+
+macro_rules! impl_network_dev {
+    ($r_type: literal , $r_struct: ty) => {
+        impl Link for $r_struct {
+            fn attrs(&self) -> &LinkAttrs {
+                self.attrs.as_ref().unwrap()
+            }
+            fn set_attrs(&mut self, attr: LinkAttrs) {
+                self.attrs = Some(attr);
+            }
+            fn r#type(&self) -> &'static str {
+                $r_type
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/resource/src/network/utils/link/manager.rs
```diff
@@ -181,119 +181,40 @@ fn parse_bridge(mut ibs: Vec<InfoBridge>) -> Bridge {
...
 
-#[derive(Debug, PartialEq, Eq, Clone, Default)]
-pub struct Vlan {
-    attrs: Option<LinkAttrs>,
+macro_rules! define_and_impl_network_dev {
+    ($r_type: literal , $r_struct: tt) => {
+        #[derive(Debug, PartialEq, Eq, Clone, Default)]
+        pub struct $r_struct {
+            attrs: Option<LinkAttrs>,
+        }
```
studychao, line 206: What about 
```
    /// on create only
    pub peer_name: String,
```
in `pub struct` definition
for `MacVlan`,  `IpVlan` and `Veth`.

studychao (commented):
Thanks bin, only one small comments.

src/runtime-rs/crates/resource/src/network/utils/link/manager.rs
liubin: I didn't find any reference for this field, is it ready for feature usage?

bergwolf: yup, we can always add it back if needed.

bergwolf (approved):
lgtm!

liubin:
@studychao PTAL.

studychao (approved):
Thanks Bin

src/runtime-rs/crates/resource/src/network/utils/link/manager.rs
studychao: I have checked `peer_name` and it is not used now.
We could remove it now and add it back if needed.

PR #5186, opened 2022-09-19
runtime-rs: use Path.is_file to check regular files

liubin:
In src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs, there are two ways used to check if a file is a regular file:

- one is using `stat` and its' mode
- one is using [`Path`'s `is_file()`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.is_file)

The `Path`'s `is_file()` is enough to check if a file is a regular file.


bergwolf (approved):
lgtm!

snir911 (approved):

Issue #5191, opened 2022-09-19
hugepages cgroups files paths doesn't exist


snir911:
# Description of problem

`Error: CreateContainer failed: unable to read a control group file caused by: Os { code: 2, kind: NotFound, message: "No such file or directory" }: unknown`

This happens when guest/podvm supported different set of hugepage sizes of the host (common use-case for peer-pods)

# Expected result

pod is running

# Actual result
`Error: CreateContainer failed: unable to read a control group file caused by: Os { code: 2, kind: NotFound, message: "No such file or directory" }: unknown`
and pod is not running

# Further information

to reproduce:
* set k8s cluster with cri-o on AWS t2.2xlarge
*  [Deploy](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/install) peer pods and try to run some workload (use podvm t2.small flavor)

PR #5192, opened 2022-09-19
CCv0: agent: validate hugepage size is supported

snir911:
before setting a limit, otherwise paths may not be found. guest supporting different hugepage size is more likely with peer-pods where podvm may use different flavor.

(sent only against CCv0 however, it may make sense to port also to main)
Fixes: #5191

### DIFF
--- a/src/agent/rustjail/src/cgroups/fs/mod.rs
+++ b/src/agent/rustjail/src/cgroups/fs/mod.rs
@@ -252,19 +252,28 @@ fn set_devices_resources(
 }
 
 fn set_hugepages_resources(
-    _cg: &cgroups::Cgroup,
+    cg: &cgroups::Cgroup,
     hugepage_limits: &[LinuxHugepageLimit],
     res: &mut cgroups::Resources,
 ) {
     info!(sl!(), "cgroup manager set hugepage");
     let mut limits = vec![];
+    let hugetlb_controller = cg.controller_of::<HugeTlbController>();
 
     for l in hugepage_limits.iter() {
-        let hr = HugePageResource {
-            size: l.page_size.clone(),
-            limit: l.limit,
-        };
-        limits.push(hr);
+        if hugetlb_controller.is_some() && hugetlb_controller.unwrap().size_supported(&l.page_size)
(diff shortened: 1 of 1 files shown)


pmores (approved):
Looks good to me!

snir911:
@liubin do you think it should be ported  also to main?

gkurz:
> (sent only against CCv0 however, it may make sense to port also to main)

I'm a bit confused by the workflow... if this PR fixes something in main, why don't we merge it in main first and then backport to CCv0 ?

snir911:
> > (sent only against CCv0 however, it may make sense to port also to main)
> 
> I'm a bit confused by the workflow... if this PR fixes something in main, why don't we merge it in main first and then backport to CCv0 ?

The reason is that i know it's essential for peer-pods where host and podvm may support different hugepage sizes. I was not even sure it's possible in kata where guest is an actual virtual machine which eventually using/sharing the same cpu

snir911:
updated:
```
@@ -261,7 +261,7 @@ fn set_hugepages_resources(
     let hugetlb_controller = cg.controller_of::<HugeTlbController>();
 
     for l in hugepage_limits.iter() {
-        if hugetlb_controller.is_none() || hugetlb_controller.unwrap().size_supported(&l.page_size)
+        if hugetlb_controller.is_some() && hugetlb_controller.unwrap().size_supported(&l.page_size)
         {
             let hr = HugePageResource {
                 size: l.page_size.clone(),
@@ -271,7 +271,7 @@ fn set_hugepages_resources(
         } else {
             warn!(
                 sl!(),
-                "{} page size is not supported, dropping requested limit", l.page_size
+                "{} page size support cannot be verified, dropping requested limit", l.page_size
             );
         }
     }
```
/test
@gkurz

gkurz (approved):
LGTM and it was a good opportunity to read some more rust code ! :smile:

Thanks @snir911 !

amshinde:
@snir911 LGTM. I would suggest first get this applied to main and then backport to CCV0. I think we have seen similar errors on the main branch.

Issue #5226, opened 2022-09-22
stable-2.5 arm64 maxvcpus only support 8 cores


YchauWang:
# Description of problem

Our arm version production need to support more cores than the 8 cores. The GIC v2 limited, and set it to 8.

# Expected result

Creating a kata vm with 16vcpu or more on our arm server.

# Actual result

The arm kata vm maxvcpus only support 8 cores.

PR #5227, opened 2022-09-22
hypervisor:arm: gic v2 max support 8 arm core

YchauWang:
arm64 maxvcpus only support 8 arm cores, limited by gicv2. The gic v3/v4 support more arm cores. This patch update the gic version.

Fixes: #5226

### DIFF
--- a/src/runtime/virtcontainers/qemu_arm64.go
+++ b/src/runtime/virtcontainers/qemu_arm64.go
@@ -30,7 +30,7 @@ const defaultQemuMachineType = QemuVirt
 
 const qmpMigrationWaitTimeout = 10 * time.Second
 
-const defaultQemuMachineOptions = "usb=off,accel=kvm,gic-version=host"
+const defaultQemuMachineOptions = "usb=off,accel=kvm,gic-version=max"
 
 var kernelParams = []Param{
 	{"iommu.passthrough", "0"},


liubin:
@jongwu PTAL

jongwu:
This change is wrong. We should set gic-version as the same with host. We can't support gicv3 in guest if the host only have gicv2.

YchauWang:
> This change is wrong. We should set gic-version as the same with host. We can't support gicv3 in guest if the host only have gicv2.

Hi @jongwu, I check the gic-version valid values are 3, 2, host, max on over servers. But the `gic-version=host` config used the V2 GIC. I forced using the `gic-version=3` and `gic-version=max`, it is ok. Is the `max` to the high `gic-version` of our server supporting?

jongwu:
Hi @YchauWang -, I check for the Qeme code [here](https://github.com/qemu/qemu/blob/6338c30111d596d955e6bc933a82184a0b910c43/hw/arm/virt.c#L1795). If no "kernel-irqchip=off" set, "max" is the same with "host".  It's strange that you get different result between "max" and "host".
Also, I'm confused with that you get gicv2 when you run kata on a server support gicv3? As qemu will probe gic version through kvm before the final version is decided.

YchauWang:
> Hi @YchauWang -, I check for the Qeme code [here](https://github.com/qemu/qemu/blob/6338c30111d596d955e6bc933a82184a0b910c43/hw/arm/virt.c#L1795). If no "kernel-irqchip=off" set, "max" is the same with "host". It's strange that you get different result between "max" and "host". Also, I'm confused with that you get gicv2 when you run kata on a server support gicv3? As qemu will probe gic version through kvm before the final version is decided.

It is very strange for me too. I keep default config using `host`, it is `v2`, but my arm server is support v3. I just want to use the high version of gic-version. And how make sure max support value the gic-version on the arm server?

jongwu:
Hi @YchauWang -, I think both "max" and "host" can do.

We should find the real root cause of it before we decide if there need a change for the gic-version in kata. 

As kata is a virtual machine, you can debug this issue by running a pure Qemu VM on your server. You can use gdb or print log to find what happen inside Qemu.

YchauWang:
Thanks @jongwu, I will use the hard version of `gic-version`, such as `v2` or `v3`, to our production.

Issue #5172, opened 2022-09-15
support hotplug/hot-unplug VFIO Device in container running state


soulseen:
**Is your feature request related to a problem? Please describe.**

In some cases, we need to hot plug/unplug GPUs Device(or other VFIO devices) dynamically into kata container when it is running, to implement the GPUs dynamically assign between different kata containers.

**Describe the solution you'd like**

AFAIK we are already implemented the [DeviceManager](https://github.com/kata-containers/kata-containers/blob/6b0dc60dda37251f7c4b25ff1cfb10ed11dc88e4/src/runtime/virtcontainers/device/api/interface.go#L89) interface. so,  I think we could add a new Kata-specific sub-command like:
```
kata-runtime hotplug $container $device
```
that would allow users to hotplug or hot-unplug VFIO Devices.

fengwang666:
Exposing an interface through kata-runtime is potentially an option. Can you explain a bit more about the use case? Sounds like the GPU hotplug/unplug lifecycle is different from the pod/containers? How would your container application handle the scenario when a GPU suddenly becomes unavailable?

soulseen:
yes, the GPU hotplug/unplug lifecycle is independent of the pod/containers. We have some applications that only need the GPU at certain times, and for the times when the GPU is not needed, we want to be able to hotplug/unplug the GPU to other applications.

soulseen:
yes, as you said, in the view of decoupling, GPU resources will be independent of k8s,  and they will be controlled by other components that will hotplug GPU to applications.

soulseen:
@fengwang666 sounds good, But this approach will cause the pod/container to be recreated, this is not expected.

zvonkok:
@soulseen Can you elaborate more on the use-case, as @bergwolf already mentioned, this is not something Kubernetes can achieve and will break the semantics of how devices are handled in a container. What is the concern here? Have a look at our recent blog post on how to improve GPU utilization in Kubernetes: https://developer.nvidia.com/blog/improving-gpu-utilization-in-kubernetes/
BTW. we also added a new resource management API to Kube please take a  look: https://github.com/kubernetes/enhancements/pull/3064

bergwolf:
@zvonkok Thanks for the links to KEP-3063! It would be great to have something flexible like KEP-3063. One quick question, so with KEP-3063, is the CDI the only interface that kata-containers need to work with? I'm thinking we should start to integrate with CDI to prepare for it.

[...]

PR #5236, opened 2022-09-23
support hotplug/hot-unplug device

soulseen:
In serverless scenarios, we need to hot plug/unplug GPUs Device dynamically into kata container without container restart, to implement the GPUs dynamically assign between different kata containers. 
we will manage GPU resources independent of the upper-level container manager(containerd/k8s), and they will be controlled by other components that will hotplug GPU to applications.
Fixes: https://github.com/kata-containers/kata-containers/issues/5172

This PR is aimed at expose the attach/detach interface to the user via `kata-runtime` command, then allowing users to hotplug or hot-unplug Devices. it is similar to `kata -runtime direct-volume` command:

```
kata-runtime device list --sandbox-id=${sandbox-id}
kata-runtime device attach --sandbox-id=$sandbox-id --device=$path_to_device"
kata-runtime device detach --sandbox-id=$sandbox-id --device=$path_to_device"
```

Signed-off-by: soulseen <zhuxiaoyang1996@gmail.com>

### DIFF
--- a/src/runtime/cmd/kata-runtime/kata-device.go
+++ b/src/runtime/cmd/kata-runtime/kata-device.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2022 NetEase Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	containerdshim "github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/katautils"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils/shimclient"
+	"github.com/urfave/cli"
+)
+
+var (
+	devPath string
+)
+
+var deviceSubCmds = []cli.Command{
+	listDeviceCommand,
+	attachDeviceCommand,
+	detachDeviceCOmmand,
+}
+
+var kataDeviceCommand = cli.Command{
+	Name:        "device",
+	Usage:       "attach or detach device to or from Kata Containers",
+	Subcommands: deviceSubCmds,
+	Action: func(context *cli.Context) {
+		cli.ShowSubcommandHelp(context)
+	},
+}
+
+var listDeviceCommand = cli.Command{
+	Name:  "list",
+	Usage: "list all assigned device",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for getting the devices",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+	},
+	Action: func(c *cli.Context) error {
+		// verify sandbox exists:
+		if err := katautils.VerifyContainerID(sandboxID); err != nil {
+			return err
+		}
+
+		url := containerdshim.DeviceUrl
+
+		body, err := shimclient.DoGet(sandboxID, defaultTimeout, url)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println(string(body))
+		return nil
+	},
+}
+
+var attachDeviceCommand = cli.Command{
+	Name:  "attach",
+	Usage: "attach the device to sandbox",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for attach the device",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.StringFlag{
+			Name:        "device",
+			Usage:       "absolute path of the device on the host",
+			Required:    true,
+			Destination: &devPath,
+		},
+	},
+	Action: func(c *cli.Context) error {
+		// verify sandbox exists:
+		if err := katautils.VerifyContainerID(sandboxID); err != nil {
+			return err
+		}
+
+		resizeReq := containerdshim.DeviceRequest{
+			DevicePath: devPath,
+		}
+		encoded, err := json.Marshal(resizeReq)
+		if err != nil {
+			return err
+		}
+
+		return shimclient.DoPut(sandboxID, defaultTimeout*10, containerdshim.DeviceUrl, "application/json", encoded)
+	},
+}
+
+var detachDeviceCOmmand = cli.Command{
+	Name:  "detach",
+	Usage: "detach the device from sandbox",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for detach the device",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.StringFlag{
+			Name:        "device",
+			Usage:       "absolute path of the device on the host",
+			Required:    true,
+			Destination: &devPath,
+		},
+	},
+	Action: func(c *cli.Context) error {
+		// verify sandbox exists:
+		if err := katautils.VerifyContainerID(sandboxID); err != nil {
+			return err
+		}
+
+		resizeReq := containerdshim.DeviceRequest{
+			DevicePath: devPath,
+		}
+		encoded, err := json.Marshal(resizeReq)
+		if err != nil {
+			return err
+		}
+
+		return shimclient.DoDelete(sandboxID, defaultTimeout*10, containerdshim.DeviceUrl, "application/json", encoded)
+	},
+}
--- a/src/runtime/pkg/containerd-shim-v2/shim_management.go
+++ b/src/runtime/pkg/containerd-shim-v2/shim_management.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/grpc/codes"
 
 	cdshim "github.com/containerd/containerd/runtime/v2/shim"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/katautils"
 	mutils "github.com/kata-containers/kata-containers/src/runtime/pkg/utils"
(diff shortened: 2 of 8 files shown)


soulseen:
@liubin @bergwolf @fengwang666 hi, PTAL, thanks.

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+	},
+}
+
+var listDeviceCommand = cli.Command{
+	Name:  "list",
+	Usage: "list all assigned device",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for getting the iptables",
```
Ji-Xinyou, line 42: I think you may forget to replace the `Usage`

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+	},
+}
+
+var attachDeviceCommand = cli.Command{
+	Name:  "attach",
+	Usage: "attach the device to sandbox",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for getting the iptables",
```
Ji-Xinyou, line 76: Same w/ above

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+	},
+}
+
+var detachDeviceCOmmand = cli.Command{
+	Name:  "detach",
+	Usage: "detach the device from sandbox",
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for getting the iptables",
```
Ji-Xinyou, line 110: Same

src/runtime/pkg/katautils/utils_unix.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if !stat.IsDir() {
+		dev, err := DeviceFromPath(devPath)
+		if err != nil {
+			return devInfo, err
+		}
+		return *dev, nil
+	}
+
+	return devInfo, nil
```
Ji-Xinyou, line 30: This returns a default value. And further be attached by sandbox, it this an expected result, just being curious.

src/runtime/cmd/kata-runtime/kata-device.go
Ji-Xinyou: Maybe "the target sandbox to ______ devices"

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -241,6 +243,82 @@ func (s *service) genericIPTablesHandler(w http.ResponseWriter, r *http.Request,
...
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		logger.WithError(err).Error("failed to read request body")
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte(err.Error()))
+		return
+	}
+
+	switch r.Method {
+	case http.MethodGet:
```
bergwolf, line 261: Not on this PR particularly, but we really need more abstraction in the shim management code, and a place to clearly define/document the APIs for each of these methods.

src/runtime/cmd/kata-runtime/kata-device.go
bergwolf, line 42: `iptables`?

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.StringFlag{
+			Name:        "device",
+			Usage:       "the absolute path of device",
```
bergwolf, line 48: Is it the path of the host or the guest?

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
```
bergwolf, line 82: Is it the path of the host or the guest?

src/runtime/pkg/katautils/utils_unix.go
soulseen: updated the case when the device path is a directory, please review again.

src/runtime/cmd/kata-runtime/kata-device.go
soulseen: updated, absolute path of the device on the host

src/runtime/pkg/containerd-shim-v2/shim_management.go
Ji-Xinyou: I agree, IMO we should maybe hide the usage of http `PUT/POST/DELETE/GET` below some other functions like `Set/Update/Remove/Query` to clarify the meaning and easier to develop?

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for getting the devices",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.StringFlag{
+			Name:        "device",
+			Usage:       "absolute path of the device on the host",
+			Destination: &devPath,
```
Ji-Xinyou, line 49: Why listing assigned device needs a devPath on host?

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for attach the device",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.StringFlag{
+			Name:        "device",
+			Usage:       "absolute path of the device on the host",
+			Destination: &devPath,
```
Ji-Xinyou, line 83: I think when attaching a `device`, the `device` flag is always required.

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,136 @@
...
+		cli.StringFlag{
+			Name:        "sandbox-id",
+			Usage:       "the target sandbox for detach the device",
+			Required:    true,
+			Destination: &sandboxID,
+		},
+		cli.StringFlag{
+			Name:        "device",
+			Usage:       "absolute path of the device on the host",
+			Destination: &devPath,
```
Ji-Xinyou, line 117: Same as above.

src/runtime/pkg/katautils/utils_unix.go
```diff
@@ -0,0 +1,82 @@
...
+	case unix.S_IFBLK:
+		devType = blockDevice
+	case unix.S_IFCHR:
+		devType = charDevice
+	case unix.S_IFIFO:
+		devType = fifoDevice
+	default:
+		return nil, fmt.Errorf("not a device node")
+	}
+	fm := os.FileMode(mode &^ unix.S_IFMT)
```
Ji-Xinyou, line 69: **nit**: I suggest using `filemode` as the variable's name since `fm` is not a idiomatic abbreviation.

Also you may add a comment to be more clear that `mode &^ S_IFMT` will **ignore the filetype (socket, ..., etc.) and only keep the permission**

Ji-Xinyou (commented):
Thanks @soulseen for the PR :)

A few more comments down below.

Ji-Xinyou:
Generally LGTM

soulseen:
@bergwolf @liubin @amshinde PTAL

src/runtime/cmd/kata-runtime/kata-device.go
```diff
@@ -0,0 +1,133 @@
...
+
+var deviceSubCmds = []cli.Command{
+	listDeviceCommand,
+	attachDeviceCommand,
+	detachDeviceCOmmand,
+}
+
+var kataDeviceCommand = cli.Command{
+	Name:        "device",
+	Usage:       "attach ot detach device to kata container",
```
liubin, line 29:
```diff
- "attach ot detach device to kata container"
+ "attach or detach device to or from Kata Containers"
```

liubin:
Thank you @soulseen, I have no comments about this PR, but think this feature also needs an integration test to verify the function is correct.

soulseen:
> Thank you @soulseen, I have no comments about this PR, but think this feature also needs an integration test to verify the function is correct.

@liubin, yes I think so, although it verifies fine locally, and may I ask how I can add an integration test for this PR because it needs GPU or other vfio-devices?

liubin:
@soulseen does need a GPU device? Can use another type of device? For example, simply create a device by `mknod` and pass it to the Guest.

And this is only a usage question: how do you let containers in the guest know a new device is added to the guest? Did you mount guest `/dev` into containers?

soulseen:
@liubin In this scenario, the device should be pre-allocated to the container(pod), then simply hot-unplug when the Device is idle and hot-plugin in when needed.
if only attach to VM without mounting to the container, I think the container can not aware of it.

soulseen:
> @soulseen does need a GPU device? Can use another type of device? For example, simply create a device by `mknod` and pass it to the Guest.

good point,  let me try.

bergwolf (commented):
Just one more comment, otherwise lgtm. Thanks!

Issue #5240, opened 2022-09-23
runtime: can not run containers in parallel


yaoyinnan:
When running more than 5 containers in parallel, we have seen a lot of errors. The more containers in parallel, the more errors we have (and different errors).

```
5.15.63
FATA[0030] failed to create shim task: open /sys/fs/cgroup/systemd/vc/tasks: no such file or directory: not found
FATA[0036] failed to create shim task: Failed to Check if grpc server is working: rpc error: code = DeadlineExceeded desc = timed out connecting to vsock 1541455645:1024: unknown
```

For detail, please refer to https://gist.github.com/norbjd/1773b2930227d89c347f223f1d9b4ff0. (From norbjd)

PR #5241, opened 2022-09-23
runtime: add SingleContainer when obtaining OCI Spec

yaoyinnan:
When creating a cgroup, add a SingleContainer when obtaining the OCI Spec to apply to ctr, podman, etc. If the CgroupPath is not obtained from the container config, the sandbox id is used as the cgroup path.

Fixes: #5240

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -2576,11 +2576,11 @@ func (s *Sandbox) GetPatchedOCISpec() *specs.Spec {
 		return nil
 	}
 
-	// get the container associated with the PodSandbox annotation. In Kubernetes, this
-	// represents the pause container. In Docker, this is the container.
-	// On Linux, we derive the group path from this container.
+	// Get the container associated with the PodSandbox annotation.
+	// In Kubernetes, this represents the pause container. In Docker, this is the container.
+	// On Linux, we derive the cgroup path from this container.
 	for _, cConfig := range s.config.Containers {
-		if cConfig.Annotations[annotations.ContainerTypeKey] == string(PodSandbox) {
+		if ContainerType(cConfig.Annotations[annotations.ContainerTypeKey]).IsSandbox() {
 			return cConfig.CustomSpec
 		}
 	}


norbjd:
Thanks! It solves one of our issues:

```shell
nerdctl pull ubuntu

for i in `seq 20`
do
  nerdctl run --rm --runtime io.containerd.run.kata.v2 ubuntu uname -r &
done
```

Extract of output (notice the errors) **before** the PR with kata 2.5.1:

```
5.15.63         
5.15.63
...
FATA[0030] failed to create shim task: open /sys/fs/cgroup/systemd/vc/tasks: no such file or directory: not found
FATA[0031] failed to create shim task: open /sys/fs/cgroup/cpuset/vc/cpuset.mems: no such file or directory: not found
```

**After** these changes, there are no more errors.

src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -61,12 +68,44 @@ func ValidCgroupPathV2(path string, systemdCgroup bool) (string, error) {
 		return filepath.Clean(path), nil
 	}
 
+	if path == "" {
```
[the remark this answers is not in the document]
norbjd: Do you think that it is possible that because this method is not thread-safe, we are experiencing also other issues (like `timed out connecting to vsock`), as shown here: https://gist.github.com/norbjd/1773b2930227d89c347f223f1d9b4ff0#30-containers-in-parallel?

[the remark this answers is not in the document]
yaoyinnan: Thanks for your suggestion, I will follow this to find the cause of the error and fix it.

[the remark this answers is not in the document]
yaoyinnan: I still think random strings should be used, because the cgroup path is empty when nerdctl runs, it will cause conflicts between cgroups of multiple sandboxes, which itself is not correct.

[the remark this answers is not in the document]
yaoyinnan: We discussed and get the results on Slack. The problem of `timed out connecting to vsock` was caused by running kata in a virtual machine. This problem was avoided on bare metal.

src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -43,28 +45,36 @@ func ValidCgroupPathV1(path string, systemdCgroup bool) (string, error) {
 	return filepath.Join(DefaultResourceControllerID, filepath.Clean("/"+path)), nil
 }
 
-// ValidCgroupPathV2 returns a valid cgroup path for cgroup v2.
-// see https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
-func ValidCgroupPathV2(path string, systemdCgroup bool) (string, error) {
+func AbsolutePath(path string) string {
```
liubin, line 48: Are `AbsolutePath` and `RandStr` need to be package exported names?

liubin (approved):

src/runtime/pkg/resourcecontrol/utils_linux.go
yaoyinnan: fixed

src/runtime/pkg/resourcecontrol/utils_linux_test.go
```diff
@@ -102,74 +102,7 @@ func TestValidCgroupPathV1(t *testing.T) {
 		} else {
 			assert.True(
 				strings.HasPrefix(path, DefaultResourceControllerID),
-				"%v should have prefix /%v", path, DefaultResourceControllerID)
-		}
-	}
-
-}
-
-func TestValidCgroupPathV2(t *testing.T) {
```
bergwolf, line 111: When removing the old code's UT, please add new UTs for the added code as well.

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -134,8 +134,14 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 	var cgroup interface{}
 	var cgroupPath string
 
+	// When the path is empty, a random string is generated
+	// as the ResourceControllerID to prevent duplicates.
+	if path == "" || path == "/" {
+		path = randStr(32)
```
bergwolf, line 140: While it lowers the chance of a collision, it is still possible to get two identical paths here. I tend to agree with @fengwang666 that we need to make the function thread-safe to fix the root cause.

src/runtime/pkg/resourcecontrol/utils_linux_test.go
yaoyinnan: `ValidCgroupPathV1()` and `ValidCgroupPathV2()` were merged into `ValidCgroupPath()`.

src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -43,28 +45,36 @@ func ValidCgroupPathV1(path string, systemdCgroup bool) (string, error) {
...
-		return "", fmt.Errorf("malformed systemd path '%v': expected to be of form 'slice:prefix:name'", path)
-	}
+	return filepath.Join("/", path)
+}
 
-	// In the case of an absolute path (starting with /), the runtime MUST
-	// take the path to be relative to the cgroups mount point.
-	if filepath.IsAbs(path) {
-		return filepath.Clean(path), nil
+func RandStr(n int) string {
```
c3d, line 53: Why do you need to invent a function to generate a random string? Other places do something like:
```
	sandboxId := uuid.Generate().String()
```

Wouldn't that be enough?

c3d (changes requested):
Hi @yaoyinnan. I believe we should use existing UUID-generating code rather than invent a new way to generate random strings.

src/runtime/pkg/resourcecontrol/utils_linux.go
yaoyinnan: Sorry, I don't know a better way, I have replaced it.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -688,6 +689,9 @@ func (s *Sandbox) createResourceController() error {
 		//TODO: in Docker or Podman use case, it is reasonable to set a constraint. Need to add a flag
 		// to allow users to configure Kata to constrain CPUs and Memory in this alternative
 		// scenario. See https://github.com/kata-containers/runtime/issues/2811
+	} else {
+		// Generate a random string as the cgroup path to prevent duplicates.
+		cgroupPath = uuid.Generate().String()
```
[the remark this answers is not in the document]
yaoyinnan: Perhaps a more correct way is to report an error?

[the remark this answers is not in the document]
bergwolf: How did you reproduce it? I wonder if it is because some container annotation is not set properly. We should always be able to find the podsandbox container when creating a new sandbox in the current kata workflow.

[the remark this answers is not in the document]
smile-luobin: I have the same issue.  
just use `ctr run --runtime "io.containerd.kata.v2"`, the `containerType` will be `singleContainer`, so `cgroupPath` will be `""`. 

why choose to use a generated random string instead of sandbox.id? 
As far as I know: 

- using randStr, the target cgroup is '/sys/fs/cgroup/vc/{randStr}'; 
- using `sandbox.Id`, the target cgroup is '/sys/fs/cgroup/vc/{sandbox.Id}';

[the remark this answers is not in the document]
yaoyinnan: @bergwolf Yes, it should be that SingleContainer was missed when judging the ContainerType of the annotation.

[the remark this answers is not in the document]
yaoyinnan: @smile-luobin Yes, it is better to directly use the sandbox id as the cgroup path, I modified it.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -689,6 +689,12 @@ func (s *Sandbox) createResourceController() error {
 		//TODO: in Docker or Podman use case, it is reasonable to set a constraint. Need to add a flag
 		// to allow users to configure Kata to constrain CPUs and Memory in this alternative
 		// scenario. See https://github.com/kata-containers/runtime/issues/2811
+	} else {
```
bergwolf, line 692: We've handled the `SingleContainer` case. Can we just fail here if spec is not found? Otherwise, I wonder how we can handle other spec-related operations.

bergwolf: In fact, there are places we assume `GetPatchedOCISpec` can always find the spec. Can you please change `GetPatchedOCISpec` to default to the first container in the sandbox if no proper annotations are found, and change the callers accordingly? Let's make sure we have such a dependency applied.

yaoyinnan: @bergwolf Sometimes it is not possible to use the first container in the sandbox as the OCI Spec, because some UTs use empty ContainerConfig to create sandboxes, and if they are modified, the default values ​​in most UTs will need to be changed. Here I am modifying to print debug information in the log instead of returning err. Use the sandbox id as the default cgroup path when the cgroupPath cannot be obtained from the OCI Spec.

bergwolf: @yaoyinnan Let's not add meaningless code just for UTs. If UTs need to be fixed, let's fix them.

c3d (commented):
Hi @yaoyinnan, overall, this looks like good progress. I somewhat agree with @bergwolf 's comments, so I'll wait for the next iteration for a final approval, but from my PoV, you are on the right track. Thanks!

yaoyinnan:
@bergwolf @fengwang666 @c3d @fidencio  This PR has basically been completed. Can you review and push to approve this PR for me?

src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -21,9 +21,9 @@ import (
 // DefaultResourceControllerID runtime-determined location in the cgroups hierarchy.
 const DefaultResourceControllerID = "/vc"
 
-// ValidCgroupPathV1 returns a valid cgroup path for cgroup v1.
+// ValidCgroupPath returns a valid cgroup path.
 // see https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
-func ValidCgroupPathV1(path string, systemdCgroup bool) (string, error) {
+func ValidCgroupPath(path string, systemdCgroup bool) (string, error) {
```
jepio, line 26: The code of this function makes no sense because the second argument is `systemdCgroup := IsSystemdCgroup(path)`.

So the two first branches `if IsSystemdCgroup(path)` and `if systemdCgroup` perform the same check with different returns.

Please fix this somehow.

src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -21,9 +21,9 @@ import (
...
 	if IsSystemdCgroup(path) {
 		return path, nil
```
jepio, line 28: For CgroupsV2 this previously always performed a transformation:
```
return filepath.Join("/", path), nil
```
Why has this been removed? I would prefer if the commit description or PR description explained this.

src/runtime/pkg/resourcecontrol/utils_linux.go
yaoyinnan: Moved to #5153 to resolve the issue.

src/runtime/pkg/resourcecontrol/utils_linux.go
yaoyinnan: Moved to #5153 to resolve the issue.

yaoyinnan:
Moved to #8974 for processing

Issue #5253, opened 2022-09-27
dragonball: linux-loader version need to be updated to 0.6.0


studychao:
Since linux-loader 0.4.0 and 0.5.0 is yanked due to a null terminator bug(furture information : https://github.com/rust-vmm/linux-loader/blob/main/CHANGELOG.md ), we need to update linux-loader to 0.6.0 in Dragonball.

PR #5254, opened 2022-09-27
Dragonball: update linux_loader to 0.6.0

studychao:
Since linux-loader 0.4.0 and 0.5.0 is yanked due to null terminator bug(furture information : https://github.com/rust-vmm/linux-loader/blob/main/CHANGELOG.md ), we need to update linux-loader to 0.6.0.

And as_str() function should also be changed.

fixes: #5253


Tim-0731-Hzt (approved):

liubin (approved):

Tim-0731-Hzt:
/(re)test-fedora

Issue #5255, opened 2022-09-27
agent: Unnecessary metrics are registered by default


Yuan-Zhuo:
Currently the agent uses `prometheus::gather` to collect metrics, which will be based on `DEFAULT_REGISTRY`. However, as shown below, `DEFAULT_REGISTRY` will give us insight into memory and CPU consumption, open file descriptors of the running Rust process, etc. These statistics are unnecessary for kata-monitor's **well-designed metrics**, many of which are repeated and polluted final statistics. Therefore, **a custom register** should be used to cleanly customize kata's metrics.

```
// ... kata_agent & kata_guest metrics ...

// ... unnecessary metrics ...
# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 0
# HELP process_max_fds Maximum number of open file descriptors.
# TYPE process_max_fds gauge
process_max_fds 1048576
# HELP process_open_fds Number of open file descriptors.
# TYPE process_open_fds gauge
process_open_fds 10
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 1921024
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1664248997
# HELP process_threads Number of OS threads in the process.
# TYPE process_threads gauge
process_threads 2
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 78172160
```

PR #5256, opened 2022-09-27
agent: Eliminate unnecessary metrics

Yuan-Zhuo:
DEFAULT_REGISTRY pre-registers many metrics that we don't need or have duplicated. This PR uses a custom register for metrics without interference and ensures that the registration process is executed only once when the program is running.

Fixes: #5255

### DIFF
--- a/src/agent/src/metrics.rs
+++ b/src/agent/src/metrics.rs
@@ -5,10 +5,11 @@
 
 extern crate procfs;
 
-use prometheus::{Encoder, Gauge, GaugeVec, IntCounter, TextEncoder};
+use prometheus::{Encoder, Gauge, GaugeVec, IntCounter, Opts, Registry, TextEncoder};
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use slog::warn;
+use std::sync::Mutex;
 use tracing::instrument;
 
 const NAMESPACE_KATA_AGENT: &str = "kata_agent";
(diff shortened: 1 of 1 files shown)


src/agent/src/metrics.rs
```diff
@@ -90,6 +103,59 @@ pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
     Ok(String::from_utf8(buffer)?)
 }
 
+#[instrument]
+fn register_metrics() {
+    REGISTRY
+        .register(Box::new(AGENT_SCRAPE_COUNT.clone()))
+        .expect(REGISTER_MSG);
```
bergwolf, line 110: Please let's handle errors instead of panicing here.

Yuan-Zhuo: Fixed

src/agent/src/metrics.rs
```diff
@@ -23,55 +25,70 @@ macro_rules! sl {
...
 
-    static ref     GUEST_MEMINFO: GaugeVec =
-    prometheus::register_gauge_vec!(format!("{}_{}",NAMESPACE_KATA_GUEST,"meminfo") , "Statistics about memory usage in the system.", &["item"]).unwrap();
+    static ref GUEST_MEMINFO: GaugeVec =
+    GaugeVec::new(Opts::new(format!("{}_{}",NAMESPACE_KATA_GUEST,"meminfo"), "Statistics about memory usage in the system."), &["item"]).unwrap();
 }
 
 #[instrument]
 pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
+    let handle = thread::spawn(|| {
```
bergwolf, line 83: Why do we need a new thread instead of handling it through tokio runtime?

Ji-Xinyou: I think @bergwolf means that you should use
```rust
tokio::spawn(...)
```
instead of 
```rust
thread::spawn(...)
```

Ji-Xinyou: I have tested `tokio::task::spawn`, which is more light-weight, finding that this will **not** cause the main thread to exit.
The example code is as follows
```rust
  1 use tokio::time::{self, Duration};
  2
  3 #[tokio::main]
  4 async fn main() {
  5 │   tokio::task::spawn(async move {
  6 │   │   println!("inside tokio spawn");
  7 │   │   panic!("die");
  8 │   });
  9 │   time::sleep(Duration::from_secs(1)).await;
 10 │   println!("Hello, world!");
 11 }
```

The output will be 
```text
inside tokio spawn
thread 'tokio-runtime-worker' panicked at 'die', src/main.rs:7:9
Hello, world!
```

Yuan-Zhuo: Thanks, I updated the PR by tokio::task::spawn.

src/agent/src/metrics.rs
```diff
@@ -23,55 +25,70 @@ macro_rules! sl {
...
 
 #[instrument]
-pub fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
+pub async fn get_metrics(_: &protocols::agent::GetMetricsRequest) -> Result<String> {
+    let handle_init = tokio::task::spawn(async move {
+        INIT_REGISTER.call_once_force(|_| {
+            register_metrics().unwrap();
+        });
+    });
+    if handle_init.await.is_err() {
```
Tim-0731-Hzt, line 88: Nit:
```rust
tokio::task::spawn(async move {
    INIT_REGISTER.call_once_force(|_| {
        register_metrics().unwrap();
    });
})
.await
.map_err(|_| anyhow!("failed to init register"))?;
```

Yuan-Zhuo: Fixed

Ji-Xinyou (approved):
Generally LGTM, thanks @Yuan-Zhuo.

Ji-Xinyou (approved):
Generally LGTM, thanks @Yuan-Zhuo.

bergwolf (approved):
lgtm! Thanks!

liubin (approved):

Issue #5261, opened 2022-09-28
stable-2.5: cargo vendor failure


bergwolf:
See https://github.com/kata-containers/kata-containers/actions/runs/3137146040/jobs/5094941441

After `cargo vendor`, we have diff in the tree:
```
./hack/tree_status.sh
tree is dirty, please commit all changes

 M src/agent/Cargo.lock
diff --git a/src/agent/Cargo.lock b/src/agent/Cargo.lock
index c970a3c69..0f4cf6b9d 100644
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -822,6 +822,17 @@ dependencies = [
make[1]: *** [Makefile:672: vendor] Error 1
make: *** [Makefile:33: vendor-runtime] Error 2
  "memoffset",
 ]
 
+[[package]]
+name = "nix"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "195cdbc1741b8134346d515b3a56a1c94b09127[580](https://github.com/kata-containers/kata-containers/actions/runs/3137146040/jobs/5094941441#step:10:581)09cfd53f99ea0f57b065fc"
+dependencies = [
+ "bitflags",
+ "cfg-if 1.0.0",
+ "libc",
+]
+
 [[package]]
 name = "ntapi"
 version = "0.3.7"
make[1]: Leaving directory '/home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime'
Error: Process completed with exit code 2.
```

PR #5262, opened 2022-09-28
stable-2.5: fix cargo vendor

bergwolf:
Looks like cargo vendor searches for all needed dependencies and we have left over nix 0.24.2

Fixes: #5261

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -823,6 +823,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "nix"
+version = "0.24.2"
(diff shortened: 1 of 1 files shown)


snir911:
sorry, closed by mistake

snir911 (approved):

snir911:
@bergwolf static checks failed on [test_find_process](https://github.com/kata-containers/kata-containers/actions/runs/3141067813/jobs/5103104658), re-run seems to pass, however, does it ring a bell? i wonder if we missed something

bergwolf:
@snir911 Didn't see it before. But I wonder if the EBADF failure is related to tokio.

PR #5267, opened 2022-09-28
# Kata Containers 3.0.0-rc1

gkurz:
- kata-deploy: support runtime-rs for kata deploy
- Last backport for 3.0-rc1
- stable-3.0: backport runtime/runtime-rs dependency updates

9afb8e5bac6a release: Adapt kata-deploy for 3.0.0-rc1
b0c5f040f02f runtime-rs: set agent timeout to 0 for stream RPCs
d44e39e0592c runtime-rs: fix incorrect comments
43b0e95800f6 runtime: store the user name in hypervisor config
81801888a29f runtime: make StopVM thread-safe
fba39ef32d03 runtime: add more debug logs for non-root user operation
63309514cacc runtime-rs: drop dependency on rustc-serialize
e229a03cc814 runtime: update runc dependency
d663f110d737 kata-deploy: get the config path from cri options
c6b3dcb67d5f kata-deploy: support kata-deploy for runtime-rs
a394761a5cc7 kata-deploy: add installation for runtime-rs


bergwolf (approved):

bergwolf:
https://github.com/kata-containers/kata-containers/actions/runs/3142238852

tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
```diff
@@ -18,7 +18,7 @@ spec:
           katacontainers.io/kata-runtime: cleanup
       containers:
       - name: kube-kata-cleanup
-        image: quay.io/kata-containers/kata-deploy:latest
+        image: quay.io/kata-containers/kata-deploy:3.0.0-rc1
```
bergwolf, line 21: Looks like we should not change this by hand. Instead, the `test-kata.sh` script changes it automatically for stable releases. While for non-stable releases, we are using `latest` here. Are the diffs generated by script?

bergwolf: I think that's why https://github.com/kata-containers/kata-containers/actions/runs/3142238852 fails, because we do not have `:3.0.0-rc1` kata-deploy image yet (which is created when the new tag is pushed).

gkurz: > Looks like we should not change this by hand. Instead, the `test-kata.sh` script changes it automatically for stable releases. While for non-stable releases, we are using `latest` here. Are the diffs generated by script?

Yes all the changes in this PR were generated by the following scriplet without any further action from my part:
```
$ export NEW_VERSION=3.0.0-rc1
$ export BRANCH=stable-3.0
$ ./update-repository-version.sh -p "$NEW_VERSION" "$BRANCH"
```

gkurz: > I think that's why https://github.com/kata-containers/kata-containers/actions/runs/3142238852 fails, because we do not have `:3.0.0-rc1` kata-deploy image yet (which is created when the new tag is pushed).

It seems you're right but how did we get there ?!?

gkurz: This is supposedly fixed by https://github.com/kata-containers/kata-containers/pull/5284

gkurz:
The release script doesn't update the PR description. I'll just close this one and let the script create a new one.

Issue #5275, opened 2022-09-29
readme.md: no need to mention general libraries


bergwolf:
There are duplicated items while we do not need to mention it at all. See https://github.com/kata-containers/kata-containers#main-components

PR #5276, opened 2022-09-29
readme: remove libraries mentioning

bergwolf:
There are two duplicated mentioning of the rust libraries in README.md. Let's just remove them all as the section is intended to list out core Kata components rather than general libraries.

Fixes: #5275


liubin (approved):

liubin:
Static check failed due to:

> 2022-09-29T03:08:39.8470038Z ERROR: Document src/libs/README.md is not referenced

The `src/libs` may be moved from `Main components` to `Additional components`.

GabyCT (approved):

Issue #5283, opened 2022-09-29
Bogus check in "tools/packaging/release/update-repository-version.sh"


gkurz:
While rolling out 3.0.0-rc1, I ended up in a situation where the version bump PR could not be merged because the required `kata-deploy` CI check was failing. The failure happens because the test tries to pull the `quay.io/kata-containers/kata-deploy:3.0.0-rc1` image which only gets available after the release was cut out (chicken and egg problem).

The `tools/packaging/release/update-repository-version.sh` script is indeed generating an extra commit to fix versions:

```
-        image: quay.io/kata-containers/kata-deploy:latest
+        image: quay.io/kata-containers/kata-deploy:3.0.0-rc1
```

This isn't the expected behavior : an -rc release should continue to rely on the `latest` tag to avoid the chicken and egg problem.

Debugging points to this check:

```
		elif [ "${new_version}" != *"rc"* ]; then
```

where `*"rc"*` gets expanded to the `src` directory, which is clearly unintended.

Check should be `[[ ! "${new_version}" =~ rc ]]` actually.

PR #5286, opened 2022-09-29
tools: release: fix bogus version check

gkurz:
Shell expands `*"rc"*` to the top-level `src` directory. This results in comparing a version with a directory name. This doesn't make sense and causes the script to choose the wrong branch of the `if`.

The intent of the check is actually to detect `rc` in the version.

Fixes: #5283
(cherry picked from commit 421729f99136522a76ce199947a0a197a1d07801)

### DIFF
--- a/tools/packaging/release/update-repository-version.sh
+++ b/tools/packaging/release/update-repository-version.sh
@@ -191,7 +191,7 @@ bump_repo() {
 
 				need_commit=true
 			fi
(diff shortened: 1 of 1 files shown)


littlejawa (approved):

c3d (approved):
/lgtm. Thanks @gkurz

wainersm (approved):
Shell script can be really dangerous sometimes :)

Issue #5287, opened 2022-09-29
dragonball: runtime-rs build fail on Arm


jongwu:
the error report when build runtime-rs on Arm:

> 12:01:00 error[E0599]: no method named `as_str` found for reference `&Cmdline` in the current scope
12:01:01    --> /home/jenkins/workspace/kata-containers-2.0-ubuntu20.04-ARM-unit/go/src/github.com/kata-containers/kata-containers/src/dragonball/src/vm/aarch64.rs:139:21
12:01:01     |
12:01:01 139 |             cmdline.as_str(),
12:01:01     |                     ^^^^^^ help: there is an associated function with a similar name: `as_cstring`

studychao:
Thanks!!

PR #5288, opened 2022-09-29
dragonball: fix no "as_str" error on Arm

jongwu:
Cmdline struct update in the latest linux-loader lib and its as_str method is changed to as_cstring, thus we need fix it according whereas the old as_str method is used.

Fixes: #5287


studychao:
Thanks!

studychao (approved):

Ji-Xinyou (approved):

gkurz (approved):
LGTM. Thanks @jongwu !

PR #5295, opened 2022-09-30
Introduce container-monitor: expose more container stats

egernst:
## Overview:

    container-monitor: Introduction of tool for gathering conatiner stats

    container-monitor is created to facilitate exposing container specific
    stats which would typically be exposed by CADvisor in a kubernetes
    environment. These are already available via task service API today, but
    are not exposed through CRI interface.

    Because of this, we'll introduce a tool that will present a prometheus
    endpoint, and a Collector for gathering stats. Each time Collect is
    invoked, we will get a list of containers from containerd, and then
    connect to each container's shim in order to get stats. We then present
    these metrics with the labels namespace, container-name and pod-name.

    These labels are what a kubernetes user would see (ie, what's presented
    "north" of CRI).

    Usage:

    $ ./container-monitor --help
    NAME:
       container-monitor - A new cli application

    USAGE:
       container-monitor [global options] command [command options]
    [arguments...]

    COMMANDS:
       help, h  Shows a list of commands or help for one command

    GLOBAL OPTIONS:
       --address value, -a value      path to the containerd socket (default: "/run/containerd/containerd.sock")
       --namespace value, --ns value  the namespace to get container stats from (default: "default")
       --server-port value, -p value  The address the server listens on for HTTP requests. (default: "8090")
       --log-level value              Logging level (trace/debug/info/warn/error/fatal/panic). (default: "warn")
       --help, -h                     show help

## to do:

- add documentation
- scrub metric naming to match existing endpoints (cadvisor?)
- open issue to introduce
- add code attributions
- consider moving to a separate repository in kata org?


zvonkok:
I haven't dug into the code but have a general question, how does this compare to: https://github.com/containers/conmon?

fidencio:
/cc @littlejawa @haircommander 

Eric, do you see this as a "generic" enough tool that could also be used with CRI-O?

Also, we do have a KEP open to have those exposed in the CRI, which was raised by Peter Hunt.

haircommander:
> I haven't dug into the code but have a general question, how does this compare to: https://github.com/containers/conmon?

scope of these two is different: conmon is for monitoring a container at runtime to write logs and handle exits. this is for exposing prom metrics



> /cc @littlejawa @haircommander
> 
> Eric, do you see this as a "generic" enough tool that could also be used with CRI-O?

I wouldn't. It uses the task API, which CRI-O doesn't support.

egernst:
Well, CRIO does support the task API -- the gap I see is that I'm using the containerd client to get the list of containers. Either way -- I don't see this as being very generic, and I see this as a stop gap until an alternative to cadvisor is utilized in kubelet.  There are KEPs that are WIP in this space, but it'll be a while before these kinds of metrics are made available in the case of sandboxed runtimes.

Issue #5316, opened 2022-10-03
Shim: Pagesize not set correctly for hugetlb stats


Champ-Goblem:
The containerd stats method and metrics API are broken with Kata 2.5.x, the stats fail to load and the metrics API responds with an error:
```
* collected metric "container_hugetlb_usage_bytes" { label:<name:"container_id" value:"fd195eaa2d63604896d2b0c5c851eb1ca29f9a02379c10730fedabfa1f39cba8" > label:<name:"namespace" value:"k8s.io" > label:<name:"page" value:"" > label:<name:"runtime" value:"io.containerd.kata.v2" > gauge:<value:0 > } was collected before with the same name and label values
* collected metric "container_hugetlb_failcnt_total" { label:<name:"container_id" value:"fd195eaa2d63604896d2b0c5c851eb1ca29f9a02379c10730fedabfa1f39cba8" > label:<name:"namespace" value:"k8s.io" > label:<name:"page" value:"" > label:<name:"runtime" value:"io.containerd.kata.v2" > gauge:<value:0 > } was collected before with the same name and label values
* collected metric "container_hugetlb_max_bytes" { label:<name:"container_id" value:"fd195eaa2d63604896d2b0c5c851eb1ca29f9a02379c10730fedabfa1f39cba8" > label:<name:"namespace" value:"k8s.io" > label:<name:"page" value:"" > label:<name:"runtime" value:"io.containerd.kata.v2" > gauge:<value:0 > } was collected before with the same name and label values
```

This seems to be down to the conversion from the stats reported by the agent RPC `StatsContainer` where the field `Pagesize` is not completed by the `setHugetlbStats` method. In the case where multiple sized table stats are reported, this causes containerd to register two metrics with the same label set, rather than each being partitioned by the `page` label, causing the error seen above.

This can be fixed by ensuring the `Pagesize` field is set when converting the agent hugetable stats

PR #5317, opened 2022-10-03
shim: Ensure pagesize is set when reporting hugetlb stats

Champ-Goblem:
The containerd stats method and metrics API are broken with Kata 2.5.x, the stats fail to load and the metrics API responds with status code 500

This seems to be down to the conversion from the stats reported by the agent RPC `StatsContainer` where the field `Pagesize` is not completed by the `setHugetlbStats` method. In the case where multiple sized tables stats are reported, this causes containerd to register two metrics with the same label set, rather than each being partitioned by the `page` label.

Fixes: #5316

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/metrics.go
+++ b/src/runtime/pkg/containerd-shim-v2/metrics.go
@@ -51,13 +51,14 @@ func statsToMetrics(stats *vc.ContainerStats) *cgroupsv1.Metrics {
 
 func setHugetlbStats(vcHugetlb map[string]vc.HugetlbStats) []*cgroupsv1.HugetlbStat {
 	var hugetlbStats []*cgroupsv1.HugetlbStat
-	for _, v := range vcHugetlb {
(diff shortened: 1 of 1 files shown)


fidencio:
@Champ-Goblem, the commit message has to be adjusted as the lines are too long.
Apart from this, lgtm!  This should also be backported to the stable-2.5 branch once merged here.

Champ-Goblem:
I think all it was missing was the commit being signed off which I've updated now, should be below the 150 line limit

fidencio:
@Champ-Goblem, would be possible to add some tests for this?  Expanding https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/metrics_test.go would be a good way to do so.

liubin (approved):

liubin:
thank you @Champ-Goblem

fidencio (approved):
lgtm, thanks @Champ-Goblem!

Issue #5077, opened 2022-09-01
`make kata-tarball` doesn't work behind a proxy


fidencio:
We've figured out that `make kata-tarball` doesn't work behind proxy.

For various targets https_proxy / http_proxy / no_proxy is not exported to the container where the build actually happens.
This is a blocker for switching our CIs to test exactly the same builds we'd ship, mainly considering some of our CIs may run behind a firewall, using proxy. :-(

PR #5323, opened 2022-10-04
packaging: Mount $HOME/.docker in the 1st layer container

fidencio:
In order to ensure that the proxy configuration is passed to the 2nd layer container, let's ensure the $HOME/.docker/config.json file is exposed inside the 1st layer container.

For some reason which I still don't fully understand exporting https_proxy / http_proxy / no_proxy was not enough to get those variables exported to the 2nd layer container.

In this commit we're creating a "$HOME/.docker" directory, and removing it after the build, in case it doesn't exist yet.  The reason we do this is to avoid docker not running in case "$HOME/.docker" doesn't exist.

This was not tested with podman, but if there's an issue with podman, the issue was already there beforehand and should be treated as a different problem than the one addressed in this commit.

Fixes: #5077

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -31,6 +31,12 @@ if [ ${docker_gid} == ${gid} ]; then
 	docker_gid=""
 fi
 
+remove_dot_docker_dir=false
+if [ ! -d "$HOME/.docker" ]; then
+	mkdir $HOME/.docker
+	remove_dot_docker_dir=true
+fi
+
 docker build -q -t build-kata-deploy \
 	--build-arg IMG_USER="${USER}" \
 	--build-arg UID=${uid} \
@@ -39,9 +45,14 @@ docker build -q -t build-kata-deploy \
 	"${script_dir}/dockerbuild/"
 
 docker run \
+	-v $HOME/.docker:/root/.docker \
 	-v /var/run/docker.sock:/var/run/docker.sock \
 	--env CI="${CI:-}" \
 	--env USER=${USER} -v "${kata_dir}:${kata_dir}" \
 	--rm \
 	-w ${script_dir} \
 	build-kata-deploy "${kata_deploy_create}" $@
+
+if [ $remove_dot_docker_dir == true ]; then
+	rm -rf "$HOME/.docker"
+fi


fidencio:
I've tested this in a machine behind proxy, the same machine used for the TDX Ci, and it works as expected.

stevenhorsman (approved):
LGTM. Thanks!

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
```diff
@@ -39,9 +41,12 @@ docker build -q -t build-kata-deploy \
...
 docker run \
+	-v $HOME/.docker:/root/.docker \
 	-v /var/run/docker.sock:/var/run/docker.sock \
 	--env CI="${CI:-}" \
 	--env USER=${USER} -v "${kata_dir}:${kata_dir}" \
 	--rm \
 	-w ${script_dir} \
 	build-kata-deploy "${kata_deploy_create}" $@
+
+rmdir --ignore-fail-on-non-empty "$HOME/.docker"
```
wainersm, line 52: `$HOME/.docker` doesn't exist then it is created in line 34. Suppose that on 2nd container some file get created in that directory, thus, this line won't remove  `$HOME/.docker` (but it should). Sorry I am being a little paranoid here, this situation might not happen.... but... you never know for sure :)

fidencio: I'm not aware of any docker command that would create a file in that directory, particularly considering the use-cases we have.

What's your preference?  Add a new flag together with the directory creation and then remove based on the flag?

fidencio:
```sh
diff --git a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
index 038c917cf..b98434fa5 100755
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -31,7 +31,11 @@ if [ ${docker_gid} == ${gid} ]; then
        docker_gid=""
 fi
 
-[ -d "$HOME/.docker" ] || mkdir $HOME/.docker
+remove_dot_docker_dir=false
+if [ ! -d "$HOME/.docker" ]; then
+       mkdir $HOME/.docker
+       remove_dot_docker_dir=true
+fi
 
 docker build -q -t build-kata-deploy \
        --build-arg IMG_USER="${USER}" \
@@ -49,4 +53,6 @@ docker run \
        -w ${script_dir} \
        build-kata-deploy "${kata_deploy_create}" $@
 
-rmdir --ignore-fail-on-non-empty "$HOME/.docker"
+if [ $remove_dot_docker_dir ]; then
+       rm -rf "$HOME/.docker"
+fi
```

wainersm: Yes, I thought about using a flag.
`+if [ $remove_dot_docker_dir ]; then` does shell eval this correctly if remove_dot_docker_dir is false/true (instead of 0 or 1)?

fidencio:
```
+if [ $remove_dot_docker_dir == true ]; then
+       rm -rf "$HOME/.docker"
+fi
```

This is the correct check :-)

fidencio: Let me update the PR

wainersm (approved):
Thanks for addressing my comment @fidencio ! LGTM.

PR #5324, opened 2022-10-04
CC | packaging: Mount $HOME/.docker in the 1st layer container

fidencio:
In order to ensure that the proxy configuration is passed to the 2nd layer container, let's ensure the $HOME/.docker/config.json file is exposed inside the 1st layer container.

For some reason which I still don't fully understand exporting https_proxy / http_proxy / no_proxy was not enough to get those variables exported to the 2nd layer container.

In this commit we're creating a "$HOME/.docker" directory, and removing it after the build, in case it doesn't exist yet.  The reason we do this is to avoid docker not running in case "$HOME/.docker" doesn't exist.

This was not tested with podman, but if there's an issue with podman, the issue was already there beforehand and should be treated as a different problem than the one addressed in this commit.

Fixes: #5077

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -31,6 +31,12 @@ if [ ${docker_gid} == ${gid} ]; then
 	docker_gid=""
 fi
 
+remove_dot_docker_dir=false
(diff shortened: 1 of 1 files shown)


fidencio:
Adding the `do-not-merge` label till the PR raised for `main` gets merged.

fidencio:
I've added a `Depends-on` on this PR so we can test https://github.com/kata-containers/tests/pull/5168

Issue #5364, opened 2022-10-06
Clean up stale storage object reference for failed mount


fengwang666:
In the kata agent [add_storages](https://github.com/kata-containers/kata-containers/blob/63959b0be6f4775eb8279b1e901a238b4a47bf30/src/agent/src/mount.rs#L723) method, if the storage mount fails for some reason (e.g. filesystem was not created beforehand), the container creation fails but a retry would succeed because the failed storage is considered [not new and skipped](https://github.com/kata-containers/kata-containers/blob/63959b0be6f4775eb8279b1e901a238b4a47bf30/src/agent/src/mount.rs#L738-L741). We need to update the logic to remove the reference if the mount failed.

PR #5365, opened 2022-10-07
agent: reduce reference count for failed mount

fengwang666:
The kata agent adds a reference for each storage object before mount and skip mount again if the storage object is known. We need to remove the object reference if mount fails.

Fixes: #5364


bergwolf (approved):
lgtm!

liubin (approved):

Issue #5366, opened 2022-10-07
container termination bug in kata agent 


fengwang666:
When kata agent [signals](https://github.com/kata-containers/kata-containers/blob/cbe5e324ae3ed6d7ea4c12aa4e49fe6b67f1b13d/src/agent/src/rpc.rs#L370) the container process to exit, signal might encounter ESRCH and result in an [early exit ](https://github.com/kata-containers/kata-containers/blob/cbe5e324ae3ed6d7ea4c12aa4e49fe6b67f1b13d/src/agent/src/rpc.rs#L393). This is problematic as all the logic after line [396](https://github.com/kata-containers/kata-containers/blob/cbe5e324ae3ed6d7ea4c12aa4e49fe6b67f1b13d/src/agent/src/rpc.rs#L396-L440) are skipped and the shim process can leak in the scenario described in https://github.com/kata-containers/kata-containers/issues/3913

PR #5367, opened 2022-10-07
agent: don't exit early if signal fails due to ESRCH

fengwang666:
ESRCH usually means the process has exited. In this case, the execution should continue to kill remaining container processes.

Fixes: #5366

### DIFF
--- a/src/agent/rustjail/Cargo.toml
+++ b/src/agent/rustjail/Cargo.toml
@@ -12,7 +12,7 @@ serde_derive = "1.0.91"
 oci = { path = "../../libs/oci" }
 protocols = { path ="../../libs/protocols" }
 caps = "0.5.0"
-nix = "0.23.0"
+nix = "0.24.2"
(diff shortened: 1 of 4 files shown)


src/agent/rustjail/Cargo.toml
```diff
@@ -12,7 +12,7 @@ serde_derive = "1.0.91"
 oci = { path = "../../libs/oci" }
 protocols = { path ="../../libs/protocols" }
 caps = "0.5.0"
-nix = "0.23.0"
+nix = "0.24.2"
```
bergwolf, line 15: The `nix` update is causing compilation errors.

bergwolf (approved):

bergwolf:
`jenkins-ci-ppc64le-ubuntu-20-04` ci job is stuck and we can ignore it for now.

lifupan (approved):

PR #5371, opened 2022-10-07
Do not merge: static-checks test

stevenhorsman:
Test PR to check if the intel link fails the static checks on main


liubin:
Yes, it seems that accessing intel.com will get 403.

stevenhorsman:
The static-checks have both passed now, so I think https://github.com/kata-containers/tests/pull/5186 did the trick!

Issue #5373, opened 2022-10-07
osbuilder: specify Attestation Agent tag for rootfs build


Alex-Carter01:
Now that the AA has a 0.1.0 tag, we can specify this in the versions file, and select it when building for the rootfs.

PR #5374, opened 2022-10-07
osbuilder: specify Attestation Agent tag for rootfs build

Alex-Carter01:
Add tag entry to the attestation agent entry of the versions file. Checkout tag commit after cloning AA in rootfs builder.

Fixes: #5373
Fixes: kata-containers#5373
Signed-Off-By: Alex Carter <alex.carter@ibm.com>

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -177,7 +177,7 @@ externals:
   attestation-agent:
     description: "Provide attested key unwrapping for image decryption"
     url: "https://github.com/confidential-containers/attestation-agent"
-    branch: "main"
+    version: "e930d362c24bad62b5fa7d5079a9dc912dd78fa4"
 
   cni-plugins:
     description: "CNI network plugins"
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -670,10 +670,13 @@ EOF
 			AA_KBC_PARAMS="offline_sev_kbc::null" envsubst < "${script_dir}/agent-config.toml.in" | tee "${ROOTFS_DIR}/etc/agent-config.toml"
 		fi
 		attestation_agent_url="$(get_package_version_from_kata_yaml externals.attestation-agent.url)"
-		attestation_agent_branch="$(get_package_version_from_kata_yaml externals.attestation-agent.branch)"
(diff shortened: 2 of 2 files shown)


fidencio (changes requested):
@Alex-Carter01, I've left a few suggestions, please, let me know what do you think about those.

Alex-Carter01:
Removed branch information and switched to directly cloning the tag

versions.yaml
```diff
@@ -177,7 +177,7 @@ externals:
   attestation-agent:
     description: "Provide attested key unwrapping for image decryption"
     url: "https://github.com/confidential-containers/attestation-agent"
-    branch: "main"
+    tag: "v0.1.0"
```
stevenhorsman, line 180: I think the model from the other components in versions.yaml is to call it `versions:` and then it can be a release, or commit. As I mentioned in slack, I think ‘pinning’ the attestation-agent rather than just pulling `main` is a good  start, but my concern about just using a tag is when would the tag be created & bumped? If it isn’t until the release we won’t be able to do any integration tests of new features, whereas using a commit would let us pull in a version of our choice?

For example right now Ding has added cosign support into attestation-agent, which we'd ideally test with kata as soon as possible, but it isn't in the tagged version. Instead we can pull in the commit (which would trigger the kata-payload as a bonus) and check that it works, then that test would like us judge whether that commit is eligible to be tagged for the next release. As we get to the release date, then tagging a commit and swapping the version.yaml reference to the tag sounds fine, but after code freeze ends we'd probably want to start picking up new features again.
Does this make sense?

Alex-Carter01:
Whoops never commented. Switched to version where you should be able to specify a tag or commit hash

versions.yaml
```diff
@@ -177,7 +177,7 @@ externals:
   attestation-agent:
     description: "Provide attested key unwrapping for image decryption"
     url: "https://github.com/confidential-containers/attestation-agent"
-    branch: "main"
+    version: "v0.1.0"
```
stevenhorsman, line 180: Thanks for the update! Can I request we try using version `e930d362c24bad62b5fa7d5079a9dc912dd78fa4` at that is the current latest in main and I don't want us to be missing function going back to the old tag?

Alex-Carter01: Sure, good catch :), updated to the hash.

stevenhorsman (approved):
LGTM. Thanks!

PR #5377, opened 2022-10-08
stable-3.0: backport agent fixes

bergwolf:
* agent: reduce reference count for failed mount
* agent: don't exit early if signal fails due to ESRCH


liubin (approved):

fidencio (approved):
lgtm, thanks @bergwolf!

Issue #5385, opened 2022-10-09
Built-in Sandbox: Add Dragonball build, clippy and format into CI


studychao:
Add Dragonball build, clippy and format into CI.

Due to issue #5378, we skip Dragonball UT test. After #5378 is solved, we will add UT test back.

Issue #5378, opened 2022-10-08
Build-in Sandbox: Enhance Static-Check of CI to support nested virtualization. 


wllenyj:
Currently, Dragonball's Unit-Testing has not been added to the static checking. The pr #4961 added Dragonball components to Makefile. But [the Static checks / test](https://github.com/kata-containers/kata-containers/actions/runs/3107229536/jobs/5035089044) fail. The reason is, there is no /dev/kvm. 
So we should enhance Static-Check of CI to support nested virtualization. 

### Options
1. Use [self-hosted-runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) for `Static Check`.
   Pros: Change small, This scheme has been used in [Dragonball-sandbox](https://github.com/openanolis/dragonball-sandbox/blob/main/.github/workflows/test.yaml), and the same scheme and machine can be reused.

2. Move the Unit-Testing of the Dragonball to `jenkins-ci` from the `Github Actions`.
   Pros: More unified management of machines.

### Details for Option 1
1. We should add a New Runner in `kata-containers` repository. See https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners#adding-a-self-hosted-runner-to-a-repository
2. Download the runner application and install it on your self-hosted runner machine.
3. Modify your workflow file to use `self-hosted runners`. See https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow. 

### Details for Option 2
1. Remove the `test` in Makefile of Dragonball.
2. Add `cargo test` after `cargo build` for Dragonball in `kata-containers/tests` repo. (or some other location)

wllenyj:
@bergwolf @jon @GabyCT Any suggestions? Thanks.

bergwolf:
Looks like the self-hosted runner is pretty clean that it doesn't require any PAT to integrate with, am I understanding it correctly?

wllenyj:
> Looks like the self-hosted runner is pretty clean that it doesn't require any PAT to integrate with, am I understanding it correctly?

I think so, it looks more clean.

PR #4961, opened 2022-08-22
Built-in Sandbox: add more unit tests for dragonball

wllenyj:
Added more unit test for vmm actions.
Added dragonball components to Makefile.
Fixed clippy warning.

PR #5386, opened 2022-10-09
Dragonball: Add Dragonball build, clippy and format into CI

studychao:
Add Dragonball build, clippy and format into CI.

Due to issue #5378, we skip Dragonball UT test. After #5378 is solved, we will add UT test back.

fixes: #5385


studychao:
close this since it's duplicated with #4961

Issue #4499, opened 2022-06-21
kata-ctl: Rewrite the kata-runtime utility program in rust


jodh-intel:
## TL;DR

Tracking project: https://github.com/orgs/kata-containers/projects/34/views/1

## Background

For Kata 2.x, we provide the [`kata-runtime` utility program](https://github.com/kata-containers/kata-containers/blob/main/docs/design/architecture/README.md#utility-program). Like the host side of Kata 2.x, this is written in golang.

## Upcoming changes

When the [`runtime-rs` PRs](https://github.com/kata-containers/kata-containers/pulls?q=is%3Apr+is%3Aopen+label%3Aruntime-rs) land in preparation for Kata 3.x, all components will be written in rust... except for the utility program.

## Rationale

We don't want to have to rely on "yet another toolchain" when we switch to a (mostly) pure rust environment:

- Extra complexity.
- Will increase the time to run the CI even further.

## Do we still need the utility program?

Yes, since it provides a number of useful features for users, admins and developers:

```bash
$ kata-runtime | sed -e '/./{H;$!d;}' -e 'x;/COMMANDS:/!d;'

COMMANDS:
   version            display version details
   check, kata-check  tests if system can run Kata Containers
   env, kata-env      display settings. Default to TOML
   exec               Enter into guest by debug console
   metrics            gather metrics associated with infrastructure used to run a sandbox
   factory            manage vm factory
   direct-volume      directly assign a volume to Kata Containers to manage
   iptables           get or set iptables within the Kata Containers guest
   help, h            Shows a list of commands or help for one command
```

> **Notes:**
>
> - `check` is semi-essential for users to allow them to determine if their hardware is capable of running Kata Containers.
> - `env` is very useful for debugging.
> - `exec` is an essential debug and development tool.

## Plan

- Rewrite [`kata-runtime`](https://github.com/kata-containers/kata-containers/tree/main/src/runtime/cmd/kata-runtime) in rust.
- Rename it `kata-ctl`.
  The current name is historical and arguably confusing for new users as it appears to be the runtime, but it isn't (see https://github.com/kata-containers/kata-containers/blob/main/docs/design/architecture/README.md#utility-program for further details on this).

## Thoughts

We _could_ consider merging `kata-runtime`/`kata-ctl` with the [`kata-agent-ctl`](https://github.com/kata-containers/kata-containers/tree/main/src/tools/agent-ctl) tool.

### Advantages

[...]

Issue #5351, opened 2022-10-06
kata-ctl: docs: Write some


jodh-intel:
The `kata-ctl` tool needs documentation.

**Question:**

Should the doc ask the user to run with `--help`, like the [`agent-ctl`](https://github.com/kata-containers/kata-containers/tree/main/src/tools/agent-ctl#full-details) docs specify? That means the doc won't "bit rot", but it also means that users are less likely to even know that the utility provides commands to do _x_, _y_ and _z_. As such, I think we should consider "fully documenting" the tool (and trying to remember to ask the developer to update the docs when PRs add new sub-commands / options / change behaviour.

Issue #5334, opened 2022-10-06
Add kata-ctl tool to main repo


jodh-intel:
As shown on https://github.com/kata-containers/kata-containers/issues/4499#issuecomment-1211769717, we've been working on a rust rewrite of the `kata-runtime` command in a fork. Now that the tool is starting to take shape, we need to add it to the main repo to make it easier for the community to start contributing.

I plan to raise a PR on `main` once https://github.com/jodh-intel/kata-containers/pull/10 has landed in my branch.

jodh-intel:
Fixed by #5388.

PR #5388, opened 2022-10-10
kata-ctl: Move development to main branch

jodh-intel:
As shown on https://github.com/kata-containers/kata-containers/issues/4499#issuecomment-1211769717, we've been developing a replacement for the `kata-runtime` utility command, written in rust.

We'd now like the community to get actively involvement in this project, so this PR is a roll-up of the PRs developed on a separate branch.

**_Caveat emptor_**: `kata-ctl` is still a work in progress. This PR is simply moving the code into the `main` branch, so that we can accelerate development. `kata-ctl` is **not** yet a replacement for the existing golang [`kata-runtime` utility command](https://github.com/kata-containers/kata-containers/blob/main/docs/design/architecture/README.md#utility-program).

For further details, see:

- [The original issue](https://github.com/kata-containers/kata-containers/issues/4499#issuecomment-1211769717)
- [The `kata-ctl` GitHub tracking project](https://github.com/orgs/kata-containers/projects/34/views/1)

  If you'd like to get involved, tal at the issues in the project and assign yourself or add a comment! :smile:

Fixes: #4499, #5334, #5351.


cmaf (commented):
@jodh-intel I'm approving the README and docs since I previously reviewed everything else, but I'll leave the official approval to others

fidencio (commented):
The code doesn't do much nowadays and the main intent of the PR is actually to have this into the main branch to have the development done there.

I've left a few questions about the version of the package / tool as it's rather confusing for me.

Ji-Xinyou:
Generally nice work and LGTM! This would be nice for interaction between runtime and kata-ctl.
However, the version problem @fidencio mentioned indeed needs some discussion.

src/tools/kata-ctl/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+[package]
+name = "kata-ctl"
+version = "0.0.1"
```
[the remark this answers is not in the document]
jodh-intel: This is just standard rust manifest versioning. If you run the tool, it shows something like this:

```
$ kata-ctl
3.1.0-alpha0-383b7db704fe3919841fef047bbc70a3dd605516-0.0.1
```

Hence, the tool contains both the Kata version and it's own tooling version (and that full version string is valid semver). We can obviously change this, but as mentioned on the description of this PR, we just want to get this landed at this stage to allow more contributions :smile:

src/tools/kata-ctl/Makefile
```diff
@@ -0,0 +1,64 @@
...
+
+include ../../../utils.mk
+
+PROJECT_NAME = Kata Containers
+PROJECT_URL = https://github.com/kata-containers
+PROJECT_COMPONENT = kata-ctl
+
+TARGET = $(PROJECT_COMPONENT)
+
+VERSION_FILE := ./VERSION
```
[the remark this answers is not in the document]
jodh-intel: See above.

src/tools/kata-ctl/README.md
```diff
@@ -0,0 +1,49 @@
...
+## Build the tool
+
+```bash
+$ make
+```
+
+## Install the tool
+
+```bash
+$ make install
```
liubin, line 27: Don't need `sudo`? Normally `$` means a non-root user.

liubin (approved):

liubin:
@jodh-intel Shouldn't add `kata-ctl` to the top-level Makefile?

https://github.com/kata-containers/kata-containers/blob/ffdd7e1ad8da4dff39949639601ad95ecf7f44c9/Makefile#L18-L21

jodh-intel:
Good catch @liubin! Fixed :smile:

fidencio (approved):
lgtm, thanks!

src/tools/kata-ctl/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+[package]
+name = "kata-ctl"
+version = "0.0.0"
+authors = ["The Kata Containers community <kata-dev@lists.katacontainers.io>"]
+edition = "2021"
```
wainersm, line 10: Hi @jodh-intel ! Amazing initiative btw!

Regarding the rust edition, it seems all the other rust components are still on 2018 edition. Could the kata-ctl be built also on that edition?

src/tools/kata-ctl/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+fn real_main() -> Result<()> {
+    let name = crate_name!();
+
+    let app = App::new(name)
+        .version(crate_version!())
+        .about(ABOUT_TEXT)
+        .long_about(DESCRIPTION_TEXT)
+        .subcommand(
+            SubCommand::with_name("check")
+            .about("tests if system can run Kata Containers")
```
wainersm: Nit: s/tests/test to make it consistent with the other descriptions

src/tools/kata-ctl/src/arch/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[cfg(target_arch = "x86_64")]
+    let result = x86_64::check(global_args);
+
+    #[cfg(not(any(
+        target_arch = "aarch64",
+        target_arch = "powerpc64le",
+        target_arch = "s390x",
+        target_arch = "x86_64"
+    )))]
+    panic!("unknown architecture");
```
likebreath: For common functions that all architecture needs to provide and share the same signature, a nicer way is exposing these common functions directly from the arch-specific module with the right `cfg` guard, for example:
```
pub use x86_64::check;

#[cfg(target_arch = "aarch64")]
pub use aarch64::check;

...
```
In this way, the actual interface function of the crate (e.g. `arch::check()`) will be better abstracted and simpler. This will also report error at compile time for unsupported architectures (instead of a panic at runtime).

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -7,11 +7,47 @@
...
+        // TODO: Perform checks based on hypervisor type
+        // TODO: Add more information to output (see kata-check in go tool); adjust formatting
+        let missing_cpu_attributes = check::check_cpu_attribs(&cpu_info, CPU_ATTRIBS_INTEL)?;
+        if missing_cpu_attributes.len() > 0 {
+            eprintln!("WARNING: Missing CPU attributes {:?}", missing_cpu_attributes);
+        }
+        let missing_cpu_flags = check::check_cpu_flags(&cpu_flags, CPU_FLAGS_INTEL)?;
+        if missing_cpu_flags.len() > 0 {
+            eprintln!("WARNING: Missing CPU flags {:?}", missing_cpu_flags);
+        }
```
likebreath, line 41: The `log` crate is a good option for handling logging with different verbose levels, say 'warn', 'error', 'debug', etc. It can also be exposed by a CLI parameter for users to decide the logging level.

likebreath (commented):
Generally looking good to me. Couple of comments below.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
jodh-intel: @likebreath - thanks. I've raised an issue to use the Kata logging crate (which uses `slog`):

- https://github.com/kata-containers/kata-containers/issues/5401

src/tools/kata-ctl/src/arch/mod.rs
jodh-intel: Hi @likebreath - I'm confused. Firstly, that panic line was replaced on commit 7c9f9a5a1dae3f9c2c16748b6d5af9fc134abe65 so I'm not sure why/how you were seeing that. Secondly, we're already using cfg guards. The code I think you might be referring to (now) is:

```rust
#[cfg(not(any(
    target_arch = "aarch64",
    target_arch = "powerpc64le",
    target_arch = "s390x",
    target_arch = "x86_64"
)))]
compile_error!("unknown architecture");
```

... and I think that does exactly what we want: it will fail to compile on unknown platforms.

jodh-intel: Ah - I see what you're saying about the cfg guard now. Must be time for :coffee: :smile:

src/tools/kata-ctl/README.md
jodh-intel: It currently installs to the cargo bin dir, so this is correct. However, I agree we should start thinking about a proper install so I've raised https://github.com/kata-containers/kata-containers/issues/5403.

PR #5407, opened 2022-10-12
image_rpc: Make `init_attestation_agent` fail gracefully

### LINKED ISSUE #5406: `init_attestation_agent` will panic the agent on failure
# Description of problem

`init_attestation_agent` may fail for a number of reasons, for example, `/tmp/ocicrypt_config.json` is inaccessible or storage is full, `/usr/local/bin/attestation-agent` doesn't exist or is inaccessible, there are not enough system resources (e.g., memory) to launch a new process, etc.

# Expected result

`pull_image` return an error if `init_attestation_agent` fails.

# Actual result

`pull_image` panics instead.

# Further information

This is because `init_attestation_agent` calls `unwrap()` to handle errors instead of, for example, propagating them to the caller.

wedsonaf:
Without this, a failure in `init_attestion_agent` would lead to a panic.

Fixes #5406

### DIFF
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -177,7 +177,7 @@ impl ImageService {
 
     // If we fail to start the AA, Skopeo/ocicrypt won't be able to unwrap keys
     // and container decryption will fail.
-    fn init_attestation_agent() {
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @wedsonaf!

PR #5400, opened 2022-10-12
image_rpc: avoid double and triple indirections

wedsonaf:
This also slightly improves readability by decluttering the function declaration and call site.

Signed-off-by: Wedson Almeida Filho <walmeida@microsoft.com>

PR #5410, opened 2022-10-13
agent: change `AGENT_CONFIG`'s lazy type to just `AgentConfig`

### LINKED ISSUE #5409: agent: simplify `AGENT_CONFIG` usage
Given that `AGENT_CONFIG` is never modified after initialisation and that it already has a `'static` lifetime, it doesn't need to be wrapped in `RwLock` and `Arc`.

wedsonaf:
Since it is never modified, it doesn't really need a lock of any kind. Removing the `RwLock` wrapper allows us to remove all `.read().await` calls when accessing it.

Additionally, `AGENT_CONFIG` already has a static lifetime, so there is no need to wrap it in a ref-counted heap allocation.

Fixes #5409

### DIFF
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -206,12 +206,12 @@ impl ImageService {
     async fn pull_image(&self, req: &image::PullImageRequest) -> Result<String> {
         env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
 
-        let https_proxy = &AGENT_CONFIG.read().await.https_proxy;
+        let https_proxy = &AGENT_CONFIG.https_proxy;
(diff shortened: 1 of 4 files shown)


fidencio:
@wedsonaf, please, rebase this PR atop of the CCv0 branch.

wedsonaf:
> @wedsonaf, please, rebase this PR atop of the CCv0 branch.

Done.

It was off of CCv0 to begin with, but there was a conflict after merging #5400.

fidencio:
@wedsonaf, I see the patch touches some code that is not specific to CC.  I'd advise to check whether this also affects the `main` branch and, if it does, open a PR there first and then we have it as part of the CC branch via the weekly rebases.

wedsonaf:
Merged into main as part of https://github.com/kata-containers/kata-containers/pull/7202

PR #5416, opened 2022-10-13
WIP: Update how-to-run-kata-containers-with-SNP-VMs.md

zvonkok:


zvonkok:
@jpecholt FYI

jpecholt:
Thanks!

docs/how-to/how-to-run-kata-containers-with-SNP-VMs.md
```diff
@@ -50,7 +50,7 @@ $ qemu_commit="$(get_from_kata_deps "assets.hypervisor.qemu.snp.commit")"
 $ git clone -b "${qemu_branch}" "${qemu_url}"
 $ pushd qemu
 $ git checkout "${qemu_commit}"
-$ ./configure --target-list=x86_64-softmmu --enable-debug
+$ ./configure --enable-virtfs --target-list=x86_64-softmmu --enable-debug
```
fidencio, line 53: @zvonkok, by the moment you add this it makes me think you're using VirtioFS instead of 9P, and if that's really the case https://github.com/kata-containers/kata-containers/blob/a9df38e4717c19cc436c5c5044aacb9792f77f14/docs/how-to/how-to-run-kata-containers-with-SNP-VMs.md?plain=1#L77-L79 should also be updated.

It'd be good to understand better the issues @jpecholt had with VirtioFS and how you solved those. :-)

fidencio: https://github.com/qemu/qemu/blob/f45cc81911adc7726e8a2801986b6998b91b816e/meson_options.txt#L218-L219

Indeed.  I wonder if `auto` default made it not needed for @jpecholt.  In any case, better to have it explicitly set.

fidencio (approved):
lgtm, thanks @zvonkok!

cmaf (commented):
Thanks @zvonkok, I approve of the change but it would be helpful if you could add a commit message describing it.

docs/how-to/how-to-run-kata-containers-with-SNP-VMs.md
jpecholt: Thanks for looking that up. Definitely makes sense to set it then.

zvonkok: @jpecholt @fidencio If the needed libraries are installed on the host, QEMU will pick it up and enable it. If not installed and you do not enable the flag, QEMU will just ignore it, and you end up without `9p` support. 
Enabling it explicitly will fail if the needed libs are not installed so this way we can be sure that it gets build.

zvonkok:
@cmaf I added a brief explanation to the commit message, PTAL

docs/how-to/how-to-run-kata-containers-with-SNP-VMs.md
jpecholt: I feared something like this might be the case so thanks for fixing it!

GabyCT (approved):

PR #5422, opened 2022-10-13
image_rpc: always call `verify_cid` after determining the cid

### LINKED ISSUE #5421: image_rs: unverified container id
# Description of problem

When a container id isn't specified by the rpc caller, `pull_image` derives one from the image path. However, it fails to verify the result, allowing the untrusted caller (e.g., the host) undue control over paths handled by the agent.

For example, calling `pull_image` with `container_id=""` and `image=".."` and skopeo present is problematic.

# Expected result

Agent would reject the request.

# Actual result

Agent accepts the request, and tries to pull an image into `/tmp/../image_oci/`. Note that this out of `/tmp`.

Once that fails, the following call is made: https://github.com/kata-containers/kata-containers/blob/7dd8b78d4434c3c9039b594cd131f6bcffd17ed0/src/agent/src/image_rpc.rs#L116 

Where `tmp_cid_path` is `/tmp/..` which is obviously just `/`.

# Further information

In addition to fixing the immediate problem, I think we should start to think about systematic ways to avoid such issues as we do a lot of path manipulation in this code.

wedsonaf:
Prior to this patch, we were missing a call to `verify_cid` when the cid was derived from the image path, which meant that the host could specify something like "prefix/..", and we would use ".." as the cid. Paths derived from this (e.g., `bundle_path`) would not be at the intended tree.

This patch factors the code out of `pull_image` so that it can be more easily tested. Tests are added for a number of cases.

Fixes #5421

### DIFF
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -205,6 +205,24 @@ impl ImageService {
             .unwrap();
     }
 
+    /// Determines the container id (cid) to use for a given request.
+    ///
+    /// If the request specifies a non-empty id, use it; otherwise derive it from the image path.
+    /// In either case, verify that the chosen id is valid.
+    fn cid_from_request(req: &image::PullImageRequest) -> Result<String> {
+        let req_cid = req.get_container_id();
+        let cid = if !req_cid.is_empty() {
+            req_cid.to_string()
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @wedsonaf1

Issue #5424, opened 2022-10-14
kata-ctl does not work on s390


liubin:
> 17:00:57 Caused by:
17:00:57   process didn't exit successfully: `/home/jenkins/workspace/kata-containers-2.0-ubuntu-s390x-unit-PR/go/src/github.com/kata-containers/kata-containers/src/tools/kata-ctl/target/release/build/ring-bd04ecb679c26bb9/build-script-build` (exit status: 101)
17:00:57   --- stderr
17:00:57   thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jenkins/.cargo/registry/src/github.com-eae4ba8cbf2ce1c7/ring-0.16.20/build.rs:358:10
17:00:57   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
17:00:57 warning: build failed, waiting for other jobs to finish...

PR #5428, opened 2022-10-14
kata-ctl: remove tls support

liubin:
TLS needs native packages/crates to support, but s390x don't support ring crate, so disable TLS to let kata-ctl run on s390x.

Fixes: #5424

### DIFF
--- a/src/tools/kata-ctl/Cargo.toml
+++ b/src/tools/kata-ctl/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0.31"
 clap = { version = "3.2.20", features = ["derive", "cargo"] }
-reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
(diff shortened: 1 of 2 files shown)


liubin:
> thread 'check::tests::check_latest_version' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.github.com")), port: None, path: "/repos/kata-containers/kata-containers/releases/latest", query: None, fragment: None }, source: hyper::Error(Connect, "invalid URL, scheme is not http") }', src/check.rs:196:64

Check needs access HTTPS, TLS can't be removed...

jodh-intel (approved):
Thanks @liubin!

lgtm

I've raised #5435 so we can look into this in more detail...

liubin:
@jodh-intel @cmaf I should close this one as [this](https://github.com/kata-containers/kata-containers/pull/5428#issuecomment-1279051224) has said, we should consider another approach.

Issue #5429, opened 2022-10-14
kata-ctl: add src/ops/version.rs to .gitignore


liubin:
`src/ops/version.rs` needs to be added to `.gitignore`.

PR #5430, opened 2022-10-14
kata-ctl: ignore src/ops/version.rs and Cargo.lock

liubin:
src/ops/version.rs and Cargo.lock are generated when building, they should be ignored.

Fixes: #5429

### DIFF
--- a/src/tools/kata-ctl/.gitignore
+++ b/src/tools/kata-ctl/.gitignore
@@ -1,1 +1,3 @@
+Cargo.lock
+src/ops/version.rs
 /vendor/


fidencio (approved):
lgtm, thanks @liubin!

Ji-Xinyou (approved):
LGTM

amshinde:
@liubin Just seeing this PR. I have raised another PR where `src/ops/version.rs` is ignored and Carlo.lock has been added to version control. We are practicing the same for other components. Since kata-ctl is a binary, it makes sense to add Cargo.lock to version control.

Here is what the rust recommendation states(https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html):
```
If you’re building a non-end product, such as a rust library that other rust [packages](https://doc.rust-lang.org/cargo/appendix/glossary.html#package) will depend on, put Cargo.lock in your .gitignore. If you’re building an end product, which are executable like command-line tool or an application, or a system library with crate-type of staticlib or cdylib, check Cargo.lock into git.
```

liubin:
@amshinde I will close this one when your PR is ready.

Issue #5425, opened 2022-10-14
packaging: virtiofsd: Build virtiofsd inside a 2nd layer container


fidencio:
When moving to [building the CI artefacts using the kata-deploy scripts](https://github.com/kata-containers/tests/pull/5168),
we've noticed that the build would fail on any machine where the tarball
wasn't officially provided.

This happens as rust is missing from the 1st layer container.  However,
it's a very common practice to leave the 1st layer container with the
minimum possible dependencies and install whatever is needed for
building a specific component in a 2nd layer container, which virtiofsd
never had

PR #5426, opened 2022-10-14
virtiofsd: Build inside a container

fidencio:
When moving to building the CI artefacts using the kata-deploy scripts, we've noticed that the build would fail on any machine where the tarball wasn't officially provided.

This happens as rust is missing from the 1st layer container.  However, it's a very common practice to leave the 1st layer container with the minimum possible dependencies and install whatever is needed for building a specific component in a 2nd layer container, which virtiofsd never had.

In this commit we introduce the second layer containers (yes, comtainers), one for building virtiofsd using musl, and one for building virtiofsd using glibc.  The reason for taking this approach was to actually simplify the scripts and avoid building the dependencies (libseccomp, libcap-ng) using musl libc.

Fixes: #5425

Signed-off-by: Fabiano Fidêncio <fabiano.fidencio@intel.com>

PR #5437, opened 2022-10-15
CC | backport | virtiofsd: Build inside a container

fidencio:
When moving to building the CI artefacts using the kata-deploy scripts, we've noticed that the build would fail on any machine where the tarball wasn't officially provided.

This happens as rust is missing from the 1st layer container. However, it's a very common practice to leave the 1st layer container with the minimum possible dependencies and install whatever is needed for building a specific component in a 2nd layer container, which virtiofsd never had.

In this commit we introduce the second layer containers (yes, comtainers), one for building virtiofsd using musl, and one for building virtiofsd using glibc. The reason for taking this approach was to actually simplify the scripts and avoid building the dependencies (libseccomp, libcap-ng) using musl libc.

Fixes: https://github.com/kata-containers/kata-containers/issues/5425

Signed-off-by: Fabiano Fidêncio [fabiano.fidencio@intel.com](mailto:fabiano.fidencio@intel.com)

---

This is a backport of https://github.com/kata-containers/kata-containers/pull/5426

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -321,35 +321,23 @@ parts:
 
   virtiofsd:
     plugin: nil
-    after: [godeps, rustdeps]
+    after: [godeps, rustdeps, docker]
     override-build: |
       source "${SNAPCRAFT_PROJECT_DIR}/snap/local/snap-common.sh"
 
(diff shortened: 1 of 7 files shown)


stevenhorsman (approved):
LGTM. Thanks!

Issue #5435, opened 2022-10-14
kata-ctl: Fix check on s390x


jodh-intel:
Create a proper fix for #5424 / #5428.

/cc @BbolroC

jodh-intel:
Related: #5438.

Issue #5402, opened 2022-10-12
kata-ctl: Simplify check function


jodh-intel:
Simplify the main `check()` function as suggested by @likebreath in https://github.com/kata-containers/kata-containers/pull/5388#discussion_r992850144.

Issue #5438, opened 2022-10-15
kata-ctl: Re-enable network check on s390x


jodh-intel:
The kata-ctl network check needs to be disabled on s390x using #5435 as that platform apparently doesn't support rust-tls (yet).

If and when rust-tls is supported, the `reqwest` HTTP crate can be used unconditionally by `kata-ctl`.

/cc @BbolroC.

BbolroC:
Thanks for looking into the issue. I will keep an eye on the support.

hbrueckner:
fyi, the TLS library used by rust-tls, `ring` lacks support for s390x. There is s390x basic support PR pending since last year: https://github.com/briansmith/ring/pull/1297

Not sure if there are other TLS alternatives for doing the tests or if we need to work and wait on above dependency.

hbrueckner:
Hi @jodh-intel 

I briefly looked at https://github.com/seanmonstar/reqwest and it seems that `reqwest` can use the native TLS crate (openssl) and being `rustls-tls` an optional feature. For s390x, I wonder whether simplying removing the `rustls-tls` feature would be the easiest option to re-enable the test case. Wdyt?

jodh-intel:
@hbrueckner, @BbolroC - I had problems using reqwest with the native libs on x86_64 due to our rust toolchain using musl as the libc, hence switched to rust-tls.

I'm trying to land the admittedly rather ugly but I think working fix (#5439) to unbreak the Kata CI.

However, it would be great if one of you who has access to s380x h/w could raise a follow-up PR that fixed the problem "more properly". Maybe we could have something like this in the manifest:

```toml
[target.'cfg(not(target_arch = "s390x"))'.dependencies]
reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }

[target.'cfg(target_arch = "s390x")'.dependencies]
reqwest = { version = "0.11", default-features = false, features = ["json", "blocking"] }
```

But I think you'd still need to find a solution to the musl problem. An alternative might be that we build `kata-ctl` using the standard libc though as I mentioned on https://github.com/jodh-intel/kata-containers/pull/10#pullrequestreview-1133119110.

PR #5439, opened 2022-10-15
kata-ctl: Disable network check on s390x

jodh-intel:
s390x apparently does not support rust-tls, which is required by the
network check (due to the `reqwest` crate dependency).

Disable the network check on s390x until we can find a solution to the
problem.

> **Note:**
>
> This fix is assumed to be a temporary one until we find a solution.
> Hence, I have not moved the network check code (which should be entirely
> generic) into an architecture specific module.

Fixes: #5435.

-----
Also fixes:

kata-ctl: arch: Improve check call

Rework the architecture-specific `check()` call by moving all the
conditional logic out of the function.

Fixes: #5402.

### DIFF
--- a/src/tools/kata-ctl/Cargo.toml
+++ b/src/tools/kata-ctl/Cargo.toml
@@ -12,9 +12,12 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0.31"
 clap = { version = "3.2.20", features = ["derive", "cargo"] }
-reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 serde_json = "1.0.85"
 thiserror = "1.0.35"
+
+# See: https://github.com/kata-containers/kata-containers/issues/5438
+[target.'cfg(not(target_arch = "s390x"))'.dependencies]
+reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 
 [dev-dependencies]
 semver = "1.0.12"
--- a/src/tools/kata-ctl/src/arch/mod.rs
+++ b/src/tools/kata-ctl/src/arch/mod.rs
@@ -7,36 +7,32 @@ use anyhow::Result;
 
 #[cfg(target_arch = "aarch64")]
 pub mod aarch64;
+#[cfg(target_arch = "aarch64")]
+pub use aarch64 as arch_specific;
 
 #[cfg(target_arch = "powerpc64le")]
 pub mod powerpc64le;
+#[cfg(target_arch = "powerpc64le")]
+pub use powerpc64le as arch_specific;
 
 #[cfg(target_arch = "s390x")]
 pub mod s390x;
+#[cfg(target_arch = "s390x")]
+pub use s390x as arch_specific;
 
 #[cfg(target_arch = "x86_64")]
 pub mod x86_64;
+#[cfg(target_arch = "x86_64")]
+pub use x86_64 as arch_specific;
+
+#[cfg(not(any(
+    target_arch = "aarch64",
+    target_arch = "powerpc64le",
+    target_arch = "s390x",
+    target_arch = "x86_64"
+)))]
+compile_error!("unknown architecture");
 
 pub fn check() -> Result<()> {
-    #[cfg(target_arch = "aarch64")]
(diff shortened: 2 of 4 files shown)


liubin:
Some clippy warnings for [s390x](http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-unit-PR/487/console):

```
22:27:30 error[E0432]: unresolved import `serde_json`
22:27:30   --> src/check.rs:16:5
22:27:30    |
22:27:30 16 | use serde_json::Value;
22:27:30    |     ^^^^^^^^^^ use of undeclared crate or module `serde_json`
22:27:30 
22:27:30 error[E0432]: unresolved import `thiserror`
22:27:30  --> src/args.rs:8:5
22:27:30   |
22:27:30 8 | use thiserror::Error;
22:27:30   |     ^^^^^^^^^ use of undeclared crate or module `thiserror`
22:27:30 
22:27:30 error: cannot determine resolution for the derive macro `Error`
22:27:30   --> src/args.rs:44:23
22:27:30    |
22:27:30 44 | #[derive(Debug, Args, Error)]
22:27:30    |                       ^^^^^
22:27:30    |
22:27:30    = note: import resolution is stuck, try simplifying macro imports
22:27:30 
22:27:30 error: cannot find attribute `error` in this scope
22:27:30   --> src/args.rs:45:3
22:27:30    |
22:27:30 45 | #[error("Argument is not valid")]
22:27:30    |   ^^^^^
22:27:30 
22:27:30 error: unused import: `std::collections::HashMap`
22:27:30   --> src/check.rs:17:5
22:27:30    |
22:27:30 17 | use std::collections::HashMap;
22:27:30    |     ^^^^^^^^^^^^^^^^^^^^^^^^^
22:27:30    |
22:27:30    = note: `-D unused-imports` implied by `-D warnings`
22:27:30 
```

hbrueckner:
Hi @jodh-intel , @BbolroC ,

> s390x apparently does not support rust-tls, which is required by the network check (due to the `reqwest` crate dependency).
> 
> Disable the network check on s390x until we can find a solution to the problem.

I would say, it is not `rust-tls` but rather its dependency `ring` where there is s390x basic support pending since last year: https://github.com/kata-containers/kata-containers/pull/5439#issue-1410132991

Not sure if there are some alternatives for doing the tests or if we need to work and wait on above dependency.

dborquez (approved):
lgtm, thanks @jodh-intel nice fix

cmaf (approved):

dborquez:
looks like more use sections need to be skipped for s390x, so WIP

liubin:
@jodh-intel @hbrueckner maybe [`native-tls`](https://docs.rs/reqwest/latest/reqwest/#optional-features) is valued a try.

jodh-intel:
@liubin, @hbrueckner - possibly, but see #5438. I'm going to merge this PR to unblock the CI but I agree that someone with s390x experience could probably fix this in a better way (please do! ;)

BbolroC:
Hi, @jodh-intel. I still do not get the point `to unblock the CI` even if some of the CI jobs are failing (it has been merged, though). The errors `never used` will be fixed on another PR or you want me or @hbrueckner to fix the errors?

jodh-intel:
@BbolroC - I merged this PR because all of the required checks have passed. But yes, there are clearly still problems on s390x so if you could look into those, that would be great. I see that @dborquez has raised #5445, but even that is still failing on s390x.

hbrueckner:
Hi @jodh-intel ,

> @BbolroC - I merged this PR because all of the required checks have passed. But yes, there are clearly still problems on s390x so if you could look into those, that would be great. I see that @dborquez has raised #5445, but even that is still failing on s390x.

I looked a bit into this issue and also to re-enable the CI tests on s390x. Two things that I need to sort out: there are some missing `[cfg(test)]` annotations leading to unused function etc. issues. Also cpu related tests do not (yet) exist on s390x. The most strange thing is that you proposed dependency solution in `Cargo.toml` does not work locally for me.

However, the good thing is that s390 use libc (not musl) so the `native-tls` approach is fine.

Let me know if I should open a PR for what I have?

hbrueckner:
@jodh-intel solved above issues, here we go #5447 . Please have a look. Thanks.

PR #5447, opened 2022-10-18
kata-ctl: Re-enable network tests on s390x (fixes 5438)

hbrueckner:
Re-enable network tests on s390x. On s390x, `reqwest` uses `native-tls` instead of `rustls-tls` and its `ring` dependencies that is not available on s390x. Further, `workspace.resolver` was necessary to allow `cfg` checks in `Cargo.toml`.

To let the tests running on s390x, further annotations where required. With that PR, test results look like:
````
git/kata-containers/kata-containers/src/tools/kata-ctl# make test
../../../utils.mk:153: "WARNING: s390x-unknown-linux-musl target is unavailable"
    Finished release [optimized] target(s) in 0.05s
     Running unittests (target/s390x-unknown-linux-gnu/release/deps/kata_ctl-7ef55a9a87c85ded)

running 7 tests
test check::tests::check_version_by_empty_url ... ok
test check::tests::check_version_by_garbage_url ... ok
test check::tests::check_version_by_invalid_url ... ok
test check::tests::test_get_cpu_flags_empty_input ... ok
test check::tests::test_get_cpu_info_empty_input ... ok
test ops::version::tests::test_get ... ok
test check::tests::check_latest_version ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.78s
````

Fixes: #5438 

Signed-off-by: Hendrik Brueckner <brueckner@linux.ibm.com>

Cc: @BbolroC @jodh-intel

### DIFF
--- a/src/tools/kata-ctl/Cargo.toml
+++ b/src/tools/kata-ctl/Cargo.toml
@@ -12,12 +12,9 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0.31"
 clap = { version = "3.2.20", features = ["derive", "cargo"] }
+reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 serde_json = "1.0.85"
 thiserror = "1.0.35"
-
-# See: https://github.com/kata-containers/kata-containers/issues/5438
-[target.'cfg(not(target_arch = "s390x"))'.dependencies]
-reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 
 [dev-dependencies]
 semver = "1.0.12"
--- a/src/tools/kata-ctl/Cargo.toml
+++ b/src/tools/kata-ctl/Cargo.toml
@@ -3,6 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+[workspace]
+resolver = "2"
+
 [package]
 name = "kata-ctl"
 version = "0.0.1"
@@ -12,9 +15,14 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0.31"
 clap = { version = "3.2.20", features = ["derive", "cargo"] }
-reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 serde_json = "1.0.85"
 thiserror = "1.0.35"
+
+[target.'cfg(target_arch = "s390x")'.dependencies]
+reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "native-tls"] }
+
+[target.'cfg(not(target_arch = "s390x"))'.dependencies]
+reqwest = { version = "0.11", default-features = false, features = ["json", "blocking", "rustls-tls"] }
 
 [dev-dependencies]
 semver = "1.0.12"
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -13,7 +13,7 @@ use std::collections::HashMap;
 const KATA_GITHUB_URL: &str =
     "https://api.github.com/repos/kata-containers/kata-containers/releases/latest";
 
-#[cfg(any(target_arch = "x86_64"))]
+#[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
 fn get_cpu_info(cpu_info_file: &str) -> Result<String> {
     let contents = std::fs::read_to_string(cpu_info_file)?;
     Ok(contents)
@@ -21,7 +21,7 @@ fn get_cpu_info(cpu_info_file: &str) -> Result<String> {
 
 // get_single_cpu_info returns the contents of the first cpu from
 // the specified cpuinfo file by parsing based on a specified delimiter
-#[cfg(any(target_arch = "x86_64"))]
+#[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
 pub fn get_single_cpu_info(cpu_info_file: &str, substring: &str) -> Result<String> {
     let contents = get_cpu_info(cpu_info_file)?;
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by BbolroC** [comment]
@liubin It would be grateful if you could have any chance to review this. Thanks.

**Comment by hbrueckner** [comment]
Hi @jodh-intel ,

based on some CI failures, I had to also exclude aarch64 for some tests (mostly cpu related where I think the arch-specifics are not yet there). Also we spot some [missing packages](http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-unit-PR/495/consoleText) being installed on the s390x CI workers (unfortunately, no env setup for [running unit tests](https://github.com/kata-containers/tests/blob/main/.ci/jenkins_job_build.sh#L217). 

I have asked @BbolroC to update the CI images and trigger a retest if ready.

**Review by dborquez — APPROVED** [review]
lgtm, thank you @hbrueckner

**Comment by hbrueckner** [comment]
There is yet another [static checks CI failure](https://github.com/kata-containers/kata-containers/actions/runs/3273990496/jobs/5389378543) likely due to the wrong `[cfg(test)]` annotations. I will provide an update along with a basic s390x cpu check implementation to make it easier. Stay tuned.

File: src/tools/kata-ctl/src/check.rs [hunk 1]
```diff
@@ -6,32 +6,22 @@
...
-    target_arch = "powerpc64le",
-    target_arch = "x86_64"
-))]
 use std::collections::HashMap;
-use std::fs;
 
 const KATA_GITHUB_URL: &str =
     "https://api.github.com/repos/kata-containers/kata-containers/releases/latest";
 
+#[cfg(test)]
```
dborquez, line 16: @hbrueckner this is not a module test, so I think you should remove the #cfg(test) annotation.

src/tools/kata-ctl/src/check.rs
```diff
@@ -6,32 +6,22 @@
...
 fn get_cpu_info(cpu_info_file: &str) -> Result<String> {
-    let contents = fs::read_to_string(cpu_info_file)?;
+    let contents = std::fs::read_to_string(cpu_info_file)?;
     Ok(contents)
 }
 
 // get_single_cpu_info returns the contents of the first cpu from
 // the specified cpuinfo file by parsing based on a specified delimiter
+#[cfg(test)]
```
dborquez, line 24: @hbrueckner the same situation as before, this function is not a module test

cmaf (commented):
Hi @hbrueckner, unless I'm misunderstanding something, it seems like the `#[cfg(test)]` annotations should not be with those non-test functions. Is there a reason it was added there that I am missing?

@dborquez tested this to confirm and may have something to add

src/tools/kata-ctl/src/check.rs
```diff
@@ -51,6 +41,7 @@ pub fn get_single_cpu_info(cpu_info_file: &str, substring: &str) -> Result<Strin
 
 // get_cpu_flags returns a string of cpu flags from cpuinfo, passed in
 // as a string
+#[cfg(test)]
```
dborquez, line 44: @hbrueckner the same case, this function is not a module test.

dborquez (changes requested):

src/tools/kata-ctl/Cargo.toml
```diff
@@ -15,7 +18,9 @@ clap = { version = "3.2.20", features = ["derive", "cargo"] }
 serde_json = "1.0.85"
 thiserror = "1.0.35"
 
-# See: https://github.com/kata-containers/kata-containers/issues/5438
+[target.'cfg(target_arch = "s390x")'.dependencies]
```
liubin, line 21: Is it possible to let both s390x and others use `native-tls`?

src/tools/kata-ctl/src/check.rs
hbrueckner: You are right, the static checks CI job failed due to this... I will push an update later today.

src/tools/kata-ctl/src/check.rs
hbrueckner: correct, will change that.

src/tools/kata-ctl/src/check.rs
hbrueckner: dito.

src/tools/kata-ctl/Cargo.toml
hbrueckner: @jodh-intel ran into issues with openssl and musl. s390x uses libc instead. I guess, ppc would be same as s390x; not sure about aarch64.

jodh-intel: @liubin - as @hbrueckner is finding, musl is the problem.

We could just link to glibc to allow native TLS since `kata-ctl` isn't running inside the container. But if we do that, we should link consistently to the same libc for all host-side elements imho (including runtime-rs/dragonball, trace-forwarder, agent-ctl, _etc_). But that's a bigger topic we should probably discuss at the AC meeting.

hbrueckner:
Hi @cmaf , @dborquez 

> Hi @hbrueckner, unless I'm misunderstanding something, it seems like the `#[cfg(test)]` annotations should not be with those non-test functions. Is there a reason it was added there that I am missing?

Initially, they seemed to be related to the test function when I had the first look into the issue from an s390x perpective. Based on CI failure and debugging, you are right, these annotations are wrong. I have corrected them in an updated patches few minutes ago.

> 
> @dborquez tested this to confirm and may have something to add

Thank you!

src/tools/kata-ctl/Cargo.toml
hbrueckner: :+1: to have this as a follow-up discussion.

dborquez (approved):

cmaf (commented):
Thanks @hbrueckner.  Could you update your commits so that they all have messages? The test is failing because one is missing.

hbrueckner:
Hi @dborquez , @cmaf , @jodh-intel 

> Thanks @hbrueckner. Could you update your commits so that they all have messages? The test is failing because one is missing.

I have spent some more cycles to re-work the commits to remove the "temporary" `[cfg(test)]` usage. Also from the still present CI failures for arm, I have excluded initially all architectures from the cpu checks. The last commit adds the basic cpu check for s390 and re-enables testing. Could be a blueprint for other architectures when providing their arch-specific check.

PR #5450, opened 2022-10-19
feat(Tracing): tracing in Rust runtime

Ji-Xinyou:
1. Instrumenting currently only on major booting process, i.e. before sandbox creation to container creation finished
* all instrumenting spans are child spans of a long running rootspan
* tracer.rs supports spans without guard entering and exiting at any place, this supports the long running rootspan

2. Add some debug trait implementation (maybe not complete) since #[instrument] needs this, and skip instrumenting the parameters is not a good choice.

3. there is one thing `TODO`, some crates, `e.g. hyper`, uses tracing and send traces to jaeger, we may offer an option to filter the traces from other crates

### DIFF
--- a/src/runtime-rs/crates/hypervisor/Cargo.toml
+++ b/src/runtime-rs/crates/hypervisor/Cargo.toml
@@ -28,6 +28,7 @@ vmm-sys-util = "0.11.0"
 rand = "0.8.4"
 path-clean = "1.0.1"
 lazy_static = "1.4"
+tracing = "0.1.36"
 
 kata-sys-util = { path = "../../../libs/kata-sys-util" }
 kata-types = { path = "../../../libs/kata-types" }
--- a/src/runtime-rs/crates/runtimes/Cargo.toml
+++ b/src/runtime-rs/crates/runtimes/Cargo.toml
@@ -12,6 +12,11 @@ netns-rs = "0.1.0"
 slog = "2.5.2"
 slog-scope = "4.4.0"
 tokio = { version = "1.28.1", features = ["rt-multi-thread"] }
+tracing = "0.1.36"
+tracing-opentelemetry = "0.18.0"
+opentelemetry = { version = "0.18.0", features = ["rt-tokio-current-thread", "trace", "rt-tokio"] }
+opentelemetry-jaeger = { version = "0.17.0", features = ["rt-tokio", "hyper_collector_client", "collector_client"] }
+tracing-subscriber = { version = "0.3", features = ["registry", "std"] }
 hyper = { version = "0.14.20", features = ["stream", "server", "http1"] }
 hyperlocal = "0.8"
 serde_json = "1.0.88"
--- a/src/runtime-rs/crates/runtimes/src/manager.rs
+++ b/src/runtime-rs/crates/runtimes/src/manager.rs
@@ -24,7 +24,8 @@ use persist::sandbox_persist::Persist;
 use resource::{cpu_mem::initial_size::InitialSizeManager, network::generate_netns_name};
 use shim_interface::shim_mgmt::ERR_NO_SHIM_SERVER;
 use tokio::fs;
-use tokio::sync::{mpsc::Sender, RwLock};
+use tokio::sync::{mpsc::Sender, Mutex, RwLock};
+use tracing::instrument;
 #[cfg(feature = "virt")]
 use virt_container::{
     sandbox::{SandboxRestoreArgs, VirtSandbox},
@@ -34,23 +35,39 @@ use virt_container::{
 #[cfg(feature = "wasm")]
 use wasm_container::WasmContainer;
 
-use crate::shim_mgmt::server::MgmtServer;
+use crate::{
+    shim_mgmt::server::MgmtServer,
+    tracer::{KataTracer, ROOTSPAN},
+};
 
 struct RuntimeHandlerManagerInner {
     id: String,
     msg_sender: Sender<Message>,
+    kata_tracer: Arc<Mutex<KataTracer>>,
     runtime_instance: Option<Arc<RuntimeInstance>>,
+}
+
+impl std::fmt::Debug for RuntimeHandlerManagerInner {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RuntimeHandlerManagerInner")
+            .field("id", &self.id)
+            .field("msg_sender", &self.msg_sender)
+            .finish()
+    }
 }
 
 impl RuntimeHandlerManagerInner {
     fn new(id: &str, msg_sender: Sender<Message>) -> Result<Self> {
+        let tracer = KataTracer::new();
         Ok(Self {
             id: id.to_string(),
             msg_sender,
+            kata_tracer: Arc::new(Mutex::new(tracer)),
             runtime_instance: None,
         })
     }
 
+    #[instrument]
     async fn init_runtime_handler(
         &mut self,
         spec: &oci::Spec,
@@ -72,9 +89,22 @@ impl RuntimeHandlerManagerInner {
             _ => return Err(anyhow!("Unsupported runtime: {}", &config.runtime.name)),
         };
         let runtime_instance = runtime_handler
-            .new_instance(&self.id, self.msg_sender.clone(), config)
+            .new_instance(&self.id, self.msg_sender.clone(), config.clone())
             .await
             .context("new runtime instance")?;
+
+        // initilize the trace subscriber
+        if config.runtime.enable_tracing {
+            let mut tracer = self.kata_tracer.lock().await;
+            if let Err(e) = tracer.trace_setup(
+                &self.id,
+                &config.runtime.jaeger_endpoint,
+                &config.runtime.jaeger_user,
+                &config.runtime.jaeger_password,
+            ) {
+                warn!(sl!(), "failed to setup tracing, {:?}", e);
+            }
+        }
 
         // start sandbox
         runtime_instance
@@ -86,6 +116,7 @@ impl RuntimeHandlerManagerInner {
         Ok(())
     }
 
+    #[instrument]
     async fn try_init(
         &mut self,
         spec: &oci::Spec,
@@ -149,6 +180,7 @@ impl RuntimeHandlerManagerInner {
             netns,
             network_created,
         };
+
         self.init_runtime_handler(spec, state, network_env, dns, Arc::new(config))
             .await
             .context("init runtime handler")?;
@@ -171,10 +203,21 @@ impl RuntimeHandlerManagerInner {
     fn get_runtime_instance(&self) -> Option<Arc<RuntimeInstance>> {
         self.runtime_instance.clone()
     }
+
+    fn get_kata_tracer(&self) -> Arc<Mutex<KataTracer>> {
+        self.kata_tracer.clone()
+    }
 }
 
 pub struct RuntimeHandlerManager {
     inner: Arc<RwLock<RuntimeHandlerManagerInner>>,
+}
+
+// todo: a more detailed impl for fmt::Debug
+impl std::fmt::Debug for RuntimeHandlerManager {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RuntimeHandlerManager").finish()
+    }
 }
 
 impl RuntimeHandlerManager {
@@ -243,6 +286,12 @@ impl RuntimeHandlerManager {
             .ok_or_else(|| anyhow!("runtime not ready"))
     }
 
+    async fn get_kata_tracer(&self) -> Result<Arc<Mutex<KataTracer>>> {
+        let inner = self.inner.read().await;
+        Ok(inner.get_kata_tracer())
+    }
+
+    #[instrument]
     async fn try_init_runtime_instance(
         &self,
         spec: &oci::Spec,
@@ -253,6 +302,7 @@ impl RuntimeHandlerManager {
         inner.try_init(spec, state, options).await
     }
 
+    #[instrument(parent = &*(ROOTSPAN))]
     pub async fn handler_message(&self, req: Request) -> Result<Response> {
         if let Request::CreateContainer(container_config) = req {
             // get oci spec
@@ -291,6 +341,7 @@ impl RuntimeHandlerManager {
         }
     }
 
+    #[instrument(parent = &(*ROOTSPAN))]
     pub async fn handler_request(&self, req: Request) -> Result<Response> {
         let instance = self
             .get_runtime_instance()
@@ -320,6 +371,11 @@ impl RuntimeHandlerManager {
             Request::ShutdownContainer(req) => {
                 if cm.need_shutdown_sandbox(&req).await {
                     sandbox.shutdown().await.context("do shutdown")?;
+
+                    // stop the tracer collector
+                    let kata_tracer = self.get_kata_tracer().await.context("get kata tracer")?;
+                    let tracer = kata_tracer.lock().await;
+                    tracer.trace_end();
                 }
                 Ok(Response::ShutdownContainer)
             }
@@ -388,6 +444,7 @@ impl RuntimeHandlerManager {
 /// 3. shimv2 create task option
 /// 4. If above three are not set, then get default path from DEFAULT_RUNTIME_CONFIGURATIONS
 /// in kata-containers/src/libs/kata-types/src/config/default.rs, in array order.
+#[instrument]
 fn load_config(spec: &oci::Spec, option: &Option<Vec<u8>>) -> Result<TomlConfig> {
     const KATA_CONF_FILE: &str = "KATA_CONF_FILE";
     let annotation = Annotation::new(spec.annotations.clone());
--- a/src/runtime-rs/crates/runtimes/src/tracer.rs
+++ b/src/runtime-rs/crates/runtimes/src/tracer.rs
@@ -1,0 +1,169 @@
+// Copyright (c) 2019-2022 Alibaba Cloud
+// Copyright (c) 2019-2022 Ant Group
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use std::cmp::min;
+use std::sync::Arc;
+
+use anyhow::Result;
+use lazy_static::lazy_static;
+use opentelemetry::global;
+use opentelemetry::runtime::Tokio;
+use tracing::{span, subscriber::NoSubscriber, Span, Subscriber};
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::Registry;
+
+const DEFAULT_JAEGER_URL: &str = "http://localhost:14268/api/traces";
+
+lazy_static! {
+    /// The ROOTSPAN is a phantom span that is running by calling [`trace_enter_root()`] at the background
+    /// once the configuration is read and config.runtime.enable_tracing is enabled
+    /// The ROOTSPAN exits by calling [`trace_exit_root()`] on shutdown request sent from containerd
+    ///
+    /// NOTE:
+    ///     This allows other threads which are not directly running under some spans to be tracked easily
+    ///     within the entire sandbox's lifetime.
+    ///     To do this, you just need to add attribute #[instrment(parent=&(*ROOTSPAN))]
+    pub static ref ROOTSPAN: Span = span!(tracing::Level::TRACE, "root-span");
+}
+
+/// The tracer wrapper for kata-containers
+/// The fields and member methods should ALWAYS be PRIVATE and be exposed in a safe
+/// way to other modules
+unsafe impl Send for KataTracer {}
+unsafe impl Sync for KataTracer {}
+pub struct KataTracer {
+    subscriber: Arc<dyn Subscriber + Send + Sync>,
+    enabled: bool,
+}
+
+impl Default for KataTracer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl KataTracer {
+    /// Constructor of KataTracer, this is a dummy implementation for static initialization
+    pub fn new() -> Self {
+        Self {
+            subscriber: Arc::new(NoSubscriber::default()),
+            enabled: false,
+        }
+    }
+
+    /// Set the tracing enabled flag
+    fn enable(&mut self) {
+        self.enabled = true;
+    }
+
+    /// Return whether the tracing is enabled, enabled by [`trace_setup`]
+    fn enabled(&self) -> bool {
+        self.enabled
+    }
+
+    /// Call when the tracing is enabled (set in toml configuration file)
+    /// This setup the subscriber, which maintains the span's information, to global and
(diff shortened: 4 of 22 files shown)


Ji-Xinyou:
@jodh-intel PTAL, the tracing in runtime is a little bit different from other components considering we need to align the lifetime of trace collectors and runtime.

For the rootspan, we need to let it live at least as long as the runtime, therefore, I hacked the tokio::tracing to not using the RAII guard to handle the lifetime of a span. But IMO this may not be the best solution?

Also, some crates like `hyper` is emitting traces, is there an elegant way to filter that out?

docs/tracing.md
```diff
@@ -89,6 +89,25 @@ Containers agent:
   The [Jaeger "all-in-one" Docker image][jaeger-all-in-one] method
   is the quickest and simplest way to run the collector for testing.
 
+  - It is worth noting that the **default collector endpoint** URL 
+    is set to http://localhost:14268/api/traces
+    When custom endpoint URL is used, make sure you are setting the correct
+    **collector** port
```
jodh-intel, line 95: Nit: Missing period:

```suggestion
    **collector** port.
```

docs/tracing.md
```diff
@@ -89,6 +89,25 @@ Containers agent:
...
+
+  - Following is a example command to run all Jaeger parts in a docker
+    container. You need to visit `localhost:16686` to check the web-UI of Jaeger
+    ```shell
+    docker run -d -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 \
```
jodh-intel, line 100: Missing `$ ` prompt (https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#code-blocks):

```suggestion
    $ docker run -d -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 \
```

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+    static ref KATATRACER: Arc<Mutex<KataTracer>> = Arc::new(Mutex::new(KataTracer::new()));
+
+    /// The ROOTSPAN is a phantom span that is running by calling [`trace_enter_root()`] at the background
+    /// once the configuration is read and config.runtime.enable_tracing is enabled
+    /// The ROOTSPAN exits by calling [`trace_exit_root()`] on shutdown request sent from containerd
+    ///
+    /// NOTE:
+    ///     This allows other threads which are not directly running under some spans to be tracked easily
+    ///     within the entire sandbox's lifetime.
+    ///    To do this, you just need to add attribute #[instrment(parent=&(*ROOTSPAN))]
```
jodh-intel, line 30: Nit: indentation:

```suggestion
    ///     within the entire sandbox's lifetime.
    ///     To do this, you just need to add attribute #[instrment(parent=&(*ROOTSPAN))]
```

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+/// The fields and member methods should ALWAYS be PRIVATE and be exposed in a safe
+/// way to other modules
+unsafe impl Send for KataTracer {}
+unsafe impl Sync for KataTracer {}
+struct KataTracer {
+    subscriber: Arc<dyn Subscriber + Send + Sync>,
+    enabled: bool,
+}
+
+//todo: remove hyper's tracing
```
jodh-intel, line 45: ?

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// let the subscriber exit the span, this has to be called in pair to enter(span)
+fn exit(span: &Span) {
+    let kt = KATATRACER.lock().unwrap();
+    let id: Option<span::Id> = span.into();
+    kt.subscriber.exit(&id.unwrap());
+}
+
+/// Varifying the configuration of jaeger and setup the default value
+fn varify_jaeger_config(endpoint: &str, username: &str, passwd: &str) -> Result<String> {
```
jodh-intel, line 146: This looks like it might a typo ("verify"?):

```suggestion
/// Verifying the configuration of jaeger and setup the default value
fn verify_jaeger_config(endpoint: &str, username: &str, passwd: &str) -> Result<String> {
```

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+    if username.is_empty() && !passwd.is_empty() {
+        warn!(
+            sl!(),
+            "Jaeger password with empty username is now allowed, tracing is NOT enabled"
```
jodh-intel, line 150:
```suggestion
            "Jaeger password with empty username is not allowed, tracing is NOT enabled"
```

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+        );
+        return Err(anyhow::anyhow!(""));
```
jodh-intel, line 152: Can you add some sort of error string here.

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // set the default endpoint address, this expects a jaeger-collector running on localhost:14268
+    let endpt = if endpoint.is_empty() {
+        "http://localhost:14268/api/traces"
```
jodh-intel, line 157: Might be worth making this URL a top-level `const` to give it more visibility.

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+use lazy_static::lazy_static;
+use opentelemetry::global;
+use opentelemetry::runtime::Tokio;
+use tracing::{span, subscriber::NoSubscriber, Span, Subscriber};
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::Registry;
+
+lazy_static! {
+    /// The static KATATRACER is the variable to provide public functions to outside modules
+    static ref KATATRACER: Arc<Mutex<KataTracer>> = Arc::new(Mutex::new(KataTracer::new()));
```
jodh-intel, line 21: I'd consider adding an underscore to make this token easier to read:

```suggestion
    static ref KATA_TRACER: Arc<Mutex<KataTracer>> = Arc::new(Mutex::new(KataTracer::new()));
```

jodh-intel (commented):
Nice work @Ji-Xinyou !! Just a few minor points...

---

I can't test this until next week realistically, but maybe others can. One other consideration is the tests in:

- https://github.com/kata-containers/tests/tree/main/functional/tracing

/cc @cmaf

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Since some crates like `hyper` are also emitting traces to jaeger's collector, we may offer a way to filter out any traces from other crates.

But I have not found an elegant way of doing this, so currently, other traces will show separately w.r.t ROOTSPAN(which we really care about). Luckily, the ROOTSPAN is not affected and the spans under it are clean.

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -106,6 +122,20 @@ impl RuntimeHandlerManagerInner {
...
+
+        // note that trace_setup() will only be called when enable_tracing is true
+        if config.runtime.enable_tracing
+            && trace_setup(
+                &self.id,
+                &config.runtime.jaeger_endpoint,
+                &config.runtime.jaeger_user,
+                &config.runtime.jaeger_password,
+            )
+            .is_ok()
```
quanweiZhou, line 134: What if the `trace_setup()` returns an error, raises the error, or maybe prints to log?

src/runtime-rs/crates/runtimes/src/tracer.rs
quanweiZhou, line 21: I am wondering if it is possible to maintain the global tracer life cycle through the manager without using the global variable KATATRACER

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn enabled(&self) -> bool {
+        self.enabled
+    }
+}
+
+/// Call when the tracing is enabled (set in toml configuration file)
+/// This setup the subscriber, which maintains the span's information, to global and
+/// inside KATATRACER.
+///
+/// Note that the span will be noop(not collected) if a valid subscriber is set
```
quanweiZhou, line 70: typo?  if an invalid subscriber is set

quanweiZhou (commented):
Thanks, @Ji-Xinyou, few suggestions

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Currently I haven't think of a way to get this done, since when entering and exiting both needs some information (now only enabled), but I wrap this to a struct for extensibility.

docs/tracing.md
Ji-Xinyou: Fixed.

docs/tracing.md
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: Fixed.

src/runtime-rs/crates/runtimes/src/tracer.rs
Ji-Xinyou: Fixed.

Ji-Xinyou:
Thanks for the review @jodh-intel @quanweiZhou ! I will be looking at the tracing test soon.

jodh-intel (approved):
Thanks @Ji-Xinyou.

lgtm

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use anyhow::Result;
+use lazy_static::lazy_static;
+use opentelemetry::global;
+use opentelemetry::runtime::Tokio;
+use tracing::{span, subscriber::NoSubscriber, Span, Subscriber};
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::Registry;
+
+const DEFAULT_JAEGER_URL: &str = "http://localhost:14268/api/traces";
```
quanweiZhou, line 19: I think the tracer can define as a struct, and as the `struct RuntimeHanderManagerInner{}` member variable, instead of using a global variable.

Ji-Xinyou: The problem is that we want the tracing start as soon as the containerd's request came in. If we make that a member field, we can only start tracing after the runtimehandler is initialized.

src/runtime-rs/crates/hypervisor/Cargo.toml
```diff
@@ -23,6 +23,7 @@ slog-scope = "4.4.0"
 thiserror = "1.0"
 tokio = { version = "1.8.0", features = ["sync"] }
 vmm-sys-util = "0.10.0"
+tracing = "0.1.36"
```
Tim-0731-Hzt, line 26: Could put them in alphabetical order?

src/runtime-rs/crates/runtimes/Cargo.toml
```diff
@@ -11,6 +11,11 @@ lazy_static = "1.4.0"
 slog = "2.5.2"
 slog-scope = "4.4.0"
 tokio = { version = "1.8.0", features = ["rt-multi-thread"] }
+tracing = "0.1.36"
+tracing-opentelemetry = "0.18.0"
+opentelemetry = { version = "0.18.0", features = ["rt-tokio-current-thread", "trace", "rt-tokio"] }
+opentelemetry-jaeger = { version = "0.17.0", features = ["rt-tokio", "hyper_collector_client", "collector_client"] }
+tracing-subscriber = { version = "0.3", features = ["registry", "std"] }
```
Tim-0731-Hzt, line 18: same as above

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -106,6 +122,22 @@ impl RuntimeHandlerManagerInner {
         };
 
         let config = load_config(spec, options).context("load config")?;
+
+        // note that trace_setup() will only be called when enable_tracing is true
+        if config.runtime.enable_tracing
+        {
+            if let Ok(_) = trace_setup(
```
Tim-0731-Hzt, line 129: I think it's better to change to 
``` rust
if trace_setup(
    &self.id,
    &config.runtime.jaeger_endpoint,
    &config.runtime.jaeger_user,
    &config.runtime.jaeger_password,
)
.is_ok()
{
    trace_enter_root();
} else {
    warn!(sl!(), "Failed to setup tracing, but tracing is enabled");
}
```

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,166 @@
...
+pub fn trace_setup(
+    sid: &str,
+    jaeger_endpoint: &str,
+    jaeger_username: &str,
+    jaeger_password: &str,
+) -> Result<()> {
+    let mut kt = KATA_TRACER.lock().unwrap();
+
+    // If varify jaeger config returns an error, it means that the tracing should not be enabled
+    let endpoint = verify_jaeger_config(jaeger_endpoint, jaeger_username, jaeger_password)?;
```
Tim-0731-Hzt, line 82: add `.context()` after this

Ji-Xinyou:
I'll be working on this tracing feature in 1 week or so. Thanks for the reviews and I will fix all of them ASAP.

I will also be looking at some possible redesigns and refactorings though when I commited this I have tried many.

YushuoEdge:
Hi @Ji-Xinyou , any update for this PR?

Ji-Xinyou:
> Hi @Ji-Xinyou , any update for this PR?

I will be working on this tomorrow xD

src/runtime-rs/crates/shim/src/bin/main.rs
```diff
@@ -142,7 +144,11 @@ fn real_main() -> Result<()> {
         Action::Delete(args) => {
             let mut shim = ShimExecutor::new(args);
             let rt = get_tokio_runtime().context("get tokio runtime")?;
-            rt.block_on(shim.delete())?
+            rt.block_on(shim.delete())?;
+
+            // XXX: end tracing when containerd asks kata to stop
+            // XXX: stop here to trace deletion process
+            trace_end();
```
YushuoEdge, line 151: This function is called in a new process, seems it will not end the former global tracer.

YushuoEdge (changes requested):
Thanks @Ji-Xinyou ! Some comments below:

YushuoEdge:
@Ji-Xinyou @quanweiZhou @Tim-Zhang @xyc-grace PTAL! 
I have raised a temporary commit based on the original implementation for discussion. The main changes are:

1.  Make `KataTracer` a member field of `RuntimeHandlerManagerInner`.
The timing of calling `RuntimeManagerInner::new()` is about the same as that of `shim.run()`, which almost guarantees that we can generate trace span from the beginning. 
2. Modify the meaning of enabled in struct KataTracer.
When we have determined that the trace is in progress, we will change this variable to true. In this way, we can better guarantee that `trace_enter_root()` and `trace_exit_root()` are called in pairs.
3. Determine when Trace starts and ends
We should not call `trace_end()` when `shim.delete()`, since this starts a new process, and it should not be possible to shut down the tracer in the original process. Now `trace_setup` and `trace_enter_root` are calling in `init_runtime_handler` (these two have been merged into one function), while `trace_end` and `trace_exit_root` are calling when the sandbox shutdown request is returning.

**There is still one thing I am not sure:**
Currently, we use `#[instrument]` to generate span in this PR. In this way, seems we will always do the span generation no matter whether we have enabled tracing. This is the most accurate and convenient for the async function's span generation. But I'm not sure if this will bring a lot of overhead, and whether we should switch to generating trace span only when tracing is turned on.

Ji-Xinyou:
> **There is still one thing I am not sure:** Currently, we use `#[instrument]` to generate span in this PR. In this way, seems we will always do the span generation no matter whether we have enabled tracing. This is the most accurate and convenient for the async function's span generation. But I'm not sure if this will bring a lot of overhead, and whether we should switch to generating trace span only when tracing is turned on.

You may use this approach (just an example)
```Rust
if enable_trace {
        let subscriber = FmtSubscriber::builder()
            .with_max_level(Level::TRACE)
            .finish();

        tracing::subscriber::set_global_default(subscriber)
            .expect("setting default subscriber failed");
}
```

IMO if no subscriber is set, no trace will be generated (Essentially they are noops).

Ji-Xinyou:
The modification by @YushuoEdge lgtm.

Ji-Xinyou:
> And I realized that we still do know how much overhead is here. So why don't us try this way first. If there is overhead we can not bear, we can try to transfer to another way then. And also, maybe we should take care to add the attribute now.

I agree. As far as I know, if no subscriber is set, the trace point has no overhead.

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -87,10 +92,23 @@ impl RuntimeHandlerManagerInner {
...
             .await
             .context("new runtime instance")?;
 
+        // initilize the trace subscriber
+        if config.runtime.enable_tracing {
+            let mut tracer = self.kata_tracer.lock().await;
+            if let Err(e) = tracer.trace_setup(
+                &self.id,
+                &config.runtime.jaeger_endpoint,
+                &config.runtime.jaeger_user,
```
xyc-grace, line 105: I'm wondering if a jaeger username with a jaeger password is required? Is it required by the crate or just we made the rule?

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -63,85 +59,88 @@ impl KataTracer {
...
+        jaeger_endpoint: &str,
+        jaeger_username: &str,
+        jaeger_password: &str,
+    ) -> Result<()> {
+        // If varify jaeger config returns an error, it means that the tracing should not be enabled
+        let endpoint = verify_jaeger_config(jaeger_endpoint, jaeger_username, jaeger_password)?;
+
+        // derive a subscriber to collect span info
+        let tracer = opentelemetry_jaeger::new_collector_pipeline()
+            .with_service_name(format!("kata-sb-{}", &sid[0..min(8, sid.len())]))
```
xyc-grace, line 80: Is 8 bits sufficient to distinguish sandboxes, especially for business scenarios with many instances.

xyc-grace:
Nice work! @YushuoEdge Just a few minor questions.

src/runtime-rs/crates/runtimes/src/manager.rs
YushuoEdge: `jaeger_user` and `jaeger_password` are optional. They are set for the authentication of the exporter.  I think that, sometimes, the upper collector needs this kind of information to identify us(the exporter). Thus we should better provide the configuration here.

src/runtime-rs/crates/runtimes/src/tracer.rs
YushuoEdge: We choose the first 8 characters of sid as unique id here. I think it is enough.

Ji-Xinyou: we could surely use more, but eight characters make it already very rarely to conflict.

prob: (1/36**8) for two of them to collide.

docs/tracing.md
```diff
@@ -89,6 +89,25 @@ Containers agent:
   The [Jaeger "all-in-one" Docker image][jaeger-all-in-one] method
   is the quickest and simplest way to run the collector for testing.
 
+  - It is worth noting that the **default collector endpoint** URL 
+    is set to http://localhost:14268/api/traces
+    When custom endpoint URL is used, make sure you are setting the correct
+    **collector** port.
+
+  - Following is a example command to run all Jaeger parts in a docker
```
cmaf, line 97: We previously did not include the command and instead directed users to Jaeger's instructions since this command has changed over time and theirs is the most current. I do not mind keeping it in though.

src/runtime-rs/crates/runtimes/src/tracer.rs
```diff
@@ -0,0 +1,170 @@
...
+    enabled: bool,
+}
+
+impl Default for KataTracer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+//todo: remove hyper's tracing
```
cmaf, line 48: Could you possibly add a link to a descriptive issue here?

cmaf (approved):
Thanks @Ji-Xinyou! I agree with @jodh-intel to take a look at the tracing test and agent shutdown test.

Are there plans to instrument more functions after this PR?

docs/tracing.md
Ji-Xinyou: I agree with @cmaf , this could be a link to jaeger's instruction. cc @YushuoEdge

Ji-Xinyou:
> Thanks @Ji-Xinyou! I agree with @jodh-intel to take a look at the tracing test and agent shutdown test.
> 
> Are there plans to instrument more functions after this PR?

The test part for 3.0 and 2.0 is a little bit different since the tracing logic differs a little bit. I think the tests could be added later on.

Instrumenting more function is extremely easy with procedural macros in Rust, currently the instructmented functions are all on the critical path of kata's booting.

Ji-Xinyou:
cc @YushuoEdge , please fix the conflict before merging this.

YushuoEdge:
Thanks @cmaf @Ji-Xinyou, and sorry for the late reply.

The comments and conflicts have been fixed!

PR #5455, opened 2022-10-19
agent: validate hugepage size is supported

snir911:
before setting a limit, otherwise paths may not be found. guest supporting different hugepage size is more likely with peer-pods where podvm may use different flavor.

Forward-port-from: https://github.com/kata-containers/kata-containers/pull/5192
Fixes: #5191


littlejawa (approved):
LGTM
Thanks @snir911 !

liubin (approved):

Issue #5330, opened 2022-10-06
CC | Shall we consider tagging "a release" from the CC branch to be used with the confidential containers operator?


fidencio:
While I'm not big fond of having a tag set on a temporary branch, having such a thing would help us to automate the already scripted process to generate a payload image to be used with the confidential containers operator.

The current process is manual, as in, someone actually calling `make cc-payload`, and this could be part of a GitHub action (as already done for the `main` branch).

/cc @kata-containers/architecture-committee @stevenhorsman @wainersm @bpradipt

wainersm:
Hi @fidencio 

> While I'm not big fond of having a tag set on a temporary branch, having such a thing would help us to automate the already scripted process to generate a payload image to be used with the confidential containers operator.
> 
> The current process is manual, as in, someone actually calling `make cc-payload`, and this could be part of a GitHub action (as already done for the `main` branch).
> 

I never ran `make cc-payload` and I don't understand the process completely. What I know it that the payload get the SHA1 of the repository on its name. Could you expand on how tagging would help?

When you say automate the process, are you thinking in when we release alpha, RC, GA of the payload or as in a CI (i.e. generate the payload as new code is merged in kata)? 

> /cc @kata-containers/architecture-committee @stevenhorsman @wainersm @bpradipt

fidencio:
> I never ran `make cc-payload` and I don't understand the process completely. What I know it that the payload get the SHA1 of the repository on its name. Could you expand on how tagging would help?

Tagging would help on triggering a GitHub action that would generate the payload image.  We can use a tag, or simply have a new action to do so.  Tagging could be easier for folks to consume, but not necessarily needed.

> When you say automate the process, are you thinking in when we release alpha, RC, GA of the payload or as in a CI (i.e. generate the payload as new code is merged in kata)?

Right now nothing on `CCv0` is related to the `main` branch tags.  We'd have our own tags to trigger a GitHub action, to then generate the tarball (without the need of someone running `make cc-payload`) and uploading that to quay.io.

Does this make sense?  We're basically trying to remove the need of a human being running `make cc-payload`. :-)

PR #5459, opened 2022-10-19
CC | action: Automate CC payload release

fidencio:
Let's create a GitHub action to automate the Kata Containers payload generation for the Confidential Containers project.

This GitHub action builds the artefacts (in parallel), merges them into a single tarball, generates the payload with the resulting tarball, and uploads the payload to the Confidential Containers quay.io.

It expects the tags to be used to be in the `CC-x.y.z` format, with x, y, and z being numbers.

Fixes: #5330

### DIFF
--- a/.github/workflows/cc-payload.yaml
+++ b/.github/workflows/cc-payload.yaml
@@ -1,0 +1,87 @@
+name: Publish Kata Containers payload for Confidential Containers
+on:
+  push:
+    tags:
+      - 'CC\-[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  build-asset:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        asset:
+          - cc-cloud-hypervisor
+          - cc-kernel
+          - cc-qemu
+          - cc-rootfs-image
+          - cc-shim-v2
+          - cc-virtiofsd
+          - cc-sev-kernel
+          - cc-sev-ovmf
+          - cc-sev-rootfs-initrd
+          - cc-tdx-kernel
+          - cc-tdx-qemu
+          - cc-tdx-td-shim
+          - cc-tdx-tdvf
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build ${{ matrix.asset }}
+        run: |
+          make "${KATA_ASSET}-tarball"
+          build_dir=$(readlink -f build)
+          # store-artifact does not work with symlink
+          sudo cp -r "${build_dir}" "kata-build"
+        env:
+          KATA_ASSET: ${{ matrix.asset }}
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
LGTM. What are your thoughts about pushing a `latest` tag at the same time. Not necessarily for this PR, but in general?

fidencio:
> LGTM. What are your thoughts about pushing a `latest` tag at the same time. Not necessarily for this PR, but in general?

Not really comfortable doing that right now, as we cannot just add a `latest` there, but rather tag as `kata-containers-latest`, considering we share the namespace with enclave-cc.  I'd rather switch to passing the tag used instead.  For instance, tagginc `kata-containers-0.2.0` there (which could also cause confusion, considering the version wouldn't match the Kata Containers project version).

Let me sleep on this suggestion for a day or two. :-)

stevenhorsman:
> > LGTM. What are your thoughts about pushing a `latest` tag at the same time. Not necessarily for this PR, but in general?
> 
> Not really comfortable doing that right now, as we cannot just add a `latest` there, but rather tag as `kata-containers-latest`, considering we share the namespace with enclave-cc. I'd rather switch to passing the tag used instead. For instance, tagginc `kata-containers-0.2.0` there (which could also cause confusion, considering the version wouldn't match the Kata Containers project version).
> 
> Let me sleep on this suggestion for a day or two. :-)

Yeah, it's an interesting one. I guess the question comes down to who is consuming the payload Image. At the moment I have two 'users':
- The operator in it's cc runtime - I think this fine as we do/plan to do a separate operator release which wraps this, so having it be manually updated there seems fine for now
- The cloud-api-adaptor payload image - this is potentially the more interesting one as it is is based on our kata payload image e.g: https://github.com/confidential-containers/cloud-api-adaptor/blob/445d72562267c61583c4e923332c92c6689819de/install/runtime-payload/Dockerfile#L2. However given that #5461 provides a CI 'latest' build that we could potentially use for operator CI and we are only planning to release every six weeks then manually adding it to the mental checklist to update seems fine to me. As I've chatted with Pradipta, longer term we might want to consider adding the kata-remote config to the payload image and then remove the CAA payload entirely, but that's a question for the future once peer pods is a proper part of the coco release.
Both of these users are likely to be done by community members, so if there aren't any 'outsiders' picking it up then needing to understand the sha of the released payload image shouldn't be too difficult and is maybe something we could add to release notes?

**TL;DR: ** - I think I agree - with #5461 an the current use cases, a release latest tag is not important at the moment. 
Thanks for indulging my brain turning this one over!

fidencio:
Removed the `do-not-merge` label as I fixed @wainersm's comments from #5461 here as well.

wainersm (approved):

Issue #5460, opened 2022-10-19
CC | Publish a Kata Containers payload on every CCv0 push


fidencio:
Let's have a GitHub action to publish the Kata Containers payload, after
every push to the CCv0 branch, to the Confidential Containers
`runtime-payload-ci` registry.

The intention of this action is to allow developers to test new
features, and easily bisect breakages that could've happened during the
development process.  Ideally we'd have a CI/CD pipeline where every
single change would be tested with the operator, but we're not yet
there.  In any case, this work would still be needed. :-)

It's very important to mention that this should be carefully considered
on whether it should or should not be merged back to `main`, as the flow
of PRs there is way higher than what we currently have as part of the
CCv0 branch.

PR #5461, opened 2022-10-19
CC | actions: Publish a payload on every CCv0 push

fidencio:
Let's have a GitHub action to publish the Kata Containers payload, after
every push to the CCv0 branch, to the Confidential Containers
`runtime-payload-ci` registry.

The intention of this action is to allow developers to test new
features, and easily bisect breakages that could've happened during the
development process.  Ideally we'd have a CI/CD pipeline where every
single change would be tested with the operator, but we're not yet
there.  In any case, this work would still be needed. :-)

It's very important to mention that this should be carefully considered
on whether it should or should not be merged back to `main`, as the flow
of PRs there is way higher than what we currently have as part of the
CCv0 branch.

Fixes: #5460

### DIFF
--- a/.github/workflows/cc-payload-after-push.yaml
+++ b/.github/workflows/cc-payload-after-push.yaml
@@ -1,0 +1,87 @@
+name: CI | Publish Kata Containers payload for Confidential Containers
+on:
+  push:
+    branches:
+      - CCv0
+
+jobs:
+  build-asset:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        asset:
+          - cc-cloud-hypervisor
+          - cc-kernel
+          - cc-qemu
+          - cc-rootfs-image
+          - cc-shim-v2
+          - cc-virtiofsd
+          - cc-sev-kernel
+          - cc-sev-ovmf
+          - cc-sev-rootfs-initrd
+          - cc-tdx-kernel
+          - cc-tdx-qemu
+          - cc-tdx-td-shim
+          - cc-tdx-tdvf
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build ${{ matrix.asset }}
+        run: |
+          make "${KATA_ASSET}-tarball"
+          build_dir=$(readlink -f build)
+          # store-artifact does not work with symlink
+          sudo cp -r "${build_dir}" "kata-build"
+        env:
+          KATA_ASSET: ${{ matrix.asset }}
+          TAR_OUTPUT: ${{ matrix.asset }}.tar.gz
+
+      - name: store-artifact ${{ matrix.asset }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: kata-artifacts
+          path: kata-build/kata-static-${{ matrix.asset }}.tar.xz
+          retention-days: 1
+          if-no-files-found: error
+
+  create-kata-tarball:
+    runs-on: ubuntu-latest
+    needs: build-asset
+    steps:
+      - uses: actions/checkout@v3
+      - name: get-artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: kata-artifacts
+          path: kata-artifacts
+      - name: merge-artifacts
+        run: |
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-merge-builds.sh kata-artifacts
+      - name: store-artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: kata-static-tarball
+          path: kata-static.tar.xz
+          retention-days: 1
+          if-no-files-found: error
+
+  kata-payload:
+    needs: create-kata-tarball
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: ${{ secrets.COCO_QUAY_DEPLOYER_USERNAME }}
+          password: ${{ secrets.COCO_QUAY_DEPLOYER_PASSWORD }}
+
+      - uses: actions/checkout@v3
+      - name: get-kata-tarball
+        uses: actions/download-artifact@v3
+        with:
+          name: kata-static-tarball
+
+      - name: build-and-push-kata-payload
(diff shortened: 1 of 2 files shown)


.github/workflows/cc-payload-after-push.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - uses: actions/checkout@v3
+      - name: get-kata-tarball
+        uses: actions/download-artifact@v3
+        with:
+          name: kata-static-tarball
+      - name: build-and-push-kata-payload
+        id: build-and-push-kata-payload
+        run: |
+          docker login -u ${{ secrets.COCO_QUAY_DEPLOYER_USERNAME }} p ${{ secrets.COCO_QUAY_DEPLOYER_PASSWORD }} quay.io
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh $(pwd)/kata-static.tar.xz "quay.io/repository/confidential-containers/runtime-payload-ci" "kata-containers-latest"
```
stevenhorsman, line 82: This is probably not helpful, but I'm not sure about `kata-containers-latest` as the tag. I with there is merit to having a latest 'release' and a latest 'CI' build, so if we do adopt that I'd want to let users be clear which is which. I'm not sure if this makes sense and is fair?

stevenhorsman (approved):
LGTM to me. One question about the tag

.github/workflows/cc-payload-after-push.yaml
fidencio: That's a totally fair question.

So, a few things to keep in mind here.
1. Although we're using `quay.io/confidential-containers` as the place to store the payload images, we do use different registries / repos / whatever it's called for **official** and **CI** images.
  * **official**: `quay.io/confidential-containers/runtime-payload`
  * **CI**: `quay.io/confidential-containers/runtime-payload-ci`
2. I see this also being used by `enclave-cc`, thus just tagging `latest` could cause confusion.  Thus I've gone for `kata-containers-latest` and I hope `enclave-cc` would go for  `enclave-cc-latest` on their side.
3. I do **not** expect users to be using anything from the `quay.io/confidential-containers/runtime-payload-ci` registry, that's meannt for developers, for testing and CI purpose.

With everything said above, do you still think we'll have a confusion even with the "repos" / registries being different?

stevenhorsman: Cool - thanks for that explanation. I did spot that is was in a slightly different repo (with `-ci` appended), but then promptly forgot after looking at the rest of the PR. I think this is totally fine to start with and we can always adjust later if the use case come up rather than trying to predict what might happen. Sorry for the noise!

.github/workflows/cc-payload-after-push.yaml
```diff
@@ -0,0 +1,82 @@
...
+          - cc-sev-kernel
+          - cc-sev-ovmf
+          - cc-sev-rootfs-initrd
+          - cc-tdx-kernel
+          - cc-tdx-qemu
+          - cc-tdx-td-shim
+          - cc-tdx-tdvf
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install docker
```
wainersm, line 27: Github ubuntu-latest runner comes already with docker installed and configured.

.github/workflows/cc-payload-after-push.yaml
```diff
@@ -0,0 +1,82 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: get-kata-tarball
+        uses: actions/download-artifact@v3
+        with:
+          name: kata-static-tarball
+      - name: build-and-push-kata-payload
+        id: build-and-push-kata-payload
+        run: |
+          docker login -u ${{ secrets.COCO_QUAY_DEPLOYER_USERNAME }} p ${{ secrets.COCO_QUAY_DEPLOYER_PASSWORD }} quay.io
```
wainersm, line 81: Never used myself, but there is an action to login in registries. See https://github.com/confidential-containers/cloud-api-adaptor/blob/staging/.github/workflows/image.yaml#L42 as an example.

.github/workflows/cc-payload-after-push.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - uses: actions/checkout@v3
+      - name: Install docker
+        run: |
+          curl -fsSL https://test.docker.com -o test-docker.sh
+          sh test-docker.sh
+
+      - name: Build ${{ matrix.asset }}
+        run: |
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-copy-yq-installer.sh
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh --build="${KATA_ASSET}"
```
wainersm, line 35: why not use make targets?

.github/workflows/cc-payload-after-push.yaml
```diff
@@ -0,0 +1,82 @@
...
+          build_dir=$(readlink -f build)
+          # store-artifact does not work with symlink
+          sudo cp -r "${build_dir}" "kata-build"
+        env:
+          KATA_ASSET: ${{ matrix.asset }}
+          TAR_OUTPUT: ${{ matrix.asset }}.tar.gz
+
+      - name: store-artifact ${{ matrix.asset }}
```
wainersm, line 43: Artifacts lives in storage for 90 days by default. I think we should reduce it to 1 day or perhaps a few hours? So that we don't hit the storage limits which I don't know what is :)

See: retention policy in https://github.com/actions/upload-artifact

.github/workflows/cc-payload-after-push.yaml
wainersm: Just added a description for the repository, see https://quay.io/repository/confidential-containers/runtime-payload-ci?tab=info .

fidencio:
@wainersm, I've addressed all your comments.

wainersm (approved):
Great! Thanks @fidencio !

fidencio:
As the CIs failing are not related to this PR, at all, I'm going ahead and merging it.

Issue #5464, opened 2022-10-19
qemu: Split static-build Dockerfile in a way that we can cache the builder image


fidencio:
Differently than every single other bit that's part of our repo, QEMU
has been using a single Dockerfile that prepares an environment where
the project can be built, but *also* building the project as part of
that very same Dockerfile.

This is a problem, for several different reasons, including:
* It's very hard to have a reproducible build if you don't have an
  archived image of the builder
* One cannot cache / ipload the image of the builder, as that contains
  already a specific version of QEMU
* Every single CI run we end up building the builder image, which
  includes building dependencies (such as liburing)

Let's split the logic into a new build script, and pass the build script
to be executed inside the builder image, which will be only responsible
for providing an environment where QEMU can be built.

PR #5465, opened 2022-10-19
qemu: Re-work static-build Dockerfile

fidencio:
Differently than every single other bit that's part of our repo, QEMU has been using a single Dockerfile that prepares an environment where the project can be built, but *also* building the project as part of that very same Dockerfile.

This is a problem, for several different reasons, including:
* It's very hard to have a reproducible build if you don't have an archived image of the builder
* One cannot cache / ipload the image of the builder, as that contains already a specific version of QEMU
* Every single CI run we end up building the builder image, which includes building dependencies (such as liburing)

Let's split the logic into a new build script, and pass the build script to be executed inside the builder image, which will be only responsible for providing an environment where QEMU can be built.

Fixes: #5464

### DIFF
--- a/tools/packaging/static-build/qemu/build-base-qemu.sh
+++ b/tools/packaging/static-build/qemu/build-base-qemu.sh
@@ -9,6 +9,8 @@ set -o nounset
 set -o pipefail
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+readonly repo_root_dir="$(cd "${script_dir}/../../../.." && pwd)"
+readonly qemu_builder="${script_dir}/build-qemu.sh"
 
 source "${script_dir}/../../scripts/lib.sh"
 source "${script_dir}/../qemu.blacklist"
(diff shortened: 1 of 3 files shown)


fidencio:
By the way, I didn't find a good way to split this into more organised commits.  The way it's been done, at least, allow us to reuse the builder image whenever we have to build QEMU.

tools/packaging/static-build/qemu/build-base-qemu.sh
```diff
@@ -9,6 +9,8 @@ set -o nounset
 set -o pipefail
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+readonly repo_root_dir="$(cd "${script_dir}/../../../.." && pwd)"
+readonly qemu_builder="${script_dir}/build-qemu.sh"
```
stevenhorsman, line 13: Do we need the `readonly qemu_builder="${script_dir}/build-qemu.sh"` if we are running the whole file path in the 'docker' run to build it?

stevenhorsman (approved):
A couple of questions if things are still needed/used, but it makes sense to me overall. Thanks!

tools/packaging/static-build/qemu/build-base-qemu.sh
fidencio: Not really, good catch.

dborquez (approved):
thank you @fidencio , LGTM

Issue #5369, opened 2022-10-07
runtime-rs: implement direct volume handler for kata-ctl


Tim-0731-Hzt:
implement direct volume handler (volume stat and volume add) with unit test for kata-ctl

jodh-intel:
This is a duplicate of #5341.

Tim-0731-Hzt:
Same as #5370

Tim-0731-Hzt:
Yes! 
1. There are four operations in direct volume(`add`, `remove`, `stat`, `resize`). To handle `stat` and `resize`, you can see https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/shim_management.go#L266-L267, and this could be implemented in `runtime-rs/crates/runtimes/src/shim_mgmt`.  To handle `add`, see https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/container.go#L615.
2. `runtime/cmd/kata-runtime/kata-volume.go` and `runtime/pkg/direct-volume` is reference to #5341, which is used for sending the direct volume operations intead of handling.

tzY15368:
@Tim-0731-Hzt  Thinks for the info! It looks like runtime-rs's version of sandbox doesn't support stat queries yet, nor does the agent have such a trait in `runtime-rs/crates/agent/src/libs.rs` for runtime-rs to directly call. I'm wondering if I missed something🤔

Tim-0731-Hzt:
> @Tim-0731-Hzt Thanks for the info! It looks like runtime-rs's version of sandbox doesn't support direct-volume stat queries or resize yet, nor does the agent have such a trait in `runtime-rs/crates/agent/src/libs.rs` for runtime-rs to directly call. I'm wondering if I missed something🤔

Hi @tzY15368 . Yes, currently, runtime-rs doesn't support these functionalities, I think you need to define the trait and implement it.

tzY15368:
@Tim-0731-Hzt Hi Tim, I've implemented the runtime-rs and sandbox part of this issue, and just realized that `resize_volume` isn't implemented in the agent either, so I looked back at the old agent repo https://github.dev/kata-containers/agent trying to find what this command actually did, but didn't really find much relevant info. Could you share some insight on this?

Tim-0731-Hzt:
Hi @tzY15368. I just found this issue #3694, you are right, It's not implemented yet. I will change this issue to implement the `volume stat` only.

[...]

PR #5467, opened 2022-10-20
Feat: implementation of kata-ctl direct-volume operations

### LINKED ISSUE #5341: kata-ctl: direct-volume: Implement with tests
Implement the `kata-runtime direct-volume` command with unit tests.

tzY15368:
This change adds support for the `direct-volume` subcommand in kata-ctl, and implements some of the directly-dependent functions in the runtime-rs codebase, namely `add` and `remove`. `stat` and `resize` requires further changes to both sandbox-rs & agent, and will be implemented in #5369 . The patches for dbs-* in `Cargo.toml` are added because otherwise the code won't build.

Fixes #5341 

Signed-off-by: Tingzhou Yuan <tzyuan15@bu.edu>


src/tools/kata-ctl/src/args.rs
```diff
@@ -84,3 +84,46 @@ pub enum IpTablesArguments {
...
+#[derive(Debug, Args)]
+pub struct DirectVolStatsArgs {
+    pub volume_path: String,
+}
+
+#[derive(Debug, Args)]
+pub struct DirectVolResizeArgs {
+    pub volume_path: String,
+    pub resize_size: u64,
+}
```
Tim-0731-Hzt, line 129: add a new line after this

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+            return Ok(String::from(path.file_name().to_str().unwrap()));
+        }
+    }
+    return Err(anyhow!("no sandbox found for {}".to_string()))?;
+}
+
+#[cfg(test)]
+mod tests {
+    #[allow(unused_macros)]
+    macro_rules! skip_if_root {
```
Ji-Xinyou, line 95: there are skip_if_root and skip_if_not_root macros under `libs/test-utils`, you may use that instead of adding a new one.

src/runtime-rs/crates/runtimes/src/shim_mgmt/handlers.rs
```diff
@@ -50,3 +55,17 @@ async fn agent_url_handler(
         .unwrap_or_else(|_| String::from(""));
     Ok(Response::new(Body::from(agent_sock)))
 }
+
+async fn direct_volume_stats_handler(
+    _sandbox: Arc<dyn Sandbox>,
+    _req: Request<Body>,
+) -> Result<Response<Body>> {
+    Ok(Response::new(Body::from("is sandbox fully implemented?")))
```
Tim-0731-Hzt, line 63: `Direct volume` depend on the functionality of block device hotplug. This pr is #5441 is doing this.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,78 @@
...
+use url;
+
+const TIMEOUT: Duration = Duration::new(5, 0);
+const CONTENT_TYPE_JSON: &str = "application/json";
+
+pub fn handle_direct_volume(vol_cmd: DirectVolumeCommand) -> Result<()> {
+    let command = vol_cmd.directvol_cmd;
+    match command {
+        DirectVolSubcommand::Add(args) => {
+            runtimes::direct_volume::add(&args.volume_path, &args.mount_info)?;
```
Tim-0731-Hzt, line 17: To handle the `add` command in `kata-ctl`, you need to write the mount info of a direct volume into a filesystem path. It seems you are trying to call the method in `runtime-rs` directly. `runtime-rs` need to get the mount info from the file under the directory, then it could be able to handle it (this functionality is to solve #5369 ).

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+        DirectVolSubcommand::Remove(args) => {
+            runtimes::direct_volume::remove(&args.volume_path)?;
```
Tim-0731-Hzt, line 20: `Remove` need to delete the direct volume path including all the files inside it.

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+fn join_path(volume_path: &String) -> PathBuf {
+    let urlencoded_vol_path = base64::encode(volume_path.as_bytes());
+    let mount_info_file_path = Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(urlencoded_vol_path);
+    return mount_info_file_path;
+}
+
+// add writes the mount info (json string) of a direct volume into a filesystem path known to Kata Container.
+pub fn add(volume_path: &String, mount_info: &String) -> Result<()> {
```
Tim-0731-Hzt, line 36: It's better to put this function to `kata-ctl`

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    let unserialized_mount_info: VolumeMountInfo = serde_json::from_str(mount_info)?;
+
+    let mount_info_file_path = mount_info_dir_path.join(MOUNT_INFO_FILE_NAME);
+    let serialized_mount_info = serde_json::to_string(&unserialized_mount_info)?;
+    fs::write(mount_info_file_path, serialized_mount_info)?;
+    Ok(())
+}
+
+// remove deletes the direct volume path including all the files inside it.
+pub fn remove(volume_path: &String) -> Result<()> {
```
Tim-0731-Hzt, line 59: It's better to put this function to `kata-ctl`

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+
+const KATA_DIRECT_VOLUME_ROOT_PATH: &str = "/run/kata-containers/shared/direct-volumes";
+const MOUNT_INFO_FILE_NAME: &str = "mountInfo.json";
+
+pub const DIRECT_VOLUME_STAT_URL: &str = "/direct-volume/stats";
+pub const DIRECT_VOLUME_RESIZE_URL: &str = "/direct-volume/resize";
+pub const DIRECT_VOLUME_PATH_KEY: &str = "path";
+
+#[derive(Serialize, Deserialize)]
+pub struct VolumeMountInfo {
```
Tim-0731-Hzt, line 16: As this struct needs to be used by both `runtime-rs` and `kata-ctl`, I think it's better to put it under `libs/kata-types/src/mount.rs`, and rename it to `DirectVolumeMountInfo`.

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
Tim-0731-Hzt, line 36: It's better to put `add` and `remove` under `volume_ops.rs`.

liubin:
Hi @tzY15368 I think you need to refactor your commits:

- [this](https://github.com/kata-containers/kata-containers/pull/5467/commits/5d34826f60439ab08499ce120f56b369dbd75f91) seems only changed the `use`
- and one merge commit
- and one with the title `kata-ctl: style: enforce code format`, which should usually be included in commits introducing the format issue.

For each commit, it must:

- without compile/test/clippy errors/warnings
- function complete and workable, includes tests if the function is first implemented.

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub volume_type: String,
+    pub device: String,
+    pub fs_type: String,
+    pub metadata: HashMap<String, String>,
+    pub options: Vec<String>,
+}
+#[derive(Serialize, Deserialize)]
```
liubin, line 23: Add a blank line before type definition.

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+pub struct ResizeRequest {
+    pub size: u64,
+    pub volume_path: String,
+}
+
+fn join_path(volume_path: &String) -> PathBuf {
+    let urlencoded_vol_path = base64::encode(volume_path.as_bytes());
+    let mount_info_file_path = Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(urlencoded_vol_path);
```
liubin, line 31: In Rust this should be:

```rust
    let mount_info_file_path = Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(urlencoded_vol_path);
    mount_info_file_path
```

Or event

```rust
    Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(urlencoded_vol_path)
```

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+// add writes the mount info (json string) of a direct volume into a filesystem path known to Kata Container.
+pub fn add(volume_path: &String, mount_info: &String) -> Result<()> {
+    let mount_info_dir_path = join_path(volume_path);
+    // check if path exists and is dir, if not, create this dir
+    if !mount_info_dir_path.exists() {
+        fs::create_dir_all(&mount_info_dir_path)?;
+    } else if !mount_info_dir_path.is_dir() {
+        return Err(anyhow!(
+            "{} is not a directory",
+            mount_info_dir_path.to_str().unwrap()
```
liubin, line 44: This may be enough to display `mount_info_dir_path`' value:

```rust
anyhow!(
            "{:?} is not a directory",
            mount_info_dir_path);
```

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+        ));
+    }
+
+    // This behavior of deserializing and serializing comes from `src/runtime/pkg/direct-volume/utils.go/add`
```
liubin, line 48: If you want add a reference, please use the full GitHub link like `https://github.com/kata-containers/kata-containers/blob/cd27ad144e1a111cb606015c5c9671431535e644/src/runtime/pkg/direct-volume/utils.go#L57-L79` with:

- the full path
- commit hash
- line range

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+pub fn remove(volume_path: &String) -> Result<()> {
+    let path = join_path(volume_path);
+    // removes path and any children it contains.
+    fs::remove_dir_all(path)?;
+    Ok(())
+}
+
+pub fn get_volumn_mount_info(volume_path: &String) -> Result<VolumeMountInfo> {
+    let mount_info_file_path = join_path(volume_path).join(MOUNT_INFO_FILE_NAME);
+    println!(
```
liubin, line 68: No debug code left before this PR is ready to merge.

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    let dir_path = join_path(volume_path);
+    let paths = fs::read_dir(dir_path)?;
+    for path in paths {
+        let path = path?;
+        // compare with MOUNT_INFO_FILE_NAME
+        if path.file_name() == MOUNT_INFO_FILE_NAME {
+            return Ok(String::from(path.file_name().to_str().unwrap()));
+        }
+    }
+    return Err(anyhow!("no sandbox found for {}".to_string()))?;
```
liubin, line 89: the placeholder `{}` has no value specified.

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -48,9 +48,9 @@ pub fn handle_check(checkcmd: CheckArgument) -> Result<()> {
     Ok(())
 }
 
-pub fn handle_check_volume() -> Result<()> {
-    Ok(())
-}
+// pub fn handle_check_volume() -> Result<()> {
```
liubin, line 51: Delete unused code.

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub volume_path: String,
+}
+
+fn join_path(volume_path: &String) -> PathBuf {
+    let urlencoded_vol_path = base64::encode(volume_path.as_bytes());
+    let mount_info_file_path = Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(urlencoded_vol_path);
+    return mount_info_file_path;
+}
+
+// add writes the mount info (json string) of a direct volume into a filesystem path known to Kata Container.
```
liubin, line 35: Kata Container -> Kata Containers

src/runtime-rs/crates/runtimes/src/direct_volume/mod.rs
tzY15368: Should the constants and helper functions (basically everything) be moved to `src/runtime-rs/crates/runtimes/src/direct_volume/volume_ops.rs` 
 as well? It looks like I should only use it to export modules, like the role __init__.py plays in Python.

Tim-0731-Hzt: yes, I think so. Beucase all the functions in `mod.rs` is to implement the subcommand of `kata-ctl`, it's better to put it there.

tzY15368:
> Hi @tzY15368 I think you need to refactor your commits:
> 
> * [this](https://github.com/kata-containers/kata-containers/pull/5467/commits/5d34826f60439ab08499ce120f56b369dbd75f91) seems only changed the `use`
> * and one merge commit
> * and one with the title `kata-ctl: style: enforce code format`, which should usually be included in commits introducing the format issue.
> 
> For each commit, it must:
> 
> * without compile/test/clippy errors/warnings
> * function complete and workable, includes tests if the function is first implemented.

Hi @liubin thanks for the feedback! Looking back at my commits, I agree that the newer ones do look a bit messy. However I'm not so sure what I should do at this point, I did turn "format on save" on, so format issues shouldn't appear anymore. Perhaps I should squash all the existing minor refactors and new suggested ones from code reviews together? Please share some advice😀

Tim-0731-Hzt:
> * ith the title `ka`

you can rebase each commit, and use `cargo fmt` and `cargo clippy`

liubin:
@tzY15368 as @Tim-0731-Hzt has said, you can look around `git rebase -i`, with which you can:

- use the latest main branch as your branch
- change commit orders
- modify a separate commit without touching other commits
- squash(combine) commits into one.

And another choice may be to split a large PR into many small and separated ones.

tzY15368:
> @tzY15368 as @Tim-0731-Hzt has said, you can look around `git rebase -i`, with which you can:
> 
> * use the latest main branch as your base branch to avoid merge commit
> * change commit orders
> * modify a separate commit without touching other commits
> * squash(combine) commits into one.
> 
> And another choice may be to split a large PR into many small and separated ones.

Thanks for the instructions! I'll fix my commits😀

src/runtime-rs/crates/runtimes/src/direct_volume/volume_ops.rs
```diff
@@ -0,0 +1,154 @@
+use anyhow::{anyhow, Ok, Result};
```
Tim-0731-Hzt, line 1: I think it's better to put everything in this file into `src/tools/kata-ctl/src/ops/volume_ops.rs` cause it is implementing the subcommand of `kata-ctl`

tzY15368: Thanks for the advice! Moving things to client side would make much sense. However, I did move `ResizeRequest` to `runtimes/src/shimmgmt/handlers.rs` and make some of the constants & mods public as this would avoid cyclic dependency issues and I think it's best if these are kept in server side.

tzY15368:
Also, the failed CI static check mentioned 
```
ERROR: Required SPDX license header check ('SPDX-License-Identifier: Apache-2.0') failed for the following files:
src/tools/kata-ctl/src/ops/volume_ops.rs
...
```
Should I just add the line `// SPDX-License-Identifier: Apache-2.0` to the first line of the mentioned files? @Tim-0731-Hzt  Thanks in advance 😀

Tim-0731-Hzt:
> Also, the failed CI static check mentioned
> 
> ```
> ERROR: Required SPDX license header check ('SPDX-License-Identifier: Apache-2.0') failed for the following files:
> src/tools/kata-ctl/src/ops/volume_ops.rs
> ...
> ```
> 
> Should I just add the line `// SPDX-License-Identifier: Apache-2.0` to the first line of the mentioned files? @Tim-0731-Hzt Thanks in advance 😀

Yes. You can also add the copyright above that, it could be you uni or yourself.

src/runtime-rs/crates/runtimes/src/shim_mgmt/handlers.rs
```diff
@@ -50,3 +63,24 @@ async fn agent_url_handler(
...
+    )))
+}
+
+async fn direct_volume_resize_handler(
+    _sandbox: Arc<dyn Sandbox>,
+    _req: Request<Body>,
+) -> Result<Response<Body>> {
+    // Same as `direct_volume_stats_handler`
+    Ok(Response::new(Body::from(
+        "sandbox doesn't support direct-volume-resize yet",
```
jodh-intel, line 84: Could you add the issue URL in line 72 to the error message?

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+
+async fn resize(volume_path: &str, size: u64) -> Result<()> {
+    let sandbox_id = get_sandbox_id_for_volume(volume_path)?;
+    let mount_info = get_volumn_mount_info(volume_path)?;
+    let resize_req = runtimes::ResizeRequest {
+        size,
+        volume_path: mount_info.device,
+    };
+    let encoded = serde_json::to_string(&resize_req)?;
+    let shim_client = runtimes::MgmtClient::new(sandbox_id, Some(TIMEOUT)).unwrap();
```
jodh-intel, line 50: Since this function returns a `Result`, can you replace the `.unwrap()` with `?`. Same comment for other calls to `.unwrap()`.

See: https://github.com/kata-containers/kata-containers/blob/main/docs/code-pr-advice.md#rust-language-specifics

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+use hyper;
+use kata_types::mount::DirectVolumeMountInfo;
+use std::{fs, path::Path, path::PathBuf, time::Duration};
+use url;
+
+use runtimes;
+
+const TIMEOUT: Duration = Duration::new(5, 0);
+const CONTENT_TYPE_JSON: &str = "application/json";
+const MOUNT_INFO_FILE_NAME: &str = "mountInfo.json";
```
jodh-intel, line 19: This file should reference `mountInfoFileName` defined in `src/runtime/pkg/direct-volume/utils.go` rather than defining its own version.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+const KATA_DIRECT_VOLUME_ROOT_PATH: &str = "/run/kata-containers/shared/direct-volumes";
```
jodh-intel, line 20: LIkewise, `kataDirectVolumeRootPath` sould be used instead of this version.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    match command {
+        DirectVolSubcommand::Add(args) => {
+            add(&args.volume_path, &args.mount_info)?;
+        }
+        DirectVolSubcommand::Remove(args) => {
+            remove(&args.volume_path)?;
+        }
+        DirectVolSubcommand::Stats(args) => {
+            let result = executor::block_on(stats(&args.volume_path))?;
+            println!("{}", &result);
```
jodh-intel, line 33: I guess this is fine for now. But it may be better to make this function return `Result<Option<String>>`. Then, all the commands except this one can return `Result<None>`, but `Stats` can return `Result<Some<String>>`.

You'd need to rename this function and have `handle_direct_volume(...)`:

1. Call the newly named function.
2. check for errors using `?`
3. Unpack the `Result` to get at the `Option` value.
4. If not `None`, print the value.

Structuring the code like that will help with future work to support but logging and various different output formats. It would also mean you can unit test this function more easily.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+
+// remove deletes the direct volume path including all the files inside it.
+pub fn remove(volume_path: &str) -> Result<()> {
+    let path = join_path(volume_path);
+    // removes path and any children it contains.
+    fs::remove_dir_all(path)?;
+    Ok(())
+}
+
+pub fn get_volumn_mount_info(volume_path: &str) -> Result<DirectVolumeMountInfo> {
```
jodh-intel, line 119: Typo?

```suggestion
pub fn get_volume_mount_info(volume_path: &str) -> Result<DirectVolumeMountInfo> {
```

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    let b64_encoded_path = base64::encode(volume_path.as_bytes());
+    Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(b64_encoded_path)
+}
+
+// add writes the mount info (json string) of a direct volume into a filesystem path known to Kata Containers.
+pub fn add(volume_path: &str, mount_info: &str) -> Result<()> {
+    let mount_info_dir_path = join_path(volume_path);
+    // check if path exists and is dir, if not, create this dir
+    if !mount_info_dir_path.exists() {
+        fs::create_dir_all(&mount_info_dir_path)?;
```
jodh-intel, line 95: This is racy. It would be safer to remove the check and call `create_dir_all()` unconditionally.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+use url;
+
+use runtimes;
+
+const TIMEOUT: Duration = Duration::new(5, 0);
+const CONTENT_TYPE_JSON: &str = "application/json";
+const MOUNT_INFO_FILE_NAME: &str = "mountInfo.json";
+const KATA_DIRECT_VOLUME_ROOT_PATH: &str = "/run/kata-containers/shared/direct-volumes";
+
+pub fn handle_direct_volume(vol_cmd: DirectVolumeCommand) -> Result<()> {
```
jodh-intel, line 22: Is this a privileged command? If it needs to be run as `root`, we should call `nix::unistd::Uid::effective().is_root()` and return a helpful message to the user in the `Result` if they are not the super-user.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    let shim_client = runtimes::MgmtClient::new(sandbox_id, Some(TIMEOUT)).unwrap();
+
+    let url = runtimes::DIRECT_VOLUMN_RESIZE_URL;
+    let response = shim_client
+        .post(url, &String::from(CONTENT_TYPE_JSON), &encoded)
+        .await?;
+    if response.status() != hyper::StatusCode::OK {
+        return Err(anyhow!(
+            "status {}: failed to resize volume",
+            response.status()
```
jodh-intel, line 59: Nit: The user doesn't know how this is implemented, so I'd drop the status message and just format it as an error. Something like:

```suggestion
            "failed to resize volume: {:?}",
            response.status()
```

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    // turn body into string
+    let body = hyper::body::to_bytes(response.into_body()).await?;
+    let body = String::from_utf8(body.to_vec()).unwrap();
+    Ok(body)
+}
+
+// join_path joins user provided volumepath with kata direct-volume root path
+fn join_path(volume_path: &str) -> PathBuf {
+    let b64_encoded_path = base64::encode(volume_path.as_bytes());
+    Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(b64_encoded_path)
```
jodh-intel, line 87: Rather than hard-coding the prefix in this function, it would be better to specify the value as a parameter. That would make unit testing much easier. You could then do something like this:

```rust
// XXX: This function can be fully unit tested
fn join_path_with_prefix(prefix: &str, volume_path: &str) -> PathBuf { ... }

// XXX: This wrapper function will be called by the main code. It can also be tested, but only when running as root.
fn join_path(volume_path: &str) -> PathBuf {
    join_path_with_prefix(KATA_DIRECT_VOLUME_ROOT_PATH, volume_path);
}
```

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    shim_client.get(&req_url).await?;
+    let response = shim_client.get(&req_url).await?;
+    // turn body into string
+    let body = hyper::body::to_bytes(response.into_body()).await?;
+    let body = String::from_utf8(body.to_vec()).unwrap();
+    Ok(body)
+}
+
+// join_path joins user provided volumepath with kata direct-volume root path
+fn join_path(volume_path: &str) -> PathBuf {
```
jodh-intel, line 85: - What if `volume_path == ""`?
- Should `volume_path` be an absolute path (starting with `/`)? This would be preferred from a security perspective, but if the value should be absolute, this function should return `Result<PathBuf>` so it can detect bad paths like `../../foo`.

Adding table driven tests are great for shaking out and handling corner cases and invalid values like this: https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md#rust-table-driven-tests

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    let b64_encoded_path = base64::encode(volume_path.as_bytes());
+    Path::new(KATA_DIRECT_VOLUME_ROOT_PATH).join(b64_encoded_path)
+}
+
+// add writes the mount info (json string) of a direct volume into a filesystem path known to Kata Containers.
+pub fn add(volume_path: &str, mount_info: &str) -> Result<()> {
```
jodh-intel, line 91: Same comment about `volume_path`. What if it was `../../some/where/bad`? I'd suggest:

- Looking at https://github.com/kata-containers/kata-containers/tree/main/src/libs/safe-path and using the types available, or ...
- Calling a common setup function in all the functions that use path parameters to check their validity.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+            println!("{}", &result);
+        }
+        DirectVolSubcommand::Resize(args) => {
+            executor::block_on(resize(&args.volume_path, args.resize_size))?;
+        }
+    }
+    Ok(())
+}
+
+async fn resize(volume_path: &str, size: u64) -> Result<()> {
```
jodh-intel, line 42: All paths (in fact all parameters) should be treated as suspicious by default - see comments below.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,212 @@
...
+    let unserialized_mount_info: DirectVolumeMountInfo = serde_json::from_str(mount_info)?;
+
+    let mount_info_file_path = mount_info_dir_path.join(MOUNT_INFO_FILE_NAME);
+    let serialized_mount_info = serde_json::to_string(&unserialized_mount_info)?;
+    fs::write(mount_info_file_path, serialized_mount_info)?;
+    Ok(())
+}
+
+// remove deletes the direct volume path including all the files inside it.
+pub fn remove(volume_path: &str) -> Result<()> {
```
jodh-intel, line 112: What if `volume_path == "/"`?

jodh-intel (commented):
Thanks @tzY15368 ! This is an impressive piece of work :smile: A few comments...

src/tools/kata-ctl/src/ops/volume_ops.rs
tzY15368: Could you give some hint on how to do this?🤔 I'm thinking that one way to do this is to move `mountInfoFileName` and `kataDirectVolumeRootPath` from `src/runtime/pkg/direct-volume/utils.go` to some shared makefile and do code generation upon compilation. 
Or maybe I should just update the katactl's makefile to fetch the actual value of these variables each time katactl is compiled? But this would break the moment the original go code gets changed.

src/tools/kata-ctl/src/ops/volume_ops.rs
tzY15368: Yes I did forget about this 😥 `add` and `remove` needs to operate on dirs created by root,  `stats` and `resize` accesses Unix domain sockets created by root, so yes this is a privileged command😀

src/tools/kata-ctl/src/ops/volume_ops.rs
tzY15368: The join_path function in `add` and `remove` encode `volume_path` into b64 string [also in kata-runtime](https://github.com/kata-containers/kata-containers/blob/cd27ad144e1a111cb606015c5c9671431535e644/src/runtime/pkg/direct-volume/utils.go#L59) so I think some of the mentioned issues can be mitigated, however since b64 permits `/` character, `safe-path::scoped_join`  is indeed necessary😀

src/tools/kata-ctl/src/ops/volume_ops.rs
tzY15368: I think for `stats` and `resize` commands which send HTTP requests to runtime-rs, it would be better if the input validation happend on the `runtime-rs` side, since a malicious user would still be able to forge a malicious request on the client side.

src/tools/kata-ctl/src/ops/volume_ops.rs
jodh-intel: Hi @tzY15368 - just re-read my comment and realised it certainly could be perceived as confusing! :smile: Sorry about that. What I meant was that if the runtime is generating that file, we should use the same definition as the runtime. But if the rust runtime (runtime-rs) isn't yet generating that file (as #5369 is still open) I guess we should put those definitions in `src/lib/kata*` somewhere. But wdyt @Tim-0731-Hzt ?

My worry is that if we create the const values in the kata-ctl tool, they might get re-added for #5369, and _those_ versions might end up changing at some point in the future leading to a disconnect (aka bugs ;)

Tim-0731-Hzt: how about put them in `src/libs/kata_types/src/mount.rs` ?

tzY15368: Thanks for the clarification! 
> how about put them in `src/libs/kata_types/src/mount.rs` ?
Putting them in `kata-types` does make a lot of sense, will do right away 😀

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,290 @@
...
+
+// get_sandbox_id_for_volume finds the id of the first sandbox found in the dir.
+// We expect a direct-assigned volume is associated with only a sandbox at a time.
+pub fn get_sandbox_id_for_volume(volume_path: &str) -> Result<String> {
+    let dir_path = join_path(KATA_DIRECT_VOLUME_ROOT_PATH, volume_path)?;
+    let paths = fs::read_dir(dir_path)?;
+    for path in paths {
+        let path = path?;
+        // compare with MOUNT_INFO_FILE_NAME
+        if path.file_name() != KATA_MOUNT_INFO_FILE_NAME {
```
jodh-intel, line 143: Inverting the logic would remove one level of indentation, which would make the code clearer, I think.

```rust
if path.file_name() == KATA_MOUNT_INFO_FILE_NAME {
    continue;
}

let file_name = path.file_name();

// ...
```

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,290 @@
...
+    #[test]
+    fn test_path_escape() {
+        // L3BhdGgvdG8vdm9s is the golang output of
+        // b64.URLEncoding.EncodeToString([]byte("/path/to/vol"))
+        let path = String::from("/path/to/vol");
+        let path_bytes = path.as_bytes();
+        let got = base64::encode(path_bytes);
+        let expected = String::from("L3BhdGgvdG8vdm9s");
+        assert!(got.eq(&expected));
+    }
```
jodh-intel, line 175: I'm not really sure why this test is required as it's not testing Kata code...?

But what we could use is a `test_join_path()` function that would pass in various volume paths and validate that the base64-encoded result is as expected.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,290 @@
...
+    let shim_client = runtimes::MgmtClient::new(sandbox_id, Some(TIMEOUT))?;
+    let response = shim_client.get(&req_url).await?;
+    // turn body into string
+    let body = hyper::body::to_bytes(response.into_body()).await?;
+    let body = String::from_utf8(body.to_vec())?;
+    Ok(Some(body))
+}
+
+// join_path joins user provided volumepath with kata direct-volume root path
+fn join_path(prefix: &str, volume_path: &str) -> Result<PathBuf> {
```
jodh-intel, line 89: Could you add some table-based unit tests for this function?

jodh-intel (commented):
Thanks @tzY15368 - a few comments...

src/tools/kata-ctl/src/ops/volume_ops.rs
tzY15368: I think this is added in https://github.com/kata-containers/kata-containers/pull/5467/commits/9b01957f8fc23bf60e5c2292763b3bd93185a90e 😀

src/tools/kata-ctl/src/ops/volume_ops.rs
tzY15368: This was added initially to verify that the rust port would behave exactly as it would in go, but since we have tests for join_path, this is indeed no longer needed. I'll remove them right away.

src/runtime-rs/crates/runtimes/src/shim_mgmt/client.rs
```diff
@@ -74,13 +70,19 @@ impl MgmtClient {
             .uri(url)
             .header("content-type", content_type)
             .body(body)?;
-        let work = self.client.request(req);
+        return self.send_request(req).await;
+    }
+
+    async fn send_request(&self, req: Request<Body>) -> Result<Response<Body>> {
+        let msg: String = format!("Request ({:?}) to uri {:?}", req.method(), req.uri());
```
tzY15368, line 77: @jodh-intel Hi James, thanks for the review, it really helped ironing out the details. I have one minor question with how logging in Rust should work tho. I couldn't allocate error messages with request info after the request was sent due to the borrowing mechanism in Rust, so I had to pre-build the logging message in line 77. This is more or less cumbersome since most of the time the request would take the OK branch, could you perhaps share some advice on this?😀

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,294 @@
...
+        let tests = &[
+            TestData {
+                rootfs: root_fs_str,
+                volume_path: "",
+                result: Err(anyhow!("volume path must not be empty")),
+            },
+            TestData {
+                rootfs: root_fs_str,
+                volume_path: "../../etc/passwd",
+                result: Ok(root_fs.join("Li4vLi4vZXRjL3Bhc3N3ZA==")),
```
jodh-intel, line 236: This is fine, but for clarity I'd consider creating a few variables to hold the original string version of `volume_path`, then calling `base64::encode()` so that the only visible strings are the human-readable ones. Something like:

```rust
let relative_passwd_path = "../../etc/passwd";
let b64_relative_passwd_path = base64::encode(relative_passwd_path.clone());

// ...

let tests = &[
    TestData {
        rootfs: root_fs_str,
        volume_path: relative_passwd_path,
        result: Ok(root_fs.join(b64_relative_passwd_path),
    }

    // ...
```

jodh-intel (commented):
Thanks @tzY15368 - A few nits...

Also, building this PR (now) fails for me due to (https://github.com/kata-containers/kata-containers/pull/5637), so please can you rebase to pull in the latest changes to `main`.

src/tools/kata-ctl/Cargo.toml
```diff
@@ -19,6 +19,18 @@ serde_json = "1.0.85"
 thiserror = "1.0.35"
 privdrop = "0.5.2"
 nix = "0.25.0"
+hyper = { version = "0.14.20", features = ["stream", "server", "http1"] }
+url = "2.3.1"
+futures = "0.3.24"
+base64 = "0.13.0"
+tempfile = "3.3.0"
+serde = { version = "1.0", features = ["derive"] }
```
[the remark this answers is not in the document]
tzY15368: > Thanks @tzY15368 - A few nits...
> 
> Also, building this PR (now) fails for me due to (#5637), so please can you rebase to pull in the latest changes to `main`.

Thanks for the suggestions!  I did a rebase yesterday, but the build still fails for me. Here's the way to reproduce the error:
1. `git checkout main`
2. add `runtimes = { path = "../../runtime-rs/crates/runtimes" } ` to kata-ctl/Cargo.toml
3. The error message is 
```
error[E0308]: mismatched types
   --> /home/tzyuan/.cargo/registry/src/github.com-1ecc6299db9ec823/dbs-interrupt-0.2.1/src/kvm/mod.rs:218:30
    |
218 |             .set_gsi_routing(irq_routing)
    |                              ^^^^^^^^^^^ expected struct `kvm_bindings::x86::bindings::kvm_irq_routing`, found struct `kvm_irq_routing`
    |
    = note:      expected reference `&kvm_bindings::x86::bindings::kvm_irq_routing`
            found mutable reference `&mut kvm_irq_routing`
    = note: perhaps two different versions of crate `kvm_bindings` are being used?
```

jodh-intel (commented):
Thanks @tzY15368 - A few more comments...

Also, I think this still needs a rebase?

jodh-intel (approved):
Thanks @tzY15368.

lgtm!

jodh-intel:
@amshinde, @cmaf, @dborquez - ptal.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+async fn stats(volume_path: &str) -> Result<Option<String>> {
+    let sandbox_id = get_sandbox_id_for_volume(volume_path)?;
+    let mount_info = get_volume_mount_info(volume_path)?;
+
+    let req_url =
+        url::form_urlencoded::Serializer::new(String::from(runtimes::DIRECT_VOLUMN_STATS_URL))
+            .append_pair(runtimes::DIRECT_VOLUMN_PATH_KEY, &mount_info.device)
+            .finish();
+
+    // sid: sandbox id
```
liubin, line 83: Nit: remove this

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// add writes the mount info (json string) of a direct volume into a filesystem path known to Kata Containers.
+pub fn add(volume_path: &str, mount_info: &str) -> Result<Option<String>> {
+    let mount_info_dir_path = join_path(KATA_DIRECT_VOLUME_ROOT_PATH, volume_path)?;
+
+    // create directory if missing
+    fs::create_dir_all(&mount_info_dir_path)?;
+
+    if !mount_info_dir_path.is_dir() {
```
liubin, line 110: Nit: is `mount_info_dir_path.is_dir()` needed? Line 108 should ensure that the `mount_info_dir_path` is a directory.

src/tools/kata-ctl/src/ops/volume_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+use nix;
+use reqwest::StatusCode;
+use safe_path;
+use std::{fs, path::PathBuf, time::Duration};
+use url;
+
+use agent::ResizeVolumeRequest;
+use runtimes;
+
+const TIMEOUT: Duration = Duration::from_millis(200);
```
liubin, line 22: Storage operations may take a lot of time, as a timeout value, 200ms may be too small.

jodh-intel:
Hi @tzY15368 - please could you update to resolve the conflict?

jodh-intel:
Hi @tzY15368 - please could you rebase this PR to resolve the conflicts?

Tim-0731-Hzt (approved):
Thanks @tzY15368

liubin:
Is this one ready to merge? @Ji-Xinyou @tzY15368

tzY15368:
Hi @liubin I looked into the failed CIs and I think none of errors are introduced by this PR. I think this is ready to merge😄

Issue #5477, opened 2022-10-20
CC | Avoid rebuilding the OVMF / TDVF builder image


fidencio:
The container used to build the OVMF / TDVF artefacts has been built on every single PR, which adds a reasonable time to the CI and doesn't help with the reproducibility of the artefacts.

To avoid such issues, let's store the OVMF / TDVF builder image in the Kata Containers quay.io, re-use it whenever is needed, and make sure to upload a new image whenever a change happens in its Dockerfile.

PR #5482, opened 2022-10-20
CC | Avoid rebuilding the containers used for building the artefacts

fidencio:
This PR covers a work done for avoiding building, on every pull-request, containers used to build the artefacts we produce.

This basically covers the following builders:
* kata-deploy (the container that will spawn the other containers)
  * #5475 
* kernel
  * #5476
* ovmf / tdvf
  * #5477
* shim-v2
  * #5478
* td-shim
  * #5479
* virtiofsd
  * #5480
* QEMU
  * #5476 

This work doesn't cover:
* rootfs builders
* initrd builders
* agent builder

The image upload is done to the https://quay.io/kata-containers/cc-builders registry on every change done to the Dockerfile used to build the artefacts.

For now we have no expiration policy set, but that's a consideration for a different issue / PR.

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -56,6 +56,7 @@ docker run \
 	--env AA_KBC="${AA_KBC:-}" \
 	--env KATA_BUILD_CC="${KATA_BUILD_CC:-}" \
 	--env INCLUDE_ROOTFS="$(realpath "${INCLUDE_ROOTFS:-}" 2> /dev/null || true)" \
+	--env PUSH_TO_REGISTRY="${PUSH_TO_REGISTRY:-"no"}" \
 	-v "${kata_dir}:${kata_dir}" \
 	--rm \
 	-w ${script_dir} \
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -16,6 +16,8 @@ kata_deploy_create="${script_dir}/kata-deploy-binaries.sh"
 uid=$(id -u ${USER})
 gid=$(id -g ${USER})
 
+source "${script_dir}/../../scripts/lib.sh"
+
 if [ "${script_dir}" != "${PWD}" ]; then
 	ln -sf "${script_dir}/build" "${PWD}/build"
 fi
@@ -37,7 +39,9 @@ if [ ! -d "$HOME/.docker" ]; then
 	remove_dot_docker_dir=true
 fi
 
-docker build -q -t build-kata-deploy \
+container_image="${CC_BUILDER_REGISTRY}:build-kata-deploy-$(get_last_modification ${kata_dir} ${script_dir})"
+
+docker pull "${container_image}" || docker build -q -t "${container_image}" \
 	--build-arg IMG_USER="${USER}" \
 	--build-arg UID=${uid} \
 	--build-arg GID=${gid} \
@@ -60,7 +64,7 @@ docker run \
 	-v "${kata_dir}:${kata_dir}" \
 	--rm \
 	-w ${script_dir} \
-	build-kata-deploy "${kata_deploy_create}" $@
+	"${container_image}" "${kata_deploy_create}" $@
 
 if [ $remove_dot_docker_dir == true ]; then
 	rm -rf "$HOME/.docker"
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
@@ -41,12 +41,15 @@ fi
 
 container_image="${CC_BUILDER_REGISTRY}:build-kata-deploy-$(get_last_modification ${kata_dir} ${script_dir})"
 
-docker pull "${container_image}" || docker build -q -t "${container_image}" \
-	--build-arg IMG_USER="${USER}" \
-	--build-arg UID=${uid} \
-	--build-arg GID=${gid} \
-	--build-arg HOST_DOCKER_GID=${docker_gid} \
-	"${script_dir}/dockerbuild/"
+docker pull "${container_image}" || \
+	(docker build -q -t "${container_image}" \
+		--build-arg IMG_USER="${USER}" \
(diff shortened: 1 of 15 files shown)


stevenhorsman (approved):
This looks great - the idea that the build containers will be built as part of the cc-payload-after-push GHA and therefore not need to be done in the CI code anymore (unless get_last_modification returns a commit that hasn’t been ‘cached’ should hopefully save a lot of time. Thanks!

fidencio:
I'm adding a `do-not-merge` to give @wainersm's a chance to review this as well on Monday, as he's been working on something related to this as well.

Thanks for the review @stevenhorsman!

tools/packaging/scripts/lib.sh
```diff
@@ -121,3 +122,14 @@ get_last_modification() {
...
 }
+
+# $1 - The tag to be pushed to the registry
+# $2 - "yes" to use sudo, "no" otherwise
+push_to_registry() {
+	local tag="${1}"
+	local use_sudo="${2:-"yes"}"
+
+	if [ "${PUSH_TO_REGISTRY}" == "yes" ]; then
+		[ "${use_sudo}" == "yes" ] && sudo docker push ${tag} || docker push ${tag}
```
wainersm, line 133: This is a good example of `A && B || C` is not equal to `if A then B else C`. Because if B fails `sudo docker push ${tag}` then it will run C (`docker push ${tag}`) as well.

For example:
```
$ tag=foobar
$ use_sudo=yes
$ [ "${use_sudo}" == "yes" ] && sudo docker push ${tag} || docker push ${tag}
Using default tag: latest
The push refers to repository [docker.io/library/foobar]
An image does not exist locally with the tag: foobar
Using default tag: latest
The push refers to repository [docker.io/library/foobar]
An image does not exist locally with the tag: foobar

```

fidencio: Ack, expanded that locally!

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
```diff
@@ -37,7 +39,9 @@ if [ ! -d "$HOME/.docker" ]; then
 	remove_dot_docker_dir=true
 fi
 
-docker build -q -t build-kata-deploy \
+container_image="${CC_BUILDER_REGISTRY}:build-kata-deploy-$(get_last_modification ${script_dir}/dockerbuild/Dockerfile)"
```
wainersm, line 42: I didn't see the upcoming commits yet, so this comment might not apply: shouldn't `get_last_modification` return `${sha}-dirty` when the file is modified but not commited? Otherwise, if I want to test a change I did locally in the Dockerfile then I will need to necessarily make a commit.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
```diff
@@ -41,12 +41,15 @@ fi
...
-	"${script_dir}/dockerbuild/"
+docker pull "${container_image}" || \
+	(docker build -q -t "${container_image}" \
+		--build-arg IMG_USER="${USER}" \
+		--build-arg UID=${uid} \
+		--build-arg GID=${gid} \
+		--build-arg HOST_DOCKER_GID=${docker_gid} \
+		"${script_dir}/dockerbuild/" && \
+ 	 # No-op unless PUSH_TO_REGISTRY is exported as "yes"
+	 push_to_registry "${container_image}" "no")
```
wainersm, line 52: This seems the only push where it doesn't need sudo. Why? Asking just to understand :)

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
fidencio: Hmm.  I didn't think about this case, as generally when I develop I do that committing things.

I can add a logic to have a `dirty`, no issues.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
fidencio: I also don't know why this one doesn't need sudo, but I've decided to not change that to avoid complicating things. :-)

tools/packaging/static-build/shim-v2/build.sh
```diff
@@ -12,18 +12,22 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
 
+source "${script_dir}/../../scripts/lib.sh"
 
 GO_VERSION=${GO_VERSION}
 RUST_VERSION=${RUST_VERSION}
 
 DESTDIR=${DESTDIR:-${PWD}}
 PREFIX=${PREFIX:-/opt/kata}
-container_image="shim-v2-builder"
+container_image="${CC_BUILDER_REGISTRY}:shim-v2-${GO_VERSION}-${RUST_VERSION}-$(get_last_modification ${script_dir}/Dockerfile)"
```
wainersm, line 22: Good point. It is not a change on dockerfile that will require a re-build of the image.

Nit: maybe prepend ${GO_VERSION} with `go` and `rust` likewise, i.e., `shim-v2-go${GO_VERSION}-rust${RUST_VERSION}`

fidencio: One easy way to avoid that is checking for any change in the folder where the files live.  That would cover this case.

tools/packaging/scripts/lib.sh
```diff
@@ -119,5 +119,8 @@ get_config_version() {
 get_last_modification() {
 	local file="${1}"
 
-	echo "$(git log -1 --pretty=format:"%H" ${file})"
+	dirty=""
+	[ $(git status --porcelain | grep modified | grep "${file}") ] && dirty="-dirty"
```
wainersm, line 124: With git 2.34.3 on Fedora 34 `git status --porcelain` returns the modified lines with prefix `M` instead of `modified`.

tools/packaging/static-build/shim-v2/build.sh
wainersm: It is a good solution.

fidencio:
@wainersm, I've addressed all your comments, thanks for the review!

fidencio:
It seems I broke the `get_last_modification` with the fix up.  I will fix this later Today.

wainersm:
@fidencio I re-reviewed it, looks great. Feel free to merge it after the CI pass.

Issue #5462, opened 2022-10-19
CC | Do **not** build the runtime-rs as part of the CCv0 branch / tests / payload


fidencio:
Regardless of how awesome is the fact we have a shim-v2 with an embedded VMM, we could save some time by **not** building it, as this is not used as part of the CCv0 branch (yet).

wainersm:
That's a good idea @fidencio , mainly because the runtime-rs build takes considerable time as I reported in https://github.com/kata-containers/kata-containers/issues/5449

PR #5483, opened 2022-10-20
CC | packaging: Don't build runtime-rs if no RUST_VERSION is provided

fidencio:
As the CCv0 effort is not using the runtime-rs, let's add a mechanism to avoid building it.

The easiest way to do so, is to simply *not* build the runtime-rs if the RUST_VERSION is not provided, and then not providing the RUST_VERSION as part of the cc-shim-v2-tarball target.

Fixes: #5462

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -149,9 +149,7 @@ install_cc_qemu() {
 #Install all components that are not assets
 install_cc_shimv2() {
 	GO_VERSION="$(yq r ${versions_yaml} languages.golang.meta.newest-version)"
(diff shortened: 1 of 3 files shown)


stevenhorsman (approved):
The logic flow and naming of install_go_rust.sh bothers me a little bit, as it isn't super obvious to understand that it won't install rust depending on an env var set several layers away, but I don't think that's a blocker. Hopefully it should help reduce the CI time further, so thank you!

fidencio:
> The logic flow and naming of install_go_rust.sh bothers me a little bit, as it isn't super obvious to understand that it won't install rust depending on an env var set several layers away, but I don't think that's a blocker. Hopefully it should help reduce the CI time further, so thank you!

I agree, for now this is a "hack" for the CCv0 branch.  Once we move everything to `main` this is one bit that we should re-evaluate on how it's done.  For now, I've tried to keep the changes to the minimum ones possible.

PR #5508, opened 2022-10-25
CC | Fix up issues with cached images

fidencio:
lib.sh: Fix get_last_modification()

The ${file} path is an absolute path, as /home/fidencio/..., while the
result of the `git status --porcelain` is a path relative to the
${repo_root_dir}.  Because of this, the logic to adding `-dirty` to the
image name would never work.

Let's fix this by removing the ${repo_root_dir} from the ${file} when
grepping for it.

---

Revert "packaging: Add infra to push the kata-deploy builder image"

This reverts commit fe8b246ae4f3945fc5816cefcee5cf67bf233a59.

The reason this has to be reverted is because we cannot cache an image
that has a specific user, uid, gid, docker_host_id, and expect that to
work equally on different machines.  Unfortunately, this is one of the
images that cannot be cached at all.

---

Revert "packaging: Use existing image for the kata-deploy-build"

This reverts commit c1aac0cdeab66b2ee7f2e2fd822d837c3dcd544f.

The reason this has to be reverted is because we cannot cache an image
that has a specific user, uid, gid, docker_host_id, and expect that to
work equally on different machines.  Unfortunately, this is one of the
images that cannot be cached at all.

---

initramfs: Build dependencies as part of the Dockerfile

This will help to not have to build those on every CI run, and rather
take advantage of the cached image.

### DIFF
--- a/tools/packaging/scripts/lib.sh
+++ b/tools/packaging/scripts/lib.sh
@@ -126,7 +126,7 @@ get_last_modification() {
 	git config --global --add safe.directory ${repo_root_dir} &> /dev/null
 
 	dirty=""
-	[ $(git status --porcelain | grep "${file}" | wc -l) -gt 0 ] && dirty="-dirty"
+	[ $(git status --porcelain | grep "${file#${repo_root_dir}/}" | wc -l) -gt 0 ] && dirty="-dirty"
 
 	echo "$(git log -1 --pretty=format:"%H" ${file})${dirty}"
 }
--- a/tools/packaging/static-build/initramfs/build.sh
+++ b/tools/packaging/static-build/initramfs/build.sh
@@ -32,7 +32,7 @@ package_output_dir="${package_output_dir:-}"
 [ -n "${lvm2_repo}" ] || die "Failed to get lvm2 repo"
 [ -n "${lvm2_version}" ] || die "Failed to get lvm2 version"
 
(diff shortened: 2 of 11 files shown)


tools/packaging/scripts/lib.sh
```diff
@@ -126,7 +126,7 @@ get_last_modification() {
 	git config --global --add safe.directory ${repo_root_dir} &> /dev/null
 
 	dirty=""
-	[ $(git status --porcelain | grep "${file}" | wc -l) -gt 0 ] && dirty="-dirty"
+	[ $(git status --porcelain | grep "${file#${repo_root_dir}/}" | wc -l) -gt 0 ] && dirty="-dirty"
```
wainersm, line 129: Good catch. I missed that on my review.

tools/packaging/static-build/initramfs/build.sh
```diff
@@ -32,7 +32,7 @@ package_output_dir="${package_output_dir:-}"
 [ -n "${lvm2_repo}" ] || die "Failed to get lvm2 repo"
 [ -n "${lvm2_version}" ] || die "Failed to get lvm2 version"
 
-container_image="${CC_BUILDER_REGISTRY}:initramfs-cryptsetup-${cryptsetup_version}-lvm2-${lvm2_version}-$(get_last_modification ${repo_root_dir} ${script_dir})"
+container_image="${CC_BUILDER_REGISTRY}:initramfs-cryptsetup-${cryptsetup_version}-lvm2-${lvm2_version}-$(get_last_modification ${repo_root_dir} ${script_dir})-$(uname -m)"
```
wainersm, line 35: yeah, another good catch...

PR #5510, opened 2022-10-26
runtime-rs: generate config files with the default target

bergwolf:
Right now it is not generated with a simple `make`.


snir911 (approved):

fidencio (approved):
lgtm, thanks @bergwolf!

PR #5513, opened 2022-10-26
versions: bump golangci-lint version

bergwolf:
There is little point to maintain backward compatibility for golangci-lint. Let's just use a unified version of it.


snir911 (approved):
thanks!

dborquez (approved):
lgtm, thanks @bergwolf

Issue #5514, opened 2022-10-26
docs: Fix volumeMounts in usage example


cmaf:
In the [SGX documentation](https://github.com/kata-containers/kata-containers/blob/main/docs/use-cases/using-Intel-SGX-and-kata.md#usage) under the usage section, the containers section should have the `volumeMounts` as well. Otherwise, the volume is not mounted and the enclave device is not available, causing the demo job report an error in the logs.

PR #5515, opened 2022-10-26
docs: Fix volumeMounts in SGX usage example

cmaf:
The /dev/sgx is not mounted and the enclave is not available, causing the demo job to report an error in the logs. Add volumeMounts to container in order to have the device available in the container.

Fixes: #5514


fidencio:
cc @mithy

fidencio (approved):
lgtm, thanks @cmaf!

dborquez (approved):
LGTM, thans @cmaf

mythi:
The fundamental problem is that the helloworld image is too old. The right fix would be to get rid of the mounts and use better sample image.

PR #5516, opened 2022-10-26
DO-NOT-MERGE: clh: Support QAT

fidencio:
This patch is more a request for opinions / help then a patch itself, and it tries to add support to QAT accelerators, using Cloud Hypervisor, with the newest QAT device plugin container provided by Intel.

There are several things here that I don't fully understand and that I'd like to get the help from someone more experienced (bergwolf, zvonko, mikko, archana) on this.

First thing, the way the QAT device plugin container works is that it requires the QAT device to be plugged behind an IOMMU bar.  In order to do so with Cloud Hypervisor, we have to add a second PCI segment, and on this PCI segment plug in the QAT device.  We need to do this because either all the devices in that PCI segmnet use IOMMU, or none of the devices do.

Looking at the agent error below, it makes me think that the agent is only scanning the 0000 bus, and it should be expanded to scan other buses.
```
Events:
  Type     Reason     Age              From               Message
  ----     ------     ----             ----               -------
  Normal   Scheduled  11s              default-scheduler  Successfully assigned default/openssl-qat-engine-dpgithub-june16-clh to 984fee00bf41
  Normal   Pulled     9s               kubelet            Successfully pulled image "localhost/openssl-qat-engine-dpgithub-june16:latest" in 10.788116ms
  Normal   Pulling    5s (x2 over 9s)  kubelet            Pulling image "localhost/openssl-qat-engine-dpgithub-june16:latest"
  Normal   Pulled     5s               kubelet            Successfully pulled image "localhost/openssl-qat-engine-dpgithub-june16:latest" in 11.126763ms
  Warning  Failed     2s (x2 over 6s)  kubelet            Error: CreateContainer failed: Timeout after 3s waiting for uevent PciMatcher { devpath: "/devices/pci0000:00/0000:00:01.0" }: unknown
```

However, one thing that bugs me quite a lot is that if we simply don't pass the device to the agent, as forced in this patch, everything simply works as expected and the QAT device plugin container will do everything needed magically, and things will work as expected.  With this in mind, I started questioning whether actually scanning for more buses would be the right approach, instead of simply filtering out such devices.  But then, if we filter out the devices, how to do that in a programatically way? :-)

So, this is where I hit my knowledge limitation, and where I'd appreciate feedback and ideas from others.

QAT plugin container: https://hub.docker.com/r/intel/intel-qat-plugin

Fixes: #0000

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -488,6 +488,16 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
 		}
 	}
 
+	if clh.config.IOMMU {
+		if clh.vmconfig.Platform == nil {
+			clh.vmconfig.Platform = chclient.NewPlatformConfig()
+		}
+
+		clh.vmconfig.Platform.SetNumPciSegments(2)
+		iommuSegments := []int32{1}
+		clh.vmconfig.Platform.SetIommuSegments(iommuSegments)
+	}
+
 	// Create the VM memory config via the constructor to ensure default values are properly assigned
 	clh.vmconfig.Memory = chclient.NewMemoryConfig(int64((utils.MemUnit(clh.config.MemorySize) * utils.MiB).ToBytes()))
 	// shared memory should be enabled if using vhost-user(kata uses virtiofsd)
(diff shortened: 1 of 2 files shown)


fidencio:
Let me tag @zvonkok, @bergwolf, @mythi, @amshinde, @sboeuf, @rbradford to kick off the discussion.

zvonkok:
@fidencio Just for understanding you're running nested IOMMU, since the DP assumes a QAT behind an IOMMU. The end-goal is to run the DP or a workload that can leverage QAT?

zvonkok:
If the DP is providing the extended resource to the cluster, how are the workloads going to consume the QAT device that is published via DP?

mythi:
> @fidencio Just for understanding you're running nested IOMMU, since the DP assumes a QAT behind an IOMMU. The end-goal is to run the DP or a workload that can leverage QAT?

The end goal is to run a workload both on bare metal and in kata the same way. The workload (`qatlib` HAL) checks that the device it consumes belongs to an IOMMU group and is the only device in it. The device plugin exposes `/dev/vfio/*` devices.

zvonkok:
@mythi Ok, then the goal is not to run the DP but the workload; I am asking because we have the very same use case with the GPUs. Our goal was to run the same CUDA container on bare metal and Kata. 

We're passing through the PF or VF of a GPU  into the VM and setting. `vfio_mode="guest-kernel"`, package the drivers in the guest OS image so that it appears to the container as an "usual" GPU device. The NVIDIA container-toolkit will do all the necessary steps to make the GPU available in the container. 

I am not too familiar with QAT but could you not create a VF, pass it through into the VM and do the needed configuration in the guest OS (what the DP is doing behind the back) and have the container consume the resulting instance?

Issue #5517, opened 2022-10-26
CC | Allow passing a specific builder image for the components that are caching the builder image


fidencio:
Right now we have support for caching the builder image for the following components:
* initramfs
* kernel
* ovmf
* qemu
* shim-v2
* td-shim
* virtiofsd

Let's allow the user to specify a builder image to be used when building those components, as this is a very first step towards reproducible builds.

PR #5518, opened 2022-10-26
CC | packaging: Allow passing a container builder to the scripts

fidencio:
This, combined with the effort of caching builder images *and* only performing the build itself inside the builder images, is the very first step for reproducible builds for the project.

Reproducible builds are quite important when we talk about Confidential Containers, as users may want to verify the content used / provided by the CSPs, and this is the first step towards that direction.

Fixes: #5517


stevenhorsman (approved):
LGTM. Is the idea that we could export the env vars that point to the various builder container images in the release workflow (or similar)?

fidencio:
> LGTM. Is the idea that we could export the env vars that point to the various builder container images in the release workflow (or similar)?

Right now we're not there yet, but I hope one day we can do a release and say:
* Components were built using:
  * Image A
  * image B
  * ...

And then when someone goes back to debug a problem, or simply to rebuild the artefacts by their own (let s say, to audit the components being used), then they can use the images mentioned by the release notes and end up with exactly the same binaries.

Makes sense?

stevenhorsman:
> > LGTM. Is the idea that we could export the env vars that point to the various builder container images in the release workflow (or similar)?
> 
> Right now we're not there yet, but I hope one day we can do a release and say:
> 
> * Components were built using:
>   
>   * Image A
>   * image B
>   * ...
> 
> And then when someone goes back to debug a problem, or simply to rebuild the artefacts by their own (let s say, to audit the components being used), then they can use the images mentioned by the release notes and end up with exactly the same binaries.
> 
> Makes sense?

Cool, that addition of the build details in the release notes so we know what to reproduce is the bit I was missing. It makes total sense. Thanks!

Issue #5522, opened 2022-10-26
 Segmentation fault on Rust compilation of virtiofsd targeting s390x-unknown-linux-gnu


BbolroC:
# Description of problem

The static built binary for `virtiofsd` does not work due to segmentation fault. It hinders `s390x` from using the file system.

How to reproduce an error (tested on Ubuntu20.04):

1. Install necessary packages:

```
$ sudo apt install && sudo apt install -y build-essential libcap-ng-dev libseccomp-dev
```

2. Install rust:

```
$ curl https://sh.rustup.rs -sSf | sh
$ source "$HOME/.cargo/env"
```

3. Clone a repository

```
$ git clone --depth 1 -b v1.3.0 https://gitlab.com/virtio-fs/virtiofsd
$ cd virtiofsd
```

4. Configure environment variables:

```
$ export RUSTFLAGS='-C target-feature=+crt-static -C link-self-contained=yes'
$ export LIBSECCOMP_LINK_TYPE=static
$ export LIBSECCOMP_LIB_PATH=/usr/lib/s390x-linux-gnu
$ export LIBCAPNG_LINK_TYPE=static
$ export LIBCAPNG_LIB_PATH=/usr/lib/s390x-linux-gnu
``` 

5. Build a source:

```
$ cargo build --release --target s390x-unknown-linux-gnu
```

6. Run a executable

```
$ ./target/s390x-unknown-linux-gnu/release/virtiofsd
Segmentation fault (core dumped)
```

# Expected result

A proper output like

```
$ ./target/s390x-unknown-linux-gnu/release/virtiofsd
error: The following required arguments were not provided:
    --fd <fd>
    --socket <socket>
    --socket-path <socket-path>

USAGE:
    virtiofsd --cache <cache> --fd <fd> --inode-file-handles=<inode-file-handles> --log-level <log-level> --sandbox <sandbox> --seccomp <seccomp> --socket <socket> --socket-path <socket-path> --thread-pool-size <thread-pool-size>                                                    

For more information try --help
```

# Actual result

Segmentation fault (stated above)

# Further information

Here is a dump log via `valgrind`:

```
==13103== Memcheck, a memory error detector
==13103== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==13103== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==13103== Command: target/s390x-unknown-linux-gnu/release/virtiofsd
==13103== 
[...]

PR #5523, opened 2022-10-26
virtiofsd: Not use "link-self-contained=yes" on s390x

BbolroC:
The compile option `link-self-contained=yes` asks rustc to use C library startup object files that come with the compiler, which are not available on the target `s390x-unknown-linux-gnu`. A build does not contain any startup files leading to a broken executable entry point (causing segmentation fault). This PR is to make the compile option able to be skipped for `s390x`. 

Fixes: #5522


hbrueckner:
Thanks @BbolroC !

LGTM

snir911 (approved):

Issue #5178, opened 2022-09-16
CC | agent: Allow the aa_kbc_params to be passed through in the command line


stevenhorsman:
**Which feature do you think can be improved?**

Currently the only way to select and pick between the different kbc options is using the `aa_kbc_params` parameter in the kata agent config.toml file, which is 'baked into' the rootfs. This means that if a user wanted to switch between KBC (or stop using one at all), or update their KBS URI they'd need to rebuild the kata guest image, which isn't that simple.

**How can it be improved?**

If we allowed the option to be passed through the kata-container's configuration's kernel_params field with a new  `agent.aa_kbc_params` parameter, this would enable us to make those changes just from the host side without having to rebuild the image.

I think this is probably required for any remote attestation scenario we want to support in the first release, e.g. with EAA, but I don't know if that is in scope.
I'm not 100% sure how the interaction with the offline SEV KBC and the simple KBS works. I think it is remote attestation, but it gets the URI and config via other fields in the kata-containers config (e.g. `guest_pre_attestation_proxy`), but @ryansavino & @fitzthum can correct me and/or confirm.

**Additional Information**

The kernel_params passed from the command line are measured, so I thought that this would be fine from a trust model point of view, but @fitzthum has mentioned that:
> The kbc parameter actually isn't trusted so really we should find a way to pass it such that it doesn't change the measurement.

Unfortunately I don't know enough about the measurement and what should be in the TCB to understand if there are big issues with this approach. I naively thought that if we updated the config to talk to a different remote KBC URI we'd want to have that as part of the measurement, but it seems this assumption isn't correct, so we might want to take a different approach.

fitzthum:
> I'm not 100% sure how the interaction with the offline SEV KBC and the simple KBS works. I think it is remote attestation, but it gets the URI and config via other fields in the kata-containers config (e.g. guest_pre_attestation_proxy), but @ryansavino & @fitzthum can correct me and/or confirm.

For the first release we are using the `offline_sev_kbc`, which is remote attestation via `simple_kbs` but it's offline attestation, meaning that there is no persistent communication between the guest and the kbs. Basically we just inject some secrets at boot and then the KBC reads them. So the `offline_sev_kbc` doesn't have any special parameters. It's the shim that communicates with the KBS. 

[...]

Issue #5173, opened 2022-09-15
agent: Let agent kernel_params override config file


stevenhorsman:
**Which feature do you think can be improved?**

At the moment the way that the agent config file works (based on my understanding at least) is that if an agent config files is specified on the kata configuration `kernel_params` then all the other kernel_params (either specified directly, or derived from other config, like the `debug_console_vport`) will be ignored.

**How can it be improved?**

The agent config toml in written into the guest rootfs image, which means that any users who want to modify it need to re-build the guest image, which isn't straight forward, it also is misleading/confusing when other kata configuration settings are ignored.

I think that a better & more flexible solution is for agent configsin kernel_params cmdline to have higher priority which can overwrite the agent config in rootfs. For the confidential containers case, my understanding is that the kernel_params cmdline are measured for attestation purposes, so I don't think there is a significant security gap introduced by this approach.

@sameo - is there something I've missed here?

sameo:
@stevenhorsman I think the order should be, in decreasing priority:

1. The agent config file passed from the `cmdline`
2. The kernel_params, if there are no agent config file passed through the `cmdline`
3. A default agent config file if neither `kernel_params` nor an agent config file path is specified.

I *think* we don't support step 3 at the moment, right?

sameo:
> @stevenhorsman I think the order should be, in decreasing priority:
> 
> 1. The agent config file passed from the `cmdline`
> 2. The kernel_params, if there are no agent config file passed through the `cmdline`
> 3. A default agent config file if neither `kernel_params` nor an agent config file path is specified.
> 
> I _think_ we don't support step 3 at the moment, right?

I'll take that back, I think config changes passed through the cmdline should take the top level precedence as it gives more flexibility and an easier user experience wo compromising security as the cmdline will be measured.

So, I'm ok with what you initially proposed:

> I think that a better & more flexible solution is for agent configs in kernel_params cmdline to have higher priority which can overwrite the agent config in rootfs.

wainersm:
hi @stevenhorsman! thanks for bringing this discussion up.

My cents:

Months ago I submitted a PR to document the agent configuration options (https://github.com/kata-containers/kata-containers/pull/4016). At that time a comment of @jodh-intel caught my attention (https://github.com/kata-containers/kata-containers/pull/4016#issuecomment-1110840875); let me copy the comment here:
```
[...]

PR #5533, opened 2022-10-28
CC | config: Export aa_kbc_params to be set in guest kernel command line

arronwy:
As we discussed in #5178, user need set aa_kbc_params config without modify kata guest image, since kernel params is also measured in TEE boot flow, we make aa_kbc_params can be parsed through kernel cmdline.

Fixes: #5178

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -27,6 +27,7 @@ const CONTAINER_PIPE_SIZE_OPTION: &str = "agent.container_pipe_size";
 const UNIFIED_CGROUP_HIERARCHY_OPTION: &str = "agent.unified_cgroup_hierarchy";
 const CONFIG_FILE: &str = "agent.config_file";
 const CONTAINER_POLICY_FILE: &str = "agent.container_policy_file";
+const AA_KBC_PARAMS: &str = "agent.aa_kbc_params";
 const HTTPS_PROXY: &str = "agent.https_proxy";
 const NO_PROXY: &str = "agent.no_proxy";
 const ENABLE_DATA_INTEGRITY: &str = "agent.data_integrity";
@@ -316,6 +317,7 @@ impl AgentConfig {
                 get_container_policy_path_value
             );
 
+            parse_cmdline_param!(param, AA_KBC_PARAMS, config.aa_kbc_params, get_string_value);
             parse_cmdline_param!(param, HTTPS_PROXY, config.https_proxy, get_url_value);
             parse_cmdline_param!(param, NO_PROXY, config.no_proxy, get_string_value);
             parse_cmdline_param!(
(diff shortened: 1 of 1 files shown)


GabyCT (approved):

fidencio (approved):
lgtm, thanks @arronwy!

@stevenhorsman, mind to also provide your feedback?

src/agent/src/config.rs
```diff
@@ -30,6 +30,7 @@ const CONTAINER_POLICY_FILE: &str = "agent.container_policy_file";
 const HTTPS_PROXY: &str = "agent.https_proxy";
 const NO_PROXY: &str = "agent.no_proxy";
 const ENABLE_DATA_INTEGRITY: &str = "agent.data_integrity";
+const AA_KBC_PARAMS: &str = "agent.aa_kbc_params";
```
stevenhorsman, line 33: Nit: It might be worth considering moving this line above `const HTTPS_PROXY: &str = "agent.https_proxy";` so that it is consistent with the ordering in the existing code that reads this from the agent config toml file e.g.:
```
#[derive(Debug)]
pub struct AgentConfig {
    pub debug_console: bool,
...
    pub container_policy_path: String,
    pub aa_kbc_params: String,
    pub https_proxy: String,
...
}
```

This applies to most of the changes here, but I didn't want to spam this comment in every line!

fitzthum (approved):
Nice. Do we also need to fix #5173 to use this when we have config file set?

stevenhorsman (approved):
Thanks for the PR Arron, it looks good!
The only thing I would have done different in this PR is changing the order of the code, which I left a Nit to try and explain. The existing code has the AA_KBS_PARAMS logic between `container_policy_path` and `https_proxy`, so that is the natural place that I would look for it everywhere, whereas the new code is added to the end (which is logically and fair). That's just my preferences, so not necessary to change before this is merged!

src/agent/src/config.rs
arronwy: Thanks @stevenhorsman , make sense for me, fixed now :-)

stevenhorsman:
> Nice. Do we also need to fix #5173 to use this when we have config file set?

Hey Tobin - sorry I missed this. In short I think yes. Given at the moment we are only using the toml file for setting this parameter and the endpoints (and we only have it in the rootfs for the `offline_sev_kbc` KBC in the 'normal' kata-payload), then I'm not sure if we immediately need this for the 0.2 release. I think we should consider removing `aa_kbc_params` from a 'default' agent config we create now as well and adjust our CI tests to use this new functionality, but that is probably another issue.

fitzthum:
> I think we should consider removing aa_kbc_params from a 'default' agent config we create now.

I think just allowing the config to be overwritten might actually be the easiest option.

I think we might just need to delete this line https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/config.rs#L249 and that's it. I will submit a PR for it tomorrow if nobody else has gotten to it by then.

Issue #5507, opened 2022-10-25
kata-deploy for arm64


liyimeng:
I try to deploy kata on am arm64 machine using kata-deploy, but it seem the default image quay.io/kata-containers/kata-deploy:3.0.0 is amd64 only, do I miss something?

Thanks a lot in advanced!

stevenhorsman:
This has been address now the kata-deploy is multi-arch

PR #5534, opened 2022-10-28
kata-deploy: fix build for arm64

liyimeng:
Current build script hard written to x86_64.

### DIFF
--- a/tools/packaging/kata-deploy/Dockerfile
+++ b/tools/packaging/kata-deploy/Dockerfile
@@ -4,18 +4,9 @@
 
 # Specify alternative base image, e.g. clefos for s390x
 ARG IMAGE
-FROM ${IMAGE:-registry.centos.org/centos}:7
-ARG KATA_ARTIFACTS=./kata-static.tar.xz
+FROM ${IMAGE:-docker.io/centos}:7
+ARG KATA_ARTIFACTS=./local-build/kata-static.tar.xz
 ARG DESTINATION=/opt/kata-artifacts
-
-COPY ${KATA_ARTIFACTS} ${WORKDIR}
-
-RUN \
-yum -y update && \
-yum -y install xz && \
-yum clean all && \
-mkdir -p ${DESTINATION} && \
-tar xvf ${KATA_ARTIFACTS} -C ${DESTINATION}
 
 # hadolint will deny echo -e, heredocs don't work in Dockerfiles, shell substitution doesn't work with $'...'
 RUN \
@@ -26,4 +17,6 @@ echo "gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packa
 yum -y install kubectl && \
 yum clean all
 
+RUN mkdir -p ${DESTINATION}
+ADD ${KATA_ARTIFACTS} ${DESTINATION}/
 COPY scripts ${DESTINATION}/scripts
--- a/tools/packaging/static-build/firecracker/build-static-firecracker.sh
+++ b/tools/packaging/static-build/firecracker/build-static-firecracker.sh
@@ -37,5 +37,5 @@ git fetch
 git checkout ${firecracker_version}
 sudo ./tools/devtool --unattended build --release
 
-ln -sf ./build/cargo_target/x86_64-unknown-linux-musl/release/firecracker ./firecracker-static
-ln -sf ./build/cargo_target/x86_64-unknown-linux-musl/release/jailer ./jailer-static
+ln -sf ./build/cargo_target/$(uname -m)-unknown-linux-musl/release/firecracker ./firecracker-static
+ln -sf ./build/cargo_target/$(uname -m)-unknown-linux-musl/release/jailer ./jailer-static
--- a/tools/packaging/kata-deploy/.dockerignore
+++ b/tools/packaging/kata-deploy/.dockerignore
@@ -1,1 +1,1 @@
-local-build
+local-build/build


liyimeng:
This need a arm64 machine to build natively, not cross compilation tested yet.

liyimeng:
https://github.com/kata-containers/kata-containers/issues/5507

liyimeng:
Also update kata-deploy base images into docker.io/centos:7. which have multi-arch suported. 
Meanwhile,  optimised kata-deploy images size.

tools/packaging/kata-deploy/Dockerfile
```diff
@@ -26,4 +17,6 @@ echo "gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packa
 yum -y install kubectl && \
 yum clean all
 
+RUN mkdir -p ${DESTINATION}
+ADD ${KATA_ARTIFACTS} ${DESTINATION}/
```
liyimeng, line 21: CI tools dose not like this:

```
tools/packaging/kata-deploy/Dockerfile:21 DL3020 error: Use COPY instead of ADD for files and folders
[1410](https://github.com/kata-containers/kata-containers/actions/runs/3345467724/jobs/5566985626#step:11:1411)
ERROR: failed to check Dockerfile 'tools/packaging/kata-deploy/Dockerfile'
```

Not sure why this policy is enforced. Using ADD instead here, docker will automatically extract files into target, hence reduce the size of the images with about 100MB.

tools/packaging/kata-deploy/Dockerfile
```diff
@@ -4,19 +4,10 @@
 
 # Specify alternative base image, e.g. clefos for s390x
 ARG IMAGE
-FROM ${IMAGE:-registry.centos.org/centos}:7
-ARG KATA_ARTIFACTS=./kata-static.tar.xz
+FROM ${IMAGE:-docker.io/centos}:7
```
fidencio, line 7: What's the reason for switching this to docker.io?  registry.centros.org doesn't provide an ARM image?

This should be explained as part of the commit message.

tools/packaging/kata-deploy/Dockerfile
```diff
@@ -4,19 +4,10 @@
...
+ARG KATA_ARTIFACTS=./local-build//kata-static.tar.xz
```
fidencio, line 8: Could you explain why we need this?

tools/packaging/static-build/firecracker/build-static-firecracker.sh
```diff
@@ -37,5 +37,5 @@ git fetch
 git checkout ${firecracker_version}
 sudo ./tools/devtool --unattended build --release
 
-ln -sf ./build/cargo_target/x86_64-unknown-linux-musl/release/firecracker ./firecracker-static
-ln -sf ./build/cargo_target/x86_64-unknown-linux-musl/release/jailer ./jailer-static
+ln -sf ./build/cargo_target/$(uname -m)-unknown-linux-musl/release/firecracker ./firecracker-static
+ln -sf ./build/cargo_target/$(uname -m)-unknown-linux-musl/release/jailer ./jailer-static
```
fidencio, line 41: Thiese chamges would be better to come as a different commit, still par of this very same PR.

fidencio (commented):
I've left a few questions on the PR.

tools/packaging/kata-deploy/Dockerfile
```diff
@@ -4,19 +4,10 @@
...
 ARG DESTINATION=/opt/kata-artifacts
 
-COPY ${KATA_ARTIFACTS} ${WORKDIR}
-
-RUN \
-yum -y update && \
```
fidencio, line 14: Why did you remove this?

We should always be looking for package fixes, which would come from a yum update.

tools/packaging/kata-deploy/Dockerfile
liyimeng: @fidencio  I test it. registry.centos.org/centos is not a manifest tag, i.e. it only provides x86_64 under that tag. docker.io provide all archs, hence go with docker.io

tools/packaging/kata-deploy/Dockerfile
liyimeng: @fidencio I test it. registry.centos.org/centos is not a manifest tag, i.e. it only provides x86_64 under that tag. docker.io provide all archs, hence go with docker.io

tools/packaging/kata-deploy/Dockerfile
liyimeng: Make it east to for developer. With this change, developer just run docker build after all staic files built, no need an extra step to move the static file into target place. Less work, less problem.

tools/packaging/kata-deploy/Dockerfile
liyimeng: After replacing COPY with ADD, this become necessary. Remove it to save image size. 
See reference https://github.com/kata-containers/kata-containers/pull/5534#discussion_r1009103543

tools/packaging/kata-deploy/Dockerfile
liyimeng: addressed with new commit message.

tools/packaging/static-build/firecracker/build-static-firecracker.sh
liyimeng: It is kind of minor change, I have no opinion. If it is a better kata coding style, not a problem for me.

liyimeng: addressed in new update as suggested @fidencio

tools/packaging/kata-deploy/Dockerfile
liyimeng: If we need fixes, like emergency security fixes, we should update the base image instead of using unpredictable yum update.  In a container world, immutable is important feature. Just my two cents though.

liyimeng:
@fidencio any further comments?

tools/packaging/kata-deploy/Dockerfile
fidencio: And we're working on a different way to address this.

There's been a whole bunch of work in the CCv0 branch, which will end up merged also in the `main` branch, to cache the images used and provide those as part of the release process, so users can easily reproduce the build.

For now, please, leave it as it's.

tools/packaging/kata-deploy/Dockerfile
fidencio: This seems to be an static check warning that we could ignore.

fidencio:
In general this work looks very good to me.
@wainersm, what's the best way to ignore the hadolint error in a specific file?

tools/packaging/kata-deploy/Dockerfile
fidencio: @liyimeng, any further updates here?

liyimeng: @fidencio Not sure if I am getting you. Do you want to get back line 11-14, which I removed in this commit?

liyimeng:
What to do to get this move on?

tools/packaging/kata-deploy/Dockerfile
fidencio: Yes.

liyimeng: hmm, I still think replace these with line 20 21 as below is more elegant way to do thing. If you insist, I could put pack line line 13 and 14. 

Honestly, If we do care security, this image should run rootless and something other more proper solution. This update dose not add much value here.

tools/packaging/kata-deploy/Dockerfile
liyimeng: @wainersm  may you please help us to get this move on.

liyimeng (commented):
@wainersm Please kindly help us out.

fidencio:
This one has been solved in a different way in the past releases, and we're closing it for now.
Please, feel free to open a new PR in case something was not addressed.

Issue #5539, opened 2022-10-31
runtime-rs: use netns-rs crate to handle network namespace


Tim-0731-Hzt:
use netns-rs crate to handle network namespace

PR #5540, opened 2022-10-31
runtime-rs: use netns-rs to handle netnamespace

Tim-0731-Hzt:
use netns-rs crate to handle network namespace

Fixes:#5539


src/runtime-rs/crates/resource/src/network/network_with_netns.rs
```diff
@@ -74,7 +74,9 @@ impl NetworkWithNetns {
 impl Network for NetworkWithNetns {
     async fn setup(&self, h: &dyn Hypervisor) -> Result<()> {
         let inner = self.inner.read().await;
-        let _netns_guard = netns::NetnsGuard::new(&inner.netns_path).context("net netns guard")?;
+        if let Some(netns) = &inner.netns_option {
+            netns.enter().context("failed to enter network namespace")?;
```
liubin, line 78: Will `netns` be dropped after exit from `if let Some()` scope? Should the `e.endpoint.attach()` run in the netns?

Tim-0731-Hzt: I think it doesn't matter, `netns.enter()` will put the calling thread into the `netns`. Yes, all runtime the threads which do the operation using `netlink` should be set in `netns`

quanweiZhou: I'am agree with @liubin , the `netns` should drop after exit, and his problem occurs many times in patch.

Tim-0731-Hzt: The thing is after handling the link, we need to set the runtime threads to back to the previous namespace

Tim-0731-Hzt: The thing is we need to set the thread back to previous namespace after handling the link.

PR #5546, opened 2022-10-31
CC | runtime: Enable kernel hashes for all SEV guests

### LINKED ISSUE #5307: CC SEV: Allow empty keysets
Today using SEV with CC requires you to use a KBS even if you are using an unencrypted container. The only way to skip connecting to the KBS is to turn of `guest_pre_attestation`, but this will remove the `kernel_hashes_enabled` argument from the QEMU command line so you will have to switch firmware from the AmdSev package to the standard Ovmf package. This is not ideal. 

Instead, we should update the shim to skip connecting to the KBS if the keyset is empty. We might also want to require the GUID to be empty. This might make more sense when we switch to the online KBC.

CC: @ryansavino

jimcadden:
This commit adds the `kernel-hashes=on` flag to the QEMU command line for all SEV guests (previously, this was only enabled for SEV guests with `guest_pre_attestation=on`. This change allows the AmdSev firmware to be used for both encrypted and non-encrypted container images.

**Note:** 
This change makes the AmdSev OVMF build a requirement for all SEV guests. The standard host OVMF package will no longer work.

Fixes #5307 

Signed-off-by: Jim Cadden <jcadden@ibm.com>

### DIFF
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -349,6 +349,7 @@ func (q *qemuAmd64) appendSEVObject(devices []govmmQemu.Device, firmware, firmwa
 				CBitPos:         cpuid.AMDMemEncrypt.CBitPosition,
 				ReducedPhysBits: cpuid.AMDMemEncrypt.PhysAddrReduction,
 				SevPolicy:       config.Policy,
+				SevKernelHashes: true,
 			}), "", nil
 	}
 }


fitzthum (approved):
Nice. This turned out to be simpler than expected. 

If it isn't already clear, this will allow people to launch unencrypted images with CC+SEV without having to use a KBS. Right now you have to use a KBS even if you don't need any secrets.

jimcadden:
@stevenhorsman any idea what might be going wrong with the static check? 

Here's the error:

> sandbox::tests::test_find_process' panicked at 'called `Result::unwrap()` on an `Err` value: fail to create cgroup manager with path /some_id: Bad file descriptor (os error 9)', src/sandbox.rs:720:14

stevenhorsman:
> @stevenhorsman any idea what might be going wrong with the static check?
> 
> Here's the error:
> 
> > sandbox::tests::test_find_process' panicked at 'called `Result::unwrap()` on an `Err` value: fail to create cgroup manager with path /some_id: Bad file descriptor (os error 9)', src/sandbox.rs:720:14

Hey Jim, It looks like you are about 70 commits behind `CCv0`, so I'd try rebasing and hopefully that will resolve the issue?

fidencio (approved):
Approving on behalf of @fitzthum.

fidencio:
I'm merging this as it's right now, bypassing the static checks issue which is being tracked on https://github.com/kata-containers/kata-containers/issues/5585 and has a possible solution already proposed https://github.com/kata-containers/kata-containers/pull/5586

jimcadden:
Thank you @stevenhorsman and @fidencio

PR #5552, opened 2022-11-01
Sig ver param

stevenhorsman:
Add `enable_signature_verification` config support

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -31,6 +31,7 @@ const AA_KBC_PARAMS: &str = "agent.aa_kbc_params";
 const HTTPS_PROXY: &str = "agent.https_proxy";
 const NO_PROXY: &str = "agent.no_proxy";
 const ENABLE_DATA_INTEGRITY: &str = "agent.data_integrity";
+const ENABLE_SIGNATURE_VERIFICATION: &str = "agent.enable_signature_verification";
 
 const DEFAULT_LOG_LEVEL: slog::Level = slog::Level::Info;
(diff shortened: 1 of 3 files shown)


src/agent/src/config.rs
```diff
@@ -179,6 +182,7 @@ impl Default for AgentConfig {
             https_proxy: String::from(""),
             no_proxy: String::from(""),
             data_integrity: false,
+            enable_signature_verification: true,
```
fitzthum, line 185: It seems like https://github.com/kata-containers/tests/issues/5228 is saying that the default should be false?

stevenhorsman: Nope - the default is true (for security reasons), but we turn it off with the default kata-config for backwards compatibility

fidencio (approved):
lgtm, thanks @stevenhorsman!

@arronwy, mind to also take a look?

stevenhorsman:
> @arronwy, mind to also take a look?

That would be great. I've heard that image-rs are considering making signature verification enabled by default soon, so I think this would allow that change to happen seamlessly for us, but let me know if you can spot something different.

src/agent/src/config.rs
fitzthum: I wouldn't exactly call this security critical, but I guess it's not too big of a deal either way.

fitzthum (approved):

arronwy (approved):
Yes, Thanks @stevenhorsman , for security consideration `image-rs` will enable it by default, with this PR, user can easy to switch on/off this feature if needed.

Issue #5555, opened 2022-11-01
runk: Cannot kill containers using crictl


ManaSugi:
# Description of problem

Cannot kill containers using crictl due to the following error message.

```
ERROR: container 9163a10e962146bfc9e72d8d549b51956d7bd69429a9a66852b16b19b037948c can't be killed because it is Stopped\n: unknown"
```

This error occurs because `runk kill --all` fails when the container already is stopped.
A low-level container runtime should skip the error handling when `kill` operation was called with `--all` option.

This bug was introduced at the commit https://github.com/kata-containers/kata-containers/commit/a6fbaac1bdc001ee88b4e0d11e1eb0435e8722d6, so we need to back to the previous behavior.

# Expected result

```bash
$ sudo crictl run -r runk container.json sandbox.json
$ sudo crictl --debug rmp --force --all
DEBU[0000] get runtime connection
Stopped sandbox 360cf8381fdb085494d592a959b8ed56c9a3e2f519f53ebe5985aaba9f72f0e5
Removed sandbox 360cf8381fdb085494d592a959b8ed56c9a3e2f519f53ebe5985aaba9f72f0e5
```

# Actual result

```bash
$ sudo crictl run -r runk container.json sandbox.json
$ sudo crictl --debug rmp --force --all
DEBU[0000] get runtime connection
E1101 19:48:11.733237 2417138 remote_runtime.go:248] "StopPodSandbox from runtime service failed" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded" podSandboxID="7a754f033ccd669e1031a3c9f38b769e84886b0937795c53bcec9518a253c729"
stopping the pod sandbox "7a754f033ccd669e1031a3c9f38b769e84886b0937795c53bcec9518a253c729" failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded

$ journalctl --no-pager -e -t containerd
[...]

PR #5556, opened 2022-11-01
runk: Ignore an error when calling kill cmd with --all option

ManaSugi:
Ignore an error handling that is triggered when the kill command is called
with `--all option` to the stopped container.

High-level container runtimes such as containerd call the kill command with
`--all` option in order to terminate all processes inside the container
even if the container already is stopped. Hence, a low-level runtime
should allow `kill --all` regardless of the container state like runc.

This commit reverts to the previous behavior.

Fixes: https://github.com/kata-containers/kata-containers/issues/5555

### DIFF
--- a/src/tools/runk/libcontainer/src/container.rs
+++ b/src/tools/runk/libcontainer/src/container.rs
@@ -83,14 +83,6 @@ impl Container {
     }
 
     pub fn kill(&self, signal: Signal, all: bool) -> Result<()> {
-        if self.state == ContainerState::Stopped {
-            return Err(anyhow!(
-                "container {} can't be killed because it is {:?}",
-                self.status.id,
-                self.state
-            ));
-        }
-
         if all {
             let pids = self.processes()?;
             for pid in pids {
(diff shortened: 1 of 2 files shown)


liubin (approved):

liubin:
Thank you @ManaSugi, I [added an issue](https://github.com/kata-containers/tests/issues/5232) in tests repo to ensure we will not run into such a regression.

fidencio (approved):
lgtm, thanks @ManaSugi!

GabyCT (approved):

cyyzero (approved):
LGTM👍

ManaSugi:
> I [added an issue](https://github.com/kata-containers/tests/issues/5232) in tests repo to ensure we will not run into such a regression.

@liubin Thank you for creating the issue!

> I can help add test cases in [runk-tests.sh](https://github.com/kata-containers/tests/blob/66498273f9ed78641ac40de5d50d3eeff877ed41/integration/containerd/runk/runk-tests.sh).

@cyyzero Awesome! Could you add the test?, thanks!

PR #5562, opened 2022-11-01
CC | Agent: Allow agent config to be overwritten

fitzthum:
I think this is all we need to allow the cmdline to overwrite the agent config. This means we can update config parameters without having to rebuilt the rootfs. Does not change trust model since the command line is measured.

@stevenhorsman  @sameo  @arronwy
Fixes: #5173

Signed-off-by: Tobin Feldman-Fitzthum <tobin@ibm.com>

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -240,6 +240,10 @@ impl AgentConfig {
         let mut config: AgentConfig = Default::default();
         let cmdline = fs::read_to_string(file)?;
         let params: Vec<&str> = cmdline.split_ascii_whitespace().collect();
+
+        let mut using_config_file = false;
+        // Check if there is config file before parsing params that might
+        // override values from the config file.
         for param in params.iter() {
             // If we get a configuration file path from the command line, we
             // generate our config from it.
@@ -247,9 +251,13 @@ impl AgentConfig {
             // or if it can't be parsed properly.
             if param.starts_with(format!("{}=", CONFIG_FILE).as_str()) {
                 let config_file = get_string_value(param)?;
-                return AgentConfig::from_config_file(&config_file);
+                config = AgentConfig::from_config_file(&config_file)?;
+                using_config_file = true;
+                break;
             }
+        }
 
+        for param in params.iter() {
             // parse cmdline flags
             parse_cmdline_param!(param, DEBUG_CONSOLE_FLAG, config.debug_console);
             parse_cmdline_param!(param, DEV_MODE_FLAG, config.dev_mode);
@@ -345,7 +353,9 @@ impl AgentConfig {
         }
 
         // We did not get a configuration file: allow all endpoints.
-        config.endpoints.all_allowed = true;
+        if !using_config_file {
+            config.endpoints.all_allowed = true;
+        }
 
         Ok(config)
     }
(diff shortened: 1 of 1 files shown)


stevenhorsman:
The commit check doesn't like `|` for the subsystem, so we tend to just use that in the issue/PR name, so changing to `agent` should fix that.

stevenhorsman (approved):
This LGTM - thanks Tobin!

fidencio (approved):
lgtm, thanks @fitzthum!

stevenhorsman:
From having a look at the failing tests it look like it is the `Test agent API endpoint can be restricted` test that is failing. As far as I can tell from the logs the exec is successfully blocked, but it's the search for the `ExecProcessRequest is blocked` message that is failing (and I can't see in the logs), but I can't immediately spot what could have caused this logging issue though.

Xynnn007:
Hi, sry I'm not quite familiar with this. I have a small question here. https://github.com/fitzthum/kata-containers/blob/params_cmdline/src/agent/src/config.rs#L228-#L233 (This pr) still allows env to assign a config-path without any overwriting by commandline. Is this OK?

fidencio:
Xynnn comment's , which was deleted for some reason, does make sense and we should consider parsing other arguments when the config comes from the arguments.

src/agent/src/config.rs
```diff
@@ -239,16 +239,22 @@ impl AgentConfig {
...
+        // override values from the config file.
         for param in params.iter() {
             // If we get a configuration file path from the command line, we
             // generate our config from it.
             // The agent will fail to start if the configuration file is not present,
             // or if it can't be parsed properly.
             if param.starts_with(format!("{}=", CONFIG_FILE).as_str()) {
                 let config_file = get_string_value(param)?;
-                return AgentConfig::from_config_file(&config_file);
+                config = AgentConfig::from_config_file(&config_file)?;
```
fidencio, line 252: Xynnn comment was that although we're doing this here, and it's fine, we're totally ignoring the fact that an agent config could be specified via command line arguments, and we'd still have to consider environment variables changing the config file.

With this in mind, we need to have an agreement on precedence.  If an agent config is passed via command-line, but **also** passed via environment variable, which one should be the one considered?  Whichever is chosen, we should warn / info about this happening and make the user aware that one was chosen over the other.

Also, we need to have tests considering this scenario.  Very nice catch from Xynnn!

fidencio (changes requested):

src/agent/src/config.rs
fidencio: Let me take this back, I didn't initially got that the only way the configuration file is passed via `-c config_file` is when using peer pods, as the kernel parameters wouldn't be present in the VM.  This code doesn't conflict with what we currently have and my comment above **does not apply**!

Maybe that was exactly the realisation that Xynnn had and the reason the comment was deleted :-)

fidencio (approved):

src/agent/src/config.rs
```diff
@@ -239,16 +239,22 @@ impl AgentConfig {
...
+                break;
```
fidencio, line 253: I think I understand what's happening here and why we're not being able to block the `ExecProcessRequest`.

Before this patch, we'd get the agent configuration and simply return, not going through the rest of this function.  However, with this patch what happens is that we end up reaching the following line at the end of this function: https://github.com/kata-containers/kata-containers/blob/26075bc9793144341894dc5bdeed40216c974a5a/src/agent/src/config.rs#L351-L352

The simplest way to solve this would adding a way to check whether the default agent config was used, and then only set the `config.endpoints.all_allowed = true;` in case we're did **not** get an agent configuration file as a parameter.

@fitzthum, mind to rework this bit and ensure we cover this as part of our tests?

fidencio (changes requested):

src/agent/src/config.rs
stevenhorsman: I wonder if an even simpler fix might be to just check `if config.endpoints.allowed.is_empty()` around L352. The only problem I can see there is if we've said in the agent config that we don't want any endpoints to be allowed, but I'm pretty sure the agent wouldn't function then

stevenhorsman: @fitzthum - not that I'm in a made rush to get downstream changes in after this or anything before a deadline, but I've created a PR with a unit test and possible fix: https://github.com/fitzthum/kata-containers/pull/1 @fidencio - your input would be welcome too.

PR #5570, opened 2022-11-02
runtime-rs:add hypervisor interface capabilities

Tim-0731-Hzt:
add hypervisor interface capabilities

1. be able to check does hypervisor support use block device, block
device hotplug, multi-queue, and share file

2. be able to set the hypervisor capability of using block device, block
device hotplug, multi-queue, and share file

Fixes: #5569

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
@@ -16,7 +16,10 @@ use dragonball::{
     vm::VmConfigInfo,
 };
 use kata_sys_util::mount;
-use kata_types::config::hypervisor::Hypervisor as HypervisorConfig;
+use kata_types::{
+    capabilities::{Capabilities, CapabilityBits},
+    config::hypervisor::Hypervisor as HypervisorConfig,
+};
 use persist::{sandbox_persist::Persist, KATA_PATH};
 use std::{collections::HashSet, fs::create_dir_all, path::PathBuf};
 
@@ -58,10 +61,19 @@ pub struct DragonballInner {
 
     /// cached block device
     pub(crate) cached_block_devices: HashSet<String>,
+
+    /// dragonball capabilities
+    pub(crate) capabilities: Capabilities,
 }
 
 impl DragonballInner {
     pub fn new() -> DragonballInner {
+        let mut capabilities = Capabilities::new();
+        capabilities.set(
+            CapabilityBits::BlockDeviceSupport
+                | CapabilityBits::BlockDeviceHotplugSupport
+                | CapabilityBits::FsSharingSupport,
+        );
         DragonballInner {
             id: "".to_string(),
             vm_path: "".to_string(),
@@ -74,6 +86,7 @@ impl DragonballInner {
             vmm_instance: VmmInstance::new(""),
             run_dir: "".to_string(),
             cached_block_devices: Default::default(),
+            capabilities,
         }
     }
 
@@ -351,6 +364,7 @@ impl Persist for DragonballInner {
             run_dir: hypervisor_state.run_dir,
             pending_devices: vec![],
             cached_block_devices: hypervisor_state.cached_block_devices,
+            capabilities: Capabilities::new(),
         })
     }
 }
--- a/src/runtime-rs/crates/resource/src/manager_inner.rs
+++ b/src/runtime-rs/crates/resource/src/manager_inner.rs
@@ -71,12 +71,21 @@ impl ResourceManagerInner {
         for dc in device_configs {
             match dc {
                 ResourceConfig::ShareFs(c) => {
-                    let share_fs = share_fs::new(&self.sid, &c).context("new share fs")?;
-                    share_fs
-                        .setup_device_before_start_vm(self.hypervisor.as_ref())
-                        .await
-                        .context("setup share fs device before start vm")?;
-                    self.share_fs = Some(share_fs);
+                    self.share_fs = if self
+                        .hypervisor
+                        .capabilities()
+                        .await?
+                        .is_fs_sharing_supported()
+                    {
+                        let share_fs = share_fs::new(&self.sid, &c).context("new share fs")?;
+                        share_fs
+                            .setup_device_before_start_vm(self.hypervisor.as_ref())
+                            .await
+                            .context("setup share fs device before start vm")?;
+                        Some(share_fs)
+                    } else {
+                        None
+                    };
                 }
                 ResourceConfig::Network(c) => {
                     let d = network::new(&c).await.context("new network")?;
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -102,6 +102,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitmask-enum"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd9e32d7420c85055e8107e5b2463c4eeefeaac18b52359fe9f9c08a18f342b2"
+dependencies = [
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -715,6 +725,7 @@ dependencies = [
 name = "kata-types"
 version = "0.1.0"
 dependencies = [
+ "bitmask-enum",
  "byte-unit",
  "glob",
  "lazy_static",
--- a/src/libs/Cargo.lock
+++ b/src/libs/Cargo.lock
@@ -47,6 +47,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
+name = "bitmask-enum"
+version = "2.1.0"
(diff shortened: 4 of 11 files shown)


src/libs/kata-types/src/capablities.rs
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2019-2022 Alibaba Cloud
+// Copyright (c) 2019-2022 Ant Group
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use bitmask_enum::bitmask;
+
+#[bitmask(u8)]
+enum Bitmask {
```
Ji-Xinyou, line 10: IMO, you should add document comments on these four variables to explain what they mean.

Ji-Xinyou (approved):
Generally LGTM, Thanks @Tim-0731-Hzt

studychao:
Could you add more introduction in issue / PR to explain what does this PR do?

pmores:
I agree with @studychao , to be able to review I'd appreciate an explanation of what the idea is.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -71,12 +71,21 @@ impl ResourceManagerInner {
...
+                        .is_fs_sharing_supported()
+                    {
+                        let share_fs = share_fs::new(&self.sid, &c).context("new share fs")?;
+                        share_fs
+                            .setup_device_before_start_vm(self.hypervisor.as_ref())
+                            .await
+                            .context("setup share fs device before start vm")?;
+                        Some(share_fs)
+                    } else {
+                        None
```
studychao, line 87: If SharerFS is not supported for the hypervisor, why not return an error?

Besides, I think this part is not related to the establishment of capabilities but the actual usage of capabilities in runtime, which should be  separated into other PR.

studychao (commented):
Thanks. A few comments.

src/libs/kata-types/src/capablities.rs
```diff
@@ -0,0 +1,106 @@
...
```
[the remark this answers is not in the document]
Tim-0731-Hzt: This pr is to align with the go runtime, you can refer to [capabilities.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/types/capabilities.go). it will not show all the differencies between the hypervisors, we only call this method when sometimes we need to perform some customized actions for different hypervisors.

src/runtime-rs/crates/resource/src/manager_inner.rs
Tim-0731-Hzt: As we are going to support multiple hypervisors, we need to consider about the use case of firecracker. It doesn't support using virtio-fs, we can only use it devicemapper snapshotter to achieve that.

src/libs/kata-types/src/capablities.rs
Ji-Xinyou: Also, the enum name should not be `BItmask`, it should be something like `CapabilitiyBits`

Tim-0731-Hzt: Agree. Thanks !

src/libs/kata-types/src/capablities.rs
```diff
@@ -0,0 +1,106 @@
...
+//
+
+use bitmask_enum::bitmask;
+
+#[bitmask(u8)]
+enum CapabilityBits {
+    blockDeviceSupport,
+    blockDeviceHotplugSupport,
+    multiQueueSupport,
+    fsSharingSupported,
```
liubin, line 14: - Others are using `Support` suffix, but `fsSharing` is using `Supported`.
- Should they use an uppercase capital? For example `BlockDeviceSupport`

src/libs/kata-types/src/capablities.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+/// Capabilities describe a virtcontainers hypervisor capabilities through a bit mask.
+#[derive(Debug, Clone)]
+pub struct Capabilities {
+    /// flags
```
liubin, line 20: No meaning comment

src/libs/kata-types/src/capablities.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl Capabilities {
+    /// new Capabilities struct
+    pub fn new() -> Self {
+        Capabilities {
+            flags: CapabilityBits { bits: 0 },
+        }
+    }
+
+    /// is_block_device_supported tells if an hypervisor supports block devices.
```
liubin, line 38: Considering future extensions, these `is_xxx` and `set_xxx` can be generated by a macro.
But you can leave it in another PR.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -58,10 +58,17 @@ pub struct DragonballInner {
...
     pub(crate) cached_block_devices: HashSet<String>,
+
+    /// dragonball capabilities
+    pub(crate) capabilities: Capabilities,
 }
 
 impl DragonballInner {
     pub fn new() -> DragonballInner {
+        let mut capabilities = Capabilities::new();
+        capabilities.set_block_device_support();
```
liubin, line 69: Is this reasonable?

```rust
let flags = CapabilityBits::fsSharingSupported | CapabilityBits::blockDeviceSupport | ... ;
capabilities.set(flags);
```

And

```rust
impl Capabilities {
    pub fn set(&mut self, flags: CapabilityBits) {
        self.flags = self.flags.or(flags);
        // or self.flags = flags;
    }
}
```

src/libs/kata-types/src/capablities.rs
Tim-0731-Hzt: fixed

src/libs/kata-types/src/capablities.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
Tim-0731-Hzt: Thanks, fixed

src/runtime-rs/crates/resource/src/manager_inner.rs
pmores: I agree with @Tim-0731-Hzt that the way the code is now we might not be able to return an error here.  However I also share @studychao 's intuition that returning an error would be the right thing to do.

I think the problem might be that we never check AFAICT whether `SharedFsInfo::shared_fs` is `Some()`.  If it is then I believe we should return an error here since that means that someone configured kata to request a shared fs but they're also using a hypervisor that doesn't support it.

In `VirtSandbox::prepare_for_start_sandbox()` we push a `ResourceConfig::ShareFs` instance onto `resource_configs` unconditionally.  I think that if we only did this when `hypervisor_config.shared_fs.shared_fs` was not `None` we could afford return an error here.  We could alternatively also do the check in `ResourceManagerInner` but I believe that that's unnecessarily late and if we want to do the check, then `VirtSandbox` is a clean place to do so.

Does that make sense, or am I completely off the mark?

pmores:
There's a typo in the `src/libs/kata-types/src/capablities.rs` file name.  Normally I'm not strict about typos but I think we might consider fixing this one as it then translates to the name of the module and spreads through the codebase.

Ji-Xinyou:
> There's a typo in the `src/libs/kata-types/src/capablities.rs` file name. Normally I'm not strict about typos but I think we might consider fixing this one as it then translates to the name of the module and spreads through the codebase.

100% agreed.

src/runtime-rs/crates/resource/src/manager_inner.rs
Tim-0731-Hzt: We do check `sharefs` is `Some`, [see this](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs#L39). If the hypervisor doesn't support share-fs, we need to use `copy file` to copy the file into the guest.

pmores: Right, so I'll have to apologise for a bad wording first - when I said "we never check whether `SharedFsInfo::shared_fs` is `Some()`" I really meant "we never check _before the code under review executes_". So you're right, `ShareFsVolume` does check, but I believe that it only runs later on, after `ResourceManagerInner::prepare_before_start_vm()` has exited, is that correct?  If so then its checking can't be helpful for `prepare_before_start_vm()`.

And please don't get me wrong, I'm not arguing that the code is wrong. I'm just trying to a) improve my understanding and b) make sure that the case where a user is running with a hypervisor that doesn't support shared fs but they also set `shared_fd = "virtiofs-<something>"' in their `configuration.toml` is handled reasonably. I believe it should fail early and with a good error message to which end I think it could be good if `SharedFsInfo::shared_fs` was also checked either here in `ResourceManagerInner::prepare_before_start_vm()` along with hypervisor capabilities, or in `VirtSandbox::prepare_for_start_sandbox()` along the line of

Tim-0731-Hzt: Yes, it was checked when try to create a container. 
Yes, we need to check if user config `shared_fd = "virtiofs-<something>` for a hypervisor doesn't support shared-fs, but I think it is better to be handled by the configuration plugin, like [dragonball](https://github.com/kata-containers/kata-containers/blob/main/src/libs/kata-types/src/config/hypervisor/dragonball.rs) and [qemu](https://github.com/kata-containers/kata-containers/blob/main/src/libs/kata-types/src/config/hypervisor/qemu.rs) does.

pmores: Alright, fair enough, thanks @Tim-0731-Hzt !

liubin (approved):

PR #5573, opened 2022-11-02
runtime-rs: blanks filled & fixes made to virtiofsd launch

### LINKED ISSUE #5572: runtime-rs cannot start virtiofsd
This is a known problem, virtiofsd support in runtime-rs is not complete. A fix should at least make it possible for virtiofsd to launch in common scenarios.

pmores:
The 'config' argument to ShareVirtioFsStandalone::new() is now actually used, taking care of an explicit TODO.

If a shared path doesn't exist in ShareVirtioFsStandalone::virtiofsd_args() it is now created instead of returning an error, thus following ShareVirtioFsInline's suit.

The '-o vhost_user_socket=...' command line argument doesn't seem to be supported by newer versions of virtiofsd so we replace it with '--socket-path' which should be functionally equivalent according to docs.

Fixes #5572

Signed-off-by: Pavel Mores <pmores@redhat.com>

### DIFF
--- a/src/runtime-rs/crates/resource/src/share_fs/share_virtio_fs_standalone.rs
+++ b/src/runtime-rs/crates/resource/src/share_fs/share_virtio_fs_standalone.rs
@@ -21,7 +21,7 @@ use tokio::{
 };
 
 use super::{
-    share_virtio_fs::generate_sock_path, utils::get_host_ro_shared_path,
+    share_virtio_fs::generate_sock_path, utils::ensure_dir_exist, utils::get_host_ro_shared_path,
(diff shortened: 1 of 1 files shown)


Tim-0731-Hzt (approved):
Thanks @pmores LGTM

Ji-Xinyou (approved):
Nice catch, Thanks @pmores , LGTM!

pmores:
Thanks for your reviews @Tim-0731-Hzt @Ji-Xinyou !  I just pushed an update but that's just to placate `cargo fmt`, no substantive changes.

liubin (approved):

PR #5574, opened 2022-11-02
CC | clh: Increase the timeouts when using Conf Guests

fidencio:
Launching a pod with measured boot enabled seems to be taking longer than expected with Cloud Hypervisor, which leads to hitting a timeout limit.

Let's double those timeout limits for now.


stevenhorsman (approved):
LGTM Thanks!

cmaf (approved):
thanks @fidencio

fidencio:
We're failing to compile the agent on the TDX machine, but that's not a blocker for this PR, rather an internet issue in the world behind the firewall.

Issue #5563, opened 2022-11-01
CC | Build and use a TDX specific guest image


fidencio:
As we're switching to using EAA KBC with TDX, we should build, distribute, and use an image that supports that.

PR #5580, opened 2022-11-03
CC | config: Add specific config for TDX + EAA KBC

fidencio:
As we're switching TDX to using EAA KBC instead of OfflineFS KBC, let's
add the configuration files needed for testing this before we fully
switch TDX to using such an image.

Fixes: #5563


stevenhorsman (approved):
LGTM and nice to get the images a bit smaller!

fidencio:
I'm merging this as it's right now, bypassing the static checks issue which is being tracked on https://github.com/kata-containers/kata-containers/issues/5585 and has a possible solution already proposed https://github.com/kata-containers/kata-containers/pull/5586

PR #5586, opened 2022-11-04
github: Parallelise static checks

fidencio:
Although introducing an awful amount of code duplication, let's
parallelise the static checks in order to reduce its time and the space
used in the VMs running those.

While I understand there may be ways to make the whole setup less
repetitive and error prone, I'm taking the approach of:
* Make it work
* Make it right
* Make it fast

So, it's clear that I'm only attempting to make it work, and I'd
appreciate community help in order to improve the situation here.  But,
for now, this is a stopgap solution.

JFYI, the time needed for run the tests on the `main` branch went down
from ~110 minutes to ~60 minutes.  Plus, we're not running those on a
single VM anymore, which decreases the chance to hit the space limit.

Reference: https://github.com/kata-containers/kata-containers/actions/runs/3393468605/jobs/5640842041

Ideally, each one of the following tests should be also split into
smaller tests, each test for one component, for instance.
* static-checks
* compiler-checks
* unit-tests
* unit-tests-as-root

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -8,7 +8,7 @@ on:
 
 name: Static checks
 jobs:
-  test:
+  check-vendored-code:
     strategy:
       matrix:
         go-version: [1.16.x, 1.17.x]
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @fidencio.

lgtm as a stop-gap!

liubin:
I'd like to say this is great work.
The most time cost is ut now, likely to take 60m, next step we can make it run-on-change: only run ut in sub-directory that has anything changed, if we only modified the runtime code, the ut of the agent should not run.

stevenhorsman (approved):
As mentioned on slack, I think we might be able to use something like https://docs.github.com/en/actions/creating-actions/creating-a-composite-action to reduce the duplication, but as a short term fix this is fine. Thanks!

Issue #5585, opened 2022-11-04
Static checks failing due to the lack of space in the GitHub Action runner


fidencio:
We've noticed a bunch of failures in the CCv0 branch related to the static checks.  When taking a look at the action we've noticed the surprising message:
```
[test (1.16.x, ubuntu-20.04)](https://github.com/kata-containers/kata-containers/actions/runs/3384930657/jobs/5632798318#step:1:37)
You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 75 MB
```

This should be a good incentive for us to go ahead an break those static checks, which are currently taking ~2 hours, and way too much space, into smaller chunks that can be done in parallel in a shorter amount of time, and hopefully not consuming that much space.

While I understand the need for having this fixed on `main`, and then backported toi the `CCv0` branch, I don't see myself working on this anytime soon, and I'd like to suggests disabling such tests from the `CCv0` branch, at least till the release is over.

PR #5592, opened 2022-11-07
CC | github: Parallelise static checks

fidencio:
Although introducing an awful amount of code duplication, let's parallelise the static checks in order to reduce its time and the space used in the VMs running those.

While I understand there may be ways to make the whole setup less repetitive and error prone, I'm taking the approach of:
* Make it work
* Make it right
* Make it fast

So, it's clear that I'm only attempting to make it work, and I'd appreciate community help in order to improve the situation here.  But, for now, this is a stopgap solution.

JFYI, the time needed for run the tests on the `main` branch went down from ~110 minutes to ~60 minutes.  Plus, we're not running those on a single VM anymore, which decreases the change to hit the space limit.

Reference: https://github.com/kata-containers/kata-containers/actions/runs/3393468605/jobs/5640842041

Ideally, each one of the following tests should be also split into smaller tests, each test for one component, for instance.
* static-checks
* compiler-checks
* unit-tests
* unit-tests-as-root

Fixes: #5585

(cherry picked from commit 40d514aa2c73ddca2ff0e3bb5615d62fd77e5ada)

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -8,7 +8,7 @@ on:
 
 name: Static checks
 jobs:
-  test:
+  check-vendored-code:
     strategy:
       matrix:
         go-version: [1.16.x, 1.17.x]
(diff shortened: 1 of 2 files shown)


fidencio:
While here, let's also backport the snap CI fixes.

PR #5597, opened 2022-11-07
clh: avoid race condition when stopping clh

### LINKED ISSUE #5622: clh: StopVM is called twice when the VM stopped
# Description of problem

StopVM calls ShutdownVMM command triggers virtiofsd quit. As a result StopVM is invoked again from another go-routine at the same time leading to potential issues.

```
level=info msg="Stopping sandbox in the VM" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers subsystem=sandbox
level=info msg="Stopping VM" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers subsystem=sandbox
level=info msg="Stop Sandbox" function=StopVM name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=debug msg="Stopping Cloud Hypervisor" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=info msg="failed to get OOM event from sandbox" error="rpc error: code = Internal desc = \"\"" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=containerd-kata-shim-v2
level=info msg="virtiofsd quits" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=virtiofsd
level=info msg="Stop Sandbox" function=StopVM name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=debug msg="Stopping Cloud Hypervisor" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="clh.VmmPingGet API call failed" error="500 " name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="clh.VmmPingGet API call failed" error="Get \"http://localhost/api/v1/vmm.ping\": dial unix /run/vc/vm/a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4/clh-api.sock: connect: no such file or directory" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="clh.VmmPingGet API call failed" error="Get \"http://localhost/api/v1/vmm.ping\": dial unix /run/vc/vm/a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4/clh-api.sock: connect: no such file or directory" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="clh.VmmPingGet API call failed" error="Get \"http://localhost/api/v1/vmm.ping\": dial unix /run/vc/vm/a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4/clh-api.sock: connect: no such file or directory" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="clh.VmmPingGet API call failed" error="Get \"http://localhost/api/v1/vmm.ping\": dial unix /run/vc/vm/a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4/clh-api.sock: connect: no such file or directory" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="clh.VmmPingGet API call failed" error="Get \"http://localhost/api/v1/vmm.ping\": dial unix /run/vc/vm/a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4/clh-api.sock: connect: no such file or directory" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=debug msg="stop virtiofsDaemon" name=containerd-shim-v2 pid=110797 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=cloudHypervisor
level=warning msg="kill virtiofsd failed" error="no such process" name=containerd-shim-v2 pid=0 sandbox=a8753e39c0a5f0ed661d9899b042a917dbc8968b8ad045c2d54af22079dad3f4 source=virtcontainers/hypervisor subsystem=virtiofsd
```

# Expected result

StopVM should be called only once

# Actual result

StopVM is called twice

alex-matei:
clh: don't try to stop clh multiple times
clh: fast exit from isClhRunning if the process was stopped
clh: return faster with dead Cloud Hypervisor process from isClhRunning
qemu: set stopped only if StopVM is successful
qemu: early exit from Check if the process was stopped

Avoid executing StopVM concurrently when virtiofs dies as a result of clh being stopped in StopVM.
Proactively check if Cloud Hypervisor process is dead in isClhRunning
Set stopped to 1 in clh and qemu only if StopVM returns without an error.
Fast exit from Check for both clh and qemu if the process was stopped, uses atomic operations instead of acquiring a mutex.
This stops isClhRunning from generating a deadlock by trying to reacquire an already-acquired lock when called via StopVM->terminate.

Fixes #5622, #5623, #5624, #5625

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -1405,16 +1405,16 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 		return false, nil
 	}
 
-	if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
-		return false, nil
-	}
-
 	timeStart := time.Now()
 	cl := clh.client()
 	for {
+		err := syscall.Kill(pid, syscall.Signal(0))
+		if err != nil {
+			return false, nil
+		}
 		ctx, cancel := context.WithTimeout(context.Background(), clh.getClhAPITimeout()*time.Second)
-		defer cancel()
-		_, _, err := cl.VmmPingGet(ctx)
+		_, _, err = cl.VmmPingGet(ctx)
+		cancel()
 		if err == nil {
 			return true, nil
 		} else {
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -24,6 +24,8 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -256,6 +258,8 @@ type cloudHypervisor struct {
 	vmconfig        chclient.VmConfig
 	state           CloudHypervisorState
 	config          HypervisorConfig
+	stopped         int32
+	mu              sync.Mutex
 }
 
 var clhKernelParams = []Param{
@@ -1081,9 +1085,21 @@ func (clh *cloudHypervisor) ResumeVM(ctx context.Context) error {
 
 // StopVM will stop the Sandbox's VM.
 func (clh *cloudHypervisor) StopVM(ctx context.Context, waitOnly bool) (err error) {
+	clh.mu.Lock()
+	defer func() {
+		if err == nil {
+			atomic.StoreInt32(&clh.stopped, 1)
+		}
+		clh.mu.Unlock()
+	}()
 	span, _ := katatrace.Trace(ctx, clh.Logger(), "StopVM", clhTracingTags, map[string]string{"sandbox_id": clh.id})
 	defer span.End()
 	clh.Logger().WithField("function", "StopVM").Info("Stop Sandbox")
+	if atomic.LoadInt32(&clh.stopped) != 0 {
+		clh.Logger().Info("Already stopped")
+		return nil
+	}
+
 	return clh.terminate(ctx, waitOnly)
 }
 
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -1401,6 +1401,10 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 
 	pid := clh.state.PID
 
+	if atomic.LoadInt32(&clh.stopped) != 0 {
+		return false, nil
+	}
+
 	if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
 		return false, nil
 	}
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -969,7 +969,7 @@ func (q *qemu) waitVM(ctx context.Context, timeout int) error {
 }
 
 // StopVM will stop the Sandbox's VM.
-func (q *qemu) StopVM(ctx context.Context, waitOnly bool) error {
+func (q *qemu) StopVM(ctx context.Context, waitOnly bool) (err error) {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 	span, _ := katatrace.Trace(ctx, q.Logger(), "StopVM", qemuTracingTags, map[string]string{"sandbox_id": q.id})
@@ -983,7 +983,9 @@ func (q *qemu) StopVM(ctx context.Context, waitOnly bool) error {
 
 	defer func() {
 		q.cleanupVM()
-		q.stopped = true
+		if err == nil {
+			q.stopped = true
+		}
 	}()
 
 	if q.config.Debug && q.qemuConfig.LogFile != "" {
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -21,6 +21,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 	"unsafe"
@@ -109,7 +110,7 @@ type qemu struct {
 
 	nvdimmCount int
 
-	stopped bool
+	stopped int32
 
 	mu sync.Mutex
 }
@@ -976,7 +977,7 @@ func (q *qemu) StopVM(ctx context.Context, waitOnly bool) (err error) {
 	defer span.End()
 
 	q.Logger().Info("Stopping Sandbox")
-	if q.stopped {
+	if atomic.LoadInt32(&q.stopped) != 0 {
 		q.Logger().Info("Already stopped")
 		return nil
 	}
@@ -984,7 +985,7 @@ func (q *qemu) StopVM(ctx context.Context, waitOnly bool) (err error) {
 	defer func() {
 		q.cleanupVM()
 		if err == nil {
-			q.stopped = true
+			atomic.StoreInt32(&q.stopped, 1)
 		}
 	}()
 
@@ -2570,7 +2571,7 @@ func (q *qemu) toGrpc(ctx context.Context) ([]byte, error) {
 func (q *qemu) Save() (s hv.HypervisorState) {
 
 	// If QEMU isn't even running, there isn't any state to Save
-	if q.stopped {
+	if atomic.LoadInt32(&q.stopped) != 0 {
 		return
 	}
 
@@ -2621,6 +2622,10 @@ func (q *qemu) Load(s hv.HypervisorState) {
 }
 
 func (q *qemu) Check() error {
+	if atomic.LoadInt32(&q.stopped) != 0 {
+		return fmt.Errorf("qemu is not running")
+	}
+
 	q.memoryDumpFlag.Lock()
 	defer q.memoryDumpFlag.Unlock()
 


gkurz (commented):
>  clh: avoid race condition when stopping clh
> 
> qemu: early exit from Check if the process was stopped
> 

Hi @alex-matei !

Please have separate commits for clh or qemu. This will ease overall maintenance and backport efforts for downstream consumers.

> When virtiofs dies because clh is stopped from StopVM stop
> calling StopVM again.
> 

Is this fixing a bug ? If so this deserves a separate commit as well since other changes seem to be enhancements.

> Prevent a 10 second timeout in isClhRunning(10) by checking
> that the process has died after each iteration.
> 
> Fast exit from Check for both clh and qemu if the process
> was stopped
> 

You should create an issue and add a `Fixes:` tag to the commit message.

> Signed-off-by: Alexandru Matei <alexandru.matei@uipath.com>

src/runtime/virtcontainers/clh.go
```diff
@@ -1081,9 +1085,20 @@ func (clh *cloudHypervisor) ResumeVM(ctx context.Context) error {
 
 // StopVM will stop the Sandbox's VM.
 func (clh *cloudHypervisor) StopVM(ctx context.Context, waitOnly bool) (err error) {
+	clh.mu.Lock()
+	defer func() {
+		atomic.StoreInt32(&clh.stopped, 1)
```
gkurz, line 1090: This means that the VM will always be considered as stopped when `StopVM()` returns, even if something goes wrong in `clh.terminate(ctx, waitOnly)`... this seems incorrect. This state should only be set if we know the VM has actually stopped.

src/runtime/virtcontainers/clh.go
```diff
@@ -1401,6 +1421,11 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 			clh.Logger().WithError(err).Warning("clh.VmmPingGet API call failed")
 		}
 
+		err = syscall.Kill(pid, syscall.Signal(0))
+		if err != nil {
+			return false, nil
+		}
+
```
gkurz, line 1428: IIUC this is the `Prevent a 10 second timeout` optimization. What about doing this at the beginning of the block ? This will allow to drop the same check that is done before the loop.

Also, this would deserve its own commit.

src/runtime/virtcontainers/clh.go
alex-matei: You are right, I missed that. The same issue I think applies as well to qemu, stopped is set at the end without checking the error.

src/runtime/virtcontainers/clh.go
alex-matei: Correct, i'll fix it.

src/runtime/virtcontainers/qemu.go
```diff
@@ -976,14 +977,14 @@ func (q *qemu) StopVM(ctx context.Context, waitOnly bool) error {
 	defer span.End()
 
 	q.Logger().Info("Stopping Sandbox")
-	if q.stopped {
+	if atomic.LoadInt32(&q.stopped) != 0 {
 		q.Logger().Info("Already stopped")
```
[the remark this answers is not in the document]
alex-matei: This is not a really a bug regarding thread-safety. Instead of acquiring the mutex (q.mu) I used atomic to speed things a bit and to keep the logic the same as the one in clh. Atomic ops are required in clh, StopVM calls terminate isClhRunning while holding the mutex so trying to acquire it again in order to check stopped will trigger a deadlock. I'll move qemu bits to another commit.

src/runtime/virtcontainers/clh.go
alex-matei: Fixed in qemu and clh.

src/runtime/virtcontainers/clh.go
alex-matei: Fixed.

alex-matei:
Thanks @gkurz. I created separate commits for each bug/enhancement and added corresponding issues. Please take another look when you can.

src/runtime/virtcontainers/clh.go
```diff
@@ -1405,17 +1405,16 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
...
 	}
 
-	err := syscall.Kill(pid, syscall.Signal(0))
-	if err != nil {
-		return false, nil
-	}
-
 	timeStart := time.Now()
 	cl := clh.client()
 	for {
```
likebreath, line 1410: I think this commit message (`clh: prevent 10 seconds timeout when checking for clh`) also needs to be refined to explain the purpose of the patch.

To my understanding:

```
clh: Return faster with dead Cloud Hypervisor process from `isClhRunning()`

Through more proactively checking if Cloud Hypervisor process is dead, this patch
provides a faster path for `isClhRunning()`.
```

likebreath (changes requested):
@alex-matei Thank you for the contribution. The changes to clh bits look fine to me, but I think the commit message needs to be rewritten. Please find my detailed comments below.

src/runtime/virtcontainers/clh.go
```diff
@@ -1385,7 +1401,12 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 
 	pid := clh.state.PID
 
-	if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
+	if atomic.LoadInt32(&clh.stopped) != 0 {
+		return false, nil
+	}
+
+	err := syscall.Kill(pid, syscall.Signal(0))
+	if err != nil {
```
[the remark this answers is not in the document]
alex-matei: Fixed.

src/runtime/virtcontainers/clh.go
```diff
@@ -24,6 +24,8 @@ import (
 	"regexp"
```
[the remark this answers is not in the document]
alex-matei: Fixed.

src/runtime/virtcontainers/clh.go
alex-matei: You're right, I changed the commit message.

alex-matei:
Thanks @likebreath. I updated the PR based on your review. Please take a look when you can.

src/runtime/virtcontainers/clh.go
```diff
@@ -1385,7 +1401,8 @@ func (clh *cloudHypervisor) isClhRunning(timeout uint) (bool, error) {
 
 	pid := clh.state.PID
 
-	if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
+	err := syscall.Kill(pid, syscall.Signal(0))
+	if err != nil {
```
likebreath: Nit: this should go into the second commit.

likebreath (commented):
@alex-matei Looking good. I only have one additional nit. I approved the work-flow to run, and will also run CI jobs later when all GH actions passed. Please keep an eye on them and ensure they all pass. Anything else I can help with, please let me know. Thank you.

src/runtime/virtcontainers/clh.go
alex-matei: Fixed.

likebreath (approved):

likebreath:
Approved for the changes to clh related code base. Additional reviews needed for the qemu bits.

Also, running the CI jobs. /test

jodh-intel (approved):
Thanks @alex-matei.

lgtm

GabyCT (approved):

gkurz (approved):
QEMU bits LGTM. Thanks @alex-matei !

Issue #5588, opened 2022-11-05
runtime-rs: bind mount volumes in sandbox level


justxuewei:
Each container has an exclusive mountpoint for the same file on the host. For example, two containers want to use "/home/example/downloads" on the host. Assume that "aaaa" and "bbbb" are those container ids, respectively. The runtime will create two mountpoints on the host runtime directory:

- for "aaaa" container: /run/kata-containers/shared/sandboxes/$pod/rw/passthrough/aaaa-{uuid}-downloads.
- for "bbbb" container: /run/kata-containers/shared/sandboxes/$pod/rw/passthrough/bbbb-{uuid}-downloads.

In this way, the overall architecture looks like this:

<img src="https://user-images.githubusercontent.com/13026316/200104660-b994cefb-acc4-43e4-b801-20a800d7c850.png" width="600px" />

This architecture is straightforward and intuitive. However, one disadvantage is that those mountpoints have their cache, leading to inconsistent data issues when a container writes something into the "downloads". Another container reads something from the "downloads" at the same time. It may not see those modifications.

One possible solution is sharing the same directory in sandbox level. As shown in the below picture, we only create a shared mountpoint for the downloads directory. Thus, modifications are visible to all containers due to a unified cache.

<img src="https://user-images.githubusercontent.com/13026316/200104672-3c5e72b7-254d-4cad-aa32-c3bd5b4f22ba.png" width="600px" />

There is more work to be done on the sandbox side, such as umount the volumes to which no containers reference, etc.

liubin:
This may give you some idea https://github.com/kata-containers/kata-containers/pull/1463

justxuewei:
> This may give you some idea #1463

Thank you, I'll take a look at this.

PR #5607, opened 2022-11-08
runtime-rs: bind mount volumes in sandbox level

justxuewei:
This PR fixes the issue mentioned in #5588. In this pull request all shared volumes, except for rootfs, are bind mounted in sandbox level. That is all volumes referred to the same source on the host are bind mounted only once until no container references to it.

It supports dynamic permission controls as well. For example, when a volume mounted with readonly permission and a new coming container needs readwrite permission, the volume should be upgraded to readwrite permission. At the same time, if readwrite permission is no longer needed, then the volume should be downgraded.

Fixes: #5588 

Signed-off-by: Xuewei Niu <justxuewei@apache.org>

### DIFF
--- a/src/runtime-rs/crates/resource/src/share_fs/mod.rs
+++ b/src/runtime-rs/crates/resource/src/share_fs/mod.rs
@@ -15,10 +15,10 @@ pub use utils::{do_get_guest_path, do_get_guest_share_path, get_host_rw_shared_p
 mod virtio_fs_share_mount;
 use virtio_fs_share_mount::VirtiofsShareMount;
 
-use std::sync::Arc;
+use std::{path::PathBuf, sync::Arc};
 
 use agent::Storage;
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, Context, Ok, Result};
 use async_trait::async_trait;
 use hypervisor::Hypervisor;
 use kata_types::config::hypervisor::SharedFsInfo;
@@ -43,8 +43,16 @@ pub trait ShareFs: Send + Sync {
     async fn setup_device_before_start_vm(&self, h: &dyn Hypervisor) -> Result<()>;
     async fn setup_device_after_start_vm(&self, h: &dyn Hypervisor) -> Result<()>;
     async fn get_storages(&self) -> Result<Vec<Storage>>;
+
+    /// Get mounted info from ShareFs.
+    /// The source is an original path on the host (not in the `/run/kata-containers/...`).
+    async fn get_mounted_info(&self, source: &str) -> Option<MountedInfo>;
+    /// Set mounted info to ShareFS.
+    /// The source is an original path on the host (not in the `/run/kata-containers/...`).
+    async fn set_mounted_info(&self, source: &str, mounted_info: MountedInfo) -> Result<()>;
 }
 
+#[derive(Debug)]
 pub struct ShareFsRootfsConfig {
     // TODO: for nydus v5/v6 need to update ShareFsMount
     pub cid: String,
@@ -54,6 +62,7 @@ pub struct ShareFsRootfsConfig {
     pub is_rafs: bool,
 }
 
+#[derive(Debug)]
 pub struct ShareFsVolumeConfig {
     pub cid: String,
     pub source: String,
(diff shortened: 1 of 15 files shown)


justxuewei:
Hi @liubin, I found some errors making ci failure in compiling Dragonball. However, this pull request didn't change Dragonball at all. How to fix this? Thanks.

justxuewei:
I rebased the latest main branch to fix Dragonball compiling issues, could you help me trigger the ci tests? Thanks @liubin @lifupan

src/runtime-rs/crates/resource/src/share_fs/mod.rs
```diff
@@ -38,13 +38,29 @@ pub const PASSTHROUGH_FS_DIR: &str = "passthrough";
 const RAFS_DIR: &str = "rafs";
 
 #[async_trait]
-pub trait ShareFs: Send + Sync {
+pub trait ShareFs: Send + Sync + Debug {
```
bergwolf, line 41: Why is `Debug` required?

src/runtime-rs/crates/resource/src/share_fs/mod.rs
```diff
@@ -69,10 +86,61 @@ pub struct ShareFsMountResult {
...
+    }
+}
+
 #[async_trait]
-pub trait ShareFsMount: Send + Sync {
+pub trait ShareFsMount: Send + Sync + Debug {
     async fn share_rootfs(&self, config: ShareFsRootfsConfig) -> Result<ShareFsMountResult>;
     async fn share_volume(&self, config: ShareFsVolumeConfig) -> Result<ShareFsMountResult>;
+    /// Upgrade to readwrite permission
+    async fn upgrade(&self, file_name: &str) -> Result<()>;
```
bergwolf, line 139: Please name it explicitly. Something like `upgrade_rw` and `dowgrade_ro` sound better.

src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
```diff
@@ -166,4 +169,36 @@ impl ShareFsMount for VirtiofsShareMount {
...
+        bind_remount(&host_dest, true)
+            .context("remount readonly directory with readonly permission")?;
+        Ok(())
+    }
+
+    async fn umount(&self, file_name: &str) -> Result<()> {
+        let host_dest = do_get_host_path(file_name, &self.id, "", true, false);
+        umount_timeout(&host_dest, 0).context("Umount readonly host dest")?;
+        let host_dest = do_get_host_path(file_name, &self.id, "", true, true);
+        umount_timeout(&host_dest, 0).context("Umount readwrite host dest")?;
```
bergwolf, line 201: The umount event is propagated to the `ro` virtiofs mount subtree. We only need to umount the rw directory bindmount.

src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
```diff
@@ -59,36 +68,91 @@ impl ShareFsVolume {
...
+                            .await
+                            .context("upgrade mount")?;
+                    }
+                    if readonly {
+                        mounted_info.ro_ref_count += 1;
+                    } else {
+                        mounted_info.rw_ref_count += 1;
+                    }
+                    share_fs
+                        .set_mounted_info(&m.source, mounted_info)
```
bergwolf, line 104: Do we need any lock when manipulating mounted_info? There are a whole bunch of race windows here.

src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
```diff
@@ -59,36 +68,91 @@ impl ShareFsVolume {
...
+                        })
+                        .await
+                        .context("mount shared volume")?;
+                    let mounted_info = MountedInfo::new(
+                        PathBuf::from_str(&mount_result.guest_path)
+                            .context("convert guest path")?,
+                        readonly,
+                    );
+                    share_fs
+                        .set_mounted_info(&m.source, mounted_info)
```
bergwolf, line 135: The same locking concern applies here too.

src/runtime-rs/crates/resource/src/share_fs/mod.rs
justxuewei: Adding the `Debug` is for debugging purposes when developing this feature. And it is not required for now. I will remove it from the code.

src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
justxuewei: IMO, we don't need an extra lock for `mounted_info`, because `ResourceManager::inner` whose type is `Arc<RwLock<ResourceManagerInner>>` has a RwLock which ensures we can manipulate it safely.

justxuewei: Yes, you are right. I am working on it.

justxuewei:
Thanks for your reviews @bergwolf. I've pushed a new commit to fix the issues you mentioned above. Please take a look.

bergwolf (approved):
lgtm! Thanks!

lifupan (approved):

PR #5610, opened 2022-11-08
runtime: prevent waiting 50 ms minimum for a process exit

### LINKED ISSUE #5617: Kata runtime should acknowledge as soon as possible when a process exits
Kata runtime waits at least 50 ms for a process before actually checking that the process has exited. The runtime should acknowledge as soon as possible that a process finished execution.

alex-matei:
runtime: prevent waiting 50 ms minimum for a process exit

Use pidfd_open and poll on newer versions of Linux to wait for the process to exit. For older versions fall back to polling every 50 ms.

Fixes #5617 

### DIFF
--- a/src/runtime/virtcontainers/utils/utils.go
+++ b/src/runtime/virtcontainers/utils/utils.go
@@ -346,10 +346,6 @@ func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) b
 	}
 }
 
-func waitForProcessCompletion(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
-	return waitProcessUsingWaitLoop(pid, timeoutSecs, logger)
-}
-
 // WaitLocalProcess waits for the specified process for up to timeoutSecs seconds.
 //
 // Notes:
--- a/src/runtime/virtcontainers/utils/utils.go
+++ b/src/runtime/virtcontainers/utils/utils.go
@@ -375,7 +375,18 @@ outer:
 	if pidRunning {
 		// Force process to die
 		if err = syscall.Kill(pid, syscall.SIGKILL); err != nil {
+			if err == syscall.ESRCH {
+				logger.WithField("pid", pid).Warnf("process already finished")
+				return nil
+			}
 			return fmt.Errorf("Failed to stop process %v: %s", pid, err)
+		}
+
+		for {
+			_, err := syscall.Wait4(pid, nil, 0, nil)
+			if err != syscall.EINTR {
+				break
+			}
 		}
 	}
 
--- a/src/runtime/virtcontainers/utils/utils.go
+++ b/src/runtime/virtcontainers/utils/utils.go
@@ -25,6 +25,8 @@ const cpBinaryName = "cp"
 
 const fileMode0755 = os.FileMode(0755)
 
+const maxWaitDelay = 50 * time.Millisecond
+
 // The DefaultRateLimiterRefillTime is used for calculating the rate at
 // which a TokenBucket is replinished, in cases where a RateLimiter is
 // applied to either network or disk I/O.
@@ -307,11 +309,18 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
 }
 
 func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
-	secs := time.Duration(timeoutSecs)
-	timeout := time.After(secs * time.Second)
+	secs := time.Duration(timeoutSecs) * time.Second
+	timeout := time.After(secs)
+	delay := 1 * time.Millisecond
 
 	for {
-		// Check if the process is running periodically to avoid a busy loop
+		// Wait4 is used to reap and check that a child terminated.
+		// Without the Wait4 call, Kill(0) for a child will always exit without
+		// error because the process isn't reaped.
+		// Wait4 return ECHLD error for non-child processes. Kill(0) is meant
+		// to address this case, once the process is reaped by init process,
+		// the call will return ESRCH error.
+
 		// "A watched pot never boils" and an unwaited-for process never appears to die!
 		waitedPid, err := syscall.Wait4(pid, nil, syscall.WNOHANG, nil)
 
@@ -324,7 +333,12 @@ func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) b
 		}
 
 		select {
-		case <-time.After(50 * time.Millisecond):
+		case <-time.After(delay):
+			delay = delay * 5
+
+			if delay > maxWaitDelay {
+				delay = maxWaitDelay
+			}
 		case <-timeout:
 			logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
 			return true
--- a/src/runtime/virtcontainers/utils/utils.go
+++ b/src/runtime/virtcontainers/utils/utils.go
@@ -306,6 +306,36 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
 	}
 }
 
+func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	secs := time.Duration(timeoutSecs)
+	timeout := time.After(secs * time.Second)
+
+	for {
+		// Check if the process is running periodically to avoid a busy loop
+		// "A watched pot never boils" and an unwaited-for process never appears to die!
+		waitedPid, err := syscall.Wait4(pid, nil, syscall.WNOHANG, nil)
+
+		if waitedPid == pid && err == nil {
+			return false
+		}
+
+		if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
+			return false
+		}
+
+		select {
+		case <-time.After(50 * time.Millisecond):
+		case <-timeout:
+			logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
+			return true
+		}
+	}
+}
+
+func waitForProcessCompletion(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	return waitProcessUsingWaitLoop(pid, timeoutSecs, logger)
+}
+
 // WaitLocalProcess waits for the specified process for up to timeoutSecs seconds.
 //
 // Notes:
@@ -334,43 +364,7 @@ func WaitLocalProcess(pid int, timeoutSecs uint, initialSignal syscall.Signal, l
 		}
 	}
 
-	pidRunning := true
-
-	secs := time.Duration(timeoutSecs)
-	timeout := time.After(secs * time.Second)
-
-	// Wait for the VM process to terminate
-outer:
-	for {
-		select {
-		case <-time.After(50 * time.Millisecond):
-			// Check if the process is running periodically to avoid a busy loop
-
-			var _status syscall.WaitStatus
-			var _rusage syscall.Rusage
-			var waitedPid int
-
-			// "A watched pot never boils" and an unwaited-for process never appears to die!
-			waitedPid, err = syscall.Wait4(pid, &_status, syscall.WNOHANG, &_rusage)
-
-			if waitedPid == pid && err == nil {
-				pidRunning = false
-				break outer
-			}
-
-			if err = syscall.Kill(pid, syscall.Signal(0)); err != nil {
-				pidRunning = false
-				break outer
-			}
-
-			break
-
-		case <-timeout:
-			logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
-
-			break outer
-		}
-	}
+	pidRunning := waitForProcessCompletion(pid, timeoutSecs, logger)
 
 	if pidRunning {
 		// Force process to die
--- a/src/runtime/virtcontainers/utils/utils_linux.go
+++ b/src/runtime/virtcontainers/utils/utils_linux.go
@@ -14,8 +14,10 @@ import (
 	"os"
 	"strings"
 	"syscall"
+	"time"
 	"unsafe"
 
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
@@ -150,4 +152,61 @@ func IsAPVFIOMediatedDevice(sysfsdev string) bool {
 		}
 	}
 	return false
+}
+
+func waitProcessUsingPidfd(pid int, timeoutSecs uint, logger *logrus.Entry) (bool, error) {
+	pidfd, err := unix.PidfdOpen(pid, 0)
+
+	if err != nil {
+		if err == unix.ESRCH {
+			return false, nil
+		}
+
+		return true, err
+	}
+
+	defer unix.Close(pidfd)
+	var n int
+
+	maxDelay := time.Duration(timeoutSecs) * time.Second
+	end := time.Now().Add(maxDelay)
+
+	for {
+		remaining := time.Until(end).Milliseconds()
+		if remaining < 0 {
+			remaining = 0
+		}
+
+		n, err = unix.Poll([]unix.PollFd{{Fd: int32(pidfd), Events: unix.POLLIN}}, int(remaining))
+		if err != unix.EINTR {
+			break
+		}
+	}
+
+	if err != nil || n != 1 {
+		logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
+		return true, err
+	}
+
+	for {
+		err := unix.Waitid(unix.P_PIDFD, pidfd, nil, unix.WEXITED, nil)
+		if err == unix.EINVAL {
+			err = unix.Waitid(unix.P_PID, pid, nil, unix.WEXITED, nil)
+		}
+
+		if err != unix.EINTR {
+			break
+		}
+	}
+	return false, nil
+}
+
+func waitForProcessCompletion(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	pidRunning, err := waitProcessUsingPidfd(pid, timeoutSecs, logger)
+
+	if err == unix.ENOSYS {
+		pidRunning = waitProcessUsingWaitLoop(pid, timeoutSecs, logger)
+	}
+
+	return pidRunning
 }
--- a/src/runtime/virtcontainers/utils/utils_darwin.go
+++ b/src/runtime/virtcontainers/utils/utils_darwin.go
@@ -5,6 +5,12 @@
 
 package utils
 
+import "github.com/sirupsen/logrus"
+
 func GetDevicePathAndFsTypeOptions(mountPoint string) (devicePath, fsType string, fsOptions []string, err error) {
 	return
+}
+
+func waitForProcessCompletion(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	return waitProcessUsingWaitLoop(pid, timeoutSecs, logger)
 }


gkurz:
> runtime: prevent waiting 50 ms minimum for a process exit
> 

Hi @alex-matei ! Not only this drops this extra delay but, most importantly, the use of more idiomatic patterns greatly improve readability.  Thanks for this submission.

For this to get merged, you need to create an issue and add a corresponding `Fixes:` tag to the commit message.

Rest looks good to me.

> Signed-off-by: Alexandru Matei [alexandru.matei@uipath.com](mailto:alexandru.matei@uipath.com)

alex-matei:
Thanks @gkurz. Updated with the corresponding issue.

c3d (approved):
Looks good to me. Thanks @alex-matei

gkurz (dismissed):
LGTM. Thanks @alex-matei !

gkurz:
> I looked at the failing test. The qemu-system-x86_64 parent pid is pid 1 and waiting for it using Process.Wait (which internally uses waitid) doesn't work for non-child processes.

The QEMU process being parented to pid 1 is the consequence of the `-daemonize` argument passed on the command line. This means that it is up to pid 1 to do the housekeeping i.e. call wait() not up to the shim process.

> I will try to rewrite using pidfd_open and fall back to polling using signal 0 for older versions of Linux. (wait4 in current implementation also doesn't work for non-child processes).

In any case, you must keep a path that does call `wait4()` in case where the hypervisor isn't daemonized.

alex-matei:
Updated the code to handle all the cases.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -351,27 +382,61 @@ outer:
...
+			return fmt.Errorf("Failed to send initial signal %v to process %v: %v", initialSignal, pid, err)
 		}
 	}
 
+	pidRunning, err := waitUsingPidfd(pid, timeoutSecs, logger)
+
+	if err == unix.ENOSYS {
+		pidRunning = waitUsingLegacy(pid, timeoutSecs, logger)
+	}
+
```
gkurz, line 439: If the hypervisor isn't daemonized, the shim is the parent of the hypervisor process and it thus **must** `wait4()` or a variant on its child process in some way. That is, the pidfd path must ensure a terminated QEMU child is ripped as well.

When the pidfd points to a child, it is possible to `waitid(P_PIDFD)`. This could be added to `waitUsingPidfd()` but I don't see any API to do that in the go library, apart from `syscall.Syscall6(syscall.SYS_WAITID, ...)`

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,43 +306,74 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
+	errc := make(chan error, 1)
+
+	go func() {
+		p, err := os.FindProcess(pid)
+		if err == nil {
+			_, err = p.Wait()
+		}
+		errc <- err
+	}()
+
```
gkurz, line 362: I missed that during the first review but there's a problem here. Let's suppose the hypervisor process is stopped with kill(SIGSTOP), this `p.Wait()` call will block until the process is runnable again, i.e. at some random time. It doesn't look great that `waitUsingLegacy()` might return and leave the goroutine behind... the result of `p.Wait()` would be lost IIUC.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -351,27 +382,61 @@ outer:
 			var waitedPid int
 
 			// "A watched pot never boils" and an unwaited-for process never appears to die!
-			waitedPid, err = syscall.Wait4(pid, &_status, syscall.WNOHANG, &_rusage)
+			waitedPid, err := syscall.Wait4(pid, &_status, syscall.WNOHANG, &_rusage)
```
gkurz, line 386: Since you keep these periodic non-blocking `wait4()` calls, you don't need the extra goroutine... You should even avoid it since it would be racy and we'd never on which path `wait4()` actually succeeded. Maybe not a bug but definitely confusing.

gkurz:
@alex-matei The pidfd approach looks neat but all systems doesn't support it and you still need a fallback with a `wait4()` based solution for non-daemonized hypervisors and `kill(0)` for daemonized ones. This doesn't really simplify the code in the end, which is a greater concern than gaining 50 ms on a not especially hot path.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Yeah, that goroutine will remain blocked until the process exits, no way to cancel waiting, this a problem in general in linux with waiting for pids.

src/runtime/virtcontainers/utils/utils.go
alex-matei: I don't know if this as a problem, as it is right now WaitLocalProcess is used only to wait for that process to exit, not to reap child processes ("init" will reap them after reparenting). There are a two other places in runtime where syscall.Kill with SIGKILL is called and there is no reaping done.

I could try to add reaping for all cases.

alex-matei:
> @alex-matei The pidfd approach looks neat but all systems doesn't support it and you still need a fallback with a `wait4()` based solution for non-daemonized hypervisors and `kill(0)` for daemonized ones. This doesn't really simplify the code in the end, which is a greater concern than gaining 50 ms on a not especially hot path.

50 ms might not seem a lot, but for short lived pods, like 500ms-1sec, it is  5-10% of the total running time. This function is used in shutdown in almost all hypervisor except qemu, not only startup but also shutdown should be fast. There are scenarios like Kubernetes where the resources are released after the pod dies, so this impacts container density and autoscaling behavior. Pidfd is present since Linux 5.3 and most distributions use newer versions.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Fixed, added reaping for pidfd and processes that are killed.`

src/runtime/virtcontainers/utils/utils.go
alex-matei: I removed that part, and left wait4 & signal 0.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Removed the goroutine.

alex-matei:
Thanks @gkurz for review. I updated the code based on your suggestions, I think i covered all the edge cases.

alex-matei:
Updated to fix the darwin tests.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -334,49 +375,15 @@ func WaitLocalProcess(pid int, timeoutSecs uint, initialSignal syscall.Signal, l
...
-			if waitedPid == pid && err == nil {
-				pidRunning = false
-				break outer
-			}
-
-			if err = syscall.Kill(pid, syscall.Signal(0)); err != nil {
-				pidRunning = false
-				break outer
-			}
-
```
gkurz, line 365: @alex-matei since the purpose of this patch is to prevent a potentially unneeded 50ms delay, what about doing just that and do the checks before the `select {}` ?

alex-matei: Good point, I changed their order.

gkurz: Err... sorry if I wasn't clear. My suggestion is to do just that simple change to the existing code instead of a massive rewrite.

alex-matei: It doesn't completely solve the issue if the hypervisor quits after 10 ms for example. You will still wait for 40 ms.

gkurz: > It doesn't completely solve the issue if the hypervisor quits after 10 ms for example. You will still wait for 40 ms more than needed.

True. I've been thinking a bit harder and I agree it would be good to switch to pidfd as you're proposing, since it fully addresses the issue. But I'd really like this commit to be broken down in smaller pieces, for easier review and overall maintainability. This is generally the first reason for me pushing back :smiley:

I'll add some more comments.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,15 +306,57 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
 	}
 }
 
+func reap(pid int) {
+	var _status syscall.WaitStatus
+	var _rusage syscall.Rusage
+	for {
+		_, err := syscall.Wait4(pid, &_status, unix.WNOHANG, &_rusage)
```
gkurz, line 313: You can pass `nil` instead of the unused `_status` and `_rusage` variables.

gkurz: If the process hasn't exited yet, wait4(WNOHANG) will just return 0 with no error and reap() will return without having reaped anything. You should just pass 0 to the flags.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,15 +306,57 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
+	for {
+		// Check if the process is running periodically to avoid a busy loop
+
+		var _status syscall.WaitStatus
+		var _rusage syscall.Rusage
+		var waitedPid int
+
+		// "A watched pot never boils" and an unwaited-for process never appears to die!
+		waitedPid, err := syscall.Wait4(pid, &_status, syscall.WNOHANG, &_rusage)
+
```
gkurz, line 333: Same remark for _status and _rusage.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,15 +306,57 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
-// - If the initial signal is zero, the specified process is assumed to be
-//   attempting to stop itself.
-// - If the initial signal is not zero, it will be sent to the process before
-//   checking if it is running.
-// - If the process has not ended after the timeout value, it will be forcibly killed.
+//   - If the initial signal is zero, the specified process is assumed to be
+//     attempting to stop itself.
+//   - If the initial signal is not zero, it will be sent to the process before
+//     checking if it is running.
+//   - If the process has not ended after the timeout value, it will be forcibly killed.
```
gkurz, line 359: Can you get rid of this space damage ?

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -334,49 +376,15 @@ func WaitLocalProcess(pid int, timeoutSecs uint, initialSignal syscall.Signal, l
...
-
-			break
-
-		case <-timeout:
-			logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
-
-			break outer
-		}
-	}
+	pidRunning := waitForProcessCompletion(pid, timeoutSecs, logger)
```
gkurz, line 379: First patch would also introduce waitForProcessCompletion() that only calls waitProcessUsingWaitLoop().

src/runtime/virtcontainers/utils/utils_linux.go
```diff
@@ -152,3 +154,50 @@ func IsAPVFIOMediatedDevice(sysfsdev string) bool {
...
+			break
+		}
+	}
+
+	if err != nil || n != 1 {
+		logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
+		return true, err
+	}
+
+	reap(pid)
```
gkurz, line 191: Let's go pidfd all the way and use a variant based on Waitid(P_PIDFD) instead of Wait4() ?

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -334,49 +376,15 @@ func WaitLocalProcess(pid int, timeoutSecs uint, initialSignal syscall.Signal, l
...
 
 	if pidRunning {
 		// Force process to die
 		if err = syscall.Kill(pid, syscall.SIGKILL); err != nil {
 			return fmt.Errorf("Failed to stop process %v: %s", pid, err)
 		}
+
+		reap(pid)
```
gkurz, line 387: Hmm... this looks like a bug fix since without it a force killed hypervisor process will stay zombie until the parent terminates. It definitely deserve its own commit.
Also I think it is possible to come up with a more native implementation for linux using waitid(), i.e. you can open-code the wait4() based varient here.

src/runtime/virtcontainers/utils/utils_linux.go
```diff
@@ -152,3 +154,50 @@ func IsAPVFIOMediatedDevice(sysfsdev string) bool {
...
+
+func waitForProcessCompletion(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	pidRunning, err := waitProcessUsingPidfd(pid, timeoutSecs, logger)
+
+	if err == unix.ENOSYS {
+		pidRunning = waitProcessUsingWaitLoop(pid, timeoutSecs, logger)
+	}
+
+	return pidRunning
+}
```
gkurz, line 203: Final commit would move the trivial waitForProcessCompletion() to the darwin file and introduce the linux one along with the pidfd implementation.

Thanks @alex-matei for taking care of this and I'll try to have it merged in a timely manner.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Fixed.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Fixed.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Fixed.

src/runtime/virtcontainers/utils/utils.go
alex-matei: I split the PR in 3 commits.

src/runtime/virtcontainers/utils/utils_linux.go
alex-matei: Correct, modified to use pidfd for waiting.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Moved the code to a separate commit.

src/runtime/virtcontainers/utils/utils_linux.go
alex-matei: Fixed.

alex-matei:
Thanks @gkurz for review. I made a couple of changes, please take a look.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,15 +306,57 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
+
+		select {
+		case <-time.After(50 * time.Millisecond):
+		case <-timeout:
+			logger.Warnf("process %v still running after waiting %ds", pid, timeoutSecs)
+			return true
+		}
+	}
+}
+
```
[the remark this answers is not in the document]
alex-matei: Done.

src/runtime/virtcontainers/utils/utils_linux.go
alex-matei: On further analysis i changed it back to wait4,  It looks like pidfd_open is introduced during kernel 5.3 and waitid with P_PIDFD is introduced in 5.4. For versions in-between waitid will return EINVAL, which is not what we want.

alex-matei: In the end, I finally settled for waitid with P_PIDFD and if it returns EINVAL it will fall back to waitid with P_PID

gkurz:
Hi @alex-matei ! Great thanks for your changes. We're almost there :smiley: , just a few things to address.

First two commits are missing a changelog and cause the required commit message check to fail :
```
  2a5a854f8a420edd305c46efc34c65461033d225    Commit body is missing
  5d1eba2e8bdd4e7a3ce8bff1e843d96d0b2a12ce    Commit body is missing
```

Arguably, the `runtime: reap force killed processes` commit could have its own issue (and the corresponding Fixes tag in its changelog) since this is a bug fix that could be backported to stable releases, while the rest is rather an enhancement.

Apart from that the changes LGTM and I'll approve when the above remarks have been addressed.

alex-matei:
Hi @gkurz, Thanks! I completely missed that. I updated the commit messages.

gkurz (approved):
LGTM. Thanks @alex-matei !

alex-matei:
Forgot an import for Darwin, updated.

gkurz (approved):
> Forgot an import for Darwin, updated.

I had forgot about it too :smiley: . I confirm my approval.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,6 +306,36 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
+	for {
+		// Check if the process is running periodically to avoid a busy loop
+		// "A watched pot never boils" and an unwaited-for process never appears to die!
+		waitedPid, err := syscall.Wait4(pid, nil, syscall.WNOHANG, nil)
+
+		if waitedPid == pid && err == nil {
+			return false
+		}
+
+		if err := syscall.Kill(pid, syscall.Signal(0)); err != nil {
```
c3d, line 322: This may be related to the "watched pot never dies" comment above, but why do you need both `wait4` and `kill(sig, 0)` in the same loop?

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,6 +306,36 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
+			return false
+		}
+
+		select {
+		case <-time.After(50 * time.Millisecond):
```
c3d, line 327: In the end, what you are trying to do is to turn a `do { sleep(); } while (!dead)` into a `while (!dead) { sleep(); }`

That makes sense, but it's a win only if the process dies very quickly.

What about having a `time.After(delay)` where delay starts at, say, 1ms, and the next delay is max(delay*5, 50). So you'd have 1, 5, 25, 50 wait times, and then loop every 50ms. That way, you reap a quickly dying process very quickly (which I think was your intent) even if it takes, say, 5ms to die and no 0.25.

src/runtime/virtcontainers/utils/utils_linux.go
```diff
@@ -151,3 +153,59 @@ func IsAPVFIOMediatedDevice(sysfsdev string) bool {
...
+		return true, err
+	}
+
+	defer unix.Close(pidfd)
+
+	start := time.Now()
+	var n int
+
+	for {
+		remaining := int(timeoutSecs*1000) - int(time.Since(start).Milliseconds())
```
c3d, line 174: It's a bit disconcerting to mix `time.Duration` and times in milliseconds represented as `int`

What about something like:

```
maxDelay := time.Duration(timeoutSecs * time.Second)
end := time.Now().Add(maxDelay)
for {
   remaining := time.Until(end).Milliseconds()
[...]
   Poll(..., int(remaining))
```

I would find this more readable than multiplications by 1000 and similar.

src/runtime/virtcontainers/utils/utils_linux.go
```diff
@@ -151,3 +153,59 @@ func IsAPVFIOMediatedDevice(sysfsdev string) bool {
...
+			remaining = 0
+		}
+
+		n, err = unix.Poll([]unix.PollFd{{Fd: int32(pidfd), Events: unix.POLLIN}}, remaining)
+		if err != unix.EINTR {
+			break
+		}
+	}
+
+	if err != nil || n != 1 {
```
c3d, line 185: Did you observe the case `n != 1`? What is it for?

To me, the underlying syscall could get you -1, 0 or 1. The case where you'd get 0, you'd stay in the loop because of the `EINTR` test. The case where you'd get -1, presumably `err` would be set. So I think the test is defensive coding. Not necessarily a bad thing, but might warrant a comment.

c3d (commented):
Hi @alex-matei, thanks a lot for taking the time to iterate that and respond to @gkurz's points.

I've made a few remarks for things that either I did not understand, or minor changes to the code that I think could improve it a little. It does look like a good improvement, but as you'll see in one of my comments, I think we can make it even more efficient with a little tweak.

Let me know what you think.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -375,8 +375,19 @@ outer:
...
 		if err = syscall.Kill(pid, syscall.SIGKILL); err != nil {
+			if err == syscall.ESRCH {
+				logger.WithField("pid", pid).Warnf("kill encounters ESRCH, process already finished")
+				return nil
+			}
 			return fmt.Errorf("Failed to stop process %v: %s", pid, err)
 		}
+
+		for {
+			_, err := syscall.Wait4(pid, nil, 0, nil)
```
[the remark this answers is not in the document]
gkurz: The existing code base uses `syscall.Wait4`. Not sure `syscall.Waitpid` is available everywhere, e.g. Darwin.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,6 +306,36 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
...
 }
 
+func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	secs := time.Duration(timeoutSecs)
+	timeout := time.After(secs * time.Second)
+
+	for {
+		// Check if the process is running periodically to avoid a busy loop
+		// "A watched pot never boils" and an unwaited-for process never appears to die!
+		waitedPid, err := syscall.Wait4(pid, nil, syscall.WNOHANG, nil)
```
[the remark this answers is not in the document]
gkurz: `syscall.Wait4` pre-exists this patch. It is just code movement. And like with the other location, this is generic code that needs to run on platforms that might don't have `syscall.Waitpid`.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,6 +306,36 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
 	}
 }
 
+func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	secs := time.Duration(timeoutSecs)
+	timeout := time.After(secs * time.Second)
+
+	for {
+		// Check if the process is running periodically to avoid a busy loop
+		// "A watched pot never boils" and an unwaited-for process never appears to die!
```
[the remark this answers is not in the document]
gkurz: > I don't understand that comment at all, nor the logic in the loop (which appears really similar to what was there before). 

This is code refactoring. The comment and the logic mostly come from the existing code base.

> Is it saying that if you don't have a `wait4`, then `kill(pid, 0)` always appear successful?

Not exactly. It says that `kill(pid, 0)` will succeed until the process is reaped by its parent.

The confusion in this code comes from the fact it must support the hypervisor process being a direct child or not. The latter case, e.g. QEMU with the current code base, can only be achieved with `kill(0, pid)` when you don't have pidfds.
The direct child case cannot be handled by `kill(0, pid)` alone : something must reap the child first.

I don't like the comment either because if seems to indicate that `wait4` is there _just_ to ensure `kill(0)` will return ESRCH... while the important aspect of `wait4` is primarily to reap dead child resources.

Also this `kill(0)` construct is fragile as we have no guarantees that pid hasn't been recycled : things would be much more sane if hypervisor processses were always direct children because the pid is guaranteed to be valid until waited for.

src/runtime/virtcontainers/utils/utils.go
gkurz: This would be much better indeed but I'd like this improvement to sit in a separate commit (either preliminary or follow-up).

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -375,8 +375,19 @@ outer:
 	if pidRunning {
 		// Force process to die
 		if err = syscall.Kill(pid, syscall.SIGKILL); err != nil {
+			if err == syscall.ESRCH {
+				logger.WithField("pid", pid).Warnf("kill encounters ESRCH, process already finished")
```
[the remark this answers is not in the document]
alex-matei: Fixed.

src/runtime/virtcontainers/utils/utils.go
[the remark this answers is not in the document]
alex-matei: Waitpid/SYS_WAITPID is not defined for Darwin in syscall or golang.org/x/sys/unix. syscall.Wait4 is the function used internally in go runtime library.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -306,6 +306,36 @@ func ConvertAddressFamily(family int32) pbTypes.IPFamily {
 	}
 }
 
+func waitProcessUsingWaitLoop(pid int, timeoutSecs uint, logger *logrus.Entry) bool {
+	secs := time.Duration(timeoutSecs)
+	timeout := time.After(secs * time.Second)
```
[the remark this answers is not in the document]
alex-matei: Fixed.

src/runtime/virtcontainers/utils/utils.go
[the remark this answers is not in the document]
alex-matei: I think the original intent was for wait4 to handle the case for child processes and `kill(0, pid)` for the non-child processes. I removed the comment, it is confusing.

src/runtime/virtcontainers/utils/utils.go
alex-matei: `Wait4` call will succeed for a child process (and reap the child also). Using only `kill(0, pid)` will return nil as long the process isn't reaped, so without Wait4 the function will timeout. For non-child processes, using wait4 will fail. kill(0, pid) will fail with ESRCH error code once the process is reaped by `init` and the function will return false.

src/runtime/virtcontainers/utils/utils.go
alex-matei: Nice, added exponential backoff in a new commit.

src/runtime/virtcontainers/utils/utils_linux.go
alex-matei: Fixed.

alex-matei:
Thanks @c3d and @gkurz. I followed your suggestions and updated the code. Please take a look.

src/runtime/virtcontainers/utils/utils.go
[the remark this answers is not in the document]
gkurz: Yes. Ideally we'd want to rely on the go runtime library `Cmd.Wait()` or `Process.Wait()` to handle such platform details but this would require more changes that go beyond the goal of this PR.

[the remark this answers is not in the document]
c3d: OK. So let's keep it as is. I'm surprised you only have the non-POSIX syscall.

src/runtime/virtcontainers/utils/utils.go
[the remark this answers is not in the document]
c3d: @gkurz I like your explanation, it is clear and makes for a nice addition to the comment.
With that addition, we could keep "A watched pot never boils" comment, in case it's a "mind trigger" for the original authors. It just needs to be explained.

src/runtime/virtcontainers/utils/utils.go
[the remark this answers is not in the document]
c3d: Ack. Resolving this thread.

src/runtime/virtcontainers/utils/utils.go
c3d: @alex-matei If you have time to rephrase the comment to explain this logic. The wording of the comment above seems even more concise than @gkurz's, so it might be the content of the comment.

c3d (approved):
With the changes, looks good to me.

If you don't mind, it might be nice to use the text of one of your PR comments as the basis for a code comment replacing / adding to the "watched pot never boils" one. But this is not blocking and won't change my approval.

src/runtime/virtcontainers/utils/utils.go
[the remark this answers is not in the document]
gkurz: Nit: the comment gets removed by the next patch (runtime: use exponential backoff for process wait) instead of this patch. Also, I'd still leave a comment that explains why we're looping on both kill(0) (non child process) and wait4(WNOHANG) (child process).

alex-matei:
Updated, I added a comment to document the code behavior.

gkurz:
The only candidate for backport in this PR is https://github.com/kata-containers/kata-containers/commit/92ebe61fea0d1035914538e24add0cada0fe5c32 which fixes #5739.

Issue #5614, opened 2022-11-08
runtime-rs: delete all cargo patch in runtime-rs and Dragonball


studychao:
The cargo patch in the `cargo.toml` seems to cause the whole runtime-rs building time longer (#5449) and also makes it harder to build runtime-rs in an environment without the network (#5527 and #5526)

We should delete all patches from the `cargo.toml` file and publish all the crates that was once patched.

Issue #5527, opened 2022-10-27
Missing vendored code for dragonball in latest release, breaks build


etrunko:
# Description of problem

Version 3.0.0

Trying to build latest release with dragonball enabled for Fedora. Build fails with the following message:

```
# FIXME: This line will be removed when we solve the vm-memory dependency problem in Dragonball Sandbox
cargo update -p vm-memory:0.8.0 --precise 0.7.0
    Updating git repository `https://github.com/openanolis/dragonball-sandbox.git`
warning: spurious network error (2 tries remaining): failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12)
warning: spurious network error (1 tries remaining): failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12)
error: failed to load source for dependency `dbs-arch`
Caused by:
  Unable to update https://github.com/openanolis/dragonball-sandbox.git?rev=7a8e832b53d66994d6a16f0513d69f540583dcd0
Caused by:
  failed to clone into: /builddir/.cargo/git/db/dragonball-sandbox-d7838976048a7b3e
Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
Caused by:
  failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12)
make: *** [Makefile:9: build] Error 101
error: Bad exit status from /var/tmp/rpm-tmp.NM2WoF (%build)
```

We don't have access to the external internet while building on Fedora infrastructure. This is the reason why we include all vendor code in a separate tarball.

# Expected result

Build passes.

# Actual result

Build fails.

# Further information

https://koji.fedoraproject.org/koji/taskinfo?taskID=93491159

liubin:
@studychao PTAL

studychao:
I think this is also introduced by patch under Cargo.toml.

Sorry for the inconvenience and we are working on it.

Will fix this problem in the next release

Issue #5526, opened 2022-10-27
Missing dbs-upcall in vendored sources causes build to fail


etrunko:
# Description of problem

Verision 3.0.0.

Trying to build the latest release for Fedora, i got this error:
```
error: failed to resolve patches for `https://github.com/rust-lang/crates.io-index`

Caused by:
  patch for `dbs-upcall` in `https://github.com/rust-lang/crates.io-index` failed to resolve

Caused by:
  The patch location `https://github.com/openanolis/dragonball-sandbox.git?rev=7a8e832b53d66994d6a16f0513d69f540583dcd0` does not appear to contain any packages matching the name `dbs-upcall`.
make: *** [Makefile:366: target/x86_64-unknown-linux-musl/release/containerd-shim-kata-v2] Error 101
```

# Further information

https://koji.fedoraproject.org/koji/taskinfo?taskID=93486905

studychao:
Thanks, I'll try to handle this

studychao:
Hi @etrunko sorry for the inconvenience, this is a quick report about what happening and how you could solve this. I will take a deeper look into this tomorrow.

## Reason

It seems like `cargo vendor` only vendor the dependency needed as default and `dbs-upcall` is not marked as a default dependency so it's not collected into the vendor source when running `cargo vendor` (since we want to have upcall and acpi two ways to do hotplugging in Dragonball).

But the `patch` in Cargo.toml will try to find `dbs-upcall` under the vendor crates but it could not find `dbs-upcall` so this error pop up.

I will take more deeper look into the mechanism future.

## Simple solution
Since `dbs-upcall` is not default needed and related functions are still under development, you could add a patch to delete the `dbs-upcall = { git = "https://github.com/openanolis/dragonball-sandbox.git", rev = "7a8e832b53d66994d6a16f0513d69f540583dcd0" }` line from `/kata/src/runtime-rs/Cargo.toml` and `/kata/src/dragonball/Cargo.toml` and I've tried in my environment this could solve the problem.

## Follow-up actions
1. We will delete all the `patch` under Cargo.toml since the patch mechanism seems to cause lots of problems and publish all the crates into crate.io.
2. I'll take a deeper look at how `cargo vendor` and `patch` work and hopefully I could add this kind of check into the Kata CI.

Sorry for the inconvenience, and please do let me know the solution works or if there anything else I could help

etrunko:
Thansk for investigating @studychao.

[...]

Issue #5449, opened 2022-10-18
Speed up the runtime-rs build


wainersm:
**Which feature do you think can be improved?**

The build time of runtime-rs (shim-v2) with kata-deploy.

On kata-containers/kata-containers:
```
$ make shim-v2-tarball
make shim-v2-tarball-build
make[1]: Entering directory '/home/wmoschet/go/src/github.com/kata-containers/kata-containers'
/home/wmoschet/go/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-binaries-in-docker.sh  --build=shim-v2
sha256:a491f457f3dc0df209fb0e164188d6794f2b0c5baf60934df9ee3c39af5c93e7
Build kata version 3.1.0-alpha0: shim-v2
INFO: DESTDIR /home/wmoschet/go/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/shim-v2/destdir
Sending build context to Docker daemon  8.704kB
Step 1/9 : FROM ubuntu:20.04
 ---> 20fffa419e3a
Step 2/9 : ENV DEBIAN_FRONTEND=noninteractive
 ---> Using cache
 ---> e880554a2fed
Step 3/9 : RUN apt-get update &&     apt-get install -y --no-install-recommends         build-essential         ca-certificates         curl         gcc         git         make         musl-tools         sudo &&     apt-get clean && rm -rf /var/lib/apt/lists/
 ---> Using cache
 ---> a8070cdc4a5f
Step 4/9 : COPY install_go_rust.sh /usr/bin/install_go_rust.sh
 ---> Using cache
 ---> 80d8120f7e78
Step 5/9 : ARG GO_VERSION
 ---> Using cache
 ---> d6e93011df5e
Step 6/9 : ARG RUST_VERSION
 ---> Using cache
 ---> 186e27a02449
Step 7/9 : RUN install_go_rust.sh "${GO_VERSION}" "${RUST_VERSION}"
 ---> Using cache
 ---> ba08a0950c2c
Step 8/9 : ENV PATH=/usr/local/go/bin:${PATH}
 ---> Using cache
 ---> aad3b1fc8cf6
Step 9/9 : ENV PATH=/root/.cargo/bin/:${PATH}
 ---> Using cache
 ---> f752bacedcd3
Successfully built f752bacedcd3
Successfully tagged shim-v2-builder:latest
[...]

PR #5615, opened 2022-11-08
runtime-rs: delete all cargo patches

studychao:
The cargo patch in the cargo.toml seems to cause the whole runtime-rs building time longer and also makes it harder to build runtime-rs in an environment without the network

We should delete all patches from the cargo.toml file and publish all the crates that was once patched.

fixes: #5614 #5527 #5526 #5449

### DIFF
--- a/src/dragonball/Cargo.toml
+++ b/src/dragonball/Cargo.toml
@@ -12,15 +12,15 @@ edition = "2018"
 [dependencies]
 arc-swap = "1.5.0"
 bytes = "1.1.0"
-dbs-address-space = "0.1.0"
-dbs-allocator = "0.1.0"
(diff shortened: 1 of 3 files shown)


Tim-0731-Hzt (approved):

Tim-0731-Hzt (commented):
The version of dependency `dbs-utils` in `crates/hypervisor/Cargo.toml` should also be changed

studychao:
> The version of dependency `dbs-utils` in `crates/hypervisor/Cargo.toml` should also be changed

Thanks, changed.

Ji-Xinyou (approved):

quanweiZhou (approved):
LGTM

Issue #5017, opened 2022-08-28
runtime-rs: add kata-monitor support for runtime-rs


YushuoEdge:
### Introduction
`kata-monitor` is a component used to collect the running status of kata containers. In Kata 2.0, `kata-monitor` is implemented as a daemon to collect and expose metrics(including events) related to all kata workloads running on the same host. It can detect all kata runtimes(`containerd-shim-kata-v2`) and expose http endpoints for user to retrieval the data. All the provided metrics are in Prometheus format. The available metrics included:

- Kata runtime metrics
- Kata agent metrics
- Kata guest OS metrics
- Hypervisor metrics
- Firecracker metrics
- Kata monitor metrics

### Requirements
We hope to add `kata-monitor` support for `runtime-rs`. Considering that the architecture has changed in Kata 3.0, the original design and implementation may not be able to migrate smoothly. Therefore, we hope to first implement the metrics collection method of `kata-monitor`, and then continue to expand the types of metrics that can be collected. Specifically, we are going to complete at least one metric of different components. At the same time, these metrics should cover all the types. For example:

- Kata runtime: `kata_shim_pause_duration_seconds`[`SUMMARY`]
- Kata agent: `kata_agent_process_cpu_seconds`[`COUNTER`]
- Hypervisor(Dragonball): `kata_hypervisor_proc_status`[`GAUGE`]
- Kata guest OS:  `kata_guest_vm_stat`[`GAUGE`]
- Kata monitor: `kata_monitor_scrape_durations_histogram_milliseconds`[`HISTOGRAM`]

#### Some docs you can refer to ahead of time:

- kata monitor: https://github.com/kata-containers/kata-containers/tree/main/src/runtime/cmd/kata-monitor
- Kata 2.0 metrics design: https://github.com/kata-containers/kata-containers/blob/main/docs/design/kata-2-0-metrics.md#metric-types
- Kata 3.0 architecture: https://github.com/kata-containers/kata-containers/tree/main/docs/design/architecture_3.0

Since we need to add more metrics/events later, it is necessary to carefully consider the scalability of metrics collection method and how those metrics are managed. Also, we should consider minimizing the impact on the kata workloads.

PR #5643, opened 2022-11-10
runtime-rs: add kata-monitor support for runtime-rs

Yuan-Zhuo:
The previous kata-monitor in golang could not communicate with runtime-rs to gather metrics.
Here is kata-monitor in rust to gather metrics from runtime-rs and monitor itself.

Fixes: #5017

### DIFF
--- a/src/runtime-rs/Cargo.lock
+++ b/src/runtime-rs/Cargo.lock
@@ -854,11 +854,10 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "form_urlencoded"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fc25a87fa4fd2094bffb06925852034d90a17f0d1e05197d4956d3555752191"
+checksum = "a9c384f161156f5260c24a097c56119f9be8c798586aecc13afbcbe7b7e26bf8"
 dependencies = [
- "matches",
  "percent-encoding",
 ]
 
@@ -1092,6 +1091,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "h2"
+version = "0.3.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5f9f29bc9dda355256b2916cf526ab02ce0aeaaaf2bad60d65ef3f12f11dd0f4"
+dependencies = [
+ "bytes 1.1.0",
+ "fnv",
+ "futures-core",
+ "futures-sink",
+ "futures-util",
+ "http",
+ "indexmap",
+ "slab",
+ "tokio",
+ "tokio-util",
+ "tracing",
+]
+
+[[package]]
 name = "hashbrown"
 version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1171,6 +1189,7 @@ dependencies = [
  "futures-channel",
  "futures-core",
  "futures-util",
+ "h2",
  "http",
  "http-body",
  "httparse",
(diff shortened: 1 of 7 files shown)


studychao:
Thanks！ @YushuoEdge Could you please take a look at this?

YushuoEdge:
> Thanks！ @YushuoEdge Could you please take a look at this?

Sure, very glad to.

docs/design/README.md
```diff
@@ -11,6 +11,7 @@ Kata Containers design documents:
 - [Host cgroups](host-cgroups.md)
 - [`Inotify` support](inotify.md)
 - [Metrics(Kata 2.0)](kata-2-0-metrics.md)
+- [Metrics(Kata 3.0)](kata-3-0-metrics.md)
```
liubin, line 14: Golang runtime also provided in 3.0, so the difference is not the version, but the runtime, golang or rust. so the title can changed to golang runtime and rust runtime.

bergwolf:
The PR combines two big changes into one. Could you split them into two PRs so that they can be discussed/reviewed separately?

To be specific, I am having the following questions:
1. What is the benefit of introducing a new kata-monitor-rs? 
2. Does the runtime-rs metrics work with kata-monitor-go?
3. Does kata-monitor-rs work with the go runtime?

Tim-0731-Hzt:
> The PR combines two big changes into one. Could you split them into two PRs so that they can be discussed/reviewed separately?
> 
> To be specific, I am having the following questions:
> 
> 1. What is the benefit of introducing a new kata-monitor-rs? What's the relationship with kata-monitor-go?
> 2. Does the runtime-rs metrics work with kata-monitor-go?
> 3. Does kata-monitor-rs work with the go runtime?

As runtime-rs is going to be the default runtime in the future, Maybe is better to implement everything in Rust? Just  like implementing `kata-ctl`

YushuoEdge:
Yes, it would be better to separate this PR into two. One PR contains the part of `kata-monitor-rs` binary, and the other PR contains the part that supports metrics collection in `runtime-rs`.

Yuan-Zhuo:
> The PR combines two big changes into one. Could you split them into two PRs so that they can be discussed/reviewed separately?
> 
> To be specific, I am having the following questions:
> 
> 1. What is the benefit of introducing a new kata-monitor-rs? What's the relationship with kata-monitor-go?
> 2. Does the runtime-rs metrics work with kata-monitor-go?
> 3. Does kata-monitor-rs work with the go runtime?

Thanks for your review! 

Currently, `kata-monitor-go` can only get the metrics of `runtime-go`. In contrast, `kata-monitor-rs` can only get the metircs of `runtime-rs`.

This is mainly because `kata-monitor-go` and `kata-monitor-rs` respectively use the following interfaces(`shimclient`) provided by `runtime-go` and `runtime-rs` to send/receive http requests, making them tightly coupled with each other.

```
// kata-monitor-go
body, err := shimclient.DoGet(sandboxID, defaultTimeout, containerdshim.MetricsUrl)

// kata-monitor-rs
let shim_response = shim_client.get(METRICS_URI).await?;
```

And I have spilt the previous pr into this pr and https://github.com/kata-containers/kata-containers/pull/5797.

bergwolf:
@Yuan-Zhuo Because both runtime-rs and runtime-go provide http-based matrics endpoints, I don't think the client implementation is a blocker to unifying them. However, one blocker we might have is if runtime-rs intends to provide a different set of metrics than runtime-go. Do we have such expectations?

Yuan-Zhuo:
@bergwolf I hadn't given this much thought before, and I'm also starting to wonder whether `monitor-rs` is really needed. 
I think the second blocker you mentioned is fine, because runtime's http response is exactly the corresponding metrics string and no additional processing is required for monitor.
And another problem with getting `runtime-rs` metrics based on `monitor-go` is that `monitor-go` needs to identify whether the sandbox that the user requests to query metrics comes from `runtime-go` or `runtime-rs`.

Yuan-Zhuo:
`runtime-go`("/run/vc/sbs") and `runtime-rs`("/run/kata") use different sandbox's storage paths, so `monitor-go` cannot obtain the metrics in `runtime-rs` now.
To support this, should we make this path the same in both components, or determine which runtime version the requested container belongs to by checking which path exists in `monitor-go`?

Issue #5667, opened 2022-11-16
runtime: commonDialer does not exit when connection times out and succeeds


gaohuatao-1:
# Description of problem
The function commonDialer shown belows will not exit when timer expired and the dialFunc succeeds.
https://github.com/kata-containers/kata-containers/blob/6f9685fbf5a264f1babbbdb047e90baa53f725ec/src/runtime/virtcontainers/pkg/agent/protocols/client/client.go#L313

In our project, we call this function to establish a connection with kata-agent and export logs from VM to kata-shim on the host. The func commonDialer is called as the underlying function and  an unexpected error is triggered.

By analyzing the stack, the program blocks at line 383 When the above scenario is triggered.
![image](https://user-images.githubusercontent.com/22569211/202131255-9601002f-352b-4487-9cc0-a54b2a6559fe.png)

# Expected result
If the set timeout expires, the function commonDialer should exit.

# Actual result

When the timeout is over, the func commondialer can exit.

PR #5668, opened 2022-11-16
runtime: fix commonDialer cannot exit

gaohuatao-1:
Fixes: #5667

Signed-off-by: gaohuatao <gaohuatao@bytedance.com>

### DIFF
--- a/src/runtime/virtcontainers/pkg/agent/protocols/client/client.go
+++ b/src/runtime/virtcontainers/pkg/agent/protocols/client/client.go
@@ -312,7 +312,8 @@ func parseGrpcHybridVSockAddr(sock string) (string, uint32, error) {
 // https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md
 func commonDialer(timeout time.Duration, dialFunc func() (net.Conn, error), timeoutErrMsg error) (net.Conn, error) {
 	t := time.NewTimer(timeout)
-	cancel := make(chan bool)
+	// Create a buffered channel so that notifying subroutine is not blocked
+	cancel := make(chan bool, 1)
 	ch := make(chan net.Conn)
 	go func() {
 		for {


src/runtime/virtcontainers/pkg/agent/protocols/client/client.go
```diff
@@ -345,7 +345,8 @@ func commonDialer(timeout time.Duration, dialFunc func() (net.Conn, error), time
 			return nil, timeoutErrMsg
 		}
 	case <-t.C:
-		cancel <- true
+		// Do not block, just notify the sub-goroutin to exit
+		close(cancel)
```
bergwolf, line 349: I'd prefer fixing it by adding a buffer to the cancel channel. Something like
```
cancel := make(chan bool, 1)
```

gaohuatao-1: thanks, fixed!

src/runtime/virtcontainers/pkg/agent/protocols/client/client.go
```diff
@@ -312,7 +312,8 @@ func parseGrpcHybridVSockAddr(sock string) (string, uint32, error) {
 // https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md
 func commonDialer(timeout time.Duration, dialFunc func() (net.Conn, error), timeoutErrMsg error) (net.Conn, error) {
 	t := time.NewTimer(timeout)
-	cancel := make(chan bool)
+	// Do not block, just notify the sub-goroutin to exit
```
amshinde, line 315: lgtm. Maybe you can change this comment to "Create a buffered channel so that notifying subroutine is not blocked"

amshinde (approved):

amshinde:
@bergwolf Can you have another look at this?

bergwolf (approved):
Thanks!

gaohuatao-1:
@bergwolf PTAL. Thanks

alex-matei:
I've closed #8151, it is the same problem. I'm still investigating why firecracker connection times out so I'll let the associated issue #8150 open.

Issue #5674, opened 2022-11-17
Race condition between wait thread and monitor thread


fengwang666:
I encounter a weird edge case. The kata agent just died, then the shim/qemu processes are stuck forever, occupying the cpu/memory resource. There is a sandbox watcher that is suppose to watch both the vmm and the agent liveness and force delete the sandbox if the agent is dead. But in this case, I suspect the reason why watcher didn’t force delete the sandbox is because of a race between [the monitor thread killing sandbox](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L118-L133) and [wait thread cancelling monitor thread](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/wait.go#L74-L77). Cancelling the watcher in the wait thread might need to be revisited.

PR #5675, opened 2022-11-17
runtime: force kill sandbox if agent is dead

fengwang666:
If the agent is dead while the sandbox is exiting, there can be a race condition between the wait thread and monitor thread that the wait thread cancels the monitor thread before it handles the agent liveness error, causing the sandbox stuck in termination.

Fixes: #5674


src/runtime/virtcontainers/agent.go
```diff
@@ -174,6 +174,9 @@ type agent interface {
 	// markDead tell agent that the guest is dead
 	markDead(ctx context.Context)
 
+	//isDead returns whether the guest is dead
```
liubin, line 177: Nit: add a space after `//`

amshinde (approved):

fengwang666:
My previous reasoning is not correct. close this PR.

Issue #5652, opened 2022-11-11
CC | agent: Add support for enabling authenticated registry support when pulling images


stevenhorsman:
In order to enable pulling the authentication info from the secure channel, we need to set `ImageConfig.auth` to be true (and it defaults to false) in the image client.

I think the best logic to follow (and Ding agrees in https://github.com/confidential-containers/image-rs/pull/72#issuecomment-1299528725) is to set ImageConfig.auth=true in the kata-agent's image_rpc.rs, if we are using an attestation-agent KBC (ie aa_kbc_params is set). This should mean that the secure channel is always available to attempt to retrieve the credentials from.

PR #5680, opened 2022-11-17
CC | agent: Add auth reg support to the agent

stevenhorsman:
agent: Add auth reg support
    
If the attestation-agent is used then enable image_client_auth
to enable the attempt to get registry credentials for the pull

Fixes: #5652

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -274,7 +274,7 @@ dependencies = [
  "serde_json",
  "sha2 0.10.5",
  "shadow-rs 0.16.3",
- "strum 0.24.1",
+ "strum",
  "tonic-build 0.8.0",
 ]
 
@@ -322,7 +322,7 @@ dependencies = [
  "matchit",
  "memchr",
  "mime",
- "percent-encoding 2.1.0",
+ "percent-encoding",
  "pin-project-lite",
  "serde",
  "sync_wrapper",
@@ -1247,7 +1247,7 @@ dependencies = [
  "der 0.3.5",
  "elliptic-curve",
  "hmac 0.11.0",
- "signature 1.3.2",
+ "signature",
 ]
 
 [[package]]
@@ -1256,7 +1256,7 @@ version = "1.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e9c280362032ea4203659fc489832d0204ef09f247a0506f170dafcac08c369"
 dependencies = [
- "signature 1.3.2",
+ "signature",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


stevenhorsman:
Now the upstream change in image-rs has merged I've updated this to point to the correct branch.

fitzthum (approved):
Cool. So there are no additional config flags for this? We just support it as long as we have he AA. That seems fine given that authenticated registry support doesn't block pulling standard container images.

stevenhorsman:
> Cool. So there are no additional config flags for this? We just support it as long as we have he AA. That seems fine given that authenticated registry support doesn't block pulling standard container images.

Yeah - I was keen to avoid yet another config flag and based on the assumption that the only negative that happens if we turn the setting on in image-rs is that we might make an extra request to the AA (confirmation of this should be in the integration test's existing cases), then I/we figured that the AA check would be the simplest solution that should always pick up the credentials if available.

fitzthum:
Yeah it's kind of interesting that we don't have a designated CC flag in the guest, although it's not clear that this would actually simplify things. One question is if we would have any issue with AA's the don't support the get_resource endpoint. I'm not sure how `image-rs` is implemented here, but is this going to fail gracefully if get_resource is unimplemented or are we not going to be able to unpack any containers?

stevenhorsman:
> One question is if we would have any issue with AA's the don't support the get_resource endpoint. I'm not sure how `image-rs` is implemented here, but is this going to fail gracefully if get_resource is unimplemented or are we not going to be able to unpack any containers? Also what happens if we use fs_kbc but there is no credentials file on the filesystem?

I hope these won't be a problem - I think these were covered when I chatted to Ding (@xynnn007) during review of the image-rs and we think they should be handled okay, but might have missed something. Our integration tests will cover the second scenario of when we don't have a credentials file, but we are just using the offline_fs_kbc which does work with the get_resource endpoint.

fitzthum:
We will be switching away from the `offline_sev_kbc` shortly, which I think is the only KBC that doesn't have get resource.

Xynnn007:
> One question is if we would have any issue with AA's the don't support the get_resource endpoint. I'm not sure how `image-rs` is implemented here, but is this going to fail gracefully if get_resource is unimplemented or are we not going to be able to unpack any containers? Also what happens if we use fs_kbc but there is no credentials file on the filesystem?

Hi Tobin, now if `get_resource` is not supported or fails there will be a warning and will use `Anonymous` auth instead to try to access the registry.

fidencio:
Let's merge this one now.  From the TDX side we're hitting an issue with flannel which was just proposed to be backported by @BbolroC.

fidencio (approved):
lgtm, thanks @stevenhorsman!

fitzthum:
btw @stevenhorsman can you add something to the release notes about this?

stevenhorsman:
> btw @stevenhorsman can you add something to the release notes about this?

Yes - good point. I'm not sure if we have a separate section for details apart from the quickstart, but I'll prepare something and we can knock it into shape as we get close to the release!

stevenhorsman:
https://github.com/confidential-containers/documentation/pull/83 created for the release notes starter :)

fitzthum:
Yeah would be good to add something to the quickstart guide at some point as well.

Issue #5694, opened 2022-11-18
vhost user fs queue size is missing under qemu


darfux:
**Which feature do you think can be improved?**

queue size of vhost user fs when using qemu

**How can it be improved?**

The default value of qemu  is 128 (https://github.com/qemu/qemu/blob/master/hw/virtio/vhost-user-fs.c#L292) , set it to 1024 by default and make it configurable

PR #5695, opened 2022-11-18
runtime: Support virtiofs queue size for qemu and make it configurable

darfux:
The default vhost-user-fs queue-size of qemu is 128 now. Set it to 1024 by default which is same as clh. Also make this value configurable.

Fixes: #5694

Signed-off-by: liyuxuan.darfux <liyuxuan.darfux@bytedance.com>


src/runtime/Makefile
```diff
@@ -193,6 +193,7 @@ DEFVALIDVIRTIOFSDAEMONPATHS := [\"$(DEFVIRTIOFSDAEMON)\"]
 #if value is 0, DAX is not enabled
 DEFVIRTIOFSCACHESIZE ?= 0
 DEFVIRTIOFSCACHE ?= auto
+DEFVIRTIOFSQUEUESIZE ?= 1024
```
liubin, line 196: Is 1024 an appropriate value?

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -1503,6 +1504,11 @@ func (vhostuserDev VhostUserDevice) QemuFSParams(config *Config) []string {
 	deviceParams = append(deviceParams, driver)
 	deviceParams = append(deviceParams, fmt.Sprintf("chardev=%s", vhostuserDev.CharDevID))
 	deviceParams = append(deviceParams, fmt.Sprintf("tag=%s", vhostuserDev.Tag))
+	queueSize := uint32(1024)
+	if vhostuserDev.QueueSize != 0 {
```
liubin, line 1508: Does this keep the minimum change(Not sure setting a default 1024 does matter )?

```golang
if vhostuserDev.QueueSize != 0 {
	deviceParams = append(deviceParams, fmt.Sprintf("queue-size=%d", vhostuserDev.QueueSize))
}
```

src/runtime/Makefile
darfux: CLH uses 1024 now： https://github.com/kata-containers/kata-containers/blob/3bb145c63ad1fe8c335d28029ee8ad8ca5f4f340/src/runtime/virtcontainers/clh.go#L1589

src/runtime/pkg/govmm/qemu/qemu.go
darfux: As comment above, here uses 1024 to keep same with CLH. Qemu's default 128 is a bit small as FUSE uses a 256 pages limit:
https://github.com/torvalds/linux/blob/v6.0/fs/fuse/fuse_i.h#L39

liubin (approved):

src/runtime/Makefile
fidencio: 1024 is the default value, AFAIR.

fidencio (approved):
lgtm, thanks @darfux!

fidencio:
@BbolroC, s390x main is failing, mind to check whether this could be related to this PR?

BbolroC:
> @BbolroC, s390x main is failing, mind to check whether this could be related to this PR?

@fidencio ah, sorry for the late check. The error has nothing to do with this PR. It is a flaky one due to the cni configuration. 

I will fix this after the current task is finished. Thanks for the notification.

PR #5703, opened 2022-11-20
docs: update doc "NVIDIA-GPU-passthrough-and-Kata"

kinderyj:
According to the NVIDIA-GPU-passthrough-and-Kata doc, user always forget to make the hook shell executable, as a result, the hook shell can't be executed successfully, and then the kata container with GPU passthrough will not start successfully. 

We should make sure the hook shell nvidia-container-toolkit is executable.

Fixes: https://github.com/kata-containers/kata-containers/issues/5594

Signed-off-by: Matt Wang <kinder_yj@hotmail.com>


kinderyj:
@Ji-Xinyou @zvonkok

Issue #5706, opened 2022-11-21
refactor: add conditional compile for virt-sandbox persist


Tim-0731-Hzt:
add conditional compile for virt-sandbox persist

PR #5707, opened 2022-11-21
Refactor(runtime-rs): add conditional compile for virt-sandbox persist

Tim-0731-Hzt:
code refactoring, add conditional compile for virt-sandbox persist

Fixes: #5706

### DIFF
--- a/src/runtime-rs/crates/runtimes/src/manager.rs
+++ b/src/runtime-rs/crates/runtimes/src/manager.rs
@@ -20,11 +20,12 @@ use kata_types::{annotations::Annotation, config::TomlConfig};
 use linux_container::LinuxContainer;
 use persist::sandbox_persist::Persist;
 use tokio::sync::{mpsc::Sender, RwLock};
-use virt_container::sandbox::SandboxRestoreArgs;
-use virt_container::sandbox::VirtSandbox;
-use virt_container::sandbox_persist::{SandboxState, SandboxTYPE};
 #[cfg(feature = "virt")]
-use virt_container::VirtContainer;
+use virt_container::{
+    sandbox::{SandboxRestoreArgs, VirtSandbox},
+    sandbox_persist::SandboxState,
+    VirtContainer,
+};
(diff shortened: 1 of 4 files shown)


studychao (commented):
I'm good with the commit itself.
But I'm curious about why changing to conditional compilation when deciding sandbox type instead of just using `sandbox_state.sandbox_type`

Tim-0731-Hzt:
> I'm good with the commit itself. But I'm curious about why changing to conditional compilation when deciding sandbox type instead of just using `sandbox_state.sandbox_type`

For other sandbox types like Linux and wasm, the function will not be called. So, I think it's better to conditionally compile it

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -153,26 +154,23 @@ impl RuntimeHandlerManager {
...
+        #[cfg(feature = "virt")]
+        {
+            let sandbox = VirtSandbox::restore(sandbox_args, sandbox_state)
+                .await
+                .context("failed to restore the sandbox")?;
+            sandbox
+                .cleanup(&inner.id)
+                .await
+                .context("failed to cleanup the resource")?;
+        }
```
bergwolf, line 173: What if we enable all `linux`, `wasm` and `virt` features, and let upper layer decide which type of sandbox to create?

studychao: +1

Ji-Xinyou:
It would be more clear to change commit title to `refactor(runtime-rs): add conditional compile for virt-sandbox persist`

src/runtime-rs/crates/runtimes/src/manager.rs
Tim-0731-Hzt: fixed

quanweiZhou (approved):
LGTM

studychao:
@Tim-0731-Hzt There is one ubuntu CI not passing, could you take a look?

Issue #5642, opened 2022-11-10
Introducing upcall for built-in sandbox Dragonball


studychao:
Upcall mechanism was introduced in the AC meeting and now we will start to ship the code to the community which will include the upcall itself and Kata CI integration for upcall since it needs kernel patches to enable upcall.

## What is upcall?
Upcall is a direct communication tool between VMM and guest developed upon vsock. The server side of the upcall is a driver in guest kernel (kernel patches are needed for this feature) and it'll start to serve the requests after the kernel starts. And the client side is in Dragonball VMM , it'll be a thread that communicates with vsock through uds.

We want to keep the lightweight of the VM through the implementation of the upcall.

![image](https://user-images.githubusercontent.com/31121654/201098316-87e0a39f-4344-4b2d-9d52-b2df29ddd2a1.png)

## What can upcall do?

We define specific operations in the device manager service (one of the services in upcall we developed) to perform device hotplug / hot-unplug including vCPU hotplug, virtio-mmio hotplug, and memory hotplug. We have accomplished device hotplug / hot-unplug directly through upcall in order to avoid the virtualization of ACPI to minimize virtual machines' overhead. And there could be many other uses if other services are implemented.

## How to enable upcall?

Upcall needs a server in the guest kernel which will be several kernel patches for the upcall server itself and different services registered in the upcall server. It's currently tested on upstream Linux kernel 5.10.
Also, a client-side is also needed in VMM. Dragonball will open-source how to implement the client side for upcall.

PR #5712, opened 2022-11-21
Dragonball: introduce upcall

studychao:
Upcall is a direct communication tool between VMM and guest developed upon vsock. The server side of the upcall is a driver in the guest kernel (kernel patches are needed for this feature) and will start to serve the requests after the kernel starts. And the client side is in Dragonball VMM , it'll be a thread that communicates with vsock through uds.

We want to keep the lightweight of the VM through the implementation of the upcall, through which we could achieve vCPU hotplug, virtio-mmio hotplug without implementing complex and heavy virtualization features such as ACPI virtualization.

fixes: #5642

---

This PR is still a WIP since we need to integrate those kernel patches with Kata guest os generation tool and Kata CI. Those work will finish in following days.

- [x] Only apply those as an experimental dragonball patch
- [x] Create a new runtime class for using this patched kernel
may not need a new runtime class and reason explained below.
- [x] Add the "experimental" kernel to what we distribute with Kata Deploy.
issue #5859 tracks this.


fidencio:
@studychao, please, when working on the integration of those, I'd suggest to:
1. Only apply those as an experimental dragonball patch
2. Create a new runtime class for using this patched kernel
3. Add the "experimental" kernel to what we distribute with Kata Deploy.

studychao:
> @studychao, please, when working on the integration of those, I'd suggest to:
> 
> 1. Only apply those as an experimental dragonball patch
> 2. Create a new runtime class for using this patched kernel
> 3. Add the "experimental" kernel to what we distribute with Kata Deploy.

Thanks @fidencio ! I'll take those advice in mind while doing integration work:)

studychao:
Hi @fidencio , I have finished the most part of the integration work and please help to review the PR.
For your three suggestions, 

> 1.Only apply those as an experimental dragonball patch

I have added upcall patches support into build_kernel.sh and only when `-e` and `-t dragonball` are both enabled which is included in this PR

> 2. Create a new runtime class for using this patched kernel

This one I am not quite get it, where should I add a new runtime class for upcall guest kernel?

> 3.Add the "experimental" kernel to what we distribute with Kata Deploy.

I'll do this in a separate PR for Kata Deploy

Thanks for your review:)

studychao:
PTAL :) @ZizhengBian @liubin @bergwolf @jiangliu @egernst

studychao:
As discussed with @fidencio , he provides an example for runtimeclass (https://github.com/kata-containers/kata-containers/pull/5785/files). Thanks!

However, I think upcall may not be a suitable choice for another runtimeclass.
The reason is that in the previous PR, no matter QEMU with or without Nydus, the Kata Container itself is complete in function (as far as I know, Nydus is a performance bouns) . But upcall provides device hotplug for Dragonball which is a fundamental function so we think it's not a good choice to provide two version of Kata Containers + Dragonball with or without upcall which will cause the Dragonball without upcall one loses the device hotplug function.
So we would like to provide only one runtimeclass with upcall ;)

studychao:
As for the " Add the "experimental" kernel to what we distribute with Kata Deploy." requirement, we need to first add upcall into the static tarball Kata delivered in order to use Kata Deploy. But right now I think this PR focus on introducing upcall to the Kata , and later PRs () could use upcall to enable cpu hotplug, block device hotplug and etc. So we would like to add upcall into Kata and test it stable in CI for doing device hotplug jobs, and then we will add upcall guest kernel to the tarball.

I will create an issue for add upcall guest kernel into static tarball to track this .

tools/packaging/kernel/build-kernel.sh
```diff
@@ -538,6 +538,16 @@ main() {
 
 	[ -z "${subcmd}" ] && usage 1
 
+	if [[ ${build_type} == "experimental" ]] && [[ ${hypervisor_target} == "dragonball" ]]; then
+		build_type="dragonball-experimental"
+		if [ -n "$kernel_version" ];  then
+			kernel_ver=$(get_major_kernel_version "${version}")
```
fidencio, line 544:
```suggestion
			kernel_ver=$(get_major_kernel_version "${kernel_version}")
```

fidencio: Then here you declare `kernel_ver` but does not actually use it, which makes me thing that ....

tools/packaging/kernel/build-kernel.sh
```diff
@@ -538,6 +538,16 @@ main() {
...
+			if [[ ${kernel_version} != "5.10" ]]; then
```
fidencio, line 545: Here you actually would like to check
```suggestion
			if [ "${kernel_ver}" != "5.10" ]; then
```

fidencio (commented):
@studychao, I've left some comments, please, take a look at them.

tools/packaging/kernel/build-kernel.sh
studychao: good catch!!! I've changed `kernel_ver` to `kernel_major_version` and also `if [[ ${kernel_version} != "5.10" ]]; then` to `if [[ ${kernel_major_version} != "5.10" ]]; then`

tools/packaging/kernel/build-kernel.sh
studychao: good catch!!! I've changed kernel_ver to kernel_major_version and also if [[ ${kernel_version} != "5.10" ]]; then to if [[ ${kernel_major_version} != "5.10" ]]; then

studychao:
Hi @fidencio , I have modified the PR accordin to your advice. 
Please take a look at them :)

fidencio (approved):
lgtm, thanks @studychao!

tools/packaging/kernel/patches/5.10.x/dragonball-experimental/0002-upcall-introduce-device-manager-upcall-service.patch
```diff
@@ -0,0 +1,326 @@
...
+After a request is sent to device manager service, db_devmgr_handler
+will start handle the request. A kthread `db_devmgr_server` will be
+created and it will send CONNECT message to the client side to notify
+the client start sending message for device management operations.
+`db_devmgr_process` will be used for determining which device operations
+will be triggered through msg_type. `get_action` will find out the
+action for dealing with the operation and `action` fn will execute the
+actual device management operation in the device manager service.
+
+Signed-off-by: Chao Wu <chaowu@linux.alibaba.com>
```
ZizhengBian, line 20: The sob here is incomplete, right?

studychao: Thanks, it's added.

ZizhengBian:
LGTM

ZizhengBian (approved):

liubin (approved):

bergwolf (approved):

PR #5714, opened 2022-11-21
runtime: don't fail mkdir if the folder is already created by another process

### LINKED ISSUE #5713: container creation fails when trying to mkdir /run/vc/vm
# Description of problem

When trying to create multiple containers at the same time some of them might fail because multiple processes try to create the folder /run/vc/vm.

alex-matei:
In case a folder is created by another process continue without generating an error.

Fixes #5713

### DIFF
--- a/src/runtime/virtcontainers/utils/utils.go
+++ b/src/runtime/virtcontainers/utils/utils.go
@@ -399,7 +399,7 @@ func MkdirAllWithInheritedOwner(path string, perm os.FileMode) error {
 		info, err := os.Stat(curPath)
 
 		if err != nil {
-			if err = os.Mkdir(curPath, perm); err != nil {
+			if err = os.MkdirAll(curPath, perm); err != nil {
 				return fmt.Errorf("mkdir call failed: %v", err.Error())
 			}
 			if err = syscall.Chown(curPath, uid, gid); err != nil {


src/runtime/virtcontainers/utils/utils.go
```diff
@@ -397,26 +397,38 @@ func MkdirAllWithInheritedOwner(path string, perm os.FileMode) error {
...
+
+		err = os.Mkdir(curPath, perm)
+		if err == nil {
 			if err = syscall.Chown(curPath, uid, gid); err != nil {
 				return fmt.Errorf("chown syscall failed: %v", err.Error())
 			}
 			continue
 		}
 
+		if !os.IsExist(err) {
```
amshinde, line 421: You dont need to check for if the directory exists here again. Just check for in the else clause of the previous if statement.

alex-matei: Thanks @amshinde, updated.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -399,12 +399,14 @@ func MkdirAllWithInheritedOwner(path string, perm os.FileMode) error {
 		info, err := os.Stat(curPath)
 
 		if err != nil {
-			if err = os.Mkdir(curPath, perm); err != nil {
+			err = os.Mkdir(curPath, perm)
```
bergwolf, line 402: What about just calling `os.MkdirAll` and do not care about EEXIST errors?

alex-matei: Good idea, changed the code to use MkdirAll.

gkurz (approved):
LGTM. Thanks @alex-matei !

fidencio (approved):
lgtm, thanks @alex-matei!

bergwolf (approved):
lgtm!

Issue #5726, opened 2022-11-23
Add nydusd to static release tarball


liubin:
Add nydusd to `/opt/kata/libexec/` alongside virtiofsd.

PR #5727, opened 2022-11-23
package: add nydus to release artifacts

liubin:
Install nydus related binaries under /opt/kata/libexec/

Fixes: #5726

### DIFF
--- a/.github/workflows/kata-deploy-push.yaml
+++ b/.github/workflows/kata-deploy-push.yaml
@@ -25,6 +25,7 @@ jobs:
           - rootfs-image
           - rootfs-initrd
           - virtiofsd
+          - nydus
     steps:
       - uses: actions/checkout@v2
       - name: Install docker
--- a/.github/workflows/kata-deploy-test.yaml
+++ b/.github/workflows/kata-deploy-test.yaml
@@ -50,6 +50,7 @@ jobs:
           - cloud-hypervisor
           - firecracker
           - kernel
+          - nydus
           - qemu
           - rootfs-image
           - rootfs-initrd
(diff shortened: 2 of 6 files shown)


bergwolf (approved):
lgtm! Thanks!

fidencio (approved):
lgtm, thanks @liubin!

fidencio:
> @liubin, @bergwolf, do you think we should bother about having a specific config file for nydus instead of virtiofsd?

I'm asking as I think it'd be beneficial for users to simply use a different runtimeClass, such as `kata-nydus`, and be able to simply test it out.

liubin:
@fidencio The nydusd is only for file sharing, we have already supported it by configuring the `configuraiton.toml`:

https://github.com/kata-containers/kata-containers/blob/5cbf879659532cad3df084eb3c89a584ac19d2ba/src/runtime/config/configuration-qemu.toml.in#L176-L180


The `runtimeClass` is more related to snapshotter, that is what [`nydus-snapshotter`](https://github.com/containerd/nydus-snapshotter) is doing, so IMO there no more changes are needed.

In the new containerd (from 1.7), it [supports specifying snapshotter for runtime](https://github.com/containerd/containerd/pull/6899), so we can use different snapshotters for different runtimes.

fidencio:
> @fidencio The nydusd is only for file sharing, we have already supported it by configuring the `configuraiton.toml`:

Let me rephrase what I wanted to say.

I think we should have a configuration using nydus by default, something like configuration-qemu-nydus.toml, and then add a runtimeClass to use such configuration.  So we don't have users having to connect to the node and changing configuration in order to be able to test nydus.

Issue #123, opened 2020-01-16
actions: release: fail to upload kata tarball to github


jcvenegas:
# Description of problem

```
Run tag=$(echo $GITHUB_REF | cut -d/ -f3-)
Cloning into 'runtime'...
Attaching 1 asset...
Attaching these assets failed:
../release-candidate/kata-static-1.10.0-x86_64.tar.xz

Error uploading release asset: Not Found (HTTP 404)
Not Found
##[error]Process completed with exit code 1.
```
# Expected result

action is able to upload the tarball

jcvenegas:
This was an issue we got in release, it seems that was a github api error, not seen anymore.

PR #5735, opened 2022-11-23
kata-ctl: Add --list option

### LINKED ISSUE #5344: kata-ctl: check: Add "--list" option
Implement the `kata-runtime check --list` command with unit tests.

This is a new command that lists all available checks. Fields to display:

- name of check.
- architecture the check runs on (or `all`).
- whether the check requires elevated privileges.
- whether the check would run for a particular class / "privilege level" of user (`root` or non-`root`).

**Question:** should this be `check --list` or `check list` (or both)?

dborquez:
Implement the kata-runtime check --list command with unit tests

Fixes #5344

### DIFF
--- a/src/tools/kata-ctl/src/arch/mod.rs
+++ b/src/tools/kata-ctl/src/arch/mod.rs
@@ -3,8 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use anyhow::Result;
-
 #[cfg(target_arch = "aarch64")]
 pub mod aarch64;
 #[cfg(target_arch = "aarch64")]
@@ -32,7 +30,3 @@ pub use x86_64 as arch_specific;
     target_arch = "x86_64"
 )))]
 compile_error!("unknown architecture");
-
-pub fn check() -> Result<()> {
-    arch_specific::check()
-}
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -8,6 +8,7 @@ pub use arch_specific::*;
 
 mod arch_specific {
     use crate::check;
+    use crate::types::*;
     use anyhow::{anyhow, Result};
 
     const PROC_CPUINFO: &str = "/proc/cpuinfo";
@@ -16,8 +17,19 @@ mod arch_specific {
     const CPU_FLAGS_INTEL: &[&str] = &["lm", "sse4_1", "vmx"];
     const CPU_ATTRIBS_INTEL: &[&str] = &["GenuineIntel"];
 
-    // check cpu
-    fn check_cpu() -> Result<()> {
+    // List of check functions
+    static CHECK_LIST: &[CheckItem] = &[CheckItem {
+        name: CheckType::CheckCpu,
+        descr: "This parameter performs the cpu check",
+        fp: check_cpu,
+        perm: PermissionType::NonPrivileged,
+    }];
+
+    pub fn get_checks() -> &'static [CheckItem<'static>] {
+        CHECK_LIST
+    }
+
+    fn check_cpu(_args: &str) -> Result<()> {
         println!("INFO: check CPU: x86_64");
 
         let cpu_info = check::get_single_cpu_info(PROC_CPUINFO, CPUINFO_DELIMITER)?;
@@ -39,16 +51,6 @@ mod arch_specific {
         if !missing_cpu_flags.is_empty() {
             eprintln!("WARNING: Missing CPU flags {:?}", missing_cpu_flags);
         }
-
-        Ok(())
-    }
-
-    pub fn check() -> Result<()> {
-        println!("INFO: check: x86_64");
-
-        let _cpu_result = check_cpu();
-
-        // TODO: collect outcome of tests to determine if checks pass or not
 
         Ok(())
     }
--- a/src/tools/kata-ctl/src/args.rs
+++ b/src/tools/kata-ctl/src/args.rs
@@ -64,6 +64,9 @@ pub enum CheckSubCommand {
 
     /// List all official and pre-release packages
     IncludeAllReleases,
+
+    /// List all available checks
+    List,
 }
 
 #[derive(Debug, Args)]
--- a/src/tools/kata-ctl/src/ops/check_ops.rs
+++ b/src/tools/kata-ctl/src/ops/check_ops.rs
@@ -3,15 +3,69 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use crate::arch;
-use crate::check;
-use crate::ops::version;
+use crate::arch::x86_64::get_checks;
 
 use crate::args::{CheckArgument, CheckSubCommand, IptablesCommand, MetricsCommand};
 
-use anyhow::Result;
+use crate::check;
+
+use crate::ops::version;
+
+use crate::types::*;
+
+use anyhow::{anyhow, Result};
 
 const NAME: &str = "kata-ctl";
+
+// This function retrieves the cmd function passes as argument
+fn get_builtin_check_func(name: CheckType) -> Result<BuiltinCmdFp> {
+    let check_list = get_checks();
+
+    for check in check_list {
+        if check.name.eq(&name) {
+            return Ok(check.fp);
+        }
+    }
+
+    Err(anyhow!("Invalid command: {:?}", name))
+}
+
+// This function is called from each 'kata-ctl check' argument section
+fn handle_builtin_check(check: CheckType, args: &str) -> Result<()> {
+    let f = match get_builtin_check_func(check) {
+        Ok(fp) => fp,
+        Err(e) => return Err(e),
+    };
+
+    f(args)
+}
+
+fn get_client_cmd_details() -> Vec<String> {
+    let mut cmds = Vec::new();
+    let check_list = get_checks();
+
+    for cmd in check_list {
+        cmds.push(format!("{} ({}. Mode: {})", cmd.name, cmd.descr, cmd.perm));
+    }
+
+    cmds
+}
+
+fn print_check_list() -> Result<()> {
+    let cmds = get_client_cmd_details();
+
+    if cmds.is_empty() {
+        println!("Checks not found!\n");
+
+        return Ok(());
+    }
+
+    cmds.iter().for_each(|n| println!(" - {}", n));
+
+    println!();
+
+    Ok(())
+}
 
 pub fn handle_check(checkcmd: CheckArgument) -> Result<()> {
     let command = checkcmd.command;
@@ -19,7 +73,7 @@ pub fn handle_check(checkcmd: CheckArgument) -> Result<()> {
     match command {
         CheckSubCommand::All => {
             // run architecture-specific tests
-            arch::check()?;
+            handle_builtin_check(CheckType::CheckCpu, "")?;
 
             // run code that uses network checks
             check::run_network_checks()?;
@@ -27,12 +81,15 @@ pub fn handle_check(checkcmd: CheckArgument) -> Result<()> {
 
         CheckSubCommand::NoNetworkChecks => {
             // run architecture-specific tests
-            arch::check()?;
+            handle_builtin_check(CheckType::CheckCpu, "")?;
         }
 
         CheckSubCommand::CheckVersionOnly => {
-            // retrieve latest release
-            check::check_version()?;
+            handle_version()?;
+        }
+
+        CheckSubCommand::List => {
+            print_check_list()?;
         }
         CheckSubCommand::OnlyListReleases => {
             // retrieve official release
@@ -83,6 +140,7 @@ pub fn handle_metrics(_args: MetricsCommand) -> Result<()> {
 
 pub fn handle_version() -> Result<()> {
     let version = version::get().unwrap();
+
     println!("{} version {:?} (type: rust)", NAME, version);
     Ok(())
 }
--- a/src/tools/kata-ctl/src/types.rs
+++ b/src/tools/kata-ctl/src/types.rs
@@ -1,0 +1,35 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::Result;
+use strum_macros::EnumString;
+
+// Builtin check command handler type.
+pub type BuiltinCmdFp = fn(args: &str) -> Result<()>;
+
+// CheckType encodes the name of each check provided by kata-ctl.
+#[derive(Debug, strum_macros::Display, EnumString, PartialEq)]
+pub enum CheckType {
+    CheckCpu,
+    CheckNetwork,
+}
+
+// PermissionType is used to show whether a check needs to run with elevated (super-user)
+// privileges, or whether it can run as normal user.
+#[derive(strum_macros::Display, EnumString, PartialEq)]
+pub enum PermissionType {
+    Privileged,
+    NonPrivileged,
+}
+
+// CheckItem is used to encode the check metadata that each architecture
+// returns in a list of CheckItem's using the architecture implementation
+// of get_checks().
+pub struct CheckItem<'a> {
+    pub name: CheckType,
+    pub descr: &'a str,
+    pub fp: BuiltinCmdFp,
+    pub perm: PermissionType,
+}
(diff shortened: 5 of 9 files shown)


src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+        arch: ArchType::x86_64,
+        perm: PermissionType::NonPrivileged,
+    },
+
+    CheckItem {
+        name: "check-cpu",
+        descr: "This parameter performs the cpu check",
+        fp: check,
+        ut: UserType::NonRoot,
+        arch: ArchType::s390x,
```
jodh-intel, line 83: This is getting quite complex but rather than having a single list of `CheckItem` checks (`CHECK_LIST`), I think it would be better to have one `CHECK_LIST` per architecture. We can then have the generic code call something like `fn get_checks() -> Result<Vec<CheckItem>>`. That way, the generic code would call `get_checks()` and only get a list of the checks for the current architecture. It could either display the details of that list (for `--list`), or execute the functions in order for the normal mode of operation.

That also implies that `CheckItem`:

- Doesn't need the specify `arch`.
- Does need `fp` to point to a platform-specific check function.

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+
+        match self {
+            ArchType::aarch64 => write!(f, "aarch64"),
+            ArchType::powerpc64le => write!(f, "powerpc64le"),
+            ArchType::s390x => write!(f, "s390x"),
+            ArchType::x86_64 => write!(f, "x86_64"),
+        }
+    }
+}
```
jodh-intel, line 117: This is fine. An alternative is to use the `strum` and `strum-macros` crates to `#[derive(strum_macros::Display)]` on `ArchType` to avoid having to define this function. See https://docs.rs/strum_macros/0.24.3/strum_macros/derive.Display.html.

jodh-intel (commented):
Thanks @dborquez ! I'd like to take another look at this tomorrow but here are a few initial comments and thoughts...

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+        descr: "This parameter performs the cpu check",
+        fp: check,
+        ut: UserType::NonRoot,
+        arch: ArchType::x86_64,
+        perm: PermissionType::NonPrivileged,
+    },
+
+    CheckItem {
+        name: "check-network",
+        descr: "This parameter performs network check",
```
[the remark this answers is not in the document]
dborquez: @jodh-intel thank you, currently `network-check `is the slot for the function you mention (PR #123 ) but let me know if I'm wrong.

src/tools/kata-ctl/src/arch/mod.rs
```diff
@@ -33,6 +33,17 @@ pub use x86_64 as arch_specific;
 )))]
 compile_error!("unknown architecture");
 
-pub fn check() -> Result<()> {
-    arch_specific::check()
+pub fn get_arch_type() -> String {
+
+    let machine_kind = if cfg!(target_arch = "x86_64") {
```
Ji-Xinyou, line 38: nit: I would say use
```Rust
if cond1 {
    "str1"
} else if cond2 {
    "str2"
}.to_string()
```
instead of 
```Rust
let _str = if cond1 {
    "str1"
} else if cond2 {
    "str2"
};
_str.to_string()
```

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+    for check in CHECK_LIST {
+        if check.name.eq(name) {
+            return Ok(check.fp);
+        }
+    }
+
+    Err(anyhow!("Invalid command: {:?}", name))
+}
+
+// This function is called from each CLI argument section
```
Ji-Xinyou, line 130: Since the return value is a tuple, I think you may need to add what the `Result` and `bool` actually stands for in the comment.

Ji-Xinyou (commented):
Generally LGTM, but some nits....

You need a cargo fmt to work around the fmt problems.

cmaf (commented):
@dborquez if I understand correctly, this PR creates --list that will list checks for the specific arch you're using without actually performing the checks, is that right?

This seems useful on an initial run to make sure the test isn't performing undesired tests (like network checks), but I'm wondering if it would be useful to be able to specify the architecture you want to list checks for and default to your current architecture for development & debugging.

src/tools/kata-ctl/src/ops/check_ops.rs
dborquez: @jodh-intel `strum_macros` used instead of implementing the Display trait.

src/tools/kata-ctl/src/ops/check_ops.rs
dborquez: Thank you @Ji-Xinyou, I opt to only use Result as return value, because the bool returned value is not really useful.

src/tools/kata-ctl/src/ops/check_ops.rs
dborquez: @jodh-intel thank you, the lifetimes are really important regarding the scopes.

dborquez: @jodh-intel thank you, the structure is now compacted

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+    Privileged,
+    NonPrivileged,
+}
+
+type BuiltinCmdFp = fn(args: &str) -> Result<()>;
+
+
+struct CheckItem {
+    name: &'static str,
+    descr: &'static str,
```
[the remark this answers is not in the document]
dborquez: @jodh-intel thank you, the lifetimes are really important regarding the scopes.

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+    x86_64,
+    s390x,
+    aarch64,
+    powerpc64le,
+}
+
+enum UserType {
+    Root,
+    NonRoot,
+}
```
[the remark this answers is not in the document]
dborquez: @jodh-intel thanks, done

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+
+enum PermissionType {
+    Privileged,
+    NonPrivileged,
+}
+
+type BuiltinCmdFp = fn(args: &str) -> Result<()>;
+
+
```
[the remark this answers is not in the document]
dborquez: @jodh-intel thanks

src/tools/kata-ctl/src/arch/mod.rs
```diff
@@ -33,6 +33,17 @@ pub use x86_64 as arch_specific;
 )))]
 compile_error!("unknown architecture");
 
-pub fn check() -> Result<()> {
-    arch_specific::check()
+pub fn get_arch_type() -> String {
```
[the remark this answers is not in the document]
dborquez: `ArchType `was removed as the CheckItem will be defined per architecture

src/tools/kata-ctl/src/arch/mod.rs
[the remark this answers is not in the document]
dborquez: This function is not more needed as we opt to define `CheckItem `per architecture

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,44 +1,198 @@
...
+    aarch64,
+    powerpc64le,
+}
+
+enum UserType {
+    Root,
+    NonRoot,
+}
+
+enum PermissionType {
```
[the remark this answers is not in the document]
dborquez: @jodh-intel thank you, comment added.

src/tools/kata-ctl/src/arch/mod.rs
dborquez: @Ji-Xinyou thank you for your feedback, the` machine_kind` variable was removed because we opt to define a `CheckItem` list per architecture.

dborquez:
> if I understand correctly, this PR creates `--list` that will list checks for the specific arch you're using without actually performing the checks, is that right?

@cmaf thank you for your question. Yes, the `--list` argument lists the checks for the architecture that kata-ctl is running on.

> This seems useful on an initial run to make sure the test isn't performing undesired tests (like network checks), but I'm wondering if it would be useful to be able to specify the architecture you want to list checks for and default to your current architecture for development & debugging. Or, list all checks for all architectures since it is getting kind of complex to have the `list` command run architecture-specific code. Maybe it isn't overly complex but it does seem like it will be confusing for future contributors to maintain.

@cmaf the result of `kata-ctl check list` is a list of available checks for the architecture where `kata-ctl` is running on.

src/tools/kata-ctl/src/types.rs
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::{anyhow, Result};
+use strum_macros::Display;
+
+// Builtin check command handler type
+pub type BuiltinCmdFp = fn(args: &str) -> Result<()>;
+
+// Used to show whether a check needs to run with elevated (super-user) privileges, or whether it can run
```
jodh-intel, line 12: Truncated line?

src/tools/kata-ctl/src/types.rs
```diff
@@ -0,0 +1,25 @@
...
+#[derive(strum_macros::Display)]
+pub enum PermissionType {
+    Privileged,
+    NonPrivileged,
+}
+
+// Command metadata for the checks
+pub struct CheckItem<'a> {
```
jodh-intel, line 20: Could you add a comment explaining that each `CheckItem` is implicitly architecture-specific: each architecture returns a list of `CheckItem`'s using that architectures implementation of `get_checks()`.

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,42 +1,101 @@
...
+        if check.name.eq(name) {
+            return Ok(check.fp);
+        }
+    }
+
+    Err(anyhow!("Invalid command: {:?}", name))
+}
+
+// This function is called from each 'kata-ctl check' argument section
+fn handle_builtin_check(check: &str, args: &str) -> Result<()> {
```
jodh-intel, line 34: I think it would be better to create an enum for the check name. Something like:

```rust
#[derive(Debug, Clone, strum_macros::Display, EnumString)]
enum CheckType {
  CheckCpu,
  CheckNetwork,
  // ...
};
```

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,42 +1,101 @@
...
+fn print_check_list() -> Result<()> {
+    let cmds = get_client_cmd_details();
+
+    if cmds.is_empty() {
+        println!("Checks not found!\n");
+
+        return Ok(());
+    }
+
+    cmds.iter().for_each(|n| println!(" - {}", n));
```
jodh-intel, line 63: Cute! :smile:

jodh-intel (commented):
Thanks @dborquez - a few comments...

jodh-intel:
> I'm wondering if it would be useful to be able to specify the architecture you want to list checks for and default to your current architecture for development & debugging.

@cmaf - I agree that could be useful. But how about we look at that as a follow-on PR?

cmaf (approved):

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -16,7 +18,20 @@ mod arch_specific {
     const CPU_FLAGS_INTEL: &[&str] = &["lm", "sse4_1", "vmx"];
     const CPU_ATTRIBS_INTEL: &[&str] = &["GenuineIntel"];
 
-    // check cpu
+    // List of check functions
+    static CHECK_LIST: &[CheckItem] = &[
+        CheckItem {
+            name: CheckType::CheckCpu,
+            descr: "This parameter performs the cpu check",
+            fp: check,
```
jodh-intel, line 26:
```suggestion
            fp: check_cpu,
```

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -43,9 +58,8 @@ mod arch_specific {
         Ok(())
     }
 
-    pub fn check() -> Result<()> {
-        println!("INFO: check: x86_64");
-
+    pub fn check(args: &str) -> Result<()> {
```
jodh-intel, line 61: - I think we can drop all the arch-specific `check()` functions now as they are replaced by `get_checks()`.
- I'm not sure we need `args` currently so we can either remove it and add when we need it, or keep it. But if we keep it, you'll need to change `check_cpu()` to be `check_cpu(args: &str)`.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -43,9 +58,8 @@ mod arch_specific {
...
+
```
jodh-intel, line 62: Nit: There is extraneous whitespace on this line (although it doesn't show on GH). I'd recommend getting `rustfmt` set up to automatically remove things like that.

src/tools/kata-ctl/src/args.rs
```diff
@@ -58,6 +58,9 @@ pub enum CheckSubCommand {
 
     /// Only compare the current and latest available versions
     CheckVersionOnly,
+
+    /// List all available checks
+    List,
```
jodh-intel, line 63: More extraneous whitespace here.

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -1,42 +1,100 @@
 // Copyright (c) 2022 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
-//
+//00
```
jodh-intel, line 4: Can you remove this line.

src/tools/kata-ctl/src/types.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// PermissionType is used to show whether a check needs to run with elevated (super-user)
+// privileges, or whether it can run as normal user.
+#[derive(strum_macros::Display, EnumString, PartialEq)]
+pub enum PermissionType {
+    Privileged,
+    NonPrivileged,
+}
+// CheckItem is used to encode the check metadata that each architecture
```
jodh-intel, line 26: Trailing whitespace.

jodh-intel (commented):
Thanks @dborquez - a few comments.

Please can you fix the warnings so that `make` runs successfully too?

src/tools/kata-ctl/src/arch/x86_64/mod.rs
dborquez: Yes I share this decision

src/tools/kata-ctl/src/arch/x86_64/mod.rs
dborquez: Yeah I noticed cargo fmt fixed this one

src/tools/kata-ctl/src/args.rs
dborquez: thanks @jodh-intel, fixed

src/tools/kata-ctl/src/ops/check_ops.rs
dborquez: @jodh-intel thanks

jodh-intel:
Thanks @dborquez - Please can you fix the conflict in `Cargo.lock` though - the code isn't currently buildable.

jodh-intel:
Hi @dborquez - Could you resolve the conflicts on this branch caused by #5786 landing.

jodh-intel (approved):
Thanks @dborquez.

lgtm!

studychao:
image
Seems like this PR breaks ARM CI. Could your help to fix it? @dborquez 

cc @jodh-intel  @cmaf

jongwu:
Thanks @studychao

PR #5736, opened 2022-11-23
runtime: Start QEMU undaemonized and get logs

### LINKED ISSUE #5780: QEMU error logs are missing
When QEMU fails for some reason, the error message it generates is lost. No trace of it can be found in the kata logs. This seriously hampers the capacity to investigate issues.

There is some logic that passes `-D /some/path/qemu.log` and collects the content of the file when QEMU terminates. `-D` is not meant for this : it goes along with `-d` which is only useful for debugging/tracing specific parts of QEMU, most of which aren't even actioned when running with KVM.

QEMU error messages should be collected from stderr. Unfortunately, kata also passes `-daemonize`, in which case QEMU redirects stderr to /dev/null... 

Discussions with the QEMU community revealed that:
- the `-daemonized` option isn't recommended. It is legacy stuff only kept to not break existing setups.
- even if we hack QEMU to write errors in the log file passed to `-D`, there isn't any size limiting and an malicious workload could use that to fill up the filesystem
- libvirt which is by far the first QEMU user doesn't use it on purpose for this reason

Relying on a mix of feature that were clearly not designed with proper error reporting in mind, either not used by anyone (-daemonize) or only used by a handful of QEMU hackers (-d and -D) looks very fragile.

QEMU top programmers warmly recommend to do what libvirt does : not daemonize and pass a pipe to collect the logs.

gkurz:
The goal of this PR is to collect error messages emitted by QEMU. This requires to not put `-daemonize` on the QEMU command line otherwise it won't produce any output. This has several impacts:
- no guarantees that QEMU is functional when LaunchQemu() returns
- QEMU becomes a direct child that should be reaped when terminated

The QMP connection is now pre-established. The listening socket is then passed to QEMU : the shim just has to wait for the QMP greeting or a disconnect, similarly as what we get with the current code.

QEMU is now started with [Cmd.Start](https://pkg.go.dev/os/exec#Cmd.Start) instead of [Cmd.Run](https://pkg.go.dev/os/exec#Cmd.Run). The Cmd structure is kept around until QEMU is terminated and waited for with [Cmd.Wait](https://pkg.go.dev/os/exec#Cmd.Wait).

Some logic is added to collect everything QEMU outputs on stderr and pass it to the logging infra.

All new paths introduced by these series are directly or indirectly controlled by the Knobs.Daemonize flag,
which is only flipped in the last commit when all the rest is in place. Note this wasn't made configurable and
the paths for the daemonized case are preserved. This is on purpose. When everyone is happy with this change,
another PR will be created to remove all the code involved in `-daemonize`.

Fixes #5780

### DIFF
--- a/src/runtime/pkg/govmm/qemu/examples_test.go
+++ b/src/runtime/pkg/govmm/qemu/examples_test.go
@@ -27,13 +27,16 @@ func Example() {
 	// resources
 	params = append(params, "-m", "370", "-smp", "cpus=2")
 
-	// LaunchCustomQemu should return as soon as the instance has launched as we
-	// are using the --daemonize flag.  It will set up a unix domain socket
-	// called /tmp/qmp-socket that we can use to manage the instance.
-	_, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
+	// LaunchCustomQemu should return immediately. We must then wait
+	// the returned process to terminate as we are using the --daemonize
+	// flag.
+	// It will set up a unix domain socket called /tmp/qmp-socket that we
+	// can use to manage the instance.
+	proc, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
 	if err != nil {
 		panic(err)
 	}
+	proc.Wait()
 
 	// This channel will be closed when the instance dies.
 	disconnectedCh := make(chan struct{})
--- a/src/runtime/pkg/govmm/qemu/examples_test.go
+++ b/src/runtime/pkg/govmm/qemu/examples_test.go
@@ -32,7 +32,7 @@ func Example() {
 	// flag.
 	// It will set up a unix domain socket called /tmp/qmp-socket that we
 	// can use to manage the instance.
-	proc, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
+	proc, _, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
 	if err != nil {
 		panic(err)
 	}
--- a/src/runtime/pkg/govmm/qemu/qemu.go
+++ b/src/runtime/pkg/govmm/qemu/qemu.go
@@ -14,7 +14,6 @@
 package qemu
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"log"
@@ -2985,12 +2984,8 @@ func (config *Config) appendFwCfg(logger QMPLog) {
 //
 // The Config parameter contains a set of qemu parameters and settings.
 //
-// This function writes its log output via logger parameter.
-//
-// The function will block until the launched qemu process exits.  "", nil
-// will be returned if the launch succeeds.  Otherwise a string containing
-// the contents of stderr + a Go error object will be returned.
-func LaunchQemu(config Config, logger QMPLog) (string, error) {
+// See LaunchCustomQemu for more information.
+func LaunchQemu(config Config, logger QMPLog) (*exec.Cmd, error) {
 	config.appendName()
 	config.appendUUID()
 	config.appendMachine()
@@ -3013,7 +3008,7 @@ func LaunchQemu(config Config, logger QMPLog) (string, error) {
 	config.appendSeccompSandbox()
 
 	if err := config.appendCPUs(); err != nil {
-		return "", err
+		return nil, err
 	}
 
 	ctx := config.Ctx
@@ -3044,16 +3039,14 @@ func LaunchQemu(config Config, logger QMPLog) (string, error) {
 //
 // This function writes its log output via logger parameter.
 //
-// The function will block until the launched qemu process exits.  "", nil
-// will be returned if the launch succeeds.  Otherwise a string containing
-// the contents of stderr + a Go error object will be returned.
+// The function returns cmd, nil where cmd is a Go exec.Cmd object
+// representing the QEMU process if launched successfully. Otherwise
+// nil, err where err is a Go error object is returned.
 func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*os.File,
-	attr *syscall.SysProcAttr, logger QMPLog) (string, error) {
+	attr *syscall.SysProcAttr, logger QMPLog) (*exec.Cmd, error) {
 	if logger == nil {
 		logger = qmpNullLogger{}
 	}
-
-	errStr := ""
 
 	if path == "" {
 		path = "qemu-system-x86_64"
@@ -3068,15 +3061,12 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
 
 	cmd.SysProcAttr = attr
 
-	var stderr bytes.Buffer
-	cmd.Stderr = &stderr
 	logger.Infof("launching %s with: %v", path, params)
 
-	err := cmd.Run()
+	err := cmd.Start()
 	if err != nil {
 		logger.Errorf("Unable to launch %s: %v", path, err)
-		errStr = stderr.String()
-		logger.Errorf("%s", errStr)
+		return nil, err
 	}
-	return errStr, err
+	return cmd, nil
 }
--- a/src/runtime/pkg/govmm/qemu/qemu.go
+++ b/src/runtime/pkg/govmm/qemu/qemu.go
@@ -2340,6 +2340,9 @@ type QMPSocket struct {
 	// Type is the socket type (e.g. "unix").
 	Type QMPSocketType
 
+	// QMP listener file descriptor to be passed to qemu
+	FD *os.File
+
 	// Name is the socket name.
 	Name string
 
@@ -2352,7 +2355,8 @@ type QMPSocket struct {
 
 // Valid returns true if the QMPSocket structure is valid and complete.
 func (qmp QMPSocket) Valid() bool {
-	if qmp.Type == "" || qmp.Name == "" {
+	// Exactly one of Name of FD must be set.
+	if qmp.Type == "" || (qmp.Name == "") == (qmp.FD == nil) {
 		return false
 	}
 
@@ -2692,7 +2696,13 @@ func (config *Config) appendQMPSockets() {
 			continue
 		}
 
-		qmpParams := append([]string{}, fmt.Sprintf("%s:%s", q.Type, q.Name))
+		var qmpParams []string
+		if q.FD != nil {
+			qemuFDs := config.appendFDs([]*os.File{q.FD})
+			qmpParams = append([]string{}, fmt.Sprintf("%s:fd=%d", q.Type, qemuFDs[0]))
+		} else {
+			qmpParams = append([]string{}, fmt.Sprintf("%s:path=%s", q.Type, q.Name))
+		}
 		if q.Server {
 			qmpParams = append(qmpParams, "server=on")
 			if q.NoWait {
--- a/src/runtime/pkg/govmm/qemu/qemu.go
+++ b/src/runtime/pkg/govmm/qemu/qemu.go
@@ -16,6 +16,7 @@ package qemu
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
@@ -2985,7 +2986,7 @@ func (config *Config) appendFwCfg(logger QMPLog) {
 // The Config parameter contains a set of qemu parameters and settings.
 //
 // See LaunchCustomQemu for more information.
-func LaunchQemu(config Config, logger QMPLog) (*exec.Cmd, error) {
+func LaunchQemu(config Config, logger QMPLog) (*exec.Cmd, io.ReadCloser, error) {
 	config.appendName()
 	config.appendUUID()
 	config.appendMachine()
@@ -3008,7 +3009,7 @@ func LaunchQemu(config Config, logger QMPLog) (*exec.Cmd, error) {
 	config.appendSeccompSandbox()
 
 	if err := config.appendCPUs(); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	ctx := config.Ctx
@@ -3039,11 +3040,12 @@ func LaunchQemu(config Config, logger QMPLog) (*exec.Cmd, error) {
 //
 // This function writes its log output via logger parameter.
 //
-// The function returns cmd, nil where cmd is a Go exec.Cmd object
-// representing the QEMU process if launched successfully. Otherwise
-// nil, err where err is a Go error object is returned.
+// The function returns cmd, reader, nil where cmd is a Go exec.Cmd object
+// representing the QEMU process and reader a Go io.ReadCloser object
+// connected to QEMU's stderr, if launched successfully. Otherwise
+// nil, nil, err where err is a Go error object is returned.
 func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*os.File,
-	attr *syscall.SysProcAttr, logger QMPLog) (*exec.Cmd, error) {
+	attr *syscall.SysProcAttr, logger QMPLog) (*exec.Cmd, io.ReadCloser, error) {
 	if logger == nil {
 		logger = qmpNullLogger{}
 	}
@@ -3061,12 +3063,17 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
 
 	cmd.SysProcAttr = attr
 
+	reader, err := cmd.StderrPipe()
+	if err != nil {
+		logger.Errorf("Unable to connect stderr to a pipe")
+		return nil, nil, err
+	}
 	logger.Infof("launching %s with: %v", path, params)
 
-	err := cmd.Start()
+	err = cmd.Start()
 	if err != nil {
 		logger.Errorf("Unable to launch %s: %v", path, err)
-		return nil, err
+		return nil, nil, err
 	}
-	return cmd, nil
+	return cmd, reader, nil
 }
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -14,6 +14,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"net"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -368,7 +369,6 @@ func (q *qemu) createQmpSocket() ([]govmmQemu.QMPSocket, error) {
 	return []govmmQemu.QMPSocket{
 		{
 			Type:   "unix",
-			Name:   q.qmpMonitorCh.path,
 			Server: true,
 			NoWait: true,
 		},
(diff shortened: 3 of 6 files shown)


fidencio:
@fengwang666, mind to take a look at this one?

src/runtime/virtcontainers/qemu.go
```diff
@@ -667,6 +672,18 @@ func (q *qemu) CreateVM(ctx context.Context, id string, network Network, hypervi
 		qemuConfig.Devices = q.arch.appendPCIeRootPortDevice(qemuConfig.Devices, hypervisorConfig.PCIeRootPort, memSize32bit, memSize64bit)
 	}
 
+	if q.config.Debug && !qemuConfig.Knobs.Daemonize {
+		qemuConfig.LogFunc =
+			func(qemuPid int, qemuStderr io.ReadCloser) {
```
fengwang666, line 677: Would it be better to also pass in the sandbox id instead given that most of the logs have sandbox id?

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3058,15 +3071,21 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
...
 		logger.Errorf("Unable to launch %s: %v", path, err)
+		stderr := new(bytes.Buffer)
+		io.Copy(stderr, stderrIn)
 		errStr = stderr.String()
 		logger.Errorf("%s", errStr)
+		return nil, errStr, err
+	}
+
+	if logFunc != nil {
+		go logFunc(cmd.Process.Pid, stderrIn)
```
fengwang666, line 3088: This go routine needs to be running before line 3084, right? Otherwise the go routine might never start. Maybe move it to line 3079 and synchronize it with the main thread.

src/runtime/virtcontainers/qemu.go
```diff
@@ -667,6 +672,18 @@ func (q *qemu) CreateVM(ctx context.Context, id string, network Network, hypervi
 		qemuConfig.Devices = q.arch.appendPCIeRootPortDevice(qemuConfig.Devices, hypervisorConfig.PCIeRootPort, memSize32bit, memSize64bit)
 	}
 
+	if q.config.Debug && !qemuConfig.Knobs.Daemonize {
```
fengwang666, line 675: Why check `q.config.Debug`? Don't we always want the qemu error log?

src/runtime/virtcontainers/qemu.go
```diff
@@ -1088,6 +1178,13 @@ func (q *qemu) cleanupVM() error {
 			}).Debug("successfully removed the non root user")
 	}
 
+	if q.qemuCmd != nil {
+		if err := q.qemuCmd.Wait(); err != nil {
```
fengwang666, line 1182: Maybe I miss something here, but I thought Wait() will block forever until the QEMM process exists. In that case, don't you need to call Wait() in a separate goroutine?

fengwang666 (commented):
This is great! Thanks for making this change.
A couple additional comments:
1. Have you tested this change end to end and verified the error logs are now in containerd journal?
2. I know there are folks working on qemu support in the rust-shim. Are they aware of this change? cc @jodh-intel @pmores

gkurz:
> This is great! Thanks for making this change. A couple additional comments:
> 

Hi Feng ! Thanks for the review and best wishes for 2023 :smile:

>     1. Have you tested this change end to end and verified the error logs are now in containerd journal?
> 

I've tested this change in openshift and I do see QEMU errors in the journal.

>     2. I know there are folks working on qemu support in the rust-shim. Are they aware of this change? cc @jodh-intel @pmores

@pmores is aware of this work. I'm also tracking the addition of QEMU support in the rust shim. We should be good.

src/runtime/pkg/govmm/qemu/qemu.go
gkurz: > This go routine needs to be running before line 3084, right? Otherwise the go routine might never start. Maybe move it to line 3079 and synchronize it with the main thread.

Line 3084 is on the error path when QEMU couldn't be started. We don't need a goroutine in this case. Just collect the errors from `cmd.Start()` ... but thanks to your remark I know realize there's more to it : if QEMU didn't start then we don't need to collect stderr. I'll simplify the error path.

src/runtime/virtcontainers/qemu.go
gkurz: > Why check `q.config.Debug`? Don't we always want the qemu error log?

I've assumed this was wanted since it is checked in the existing code for `-D qemu.log`, but I'd be glad to always get QEMU errors if you think it is preferable.

src/runtime/virtcontainers/qemu.go
fengwang666: Yeah. Would it be better to call `Wait()` in the `stopVM` method instead of `cleanupVM`?

src/runtime/virtcontainers/qemu.go
gkurz: I'll remove the check in the next version.

src/runtime/virtcontainers/qemu.go
gkurz: After some more thoughts, I'll move Wait() back to govmm but put it in a goroutine.

gkurz:
I had to rework the logic in order to call Wait() at the right place (patch 3 and following). I've also added some extra patches to remove all the code that gets obsoleted by the _undaemonization_.

Revelant changes:
- govmm returns the pid of QEMU instead of an *exec.Cmd
- call Wait() from a goroutine started by LaunchCustomQemu()
- call the log function from this goroutine **before** calling Wait() since it will close the pipes
- collect QEMU messages unconditionally
- map QEMU error and warnings to the equivalent logger levels
- drop all the code related to the  `-pidfile`, `-D` and `-daemonize` QEMU options

PTAL @fengwang666

gkurz:
The `jenkins-ci-fedora-35-x86_64-crio_k8s` check is failing because of the disappearance of the pid file. I'll post a PR for the test repo and add a depend-on tag.

gkurz:
> The `jenkins-ci-fedora-35-x86_64-crio_k8s` check is failing because of the disappearance of the pid file. I'll post a PR for the test repo and add a depend-on tag.

Hmmm... I already have https://github.com/kata-containers/tests/pull/5324 that depends-on this PR. I'll just move the cleanup patches to some subsequent PR then.

src/runtime/virtcontainers/qemu.go
```diff
@@ -882,20 +964,31 @@ func (q *qemu) StartVM(ctx context.Context, timeout int) error {
...
-			if err == nil {
-				strErr += string(b)
-			}
+		q.Logger().WithError(err).Error("failed to launch qemu")
+		return fmt.Errorf("failed to launch qemu: %s", err)
+	}
+	if q.qemuConfig.Knobs.Daemonize {
+		// Wait for QEMU to daemonize itself. Don't check errors in
+		// case govmm has already reaped the upper process.
+		syscall.Wait4(qemuPid, nil, 0, nil)
```
bergwolf, line 975: IIUC, the pid returned by `LaunchQemu` is the qemu command line process pid that is not the daemonized qemu main thread id. And the pid is already waited in `LaunchCustomQemu`.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3034,17 +3049,14 @@ func LaunchQemu(config Config, logger QMPLog) (string, error) {
...
 // This function writes its log output via logger parameter.
 //
-// The function will block until the launched qemu process exits.  "", nil
-// will be returned if the launch succeeds.  Otherwise a string containing
-// the contents of stderr + a Go error object will be returned.
+// The function will return pid, nil if the launch succeeds.  Otherwise
+// 0 + a Go error object will be returned.
 func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*os.File,
-	attr *syscall.SysProcAttr, logger QMPLog) (string, error) {
+	attr *syscall.SysProcAttr, logger QMPLog, logFunc LogFunc) (int, error) {
```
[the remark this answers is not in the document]
gkurz: I'll look into it.

src/runtime/pkg/govmm/qemu/examples_test.go
```diff
@@ -27,13 +28,15 @@ func Example() {
...
-	_, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
+	// LaunchCustomQemu should return immediately. We must then wait4
+	// the returned pid to terminate as we are using the --daemonize flag.
+	// It will set up a unix domain socket called /tmp/qmp-socket that we
+	// can use to manage the instance.
+	pid, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil, nil)
 	if err != nil {
 		panic(err)
 	}
+	syscall.Wait4(pid, nil, 0, nil)
```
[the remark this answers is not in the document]
gkurz: syscall.Wait4() is used as a synchronization point : we want to be sure that the top level process has exited before going forward.

src/runtime/virtcontainers/qemu.go
gkurz: Same remark here : qemuPid corresponds to the top level QEMU process. We're just synchronizing with its termination to be sure that the daemonized QEMU is functional.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -2352,7 +2355,7 @@ type QMPSocket struct {
 
 // Valid returns true if the QMPSocket structure is valid and complete.
 func (qmp QMPSocket) Valid() bool {
-	if qmp.Type == "" || qmp.Name == "" {
+	if qmp.Type == "" || (qmp.Name == "" && qmp.FD == nil) {
```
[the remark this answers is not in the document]
gkurz: > The `false` condition should be `qmp.Type == "" || qmp.Name == "" || qmp.FD == nil`, correct? Otherwise, why do you even need to check whether `qmd.FD` is nil.

No. Either we pass the path that QEMU should bind the server socket to or we pass the fd of an already bound socket. In the latter case, the path already exists on the VFS and QEMU thus doesn't need it. It is thus valid to have one of Name of FD to be unset if the other one is set.

QEMU actually even prevents to pass both the path and the fd on the command line :

```
$ qemu-system-x86_64 -qmp unix:foo,fd=3,server,nowait
qemu-system-x86_64: -qmp unix:foo,fd=3,server,nowait: Exactly one of 'path', 'fd' or 'host' required
```

A stricter check is :

```
if qmp.Type == "" || (qmp.Name == "" && qmp.FD == nil) || (qmp.Name != "" && qmp.FD != nil) {
```

I'll rework this part to make it more obvious.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -2692,7 +2692,13 @@ func (config *Config) appendQMPSockets() {
 			continue
 		}
 
-		qmpParams := append([]string{}, fmt.Sprintf("%s:%s", q.Type, q.Name))
+		var qmpParams []string
+		if q.FD != nil {
```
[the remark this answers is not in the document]
gkurz: This is the first patch. Caller is still passing the socket name at this point and I want to preserve bisectability. This path should be removed in a subsequent cleanup PR though.

src/runtime/pkg/govmm/qemu/examples_test.go
```diff
@@ -27,13 +28,15 @@ func Example() {
 	// resources
 	params = append(params, "-m", "370", "-smp", "cpus=2")
 
-	// LaunchCustomQemu should return as soon as the instance has launched as we
-	// are using the --daemonize flag.  It will set up a unix domain socket
-	// called /tmp/qmp-socket that we can use to manage the instance.
-	_, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
+	// LaunchCustomQemu should return immediately. We must then wait4
+	// the returned pid to terminate as we are using the --daemonize flag.
```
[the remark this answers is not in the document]
gkurz: I've only made changes to this file that reflected the API changes. The example still passes `-daemonize` actually (look at line 26 above) so the comment must be kept. I can possibly adapt the example to start an non-daemonized QEMU later on.

src/runtime/pkg/govmm/qemu/qemu.go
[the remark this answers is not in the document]
c3d: FWIW, I think you can rewrite the stricter check as:
```
if qmp.Type == "" || (qmp.Name == "") != (qmp.FD == nil)
```

src/runtime/virtcontainers/qemu.go
```diff
@@ -812,6 +813,59 @@ func (q *qemu) setupVirtioMem(ctx context.Context) error {
 	return err
 }
 
+// setupEarlyQmpConnection creates a listener socket to be passed to QEMU
+// as a QMP listening endpoint. An initial connection is established, to
+// be used as the QMP client socket. This allows to detect an early failure
+// of QEMU instead of looping on connect until some timeout expires.
+func (q *qemu) setupEarlyQmpConnection() (net.Conn, error) {
+	monitorSockPath := q.qemuConfig.QMPSockets[0].Name
```
c3d, line 841: Why is it `QMPSockets[0]`? Is there any situation where we expect multiple QMP sockets?

src/runtime/virtcontainers/qemu.go
```diff
@@ -812,6 +813,59 @@ func (q *qemu) setupVirtioMem(ctx context.Context) error {
...
+
+	qmpListener, err := net.Listen("unix", monitorSockPath)
+	if err != nil {
+		q.Logger().WithError(err).Errorf("Unable to listen on unix socket address (%s)", monitorSockPath)
+		return nil, err
+	}
+
+	// A duplicate fd of this socket will be passed to QEMU. We must
```
c3d, line 849: The comment is useful, but it is incomplete:
* Who does the duplication
* Why is the duplication needed?

I'm trying to figure that out, because `appendFDs` clearly assumes that all file descriptors are consecutive and match the `os.File`. If there is an implicit `dup()` call, then it presumably breaks this assumption.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -2989,7 +2989,7 @@ func (config *Config) appendFwCfg(logger QMPLog) {
 // The function will block until the launched qemu process exits.  "", nil
```
c3d, line 2995: The return value needs to be updated in the comment

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3043,17 +3043,14 @@ func LaunchQemu(config Config, logger QMPLog) (string, error) {
 //
 // This function writes its log output via logger parameter.
 //
-// The function will block until the launched qemu process exits.  "", nil
-// will be returned if the launch succeeds.  Otherwise a string containing
-// the contents of stderr + a Go error object will be returned.
+// The function will return pid, nil if the launch succeeds.  Otherwise
+// 0 + a Go error object will be returned.
```
c3d, line 3053:
```suggestion
// 0, err where err is a Go error object will be returned.
```

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3071,11 +3068,15 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
...
+	err := cmd.Start()
 	if err != nil {
 		logger.Errorf("Unable to launch %s: %v", path, err)
-		errStr = stderr.String()
-		logger.Errorf("%s", errStr)
+		return 0, err
 	}
-	return errStr, err
+
+	go func() {
```
c3d, line 3082: The commit description indicates that you intend to _defer_, so I wonder if the intent was to have a `defer func()` here. If not, I do not understand how that is supposed to work, and a comment would be welcome.

src/runtime/virtcontainers/qemu.go
```diff
@@ -2568,15 +2580,8 @@ func (q *qemu) Cleanup(ctx context.Context) error {
...
-		q.Logger().WithError(err).Error("Could not read qemu pid file")
-		return []int{0}
-	}
-
-	pid, err := strconv.Atoi(strings.Trim(string(data), "\n\t "))
-	if err != nil {
-		q.Logger().WithError(err).Error("Could not convert string to int")
+	pid := q.state.Pid
+	if pid == 0 {
 		return []int{0}
```
c3d, line 2604: I wonder why we return an array with a zero inside here? This seems to sort of rely on the fact that 0 is an error value of `GetHypervisorPid`, but it's not very clean, and it does not match what `GetHypervisorPid` actually considers the error case:

```
()
	if len(pids) == 0 {
		return 0
	}
	return pids[0]
}
```

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3075,6 +3080,9 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
 	}
 
 	go func() {
+		if logFunc != nil {
+			logFunc(cmd.Process.Pid, stderrIn)
+		}
```
c3d, line 3085: OK, that part of the commit answers my comment on what you intended to do with this goroutine.

src/runtime/virtcontainers/qemu.go
```diff
@@ -685,6 +687,23 @@ func (q *qemu) CreateVM(ctx context.Context, id string, network Network, hypervi
...
 
+	if !qemuConfig.Knobs.Daemonize {
+		qemuConfig.LogFunc = func(pid int, stderr io.ReadCloser) {
+			q.Logger().Infof("Start logging QEMU (qemuPid=%d)", pid)
+			scanner := bufio.NewScanner(stderr)
+			for scanner.Scan() {
+				warnRE := regexp.MustCompile("^[^:]+: warning:")
+				text := scanner.Text()
+				if warnRE.MatchString(text) {
+					q.Logger().WithField("qemuPid", pid).Warning(text)
```
c3d, line 698: Unless I'm mistaken, you are keeping the `warning: ` stanza here. Maybe strip it from the text before passing it to `Warning()`?

c3d (approved):
Hi @gkurz, thanks a lot for this series.

Overall, it looks good enough for me to approve it, but I made a few suggestions that could be addressed in a follow-up.

Is there a CI test with / without the daemonize option, just making sure that we can still run with the two modes?

fengwang666 (dismissed):
LGTM. Thanks! @gkurz

src/runtime/virtcontainers/qemu.go
gkurz: The array was added in `govmm` before it got imported into the main `kata-containers` repo (see https://github.com/kata-containers/govmm/commit/6fe338d604357b10944220d0cac39f5712ba3f5e). I agree with you that this doesn't seem to be needed by the current code base. Maybe we can get rid of the complexity in a subsequent PR.

src/runtime/virtcontainers/qemu.go
gkurz: > The comment is useful, but it is incomplete:
> 
>     * Who does the duplication
> 

https://pkg.go.dev/net#TCPListener.File at line 841

>     * Why is the duplication needed?
> 

This is the only way I could find to get a file descriptor number for the listening socket. Not sure why `File()` calls `dup()` internally though.

> 
> I'm trying to figure that out, because `appendFDs` clearly assumes that all file descriptors are consecutive and match the `os.File`. If there is an implicit `dup()` call, then it presumably breaks this assumption.

I must confess that I didn't dig into the details that much, but assuming that all file descriptors are consecutive looks like a terrible idea... I've added some traces...

```
	for i := range fds {
		fdInts = append(fdInts, oldLen+3+i)
		global_logger.Infof(" appendFDs fds[%d]=%v (fd %d) -> [%d]", i, fds[i], fds[i].Fd(), oldLen+3+i)
	}
```

...  and "input" descriptors are definitely not consecutive

```
appendFDs [0] (&{0xc0003cc000}) 11 -> [3]

appendFDs [0] (&{0xc0003cc4e0}) 22 -> [4]

appendFDs [0] (&{0xc0003cc480}) 21 -> [5]
```

I think that the final descriptors are made consecutive (starting from 3) later on with `dup2()`.

src/runtime/pkg/govmm/qemu/qemu.go
gkurz: > The commit description indicates that you intentionally defer the `Wait` to a goroutine. But since there is no channel or anything to communicate with that goroutine, I do not understand what you expect the `Wait()` to actually do for you?
> 

`Wait` must be called for reaping purpose, especially in the case QEMU isn't a daemon.

> Maybe it's a hook for a later commit? If so it would be nice for the commit message to explain. If there is another reason for that `Wait()`, I must say it is mysterious to me.
> 
> Update: It is indeed used for logging in a later commit.

Yeah. The idea is that `Wait` also closes the `stderr` pipe that will be added for logging. As explained at https://go.dev/src/os/exec/exec.go#L780 , `It is thus incorrect to call Wait before all reads from the pipe have completed.`

I can document that.

src/runtime/virtcontainers/qemu.go
gkurz: 0 is guaranteed to never be a valid pid : []int{} and []int{0} are thus kind of _equivalent_ . But I agree this looks messy. I'll fix that.

gkurz:
> Hi @gkurz, thanks a lot for this series.
> 
> Overall, it looks good enough for me to approve it, but I made a few suggestions that could be addressed in a follow-up.
> 

I shall actually re-factor the commits a bit. Especially get rid of the complexity of having both the legacy and the pre-connected QMP connection which seem to have confused reviewers.

> Is there a CI test with / without the daemonize option, just making sure that we can still run with the two modes?

I haven't looked into that yet. I'd rather get rid of the daemonize option entirely BTW.

gkurz:
Huge rework again. Highlights :
- re-order patches  : first handle the QMP socket, then launch QEMU with Start(), then un-daemonize and finally add logging
- revert to LaunchQemu returning an exec.Cmd instead of a pid
- move the goroutine that handles the logging and Wait to virtcontainers

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3058,15 +3063,17 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
...
+	}
 	logger.Infof("launching %s with: %v", path, params)
 
-	err := cmd.Run()
+	err = cmd.Start()
 	if err != nil {
 		logger.Errorf("Unable to launch %s: %v", path, err)
-		errStr = stderr.String()
-		logger.Errorf("%s", errStr)
+		return nil, nil, err
```
c3d, line 3076: Why don't you return the reader in that case? Do we assume that if we failed to start, there is nothing on `stderr`?

src/runtime/virtcontainers/qemu.go
c3d: That worries me a little 😄 

@sameo You seem to have written `appendFDs`. Could you explain the rationale for why the FDs are seen as being consecutive, and whether we need to anything special to keep that invariant true?

c3d: @gkurz I think it might be worth adding your observations to the comment. But let's wait for @sameo's explanations, I may be reading the meaning of the returned `fdInts` incorrectly.

c3d (approved):
Still approved after the changes.

c3d:
@gkurz You still have `do-not-merge` set. Any reason for that?

src/runtime/pkg/govmm/qemu/qemu.go
gkurz: Yes. AFAICT from the source, the go library doesn't write anything on stderr. Moreover it closes all the pipes on errors. I can't think of anything else writing to file descriptor 2.

gkurz:
> @gkurz You still have `do-not-merge` set. Any reason for that?

Not anymore. It's an oversight.

fengwang666 (approved):
Thanks! @gkurz

gkurz:
Not being able to collect QEMU logs is a bug. It deserves to be backported to stable-3.0.

Issue #5741, opened 2022-11-24
kernel: CONFIG_X86_SGX is not available in older kernels 


studychao:
`CONFIG_X86_SGX` is introduced after kernel 5.11, and that config is a default x86_64 config for Kata `build-kernel.sh` script.
But if we use `-v` to specify any kernel version below 5.11 will cause an inevitable error because `CONFIG_X86_SGX` is not supported in older kernels and that may cause problem for the situation if we need kernel version below 5.11.

So I propose to put `CONFIG_X86_SGX` into `whitelist.conf` to avoid break building guest kernel below 5.11.

PR #5742, opened 2022-11-24
kernel: add CONFIG_X86_SGX into whitelist

studychao:
CONFIG_X86_SGX is introduced after kernel 5.11, and that config is a default x86_64 config for Kata build-kernel.sh script. But if we use -v to specify any kernel version below 5.11 will cause an inevitable error because CONFIG_X86_SGX is not supported in older kernels and that may cause problem for the situation if we need kernel version below 5.11.

So I propose to put CONFIG_X86_SGX into whitelist.conf to avoid break building guest kernel below 5.11.

fixes: #5741


studychao:
@fidencio  Please help to take a look at this:)

fidencio (approved):
lgtm, thanks @studychao!

studychao:
Thanks! @fidencio

liubin (approved):

PR #5744, opened 2022-11-24
CC | agent: Update the merge_oci_process to properly manage the env variables

Jordan9500:
agent: Update the merge_oci_process to properly manage the env variables
Loop through the images enviroment variables, checking if it exists inside the target. If it does then do not append it. Add unit tests for correctly merging the env variables of the pod yaml and image itself in the container and image process

Signed-off-by: Jordan Jackson <jordan.jackson@ibm.com>


stevenhorsman (approved):
LGTM. Thanks Jordan!

PR #5749, opened 2022-11-24
tools: Use virtiofsd cache

### LINKED ISSUE #5356: CC | Use cached build for the cc-virtiofsd-tarball when possible
One way to speed up the current CC CIs is by caching the artefacts that don't change frequently and simply download and use them whenever they're needed.

This issue is for using the cached tarball whenever it's possible, instead of building the component.

GabyCT:
This PR allow us to use the virtiofsd cache instead of building from source.

Fixes #5356

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

### DIFF
--- a/tools/packaging/static-build/cache_components.sh
+++ b/tools/packaging/static-build/cache_components.sh
@@ -48,6 +48,16 @@ create_cache_asset() {
 	cat "sha256sum-${component_name}"
 }
 
+cache_virtiofsd_artifacts() {
+	local binary_name="virtiofsd"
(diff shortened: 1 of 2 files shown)


tools/packaging/static-build/cache_components.sh
```diff
@@ -65,7 +76,7 @@ main() {
 	local cloud_hypervisor_component="${cloud_hypervisor_component:-}"
 	local qemu_component="${qemu_component:-}"
 	local OPTIND
-	while getopts ":cqh:" opt
+	while getopts ":cqhv:" opt
```
liubin, line 79: It's better define the var at line 78:

```shell
	local virtiofsd_component="${virtiofsd_component:-}"
```

PR #5754, opened 2022-11-25
fix(clippy): fix clippy checks

Ji-Xinyou:
Fix clippy checks for components in need.

### DIFF
--- a/src/agent/rustjail/src/mount.rs
+++ b/src/agent/rustjail/src/mount.rs
@@ -32,7 +32,7 @@ use crate::log_child;
 
 // Info reveals information about a particular mounted filesystem. This
 // struct is populated from the content in the /proc/<pid>/mountinfo file.
-#[derive(std::fmt::Debug, PartialEq)]
+#[derive(std::fmt::Debug, PartialEq, Eq)]
 pub struct Info {
     mount_point: String,
     optional: String,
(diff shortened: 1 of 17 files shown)


liubin:
Which version of rust get these warnings when `cargo clippy`?

Ji-Xinyou:
> Which version of rust gets these warnings when `cargo clippy`?

I think it was 1.64?

liubin:
I mean we should only fix clippy warnings under the version we are using now.

https://github.com/kata-containers/kata-containers/blob/1da2d0603c882073b38fb586f6b46de51f584685/versions.yaml#L311-L319

@kata-containers/architecture-committee In which steps should we keep upgrading Rust version?

Ji-Xinyou:
> I mean we should only fix clippy warnings under the version we are using now.
> 
> https://github.com/kata-containers/kata-containers/blob/1da2d0603c882073b38fb586f6b46de51f584685/versions.yaml#L311-L319
> 
> @kata-containers/architecture-committee In which steps should we keep upgrading Rust version?

IMO the clippy checks are mostly back-compatable with previous rustc so the clippy suggestions could be independently fixed w.r.t. the rustc version.

Tim-Zhang:
Hi @Ji-Xinyou, I find this pr is the oldest PR that try to fix clippy warnings. Could you please rewrite it for [rust 1.66](https://github.com/kata-containers/kata-containers/pull/5967) and resolve the conflict?

Issue #5764, opened 2022-11-25
Some CIs are failing unintentionally


Ji-Xinyou:
Some tests are failing unintentionally

Issue #5737, opened 2022-11-24
dragonball UT failure


bergwolf:
See https://github.com/kata-containers/kata-containers/actions/runs/3525691666/jobs/5912721564

> thread 'address_space_manager::tests::test_create_address_space' panicked at 'called `Result::unwrap()` on an `Err` value: MmapGuestMemory(SeekEnd(Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }))', src/address_space_manager.rs:780:71

bergwolf:
While `RawFd` is convenient to use, we should avoid using it directly as much as possible as it bypasses all the rust ownership checking and life cycle management. We've seen random failures w.r.t. file descriptors messed up before in the agent. This alarms the runtime-rs code as well.

liubin:
Similar issues:

- #4103
- #4248
- #5764

Issue #4248, opened 2022-05-13
watcher::tests::create_tmpfs randomly fails on s390x


fidencio:
```
thread 'watcher::tests::create_tmpfs' panicked at 'assertion failed: is_mounted(WATCH_MOUNT_POINT_PATH).unwrap()', src/watcher.rs:1279:9
```

This has been observed in the following CI run: http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-PR/1658/console

jodh-intel:
Dup of #2833.

jodh-intel:
Seen again on #8155.

gkurz:
Seen again on #8287 which is currently gating the 3.2.0 and 3.3.0-alpha0 releases.

BbolroC:
Unfortunately this is something I cannot reproduce, but happens from time to time. I would say it is an infra issue. I will open up an internal discussion anyhow. Thanks.

BbolroC:
I will raise a PR to skip the test on s390x.

Issue #4103, opened 2022-04-15
test_do_write_stream fails frequently


liubin:
Sometimes static check complains this error(Full link: https://github.com/kata-containers/kata-containers/runs/6014924735?check_suite_focus=true):

It's not a 100% issue.

```

test rpc::tests::test_append_guest_hooks ... ok
test rpc::tests::test_add_arp_neighbors ... ok
test rpc::tests::test_get_memory_info ... ok
test rpc::tests::test_load_kernel_module ... ok
test rpc::tests::test_update_interface ... ok
test rpc::tests::test_update_routes ... ok
test rpc::tests::test_verify_cid ... ok
test sandbox::tests::add_and_get_container ... ok
thread 'sandbox::tests::add_guest_hooks' panicked at 'assertion failed: s.hooks.as_ref().unwrap().prestart.len() == 1', src/sandbox.rs:798:9
stack backtrace:
test rpc::tests::test_do_write_stream ... ok
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:48:5
   3: kata_agent::sandbox::tests::add_guest_hooks::{{closure}}::{{closure}}
             at ./src/sandbox.rs:798:9
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/future/mod.rs:80:19
   5: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/future/future.rs:119:9
   6: tokio::runtime::basic_scheduler::Inner<P>::block_on::{{closure}}::{{closure}}
             at /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.15.0/src/runtime/basic_scheduler.rs:240:62
   7: tokio::coop::with_budget::{{closure}}
[...]

PR #5765, opened 2022-11-25
fix(kata-ctl): fix test w.r.t tmp file creation

Ji-Xinyou:
Some ci are not passing due to file creation error. The failure is NOT the intention of the tests.

Fixes: #5764
Signed-off-by: Ji-Xinyou <jerryji0414@outlook.com>

### DIFF
--- a/src/tools/kata-ctl/src/utils.rs
+++ b/src/tools/kata-ctl/src/utils.rs
@@ -62,14 +62,15 @@ mod tests {
         let dir = tempdir().unwrap();
         let file_path = dir.path().join("proc-version");
         let path = file_path.clone();
-        let mut file = fs::File::create(file_path).unwrap();
-        writeln!(
-            file,
-            "Linux version 5.15.0-75-generic (buildd@lcy02-amd64-045)"
-        )
-        .unwrap();
-        let kernel = get_kernel_version(path.to_str().unwrap()).unwrap();
-        assert_eq!(kernel, "5.15.0-75-generic");
+        if let Ok(mut file) = fs::File::create(file_path) {
+            writeln!(
+                file,
+                "Linux version 5.15.0-75-generic (buildd@lcy02-amd64-045)"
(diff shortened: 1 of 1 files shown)


src/tools/kata-ctl/src/utils.rs
```diff
@@ -62,14 +62,15 @@ mod tests {
         let dir = tempdir().unwrap();
         let file_path = dir.path().join("proc-version");
         let path = file_path.clone();
-        let mut file = fs::File::create(file_path).unwrap();
```
liubin, line 65: In which case does the creating of file may fail?

Ji-Xinyou: Currently I have no idea. But from the CI's console output, this is where this test normally fails.

liubin: After this patch, the test will test nothing in case creating a temp file fails.
Do you have any logs?

Ji-Xinyou: Unfortunately there was no logs, it only said the unwrap causes the panic.
Currently, I only see this operation fails on s390x and arm, but I have no clue about the reasons.

liubin: @Ji-Xinyou I think this may only fail in GitHub actions, we have some similar errors:

- https://github.com/kata-containers/kata-containers/issues/4103
- https://github.com/kata-containers/kata-containers/issues/4248
- https://github.com/kata-containers/kata-containers/issues/5737

Ji-Xinyou: It is rlly weird...... It seems like it happens randomly

cmaf (commented):
@Ji-Xinyou I'm not seeing where it is failing due to file creation. I agree with @liubin, if file creation fails the test will not test anything.

Issue #5763, opened 2022-11-25
create a composite action to reduce the duplication for GitHub actions


liubin:
> As mentioned on slack, I think we might be able to use something like https://docs.github.com/en/actions/creating-actions/creating-a-composite-action to reduce the duplication, but as a short term fix this is fine. Thanks!

_Originally posted by @stevenhorsman in https://github.com/kata-containers/kata-containers/pull/5586#pullrequestreview-1169627713_

PR #5766, opened 2022-11-25
actions: use composite action to refactor static checks

liubin:
Using composite action to reduce the duplication.

Fixes: #5763


liubin:
All the checks almost have the same code style, only the last command is different, so instead of using composite action, using [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) will be easier.

Finally I chose to:

- split the static check for Dragonball into a new file, it has different format with others
- use matrix to run all other checks: 
          - "make vendor"
          - "make static-checks"
          - "make check"
          - "make test"
          - "sudo -E PATH=\"$PATH\" make test"

@fidencio @stevenhorsman

stevenhorsman (approved):
LGTM and a nice simple solution with matrix :) Thanks @liubin

PR #5769, opened 2022-11-25
kata-ctl: add host check for aarch64

jongwu:
For now, we can check if host support running kata by check if "/dev/kvm" exist on aarch64.

Fixes: #5768

### DIFF
--- a/src/tools/kata-ctl/src/arch/aarch64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/aarch64/mod.rs
@@ -8,8 +8,18 @@ pub use arch_specific::*;
 
 mod arch_specific {
     use anyhow::Result;
+    use std::path::Path;
+
+    const KVM_DEV: &str = "/dev/kvm";
 
     pub fn check() -> Result<()> {
-        unimplemented!("Check not implemented in aarch64")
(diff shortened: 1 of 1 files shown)


src/tools/kata-ctl/src/arch/aarch64/mod.rs
```diff
@@ -8,8 +8,18 @@ pub use arch_specific::*;
...
     use anyhow::Result;
+    use std::path::Path;
+
+    const KVM_DEV: &str = "/dev/kvm";
 
     pub fn check() -> Result<()> {
-        unimplemented!("Check not implemented in aarch64")
+        println!("INFO: check: aarch64");
+        if Path::new(KVM_DEV).exists() {
+            println!("Kata Containers can run on this host\n");
```
Ji-Xinyou, line 18: nit: May be add a hint that, with the VMMs which support aarch64.

liubin (approved):

src/tools/kata-ctl/src/arch/aarch64/mod.rs
liubin: @Ji-Xinyou Line 16 prints aarch64, so I think we can just merge it.

Issue #5770, opened 2022-11-25
CC | export kernel_version in the function responsible for building TEEs kernel


fidencio:
Now that we're caching the kernel, we're relying on the kernel version
 being exported.  This is already done for the CC kernel, but not for the
 TEE specific ones.

PR #5771, opened 2022-11-25
CC | build: export kernel_version to be used with its build script

fidencio:
Now that we're caching the kernel, we're relying on the kernel version
being exported.  This is already done for the CC kernel, but not for the
TEE specific ones.

Fixes: #5770


stevenhorsman (approved):
LGTM - thanks!

PR #5776, opened 2022-11-28
ci: let static checks don't depend on build

liubin:
Since the static-checks only need some config/version files to be generated, we can add a new target/rule to create the code required to let static-checks pass.

Fixes: #5777

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ TOOLS += log-parser
 TOOLS += runk
 TOOLS += trace-forwarder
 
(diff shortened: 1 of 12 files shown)


liubin:
Time needed with `build` is 30m+, with the `static-checks-build`, only takes 3m+.

fidencio (approved):
lgtm, thanks @liubin!

stevenhorsman (approved):
This looks reasonable to me and it's a great improvement in the time taken. Thanks @liubin

GabyCT (approved):

Issue #5787, opened 2022-11-29
runtime-rs: add support for oci spec


YushuoEdge:
According to oci runtime spec, we could execute some hook operations during container lifecycle:
https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle

In runtime-go, we are now supporting prestart hook, CreateRuntime hook, poststart hook and poststop hook, and we expect to complementary this kind of ability in runtime-rs.

YushuoEdge:
BTW, I would like to ask if we have plans to support all hooks in runtime spec?

bergwolf:
One major difference between runc and kata is that we have a vmm running that is supposed to be part of the container runtime environment. To be compatible with https://github.com/opencontainers/runtime-spec/blob/main/config.md#summary, I would suggest that we creat a table somewhere to document how we deal with different hooks (remember that we have guest hooks too.), something like the following table:

|Hook Name|Hook Type | Hook Path|Exec Place | Exec Time |
|---|---|---|---|---|
|Prestart (deprecated)| OCI hook |host runtime namespace | host runtime namespace | After VM is started, before container is created |
|CreateRuntime| OCI hook | host runtime namespace | host runtime namespace | After VM is started, before container is created, after prestart hooks |
|CreateContainer| OCI hook | host runtime namespace | host vmm namespace | After VM is started, before container is created, after createruntime hooks|
|StartContainer| OCI hook | guest container namespace | guest container namespace | After container is created, before container is started| 
|Poststart| OCI hook | host runtime namespace | host runtime namespace | After container is started, before start operation returns|
|Poststop| OCI hook | host runtime namespace | host runtime namespace | After container is deleted, before delete operation returns|
|guest Prestart| Kata hook | guest agent namespace | guest agent namespace | During start operation, before the container command is executed|
|guest Poststart| Kata hook | guest agent namespace | guest agent namespace | During start operation, after the container command is executed, before start operation returns|
|guest Poststop| Kata hook | guest agent namespace | guest agent namespace | During delete operation, after the container command is deleted, before delete operation returns|

One design choice that we need to make is that for CreateContainer hooks, OCI requires to run them inside the container namespace but the hook executable path is in the host runtime, which is a non-starter for VM-based containers. So I propose that we keep running it in the host runtime namespace. Other OCI hooks are found and run in proper namespace as required by the OCI runtime.

[...]

PR #5788, opened 2022-11-29
runtime-rs: add oci hook support

YushuoEdge:
In this pr, we are going to support OCI Hooks in runtime-rs. Besides `prestart` / `poststart` / `poststop` / `create_runtime hooks`, which are supported in runtime-go, we also support `create_container` / `start_container` hooks in this pr.

Description of  OCI hooks in OCI spec: https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle

Fixes: #5787

### DIFF
--- a/docs/design/hooks-handling.md
+++ b/docs/design/hooks-handling.md
@@ -49,7 +49,7 @@ The table below summarized when and where those different hooks will be executed
 |---|---|---|---|---|
 | `Prestart(deprecated)` | OCI hook | host runtime namespace | host runtime namespace | After VM is started, before container is created. |
 | `CreateRuntime` | OCI hook | host runtime namespace | host runtime namespace | After VM is started, before container is created, after `Prestart` hooks. |
-| `CreateContainer` | OCI hook | host runtime namespace | host vmm namespace | After VM is started, before container is created, after `CreateRuntime` hooks. |
+| `CreateContainer` | OCI hook | host runtime namespace | host vmm namespace* | After VM is started, before container is created, after `CreateRuntime` hooks. |
 | `StartContainer` | OCI hook | guest container namespace | guest container namespace | After container is created, before container is started. |
 | `Poststart` | OCI hook | host runtime namespace | host runtime namespace | After container is started, before start operation returns. |
 | `Poststop` | OCI hook | host runtime namespace | host runtime namespace | After container is deleted, before delete operation returns. |
@@ -59,4 +59,5 @@ The table below summarized when and where those different hooks will be executed
 
 + `Hook Path` specifies where hook's path be resolved.
 + `Exec Place` specifies in which namespace those hooks can be executed.
+  + For `CreateContainer` Hooks, OCI requires to run them inside the container namespace while the hook executable path is in the host runtime, which is a non-starter for VM-based containers. So we design to keep them running in the *host vmm namespace.* 
 + `Exec Time` specifies at which time point those hooks can be executed.
--- a/docs/design/hooks-handling.md
+++ b/docs/design/hooks-handling.md
@@ -1,0 +1,62 @@
+# Kata Containers support for `Hooks`
+
+## Introduction
+
+During container's lifecycle, different Hooks can be executed to do custom actions. In Kata Containers, we support two types of Hooks, `OCI Hooks` and `Kata Hooks`.
+
+### OCI Hooks
+
+The OCI Spec stipulates six hooks that can be executed at different time points and namespaces, including `Prestart Hooks`, `CreateRuntime Hooks`, `CreateContainer Hooks`, `StartContainer Hooks`, `Poststart Hooks` and `Poststop Hooks`. We support these types of Hooks as compatible as possible in Kata Containers.
+
+The path and arguments of these hooks will be passed to Kata for execution via `bundle/config.json`. For example:
+```
+...
+"hooks": {
+  "prestart": [
+    {
+      "path": "/usr/bin/prestart-hook",
+      "args": ["prestart-hook", "arg1", "arg2"],
+      "env":  [ "key1=value1"]
+    }
+  ],
+  "createRuntime": [
+    {
+      "path": "/usr/bin/createRuntime-hook",
+      "args": ["createRuntime-hook", "arg1", "arg2"],
+      "env":  [ "key1=value1"]
+    }
+  ]
+}
+...
+```
+
+### Kata Hooks
+
+In Kata, we support another three kinds of hooks executed in guest VM, including `Guest Prestart Hook`, `Guest Poststart Hook`, `Guest Poststop Hook`.
+
+The executable files for Kata Hooks must be packaged in the *guest rootfs*. The file path to those guest hooks should be specified in the configuration file, and guest hooks must be stored in a subdirectory of `guest_hook_path` according to their hook type. For example:
+
++ In configuration file:
+```
+guest_hook_path="/usr/share/hooks"
+```
++ In guest rootfs, prestart-hook is stored in `/usr/share/hooks/prestart/prestart-hook`.
+
+## Execution
+The table below summarized when and where those different hooks will be executed in Kata Containers:
+
+| Hook Name | Hook Type | Hook Path | Exec Place | Exec Time |
+|---|---|---|---|---|
+| `Prestart(deprecated)` | OCI hook | host runtime namespace | host runtime namespace | After VM is started, before container is created. |
+| `CreateRuntime` | OCI hook | host runtime namespace | host runtime namespace | After VM is started, before container is created, after `Prestart` hooks. |
+| `CreateContainer` | OCI hook | host runtime namespace | host vmm namespace | After VM is started, before container is created, after `CreateRuntime` hooks. |
+| `StartContainer` | OCI hook | guest container namespace | guest container namespace | After container is created, before container is started. |
+| `Poststart` | OCI hook | host runtime namespace | host runtime namespace | After container is started, before start operation returns. |
+| `Poststop` | OCI hook | host runtime namespace | host runtime namespace | After container is deleted, before delete operation returns. |
+| `Guest Prestart` | Kata hook | guest agent namespace | guest agent namespace | During start operation, before container command is executed. |
+| `Guest Poststart` | Kata hook | guest agent namespace | guest agent namespace | During start operation, after container command is executed, before start operation returns. |
+| `Guest Poststop` | Kata hook | guest agent namespace | guest agent namespace | During delete operation, after container is deleted, before delete operation returns. |
+
++ `Hook Path` specifies where hook's path be resolved.
++ `Exec Place` specifies in which namespace those hooks can be executed.
++ `Exec Time` specifies at which time point those hooks can be executed.
--- a/src/agent/rustjail/src/lib.rs
+++ b/src/agent/rustjail/src/lib.rs
@@ -835,6 +835,45 @@ mod tests {
                         Timeout: 10,
                         ..Default::default()
                     }])),
+                    CreateRuntime: protobuf::RepeatedField::from(Vec::from([grpc::Hook {
+                        Path: String::from("createruntimepath"),
+                        Args: protobuf::RepeatedField::from(Vec::from([
+                            String::from("arg1"),
+                            String::from("arg2"),
+                        ])),
+                        Env: protobuf::RepeatedField::from(Vec::from([
+                            String::from("env1"),
+                            String::from("env2"),
+                        ])),
+                        Timeout: 10,
+                        ..Default::default()
+                    }])),
+                    CreateContainer: protobuf::RepeatedField::from(Vec::from([grpc::Hook {
+                        Path: String::from("createcontainerpath"),
+                        Args: protobuf::RepeatedField::from(Vec::from([
+                            String::from("arg1"),
+                            String::from("arg2"),
+                        ])),
+                        Env: protobuf::RepeatedField::from(Vec::from([
+                            String::from("env1"),
+                            String::from("env2"),
+                        ])),
+                        Timeout: 10,
+                        ..Default::default()
+                    }])),
+                    StartContainer: protobuf::RepeatedField::from(Vec::from([grpc::Hook {
+                        Path: String::from("startcontainerpath"),
+                        Args: protobuf::RepeatedField::from(Vec::from([
+                            String::from("arg1"),
+                            String::from("arg2"),
+                        ])),
+                        Env: protobuf::RepeatedField::from(Vec::from([
+                            String::from("env1"),
+                            String::from("env2"),
+                        ])),
+                        Timeout: 10,
+                        ..Default::default()
+                    }])),
                     ..Default::default()
                 },
                 result: oci::Hooks {
@@ -864,7 +903,24 @@ mod tests {
                         env: Vec::from([String::from("env1"), String::from("env2")]),
                         timeout: Some(10),
                     }]),
-                    ..Default::default()
+                    create_runtime: Vec::from([oci::Hook {
+                        path: String::from("createruntimepath"),
+                        args: Vec::from([String::from("arg1"), String::from("arg2")]),
+                        env: Vec::from([String::from("env1"), String::from("env2")]),
+                        timeout: Some(10),
+                    }]),
+                    create_container: Vec::from([oci::Hook {
+                        path: String::from("createcontainerpath"),
+                        args: Vec::from([String::from("arg1"), String::from("arg2")]),
+                        env: Vec::from([String::from("env1"), String::from("env2")]),
+                        timeout: Some(10),
+                    }]),
+                    start_container: Vec::from([oci::Hook {
+                        path: String::from("startcontainerpath"),
+                        args: Vec::from([String::from("arg1"), String::from("arg2")]),
+                        env: Vec::from([String::from("env1"), String::from("env2")]),
+                        timeout: Some(10),
+                    }]),
                 },
             },
             TestData {
@@ -897,6 +953,45 @@ mod tests {
                         Timeout: 10,
                         ..Default::default()
                     }])),
+                    CreateRuntime: protobuf::RepeatedField::from(Vec::from([grpc::Hook {
+                        Path: String::from("createruntimepath"),
+                        Args: protobuf::RepeatedField::from(Vec::from([
+                            String::from("arg1"),
+                            String::from("arg2"),
+                        ])),
+                        Env: protobuf::RepeatedField::from(Vec::from([
+                            String::from("env1"),
+                            String::from("env2"),
+                        ])),
+                        Timeout: 10,
+                        ..Default::default()
+                    }])),
+                    CreateContainer: protobuf::RepeatedField::from(Vec::from([grpc::Hook {
+                        Path: String::from("createcontainerpath"),
+                        Args: protobuf::RepeatedField::from(Vec::from([
+                            String::from("arg1"),
+                            String::from("arg2"),
+                        ])),
+                        Env: protobuf::RepeatedField::from(Vec::from([
+                            String::from("env1"),
+                            String::from("env2"),
+                        ])),
+                        Timeout: 10,
+                        ..Default::default()
+                    }])),
+                    StartContainer: protobuf::RepeatedField::from(Vec::from([grpc::Hook {
+                        Path: String::from("startcontainerpath"),
+                        Args: protobuf::RepeatedField::from(Vec::from([
+                            String::from("arg1"),
+                            String::from("arg2"),
+                        ])),
+                        Env: protobuf::RepeatedField::from(Vec::from([
+                            String::from("env1"),
+                            String::from("env2"),
+                        ])),
+                        Timeout: 10,
+                        ..Default::default()
+                    }])),
                     ..Default::default()
                 },
                 result: oci::Hooks {
@@ -913,7 +1008,24 @@ mod tests {
                         env: Vec::from([String::from("env1"), String::from("env2")]),
                         timeout: Some(10),
                     }]),
-                    ..Default::default()
+                    create_runtime: Vec::from([oci::Hook {
+                        path: String::from("createruntimepath"),
+                        args: Vec::from([String::from("arg1"), String::from("arg2")]),
+                        env: Vec::from([String::from("env1"), String::from("env2")]),
+                        timeout: Some(10),
+                    }]),
+                    create_container: Vec::from([oci::Hook {
+                        path: String::from("createcontainerpath"),
+                        args: Vec::from([String::from("arg1"), String::from("arg2")]),
+                        env: Vec::from([String::from("env1"), String::from("env2")]),
+                        timeout: Some(10),
+                    }]),
+                    start_container: Vec::from([oci::Hook {
+                        path: String::from("startcontainerpath"),
+                        args: Vec::from([String::from("arg1"), String::from("arg2")]),
+                        env: Vec::from([String::from("env1"), String::from("env2")]),
+                        timeout: Some(10),
+                    }]),
                 },
             },
         ];
--- a/src/agent/rustjail/src/lib.rs
+++ b/src/agent/rustjail/src/lib.rs
@@ -156,6 +156,8 @@ fn hooks_grpc_to_oci(h: &grpc::Hooks) -> oci::Hooks {
 
     let create_runtime = hook_grpc_to_oci(h.CreateRuntime.as_ref());
 
+    let create_container = hook_grpc_to_oci(h.CreateContainer.as_ref());
+
     let poststart = hook_grpc_to_oci(h.Poststart.as_ref());
 
     let poststop = hook_grpc_to_oci(h.Poststop.as_ref());
@@ -163,6 +165,7 @@ fn hooks_grpc_to_oci(h: &grpc::Hooks) -> oci::Hooks {
     oci::Hooks {
         prestart,
         create_runtime,
+        create_container,
         poststart,
         poststop,
     }
--- a/src/agent/rustjail/src/lib.rs
+++ b/src/agent/rustjail/src/lib.rs
@@ -154,12 +154,15 @@ fn hook_grpc_to_oci(h: &[grpcHook]) -> Vec<oci::Hook> {
 fn hooks_grpc_to_oci(h: &grpc::Hooks) -> oci::Hooks {
     let prestart = hook_grpc_to_oci(h.Prestart.as_ref());
 
+    let create_runtime = hook_grpc_to_oci(h.CreateRuntime.as_ref());
+
     let poststart = hook_grpc_to_oci(h.Poststart.as_ref());
 
     let poststop = hook_grpc_to_oci(h.Poststop.as_ref());
 
     oci::Hooks {
         prestart,
+        create_runtime,
         poststart,
         poststop,
     }
@@ -860,6 +863,7 @@ mod tests {
                         env: Vec::from([String::from("env1"), String::from("env2")]),
                         timeout: Some(10),
                     }]),
+                    ..Default::default()
                 },
             },
             TestData {
@@ -908,6 +912,7 @@ mod tests {
                         env: Vec::from([String::from("env1"), String::from("env2")]),
                         timeout: Some(10),
                     }]),
+                    ..Default::default()
                 },
             },
         ];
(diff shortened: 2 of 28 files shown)


src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+        .stderr(Stdio::piped())
+        .spawn()?;
+
+    let state = serde_json::to_string(st)?;
+    let join_handle = tokio::spawn(async move {
+        // write state to stdin of child process
+        if let Some(mut stdin) = child.stdin.take() {
+            match stdin.write_all(state.as_bytes()).await {
+                Ok(_) => {}
+                Err(e) => {
```
Tim-0731-Hzt, line 101: Should this return an error?

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+    PostStart,
+    PostStop,
+}
+
+// execute hooks in oci::Hooks according to hook types
+pub async fn run_hooks(spec: &oci::Spec, st: &oci::State, hk_type: HookType) -> Result<()> {
+    if let Some(spec_hooks) = &spec.hooks {
+        let hooks = match hk_type {
+            HookType::PreStart => {
+                if !spec_hooks.prestart.is_empty() {
```
liubin, line 27: I think this log and belows have no meaning, so delete them?

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+                    .ok_or_else(|| anyhow!("hook exit status has no code"))?;
+                if exit_code != 0 {
+                    error!(
+                        sl!(),
+                        "hook {} exit code: {}, error message: {}",
+                        &binary_path.display(),
+                        exit_code,
+                        err
+                    );
+                    return Err(anyhow!(nix::Error::UnknownErrno));
```
liubin, line 146: Is it better to return more detailed error like lines 152-156?

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+
+// check and reserve valid environment variables
+// invalid env var may cause panic, refer to https://doc.rust-lang.org/std/env/fn.set_var.html#panics
+// key should not:
+//  * contain NUL character '\0'
+//  * contain ASCII equal sign '='
+//  * be empty
+// value should not:
+//  * contain NUL character '\0'
+fn valid_env(e: &str) -> Option<(&str, &str)> {
```
liubin, line 174: Agent has the same code, and should use a common function laid in `src/libs`.

But you can leave it in another PR.

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -111,6 +117,12 @@ impl RuntimeHandlerManagerInner {
             .await
             .context("init runtime handler")?;
 
+        // CreateRuntime Hooks:
+        // * should be run in runtime namespace
+        // * should be run when creating the runtime
+        // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#createruntime-hooks
```
liubin, line 123: Don't use `main`/`master` in GitHub URL, use tag/commit instead.

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -205,7 +218,22 @@ impl RuntimeHandlerManager {
             let bundler_path = format!("{}/{}", req.bundle, oci::OCI_SPEC_CONFIG_FILE_NAME);
             let spec = oci::Spec::load(&bundler_path).context("load spec")?;
 
-            self.try_init_runtime_instance(&spec, &req.options)
+            // Prestart Hooks [DEPRECATED in newest oci spec]:
+            // * should be run in runtime namespace
+            // * should be run after "start", but if Prestart Hook and CreateRuntime Hook are both supported
+            //   it should be run before CreateRuntime Hook
+            // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#prestart
```
liubin, line 225: Ditto

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -89,6 +91,22 @@ impl ContainerManager for VirtContainerManager {
                 let c = containers
                     .remove(container_id)
                     .ok_or_else(|| Error::ContainerNotFound(container_id.to_string()))?;
+
+                // Poststop Hooks:
+                // * should be run in runtime namespace
+                // * should be run after the container is deleted but before delete operation returns
+                // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#poststop
```
liubin, line 98: Ditto

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        // read info from stdout of child process
+        if let Some(stdout) = child.stdout.as_mut() {
+            let mut out = String::new();
+            match stdout.read_to_string(&mut out).await {
+                Ok(_) => {
+                    info!(sl!(), "child stdout: {:?}", out.as_str());
+                }
+                Err(e) => {
```
Tim-0731-Hzt, line 114: same as above

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -192,6 +210,22 @@ impl ContainerManager for VirtContainerManager {
             .get(container_id)
             .ok_or_else(|| Error::ContainerNotFound(container_id.clone()))?;
         c.start(process).await.context("start")?;
+
+        // Poststart Hooks:
+        // * should be run in runtime namespace
+        // * should be run after user-specific command is executed but before start operation returns
+        // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#poststart
```
liubin, line 217: Ditto

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -89,6 +91,22 @@ impl ContainerManager for VirtContainerManager {
...
                     .remove(container_id)
                     .ok_or_else(|| Error::ContainerNotFound(container_id.to_string()))?;
+
+                // Poststop Hooks:
+                // * should be run in runtime namespace
+                // * should be run after the container is deleted but before delete operation returns
+                // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#poststop
+                let c_spec = c.spec().await;
+                let state = oci::State {
+                    version: String::from("ociVersion"),
```
liubin, line 101: The `version` should a real version number, but not the `ociVersion` itself.

```json
    "state": {
        "ociVersion": "0.2.0",
        "id": "oci-container1",
        "status": "creating",
        "pid": 4422,
        "bundle": "/containers/redis",
        "annotations": {
            "myKey": "myValue"
        }
    }
```

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -192,6 +210,22 @@ impl ContainerManager for VirtContainerManager {
...
             .ok_or_else(|| Error::ContainerNotFound(container_id.clone()))?;
         c.start(process).await.context("start")?;
+
+        // Poststart Hooks:
+        // * should be run in runtime namespace
+        // * should be run after user-specific command is executed but before start operation returns
+        // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#poststart
+        let c_spec = c.spec().await;
+        let state = oci::State {
+            version: String::from("ociVersion"),
```
liubin, line 220: Ditto

src/runtime-rs/crates/runtimes/common/Cargo.toml
```diff
@@ -26,3 +26,4 @@ agent = { path = "../../agent" }
 kata-sys-util = { path = "../../../../libs/kata-sys-util" }
 kata-types = { path = "../../../../libs/kata-types" }
 oci = { path = "../../../../libs/oci" }
+logging = { path = "../../../../libs/logging"}
```
Tim-0731-Hzt, line 29: add a new line after this

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
YushuoEdge: We might need some log to indicate that the hook is running?

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+            Err(e) => Err(anyhow!(
+                "wait child process error: {} {}",
+                e,
+                e.raw_os_error().unwrap()
+            )),
+        }
+    });
+
+    match tokio::time::timeout(execution_timeout, join_handle).await {
+        Ok(r) => r.unwrap(),
```
Tim-0731-Hzt, line 161: how about make it to `Ok(r) => r.context("context")?`

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
YushuoEdge: The problem here is that there may be many kinds of errors, and it is difficult for us to determine which one.

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+            HookType::PostStop => {
+                if !spec_hooks.poststop.is_empty() {
+                    info!(sl!(), "poststop hook");
+                }
+                &spec_hooks.poststop
+            }
+        };
+
+        for h in hooks.iter() {
+            run_hook(h, st).await?;
```
Tim-0731-Hzt, line 53: suggest adding `.context()`

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    Ok(())
+}
+
+// execute a specific hook, regardless of hook type
+pub async fn run_hook(h: &oci::Hook, st: &oci::State) -> Result<()> {
+    info!(sl!(), "runtime action: execute hook");
+
+    // check whether the binary file called by the hook exists
+    let binary_path = PathBuf::from(h.path.as_str()).canonicalize()?;
```
Tim-0731-Hzt, line 65: suggest adding `.context()`

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -89,6 +91,22 @@ impl ContainerManager for VirtContainerManager {
...
+                let c_spec = c.spec().await;
+                let state = oci::State {
+                    version: String::from("ociVersion"),
+                    id: c.container_id.to_string(),
+                    status: oci::ContainerState::Stopped,
+                    pid: c.pid().await as i32,
+                    bundle: c.config().await.bundle.clone(),
+                    annotations: c_spec.annotations.clone(),
+                };
+                run_hooks(&c_spec, &state, HookType::PostStop).await?;
```
Tim-0731-Hzt, line 108: suggest adding `.context("run post stop")`

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
YushuoEdge: We probably don't need to return error here. I think it is best not to hinder the continued operation of the runtime because of the execution of the hook. But the log type here may indeed need to be changed.

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
Ji-Xinyou: The exitcode should be a Kernel Errno, you can convert that to Error. I think libc or nix can do that.

YushuoEdge: I'm not sure whether it's a Kernel Errno at all here, and I need sometime to confirm ...

src/runtime-rs/crates/runtimes/common/src/runtimes_utils/host_hooks.rs
YushuoEdge: I have found that there is also some code implementing `execute_hooks` in `src/libs`:
https://github.com/kata-containers/kata-containers/blob/main/src/libs/kata-sys-util/src/hooks.rs

Maybe I will use that code to refactor this PR. Or before that, also refactor the agent's hook execution with the same code (:

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -192,6 +210,22 @@ impl ContainerManager for VirtContainerManager {
...
+        // Poststart Hooks:
+        // * should be run in runtime namespace
+        // * should be run after user-specific command is executed but before start operation returns
+        // * spec details: https://github.com/opencontainers/runtime-spec/blob/main/config.md#poststart
+        let c_spec = c.spec().await;
+        let state = oci::State {
+            version: String::from("ociVersion"),
+            id: c.container_id.to_string(),
+            status: oci::ContainerState::Running,
+            pid: c.pid().await as i32,
```
bergwolf, line 223: As per https://github.com/kata-containers/kata-containers/issues/5321, docker relies on the pid of the container to find out the container namespace. So please let's set state.pid to the vmm thread id as the vmm is the only process that is inside the container netns on the host.

In the meantime, it means that we need to run the prestart hooks after the vmm is launched.

bergwolf:
I would like to ask for the feature to be discussed and design decisions to be made before merging the PR. See  https://github.com/kata-containers/kata-containers/issues/5787#issuecomment-1331748449

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
YushuoEdge: Does it mean that state.pid of all OCI hooks should be set to vmm thread/process id?

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -192,6 +210,22 @@ impl ContainerManager for VirtContainerManager {
             .get(container_id)
             .ok_or_else(|| Error::ContainerNotFound(container_id.clone()))?;
         c.start(process).await.context("start")?;
+
+        // Poststart Hooks:
+        // * should be run in runtime namespace
```
Tim-0731-Hzt, line 215: IIUC, runtime namespace means container namespace, do you need to enter the network namespace?  like `netns::NetnsGuard::new` Do I missing something?

YushuoEdge: According to the spec, `runtime namespace` is not `container namespace`.

Tim-0731-Hzt, line 215: IIUC, runtime namespace is container namespace, should enter network namespace?

Tim-0731-Hzt: Thanks @YushuoEdge , I see that

bergwolf:
@YushuoEdge Thanks for renewing the PR! Could you please add a design doc in the `docs` directory to explain the hook handling logic? The table in https://github.com/kata-containers/kata-containers/issues/5787#issuecomment-1331748449 can be included there as well.

YushuoEdge:
@bergwolf The design document has been added, PTAL！

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
YushuoEdge: Updated!

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -151,6 +190,16 @@ impl Sandbox for VirtSandbox {
...
 
+        // execute pre-start hook functions, including Prestart Hooks and CreateRuntime Hooks
+        let (prestart_hooks, create_runtime_hooks) = match spec.hooks.as_ref() {
+            Some(hooks) => (hooks.prestart.clone(), hooks.create_runtime.clone()),
+            None => (Vec::new(), Vec::new()),
+        };
+        self.execute_prestart_hooks(&prestart_hooks, &create_runtime_hooks, state)
+            .await?;
+
+        // TODO: if prestart_hooks is empty, rescan the network endpoints
```
liubin, line 201: Why leave this comment?

YushuoEdge: Sorry, it should be "is not". I will fix it.
This comment is related to https://github.com/kata-containers/kata-containers/pull/5972. 
As networks may be configured through `Prestart Hook`, so we should rescan the network if `prestart_hook` is not empty. Hotplug network endpoints has not supported yet and the rescan operation relies on that capability,  so I leave this comment as a TODO action.

docs/design/hooks-handling.md
```diff
@@ -0,0 +1,62 @@
...
+guest_hook_path="/usr/share/hooks"
+```
++ In guest rootfs, prestart-hook is stored in `/usr/share/hooks/prestart/prestart-hook`.
+
+## Execution
+The table below summarized when and where those different hooks will be executed in Kata Containers:
+
+| Hook Name | Hook Type | Hook Path | Exec Place | Exec Time |
+|---|---|---|---|---|
+| `Prestart(deprecated)` | OCI hook | host runtime namespace | host runtime namespace | After VM is started, before container is created. |
```
Tim-0731-Hzt, line 50: I am confused about this, if it is deprecated, why do we still implement it?

liubin: Users may still use them.

bergwolf: It is deprecated by the spec, but there are still many users so we have to support it.

docs/design/hooks-handling.md
```diff
@@ -0,0 +1,62 @@
...
+| `StartContainer` | OCI hook | guest container namespace | guest container namespace | After container is created, before container is started. |
+| `Poststart` | OCI hook | host runtime namespace | host runtime namespace | After container is started, before start operation returns. |
+| `Poststop` | OCI hook | host runtime namespace | host runtime namespace | After container is deleted, before delete operation returns. |
+| `Guest Prestart` | Kata hook | guest agent namespace | guest agent namespace | During start operation, before container command is executed. |
+| `Guest Poststart` | Kata hook | guest agent namespace | guest agent namespace | During start operation, after container command is executed, before start operation returns. |
+| `Guest Poststop` | Kata hook | guest agent namespace | guest agent namespace | During delete operation, after container is deleted, before delete operation returns. |
+
++ `Hook Path` specifies where hook's path be resolved.
++ `Exec Place` specifies in which namespace those hooks can be executed.
++ `Exec Time` specifies at which time point those hooks can be executed.
```
bergwolf, line 62: Please explain the design choice of  the `CreateContainer` hooks as in https://github.com/kata-containers/kata-containers/issues/5787#issuecomment-1331748449 too. We need to clearly document the inconsistent OCI-hook behavior that we choose to implement.

> One design choice that we need to make is that for CreateContainer hooks, OCI requires to run them inside the container namespace but the hook executable path is in the host runtime, which is a non-starter for VM-based containers. So I propose that we keep running it in the host vmm namespace. Other OCI hooks are found and run in the proper namespace as required by the OCI runtime.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -119,11 +120,49 @@ impl VirtSandbox {
 
         Ok(resource_configs)
     }
+
+    async fn execute_prestart_hooks(
```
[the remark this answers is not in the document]
YushuoEdge: I wrote this function just for binding this two kinds of hooks like `PrestartHookFunc` in runtime-go.  `prestart_hooks` and `create_runtime_hooks` refers to two specific kinds of hooks, while `execute_prestart_hooks` refers to execution time point.
But I agree that this name should be changed for understanding (:

docs/design/hooks-handling.md
YushuoEdge: update!

src/libs/protocols/protos/oci.proto
```diff
@@ -161,11 +161,20 @@ message Hooks {
...
+	// StartContainer is a list of hooks to be run after container is created, but before it is started.
+	repeated Hook StartContainer = 4  [(gogoproto.nullable) = false]; 
+
 	// Poststart is a list of hooks to be run after the container process is started.
-	repeated Hook Poststart = 2  [(gogoproto.nullable) = false];
+	repeated Hook Poststart = 5  [(gogoproto.nullable) = false];
 
 	// Poststop is a list of hooks to be run after the container process exits.
-	repeated Hook Poststop = 3  [(gogoproto.nullable) = false];
+	repeated Hook Poststop = 6  [(gogoproto.nullable) = false];
```
bergwolf, line 177: When editing agent protocols, please do not insert things in the middle of a struct. It will break agent protocol compatibility (which is already in a bad state.)

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -57,12 +63,37 @@ impl ContainerManager for VirtContainerManager {
...
             self.agent.clone(),
             self.resource_manager.clone(),
         )
         .context("new container")?;
 
+        // CreateContainer Hooks:
+        // * should be run in vmm namespace (hook path in runtime namespace)
+        // * should be run after the vm is started, before container is created, and after CreateRuntime Hooks
+        // * spec details: https://github.com/opencontainers/runtime-spec/blob/c1662686cff159595277b79322d0272f5182941b/config.md#createcontainer-hooks
+        let vmm_master_tid = self.hypervisor.get_vmm_master_tid().await?;
```
bergwolf, line 77: Please add a helper function to run the hook so that we don't have to maintain the same piece of code at different places.

YushuoEdge: I used to think the `hookstates.execute_hooks(...)` is already the helper function, since we should construct different oci::State for different kinds of hooks. I wonder, do we need to consider the hook type in the helper function? By that way, we should construct different oci::State according the hook type in that function?

src/libs/protocols/protos/oci.proto
YushuoEdge: OK! I will fix it ASAP.

bergwolf (approved):
Thanks a lot!

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -149,6 +189,16 @@ impl Sandbox for VirtSandbox {
...
 
+        // execute pre-start hook functions, including Prestart Hooks and CreateRuntime Hooks
+        let (prestart_hooks, create_runtime_hooks) = match spec.hooks.as_ref() {
+            Some(hooks) => (hooks.prestart.clone(), hooks.create_runtime.clone()),
+            None => (Vec::new(), Vec::new()),
+        };
+        self.execute_oci_hook_functions(&prestart_hooks, &create_runtime_hooks, state)
+            .await?;
+
+        // TODO: if prestart_hooks is not empty, rescan the network endpoints
```
Tim-0731-Hzt, line 200: Please raise an issue for this

Tim-0731-Hzt (approved):
LGTM

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
YushuoEdge: Thanks! Fixed.

src/agent/rustjail/src/lib.rs
```diff
@@ -154,12 +154,15 @@ fn hook_grpc_to_oci(h: &[grpcHook]) -> Vec<oci::Hook> {
...
     let prestart = hook_grpc_to_oci(h.Prestart.as_ref());
 
+    let create_runtime = hook_grpc_to_oci(h.CreateRuntime.as_ref());
+
     let poststart = hook_grpc_to_oci(h.Poststart.as_ref());
 
     let poststop = hook_grpc_to_oci(h.Poststop.as_ref());
 
     oci::Hooks {
         prestart,
```
c3d, line 163: I'm a bit confused. The commit message and description seem wrong: they claim that you add support for OCI hooks, but the code apparently already had `prestart`, `poststart` and `poststop`. Sho

YushuoEdge: Thanks for reviewing!

The description of this pr is actually outdated, I am sorry for that have made you confused and I have update the description.  `prestart hooks`, `poststart hooks`, `poststop hooks` are those hooks that have been supported in runtime-go, so I guess this is the reason why the code already have `prestart`, `poststart`, and `poststop`. And `create_runtime hooks` will not be executed in the guest, so I guess this is the reason why it does not appear in the original code.

But actually, `prestart`, `poststart` and `poststop` *OCI hooks* will not execute in the guest either, the reason I tend to keep them here is for better log output and for future changes. WDYT?

I hope my explanation can make you clean😆

c3d: I think that the new wording is much better, thank you @YushuoEdge

docs/design/hooks-handling.md
```diff
@@ -0,0 +1,62 @@
+# Kata Containers support for `Hooks`
```
c3d, line 1: If you do a new commit for any other reason, I would suggest reordering your commits to put this one first.

src/agent/rustjail/src/lib.rs
```diff
@@ -860,6 +864,7 @@ mod tests {
                         env: Vec::from([String::from("env1"), String::from("env2")]),
                         timeout: Some(10),
                     }]),
+                    ..Default::default()
```
c3d, line 867: I find the use of `default` here inconsistent with how it's done for other hooks. Could you either:

1. Expand the initialization like for the other hooks, or
2. Explain in a comment why defaults are OK for the newly created hooks.

src/libs/protocols/src/trans.rs
```diff
@@ -974,6 +977,18 @@ impl From<crate::oci::Hooks> for oci::Hooks {
         for hook in from.take_Prestart().to_vec() {
             prestart.push(hook.into())
         }
+        let mut create_runtime = Vec::new();
```
c3d, line 980: While this is similar to the code that existed before, this is not super idiomatic. You could use something like:

```
        let create_container = from
            .take_CreateContainer()
            .into_iter()
            .map(|i| i.into())
            .collect();
```

This dispense with the need to make things `mut`.

docs/design/hooks-handling.md
YushuoEdge: Update!

src/agent/rustjail/src/lib.rs
YushuoEdge: I have expanded the initialization!

src/libs/protocols/src/trans.rs
YushuoEdge: Great suggestion! Thanks!

studychao (approved):
Really nice work. Thanks @YushuoEdge

PR #5803, opened 2022-11-30
CC | Add annotations for SEV/kbs controls at the pod level (DONOTMERGE)

### LINKED ISSUE #5782: CC | Enable CoCo configuration changes through Kubernetes annotations
The user needs controls to configure CoCo pods on a case-by-case basis. This should include k8s annotations to enable `confidential_guest`, `guest_pre_attestation`, and to set the SEV(-ES) variables such as KBS URI, MODE, POLICY, and KEYSET. 

CC @dubek @fitzthum @Alex-Carter01

jimcadden:
Fixes #5782 

Depends: #5665


jimcadden:
cc. @dubek @fitzthum @ryansavino @Alex-Carter01

fidencio:
Hey @jimcadden, just a FYI, as this PR brings in https://github.com/kata-containers/kata-containers/pull/5665, and that PR will have to be rebased due to the conflict, I think you could have the second patch here as part of that one, so we can avoid one round of rebase-merge, rebase-merge.

Up to you to take whatever approach you prefer.

src/runtime/virtcontainers/pkg/annotations/annotations.go
```diff
@@ -24,6 +24,25 @@ const (
 	SandboxConfigPathKey = kataAnnotationsPrefix + "config_path"
 )
 
+// Annotations related to Confidential Containers (CoCo)
+const (
+	cocoAnnotationsPrefix      = "io.confidentialcontainers."
```
fidencio, line 29: As this is / will be part of the default Kata Containers configuration at some point, I'd prefer using the "io.katacontainers.*" prefix.  With this said, you don't have to add this new block for CoCo, but just increase / expand the one for kata, adding the `kataAnnotPreAttestationPrefix` and the  `kataAnnotSevPrefix`, and then keeping the PreAttestation / GuestPolicy in this const block here.

fidencio: Also, if we already have some tests for specific annotations (please, check the `annotations_tests.go` file, it'd be good to add some tests there to ensure we don't end up regressing on this at some point.

jimcadden:
Thanks @fidencio 

> I think you could have the second patch here as part of that one, so we can avoid one round of rebase-merge, rebase-merge.

Makes sense. I will make the change prefix/test changes mentioned above and push the updated patch to #5665, and then close this PR.

src/runtime/virtcontainers/pkg/annotations/annotations.go
jimcadden: > ...to ensure these newly introduced annotations will be present when Kata Containers is deployed via the CC Operator.

@fidencio I believe only `io.katacontainers.hypervisor.*` annotations are require to be explicitly enabled. That is, if we did `io.katacontainers.cc.*` the CC annotation would be enabled by default. Am I mistaken?

fidencio: Not really.  By default all the annotations but that one in the Makefile are disabled.

jimcadden: That is what I thought as well. However, in doing this patch I was to test the new annotations without explicitly enabling them in the configuration file. I dug deeper and the only place I found where annotation filtering took place was here in [src/runtime/pkg/oci/untils.go:385](https://github.com/kata-containers/kata-containers/blob/8246de821ff327beb750b6239ac6c9cc6272f217/src/runtime/pkg/oci/utils.go#L385) (note the `KataAnnotationHypervisorPrefix` filter, which is what I refer to in my previous comment) . As far as I can tell, this code isn't executed with a launch via `kubectl apply -f `. In addition, I could not find another place where annotations are being filtered. I would expect to see something in `virtcontiners/sandbox.go` where the annotations are being applied. It's very possible I'm missing something here...

fidencio: Nice finding, that's a surprise for me.

So, that's good to know that we're fine and don't have to enable anything else.  Thanks for double-checking this @jimcadden!

jimcadden:
Closing this PR and adding the patch to #5665

PR #5805, opened 2022-11-30
add volume expand support for direct volume resize

### LINKED ISSUE #5804: Add volume expand support for direct volume
When direct-volume is introduced, only create and delete operation is implemented {https://github.com/kata-containers/kata-containers/issues/3454}. This issue is for adding the volume expand implementation to support volume resize in kata agent.

shuochen0311 is working on this issue.

shuochen0311:
agent: add volume expand support for direct volume resize

When direct volume resize happens, we should rescan the pci device metadata and expand the existing mount point inside kata agent. This change add the implementation for it.

Fixes #5804 

Signed-off-by: Shuo Chen <shuo.chen@databricks.com>

### DIFF
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -10,6 +10,7 @@ use std::io::{BufRead, BufReader, Write};
 use std::iter;
 use std::os::unix::fs::{MetadataExt, PermissionsExt};
 use std::path::Path;
+use std::process::Command;
 use std::str::FromStr;
 use std::sync::Arc;
 
@@ -48,6 +49,8 @@ const RW_MASK: u32 = 0o660;
 const RO_MASK: u32 = 0o440;
 const EXEC_MASK: u32 = 0o110;
 const MODE_SETGID: u32 = 0o2000;
+
+const MTAB_PATH: &str = "/etc/mtab";
 
 #[rustfmt::skip]
 lazy_static! {
@@ -1008,6 +1011,40 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
     fs::create_dir_all(dir).context(format!("create_dir_all {:?}", dir))?;
 
     fs::File::create(path).context(format!("create empty file {:?}", path))?;
+
+    Ok(())
+}
+
+/// Parse mtab and return the device which is mounted at a given directory
+pub fn get_mount_device(mount_dir: &Path) -> Result<String> {
+    let dir = mount_dir.to_string_lossy().into_owned();
+    let f = File::open(MTAB_PATH)?;
+    let reader = BufReader::new(f);
+
+    for line in reader.lines() {
+        let tem_line = line?;
+        let parts: Vec<&str> = tem_line.split_whitespace().collect();
+        if parts.contains(&dir.as_str()) && !parts.is_empty() {
+            return Ok(String::from(parts[0]));
+        }
+    }
+
+    return Err(anyhow!("can't find mount point {:?}", dir));
+}
+
+#[instrument]
+pub fn expand_file_system(mountpoint: &String, logger: Logger) -> Result<()> {
+    let path = Path::new(mountpoint);
+    match get_mount_device(path) {
+        Ok(device_path) => {
+            info!(logger, "resize to file system of device {:?}", device_path);
+            Command::new("resize2fs")
+                .arg(device_path)
+                .output()
+                .expect("failed to execute file system expand");
+        }
+        Err(e) => return Err(e),
+    };
 
     Ok(())
 }
--- a/src/agent/src/pci.rs
+++ b/src/agent/src/pci.rs
@@ -4,6 +4,7 @@
 //
 use std::convert::TryInto;
 use std::fmt;
+use std::fs;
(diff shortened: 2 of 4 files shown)


src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -262,8 +263,8 @@ func (s *service) startManagementServer(ctx context.Context, ociSpec *specs.Spec
 	m := http.NewServeMux()
 	m.Handle(MetricsUrl, http.HandlerFunc(s.serveMetrics))
 	m.Handle(AgentUrl, http.HandlerFunc(s.agentURL))
-	m.Handle(DirectVolumeStatUrl, http.HandlerFunc(s.serveVolumeStats))
-	m.Handle(DirectVolumeResizeUrl, http.HandlerFunc(s.serveVolumeResize))
+	m.Handle("/"+DirectVolumeStatUrl, http.HandlerFunc(s.serveVolumeStats))
+	m.Handle("/"+DirectVolumeResizeUrl, http.HandlerFunc(s.serveVolumeResize))
```
liubin, line 267: Can you move the `/` to `DirectVolumeStatUrl`? Like `IPTablesUrl`.

src/agent/src/mount.rs
```diff
@@ -1012,6 +1015,40 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
...
+    }
+
+    return Err(anyhow!("can't find mount point {:?}", dir));
+}
+
+#[instrument]
+pub fn expand_file_system(mountpoint: &String, logger: Logger) -> Result<()> {
+    let path = Path::new(mountpoint);
+    match get_mount_device(path) {
+        Ok(device_path) => {
```
liubin, line 1039: Since you are doing nothing in case of error, I think this will save a `match` statement:

```rust
let device_path = get_mount_device(path)?;
......
```

src/agent/src/mount.rs
```diff
@@ -1012,6 +1015,40 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
...
+            return Ok(String::from(parts[0]));
+        }
+    }
+
+    return Err(anyhow!("can't find mount point {:?}", dir));
+}
+
+#[instrument]
+pub fn expand_file_system(mountpoint: &String, logger: Logger) -> Result<()> {
+    let path = Path::new(mountpoint);
```
liubin, line 1037: `get_mount_device` accepts a `Path` parameter, here you convert a `String` to `Path` and in `get_mount_device` you convert it to `String` back, so why not just use a `String` or `&str` parameter for `get_mount_device` fn?

src/agent/src/mount.rs
```diff
@@ -1012,6 +1015,40 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
...
+/// Parse mtab and return the device which is mounted at a given directory
+pub fn get_mount_device(mount_dir: &Path) -> Result<String> {
+    let dir = mount_dir.to_string_lossy().into_owned();
+    let f = File::open(MTAB_PATH)?;
+    let reader = BufReader::new(f);
+
+    for line in reader.lines() {
+        let tem_line = line?;
+        let parts: Vec<&str> = tem_line.split_whitespace().collect();
+        if parts.contains(&dir.as_str()) && !parts.is_empty() {
```
liubin, line 1027: Here should use the exact field to check the `mount_dir`.

src/agent/src/pci.rs
```diff
@@ -182,6 +187,13 @@ impl FromStr for Path {
     }
 }
 
+pub fn rescan_pci_meta() -> anyhow::Result<()> {
+    match fs::write(PCI_RESCAN_FILE, PCI_RESCAN_SIGNAL) {
```
liubin, line 191: This can be simplified by
```rust
fs::write(PCI_RESCAN_FILE, PCI_RESCAN_SIGNAL)
```
You will format the error in `rpc.rs`, so here even the `anyhow!` is not needed.

src/agent/src/mount.rs
```diff
@@ -1012,6 +1015,40 @@ fn ensure_destination_file_exists(path: &Path) -> Result<()> {
...
 }
 
+/// Parse mtab and return the device which is mounted at a given directory
+pub fn get_mount_device(mount_dir: &Path) -> Result<String> {
+    let dir = mount_dir.to_string_lossy().into_owned();
+    let f = File::open(MTAB_PATH)?;
+    let reader = BufReader::new(f);
+
+    for line in reader.lines() {
+        let tem_line = line?;
```
amshinde, line 1025: Why do you need this assignment and esp the error check ? at the end

src/agent/src/mount.rs
amshinde: Yes, looks like this should be an exact match.

src/agent/src/pci.rs
amshinde: Agreed. Though this should be ```fs::write(PCI_RESCAN_FILE, PCI_RESCAN_SIGNAL)?;```

PR #5807, opened 2022-12-01
runtime: Add identification in version for runtime-rs

liubin:
Now we are supporting two runtime/shim, the go version, and the rust version, for debug purposes, we can
add an identification in the version info
to tell us which runtime/shim is used.

Fixes: #5906


studychao (approved):
Thanks @liubin , LGTM

snir911 (approved):

Issue #5806, opened 2022-12-01
Add identification in version for runtime-rs


liubin:
Now we are supporting two runtime/shim, the go version, and the rust version, for debug purposes, we can add an identification in the version info to tell us which runtime/shim is used.

PR #5815, opened 2022-12-01
runtime-rs:add identification in version for runtime-rs

studychao:
Since we now have 2 runtime, go runtime and rust runtime, and sometimes it's hard for identifying which version of runtime being used.

In order to support debug runtime, we add identification for runtime-rs version command (containerd-shim-kata-v2 --version) and an extra (rust) will be displayed in version when using rust runtime.

fixes:#5806


studychao:
emmm I realize this is not a good approach for modifying version, I will redo it.

Now I will first put this PR as a draft

studychao:
I'll close this since it's dup as #5807

PR #5821, opened 2022-12-02
fix exited container bundle not exist causing new container start failed

### LINKED ISSUE #5819: persist file left by previous container cause other container in the pod start failed
hey guys,
I met a problem in testing.
As we use persist.json file to store pod container information, if a container start failed but its info is not cleaned up in persist.json due to some unexpected circumstances, when we start other containers in the same pod, fetchContainers function will range all containers in the persist.json file and get their config.json from bundle path, as previous container start failed, the bundle path is not exist, which will cause other containers start failed.
image
So why we have to get all containers config.json even though it is not exist anymore? Is it possible to ingore the error if GetContainerSpec failed?

Vanient:
reason:If a previous container exited unexpectedly and bundle path is cleaned up(which is cleaned by docker or isulad rm) but persist.json still exists(which should be cleaned by kata), fetchSandbox will range all containers in pod persist.json to get config.json under bundle during new container in pod starting, it will return error as exited container bundle path not exist anymore, causing new container start failed. So we should ignore exited container bundle path not exist error during new one start.

fix #5819

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -2465,6 +2465,10 @@ func (s *Sandbox) fetchContainers(ctx context.Context) error {
 		// Add spec from bundle path
 		spec, err := compatoci.GetContainerSpec(contConfig.Annotations)
 		if err != nil {
+			if os.IsNotExist(err) {
+				s.Logger().Errorf("get container %s spec err: %v", contConfig.ID, err)
+				continue
+			}
 			return err
 		}
 		contConfig.CustomSpec = &spec


liubin:
Thank you @Vanient, you need to fix your commit message by referring to https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

liubin:
@Vanient there are still some commit message errors, please check the Details of "Commit Message Check".

Vanient:
> @Vanient there are still some commit message errors, please check the Details of "Commit Message Check".

@liubin Thanks, updated

Issue #5360, opened 2022-10-06
CC | Use cached build for the cc-tdx-td-shim-tarball when possible


fidencio:
One way to speed up the current CC CIs is by caching the artefacts that don't change frequently and simply download and use them whenever they're needed.

This issue is for using the cached tarball whenever it's possible, instead of building the component.

PR #5828, opened 2022-12-02
CC | Add support for caching firmwares

fidencio:
As we're already doing for some components, let's also add support for
caching firmwares. TD-Shim and TDVF are the ones supported for now.

Fixes: #5360, #5361


tools/packaging/static-build/cache_components.sh
```diff
@@ -70,7 +90,17 @@ Usage: $0 "[options]"
 	Options:
 		-c	Cloud hypervisor cache
 		-k	Kernel cache
+			* Can receive a TEE environnment variable value, valid values are:
+			  * tdx
+			  If no TEE environment is passed, QEMU is built without TEE support.
```
stevenhorsman, line 95:
```suggestion
			  If no TEE environment is passed, the kernel is built without TEE support.
```
I think this should be referencing the kernel, rather than QEMU here?

stevenhorsman (approved):
Apart from one question this LGTM. Thanks!

arronwy (approved):
Thanks a lot @fidencio ! , verified both TDX CI for cloud hypervisor and qemu use cached td-shim, TDVF  and works well!

Issue #5740, opened 2022-11-24
Use rust version of runtime to create busybox containers with high cpu usage.


chentt10:
I used the source compiled runtime-rs and only modified the image and kernel in the hypervisor.dragonball section of the default configuration `/usr/share/defaults/kata-containers/configuration.toml`. image and kernel are from the latest release.

# System-related information and create container commands
``` shell
root@ubuntu:~# cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.5 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
root@ubuntu:~# uname -a
Linux ubuntu 5.4.0-42-generic #46~18.04.1-Ubuntu SMP Fri Jul 10 07:21:24 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
root@ubuntu:~# ctr run --runtime io.containerd.run.kata.v2 -t --rm docker.io/library/busybox:latest hello sh
/ #
```

# After creating the container, check the cpu usage  
```
root@ubuntu:/opt/kata/bin# top 
top - 03:42:51 up 24 min,  0 users,  load average: 1.00, 0.73, 0.52
Tasks: 235 total,   1 running, 152 sleeping,   0 stopped,   0 zombie
%Cpu(s):  0.1 us, 15.8 sy,  0.0 ni, 84.0 id,  0.0 wa,  0.0 hi,  0.2 si,  0.0 st
KiB Mem :  4002268 total,  3021036 free,   314844 used,   666388 buff/cache
KiB Swap:  2097148 total,  2097148 free,        0 used.  3350108 avail Mem

   PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
  2560 root      20   0 2152320 103380 102744 S 113.0  2.6   3:00.62 containerd-shim

root@ubuntu:/opt/kata/bin# top -Hp 2560
[...]

PR #5830, opened 2022-12-03
runtime-rs: fix high cpu

quanweiZhou:
Fixed the issue when using nonblocking, the `tokio::io::copy()` needing to handle EAGAIN, resulting in high CPU usage.

Fixes: #5740
Signed-off-by: Quanwei Zhou <quanweiZhou@linux.alibaba.com>

### DIFF
--- a/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/shim_io.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/shim_io.rs
@@ -6,7 +6,11 @@
 
 use std::{
     io,
-    os::unix::{io::FromRawFd, net::UnixStream as StdUnixStream},
+    os::unix::{
+        io::{FromRawFd, RawFd},
+        net::UnixStream as StdUnixStream,
+        prelude::AsRawFd,
+    },
     pin::Pin,
     task::Context as TaskContext,
     task::Poll,
@@ -52,8 +56,21 @@ impl ShimIo {
             "new shim io stdin {:?} stdout {:?} stderr {:?}", stdin, stdout, stderr
         );
 
+        let set_flag_with_blocking = |fd: RawFd| {
+            let flag = unsafe { libc::fcntl(fd, libc::F_GETFL) };
+            let ret = unsafe { libc::fcntl(fd, libc::F_SETFL, flag & !libc::O_NONBLOCK) };
+            if ret < 0 {
+                error!(sl!(), "failed to set fcntl for fd {} error {}", fd, ret);
+            }
+        };
+
         let stdin_fd: Option<Box<dyn AsyncRead + Send + Unpin>> = if let Some(stdin) = stdin {
             info!(sl!(), "open stdin {:?}", &stdin);
+
+            // Since the stdin peer point (which is hold by containerd) could not be openned
+            // immediately, which would block here's open with block mode, and we wouldn't want to
+            // block here, thus here opened with nonblock and then reset it to block mode for
+            // tokio async io.
             match OpenOptions::new()
                 .read(true)
                 .write(false)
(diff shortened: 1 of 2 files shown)


src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/shim_io.rs
```diff
@@ -61,7 +73,10 @@ impl ShimIo {
                 .open(&stdin)
                 .await
             {
-                Ok(file) => Some(Box::new(file)),
+                Ok(file) => {
+                    set_flag_with_blocking(file.as_raw_fd());
```
Ji-Xinyou, line 77: Since this is not intuitive, maybe add a comment like
```Rust
// Set it to blocking to avoid infinitely handling EAGAIN when the reader is empty
```

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
```diff
@@ -133,22 +133,14 @@ impl Process {
...
-                    }
-                };
-                break;
-            }
+            match tokio::io::copy(&mut reader, &mut writer).await {
+                Err(e) => {
+                    warn!(logger, "run_io_copy: failed to copy stream: {}", e);
+                }
+                Ok(length) => {
+                    warn!(logger, "run_io_copy: stop to copy stream length {}", length)
```
Ji-Xinyou, line 141: I think the stream copy future terminates on this. Maybe the log could be at `debug` level or `info` level, what do you think.

Ji-Xinyou (approved):
Nice work and LGTM! Thanks @quanweiZhou. Only some nits below.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/shim_io.rs
quanweiZhou: Thanks, @Ji-Xinyou already added a comment.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
quanweiZhou: Thanks, @Ji-Xinyou change the `warn!()` to `info!()`.

Tim-0731-Hzt (approved):
thanks!  LGTM

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/shim_io.rs
```diff
@@ -61,7 +73,11 @@ impl ShimIo {
                 .open(&stdin)
                 .await
             {
-                Ok(file) => Some(Box::new(file)),
+                Ok(file) => {
+                    // Set it to blocking to avoid infinitely handling EAGAIN when the reader is empty
+                    set_flag_with_blocking(file.as_raw_fd());
```
bergwolf, line 78: Why not just remove the `O_NONBLOCK` flag from the above open options?

lifupan: Since the stdin peer point (which is hold by containerd) could not be openned immediately, which would block here's open with block mode, and we wouldn't want to block here, thus here opened with nonblock and then reset it to block mode for tokio async io. 

@quanweiZhou  You'd better comment it clearly why opened it with  "O_NONBLOCK" mode.

quanweiZhou: Thanks, @bergwolf @lifupan add comment for this.

bergwolf: Good point!

bergwolf (approved):

Issue #5778, opened 2022-11-28
Drop the `Option` for LinuxContainer.cgroup_manager


liubin:
> Unrelated to this PR, but I wonder why wouldn't we always have a cgroup manager for a container. IOW, can we drop the `Option` here? It will simplify the code a lot by not having to check for NONE everywhere cgroup_manager is used.

_Originally posted by @bergwolf in https://github.com/kata-containers/kata-containers/pull/5154#discussion_r1033116090_

PR #5848, opened 2022-12-06
agent: Drop the Option for LinuxContainer.cgroup_manager

Yuan-Zhuo:
Cgroup manager for a container will always be created.
Thus, dropping the option for LinuxContainer.cgroup_manager is feasible and could simplify the code.

Fixes: #5778


bergwolf (approved):
Nice cleanup!

Issue #5849, opened 2022-12-06
CC: SNP OVMF case for kata-deploy


Alex-Carter01:
SNP needs two builds of ovmf: the AmdSev build and the normal x86_64 build.

Needs cases in the makefile, kata-deploy scripts.

Currently building those two ovmfs have conflicting paths and overwrite, so writing them to slightly different paths.

jepio:
> SNP needs two builds of ovmf: the AmdSev build and the normal x86_64 build.

Why would SNP need two builds of ovmf?

fitzthum:
> Why would SNP need two builds of ovmf?

SNP only needs the AmdSev build, which has some extra stuff for kernel hashes support. Maybe in the future this could be added to the main Ovmf build.

Anyway, we did this a long time ago.

ariel-adam:
Following comment from Tobin closing this issue

PR #5850, opened 2022-12-06
CC: kata-deploy add x86 ovmf build for SNP

Alex-Carter01:
SNP needs two builds of ovmf: the AmdSev build and the normal x86_64 build.

Adds target for vanilla ovmf build for snp

Adding another make target / kata-deploy function, and fixing the ovmf builder so these builds dont overlap.

Fixes: kata-containers#5849

### DIFF
--- a/.github/workflows/cc-payload-after-push-amd64.yaml
+++ b/.github/workflows/cc-payload-after-push-amd64.yaml
@@ -19,6 +19,7 @@ jobs:
           - cc-virtiofsd
           - cc-sev-kernel
           - cc-sev-ovmf
+          - cc-x86_64-ovmf
           - cc-sev-rootfs-initrd
           - cc-tdx-kernel
           - cc-tdx-rootfs-image
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -17,6 +17,7 @@ EXTRA_TARBALL=cc-cloud-hypervisor-tarball \
 	cc-tdx-td-shim-tarball \
 	cc-tdx-tdvf-tarball \
 	cc-sev-ovmf-tarball \
+	cc-x86_64-ovmf-tarball \
 	cc-sev-rootfs-initrd-tarball \
 	cc-tdx-rootfs-image-tarball
 endif
@@ -151,4 +152,7 @@ cc-tdx-tdvf-tarball:
 	${MAKE} $@-build
 
 cc-sev-ovmf-tarball:
+	${MAKE} $@-build
+
+cc-x86_64-ovmf-tarball:
 	${MAKE} $@-build
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -105,6 +105,7 @@ options:
 	cc-shimv2
 	cc-virtiofsd
 	cc-sev-ovmf
+	cc-x86_64-ovmf
 EOF
 
 	exit "${return_code}"
@@ -540,6 +541,10 @@ install_cc_sev_ovmf(){
  	install_cc_tee_ovmf "sev" "edk2-sev.tar.gz"
 }
 
+install_cc_x86_64_ovmf(){
+ 	install_cc_tee_ovmf "x86_64" "edk2-x86_64.tar.gz"
+}
+
 #Install guest image
 install_image() {
 	info "Create image"
@@ -709,6 +714,8 @@ handle_build() {
 	cc-tdx-tdvf) install_cc_tdx_tdvf ;;
 
 	cc-sev-ovmf) install_cc_sev_ovmf ;;
+	
+	cc-x86_64-ovmf) install_cc_x86_64_ovmf ;;
 
 	cloud-hypervisor) install_clh ;;
 
--- a/tools/packaging/static-build/ovmf/build-ovmf.sh
+++ b/tools/packaging/static-build/ovmf/build-ovmf.sh
@@ -79,9 +79,11 @@ popd
 
 info "Install fd to destdir"
 install_dir="${DESTDIR}/${PREFIX}/share/ovmf"
-if [ "${ovmf_build}" == "tdx" ]; then
-	install_dir="$DESTDIR/$PREFIX/share/tdvf"
-fi
+case "${ovmf_build}" in
+	"tdx")
+		install_dir="$DESTDIR/$PREFIX/share/tdvf"
+		;;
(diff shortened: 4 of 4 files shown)


fidencio:
@Alex-Carter01, do you want / need the OVMF to be part of the payload generated for / consumed by the Operator?
If not, then what you have is fine.  Otherwise you'll have to also update the following files:
* https://github.com/kata-containers/kata-containers/blob/CCv0/.github/workflows/cc-payload-after-push-amd64.yaml
* https://github.com/kata-containers/kata-containers/blob/CCv0/.github/workflows/cc-payload.yaml

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -251,6 +252,10 @@ install_cc_sev_ovmf(){
  	install_cc_tee_ovmf "sev" "edk2-sev.tar.gz"
 }
 
+install_cc_x86_ovmf(){
+ 	install_cc_tee_ovmf "" "edk2-x86_64.tar.gz"
```
wainersm, line 256: Hi @Alex-Carter01 , to use `install_cc_tee_ovmf()` you will need to pass a `tee` value otherwise the function will not get the ovmf url/version correctly (see https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L469)

Alex-Carter01:
Rebasing to fix merge conflicts, specifying ovmf_build so the case in the static builder is actually hit.

Alex-Carter01:
Will want something like #6120 to let the CI cache this, should that be a part of this PR?

Also will be adding to operator yamls

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -79,9 +79,14 @@ popd
 
 info "Install fd to destdir"
 install_dir="${DESTDIR}/${PREFIX}/share/ovmf"
-if [ "${ovmf_build}" == "tdx" ]; then
-	install_dir="$DESTDIR/$PREFIX/share/tdvf"
-fi
+case "${ovmf_build}" in
+	"tdx")
+		install_dir="$DESTDIR/$PREFIX/share/tdvf"
```
fitzthum, line 84: Is there a reason we are putting the firmware files in `share` rather than in `bin`?  @fidencio

tools/packaging/static-build/ovmf/build-ovmf.sh
```diff
@@ -79,9 +79,14 @@ popd
...
+		;;
+	"x86_64")
+		install_dir="$DESTDIR/$PREFIX/share/ovmfx86_64"
```
fitzthum, line 87: I would prefer to have the two versions of ovmf be called `AMDSEV.fd` for the AmdSevPkg and `OVMF.fd` for the normal x64 package. This would match a bit better with the conventions that I have seen and there would be no need for this additional `ovmfx86_64` directory.

fitzthum (commented):
Looks good, but I have some questions about some of the naming conventions.

.github/workflows/cc-payload-after-push-amd64.yaml
```diff
@@ -19,6 +19,7 @@ jobs:
           - cc-virtiofsd
           - cc-sev-kernel
           - cc-sev-ovmf
+          - cc-x86-ovmf
```
larrydewey, line 22: Not a blocker by any means, but would it be good to specify `x86_64` vs `x86` here?

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -18,6 +18,7 @@ EXTRA_TARBALL=cc-cloud-hypervisor-tarball \
 	cc-tdx-td-shim-tarball \
 	cc-tdx-tdvf-tarball \
 	cc-sev-ovmf-tarball \
+	cc-x86-ovmf-tarball \
```
larrydewey, line 21: Same as previous statement

tools/packaging/static-build/ovmf/build-ovmf.sh
larrydewey: I agree with @fitzthum. That would make the most sense here.

larrydewey (approved):
Just a few suggestions regarding the naming convention which honestly applies to the entire set of variables, function, string literals, etc.

Alex-Carter01:
Renamed everything to x86_64 and opened #6338 to have the AmdSevPkg build AMDSEV.fd

fitzthum (approved):
Ok, I think we're good here.

fitzthum:
@wainersm have your concerns been addressed here?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
Alex-Carter01: Has this been resolved? I now specify x86_64 ovmf_build

src/runtime/Makefile
```diff
@@ -124,7 +124,7 @@ FIRMWAREVOLUMEPATH :=
 TDVFFIRMWAREPATH := $(PREFIXDEPS)/share/tdvf/OVMF_CODE.fd
 TDVFFIRMWAREVOLUMEPATH := $(PREFIXDEPS)/share/tdvf/OVMF_VARS.fd
 TDSHIMFIRMWAREPATH := ${PREFIXDEPS}/share/td-shim/td-shim.bin
-SEVFIRMWAREPATH := $(PREFIXDEPS)/share/ovmf/OVMF.fd
+SEVFIRMWAREPATH := $(PREFIXDEPS)/share/ovmf/AMDSEV.fd
```
jepio, line 127: this change is already in CCv0. Might it be a good idea to rebase this commit?

tools/packaging/static-build/ovmf/build-ovmf.sh
jepio: qemu traditionally keeps OVMF/firmware files in `/usr/share/...`

jepio:
It's rather unlucky, but the commit message parser for `Fixes:` does not accept a repo name as a prefix. Can you update it to `Fixes: #5849`?

jepio (approved):
I believe this is good to go - but the commit message needs to be fixed so that static checks pass.

Alex-Carter01:
fixing 'fixes' in commit message for static checks

ryansavino (approved):
LGTM, thanks Alex.

stevenhorsman (approved):
passes the tests and has a few SME reviews, so I think it's ready to be merged

Issue #5871, opened 2022-12-09
kata-sys-util: umounting by umount2 encounters "no such a file or directory" issue


justxuewei:
`umount2`, which you may refer to [this link](https://github.com/kata-containers/kata-containers/blob/5ef7ed72aec6a606eb9642a86c48e5d236576c79/src/libs/kata-sys-util/src/mount.rs#L762), returns "no such a file or directory" error when umounting a mount point. The root cause is that the strings in the Rust don't have `\0` at the end, but C does.

```rust
let path_ptr = path.as_ref().as_os_str().as_bytes().as_ptr() as *const c_char;
// == snipped ==
if unsafe { libc::umount2(path_ptr, flags) } < 0 {
    Err(io::Error::last_os_error())
} else {
    Ok(())
}
```

As aforementioned, the `libc::umount2` couldn't get the correct path in the above snippet.

PR #5873, opened 2022-12-09
kata-sys-util: fix issues where umount2 couldn't get the correct path

justxuewei:
Strings in Rust don't have \0 at the end, but C does, which leads to `umount2`
in the libc can't get the correct path. Besides, calling `nix::mount::umount2`
to avoid using an unsafe block is a robust solution.

Fixes: https://github.com/kata-containers/kata-containers/issues/5871


justxuewei:
@lifupan @jiangliu @liubin @bergwolf PTAL, thanks.

bergwolf (approved):
lgtm!

lifupan (approved):

Issue #5874, opened 2022-12-09
Move shim management client side to libs


Ji-Xinyou:
The client side is a public interface to be called from, it should be put in libs and included as needed.

PR #5875, opened 2022-12-09
refactor(shim-mgmt): move client side to libs

Ji-Xinyou:
The client side is moved to libs. This is to solve the problem that including clients will bring about messy dependencies.

Fixes: #5874
Signed-off-by: Ji-Xinyou <jerryji0414@outlook.com>

### DIFF
--- a/src/libs/shim-interface/src/lib.rs
+++ b/src/libs/shim-interface/src/lib.rs
@@ -1,0 +1,66 @@
+// Copyright (c) 2022 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! shim-interface is a common library for different components of Kata Containers
+//! to make function call through services inside the runtime(runtime-rs runtime).
+//!
+//! Shim management:
+//! Currently, inside the shim, there is a shim management server running as the shim
+//! starts, working as a RESTful server. To make function call in runtime from another
+//! binary, using the utilities provided in this library is one of the methods.
+//!
+//! You may construct clients by construct a MgmtClient and let is make specific
+//! HTTP request to the server. The server inside shim will multiplex the request
+//! to its corresponding handler and run certain methods.
+
+use std::path::Path;
+
+use anyhow::{anyhow, Result};
+
+pub mod shim_mgmt;
+
+pub const KATA_PATH: &str = "/run/kata";
+pub const SHIM_MGMT_SOCK_NAME: &str = "shim-monitor.sock";
+
+// return sandbox's storage path
+pub fn sb_storage_path() -> String {
+    String::from(KATA_PATH)
+}
+
+// returns the address of the unix domain socket(UDS) for communication with shim
+// management service using http
+// normally returns "unix:///run/kata/{sid}/shim_monitor.sock"
+pub fn mgmt_socket_addr(sid: &str) -> Result<String> {
+    if sid.is_empty() {
+        return Err(anyhow!(
+            "Empty sandbox id for acquiring socket address for shim_mgmt"
+        ));
+    }
+
+    let p = Path::new(&sb_storage_path())
+        .join(sid)
+        .join(SHIM_MGMT_SOCK_NAME);
+
+    if let Some(p) = p.to_str() {
+        Ok(format!("unix://{}", p))
+    } else {
+        Err(anyhow!("Bad socket path"))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_mgmt_socket_addr() {
+        let sid = "414123";
+        let addr = mgmt_socket_addr(sid).unwrap();
+        assert_eq!(addr, "unix:///run/kata/414123/shim-monitor.sock");
+
+        let sid = "";
+        assert!(mgmt_socket_addr(sid).is_err());
+    }
+}
--- a/src/libs/shim-interface/src/shim_mgmt/mod.rs
+++ b/src/libs/shim-interface/src/shim_mgmt/mod.rs
@@ -1,0 +1,24 @@
+// Copyright (c) 2022 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// The shim management client module
+pub mod client;
+
+/// The key for direct volume path
+pub const DIRECT_VOLUME_PATH_KEY: &str = "path";
+/// URL for stats direct volume
+pub const DIRECT_VOLUME_STATS_URL: &str = "/direct-volume/stats";
+/// URL for resizing direct volume
+pub const DIRECT_VOLUME_RESIZE_URL: &str = "/direct-volume/resize";
+/// URL for querying agent's socket
+pub const AGENT_URL: &str = "/agent-url";
+/// URL for operation on guest iptable (ipv4)
+pub const IP_TABLE_URL: &str = "/iptables";
+/// URL for operation on guest iptable (ipv6)
+pub const IP6_TABLE_URL: &str = "/ip6tables";
+/// URL for querying metrics inside shim
+pub const METRICS_URL: &str = "/metrics";
+
+pub const ERR_NO_SHIM_SERVER: &str = "Failed to create shim management server";
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner_hypervisor.rs
@@ -14,7 +14,7 @@ use kata_types::capabilities::Capabilities;
 
 use super::inner::DragonballInner;
 use crate::{utils, VcpuThreadIds, VmmState};
-use persist::KATA_PATH;
+use shim_interface::KATA_PATH;
 const DEFAULT_HYBRID_VSOCK_NAME: &str = "kata.hvsock";
 
 fn get_vsock_path(root: &str) -> String {
--- a/src/runtime-rs/crates/persist/src/lib.rs
+++ b/src/runtime-rs/crates/persist/src/lib.rs
@@ -7,9 +7,9 @@
 pub mod sandbox_persist;
 use anyhow::{anyhow, Context, Ok, Result};
 use serde::de;
+use shim_interface::KATA_PATH;
 use std::{fs::File, io::BufReader};
 
-pub const KATA_PATH: &str = "/run/kata";
 pub const PERSIST_FILE: &str = "state.json";
 use kata_sys_util::validate::verify_id;
 use safe_path::scoped_join;
--- a/src/runtime-rs/crates/runtimes/src/manager.rs
+++ b/src/runtime-rs/crates/runtimes/src/manager.rs
@@ -19,6 +19,7 @@ use kata_types::{annotations::Annotation, config::TomlConfig};
 #[cfg(feature = "linux")]
 use linux_container::LinuxContainer;
 use persist::sandbox_persist::Persist;
+use shim_interface::shim_mgmt::ERR_NO_SHIM_SERVER;
 use tokio::sync::{mpsc::Sender, RwLock};
 #[cfg(feature = "virt")]
 use virt_container::{
@@ -117,7 +118,9 @@ impl RuntimeHandlerManagerInner {
         let shim_mgmt_svr = MgmtServer::new(
             &self.id,
             self.runtime_instance.as_ref().unwrap().sandbox.clone(),
-        );
+        )
+        .context(ERR_NO_SHIM_SERVER)?;
+
         tokio::task::spawn(Arc::new(shim_mgmt_svr).run());
         info!(sl!(), "shim management http server starts");
 
--- a/src/runtime-rs/crates/runtimes/src/shim_mgmt/mod.rs
+++ b/src/runtime-rs/crates/runtimes/src/shim_mgmt/mod.rs
@@ -4,6 +4,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-pub mod client;
+//! The server side of shim management implementation, receive HTTP
+//! requests and multiplex them to corresponding functions inside shim
+//!
+//! To call services in a RESTful convention, use the client
+//! from libs/shim-interface library
+
 mod handlers;
 pub mod server;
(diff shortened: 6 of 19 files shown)


Ji-Xinyou:
Just an explication, the extra directory under shim-services, i.e. shim-mgmt, is reserved for future service interfaces.

src/libs/shim-services/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+// return sandbox's storage path
+pub fn sb_storage_path() -> String {
+    String::from(KATA_PATH)
+}
+
+// returns the address of the unix domain socket(UDS) for communication with shim
+// management service using http
+// normally returns "unix:///run/kata/{sid}/shim_monitor.sock"
+pub fn mgmt_socket_addr(sid: String) -> String {
```
Tim-0731-Hzt, line 33: Should the `sid` value be verified in this function? what if `sid = "../../tmp/malicous-hacker-directory/"` ? it can use https://github.com/kata-containers/kata-containers/tree/main/src/libs/safe-path

Ji-Xinyou: True, I think we also need to check some of the possible places where we can appropriately use this.

Ji-Xinyou: No, this socket address is used when some endpoint want to connect it. If the address is not valid, the connection will surely fail.

Tim-0731-Hzt: Yes, but I think you need to check if that is valid and return some error information for users

Ji-Xinyou: Still I dont think that is necessary. This functions is equivalent to formatting a `sid` to a `uds`. It is the caller's responsibility to check the validity of socket address.

Tim-0731-Hzt: fair enough

src/libs/shim-services/src/shim_mgmt/client.rs
```diff
@@ -10,13 +11,13 @@
...
 
-use super::server::mgmt_socket_addr;
+use crate::mgmt_socket_addr;
 use anyhow::{anyhow, Context, Result};
 use hyper::{Body, Client, Method, Request, Response};
 use hyperlocal::{UnixClientExt, UnixConnector, Uri};
 
 /// Shim management client with timeout
-pub struct MgmtClient {
+struct MgmtClient {
```
studychao, line 20: Why `MgmtClient` is not public? What if other crate want to use this `struct`

src/libs/shim-services/src/shim_mgmt/mod.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2019-2022 Alibaba Cloud
```
studychao, line 1: same here

src/runtime-rs/crates/hypervisor/src/dragonball/inner_hypervisor.rs
```diff
@@ -14,7 +14,7 @@ use kata_types::capabilities::Capabilities;
 
 use super::inner::DragonballInner;
 use crate::{utils, VcpuThreadIds, VmmState};
-use persist::KATA_PATH;
+use shim_services::KATA_PATH;
```
studychao, line 17: NIT: It's kinda strange that `inner_hypervisor` use `shim_service`, could you find some better space to put `KATA_PATH`?

src/runtime-rs/crates/persist/src/lib.rs
```diff
@@ -7,9 +7,9 @@
 pub mod sandbox_persist;
 use anyhow::{anyhow, Context, Ok, Result};
 use serde::de;
+use shim_services::KATA_PATH;
```
studychao, line 10: same as above

studychao (commented):
Thanks @Ji-Xinyou , a few comments

src/runtime-rs/crates/hypervisor/src/dragonball/inner_hypervisor.rs
Ji-Xinyou: I agree, but there would be another problem.
1. this variable has to reside in a library.
2. if the variable is not in shim_services, then shim_services will introduce another library as dependency.
I think the library should not be dependent on any kata libararies? Maybe this could be resolved in another PR?

src/libs/shim-services/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+
+// return sandbox's storage path
+pub fn sb_storage_path() -> String {
+    String::from(KATA_PATH)
+}
+
+// returns the address of the unix domain socket(UDS) for communication with shim
+// management service using http
+// normally returns "unix:///run/kata/{sid}/shim_monitor.sock"
+pub fn mgmt_socket_addr(sid: &str) -> String {
```
jodh-intel, line 33: Should this return `Result<String>` to handle the case where `sid == ""`?

src/libs/shim-services/src/shim_mgmt/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+/// The shim management client module
+pub mod client;
+
+/// The key for direct volumn path
+pub const DIRECT_VOLUMN_PATH_KEY: &str = "path";
+/// URL for stats direct volumn
+pub const DIRECT_VOLUMN_STATS_URL: &str = "/direct-volumn/stats";
+/// URL for resizing direct volumn
+pub const DIRECT_VOLUMN_RESIZE_URL: &str = "/direct-volumn/resize";
```
jodh-intel, line 14: I know we have this identifier in the repo already but I'm wondering if it's a typo? Should it be...

```suggestion
/// URL for resizing direct volume
pub const DIRECT_VOLUME_RESIZE_URL: &str = "/direct-volume/resize";
```

src/libs/shim-services/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "shim-services"
```
jodh-intel, line 2: I think it would be helpful somewhere to state that we're talking about the runtime-rs (runtime) shim to avoid confusion with any other shims ;) It's the same sort of issue as when someone mentions "proxy", "handler", "interface", "protocol", I always think, "which one?" ;)

src/runtime-rs/crates/hypervisor/src/dragonball/inner_hypervisor.rs
Ji-Xinyou: I have adopted @jodh-intel 's advice, use another name as `shim-interface`, as further interfaces facing other binaries could be put their. Variables like `KATA_PATH` makes more sense as well.

src/libs/shim-services/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+//!
+//! You may construct clients by construct a MgmtClient and let is make specific
+//! HTTP request to the server. The server inside shim will multiplex the request
+//! to its corresponding handler and run certain methods.
+
+use std::path::Path;
+
+pub mod shim_mgmt;
+
+pub const KATA_PATH: &str = "/run/kata";
```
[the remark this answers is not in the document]
Ji-Xinyou: In my commit, I move this from `persist` to libs, since the `KATA_PATH` should be globally accessible and remains consistent(i.e. single definition).

src/libs/shim-services/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub const SHIM_MGMT_SOCK_NAME: &str = "shim-monitor.sock";
```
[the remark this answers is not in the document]
Ji-Xinyou: It could be, but the original name w/o `kata-` prefix was the behavior of Kata version 2.0.

src/libs/shim-services/src/lib.rs
Ji-Xinyou: Thanks, fixed

src/libs/shim-services/Cargo.toml
Ji-Xinyou: I think the interface would be more generic and appropriate. Since this works as a interface facing outside binaries.

src/libs/shim-services/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2019-2022 Alibaba Cloud
```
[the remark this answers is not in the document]
Ji-Xinyou: Fixed

src/libs/shim-services/src/shim_mgmt/client.rs
Ji-Xinyou: Fixed

src/libs/shim-services/src/shim_mgmt/mod.rs
Ji-Xinyou: Fixed, nice catch.

src/libs/shim-interface/src/lib.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) s022 Alibaba Cloud
```
studychao, line 1: typo? s022?

src/runtime-rs/crates/runtimes/src/shim_mgmt/mod.rs
```diff
@@ -4,6 +4,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-pub mod client;
+//! The server side of shim management implementation, receive HTTP
+//! requests and multiplex them to corresponding functions inside shim
+//!
+//! To call services in a RESTful convention, use the client
+//! from libs/shim-services library
```
studychao, line 11: Does this rename to shim-interface?

src/libs/shim-interface/src/shim_mgmt/mod.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) -2022 Alibaba Cloud
```
studychao, line 1: type `-2022`

studychao (commented):
Thanks @Ji-Xinyou , a few more comments.

src/libs/shim-interface/src/lib.rs
Ji-Xinyou: fixed

src/runtime-rs/crates/runtimes/src/shim_mgmt/mod.rs
Ji-Xinyou: fixed

src/libs/shim-interface/src/shim_mgmt/mod.rs
Ji-Xinyou: fixed

studychao (approved):
LGTM, thanks.

studychao:
@jodh-intel please take a look at this

src/libs/shim-interface/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    if sid.is_empty() {
+        return Err(anyhow!(
+            "Empty sandbox id for acquiring socket address for shim_mgmt"
+        ));
+    }
+
+    let p = Path::new(&sb_storage_path())
+        .join(sid)
+        .join(SHIM_MGMT_SOCK_NAME);
+    Ok(format!("unix://{}", p.to_string_lossy()))
```
jodh-intel, line 45: Nit: Now that this returns a `Result`, it might be worth calling `p.to_str()` and returning an error if the conversion fails?

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -117,7 +117,8 @@ impl RuntimeHandlerManagerInner {
         let shim_mgmt_svr = MgmtServer::new(
             &self.id,
             self.runtime_instance.as_ref().unwrap().sandbox.clone(),
-        );
+        )
+        .context("Failed to spawn shim_mgmt_svr")?;
```
jodh-intel, line 121: Nit: consistency: there are a few references to "shim_mgmt" in errors, but also (the much clearer imho) `"Failed to create shim-mgmt server"`. I'd consider making all the error messages consistent and spelling the message out in full by using a `const` to avoid duplication and ensure consistency. Something like:

```
const ERR_NO_SHIM_SERVER: &str = "Failed to create shim management server";
```

src/libs/shim-interface/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+//! to make function call through services inside the shim.
+//!
+//! Shim management:
+//! Currently, inside the shim, there is a shim management server running as the shim
+//! starts, working as a RESTful server. To make function call in shim from another
+//! binary, using the utilities provided in this library is one of the methods.
+//!
+//! You may construct clients by construct a MgmtClient and let is make specific
+//! HTTP request to the server. The server inside shim will multiplex the request
+//! to its corresponding handler and run certain methods.
```
jodh-intel, line 16: Nit: I still think this explanatory text should mention "runtime" somewhere ;)

jodh-intel (approved):
Thanks @Ji-Xinyou. A few nits, but overall...

lgtm

src/libs/shim-interface/src/lib.rs
Ji-Xinyou: I think you are right, fixed.

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: I agree, but I also find out a problem of kata right now.

We currently are abusing the `anyhow`, instead of implementing our own `Error` types. Abusing crates like `anyhow` and `thiserror` 's formatting tricks are not idiomatic IMO.

WDYT @jodh-intel

src/libs/shim-interface/src/lib.rs
Ji-Xinyou: Fixed

Ji-Xinyou:
Thanks for the review @jodh-intel , I fixed the nits you mentioned and raised a question in one of them.

src/runtime-rs/crates/runtimes/src/manager.rs
jodh-intel: @Ji-Xinyou - yes, I tend to agree. Could you raise an issue so this topic can be discussed separately?

jodh-intel:
Thanks @Ji-Xinyou ! Please could you rebase to resolve the lock file conflict?

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: @jodh-intel I have raised an issue. https://github.com/kata-containers/kata-containers/issues/5918

jodh-intel: Thanks @Ji-Xinyou ! :smile:

Issue #5872, opened 2022-12-09
runtime-rs: failed to run/start a container from bundle with runtime-rs


YushuoEdge:
### Command
```
$ ls ./bundle
config.json  rootfs
$ sudo ctr run -d --runtime io.containerd.kata.v2 --config bundle/config.json test_kata
ctr: failed to create shim task: Others("failed to handler message create container\n\nCaused by:\n    0: create\n    1: handler rootfs\n    2: unsupported rootfs mounts count 0"): unknown


$ sudo /usr/local/bin/containerd-shim-kata-v2 --version
Kata Containers containerd shim (Rust): id: io.containerd.kata.v2, version: 3.1.0-alpha0, commit: 39394fa2a8d1a832163cdb8b97b2bf235648be45-dirty
```

##### Rootfs
Use the script to create rootfs.
```
ctr i pull quay.io/prometheus/busybox:latest
ctr i export rootfs.tar quay.io/prometheus/busybox:latest

rootfs_tar=rootfs.tar
bundle_dir="./bundle"
mkdir -p "${bundle_dir}"

# extract busybox rootfs
rootfs_dir="${bundle_dir}/rootfs"
mkdir -p "${rootfs_dir}"
layers_dir="$(mktemp -d)"
tar -C "${layers_dir}" -pxf "${rootfs_tar}"
for ((i=0;i<$(cat ${layers_dir}/manifest.json | jq -r ".[].Layers | length");i++)); do
  tar -C ${rootfs_dir} -xf ${layers_dir}/$(cat ${layers_dir}/manifest.json | jq -r ".[].Layers[${i}]")
done
```

##### config.json
Use `runc spec` to generate config.json.
```
{
    "ociVersion":"1.0.2-dev",
    "process":{
        "terminal":true,
        "user":{
            "uid":0,
            "gid":0
        },
        "args":[
            "sh"
        ],
        "env":[
            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
            "TERM=xterm"
        ],
        "cwd":"/",
        "capabilities":{
            "bounding":[
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
            ],
            "effective":[
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
            ],
            "permitted":[
                "CAP_AUDIT_WRITE",
[...]

PR #5877, opened 2022-12-10
runtime-rs: enable start container from bundle

Tim-0731-Hzt:
enable start container from bundle in this way

$ ls ./bundle
config.json  rootfs
$ sudo ctr run -d --runtime io.containerd.kata.v2 --config bundle/config.json test_kata

Fixes:#5872

### DIFF
--- a/docs/how-to/containerd-kata.md
+++ b/docs/how-to/containerd-kata.md
@@ -257,6 +257,48 @@ This launches a BusyBox container named `hello`, and it will be removed by `--rm
 The `--cni` flag enables CNI networking for the container. Without this flag, a container with just a
 loopback interface is created.
 
+### Launch containers using `ctr` command line with rootfs bundle
+
+#### Get rootfs
+Use the script to create rootfs
+```bash
+ctr i pull quay.io/prometheus/busybox:latest
+ctr i export rootfs.tar quay.io/prometheus/busybox:latest
+
+rootfs_tar=rootfs.tar
+bundle_dir="./bundle"
+mkdir -p "${bundle_dir}"
+
+# extract busybox rootfs
+rootfs_dir="${bundle_dir}/rootfs"
+mkdir -p "${rootfs_dir}"
+layers_dir="$(mktemp -d)"
+tar -C "${layers_dir}" -pxf "${rootfs_tar}"
+for ((i=0;i<$(cat ${layers_dir}/manifest.json | jq -r ".[].Layers | length");i++)); do
+  tar -C ${rootfs_dir} -xf ${layers_dir}/$(cat ${layers_dir}/manifest.json | jq -r ".[].Layers[${i}]")
+done
+```
+#### Get `config.json`
+Use runc spec to generate `config.json`
+```bash
+cd ./bundle/rootfs
+runc spec
+mv config.json ../
+```
+Change the root `path` in `config.json` to the absolute path of rootfs
+
+```JSON
+"root":{
+    "path":"/root/test/bundle/rootfs",
+    "readonly": false
+},
+```
+
+#### Run container
+```bash
+sudo ctr run -d --runtime io.containerd.run.kata.v2 --config bundle/config.json hello
+sudo ctr t exec --exec-id ${ID} -t hello sh
+```
 ### Launch Pods with `crictl` command line
 
 With the `crictl` command line of `cri-tools`, you can specify runtime class with `-r` or `--runtime` flag.
--- a/src/runtime-rs/crates/resource/src/manager.rs
+++ b/src/runtime-rs/crates/resource/src/manager.rs
@@ -68,11 +68,14 @@ impl ResourceManager {
     pub async fn handler_rootfs(
         &self,
         cid: &str,
+        root: &oci::Root,
         bundle_path: &str,
         rootfs_mounts: &[Mount],
     ) -> Result<Arc<dyn Rootfs>> {
         let inner = self.inner.read().await;
-        inner.handler_rootfs(cid, bundle_path, rootfs_mounts).await
+        inner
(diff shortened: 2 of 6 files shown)


src/runtime-rs/crates/resource/src/rootfs/share_fs_rootfs.rs
```diff
@@ -23,14 +23,21 @@ impl ShareFsRootfs {
...
+        is_mounted: bool,
     ) -> Result<Self> {
-        let bundle_rootfs = format!("{}/{}", bundle_path, ROOTFS);
-        rootfs.mount(&bundle_rootfs).context(format!(
-            "mount rootfs from {:?} to {}",
-            &rootfs, &bundle_rootfs
-        ))?;
-
+        let bundle_rootfs = if is_mounted {
+            bundle_path.to_string()
```
quanweiZhou, line 30: I'm confused about why use bundle_path as the rootfs source.

src/runtime-rs/crates/resource/src/rootfs/mod.rs
```diff
@@ -51,16 +51,38 @@ impl RootFsResource {
...
         hypervisor: &dyn Hypervisor,
         sid: &str,
         cid: &str,
+        root: oci::Root,
         bundle_path: &str,
         rootfs_mounts: &[Mount],
     ) -> Result<Arc<dyn Rootfs>> {
         match rootfs_mounts {
+            // if rootfs_mounts is empty
+            mounts_vec if mounts_vec.is_empty() => {
```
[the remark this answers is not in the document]
Tim-0731-Hzt: This is because `mount_vec` will include the information of containerd snapshot directory, if we are using bundle to create container, we don't need the containerd snapshotter

src/runtime-rs/crates/resource/src/rootfs/share_fs_rootfs.rs
Tim-0731-Hzt: If we use a bundle to start the container, we only need to share the bundle to the shared path between host and guest.

src/runtime-rs/crates/resource/src/rootfs/mod.rs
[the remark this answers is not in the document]
YushuoEdge: I am not sure whether `mounts_vec` is always empty when starting container with bundle. Maybe we should figure it out ...

src/runtime-rs/crates/resource/src/rootfs/share_fs_rootfs.rs
YushuoEdge: I am also confused here. What if the rootfs is not mounted under `bundle/rootfs`, will report an error?
And can mounted nydus-rootfs be supported?

src/runtime-rs/crates/resource/src/manager.rs
```diff
@@ -68,11 +68,14 @@ impl ResourceManager {
     pub async fn handler_rootfs(
         &self,
         cid: &str,
+        root: oci::Root,
```
YushuoEdge, line 71: What about use &oci::Root?

YushuoEdge (commented):
Thanks! @Tim-0731-Hzt Some comments...

src/runtime-rs/crates/resource/src/rootfs/mod.rs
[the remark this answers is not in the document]
Tim-0731-Hzt: yes, I think so, because we don't rely on contained snapshotter if we start container in this way ?

src/runtime-rs/crates/resource/src/manager.rs

src/runtime-rs/crates/resource/src/rootfs/share_fs_rootfs.rs
Tim-0731-Hzt: The user should give the absolute path of the `rootfs`. I think It should work with nydus.

bergwolf (approved):

src/runtime-rs/crates/resource/src/rootfs/share_fs_rootfs.rs
```diff
@@ -23,14 +23,21 @@ impl ShareFsRootfs {
...
+        rootfs: Option<&Mount>,
+        is_mounted: bool,
     ) -> Result<Self> {
-        let bundle_rootfs = format!("{}/{}", bundle_path, ROOTFS);
-        rootfs.mount(&bundle_rootfs).context(format!(
-            "mount rootfs from {:?} to {}",
-            &rootfs, &bundle_rootfs
-        ))?;
-
+        let bundle_rootfs = if is_mounted {
```
liubin, line 29: Is `is_mounted == false` == `rootfs.is_some()`? so you can save the `is_mounted` parameter.

Tim-0731-Hzt: Thanks, fixed

docs/how-to/containerd-kata.md
```diff
@@ -257,6 +257,48 @@ This launches a BusyBox container named `hello`, and it will be removed by `--rm
...
+Change the root `path` in `config.json` to the absolute path of rootfs
+
+```JSON
+"root":{
+    "path":"/root/test/bundle/rootfs",
+    "readonly": false
+},
+```
+
+Run container
```
liubin, line 297: Is `Run container` a header? just like `Get rootfs` and `Get config.json`.
You can change this line to `#### Run container`.

liubin (approved):

docs/how-to/containerd-kata.md
Tim-0731-Hzt: fixed

Issue #5895, opened 2022-12-13
CC | Use prebuilt containerd artefacts


fidencio:
In order to avoid building the containerd binaries on every single run
of our CI, let's simply rely on the fact that now we're releasing the
"same" artefacts containerd normally releases from our own fork.

See:
https://github.com/confidential-containers/containerd/releases/tag/v1.6.8.1

PR #5896, opened 2022-12-13
versions: Use prebuilt containerd artefacts

fidencio:
In order to avoid building the containerd binaries on every single run of our CI, let's simply rely on the fact that now we're releasing the "same" artefacts containerd normally releases from our own fork.

See:
https://github.com/confidential-containers/containerd/releases/tag/v1.6.8.1

Fixes: #5895


stevenhorsman (approved):
LGTM. Thanks for the great change in speeding up that part of the build!

liubin (approved):

Issue #5868, opened 2022-12-08
After the node restarts, the pod of kata depoy starts to occur an error


singhwang:
# Description of problem

1. Create a kubernetes cluster and ensure that only one node in the cluster can run pod of type kata.
2. In this cluster, use kata-deploy to install kata containers and make sure the installation is successful.
3. Use the kata-qemu runtimeclass to create a pod of type kata and ensure that it is running state.
4. Restart the node with the kata type pod, wait for the restart to complete, and observe the status of the pod corresponding to kata-depoy

# Expected result

After restarting the node, the state of the pod corresponding to kata-depoy is Running.

# Actual result

After restarting the node, the state of the pod corresponding to kata-depoy is CrashLoopBackOff. Check its log and find some problems.

copying kata artifacts onto host
cp: cannot create regular file '/opt/kata/bin/qemu-system-x86_64': Text file busy
cp: cannot create regular file '/opt/kata/bin/containerd-shim-kata-v2': Text file busy
cp: cannot create regular file '/opt/kata/libexec/virtiofsd': Text file busy

# Further information

Although restarting node may take a few more tries , the probability of occurrence is still very high.

PR #5913, opened 2022-12-15
stable-3.0 | kata-deploy: Fix the pod of kata deploy starts to occur an error

singhwang:
kata-deploy: Fix after the node restarts, the pod of kata depoy starts to report an error

If a pod of type kata is deployed on a machine, after the machine restarts, the pod status corresponding to kata-deploy will be CrashLoopBackOff. The reason is that kata-deploy will update the installation files in the /opt/kata/ directory. At this time, the pod of the kata type is started, and some files in this directory are occupied and are not allowed to be updated. The -u parameter of the cp command is just suitable for solving it.

The following is the error message before modification, and it has been resolved after modification.

$ kubectl logs -f kata-deploy-hjtc6 -n kube-system
cp: cannot create regular file '/opt/kata/bin/qemu-system-x86_64': Text file busy
cp: cannot create regular file '/opt/kata/bin/containerd-shim-kata-v2': Text file busy
cp: cannot create regular file '/opt/kata/libexec/virtiofsd': Text file busy

Fixes: #5868 for stable-3.0 branch


fidencio (approved):
lgtm, thanks @singhwang!

bergwolf (approved):

Issue #5903, opened 2022-12-14
agent: support list and update infiniband interface


Bevisy:
Now kata-agent currently only supports Ethernet interface NIC operation, due to the support for physical NICs such as Mellanox Infiniband, it is necessary to add support for Infiniband interface NICs.
The difference between Ethernet links and Infiniband links is as follows.:
```json
[
{
    "ifindex": 2,
    "ifname": "eth0",
    "flags": ["BROADCAST", "MULTICAST", "UP", "LOWER_UP"],
    "mtu": 1500,
    "qdisc": "fq_codel",
    "operstate": "UP",
    "linkmode": "DEFAULT",
    "group": "default",
    "txqlen": 1000,
    "link_type": "ether",
    "address": "0a:58:0a:e8:06:79",
    "broadcast": "ff:ff:ff:ff:ff:ff",
    "promiscuity": 0,
    "min_mtu": 68,
    "max_mtu": 65535,
    "inet6_addr_gen_mode": "eui64",
    "num_tx_queues": 1,
    "num_rx_queues": 1,
    "gso_max_size": 65536,
    "gso_max_segs": 65535,
    "parentbus": "virtio",
    "parentdev": "virtio5"
  },
  {
    "ifindex": 3,
    "ifname": "ib0",
    "flags": ["BROADCAST", "MULTICAST"],
    "mtu": 4092,
    "qdisc": "noop",
    "operstate": "DOWN",
    "linkmode": "DEFAULT",
    "group": "default",
    "txqlen": 256,
    "link_type": "infiniband",
    "address": "00:00:09:26:fe:80:00:00:00:00:00:00:02:00:00:00:00:00:00:19",
    "broadcast": "00:ff:ff:ff:ff:12:40:1b:ff:ff:00:00:00:00:00:00:ff:ff:ff:ff",
    "promiscuity": 0,
    "min_mtu": 68,
    "max_mtu": 65520,
    "linkinfo": {
      "info_kind": "ipoib",
      "info_data": { "key": "0xffff", "mode": "datagram", "umcast": "0000" }
    },
    "inet6_addr_gen_mode": "eui64",
    "num_tx_queues": 8,
    "num_rx_queues": 1,
    "gso_max_size": 65536,
    "gso_max_segs": 65535,
    "parentbus": "pci",
    "parentdev": "0000:02:00.0"
  }
]
```

Bevisy:
[...]

PR #5916, opened 2022-12-15
Feat(agent): support IPoIB when using infiniband physical card

Bevisy:
Impl list and update interface for IB interface

Fixes: #5903


Bevisy:
> Thanks @Bevisy ! Could you add a howto document to explain how users can use IPoIB with kata?

Ok, I will prepare the document and add some other associated changes

PR #5947, opened 2022-12-22
runtime/Makefile: Get some bits happy on darwin

dcantah:
Substitution in the yq install script doesn't like zsh, and additionally
the version of yq we're using doesn't have a darwin/arm64 build so grab
the amd64 version and let rosetta work its magic.

This also swaps to abspath from readlink -m for the printing of what binaries
to install, as the -m flag doesn't exist on the BSD variant, and this
should be the same behavior.

Fixes: https://github.com/kata-containers/kata-containers/issues/5970

### DIFF
--- a/ci/install_yq.sh
+++ b/ci/install_yq.sh
@@ -43,6 +43,16 @@ function install_yq() {
 	"aarch64")
 		goarch=arm64
 		;;
+	"arm64")
+		# If we're on an apple silicon machine, just assign amd64. 
+		# The version of yq we use doesn't have a darwin arm build, 
+		# but Rosetta can come to the rescue here.
+		if [ $goos == "Darwin" ]; then 
+			goarch=amd64
+		else 
+			goarch=arm64
+		fi
+		;;
 	"ppc64le")
 		goarch=ppc64le
 		;;
@@ -64,7 +74,7 @@ function install_yq() {
 	fi
 
 	## NOTE: ${var,,} => gives lowercase value of var
-	local yq_url="https://${yq_pkg}/releases/download/${yq_version}/yq_${goos,,}_${goarch}"
+	local yq_url="https://${yq_pkg}/releases/download/${yq_version}/yq_${goos}_${goarch}"
 	curl -o "${yq_path}" -LSsf "${yq_url}"
 	[ $? -ne 0 ] && die "Download ${yq_url} failed"
 	chmod +x "${yq_path}"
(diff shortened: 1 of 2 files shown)


dcantah:
cc @egernst

liubin:
@dcantah thanks for your PR.
But you need to create an issue and add `Fixes: #xxx` in your commit message.

Please refer to https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

dcantah:
Oop, should've read. Sure, I'll do sometime soon

liubin:
It is a bit of difference between `-f` and `-m` option:

```
  -f, --canonicalize            canonicalize by following every symlink in
                                every component of the given name recursively;
                                all but the last component must exist
  -m, --canonicalize-missing    canonicalize by following every symlink in
                                every component of the given name recursively,
                                without requirements on components existence
```

If the file doesn't exist, `$(readlink -f)` will return empty. For example, in my PC there is no `/usr/local/bin/data/` directory, run `make show-summary`, and I'll see:

```
	binaries to install :
	 - /usr/local/bin/kata-runtime
	 - /go/src/github.com/kata-containers/kata-containers/src/runtime/containerd-shim-kata-v2
	 - /usr/local/bin/kata-monitor
	 - 
```

In contrast, `readlink -m` will return:

```
	binaries to install :
	 - /usr/local/bin/kata-runtime
	 - /go/src/github.com/kata-containers/kata-containers/src/runtime/containerd-shim-kata-v2
	 - /usr/local/bin/kata-monitor
	 - /usr/local/bin/data/kata-collect-data.sh
```

IMO, since this part is only for printing where the files will be installed, can we just use `abspath` to replace `shell readlink -m`? Then I get this:

```
	binaries to install :
	 - /usr/local/bin/kata-runtime
	 - /usr/local/bin/containerd-shim-kata-v2
	 - /usr/local/bin/kata-monitor
	 - /usr/local/bin/data/kata-collect-data.sh
```

Indeed my `/usr/local/bin/containerd-shim-kata-v2` is a link to my work directory `/go/src/github.com/kata-containers/kata-containers/src/runtime/containerd-shim-kata-v2`.

dcantah:
@liubin Sorry for the delay here, finally back to my darwin box. Yea my setup is odd as I have all of these paths, so it wasn't as obvious, but I've swapped to abspath here as you've suggested

liubin (approved):

egernst (approved):

Issue #5948, opened 2022-12-23
Randomly fail in kata-ctl check_latest_version test


liubin:
> ---- check::tests::check_latest_version stdout ----
thread 'check::tests::check_latest_version' panicked at 'no entry found for key', src/check.rs:207:23
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_display
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:72:5
   3: core::panicking::panic_str
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:56:5
   4: core::option::expect_failed
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/option.rs:1854:5
   5: kata_ctl::check::tests::get_kata_version_by_url
   6: kata_ctl::check::tests::check_latest_version
   7: core::ops::function::FnOnce::call_once
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

PR #5951, opened 2022-12-23
kata-ctl: skip test if access GitHub.com fail

liubin:
This commit will call `error_for_status` after `send`, this call
will generate errors if status code between 400-499 and 500-599.

And sometime access github.com will fail, in this case we can
skip the test to prevent the CI failing.

Fixes: #5948

### DIFF
--- a/src/tools/kata-ctl/Cargo.lock
+++ b/src/tools/kata-ctl/Cargo.lock
@@ -1382,6 +1382,7 @@ dependencies = [
  "seccompiler",
  "serde",
  "serde_json",
+ "shim-interface",
  "slog",
  "slog-scope",
  "thiserror",
@@ -2188,6 +2189,7 @@ dependencies = [
  "safe-path",
  "serde",
  "serde_json",
+ "shim-interface",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


bergwolf (approved):

Tim-0731-Hzt (approved):

studychao (approved):

fidencio (approved):
lgtm, thanks @liubin!

fidencio:
Just adding more context here:

GitHub has an API rate limiting (please, see: https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting) and when that's reached the content returned [here](https://github.com/kata-containers/kata-containers/blob/a3e1257708dff4c16cc1d889f0a8c1c3f362197f/src/tools/kata-ctl/src/check.rs#L200) is something like:
```
{"message": String("API rate limit exceeded for 83.240.61.87. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"), "documentation_url": String("https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting")}
```

This then causes a panick at [here](https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/check.rs#L207):

Issue #5952, opened 2022-12-23
In the arm environment, an error occurred when building the kata static installation package


singhwang:
# Description of problem
In the arm environment, execute the command to enter the local-build directory.
cd $GOPATH/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build
1. An error occurred when executing ./kata-deploy-binaries-in-docker.sh --build=firecracker
2. An error occurred when executing ./kata-deploy-binaries-in-docker.sh --build=virtiofsd

# Expected result
1. firecracker can be built normally without errors in the arm environment.
2. virtiofsd can be built normally without errors in the arm environment.

# Actual result
1. An error occurred when executing ./kata-deploy-binaries-in-docker.sh --build=firecracker
......
[Firecracker devtool] Build successful.
[Firecracker devtool] Firecracker and Jailer binaries placed under /data/gopath/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/all/b
uilddir/firecracker/build/cargo_target/aarch64-unknown-linux-musl/release
[Firecracker devtool] Seccompiler-bin binary placed under /data/gopath/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/all/builddir/f
irecracker/build/seccompiler/aarch64-unknown-linux-musl/release
[Firecracker devtool] Rebase_snap binary placed under /data/gopath/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/all/builddir/firec
racker/build/rebase-snap/aarch64-unknown-linux-musl/release
INFO: Install static firecracker
install: cannot stat 'firecracker/firecracker-static': No such file or directory

2. An error occurred when executing ./kata-deploy-binaries-in-docker.sh --build=virtiofsd
......
cargo: command not found
......
After fixing the above errors, continue to report errors
Compiling libseccomp-sys v0.2.1
   Compiling virtio-bindings v0.1.0
   Compiling termcolor v1.1.3
   Compiling lazy_static v1.4.0
   Compiling humantime v2.1.0
   Compiling textwrap v0.11.0
error: could not find native static library `seccomp`, perhaps an -L flag is missing?

   Compiling aho-corasick v0.7.18
error: could not compile `libseccomp-sys` due to previous error
warning: build failed, waiting for other jobs to finish...

# Further information
[...]

PR #5953, opened 2022-12-23
stable-3.0 | kata-deploy: Fix building the kata static arm package occurred an error

singhwang:
When building the kata static arm package, the two stages of firecracker and virtiofsd report errors.

Fixes: #5952


singhwang:
The modification for firecracker is specially used to solve the error reporting part of firecracker in the issue. This part of the modification also considers the x86 environment.

singhwang:
For the modification of Dockerfile and build-static-virtiofsd.sh, it mainly solves the problem that the build environment is not ready under arm. In the x86 environment, the logic of the script is to directly download the official compiled package, but in the arm environment, it needs to be compiled in the rust environment, refer to the document https://gitlab.com/virtio-fs/virtiofsd , the ubuntu compilation environment needs to install libcap-ng-dev and libseccomp-dev two dependent packages. At the same time, before using cargo in build-static-virtiofsd.sh, the logic of rust installation is added, otherwise an error will be reported here and cannot be found. to the cargo command.

liyimeng:
Similar fix is proposed in https://github.com/kata-containers/kata-containers/pull/5534, but seem not easy to get through :(

singhwang:
> #5534

It seems that we encountered a similar issue. For the modification of build-static-firecracker.sh, our purpose is the same, but I also made corresponding modifications to Dockerfile and build-static-virtiofsd.sh for virtiofsd.

singhwang:
@liyimeng It seems that we encountered a similar issue. For the modification of build-static-firecracker.sh, our purpose is the same, but I also made corresponding modifications to Dockerfile and build-static-virtiofsd.sh for virtiofsd. Unfortunately, reviewers has not given me the opportunity to test.

singhwang:
Can the reviewers trigger the 10 expected checks? Their status has been Expected — Waiting for status to be reported.

singhwang:
The second problem 2 has been officially fixed in version 3.01

Issue #5789, opened 2022-11-29
Refactor how to call iptables command in agent


liubin:
We have some iptables-related changes:

- https://github.com/kata-containers/kata-containers/pull/5612
- https://github.com/kata-containers/kata-containers/pull/5693

These maybe distro related. For new versions, iptables may shift to nftables.

- https://wiki.debian.org/iptables#Current_status
- https://developers.redhat.com/blog/2020/08/18/iptables-the-two-variants-and-their-relationship-with-nftables#

So the iptables command may be a link to either `iptables-legacy` or `iptables-nft`, or the link is not created yet, in this case, we should also check if `iptables-legacy` or `iptables-nft` exists.

Rouzip:
Hi, I am interested in this problem, but I am new in rust, can I have a try?

liubin:
@Rouzip welcome.

Rouzip:
Sorry for the late response.
This issue requires me add check for `iptables-legacy-*` and `iptables-nft-*` in `/usr/sbin` and `/sbin` like this:
```rust
if is ipv6 {
    check /usr/sbin/ip6tables-save
    check /usr/sbin/ip6tables-nft-save
    check /usr/sbin/ip6tables-legacy-save
    check /sbin/ip6tables-save
    check /sbin/ip6tables-nft-save
    check /sbin/ip6tables-legacy-save
} else {
   same logic as ipv6
}
```
Is my understanding of this issue correct?

liubin:
@Rouzip, yes, you are correct.

Rouzip:
I have add the check for iptables-legacy and iptables-nft, but I don't know any better style in Rust. Should I put these string into a array and filter them?

liubin:
@Rouzip you can try to create an PR and talk on the PR.

Rouzip:
> @Rouzip you can try to create an PR and talk on the PR.

Ok, we can talk [there](https://github.com/kata-containers/kata-containers/pull/5956).

Prajna-AS:
Hi @liubin ,
Is this issue still open to work on? I saw there was a closed PR linked to this issue , was wondering if there has been any discussion about the implementation

PR #5956, opened 2022-12-25
agent: add iptable-legacy and iptables-nft check

Rouzip:
Check if iptables-legacy or iptables-nft exists under /usr/sbin and /sbin. Also fix type issue in previous [commit](https://github.com/kata-containers/kata-containers/blob/1511587a9a44226c04d51aba5991bd0e75c8809d/src/agent/src/rpc.rs#L90).

Fixes: #5789
Signed-off-by: Rouzip <1226015390@qq.com>

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -87,12 +87,28 @@ const MODPROBE_PATH: &str = "/sbin/modprobe";
 /// or /usr/sbin, we need to check both of them
 const USR_IPTABLES_SAVE: &str = "/usr/sbin/iptables-save";
 const IPTABLES_SAVE: &str = "/sbin/iptables-save";
-const USR_IPTABLES_RESTORE: &str = "/usr/sbin/iptables-store";
+const USR_IPTABLES_LEGACY_SAVE: &str = "/usr/sbin/iptables-legacy-save";
+const IPTABLES_LEGACY_SAVE: &str = "/sbin/iptables-legacy-save";
+const USR_IPTABLES_NFT_SAVE: &str = "/usr/sbin/iptables-nft-save";
+const IPTABLES_NFT_SAVE: &str = "/sbin/iptables-nft-save";
+const USR_IPTABLES_RESTORE: &str = "/usr/sbin/iptables-restore";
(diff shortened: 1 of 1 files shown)


Rouzip:
I have add the check for iptables-legacy and iptables-nft, but I don't know any better style in Rust. Should I put these string into a array and filter them or any better idea?

Ji-Xinyou:
> I have add the check for iptables-legacy and iptables-nft, but I don't know any better style in Rust. Should I put these string into a array and filter them or any better ideas?

IMO, putting them into array and sequencially scan the binaries would be a good practice. @Rouzip

c3d:
Since we control what we put in the guest image, what is the benefit of supporting a multiplicity of binaries here? I'm concerned about fragility and security risks.

Rouzip:
> > I have add the check for iptables-legacy and iptables-nft, but I don't know any better style in Rust. Should I put these string into a array and filter them or any better ideas?
> 
> IMO, putting them into array and sequencially scan the binaries would be a good practice. @Rouzip

If I use array, maybe I need add a default binary path? And the binary order in this array should have a disscuss?

Ji-Xinyou:
> > > I have add the check for iptables-legacy and iptables-nft, but I don't know any better style in Rust. Should I put these string into a array and filter them or any better ideas?
> > 
> > 
> > IMO, putting them into array and sequencially scan the binaries would be a good practice. @Rouzip
> 
> If I use array, maybe I need add a default binary path? And the binary order in this array should have a disscuss?

I think we need a discuss. AFAIK, for setting and getting the iptables, these binaries work almost the same. But I am not 100% sure about this. But the default should always be `iptables` binary.

Ji-Xinyou:
> Since we control what we put in the guest image, what is the benefit of supporting a multiplicity of binaries here? I'm concerned about fragility and security risks.

The agent is in the guest image instead of the container image. There are some distros that puts the iptables binary irregularly(not exactly this PR is about).

But I agree that this might bring about some fragility and security risks, this needs a further discussion.

Rouzip:
Will we have a default binary in the list? @liubin

Issue #5961, opened 2022-12-28
CC: pulling image by digest fails


snir911:
# Description of problem

image pull of images by digest such as `gcr.io/google-containers/echoserver@sha256:cb5c1bddd1b5665e1867a7fa1b5fa843a47ee433bbb75d4293888b71def53229` fails, although oci distribution spec allows it and the [oci-distribution](https://github.com/krustlet/oci-distribution) crate which is used by image-rs supports it.

# Expected result

image being pulled by the agent

# Actual result

image pull fails

# Further information

https://cloud.google.com/architecture/using-container-image-digests-in-kubernetes-manifests
https://github.com/krustlet/oci-distribution

PR #5962, opened 2022-12-28
CCv0: agent: validate cid using oci-distribution crate

snir911:
which is also used by image-rs, that will allow support in container images digest

Fixes: #5961

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -2385,6 +2385,7 @@ dependencies = [
  "netlink-sys",
  "nix 0.24.3",
  "oci",
+ "oci-distribution",
  "openssl",
  "opentelemetry",
  "procfs 0.12.0",
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -24,6 +24,7 @@ serial_test = "0.5.1"
 kata-sys-util = { path = "../libs/kata-sys-util" }
 kata-types = { path = "../libs/kata-types" }
 url = "2.2.2"
+oci-distribution = "0.9.4"
 
 # Async helpers
 async-trait = "0.1.42"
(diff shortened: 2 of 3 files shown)


fidencio (approved):
lgtm, thanks @snir911!

@mythi, I think this should also be replicated in the enclave-cc code.

stevenhorsman (approved):
I guess this needs a rebase after the rust bump changes, but apart from that it looks great, thanks @snir911!
A question that occurred to me looking at this, that I thought was worth asking is whether the `oci-distribution` crate (and therefore image-rs I guess) can work for an image without a tag, and assume default, or if it's rejected? No worries if you don't know, I can add it to my list!

snir911:
@stevenhorsman it doesn't, is that a problem?
rebased
/test

stevenhorsman:
> @stevenhorsman it doesn't, is that a problem?

Thanks for the info. I don't think it is a problem, it's just potentially something to validate still works with a 'standard' Kubernetes pod descriptor which doesn't specify the full image with tag, but I'm guessing that the CRI resolves before it gets to us. Anyway it sounds like you are just moving existing validation forward from `image-rs` into the `kata-agent`, so won't effect the functionality at the moment.

stevenhorsman:
@snir911 - it looks like we need the new `oci-distribution` dependency updated in the Cargo.lock file as per the static check failure:
```
index c6fb1ec19..2407a201b 100644
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -2385,6 +2385,7 @@ dependencies = [
  "netlink-sys",
  "nix 0.24.3",
  "oci",
+ "oci-distribution",
  "openssl",
  "opentelemetry",
  "procfs 0.12.0",
```

stevenhorsman:
@snir911 - FYI I think we need another look at this. All the tests seem to be failing with rejected container id eg. http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-x86_64-CC_CRI_CONTAINERD_K8S-PR/476/consoleFull has 88 instances of `invalid container ID: whole reference doesn't match`.

Maybe we should add the values of `id` and `reference.whole()` to the error message for debug?

PR #5972, opened 2023-01-03
fix moby prestart hook handling

bergwolf:
Moby relies on the prestart hook to configure its container network endpoints. We should handle it properly so that Kata can once again work with moby after https://github.com/moby/moby/commit/547da0d575519a197fbf4e0543555d508d176084


src/runtime/virtcontainers/sandbox.go
```diff
@@ -1234,9 +1234,24 @@ func (s *Sandbox) startVM(ctx context.Context) (err error) {
 		return err
 	}
 
+	if prestartHookFunc != nil {
+		hid, err := s.GetHypervisorPid()
+		if err != nil {
+			return err
+		}
+		s.Logger().Infof("hypervisor pid is %v", hid)
+		ctx = context.WithValue(ctx, "hypervisor-pid", hid)
```
fidencio, line 1243: static checks are complaining about:
```
Error: sandbox.go:1243:32: SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)
		ctx = context.WithValue(ctx, "hypervisor-pid", hid)
		                             ^
make: *** [Makefile:41: static-checks] Error 1
```

fidencio (commented):
@bergwolf, very nice work!

static checks are complaining a little bit, but that's easy to fix.
I'll give it a try locally with moby and nerdctl and make sure that both are still working as expected before approving!

fidencio:
Please, let's have the tests running meanwhile, just to make sure those are passing with the patch as it is.

studychao:
Should this PR also backport to runtime-rs 's OCI hook implementation? cc @YushuoEdge

bergwolf:
@studychao yep, I was hoping that @YushuoEdge 's runtime-rs hook PR could do similar things. Hopefully runtime-rs should have a cleaner life cycle API to allow running hooks between vm start and netns scanning.

YushuoEdge:
> @studychao yep, I was hoping that @YushuoEdge 's runtime-rs hook PR could do similar things. Hopefully runtime-rs should have a cleaner life cycle API to allow running hooks between vm start and netns scanning.

OK! I'll do a similar modification in runtime-rs's hook implementation!

YushuoEdge:
Hi @bergwolf, runtime-rs+dragonball does not yet support hotplug endpoints, and it may take some time for the complete recan operations to be supported. I think we may fix the issue in runtime-rs after we can hotplug endpoints. (But I will try to make the lifecycle API cleaner before that 😁)

src/runtime/virtcontainers/sandbox.go
```diff
@@ -1234,9 +1237,24 @@ func (s *Sandbox) startVM(ctx context.Context) (err error) {
...
+			return err
+		}
+	}
+
 	// In case of vm factory, network interfaces are hotplugged
 	// after vm is started.
-	if s.factory != nil {
+	// In case of prestartHookFunc, network config might have been changed.
+	// We need to rescan and handle the change.
+	if s.factory != nil || prestartHookFunc != nil {
```
YushuoEdge, line 1257: Seems the prestartHookFunc will never be nil in this function, although there is not necessarily a real “prestart hook” (which means even if the prestart hook is not actually executed, we will always scan the netns once).
I think the program can run without errors, but I'm not sure whether this is as expected.

bergwolf: From VC API point of view, IMO it's better to check for nil `prestartHookFunc` although right now the caller always sets it.

YushuoEdge: Indeed it is.

fidencio:
@bergwolf, may I ask for your containerd configuration as I'm trying to test this locally on my side?

corhere:
@fidencio No containerd configuration is needed to use Kata (or any shimv2 runtime) with Moby. (Configuration is only needed for `cri-containerd`.) `docker run --runtime io.containerd.kata.v2` will Just Work so long as `containerd-shim-kata-v2` is on the containerd process' `PATH`.

bergwolf:
@fidencio As @corhere explained, there is no containerd config required. In my test setup, I was using a fresh install of both moby and containerd with the following command:
```
curl -fsSL https://get.docker.com -o get-docker.sh
sudo CHANNEL=test sh get-docker.sh
```

src/runtime/virtcontainers/network_linux.go
```diff
@@ -252,6 +252,22 @@ func (n *LinuxNetwork) removeSingleEndpoint(ctx context.Context, s *Sandbox, idx
...
 
+func (n *LinuxNetwork) endpointAlreadyAdded(netInfo *NetworkInfo) bool {
+	for _, ep := range n.eps {
+		// Existing endpoint
+		if ep.Name() == netInfo.Iface.Name {
+			return true
+		}
+		pair := ep.NetworkPair()
+		// Existing virtual endpoints
+		if pair != nil && (pair.TapInterface.Name == netInfo.Iface.Name || pair.TapInterface.TAPIface.Name == netInfo.Iface.Name) {
```
liubin, line 263: Should compare `netInfo.Iface.Name` with `pair.TapInterface.Name`/`pair.TapInterface.TAPIface.Name`, or `netPair.VirtIface.Name`? 

Through https://github.com/kata-containers/kata-containers/blob/a55eb78c32e7461161b77b0663b2a737f4a83fba/src/runtime/virtcontainers/network.go#L354-L356 , the `netPair.VirtIface.Name`'s value will be set using the value of `netInfo.Iface.Name`.

bergwolf: Let's match them all as I got `netInfo.Iface.Name == tap0_kata` here when testing.

fidencio (approved):
lgtm, thanks @bergwolf!

fidencio:
> @fidencio As @corhere explained, there is no containerd config required. In my test setup, I was using a fresh install of both moby and containerd with the following command:
> 
> ```
> curl -fsSL https://get.docker.com -o get-docker.sh
> sudo CHANNEL=test sh get-docker.sh
> ```

What I ended up hitting was an agent crash not related to this work, and that drove me crazy till I figure out what was happening.

I finally was able to test using an older rootfs, and everything works as expected.  Thanks a whole lot for this work and I cannot believe we're finally back to using docker with Kata Containers, that's simply freaking amazing! :-)

fidencio:
> What I ended up hitting was an agent crash not related to this work, and that drove me crazy till I figure out what was happening.

And I cannot reproduce the issue when running this patch rebased atop of main.
We're good to go with this one, IMHO.  Again, thanks for the work @bergwolf!

Issue #5890, opened 2022-12-13
agent: foreign `CC` incompatible with libgit2


jtumber-ibm:
# Description of problem

Related to the CCv0 branch but better to fix in main.

When cross-compiling the `kata-agent` the process is to specify the `ARCH` and `CC` for the required target. i.e. `s390x` and `s390x-linux-gnu-gcc`. This will setup the required `RUSTFLAGS` to specify the linker. However doing so break the build process.

# Expected result

kata-agent IBM S/390 binary 

# Actual result

Build fails to compile `image-rs` and `attestation-agent` (CCv0). With errors such as `error adding symbols: file in wrong format`. 

# Further information

This is due to specifying a foreign `CC` which is incompatible with libgit2. 

I will submit a PR to resolve this.

PR #5977, opened 2023-01-04
CC | agent: unset `CC` for cross-build

stevenhorsman:
When `HOST_ARCH` != `ARCH` unset `CC`

Specifying a foreign CC is incompatible with building libgit2. Thus after the RUSTFLAGS linker has been set we can safely unset CC to avoid passing this value through the build.

Fixes: #5890

Signed-off-by: James Tumber <james.tumber@ibm.com>
Cherry-picked: 087515a


fidencio (approved):
lgtm, thanks @stevenhorsman!

Issue #5985, opened 2023-01-05
nydusd: Make netns operations buildable on !linux


dcantah:
To make virtcontainers buildable on darwin/other plats we need to not do net namespace operations in virtcontainers with no buildtags/stubs for other platforms. 

This is a continuation of splitting up https://github.com/kata-containers/kata-containers/pull/4296/commits into individual PRs.

PR #5986, opened 2023-01-05
nydus: net-ns handling needs to be only executed on Linux hosts

dcantah:
Fixes: #5985

With nydus not being its own pkg, it is challenging to implement cleanly in a virtcontainers package that isn't necesarily Linux-only. The existing code utilizes network namespace code in order to ensure nydus is launched in the host netns. This is very Linux specific - so let's make sure we only carry this out in a linux specific file.

In the Darwin case, to allow for compilation at least, let's add a stub for doNetNS. Ideally the nydus and vc code can be refactored / decoupled.

### DIFF
--- a/src/runtime/virtcontainers/nydusd_linux.go
+++ b/src/runtime/virtcontainers/nydusd_linux.go
@@ -1,0 +1,21 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package virtcontainers
+
+import (
+	"os/exec"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+)
+
+const shimNsPath = "/proc/self/ns/net"
+
+func startInShimNS(cmd *exec.Cmd) error {
+	// Create nydusd in shim netns as it needs to access host network
+	return doNetNS(shimNsPath, func(_ ns.NetNS) error {
+		return cmd.Start()
+	})
+}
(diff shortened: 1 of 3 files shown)


dcantah:
Probably make an issue/PR to try and get nydusd into its own package.. That's mostly orthogonal to the getting darwin/other platforms to build work, but it's very odd (albeit I haven't looked at this codebase in awhile) that it's shoved in virtcontainers.

src/runtime/virtcontainers/nydusd_linux.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2017 Intel Corporation
```
liubin, line 1: This file and `nydusd_other.go` have different copyright headers.

dcantah: The other file was just work moved to a new spot, does that constitute a new header?

egernst: agreed.  I think what Danny has here makes sense @liubin.  _other is a new file, whereas _linux is just a move of existing code. I would follow same pattern?

src/runtime/virtcontainers/nydusd_other.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2022 Apple Inc.
```
egernst, line 1: happy new year?

dcantah: LOL, we talked about the same in https://github.com/kata-containers/kata-containers/pull/5984#discussion_r1062670516

dcantah: Need at least two weeks to adjust to there being a 3

src/runtime/virtcontainers/nydusd_linux.go
liubin: @dcantah @egernst Ok.

egernst (approved):

liubin (approved):

Issue #6002, opened 2023-01-08
virtcontainers: fs_share skeleton for Darwin 


dcantah:
Continuing work to get the runtime building on Darwin https://github.com/kata-containers/kata-containers/pull/4296/commits.

This tracks getting a skeleton up of the [fs_share functionality](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/fs_share_linux.go) so we can at least build on Darwin

PR #6003, opened 2023-01-08
virtcontainers: fs_share: Add Darwin skeleton

dcantah:
Fixes: #6002

As a first pass for testing, let's add a skeleton for filesystem sharing support on Darwin.


dcantah:
cc @mxpv

egernst (approved):

PR #6009, opened 2023-01-09
Dragonball: add cpu resize ability

studychao:
Add cpu resize ability upon upcall communication channel. Runtime could use ResizeVcpu VmmAction and pass the desired vCPU number to the Dragonball hypervisor.
Dragonball will trigger the device manager service in guest kernel's upcall server to do cpu resize.

fixes: kata-containers#6008

### DIFF
--- a/src/dragonball/src/vcpu/vcpu_impl.rs
+++ b/src/dragonball/src/vcpu/vcpu_impl.rs
@@ -214,10 +214,20 @@ pub enum VcpuResponse {
     CacheRevalidated,
 }
 
+#[derive(Debug, PartialEq)]
+/// Vcpu Hotplug Result returned from the guest
+pub enum VcpuResizeResult {
+    /// All vCPU hotplug / hot-unplug operations are successful
+    Success = 0,
+    /// vCPU hotplug / hot-unplug failed
+    Failed = 1,
+}
+
 /// List of events that the vcpu_state_sender can send.
 pub enum VcpuStateEvent {
-    /// (result, response) for hotplug, result 0 means failure, 1 means success.
-    Hotplug((i32, u32)),
+    /// (result, response) for hotplug / hot-unplugged.
+    /// response records how many cpu has successfully being hotplugged / hot-unplugged.
+    Hotplug((VcpuResizeResult, u32)),
 }
 
 /// Wrapper over vCPU that hides the underlying interactions with the vCPU thread.
--- a/src/dragonball/src/vm/mod.rs
+++ b/src/dragonball/src/vm/mod.rs
@@ -4,6 +4,7 @@
 use std::io::{self, Read, Seek, SeekFrom};
 use std::ops::Deref;
 use std::os::unix::io::RawFd;
+
 use std::sync::{Arc, Mutex, RwLock};
 
 use dbs_address_space::AddressSpace;
@@ -22,6 +23,8 @@ use vmm_sys_util::eventfd::EventFd;
 
 #[cfg(all(feature = "hotplug", feature = "dbs-upcall"))]
 use dbs_upcall::{DevMgrService, UpcallClient};
+#[cfg(feature = "hotplug")]
+use std::sync::mpsc::Sender;
 
 use crate::address_space_manager::{
     AddressManagerError, AddressSpaceMgr, AddressSpaceMgrBuilder, GuestAddressSpaceImpl,
@@ -35,6 +38,8 @@ use crate::event_manager::EventManager;
 use crate::kvm_context::KvmContext;
 use crate::resource_manager::ResourceManager;
 use crate::vcpu::{VcpuManager, VcpuManagerError};
+#[cfg(feature = "hotplug")]
+use crate::vcpu::{VcpuResizeError, VcpuResizeInfo};
 #[cfg(target_arch = "aarch64")]
 use dbs_arch::gic::Error as GICError;
 
@@ -795,7 +800,30 @@ impl Vm {
         } else if self.is_upcall_client_ready() {
             Ok(DeviceOpContext::create_hotplug_ctx(self, epoll_mgr))
         } else {
-            Err(StartMicroVmError::UpcallNotReady)
+            Err(StartMicroVmError::UpcallServerNotReady)
+        }
+    }
+
+    /// Resize MicroVM vCPU number
+    #[cfg(feature = "hotplug")]
+    pub fn resize_vcpu(
+        &mut self,
+        config: VcpuResizeInfo,
+        sync_tx: Option<Sender<bool>>,
+    ) -> std::result::Result<(), VcpuResizeError> {
+        if self.upcall_client().is_none() {
+            Err(VcpuResizeError::UpcallClientMissing)
+        } else if self.is_upcall_client_ready() {
+            if let Some(vcpu_count) = config.vcpu_count {
+                self.vcpu_manager()
+                    .map_err(VcpuResizeError::Vcpu)?
+                    .resize_vcpu(vcpu_count, sync_tx)?;
+
+                self.vm_config.vcpu_count = vcpu_count;
+            }
+            Ok(())
+        } else {
+            Err(VcpuResizeError::UpcallServerNotReady)
         }
     }
 
--- a/src/dragonball/src/api/v1/vmm_action.rs
+++ b/src/dragonball/src/api/v1/vmm_action.rs
@@ -35,6 +35,9 @@ pub use crate::device_manager::virtio_net_dev_mgr::{
 #[cfg(feature = "virtio-vsock")]
 pub use crate::device_manager::vsock_dev_mgr::{VsockDeviceConfigInfo, VsockDeviceError};
 
+#[cfg(feature = "hotplug")]
+pub use crate::vcpu::{VcpuResizeError, VcpuResizeInfo};
+
 use super::*;
 
 /// Wrapper for all errors associated with VMM actions.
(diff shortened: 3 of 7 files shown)


src/dragonball/src/vm/mod.rs
```diff
@@ -799,6 +804,30 @@ impl Vm {
         }
     }
 
+    /// Resize MicroVM vCPU number
+    #[cfg(feature = "hotplug")]
+    pub fn resize_vcpu(
+        &mut self,
+        config: VcpuResizeInfo,
+        sync_tx: Option<Sender<bool>>,
```
YushuoEdge, line 812: I am curious about how this parameter(sync_tx) works? 
( since now we always get a `None`, and we just pass it to vcpu_manager.resize_vcpu(..))

YushuoEdge (approved):
LGTM except for one small question (

src/dragonball/src/vm/mod.rs
studychao: Right now we are using async way to do cpu resize, later we could add sync way to do cpu resize through this Sender

studychao:
This is a joint work with @Ji-Xinyou

YushuoEdge:
@studychao @Ji-Xinyou Thanks for the nice work! I think besides on container start, we should also update cpu/mem resources on container update (? Could you add these code altogether to this PR?

YushuoEdge (commented):
@studychao @Ji-Xinyou Thanks for the nice work! I think besides on container start, we should also update cpu/mem resources on container update (? Could you add these code altogether to this PR?

src/runtime-rs/crates/resource/src/cpu_mem/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Cpu information wrapper
+#[derive(Default, Debug, Clone, Copy)]
+pub struct CpuResource {
+    /// The default vcpu # when boot
+    default_vcpu: i32,
+
+    /// Current number of vCPUs
+    /// when boot, the number is set to *default_vcpus*
+    current_vcpu: i32,
+    // TODO(shuoyu.ys): cpuset mapping relation
```
YushuoEdge, line 27: Maybe we should delete this comment at this time 🙈

src/runtime-rs/crates/resource/src/cpu_mem/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+            default_vcpu: hypervisor_config.cpu_info.default_vcpus,
+            current_vcpu: hypervisor_config.cpu_info.default_vcpus,
+        })
+    }
+}
+
+/// Memory information wrapper
+#[derive(Default, Debug, Clone, Copy)]
+pub struct MemResource {
+    /// The default vcpu # when boot
```
YushuoEdge, line 47: The default memory size, in MiB

Ji-Xinyou:
> @studychao @Ji-Xinyou Thanks for the nice work! I think besides on container start, we should also update cpu/mem resources on container update (? Could you add these code altogether to this PR?

I am not seeing the `update` operation in the containerd's Request, i.e. in the `Request` enum.

YushuoEdge:
> > @studychao @Ji-Xinyou Thanks for the nice work! I think besides on container start, we should also update cpu/mem resources on container update (? Could you add these code altogether to this PR?
> 
> I am not seeing the `update` operation in the containerd's Request, i.e. in the `Request` enum.

Here we can find the request:
https://github.com/kata-containers/kata-containers/blob/40c69043245d0326080f48b2d295d31eee52e385/src/runtime/vendor/github.com/containerd/containerd/runtime/v2/task/shim.pb.go#L620-L627

And there is a UpdateContainer in VCSandbox interface in runtime. (If we have not implement this interface in runtime-rs, maybe we can open another issue.)

src/runtime-rs/crates/resource/src/cpu_mem/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+            default_mem_mb: hypervisor_config.memory_info.default_memory,
+            current_mem_mb: hypervisor_config.memory_info.default_memory,
+            enable_guest_swap: hypervisor_config.memory_info.enable_guest_swap,
+        })
+    }
+}
+
+/// Generic sandbox Cpu and Memory information wrapper, used as a query intermediate
+#[derive(Default, Debug, Clone, Copy)]
+pub struct CpuMemResource {
```
Tim-0731-Hzt, line 75: just ooi, Why the CPU and memory resources should be put together? Why we cannot do like this
```
__resource
|__src
     |__cpu
     |__mem
```

src/runtime-rs/crates/resource/src/cpu_mem/mod.rs
Ji-Xinyou: I agree 😄

Ji-Xinyou:
> > > @studychao @Ji-Xinyou Thanks for the nice work! I think besides on container start, we should also update cpu/mem resources on container update (? Could you add these code altogether to this PR?
> > 
> > 
> > I am not seeing the `update` operation in the containerd's Request, i.e. in the `Request` enum.
> 
> Here we can find the request:
> 
> https://github.com/kata-containers/kata-containers/blob/40c69043245d0326080f48b2d295d31eee52e385/src/runtime/vendor/github.com/containerd/containerd/runtime/v2/task/shim.pb.go#L620-L627
> 
> And the service in runtime: https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/containerd-shim-v2/service.go#L1011-L1043
> 
> And we can find the service in runtime-rs: https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs#L248-L257

You are right, this should be fixed.

src/runtime-rs/crates/resource/src/cpu_mem/mod.rs
Ji-Xinyou: Normally the resources of cpu and memory are considered at the same time, splitting them IMO makes no difference. Providing separate APIs for querying cpu resource and memory resource will be enough.

studychao:
As discussed with @Ji-Xinyou , he will follow up the implementation of runtime part in other PR. So this PR will only focus on the hypervisor part, and after this PR merged, @Ji-Xinyou could continue the work on runtime.

issue #6159 is for following up the runtime implementation of cpu resize.

liubin (approved):

src/dragonball/src/vcpu/vcpu_impl.rs
```diff
@@ -216,7 +216,7 @@ pub enum VcpuResponse {
 
 /// List of events that the vcpu_state_sender can send.
 pub enum VcpuStateEvent {
-    /// (result, response) for hotplug, result 0 means failure, 1 means success.
+    /// (result, response) for hotplug, result 0 means success, 1 means failure.
     Hotplug((i32, u32)),
```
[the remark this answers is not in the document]
studychao: Thanks, added.

Tim-0731-Hzt (approved):

src/dragonball/src/vcpu/vcpu_impl.rs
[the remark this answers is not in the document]
Ji-Xinyou: I would suggest adding this
```Rust
const HotplugSuccess: i32 = 0;
const HotplugFailed: i32 = 1;
```

src/dragonball/src/vm/mod.rs
```diff
@@ -795,8 +800,32 @@ impl Vm {
...
+        }
+    }
+
+    /// Resize MicroVM vCPU number
+    #[cfg(feature = "hotplug")]
+    pub fn resize_vcpu(
+        &mut self,
+        config: VcpuResizeInfo,
+        sync_tx: Option<Sender<bool>>,
+    ) -> std::result::Result<(), VcpuResizeError> {
```
Ji-Xinyou, line 813: Nit, whynot:
```Rust
if self.upcall_client().is_none() {
        return Err(VcpuResizeError::UpcallClientMissing)
} else if self.is_upcall_client_ready() {
        if let Some(vcpu_count) = config.vcpu_count {
                self.vcpu_manager()
                    .map_err(VcpuResizeError::Vcpu)?
                    .resize_vcpu(vcpu_count, sync_tx)?;

                self.vm_config.vcpu_count = vcpu_count;
                Ok(())
        }
} else {
        Err(VcpuResizeError::UpcallServerNotReady)
}
```

Ji-Xinyou (approved):
Generally LGTM! Some nits below

src/dragonball/src/vcpu/vcpu_impl.rs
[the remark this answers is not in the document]
liubin: Normally returning value 0 means success, is defining it as const unnecessary?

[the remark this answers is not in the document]
Ji-Xinyou: @liubin, I think this is a nit, not defining those consts will definitely be fine. My opinions are as follows.

In C I would say this would be unnecessary, since that is the convention.

But here we have already put the `Hotplug` as a product type of i32 and u32, Things like
```Rust
if let Hotplug((x, y)) == some_func_returns_hotplug() {
        match x {
            0 => handle_success(),
            1 => handled_failure(),
        }
}
```
would be hard to read, and prone to errors(by developers) if multiple functions use this.

Moreover, if all `VcpuStateEvent`s are gonna use 0 and 1 as sentinel values, `VcpuStateEventSuccess` would be a better name.

src/dragonball/src/vm/mod.rs
studychao: thanks, changed.

studychao:
@liubin @Ji-Xinyou Thanks for your suggestions. For the `response` part, I have created an enum called `VcpuResizeResult` to indicate the result status instead of only using i32 to indicate.

Issue #6016, opened 2023-01-10
runtime-rs: use hypervisor.enable_debug to control log output


YushuoEdge:
Currently, when users want kata-agent's log and guest kernel's log, they need to append kernel_params in the configuration file. However, whether to output these kinds of logs should be controlled by [hypervisor].enable_debug.

This issue can be regarded as a sub-issue of https://github.com/kata-containers/kata-containers/issues/6000.

PR #6017, opened 2023-01-10
runtime-rs: use hypervisor.enable_debug to control log output

YushuoEdge:
Currently, when users want kata-agent's log and guest kernel's log with runtime-rs+dragonball, they need to append kernel_params(agent.log_vport=1025 console=ttyS1) in the configuration file.

Now we change to use hypervisor.enable_debug to control the log output.

Fixes: #6016

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
@@ -6,8 +6,8 @@
 
 use super::vmm_instance::VmmInstance;
 use crate::{
-    device::Device, hypervisor_persist::HypervisorState, kernel_param::KernelParams, VmmState,
-    DEV_HUGEPAGES, HUGETLBFS, HYPERVISOR_DRAGONBALL, SHMEM, VM_ROOTFS_DRIVER_BLK,
+    device::Device, hypervisor_persist::HypervisorState, kernel_param::KernelParams, Param,
+    VmmState, DEV_HUGEPAGES, HUGETLBFS, HYPERVISOR_DRAGONBALL, SHMEM, VM_ROOTFS_DRIVER_BLK,
 };
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
@@ -18,7 +18,10 @@ use dragonball::{
 use kata_sys_util::mount;
 use kata_types::{
     capabilities::{Capabilities, CapabilityBits},
-    config::hypervisor::Hypervisor as HypervisorConfig,
+    config::{
+        default::DEFAULT_AGENT_LOG_PORT, hypervisor::Hypervisor as HypervisorConfig,
+        LOG_VPORT_OPTION,
+    },
 };
 use persist::sandbox_persist::Persist;
 use shim_interface::KATA_PATH;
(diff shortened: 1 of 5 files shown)


src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -100,11 +103,22 @@ impl DragonballInner {
...
-        let mut kernel_params = KernelParams::new(self.config.debug_info.enable_debug);
+        let mut kernel_params = KernelParams::new();
         kernel_params.append(&mut KernelParams::new_rootfs_kernel_params(&rootfs_driver));
         kernel_params.append(&mut KernelParams::from_string(
             &self.config.boot_info.kernel_params,
         ));
+
+        if self.config.debug_info.enable_debug {
+            // export guest kernel log
+            kernel_params.push(Param::new("console", "ttyS1"));
```
Tim-0731-Hzt, line 114: suggest adding const

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
```diff
@@ -60,10 +55,6 @@ impl KernelParams {
             Param::new("systemd.mask", "systemd-networkd.service"),
         ];
 
-        if debug {
-            params.push(Param::new(LOG_VPORT_OPTION, VSOCK_LOGS_PORT));
```
Tim-0731-Hzt, line 64: I think it's better to export the guest kernel log here. This function is used as the util function for the kernel parameters.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
YushuoEdge: I have considered not to make them const value for the following reason:

- console and ttyS1 seems to be used only once
- there is no special meaning to this two string, if we add const, we just change the letter case

If you think that does not make sense, or we should always apply const for a const string, I will fix it.

Tim-0731-Hzt: fair enough

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
YushuoEdge: Yes, as I have thought that `agent.log_vport` will not used for all hypervisor/agent, so I move this branch logic to dragonball's implementation to make KernelParams works as a more independent module.

If we still want to export guest kernel log here, we should at least determine which vmm is currently in use. And as KernelParam can not fetch the config, we may need to add a parameter to this function. Thus it would be like:
in KernelParam:
```
pub(crate) fn new(debug: bool, hypervisor: &str) -> self {
	...
	if debug {
		if hypervisor == "dragonball" {
			...
		}
	}
}
```

in DragonballInner:
```
	...
	KernelParam::new(config.enable_debug, "dragonball");
	...
```

Tim-0731-Hzt: ye, I see, I just check the `go runtime`, `agent.log_vport` will only work for `firecracker`.

Issue #6021, opened 2023-01-10
networking: With new prestartHookFunc scanning, internetworking model of none causes container to fail to start


stevenhorsman:
I'm not 100% sure about this problem's reproduce in 'vanilla' kata-containers, but we've hit this error in peer pod https://github.com/confidential-containers/cloud-api-adaptor/issues/505, with the following logs
```
Jan 10 08:34:29 ibm-caa-worker kata[261107]: time="2023-01-10T08:34:29.773508281Z" level=info msg="calling remote hypervisor StartVM (timeout: 600)"
Jan 10 08:35:12 ibm-caa-worker kata[261107]: time="2023-01-10T08:35:12.701441048Z" level=info msg="hypervisor pid is 261107" name=containerd-shim-v2 pid=261107 sandbox=81a3b027a709db0549e4ac7d0cae84cd8f3e909a3d71895c97a753170091c77f source=virtcontainers subsystem=sandbox
Jan 10 08:35:12 ibm-caa-worker kata[261107]: time="2023-01-10T08:35:12.702363934Z" level=info msg="veth interface found" name=containerd-shim-v2 pid=261107 sandbox=81a3b027a709db0549e4ac7d0cae84cd8f3e909a3d71895c97a753170091c77f source=virtcontainers subsystem=network
Jan 10 08:35:12 ibm-caa-worker kata[261107]: time="2023-01-10T08:35:12.702417759Z" level=info msg="Attaching endpoint" endpoint-type=virtual hotplug=true name=containerd-shim-v2 pid=261107 sandbox=81a3b027a709db0549e4ac7d0cae84cd8f3e909a3d71895c97a753170091c77f source=virtcontainers subsystem=network
Jan 10 08:35:12 ibm-caa-worker kata[261107]: time="2023-01-10T08:35:12.702458597Z" level=error msg="Error bridging virtual ep" error="Invalid internetworking model" name=containerd-shim-v2 pid=261107 sandbox=81a3b027a709db0549e4ac7d0cae84cd8f3e909a3d71895c97a753170091c77f source=virtcontainers subsystem=network
```

It's in sandbox networking, so way outside of my comfort zone, but I think the issue boils down to:
- https://github.com/kata-containers/kata-containers/commit/578a9c25f05c53525319a968f379ca718e8c9db7 added:
```
// We need to rescan and handle the change.
	if s.factory != nil || prestartHookFunc != nil {
```
[...]

PR #6022, opened 2023-01-10
network: Handle InterworkingModel of none

stevenhorsman:
- Add a case to handle `NetXConnectNoneModel` in `xConnectVMNetwork`

Fixes: #6021

### DIFF
--- a/src/runtime/virtcontainers/network_linux.go
+++ b/src/runtime/virtcontainers/network_linux.go
@@ -564,6 +564,8 @@ func xConnectVMNetwork(ctx context.Context, endpoint Endpoint, h Hypervisor) err
 	case NetXConnectTCFilterModel:
 		networkLogger().Info("connect TCFilter to VM network")
 		err = setupTCFiltering(ctx, endpoint, queues, disableVhostNet)
+	case NetXConnectNoneModel:
+		// internetworking_model is none, so do nothing?
 	default:
 		err = fmt.Errorf("Invalid internetworking model")
 	}


stevenhorsman:
@bergwolf - would you be able to take a look at this (and the issue). I'm in over my head here, but I thought I'd try and see if I could help with the problem. I suspect there might be a much better way to handle this, but in case there isn't I can also add support for `NetXConnectNoneModel ` into `xDisconnectVMNetwork`?

Thanks!

Jordan9500:
I have tested these changes on my worker. However, I am instead getting this error: 
```
  Warning  FailedCreatePodSandBox  1s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: Unsupported network interface: vxlan: unknown
```
Linked to this line of code: https://github.com/kata-containers/kata-containers/blob/0ec4aa1a86269f4cedcc637321fddc7da641cea9/src/runtime/virtcontainers/network_linux.go#L171

stevenhorsman:
Peer pods does use `vxlan`, but that feels like a bigger leap to add a dead if case for, so I think stopping it running at a higher level might be needed? It also looked like in the previous path we were going into the veth interface type before the change I made, so I've definitely proven that I don't understand what is going on! Maybe there are different endpoints with multiple interface types we need to handle differently?

stevenhorsman:
Now we have an expert looking at a better solution, I won't waste much more time on this PR

Issue #6033, opened 2023-01-11
CC | packaging: Take toolchain version when using the td-shim builder image


fidencio:
As the toolchain is installed in the image itself, we *must* take the
 toolchain into consideration when deciding whether to use a cached image
 or building a new one.

PR #6034, opened 2023-01-11
CC | packaging: Take toolchain version when using the td-shim builder image

fidencio:
As the toolchain is installed in the image itself, we *must* take the toolchain into consideration when deciding whether to use a cached image or building a new one.

Fixes: #6033


stevenhorsman (approved):
LGTM! Thanks

PR #6035, opened 2023-01-11
tools: add --locked option for cargo install

liubin:
There is a broken release of cgroup-rs, but cargo install will not use the version in Cargo.lock, so add the `--locked` option to use the version specified in the Cargo.toml

Fixes: #5376


Tim-Zhang (approved):
LGTM, thanks @liubin

PR #6037, opened 2023-01-11
runtime: fix up disable_netns handling

bergwolf:
With `disable_netns=true`, we should never scan the sandbox netns which is the host netns in such case.

Fixes: #6021


liubin (approved):

dcantah (approved):

stevenhorsman (approved):
LGTM. Thanks for the quick fix @bergwolf!

amshinde (approved):

Issue #6059, opened 2023-01-12
runtime: all APIs are hang in the service.mu


ls-ggg:
# Description of problem
Deleting the container timed out, the containerd log is as follows:
{"error":"context deadline exceeded: unknown","level":"error","msg":"get state for $containerId","time":""}

After checking, it is found that all APIs will be hang in the mutex(s.mu.Lock())

PR #6060, opened 2023-01-12
 runtime:all APIs are hang in the service.mu

ls-ggg:
When the vmm process exits abnormally, a goroutine sets s.monitor
to null in the 'watchSandbox' function without getting service.mu,
This will cause another goroutine to block when sending a message
to s.monitor, and it holds service.mu, which leads to a deadlock.
For example, the wait function in the file
.../pkg/containerd-shim-v2/wait.go will send a message to s.monitor
after obtaining service.mu, but s.monitor may be null at this time

 Fixes: #6059

Signed-off-by: ls <335814617@qq.com>

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/wait.go
+++ b/src/runtime/pkg/containerd-shim-v2/wait.go
@@ -120,10 +120,12 @@ func watchSandbox(ctx context.Context, s *service) {
 	if err == nil {
 		return
 	}
-	s.monitor = nil
 
 	s.mu.Lock()
 	defer s.mu.Unlock()
+
+	s.monitor = nil
+
 	// sandbox malfunctioning, cleanup as much as we can
 	shimLog.WithError(err).Warn("sandbox stopped unexpectedly")
 	err = s.sandbox.Stop(ctx, true)


ls-ggg:
For example, the wait function in the file .../runtime/pkg/containerd-shim-v2/wait.go will send a message to s.monitor after obtaining service.mu, but s.monitor may be null at this time

src/runtime/pkg/containerd-shim-v2/wait.go
```diff
@@ -120,7 +120,6 @@ func watchSandbox(ctx context.Context, s *service) {
 	if err == nil {
 		return
 	}
-	s.monitor = nil
```
fengwang666, line 123: I'm not sure if there is any unintended side effect here caused by removing line 123. Why not acquire the lock before setting s.monitor to nil? Seems a less invasive change to me.

ls-ggg: Setting s.monitor to null after acquiring the lock can indeed solve the problem, but I think s.monitor should be released correctly in sandbox.Stop。

> I'm not sure if there is any unintended side effect here caused by removing line 123. @bergwolf probably has more context. Why not acquire the lock before setting s.monitor to nil? Seems a less invasive change to me.

ls-ggg: I've stress-tested this commit in our test environment and haven't seen any issues so far.

ls-ggg: @fengwang666 Thanks, I made a mistake and I have changed to set s.monitor to null after acquiring the lock

liubin:
Thank you @ls-ggg, can you squash your two commits into one and make sure that the commit line length is under 70 for reading?

You can refer to https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

ls-ggg:
> Thank you @ls-ggg, can you squash your two commits into one and make sure that the commit line length is under 70 for reading?
> 
> You can refer to https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

@liubin Thanks, I have merged two commits into one, and the commit line length is less than 70.

fengwang666 (approved):
LGTM. Thanks!

bergwolf (approved):
Good catch! `s.monitor` is not supposed to be protected by `s.mu`, but we ended up having to access it with `s.mu` at hand. So your fix is definitely needed. Thanks!

Issue #6071, opened 2023-01-13
Can't install kata by kata-manager.sh with installed containerd


darfux:
# Description of problem

Under an environment with containerd installed,
run `kata-manager.sh` with the `only_kata` option:
```
$ bash kata-manager.sh -o
```

# Expected result

Install kata along with containerd successfully

# Actual result

pre-checks fails with:
```
INFO: Running pre-checks
ERROR: containerd already installed
```

PR #6072, opened 2023-01-13
utils:kata-manager: Don't check containerd when `only_kata` is enabled

darfux:
Fixes: #6071

Signed-off-by: Li Yuxuan <liyuxuan.darfux@bytedance.com>

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -269,16 +269,20 @@ containerd_installed()
 
 pre_checks()
 {
+	local skip_containerd="${1:-}"
+
 	info "Running pre-checks"
 
 	command -v "${kata_shim_v2}" &>/dev/null \
 		&& die "Please remove existing $kata_project installation"
 
-	local ret
+	if [ "$skip_containerd" != "true" ]; then
+		local ret
 
-	{ containerd_installed; ret=$?; } || true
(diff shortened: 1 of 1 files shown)


utils/kata-manager.sh
```diff
@@ -269,16 +269,20 @@ containerd_installed()
 
 pre_checks()
 {
+	local only_kata="${1:-}"
```
amshinde, line 272: Maybe  use `skip-containerd` here instead of `only_kata` to make things clearer.

utils/kata-manager.sh
```diff
@@ -269,16 +269,20 @@ containerd_installed()
...
 
 	command -v "${kata_shim_v2}" &>/dev/null \
 		&& die "Please remove existing $kata_project installation"
 
-	local ret
+	if [ "$only_kata" != "true" ]; then
+		local ret
 
-	{ containerd_installed; ret=$?; } || true
+		{ containerd_installed; ret=$?; } || true
```
amshinde, line 282: Check the use of space here instead of tab.

darfux: Here is tab actually 🧐 I'm not sure why it was highlighted

darfux:
Hi @amshinde , I've resolved the conflicts. PTAL😄

Issue #5975, opened 2023-01-04
runtime-rs doesn't cleanup the /run/kata-containers/shared/sandboxes/[pid] after destoryed


studychao:
I have tried this command 
```
ctr run --runtime=io.containerd.kata.v2 --rm docker.io/library/busybox:latest kata-test-2002 uname -r
```

After the Kata is destoryed, the `/run/kata-containers/shared/sandboxes/[pid]` still exists.

![image](https://user-images.githubusercontent.com/31121654/210556209-89e2b068-19a0-482d-b5c5-8a1e38ff3286.png)

PR #6080, opened 2023-01-16
runtime-rs: cleanup kata host share path

Tim-0731-Hzt:
cleanup the /run/kata-containers/shared/sandboxes/pid path

Fixes:#5975

### DIFF
--- a/src/runtime-rs/crates/resource/src/manager.rs
+++ b/src/runtime-rs/crates/resource/src/manager.rs
@@ -101,9 +101,9 @@ impl ResourceManager {
         inner.update_cgroups(cid, linux_resources).await
     }
 
-    pub async fn delete_cgroups(&self) -> Result<()> {
+    pub async fn cleanup(&self) -> Result<()> {
         let inner = self.inner.read().await;
-        inner.delete_cgroups().await
+        inner.cleanup().await
     }
 }
 
--- a/src/runtime-rs/crates/resource/src/manager_inner.rs
+++ b/src/runtime-rs/crates/resource/src/manager_inner.rs
@@ -8,7 +8,7 @@ use std::{sync::Arc, thread};
 
 use crate::resource_persist::ResourceState;
 use agent::{Agent, Storage};
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, Context, Ok, Result};
 use async_trait::async_trait;
 use hypervisor::Hypervisor;
 use kata_types::config::TomlConfig;
@@ -233,8 +233,22 @@ impl ResourceManagerInner {
             .await
     }
 
-    pub async fn delete_cgroups(&self) -> Result<()> {
-        self.cgroups_resource.delete().await
+    pub async fn cleanup(&self) -> Result<()> {
+        // clean up cgroup
+        self.cgroups_resource
+            .delete()
+            .await
+            .context("delete cgroup")?;
+        // clean up share fs mount
+        if let Some(share_fs) = &self.share_fs {
+            share_fs
+                .get_share_fs_mount()
+                .cleanup(&self.sid)
+                .await
+                .context("failed to cleanup host path")?;
+        }
+        // TODO cleanup other resources
+        Ok(())
     }
 
     pub async fn dump(&self) {
--- a/src/runtime-rs/crates/resource/src/share_fs/mod.rs
+++ b/src/runtime-rs/crates/resource/src/share_fs/mod.rs
@@ -131,6 +131,8 @@ pub trait ShareFsMount: Send + Sync {
     async fn umount_volume(&self, file_name: &str) -> Result<()>;
     /// Umount the rootfs
     async fn umount_rootfs(&self, config: &ShareFsRootfsConfig) -> Result<()>;
+    /// Clean up share fs mount
+    async fn cleanup(&self, sid: &str) -> Result<()>;
 }
 
 pub fn new(id: &str, config: &SharedFsInfo) -> Result<Arc<dyn ShareFs>> {
--- a/src/runtime-rs/crates/resource/src/share_fs/utils.rs
+++ b/src/runtime-rs/crates/resource/src/share_fs/utils.rs
@@ -59,6 +59,10 @@ pub fn get_host_rw_shared_path(sid: &str) -> PathBuf {
     Path::new(KATA_HOST_SHARED_DIR).join(sid).join("rw")
 }
 
+pub fn get_host_shared_path(sid: &str) -> PathBuf {
+    Path::new(KATA_HOST_SHARED_DIR).join(sid)
+}
+
 fn do_get_guest_any_path(
     target: &str,
     cid: &str,
--- a/src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
+++ b/src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
@@ -7,7 +7,7 @@
 use agent::Storage;
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
-use kata_sys_util::mount::{bind_remount, umount_timeout};
+use kata_sys_util::mount::{bind_remount, umount_all, umount_timeout};
 use kata_types::k8s::is_watchable_mount;
 use kata_types::mount;
 use nix::sys::stat::stat;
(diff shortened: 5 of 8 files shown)


src/runtime-rs/crates/resource/src/share_fs/mod.rs
```diff
@@ -131,6 +131,7 @@ pub trait ShareFsMount: Send + Sync {
     async fn umount_volume(&self, file_name: &str) -> Result<()>;
     /// Umount the rootfs
     async fn umount_rootfs(&self, config: &ShareFsRootfsConfig) -> Result<()>;
+    async fn clean_up(&self, sid: String) -> Result<()>;
```
justxuewei, line 134: As other functions do, how about passing sid's reference?

src/runtime-rs/crates/resource/src/manager.rs
```diff
@@ -105,6 +105,11 @@ impl ResourceManager {
         let inner = self.inner.read().await;
         inner.delete_cgroups().await
     }
+
+    pub async fn clean_up_hostpath(&self) -> Result<()> {
```
liubin, line 109: How about changing `clean_up` to `cleanup`, there are already many `cleanup`s.

src/runtime-rs/crates/resource/src/share_fs/mod.rs
liubin: Ditto.

src/runtime-rs/crates/resource/src/manager.rs
liubin: If we may need cleanup other resources in the future, I'd like the suggestion to add a method named `cleanup` which may do many cleanups including host path by calling private methods like:

```rust
pub cleanup(){
    self.cleanup_hostpath;
    self.cleanup_vlume;
    self.cleanup_network;
    ... ...
}
```

Just like https://github.com/kata-containers/kata-containers/pull/6080/files#diff-c62b06f33ee88853baa89dbb1c7b84e5da4d75e18172938d29475a81b0289f6bR262 what has done.

src/runtime-rs/crates/resource/src/share_fs/mod.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/resource/src/manager.rs

justxuewei (approved):
Lgtm, thanks!

src/runtime-rs/crates/resource/src/share_fs/utils.rs
```diff
@@ -59,6 +59,10 @@ pub fn get_host_rw_shared_path(sid: &str) -> PathBuf {
     Path::new(KATA_HOST_SHARED_DIR).join(sid).join("rw")
 }
 
+pub fn get_host_shared_path(sid: &str) -> PathBuf {
```
YushuoEdge, line 62: suggest rename to get_host_sbs_shared_path.

src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
```diff
@@ -224,4 +225,18 @@ impl ShareFsMount for VirtiofsShareMount {
...
+        // Unmount ro path
+        let host_ro_dest = get_host_ro_shared_path(sid);
+        umount_all(host_ro_dest.clone(), true).context("failed to umount ro path")?;
+        fs::remove_dir_all(host_ro_dest).context("failed to remove ro path")?;
+        // As the rootfs and volume have been umounted before calling this function, so just remove the rw dir directly
+        let host_rw_dest = get_host_rw_shared_path(sid);
+        fs::remove_dir_all(host_rw_dest).context("failed to remove rw path")?;
+        // remove the host share directory
+        let host_path = get_host_shared_path(sid);
+        fs::remove_dir_all(host_path).context("failed to remove host shared path")?;
```
YushuoEdge, line 239: I am curious why we need to remove dir three times? Seems only remove host_path is ok, by this, host_ro_dest and host_rw_dest will also be removed.

Tim-0731-Hzt: I think this is because when cleaning up the host share path, there are multiple directories to be handled, which is step by step. And later, we are going to support sandbox bind mount as well, so removing each directory when it is ready to delete makes more sense here.

src/runtime-rs/crates/resource/src/share_fs/utils.rs
Tim-0731-Hzt: Maybe the name is too long? I think maybe other developers should know the share path is only for sandbox?

YushuoEdge: fair enough

src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
YushuoEdge: ok

YushuoEdge (approved):

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -269,9 +259,24 @@ impl Sandbox for VirtSandbox {
...
+            .await
+            .context("delete cgroups")?;
+
+        info!(sl!(), "delete hypervisor");
+        self.hypervisor
+            .cleanup()
+            .await
+            .context("delete hypervisor")?;
+
+        info!(sl!(), "clean up host path");
```
quanweiZhou, line 275: Is it possible to provide a  `cleanup()` for the resource manager, including the `delete_cgroups` and `cleanup_hostpath`.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -237,6 +237,17 @@ impl ResourceManagerInner {
         self.cgroups_resource.delete().await
     }
 
+    pub async fn cleanup_hostpath(&self) -> Result<()> {
+        if let Some(share_fs) = &self.share_fs {
+            share_fs
+                .get_share_fs_mount()
+                .cleanup_hostpath(&self.sid)
```
quanweiZhou, line 244: Suggest providing a `cleanup()` for  the `share_fs` and encapsulated `get_share_fs_mount().cleanup_hostpath(&self.sid)` in the `cleanup()`.

quanweiZhou (commented):
Thanks, @Tim-0731-Hzt, few comments.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/resource/src/manager_inner.rs
Tim-0731-Hzt: fixed

quanweiZhou (approved):

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -233,8 +233,22 @@ impl ResourceManagerInner {
...
+            .context("delete cgroup")?;
+        // clean up share fs mount
+        if let Some(share_fs) = &self.share_fs {
+            share_fs
+                .get_share_fs_mount()
+                .cleanup(&self.sid)
+                .await
+                .context("failed to cleanup host path")?;
+        }
+        // TODO cleanup other resources
```
studychao, line 250: Is there any resource we know now that need to be cleaned up?

src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
```diff
@@ -224,4 +225,18 @@ impl ShareFsMount for VirtiofsShareMount {
 
         Ok(())
     }
+
+    async fn cleanup(&self, sid: &str) -> Result<()> {
+        // Unmount ro path
+        let host_ro_dest = get_host_ro_shared_path(sid);
+        umount_all(host_ro_dest.clone(), true).context("failed to umount ro path")?;
+        fs::remove_dir_all(host_ro_dest).context("failed to remove ro path")?;
+        // As the rootfs and volume have been umounted before calling this function, so just remove the rw dir directly
```
studychao, line 234: I am curious why or which part of code could maek sure that the rootfs and volume have been umounted before calling this function?

studychao (approved):
LGTM, just 2 small questions.

src/runtime-rs/crates/resource/src/manager_inner.rs
Tim-0731-Hzt: like some block devices, network endpoints need to be removed

src/runtime-rs/crates/resource/src/share_fs/virtio_fs_share_mount.rs
Tim-0731-Hzt: we do it here https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs#L252

Issue #6054, opened 2023-01-12
Dragonball: Logs are filled with "upcall" reconnection warnings


justxuewei:
I found logs are filled with "upcall" reconnection warnings after rebasing the latest main branch. 900 of 1000 lines print the same warning stuff as shown below.

```
Jan 12 13:52:42 - containerd[2503079]: {"msg":"upcall reconnect server...","level":"INFO","ts":"2023-01-12T05:52:42.424545255Z","source":"42d19c68af8104ad7e4b108ea244ae738243e0a60e258a373afbe020e989f628","subsystem":"root","pid":"3763576","version":"0.1.0","name":"kata-runtime"}
Jan 12 13:52:42 - containerd[2503079]: {"msg":"upcall connect server check failed, upcall server connect error: connection reset","level":"WARN","ts":"2023-01-12T05:52:42.424790601Z","version":"0.1.0","subsystem":"root","pid":"3763576","source":"42d19c68af8104ad7e4b108ea244ae738243e0a60e258a373afbe020e989f628","name":"kata-runtime"}
```

How can I prevent this from happening? /cc @Tim-0731-Hzt @studychao

studychao:
The main reason for that is you are not using guest kernel with upcall kernel patch.

And it's true that we should optimize the logic here to avoid too many logs for upcall.

I will do that optimization and for now you can either use a guest kernel with upcall patches or manually comment out this log

justxuewei:
Thanks for pointing out the workarounds!

justxuewei:
Hi @studychao, could you elaborate a bit on how to patch the "upcall"? I noticed that the patches provided by the Dragonball were tested on the "4.19" kernel. I tried to build my own "4.19.269" kernel with a tool named "build-kernel", but it failed unexpectedly. Then I tried to build a "5.10.25" kernel, but no luck here either. I'll list all of the steps and logs when building the kernels.

---

The process of building the "4.19.269" kernel is:

```bash
# failed
$ ./build-kernel.sh -v 4.19.269 setup
```

The logs of `./build-kernel.sh -v 4.19.269 setup` are:

```
linux-4.19.269.tar.xz: OK
INFO: Apply patches from /home/nxw/developer/kata-containers/tools/packaging/kernel/patches/4.19.x
INFO: Found 6 patches
[...]

PR #6081, opened 2023-01-16
upcall: add document for upcall

studychao:
In order for users to get better understand of upcall feature, we add this document for upcall to illustrate what is upcall and how to enable upcall.

fixes: #6054

### DIFF
--- a/src/dragonball/docs/upcall.md
+++ b/src/dragonball/docs/upcall.md
@@ -1,0 +1,30 @@
+# `Upcall`
+
+## What is `Upcall`?
+
+`Upcall` is a direct communication tool between VMM and guest developed upon `vsock`. The server side of the `upcall` is a driver in guest kernel (kernel patches are needed for this feature) and it'll start to serve the requests after the kernel starts. And the client side is in Dragonball VMM , it'll be a thread that communicates with `vsock` through `uds`.
+
+We want to keep the lightweight of the VM through the implementation of the `upcall`.
+
+![architecture overview](images/upcall-architecture.svg)
+## What can `upcall` do?
+
+We define specific operations in the device manager service (one of the services in `upcall` we developed) to perform device hotplug / hot-unplug including vCPU hotplug, `virtio-mmio` hotplug, and memory hotplug. We have accomplished device hotplug / hot-unplug directly through `upcall` in order to avoid the virtualization of ACPI to minimize virtual machines overhead. And there could be many other uses if other services are implemented.
+
+## How to enable `upcall`?
+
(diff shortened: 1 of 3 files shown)


GabyCT (approved):

justxuewei:
I followed the document and successfully built the kernel with the `upcall` patches, thanks!

justxuewei (approved):

studychao:
> I followed the document and successfully built a kernel with the `upcall` patches, thanks!

My pleasure:)

src/dragonball/docs/upcall.md
```diff
@@ -0,0 +1,30 @@
...
+
+## How to enable `upcall`?
+
+`Upcall` needs a server in the guest kernel which will be several kernel patches for the `upcall` server itself and different services registered in the `upcall` server. It's currently tested on upstream Linux kernel 5.10.
+
+To make it easy for users to use, we have open-source the `upcall` guest patches in [Dragonball experimental guest patches](../../../tools/packaging/kernel/patches/5.10.x/dragonball-experimental) and develop `upcall` support in [Kata guest kernel building script](../../../tools/packaging/kernel/build-kernel.sh).
+
+You could use following command to download the upstream kernel (currently Dragonball uses 5.10.25) and put the `upcall` patches and other Kata patches into kernel code.
+
+`sh build-kernel.sh -e -t dragonball -f setup`
```
bergwolf, line 22: Sorry for being late in the party, but I think we should have a single experimental kernel. It means that if we ever need other experimental kernel patches, we can put them all together instead of having to set up an experimental kernel for each different features. wdyt? /cc @fidencio @egernst 

btw, for upcall, it might actually be useful for other hypervisors too, e.g., firecracker?

studychao: Hi @bergwolf, sorry that I just notice your comments.
1. Right now, we will put all Dragonball experimental features (right now only upcall) into Dragonball experimental guest kernel. I am not quite familiar with other experimental kernel features but I think those patches may target different kernel version so it will be kinda hard to put all experimental features into just one experimental kernel 

2. Yes, ideally firecracker could use upcall to do things like cpu hot plug but they need to implement upcall client side in the VMM first.

bergwolf: @studychao, fair enough. I took a look at the `tools/packaging/kernel` directory and it has evolved a lot since the last time I looked. IMO it is quite a burden to maintain so many kernel patches and versions separately but looks like folks are just fine with it. So let's live with it ;)

bergwolf (approved):

Issue #6093, opened 2023-01-17
kata-runtime: Build (and implement) `kata-runtime check` on Darwin 


dcantah:
Continuing the work started/laid out in https://github.com/kata-containers/kata-containers/issues/3004 we should get kata-runtime building as well as add an implementation of the check command for Darwin as it's much different than Linux.

PR #6094, opened 2023-01-17
runtime: Implement `kata-runtime check` on Darwin

dcantah:
Fixes: #6093 

Add logic for kata-check on Darwin. This will do some simple verifications
to check if you're on a high enough build, and some sysctl checks for if
the processor supports virt, has it disabled, or if you're already in a
guest (no nested support in virt framework yet). To begin working on this 
we need to first get kata-runtime building on Darwin which involved splitting 
things up between linux/darwin as the checks necessary vary quite a bit..

This won't build at the moment, due to https://github.com/opencontainers/runc/pull/3697 
so leaving as draft until resolved


dcantah:
I'm just going to open this even though it doesn't build because of the runc issue. cc @egernst @mxpv

Issue #6102, opened 2023-01-18
versions: Bump QEMU to v7.2.0


fidencio:
v7.2.0 has been released in december 2022, see the changelog: https://wiki.qemu.org/ChangeLog/7.2

PR #6103, opened 2023-01-18
versions: Bump QEMU to v7.2.0

fidencio:
As QEMU released its v7.2.0 version in December last year, last do the bump on our side.

A few configuration options have been removed between the v6.2.0 (the vcersion we currently use) and v7.2.0, so those have also been dropped from our configure-hypervison.sh script (for this specific version).

Fixes: #6102

### DIFF
--- a/tools/packaging/scripts/configure-hypervisor.sh
+++ b/tools/packaging/scripts/configure-hypervisor.sh
@@ -242,7 +242,9 @@ generate_qemu_options() {
 	# Disable graphical network access
 	qemu_options+=(size:--disable-vnc)
 	qemu_options+=(size:--disable-vnc-jpeg)
-	qemu_options+=(size:--disable-vnc-png)
+	if ! gt_eq "${qemu_version}" "7.2.0" ; then
+		qemu_options+=(size:--disable-vnc-png)
+	fi
 	qemu_options+=(size:--disable-vnc-sasl)
 
 	# Disable PAM authentication: it's a feature used together with VNC access
@@ -330,13 +332,16 @@ generate_qemu_options() {
 	# building QEMU.
 	case "$arch" in
 	aarch64)
+		qemu_options+=(functionality:--disable-virtiofsd)
 		;;
 	x86_64)
+		qemu_options+=(functionality:--disable-virtiofsd)
 		;;
 	ppc64le)
 		qemu_options+=(functionality:--enable-virtiofsd)
 		;;
 	s390x)
+		qemu_options+=(functionality:--disable-virtiofsd)
 		;;
 	esac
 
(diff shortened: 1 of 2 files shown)


bergwolf (approved):
lgtm!

tools/packaging/scripts/configure-hypervisor.sh
```diff
@@ -330,13 +332,16 @@ generate_qemu_options() {
 	# building QEMU.
 	case "$arch" in
 	aarch64)
+		qemu_options+=(functionality:--disable-virtiofsd)
```
gkurz, line 335: Maybe worth mentioning the motivation (we've switched to Rust implementation) in the commit message ?

tools/packaging/scripts/configure-hypervisor.sh
```diff
@@ -330,13 +332,16 @@ generate_qemu_options() {
...
 		;;
 	x86_64)
+		qemu_options+=(functionality:--disable-virtiofsd)
 		;;
 	ppc64le)
 		qemu_options+=(functionality:--enable-virtiofsd)
 		;;
```
gkurz, line 342: AFAICT the Rust implemtation also exist for ppc64le. Why do we still need the C version ?

versions.yaml
```diff
@@ -88,8 +88,8 @@ assets:
     qemu:
       description: "VMM that uses KVM"
       url: "https://github.com/qemu/qemu"
```
gkurz, line 90: This is the official mirror. The official repo is https://gitlab.com/qemu-project/qemu actually. Just per curiosity (not asking for a change) : why not using the official repo ?

tools/packaging/scripts/configure-hypervisor.sh
fidencio: AFAIR @Amulyam24 hit some issues with the ppc64le version of virtiofsd, and as I don't have access to different HW platforms, I'm relying on the maintainers to do the switch whenever things are all working for them.

versions.yaml
fidencio: So, the whole thing boils down to the time GitLab was not yet being used and QEMU infra would go down and block Kata Containers project.  We can open a different PR for addressing this, if you feel the need.

tools/packaging/scripts/configure-hypervisor.sh
fidencio: Done, thanks!

versions.yaml
gkurz: Thanks for the explanation. I think we can live with that.

gkurz (approved):
I forgot to mention a tiny nit in the commit message ("vcersion") but the rest looks good so I'm approving this PR right away. Thanks @fidencio !

tools/packaging/scripts/configure-hypervisor.sh
Amulyam24: Hi @fidencio @gkurz. Yes, that's right. There are some issues with running the rust implementation on ppc64le. The fixes are merged in the official repo. Planning to switch to it soon once the other issues are resolved.

liubin (approved):
Just leave a comment as a reference https://github.com/qemu/qemu/blob/a8d6abe1292e1db1ad9be5b2b124b9c01bcda094/docs/about/deprecated.rst#virtiofsd

PR #6104, opened 2023-01-18
CC: Remove skopeo umoci

stevenhorsman:
- Remove skopeo and umoci code from the agent
- Remove associated agent configuration
- Remove the options to install the skopeo and umoci binaries into the rootfs

### DIFF
--- a/docs/how-to/how-to-build-and-test-ccv0.md
+++ b/docs/how-to/how-to-build-and-test-ccv0.md
@@ -68,7 +68,7 @@ $ chmod u+x ccv0.sh
         - Configure Kata to use containerd and for debug and confidential containers features to be enabled (including
           enabling console access to the Kata guest shell, which should only be done in development)
         - Create, build and install a rootfs for the Kata hypervisor to use. For 'CCv0' this is currently based on Ubuntu
-        20.04 and has extra packages like `umoci` added.
+        20.04.
         - Build the Kata guest kernel
         - Install the hypervisor (in order to select which hypervisor will be used, the `KATA_HYPERVISOR` environment
         variable can be used to select between `qemu` or `cloud-hypervisor`)
@@ -176,8 +176,6 @@ there.
         ```
         total 72
         -rw-r--r--  1 root root  2977 Jan 20 10:03 config.json
-        -rw-r--r--  1 root root   372 Jan 20 10:03 umoci.json
-        -rw-r--r--  1 root root 63584 Jan 20 10:03 sha256_be9faa75035c20288cde7d2cdeb6cd1f5f4dbcd845d3f86f7feab61c4eff9eb5.mtree
         drwxr-xr-x 12 root root   240 Jan 20 10:03 rootfs
         ```
         which shows how the image has been pulled and then unbundled on the guest.
@@ -279,8 +277,6 @@ the `ccv0.sh` script to automatically fill in the variables:
   total 72
   drwxr-xr-x 10 root root   200 Jan  1  1970 rootfs
   -rw-r--r--  1 root root  2977 Jan 20 16:45 config.json
-  -rw-r--r--  1 root root   372 Jan 20 16:45 umoci.json
-  -rw-r--r--  1 root root 63584 Jan 20 16:45 sha256_be9faa75035c20288cde7d2cdeb6cd1f5f4dbcd845d3f86f7feab61c4eff9eb5.mtree
   ```
 - Leave the Kata shell by running:
   ```bash
--- a/docs/how-to/how-to-build-and-test-ccv0.md
+++ b/docs/how-to/how-to-build-and-test-ccv0.md
@@ -297,9 +297,9 @@ In our test repository there are three tagged images:
 
 | Test Image | Base Image used | Signature status | GPG key status |
 | --- | --- | --- | --- |
-| `quay.io/kata-containers/confidential-containers:signed` | `busybox:1.33.1` | [signature](https://github.com/kata-containers/tests/tree/CCv0/integration/confidential/fixtures/quay_verification/signatures.tar) embedded in kata rootfs |  [public key](https://github.com/kata-containers/tests/tree/CCv0/integration/confidential/fixtures/quay_verification/public.gpg) embedded in kata rootfs |
+| `quay.io/kata-containers/confidential-containers:signed` | `busybox:1.33.1` | [signature](https://github.com/kata-containers/tests/tree/CCv0/integration/confidential/fixtures/quay_verification/x86_64/signatures.tar) embedded in kata rootfs |  [public key](https://github.com/kata-containers/tests/tree/CCv0/integration/confidential/fixtures/quay_verification/x86_64/public.gpg) embedded in kata rootfs |
 | `quay.io/kata-containers/confidential-containers:unsigned` | `busybox:1.33.1` | not signed | not signed |
-| `quay.io/kata-containers/confidential-containers:other_signed` | `nginx:1.21.3` | [signature](https://github.com/kata-containers/tests/tree/CCv0/integration/confidential/fixtures/quay_verification/signatures.tar) embedded in kata rootfs | GPG key not kept |
+| `quay.io/kata-containers/confidential-containers:other_signed` | `nginx:1.21.3` | [signature](https://github.com/kata-containers/tests/tree/CCv0/integration/confidential/fixtures/quay_verification/x86_64/signatures.tar) embedded in kata rootfs | GPG key not kept |
 
 Using a standard unsigned `busybox` image that can be pulled from another, *unprotected*, `quay.io` repository we can
 test a few scenarios.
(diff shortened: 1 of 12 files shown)


stevenhorsman:
`do-not-merge` whilst we are in code freeze

stevenhorsman:
FYI @arronwy - we are in code freeze, so to be clear this is for 0.4, but I thought I'd prep an early PR to give you a chance to have a look before Spring Festival if you want to. Thanks!

src/agent/src/image_rpc.rs
```diff
@@ -211,7 +120,7 @@ impl ImageService {
         let cid = if !req_cid.is_empty() {
             req_cid.to_string()
         } else if let Some(last) = req.get_image().rsplit('/').next() {
-            // ':' have special meaning for umoci during upack
+            // ':' not valid for container id
             last.replace(':', "_")
```
arronwy, line 124: After remove umoci, we don't need this replace operate. If cid is empty, we will use this info as image name which can associate with tags or digest

docs/how-to/how-to-build-and-test-ccv0.md
```diff
@@ -176,7 +176,6 @@ there.
         ```
         total 72
         -rw-r--r--  1 root root  2977 Jan 20 10:03 config.json
-        -rw-r--r--  1 root root   372 Jan 20 10:03 umoci.json
         -rw-r--r--  1 root root 63584 Jan 20 10:03 sha256_be9faa75035c20288cde7d2cdeb6cd1f5f4dbcd845d3f86f7feab61c4eff9eb5.mtree
```
arronwy, line 179: mtree file is also created by `umoci`, with `image-rs` we only have `config.json` and `rootfs`

docs/how-to/how-to-build-and-test-ccv0.md
```diff
@@ -279,7 +278,6 @@ the `ccv0.sh` script to automatically fill in the variables:
   total 72
   drwxr-xr-x 10 root root   200 Jan  1  1970 rootfs
   -rw-r--r--  1 root root  2977 Jan 20 16:45 config.json
-  -rw-r--r--  1 root root   372 Jan 20 16:45 umoci.json
   -rw-r--r--  1 root root 63584 Jan 20 16:45 sha256_be9faa75035c20288cde7d2cdeb6cd1f5f4dbcd845d3f86f7feab61c4eff9eb5.mtree
```
arronwy, line 281: Same comments as above.

arronwy (approved):
Thanks @stevenhorsman , LGTM! Only minor suggestions inline.

src/agent/src/image_rpc.rs
stevenhorsman: Hey Arron, do you mean replacing `last.replace(':', "_")` with just `last.to_string()` in this `else if`? I initially did this and updates the two unit tests that check the replacement of `:` with `_`, but we got an error from validate.rs (https://github.com/kata-containers/kata-containers/blob/40c69043245d0326080f48b2d295d31eee52e385/src/libs/kata-sys-util/src/validate.rs#L7-L28) 
`invalid container ID: "a:b"`
That's why I added the replace back in, but maybe there is a better way to solve this?

docs/how-to/how-to-build-and-test-ccv0.md
stevenhorsman: Thanks, removed

docs/how-to/how-to-build-and-test-ccv0.md
stevenhorsman: Thanks, removed

src/agent/src/image_rpc.rs
arronwy: > Hey Arron, do you mean replacing `last.replace(':', "_")` with just `last.to_string()` in this `else if`? I initially did this and updates the two unit tests that check the replacement of `:` with `_`, but we got an error from validate.rs (
> 
> https://github.com/kata-containers/kata-containers/blob/40c69043245d0326080f48b2d295d31eee52e385/src/libs/kata-sys-util/src/validate.rs#L7-L28
> 
> )
> `invalid container ID: "a:b"`
> That's why I added the replace back in, but maybe there is a better way to solve this?

Hi Steven, my fault, this commit `cde438ceb781888a8e6eadd9d31033efb0a3628c` "image_rpc: always call `verify_cid` after determining the cid" added cid check when the cid was derived from the image path. The logic here is correct.

src/agent/src/image_rpc.rs
```diff
@@ -235,9 +144,7 @@ impl ImageService {
 
         let cid = Self::cid_from_request(req)?;
         let image = req.get_image();
-        // Can switch to use cid directly when we remove umoci
-        let v: Vec<&str> = image.rsplit('/').collect();
-        if !v[0].is_empty() && v[0].starts_with("pause:") {
+        if !cid.is_empty() && cid.starts_with("pause:") {
```
arronwy, line 147: Since we added the cid check when it was derived from the image name and `:`is replaced with `_` in image name , if we switch to use cid, it may can not detect `pause:`  Sorry for misleading comments 
`// Can switch to use cid directly when we remove umoci`

stevenhorsman: Good spot, I'll switch to check for `pause` just in case.

src/agent/src/image_rpc.rs
```diff
@@ -59,95 +57,6 @@ impl ImageService {
...
-
-        // To save space delete the oci image after unpack
-        fs::remove_dir_all(&tmp_cid_path)?;
-
-        Ok(())
-    }
-
     // pause image is packaged in rootfs for CC
     fn unpack_pause_image(cid: &str) -> Result<()> {
         let cc_pause_bundle = Path::new(KATA_CC_PAUSE_BUNDLE);
```
snir911, line 62: having it means pause image bundle still needs to embedded in the podvm image?

src/agent/src/image_rpc.rs
snir911: FWIW i think assuming "pause:" is part of pause image name may be wrong is some cases like okd or openshift

src/agent/src/image_rpc.rs
stevenhorsman: Yes, we could consider making an optional setting for this as part of another PR, but the pause bundle being on the guest was introduced in https://github.com/kata-containers/kata-containers/issues/4882 and was to prevent the potential security risks of having the pause image be provided by the host side, which doesn't change with the removal of skopeo.

src/agent/src/image_rpc.rs
stevenhorsman: This might be worth raising as a separate issue as it would introduce the vulnerability of the pause image being provided by the host in non-peer pods CoCo. It might also help indicate why you guys have seen the behaviour of the pause image being pulled, rather than using the built in version if the image name doesn't match this?

src/agent/src/image_rpc.rs
snir911: but it will still require building umoci and using umoci at image build right?

src/agent/src/image_rpc.rs
snir911: Yes, it's related i guess, i wasn't able to follow the whole workflow. AFAIU normally pause container should be as any other container, in crio's configuration you can even set custom pause_command, I'll open an issue once I'll figure out what exactly is the issue:P

src/agent/src/image_rpc.rs
stevenhorsman: Yeah - that's correct - with this PR skopeo and umoci are still used for image build, but not built into the image, so doesn't inflate the size of the guest image, which is more important for the non-peer pods case, but should hopefully help peer pod VM boot time a little bit too.

src/agent/src/image_rpc.rs
c3d: Why is the `cid.is_empty()` test needed if `cid.starts_with` would rule that case out anyway?

stevenhorsman: Good point - I've remove the `is_empty()` check. Thanks!

fidencio (approved):
lgtm, thanks @stevenhorsman!

src/agent/src/image_rpc.rs
```diff
@@ -257,72 +164,58 @@ impl ImageService {
                 Err(_) => info!(sl!(), "Attestation Agent already running"),
             }
         }
+        // If the attestation-agent is being used, then enable the authenticated credentials support
```
wainersm, line 167: hi @stevenhorsman , here you switch two blocks so that attestation-agent config initilization is done before signature. Does it fix any issue? or it is just your preference?

stevenhorsman: No issue, I just thought that putting all the bits that were based on `aa_kbc_params` was more user friendly.

wainersm (approved):
Thank you very much @stevenhorsman ! Removing those stuffs is a hell of work!

sameo (approved):
Thanks a lot @stevenhorsman !

Issue #6107, opened 2023-01-19
tests: check_latest_version failed


liubin:
```
test utils::tests::test_kernel_version_valid_input ... ok
thread 'check::tests::check_latest_version' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unexpected character 'C' while parsing major version number")', src/check.rs:269:51
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
test check::tests::check_latest_version ... FAILED
             at ./rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:65:14
   2: core::result::unwrap_failed
             at ./rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/result.rs:1791:5
```

The first child of releases is named `CC-0.3.0`, this is not a valid `semver` version.

There is no meaning to check the tag name, so I'd like to delete the semver verification.

PR #6108, opened 2023-01-19
kata-ctl: remove version check in tests

liubin:
check_latest_version currently check the first tag_name returned by the releases from GitHub, but in some cases it may be not a valid semver like `CC-1.2.3`.

Indeed the check of the version/tag is not required and can be deleted.

Fixes: #6107

### DIFF
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -194,7 +194,6 @@ pub fn check_official_releases() -> Result<()> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use semver::Version;
 
     #[test]
     fn test_get_cpu_info_empty_input() {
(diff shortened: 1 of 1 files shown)


fidencio:
@liubin, I wonder if another and more complicated way to do this would be to actually ignore any kind of release that doesn't follow semver.

The reason we didn't go for semver as part of the CoCo is because people *should not* be using that till we have everthing merged back to main, and then we follow normal Kata Containers releases as usual.

WDYT?  What would be your preference?

jodh-intel:
Why is CC not using semver versions?

fidencio:
> Why is CC not using semver versions?

```
The reason we didn't go for semver as part of the CoCo is because people should not be using that till we have everthing merged back to main, and then we follow normal Kata Containers releases as usual.
```

jodh-intel:
semver supports pre release and patch release versions, so I'm still confused.

fidencio:
> semver supports pre release and patch release versions, so I'm still confused.

So, lete's say we'd go for the CC 0.1.0 release, how to cut that release and not confuse folks about the official Kata Containers 0.1.0 release?  Also, CC branch is a workaround that should not be used by general public.

fidencio:
@liubin, @jodh-intel, we can drop this patch as I deleted the releases of CC-x.y.z from the Kata Containers side.  We'll keep tagging those, but **not** cut any release.

liubin:
> @liubin, @jodh-intel, we can drop this patch as I deleted the releases of CC-x.y.z from the Kata Containers side. We'll keep tagging those, but **not** cut any release.

Close as @fidencio said.

PR #6114, opened 2023-01-20
Skip version check part

wenhuizhang:
The version check makes error, skip this part


liubin:
@wenhuizhang can you explain why you need this change?

liubin:
Guess that you do have not a `yq` installed correctly, the make command will use `yq` to get the golang version from the `versions.yaml` and check it. If you haven't a `yq` command, then you can't get the version, so the make will fail.

Please see `ci/install_yq.sh` for details.

PR #6116, opened 2023-01-20
WIP CC: Prepare kernel/initrd image for IBM SE (s390x)

BbolroC:
This is to adapt config for SE specific kernel/initrd image for IBM secure execution (TEE).

Fixes: #0000


BbolroC:
I would like to get harsh review (especially the kernel build part) from @wainersm @fidencio @hbrueckner and @fitzthum. Thanks a lot!

tools/packaging/guest-image/build_image.sh
```diff
@@ -39,17 +39,19 @@ build_initrd() {
...
 		kernel_version=${kernel_version#v}
 		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-sev-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
-		sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}" KERNEL_MODULES_DIR="${module_dir}"
 	else
-		sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}"
+		kernel_version="$(get_from_kata_deps "assets.kernel.version")"
+		kernel_version=${kernel_version#v}
+		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
 	fi
+	sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}" KERNEL_MODULES_DIR="${module_dir}"
```
fidencio: Related to the comment made above ...

tools/packaging/guest-image/build_image.sh
```diff
@@ -39,13 +39,19 @@ build_initrd() {
...
-		config_version=$(get_config_version)
 		kernel_version="$(get_from_kata_deps "assets.kernel.sev.version")"
 		kernel_version=${kernel_version#v}
 		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-sev-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
+	else
+		kernel_version="$(get_from_kata_deps "assets.kernel.version")"
+		kernel_version=${kernel_version#v}
+		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
+	fi
+	if [ "${KATA_BUILD_CC}" == "yes" ]; then
```
fidencio, line 54: This part here will cause a behaviour change, as we'll start trying to add kernel modules even for the targets that do not use it (TDX and vanilla).

Have you done some tests to check whether this will cause any regression?

tools/packaging/kernel/build-kernel.sh
```diff
@@ -409,7 +409,7 @@ build_kernel() {
 	arch_target=$(arch_to_kernel "${arch_target}")
 	pushd "${kernel_path}" >>/dev/null
 	make -j $(nproc ${CI:+--ignore 1}) ARCH="${arch_target}"
-	if [ "${conf_guest}" == "sev" ]; then
+	if [ "${KATA_BUILD_CC}" == "yes" ]; then
```
fidencio, line 412: This will also cause a change in the behaviour, and the same question applies.

fidencio (commented):
@BbolroC, I've left some questions.

tools/packaging/guest-image/build_image.sh
BbolroC: yep, I've tested newly built kernel, image and initrd without errors. But it is true that those additional built modules are only needed for cc-initrd. That is why I left `harsh review` for better solution. Thanks.

BbolroC: If SE (secure execution for IBM Z) requires some special modules like SEV, we could make a dedicated target like `cc-se-kernel-tarball`, but the thing is that a common initrd for CC does not work without those crypto modules.

BbolroC: Or why don't we make a dummy target which is only used for cc-initrd?

BbolroC:
I should make a dummy target not to break the existing build:

```
*** The present kernel configuration has modules disabled.
*** To use the module feature, please run "make menuconfig" etc.
*** to enable CONFIG_MODULES.
```

tools/packaging/guest-image/build_image.sh
Alex-Carter01: Any chance this kernel module is only needed for offline_fs_kbc and it could also delineate based on AA_KBC?

BbolroC: I decided to go with a SE specific dummy kernel build. The regression should be successful. (Please check my recent brand new changes)

tools/packaging/guest-image/build_image.sh
```diff
@@ -47,6 +47,13 @@ build_initrd() {
 		kernel_version=${kernel_version#v}
 		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-sev-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
 		sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}" KERNEL_MODULES_DIR="${module_dir}"
+	elif [[ "${AA_KBC:-}" == "offline_fs_kbc" ]] && [[ "${arch_target}" == "s390x" ]]; then
+		# IBM Z SE (secure execution) builds initrd with rootfs including common crypto kernel modules
+		config_version=$(get_config_version)
+		kernel_version="$(get_from_kata_deps "assets.kernel.se.version")"
+		kernel_version=${kernel_version#v}
+		module_dir="${repo_root_dir}/tools/packaging/kata-deploy/local-build/build/cc-se-kernel/builddir/kata-linux-${kernel_version}-${config_version}/lib/modules/${kernel_version}"
+		sudo -E PATH="$PATH" make rootfs ROOTFS_BUILD_DEST="${rootfs_build_dest}" KERNEL_MODULES_DIR="${module_dir}"
```
fitzthum, line 56: So just to clarify, we need these kernel modules to be loadable? It seems like it would be much simpler to have the builtin to the kernel.

BbolroC: But, yes, I agree that the builtin approach looks much simpler. Thanks! :wink:

BbolroC: @fidencio just to make sure that the built-in approach, Tobin mentioned, is something I had done before the current change which triggers the error message above?

> ```
> *** The present kernel configuration has modules disabled.
> *** To use the module feature, please run "make menuconfig" etc.
> *** to enable CONFIG_MODULES.
> ```

fidencio: https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kernel/configs/fragments/s390/crypto.conf isn't those built-in already?

BbolroC: I meant the crypt modules from https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kernel/configs/fragments/common/confidential_containers/cryptsetup.conf which is included when `KATA_BUILD_CC` is set.

BbolroC: So do you think including all the modules from https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kernel/configs/fragments/common/confidential_containers/cryptsetup.conf in https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kernel/configs/fragments/common/confidential_containers/cryptsetup.conf would be a simple solution? :wink:

BbolroC:
Hmm.. I think my hypothesis would be wrong. Let me revisit and dig into the problem again and request the review afterwards. Thanks!

BbolroC:
Update: it seems there was a mistake for configuration (unfortunately, I have no idea for which one). The existing target `rootf-initrd-tarball` works with `cc-kernel-tarball`. Sorry for disruption, guys. I will close this PR and the issue.

PR #6122, opened 2023-01-24
runtime/agent: Add log_level to the configuration of runtime and kata-agent 

yumingqiao:
This PR introduced the `log_level` configuration for `runtime` and `kata-agent` to enable more granular control of which log lines are produced. The `log_level` configuration takes precedence over `debug` if both are set. Currently, kata-agent's log is streamed out by `consoleWatcher` and is [always logged with Debug](https://github.com/kata-containers/kata-containers/blob/54f2b296e373f24722dc82045b5f1b6315feadad/src/runtime/virtcontainers/sandbox.go#L1058). This PR improves this by parsing the log line and log using the same log level used in `kata-agent`. 

In addition to the feature added, this PR has the following small changes: 
1. This PR sets the default `log_level` to be `warn` for both runtime and agent.
2. This PR makes the `consoleWatcher` always start. As consoleWatcher's job is to stream the kata agent's log, if people are worried about too many logs from agent, now we can control it by using the correct log_level.

A rewritten and some enhancements based from https://github.com/kata-containers/kata-containers/pull/4148.

Fixes: https://github.com/kata-containers/kata-containers/issues/366
Signed-off-by: yumingqiao qym316@gmail.com

### DIFF
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -6,6 +6,7 @@
 package virtcontainers
 
 import (
+	"context"
 	b64 "encoding/base64"
 	"encoding/json"
 	"errors"
@@ -19,6 +20,7 @@ import (
 	"time"
 
 	"github.com/docker/go-units"
+
 	"github.com/kata-containers/kata-containers/src/runtime/pkg/device/api"
 	"github.com/kata-containers/kata-containers/src/runtime/pkg/device/config"
 	volume "github.com/kata-containers/kata-containers/src/runtime/pkg/direct-volume"
@@ -32,8 +34,6 @@ import (
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/rootless"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/types"
 	"github.com/kata-containers/kata-containers/src/runtime/virtcontainers/utils"
-
-	"context"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -265,6 +265,7 @@ func ephemeralPath() string {
 // KataAgentConfig is a structure storing information needed
 // to reach the Kata Containers agent.
 type KataAgentConfig struct {
+	LogLevel           string
 	KernelModules      []string
 	ContainerPipeSize  uint32
 	DialTimeout        uint32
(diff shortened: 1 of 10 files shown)


yumingqiao:
cc @fengwang666

yumingqiao:
Hi @jodh-intel @liubin @bergwolf , gentle reminder for you guys to take a look, thanks!

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -309,13 +310,26 @@ func (k *kataAgent) longLiveConn() bool {
 	return k.keepConn
 }
 
+// agentLogLevel calculates the log level, LogLevel precedes the Debug flag.
+func agentLogLevel(config KataAgentConfig) string {
+	level := ""
+	if config.Debug {
```
YushuoEdge, line 316: this means we still use `enable_debug` in agent config to control log_level?

src/runtime/virtcontainers/sandbox.go
```diff
@@ -1206,14 +1242,12 @@ func (s *Sandbox) startVM(ctx context.Context, prestartHookFunc func(context.Con
 
 	s.Logger().Info("Starting VM")
 
-	if s.config.HypervisorConfig.Debug {
```
YushuoEdge, line 1209: Seems we could not remove this line. The serial device (console channel) are created based on `HypervisorConfig.Debug`, which means, if this param is not set, we can not reading kata-agent's log (?

YushuoEdge (commented):
Thanks! @yumingqiao I have some comments (or questions):

src/runtime/virtcontainers/kata_agent.go
yumingqiao: The current code uses `enable_debug` to control kata-agent's log level, too. This PR just makes the kata-agent log level configurable (instead of the current one that's always `debug`).
```
if config.Debug {
		params = append(params, Param{Key: "agent.log", Value: "debug"})
	}
```

src/runtime/virtcontainers/sandbox.go
yumingqiao: We need the consoleWatcher to read the kata-agent's log, you are correct. My PR however makes the consoleWatcher always enabled. As I made in the PR description, if the user thinks there are too many logs from kata-agent, the user can use the kata-agent's log level to tune.

liubin:
Can we merge `enable_debug` and `log_level` into one in the configuration files?
/cc @kata-containers/architecture-committee @kata-containers/kata-containers-maintainer

yumingqiao:
> Can we merge `enable_debug` and `log_level` into one in the configuration files?
Reading the discussion in https://github.com/kata-containers/kata-containers/pull/4148 gives me the impression that we would like to have both `enable_debug` and `log_level`, as @fidencio mentioned
> We should not break the current approach, but having the possibility to set the log_level right now is desired.

stevenhorsman:
Hey, quick question about how the agent's `log_level = warn` setting in the config file interacts with the current method of setting the agent log level - adding `agent.log=<level>` to the kernel_params. It is supposed to override it, or replace it. Do we need to update the documentation to point to this new config? Thanks?

PR #6140, opened 2023-01-27
virtiofsd: fix the build on ppc64le

Amulyam24:
`link-self-contained` is not supported on ppc64le rust target.
Hence, do not pass it while building virtiofsd.

Fixes: #6195

### DIFF
--- a/tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
+++ b/tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
@@ -57,6 +57,7 @@ init_env() {
 			LIBC="gnu"
 			ARCH="powerpc64le"
 			ARCH_LIBC=${ARCH}-linux-${LIBC}
+			extra_rust_flags=""
 			;;
 		"s390x")
 			LIBC="gnu"


tools/packaging/static-build/virtiofsd/build-static-virtiofsd.sh
```diff
@@ -57,6 +57,7 @@ init_env() {
 			LIBC="gnu"
 			ARCH="powerpc64le"
 			ARCH_LIBC=${ARCH}-linux-${LIBC}
+			extra_rust_flags=""
```
cmaf, line 60: Hi @Amulyam24, is this line necessary?

Amulyam24: Hi @cmaf, yes it is necessary because the extra rust flag `-C link-self-contained=yes` is no more supported on the ppc64le rust target which is causing the failure while compiling virtiofsd
```
error: option `-C link-self-contained` is not supported on this target
```

versions.yaml
```diff
@@ -290,13 +290,13 @@ externals:
   virtiofsd:
     description: "vhost-user virtio-fs device backend written in Rust"
     url: "https://gitlab.com/virtio-fs/virtiofsd"
-    version: "v1.3.0"
+    version: "v1.5.0"
```
Amulyam24, line 293: There's a PR already opened for virtiofsd bump. Added this change for time being to test as using the Depends-on was causing failure.

Amulyam24:
The initial plan for this PR was to switch to the rust version of virtiofsd since the necessary fixes for ppc64le have been added in v1.5.0. However, seeing some issues while testing. For now, changed the PR to unblock the PowerCI. Will fix the rust virtiofsd in another PR.

gkurz (approved):
LGTM. Thanks @Amulyam24 !

Amulyam24:
Note: The ppc64le failure is due to the flannel version dependency which will be fixed by https://github.com/kata-containers/tests/pull/5406

GabyCT (approved):

cmaf (approved):

Issue #6142, opened 2023-01-27
shim-v2/build.sh: Only build runtime-rs for the supported arches


fidencio:
Right now only x86_64 and ARM are supported, and this will cause issues when we try to use kata-deploy scripts as part of our CI.

See: https://github.com/kata-containers/tests/pull/5409#issuecomment-1406277333

PR #6143, opened 2023-01-27
shim-v2/build.sh: Only build runtime-rs for the supported arches

fidencio:
Right now only x86_64 and ARM are supported.

Fixes: #6142

### DIFF
--- a/src/runtime-rs/Makefile
+++ b/src/runtime-rs/Makefile
@@ -28,11 +28,17 @@ default:
 test:
 	@echo "s390x is not currently supported"
 	exit 0
+install:
+	@echo "s390x is not currently supported"
+	exit 0
 else ifeq ($(ARCH), powerpc64le)
 default:
 	@echo "PowerPC 64 LE is not currently supported"
 	exit 0
 test:
+	@echo "PowerPC 64 LE is not currently supported"
+	exit 0
+install:
 	@echo "PowerPC 64 LE is not currently supported"
 	exit 0
 else
@@ -41,6 +47,7 @@ default: runtime show-header
 ##TARGET test: run cargo tests
 test:
 	@cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
+install: install-runtime install-configs
 endif
 
 ifeq (,$(realpath $(ARCH_FILE)))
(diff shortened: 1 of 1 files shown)


fidencio:
Maybe a better approach would be to do exactly what we do for s390x in the runtime-rs Makefile.  Let me switch.

tools/packaging/static-build/shim-v2/build.sh
```diff
@@ -27,15 +27,17 @@ if [ ${arch} = "ppc64le" ]; then
...
-sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
-	-w "${repo_root_dir}/src/runtime-rs" \
-	"${container_image}" \
-	bash -c "git config --global --add safe.directory ${repo_root_dir} && make PREFIX=${PREFIX} QEMUCMD=qemu-system-${arch}"
-
-sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
-	-w "${repo_root_dir}/src/runtime-rs" \
-	"${container_image}" \
-	bash -c "git config --global --add safe.directory ${repo_root_dir} && make PREFIX="${PREFIX}" DESTDIR="${DESTDIR}" install"
+if [ "${arch}" != "ppc64" ] || [ "${arch}" != "s390x" ]; then
```
liubin, line 30: Should

```diff
- if [ "${arch}" != "ppc64" ] || [ "${arch}" != "s390x" ]; then
+ if [ "${arch}" != "ppc64" ] && [ "${arch}" != "s390x" ]; then
```

studychao (approved):
Thanks:)

tools/packaging/static-build/shim-v2/build.sh
studychao: +1

Tim-0731-Hzt (approved):

BbolroC:
@fidencio has allowed me to take over this PR and finalize it. I got a target `install` included in the conditional branch as well because `make install` is called for `CCv0` (to which the PR is supposed to be back-ported).
Unless the target is included, a target `runtime` is triggered by a `install` - `install-runtime` chain, which breaks the CI on s390x and ppcle64.

cmaf (approved):

PR #6145, opened 2023-01-27
packaging: Cache the container used to build the kata-deploy artefacts

fidencio:
As we'll move the CI to be using the kata-deploy scripts to build the components, let's cache the images used and improve a little bit the building time of those components, while also having a realiable way to reproduce issues when debugging any sort of issues.

This has been in place as part of the `CCv0` branch for a while and is a forward-port of that.

### DIFF
--- a/.github/workflows/kata-deploy-push.yaml
+++ b/.github/workflows/kata-deploy-push.yaml
@@ -28,6 +28,13 @@ jobs:
           - virtiofsd
           - nydus
     steps:
+      - name: Login to Kata Containers quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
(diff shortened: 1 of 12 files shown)


fidencio:
I am not sure why the GitHub actions are failing, as we do have the secrets needed as part of our repo.

fidencio:
```
Anyone with collaborator access to the repositories with access to a secret or variable can use it for Actions. They are not passed to workflows that are triggered by a pull request from a fork.
```

This is from the secrets page on GitHub admin view.  So, I think those tests will fail till we have them merged, which sucks quite a lot.

fidencio:
I'm closing this in favour of https://github.com/kata-containers/kata-containers/pull/6310, which comes from the Kata Containers repo instead of a fork.  By doing this we at least can ensure the actions will be passing as expected.

Issue #6154, opened 2023-01-30
docker exec hang


bergwolf:
Reported by Jim Carroll from Mirantis:

terminal 1:
```docker run --rm -it --name katatest --runtime io.containerd.kata.v2 alpine```
terminal 2:
```docker exec -it katatest /bin/sh```
everything seems fine.
Until I try to exit the exec session, at which point the following happens:
* the exec’d session hangs (tty is effectively frozen)
* the original docker run session exits
* docker ps shows that no containers are running

I have to kill the exec command still hung in terminal 2 to regain control of that lost soul

liubin:
Should this be closed? @bergwolf

neersighted:
Not until we merge the PR in Moby, in my opinion. It looks like GitHub has even set it up to auto-close.

neersighted:
Looks like the auto-close doesn't work across orgs, despite the Web UI claiming so. Anyway, we've merged the fix for the Moby master branch, and the fix for 23.0 is in review for the next patch. I'd suggest this can be closed and those interested in the fixes can monitor the related PRs.

bergwolf:
Thanks @neersighted ! Let's close this as https://github.com/moby/moby/pull/44892 has been merged.

PR #6155, opened 2023-01-30
runtime: return fake pid for exec process exit events

bergwolf:
We cannot return the container ID 'cause that can make moby think that the container process itself exits. Let's return a fake pid 0 here.

Fixes: #6154

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/utils.go
+++ b/src/runtime/pkg/containerd-shim-v2/utils.go
@@ -22,9 +22,15 @@ import (
 )
 
 func cReap(s *service, status int, id, execid string, exitat time.Time) {
+	var pid uint32
+	if execid == "" {
+		pid = s.hpid
+	} else {
+		pid = 0
+	}
 	s.ec <- exit{
 		timestamp: exitat,
-		pid:       s.hpid,
+		pid:       pid,
 		status:    status,
 		id:        id,
 		execid:    execid,


src/runtime/pkg/containerd-shim-v2/utils.go
```diff
@@ -22,9 +22,15 @@ import (
 )
 
 func cReap(s *service, status int, id, execid string, exitat time.Time) {
+	var pid uint32
+	if execid == "" {
+		pid = s.hpid
+	} else {
+		pid = 0
```
liubin, line 29: Can moby handle the pid of 0 correctly?

corhere: Yes. [This is the only place in the Moby codebase which does anything with PIDs aside from reporting their values in Engine API responses.](https://github.com/moby/moby/blob/b54af02b51b0dfe2e2863e8f1647aaa27936c274/daemon/monitor.go#L158-L161) It's also the root cause of the bug which this PR is working around.

corhere (approved):
This should work around the issue just fine. The root cause is a bug in Moby, which https://github.com/moby/moby/pull/44888 should fix.

neersighted:
I would suggest rejecting this PR in favor of a fix in Moby itself, as Cory has quite convincingly shown the fault is there. This could certainly be brought into Kata to enable support sooner (e.g. on the 20.10 branch), but I think we should have no issues backporting @corhere's fix to 23.0 which will be released shortly.

neersighted:
It was just pointed out to me that https://github.com/moby/moby/pull/43800 is not backported to 20.10, in which case I think there is no value working around what is a bug in Moby in Kata, assuming we backport @corhere's fix to the Moby 23.0 branch.

liubin (approved):

src/runtime/pkg/containerd-shim-v2/utils.go
liubin: @corhere thank you for the clarification.

bergwolf:
Thanks a lot @corhere @neersighted ! Let's drop this in favor of https://github.com/moby/moby/pull/44888 !

Issue #6077, opened 2023-01-13
Use cgroup.kill on 5.14+ kernels 


dcantah:
In [5.14+](https://lwn.net/Articles/855924/) you can write to a cgroup.kill file that will send a SIGKILL to every process in the group for you. This can avoid the Freeze -> Send signal -> Thaw dance that we employ here in a few spots:

1. https://github.com/kata-containers/kata-containers/blob/main/src/agent/rustjail/src/container.rs#L1122-L1130
2. https://github.com/kata-containers/kata-containers/blob/67f0fd505dc55d5913dded0fdda442575c765788/src/agent/src/rpc.rs#L424-L455

We'll need this change https://github.com/kata-containers/cgroups-rs/pull/97 which is being pulled in with https://github.com/kata-containers/kata-containers/pull/6040.

PR #6156, opened 2023-01-30
agent: Use cgroup.kill where supported

dcantah:
Fixes: #6077

In 5.14+ you can write to a cgroup.kill file that will send a SIGKILL to every process in the group for you. Support for this was recently added to cgroups-rs and brought in here so we can make use of it. This can avoid the Freeze -> Send signal -> Thaw dance that we use here, as well as looping through the pids of a cgroup and sending SIGKILL manually.

### DIFF
--- a/src/agent/rustjail/src/cgroups/fs/mod.rs
+++ b/src/agent/rustjail/src/cgroups/fs/mod.rs
@@ -180,6 +180,11 @@ impl CgroupManager for Manager {
         Ok(())
     }
 
+    fn kill(&self) -> Result<()> {
+        self.cgroup.kill()?;
+        Ok(())
+    }
+
     fn destroy(&mut self) -> Result<()> {
         let _ = self.cgroup.delete();
         Ok(())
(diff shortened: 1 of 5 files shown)


dcantah:
Some of these failures seem to be occurring on other PRs (that don't touch the areas that are actually failing it seems).

```
1. thread 'ops::volume_ops::tests::test_add_remove' panicked at 'remove failed: Permission denied (os error 13)', src/ops/volume_ops.rs:test ops::version::tests::test_get

2. thread 'watcher::tests::create_tmpfs' panicked at 'assertion failed: is_mounted(mount_point).unwrap()', src/watcher.rs:1306:13
```

dcantah:
@liubin Almost forgot about this, could you re-kick off the CI? Or perhaps I should probably rebase to bring in any improvements since as well

liubin (approved):

egernst (approved):

dcantah:
Is this good to check in?

PR #6158, opened 2023-01-30
CC | Take runtime-rs into consideration when building and caching the shim-v2 tarball

fidencio:
We've disabled building shim-v2 as part of https://github.com/kata-containers/kata-containers/commit/299829aec09a1f237b85bc6cebd49980766ef33f. However, now TDX support will be implemented there and we must start considering it for building and caching the shim-v2 tarball.

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -330,9 +330,11 @@ install_cc_qemu() {
 #Install all components that are not assets
 install_cc_shimv2() {
 	local shim_v2_last_commit="$(get_last_modification "${repo_root_dir}/src/runtime")"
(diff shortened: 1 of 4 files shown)


fidencio:
/cc @studychao @wainersm 

@studychao, we need to consider a few other things as well.  Will you be using the Kernel with upcall support for CCv0?

studychao:
> /cc @studychao @wainersm
> 
> @studychao, we need to consider a few other things as well. Will you be using the Kernel with upcall support for CCv0?

For now, we could just use the vanilla TDX guest kernel as other hypervisors use. 
upcall only enable resize ability for VM and we need some talk internally to discuss whether it is a fit for TDX environment, so please ignore upcall for CC for now.

Thanks!

studychao (approved):
LGTM, thanks @fidencio :)

stevenhorsman (approved):
LGTM. Thanks @fidencio!

PR #6161, opened 2023-01-30
firecracker: Support disk rate limiter

### LINKED ISSUE #6160: Add disk rate limiter support for Firecracker

Firecracker rate limiter support.

**How can it be improved?**

Currently for Firecracker VMM only network tx/rx rate limiter is supported. Though Firecracker itself supports also disk bandwidth/operations rate limiter, and there are corresponding values in the hypervisor configuration structure, these values are not passed to the Firecracker. 

**Additional Information**

Anything else to add? No.

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)? Yes

akoptelov:
This PR adds code that passes disk limiter parameters from KC configuration to Firecracker.

This fixes #6160.


akoptelov:
I need to check that disk rate values are correctly scaled for Firecracker (current assumption is that Firecracker uses bytes, but that should be double-checked).

akoptelov:
Beep...

PR #6166, opened 2023-01-30
Minor cleanups in make file

amshinde:


Makefile
```diff
@@ -45,10 +45,8 @@ docs-url-alive-check:
 
 .PHONY: \
 	all \
-	binary-tarball \
+	kata-tarball \
+	install-tarball \
 	default \
-	install-binary-tarball \
```
cmaf, line 50: @amshinde the commit message is a little unclear, were these targets supposed to be added to the tools/packaging/kata-deploy/local-build Makefile?

cmaf (commented):
@amshinde one question about the removal of targets from the Makefile

Makefile
amshinde: I think it should matter either way since this Makefile has an include for the other one. I added in the top level Makefile for them to be easily discoverable from end user point of view.

dborquez (approved):
lgtm, thank you @amshinde

GabyCT (approved):

cmaf:
@amshinde this needs a rebase

PR #6174, opened 2023-01-31
runtime: Drop QEMU log file support

### LINKED ISSUE #6173: QEMU log file should not be used
The QEMU log file isn't limited in size, nor rotated in any way. A container running inside the VM  could thus potentially fill up the disk where the file resides (e.g. on openshift the file is on a 15 GiB tmpfs partition => 15 GiB of RAM per VM). Also, the log file isn't really useful unless debugging QEMU internals, which is best performed outside of a kata environment.

In order to avoid potential DoS on the host in the future, we should just drop the related code. This won't break any user experience since it never worked before :  QEMU error messages are only collected since PR #5736.

gkurz:
The QEMU log file is essentially about fine grain tracing of QEMU internals and mostly useful for developpers, not production. Notably, the log file isn't limited in size, nor rotated in any way. It means that a container running in the VM could possibly flood the log file with a guest triggerable trace. For example, on openshift, the log file is supposed to reside on a per-VM 14 GiB tmpfs mount. This means that each pod running with the kata runtime could potentially consume this amount of host RAM which is not acceptable.

Error messages are best collected from QEMU's stderr as kata is doing now since PR #5736 was merged. Drop support for the QEMU log file because it doesn't bring any value but can certainly do harm.

Fixes #6173

### DIFF
--- a/src/runtime/pkg/govmm/qemu/qemu.go
+++ b/src/runtime/pkg/govmm/qemu/qemu.go
@@ -2628,9 +2628,6 @@ type Config struct {
 	// PidFile is the -pidfile parameter
 	PidFile string
 
-	// LogFile is the -D parameter
-	LogFile string
-
 	qemuParams []string
 }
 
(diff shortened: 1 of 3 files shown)


liubin (approved):

bergwolf (approved):
Thanks @gkurz !

gkurz:
Adding the bug label as the use of the QEMU log file is unsafe and should be backported to stable-3.0.

PR #6175, opened 2023-01-31
Qemu logs for stable 3.0

### LINKED ISSUE #5780: QEMU error logs are missing

### LINKED ISSUE #6173: QEMU log file should not be used

gkurz:
This backports recent work on QEMU logs:
- don't daemonize QEMU
- collect QEMU logs from stderr
- drop QEMU log file support (unsafe, can fill up host storage)

Fixes #5780 for stable-3.0
Fixes #6173 for stable-3.0

### DIFF
--- a/src/runtime/pkg/govmm/qemu/examples_test.go
+++ b/src/runtime/pkg/govmm/qemu/examples_test.go
@@ -27,13 +27,16 @@ func Example() {
 	// resources
 	params = append(params, "-m", "370", "-smp", "cpus=2")
 
-	// LaunchCustomQemu should return as soon as the instance has launched as we
-	// are using the --daemonize flag.  It will set up a unix domain socket
-	// called /tmp/qmp-socket that we can use to manage the instance.
-	_, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
+	// LaunchCustomQemu should return immediately. We must then wait
+	// the returned process to terminate as we are using the --daemonize
+	// flag.
+	// It will set up a unix domain socket called /tmp/qmp-socket that we
+	// can use to manage the instance.
+	proc, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil, nil)
 	if err != nil {
 		panic(err)
 	}
+	proc.Wait()
 
 	// This channel will be closed when the instance dies.
 	disconnectedCh := make(chan struct{})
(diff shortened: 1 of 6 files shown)


c3d:
@gkurz Please add the required porting labels.

c3d (approved):
This looks good to me. Apparently, backporting was relatively trivial.

In addition to the issues that @gkurz listed, let me share with others a bit of additional context. This is a relatively important backport for OpenShift supportability, since it allows us to get "decent" logging out of `qemu`, which was entirely missing due to daemonization.

bergwolf (approved):
Thanks @gkurz !

snir911 (approved):
Thanks Greg!
nit: notice that you're sign-off appear twice probably because it was cherry-picked

Issue #6189, opened 2023-02-02
qemu upgrade to 7.2.0 causing SEV containers to fail launch


ryansavino:
# Description of problem

Upgrade to qemu 7.2.0 contains a bug that prevents kernel hashes option from working correctly. SEV containers fail to launch.

# Expected result

Successful SEV container launch using kata runtime class.

# Actual result

SEV containers fail to launch.

c3d:
@ryansavino Do you have a pointer to the `qemu` bug?

c3d:
> @ryansavino Do you have a pointer to the `qemu` bug?

This would be it?

https://lore.kernel.org/qemu-devel/da39abab9785aea2a2e7652ed6403b6268aeb31f.camel@linux.ibm.com/

ryansavino:
> > @ryansavino Do you have a pointer to the `qemu` bug?
> 
> This would be it?
> 
> https://lore.kernel.org/qemu-devel/da39abab9785aea2a2e7652ed6403b6268aeb31f.camel@linux.ibm.com/

@c3d Yep. @dubek gave a little more detail here:
https://github.com/kata-containers/kata-containers/pull/6190#issuecomment-1415397452

I assume you probably already saw that.

dubek:
Note: the QEMU maintainer sent a pull request which includes the revert of the offending patches:
https://lore.kernel.org/qemu-devel/20230302082343.560446-1-mst@redhat.com/
(we need patches 02-08 in that series)

@fidencio : Do we wait until these patches land in the master branch and then we copy them over to the `patches` dir?

stevenhorsman:
SEV is not supported now

PR #6190, opened 2023-02-02
[main] versions: downgrade qemu from 7.2.0 to 7.1.0

ryansavino:
A known bug in qemu 7.2.0 causes a problem handling the kernel hashes argument and causes SEV container launching to fail.

Fixes: #6189

### DIFF
--- a/tools/packaging/scripts/configure-hypervisor.sh
+++ b/tools/packaging/scripts/configure-hypervisor.sh
@@ -242,7 +242,7 @@ generate_qemu_options() {
 	# Disable graphical network access
 	qemu_options+=(size:--disable-vnc)
 	qemu_options+=(size:--disable-vnc-jpeg)
-	if ! gt_eq "${qemu_version}" "7.2.0" ; then
+	if ! gt_eq "${qemu_version}" "7.1.0" ; then
 		qemu_options+=(size:--disable-vnc-png)
 	fi
 	qemu_options+=(size:--disable-vnc-sasl)
@@ -358,7 +358,7 @@ generate_qemu_options() {
 	qemu_options+=(size:--disable-vde)
 
 	# Don't build other options which can't be depent on build server.
-	if ! gt_eq "${qemu_version}" "7.2.0" ; then
+	if ! gt_eq "${qemu_version}" "7.1.0" ; then
 		qemu_options+=(size:--disable-xfsctl)
 		qemu_options+=(size:--disable-libxml2)
 	fi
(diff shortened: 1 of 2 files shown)


fidencio:
is there a patch for that upstream?  I'd rather add a patch than do the downgrade on main.

amshinde:
@ryansavino As @fidencio mentioned, is there a patch to fix this. We would rather have a patch fix rather than a downgrade if possible. Also, can you mention the bug report that filed upstream?

ryansavino:
> @ryansavino As @fidencio mentioned, is there a patch to fix this. We would rather have a patch fix rather than a downgrade if possible. Also, can you mention the bug report that filed upstream?

We're merging the downgrade into CCv0 directly per @fidencio's comment on #6191. This PR for main can get left on the wayside as we wait for the qemu upstream patch. @dubek @fitzthum do you have the qemu bug report for this?

dubek:
There's no bug report that I'm aware of.

[This qemu-devel thread](https://lore.kernel.org/qemu-devel/da39abab9785aea2a2e7652ed6403b6268aeb31f.camel@linux.ibm.com/) has different POVs on this issue and approaches to solve it in the 8.x cycle.

Note that the RNG seed feature which was added in 7.2 both broke SEV and also has another bug in its implementation (decompressed kernel clobbers setup_data). [The fix for the other bug](https://lore.kernel.org/qemu-devel/20221230220725.618763-1-Jason@zx2c4.com/) solves the hash verification issue for the kernel file but breaks the hash of the kernel cmdline -- so it still breaks SEV. It wasn't merged to master yet...

If we want to keep qemu 7.2, we can work around the bug by adding `-machine pc-q35-7.1` to the QEMU command-line (shim modification?). This can be done for SEV only (in branch CCv0?).

Another option is to build qemu 7.2 with a patch file; the patch at the beginning of the thread mentioned above solves the issue for SEV, but it's not accepted upstream (and won't be accepted, because it simply disables a feature which other setups need; we don't need it because guests with OVMF don't need this feature at all). Are we OK adding a non-upstreamed patch?

fidencio:
> Are we OK adding a non-upstreamed patch?

I'd say no.  Still, as we have discussed, we can wait till this gets solved upstream to actually have it backported here as we're not that close to making a stable release out of the main branch.

c3d:
@gkurz Could you take a look at this one?

gkurz:
> @gkurz Could you take a look at this one?

QEMU community seems to have a consensus on simply [reverting  the RNG seed mess](https://lore.kernel.org/qemu-devel/20230208211212.41951-1-mst@redhat.com/). The fixes will be applied to the master branch for QEMU 8.0 but they are expected to be valid for QEMU 7.2 as well.

PR #6191, opened 2023-02-02
[CCv0] versions: downgrade qemu from 7.2.0 to 7.1.0

ryansavino:
A known bug in qemu 7.2.0 causes a problem handling the kernel hashes argument and causes SEV container launching to fail.

Fixes: #6189


fidencio (approved):
lgtm, thanks @ryansavino!

For the main branch, I'd rather wait for an official patch and add the patch to the Kata Containers tree.  For the CCv0 branch, we should proceed with this patch!

PR #6193, opened 2023-02-02
runtime-rs: ignor "no such process" error when delete cgroup for a thread to let it go

jongwu:
It lead to "no such process" error when delete cgroup for a thread which may exit already. We can ignor this error because we need not do the delete op for a non-exist thread. This error can happen on Arm CI regularily. 

To achieve it, we should retrive the Os error code from the return value of the "remove_task". But it seems there is something to be improved in cgroups-fs, thus this PR depends on another [PR](https://github.com/kata-containers/cgroups-rs/pull/108)

Fow now, I temporarily switch the cgroups-fs to my own fork to do the test.


src/runtime-rs/crates/resource/src/cgroups/mod.rs
```diff
@@ -109,7 +111,19 @@ impl CgroupsResource {
     /// overhead_cgroup_manager to the parent and then delete the cgroups.
     pub async fn delete(&self) -> Result<()> {
         for cg_pid in self.cgroup_manager.tasks() {
-            self.cgroup_manager.remove_task(cg_pid)?;
+            if let Err(error) = self.cgroup_manager.remove_task(cg_pid) {
+                match error.source() {
+                    Some(err) => match err.downcast_ref::<io::Error>() {
+                        Some(e) => {
+                            if e.raw_os_error() != Some(3) {
```
studychao, line 118: Could you use a const to describe what does the `3` mean here?

studychao (commented):
Thanks @jongwu , please take a look at the comment.

src/runtime-rs/crates/resource/src/cgroups/mod.rs
jongwu: Sure, it's better.

src/runtime-rs/crates/resource/src/cgroups/mod.rs
```diff
@@ -109,7 +113,22 @@ impl CgroupsResource {
     /// overhead_cgroup_manager to the parent and then delete the cgroups.
     pub async fn delete(&self) -> Result<()> {
         for cg_pid in self.cgroup_manager.tasks() {
-            self.cgroup_manager.remove_task(cg_pid)?;
+            // For now, we can't guarantee that the thread in cgroup_manager does still
+            // exist. Once it exit, we should ignor that error returned by remove_task
+            // to let it go.
+            if let Err(error) = self.cgroup_manager.remove_task(cg_pid) {
```
yaoyinnan, line 119: It is suggested that `remove_task_by_tgid` should be used instead of `remove_task` to adapt to cgroup v2. Cgroup v2 currently only supports deleting tasks in `thread mode`.

In order to shield the caller from judging the cgroup mode, https://github.com/kata-containers/cgroups-rs/pull/104 has been modified.

yaoyinnan (changes requested):
Thanks @jongwu, please take a look at the comment.

In which `arm dragonball` CI job did the error you describe appear? Can you share the link?

src/runtime-rs/crates/resource/src/cgroups/mod.rs
jongwu: OK, thanks @yaoyinnan

jongwu:
> Thanks @jongwu, please take a look at the comment.
> 
> In which `arm dragonball` CI job did the error you describe appear? Can you share the link?

This bug will lead to the failure of kata stability test, the error info like:

> 19:39:36 Start iteration 1 of 2
19:39:36 Running...
19:39:36 Checking 0 containers have all relevant components
19:39:51 And we have hit the max 20 containers
19:39:51 Checking 20 containers have all relevant components
19:39:52 [soak_parallel_rm.sh:237] INFO: Wait until the containers gets removed
19:40:07 Checking 0 containers have all relevant components
19:40:08 Wrong number of shims running (0 != 1) - stopping

see http://jenkins.katacontainers.io/job/kata-containers-dragonball-ubuntu-Arm-20.04-PR/278/consoleFull

src/runtime-rs/crates/resource/src/cgroups/mod.rs
YushuoEdge: Hi @yaoyinnan , as we discussed earlier, `remove_task_by_tgid` may move redundant tasks to root cgroup, and this is not what we expected. Ignoring this error message maybe more reasonable.

src/runtime-rs/crates/resource/src/cgroups/mod.rs
```diff
@@ -109,7 +113,22 @@ impl CgroupsResource {
...
+                match error.source() {
+                    Some(err) => match err.downcast_ref::<io::Error>() {
+                        Some(e) => {
+                            if e.raw_os_error() != Some(OS_ERROR_NO_SUCH_PROCESS) {
```
YushuoEdge, line 123: Can we add a ErrorKind in cgroup-rs?

YushuoEdge (commented):
Thanks @jongwu @yaoyinnan !! Hope the CI issue can be solved ASAP!!

src/runtime-rs/crates/resource/src/cgroups/mod.rs
jongwu: Hi @YushuoEdge -, What ErrorKind do you mean? no such process error?

YushuoEdge: Oh, true, NoSuchProcess should not distinguished in `remove_task`. The current approach to handle this error may be the best solution！

studychao:
Hi @jongwu , I have merged the dependent PR in cgroup-rs, please rebase this PR.

Also, we have found out that this problem is also in x86 runtime-rs + Dragonball, so it's kinda important to merge this PR as soon as possible:) Thanks!

src/runtime-rs/crates/resource/src/cgroups/mod.rs
jongwu: Thanks @YushuoEdge -, I really love this change, but I can't reslove the compile error😞. As this PR is emergency, we can improve it later.

YushuoEdge: OK！

studychao:
@jongwu  Hi, cgroup-rs has release v0.3.1 to include your PR. Please update this PR and we could follow up to merge this.

jongwu:
Great! Thanks @studychao

jongwu:
All tests pass. Can this be approved? @studychao @YushuoEdge

studychao (approved):
Thanks! @jongwu

yaoyinnan (approved):
Thanks! @jongwu

yaoyinnan (approved):
Thanks! @jongwu

Tim-0731-Hzt (approved):

quanweiZhou (commented):
Thanks @jongwu, LGTM.

PR #6209, opened 2023-02-03
Action check kernel config version

amshinde:
github-action: Add step to verify kernel config version id updated
    
The version mentioned in the `kata_config_version` needs to be
updated for any kernel config change or changed to the patches applied.
Without this, CI would not test with the latest kernel changes.
We use to enforce this earlier as part of CI when `packaging` was
a standalone repo.
Add back this check as part of a github action so that the check is
performed early on instead of a CI job.

Fixes: kata-containers/tests#5419

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -27,7 +27,6 @@ jobs:
       target_branch: ${{ github.base_ref }}
     steps:
     - name: Checkout code
-      if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
       uses: actions/checkout@v3
       with:
         fetch-depth: 0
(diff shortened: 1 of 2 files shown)


cmaf (approved):
thanks @amshinde

GabyCT (approved):

.github/workflows/static-checks.yaml
```diff
@@ -38,6 +37,22 @@ jobs:
         go-version: 1.19.3
       env:
         GOPATH: ${{ runner.workspace }}/kata-containers
+    - name: Check kernel config version
+      run: |
+        cd "${{ github.workspace }}/src/github.com/${{ github.repository }}"
+        kernel_dir="tools/packaging/kernel/"
+        kernel_version_file="${kernel_dir}kata_config_version"
+        modified_files=$(git diff --name-only origin/main..HEAD)
+        result=$(git whatchanged origin/main..HEAD "${kernel_dir}" >>"/dev/null")
```
liubin, line 46: Nit: var `result` is not used.

Issue #6211, opened 2023-02-03
No docker-buildx-plugin in docker image ubuntu:20.04 for s390x|ppc64le


BbolroC:
# Description of problem

ATM, a package `docker-buildx-plugin` is not available for docker upstream `ubuntu:20.04` for s390x and ppc64le. There is no configuration for an environment variable `VERSION` at https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/local-build/dockerbuild/Dockerfile#L21, which makes it stick to `latest`. This leads to adding `docker-buildx-plugin` up. As a quick remedy (temporary solution because it is blocking many PRs), it is necessary to set `VERSION` to `v20.10` for the affected architectures in Dockerfile.

PR #6212, opened 2023-02-03
CI: Make docker version stick to v20.10 in ubuntu:20.04 for s390x|ppc64le

BbolroC:
This is to make a docker version to v20.10 in docker upstream image ubuntu:20.04 for s390x|ppc64le.

Fixes: #6211

### DIFF
--- a/tools/packaging/kata-deploy/local-build/dockerbuild/Dockerfile
+++ b/tools/packaging/kata-deploy/local-build/dockerbuild/Dockerfile
@@ -8,6 +8,7 @@ ENV INSTALL_IN_GOPATH=false
 
 COPY install_yq.sh /usr/bin/install_yq.sh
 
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
This looks fine to me at the moment as the only package dropped as a result of this is `docker-buildx-plugin`, which we obviously aren't using on s390x as it isn't available!

gkurz (approved):
LGTM. Thanks @BbolroC !

Amulyam24:
> It seems `ppc64le` is also affected. @Amulyam24 is it okay to include `ppc64le` in the conditional like below?
> 
> > if uname -m | grep -Eq 's390x|ppc64le'; then export VERSION="v20.10"; fi && \

Absolutely, Thanks @BbolroC !

GabyCT (approved):

Issue #6213, opened 2023-02-03
runtime-rs: unable to delete cgroup causes the number of shims to not correspond


yaoyinnan:
In the arm CI of dragonball, the cleanup of `containerd.sock` is not complete, the error info like:

```
19:39:36 Start iteration 1 of 2
19:39:36 Running...
19:39:36 Checking 0 containers have all relevant components
19:39:51 And we have hit the max 20 containers
19:39:51 Checking 20 containers have all relevant components
19:39:52 [soak_parallel_rm.sh:237] INFO: Wait until the containers gets removed
19:40:07 Checking 0 containers have all relevant components
19:40:08 Wrong number of shims running (0 != 1) - stopping
```

The log of containerd shows that an error is reported for delete cgroup:

```
Feb 02 13:35:47 ubuntu20-937060 containerd[116803]: {"msg":"delete cgroup","level":"INFO","ts":"2023-02-02T13:35:47.419545268Z","source":"kata-zuyZ03","subsystem":"virt-container","pid":"118353","version":"0.1.0","name":"kata-runtime"}
Feb 02 13:35:47 ubuntu20-937060 containerd[116803]: time="2023-02-02T13:35:47.443457289Z" level=error msg="failed to shutdown shim task" error="Others(\"failed to handler message handler request\\n\\nCaused by:\\n    0: do shutdown\\n    1: delete cgroups\\n    2: unable to write to a control group file tasks, value 118358 caused by: Os { code: 3, kind: Uncategorized, message: \\\"No such process\\\" }\"): unknown" id=kata-zuyZ03
```

PR #6214, opened 2023-02-03
runtime-rs: use remove_task_by_tgid instead of remove_task in delete cgroup

yaoyinnan:
Use remove_task_by_tgid instead of remove_task in delete cgroup, to move all threads of the process to root according to tgid.

Fixes: #6213

### DIFF
--- a/src/runtime-rs/crates/resource/src/cgroups/mod.rs
+++ b/src/runtime-rs/crates/resource/src/cgroups/mod.rs
@@ -108,8 +108,8 @@ impl CgroupsResource {
     /// delete will move the running processes in the cgroup_manager and
     /// overhead_cgroup_manager to the parent and then delete the cgroups.
     pub async fn delete(&self) -> Result<()> {
-        for cg_pid in self.cgroup_manager.tasks() {
-            self.cgroup_manager.remove_task(cg_pid)?;
+        for cg_pid in self.cgroup_manager.procs() {
+            self.cgroup_manager.remove_task_by_tgid(cg_pid)?;
         }
 
         self.cgroup_manager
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/resource/src/cgroups/mod.rs
```diff
@@ -108,17 +108,17 @@ impl CgroupsResource {
     /// delete will move the running processes in the cgroup_manager and
     /// overhead_cgroup_manager to the parent and then delete the cgroups.
     pub async fn delete(&self) -> Result<()> {
-        for cg_pid in self.cgroup_manager.tasks() {
-            self.cgroup_manager.remove_task(cg_pid)?;
+        for cg_pid in self.cgroup_manager.procs() {
```
YushuoEdge, line 111: I just wonder if we can remove all tasks when we just remove all pids in `procs` file?

YushuoEdge (commented):
Thanks yinnan! @yaoyinnan

BbolroC:
Regarding the CI errors for `[s390x|ppc64le]-snap`, it has been resolved by https://github.com/kata-containers/kata-containers/pull/6212. Please rebase this onto main.

yaoyinnan:
> Regarding the CI errors for `[s390x|ppc64le]-snap`, it has been resolved by #6212. Please rebase this onto main.

Thank you.

YushuoEdge:
Hi @yaoyinnan , we think error "no such process" is allowed when we remove_task, and let's just ignore this error to fix the issue instead of  using remove_task_by_tgid.

studychao:
As discussed with @yaoyinnan  & @YushuoEdge , We should use #6193 to fix the cgroup problem, so I will close this PR.

PR #6231, opened 2023-02-07
osbuilder: ungzip initrd when extracting

jepio:
The initrd that is built is gzip compressed, it needs to be ungzipped before passing to cpio.

### DIFF
--- a/tools/osbuilder/Makefile
+++ b/tools/osbuilder/Makefile
@@ -100,7 +100,7 @@ $(ROOTFS_BUILD_DEST)/.%$(ROOTFS_MARKER_SUFFIX):: rootfs-builder/%
 .PRECIOUS: $(ROOTFS_BUILD_DEST)/.dracut$(ROOTFS_MARKER_SUFFIX)
 $(ROOTFS_BUILD_DEST)/.dracut$(ROOTFS_MARKER_SUFFIX): $(TARGET_INITRD)
 	mkdir -p $(TARGET_ROOTFS)
-	(cd $(TARGET_ROOTFS); cat $< | cpio --extract --preserve-modification-time --make-directories)
+	(cd $(TARGET_ROOTFS); zcat $< | cpio --extract --preserve-modification-time --make-directories)
 	@touch $@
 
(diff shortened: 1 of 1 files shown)


tools/osbuilder/Makefile
```diff
@@ -100,7 +100,7 @@ $(ROOTFS_BUILD_DEST)/.%$(ROOTFS_MARKER_SUFFIX):: rootfs-builder/%
 .PRECIOUS: $(ROOTFS_BUILD_DEST)/.dracut$(ROOTFS_MARKER_SUFFIX)
 $(ROOTFS_BUILD_DEST)/.dracut$(ROOTFS_MARKER_SUFFIX): $(TARGET_INITRD)
 	mkdir -p $(TARGET_ROOTFS)
-	(cd $(TARGET_ROOTFS); cat $< | cpio --extract --preserve-modification-time --make-directories)
+	(cd $(TARGET_ROOTFS); zcat $< | cpio --extract --preserve-modification-time --make-directories)
```
fidencio, line 103: The change itself makes sense, and just for completeness, here's where the initrd is gzipped: https://github.com/kata-containers/kata-containers/blob/8ae14f6a5565b7e759be9b9e0643c818e599bf4d/tools/osbuilder/initrd-builder/initrd_builder.sh#L79

Some of the checks in our CI are going to complain that there's no `Fixes: #NNN` in your commit message, as the community requires an issue to be created and linked to the PR's commit message.

Last but not least, I wonder why the CI is not hitting this case at all.

jepio: I looked into this, and there is some confusing with the makefile targets:
* the default target should be `all`, not `$(ROOTFS_BUILD_DEST)/.dracut$(ROOTFS_MARKER_SUFFIX)`
* `$(ROOTFS_BUILD_DEST)/.dracut$(ROOTFS_MARKER_SUFFIX)` only be buildable when `BUILD_METHOD=dracut` is set.

So my fix is wrong, for the dracut case the initrd is actually not compressed. I'll open a new PR.

PR #6234, opened 2023-02-07
clh: Enforce API timeout only for vm.boot request

alex-matei:
launchClh already has a timeout of 10seconds for launching clh, e.g. if launchClh or setupVirtiofsDaemon takes a few seconds the context's deadline will already be expired by the time it reaches bootVM

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -652,9 +652,6 @@ func (clh *cloudHypervisor) StartVM(ctx context.Context, timeout int) error {
 	span, _ := katatrace.Trace(ctx, clh.Logger(), "StartVM", clhTracingTags, map[string]string{"sandbox_id": clh.id})
 	defer span.End()
 
-	ctx, cancel := context.WithTimeout(context.Background(), clh.getClhAPITimeout()*time.Second)
-	defer cancel()
-
 	clh.Logger().WithField("function", "StartVM").Info("starting Sandbox")
 
 	vmPath := filepath.Join(clh.config.VMStorePath, clh.id)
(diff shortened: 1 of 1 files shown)


src/runtime/virtcontainers/clh.go
```diff
@@ -693,6 +690,9 @@ func (clh *cloudHypervisor) StartVM(ctx context.Context, timeout int) error {
 	}
 	clh.state.PID = pid
 
+	ctx, cancel := context.WithTimeout(context.Background(), clh.getClhAPITimeout()*time.Second)
```
cmaf, line 693: Is it possible to create the context using the existing one (ctx passed in) as the parent context? I didn't have time to extensively look through the code but a quick search showed that subsequent calls from StartVM and beyond create trace spans from this context, and when a new context from background is used it creates a span with no parent unattached to the root. However, I'm not sure if the parent context[s] also had a timeout and if this conflicts. @jodh-intel 

Despite that, I'm approving since this use of context was already approved and you're only changing the location here.

cmaf (approved):
thanks @alex-matei, approving but with 1 comment/question

cmaf (commented):
@alex-matei, could you please add a "Fixes #XXXX" line to the commit message?

Details for patch formatting can be found here: https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

src/runtime/virtcontainers/clh.go
liubin: Agree with @cmaf, use the input `ctx` can keep passing information to the next function calls.

alex-matei: Got it, fixed.

jodh-intel (approved):
Thanks @alex-matei.

lgtm

PR #6238, opened 2023-02-08
tools: add nerdctl as a new container engine

changweige:
For users that do not have docker installed locally, nerdctl which is a containerd community client is very convenient, especially for containerd and its sub-projects developers.

In addition, try to format the rootfs.sh script making it neat.

Signed-off-by: Changwei Ge <gechangwei@bytedance.com>

### DIFF
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -64,8 +64,7 @@ typeset ROOTFS_DIR
 typeset init=
 
 #$1: Error code if want to exit different to 0
-usage()
-{
+usage() {
 	error="${1:-0}"
 	cat <<EOF
 
@@ -161,7 +160,7 @@ architectures:
 https://github.com/kata-containers/kata-containers/tree/main/tools/osbuilder#platform-distro-compatibility-matrix
 
 EOF
-exit "${error}"
+	exit "${error}"
 }
 
 get_distros() {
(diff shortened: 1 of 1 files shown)


tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -371,6 +368,8 @@ build_rootfs_distro()
 		elif [ -n "${USE_PODMAN}" ]; then
 			container_engine="podman"
 			engine_build_args+=" --runtime ${DOCKER_RUNTIME}"
+		elif [ -n "${USE_NERDCTL}" ]; then
```
liubin, line 371: Please can you add `USE_NERDCTl` in `tools/osbuilder/README.md` about how to use it and the use case, for example, combinate it with buildkit.

liubin: And please add an option description to the help message https://github.com/kata-containers/kata-containers/blob/f1381eb361aa2202d48607df8cfc6dbfe6db580c/tools/osbuilder/rootfs-builder/rootfs.sh#L151-L157

changweige: Sure. will add more information and an introduction for related docs within this PR. Besides, I feel it's better to split this PR into two one of which is only formatting the script.

amshinde: Yes please, this should be split into two separate commits. Please include all the formatting changes in a separate commit.

changweige:
Depends on #6255

Issue #6197, opened 2023-02-02
runtime: cloud-hyperviror hugepages kernel-params always take effect, when the trigger 'enable_hugepages' is 'false'


shippomx:
# Description of problem
hugepage anntotation trigger dosen't work if `io.katacontainers.config.hypervisor.kernel_params` has been set to false.

create a sandbox with this annotations:
```json
"annotations": {
    "io.katacontainers.config.hypervisor.enable_hugepages": "false",
    "io.katacontainers.config.hypervisor.enable_mem_prealloc": "false",
    "io.katacontainers.config.hypervisor.default_memory": "4098",
    "io.katacontainers.config.hypervisor.kernel_params": "default_hugepagesz=2M hugepagesz=2M hugepages=1024 hugepagesz=1G hugepages=1 ",
    "io.katacontainers.config.runtime.sandbox_cgroup_only": "false"
  },
```

# Expected result
The sandbox doesn't use the hugepages resources. And cloud-hypervisor kernel-params should not contain hugepages params.

# Actual result
And the `VM config` log is：

```
[...]

PR #6258, opened 2023-02-10
runtime: clh always enable hugetlb with kernel params

shippomx:
If pod annotataion with hugepages kernel params, whenever the trigger
    'enable_hugepges' is, cloud-hypervisor will always enable hugetlb feature.

    Fixes: #6197


shippomx:
@liangxianlong

Issue #6261, opened 2023-02-10
Incomplete kata-agent log messages on console when using initrd with AGENT_INIT=yes


jepio:
# Description of problem

I noticed that when running kata with an initrd built with `AGENT_INIT=yes`, I don't see any log messages from kata-agent on the console. For e.g. if an invalid `agent.config_file=` path is provided through the kernel command line, kata-agent panics but the panic can't be seen (only a "attempted to kill init" from the kernel). This also includes other more verbose json log messages that kata-agent prints when working.

All these messages are visible when using systemd (`AGENT_INIT=no`), so I investigated and found that when an initramfs is in use, the kernel does not mount devtmpfs before running `/init`. So kata-agent's stdout/stderr are not connected to `/dev/console`. The kernel even complains about this (`Warning: unable to open an initial console.`).

# Expected result

See kata-agent log messages on console (or forwarded to host) when kata-agent is the init process.

# Actual result

No log messages on console from kata-agent.

# Further information

This is fixed by providing some standard device nodes inside the initrd. PR will follow.

PR #6262, opened 2023-02-10
osbuilder: Include minimal set of device nodes in ubuntu initrd

jepio:
When starting an initrd the kernel expects to find /dev/console in the initrd, so that it can connect it as stdin/stdout/stderr to the /init process. If the device node is missing the kernel will complain that it was unable to open an initial console. If kata-agent is the initrd init process, it will also result in log messages not being logged to console and thus not forwarded to host syslog.

Add a set of standard device nodes for completeness, so that console logging works.

Fixes: #6261
Signed-off-by: Jeremi Piotrowski <jpiotrowski@microsoft.com>

### DIFF
--- a/tools/osbuilder/rootfs-builder/ubuntu/Dockerfile.in
+++ b/tools/osbuilder/rootfs-builder/ubuntu/Dockerfile.in
@@ -6,6 +6,8 @@ ARG IMAGE_REGISTRY=docker.io
 FROM ${IMAGE_REGISTRY}/ubuntu:@OS_VERSION@
 @SET_PROXY@
 
+# makedev tries to mknod from postinst
+RUN [ -x /usr/bin/systemd-detect-virt ] || ( echo "echo docker" >/usr/bin/systemd-detect-virt && chmod +x /usr/bin/systemd-detect-virt )
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


amshinde (approved):
lgtm. @jepio Thanks for fixing this!

liubin (approved):

GabyCT (approved):

jepio:
Argh, makedev deb tries to create device nodes in the host during postinst. I'll check if there's a way to disable that.

jepio:
Pushed a fix, i don't know if tests will automatically rerun or someone needs to trigger.

fidencio (approved):
lgtm, thanks @jepio!

I'm restarting the jobs that didn't pass as the failures were not related to this PR.

jepio:
This was my fault - the previous version still had the same issue under podman. I've pushed a fix that should work this time.
@Amulyam24 could you retrigger the tests?

dcantah (approved):

fidencio (approved):
lgtm, thanks @jepio!

The kata-deploy action issues are not related and have been already solved.
I don't think it's worth asking you to rebase the PR and re-run the CI just to get the green ticks there.  This, I'm merging it as it is.

PR #6272, opened 2023-02-13
runtime: tracing: Fix missing ctx return

### LINKED ISSUE #6271: missing ctx return in creating a trace span in CLH StartVM()
We should return the context in the creation of a trace span to maintain the correct ordering of calls in tracing output

cmaf:
Normally we return the context when creating a trace span so that the ordering of spans w.r.t. calls is maintained in tracing output. Add missing context for StartVM() for Cloud Hypervisor.

Fixes #6271


fidencio (approved):
lgtm, thanks @cmaf!

jodh-intel (approved):
Thanks @cmaf.

lgtm

Issue #6276, opened 2023-02-14
Dragonball TDX: add is_tdx_enable to identify tdx


studychao:
In order to disable or enable some features when running tdx vms, we
need to add is_tdx_enable() function.

PR #6277, opened 2023-02-14
CC | Dragonball TDX: add is_tdx_enabled to identify tdx VM type

studychao:
In order to disable or enable some features when running tdx vms, we need to add is_tdx_enabled() function to identify whether the VM confidiential type is TDX.

fixes: #6276

Signed-off-by: fengshifang <fengshifang@linux.alibaba.com>

### DIFF
--- a/src/dragonball/src/vm/mod.rs
+++ b/src/dragonball/src/vm/mod.rs
@@ -350,6 +350,15 @@ impl Vm {
         instance_state == InstanceState::Running
     }
 
+    /// return true if VM confidential type is TDX
+    pub fn is_tdx_enabled(&self) -> bool {
+        let shared_info = self
+            .shared_info()
+            .read()
+            .expect("failed to get instance state, because shared info is poisoned lock");
+        shared_info.is_tdx_enabled()
+    }
+
     /// Save VM instance exit state
     pub fn vm_exit(&self, exit_code: i32) {
         if let Ok(mut info) = self.shared_info.write() {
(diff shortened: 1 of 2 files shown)


src/dragonball/src/vm/mod.rs
```diff
@@ -350,6 +350,15 @@ impl Vm {
         instance_state == InstanceState::Running
     }
 
+    /// Return true if TDX is enabled
+    pub fn is_tdx_enable(&self) -> bool {
+        let shared_info = self
+            .shared_info()
+            .read()
+            .expect("failed to get instance state, because shared info is poisoned lock");
```
fidencio, line 358: Checking for the TDX or TDX Seam module wouldn't be a better way to do this?

studychao: Hi, here we are not checking whether the hardware has TDX cabability.
Instead, we are just checking the VM type is TDX VM or normal VM (The information should be passed by the runtime)
I will write more accurate comments and commit message and please tell me if that makes sense.

studychao: Optimized. Please take a look.

fidencio (approved):
lgtm, thanks @studychao!

c3d:
> Thanks @studychao. Could you please rename the function to `is_tdx_enabled`, since I believe it checks the state but does not change it?

Looking at the code, I wonder if `has_tdx` wouldn't be a better name, since at least one of the checks seems to be static, not dynamic.

studychao:
Hi @c3d , thanks for your advice. I think `is_tdx_enabled` is better and I have adjusted the function name to `is_tdx_enabled`.

GabyCT (approved):

PR #6289, opened 2023-02-15
feat(runtime): vcpu resize capability

Ji-Xinyou:
## Content
* This PR allows Kata Container 3.0(runtime-rs) to resize the vCPU on the fly, currently only working with dragonball.
  * Note that this has to work with `dragonball upcall patches` [[doc](https://github.com/kata-containers/kata-containers/blob/main/src/dragonball/docs/upcall.md)] and `dragonball upcall feature`[in Cargo.toml].
  * But to work with QEMU, I think we only need to support QMP cpu hotplugging in the runtime (TODO).

## How to test
* In the container spec, you add `cpu_quota` field and `cpu_period` field to `linux[resources]`. For more info, see [here](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto)
    * For example, a `cpu_quota` of `200000` and `cpu_period` of `100000` should result in 2 (200000/100000, rounded up) more vcpus in the VM.
* Set the `default_maxvcpu` field in the kata configuration toml file
* In your pod, create and start the container with the spec.
    * The resources will be updated on `Start`, `Create` and `Update`.
* You could check the cpu number with command `sudo crictl exec <container-id> cat /proc/cpuinfo` with `crictl` or similar cli tools.


src/runtime-rs/crates/runtimes/common/src/runtime_handler.rs
```diff
@@ -41,3 +42,27 @@ pub trait RuntimeHandler: Send + Sync {
...
+
+impl RuntimeInstance {
+    // NOTE: if static resource management is configured, a warning is logged
+    // hotplug vcpu/memory, and the cpu will not be updated since the sandbox
+    // resource should be static
+    //
+    // The updated resource is calculated from:
+    //   - vcpu: the sum of each ctr, plus default vcpu
+    //   - memory: the sum of each ctr, plus default memory, and setup swap
+    pub async fn update_sandbox_resource(&self) -> Result<()> {
```
Tim-0731-Hzt, line 54: I think `update_sandbox_resource` only works with `virt-containers`. If the runtime instance is Linux or wasm, I think it can't work. My suggestion is to implement it in `container_manager`, because you can also fetch `resource_manager`  there, then you don't need to take it from` virt_sandbox`

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -272,4 +277,107 @@ impl ContainerManager for VirtContainerManager {
         process.process_type == ProcessType::Container
             && process.container_id.container_id == self.sid
     }
+
+    // if guest_swap is true, add swap to memory_sandbox
+    // returns `(memory_sandbox_b, swap_sandbox_b)` (in Byte)
+    async fn mem_required(&self) -> Result<(u64, i64)> {
```
Tim-0731-Hzt, line 283: Could you leave this on another pr? As this is for memory

Tim-0731-Hzt (commented):
Thanks @Ji-Xinyou , a  few comments

src/runtime-rs/crates/runtimes/common/src/runtime_handler.rs
Tim-0731-Hzt: > I think the implementation of `update_sandbox_resource` only works with `virt-containers`. If the runtime instance is Linux or wasm, I think it can't work. My suggestion is to implement it in `container_manager`, because you can also fetch `resource_manager` there, then you don't need to take it from` virt_sandbox`

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
Ji-Xinyou: I will leave this as a dummy one.

src/runtime-rs/crates/runtimes/common/src/runtime_handler.rs
Tim-0731-Hzt: But my point is this operation only works for virt-containers, if the runtime instance is Linux or wasm, this won't work

Ji-Xinyou: > But my point is this operation only works for virt-containers, if the runtime instance is Linux or wasm, this won't work

This works with other type of container, since a sandbox often manages more than one containers.

Tim-0731-Hzt: `let nr_vcpus = self.container_manager.vcpus_required().await?;` I think only virt-container would have vcpu

Ji-Xinyou: > `let nr_vcpus = self.container_manager.vcpus_required().await?;` I think only virt-container would have vcpu

This `vcpu` is actually the rounded `cpu_quota / cpu_period` value, this semantic applys to all kinds of container. Do you think changing this to `cpu_required()` will be fine?

Tim-0731-Hzt: yes, I agree

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -226,6 +226,13 @@ impl RuntimeHandlerManager {
                 .await
                 .context("create container")?;
 
+            // update sandbox resource when creating container
+            // the function is idempotent
+            instance
+                .update_sandbox_resource()
```
YushuoEdge, line 232: Hi @Ji-Xinyou, I still recommend not to do here. For my concern:

- I have not found any cgroup operations in your code, but actually we need to do that after vcpu hotpluged. Since we will create new threads, and they need to be constrained. (If you have wriiten those code, please point that to me, thanks~)
- Assumed we have to do cgroup operations, so under current implementation, for one create_container request, we have to do cgroup operations twice. And cgroup operations are always the bottlenecks in high elasticity and high density senarios.

Ji-Xinyou: * In create container, the cgroup is updated.
* In start container, the cgroup does not need to be updated.
  * case1: a created but unstarted container, whose cgroup has already been updated
  * case2: a started but paused container, whose cgroup is setup upon creation
* In update container, the cgroup is updated.

In current implementation, I think the cgroup is updated only once since the sandbox resource update does not enforce the cgroup update.

Tim-0731-Hzt: I think @YushuoEdge 's point is you need to update cgroup to constrain Vmm and vcpu threads after `update_sandbox_resource`. As cgroup is been updated in `create_container`once, so it will become twice

Tim-0731-Hzt: You can check how kata 2.0 did this https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/sandbox.go#L1376

Ji-Xinyou: > I think @YushuoEdge 's point is you need to update cgroup to constrain Vmm and vcpu threads after `update_sandbox_resource`. As cgroup is been updated in `create_container`once, it will become twice when you do it after `update_sandbox_resource`

1. The cgroup information like `cpu_quota` and `cpu_period` is set upon container creation and update(in create_container and update_container), we DO NOT need to modify this when we update the sandbox resource, all it does it to `hotplug and record` cpu/mem resource.
2. As you mentioned in 2.0, 2.0 update the resource AFTER the container creation. The reason is just like I said, this update DOES NOT modify the cgroup. However, it does update the `cpuset`, [see here](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/sandbox.go#L2236). Since we are not supporting cpu set in the PR, this is not a problem.

YushuoEdge: We need to write newly spawned threads to cgroup controller, right?

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl CpuResource {
+    pub fn new(config: Arc<TomlConfig>) -> Result<Self> {
+        let hypervisor_name = config.runtime.hypervisor_name.clone();
+        let hypervisor_config = config
+            .hypervisor
+            .get(&hypervisor_name)
+            .context("failed to get hypervisor")?;
+        Ok(Self {
+            default_vcpu: hypervisor_config.cpu_info.default_vcpus,
```
Tim-0731-Hzt, line 28: I think you don't need a struct to store this value because it's easy to get it from the config. Also, I think you don't need the `CpuResource` struct, you can store the `current_vcpu` in the resource manager.

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: > @Ji-Xinyou We need to write newly spawned threads to cgroup controller, right?

Hmm, I am not 100% sure. But I think the logic is like this.

The `spec` specifies each container's fair share of cpu, lets say container `C1` has a share of 2 vCPUs and `C2` has a share of 1 vCPUs. Their cgroup information are **contraint in the VM**, i.e., inside the VM, they are sharing the vCPU with the ratio of `2:1`.

What we really care is on the host side, which is the actual share of the container. Take QEMU as an example, each vCPU is actually a thread. In Linux CFS, the thread with the same parameters (e.g. nice value) are sharing the harts (hardware threads) **fairly**.

Inside the container, they are sharing the vCPU correctly, which means they are sharing the vCPU threads correctly, which further means they are sharing the host-side harts correctly.

Therefore, my current opinion would be `if we are not supporting cpuset, the newly spawned threads will be fine if they are correctly handled by the guest cgroup (which I think is the case).`

src/runtime-rs/crates/runtimes/common/src/runtime_handler.rs
quanweiZhou: @Tim-0731-Hzt I agree with @Ji-Xinyou that this is the resource of a `sandbox` rather than a container, sandbox resource is the summary of the sandbox, it works for Virt-container/Linux-container/Wasm-container. For Virt-container it is the host resource such as CPU and memory cgroup. For Linux-container it is the Pod Level resource such as CPU, and memory cgroup. 
All the differences between Virt-container/Linux-container/Wasm-container are implemented in the Sandbox.

For the design, I have some suggestions @Ji-Xinyou 
1. The `update_sandbox_resource()` should be implemented in the sandbox, the sandbox can get the total CPU from the cgroup resource, which saves all the Linux resources for all containers.
2. The design of `CpuMemResource` is a little thin, and it is recommended to combine it with `CgroupsResource` to look appropriate

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -276,6 +283,13 @@ impl RuntimeHandlerManager {
                     .start_process(&process_id)
                     .await
                     .context("start process")?;
+
+                // update sandbox resource in case when starting a paused container
+                instance
+                    .update_sandbox_resource()
```
quanweiZhou, line 289: I think this is no need to `update_sandbox_resource()` when starting the process, the CPU and memory resources are updated when creating or updating the container.

src/runtime-rs/crates/runtimes/common/src/runtime_handler.rs
Tim-0731-Hzt: @quanweiZhou @Ji-Xinyou I think I didn't express my opinion clearly. My concern is if you call `instance.update_sandbox_resource()`, if the runtime instance is virt-container, the current design works fine. If the runtime instance is wasm or Linux, `update_sandbox_resource()` may not work the same way as virt-container. wasm and Linux might have different implementations for `update_sandbox_resource()`

Tim-0731-Hzt: So my suggestion is to implement `update_sandbox_resource` for different runtime instance

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: This is for the case like
```
1. stop a container (the resource will not be deallocated)
2. start a container (the resource previously allocated is enough, so this start will not allocate any resource)
```

quanweiZhou: @Ji-Xinyou As you say for case 2 

> This is for the case like
> 
> ```
> 1. stop a container (the resource will not be deallocated)
> 2. start a container (the resource previously allocated is enough, so this start will not allocate any resource)
> ```
@Ji-Xinyou As you say in case 2 "the resource previously allocated is enough, so this start will not allocate any resource", so I think no need to call `update_sandbox_resource()`.

src/runtime-rs/crates/runtimes/src/manager.rs
YushuoEdge: Refer to https://github.com/kata-containers/kata-containers/blob/main/docs/design/host-cgroups.md
I mean, taking `sandbox_cgroup_only=false` and `cgroup v1` as an example, in the host, for each pod, we will have two cgroup_paths, /kubepods/kata_<sandbox_id> and /kata_overhead/<sandbox_id>. We name them as cgroup_controller and overhead_controller in runtime's code. We have to write the tid to each path's `tasks` so that the resources[on the host] each thread can use can be constrained.

When we have hotpluged new vcpus, there have spawned new vcpu threads. But the newly threads did not appear in any cgroup controller now, which indicates that the newly spawned vcpu threads can use any threads they want on the host.

As you have mentioned, although we do not have cpuset support now, but I think we should implement a more general  methods, which can more suitable for all host resources' management. Otherwise, we still need to refactor the code logic.

YushuoEdge: In addition, under your current implementation, we are experiencing:
1. start vm with default resources
2. update the host resources that pod can use when create container, sometimes we may need to set contrain in the guest, according to container's spec
3. update the vm's resources

There exists a logic problem, when we set resources' limits for each container in the guest, we're setting cgroups based on resources that there should have existed. (As from upper layer's perspective, we are setting cgroups according the host's resources.) But there did not exist enough in reality. This will lead to error.  Still, taking cpuset as an example,
If we already start the vm with default 1 vcpus, then there comes a container with 2 vcpus requested. If we have to set cpuset in the guest, assumping we are going to set `0-1` in the guest, then the guest OS will report error.

Same cases will also happen when containers request for hugepage, etc.

So, what I mean here is, we should finish preparing resources before we set cgroup, which is more natural logic.

src/runtime-rs/crates/runtimes/common/src/runtime_handler.rs
YushuoEdge: Please correct me if I am wrong. IMO, we have an abstraction, `ResourceManager`, which is independent from `Sandbox` and `ContainerManager`. I think we can implement the real `update_sandbox_resource()` in `ResourceManager`. Calling this function from `Sandbox` or `ContainerManager` are all ok for me. (If we calling the function from ContainerManager, we can use a sandbox_id parameter.)

src/runtime-rs/crates/runtimes/src/manager.rs
Ji-Xinyou: > @Ji-Xinyou As you say for case 2
> 
> > This is for the case like
> > ```
> > 1. stop a container (the resource will not be deallocated)
> > 2. start a container (the resource previously allocated is enough, so this start will not allocate any resource)
> > ```
> 
> @Ji-Xinyou As you say in case 2 "the resource previously allocated is enough, so this start will not allocate any resource", so I think no need to call `update_sandbox_resource()`.

I remember this was the 2.0 behavior.
I guess this was for those VMMs who support hot-unplug. To handle your case, we could only update the resource for such VMM, wdyt?

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -325,6 +325,61 @@ impl DragonballInner {
...
+        // cannot exceed maximum value
+        if new_vcpus > self.config.cpu_info.default_maxvcpus {
+            return Err(anyhow!("resize vcpu error: cannot greater than maxvcpus"));
+        }
+
+        Ok((current_vcpus, new_vcpus))
+    }
+
+    // do the check before resizing, returns Result<(old, new)>
+    pub async fn resize_vcpu(&self, old_vcpus: u32, new_vcpus: u32) -> Result<(u32, u32)> {
```
Tim-0731-Hzt, line 348: `pub(crate) async fn`

src/runtime-rs/crates/runtimes/src/manager.rs
quanweiZhou: Currently, the [`cpus_required()`](https://github.com/openanolis/kata-containers/blob/runtime_vcpu_resize/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs#L293) calculate all to container CPU without check the container state, so the CPU number only changes for `Create/Update/Delete Container` not the `Start Container`. If you want to include the `Pause State` of a container, the `cpus_required()` should consider the container state too, but I wonder if this case is needed.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/manager.rs
```diff
@@ -272,4 +277,54 @@ impl ContainerManager for VirtContainerManager {
...
+    // calculates the total required vpus by adding each containers' requirement within the pod
+    async fn cpus_required(&self) -> Result<u32> {
+        let mut total_vcpu = 0;
+        let mut cpuset_count = 0;
+        let cpu_mem_info = self.resource_manager.sandbox_cpu_mem_info().await;
+
+        let containers = self.containers.read().await;
+        for c in containers.values() {
+            if let Some(resource) = &c.linux_resources {
+                if let Some(cpu) = &resource.cpu {
```
Tim-0731-Hzt, line 301: @Ji-Xinyou @quanweiZhou we should also check the container state here, refer to [go runtime](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/sandbox.go#L2197). If the container state is stopped, we don't need to count the CPU here. To get the container state, we have to do it in the container manager, so I think it's inappropriate to implement it in Sandbox, wdyt @quanweiZhou ?

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -311,6 +325,13 @@ impl RuntimeHandlerManager {
             }
             Request::UpdateContainer(req) => {
                 cm.update_container(req).await.context("update container")?;
+
+                // update the sandbox's resource when a container's spec is updated
+                instance
```
Tim-0731-Hzt, line 330: @Ji-Xinyou Did you update the `oci spec` when the update has been called? It seems you are still using the original one?

Ji-Xinyou:
@YushuoEdge , the latest commit has no commit body and the message check has failed. Please fix that.

YushuoEdge:
> @YushuoEdge , the latest commit has no commit body and the message check has failed. Please fix that.

fixed

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,166 @@
...
+        Ok(Self {
+            current_vcpu: Arc::new(RwLock::new(hypervisor_config.cpu_info.default_vcpus as u32)),
+            container_cpu_resources: Arc::new(RwLock::new(HashMap::new())),
+        })
+    }
+
+    pub(crate) async fn update_cpu_resources(
+        &self,
+        cid: &str,
+        linux_resources: Option<&LinuxResources>,
```
Tim-0731-Hzt, line 46: I think you can just pass `LinuxCpu` into this function

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,166 @@
...
+                        resources.remove(cid);
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    // calculates the totoal required vcpus by adding each containers' requirements within the pod
```
Tim-0731-Hzt, line 105: typo: `total` `each container's`

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,166 @@
...
+    // do hotplug the vcpu
+    async fn do_update_cpu_resources(
+        &self,
+        new_vcpus: u32,
+        op: ResourceUpdateOp,
+        hypervisor: &dyn Hypervisor,
+        agent: &dyn Agent,
+    ) -> Result<u32> {
+        let old_vcpus = self.current_vcpu().await;
+        if (op == ResourceUpdateOp::Add && old_vcpus > new_vcpus)
```
Tim-0731-Hzt, line 139: suggest adding some comments for this

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            return Ok(old_vcpus);
+        }
+
+        let (old, new) = hypervisor
+            .resize_vcpu(old_vcpus, new_vcpus)
+            .await
+            .context("resize vcpus")?;
+
+        if old < new {
```
Tim-0731-Hzt, line 150: I am confusing why we do online only if  old < new, suggest adding some comment for this

src/runtime-rs/crates/resource/src/cpu_mem/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2019-2022 Alibaba Cloud
+// Copyright (c) 2019-2022 Ant Group
```
Tim-0731-Hzt, line 2: 2023

Tim-0731-Hzt (commented):
Thanks @YushuoEdge , few comments

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -264,6 +259,55 @@ impl ResourceManagerInner {
         self.rootfs_resource.dump().await;
         self.volume_resource.dump().await;
     }
+
+    pub async fn update_linux_resource(
```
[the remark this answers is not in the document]
YushuoEdge: Currently we have just considered cpu, memory, and host cgroups.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -264,6 +259,55 @@ impl ResourceManagerInner {
...
+        &self,
+        cid: &str,
+        linux_resources: Option<&LinuxResources>,
+        op: ResourceUpdateOp,
+    ) -> Result<Option<LinuxResources>> {
+        // if static_sandbox_resource_mgmt, we will not have to update sandbox's cpu or mem resource
+        if !self.toml_config.runtime.static_sandbox_resource_mgmt {
```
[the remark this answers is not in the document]
YushuoEdge: Yeah, we think it should be skipped.

`static_sandbox_resource_mgmt` means we will not modify sandbox resources anymore, which means we will not do any hotplug operations.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -264,6 +259,55 @@ impl ResourceManagerInner {
...
+            }
+        };
+
+        // clear the cpuset
+        // or example, if there are only 5 vcpus now, and the cpuset in LinuxResources is 0-2,6, guest os will report
+        // error when creating the container. so we choose to clear the cpuset here.
+        if let Some(cpu) = &mut resources.cpu {
+            cpu.cpus = String::new();
+        }
+
```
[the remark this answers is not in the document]
YushuoEdge: I think we may need to update something else if we want to do the resource allocation in the container level.

src/runtime-rs/crates/resource/src/manager_inner.rs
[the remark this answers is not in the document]
Tim-0731-Hzt: If device won't be handled here, maybe consider another name?

src/runtime-rs/crates/resource/src/manager_inner.rs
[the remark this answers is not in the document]
Tim-0731-Hzt: How about hot unplug, if the container exit ?

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub(crate) container_cpu_resources: Arc<RwLock<HashMap<String, LinuxContainerCpuResources>>>,
+}
+
+impl CpuResource {
+    pub fn new(config: Arc<TomlConfig>) -> Result<Self> {
+        let hypervisor_name = config.runtime.hypervisor_name.clone();
+        let hypervisor_config = config
+            .hypervisor
+            .get(&hypervisor_name)
+            .context("failed to get hypervisor")?;
```
quanweiZhou, line 36: Suggesting adding  `hypervisor name` for the context.

quanweiZhou (commented):
Thanks, @Ji-Xinyou @YushuoEdge Few comments.

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,166 @@
...
+        let (old, new) = hypervisor
+            .resize_vcpu(old_vcpus, new_vcpus)
+            .await
+            .context("resize vcpus")?;
+
+        if old < new {
+            let add = new - old;
+            info!(sl!(), "request to onlineCpuMem with {:?} cpus", add);
+
+            agent
```
Tim-0731-Hzt, line 154: I think you could call this function after both CPU and memory resource is updated, then we only need to call it once.

src/runtime-rs/crates/resource/src/manager_inner.rs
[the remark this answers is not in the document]
YushuoEdge: If the static_sandbox_resource_mgmt, we will not do the hot unplug when the container exits.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -264,6 +259,55 @@ impl ResourceManagerInner {
...
+                .await?;
+        }
+
+        // we should firstly update the vcpus and mems, and then update the host cgroups
+        self.cgroups_resource
+            .update_cgroups(cid, linux_resources, op, self.hypervisor.as_ref())
+            .await?;
+
+        // update the linux resources for agent
+        self.agent_linux_resources(linux_resources)
```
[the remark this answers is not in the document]
YushuoEdge: I am not sure whether it would be great if the name is changed to `update_agent_linux_resources`, since we will not truly update the linux resources in the guest. We just update the contents in the spec.

src/agent/src/sandbox.rs
```diff
@@ -432,22 +434,29 @@ fn online_resources(logger: &Logger, path: &str, pattern: &str, num: i32) -> Res
...
-            num - onlined_count,
-        );
+            num - onlined_cpu_count,
+        )
+        .context("online cpu resource")?;
 
-        onlined_count += r?;
-        if onlined_count == num {
+        onlined_cpu_count = onlined_cpus().context("onlined cpu count")?;
+        if onlined_cpu_count >= num {
```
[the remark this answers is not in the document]
YushuoEdge: This is also a sanity check, max number of onlined_cpu_count here is num (If we do proper operation in runtime).

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2100,9 +2100,8 @@ func (s *Sandbox) updateResources(ctx context.Context) error {
 	s.Logger().Debugf("Request to hypervisor to update oldCPUs/newCPUs: %d/%d", oldCPUs, newCPUs)
 	// If the CPUs were increased, ask agent to online them
 	if oldCPUs < newCPUs {
-		vcpusAdded := newCPUs - oldCPUs
-		s.Logger().Debugf("Request to onlineCPUMem with %d CPUs", vcpusAdded)
-		if err := s.agent.onlineCPUMem(ctx, vcpusAdded, true); err != nil {
+		s.Logger().Debugf("Request to onlineCPUMem with %d CPUs", newCPUs)
+		if err := s.agent.onlineCPUMem(ctx, newCPUs, true); err != nil {
```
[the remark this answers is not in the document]
YushuoEdge: This modification is for compatibility with dragonball (which will do the online operation for us). 

The original meaning: num of vcpus we should do online operation this time.
Meaning after this modification: the total num of vcpus being onlined we should guaranteed after this operation.

If we do not modify this rpc, and we do not want to judge which vmm we are currently using, since dragonball may online part of the vcpus for us, the agent may not be able to online a sufficient number of vcpus when do the online operation. This may cause an error to be reported almost every time the agent does an online operation when dragonball is used as the hypervisor.

src/runtime-rs/crates/resource/src/manager_inner.rs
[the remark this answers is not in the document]
YushuoEdge: I have checked the `handler_device` function. We handle the devices based on `LinuxDevice`, but here, we handle the resources based on `LinuxResources`. This function should not conclude the device handling.

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
YushuoEdge: I think we can move it out when we work on memory resize capability. Actually, I am not sure how to I am not very clear about what kind of judgment conditions should be used to modify the OnlineRequest.

quanweiZhou (approved):

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,167 @@
...
+        new_vcpus: u32,
+        op: ResourceUpdateOp,
+        hypervisor: &dyn Hypervisor,
+        agent: &dyn Agent,
+    ) -> Result<u32> {
+        let old_vcpus = self.current_vcpu().await;
+
+        // when adding vcpus, ignore old_vcpus > new_vcpus
+        // when deleting vcpus, ignore old_vcpus < new_vcpus
+        if (op == ResourceUpdateOp::Add && old_vcpus > new_vcpus)
```
justxuewei, line 140: Are there any inspections for the `ResourceUpdateOp::Update` option missing?

YushuoEdge: We can not do this kind of check for ResourceUpdateOp, for we do not know whether the resources are required more or less.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -340,6 +334,57 @@ impl ResourceManagerInner {
...
         self.volume_resource.dump().await;
     }
+
+    pub async fn update_linux_resource(
+        &self,
+        cid: &str,
+        linux_resources: Option<&LinuxResources>,
+        op: ResourceUpdateOp,
+    ) -> Result<Option<LinuxResources>> {
+        let linux_cpus = || -> Option<&LinuxCpu> { linux_resources.as_ref()?.cpu.as_ref() }();
```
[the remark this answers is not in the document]
YushuoEdge: Using a closure can avoid handling error type conflict.

docs/design/vcpu-handling-runtime-rs.md
```diff
@@ -0,0 +1,51 @@
...
+* `CreateVM`: When creating a sandbox, we need to know how many vCPUs to start the VM with.
+* `CreateContainer`: When creating a new container in the VM, we may need to hot-plug the vCPUs according to the requirements in container's spec.
+* `UpdateContainer`: When receiving the `UpdateContainer` request, we may need to update the vCPU resources according to the new requirements of the container.
+* `DeleteContainer`: When a container is removed from the VM, we may need to hot-unplug the vCPUs to reclaim the vCPU resources introduced by the container.
+
+## On what basis do we calculate the number of vCPUs?
+When I calculate the number of vCPUs, We have three data sources, the `default_vcpus` and `default_maxvcpus` specified in the configuratio file (named `TomlConfig` later in the doc), the `io.kubernetes.cri.sandbox-cpu-quota`, `io.kubernetes.cri.sandbox-cpu-period` and `io.kubernetes.cri.sandbox-cpu-shares` annotations passed by the upper layer runtime, and the corresponding CPU resource part in the container's spec for the container when `CreateContainer`/`UpdateContainer`/`DeleteContainer` is requested.
```
studychao, line 15: `I` -> `Kata`?

docs/design/vcpu-handling-runtime-rs.md
```diff
@@ -0,0 +1,51 @@
...
+   * We think quota describes the cpu time slice that a cgroup can use, and cpuset describes the actual cpu number that a cgroup can use. Quota can better describe the size of the CPU time slice that a cgroup actually wants to use. The cpuset only describes which CPUs the cgroup can use, but the cgroup can use the specified CPU but consumes a smaller time slice, so the quota takes precedence over the cpuset.
+   * On the one hand, when both cpu quota and cpuset are specified, we will calculate the number of vCPUs based on cpu quota and ignore cpuset. On the other hand, if cpu quota was used to control the number of vCPUs in the past, and only cpuset was updated during `UpdateContainer`, we will not adjust the number of vCPUs at this time.
+3. `StaticSandboxResourceMgmt` controls hotplug.
+   * Some VMMs and kernels of some architectures do not support hotplugging. We can accommodate this situation through `StaticSandboxResourceMgmt`. When `StaticSandboxResourceMgmt = true` is set, we don't make any further attempts to update the number of vCPUs after booting.
+
+
+[1]: https://github.com/kubernetes/kubernetes/pull/104886
```
studychao, line 50: Both links are for Kubernetes? I suppose there should be one for containerd?

docs/design/vcpu-handling-runtime-rs.md
```diff
@@ -0,0 +1,51 @@
...
+> In order to adapt to this beneficial change and be compatible with the past, we have implemented the new vCPUs handling way in `runtime-rs`, which is slightly different from the original `runtime-go`'s design.
+
+## When do we need to handle vCPUs size?
+vCPUs sizing should be determined by the container workloads. So throughout the life cycle of Kata Containers, there are several points in time when we need to think about how many vCPUs should be at the time. Mainly including the time points of `CreateVM`, `CreateContainer`, `UpdateContainer`, and `DeleteContainer`.
+* `CreateVM`: When creating a sandbox, we need to know how many vCPUs to start the VM with.
+* `CreateContainer`: When creating a new container in the VM, we may need to hot-plug the vCPUs according to the requirements in container's spec.
+* `UpdateContainer`: When receiving the `UpdateContainer` request, we may need to update the vCPU resources according to the new requirements of the container.
+* `DeleteContainer`: When a container is removed from the VM, we may need to hot-unplug the vCPUs to reclaim the vCPU resources introduced by the container.
```
studychao, line 12: Could this configurable? Doing cpu hotplug and hot-unplug too often may introduce stability problem to the worklord

docs/design/vcpu-handling-runtime-rs.md
studychao, line 15: Why `cpu share` has impact on cpu size?

docs/design/vcpu-handling-runtime-rs.md
```diff
@@ -0,0 +1,51 @@
+# Virtual machine vCPU sizing in Kata Containers 3.0
+
+> Preview: 
+> [Kubernetes(since 1.23)][1] and [Containerd(since 1.6.0-beta4)][2] will help calculate `Sandbox Size` info and pass it to Kata Containers through annotations.
+> In order to adapt to this beneficial change and be compatible with the past, we have implemented the new vCPUs handling way in `runtime-rs`, which is slightly different from the original `runtime-go`'s design.
```
studychao, line 5: nit : could you describe what behavior is different from go runtime? I assume is the support for `Sandbox Size` part ?

studychao (commented):
Nice work ! A few comments for the document.

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,187 @@
...
+            .resize_vcpu(old_vcpus, new_vcpus)
+            .await
+            .context("resize vcpus")?;
+
+        if old < new {
+            let add = new - old;
+            info!(sl!(), "request to onlineCpuMem with {:?} cpus", add);
+
+            agent
+                .online_cpu_mem(OnlineCPUMemRequest {
```
studychao, line 176: Why this request is called `OnlineCPUMemRequest` ? Do we want to interwine cpu & memory online into one single request? What's the reason for this mixed design?

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        info!(
+            sl!(),
+            "(from cfs_quota&cfs_period)get vcpus count {}", total_vcpu
+        );
+        Ok(total_vcpu)
+    }
+
+    // do hotplug the vcpu
```
studychao, line 144: nit: do hotplug & hot-unplug the vcpus

src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
```diff
@@ -100,8 +100,13 @@ impl StaticResourceManager {
             .hypervisor
             .get_mut(hypervisor_name)
             .context("failed to get hypervisor config")?;
-        hv.cpu_info.default_vcpus += self.resource.vcpu as i32;
-        hv.memory_info.default_memory += self.resource.mem_mb;
+
+        if self.resource.vcpu > 0 {
+            hv.cpu_info.default_vcpus = self.resource.vcpu as i32
```
studychao, line 105: Why `+=` is replaced with `=` here but `mem_mb` calculation stays the same way?

src/runtime-rs/crates/resource/src/manager.rs
```diff
@@ -4,22 +4,26 @@
...
 //
 
-use crate::network::NetworkConfig;
-use crate::resource_persist::ResourceState;
-use crate::{manager_inner::ResourceManagerInner, rootfs::Rootfs, volume::Volume, ResourceConfig};
-use agent::types::Device;
-use agent::{Agent, Storage};
 use anyhow::Result;
 use async_trait::async_trait;
+
```
studychao, line 9: suggest clean up the crates dependency here. 
rules as above

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -4,13 +4,12 @@
...
 //
 
-use std::{sync::Arc, thread, vec};
-
-use crate::{network::NetworkConfig, resource_persist::ResourceState};
-use agent::{types::Device, Agent, Storage};
 use anyhow::{anyhow, Context, Ok, Result};
 use async_trait::async_trait;
 
+use std::{sync::Arc, thread};
```
studychao, line 10: clean up crate dependencies.

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -354,6 +347,57 @@ impl ResourceManagerInner {
...
+
+        // update the linux resources for agent
+        self.agent_linux_resources(linux_resources)
+    }
+
+    fn agent_linux_resources(
+        &self,
+        linux_resources: Option<&LinuxResources>,
+    ) -> Result<Option<LinuxResources>> {
+        let mut resources = match linux_resources {
```
studychao, line 385: nit : Will `if let` better here instead of using `match`?

src/runtime-rs/crates/resource/src/manager_inner.rs
```diff
@@ -354,6 +347,57 @@ impl ResourceManagerInner {
...
+            Some(linux_resources) => linux_resources.clone(),
+            None => {
+                return Ok(None);
+            }
+        };
+
+        // clear the cpuset
+        // or example, if there are only 5 vcpus now, and the cpuset in LinuxResources is 0-2,6, guest os will report
```
studychao, line 393: `or example` -> `for example`

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -6,7 +6,7 @@
 
 use std::{path::PathBuf, str::from_utf8, sync::Arc};
 
-use crate::{shim_mgmt::server::MgmtServer, static_resource::StaticResourceManager};
+use crate::shim_mgmt::server::MgmtServer;
```
studychao, line 9: clean up.

studychao (commented):
Thanks, a few comments

src/libs/protocols/protos/agent.proto
```diff
@@ -366,7 +366,8 @@ message OnlineCPUMemRequest {
 	// resources are connected asynchronously and the agent returns immediately.
 	bool wait = 1;
 
-	// NbCpus specifies the number of CPUs that were added and the agent has to online.
+	// NbCpus specifies the number of CPUs that should be onlined in the guest.
+	// Special value 0 means agent will skip this check.
```
studychao, line 370: Where do you use `0` to skip the check in this PR?

docs/design/vcpu-handling-runtime-rs.md
YushuoEdge: The `cpu share` has no impact on the cpu size, it will just be passed to runtime from upper layer. I have deleted it.

docs/design/vcpu-handling-runtime-rs.md
YushuoEdge: The way we support sandbox size is different. In runtime-go, only we open the static_sandbox_resource_mgmt, the sandbox size will be effective.

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        let resources = self.container_cpu_resources.read().await;
+        for (_, cpu_resource) in resources.iter() {
+            let vcpu = cpu_resource.get_vcpus().unwrap_or(0) as u32;
+            cpuset_vcpu.extend(cpu_resource.cpuset().iter());
+            total_vcpu += vcpu;
+        }
+
+        // contrained only by cpuset
+        if total_vcpu == 0 && !cpuset_vcpu.is_empty() {
```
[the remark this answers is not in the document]
YushuoEdge: Only total_vcpu==0, cpuset will be effective. And total_vcpu is the sum of cpu_quota.

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
YushuoEdge: This request is inherited from runtime-go.

src/runtime-rs/crates/resource/src/manager_inner.rs
YushuoEdge: Because we only modify the cpuset now, it seems if let is better. But there are possibilities we will modify other fields, which may cause a lot of `if let`. Thus we use match here. Thanks for pointing this.

docs/design/vcpu-handling-runtime-rs.md
YushuoEdge: Thanks! Fixes.

docs/design/vcpu-handling-runtime-rs.md
YushuoEdge: Sorry, fixed.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -327,6 +327,50 @@ impl DragonballInner {
...
+            return Err(anyhow!("resize vcpu error: cannot greater than maxvcpus"));
+        }
+
+        Ok((current_vcpus, new_vcpus))
+    }
+
+    // do the check before resizing, returns Result<(old, new)>
+    pub(crate) async fn resize_vcpu(&self, old_vcpus: u32, new_vcpus: u32) -> Result<(u32, u32)> {
+        let (old_vcpus, new_vcpus) = self.precheck_resize_vcpus(old_vcpus, new_vcpus)?;
+        if old_vcpus == new_vcpus {
```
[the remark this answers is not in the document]
YushuoEdge: Thanks, fixed!

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -327,6 +327,50 @@ impl DragonballInner {
...
+            info!(sl!(), "resize_vcpu: no need to resize vcpus");
```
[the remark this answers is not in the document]
YushuoEdge: Thanks, fixed!

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) 2019-2022 Alibaba Cloud
+// Copyright (c) 2019-2022 Ant Group
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use agent::{Agent, OnlineCPUMemRequest};
```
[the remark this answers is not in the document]
YushuoEdge: Thanks! Fixed.

src/runtime-rs/crates/resource/src/cpu_mem/cpu.rs
YushuoEdge: Thanks, fixed.

src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
YushuoEdge: Thanks, fixed.

src/runtime-rs/crates/resource/src/manager.rs
YushuoEdge: Thanks, fixed.

src/runtime-rs/crates/resource/src/manager_inner.rs
YushuoEdge: Thanks, fixed.

src/runtime-rs/crates/resource/src/manager_inner.rs
YushuoEdge: Thanks, fixed.

src/libs/protocols/protos/agent.proto
YushuoEdge: in L266-L279 src/agent/src/sandbox.rs

src/runtime-rs/crates/runtimes/src/manager.rs
YushuoEdge: Thanks, fixed.

Tim-0731-Hzt (approved):

Issue #6291, opened 2023-02-15
runtime-rs: Add more default config paths


fidencio:
This is not ideal, but this is what it is right now, as the
configurations end up being loaded from that list in case
`KATA_CONF_FILE` is not set.

As part of our CI, as part of kata-deploy, and as part of the payload we
provide to Confidential Containers, we do not set `KATA_CONF_FILE`, thus
expanding the list seems like the most reasonable approach to take, at
least till github.com/kata-containers/kata-containers#3961 gets solved.

While expanding the list, let's also make sure to add the
`-dragonball.toml` as first options as nothing guarantees that the
default configuration will be pointing to that one.

Issue #6290, opened 2023-02-15
runtime-rs: Use the correct kernel name


fidencio:
When calling `MAKE_KERNEL_NAME` we're considering the default kernel
name will be `vmlinux.container` or `vmlinuz.container`, which is not
the case as the runtime-rs, when used with dragonball, relies on the
`vmlinu[zx]-dragonball-experimental.container` kernel.

Other hypervisors will have to introduce a similar
`MAKE_KERNEL_NAME_${HYPERVISOR}` to adapt this to the kernel they want
to use, similarly to what's already done for the go runtime.

By doing this we also ensure that no changes in the configuration file
will be required to run runtime-rs, with dragonball, as part of our CI
or as part of kata-deploy.

Issue #3961, opened 2022-03-25
runtime-rs: support shimv2 create task option


quanweiZhou:
runtime-rs: support shimv2 create task option, relay on `shim.proto`
from https://github.com/kata-containers/kata-containers/pull/3879#discussion_r831941742

Options:
https://github.com/containerd/containerd/blob/main/pkg/runtimeoptions/v1/api.proto#L19

studychao:
This issue is already fixed by : https://github.com/kata-containers/kata-containers/commit/d663f110d737a1362d3d49bf58aaea9fd7592271

studychao:
reopen this since we need a commit to fix the comment area above `load_config`

PR #6292, opened 2023-02-15
runtime-rs: fix default kernel location and add more default config paths

fidencio:
---

build: Use the correct kernel name

When calling `MAKE_KERNEL_NAME` we're considering the default kernel
name will be `vmlinux.container` or `vmlinuz.container`, which is not
the case as the runtime-rs, when used with dragonball, relies on the
`vmlinu[zx]-dragonball-experimental.container` kernel.

Other hypervisors will have to introduce a similar
`MAKE_KERNEL_NAME_${HYPERVISOR}` to adapt this to the kernel they want
to use, similarly to what's already done for the go runtime.

By doing this we also ensure that no changes in the configuration file
will be required to run runtime-rs, with dragonball, as part of our CI
or as part of kata-deploy.

Fixes: #6290

---

kata-types/config: Add more default config paths

This is not ideal, but this is what it is right now, as the
configurations end up being loaded from that list in case
`KATA_CONF_FILE` is not set.

As part of our CI, as part of kata-deploy, and as part of the payload we
provide to Confidential Containers, we do not set `KATA_CONF_FILE`, thus
expanding the list seems like the most reasonable approach to take, at
least till github.com/kata-containers/kata-containers#3961 gets solved.

While expanding the list, let's also make sure to add the
`-dragonball.toml` as first options as nothing guarantees that the
default configuration will be pointing to that one.

Fixes: #6291

---

### DIFF
--- a/src/runtime-rs/Makefile
+++ b/src/runtime-rs/Makefile
@@ -203,7 +203,7 @@ ifneq (,$(DBCMD))
     DEFNETWORKMODEL_DB := tcfilter
     KERNELPARAMS = console=ttyS1 agent.log_vport=1025
 	KERNELTYPE_DB = uncompressed
-    KERNEL_NAME_DB = $(call MAKE_KERNEL_NAME,$(KERNELTYPE_DB))
+    KERNEL_NAME_DB = $(call MAKE_KERNEL_NAME_DB,$(KERNELTYPE_DB))
     KERNELPATH_DB = $(KERNELDIR)/$(KERNEL_NAME_DB)
     DEFSANDBOXCGROUPONLY = true
     RUNTIMENAME := virt_container
@@ -371,8 +371,8 @@ endef
 
 # Returns the name of the kernel file to use based on the provided KERNELTYPE.
 # $1 : KERNELTYPE (compressed or uncompressed)
-define MAKE_KERNEL_NAME
-$(if $(findstring uncompressed,$1),vmlinux.container,vmlinuz.container)
+define MAKE_KERNEL_NAME_DB
+$(if $(findstring uncompressed,$1),vmlinux-dragonball-experimental.container,vmlinuz-dragonball-experimental.container)
 endef
 
 .DEFAULT_GOAL := default


src/libs/kata-types/src/config/default.rs
```diff
@@ -14,8 +14,14 @@ use lazy_static::lazy_static;
 lazy_static! {
     /// Default configuration file paths, vendor may extend the list
     pub static ref DEFAULT_RUNTIME_CONFIGURATIONS: Vec::<&'static str> = vec![
+        "/etc/kata-containers/configuration-dragonball.toml",
+        "/usr/share/defaults/kata-containers/configuration-dragonball.toml",
+        "/opt/kata/share/defaults/kata-containers/configuration-dragonball.toml",
+        "/opt/confindetial-containers/share/defaults/kata-containers/configuration-dragonball.toml",
```
studychao, line 20: Is creating a soft link from `/../../configuration.toml` to `../../configuration-dragonball.toml` during CI environment builtup process or user installation process a better way?

studychao (commented):
Hi @fidencio , LGTM on the kernel name one.
And a little question on the default config path

src/runtime-rs/Makefile
```diff
@@ -371,8 +371,8 @@ endef
 
 # Returns the name of the kernel file to use based on the provided KERNELTYPE.
 # $1 : KERNELTYPE (compressed or uncompressed)
-define MAKE_KERNEL_NAME
-$(if $(findstring uncompressed,$1),vmlinux.container,vmlinuz.container)
+define MAKE_KERNEL_NAME_DB
+$(if $(findstring uncompressed,$1),vmlinux-dragonball-experimental.container,vmlinuz-dragonball-experimental.container)
```
Ji-Xinyou, line 375: Will this affect that the runtime-rs will support more hypervisor in the future? It seems that this assumes runtime-rs only works with dragonball (although it does now).

src/libs/kata-types/src/config/default.rs
fidencio: So, it depends a lot.  CI does that, but kata-deploy and in the payload we provide to Confidential Containers we don't do that.

The reason for that is because the default runtime and the default VMM are still the go one and QEMU.  So, IOW, if we just link `configuration-dragonball.toml` to `configuration.toml` we break the current behaviour and the current assumption of the users.

src/runtime-rs/Makefile
studychao: This one has a `DB` postfix.
So I believe when CLH/QEMU/other hypervisors are supported, they could introduce xxxx_CLH/xxxx_QEMU

fidencio: What other VMMs will have to do is create a new `MAKE_KERNEL_NAME` for their own usage.  But that's something they'll have to do when they become supported (which is not the case now).

src/libs/kata-types/src/config/default.rs
studychao: Will this change cause go runtime to use `configuration-dragonball.toml` since it's the first option? (If go runtime and rust runtime are both installed in some cases? I think that is a situation in kata-deploy if I remember correctly)

fidencio: Yep, that's exactly the situation with kata-deploy, where we'll have both the go and the rust runtime "installed" at the same time.

The changes that I did were to avoid having any kind of conflict in that specific case (which is also the case for Confidential Containers).

If we just linked `configuration-dragonball.toml` toi `configuration.toml` the `kata` runtime class would pick the go runtime, to use with QEMU, but using the DB configuration, and that would simply explode. :-)

Is there something else that I could try to explain in a better way?

fidencio: By the way, http://jenkins.katacontainers.io/job/kata-containers-dragonball-tests-ubuntu-20.04-PR/372/console now has passed the point of the issues that I had before this PR. \o/

src/runtime-rs/Makefile
Ji-Xinyou: > What other VMMs will have to do is create a new `MAKE_KERNEL_NAME` for their own usage. But that's something they'll have to do when they become supported (which is not the case now).

I see, then this may affect the work related to @jodh-intel and @pmores. But I think this is good.

src/libs/kata-types/src/config/default.rs
studychao: ohhh I got it. I misunderstood that `DEFAULT_RUNTIME_CONFIGURATIONS ` will also be used in go-runtime. So I thought this will influence go runtiem cause they will also choose dragonball config as default. My mistake, sorry.

But if later CLH/QEMU introduce their support for runtime-rs, will this change here makes it hard for them to use their config? Maybe a FIXME or TODO comments leave here?

fidencio: > But if later CLH/QEMU introduce their support for runtime-rs, will this change here makes it hard for them to use their config?

That's a valid point, but I sincerely hope we are able to get rid of this by having https://github.com/kata-containers/kata-containers/issues/3961 implemented.

fidencio: I've updated the patch with a TODO comment.
Thanks for the review!

studychao:
Thanks！LGTM：）

studychao (approved):

studychao:
@fidencio seems like dragonball-ci is broken, maybe because how the kata test choose guest kernel, I'll also help to take a look at this.

studychao:
it's kinda strange that this PR breaks k8s-seccomp.bats in both Dragonball CIs.. I will do a retest

src/libs/kata-types/src/config/default.rs
studychao: Thanks!

fidencio:
> it's kinda strange that this PR breaks k8s-seccomp.bats in both Dragonball CIs.. I will do a retest

I'd like to check whether it will also break with my current tests PR (I'm in the process of doing that, right now fixing power issues), if the tests pass with the tests PR better, and we just hold this one till we get the other one solved.

amshinde (approved):

src/libs/kata-types/src/config/default.rs
```diff
@@ -13,9 +13,19 @@ use lazy_static::lazy_static;
 
 lazy_static! {
     /// Default configuration file paths, vendor may extend the list
+    /// TODO:
+    /// * Drop the `-dragonvall.toml` configs once
```
jodh-intel, line 17: Nit: typo.

```suggestion
    /// * Drop the `-dragonball.toml` configs once
```

src/libs/kata-types/src/config/default.rs
```diff
@@ -13,9 +13,19 @@ use lazy_static::lazy_static;
...
+    ///   implemented.
     pub static ref DEFAULT_RUNTIME_CONFIGURATIONS: Vec::<&'static str> = vec![
+        "/etc/kata-containers/configuration-dragonball.toml",
+        "/usr/share/defaults/kata-containers/configuration-dragonball.toml",
+        "/opt/kata/share/defaults/kata-containers/configuration-dragonball.toml",
+        "/opt/confindetial-containers/share/defaults/kata-containers/configuration-dragonball.toml",
         "/etc/kata-containers/configuration.toml",
         "/usr/share/defaults/kata-containers/configuration.toml",
+        "/opt/kata/share/defaults/kata-containers/configuration.toml",
+        "/opt/confindetial-containers/share/defaults/kata-containers/configuration.toml",
```
jodh-intel, line 28: Typo bug:

```suggestion
        "/opt/confidential-containers/share/defaults/kata-containers/configuration.toml",
```

src/runtime-rs/Makefile
```diff
@@ -203,7 +203,7 @@ ifneq (,$(DBCMD))
     DEFNETWORKMODEL_DB := tcfilter
     KERNELPARAMS = console=ttyS1 agent.log_vport=1025
 	KERNELTYPE_DB = uncompressed
-    KERNEL_NAME_DB = $(call MAKE_KERNEL_NAME,$(KERNELTYPE_DB))
+    KERNEL_NAME_DB = $(call MAKE_KERNEL_NAME_DB,$(KERNELTYPE_DB))
```
jodh-intel, line 206: I know this is following the existing pattern, but this name is potentially a little confusing: I read "DB" as "database" initially :smile:

src/runtime-rs/Makefile
jodh-intel: Thanks for the ping @Ji-Xinyou.

I think this change is fine, but if we need to add further logic, I'd be in favour of making the `Makefile` call a utility script to return the name rather than having the logic in the `Makefile`. That would be clearer and easier to maintain.

jodh-intel (commented):

GabyCT (approved):

fidencio:
This PR depends-on https://github.com/kata-containers/tests/pull/5477, let's merge the tests one first and only after that we merge this one.

fidencio:
I dropped the changed related to adding more default paths, as this is actually not needed.
The other changes are still needed, though, as we progress on merging CCv0 content to main.

Issue #6293, opened 2023-02-15
packaging/shim-v2: Only change the config if the file exists


fidencio:
Let's not try to sed a file that doesn't exist, which may be the case
depending on the architecture we're building the shim-v2 for.

PR #6294, opened 2023-02-15
packaging/shim-v2: Only change the config if the file exists

fidencio:
Let's not try to sed a file that doesn't exist, which may be the case depending on the architecture we're building the shim-v2 for.

This is a partial-forward port of
f24c47ea47e39bdc28bb740d5281eba21fa5727d.

Fixes: #6293


dborquez (approved):
lgtm, thanks @fidencio

Ji-Xinyou (approved):
LGTM

BbolroC (approved):
lgtm, thanks a lot! :wink:

Issue #6132, opened 2023-01-26
Implement generic kernel module checks


jodh-intel:
> **Note:**
>
> This issue is to ensure that a number of hardware-agnostic kernel modules are loaded (unlike the related, but Intel specific #5332).

Requirements for this issue:

- Ensure the `vhost` module is loaded or available.
- Ensure the `vhost_net` module is loaded or available.
- Ensure the `vhost_vsock` module is loaded or available.

This breaks down into the following sub-tasks:

- [ ] Rewrite [`RunningOnVMM()`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor.go#L858) in rust.
- [ ] Rewrite [`checkKernelModules()`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-check.go#L218) in rust.
- [ ] Move the agent's [`load_kernel_module()`](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/rpc.rs#L2002) function into [`src/libs/kata-sys-util`](https://github.com/kata-containers/kata-containers/tree/main/src/libs/kata-sys-util) so it can be used by the agent and `kata-ctl`.
- [ ] Rewrite [`haveKernelModule()`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-check.go#L135) in rust (depends on `load_kernel_module()`).

## Notes

- The golang code will call `modprobe $module` if the command is running as `root`. We should consider **not** doing this by default even when running as `root` to avoid the command changing the system state. The command should fail gracefully with a clear message stating the admin should run `sudo modprobe $module` before re-running this command. However, we should provide a command-line option to allow the admin to get similar behaviour to the existing golang `kata-runtime check` (particularly for backwards compatibility - see #6090).

- All functions need unit tests (even if those tests simply check that, for example, attempting to `modprobe` a kernel module will fail if the user running the command is non-`root`). Again, care needs to be taken about actually calling `sudo modprobe $module` in a CI environment so we may need to have the code include some test-conditional logic:
  ```rust
     #[cfg(test)]
    // Report that "modprobe $module" *would* be run.

    #[cfg(not(test))]
    // Actually call "modprobe $module"
  ```

cmaf:
This issue will be closed by https://github.com/kata-containers/kata-containers/pull/6800, making https://github.com/kata-containers/kata-containers/pull/6944

PR #6295, opened 2023-02-15
kata-ctl: checks for kvm, kvm_intel modules loaded

### LINKED ISSUE #5332: kata-ctl: check: x86_64: Write kernel module checks
This needs to handle the scenario where the running kernel does not have the appropriate module loaded (so it will need to call `/sbin/modprobe`). See the golang code for further details.

See: #4499

dborquez:
Ensure that kvm and kvm_intel modules are loaded

Fixes #5332

### DIFF
--- a/src/tools/kata-ctl/src/arch/aarch64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/aarch64/mod.rs
@@ -20,7 +20,7 @@ mod arch_specific {
 
     // List of check functions
     static CHECK_LIST: &[CheckItem] = &[CheckItem {
-        name: CheckType::CheckCpu,
+        name: CheckType::Cpu,
         descr: "This parameter performs the host check",
         fp: check,
         perm: PermissionType::NonPrivileged,
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -21,16 +21,43 @@ mod arch_specific {
     const CPUINFO_FLAGS_TAG: &str = "flags";
     const CPU_FLAGS_INTEL: &[&str] = &["lm", "sse4_1", "vmx"];
     const CPU_ATTRIBS_INTEL: &[&str] = &["GenuineIntel"];
+    const VMM_FLAGS: &[&str] = &["hypervisor"];
+
     pub const ARCH_CPU_VENDOR_FIELD: &str = check::GENERIC_CPU_VENDOR_FIELD;
     pub const ARCH_CPU_MODEL_FIELD: &str = check::GENERIC_CPU_MODEL_FIELD;
 
     // List of check functions
-    static CHECK_LIST: &[CheckItem] = &[CheckItem {
-        name: CheckType::CheckCpu,
-        descr: "This parameter performs the cpu check",
-        fp: check_cpu,
-        perm: PermissionType::NonPrivileged,
-    }];
+    static CHECK_LIST: &[CheckItem] = &[
+        CheckItem {
+            name: CheckType::Cpu,
+            descr: "This parameter performs the cpu check",
+            fp: check_cpu,
+            perm: PermissionType::NonPrivileged,
+        },
+        CheckItem {
+            name: CheckType::KernelModules,
+            descr: "This parameter performs the kvm check",
+            fp: check_kernel_modules,
+            perm: PermissionType::NonPrivileged,
+        },
+    ];
+
+    static MODULE_LIST: &[KernelModule] = &[
+        KernelModule {
+            name: "kvm",
+            parameter: KernelParam {
+                name: "kvmclock_periodic_sync",
+                value: KernelParamType::Simple("Y"),
+            },
+        },
+        KernelModule {
+            name: "kvm_intel",
+            parameter: KernelParam {
+                name: "unrestricted_guest",
+                value: KernelParamType::Predicate(unrestricted_guest_param_check),
+            },
+        },
+    ];
 
     pub fn get_checks() -> Option<&'static [CheckItem<'static>]> {
         Some(CHECK_LIST)
@@ -140,6 +167,120 @@ mod arch_specific {
         }
 
         Ok(GuestProtection::NoProtection)
+    }
+
+    fn running_on_vmm() -> Result<bool> {
+        match check::get_single_cpu_info(check::PROC_CPUINFO, CPUINFO_DELIMITER) {
+            Ok(cpu_info) => {
+                // check if the 'hypervisor' flag exist in the cpu features
+                let missing_hypervisor_flag = check::check_cpu_attribs(&cpu_info, VMM_FLAGS)?;
+
+                if missing_hypervisor_flag.is_empty() {
+                    return Ok(true);
+                }
+            }
+            Err(e) => {
+                return Err(anyhow!(
+                    "Unable to determine if the OS is running on a VM: {}: {}",
+                    e,
+                    check::PROC_CPUINFO
+                ));
+            }
+        }
+
+        Ok(false)
+    }
+
+    // check the host kernel parameter value is valid
+    // and check if we are running inside a VMM
+    fn unrestricted_guest_param_check(
+        module: &str,
+        param_name: &str,
+        param_value_host: &str,
+    ) -> Result<()> {
+        let expected_param_value: char = 'Y';
+
+        let running_on_vmm_alt = running_on_vmm()?;
+
+        if running_on_vmm_alt {
+            let msg = format!("You are running in a VM, where the kernel module '{}' parameter '{:}' has a value '{:}'. This causes conflict when running kata.",
+                module,
+                param_name,
+                param_value_host
+            );
+            return Err(anyhow!(msg));
+        }
+
+        if param_value_host == expected_param_value.to_string() {
+            Ok(())
+        } else {
+            let error_msg = format!(
+                "Kernel Module: '{:}' parameter '{:}' should have value '{:}', but found '{:}.'.",
+                module, param_name, expected_param_value, param_value_host
+            );
+
+            let action_msg = format!("Remove the '{:}' module using `rmmod` and then reload using `modprobe`, setting '{:}={:}'",
+                module,
+                param_name,
+                expected_param_value
+            );
+
+            return Err(anyhow!("{} {}", error_msg, action_msg));
+        }
+    }
+
+    fn check_kernel_param(
+        module: &str,
+        param_name: &str,
+        param_value_host: &str,
+        param_type: KernelParamType,
+    ) -> Result<()> {
+        match param_type {
+            KernelParamType::Simple(param_value_req) => {
+                if param_value_host != param_value_req {
+                    return Err(anyhow!(
+                        "Kernel module '{}': parameter '{}' should have value '{}', but found '{}'",
+                        module,
+                        param_name,
+                        param_value_req,
+                        param_value_host
+                    ));
+                }
+                Ok(())
+            }
+            KernelParamType::Predicate(pred_func) => {
+                pred_func(module, param_name, param_value_host)
+            }
+        }
+    }
+
+    fn check_kernel_modules(_args: &str) -> Result<()> {
+        println!("INFO: check kernel modules for: x86_64");
+
+        for module in MODULE_LIST {
+            let module_loaded =
+                check::check_kernel_module_loaded(module.name, module.parameter.name);
+
+            match module_loaded {
+                Ok(param_value_host) => {
+                    let parameter_check = check_kernel_param(
+                        module.name,
+                        module.parameter.name,
+                        &param_value_host,
+                        module.parameter.value.clone(),
+                    );
+
+                    match parameter_check {
+                        Ok(_v) => println!("{} Ok", module.name),
+                        Err(e) => return Err(e),
+                    }
+                }
+                Err(err) => {
+                    eprintln!("WARNING {:}", err.replace('\n', ""))
+                }
+            }
+        }
+        Ok(())
     }
 }
 
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -9,6 +9,10 @@ use anyhow::{anyhow, Result};
 use reqwest::header::{CONTENT_TYPE, USER_AGENT};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+
+#[cfg(any(target_arch = "x86_64"))]
+use std::process::{Command, Stdio};
+
 #[derive(Debug, Deserialize, Serialize, PartialEq)]
 struct Release {
     tag_name: String,
@@ -16,6 +20,12 @@ struct Release {
     created_at: String,
     tarball_url: String,
 }
+
+#[allow(dead_code)]
+const MODPROBE_PATH: &str = "/sbin/modprobe";
+
+#[allow(dead_code)]
+const MODINFO_PATH: &str = "/sbin/modinfo";
 
 const KATA_GITHUB_RELEASE_URL: &str =
     "https://api.github.com/repos/kata-containers/kata-containers/releases";
@@ -29,6 +39,7 @@ const ERR_NO_CPUINFO: &str = "cpu_info string is empty";
 
 #[allow(dead_code)]
 pub const GENERIC_CPU_VENDOR_FIELD: &str = "vendor_id";
+
 #[allow(dead_code)]
 pub const GENERIC_CPU_MODEL_FIELD: &str = "model name";
 
@@ -36,8 +47,8 @@ pub const GENERIC_CPU_MODEL_FIELD: &str = "model name";
 pub const PROC_CPUINFO: &str = "/proc/cpuinfo";
 
 #[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
-fn get_cpu_info(cpu_info_file: &str) -> Result<String> {
-    let contents = std::fs::read_to_string(cpu_info_file)?;
+fn read_file_contents(file_path: &str) -> Result<String> {
+    let contents = std::fs::read_to_string(file_path)?;
     Ok(contents)
 }
 
@@ -45,7 +56,7 @@ fn get_cpu_info(cpu_info_file: &str) -> Result<String> {
 // the specified cpuinfo file by parsing based on a specified delimiter
 #[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
 pub fn get_single_cpu_info(cpu_info_file: &str, substring: &str) -> Result<String> {
-    let contents = get_cpu_info(cpu_info_file)?;
+    let contents = read_file_contents(cpu_info_file)?;
 
     if contents.is_empty() {
         return Err(anyhow!(ERR_NO_CPUINFO));
@@ -57,7 +68,6 @@ pub fn get_single_cpu_info(cpu_info_file: &str, substring: &str) -> Result<Strin
         .ok_or("error splitting contents of cpuinfo")
         .map_err(|e| anyhow!(e))?
         .to_string();
-
     Ok(result)
 }
 
@@ -70,7 +80,7 @@ pub fn get_cpu_flags(cpu_info: &str, cpu_flags_tag: &str) -> Result<String> {
     }
 
     if cpu_flags_tag.is_empty() {
-        return Err(anyhow!("cpu flags delimiter string is empty"));
+        return Err(anyhow!("cpu flags delimiter string is empty"))?;
     }
 
     let subcontents: Vec<&str> = cpu_info.split('\n').collect();
@@ -220,6 +230,86 @@ pub fn check_official_releases() -> Result<()> {
     }
 
     Ok(())
+}
+
+#[cfg(any(target_arch = "x86_64"))]
+pub fn check_kernel_module_loaded(module: &str, parameter: &str) -> Result<String, String> {
+    const MODPROBE_PARAMETERS_DRY_RUN: &str = "--dry-run";
+    const MODPROBE_PARAMETERS_FIRST_TIME: &str = "--first-time";
+    const MODULES_PATH: &str = "/sys/module";
+
+    let status_modinfo_success;
+
+    // Partial check w/ modinfo
+    // verifies that the module exists
+    match Command::new(MODINFO_PATH)
+        .arg(module)
+        .stdout(Stdio::piped())
+        .output()
+    {
+        Ok(v) => {
+            status_modinfo_success = v.status.success();
+
+            // The module is already not loaded.
+            if !status_modinfo_success {
+                let msg = String::from_utf8_lossy(&v.stderr).replace('\n', "");
+                return Err(msg);
+            }
+        }
+        Err(_e) => {
+            let msg = format!(
+                "Command {:} not found, verify that `kmod` package is already installed.",
+                MODINFO_PATH,
+            );
+            return Err(msg);
+        }
+    }
+
+    // Partial check w/ modprobe
+    // check that the module is already loaded
+    match Command::new(MODPROBE_PATH)
+        .arg(MODPROBE_PARAMETERS_DRY_RUN)
+        .arg(MODPROBE_PARAMETERS_FIRST_TIME)
+        .arg(module)
+        .stdout(Stdio::piped())
+        .output()
+    {
+        Ok(v) => {
+            // a successful simulated modprobe insert, means the module is not already loaded
+            let status_modprobe_success = v.status.success();
+
+            if status_modprobe_success && status_modinfo_success {
+                // This condition is true in the case that the module exist, but is not already loaded
+                let msg = format!("The kernel module `{:}` exist but is not already loaded. Try reloading it using 'modprobe {:}=Y'",
+                        module, module
+                    );
+                return Err(msg);
+            }
+        }
+
+        Err(_e) => {
+            let msg = format!(
+                "Command {:} not found, verify that `kmod` package is already installed.",
+                MODPROBE_PATH,
+            );
+            return Err(msg);
+        }
+    }
+
+    let module_path = format!("{}/{}/parameters/{}", MODULES_PATH, module, parameter);
+
+    // Here the currently loaded kernel parameter value
+    // is retrieved and returned on success
+    match read_file_contents(&module_path) {
+        Ok(result) => Ok(result.replace('\n', "")),
+        Err(_e) => {
+            let msg = format!(
+                "'{:}' kernel module parameter `{:}` not found.",
+                module, parameter
+            );
+            Err(msg)
+        }
+    }
 }
 
 #[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
@@ -412,5 +502,65 @@ mod tests {
         assert!(!v.major.to_string().is_empty());
         assert!(!v.minor.to_string().is_empty());
         assert!(!v.patch.to_string().is_empty());
+    }
+
+    #[cfg(any(target_arch = "x86_64"))]
+    #[test]
+    fn check_module_loaded() {
+        #[allow(dead_code)]
+        #[derive(Debug)]
+        struct TestData<'a> {
+            module_name: &'a str,
+            param_name: &'a str,
+            param_value: &'a str,
+            result: Result<String>,
+        }
+
+        let tests = &[
+            // Failure scenarios
+            TestData {
+                module_name: "",
+                param_name: "",
+                param_value: "",
+                result: Err(anyhow!("modinfo: ERROR: Module {} not found.", "")),
+            },
+            TestData {
+                module_name: "kvm",
+                param_name: "",
+                param_value: "",
+                result: Err(anyhow!(
+                    "'{:}' kernel module parameter `{:}` not found.",
+                    "kvm",
+                    ""
+                )),
+            },
+            // Success scenarios
+            TestData {
+                module_name: "kvm",
+                param_name: "kvmclock_periodic_sync",
+                param_value: "Y",
+                result: Ok("Y".to_string()),
+            },
+        ];
+
+        for (i, d) in tests.iter().enumerate() {
+            let msg = format!("test[{}]: {:?}", i, d);
+            let result = check_kernel_module_loaded(d.module_name, d.param_name);
+            let msg = format!("{}, result: {:?}", msg, result);
+
+            if d.result.is_ok() {
+                assert_eq!(
+                    result.as_ref().unwrap(),
+                    d.result.as_ref().unwrap(),
+                    "{}",
+                    msg
+                );
+                continue;
+            }
+
+            let expected_error = format!("{}", &d.result.as_ref().unwrap_err());
+            let actual_error = result.unwrap_err().to_string();
+            assert!(actual_error == expected_error, "{}", msg);
+        }
     }
 }
--- a/src/tools/kata-ctl/src/types.rs
+++ b/src/tools/kata-ctl/src/types.rs
@@ -12,8 +12,9 @@ pub type BuiltinCmdFp = fn(args: &str) -> Result<()>;
 // CheckType encodes the name of each check provided by kata-ctl.
 #[derive(Debug, strum_macros::Display, EnumString, PartialEq)]
 pub enum CheckType {
-    CheckCpu,
-    CheckNetwork,
+    Cpu,
+    Network,
+    KernelModules,
 }
 
 // PermissionType is used to show whether a check needs to run with elevated (super-user)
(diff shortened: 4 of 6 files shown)


src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -61,4 +69,9 @@ mod arch_specific {
 
         Ok(())
     }
+
+    fn check_kernel_modules(_args: &str) -> Result<()> {
```
jodh-intel, line 73: @dborquez - When this function is implemented, it would call a generic function that will check if a kernel module is loaded (or if the feature is built into the kernel). Something like:

```rust
check_kernel_module_loaded("kvm")?;
check_kernel_module_loaded("kvm_intel")?;
```

That way, the PR for #6132 could also call this generic function. For example:

```rust
check_kernel_module_loaded("vhost_net")?;
```

In reality, this generic function will need to accept more than just the module name because we need to also check that module parameters are correct too (for example `kvm_intel` should be loaded with the `unrestricted_guest=Y` parameter for that module).

jodh-intel (commented):
Thanks @dborquez - looks good so far.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
dborquez: @jodh-intel thank you, I have added a `KernelModule` type structure that wraps the _name_ of the module and a map with the _(parameter,value)_ pair. So `check_kernel_module_loaded()` accepts this type of data.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -59,6 +89,22 @@ mod arch_specific {
...
+
+        println!("INFO: check kernel modules: x86_64");
+
+        for module in MOD_LIST.iter() {
+            let _k = check::check_kernel_module_loaded(module);    
+            let res = check_kernel_module_parameters(module);
+
+            // TODO: use (k,res) to generate a Result
+            println!("{res:?}");
+        }
```
cmaf, line 105: Right now, the program will panic if check_kernel_module_loaded fails, is that correct?

cmaf (commented):
@dborquez Have a question regarding failure

src/tools/kata-ctl/src/arch/x86_64/mod.rs
c3d: @cmaf: It looks like the intent was to use `check_kernel_module_loaded(module)?`, according to @jodh-intel's comment above:

> @dborquez - When this function is implemented, it would call a generic function that will check if a kernel module is loaded (or if the feature is built into the kernel). Something like:
> 
> check_kernel_module_loaded("kvm")?;
> check_kernel_module_loaded("kvm_intel")?;
> That way, the PR for https://github.com/kata-containers/kata-containers/issues/6132 could also call this generic function. For example:
> 
> check_kernel_module_loaded("vhost_net")?;
> In reality, this generic function will need to accept more than just the module name because we need to also check that module parameters are correct too (for example kvm_intel should be loaded with the unrestricted_guest=Y parameter for that module).
> 
>

dborquez: thanks @cmaf and @c3d , thank you for your inputs. The PR was wip, but it has been updated. 
First step is performed by the `check_kernel_module_loaded` function that verifies the module is loaded.
The second steps is performed by the `check_kernel_module_parameters` function that verifies that the parameters used to load the module are the required.

At the end the function returns a list of text lines consisting of the correct loaded modules and with the incorrect modules / parameters observed.

cmaf (approved):
thanks @dborquez

GabyCT (approved):

cmaf:
@dborquez I think we may unfortunately need a `#[cfg(any(target_arch = "x86_64"))]` annotation for the generic check until the calls for the other architectures are implemented

dborquez:
> [cfg(any(target_arch = "x86_64"))]

Thank you @cmaf , I have added the annotation

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +206,101 @@ pub fn check_official_releases() -> Result<()> {
     Ok(())
 }
 
+#[cfg(any(target_arch = "x86_64"))]
+pub fn check_kernel_module_loaded(module: &KernelModule) -> Result<()> {
+    const MODPROBE_PARAMETERS: &str = "-n --first-time";
```
c3d, line 211: That seems wrong, you are passing two arguments as a single `.arg()`.

```
> modprobe -n --first-time fuse
modprobe: ERROR: could not insert 'fuse': Module already in kernel
> modprobe '-n --first-time' fuse
modprobe: invalid option -- ' '
```

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +206,101 @@ pub fn check_official_releases() -> Result<()> {
...
+    let output_modinfo = Command::new(MODINFO_PATH)
+        .arg(module.name.clone())
+        .stdout(Stdio::piped())
+        .output()?;
+
+    let status_modinfo = output_modinfo.status;
+
+    // Partial check w/ modprobe
+    let output_modprobe = Command::new(MODPROBE_PATH)
+        .arg(MODPROBE_PARAMETERS)
```
c3d, line 223: This is the location where you pass two arguments with a single `.arg()`.

c3d (changes requested):
I think there is a problem with the argument passing to `modprobe`.

Otherwise, looks good.

jodh-intel (commented):
Thanks @dborquez. I'm wondering if we actually need to use a hash for the kernel params. We will never have many entries to check and I think a statically defined list would be simpler. The problem is how to handle that awkard `unrestricted_guest`. We can either have some special case logic or if we want to keep the design as clean as possible, you could define a `KernelModule` as a name and an optional list of parameters where each parameter is a name and a corresponding enum. The simple case would use an enum containing a `String`. In that case we'd make sure the kernel parameter was specified with the value specified. But we could add another enum type that contains a predicate function that returns a `bool` where `true` would mean "kernel module parameter check passed." We could then create a predicate for `unrestricted_guest` to check the value *and* whether we're already running in a VM.

Some of this could be done in future PRs, but wdyt about replacing the hash with a list?

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -19,12 +21,39 @@ mod arch_specific {
...
+                name: "kvm".to_string(),
+                params: HashMap::from([
+                    ("kvmclock_periodic_sync".to_string(),"Y".to_string()),
+                ]),
+            },
+
+            KernelModule {
+                name: "kvm_intel".to_string(),
+                params: HashMap::from([
+                    ("unrestricted_guest".to_string(),"Y".to_string()),
```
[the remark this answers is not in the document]
dborquez: thank you @jodh-intel, fixed.

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +208,110 @@ pub fn check_official_releases() -> Result<()> {
...
+        }
+        None => Err(anyhow!("Process terminated by signal")),
+    }
+}
+
+pub fn check_kernel_module_parameters(
+    module: &KernelModule,
+) -> std::result::Result<std::result::Result<std::string::String, anyhow::Error>, ()> {
+    const MODULES_PATH: &str = "/sys/module";
+    const CAT_PATH: &str = "/usr/bin/cat";
```
jodh-intel, line 260: We need to minimise use of any external commands. `modprobe` can't really be avoided but can you remove the use of `cat` and use https://doc.rust-lang.org/std/fs/fn.read_to_string.html instead? That would simplify the logic too I think.

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +208,110 @@ pub fn check_official_releases() -> Result<()> {
...
+                    .stdout(Stdio::piped())
+                    .output()?;
+
+                let status = output.status;
+
+                if !status.success() {
+                    match status.code() {
+                        Some(_code) => {
+                            let msg = format!(
+                                "The \"{}\" parameter seems to be missing in the \"{}\" kernel module.",
```
jodh-intel, line 283: Nit: If you use `{:?}` rather than `\"{}\"`, the value will be quoted automatically.

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +208,110 @@ pub fn check_official_releases() -> Result<()> {
...
+                                parameter, module.name
+                            );
+                            Err(anyhow!(msg))
+                        },
+                        None => {
+                            Err(anyhow!("Process terminated by signal"))
+                        }
+                    }
+                }
+                else {
```
jodh-intel, line 293: Nit: Weird formatting.

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +208,110 @@ pub fn check_official_releases() -> Result<()> {
...
+            let msg = format!("{} {}", ERR_MODPROBE, module.name);
+            Err(anyhow!(msg))
+        }
+        None => Err(anyhow!("Process terminated by signal")),
+    }
+}
+
+pub fn check_kernel_module_parameters(
+    module: &KernelModule,
+) -> std::result::Result<std::result::Result<std::string::String, anyhow::Error>, ()> {
```
jodh-intel, line 258: That return type is a bit scary tbh! You could simplify a little (`std::string::String` -> `String`, and maybe `use anyhow::Error`), but it also looks really odd. fwics this is due to `beau_collector`. An alternative would be to simply return a list of results, but wrap it in a `Result` so the caller can check if the function itself worked:

```
```suggestion
pub fn check_kernel_module_parameters(module: &KernelModule) -> Result<Vec<Result<()>> {
```

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +208,110 @@ pub fn check_official_releases() -> Result<()> {
...
+                    match status.code() {
+                        Some(_code) => {
+                            let msg = format!(
+                                "The \"{}\" parameter seems to be missing in the \"{}\" kernel module.",
+                                parameter, module.name
+                            );
+                            Err(anyhow!(msg))
+                        },
+                        None => {
+                            Err(anyhow!("Process terminated by signal"))
```
jodh-intel, line 289: Nit: It might be worth adding the signal name/number to the error here.

src/tools/kata-ctl/src/check.rs
```diff
@@ -25,6 +34,8 @@ const USER_AGT: &str = "kata";
 
 #[allow(dead_code)]
 const ERR_NO_CPUINFO: &str = "cpu_info string is empty";
+const ERR_LSMOD: &str = "check_kernel_module_loaded return:";
+const ERR_MODPROBE: &str = "check_kernel_module_loaded return: module not loaded:";
```
jodh-intel, line 38: Can you remove the function names from these errors [*]. I'm not totally sure we need these consts as they don't specify the actual error ("module not found", "module not loaded"?)

---

[*] - it's frustrating that rust _still_ doesn't provide an equivalent to C's `__func__`, although it does provide `file!()` and `line!()`.

jodh-intel (commented):
Thanks @dborquez - a few comments...

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -19,12 +20,41 @@ mod arch_specific {
...
+            fp: check_cpu,
+            perm: PermissionType::NonPrivileged,
+        },
+        CheckItem {
+            name: CheckType::CheckKernelModules,
+            descr: "This parameter performs the kvm check",
+            fp: check_kernel_modules,
+            perm: PermissionType::NonPrivileged,
+        },
+    ];
```
cmaf, line 36: @dborquez, FYI, when I added a 3rd CheckType I got a compiler warning about having the same prefix, so my PR will change these to Cpu and KernelModules from CheckCpu and CheckKernelModules, but you could change it now if you want

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +203,64 @@ pub fn check_official_releases() -> Result<()> {
     Ok(())
 }
 
+#[cfg(any(target_arch = "x86_64"))]
+pub fn check_kernel_module_loaded(module: &str, parameter: &str) -> Result<String, String> {
+    const MODPROBE_PARAMETERS_DRY_RUN: &str = "--dry-run";
+    const MODPROBE_PARAMETERS_FIRST_TIME: &str = "--first-time";
+    const MODULES_PATH: &str = "/sys/module";
+    const CAT_PATH: &str = "/usr/bin/cat";
```
jodh-intel, line 211: A reminder that this can now be removed.

src/tools/kata-ctl/src/types.rs
```diff
@@ -33,3 +34,28 @@ pub struct CheckItem<'a> {
...
+    pub name: &'a str,
+    pub value: KernelParamType<'a>,
+}
+
+// KernelModule is used to describe a kernel module along with its required parameters.
+// Parameters is a map,value structure
+// #[derive(Clone, Debug)]
+pub struct KernelModule<'a> {
+    pub name: &'a str,
+    pub parameter: KernelParam<'a>,
```
jodh-intel, line 60: This is fine for now. There is a chance in the future that we may need to check >1 parameters for a single module. If and when that happens, we should probably make this a `Vec<KernelParam<'a>>`.

src/tools/kata-ctl/src/types.rs
```diff
@@ -33,3 +34,28 @@ pub struct CheckItem<'a> {
     pub fp: BuiltinCmdFp,
     pub perm: PermissionType,
 }
+
+// Builtin module parameter check handler type
+pub type BuiltinModuleParamFp = fn(args: &str) -> bool;
```
jodh-intel, line 39: I think this would be better as:

```suggestion
pub type BuiltinModuleParamFp = fn(args: &str) -> Result<()>;
```

Since a `Result` is a boolean, we can check for `Ok())` to mean `true` and Err(e)` to mean `false`. And if the predicate function fails, we can display that error to the user.

jodh-intel (commented):
Thanks @dborquez - A few comments that based on our meeting yesterday.

src/tools/kata-ctl/src/types.rs
dborquez: @jodh-intel  thank you for the suggestion

src/tools/kata-ctl/src/arch/x86_64/mod.rs
dborquez: @cmaf do you want to change `CheckKernelModules `to `KernelModules `?

src/tools/kata-ctl/src/check.rs
dborquez: @jodh-intel fixed, thanks!

src/tools/kata-ctl/src/check.rs
dborquez: @jodh-intel I have added the `stdout` and `stderr` to the error returned.

src/tools/kata-ctl/src/check.rs
dborquez: @jodh-intel Thank you, fixed

src/tools/kata-ctl/src/check.rs
dborquez: @jodh-intel fixed, thank you

src/tools/kata-ctl/src/check.rs
dborquez: @jodh-intel fixed, thanks!

src/tools/kata-ctl/src/check.rs
dborquez: @jodh-intel fixed, thank you!

src/tools/kata-ctl/src/arch/x86_64/mod.rs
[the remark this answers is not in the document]
dborquez: @jodh-intel done, thanks!

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +212,81 @@ pub fn check_official_releases() -> Result<()> {
     Ok(())
 }
 
+#[cfg(any(target_arch = "x86_64"))]
+pub fn check_kernel_module_loaded(module: &str, parameter: &str) -> Result<String, String> {
```
amshinde, line 216: Can you add a comment here listing out the steps you are using for checking the module?

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +212,81 @@ pub fn check_official_releases() -> Result<()> {
...
+    // Partial check w/ modinfo
+    let output_modinfo = Command::new(MODINFO_PATH)
+        .arg(module)
+        .output()
+        .expect("failed to execute process");
+
+    let status_modinfo = output_modinfo.status;
+
+    // Partial check w/ modprobe
+    let output_modprobe = Command::new(MODPROBE_PATH)
```
amshinde, line 230: why are you doing a modprobe here? Looks like the function not only checks but loads the module as well. Can you change the function name to reflect this ? Something like check_and _load_kernel_module.
cc @jodh-intel

amshinde: Also, if the modinfo succeeds, do you need to perform a modprobe?

src/tools/kata-ctl/src/check.rs
```diff
@@ -197,6 +212,81 @@ pub fn check_official_releases() -> Result<()> {
...
+        .expect("failed to execute process");
+
+    let status_modprobe = output_modprobe.status;
+
+    // Return module parameter value on success
+    if status_modinfo.success() && !status_modprobe.success() {
+        // Get the parameter value
+        let module_path = format!("{}/{}/parameters/{}", MODULES_PATH, module, parameter);
+
+        match get_single_cpu_info(&module_path, parameter) {
```
amshinde, line 245: I am confused, why are you using `get_single_cpu_info` here ?

src/tools/kata-ctl/src/check.rs
dborquez: @amshinde:

This function receives two arguments:

- a module name
- a module parameter name

The function determines two things:
- if the module name passed exists
- if the module is currently loaded

In case the module exist and is already loaded, the function returns the value of the parameter passed as argument.

1) why are you doing a modprobe here? 

I've passed two arguments to modprobe:
- --dry-run
- --first-time

`--first-time ` causes modprobe to fail if the module passed is already loaded, and the `--dry-run` flag causes `modprobe` to avoid loading it or removing [1], this is because the function should not load or unload any module passed.

2) modinfo is used to determine if a module exist.

The combination of `modinfo` and `modprobe` are as follows:

`modinfo $module >/dev/null 2>/dev/null && ! modprobe --dry-run --first-time $module`

[1]: https://man7.org/linux/man-pages/man8/modprobe.8.html

src/tools/kata-ctl/src/check.rs
dborquez: Basically `get_single_cpu_info` reads a file passed as an argument and looks for a string within that file.
I used this function to read the parameter value of the module.

src/tools/kata-ctl/src/check.rs
amshinde: Ok, thanks for the explanation. Would be helpful to summarise a bit in the comments.

src/tools/kata-ctl/src/check.rs
amshinde: I suggest you use a separate utility function to do so, you could refactor a bit to make `get_single_cpu_info` to make use of it if you want.

jodh-intel: I agree with @amshinde.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -61,4 +87,95 @@ mod arch_specific {
...
+        let missing_hypervisor_flag = check::check_cpu_attribs(&cpu_info, VMM_FLAGS)?;
+
+        if !missing_hypervisor_flag.is_empty() {
+            return Ok(true);
+        }
+
+        Ok(false)
+    }
+    fn unrestricted_guest_param_check(current_value: &str) -> Result<()> {
+        let running_on_vmm = running_on_vmm().unwrap();
```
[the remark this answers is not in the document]
dborquez: @amshinde yes, the error propagation is fixed

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -61,4 +87,95 @@ mod arch_specific {
...
+        const EXPECTED_PARAM_VALUE: char = 'Y';
```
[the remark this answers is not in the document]
dborquez: @amshinde thanks, the expected parameter value is now declared as non mutable variable.

src/tools/kata-ctl/src/check.rs
dborquez: comments added, thanks @amshinde

src/tools/kata-ctl/src/check.rs
dborquez: @amshinde, it was fixed by renaming the function `get_cpu_info()` to `read_file_contents()`

src/tools/kata-ctl/src/check.rs
dborquez: @c3d thank you for your inputs, the arguments are now separated in two `.arg()` argument values.

src/tools/kata-ctl/src/arch/aarch64/mod.rs
```diff
@@ -19,7 +19,7 @@ mod arch_specific {
 
     // List of check functions
     static CHECK_LIST: &[CheckItem] = &[CheckItem {
-        name: CheckType::CheckCpu,
+        name: CheckType::Cpu,
         descr: "This parameter performs the host check",
```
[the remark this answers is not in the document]
dborquez: The original nomenclature was `CheckCpu`, but the `Check` prefix was dropped because rust complained with the error:
`the field is never read`

dborquez:
@c3d could you give the PR a second review please?

src/tools/kata-ctl/src/arch/x86_64/mod.rs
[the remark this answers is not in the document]
c3d: @dborquez I still see an `unwrap()` here, so you will still panic.

Error propagation would be:

```
let running_on_vmm = running_on_vmm()?;
```

c3d (changes requested):
Still a small problem with error propagation…

src/tools/kata-ctl/src/check.rs
```diff
@@ -222,6 +232,72 @@ pub fn check_official_releases() -> Result<()> {
...
+            // Module not found
+            if code != 0 {
+                let stderr = String::from_utf8_lossy(&output_modinfo.stderr);
+                let msg = format!("{}", stderr);
+                return Err(anyhow!(msg).to_string());
+            }
+
+            // Module not loaded
+            let msg = format!("Module not found: {}", module);
+            Err(anyhow!(msg).to_string())
```
amshinde, line 288: format! returns a string, do you need the to_string() call here and below?

amshinde (approved):
lgtm overall. Just a minor comment.

src/tools/kata-ctl/src/check.rs
dborquez: yeah you are right, I removed the `anyhow()` call, thanks @amshinde

jodh-intel (approved):
Thanks @dborquez.

lgtm

jodh-intel:
@c3d - Could you take another look at the latest revision of this?

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -141,6 +168,100 @@ mod arch_specific {
...
+                    } else {
+                        return Err(anyhow!(
+                            "{} does not meet the expected value {}",
+                            current_value,
+                            expected_param_value
+                        ));
+                    }
+                }
+                Err(anyhow!(
+                    "{}, but looks like the working environment is running on a VM",
```
c3d, line 208: Same comment as above (ideally, a message should be formulated to give actionable information to the user), but also I wonder if you should not pass `current_value` here, since this is the part that varies from run to run.

c3d (commented):
I'm really confused by this new iteration.

I was only suggesting to replace `running_on_vmm().unwrap()`, which would panic, with `running_on_vmm()?` which would propagate the error.

Instead, we now have rather large `match running_on_vmm() { Ok(ok) {…} Err(err) {…} } which introduce new error messages along some paths, and I do not understand the error messages we would get at all (they give no indication about what failed, and seem to report 'Y' or 'N' in thir output).

Maybe I am misunderstanding the intent?

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -141,6 +168,100 @@ mod arch_specific {
...
+
+        match running_on_vmm() {
+            Ok(running_on_vmm) => {
+                if !running_on_vmm {
+                    // check the value of the parameter and check if we are running inside a VMM
+                    if current_value == expected_param_value.to_string() {
+                        return Ok(());
+                    } else {
+                        return Err(anyhow!(
+                            "{} does not meet the expected value {}",
```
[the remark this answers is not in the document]
jodh-intel: @c3d - I agree this does look a bit confusing. However, when this error propagates back, the user will end up seeing something like this (see `check_kernel_modules()`):

```
ERROR: "Kernel Module: kvm Parameter: kvmclock_periodic_sync = Y does not meet the expected value: N"
```

That contains all the detail, but taking another look I am wondering if a better approach would be to not "chain" together errors like this and instead change the definition of `BuiltinModuleParamFp`:

```diff
diff --git a/src/tools/kata-ctl/src/types.rs b/src/tools/kata-ctl/src/types.rs
index f71daece0..26f5954d8 100644
--- a/src/tools/kata-ctl/src/types.rs
+++ b/src/tools/kata-ctl/src/types.rs
@@ -35,8 +35,18 @@ pub struct CheckItem<'a> {
     pub perm: PermissionType,
 }

-// Builtin module parameter check handler type
-pub type BuiltinModuleParamFp = fn(args: &str) -> Result<()>;
+// Builtin module parameter check handler type.
+//
+// BuiltinModuleParamFp represents a predicate function to determine if a
+// kernel parameter _value_ is as expected. If not, the returned Error will
+// explain what is wrong.
+//
+// Parameters:
+//
+// - module: name of kernel module.
+// - param: name of parameter for the kernel module.
+// - value: value of the kernel parameter.
+pub type BuiltinModuleParamFp = fn(module: &str, param: &str, value: &str) -> Result<()>;

 // KernelParamType encodes the value and a handler
 // function for kernel module parameters
```

This way, the BuiltinModuleParamFp can return the full error message (containing the kernel module and kernel param name. That would also simplify  `check_kernel_modules()` as it could just call:

```diff
diff --git a/src/tools/kata-ctl/src/arch/x86_64/mod.rs b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
index c2501cbef..e8a059c75 100644
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -243,15 +243,7 @@ mod arch_specific {

                     match parameter_check {
                         Ok(_v) => println!("{} Ok", module.name),
-
-                        Err(parameter_status) => {
-                            return Err(anyhow!(
-                                "Kernel Module: {} Parameter: {:} = {:}",
-                                module.name,
-                                module.parameter.name.to_string(),
-                                parameter_status,
-                            ));
-                        }
+                        Err(e) => return Err(e),
                     }
                 }
```

I'd also tweak the format slightly. Maybe something like:

```
ERROR: Kernel module 'kvm': parameter 'kvmclock_periodic_sync' should have value 'Y', but found 'N'.
```

To be super helpful, we could explain how the user/admin might be able to resolve the problem. Something like:

```
ERROR: Kernel module 'kvm': parameter 'kvmclock_periodic_sync' should have value 'Y', but found 'N'.
 (try running "sudo modprobe -r kvm ; sudo modprobe kvm kvmclock_periodic_sync=Y")
```

However, those commands may not work due to module deps so maybe we should just say, something like:

```
ERROR: Kernel module 'kvm': parameter 'kvmclock_periodic_sync' should have value 'Y', but found 'N': remove the 'kvm' module using `rmmod` and then reload using `modprobe`, setting 'kvmclock_periodic_sync=Y' ?
```

> **Note:**
>
> Although GitHub formats them across multiple lines, I think all these errors should be on a single-line.

src/tools/kata-ctl/src/check.rs
```diff
@@ -222,6 +232,71 @@ pub fn check_official_releases() -> Result<()> {
...
+    const MODPROBE_PARAMETERS_DRY_RUN: &str = "--dry-run";
+    const MODPROBE_PARAMETERS_FIRST_TIME: &str = "--first-time";
+    const MODULES_PATH: &str = "/sys/module";
+
+    // Partial check w/ modinfo
+    // verifies that the module exists
+    let output_modinfo = Command::new(MODINFO_PATH)
+        .arg(module)
+        .output()
+        .expect("failed to execute process");
```
amshinde, line 246: @dborquez Was just looking at this code again, and I realized that this will again panic in case modinfo fails. Its better to return an error here with something like `map_err` and avoiding the `expect` call. 
cc @jodh-intel 

Similarly for the expect usage below.

dborquez: ok @amshinde let me check it out

src/tools/kata-ctl/src/arch/x86_64/mod.rs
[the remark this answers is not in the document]
dborquez: @c3d I appreciate your suggestion to improve the error messages and propagate the error from just one source.
@jodh-intel thank you for this better approach to describe the error.

src/tools/kata-ctl/src/check.rs
```diff
@@ -222,6 +232,78 @@ pub fn check_official_releases() -> Result<()> {
...
+    const MODPROBE_PARAMETERS_FIRST_TIME: &str = "--first-time";
+    const MODULES_PATH: &str = "/sys/module";
+
+    // Partial check w/ modinfo
+    // verifies that the module exists
+    let output_modinfo = Command::new(MODINFO_PATH)
+        .arg(module)
+        .stdout(Stdio::piped())
+        .output()
+        .expect("failed to execute modinfo command");
```
jodh-intel, line 247: @dborquez - yes, as @amshinde says, we should be returning the error, not panicking:

```suggestion
        .output()?;
```

src/tools/kata-ctl/src/check.rs
```diff
@@ -222,6 +232,78 @@ pub fn check_official_releases() -> Result<()> {
...
+
+    // Partial check w/ modprobe
+    // check that the module is already loaded
+    let output_modprobe = Command::new(MODPROBE_PATH)
+        .arg(MODPROBE_PARAMETERS_DRY_RUN)
+        .arg(MODPROBE_PARAMETERS_FIRST_TIME)
+        .arg(module)
+        .stdout(Stdio::piped())
+        .output()
+        .expect("failed to execute modprobe command");
```
jodh-intel, line 259: Same here:

```suggestion
        .output()?;
```

src/tools/kata-ctl/src/arch/x86_64/mod.rs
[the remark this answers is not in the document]
c3d: @dborquez Do you mind doing another iteration with @jodh-intel's suggestions?

src/tools/kata-ctl/src/check.rs
c3d: What about adding a `.context("executing modinfo command")` ?

src/tools/kata-ctl/src/arch/x86_64/mod.rs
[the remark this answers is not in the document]
dborquez: @c3d I've opted to use `match` and format the error message, but then I've changed of mind to adopt your suggestion, thanks.

dborquez:
> Still a small problem with error propagation…

@c3d could you please perform a re-review?

jodh-intel:
Hi @c3d - I'm happy with the latest updates, so could you tal and ack if you're happy? ;)

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -141,6 +168,120 @@ mod arch_specific {
...
+                // check if the 'hypervisor' flag exist in the cpu features
+                let missing_hypervisor_flag = check::check_cpu_attribs(&cpu_info, VMM_FLAGS)?;
+
+                if missing_hypervisor_flag.is_empty() {
+                    return Ok(true);
+                }
+            }
+            Err(e) => {
+                return Err(anyhow!(
+                    "Unable to determine if the OS is running on a VM because this error: {}: {}",
```
c3d, line 184:
```suggestion
                  "Unable to determine if the OS is running on a VM: {}: {}",
```

("because this error" is not grammatically correct, I believe).

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -141,6 +168,120 @@ mod arch_specific {
...
+        module: &str,
+        param_name: &str,
+        param_value_host: &str,
+    ) -> Result<()> {
+        let expected_param_value: char = 'Y';
+
+        let running_on_vmm_alt = running_on_vmm()?;
+
+        if running_on_vmm_alt {
+            let msg = format!("You are running in a VM, where the kernel module '{}' parameter '{:}' have a value '{:}', this causes conflict when running kata.",
```
c3d, line 206: Much improved error message, but needs minor grammatical adjustments.

```suggestion
          let msg = format!("You are running in a VM, where the kernel module '{}' parameter '{:}' has a value '{:}'. This causes conflict when running kata.",
```

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -141,6 +168,120 @@ mod arch_specific {
...
+                param_value_host
+            );
+            return Err(anyhow!(msg));
+        }
+
+        if param_value_host == expected_param_value.to_string() {
+            Ok(())
+        } else {
+            let error_msg = format!(
+                "Kernel Module: '{:}' parameter '{:}' should have value '{:}', but found '{:}.'.",
```
c3d, line 218: Looks good.

Issue #6299, opened 2023-02-16
runtime-rs: handle system directories bind volume


YushuoEdge:
Now, we will mount those system directories on the host if the mount type is specified as "bind", and then, they will be mounted in the guest through virtiofs. It does not make sense, and sometimes it will cause issues for the filesystem type under those directories is KataShared instead of sysfs/tmpfs/cgroup...

This issue may be also related to footloose CI.

PR #6300, opened 2023-02-16
runtime-rs: handle sys_dir bind volume

YushuoEdge:
For some cases, users will mount system directories as bind volume. We should not bind mount these kind of directories in the host as it does not make sense.

Fixes: #6299

### DIFF
--- a/src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
+++ b/src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
@@ -17,6 +17,8 @@ use super::Volume;
 use crate::share_fs::{MountedInfo, ShareFs, ShareFsVolumeConfig};
 use kata_types::mount;
 
+const SYS_MOUNT_PREFIX: [&str; 2] = ["/proc", "/sys"];
+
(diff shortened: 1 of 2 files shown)


Tim-0731-Hzt (approved):

src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
```diff
@@ -252,6 +255,21 @@ fn is_host_device(dest: &str) -> bool {
...
 }
 
+// Skip mounting certain system paths("/sys/*", "/proc/*")
+// from source on the host side into the container as it does not
+// make sense to do so.
+// Agent will support this kind of bind mount.
+fn is_system_mount(src: &str) -> bool {
+    for p in SYS_MOUNT_PREFIX {
+        let mut sub_dir_p = p.to_string();
+        sub_dir_p.push('/');
```
Tim-0731-Hzt, line 265: You could use `format` to do this

YushuoEdge: Thanks! Fixed

liubin (approved):

Issue #6259, opened 2023-02-10
kata-manager: Fails on cgroupsv2 only systems (Fedora)


jodh-intel:
## Problem

Installing using [`kata-manager`](https://github.com/kata-containers/kata-containers/blob/main/utils/README.md) (the "automatic" installation method") fails on systems that only provide cgroups v2, such as Fedora:

```bash
$ grep -i ^ID= /etc/os-release                                                                                                                         main [0]
ID=fedora
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh)"
ERROR: Kata Containers does not yet fully support cgroups v2 - see https://github.com/kata-containers/kata-containers/issues/927
```

However, the error is bogus as Kata does now support host-side cgroups v2 (see #4397 / #3073, #927).

## Workaround

The workaround is simple: add the `-f` option (not carefully that the option comes **before** the closing double quote):

```bash
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh) -f"
```

## TODO

`kata-manager` needs to be updated to remove the restriction:

```diff
diff --git a/utils/kata-manager.sh b/utils/kata-manager.sh
index 1cb65145f..166af9e2b 100755
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -238,23 +238,6 @@ Advice:
 EOF
 }

-# Determine if the system only supports cgroups v2.
-#
-# - Writes "true" to stdout if only cgroups v2 are supported.
-# - Writes "false" to stdout if cgroups v1 or v1+v2 are available.
-# - Writes a blank string to stdout if cgroups are not available.
-only_supports_cgroups_v2()
-{
- local v1=$(mount|awk '$5 ~ /^cgroup$/ { print; }' || true)
- local v2=$(mount|awk '$5 ~ /^cgroup2$/ { print; }' || true)
-
- [ -n "$v1" ] && [ -n "$v2" ] && { echo "false"; return 0; } || true
- [ -n "$v1" ] && { echo "false"; return 0; } || true
- [ -n "$v2" ] && { echo "true"; return 0; } || true
-
- return 0
-}
-
 # Return 0 if containerd is already installed, else return 1.
 containerd_installed()
 {
@@ -280,13 +263,6 @@ pre_checks()

        [ "$ret" -eq 0 ] && die "$containerd_project already installed"

[...]

PR #6301, opened 2023-02-16
Remove install scripts dependency on jq

gabevenberg:
Jq is not installed by default on most distrubutions, and with #6259 encouraging the use of the -f option, it was found that kata-manager does not check for its presence, even though it is needed for the script to work. In addition to doing the checks needed for the script to work even when -f is passed, it would be nice to replace the dependence on jq with one on python, as the latter is more widely found on systems.

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -103,7 +103,7 @@ for tag in response:
 	#   so attempt to perform a semver sort manually.
 	# - Pre-releases are excluded via the select() call.
 	local latest=$(curl -sL "$url" |\
-		python3 -c "$python_jq" |\
+		python -c "$python_jq" |\
 		sort -t "." -k1,1n -k2,2n -k3,3n |\
 		tail -1 || true)
 
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -91,13 +91,19 @@ github_get_latest_release()
 {
 	local url="${1:-}"
 
+	#parses the json response from github. Python is on every machine out there, not jq.
+	local python_jq="import json, sys
+response=json.loads(sys.stdin.read())
+for tag in response:
+    if not ('-' in tag['tag_name']):
+        print(tag['tag_name'])"
 	# Notes:
 	#
 	# - The sort(1) call; none of the standard utilities support semver
 	#   so attempt to perform a semver sort manually.
 	# - Pre-releases are excluded via the select() call.
 	local latest=$(curl -sL "$url" |\
-		jq -r '.[].tag_name | select(contains("-") | not)' |\
+		python3 -c "$python_jq" |\
 		sort -t "." -k1,1n -k2,2n -k3,3n |\
 		tail -1 || true)
 
@@ -155,9 +161,18 @@ github_get_release_file_url()
 
 	local download_url
 
+	#parses the json response from github. Python is on every machine out there, but not jq.
+	local python_jq="import json
+import sys
+response=json.loads(sys.stdin.read())
+version=sys.argv[1]
+for tag in response:
+    if tag['tag_name']==version:
+        for file in tag['assets']:
+            print(file['browser_download_url'])"
+
 	download_url=$(curl -sL "$url" |\
-		jq --arg version "$version" \
-		-r '.[] | select(.tag_name == $version) | .assets[].browser_download_url' |\
+		python -c "$python_jq", $version|\
 		grep "/${regex}$")
 
 	[ -z "$download_url" ] && die "Cannot determine download URL for version $version ($url)"
@@ -299,7 +314,6 @@ check_deps()
 
 	elems+=("curl:curl")
 	elems+=("git:git")
-	elems+=("jq:jq")
 	elems+=("tar:tar")
 
 	local pkgs_to_install=()


fidencio:
@gabevenberg, thanks for the contribution, but I'd rather not have it in, for a few reasons.
1) We use `jq` in several parts of the code, not only in the part you're placing
2) We use `jq` also in the tests repo
3) Installing python takes ~25MB, while installing jq takes ~1MB.

```
root@a675538dcf61:/# apt install jq
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libjq1 libonig5
The following NEW packages will be installed:
  jq libjq1 libonig5
0 upgraded, 3 newly installed, 0 to remove and 5 not upgraded.
Need to get 357 kB of archives.
After this operation, 1087 kB of additional disk space will be used.
Do you want to continue? [Y/n] n
Abort.

root@a675538dcf61:/# apt install python3
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libexpat1 libmpdec3 libpython3-stdlib libpython3.10-minimal libpython3.10-stdlib libreadline8 libsqlite3-0 media-types python3-minimal python3.10 python3.10-minimal readline-common
Suggested packages:
  python3-doc python3-tk python3-venv python3.10-venv python3.10-doc binutils binfmt-support readline-doc
The following NEW packages will be installed:
  libexpat1 libmpdec3 libpython3-stdlib libpython3.10-minimal libpython3.10-stdlib libreadline8 libsqlite3-0 media-types python3 python3-minimal python3.10 python3.10-minimal readline-common
0 upgraded, 13 newly installed, 0 to remove and 5 not upgraded.
Need to get 6494 kB of archives.
After this operation, 23.4 MB of additional disk space will be used.
```

utils/kata-manager.sh
```diff
@@ -91,13 +91,19 @@ github_get_latest_release()
...
+    if not ('-' in tag['tag_name']):
+        print(tag['tag_name'])"
 	# Notes:
 	#
 	# - The sort(1) call; none of the standard utilities support semver
 	#   so attempt to perform a semver sort manually.
 	# - Pre-releases are excluded via the select() call.
 	local latest=$(curl -sL "$url" |\
-		jq -r '.[].tag_name | select(contains("-") | not)' |\
+		python3 -c "$python_jq" |\
```
jodh-intel, line 106: I know that most distros (are moving|have moved) to python3, but there might be some that only install python2 by default, which might break this?

utils/kata-manager.sh
```diff
@@ -155,9 +161,18 @@ github_get_release_file_url()
...
+version=sys.argv[1]
+for tag in response:
+    if tag['tag_name']==version:
+        for file in tag['assets']:
+            print(file['browser_download_url'])"
+
 	download_url=$(curl -sL "$url" |\
-		jq --arg version "$version" \
-		-r '.[] | select(.tag_name == $version) | .assets[].browser_download_url' |\
+		python -c "$python_jq", $version|\
```
jodh-intel, line 175: - Consistency: above, we use `python3`, not `python`.
- As above, `/usr/bin/python` might be python2 or python3. On Fedora 37 it's python3.

jodh-intel (commented):
Thanks @gabevenberg, and well done for being the first to raise a PR! ;)

I've added some comments about python. It's an awkward one. The most robust approach would probably be to:

1) Check for `jq`. If found, use it, else try to install it.
2) If we can't use/install jq, revert to python.
3) First look for python3 and if found use that.
4) If no python3, look for python and if found use that.
5) Give up with an error message.

---

Regarding the CI complaining, yes, our commit format is somewhat picky ;)

The things you need to do to get the CI to be happy are:

- Change your [SOB](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#certificate-of-origin) [1] line to use dashes rather than spaces.
- Add a ["subsystem"](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#subsystem) (aka tag, something like "utils:" for this PR) to the first (summary) line.

The final commit should look something like:

```
utils: replaced jq with simple python3 in kata-manager

Most distributions do not have `jq` installed by default. Rather then
have a dependency that is only needed for the installation script and is
otherwise unlikely to be used, we can replace our use of `jq` for a small
embedded python program, something that is default on most systems.

Fixes #6270

Signed-off-by: Gabe Venberg <gabevenberg@gmail.com>
```

Also, we need to remove the merge commit as we don't use those (we use rebasing with a force push to update PR branches):

```bash
# Remove the merge commit
$ git reset HEAD^

# Clean up your working tree
$ git checkout -f
$ git clean -dfx

# Update your local copy of the main branch
$ git checkout main
$ git pull

# Now rebase your branch on top of main
$ git checkout jq-replace
$ git rebase -i main
```

Then,

- Change `pick` to `reword`, then save+quit editor.
- Update the commit to add the `utils:` prefix to line 1.
- Update the `Signed-off-by:` line to add the dashes.
- Save and quit.
- Finally, run: `git push -f` to update your PR branch.

---

[1] - Yeah, there is an amusing alternative definition but we're not using it in that context! Ditto for git of course :smile:

utils/kata-manager.sh
gabevenberg: Good catch on the consistency. To the best of my knowledge, the script will work on both python2 and 3

gabevenberg:
@fidencio no problem, most of this was to get comfortable with your CI checks ;). 

However, I would like to point out that according to a quick `grep -R --files-with-matches --binary-files=without-match "jq" ./`, jq is only used as a build dependency, with the exception of kata-manager.sh. If kata-manager did not require jq, the end user would not require it.

Secondly, yes, python is an order of magnitude larger dependency than jq, but it will almost always already exist. The only distribution I can find that doesnt include python is some of the most minimal, container-oriented alpine images. All the distributions are documentation says we support has python as part of the base install, none of them have jq as base.

nonetheless, its honestly no skin off my back if we want to close this. Or I can just check for the presence of both python and jq, using whatever is available. regardless, Ill be opening a new PR to fix #6302, arguably the root issue behind what I was trying to fix here. (didnt realize that the script was supposed to be checking for jq in the first place.)

jodh-intel:
@gabevenberg - Let's close this one now that you resolved the issue in a different way on #6309. We can always re-open at a later date if we decide to fully remove `jq` as a dependency.

Issue #6303, opened 2023-02-16
rootfs: Ship the rootfs-image we test


fidencio:
This is an interesting one, as we currently ship our rootfs-image based on `clearlinux`, but **all** the tests in our Ci using the rootfs-image actually use `ubuntu` as the base distro.

I'd like to suggest that, right now, we switch the default that we ship to our users to exactly the same of what we test as part of our CI, and later on we can discuss whether `ubuntu` is the right choice or whether we should go to something smaller, such as alpine.

PR #6304, opened 2023-02-16
versions: Use ubuntu as the default distro for the rootfs-image

fidencio:
Currently ubuntu is already the default distro for all the architectures but x86_64, which uses clearlinux.  However, our CI does *not* test the clearlinux image we ship.

Taking a look at our CI code [0], we've been using ubuntu as base for the tests for a few years already, if not forever.

The minimum we can do is to switch to distributing ubuntu, as the tested rootfs-image, and then decide later on whether we should switch back to clearlinux (once we switch our CI to using that, and make sure all tests will be green), or if we move to slimmer distro, such as alpine.

[0]: https://github.com/kata-containers/tests/blob/0a39dd1a01dc135e76f874b9475a33c5f54afcca/.ci/install_kata_image.sh#L44

Fixes: #6303

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -129,7 +129,7 @@ assets:
         name: "ubuntu"
         version: "latest"
       x86_64:
-        name: &default-image-name "clearlinux"
+        name: &default-image-name "ubuntu"
         version: "latest"
     meta:
       image-type: *default-image-name


chavafg (approved):
make sense to ship what we are currently testing.

amshinde (approved):

Issue #6210, opened 2023-02-03
Increment kata_config_version for kernel config option changes 


amshinde:
Dup of https://github.com/kata-containers/tests/issues/5419.

amshinde:
Creating this one for a linked issue for PR.

PR #6305, opened 2023-02-17
actions: Use `git-diff` to get changes in kernel dir

amshinde:
Use `git-diff` instead of legacy `git-whatchanged` to get differences in the packaging/kernel directory. This also fixes a bug by grepping for the kernel directory in the output of the git command.

Fixes: #6210


fidencio (approved):
lgtm, thanks @amshinde!

GabyCT (approved):

Issue #6307, opened 2023-02-17
Unnecessary Send/Sync implement should be removed


Tim-Zhang:
https://github.com/kata-containers/kata-containers/blob/c1602c848a3d908c3da983bf00092204cc4945cc/src/runtime-rs/crates/runtimes/virt_container/src/lib.rs#L38

Send and Sync are automatically derived traits, if a type is composed entirely of Send or Sync types, then it is Send or Sync. Almost all primitives are Send and Sync, so we don't need to implement them manually most of the time

PR #6308, opened 2023-02-17
runtime-rs: remove unnecessary Send/Sync trait implement

Tim-Zhang:
Send and Sync are automatically derived traits,
if a type is composed entirely of Send or Sync types, then it is Send or Sync. Almost all primitives are Send and Sync,
so we don't need to implement them manually most of the time.

Fixes: #6307


quanweiZhou (approved):
LGTM

liubin (approved):

Issue #6270, opened 2023-02-13
kata-manager.sh assumes jq is installed


gabevenberg:
# problem
installing kata via kata-manager.sh will call `jq` without checking whether `jq` is installed. This can lead to a confusing error message implying that fetching from github itself failed.

# fix
add a more user friendly error message when jq is not found.

jodh-intel:
Thanks for raising @gabevenberg. I agree - we can add an extra check and if not found explain how the user can install the package/command.

It might be getting a bit baroque but another approach could be to fall back to using `python` or `python3`, as these have  JSON support by default.

jodh-intel:
Issue added to the outreach tasks ideas page: https://github.com/kata-containers/community/wiki/NDSU---Outreach-2023-work-item-ideas

gabevenberg:
> Thanks for raising @gabevenberg. I agree - we can add an extra check and if not found explain how the user can install the package/command.
> 
> It might be getting a bit baroque but another approach could be to fall back to using `python` or `python3`, as these have JSON support by default.

This is true. From cursory examination, jq is not a default tool on ubuntu, debian, or fedora. Python3, however, is everywhere.

gabevenberg:
> Thanks for raising @gabevenberg. I agree - we can add an extra check and if not found explain how the user can install the package/command.
> 
> It might be getting a bit baroque but another approach could be to fall back to using `python` or `python3`, as these have JSON support by default.

So, after re-implementing the jq stuff in python, I realized we are checking (and on some systems, even attempting to install) jq, but due to passing the -f flag to workaround #6259, the entire check gets bypassed. I can open a PR to remove the dependency on jq still, or we can just wait of #6259 to be fixed. (or do we want things like curl and jq, that are needed for the script to work at all, to be checked regardless of -f?)

jodh-intel:
> but due to passing the -f flag to workaround https://github.com/kata-containers/kata-containers/issues/6259, the entire check gets bypassed.

@gabevenberg - looking at the code again, yes, another solution would be to simply call `check_deps()` before `pre_checks()` as that is really what the script should do:

```diff
diff --git a/utils/kata-manager.sh b/utils/kata-manager.sh
index 1cb65145f..aeda885a9 100755
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -345,8 +345,8 @@ setup()

        [ "$force" = "true" ] && return 0

-       pre_checks
        check_deps
+       pre_checks
[...]

PR #6309, opened 2023-02-17
utils: always check some dependencies.

### LINKED ISSUE #6302: Kata Manager should check for dependencies it needs itself even with -f.
Currently, certain dependency checks like curl and tar are skipped when the -f option is passed to kata-manager.sh. This leads to the script failing in ways that may be unintuitive to the user. We should really be performing those checks even when -f is passed.

gabevenberg:
Every dependency in check_deps is used inside the script (apart from git, which may be a historical artifact), and therefore should be checked even when the -f option is passed to the script. Simply changed at what point check_deps is called in order to always run it.

Fixes #6302.

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -327,7 +327,7 @@ check_deps()
 		debian|ubuntu) sudo apt-get -y install $packages ;;
 		fedora) sudo dnf -y install $packages ;;
 		opensuse*|sles) sudo zypper install -y $packages ;;
-		*) die "Unsupported distro: $ID"
+		*) die "Unsupported distro: $ID, install $packages manually and re-run"
 	esac
 }
 
@@ -343,10 +343,12 @@ setup()
 
 	source /etc/os-release || source /usr/lib/os-release
 
+	#these dependencies are needed inside this script, and should be checked regardless of the -f option.
+	check_deps
+
 	[ "$force" = "true" ] && return 0
 
 	pre_checks
-	check_deps
 }
 
 # Download the requested version of the specified project.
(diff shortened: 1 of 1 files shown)


utils/kata-manager.sh
```diff
@@ -327,7 +327,7 @@ check_deps()
 		debian|ubuntu) sudo apt-get -y install $packages ;;
 		fedora) sudo dnf -y install $packages ;;
 		opensuse*|sles) sudo zypper install -y $packages ;;
-		*) die "Unsupported distro: $ID"
+		*) die "Unsupported distro: $ID, install $packages manually and re-run"
```
fidencio, line 330: Just a tip from the user's perspective, and I know it's been there before your PR. :-)

Drop the "Unsupported distro: " part of the message.  It would make a user run away thinking the whole kata-containers is not supported on that distro, which is not the case.

gabevenberg: Done!

jodh-intel (approved):
Nice! Thanks @gabevenberg!

lgtm

jodh-intel:
Let's fire up the CI (see https://github.com/kata-containers/community/wiki/Controlling-the-CI for a list of commands btw)...

/test

jodh-intel:
> this would also fix https://github.com/kata-containers/kata-containers/issues/6270, btw. @jodh-intel, do I need to make a dummy commit to get gh to close that issue when this is closed, or do we just manually close it?

@gabevenberg - I'll close that manually in a sec. An alternative is to specify multiple "Fixes: #" comment lines so that GH will close all referenced issues when the PR is merge.

PR #6310, opened 2023-02-17
packaging: Cache the container used to build the kata-deploy artefacts

fidencio:
As we'll move the CI to be using the kata-deploy scripts to build the components, let's cache the images used and improve a little bit the building time of those components, while also having a realiable way to reproduce issues when debugging any sort of issues.

This has been in place as part of the CCv0 branch for a while and is a forward-port of that.


BbolroC (approved):
LGTM, thanks! :wink:

arronwy (approved):
Thanks @fidencio ,  LGTM! It helps a lots for the CI,

Issue #6311, opened 2023-02-17
virtiofsd: update valid path for ppc64le


Amulyam24:
On Power, C version of virtiofsd is being installed to
```
Installing qemu-system-ppc64 to /usr/bin
Installing tools/virtiofsd/virtiofsd to /usr/libexec/qemu
```
Switch to the path instead of using a link `/usr/libexec/kata-qemu/virtiofsd -> /usr/libexec/qemu/virtiofsd`

PR #6312, opened 2023-02-17
virtiofsd: update to a valid path on ppc64le

Amulyam24:
Currently the symbolic link for virtiofsd which is used as a valid path is not updated on every CI run. Fix it by using the actual path of installation.

Fixes: #6311


fidencio (approved):
lgtm, thanks @Amulyam24!

GabyCT (approved):

dcantah (approved):

Issue #6324, opened 2023-02-20
Dragonball: update only config in config manager


HerlinCoder:
**Which feature do you think can be improved?**

Dragonball config manager
In config manager, insert_or_update is described to insert or update configuration information for a device. But actually it create a new DeviceConfigInfo to replace the old one , resulting in the device is not being preserved.

**How can it be improved?**

In order to solve this problem, we could simply change the action in this function, from using new device info to only updating config in the old device info.

PR #6325, opened 2023-02-20
dragonball: config_manager: preserve device when update

HerlinCoder:
DeviceConfigInfo contains config and device, so when we want to do update we could simply update config part of the info, and device would not be changed during update.

Fixes: #6324


studychao (approved):
LGTM

wllenyj (approved):

Tim-0731-Hzt (approved):
LGTM

PR #6327, opened 2023-02-20
runtime/arm64: Enable virtio mem 

jongwu:
For now, there is no other method than SHPC to hotplug pci device into pci-bridge. But SHPC is harmfull sometime, thus we leave this choice to user to decide if SHPC should be enabled.

As virtio-mem device is hotplugged into pci-bridge, once it enabled, SHPC should be enabled either.

Also, thee memory size allocated to NVDIMM should be added on top of host memory size as the total memory. If not, we can't pass the qemu check when hotplug virtio-mem.


fidencio:
@jongwu, could you rebase the PR?  It seems the kata-deploy issues are the ones already fixed on main.

Issue #6337, opened 2023-02-21
Rename SEV firmware to AMDSEV.fd


Alex-Carter01:
Renaming the output binary from the AmdSevPkg from `OVMF.fd` to `AMDSEV.fd` so it does not conflict with the base x86_64 build.

Based on discussion on #5850 

Creating a new issue/PR to make it more intuitive than seeing an SEV change as part of a commit labelled for SNP.

@larrydewey @fitzthum

PR #6338, opened 2023-02-21
CC: Rename SEV firmware to AMDSEV.fd

Alex-Carter01:
Renaming the output binary from the AmdSevPkg from OVMF.fd to AMDSEV.fd so it does not conflict with the base x86_64 build.

Changing install name in ovmf static builder and the location in the sev config file.

Fixes: #6337


fitzthum (approved):

Alex-Carter01:
linking the correct new issue. had pointed to the snp x86_64 ovmf issue

larrydewey (approved):

fidencio (approved):
lgtm, thanks @Alex-Carter01!

fidencio:
We can get this one merged as it doesn't affect TDX, and the queue is extremely long Today.

Issue #6320, opened 2023-02-20
In the arm64 environment, there is an error in the shim-v2 part of building the kata static installation package


singhwang:
# Description of problem
In the arm64 environment, execute the command to enter the local-build directory.
`cd $GOPATH/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build`
An error occurred when executing `./kata-deploy-binaries-in-docker.sh --build=shim-v2` or `make shim-v2-tarball`

# Expected result
Shim-v2 can be built normally without errors in the arm64 environment.

# Actual result
An error occurred when executing `./kata-deploy-binaries-in-docker.sh --build=shim-v2` or `make shim-v2-tarball`
```
error occurred: Failed to find tool. Is `aarch64-linux-musl-gcc` installed?
```

# Further information
The root cause of this problem is that the `aarch64-linux-musl-gcc` is missing in the arm64 environment, so `aarch64-linux-musl-gcc` needs to be installed and installed from source code.

PR #6340, opened 2023-02-21
kata-deploy: Ensure go binaries can run on Ubuntu 20.04

fidencio:
---

Revert "kata-deploy: Fix kata static firecracker arm64 package build error"

This reverts commit 697ec8e578f32c238adf7c14262f6bf1437eadd7.

Golang binaries are built statically by default, unless linking against
CGO, which we do.  In this case we dynamically link against glibc,
causing us troubles when running a binary built with Ubuntu 22.04 on
Ubuntu 20.04 (which will still be supported for the next few years ...)

---

kata-deploy: Fix kata static shim-v2 arm64 package build error

When building the kata static arm64 package, the stages of shim-v2 report errors.

Fixes: #6320
Signed-off-by: SinghWang <wangxin_0611@126.com>

---

### DIFF
--- a/tools/packaging/static-build/shim-v2/Dockerfile
+++ b/tools/packaging/static-build/shim-v2/Dockerfile
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM ubuntu:22.04
+FROM ubuntu:20.04
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
--- a/tools/packaging/static-build/shim-v2/install_go_rust.sh
+++ b/tools/packaging/static-build/shim-v2/install_go_rust.sh
@@ -58,6 +58,10 @@ case "${ARCH}" in
 	aarch64)
 		goarch=arm64
 		LIBC=musl
+		# This is a hack needed as part of Ubuntu 20.04
+		if [ ! -f /usr/bin/aarch64-linux-musl-gcc ]; then
+			ln -sf /usr/bin/musl-gcc /usr/bin/aarch64-linux-musl-gcc
+		fi
 		;;
 	ppc64le) 
 		goarch=${ARCH}


jongwu:
IMO, we don't need build musl from source for arm64. We can get musl-gcc by apt install musl-gcc and then create a link file aarch64-linux-musl-gcc based on it. Also I think it's better fix it in install_go_rust.go as we can reuse this change elsewhere.

fidencio:
> IMO, we don't need build musl from source for arm64. We can get musl-gcc by apt install musl-gcc and then create a link file aarch64-linux-musl-gcc based on it. Also I think it's better fix it in install_go_rust.go as we can reuse this change elsewhere.

The problem that I have using this as part `install_go_rust.sh` is that the script is not aware of which distro one is using.

fidencio:
@jongwu, also, this is what I'm getting on ubuntu 20.04:
```
root@269e793d267a:/# apt-get install musl-gcc
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package musl-gcc
```

jongwu:
Oh, my fault, should be "apt install -y musl-tools" then musl-gcc will appear under /usr/bin

singhwang:
@fidencio @jongwu If you decide to use ubuntu 20.04, it is indeed easier to change this way, and it is simpler than building from source code I submitted before.

fidencio:
> @fidencio @jongwu If you decide to use ubuntu 20.04, it is indeed easier to change this way, and it is simpler than building from source code I submitted before.

I'm keeping your Signed-off-by and Ownershop of the patch, thanks to the nice work you done with the previous PR.

Amulyam24 (approved):
LGTM, Thanks @fidencio!

stevenhorsman (approved):
LGTM. Thanks!

singhwang (approved):

singhwang:
> > @fidencio @jongwu If you decide to use ubuntu 20.04, it is indeed easier to change this way, and it is simpler than building from source code I submitted before.
> 
> I'm keeping your Signed-off-by and Ownershop of the patch, thanks to the nice work you done with the previous PR.

Thanks.

Issue #6341, opened 2023-02-21
CCv0: Dynamic changes to ConfigMap/DownwardAPI/Secrets volumes are not propagated for remote hypervisor


bpradipt:
Kubernetes ConfigMap/DownwardAPI/Secrets volumes support dynamic updates of their contents.  The updated contents eventually propagates to a corresponding volume of the pod.

In the case of remote hypervisor, however, the contents of ConfigMap/DownwardAPI/Secrets are copied only once at pod creation time, and never copied again when they are updated.

This is because the current implementation only copies the volume contents once.
https://github.com/kata-containers/kata-containers/blob/b91337afef54d94e7f6458a0d1548e921dab6ea0/src/runtime/virtcontainers/fs_share_linux.go#L270

Ref: https://github.com/confidential-containers/cloud-api-adaptor/issues/574

ariel-adam:
@bpradipt are the 2 merged PRs (above) covering this issue or there is still work to be done?

stevenhorsman:
Fixes by #6342

PR #6342, opened 2023-02-21
runtime: propagate configmap, secrets, downward-api etc changes for remote-hyp

bpradipt:
For remote hypervisor, the configmap, secrets, downward-api or project-volumes are copied from host to guest. This patch watches for changes to the host files and copies the changes to the guest.

Note that configmap updates takes significantly longer than updates via downward-api. This is similar across runc and Kata runtimes.

Fixes: #6341

### DIFF
--- a/src/runtime/virtcontainers/fs_share_linux.go
+++ b/src/runtime/virtcontainers/fs_share_linux.go
@@ -14,9 +14,12 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"sync"
 	"syscall"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -31,15 +34,26 @@ func unmountNoFollow(path string) error {
 }
 
 type FilesystemShare struct {
-	sandbox *Sandbox
+	sandbox            *Sandbox
+	watcher            *fsnotify.Watcher
+	srcDstMap          map[string]string
+	watcherDoneChannel chan bool
 	sync.Mutex
 	prepared bool
 }
 
 func NewFilesystemShare(s *Sandbox) (FilesystemSharer, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("Creating watcher returned error %w", err)
+	}
+
 	return &FilesystemShare{
-		prepared: false,
-		sandbox:  s,
+		prepared:           false,
+		sandbox:            s,
+		watcherDoneChannel: make(chan bool),
+		srcDstMap:          make(map[string]string),
+		watcher:            watcher,
 	}, nil
 }
 
@@ -266,8 +280,46 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
 			}
 
 			dstPath := filepath.Join(guestPath, srcPath[len(srcRoot):])
+			f.Logger().Infof("ShareFile: Copying file from src (%s) to dest (%s)", srcPath, dstPath)
+			//TODO: Improve the agent protocol, to handle the case for existing symlink.
+			// Currently for an existing symlink, this will fail with EEXIST.
+			err = f.sandbox.agent.copyFile(ctx, srcPath, dstPath)
+			if err != nil {
+				f.Logger().WithError(err).Error("Failed to copy file")
+				return err
+			}
 
-			return f.sandbox.agent.copyFile(ctx, srcPath, dstPath)
+			// Add fsNotify watcher for volume mounts
+			if strings.Contains(srcPath, "kubernetes.io~configmap") ||
+				strings.Contains(srcPath, "kubernetes.io~secrets") ||
+				strings.Contains(srcPath, "kubernetes.io~projected") ||
+				strings.Contains(srcPath, "kubernetes.io~downward-api") {
+
+				// fsNotify doesn't add watcher recursively.
+				// So we need to add the watcher for directories under kubernetes.io~configmap, kubernetes.io~secrets,
+				// kubernetes.io~downward-api and kubernetes.io~projected
+				if info.Mode().IsDir() {
+					// The cm dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~configmap/foo/{..data, key1, key2,...}
+					// The secrets dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~secrets/foo/{..data, key1, key2,...}
+					// The projected dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~projected/foo/{..data, key1, key2,...}
+					// The downward-api dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~downward-api/foo/{..data, key1, key2,...}
+					f.Logger().Infof("ShareFile: srcPath(%s) is a directory", srcPath)
+					err := f.watchDir(srcPath)
+					if err != nil {
+						f.Logger().WithError(err).Error("Failed to watch directory")
+						return err
+					}
+				} else {
+					f.Logger().Infof("ShareFile: srcPath(%s) is not a directory", srcPath)
+				}
+				// Add the source and destination to the global map which will be used by the event loop
+				// to copy the modified content to the destination
+				f.Logger().Infof("ShareFile: Adding srcPath(%s) dstPath(%s) to srcDstMap", srcPath, dstPath)
+				f.srcDstMap[srcPath] = dstPath
+
+			}
+
+			return nil
 		}
 
 		if err := filepath.WalkDir(srcRoot, walk); err != nil {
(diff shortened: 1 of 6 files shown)


bpradipt:
cc @yoheiueda

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -507,3 +557,225 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+		f.Logger().Infof("watchDir: Adding fsnotify watcher for dir (%s)", source)
+		err := f.watcher.Add(source)
+		if err != nil {
+			f.Logger().Info("watchDir: Failed to add watcher to list")
+			return err
+		}
+	}
+
+	for _, v := range watchList {
+		if v != source {
```
stevenhorsman, line 590: Hey Pradipta - apologies for a probably dumb question, but it might help my understanding of how this is properly working.
My assumption is that `for _, v := range watchList {` will go through every source added to the watcher and so if you already had [a,b,c] added to the watchlist and called watchDir(c) the logic would be to compare each of the watched files against source e.g.
- a != c, so let's add it which would return an error as it is already watched
Have I missed something important here in the flow? The alternative I see is to loop through the watchList looking for source and then it is isn't found (after the loop) to add it?

stevenhorsman (commented):
From a non-expert view this looks reasonable, though I have a question about the watcher adding to help me understand it. @bpradipt - is there a decent test case I can use when pulling this code into a worker running the CAA (maybe on libvirt?) to check it working in the intended place, or any example you have that you could share? Thanks!

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: We didn't face any issues since there were no duplicate directories so the condition didn't occur. But I understand what you have pointed out.  Thanks for pointing out.

bpradipt:
> From a non-expert view this looks reasonable, though I have a question about the watcher adding to help me understand it. @bpradipt - is there a decent test case I can use when pulling this code into a worker running the CAA (maybe on libvirt?) to check it working in the intended place, or any example you have that you could share? Thanks!

You can use this sample pod yaml and try with any CAA implementation (eg.libvirt)
Once the POD is running, you can edit the Configmap values like changing the value of `key1` to `value11` and checking the value under `/etc/foocm`.
Likewise you can change the annotations or labels values by editing the pod spec and then look for updated values under 
`/etc/fooinfo`
Configmap updates takes roughly 1min. This is based on kubelet syncloop timing. Annotations/Lables typically take 1-2s for the updates to be visible.
I don't know where to add the test case. Please let me know and I'll happy to add an e2e test case.


```
apiVersion: v1
kind: ConfigMap
metadata:
  name: myconfigmap
data:
  key1: "value1"
  key2: "value2"

---

apiVersion: v1
kind: Pod
metadata:
  name: mypod
  annotations:
    key1: "value1"
    key2: "value2"
  labels:
    key1: "value1"
    key2: "value2"
spec:
  runtimeClassName: kata-remote-cc
  containers:
  - name: mypod
    image: quay.io/centos/centos:stream9
    args: [ "sleep" , "40000"]
    volumeMounts:
    - name: foocm
      mountPath: "/etc/foocm"
    - name: fooinfo
      mountPath: "/etc/fooinfo"

  volumes:
  - name: foocm
    configMap:
      name: myconfigmap
  - name: fooinfo
    downwardAPI:
      items:
        - path: "labels"
          fieldRef:
            fieldPath: metadata.labels
        - path: "annotations"
          fieldRef:
            fieldPath: metadata.annotations
```

stevenhorsman:
> You can use this sample pod yaml and try with any CAA implementation (eg.libvirt)

Thanks for the info
> I don't know where to add the test case. Please let me know and I'll happy to add an e2e test case.

I guess from a peer pod point of view it would be great to have this test case in https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/test/e2e as this sounds like a scenario that we would want to regression test for each build to ensure we don't break with a change either in kata, or CAA along the line? We might need a bit of help from @wainersm to make sure we do it in the right way though! :)

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: Updated the code. Thanks @stevenhorsman

stevenhorsman: Thanks! Looks good

bpradipt:
> > You can use this sample pod yaml and try with any CAA implementation (eg.libvirt)
> 
> Thanks for the info
> 
> > I don't know where to add the test case. Please let me know and I'll happy to add an e2e test case.
> 
> I guess from a peer pod point of view it would be great to have this test case in https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/test/e2e as this sounds like a scenario that we would want to regression test for each build to ensure we don't break with a change either in kata, or CAA along the line? We might need a bit of help from @wainersm to make sure we do it in the right way though! :)

Adding the test to the e2e repo of CAA makes sense. We can use the issue in CAA repo to track the e2e test case - https://github.com/confidential-containers/cloud-api-adaptor/issues/574

stevenhorsman:
@bpradipt - any chance you can rebase this now the `main` merge has gone in and then we'll re-run the tests and hopefully move this forward

stevenhorsman:
@bpradipt - in case you hadn't seen it we are hitting some errors in the unit tests for this PR:
```
--- FAIL: TestPreAddDevice (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x146ffbd]

goroutine 759 [running]:
testing.tRunner.func1.2({0x16f9400, 0x23183b0})
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1396 +0x372
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1399 +0x5f0
panic({0x16f9400, 0x23183b0})
	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/panic.go:890 +0x262
github.com/kata-containers/kata-containers/src/runtime/virtcontainers.(*Container).mountSharedDirMounts(0xc0007622c0, {0x1ae3900?, 0xc00003c1b8}, 0x0?, 0x0?)
	/home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime/virtcontainers/container.go:534 +0x23d
github.com/kata-containers/kata-containers/src/runtime/virtcontainers.TestPreAddDevice(0x0?)
	/home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/runtime/virtcontainers/sandbox_test.go:1189 +0xb46
testing.tRunner(0xc000487380, 0x18da890)
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1446 +0x217
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1493 +0x75e
FAIL	github.com/kata-containers/kata-containers/src/runtime/virtcontainers	1.542s
```

stevenhorsman:
The TDX integration test is a known issue, so shouldn't block this from merging.

stevenhorsman (approved):
My comment has been addressed and the tests are passing, so I'm happy enough with this. I'd like to see an integration test in the cloud-api-adaptor, but that comes later. Thanks @bpradipt

stevenhorsman:
@yoheiueda - if you get a chance would you be able to review this PR please?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -507,3 +557,225 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+				f.Logger().Debugf("copyFilesFromDataDir: dataDir(%s), path(%s), rel(%s)", dataDir, path, rel)
+				// Form the destination path in the guest
+				dstFile := filepath.Join(dst, rel)
+				f.Logger().Infof("copyFilesFromDataDir: Copying file %s to dst %s", path, dstFile)
+				err = f.sandbox.agent.copyFile(context.Background(), path, dstFile)
+				if err != nil {
+					f.Logger().Infof("copyFilesFromDataDir: Error in copying file %v", err)
+					return err
+				}
+				f.Logger().Infof("copyFilesFromDataDir: Successfully copied file (%s)", path)
```
yoheiueda, line 762: The old directory (`..2023_02_09_06_40_51.2326009790`) in the target is updated with the content of the new directory (`..2023_02_13_12_35_49.1380323032`), instead of creating a new directory (`..2023_02_13_12_35_49.1380323032`) and atomically updating the symlink (`..data`). Right?

So, there is a possibility that containers see contents being updated.

I think it is OK for now, since the current agent API doesn't have functions like `rename` and `remove`.

It looks like calling `CopyFile`for an existing symlink will fail due to `EEXIST`.

https://docs.rs/nix/latest/nix/unistd/fn.symlinkat.html
https://github.com/kata-containers/kata-containers/blob/4c810c7933a1ef044eca4cc31d1e90e79df6ac61/src/agent/src/rpc.rs#L2069-L2078

How about adding a TODO comment here about the current limitation?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -507,3 +557,225 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+
+}
+
+func (f *FilesystemShare) StartFileEventWatcher(ctx context.Context) error {
+
+	// Start event loop if watchList is not empty
+	if (f.watcher != nil) || len(f.watcher.WatchList()) == 0 {
+		f.Logger().Info("StartFileEventWatcher: No watches found, returning")
+		return nil
+	}
```
yoheiueda, line 612:
```
f.watcher != nil
```
I think this should be
```
f.watcher == nil
```
Otherwise, `StartFileEventWatcher` immediately returns in my setup. 

Alternatively:
```
	if !(f.watcher != nil && len(f.watcher.WatchList() > 0) {
```

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -266,8 +280,44 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
...
+			if err != nil {
+				f.Logger().WithError(err).Error("Failed to copy file")
+				return err
+			}
+
+			// Add fsNotify watcher for volume mounts
+			if strings.Contains(srcPath, "kubernetes.io~configmap") ||
+				strings.Contains(srcPath, "kubernetes.io~secrets") ||
+				strings.Contains(srcPath, "kubernetes.io~projected") ||
+				strings.Contains(srcPath, "kubernetes.io~downward-api") {
```
yoheiueda, line 294: A `kubernetes.io~*` directory name should be already in `srcRoot`, so I don't think we need to check every `srcPath` in `WalkDir`.

We can check `srcRoot` at Line 259, and set a flag, which will be referenced in `WalkDir`.

Possibly we can do more strict checking here to reduce false-positives like this.
https://go.dev/play/p/JNNXZSmzLkS

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -507,3 +557,225 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+	f.Logger().Infof("watchDir: Add fsnotify watcher for dir (%s)", source)
+	watchList := f.watcher.WatchList()
+	if len(watchList) == 0 {
+		f.Logger().Infof("watchDir: Adding fsnotify watcher for dir (%s)", source)
+		err := f.watcher.Add(source)
+		if err != nil {
+			f.Logger().Info("watchDir: Failed to add watcher to list")
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
bpradipt: Right.. I'll fix it in the next iteration.

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: Nice catch. My bad. I'll fix it in the next iteration

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: For now, let's keep the existing mechanism as it's bit easier to understand. Going forward we can optimise this path.

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: I'll add a TODO for now

src/runtime/virtcontainers/fs_share_linux.go
yoheiueda: Sure. Go ahead.

bpradipt:
@yoheiueda ptal

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -31,15 +34,26 @@ func unmountNoFollow(path string) error {
...
+	srcDstMap          map[string]string
+	watcherDoneChannel chan bool
 	sync.Mutex
 	prepared bool
 }
 
 func NewFilesystemShare(s *Sandbox) (FilesystemSharer, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to add watcher")
```
wainersm, line 48: Can we print the original error as well?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -266,8 +280,46 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
 			}
 
 			dstPath := filepath.Join(guestPath, srcPath[len(srcRoot):])
+			f.Logger().Infof("ShareFile: Copying file from src (%s) and dest (%s)", srcPath, dstPath)
```
wainersm, line 283: `src (%s) and dest (%s)`  -> `src (%s) to dest (%s)`

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -507,3 +559,217 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+	for _, v := range watchList {
+		if v == source {
+			f.Logger().Infof("watchDir: Watcher for dir(%s) is already present", source)
+			return nil
+		}
+	}
+
+	err := f.watcher.Add(source)
+	if err != nil {
+		f.Logger().Info("watchDir: Failed to add watcher to list")
```
wainersm, line 592: Return the original error here as well.

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -507,3 +559,217 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+	var re = regexp.MustCompile(`(?m)\s*[0-9]{4}_[0-9]{2}_[0-9]{2}_[0-9]{2}_[0-9]{2}_[0-9]{2}.[0-9]{10}$`)
+
+	f.Logger().Debugf("StartFileEventWatcher: srcDstMap dump %v", f.srcDstMap)
+
+	// This is the event loop to watch for fsNotify events and copy the contents to the guest
+	for {
+		select {
+		case event, ok := <-f.watcher.Events:
+			if !ok {
+				return fmt.Errorf("StartFileEventWatcher: Error in receiving events")
```
wainersm, line 617: Return the original error here too?

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: I'll add the err to the Errorf statement.

bpradipt:
> https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/fs_share_linux_test.go

@wainersm honestly I couldn't figure out what new test to add when this topic came up last time :-( . The core method is  exercised via container_test.go (mountSharedDirMounts). 
Any specific suggestion you have in mind ? Else can we focus on the integration test for this fix ?

yoheiueda (approved):
LGTM

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: I don't think there is any way :-( - https://pkg.go.dev/github.com/fsnotify/fsnotify#section-readme
There is watcher.Errors which is handled explicitly in subsequent part of the code.

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: Done

stevenhorsman:
@UnmeshDeodhar - We are still hitting the kubernetes init issues even after Ryan said you'd fixed it, so can you take another look at the failures here please. Thank you!

Issue #6353, opened 2023-02-22
osbuilder: add cc_kbc module to Attestation Agent build in SEV initrd for SNP support


Alex-Carter01:
We are planning to have SNP share an initrd with SEV. Currently the rootfs for SEV contains an AA build with `online_sev_kbc`. SNP will be using the `cc_kbc`. We can build the AA with both KBC modules in the SEV case and it will be sufficient for SNP. SNP will temporarily need @fitzthum 's fork  with an SNP attester.

PR #6354, opened 2023-02-22
osbuilder: add cc_kbc module to AA build in SEV initrd for SNP support

Alex-Carter01:
SNP will also be using the same initrd as SEV, so adding the cc_kbc to the attestation agent that is built inside the rootfs for SEV.

Switching attestation agent source to tobin's fork with an SNP attester.

Fixes: #6353


versions.yaml
```diff
@@ -190,8 +190,8 @@ externals:
 
   attestation-agent:
     description: "Provide attested key unwrapping for image decryption"
-    url: "https://github.com/confidential-containers/attestation-agent"
-    version: "v0.3.0"
+    url: "https://github.com/fitzthum/attestation-agent"
+    version: "snp-attester"
```
fidencio, line 194: I'm not sure if you're trying to do this only for testing purposes, then it's okay, but we won't switch the whole project to using the local repo and a specific branch of the attestation-agent.

What I reocmmend is getting that branch merged into the attestation-agent repo, and once that happens we proceed with this PR.

fidencio (changes requested):
Thanks @Alex-Carter01, I've left a comment.

PR #6368, opened 2023-02-24
agent: don't set permission of existing directory in copy_file

### LINKED ISSUE #6367: agent: copy_file unconditionally changes permission of existing parent directory
The handler of the `CopyFile` agent API unconditionally changes the permission of the parent directory of a target file.

`copy_file` copies a specified file in the host to the guest. When the parent of the target file path does not exist, `agent` creates the parent directory using ` fs::create_dir_all`. Then, `agent` set the permission of the parent directory based on `dir_mode` parameter of the `CopyFile` agent API.

This logic leads to override the permission of the parent directory, even when the parent directory already exists.  `copy_file` should not change the permission of the parent directory, when the parent directory already exists.

https://github.com/kata-containers/kata-containers/blob/44a780f262883ce5e1f3377a844a5320d9713b51/src/agent/src/rpc.rs#L1880-L1896

`CopyFile` is used in peer pods, and this issue causes incorrect directory permissions of Kubernetes Volumes.
https://github.com/confidential-containers/cloud-api-adaptor/issues/616

yoheiueda:
This patch fixes the issue that do_copy_file changes the directory permission of the parent directory of a target file, even when the parent directory already exists.

Fixes #6367

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -1877,23 +1877,18 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
         ));
     }
 
-    let parent = path.parent();
-
-    let dir = if let Some(parent) = parent {
-        parent.to_path_buf()
-    } else {
-        PathBuf::from("/")
-    };
-
-    fs::create_dir_all(&dir).or_else(|e| {
-        if e.kind() != std::io::ErrorKind::AlreadyExists {
-            return Err(e);
+    if let Some(parent) = path.parent() {
+        if !parent.exists() {
+            let dir = parent.to_path_buf();
+            if let Err(e) = fs::create_dir_all(&dir) {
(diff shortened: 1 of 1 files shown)


liubin (approved):

YushuoEdge (approved):

stevenhorsman (approved):
LGTM, so once the tests are passing I think we can merge this. Thanks @yoheiueda!

stevenhorsman:
It looks like the two failing tests are `dragonball` on ARM (which is failing for other PRs and not required), and qemu-devmapper, which is failing with the:
```
INFO: Install Kubernetes
Install Kubernetes components
Warning: apt-key output should not be parsed (stdout is not a terminal)
gpg: no valid OpenPGP data found.
```
issue, which I think it due to a google cloud issues, so we are probably blocked until that is resolved.

PR #6371, opened 2023-02-24
Merge pull request #6259 utils/kata-manager.sh

pembek01:
Removed the part in the `kata-manager.sh` script that checked if the host machine only supported cgroups v2.

Issue #6259

Alec Pemberton <pembek1901@gmail.com>

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -238,23 +238,6 @@ Advice:
 EOF
 }
 
-# Determine if the system only supports cgroups v2.
-#
-# - Writes "true" to stdout if only cgroups v2 are supported.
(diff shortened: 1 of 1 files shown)


jodh-intel:
Thanks @pembek01. The commit format is almost correct but you'll need to add in the SoB comment and the subsystem.

You can use `git commit --amend` to drop you into your editor. Then, you can change the commit message to something like this:

```
utils: kata-manager: Remove redundant cgroups v2 checks

Removed the part in the `kata-manager.sh` script that checked if the host
machine only supported cgroups v2 (as that restriction was overcome on #4397).

Fixes: #6259.

```

Save and quit the editor, then you'll need to "force push" to update this PR. On the command line, you'd do that as follows:

```bash
$ git push -f
```

PR #6377, opened 2023-02-24
utils: Remove kata-manager.sh cgroups v2 check

pembek01:
Removed the part in the `kata-manager.sh` script that checks if the host system only runs cgroups v2.

Fixes: #6259.

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -238,23 +238,6 @@ Advice:
 EOF
 }
 
-# Determine if the system only supports cgroups v2.
-#
-# - Writes "true" to stdout if only cgroups v2 are supported.
-# - Writes "false" to stdout if cgroups v1 or v1+v2 are available.
-# - Writes a blank string to stdout if cgroups are not available.
-only_supports_cgroups_v2()
-{
-	local v1=$(mount|awk '$5 ~ /^cgroup$/ { print; }' || true)
-	local v2=$(mount|awk '$5 ~ /^cgroup2$/ { print; }' || true)
-
-	[ -n "$v1" ] && [ -n "$v2" ] && { echo "false"; return 0; } || true
-	[ -n "$v1" ] && { echo "false"; return 0; } || true
-	[ -n "$v2" ] && { echo "true"; return 0; } || true
-
-	return 0
-}
-
 # Return 0 if containerd is already installed, else return 1.
 containerd_installed()
 {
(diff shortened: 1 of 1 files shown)


jodh-intel (commented):
Thanks @pembek01.

We're 99% there now but please can you edit the commit and remove the unecessary first line ("Merge pull request #6259 utils/kata-manager") as that will still cause the CI to fail. The very first line needs to be the one with the "subsystem" (aka ""utils: Remove kata-manager.sh cgroups v2 check").

Also, you can "re-use" this PR - you don't need to close this PR and create a new one. The following should work:

```bash
$ git commit --amend
$ git push -f
```

pembek01:
Sounds good, I will change that after class.

jodh-intel (approved):
Thanks @pembek01 !

lgtm

jodh-intel:
@BbolroC, @Amulyam24 - I think the s390x/ppc64 CI's failure was transitory. I've restarted the jobs from jenkins (to avoid restarting multiple CI's) but the jobs don't appear to be running/scheduled?!? Could you tal?

Amulyam24:
Hi @jodh-intel, I can see that that jobs are currently queued. They should start running once the node is free 👍🏼

jodh-intel:
Thanks @Amulyam24 - I see that GH has now changed status, so I think I was being too impatient ;)

BbolroC:
@jodh-intel Hi, the Jenkins system is not that responsive sometimes. I am sure why, but I've confirmed that a request is handled in the end. 😉

jodh-intel:
Thanks @BbolroC ! ;)

jodh-intel:
Could we get one more quick review on this (small) PR so we can land it folks?

Amulyam24 (approved):
LGTM, Thanks @pembek01!

GabyCT (approved):

c3d:
@jodh-intel @GabyCT Do you think this needs backporting? My rationale is that earlier releases historically did need the check. Do you agree.

jodh-intel:
Hi @c3d - good call! I agree we shouldn't backport.

Thanks @pembek01 ! Let's land this...

___

It's worth pointing out that the latest version of this script is used by the ["automatic" installation method](https://github.com/kata-containers/kata-containers/blob/main/utils/README.md). So when this PR lands, if someone runs this and forces the installation of an old version of Kata that does not support cgroups v2 on the host, they might get confused. That said, it's unlikely as the user would have to explicitly force the installation of an old version with the `-f` option (the default is to install the latest version of Kata).

PR #6379, opened 2023-02-27
kata-ctl: add generic kvm check & unit test

### LINKED ISSUE #5338: kata-ctl: check: x86_64: Write KVM check
Implement the KVM check as done in the golang version.

cmaf:
Add kvm check using ioctl macro to create a syscall that checks the kvmapi version and if creation of a vm is successful.

Check in a unit test that kvm test fails when run as non-root and when device specified is not /dev/kvm.

Fixes #5338

### DIFF
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -13,7 +13,7 @@ mod arch_specific {
     use crate::check::{GuestProtection, ProtectionError};
     use crate::types::*;
     use crate::utils;
-    use anyhow::{anyhow, Result};
+    use anyhow::{anyhow, Context, Result};
     use nix::unistd::Uid;
     use std::fs;
     use std::path::Path;
@@ -40,6 +40,12 @@ mod arch_specific {
             descr: "This parameter performs the kvm check",
             fp: check_kernel_modules,
             perm: PermissionType::NonPrivileged,
+        },
+        CheckItem {
+            name: CheckType::KvmIsUsable,
+            descr: "This parameter performs check to see if KVM is usable",
+            fp: check_kvm_is_usable,
+            perm: PermissionType::Privileged,
         },
     ];
 
@@ -112,6 +118,15 @@ mod arch_specific {
 
     pub fn get_cpu_details() -> Result<(String, String)> {
         utils::get_generic_cpu_details(check::PROC_CPUINFO)
+    }
+
+    // check if kvm is usable
+    fn check_kvm_is_usable(_args: &str) -> Result<()> {
+        println!("INFO: check if kvm is usable: x86_64");
+
+        let result = check::check_kvm_is_usable_generic();
+
+        result.context("KVM check failed")
     }
 
     pub const TDX_SYS_FIRMWARE_DIR: &str = "/sys/firmware/tdx_seam/";
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -6,6 +6,10 @@
 // Contains checks that are not architecture-specific
 
 use anyhow::{anyhow, Result};
+use nix::fcntl::{open, OFlag};
+use nix::sys::stat::Mode;
+use nix::unistd::close;
+use nix::{ioctl_write_int_bad, request_code_none};
 use reqwest::header::{CONTENT_TYPE, USER_AGENT};
 use serde::{Deserialize, Serialize};
 use std::fmt;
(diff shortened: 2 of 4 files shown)


GabyCT (approved):

src/tools/kata-ctl/src/check.rs
```diff
@@ -129,6 +133,55 @@ pub fn run_network_checks() -> Result<()> {
     Ok(())
 }
 
+// Set of basic checks for kvm. Architectures should implement more specific checks if needed
+#[cfg(any(target_arch = "x86_64"))]
+pub fn check_kvm_is_usable_generic(kvm_device: &str, machine_type: i32) -> Result<()> {
```
amshinde, line 138: If this is architecture specific to x86, would it make sense to move most of this code into the arch specific module and have this function call a wrapper function in there.
cc @jodh-intel

amshinde (commented):
lgtm @cmaf. Just a few minor comments.

src/tools/kata-ctl/src/check.rs
```diff
@@ -129,6 +133,55 @@ pub fn run_network_checks() -> Result<()> {
...
+    // check for root user
+    if !nix::unistd::Uid::effective().is_root() {
+        return Err(anyhow!("Will not perform kvm checks as non root user"));
+    }
+
+    // we do not want to erroneously create syscalls to any device besides /dev/kvm
+    if kvm_device != "/dev/kvm" {
```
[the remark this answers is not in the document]
cmaf: @amshinde I could remove it, but then we really have no unit tests for it

src/tools/kata-ctl/src/check.rs
cmaf: @amshinde it isn't, it isn't implemented for the architectures yet so I added this to avoid having the CI fail

src/tools/kata-ctl/src/check.rs
```diff
@@ -129,6 +133,55 @@ pub fn run_network_checks() -> Result<()> {
...
+    let flags = OFlag::O_RDWR | OFlag::O_CLOEXEC;
+    let fd = open(kvm_device, flags, mode)?;
+
+    // check kvm api version
+    ioctl_write_int_bad!(
+        kvm_api_version,
+        request_code_none!(KVM_IOCTL_ID, KVM_GET_API_VERSION)
+    );
+    let v = unsafe { kvm_api_version(fd, 0)? };
+    // per kvm api documentation, this number should always be 12
```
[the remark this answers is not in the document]
cmaf: thanks, updated

src/tools/kata-ctl/src/check.rs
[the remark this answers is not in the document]
cmaf: Also on rereading the documentation, I'm going to remove the vm type and enforce default, if an arch needs something more in depth it can be implemented separately

src/tools/kata-ctl/src/check.rs
```diff
@@ -129,6 +133,53 @@ pub fn run_network_checks() -> Result<()> {
...
+    // check kvm api version
+    ioctl_write_int_bad!(
+        kvm_api_version,
+        request_code_none!(KVM_IOCTL_ID, KVM_GET_API_VERSION)
+    );
+    // 0 is not used but required to produce output
+    let v = unsafe { kvm_api_version(fd, 0)? };
+    // per kvm api documentation, this number should always be 12
+    // https://www.kernel.org/doc/html/latest/virt/kvm/api.html#kvm-get-api-version
+    if v != 12 {
```
jodh-intel, line 166: Could you make this magic number a top-level `const` for better visibility (and put the doc link up there with it)?

src/tools/kata-ctl/src/check.rs
```diff
@@ -129,6 +133,53 @@ pub fn run_network_checks() -> Result<()> {
...
+        return Err(anyhow!("KVM API version is not correct"));
+    }
+
+    // check if you can create vm
+    ioctl_write_int_bad!(
+        kvm_create_vm,
+        request_code_none!(KVM_IOCTL_ID, KVM_CREATE_VM)
+    );
+    // 0 is default machine type
+    let _vmfd = unsafe { kvm_create_vm(fd, 0)? };
```
jodh-intel, line 176: Could you see if you can add in the check to determine if another hypervisor has locked the device, as we do in the golang version:

- https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-check.go#L491..L492

jodh-intel (commented):
Thanks @cmaf - a couple of comments.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -61,4 +69,16 @@ mod arch_specific {
...
+    // check if kvm is usable
+    fn check_kvm_is_usable(_args: &str) -> Result<()> {
+        println!("INFO: check if kvm is usable: x86_64");
+
+        let result = check::check_kvm_is_usable_generic();
+        if result.is_err() {
+            eprintln!("FAILED: KVM CHECK FAILED: {:?}", result);
+        }
+
+        Ok(())
```
jodh-intel, line 82: Shouldn't this function should return the `result`? Overall, the command should return 1 if any checks fail.

jodh-intel (commented):
Thanks @cmaf - one more comment...

src/tools/kata-ctl/src/check.rs
```diff
@@ -332,6 +383,16 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_check_kvm_is_usable_generic() {
+        if !nix::unistd::Uid::effective().is_root() {
```
liubin, line 388: We have a marcro [`skip_if_not_root`](https://github.com/kata-containers/kata-containers/blob/629a31ec6e9ee5744f53ed03a0eb9a2fa69f0b84/src/agent/src/network.rs#L90-L94) for running tests for root user.

src/tools/kata-ctl/src/check.rs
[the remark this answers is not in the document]
cmaf: updated

src/tools/kata-ctl/src/check.rs
cmaf: @jodh-intel updated

src/tools/kata-ctl/src/check.rs
cmaf: @jodh-intel updated but still testing

src/tools/kata-ctl/src/check.rs
cmaf: @liubin thanks, updated

cmaf:
@amshinde @jodh-intel @liubin I only partially updated this and am still testing, please hold off on reviews

src/tools/kata-ctl/src/check.rs
cmaf: Resolved

src/tools/kata-ctl/src/arch/x86_64/mod.rs
cmaf: @jodh-intel in this test, failures aren't collected, a single failure cancels the rest of the test. It shouldn't get to this point if there are any failures. Could you re-review here to see if I have the error handling right?

Also this is a simple pass/fail with no result to return. Should I be returning 0 here? We actually have a significant amount of existing code not returning anything.

jodh-intel: Hi @cmaf - we need `kata-ctl check ...` to return 1 on any failure so all the code should be returning an `Err` if any failure is detected. I think I see where the confusion is occurring - you're returning `Ok()` here to mean "the check ran successfully (and successfully detected an error/failure)" but we need it to just fail with the error so I'd do something like:

```diff
diff --git a/src/tools/kata-ctl/src/arch/x86_64/mod.rs b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
index bb162b984..d4a0c9e88 100644
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -13,7 +13,7 @@ mod arch_specific {
     use crate::check::{GuestProtection, ProtectionError};
     use crate::types::*;
     use crate::utils;
-    use anyhow::{anyhow, Result};
+    use anyhow::{anyhow, Context, Result};
     use nix::unistd::Uid;
     use std::fs;
     use std::path::Path;
@@ -125,11 +125,8 @@ mod arch_specific {
         println!("INFO: check if kvm is usable: x86_64");

         let result = check::check_kvm_is_usable_generic();
-        if result.is_err() {
-            eprintln!("FAILED: KVM CHECK FAILED: {:?}", result);
-        }

-        Ok(())
+        result.context("KVM check failed")
     }

     pub const TDX_SYS_FIRMWARE_DIR: &str = "/sys/firmware/tdx_seam/";
```


> Also this is a simple pass/fail with no result to return.

If this function fails, we just need to return the `Err`, but if it passes, yes, we just return `Ok(())`.

> We actually have a significant amount of existing code not returning an actual value.

Let's discuss in the meeting later today. It's ok if the functions are returning `Ok(())` but if they are not failing when they should be, we need to resolve that.

cmaf: Hi @jodh-intel, I just realized my first response was as if this comment was written for the code in check.rs, sorry! I agree with your assessment.

cmaf: related: https://github.com/kata-containers/kata-containers/issues/6783

cmaf: @jodh-intel  Updated

liubin (approved):

jodh-intel (approved):
Thanks @cmaf.

lgtm

jodh-intel (approved):
Thanks @cmaf.

lgtm

Issue #6390, opened 2023-02-28
workflows: Do not install docker as part of the GitHub actions


fidencio:
The latest ubuntu runners already have docker installed and trying to install it manually will cause the following issue:
```
Run curl -fsSL https://test.docker.com/ -o test-docker.sh
# Executing docker install script, commit: 66474034[5](https://github.com/kata-containers/kata-containers/actions/runs/4297456061/jobs/7490451535#step:3:6)47a9[6](https://github.com/kata-containers/kata-containers/actions/runs/4297456061/jobs/7490451535#step:3:7)caa0a25be56051ff8b726a1b28
Warning: the "docker" command appears to already exist on this system.

If you already have Docker installed, this script can cause trouble, which is
why we're displaying this warning and provide the opportunity to cancel the
installation.

If you installed the current Docker package using this script and are using it
again to update Docker, you can safely ignore this message.

You may press Ctrl+C now to abort this script.
+ sleep 20
+ sudo -E sh -c apt-get update -qq >/dev/null
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy Release' is no longer signed.
```

PR #6391, opened 2023-02-28
workflows: Do not install docker

fidencio:
The latest ubuntu runners already have docker installed and trying to install it manually will cause the following issue:
```
Run curl -fsSL https://test.docker.com/ -o test-docker.sh
Warning: the "docker" command appears to already exist on this system.

If you already have Docker installed, this script can cause trouble, which is
why we're displaying this warning and provide the opportunity to cancel the
installation.

If you installed the current Docker package using this script and are using it
again to update Docker, you can safely ignore this message.

You may press Ctrl+C now to abort this script.
+ sleep 20
+ sudo -E sh -c apt-get update -qq >/dev/null
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy Release' is no longer signed.
```

Fixes: #6390

### DIFF
--- a/.github/workflows/kata-deploy-push.yaml
+++ b/.github/workflows/kata-deploy-push.yaml
@@ -29,12 +29,6 @@ jobs:
           - nydus
     steps:
       - uses: actions/checkout@v2
-      - name: Install docker
-        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
-        run: |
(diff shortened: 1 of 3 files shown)


jepio (approved):

amshinde (approved):

stevenhorsman (approved):
LTGM. Thanks

Issue #6343, opened 2023-02-21
workflows: Publish kata-deploy payload after a merge


fidencio:
For the architectures we know that `make kata-tarball` works as
expected, let's start publishing the kata-deploy payload after each
merge.

This will help to:
* Easily test the content of current `main` or `stable-*` branch
* Easily bisect issues
* Start providing some sort of CI/CD content pipeline for those who
  need that

Later on, when we have the Confidential Containers work merged
into main, we'll have this as a base for the Confidentia Containers
CI/CD.

PR #6400, opened 2023-03-02
workflows:  Fixes for the `payload-after-push` action

fidencio:
---

workflows: Fix action name

We have a few actions in the `payload-after-push.*.yaml` that are
referring to Confidential Containers, but they should be referring to
Kata Containers instead.

---

workflows: Fix the path of imported workflows

In `payload-after-push.yaml` we ended up mentioning cc-*.yaml workflows,
which are non existent in the main branch.

Let's adapt the name to the correct ones.

Fixes: #6343

---

### DIFF
--- a/.github/workflows/payload-after-push-amd64.yaml
+++ b/.github/workflows/payload-after-push-amd64.yaml
@@ -75,7 +75,7 @@ jobs:
     needs: create-kata-tarball
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


stevenhorsman (approved):
LGTM. Thanks!

jodh-intel (approved):
Thanks @fidencio,

lgtm

arronwy (approved):
LGTM!

PR #6402, opened 2023-03-02
workflows: Yet more fixes for publishing the kata-deploy payload after every PR merged

fidencio:
---

kata-deploy: Allow passing IMG_NAME and IMG_TAG

Let's break the IMAGE build parameter into IMG_NAME and IMG_TAG, as it
makes it easier to replace the default CentOS image by something else.

Spoiler alert, the default CentOS image is **not** multi-arch, and we do
want to support at least aarch64 and s390x in the near term future.

---

kata-deploy: Use different images for s390x and aarch64

As the image provided as part of registry.centos.org is not a multi-arch
one, at least not for CentOS 7, we need to expand the script used to
build the image to pass images that are known to work for s390x (ClefOS)
and aarch64 (CentOS, but coming from dockerhub).

---

kata-deploy: Fix path to the Dockerfile

As part of bd1ed26c8d0ef8da99bc2ffffb099f93f73a3880, we've pointed to
the Dockerfile that's used in the CC branch, which is wrong.

For what we're doing on main, we should be pointing to the one under the
`kata-deploy` folder, and not the one under the non-existent
`kata-deploy-cc` one.

Fixes: #6343

---

### DIFF
--- a/tools/packaging/kata-deploy/Dockerfile
+++ b/tools/packaging/kata-deploy/Dockerfile
@@ -3,8 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Specify alternative base image, e.g. clefos for s390x
-ARG IMAGE
-FROM ${IMAGE:-registry.centos.org/centos}:7
+ARG BASE_IMAGE_NAME=registry.centos.org/centos
+ARG BASE_IMAGE_TAG=7
+FROM $BASE_IMAGE_NAME:$BASE_IMAGE_TAG
 ARG KATA_ARTIFACTS=./kata-static.tar.xz
 ARG DESTINATION=/opt/kata-artifacts
 
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
@@ -20,8 +20,8 @@ IMAGE_TAG="${REGISTRY}:kata-containers-$(git rev-parse HEAD)-$(uname -m)"
 echo "Building the image"
 if [ "$(uname -m)" = "s390x" ]; then
 	docker build \
-		--build-arg IMG_NAME=clefos \
-		--build-arg IMG_TAG=7 \
+		--build-arg BASE_IMAGE_NAME=clefos \
+		--build-arg BASE_IMAGE_TAG=7 \
 		--tag ${IMAGE_TAG} .
 else
 	docker build --tag ${IMAGE_TAG} .
@@ -36,8 +36,8 @@ if [ -n "${TAG}" ]; then
 	echo "Building the ${ADDITIONAL_TAG} image"
 	if [ "$(uname -m)" = "s390x" ]; then
 		docker build \
-			--build-arg IMG_NAME=clefos \
-			--build-arg IMG_TAG=7 \
+			--build-arg BASE_IMAGE_NAME=clefos \
+			--build-arg BASE_IMAGE_TAG=7 \
 			--tag ${ADDITIONAL_TAG} .
 	else
 		docker build --tag ${ADDITIONAL_TAG} .
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-KATA_DEPLOY_DIR="`dirname ${0}`/../../kata-deploy-cc"
+KATA_DEPLOY_DIR="`dirname ${0}`/../../kata-deploy"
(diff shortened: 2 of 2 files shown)


tools/packaging/kata-deploy/Dockerfile
```diff
@@ -3,8 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Specify alternative base image, e.g. clefos for s390x
-ARG IMAGE
-FROM ${IMAGE:-registry.centos.org/centos}:7
+ARG IMG_NAME=registry.centos.org/centos
```
c3d, line 6: Just curious, why not keep the name `IMAGE` as it was and add `IMAGE_TAG`, since this is apparently what you are doing?

c3d: Also, I'm trying to assess if there is some important corner case of shifting from shell default value to ARG default value… The new approach seems definitely cleaner.

tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
```diff
@@ -18,14 +18,23 @@ pushd ${KATA_DEPLOY_DIR}
...
 echo "Building the image"
-if [ "$(uname -m)" = "s390x" ]; then
-	docker build \
-		--build-arg IMG_NAME=clefos \
-		--build-arg IMG_TAG=7 \
-		--tag ${IMAGE_TAG} .
-else
-	docker build --tag ${IMAGE_TAG} .
-fi
+case $(uname -m) in
```
c3d, line 21: I find it very confusing that, both in the old and new code, we use *both* `IMG_TAG` and `IMAGE_TAG`. Given that it looks like the script is indeed using `IMAGE_TAG`, that reinforce my suggestion to name the dockerfile ARG as `IMAGE_TAG` as well, no?

tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
```diff
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-KATA_DEPLOY_DIR="`dirname ${0}`/../../kata-deploy-cc"
```
c3d, line 8: Curious why the target issue #6343 is already shown as closed? Is it the correct one?

tools/packaging/kata-deploy/Dockerfile
fidencio: I am basing the work on what's already part of the CCv0 branch.  I can change it, no problem.

tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
fidencio: IMAGE_TAG is the tag you want your built image to receive, for instance, `kata-containers-x86_64`.
IMG_TAG, inside the script, is the `7` in case of `centos:7`.

tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
fidencio: Out of lazyness, to be honest, as all the fixes are actually related to that issue.

c3d: OK.

tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
c3d: Oh, I totally misinterpreted this. Hmm, maybe this suggest we should have a "source" and "destination" prefix to some of those? Or are you comfortable enough that the code is maintainable as is?

fidencio: SRc / DEST is also not the most clear or accurate option.
Let me iterate on this one and I'll count on you for a second pass to make sure we can make it cleaner / easier to maintain.

c3d: Thanks. It can easily be deferred if the change as is unblocks something else.

fidencio: I've changed the IMG_NAME and IMG_TAG to BASE_IMAGE_NAME and BASE_IMAGE_TAG, which makes it easier to understand that those are related to the container used as the basis for the kata-deploy daemonset.

IMAGE_TAG was kept as it is, as it was already like that, as this patch set is not directly touching it, and as I do believe the changes mentioned above make it clear enough to avoid any confusion.

Let me know what you think.

jongwu (approved):

jongwu:
Thanks @fidencio

fidencio:
This is being affected by https://github.com/kata-containers/ci/issues/530

c3d (approved):
Looks perfect now, thanks a lot @fidencio

PR #6406, opened 2023-03-03
static checks workflow improvements

jepio:
- remove all TRAVIS_XXX variables (except TRAVIS because it is still used in kata-containers/tests)
- free some disk space to prevent the sudo test job from running out of space (affects mostly CCv0 branch)
- simplify setting GOPATH

Since these apply to main, I'm submitting against main but would be nice to have them flow into CCv0 soon.

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -21,6 +21,7 @@ jobs:
     env:
       RUST_BACKTRACE: "1"
       target_branch: ${{ github.base_ref }}
+      GOPATH: ${{ github.workspace }}
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
@@ -31,8 +32,6 @@ jobs:
       uses: actions/setup-go@v3
       with:
         go-version: 1.19.3
-      env:
-        GOPATH: ${{ runner.workspace }}/kata-containers
     - name: Check kernel config version
       run: |
         cd "${{ github.workspace }}/src/github.com/${{ github.repository }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/static-checks.yaml
```diff
@@ -19,13 +19,17 @@ jobs:
           - "make test"
           - "sudo -E PATH=\"$PATH\" make test"
     env:
+      # Still used in kata-containers/tests:.ci/setup.sh to select minimal
+      # package set for install
```
fidencio, line 23: I **think** this can be removed, but I need to take a look at the jenkins jobs definitions to be 100% sure.
Poke me again about this if I don't get back to you in a day or so.

fidencio (commented):
This looks sensible to me, but we need to take a look at the jobs definition to make sure we can get rid of those.

.github/workflows/static-checks.yaml
fidencio: As far I `grep` could help me, we pretty much don't use the `TRAVIS` variable anywhere.
I'd say we can go ahead and remove it, and also cleanup the tests and ci repos, whenever it's mentioned.

jepio: ok, i've opened https://github.com/kata-containers/tests/pull/5518 & https://github.com/kata-containers/ci/pull/533, and will update this commit to also remove the usage here.

jepio:
@fidencio
Since https://github.com/kata-containers/tests/pull/5518 is merged, this is also ready

fidencio (approved):
lgtm, thanks @jepio!

stevenhorsman (approved):
LGTM. Thanks so much @jepio!

stevenhorsman:
I'm guessing that is was targetted at `CCv0` at some point, which is why we have the extra CC test jobs pending? They can be ignored safely for merging into `main`

jepio:
#6479 shares the exact same commits and is targeted at CCv0

PR #6411, opened 2023-03-04
Add support for ephemeral mounts to occupy entire sandbox's memory

wlan0:
On hotplug of memory as containers are started, remount all ephemeral mounts with size option set to the total sandbox memory

Fixes: https://github.com/kata-containers/kata-containers/issues/6417


fengwang666:
Doesn't this change introduce a behavior difference between kata and runc from an application's perspective? Let's walk through a scenario: 
I have a pod with a spec containing a container requiring 4GB of memory and a memory-backed emptydir which the container expects to write 4 GB of data. With runc runtime class, I deploy this pod spec on a node with 16 GB of memory (note: the empty dir memory is not calculated as part of the container memory).
Now I want to try out kata runtime class. The same use case is unlikely to work. By default the kata guest VM will be allocated of 6 GB of memory (2GB default + 4GB container), which is not enough for the use case requirement (4GB container + 4GB emptydir).

IMO a more correct way to scale up the emptydir size is to increase the default sandbox memory. The emptydir memory needs to be accounted in the case of kata runtime class and counting it as part of the container resource introduces a divergence from runc. 

cc @egernst

c3d:
> Doesn't this change introduce a behavior difference between kata and runc from an application's perspective?[...]
> IMO a more correct way to scale up the emptydir size is to increase the default sandbox memory. The emptydir memory needs to be accounted in the case of kata runtime class and counting it as part of the container resource introduces a divergence from runc.

In the example given in #6417, we only have a container size specified. There is no size indication for the temporary dir other than the size of the container. My guess is that with runc, in that case, we will also give the tmpfs a size that will be limited by the cgroup for the container, so it seems to me that we should expect something a bit close to what is being implemented here.

wlan0:
>In the example given in https://github.com/kata-containers/kata-containers/issues/6417, we only have a container size specified. There is no size indication for the temporary dir other than the size of the container. My guess is that with runc, in that case, we will also give the tmpfs a size that will be limited by the cgroup for the container, so it seems to me that we should expect something a bit close to what is being implemented here. 

That is correct. tmpfs mount size will be limited to the cgroup of the container as noted here: (https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run)

wlan0:
@fengwang666 @c3d - pinging to bump it up to your attention.

src/agent/src/mount.rs
```diff
@@ -240,6 +240,70 @@ async fn ephemeral_storage_handler(
...
+                    for (_, opt) in storage.options.iter().enumerate() {
+                        if opt.starts_with(FS_GID) {
+                            continue;
+                        }
+                        opts.push(opt)
+                    }
+                    let mount_path = Path::new(&storage.mount_point);
+                    let src_path = Path::new(&storage.source);
+
+                    let (flags, options) = parse_mount_flags_and_options(opts);
```
c3d, line 276: Do we need to have a `remount` option here? If not why not?

c3d (approved):
Overall, this looks good to me, although I did not actually test it.

I have a couple of questions regarding the remount.

src/agent/src/mount.rs
wlan0: all mount flags are passed in from the runtime. Including the `remount` flag

wlan0: https://github.com/kata-containers/kata-containers/pull/6411/files#diff-a6e45adf0a559c18c61d84b1ea3225f62c1cf6166c7bf1f5b24e5c034921db09R2152

fengwang666 (approved):

src/agent/src/mount.rs
c3d: Thanks. I wonder how I missed this.

src/agent/src/mount.rs
```diff
@@ -240,6 +240,70 @@ async fn ephemeral_storage_handler(
...
+            "storage-type" => handler_name.to_owned()));
+
+        match handler_name.as_str() {
+            DRIVER_EPHEMERAL_TYPE => {
+                fs::create_dir_all(Path::new(&storage.mount_point))?;
+
+                if storage.options.is_empty() {
+                    continue;
+                } else {
+                    // assume that fsGid has already been set
```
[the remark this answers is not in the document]
c3d: Thanks. So it's mostly an optimization.

c3d (approved):
Looks good to me.

wlan0:
failing because it is not `gofmt`ed with `-s` flag. Fixing and force-pushing

wlan0:
The power pc tests in ci failure seems like a flake

```

The reset process does not clean your kubeconfig files and you must remove them manually.
Please, check the contents of the $HOME/.kube/config file.
[cleanup_env.sh:36] INFO: Teardown the registry server
[cleanup_env.sh:275] INFO: restart docker service
[cleanup_env.sh:90] INFO: remove local registry server: kata-registry
[cleanup_env.sh:40] INFO: Stop containerd service
[cleanup_env.sh:43] INFO: Remove network devices
[cleanup_env.sh:45] INFO: remove device: cni0
[cleanup_env.sh:45] INFO: remove device: flannel.1
bash: /home/jenkins/workspace/kata-containers-2.0-ubuntu-ppc64le-main-PR/go/src/github.com/kata-containers/tests/integration/kubernetes/cleanup_bare_metal_env.sh false: No such file or directory
[cleanup_env.sh:55] INFO: Check no kata processes are left behind after reseting kubernetes
3298438
ERROR: Found unexpected /usr/local/bin/containerd-shim-kata-v2 present

script:
  name: '/home/jenkins/workspace/kata-containers-2.0-ubuntu-ppc64le-main-PR/go/src/github.com/kata-containers/tests/integration/kubernetes/cleanup_env.sh'
  pid: 3298369
  directory: '/home/jenkins/workspace/kata-containers-2.0-ubuntu-ppc64le-main-PR/go/src/github.com/kata-containers/tests/integration/kubernetes'
  details: 'drwxrwxr-x 6 jenkins jenkins ? 4096 Mar 13 11:25 /home/jenkins/workspace/kata-containers-2.0-ubuntu-ppc64le-main-PR/go/src/github.com/kata-containers/tests/integration/kubernetes'
```

Issue #6415, opened 2023-03-06
SEV kernel is built twice


UnmeshDeodhar:
Kernel is built two times in SEV jobs:
1. In tools\packaging\kata-deploy\local-build\Makefile: 16
2. When building initrd: tools\packaging\kata-deploy\local-build\Makefile: 120

PR #6416, opened 2023-03-06
packaging: fix sev kernel build

UnmeshDeodhar:
Remove duplicate SEV kernel builds.
Fixes: #6415

Create a directory where kernel modules tarball will be extracted.
Fixes: #6418

### DIFF
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -13,7 +13,6 @@ ARCH := $(shell uname -m)
 ifeq ($(ARCH), x86_64)
 EXTRA_TARBALL=cc-cloud-hypervisor-tarball \
 	cc-tdx-kernel-tarball \
-	cc-sev-kernel-tarball \
 	cc-tdx-qemu-tarball \
 	cc-tdx-td-shim-tarball \
 	cc-tdx-tdvf-tarball \
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -421,7 +421,8 @@ install_cached_kernel_component() {
 		"kata-static-cc-sev-kernel-modules.tar.xz" \
 		"${workdir}/kata-static-cc-sev-kernel-modules.tar.xz" \
 	|| return 1
-		
+	
+	mkdir -p "${module_dir}"
 	tar xvf "${workdir}/kata-static-cc-sev-kernel-modules.tar.xz" -C  "${module_dir}" && return 0
 
 	return 1


UnmeshDeodhar:
@ryansavino @fitzthum please review and let me know if this look good.

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -13,7 +13,6 @@ ARCH := $(shell uname -m)
 ifeq ($(ARCH), x86_64)
 EXTRA_TARBALL=cc-cloud-hypervisor-tarball \
 	cc-tdx-kernel-tarball \
-	cc-sev-kernel-tarball \
```
ryansavino, line 16: Missing `\` on line 121?

I'm also wondering if we should force the kernel build every time for `cc-sev-rootfs-initrd-tarball`. If the kernel tarball has previously been built, maybe we can just check to see if a specific directory exists. Not sure how the maintainers/stakeholders feel about this. @fidencio @wainersm @fitzthum

ryansavino (changes requested):

tools/packaging/kata-deploy/local-build/Makefile
UnmeshDeodhar: @ryansavino, so with this change, kernel will not be built earlier in the test. So when we are building initrd, we can force kernel build.

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -118,7 +117,7 @@ cc-qemu-tarball:
 cc-rootfs-image-tarball:
 	${MAKE} $@-build
 
-cc-sev-rootfs-initrd-tarball: cc-sev-kernel-tarball
+cc-sev-rootfs-initrd-tarball: cc-sev-kernel-tarball \
```
jepio, line 120: why the backslash?

UnmeshDeodhar: Removed it.

tools/packaging/kata-deploy/local-build/Makefile
UnmeshDeodhar: @fitzthum yes. We can skip the first build. During second build, we check if we have a cached tarball that we can download. If yes, download and use that tarball. Else, build it from scratch.

fitzthum (approved):

tools/packaging/kata-deploy/local-build/Makefile
Alex-Carter01: Using a cached kernel build should be fine for the sev initrd

UnmeshDeodhar: Right, but calling kernel build twice becomes redundant. That is why I am removing the first kernel build call and kernel will be built during initrd build. Does that look okay to you? @Alex-Carter01

wainersm: An alternative solution, as pointed out by @ryansavino, would be to depend on a previously built directory. E.g. the following work out:

```
diff --git a/tools/packaging/kata-deploy/local-build/Makefile b/tools/packaging/kata-deploy/local-build/Makefile
index 28202e90e..1ffb82713 100644
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -117,7 +117,9 @@ cc-qemu-tarball:
 cc-rootfs-image-tarball:
        ${MAKE} $@-build
 
-cc-sev-rootfs-initrd-tarball: cc-sev-kernel-tarball
+$(MK_DIR)/build/cc-sev-kernel/builddir: cc-sev-kernel-tarball
+
+cc-sev-rootfs-initrd-tarball: $(MK_DIR)/build/cc-sev-kernel/builddir
        ${MAKE} $@-build
 
 cc-tdx-rootfs-image-tarball:
```

The approach proposed on this PR has the caveats which are:
a) If the kernel was built already for some obscure reason (e.g. the developer built it manually), it will still be built twice:
```
make cc-sev-kernel-tarball && make cc-sev-rootfs-initrd-tarball
```
b) The sev kernel is included in the cc tarball only and only if the sev initrd is built, it might be error prone

And the caveats of depending on the build directory as I showed above:

a) It doesn't depend on a specific kernel version so that the build directory may exist but the initrd requires a newer kernel version, therefore the initrd build will fail (I really hope it fails!)

In the end it is a matter of preference. I'm okay to stick with the approach proposed on this PR. I just wanted to raise the alternative to keep you confused :P

wainersm (approved):

tools/packaging/kata-deploy/local-build/Makefile
UnmeshDeodhar: @wainersm my thoughts about the caveats of the approach suggested in this PR:
- Only kernel should not be built on these CI systems for any obscure reason. If someone wants to use AMD SEV, they need the kernel and the initrd. Thus, I can't think of any scenario where only AMD SEV kernel needs to be built. Thus, if someone wants to use SEV, they can just call the initrd build and we will have two binaries.
On the approach you suggested, I think I agree with you. 
Personally, I prefer the approach suggested in this PR.

wainersm:
hey @UnmeshDeodhar , thanks for the contribution!

Issue #6385, opened 2023-02-28
Kata-manager.sh -o option still aborts when containerd is pre-installed.


gabevenberg:
`pre_checks()` tests for the presence of containerd, exiting early, as shown here:

```
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh) -o"
INFO: Checking dependencies
INFO: Running pre-checks
ERROR: containerd already installed
```
Pre checks should either not check containerd at all, or should be aware of `$only_kata`.

jodh-intel:
Good catch @gabevenberg !!

PR #6419, opened 2023-03-07
make only_kata work without -f

gabevenberg:
barebones fix for #6385, simply passes through only_kata to pre_check and adds an additional conditional before aborting the install due to pre-existing containerd.

However, should we be aborting on pre-existing containerd, or should only_kata be set based on whether containerd is already installed? (so that containerd install would be skipped if it was detected to already be installed.)

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -254,8 +254,13 @@ pre_checks()
 {
 	info "Running pre-checks"
 
+	local only_kata="${1:-}"
+	[ -z "$only_kata" ] && die "no only_kata value"
+
 	command -v "${kata_shim_v2}" &>/dev/null \
 		&& die "Please remove existing $kata_project installation"
+
+	[only_kata = "false" ] && return 0
 
 	local ret
 
@@ -315,6 +320,9 @@ setup()
 	local force="${2:-}"
 	[ -z "$force" ] && die "no force value"
 
+	local only_kata="${3:-}"
+	[ -z "$only_kata" ] && die "no only_kata value"
+
 	[ "$cleanup" = "true" ] && trap cleanup EXIT
 
 	source /etc/os-release || source /usr/lib/os-release
@@ -324,7 +332,7 @@ setup()
 
 	[ "$force" = "true" ] && return 0
 
-	pre_checks
+	pre_checks "$only_kata"
 }
 
 # Download the requested version of the specified project.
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Good catch! Thanks @gabevenberg !

lgtm

utils/kata-manager.sh
```diff
@@ -254,14 +254,17 @@ pre_checks()
...
+
 	command -v "${kata_shim_v2}" &>/dev/null \
 		&& die "Please remove existing $kata_project installation"
 
 	local ret
 
 	{ containerd_installed; ret=$?; } || true
 
-	[ "$ret" -eq 0 ] && die "$containerd_project already installed"
+	[ "$ret" -eq 0 ] && [only_kata = "false" ] && die "$containerd_project already installed"
```
fidencio, line 267: It would be better to check for `only_kata` at the start of the function.
If it's true, we simply return 0, if not we proceed with function logic.  By doing this we avoid actually checking something that will never be used in case `only_kata` is true.

fidencio (commented):
I've left a comment.

utils/kata-manager.sh
gabevenberg: Sounds good, still keeping the kata-is-already-installed check, though. Fixed.

utils/kata-manager.sh
```diff
@@ -260,11 +260,13 @@ pre_checks()
...
 		&& die "Please remove existing $kata_project installation"
 
+	[only_kata = "false" ] && return 0
+
 	local ret
 
 	{ containerd_installed; ret=$?; } || true
 
-	[ "$ret" -eq 0 ] && [only_kata = "false" ] && die "$containerd_project already installed"
+	[ "$ret" -eq 0 ] && die "$containerd_project already installed"
```
amshinde: @gabevenberg Can you squash this commit with the previous one?

utils/kata-manager.sh
```diff
@@ -254,14 +254,17 @@ pre_checks()
 {
 	info "Running pre-checks"
 
+	local only_kata="${1:-}"
```
amshinde, line 257: I would prefer calling this as "skip_containerd" instead of "only_kata" with the default value being false.

gabevenberg: Only kata is what it is called way back when the cli options are parsed. Do you want me to renaime it all the way back, or just in pre-checks?

utils/kata-manager.sh
gabevenberg: done.

utils/kata-manager.sh
amshinde: I would rename it everywhere.

gabevenberg: like that?

fidencio (approved):
lgtm, thanks @gabevenberg!

amshinde (approved):

cmaf (approved):

Issue #6422, opened 2023-03-07
containerd-shim-kata failed to cleanup cgroup causing cgroup file leaks


yahaa:
# Description of problem
When delete pod running on kata-container-runtime, I found that the containerd report some error log, details  as followed:
![image](https://user-images.githubusercontent.com/16364111/223395615-26a928d0-fda6-4304-8d25-9e877d6fd54e.png)

and there are many directories under the /sys/fs/cgroup/kata_overhead directory that have not been deleted, details as followd:
![image](https://user-images.githubusercontent.com/16364111/223397006-402370c5-f7ce-47ba-ae05-dc5d997f6d9d.png)

# Expected result

* After pod deleted, related cgroup files are deleted normally
* The command `journalctl -t kata -f| grep "failed to cleanup the"` not found any error log

# Actual result
Detail informetion see `Description of problem`

# Further information
##  Kata Containers containerd shim version

```txt
Kata Containers containerd shim (Golang): id: "io.containerd.kata.v2", version: 3.1.0-rc0, commit: 0749657c737adbe7b6be648b93678c496c58195d
```
## Kernel 
```txt
Linux w 5.19.0-32-generic #33~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Jan 30 17:03:34 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

PR #6423, opened 2023-03-07
runtime: fix containerd-shim-kata failed to cleanup related cgroup files

yahaa:
Fixes: #6422

I found that the cause of this issue (#6422) is that The LinuxCgroup type returned by NewSandboxResourceController and LoadResourceController is inconsistent

In the [NewSandboxResourceController](https://github.com/kata-containers/kata-containers/blob/cc1821fb8bedb707ad87d914ff2a0df758ba2f04/src/runtime/pkg/resourcecontrol/cgroups.go#L173) function
```go
func NewSandboxResourceController(path string, resources *specs.LinuxResources, sandboxCgroupOnly bool) (ResourceController, error) {
	sandboxResources := *resources
	sandboxResources.Devices = append(sandboxResources.Devices, sandboxDevices()...)
        // ...
	if !IsSystemdCgroup(path) || !sandboxCgroupOnly {
                // this condition always true if  we set  sandbox_cgroup_only=false in the config file /etc/kata-containers/configuration.toml
               // so NewResourceController func will create an cgroup using cgroupsv2.NewManager
		return NewResourceController(path, &sandboxResources)
	}
        // ...

```

In the [LoadResourceController](https://github.com/kata-containers/kata-containers/blob/cc1821fb8bedb707ad87d914ff2a0df758ba2f04/src/runtime/pkg/resourcecontrol/cgroups.go#L228) function(this function will be call on pod deleting)
```go
func LoadResourceController(path string) (ResourceController, error) {
	var err error
	var cgroup interface{}

	//  ...
	if cgroups.Mode() == cgroups.Legacy || cgroups.Mode() == cgroups.Hybrid {
		// ...
	} else if cgroups.Mode() == cgroups.Unified {
                 // this condition always true because the function  IsSystemdCgroup(path) always return true
                //  so LoadResourceController func will return an cgroup using cgroupsv2.LoadSystemd
		if IsSystemdCgroup(path) {
			slice, unit, err := getSliceAndUnit(path)
			if err != nil {
				return nil, err
			}
			cgroup, err = cgroupsv2.LoadSystemd(slice, unit)
			if err != nil {
				return nil, err
			}
		} else {
			//...
		}
	} else {
        // ...
        }
        // ...
}
```

* edit configuration file such as /etc/kata-containers/configuration.toml, set sandbox_cgroup_only=false
* delete some pod that running on containerd-shim-kata

The solution to this issue is to make the cgroup types returned by NewSandboxResourceController and LoadResourceController be the same.

Signed-off-by: yahaa <1477765176@qq.com>

### DIFF
--- a/src/runtime/pkg/resourcecontrol/cgroups.go
+++ b/src/runtime/pkg/resourcecontrol/cgroups.go
@@ -41,10 +41,11 @@ func RenameCgroupPath(path string) (string, error) {
 }
 
 type LinuxCgroup struct {
-	cgroup  interface{}
-	path    string
-	cpusets *specs.LinuxCPU
-	devices []specs.LinuxDeviceCgroup
+	cgroup            interface{}
+	path              string
+	cpusets           *specs.LinuxCPU
+	devices           []specs.LinuxDeviceCgroup
+	sandboxCgroupOnly bool
 
 	sync.Mutex
 }
@@ -225,7 +226,7 @@ func NewSandboxResourceController(path string, resources *specs.LinuxResources, 
 	}, nil
 }
 
-func LoadResourceController(path string) (ResourceController, error) {
+func LoadResourceController(path string, sandboxCgroupOnly bool) (ResourceController, error) {
 	var err error
 	var cgroup interface{}
 
@@ -241,7 +242,7 @@ func LoadResourceController(path string) (ResourceController, error) {
 			return nil, err
 		}
 	} else if cgroups.Mode() == cgroups.Unified {
-		if IsSystemdCgroup(path) {
+		if IsSystemdCgroup(path) && sandboxCgroupOnly {
 			slice, unit, err := getSliceAndUnit(path)
 			if err != nil {
 				return nil, err
@@ -261,8 +262,9 @@ func LoadResourceController(path string) (ResourceController, error) {
 	}
 
 	return &LinuxCgroup{
-		path:   path,
-		cgroup: cgroup,
+		sandboxCgroupOnly: sandboxCgroupOnly,
+		path:              path,
+		cgroup:            cgroup,
 	}, nil
 }
 
@@ -275,7 +277,7 @@ func (c *LinuxCgroup) Delete() error {
 	case cgroups.Cgroup:
 		return cg.Delete()
 	case *cgroupsv2.Manager:
-		if IsSystemdCgroup(c.ID()) {
+		if IsSystemdCgroup(c.ID()) && c.sandboxCgroupOnly {
 			if err := cg.DeleteSystemd(); err != nil {
 				return err
 			}
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -2265,14 +2265,16 @@ func (s *Sandbox) resourceControllerDelete() error {
 		return nil
 	}
 
-	sandboxController, err := resCtrl.LoadResourceController(s.state.SandboxCgroupPath)
+	sandboxController, err := resCtrl.LoadResourceController(s.state.SandboxCgroupPath, s.config.SandboxCgroupOnly)
 	if err != nil {
 		return err
 	}
 
 	resCtrlParent := sandboxController.Parent()
-	if err := sandboxController.MoveTo(resCtrlParent); err != nil {
-		return err
+	if resCtrlParent != "." {
+		if err := sandboxController.MoveTo(resCtrlParent); err != nil {
+			return err
+		}
 	}
 
 	if err := sandboxController.Delete(); err != nil {
@@ -2280,7 +2282,7 @@ func (s *Sandbox) resourceControllerDelete() error {
 	}
 
 	if s.state.OverheadCgroupPath != "" {
-		overheadController, err := resCtrl.LoadResourceController(s.state.OverheadCgroupPath)
+		overheadController, err := resCtrl.LoadResourceController(s.state.OverheadCgroupPath, s.config.SandboxCgroupOnly)
 		if err != nil {
 			return err
 		}


jepio:
Hi @yahaa,
how does this fix the issue? can you extend the commit message to explain what the issue is and how you are fixing it.
The added `sandboxCgroupOnly` parameter/config field is not set to true anywhere, so it looks like you're effectively removing some code (all the blocks that check `sandboxCgroupOnly` are false), yes?

yahaa:
> Hi @yahaa, how does this fix the issue? can you extend the commit message to explain what the issue is and how you are fixing it. The added `sandboxCgroupOnly` parameter/config field is not set to true anywhere, so it looks like you're effectively removing some code (all the blocks that check `sandboxCgroupOnly` are false), yes?

Hi @jepio Sorry for the little information I provided. 

 I found that the cause of this issue (#6422) is that The LinuxCgroup type returned by NewSandboxResourceController and LoadResourceController is inconsistent

In the [NewSandboxResourceController](https://github.com/kata-containers/kata-containers/blob/cc1821fb8bedb707ad87d914ff2a0df758ba2f04/src/runtime/pkg/resourcecontrol/cgroups.go#L173) function
```go
func NewSandboxResourceController(path string, resources *specs.LinuxResources, sandboxCgroupOnly bool) (ResourceController, error) {
	sandboxResources := *resources
	sandboxResources.Devices = append(sandboxResources.Devices, sandboxDevices()...)
        // ...
	if !IsSystemdCgroup(path) || !sandboxCgroupOnly { 
                // this condition always true if  we set  sandbox_cgroup_only=false in the config file /etc/kata-containers/configuration.toml
               // so NewResourceController func will create an cgroup using cgroupsv2.NewManager
		return NewResourceController(path, &sandboxResources) 
	}
        // ...

```

In the [LoadResourceController](https://github.com/kata-containers/kata-containers/blob/cc1821fb8bedb707ad87d914ff2a0df758ba2f04/src/runtime/pkg/resourcecontrol/cgroups.go#L228) function(this function will be call on pod deleting)
```go
func LoadResourceController(path string) (ResourceController, error) {
	var err error
	var cgroup interface{}

	//  ...
	if cgroups.Mode() == cgroups.Legacy || cgroups.Mode() == cgroups.Hybrid {
		// ...
	} else if cgroups.Mode() == cgroups.Unified {
                 // this condition always true because the function  IsSystemdCgroup(path) always return true
                //  so LoadResourceController func will return an cgroup using cgroupsv2.LoadSystemd
		if IsSystemdCgroup(path) {
			slice, unit, err := getSliceAndUnit(path)
			if err != nil {
				return nil, err
			}
			cgroup, err = cgroupsv2.LoadSystemd(slice, unit)
			if err != nil {
				return nil, err
			}
		} else {
			//...
		}
	} else {
        // ...
        }
        // ...
}
```

So when the pod is deleted, we will encounter the issue mentioned in #6422

The solution to this issue is to make the cgroup types returned by NewSandboxResourceController and LoadResourceController be the same.



And the values of `sandboxCgroupOnly` fields will come from  the config file /etc/kata-containers/configuration.toml

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -241,7 +242,7 @@ func LoadResourceController(path string) (ResourceController, error) {
 			return nil, err
 		}
 	} else if cgroups.Mode() == cgroups.Unified {
-		if IsSystemdCgroup(path) {
+		if IsSystemdCgroup(path) && sandboxCgroupOnly {
```
jepio, line 245: What about when `cgroups.Mode() != cgroups.Unified`? Doesn't that hit the same bug?

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2265,22 +2265,24 @@ func (s *Sandbox) resourceControllerDelete() error {
...
-	sandboxController, err := resCtrl.LoadResourceController(s.state.SandboxCgroupPath)
+	sandboxController, err := resCtrl.LoadResourceController(s.state.SandboxCgroupPath, s.config.SandboxCgroupOnly)
 	if err != nil {
 		return err
 	}
 
 	resCtrlParent := sandboxController.Parent()
-	if err := sandboxController.MoveTo(resCtrlParent); err != nil {
-		return err
+	if resCtrlParent != "." {
```
jepio, line 2274: This is an unrelated fix right? This might make more sense as a separate commit with a good description.

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -275,7 +277,7 @@ func (c *LinuxCgroup) Delete() error {
 	case cgroups.Cgroup:
 		return cg.Delete()
 	case *cgroupsv2.Manager:
-		if IsSystemdCgroup(c.ID()) {
+		if IsSystemdCgroup(c.ID()) && c.sandboxCgroupOnly {
```
jepio, line 280: Whats about when we're using `cgroups.Cgroup` as the manager, does that also need the same fix?

jepio (commented):
Can you update the commit message to include the information you posted as a comment above?

yahaa:
> Can you update the commit message to include the information you posted as a comment above?

I have added it  on commit message.

src/runtime/pkg/resourcecontrol/cgroups.go
yahaa: That's probably not a bug, `NewSandboxResourceController` has handle this case, worst case is that `LoadResourceController` will return `nil, ErrCgroupMode`

src/runtime/pkg/resourcecontrol/cgroups.go
yahaa: Maybe we need to do more tests, but i can't construct a case where `cgroups.Cgroup` acts as a manager in my environment.

src/runtime/virtcontainers/sandbox.go
yahaa: This should be related, when i set sandbox_cgroup_only=true in /etc/kata-containers/configuration.toml, `sandboxController` will use `cgroupsv2.LoadSystemd` as the manager, but the `sandboxController.Parent()` return `.`, that cause `sandboxController.MoveTo(resCtrlParent)` return `cgroupsv2.ErrInvalidGroupPath
` error.

My solution is to skip it, but I'm not sure if that is a good solution.

src/runtime/pkg/resourcecontrol/cgroups.go
jepio: What is your environment? Try Ubuntu 20.04.

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -41,10 +41,11 @@ func RenameCgroupPath(path string) (string, error) {
...
 type LinuxCgroup struct {
-	cgroup  interface{}
-	path    string
-	cpusets *specs.LinuxCPU
-	devices []specs.LinuxDeviceCgroup
+	sandboxCgroupOnly bool
+	cgroup            interface{}
+	path              string
+	cpusets           *specs.LinuxCPU
+	devices           []specs.LinuxDeviceCgroup
```
jepio, line 48: govet is not happy with the placement of the `sandboxCgroupOnly` field, try the following:
```suggestion
	cgroup            interface{}
	path              string
	cpusets           *specs.LinuxCPU
	devices           []specs.LinuxDeviceCgroup
	sandboxCgroupOnly bool
```

src/runtime/pkg/resourcecontrol/cgroups.go
yahaa: After setting up an environment with Ubuntu 20.04, I will try it.

src/runtime/pkg/resourcecontrol/cgroups.go
jepio: @yahaa you'll need to fix the commit message for this commit, see the checks done here: https://github.com/kata-containers/kata-containers/actions/runs/4376736826/jobs/7675992361

jepio: So it needs a signoff, and a subsystem in the subject line. For the first commit you don't need to copy paste code in, but a description of what going on is helpful.

jepio:
Hi @yahaa,
would you be able to squash the two commits into one and edit the commit message to stay within the line length limit? It does not need to contain code samples, a description of what is happening is enough.

yahaa:
> Hi @yahaa, would you be able to squash the two commits into one and edit the commit message to stay within the line length limit? It does not need to contain code samples, a description of what is happening is enough.

Okay, thank you for the reminder.

gkurz:
> > Can you update the commit message to include the information you posted as a comment above?
> 
> I have added it on commit message.

Hi @yahaa ! Thanks for working on this.

The commit message could maybe provide some more clarifications, e.g.

```
When creating a new resource controller and `sandbox_cgroup_only` is false, cgroupfs is assumed (see https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/resourcecontrol/cgroups.go#L172).
Make sure this assumption is honored by the rest of the code.
```

Basically this means that `sandbox_cgroup_only == false` should follow the same paths as `IsSystemdCgroup() == false`.
Maybe this assumption could be consolidated in a helper and use it instead of `IsSystemdCgroup()` ?

gkurz:
Please propagate the API change of `LoadResourceController` to `cgroups_darwin.go` as well.

kinderyj:
WIP?  I have the same issue in kata 3.2.0

kinderyj:
"Is this a work in progress (WIP)? I'm experiencing the same issue in kata 3.2.0."

kinderyj:
Is this a work in progress (WIP)? I'm experiencing the same issue in kata 3.2.0.

Issue #6248, opened 2023-02-09
runtime-rs: resize pty error


YushuoEdge:
When I try to exec command in the container with runtime-rs and nerdctl, there will report an resize pty error, like:
```
$ nerdctl exec -it 2124ed776abb /bin/bash
ERRO[0000] resize pty                                    error="Others(\"failed to handler message handler request\\n\\nCaused by:\\n    0: resize pty\\n    1: resize pty\\n    2: resize pty\\n    3: rpc status: code: UNAVAILABLE message: \\\"\\\\\\\"invalid argument: Invalid exec id\\\\\\\"\\\"\"): unknown"
[root@clr-e9b4aae9f214405fb530d97c197f7bce /]#
```

Tim-0731-Hzt:
Could you provide the environment you use?

YushuoEdge:
```
$ nerdctl --version
nerdctl version 1.1.0
$ containerd --version
containerd github.com/containerd/containerd v1.7.0-beta.0 d878d7dda8d97f7f9d1d50b4a5d36d2d8fae9162
$ containerd-shim-kata-v2 --version
Kata Containers containerd shim (Rust): id: io.containerd.kata.v2, version: 3.1.0-alpha1, commit: 91bb2cdf8944ed99583c6a1a05d7251b607ac601
```

PR #6426, opened 2023-03-08
bugfix: modify tty_win info in runtime when handling ResizePtyRequest

YushuoEdge:
Currently, we only create the new exec process in runtime, this will cause error when the following requests needing to be handled:

- Task: exec process
- Task: resize process pty
- ...

The agent do not do_exec_process when we handle ExecProcess, thus we can not find any process information in the guest when we handle ResizeProcessPty. This will report an error.
    
In this commit, the handling process is modified to the:
    * Modify process tty_win information in runtime
    * If the exec process is not running, we just return. And the truly pty_resize will happen when start_process

Fixes: #6248

### DIFF
--- a/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
@@ -344,20 +344,33 @@ impl Container {
         height: u32,
     ) -> Result<()> {
         let logger = logger_with_process(process);
-        let inner = self.inner.read().await;
+        let mut inner = self.inner.write().await;
         if inner.init_process.get_status().await != ProcessStatus::Running {
             warn!(logger, "container is not running");
             return Ok(());
         }
-        self.agent
-            .tty_win_resize(agent::TtyWinResizeRequest {
-                process_id: process.clone().into(),
-                row: height,
-                column: width,
-            })
-            .await
-            .context("resize pty")?;
-        Ok(())
+
+        if process.exec_id.is_empty() {
+            inner.init_process.height = height;
+            inner.init_process.width = width;
+        } else if let Some(exec) = inner.exec_processes.get_mut(&process.exec_id) {
+            exec.process.height = height;
+            exec.process.width = width;
+
+            // for some case, resize_pty request should be handled while the process has not been started in agent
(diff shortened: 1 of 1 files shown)


YushuoEdge:
@quanweiZhou @Tim-0731-Hzt  PTAL!

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -344,20 +344,35 @@ impl Container {
...
+                return Ok(());
+            }
+        } else {
+            warn!(
+                sl!(),
+                "could not find process {} in container {}",
+                process.exec_id(),
+                process.container_id()
+            );
+            return Ok(());
```
Tim-0731-Hzt, line 372: why should we return OK if the process is invalid

quanweiZhou (approved):
Thanks, @YushuoEdge LGTM.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -344,20 +344,35 @@ impl Container {
...
-            .tty_win_resize(agent::TtyWinResizeRequest {
-                process_id: process.clone().into(),
-                row: height,
-                column: width,
-            })
-            .await
-            .context("resize pty")?;
-        Ok(())
+
+        if process.exec_id.is_empty() {
```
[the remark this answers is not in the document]
YushuoEdge: Actually, I wrote it here just imitating what is wrote in agent, see: https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/sandbox.rs#L239-L252

The request comes from upper layer, and maybe there are some cases the request did not contain eid.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
YushuoEdge: I don't think this is on the critical path, so I just return `Ok(())` here, like in the above, we will not return `Err(...)` when the container is not in Running state.

YushuoEdge: OK, fixed.

Tim-0731-Hzt (approved):

Issue #6430, opened 2023-03-08
runtime-rs: ch: Add confidential guest support


jodh-intel:
Add the ability for the runtime-rs Cloud Hypervisor (CH) feature ([currently disabled](https://github.com/kata-containers/kata-containers/pull/6201)) to generate a CH configuration for confidential guests.

jodh-intel:
Blocked on #6429.

jodh-intel:
Now unblocked.

Note that this issue simply adds the generic confidential guest handling and does not enable Intel TDX. That work will be convered on #6448.

PR #6201, opened 2023-02-02
runtime-rs: Add basic CH implementation

jodh-intel:
Add a basic runtime-rs `Hypervisor` trait implementation for Cloud Hypervisor (CH).

> **Note:**
>
> This only supports a default Kata configuration for CH currently.

Fixes: #5242.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

Issue #6429, opened 2023-03-08
runtime-rs: Cloud Hypervisor now fails with netns error


jodh-intel:
(... aaaand this is what happens when we have features disabled by default (_mea culpa_ :smile:))

It's looking like #5788 has broken the CH external hypervisor feature (#6201). Enabling CH and testing results in a panic due to the following not being implemented:

```rust
pub(crate) async fn get_vmm_master_tid(&self) -> Result<u32> {
    todo!()
}
```

If I update that code to return the PID of the hypervisor as `Ok(pid)`, runtime-rs crashes as follows:

```bash
ctr: failed to create shim task: Others("failed to handler message create container\n\nCaused by:\n    0: vmm netns guard\n    1: open new netns path /proc/675845/task/675850
/ns/net\n    2: No such file or directory (os error 2)"): unknown
```

Any thoughts @YushuoEdge? Is there anything else the external hypervisors need to do to work with the new API?

ananos:
I think it will definitely break FC. If I understand correctly it is related to this: #6384

YushuoEdge:
Hi @jodh-intel @ananos , this API is added when we need to support OCI hook. Since docker hook(CreateContainer hook) rely on the pid of the container to find out the container namespace on the host. So in runtime-rs + dragonball, we set state.pid to the vmm master thread id for that hook, as it is the only thread in the namespace.

jodh-intel:
Hi @YushuoEdge - Just to confirm, are you planning to raise a PR to resolve this? I can try testing your fix if you haven't tried it with CH?

YushuoEdge:
> Hi @YushuoEdge - Just to confirm, are you planning to raise a PR to resolve this? I can try testing your fix if you haven't tried it with CH?

Yes, I have not tested with CH yet. It seems that CH has not implemented netns related contents when startup, so I am not sure whether it is appropriate to do like the commit. If it's better to discuss under PR, maybe I can raise a fix PR first.

YushuoEdge:
Hi @jodh-intel , I have split the task into two issues: https://github.com/kata-containers/kata-containers/issues/6441, https://github.com/kata-containers/kata-containers/pull/6443. And I have already raised a PR for the second issue, so that your development will not be blocked.

jodh-intel:
Fixed by #6443. Thanks @YushuoEdge ! :smile:

PR #6439, opened 2023-03-09
runtime-rs: ch: Implement confidential guest handling

jodh-intel:
This change provides a preliminary implementation for the Cloud Hypervisor (CH) feature ([currently
disabled](https://github.com/kata-containers/kata-containers/pull/6201)) to allow it to generate the CH configuration for handling confidential guests.

This change also introduces concrete errors using the `thiserror` crate (see `src/runtime-rs/crates/hypervisor/ch-config/src/errors.rs`) and a lot of unit tests for the conversion code that generates the CH configuration from the generic Hypervisor configuration.

Fixes: #6430.


jodh-intel:
Raising to get some visibility on this. However, it's currently DNM as it's blocked on #6429: once we have a fix for that, I can re-test with CH.

**Note to self**: I also need to remove the `FIXME` temporary commit to enable CH by default :smile:

jodh-intel:
Note to reviewers as it's easy to miss this: most of the changes are in `convert.rs`, which GH hides by default due to the size of the diff. Hence, youv'e got to click the "Load diff" link.

jodh-intel:
Re-tested with CH feature enabled now that #6443 has landed and removed the temporary commit, so this is ready for review.

jodh-intel:
Notes:

- This code is still for the disabled-by-default Cloud Hypervisor feature (see #6201).
- Intel TDX detection will be added on #6448.

jodh-intel:
Notes:

- This code is still for the disabled-by-default Cloud Hypervisor feature (see #6201).
- Intel TDX detection will be added on #6448.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -229,7 +229,13 @@ impl CloudHypervisorInner {
     async fn cloud_hypervisor_launch(&mut self, _timeout_secs: i32) -> Result<()> {
         self.cloud_hypervisor_ensure_not_launched().await?;
 
-        let debug = false;
+        let cfg = self
+            .config
+            .as_ref()
+            .ok_or("no hypervisor config for CH")
+            .map_err(|e| anyhow!(e))?;
```
c3d, line 257: Do we want a hard error for a missing config, or wouldn't that be a case where you would want to set `debug` to `true`?

jodh-intel: @c3d - That's a good question but that code is used a few times and is doing "double time" as here we want to know:

- That this API was called after we set the generic hypervisor config passed to us by runtime-rs (in `set_hypervisor_config()`). 
- If debug was requested.

I don't like the duplication but I'm being paranoid to protect against runtime-rs calling the hypervisor methods in the wrong order.

src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
```diff
@@ -17,121 +17,214 @@ use std::convert::TryFrom;
...
         let cfg = n.cfg;
-        let vsock_socket_path = n.vsock_socket_path;
-        let sandbox_path = n.sandbox_path;
+
+        let debug = cfg.debug_info.enable_debug;
+        let confidential_guest = cfg.security_info.confidential_guest;
+
+        let tdx_enabled = n.tdx_enabled;
+
+        let vsock_socket_path = if n.vsock_socket_path.is_empty() {
```
c3d, line 48: I find this style disconcerting, relative to

```
let vsock_socket_path = n.vsock_socket_path
if vsock_socket_path.is_empty() {
            return Err(VmConfigError::EmptyVsockSocketPath);
}
```

src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
```diff
@@ -17,121 +17,214 @@ use std::convert::TryFrom;
...
-        } else {
-            None
-        };
-
-        let pmem = if use_initrd {
+        let pmem = if use_initrd || confidential_guest {
             None
         } else {
-            let pmem = PmemConfig::try_from(&boot_info)?;
+            let pmem = PmemConfig::try_from(&boot_info).map_err(VmConfigError::PmemError)?;
```
c3d, line 93: I wonder if it would not be better to use a `.context()` here instead, to avoid losing the original error.

cmaf (approved):

amshinde (approved):

src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
jodh-intel: @c3d - This is a bit subtle but it's not losing the original error, it's returning it: `PmemConfig::try_from()` returns a `PmemConfigError` and the `map_err()` is returning that original error by wrapping it in a `VmConfigError::PmemError`, which is defined as:

```rust
#[error("Pmem config error: {0}")]
PmemError(PmemConfigError),
```

Issue #6442, opened 2023-03-10
runtime-rs: should consider external vmm when trying to get netns in host


YushuoEdge:
Now, when we try to get host netns, we just use the path `/proc/{shim.pid}/task/{vmm_master.tid}/net`, which could not work with external vmm.

Related to: https://github.com/kata-containers/kata-containers/issues/6429

PR #6443, opened 2023-03-10
bugfix: add get_ns_path API for Hypervisor

YushuoEdge:
For external hypervisors(qemu, cloud-hypervisor, ...), the ns they launch vm in is different from internal hypervisor(dragonball). And when we doing CreateContainer hook, we will rely on the netns path. So we add a get_ns_path API.

Fixes: #6442

/cc @jodh-intel @pmores @amshinde


jodh-intel (approved):
Thanks very much @YushuoEdge !! I've tested CH works with this PR when the CH feature is enabled :smile:

lgtm

Issue #6449, opened 2023-03-12
Support multiple architecture publishing


singhwang:
**Which feature do you think can be improved?**

The current release does not support multi-architecture publishing. The community can consider adding this feature to the release.

**How can it be improved?**

Multi-architecture publishing is supported in release, including kata-deploy container image and kata-static compression package.

PR #6450, opened 2023-03-12
main | release: Fix multi-arch publishing is not supported

singhwang:
release: Fix multi-arch publishing is not supported

When release is published, kata-deploy image and kata-static package can support multi-arch publishing.

Fixes: #6449

Signed-off-by: Singh Wang <wangxin_0611@126.com>

### DIFF
--- a/.github/workflows/release-amd64.yaml
+++ b/.github/workflows/release-amd64.yaml
@@ -1,0 +1,63 @@
+name: Publish Kata release artifacts for amd64
+on:
+  workflow_call:
+    inputs:
+      target-arch:
+        required: true
+        type: string
+
+jobs:
+  build-kata-static-tarball-amd64:
+    uses: ./.github/workflows/build-kata-static-tarball-amd64.yaml
+
+  kata-deploy:
+    needs: build-kata-static-tarball-amd64
+    runs-on: ubuntu-latest
+    steps:
+      - name: Login to Kata Containers docker.io
+        uses: docker/login-action@v3
+        with:
+          registry: docker.io
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Login to Kata Containers quay.io
+        uses: docker/login-action@v3
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_DEPLOYER_USERNAME }}
+          password: ${{ secrets.QUAY_DEPLOYER_PASSWORD }}
+
+      - uses: actions/checkout@v3
+      - name: get-kata-tarball
+        uses: actions/download-artifact@v3
+        with:
+          name: kata-static-tarball-amd64
+
+      - name: build-and-push-kata-deploy-ci-amd64
+        id: build-and-push-kata-deploy-ci-amd64
+        run: |
+          tag=$(echo $GITHUB_REF | cut -d/ -f3-)
+          pushd $GITHUB_WORKSPACE
+          git checkout $tag
+          pkg_sha=$(git rev-parse HEAD)
+          popd
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh \
+            $(pwd)/kata-static.tar.xz "docker.io/katadocker/kata-deploy-ci" \
+            "${pkg_sha}-${{ inputs.target-arch }}"
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh \
+            $(pwd)/kata-static.tar.xz "quay.io/kata-containers/kata-deploy-ci" \
+            "${pkg_sha}-${{ inputs.target-arch }}"
+
+      - name: push-tarball
+        run: |
+          # tag the container image we created and push to DockerHub
+          tag=$(echo $GITHUB_REF | cut -d/ -f3-)
+          tags=($tag)
+          tags+=($([[ "$tag" =~ "alpha"|"rc" ]] && echo "latest" || echo "stable"))
+          for tag in ${tags[@]}; do
+            docker tag docker.io/katadocker/kata-deploy-ci:${{steps.build-and-push-kata-deploy-ci-amd64.outputs.PKG_SHA}}-${{ inputs.target-arch }} docker.io/katadocker/kata-deploy:${tag}-${{ inputs.target-arch }}
+            docker tag quay.io/kata-containers/kata-deploy-ci:${{steps.build-and-push-kata-deploy-ci-amd64.outputs.PKG_SHA}}-${{ inputs.target-arch }} quay.io/kata-containers/kata-deploy:${tag}-${{ inputs.target-arch }}
+            docker push docker.io/katadocker/kata-deploy:${tag}-${{ inputs.target-arch }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release-arm64.yaml
```diff
@@ -0,0 +1,167 @@
...
+          $(pwd)/kata-static.tar.xz "docker.io/katadocker/kata-deploy-ci" \
+          "${pkg_sha}-${{ inputs.target-arch }}"
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh \
+          $(pwd)/kata-static.tar.xz "quay.io/kata-containers/kata-deploy-ci" \
+          "${pkg_sha}-${{ inputs.target-arch }}"
+          mkdir -p packaging/kata-deploy
+          ln -s $GITHUB_WORKSPACE/tools/packaging/kata-deploy/action packaging/kata-deploy/action
+          echo "PKG_SHA=${pkg_sha}" >> $GITHUB_OUTPUT
+
+      - name: test-kata-deploy-ci-arm64-in-aks
```
bergwolf, line 121: Does ASK support this?

singhwang: @bergwolf This piece is not supported. I didn’t communicate well with @fidencio before. I will adjust this and resubmit later.

singhwang: @bergwolf I have resubmitted the code.

fidencio (commented):
@singhwang, nice work, I've left a comment.

.github/workflows/release-s390x.yaml
```diff
@@ -0,0 +1,157 @@
...
+          tags=($tag)
+          tags+=($([[ "$tag" =~ "alpha"|"rc" ]] && echo "latest" || echo "stable"))
+          for tag in ${tags[@]}; do \
+            docker tag docker.io/katadocker/kata-deploy-ci:${{steps.build-and-push-kata-deploy-ci-s390x.outputs.PKG_SHA}}-${{ inputs.target-arch }} docker.io/katadocker/kata-deploy:${tag}-${{ inputs.target-arch }} && \
+            docker tag quay.io/kata-containers/kata-deploy-ci:${{steps.build-and-push-kata-deploy-ci-s390x.outputs.PKG_SHA}}-${{ inputs.target-arch }} quay.io/kata-containers/kata-deploy:${tag}-${{ inputs.target-arch }} && \
+            docker push docker.io/katadocker/kata-deploy:${tag}-${{ inputs.target-arch }} && \
+            docker push quay.io/kata-containers/kata-deploy:${tag}-${{ inputs.target-arch }}; \
+          done
+
+  upload-static-tarball:
```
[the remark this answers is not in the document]
singhwang: @fidencio Regarding your first question, do you want to say that we don't need to publish the kata static compression package of s390x on github?

Regarding the second question, I don't quite understand what you mean?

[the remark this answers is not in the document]
singhwang: > I see this was not changed as hub doesn't provide any binary to s390x, correct?
> 
> Wouldn't make sense to just use amd64 for the arm case as well, and move this one to the generic yaml file?

The first question is that there is currently no kata static package for s390x on github. The modified logic will be compiled and uploaded to github.

The second question, according to my understanding of kata documents, plus the yaml file of payload-after-push.yaml provided by you before, I think it is very necessary to provide kata static package and kata-deploy image for arm64 environment . I used to be a user, and I deeply felt how bad the kata static package and kata deploy image experience without the arm64 environment was, so I had to compile and produce these two very important products by myself.

fidencio:
I've rebased this one atop of current `main` as a bunch of work got in that caused conflicts on this one.
As the work was added by myself, I consider it easier for me to rebase than leaving the burden on the contributor.

Please, I'd like to get reviews on this one as it'd be really nice to have it as part of the release I'm close to cut.

jongwu (approved):

.github/workflows/release.yaml
```diff
@@ -81,8 +89,36 @@ jobs:
           GITHUB_TOKEN=${{ secrets.GIT_UPLOAD_TOKEN }} hub release edit -m "" -a "${tarball}" "${tag}"
           popd
 
+      - name: download-artifacts-arm64
```
jepio, line 92: shouldn't this whole thing just be a for loop:
```
for arch in arm64 amd64 s390x; do
   ...
done
```

.github/workflows/release-amd64.yaml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          # tag the container image we created and push to DockerHub
+          tag=$(echo $GITHUB_REF | cut -d/ -f3-)
+          tags=($tag)
+          tags+=($([[ "$tag" =~ "alpha"|"rc" ]] && echo "latest" || echo "stable"))
+          for tag in ${tags[@]}; do \
+            docker tag docker.io/katadocker/kata-deploy-ci:${{steps.build-and-push-kata-deploy-ci-amd64.outputs.PKG_SHA}}-${{ inputs.target-arch }} docker.io/katadocker/kata-deploy:${tag}-${{ inputs.target-arch }} && \
+            docker tag quay.io/kata-containers/kata-deploy-ci:${{steps.build-and-push-kata-deploy-ci-amd64.outputs.PKG_SHA}}-${{ inputs.target-arch }} quay.io/kata-containers/kata-deploy:${tag}-${{ inputs.target-arch }} && \
+            docker push docker.io/katadocker/kata-deploy:${tag}-${{ inputs.target-arch }} && \
+            docker push quay.io/kata-containers/kata-deploy:${tag}-${{ inputs.target-arch }}; \
```
jepio, line 47: why the chaining with `&&`? these can just be independent lines.
same in other files

.github/workflows/release-arm64.yaml
```diff
@@ -0,0 +1,63 @@
...
+        id: build-and-push-kata-deploy-ci-arm64
+        run: |
+          tag=$(echo $GITHUB_REF | cut -d/ -f3-)
+          pushd $GITHUB_WORKSPACE
+          git checkout $tag
+          pkg_sha=$(git rev-parse HEAD)
+          popd
+          ./tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh \
+          $(pwd)/kata-static.tar.xz "docker.io/katadocker/kata-deploy-ci" \
+          "${pkg_sha}-${{ inputs.target-arch }}"
```
jepio, line 47: if these are line continuations then it would be nice to indent them.

same elsewhere in this PR

.github/workflows/release.yaml
fidencio: To do this using GHA is not as simple as it'd be to do it in bash, unfortunately.

.github/workflows/release-amd64.yaml
fidencio: Right now this is following what's been already there.
I can change, I just am not sure if it will work as expected using the GHA.

.github/workflows/release-arm64.yaml
fidencio: Ack!

jepio (approved):

PR #6458, opened 2023-03-14
runtime-rs: update load_config comment

studychao:
Since shimv2 create task option is already implemented, we need to update the corresponding comments.

Also, the ordering is also updated to fit with the code.

fixes: #3961

### DIFF
--- a/src/runtime-rs/crates/runtimes/src/manager.rs
+++ b/src/runtime-rs/crates/runtimes/src/manager.rs
@@ -353,9 +353,10 @@ impl RuntimeHandlerManager {
 
 /// Config override ordering(high to low):
 /// 1. podsandbox annotation
-/// 2. shimv2 create task option
-/// TODO: https://github.com/kata-containers/kata-containers/issues/3961
-/// 3. environment
+/// 2. environment variable
+/// 3. shimv2 create task option
+/// 4. If above three are not set, then get default path from DEFAULT_RUNTIME_CONFIGURATIONS
+/// in kata-containers/src/libs/kata-types/src/config/default.rs, in array order.
 fn load_config(spec: &oci::Spec, option: &Option<Vec<u8>>) -> Result<TomlConfig> {
     const KATA_CONF_FILE: &str = "KATA_CONF_FILE";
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -353,9 +353,8 @@ impl RuntimeHandlerManager {
 
 /// Config override ordering(high to low):
 /// 1. podsandbox annotation
-/// 2. shimv2 create task option
-/// TODO: https://github.com/kata-containers/kata-containers/issues/3961
-/// 3. environment
+/// 2. environment variable
+/// 3. shimv2 create task option
```
fidencio, line 357: Please, add `/// 4. Default locations, based on https://github.com/kata-containers/kata-containers/blob/dc42f0a33b4e60750cda9aba241d25a88cbc65c2/src/libs/kata-types/src/config/default.rs#L14-L20

studychao: Thanks, changed.

fidencio (approved):
lgtm, thanks @studychao!

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -353,9 +353,10 @@ impl RuntimeHandlerManager {
...
+/// 4. If above three are not set, then default path in kata-containers/src/libs/kata-types/src/config/default.rs
```
c3d, line 358:
```suggestion
/// 4. If above three are not set, then get default path from DEFAULT_RUNTIME_CONFIGURATIONS
///    in kata-containers/src/libs/kata-types/src/config/default.rs, in array order
```

c3d (approved):
Looks good to me, but I suggested a slight wording change to indicate that `DEFAULT_RUNTIME_CONFIGURATIONS` itself has an order, since it's an array.

quanweiZhou (approved):

src/runtime-rs/crates/runtimes/src/manager.rs
studychao: Thanks, changed.

PR #6467, opened 2023-03-15
qemu/arm64: disable image nvdimm once no firmware offered

jongwu:
For now, image nvdimm on qemu/arm64 depends on UEFI/ACPI, so if there
    is no firmware offered, it should be disabled.

### DIFF
--- a/src/runtime/pkg/katautils/config.go
+++ b/src/runtime/pkg/katautils/config.go
@@ -768,6 +768,16 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		kataUtilsLogger.Info("Setting 'disable_image_nvdimm = true' as microvm does not support NVDIMM")
 	}
 
+	// Nvdimm can only be support when UEFI/ACPI is enabled on arm64, otherwise disable it.
+	if goruntime.GOARCH == "arm64" && firmware == "" {
+		if p, err := h.PFlash(); err == nil {
+			if len(p) == 0 {
+				h.DisableImageNvdimm = true
+				kataUtilsLogger.Info("Setting 'disable_image_nvdimm = true' if there is no firmware specified")
+			}
+		}
+	}
+
 	blockDriver, err := h.blockDeviceDriver()
 	if err != nil {
 		return vc.HypervisorConfig{}, err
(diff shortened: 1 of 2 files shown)


fidencio:
@jongwu, this is a very good starting point.  However, we still need to make sure that:
* We build and ship OVMF as part of the tests, using the kata-deploy local-build scripts
* We build and ship OVMF as part of the releases and other github actions.

Would you be able to include the second item as part of this PR?

jongwu:
Hi @fidencio -, I'm not familiar with the grammar of the github action. I think qemu OVMF has been built when build qemu. What we need is just give the right path in kata configuration. Where should I set it base on github action? And how can I check the github action build result?

fidencio:
As part of the GitHub actions, we need to ensure we build OVMF here:
* https://github.com/kata-containers/kata-containers/blob/dc42f0a33b4e60750cda9aba241d25a88cbc65c2/.github/workflows/payload-after-push-arm64.yaml#L14-L22
  * In order to do that, we need to forward port this patch from `CCv0` branch: https://github.com/kata-containers/kata-containers/pull/4170

With those two patches, we start building OVMF and pushing it as part of each PR merged.

Then, to have this fixed as part of the release, we need to change:
* https://github.com/kata-containers/kata-containers/blob/dc42f0a33b4e60750cda9aba241d25a88cbc65c2/.github/workflows/release.yaml#L12-L22

With those changes we'd be good to go on the `kata-containers`  repo.

After that, we can discuss the changes needed on the `tests`  repo.

Let me know if the instructions are not clear, or if I can help answering any other question.

fidencio:
> 

Oh, I thought you actually needed the OVMF built for ARM as well.
There's no difference, and if you think that's enough, then that's enough.

jongwu:
Hi @fidencio -, I disable image nvdimm once there is no firmware offered on arm64. I think it's reasonable that user can run kata without firmware by default. Can I get your approval?

fidencio (approved):
lgtm, thanks @jongwu!

jongwu:
Help TAL this? @liubin @jodh-intel. Thanks!

liubin (approved):

jodh-intel (approved):
Thanks @jongwu.

lgtm

Issue #6247, opened 2023-02-09
runtime-rs: wrong username and hostname in the container


YushuoEdge:
Now, name for hostname is not set when we use `runtime-rs` + `dragonball`. And there are also user name issues in the container. For example, when we enter the container with crictl exec, we will get:
```
$ crictl exec -it container_id /bin/bash
DEBU[0000] get runtime connection
DEBU[0000] ExecRequest: &ExecRequest{ContainerId:7ea30e061dc5b,Cmd:[/bin/bash],Tty:true,Stdin:true,Stdout:true,Stderr:false,}
DEBU[0000] ExecResponse: &ExecResponse{Url:http://127.0.0.1:42133/exec/QbwH-Rte,}
DEBU[0000] Exec URL: http://127.0.0.1:42133/exec/QbwH-Rte
DEBU[0000] StreamOptions: {0xc000012010 0xc000012018 0xc000012020 true <nil>}
groups: cannot find name for group ID 1000
I have no name!@clr-95e8f593545b466ea7883d33ef78a216:/$
I have no name!@clr-95e8f593545b466ea7883d33ef78a216:/$
I have no name!@clr-95e8f593545b466ea7883d33ef78a216:/$
```

Using nerdctl:
```
$ nerdctl run --rm -it --runtime io.containerd.kata.v2 centos
[root@clr-6471b5fe9eb04287bae0366771d08a3c /]# cat /etc/hostname
9c743a7d3f73
```

PR #6477, opened 2023-03-16
bugfix: set hostname in CreateSandboxRequest

YushuoEdge:
Setting hostname according to the spec.

The wrong username is because special systemd process is used as init process.

Fixes: #6247


studychao (approved):
LGTM

quanweiZhou (approved):
LGTM

Tim-0731-Hzt (approved):

Issue #6480, opened 2023-03-16
packaging: Add the ability to cache and install cached builds when building them using the Kata Deploy local builds


fidencio:
As the CI is slowly switching to using the Kata Deploy local build scripts to build each component, which helps to test exactly what we ship as part of our releases, we better start caching the artefacts and consuming those when it makes sense.

This has already been done as part of the  `CCv0` branch and we were able to get the build time goind down from 80~90 minutes to ~20, or even to > 10 minutes in the best case scenarios.

PR #6481, opened 2023-03-16
packaging / kata-deploy builds:  Add the ability to cache and consume cached components

fidencio:
As the CI is slowly switching to using the Kata Deploy local build scripts to build each component, which helps to test exactly what we ship as part of our releases, we better start caching the artefacts and consuming those when it makes sense.

This has already been done as part of the CCv0 branch and we were able to get the build time goind down from 80~90 minutes to ~20, or even to > 10 minutes in the best case scenarios.

Please, take a look at each invidivual commit for more details.

Fixes: #6480


fidencio:
Here you can find an example of the caching jobs already working:
* Cloud Hypervisor: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-clh-x86_64/1/
* Firecracker: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-firecracker-x86_64/2/
* Kernel:
  * Dragonball Experimental: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-kernel-dragonball-experimental-x86_64/1/
  * Vanilla: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-kernel-x86_64/1/
* Nydus: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-nydus-x86_64/2/
* QEMU: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-qemu--x86_64/2/
  * The QEMU jobs has an extra "-" between `qemu` and the `arch`, that's like that till this PR gets merged, as the wanted / proper name clashes with the current job used for caching the QEMU binary
* RootFS
  * Image: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-rootfs-image-x86_64/1/
  * Initrd: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-rootfs-initrd-x86_64/1/
* Shim v2: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-shim-v2-x86_64/1/
* VirtioFS: http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-virtiofsd-x86_64/2/

tools/packaging/static-build/cache_components_main.sh
stevenhorsman, line 1: Is the idea that that the `CCv0` version of this will be renamed `cache_components_CCv0.sh` to reflect that there are some differences? I guess there is a fair amount of duplicated code if this is the case, but we can probably review that as we start to push things from `CCv0` down into `main`.

tools/packaging/static-build/cache_components_main.sh
```diff
@@ -48,6 +49,19 @@ cache_qemu_artifacts() {
 	create_cache_asset "${qemu_tarball_name}" "${current_qemu_version}-${qemu_sha}" "${current_qemu_image}"
 }
 
+cache_rootfs_artifacts() {
+	local osbuilder_last_commit="$(get_last_modification "${repo_root_dir}/tools/osbuilder")"
```
stevenhorsman, line 53: In the install_image logic you have `sed s/-dirty//` as the rootfs artefacts are generated in the osbuilder directory. I think it is needed here as well to match them up?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -187,9 +187,19 @@ install_experimental_kernel() {
 
 # Install static qemu asset
 install_qemu() {
-	info "build static qemu"
 	export qemu_repo="$(yq r $versions_yaml assets.hypervisor.qemu.url)"
 	export qemu_version="$(yq r $versions_yaml assets.hypervisor.qemu.version)"
+
+	install_cached_tarball_component \
+		"QEMU" \
+		"${jenkins_url}/job/kata-containers-main-qemu-$(uname -m)/${cached_artifacts_path}" \
```
stevenhorsman, line 235: I noticed that there are jobs for both: 
http://jenkins.katacontainers.io/view/all/job/kata-containers-main-qemu--x86_64 and
http://jenkins.katacontainers.io/view/all/job/kata-containers-main-qemu-x86_64
Is the first one needed?

stevenhorsman (commented):
Generally looking good, I just had a few comments/questions:

tools/packaging/static-build/cache_components_main.sh
fidencio: Not really, the idea is to start moving the CCv0 content to this one.

The only reason I didn't leave this with the same name is to avoid complications for those doing the weekly merges, till the point we stop using the CCv0 branch.

Do you think this approach is reasonable?

tools/packaging/static-build/cache_components_main.sh
fidencio: Hmm.  I actually can remove the `-dirty` everywhere, as this is only needed for the case we generate the root hashes.
Thanks for spotting this!

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
fidencio: The idea once this one gets merged is to remove the latter completely, as it's only used by our "old cache mechanism".

Makes sense?

tools/packaging/static-build/cache_components_main.sh
stevenhorsman: That's fine - if you didn't want to make the tail wag the dog then I'm happy if we rename the CCv0 version instead before the next merge and change this to `cache_components.sh`?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
stevenhorsman: Will the `qemu--` job be renamed then. as this is pointing to the latter at the moment?

fidencio: Yep, that's exactly the idea.  I just didn't do it right now as the PR switching to using kata-deploy local-builds is not yet merged on the tests repo.  Once that one gets merged I'll remove the original and rename the qemu-- one.

tools/packaging/static-build/cache_components_main.sh
fidencio: The problem is that we need to adjust this in a few other places, plus change the jobs in Jenkins.
As time is a severe constraint right now, I'd rather not mess with this as this will be gone sooner than later (hopefully!).

stevenhorsman: Sure, that's totally fine!

stevenhorsman (approved):
LGTM. Thanks for the update and explanations!

tools/packaging/scripts/lib.sh
```diff
@@ -134,3 +134,44 @@ get_kernel_image_name() {
...
+	files="$(echo $files | tr ' ' '\n' | LC_ALL=C sort -u)"
+	# Concate the files and calculate a hash.
+	shasum="$(cat $files | sha256sum -b)" || true
+	if [ -n "$shasum" ];then
+		# Return only the SHA field.
+		echo $(awk '{ print $1 }' <<< $shasum)
+	fi
+}
+
+calc_qemu_files_sha256sum() {
```
wainersm, line 168: I think here we should include `static-build/qemu` too, shouldn't us?

It seems a dejavu... :)

fidencio: Not really.

We're differentiating two things here:
* The image that's used to build QEMU -- this is `static-build/qemu`, and this is covered by `get_last_modification`on that folder
* The QEMU patches and QEMU version themselves -- this is covered by the `calc_qemu_files_sha256sum`

In the end we end up using both information when deciding whether to build QEMU or not, but I don't see a particular reason to have those two combined here.

PR #6491, opened 2023-03-17
CC | kata-deploy-push more space

jepio:
Remove some tools from runner and some temp files during build so that we (hopefully) don't run out of disk space so often.

### DIFF
--- a/.github/workflows/kata-deploy-push.yaml
+++ b/.github/workflows/kata-deploy-push.yaml
@@ -72,6 +72,10 @@ jobs:
   make-kata-tarball:
     runs-on: ubuntu-latest
     steps:
+      - name: Free disk space
+        run: |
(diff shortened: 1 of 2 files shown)


stevenhorsman:
The non-TEE k8s test failed on:
```
not ok 12 [cc][agent][kubernetes][containerd] Test cannot pull an image from an authenticated registry without credentials
```
but the logs show it didn't ever get to doing the pull, so I've requeued to see what happens next time.

stevenhorsman (approved):
LGTM. Thanks!

stevenhorsman:
Given that only the tdx test is failing, which has a known issue, once we get a second review I think we can merge this.

Tim-0731-Hzt (approved):
Thanks @jepio LGTM

PR #6515, opened 2023-03-23
fix: bash syntax error

xyz-li:
Fix `return: cleanup_and_fail: numeric argument required`.


c3d (approved):
Looks good to me

fidencio (changes requested):
lgtm, thanks @xyz-li!

fidencio (approved):

PR #6523, opened 2023-03-24
gha: Add support for testing `kubernetes`

fidencio:


.github/workflows/kata-containers-ci-on-push.yaml
c3d, line 1: I love it that you are trying to break our current record for number of uses of the word "fixup!" in a commit message ✈️

PR #6528, opened 2023-03-27
arm/runtime: bring back pci shpc for qemu on arm

jongwu:
Currently, shpc is the only way for pci hotplug for qemu on Arm and there is no pci-acpi hotplug like x86 does for now.

We have disabled shpc drive in guest kernel which removes the 5s delay when perform shpc hotplug , but we may need pci bus rescan for device discovery explicitly. I don't do the recan op in a common path but only in "find_netlink" as I just encounter this case for now and may improve it later some way ( I have no idea for now).

I test this patch by using "nerdctl run --runtime io.containerd.kata.v2 --rm -it docker.io/library/ubuntu:latest bash". Without the first commit, network interface won't be hot added to kata and without the second commit, network device can't be found.

### DIFF
--- a/src/agent/src/netlink.rs
+++ b/src/agent/src/netlink.rs
@@ -16,6 +16,7 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::ops::Deref;
 use std::str::{self, FromStr};
 
+#[derive(Clone)]
 /// Search criteria to use when looking for a link in `find_link`.
 pub enum LinkFilter<'a> {
     /// Find by link name.
@@ -68,7 +69,9 @@ impl Handle {
         // target link. filter using name or family is supported, but
         // we cannot use that to find target link.
         // let's try if hardware address filter works. -_-
-        let link = self.find_link(LinkFilter::Address(&iface.hwAddr)).await?;
+        let link = self
+            .find_link_wrap(LinkFilter::Address(&iface.hwAddr))
+            .await?;
 
         // Bring down interface if it is UP
         if link.is_up() {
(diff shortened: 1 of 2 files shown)


jongwu:
Hello @dgibson -, Can I get some comments from you?

dgibson:
@jongwu I'm not really working on Kata any more.  I really don't recommend SHPC, that 5s delay is baked into the protocol.  It's basically fundamentally unsuitable for virtualized hardware. By bypassing SHPC by removing the module, you're not really doing SHPC any more, you're doing an ad-hoc hotplug by manual rescan which may or may not work.

Performing the manual PCI rescan is also pretty dangerous - the agent doesn't really have any good way of knowing when qemu is sufficiently complete in the hotplug process for it to actually pick up the new devices.  I've also found in practice that races between that rescan and other PCI operations during boot can cause a number of nasty side effects, up to and including making a device entirely unusable until a *host* reboot (only with passed through host devices, obviously).

I have no good solutions for you here.  Honestly I think to make this work in any way reliably you're going to need to look at the kernel and find, fix, or implement a hot-plug protocol which actually works without delays that Kata can't accommodate.

jongwu:
I'm really appreciate with your explaination. Yeah, SHPC is harmfull and it's better to discard it in virtualization senario.

It's correct way to create a new protocal or fix it in kernel, however, is difficult and takes time. I will find other way to resolve the pci hotplug issue on Arm/kata. 

Thanks! @dgibson

dgibson:
Good luck.  I suspect your best options are to see if the support for either ACPI or PCI-E hotplug can be gotten working on arm.

jongwu:
I hope someone can enable ACPI PCI hotplug in qemu part. For now I can only use pcie-root-port to do the PCI device hotplug (I assume that there is no legacy PCI device that can't hotplugged into pcie-root-port in kata) thus make pcie-root-port instead of pci-bridge as the default bridge in kata on Arm. Hope it works.

Issue #6529, opened 2023-03-27
CC | QEMU TDX jobs are constantly failing


fidencio:
@arronwy figured out the failures are happening as dcap, one dependency of rats-tls, has been updated and the update is incompatible with the host OS used as part of the CI jobs.

We need to ensure a known to be working version of dcap (1.15) is used instead.

PR #6532, opened 2023-03-27
rootfs: Ensure we use a compatible version of dcap with rats-tls

fidencio:
DCAP has upgraded to 1.16, which is not compatible with the host OS used as part of our CI (2022ww44).  Let's ensure DCAP 1.15 is used instead.

Fixes: #6529


fidencio:
Cloud Hypervisor TDX passed but didn't report back, again: http://10.112.240.208:8080/job/kata-containers-CCv0-cloud-hypervisor-TDX-CRI_CONTAINERD_K8S-PR/803/

Issue #6582, opened 2023-04-03
Nydus prefetch_files feature needed.


Apokleos:
**Is your feature request related to a problem? Please describe.**
In general, when a container reads data, Nydus needs to request image data from the image center. 
However, sometimes it is necessary to speed up the data request efficiency, such as when Kata pod starting, it is necessary to ensure fast access to image base data as much as possible. 
Nydus itself provides a prefetch mechanism, which can prefetch data and put it in the cache. When the container needs to read it, it can directly read the data from the cache. 

Currently, however, the related implementation of Nydus does not support this feature.

So we need to  enable this feature for kata/nydus.

**Describe the solution you'd like**

PR #6585, opened 2023-04-03
nydus_rootfs/prefetch_files: add prefetch_files for RAFS

Apokleos:
nydus_rootfs/prefetch_files: add prefetch_files for RAFS

A sandbox annotation used to specify prefetch_files.list path the container image being used, and runtime will pass
it to Hypervisor to search for corresponding prefetch file. format looks like:
"io.katacontainers.config.hypervisor.prefetch_files.list" = /path/to/<uid>/xyz.com/fedora:36/prefetch_file.list

Fixes: #6582

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/mod.rs
+++ b/src/libs/kata-types/src/config/hypervisor/mod.rs
@@ -979,6 +979,13 @@ pub struct Hypervisor {
     #[serde(default, flatten)]
     pub shared_fs: SharedFsInfo,
 
+    /// A sandbox annotation used to specify prefetch_files.list host path container image
+    /// being used, and runtime will pass it to Hypervisor to  search for corresponding
+    /// prefetch list file:
+    ///   prefetch_list_path = /path/to/<uid>/xyz.com/fedora:36/prefetch_file.list
+    #[serde(default)]
+    pub prefetch_list_path: String,
+
     /// Vendor customized runtime configuration.
     #[serde(default, flatten)]
     pub vendor: HypervisorVendor,
@@ -1022,6 +1029,10 @@ impl ConfigOps for Hypervisor {
                 hv.network_info.adjust_config()?;
                 hv.security_info.adjust_config()?;
                 hv.shared_fs.adjust_config()?;
+                resolve_path!(
+                    hv.prefetch_list_path,
+                    "prefetch_list_path `{}` is invalid: {}"
+                )?;
             } else {
                 return Err(eother!("Can not find plugin for hypervisor {}", hypervisor));
             }
@@ -1056,6 +1067,10 @@ impl ConfigOps for Hypervisor {
                     "Hypervisor control executable `{}` is invalid: {}"
                 )?;
                 validate_path!(hv.jailer_path, "Hypervisor jailer path `{}` is invalid: {}")?;
+                validate_path!(
+                    hv.prefetch_list_path,
+                    "prefetch_files.list path `{}` is invalid: {}"
+                )?;
             } else {
                 return Err(eother!("Can not find plugin for hypervisor {}", hypervisor));
             }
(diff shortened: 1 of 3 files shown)


src/runtime-rs/crates/resource/src/rootfs/nydus_rootfs.rs
```diff
@@ -151,3 +160,75 @@ impl Rootfs for NydusRootfs {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs::{self, File};
+
+    #[tokio::test]
+    async fn test_get_nydus_prefetch_files() {
+        let nydus_image_key = "xxx001".to_string();
+        let nydus_prefetch_path = "/tmp/nudus001/ooox01".to_string();
```
liubin, line 212: It's better to use `tempfile::tempdir` to save test content.

src/runtime-rs/crates/resource/src/rootfs/nydus_rootfs.rs
```diff
@@ -151,3 +160,75 @@ impl Rootfs for NydusRootfs {
...
+    if !prefetch_list_path.is_file() {
+        debug!(
+            sl!(),
+            "nydus prefetch list file {:?} not a regular file", &prefetch_list_path
+        );
+
+        return None;
+    }
+
+    return Some(prefetch_list_path.to_str().unwrap().to_string());
```
liubin, line 201: Avoid using `unwrap()`, using `and_then` of an Option object may be ok.

src/runtime-rs/crates/resource/src/rootfs/nydus_rootfs.rs
Apokleos: Thx, fixed

src/runtime-rs/crates/resource/src/rootfs/nydus_rootfs.rs
Apokleos: > Wait, why is the option object necessary to start with?

Thx, the Option Object chosen as the result of the annotation sometimes maybe empty. And we need ensure the `prefetch_files_list of None` case be handled correctly.

src/runtime-rs/crates/resource/src/rootfs/nydus_rootfs.rs
```diff
@@ -151,3 +156,67 @@ impl Rootfs for NydusRootfs {
...
+// As the result of caller `rafs_mount`, it returns `Option<String>`.
+async fn get_nydus_prefetch_files(nydus_prefetch_path: String) -> Option<String> {
+    // nydus_prefetch_path is an annotation and pod with it will indicate
+    // that prefetch_files will be included.
+    if nydus_prefetch_path.is_empty() {
+        info!(sl!(), "nydus prefetch files path not set, just skip it.");
+
+        return None;
+    }
+
```
lifupan, line 170: It's better to deal with the path's symbol link here.

liubin (approved):

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -979,6 +979,13 @@ pub struct Hypervisor {
     #[serde(default, flatten)]
     pub shared_fs: SharedFsInfo,
 
+    /// A sandbox annotation used to specify prefetch_files.list host path container image
+    /// being used, and runtime will pass it to Hypervisor to  search for corresponding
+    /// prefetch list file:
+    ///   prefetch_list_path = /path/to/<uid>/xyz.com/fedora:36/prefetch_file.list
+    #[serde(default)]
+    pub prefetch_list_path: String,
```
Tim-0731-Hzt, line 987: Can we configure this path via `configuration.toml` ?

Apokleos: Thanks.
It will be updated into configuration, but generally it will not be directly written in the configuration because the prefetch_files path stored for each pod is different.
So, we can only do this through annotation to specify the prefetch files path.

Issue #6588, opened 2023-04-03
tools: Avoid building the kernel twice


zvonkok:
`static-build/kernel/build.sh` runs two kernel build commands `build` and `install`

```
sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
	-w "${PWD}" \
	"${container_image}" \
	bash -c "${kernel_builder} $* build"

sudo docker run --rm -i -v "${repo_root_dir}:${repo_root_dir}" \
	-w "${PWD}" \
	--env DESTDIR="${DESTDIR}" --env PREFIX="${PREFIX}" \
	"${container_image}" \
	bash -c "${kernel_builder} $* install"
```

This translates to `build-kernel.sh <arguments> build` and `build-kernel.sh <arguments> install`

```
case "${subcmd}" in
		build)
			build_kernel "${kernel_path}"
			;;
		install)
			build_kernel "${kernel_path}"
			install_kata "${kernel_path}"
			;;
		setup)
			setup_kernel "${kernel_path}"
			[ -d "${kernel_path}" ] || die "${kernel_path} does not exist"
			echo "Kernel source ready: ${kernel_path} "
			;;
		*)
			usage 1
			;;
```

This results in building the kernel twice

PR #6590, opened 2023-04-03
tools: Avoid building the kernel twice

zvonkok:
Fixes: #6588

`install_kata` does not set arch so setting it like in `build_kernel`


fidencio (approved):
lgtm, thanks @zvonkok!

cmaf (approved):
thanks @zvonkok

zvonkok:
@cmaf Done.

amshinde (approved):

Issue #6592, opened 2023-04-03
CI | GHA | kernel build check fails when trying to push image on quay


UnmeshDeodhar:
Kernel build check fails when trying to push image to quay:
https://github.com/kata-containers/kata-containers/actions/runs/4600815573/jobs/8127897852?pr=6591
It errors with `unauthorized: access to the requested resource is not authorized`

As per discussion with @fidencio on kata-dev slack this is not an expected behaviour.

PR #6596, opened 2023-04-04
gha: Only push images to registry after merging a PR

fidencio:
56331bd7bcfbc6aabd8a732a10c544ba3df84622 oversaw the fact that we mistakenly tried to push the build containers to the registry for a PR, rather than doing so only when the code is merged.

As the workflow is now shared between different actions, let's introduce an input variable to specify which are the cases we actually need to perform a push to the registry.

Fixes: #6592

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -9,6 +9,10 @@ on:
       tarball-suffix:
         required: false
         type: string
+      push-to-registry:
+        required: false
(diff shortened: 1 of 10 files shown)


wainersm (approved):
The only doubt I had was whether `head_sha` is the repo's HEAD sha or pull request's sha. I talked in private with @fidencio and it is really the later case, i.e., `head_sha` is inherited from the "Commit Message Check" job.

Good work @fidencio !

GabyCT (approved):

Issue #6600, opened 2023-04-04
gha: add cc-snp-qemu asset to amdx86_64 payload builder


Alex-Carter01:
Last SNP component needed in the x86 payload is Qemu. I missed adding an entry in #6062.
There are already entries for the shim and x86 ovmf build, so after this all of the base SNP components will be in the CI bundles.

PR #6062, opened 2023-01-12
kata-deploy: SNP version of Qemu

Alex-Carter01:
An SNP runtimeclass needs a build of Qemu from https://github.com/AMDESE/qemu/tree/snp-v3.

So a new target needs to be added to add it to a kata-deploy bundle.

Building requires a qemu no_patches file

The qemu builders work simplest with a tag rather than a branch and commit hash, so I made a fork of the AMDESE repo, but we could alternatively use a qemu fork in the CoCo group or the confidential-containers-demo already has a qemu repo.

Fixes: #6061
Signed-Off-By: Alex Carter <alex.carter@ibm.com>

PR #6602, opened 2023-04-04
CC | gha: add SNP components to amdx86_64 payload workflows

Alex-Carter01:
Last SNP component needed in the x86 payload is Qemu. I missed adding an entry in #6062.

Fixes: #6600

### DIFF
--- a/.github/workflows/cc-payload-after-push-amd64.yaml
+++ b/.github/workflows/cc-payload-after-push-amd64.yaml
@@ -20,6 +20,7 @@ jobs:
           - cc-sev-kernel
           - cc-sev-ovmf
           - cc-x86_64-ovmf
+          - cc-snp-qemu
           - cc-sev-rootfs-initrd
           - cc-tdx-kernel
           - cc-tdx-rootfs-image
--- a/.github/workflows/cc-payload-amd64.yaml
+++ b/.github/workflows/cc-payload-amd64.yaml
@@ -19,6 +19,8 @@ jobs:
           - cc-virtiofsd
           - cc-sev-kernel
           - cc-sev-ovmf
+          - cc-x86_64-ovmf
+          - cc-snp-qemu
           - cc-sev-rootfs-initrd
           - cc-tdx-kernel
           - cc-tdx-rootfs-image


jepio (approved):

fidencio:
@Alex-Carter01,

Here's a suggestion on how to improve your commit message:
````
gha: Add cc-snp-qemu asset to the payload builder

The last SNP component needed in the x86 payload is QEMU, but I missed adding an entry
for it as part of https://github.com/kata-containers/kata-containers/pull/6062.

Fixes: #6600

Signed-Off-By: Alex Carter <AlexCarter@ibm.com>
````

File: .github/workflows/cc-payload-after-push-amd64.yaml [hunk 1]
```diff
@@ -20,6 +20,7 @@ jobs:
           - cc-sev-kernel
           - cc-sev-ovmf
           - cc-x86_64-ovmf
+          - cc-snp-qemu
```
fidencio, line 23: This will make `snp-qemu` part of the payload generated after every push, and I think this is **also** something that you want.

However, don't you want to also add this as part of the release?  If so, https://github.com/kata-containers/kata-containers/blob/CCv0/.github/workflows/cc-payload-amd64.yaml is where you should be looking at.

fidencio (changes requested):
I've left some comments.

Also, please, when targetting the CCv0 branch add the  `CC | ` prefix (or similar).

.github/workflows/cc-payload-after-push-amd64.yaml
Alex-Carter01: Oh, I hadn't seen there was a separate release workflow, so I will add that to this PR for qemu, and will also add the x86_64 ovmf build that SNP needs

Alex-Carter01:
Also adding to the release workflow. Also added the ovmf piece.

fidencio:
> Also adding to the release workflow. Also added the ovmf piece.

Cool. Could you change the commit message to?

```
gha: Add SNP components to the cc-payload workflows

Adding SNP components needed to the x86 payload push and release payloads.

QEMU is needed in both the after-push payload and release payload, while OVMF is only
missing from the release workflow.

Fixes: #6600

Signed-Off-By: Alex Carter <AlexCarter@ibm.com>
```

fidencio (approved):
lgtm, thanks @Alex-Carter01!

fidencio:
s390x failure happened on a unit test that's known to be flaky and is not related to this PR at all.

Issue #6606, opened 2023-04-05
CI | GHA | Use a D4s_v5 instance type


fidencio:
We're currently using D4s_v3, but we can bump to using D4s_v5 as it's machine with newer CPUs and with the very same costs.

Thanks @jepio  for the tip.

PR #6607, opened 2023-04-05
gha: aks: Use D4s_v5 instance

fidencio:
It's been pointed out that D4s_v5 instances are more powerful than the D4s_v3 ones, and have the very same price.  With this in mind, let's switch to the newer machines.

Fixes: #6606


GabyCT (approved):

wainersm (approved):
Yeah, If I read the docs correctly they have same amount of vCPU and Memory

Issue #6611, opened 2023-04-06
gha | CI | k8s-on-aks | The name of the AKS cluster is not valid


fidencio:
This happened as we were depending on `pr_number` when running on `pull_request`, however this is not easily available when running on `workflow_run`, and I ended up not seeing this part also needed to be changed when switching to using `workflow_run`.

PR #6612, opened 2023-04-06
gha: k8s-on-aks: Fix cluster name

fidencio:
This was missed from the last series, as GHA will use the "target branch" yaml file to start the workflow.

Basically we changed the name of the cluster created to stop relying on the PR number, as that's not easily accessible on `workflow_run`.

Fixes: #6611


BbolroC (approved):
lgtm, thanks!

gkurz (approved):
LGTM. Thanks @fidencio !

studychao (approved):
Thanks!

PR #5964, opened 2022-12-30
kata-runtime: add rust runtime path for kata-runtime exec

Tim-0731-Hzt:
add rust runtime path for kata-runtime exec

Fixes:#5963
Signed-off-by: Zhongtao Hu <zhongtaohu.tim@linux.alibaba.com>

PR #6623, opened 2023-04-06
runtime: Don't create socket file in /run/kata

### LINKED ISSUE #6622: runtime: No space left on device for /run folder
# Description of problem

After creating and destroying thousands of containers in a loop, the /run folder will have no inodes available and creation of new containers will generate a 'no space left on device' error.

alex-matei:
The socket file for shim management is created in /run/kata and it 
isn't deleted after the container is stopped. After running and stopping 
thousands of containers /run folder will run out of space.

Fixes #6622


src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -312,9 +312,15 @@ func GetSandboxesStoragePathRust() string {
 	return "/run/kata"
 }
 
-// SocketAddress returns the address of the unix domain socket for communicating with the
+// ServerSocketAddress returnes the address of the unix domain socket the shim management endpoint
```
gkurz, line 315: Hi @alex-matei !

Thanks for addressing this annoying issue ! A few comments below.

I had hard times to understand how the address could differ between server and client, which doesn't quite make sense in the first place. This all boils down to #5964 : the original `kata-runtime` in go is still being used with runtime-rs. This requires the shim management code to be able to provide "/run/kata" as the path to be used when a go client wants to connect to the rust runtime. A few comments to make this clear would definitely help.

Also, this is hardcoding the creation of the socket path in yet another place. What about consolidating this a bit ?

```suggestion
-// ServerSocketAddress returnes the address of the unix domain socket the shim management endpoint
-// should listen
+// SocketPath returns the path of the socket using the given storagePath
+func SocketPath(id string, storagePath string) string {
+       return filepath.Join(string(filepath.Separator), storagePath, id, "shim-monitor.sock")
+}
+
+// SocketPath returns the path of the socket to be used with the go runtime
+func SocketPathGo(id string) string {
+       return SocketPath(id, GetSandboxesStoragePath())
+}
+
+// SocketPath returns the path of the socket to be used with the rust runtime
+func SocketPathRust(id string) string {
+       return SocketPath(id, GetSandboxesStoragePathRust())
+}
+
+// ServerSocketAddress returns the address of the unix domain socket the shim management endpoint
+// should listen.
+// NOTE: this code is only called by the go shim management implementation.
 func ServerSocketAddress(id string) string {
-       return fmt.Sprintf("unix://%s", filepath.Join(string(filepath.Separator), GetSandboxesStoragePath(), id, "shim-monitor.sock"))
+       return fmt.Sprintf("unix://%s", SocketPathGo(id))
 }
 
 // ClientSocketAddress returns the address of the unix domain socket for communicating with the
 // shim management endpoint
+// NOTE: this code allows various go clients, e.g. kata-runtime or kata-monitor commands, to
+// connect to the rust shim management implementation.
 func ClientSocketAddress(id string) string {
        // get the go runtime uds path
-       socketPath := filepath.Join(string(filepath.Separator), GetSandboxesStoragePath(), id, "shim-monitor.sock")
+       socketPath := SocketPathGo(id)
        // if the path not exist, use the rust runtime uds path instead
        if _, err := os.Stat(socketPath); err != nil {
-               return fmt.Sprintf("unix://%s", filepath.Join(string(filepath.Separator), GetSandboxesStoragePathRust(), id, "shim-monitor.sock"))
+               socketPath = SocketPathRust(id)
        }
        return fmt.Sprintf("unix://%s", socketPath)
 }
```

alex-matei: Thanks @gkurz, it looks good, definitely easier to follow. I shall commit you changes and squash them in one commit.

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -312,7 +312,46 @@ func GetSandboxesStoragePathRust() string {
 	return "/run/kata"
 }
 
-// ServerSocketAddress returnes the address of the unix domain socket the shim management endpoint
+-// ServerSocketAddress returnes the address of the unix domain socket the shim management endpoint
```
gkurz, line 315: Hmm... something went wrong. All lines now start with `+` sign...

alex-matei: Fixed, should be ok now.

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -312,14 +312,38 @@ func GetSandboxesStoragePathRust() string {
 	return "/run/kata"
 }
 
-// SocketAddress returns the address of the unix domain socket for communicating with the
+// SocketPath returns the path of the socket using the given storagePath
+func SocketPath(id string, storagePath string) string {
+	return filepath.Join(string(filepath.Separator), storagePath, id, "shim-monitor.sock")
+}
+
+// SocketPath returns the path of the socket to be used with the go runtime
```
gkurz, line 320: Oops... my bad. Please s/SocketPath/SocketPathGo.

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -312,14 +312,38 @@ func GetSandboxesStoragePathRust() string {
...
+func SocketPathGo(id string) string {
+	return SocketPath(id, GetSandboxesStoragePath())
+}
+
+// SocketPath returns the path of the socket to be used with the rust runtime
```
gkurz, line 325: and s/SocketPath/SocketPathRust/ here. And then we should be good ! :smile:

src/runtime/pkg/containerd-shim-v2/shim_management.go
alex-matei: Done.

src/runtime/pkg/containerd-shim-v2/shim_management.go
alex-matei: Done.

gkurz (approved):
Great ! Thanks @alex-matei !

Would you be interested into going all the way through by backporting this patch to the stable-3.1 branch ?

alex-matei:
@gkurz Yes, sure. I'll create a backport PR.

bergwolf (approved):
lgtm!

The CI failure is unrelated to the change here but because of a newly added build target `qemu-tdx-experimental-tarball`. Could you rebased the PR on top of the latest main to get rid of the failure? Thanks!

Issue #6628, opened 2023-04-06
CI | gha | k8s-on-aks | Set {create,delete}_aks as steps


fidencio:
We've been currently using {create,delete}_aks as jobs.  However, it
means that if the tests fail we'll end up deleting the AKS cluster (as
expected), but not having a way to recreate the cluster without
re-running all jobs, which is a waste of resources.

PR #6629, opened 2023-04-06
gha: k8s-on-aks: Set {create,delete}_aks as steps

fidencio:
We've been currently using {create,delete}_aks as jobs.  However, it means that if the tests fail we'll end up deleting the AKS cluster (as expected), but not having a way to recreate the cluster without re-running all jobs, which is a waste of resources.

Fixes: #6628

### DIFF
--- a/.github/workflows/ci-on-push.yaml
+++ b/.github/workflows/ci-on-push.yaml
@@ -1,14 +1,9 @@
 name: Kata Containers CI
 on:
-  workflow_run:
-    workflows:
-      - Commit Message Check
-    types:
-      - completed
+  pull_request_target:
 
 jobs:
   build-kata-static-tarball-amd64:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-on-push.yaml
```diff
@@ -1,6 +1,8 @@
 name: Kata Containers CI
 on:
   pull_request_target:
+    branches:
```
wainersm, line 4: Good. On another PR that I reviewed I was about to ask you why not set this property, but I ended up not asking because all this stuffs is already too confusing :)

fidencio: You should have done.  I didn't do simply because I forgot to. :-/

fengwang666 (approved):

PR #6637, opened 2023-04-07
rustjail: Use CPUWeight with systemd and CgroupsV2

### LINKED ISSUE #6636: Cannot create pods when guest is systemd+CgroupsV2
Creation of a pod requesting 2 CPU shares fails when the guest is running with systemd in CgroupsV2 mode :

```
[test@virtlab702 ~]$ sudo crictl runp -r kata -T 1000s pod.json 
E0407 09:28:06.462216 1441654 remote_runtime.go:222] "RunPodSandbox from runtime service failed" err=<
	rpc error: code = Unknown desc = CreateContainer failed: failed to set unit properties crio-9b7e658bd775b751846527ea0f9f78eddcec4a3ac8d51628b3e275874d9c177c.scope
	
	Caused by:
	    org.freedesktop.DBus.Error.InvalidArgs: Value specified in CPUShares is out of range: unknown
```

According to https://github.com/containers/crun/blob/main/crun.1.md#cgroup-v2 , CPU shares are superseded by CPU weight.

The agent indeed does the calculation to convert from shares to weight :

```
// ref: https://github.com/containers/crun/blob/main/crun.1.md#cgroup-v2
// [2-262144] to [1-10000]
fn get_unified_cpushares(shares: u64) -> u64 {
    if shares == 0 {
        return 100;
    }

    1 + ((shares - 2) * 9999) / 262142
}
```

and gets 1 as expected, but it then passes it to systemd as `CPUShares` instead of `CPUWeight` :

```
    // v2:
    // cpu.shares <-> CPUShares
    // cpu.period <-> CPUQuotaPeriodUSec
    // cpu.period & cpu.quota <-> CPUQuotaPerSecUSec
    fn unified_apply(
        cpu_resources: &LinuxCpu,
        properties: &mut Properties,
        systemd_version: &str,
    ) -> Result<()> {
        if let Some(shares) = cpu_resources.shares {
            let unified_shares = get_unified_cpushares(shares);
            properties.push(("CPUShares", Value::U64(unified_shares)));
        }
```

Systemd then rejects the value because CPU shares should belong to [2-262144].

gkurz:
The CPU shares property belongs to CgroupsV1. CgroupsV2 uses CPU weight instead. The correct value is computed in the latter case but it is passed to systemd using the legacy property. Systemd rejects the request and the agent exists with the following error :

        Value specified in CPUShares is out of range: unknown

Replace the "shares" wording with "weight" in the CgroupsV2 code to avoid confusions. Use the "CPUWeight" property since this is what systemd expects in this case.

Fixes #6636

References:

https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#CPUWeight=weight https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#systemd%20252 https://github.com/containers/crun/blob/main/crun.1.md#cpu-controller

### DIFF
--- a/src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
+++ b/src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
@@ -71,7 +71,7 @@ impl Cpu {
     }
 
     // v2:
-    // cpu.shares <-> CPUShares
+    // cpu.shares <-> CPUWeight
     // cpu.period <-> CPUQuotaPeriodUSec
     // cpu.period & cpu.quota <-> CPUQuotaPerSecUSec
     fn unified_apply(
@@ -80,8 +80,8 @@ impl Cpu {
         systemd_version: &str,
     ) -> Result<()> {
         if let Some(shares) = cpu_resources.shares {
-            let unified_shares = get_unified_cpushares(shares);
-            properties.push(("CPUShares", Value::U64(unified_shares)));
+            let weight = shares_to_weight(shares);
+            properties.push(("CPUWeight", Value::U64(weight)));
         }
 
         if let Some(period) = cpu_resources.period {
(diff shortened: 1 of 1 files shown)


littlejawa (approved):
/lgtm
Thanks @gkurz

fidencio (approved):
lgtm, thanks @gkurz!

Issue #4693, opened 2022-07-20
runtime-rs: support nerdctl net


quanweiZhou:
nredctl + runc can load bridge net
```
# nerdctl run --rm -it --runtime io.containerd.runc.v2 --net bridge busybox ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0@if8: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue
    link/ether f6:66:9d:23:b6:21 brd ff:ff:ff:ff:ff:ff
    inet 10.4.0.6/24 brd 10.4.0.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::f466:9dff:fe23:b621/64 scope link
       valid_lft forever preferred_lft forever
```

nredctl + runc can't load bridge net
go runtime PR: #3670
```
# nerdctl run --rm -it --runtime io.containerd.kata.v2 --net bridge busybox sh
/ # ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
```

PR #6639, opened 2023-04-09
runtime-rs: enable nerdctl to setup cni plugin

Tim-0731-Hzt:
1. when we use nerdctl to setup network for kata, no netns is created by
  nerdctl, kata need to create netns by its own
  
2. after start VM, nerdctl will call cni plugin via oci hook, we need to
rescan the netns after the interfaces have been created, and hotplug
the network device into the VM
  
Fixes:#4693


src/libs/oci/src/lib.rs
```diff
@@ -192,7 +192,10 @@ pub struct Hook {
 pub struct Hooks {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub prestart: Vec<Hook>,
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    #[serde(
+        rename(deserialize = "createRuntime"),
```
YushuoEdge, line 196: The following create_container, start_container may also need ‘rename(...)’

src/runtime-rs/crates/resource/src/manager.rs
```diff
@@ -50,9 +50,12 @@ impl ResourceManager {
...
     }
 
-    pub async fn prepare_before_start_vm(&self, device_configs: Vec<ResourceConfig>) -> Result<()> {
+    pub async fn prepare_before_or_after_start_vm(
+        &self,
+        device_configs: Vec<ResourceConfig>,
+    ) -> Result<()> {
         let mut inner = self.inner.write().await;
-        inner.prepare_before_start_vm(device_configs).await
+        inner.prepare_before_or_after_start_vm(device_configs).await
```
YushuoEdge, line 58: I think there is no need to change the name.

src/runtime-rs/crates/resource/src/network/network_with_netns.rs
```diff
@@ -120,6 +124,21 @@ impl Network for NetworkWithNetns {
...
+        {
+            let _netns_guard =
+                netns::NetnsGuard::new(&inner.netns_path).context("net netns guard")?;
+            for e in &inner.entity_list {
+                e.endpoint.detach(h).await.context("detach")?;
+            }
+        }
+        let netns = get_from_path(inner.netns_path.clone())?;
+        netns.remove()?;
+        fs::remove_dir_all(inner.netns_path.clone()).context("failed to remove netns path")?;
```
YushuoEdge, line 139: I wonder why do we need to remove the dir again after calling netns.remove()?

src/runtime-rs/crates/runtimes/common/src/sandbox.rs
```diff
@@ -12,6 +12,7 @@ pub trait Sandbox: Send + Sync {
     async fn start(
         &self,
         netns: Option<String>,
+        network_created: bool,
```
YushuoEdge, line 15: Can we use some check function instead of passing this parameter?

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -129,7 +114,39 @@ impl RuntimeHandlerManagerInner {
...
+                        continue;
+                    }
+
+                    if !ns.path.is_empty() {
+                        netns_path = Some(ns.path.clone());
+                        break;
+                    }
+                }
+            }
+            // if we get empty netns from oci spec, we need to create netns for the VM
```
YushuoEdge, line 137: Ditto. I think using some check function and doing similar operation later when we handle 'netns' is more appropriate. Because I think parameters in sandbox.start() are very important, we should add with restraint as much as possible. Otherwise it would be very easy to introduce the clippy hint of too_many_arugument.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -188,15 +196,33 @@ impl Sandbox for VirtSandbox {
         info!(sl!(), "start vm");
 
         // execute pre-start hook functions, including Prestart Hooks and CreateRuntime Hooks
-        let (prestart_hooks, create_runtime_hooks) = match spec.hooks.as_ref() {
-            Some(hooks) => (hooks.prestart.clone(), hooks.create_runtime.clone()),
-            None => (Vec::new(), Vec::new()),
+        let (prestart_hooks, create_runtime_hooks, has_prestart_hooks) = match spec.hooks.as_ref() {
```
YushuoEdge, line 199: I am not sure the parameter has_prestart_hooks is correct here. Sometimes the hooks may only contain start_container hooks, for example, which is not prestart_hook. And I following operation seems to be strongly related to create_runtime hooks.

YushuoEdge (commented):
Thanks @Tim-0731-Hzt ! Some comments below:

src/runtime-rs/crates/resource/src/network/network_with_netns.rs
Tim-0731-Hzt: because `netns.remove()` will only do the umount, the netns path won't be removed

src/runtime-rs/crates/resource/src/manager.rs
Tim-0731-Hzt: After the VM starts, I think it's inappropriate to call it `prepare_before_start_vm`

src/runtime-rs/crates/runtimes/common/src/sandbox.rs
Tim-0731-Hzt: As we need to determine if the `netns` is created by kata or contained, the best way is to do this is to check the existence of `netns` in OCI spec. If `netns` is None, kata will create it and set `network_created` to true, otherwise, to None, so I think no more check functions are needed.

src/runtime-rs/crates/runtimes/src/manager.rs
Tim-0731-Hzt: How about put `netns `and `network_created` into a struct `network_config`, and passing it to `start`. Because it's hard to check if the network is created in `start()`.

src/libs/oci/src/lib.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -195,8 +203,27 @@ impl Sandbox for VirtSandbox {
         self.execute_oci_hook_functions(&prestart_hooks, &create_runtime_hooks, state)
             .await?;
 
-        // TODO: if prestart_hooks is not empty, rescan the network endpoints(rely on hotplug endpoints).
-        // see: https://github.com/kata-containers/kata-containers/issues/6378
+        let has_prestart_hooks = !prestart_hooks.is_empty() || !create_runtime_hooks.is_empty();
```
quanweiZhou, line 206: Suggest wrapping into the function `process_after_start_vm`.

src/runtime-rs/crates/resource/src/network/network_with_netns.rs
```diff
@@ -120,6 +124,21 @@ impl Network for NetworkWithNetns {
         }
         Some(endpoint)
     }
+
+    async fn remove(&self, h: &dyn Hypervisor) -> Result<()> {
+        let inner = self.inner.read().await;
```
quanweiZhou, line 129: For the netns not created by Kata, I think no need to remove them.

quanweiZhou (commented):
Thanks, @Tim-0731-Hzt few comments

src/runtime-rs/crates/resource/src/network/utils/netns.rs
```diff
@@ -50,6 +51,20 @@ impl Drop for NetnsGuard {
     }
 }
 
+// generate the network namespace name
+pub fn generate_netns_name() -> String {
```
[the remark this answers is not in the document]
Tim-0731-Hzt: This is hard to do, as the name generate is random.

src/runtime-rs/crates/resource/src/network/network_with_netns.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/resource/src/manager.rs
Tim-0731-Hzt: fixed

quanweiZhou: The function name `prepare_before_or_after_start_vm` is not updated.

src/runtime-rs/crates/runtimes/src/manager.rs
Tim-0731-Hzt: I added `SandboxNetworkEnv `struct to store them

quanweiZhou (approved):

src/runtime-rs/crates/runtimes/common/src/sandbox.rs
```diff
@@ -7,14 +7,20 @@
 use anyhow::Result;
 use async_trait::async_trait;
 
+#[derive(Clone)]
+pub struct SandboxNetWorkEnv {
```
liubin, line 11: Nit: `SandboxNetWorkEnv` or `SandboxNetworkEnv`

liubin (approved):

src/runtime-rs/crates/runtimes/common/src/sandbox.rs
Tim-0731-Hzt: fixed

PR #6641, opened 2023-04-11
Backport cgroup fixes to 3.1

gkurz:
Backport #6562 and #6637. This is needed to use cgroups v2 with systemd in the guest.


jepio (approved):
LGTM, thanks.

Issue #6643, opened 2023-04-12
agent: support drop cache


xyc-grace:
When attempting to remove containers, it's possible to encounter an 'EBUSY' error due to unreleased guest file descriptors. These unreleased descriptors prevent the host VirtIO-FS file descriptor from being released, resulting in the 'EBUSY' error. This issue can be particularly common in production environments.

xyc-grace:
The 'EBUSY' problem has not yet occurred in the kata community, and its root cause can be traced to VirtIO-FS. Therefore, it's best to monitor whether it arises in kata and wait for the issue to be resolved in VirtIO-FS.

PR #6644, opened 2023-04-12
agent: support drop cache

xyc-grace:
This PR adds a 'drop_cache' api to the agent, which releases the unreleased file descriptors and resolves the 'EBUSY' error when removing containers.

Fixes: #6643

### DIFF
--- a/src/agent/src/util.rs
+++ b/src/agent/src/util.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use futures::StreamExt;
 use std::io;
 use std::io::ErrorKind;
@@ -15,6 +15,7 @@ use tracing::instrument;
 
 // Size of I/O read buffer
 const BUF_SIZE: usize = 8192;
+const DROP_CACHE_FILE: &str = "/proc/sys/vm/drop_caches";
 
 // Interruptable I/O copy using readers and writers
 // (an interruptable version of "io::copy()").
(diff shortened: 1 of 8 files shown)


src/agent/src/util.rs
```diff
@@ -72,6 +73,18 @@ pub async fn get_vsock_stream(fd: RawFd) -> Result<VsockStream> {
     Ok(stream?)
 }
 
+// Drop cache in guest
+// drop value should be in {1, 2, 3}
```
YushuoEdge, line 77: Could you add some comment to explain what the meaning of this value?

YushuoEdge (commented):
Thanks @xyc-grace !  Just one small problem：

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs
```diff
@@ -183,6 +184,17 @@ impl ContainerInner {
                 }
             })?;
 
+        self.agent
+            .drop_cache(DropCacheRequest {
+                drop_option: 2,
+                ..Default::default()
```
Apokleos, line 190: Thx @xyc-grace 
Some comments about the reason of  drop_option(2) is needed here.

src/agent/src/util.rs
```diff
@@ -72,6 +73,22 @@ pub async fn get_vsock_stream(fd: RawFd) -> Result<VsockStream> {
     Ok(stream?)
 }
 
+/// Drop cache in guest
+/// * Args:
+/// - drop: determins which cache will be droped. Value should be in {1, 2, 3}
```
Tim-0731-Hzt, line 78: typo: determines

src/agent/src/util.rs
```diff
@@ -72,6 +73,22 @@ pub async fn get_vsock_stream(fd: RawFd) -> Result<VsockStream> {
...
+/// 1: page cahe is dropped
```
Tim-0731-Hzt, line 79: typo: cache

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs
```diff
@@ -183,6 +184,17 @@ impl ContainerInner {
                 }
             })?;
 
+        self.agent
+            .drop_cache(DropCacheRequest {
+                drop_option: 2,
```
Tim-0731-Hzt, line 189: In which case, the `drop_option` will be 1 or 3 ?

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs
studychao: I got the same question here.

src/runtime-rs/crates/agent/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub mod kata;
 mod log_forwarder;
 mod sock;
 pub mod types;
+use protocols::agent::DropCacheRequest;
```
YushuoEdge, line 16: The agent related Request should be better defined in 'types', which is under 'crates/agent/src/type.rs' now, like the other Request.
This situation maybe changed in the future, as we do not have to support grpc right now. But I think It might be easier to do some same modifications by then.

xyc-grace:
The 'EBUSY' problem has not yet arisen in the kata community, and its root cause can be traced to VirtIO-FS. Therefore, it's best to close this pull request and wait for the issue to be resolved in VirtIO-FS.

PR #6658, opened 2023-04-13
osbuilder: Enable dbus in the dracut case

### LINKED ISSUE #6657: Kata agent fails to connect to systemd
When the guest image has systemd as init (as created with dracut), the agent terminates and the following error is returned :

```
Caused by:
    0: Establishing a D-Bus connection
    1: I/O error: No such file or directory (os error 2)
    2: No such file or directory (os error 2): unknown 
```

This is the yet another consequence of #5154 : D-Bus should be enabled in the guest if the agent needs to talk to systemd.

gkurz:
The agent now offloads cgroup configuration to systemd when possible. This requires to enable D-Bus in order to communicate with systemd.

Fixes #6657

### DIFF
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -475,6 +475,8 @@ prepare_overlay()
 	# Kata systemd unit file
 	mkdir -p ./etc/systemd/system/basic.target.wants/
 	ln -sf /usr/lib/systemd/system/kata-containers.target ./etc/systemd/system/basic.target.wants/kata-containers.target
+	mkdir -p ./etc/systemd/system/kata-containers.target.wants/
+	ln -sf /usr/lib/systemd/system/dbus.socket  ./etc/systemd/system/kata-containers.target.wants/dbus.socket
 	popd  > /dev/null
 }
 
--- a/tools/osbuilder/dracut/dracut.conf.d/05-base.conf
+++ b/tools/osbuilder/dracut/dracut.conf.d/05-base.conf
@@ -14,4 +14,4 @@ hostonly_cmdline="no"
 # create reproducible images
 reproducible="yes"
 # dracut modules to include (NOTE: these are NOT kernel modules)
-dracutmodules="kernel-modules udev-rules syslog systemd"
+dracutmodules="kernel-modules udev-rules syslog systemd dbus"


jepio (approved):

fidencio (approved):
lgtm, thanks @gkurz!

gkurz:
> lgtm, thanks @gkurz!

Thanks @jepio and @fidencio ! :smile:

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -475,6 +475,8 @@ prepare_overlay()
 	# Kata systemd unit file
 	mkdir -p ./etc/systemd/system/basic.target.wants/
 	ln -sf /usr/lib/systemd/system/kata-containers.target ./etc/systemd/system/basic.target.wants/kata-containers.target
+	mkdir -p ./etc/systemd/system/kata-containers.target.wants/
+	ln -sf /usr/lib/systemd/system/dbus.socket  ./etc/systemd/system/kata-containers.target.wants/dbus.socket
```
c3d, line 479: I know this was done that way before, but why is there an absolute path for /usr but a relative one for ./etc? I assume that it's because it's a symbolic link and you want the path as it exists on the target?

c3d (approved):
FWIW, looks good to me, thanks @gkurz

tools/osbuilder/rootfs-builder/rootfs.sh
gkurz: > I know this was done that way before, but why is there an absolute path for /usr but a relative one for ./etc? I assume that it's because it's a symbolic link and you want the path as it exists on the target?

The first argument to `ln -s` is basically just a string that will be parsed at runtime when the symbolic link is traversed. The second argument is the path of the actual symbolic link we're creating. So yes, your assumption is correct.

Vlad1mir-D: @gkurz Why not to insert the same logic here: https://github.com/gkurz/kata-containers/blob/eb1762e813c52d5b593bacb04932d6fbf0ca3a19/tools/osbuilder/rootfs-builder/rootfs.sh#L630?

gkurz: > @gkurz Why not to insert the same logic here: https://github.com/gkurz/kata-containers/blob/eb1762e813c52d5b593bacb04932d6fbf0ca3a19/tools/osbuilder/rootfs-builder/rootfs.sh#L630?

Hi @Vlad1mir-D ! Thanks for reporting this. I simply missed it :disappointed: . Let's address this with #6677 .

PR #6659, opened 2023-04-13
osbuilder: Enable dbus in the dracut case (backport for 3.1)

### LINKED ISSUE #6657: Kata agent fails to connect to systemd

gkurz:
The agent now offloads cgroup configuration to systemd when possible. This requires to enable D-Bus in order to communicate with systemd.

Fixes #6657

(cherry picked from commit eb1762e813c52d5b593bacb04932d6fbf0ca3a19)

### DIFF
--- a/tools/osbuilder/dracut/dracut.conf.d/05-base.conf
+++ b/tools/osbuilder/dracut/dracut.conf.d/05-base.conf
@@ -14,4 +14,4 @@ hostonly_cmdline="no"
 # create reproducible images
 reproducible="yes"
 # dracut modules to include (NOTE: these are NOT kernel modules)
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @gkurz!

jepio (approved):

jepio:
thanks for the quick backport

Issue #5339, opened 2022-10-06
kata-ctl: env: Implement with unit tests


jodh-intel:
Implement the `kata-runtime env` command with unit tests.

See: https://github.com/kata-containers/kata-containers/issues/4499.

Issue #6090, opened 2023-01-17
kata-ctl: does kata-ctl plan to support runtime(golang)?


YushuoEdge:
Currently, we have several tools to help using kata containers. And for a long time, we may maintain two kinds of runtime (`runtime-go` and `runtime-rs`). However, different tools and runtimes are not fully compatible. For example, if we are using `runtime-rs`, some functions of kata-monitor will be invalid.

Kata-ctl seems to plan to be a unified tool for kata to reduce the number of tools for ease of use. I wonder if we need to make kata-ctl also support runtime-go so that users will not need to choose the tools according to the runtime they are using. Also, some of our CIs are using those tools,  these CIs have to determine which runtime is used when they are triggered, which cause some inconvenience.

liubin:
IMO we should use `kata-ctl` to replace `kata-runtime` completely to reduce maintenance costs, but golang runtime will live for a long time, so `kata-ctl` should be compatible with golang runtime.

jodh-intel:
Thanks for raising @YushuoEdge.

## Requirements

There are two potential requirements here fwics:

1) Make `kata-ctl` work *with* the existing `runtime-go` (aka ensure `kata-ctl` can communicate with the golang Kata runtime).

2) Make `kata-ctl` CLI compatible with "`kata-runtime check`" (support identical CLI commands and options).

3) Make `kata-ctl` output compatible with "`kata-runtime check`" (support identical output format.

Just to be clear, I believe we are talking here about (1) only, but read on...

## Full compatibility?

One of the reasons for writing `kata-ctl` was to improve on the existing "`kata-runtime check`" utility, but we didn't necessarily plan to be CLI/output compatible as that could limit the scope of `kata-ctl`.

### Compat script

We certainly could make `kata-ctl` CLI/output compatible if that is also required, but iff that is necessary, the best solution might be to provide a script in the Kata 3.x release archives called `kata-runtime` that only supports the `runtime-go` commands (`kata-runtime [check|env|etc...]`). That script would then call `kata-ctl` passing it the appropriate (new) options.

For example, we could always add a `kata-ctl --compat-mode <cmd> <options...>` or similar and make the `kata-runtime` script call `kata-ctl` like that.

## More input required

[...]

Issue #5954, opened 2022-12-23
runtime-rs changes Kata config file format


jodh-intel:
Working on #5242 I spotted a config change that runtime-rs introduces.

In the Kata 2.x TOML config file, we have sections like this:

```toml
[hypervisor.foo]
# ...

[agent.kata]
# ...

[runtime]
# ...
```

... but with runtime-rs, we seem to now have:

```toml
[hypervisor.dragonball]
# ...

[agent.kata]
# ...

[runtime]

# ...
name="virt_container"
hypervisor_name="dragonball"  # XXX: <-- Redundant information.
agent_name="kata"             # XXX: <-- Redundant information.

# ...
```

As shown by the comment, it's unclear why we need runtime.hypervisor_name given that the TOML format itself describes that value as `dragonball` (from the `[hypervisor.dragonball]` heading). In other words, specifying `runtime.hypervisor_name` is redundant - it's already been specified.

Why is this needed? The current runtime-rs code assumes this new format which means existing 2.x config files for stand-alone hypervisors do not work unless their config files are modified to include the redundant `runtime.hypervisor_name` and `runtime.agent_name`.

Specifically, the problem is `Annotation::update_config_by_annotation()` which does this:

```rust
if config.runtime.hypervisor_name.is_empty() {
    config.runtime.hypervisor_name = DEFAULT_HYPERVISOR.to_string()
}
```

... whereas the code should be looking at `config.hypervisor` as that contains the details in the `[hypervisor.name]` TOML section.

> **Note:** The same logic applies for determining the value of "agent".

See:

- https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/config/configuration-dragonball.toml.in
- https://github.com/kata-containers/kata-containers/blob/main/src/libs/kata-types/src/annotations/mod.rs#L451..L456

/cc @studychao, @jiangliu, @pmores.

studychao:
Thanks james for raising this:)

@Tim-0731-Hzt is helping to look at this.

pmores:
I ran into this too but I didn't realise this was a departure from the previous state.  I assumed that it was possible to have several hypervisors configured in a single config file (i.e. to have several `[hypervisor.XXX]` tables), with `runtime.hypervisor_name` selecting the active one among them.

PR #6660, opened 2023-04-14
Implement the "kata-ctl env" command

amshinde:
Command implements functionality to get user environment settings.
    
Fixes: #5339

### DIFF
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -8,6 +8,7 @@
 use anyhow::{anyhow, Result};
 use reqwest::header::{CONTENT_TYPE, USER_AGENT};
 use serde::{Deserialize, Serialize};
+use std::fmt;
 use thiserror::Error;
 
 #[cfg(any(target_arch = "x86_64"))]
@@ -145,6 +146,19 @@ pub enum GuestProtection {
     Snp,
     Pef,
     Se,
+}
+
+impl fmt::Display for GuestProtection {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            GuestProtection::Tdx => write!(f, "tdx"),
+            GuestProtection::Sev => write!(f, "sev"),
+            GuestProtection::Snp => write!(f, "snp"),
+            GuestProtection::Pef => write!(f, "pef"),
+            GuestProtection::Se => write!(f, "se"),
+            GuestProtection::NoProtection => write!(f, "none"),
+        }
+    }
 }
 
 #[allow(dead_code)]
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -47,7 +47,6 @@ pub const GENERIC_CPU_MODEL_FIELD: &str = "model name";
 #[allow(dead_code)]
 pub const PROC_CPUINFO: &str = "/proc/cpuinfo";
 
-#[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
 fn read_file_contents(file_path: &str) -> Result<String> {
     let contents = std::fs::read_to_string(file_path)?;
     Ok(contents)
@@ -55,7 +54,6 @@ fn read_file_contents(file_path: &str) -> Result<String> {
 
 // get_single_cpu_info returns the contents of the first cpu from
 // the specified cpuinfo file by parsing based on a specified delimiter
-#[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
 pub fn get_single_cpu_info(cpu_info_file: &str, substring: &str) -> Result<String> {
     let contents = read_file_contents(cpu_info_file)?;
 
--- a/src/tools/kata-ctl/src/ops/env_ops.rs
+++ b/src/tools/kata-ctl/src/ops/env_ops.rs
@@ -449,7 +449,7 @@ pub fn handle_env(env_args: EnvArgument) -> Result<()> {
         Box::new(io::stdout())
     };
 
-    let (toml_config, _) = TomlConfig::load_raw_from_file("").context("load toml config")?;
+    let (toml_config, _) = TomlConfig::load_from_default().context("load toml config")?;
 
     let env_info = get_env_info(&toml_config)?;
 
--- a/src/tools/kata-ctl/src/ops/env_ops.rs
+++ b/src/tools/kata-ctl/src/ops/env_ops.rs
@@ -1,0 +1,431 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Contains checks that are not architecture-specific
+
+use crate::arch::arch_specific;
+use crate::args::EnvArgument;
+use crate::check;
+use crate::ops::version;
+use crate::utils;
+use kata_types::config::TomlConfig;
+
+use anyhow::{anyhow, Context, Result};
+use serde::{Deserialize, Serialize};
+use std::process::Command;
+use sys_info;
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct HostInfo {
+    #[serde(default)]
+    available_guest_protection: String,
+    #[serde(default)]
+    kernel: String,
+    #[serde(default)]
+    architecture: String,
+    #[serde(default)]
+    vm_container_capable: bool,
+    #[serde(default)]
+    support_vsocks: bool,
+    #[serde(default)]
+    distro: DistroInfo,
+    #[serde(default)]
+    cpu: CPUInfo,
+    #[serde(default)]
+    memory: MemoryInfo,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct DistroInfo {
+    #[serde(default)]
+    name: String,
+    #[serde(default)]
+    version: String,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct CPUInfo {
+    #[serde(default)]
+    vendor: String,
+    #[serde(default)]
+    model: String,
+    #[serde(default)]
+    cpus: usize,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct MemoryInfo {
+    #[serde(default)]
+    total: u64,
+    #[serde(default)]
+    available: u64,
+    #[serde(default)]
+    free: u64,
+}
+
+// Semantic version for the output of the command.
+//
+// XXX: Increment for every change to the output format
+// (meaning any change to the EnvInfo type).
+const FORMAT_VERSION: &str = "0.0.1-kata-ctl";
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct MetaInfo {
+    #[serde(default)]
+    version: String,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct VersionInfo {
+    #[serde(default)]
+    semver: String,
+    #[serde(default)]
+    commit: String,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct RuntimeConfigInfo {
+    #[serde(default)]
+    path: String,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct RuntimeInfo {
+    #[serde(default)]
+    path: String,
+    #[serde(default)]
+    guest_selinux_label: String,
+    #[serde(default)]
+    pub experimental: Vec<String>,
+    #[serde(default)]
+    debug: bool,
+    #[serde(default)]
+    trace: bool,
+    #[serde(default)]
+    disable_guest_seccomp: bool,
+    #[serde(default)]
+    disable_new_net_ns: bool,
+    #[serde(default)]
+    sandbox_cgroup_only: bool,
+    #[serde(default)]
+    static_sandbox_resource_mgmt: bool,
+    #[serde(default)]
+    config: RuntimeConfigInfo,
+    #[serde(default)]
+    version: VersionInfo,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct AgentInfo {
+    #[serde(default)]
+    debug: bool,
+    #[serde(default)]
+    trace: bool,
+}
+// KernelInfo stores kernel details
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct KernelInfo {
+    #[serde(default)]
+    path: String,
+    #[serde(default)]
+    parameters: String,
+}
+
+// InitrdInfo stores initrd image details
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct InitrdInfo {
+    #[serde(default)]
+    path: String,
+}
+
+// ImageInfo stores root filesystem image details
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct ImageInfo {
+    #[serde(default)]
+    path: String,
+}
+
+// HypervisorInfo stores hypervisor details
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct HypervisorInfo {
+    #[serde(default)]
+    machine_type: String,
+    #[serde(default)]
+    machine_accelerators: String,
+    #[serde(default)]
+    version: String,
+    #[serde(default)]
+    path: String,
+    #[serde(default)]
+    block_device_driver: String,
+    #[serde(default)]
+    entropy_source: String,
+    #[serde(default)]
+    shared_fs: String,
+    #[serde(default)]
+    virtio_fs_daemon: String,
+    #[serde(default)]
+    msize_9p: u32,
+    #[serde(default)]
+    memory_slots: u32,
+    #[serde(default)]
+    pcie_root_port: u32,
+    #[serde(default)]
+    hotplug_vfio_on_rootbus: bool,
+    #[serde(default)]
+    debug: bool,
+    #[serde(default)]
+    enable_iommu: bool,
+    #[serde(default)]
+    enable_iommu_platform: bool,
+    #[serde(default)]
+    default_vcpus: i32,
+    #[serde(default)]
+    cpu_features: String,
+}
+
+// EnvInfo collects all information that will be displayed by the
+// env command.
+//
+// XXX: Any changes must be coupled with a change to formatVersion.
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct EnvInfo {
+    #[serde(default)]
+    kernel: KernelInfo,
+    #[serde(default)]
+    meta: MetaInfo,
+    #[serde(default)]
+    image: ImageInfo,
+    #[serde(default)]
+    initrd: InitrdInfo,
+    #[serde(default)]
+    hypervisor: HypervisorInfo,
+    #[serde(default)]
+    runtime: RuntimeInfo,
+    #[serde(default)]
+    host: HostInfo,
+    #[serde(default)]
+    agent: AgentInfo,
+}
+
+pub fn get_meta_info() -> MetaInfo {
+    MetaInfo {
+        version: String::from(FORMAT_VERSION),
+    }
+}
+
+pub fn get_memory_info() -> Result<MemoryInfo> {
+    let mem_info = sys_info::mem_info().context("get host memory information")?;
+    Ok(MemoryInfo {
+        total: mem_info.total,
+        available: mem_info.avail,
+        free: mem_info.free,
+    })
+}
+
+fn get_host_info() -> Result<HostInfo> {
+    let host_kernel_version = utils::get_kernel_version(utils::PROC_VERSION_FILE)?;
+    let (host_distro_name, host_distro_version) =
+        utils::get_distro_details(utils::OS_RELEASE, utils::OS_RELEASE_CLR)?;
+    let (cpu_vendor, cpu_model) = utils::get_generic_cpu_details(check::PROC_CPUINFO)?;
+
+    let host_distro = DistroInfo {
+        name: host_distro_name,
+        version: host_distro_version,
+    };
+
+    let cores: usize = std::thread::available_parallelism()
+        .context("get available parallelism")?
+        .into();
+
+    let host_cpu = CPUInfo {
+        vendor: cpu_vendor,
+        model: cpu_model,
+        cpus: cores,
+    };
+
+    let memory_info = get_memory_info()?;
+
+    let guest_protection =
+        arch_specific::available_guest_protection().map_err(|e| anyhow!(e.to_string()))?;
+
+    let guest_protection = guest_protection.to_string();
+
+    let support_vsocks = utils::supports_vsocks(utils::VHOST_VSOCK_DEVICE)?;
+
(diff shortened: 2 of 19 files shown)


amshinde:
cc @jodh-intel

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+        static_sandbox_resource_mgmt: toml_config.runtime.static_sandbox_resource_mgmt,
+        config: RuntimeConfigInfo { path: toml_path },
+    }
+}
+
+pub fn get_agent_info(toml_config: &TomlConfig) -> Result<AgentInfo> {
+    let agent_config = toml_config
+        .agent
+        .get(&toml_config.runtime.agent_name)
+        .ok_or("could not find agent config in configuration")
```
jodh-intel, line 321: I hit this branch in testing:

```bash
$ grep '^\[agent' /etc/kata-containers/configuration.toml
[agent.kata]
sudo $(which kata-ctl)  env
ERROR: "could not find agent config in configuration"
```

However, this behaviour is incorrect due to an existing bug in the runtime-rs implementation - see https://github.com/kata-containers/kata-containers/issues/5954.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+    let (toml_config, _) = TomlConfig::load_raw_from_file("").context("load toml config")?;
+
+    let env_info = get_env_info(&toml_config)?;
+
+    if env_args.json {
+        let serialized_json = serde_json::to_string_pretty(&env_info).unwrap();
+        println!("Env:\n{}", serialized_json);
+    } else {
+        let toml = toml::to_string_pretty(&env_info).unwrap();
+        println!("Env:\n{}", toml);
```
jodh-intel, line 441: Can you remove the "Env:" output as it invalidates the output (TOML or JSON).

src/tools/kata-ctl/src/check.rs
```diff
@@ -137,6 +138,19 @@ pub enum GuestProtection {
     Se,
 }
 
+impl fmt::Display for GuestProtection {
```
jodh-intel, line 141: This is fine for now, but I think `GuestProtection` should be defined per platform as we do for the checks via the `get_checks()` call. I've raised #6662 so we can handle this later.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Contains checks that are not architecture-specific
+
+#![allow(dead_code)]
+
+use crate::args::EnvArgument;
+use crate::ops::version;
+//use anyhow::Result;
```
jodh-intel, line 12: Commented out code.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+    required_cpu_attribs: HashMap<String, String>,
+    required_kernel_modules: HashMap<String, KernelModule>,
+    cpu_info_file: String,
+}
+
+// Semantic version for the output of the command.
+//
+// XXX: Increment for every change to the output format
+// (meaning any change to the EnvInfo type).
+const FORMAT_VERSION: &str = "1.0.26";
```
jodh-intel, line 89: This is an awkard one. I can see this has been carried over from [the golang implementation](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-env.go#L32), but since the format and content of `kata-ctl env` is different to `kata-runtime env`, I think we should probably reset this to `0.0.1` and maybe add a patch suffix (`0.0.1-kata-ctl`) to allow the caller to identify that this is not actually `kata-runtime check` that has suddenly been "downgraded".

See: https://github.com/kata-containers/kata-containers/issues/6090

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+    Ok(version)
+}
+
+pub fn get_hypervisor_info(
+    toml_config: &TomlConfig,
+) -> Result<(HypervisorInfo, ImageInfo, KernelInfo, InitrdInfo)> {
+    let hypervisor_config = toml_config
+        .hypervisor
+        .get(&toml_config.runtime.hypervisor_name)
+        .ok_or("could not find hypervisor config in configuration")
```
jodh-intel, line 353: Same issue as the agent above. See: https://github.com/kata-containers/kata-containers/issues/5954

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+    let support_v_socks = utils::supports_vsocks(utils::VHOST_VSOCK_DEVICE)?;
+
+    Ok(HostInfo {
+        kernel: host_kernel_version,
+        architecture: String::from(std::env::consts::ARCH),
+        distro: host_distro,
+        cpu: host_cpu,
+        memory: memory_info,
+        available_guest_protection: guest_protection,
+        vm_container_capable: true, // TODO: implement
```
jodh-intel, line 279: Could you raise an issue and add the URL here for this?

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+    toml_config: &TomlConfig,
+) -> Result<(HypervisorInfo, ImageInfo, KernelInfo, InitrdInfo)> {
+    let hypervisor_config = toml_config
+        .hypervisor
+        .get(&toml_config.runtime.hypervisor_name)
+        .ok_or("could not find hypervisor config in configuration")
+        .map_err(|e| anyhow!(e))?;
+
+    let version =
+        get_command_version(&hypervisor_config.path).context("error getting hypervisor version")?;
```
jodh-intel, line 357: This is being displayed on multiple lines. On my systems:

## CH

  ```bash
  $ /usr/local/bin/cloud-hypervisor --version|tee /dev/tty|wc -l
  cloud-hypervisor v30.0-6-g04ee1efa-dirty
  1
  ```

But the output shown is:

```toml
[hypervisor]
machine_type = ''
machine_accelerators = ''
version = '''
cloud-hypervisor v30.0-6-g04ee1efa-dirty
'''
path = '/usr/local/bin/cloud-hypervisor'
```

## QEMU

```bash
$ /opt/kata/bin/qemu-system-x86_64 --version|tee /dev/tty|wc -l
QEMU emulator version 7.2.0 (kata-static)
Copyright (c) 2003-2022 Fabrice Bellard and the QEMU Project developers
2
```

```toml
[hypervisor]
machine_type = 'q35'
machine_accelerators = ''
version = '''
QEMU emulator version 7.2.0 (kata-static)
Copyright (c) 2003-2022 Fabrice Bellard and the QEMU Project developers
'''
path = '/opt/kata/bin/qemu-system-x86_64'
```

`kata-runtime env` displays the NL as `\n` whilst maintaining a single line:

```toml
[Hypervisor]
  MachineType = "q35"
  Version = "QEMU emulator version 7.2.0 (kata-static)\nCopyright (c) 2003-2022 Fabrice Bellard and the QEMU Project developers"
  Path = "/opt/kata/bin/qemu-system-x86_64"
```

jodh-intel (commented):
Thanks for raising @amshinde !

A few comments inline. Also, if I run this without `root` perms, I get an error:

  ```bash
  $ kata-ctl env
  ERROR: NoPerms
  $ echo $?
  1
  ```

  This is coming from `ProtectionError` in `src/check.rs`, but rather than showing `NoPerms`, the error displayed should be:

```
ERROR: No permission to check guest protection
```

From the `#[error]` here: https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/check.rs#L140..L144.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,445 @@
...
+            None => String::new(),
+        };
+    }
+
+    RuntimeInfo {
+        // TODO: Needs to be implemented
+        path: "TBD".to_string(),
+        version,
+        experimental: toml_config.runtime.experimental.clone(),
+        guest_se_linux_label: String::from("not implemented yet"),
```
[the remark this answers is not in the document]
amshinde: Created https://github.com/kata-containers/kata-containers/issues/6667 for this.

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: Yes, came accross this issue as well. However I tested this PR with the dragonball config introduced in runtime-rs and it works with that.

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: Done, some debug statement left behind.

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: This shows correctly in the json output. Seems to be an issue with the serde toml encoder. Let me see if I can fix this.

amshinde: @jodh-intel This is fixed now. It was caused by `toml:to_string_pretty`. I am using toml::to_string instead now, the string now shows correctly.  I feel this is a bug with the pretty output. 
In case of toml, I see that I didnt really need the pretty output, but I do need that in case of json.

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: Fixed

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: Fixed

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: @jodh-intel I needed to implement this was as the agent field in the TomlConfig is a HashMap keyed on the agent name and not a single agent config structure:
https://github.com/kata-containers/kata-containers/blob/e90989b16b8ea8cfd0c7481f1c2406534d2b9699/src/libs/kata-types/src/config/mod.rs#L104
Similarly for the hypervisor field.  This would need to be resolved in how the toml is parsed and stored in the `kata-types` crate. Once it is resolved there, I can change it here. Wdyt?

amshinde: @jodh-intel I have addressed this in a following commit. For configs where agent_name and hypervisor_name are absent, I resort to using the first entry in the hashmap that is meant to store the agent and hypervisor configs by runtime-rs.

fidencio:
@amshinde, could you rebase your PR atop of current main?  This will allow the actions added in the past week to be found.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+    #[serde(default)]
+    available_guest_protection: String,
+    #[serde(default)]
+    kernel: String,
+    #[serde(default)]
+    architecture: String,
+    #[serde(default)]
+    vm_container_capable: bool,
+    #[serde(default)]
+    support_v_socks: bool,
```
jodh-intel, line 34: Nit: I'd consider renaming to:

```suggestion
    support_vsocks: bool,
```

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+    #[serde(default)]
+    path: String,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct RuntimeInfo {
+    #[serde(default)]
+    path: String,
+    #[serde(default)]
+    guest_se_linux_label: String,
```
jodh-intel, line 114: Same here:

```suggestion
    guest_selinux_label: String,
```

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+pub fn handle_env(env_args: EnvArgument) -> Result<()> {
+    let (toml_config, _) = TomlConfig::load_raw_from_file("").context("load toml config")?;
+
+    let env_info = get_env_info(&toml_config)?;
+
+    if env_args.json {
+        let serialized_json = serde_json::to_string_pretty(&env_info).unwrap();
+        println!("{}", serialized_json);
+    } else {
+        let toml = toml::to_string(&env_info).unwrap();
```
jodh-intel, line 463: - Might be worth adding a comment explaining why we're not using pretty here.
- Could you replace the `.unwrap()` calls with `?` here.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+    Ok(env_info)
+}
+
+pub fn handle_env(env_args: EnvArgument) -> Result<()> {
+    let (toml_config, _) = TomlConfig::load_raw_from_file("").context("load toml config")?;
+
+    let env_info = get_env_info(&toml_config)?;
+
+    if env_args.json {
+        let serialized_json = serde_json::to_string_pretty(&env_info).unwrap();
```
jodh-intel, line 460:
```suggestion
        let serialized_json = serde_json::to_string_pretty(&env_info)?;
```

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+        hypervisor: hypervisor_info,
+        host: host_info,
+        agent: agent_info,
+    };
+
+    Ok(env_info)
+}
+
+pub fn handle_env(env_args: EnvArgument) -> Result<()> {
+    let (toml_config, _) = TomlConfig::load_raw_from_file("").context("load toml config")?;
```
jodh-intel, line 455: This looks a bit odd as there is no file specified. Is there an alternative (`TomlConfig::Default()` or similiar?)

jodh-intel (commented):
Thanks @amshinde.

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: @jodh-intel The current API does not have that, if the config_file provided to load_raw_from_file(config_file) is empty, it is only then that the kata-types crate searches for config file in the order configured.

On a side note, I was thinking it will be useful to support passing a configuration file path to the kata-ctl env, that way the tool can show environment wrt to the config file provided. Wdyt?

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: > Thanks for raising @amshinde !
> 
> A few comments inline. Also, if I run this without `root` perms, I get an error:
> 
> ```shell
> $ kata-ctl env
> ERROR: NoPerms
> $ echo $?
> 1
> ```
> 
> This is coming from `ProtectionError` in `src/check.rs`, but rather than showing `NoPerms`, the error displayed should be:
> 
> ```
> ERROR: No permission to check guest protection
> ```
> 
> From the `#[error]` here: https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/check.rs#L140..L144.

@jodh-intel Will be looking into this, but I am not sure what I need to be doing differently here. I simple invoke `available_guest_protection` as 
```
let guest_protection: String = arch_specific::available_guest_protection()?.to_string();
```
I would assume that `?` operator would cause the string for the error to return here. Do you know how this should be handled? I'll continue to take a look at this tomorrow.

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: fixed

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: fixed

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: I have fixed this now. I had to explicitly call to_string() on the ProtectionError returned to fix this. I now this this:
```
cargo run -- env 
ERROR: Error {
    context: "get host information",
    source: "No permission to check guest protection",
}
```

dborquez (approved):
LGTM

cmaf (commented):
@amshinde, could you explain why you're implementing this in src/ops/env_ops.rs rather than something like src/ops.rs similar to check.rs? 

I'm not saying it should be one way or the other, I have lost track of why we have this distinction. Early on in eb5423cb7f we made a change in the CLI that increased the number of files for command and argument handling and some of the kata-ctl functionality is now in /ops too. Could anyone clarify the reasoning?

cc @dborquez  @jodh-intel

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Contains checks that are not architecture-specific
+
+#![allow(dead_code)]
```
jodh-intel, line 8: Can this be removed (now)?

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+        let path_str = p.to_str();
+        toml_path = match path_str {
+            Some(s) => String::from(s),
+            None => String::new(),
+        };
+    }
+
+    RuntimeInfo {
+        // TODO: Needs to be implemented
+        path: "TBD".to_string(),
```
jodh-intel, line 301: Again, could you maybe change "TBD" to a GH issue URL so we don't forget this.

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+        version,
+        experimental: toml_config.runtime.experimental.clone(),
+        // TODO: See https://github.com/kata-containers/kata-containers/issues/6667
+        guest_selinux_label: String::from("not implemented yet"),
```
jodh-intel, line 305: Thanks - how about putting the URL in the output string too?

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+
+pub fn get_agent_info(toml_config: &TomlConfig) -> Result<AgentInfo> {
+    if toml_config.agent.len() == 0 {
+        return Err(anyhow!("Missing agent config"));
+    }
+
+    // Assign the first entry to the agent config after checking hashmap is not empty, to make this
+    // work for configs where agent_name is absent.
+    // This is a workaround for https://github.com/kata-containers/kata-containers/issues/5954
+    let (_, mut agent_config) = toml_config.agent.iter().next().unwrap();
```
jodh-intel, line 324: Since the function returns a `Result`, can you replace the `unwrap()`. with `?` here?

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -0,0 +1,468 @@
...
+    toml_config: &TomlConfig,
+) -> Result<(HypervisorInfo, ImageInfo, KernelInfo, InitrdInfo)> {
+    if toml_config.hypervisor.len() == 0 {
+        return Err(anyhow!("Missing hypervisor config"));
+    }
+
+    // Assign the first entry to the hypervisor config as above after checking hashmap is not empty, to make this
+    // work for configs where hypervisor_name is absent.
+    // This is a workaround for https://github.com/kata-containers/kata-containers/issues/5954
+    let (_, mut hypervisor_config) = toml_config.hypervisor.iter().next().unwrap();
```
jodh-intel, line 368: Again could you `.unwrap()` -> `?`?

src/tools/kata-ctl/src/ops/env_ops.rs
jodh-intel: Sounds good, but feel free to raise an issue and add that on a separate PR if you want.

jodh-intel (commented):
Thanks @amshinde - a few more comments...

src/tools/kata-ctl/src/ops/env_ops.rs
amshinde: The unwrap was for getting the value from Option fetched from next(). I had made sure this does not panic after checking for the length of the hashmap itself.
I'll rewrite this to avoid using unwrap() altogether.

amshinde:
@jodh-intel I have addressed all your comments. Please take a look.

jodh-intel (commented):
Thanks @amshinde. This is looking good, but I found a bug that I think needs fixing first:

```bash
$ target/x86_64-unknown-linux-musl/release/kata-ctl env
ERROR: Error {
    context: "get host information",
    source: "No permission to check guest protection",
}
```

I still think we should make this error clearer and tell the user they need to run the command as `root` / use `sudo`. But even when I do that:

```bash
$ sudo target/x86_64-unknown-linux-musl/release/kata-ctl env
ERROR: Error {
    context: "get host information",
    source: Os {
        code: 2,
        kind: NotFound,
        message: "No such file or directory",
    },
}
```

This much more cryptic error is due to `vhost_vsock` not being loaded. Once I'd `sudo modprobe vhost_vsock`, I got the expected output.

amshinde:
@jodh-intel Resolved those issues. Had to make a new change regarding fetching cpu details work accross all architectures. PTAL.

src/tools/kata-ctl/src/utils.rs
```diff
@@ -144,9 +141,14 @@ pub fn get_generic_cpu_details(cpu_info_file: &str) -> Result<(String, String)>
     Ok((vendor, model))
 }
 
-const VHOST_VSOCK_DEVICE: &str = "/dev/vhost-vsock";
+pub const VHOST_VSOCK_DEVICE: &str = "/dev/vhost-vsock";
 pub fn supports_vsocks(vsock_path: &str) -> Result<bool> {
-    let metadata = fs::metadata(vsock_path)?;
+    let metadata = fs::metadata(vsock_path).map_err(|err| {
+        anyhow!(
+            "Host system does not support vhost-vsock: {}",
```
jodh-intel, line 148:
```suggestion
            "Host system does not support vhost-vsock (try running `sudo modprobe vhost_vsock`): {}",
```

jodh-intel (approved):
Thanks @amshinde. I've made an optional comment about making the vhost-vsock error clearer to users but overall...

lgtm

ftr, here's the latest error you get if the `vhost_vsock` kernel module isn't loaded:

```bash
$ sudo target/debug/kata-ctl env
ERROR: Error {
    context: "get host information",
    source: "Host system does not support vhost-vsock: No such file or directory (os error 2)",
}
$ echo $?
1
```

To resolve:


```bash
$ sudo modprobe vhost_vsock
$ sudo target/debug/kata-ctl env &>/dev/null; echo $?
0
```

jodh-intel:
\o/ Thanks @amshinde ! :smile:

Issue #6663, opened 2023-04-14
gpu: Do not pass-through PCI (Host) Bridges


zvonkok:
On some systems a GPU is in a IOMMU group with a PCI Bridge and PCI Host Bridge. Per default no PCI Bridge needs to be passed-through. 

```
IOMMU Group 44:
        40:03.0 Host bridge [0600]: Advanced Micro Devices, Inc. [AMD] Starship/Matisse PCIe Dummy Host Bridge [1022:1482]
        40:03.1 PCI bridge [0604]: Advanced Micro Devices, Inc. [AMD] Starship/Matisse GPP Bridge [1022:1483]
        41:00.0 3D controller [0302]: NVIDIA Corporation GA100 [A100 PCIe 80GB] [10de:20b5] (rev a1)
```
When scanning the IOMMU group, ignore devices with a `0x60` class ID prefix. 
```
Host Bridge Class ID: 0x6000
PCI Bridge Class ID: 0x6040
```

PR #6664, opened 2023-04-14
gpu: Do not pass-through PCI (Host) Bridges

zvonkok:
On some systems a GPU is in a IOMMU group with a PCI Bridge and PCI Host Bridge. Per default no PCI Bridge needs to be passed-through. When scanning the IOMMU group, ignore devices with a 0x60 class ID prefix.

Fixes: #6663


fidencio (approved):
looks sensible to me, thanks @zvonkok!

fengwang666 (approved):

PR #6670, opened 2023-04-15
cache-components: Fix caching of TDVF and QEMU for TDX

fidencio:
---

cached-components: Fix TDX QEMU caching

TDX QEMU caching is not working as expected, as we're checking for its
version looking at "assets.hypervisor.${QEMU_FLAVOUR}.version", which is
correct for standard QEMU. However, for TDX QEMU we should be checking
for "assets.hypervisor.${QEMU_FLAVOUR}.tag"

---

cache-components: Fix TDVF caching

TDVF caching is not working as the tarball name is incorrect. The result
expected is kata-static-tdvf.tar.xz, but it's looking for
kata-static-tdx.tar.xz.

This happens as a logic to convert tdx -> tdvf has been added as part of
the building scripts, but I missed doing this as part of the caching
scripts.

---


jepio (approved):
LGTM

snir911 (approved):
LGTM

Issue #6679, opened 2023-04-18
A unit test bug about `Bad file descriptor (os error 9)` 


Tim-Zhang:
Run `cargo test` in agent folder, Got:
```
test tests::test_create_logger_task ... ok
test sandbox::tests::update_shared_pidns ... FAILED
test sandbox::tests::unset_sandbox_storage ... ok
test rpc::tests::test_do_write_stream ... ok
test watcher::tests::create_tmpfs ... ok
test uevent::tests::test_create_logger_task2 ... ok
test watcher::tests::make_target_path ... ok
test watcher::tests::delete_file ... ok
test watcher::tests::spawn_thread ... ok
test watcher::tests::test_copy ... ok
test watcher::tests::test_single_file_check ... ok
test watcher::tests::test_empty_sourcedir_check ... ok
test watcher::tests::test_watch_entries ... ok
test watcher::tests::verify_container_cleanup_watching ... ok
test watcher::tests::watch_directory ... ok
test sandbox::tests::test_create_logger_task1 ... ok
test watcher::tests::watch_directory_with_symlinks ... ok
test watcher::tests::watch_file ... ok
test util::tests::test_interruptable_io_copier_eof ... ok
test watcher::tests::watch_directory_verify_dir_removal ... ok
test util::tests::test_interruptable_io_copier_shutdown ... ok
test watcher::tests::watch_directory_too_large ... ok
test util::tests::test_interruptable_io_copier_reader ... ok

failures:

---- sandbox::tests::update_shared_pidns stdout ----
thread 'sandbox::tests::update_shared_pidns' panicked at 'called `Result::unwrap()` on an `Err` value: fail to create cgroup manager with path /some_id: Bad file descriptor (os error 9)', src/sandbox.rs:717:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    sandbox::tests::update_shared_pidns

test result: FAILED. 110 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.09s

error: test failed, to rerun pass `--bin 
```

The problem cannot reproduce every time, but it often occurs in 10 times

Thanks @zvonkok for giving the initial bug report.

PR #6680, opened 2023-04-18
agent: Fix ut issue caused by fd double closed

Tim-Zhang:
Never ever try to close the same fd double times, even in a unit test.

A file descriptor is a number which will be reused, so when you close the same number twice you may close another file descriptor in the second time and then there will be an error 'Bad file descriptor (os error 9)' while the wrongly closed fd is being used.

Fixes: #6679

### DIFF
--- a/src/agent/src/main.rs
+++ b/src/agent/src/main.rs
@@ -442,9 +442,8 @@ mod tests {
             let msg = format!("test[{}]: {:?}", i, d);
             let (rfd, wfd) = unistd::pipe2(OFlag::O_CLOEXEC).unwrap();
             defer!({
-                // rfd is closed by the use of PipeStream in the crate_logger_task function,
-                // but we will attempt to close in case of a failure
-                let _ = unistd::close(rfd);
+                // XXX: Never try to close rfd, because it will be closed by PipeStream in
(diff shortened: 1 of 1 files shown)


gkurz (approved):
Wow this could certainly induce subtle bugs ! Thanks @Tim-Zhang !

gkurz:
BTW I'm a bit surprised to see that it is also _implementable_  in Rust... I would have expected the Rust compiler to forbid `rfd` from being passed to both `create_logger_task()` and the block inside the `defer!()` macro.

Tim-Zhang:
> BTW I'm a bit surprised to see that it is also _implementable_ in Rust... I would have expected the Rust compiler to forbid `rfd` from being passed to both `create_logger_task()` and the block inside the `defer!()` macro.

Unfortunately, there is no such magic currently 😄

lifupan (approved):

c3d:
> BTW I'm a bit surprised to see that it is also _implementable_ in Rust... I would have expected the Rust compiler to forbid `rfd` from being passed to both `create_logger_task()` and the block inside the `defer!()` macro.

To be fair, I think this is only true for something like a `RawFd`, where you are telling Rust "I know what I'm doing".

Issue #6682, opened 2023-04-19
Upgrade to Cloud Hypervisor v32.0 


likebreath:
This release has been tracked in our [roadmap project](https://github.com/orgs/cloud-hypervisor/projects/6) as iteration v32.0. The following user visible changes have been made:

### Increased PCI Segment Limit

The maximum number of PCI segments that can be used is now 96 (up from 16).

### API Changes

* The VmmPingResponse now includes the PID as well as the build details.  

### Notable Bug Fixes

* Ignore and warn TAP FDs sent via the HTTP request body
* Properly preserve and close valid FDs for TAP devices
* Only use `KVM_ARM_VCPU_PMU_V3` if available
* Only touch the tty flags if it's being used
* Fix seccomp filter lists for vhost-user devices
* The number of vCPUs is capped at the hypervisor maximum
* Fixes for TTY reset
* CPU topology fixes on MSHV
* Seccomp fixes for older distributions

Details can be found here: https://github.com/cloud-hypervisor/cloud-hypervisor/releases/tag/v32.0

PR #6683, opened 2023-04-19
Upgrade to Cloud Hypervisor v31.1

likebreath:
Cloud Hypervisor published a bug fix release v33.1 with the following issues being addressed:
* Ignore and warn TAP FDs sent via the HTTP request body
* Properly preserve and close valid FDs for TAP devices
* Only use `KVM_ARM_VCPU_PMU_V3` if available
* Only touch the tty flags if it's being used
* Fix seccomp filter lists for vhost-user devices

Details can be found here: https://github.com/cloud-hypervisor/cloud-hypervisor/releases/tag/v31.1

Fixes: #6682


likebreath:
Note: multiple CI jobs are failing, but they are apparently not related to the PR changes.

Tim-Zhang:
@likebreath yeah, I found that either.

PR #6686, opened 2023-04-19
gha: Create Mariner host as part of k8s tests

sprt:
This PR kickstarts the onboarding of Mariner onto the CI.

The current testing setup only supports running Kata on top of an Ubuntu host. This adds Mariner to the matrix of testable hosts for k8s tests, with Cloud Hypervisor as a VMM.

As preparation for the upcoming PR (#6839) that will change the actual test code (rather than workflow YAMLs), this also introduces a new file `setup.sh` that will be used to set host-specific parameters at test time.

Fixes: #6961

### DIFF
--- a/.github/workflows/run-k8s-tests-on-aks.yaml
+++ b/.github/workflows/run-k8s-tests-on-aks.yaml
@@ -17,10 +17,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        host_os:
+          - ubuntu
         vmm:
           - clh
           - dragonball
           - qemu
+        include:
+          - host_os: cbl-mariner
+            vmm: clh
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -46,7 +51,8 @@ jobs:
             -n "${{ github.event.pull_request.number }}-${{ github.event.pull_request.head.sha }}-${{ matrix.vmm }}-amd64" \
             -s "Standard_D4s_v5" \
             --node-count 1 \
-            --generate-ssh-keys
+            --generate-ssh-keys \
+            ${{ matrix.host_os == 'cbl-mariner' && '--os-sku mariner --workload-runtime KataMshvVmIsolation' }}
 
       - name: Install `bats`
         run: |
@@ -79,10 +85,11 @@ jobs:
           sleep 240s
 
           pushd tests/integration/kubernetes
-          sed -i -e 's|runtimeClassName: kata|runtimeClassName: kata-${{ matrix.vmm }}|' runtimeclass_workloads/*.yaml
+          bash setup.sh
           bash run_kubernetes_tests.sh
           popd
         env:
+          KATA_HOST_OS: ${{ matrix.host_os }}
           KATA_HYPERVISOR: ${{ matrix.vmm }}
 
       - name: Delete AKS cluster
--- a/tests/integration/kubernetes/run_kubernetes_tests.sh
+++ b/tests/integration/kubernetes/run_kubernetes_tests.sh
@@ -54,6 +54,10 @@ else
 	)
 fi
 
+if [ ${KATA_HOST_OS} == "cbl-mariner" ]; then
+	exit 0
+fi
+
 # we may need to skip a few test cases when running on non-x86_64 arch
 arch_config_file="${kubernetes_dir}/filter_out_per_arch/${TARGET_ARCH}.yaml"
 if [ -f "${arch_config_file}" ]; then
--- a/tests/integration/kubernetes/setup.sh
+++ b/tests/integration/kubernetes/setup.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+# Copyright (c) 2023 Microsoft Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+kubernetes_dir=$(dirname "$(readlink -f "$0")")
+
+set_runtime_class() {
+    sed -i -e "s|runtimeClassName: kata|runtimeClassName: kata-${KATA_HYPERVISOR}|" ${kubernetes_dir}/runtimeclass_workloads/*.yaml
+}
+
+main() {
+    set_runtime_class
+}
+
+main "$@"


tests/integration/kubernetes/setup.sh
```diff
@@ -34,4 +41,5 @@ set_image_path() {
 
 set_runtime_class
 set_hypervisor_path
+set_kernel_path
```
sprt, line 44: Might also need the following kernel params: `systemd.legacy_systemd_cgroup_controller=yes systemd.unified_cgroup_hierarchy=0`

tests/integration/kubernetes/setup.sh
```diff
@@ -23,6 +23,13 @@ set_hypervisor_path() {
     fi
 }
 
+set_kernel_path() {
+    if [[ "${KATA_HOST_OS}" = "cbl-mariner" ]]; then
+        mariner_kernel_path="/boot/vmlinux.bin"
+        find runtimeclass_workloads/*.yaml -exec yq write -i {} 'metadata.annotations[io.katacontainers.config.hypervisor.kernel]' "${mariner_kernel_path}" \;
```
sprt, line 29: FIXME: This won't have an effect because `kernel` isn't in `enable_annotations`.

tests/integration/kubernetes/setup.sh
```diff
@@ -16,6 +16,13 @@ set_runtime_class() {
     sed -i -e "s|runtimeClassName: kata|runtimeClassName: kata-${KATA_HYPERVISOR}|" runtimeclass_workloads/*.yaml
 }
 
+set_hypervisor_path() {
+    if [[ "${KATA_HOST_OS}" = "cbl-mariner" ]]; then
+        mariner_clh_path="/usr/bin/cloud-hypervisor"
+        find runtimeclass_workloads/*.yaml -exec yq write -i {} 'metadata.annotations[io.katacontainers.config.hypervisor.path]' "${mariner_clh_path}" \;
```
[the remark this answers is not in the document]
sprt: Actually, to address this and other comments, how about in `kata-deploy.sh` we configure Kata to 1) set `valid_hypervisor_paths = ["/*"]` and 2) set `enable_annotations = ".*"`?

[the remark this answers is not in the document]
fidencio: I've been going back and forth on this one.

Ideally we could have `/usr/bin/cloud-hypervisor` added as part of the `valid_hypervisor_path`, enable the annotation, and set it accordingly.

One thing that makes me worried here is if we have different versions of cloud-hypervisor.
Let's say the one we provide is 32.0, while the one supported on Mariner is 27.0.  There may be API changes in the runtime code that will simply make 27.0 not work.

cc @likebreath for his opinion on this one.

tests/integration/kubernetes/setup.sh
```diff
@@ -23,6 +23,13 @@ set_hypervisor_path() {
     fi
 }
 
+set_kernel_path() {
+    if [[ "${KATA_HOST_OS}" = "cbl-mariner" ]]; then
+        mariner_kernel_path="/boot/vmlinux.bin"
```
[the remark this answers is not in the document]
fidencio: This, as the hypervisor_path, will have to be done via annotation.

So, all the comments above also apply here.

tests/integration/kubernetes/setup.sh
fidencio: Also has to be handled via annotations.

tests/integration/kubernetes/setup.sh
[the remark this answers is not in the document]
sprt: > One thing that makes me worried here is if we have different versions of cloud-hypervisor.
Let's say the one we provide is 32.0, while the one supported on Mariner is 27.0. There may be API changes in the runtime code that will simply make 27.0 not work.

Cc @liuw - this PR deploys Mariner's CLH during CI instead of using the upstream version. Thoughts?

sprt:
@fidencio You got it right except for the rootfs part: since the test scripts seem hardcoded to use an image, I use an image to test Mariner as well - we don't expect to run into issues compared to initrd.

tests/integration/kubernetes/setup.sh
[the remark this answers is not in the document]
liuw: If you want to track upstream Kata then you should also track upstream Cloud Hypervisor.

fidencio:
> I'll ask the question: if we need to pass all these changes as annotatations, would adding a full kata config ("mariner") that is used by tests work better?

Hmm.  That would make sense.  Let me think a little bit about this (and also about having a drop-in file with the needed config), and I will get back to you Tomorrow.

fidencio:
So, what our CI does is:
1. build each one of the artefacts
2.  generate the kata-static.tar.xz
3. generate the kata-deploy payload from the kata-static.tar.xz
4. run the tests with the kata-deploy 

I would **not** be okay on having a "shipped by default" mariner specific config.  However, between 3 and 4, where we could easily generate a new container image where we'd add a mariner specific config.  That should work, without adding too much to the current process.

@sprt, @jepio, WDYT about this suggestion?

tests/integration/kubernetes/setup.sh
sprt: Not need, cgroups v1 is the default in Mariner.

sprt:
In the meantime, we've decided for now to keep the pinned CH version instead of using the one shipped with Mariner. So commit [tests: Use baked-in Cloud Hypervisor with Mariner](https://github.com/kata-containers/kata-containers/pull/6686/commits/bb8f610cb11e34b658a64ad5c9f18faf6465e3c9) would be reverted.

> So, what our CI does is:
> 
> 1. build each one of the artefacts
> 2. generate the kata-static.tar.xz
> 3. generate the kata-deploy payload from the kata-static.tar.xz
> 4. run the tests with the kata-deploy
> 
> I would **not** be okay on having a "shipped by default" mariner specific config. However, between 3 and 4, where we could easily generate a new container image where we'd add a mariner specific config. That should work, without adding too much to the current process.
> 
> @sprt, @jepio, WDYT about this suggestion?

Given that the only remaining config change is the kernel path, that approach seems like a big change just to support one use case. Also we expect all our kernel changes to be upstreamed eventually. So as a stopgap, passing an environment variable (`HOST_OS`?) to `kata-deploy.sh` to control `enable_annotations` seems easier. Thoughts? @fidencio @jepio

fidencio:
> Given that the only remaining config change is the kernel path, that approach seems like a big change just to support one use case. Also we expect all our kernel changes to be upstreamed eventually. So as a stopgap, passing an environment variable (`HOST_OS`?) to `kata-deploy.sh` to control `enable_annotations` seems easier. Thoughts?

I cannot think on any possible drawback with this approach, @sprt!

tests/integration/kubernetes/setup.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+set_runtime_class() {
+    sed -i -e "s|runtimeClassName: kata|runtimeClassName: kata-${KATA_HYPERVISOR}|" runtimeclass_workloads/*.yaml
+}
+
+set_runtime_class
```
wainersm, line 14: Nit: could you move this call to a `main()` method then `main $@`? This is a "standard" that we try hard for the scripts to follow. @jodh-intel and I will be pretty happy with that change :)

tests/integration/kubernetes/setup.sh
```diff
@@ -0,0 +1,14 @@
...
+# Copyright (c) 2023 Microsoft Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+set_runtime_class() {
+    sed -i -e "s|runtimeClassName: kata|runtimeClassName: kata-${KATA_HYPERVISOR}|" runtimeclass_workloads/*.yaml
```
wainersm, line 11: This sed is error prone because the paths to files are relative. Could you change it to use absolute path?

tests/integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -62,6 +62,10 @@ if [ ${KATA_HYPERVISOR} == "qemu-snp" ]; then
 	exit 0
 fi
 
+if [ ${KATA_HOST_OS} == "cbl-mariner" ]; then
```
wainersm, line 65: Is there a way to probe k8s to determine the node's OS so that we don't need to introduce another environment variable?

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -46,7 +51,8 @@ jobs:
             -n "${{ github.event.pull_request.number }}-${{ github.event.pull_request.head.sha }}-${{ matrix.vmm }}-amd64" \
             -s "Standard_D4s_v5" \
             --node-count 1 \
-            --generate-ssh-keys
+            --generate-ssh-keys \
+            ${{ matrix.host_os == 'cbl-mariner' && '--os-sku mariner --workload-runtime KataMshvVmIsolation' }}
```
wainersm, line 55: is mariner == cbl-mariner? or are we talking about different flavors?

I am asking because I have another suggestion. Rename `host_os` from cbl-mariner to mariner on line 27. Here we simply pass `--os-sku ${{ matrix.host_os }}` (I assume that `--os-sku ubuntu` is valid). This way adding a new host OS is easy. Then one extra line will be needed for passing the `--workload-runtime`.

wainersm:
I think I finished my review, @sprt, thanks!

I got a concern though. I'm a regular kata devel without access to Azure. I submit a PR that breaks the tests only in mariner, how can I reproduce the testing environment on my machine for debugging purpose?

tests/integration/kubernetes/setup.sh
sprt: Done

tests/integration/kubernetes/setup.sh
sprt: Done

tests/integration/kubernetes/run_kubernetes_tests.sh
sprt: We could do that by extracting the `osImage` field from the node description indeed, however the value returned by that field on Mariner isn't stable yet, so to avoid inadvertently breaking the CI with a new Mariner release, I think it's better to keep this environment variable for now.

.github/workflows/run-k8s-tests-on-aks.yaml
sprt: Yes, Mariner is just a shorthand for CBL-Mariner. I use `cbl-mariner` to match #6465 and because I'd find it confusing to refer to both `cbl-mariner` and `mariner` in the codebase. This way we can simply pass down the `host_os` value to determine which image/initrd to build.

sprt:
> I got a concern though. I'm a regular kata devel without access to Azure. I submit a PR that breaks the tests only in mariner, how can I reproduce the testing environment on my machine for debugging purpose?

Yes, with the move to GHA, reproducing the test environment is a use case that we'll want to investigate further for Mariner and Ubuntu. Another option in the future would be to preserve the cluster on failure and allow admins to temporarily give the author access to the cluster for debugging.

tests/integration/kubernetes/run_kubernetes_tests.sh
jepio: checking `/etc/os-release` contents would also work

jepio (approved):
lgtm

jepio:
Could you open an issue on this repo, and specify it as `Fixes: #<issue>` to that the commit message check passes. I don't think the static check accepts `Fixes` with a repo specification.

tests/integration/kubernetes/run_kubernetes_tests.sh
sprt: Yes, I believe that's how `osImage` is sourced. However, with the new [Azure Linux](https://aka.ms/azurelinux), it's still TBD what that file would contain longer term.

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -76,10 +82,11 @@ jobs:
         timeout-minutes: 30
         run: |
           pushd tests/integration/kubernetes
-          sed -i -e 's|runtimeClassName: kata|runtimeClassName: kata-${{ matrix.vmm }}|' runtimeclass_workloads/*.yaml
+          bash setup.sh
```
fidencio, line 85: Thanks for making this into a bash script!

fidencio (approved):
lgtm, thanks @sprt!

Issue #6695, opened 2023-04-21
kata-manager script fails to download and install containerd


amshinde:
While downloading containerd, we see the following error:
```
INFO: Downloading containerd release (latest version)
curl: (3) URL using bad/illegal format or missing URL
ls: cannot access '/tmp/tmp.5cTBbEaqSj/containerd-1.7.0-linux-amd64.tar.gz'$'\n''containerd-static-1.7.0-linux-amd64.tar.gz': No such file or directory
ERROR: Cannot determine containerd release file
 ```

amshinde:
On further investigation, I see this may be caused due to a recent change in the way containerd publishes its released packages.
There seems to be an additional static package included in the release as well. 

Debugging the script I found out that due to this change, we do not get a single download url for containerd, instead we get two from the `github_download_release()`. This function now returns:
`https://github.com/containerd/containerd/releases/download/v1.7.0/containerd-1.7.0-linux-amd64.tar.gz https://github.com/containerd/containerd/releases/download/v1.7.0/containerd-static-1.7.0-linux-amd64.tar.gz`
instead of a single url, causing the download to fail.

cc @jodh-intel

PR #6696, opened 2023-04-21
kata-manager: Fix containerd download

amshinde:
Newer containerd releases have an additional static package published. Because of this,  download_url contains two urls causing curl to fail. To resolve this, pick the first url from the containerd releases to download containerd.

Fixes: #6695


jodh-intel (approved):
Thanks @amshinde!

lgtm

jodh-intel:
This bug is stopping users install Kata, so please can I get one more ack here ;)

gkurz (approved):
LGTM. Thanks @amshinde !

PR #6706, opened 2023-04-23
feat(runtime-rs): introduce huge page mode to select VM RAM's backend

guixiongwei:
Currently, we support two backends: thp and hugetlbfs

Fixes: #6703

This commit allow us to specific the hugepage backend while enable hugepage.
Currently, we support two backends: thp and hugetlbfs, the default is hugetlbfs.


src/libs/kata-types/src/annotations/mod.rs
```diff
@@ -227,6 +227,9 @@ pub const KATA_ANNO_CFG_HYPERVISOR_MEMORY_PREALLOC: &str =
 /// A sandbox annotation to specify if the memory should be pre-allocated from huge pages.
 pub const KATA_ANNO_CFG_HYPERVISOR_HUGE_PAGES: &str =
     "io.katacontainers.config.hypervisor.enable_hugepages";
+/// A sandbox annotation to specify huge page mode.
+pub const KATA_ANNO_CFG_HYPERVISOR_HUGE_PAGES_MODE: &str =
+    "io.katacontainers.config.hypervisor.enable_hugepages_mode";
```
Apokleos, line 232: Thx.
What's about changing the boolean of `self.config.memory_info.enable_hugepages` to String ? 
And if so,  it'll avoid adding another config setting item for hugepage in configuration.

guixiongwei: Hi, I changed the self.config.memory_info.enable_hugepages to self.config.memory_info.hugepages of type string. Just change code part, If this change fine, I would modify the rest parts, like toml and docs.

src/libs/kata-types/src/annotations/mod.rs
```diff
@@ -226,7 +226,7 @@ pub const KATA_ANNO_CFG_HYPERVISOR_MEMORY_PREALLOC: &str =
     "io.katacontainers.config.hypervisor.enable_mem_prealloc";
 /// A sandbox annotation to specify if the memory should be pre-allocated from huge pages.
 pub const KATA_ANNO_CFG_HYPERVISOR_HUGE_PAGES: &str =
-    "io.katacontainers.config.hypervisor.enable_hugepages";
+    "io.katacontainers.config.hypervisor.hugepages";
```
c3d, line 229: It's called `enable_hugepages` [in the Go version](https://github.com/kata-containers/kata-containers/blob/3d8688f92e68630b15d1bf8b4888857fd8d472b1/src/runtime/virtcontainers/pkg/annotations/annotations.go#L170). What is the rationale for having a different name in the Rust version?

src/libs/kata-types/src/annotations/mod.rs
```diff
@@ -736,14 +736,9 @@ impl Annotation {
...
-                    KATA_ANNO_CFG_HYPERVISOR_HUGE_PAGES => match self.get_value::<bool>(key) {
-                        Ok(r) => {
-                            hv.memory_info.enable_hugepages = r.unwrap_or_default();
-                        }
-                        Err(_e) => {
-                            return Err(bool_err);
-                        }
-                    },
+                    KATA_ANNO_CFG_HYPERVISOR_HUGE_PAGES => {
+                        hv.memory_info.hugepages = value.to_string();
```
c3d, line 740: The local style is to validate annotations here, see for example line 681:

```
                        hv.machine_info.validate_entropy_source(value)?;
```

I believe that this should be done independently of a possibly later hypervisor-specific check as you do for Dragonball.

src/libs/kata-types/src/config/hypervisor/dragonball.rs
```diff
@@ -191,6 +192,15 @@ impl ConfigPlugin for DragonballConfig {
                     MIN_DRAGONBALL_MEMORY_SIZE_MB
                 ));
             }
+            if !db.memory_info.hugepages.is_empty()
```
c3d, line 195: That test does not seem Dragonball-specific, frankly. If that's the only test being done, it can probably move in the config parsing code.

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -551,13 +556,15 @@ pub struct MemoryInfo {
...
-    /// when you want to use vhost-user network stacks within the container. This will automatically
-    /// result in memory pre allocation.
+    /// The default is an Empty String which will disable this feature. The hugepage backend:
+    /// - hugetlbfs: This will result in the VM memory being allocated using hugetlbfs backend.
+    /// - thp: This will result in the VM memory being allocated using transprant huge page backend.
+    /// This is useful when you want to use vhost-user network stacks within the container.
+    /// This will automatically result in memory pre allocation.
     #[serde(default)]
-    pub enable_hugepages: bool,
+    pub hugepages: String,
```
c3d, line 567: I believe that `String` is the wrong type here. It should be an three-state enum, which would be much more efficient, clearer as to the intent, and less error prone.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -188,8 +189,12 @@ impl DragonballInner {
 
     fn set_vm_base_config(&mut self) -> Result<()> {
         let serial_path = [&self.run_dir, "console.sock"].join("/");
-        let (mem_type, mem_file_path) = if self.config.memory_info.enable_hugepages {
-            (String::from(HUGETLBFS), String::from(DEV_HUGEPAGES))
+        let (mem_type, mem_file_path) = if !self.config.memory_info.hugepages.is_empty() {
+            match self.config.memory_info.hugepages.as_str() {
```
c3d, line 193: This clearly shows that you expect an enum in the config

c3d (changes requested):
Overall, good idea, but two concerns justifying a request to rework this a bit:

1. Backward compatibility of annotations with Go runtime suggests you should still support `enable_hugepages`. I give a suggestion how to support both your new syntax and the old one relatively easily.
2. Configuration as string instead of an enum.

c3d:
> Overall, good idea, but two concerns justifying a request to rework this a bit:
> 
> 1. Backward compatibility of annotations with Go runtime suggests you should still support `enable_hugepages`. I give a suggestion how to support both your new syntax and the old one relatively easily.
> 2. Configuration as string instead of an enum.

@Apokleos please note that I went against your recommendation in my own review. I think that `String` is a bad choice here, it should really be a tri-state enum (we are not getting a path or anything like that)

src/libs/kata-types/src/annotations/mod.rs
yipengyin: https://github.com/kata-containers/kata-containers/blob/7fdaab49bcce816cd0f928677cf67dcea9548ec3/src/libs/kata-types/src/config/hypervisor/mod.rs#L600-L612

src/libs/kata-types/src/annotations/mod.rs
studychao: I agree with this comment, I think we should keep backward compatible in Rust runtime.
@c3d has presented an easy way to do so.

src/libs/kata-types/src/annotations/mod.rs
studychao: That's a good place to start I think

src/libs/kata-types/src/config/hypervisor/mod.rs
studychao: Agree , an enum would be clearer.

studychao (commented):
Thanks, a few comments.

studychao:
Also, there is some CI errors, please help fix them.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -188,8 +189,12 @@ impl DragonballInner {
...
+                HUGE_PAGE_MODE_THP => (String::from(HUGE_SHMEM), String::from("")),
```
studychao, line 194: Besides `hugeshmem`, Dragonball could also use `hugeanon`, `hugemmap` and `hugememfd` to trigger transparent huge page.

studychao: image

yipengyin: THP couldn't represent all of these types. Should we expose all of these memory backend to runtime?

Apokleos:
> > Overall, good idea, but two concerns justifying a request to rework this a bit:
> > 
> > 1. Backward compatibility of annotations with Go runtime suggests you should still support `enable_hugepages`. I give a suggestion how to support both your new syntax and the old one relatively easily.
> > 2. Configuration as string instead of an enum.
> 
> @Apokleos please note that I went against your recommendation in my own review. I think that `String` is a bad choice here, it should really be a tri-state enum (we are not getting a path or anything like that)
>

Hi @c3d Thx first, and sorry  late to reply.
And, I don't think adding extra param is a good way.  That was my initial idea.

IMO, If one parameter does not fully describe it, it means that there is a problem with this parameter. I think instead of adding an extra param to represent it, it's better to use this one to clearly express the whole thing. There is no need to separate multiple steps to do it.
I give the suggestion based on the above idea.

src/libs/kata-types/src/config/hypervisor/mod.rs
Apokleos: Yeah, I agree. 
Enum type maybe is a better one?

src/libs/kata-types/src/annotations/mod.rs
Apokleos: > It's called `enable_hugepages` [in the Go version](https://github.com/kata-containers/kata-containers/blob/3d8688f92e68630b15d1bf8b4888857fd8d472b1/src/runtime/virtcontainers/pkg/annotations/annotations.go#L170). What is the rationale for having a different name in the Rust version? I understand that you have a string now, but I'm concerned that someone passing `enable_hugepages=true` would get no effect with the Rust runtime.
> 
> My recommendation would be to accept both `enable_hugepages` and `hugepages`, and at least for the latter, accept `true` (same as `hugetlbfs`) and `false` as well as `thp`, `hugetlbfs` and `none` (same as `false`). This would provide a behaviour that is compatible both with your current preference as well as with historical input.
>
Using both `enable_hugepages` and other parameters such as `hugepages` at the same time may also lead to user misconfiguration. Following the usage habit of the kata in golang, only configuring `enable_hugepages` but forgetting to configure `hugepages` may also result in the functionality not achieving the expected effect.

So, I agree with the idea that `hugepages` with Enum and make it compatible with both current and historic input.

guixiongwei: @Apokleos  I think it is hard to provide a behaviour that keep backward compatible in Rust runtime in one parameter. In the current version , enable_hugepages is bool type(true and false) but the new parameter should provide string type("thp" "hugetlbfs"). These two types cannot be configured as one parameter.
IMO we need two parameter to make it compatible with both current and historic input.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
studychao: make sense. Then the current code is good for me.

Apokleos (approved):
LGTM

studychao:
For the annotations, in light of everyone's opinions, let's support two annotations to keep backward-compability:

- `enable_hugepages`:  bool, default false. true will enable hugetlbfs

- `hugepages`: enum, default None. accept thp, hugetlbfs and none 

@guixiongwei @yipengyin Please help update this PR :)

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -551,13 +556,15 @@ pub struct MemoryInfo {
...
 
-    /// Enable huge pages for VM RAM, default false
+    /// Huge pages support for VM RAM.
     ///
-    /// Enabling this will result in the VM memory being allocated using huge pages. This is useful
-    /// when you want to use vhost-user network stacks within the container. This will automatically
-    /// result in memory pre allocation.
+    /// The default is an Empty String which will disable this feature. The hugepage backend:
+    /// - hugetlbfs: This will result in the VM memory being allocated using hugetlbfs backend.
+    /// - thp: This will result in the VM memory being allocated using transprant huge page backend.
```
c3d, line 563:
```suggestion
    /// - thp: This will result in the VM memory being allocated using transparent huge page backend.

```

src/libs/kata-types/src/annotations/mod.rs
c3d: > @Apokleos I think it is hard to provide a behaviour that keep backward compatible in Rust runtime in one parameter. In the current version , enable_hugepages is bool type(true and false) but the new parameter should provide string type("thp" "hugetlbfs"). These two types cannot be configured as one parameter. IMO we need two parameter to make it compatible with both current and historic input.

Since we parse the text input, we can parse `true` as meaning `hugetlbfs`.

c3d: > Using both `enable_hugepages` and other parameters such as `hugepages` at the same time may also lead to user misconfiguration. Following the usage habit of the kata in golang, only configuring `enable_hugepages` but forgetting to configure `hugepages` may also result in the functionality not achieving the expected effect.

What kind of misconfiguration are you thinking about? To me, this is really an optimization in how you allocate memory. If you say "enable hugepages" and get the same thing as before, but not the new fancy stuff, that's not misconfiguration. It's not a problem that you need to say something new to enable the new option. So I do not understand the objection.

c3d (commented):
@studychao asked me to take another look, but I don't see a code change. So I made a few comments, but I'm not entirely sure what the question was ;-)

c3d:
> This commit allow us to specific the hugepage backend while enable hugepage.
nit: "allows us", "while enable" -> "when enabling"


> Currently, we support two backends: thp and hugetlbfs, the default is hugetlbfs.

I believe that the commit message is misleading. The commit actually changes the configuration, notably the spelling of `enable_hugepages`. I'm frankly not in favor of that.

At the very least, if you _do_ change the spelling and reject the old configuration, you need to change the documentation accordingly, and explain in your commit message the rationale for the change, something like: "The semantics of `enable_hugepages` is broken because <insert reason>. Someone setting it to `true` would expect <expected-result> and instead gets <unexpected-result>. For that reason, we need to change it to a string that lets us control the setting correctly. Setting it to <setting 1> has <effect 1> by selecting <hypervisor 1 option 1> on <hypervisor 1>, <hypervisor 2 option 1> on <hypervisor 2> and <other cases>. Setting it to <setting 2> has <effect 2> by selecting <hv1 option 2> on <hv1> and <hv2 option 2> on <hv2>. As a result, users will get <better performance | increased usability | more consistent behavior...>. The documentation is modified accordingly, to indicate that `enable_hugepages` is no longer supported and explaining how to convert existing options to the new model.

The other path is to _not_ change the spelling of the optoin, but to extend the range of accepted input. Today, it's only `true` or `false`. You could add finer-grained control either by adding a second option that provides the missing information (e.g. `hugepage_type`) or by accepting new values as input to the existing one (e.g. `enable_hugepages = thp`). That would make it easier to document the change.

Hope this helps.

c3d:
@guixiongwei @studychao I am not necessarily monitoring this PR super-closely. Please ping me on Slack so that I don't delay things forever, and sorry about the delay so far.

yipengyin:
Hey guys @c3d @Apokleos @studychao . Since @guixiongwei is occupied by other issue. I have tweaked this patch a little bit:
1. Now, the huge page type is an enum.
2. I use another config item to select the huge page type. I believe its semantics is more clear. 
    Of course, it's possible to extend the item`enable_hugepages` to accept `true`, `false`, `thp` and `hugetlbfs`. But I think it's a little bit weird.
3. If `enable_hugepages` is `false`, it will disable all kind of huge page.

What do you think of this patch.

BTW, the cargo lock seems no to be up-to-update. I'm not sure how to update it.

src/libs/kata-types/src/annotations/mod.rs
```diff
@@ -740,14 +743,28 @@ impl Annotation {
...
                         Err(_e) => {
                             return Err(bool_err);
                         }
                     },
+                    KATA_ANNO_CFG_HYPERVISOR_HUGEPAGE_TYPE => {
+                        match self.get_value::<HugePageType>(key) {
+                            Ok(r) => {
+                                hv.memory_info.hugepage_type = r.unwrap_or_default();
+                            }
+                            Err(_e) => {
```
studychao, line 760: is it possible to print out the error?

studychao (commented):
overall LGTM, just two comments.

studychao:
Also, could you update the commit message to cover that we introduce the implementation of two annotations in this commit with annotation names and how they work together.

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -577,12 +597,18 @@ pub struct MemoryInfo {
...
-    /// result in memory pre allocation.
+    /// Enabling this will result in the VM memory being allocated using huge pages.
+    /// Its backend type is specified by item "hugepage_type"
     #[serde(default)]
     pub enable_hugepages: bool,
 
+    /// Select huge page type, default "hugetlbfs"
+    /// Following huge types are supported:
+    /// - hugetlbfs
+    /// - thp
```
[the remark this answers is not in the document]
yipengyin: `Select huge page type, default "hugetlbfs"`
I already stated the default type here.

src/libs/kata-types/src/annotations/mod.rs
yipengyin: done

yipengyin:
> Also, could you update the commit message to cover that we introduce the implementation of two annotations in this commit with annotation names and how they work together.

> Besides, seems like you forget to update the Cargo.lock file for agent.

All of them are done

studychao:
@c3d Please take a look

quanweiZhou (approved):
Thanks, @guixiongwei, LGTM.

c3d (approved):
Looks good to me in this iteration. Thanks a lot, @guixiongwei and @yipengyin, and sorry for the time it took to review.

Issue #6640, opened 2023-04-10
CC | Configurable `Policy.json`, sigstore config for simple signing and registry credential file


Xynnn007:
**Which feature do you think can be improved?**

Now we support KBS Resource URI to import/retrieve different resources from KBS side, including cosign public keys, pgp public keys, etc. However three important things are ignored:
- [image security policy file](https://github.com/confidential-containers/image-rs/blob/main/test_data/simple-signing-scheme/security_policy.json): so-called `policy.json`. This policy defines the allowlis, denylist, signature verification rules of the container images to pull. It will be fetched from the specified URI and used when `security_validate` is enabled, as [here](https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/image_rpc.rs#L191). Now its URI by default is `kbs:///default/security-policy/test` as [here](https://github.com/confidential-containers/image-rs/blob/main/src/config.rs#L17), which means that by default this resource should be fetched from KBS with this uri.
- [sigstore config for simple signing](https://github.com/confidential-containers/image-rs/blob/main/test_data/simple-signing-scheme/sigstore_config.yaml): This yaml configuration file defines where to find the signature files for [simple signing scheme](https://github.com/confidential-containers/image-rs/tree/main/src/signature/mechanism/simple). If simple signing is involved in the `image security policy file` (`"type": "signedBy"`), this file will be fetched from the specified URI. By default the URI is [`kbs:///default/sigstore-config/test`](https://github.com/confidential-containers/image-rs/blob/main/src/config.rs#L24), which means that if simple signing is to be used, a corresponding sigstore config yaml should be prepared with such a URI.
- [registry auth file](https://github.com/confidential-containers/attestation-agent/blob/main/src/kbc_modules/sample_kbc/auth.json): This file contains credential information to access private container registries. By default the URI is ["kbs:///default/credential/test"](https://github.com/confidential-containers/image-rs/blob/main/src/config.rs#L32), which means it will be fetched from KBS.

Image-rs now provides an config API (`file_paths` parameter in [this structure](https://github.com/confidential-containers/image-rs/blob/main/src/config.rs#L36)) to change URIs of the mentioned resources, which means that they can be either other KBS URIs, or from the guest filesystem with URI like `"file:///etc/example.conf"`. However, we have not implement this in kata-agent.

**How can it be improved?**

To support configurable URIs of these resources, the basic idea is
[...]

PR #6707, opened 2023-04-24
agent: add container launch control parameters from kernel commandline

Xynnn007:
Hi, I'm not sure whether it is still ok to make PR to CCv0 branch. If not, please tell me the correct thing to do.

##### About this PR

Add three commdline parameters to control the container launch process

`image_policy_file` is the URI of the image security policy file which is used to control the allowlist and denylist of the containers to be run inside the guest.

`image_registry_auth_file` is the URI of the auth file which is needed to access an private registry.

`simple_signing_sigstore_config` is a configuration file that shows where the signature files are when simple signing is enabled.

All the config items support both `kbs://..` scheme, `file://..` scheme, or a direct absolute path `/...`, which means either the file is to be fetched from a remote KBS that `aa_kbc_params` points to, or from the local filesystem of the guest.

Fixes: #6640

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -31,6 +31,9 @@ const HTTPS_PROXY: &str = "agent.https_proxy";
 const NO_PROXY: &str = "agent.no_proxy";
 const ENABLE_DATA_INTEGRITY: &str = "agent.data_integrity";
 const ENABLE_SIGNATURE_VERIFICATION: &str = "agent.enable_signature_verification";
+const IMAGE_POLICY_FILE: &str = "agent.image_policy";
+const IMAGE_REGISTRY_AUTH_FILE: &str = "agent.image_registry_auth";
+const SIMPLE_SIGNING_SIGSTORE_CONFIG: &str = "agent.simple_signing_sigstore_config";
 
 const DEFAULT_LOG_LEVEL: slog::Level = slog::Level::Info;
 const DEFAULT_HOTPLUG_TIMEOUT: time::Duration = time::Duration::from_secs(3);
@@ -89,6 +92,9 @@ pub struct AgentConfig {
     pub no_proxy: String,
     pub data_integrity: bool,
     pub enable_signature_verification: bool,
+    pub image_policy_file: String,
+    pub image_registry_auth_file: String,
+    pub simple_signing_sigstore_config: String,
 }
 
 #[derive(Debug, Deserialize)]
@@ -110,6 +116,9 @@ pub struct AgentConfigBuilder {
     pub no_proxy: Option<String>,
     pub data_integrity: Option<bool>,
     pub enable_signature_verification: Option<bool>,
+    pub image_policy_file: Option<String>,
+    pub image_registry_auth_file: Option<String>,
+    pub simple_signing_sigstore_config: Option<String>,
 }
 
 macro_rules! config_override {
@@ -177,6 +186,9 @@ impl Default for AgentConfig {
             no_proxy: String::from(""),
             data_integrity: false,
             enable_signature_verification: true,
+            image_policy_file: String::from(""),
+            image_registry_auth_file: String::from(""),
+            simple_signing_sigstore_config: String::from(""),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Xynnn007:
Thanks, Tobin!

> One suggestion would be to simplify the interface by just supplying repository/id from the command line. These could be used to generate the three URIs. Do you think that would be simpler? It is slightly less flexible.

To my understand you might want to every `repository` or `id` to play the role as an identity of `pod`. In my opinion the kernel parameter here to specify these configurations are underlying things and they'd better be flexible enough. For the convenience of users, we can consider a CLI tool or something that receives a simple request from user and translate that request into a complex one.

My core point of view is that the underlying things should be as extensible as possible, so that only encapsulation is required in the subsequent development process, so as to avoid repeated development at the bottom layer in order to expand a certain feature.

Xynnn007:
> Maybe we could implement this in the shim? Rather than setting the kernel params directly the user could set an annotation that specifies some kind of pod identity. The shim would use this to formulate the path for the configs and add them to the kernel params.

I like this idea. Maybe the process can be described as the following:
1. shim parses the identity of the pod and the address of KBS from address. Let's assume the identity is "A" and the address "addr"
2. shim derives the three parameters: `kbs://add/<repo>/security-policy/A`, `kbs://add/<repo>/credential/A` and `kbs://add/<repo>/sigstore-config/A`. These three parameters can then be injected by kernel commandline.

Up to know, AS can already know the requested policy and the identity of the pod. We can think whether we also need to embed the identity into CONFIGID-like fields before launch the VM.

stevenhorsman:
Thanks Ding, this PR looks good and would help a use-case that we've heard in the peer pods community where people want to effectively use the offline_fs_kbc for image pull credentials, but cc_kbc for pulling runtime secrets/decryption keys from remote attestation.

From what I understand this would allow us to just embed the cc_kbc module built attestation-agent into the peer pod VM image and then rather than use the offline_fs_kbc writing our credentials to `/etc/aa-offline_fs_kbc-resources.json`, we can update our dynamic image pull secret injector to write to somewhere like `/etc/containers/auth.json` and then set `image_registry_auth_file=file:///etc/containers/auth.json` in `/etc/agent-config.toml` in the peer pod VM image.

I guess with the remote-hypervisor we don't use annotations, or the shim to set the kernel_params, so Tobin's question isn't so important. I do prefer the idea of the annotation over overloading the kernel_params even more than we already do though!

Xynnn007:
@stevenhorsman  Sounds great to me Steve!

@fitzthum Hey Tobin, can we get this PR approved/merged now? It can be done in future about the derivation of the kernel commandline.

stevenhorsman (approved):
LGTM - thanks @Xynnn007.
Just one question - will we need to bump image-rs in order to test this, or is the 0.6 release version compatible with this?

Xynnn007:
> LGTM - thanks @Xynnn007.
> Just one question - will we need to bump image-rs in order to test this, or is the 0.6 release version compatible with this?

v0.6.0 is compatible with this.

stevenhorsman:
> > LGTM - thanks @Xynnn007.
> > Just one question - will we need to bump image-rs in order to test this, or is the 0.6 release version compatible with this?
> 
> v0.6.0 is compatible with this.

Thanks - Do you think it is worth adding some CI tests for this behaviour and then we can repurpose them to check the annotation when that gets added?

Xynnn007:
> > > LGTM - thanks @Xynnn007.
> > > Just one question - will we need to bump image-rs in order to test this, or is the 0.6 release version compatible with this?
> > 
> > 
> > v0.6.0 is compatible with this.
> 
> Thanks - Do you think it is worth adding some CI tests for this behaviour and then we can repurpose them to check the annotation when that gets added?

Ci test sounds good to me. With the base CI we can extend it into different aims, including the stated annotation test. Furthermore, any potential test in future that would have a specific parameter in kernel parameter would benefit from this, like the dm-verity for encrypted rootfs.

stevenhorsman:
> Ci test sounds good to me. With the base CI we can extend it into different aims, including the stated annotation test. Furthermore, any potential test in future that would have a specific parameter in kernel parameter would benefit from this, like the dm-verity for encrypted rootfs.

Would you be able to raise an issue on the tests repo for this? I won't have capacity to look at it myself in the short-term, but might be able to find someone else to help and otherwise it would be good to track so I don't forget about it.

fitzthum (approved):
LGTM. This is against CCv0 so we may need to introduce again in main before too long.

fitzthum:
One thing to mention is that we may need to think about the best way to handle some of these features from the KBS side. For instance, we are introducing the ability for different pods to request different resources, but we don't really have any logic to group resources together in the KBS.

stevenhorsman:
I flagged this up on last weeks CoCo call as available to review and there have been no takers, so I think we can merge it in now.

Issue #5343, opened 2022-10-06
kata-ctl: iptables: Implement with tests


jodh-intel:
Implement the `kata-runtime iptables` command with unit tests.

See: https://github.com/kata-containers/kata-containers/issues/4499.

jodh-intel:
This will require implementing [`handle_iptables()`](https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/ops/check_ops.rs#L115) along the lines of [`set_iptables()`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs#L281).

pembek01:
@jodh-intel did you happen to find anything out to help me with this issue?

jodh-intel:
Hi @pembek01 - I put a comment on slack about this issue and #6230. Looking at these issues, I think it would be helpful if I gave a presentation on the Kata architecture and explained a bit about how these two tools are structured. I plan to do that on Friday 10th (tomorrow).

One other link that might be useful to look at today is the golang implementation:

- https://github.com/kata-containers/kata-containers/blob/main/src/runtime/cmd/kata-runtime/kata-iptables.go#L101..L118

jodh-intel:
This breaks down to the following tasks:

- Add a new `iptables` sub-command to `kata-ctl` so we can run `kata-ctl iptables get --sandbox-id foo` and `kata-ctl iptables set --sandbox-id foo file`.

  This will be equivalent to `kata-runtime iptables [get|set] ...`.

- For the `kata-ctl iptables set`, you'll need to read a file specified by the user. I'd recommend using
  `std::fs::read_to_string(filename)`.

- You'll then need to connect to the agent. The `agent-ctl` tool does this, but in an very low-level way so instead, you'll need to create an agent connection like `runtime-rs` does in [`new_agent()`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/runtimes/virt_container/src/lib.rs#L153):
  ```
  let agent: KataAgent = KataAgent::new(...);
  ```

- You'll then need to call `agent.get_ip_tables()` and `agent.set_ip_tables()` methods.

[...]

PR #6708, opened 2023-04-24
kata-ctl: added iptables subcommand file - WIP - #5343

pembek01:
Added a sub command file named "iptables.rs" to kata-ctl so users can get and set ip-tables rules.This is a Work in Progress yet but I believe I am going in the right direction. I commented on most of the lines just for your convenience so you can understand my thought process behind everything. If this is code is correct I will delete some of the comments that are not necessary.

Fixes: #5343 


src/tools/kata-ctl/src/iptables.rs
```diff
@@ -0,0 +1,153 @@
...
+    if let Err(e) = new_main() {
+        eprintln!("Error: {}", e);
+        process::exit(1);
+    }
+}
+
+// Define the function signature. It returns `Result<(), Box<dyn std::error::Error>>`,
+// which means it can either return an `Ok(())` value indicating success or an `Err` value
+// containing a boxed error type that implements the `std::error::Error` trait.
+fn new_main() -> Result<(), Box<dyn std::error::Error>> {
```
jodh-intel, line 22: You don't need to create any main functions as that's already handled by https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/main.rs.

What you need to do is implement the `handle_iptables()` function called here:

https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/main.rs#L34

See my comment here: https://github.com/kata-containers/kata-containers/issues/5343#issuecomment-1446709040.

You'll also need to add the command-line options/args, but again, only by adding to what is already in `src/main.rs` (see the `IptablesCommand` structure).

What I would do though is move the existing "stub" / "nop" `handle_iptables()` function from `src/ops/check_ops.rs` into your new `src/iptables.rs` file (it got added to the `check_ops.rs` file originally by mistake I think ;)

You'll also need to add `mod iptables;` to `src/main.rs` as currently this file is _not_ being built as part of `kata-ctl` as it isn't being referenced by any of the existing code (and in fact fails to compile due to the "floating match" (see comment below).

src/tools/kata-ctl/src/iptables.rs
```diff
@@ -0,0 +1,153 @@
...
+                ),
+        )
+        .get_matches(); // Parse the command line arguments and return a `clap::ArgMatches` struct.
+
+    // Return an `Ok` value indicating success.
+    Ok(())
+}
+
+//checking for subcommand entered form user 
+    match matches.subcommand() {
```
jodh-intel, line 75: This doesn't seem to be part of a function? I think you probably want to add the logic to the `handle_iptables()` fuction.

jodh-intel (commented):
Thanks for raising @pembek01!

The good news is that you don't need the code to be this complcated ;) I've added some comments that hopefully will help you simplify the PR further.

/cc @tg5788re.

jodh-intel:
Closing this PR as you've raised a new one #6716

Issue #6717, opened 2023-04-25
support keep_abnormal in runtime-rs


mzweilz:
In some debugging scenarios, we may want to keep the runtime process or abnormal information. So we add a configuration keep_abnormal to make it work.

PR #6718, opened 2023-04-25
runtime-rs: support keep_abnormal in toml config

mzweilz:
This patch adds keep_abnormal in runtime config. If keep_abnormal = true, it means that 1) if the runtime exits abnormally, the cleanup process will be skipped, and 2) the runtime will not exit even if the health check fails.

This option is typically used to retain abnormal information for debugging and should NOT be enabled by default.

Fixes: #6717

### DIFF
--- a/src/libs/kata-types/src/config/runtime.rs
+++ b/src/libs/kata-types/src/config/runtime.rs
@@ -130,6 +130,12 @@ pub struct Runtime {
     /// Vendor customized runtime configuration.
     #[serde(default, flatten)]
     pub vendor: RuntimeVendor,
+
+    /// If keep_abnormal is enabled, it means that 1) if the runtime exits abnormally, the cleanup process
+    /// will be skipped, and 2) the runtime will not exit even if the health check fails.
+    /// This option is typically used to retain abnormal information for debugging.
+    #[serde(default)]
+    pub keep_abnormal: bool,
 }
 
 impl ConfigOps for Runtime {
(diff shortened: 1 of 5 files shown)


src/libs/kata-types/src/config/runtime.rs
```diff
@@ -130,6 +130,12 @@ pub struct Runtime {
     /// Vendor customized runtime configuration.
     #[serde(default, flatten)]
     pub vendor: RuntimeVendor,
+
+    /// If keep_abnormal is enabled, it means that 1) if the runtime exits abnormally, the cleanup process
+    /// will be skipped, and 2) the runtime will not exit even if the health check fails.
+    /// This option is typically used to retain abnormal information for debugging.
+    #[serde(default)]
+    pub keep_abnormal: bool,
```
Tim-0731-Hzt, line 138: sugguest adding this to configuration.toml

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -190,9 +191,16 @@ impl RuntimeHandlerManager {
         let sender = inner.msg_sender.clone();
         let sandbox_state = persist::from_disk::<SandboxState>(&inner.id)
             .context("failed to load the sandbox state")?;
+
+        let config = if let Ok(spec) = load_oci_spec() {
```
YushuoEdge, line 195: Please add some comments on where the spec comes from.

YushuoEdge (commented):
Thanks, @mzweilz !

src/libs/kata-types/src/config/runtime.rs
mzweilz: done

Tim-0731-Hzt (approved):

quanweiZhou (approved):

PR #6721, opened 2023-04-25
Improve TestQemuCreateVM coverage

tklengyel:
In this PR we are improving the code-coverage of virtcontainers/qemu.go, issue https://github.com/kata-containers/kata-containers/issues/267

The additional coverage is only on the paths related to CreateVM. Additional tests are still required so the issue shouldn't yet be closed or the PR merged yet.

Credit PR to Hackathon Team3


src/runtime/virtcontainers/qemu_test.go
```diff
@@ -76,40 +76,112 @@ func TestQemuKernelParameters(t *testing.T) {
...
+		assert.NoError(err, msg)
+
+		if d.configMatch {
+			assert.Exactly(d.config, q.config, msg)
+		}
+
+		mem := q.GetTotalMemoryMB(context.Background())
+		assert.True(mem > 0)
+
+		err = q.canDumpGuestMemory("/tmp")
```
jodh-intel, line 162: If this dump works, the file in `/tmp` could be very large. Please create a temporary directory for this. The doc below shows how to create a tmpdir that is automatically removed once the test finishes:

- https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md#temporary-files

src/runtime/virtcontainers/qemu_test.go
```diff
@@ -76,40 +76,112 @@ func TestQemuKernelParameters(t *testing.T) {
...
+	data[11].config.BootFromTemplate = true
+	data[12].config.BootToBeTemplate = true
+	data[12].config.SharedFS = config.VirtioFS
+	data[12].expectError = true
+	data[13].config.FileBackedMemRootDir = "/tmp/xyzabc"
+	data[13].config.HugePages = true
+	data[14].config.SharedFS = config.VirtioFS
+	data[15].config.BlockDeviceDriver = ""
+	data[16].config.SharedFS = config.VirtioFSNydus
+	data[17].config.VMid = "testSandbox"
```
jodh-intel, line 134: This is admittedly very compact, but I find it very difficult to read.

Please could you change this to match more like what is shown in https://github.com/kata-containers/kata-containers/blob/main/docs/Unit-Test-Advice.md#golang-table-driven-tests by using partial struct initialisation:

```go
data := []testData{
    {},
    {config: HypervisorConfig{SeccompSandbox: "enable=1"}},
    {config: HypervisorConfig{ImagePath: testQemuImagePath}},

    // ...
}
```

This will also allow you to remove the `for` loop with the hard-coded upper index bound above, so the code will be safer and easier to maintain.

src/runtime/virtcontainers/qemu_test.go
tklengyel: // canDumpGuestMemory check if can do a guest memory dump operation.
// for now it only ensure there must be double of VM size for free disk spaces
func (q *qemu) canDumpGuestMemory(dumpSavePath string) error {

src/runtime/virtcontainers/qemu_test.go
```diff
@@ -76,40 +76,112 @@ func TestQemuKernelParameters(t *testing.T) {
...
+	type testData struct {
+		config      HypervisorConfig
+		expectError bool
+		configMatch bool
+	}
+
+	var data []testData
+
+	// create test configs with default settings
+	for i := 0; i < 18; i++ {
```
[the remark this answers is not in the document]
tklengyel: Why is that a problem?

src/runtime/virtcontainers/qemu_test.go
tklengyel: This is all fine if all you are setting is a single config entry. It becomes ugly once you start adding more config options:

```
    data := []testData{
        { false, true, newQemuConfig() },
        { false, false, HypervisorConfig{ SeccompSandbox = "enable=1" } },
        { false, false, HypervisorConfig {  InitrdPath = "",
                                            ImagePath = testQemuImagePath } },
        { false, false, HypervisorConfig { Debug = true },

    }
```

I find this format to be a lot harder to read then default-initializing all test-configs and then just changing options by index. With partial init only you are also losing the check `assert.Exactly(d.config, q.config, msg)` because it fails for every one except the default one generated by `newQemuConfig()`. Furthermore, without default settings being present in every config the code coverage is lower because the code bails earlier, so you end up having to define the same settings over and over and over for every config further making it harder to understand which option is actually being tested.

jodh-intel: True, but when we start adding large chunks we tend to create a few variables to make this clearer. For example see how we use `longPath`, `longestPath` and `pathTooLong` here:

- https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/utils/utils_test.go#L276..L292

src/runtime/virtcontainers/qemu_test.go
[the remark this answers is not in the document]
jodh-intel: Because it's fragile code that is hard-coding a magic number. That number and the `data[index]` block below this needs to be kept in-sync with this magic number.

In Kata, we try to avoid this as it makes maintenance more difficult.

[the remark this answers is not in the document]
tklengyel: I could move the number to a standalone variable like "NumberOfTestConfigs" (in C I would just use a `#define`, not sure what's the go equivalent). The other altnernative is really to just have 18 lines repeating the same thing which I find silly.

src/runtime/virtcontainers/qemu_test.go
tklengyel: Sure, I can do that, IMHO this approach adds more lines than necessary but if you think it makes it more readable I won't argue

jodh-intel (approved):
Thanks @tklengyel!

lgtm

src/runtime/virtcontainers/qemu_test.go
```diff
@@ -636,3 +753,27 @@ func TestQemuSetConfig(t *testing.T) {
...
+			VMStorePath:  sandbox.store.RunVMStoragePath(),
+			RunStorePath: sandbox.store.RunStoragePath(),
+		},
+		virtiofsDaemon: &virtiofsdMock{},
+	}
+
+	err = q.CreateVM(context.Background(), sandbox.id, network, &sandbox.config.HypervisorConfig)
+	assert.NoError(err)
+
+	err = q.StartVM(context.Background(), 10)
```
amshinde, line 777: Doesnt this actually start the qemu process? Make sure there are no processes left behind after the test is executed.

tklengyel: No, this should actually error out with `failed to launch qemu: fork/exec /tmp/vc-tmp-623275751/hypervisor: permission denied`. That's why we have `assert.Error(err)`.

amshinde: My bad, overlooked the assert for error.

amshinde (approved):

amshinde:
@tklengyel The s390x CI that tests the unit tests on a s390 machine is repeatedly failing. I see this in the error log:
```
13:19:03 time="2023-05-11T20:18:47Z" level=error msg="Failed to connect to QEMU instance" error="dial unix testSandbox/qmp.sock: connect: no such file or directory" source=virtcontainers/hypervisor subsystem=qemu
13:19:03 --- FAIL: TestQemuCreateVM (0.04s)
13:19:03     qemu_test.go:198: 
13:19:03         	Error Trace:	/home/jenkins/workspace/kata-containers-2.0-ubuntu-s390x-unit-PR/go/src/github.com/kata-containers/kata-containers/src/runtime/virtcontainers/qemu_test.go:198
13:19:03         	Error:      	Received unexpected error:
13:19:03         	            	S390x does not support appending a vIOMMU
13:19:03         	Test:       	TestQemuCreateVM
13:19:03         	Messages:   	test[4]
```
Link: http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-s390x-unit-PR/1575/console

Will you be able to take a look at this?

tklengyel:
I removed the offending config as I'm not sure how I could special case that this config is expected to fail only on select platforms.

PR #6726, opened 2023-04-25
kata-sys-util: Move CPU info functions

### LINKED ISSUE #6383: Move cpu info functions to kata-sys-utils
## Background

[`kata-ctl`](https://github.com/kata-containers/kata-containers/tree/main/src/tools/kata-ctl) provides a few functions that handle CPU flags:

- [`get_single_cpu_info`](https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/check.rs#L46)
- [`get_cpu_flags`](https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/check.rs#LL66C8-L66C21)

## Use cases

It would be useful for the [CH integration work](https://github.com/kata-containers/kata-containers/issues/6263) if these functions were available in [`kata-sys-util`](https://github.com/kata-containers/kata-containers/tree/main/src/libs/kata-sys-util/src) (maybe in a new  `cpu.rs` file).

Also, it would be useful to add some more utility functions so that the consumer could make a single call that would return all flags in the file and then provide a way to check quickly if a flag exists in the returned list. Something like (but not necessarily identical to the following):

```rust
let cpu_flags_list: Vec<String> = get_cpu_flags_vec();

if cpu_flags_list.contains("foo_flag".to_string()) {
    // ...
}

// Or to avoid converting the arg to a string:
if cpu_flags_list.iter().any(|v| v == "foo_flag")) {
    // ...
}
```

Hence...

## Tasks

- [ ] Move cpu info functions from `kata-ctl` to `kata-sys-util` crate.
- [ ] Add a few extra utility functions to make common usage simpler.

nathanwhyte:
Move get_single_cpu_info and get_cpu_flags into kata-sys-util. Add new functions that get a list of flags and check if a flag exists in that list.

Fixes #6383


jodh-intel:
Thanks for raising @nathanwhyte.

/cc @amshinde.

/test

amshinde (approved):

jodh-intel (approved):
Thanks @nathanwhyte.

lgtm

amshinde:
@nathanwhyte I approved this earlier, but now that I came back to this, I see that you will need to drop the moved functions from the kata-ctl source as well, and change all the invocations in kata-ctl to make use of kata-sys-util instead.

amshinde (changes requested):
@nathanwhyte I approved this earlier, but now that I came back to this, I see that you will need to drop the moved functions from the kata-ctl source as well, and change all the invocations in kata-ctl to make use of kata-sys-util instead.

amshinde:
@nathanwhyte Any updates on this?

amshinde:
@nathanwhyte I have included your commit from this PR in another PR [7236](https://github.com/kata-containers/kata-containers/pull/7236) that needed it to work. PR 7236 has been merged with your commits in it. This PR is no longer needed, hence closing this one.

Issue #6732, opened 2023-04-26
when an error occurs and the function returns early by calling return, the defer statements may not actually take effect


Bevisy:
for example:
```go
// virtcontainers/api.go L56
func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, factory Factory, prestartHookFunc func(context.Context) error) (_ *Sandbox, err error) {
	span, ctx := katatrace.Trace(ctx, virtLog, "createSandboxFromConfig", apiTracingTags)
	defer span.End()

	// Create the sandbox.
	s, err := createSandbox(ctx, sandboxConfig, factory)

	if err != nil {
		return nil, err
	}

	// Cleanup sandbox resources in case of any failure
	defer func() {
		if err != nil {
			s.Delete(ctx)
		}
	}()
...
```
https://github.com/kata-containers/kata-containers/blob/509bc8b6c8aaa573fe96e74e46e44f0bbe2d0662/src/runtime/virtcontainers/api.go#L62

We have found multiple such syntax issues.

bergwolf:
It looks like the problem is that `createSandbox` is not self-contained. The deferred cleanup is expected to clean up a sandbox when something bad happens after the sandbox is created. So `createSandboxFromConfig` looks fine but `ceateSandbox` should be made self-contained such that if it fails, it needs to do proper cleanup to make sure there is no side effects left behind.

```
// virtcontainers/api.go L56
func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, factory Factory, prestartHookFunc func(context.Context) error) (_ *Sandbox, err error) {
	span, ctx := katatrace.Trace(ctx, virtLog, "createSandboxFromConfig", apiTracingTags)
	defer span.End()

	// Create the sandbox.
	s, err := createSandbox(ctx, sandboxConfig, factory)

	if err != nil {
		return nil, err
	}

	// Cleanup sandbox resources in case of any failure
	defer func() {
		if err != nil {
			s.Delete(ctx)
		}
	}()
...
```

Bevisy:
> It looks like the problem is that `createSandbox` is not self-contained. The deferred cleanup is expected to clean up a sandbox when something bad happens after the sandbox is created. So `createSandboxFromConfig` looks fine but `ceateSandbox` should be made self-contained such that if it fails, it needs to do proper cleanup to make sure there is no side effects left behind.
> 
> ```
> // virtcontainers/api.go L56
[...]

PR #6733, opened 2023-04-26
runtime: fix the issue of early return before defer

Bevisy:
1. adjust the positions of "defer" and "return"
2. return "s" not nil even if the sandbox create failed for defer func to clean resource

Fixes: #6732


iced-duck:
@katacontainersbot

Issue #6455, opened 2023-03-14
[BUG] virtiofsd leaks file descriptors during container crash loop


fengwang666:
When a container crashes and restarts, `virtiofsd` doesn't release the file descriptors it opens in the previous container instance. So if a pod is in a crash loop, eventually `virtiofsd` exhausts the file descriptors on the node and impacts other pods on the node. Admittedly, in a normal environment pods shouldn't be in crashloop forever. However, since kata is designed for running untrusted workload in a multi-tenant env. This potentially could be used to perform a DOS attack on the node by a malicious attacker.  

The root cause is unknown. My guess is the kata shim probably should (but doesn't) ask virtiofsd to close all the file descriptors as the container exits. Need to look into it a bit more. 

#### Env:
```
[root@ip-10-10-10-222 /]# /opt/kata/bin/kata-runtime --version
kata-runtime  : 2.5.2
   commit   : 4b39dc0a390584d2ee21072cca7707f4ee7f56c5
   OCI specs: 1.0.2-dev
```
```
[root@ip-10-10-10-222 /]# /opt/kata/libexec/virtiofsd --version
virtiofsd backend 1.3.0
```
```
[root@ip-10-10-10-222 /]# /opt/kata/bin/cloud-hypervisor --version
cloud-hypervisor v26.0
```
#### sample pod to deploy
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    run: fd-leak-test
  name: fd-leak-test
spec:
  replicas: 1
  selector:
    matchLabels:
      run: fd-leak-test
  template:
    metadata:
      labels:
        run: fd-leak-test
    spec:
      runtimeClassName: kata-runtime
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: kubernetes.io/hostname
                    operator: In
                    values:
                      -[your node]
      containers:
        - image: alpine
          imagePullPolicy: Always
          name: fd-leak
          command: ['sh', '-c', 'echo "test exit" && sleep 10 && exit 1']
          ports:
            - containerPort: 80
              protocol: TCP
          resources:
            requests:
              cpu: 500m
      restartPolicy: Always
```

#### verify the bug

Find the sandbox that is crashing, find the virtiofsd processes associated with it. Only one of the two is leaking file desci

[...]

PR #6738, opened 2023-04-26
runtime: Fix virtiofs fd leak

fengwang666:
The kata runtime invokes removeStaleVirtiofsShareMounts after a container is stopped to clean up the stale virtiofs file caches.

Fixes: #6455

### Implementation notes
In my original implementation, I invoked the `RemoveStaleVirtiofsShareMounts` rpc call inside the [UnshareFile](https://github.com/kata-containers/kata-containers/blob/9794c52c6517badf2e3b6f665883d11b37a3b62e/src/runtime/virtcontainers/fs_share_linux.go#L296) method in the fs_share_linux.go and pass in the file path in the rpc call. However turns out this implementation is not very robust because it appears there might be a race where the file inside the guest doesn't expire yet when the `RemoveStaleVirtiofsShareMounts` request arrives. Therefore I change the implementation to scan through all the mounts starting with "/run/kata-containers/shared/containers/" and run `stat` against each of them, which is much more robust against potential delays in virtiofsd.


fengwang666:
@bergwolf @egernst @gkurz Can you help review this fix?

liubin (approved):

amshinde (approved):

src/agent/src/rpc.rs
```diff
@@ -829,6 +831,29 @@ impl agent_ttrpc::AgentService for AgentService {
...
+        is_allowed!(req);
+        let mount_infos = parse_mount_table("/proc/self/mountinfo")
+            .map_err(|e| ttrpc_error!(ttrpc::Code::INTERNAL, e))?;
+        for m in &mount_infos {
+            if m.mount_point.starts_with(KATA_GUEST_SHARE_DIR) {
+                // stat the mount point, virtiofs daemon will remove the stale cache and release the fds if the mount point doesn't exist any more.
+                // More details in https://github.com/kata-containers/kata-containers/issues/6455#issuecomment-1477137277
+                match stat::stat(Path::new(&m.mount_point)) {
+                    Ok(_) => info!(sl!(), "stat {} success", m.mount_point),
+                    Err(e) => info!(sl!(), "stat {} failed: {}", m.mount_point, e),
```
gkurz, line 849: Not sure it helps much to say that `stat()` succeeded or failed. We essentially want to know if some stale mounts point are removed. Also should we expect a specific error from virtiofsd in this case ?

fengwang666: @gkurz No. There isn't a specific error returned from virtiofsd. The reason I added the log is to indicate whether a mount is stale or not.

Issue #5561, opened 2022-11-01
Error message of "Check Body Line Length" is not accurate


liubin:
The regex and error message are:

https://github.com/kata-containers/kata-containers/blob/acd72c44d4c9c2e1fb72544d493ca30d82681b6b/.github/workflows/commit-message-check.yaml#L78-L79

But if the commit message body is empty or has no "Signed-off-by", the error message is always "Body line too long (max 150)".

nitinmewar:
what error message would be accurate? 
should we add another step for these error messages or fix them in "Check Body Line Length" only?

byron-marohn:
We're interested in working on this for the hackathon and will start looking at this tomorrow.

PR #6739, opened 2023-04-26
gha: Fix Body Line Length action flagging empty body commit messages

byron-marohn:
Change the Body Line Length workflow to not trigger when the commit message contains only a message without a body. Other workflows will flag the missing body sections, and it was confusing to have an error message that said 'Body line too long (max 150)' when this was not actually the case.

Fixes: #5561

---

This basically just adds a match for a single line without a new line, OR'd with the existing logic. 
We created a few test cases to try this out here - https://regexr.com/7cq1d

### DIFF
--- a/.github/workflows/commit-message-check.yaml
+++ b/.github/workflows/commit-message-check.yaml
@@ -62,6 +62,9 @@ jobs:
         #   to be specified at the start of the regex as the action is passed
         #   the entire commit message.
         #
+        # - This check will pass if the commit message only contains a subject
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @byron-marohn.

lgtm

/cc @liubin

liubin (approved):
Thank you @byron-marohn

liubin:
@Tim-Zhang would you like take a look?

/test

PR #6742, opened 2023-04-27
runtime: remove overriding ARCH value by default for ppc64le

Amulyam24:
Currently, `ARCH` value is being set to `powerpc64le` by default. `powerpc64le` is only right in the context of rust and any operation which might use this variable for a different purpose might fail on `ppc64le`.

Fixes: #6741


fidencio (approved):
lgtm, thanks @Amulyam24!

Amulyam24:
All the required jobs have passed, hope this can be merged now @fidencio @amshinde

Issue #6743, opened 2023-04-27
runtime-rs: incomplete resource deletion when container stopped


YushuoEdge:
Currently, the container resource deletion is incomplete in runtime-rs, including the following issues:

- No resource deletion in delete_process, which means if a container is created but never started, the container resource will not be cleaned.
- Cleanup resource only happens in Container without notifying ResourceManager. This will make the information inconsistent.

PR #6744, opened 2023-04-27
bugfix: fix the incomplete delete operation of runtime-rs

YushuoEdge:
This PR consists of three commits:

- The first commit solves the problem of resource deletion of Containers that are created but never started.
- The second commit solves the problem of checking states in start_container by mistake.
- The third commit solves the problem of inconsistent information stored after the cleanup of container's rootfs and volumes.

Fixes: #6743

### DIFF
--- a/src/runtime-rs/crates/resource/src/rootfs/mod.rs
+++ b/src/runtime-rs/crates/resource/src/rootfs/mod.rs
@@ -4,31 +4,35 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+mod block_rootfs;
 mod nydus_rootfs;
 mod share_fs_rootfs;
+
+use std::{sync::Arc, vec::Vec};
+
 use agent::Storage;
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
-use kata_types::mount::Mount;
-mod block_rootfs;
 use hypervisor::{device::device_manager::DeviceManager, Hypervisor};
-use std::{sync::Arc, vec::Vec};
+use kata_sys_util::rand::RandomBytes;
+use kata_types::mount::Mount;
 use tokio::sync::RwLock;
 
-use crate::share_fs::ShareFs;
-
 use self::{block_rootfs::is_block_rootfs, nydus_rootfs::NYDUS_ROOTFS_TYPE};
+use crate::share_fs::ShareFs;
 
 const ROOTFS: &str = "rootfs";
 const HYBRID_ROOTFS_LOWER_DIR: &str = "rootfs_lower";
 const TYPE_OVERLAY_FS: &str = "overlay";
 #[async_trait]
 pub trait Rootfs: Send + Sync {
+    fn id(&self) -> String;
+
     async fn get_guest_rootfs_path(&self) -> Result<String>;
     async fn get_rootfs_mount(&self) -> Result<Vec<oci::Mount>>;
     async fn get_storage(&self) -> Option<Storage>;
-    async fn cleanup(&self, device_manager: &RwLock<DeviceManager>) -> Result<()>;
     async fn get_device_id(&self) -> Result<Option<String>>;
+    async fn cleanup(&self, device_manager: &RwLock<DeviceManager>) -> Result<()>;
 }
 
 #[derive(Default)]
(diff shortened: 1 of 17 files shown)


src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -305,10 +306,70 @@ impl Container {
...
+
+        // ignore the stop when container status has already been stopped.
+        let status = inner.init_process.get_status().await;
+        if status == ProcessStatus::Stopped {
+            return Ok(());
+        }
+
+        match container_process.process_type {
+            ProcessType::Container => {
+                inner
```
Tim-0731-Hzt, line 318: As both `Exec` and `Container`  `ProcessType` will do `stop_process,` I think you can just do it once, and use with_context here for different process type

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -305,10 +306,70 @@ impl Container {
 
     pub async fn stop_process(&self, container_process: &ContainerProcess) -> Result<()> {
         let mut inner = self.inner.write().await;
-        inner
-            .stop_process(container_process, true)
+
+        // ignore the stop when container status has already been stopped.
+        let status = inner.init_process.get_status().await;
+        if status == ProcessStatus::Stopped {
```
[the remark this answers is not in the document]
YushuoEdge: This function will be called from wait_process and delete_process. If the user calls `kill` and `delete`, then this function will be called twice. We need to ignore the second.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
YushuoEdge: fixed

src/runtime-rs/crates/resource/src/rootfs/mod.rs
```diff
@@ -139,3 +155,8 @@ impl RootFsResource {
 fn is_single_layer_rootfs(rootfs_mounts: &[Mount]) -> bool {
     rootfs_mounts.len() == 1
 }
+
+pub fn generate_rootfs_id() -> String {
```
Tim-0731-Hzt, line 159: pub(crate)

YushuoEdge: Fixed. Thanks!

src/runtime-rs/crates/resource/src/volume/mod.rs
```diff
@@ -116,3 +130,8 @@ fn is_skip_volume(_m: &oci::Mount) -> bool {
     // TODO: support volume check
     false
 }
+
+pub fn generate_volume_id() -> String {
```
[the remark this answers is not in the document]
YushuoEdge: Fixes. Thanks!

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -305,10 +306,65 @@ impl Container {
...
+                    Ok(agent::Empty::new())
+                } else {
+                    Err(e)
+                }
+            })?;
+
+        let mut inner = self.inner.write().await;
+
+        // cleanup rootfs and volumes, and notify resource manager to keep states persistent
+        let cleaned_rootfs = inner.clean_rootfs().await?;
```
Tim-0731-Hzt, line 360: We could consider doing this with the `resource manager`. As for block volume and block rootfs cleanup, we need to use the `device manager `to do this. Then, we need to pass the `device manager` as the parameter for the cleanup function. But in the current implementation, it's unable to fetch `device manager`. So, my suggestion is to pass `Vec<Arc<dyn Rootfs>>` and `Vec<Arc<dyn Volume>>` to `resource manager`.

YushuoEdge: Update.
Now the cleanup operation is doing in `ResourceManager`, and also the unremoved rootfs and volumes will be returned to `Container`. WDYT?

Issue #6709, opened 2023-04-24
kata-ctl: Use a single timeout value and provide CLI timeout option


jodh-intel:
The [`kata-ctl`](https://github.com/kata-containers/kata-containers/tree/main/src/tools/kata-ctl) tool currently hard-codes multiple timeout values:

```bash
$ cd src/tools/kata-ctl/
$ grep -r 'const.*TIMEOUT' src/                                                                                                                 main [0]
src/ops/volume_ops.rs:const TIMEOUT: Duration = Duration::from_millis(2000);
src/ops/exec_ops.rs:const KATA_AGENT_VSOCK_TIMEOUT: u64 = 5;
src/ops/exec_ops.rs:const TIMEOUT: Duration = Duration::from_millis(2000);
```

We need a single timeout value used by all parts of the code. We also need a CLI option that allows the timeout to be specified manually (an override).

PR #6745, opened 2023-04-27
kata-ctl: Use a single timeout value and provide CLI timeout option

Liu-Weijie:
Use static variable to set all timeout in kata-ctl

Credit PR to Hackathon Team3

Fixes: #6709

### DIFF
--- a/src/tools/kata-ctl/src/args.rs
+++ b/src/tools/kata-ctl/src/args.rs
@@ -160,6 +160,5 @@ pub struct ExecArguments {
 
 #[derive(Debug, Args)]
 pub struct TimeoutArguments {
-    /// timeout value
     pub timeout: u64,
 }
--- a/src/tools/kata-ctl/src/args.rs
+++ b/src/tools/kata-ctl/src/args.rs
@@ -39,6 +39,9 @@ pub enum Commands {
 
     /// Display version details
     Version,
+
+    /// Set Timeout value
+    Timeout(TimeoutArguments),
 }
 
 #[derive(Debug, Args, Error)]
@@ -153,4 +156,10 @@ pub struct ExecArguments {
     #[clap(short = 'p', long = "kata-debug-port", default_value_t = 1026)]
     /// kata debug console vport same as configuration, default is 1026.
     pub vport: u32,
+}
+
+#[derive(Debug, Args)]
+pub struct TimeoutArguments {
+    /// timeout value
+    pub timeout: u64,
 }
--- a/src/tools/kata-ctl/src/args.rs
+++ b/src/tools/kata-ctl/src/args.rs
@@ -12,6 +12,10 @@ use thiserror::Error;
 pub struct KataCtlCli {
     #[clap(subcommand)]
     pub command: Commands,
+    /// Set timeout value (seconds)
+    /// default value: 1 second
+    #[arg(short = 't', long = "timeout")]
+    pub timeout: Option<u64>,
 }
 
 #[derive(Debug, Subcommand)]
@@ -39,9 +43,6 @@ pub enum Commands {
 
     /// Display version details
     Version,
-
-    /// Set Timeout value
-    Timeout(TimeoutArguments),
 }
 
 #[derive(Debug, Args, Error)]
@@ -156,9 +157,4 @@ pub struct ExecArguments {
     #[clap(short = 'p', long = "kata-debug-port", default_value_t = 1026)]
     /// kata debug console vport same as configuration, default is 1026.
     pub vport: u32,
-}
-
-#[derive(Debug, Args)]
-pub struct TimeoutArguments {
-    pub timeout: u64,
 }
--- a/src/tools/kata-ctl/src/ops/check_ops.rs
+++ b/src/tools/kata-ctl/src/ops/check_ops.rs
@@ -13,12 +13,6 @@ use crate::ops::version;
 
 use crate::types::*;
 
-use std::sync::Mutex;
-
-lazy_static! {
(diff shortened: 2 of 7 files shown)


src/tools/kata-ctl/src/args.rs
```diff
@@ -145,3 +148,9 @@ pub struct ExecArguments {
     /// kata debug console vport same as configuration, default is 1026.
     pub vport: u32,
 }
+
+#[derive(Debug, Args)]
+pub struct TimeoutArguments {
+    /// timeout value
```
jodh-intel, line 154: I'd drop this comment as the code is clear enough, I think.

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -129,3 +133,9 @@ pub fn handle_version() -> Result<()> {
     println!("{} version {:?} (type: rust)", NAME, version);
     Ok(())
 }
+
+pub fn handle_timeout(_args: TimeoutArguments) -> Result<()> {
+    unsafe { KATA_TIMEOUT = _args.timeout };
```
jodh-intel, line 138: There is a bit more overhead in doing so, but I'd recommend using the [`lazy_static`](https://docs.rs/lazy_static/latest/lazy_static) crate instead of this as it avoids the `unsafe` block and is the standard way to manage globals like this. Also, you need to rename `_args` to `args` as you are actually using this parameter (the underscore is used to mean, " I'm not using this thing").

Something like:

```rust
  lazy_static! {
      pub static ref KATA_TIMEOUT_SECS: u64 = 1;
  }

// ...

// Access the global
let timeout_ref = KATA_TIMEOUT.clone();
let mut timeout = timeout_ref;

// And modify it
timeout = 99;
```

jodh-intel (commented):
Thanks @Liu-Weijie - a few comments...

src/tools/kata-ctl/src/ops/check_ops.rs
Liu-Weijie: Thanks your suggestions. @jodh-intel. I am new to rust. To avoid using unsafe, I use mutex to access the shared variable. Using global variable is the way I can think of. Is there a way that doesn't introduce additional overhead？Looking for your advice

src/tools/kata-ctl/src/ops/check_ops.rs
```diff
@@ -5,14 +5,22 @@
...
 use crate::check;
 
 use crate::ops::version;
 
 use crate::types::*;
 
+use std::sync::Mutex;
+
+lazy_static! {
+    pub static ref KATA_TIMEOUT_SECS: Mutex<u64> = Mutex::new(1);
```
jodh-intel, line 21: This is fine but I'm not sure we actually need a mutex here as the tool is "single shot" and not threaded or async.

src/tools/kata-ctl/src/args.rs
```diff
@@ -39,6 +39,9 @@ pub enum Commands {
 
     /// Display version details
     Version,
+
+    /// Set Timeout value
+    Timeout(TimeoutArguments),
```
jodh-intel, line 44: The timeout member needs to be part of `KataCtlCli` rather than here as we want to command line to parse like this:

```bash
$ kata-ctl --timeout 1234 exec ...
```

Not like this:

```bash
$ kata-ctl timeout 1234
```

jodh-intel (commented):
Thanks @Liu-Weijie - a few more comments...

src/tools/kata-ctl/src/args.rs
Liu-Weijie: Sure, I will try it

src/tools/kata-ctl/src/args.rs
amshinde: @jodh-intel I think this appears in the cli output as explanatory comments. 
I would keep the comment and explain what it does here.

jodh-intel: Oh, if that's the case, then I guess we need to keep it. But please state what unit the timeout value uses (integer seconds?)

Liu-Weijie: Maybe I can revert this comment and add timeout unit

src/tools/kata-ctl/src/args.rs
Liu-Weijie: I added timeout as option in kata ctl cli. Now it can be used as follow:

```kata-ctl --timeout <num> exec ...```

src/tools/kata-ctl/src/ops/check_ops.rs
amshinde: Agree, I dont think we need a mutex here. The kata-ctl code is not threaded, so this is not required.

amshinde:
@Liu-Weijie The code is looking good. There is just one oustanding comment about dropping the usage of mutex that needs to be addressed here. Could you take a look. Also, can you now squash all the commits into a single commit?

Liu-Weijie:
> @Liu-Weijie The code is looking good. There is just one oustanding comment about dropping the usage of mutex that needs to be addressed here. Could you take a look. Also, can you now squash all the commits into a single commit?

Sure, I will do it later

src/tools/kata-ctl/src/ops/check_ops.rs
Liu-Weijie: I modified the implementation, I used mutex because I wanted to safely access global variables, Now I pass timeout value as a parameter into the function

Issue #6425, opened 2023-03-07
Make kata-manager show version when installing the latest


jodh-intel:
The [`kata-manager`](https://github.com/kata-containers/kata-containers/blob/main/utils/kata-manager.sh) script will install the latest version of Kata by default. However, when it does this, although it determines what the latest version actually _is_, it doesn't show the user - all they see is "(latest version)".

Hence, it would be useful if it actually disiplayed the [semver](https://semver.org/) version along with the text "latest version".

Related, bonus points if the script could determine the latest version, display it, and then exit (aka a "dry run" mode). That could be useful for users and admins.

PR #6746, opened 2023-04-27
Make kata-manager show version when installing the latest

Liu-Weijie:
In this PR, I use curl command get the tag and extract version from the tag to show actual latest version

Credit PR to Hackathon Team3

Fixes: #6425

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -365,6 +365,16 @@ github_download_package()
 	echo "${version}:${file}"
 }
 
+# get the actual latest version of the module
+get_latest_version()
+{
+	local release_url="${1:-}"
+
+	local latest_version=$(curl -sL "$release_url/latest" | grep "tag_name" | sed -E 's/.*"([^"]+)".*/\1/')
+
+	echo "$latest_version"
+}
+
 install_containerd()
 {
 	local requested_version="${1:-}"
@@ -374,7 +384,10 @@ install_containerd()
 	local version_desc="latest version"
 	[ -n "$requested_version" ] && version_desc="version $requested_version"
 
-	info "Downloading $project release ($version_desc)"
+	local version=$(get_latest_version \
+		"$containerd_releases_url")
+
+	info "Downloading $project release ($version)"
 
 	local results=$(github_download_package \
 		"$containerd_releases_url" \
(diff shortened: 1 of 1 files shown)


utils/kata-manager.sh
```diff
@@ -374,7 +374,9 @@ install_containerd()
 	local version_desc="latest version"
 	[ -n "$requested_version" ] && version_desc="version $requested_version"
 
-	info "Downloading $project release ($version_desc)"
+	local latest_version=$(curl -s "$containerd_releases_url/latest" | grep "tag_name" | sed -E 's/.*"([^"]+)".*/\1/')
```
jodh-intel, line 377: I'd add `-L` "just in case":

```suggestion
	local latest_version=$(curl -sL "$containerd_releases_url/latest" | grep "tag_name" | sed -E 's/.*"([^"]+)".*/\1/')
```

utils/kata-manager.sh
```diff
@@ -518,7 +520,9 @@ install_kata()
 	local version_desc="latest version"
 	[ -n "$requested_version" ] && version_desc="version $requested_version"
 
-	info "Downloading $project release ($version_desc)"
+	local latest_version=$(curl -s "$kata_releases_url/latest" | grep "tag_name" | sed -E 's/.*"([^"]+)".*/\1/')
```
jodh-intel, line 523:
```suggestion
	local latest_version=$(curl -sL "$kata_releases_url/latest" | grep "tag_name" | sed -E 's/.*"([^"]+)".*/\1/')
```

It might be worth creating a small function to avoid the duplication, and avoid potential future breakage if we forget to update one of the instances.

jodh-intel (commented):
Thanks @Liu-Weijie - a few comments...

Liu-Weijie:
> Thanks @Liu-Weijie - a few comments...

Your suggestion is reasonable, I have modified it

amshinde (commented):
@Liu-Weijie lgtm. Thanks for making the suggested changes. Can you squash the two commits into a single one ?

Liu-Weijie:
> @Liu-Weijie lgtm. Thanks for making the suggested changes. Can you squash the two commits into a single one ?

Done

amshinde (approved):

PR #6747, opened 2023-04-27
gha: Add ci-on-pr.yaml and skip CI for certains PRs

tklengyel:
Credit to Hackathon Team3

cc: @jodh-intel, @amshinde.

Fixes: #3963

### DIFF
--- a/.github/workflows/ci-on-pr-labels.yaml
+++ b/.github/workflows/ci-on-pr-labels.yaml
@@ -27,12 +27,6 @@ jobs:
     uses: ./.github/workflows/PR-wip-checks.yaml
     secrets: inherit
 
-  run-require-pr-porting-labels:
-    needs: [skip]
-    if: needs.skip.outputs.skip != '1'
-    uses: ./.github/workflows/require-pr-porting-labels.yaml
-    secrets: inherit
-
   run-add-backport-label:
     needs: [skip]
     if: needs.skip.outputs.skip != '1'
--- a/.github/workflows/ci-on-pr-labels.yaml
+++ b/.github/workflows/ci-on-pr-labels.yaml
@@ -1,0 +1,40 @@
+# Copyright (c) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+name: Kata Containers PR CI on label change
+on:
+  pull_request:
+    types:
+      - labeled
+      - unlabeled
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  skip:
+    uses: ./.github/workflows/skip-pr-check.yaml
+    with:
+      pr: ${{ github.event.pull_request.number }}
+    secrets: inherit
+
+  run-PR-wip-checks:
+    needs: [skip]
+    if: needs.skip.outputs.skip != '1'
+    uses: ./.github/workflows/PR-wip-checks.yaml
+    secrets: inherit
+
+  run-require-pr-porting-labels:
+    needs: [skip]
+    if: needs.skip.outputs.skip != '1'
+    uses: ./.github/workflows/require-pr-porting-labels.yaml
+    secrets: inherit
+
+  run-add-backport-label:
+    needs: [skip]
+    if: needs.skip.outputs.skip != '1'
+    uses: ./.github/workflows/add-backport-label.yaml
+    secrets: inherit
--- a/.github/workflows/move-issues-to-in-progress.yaml
+++ b/.github/workflows/move-issues-to-in-progress.yaml
@@ -6,10 +6,22 @@
 name: Move issues to "In progress" in backlog project when referenced by a PR
 
 on:
-  workflow_call:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
 
 jobs:
+  skip:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./.github/workflows/skip-pr-check.yaml
+        with:
+          pr: ${{ github.event.pull_request.number }}
+
   move-linked-issues-to-in-progress:
+    needs: skip
+    if: needs.skip.outputs.skip != '1'
     runs-on: ubuntu-latest
     steps:
       - name: Install hub
--- a/.github/workflows/move-issues-to-in-progress.yaml
+++ b/.github/workflows/move-issues-to-in-progress.yaml
@@ -6,17 +6,13 @@
 name: Move issues to "In progress" in backlog project when referenced by a PR
 
 on:
-  pull_request_target:
-    types:
-      - opened
-      - reopened
+  workflow_call:
 
 jobs:
   move-linked-issues-to-in-progress:
     runs-on: ubuntu-latest
     steps:
       - name: Install hub
-        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
         run: |
           HUB_ARCH="amd64"
           HUB_VER=$(curl -sL "https://api.github.com/repos/github/hub/releases/latest" |\
@@ -27,7 +23,6 @@ jobs:
           sudo install hub /usr/local/bin
 
       - name: Install hub extension script
-        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
         run: |
           # Clone into a temporary directory to avoid overwriting
           # any existing github directory.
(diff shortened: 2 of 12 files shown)


amshinde (approved):

.github/workflows/add-pr-force-skip-ci-label.yaml
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2022 Intel Corporation
```
amshinde, line 1: I would just update the year to 2023 here.

jodh-intel (commented):
Thanks for raising @tklengyel. This is very neat but I think there is a problem. If we add the `force-skip-ci` label, the static checks script (which we need to run on all files) won't run. See:

- https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/static-checks.yaml#L78..L81

One solution would be to modify the workflow file above and add the `force-skip-ci` label _after_ the static checks have run. The problem there though is that I believe all the GH actions run simultaneously so that doesn't really resolve the problem.

It looks like the solution might be to make all the other action/workflow files block on the static checks one using the `jobs.<job_id>.needs` keyword as documented here:

- https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow

tklengyel:
FYI jobs run in parallel as well, you have to specify for each job which other jobs its waiting on using the `needs` parameter. Converting all existing workflows into jobs however would normally require you to move them into a single yaml file. You can work around that by using making existing workflows [callable](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) and have a single ci workflow that runs on the existing hooks and that specifies in what order you want to execute the other callable workflows. That would be my recommendation because then you can also run workflows from the command line too using `gh workflow`.

tklengyel:
OK, I reworked the PR CI flow to have a master workflow `ci-on-pr.yaml` that decides whether to (re-)run workflows that previously had relied on the `force-skip-ci` label. There was some circularity in event triggers due some of the existing workflows applying labels that would then trigger the master workflow again, so I decided to have a separate master workflow `ci-on-pr-labels.yaml` that only (re-)runs workflows that deal with labels.

jodh-intel (approved):
Thanks @tklengyel. This is very neat.

lgtm, but please can you rebase to resolve the conflict?

Aside: I do wonder if it might be possible as a future enhancement to invert the logic so that rather than having to list each actual action in `.github/workflows/ci-on-pr.yaml`, it would be possible to have each actual action simply reference `.github/workflows/ci-on-pr.yaml`. That might be "safer" in the sense that it is potentially easy to create a new action but forget to update the list of actions to run (in `.github/workflows/ci-on-pr.yaml`).

tklengyel:
@jodh-intel The only way to do that would be for each individual workflow to have their own triggers and call the check workflow. That would result in the check workflow being run separately for each, which is just pure overhead. IMHO this is the optimal setup in terms of runtime cycles burned and to add an extra call to the top workflow is minimal effort.

I'll rebase shortly.

jodh-intel:
Thanks @tklengyel!

.github/workflows/ci-on-pr-labels.yaml
```diff
@@ -0,0 +1,40 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  skip:
+    uses: ./.github/workflows/skip-pr-check.yaml
+    with:
+      pr: ${{ github.event.pull_request.number }}
+    secrets: inherit
```
fidencio, line 22: I'm not confident this will work.
When we're dealing with GitHub actions targetting `pull_request`, no secrets are passed down to workflows, apart from `GITHUB_TOKEN`, and this is documented here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories

.github/workflows/move-issues-to-in-progress.yaml
```diff
@@ -37,11 +32,9 @@ jobs:
           popd &>/dev/null
 
       - name: Checkout code to allow hub to communicate with the project
-        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
         uses: actions/checkout@v2
 
       - name: Move issue to "In progress"
-        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
         env:
           GITHUB_TOKEN: ${{ secrets.KATA_GITHUB_ACTIONS_TOKEN }}
```
fidencio, line 39: This will stop working.

fidencio (changes requested):
I've left a few comments as I think this will not do exactly what @tklengyel expects it to do. :-)

Please, take a look at the comments, and let me know WDYT.

.github/workflows/ci-on-pr-labels.yaml
tklengyel: Indeed that seems to be case

.github/workflows/move-issues-to-in-progress.yaml
tklengyel: Did this ever work then? According to the documentation no runner gets secrets when triggered from a forked repo, it doesn't say that it applies only to workflow calls AFAICT.

.github/workflows/ci-on-pr-labels.yaml
tklengyel: Actually, I think the documentation is a bit misleading. It says "secrets are not passed to the runner when a workflow is triggered from a forked repository." That means you fork the repo and trigger the workflow on the fork. That's not what happens during PRs. The PR triggers workflows in the repo you are trying to merge to, not the fork. So this limitation doesn't apply AFAICT.

tklengyel: See description at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull-request-events-for-forked-repositories: "No pull request events occur on the forked repository."

tklengyel: To double-check I've setup an experiment at https://github.com/tklengyel/pr_secret_test/pulls to check whether a PR has access to secrets when triggering a workflow. It doesn't and no matter whether the workflow was triggered directly by the PR event or down the line via workflow calls.

That means the workflows already in place with a PR trigger in kata-containers that use any secret token other then the default GITHUB_TOKEN are broken.

.github/workflows/move-issues-to-in-progress.yaml
tklengyel: I don't think this ever worked.

.github/workflows/require-pr-porting-labels.yaml
```diff
@@ -44,7 +35,6 @@ jobs:
           popd &>/dev/null
 
       - name: Stop PR being merged unless it has a correct set of porting labels
-        if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
         env:
           GITHUB_TOKEN: ${{ secrets.KATA_GITHUB_ACTIONS_TOKEN }}
```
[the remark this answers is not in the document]
tklengyel: I don't think this ever worked.

.github/workflows/ci-on-pr-labels.yaml
tklengyel: Wow, OK. There is a difference between `pull_request` and `pull_request_target` triggers. With `pull_request_target` the workflow will have access to the secrets. What a pain :)

tklengyel:
@fidencio Thanks for pointing out the issue, that was very subtle. I removed the workflows that used `pull_request_target` as triggers from the master `ci-on-pr` workflows and those will run separately with their own trigger and just individually call `skip-pr-check`.

fidencio:
> @fidencio Thanks for pointing out the issue, that was very subtle. I removed the workflows that used `pull_request_target` as triggers from the master `ci-on-pr` workflows and those will run separately with their own trigger and just individually call `skip-pr-check`.

Thanks for the change, @tklengyel!
I've been dealing with "not-so-well-expected-or-defined" behaviours from GHA in the past month or so, so my eyes are getting trained to see the traps I've fallen into. :-)

fidencio:
@tklengyel, one other things that I'd like to ask you, do we plan to block PRs to get merged based on `workflow_call` operations?  If so, that's a no-go, as jobs based on `workflow_call` events simply do not show up in the settings to make those required. :-/

And, yes, that's yet another thing from GHA.

tklengyel:
> do we plan to block PRs to get merged based on `workflow_call` operations? 

I assume you would mark the master `ci-on-pr` workflow as something Required to pass, not the sub-workflows that get triggered via `workflow_call`.

Issue #6758, opened 2023-04-28
CI | GHA | sev k8s test runner failing


ryansavino:
Error:
```
Run kubectl delete -k tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
error: must build at directory: 'tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml': file is not directory
Error: Process completed with exit code 1.
```

PR #6759, opened 2023-04-28
gha: sev: fix for kata-deploy error

ryansavino:
kubectl commands need a '-f' instead of a '-k'

Fixes: #6758


wainersm (approved):
Thanks for the quick fix @ryansavino !

sameo (approved):

Issue #5401, opened 2022-10-12
kata-ctl: Switch to slog for output and logging


jodh-intel:
`kata-ctl` currently writes direct to stdout/stderr using `println!()`/`eprintln!()`. Switch to using the [Kata logging crate](https://github.com/kata-containers/kata-containers/tree/main/src/libs/logging) for consistency and to benefit from structured logging.

jodh-intel:
Related: #5878.

jodh-intel:
I'm assigning this to the [current outreach team](https://github.com/kata-containers/community/wiki/Outreach---Internship-programs).

cmaf:
Related to https://github.com/kata-containers/kata-containers/issues/6654

jodh-intel:
Removed `outreach` label as the [current Outreach team](https://github.com/kata-containers/community/wiki/Outreach---Internship-programs#ndsu-capstone-class-january---may-2023) are not looking at this issue now.

For minimal examples of how to complete this task, see the call like this...

```rust
let (logger, _guard) = logging::create_logger(name, crate_name!(), log_level, writer);
```

... in the following:

- https://github.com/kata-containers/kata-containers/blob/main/src/tools/agent-ctl/src/main.rs
- https://github.com/kata-containers/kata-containers/blob/main/src/tools/trace-forwarder/src/main.rs

See also:

- https://github.com/jamesodhunt/rust-slog-structured-logging

byron-marohn:
We're going to start working on this one as part of the hackathon. Also looking at 6654 but not sure yet if we're going to try to tackle the whole scope there too at the same time.

Issue #6654, opened 2023-04-12
kata-ctl: add option for verbose output and otherwise improve output


cmaf:
**Which feature do you think can be improved?**

In the kata-ctl tool, currently we do not have verbose output nor a way to suppress all output except failures.

**How can it be improved?**

Add command line option for verbose output at minimum, and consider suppressing all output except failures by default.

**Additional Information**

Related to https://github.com/kata-containers/kata-containers/issues/5401

PR #6764, opened 2023-04-28
kata-ctl: Switch to slog logging; add --log-level and --json-logging arguments

byron-marohn:
Fixes: #5401, #6654

- Switch kata-ctl from eprintln!()/println!() to structured logging via the logging library which uses slog.
- Adds a new create_term_logger() library call which enables printing log messages to the terminal via a less verbose / more human readable terminal format with colors.
- Adds --log-level argument to select the minimum log level of printed messages.
- Adds --json-logging argument to switch to logging in JSON format.

### DIFF
--- a/src/tools/kata-ctl/src/arch/aarch64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/aarch64/mod.rs
@@ -11,6 +11,7 @@ mod arch_specific {
     use crate::types::*;
     use crate::utils;
     use anyhow::Result;
+    use slog::{info, o, warn};
     use std::path::Path;
 
     const KVM_DEV: &str = "/dev/kvm";
@@ -18,6 +19,12 @@ mod arch_specific {
     pub const ARCH_CPU_VENDOR_FIELD: &str = "CPU implementer";
     #[allow(dead_code)]
     pub const ARCH_CPU_MODEL_FIELD: &str = "CPU architecture";
+
+    macro_rules! sl {
+        () => {
+            slog_scope::logger().new(o!("subsystem" => "aarch64"))
+        };
+    }
 
     // List of check functions
     static CHECK_LIST: &[CheckItem] = &[CheckItem {
@@ -28,11 +35,14 @@ mod arch_specific {
     }];
 
     pub fn check(_args: &str) -> Result<()> {
-        println!("INFO: check: aarch64");
+        info!(sl!(), "check: aarch64");
         if Path::new(KVM_DEV).exists() {
-            println!("Kata Containers can run on this host\n");
+            info!(sl!(), "Kata Containers can run on this host\n");
         } else {
-            eprintln!("WARNING: Kata Containers can't run on this host as lack of virtulization support\n");
+            warn!(
+                sl!(),
+                "Kata Containers can't run on this host as lack of virtulization support\n"
+            );
         }
 
         Ok(())
--- a/src/tools/kata-ctl/src/args.rs
+++ b/src/tools/kata-ctl/src/args.rs
@@ -12,6 +12,25 @@ use thiserror::Error;
 pub struct KataCtlCli {
     #[clap(subcommand)]
     pub command: Commands,
+    #[clap(short, long, value_enum, value_parser = parse_log_level)]
+    /// Sets the minimum log level required for log messages to be displayed. Default is 'info'.
+    /// Valid values are: trace, debug, info, warning, error, critical
+    pub log_level: Option<slog::Level>,
+    #[clap(short, long, action)]
+    /// If enabled, log messages will be JSON formatted for easier machine parsing
+    pub json_logging: bool,
+}
+
+fn parse_log_level(arg: &str) -> Result<slog::Level, String> {
+    match arg {
+        "trace" => Ok(slog::Level::Trace),
+        "debug" => Ok(slog::Level::Debug),
+        "info" => Ok(slog::Level::Info),
+        "warning" => Ok(slog::Level::Warning),
+        "error" => Ok(slog::Level::Error),
+        "critical" => Ok(slog::Level::Critical),
+        _ => Err("Must be one of [trace, debug, info, warning, error, critical]".to_string()),
+    }
 }
 
 #[derive(Debug, Subcommand)]
--- a/src/tools/kata-ctl/src/main.rs
+++ b/src/tools/kata-ctl/src/main.rs
@@ -11,7 +11,8 @@ mod types;
 mod utils;
 
 use anyhow::Result;
-use clap::Parser;
+use clap::{crate_name, Parser};
+use std::io;
 use std::process::exit;
 
 use args::{Commands, KataCtlCli};
@@ -22,11 +23,28 @@ use ops::check_ops::{
 use ops::env_ops::handle_env;
 use ops::exec_ops::handle_exec;
 use ops::volume_ops::handle_direct_volume;
+use slog::{error, o};
+
+macro_rules! sl {
+    () => {
+        slog_scope::logger().new(o!("subsystem" => "kata-ctl_main"))
+    };
+}
 
 fn real_main() -> Result<()> {
     let args = KataCtlCli::parse();
 
-    match args.command {
+    let log_level = args.log_level.unwrap_or(slog::Level::Info);
+
+    let (logger, _guard) = if args.json_logging {
+        logging::create_logger(crate_name!(), crate_name!(), log_level, io::stdout())
+    } else {
+        logging::create_term_logger(log_level)
+    };
+
+    let _guard = slog_scope::set_global_logger(logger);
+
+    let res = match args.command {
         Commands::Check(args) => handle_check(args),
         Commands::DirectVolume(args) => handle_direct_volume(args),
         Commands::Exec(args) => handle_exec(args),
@@ -35,12 +53,19 @@ fn real_main() -> Result<()> {
         Commands::Iptables(args) => handle_iptables(args),
         Commands::Metrics(args) => handle_metrics(args),
         Commands::Version => handle_version(),
+    };
+
+    // Log errors here, then let the logger go out of scope in main() to ensure
+    // the asynchronous drain flushes all messages before exit()
+    if let Err(e) = &res {
+        error!(sl!(), "{:#?}", e);
     }
+
+    res
 }
 
 fn main() {
-    if let Err(e) = real_main() {
-        eprintln!("ERROR: {:#?}", e);
+    if let Err(_e) = real_main() {
         exit(1);
     }
 }
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -423,10 +423,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs-next"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b98cf8ebf19c3d1b223e151f99a4f9f0690dca41414773390fc824184ac833e1"
+dependencies = [
+ "cfg-if 1.0.0",
+ "dirs-sys-next",
+]
+
+[[package]]
 name = "dirs-sys"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b1d1d91c932ef41c0f2663aa8b0ca0342d444d842c06914aa0a7e352d0bada6"
+dependencies = [
+ "libc",
+ "redox_users",
+ "winapi",
+]
+
+[[package]]
+name = "dirs-sys-next"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4ebda144c4fe02d1f7ea1a7d9641b6fc6b580adcfa024ae48797ecdeb6825b4d"
 dependencies = [
  "libc",
  "redox_users",
@@ -949,6 +970,7 @@ dependencies = [
  "slog-async",
  "slog-json",
  "slog-scope",
+ "slog-term",
 ]
 
 [[package]]
@@ -1763,6 +1785,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustversion"
+version = "1.0.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4f3208ce4d8448b3f3e7d168a73f5e0c43a61e32930de3bceeccedb388b6bf06"
+
+[[package]]
 name = "ryu"
 version = "1.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1952,6 +1980,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "slog-term"
+version = "2.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "87d29185c55b7b258b4f120eab00f48557d4d9bc814f41713f449d35b0f8977c"
+dependencies = [
+ "atty",
+ "slog",
+ "term",
+ "thread_local",
+ "time 0.3.11",
+]
+
+[[package]]
 name = "smallvec"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2032,6 +2073,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "term"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c59df8ac95d96ff9bede18eb7300b0fda5e5d8d90960e76f8e14ae765eedbf1f"
+dependencies = [
+ "dirs-next",
+ "rustversion",
+ "winapi",
+]
+
+[[package]]
 name = "termcolor"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2102,7 +2154,14 @@ dependencies = [
  "itoa",
  "libc",
  "num_threads",
+ "time-macros",
 ]
+
+[[package]]
+name = "time-macros"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "42657b1a6f4d817cda8e7a0ace261fe0cc946cf3a80314390b22cc61ae080792"
 
 [[package]]
 name = "tokio"
--- a/src/libs/logging/Cargo.toml
+++ b/src/libs/logging/Cargo.toml
@@ -15,6 +15,7 @@ serde_json = "1.0.73"
 #   (by stopping the compiler from removing log calls).
 slog = { version = "2.5.2", features = ["dynamic-keys", "max_level_trace", "release_max_level_debug"] }
 slog-json = "2.4.0"
+slog-term = "2.9.0"
 slog-async = "2.7.0"
 slog-scope = "4.4.0"
 
--- a/src/libs/logging/src/lib.rs
+++ b/src/libs/logging/src/lib.rs
@@ -39,6 +39,28 @@ const LOG_LEVELS: &[(&str, slog::Level)] = &[
 
 const DEFAULT_SUBSYSTEM: &str = "root";
 
+// Creates a logger which prints output as human readable text to the terminal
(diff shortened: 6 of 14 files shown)


src/tools/kata-ctl/src/arch/aarch64/mod.rs
```diff
@@ -27,11 +33,11 @@ mod arch_specific {
     }];
 
     pub fn check(_args: &str) -> Result<()> {
-        println!("INFO: check: aarch64");
+        info!(sl!(), "check: aarch64");
         if Path::new(KVM_DEV).exists() {
             println!("Kata Containers can run on this host\n");
```
byron-marohn, line 38: One question we had - we initially elected not to port over every single println!() call into the logging framework. We only included lines that had WARNING or INFO, etc. in their text. 

Would you like us to port over the rest of these as well? Basically - converting all println!() calls?

src/tools/kata-ctl/src/main.rs
```diff
@@ -21,25 +22,54 @@ use ops::check_ops::{
...
 
 fn main() {
-    if let Err(e) = real_main() {
-        eprintln!("ERROR: {:#?}", e);
-        exit(1);
-    }
+   if let Err(_e) = real_main() {
+       exit(1);
+   }
 }
```
byron-marohn, line 72: Note that this strange looking change was necessary because the error! macro no longer works once the logging variable goes out of scope in real_main(). This structure ensures that the error is printed before that happens.

src/tools/kata-ctl/src/arch/aarch64/mod.rs
cmaf: @byron-marohn IMO, if you already ported some of a similar type, you should probably cover all of them. In this case I would use INFO here

src/tools/kata-ctl/src/main.rs
```diff
@@ -21,25 +22,54 @@ use ops::check_ops::{
...
+
+
+
```
cmaf, line 75: @byron-marohn please remove extra whitespace

cmaf (commented):
Thanks @byron-marohn and others! I left a couple comments and have a few more:
1. If you have time, could you go through and classify the rest of the println!()s as one of the levels?
2. Could you resolve the conflicts
3.  Don't forget to add all the necessary co-authors - not sure if your co-author vs. signed-off-by divisions were intentional

src/tools/kata-ctl/src/main.rs
jodh-intel: Could you add a comment in the code to explain that as it does indeed look odd! :smile:

jodh-intel (commented):
@byron-marohn - This is very nice indeed. One comment - could you update the code so that the "Ok" messages are also handled. For example, the last two lines of output below are just "raw" output rather than slog output I think?:

```bash
$ kata-ctl check all
subsystem: x86_64
 May 03 10:46:00.477 INFO check CPU: x86_64
 May 03 10:46:00.477 INFO check kernel modules for: x86_64
kvm Ok        # <--- XXX: Not using slog logging?
kvm_intel Ok  # <--- XXX: Not using slog logging?
```

jodh-intel:
@byron-marohn - could you also rebase to resolve the conflict?

byron-marohn:
> Thanks @byron-marohn and others! I left a couple comments and have a few more:
> 
> 1. If you have time, could you go through and classify the rest of the println!()s as one of the levels?
> 2. Could you resolve the conflicts
> 3. Don't forget to add all the necessary co-authors - not sure if your co-author vs. signed-off-by divisions were intentional

We'll work on these things, thanks for the review. 

The co-authorship decisions were intentional. We split the work on the hackathon between the two projects - version checking tool and these kata-ctl improvements. Some team members contributed code to one and only contributed review to the other (and thus don't have the co-author tag but do have signed-off-by since they were involved).

byron-marohn:
@jodh-intel 
When making changes to this, would you like us to amend the existing commit and force-push that to this branch, or would you like us to make a 2nd commit with those fixes?

jodh-intel:
Hi @byron-marohn - Please could you resolve the commits, then rebase and squash down to a single commit.

jodh-intel:
Hi @byron-marohn - please could you rebase to resolve the conflicts and possibly fix the messages not using the `info!()` log calls?

jodh-intel (approved):

amshinde (approved):

amshinde:
@byron-marohn Static checks are failing: https://github.com/kata-containers/kata-containers/actions/runs/4931999547/jobs/8829290945?pr=6764

Cargo is complaining of tree being dirty. Can you commit any changes that were made to `Cargo.lock` as part of your changes?

byron-marohn:
Interesting - it's the file at `src/agent/Cargo.lock`, which we didn't change directly. Looks like it's picking up the transitive dependency change to the logging library. 

Rebuilt the agent and added the new Cargo.lock. Might be this pops up somewhere else too, guess we'll see.

src/tools/kata-ctl/src/arch/aarch64/mod.rs
byron-marohn: Done

src/tools/kata-ctl/src/main.rs
byron-marohn: Done

src/tools/kata-ctl/src/main.rs
byron-marohn: Done

byron-marohn:
Also just shortened commit message subject - I see that check was failing too.

byron-marohn:
+ another push to run cargo fmt

jodh-intel:
Hi @byron-marohn and @lucasphillips - The static checks are failing on a couple of code nits:

```
error: cannot find macro `warn` in this scope
   --> src/check.rs:592:13
    |
592 |             warn!(
    |             ^^^^
    |
    = help: consider importing one of these items:
            slog::warn
            slog_scope::warn
    = note: `warn` is in scope, but it is an attribute: `#[warn]`

error: could not compile `kata-ctl` due to previous error
warning: build failed, waiting for other jobs to finish...
error: unneeded `return` statement
  --> src/main.rs:68:5
   |
68 |     return res;
   |     ^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
   = note: `-D clippy::needless-return` implied by `-D warnings`
   = help: remove `return`

error: could not compile `kata-ctl` due to previous error
```

byron-marohn:
I'm having some trouble with the warn! macro one, really bizarre - 

The file check.rs clearly uses warn!. But when I import it, the code fails to build with an used imports warning
```
error: unused import: `warn`
  --> src/check.rs:15:21
   |
15 | use slog::{info, o, warn};
   |                     ^^^^
   |

error: could not compile `kata-ctl` due to previous error
make: *** [Makefile:39: build] Error 101
```

Any idea why this would be? Clearly it is used... it's in a function marked `#test`, not sure if this makes a difference.

amshinde:
@byron-marohn If you are using warn only in the tests and not in the rest of the check.rs, import it only in the tests module. Something like:
https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/utils.rs#L160
`tempdir` is only used in tests and its only imported in tests.

lucasphillips:
> @byron-marohn If you are using warn only in the tests and not in the rest of the check.rs, import it only in the tests module. Something like: https://github.com/kata-containers/kata-containers/blob/main/src/tools/kata-ctl/src/utils.rs#L160 `tempdir` is only used in tests and its only imported in tests.

Oh, this makes me wonder if we should be using structured logging in the tests in the first place. Do we actually want to use slog for test messages, or just the actual kata-ctl output?

src/tools/kata-ctl/src/args.rs
```diff
@@ -12,6 +12,35 @@ use thiserror::Error;
 pub struct KataCtlCli {
     #[clap(subcommand)]
     pub command: Commands,
+    #[clap(short, long, value_enum)]
+    /// Sets the minimum log level required for log messages to be displayed. Default is 'info'
+    pub log_level: Option<LevelWrapper>,
```
Ji-Xinyou, line 17: Why not just use Option<String> and implement a branch on that string, this struct seems unnecessary.

src/tools/kata-ctl/src/main.rs
```diff
@@ -22,11 +23,32 @@ use ops::check_ops::{
 use ops::env_ops::handle_env;
 use ops::exec_ops::handle_exec;
 use ops::volume_ops::handle_direct_volume;
+use slog::{error, o};
+
+macro_rules! sl {
+    () => {
+        slog_scope::logger().new(o!("subsystem" => "main"))
```
Ji-Xinyou, line 30: why is this subsystem name main

src/tools/kata-ctl/src/main.rs
```diff
@@ -22,11 +23,32 @@ use ops::check_ops::{
...
     let args = KataCtlCli::parse();
 
-    match args.command {
+    let log_level = match args.log_level {
+        None => slog::Level::Info,
+        Some(level_arg) => level_arg.get_slog_level(),
+    };
+
+    let (logger, _guard);
+    if args.json_logging {
```
Ji-Xinyou, line 43: Do not write this kind of initialization.

Use
```Rust
let (logger, _guard) = if args.json_logging {
...
} else {
...
}
```
instead.

src/tools/kata-ctl/src/main.rs
```diff
@@ -35,12 +57,19 @@ fn real_main() -> Result<()> {
         Commands::Iptables(args) => handle_iptables(args),
         Commands::Metrics(args) => handle_metrics(args),
         Commands::Version => handle_version(),
+    };
+
+    // Log errors here, then let the logger go out of scope in main() to ensure
+    // the asynchronous drain flushes all messages before exit()
+    if let Err(ref e) = res {
```
Ji-Xinyou, line 64: why `ref e`? `error!()` could take a reference.

src/tools/kata-ctl/src/main.rs
```diff
@@ -35,12 +57,19 @@ fn real_main() -> Result<()> {
...
+        error!(sl!(), "{:#?}", e);
     }
+
+    return res;
 }
 
 fn main() {
-    if let Err(e) = real_main() {
-        eprintln!("ERROR: {:#?}", e);
+    if let Err(_e) = real_main() {
```
Ji-Xinyou, line 72: In `real_main` you print the error, I mean that is not the Rust way of error handling.

Conventionally, you return an error in `real_main`, handle it in `main` and print it out.
Like
```Rust
fn main() {
    if let Err(e) = real_main() {
        error!("{:?}", e);
    }
}
```

Ji-Xinyou (changes requested):
Some changes needed.

src/tools/kata-ctl/src/main.rs
byron-marohn: Agreed - and we tried to get this to work. However - the logger is only in scope in the real_main() function - when it goes out of scope, putting this error! print in main no longer works. I believe there are only three options: 
- The way it is now, with the print in real_main while the logger is still in scope. 
- Revert this specific print to not use slog
- Pull enough of the argument parsing logic into this function so that the logger can be set up here, so the print is valid

Definitely open to other suggestions - this is ugly, and we added a comment about it. Let me know if that comment needs additional detail.

src/tools/kata-ctl/src/main.rs
byron-marohn: Fixed to `Err(e) = &res`, thanks.

src/tools/kata-ctl/src/args.rs
byron-marohn: Thanks for the suggestion - I tried this initially and couldn't get it to work. Finally figured it out and it's smaller, better code with no extra struct using clap's `value_parser` feature.

src/tools/kata-ctl/src/main.rs
byron-marohn: This I could use your help with - we named the "subsystems" whatever file they were in. I'm going to change this to `kata-ctl_main`, but without understanding how this logging is used elsewhere I'm not sure what values would be most useful.

src/tools/kata-ctl/src/main.rs
byron-marohn: Done, thanks

byron-marohn:
Changes made, thanks for your review & comments @Ji-Xinyou. I added some replies to individual comments. Fixed 3/5, I think 1/5 should probably stay as-is unless you have a better idea, and last 1/5 open to suggestion on subsystem names. 

Here's the diff vs. the previous to ease review: 
```diff
diff --git a/src/tools/kata-ctl/src/args.rs b/src/tools/kata-ctl/src/args.rs
index 9f7ee7489..8049ae31b 100644
--- a/src/tools/kata-ctl/src/args.rs
+++ b/src/tools/kata-ctl/src/args.rs
@@ -12,34 +12,24 @@ use thiserror::Error;
 pub struct KataCtlCli {
     #[clap(subcommand)]
     pub command: Commands,
-    #[clap(short, long, value_enum)]
-    /// Sets the minimum log level required for log messages to be displayed. Default is 'info'
-    pub log_level: Option<LevelWrapper>,
+    #[clap(short, long, value_enum, value_parser = parse_log_level)]
+    /// Sets the minimum log level required for log messages to be displayed. Default is 'info'.
+    /// Valid values are: trace, debug, info, warning, error, critical
+    pub log_level: Option<slog::Level>,
     #[clap(short, long, action)]
     /// If enabled, log messages will be JSON formatted for easier machine parsing
     pub json_logging: bool,
 }

-#[derive(clap::ValueEnum, Clone, Debug)]
-pub enum LevelWrapper {
-    Trace,
-    Debug,
-    Info,
-    Warning,
-    Error,
-    Critical,
-}
-
-impl LevelWrapper {
-    pub fn get_slog_level(&self) -> slog::Level {
-        match self {
-            LevelWrapper::Trace => slog::Level::Trace,
-            LevelWrapper::Debug => slog::Level::Debug,
-            LevelWrapper::Info => slog::Level::Info,
-            LevelWrapper::Warning => slog::Level::Warning,
-            LevelWrapper::Error => slog::Level::Error,
-            LevelWrapper::Critical => slog::Level::Critical,
-        }
+fn parse_log_level(arg: &str) -> Result<slog::Level, String> {
+    match arg {
+        "trace" => Ok(slog::Level::Trace),
+        "debug" => Ok(slog::Level::Debug),
+        "info" => Ok(slog::Level::Info),
+        "warning" => Ok(slog::Level::Warning),
+        "error" => Ok(slog::Level::Error),
+        "critical" => Ok(slog::Level::Critical),
+        _ => Err("Must be one of [trace, debug, info, warning, error, critical]".to_string()),
     }
 }

diff --git a/src/tools/kata-ctl/src/check.rs b/src/tools/kata-ctl/src/check.rs
index dc587932c..d531425c2 100644
--- a/src/tools/kata-ctl/src/check.rs
+++ b/src/tools/kata-ctl/src/check.rs
@@ -408,6 +408,7 @@ pub fn check_kernel_module_loaded(module: &str, parameter: &str) -> Result<Strin
 mod tests {
     use super::*;
     use semver::Version;
+    use slog::warn;
     use std::fs;
     use std::io::Write;
     use tempfile::tempdir;
diff --git a/src/tools/kata-ctl/src/main.rs b/src/tools/kata-ctl/src/main.rs
index 053f84c1e..5495d0279 100644
--- a/src/tools/kata-ctl/src/main.rs
+++ b/src/tools/kata-ctl/src/main.rs
@@ -27,25 +27,21 @@ use slog::{error, o};

 macro_rules! sl {
     () => {
-        slog_scope::logger().new(o!("subsystem" => "main"))
+        slog_scope::logger().new(o!("subsystem" => "kata-ctl_main"))
     };
 }

 fn real_main() -> Result<()> {
     let args = KataCtlCli::parse();

-    let log_level = match args.log_level {
-        None => slog::Level::Info,
-        Some(level_arg) => level_arg.get_slog_level(),
-    };
+    let log_level = args.log_level.unwrap_or(slog::Level::Info);

-    let (logger, _guard);
-    if args.json_logging {
-        let writer = io::stdout();
-        (logger, _guard) = logging::create_logger(crate_name!(), crate_name!(), log_level, writer);
+    let (logger, _guard) = if args.json_logging {
+        logging::create_logger(crate_name!(), crate_name!(), log_level, io::stdout())
     } else {
-        (logger, _guard) = logging::create_term_logger(log_level);
-    }
+        logging::create_term_logger(log_level)
+    };
+
     let _guard = slog_scope::set_global_logger(logger);

     let res = match args.command {
@@ -61,11 +57,11 @@ fn real_main() -> Result<()> {

     // Log errors here, then let the logger go out of scope in main() to ensure
     // the asynchronous drain flushes all messages before exit()
-    if let Err(ref e) = res {
+    if let Err(e) = &res {
         error!(sl!(), "{:#?}", e);
     }

-    return res;
+    res
 }

 fn main() {
```

src/tools/kata-ctl/src/main.rs
Ji-Xinyou: I think it's fine to leave it like this. The only way out is to initialize the logger before your call to `real_main()`, but thats not better lol.

Ji-Xinyou:
> Changes made, thanks for your review & comments @Ji-Xinyou. I added some replies to individual comments. Fixed 3/5, I think 1/5 should probably stay as-is unless you have a better idea, and last 1/5 open to suggestion on subsystem names.
> 
> Here's the diff vs. the previous to ease review:

Thanks for your work, I think the code is good enough for now.

Ji-Xinyou (approved):

jodh-intel:
Hi @byron-marohn - Unfortunately the (required) s390x CI is failing as the s390x version of the code is missing the appropriate `use` statements for `slog`:

```
16:40:04 error: cannot find macro `info` in this scope
16:40:04   --> src/arch/s390x/mod.rs:37:9
16:40:04    |
16:40:04 37 |         info!(sl!(), "check CPU: s390x");
16:40:04    |         ^^^^
16:40:04    |
16:40:04    = help: consider importing one of these items:
16:40:04            slog::info
16:40:04            slog_scope::info
16:40:04 
16:40:04 error: cannot find macro `warn` in this scope
16:40:04   --> src/arch/s390x/mod.rs:51:13
16:40:04    |
16:40:04 51 |             warn!(sl!(), "Missing CPU flags {:?}", missing_cpu_features);
16:40:04    |             ^^^^
16:40:04    |
16:40:04    = help: consider importing one of these items:
16:40:04            slog::warn
16:40:04            slog_scope::warn
16:40:04    = note: `warn` is in scope, but it is an attribute: `#[warn]`
16:40:04 
16:40:04 error: cannot find macro `info` in this scope
16:40:04   --> src/arch/s390x/mod.rs:58:9
16:40:04    |
16:40:04 58 |         info!(sl!(), "check: s390x");
16:40:04    |         ^^^^
16:40:04    |
16:40:04    = help: consider importing one of these items:
16:40:04            slog::info
16:40:04            slog_scope::info
16:40:04 
16:40:04 error: unused macro definition: `sl`
16:40:04   --> src/arch/s390x/mod.rs:24:18
16:40:04    |
16:40:04 24 |     macro_rules! sl {
16:40:04    |                  ^^
16:40:04    |
16:40:04    = note: `-D unused-macros` implied by `-D warnings`
16:40:04 
16:40:04 error: could not compile `kata-ctl` due to 4 previous errors
```

byron-marohn:
Rebased and added this diff - hopefully now we'll be good? 
```diff
+++ b/src/tools/kata-ctl/src/arch/s390x/mod.rs
@@ -13,6 +13,7 @@ mod arch_specific {
     use crate::utils;
     use anyhow::{anyhow, Result};
     use nix::unistd::Uid;
+    use slog::{info, o, warn};
     use std::collections::HashMap;
     use std::io::BufRead;
     use std::io::BufReader;
```

byron-marohn:
@jodh-intel I think this fixed the actual issues - but there are still a few failing tests. I looked through the logs there and they don't look related but I don't have a good eye for this. 

Any suggestions on how to move this one forward?

cmaf (approved):

jodh-intel:
Gah - sorry @byron-marohn but the branch is now conflicted. Could you possibly rebase and repush to resolve?

PR #6768, opened 2023-04-29
build: Kata-manager should load required VSOCK kernel modules

Divyansh-Raina:
Added a check for running the required sudo commands and verifying the status of those commands

Issue #6696

Signed-off-by: Divyansh Raina <divyansh.raina@intel.com>


utils/kata-manager.sh
```diff
@@ -10,6 +10,15 @@ set -o nounset
...
 set -o errtrace
 
+#Added code :- simple checker, just sudo would require a 
+# passwword if sudo is not used before a restart
+if sudo modprobe vhost_vsock && sudo modprobe vhost_net; then
+  echo "VSOCK sockets and are supported"
+else
+  echo "Error: Host environment is too old to support VSOCK sockets"
+  exit 1
+fi
```
fidencio, line 20: Aren't those automatically loaded by the VMMs whenever a new enough kernel is used?

jodh-intel: I have some Ubuntu QEMU VM's that don't load the kernel modules so I do think this check should be performed.

utils/kata-manager.sh
```diff
@@ -10,6 +10,15 @@ set -o nounset
 set -o pipefail
 set -o errtrace
 
+#Added code :- simple checker, just sudo would require a 
+# passwword if sudo is not used before a restart
+if sudo modprobe vhost_vsock && sudo modprobe vhost_net; then
+  echo "VSOCK sockets and are supported"
+else
+  echo "Error: Host environment is too old to support VSOCK sockets"
```
jodh-intel, line 18: Could you send the message to standard error by using the `die()` function:

```suggestion
  die "Host environment is too old to support VSOCK sockets"
```

utils/kata-manager.sh
```diff
@@ -10,6 +10,15 @@ set -o nounset
 set -o pipefail
 set -o errtrace
 
+#Added code :- simple checker, just sudo would require a 
+# passwword if sudo is not used before a restart
+if sudo modprobe vhost_vsock && sudo modprobe vhost_net; then
+  echo "VSOCK sockets and are supported"
```
jodh-intel, line 16: I don't think we need a message for the success case so I'd drop this. It might then make sense to invert the logic so that if we couldn't load those modules, we fail.

Finally, it might be an idea to structure this as a `for` loop. It takes up more lines of code but is clearer and easier to update if we decide to add further kernel modules to the list. Something like:

```bash
local -a modules=()

modules+=(vhost_vsock)
modules+=(vhost_net)

local module

for module in ${modules[@]}
do
    sudo modprobe "$module" || die "Host kernel does not support VSOCK sockets"
done
```

jodh-intel (commented):
Thanks @Divyansh-Raina - a few comments...

amshinde:
ping @Divyansh-Raina

jodh-intel:
Hi @Divyansh-Raina - please can you address the comment. Also, you need to change the format of your commit slightly (from "Issue #6696" to the GitHub standard and CI-mandated, "Fixes: #6696"). See:

- https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

Issue #6771, opened 2023-05-03
Provide better context when agent config is missing.


jepio:
When `agent.config_file=<path>` in kernel cmdline refers to a missing file, the agent panics but does not provide context why.
Fix this by enriching the errors with some local context.

PR #6773, opened 2023-05-03
agent: Add context to errors that may occur when AgentConfig file is …

jepio:
…missing

When the agent config file is missing, the panic message says "no such file or directory" but doesn't inform the user about which file was missing. Add context to the parsing (with filename) and to the from_config_file() calls (with information where the path is coming from).

Fixes: #6771

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -200,7 +200,7 @@ impl AgentConfig {
         let config_position = args.iter().position(|a| a == "--config" || a == "-c");
         if let Some(config_position) = config_position {
             if let Some(config_file) = args.get(config_position + 1) {
-                return AgentConfig::from_config_file(config_file);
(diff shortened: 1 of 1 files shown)


liubin (approved):

jepio:
Something is systematically wrong about the test, i'm debugging with some changes to kata-containers/tests.

/test-vfio

src/agent/src/config.rs
```diff
@@ -304,7 +305,8 @@ impl AgentConfig {
 
     #[instrument]
     pub fn from_config_file(file: &str) -> Result<AgentConfig> {
-        let config = fs::read_to_string(file)?;
+        let config =
+            fs::read_to_string(file).context(format!("Failed to read config file {}", file))?;
```
Tim-Zhang, line 309: How about using with_context instead? it may avoid a heap allocation.

Issue #6730, opened 2023-04-26
Handle context correctly in StopVM implementation of remote hypervisor


bpradipt:
The initial context is not handled in StopVM - https://github.com/kata-containers/kata-containers/blob/CCv0/src/runtime/virtcontainers/remote.go#L135

PR #6774, opened 2023-05-03
CC | runtime: Add configurable context timeout for StopVM

bpradipt:
Add configurable context timeout for StopVM similar to StartVM

Fixes: #6730

### DIFF
--- a/src/runtime/virtcontainers/remote.go
+++ b/src/runtime/virtcontainers/remote.go
@@ -134,6 +134,15 @@ func (rh *remoteHypervisor) AttestVM(ctx context.Context) error {
 
 func (rh *remoteHypervisor) StopVM(ctx context.Context, waitOnly bool) error {
 
+	// waitOnly doesn't make sense for remote hypervisor and suited for local hypervisor.
+	// Instead use a similar logic like StartVM to handle StopVM with timeout.
+
+	timeout := defaultMinTimeout
+
+	if rh.config.RemoteHypervisorTimeout > 0 {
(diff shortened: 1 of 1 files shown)


src/runtime/virtcontainers/remote.go
```diff
@@ -134,6 +134,10 @@ func (rh *remoteHypervisor) AttestVM(ctx context.Context) error {
 
 func (rh *remoteHypervisor) StopVM(ctx context.Context, waitOnly bool) error {
 
+	// waitOnly doesn't make sense for remote hypervisor and suited for local hypervisor.
+	// Instead use a similar logic like StartVM to handle StopVM with timeout.
+	timeout := int(rh.config.RemoteHypervisorTimeout)
```
jepio, line 139: Does this need the same logic as `StartVM`:
```
	minTimeout := defaultMinTimeout
	if rh.config.RemoteHypervisorTimeout > 0 {
		minTimeout = int(rh.config.RemoteHypervisorTimeout)
	}
```

jepio (approved):

src/runtime/virtcontainers/remote.go
bpradipt: I'll modify the code to use the defaultMinTimeout

fidencio: @bpradipt, if you're already modifying the code, please, also change "Ad" -> "Add" in the commit message.

stevenhorsman (approved):
LGTM, after the fix has gone in (and assuming the tests pass). Thanks Pradipta and Jeremi!

Issue #6777, opened 2023-05-04
cache: Make sure we cache (and use the cached artefact) for the gpu related kernels


fidencio:
Right now @zvonkok added support for building different kernels with nvidia drivers, but those are not yet cached, meaning that every time the GHA runs those are built from scratch, adding unneeded ~20 minutes to each job.

PR #6778, opened 2023-05-04
cache: Update the KERNEL_FLAVOUR list to include nvidia-gpu

fidencio:
We need to make sure that, when caching a `-nvidia-gpu` kernel, we still
look at the version of the base kernel used to build the nvidia-gpu
drivers, as the ${vendor}-gpu kernels are based on already existing
entries in the versions.yaml file and do not require a new entry to be
added.

Fixes: #6777

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -22,9 +22,9 @@ jobs:
           - kernel-sev
           - kernel-dragonball-experimental
           - kernel-tdx-experimental
-          - kernel-gpu
-          - kernel-gpu-snp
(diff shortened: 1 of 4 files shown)


zvonkok:
LGTM agree on the renaming, there are going to be follow up PRs to add CONFIG settings that may be incompatible with other GPU vendors.

fidencio:
The static checks will fail due to the artefacts renaming, it's expected.

stevenhorsman (approved):
Looks reasonable to me. Thanks!

amshinde (approved):

fidencio:
I rebased this one atop of current `main` in order to solve conflicts.

/test

PR #6799, opened 2023-05-11
shim: do not return error if deleting already removed sandbox

tanweernoor:
If the sandbox doesn't exist, let's just return 0 rather than throwing an error back to the caller.

 Fixes: 6796

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/errors.go
+++ b/src/runtime/pkg/containerd-shim-v2/errors.go
@@ -58,7 +58,8 @@ func isInvalidArgument(err error) bool {
 
 func isNotFound(err error) bool {
 	return err == vc.ErrNoSuchContainer || err == syscall.ENOENT ||
-		strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "not exist")
+		strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "not exist") ||
+		strings.Contains(err.Error(), "no such file or directory")
 }
 
 func isGRPCErrorCode(code codes.Code, err error) bool {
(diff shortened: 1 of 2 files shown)


c3d (approved):
With the changes in commit message, looks good to me.

c3d (approved):
With the changes in commit message, looks good to me.

src/runtime/pkg/containerd-shim-v2/errors.go
```diff
@@ -58,7 +58,8 @@ func isInvalidArgument(err error) bool {
 
 func isNotFound(err error) bool {
 	return err == vc.ErrNoSuchContainer || err == syscall.ENOENT ||
-		strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "not exist")
+		strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "not exist") ||
+		strings.Contains(err.Error(), "no such file or directory")
```
c3d, line 62: Note: `isNotFound` is also called in the `StopContainer` case. However, I think that the added `no such file or directory` scenario also applies in that case.

src/runtime/pkg/containerd-shim-v2/service.go
```diff
@@ -445,6 +450,11 @@ func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (_ *taskAP
 
 	start := time.Now()
 	defer func() {
+		// if sandbox/container is not found in path
+		// it means that it is already deleted
+		if err != nil && isNotFound(err) {
```
amshinde, line 455: This may make sense in the kill/cleanup case, but in case of start/exec, if the container is not found or there is an error case, the error needs to be propogated. Setting the err to nil here will hide errors during container lifecycle.

tanweernoor: makes sense, updated

haslersn:
Why was this never merged?

Issue #6727, opened 2023-04-25
Implement functionality to check of host is capable of running a virtual machine


amshinde:
Implement functionality to check of host is capable of running a virtual machine as part of the kata-ctl tool.

PR #6800, opened 2023-05-11
kata-ctl: Check for vm capability

amshinde:
Implement functionality to add to the env output if the host is capable
of running a VM.
    
Fixes: #6727

 Adding vhost and vhost-net to the kernel modules. These do not require
 any kernel module parameters to be checked. Currently, kernel params is
 a required field. Make this as optional. Could make this as <Option>,
 but making this a slice instead, as a module could have multiple kernel
 params. Refactor the function that checks are for kernel modules into
 two with one specifically checking if the module is loaded and other
 checking for module parameters.

### DIFF
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -343,6 +343,23 @@ mod arch_specific {
         }
         Ok(())
     }
+
+    pub fn host_is_vmcontainer_capable() -> Result<bool> {
+        let mut count = 0;
+        if check_cpu("check_cpu").is_err() {
+            count += 1;
+        };
+
+        if check_kernel_modules("check_modules").is_err() {
+            count += 1;
+        };
+
+        if count == 0 {
+            return Ok(true);
+        };
+
+        Err(anyhow!("System is not capable of running a VM"))
+    }
 }
 
 #[cfg(target_arch = "x86_64")]
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -238,13 +238,9 @@ mod arch_specific {
 
         let running_on_vmm_alt = running_on_vmm()?;
 
+        // Kernel param "unrestricted_guest" is not required when running under a hypervisor
         if running_on_vmm_alt {
-            let msg = format!("You are running in a VM, where the kernel module '{}' parameter '{:}' has a value '{:}'. This causes conflict when running kata.",
-                module,
-                param_name,
-                param_value_host
-            );
-            return Err(anyhow!(msg));
+            return Ok(());
         }
 
         if param_value_host == expected_param_value.to_string() {
--- a/src/tools/kata-ctl/src/arch/x86_64/mod.rs
+++ b/src/tools/kata-ctl/src/arch/x86_64/mod.rs
@@ -59,17 +59,29 @@ mod arch_specific {
     static MODULE_LIST: &[KernelModule] = &[
         KernelModule {
             name: "kvm",
-            parameter: KernelParam {
+            params: &[KernelParam {
                 name: "kvmclock_periodic_sync",
                 value: KernelParamType::Simple("Y"),
-            },
+            }],
         },
         KernelModule {
             name: "kvm_intel",
-            parameter: KernelParam {
+            params: &[KernelParam {
                 name: "unrestricted_guest",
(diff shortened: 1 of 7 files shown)


amshinde:
cc @jodh-intel PTAL. I have refactored the code a bit to make the kernel param a list to account for modules that may have none or multiple parameters. I plan to add functionality to mark  modules as required/non-required. WDYT?
I am now using the checks in the env command to determine if the host is capable of running a VM. But I realised that these checks produce output to stdout/err, and this messes the env output. I am planning to refactor further so that these functions return a result and the output for check command is handled higher up. Will need to look into this. Will push a commit to this PR itself.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -234,6 +246,38 @@ mod arch_specific {
...
+                    )
+                })?;
+
+            check_kernel_param(
+                kernel_module.name,
+                param.name,
+                &param_value_host,
+                param.value.clone(),
+            )
+            .map_err(|e| anyhow!("Error : {}", e))?;
```
jodh-intel, line 276: Can you remove the error prefix as that is handled by `main()`:

```suggestion
            .map_err(|e| anyhow!("e))?;
```

jodh-intel (commented):
Thanks @amshinde !

A slight niggle though if the kernel modules aren't loaded:

```bash
$ sudo kata-ctl check all
INFO: check CPU: x86_64
INFO: check kernel modules for: x86_64
kvm Ok
kvm_intel Ok
WARNING The kernel module `vhost` exist but is not already loaded. Try reloading it using 'modprobe vhost=Y'
WARNING The kernel module `vhost_net` exist but is not already loaded. Try reloading it using 'modprobe vhost_net=Y'
WARNING The kernel module `vhost_vsock` exist but is not already loaded. Try reloading it using 'modprobe vhost_vsock=Y'
```

Note that those kernel modules should not have `=Y` appended as that invalidates the names of the modules.

src/tools/kata-ctl/src/arch/x86_64/mod.rs
```diff
@@ -46,17 +46,29 @@ mod arch_specific {
...
+            name: "vhost",
+            params: &[],
+        },
+        KernelModule {
+            name: "vhost_net",
+            params: &[],
+        },
+        KernelModule {
+            name: "vhost_vsock",
+            params: &[],
```
cmaf, line 71: You could consider an Option type here

cmaf (approved):
thanks @amshinde

src/tools/kata-ctl/src/ops/env_ops.rs
```diff
@@ -255,6 +255,12 @@ fn get_host_info() -> Result<HostInfo> {
 
     let guest_protection = guest_protection.to_string();
 
+    let mut vm_container_capable = true;
+
+    if arch_specific::host_is_vmcontainer_capable().is_err() {
+        vm_container_capable = false;
+    }
```
jodh-intel, line 262: Testing this PR in a VM that can run a container, `env` shows `vm_container_capable = false` for me. Adding some debug shows the problem is:

```
check_kernel_modules error: Err(You are running in a VM, where the kernel module 'kvm_intel' parameter 'unrestricted_guest' has a value 'Y'. This causes conflict when running kata.)
```

jodh-intel (commented):
Thanks @amshinde - ptal at my comment.

src/tools/kata-ctl/src/ops/env_ops.rs
jodh-intel: ftr, the QEMU VM I'm running `kata-ctl` inside of is created like this:

```bash
$ qemu-system-x86_64 -hda ./ubuntu-22.04.qcow2 -name ubuntu-22.04.qcow2 -smp cpus=4 -m 8192 -cpu host,host-phys-bits -enable-kvm -net nic,model=virtio -net user,hostfwd=tcp::1826-:22,hostfwd=tcp::1679-:16686 -device virtio-rng-pci -qmp unix:/tmp/qmp.sock,server,nowait -nographic -serial mon:stdio -device vhost-vsock-pci,id=vhost-vsock-pci0,guest-cid=3
```

amshinde: @jodh-intel I have added a fix for the above issue so that it is safe to ignore the `unrestricted_guest` parameter while running in a VM.

jodh-intel (approved):
Thanks @amshinde.

lgtm

Issue #6803, opened 2023-05-11
docs instruct to use a stale Rust version


pmores:
Build instructions for runtime-rs lead users to install Rust 1.62.  Current runtime-rs is not buildable anymore with 1.62 and 1.68 is required.

PR #6804, opened 2023-05-11
runtime-rs: fix building instructions to use correct required Rust ve…

pmores:
…rsion

Fixes: #6803

### DIFF
--- a/docs/install/kata-containers-3.0-rust-runtime-installation-guide.md
+++ b/docs/install/kata-containers-3.0-rust-runtime-installation-guide.md
@@ -49,14 +49,14 @@ Follow the [`kata-deploy`](../../tools/packaging/kata-deploy/README.md).
 
 * Download `Rustup` and install  `Rust`
     > **Notes:**
-    > For Rust version, please see [`versions.yaml`](../../versions.yaml)  file's rust section.
+    > For Rust version, please set `RUST_VERSION` to the value of `languages.rust.meta.newest-version key` in [`versions.yaml`](../../versions.yaml) or, if `yq` is available on your system, run `export RUST_VERSION=$(yq read versions.yaml languages.rust.meta.newest-version)`.
 
     Example for `x86_64`
     ```
     $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
     $ source $HOME/.cargo/env
-    $ rustup install 1.62.0
-    $ rustup default 1.62.0-x86_64-unknown-linux-gnu
+    $ rustup install ${RUST_VERSION}
+    $ rustup default ${RUST_VERSION}-x86_64-unknown-linux-gnu
     ```
 
 * Musl support for fully static binary


stevenhorsman (commented):
I'm not against this change, but I don't think the docs are currently completely wrong. In order to not have to keep having to bump the version each time the documentation does say 
> Download Rustup and install Rust
> 
> Notes: For Rust version, please see [versions.yaml](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml) file's rust section.
> 
> Example for x86_64
> 
> ```
> $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
> $ source $HOME/.cargo/env
> $ rustup install 1.62.0
> $ rustup default 1.62.0-x86_64-unknown-linux-gnu
> ```

So it tells users to check the versions.yaml file for the version to use.

I guess my question is do we want to update this doc ever time we bump the versions.yaml?
Or is there something else we can do to clarify it e.g. remove a specific version from the doc completely by having something like:

> Example for x86_64
> ```
> $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
> $ source $HOME/.cargo/env
> $ rustup install <rust version>
> $ rustup default <rust version>-x86_64-unknown-linux-gnu
> ```

pmores:
I get the point, thanks, and I'm not insisting on any particular solution.  I do think that having a stale version literal in the docs is rather misleading but how to handle this is probably a matter of taste and practical consideration that I'm not necessarily familiar with. For instance, if Rust version is bumped once a year I'd guess making a corresponding docs update as a part of the procedure would likely be acceptable, if it's done every month then probably not so much.

stevenhorsman:
> I get the point, thanks, and I'm not insisting on any particular solution. I do think that having a stale version literal in the docs is rather misleading but how to handle this is probably a matter of taste and practical consideration that I'm not necessarily familiar with. For instance, if Rust version is bumped once a year I'd guess making a corresponding docs update as a part of the procedure would likely be acceptable, if it's done every month then probably not so much.

Yeah, I'm sorry if my comment came across as dismissive - I completely agree that the doc could be better, I was just trying to start a conversation about alternatives.

I've had a quick look at it seems like the Rust version has been bumped 4 times in the last year, but rust has a six-week release cadence and we might want to be more proactive about bumping some of our dependencies, so it depends if we can remember to update the doc as part of the PRs that bump the rust version - we could possibly add a note into the versions.yaml to remind any 'bumpers' to match the doc if we go down that route?

pmores:
Now it's my turn to apologise, if I sounded somehow offended that wasn't my intention at all (I'm not a native English speaker).  Anyway, as someone who got mislead by the obsolete version literal just this morning, the important distinction to me would be to either make sure that the version in the docs is always up-to-date, or not have a version literal in the docs at all.

In that sense, what you suggested above:

```
$ rustup install <rust version>
$ rustup default <rust version>-x86_64-unknown-linux-gnu
```
coupled with clear instructions how to get `<rust version>` (ideally in the previous paragraph, at any case not too far away) would totally solve the problem from my perspective as well.

pmores:
So how about:
- change the Note text to "For Rust version, please see the `languages.rust.meta.newest-version` key in  [versions.yaml](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml)"
- if `yq` usage is acceptable, add "... or run `yq read versions.yaml languages.rust.meta.newest-version`."
- and replace the version string literal in the example commands with `<rust version>` as you proposed?

stevenhorsman:
> I hesitated about `yq` usage since it doesn't seem to included in all distros but if `yq` availability is a safe assumption then `export RUST_VERSION=$(yq read versions.yaml languages.rust.meta.newest-version)` is perfect and fits nicely with Kata docs habit of giving literal scripts to do things.
> 
> I'll make the changes and re-push.

Yes, I think adding the caveat of ("if `yq` is available on your platform, ...."). I think we use yq in a fair number of the kata scripts to do exactly this thing, so there is a chance it might be installed for developers.

Thanks very much!

pmores:
Pushed a new version that includes what we agreed on (I believe :-)).  Please take a look and let me know if further changes are necessary.

Tim-0731-Hzt (approved):
Thanks @pmores LGTM

studychao (approved):
LGTM, but there is a commit Sign-off-by complain in the CI, please help fix it.

The reason for that is the author of the commit (pvl@ibm-p8-06-fsp.mgmt.pnr.lab.eng.rdu2.redhat.com) and the signed-off-by (pmores@redhat.com) is not the same.

stevenhorsman (approved):
LGTM. Thanks for the update @pmores!

PR #6806, opened 2023-05-11
Dragonball: support vcpu hotplug on aarch64

### LINKED ISSUE #6010: Dragonball: add cpu resize ability for aarch64 platform
We will add cpu resize aarch64 platform support with upcall

xuejun-xj:
This PR adds support of vcpu hotplug/hotunplug on aarch64. It is implemented with dragonball upcall.
Besides, add three patches of guest kernel in dragonball-experimental directory.

fix #6010


jongwu:
Thanks @xuejun-xj -, some questions here:

- As there is no ACPI enabled, DeviceTree is used to pass device info to kernel. But DeviceTree only containes the boot cpu info. As to enable Cpu Hotplug, should we consider the impact on DT? If pass max cpu number from cmdline, is there any conflict with DT?
- Is there any plan to post upcall solusion to kernel community?

xuejun-xj:
> Thanks @xuejun-xj -, some questions here:
> 
> * As there is no ACPI enabled, DeviceTree is used to pass device info to kernel. But DeviceTree only containes the boot cpu info. As to enable Cpu Hotplug, should we consider the impact on DT? If pass max cpu number from cmdline, is there any conflict with DT?
> * Is there any plan to post upcall solusion to kernel community?

hello @jongwu ! Thanks for your questions.
For the first question:
1. When we initialize vcpu_manager, `self.vcpu_infos` is initialized with `max_vcpu_count` size.
2. Create vcpu call trace: init_microvm -> create_boot_vcpus -> create_vcpus -> create_vcpu. In this PR, we will pass `max_vcpu_count` as boot vcpu number to `create_vcpus` method, and then dragonball creates `max_vcpu_count` vcpufds. In `create_vcpu`, `Vcpu` struct, which contains vcpufd, will be saved in `self.vcpu_infos`.
3. In fdt, when creating cpu nodes, we iterates `vcpu_mpidr` vector which is create from `vcpu_manager.vcpus()`. This method uses `self.vcpu_infos` to return the `Vcpu` vector. As I have explained above, `self.vcpu_infos` is correctly initialized with `max_vcpu_count`, and fdt won't face the conflict of vcpu numbers.
4. As we have pass the max_vcpu_count into guest from cmdline, we may see this situation in guest sysfs:
    > In `/sys/devices/system/cpu/` directory, it contains `max_vcpu_count` cpuX directory.
    > Only a few cpu is online, it is decided by boot_vcpu_count.

src/dragonball/src/vm/aarch64.rs
```diff
@@ -96,6 +97,30 @@ impl Vm {
...
+            .kernel_cmdline_mut();
+
+        cmdline
+            .insert(
+                "dragonball_bootcpus",
+                &format!("{}", self.vm_config.vcpu_count),
+            )
+            .map_err(StartMicroVmError::AddCmdlineCpuCountError)?;
+        cmdline
+            .insert("maxcpus", &format!("{}", self.vm_config.max_vcpu_count))
```
jongwu, line 116: IMO, maxcpus means present cpu number not the possible cpu but you want to pass the possible cpu number here. Am I right?

studychao:
> Thanks @xuejun-xj -, some questions here:
> 
> * As there is no ACPI enabled, DeviceTree is used to pass device info to kernel. But DeviceTree only containes the boot cpu info. As to enable Cpu Hotplug, should we consider the impact on DT? If pass max cpu number from cmdline, is there any conflict with DT?
> * Is there any plan to post upcall solusion to kernel community?

Hi @jongwu ,for the second question, currently our main focus is to enable upcall in Kata Container community through experimental kernel. 
We have plan for post the solution to kernel community but currently we don't have exact schedule for that.

jongwu:
> > Thanks @xuejun-xj -, some questions here:
> > 
> > * As there is no ACPI enabled, DeviceTree is used to pass device info to kernel. But DeviceTree only containes the boot cpu info. As to enable Cpu Hotplug, should we consider the impact on DT? If pass max cpu number from cmdline, is there any conflict with DT?
> > * Is there any plan to post upcall solusion to kernel community?
> 
> Hi @jongwu ,for the second question, currently our main focus is to enable upcall in Kata Container community through experimental kernel. We have plan for post the solution to kernel community but currently we don't have exact schedule for that.

@studychao -, OK, Thanks your reply.

src/dragonball/src/vm/aarch64.rs
xuejun-xj: "maxcpus" here is used to initialize `setup_max_cpus` variable in kernel. See [maxcpus cmdline](https://github.com/torvalds/linux/blob/f1fcbaa18b28dec10281551dfe6ed3a3ed80e3d6/kernel/smp.c#L920).
According to the comment in kernel, I think you're right. See [setup_max_cpus definition](https://github.com/torvalds/linux/blob/f1fcbaa18b28dec10281551dfe6ed3a3ed80e3d6/kernel/smp.c#L879).

jongwu: Thanks @xuejun-xj -,

So this solusion replies on the "bootvcpu" and "maxvcpu" passed from cmdline which is customized for kata/DB and is hard for a common scenario eg. for a standalone VM.

Is it possible to reuse DT to pass those info then we can extend this feature for a wider use case?

xuejun-xj: Yeah, that's a good idea. Thanks @jongwu :)
As DT contains `max_vcpu_count` cpu nodes, how about adding an attribute to indicate whether this cpu should be activated at booting procedure?
The core problem is to pass the correct "boot_vcpu_count" to bringup_nonboot_cpus function. Do you have any better suggestions about this?
By the way, do you know what's the difference between possible cpu and present cpu? In which case the number of them are unequal.

jongwu: > Yeah, that's a good idea. Thanks @jongwu :) As DT contains `max_vcpu_count` cpu nodes, how about adding an attribute to indicate whether this cpu should be activated at booting procedure?
> 
I think it make sense.
> 
> The core problem is to pass the correct "boot_vcpu_count" to bringup_nonboot_cpus function. Do you have any better suggestions about this?
> 
Maybe this issue can be resolved gracefully by getting boot_cpu number from DT if we can extend use of DT (may violate DT spec)
> 
> By the way, do you know what's the difference between possible cpu and present cpu? And in which case the number of them are unequal?
> 
The meaning for possible & present cpu in x86 is easy to understand. The present cpu means the plugged cpu. The possible cpu means the plugged cpu number plus the cpu number that will be plugged in the future. There is a option "smp" in qemu, like: `-smp 3,sockets=1,cores=15,threads=1,maxcpus=15` where "3" means present cpu number, maxcpus means possible cpu number. At first, there is only 3 vcpu in the guest, for now the preset vcpu is not equal to possible cpu, unless you hotplug 12 more vcpu into guest.
 
As Jame's patch about "support vcpu hotplug on arm64" for linux kernel is under review, where the meaning of possible cpu and present cpu is changed but is confusing on arm side. We can leave it aside for now, I think.

xuejun-xj: @jongwu Thanks for explanation! :)
> Maybe this issue can be resolved gracefully by getting boot_cpu number from DT if we can extend use of DT (may violate DT spec)

I will update the implementation about this extension recently.

> The meaning for possible & present cpu in x86 is easy to understand. The present cpu means the plugged cpu. The possible cpu means the plugged cpu number plus the cpu number that will be plugged in the future. There is a option "smp" in qemu, like: -smp 3,sockets=1,cores=15,threads=1,maxcpus=15 where "3" means present cpu number, maxcpus means possible cpu number. At first, there is only 3 vcpu in the guest, for now the preset vcpu is not equal to possible cpu, unless you hotplug 12 more vcpu into guest.
> 
> As Jame's patch about "support vcpu hotplug on arm64" for linux kernel is under review, where the meaning of possible cpu and present cpu is changed but is confusing on arm side. We can leave it aside for now, I think.

May I understand like this?
- possible_cpu: The max cpu number that can be plugged. Maybe it is decided physical cpu devices?
- present_cpu: The max cpu number that current system supports.
- online_cpu: The cpu number that is currently running on the system.
Am I right?

xuejun-xj: Hi @jongwu. I have thinked of two available solutions:
1. Use a single attribute, like "boot-activated-cpu-number", under /cpus/ node.
2. Use an attribute, like "boot-activated", under each /cpus/cpuX/ node.

IMO, maybe the first solution is better. Is it necessary to use attribute under each /cpus/cpuX? What do you think about it?

jongwu: > @jongwu Thanks for explanation! :)
> 
> > Maybe this issue can be resolved gracefully by getting boot_cpu number from DT if we can extend use of DT (may violate DT spec)
> 
> I will update the implementation about this extension recently.
> 
> > The meaning for possible & present cpu in x86 is easy to understand. The present cpu means the plugged cpu. The possible cpu means the plugged cpu number plus the cpu number that will be plugged in the future. There is a option "smp" in qemu, like: -smp 3,sockets=1,cores=15,threads=1,maxcpus=15 where "3" means present cpu number, maxcpus means possible cpu number. At first, there is only 3 vcpu in the guest, for now the preset vcpu is not equal to possible cpu, unless you hotplug 12 more vcpu into guest.
> > As Jame's patch about "support vcpu hotplug on arm64" for linux kernel is under review, where the meaning of possible cpu and present cpu is changed but is confusing on arm side. We can leave it aside for now, I think.
> 
> May I understand like this?
> 
> * possible_cpu: The max cpu number that can be plugged. Maybe it is decided physical cpu devices?
> * present_cpu: The max cpu number that current system supports.
> * online_cpu: The cpu number that is currently running on the system.
> 
> Am I right?

Here are some reference for you:

- The previous implementation in qemu from Huawei
   - https://patchwork.kernel.org/project/qemu-devel/cover/20200613213629.21984-1-salil.mehta@huawei.com/
- patch in review, kernel from Arm:
   - https://lwn.net/Articles/922127/
- qemu from Huawei
   - https://github.com/salil-mehta/qemu/tree/virt-cpuhp-armv8/rfc-v1-port29092022

jongwu: Another concern is if the implemention here will conflict with the future implemention from kernel. We need take care about it.

xuejun-xj: > Another concern is if the implemention here will conflict with the future implemention from kernel. We need take care about it.

Thanks for your suggestions! I will browse the reference and handle it carefullly.

xuejun-xj:
Hi @jongwu , I have pushed new series. Please review.

It does these changes:
1. Update the comment of init_microvm method.
2. Update dbs-boot to v0.4.0. It simplify the interface of create_fdt. And `configure_system` is never needed. By the way, it also provides reserved interfaces needed by #6969.
3. Construct a vecter called `vcpu_boot_onlined` to pass into fdt, indicating whether this cpu should be onlined at first boot. Temporarily, it simply thinks the first vcpu_count vCPUs should be onlined.
4. According to the fdt changes, kernel should parse "boot-onlined" property and pass correct cpu count into bringup_nonboot_cpus.

The changes in dragonball-sandbox is in this PR: https://github.com/openanolis/dragonball-sandbox/pull/280

src/dragonball/src/vm/aarch64.rs
```diff
@@ -138,6 +103,50 @@ impl Vm {
...
+        let vcpu_boot_onlined = vec![];
+        let vpmu_feature = vcpu_manager.vpmu_feature();
+        // This configuration is used for passing cache information into guest.
+        // TODO: dragonball-sandbox #274; kata-containers #6969
+        let cache_passthrough_enabled = false;
+        let fdt_vcpu_info = FdtVcpuInfo::new(
+            vcpu_mpidr,
+            vcpu_boot_onlined,
+            vpmu_feature,
+            cache_passthrough_enabled,
```
jongwu, line 131: Out of curiosity, what's cache_passthrough?

xuejun-xj: This is used as a switch to pass cache information into guest. See #6969.

tools/packaging/kernel/patches/5.10.x/dragonball-experimental/0007-smp-update-bringup_nonboot_cpus-parameters.patch
```diff
@@ -0,0 +1,136 @@
...
++}
++
+ static u64 __init of_get_cpu_mpidr(struct device_node *dn)
+ {
+ 	const __be32 *cell;
+@@ -654,6 +673,7 @@ static void __init of_parse_and_init_cpus(void)
+ 	struct device_node *dn;
+ 
+ 	for_each_of_cpu_node(dn) {
++		of_get_cpu_boot_onlined(dn);
```
jongwu, line 98: nit: will always print err in non-DB senario.

src/dragonball/src/vm/aarch64.rs
jongwu: OK

tools/packaging/kernel/patches/5.10.x/dragonball-experimental/0007-smp-update-bringup_nonboot_cpus-parameters.patch
xuejun-xj: Maybe wrapper it with a CONFIG_DRAGONBALL_HOTPLUG_CPU?

jongwu (approved):
Looks good to me. Thanks! @xuejun-xj

tools/packaging/kernel/patches/5.10.x/dragonball-experimental/0007-smp-update-bringup_nonboot_cpus-parameters.patch
jongwu: Yeah, it's better.

xuejun-xj:
> Looks good to me. Thanks! @xuejun-xj

My pleasure:)

tools/packaging/kernel/patches/5.10.x/dragonball-experimental/0007-smp-update-bringup_nonboot_cpus-parameters.patch
```diff
@@ -0,0 +1,136 @@
...
+index 25240fb2df94..1b305a03582e 100644
+--- a/kernel/smp.c
++++ b/kernel/smp.c
+@@ -801,17 +801,26 @@ void __init setup_nr_cpu_ids(void)
+ 	nr_cpu_ids = find_last_bit(cpumask_bits(cpu_possible_mask),NR_CPUS) + 1;
+ }
+ 
++/* Setup number of CPUs to activate */
++unsigned int boot_onlined_cpu = 0;
++EXPORT_SYMBOL(boot_onlind_cpu);
```
studychao, line 112: typo?`boot_onlind_cpu`

studychao (approved):
LGTM except one typo

tools/packaging/kernel/patches/5.10.x/dragonball-experimental/0007-smp-update-bringup_nonboot_cpus-parameters.patch
xuejun-xj: Thanks, @studychao. As there is no module will use this symbol, I have delete this line in new push.

PR #6807, opened 2023-05-11
cache: Fix nvidia-gpu version

fidencio:
c9bf7808b6345c7902c46ad11204fc076fb7b894 introduced the logic to properly get the version of nvidia-gpu kernels, but one important part was dropped during the rebase into main, which is actually getting the correct version of the kernel.

Fixing this now, and using the old issue as reference.

Fixes: #6777


stevenhorsman (approved):
LGTM

jepio (approved):

Issue #6555, opened 2023-03-30
CI | GHA | Create a PoC for running VFIO tests on external runners


fidencio:
A proof-of-concept has to be created to run the equivalent of `make vfio` tests, but using GitHub actions, on an external runner.

The expected outcome is:
* A PR with the PoC
* Issues created for the failing tests per architecture
* The owner of this task is responsible for creating new issues for each one of the targets being tested.

PR #6808, opened 2023-05-11
gha: Add a skeleton for the VFIO job

fidencio:
As the very first step to migrate the VFIO test job from Jenkins to GitHub Actions, let's just create a very basic skeleton for the job, which will do nothing and always pass.

Once this is merged, we can start working on the `tests/vfio/run.sh` script, which then will be responsible for performing the tests, and we'll be targetting a 1:1, bug-by-bug, implementation, when compared to the one used in jenkins.

Fixes: #6555


fidencio:
Dropping this one after a convo with @jepio, as we'll get the VFIO job to run on a Azure VM.

The approach will be slower, but the fact we don't have to rely on a single BM machine that's also used for a different job (the metrics one) will for sure compensate the sloweness.

PR #6810, opened 2023-05-11
gha: Enable SEV-SNP tests on main

fitzthum:
Enable SNP tests. Skip the CPU constraints tests due to hotplug limitations.

### DIFF
--- a/tests/integration/kubernetes/k8s-cpu-ns.bats
+++ b/tests/integration/kubernetes/k8s-cpu-ns.bats
@@ -11,7 +11,9 @@ load "${BATS_TEST_DIRNAME}/tests_common.sh"
 setup() {
 	[ "${KATA_HYPERVISOR}" == "firecracker" ] && skip "test not working see: ${fc_limitations}"
 	[ "${KATA_HYPERVISOR}" == "dragonball" ] && skip "test not working see: ${dragonball_limitations}"
-	[ "${KATA_HYPERVISOR}" == "qemu-tdx" ] && skip "TEEs do not support memory / CPU hotplug"
+	( [ "${KATA_HYPERVISOR}" == "qemu-tdx" ] || [ "${KATA_HYPERVISOR}" == "qemu-snp" ] || [ "${KATA_HYPERVISOR}" == "qemu-sev" ] ) \
+		&& skip "TEEs do not support memory / CPU hotplug"
+
 
 	pod_name="constraints-cpu-test"
 	container_name="first-cpu-container"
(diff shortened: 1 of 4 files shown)


tests/integration/kubernetes/k8s-cpu-ns.bats
```diff
@@ -12,6 +12,7 @@ setup() {
 	[ "${KATA_HYPERVISOR}" == "firecracker" ] && skip "test not working see: ${fc_limitations}"
 	[ "${KATA_HYPERVISOR}" == "dragonball" ] && skip "test not working see: ${dragonball_limitations}"
 	[ "${KATA_HYPERVISOR}" == "qemu-tdx" ] && skip "TEEs do not support memory / CPU hotplug"
+	[ "${KATA_HYPERVISOR}" == "qemu-snp" ] && skip "Memory / CPU hotplug support not implemented for SEV-SNP"
```
fidencio, line 15: Could you extend the conditional and / or use the same message for both TDX / SNP / SEV?

tests/integration/kubernetes/k8s-cpu-ns.bats
```diff
@@ -29,6 +30,7 @@ setup() {
...
```
fidencio, line 33: Same here.

tests/integration/kubernetes/k8s-cpu-ns.bats
```diff
@@ -74,6 +76,7 @@ teardown() {
...
```
fidencio, line 79: Same here.

fidencio (changes requested):
Tests passed, yay!

I've left one comment, and I think you could, as a different commit but still part of the same PR, enable the SEV tests as well.

WDYT?

fidencio (changes requested):
Tests passed, yay!

I've left one comment, and I think you could, as a different commit but still part of the same PR, enable the SEV tests as well.

WDYT?

fidencio (approved):
lgtm, thanks @fitzthum!

fitzthum:
Passing with SEV and SNP. Not sure about issue with `run-k8s-tests-on-aks / run-k8s-tests (qemu)`. Is that common? 

Looks like a lot of the Jenkins jobs are getting stuck looking for apt packages.

fidencio:
> Passing with SEV and SNP. Not sure about issue with `run-k8s-tests-on-aks / run-k8s-tests (qemu)`. Is that common?

Common is a hard word. :-)
It's been flaky and we need to debug that.  Whenever you see those errors, please, open an issue pointing to which test is failing and that will help me to debug it further.

> Looks like a lot of the Jenkins jobs are getting stuck looking for apt packages.

Yeah, Yesterday was not a good day for `apt`.  Let's hope Today it does better.

fidencio:
@fitzthum, I've solved the conflicts, as those were originated by one of my PRs that got merged first, and force-pushed to your branch.

stevenhorsman (approved):
LGTM assuming that the tests currently running pass. Thanks @fitzthum

fidencio:
For the first time, all the tests were executed in the TEEs we support on x86_64. \o/

PR #6811, opened 2023-05-11
cache: More fixes to nvidia-gpu kernels caching

fidencio:
---

cache: Fix nvidia-gpu-tdx-experimental cache URL

We were passing "kernel-nvidia-gpu-tdx", missing the "-experimental"
part, leading to a non-valid URL.

---

cache: Fix nvidia-snp caching version

All the kernel-foo instances, such as "kernel-sev" or "kernel-snp",
should be transformed into "kernel.foo" when looking at the
versions.yaml file.

This was already done for SEV, but missed on the SNP case.

Fixes: #6777 

---


jepio (approved):

stevenhorsman (approved):
LGTM. Thanks

Issue #6814, opened 2023-05-12
CC | actions: Making runtime payload builds multi-arch


stevenhorsman:
Currently our [runtime payload](https://quay.io/repository/confidential-containers/runtime-payload) and [runtime-payload-ci](https://quay.io/repository/confidential-containers/runtime-payload-ci?tab=tags) have a multi-arch manifest for latest, but not for individual commit tags. As we reference these tags in our operator installs, it means we have an x86 and s390x version of the [ccruntime](https://github.com/confidential-containers/operator/tree/main/config/samples/ccruntime) and as I'm looking to get peer pods using the coco operator properly we have the same issue there where we will need peer-pods kustomize's for each architecture to reference the runtime-payload. I can see this problem getting worse if we add other architecture e.g. ARM where we have a proliferation of architectures multiplying up every cc runtime type (I do appreciate that we have the s390x version to modify the supported runtimeClasses, not just the runtime-payload).

Instead of getting into this situation I think we should make the runtime-payload and runtime-payload-ci commit images multi-arch, so then we could just point to that without have to have the -s390x  and -x86_64 appended in the kustomize yamls and reduced how many ccruntimes we'll need.

I think this will also be useful in `main` kata-containers, so once we have proved this process on `CCv0` I'll probably do similar to the `kata-deploy-ci`

stevenhorsman:
Closed with #6815 merging

stevenhorsman:
The running of this workflow after it merged revealed another problem where we having the inconsistent architecture tagging in our commit builds, so I've raised https://github.com/kata-containers/kata-containers/issues/6820 to cover fixing this in `main` and will then cherry-pick it into `CCv0` once it has been tested there.

PR #6815, opened 2023-05-12
actions:  multi-arch runtime-payload

stevenhorsman:
- Create multi-arch manifests for the ci and release runtime-payload that are tagged with the commit, for use in the operator

Fixes: #6814


fidencio (approved):
lgtm, thanks @stevenhorsman!

Issue #6820, opened 2023-05-12
kata-deploy: Inconsistency in image tagging


stevenhorsman:
When building the kata-deploy payload [images](https://quay.io/repository/kata-containers/kata-deploy-ci?tab=tags) we have inconsistency in the tagging.

The two tags we push/update per architecture for each workflow are `kata-containers-<commit-sha>-<arch>` and update `kata-containers-<arch>`, however for `x86_64/amd64` are are inconsistent with the architecture, so produce `kata-containers-7fdaab49bcce816cd0f928677cf67dcea9548ec3-x86_64` and `kata-containers-amd64` 

`amd64`seems to be the standard arch for container images, so we should switch to that.

stevenhorsman:
There is a bug in the PR I made for this, so re-opening.

PR #6821, opened 2023-05-12
deploy: Fix arch in image tag

stevenhorsman:
`uname -m` produces `x86_64`, but container image convention is to use `amd64`, so update this in the tag

Fixes: #6820


fidencio (approved):
lgtm, thanks @stevenhorsman!

gkurz (approved):
LGTM. Thanks @stevenhorsman !

Issue #6770, opened 2023-05-03
kata-deploy: Instead of building Firecracker, use the binaries provided by the project's official release


fidencio:
https://github.com/firecracker-microvm/firecracker/releases provides the binaries for x86_64 and aarch64.  With this in mind, let's start consuming the official binaries instead of building our own.

PR #6825, opened 2023-05-12
kata-deploy: Build improvements

fidencio:
Let's start downloading, instead of building, the binaries for Cloud Hypervisor (aarch64) and Firecracker.

As we're here, we've also fixed the nydus package download for aarch64.

Fixes: #6770

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -436,6 +436,8 @@ install_virtiofsd() {
 
 # Install static nydus asset
 install_nydus() {
+	[ "${ARCH}" == "aarch64" ] && ARCH=arm64
+
 	install_cached_tarball_component \
 		"nydus" \
 		"${jenkins_url}/job/kata-containers-main-nydus-$(uname -m)/${cached_artifacts_path}" \
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -377,10 +377,12 @@ install_qemu_snp_experimental() {
 
 # Install static firecracker asset
 install_firecracker() {
+	local firecracker_version=$(get_from_kata_deps "assets.hypervisor.firecracker.version")
+
 	install_cached_tarball_component \
 		"firecracker" \
 		"${jenkins_url}/job/kata-containers-main-firecracker-$(uname -m)/${cached_artifacts_path}" \
-		"$(get_from_kata_deps "assets.hypervisor.firecracker.version")" \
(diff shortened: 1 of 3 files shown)


tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -436,6 +436,8 @@ install_virtiofsd() {
 
 # Install static nydus asset
 install_nydus() {
+	[ "${ARCH}" == "aarch64" ] && ARCH=arm64
+
 	install_cached_tarball_component \
 		"nydus" \
 		"${jenkins_url}/job/kata-containers-main-nydus-$(uname -m)/${cached_artifacts_path}" \
```
stevenhorsman, line 445: I wonder if it's worth updating the name of the nydus cache job from kata-containers-main-nydus-aarch64 to kata-containers-main-nydus-arm64 to match with the nydus release?

stevenhorsman (approved):
- I can see from the nydus release that we are using (https://github.com/dragonflyoss/image-service/releases/tag/v2.2.0) that the release has the naem `arm64.tgz`
- I can see that the CLH version we are using (https://github.com/cloud-hypervisor/cloud-hypervisor/releases/tag/v31.0) has an aarch64 release
- I can see from the Firecracker release https://github.com/firecracker-microvm/firecracker/releases/tag/v1.3.1 that there are aarch64 and x86_64 binaries, which I believe are the only supported platforms.

Just one minor nit, but LGTM. Thanks!

stevenhorsman (approved):
- I can see from the nydus release that we are using (https://github.com/dragonflyoss/image-service/releases/tag/v2.2.0) that the release has the naem `arm64.tgz`
- I can see that the CLH version we are using (https://github.com/cloud-hypervisor/cloud-hypervisor/releases/tag/v31.0) has an aarch64 release
- I can see from the Firecracker release https://github.com/firecracker-microvm/firecracker/releases/tag/v1.3.1 that there are aarch64 and x86_64 binaries, which I believe are the only supported platforms.

Just one minor nit, but LGTM. Thanks!

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
fidencio: That's a good point, but I think it'd be then incosistent with the rest of the jobs.

I have no preference here, as I think both cases are not the optimal ones.  WDYT?

stevenhorsman: I'm fine with it as it currently is, as you say neither solutions are perfect given the inconsistency in nydus, so I thought the comment was worth discussing about whether we align with nydus, or our other jobs, but I wasn't convinced that would be better!

bergwolf (approved):

PR #6839, opened 2023-05-15
tests: Enable running k8s tests on Mariner

### LINKED ISSUE #6840: CI | GHA | mariner | Enable running k8s tests on Mariner
Because of a GHA limitation, we had to split the Mariner PR into two. This second issue is related to the actual test code, rather than the workflow YAML. The goal is to adapt the test code to enable the tests to actually run on Mariner.

sprt:
Depends on: ~#6686~ ~#7027~

This PR adapts the test code so that we can actually run k8s tests on Mariner. The change is essentially summarized as below:

1) We allow building the Mariner rootfs initrd as part of CI, and we specify it at test time using an annotation.
2) Again using an annotation, we specify the custom bleeding-edge kernel used by Mariner. This is enabled by setting the `enable_annotations` config in `kata-deploy.sh`.
3) Build Cloud Hypervisor with glibc and the `mshv` feature for Mariner.

Fixes #6840.

### DIFF
--- a/tests/integration/kubernetes/run_kubernetes_tests.sh
+++ b/tests/integration/kubernetes/run_kubernetes_tests.sh
@@ -54,10 +54,6 @@ else
 	)
 fi
 
-if [ ${KATA_HOST_OS} == "cbl-mariner" ]; then
-	exit 0
-fi
-
 # we may need to skip a few test cases when running on non-x86_64 arch
 arch_config_file="${kubernetes_dir}/filter_out_per_arch/${TARGET_ARCH}.yaml"
 if [ -f "${arch_config_file}" ]; then
--- a/tests/integration/kubernetes/setup.sh
+++ b/tests/integration/kubernetes/setup.sh
@@ -14,6 +14,13 @@ set_runtime_class() {
     sed -i -e "s|runtimeClassName: kata|runtimeClassName: kata-${KATA_HYPERVISOR}|" ${kubernetes_dir}/runtimeclass_workloads/*.yaml
 }
 
+set_kernel_path() {
+    if [[ "${KATA_HOST_OS}" = "cbl-mariner" ]]; then
+        mariner_kernel_path="/usr/share/cloud-hypervisor/vmlinux.bin"
+        find ${kubernetes_dir}/runtimeclass_workloads/*.yaml -exec yq write -i {} 'metadata.annotations[io.katacontainers.config.hypervisor.kernel]' "${mariner_kernel_path}" \;
+    fi
+}
+
 set_initrd_path() {
     if [[ "${KATA_HOST_OS}" = "cbl-mariner" ]]; then
         initrd_path="/opt/kata/share/kata-containers/kata-containers-initrd-cbl-mariner.img"
@@ -24,6 +31,7 @@ set_initrd_path() {
 main() {
     INSTALL_IN_GOPATH=false bash "${repo_root_dir}/ci/install_yq.sh"
     set_runtime_class
+    set_kernel_path
     set_initrd_path
 }
 
--- a/tests/integration/kubernetes/setup.sh
+++ b/tests/integration/kubernetes/setup.sh
@@ -8,13 +8,23 @@ set -o nounset
 set -o pipefail
 
 kubernetes_dir=$(dirname "$(readlink -f "$0")")
+repo_root_dir="$(cd "${kubernetes_dir}/../../../" && pwd)"
 
 set_runtime_class() {
     sed -i -e "s|runtimeClassName: kata|runtimeClassName: kata-${KATA_HYPERVISOR}|" ${kubernetes_dir}/runtimeclass_workloads/*.yaml
 }
 
+set_initrd_path() {
+    if [[ "${KATA_HOST_OS}" = "cbl-mariner" ]]; then
+        initrd_path="/opt/kata/share/kata-containers/kata-containers-initrd-cbl-mariner.img"
+        find ${kubernetes_dir}/runtimeclass_workloads/*.yaml -exec yq write -i {} 'metadata.annotations[io.katacontainers.config.hypervisor.initrd]' "${initrd_path}" \;
+    fi
+}
+
 main() {
+    INSTALL_IN_GOPATH=false bash "${repo_root_dir}/ci/install_yq.sh"
     set_runtime_class
+    set_initrd_path
 }
 
 main "$@"
--- a/tests/integration/kubernetes/setup.sh
+++ b/tests/integration/kubernetes/setup.sh
@@ -29,7 +29,6 @@ set_initrd_path() {
 }
 
 main() {
-    INSTALL_IN_GOPATH=false bash "${repo_root_dir}/ci/install_yq.sh"
     set_runtime_class
     set_kernel_path
     set_initrd_path
--- a/tools/packaging/guest-image/build_image.sh
+++ b/tools/packaging/guest-image/build_image.sh
@@ -22,45 +22,44 @@ readonly osbuilder_dir="$(cd "${repo_root_dir}/tools/osbuilder" && pwd)"
 export GOPATH=${GOPATH:-${HOME}/go}
 
 arch_target="$(uname -m)"
-final_image_name="kata-containers"
-final_initrd_name="kata-containers-initrd"
+final_artifact_name="kata-containers"
 image_initrd_extension=".img"
 
 build_initrd() {
 	info "Build initrd"
-	info "initrd os: $initrd_distro"
-	info "initrd os version: $initrd_os_version"
+	info "initrd os: $os_name"
+	info "initrd os version: $os_version"
 	sudo -E PATH="$PATH" make initrd \
-		DISTRO="$initrd_distro" \
+		DISTRO="$os_name" \
 		DEBUG="${DEBUG:-}" \
-		OS_VERSION="${initrd_os_version}" \
+		OS_VERSION="${os_version}" \
 		ROOTFS_BUILD_DEST="${builddir}/initrd-image" \
 		USE_DOCKER=1 \
 		AGENT_INIT="yes"
-	mv "kata-containers-initrd.img" "${install_dir}/${initrd_name}"
+	mv "kata-containers-initrd.img" "${install_dir}/${artifact_name}"
 	(
 		cd "${install_dir}"
-		ln -sf "${initrd_name}" "${final_initrd_name}${image_initrd_extension}"
+		ln -sf "${artifact_name}" "${final_artifact_name}${image_initrd_extension}"
 	)
 }
 
 build_image() {
 	info "Build image"
-	info "image os: $img_distro"
-	info "image os version: $img_os_version"
+	info "image os: $os_name"
+	info "image os version: $os_version"
 	sudo -E PATH="${PATH}" make image \
-		DISTRO="${img_distro}" \
+		DISTRO="${os_name}" \
 		DEBUG="${DEBUG:-}" \
 		USE_DOCKER="1" \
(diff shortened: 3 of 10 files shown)


tests/integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -62,6 +62,10 @@ if [ ${KATA_HYPERVISOR} == "qemu-snp" ]; then
 	exit 0
 fi
 
+if [ ${KATA_HOST_OS} == "cbl-mariner" ]; then
+	exit 0
+fi
+
```
sprt, line 68: Will revert this after #6686 is merged.

tools/packaging/guest-image/build_image.sh
```diff
@@ -80,28 +86,43 @@ EOF
 }
 
 main() {
+	local host_oses=("cbl-mariner" "ubuntu")
```
wainersm, line 89: I think we will need to re-think the build of rootfs with marine as it is breaking the standard.

Currently, from a developer/CI point of view, the build process starts with a make command like:
```
$ rootfs-initrd-tarball
```

The kata-deploy's makefile targets are defined in `tools/packaging/kata-deploy/local-build/Makefile`. For rootfs builds, the target pattern name is `rootfs-(initrd|image)-[flavor]-tarball` where `flavor` can be anything. For example, for Confidential Containers we use `flavor` to build TEE-specifc rootfs, for example, AMD SEV (so `rootfs-initrd-sev-tarball` is the target) and others will come. In the past we used to build the guest OS for various OS (ubuntu, alpine, fedora...) however we hadn't covered on CI resulting on broken variants, then the community decide to keep on the ubuntu-based. That's why we don't have a `rootfs-initrd-fedora-tarball`, for example. However, IMHO, with cbl-mariner we would start building more OS variants as long as they are maintained and tested on CI. Okay, having said all that, I suggest to create a `rootfs-initrd-mariner-tarball` target so that a developer/CI can build it as:

```
$ make rootfs-initrd-mariner-tarball
```

Next you need to adapt the `tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh` script to accept the `--build=rootfs-initrd-mariner`. In particular, have a look at `install_initrd()` which the function that calls the rootfs builder script (tools/packaging/guest-image/build_image.sh), i.e., this own script should be changed to receive the `initrd_distro` and `initrd_distro_version`. It is the caller of `build_image.sh`  that should read the distro name and version from `versions.yaml`.

wainersm: FYI @fidencio

fidencio: I wonder if we also should make it more explicit the default ones (as in rootfs-initrd-alpine-tarball, and rootfs-image-ubuntu-tarball), or if this is a little bit too much.

fidencio:
The patchset is clean, easy to read, easy to review, thanks for that @sprt.
I agree with what Wainer brought up, and that's my only point as well.

sprt:
Note: switched to using an initrd as per Mariner's defaults.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -64,6 +64,15 @@ function install_artifacts() {
 	chmod +x /opt/kata/bin/*
 	[ -d /opt/kata/runtime-rs/bin ] && \
 		chmod +x /opt/kata/runtime-rs/bin/*
+
+	# Allow Mariner to use custom configuration.
+	if [ "${HOST_OS}" == "cbl-mariner" ]; then
+		config_path="/opt/kata/share/defaults/kata-containers/configuration-clh.toml"
+		clh_path="/opt/kata/bin/cloud-hypervisor-glibc"
+		sed -i -E 's|(enable_annotations) = .+|\1 = ["initrd", "kernel"]|' "${config_path}"
```
jepio, line 72: this removes the default `enable_iommu` annotation from the list

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -64,6 +64,15 @@ function install_artifacts() {
...
+		sed -i -E "s|(valid_hypervisor_paths) = .+|\1 = [\"${clh_path}\"]|" "${config_path}"
+		sed -i -E "s|(path) = \".+/cloud-hypervisor\"|\1 = \"${clh_path}\"|" "${config_path}"
```
jepio, line 74: if we're doing this here, then we could as well specify the mariner kernel and initrd here instead of annotations?

jepio (commented):
I know that you're trying to express "for mariner, run the same guest os as host os", but the resulting versions.yaml does not make much sense.

The naming is backwards: `HOST_OS` actually configures the guest OS. `KATA_HOST_OS` is leaking from the test scripts, and from the point of view of kata-deploy we should be giving users a chance to configure the `GUEST_OS`.

sprt:
> I know that you're trying to express "for mariner, run the same guest os as host os", but the resulting versions.yaml does not make much sense.
> 
> The naming is backwards: `HOST_OS` actually configures the guest OS. `KATA_HOST_OS` is leaking from the test scripts, and from the point of view of kata-deploy we should be giving users a chance to configure the `GUEST_OS`.

Hmm, I see versions.yaml as outputting an initrd/image guest OS based on the "inputs" you pass it. So in the same way `architecture` "takes" `x86_64, aarch64, ...`, now `host_os` takes `ubuntu, cbl-mariner`, which give you a guest OS. Does that make sense?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -182,12 +205,26 @@ install_initrd() {
...
+
+#Install Mariner guest initrd
+install_initrd_mariner() {
+	install_initrd "cbl-mariner"
 }
 
 #Install guest initrd for sev
 install_initrd_sev() {
-	install_initrd "initrd-sev" "sev"
+	install_initrd "${DEFAULT_HOST_OS}" "sev"
```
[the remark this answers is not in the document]
sprt: Thanks for catching this! Fixed.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -396,26 +433,44 @@ install_firecracker() {
...
-	sudo install -D --owner root --group root --mode 0744 cloud-hypervisor/cloud-hypervisor "${destdir}/opt/kata/bin/cloud-hypervisor"
+	sudo install -D --owner root --group root --mode 0744 cloud-hypervisor/cloud-hypervisor "${destdir}/opt/kata/bin/cloud-hypervisor${suffix}"
+}
+
+# Install static cloud-hypervisor asset
+install_clh() {
+	if [[ "${ARCH}" == "x86_64" ]]; then
+		features="mshv,tdx"
+	else
+		features="mshv"
```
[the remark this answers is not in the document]
sprt: Thanks for catching - fixed

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -396,26 +433,44 @@ install_firecracker() {
...
 		&& return 0
 
-	if [[ "${ARCH}" == "x86_64" ]]; then
-		export features="tdx"
-	fi
+	export libc
+	export features
 
 	info "build static cloud-hypervisor"
 	"${clh_builder}"
```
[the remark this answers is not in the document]
sprt: Done

tools/packaging/kata-deploy/scripts/kata-deploy.sh
sprt: Fixed

tools/packaging/kata-deploy/scripts/kata-deploy.sh
sprt: I'm only doing this here because these are impossible to set via annotations. I like `setup.sh` since it can also be cleanly reused for other environments.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -64,6 +64,15 @@ function install_artifacts() {
 	chmod +x /opt/kata/bin/*
 	[ -d /opt/kata/runtime-rs/bin ] && \
 		chmod +x /opt/kata/runtime-rs/bin/*
+
+	# Allow Mariner to use custom configuration.
+	if [ "${HOST_OS}" == "cbl-mariner" ]; then
```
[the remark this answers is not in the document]
sprt: Thanks! Fixed

tests/integration/kubernetes/setup.sh
```diff
@@ -9,12 +9,34 @@ set -o pipefail
 
 kubernetes_dir=$(dirname "$(readlink -f "$0")")
 
+repo_root_dir="$(cd "${kubernetes_dir}/../../../" && pwd)"
+source "${repo_root_dir}/tools/packaging/scripts/lib.sh"
+install_yq
```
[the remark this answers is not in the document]
sprt: Reverted to `install_yq` because of https://github.com/kata-containers/kata-containers/actions/runs/5260204827/jobs/9507373999

sprt:
Implemented @jepio's suggestions to avoid propagating `HOST_OS` through the build process and flatten `versions.yaml`.

versions.yaml
```diff
@@ -122,17 +122,20 @@ assets:
     url: "https://github.com/kata-containers/kata-containers/tools/osbuilder"
     architecture:
       aarch64:
-        name: "ubuntu"
-        version: "latest"
+        name:  &default-image-name "ubuntu"
```
jepio, line 125:
```suggestion
        name: &default-image-name "ubuntu"
```

sprt: Done

jepio (approved):
@fidencio @wainersm
i think this is in good shape now - would you take another look?

jepio:
The metrics gha job is fixed in main, i would not rebase again just for that.

versions.yaml
```diff
@@ -122,17 +122,20 @@ assets:
...
+        name: *default-image-name
+        version: *default-image-version
       x86_64:
-        name: &default-image-name "ubuntu"
-        version: "latest"
+        name: *default-image-name
+        version: *default-image-version
+        tdx:
+          name: *default-image-name
+          version: *default-image-version
```
fidencio, line 138: I'm not sure why we're adding TDX here.

fidencio:
In general it looks good to me, just would like to have an answer on one specific bit.

versions.yaml
sprt: This is a suggestion from @jepio, which makes sense if you follow the code path from [`install_image_tdx()`](https://github.com/kata-containers/kata-containers/pull/6839/commits/532755ce31d069dafaa1427c55757f0973381647#diff-4e9dffeafb07dc536a419fa1c12eda2ed1fba26d98f3987694a13fafecc4362dR181). This allows us to simply get the guest OS name/version [based on a "variant" field](https://github.com/kata-containers/kata-containers/pull/6839/commits/532755ce31d069dafaa1427c55757f0973381647#diff-4e9dffeafb07dc536a419fa1c12eda2ed1fba26d98f3987694a13fafecc4362dR169-R175) instead of having a bunch of [if statements in the image builder](https://github.com/kata-containers/kata-containers/pull/6839/commits/532755ce31d069dafaa1427c55757f0973381647#diff-c4e252149c7d3f7767fa98bf90ccef0e2cfe8a9a9065f08b1cb181df001ea0b8L113-L123).

fidencio: Makes sense.

fidencio (approved):
lgtm, thanks @sprt!

Issue #6553, opened 2023-03-30
gpu: Build and Ship an GPU enabled Kernel with every Kata release


zvonkok:
For easier integration of GPU-related workloads on Kata, it would be nice to ship a GPU-enabled kernel with every Kata release. This ensures that it is tested, and we can be sure that it works with the release it was shipped with. 
I am in the process of adding GPU tests to CI, which will make sure that the Kernel is working correctly. 

We already have several pipelines to build various kernels TDX, SNP, I would like to make sure that we're always using the correct set of artifacts to run "any" workload on Kata.

PR #6845, opened 2023-05-16
gpu: Rename the last bits from `gpu` to `nvidia-gpu`

fidencio:
Let's specifically name the `gpu` runtime class as `nvidia-gpu`.  By
doing this we keep the door open and ease the life of the next vendor
adding GPU support for Kata Containers.

Fixes: #6553


wainersm (approved):
Thanks @fidencio !

Renaming gpu to nvidia-gpu is a right decision IMO.

stevenhorsman (approved):
LGTM. Thanks!

Issue #6843, opened 2023-05-16
BindDevicetoVFIO cause abnormal exit of shim when config i40e to iavf


liangxianlong:
**Which feature do you think can be improved?**
Use a new method to bind device to vfio:
echo "vfio-pci" > /sys/bus/pci/devices/${bdf}/driver_overwride
echo ${bdf} > /sys/bus/pci/devices/${bdf}/driver/unbind
echo ${bdf} > /sys/bus/pci/drivers_probe

PR #6855, opened 2023-05-17
runtime: bind and unbind vfio with driver_probe

liangxianlong:
Make vfio driver switching more convenient and compatible.

Fixes: #6843

### DIFF
--- a/src/runtime/pkg/device/drivers/vfio.go
+++ b/src/runtime/pkg/device/drivers/vfio.go
@@ -22,14 +22,13 @@ import (
 
 // bind/unbind paths to aid in SRIOV VF bring-up/restore
 const (
-	pciDriverUnbindPath = "/sys/bus/pci/devices/%s/driver/unbind"
-	pciDriverBindPath   = "/sys/bus/pci/drivers/%s/bind"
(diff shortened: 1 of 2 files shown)


fidencio:
@zvonkok, mind to take a look at this one?

zvonkok:
Will do on Monday, public holiday and long weekend.

amshinde:
@liangxianlong Thanks for the PR. Sorry for the delay in review.  I am trying to understand the root cause for the bug you are seeing.
The current implementation was tested and would work for a physical device/virtual function that has a driver assigned to it on the host, I can see it may not work for a physical device that does not have a driver assigned on the host. 
Is this what is happening in your case?
With the change you have implemented, I want to make sure vfio driver assignment works in all cases:
1) Physical device/VF assigned to host driver
2) Physical device/VF not using a driver on the host. 

Have you tested your code in the above scenarios? I'll try to test this on my end as well.

amshinde:
@liangxianlong Since I havent heard from you, I have raised a PR that includes this change that I needed to add another network related feature. I have attributed the commit with you as a contributor:https://github.com/kata-containers/kata-containers/pull/8597/commits/17a8c573777ae04a2558eef9879347e4daae56a4

Issue #6857, opened 2023-05-17
gha: tdx: Cleanup must use the k3s overlay


fidencio:
As the TDX CI is running on a k3s deployed cluster, the cleanup should be using the k3s overlay (as already done for the kata-deploy scripts).

Issue #6760, opened 2023-04-28
gha: consolidate the CI workflows for TDX, SEV and SNP.


wainersm:
**Which feature do you think can be improved?**

The github CI workflows for TDX, SEV and SNP. 

**How can it be improved?**

They can be consolidated into a single yaml file as described in https://github.com/kata-containers/kata-containers/pull/6712#discussion_r1179516764 so that we remove duplicated code.

Cc @fidencio @ryansavino

fidencio:
https://github.com/kata-containers/kata-containers/pull/6712#discussion_r1182375126

PR #6858, opened 2023-05-17
gha: tdx: Use the k3s overlay for kata-cleanup

fidencio:
As the TDX CI runs on k3s, we must ensure the cleanup, as already done for the deploy, used the k3s overlay.

Fixes: #6857


gkurz (approved):
LGTM. Thanks @fidencio !

wainersm:
While ago I opened https://github.com/kata-containers/kata-containers/issues/6760 suggesting to consolidate the sev, snp and tdx into a single workflow. We will need to deal with the fact that SEV (at least) have k8s installed via kubeadm over baremetal, so those rancher specific bits will need to be refactored. Anyway, this change looks fine with the current workflow. Thanks @fidencio !

studychao (commented):
Thanks!

Issue #6842, opened 2023-05-16
kata-agent nic info sync raise "Received a netlink error message File exists (os error 17)" error


yiya1989:
# Description of problem
pod has two nic(eth1 and eth0),  we expect kata sync these two nic info into vm(eth0 eth1), but raise File exists (os error 17).
It works well of kata-2.3, but it failed of kata-2.5.

Network info in ns of physical machine:
eth1,  mac: BBB
eth0, mac: AAA

Network info of vm :
eth0,  mac: BBB
eth1, mac: AAA

so when kata agent sync info from host's network ns to vm's network,  it raise mac duplicated(`File exists (os error 17)`)

# Expected result
kata works well.
Network info in ns of physical machine:
eth1,  mac: BBB
eth0, mac: AAA

Network info of vm :
eth0,  mac: AAA
eth1, mac: BBB

# Actual result
kata agent raise error:
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim: "update interface: Received a netlink error message File exists (os error 17)": unknown

# Further information
I see a related commit:
[Make the network API non Linux specific](https://github.com/kata-containers/kata-containers/pull/3597)
https://github.com/kata-containers/kata-containers/commit/2edea883698bddcc2b17f24ac3c516382027d753

**The old code is collect all endpoint, and then sort endpoints by name,  and then attatch the endpoints.
But the new code is collect one the network info and then attatch this endppint, finally sort endpoints by name after all endpoints attatched, so lead to the difference of vm's nic order with ns's nic order**

The related code is as bellow:
```
func createEndpointsFromScan(networkNSPath string, config *NetworkConfig) ([]Endpoint, error) {
	var endpoints []Endpoint

	idx := 0
	for _, link := range linkList {
                 ....
		if err := doNetNS(networkNSPath, func(_ ns.NetNS) error {
			endpoint, errCreate = createEndpoint(netInfo, idx, config.InterworkingModel, link)
			return errCreate
		}); err != nil {
			return []Endpoint{}, err
		}

		endpoints = append(endpoints, endpoint)
	}

	sort.Slice(endpoints, func(i, j int) bool {
		return endpoints[i].Name() < endpoints[j].Name()
	})
        ...
}

// Add adds all needed interfaces inside the network namespace.
func (n *Network) Add(ctx context.Context, config *NetworkConfig, s *Sandbox, hotplug bool) ([]Endpoint, error) {
        ...
[...]

PR #6860, opened 2023-05-17
netlink: Fix the issue of update_interface

lifupan:
When updating an interface, there's maybe an existed interface whose name would be the same with the updated required name, thus it would update failed with interface name existed error. Thus we should rename the existed interface with an temporary name and swap it with the previouse interface name last.

Fixes: #6842


mxpv (approved):

Tim-Zhang (approved):
LGTM

PR #6863, opened 2023-05-17
kata-shim: Fix the issue of update_interface error

yiya1989:
We collect all the netInfo, and then collect all the endpoints. After all the endpoints collectted, we sort endpoints by name, and then attatch them to vmm.
After this action, we can make sure that endpoints sequence is same with the vm's nic sequence.

Fixes: #6842


yiya1989:
can any one approval these test? 3ks

yiya1989:
@Tim-Zhang @lifupan  This PR is another way to fix https://github.com/kata-containers/kata-containers/issues/6842 by sort the nics in kata-shim, which is fit for the kata's old endpoint code logic. Can you help me to review it?

Issue #6865, opened 2023-05-17
gha: tests: Merge "Deploy kata-deploy" and "Run tests" steps


fidencio:
Those are the steps we can see here:
https://github.com/kata-containers/kata-containers/blob/9630c13ac0cc6087c40cf5cb93afa71a4a7d8752/.github/workflows/run-k8s-tests-on-aks.yaml#L64-L83

The reason for merging those steps is because if something goes awry and we want to re-run the tests, we really would like to have kata-deploy properly redeployed.  However,. considering the case this step passed in the first run, it'll just go through it in the second run without actually running it. :-/

Yet another weirdness from GitHub actions.

Issue #6649, opened 2023-04-12
gha: ci-on-push: Errors to start the first test are randomly happening


fidencio:
```
# Events:
#   Type     Reason                  Age   From               Message
#   ----     ------                  ----  ----               -------
#   Warning  FailedScheduling        91s   default-scheduler  0/1 nodes are available: 1 node(s) didn't match Pod's node affinity/selector. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
#   Normal   Scheduled               80s   default-scheduler  Successfully assigned default/handlers to aks-nodepool1-41500360-vmss000000
#   Warning  FailedCreatePodSandBox  80s   kubelet            Failed to create pod sandbox: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix /run/containerd/containerd.sock: connect: no such file or directory"
#   Warning  FailedCreatePodSandBox  32s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: Failed to Check if grpc server is working: rpc error: code = DeadlineExceeded desc = timed out 
```

I've faced this both in the TDX CI, and also on the AKS Ci.

/cc @wainersm @stevenhorsman @bergwolf

stevenhorsman:
Just to confirm, is this for vanilla kata - so not with the custom fork of containerd?
Does AKS/ rancher use the default `CONTAINER_RUNTIME_ENDPOINT` value? anyway we can get more debug?

PR #6866, opened 2023-05-17
gha: k8s: Make the tests more reliable

fidencio:
We like it or not, every now and then we'll have to deal with flaky tests, and our tests using GHA are not exempt from that fact.

With this simple commit, we're trying to improve the reliability of the tests in a few different fronts:

* Giving enough time for the script used by kata-deploy to be executed
  * We've hit issues as the kata-deploy pod is considered "Ready" at the moment it starts running, not when it finishes the needed setup. We should also be looking on how to solve this on the kata-deploy side but, for now, let's ensure our tests do not break with the current kata-deploy behavior.

* Merging the "Deploy kata-deploy" and "Run tests" steps
  * We've hit issues re-running tests and seeing even more failures than the ones we're trying to debug, as a step will simply be taken as succeeded as part of the re-run, in case it was successful executed as part of the first run.  This causes issues with the kata-deploy deployment, as the tests would start running before even having the node set up for running Kata Containers.

Fixes: #6865 #6649

### DIFF
--- a/.github/workflows/run-k8s-tests-on-aks.yaml
+++ b/.github/workflows/run-k8s-tests-on-aks.yaml
@@ -61,7 +61,8 @@ jobs:
         run: |
           az aks get-credentials -g "kataCI" -n ${{ github.event.pull_request.number }}-${{ github.event.pull_request.head.sha }}-${{ matrix.vmm }}-amd64
 
-      - name: Deploy kata-deploy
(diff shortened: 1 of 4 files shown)


GabyCT (approved):

stevenhorsman (approved):
LGTM. Thanks!

Issue #6884, opened 2023-05-18
CCv0: Do not update TDX components, as the TDX jenkins CI is running on a 2022ww44 TDX stack


fidencio:
"What's a 2022ww44 TDX stack?", someone would ask, and the answer is that TDX development has been made publicly available, as a full stack that will work end-to-end, in interval of a few weeks between each drop.

Current as part of the CCv0, we're running a Ci that has a system based on the 2022ww44, but on `main` we updated the TDX artefacts to the `2023ww01`, and this was mistakenly brought to the `CCv0` branch (if we'd play the blame games, the reviewer didn't catch this, and the reviewer happens to be myself).

We should revert the patches related to this and ensure that the TDX CI will be up and running in order to avoid any possible issues during the freeze / release time.

stevenhorsman:
I believe this was resolved with #6886

PR #6886, opened 2023-05-18
CC: tdx: Stick to the 2022ww44 TDX stack for the CCv0 branch

fidencio:
As the Jenkins TDX CI is running on a system with a TDX stack called
"2022ww44", we should keep the QEMU / kernel / OVMF versions matching
what's provided in that stack.

The reason we were able to update this on `main` is because the GHA TDX
CI is running on a TDX stack called "2023ww01", but we have decided to
NOT take the bullet, NOT updating the Jenkins CI in order to avoid
unexepected breakages.

This regression was introduced as part of the last CCv0 merge to main,
and would've been caught by the CI, and should've been caught by the
reviewer (myself :-)), but CI was having a hard time to even build the
compoenents and I wrote in the PR and I'm quoting it here: "I rather
deal with possible breakages on this later on, than block this PR to get
in." ... and here we are. :-)

Fixes: #6884


GabyCT (approved):

fidencio:
TDX machine is having a really hard time with network.
I did a `make cc-tdx-kernel-tarball`, `make cc-tdx-qemu-tarball`, and `make cc-tdx-tdvf-tarball` locally, and I could get the artefacts properly built.

I'd propose to get this one merged, get the cache of those artefacts running, and that will help our situation a little bit.

stevenhorsman (approved):
Reverting these changes makes sense base don the description

Issue #6897, opened 2023-05-19
cached components: OVMF cache is not working


fidencio:
`make ovmf-tarball` will rebuild OVMF instead of using the cached one, in cases where the cache is valid.

fidencio:
Note: tdvf and ovmf-sev are working as expected.

fidencio:
I think I see what's going on.
We try to download it from a specific URL: `${jenkins_url}/job/kata-containers-main-ovmf-${ovmf_type}-$(uname -m)/${cached_artifacts_path}`

As `ovmf_type` is x86_64, we end up with a URL like `main-ovmf-x86_64-x86_64`.  I'd say it's not worth patching the code right now, but rather simply changing the cache job URL.

fidencio:
No, that was not the issue.  OVMF_FLAVOUR was wrong in http://jenkins.katacontainers.io/view/Cached%20jobs/job/kata-containers-main-ovmf-x86_64-x86_64/, it should've been "x86_64", but it was set to "ovmf".

I've changed it and restarted the job.

fidencio:
No, it actually requires a PR to fix the cache component part. :-/.
I'm assigning this one to myself.

PR #6898, opened 2023-05-19
cache: Fix OVMF caching

fidencio:
OVMF has been cached, but it's not been used from cache as the `version` set in the cached builds has always been empty.

The reason for that is because we've been trying to look for `externals.ovmf.ovmf.version`, while we should be actually looking for `externals.ovmf.x86_64.version`.

Setting `x86_64` as the OVMF_FLAVOUR would cause another bug, as the expected tarball name would then be `kata-static-x86_64.tar.xz`, instead of `kata-static-ovmf-x86_64.tar.xz`.

With everything said, let's simplify the OVMF_FLAVOUR usage, by using it as it's passed, and only adapting the tarball name for the TDVF case, which is the abnormal one.

Fixes: #6897

### DIFF
--- a/tools/packaging/static-build/cache_components_main.sh
+++ b/tools/packaging/static-build/cache_components_main.sh
@@ -77,9 +77,8 @@ cache_nydus_artifacts() {
 
 cache_ovmf_artifacts() {
 	local current_ovmf_version="$(get_from_kata_deps "externals.ovmf.${OVMF_FLAVOUR}.version")"
-	[ "${OVMF_FLAVOUR}" == "tdx" ] && OVMF_FLAVOUR="tdvf"
-	[ "${OVMF_FLAVOUR}" == "sev" ] && OVMF_FLAVOUR="ovmf-sev"
-	local ovmf_tarball_name="kata-static-${OVMF_FLAVOUR}.tar.xz"
+	local ovmf_tarball_name="kata-static-ovmf-${OVMF_FLAVOUR}.tar.xz"
+	[ "${OVMF_FLAVOUR}" == "tdx" ] && ovmf_tarball_name="kata-static-tdvf.tar.xz"
 	local current_ovmf_image="$(get_ovmf_image_name)"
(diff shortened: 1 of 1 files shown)


stevenhorsman:
> The reason for that is because we've been trying to look for externals.ovmf.ovmf.version, while we should be actually looking for externals.ovmf.x86_64.version.

Just to check I understand - It looks like OVMF_FLAVOUR was incorrectly set to `ovmf` in the cache job config rather than the source code as I can't see where it was set to `omvf` in there?

fidencio:
> Just to check I understand - It looks like OVMF_FLAVOUR was incorrectly set to `ovmf` in the cache job config rather than the source code as I can't see where it was set to `omvf` in there?

Yes, that's 100% correct.

At some point we should start versioning the cache scripts, in order to keep track of things like this one.

stevenhorsman (approved):
LGTM. Thanks for the clarification

tools/packaging/static-build/cache_components_main.sh
```diff
@@ -77,9 +77,8 @@ cache_nydus_artifacts() {
 
 cache_ovmf_artifacts() {
 	local current_ovmf_version="$(get_from_kata_deps "externals.ovmf.${OVMF_FLAVOUR}.version")"
-	[ "${OVMF_FLAVOUR}" == "tdx" ] && OVMF_FLAVOUR="tdvf"
-	[ "${OVMF_FLAVOUR}" == "sev" ] && OVMF_FLAVOUR="ovmf-sev"
-	local ovmf_tarball_name="kata-static-${OVMF_FLAVOUR}.tar.xz"
+	local ovmf_tarball_name="kata-static-ovmf-${OVMF_FLAVOUR}.tar.xz"
+	[ "${OVMF_FLAVOUR}" == "tdx" ] && ovmf_tarball_name="kata-static-tdvf.tar.xz"
```
wainersm, line 81: Just curious: is it conceptually wrong call it "ovmf-tdvf"?

wainersm (approved):

stevenhorsman:
All the tests have passed, so I think this can be merged.

Issue #6899, opened 2023-05-19
gha: payload after push needs to log into quay.io, but is currently not doing it


fidencio:
The payload after push action is broken as it requires to push the container used to build the payload to quay.io.  However, during the refactor we did in the GHA to reuse files, this part ended up missed.

PR #6900, opened 2023-05-19
gha: payload-after-push: Pass secrets down

fidencio:
The "build-assets-${arch}" jobs need to have access to the secrets in order to log into the container registry in the cases where "push-to-registry", which is used to push the builder containers to quay.io, is set to "yes".

Now that "build-assets-${arch}" pass the secrets down, we need to log into the container registry in the "build-kata-static-tarball-${arch}" files, in case "push-to-registry" is set to "yes".

Fixes: #6899


stevenhorsman (approved):
LGTM. Thanks!

wainersm (approved):
Thanks @fidencio !

Issue #6901, opened 2023-05-19
CC | local-build: Bring back root_hash logic


fidencio:
I cannot easily pin-point which commit dropped it, but my gut feeling is
that it's the result of an erroneous conflict resolution when merging
content from main to the CCv0 branch.

Regardless of when / why it happened, as the root_hash logic ended up
being dropped, workflows that depend on that are now failing.

PR #6902, opened 2023-05-19
CC | local-build: Bring back root_hash logic

fidencio:
I cannot easily pin-point which commit dropped it, but my gut feeling is that it's the result of an erroneous conflict resolution when merging content from main to the CCv0 branch.

Regardless of when / why it happened, as the root_hash logic ended up being dropped, workflows that depend on that are now failing.

With everything said in mind, let's bring the logic back.

Fixes: #6901


stevenhorsman (approved):
Yeah, this got dropped in the 7 week old merge PR https://github.com/kata-containers/kata-containers/pull/6559/files#diff-4e9dffeafb07dc536a419fa1c12eda2ed1fba26d98f3987694a13fafecc4362d

Thanks for spotting and adding back!

wainersm:
@fidencio it is a pieced I never touched. It seems this logic is meant to overwrite the one in `install_cached_cc_shim_v2()`, is it correct? Anyway, if the logic existed before and got wiped out by the merge, let's return to the previous state.

wainersm (approved):

stevenhorsman:
All the tests are passing apart from the tdx ones, which have a known error, so I think it's safe to merge this

Issue #6908, opened 2023-05-19
gha | deploy: kata-deploy-build-and-upload-payload can silently fail 


stevenhorsman:
I think this is valid on `main`, but have only seen it on CCv0 as the main payload deploy isn't running at the moment.
The complaint might be valid for other steps, but in specifically `build-and-push-kata-payload` the `kata-deploy-build-and-upload-payload.sh` script can fail, but the step is still marked as successful and moves on to the next step which then fails
e.g. https://github.com/kata-containers/kata-containers/actions/runs/5026720857/jobs/9017815564 (this example is s390x, but it failed on x86 as well) the build of the image failed (I think due to Ubuntu being so flakey), but the logs show:
```
Err:1 https://packages.cloud.google.com/apt kubernetes-xenial InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B53DC80D13EDEF05
Reading package lists...
W: GPG error: https://packages.cloud.google.com/apt kubernetes-xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B53DC80D13EDEF05
E: The repository 'https://apt.kubernetes.io/ kubernetes-xenial InRelease' is not signed.
The command '/bin/bash -o pipefail -c apt-get update && apt-get install -y --no-install-recommends apt-transport-https ca-certificates curl xz-utils systemd && mkdir -p /etc/apt/keyrings/ && curl -fsSLo /etc/apt/keyrings/kubernetes-archive-keyring.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg && echo "deb [signed-by=/etc/apt/keyrings/kubernetes-archive-keyring.gpg] https://apt.kubernetes.io/ kubernetes-xenial main" | tee /etc/apt/sources.list.d/kubernetes.list && apt-get update && apt-get install -y --no-install-recommends kubectl && apt-get clean && rm -rf /var/lib/apt/lists/ && mkdir -p ${DESTINATION} && tar xvf ${KATA_ARTIFACTS} -C ${DESTINATION} && rm ${WORKDIR}/${KATA_ARTIFACTS}' returned a non-zero code: 100
Pushing the image to the registry
The push refers to repository [quay.io/confidential-containers/runtime-payload-ci]
tag does not exist: quay.io/confidential-containers/runtime-payload-ci:kata-containers-2ba029f3e1f3d7ac60d192582cbce2954ce3639e-s390x
```
[...]

PR #6913, opened 2023-05-22
local-build: Standardise what's set for the local build scripts

fidencio:
We've a discrepancy on what's set along the scripts used to build the Kata Cotainers artefacts locally.

Some of those were missing a way to easily debug them in case of a failure happens, but one specific one (build-and-upload-payload.sh) could actually silently fail.

All of those have been changed as part of this commut.

Fixes: #6908

### DIFF
--- a/tools/packaging/kata-deploy/Dockerfile
+++ b/tools/packaging/kata-deploy/Dockerfile
@@ -18,7 +18,7 @@ RUN \
 apt-get update && \
 apt-get install -y --no-install-recommends apt-transport-https ca-certificates curl xz-utils systemd && \
 mkdir -p /etc/apt/keyrings/ && \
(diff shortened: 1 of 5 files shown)


fidencio:
Adding the needs-backport here as it needs to be backported to the CCv0 branch.

snir911 (approved):
LGTM, thanks!

fidencio:
Nice that the first run of the tests already exposed the issue: https://github.com/kata-containers/kata-containers/actions/runs/5042650167/jobs/9043971122?pr=6913

I've updated this series to follow the suggestion provided by dims, here: https://github.com/kubernetes/k8s.io/pull/4837#issuecomment-1446426585

fidencio:
I've force-pushed here in order to make the commit message check happy.

jepio (approved):
LGTM, I'm going to make the same change for the jenkins jobs.

PR #6943, opened 2023-05-23
gha: release: login-action: Don't specify docker.io registry

fidencio:
For some bizarre reason, the login-action will simply fail to authenticate to docker.io in it's specified as a registry.  The way to proceed, instead, is to *not* specify any registry as it'd be used by default.

Fixes: #6943


fengwang666 (approved):
LGTM

cmaf (approved):

GabyCT (approved):

amshinde (approved):

Issue #6946, opened 2023-05-24
gha: release: Fix tagging the release payload


fidencio:
We previously were doing:
* Create a new image on kata-deploy-ci using the commit hash of the latest tag
  * This was used to test on AKS, which is no longer needed as we test on AKS on every PR
* Create a new image on kata-deploy using the release tag and "latest" or "stable", by tagging the kata-deploy-ci image accordingly

As part of cfe63527c5328611cac647d0c10916bef6cf0408, we broke the workflow described above, as in the first step we would save the PKG_SHA to be used in the second step, but that part ended up being removed.

Anyways, this back and forth is not needed anymore and we can simplify
the process by doing:
* Create a new image on kata-deploy, using:
  - The tag received as ref from the event that triggered this worklow
  - "latest" or "stable" tag, depending on whether it's a stable release or not

PR #6947, opened 2023-05-24
gha: release: Simplify the process for tagging the payload

fidencio:
We previously were doing:
* Create a new image on kata-deploy-ci using the commit hash of the latest tag
  * This was used to test on AKS, which is no longer needed as we test on AKS on every PR
* Create a new image on kata-deploy using the release tag and "latest" or "stable", by tagging the kata-deploy-ci image accordingly

As part of cfe63527c5328611cac647d0c10916bef6cf0408, we broke the workflow described above, as in the first step we would save the PKG_SHA to be used in the second step, but that part ended up being removed.

Anyways, this back and forth is not needed anymore and we can simplify the process by doing:
* Create a new image on kata-deploy, using:
  - The tag received as ref from the event that triggered this worklow
  - "latest" or "stable" tag, depending on whether it's a stable release or not

Fixes: #6946

### DIFF
--- a/.github/workflows/release-amd64.yaml
+++ b/.github/workflows/release-amd64.yaml
@@ -36,27 +36,16 @@ jobs:
       - name: build-and-push-kata-deploy-ci-amd64
         id: build-and-push-kata-deploy-ci-amd64
         run: |
-          tag=$(echo $GITHUB_REF | cut -d/ -f3-)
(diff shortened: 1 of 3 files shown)


jepio (approved):

stevenhorsman (approved):
LGTM. Thanks!

PR #6953, opened 2023-05-24
gha: Add nightly jobs

fidencio:
The idea of this PR is to add nightly jobs for the bits we care, and ensure we have something similar to the "Green CI" effort, but as part of GitHub.

This is a draft PR, the idea is to gather feedback from the interested parts.

One thing to mention, how do we track it?  My idea was to add a badge on our `kata-containers/kata-containers/README.md`, and from there folks can monitor the state of the CI.

### DIFF
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -1,0 +1,20 @@
+name: Kata Containers Nightly CI
+on:
+  schedule:
+    cron: '0 0 * * *'
+
+env:
+  COMMIT_HASH: ${GITHUB_REF}
+
+jobs:
+  set-fake-pr-number:
+    runs-on: ubuntu-latest
+    run: |
+      echo "PR_NUMBER=$(date +%Y%m%d%H%M%S)" >> "$GITHUB_ENV"
+
+  kata-containers-ci-on-push:
+    uses: ./.github/workflows/ci.yaml
+    with:
+      commit-hash: ${{ env.COMMIT_HASH }}
+      pr-number: ${{ env.PR_NUMBER }}
+      tag: ${{ env.PR_NUMBER }}-${{ env.COMMIT_HASH }}-nightly
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,6 @@ jobs:
     secrets: inherit
 
   run-k8s-tests-on-aks:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
     needs: publish-kata-deploy-payload-amd64
     uses: ./.github/workflows/run-k8s-tests-on-aks.yaml
     with:
@@ -42,7 +41,6 @@ jobs:
     secrets: inherit
 
   run-k8s-tests-on-sev:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
     needs: publish-kata-deploy-payload-amd64
     uses: ./.github/workflows/run-k8s-tests-on-sev.yaml
     with:
@@ -52,7 +50,6 @@ jobs:
       commit-hash: ${{ inputs.commit-hash }}
 
   run-k8s-tests-on-snp:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
     needs: publish-kata-deploy-payload-amd64
     uses: ./.github/workflows/run-k8s-tests-on-snp.yaml
     with:
@@ -63,7 +60,6 @@ jobs:
       commit-hash: ${{ inputs.commit-hash }}
 
   run-k8s-tests-on-tdx:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
     needs: publish-kata-deploy-payload-amd64
     uses: ./.github/workflows/run-k8s-tests-on-tdx.yaml
     with:
@@ -73,7 +69,6 @@ jobs:
       commit-hash: ${{ inputs.commit-hash }}
 
   run-metrics-tests:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
     needs: build-kata-static-tarball-amd64
     uses: ./.github/workflows/run-metrics.yaml
     with:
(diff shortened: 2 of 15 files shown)


stevenhorsman:
The code looks good to me at initial review and I think it make sense to have the nightly runs to get the daily baseline checks. One thing to think about for the bare-metal nodes - is there a clash between the timing of these runs and the Jenkins based ones that we need to think about (before Jenkins goes away completely) to avoid queueing?

fidencio:
> One thing to think about for the bare-metal nodes - is there a clash between the timing of these runs and the Jenkins based ones that we need to think about (before Jenkins goes away completely) to avoid queueing?

That's an excellent point.  And, to be fair, it makes me think that we should be really careful on how we use the same node for Jenkins and GHA, as that may lead to unexpected results.

I know that for the TDX node, we're only using that node for the GHA.  @UnmeshDeodhar, what's the situation with the SEV / SNP nodes?

We'll also have to properly think about this when we migrate the metrics job.  The VFIO one could be done on a VM (right, @jepio? :-)).

jepio:
> We'll also have to properly think about this when we migrate the metrics job. The VFIO one could be done on a VM (right, @jepio? :-)).

Right, the VFIO can be done (double) nested in an Azure VM, we just need those dynamic self-hosted runners (garm) for that. It should be even less problematic when we don't need to build inside the nested vm, because we just fetch the artifacts built earlier in the pipeline.

stevenhorsman:
Note something that should have a nightly test, but something that I think we'd want to consider a status badge for is the Runtime payload build: https://github.com/kata-containers/kata-containers/actions/workflows/payload-after-push.yaml as we have had 2+ month periods when it has been broken and as it is an automatic workflow that happens after things have been merge I suspect most people don't think about it too much (or maybe just me!)

fidencio:
> Note something that should have a nightly test, but something that I think we'd want to consider a status badge for is the Runtime payload build: https://github.com/kata-containers/kata-containers/actions/workflows/payload-after-push.yaml as we have had 2+ month periods when it has been broken and as it is an automatic workflow that happens after things have been merge I suspect most people don't think about it too much (or maybe just me!)

I've opened a PR for this one, as it's simpler: https://github.com/kata-containers/kata-containers/pull/7009

fidencio:
> I know that for the TDX node, we're only using that node for the GHA. @UnmeshDeodhar, what's the situation with the SEV / SNP nodes?

@UnmeshDeodhar, @ryansavino :-)

UnmeshDeodhar:
@fidencio AMD Rome and Milan nodes are being exclusively used for kata ci (github runner)

.github/workflows/ci-nightly.yaml
```diff
@@ -0,0 +1,20 @@
...
+    cron: '0 0 * * *'
+
+env:
+  COMMIT_HASH: ${GITHUB_REF}
+
+jobs:
+  set-fake-pr-number:
+    runs-on: ubuntu-latest
+    run: |
+      echo ::set-env name=PR_NUMBER::$(date +%Y%m%d%H%M%S)
```
stevenhorsman, line 13: IIRC - I think that I saw `set-env` being deprecated and replaced with using the environment variable file? https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+      repo: ${{ github.repository_owner }}/kata-deploy-ci
+      tag: ${{ inputs.tag }}-amd64
+      commit-hash: ${{ inputs.commit-hash }}
+
+  run-metrics-tests:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
+    needs: build-kata-static-tarball-amd64
+    uses: ./.github/workflows/run-metrics.yaml
+    with:
+      tarball-suffix: -${{ inpits.tag }}-amd64
```
stevenhorsman, line 80: Typo on `inputs`

stevenhorsman (commented):
At first pass I just picked up a couple of issues

.github/workflows/ci-nightly.yaml
fidencio: Fixed, thanks for the pointer!

.github/workflows/ci.yaml
fidencio: FIxed, thanks for catching my tpyos. :-)

fidencio:
Add the `Fixes: #`  to make our static checks happy.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,81 @@
...
+    with:
+      tarball-suffix: -${{ inputs.tag }}
+      registry: ghcr.io
+      repo: ${{ github.repository_owner }}/kata-deploy-ci
+      tag: ${{ inputs.tag }}-amd64
+      commit-hash: ${{ inputs.commit-hash }}
+    secrets: inherit
+
+  run-k8s-tests-on-aks:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
```
stevenhorsman, line 34: Is there any issues with the `ok-to-test` checks using the pull request event given that I think the nightly CI runs this path?

stevenhorsman (approved):
LGTM. Are you planning a follow up to add the badges for the nightly run to the README once it is working?

fidencio:
> LGTM. Are you planning a follow up to add the badges for the nightly run to the README once it is working?

Yep, that's my plan, but I'd like to have them stable before doing so.

.github/workflows/ci.yaml
```diff
@@ -31,7 +31,6 @@ jobs:
     secrets: inherit
 
   run-k8s-tests-on-aks:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
```
wainersm: @fidencio are we going to leave without the 'ok-to-test' or you plan to move to somewhere else?

fidencio: Those were moved here: https://github.com/kata-containers/kata-containers/blob/e067d183336ce0bb4f79d922787d44bc1e6144f4/.github/workflows/ci-on-push.yaml#L21

So, in order to start the CI you still have to set `ok-to-test`

wainersm: oh, my old eyes didn't see it. thanks!

wainersm (approved):
Thanks @fidencio ! As the talked the other day I will be able to monitor and report those jobs alongside the Jenkins ones at the CoCo meetings.

Issue #6957, opened 2023-05-24
kata-deploy: Improve shim backup / restore


stevenhorsman:
Some changes were made to the cc version of kata-delpoy to improve the shim backup and restore: https://github.com/kata-containers/kata-containers/pull/5045/commits/fb711e0e8e9bca07e852b49ffd84540b08ac9da1

These are worth bring forward into the `main` version as well

PR #6958, opened 2023-05-24
kata-deploy: Improve shim backup / restore

fidencio:
We're currently backing up and restoring all the possible shim files, but the default one ("containerd-shim-kata-v2").

Let's ensure this is also backed up and restored.

Fixes: #6957


stevenhorsman (approved):
LGTM. Thanks

jepio (approved):

Issue #6964, opened 2023-05-25
gha: Timeout in AKS tests


stevenhorsman:
During testing of https://github.com/kata-containers/kata-containers/pull/6952 - just a doc change I hit timeout issues with the AKS clh and qemu tests
https://github.com/kata-containers/kata-containers/actions/runs/5068536767/jobs/9101418128?pr=6952, so I think we need to look at this, or make the tests non-required

PR #6965, opened 2023-05-25
gha: Increaste timeout for AKS jobs and give more time to start running the tests

fidencio:
---

gha: aks: Increase the timeout time

We've seen tests being aborted close to the end of the run due to the
timeout.  Let's increase it, avoiding to hit such cases again..

Fixes: #6964

---

gha: aks: Wait longer to start running the tests

We're still facing issues related to the time taken to deploy the
kata-deplot daemonset and starting to run the tests.

Ideally, we should solve this with a readiness probe, and that's the
approach we want to take in the future.  However, for now, let's just
make sure those tests are not on the way of the community.

---


stevenhorsman (approved):
🤞 - this is enough of a timeout this time!

PR #6959, opened 2023-05-24
runtime: sending SIGKILL to qemu

beraldoleal:
There is a race condition when virtiofsd is killed without finishing all the clients. Because of that, when a pod is stopped, QEMU detects virtiofsd is gone, which is legitimate.

Sending a SIGTERM first before killing could introduce some latency during the shutdown.

One alternative is to just kill QEMU during this phase.

Fixes #6757.

PR #6967, opened 2023-05-25
runtime: sending SIGKILL to qemu (backport to 3.1)

beraldoleal:
There is a race condition when virtiofsd is killed without finishing all the clients. Because of that, when a pod is stopped, QEMU detects virtiofsd is gone, which is legitimate.

Sending a SIGTERM first before killing could introduce some latency during the shutdown.

Backport of #6959.

### DIFF
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -1108,22 +1108,21 @@ func (q *qemu) StopVM(ctx context.Context, waitOnly bool) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


gkurz (commented):
It is generally appreciated to have the `cherry picked from commit XXXX` line in backport commits :smile:

beraldoleal:
> It is generally appreciated to have the `cherry picked from commit XXXX` line in backport commits smile

Fixed with the force-push.

gkurz (approved):
LGTM. Thanks @beraldoleal !

Issue #6960, opened 2023-05-24
gha: Intermittent failures to create a pod


sprt:
Per #6772, we're running into errors creating pods from time to time (seemingly mostly on AKS but not only):

QEMU/CLH:
```
  Warning  FailedCreatePodSandBox  57s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: Failed to Check if grpc server is working: rpc error: code = DeadlineExceeded desc = timed out connecting to vsock 2179187993:1024: unknown
```

Dragonball:
```
  Warning  FailedCreatePodSandBox  11s (x7 over 116s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: Others("failed to handler message try init runtime instance\n\nCaused by:\n    0: init runtime handler\n    1: start sandbox\n    2: connect\n    3: connect agent server\n    4: connect\n    5: cannot connect to agent ttrpc server ConnectConfig { dial_timeout_ms: 10, reconnect_timeout_ms: 3000 }"): unknown
```

PR #6968, opened 2023-05-25
gha: Ensure kata-deploy completed before moving on

sprt:
The previous approach from e8f81ee proved unreliable, and in my own testing it could somehow kill the cluster. So instead this uses an init container to control the Ready state of the pod.

Fixes: #6960

### DIFF
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -78,7 +78,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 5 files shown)


fidencio (commented):
Please, also update https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml

jepio (approved):
Interesting approach, does this need to be changed in the CoCo operator too?

LGTM (with Fabiano's comment addressed).

Issue #6971, opened 2023-05-26
gha: Make it easier to test workflow changes


sprt:
With GHA, we can't test workflow YAML changes in the same PR where we make those changes. One way to address this would be to make every `run` step its own Bash script.

PR #6972, opened 2023-05-26
gha: aks: Extract `run` commands to a script

sprt:
Github Actions reads and runs workflow files from the main branch, rather than from the PR branch. This means that PRs that modify workflow files aren't being tested with the updated workflows coming from the PR, but rather with the old workflows from the main branch. AFAIK, this behavior isn't avoidable for workflow files (but is for other scripts).

This makes it very hard to reliably test workflow changes before they're actually merged into main and leads to issues that we have to hotifx (see #6983, #6995).

This PR aims to mitigate that by extracting the commands used in workflows to a separate script file. The way our CI is set up, those script files are read from the PR branch and thus changes would be reflected in the CI checks.

Fixes: #6971

### DIFF
--- a/.github/workflows/run-k8s-tests-on-aks.yaml
+++ b/.github/workflows/run-k8s-tests-on-aks.yaml
@@ -27,83 +27,44 @@ jobs:
           - host_os: cbl-mariner
             vmm: clh
     runs-on: ubuntu-latest
+    env:
+      DOCKER_REGISTRY: ${{ inputs.registry }}
+      DOCKER_REPO: ${{ inputs.repo }}
+      DOCKER_TAG: ${{ inputs.tag }}
+      GH_PR_NUMBER: ${{ github.event.pull_request.number }}
+      KATA_HOST_OS: ${{ matrix.host_os }}
+      KATA_HYPERVISOR: ${{ matrix.vmm }}
     steps:
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Download Azure CLI
-        run: |
-          curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
-          # The aks-preview extension is required while the Mariner Kata host is in preview.
-          az extension add --name aks-preview
+        run: bash tests/integration/gha-run.sh install-az-cli
 
       - name: Log into the Azure account
-        run: |
-          az login \
-            --service-principal \
-            -u "${{ secrets.AZ_APPID }}" \
-            -p "${{ secrets.AZ_PASSWORD }}" \
-            --tenant "${{ secrets.AZ_TENANT_ID }}"
-
-      - name: Format cluster name
-        run: |
-          rev=$(git rev-parse --short=12 HEAD)
-          echo "cluster_name=${{ github.event.pull_request.number }}-$rev-${{ matrix.vmm }}-${{ matrix.host_os }}-amd64" >> $GITHUB_ENV
+        run: bash tests/integration/gha-run.sh login-azure
+        env:
+          AZ_APPID: ${{ secrets.AZ_APPID }}
+          AZ_PASSWORD: ${{ secrets.AZ_PASSWORD }}
+          AZ_TENANT_ID: ${{ secrets.AZ_TENANT_ID }}
 
       - name: Create AKS cluster
-        run: |
-          az aks create \
-            -g "kataCI" \
-            -n "${{ env.cluster_name }}" \
-            -s "Standard_D4s_v5" \
-            --node-count 1 \
-            --generate-ssh-keys \
-            $([ "${{ matrix.host_os == 'cbl-mariner' }}" = "true" ] && echo "--os-sku mariner --workload-runtime KataMshvVmIsolation")
+        run: bash tests/integration/gha-run.sh create-cluster
 
       - name: Install `bats`
-        run: |
-          sudo apt-get update
-          sudo apt-get -y install bats
+        run: bash tests/integration/gha-run.sh install-bats
 
       - name: Install `kubectl`
-        run: |
-          sudo az aks install-cli
+        run: bash tests/integration/gha-run.sh install-kubectl
 
       - name: Download credentials for the Kubernetes CLI to use them
-        run: |
-          az aks get-credentials -g "kataCI" -n "${{ env.cluster_name }}"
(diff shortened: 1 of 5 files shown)


Redent0r (approved):
LGTM. Solves issue where you cant test actions changes until merged, and is cleaner too

fidencio:
@sprt, this lgtm as the first step, but I'd love to see the tee specific tests also taking advantage of this.

sprt:
> @sprt, this lgtm as the first step, but I'd love to see the tee specific tests also taking advantage of this.

Done, PTAL

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -27,78 +27,44 @@ jobs:
           - host_os: cbl-mariner
             vmm: clh
     runs-on: ubuntu-latest
+    env:
+      AZ_APPID: ${{ secrets.AZ_APPID }}
+      AZ_PASSWORD: ${{ secrets.AZ_PASSWORD }}
+      AZ_TENANT_ID: ${{ secrets.AZ_TENANT_ID }}
```
fidencio, line 33: I'm afraid those ones here would become plain text and easy to extract by having them as an env.
Wouldn't it be the case?  If not, why?

sprt: This is safe per [[1]](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) and [[2]](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow), but I did restrict the scope to only the login step to even further reduce the attack surface.

tests/integration/gha-run.sh
```diff
@@ -0,0 +1,153 @@
...
+        run_tests "sev"
+        ;;
+    run-tests-snp)
+        run_tests "snp"
+        ;;
+    run-tests-tdx)
+        run_tests "tdx"
+        ;;
+    cleanup-aks)
+        cleanup "aks"
```
jepio, line 134: make this its own function as it's different than the others

tests/integration/gha-run.sh
```diff
@@ -0,0 +1,153 @@
...
+    popd
+}
+
+function cleanup() {
+    platform="${1}"
+
+    if [ "${platform}" = "aks" ]; then
+        az aks delete \
+            -g "kataCI" \
+            -n "${GH_PR_NUMBER}-${GH_PR_SHA}-${KATA_HYPERVISOR}-amd64" \
```
jepio, line 56:
```suggestion
            -n "$(_print_cluster_name)" \
```

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -27,83 +27,45 @@ jobs:
...
       - name: Download Azure CLI
-        run: |
-          curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
-          # The aks-preview extension is required while the Mariner Kata host is in preview.
-          az extension add --name aks-preview
+        run: bash tests/integration/gha-run.sh install-az-cli
+        env:
+          AZ_APPID: ${{ secrets.AZ_APPID }}
+          AZ_PASSWORD: ${{ secrets.AZ_PASSWORD }}
+          AZ_TENANT_ID: ${{ secrets.AZ_TENANT_ID }}
```
jepio, line 48: These are scoped to the wrong step (the next one does the login), GHA has this weird "feature" that env assignments leak to following steps 😃

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -27,83 +27,45 @@ jobs:
...
 
       - name: Log into the Azure account
```
jepio, line 50: btw: any reason we're not relying on https://github.com/marketplace/actions/azure-login? @fidencio

tests/integration/gha-run.sh
```diff
@@ -0,0 +1,153 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+integration_dir="$(dirname "$(readlink -f "$0")")"
+tools_dir="${integration_dir}/../../tools"
+
+function _print_cluster_name() {
+    short_sha="$(echo "${GH_PR_SHA}" | cut -c -12)"
```
jepio, line 15: Fabiano recommended that we use `git rev-parse --short=12 HEAD` instead of string manipulation.

tests/integration/gha-run.sh
```diff
@@ -0,0 +1,153 @@
...
+
+    kubectl delete ${cleanup_spec}
+    kubectl delete -f "${tools_dir}/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml"
+    kubectl delete -f "${tools_dir}/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml"
+}
+
+function main() {
+    action="${1:-}"
+
+    case "${action}" in
```
jepio, line 87: what do you think of putting every case into a function?

.github/workflows/run-k8s-tests-on-aks.yaml
sprt: Oops, fixed!

> GHA has this weird "feature" that env assignments leak to following steps 😃

Hmm have you seen this in the wild? Can't find any docs on this.

.github/workflows/run-k8s-tests-on-aks.yaml
sprt: AFAIK Fabiano could never get that to work.

tests/integration/gha-run.sh
sprt: Done

tests/integration/gha-run.sh
sprt: Done

tests/integration/gha-run.sh
sprt: Done

fidencio (commented):
It looks good to me, and I'm super happy to approve it as soon as @jepio confirms all his comments were addressed.

.github/workflows/run-k8s-tests-on-aks.yaml
jepio: there are no docs on this, but i'm fairly sure i've seen this happen.

jepio (approved):

Issue #6978, opened 2023-05-29
Support file sharing for hypervisors that do not support virtiofs in runtime-rs


ananos:
In hypervisors that do not support virtiofs (eg Firecracker), sharing files between the host and the sandbox is only possible by copying files to the sandbox. 

The equivalent logic in the go runtime is in https://github.com/kata-containers/kata-containers/blob/099b4b0d0e3db31b9054e7240715f0d7f51f9a1c/src/runtime/virtcontainers/fs_share_linux.go#L227

This issue tracks the runtime-rs implementation for this feature.

PR #6980, opened 2023-05-29
runtime-rs: handle copy files when share_fs is not available

ananos:
In hypervisors that do not support virtiofs we have to copy files in the VM sandbox to properly setup the network (resolv.conf, hosts, and hostname).

To do that, we construct the volume as before, with the addition of an extra variable that designates the path where the file will reside in the sandbox.

In this case, we issue a `copy_file` agent request *and* we patch the spec to account for this change.

Fixes: #6978

### DIFF
--- a/src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
+++ b/src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
@@ -5,11 +5,15 @@
 //
 
 use std::{
+    fs::File,
+    io::Read,
+    os::unix::fs::MetadataExt,
     path::{Path, PathBuf},
     str::FromStr,
     sync::Arc,
 };
 
+use agent::Agent;
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use hypervisor::device::device_manager::DeviceManager;
@@ -18,6 +22,9 @@ use tokio::sync::RwLock;
 use super::Volume;
 use crate::share_fs::{MountedInfo, ShareFs, ShareFsVolumeConfig};
 use kata_types::mount;
+
+use crate::share_fs::DEFAULT_KATA_GUEST_SANDBOX_DIR;
+use crate::share_fs::PASSTHROUGH_FS_DIR;
 
 const SYS_MOUNT_PREFIX: [&str; 2] = ["/proc", "/sys"];
 
@@ -39,6 +46,7 @@ impl ShareFsVolume {
         m: &oci::Mount,
         cid: &str,
         readonly: bool,
+        agent: Arc<dyn Agent>,
     ) -> Result<Self> {
         // The file_name is in the format of "sandbox-{uuid}-{file_name}"
         let file_name = Path::new(&m.source).file_name().unwrap().to_str().unwrap();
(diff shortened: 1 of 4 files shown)


src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
```diff
@@ -60,10 +63,21 @@ impl ShareFsVolume {
                     }
                     Ok(src) => src,
                 };
+                let dest = [KATA_SHARED_SANDBOX_PATH, &file_name].join("/");
 
+                // This is where we set the value for the guest path
+                volume.guest_path = dest;
                 if src.is_file() {
-                    // TODO: copy file
```
Tim-0731-Hzt, line 65: I think it's better to implement copy file here. 
1. if `share_fs` match None, it means the hypervisor doesn't support fs_share, so we don't need to check  `is_fs_sharing_supported` again in `create`. 
3. we could simplify the `create` function, and the code will look more readable

src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
```diff
@@ -231,6 +245,17 @@ impl Volume for ShareFsVolume {
     fn get_device_id(&self) -> Result<Option<String>> {
         Ok(None)
     }
+
+    fn get_guest_path(&self) -> Result<Option<String>> {
```
Tim-0731-Hzt, line 249: I think you don't need this trait, if copy file is implemented in this file, the `guest path` will be stored in mount, `get_volume_mount` will be enough for use.

Tim-0731-Hzt (commented):
Thanks @ananos few comments

fidencio:
Sorry, at the time you opened your PR we were having issues with our CI, which leads to required tests failing without having nothing to do with this PR.

AS this has been already fixed on  main, we´d like to ask you to rebase your PR atop of the current `main` branch.

Thanks for the consideration, and sorry for the trouble.

ananos:
thanks @Tim-0731-Hzt for the comments! really appreciate your prompt response.

We weren't sure how to handle the copy request to the agent in share_fs_volume, that's why we did this trampoline thing with manager/manager_inner and container.rs. It seems that passing the agent as a parameter to the constructor of the share_fs_volume resolves this.

Let us know if this latest commit addresses your concerns. Thanks again!

src/runtime-rs/crates/resource/src/volume/share_fs_volume.rs
```diff
@@ -60,10 +64,51 @@ impl ShareFsVolume {
...
+                    &src,
+                    dest.clone()
+                );
+                agent
+                    .copy_file(r)
+                    .await
+                    .context("copy file request failed")?;
+
+                // This is where we set the value for the guest path
                 if src.is_file() {
```
Tim-0731-Hzt, line 101: Thanks @ananos , this PR generally LGTM.  One more question, should the operation above be handled in this if statement?  If it is a none regular file, I think we don't need to copy this guest ?

ananos: you're absolutely right! our bad -- should be now fixed.

ananos:
https://github.com/kata-containers/kata-containers/actions/runs/5134329580/jobs/9238180586?pr=6980#step:10:1914 <- could this be related to this: https://github.com/rust-lang/rust/issues/63033 ?

do we mind if we move to rust 1.69.0?

Tim-0731-Hzt:
> https://github.com/kata-containers/kata-containers/actions/runs/5134329580/jobs/9238180586?pr=6980#step:10:1914 <- could this be related to this: [rust-lang/rust#63033](https://github.com/rust-lang/rust/issues/63033) ?
> 
> do we mind if we move to rust 1.69.0?



It still could work without upgrading to 1.69.0. if you pass the Agent instance as a reference, just like this
```rust
pub(crate) async fn new(
      share_fs: &Option<Arc<dyn ShareFs>>,
      m: &oci::Mount,
      cid: &str,
      readonly: bool,
      agent: &dyn Agent,
  )
```
But I think it's better to use `Arc<dyn Agent>`, as it allows the Agent instance could be shared, and be dropped if no more reference to it. Maybe we should consider move to 1.69 ? cc @quanweiZhou

fidencio:
@ananos, @Tim-0731-Hzt, not exactly related to this PR, but I'd love to see a `SharedFS: none` option in the runtime-rs, which would explicitly copy the files to the guest in case users only want to use devmapper without having to pay the penalty of still having virtiofsd / nydus processes around.

fidencio (approved):
lgtm, thanks @ananos!

Issue #6982, opened 2023-05-29
gha: ci-on-push is broken


sprt:
#6686 broke the CI because of the conditional in the cluster creation step.

See: https://github.com/kata-containers/kata-containers/actions/runs/5111460297/jobs/9188819103

PR #6983, opened 2023-05-29
gha: Unbreak CI and fix cluster creation step

sprt:
This fixes the regression introduced by https://github.com/kata-containers/kata-containers/pull/6686 by properly injecting the `--os-sku mariner --workload-runtime KataMshvVmIsolation` flags.

Error reference:
https://github.com/kata-containers/kata-containers/actions/runs/5111460297/jobs/9188819103

Fixes: https://github.com/kata-containers/kata-containers/issues/6982

### DIFF
--- a/.github/workflows/run-k8s-tests-on-aks.yaml
+++ b/.github/workflows/run-k8s-tests-on-aks.yaml
@@ -52,7 +52,7 @@ jobs:
             -s "Standard_D4s_v5" \
             --node-count 1 \
             --generate-ssh-keys \
-            ${{ matrix.host_os == 'cbl-mariner' && '--os-sku mariner --workload-runtime KataMshvVmIsolation' }}
+            $([ "${{ matrix.host_os == 'cbl-mariner' }}" = "true" ] && echo "--os-sku mariner --workload-runtime KataMshvVmIsolation")
 
       - name: Install `bats`
         run: |


.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -36,6 +36,29 @@ jobs:
         run: |
           curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
 
+          # Enable the AKS Mariner Kata host while it's in preview.
+
+          az extension add --name aks-preview
+          az feature register \
+            --namespace "Microsoft.ContainerService" \
+            --name "KataVMIsolationPreview"
```
fidencio, line 44: Do we really need to do this?

I mean, can't it be done in the Azure account so we avoid having to do the registration here?

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -52,7 +75,7 @@ jobs:
             -s "Standard_D4s_v5" \
             --node-count 1 \
             --generate-ssh-keys \
-            ${{ matrix.host_os == 'cbl-mariner' && '--os-sku mariner --workload-runtime KataMshvVmIsolation' }}
+            $([ "${{ matrix.host_os == 'cbl-mariner' }}" = "true" ] && echo "--os-sku mariner --workload-runtime KataMshvVmIsolation")
```
fidencio, line 78: Ouch.  Yep, I think this one should solve it.

.github/workflows/run-k8s-tests-on-aks.yaml
sprt: Yes that's correct. I don't have access to the Azure subscription yet but if you're able to do that we can remove this part.

fidencio: Done from my side!

fidencio (approved):
lgtm, thanks @sprt!

BbolroC (approved):
lgtm, thanks @sprt

Issue #6984, opened 2023-05-30
Setting up and downloading kernel for tdx using the build-kernel.sh script does not seem to work correctly  


amshinde:
There were recent changes for the tdx kernel in the version.yaml that are not currently accounted for in the build-kernel.sh script.
My recent attempts to setup a tdx kernel to build local changes seemed to not download the tdx kernel. Instead the mainline kernel is downloaded which has no tdx-related changes.

PR #6985, opened 2023-05-30
kernel: Modify build-kernel.sh to accomodate for changes in version.yaml

amshinde:
There were recent changes for the tdx kernel in the version.yaml that are not currently accounted for in the build-kernel.sh script. Attempts to setup a tdx kernel to build local changes seemed to not download the tdx kernel. Instead the mainline kernel is downloaded which has no tdx-related changes.

The version.yaml has a new entry for tdx kernel. Use that instead for setting up and downloading the tdx kernel.

Fixes: #6984


fidencio (approved):
lgtm, thabks @amshinde!

fidencio:
Also, it's worth mentioning that this didn't affect the makefile targets we have for the kata-deploy local build scrips as, there, we already pass the correct URL.

fidencio:
Sorry, at the time you opened your PR we were having issues with our CI, which leads to required tests failing without having nothing to do with this PR.

AS this has been already fixed on  main, we´d like to ask you to rebase your PR atop of the current `main` branch.

Thanks for the consideration, and sorry for the trouble.

dborquez (approved):
LGTM, thanks @amshinde

cmaf (approved):

amshinde:
> Also, it's worth mentioning that this didn't affect the makefile targets we have for the kata-deploy local build scrips as, there, we already pass the correct URL.

Yeah I saw that. But I still think that it would have picked the wrong version,maybe 5.19.2 instead of the TDX 5.19-TDX-v2.2. Not clear how that worked, but it did work as I tested tdx with the latest release.

Issue #6992, opened 2023-05-30
agent: kata-agent should ignore mounts that already happened


zvonkok:
Using an initrd and setting KATA_INIT=yes meaning we're using the kata-agent as the init process we need to make sure that the agent is not segfaulting if mounts are already happened.

Some workloads need to configure several things in the initrd before the kata-agent starts which involves having /proc or /sys already mounted. 

Add functionality to ignore already mounted devs/fs/pseudo-fs.

PR #6993, opened 2023-05-30
agent: Ignore already mounted dev/fs/pseudo-fs

zvonkok:
Using an initrd and setting KATA_INIT=yes meaning we're using the kata-agent as the init process we need to make sure that the agent is not segfaulting if mounts  already happened.

Some workloads need to configure several things in the initrd before the kata-agent starts which involves having /proc or /sys already mounted.

Add functionality to ignore already mounted devs/fs/pseudo-fs.

Fixes: #6992

### DIFF
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -36,6 +36,7 @@ use crate::Sandbox;
 use crate::{ccw, device::get_virtio_blk_ccw_device_name};
 use anyhow::{anyhow, Context, Result};
 use slog::Logger;
+
 use tracing::instrument;
 
 pub const TYPE_ROOTFS: &str = "rootfs";
@@ -1132,6 +1133,7 @@ fn parse_options(option_list: Vec<String>) -> HashMap<String, String> {
 mod tests {
     use super::*;
     use protocols::agent::FSGroup;
+    use slog::Drain;
     use std::fs::File;
     use std::fs::OpenOptions;
     use std::io::Write;
@@ -1142,25 +1144,11 @@ mod tests {
         skip_if_not_root, skip_loop_by_user, skip_loop_if_not_root, skip_loop_if_root,
     };
 
-    // Shadow get_mounts during tests since this is only used in the test
-    // context, compiler will warn about dead-code.
-    #[allow(dead_code)]
-    fn get_mounts() -> Result<String, std::io::Error> {
-        Ok(String::from(
-            "
-            rootfs / rootfs rw,size=1694984k,nr_inodes=423746 0 0
-            proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
-            sys /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
-            dev /dev devtmpfs rw,nosuid,relatime,size=1695000k,nr_inodes=423750,mode=755 0 0
-            run /run tmpfs rw,nosuid,nodev,relatime,mode=755 0 0
-        ",
-        ))
-    }
-
     #[test]
     fn test_already_baremounted() {
-        let drain = slog::Discard;
-        let logger = slog::Logger::root(drain, o!());
+        let plain = slog_term::PlainSyncDecorator::new(std::io::stdout());
+        let logger = Logger::root(slog_term::FullFormat::new(plain).build().fuse(), o!());
+
         let test_cases = [
             ("dev", "/dev", "devtmpfs"),
             ("udev", "/dev", "devtmpfs"),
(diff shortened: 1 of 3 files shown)


sprt:
Fyi you'll have to rebase your branch on top of main to make the GHA tests pass 🙂

src/agent/src/mount.rs
```diff
@@ -168,6 +174,28 @@ pub fn baremount(
         return Err(anyhow!("need mount FS type"));
     }
 
+    let source_str = source.to_string_lossy();
+    let destination_str = destination.to_string_lossy();
+    let mounts = get_mounts().unwrap_or_else(|_| String::new());
```
amshinde, line 179: So the `/proc/mounts` file is read every time we attempt to mount a file/dir. I am wondering if we can do better, get the mounts once and store them as a static resource. This could be a reasonable approach. I dont think having stale mounts would be a concern here. But if it is, then maybe we can attempt to perform the mount first and in case of error check if the  source/dest is already mounted and then return a sucess/failure result.

bergwolf: Actually, I would prefer to keep it as is, because the mount info might change outside of kata-agent. The change is to make the `baremount` function reentrant which is good from an API PoV.

src/agent/src/mount.rs
```diff
@@ -1112,6 +1140,40 @@ mod tests {
         skip_if_not_root, skip_loop_by_user, skip_loop_if_not_root, skip_loop_if_root,
     };
 
+    #[test]
+    // Override get_mounts during tests
+    fn get_mounts() -> Result<String, std::io::Error> {
```
bergwolf, line 1145: Can we override `get_mounts()` like this?

zvonkok: Override is the wrong terminology. Old C/C++ hacker here; I feel guilty :) 

The `get_mounts` here is in the scope of the test module and hence in its namespace. When the tests are not being run, the test module is not included, and the "real" `get_mounts` is available and used. 
When tests are run, the test's `get_mounts` shadows the original function.

zvonkok: I did not want to overcomplicate things; we could also use a trait and mock that trait for testing.

bergwolf: Are you sure that this is working as expected? I tried a bit in rust play ground and it does not seem to work:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3bac5b90a52ff856320116e1ddf76277

zvonkok: @bergwolf Try this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df63af81ac9aab1fef5dfa076f8ba06d

If you do `Run` you will see a full output of a `/proc/mounts` file on the system where the code is executing. 

If you do `Test` you will see the smaller output that is specified in the `mod tets`. 

I added a ` std::assert!(false);` causing a failure, which will trigger `cargo test` to print out stdout, otherwise its suppressed, and the tool does not let me add flags to the `cargo test` command.

zvonkok:
Removed shadowed get_mounts(), added slog-term as a new crate,
slog can directly log to stdout and we can capture output
in the test-cases that are created in the function to be tested.

```
❯ cargo test -- --show-output  mount::tests::test_already_baremounted 
   Compiling kata-agent v0.1.0 (/zvonkok/github.com/zvonkok/kata-containers/src/agent)
    Finished test [unoptimized + debuginfo] target(s) in 8.00s
     Running unittests src/main.rs (target/debug/deps/kata_agent-1bf271d38a434a37)

running 1 test
Jul 07 07:18:40.438 INFO "dev" is already mounted at "/dev", subsystem: baremount
Jul 07 07:18:40.438 INFO "udev" is already mounted at "/dev", subsystem: baremount
Jul 07 07:18:40.439 INFO "proc" is already mounted at "/proc", subsystem: baremount
Jul 07 07:18:40.439 INFO "sysfs" is already mounted at "/sys", subsystem: baremount
test mount::tests::test_already_baremounted ... ok

successes:

---- mount::tests::test_already_baremounted stdout ----
testing if already mounted baremount("dev" "/dev" "devtmpfs")
testing if already mounted baremount("udev" "/dev" "devtmpfs")
testing if already mounted baremount("proc" "/proc" "proc")
testing if already mounted baremount("sysfs" "/sys" "sysfs")


successes:
    mount::tests::test_already_baremounted

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.00s
```

bergwolf (approved):
lgtm, thanks!

bergwolf:
static checks failed though :)

```
error: unused import: `slog::Drain`
  --> src/mount.rs:39:5
   |
39 | use slog::Drain;
   |     ^^^^^^^^^^^
   |
```

zvonkok:
Yeah right, and if I remove `Drain` it complains
 ```
error[E0599]: `FullFormat<PlainSyncDecorator<std::io::Stdout>>` is not an iterator
```

zvonkok:
Ding it, I need Drain only in the test context, updated!

fidencio (approved):
lgtm, thanks @zvonkok!

fidencio:
I've added the `do-not-merge` label just to make sure we have clh-devmapper, qemu-devmapper, firecracker, and the other tests passing.

Issue #6994, opened 2023-05-30
gha: Unable to create a Mariner cluster


sprt:
We're missing the `aks-preview` extension to allow creating a Mariner Kata host.

PR #6995, opened 2023-05-30
gha: Fix Mariner cluster creation

sprt:
While the Mariner Kata host is in preview, we need the `aks-preview` extension to enable the `--workload-runtime KataMshvVmIsolation` flag.

Fixes: #6994


fidencio (approved):
lgtm, thanks @sprt!

amshinde (approved):

Issue #6999, opened 2023-05-31
gha: aks: Add the host_os as part of the aks cluster's name


fidencio:
We need to do so, otherwise we'll create two clusters for testing Cloud
Hypervisor with exactly the same name, one using Ubuntu, and one using
Mariner.

PR #7000, opened 2023-05-31
gha: aks: Add the host_os as part of the aks cluster's name

fidencio:
We need to do so, otherwise we'll create two clusters for testing Cloud Hypervisor with exactly the same name, one using Ubuntu, and one using Mariner.

Fixes: #6999


BbolroC (approved):
lgtm, thanks @fidencio!

PR #6595, opened 2023-04-04
gpu: Build and Ship an GPU enabled Kernel

zvonkok:
With each release make sure we ship a GPU and TEE enabled kernel

Fixes: https://github.com/kata-containers/kata-containers/issues/6553

DependsOn: ~~https://github.com/kata-containers/kata-containers/pull/6522~~

PR #7001, opened 2023-05-31
gpu: NVIDIA gpu initrd/image build

zvonkok:
With each release make sure we ship a GPU enabled rootfs/initrd

Fixes: https://github.com/kata-containers/kata-containers/issues/6554

DependsOn: ~~https://github.com/kata-containers/kata-containers/pull/6664~~  ~~https://github.com/kata-containers/kata-containers/pull/6595~~ ~~https://github.com/kata-containers/kata-containers/pull/6993~~  ~~https://github.com/kata-containers/kata-containers/pull/6949~~


src/agent/rustjail/src/mount.rs
```diff
@@ -265,6 +265,16 @@ pub fn init_rootfs(
         }
     }
 
+
```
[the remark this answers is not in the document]
gkurz: Alright. This is a long standing issue that 9p doesn't honor use-after-unlink on open file descriptors (see https://gitlab.com/qemu-project/qemu/-/issues/103). I tried to fix that when I was actively working on QEMU 9p but it required non-trivial changes on both QEMU and guest 9p driver sides, and I didn't manage to have the 9p driver bits merged.

Good news is that things made substantial progress on the kernel side. It might be worth to resume the QEMU patches (Cc current QEMU 9p maintainer @cschoenebeck).

src/agent/rustjail/src/mount.rs
```diff
@@ -265,6 +265,16 @@ pub fn init_rootfs(
         }
     }
 
+    
+    // Mount /tmp as tmpfs with MS_NODEV and MS_STRICTATIME
```
gkurz, line 269: AFAIK virtiofs does honor the use-after-unlink case gracefully. Maybe only do this for 9p ?

gkurz:
Hi @zvonkok ! Thanks for engaging this huge piece of work.

I've only commented on the 9p workaround you pointed me to so far. I have a few general remarks though :
- the PR contains two commits with the same titles and changelogs with very few information. This makes the PR harder to review since we don't really know what the patches are supposed to achieve
- it seems that a bunch of unrelated changes are squashed together, e.g. passing `-j` to make or switch from gzip to pigz and maybe more

Could you break the PR down in clearly separated commits with informative changelogs ?

Also I've spotted a bunch of space damages that make my `git show` unhappy.

src/agent/Makefile
```diff
@@ -23,7 +23,9 @@ COMMIT_MSG = $(if $(COMMIT),$(COMMIT),unknown)
 # Exported to allow cargo to see it
 export VERSION_COMMIT := $(if $(COMMIT),$(VERSION)-$(COMMIT),$(VERSION))
 
-EXTRA_RUSTFEATURES :=
+EXTRA_RUSTFEATURES :=
```
gkurz, line 26: Trailing space damage.

src/agent/rustjail/src/mount.rs
zvonkok: This is a hack to get me moving. Essentially this shouldn't be needed, so I do not want to carry this on, regardless of the FS used.

Issue #6969, opened 2023-05-26
Dragonball: support to pass cache/numa information into guest on aarch64


xuejun-xj:
The cache/numa information in guest is missing for now. It may bring some improvement on system performance. 

**Solution**

Give the mapping relationships from pcpu to vcpu, and pass it into fdt. When being generated, it will read cache information from sysfs on host and write into fdt. The reading strategy will be implemented in [dragonball-sandbox](https://github.com/openanolis/dragonball-sandbox/issues/274) in future.
Runtime-rs will pass the numa information of host. Then dragonball pass it to create_fdt() function. The numa information will be written in fdt and will be loaded by kernel automatically.

**Limitation**

For now, we may not support correct info of "shared_cpu_list" on L1~L2 cache in guest， unless runtime ensure the correct one-to-one mapping relationship from pcpu to vcpu.

xuejun-xj:
TODO in [dragonball-sandbox](https://github.com/openanolis/dragonball-sandbox): 
- https://github.com/openanolis/dragonball-sandbox/issues/274
- https://github.com/openanolis/dragonball-sandbox/issues/275

PR #7013, opened 2023-06-01
runtime-rs: bugfix: update Cargo.lock

xuejun-xj:
When dragonball update dbs-boot crate in commit
64c764c1475fc9230b4836f5e02c81fc6e1410a7, the Cargo.lock in runtime-rs should also be updated.

Fixes: #6969


fidencio (approved):
lgtm, thanks @xuejun-xj!

studychao (approved):
LGTM

Issue #7017, opened 2023-06-01
Add SLAVE/SHARED flag in bind_mount helper


ananos:
It would be nice if we could specify the flag needed for bind mounts in `bind_mount_unchecked()` (SHARED or SLAVE). This way we could re-use the function to bind mount directories for jailed instances of external hypervisors, in addition to files/dirs for the host/sandbox sharing.

The same functionality could also be used in DBS when using block-device backed storage for the container rootfs.

PR #7018, opened 2023-06-01
runtime-rs: add parameter for propagation of (u)mount events

ananos:
Add an extra parameter in `bind_mount_unchecked` to specify the propagation type: "shared" or "slave".

Fixes: #7017

### DIFF
--- a/src/libs/kata-sys-util/src/mount.rs
+++ b/src/libs/kata-sys-util/src/mount.rs
@@ -103,6 +103,8 @@ pub enum Error {
     MountOptionTooBig,
     #[error("Path for mountpoint is null")]
     NullMountPointPath,
+    #[error("Invalid Propagation type Flag")]
+    InvalidPgMountFlag,
     #[error("Faile to open file {0} by path, {1}")]
     OpenByPath(PathBuf, io::Error),
     #[error("Can not read metadata of {0}, {1}")]
@@ -227,7 +229,13 @@ pub fn bind_remount<P: AsRef<Path>>(dst: P, readonly: bool) -> Result<()> {
     do_rebind_mount(dst, readonly, MsFlags::empty())
 }
 
-/// Bind mount `src` to `dst` in slave mode, optionally in readonly mode if `readonly` is true.
+/// Bind mount `src` to `dst` with a custom propagation type, optionally in readonly mode if
+/// `readonly` is true.
+///
+/// Propagation type: MsFlags::MS_SHARED or MsFlags::MS_SLAVE
+/// MsFlags::MS_SHARED is used to bind mount the sandbox path to enable `exec` (in case of FC
+/// jailer).
+/// MsFlags::MS_SLAVE is used on all other cases.
 ///
 /// # Safety
 /// Caller needs to ensure:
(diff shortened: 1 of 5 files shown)


src/libs/kata-sys-util/src/mount.rs
```diff
@@ -269,7 +270,12 @@ pub fn bind_mount_unchecked<S: AsRef<Path>, D: AsRef<Path>>(
     .map_err(|e| Error::BindMount(abs_src, dst.to_path_buf(), e))?;
 
     // Change into slave propagation mode.
-    mount(Some(""), dst, Some(""), MsFlags::MS_SLAVE, Some(""))
+    let propflag = match propagation {
```
Ji-Xinyou, line 273: The function's comment doc gives
```text
Bind mount `src` to `dst` in slave mode, optionally in readonly mode if `readonly` is true.
```

Can you specify the need for this change? If the change is indeed needed, please change the comment doc on line 230.

Ji-Xinyou (approved):
LGTM, thanks for the work.

src/libs/kata-sys-util/src/mount.rs
```diff
@@ -269,7 +270,12 @@ pub fn bind_mount_unchecked<S: AsRef<Path>, D: AsRef<Path>>(
     .map_err(|e| Error::BindMount(abs_src, dst.to_path_buf(), e))?;
 
     // Change into slave propagation mode.
```
fidencio, line 272: s/into slave/into the chosen/

fidencio (approved):
@ananos, thanks for the PR.

I've left one comment that's a nit and I will leave it to you to address it or not.

lgtm!

src/libs/kata-sys-util/src/mount.rs
```diff
@@ -828,7 +834,7 @@ mod tests {
             Err(Error::InvalidPath(_))
         ));
 
-        bind_mount_unchecked(tmpdir2.path(), tmpdir.path(), true).unwrap();
+        bind_mount_unchecked(tmpdir2.path(), tmpdir.path(), true, "slave").unwrap();
```
Tim-0731-Hzt, line 837: I think you could just pass `MsFlags::MS_SLAVE ` as parameter, so you don't need to match the flag type in `bind_mount_unchecked`

Ji-Xinyou: I agree, since it is not parsed from a configuration.

ananos: sure, fixed!

Tim-0731-Hzt (approved):

studychao (approved):
LGTM, thanks!

src/libs/kata-sys-util/src/mount.rs
```diff
@@ -227,7 +229,7 @@ pub fn bind_remount<P: AsRef<Path>>(dst: P, readonly: bool) -> Result<()> {
     do_rebind_mount(dst, readonly, MsFlags::empty())
 }
 
-/// Bind mount `src` to `dst` in slave mode, optionally in readonly mode if `readonly` is true.
+/// Bind mount `src` to `dst` with a custom propagation type, optionally in readonly mode if `readonly` is true.
 ///
```
Apokleos, line 233: Thx.
Only two propagation types found. I think,  add a comment about types supported may be better.

Issue #7028, opened 2023-06-03
runtime-rs: network is not setup correctly when IP-in-IP encapsulation is on


ananos:
when setting up the network, we check whether an interface has the `IFF_LOOPBACK` flag and we skip it. However, is `ipip` is enabled, then the `tunl0` appears first in the new network namespace.

This leads to Dragonball sandboxes not working with the following message:

```
FATA[0000] failed to create shim task: Others("failed to handler message try init runtime instance\n\nCaused by:\n    0: init runtime handler\n    1: start sandbox\n    2: set up device after start vm\n    3: failed to set up network\n    4: new network\n    5: new network with netns\n    6: get entity from netns\n    7: create endpoint\n    8: ioctl ethtool\n    9: EOPNOTSUPP: Operation not supported on transport endpoint"): unknown 
```

This is because we treat `tunl0` as a device interface.

A workaround is to set `net.core.fb_tunnels_only_for_init_net` to 1.

```
sysctl -w net.core.fb_tunnels_only_for_init_net=1
```
This will prevent tunl0 to appear in a new, "non-init" netns. 

Another way to go is to also check for the `IFF_NOARP` flag maybe?

ananos:
closing as duplicate of #7020

PR #7021, opened 2023-06-02
runtime-rs: ignore unconfigured network interfaces

cheriL:
Ignore unconfigured network interfaces like tunl0 when create endpoints.

Fixes: #7020

PR #7029, opened 2023-06-03
runtime-rs: skip ipip network interfaces (iff_noarp)

ananos:
when setting up the network, we check whether an interface has the IFF_LOOPBACK flag and we skip it. However, if ipip is enabled, the tunl0 appears first in the new network namespace.

During this network link discovery, we find it first, and try to add it as an endpoint.

Adding the IFF_NOARP flag makes it skip this, and proceed to the veth interface.

A workaround is to set net.core.fb_tunnels_only_for_init_net to 1.

Fixes: #7028

### DIFF
--- a/src/runtime-rs/crates/resource/src/network/network_with_netns.rs
+++ b/src/runtime-rs/crates/resource/src/network/network_with_netns.rs
@@ -173,7 +173,7 @@ async fn get_entity_from_netns(config: &NetworkWithNetNsConfig) -> Result<Vec<Ne
         let link = link::get_link_from_message(link);
         let attrs = link.attrs();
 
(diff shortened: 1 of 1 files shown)


Ji-Xinyou:
This qemu-devmapper CI is constantly failing

```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
[install_cri_containerd.sh:78] ERROR: sudo tar xfz - -C /usr/local/bin
[jenkins_job_build.sh:246] ERROR: ci/setup.sh
Build step 'Execute shell' marked build as failure
```

Seems like a deterministic problem in the CI script.

ananos:
closing as duplicate of #7021 -- that solutions seems more aligned with the go runtime.

Issue #7032, opened 2023-06-05
agent-ctl: could not compile due to dependency error


YushuoEdge:
This will fail the CI

```
error[E0061]: this function takes 3 arguments but 2 arguments were supplied
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.3/src/util_libc.rs:110:14
    |
110 |     let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC);
    |              ^^^^------------------------------------------------------------- an argument of type `u32` is missing
    |
note: function defined here
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.145/src/unix/linux_like/linux/musl/lfs64.rs:112:26
    |
112 | pub unsafe extern "C" fn open64(
    |                          ^^^^^^
help: provide the argument
    |
110 |     let fd = open(path.as_ptr() as *const _, libc::O_RDONLY | libc::O_CLOEXEC, /* u32 */);
    |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0061`.
error: could not compile `getrandom` due to previous error
warning: build failed, waiting for other jobs to finish...
make[1]: *** [Makefile:16: build] Error 101
make[1]: Leaving directory '/home/runner/work/kata-containers/kata-containers/src/github.com/kata-containers/kata-containers/src/tools/agent-ctl'
make: *** [Makefile:33: agent-ctl] Error 2
Error: Process completed with exit code 2.
```

PR #7033, opened 2023-06-05
agent-ctl: fix the compile error

YushuoEdge:
When the version of `libc` is upgraded to 0.2.145, `getrandom` could not adapt to new API, and this will make `agent-ctl` fail to compile.

We upgrade the version of `rand`, so the low version of `getrandom` will no longer need.

Fixes: #7032


studychao:
I don't think this commit is a `workaround`. Please adjust your commit message.

GabyCT (approved):

fidencio:
I've restarted the 2 required jobs that failed.

fidencio (approved):
lgtm, thanks @YushuoEdge!

Let's wait for the CI to pass, and merge it ASAP.

PR #7042, opened 2023-06-06
Revert "agent: fix the issue of exec hang with a backgroud process"

fidencio:
This reverts commit 25d2fb0fdecf5d76fcb8a4ecde5c54c0d4c3e240.

The reason we're reverting the commit is because it to check whether it's the cause for the regression on devmapper tests.

Fixes: #0000

### DIFF
--- a/src/agent/rustjail/src/process.rs
+++ b/src/agent/rustjail/src/process.rs
@@ -161,7 +161,7 @@ impl Process {
 
     pub fn notify_term_close(&mut self) {
         let notify = self.term_exit_notifier.clone();
-        notify.notify_waiters();
+        notify.notify_one();
     }
 
     pub fn close_stdin(&mut self) {
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -595,16 +595,15 @@ impl AgentService {
         let cid = req.container_id;
         let eid = req.exec_id;
 
-        let term_exit_notifier;
+        let mut term_exit_notifier = Arc::new(tokio::sync::Notify::new());
         let reader = {
             let s = self.sandbox.clone();
             let mut sandbox = s.lock().await;
 
             let p = sandbox.find_container_process(cid.as_str(), eid.as_str())?;
 
-            term_exit_notifier = p.term_exit_notifier.clone();
-
             if p.term_master.is_some() {
+                term_exit_notifier = p.term_exit_notifier.clone();
                 p.get_reader(StreamType::TermMaster)
             } else if stdout {
                 if p.parent_stdout.is_some() {
--- a/tests/integration/kubernetes/k8s-number-cpus.bats
+++ b/tests/integration/kubernetes/k8s-number-cpus.bats
@@ -9,6 +9,8 @@ load "${BATS_TEST_DIRNAME}/../../common.bash"
 load "${BATS_TEST_DIRNAME}/tests_common.sh"
 
 setup() {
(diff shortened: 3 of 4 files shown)


fidencio:
https://github.com/kata-containers/tests/issues/5682 is the issue raised, and if we can fix it on this repo, I'll move the issue here.

fidencio:
@lifupan, I don't think reverting is the way to go, but I'd like to ask if you have some cycles to take a look at the regression introduced by the commit.

lifupan:
Hi @fidencio 

I had tried to reproduce this issue, but actually I haven't by now.

I had noticed that the failed cases were not always the same one, some times the [run_kubernetes_tests.sh:153] ERROR: bats k8s-pid-ns.bats failed, sometimes failed on "[run_kubernetes_tests.sh:153] ERROR: bats k8s-copy-file.bats", I'd try to figure out the root cause later, and it wouldn't be a big issue, so I think we can just keep this PR open until I find the root cause.

fidencio:
> Hi @fidencio
> 
> I had tried to reproduce this issue, but actually I haven't by now.
> 
> I had noticed that the failed cases were not always the same one, some times the [run_kubernetes_tests.sh:153] ERROR: bats k8s-pid-ns.bats failed, sometimes failed on "[run_kubernetes_tests.sh:153] ERROR: bats k8s-copy-file.bats", I'd try to figure out the root cause later, and it wouldn't be a big issue, so I think we can just keep this PR open until I find the root cause.

Hey @lifupan, I agree. :-)
The consistent failing that I got on the systems using devmapper was on the k8s-copy-file.bats, in case that helps.

wainersm:
@lifupan @fidencio the firecracker CI jobs leverage device-mapper, they are mostly failing at this point and preventing PRs from being merged. Could we merge this revert now? An alternative is to make the firecracker jobs non-required....

fidencio:
After running tests here I can see the following behavior.

With https://github.com/kata-containers/kata-containers/commit/25d2fb0fdecf5d76fcb8a4ecde5c54c0d4c3e240 reverted, I got a green run for Cloud Hypervisor, and QEMU.

Dragonball was still failing in two tests:
* k8s-numbers-cpus
* k8s-oom

With those two tests disabled for dragonball, I got 5 consecutive green runs for the dragonball Ci:
* https://github.com/kata-containers/kata-containers/actions/runs/5493775117/jobs/10012365165
* https://github.com/kata-containers/kata-containers/actions/runs/5493775117/jobs/10012505081
* https://github.com/kata-containers/kata-containers/actions/runs/5493775117/jobs/10012609881
* https://github.com/kata-containers/kata-containers/actions/runs/5493775117/jobs/10012720546
* https://github.com/kata-containers/kata-containers/actions/runs/5493775117/jobs/10012920088

This makes me seriously believe that we should revert 25d2fb0fdecf5d76fcb8a4ecde5c54c0d4c3e240 ASAP.

fidencio:
What I'd like to do here is ...

* Have 2 issues opened. One for k8s-number-cpus and one for k8s-oom
  * Done as: https://github.com/kata-containers/kata-containers/issues/7270
  * Done as: https://github.com/kata-containers/kata-containers/issues/7271
* Skip those 2 tests
  * Done as part of this PR
* Have this PR merged

This will allow us to have a CI running ASAP.

I'll then do:
* Work on k8s-number-cpus and have it re-enabled
* Work on k8s-oom and have it re-enabled

Does it sound like a fair plan to everyone involved?

fidencio:
I am closing this one, as it's part of https://github.com/kata-containers/kata-containers/pull/7272

PR #7050, opened 2023-06-06
tests: Move tests repo to kata-containers repo

dborquez:
To continue the effort to adopt Github actions to run our tests, the
first part is moving all the tests to this repo.

Following the recommendation given in the commit 11e0099, we decided to
do not keep the tests history to avoid increasing as much as possible the
size of this repository.

### DIFF
--- a/src/tools/log-parser-rs/README.md
+++ b/src/tools/log-parser-rs/README.md
@@ -48,7 +48,7 @@ For a comprehensive (and guaranteed up to date) list, please run `log-parser-rs 
 
 ## Usage
 
-1. Make sure containerd is in [debug mode](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#enabling-full-containerd-debug)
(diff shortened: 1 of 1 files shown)


GabyCT (approved):

fidencio:
@dborquez, let's take a step back. :-)

We don't want to move the **whole** tests repo into the `kata-containers` one.  Instead, we want to move only the pieces and bits that you'll be adding tests for.

For instance, if adding tests related to k8s, then you move the tests/integration/kubernetes.  If adding tests related to metrics, then you move only tests/metrics.

Also, please, keep in mind that we don't need the majority of the scripts anymore, as the entry point will be different, as the base system may be different.  So, we'll end up cleaning up a lot the content we move hee, instead of doing a full move.

Does this sound reasonable?

cmaf (commented):
@dborquez nothing to add from me, I agree with @fidencio. There's something weird going on with the commit though, there are 2 visible and one says it can't be displayed because it doesn't share a common merge base with the base branch

Issue #7049, opened 2023-06-06
Migrate the launch-times test and helper functions to kata-containers/tests


dborquez:
To continue the effort to adopt GitHub actions to run our metrics, it is required to move the following tests/artifacts to the kata-containers/metrics repository:

- [Launch-times](https://github.com/kata-containers/tests/tree/main/metrics/time) test

dborquez:
This issue is reopened because the previous PR #7138 only prints a couple of messages, but did not enabled the the launch-times script.

PR #7052, opened 2023-06-07
tests: Move tests/metrics folder to kata-containers/tests repo

dborquez:
tests: Move tests/metrics folder to kata-containers/tests repo

To continue the effort to adopt Github actions to run our tests, the
requirement is to move all the tests/metrics folder to this repo.

Following the recommendation given in the commit 11e0099, we decided to
do not keep the tests history to avoid increasing as much as possible the
size of this repository.

Fixes: #7049


fidencio:
@dborquez, I´d  prefer doing this as part of the series that will actually use it.

Then, whenever we do this, we make sure to visit and check what parts of the scripts are actually needed.
For instance, the whole `libs/common.bash` has to be carefully examinated, as we're already using https://github.com/kata-containers/kata-containers/blob/main/tests/common.bash ...  So, grow the file that we have, only with the functions that you'll for sure need, instead of creating a new file that will have duplication.

The same goes for lib/error.sh, which has some of those functions as part of `common.bash`.

Everything under tests/metrics/ is fine, but we really need to make sure the scripts will work.

About versions.yaml, I'm not sure if we need that, or at least not in the way it's.

Let's try to get some time Tomorrow and sync about this, in case you want to.

fidencio (changes requested):
Adding "Request changes" to avoid it getting prematurely merged.

PR #7065, opened 2023-06-08
versions: tdx: Update content to the 2023ww22 release

fidencio:
2023ww22 is the latest release based on the 5.19 kernel, targetting TDX 1.0, and can be found here: https://github.com/intel/tdx-tools/tree/2023ww22

We should move our components to that as it brings in some optimisations / stability related patches.


arronwy (approved):
Thanks @fidencio verified on my TDX server with WW22 host environment

GabyCT (approved):

Issue #7059, opened 2023-06-07
Incorrect qemu build for SNP


UnmeshDeodhar:
AMD SEV-SNP requires special qemu that can be built from: https://github.com/AMDESE/qemu
Right now, we are building generic qemu for SNP as well in `install_cc_tee_qemu`

This needs to change and we should build experimental qemu for SNP.

As part of [this ](https://github.com/kata-containers/kata-containers/commit/33143eb342221aca9897543b2bf66929d3083418#diff-1c4cd801df522f4a92edbfb0fea95364ed074a391ea47c284ddc078f512f7b6aL105) change, qemu entry in the version.yaml was changed.

PR #7070, opened 2023-06-09
Fix qemu snp prefix and name

UnmeshDeodhar:
Main merge back to CCv0 caused snp qemu build to move from install_qemu to install_qemu_experimental.
Thus, reflecting this change into the snp qemu command. Also changing the prefix to /opt/confidential-containers from /opt/kata for the CCv0 branch.

Fixes: #7059 

Signed-Off-By: Unmesh Deodhar <udeodhar@amd.com>


tools/packaging/static-build/qemu/build-base-qemu.sh
```diff
@@ -31,7 +31,7 @@ info "Build ${qemu_repo} version: ${qemu_version}"
 
 http_proxy="${http_proxy:-}"
 https_proxy="${https_proxy:-}"
-prefix="${prefix:-"/opt/kata"}"
```
stevenhorsman, line 34: So my understanding, which might well be wrong, is that for CC this script is called from build-static-qemu-cc.sh, which should already have the prefix exported?

https://github.com/kata-containers/kata-containers/blob/8330fdeadfd7fb92793a642a89e5de7fb2bbdafc/tools/packaging/static-build/qemu/build-static-qemu-cc.sh#L19

fidencio: Yep, that's exactly the case, @stevenhorsman!

UnmeshDeodhar: Please correct me if I am missing something @stevenhorsman @fidencio 
But based on my understanding since QEMU build is now moved to experimental, we will now use `build-static-qemu-experimental.sh` instead of `build-static-qemu-cc.sh`
This is being called from here: https://github.com/kata-containers/kata-containers/blob/d742fe37a92a7be600f2fb2d676108697b91abad/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L810

stevenhorsman (approved):
I think this is better than needing to hard-code the prefix for snp_qemu_experimental. I still have some questions about how we are going to reconcile this to get CC support on `main`, but that's beyond the scope of this PR

UnmeshDeodhar:
@fidencio can you please confirm if tdx failures are related to this PR?

fidencio (approved):
lgtm, thanks!

Issue #7030, opened 2023-06-04
Using Symbolic link file pointing to “/proc","/sys", runtime still can mount host's /proc, /sys, /dev into containers.


InstantN:
# Description of problem
#6082 can specify a particular mount path that allows runtime mount host's /proc, /sys, and /dev into containers. This issue has been fixed by #6369. 

However, I found that the current check method can still be bypassed by setting a specific symlink directory.

Steps to reproduce:
1. Create a symbolic link directory pointing to “/”
```bash
mkdir /tmp/evil
ln -s / /tmp/evil
```
2. Modify the `host_path` in `container.yaml` mentioned in #6082.
```json
{
    "metadata": {
    	"name": "ubuntu-0",
    	"namespace": "default"
    },
    "image": {
        "image": "docker.io/library/ubuntu:latest"
    },
    "command": [
        "sleep",
        "99999999"
        ],
    "linux": {},
    "mounts": [{
	    "host_path": "/tmp/evil/proc",
	    "container_path": "/test-dir",
	    "readonly": false,
	    "propagation": 2
    }],
    "args": [],
    "working_dir": "/",
    "log_path": ""
}
```
3. The remaining steps are the same as mentioned in #6082. Use crictl + containerd to run the kata container.

# Expected result

Kata-runtime should reject mounting the host's system directory into container, such as /proc,/sys, etc. Special symlink directories need to be considered.

# Actual result

Kata mounts the host's /proc, /sys, and /dev into container with a crafted host symlink.
With mounting the host's `/proc/sysrq-trigger`, the container can cause the host to reboot by writing to this file.

# Further information
I want to confirm if this is a real issue, and if so, can I get a CVE number :)

fengwang666:
@InstantN Thanks for reporting this. Can you tell me the exact version of the kata containers you run the test on? I'll try to repro it on my side.

PR #7074, opened 2023-06-09
runtime: check symlink on a host path mount

fengwang666:
Skip the mount if it's a symlink to a host system mount

Fixes: #7030

### DIFF
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -435,13 +435,6 @@ func (c *Container) mountSharedDirMounts(ctx context.Context, sharedDirMounts, i
 	}()
 
 	for idx, m := range c.mounts {
-		// Skip mounting certain system paths from the source on the host side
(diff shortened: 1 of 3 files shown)


XDTG:
Hi @ fengwang666, there are some little problems with this patch. 

1. In `resolveSymlink()`, `os.Lstat(path)` can only obtain the status of the **last element** in the path. 
For example, if `path` is "/tmp/evil-symlink"("evil-symlink" is a symlink to "/"), the status of "evil-symlink" will be returned, with its mode containing `os.ModeSymlink`. But if `path` is "/tmp/evil-symlink/proc" or "/tmp/evil-symlink/", `os.Lstat(path)` will return the status of "/proc" or "/", without "os.ModeSymlink".

2. In `resolveSymlink()`, `filepath.Join(filepath.Dir(path), target)` is **only appropriate when the target is a relative path**. When the target is an absolute path, join them will cause an mistake.

Issue #7075, opened 2023-06-11
runtime-rs: large amounts of log will cause shim panic.


yipengyin:
```
A panic occurred at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/slog-2.7.0/src/lib.rs:1944: slog::Fuse Drain: Custom { kind: Other, error: "serde serialization error: Resource temporarily unavailable (os error 11)" }
0: shim::panic_hook::set_panic_hook::{{closure}}
1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/alloc/src/boxed.rs:2001:9
    std::panicking::rust_panic_with_hook
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:692:13
2: std::panicking::begin_panic_handler::{{closure}}
            at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:579:13
3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/sys_common/backtrace.rs:137:18
```

For instance, when add lots of large env for container in pod yaml,  `spec_value` of `ExecProcessRequest` will be enormous. So, the logger will print a large log entry for `ExecProcessRequest` when log level is set to `debug`. In such circumstance, shim will crash.

PR #7076, opened 2023-06-11
fix(runtime-rs): disable nonblock for logger fifo

yipengyin:
Since async slog thread is synchronous, writer might report EAGAIN(11) when it writes large amounts of log to nonblock fifo. And it will cause shim panic. Therefore we disable noblock for fifo.

Fixes: #7075

### DIFF
--- a/src/runtime-rs/crates/shim/src/logger.rs
+++ b/src/runtime-rs/crates/shim/src/logger.rs
@@ -4,15 +4,15 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use std::os::unix::fs::OpenOptionsExt;
-
 use anyhow::{Context, Result};
 
 use crate::Error;
 
 pub(crate) fn set_logger(path: &str, sid: &str, is_debug: bool) -> Result<slog_async::AsyncGuard> {
+    // Since slog-async-logger is a synchronous thread actually, the log fifo can not be
+    // asynchronous. Otherwise, writer might report EAGAIN(11) when it writes large amounts
+    // of log to nonblock fifo, which will cause shim panic.
     let fifo = std::fs::OpenOptions::new()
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/shim/src/logger.rs
```diff
@@ -4,15 +4,12 @@
...
-use std::os::unix::fs::OpenOptionsExt;
-
 use anyhow::{Context, Result};
 
 use crate::Error;
 
 pub(crate) fn set_logger(path: &str, sid: &str, is_debug: bool) -> Result<slog_async::AsyncGuard> {
     let fifo = std::fs::OpenOptions::new()
-        .custom_flags(libc::O_NONBLOCK)
         .create(true)
```
quanweiZhou, line 13: Suggest adding some comment.

yipengyin: done

quanweiZhou (approved):

src/runtime-rs/crates/shim/src/logger.rs
```diff
@@ -4,15 +4,15 @@
...
 //
 
-use std::os::unix::fs::OpenOptionsExt;
-
 use anyhow::{Context, Result};
 
 use crate::Error;
 
 pub(crate) fn set_logger(path: &str, sid: &str, is_debug: bool) -> Result<slog_async::AsyncGuard> {
+    // Since slog-async-logger is a synchronous thread actually, he log fifo can not be
```
studychao, line 12: `he` -> `the`

yipengyin: oops...

yipengyin: done

Issue #7077, opened 2023-06-12
agent: TERM signal from container will cause agent crash


yipengyin:
It seems that TERM signal from container might cause agent crash. I don't have easy way to reproduce the problem. But setsid when container initialized could solve the problem.

PR #7078, opened 2023-06-12
fix(agent): setsid to ensure signal segregation

yipengyin:
The signal sent by container to process might leak to kata-agent. And the TERM signal will cause agent crash. So we must setsid when container is initialized.

Fixes: #7077


quanweiZhou (approved):
Thanks, @yipengyin LGTM.

Issue #7080, opened 2023-06-12
agent: destroying systemd cgroup manager failed due to NoSuchUnit error


yoheiueda:
In peer pods, kata-agent is launched as a systemd unit, and the systemd cgroups manager is used to control cgroups.

When I remove a container managed by a kata-agent, the `destroy` method of the systemd cgroups manager (almost) always fails with the following error message.

```
rpc error: code = Internal desc = destroy cgroups

Caused by:
    0: failed to stop unit cri-containerd-67802846e448f429f9d4ac418b05ae7f10c6b977b20f71d9ac0a3b8deb5d5843.scope
    1: org.freedesktop.systemd1.NoSuchUnit: Unit cri-containerd-67802846e448f429f9d4ac418b05ae7f10c6b977b20f71d9ac0a3b8deb5d5843.scope not loaded.
```

This is because all processes in the container are killed just before the `destroy` method of the cgroup manager is called.

https://github.com/kata-containers/kata-containers/blob/29885533050d133df86d06edc1a5ee8825129f1f/src/agent/rustjail/src/container.rs#L1139-L1150

The systemd cgroup manager creates a *transient* systemd unit for a container, and systemd automatically deletes the transient unit when all processes in it exit.

https://github.com/kata-containers/kata-containers/blob/29885533050d133df86d06edc1a5ee8825129f1f/src/agent/rustjail/src/cgroups/systemd/dbus_client.rs#L78-L80

The transient unit is deleted when all processes are killed, so `stop_unit` method in the `destroy` method fails.

https://github.com/kata-containers/kata-containers/blob/29885533050d133df86d06edc1a5ee8825129f1f/src/agent/rustjail/src/cgroups/systemd/manager.rs#L83-L86

This problem is reported in https://github.com/confidential-containers/cloud-api-adaptor/issues/1072

PR #7081, opened 2023-06-12
agent: Ignore no such unit error in the destroy method of the systemd cgroups manager

yoheiueda:
The systemd cgroups manager calls the `stop_unit` method, but the target transient unit does not exist, since all processes in a container are already killed as described in #7080.

https://github.com/kata-containers/kata-containers/blob/29885533050d133df86d06edc1a5ee8825129f1f/src/agent/rustjail/src/container.rs#L1139-L1150

This patch changes the `destroy` method of systemd cgroups manager to ignore no such unit error.

Fixes: #7080

### DIFF
--- a/src/agent/rustjail/src/cgroups/systemd/manager.rs
+++ b/src/agent/rustjail/src/cgroups/systemd/manager.rs
@@ -81,7 +81,16 @@ impl CgroupManager for Manager {
     }
 
     fn destroy(&mut self) -> Result<()> {
-        self.dbus_client.stop_unit(self.unit_name.as_str())?;
+        self.dbus_client
+            .stop_unit(self.unit_name.as_str())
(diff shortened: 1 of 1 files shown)


Yuan-Zhuo:
Hello, I am doing some cleanup work on systemd cgroup code. I also found the problem you mentioned, and your error handling method is exactly what I expect. Do you agree to submit a co-signed PR?

yoheiueda:
@Yuan-Zhuo sorry for the late response. I was on vacation.

> Do you agree to submit a co-signed PR with some other fixes for systemd cgroups?

Yes! Please go ahead with your PR.

yoheiueda:
This issue will be address in https://github.com/kata-containers/kata-containers/pull/7151.

I close this PR.

Issue #7086, opened 2023-06-12
gha: `stage` isn't defined properly


sprt:
#7027 broke the GHA CI: https://github.com/kata-containers/kata-containers/actions/runs/5247059497

PR #7087, opened 2023-06-12
gha: Fix `stage` definition in matrix

sprt:
This defines `stage` as a list instead of a literal to fix the GHA CI.

Fixes: #7086


fidencio (approved):
lgtm, thanks @sprt!

jepio (approved):

PR #7095, opened 2023-06-13
docs: Update Developer Guide

### LINKED ISSUE #7094: update developer guide

GabyCT:
This PR updates the developer guide at the connect to the debug console section.

Fixes #7094

### DIFF
--- a/docs/Developer-Guide.md
+++ b/docs/Developer-Guide.md
@@ -587,10 +587,15 @@ $ sudo kata-monitor
 
 #### Connect to debug console
 
-Command `kata-runtime exec` is used to connect to the debug console.
+You need to start a container for example:
+```bash
+$ sudo ctr run  --runtime io.containerd.kata.v2 -d docker.io/library/ubuntu:latest testdebug
+```
+
+Then, you can use the command `kata-runtime exec <sandbox id>` to connect to the debug console.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/Developer-Guide.md
```diff
@@ -587,10 +587,15 @@ $ sudo kata-monitor
 
 #### Connect to debug console
 
-Command `kata-runtime exec` is used to connect to the debug console.
+You need to start a container for example:
+```bash
+$ sudo ctr run  --runtime io.containerd.kata.v2 -d docker.io/library/ubuntu:latest testdebug
+```
+
+Then, you can use the command `kata-runtime exec` to connect to the debug console.
```
stevenhorsman, line 595: I wonder if adding that the sandbox id is the parameter needed would help with clarity here?
```suggestion
Then, you can use the command `kata-runtime exec <sandbox id>` to connect to the debug console.
```

stevenhorsman (approved):
Thanks @GabyCT - I think this helps make it clearer - I just have a minor suggestion that might also be worth considering?

docs/Developer-Guide.md
GabyCT: @stevenhorsman thanks for the feedback, changes applied

dborquez (approved):
lgtm

Issue #7098, opened 2023-06-13
gha: naked Makefile definition leads to error


sprt:
#7087 didn't fix things completely, see https://github.com/kata-containers/kata-containers/actions/runs/5258732615/jobs/9503343282

PR #7099, opened 2023-06-13
tools: Fix no-op builds

sprt:
This fixes the builds of `cloud-hypervisor-glibc` and `rootfs-initrd-mariner` to properly create the `build/` directory.

Fixes: #7098


stevenhorsman (approved):
LGTM. Thanks!

Issue #7100, opened 2023-06-13
Add workflow for metrics


dborquez:
To continue the effort to migrate the test repository, we must first generate a basic github workflow to start testing metrics.

PR #7101, opened 2023-06-13
gha: ci-on-push: Run metrics tests

dborquez:
This gh-workflow prints a simple msg, but is the base for future PRs that will gradually add the jobs corresponding to the kata metrics test.

Fixes: #7100


GabyCT (approved):

fidencio (commented):
@dborquez, I´d recommend, for the sake of easily testing further interatcions, creating a bash script that you'll call as part of the `run`.

This will help us to just change the script content when actually adding the tests.
WDYT?

dborquez:
@fidencio makes sense to me, thanks, so I've added the script, and I appreciate  your review :)

.github/workflows/run-metrics-tests.yaml
```diff
@@ -0,0 +1,12 @@
...
+  workflow_call:
+
+jobs:
+  launch-times-tests:
+    runs-on: metrics
+    env:
+      GOPATH: ${{ github.workspace }}
+    steps:
+    - name: run launch times
+      run: bash tests/metics/gha-run.sh run-test-launchtimes
```
sprt: Small typo here 🙂 
```suggestion
      run: bash tests/metrics/gha-run.sh run-test-launchtimes
```

sprt (changes requested):

.github/workflows/run-metrics-tests.yaml
dborquez: @sprt thank you, typo fixed

fidencio (approved):
lgtm, thanks @dborquez!

dborquez:
@sprt, could you please re-review this PR?

PR #7106, opened 2023-06-14
Image pulling on the host

wedsonaf:
These are the changes in the kata runtime and agent that enable pulling images on the host and exposing layers as block devices to the VM, which are then mounted and overlayed to produce the final image. This layering allows sharing by containers with common layers, which is a desirable to reduce the memory footprint (and similar to what the overlay snapshotter does on the host). There is no dependency on image-rs.

The changes here are independent of which file system is used. Everything is driven by mount options, which can be used with any file system and snapshotter. (I believe this preferred over hard-coding file system names and functions in generic kata code.)

We have a reference snapshotter (tardev) that uses tarfs as the file system -- both are available here: https://github.com/kata-containers/tardev-snapshotter. We also have a fuse version of tarfs that we'll upload soon, as well as changes needed to enable dm-verity, the PRs for which will be uploaded once this is merged.

The kernel module is 664 LoC and has been tested in 5.15 and 6.1 kernels (the two latest LTSs). The fuse version is written in Rust and is currently 327 LoC.  The tardev snapshotter is written in Rust and totals 1095 LoC. It does not depend on virtiofs at all; it downloads layers, decompresses them, and appends an index to allow for efficient traversal by the fs impls. (In a subsequent PR, it will also append a dm-verity hash tree.)

This is the initial implementation of what I described in the Jan 26th meeting.

### DIFF
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -813,10 +813,14 @@ fn parse_mount_flags_and_options(options_vec: Vec<&str>) -> (MsFlags, String) {
                     }
                 }
                 None => {
+                    if opt.starts_with("io.katacontainers.") {
+                        continue;
+                    }
+
                     if !options.is_empty() {
                         options.push_str(format!(",{}", opt).as_str());
                     } else {
-                        options.push_str(opt.to_string().as_str());
+                        options.push_str(opt);
                     }
                 }
             };
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -340,8 +340,33 @@ pub async fn update_ephemeral_mounts(
 async fn overlayfs_storage_handler(
     logger: &Logger,
     storage: &Storage,
+    cid: Option<&str>,
     _sandbox: Arc<Mutex<Sandbox>>,
 ) -> Result<String> {
+    if storage
+        .options
+        .iter()
+        .any(|e| e == "io.katacontainers.fs-opt.overlay-rw")
+    {
+        let cid = cid.ok_or_else(|| anyhow!("No container id in rw overlay"))?;
+        let cpath = Path::new(crate::rpc::CONTAINER_BASE).join(cid);
+        let work = cpath.join("work");
+        let upper = cpath.join("upper");
+
+        fs::create_dir_all(&work).context("Creating overlay work directory")?;
+        fs::create_dir_all(&upper).context("Creating overlay upper directory")?;
+
+        let mut storage = storage.clone();
+        storage.fstype = "overlay".into();
+        storage
+            .options
+            .push(format!("upperdir={}", upper.to_string_lossy()));
+        storage
+            .options
+            .push(format!("workdir={}", work.to_string_lossy()));
+        return common_storage_handler(logger, &storage);
+    }
+
     common_storage_handler(logger, storage)
 }
 
@@ -869,7 +894,7 @@ pub async fn add_storages(
                 ephemeral_storage_handler(&logger, &storage, sandbox.clone()).await
             }
             DRIVER_OVERLAYFS_TYPE => {
-                overlayfs_storage_handler(&logger, &storage, sandbox.clone()).await
+                overlayfs_storage_handler(&logger, &storage, cid.as_deref(), sandbox.clone()).await
             }
             DRIVER_MMIO_BLK_TYPE => {
                 virtiommio_blk_storage_handler(&logger, &storage, sandbox.clone()).await
--- a/src/runtime/pkg/govmm/qemu/qmp.go
+++ b/src/runtime/pkg/govmm/qemu/qmp.go
@@ -802,7 +802,12 @@ func (q *QMP) blockdevAddBaseArgs(driver string, blockDevice *BlockDevice) map[s
 // used to name the device.  As this identifier will be passed directly to QMP,
 // it must obey QMP's naming rules, e,g., it must start with a letter.
 func (q *QMP) ExecuteBlockdevAdd(ctx context.Context, blockDevice *BlockDevice) error {
-	args := q.blockdevAddBaseArgs("host_device", blockDevice)
+	var args map[string]interface{}
+	if fi, err := os.Stat(blockDevice.File); err == nil && fi.Mode().IsRegular() {
+		args = q.blockdevAddBaseArgs("file", blockDevice)
+	} else {
+		args = q.blockdevAddBaseArgs("host_device", blockDevice)
+	}
 
 	return q.executeCommand(ctx, "blockdev-add", args, nil)
 }
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -907,6 +907,8 @@ func (k *kataAgent) removeIgnoredOCIMount(spec *specs.Spec, ignoredMounts map[st
 	for _, m := range spec.Mounts {
 		if _, found := ignoredMounts[m.Source]; found {
 			k.Logger().WithField("removed-mount", m.Source).Debug("Removing OCI mount")
+		} else if HasOption(m.Options, vcAnnotations.IsFileSystemLayer) {
+			k.Logger().WithField("removed-mount", m.Source).Debug("Removing layer")
 		} else {
 			mounts = append(mounts, m)
 		}
@@ -1293,12 +1295,16 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
 
 	// Block based volumes will require some adjustments in the OCI spec, and creation of
 	// storage objects to pass to the agent.
-	volumeStorages, err := k.handleBlkOCIMounts(c, ociSpec)
+	layerStorages, volumeStorages, err := k.handleBlkOCIMounts(c, ociSpec)
 	if err != nil {
 		return nil, err
 	}
 
 	ctrStorages = append(ctrStorages, volumeStorages...)
+
+	// Layer storage objects are prepended to the list so that they come _before_ the
+	// rootfs because the rootfs depends on them (it's an overlay of the layers).
+	ctrStorages = append(layerStorages, ctrStorages...)
 
 	grpcSpec, err := grpc.OCItoGRPC(ociSpec)
 	if err != nil {
@@ -1611,9 +1617,10 @@ func (k *kataAgent) createBlkStorageObject(c *Container, m Mount) (*grpc.Storage
 // handleBlkOCIMounts will create a unique destination mountpoint in the guest for each volume in the
 // given container and will update the OCI spec to utilize this mount point as the new source for the
 // container volume. The container mount structure is updated to store the guest destination mountpoint.
-func (k *kataAgent) handleBlkOCIMounts(c *Container, spec *specs.Spec) ([]*grpc.Storage, error) {
+func (k *kataAgent) handleBlkOCIMounts(c *Container, spec *specs.Spec) ([]*grpc.Storage, []*grpc.Storage, error) {
 
 	var volumeStorages []*grpc.Storage
+	var layerStorages []*grpc.Storage
 
 	for i, m := range c.mounts {
 		id := m.BlockDeviceID
@@ -1629,7 +1636,12 @@ func (k *kataAgent) handleBlkOCIMounts(c *Container, spec *specs.Spec) ([]*grpc.
 		// Create Storage structure
 		vol, err := k.createBlkStorageObject(c, m)
 		if vol == nil || err != nil {
-			return nil, err
+			return nil, nil, err
+		}
+
+		if HasOption(m.Options, vcAnnotations.IsFileSystemLayer) {
+			layerStorages = append(layerStorages, vol)
+			continue
 		}
 
 		// Each device will be mounted at a unique location within the VM only once. Mounting
@@ -1660,7 +1672,7 @@ func (k *kataAgent) handleBlkOCIMounts(c *Container, spec *specs.Spec) ([]*grpc.
 		volumeStorages = append(volumeStorages, vol)
 	}
 
-	return volumeStorages, nil
+	return layerStorages, volumeStorages, nil
 }
 
 // handlePidNamespace checks if Pid namespace for a container needs to be shared with its sandbox
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -1662,6 +1662,10 @@ func (k *kataAgent) handleBlkOCIMounts(c *Container, spec *specs.Spec) ([]*grpc.
 				"new-source":      path,
 			}).Debug("Replacing OCI mount source")
 			spec.Mounts[idx].Source = path
+			if HasOption(spec.Mounts[idx].Options, vcAnnotations.IsFileBlockDevice) {
+				// The device is already mounted, just bind to path in container.
+				spec.Mounts[idx].Options = []string{"bind"}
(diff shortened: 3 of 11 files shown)


wedsonaf:
Updated commit messages to mention https://github.com/confidential-containers/community/issues/66

wedsonaf:
Removed usage of `strings.CutPrefix` as it's only available in go 1.20+.

fitzthum:
@wedsonaf how does this relate to https://github.com/confidential-containers/documentation/issues/125?

wedsonaf:
> @wedsonaf how does this relate to [confidential-containers/documentation#125](https://github.com/confidential-containers/documentation/issues/125)?

It's an implementation of the design documented there.

That design is the same as what I proposed in January, except that they use a modified version of erofs (to support 512-byte data alignment, which is what tar files offer) for the file system and Nydus for the snapshotter. (See [this slide](https://docs.google.com/presentation/d/1PU1XizarZt22eiVlQ_ug3bjtViST3qlya9OZQPWl8IM/edit#slide=id.p4) from Gerry's presentation after our first round of discussions.)

The Nydus/erofs combination demoed last week does not satisfy our requirements at MS:
1. It's not ready. Ours is ready with integrity support (with dm-verity root hashes enforced by policy). This is the first PR; dm-verity support is a smaller PR than this one (136 lines added to kata-agent).
2. Dependency on virtiofs: neither tdx- nor sev-snp-enabled guest kernels support virtiofs. So we can't really depend on it. Our implementation uses block devices, which works with tdx and sev-snp-enabled guests.
3. Performance: based on what I could see in the recording of last week's demo, images are shared via virtiofs with the guest, then a loop device (/dev/loop6 in the demo) is created, then erofs is mounted on top of that. It gets even worse if we have to use 9p instead of virtiofs. That's a lot of (arguably unnecessary) components in the data path; in our implementation in this PR, tarfs is mounted directly on the block device.
4. The modified erofs implementation is in kernel 6.4, which isn't even released yet ([currently in rc6](https://lore.kernel.org/lkml/CAHk-=wgmbjQu3bVSa5JJhEymeQhkr7N3cjg-CoNLVituUnfW8w@mail.gmail.com/)). The first LTS version is expected towards the end of the year, even then it can't be immediately used because CSPs and distros need to test first. Given that CSPs don't want to touch non-LTS kernels, this would be available for wide usage only next year. At MS we currently use 5.15. In any case, our tarfs module works on 5.15+ kernels (haven't really tested earlier kernels, but it doesn't depend on any fancy features, so I expect it to work) and is quite simple: we actually _want_ small and simple components in our TCB. For comparison, erofs is 9,040 LoC in 6.4 while tarfs is 664 LoC -- this is mostly due to features that we don't or can't use.
5. Dependency on Nydus: nydus is great but it's also big and complicated, it has 20,757 LoC in non-test files and is written in golang. Our tardev-snapshotter is written in Rust (which generally consumes less memory and is more efficient because it doesn't have a GC) and has 1,095 LoC -- it is much simpler. While it's not part of the confidential computing TCB, it is in the control plane and interacts with external untrusted components (e.g., container registries) so it's also preferable to have a reduced attack surface.

I'd also like to mention that we want containers that have common layers to be able to share them when running on the same pod. So we share individual layers as block devices. It's unclear to me what Nydus does: in the demo I didn't see individual layers. Anyway, sharing the layers separately also allows us to have combinations of encrypted/unencrypted layers -- this isn't possible if we squash all the layers.

Lastly, since our implementation doesn't depend on image-rs, it allows us to shave 8.6MB off our kata-agent binary (when pulling from the guest is not needed). The sizes of kata-agent are as follows for a release build: 31.9MB with image-rs, 23.3MB without image-rs. So we save on guest memory and reduce the amount of code in our TCB.

Having said all of the above, this PR does not preclude nydus/erofs, quite the contrary: it adds generic support for block-based layers to kata, which can be leveraged by other snapshotter/filesystems. It also doesn't touch any of the existing guest-pulling code -- it remains available for use if desired. This PR is also compatible with a proposal I posted on slack for supporting peer pods with a different snapshotter (that downloads layers to CSP disks) and a remote hypervisor that knows how to hot-add CSP disks to peer VMs.

This PR introduces 151 new lines of code, so it's quite simple and low risk.

It also allows us to have a working and simple image management implementation that works without a forked containerd in CC 0.7.0, and that can be upstreamed to main kata.

wedsonaf:
Here is a demo of running a pod with tardev-snapshotter and tarfs:
[asciicast](https://asciinema.org/a/r8d9BEU9XnWRSgPZIUD7DVKpe)

hsiangkao:
> 4\. For comparison, erofs is 9,040 LoC in 6.4 while tarfs is 664 LoC -- this is mostly due to features that we don't or can't use.

Personally I have no idea to reponse any of this (I don't think a read-only fs with more useful features are flaw, especially many features can be explicitly disabled at compile time or run time): guess what the current linux kernel LoC is? guess what the 2.6 kernel LoC is? Could you just duplicate all paths for your specific end-to-end use case?

The other points on my side are quite minor.

src/runtime/pkg/govmm/qemu/qmp.go
```diff
@@ -818,7 +818,12 @@ func (q *QMP) blockdevAddBaseArgs(driver string, blockDevice *BlockDevice) map[s
 // used to name the device.  As this identifier will be passed directly to QMP,
 // it must obey QMP's naming rules, e,g., it must start with a letter.
 func (q *QMP) ExecuteBlockdevAdd(ctx context.Context, blockDevice *BlockDevice) error {
-	args := q.blockdevAddBaseArgs("host_device", blockDevice)
+	var args map[string]interface{}
+	if fi, err := os.Stat(blockDevice.File); err == nil && fi.Mode().IsRegular() {
+		args = q.blockdevAddBaseArgs("file", blockDevice)
+	} else {
+		args = q.blockdevAddBaseArgs("device", blockDevice)
```
jepio, line 825: does qemu accept this? according to qemu protocol docs it really should be "host_device".

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1304,11 +1306,12 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
 
 	// Block based volumes will require some adjustments in the OCI spec, and creation of
 	// storage objects to pass to the agent.
-	volumeStorages, err := k.handleBlkOCIMounts(c, ociSpec)
+	layerStorages, volumeStorages, err := k.handleBlkOCIMounts(c, ociSpec)
```
jepio, line 1309: why split out the return into two slices when both of them are appended to ctrStorages anyway?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1304,11 +1306,12 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
...
 	if err != nil {
 		return nil, err
 	}
 
+	ctrStorages = append(layerStorages, ctrStorages...)
```
jepio, line 1314: why are the args flipped compared to the usual convention - is the order somehow important?

jepio:
This should target `main` instead of `CCv0`.

Do you have an example manifest that can be used to exercise the code paths in this PR?

jepio:
As far as I can tell is not so much "image-pulling-on-host" as it is: "mount additional volumes from the host as block devices into the guest based on annotations".

So this is not coupled to any specific snapshotter and would be used in https://github.com/confidential-containers/documentation/issues/125 as well, because some of the modes rely on virtio-blk and not just virtiofs.

src/runtime/pkg/govmm/qemu/qmp.go
wedsonaf: Oh, this should be `host_device`. Thanks for catching this, will fix.

src/runtime/virtcontainers/kata_agent.go
wedsonaf: The layers must come before the rootfs because the rootfs is an overlay of the layers (otherwise mounting the rootfs fails). `handleBlkOCIMounts` returns the layers separately so that we can *prepend* them to `ctrStorages`, which makes them be mounted before the rootfs (which is the first entry in `ctrStorages`).

src/runtime/virtcontainers/kata_agent.go
wedsonaf: Because we want to *prepend* to `ctrStorages`. See my previous comment on why.

wedsonaf (commented):
Thanks for the review @jepio!

src/runtime/virtcontainers/kata_agent.go
jepio: Very subtle - could you document this at line 1314

wedsonaf:
> As far as I can tell this is not so much "image-pulling-on-host" as it is: "mount additional volumes from the host as block devices into the guest based on annotations".

Yes, "image pulling on the host" is the name of the PR only. The 3 commit messages just describe what the commits do.

> So this is not coupled to any specific snapshotter and would be used in [confidential-containers/documentation#125](https://github.com/confidential-containers/documentation/issues/125) as well, because some of the modes rely on virtio-blk and not just virtiofs.

That's right. I try to say this in one of my previous messages:  _this PR does not preclude nydus/erofs, quite the contrary: it adds generic support for block-based layers to kata, which can be leveraged by other snapshotter/filesystems._

src/runtime/virtcontainers/kata_agent.go
wedsonaf: Sure, will do.

jepio:
> Regular manifests would work. To exercise this code path we need to configure containerd to use a snapshotter that returns mounts with these annotations, which is what tardev-snapshotter does.

From the code alone it's hard to figure out example annotations that need to be passed from the snapshotter. I'm not thinking about fully functional usage as container layers, but a manifest that results in host files showing up as block volumes in the guest to play around with.

src/runtime/virtcontainers/kata_agent.go
wedsonaf: Done in the latest version.

src/runtime/pkg/govmm/qemu/qmp.go
wedsonaf: Fixed in the latest version.

src/agent/src/mount.rs
```diff
@@ -309,8 +309,33 @@ pub async fn update_ephemeral_mounts(
...
+        .options
+        .iter()
+        .any(|e| e == "io.katacontainers.fs-opt.overlay-rw")
+    {
+        let cid = cid.ok_or_else(|| anyhow!("No container id in rw overlay"))?;
+        let cpath = Path::new(crate::rpc::CONTAINER_BASE).join(cid);
+        let work = cpath.join("work");
+        let upper = cpath.join("upper");
+
+        fs::create_dir_all(&work)?;
```
c3d, line 325: Maybe add a `.context("Creating work directory")` and similarly for next line, since there are multiple places that can fail in this function

wedsonaf:
> > Regular manifests would work. To exercise this code path we need to configure containerd to use a snapshotter that returns mounts with these annotations, which is what tardev-snapshotter does.
> 
> From the code alone it's hard to figure out example annotations that need to be passed from the snapshotter. I'm not thinking about fully functional usage as container layers, but a manifest that results in host files showing up as block volumes in the guest to play around with.

I see. I've been using `crictl` to test this out. The [CRI api](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L212) doesn't support specifying mount options, so there is no obvious way to do this with `crictl` without a snapshotter.

I'm going to try out other tools to see if I can find one that allows me to specify mount options in a manifest or command line.

Meanwhile, you can find the annotations the snapshotter uses here: https://github.com/kata-containers/tardev-snapshotter/blob/main/snapshotter/src/snapshotter.rs#L121

src/agent/src/mount.rs
wedsonaf: Good idea. Will do.

wedsonaf:
@jepio -- I just added a 4-line commit to this PR that enables the following without a snapshotter:

```
$ sudo ./ctr run --rm -t --runtime io.containerd.kata.v2 --mount type=ext4,src=$PWD/disk.img,dst=/example,options=ro:io.katacontainers.fs-opt.block_device=file docker.io/library/busybox:latest example
/ # mount | grep example
/dev/vda on /example type ext4 (ro,relatime)
/ # ls /example
hello.txt   lost+found
/ #
```

Note that `$PWD/disk.img` is an ext4fs image I created previously. By adding the `io.katacontainers.fs-opt.block_device=file` option to the mount, the file is added as block device and is mounted into the pod with the right fs. Then it is bind-mounted into the container as you can see in the example above.

jepio:
@wedsonaf the "Fixes:" check requires an issue in this repo, or an issue name that  starts with`github.com/kata-containers/...`. Can you create an issue and update the commit messages?

ChengyuZhu6:
Hi, @wedsonaf . Maybe you could reserve some interfaces for image-rs  to support image-rs processing image blocks with encryption and integrity verification?

ChengyuZhu6:
@wedsonaf This is the implementation of image block integrity check and mount in image-rs.  It can support all mount types, not just erofs tarfs.

wedsonaf:
Rebased to `main` after speaking to @fidencio

ChengyuZhu6:
hi @wedsonaf . It's the roadmap of image pulling with verity on the host. https://github.com/confidential-containers/confidential-containers/issues/142.
maybe we can discuss this ?

fidencio:
I sincerely think this one is good to go as it's and it'll unblock us.

Very compotent people (@wedsonaf, @ChengyuZhu6, @jiangliu), way more compotent than I ever will be, are discussing how to build a solution atop of this, but I think this very first step can be unblocked.

There's an ongoing discussion on standardising how we pass the mount options from the runtime to the agent, and that discussion is happening as part of https://github.com/confidential-containers/confidential-containers/issues/142.

Regardless of my opinions on how that should be implemented, which has close to zero value considering that I'm not as familiar with this as the folks actively working on this, I that can be refactored in the coming weeks as part of the follow-up work.

I just cut Kata Containers 3.2.0-rc0, meaning that this work will **NOT** be part of an official release till 3.3.0, in a quarter from now, and we have that time in our favour to have that sorted out.

jiangliu (approved):

fidencio:
> We do have different opinions about the message format

I'd like to have an explicit agreement from @wedsonaf, **before having this one merged**, that he agrees on continuing with the discussion and, **if needed**, adapting the code to the consensus reached in the coming discussions.

ChengyuZhu6:
Thanks! @wedsonaf

wedsonaf:
> > We do have different opinions about the message format
> 
> I'd like to have an explicit agreement from @wedsonaf, **before having this one merged**, that he agrees on continuing with the discussion and, **if needed**, adapting the code to the consensus reached in the coming discussions.

Yes, I'm happy to update this as we (the community) progress in our discussions.

fidencio:
@wedsonaf, 

@wedsonaf, as our static checks are looking for an issue here, I've "cloned" the issue from the CoCo side to here .

With this in mind, please, change the reference in the commit messages from:
```
Fixes: confidential-containers/community#66
```

to:
```
Fixes: #7536
```

I'd have done it myself, but I have no authorisation to push to your branch. :-)

fidencio (approved):
lgtm, thanks @wedsonaf!

Issue #7108, opened 2023-06-14
migrate tests/lib/common.sh functions to kata-containers/tests/common.sh


dborquez:
As part of the effort to migrate our CI metrics to utlize Github workflows, we need to simplify and merge the script `tests/lib/common.sh` script to this location `kata-containers/tests/common.sh`.

PR #7109, opened 2023-06-14
tests: Move tests helper script to this repo

dborquez:
The common.sh script includes helper functions used in our metrics tests, so we are gradually adding more metrics used in kata.

Fixes: #7108

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -7,6 +7,22 @@
 # This file contains common functions that
 # are being used by our metrics and integration tests
 
+# Kata tests directory used for storing various test-related artifacts.
+KATA_TESTS_BASEDIR="${KATA_TESTS_BASEDIR:-/var/log/kata-tests}"
+
+# Directory that can be used for storing test logs.
+KATA_TESTS_LOGDIR="${KATA_TESTS_LOGDIR:-${KATA_TESTS_BASEDIR}/logs}"
+
+# Directory that can be used for storing test data.
+KATA_TESTS_DATADIR="${KATA_TESTS_DATADIR:-${KATA_TESTS_BASEDIR}/data}"
+
+# Directory that can be used for storing cache kata components
+KATA_TESTS_CACHEDIR="${KATA_TESTS_CACHEDIR:-${KATA_TESTS_BASEDIR}/cache}"
+
+KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
+
+RUNTIME="${RUNTIME:-containerd-shim-kata-v2}"
+
 die() {
 	local msg="$*"
 	echo -e "[$(basename $0):${BASH_LINENO[0]}] ERROR: $msg" >&2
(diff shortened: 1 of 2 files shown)


GabyCT (approved):

tests/common.bash
```diff
@@ -45,3 +64,160 @@ waitForProcess() {
...
+	if [ "${KATA_HYPERVISOR}" != "dragonball" ]; then
+		HYPERVISOR_VERSION=$(${HYPERVISOR_PATH} --version | head -n1)
+	fi
+	VIRTIOFSD_PATH=$(kata-runtime kata-env --json | jq -r .Hypervisor.VirtioFSDaemon)
+
+	INITRD_PATH=$(kata-runtime kata-env --json | jq -r .Initrd.Path)
+}
+
+# Checks that processes are not running
+check_processes() {
```
fidencio, line 103: Where is it going to be used?

tests/common.bash
```diff
@@ -45,3 +64,160 @@ waitForProcess() {
...
+	done
+
+	count_tasks="$(sudo ctr t list -q | wc -l)"
+
+	if (( count_tasks > 0 )); then
+		die "Can't remove running contaienrs."
+	fi
+}
+
+restart_containerd_service() {
```
fidencio, line 191: I imagine this will be used by the metrics tests, right?

tests/common.bash
```diff
@@ -45,3 +64,160 @@ waitForProcess() {
...
+		((counter++))
+	done
+
+	[ "$counter" -ge "$retries" ] && { warn "Can't connect to containerd socket"; return 1; }
+
+	clean_env_ctr
+	return 0
+}
+
+restart_docker_service() {
```
fidencio, line 208: Same about this one.

fidencio (changes requested):
I've raised a bunch of questions, just to make sure we're not adding anything else but what we need, and that whatever we're adding is actually present.

tests/common.bash
dborquez: The following metrics make use of check_processes():
- disk/cassandra_kubernetes/cassandra.sh
- storage/fio-k8s/fio-test-ci.sh
- network/iperf3_kubernetes/k8s-network-metrics-iperf3.sh
- density/k8s-sysbench.sh

tests/common.bash
```diff
@@ -45,3 +64,160 @@ waitForProcess() {
...
+	restart_systemd_service_with_no_burst_limit docker || return 1
```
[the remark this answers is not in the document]
dborquez: I'll remove that one, as docker isn't currently used in our metrics

tests/common.bash
dborquez: yes, it is used by:
- metrics/density/fast_footprint.sh
- metrics/density/footprint_data.sh
- metrics/storage/webtooling.sh

cmaf (commented):
@dborquez the intention is to move this and then the tests that use it in another PR?

dborquez:
@cmaf yes, we are moving gradually the tests repo, so first we are moving the common shared scripts and then we are moving small groups of tests.

tests/common.bash
```diff
@@ -45,3 +64,143 @@ waitForProcess() {
...
+
+	count_tasks="$(sudo ctr t list -q | wc -l)"
+
+	if (( count_tasks > 0 )); then
+		die "Can't remove running contaienrs."
+	fi
+}
+
+restart_containerd_service() {
+	restart_systemd_service_with_no_burst_limit containerd || return 1
```
fidencio, line 192: This function, from common.sh, should also be added here, no?
Otherwise calling restart_containerd_service will just fail as the function above is not defined anywhere.

fidencio: Just to be super clear, from the tests repo common.sh file.

fidencio: https://github.com/kata-containers/tests/blob/main/lib/common.bash#L272-L309

tests/common.bash
dborquez: that's correct, let me fix it

tests/common.bash
dborquez: @fidencio yeah, good catch again

tests/common.bash
```diff
@@ -7,6 +7,25 @@
 # This file contains common functions that
 # are being used by our metrics and integration tests
 
+_script_dir=$(dirname "${BASH_SOURCE[0]}")
+source "${_script_dir}/error.sh"
```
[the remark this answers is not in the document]
dborquez: that's correct, thanks @fidencio

fidencio (approved):
lgtm, thanks @dborquez!

PR #7110, opened 2023-06-15
WIP: Add attestation support to main

fitzthum:
This is not ready yet, but I will put it out here to avoid any conflicts.

- [x] Add support for AA to rootfs builder (may need cleanup)
- [x] Add KBC for SEV/SNP initrd
- [ ] Add KBC for other TEEs
- [x] Add AA_KBC_PARAMS to config
- [x] Start the AA via the Kata Agent (early on) - not sure about placement
- [x] Add a test -- not sure if this works yet

See thread below for updates.

I have a pretty interesting plan for a test. Hopefully I will be able to push that tomorrow.

cc: @ryansavino @stevenhorsman

### DIFF
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -25,6 +25,7 @@ const LOG_VPORT_OPTION: &str = "agent.log_vport";
 const CONTAINER_PIPE_SIZE_OPTION: &str = "agent.container_pipe_size";
 const UNIFIED_CGROUP_HIERARCHY_OPTION: &str = "agent.unified_cgroup_hierarchy";
 const CONFIG_FILE: &str = "agent.config_file";
+const AA_KBC_PARAMS: &str = "agent.aa_kbc_params";
 
 const DEFAULT_LOG_LEVEL: slog::Level = slog::Level::Info;
 const DEFAULT_HOTPLUG_TIMEOUT: time::Duration = time::Duration::from_secs(3);
@@ -77,6 +78,7 @@ pub struct AgentConfig {
     pub tracing: bool,
     pub endpoints: AgentEndpoints,
     pub supports_seccomp: bool,
+    pub aa_kbc_params: String,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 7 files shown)


fidencio:
JFYI, I've gone through the forward port, and it does look good so far.
I'll come back for a full review once it's ready (not in a WIP state).

fitzthum:
This is still in progress btw. Running into a lot of snags with the tests but hopefully will be ready soon.

fitzthum:
Ok, I have pushed a first version of all the components.

Some notes:

- Currently this only has a chance to pass for SNP. For SEV, we need some chances to the shim to support attestation. @ryansavino is working on them. Once they are complete I can add to the test to support SEV(-ES). The test will currently run for TDX, but I am not sure if the TDX image has the necessary artifacts (i.e. the AA) for attestation to work.
- I am starting the KBS/AS as part of the test. In the past @fidencio has discussed decoupling the KBS from the Kata CI by running it in the background or doing something else. Since I don't have access to any of the CI nodes, I am not able to develop the test with this approach. We can discuss the optimal approach.
- The Azure SNP verifier can currently only be built with ubuntu > 22.04 due to a dependency on `libtss2-dev`. As written the KBS will fail to build on the SNP CI node. There are a few solutions to this.
   - Upgrade CI node to 22.04. We should do this regardless of next steps
   - Use containerized KBS
   - Update features of KBS so we can skip building the AZ-vTPM verifier, which can only be used with peer pods currently. We could make a feature for all the baremetal verifiers and one for all the peer pods verifiers, for instance.
- There are some functions in the test file that should be moved to a library file. I am not sure which of the multiple library files is appropriate. Since this test is currently the only code using these functions I have left them in place for now, but they should probably move elsewhere.
- This PR modifies the kernel command line to set the AA_KBC_PARAMS. This should be done via annotation, but that is not yet supported in main. I currently don't restore the kernel params, which might sabotage other tests.
- I picked a random location to start the AA. I welcome any suggestions about the best place for this in the Kata Agent.
- I was planning to use a generic `grpcurl` image for testing, but this has some problems. Instead I created my own testing image with `grpcurl` inside of it. I may update this to have a ttrpc client at some point. Right now this PR switches to gRPC in main because of https://github.com/confidential-containers/guest-components/issues/171, but this issue may be fixed soon.

fitzthum:
Can someone /test this

Might change approach a bit given the incoming CDH, but curious to see if this actually works.

PR #7120, opened 2023-06-16
runtime-rs: change block index to 0

### LINKED ISSUE #7119: runtime-rs: device manager block index incorrect
In device_manager.rs, block index is set to 1 which generally corresponds to vdb per the notes in the code. Is there a reason for choosing this over 0/vda?

@jodh-intel could you describe the problem you were seeing with this?

cmaf:
Change block index in SharedInfo to 0 for vda.

Fixes #7119

In device_manager.rs, block index is set to 1 which generally corresponds to vdb per the notes in the code. Is there a reason for choosing this over 0/vda?

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
@@ -26,7 +26,7 @@ const DEVICE_TYPE_BLOCK: &str = "b";
 /// block_index and released_block_index are used to search an available block index
 /// in Sandbox.
 ///
-/// @block_index generally default is 1 for <vdb>;
+/// @block_index generally default is 0 for <vda>;
 /// @released_block_index for blk devices removed and indexes will released at the same time.
 #[derive(Clone, Debug, Default)]
 struct SharedInfo {
@@ -37,7 +37,7 @@ struct SharedInfo {
 impl SharedInfo {
     fn new() -> Self {
         SharedInfo {
-            block_index: 1,
+            block_index: 0,
             released_block_index: vec![],
         }
     }


jodh-intel (approved):
Thanks @cmaf.

lgtm

dborquez (approved):

Tim-0731-Hzt:
Thanks @cmaf, yes, it's a mistake, we should set the block index to start from 0, but I am afraid you cannot only change the index to 0. This is because if you want to insert a block volume after the VM start, the current `block_index` is 0, it will be `vda` again, but we hope it is `vdb` at this stage. So, I think you might need to consider using `device manager to manage the index here, like increasing the `block_index` when set up the VM rootfs.

jodh-intel:
Hi @Tim-0731-Hzt - yep, we've realised this is too simplistic. I'm still trying to digest all the different layers in the code but it looks initially as though the hypervisor would need to tell the `DeviceManager` that it has already consumed a block index somehow. The `DeviceManager` cannot know without asking how the hypervisor plans to use the files specified in the `HypervisorConfig`. So, fwics, either the hypervisor would need to request permission from the `DeviceManager` to use >0 initial indices (starting from zero), or the hypervisor would need to be able to advertise to the `DeviceManager` if it has already consumed >0 block indices.

Tim-0731-Hzt:
> Hi @Tim-0731-Hzt - yep, we've realised this is too simplistic. I'm still trying to digest all the different layers in the code but it looks initially as though the hypervisor would need to tell the `DeviceManager` that it has already consumed a block index somehow. The `DeviceManager` cannot know without asking how the hypervisor plans to use the files specified in the `HypervisorConfig`. So, fwics, either the hypervisor would need to request permission from the `DeviceManager` to use >0 initial indices (starting from zero), or the hypervisor would need to be able to advertise to the `DeviceManager` if it has already consumed >0 block indices.

Hi @jodh-intel @cmaf , I can help with this. I have raised another PR, which use device manager to handle this, please take a look https://github.com/kata-containers/kata-containers/pull/7139

jodh-intel:
Closing in favour of #7139.

PR #7139, opened 2023-06-20
runtime-rs: change block index to 0

Tim-0731-Hzt:
1. change the block index of the device manager to start from 0, then the VM rootfs device will start from `vda` 
2. use the device manager to handle the VM rootfs

cc @cmaf @jodh-intel


jodh-intel (approved):
Thanks @Tim-0731-Hzt!

lgtm

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
```diff
@@ -310,7 +310,7 @@ mod tests {
                     ]
                     .to_vec(),
                 },
-                result: Err(anyhow!("Unsupported rootfs driver")),
+                result: Err(anyhow!("Unsupported rootfs driver foo")),
```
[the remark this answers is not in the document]
Tim-0731-Hzt: This is added in the unit test, if the incorrect rootfs device type is used, this is the expected output. 
As I  return the incorrect rootfs driver type here.
`return Err(anyhow!("Unsupported rootfs driver {}", rootfs_driver));`

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
```diff
@@ -324,15 +324,14 @@ mod tests {
                     ]
                     .to_vec(),
                 },
-                result: Err(anyhow!("Unsupported rootfs type")),
+                result: Err(anyhow!("Unsupported rootfs type foo")),
```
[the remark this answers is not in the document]
Tim-0731-Hzt: same as above

GabyCT (approved):

Tim-0731-Hzt:
> @Tim-0731-Hzt - hmm. I think there is still a problem here. When I run with ctr and devmapper...
> 
> ```shell
> $ sudo ctr run --runtime "io.containerd.kata.v2" --snapshotter devmapper --rm -t quay.io/prometheus/busybox:latest foo-dm
> ```
> 
> ..., there will be 2 block devices (VM rootfs and container rootfs). However, the hypervisor's `add_device()` method gets called in the wrong order:
> 
> * 1st call: `/dev/vda` for the **container** rootfs (`busybox` in my example above).
> * 2nd call: for `/dev/vdb` for the **VM** rootfs (`kata-containers.img`).

@jodh-intel , I tried this command, this is what I get, it seems the output is expected. We can see `vda1` has mount point `/`  `vdb` has mount point `/run/kata-containers/foo-dm/rootfs`
```shell
root@dragonX:~/github/kata-containers/src/runtime-rs# kata-runtime  exec foo-dm
root@localhost:/# 
root@localhost:/# 
root@localhost:/# 
root@localhost:/# 
root@localhost:/# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
vda    254:0    0  128M  1 disk 
└─vda1 254:1    0  124M  1 part /
vdb    254:16   0   10G  0 disk /run/kata-containers/foo-dm/rootfs
root@localhost:/# 
```
I think it's not quite possible to have the problem you mentioned, as we create containers after VM starts successfully. If the VM start, the VM rootfs must be ready, it cannot be `vdb`

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -260,37 +243,6 @@ impl DragonballInner {
             .context("put boot source")
     }
 
-    fn set_vm_rootfs(&mut self, path: &str, driver: &str) -> Result<()> {
-        info!(sl!(), "set vm rootfs {} {}", path, driver);
-        let jail_drive = self
-            .get_resource(path, DRAGONBALL_ROOT_FS)
-            .context("get resource")?;
-
-        if driver == VM_ROOTFS_DRIVER_BLK || driver == VM_ROOTFS_DRIVER_MMIO {
```
cmaf, line 269: @Tim-0731-Hzt does this check for rootfs driver happen elsewhere?

Tim-0731-Hzt: No, I realized this is a duplicate check, as we will check it in `new_rootfs_kernel_params`, so I just remove it.

jodh-intel (changes requested):
Hi @Tim-0731-Hzt - I took another look at this today and I can see what's going wrong for me (with CH).

| DM? | TDX `[*`] | VM rootfs | container rootfs | Works? |
|-|-|-|-|-|
| no | no | non-virtio-block device (pmem block device) | virtiofs device | yes |
| yes | no | non-virtio-block device (pmem block device) | `/dev/vda` (index 0) | **no** |
| yes | yes | `/dev/vda` (index 0) | `/dev/vdb` (index 1) | yes |

`[*]` - requires that the VM rootfs uses a virtio-blk device.

As the tables shows, the problem is that when I run using DM (`ctr run --snapshotter devmapper`), the container rootfs is passed as a block device. But the changes on this PR give that block device index 1 which gets converted into `/dev/vdb`. But that's incorrect for CH, since although it uses a block device (pmem), it's not a virtio-blk one, so the index should be zero (to allow the container rootfs to be mounted as `/dev/vda`).

So, I think we need a way for the hypervisor to tell runtime-rs whether it plans to use present the VM rootfs as a block device. If yes, the logic on this PR is fine, but if not, the index needs to be decremented by 1.

Tim-0731-Hzt:
> Hi @Tim-0731-Hzt - I took another look at this today and I can see what's going wrong for me (with CH).
> 
> DM?	TDX `[*`]	VM rootfs	container rootfs	Works?
> no	no	non-virtio-block device (pmem block device)	virtiofs device	yes
> yes	no	non-virtio-block device (pmem block device)	`/dev/vda` (index 0)	**no**
> yes	yes	`/dev/vda` (index 0)	`/dev/vdb` (index 1)	yes
> `[*]` - requires that the VM rootfs uses a virtio-blk device.
> 
> As the tables shows, the problem is that when I run using DM (`ctr run --snapshotter devmapper`), the container rootfs is passed as a block device. But the changes on this PR give that block device index 1 which gets converted into `/dev/vdb`. But that's incorrect for CH, since although it uses a block device (pmem), it's not a virtio-blk one, so the index should be zero (to allow the container rootfs to be mounted as `/dev/vda`).
> 
> So, I think we need a way for the hypervisor to tell runtime-rs whether it plans to use present the VM rootfs as a block device. If yes, the logic on this PR is fine, but if not, the index needs to be decremented by 1.

@jodh-intel, PTAL,  I have added a `pmem_index` for device manager, if the hypervisor is using a pmem device as rootfs, device manager will increase `pmem_index`. In this way, the container rootfs could be `/dev/vda`

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -254,14 +263,21 @@ impl DeviceManager {
...
+        if block_driver == *VM_ROOTFS_DRIVER_PMEM {
+            block_config.index = self.shared_info.get_and_set_pmem_index();
+            block_config.virt_path = format!("pmem{}", block_config.index);
+        } else {
+            // generate block device index and virt path
+            // safe here, Block device always has virt_path.
+            if let Some(virt_path) = self.get_dev_virt_path(DEVICE_TYPE_BLOCK)? {
+                block_config.index = virt_path.0;
+                block_config.virt_path = virt_path.1;
+            }
```
Apokleos, line 280: For case PMEM, a special block device case, we should handle it in a generic way as much as possible. And the index and virt_path are handled in the  get_dev_virt_path is better or maybe some other solutions.

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -25,19 +25,22 @@ pub type ArcMutexDevice = Arc<Mutex<dyn Device>>;
...
 ///
-/// @block_index generally default is 1 for <vdb>;
+/// @block_index generally default is 0 for <vda>;
 /// @released_block_index for blk devices removed and indexes will released at the same time.
+/// @pmem_index is to indicate the index of the pmem device
 #[derive(Clone, Debug, Default)]
 struct SharedInfo {
     block_index: u64,
     released_block_index: Vec<u64>,
+    pmem_index: u64,
```
Apokleos, line 35: We need carefully  to consider whether to add a `pmem_index` field like this.  And, if not so, how to handle the block_index to satisfy the pmem index requirement.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -162,6 +174,29 @@ impl VirtSandbox {
...
+                Err(anyhow!("failed to get image"))
+            }
+        }
+        .context("get image")?;
+
+        Ok(BlockConfig {
+            path_on_host: image,
+            is_readonly: true,
+            ..Default::default()
+        })
```
Apokleos, line 197: Does it need to specify the vm rootfs driver?

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
Tim-0731-Hzt: thx, fixed

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -22,57 +23,66 @@ use super::{
...
+        declare_index!(self, pmem_index, released_pmem_index)
+    }
 
-        Ok(current_index)
+    fn declare_block_index(&mut self) -> Result<u64> {
+        declare_index!(self, block_index, released_block_index)
+    }
+
+    fn release_pmem_index(&mut self, index: u64) {
+        release_index!(self, index, released_pmem_index);
```
Apokleos, line 81: IMO, the `declare_device_index` and `release_device_index` should be reserved and add an argument to  differentiate the NVDIMM type and BLOCK type, If so, dup codes will be reduced.

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -106,19 +124,23 @@ impl DeviceManager {
         if let Err(e) = result {
             match device_guard.get_device_info().await {
                 DeviceType::Block(device) => {
-                    self.shared_info.release_device_index(device.config.index);
+                    if device.config.driver_option == *KATA_NVDIMM_DEV_TYPE {
+                        self.shared_info.release_pmem_index(device.config.index)
+                    } else {
+                        self.shared_info.release_block_index(device.config.index);
+                    }
```
Apokleos, line 131: the same as L72-L81

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -202,13 +224,22 @@ impl DeviceManager {
...
+    ) -> Result<Option<(u64, String)>> {
         let virt_path = if dev_type == DEVICE_TYPE_BLOCK {
-            // generate virt path
-            let current_index = self.shared_info.declare_device_index()?;
-            let drive_name = get_virt_drive_name(current_index as i32)?;
-            let virt_path_name = format!("/dev/{}", drive_name);
-
+            let (current_index, virt_path_name) = if is_pmem {
+                let current_index = self.shared_info.declare_pmem_index()?;
+                let virt_path_name = format!("/pmem/{}", current_index);
```
Apokleos, line 235: It should be `format!("/dev/pmem{}", current_index);`

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -282,12 +312,28 @@ impl DeviceManager {
...
+            // convert the block driver to kata type
+            VIRTIO_BLOCK_MMIO => {
+                vhu_blk_config.driver_option = KATA_MMIO_BLK_DEV_TYPE.to_string();
+            }
+            VIRTIO_BLOCK_PCI => {
+                vhu_blk_config.driver_option = KATA_BLK_DEV_TYPE.to_string();
+            }
+            _ => {
+                return Err(anyhow!(
+                    "invliad driver type {}",
```
Apokleos, line 328: Sorry, the "invalid driver type" -> "unsupported driver type" is better.

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -304,10 +350,30 @@ impl DeviceManager {
...
+            VIRTIO_BLOCK_PCI => {
+                block_config.driver_option = KATA_BLK_DEV_TYPE.to_string();
+            }
+            VIRTIO_PMEM => {
+                block_config.driver_option = KATA_NVDIMM_DEV_TYPE.to_string();
+                is_pmem = true;
+            }
+            _ => {
+                return Err(anyhow!(
+                    "invliad driver type {}",
```
Apokleos, line 369: "invalid driver type" -> "unsupported driver type" is better.

Apokleos (commented):
Thx @Tim-0731-Hzt , few comments!

Apokleos (commented):
Thx @Tim-0731-Hzt , few comments!

Apokleos (commented):
Thx @Tim-0731-Hzt , few comments!

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
Tim-0731-Hzt: thx, fixed

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
Tim-0731-Hzt: fixed

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
Tim-0731-Hzt: fixed

Apokleos (approved):
LGTM ！

jodh-intel:
Hi @Tim-0731-Hzt - Thanks for updating this. I'm not sure the `pmem_index` idea will work though because for CH with TDX, the user should only need to set `confidential_guest=true` to enable a (TDX) confidential guest. However, in that scenario, we have to change the default VM rootfs config from pmem to virtio-blk. This happens "silently". In summary, the hypervisor driver needs to make that decision and runtime-rs cannot (should not) know about it as we don't want to pollute runtime-rs with details of how the hypervisors work. That's why I was thinking we could add a new hypervisor API to indictate to runtime-rs that the hypervisor intends to use a block device for the VM rootfs. That design would mean that runtime-rs would be informed (and could therefore calculate the block index correctly), but it would not need to know about the decisions the hypervisor driver is making.

Apokleos:
Hi @jodh-intel , Thx, I have some questions about this.
> Hi @Tim-0731-Hzt - Thanks for updating this. I'm not sure the `pmem_index` idea will work though because for CH with TDX, the user should only need to set `confidential_guest=true` to enable a (TDX) confidential guest. 
However, in that scenario, we have to change the default VM rootfs config from pmem to virtio-blk. This happens "silently". In summary, the hypervisor driver needs to make that decision and runtime-rs cannot (should not) know about it as we don't want to pollute runtime-rs with details of how the hypervisors work. That's why I was thinking we could add a new hypervisor API to indictate to runtime-rs that the hypervisor intends to use a block device for the VM rootfs. That design would mean that runtime-rs would be informed (and could therefore calculate the block index correctly), but it would not need to know about the decisions the hypervisor driver is making.
>

1. How to change the default VM rootfs config from pmem to virtio blk? And who will do it ?
2. I want to sync it with you that the idea of adding a new hypervisor API to inform runtime-rs(device manager?) is inserting block device into VM first and then getting its info through the API which includes block device name and index?
3. What's the gap between the pmem handling method in the TDX scenario and  others?

Tim-0731-Hzt:
> Hi @Tim-0731-Hzt - Thanks for updating this. I'm not sure the `pmem_index` idea will work though because for CH with TDX, the user should only need to set `confidential_guest=true` to enable a (TDX) confidential guest. However, in that scenario, we have to change the default VM rootfs config from pmem to virtio-blk. This happens "silently". In summary, the hypervisor driver needs to make that decision and runtime-rs cannot (should not) know about it as we don't want to pollute runtime-rs with details of how the hypervisors work. That's why I was thinking we could add a new hypervisor API to indictate to runtime-rs that the hypervisor intends to use a block device for the VM rootfs. That design would mean that runtime-rs would be informed (and could therefore calculate the block index correctly), but it would not need to know about the decisions the hypervisor driver is making.

The current PR could work with the most general cases, it solves the problem of using different block driver for VM rootfs and containers rootfs/volume. Agree with @Apokleos, As we are not familiar with the TDX, not sure this PR would work for that.

lifupan:
> Hi @Tim-0731-Hzt - Thanks for updating this. I'm not sure the `pmem_index` idea will work though because for CH with TDX, the user should only need to set `confidential_guest=true` to enable a (TDX) confidential guest. However, in that scenario, we have to change the default VM rootfs config from pmem to virtio-blk. This happens "silently". In summary, the hypervisor driver needs to make that decision and runtime-rs cannot (should not) know about it as we don't want to pollute runtime-rs with details of how the hypervisors work. That's why I was thinking we could add a new hypervisor API to indictate to runtime-rs that the hypervisor intends to use a block device for the VM rootfs. That design would mean that runtime-rs would be informed (and could therefore calculate the block index correctly), but it would not need to know about the decisions the hypervisor driver is making.

Hi @jodh-intel 

It seemed the cloud hypervisor wouldn't support pmem for confidential guest, it that true also for other hypervisor such qemu? if yes, I think we can check it in the runtime-rs and force it use virtio-blk even the configuration set it as pmem. 
If it's hypervisor depended, thus I agree to add a new api to indictate to runtime-rs that the hypervisor intends to use a block device for the VM rootfs.

jodh-intel:
hi @lifupan,

> Hi @jodh-intel
> 
> It seemed the cloud hypervisor wouldn't support pmem for confidential guest, it that true also for other hypervisor such qemu?

I think we need input from @pmores and @ananos on that question.

Tim-0731-Hzt:
I think we also need input from @yipengyin

yipengyin:
Thx @Tim-0731-Hzt . 

> This happens "silently". In summary, the hypervisor driver needs to make that decision and runtime-rs cannot (should not) know about it as we don't want to pollute runtime-rs with details of how the hypervisors work. That's why I was thinking we could add a new hypervisor API to indictate to runtime-rs that the hypervisor intends to use a block device for the VM rootfs. 

@jodh-intel @lifupan 
I do agree the runtime-rs should not know about the details of how the hypervisor work. But I suppose change the VM rootfs driver "silently" might cause confusion in practice. Can we just let the runtime-rs add the unsupported device to VM? And the hypervisor throw an error when the device is not supported.

jodh-intel:
@yipengyin - I still think the most general approach is for the hypervisor to tell runtime-rs if it intends to use the VM rootfs as a block device. That avoids having to have special variables like `pmem_index`.

Looking at this again, fwics, the `.set_hypervisor_config()` method should be part of the `Hypervisor` trait for parity with the existing `hypervisor_config()` method. If we add that, but make the setter return a `bool`, that could tell runtime-rs if the hypervisor intends to present the VM rootfs defined in `Hypervisor.boot_info`:

```diff
diff --git a/src/runtime-rs/crates/hypervisor/src/lib.rs b/src/runtime-rs/crates/hypervisor/src/lib.rs
index 2386d0924..9030b63b5 100644
--- a/src/runtime-rs/crates/hypervisor/src/lib.rs
+++ b/src/runtime-rs/crates/hypervisor/src/lib.rs
@@ -88,6 +88,7 @@ pub trait Hypervisor: Send + Sync {
     async fn get_agent_socket(&self) -> Result<String>;
     async fn disconnect(&self);
     async fn hypervisor_config(&self) -> HypervisorConfig;
+    async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool;
     async fn get_thread_ids(&self) -> Result<VcpuThreadIds>;
     async fn get_pids(&self) -> Result<Vec<u32>>;
     async fn get_vmm_master_tid(&self) -> Result<u32>;
```

```rust
impl Hypervisor for CloudHypervisor {
    async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool {
        // ...

        if need_to_specify_vm_rootfs_as_block_device(config) {
            return true;
        }

        return false;
    }
}


```

lifupan:
> @yipengyin - I still think the most general approach is for the hypervisor to tell runtime-rs if it intends to use the VM rootfs as a block device. That avoids having to have special variables like `pmem_index`.
> 
> Looking at this again, fwics, the `.set_hypervisor_config()` method should be part of the `Hypervisor` trait for parity with the existing `hypervisor_config()` method. If we add that, but make the setter return a `bool`, that could tell runtime-rs if the hypervisor intends to present the VM rootfs defined in `Hypervisor.boot_info`:
> 
> ```diff
> diff --git a/src/runtime-rs/crates/hypervisor/src/lib.rs b/src/runtime-rs/crates/hypervisor/src/lib.rs
> index 2386d0924..9030b63b5 100644
> --- a/src/runtime-rs/crates/hypervisor/src/lib.rs
> +++ b/src/runtime-rs/crates/hypervisor/src/lib.rs
> @@ -88,6 +88,7 @@ pub trait Hypervisor: Send + Sync {
>      async fn get_agent_socket(&self) -> Result<String>;
>      async fn disconnect(&self);
>      async fn hypervisor_config(&self) -> HypervisorConfig;
> +    async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool;
>      async fn get_thread_ids(&self) -> Result<VcpuThreadIds>;
>      async fn get_pids(&self) -> Result<Vec<u32>>;
>      async fn get_vmm_master_tid(&self) -> Result<u32>;
> ```
> 
> ```rust
> impl Hypervisor for CloudHypervisor {
>     async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool {
>         // ...
> 
>         if need_to_specify_vm_rootfs_as_block_device(config) {
>             return true;
>         }
> 
>         return false;
>     }
> }
> ```

Hi @Tim-0731-Hzt  @jodh-intel 
If the pmem was only used for guest rootfs, and there was no chance to used for normal block device, I think it's ok to remove the pmem index and just set the block index started from 0 or 1 depended on  set_hypervisor_config return value.

Tim-0731-Hzt:
> > @yipengyin - I still think the most general approach is for the hypervisor to tell runtime-rs if it intends to use the VM rootfs as a block device. That avoids having to have special variables like `pmem_index`.
> > Looking at this again, fwics, the `.set_hypervisor_config()` method should be part of the `Hypervisor` trait for parity with the existing `hypervisor_config()` method. If we add that, but make the setter return a `bool`, that could tell runtime-rs if the hypervisor intends to present the VM rootfs defined in `Hypervisor.boot_info`:
> > ```diff
> > diff --git a/src/runtime-rs/crates/hypervisor/src/lib.rs b/src/runtime-rs/crates/hypervisor/src/lib.rs
> > index 2386d0924..9030b63b5 100644
> > --- a/src/runtime-rs/crates/hypervisor/src/lib.rs
> > +++ b/src/runtime-rs/crates/hypervisor/src/lib.rs
> > @@ -88,6 +88,7 @@ pub trait Hypervisor: Send + Sync {
> >      async fn get_agent_socket(&self) -> Result<String>;
> >      async fn disconnect(&self);
> >      async fn hypervisor_config(&self) -> HypervisorConfig;
> > +    async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool;
> >      async fn get_thread_ids(&self) -> Result<VcpuThreadIds>;
> >      async fn get_pids(&self) -> Result<Vec<u32>>;
> >      async fn get_vmm_master_tid(&self) -> Result<u32>;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```rust
> > impl Hypervisor for CloudHypervisor {
> >     async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool {
> >         // ...
> > 
> >         if need_to_specify_vm_rootfs_as_block_device(config) {
> >             return true;
> >         }
> > 
> >         return false;
> >     }
> > }
> > ```
> 
> Hi @Tim-0731-Hzt @jodh-intel If the pmem was only used for guest rootfs, and there was no chance to used for normal block device, I think it's ok to remove the pmem index and just set the block index started from 0 or 1 depended on set_hypervisor_config return value.

@lifupan I think it could also be used for containers, agent will handle pmem device
```rust
pub async fn add_storages(
    logger: Logger,
    storages: Vec<Storage>,
    sandbox: Arc<Mutex<Sandbox>>,
    cid: Option<String>,
) -> Result<Vec<String>> {
...
DRIVER_NVDIMM_TYPE => nvdimm_storage_handler(&logger, &storage, sandbox.clone()).await,
...
}
```

yipengyin:
> @lifupan I think it could also be used for containers, agent will handle pmem device

I agree with @Tim-0731-Hzt . Although pmem device is only used for guest rootfs currently, we should not break its capability to support container rootfs or volume.

> That avoids having to have special variables like pmem_index.

@jodh-intel I don't like this hardcoded variable `pmem_index` either. However, we might need a mechanism to manage the index for different block driver. Can we introduce some abstraction for block driver @Tim-0731-Hzt ? AFAIK, `virtio-scsi` use another independent index. But I'm not sure.

lifupan:
> > @lifupan I think it could also be used for containers, agent will handle pmem device
> 
> I agree with @Tim-0731-Hzt . Although pmem device is only used for guest rootfs currently, we should not break its capability to support container rootfs or volume.
> 
> > That avoids having to have special variables like pmem_index.
> 
> @jodh-intel I don't like this hardcoded variable `pmem_index` either. However, we might need a mechanism to manage the index for different block driver. Can we introduce some abstraction for block driver @Tim-0731-Hzt ? AFAIK, `virtio-scsi` use another independent index. But I'm not sure.

Actually  we only support one block device driver for other cases such as container rootfs, volumes etc except the guest rootfs. Thus I  it's ok to only keep one block index track, and to determine the init index: if the vm rootfs block driver is the same with volumes/container rootfs, then the block index initiated with "0", otherwise, initiated it with "1"?

yipengyin:
> Actually we only support one block device driver for other cases such as container rootfs, volumes etc except the guest rootfs. 

@lifupan If this assumption hold, I'm OK with the special handling of pmem rootfs driver.

jodh-intel:
CH uses pmem, but I'd still much rather avoid "polluting" runtime-rs with details about pmem. Also, if other versions, technologies appear, we'd need to add things like `pmem_v2_index`, `foo_index`, `bar_index`, _etc_. But all we ultimately care about is the the hypervisor and the runtime deciding on the virtio-blk index, so wouldn't `set_hypervisor_config()` returning a `bool` be a much simpler and cleaner solution?

bergwolf:
Sorry for chiming in late, but why are we depending on the static mapping between device index and `/dev/vda` etc. with runtime-rs+cloud-hypervisor? IIUC, it is a limitation of virtio-mmio but cloud-hypervisor is actually using virtio-pci which should be able to get rid of the static device mapping with pci bus address.

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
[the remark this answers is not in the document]
amshinde: @Tim-0731-Hzt This should include `rootfs_driver` in the returned error and not `foo`. Could you correct this?

src/runtime-rs/crates/hypervisor/src/kernel_param.rs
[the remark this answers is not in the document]
amshinde: Same comment as above - this should include rootfs_driver in the returned error and not foo.

bergwolf:
@jodh-intel 
> CH uses pmem, but I'd still much rather avoid "polluting" runtime-rs with details about pmem. Also, if other versions and technologies appear, we'd need to add things like pmem_v2_index, foo_index, bar_index, etc. But all we ultimately care about is the the hypervisor and the runtime deciding on the virtio-blk index, so wouldn't set_hypervisor_config() returning a bool be a much simpler and cleaner solution?

FWIW, runtime may use pmem for rootfs and volume devices as well, so it is not just about the guest image. So IMO, runtime should know which type of devices are inserted into the guest. If runtime inserts pmem for the rootfs, it should keep track of it with something like `pmem_idx` and the `block_idx` can be left to virtio-blk index only.

jodh-intel:
@ananos, @pmores - do you have any thoughts on this from the FC and QEMU perspectives?

yipengyin:
Hey, guys. @Tim-0731-Hzt @jodh-intel @lifupan @Apokleos @bergwolf 
Can we reach a consensus on following matter first? Otherwise, It seems difficult to reach a conclusion. 
1. Should we treat `virito-pmem` as a general storage backend which might be used as a volume?
2. Can runtime change the vm rootfs driver silently?

In my opinion, 
1. we should treat `virtio-pmem` as a general storage backend. But I'm OK to handle it in specially if it's only used in vm rootfs. 
2. Runtime should throw an error if the vm rootfs driver is not available. Silent change might cause misunderstanding.

Tim-0731-Hzt:
> Hey, guys. @Tim-0731-Hzt @jodh-intel @lifupan @Apokleos @bergwolf Can we reach a consensus on following matter first? Otherwise, It seems difficult to reach a conclusion.
> 
> 1. Should we treat `virito-pmem` as a general storage backend which might be used as a volume?
> 2. Can runtime change the vm rootfs driver silently?
> 
> In my opinion,
> 
> 1. we should treat `virtio-pmem` as a general storage backend. But I'm OK to handle it in specially if it's only used in vm rootfs.
> 2. Runtime should throw an error if the vm rootfs driver is not available. Silent change might cause misunderstanding.

I agree with @yipengyin 
1. I think `virtio-pmem` cannot only be treated as VM rootfs. As I said before, the agent will handle this, so, there must be some use cases for that.
2. I think we need to decide which `crate `to handle the VM rootfs driver. In my current implementation, it is determined when we are loading the config. But we can also implement this in `hypervisor`

ananos:
> @ananos, @pmores - do you have any thoughts on this from the FC and QEMU perspectives?

apologies for the radio silence. 

AFAIK FC does not use the pmem backend, so it does not exhibit the issue you mention @jodh-intel. Otherwise, we would have the same issue I assume. FWIW, the changes in this PR do not affect the FC branch.

Whether it should be the runtime or the hypervisor to determine the blk index I would assume that this should be agnostic to the runtime, as it is hypervisor specific.

amshinde:
> > @yipengyin - I still think the most general approach is for the hypervisor to tell runtime-rs if it intends to use the VM rootfs as a block device. That avoids having to have special variables like `pmem_index`.
> > Looking at this again, fwics, the `.set_hypervisor_config()` method should be part of the `Hypervisor` trait for parity with the existing `hypervisor_config()` method. If we add that, but make the setter return a `bool`, that could tell runtime-rs if the hypervisor intends to present the VM rootfs defined in `Hypervisor.boot_info`:
> > ```diff
> > diff --git a/src/runtime-rs/crates/hypervisor/src/lib.rs b/src/runtime-rs/crates/hypervisor/src/lib.rs
> > index 2386d0924..9030b63b5 100644
> > --- a/src/runtime-rs/crates/hypervisor/src/lib.rs
> > +++ b/src/runtime-rs/crates/hypervisor/src/lib.rs
> > @@ -88,6 +88,7 @@ pub trait Hypervisor: Send + Sync {
> >      async fn get_agent_socket(&self) -> Result<String>;
> >      async fn disconnect(&self);
> >      async fn hypervisor_config(&self) -> HypervisorConfig;
> > +    async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool;
> >      async fn get_thread_ids(&self) -> Result<VcpuThreadIds>;
> >      async fn get_pids(&self) -> Result<Vec<u32>>;
> >      async fn get_vmm_master_tid(&self) -> Result<u32>;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```rust
> > impl Hypervisor for CloudHypervisor {
> >     async fn set_hypervisor_config(&self, config: HypervisorConfig) -> bool {
> >         // ...
> > 
> >         if need_to_specify_vm_rootfs_as_block_device(config) {
> >             return true;
> >         }
> > 
> >         return false;
> >     }
> > }
> > ```
> 
> Hi @Tim-0731-Hzt @jodh-intel If the pmem was only used for guest rootfs, and there was no chance to used for normal block device, I think it's ok to remove the pmem index and just set the block index started from 0 or 1 depended on set_hypervisor_config return value.

@lifupan @Tim-0731-Hzt Can we use the approach suggested by James here. I have talked about a similar approach in one of the comments posted earlier.
Today, pmem is only used by the rootfs, I understand that this PR attempts to add support for pmem devices that could be used as container devices, but I believe this is not really related to the problem we are trying to solve here.
As James mentioned the hypervisor could expose an API which returns whether, the VM rootfs is using block driver or not, the device manager could then use this to start with block_index of 0 or 1 accordingly.
This could solve the issue we are seeing currently with cloud-hypervisor and TDX.

Similarly, the API could be used to increment the pmem_index as well if pmem is being used for the VM rootfs.  
Does this sound like a reasonable approach?

cc @bergwolf

pmores:
I apologize for not participating so far, I was only made aware of this discussion yesterday.

> Hi @Tim-0731-Hzt - Thanks for updating this. I'm not sure the `pmem_index` idea will work though because for CH with TDX, the user should only need to set `confidential_guest=true` to enable a (TDX) confidential guest. However, in that scenario, we have to change the default VM rootfs config from pmem to virtio-blk. This happens "silently".

Isn't this "silence" part of the problem though? If a hypervisor driver has to silently override information essentially coming from `configuration.toml`, wouldn't it be better to declare the configuration as invalid in the first place?  CH validation of the config file could detect this condition early and let user know that TDX requires a block device for the VM rootfs and they need to specify that in their config file.  If we ensure that consistent information is coming straight from the config file then other parts of the runtime (device manager, hypervisor driver) can possibly be simpler.

I understand the objections against `pmem_index` however the `set_hypervisor_config()` proposal honestly doesn't seem ideal to me either - the function name suggests fairly high-level straightforward semantics but if the function also returns a very specific detailed piece of information I'm afraid it mixes abstraction levels badly.

jodh-intel:
@pmores - Regarding the automatic / silent (re-)configuration of particular settings, yes, that is a possibility. As @fidencio mentioned in the AC meeting on 2023-07-23, we could simply provide additional configuration files for reach major use-case. That would avoid having any automatic reconfiguration and would simplify switching to/from such configurations. However, it could complicate testing and there is the possibility of config file deviating as more options are added (or indeed some options removed).

>  the `set_hypervisor_config()` proposal 

I agree with your comments entirely. The `set_hypervisor_config()` idea came about as the simplest possible API change: I was incorrectly assuming that the Hypervisor API could not be changed any further. However, we discussed this in the AC meeting on Tuesday and this is happily not the case ;)

As documented in https://etherpad.opendev.org/p/Kata_Containers_2023_Architecture_Committee_Mtgs, we discussed approaches and the eventual plan is to move the rootfs logic into either the image management or device management layers. Meantime, we agreed to get this PR landed even though it (temporarily) complicates enabling `confidential_guest` for Cloud Hypervisor.

jodh-intel (approved):
Although this will complicate CH configuration temporarily, I'm approving [as discussed in the AC meeting on 2023-07-25](https://etherpad.opendev.org/p/Kata_Containers_2023_Architecture_Committee_Mtgs) on the understanding that we will simplify the rootfs handling logic, and since this PR should provide a usable system for CH (even though it complicates the configuration).

amshinde (approved):

PR #7162, opened 2023-06-22
gha: Don't automatically trigger CI

sprt:
We have GH configured so that manual approval is required for CI runs triggered by outside contributors. However, because CI is triggered by the `pull_request_target` event, this setting isn't being honored (see [1]).

This change aims to mititgate this issue by preventing PRs from triggering CI unless the `ok-to-test` label is set.

Note: For further context, we use the `pull_request_target` event and manually check out the PR branch because it is the only way to both access secrets and test incoming code changes.

Fixes: https://github.com/kata-containers/kata-containers/issues/7163

 [1]: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

### DIFF
--- a/.github/workflows/ci-on-push.yaml
+++ b/.github/workflows/ci-on-push.yaml
@@ -3,14 +3,25 @@ on:
   pull_request_target:
     branches:
       - 'main'
+    types:
+      # Adding 'labeled' to the list of activity types that trigger this event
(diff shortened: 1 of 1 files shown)


jepio (approved):

stevenhorsman (approved):
LGTM

stevenhorsman:
The AKS tests are a known problem, so this can be merged

Issue #7066, opened 2023-06-08
Migrate the Memory footprint tests, check metrics and report tool to kata-containers/tests


dborquez:
To continue the effort to adopt GitHub actions to run our metrics, it is required to move the following tests and artifacts to the kata-containers/metrics repository:

1. [Memory usage inside container](https://github.com/kata-containers/tests/blob/main/metrics/density/memory_usage_inside_container.sh)
2. [Memory usage (whole container)](https://github.com/kata-containers/tests/blob/main/metrics/density/memory_usage.sh)
3. [Memory Footprint at user space and kernel space](https://github.com/kata-containers/tests/blob/main/metrics/density/footprint_data.sh)
4. [Memory fast footprint](https://github.com/kata-containers/tests/blob/main/metrics/density/fast_footprint.sh)

PR #7174, opened 2023-06-26
metrics: Add memory footprint tests

dborquez:
This PR adds memory-foot-print tests to
kata-containers/tests folder.

Fixes: #7066


GabyCT (approved):

chavafg (approved):

GabyCT (approved):

.github/workflows/run-metrics.yaml
```diff
@@ -30,3 +30,15 @@ jobs:
...
+        run:  bash tests/metrics/gha-run.sh run-test-memory-usage-qemu
+
+      - name: run memory foot print on clh
+        run:  bash tests/metrics/gha-run.sh run-test-memory-usage-clh
+
+      - name: run memory usage inside container qemu
+        run:  bash tests/metrics/gha-run.sh run-test-memory-usage-inside-container-qemu
+
+      - name: run memory usage inside container clh
+        run:  bash tests/metrics/gha-run.sh run-test-memory-usage-inside-container-clh
```
fidencio, line 44: All of those steps could be made into a matrix, similar to what we have done here: https://github.com/kata-containers/kata-containers/blob/19890133e9b40ebe74d4f7359e61f56bedcb65ee/.github/workflows/run-k8s-tests-on-aks.yaml#L19-L28

This could simplify the process and make the code duplication smaller.

Let's **not** change it right now, though, but please, make this change in a future PR, okay?

tests/metrics/gha-run.sh
```diff
@@ -91,12 +91,32 @@ function run_test_launchtimes() {
...
 function main() {
 	action="${1:-}"
 	case "${action}" in
 		install-kata) install_kata ;;
 		run-test-launchtimes-qemu) run_test_launchtimes "qemu" ;;
 		run-test-launchtimes-clh) run_test_launchtimes "clh" ;;
+		run-test-memory-usage-qemu) run_test_memory_usage "qemu" ;;
+		run-test-memory-usage-clh) run_test_memory_usage "clh" ;;
+		run-test-memory-usage-inside-container-qemu) run_test_memory_usage_inside_container "qemu" ;;
+		run-test-memory-usage-inside-container-clh) run_test_memory_usage_inside_container "clh" ;;
```
fidencio, line 119: Ideally, we'd short-circuit this to return success and not execute the tests.

This would ensure that tests would be green when added (of course, they'd do nothing), and then you could enable them by changing the script (which is taken into consideration by GHA, differently than changes in the yaml file themselves).

I'd prefer having this done in the suggested way,  but I'll leave it up to you to decide whether you want to follow the suggestion or not.

fidencio (commented):
I didn't check the code itself, but rather the logic of the additions.

@dborquez, the last comment is important.

The one about the matrix is just for the future, and does not have to be addrssed in this PR.

.github/workflows/run-metrics.yaml
dborquez: @fidencio I take the opportunity to add the matrix in this PR.

tests/metrics/gha-run.sh
dborquez: @fidencio I took your suggestion and all the tests exits w/ zero.

fidencio (approved):
lgtm, thanks @dborquez!

Issue #7178, opened 2023-06-26
metrics: issue with permision denied  when retrieving clh version


dborquez:
`metrics/common.bash` fails to extract the `clh` version from `extract_kata_env()` function.
The problem could be solved calling `/opt/kata/bin/cloud-hypervisor --version` using sudo.

PR #7179, opened 2023-06-26
metrics: retrieve hypervisor version using sudo

dborquez:
This PR fixes a problem w/ permision denied when retrieving
the CLH version by using sudo.

Fixes: #7178


GabyCT:
@dborquez there are some files that are making reference to blogbench but I do not think they are part of this PR

Issue #7186, opened 2023-06-28
runtime-rs: direct volume report stat mount source xxx failed


Apokleos:
# Description of problem

```bash
# sudo kata-ctl direct-volume add /kubelet/kata-direct-vol-002/directvol002 "{\"device\": \"xxx\", \"volume_type\": \"directvol\", \"fs_type\": \"ext4\", \"metadata\":"{}", \"options\": []}"
# sudo ctr run -t --rm --runtime io.containerd.kata.v2 --mount type=directvol,src=/kubelet/kata-direct-vol-002/directvol002,dst=/disk002,options=rbind:rw fedora:35 kata-direct-vol-xx0628 /bin/bash
```
It fails to create the container and reports error message which indicates "stat mount source xxx failed".

# Expected result

kata container created OK

# Actual result

failed

The failure mainly caused by the encoded volume path and the `--mount type=directvol,src=/kubelet/kata-direct-vol-002/directvol002`. As the `src` will be validate with `stat`,but it's not a full path and encoded, which causes the `stat mount source /kubelet/kata-direct-vol-002/directvol002 failed` .

PR #7189, opened 2023-06-28
runtime-rs: bugfix for direct volume path's validation.

Apokleos:
The failure mainly caused by the encoded volume path and the mount/src. As the src will be validated with stat,but it's not a full path and encoded, which causes the stat mount source failed.

Fixes: #7186


quanweiZhou (approved):
Thanks, @Apokleos, LGTM.

Tim-0731-Hzt (approved):

GabyCT (approved):

Issue #7190, opened 2023-06-28
agent: agent has many unnecessary calls to `Arc::clone` 


wedsonaf:
# Description of problem

We have variants of the following pattern repeated in many places where `self` is a reference to `Self`:

```rust
let s = Arc::clone(&self.sandbox);
let mut sandbox = s.lock().await;
```

```rust
let s = self.sandbox.clone();
let mut sandbox = s.lock().await;
```

Given that `self` is immutable and guaranteed to outlive the call, cloning the `Arc` is not only unnecessary, it hurts performance because it results in two extra atomic operations in each code path: one to increment the refcount and another one to decrement it.

# Expected result

No unnecessary atomic operations.

# Actual result

A number of unnecessary atomic operations.

# Further information

The good news here is that the Rust compiler catches bugs in this area for us. As long as we don't use unsafe blocks, the compiler will complain when there are references that outlive the values they refer to.

PR #7191, opened 2023-06-28
agent: avoid unnecessary calls to `Arc::clone`

wedsonaf:
These calls cause two extra atomic instructions each time they're used, one to increment and another one to decrement the refcount.

Since we don't need them because the referred value is guaranteed to outlive the function, remove the calls.

Fixes: #7190

### DIFF
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -230,7 +230,7 @@ pub fn baremount(
 async fn ephemeral_storage_handler(
     logger: &Logger,
     storage: &Storage,
-    sandbox: Arc<Mutex<Sandbox>>,
+    sandbox: &Arc<Mutex<Sandbox>>,
 ) -> Result<String> {
     // hugetlbfs
     if storage.fstype == FS_TYPE_HUGETLB {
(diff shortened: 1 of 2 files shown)


jepio (approved):
looks sane to me.

I wonder whether some (most?) of these `&Arc<Mutex<Sandbox>>` params should actually be `&Mutex<Sandbox>` though, it feels like some of these call stacks no longer need to increment the reference count at all.

fidencio (approved):
Looks sensible to me, thanks @wedsonaf!

GabyCT (approved):

fidencio:
Wow, the CI seems to be quite flaky and I don't think it's related to this PR.

Let me re-run the tests and monitor their status.

fidencio:
@wedsonaf, could you rebase this one atop of current `main`?  Solving the conflicts is what we need to get this one merged.

wedsonaf:
> @wedsonaf, could you rebase this one atop of current `main`? Solving the conflicts is what we need to get this one merged.

Hey @fidencio, I rebased it last week but some tests seem to still be failing. Are we still seeing flakiness in the CI?

Thanks!

fidencio:
> Hey @fidencio, I rebased it last week but some tests seem to still be failing. Are we still seeing flakiness in the CI?

We were, Yesterday I got some fixes in that make things way more stable, although we're still seeing some issues every now and then to start the tests.  However, once they start, they've been solid.

PR #7193, opened 2023-06-28
stable-3.1: A bunch of kata-deploy fixes and a switch to using ubuntu as the base image

fidencio:
CentOS registry has been decomissioned between our last stable release and this one (see https://www.spinics.net/lists/centos-devel/msg22160.html, thanks @gkurz for pointing this out), which breaks our usage of CentOS as the base image.

As we've switched to using Ubuntu as the base image some time ago for the `main` branch, let's simply backport the needed patches here, instead of keep using CentOS.

### DIFF
--- a/tools/packaging/kata-deploy/Dockerfile
+++ b/tools/packaging/kata-deploy/Dockerfile
@@ -17,9 +17,9 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN \
 apt-get update && \
-apt-get install -y --no-install-recommends apt-transport-https ca-certificates curl xz-utils systemd && \
+apt-get install -y --no-install-recommends apt-transport-https ca-certificates curl gpg xz-utils systemd && \
 mkdir -p /etc/apt/keyrings/ && \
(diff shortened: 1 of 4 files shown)


tools/packaging/kata-deploy/Dockerfile
```diff
@@ -3,27 +3,28 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Specify alternative base image, e.g. clefos for s390x
-ARG IMAGE
-FROM ${IMAGE:-registry.centos.org/centos}:7
```
gkurz, line 7: This change doesn't belong to cd2aaeda2a07d2605634cc381b1bb3d68fe562a6 AFAICT. You fixed a conflict manually, right ?

Cherry picking bd1ed26c8d0ef8da99bc2ffffb099f93f73a3880 and 6267909501a1dc221e94ec2c6e80c7c0e1c083e7 seems to be enough to avoid that.

fidencio: Exactly, but I really would like to avoid cherry-picking GitHub actions that should never ever be part of this branch into here.

gkurz (approved):
LGTM with the justification that the GHA changes don't exist in stable-3.1.

Thanks @fidencio !

amshinde (approved):

Issue #7194, opened 2023-06-28
kata-deploy-stable: Switch to using the ubuntu based payload


fidencio:
As we're in the process of switching away from CentOS and using the
Ubuntu payload also for the maintained stable branch, we should make
sure we adapt the kata-deploy-stable.yaml on the main branch.

PR #7195, opened 2023-06-28
kata-deploy-stable: Switch to using the ubuntu based payload

fidencio:
As we're in the process of switching away from CentOS and using the Ubuntu payload also for the maintained stable branch, we should make sure we adapt the kata-deploy-stable.yaml on the main branch.

This is basically the same as cd2aaeda2a07d2605634cc381b1bb3d68fe562a6, but for this specific file.

Fixes: #7194


tools/packaging/kata-deploy/kata-deploy/base/kata-deploy-stable.yaml
```diff
@@ -29,7 +29,7 @@ spec:
             fieldRef:
               fieldPath: spec.nodeName
         securityContext:
-          privileged: false
+          privileged: true
```
zvonkok, line 32: Why `privileged: true`? What is blocking keeping `privileged:  false` ?

fidencio: Accessing /var/run/dbus and /run/systemd/system was not possible with privileged set to false for the Ubuntu based container, IIRC.

tools/packaging/kata-deploy/kata-deploy/base/kata-deploy-stable.yaml
```diff
@@ -6,43 +6,50 @@ metadata:
...
-        - name: kata-artifacts
-          mountPath: /opt/kata/
-        - name: dbus
-          mountPath: /var/run/dbus
-        - name: systemd
-          mountPath: /run/systemd
-        - name: local-bin
-          mountPath: /usr/local/bin/
+        - name: kube-kata
+          image: quay.io/kata-containers/kata-deploy:stable
```
gkurz, line 20: Isn't the point of this PR to use `:latest` ?

fidencio: Not really.  The point is to change the mounts we use with the stable image, so the kata-deploy daemonset doesn't break when folks try to use the "stable" image.

We want the stable file to still be pointing to the stable release.

Makes sense?

gkurz: What confuses me is that https://github.com/kata-containers/kata-containers/pull/7195/commits/4c69d516a9aff381226857992413e98611c8577c#diff-98322bfe8987923d1b67ca3949b9846ef827c37443c41fbde2dad4076c043fe7R22 changes kubelet-kata-cleanup to use `:latest` instead of `:stable`...

fidencio: Take a look at the following commits, from main:
* cd2aaeda2a07d2605634cc381b1bb3d68fe562a6
* 8f4b1df9cf301a1f91049a2d49a8ff7c80485b7c
* a56f96bb2b72ba71428c135dc790303e7cbbb0bc
* a274333248b3c66c5bda36aba818c20f67695caa
* 0e157be6f2ddb9bfcacc3c67a35f4e9c53dd7273
* 3230dec950e41c1a499a552b0a54739d079f9b97
* 4af78be13aa276cfadd4f4b859f62bdf45e5bac7

We need to adapt the stable yaml file to be able to cope with those changes, which are now as part of the stable-3.2 branch, which is what the images will be pointing to.  if we don't do that, whoever tries to use the kata-deploy-stable.yaml` file will just see it breaking.

> What confuses me is that https://github.com/kata-containers/kata-containers/commit/4c69d516a9aff381226857992413e98611c8577c#diff-98322bfe8987923d1b67ca3949b9846ef827c37443c41fbde2dad4076c043fe7R22 changes kubelet-kata-cleanup to use :latest instead of :stable...

This is me screwing things up, let me fix that, it should be `stable`.

gkurz (approved):
LGTM. Thanks @fidencio !

zvonkok (approved):
LGTM, thanks!

PR #7200, opened 2023-06-28
runtime: Add "none" as a shared_fs option

fidencio:
Currently, even when using devmapper, if the VMM supports virtio-fs /
virtio-9p, that's used to share a few files between the host and the
guest.

This needed, as we need to share with the guest contents like secrets,
certificates, and configurations, via Kubernetes objects like configMaps
or secrets, and those are rotated and must be updated into the guest
whenever the rotation happens.

However, there are still use-cases users can live with just copying
those files into the guest at the pod creation time, and for those
there's absolutely no need to have a shared filesystem process running
with no extra obvious benefit, consuming memory and even increasing the
attack surface used by Kata Containers.

For the case mentioned above, we should allow users, making it very
clear which limitations it'll bring, to run Kata Containers with
devmapper without actually having to use a shared file system, which is
already the approach taken when using Firecracker as the VMM.

Fixes: #7207

### DIFF
--- a/src/runtime/pkg/device/config/config.go
+++ b/src/runtime/pkg/device/config/config.go
@@ -81,6 +81,17 @@ const (
 
 	// VirtioFSNydus means use nydus for the shared file system
 	VirtioFSNydus = "virtio-fs-nydus"
+
+	// NoSharedFS means *no* shared file system solution will be used
+	// and files will be copied into the guest system.
+	//
+	// WARNING: This should be carefully used, and only used in very few
+	// specific cases, as any update to the mount will *NOT* be reflected
+	// during the lifecycle of the pod, causing issues with rotation of
+	// secrets, certs, or configurations via kubernetes objects like
+	// configMaps or secrets, as those will be copied into the guest at
+	// *pod* *creation* *time*.
+	NoSharedFS = "none"
 )
 
 const (
--- a/src/runtime/pkg/katautils/config.go
+++ b/src/runtime/pkg/katautils/config.go
@@ -512,7 +512,7 @@ func (h hypervisor) blockDeviceAIO() (string, error) {
 }
 
 func (h hypervisor) sharedFS() (string, error) {
-	supportedSharedFS := []string{config.Virtio9P, config.VirtioFS, config.VirtioFSNydus}
+	supportedSharedFS := []string{config.Virtio9P, config.VirtioFS, config.VirtioFSNydus, config.NoSharedFS}
 
 	if h.SharedFS == "" {
 		return config.VirtioFS, nil
@@ -1009,11 +1009,12 @@ func newClhHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		return vc.HypervisorConfig{}, err
 	}
 
-	if sharedFS != config.VirtioFS && sharedFS != config.VirtioFSNydus {
-		return vc.HypervisorConfig{}, errors.New("clh only support virtio-fs or virtio-fs-nydus")
+	if sharedFS != config.VirtioFS && sharedFS != config.VirtioFSNydus && sharedFS != config.NoSharedFS {
+		return vc.HypervisorConfig{},
+			fmt.Errorf("Cloud Hypervisor does not support %s shared filesystem option", sharedFS)
 	}
 
-	if h.VirtioFSDaemon == "" {
+	if (sharedFS == config.VirtioFS || sharedFS == config.VirtioFSNydus) && h.VirtioFSDaemon == "" {
 		return vc.HypervisorConfig{},
 			fmt.Errorf("cannot enable %s without daemon path in configuration file", sharedFS)
 	}
--- a/src/runtime/virtcontainers/qemu_arch_base.go
+++ b/src/runtime/virtcontainers/qemu_arch_base.go
@@ -51,7 +51,7 @@ type qemuArch interface {
 	kernelParameters(debug bool) []Param
 
 	//capabilities returns the capabilities supported by QEMU
-	capabilities() types.Capabilities
+	capabilities(config HypervisorConfig) types.Capabilities
 
 	// bridges sets the number bridges for the machine type
 	bridges(number uint32)
@@ -280,11 +280,13 @@ func (q *qemuArchBase) kernelParameters(debug bool) []Param {
 	return params
 }
 
-func (q *qemuArchBase) capabilities() types.Capabilities {
+func (q *qemuArchBase) capabilities(hConfig HypervisorConfig) types.Capabilities {
 	var caps types.Capabilities
 	caps.SetBlockDeviceHotplugSupport()
 	caps.SetMultiQueueSupport()
-	caps.SetFsSharingSupport()
+	if hConfig.SharedFS != config.NoSharedFS {
+		caps.SetFsSharingSupport()
+	}
 	return caps
 }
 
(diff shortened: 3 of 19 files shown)


fidencio:
I'll split this patch, but I'd like to get some initial feedback on it.

wedsonaf:
I like this. We'll need something like that for CC when running with CLH.

I was using the `ConfidentialGuest` config to disable fs sharing in CLH (https://github.com/microsoft/kata-containers/commit/3631bd3c11e61b5b1a3db9b61d44e99f0568eb0e) but I like this explicit option better, especially because we may eventually have other FS in clh-cc in the future.

src/runtime/pkg/device/config/config.go
```diff
@@ -81,6 +81,9 @@ const (
 
 	// VirtioFSNydus means use nydus for the shared file system
 	VirtioFSNydus = "virtio-fs-nydus"
+
+	// VirtioFSNone means files will be copied into the guest system
+	VirtioFSNone = "none"
```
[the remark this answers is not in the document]
fidencio: Totally, and we should also make it very clear in the configuration file.

src/runtime/pkg/katautils/config.go
```diff
@@ -1009,15 +1009,16 @@ func newClhHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		return vc.HypervisorConfig{}, err
 	}
 
-	if sharedFS != config.VirtioFS && sharedFS != config.VirtioFSNydus {
-		return vc.HypervisorConfig{}, errors.New("clh only support virtio-fs or virtio-fs-nydus")
-	}
-
-	if h.VirtioFSDaemon == "" {
+	if (sharedFS == config.VirtioFS || sharedFS != config.VirtioFSNydus) && h.VirtioFSDaemon == "" {
```
wedsonaf, line 1012: Did you mean `sharedFs == config.VirtioFSNydus` here?

src/runtime/pkg/katautils/config.go
```diff
@@ -1009,15 +1009,16 @@ func newClhHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
...
 		return vc.HypervisorConfig{},
 			fmt.Errorf("cannot enable %s without daemon path in configuration file", sharedFS)
 	}
 
+
+	if sharedFS == config.Virtio9P {
```
wedsonaf, line 1018: The change you're making here is analogous to going from an allow-list to a deny-list.

I think the consensus is that an allow-list is better though.

In this particular case, suppose we add a new Fs to kata that clh doesn't support; with a deny-list, we would wrongly accept it. With an allow-list, we would reject it. I think the latter is preferable because if it is meant to be allowed, someone will notice it and just add it here. Allowing something that shouldn't be allowed is harder to catch.

fidencio: Sure, I will change the check back to how it was, and also re-write the check above.

src/runtime/pkg/katautils/config.go
fidencio: I did. :-)

src/runtime/pkg/device/config/config.go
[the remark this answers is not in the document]
egernst: Specifically state that this is static at pod creation time, and that any updates to the mount will not be reflected during life cycle of the pod. As an example, this will result in rotation of secrets, certificates or configurations via kubernetes objects like configMaps or secrets.

src/runtime/virtcontainers/qemu_arch_base.go
```diff
@@ -168,12 +168,13 @@ type qemuArchBase struct {
...
-	disableNvdimm bool
-	dax           bool
-	legacySerial  bool
+	protection      guestProtection //nolint:structcheck
+	nestedRun       bool
+	vhost           bool
+	disableNvdimm   bool
+	dax             bool
+	legacySerial    bool
+	disableSharedFS bool
```
bergwolf, line 177: This is unused?

src/runtime/pkg/device/config/config.go
amshinde, line 86: I am fine with naming this as none, but was wondering if making as this "copy-files" instead of  "none" would be benefecial in terms of adding clarity.

fidencio:
Folks, this one is ready for review and I think I've addressed all your comments.

Please, take a second look at the PR whenever it's possible.

I've tested this locally with both Cloud Hypervisor and QEMU, which are the affected VMMs.

lifupan (approved):

wedsonaf (approved):

bergwolf (approved):
lgtm!

amshinde (approved):

src/runtime/pkg/device/config/config.go
```diff
@@ -81,6 +81,17 @@ const (
...
+	// VirtioFSNone means *no* shared file system solution will be used
+	// and files will be copied into the guest system.
+	//
+	// WARNING: This should be carefully used, and only used in very few
+	// specific cases, as any update to the mount will *NOT* be reflected
+	// during the lifecycle of the pod, causing issues with rotation of
+	// secrets, certs, or configurations via kubernetes objects like
+	// configMaps or secrets, as those will be copied into the guest at
+	// *pod* *creation* *time*.
+	VirtioFSNone = "none"
```
egernst, line 94: Why string?

src/runtime/pkg/katautils/config.go
```diff
@@ -1009,11 +1009,12 @@ func newClhHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		return vc.HypervisorConfig{}, err
 	}
 
-	if sharedFS != config.VirtioFS && sharedFS != config.VirtioFSNydus {
-		return vc.HypervisorConfig{}, errors.New("clh only support virtio-fs or virtio-fs-nydus")
+	if sharedFS != config.VirtioFS && sharedFS != config.VirtioFSNydus && sharedFS != config.VirtioFSNone {
```
egernst, line 1012: This is a bit awkward, in that we also are doing just about this same check up above in the map of supported shared-filesystem options in sharedFS.  I wonder if there's a way we can dedup/clean this up, so we don't have to make changes in multiple spots again next time.

egernst: Ah, sorry - I see that's a 'general' check, whereas this is cloud hypervisor specific check.  meh, its a bit awkward/broken abstraction that we are checking clh specific bits in a generic katautils pkg, but I guess that isn't necessarily your fault.

egernst:
This PR reminded me of a couple of broken abstractions and how large VC itself is, but neither item is your fault here.  The changes lgtm @fidencio. Are there integration tests that exercise this available already? I'd prefer we have these queued up and exercised -- I trust you tested, but I don't trust that we won't immediately break it if we don't have tests.

fidencio:
> Are there integration tests that exercise this available already?

@egernst, there are tests for the capaibilites on the hypervisor side, and I expanded those to ensure that whenever "NoSharedFS" is set, "IsFsSharingSupported" returns false.

I've expanded this on the QEMU side, and added it to the Cloud Hypervisor side.

Please, let me know if this is enough to cover your concerns.

egernst (approved):
Thanks - I like the rename - reads more natural now.  lgtm.

Issue #7210, opened 2023-06-29
runtime: Forward point watcher to propagate secrets, config maps etc


stevenhorsman:
Create a watcher for secrets / certs / config maps / etc, and copy the file into the guest

PR #7211, opened 2023-06-29
Propogate secrets, config maps etc into guest if sharedFS not available

stevenhorsman:
Fixes: #7210

### DIFF
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -8,8 +8,9 @@ use rustjail::{pipestream::PipeStream, process::StreamType};
 use tokio::io::{AsyncReadExt, AsyncWriteExt, ReadHalf};
 use tokio::sync::Mutex;
 
-use std::ffi::CString;
+use std::ffi::{CString, OsStr};
 use std::io;
+use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 use std::sync::Arc;
 use ttrpc::{
@@ -1854,6 +1855,38 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
                 std::fs::set_permissions(&dir, std::fs::Permissions::from_mode(req.dir_mode))?;
             }
         }
+    }
+
+    let sflag = stat::SFlag::from_bits_truncate(req.file_mode);
+
+    if sflag.contains(stat::SFlag::S_IFDIR) {
+        fs::create_dir(&path).or_else(|e| {
+            if e.kind() != std::io::ErrorKind::AlreadyExists {
+                return Err(e);
+            }
+            Ok(())
+        })?;
+
+        std::fs::set_permissions(&path, std::fs::Permissions::from_mode(req.file_mode))?;
+
+        unistd::chown(
+            &path,
+            Some(Uid::from_raw(req.uid as u32)),
+            Some(Gid::from_raw(req.gid as u32)),
+        )?;
+
+        return Ok(());
+    }
+
+    if sflag.contains(stat::SFlag::S_IFLNK) {
+        let src = PathBuf::from(OsStr::from_bytes(&req.data));
+        unistd::symlinkat(&src, None, &path)?;
+        let path_str = CString::new(path.as_os_str().as_bytes())?;
+
+        let ret = unsafe { libc::lchown(path_str.as_ptr(), req.uid as u32, req.gid as u32) };
+        Errno::result(ret).map(drop)?;
+
+        return Ok(());
     }
 
     let mut tmpfile = path.clone();
--- a/src/runtime/virtcontainers/fs_share_darwin.go
+++ b/src/runtime/virtcontainers/fs_share_darwin.go
@@ -56,3 +56,10 @@ func (f *FilesystemShare) ShareRootFilesystem(ctx context.Context, c *Container)
 func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Container) error {
 	return nil
 }
+
+func (f *FilesystemShare) StartFileEventWatcher(ctx context.Context) error {
+	return nil
+}
+
+func (f *FilesystemShare) StopFileEventWatcher(ctx context.Context) {
+}
--- a/src/runtime/virtcontainers/fs_share_linux.go
+++ b/src/runtime/virtcontainers/fs_share_linux.go
@@ -14,9 +14,11 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sync"
 	"syscall"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -33,14 +35,29 @@ func unmountNoFollow(path string) error {
 
 type FilesystemShare struct {
 	sandbox *Sandbox
+	watcher *fsnotify.Watcher
+	// The same volume mount can be shared by multiple containers in the same sandbox (pod)
+	srcDstMap            map[string][]string
+	srcDstMapLock        sync.Mutex
+	eventLoopStarted     bool
+	eventLoopStartedLock sync.Mutex
+	watcherDoneChannel   chan bool
 	sync.Mutex
 	prepared bool
 }
 
 func NewFilesystemShare(s *Sandbox) (FilesystemSharer, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("Creating watcher returned error %w", err)
+	}
+
 	return &FilesystemShare{
-		prepared: false,
-		sandbox:  s,
+		prepared:           false,
+		sandbox:            s,
+		watcherDoneChannel: make(chan bool),
+		srcDstMap:          make(map[string][]string),
+		watcher:            watcher,
 	}, nil
 }
 
@@ -267,8 +284,59 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
 			}
 
 			dstPath := filepath.Join(guestPath, srcPath[len(srcRoot):])
+			f.Logger().Infof("ShareFile: Copying file from src (%s) to dest (%s)", srcPath, dstPath)
+			//TODO: Improve the agent protocol, to handle the case for existing symlink.
+			// Currently for an existing symlink, this will fail with EEXIST.
+			err = f.sandbox.agent.copyFile(ctx, srcPath, dstPath)
+			if err != nil {
+				f.Logger().WithError(err).Error("Failed to copy file")
+				return err
+			}
 
-			return f.sandbox.agent.copyFile(ctx, srcPath, dstPath)
+			// Add fsNotify watcher for volume mounts
+			// Use regex for strict matching instead of strings.Contains
+			// match for kubernetes.io~configmap, kubernetes.io~secret, kubernetes.io~projected, kubernetes.io~downward-api
+			// as recommended in review comments for PR #7211
+
+			// Example directory structure for the volume mounts.
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~configmap
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~secret
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~projected
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~downward-api
+
+			// More relaxed regex for the pod UID
+			// `^/var/lib/kubelet/pods/[a-fA-F0-9\-]+/volumes/kubernetes\.io~(configmap|secret|projected|downward-api)`
+
+			//TODO: Move this to a global variable and compile only once.
+			regex := regexp.MustCompile(`^/var/lib/kubelet/pods/[a-fA-F0-9\-]{36}/volumes/kubernetes\.io~(configmap|secret|projected|downward-api)`)
+			if regex.MatchString(srcPath) {
+				// fsNotify doesn't add watcher recursively.
+				// So we need to add the watcher for directories under kubernetes.io~configmap, kubernetes.io~secret,
+				// kubernetes.io~downward-api and kubernetes.io~projected
+				if info.Mode().IsDir() {
+					// The cm dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~configmap/foo/{..data, key1, key2,...}
+					// The secret dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~secret/foo/{..data, key1, key2,...}
+					// The projected dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~projected/foo/{..data, key1, key2,...}
+					// The downward-api dir is of the form /var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~downward-api/foo/{..data, key1, key2,...}
+					f.Logger().Infof("ShareFile: srcPath(%s) is a directory", srcPath)
+					err := f.watchDir(srcPath)
+					if err != nil {
+						f.Logger().WithError(err).Error("Failed to watch directory")
+						return err
+					}
+				} else {
+					f.Logger().Infof("ShareFile: srcPath(%s) is not a directory", srcPath)
+				}
+				// Add the source and destination to the global map which will be used by the event loop
+				// to copy the modified content to the destination
+				f.Logger().Infof("ShareFile: Adding srcPath(%s) dstPath(%s) to srcDstMap", srcPath, dstPath)
+				// Lock the map before adding the entry
+				f.srcDstMapLock.Lock()
+				defer f.srcDstMapLock.Unlock()
+				f.srcDstMap[srcPath] = append(f.srcDstMap[srcPath], dstPath)
+			}
+
+			return nil
 		}
 
 		if err := filepath.WalkDir(srcRoot, walk); err != nil {
@@ -278,6 +346,7 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
 		if ignored {
 			return nil, nil
 		}
+
 	} else {
 		// These mounts are created in the shared dir
 		mountDest := filepath.Join(getMountPath(f.sandbox.ID()), filename)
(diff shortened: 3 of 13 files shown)


src/runtime/virtcontainers/container.go
```diff
@@ -529,6 +529,12 @@ func (c *Container) mountSharedDirMounts(ctx context.Context, sharedDirMounts, i
 		sharedDirMounts[sharedDirMount.Destination] = sharedDirMount
 	}
 
+	// Start the event loop to watch for file change notifications
+	// The event loop will only start if there are watches added
+	if c.sandbox.fsShare != nil {
+		go c.sandbox.fsShare.StartFileEventWatcher(ctx)
+	}
```
wedsonaf, line 536: Should we put this in the same deferred call created at the beginning of the function?

Because if we have early returns (e.g., line 477 above), then we end up not starting the watcher.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1247,6 +1247,7 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
 	sharedDirMounts := make(map[string]Mount)
 	ignoredMounts := make(map[string]Mount)
 
+	k.Logger().Info("mounting shared dir mounts")
```
wedsonaf, line 1250: Was this meant to be left here?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2181,40 +2182,57 @@ func (k *kataAgent) setGuestDateTime(ctx context.Context, tv time.Time) error {
...
+		cpReq.Data = []byte(symlink)
+
+	default:
+		return fmt.Errorf("Unsupported file type: %o", sflag)
+	}
 
 	k.Logger().WithFields(logrus.Fields{
 		"source": src,
 		"dest":   dst,
 	}).Debugf("Copying file from host to guest")
```
wedsonaf, line 2226: Maybe we want to specify the kind of thing we're copying here as opposed to calling it just a file?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2181,40 +2182,57 @@ func (k *kataAgent) setGuestDateTime(ctx context.Context, tv time.Time) error {
...
+		// TODO: Support incremental file copying instead of loading whole file into memory
+		b, err = os.ReadFile(src)
+		if err != nil {
+			return fmt.Errorf("Could not read file %s: %v", src, err)
+		}
+		cpReq.FileSize = int64(len(b))
+
+	case unix.S_IFDIR:
+
+	case unix.S_IFLNK:
```
wedsonaf, line 2212: Does the agent do the right thing when a link is passed to it?

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2181,40 +2182,57 @@ func (k *kataAgent) setGuestDateTime(ctx context.Context, tv time.Time) error {
 func (k *kataAgent) copyFile(ctx context.Context, src, dst string) error {
 	var st unix.Stat_t
 
-	err := unix.Stat(src, &st)
+	err := unix.Lstat(src, &st)
```
wedsonaf, line 2185: This is a change in behaviour: we used to just copy the contents of the pointee, now we're send the link in (potentially to a non-existing file).

Are we sure this isn't break anything?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -239,23 +254,81 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
...
+			//TODO: Improve the agent protocol, to handle the case for existing symlink.
+			// Currently for an existing symlink, this will fail with EEXIST.
+			err = f.sandbox.agent.copyFile(ctx, srcPath, dstPath)
+			if err != nil {
+				f.Logger().WithError(err).Error("Failed to copy file")
+				return err
+			}
+
+			// Add fsNotify watcher for volume mounts
+			if strings.Contains(srcPath, "kubernetes.io~configmap") ||
```
wedsonaf, line 293: Should we have tighter checks here? That is, checks the paths documented in lines 302-305 below.

Because otherwise we could have a file that happens to contain say `kubernetes.io~configmap` in its name, and it will get unexpected treatment. User may want them to copied once at start but now they're being updated unintentionally.

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -239,23 +254,81 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
...
+				return err
+			}
+
+			info, err := d.Info()
+			if err != nil {
+				return err
+			}
+
+			if !(info.Mode().IsRegular() || info.Mode().IsDir() || (info.Mode()&os.ModeSymlink) == os.ModeSymlink) {
+				f.Logger().WithField("ignored-file", srcPath).Debug("Ignoring non-regular file as FS sharing not supported")
```
wedsonaf, line 272: Given that we accept other types of files, could we update the comment to reflect that? (If we don't want to list the accepted types, perhaps we should just say that it is unsuported.)

src/runtime/virtcontainers/kata_agent.go
yoheiueda: We need a corresponding change to agent in de232b8030eba9908557ac6546b97e8044685bfc of CCv0

src/runtime/virtcontainers/kata_agent.go
yoheiueda: Yes, it changes the original behavior, but I am not sure it break any existing use cases.

One idea to keep the compatibility is that we de-reference the root of the source path.  The original implementation does not support recursive copying, so the source path is a regular file or a symlink to a regular file.

We use `unix.Stat` for the root of the source path, and determine whether it is a file or directory. If the source path is a directory, then its children are examined using `unix.Lstat`.

src/runtime/virtcontainers/fs_share_linux.go
yoheiueda: A possible tighter check was  discussed here.
https://github.com/kata-containers/kata-containers/pull/6342#discussion_r1151410817

wedsonaf: My read of this comment is that you had two points: optimising the check and making it stricter, but only the former was addressed.

Let's discuss the stricter check now then :)

src/runtime/virtcontainers/kata_agent.go
wedsonaf: @stevenhorsman would it be possible to add this commit to this PR?

fidencio: [This](https://github.com/kata-containers/kata-containers/commit/de232b8030eba9908557ac6546b97e8044685bfc) should be part of this PR as well, then.

src/runtime/virtcontainers/container.go
stevenhorsman: I've moved this up into the defer as suggested.

src/runtime/virtcontainers/kata_agent.go
stevenhorsman: Deleted

src/runtime/virtcontainers/kata_agent.go
stevenhorsman: Added the agent side commit into the PR

src/runtime/virtcontainers/fs_share_linux.go
stevenhorsman: I've removed the `non-regular` from the message

src/runtime/virtcontainers/fs_share_linux.go
yoheiueda: > Possibly we can do more strict checking here to reduce false-positives like this.
> https://go.dev/play/p/JNNXZSmzLkS

I meant this one cloud be a tighter check because it does not rely on `strings.Contains`.

stevenhorsman:
@yoheiueda @wedsonaf @bpradipt - what's the next steps here. Are the fixes I've done okay, so we can see what the remaining issues are? Thanks!

bpradipt:
> @yoheiueda @wedsonaf @bpradipt - what's the next steps here. Are the fixes I've done okay, so we can see what the remaining issues are? Thanks!

The fixes looks ok to me. 
But unfortunately I couldn't test this today as I was unable to spin up a working cluster setup today :-(. I'll try again tomorrow

src/runtime/virtcontainers/fs_share_linux.go
yoheiueda: I talked with @bpradipt about this.

This PR is cherry-picking commits from the CCv0 branch. The original commits are tested as a part of CCv0 for some time, and I think it is OK to merge the proven commits as they are into the main for now. Then, we will submit another PR to improve the code with the stricter checking.

@wedsonaf what do you think?

fidencio:
@bpradipt, @yoheiueda, @wedsonaf, any news here?

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: thanks @wedsonaf

bpradipt:
@fidencio /lgtm from my side

wedsonaf:
And apologies for the delay in following up, I was out of the office last week.

src/agent/src/rpc.rs
```diff
@@ -1936,6 +1936,38 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
...
+            &path,
+            Some(Uid::from_raw(req.uid as u32)),
+            Some(Gid::from_raw(req.gid as u32)),
+        )?;
+
+        return Ok(());
+    }
+
+    if sflag.contains(stat::SFlag::S_IFLNK) {
+        let src = PathBuf::from(String::from_utf8(req.data.clone()).unwrap());
```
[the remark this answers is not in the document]
stevenhorsman: Hey Wedson, but Rust isn't brilliant, is something like:
```rust
        let src = match String::from_utf8(req.data.clone()) {
            Err(e) => {
                return Err(anyhow!("failed to convert bytes to text: {}", e));
            }
            Ok(path) => PathBuf::from(path),
        };
```
what you had in mind, or is there something more idiomatic?

src/agent/src/rpc.rs
```diff
@@ -1936,6 +1936,38 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
...
+
+        unistd::symlinkat(&src, None, &path)?;
+        let path_str = CString::new(path.to_str().unwrap())?;
```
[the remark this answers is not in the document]
stevenhorsman: How about:
```
        let path = path
            .to_str()
            .ok_or_else(|| anyhow!("Could not convert path to string"))?;
        let path_str = CString::new(path)?;
```
?

src/agent/src/rpc.rs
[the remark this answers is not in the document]
wedsonaf: I think the `ok_or_else` you have in the next one is more idiomatic.

But given that what we want in the end is a `PathBuf`, which doesn't require utf8, I think we can avoid the conversion and allocation altogether if we avoid the `String` type. I'm thinking:

```rust
let src = PathBuf::from(OsStr::from_bytes(&req.data));
```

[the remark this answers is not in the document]
stevenhorsman: Thanks, that's super helpful. I've added the rust changes suggested as a new commit for convenience, but once they have been reviewed I'll fix it up in to the other commit.

src/agent/src/rpc.rs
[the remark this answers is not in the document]
wedsonaf: I think we can also avoid `String` here:

```rust
let path_str = CString::new(path.as_os_str().as_bytes())?;
```

Note that `to_str` requires utf8 encoding and allocates a new buffer for the string. `as_os_str().as_bytes()` doesn't convert and doesn't require new allocations.

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -239,23 +254,81 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
...
+						f.Logger().WithError(err).Error("Failed to watch directory")
+						return err
+					}
+				} else {
+					f.Logger().Infof("ShareFile: srcPath(%s) is not a directory", srcPath)
+				}
+				// Add the source and destination to the global map which will be used by the event loop
+				// to copy the modified content to the destination
+				f.Logger().Infof("ShareFile: Adding srcPath(%s) dstPath(%s) to srcDstMap", srcPath, dstPath)
+				f.srcDstMap[srcPath] = dstPath
```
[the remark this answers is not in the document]
bpradipt: @stevenhorsman I missed the notifications. On the first glance it appears @wedsonaf is right.  I will look into the code again and get back.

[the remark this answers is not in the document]
bpradipt: Relooked at my notes (thankfully it was still there) and the code. My reading of the code and understanding was that the creation of the containers in the Kata pod was sequential based on this flow and hence synchronising writes to the `srcDstMap` was not required
 api.go (createSandbox) -> sandbox.go (createContainers) -> (create all containers) container.go (create) -> kata_agent.go (createContainer) -> container.go (mountSharedDirMounts) -> fs_share_linux.go StartFileEventWatcher followed by ShareFile

And indeed there will be multiple goroutines for the event loop (StartFileEventWatcher). 

I'm trying out the changes in this PR and also discussing with @yoheiueda. Will get back with details. Thanks @wedsonaf for pointing this out.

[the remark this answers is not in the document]
bpradipt: @wedsonaf I can't confirm if there is no data race and your explanation makes sense to me. I'm trying out a fix to avoid the race and update here.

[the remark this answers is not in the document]
bpradipt: @wedsonaf I have made changes based on your suggestions which is in my fork - https://github.com/bpradipt/kata-containers/tree/pr7211
Once @stevenhorsman is back we'll update this PR with the changes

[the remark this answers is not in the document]
stevenhorsman: I've cherry-picked in your commit Pradipta, so I think this is ready for re-review @wedsonaf @bpradipt. Thanks for your patience and sorry for the hold-up whilst I was out!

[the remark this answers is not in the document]
stevenhorsman: BTW once we are happy with the commit then I'll `squash` it to stop the commit message check failure

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -666,12 +696,14 @@ func (f *FilesystemShare) StartFileEventWatcher(ctx context.Context) error {
...
 					// eg. dataDir = /var/lib/kubelet/pods/b44e3261-7cf0-48d3-83b4-6094bba95dc8/volumes/kubernetes.io~configmap/foo/..data
 
-					destination := f.srcDstMap[dataDir]
-					f.Logger().Infof("StartFileEventWatcher: Copy file from src (%s) to dst (%s)", dataDir, destination)
-					err := f.copyFilesFromDataDir(dataDir, destination)
-					if err != nil {
-						f.Logger().Infof("StartFileEventWatcher: got an error (%v) when copying file from src (%s) to dst (%s)", err, dataDir, destination)
-						return err
+					// Handle different destination for the same source
+					for _, destination := range f.srcDstMap[dataDir] {
```
wedsonaf, line 700: What prevents a concurrent writer to conflict with this read?

Unless there's some synchronization here I'm missing, you need to acquire `srcDstMapLock` here too to prevent modifications to `srcDstMap` while you're iterating over it.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1208,6 +1208,7 @@ func (k *kataAgent) rollbackFailingContainerCreation(ctx context.Context, c *Con
 	}
 }
 
+// bpradipt: This function is called to create a container
```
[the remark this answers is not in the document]
bpradipt: Yeah. I'll remove

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -589,17 +607,29 @@ func (f *FilesystemShare) watchDir(source string) error {
 
 func (f *FilesystemShare) StartFileEventWatcher(ctx context.Context) error {
 
-	// Start event loop if watchList is not empty
-	if (f.watcher == nil) || len(f.watcher.WatchList()) == 0 {
-		f.Logger().Info("StartFileEventWatcher: No watches found, returning")
+	// Check if the event loop is already started
+	if f.eventLoopStarted {
```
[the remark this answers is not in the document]
bpradipt: I see.. Thanks for the suggestion. Will do

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: You are right. Will make the changes.

src/runtime/virtcontainers/sandbox.go
```diff
@@ -614,6 +614,18 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
 		return nil, err
 	}
 
+	// Start the event loop if not already started when fs sharing is not used
+	if sandboxConfig.HypervisorConfig.SharedFS == config.NoSharedFS {
+		// Start the StartFileEventWatcher method as a goroutine
+		// to monitor the file events.
+		go func() {
```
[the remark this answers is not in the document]
bpradipt: Any suggestions on how to handle this case ?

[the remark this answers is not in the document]
wedsonaf: I think something like what's in line 584 would work: add a deferred call that stops the watcher if `retErr != nil`.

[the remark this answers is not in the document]
bpradipt: Ok, I think calling `s.fsShare.StopFileEventWatcher` in case of errors to stop the go routine should suffice.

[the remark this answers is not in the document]
bpradipt: @wedsonaf sorry somehow I missed your suggestion when I added the above message. Github didn't show me that :-( I instead added the StopFileEventWatcher in createSandbox where newSandbox is called. I'll add it as a deferred call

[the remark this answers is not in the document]
bpradipt: @wedsonaf I have updated the code based on your suggestions. Kindly take a relook and let me know your comments and suggestions for further improvements.

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -476,3 +567,238 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+					// Handle different destination for the same source
+					// Acquire srcDstMapLock before reading srcDstMap
+					f.srcDstMapLock.Lock()
+					for _, destination := range f.srcDstMap[dataDir] {
+						f.Logger().Infof("StartFileEventWatcher: Copy file from src (%s) to dst (%s)", dataDir, destination)
+						err := f.copyFilesFromDataDir(dataDir, destination)
+						if err != nil {
+							f.Logger().Infof("StartFileEventWatcher: got an error (%v) when copying file from src (%s) to dst (%s)", err, dataDir, destination)
+							f.srcDstMapLock.Unlock()
+							return err
```
wedsonaf, line 711: If we fail to copy a single instance of a data directory, we stop the whole event watcher?

In fact, this is a bit of a silent failure because it stops watching but everything else seems to be working.

I think we should continue to attempt to copy the other files and keep the watcher alive.

Alternatively, if we really want to fail, I think we should fail more loudly. Stopping the whole sandbox may seem too drastic but if we to stop its watcher, this may be preferrable than limping along.

WDYT?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -476,3 +567,238 @@ func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Containe
...
+					}
+					f.srcDstMapLock.Unlock()
+				}
+			}
+		case err, ok := <-f.watcher.Errors:
+			if !ok {
+				return fmt.Errorf("StartFileEventWatcher: Error (%v) in receiving error events", err)
+			}
+			f.Logger().Infof("StartFileEventWatcher: got an error event (%v)", err)
+			return err
```
wedsonaf, line 722: Can these errors be transient? Or do they indicate that the watcher is hopelessly broken?

src/runtime/virtcontainers/fs_share_darwin.go
```diff
@@ -56,3 +56,10 @@ func (f *FilesystemShare) ShareRootFilesystem(ctx context.Context, c *Container)
 func (f *FilesystemShare) UnshareRootFilesystem(ctx context.Context, c *Container) error {
 	return nil
 }
+
+func (f *FilesystemShare) StartFileEventWatcher(ctx context.Context) error {
+	return nil
```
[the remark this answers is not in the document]
bpradipt: Fair point. I think it will need more discussion to handle the other methods as well like (Un)ShareFile etc.

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: I think just logging the error and continuing makes sense.

src/runtime/virtcontainers/fs_share_linux.go
bpradipt: Some typical errors could be permission related or resource exhaustion. I'm not sure if permission related errors can happen here. And resource exhaustion can be a transient error.  Also given the previous comment, it might be ok to log the errors and continue.

bpradipt: Updated the code..

src/agent/src/rpc.rs
```diff
@@ -1942,6 +1943,38 @@ fn do_copy_file(req: &CopyFileRequest) -> Result<()> {
...
+
+    if sflag.contains(stat::SFlag::S_IFDIR) {
+        fs::create_dir(&path).or_else(|e| {
+            if e.kind() != std::io::ErrorKind::AlreadyExists {
+                return Err(e);
+            }
+            Ok(())
+        })?;
+
+        std::fs::set_permissions(path.clone(), std::fs::Permissions::from_mode(req.file_mode))?;
```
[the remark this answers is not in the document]
bpradipt: @stevenhorsman

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -239,23 +258,95 @@ func (f *FilesystemShare) ShareFile(ctx context.Context, c *Container, m *Mount)
...
+			// Example directory structure for the volume mounts.
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~configmap
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~secret
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~projected
+			// /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~downward-api
+
+			// More relaxed regex for the pod UID
+			// `^/var/lib/kubelet/pods/[a-fA-F0-9\-]+/volumes/kubernetes\.io~(configmap|secret|projected|downward-api)`
+
+			regex := regexp.MustCompile(`^/var/lib/kubelet/pods/[a-fA-F0-9\-]{36}/volumes/kubernetes\.io~(configmap|secret|projected|downward-api)`)
```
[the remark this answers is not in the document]
bpradipt: I have added a TODO so that we can revisit this

wedsonaf (approved):

src/agent/src/rpc.rs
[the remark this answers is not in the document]
stevenhorsman: Fixed

fidencio (approved):
lgtm, on behalf of @wedsonaf!
lgtm, on my own behalf!

thanks @stevenhorsman, @bpradipt, @yoheiueda for the work on this.  thanks @wedsonaf for the time spent on the reviews!

stevenhorsman:
> @bpradipt - are you okay for me to fixup your review commit into `runtime: propagate configmap/secrets etc changes for remote-hyp` before we merge this?

I had to anyway to get the commit checks to pass!

stevenhorsman:
> thanks @stevenhorsman, @bpradipt, @yoheiueda for the work on this. thanks @wedsonaf for the time spent on the reviews!

Agreed - big thanks to Wedson and Dan for the feedback that made this code much better and Pradipta for implementing the changes!

bpradipt:
> @bpradipt - are you okay for me to fixup your review commit into `runtime: propagate configmap/secrets etc changes for remote-hyp` before we merge this?

Sure, please go ahead

stevenhorsman:
@UnmeshDeodhar @ryansavino SEV is failing and SNP won't start on these tests (and both failed the last two nightly runs) any chance you can check the nodes please? Thanks!

ryansavino:
Looks like SEV passed. For the SNP failure, @fidencio i see your note here:
https://github.com/kata-containers/kata-containers/blob/b975c277932746c252d1c5835d6a7f92d8330d09/tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats#L15

However, I'm still seeing all the runtime classes on the node:
```
kata@amd-milan-kata-ci-ubuntu2004-001:~/ci-setup/kata-containers$ k get runtimeclasses
NAME                   HANDLER                AGE
kata                   kata                   2m40s
kata-clh               kata-clh               5d4h
kata-dragonball        kata-dragonball        5d4h
kata-fc                kata-fc                5d4h
kata-qemu              kata-qemu              5d4h
kata-qemu-nvidia-gpu   kata-qemu-nvidia-gpu   5d4h
kata-qemu-sev          kata-qemu-sev          5d4h
kata-qemu-snp          kata-qemu-snp          2m40s
kata-qemu-tdx          kata-qemu-tdx          5d4h
```

Is this the kata-deploy cleanup issue you mentioned to me in DM on slack? Any BKM to clean it up?

jiangliu (approved):

jepio:
I'll merge as the failures are unrelated to the PR. Thanks @stevenhorsman !

ryansavino:
FYI, I cleaned up the runtime classes on the node. This should hopefully pass now for other PRs.

Issue #7215, opened 2023-06-30
runtime-rs: enhancement of Device Manager for network devices management


Apokleos:
**Which feature do you think can be improved?**

Currently, the network devices have implemented, but they're alone outside the device manager, we should include them into the 
device manager for management.

Another reason is that some network devices are also PCI devices and we need to manage their corresponding PCI paths. 

**How can it be improved?**

refactor the network devices' implementation with the device manager

@lifupan @Tim-0731-Hzt @quanweiZhou

PR #7219, opened 2023-07-02
runtime-rs: enhancement of Device Manager for network endpoints.

Apokleos:
Currently the network endpoints are alone out of device manager, we should include them into the device manager for management.

If so, refactoring network endpoints' implementation is needed.

First, restruct NetworkConfig and NetworkDevice structures; Second, do implementation for virtio-net driver and add Network device into Device Manager;
Third, unified the entry with do_handle_device for each endpoint;

Fixes: #7215

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
@@ -6,6 +6,14 @@
 
 use std::fmt;
 
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+
+use crate::{
+    device::{Device, DeviceType},
+    Hypervisor as hypervisor,
+};
+
 #[derive(Clone)]
 pub struct Address(pub [u8; 6]);
 
@@ -20,20 +28,71 @@ impl fmt::Debug for Address {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct NetworkConfig {
+    /// for detach, now it's default value 0.
+    pub index: u64,
+
     /// Host level path for the guest network interface.
     pub host_dev_name: String,
+
+    /// Guest iface name for the guest network interface.
+    pub virt_iface_name: String,
 
     /// Guest MAC address.
     pub guest_mac: Option<Address>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct NetworkDevice {
     /// Unique identifier of the device
-    pub id: String,
+    pub device_id: String,
 
     /// Network Device config info
     pub config: NetworkConfig,
+}
+
+impl NetworkDevice {
+    // new creates a NetworkDevice
+    pub fn new(device_id: String, config: &NetworkConfig) -> Self {
+        Self {
+            device_id,
+            config: config.clone(),
+        }
+    }
+}
+
+#[async_trait]
+impl Device for NetworkDevice {
+    async fn attach(&mut self, h: &dyn hypervisor) -> Result<()> {
+        h.add_device(DeviceType::Network(self.clone()))
+            .await
+            .context("add network device.")?;
+
+        return Ok(());
+    }
+
+    async fn detach(&mut self, h: &dyn hypervisor) -> Result<Option<u64>> {
+        h.remove_device(DeviceType::Network(self.clone()))
+            .await
+            .context("remove network device.")?;
+
+        Ok(Some(self.config.index))
+    }
+
+    async fn get_device_info(&self) -> DeviceType {
+        DeviceType::Network(self.clone())
+    }
+
+    async fn increase_attach_count(&mut self) -> Result<bool> {
+        // network devices will not be attached multiple times, Just return Ok(false)
+
+        Ok(false)
+    }
+
+    async fn decrease_attach_count(&mut self) -> Result<bool> {
+        // network devices will not be detached multiple times, Just return Ok(false)
+
+        Ok(false)
+    }
 }
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
@@ -48,7 +48,7 @@ impl DragonballInner {
         info!(sl!(), "dragonball add device {:?}", &device);
         match device {
             DeviceType::Network(network) => self
-                .add_net_device(&network.config, network.id)
+                .add_net_device(&network.config)
                 .context("add net device"),
             DeviceType::Vfio(hostdev) => self.add_vfio_device(&hostdev).context("add vfio device"),
             DeviceType::Block(block) => self
@@ -84,6 +84,15 @@ impl DragonballInner {
         info!(sl!(), "remove device {} ", device);
 
         match device {
+            DeviceType::Network(network) => {
+                // Dragonball doesn't support remove network device, just print message.
+                info!(
+                    sl!(),
+                    "dragonball remove network device: {:?}.", network.config.virt_iface_name
+                );
+
+                Ok(())
+            }
             DeviceType::Block(block) => {
                 let drive_id = drive_index_to_id(block.config.index);
                 self.remove_block_drive(drive_id.as_str())
@@ -197,9 +206,9 @@ impl DragonballInner {
         Ok(())
     }
 
-    fn add_net_device(&mut self, config: &NetworkConfig, device_id: String) -> Result<()> {
+    fn add_net_device(&mut self, config: &NetworkConfig) -> Result<()> {
         let iface_cfg = VirtioNetDeviceConfigInfo {
-            iface_id: device_id,
+            iface_id: config.virt_iface_name.clone(),
             host_dev_name: config.host_dev_name.clone(),
             guest_mac: match &config.guest_mac {
                 Some(mac) => MacAddr::from_bytes(&mac.0).ok(),
--- a/src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
+++ b/src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
@@ -6,11 +6,12 @@
 
 #[cfg(test)]
 mod tests {
-    use anyhow::Context;
+    use std::{fs, path::Path, sync::Arc};
+
+    use anyhow::{anyhow, Context, Result};
     use netlink_packet_route::MACVLAN_MODE_PRIVATE;
     use scopeguard::defer;
-
-    use std::sync::Arc;
+    use tokio::sync::RwLock;
 
     use crate::network::{
         endpoint::{IPVlanEndpoint, MacVlanEndpoint, VlanEndpoint},
@@ -22,6 +23,41 @@ mod tests {
         network_pair::{NetworkInterface, NetworkPair, TapInterface},
         utils::link::net_test_utils::delete_link,
     };
+    use hypervisor::{device::device_manager::DeviceManager, qemu::Qemu};
+    use kata_types::config::{QemuConfig, TomlConfig};
+
+    async fn get_device_manager() -> Result<Arc<RwLock<DeviceManager>>> {
+        let path = env!("CARGO_MANIFEST_DIR");
+        let path = Path::new(path)
+            .join("../../../libs/kata-types/tests/texture/configuration-anno-0.toml");
+
+        let content = fs::read_to_string(path).context("read configuration failed")?;
+        // just for test, use x/kata-types/tests/texture/configuration-anno-0.toml as
+        // the test configuration.toml which is for qemu.
+        let hypervisor_name: &str = "qemu";
+
+        let qemu = QemuConfig::new();
+        qemu.register();
+
+        let toml_config = TomlConfig::load(&content).context("load toml config failed")?;
+        let hypervisor_config = toml_config
+            .hypervisor
+            .get(hypervisor_name)
+            .ok_or_else(|| anyhow!("failed to get hypervisor for {}", &hypervisor_name))?;
+
+        let mut hypervisor = Qemu::new();
+        hypervisor
+            .set_hypervisor_config(hypervisor_config.clone())
+            .await;
+
+        let dm = Arc::new(RwLock::new(
+            DeviceManager::new(Arc::new(hypervisor))
+                .await
+                .context("device manager")?,
+        ));
+
+        Ok(dm)
+    }
 
     // this unit test tests the integrity of MacVlanEndpoint::new()
     #[actix_rt::test]
@@ -32,6 +68,10 @@ mod tests {
         let tap_iface_name = format!("tap{}_kata", idx); // create by NetworkPair::new()
         let dummy_name = format!("dummy{}", idx);
         let vlanid = 123;
+
+        let dm = get_device_manager().await;
+        assert!(dm.is_ok());
+        let d = dm.unwrap();
 
         if let Ok((conn, handle, _)) =
             rtnetlink::new_connection().context("failed to create netlink connection")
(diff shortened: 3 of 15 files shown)


src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -20,20 +28,106 @@ impl fmt::Debug for Address {
...
+        }
+
+        Ok(Some(self.config.index))
+    }
+
+    async fn get_device_info(&self) -> DeviceType {
+        DeviceType::Network(self.clone())
+    }
+
+    async fn increase_attach_count(&mut self) -> Result<bool> {
```
Ji-Xinyou, line 104: You should add comments on the semantic of this return value, i.e. stating that returning false means we should do real attachment.

Ji-Xinyou: Moreover, imo we should only increase the attach count here, we could check the attach count in the `attach()`, if it is one, we let the VMM attach that.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -20,20 +28,106 @@ impl fmt::Debug for Address {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct NetworkConfig {
+    pub index: u64,
```
Ji-Xinyou, line 33: Add a one line comment doc for this field please.

src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
```diff
@@ -84,6 +84,9 @@ impl DragonballInner {
         info!(sl!(), "remove device {} ", device);
 
         match device {
+            DeviceType::Network(_network) => {
+                todo!()
```
Ji-Xinyou, line 88: If dbs support this we should implement this, since this is called on `detach()`. Otherwise it would panic here.

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -33,6 +65,8 @@ mod tests {
         let dummy_name = format!("dummy{}", idx);
         let vlanid = 123;
 
+        let d = get_device_manager().await.unwrap();
```
Ji-Xinyou, line 68: Although this is a test, I prefer not to use any abbreviation in our code.

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -144,6 +179,7 @@ mod tests {
         let tap_iface_name = format!("tap{}_kata", idx); // create by NetworkPair::new()
         let model_str = TC_FILTER_NET_MODEL_STR;
         let dummy_name = format!("dummy{}", idx);
+        let d = get_device_manager().await.unwrap();
```
Ji-Xinyou, line 182: ditto

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -267,6 +305,7 @@ mod tests {
         let mac_addr = String::from("02:00:CA:FE:00:04");
         let manual_virt_iface_name = format!("eth{}", idx);
         let tap_iface_name = format!("tap{}_kata", idx); // create by kata
+        let d = get_device_manager().await.unwrap();
```
Ji-Xinyou, line 308: ditto

src/runtime-rs/crates/resource/src/network/endpoint/ipvlan_endpoint.rs
```diff
@@ -4,37 +4,54 @@
...
+    endpoint_persist::{EndpointState, IpVlanEndpointState},
+    Endpoint,
+};
+use crate::network::{network_model::TC_FILTER_NET_MODEL_STR, utils, NetworkPair};
 
 // IPVlanEndpoint is the endpoint bridged to VM
 #[derive(Debug)]
 pub struct IPVlanEndpoint {
     pub(crate) net_pair: NetworkPair,
+    pub(crate) d: Arc<RwLock<DeviceManager>>,
```
Ji-Xinyou, line 35: ditto

Ji-Xinyou: By the way, is it possible to decouple the device manager's reference with the device? Or it is just the practice of using this.

Ji-Xinyou (commented):
In general this is a great work and LGTM! A few comments below.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -20,20 +28,106 @@ impl fmt::Debug for Address {
...
+        return Ok(());
+    }
+
+    async fn detach(&mut self, h: &dyn hypervisor) -> Result<Option<u64>> {
+        if self
+            .decrease_attach_count()
+            .await
+            .context("failed to decrease attach count")?
+        {
+            return Ok(None);
```
Tim-Zhang, line 89: It would be better if you leave some comments to explain why it returned early here.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -20,20 +28,106 @@ impl fmt::Debug for Address {
...
+    }
+
+    async fn increase_attach_count(&mut self) -> Result<bool> {
+        match self.attach_count {
+            0 => {
+                // do real attach
+                self.attach_count += 1;
+                Ok(false)
+            }
+            std::u64::MAX => Err(anyhow!("device was attached too many times")),
```
Tim-Zhang, line 111: is `std::` required here?

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -20,20 +28,106 @@ impl fmt::Debug for Address {
...
+            _ => {
+                self.attach_count += 1;
+                Ok(true)
+            }
+        }
+    }
+
+    async fn decrease_attach_count(&mut self) -> Result<bool> {
```
Tim-Zhang, line 119: in my opinion, returning a bool is very confusing, why don't we just return the attach_count and leave the conditional logic to callers?

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -22,6 +23,37 @@ mod tests {
...
         utils::link::net_test_utils::delete_link,
     };
+    use hypervisor::{device::device_manager::DeviceManager, qemu::Qemu};
+    use kata_types::config::{QemuConfig, TomlConfig};
+
+    async fn get_device_manager() -> Result<Arc<RwLock<DeviceManager>>> {
+        let path = env!("CARGO_MANIFEST_DIR");
+        let path = Path::new(path)
+            .join("../../../libs/kata-types/tests/texture/configuration-anno-0.toml");
+        let content = fs::read_to_string(path).unwrap();
```
Tim-Zhang, line 33: use `?` instead?

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -22,6 +23,37 @@ mod tests {
...
+        let hypervisor_name: &str = "qemu";
```
Tim-Zhang, line 34: Is type declaration required here？and what about converting it to a const?

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -22,6 +23,37 @@ mod tests {
...
+
+        let qemu = QemuConfig::new();
+        qemu.register();
+
+        let toml_config = TomlConfig::load(&content).unwrap();
```
Tim-Zhang, line 39: better to avoid unwrap

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -22,6 +23,37 @@ mod tests {
...
+        let hypervisor_config = toml_config
+            .hypervisor
+            .get(hypervisor_name)
+            .ok_or_else(|| anyhow!("failed to get hypervisor for {}", &hypervisor_name))
```
Tim-Zhang, line 43: error message duplicate here, ok_or_else is enough

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
```diff
@@ -22,6 +23,37 @@ mod tests {
...
+            .context("get hypervisor")?;
+
+        let mut hypervisor = Qemu::new();
+        hypervisor
+            .set_hypervisor_config(hypervisor_config.clone())
+            .await;
+
+        let dm = Arc::new(RwLock::new(
+            DeviceManager::new(Arc::new(hypervisor)).await.unwrap(),
```
Tim-Zhang, line 52: better to avoid unwrap

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: As a result of so many similar increase/decrease operations, I think it's better to do so as it is.
And the `Result<bool>`  seems not so good, but the refactoring work is coming soon.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: More detailed info about Device::Trait in[de]crease_attch_count have been explained in drvier/mod.rs.

src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
Apokleos: fixed

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: fixed

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -20,20 +28,117 @@ impl fmt::Debug for Address {
...
+            config: config.clone(),
+        }
+    }
+}
+
+#[async_trait]
+impl Device for NetworkDevice {
+    async fn attach(&mut self, h: &dyn hypervisor) -> Result<()> {
+        if self
+            .increase_attach_count()
```
Ji-Xinyou, line 73: The logic here is a little bit messy. The attach could be called **MULTIPLE** times, but the logic here will return **ERROR** if we are increaing attach count that is more than zero.

I would suggest a logic like this, cc @Tim-Zhang 

```Rust
  async fn attach(&mut self, h: &dyn hypervisor) -> Result<()> {
    self.increase_attach_count().await.context(".....")?;
    if self.attach_count == 1 {
        if let Err(e) = h.add_device(DeviceType::Network(self.clone())).await {
            self.decrease_attach_count().await?;
            Err(e)
        }
    }

     Ok(());
  }
```

And the increase attach count logic should look like
```Rust
async fn increase_attach_count(&mut self) -> Result<()> {
       if let self.attach_count == MAX {
          return Err(anyhow!("device was attached too many times"));
      }
     
    self.attach_count += 1;
    Ok(())
}

Comment (Apokleos) [reply src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs:73]: Hi @Ji-Xinyou, sorry for the change, my mistake for it.

But I don't think it's a good idea to rewrite the increase_attach_count as you said. 
And it's designed to do refcount and return true or false, then tell attach() whether do attach or do skip.
There's no need to refer`self.attach_count ==1`.

File: src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs [hunk 17]
```diff
@@ -20,20 +28,119 @@ impl fmt::Debug for Address {
...
+        }
+
+        Ok(Some(self.config.index))
+    }
+
+    async fn get_device_info(&self) -> DeviceType {
+        DeviceType::Network(self.clone())
+    }
+
+    async fn increase_attach_count(&mut self) -> Result<bool> {
```
Tim-0731-Hzt, line 113: IMO `attach_count` is unnecessary for the network interface because it won't be attached multiple times. Maybe consider leaving this trait blank, wdyt

src/runtime-rs/crates/resource/src/network/network_with_netns.rs
```diff
@@ -82,11 +82,11 @@ impl NetworkWithNetns {
 
 #[async_trait]
 impl Network for NetworkWithNetns {
-    async fn setup(&self, h: &dyn Hypervisor) -> Result<()> {
+    async fn setup(&self, _h: &dyn Hypervisor) -> Result<()> {
```
Tim-0731-Hzt, line 85: if this argument is not needed, I think you can just remove it.

Tim-0731-Hzt (commented):
@Apokleos Thanks for your work!

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: Thx @Tim-0731-Hzt , it will be fixed.

src/runtime-rs/crates/resource/src/network/network_with_netns.rs
Apokleos: good idea!

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: fixed

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
```diff
@@ -84,6 +84,15 @@ impl DragonballInner {
         info!(sl!(), "remove device {} ", device);
 
         match device {
+            DeviceType::Network(network) => {
+                // FIXME: now Dragonball not support remove network device, we only print some message.
+                info!(
+                    sl!(),
+                    "dragonball remove network device: {:?}.", network.config.virt_iface_name
```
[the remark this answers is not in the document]
Apokleos: fixed

Tim-0731-Hzt (approved):
Thanks @Apokleos LGTM

quanweiZhou (approved):
Thanks, @Apokleos, LGTM.

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
Apokleos: fixed

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
Apokleos: fixed

src/runtime-rs/crates/resource/src/network/endpoint/endpoints_test.rs
Apokleos: just for test case.

Issue #6720, opened 2023-04-25
Configure proxy information for containerd service as part of kata-manager setup script.


amshinde:
When the http_proxy is set, containerd service file needs to be configured with this proxy in order for `ctr` to be able to download container images. 
This configuration in the containerd systemd service (typically at `/etc/systemd/system/containerd` or `/usr/lib/systemd/system/containerd` depending on the distribution)would look like:
```
[Service]
ExecStartPre=/sbin/modprobe overlay
ExecStart=/usr/local/bin/containerd
Restart=always
Environment="HTTP_PROXY=http://9.21.61.141:3128/"
```

Take a look at : https://github.com/containerd/containerd/issues/1990#issuecomment-403414959 for reference.

PR #7220, opened 2023-07-03
kata-manager: Configure proxy information for containerd service as part of kata-manager setup script

nasrulirfan:
In containerd, when a user wants to set a proxy to download/pull an image, we want to add an environment variable in the containerd systemd service (typically at `/etc/systemd/system/containerd`) as shown below: 

```
[Service]
ExecStartPre=/sbin/modprobe overlay
ExecStart=/usr/local/bin/containerd
Restart=always
Environment="HTTP_PROXY=http://your-proxy/"
```
In this PR, we configure the proxy information for containerd service to be in part of kata-manager setup script by adding the `-p` argument.
User who wants to set a proxy information can pass the argument '-p <HTTP_PROXY>'. For example
`bash kata-manager.sh -p "http://9.21.61.141:3128"`
The HTTP PROXY value then will be passed to the configure_containerd() function. If existing HTTP Proxy exists, it will overwrite with the new proxy value. Otherwise, it will set new HTTP proxy to the containerd Environment under the [Service] category. Refer as below:
https://github.com/nasrulirfan/kata-containers/blob/8621355e0ae6cf3ff8fe41ea8e61576d85cc3d75/utils/kata-manager.sh#L514-L519

Fixes: #6720

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -408,6 +408,11 @@ configure_containerd()
 	local enable_debug="${1:-}"
 	[ -z "$enable_debug" ] && die "no enable debug value"
 
+	local http_proxy_set="${2:-}"
+	[ -z "$http_proxy_set" ] && die "no http_proxy_set value"
+
+	local http_proxy="${3:-}"
+
 	local project="$containerd_project"
 
 	info "Configuring $project"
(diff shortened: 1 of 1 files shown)


utils/kata-manager.sh
```diff
@@ -408,6 +408,11 @@ configure_containerd()
 	local enable_debug="${1:-}"
 	[ -z "$enable_debug" ] && die "no enable debug value"
 
+	local http_proxy_set="${2:-}"
```
liubin, line 411: Can reduce `http_proxy_set` and `http_proxy` into one? IMO a non-empty `http_proxy` equals `http_proxy_set=true`

nasrulirfan: Agreed. Pushed the changes.

cmaf:
Hi @nasrulirfan, could you follow @amshinde's request to squash the commits and follow our [patch format guidelines](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format)?

Issue #7221, opened 2023-07-03
static-checks-dragonballs sometimes fails in CI.


jepio:
The job has two issues:
- it wants to run on a self-hosted runner which now occasionally falls onto our amd-sev node 
- it installs cargo but doesn't correctly propagate the PATH update to subsequent steps so it fails unless rust/cargo is already installed on the self-hosted node

An example of a failing run is here: https://github.com/kata-containers/kata-containers/actions/runs/5445837890/jobs/9905672573
```
/home/kata/actions-runner/_work/_temp/8a2cfcdd-1393-447a-8bd5-79d152b07ebb.sh: line 2: cargo: command not found
Error: Process completed with exit code 127.
```

PR #7222, opened 2023-07-03
gha: dragonball: Correctly propagate PATH update

jepio:
cargo/rust is installed in one step, we need to write the PATH update to GITHUBENV so that it becomes visible in the next steps.

Fixes: #7221


fidencio (approved):
lgtm, thanks @jepio!

GabyCT (approved):

Issue #7223, opened 2023-07-03
release: The release workflow cannot download hub


fidencio:
```sh
curl -s "https://api.github.com/repos/github/hub/releases/latest" 
{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/401025/releases/latest",
  "documentation_url": "https://docs.github.com/v3/#http-redirects"
}
```

Which causes failures during the release workflow.

PR #7224, opened 2023-07-03
gha: release: Use a specific release of hub

fidencio:
ideally we should never ever use hub again, and switch to a supported / release tool instead.  However, in order to get v3.1.3 released, let's just stick to the last released version of hub, as trying to get its release is leading to:
```
curl -s "https://api.github.com/repos/github/hub/releases/latest"
{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/401025/releases/latest",
  "documentation_url": "https://docs.github.com/v3/#http-redirects"
}
```

And that breaks the release process. :-/

Fixes: #7223

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -72,8 +72,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: install hub
         run: |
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
Looks plausible to me!

bergwolf (approved):

Issue #701, opened 2020-09-10
docs: update the sandbox api doc in the kata-api-design.md


YchauWang:
Sync the sandbox api doc from the 2.0-dev. Delete and add somes apis from the kata-api-design.md doc. Add new table for "Sandbox Monitor API"

1.x issue: https://github.com/kata-containers/documentation/issues/727
1.x patch: https://github.com/kata-containers/documentation/pull/728

PR #7226, opened 2023-07-04
debug: Force image cache miss

stevenhorsman:
Hard-code last commit to force cache miss in image for testing

Fixes: #701


stevenhorsman:
We think the bug has been resolved now, so this can close

PR #7227, opened 2023-07-04
gha: aks: Wait longer to start running the tests, again

fidencio:
We're still trying to run the tests when Kata Containers is not fully deployed in the AKS cluter.  Let's increase the sleeping time yet again, this time to 5 minutes.

Fixes: #6964


sprt:
Yep I don't think increasing the timeout will work. Actually I believe `kubectl -n kube-system wait --timeout=10m --for=condition=Ready -l name=kata-deploy pod` is indeed waiting till everything has been copied, but there might be something under the hood that still hasn't been deployed fully at that point. I want to do a deep dive in this once I find the time.

PR #7231, opened 2023-07-04
Build for measured rootfs improvements

wainersm:
While helping to solve https://github.com/kata-containers/tests/issues/5701 I looked more carefully at the kernel, image and shim-v2 build scripts, and I found some opportunities of improvements. Apart from that, we've got some plans for measured rootfs on main branch as #7415.

This PR has the goals:

 * move the logic to build measured rootfs out of the kata-deploy. IMO the component's builders should handle the build paths for enabling measured rootfs
 *  ensure that when `MEASURED_BOOTFS=yes` then the expected files are found or the build process die. This is to ensure we don't silently produce bad components
 * implements one test case for current implementation


wainersm:
I set `needs-backport` in case @stevenhorsman prefers this out of the next merge PR. Of course, if those changes work and are approved :)

tools/packaging/static-build/shim-v2/build.sh
```diff
@@ -23,6 +23,19 @@ container_image="${SHIM_V2_CONTAINER_BUILDER:-$(get_shim_v2_image_name)}"
 
 EXTRA_OPTS="${EXTRA_OPTS:-""}"
 
+if [ "${MEASURED_ROOTFS}" == "yes" ]; then
+	EXTRA_OPTS+=" DEFSERVICEOFFLOAD=true"
+	info "Enable rootfs measurement config"
+
+	root_hash_file="${repo_root_dir}/tools/osbuilder/root_hash.txt"
+	[ -f "$root_hash_file" ] || \
```
stevenhorsman, line 31: I guess for the near future (and in `CCv0` now) we might have `${repo_root_dir}/tools/osbuilder/root_hash_tdx.txt` for TDX instead of `root_hash.txt`, so we'll need to update the logic to fail if none of the root hash files are found then.

stevenhorsman (approved):
The code looks good to me, but we don't currently have a way to test it in `main` which makes me a little nervous, but then again we don't have a code path that activates it either, so it's probably ok to have code review only until it's merged into `CCv0`, though I will need to update it to handle TDX root hashes then too.

wainersm:
> The code looks good to me, but we don't currently have a way to test it in `main` which makes me a little nervous, but then again we don't have a code path that activates it either, so it's probably ok to have code review only until it's merged into `CCv0`, though I will need to update it to handle TDX root hashes then too.

Yeah, I was expecting it being partially tested by the new TDX job on GHB but it seems disabled?

This is how I tested it:

* Build the vanilla kernel with support to measured rootfs:

```
make MEASURED_ROOTFS=yes kernel-tarball
```

* Build the TDX kernel with support to measured rootfs (** `MEASURED_ROOTFS=yes` is exported in https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L318):
```
make kernel-tdx-experimental-tarball
```

* Build shim-v2 with support to measured rootfs:
```
cd tools/osbuilder
wget http://jenkins.katacontainers.io/job/kata-containers-2.0-rootfs-image-cc-x86_64/lastSuccessfulBuild/artifact/artifacts/root_hash_vanilla.txt
mv root_hash_vanilla.txt root_hash.txt
cd -
make MEASURED_ROOTFS=yes shim-v2-tarball
cat build/shim-v2/destdir/opt/kata/share/defaults/kata-containers/configuration.toml | grep hash
```

* Build shim-v2 with support to measured rootfs but force a fail:
```
rm -f tools/osbuilder/root_hash.txt
make MEASURED_ROOTFS=yes shim-v2-tarball
```

* Build the TDX kernel with support to measured rootfs but force a fail:
```
rm -rf tools/packaging/kernel/initramfs.cpio.gz
# Comment the line https://github.com/kata-containers/kata-containers/blob/b2b70e6d8a5880db4655ace6b981d5ce48acc07d/tools/packaging/static-build/kernel/build.sh#L25 so that initramfs is not built
make kernel-tdx-experimental-tarball
```

tools/packaging/static-build/shim-v2/build.sh
stevenhorsman: Yeah - I think you are right that doing the backport straight away (well after we have the CCv0 CI stable 😭 ) is the best approach rather than have it mixed in with the rest of the merged changes

wainersm:
hi @arronwy , could I have your review on this one?

wainersm:
Rebased and fixed the conflicts. Apparently the logic is still working even with the new cache mechanism.

Leaving it WIP as I plan to address the last @stevenhorsman 's suggestions.

wainersm:
A pod with following annotation should be failing the measurement as I am passing an incorrect hash, but it is not:
```
apiVersion: v1
kind: Pod
metadata:
   name: busybox-cc
   annotations:
     io.katacontainers.config.hypervisor.kernel_params: "rootfs_verity.scheme=dm-verity rootfs_verity.hash=5180b1568c2ba972e4e06ee0a55976acae8329f2a5d1d2004395635e1ec4a76e"
 spec:
   runtimeClassName:
   containers:
     - name: nginx
       image: nginx
       imagePullPolicy: Always
   imagePullSecrets:
     - name: cococred
```

Am I passing the right annotation @stevenhorsman ?

wainersm:
> A pod with following annotation should be failing the measurement as I am passing an incorrect hash, but it is not:
> 
> ```
> apiVersion: v1
> kind: Pod
> metadata:
>    name: busybox-cc
>    annotations:
>      io.katacontainers.config.hypervisor.kernel_params: "rootfs_verity.scheme=dm-verity rootfs_verity.hash=5180b1568c2ba972e4e06ee0a55976acae8329f2a5d1d2004395635e1ec4a76e"
>  spec:
>    runtimeClassName:
>    containers:
>      - name: nginx
>        image: nginx
>        imagePullPolicy: Always
>    imagePullSecrets:
>      - name: cococred
> ```
> 
> Am I passing the right annotation @stevenhorsman ?

Just noticed that `runtimeClassName` is blank....

stevenhorsman:
> Am I passing the right annotation @stevenhorsman ?

Yes - that looks like it matches the kata-deploy method that adds it when measured_rootfs is enabled at the moment: https://github.com/kata-containers/kata-containers/blob/29d863350f40ecb5f1f885c5f37c22d8df6c7b12/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L588

> Just noticed that `runtimeClassName` is blank....

Yeah, that's worth looking at more 😄

wainersm:
@stevenhorsman apart from the needed to polish the commit messages/code, with this last update I close the scope of this PR. The remaining work for https://github.com/kata-containers/kata-containers/issues/7415 I will send on other PRs.

I'm leaving the `assert_logs_contain()` function to show you an interesting problem we've got: the `test_start_time` is no longer accurate to be used on  `assert_logs_contain()` because it hold the local time, not the worker time. i.e. we need to figure out the time where the pod started on the worker. I was considering to use the `creationTime` property of the pod but I read it is the timestamp created by the API server but workers and controllers might be on different timezones as well. An alternative is to connect at one random worker, get the current time, then assume it is the `test_start_time` (but then I assume all workers have the clock in sync, which I am not sure is true). Anyway, interesting stuff...I will figure out an way to make it work without over-engineering, let me know if you got ideas.

Ah, I cannot properly test this PR because it is blocked by https://github.com/kata-containers/kata-containers/issues/8318

stevenhorsman:
Thanks, it's on my list to do a review of this, but please nudge me once the blockers are removed if I haven't had a chance as the list is quite long and fast growing!

wainersm:
The test that I am adding on this PR passed (https://github.com/kata-containers/kata-containers/actions/runs/6670016832/job/18130811556?pr=7231):

```
ok 1 Test cannnot launch pod with measured boot enabled and incorrect hash
1..2
```

Removing the wip as it is ready to review.

stevenhorsman (approved):
Apart from a couple of minor issue this looks great. Thanks @wainersm!

tools/packaging/kernel/build-kernel.sh
```diff
@@ -538,7 +539,7 @@ install_kata() {
 }
 
 main() {
-	while getopts "a:b:c:deEfg:hH:k:p:t:u:v:x:" opt; do
+	while getopts "a:b:c:deEfg:hH:mk:p:t:u:v:x:" opt; do
```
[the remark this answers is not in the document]
wainersm: Fixed!

wainersm:
Rebased this PR to solve a small conflict and addressed the last comments of @stevenhorsman

fidencio (approved):
lgtm, thanks @wainersm!

Issue #7233, opened 2023-07-05
CC: kata-deploy: Adapt the SNP's QEMU binary name


fidencio:
SNP's QEMU has changed its name some time ago and, due to that, we have
been leaving the new binary behind during the uninstall process, which
lead to the Operator hanging when uninstalling.

PR #7234, opened 2023-07-05
cc: kata-deploy: Adapt the SNP's QEMU name

fidencio:
SNP's QEMU has changed its name some time ago and, due to that, we have been leaving the new binary behind during the uninstall process, which lead to the Operator hanging when uninstalling.

Fixes: #7233

### DIFF
--- a/tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh
@@ -272,7 +272,7 @@ function remove_artifacts() {
 		/opt/confidential-containers/bin/kata-runtime \
 		/opt/confidential-containers/bin/kata-collect-data.sh \
 		/opt/confidential-containers/bin/qemu-system-x86_64 \
-		/opt/confidential-containers/bin/qemu-system-x86_64-snp \
+		/opt/confidential-containers/bin/qemu-system-x86_64-snp-experimental \
 		/opt/confidential-containers/bin/qemu-system-x86_64-tdx \
 		/opt/confidential-containers/bin/qemu-system-s390x \
 		/opt/confidential-containers/bin/cloud-hypervisor \


tools/packaging/kata-deploy-cc/scripts/kata-deploy.sh
```diff
@@ -272,7 +272,7 @@ function remove_artifacts() {
 		/opt/confidential-containers/bin/kata-runtime \
 		/opt/confidential-containers/bin/kata-collect-data.sh \
 		/opt/confidential-containers/bin/qemu-system-x86_64 \
-		/opt/confidential-containers/bin/qemu-system-x86_64-snp \
+		/opt/confidential-containers/bin/qemu-system-x86_64-snp-experimental \
```
stevenhorsman, line 275: Could we just take the same approach as `main` does:
```
function remove_artifacts() {
    echo "deleting kata artifacts"
    rm -rf /opt/kata/
}
```
and delete the whole directory, or does the pre-install & post-uninstall using it cause issues with that?

fidencio: The problem on deleting the whole directory is that we'll try to delete the `containerd` binary that's also located there, and any other binary that folks would end up putting there as part of their modified pre-install / post-uninstall containers. :-/

fidencio: But one thing that can be done, but not as part of this PR, is a test to ensure that everything that's part of the image is being into account to be cleaned up ... that would not be hard to implement and that would be a very nice way to provent it happening in the future.

fidencio: Partially.

When we get rid of the containerd fork, **I imagine** that we'll still be relying on a speciifc version of containerd, which will have the patches proposed by the community in order to better support CoCo.  If that's really the case, we'll  have to keep shipping that minimal version of containerd for a reasonable amount of time (preferrably till it gets EOD and we don't have to care about folks using a non-supported version of containerd anymore).

stevenhorsman (approved):
LGTM. Thanks!

Issue #7235, opened 2023-07-05
CC: gha: workflows are not generated the root_hash files needed for measured rootfs


fidencio:
Since the work's been merged into main, and then backported to the `CCv0` branch, we've overlooked into the gha and missed adding the `MEASURED_ROOTFS=yes` env-var when building vanilla (for x86_64, s390x) and TDX kernels.

fidencio:
Fixed by https://github.com/kata-containers/kata-containers/pull/10466

PR #7237, opened 2023-07-05
cc: gha: Pass MEASURED_ROOTFS to the artefacts build

fidencio:
Since the measured rootfs work has been merged to main, and then brought in to the CCv0 via the weekly merge, we have introduced a few regressions related to how we build it / use it.

This PR attempts to make sure the artefacts are properly built, using GitHub Actions, so the feature can be used with the operator.

Fixes: #7235

### DIFF
--- a/.github/workflows/cc-payload-after-push-amd64.yaml
+++ b/.github/workflows/cc-payload-after-push-amd64.yaml
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        measured_rootfs:
+          - no
         asset:
           - cc-cloud-hypervisor
           - cc-kernel
@@ -27,6 +29,11 @@ jobs:
           - cc-tdx-qemu
           - cc-tdx-td-shim
           - cc-tdx-tdvf
+        include:
+          - measured_rootfs: yes
+            asset:
+              - cc-kernel
+              - cc-tdx-kernel
     steps:
       - name: Login to Kata Containers quay.io
         uses: docker/login-action@v2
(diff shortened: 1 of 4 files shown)


fidencio:
And we're back at measured rootfs tests failing for non-TEE. :-/

stevenhorsman:
> And we're back at measured rootfs tests failing for non-TEE. :-/

I think this is due to the shim cache jobs needing measured_rootfs set as well. Updating that now and will then re-run the tests

.github/workflows/cc-payload-after-push-amd64.yaml
```diff
@@ -27,6 +29,11 @@ jobs:
           - cc-tdx-qemu
           - cc-tdx-td-shim
           - cc-tdx-tdvf
+        include:
+          - measured_rootfs: yes
+            asset:
+              - cc-kernel
+              - cc-tdx-kernel
```
wainersm, line 36: It doesn't change the outcome but the cc-tdx-kernel is already built with `MEASURED_ROOTFS=yes` (it is explicitly set in https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L763). I am in favor of removing that export from kata-deploy-binaries.sh (can be in a follow up PR).

wainersm (approved):

.github/workflows/cc-payload-after-push-amd64.yaml
fidencio: I will do in a follow-up PR, after making sure the image is actually generated as we want.

stevenhorsman:
measured rootfs tests working again now :)

Issue #6864, opened 2023-05-17
tests: Ensure pods created for our tests are under a specific namespace


fidencio:
This will help to cleanup possible leftovers, by just deleting all the pods from the `kata-tests` namespace, for instance.

fidencio:
Some reference material: https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/

wainersm:
I think it is yet another good idea. Test frameworks like k8s e2e generates temporary namespaces for each test so that they are isolated of each other. We don't necessarily need to generate on namespace per test, but using a random non-default for the whole suite would be good.

amshinde:
Sounds like a good idea to me. 
+1

sprt:
I can work on this as we talked about.

PR #7238, opened 2023-07-05
gha: k8s: Ensure tests are running on a specific namespace

fidencio:
Let's make sure we run our tests in a specific namespace, as in case of
any kind of issue, we will just get rid of the namespace itself, which
will take care of cleaning up any leftover from failing tests.

One important thing to mention is why we can get rid of the `namespace:
${namespace}` on the tests that are already using it, and let's do it in
parts:
* namespace: default
  We can easily get rid of this as that's the default namespace where
  pods are created, so it was a no-op so far.
* namespace: test-quota-ns
  My understanding is that we'd need this in order to get a clean
  namespace where we'd be setting a quota for.  Doing this in the
  namespace that's only used for tests should **not** cause any
  side-effect on the tests, as we're running those in serial and there's
  no other pods running on the `kata-containers-k8s-tests` namespace

Last but not least, we're not dynamically creating namespaces as the
tests are not running in parallel, **never**, not in the case of having
2 tests being ran at same time, neither in the case of having 2 jobs
being scheduled to the same machine.

Fixes: #6864

### DIFF
--- a/tests/integration/kubernetes/runtimeclass_workloads/tests-namespace.yaml
+++ b/tests/integration/kubernetes/runtimeclass_workloads/tests-namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kata-containers-k8s-tests
--- a/tests/integration/gha-run.sh
+++ b/tests/integration/gha-run.sh
@@ -58,6 +58,12 @@ function get_cluster_credentials() {
 function run_tests() {
     platform="${1}"
 
+    # Emsure we're in the default namespace
+    kubectl config set-context --current --namespace=default
+
+    # Delete any spurious tests namespace that was left behind
+    kubectl delete namespace kata-containers-k8s-tests &> /dev/null || true
+
     sed -i -e "s|quay.io/kata-containers/kata-deploy:latest|${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}|g" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
     cat "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
     cat "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" | grep "${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}" || die "Failed to setup the tests image"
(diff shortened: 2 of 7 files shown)


fidencio:
I've taken the simplistic approach and I'm just creating one namespace that will be well known.  The reason for that is because we never will be running tests in parallel as part of our TEE machines.

wainersm:
We've got a couple of pods that leverage `default` or a custom namespace. The `resources-quota` seems to make sense to leverage its own namespace, it should be setting limits:

```
$ grep namespace tests/integration/kubernetes/runtimeclass_workloads/*
tests/integration/kubernetes/runtimeclass_workloads/pod-custom-dns.yaml:  namespace: default
tests/integration/kubernetes/runtimeclass_workloads/pod-oom.yaml:  namespace: default
tests/integration/kubernetes/runtimeclass_workloads/pod-quota-deployment.yaml:  namespace: test-quota-ns
tests/integration/kubernetes/runtimeclass_workloads/resource-quota.yaml:    namespace: test-quota-ns
```

Does k8s interpret "default" as the current namespace?

Other than that, I got a question. Is there a reason to create this namespace at the GHB level rather than, for example, on `tests/integration/kubernetes/run_kubernetes_tests.sh`? Putting on the later file we can at least test the test code on a local machine.

tests/integration/kubernetes/runtimeclass_workloads/tests-namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kata-containers-k8s-tests
```
wainersm, line 4: If I am not asking too much, the namespace name could be random :)

fidencio: Why? :-)

I mean, it makes things more complicated without any clear benefit.

Let's say, we create a random namespace ... as we don't run any kind of tests in parallel (as in more than one execution of the same job in the same machine at the same time), why having a random namespace would be better?

wainersm: If for some unexpected termination of the job the namespace didn't get remove, which random namespace we are still protected. But I might being too paranoiac here :)

fidencio:
> Is there a reason to create this namespace at the GHB level rather than, for example, on `tests/integration/kubernetes/run_kubernetes_tests.sh`? Putting on the later file we can at least test the test code on a local machine.

The reason we do that in the gha-run.sh scripts is to ensure that the namespace will be deleted even in case of failure.  The easiest way I found to do that was by enforcing it do be done during the cleanup, but I'm open to suggestions here.

stevenhorsman (approved):
LGTM. 🤞 that it helps with the stability

wainersm (approved):
hey, thanks @fidencio !

PR #7243, opened 2023-07-05
gha: k8s: Ensure cluster doesn't exist before creating it

sprt:
The cluster cleanup step will sometimes fail to run, meaning the next run would fail in the cluster creation step. This PR addresses that.

Example: https://github.com/kata-containers/kata-containers/actions/runs/5349582743/jobs/9867845852

Fixes: #7242

### DIFF
--- a/tests/integration/gha-run.sh
+++ b/tests/integration/gha-run.sh
@@ -31,6 +31,9 @@ function login_azure() {
 }
 
 function create_cluster() {
+    # First, ensure that the cluster didn't fail to get cleaned up from a previous run.
+    delete_cluster || true
+
     az aks create \
         -g "kataCI" \
         -n "$(_print_cluster_name)" \
@@ -115,8 +118,7 @@ function delete_cluster() {
     az aks delete \
         -g "kataCI" \
         -n "$(_print_cluster_name)" \
-        --yes \
-        --no-wait
+        --yes
 }
 
 function main() {


Ji-Xinyou (approved):
Nice fix, LGTM

tests/integration/gha-run.sh
```diff
@@ -31,6 +31,9 @@ function login_azure() {
 }
 
 function create_cluster() {
+    # First, ensure that the cluster didn't fail to get cleaned up from a previous run.
+    az aks delete -g "kataCI" -n "$(_print_cluster_name)" -y || true
```
fidencio, line 35: @sprt, I like the approach, but we need to take into consideration that:
1. Sometimes we hit this because the cluster is in the process of being deleted
2. Deleting a cluster may take several minutes

Does `az aks delete` blocks till the cluster is fully deleted?  If not, we must busy-wait till the cluster is actually removed.

jepio: `delete_cluster` passes `--no-wait` which makes it non-blocking, the default should be a blocking wait.

on that note: let's extend `delete_cluster` to support passing wait/nowait and call it from here.

fidencio: Okay, then I'd be happy to get this one in with @jepio's suggestion addressed.

Thanks for the explanations @jepio and @Ji-Xinyou!

sprt: Thanks for pointing this out. As discussed offline with Fabiano, I removed `--no-wait` from `delete_cluster()` so that subsequent runs would wait for deletion of the previous cluster. I also implemented Jeremi's suggestion to just call `delete_cluster()` here.

fidencio (approved):
lgtm, thanks @sprt!

jepio (approved):

PR #7246, opened 2023-07-06
cc: gha: Avoid building twice the components set as "include"

fidencio:
Those components are being built twice, one as part of the normal matrix assets, and the second one as part of the include.

Fixes: #7235


stevenhorsman (approved):
LGTM, but I've demonstrated that I don't understand how `include` works!

fidencio:
> but I've demonstrated that I don't understand how `include` works!

We, @stevenhorsman, we.

amshinde (approved):

Issue #7247, opened 2023-07-06
gha: Add a nightly job for all the GHA jobs we're running ("Green CI baselines")


fidencio:
As done for Jenkins, let's create a nightly job that will be used to serve as a baseline job for the jobs we run on GHA.

After those are added, we'll need an extra issue to actually put them in our main github page, and folks can easily monitor them from there.

PR #7261, opened 2023-07-06
gha: ci: Avoid using env unless it's really needed

fidencio:
de83cd9de79226b15519e07871dcdee635d501c6 tried to solve an issue, but it clearly seems that I'm using env wrongly, as what ended up being passed as input was "$VAR", instead of the content of the VAR variable.

As we can simply avoid using those here, let's do it and save us a headache.

Fixes: #7247

### DIFF
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -3,9 +3,6 @@ on:
   schedule:
     - cron: '0 0 * * *'
 
-env:
-  COMMIT_HASH: ${GITHUB_REF}
-
 jobs:
   set-fake-pr-number:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


cmaf (approved):

dborquez (approved):

.github/workflows/ci-nightly.yaml
```diff
@@ -18,6 +15,6 @@ jobs:
     needs: set-fake-pr-number
     uses: ./.github/workflows/ci.yaml
     with:
-      commit-hash: $COMMIT_HASH
-      pr-number: $PR_NUMBER
-      tag: $PR_NUMBER-$COMMIT_HASH-nightly
+      commit-hash: ${GITHUB_REF}
+      pr-number: ${PR_NUMBER}
+      tag: ${PR_NUMBE}R-${GITHUB_REF}-nightly
```
GabyCT, line 20: @fidencio is ${PR_NUMBER} instead of ${PR_NUMBE}R

dborquez:
@fidencio I've tagged w/ WIP to prevent the PR from being merged with the typo found by @GabyCT

PR #7263, opened 2023-07-07
gha: ci: More follow up fixes after adding a nightly CI

fidencio:
Please, take a look at each commit message for more details.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,5 +72,5 @@ jobs:
     needs: build-kata-static-tarball-amd64
     uses: ./.github/workflows/run-metrics.yaml
     with:
-      tarball-suffix: -${{ inputs.tag }}-amd64
+      tarball-suffix: -${{ inputs.tag }}
       commit-hash: ${{ inputs.commit-hash }}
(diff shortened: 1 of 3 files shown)


jepio (approved):

.github/workflows/ci.yaml
```diff
@@ -72,5 +72,5 @@ jobs:
     needs: build-kata-static-tarball-amd64
     uses: ./.github/workflows/run-metrics.yaml
     with:
-      tarball-suffix: -${{ inputs.tag }}-amd64
+      tarball-suffix: -${{ inputs.tag }}
```
stevenhorsman, line 75: Just to check I understand this is because the tarball_suffix is used in:
```
name: kata-static-tarball-amd64${{ inputs.tarball-suffix }}
```
so the `-amd64` is already factored into the name?

fidencio: Yep, for the `tarball-suffix` case the name will always be: `kata-static-tarball-${arch}$( inputs.tarball-suffix}}

The $arch is properly set on each of the other architecture jobs, accordingly.
The mistake I made befoire was getting confused between the tarball name, which follows what I pointed out above, and what we expect for the tags (and there we need to pass the $arch)

Was I able to explain or did I make things more confusing?

stevenhorsman: Nope, that confirmed what I was expecting. Thanks!

stevenhorsman (approved):
LGTM

PR #7264, opened 2023-07-07
gha: ci: Avoid using env also in the ci-nightly and payload-after-push

fidencio:
The latter workflow is breaking as it doesn't recognise ${GITHUB_REF}, the former would most likely break as well, but it didn't get triggered yet.

The error we're facing is:
```
Determining the checkout info
  /usr/bin/git branch --list --remote origin/${GITHUB_REF}
  /usr/bin/git tag --list ${GITHUB_REF}
  Error: A branch or tag with the name '${GITHUB_REF}' could not be found
```

Fixes: #7247

### DIFF
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -4,18 +4,10 @@ on:
     - cron: '0 0 * * *'
 
 jobs:
-  set-fake-pr-number:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Set fake PR number
-        run: |
-          echo "PR_NUMBER=$(date +%Y%m%d%H%M%S)" >> "$GITHUB_ENV"
-
   kata-containers-ci-on-push:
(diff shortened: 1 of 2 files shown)


fidencio:
As the changes here are only in the GHA, we can force skip the CI.

stevenhorsman (approved):
It looks like from the docs (https://docs.github.com/en/actions/learn-github-actions/variables) that GITHUB_REF is trigger specific, so I guess isn't available on cron triggers

jepio (approved):

wainersm (approved):

wainersm:
do you know a way to test those nightly jobs before merge? Trial & error with those workflows s**** :(

jepio:
> do you know a way to test those nightly jobs before merge? Trial & error with those workflows s**** :(

The only way i know is to add a `workflow_dispatch` trigger to manually start the workflow.

Issue #7271, opened 2023-07-08
ci: k8s: dragonball: k8s-oom is failing


fidencio:
Unfortunately, we've had a few weeks with the dragonball CI not running k8s tests, and that's my bad.

Now that we're working on re-enabling it, we've had to disable the k8s-oom test is failing, as shown below:
```
not ok 1 Test OOM events for pods
# (from function `waitForProcess' in file ../../common.bash, line 62,
#  in test file k8s-oom.bats, line 26)
#   `waitForProcess "$wait_time" "$sleep_time" "$cmd"' failed
# [bats-exec-test:38] INFO: k8s configured to use runtimeclass
# pod/pod-oom created
# pod/pod-oom condition met
# Name:                pod-oom
# Namespace:           kata-containers-k8s-tests
# Priority:            0
# Runtime Class Name:  kata-dragonball
# Service Account:     default
# Node:                aks-nodepool1-19131931-vmss000000/10.224.0.4
# Start Time:          Sat, 08 Jul 2023 08:04:08 +0000
# Labels:              <none>
# Annotations:         <none>
# Status:              Failed
# IP:                  10.244.0.33
# IPs:
#   IP:  10.244.0.33
# Containers:
#   oom-test:
#     Container ID:  containerd://bf0660cdb4cbce4c86bc411[154](https://github.com/kata-containers/kata-containers/actions/runs/5491267426/jobs/10011525519#step:9:155)b5771aaabaf495e40092984fb767611d7a4a5a
#     Image:         quay.io/kata-containers/sysbench-kata:latest
#     Image ID:      quay.io/kata-containers/sysbench-kata@sha256:f0e7681c011df[157](https://github.com/kata-containers/kata-containers/actions/runs/5491267426/jobs/10011525519#step:9:158)b483bb78eb7679f9f3aec45f997070b9393802[159](https://github.com/kata-containers/kata-containers/actions/runs/5491267426/jobs/10011525519#step:9:160)e51f06e
#     Port:          <none>
#     Host Port:     <none>
#     Command:
#       /bin/sh
#     Args:
#       -c
#       sleep 2; stress --vm 2 --vm-bytes 400M --timeout 30s
#     State:          Terminated
#       Reason:       Error
#       Exit Code:    1
#       Started:      Sat, 08 Jul 2023 08:04:10 +0000
#       Finished:     Sat, 08 Jul 2023 08:04:42 +0000
#     Ready:          False
#     Restart Count:  0
#     Limits:
[...]

PR #6876, opened 2023-05-18
runtime-rs: support Memory hotplug

Tim-0731-Hzt:
support memory hotplug for runtime-rs

PR #7273, opened 2023-07-09
bugfix: plus default_memory when calculating mem size

YushuoEdge:
Otherwise, it would be easily to OOM.

Fixes: #7271

### DIFF
--- a/src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
+++ b/src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
@@ -105,7 +105,7 @@ impl InitialSizeManager {
             hv.cpu_info.default_vcpus = self.resource.vcpu as i32
         }
         if self.resource.mem_mb > 0 {
-            hv.memory_info.default_memory = self.resource.mem_mb;
+            hv.memory_info.default_memory += self.resource.mem_mb;
         }
         Ok(())
     }


src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
```diff
@@ -105,7 +105,7 @@ impl InitialSizeManager {
             hv.cpu_info.default_vcpus = self.resource.vcpu as i32
```
fidencio, line 105: Don't you need to also do the same here?

YushuoEdge: We have discussed a lot for this point in the original PR and decided not to plus the default_vcpus for now. The reasons are:

- Except for those io operations, the kata-agent process does not consume many resources when the container is running.  But in some recent discussions, we have seen that we are trying to skip kata-agent when doing io forwarder to improve performance and stability.
- Doing so will bring us more opportunities to optimize the overall resource management of the host. Including performance, resource overselling, etc. 
- The default value is very hard to manage, not only for the users but also for runtime-rs.

So let's see what happens when we do this. Do you think it's ok?

src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
```diff
@@ -105,7 +105,7 @@ impl InitialSizeManager {
...
         }
         if self.resource.mem_mb > 0 {
-            hv.memory_info.default_memory = self.resource.mem_mb;
+            hv.memory_info.default_memory += self.resource.mem_mb;
```
Tim-0731-Hzt, line 108: I think this PR might be a duplicate, as this PR already did this operation https://github.com/kata-containers/kata-containers/pull/6876, I suggest we can get this merged ASAP.

Tim-0731-Hzt: Sorry, I didn't see the message on the slack

src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
fidencio: Yes, totally okay.  Thanks!

fidencio:
@YushuoEdge, I've force-pushed to your PR as:
* I updated the commt message
* I re-enabled the k8s=oom.bats tests

Due to this, I'm running a `/test` again.

src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
quanweiZhou, line 108: Suggest you add comments, such as the Slack discussion or golang version code reference link here.

fidencio:
@YushuoEdge, can you also add this to your patch?
```
diff --git a/tests/integration/kubernetes/k8s-oom.bats b/tests/integration/kubernetes/k8s-oom.bats
index 5f45ad78a..f89b761f8 100644
--- a/tests/integration/kubernetes/k8s-oom.bats
+++ b/tests/integration/kubernetes/k8s-oom.bats
@@ -9,15 +9,11 @@ load "${BATS_TEST_DIRNAME}/../../common.bash"
 
 setup() {
-	[ "${KATA_HYPERVISOR}" == "dragonball" ] && skip "Test is currently failing, see: https://github.com/kata-containers/kata-containers/issues/7271"
-
 	pod_name="pod-oom"
 	get_pod_config_dir
 }
 
 @test "Test OOM events for pods" {
-	[ "${KATA_HYPERVISOR}" == "dragonball" ] && skip "Test is currently failing, see: https://github.com/kata-containers/kata-containers/issues/7271"
-
 	# Create pod
 	kubectl create -f "${pod_config_dir}/$pod_name.yaml"
 
@@ -33,8 +29,6 @@ setup() {
 }
 
 teardown() {
-	[ "${KATA_HYPERVISOR}" == "dragonball" ] && skip "Test is currently failing, see: https://github.com/kata-containers/kata-containers/issues/7271"
-
 	# Debugging information
 	kubectl describe "pod/$pod_name"
 	kubectl get "pod/$pod_name" -o yaml
-- 
2.41.0
```

And then, in the commit message, I'd like to have something like:
```
runtime-rs: Add default_memory when calculating mem size

We've noticed this caused regressions with the k8s-oom tests, and then
decided to take a step back and do this in the same way it was done
before 67972ec48a40434fd7c3883c4538cc1d328ede40.

And by doing this we can re-enable the k8s-oom.bats tests, which is done
as part of this PR.

Fixes: #7271
Depends-on: github.com/kata-containers/tests#5705

```

@YushuoEdge, WDYT?

fidencio:
Also, please, rebase this one atop of the current `main`.

src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs
YushuoEdge: I have added some comments here

fidencio (approved):
lgtm, thanks @YushuoEdge!

studychao (approved):

Issue #7043, opened 2023-06-06
Add Sandbox API support into Kata Containers


studychao:
This issue is supposed to be the umbrella issue for Sandbox API support in Kata.
This will track all the todos and discussion in this issue.
 
## Background 

containerd community started designing Sandbox API since 2020 ( related issue : https://github.com/containerd/containerd/issues/4131 ). Now experimental Sandbox API implementation is shipped with containerd 1.7. And it's estimated that the stable version of Sandbox API will be shipped in containerd 2.0.

Sandbox API basically changes how we create sandboxs (i.e. VM), how we create containers, how we calculate VM resource and so many different aspects of Kata Container. It's so evolutionary that we have to act fast to take deep look into it and also conclude what Kata community needs to help building this API .

### What is Sandbox?
[ from containerd ]
The sandbox concept has the following properties in relation to containers it hosts:

Sandbox acts as a parent entity for containers, e.g. it starts first and ends last (this typically useful in (micro)VM environments, where VMs are required to be started before any other entities).
Sandbox acquires resources needed for running child containers (for instance Kubernetes creates "pause" containers to acquire IP and network namespace for child containers).

### What is Sandbox API?

Sandbox API make the sandbox first class citizen, and all the sandbox management logic will not be mixed in the Task API.
For example, before Sandbox API, when we process the Task API request `CreateContainer`, we may also mix VM creation logic in that request. 

Advantages it brings:
- No more intertwined logic with containers and sandbox (VM) management.
- No more redundant pause container in Kata Containers when using Sandbox API.
- Quicker startup time and less memory consumption with simplied architecture.

#### Stage 1
![runtime-rs+non-built-in+stage 1](https://github.com/kata-containers/kata-containers/assets/31121654/a8e6d172-bfbc-419a-aa28-428ceb26ef78)
#### Stage 2
![runtime-rs+non-built-in](https://github.com/kata-containers/kata-containers/assets/31121654/2e3f3958-33bc-4461-8975-0805a7ece893)

What's new?
- Introduce the Sandboxer (temporary name) process to manage all containers on the node in a 1:N model, replacing the shim process that was present for each POD in 1:1 model. So there is only 1 process for each POD.
- By using the plugin service architecure in runtime-rs, we could introduce Sandbox Service in sandboxer process to handle sandbox related management, Task Service in sandboxer process to handle container related management.

Why 2 stages?
[...]

PR #7274, opened 2023-07-09
[RFC] Support sandbox api

wllenyj:
Sandbox API has been implemented with experimental in containerd 1.7. Now let's follow up with the containerd implementation to support Sandbox API. 

For more details in #7043.

This PR is just an RFC, comments are welcome, no merging for now.

### DIFF
--- a/src/runtime-rs/Cargo.lock
+++ b/src/runtime-rs/Cargo.lock
@@ -501,6 +501,7 @@ dependencies = [
  "oci",
  "persist",
  "protobuf 3.2.0",
+ "protobuf-codegen 3.2.0",
  "serde_json",
  "slog",
  "slog-scope",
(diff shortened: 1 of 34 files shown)


bergwolf:
I've seen a few issues when testing this PR:
* 1. containerd create sandbox request does not pass down `plugins.cri.containerd.runtimes.kata-sandbox.options`
  * looks like a containerd issue, and it causes containerd runtime handler (kata shim) failing to search for configured config.toml file path
* 2. cannot create new pods with containerd(via CRI)
* 3. shim is not cleaned up when failing to create new sandboxes
  * when failing to create new sandboxes via kubelet/CRI, there are a lot of kata shims being created w/o being cleaned up
* 4. even with ctr, only `busybox` container can be created
  * other containers (such case `ubuntu:latest`), failed as `/etc/resolv.conf` being somehow created as a directory
* 5. container stdio is broken that it only works in one-shot mode
  * `ctr run -it` would just hang
* 6. sandbox resources are not set per pod config linux.resources
  * only the default vcpu/memory resources are allocated for the sandbox

PR #7276, opened 2023-07-10
gha: ci: Gather info about the node / pods

fidencio:
This is a very simple addition, that should be expanded by
https://github.com/kata-containers/kata-containers/pull/7295, and it's
targetting gathering more info that will help us to debug CI failures.

Fixes: #7296

### DIFF
--- a/tests/integration/gha-run.sh
+++ b/tests/integration/gha-run.sh
@@ -93,6 +93,9 @@ function run_tests() {
     kubectl -n kube-system wait --timeout=10m --for=condition=Ready -l name=kata-deploy pod
     kubectl apply -f "${tools_dir}/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml"
 
+    echo "Gather information about the nodes and pods after having kata-deploy ready"
+    get_nodes_and_pods_info
+
     # This is needed as the kata-deploy pod will be set to "Ready" when it starts running,
     # which may cause issues like not having the node properly labeled or the artefacts
     # properly deployed when the tests actually start running.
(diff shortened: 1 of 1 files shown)


fidencio:
Murphy's law, I guess, but it's not been easy to reproduce the issue.

stevenhorsman (approved):
LGTM. More debug in CI is always helpful. Thanks

fidencio:
I've just verified this one works on https://github.com/kata-containers/kata-containers/actions/runs/5520011781?pr=7299

jepio (approved):

fidencio:
I've added the `do-not-merge` label here as the test failed and I'd like to try to gather information about the failure right now.

tests/integration/gha-run.sh
```diff
@@ -93,6 +93,9 @@ function run_tests() {
     kubectl -n kube-system wait --timeout=10m --for=condition=Ready -l name=kata-deploy pod
     kubectl apply -f "${tools_dir}/packaging/kata-deploy/runtimeclasses/kata-runtimeClasses.yaml"
 
+    echo "\n\nGather information about the nodes and pods after having kata-deploy ready"
```
wainersm, line 96: I'm curious about those '\n\n' in the begin. Does Github parse it differently? or you are missing the `-e` to echo?

wainersm (approved):
Thanks @fidencio ! What I said about https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines can be in another PR if you agree with the idea.

fidencio:
I know this is printing a lot of info, and maybe we can limit this later on to get all this info only when a step fails, but for now the focus is to debug the random failures starting the tests, and it's good enough for it.

PR #7277, opened 2023-07-10
gha: ci: Use github.sha to get the last commit reference

fidencio:
As we need to pass down the commit sha to the jobs that will be
triggered from the `push` / 'schedule' events, we must be careful
on what exactly we're using there.

At first we were using ${{ github.ref }}, but this turns out to be the
**branch name**, rather than the commit hash.  In order to actually get
the commit hash, Let's use ${{ github.sha }} instead.

Fixes: #7247


stevenhorsman (approved):
LGTM. Thanks!

zvonkok (approved):
LGTM, Muito obrigado!

studychao:
LGTM!

PR #7288, opened 2023-07-11
gha: nightly: Fix long name of AKS clusters issue and make the CI easier to test

fidencio:
---

gha: nightly: Fix name size limit for AKS

Passing the commit hash as the "pr-number" has shown problematic as it
would make the AKS cluster name longer than what's accepted by AKS.

One easy way to solve this is just passing "nightly" as the PR number,
as that's only used to create the cluster.

Fixes: #7247

---

gha: nightly: Also use `workflow_dispatch` to trigger it

This is a very nice suggestion from Steve Horsman, as with that we can
manually trigger the workflow anytime we need to test it, instead of
waiting for a full day for it to be retriggered via the `schedule`
event.

---

### DIFF
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -8,6 +8,6 @@ jobs:
     uses: ./.github/workflows/ci.yaml
     with:
       commit-hash: ${{ github.sha }}
-      pr-number: ${{ github.sha }}
+      pr-number: "nightly"
       tag: ${{ github.sha }}-nightly
     secrets: inherit
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -2,6 +2,7 @@ name: Kata Containers Nightly CI
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
 
 jobs:
   kata-containers-ci-on-push:


.github/workflows/ci-nightly.yaml
```diff
@@ -1,13 +1,13 @@
...
-    - cron: '0 0 * * *'
+    - cron: '0 */2 * * *'
 
 jobs:
   kata-containers-ci-on-push:
     uses: ./.github/workflows/ci.yaml
     with:
       commit-hash: ${{ github.sha }}
-      pr-number: ${{ github.sha }}
+      pr-number: "nightly"
```
stevenhorsman, line 11: This is fine, but I guess the alternative if we want to track more clear tracking of the commit tested would be to do a `cut -c1-8` or something to get the 'short' sha out

.github/workflows/ci-nightly.yaml
```diff
@@ -1,7 +1,7 @@
 name: Kata Containers Nightly CI
 on:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '0 */2 * * *'
```
stevenhorsman, line 4: Would the other option be to add `workflow_dispatch`, so we can manually build it whenever, or does that impact the trigger path too much?

fidencio: That's a very good idea, Steve!
Let me see if I can easily do that.

fidencio: Done!

.github/workflows/ci-nightly.yaml
stevenhorsman: Great, thanks for clarifying

stevenhorsman (approved):
LGTM. Thanks!

liubin (approved):

Issue #7292, opened 2023-07-11
cached jobs: It seems rootfs-initrd-mariner has always to be built


fidencio:
It slows down the CI by quite a lot.

Please, take a look at the cached jobs, create one if needed, and let us know what's happening there.

/cc @sprt @jepio

fidencio:
@jepio got it working, thus I'm closing this one.

Thanks Jeremi!

jepio:
No not yet :D

PR #7297, opened 2023-07-11
tools: Use a consistent target name when building mariner initrd

jepio:
Currently a mixture of cbl-mariner and mariner is used when creating the mariner initrd. The kata-static tarball has mariner in the name, but the jenkins url uses cbl-mariner. This breaks cache usage.

Use mariner as the target name throughout the build, so that caching works.

Fixes: #7292


fidencio (approved):
lgtm, thanks @jepio!

GabyCT (approved):

Issue #6912, opened 2023-05-19
Resource allocation: podOverhead in runtimeClasses may not match with default base OS VM memory/CPU allocation made by the shim leading to misalignment of the view of allocated resources between kubernetes and shim


ms-mahuber:
**Disclaimer:**
  - This issue is one of the four issues forked based the discussion from [issue 6818, using static resource management for TEEs](https://github.com/kata-containers/kata-containers/issues/6818). All four issues are pre-fixed with "Resource allocation:"
  - I am aware I may be a bit deviating a bit from the style of how a bug/request should be opened, but this reflects an obvious issue observed by other peers, so I assume this is fine.

**Description:**
We should align the base memory/CPU assignment for the VM between kata-shim and our podOverhead specification while taking the host memory consumption into account as well.
  - podOverhead values are not translated down to shim level. These are basically an implicit assumption on kubernetes layer. For each pod, kubernetes substracts allocatable resources by what is specified in podOverhead. This needs to match with what the shim allocates for the VM's base OS (e.g. if you define 1 vCPU for the base VM OS and 256 MB default memory for the base VM OS, this will be part of the podOverhead) **and** for additional resources required on the host, e.g., for starting/running a cloud-hypervisor and companion processes.
  - I believe this needs to be addressed in documentation and refined across all podOverhead definitions. However this is not easy, following example state why
  - For instance, I was using cloud-hypervisor while using a kata image, not an initrd. In this case CH uses "--pmem file=" and maps the kata image as a private memory mapping in its memory map. My image was 384 MB large leading to additional 384 MB required *per pod* just by the CH process (this is by design, see the relevant [CH commit here]().
  - In the best case we would be able to exactly measure how many resources we want to reserve on host side, but also take into account which resources the kata shim will allocate for the base VM OS. While this is probably not practical across the whole solution space, we may want to give better estimates plus documentation

fidencio:
@ms-mahuber and I had a call Yesterday where we discussed this issue.

The approach we're going to take is to create a script that the user can run on their side to gather information about what would be a reasonable amount of memory and vcpu to be used with the podOverhead.

We'll also use those scripts on a known configuration (still to be decided) and update the values we currently have defined.


Now, what should this script do?
* Start one pod a certain amount of times, and gather memory / CPU consumption of the containerd-shim-kata-v2, VMMs, virtiofsd / nydusd for each run.
[...]

Issue #6910, opened 2023-05-19
Resource allocation: While request requests are not propagated to the shim, these should be prohibited to avoid misalignment of the view of allocated resources between kubernetes and shim


ms-mahuber:
**Disclaimer:**
  - This issue is one of the four issues forked based the discussion from [issue 6818, using static resource management for TEEs](https://github.com/kata-containers/kata-containers/issues/6818). All four issues are pre-fixed with "Resource allocation:"
  - I am aware I may be a bit deviating a bit from the style of how a bug/request should be opened, but this reflects an obvious issue observed by other peers, so I assume this is fine.

**Description:**
While we currently should use resource limits on kubernetes layer to request resources for workloads on shim level, we should ensure that nobody can specify resource requests. Otherwise, the view of kubernetes on allocated resources can vary drastically from what the shim actually allocates. Imagine you request 100MB of memory and define a limit of 1 GB in a pod yaml. The shim will allocate significantly more memory. While on kubernetes layer, only 100MB are allocated. I think for now this should either be clearly documented, and/or potentially solved with a workaround on kubernetes layer: We could enforce a [validating admission policy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) to prevent resource requests when using the kata or kata CoCo handler

fidencio:
@ms-mahuber and I had a call Yesterday where we discussed this issue.

The approach we're going to take is to provide a "sample validating admission policy", which we'll recommend user to use that with Kata Containers.

In the long term, k8s should expose this to us.

ms-mahuber:
After speaking with @fidencio , suggested a single solution in [issue 6911](https://github.com/kata-containers/kata-containers/issues/6911) to address both issues.

@Redent0r for visibility.

PR #7304, opened 2023-07-11
tools: Add resources webhook

### LINKED ISSUE #6911: Resource allocation: When no resource limits are specified, only the base memory/CPU is allocated to the VM and no cgroups are enforced in the VM
**Disclaimer:**
  - This issue is one of the four issues forked based the discussion from [issue 6818, using static resource management for TEEs](https://github.com/kata-containers/kata-containers/issues/6818). All four issues are pre-fixed with "Resource allocation:"
  - I am aware I may be a bit deviating a bit from the style of how a bug/request should be opened, but this reflects an obvious issue observed by other peers, so I assume this is fine.

**Description:**
This is definitely a problem. First, this theoretically does not account for resources that the workloads require (the base resources are only for the VM base OS like kata-agent and systemd, where applicable). Second, malicious workloads can exhaust the whole UVM resources.
  - A potential solution could be to always ensure a default amount of resources for workloads is provided. I tried making a patch on shim level that would introduce new configuration variables which would be used ensure that if no limits are given, a default amount of CPU and memory is allocated for the workloads (see src/runtime/pkg/oci/utils.go: SandboxConfig() where base CPU/memory is added to the requested workload CPU/memory). However, this would lead to a distorted view of the resources Kubernetes thinks are allocatable, and the resources the shim actually allocates under the plate
  - A potential better solution could be to define a default memory/CPU limit on Kubernetes layer via [LimitRanges](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/). This ensures we can define a good workload memory/CPU value and Kubernetes will be aware of it. Further, we can practically reduce the default base OS memory value of 2,048 MB for the UVM to a much smaller value that better represents the memory required for the base runtime in the UVM.
  - Another solution can be to fail when no limits are provided. Not nice.
  - Other, potential solutions exist for sure but these are beyond the scope of my knowledge

Redent0r:
This adds a webhook to set default resource requests and limits on a pod workload. Based on https://github.com/kata-containers/tests/tree/main/kata-webhook

Fixes: https://github.com/kata-containers/kata-containers/issues/6911

This pull requests adds a webhook to remediate resource allocation issues occurring between the Kubernetes layer and the kata shim (memory, CPU): with the current go-based shim implementation, [resource requests are not being passed down to the shim](https://github.com/kata-containers/kata-containers/issues/6910) due to [missing Sandbox API support](https://github.com/kata-containers/kata-containers/issues/7043).

On top of this, if no resource limits are being set, there is [no resources explicitly allocated for workloads](https://github.com/kata-containers/kata-containers/issues/6911). Resources are only implicitly allocated based on the default UVM size values (ex. default_vcpus, default_memory configuration variables). These variables are intended to reserve space for the UVM base OS (e.g. systemd, kata-agent), but not for workloads. 

The implementation of this kata webhook addresses the mentioned issues for shim implementations without Sandbox API support:
* Enforce default values for resource limits if none are set: https://github.com/kata-containers/kata-containers/issues/6911. The default values can be defined as a ConfigMap 
* Disallow resource requests: https://github.com/kata-containers/kata-containers/issues/6910 
* The implementation is based on the existing [kata webhook implementation](https://github.com/kata-containers/tests/tree/main/kata-webhook)
 
This webhook will also serve as a baseline to solve the following issue:
https://github.com/kata-containers/kata-containers/issues/6912 
With this webhook, users can define proper podOverhead where the default UVM memory and CPU requirements can be considered in addition to resource requirements on the container host. 
 
A potential improvement for this webhook could be to:
Create a configuration to enable or disable mentioned features (e.g. configMap, environment variables), making this a generic kata webhook that can be equipped with various other functionalities (e.g., solve problems on Kubernetes layer that need to be solved on this layer). With this, the new webhook implementation could be combined with the existing webhook, making runtime class injection a potential feature 

Additional notes:
- Main logic: https://github.com/kata-containers/kata-containers/pull/7304/files#diff-2b34b1875f921b33901653e9a7778baedc1d1a8f205f569626ef7b51c720fb5bR52-R65
- This webhook, as well as the runtime webhook, only work during pod creation. The clients will still be able to set resource request on an existing pod workload. We may be able to get around that by adding "UPDATE" to the [webhook operation rules](https://github.com/kata-containers/kata-containers/pull/7304/files#diff-afc6323c3e5f642df32c9031d589bb3277fa561332e5bd1b402d7b19e01aedf0R24), but I haven't tested that yet

- Some of the resources name are the same from https://github.com/kata-containers/tests/tree/main/kata-webhook. I can update the names if needed in order for the 2 webhooks to be used at the same time without name collision. 

- If they not specified, request values will take the values of limits 

- Case scenarios:

| Requests | Limits | Outcome   |
| ------------- |:-------------:| -----:|
| unset       | unset | Both requests and limits get set to default values |
| set           | unset | Both requests and limits get set to default values |
| unset       | set     | Both requests and limits get set to limits set by user |
| set           | set     | Both requests and limits get set to limits set by user |


Redent0r:
Hi @fidencio! I've been working on this with @ms-mahuber for the pass few days and we think it looks good so far. Would you mind taking a quick peek for any major suggestions needed before I set it ready for review? Thanks!

fidencio:
General comments here on how **I** would approach this, separating each item below in **one** commit in this **very same PR**.

1. Copy the webhook from https://github.com/kata-containers/tests/tree/main/kata-webhook to this repo, under `tools/` or `utils/` **as it is**.
2. Adapt webhook to this repo, so we it can be built and used from here.
2.1. It can be several commits, to make it easier for the reviewer
3.  Add the functionality to the webhook to deal with resources
3.1. This will make things easier so we can keep a single webhook for "all our purposes"
4. Match the limits to the default_max_$value of the vanilla config, in case no limits are specified
4.1. This will make sure that we don't break current use cases

Then, as different PRs, we'll have to:
1. Build it and ship and image as part of the release
2. Test it as part of our CI

Now, about the approach, I really like it and it does fit exactly what @ms-mahuber and I talked over the issues, kudos for working on that.

Last but not least, please, poke me on Slack (or even on Teams) for a review, as I am a few years behind on GitHub notifications.

fidencio:
@ms-mahuber, @Redent0r, what are your thoughts on the suggested approach?

ms-mahuber:
Sounds good. On 'Match the limits to the default_max_$value of the vanilla config, in case no limits are specified', I would suggest: Let's make the webhook configurable so that we are flexible in what limits values the webhook would set (if limits are not specified by the user in the pod yaml).

We will still need defaults for exactly the values the webhook would set if the values would not be explicitly configured for the webhook. For this, I'm not sure whether the "default_max_$value" would be the right thing. I think we will want some value where the sum of the UVM base OS mem/CPU + values set by limits are smaller than the maximum values (as those suggest how much the VM can 'grow').

Let us potentially follow up offline on this, then come up with a better description for the webhook for the community to make a next iteration step. :)

Redent0r:
Thank you both! I like those suggestions and will implement them

tools/kata-webhook/main.go
```diff
@@ -0,0 +1,177 @@
...
+	fmt.Println("setting runtime to kata: ", pod.GetNamespace(), pod.GetName())
+
+	runtimeClassEnvKey := "RUNTIME_CLASS"
+	kataRuntimeClassName := getEnvValue(runtimeClassEnvKey, "kata")
+	pod.Spec.RuntimeClassName = &kataRuntimeClassName
+
+	if getEnvValue("MANAGE_RESOURCES", "true") == "true" {
+		fmt.Println("setting default limits and request values: ", pod.GetNamespace(), pod.GetName())
+
+		hypConfig := katautils.GetDefaultHypervisorConfig()
```
Redent0r, line 87: Where to get cpu and memory limit from? Trying GetDefaultHypervisorConfig()

Redent0r: After talking to @fidencio, we opted to take the default values from toml configuration and let the user define where is the configuration. For our current setup, is `/opt/confidential-containers/share/defaults/kata-containers`. We should take `default_maxvcpus` and `default_maxmemory` and use these for cpu and memory limit in case the user does not specifies in the config map

ms-mahuber: I think the webhook should either define its own defaults/values or we should introduce new values in the config that designate default resources available for/to the containers (potentially similar to `default_memory` and `default_vcpus` which define the resources for the "UVM OS" part of the sandbox). `default_maxvcpus` and `default_maxmemory` does not sound appropriate here.

For now, I would suggest to define those values in the webhook. If the webhook gets upstreamed, one can attempt to change the main code base. The main code base could either take the resource limits e.g. set by the webhook, or defined by the user. If the webhook is not running and the user does not specify limits, default values could be used and added to the VM resources (e.g. UVM OS memory + container memory = memory size of UVM)

Redent0r: Changing to https://github.com/kata-containers/kata-containers/pull/7304/commits/c2ef95baf2fb5fb771724a91223675cc128548ee for now and open to explore better solution

Issue #7305, opened 2023-07-12
runtime-rs: Scan network if network model is "none"


justxuewei:
The expected behavior is that do not scan if the network model is specified to "none", since pods don't need network.

PR #7306, opened 2023-07-12
runtime-rs: Do not scan network if network model is "none"

justxuewei:
Skip to scan network from netns if the network model is specified to "none".

Fixes: #7305


Tim-0731-Hzt (approved):

lifupan (approved):

studychao (approved):
LGTM, thanks

Issue #6543, opened 2023-03-29
CI | GHA | Create a PoC for running non-k8s tests using an Azure VM


fidencio:
A proof-of-concept has to be created to run the equivalent of `make cri-containerd` / `make shim-v2` /  `make sandbox-cgroup` / `make stability` / `make dragonball-stability` /  `make runk` / `make vcpus` / `make rootless` tests, but using GitHub actions, on an Azure VM

The expected outcome is:
* A PR with the PoC
* A list of hypervisors being tested
* A list of passing tests per hypervisor tested
* Issues created for the failing tests per hypervisor

The owner of this task is responsible for creating new issues for each one of the targets being tested.

PR #7314, opened 2023-07-12
tests: gha: ci: Add cri-containerd tests skeleton

fidencio:
---

tests: Add `function` before each function

Let's just keep this standardised.

---

tests: common: Move a few utility functions to common.bash

Those functions were originally introduced as part of the
`metrics/gha-run.sh` file, but those will be very hand at the time we
start adding more tests.

---

gha: ci: Add cri-containerd tests skeleton

This PR builds the foundation for us to start migrating the
cri-containerd tests from Jenkins to GitHub Actions.

Right now the test does nothing and should always finish successfully.
The coming PRs will actually introduce logic to the `gha-run.sh` script
where we'll be able to run the tests and make sure those pass before
having them actually merged.

Fixes: #6543 

---

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -23,23 +23,23 @@ KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 
 RUNTIME="${RUNTIME:-containerd-shim-kata-v2}"
 
-die() {
+function die() {
 	local msg="$*"
 	echo -e "[$(basename $0):${BASH_LINENO[0]}] ERROR: $msg" >&2
 	exit 1
 }
 
-warn() {
+function warn() {
 	local msg="$*"
 	echo -e "[$(basename $0):${BASH_LINENO[0]}] WARNING: $msg"
 }
 
-info() {
+function info() {
 	local msg="$*"
 	echo -e "[$(basename $0):${BASH_LINENO[0]}] INFO: $msg"
 }
 
-handle_error() {
+function handle_error() {
 	local exit_code="${?}"
 	local line_number="${1:-}"
 	echo -e "[$(basename $0):$line_number] ERROR: $(eval echo "$BASH_COMMAND")"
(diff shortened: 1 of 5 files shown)


GabyCT (approved):

GabyCT:
@fidencio seems like metrics gha has an issue with
```
tests/metrics/gha-run.sh: line 39: compress_metrics_results_dir: command not found
```

dborquez (approved):

tests/common.bash
```diff
@@ -76,7 +76,7 @@ is_a_kata_runtime() {
 
 # Gets versions and paths of all the components
 # list in kata-env
-extract_kata_env() {
+function extract_kata_env() {
 	RUNTIME_CONFIG_PATH=$(kata-runtime kata-env --json | jq -r .Runtime.Config.Path)
 	RUNTIME_VERSION=$(kata-runtime kata-env --json | jq -r .Runtime.Version | grep Semver | cut -d'"' -f4)
 	RUNTIME_COMMIT=$(kata-runtime kata-env --json | jq -r .Runtime.Version | grep Commit | cut -d'"' -f4)
```
dborquez, line 82: Please use `sudo` to invoke `kata-runtime` command so that it works as expected.

dborquez (changes requested):

dborquez:
> compress_metrics_results_dir

@fidencio do you know if `tar` is already installed on the vm?

tests/common.bash
fidencio: I don't think that's the case, @dborquez, as we're not changing anything on this path.

The error we faced was because I mistakenly removed a function that was metrics specific and was in the common.bash.

I introduced a new commit to the PR moving that function to metrics/gha-run.sh

fidencio:
> > compress_metrics_results_dir
> 
> @fidencio do you know if `tar` is already installed on the runner?

I had mistakenly removed the function, things should be good now.

The error we faced was because I mistakenly removed a function that was metrics specific and was in the common.bash.

I introduced a new commit to the PR moving that function to metrics/gha-run.sh

dborquez (approved):
LGTM, thank you @fidencio

Issue #7319, opened 2023-07-13
metrics: fix jq parsing expression used in processMetricsBaseline


dborquez:
[checkmetrics](https://github.com/kata-containers/kata-containers/blob/ee17097e8819a59a74cc6d07bf45de348c769160/tests/metrics/cmd/checkmetrics/main.go#L74) reports the following error when [parsing metrics results](https://github.com/kata-containers/kata-containers/blob/ee17097e8819a59a74cc6d07bf45de348c769160/tests/metrics/cmd/checkmetrics/ci_worker/checkmetrics-json-clh-kata-metric8.toml#L43):

```
jq '.\"memory-footprint-inside-container\".Results | .[] |.memtotal.Result'  memory-footprint-inside-container.json
jq: error: syntax error, unexpected INVALID_CHARACTER (Unix shell quoting issues?) at <top-level>, line 1:
```

The solution is to replace the backslash that escapes double quotes, with square brackets, as follows:
`jq '.["memory-footprint-inside-container"].Results | .[] |.memtotal.Result'  memory-footprint-inside-container.json`

PR #7320, opened 2023-07-13
metrics: replace backslashes used to escape double quoted jq key expr.

dborquez:
This PR uses squared brackets in a jq expression to access key values corresponding to metric results in json format.

The values are the data inputs into the checkmetrics tool.

Fixes: #7319


fidencio (approved):
lgtm, thanks @dborquez!

cmaf (approved):

GabyCT:
@dborquez seems that there is an issue while reading the result

PR #7322, opened 2023-07-13
gha: ci: Add cri-containerd tests skeleton -- follow up 1

fidencio:
---

ci: cri-containerd: Ensure deps are installed

Let's make sure we install the needed dependencies for running the
`cri-containerd` tests.

---

ci: cri-containerd: Export GOPATH

Let's make sure this is exported, as it'll be needed in order to install
`yq`, which will be used to get the versions of the dependencies to be
installed.

---

ci: cri-containerd: Add LTS / Active versions for containerd

As we'll be testing against the LTS and the Active versions of
containers, let's add those entries to the versions.yaml file and make
sure we export what we want to use for the tests as an env var.

The approach taken should not break the current way of getting the
containerd version.

LTS and Active versions of containerd can be found at:
https://containerd.io/releases/#support-horizon

Fixes: #6543

---

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -228,6 +228,8 @@ externals:
     # instead of abstract socket, thus kata wouldn's support the containerd's
     # version older than them.
     version: "v1.6.8"
+    lts: "v1.6"
+    active: "v1.7"
 
   critools:
     description: "CLI tool for Container Runtime Interface (CRI)"
--- a/.github/workflows/run-cri-containerd-tests.yaml
+++ b/.github/workflows/run-cri-containerd-tests.yaml
@@ -17,6 +17,7 @@ jobs:
         vmm: ['clh', 'qemu']
     runs-on: garm-ubuntu-2204
     env:
+      GOPATH: ${{ github.workspace }}
       KATA_HYPERVSIOR: ${{ matrix.vmm }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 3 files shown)


versions.yaml
```diff
@@ -228,6 +228,8 @@ externals:
     # instead of abstract socket, thus kata wouldn's support the containerd's
     # version older than them.
     version: "v1.6.8"
+      lts: "v1.6"
+      active: "v1.7"
```
jepio, line 232: these should be as indented as `version:`, otherwise this is not valid yaml
```suggestion
    lts: "v1.6"
    active: "v1.7"
```

fidencio: Done, thanks for catching this up!

jepio (approved):

fidencio:
"""
ci: cri-containerd: Ensure deps are installed

Let's make sure we install the needed dependencies for running the
cri-containerd tests.
"""

This is only a placeholder, the dependencies that will be installed are basically:
* From the OS
  * jq
  * curl (already present)
* From our repo
  * yq (using the install_yq script)
* From GitHub
  * cri-containerd
  * cri-tools
  * cni plugins

We may need a few more packages, which we'll end up finding out in the way of moving the tests here, but what's listed are the most obvious ones.

stevenhorsman (approved):
LGTM. Thanks

PR #7324, opened 2023-07-13
Remove trailing whitespace from source code

c3d:
A lot of our documentation and source code have leftover trailing whitespace, which causes unnecessary diffs when saving files with an editor configured to remove them.


c3d:
@fidencio I did a pass at cleaning up trailing whitespace all over Kata. I purposefully ignored `.patch` files

fidencio:
@c3d, nice.  What about adding a test so we avoid adding any new whitespace to the code?

c3d:
> @c3d, nice. What about adding a test so we avoid adding any new whitespace to the code?

Hmm, OK, but then I will need to either also do the `.patch` files, or to have an exclude list. I'll take a look

c3d:
Pushed a new version excluding the vendored code. I will not work on something testing that trailing spaces are not committed until I know exactly what our CI expects ;-)

GabyCT (approved):

PR #4699, opened 2022-07-20
WIP: runtime: Add support for GPUDirect and GPUDirect RDMA PCIe topology Part 2

zvonkok:
Depends on: https://github.com/kata-containers/kata-containers/pull/4492

Add support for CDI https://github.com/container-orchestrated-devices/container-device-interface

PR #7325, opened 2023-07-13
gpu: Adding CDI support for cold and hot-plug of VFIO devices 

zvonkok:
We need to do proper sandbox sizing when we're doing cold-plug introduce CDI, the de-facto standard for enabling devices in containers. 

There is a new package vendored, this patch only touches two files `container.go` and `create.go`


src/runtime/pkg/containerd-shim-v2/create.go
```diff
@@ -247,6 +310,7 @@ func loadRuntimeConfig(s *service, r *taskAPI.CreateTaskRequest, anno map[string
 		configPath = os.Getenv("KATA_CONF_FILE")
 	}
 
+	shimLog.Info("### loadRuntimeConfig: configPath ", configPath)
```
jepio, line 313: should remove this

zvonkok:
@jepio Yes will create an issue and explain.

zvonkok:
@jepio Issue #7331 most of the infrastracture to support this was already merged. We just need to inject the devices into the spec and `sandbox.go` and `container.go` will take care of the rest.

src/runtime/virtcontainers/container.go
```diff
@@ -869,6 +869,23 @@ func (c *Container) create(ctx context.Context) (err error) {
 		}
 	}
 
+	// If cold-plug we've attached the devices already, do not try to
+	// attach them a second time.
+	coldPlugVFIO := (c.sandbox.config.HypervisorConfig.ColdPlugVFIO != config.NoPort)
+	if coldPlugVFIO {
+		var cntDevices []ContainerDevice
+		for _, dev := range c.devices {
+			if strings.HasPrefix(dev.ContainerPath, vfioPath) {
```
bergwolf, line 878: Does the cold-plug implementation handle **ALL** vfio devices, including network, gpu, nvme etc? I'm asking because the change here assumes that all vfio devices are already code-plugged.

bergwolf (commented):
@zvonkok thanks for working on the CDI support! Do you need additional change to make CDI work or is this the last PR for CDI? When it is done, could you add an E2E test case so that we make sure that it is working all the time?

zvonkok:
@bergwolf This is part one to add CDI for the "outer-runtime", the second step is add CDI support for the "inner-runtime". Depending on the `vfio_mode=guest-kernel | vfio` we can auto generate CDI specs inside the VM for devices that support it. 
In the case of GPUs this is already done in the images/initrds.

src/runtime/virtcontainers/container.go
zvonkok: That's a good point. At sandbox creation time I am "marking" cold-plugged devices I need to add an additional check if they are VFIO && cold-plugged.

zvonkok: Ok thinking more about this, if we set `cold_plug_vfio="root-port"` ALL vfio devices are cold-plugged to the `root-port` and hence all VFIO devices need to be removed in the container create stage. 

This PR https://github.com/kata-containers/kata-containers/pull/4699/files#diff-e6df7a8b934e487feec13185e69c9bcb0e256e513bb792141ec9162d5edad399 will introduce selectively choose were to attach each VFIO device.

zvonkok: We currently have no way of telling how individual devices are going to be hot|cold plugged. Added CDI in this PR to exactly solve this with the aforementioned PR.

bergwolf (approved):
lgtm, I'm fine with merging it first and implementing CDI support incrementally.  Thanks!

jepio (approved):

Issue #6854, opened 2023-05-17
clh does not surpport with static vfio


liangxianlong:
**Is your feature request related to a problem? Please describe.**
In some scenarios, static addition of vfio network cards is required

PR #6856, opened 2023-05-17
runtime: surpport cold addition vfio for clh

liangxianlong:
In some scenarios, cold addition of vfio network cards is required.

Fixes: #6854

PR #7339, opened 2023-07-14
runtime: support cold addition vfio for clh

liangxianlong:
In some scenarios, cold addition of vfio network cards is required.

Fixes: #6854

### DIFF
--- a/src/runtime/virtcontainers/clh_test.go
+++ b/src/runtime/virtcontainers/clh_test.go
@@ -127,6 +127,31 @@ func TestCloudHypervisorAddVSock(t *testing.T) {
 	assert.Equal(clh.vmconfig.Vsock.Socket, "path")
 }
 
+func TestCloudHypervisorAddVfioDevice(t *testing.T) {
+	Id := "vfio123"
+	path := "/dev/vfio/101"
+
+	vfDevice := config.VFIODev{
+		ID:       Id,
+		SysfsDev: path,
+	}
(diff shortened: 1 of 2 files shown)


liangxianlong:
@likebreath

src/runtime/virtcontainers/clh_test.go
```diff
@@ -127,6 +127,31 @@ func TestCloudHypervisorAddVSock(t *testing.T) {
...
+	case config.VFIODev:
+		Id := "vfio123"
+		path := "/dev/vfio/101"
+		_ = clh.addVfioDevice(s)
+
+		assert.Equal(s.ID, Id)
+		assert.Equal(s.SysfsDev, path)
+	}
+}
+
```
likebreath, line 154: I don't think the abstraction of `testCloudHypervisorAddDevice` here is making sense. This function should be removed and everything can be done from `TestCloudHypervisorAddVfioDevice`.

likebreath (changes requested):
Changes to clh.go looks good to me. One comments below about the unit test.

Also, it would be better to keep using one PR for addressing review comments, so that we all can reference previous conversations. You can always use `git push --force` to override your previous commits.

For the context, this PR is a follow-up of #6856.

jepio:
Let's wait to merge this together with https://github.com/kata-containers/kata-containers/pull/7971 which adds a test case of cold-plug vfio.

liangxianlong:
@liubin

Issue #7145, opened 2023-06-20
kata metrics get error infomation about podoverhead


liangxianlong:
# Description of problem

When use kata metrics to collect kata datas with `sandbox_cgroup_only=false`, I find `kata_shim_pod_overhead_memory_in_bytes` is wrong.

# Expected result
The kata_shim_pod_overhead_memory_in_bytes should greater than 0.

# Actual result
```bash
# root @ host-59 in /home/lxl/neo_test_images [19:46:02] 
$ curl 127.0.0.1:8090/metrics | grep kata_shim_pod_overhead_memory_in_bytes
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0# HELP kata_shim_pod_overhead_memory_in_bytes Kata Pod overhead for memory resources(bytes).
# TYPE kata_shim_pod_overhead_memory_in_bytes gauge
kata_shim_pod_overhead_memory_in_bytes{sandbox_id="0e7763d1fd2f6fc29255cb5ee7b208369d82cc5d26a80ebbc337cbdd616c62ed",cri_uid="hdilhd83dsaidwnduwk28bcsb",cri_name="lxl",cri_namespace="de"} 1.8446744073707467e+19
100  316k    0  316k    0     0  7545k      0 --:--:-- --:--:-- --:--:-- 7545k
```
The value is `1.8446744073707467e+19` which is a overflow value.
# Why?
```go
func calcOverhead(initialSandboxStats, finishSandboxStats vc.SandboxStats, initialContainerStats, finishContainersStats []vc.ContainerStats, deltaTime float64) (float64, float64) {
	hostInitCPU := initialSandboxStats.CgroupStats.CPUStats.CPUUsage.TotalUsage
	guestInitCPU := uint64(0)
	for _, cs := range initialContainerStats {
		guestInitCPU += cs.CgroupStats.CPUStats.CPUUsage.TotalUsage
	}

	hostFinalCPU := finishSandboxStats.CgroupStats.CPUStats.CPUUsage.TotalUsage
	guestFinalCPU := uint64(0)
	for _, cs := range finishContainersStats {
		guestFinalCPU += cs.CgroupStats.CPUStats.CPUUsage.TotalUsage
	}

	var guestMemoryUsage uint64
	for _, cs := range finishContainersStats {
[...]

PR #7340, opened 2023-07-14
runtime: get podoverhead from kata_overhead

liangxianlong:
kata metrics get error infomation about podoverhead

Fixes: #7145

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -1660,16 +1660,12 @@ func (s *Sandbox) Stats(ctx context.Context) (SandboxStats, error) {
 	var metrics interface{}
 	var err error
 	if !s.config.SandboxCgroupOnly {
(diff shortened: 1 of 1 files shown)


src/runtime/virtcontainers/sandbox.go
```diff
@@ -1657,9 +1657,19 @@ func (s *Sandbox) StatsContainer(ctx context.Context, containerID string) (Conta
...
 
-	metrics, err := s.sandboxController.Stat()
-	if err != nil {
-		return SandboxStats{}, err
+	var metrics interface{}
+	var err error
+	if !s.config.SandboxCgroupOnly {
+
+		metrics, err = s.overheadController.Stat()
+		if err != nil {
```
liubin, line 1665: You can check the err after `if !s.config.SandboxCgroupOnly { } else {}`, then you can save one err check.

Issue #7343, opened 2023-07-14
ci: Run k8s tests with kata-containers on debug mode


fidencio:
This will help us to get more info whenever failures happen.

Issue #7342, opened 2023-07-14
kata-deploy: Allow users to run Kata Containers on debug mode


fidencio:
Let's add an env var to the `kata-deploy.yaml` that allows user to run kata-containers in debug mode.

PR #7344, opened 2023-07-14
kata-deploy: Add a debug option to kata-deploy (and also use it as part of our CI)

fidencio:
---

kata-deploy: Give users the ability to run it on DEBUG mode

The DEBUG env var introduced to the kata-deploy / kata-cleanup yaml file
will be responsible for:
* Setting up the CRI Engine to run with the debug log level set to debug
  * The default is usually info
* Setting up Kata Containers to enable:
  * debug logs
  * debug console
  * agent logs

This will help a lot folks trying to debug Kata Containers while using
kata-deploy, and also help us to always run with DEBUG=yes as part of
our CI.

Fixes: #7342

---

ci: k8s: Enable debug when running the tests

This will help us to gather more information about Kata Containers in
case of failure.

Fixes: #7343

---

### DIFF
--- a/tests/integration/gha-run.sh
+++ b/tests/integration/gha-run.sh
@@ -77,6 +77,10 @@ function run_tests() {
     kubectl delete namespace kata-containers-k8s-tests &> /dev/null || true
 
     sed -i -e "s|quay.io/kata-containers/kata-deploy:latest|${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}|g" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
+
+    # Enable debug for Kata Containers
+    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[1].value' "\"yes\""
+
     if [ "${KATA_HOST_OS}" = "cbl-mariner" ]; then
         yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[+].name' "HOST_OS"
         yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[-1].value' "${KATA_HOST_OS}"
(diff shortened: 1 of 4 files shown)


tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -212,6 +223,14 @@ function configure_crio() {
 	for shim in "${shims[@]}"; do
 		configure_crio_runtime $shim
 	done
+
+
+	if [ "${DEBUG:-"no"}" == "yes" ]; then
+		cat <<EOF | tee -a $crio_drop_in_conf_file_debug
+[crio]
```
zvonkok, line 230: What about the case where a user already has set `log_level` to some value?

zvonkok: You could do the same logic you did for `containerd`

fidencio: So, CRI-O has this drop-in mechanism.  Our will just override what the users have.

zvonkok: Otherwise looks good to me.

tests/integration/gha-run.sh
```diff
@@ -77,6 +77,10 @@ function run_tests() {
     kubectl delete namespace kata-containers-k8s-tests &> /dev/null || true
 
     sed -i -e "s|quay.io/kata-containers/kata-deploy:latest|${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}|g" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
+
+    # Enable debug for Kata Containers
+    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[DEBUG].value' "yes"
```
fidencio, line 82: This is overriding all the env vars, not only editing the one that I want.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
fidencio: But why? :-)

We want to override it in this case anyways, and we do that by drop-in a new file.

zvonkok: Holy cow, missed the `drop_in_conf_file`, there is a clear advantage if you can read.

zvonkok (approved):
LGTM Obrigado por seu esforço

stevenhorsman (approved):
LGTM and should be helpful in problem solving. Thanks!

PR #7356, opened 2023-07-17
docs: remove the promp prefix from commands in README

aucampia:
This makes it easier for users to copy and paste the commands as they don't first have to modify them by hand.

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -53,7 +53,7 @@ behaviour.
 To show the available options without installing anything, run:
 
 ```sh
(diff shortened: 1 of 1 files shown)


stevenhorsman:
Hey @aucampia - I sympathise with your frustrations (and I also feel them and run into the same issues with some doc PRs), but the [kata-containers doc guidelines](https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#code-blocks) specify that each command line needs prefixing with $, so until we have a cross the board conversation about that and agree to update the requirements I don't think just updating individual docs is a good idea. Thanks

aucampia:
@stevenhorsman thanks for the info, closing this PR then.

Issue #7364, opened 2023-07-17
Using fsGroup with direct-assign breaks symbolic link insides container file system


alakesh:
# Description of the problem

1. Create a pod spec that has security context and uses fsGroup and persistentVolumeClaim and apply the config

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: repro
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 8Gi
  storageClassName: scaleio
---
apiVersion: <version>
kind: StatefulPod
metadata:
  name: repro
spec:
  podSpec:
    containers:
    - image: <any linux>
      command: ["/bin/sh", "-c"]
      args: ["while (true); do sleep 10; done;"]
      imagePullPolicy: Always
      name: c1
      resources:
        limits:
          cpu: "1"
          memory: 2Gi
        requests:
          cpu: 100m
          memory: 2Gi
      volumeMounts:
      - mountPath: /mnt/volume
        name: data-volume
      securityContext:
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: true
    priorityClassName: p1
    restartPolicy: Always
    runtimeClassName: microvm
    securityContext:
      fsGroup: 51020
      runAsGroup: 51020
      runAsUser: 264
    volumes:
    - name: data-volume
      persistentVolumeClaim:
        claimName: repro
  stickyIP: false
  suspend: false
  terminationPolicy:
    authority: onDelete
    serviceInstance: repro
    serviceName: repro
```
2. Inside the container

```
$ mkdir /mnt/volume/data && touch /mnt/volume/data/file && ln -s /mnt/volume/data/file /mnt/volume/link && ls -lart /mnt/volume/
$ echo "foo" > /mnt/volume/link
$ cat /mnt/volume/link
foo
$ cat /mnt/volume/data/file
```
So all looks good till now. 

3. Next we delete the pod which will result in recreating it.

4. When the pod comes up try to read the symbolic link

```
[/]$ cat /mnt/volume/link

```
This results in container status changes to error state and we see a lot of I/O error for metadata and logs in kernel dmesg.

# Expected result

The container is expected to go into "Running" state

# Actual result

It goes into "RunContainerError" state instead. 
There are lot of XFS errors in demsg

```
[Tue Jul 11 23:57:17 2023] XFS (vda): metadata I/O error in "xfs_buf_iodone_callback_error" at daddr 0x40 len 32 error 5
[...]

PR #7365, opened 2023-07-17
agent: exclude symlinks from recursive ownership change

alakesh:
currently when fsGroup is used with direct-assign, kata agent recursively changes ownership and permission for each file including symlinks. However the problem with symlinks is, the permission of the symlink itself may not be same as the underlying file. So while doing recursive ownership and permission changes we should skip symlinks.

Fixes: #7364


bergwolf (approved):
Nice catch! Thanks!

Issue #7373, opened 2023-07-18
ci: k8s: Move gha-run.sh to the kuberentes dir


fidencio:
The file belongs there, as it's only used for k8s related tests.

PR #7374, opened 2023-07-18
ci: Move `tests/integration/gha-run.sh`  to `tests/integration/kuberentes/` ... and also remove KUBECONFIG from the tdx envs

fidencio:
---

ci: tdx: Adjust KUBECONFIG

We don't need to export KUBECONFIG there.  Let's just make sure we have
the server correctly setup and avoid doing that.

---

ci: k8s: Move gha-run.sh to the kuberentes dir

The file belongs there, as it's only used for k8s related tests.

Fixes: #7373

---

Please, **do not set ok-to-test** on this one, as the changes on yaml files will only be picked up after the merge.


stevenhorsman (approved):
LGTM. Thanks

Issue #7381, opened 2023-07-19
s390x: Bring back AP devices handling with the latest PCie Topology Enhancements


zvonkok:
AP Devices were introduced in previous PRs which was based on a very old implementation of VFIO handling. With the recent advancements in PCI Topology reworks we need to add AP VFIO support for s390x. 

We need to be explicit on which port we're attaching the AP VFIO device and add handling for arch specific `attachDevice`. 

Enable cold-plugging and hot-plugging of AP VFIO devices.

PR #7382, opened 2023-07-19
s390x: Fixing device.Bus assignment

zvonkok:
Fixes: #7381

The device.Bus was reset if a specific combination of configuration parameters were not met. With the new PCIe topology this should not happen anymore


zvonkok:
Removing `HotplugVFIOonRootBus` which is obsolete with the latest PCI topology changes, users can set `cold_plug_vfio` or `hot_plug_vfio` either in the `configuration.toml` or via annotations.

zvonkok:
Since this mostly affects s390x @hbrueckner @BbolroC PTAL

src/runtime/pkg/device/manager/manager.go
```diff
@@ -195,7 +195,6 @@ func (dm *deviceManager) AttachDevice(ctx context.Context, id string, dr api.Dev
 	if !ok {
 		return ErrDeviceNotExist
 	}
-
```
BbolroC, line 198: Do we need to add this change to the commit? Thanks.

zvonkok: Of course not. Committed but not pushed.

zvonkok: Waiting for the static-checks to complete.

src/runtime/virtcontainers/sandbox.go
BbolroC, line 1: How do you think if we would have unit tests for the updated logic?

BbolroC (approved):
I've left a small knit, but overall lgtm. Thanks @zvonkok ! I appreciate your hard work.

BbolroC:
Test results for s390x (`vfio-ap`): PASS

zvonkok:
Depends on: https://github.com/kata-containers/tests/pull/5727

GabyCT (approved):

PR #7394, opened 2023-07-20
kata-deploy: Add VERSION and versions.yaml to the final tarball

fidencio:
Let's make things simpler to figure out which version of Kata
Containers has been deployed, and also which artefacts come with it.

This will help us immensely in the future, for the TEEs use case, so we
can easily know whether we can deploy a specific guest kernel for a
specific host kernel.


stevenhorsman (approved):
Looks ok to me

GabyCT (approved):

PR #7399, opened 2023-07-20
packaging/tools: Add kata-debug and use it as part of our CI

fidencio:
kata-debug is a tool that is used as part of the Kata Containers CI to gather
information from the node, in order to help debugging issues with Kata
Containers.

As one can imagine, this can be expanded and used outside of the CI context,
and any contribution back to the script is very much welcome.

The resulting container is stored at the [Kata Containers quay.io
space](https://quay.io/repository/kata-containers/kata-debug) and can
be used as shown below:
```sh
kubectl debug $NODE_NAME -it --image=quay.io/kata-containers/kata-debug:latest
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ TOOLS += trace-forwarder
 
 STANDARD_TARGETS = build check clean install static-checks-build test vendor
 
+# Variables for the build-and-publish-kata-debug target
+KATA_DEBUG_REGISTRY ?= ""
+KATA_DEBUG_TAG ?= ""
+
 default: all
 
 include utils.mk
@@ -43,6 +47,9 @@ static-checks: static-checks-build
 
 docs-url-alive-check:
 	bash ci/docs-url-alive-check.sh
+
+build-and-publish-kata-debug:
+	bash tools/packaging/kata-debug/kata-debug-build-and-upload-payload.sh ${KATA_DEBUG_REGISTRY} ${KATA_DEBUG_TAG} 
 
 .PHONY: \
 	all \
--- a/tools/packaging/kata-debug/kata-debug-build-and-upload-payload.sh
+++ b/tools/packaging/kata-debug/kata-debug-build-and-upload-payload.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+#
+# Copyright 2023 Intel
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+[ -z "${DEBUG}" ] || set -x
+set -o errexit
+set -o nounset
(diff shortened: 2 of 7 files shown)


jepio (approved):

Makefile
```diff
@@ -24,6 +24,10 @@ TOOLS += trace-forwarder
 
 STANDARD_TARGETS = build check clean install static-checks-build test vendor
 
+# Variables for the build-and-publish-kata-debug target
+REGISTRY ?= ""
+TAG ?= ""
```
stevenhorsman, line 29: Completely optional suggestion: Would it potentially help add clarity now and avoid issues later if there were named something like `KATA_DEBUG_REGISTRY` and `KATA_DEBUG_TAG`

tools/packaging/kata-debug/kata-debug-build-and-upload-payload.sh
```diff
@@ -0,0 +1,42 @@
...
+set -o pipefail
+set -o errtrace
+
+KATA_DEBUG_DIR="`dirname ${0}`"
+
+REGISTRY="${1:-"quay.io/kata-containers/kata-debug"}"
+TAG="${2:-}"
+
+arch=$(uname -m)
+[ "$arch" = "x86_64" ] && arch="amd64"
```
stevenhorsman, line 20: Should I/we raise multi-arch support will be a separate issue?

stevenhorsman (approved):
Thanks! kata-debug looks like it will be really helpful. Just to ask a basic question and check that I understand how it work - it runs a bit like a short-lived side car container in the cluster and gets the info from the worker and output it to the terminal?

tools/packaging/kata-debug/kata-debug-build-and-upload-payload.sh
fidencio: I'll take care of that in a different PR, where I will add this as part of our release process.

Makefile
fidencio: Makes sense, and I will add it.

fidencio:
> Thanks! kata-debug looks like it will be really helpful. Just to ask a basic question and check that I understand how it work - it runs a bit like a short-lived side car container in the cluster and gets the info from the worker and output it to the terminal?

Exactly.  You can see the output of that already as part of our cleanup in the GHA k8s tests.

fidencio:
AS the static-checks are running (and hopefuilly passing), I've added the force-skip-ci for the jenkins tests as nothing here can be tested there.

fidencio:
static checks failured on a totally non-related issue. :-/

Issue #7405, opened 2023-07-21
kata-deploy: Properly get the path of the versions.yaml file


fidencio:
We need to correctly get the full path of the versions.yaml file as part
of the merge-builds.sh script, as we do a `pushd` there and that leads
to a fail merging the artefacts as the `versions.yaml` file does not
exists in that path.

PR #7406, opened 2023-07-21
kata-deploy: Properly get the path of the versions.yaml file

fidencio:
We need to correctly get the full path of the versions.yaml file as part of the merge-builds.sh script, as we do a `pushd` there and that leads to a fail merging the artefacts as the `versions.yaml` file does not exists in that path.

Fixes: #7405


zvonkok (approved):
LGTM Thanks!

jepio (approved):

stevenhorsman (approved):
LGTM. Thanks

Issue #7409, opened 2023-07-21
kata-deploy: Create the runtime-classes as part of the script


fidencio:
By doing this we avoid the user one step, and the Confidential Containers operator can easily take advantage of that.

This should be done in a way that does not break comaptibily with the current kata-deploy scripts.

PR #7408, opened 2023-07-21
kata-deploy: Allow shim creation based on what's passed to the daemonset

fidencio:
Instead of hardcoding shims as part of the script, let's ensure we can allow them to be created based on environment variables passed to the daemonset.

This change brings no functionality change as the default values in the daemonset are exactly what has been used as part of the scripts.

Fixes: #7407

PR #7411, opened 2023-07-21
kata-deploy: Allow runtimeclasses to be created by the daemonset

fidencio:
Let's allow the daemonset to create the runtimeclasses, which will
decrease one manual step a user of kata-deploy should take, and also
help us in the Confidential Containers land as the Operator can just
delegate it to this script.

Fixes: #7409

**NOTE:** I have this one rebased atop of https://github.com/kata-containers/kata-containers/pull/7408 and that one should be merged first.


tests/integration/kubernetes/gha-run.sh
```diff
@@ -143,6 +144,9 @@ function cleanup() {
     kubectl delete ${deploy_spec}
     kubectl -n kube-system wait --timeout=10m --for=delete -l name=kata-deploy pod
 
+    # Let the `kata-deploy` script take care of the runtime class creation / removal
+    yq write -i "${tools_dir}/packaging/kata-deploy/kata-cleanuo/base/kata-cleanup.yaml" 'spec.template.spec.containers[0].env[4].value' "\"yes\""
```
dborquez, line 148:
```suggestion
    yq write -i "${tools_dir}/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml" 'spec.template.spec.containers[0].env[4].value' "\"yes\""
```

dborquez (changes requested):

fidencio:
/test

@dborquez @zvonkok @jepio @sprt

zvonkok (approved):
LGTM, this is great; this will make deployment far easier.

fidencio:
I've force-pushed a change as I noticed that we don't need to remove the runtimeclasses in case those are created by the daemonset itself.

tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
```diff
@@ -13,7 +13,7 @@ spec:
         labels:
           name: kubelet-kata-cleanup
     spec:
-      serviceAccountName: kata-label-node
+      serviceAccountName: kata-deploy
```
amshinde, line 16: suggestion: Suffix this with "sa" or "service-account" as the daemonset is named similarly as well which is not an issue but just makes things more explicit and avoids confusion while describing either or.

tools/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml
```diff
@@ -2,28 +2,31 @@
...
 metadata:
-  name: kata-label-node
+  name: kata-deploy
   namespace: kube-system
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: node-labeler
+  name: kata-deploy-cr
```
amshinde, line 11: suggestion : /cr/role/

tests/integration/kubernetes/gha-run.sh
```diff
@@ -173,21 +173,6 @@ function delete_cluster() {
...
-    kubectl get nodes -o wide --show-labels=true
-    echo "::endgroup::"
-    echo ""
-    echo "::group::Get all the pods running"
-    kubectl get pods -A
-    echo "::endgroup::"
-    echo ""
-    if [[ "${describe_pods}" == "yes" ]]; then
-	echo "::group::Describe all the pods"
-    	kubectl describe pods -A
```
amshinde, line 179: I believe the describe pods commands could give useful logs when there is a pod failure. You may want to reconsider this. In case of failure, the describe command ends up being quite helpful.

fidencio: We're already getting the `kubectl describe pods`  as part of a failure, this one was getting that from **all** pods in the cluster after the tests finished, and I used that to debug a few issues, but that has proven not so useful in that scenario.

The `describe pod` in case of failures is kept, and I 100% agree with you that those ones should be kept there.

Does that sound reasonable?

tools/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml
fidencio: I will change, thanks!

tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
fidencio: I will change, thanks!

tools/packaging/kata-deploy/runtimeclasses/kata-clh.yaml
```diff
@@ -0,0 +1,13 @@
+---
+kind: RuntimeClass
```
wainersm, line 2: That's a really really good idea @fidencio . Question: why we still need the kata-runtimeClasses.yaml file? Forgot to delete?

tests/integration/kubernetes/gha-run.sh
```diff
@@ -71,7 +71,7 @@ function deploy_kata() {
     sed -i -e "s|quay.io/kata-containers/kata-deploy:latest|${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}|g" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
 
     # Enable debug for Kata Containers
-    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[1].value' "\"yes\""
+    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[1].value' --tag !!str "true"
```
wainersm, line 74: Running on bash, I had to pass '!!str' because !! is going to eval to the last command in the bash's history database. Does it work fine within this script?

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -29,6 +29,41 @@ function print_usage() {
...
+	for shim in "${shims[@]}"; do
+		echo "Deleting the kata-${shim} runtime class"
+		kubectl delete -f /opt/kata-artifacts/runtimeclasses/kata-${shim}.yaml
+	done
+
+
+	if [[ "${CREATE_DEFAULT_RUNTIMECLASS:-"false"}" == "true" ]]; then
+		echo "Deleting the kata runtime class for the default shim (an alias for kata-${default_shim})"
+		cp /opt/kata-artifacts/runtimeclasses/kata-${default_shim}.yaml /tmp/kata.yaml
+		sed -i -e 's/${default_shim}/kata/g' /tmp/kata.yaml
```
wainersm, line 61: This sed doesn't work.

```
[wmoschet@wainer-laptop kata-deploy]$ default_shim="qemu"
[wmoschet@wainer-laptop kata-deploy]$ cp ./runtimeclasses/kata-${default_shim}.yaml /tmp/kata.yaml
[wmoschet@wainer-laptop kata-deploy]$ sed -i -e 's/${default_shim}/kata/g' /tmp/kata.yaml
[wmoschet@wainer-laptop kata-deploy]$ cat /tmp/kata.yaml
---
kind: RuntimeClass
apiVersion: node.k8s.io/v1
metadata:
    name: kata-qemu
handler: kata-qemu
overhead:
    podFixed:
        memory: "160Mi"
        cpu: "250m"
scheduling:
  nodeSelector:
    katacontainers.io/kata-runtime: "true"
```
Now suppose:
```
[wmoschet@wainer-laptop kata-deploy]$ sed -i -e 's/'${default_shim}'/kata/g' /tmp/kata.yaml
[wmoschet@wainer-laptop kata-deploy]$ cat /tmp/kata.yaml
---
kind: RuntimeClass
apiVersion: node.k8s.io/v1
metadata:
    name: kata-kata
handler: kata-kata
overhead:
    podFixed:
        memory: "160Mi"
        cpu: "250m"
scheduling:
  nodeSelector:
    katacontainers.io/kata-runtime: "true"
```

Which is still wrong. I think you want:

```
[wmoschet@wainer-laptop kata-deploy]$ cp ./runtimeclasses/kata-${default_shim}.yaml /tmp/kata.yaml
[wmoschet@wainer-laptop kata-deploy]$ sed -i -e 's/kata-'${default_shim}'/kata/g' /tmp/kata.yaml
[wmoschet@wainer-laptop kata-deploy]$ cat /tmp/kata.yaml 
---
kind: RuntimeClass
apiVersion: node.k8s.io/v1
metadata:
    name: kata
handler: kata
overhead:
    podFixed:
        memory: "160Mi"
        cpu: "250m"
scheduling:
  nodeSelector:
    katacontainers.io/kata-runtime: "true"
```

wainersm: btw, I didn't understand why delete the "kata" runtimeclass.

tools/packaging/kata-deploy/runtimeclasses/kata-clh.yaml
fidencio: Because users may still want to have control and have the old behaviour.  It's there for "backwards compat", let's say.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
wainersm: never mind to the previous question, I thought I was reading the `create_runtimeclasses()`

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -29,6 +29,41 @@ function print_usage() {
...
+
+	for shim in "${shims[@]}"; do
+		echo "Creating the kata-${shim} runtime class"
+		kubectl apply -f /opt/kata-artifacts/runtimeclasses/kata-${shim}.yaml
+	done
+
+	if [[ "${CREATE_DEFAULT_RUNTIMECLASS:-"false"}" == "true" ]]; then
+		echo "Creating the kata runtime class for the default shim (an alias for kata-${default_shim})"
+		cp /opt/kata-artifacts/runtimeclasses/kata-${default_shim}.yaml /tmp/kata.yaml
+		sed -i -e 's/${default_shim}/kata/g' /tmp/kata.yaml
```
wainersm, line 43: See the comment below.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
fidencio: You're correct!

tests/integration/kubernetes/gha-run.sh
fidencio: Oh, interesting.  Let me fix it.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -381,6 +381,14 @@ function reset_runtime() {
 }
 
 function main() {
+	echo "Environment variables passed to this script"
+	echo "* NODE_NAME: ${NODE_NAME}"
+	echo "* DEBUG: ${DEBUG}"
```
wainersm, line 386: Running with `set -o nounset` and `DEBUG` can be unset (because in some `if`s you set a default value `${DEBUG:-"false"}`). Don't know if the other variables are mandatory.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -71,7 +71,7 @@ function deploy_kata() {
     sed -i -e "s|quay.io/kata-containers/kata-deploy:latest|${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}|g" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
 
     # Enable debug for Kata Containers
-    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[1].value' "\"yes\""
+    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[1].value' --tag "!!str" "true"
```
wainersm, line 74: it should be in single-quotes '!!str' for bash not interpret it

.github/workflows/kata-runtime-classes-sync.yaml
```diff
@@ -0,0 +1,36 @@
+on:
```
wainersm, line 1: When you said in the other comment we would keep the all-in-one file I immediately thought "It can get un-sync easily". So That's a good idea to create a checker for this, but I think that using an action is too much. I hear voices in my mind, they say "static-checker tool.. static-checker tooolllll".

fidencio: Hmm.  There's no way I'm going to touch the static-checker right now, at least not till it gets moved to main and maybe even refactored. :-)

For now, can we just live with the test we have? :-)

tests/integration/kubernetes/gha-run.sh
fidencio: Fixed, thanks!

And, hey, I've been exposed as someone not using bash.  Well, I like the taste of fish, that's it. :-)

tools/packaging/kata-deploy/scripts/kata-deploy.sh
fidencio: Fixed, thanks!

.github/workflows/kata-runtime-classes-sync.yaml
wainersm: I am okay with that. I just suggest to open an issue to track this as a possible improvement.

wainersm (approved):
@fidencio LGTM. Thanks for addressing my comments!

Issue #7416, opened 2023-07-21
kata-deploy-binaries: Take module_dir into consideration when installed the cached kernel


fidencio:
`module_dir` has been passed to the function but was never assigned to a var, leading to errors when trying to use it.

PR #7417, opened 2023-07-21
kata-deploy-binaries: kernel_cache: Take module_dir into account

fidencio:
`module_dir` has been passed to the function but was never assigned to a var, leading to errors when trying to use it.

Fixes: #7416


dborquez (approved):
lgtm, thank you @fidencio

stevenhorsman (approved):
LGTM. Thanks

PR #7421, opened 2023-07-24
Add loongarch64 support in agent

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

Fixes: https://github.com/kata-containers/kata-containers/issues/7423

### DIFF
--- a/src/agent/src/linux_abi.rs
+++ b/src/agent/src/linux_abi.rs
@@ -12,6 +12,7 @@ use std::fs;
 
 pub const SYSFS_DIR: &str = "/sys";
 #[cfg(any(
+    target_arch = "loongarch64",
     target_arch = "powerpc64",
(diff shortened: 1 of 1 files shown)


bergwolf:
What about runtime and vmm?

zhaixiaojuan:
> What about runtime and vmm?

The basic functions have been transplanted, and the test is currently running, and the code will be uploaded after the test is completed

PR #7426, opened 2023-07-24
gha: ci: Add no-op nydus tests to our CI

fidencio:
Let's move all the nydus content from the `tests`  repo to our GHA, and create a no-op test for those.

The tests right now are being triggered but not being executed, and a different PR doing clean-ups and enabling the tests will be raised, as this one is basically a placeholder that must get in first.

There's no need to add the `ok-to-test` label as changes in the yaml file for the ci related workflows won't be triggered.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,3 +81,10 @@ jobs:
     with:
       tarball-suffix: -${{ inputs.tag }}
       commit-hash: ${{ inputs.commit-hash }}
+
+  run-nydus-tests:
+    needs: build-kata-static-tarball-amd64
+    uses: ./.github/workflows/run-nydus-tests.yaml
+    with:
+      tarball-suffix: -${{ inputs.tag }}
+      commit-hash: ${{ inputs.commit-hash }}
--- a/.github/workflows/run-nydus-tests.yaml
+++ b/.github/workflows/run-nydus-tests.yaml
@@ -1,0 +1,42 @@
+name: CI | Run nydus tests
+on:
+  workflow_call:
+    inputs:
+      tarball-suffix:
+        required: false
+        type: string
+      commit-hash:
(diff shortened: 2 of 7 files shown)


studychao (approved):
LGTM
But a simple question, since I'm new to Github Action, how could I know that the tests "are being triggered but not being executed" ?

fidencio:
> But a simple question, since I'm new to Github Action, how could I know that the tests "are being triggered but not being executed" ?

That's a good question, so let me go for the complete answer here.

The event we use for GitHub for the CI jobs is called `pull_request_target`, this event, unfortunately, does **not** allow us to make changes in the actions themselves (the .yaml files we're adding) during the same PR it's testing.

With this in mind, it lead us to have to add a basic "skeleton" for the GHA, and scripts we call.  Once the action is merged, we can modify the scripts and make sure the scripts will be running the tests, and changes will be taken into consideration.

But this does not answer your question yet.

The reason why the tests will be triggered is because we added them as part of the jobs that we run, here:
* https://github.com/kata-containers/kata-containers/pull/7426/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR85-R90

The reason why the tests will **not** be executed, is because we short-circuit them to **not** execute yet, here ... and oops, it seems I forgot to add a file to the commit.  Let me add the file and keep the explanation in the next comment. :-)

fidencio:
Okay, now that the file has been added, you will be able to see that we short-circuit the CI to **not** run the test, as shown here:
* https://github.com/kata-containers/kata-containers/pull/7426/files#diff-a2e00bbbbd492883876f17b091e9cfd845093d8a4d34a107a4a424f85aaaa2c6R16-R34

Does this answer your question, @studychao?

zvonkok (approved):
LGTM nydus++

PR #7437, opened 2023-07-25
cache: kernel: Fix kernel caching

jepio:
Kernel caching got broken for various configs by mismatched commits in the version string. SEV had additional breakage because the two cache assets used different version strings altogether. Fix both issues.


fidencio (approved):
lgtm, thanks @jepio!

PR #7440, opened 2023-07-25
CC | Adjust cache URL for TDX components

fidencio:
As CCv0 and main branches have the different versions of the TDX components, we need to adjust the URL where we're getting the cached artefacts from.


jepio (approved):
lgtm

fidencio:
All tests passed, finally.

PR #7453, opened 2023-07-25
tests: Fix `k8s-job` test

sprt:
**tests: k8s: Clean up node debuggers after running**

This deletes node debugger pods after execution since their presence may
affect tests that assume only test workloads pods are present.

For example, in `k8s-job` we wait for *any* pod to be in the `Succeeded`
state before proceeding, which causes failures.

Fixes: https://github.com/kata-containers/kata-containers/issues/7452

---

**tests: gha: Add timeout to cluster creation**

This has been intermittently taking a while lately so let's add a
timeout.

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -175,6 +175,7 @@ function get_nodes_and_pods_info() {
 	echo "::endgroup::"
     fi
     kubectl debug $(kubectl get nodes -o name) -it --image=quay.io/kata-containers/kata-debug:latest
+    kubectl get pods -o name | grep node-debugger | xargs kubectl delete
 }
 
 function main() {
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @sprt!

tests/integration/kubernetes/gha-run.sh
```diff
@@ -175,6 +175,7 @@ function get_nodes_and_pods_info() {
 	echo "::endgroup::"
     fi
     kubectl debug $(kubectl get nodes -o name) -it --image=quay.io/kata-containers/kata-debug:latest
+    kubectl get pods -o name | grep node-debugger | xargs kubectl delete
```
zvonkok, line 178: @fidencio @sprt   We could have a label for the kata-debug pods then we could omit the `grep` and `xargs` for a cleaner 
```
kubectl delete pods -l name=kata-debug-<whatever>
```

zvonkok (approved):
LGTM, just a small nit for a future improvement.

tests/integration/kubernetes/gha-run.sh
sprt: I agree it's not super clean, but you can't pass a label to `kubectl debug`, and getting the exact name of a debug pod would also look ugly.

PR #7455, opened 2023-07-25
tests: QoL improvements for running tests locally

sprt:
This is a series of patches to make running tests outside of GHA more pleasant.

Fixes: #7454

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -12,6 +12,8 @@ kubernetes_dir="$(dirname "$(readlink -f "$0")")"
 source "${kubernetes_dir}/../../common.bash"
 tools_dir="${repo_root_dir}/tools"
 
+AZ_RG="${AZ_RG:-kataCI}"
+
 function _print_cluster_name() {
     short_sha="$(git rev-parse --short=12 HEAD)"
     echo "${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64"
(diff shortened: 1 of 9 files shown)


tests/integration/kubernetes/gha-run.sh
```diff
@@ -121,6 +121,11 @@ function cleanup() {
     echo "Gather information about the nodes and pods before cleaning up the node"
     get_nodes_and_pods_info "yes"
 
+    if [ "${platform}" = "aks" ]; then
+        delete_cluster
+        return
+    fi
```
fidencio, line 131: Please, allow kata-deploy to be uninstalled from aks clusters, so we can know whether this actually working or not.

If you want to add a short-circuit to go directly to the cliuster deletion, I'm fine, but in the CI runs I'd prefer not changing the folow.

fidencio:
@sprt, **very nice series!!!**

I've left one comment, once that's adressed I'll approve it.

tests/integration/kubernetes/gha-run.sh
fidencio: Ouch, it seems this is needed as without it we've been leaving pods behind.

I am approving it, and later on we can improve it if needed.

fidencio (approved):
lgtm, thanks @sprt!

jepio (approved):
looks ok

tests/integration/kubernetes/gha-run.sh
sprt: > Please, allow kata-deploy to be uninstalled from aks clusters, so we can know whether this actually working or not.

This makes sense, hadn't thought about this.

> Ouch, it seems this is needed as without it we've been leaving pods behind.

Don't understand this one actually. Can you explain? Aren't we fine if we delete the cluster in the end?

fidencio: It seems we're failing in a cleanup step and we're never triggering the cluster delete part. :-/

PR #7459, opened 2023-07-26
kata-deploy: Drop privileged security-context

fidencio:
Let's check whether this is really needed.


fidencio:
The tests results shows we cannot drop the privileged bit, at least not for now. :-/

Issue #7460, opened 2023-07-26
CC | kata-deploy.sh is missing executable bit


jepio:
Causes issues running kata-deploy.sh from a pod.

PR #7461, opened 2023-07-26
CC | kata-deploy: Add executable bit to script

jepio:
We need the executable bit set because it is preserved into the runtime-payload-ci image.

Fixes: #7460


fidencio:
Nice, can you also do this for the `main` branch?

fidencio (approved):
lgtm, thanks @jepio!

jepio:
> Nice, can you also do this for the `main` branch?

Main branch doesn't have this issue.

fidencio:
> Main branch doesn't have this issue.

Yeah, I noticed I didn't break it there, thanks for the fix!. :-)

fidencio:
We can merge this one without running the CIs, as the change will only be tested as part of the operator tests.

Issue #7422, opened 2023-07-24
CC | Close the gap between the main and the CCv0 branch


fidencio:
We've been dealing with a lot of conflicts on the merge from main to CCv0, and that's been happening because in the past we took the approach of splitting all components we're building for CCv0.

However, now that we're merging those components back to main, it has proven really problematic and we should close this gap ASAP to avoid wasting time solving a number way higher than needed conflicts when doing those merges.

PR #7463, opened 2023-07-26
cc: kata-deploy: Configure cri-handler="cc"

fidencio:
This has been mistakenly dropped as part of
441399df1f473088703d91cff63161ee79c76502

Fixes: #7422


fidencio:
We don't have to test this as part of the CI, as the CC CI itself doesn't exercise this path.
It'll be tested against the Operator once it's merged.

jepio (approved):

Issue #7464, opened 2023-07-26
gha: dragonball: static-checks-dragonball are landing on all the self-hosted runners


fidencio:
The reason it happens is because the label used (`self-hosted`) is a common label for all the runners.

The labels available for the machine are "self-hosted" "Linux" "X64" "dragonball" , and `dragonball` should be used instead.

PR #7465, opened 2023-07-26
gha: dragonball: Run only on the dragonball labeled machine

fidencio:
Static checks for dragonball are landing on any of the self-hosted runners, and the reason for that is because "self-hosted" was the label selector used.

Let's use "dragonball" instead, as the machine has that label as well.

Fixes: #7464


UnmeshDeodhar (approved):
Thanks for this quick change. LGTM!

wainersm (approved):

sprt (approved):

studychao:
Thanks!

PR #7471, opened 2023-07-27
CC | kata-deploy: Add the runtime-classes that are not yet on main

fidencio:
This is another piece that got dropped as part of
6f552b010c51eb9e6dbcbb21cfa122f5c8d6ad36 and is causing regressions on the operator tests.

Fixes: #7422

This doesn't require the tests to run, as this script is only tested as part of the Operator CI.


zvonkok (approved):
LGTM, thanks you for you effort on this tedious work.

PR #7477, opened 2023-07-27
CC | kata-deploy: Use different shim arrays for different arches

fidencio:
On main we will not have this problem as we can easily configure which
shims will be installed according to an environment variable passed to
the kata-deploy.yaml file.

However, on CCV0, at least for now, we better keep the list of shims
separated by architecture, as we've found out that s390x CoCo Operator
CI is breaking because we try to install a shim that's not even built
for that architecture (dragonball).


BbolroC (approved):
LGTM, thanks @fidencio !

fidencio:
I added the do-not-merge label so both Choi and I can test it locally, but it will only happen Tomorrow.

fidencio:
Again, yet another case where running the CI is not useful as the test will only be checked by the operator CI.

Let's have it merged now, and then check on the operator once the payload is generated after the merge.

PR #7482, opened 2023-07-28
kata-deploy: Fix if-elif-else-statement

fidencio:
We were doing "if - else  if - else", while bash expects "if - elif - else", and that should never have happened in the first place, but it happend as part of b8b73939ea98086e1359f5cb4d5a71d940e960dc

Fixes: #7422

Again, no tests are needed to run as it will only be tested on the operator side.


BbolroC (approved):
LGTM, thanks @fidencio !

Issue #278, opened 2020-06-05
tests: Improve unit test coverage for oci/ for Kata 2.0 agent


jodh-intel:
The following [`2.0-dev` agent branch](https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/agent) files need more unit tests:

```oci/```

See: https://github.com/kata-containers/kata-containers/issues/275

cmaf:
Current coverage is <1%

PR #7483, opened 2023-07-28
runtime-rs: improving io performance using dragonball's vsock fd passthrough

frezcirno:
Currently, in the kata container, every io read/write operation requires an RPC request from the runtime to the agent.
This process involves unnecessary data copying into/from an RPC request/response, which introduces high overhead. In scenarios where containers have multiple process streams, this results in poor performance and additional CPU consumption.

To solve this issue, this PR proposes utilizing the vsock fd passthrough #7585, a newly introduced feature in the Dragonball hypervisor. This feature allows other host programs to pass a file descriptor to the Dragonball process directly as the backend of an ordinary hybrid vsock connection. The detail is depicted in the following diagram:

**Changes Made in this PR**:
1. The runtime-rs now utilizes this feature for container process io. It passes the stdin/stdout/stderr fifo from containerd to Dragonball, eliminating the need for an RPC for each io read/write operation.
2. The agent uses the vsock streams as the child process's stdin/stdout/stderr in passfd mode, eliminating the need for a pipe to bump data (in non-tty mode).
3. As this is a Dragonball-only improvement, two toml options, `use_passfd_io` and `passfd_listener_port`, are introduced to enable and configure the feature.
4. The vsock implementation in the Dragonball vmm is modified to support container io detach and attach.

**Implementation Details**:
*Agent side*:
1. The agent starts an AF_VSOCK server (the passfd_listener) based on a boot parameter. The server does `accept()` and saves the `(hostport, stream)` pairs for later use.
2. Upon receiving `CreateContainer` or `ExecProcess` requests, the agent gets the port info, finds corresponding streams, and uses them as the child process's stdin/stdout/stderr. Note that if a terminal is required, the agent spawns two tokio tasks to copy from the stdin stream to the term_master, and from the term_master to the stdout stream.

*Runtime-rs side*:
1. When creating a new container (with the container init process to run) or executing a process in an existing container, the runtime-rs opens the process io FIFO provided by containerd and connects to the passfd_listener in the guest. It then passes the fds to Dragonball, and saves the `OK <hostport>`  results.
2. The `CreateContainerRequest` and `ExecProcessRequest` carry the stream's hostport information.

*Dragonball vmm*:
1. A stream can be set in "keep" mode to allow a peer to detach and attach.

**New Protobuf Fields**:
Three extra u32 fields (stdin, stdout, and stderr stream ports) are added to the `CreateContainerRequest` and `ExecProcessRequest` structs.

This PR is part of the student open-source practice program hosted in [GitLink Code Camp](https://www.gitlink.org.cn/glcc/2023/), similar to the GSoC.
cc mentor @lifupan


frezcirno:
> @frezcirno I assumed that there should be some modification to the dragonball vmm, but it turned out to be not the case?

Hi, the main functionality of passfd has merged into dragonball in this [openanolis/dragonbal-sandbox PR #278](https://github.com/openanolis/dragonball-sandbox/pull/278) already, which is done before the original dragonball repo archives and moves into kata. 

Moreover, [another precondition of this PR](https://github.com/lifupan/dragonball-sandbox/commit/2c0087fbce4f3d829be747c56059989a1f4bc7f4), which fixes a passfd type issue, is on the way, so this pr should be merged after that.

bergwolf:
@frezcirno [another precondition of this PR](https://github.com/lifupan/dragonball-sandbox/commit/2c0087fbce4f3d829be747c56059989a1f4bc7f4) has already been merged in dragonball-sandbox. Do you mean that it needs to be ported to kata-containers (now that dragonball is part of kata)?

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
```diff
@@ -218,14 +258,18 @@ impl Process {
     }
 
     pub async fn close_io(&mut self, agent: Arc<dyn Agent>) {
-        self.wg_stdin.wait().await;
+        if let Some(_) = self.passfd_io.take() {
+            // do nothing
+        } else {
+            self.wg_stdin.wait().await;
```
bergwolf, line 264: Do we need a similar logic in the agent in the passfd case? I mean, do we need to always make sure that we drain the stdin steam before closing? Likely we have to handle it in dragonball's hvsock implementation.

bergwolf (commented):
Mostly looks good. Nice PR, thanks! A few comments are inline.

frezcirno:
> @frezcirno [another precondition of this PR](https://github.com/lifupan/dragonball-sandbox/commit/2c0087fbce4f3d829be747c56059989a1f4bc7f4) has already been merged in dragonball-sandbox. 

Sorry, but I don't find this fix in openanolis/dragonball-sandbox or in kata-containers. Is it merged? I searched for the fix change and didn't find a similar commit or PR?

> Do you mean that it needs to be ported to kata-containers (now that dragonball is part of kata)?

Yes, this PR won't work normally without that fix to dragonball.

bergwolf:
@frezcirno oops, I was looking at a different PR. Then could you help port [another precondition of this PR](https://github.com/lifupan/dragonball-sandbox/commit/2c0087fbce4f3d829be747c56059989a1f4bc7f4) to kata-containers ?

src/agent/src/rpc.rs
```diff
@@ -235,7 +237,25 @@ impl AgentService {
         let pipe_size = AGENT_CONFIG.container_pipe_size;
 
         let p = if let Some(p) = oci.process {
-            Process::new(&sl(), &p, cid.as_str(), true, pipe_size)?
+            let (stdin_stream, stdout_stream, stderr_stream, signal_stream) = take_io_streams(
```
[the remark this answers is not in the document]
frezcirno: ok, resolve with the 5th commit

src/libs/protocols/protos/agent.proto
```diff
@@ -88,6 +88,15 @@ message CreateContainerRequest {
...
 	bool sandbox_pidns = 7;
+
+	// These field are the host-side vport numbers of passfd streams
+	// pre-created by runtime-rs, and used as identifiers for the agent
+	// to select the right streams for container's stdin/stdout/stderr. 
+	// Disable the feature by setting the port to 0.
+	uint32 stdin_port = 8;
+	uint32 stdout_port = 9;
+	uint32 stderr_port = 10;
+	uint32 signal_port = 11;
```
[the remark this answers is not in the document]
frezcirno: Explanation on the signal port is added to the corresponding comments.

src/libs/protocols/protos/agent.proto
```diff
@@ -110,6 +119,15 @@ message ExecProcessRequest {
...
 	Process process = 4;
+
+	// These field are the host-side vport numbers of passfd streams
+	// pre-created by runtime-rs, and used as identifiers for the agent
+	// to select the right streams for container's stdin/stdout/stderr. 
+	// Disable the feature by setting the port to 0.
+	uint32 stdin_port = 5;
+	uint32 stdout_port = 6;
+	uint32 stderr_port = 7;
+	uint32 signal_port = 8;
```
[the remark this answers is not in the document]
frezcirno: Explanation on the signal port is added to the corresponding comments.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -69,6 +71,26 @@ impl Container {
...
             .and_then(|linux| linux.resources.clone());
 
+        let toml_config = resource_manager.config().await;
+        let passfd_listener_addr = if toml_config.runtime.use_passfd_io {
+            let agent_sock_addr = agent
+                .agent_sock_addr()
+                .await
+                .context("get agent sock addr")?;
+            let agent_sock_url = Url::parse(&agent_sock_addr).context("parse socket address")?;
+            if agent_sock_url.scheme() != "hvsock" {
```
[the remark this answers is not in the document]
frezcirno: Hello, do you mean that the `if` statement should be removed?

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
```diff
@@ -78,7 +81,44 @@ impl Process {
...
+    }
+
+    pub async fn passfd_io_start(&mut self, agent: Arc<dyn Agent>) -> Result<()> {
+        let wg = WaitGroup::new();
+        let wgw = wg.worker();
+
+        // The stdin, stdout, stderr are all passed to the hvsock (by passfd),
+        // so we need a normal stream as a `signal` to get notified when the
+        // guest side close io.
+        if let Some(passfd_io) = &mut self.passfd_io {
```
lifupan, line 109: Maybe we didn't need this signal stream to wait, since we have no temporary io copy in the io stream path and we just need it in the agent to drain out the pipe when no tty set.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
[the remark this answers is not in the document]
bergwolf: I meant that we can live with current implementation. In future when we add fd socket backend to vhost-vsock, we'll need check passfd_io for vhost-vsock as well.

src/agent/src/rpc.rs
```diff
@@ -1723,6 +1762,44 @@ async fn read_stream(reader: Arc<Mutex<ReadHalf<PipeStream>>>, l: usize) -> Resu
     Ok(content)
 }
 
+async fn take_io_streams(
```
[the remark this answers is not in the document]
frezcirno: ok, done

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -42,10 +43,11 @@ pub struct Container {
     agent: Arc<dyn Agent>,
     resource_manager: Arc<ResourceManager>,
     logger: slog::Logger,
+    passfd_listener_addr: Option<(String, u32)>,
```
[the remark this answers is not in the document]
frezcirno: That's a good idea. This info is stored in the `Hypervisor` implementation now, and a new API is exposed to retrieve the address info when creating a container.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
frezcirno: Not sure if I understand correctly. The signal stream is mainly used for the closing procedure rather than draining out the pipe. Without it, when the container process exited, the runtime-rs won't know that and will wait forever.

In the original io implementation(`start_io_and_wait`), "all io streams disconnect" is the condition that triggers runtime-rs to send a `WaitProcessRequest` to the agent and update the container process's state. In the passfd case, we need some equivalent trigger and I came up with the solution of the signal stream.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
frezcirno: Oh yes, I agree, that's a problem. 

Now I am wondering if contained will really close the stdin fifo before or after sending the `CloseIORequest` rpc? 

If so, the guest af_vsock stream will finally be closed by hvsock implementation, then we just need to delay the closing after `stdin_stream` is drained in the agent.

lifupan:
Hi @frezcirno  

I took a look and found that there are no other problems. There are two questions you need to figure out:
1) Why did you need to wait for io close before waiting for the container process? The root cause of this problem is that you need to wait for all the data in the io stream of the process to be copied before initiating wait. otherwise
      This may cause some io content to be lost. Can't see it on the console. Now we have separated the io stream separately, and there is no need to wait for the io to end on the runtime side. Initiate directly at runtime
       wait process Wait for the process on the kata-agent side to return and it will be ok.

2) close stdin should not need to be sent to the agent. Because when containerd initiates close stdin, the pipe at containerd's end should have been closed. At this time
     Our vsock should be able to sense the peer close on the guest side. At this time, for non-tty cases, the guest side should not need to process anything, because vsock fd is the stdin of the process.
     It will be closed automatically. If it is a tty case, after the guest senses that the vsock peer is closed, it can just close the tty master.

frezcirno:
@lifupan Very thanks for your review, the two problems have been resolved in the latest patch.

For the first, the `signal_{port,stream}` has been removed, now the shim will send a `WaitProcessRequest` to the agent and wait for the response, *immediately after* starting the process, in a separate thread. The agent will only respond after io is done. This ensures that process output data is not lost at exiting. During coding, I compile a timing diagram to display relative details. The changed parts are in red color. This may help.



For the second, `close stdin` won't be sent in passfd io mode now.

In addition, 
* More code comments are added.
* The patch is rebased to the main branch.
* All the code changes are reorganized into 3 main commits

---

Also, here are some basic functional tests on my machine.

Test setting:

```toml
# In configuration.toml
use_passfd_io = true
passfd_listener_port = 1027
```

Test command:

1. Start a container (using tty)
```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm -t docker.io/library/busybox:latest hello
/ # ls
bin    dev    etc    home   lib    lib64  proc   root   run    sys    tmp    usr    var
/ # exit
```

2. Start a container (not using tty)
```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/busybox:latest hello ls
bin
dev
...
```

3. Start a container (not using tty, pipe stdin)
```sh
$ printf "aaa\nbbb\nccc\n" | sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/busybox:latest hello wc
wc: standard input: Resource temporarily unavailable
        3         3        12
```

4. Start a detached container
```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello

# 4.1 Exec command in it (using tty)
$ sudo ctr t exec -t --exec-id asdasd hello sh
/ # ls
bin    dev    etc    home   lib    lib64  proc   root   run    sys    tmp    usr    var
/ # exit

# 4.2 Exec command in it (not using tty)
$ sudo ctr t exec --exec-id asdasd hello ls
bin
dev
...

# 4.3 Exec command in it (not using tty, pipe stdin)
$ printf "aaa\nbbb\nccc\n" | sudo ctr t exec --exec-id asdasd hello wc
wc: standard input: Resource temporarily unavailable
        3         3        12
```

The weird error ("standard input: Resource temporarily unavailable") is output from the guest process(wc), which does not seem to affect functionality.

frezcirno:
I discovered a new problem during the testing. I have added a new commit ("b37a727: dragonball: vsock: properly handle EPOLLHUP/EPOLLERR events") to fix it.

My latest test results:

## Setting

```toml
# In configuration.toml
use_passfd_io = true
passfd_listener_port = 1027
```

## Test cases:

1. Start container (using tty) -- OK

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm -t docker.io/library/busybox:latest hello ls
bin    dev    etc    home   lib    lib64  proc   root   run    sys    tmp    usr    var
```

2. Start container (not using tty) -- OK

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/busybox:latest hello ls
bin
dev
...
```

3. Start a container (not using tty, pipe stdin) -- OK

```sh
$ printf "aaa\nbbb\nccc\n" | sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/busybox:latest hello wc
        3         3        12
$ printf "aaa\nbbb\nccc\n" | sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/busybox:latest hello cat
aaa
bbb
ccc
```

4. Start a detached container (using tty) -- OK

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d -t docker.io/library/busybox:latest hello
$ sudo ctr c ls
CONTAINER    IMAGE                               RUNTIME
hello        docker.io/library/busybox:latest    io.containerd.run.kata.v2
$ sudo ctr t ls
TASK     PID        STATUS
hello    1049204    RUNNING
```

5. Start a detached container (not using tty) -- **FAILED, the init process stopped, working on it**

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello
$ sudo ctr c ls
CONTAINER    IMAGE                               RUNTIME
hello        docker.io/library/busybox:latest    io.containerd.run.kata.v2
$ sudo ctr t ls
TASK     PID        STATUS
hello    1027586    STOPPED
```

6. Exec command (using tty) in a detached container -- OK

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d -t docker.io/library/busybox:latest hello
$ sudo ctr t exec -t --exec-id asdasd hello sh
/ # ls
bin    dev    etc    home   lib    lib64  proc   root   run    sys    tmp    usr    var
/ # exit
```

7. Exec command (not using tty) in a detached container -- OK

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d -t docker.io/library/busybox:latest hello
$ sudo ctr t exec --exec-id asdasd hello ls
bin
dev
...
```

8. Exec command (not using tty, pipe stdin) in a detached container -- OK

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d -t docker.io/library/busybox:latest hello
$ printf "aaa\nbbb\nccc\n" | sudo ctr t exec --exec-id asdasd hello wc
        3         3        12
$ printf "aaa\nbbb\nccc\n" | sudo ctr t exec --exec-id asdasd hello cat
aaa
bbb
ccc
```

src/agent/rustjail/src/container.rs
```diff
@@ -989,13 +990,47 @@ impl BaseContainer for LinuxContainer {
...
             child_stderr = unsafe { std::process::Stdio::from_raw_fd(pseudo.slave) };
+
+            if let Some(proc_io) = &mut p.proc_io {
+                if let Some(mut stdin_stream) = proc_io.stdin.take() {
+                    let mut term_master = unsafe { File::from_raw_fd(pseudo.master) };
+                    let logger = logger.clone();
+                    tokio::spawn(async move {
+                        let res = tokio::io::copy(&mut stdin_stream, &mut term_master).await;
+                        debug!(logger, "copy from stdin to term_master end: {:?}", res);
+                        std::mem::forget(term_master); // Avoid auto closing of term_master
```
lifupan, line 1001: I had noticed that in tty mode, the runtime side didn't send the close io rpc request, and here you "std::mem::forget(term_master)", thus the tty master would closed, which mean the container process's stdin didn't  close. So was it right to close the term master here?

lifupan (commented):
Hi @frezcirno 

Most of the PR LGTM, and only two comments as below.

bergwolf:
@frezcirno Please add some documents about this feature, with [your nice graph in the comment](https://github.com/kata-containers/kata-containers/pull/7483#issuecomment-1745188228), so that other hypervisors (like CLH and QEMU) can follow up later on). Also please add some tests to make sure the feature is validated in CI (e.g., by mimicking the tests done in https://github.com/kata-containers/kata-containers/pull/7483#issuecomment-1746520081)

zvonkok:
@bergwolf The `fd` is only needed at `CreateContainer` time? What about `CreateSandbox` is there any use-case where the `fd` is needed at VM creation time?

bergwolf:
@zvonkok  I think there are mainly two use cases for the fd-based vosck backend:
1. container IO stream
2. container port forwarding

So right now both of them are needed at `CreateContainer` time. Is there any sandbox level use case you have in mind?

frezcirno:
After painstaking investigation, testing, and debugging, the issue discovered last month was finally resolved. All existing issues are clear!

In the process, we discover that the old I/O implementation of runtime-rs does not consider the fact that a container's IO can be detached and re-attached. We designed and finalized this part for the passfd io mode. Many thanks to @lifupan for mentorship and help on the Dragonball vmm.

Below are the latest smoke test results:

### Setup

```toml
# In configuration.toml
use_passfd_io = true
passfd_listener_port = 1027
```

### Test cases

#### 1. Start a container (using terminal) -- OK

Run a normal container with `-t` and a simple command.
- [x] The console should be responsive.
- [x] The command should work normally.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm -t docker.io/library/busybox:latest hello id
uid=0(root) gid=0(root) groups=0(root),10(wheel)

$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm -t docker.io/library/ubuntu:latest hello id
uid=0(root) gid=0(root) groups=0(root)

$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm -t docker.io/library/ubuntu:latest hello sh
# id
uid=0(root) gid=0(root) groups=0(root)
# exit
```

</details>

#### 2. Start a container (not using terminal) -- OK

Run a normal container with a simple command, not using a terminal.
- [x] The console should be responsive.
- [x] The command should work normally.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/busybox:latest hello id
uid=0(root) gid=0(root) groups=0(root),10(wheel)

$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/ubuntu:latest hello id
uid=0(root) gid=0(root) groups=0(root)

$ sudo ctr run --runtime io.containerd.run.kata.v2 --rm docker.io/library/ubuntu:latest hello sh
id
uid=0(root) gid=0(root) groups=0(root)
exit
```

</details>

#### 3. Start a detached container (using terminal) -- OK

Run a detached container, with a terminal.
- [x] The initial task status should be 'RUNNING'.
- [x] CPU usage shouldn't be 100%.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d -t docker.io/library/busybox:latest hello

$ sudo ctr container ls
CONTAINER    IMAGE                               RUNTIME
hello        docker.io/library/busybox:latest    io.containerd.run.kata.v2

$ sudo ctr task ls
TASK     PID        STATUS
hello    (pid number)    RUNNING
```

</details>

#### 4. Reattach to a detached container (using terminal) -- OK

Run a detached container, with a terminal.
Attach to the container.
Kill the `ctr` process to detach.
Attach to the container again.
- [x] Can be attached and re-attached.
- [x] The attached terminal should be responsive.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d -t docker.io/library/busybox:latest hello

$ sudo ctr t attach hello
/ #
/ # id
uid=0(root) gid=0(root) groups=0(root),10(wheel)
/ #
```

Now in another session:
```sh
$ sudo pkill -9 -f ctr
```

Re-attach to the same container:

```sh
Killed
$ sudo ctr t attach hello

/ # id
uid=0(root) gid=0(root) groups=0(root),10(wheel)
/ #

```

</details>

#### 5. Start a detached container (not using terminal) -- OK

Run a detached container, not using a terminal.
- [x] The container status should be 'RUNNING'.
- [x] CPU usage shouldn't be 100%.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello

$ sudo ctr container ls
CONTAINER    IMAGE                               RUNTIME
hello        docker.io/library/busybox:latest    io.containerd.run.kata.v2

$ sudo ctr task ls
TASK     PID        STATUS
hello    (pid number)    STOPPED
```

</details>

#### 6. Reattach to a detached container (not using terminal) -- OK

Run a detached container, not using a terminal.
Attach to the container.
Kill the `ctr` process to detach.
Attach to the container again.
- [x] The container should be able to be attached and re-attached.
- [x] The attached terminal should be responsive.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello

$ sudo ctr t attach hello
echo Hello
Hello
```

Now in another session:
```sh
$ sudo pkill -9 -f ctr
```

Re-attach to the same container:
```sh
Killed
$ sudo ctr t attach hello
echo Hello again
Hello again

```

</details>

#### 7. Execute command (using terminal) in an existing container -- OK

Run a container. Execute simple commands in the container using a terminal.
- [x] Can execute commands in the container.
- [x] The console should be responsive.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello

$ sudo ctr t exec -t --exec-id foobar hello id
uid=0(root) gid=0(root) groups=0(root),10(wheel)

$ sudo ctr t exec -t --exec-id foobar hello sh
/ # id
uid=0(root) gid=0(root) groups=0(root),10(wheel)
/ # exit
```

</details>

#### 8. Execute command (not using terminal) in an existing container -- OK

Run a container. Execute simple commands in the container, not using a terminal.
- [x] Can execute commands in the container.
- [x] The console should be responsive.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello

$ sudo ctr t exec --exec-id foobar hello id
uid=0(root) gid=0(root) groups=0(root),10(wheel)

$ sudo ctr t exec --exec-id foobar hello sh
id
uid=0(root) gid=0(root) groups=0(root),10(wheel)
exit
```

</details>

#### 9. Execute command (not using terminal, pipe stdin) in an existing container -- OK

Run a container. Execute simple commands in the container with pipe as stdin.
- [x] Can execute commands in the container.
- [x] The IO data should not changed.

<details>

```sh
$ sudo ctr run --runtime io.containerd.run.kata.v2 -d docker.io/library/busybox:latest hello

# Word count
$ printf "aaa\nbbb\nccc\n" | sudo ctr t exec --exec-id foobar hello wc
        3         3        12

# Large file count
$ head -c 1000000000 /dev/random > /tmp/input
$ cat /tmp/input | wc -c
1000000000
$ cat /tmp/input | sudo ctr t exec --exec-id foobar hello wc -c
1000000000
$ cat /tmp/input | sudo ctr t exec --exec-id foobar hello cat | wc -c
1000000000

# Large file copy
$ head -c 1000000000 /dev/random > /tmp/input
$ cat /tmp/input | sudo ctr t exec --exec-id foobar hello cat > /tmp/output
$ diff -q /tmp/input /tmp/output
```

</details>

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/passfd_io.rs
```diff
@@ -0,0 +1,129 @@
...
+            let hostport = passfd_connect(uds_path, passfd_port, fin.into())
+                .await
+                .context("passfd")?;
+
+            self.stdin_port = Some(hostport);
+        }
+
+        if let Some(stdout) = &self.stdout {
+            let fout = OpenOptions::new()
+                .write(true)
```
lifupan, line 100: For the stdout and stderr, it's better to open the pipe using nonblocking mode, otherwise, the open would blocked if the other side didn't open it.

lifupan (approved):

bergwolf (approved):
lgtm. Please open an issue to track the missing docs and tests, and add them in a follow-up PR. Thanks!

justxuewei:
Hi @frezcirno. It appears that two required tests were failed in every attempt, see:
- https://github.com/kata-containers/kata-containers/actions/runs/6874426727/job/18743779134?pr=7483
- https://github.com/kata-containers/kata-containers/actions/runs/6874426727/job/18743779405?pr=7483

I'd recommend you run these tests in your local environment to see what is going wrong. Please refer to [this doc](https://docs.google.com/document/d/1A71AJjY7oMefzBF-rdJo9F17yd50ceHQTH7SGkBQKhU/edit#heading=h.jtwfrr5vbwwq) for more information.

frezcirno:
> Hi @frezcirno. It appears that two required tests were failed in every attempt, see:
> 
> * https://github.com/kata-containers/kata-containers/actions/runs/6874426727/job/18743779134?pr=7483
> * https://github.com/kata-containers/kata-containers/actions/runs/6874426727/job/18743779405?pr=7483
> 
> Unfortunately, no detailed logs are printed into console. I'd recommend you run these tests in your local environment to see what is going wrong. Please refer to [this doc](https://docs.google.com/document/d/1A71AJjY7oMefzBF-rdJo9F17yd50ceHQTH7SGkBQKhU/edit#heading=h.jtwfrr5vbwwq) for more information.

Hi @justxuewei. The problem is that the `quay.io/sjenning/nginx:1.15-alpine` image used in the tests relies on procfs to redirect its stdout and stderr, which is not a perfect way and is not supported by this PR.

The `/proc/x/fd/y` is a virtual file representing the file descriptor *y* of the process with pid number *x*. Another process can open it to acquire the same duplicated file. However, if the fd is a socket, the Linux kernel forbids opening it and fails the open() call with ENXIO ("No such device or address"). Ref: [[1]](https://unix.stackexchange.com/a/296029), [[2]](https://bugzilla.kernel.org/show_bug.cgi?id=1360)

In the nginx image, `/var/log/nginx/access.log` (error.log) is a soft link to `/dev/stdout` (stderr), which is another soft link to `/proc/self/fd/1` (/2). In the context of this PR, the stdout and stderr of the process are actually sockets (of vsock type), and blocked by the kernel.

For the CI, maybe we could use another nginx image without using the procfs? AFAIK, the nginx now supports specifying `stdout` or `stderr` directly rather than `/dev/std*` or `/proc/self/fd/*` as [access_log](https://nginx.org/en/docs/http/ngx_http_log_module.html#access_log) and [error_log](https://nginx.org/en/docs/ngx_core_module.html#error_log), maybe it's time to update the config?

Thanks,

justxuewei:
> For the CI, maybe we could use another nginx image without using the procfs? AFAIK, the nginx now supports specifying stdout or stderr directly rather than /dev/std* or /proc/self/fd/* as [access_log](https://nginx.org/en/docs/http/ngx_http_log_module.html#access_log) and [error_log](https://nginx.org/en/docs/ngx_core_module.html#error_log), maybe it's time to update the config?

Hi @frezcirno, thanks for your efforts to investigate this issue. It's ok to me overall, but there is one question remaining: Is this feature enabled by default? If yes, it means some images, like aforementioned `nginx` image, are required to be adapted. Otherwise, pods can't be started as expected.

Wdyt? /cc @lifupan @bergwolf

src/agent/src/passfd_io.rs
```diff
@@ -0,0 +1,63 @@
+use anyhow::Result;
```
lifupan, line 1: You need  add a license header here.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/passfd_io.rs
```diff
@@ -0,0 +1,131 @@
+use anyhow::{anyhow, Context, Result};
```
lifupan, line 1: The same here too, please add a license header.

justxuewei:
Hi @frezcirno. Your pull request possibly breaks CI, since I raised another PR without changes at [here](https://github.com/kata-containers/kata-containers/actions/runs/7623597816/job/20765685648?pr=8881). `run-k8s-tests (ubuntu, dragonball, small)` passed.

I observed that it stucks at one of `Empty dir volumes` and `Empty dir volume when FSGroup is specified with non-root container`. Your pull request somehow breaks those tests.

You could deploy a K8s cluster on local to do the test. I have a document about how to create a test cluster. You can ping me (@Xavier) in Slack if you are interested in.

justxuewei:
@frezcirno Please rebase atop of latest main containing a CI fix.

PR #7484, opened 2023-07-28
kata-deploy: We need shim to an array, not a string

fidencio:
In order to do so, we need the `()` around the `shim_{arch}`.

Fixes: #7422


wainersm (approved):
Makes sense!

fidencio:
Again, no tests needed as it'll only be tested on the Operator side.
I wish it'd be tested here, but let's leave that to the tests on main.

Issue #7218, opened 2023-06-30
runtime-rs: Enhancement of Device PCI Topology


Apokleos:
**Which feature do you think can be improved?**
As the result of devices whose driver is virtio-blk-pci, virtio-net-pci,vfio-pci or other pci devices, are all attached on the PCI Bus, and each of them owns one unique bus_slot_function after device added into VM, and we should specify the bus_slot_funtion to the kata agent device when we do add_device. 

If we want to reach it, one choice is return the BDF directly by VMM, However, not all of VMMs will do this. 

**How can it be improved?**

A feasible solution will be  given to manage the devices on the PCI bus uniformly in the device manager.

**Additional Information**

The issue depends on the #7655 .

@lifupan @bergwolf  @jodh-intel @cmaf

PR #7489, opened 2023-07-30
runtime-rs: add pci topology for pci devices

Apokleos:
It's for kinds of pci devices which will be attached on the pci/pcie bus.

Fixes: #7218

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
@@ -18,6 +18,7 @@ use crate::{
 };
 
 use super::{
+    topology::PCIeTopology,
     util::{get_host_path, get_virt_drive_name, DEVICE_TYPE_BLOCK},
     Device, DeviceConfig, DeviceType,
 };
@@ -121,7 +122,9 @@ impl DeviceManager {
             .context("failed to find device")?;
         let mut device_guard = device.lock().await;
         // attach device
-        let result = device_guard.attach(self.hypervisor.as_ref()).await;
+        let result = device_guard
+            .attach(&mut None::<&mut PCIeTopology>, self.hypervisor.as_ref())
+            .await;
         // handle attach error
         if let Err(e) = result {
             match device_guard.get_device_info().await {
@@ -161,7 +164,10 @@ impl DeviceManager {
     pub async fn try_remove_device(&mut self, device_id: &str) -> Result<()> {
         if let Some(dev) = self.devices.get(device_id) {
             let mut device_guard = dev.lock().await;
-            let result = match device_guard.detach(self.hypervisor.as_ref()).await {
+            let result = match device_guard
+                .detach(&mut None::<&mut PCIeTopology>, self.hypervisor.as_ref())
+                .await
+            {
                 Ok(index) => {
                     if let Some(i) = index {
                         // release the declared device index
--- a/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
@@ -8,6 +8,7 @@ use std::{collections::HashMap, sync::Arc};
 
 use anyhow::{anyhow, Context, Result};
 use kata_sys_util::rand::RandomBytes;
+use kata_types::config::hypervisor::TopologyConfigInfo;
 use tokio::sync::{Mutex, RwLock};
 
 use crate::{
@@ -94,15 +95,20 @@ pub struct DeviceManager {
     devices: HashMap<String, ArcMutexDevice>,
     hypervisor: Arc<dyn Hypervisor>,
     shared_info: SharedInfo,
+    pcie_topology: Option<PCIeTopology>,
 }
 
 impl DeviceManager {
-    pub async fn new(hypervisor: Arc<dyn Hypervisor>) -> Result<Self> {
+    pub async fn new(
+        hypervisor: Arc<dyn Hypervisor>,
+        topo_config: Option<&TopologyConfigInfo>,
+    ) -> Result<Self> {
         let devices = HashMap::<String, ArcMutexDevice>::new();
         Ok(DeviceManager {
             devices,
             hypervisor,
             shared_info: SharedInfo::new().await,
+            pcie_topology: PCIeTopology::new(topo_config),
         })
     }
 
@@ -120,10 +126,11 @@ impl DeviceManager {
             .devices
             .get(device_id)
             .context("failed to find device")?;
+
         let mut device_guard = device.lock().await;
         // attach device
         let result = device_guard
-            .attach(&mut None::<&mut PCIeTopology>, self.hypervisor.as_ref())
+            .attach(&mut self.pcie_topology.as_mut(), self.hypervisor.as_ref())
             .await;
         // handle attach error
         if let Err(e) = result {
@@ -165,7 +172,7 @@ impl DeviceManager {
         if let Some(dev) = self.devices.get(device_id) {
             let mut device_guard = dev.lock().await;
             let result = match device_guard
-                .detach(&mut None::<&mut PCIeTopology>, self.hypervisor.as_ref())
+                .detach(&mut self.pcie_topology.as_mut(), self.hypervisor.as_ref())
                 .await
             {
                 Ok(index) => {
@@ -605,6 +612,7 @@ mod tests {
         BlockConfig, KATA_BLK_DEV_TYPE,
     };
     use anyhow::{anyhow, Context, Result};
+    use kata_types::config::hypervisor::TopologyConfigInfo;
     use std::sync::Arc;
     use tests_utils::load_test_config;
     use tokio::sync::RwLock;
@@ -612,6 +620,7 @@ mod tests {
     async fn new_device_manager() -> Result<Arc<RwLock<DeviceManager>>> {
         let hypervisor_name: &str = "qemu";
         let toml_config = load_test_config(hypervisor_name.to_owned())?;
+        let topo_config = TopologyConfigInfo::new(&toml_config);
         let hypervisor_config = toml_config
             .hypervisor
             .get(hypervisor_name)
@@ -623,7 +632,7 @@ mod tests {
             .await;
 
         let dm = Arc::new(RwLock::new(
-            DeviceManager::new(Arc::new(hypervisor))
+            DeviceManager::new(Arc::new(hypervisor), topo_config.as_ref())
                 .await
                 .context("device manager")?,
         ));
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
@@ -5,28 +5,24 @@
 //
 
 use std::{
-    collections::HashMap,
     fs,
     path::{Path, PathBuf},
     process::Command,
-    sync::{
-        atomic::{AtomicU8, Ordering},
-        Arc, RwLock,
-    },
 };
 
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
-use lazy_static::lazy_static;
 use path_clean::PathClean;
 
 use kata_sys_util::fs::get_base_name;
 
-use crate::device::{
-    hypervisor,
-    pci_path::{PciPath, PciSlot},
-    topology::{do_add_pcie_endpoint, PCIeTopology},
-    Device, DeviceType, PCIeDevice,
+use crate::{
+    device::{
+        pci_path::PciPath,
+        topology::{do_add_pcie_endpoint, PCIeTopology},
+        Device, DeviceType, PCIeDevice,
+    },
+    register_pcie_device, unregister_pcie_device, update_pcie_device, Hypervisor as hypervisor,
 };
 
 pub const SYS_BUS_PCI_DRIVER_PROBE: &str = "/sys/bus/pci/drivers_probe";
@@ -43,35 +39,6 @@ const SYS_CLASS_IOMMU: &str = "/sys/class/iommu";
 const INTEL_IOMMU_PREFIX: &str = "dmar";
 const AMD_IOMMU_PREFIX: &str = "ivhd";
 const ARM_IOMMU_PREFIX: &str = "smmu";
-
-lazy_static! {
-    static ref GUEST_DEVICE_ID: Arc<AtomicU8> = Arc::new(AtomicU8::new(0_u8));
-    static ref HOST_GUEST_MAP: Arc<RwLock<HashMap<String, String>>> =
-        Arc::new(RwLock::new(HashMap::new()));
-}
-
-// map host/guest bdf and the mapping saved into `HOST_GUEST_MAP`,
-// and return PciPath.
-pub fn generate_guest_pci_path(bdf: String) -> Result<PciPath> {
-    let hg_map = HOST_GUEST_MAP.clone();
-    let current_id = GUEST_DEVICE_ID.clone();
-
-    current_id.fetch_add(1, Ordering::SeqCst);
-    let slot = current_id.load(Ordering::SeqCst);
-
-    // In some Hypervisors, dragonball, cloud-hypervisor or firecracker,
-    // the device is directly connected to the bus without intermediary bus.
-    // FIXME: Qemu's pci path needs to be implemented;
-    let host_bdf = normalize_device_bdf(bdf.as_str());
-    let guest_bdf = format!("0000:00:{:02x}.0", slot);
-
-    // safe, just do unwrap as `HOST_GUEST_MAP` is always valid.
-    hg_map.write().unwrap().insert(host_bdf, guest_bdf);
-
-    Ok(PciPath {
-        slots: vec![PciSlot::new(slot)],
-    })
-}
 
 pub fn do_check_iommu_on() -> Result<bool> {
     let element = std::fs::read_dir(SYS_CLASS_IOMMU)?
@@ -479,9 +446,11 @@ impl VfioDevice {
 impl Device for VfioDevice {
     async fn attach(
         &mut self,
-        _pcie_topo: &mut Option<&mut PCIeTopology>,
+        pcie_topo: &mut Option<&mut PCIeTopology>,
         h: &dyn hypervisor,
     ) -> Result<()> {
+        register_pcie_device!(self, pcie_topo)?;
+
         if self
             .increase_attach_count()
             .await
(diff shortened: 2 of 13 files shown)


cmaf:
Hi @Apokleos, any recent updates for this?

Apokleos:
> Hi @Apokleos, any recent updates for this?
>
Hi @cmaf sorry for the late reply. 
I will update it  a few days later. As virtio-fs deivces are still outside the device manager, I need move it into dm before updating it.

Apokleos:
Hi, @amshinde @jodh-intel @cmaf @bergwolf @lifupan 

I've made some improvements to the PCIe topology support for CLH/dragonball. 
We can now fully support devices directly attached to the root bus in CLH/Dragonball (getting the PCI path of PCI devices in the Guest). We can also extend support for Qemu and GPU in the future.
Of course, we can discuss the details of the implementation to find a way that works in different scenarios.

To help everyone understand the context, I think it's important to add some relevant links and the original intention behind this feature.
[runtime-rs: Enhancement of Device PCI Topology](https://github.com/kata-containers/kata-containers/issues/7218)
[runtime-rs: add driver option ](https://github.com/kata-containers/kata-containers/pull/7558)

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,316 @@
...
+// 00/FF not included
+const PCI_BRIDGE_MAX_CAPACITY: u32 = 30;
+
+pub const DRIVER_MMIO_BLK_TYPE: &str = "mmioblk";
+pub const DRIVER_VFIO_PCI_TYPE: &str = "vfio-pci";
+const DEFAULT_PCIE_ROOT_BUS: &str = "pcie.0";
+const DEFAULT_PCIE_ROOT_BUS_ADDRESS: &str = "0000:00";
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub enum BusMode {
```
lifupan, line 87: since this bus mod only related with vfio, so is it better to keep it in vfio.rs?

Apokleos: Thx @lifupan 
I'll need to take a second look at this, because bus mode is also a factor now that we're considering PCI topology.

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,316 @@
...
+        let root_complex = PCIeRootComplex {
+            root_bus: DEFAULT_PCIE_ROOT_BUS.to_owned(),
+            root_bus_address: DEFAULT_PCIE_ROOT_BUS_ADDRESS.to_owned(),
+            root_bus_devices: Arc::new(RwLock::new(HashMap::with_capacity(
+                PCI_BRIDGE_MAX_CAPACITY as usize,
+            ))),
+        };
+
+        match hypervisor_name {
+            HYPERVISOR_DRAGONBALL | HYPERVIOR_CLH => Self {
```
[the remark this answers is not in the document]
Apokleos: More comments to be added to explain why do so.
CLH and Dragonball has similar pci topo, only one root bus, all PCI/PCIe device will be attached onto the root bus directly.  
But Qemu has complex PCIe topo, there's one separated branch to handle it.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -426,11 +428,8 @@ impl VfioDevice {
             _ => None,
         }
     }
-}
 
-#[async_trait]
-impl Device for VfioDevice {
-    async fn attach(&mut self, h: &dyn hypervisor) -> Result<()> {
+    fn set_vfio_device(&mut self) -> Result<&Self> {
```
pmores, line 393: I'm not quite clear as to the return value change, could you explain?  I don't see an `Ok()` return value in the function implementation other than `Ok(self)`, so the function always returns `self` (if it succeeds).  However, seeing as the function already takes `self` as a mutable reference, wouldn't it be more natural if it just mutated `self` as needed and returned `Result<()>` as it did so far?

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -266,15 +266,17 @@ impl VfioDevice {
...
             attach_count: 0,
             bus_mode: VfioBusMode::PCI,
             driver_type,
             config: dev_info.clone(),
             devices,
             device_options,
-        }
+        };
+
+        vfio_device.set_vfio_device().cloned()
```
pmores, line 240: Is the cloning necessary here?  If I understand correctly, `vfio_device` is local to this function so do we mind if it gets moved from into the return value?

src/runtime-rs/crates/hypervisor/src/device/mod.rs
```diff
@@ -49,7 +49,7 @@ impl fmt::Display for DeviceType {
 #[async_trait]
 pub trait Device: std::fmt::Debug + Send + Sync {
     // attach is to plug device into VM
-    async fn attach(&mut self, h: &dyn hypervisor) -> Result<()>;
+    async fn attach(&mut self, h: &dyn hypervisor) -> Result<DeviceType>;
```
pmores, line 56: I'm struggling a bit trying to understand the motivation of this change.  The function already takes `&mut self`, wouldn't it be enough for implementations just to mutate `self`?  Or is it envisioned that an implementation might return a different device than `self`?

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub root_bus_address: String,
+    pub root_bus_devices: HashMap<String, PCIeEndpoint>,
+}
+
+// PCIe Root Port
+#[derive(Clone, Debug, Default)]
+pub struct PCIeRootPort {
+    pub chassis: u8,
+    pub slot: u8,
+}
```
zvonkok, line 89: No multifunction support?

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct PCIeRootComplex {
+    pub root_bus: String,
+    pub root_bus_address: String,
+    pub root_bus_devices: HashMap<String, PCIeEndpoint>,
+}
+
+// PCIe Root Port
+#[derive(Clone, Debug, Default)]
+pub struct PCIeRootPort {
```
zvonkok, line 86: Is it going to be possible to set mem-reserver and pref-64-reserve for the root port? Does dragonball has such settings for the root port? The GPU has "huge" BARs and they need proper IOMMU memory setup. We cannot work with the default 4k window range.

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,240 @@
...
+
+#[derive(Default)]
+pub struct UpStreamPort {
+    pub root_port: PCIeRootPort,
+    pub bus_numer: u8,
+    pub port_id: String,
+}
+
+#[derive(Default)]
+pub struct DownStreamPort {
```
zvonkok, line 99: The same comment here: Is it going to be possible to set mem-reserve and pref-64-reserve for the root port? Does dragonball has such settings for the root port? The GPU has "huge" BARs and they need proper IOMMU memory setup. We cannot work with the default 4k window range.

zvonkok:
I am missing also a function to get the size of the PCI BARs of a device so we can setup the right IOMMU sizes, see go-runtime `getBARsMaxAddressableMemory` this should be done per vfio device that is attached. 
For easier management each root-port could be set to those numbers but it would be more efficient to have a correct mapping from VFIO device to pcie root-port and corresponding settings.

zvonkok:
We can start with PCIe root-port support and add later proper PCIe switch support.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
Apokleos: Thx @pmores, good point! I changed the set_vfio_device -> Result<&Self> to set_vfio_device -> Result<()>, which reduces the use of clone.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: Thx @zvonkok 
1. Setting mem-reserve and pref-64-reserve is to be supported, and based on my analysis of kata-runtime, it seems that only qemu needs to set the resource reservation, am I right?
2. AFAIK, Dragonball is a lightweight VMM and now only supports PCIe/PCI devices directly attaching onto the root bus. IOW, it has no such setting for the root port now.  [Discussion about it can be seen at slack ](https://katacontainers.slack.com/archives/C02SN6U9P1D/p1698844006082159)

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: multi-function will be added to PCIeRootPort

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: The fields related to resource reserved will be added.

Apokleos (commented):
Thx @zvonkok 
I think the PCIe Topology design can be roughly divided into two layers: the basic layer and the additional layer. It should also be divided into two phases. 
- The first phase is to implement the basic functionality, with a focus on supporting dragonball/CLH (CLH has its own management method, but it does not conflict, let it go and we make the method compatible with CLH), which has a simple linear bus form and is also the basis for supporting complex buses. 

- The second phase, with the continuous improvement of qemu in runtime-rs, PCIe Topology will be further enhanced to support qemu, which means supporting resource reservation and Huge BARs.

Apokleos:
> I am missing also a function to get the size of the PCI BARs of a device so we can setup the right IOMMU sizes, see go-runtime `getBARsMaxAddressableMemory` this should be done per vfio device that is attached. For easier management each root-port could be set to those numbers but it would be more efficient to have a correct mapping from VFIO device to pcie root-port and corresponding settings.
>
Yeah, Yes, I think this function is essential. But the function will call methods: New()  and GetAllDevices() which located in go-nvlib. I read your document about [VRA](https://github.com/kata-containers/kata-containers/blob/main/docs/design/kata-vra.md#pci-express-virtual-p2p-approval-capability) some time ago, and it mentioned `go-nvlib`. I want to confirm that is there a Rust version of nvlib?

src/runtime-rs/crates/hypervisor/src/device/mod.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_blk.rs
```diff
@@ -152,4 +162,45 @@ impl Device for BlockDevice {
             }
         }
     }
+
+    async fn register_pcie_device(&mut self, topology: &mut PCIeTopology) -> Result<()> {
```
quanweiZhou, line 166: The functions `register_pcie_device()` and `unregister_pcie_device()` are similar to `vhost_user_blk.rs`. It is recommended to optimize them.

src/runtime-rs/crates/hypervisor/src/device/topology.rs
zvonkok: You're right only QEMU needs that for now. We should keep in mind that some other HV may need this as well. If we can make it generic this would be great. 
Understood so this PR is to get Dragonball working.

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -112,49 +119,104 @@ impl DeviceManager {
...
+        // register pcie device into pcie topology
+        device
+            .lock()
+            .await
+            .register_pcie_device(&mut self.pcie_topology)
+            .await
+            .context("register pcie device failed")?;
+
+        // attach device and do update pcie device or handle error.
+        let result = device.lock().await.attach(self.hypervisor.as_ref()).await;
```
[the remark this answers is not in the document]
Apokleos: Thx @quanweiZhou Fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_blk.rs
Apokleos: Thx, Fixed

Apokleos:
> > The second phase, with the continuous improvement of qemu in runtime-rs, PCIe Topology will be further enhanced to support qemu, which means supporting resource reservation and Huge BARs.
> 
> Great keep me in the loop and we can test the QEMU part from my side.
>
> > Yeah, I think this function is essential. But the function will call methods: New() and GetAllDevices() which located in go-nvlib. I read your document about [VRA](https://github.com/kata-containers/kata-containers/blob/main/docs/design/kata-vra.md#pci-express-virtual-p2p-approval-capability) some time ago, and it mentioned go-nvlib. I want to confirm that is there a Rust version of nvlib?
> 
> This is on the roadmap, thanks for mentioning it, added as AI to my list.
>
Thx, Ok!

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: Yes, It's designed to cover all VMMs, not dedicated for some VMM, but now it's the first step to support dragonball and keep compatibility with CLH. And the next, I'll plan to support Qemu.

quanweiZhou (approved):
Thanks, @Apokleos, LGTM.

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,293 @@
...
+
+pub trait PCIeDevice: Send + Sync {
+    fn device_id(&self) -> &str;
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct PCIeEndpoint {
+    // device_id for device in device manager
+    pub device_id: String,
+    pub pci_path: PciPath,
```
studychao, line 127: miss comment?

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,293 @@
...
+            );
+            if v.device_id() == device_id {
+                return true;
+            }
+        }
+
+        false
+    }
+
+    pub fn do_inert_or_update(&mut self, pciep: &mut PCIeEndpoint) -> Result<PciPath> {
```
studychao, line 306: typo: insert

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+impl PCIeTopology {
+    // As some special case doesn't support PCIe devices, there's no need to build a PCIe Topology.
+    pub fn new(config_info: Option<&TopologyConfigInfo>) -> Option<Self> {
+        // if config_info is None, it will return None.
+        let topo_config = config_info?;
+
+        let root_complex = PCIeRootComplex {
+            root_bus: DEFAULT_PCIE_ROOT_BUS.to_owned(),
```
studychao, line 213: seems like all value in root_complex is default ones, can the default value cover all cases like the complex topology in qemu?

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,293 @@
...
+            } else {
+                target_device = Some((*k).to_string());
+                false
+            }
+        });
+
+        target_device
+    }
+
+    pub fn update_device(&mut self, ep: &PCIeEndpoint) -> Option<PciPath> {
```
studychao, line 270: update_device here means that device_id remains the same but pci_addr needs to be changed?

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,293 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/*
+The design Origins from https://github.com/qemu/qemu/blob/master/docs/pcie.txt
+
+The flatten PCIe topology. Root Bus (pcie.0)
+pcie.0 bus (Root Complex)
+----------------------------------------------------------------------------
```
studychao, line 12: this graph is kinda different from `https://github.com/qemu/qemu/blob/master/docs/pcie.txt`
Does this mean we are not gonna support device like `PCIe Root Port `, `PCIe-PCI Bridge` or `pxb-pcie`

studychao (commented):
hi @Apokleos , thanks for your well-organized PR. leave some comment here and I'll take a deeper look next morning

src/runtime-rs/crates/hypervisor/src/device/topology.rs
```diff
@@ -0,0 +1,293 @@
+//
+// Copyright (c) 2019-2023 Ant Group
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/*
+The design Origins from https://github.com/qemu/qemu/blob/master/docs/pcie.txt
```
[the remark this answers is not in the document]
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: Comments added to explain it

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: more will be added when qemu is implemented in runtime-rs

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: comments added to explain it.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -44,35 +40,6 @@ const INTEL_IOMMU_PREFIX: &str = "dmar";
 const AMD_IOMMU_PREFIX: &str = "ivhd";
 const ARM_IOMMU_PREFIX: &str = "smmu";
 
-lazy_static! {
```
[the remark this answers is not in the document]
Apokleos: Since the macro `update_pcie_device` can achieve the same functionality, the previous related code is no longer needed, so it needs to be deleted in this commit.

Apokleos (commented):
Thx @studychao fixed

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -505,6 +505,10 @@ impl Device for VfioDevice {
             None
         };
 
+        if let Some(topology) = pcie_topo {
+            self.unregister(topology).await?;
```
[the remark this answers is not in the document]
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/topology.rs
Apokleos: fixed

studychao (approved):

PR #7492, opened 2023-07-31
kata-deploy: Ensure we cover SHIMS / DEFAULT_SHIM as part of our tests

fidencio:
---

k8s: tests: Take advantage of `SHIMS` and `DEFAULT_SHIM` env vars

We don't have to do any sed to replace the runtimeclass being used by
the moment we start taking advantage of the `DEFAULT_SHIM` environment
variable exposed merged in the previous commits.

---

tests: k8s: Ensure the runtime classes are properly created

With these 2 simple checks we can ensure that we do not regress on the
behaviour of allowing the runtime classes / default runtime class to be
created by the kata-deploy payload.

---

### DIFF
--- a/tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats
+++ b/tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats
@@ -1,0 +1,28 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../common.bash"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	# We expect 2 runtime classes because:
+	# * `kata` is the default runtimeclass created, basically an alias for `kata-${KATA_HYPERVISOR}`.
+	# * `kata-${KATA_HYPERVISOR}` is the other one
+	#    * As part of the tests we're only deploying the specific runtimeclass that will be used, instead of all of them.
+	expected_runtime_classes=2
+}
+
+@test "Test runtimeclasses are being properly created" {
+	# We filter `kata-mshv-vm-isolation` out as that's present on AKS clusters, but that's not coming from kata-deploy
+	current_runtime_classes=$(kubectl get runtimeclasses | grep -v "kata-mshv-vm-isolation" | grep "kata" | wc -l)
+	[[ ${current_runtime_classes} -eq ${expected_runtime_classes} ]]
+	[[ $(kubectl get runtimeclasses | grep -q "${KATA_HYPERVISOR}") -eq 0 ]]
+}
+
+teardown() {
+	kubectl get runtimeclasses
+}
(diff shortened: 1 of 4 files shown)


stevenhorsman:
So just to double-check I'm understanding correctly, as I don't think I saw the `CREATE_DEFAULT_RUNTIMECLASS` get added - is the idea that if that is set to true then kata-deploy will create a runtimeClass of `kata` that points to the default shim?

Then this update allows us to avoid overwriting the runtimeClassName in the test yamls?

tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2018 Intel Corporation
```
stevenhorsman, line 3: nit: Copyright year

fidencio:
> So just to double-check I'm understanding correctly, as I don't think I saw the `CREATE_DEFAULT_RUNTIMECLASS` get added - is the idea that if that is set to true then kata-deploy will create a runtimeClass of `kata` that points to the default shim?
> 
> Then this update allows us to use this to avoid overwriting the runtimeClassName in the test yamls?

You got it correctly, 100% correctly.

tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats
```diff
@@ -0,0 +1,26 @@
...
+
+setup() {
+	expected_runtime_classes=2
+}
+
+@test "Test runtimeclasses are being properly created" {
+	echo "FIDENCIO | DEBUG"
+	current_runtime_classes=$(kubectl get runtimeclasses | grep "kata" | wc -l)
+	echo "FIDENCIO | current_runtime_classes: $current_runtime_classes"	
+	echo "FIDENCIO | expected_runtime_classes: $expected_runtime_classes"
```
stevenhorsman, line 19: I think you might have to force the test to fail to get BATS to show the logs? (or alter a setting elsewhere for that for the whole run?)

tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats
```diff
@@ -0,0 +1,23 @@
...
+# Copyright (c) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../common.bash"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	expected_runtime_classes=2
```
stevenhorsman, line 12: Just to double check - the reason we have two expected runtime classes are the one we are testing and `kata` which defaults to map to it?

stevenhorsman (approved):
LGTM. Thanks!

tests/integration/kubernetes/kata-deploy-ensure-runtimec-classes-created.bats
fidencio: Exaclty!  Do you want me to add a comment making it explicit? (I do think it's a good idea)

fidencio: I did.

stevenhorsman: Thanks - that might have been what I was hinting at ;)

zvonkok (approved):
LGTM, Thanks!

Issue #7493, opened 2023-07-31
vfio: vfio_mode=vfio broken if order if devices is incorrect


zvonkok:
If modeVFIO is enabled we need 1st to attach the VFIO control group device `/dev/vfio/vfio` an 2nd the actuall device(s) afterwards.Sort the devices starting with device #1 being the VFIO control group  device and the next the actuall device(s) `/dev/vfio/<group>`

Works: 
```
sudo nerdctl run  --runtime "io.containerd.kata.v2" --device /dev/vfio/vfio --device /dev/vfio/44  ...
```
Does not work 
```
sudo nerdctl run  --runtime "io.containerd.kata.v2" --device /dev/vfio/44 --device /dev/vfio/vfio  ... 
```

PR #7494, opened 2023-07-31
vfio: Fix vfio device ordering

zvonkok:
If modeVFIO is enabled we need 1st to attach the VFIO control group device `/dev/vfio/vfio` an 2nd the actuall device(s) afterwards. Sort the devices starting with device #1 being the VFIO control group device and the next the actuall device(s)
`/dev/vfio/<group>`

Fixes: #7493

### DIFF
--- a/src/runtime/virtcontainers/container.go
+++ b/src/runtime/virtcontainers/container.go
@@ -851,6 +851,21 @@ func (c *Container) checkBlockDeviceSupport(ctx context.Context) bool {
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @zvonkok!

jepio (approved):

jepio:
> If modeVFIO is enabled we need 1st to attach the VFIO control group device /dev/vfio/vfio an 2nd the actual device(s) afterwards.

Why? Like - why is there a dependency on the order during sandbox creation?

Issue #7495, opened 2023-07-31
tests: k8s: Prepare the ground for using NFD as part of our tests


fidencio:
Let's make sure we can rely on the tests passing down whether they want
to be tested using Node Feataure Discovery or not.

Right now, only the TDX job has this option set to "true", all the other
jobs have this option set to "false".

We can and have to merge this one before merging the NFD related patches
as:
1) It causes no harm in exporting this environment variable, but not
   having it used
2) It will allow us to test the NFD after this one is merged, as changes
   in the yaml file, in the case of the pull_request_target event,  are
   not taken into consideration before they're merged

Related: https://github.com/kata-containers/kata-containers/issues/7392

PR #7496, opened 2023-07-31
k8s: Rely on the USING_NFD environment variable passed by the jobs

fidencio:
Let's make sure we can rely on the tests passing down whether they want to be tested using Node Feataure Discovery or not.

Right now, only the TDX job has this option set to "true", all the other jobs have this option set to "false".

We can and have to merge this one before merging the NFD related patches as:
1) It causes no harm in exporting this environment variable, but not
   having it used
2) It will allow us to test the NFD after this one is merged, as changes
   in the yaml file, in the case of the pull_request_target event,  are
   not taken into consideration before they're merged

Fixes: #7495


fidencio:
There's no need to have `ok-to-test` set as the changes in the yaml file are not being taken into consideration before it's merged.  We're setting the `force-skip-ci` label as we're not doing any changes related to the code itself, only to the tests yaml files.

zvonkok (approved):
LGTM, would be great if the other TEEs follow this pattern.

jepio (approved):

Issue #7497, opened 2023-07-31
gha: release: `stage` must be a string


fidencio:
Otherwise we'll face the following error as part of our GHA:
```
The workflow is not valid.
kata-containers/kata-containers/.github/workflows/release-$foo.yaml
(Line: 13, Col: 14): Invalid input, stage is not defined in the
referenced workflow.
```

PR #7498, opened 2023-07-31
gha: release: `stage` must be a string

fidencio:
Otherwise we'll face the following error as part of our GHA:
```
The workflow is not valid.
kata-containers/kata-containers/.github/workflows/release-$foo.yaml
(Line: 13, Col: 14): Invalid input, stage is not defined in the
referenced workflow.
```

Fixes: #7497

NOTE: I'm adding the `force-skip-ci` label as the changes here are only touching the yaml file used for the release, and it can only be tested after being merged.  Running the tests here would be a waste of resource, and they would **NOT** test anything related to this PR.


zvonkok (approved):
LGTM, thanks!

stevenhorsman (approved):
LGTM. Thanks

Issue #7509, opened 2023-08-01
ci: k8s: Do not fail in any of "gathering info" part of the cleanup


fidencio:
Otherwise the Azure machines won't be deleted.

PR #7510, opened 2023-08-01
ci: k8s: Do not fail when gathering info on AKS nodes

fidencio:
Otherwise the VM deletion may not delete, leaving us with several machines behind.

Fixes: #7509


jepio (approved):

stevenhorsman (approved):
LGTM

Issue #7518, opened 2023-08-02
ci: gha: Also run on stable branches


fidencio:
We're about to cut a rc0 release, so let's make sure that the GHA CI will be running on stable branches.

PR #7519, opened 2023-08-02
ci-on-push: Make the CI also run for the stable-* branches

fidencio:
As we only support one stable branch, it'll be used as part of the stable-3.2 and onwards.

Fixes: #7518


jepio (approved):

zvonkok (approved):
LGTM, thanks!

Issue #7511, opened 2023-08-01
kata-deploy: systemctl does not work inside container


heilerich:
# Description of problem

Since a recent upgrade of host OS and kata-containers to `3.1.3` the kata-deploy DaemonSet does not work as expected anymore. Specifically, it fails to reload the container runtime by executing `systemctl daemon-reload` and `restart`.

# Expected result

Container runtime is reloaded.

# Actual result

systemctl exits with `Running in chroot, ignoring request:` despite the ` /var/run/dbus/system_bus_socket` being mounted correctly.

# Further information

Since systemctl is so great at providing error messages 🙄 I can only guess what is going wrong. I suspect that maybe a version mismatch between the systemctl shipped with the container and systemd running on the host has something to do with this?

# Potential solution

Without investigating the apparent communication problem of systemctl much further a quick fix seems to be just to use the hosts systemctl by just replacing all systemctl commands in the deploy script with `nsenter -t1 -m systemctl ...` since we are running as root anyways. This should also make the whole process more portable, more robust against changes on system upgrades and remove the dependency on the systemctl shipped with the container image.

Is there any reason why this shouldn't be the default?

heilerich:
We are running Flatcar Linux, a CoreOS fork

jepio:
Thank you for being a Flatcar user. We wanted to fix this issue in the past, I just opened a PR to do what you suggested.

PR #7520, opened 2023-08-02
kata-deploy: Use host's systemctl

jepio:
when interacting with systemd. We have occasionally faced issues with compatibility between the systemctl version used inside the kata-deploy container and the systemd version on the host. Instead of using a containerized systemctl with bind mounted sockets, nsenter the host and run systemctl from there. This provides less coupling between the kata-deploy container and the host.

Fixes: #7511

### DIFF
--- a/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
+++ b/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
@@ -14,6 +14,7 @@ spec:
           name: kata-deploy
     spec:
       serviceAccountName: kata-deploy-sa
+      hostPID: true
       containers:
       - name: kube-kata
         image: quay.io/kata-containers/kata-deploy:latest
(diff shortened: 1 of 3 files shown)


tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
```diff
@@ -38,6 +38,7 @@ spec:
           value: "false"
         - name: CREATE_DEFAULT_RUNTIMECLASS
           value: "false"
+        hostPID: true
         securityContext:
           privileged: true
```
fidencio, line 43: Does it still require `privileged`?

jepio: We still do. Kata-deploy logs report:
```
nsenter: cannot open /proc/1/ns/mnt: Permission denied
```
when privileged is off (but kata-deploy doesn't catch the failure?)

fidencio (approved):
lgtm, thanks @jepio!

stevenhorsman (approved):
LGTM. Thanks @jepio - do we need to make an issue in the coco-operator to do updates to match this as well?

fidencio:
> LGTM. Thanks @jepio - do we need to make an issue in the coco-operator to do updates to match this as well?

@jepio told me he started testing this on the CoCo side, so I do assume he'll create an issue for that and take care of:
* CoCo operator (pre-install / install images)
* Enclave CC (let me also loop @mythi here)

fidencio:
@jepio, I'd like to have this one backported to the stable-3.2 branch as well.

PR #7530, opened 2023-08-02
metrics: stop kata components before start a metric test.

dborquez:
This PR kills all kata components before start a new metric test.

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -186,31 +186,20 @@ function clean_env_ctr()
 	if (( count_tasks > 0 )); then
 		die "Can't remove running containers."
 	fi
-
-	kill_kata_components
 }
 
 # Kills running shim and hypervisor components
+# by using the kata-component file name.
 function kill_kata_components() {
-	local kata_bin_dir="/opt/kata/bin"
-	local shim_path="${kata_bin_dir}/containerd-shim-kata-v2"
-	local hypervisor_path="${kata_bin_dir}/qemu-system-x86_64"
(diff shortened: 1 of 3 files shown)


tests/common.bash
```diff
@@ -186,31 +186,27 @@ function clean_env_ctr()
 	if (( count_tasks > 0 )); then
 		die "Can't remove running containers."
 	fi
-
-	kill_kata_components
```
fidencio, line 190: I'd say you could still leave it here.

tests/metrics/lib/common.bash
```diff
@@ -192,6 +192,8 @@ function kill_processes_before_start()
 	CTR_PROCS=$(sudo "${CTR_EXE}" t list -q)
 	[[ -n "${CTR_PROCS}" ]] && clean_env_ctr
 
+	kill_kata_components
```
fidencio, line 195: And also add it here. :-)

fidencio (approved):
lgtm, my comment is a nit and doesn't have to be addressed before merging this one.

tests/common.bash
dborquez: The problem is that the purpose of `clean_env_ctr` is only to kill running containers, that's why we moved `kill_kata_components` to the stage where we need to stop kata running processes.

GabyCT:
@dborquez seems like it needs a daemon-reload because we are getting the following failure
```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Issue #7507, opened 2023-08-01
agent: Restrict device access at upper node of container's cgroup


justxuewei:
The OCI spec contains a list of device resources, which affects container's devices cgroup eventually. However, the rules not take affect in containers' upper cgroup node. This might cause some security issues, e.g. the container mounts cgroup and elevates its permission to access all devices in the VM.

Assume that we have a pod whose id is `$pod`, a container whose id is `$cnt`. Their cgroup paths are `/k8s.io/$pod` and `/k8s.io/$cnt` respectively. Let me give you an example, the container needs to access a block device whose major is 2 and minor is 128. An entry `b 2:128 rwm` is printed to console while you execute `cat /sys/fs/cgroup/devices/k8s.io/$cnt/devices.list`. `a *:* rwm` is printed while you execute `cat /sys/fs/cgroup/devices/k8s.io/devices.list`.

Parent's devices cgroup permission should be shrinked to avoid the issues aforementioned. Back to the example, if the container only needs the device and there are no other containers, the `k8s.io`'s `devices.list` is expected to only have one entry: `b 2:128 rwm`.

Issue #7506, opened 2023-08-01
Systemd cgroup supports devices subsystem


justxuewei:
Currently, the systemd cgroup not supports devices subsystem. We are planing to limit access to devices via devices cgroup and implement it with cgroupfs at the very begining stage. The implementation with systemd cgroup depends on this feature.

PR #7531, opened 2023-08-03
agent: Restrict device access at upper node of container's cgroup

justxuewei:
The target is to guarantee that containers couldn't escape to access extra
devices, like vm rootfs, etc.

## Motivation

Assume that there is a cgroup, such as `/A/B`. The `B` is container cgroup,
and the `A` is what we called pod cgroup. No matter what permissions are
set for the container (`B`), the `A`'s permission is always `a *:* rwm`
with current implementation. It leads that containers could acquire
permission to access to other devices in VM that not belongs to themselves. 

## Details

In order to set devices cgroup properly, the order of setting cgroups is
that the pod cgroup comes first and the container cgroup comes after.

The `Sandbox` has a new field, `devcg_info`, to save cgroup states. To
avoid setting container cgroup too early, an initialization should be done
carefully. `inited`, one of the states, is a boolean to indicate if the pod
cgroup is initialized. If no, the pod cgroup should be created firstly, and
set default permissions. After that, the pause container cgroup is created
and inherits the permissions from the pod cgroup.

If whitelist mode which allows containers to access all devices in VM is
enabled,  then device resources from OCI spec are ignored.

## Limitation

This feature not supports systemd cgroup and cgroup v2, since:

- Systemd cgroup implemented on Agent hasn't supported devices subsystem so
far, see: https://github.com/kata-containers/kata-containers/issues/7506.
- Cgroup v2's device controller depends on eBPF programs, which is out of
scope of cgroup.

Fixes: #7507


src/agent/rustjail/src/cgroups/mock.rs
```diff
@@ -72,7 +75,11 @@ impl CgroupManager for Manager {
 }
 
 impl Manager {
-    pub fn new(cpath: &str) -> Result<Self> {
+    pub fn new(
+        cpath: &str,
+        _spec: &Spec,
+        _devcg_info: Option<Arc<RwLock<DevicesCgroupInfo>>>,
```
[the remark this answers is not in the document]
justxuewei: We have an annotation to switch `crate::cgroups::mock::Manager` for testing:

```rust
#[cfg(not(test))]
use crate::cgroups::fs::Manager as FsManager;
#[cfg(test)]
use crate::cgroups::mock::Manager as FsManager;
```

The `crate::cgroups::fs::Manager` has two new arguments in this patch. I want the signature of the function to look like same as the real `Manager`.

quanweiZhou (commented):
Thanks, @justxuewei, few comments.

justxuewei:
Hi @quanweiZhou, I've push a new commit to address the issues mentioned in your comments. PTAL, thanks!

quanweiZhou (commented):
One more comment.

src/agent/rustjail/src/cgroups/fs/mod.rs
```diff
@@ -18,13 +18,14 @@ use cgroups::{
     DeviceResource, HugePageResource, MaxValue, NetworkPriority,
 };
 
-use crate::cgroups::Manager as CgroupManager;
+use crate::cgroups::{is_all_devices_rule, Manager as CgroupManager};
+#[cfg(not(test))]
```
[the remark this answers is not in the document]
justxuewei: I tried to do this, but no luck here. In order to remove the annotations, an unit test for `default_allowed_devices()` was added in the latest commit. PTAL, thanks!

quanweiZhou (approved):
Thanks, @justxuewei, LGTM.

src/agent/rustjail/src/cgroups/fs/mod.rs
```diff
@@ -1001,13 +996,135 @@ pub fn get_mounts(paths: &HashMap<String, String>) -> Result<HashMap<String, Str
...
+    ) -> Result<Self> {
+        let (paths, mounts) = Self::get_paths_and_mounts(cpath).context("Get paths and mounts")?;
+
+        // Do not expect poisoning lock
+        let mut devices_group_info = devcg_info.as_ref().map(|i| i.write().unwrap());
+        let pod_cgroup: Option<Cgroup>;
+
+        if let Some(devices_group_info) = devices_group_info.as_mut() {
+            // Cgroup path of parent of container
+            let pod_cpath = PathBuf::from(cpath)
```
lifupan, line 1025: what  about  container cgroup didn't have a parent path? such as "/container_cgroup"?

src/agent/rustjail/src/cgroups/fs/mod.rs
```diff
@@ -1020,23 +1137,144 @@ impl Manager {
...
+        };
+        cgroup.apply(&res)?;
+
+        Ok(())
+    }
+
+    /// Check if OCI spec contains a whiltlist rule.
+    ///
+    /// # Returns
+    ///
```
lifupan, line 1218: So what's the whitelist and blacklist? Maybe the white list is: deny all first and then allow the specified devices one by one; and the black list is on the opposite side: allow all first, and then deny the specified devices one by one.

justxuewei: Fixed. The previous `whitelist` is replaced with `allowed_all`. The `blacklist` is replaced with `whitelist`.

src/agent/rustjail/src/cgroups/fs/mod.rs
justxuewei: Fixed. Device cgroup won't work if cgroup path is "/" or "/{anything}".

lifupan (approved):

shaoyezh:
Hi, I am trying to allow the container to have permission to some specific devices in VM. Is there a way to configure it？

shaoyezh:
> @shaoyezh Are you using a privileged container? Could you elaborate a bit more to let me know how you are using it? Plus,I suggest you file a new issue to track the problem, where you can provide us with more details about how to reproduce it. 

Thank you for relying! I created an issue at #10392. Yes,  I used privileged container. The device is passed through to the container by using vfio. I work around this by adding my device to the DEFAULT_DEVICE. I will post how I made my change tmr.

Issue #7532, opened 2023-08-03
agent: 'async' keyword of `Namespace::setup()` is no longer needed


ManaSugi:
[`Namespace::setup()`](https://github.com/kata-containers/kata-containers/blob/7d1c48c88118ded33b890f8b8fc6f0758dd0a6d9/src/agent/src/namespace.rs#L82C5-L152C2) function has been defined an `async` function but 
it doesn't call any `async` task, so we can remove the `async` keyword.

```rust
 pub async fn setup(mut self) -> Result<Self> {
        ...

        new_thread
            .join()
            .map_err(|e| anyhow!("Failed to join thread {:?}!", e))??;

        Ok(self)
    }
}
```

Before the https://github.com/kata-containers/kata-containers/commit/ea1a173854af892e92af95790b5726d64d04edf0 was introduced, the function had used a `tokio` async thread, but currently it uses a normal thread.

PR #7533, opened 2023-08-03
agent: Remove unneeded 'async' keyword

ManaSugi:
Remove unneeded the 'async' keyword of `Namespace::setup()` because it uses just a normal thread, not an asynchronous one. The `setup` function haven't used the tokio async thread since the https://github.com/kata-containers/kata-containers/commit/ea1a173854af892e92af95790b5726d64d04edf0 was introduced, so the `async` keyword is no longer needed.

Fixes: #7532

### DIFF
--- a/src/agent/src/namespace.rs
+++ b/src/agent/src/namespace.rs
@@ -79,7 +79,7 @@ impl Namespace {
     // Note, pid namespaces cannot be persisted.
     #[instrument]
     #[allow(clippy::question_mark)]
-    pub async fn setup(mut self) -> Result<Self> {
+    pub fn setup(mut self) -> Result<Self> {
         fs::create_dir_all(&self.persistent_ns_dir)?;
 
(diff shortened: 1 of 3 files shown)


GabyCT (approved):

bergwolf:
It seems that we should use the [oneshot channel](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html) and `await` on it instead of blocking on thread joining, and thus namespace setup stays `async`.

bergwolf:
@ManaSugi Indeed, let's just remove them now. Thanks.

bergwolf:
@ManaSugi Could you rebase the PR so that we can merge it once CI is green?

bergwolf (approved):
lgtm, please rebase it though. Thanks!

ManaSugi:
@bergwolf rebased and all tests passed.

Issue #7534, opened 2023-08-03
test: redundant clone error showed in the agent unit tests using >=rustc 1.70


ManaSugi:
When I run `make check` under `agent` directory using >=rustc 1.70, the redundant clone error against the `uev_a` showed.
However, these errors are false-negatives because the `uev_a` is referred by the following code (`&uev_a`).
Hence, we should ignore the clippy.

```sh
$ make check
...

    Checking kata-agent v0.1.0 (/***/kata-containers/kata-containers/src/agent)
error: redundant clone
    --> src/device.rs:1482:30
     |
1482 |         let mut uev_b = uev_a.clone();
     |                              ^^^^^^^^ help: remove this
     |
note: cloned value is neither consumed nor mutated
    --> src/device.rs:1482:25
     |
1482 |         let mut uev_b = uev_a.clone();
     |                         ^^^^^^^^^^^^^
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
     = note: `-D clippy::redundant-clone` implied by `-D warnings`

error: redundant clone
    --> src/device.rs:1569:30
     |
1569 |         let mut uev_b = uev_a.clone();
     |                              ^^^^^^^^ help: remove this
     |
note: cloned value is neither consumed nor mutated
    --> src/device.rs:1569:25
     |
1569 |         let mut uev_b = uev_a.clone();
     |                         ^^^^^^^^^^^^^
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone

error: redundant clone
    --> src/device.rs:1594:30
     |
1594 |         let mut uev_b = uev_a.clone();
     |                              ^^^^^^^^ help: remove this
     |
note: cloned value is neither consumed nor mutated
    --> src/device.rs:1594:25
     |
1594 |         let mut uev_b = uev_a.clone();
     |                         ^^^^^^^^^^^^^
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone

error: redundant clone
    --> src/device.rs:1618:30
     |
1618 |         let mut uev_b = uev_a.clone();
     |                              ^^^^^^^^ help: remove this
     |
note: cloned value is neither consumed nor mutated
    --> src/device.rs:1618:25
     |
1618 |         let mut uev_b = uev_a.clone();
     |                         ^^^^^^^^^^^^^
[...]

PR #7535, opened 2023-08-03
agent: Allow clippy::redundant_clone in the unit tests

ManaSugi:
Allow `clippy::redundant_clone` in the agent's unit tests because rustc>=1.70 shows the errors as false-negatives. These `clone()` are required because the following codes refer to the variable, but the clippy analyzes them by mistake, using the conservative and limited approach.
Ref. https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_clone

Fixes: #7534

### DIFF
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -1467,6 +1467,7 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


bergwolf (approved):

studychao (approved):
LGTM

PR #7547, opened 2023-08-04
kata-deploy: Preliminary k0s support

### LINKED ISSUE #7548: K0S support
**Is your feature request related to a problem? Please describe.**

We need to support k0s which is another certified k8s distribution, and is very similar to k3s.

**Additional context**

This is a dummy issue for the PR https://github.com/kata-containers/kata-containers/pull/7547

stevefan1999-personal:
This patch adds k0s deployment support, ~~but this isn't tested yet. I will be running this on my own cluster soon.~~ On my own homelab, this runs pretty well, but I don't have the managed containerd auto config and instead I merged the keys on my own.

This feature requires k0s 1.27.1 or above due to a new mechanism:

> k0s managed dynamic runtime configuration[#](https://docs.k0sproject.io/v1.27.1+k0s.0/runtime/#k0s-managed-dynamic-runtime-configuration)
>
> From 1.27.1 onwards k0s enables dynamic configuration on containerd CRI runtimes. This works by k0s creating a special directory in /etc/k0s/containerd.d/ where user can drop-in partial containerd configuration snippets.
> 
> **k0s will automatically pick up these files and adds these in containerd configuration imports list**. If k0s sees the configuration drop-ins are CRI related configurations k0s will automatically collect all these into a single file and adds that as a single import file. This is to overcome some hard limitation on containerd 1.X versions. Read more at [containerd#8056](https://github.com/containerd/containerd/pull/8056)
> 

And it also needs to be k0s managed.

> User managed containerd configuration[#](https://docs.k0sproject.io/v1.27.1+k0s.0/runtime/#user-managed-containerd-configuration)
> In the default k0s generated configuration there's a "magic" comment telling k0s it is k0s managed:
> 
> \# k0s_managed=true
> If you wish to take over the configuration management remove this line.

Thus for k0s version below 1.27.1 and with a containerd.toml that doesn't have the k0s managed magic, this is effectively a no-op and won't hurt much.

Fixes #7548

### DIFF
--- a/tools/packaging/kata-deploy/README.md
+++ b/tools/packaging/kata-deploy/README.md
@@ -43,6 +43,36 @@ $ kubectl apply -f kata-rbac/base/kata-rbac.yaml
 $ kubectl apply -k kata-deploy/overlays/rke2
 ```
 
+#### [k0s] cluster
+
+For your [k0s](https://k0sproject.io/) cluster, run:
+
+```sh
+$ git clone https://github.com/kata-containers/kata-containers.git
+```
+
+Check and switch to "main", and then run:
+
+```bash
+$ cd kata-containers/tools/packaging/kata-deploy
+$ kubectl apply -f kata-rbac/base/kata-rbac.yaml
+$ kubectl apply -k kata-deploy/overlays/k0s
+```
+
+##### Note
+
+The supported version of k0s is **v1.27.1+k0s** and above, since the k0s support leverages a special dynamic containerd configuration mode:
+
+> From 1.27.1 onwards k0s enables dynamic configuration on containerd CRI runtimes. This works by k0s creating a special directory in /etc/k0s/containerd.d/ where user can drop-in partial containerd configuration snippets.
+> 
+> k0s will automatically pick up these files and adds these in containerd configuration imports list. If k0s sees the configuration drop-ins are CRI related configurations k0s will automatically collect all these into a single file and adds that as a single import file. This is to overcome some hard limitation on containerd 1.X versions. Read more at containerd#8056
+
+However, this would also require a magic string set in the beginning of the line for `/etc/k0s/containerd.toml`:
+
+```
+# k0s_managed=true
+```
+
 #### Vanilla Kubernetes cluster
 
 ##### Installing the latest image
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -74,6 +74,7 @@ function get_container_runtime() {
 	if [ "$?" -ne 0 ]; then
                 die "invalid node name"
 	fi
+
 	if echo "$runtime" | grep -qE 'containerd.*-k3s'; then
 		if host_systemctl is-active --quiet rke2-agent; then
 			echo "rke2-agent"
@@ -84,6 +85,12 @@ function get_container_runtime() {
 		else
 			echo "k3s"
 		fi
+	# Note: we assumed you used a conventional k0s setup and k0s will generate a systemd entry k0scontroller.service and k0sworker.service respectively    
+	# and it is impossible to run this script without a kubelet, so this k0s controller must also have worker mode enabled 
+	elif host_systemctl is-active --quiet k0scontroller; then
+		echo "k0s-controller"
+	elif host_systemctl is-active --quiet k0sworker; then
+		echo "k0s-worker"
 	else
 		echo "$runtime" | awk -F '[:]' '{print $1}'
 	fi
@@ -136,12 +143,17 @@ function configure_cri_runtime() {
 	crio)
 		configure_crio
 		;;
-	containerd | k3s | k3s-agent | rke2-agent | rke2-server)
-		configure_containerd
+	containerd | k3s | k3s-agent | rke2-agent | rke2-server | k0s-controller | k0s-worker)
+		configure_containerd "$1"
 		;;
 	esac
-	host_systemctl daemon-reload
-	host_systemctl restart "$1"
+	if [ "$1" == "k0s-worker" ] || [ "$1" == "k0s-controller" ]; then
+		# do nothing, k0s will automatically load the config on the fly
+		:
+	else
+		host_systemctl daemon-reload
+		host_systemctl restart "$1"
+	fi
 
 	wait_till_node_is_ready
 }
@@ -274,12 +286,15 @@ EOF
 function configure_containerd_runtime() {
 	local runtime="kata"
 	local configuration="configuration"
-	if [ -n "${1-}" ]; then
-		runtime+="-$1"
-		configuration+="-$1"
+	if [ -n "${2-}" ]; then
+		runtime+="-$2"
+		configuration+="-$2"
 	fi
 	local pluginid=cri
-	if grep -q "version = 2\>" $containerd_conf_file; then
+	
+	# if we are running k0s auto containerd.toml generation, the base template is by default version 2
+	# we can safely assume to reference the older version of cri
+	if grep -q "version = 2\>" $containerd_conf_file || [ "$1" == "k0s-worker" ] || [ "$1" == "k0s-controller" ]; then
 		pluginid=\"io.containerd.grpc.v1.cri\"
 	fi
 	local runtime_table="plugins.${pluginid}.containerd.runtimes.$runtime"
@@ -333,10 +348,10 @@ function configure_containerd() {
 	fi
 
 	# Add default Kata runtime configuration
-	configure_containerd_runtime
+	configure_containerd_runtime "$1" 
 
 	for shim in "${shims[@]}"; do
-		configure_containerd_runtime $shim
+		configure_containerd_runtime "$1" $shim
 	done
 }
 
@@ -352,7 +367,7 @@ function cleanup_cri_runtime() {
 	crio)
 		cleanup_crio
 		;;
-	containerd | k3s | k3s-agent | rke2-agent | rke2-server)
+	containerd | k3s | k3s-agent | rke2-agent | rke2-server | k0s-controller | k0s-worker)
 		cleanup_containerd
 		;;
 	esac
@@ -375,8 +390,14 @@ function cleanup_containerd() {
 
 function reset_runtime() {
 	kubectl label node "$NODE_NAME" katacontainers.io/kata-runtime-
-	host_systemctl daemon-reload
-	host_systemctl restart "$1"
+	if [ "$1" == "k0s-worker" ] || [ "$1" == "k0s-controller" ]; then
+		# do nothing, k0s will auto restart
+		:
+	else
+		host_systemctl daemon-reload
+		host_systemctl restart "$1"
+	fi
+
 	if [ "$1" == "crio" ] || [ "$1" == "containerd" ]; then
 		host_systemctl restart kubelet
 	fi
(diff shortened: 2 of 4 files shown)


stevefan1999-personal:
I've deployed the kata-deploy image as [Package kata-deploy (github.com)](https://github.com/users/stevefan1999-personal/packages/container/package/kata-deploy)

stevefan1999-personal:
Hmm, seems like there is a bit of a problem on the generated TOML

```
Aug 04 23:47:19 serv k0s[1087415]: time="2023-08-04 23:47:19" level=info msg="containerd: failed to load TOML from /etc/k0s/containerd.toml: invalid plugin key URI \"cri\" expect io.containerd.x.vx" component=containerd stream=stderr
```

fidencio:
> Hmm, seems like there is a bit of a problem on the generated TOML

This happens because we check whether we can use `version = 2` kind of notations.
It could be very simple to solve by having an extra variable passed there to enforce `version = 2`, for instsance.

Please, take a look here: https://github.com/kata-containers/kata-containers/blob/58073c23f6ac71f28c0f849e97ebcb2074572a17/tools/packaging/kata-deploy/scripts/kata-deploy.sh#L292-L294

stevefan1999-personal:
> > Hmm, seems like there is a bit of a problem on the generated TOML
> 
> This happens because we check whether we can use `version = 2` kind of notations. It could be very simple to solve by having an extra variable passed there to enforce `version = 2`, for instsance.
> 
> Please, take a look here:
> 
> https://github.com/kata-containers/kata-containers/blob/58073c23f6ac71f28c0f849e97ebcb2074572a17/tools/packaging/kata-deploy/scripts/kata-deploy.sh#L292-L294

Yes, seems like we have to check for k0s's case and assume it will be version 2

fidencio:
> Yes, seems like we have to check for k0s's case and assume it will be version 2

Cool, you've addressed it in the exactly the same way I was writing down the suggestion, amazing!

fidencio:
One last thing that I'd like to ask you, though, is, **please**, make kata-deploy fail if the version of `k0s` is not the minimum supported one, and make sure it fails with a reasonable error message.

stevefan1999-personal:
> One last thing that I'd like to ask you, though, is, **please**, make kata-deploy fail if the version of `k0s` is not the minimum supported one, and make sure it fails with a reasonable error message.

Hmm, this would be tricky as we need to parse the major and minor version altogether. I think we could split the dot and compare it parts by parts, however

stevefan1999-personal:
> @stevefan1999-personal, I can see stsatic checks / commit message check not being happy, as it doesn't like the merge commit there.
> 
> Also, please, we need to:
> 
> * Make sure all the commits have a subsystem: `kata-deploy: ...` would do the trick
> * Make sure we have an issue opened on Kata Containers side, and then add `Fixes: #NNNN` to the commit message
> 
> Do you have a strong opinion on whether having all the patches as they are or squashed? I would like to hear from you before suggesting anything.

Maybe I can squash it

fidencio:
Then, I'd prefer having all the patches squashed into a single one, as it's a short enough patch (but very very valuable :-)), and that would make things cleaner as this is very self-contented.

fidencio:
> Hmm, this would be tricky as we need to parse the major and minor version altogether. I think we could split the dot and compare it parts by parts, however

Okay, not going to make this a blocker for now.

As a different commit than everything you'll squash, I'd like to ask you to add docs to this as part of https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/README.md#install-kata-on-a-running-kubernetes-cluster and there you can make a big bold note about the minimal supported version.  WDYT?

tools/packaging/kata-deploy/README.md
```diff
@@ -43,6 +43,34 @@ $ kubectl apply -f kata-rbac/base/kata-rbac.yaml
...
+
+The supported version of k0s is **v1.27.1+k0s** and above, since the k0s support leverages a special dynamic containerd configuration mode:
+
+> From 1.27.1 onwards k0s enables dynamic configuration on containerd CRI runtimes. This works by k0s creating a special directory in /etc/k0s/containerd.d/ where user can drop-in partial containerd configuration snippets.
+> 
+> k0s will automatically pick up these files and adds these in containerd configuration imports list. If k0s sees the configuration drop-ins are CRI related configurations k0s will automatically collect all these into a single file and adds that as a single import file. This is to overcome some hard limitation on containerd 1.X versions. Read more at containerd#8056
+
+However, this would also require a magic string set in the beginning of the line. Whatever the cases, the autogenerated Kata containerd configuration would always be there at `/etc/k0s/containerd.d/kata-containers.toml`.
+
+If you cannot meet the version requirement or disabled the automatic config management, you could also choose to manually append the config `kata-containers.toml` into your own config.toml
```
fidencio, line 72: Not really, the daemonset will actually replace any manual configuration the user does.  So, I'd drop this part.

tools/packaging/kata-deploy/README.md
```diff
@@ -43,6 +43,34 @@ $ kubectl apply -f kata-rbac/base/kata-rbac.yaml
...
+
+##### Note
+
+The supported version of k0s is **v1.27.1+k0s** and above, since the k0s support leverages a special dynamic containerd configuration mode:
+
+> From 1.27.1 onwards k0s enables dynamic configuration on containerd CRI runtimes. This works by k0s creating a special directory in /etc/k0s/containerd.d/ where user can drop-in partial containerd configuration snippets.
+> 
+> k0s will automatically pick up these files and adds these in containerd configuration imports list. If k0s sees the configuration drop-ins are CRI related configurations k0s will automatically collect all these into a single file and adds that as a single import file. This is to overcome some hard limitation on containerd 1.X versions. Read more at containerd#8056
+
+However, this would also require a magic string set in the beginning of the line. Whatever the cases, the autogenerated Kata containerd configuration would always be there at `/etc/k0s/containerd.d/kata-containers.toml`.
```
fidencio, line 70: I'm not sure if I understand this comment here.  What's the "magic string" that must be set in the beginning of the line?

tools/packaging/kata-deploy/README.md
```diff
@@ -43,6 +43,34 @@ $ kubectl apply -f kata-rbac/base/kata-rbac.yaml
...
 
+#### [k0s] cluster
+
+For your [k0s](https://k0sproject.io/) cluster, run:
+
+```sh
+$ git clone https://github.com/kata-containers/kata-containers.git
+```
+
+Check and switch to the stable branch of your choice, if wanted, and then run:
```
fidencio, line 54: I'd make a note that, **for now**, this will only work on the `main` branch, and that "stable" is not supported.

fidencio (commented):
I've left a few more comments.

tools/packaging/kata-deploy/README.md
```diff
@@ -43,6 +43,40 @@ $ kubectl apply -f kata-rbac/base/kata-rbac.yaml
...
+The supported version of k0s is **v1.27.1+k0s** and above, since the k0s support leverages a special dynamic containerd configuration mode:
+
+> From 1.27.1 onwards k0s enables dynamic configuration on containerd CRI runtimes. This works by k0s creating a special directory in /etc/k0s/containerd.d/ where user can drop-in partial containerd configuration snippets.
+> 
+> k0s will automatically pick up these files and adds these in containerd configuration imports list. If k0s sees the configuration drop-ins are CRI related configurations k0s will automatically collect all these into a single file and adds that as a single import file. This is to overcome some hard limitation on containerd 1.X versions. Read more at containerd#8056
+
+However, this would also require a magic string set in the beginning of the line for `/etc/k0s/containerd.toml`:
+
+```
+# k0s_managed=true
```
fidencio, line 73: Aha, perfect!

tools/packaging/kata-deploy/README.md
fidencio: Yeah. I'd still drop this paragraph and the one above.  If the conditions are not met, they are not met, and you don't rely on  kata-deploy.  That's better than telling folks to do manual steps, which may end up overriden by the daemonset.

fidencio:
I'll do the tests here from my side, but I'll do this on Monday (already too late on my timezone)
After the result of the tests, please, squash the commits as we discussed, and then we can have it in on Monday, sounds fair?

stevefan1999-personal:
> I'll do the tests here from my side, but I'll do this on Monday (already too late on my timezone)
> After the result of the tests, please, squash the commits as we discussed, and then we can have it in on Monday, sounds fair?

Sure. I'm already using it on my k0s main cluster

fidencio:
@stevefan1999-personal, I've just tested this one locally and it works like a charm.

Please, squash the commits as we discussed and let's have it merged!

stevefan1999-personal:
> @stevefan1999-personal, please, we need the following.
> 
> 1) Squash all the patches you have after a rebase on `main`
> 2) Use a descriptive commit message, pointing to the issue it's fixing, something like:
> ```
> kata-deploy: Add k0s support
> 
> Add k0s support to kata-deploy, in the very same way kata-containers
> already supports k3s, and rke2.
> 
> k0s support requires v1.27.1, which is noted as part of the kata-deploy
> documentation, as it's the way to use dynamic configuration on
> containerd CRI runtimes.
> 
> This support will only be part of the `main` branch, as it's not a bug
> fix that can be backported to the `stable-3.2` branch, and this is also
> noted as part of the documentation.
> 
> Fixes: #7548
> ```
> 3) The patch must be signoed-off-by yourself.

Thanks!

fidencio:
@stevefan1999-personal, sorry to bother you again, but can you rebase atop of the current `main`?
This will ensure that the fixes for the static-checks are already merged, and will be taken into consideration by our CI.

fidencio (approved):
lgtm, thanks @stevefan1999-personal!

fidencio:
It's weird that all the tests on k8s are failing, let me hold this one a little bit and do some debug locally.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -274,12 +286,15 @@ EOF
...
+		runtime+="-$2"
+		configuration+="-$2"
 	fi
 	local pluginid=cri
-	if grep -q "version = 2\>" $containerd_conf_file; then
+	
+	# if we are running k0s auto containerd.toml generation, the base template is by default version 2
+	# we can safely assume to reference the older version of cri
+	if [ grep -q "version = 2\>" $containerd_conf_file ] || [ "$1" == "k0s-worker" ] || [ "$1" == "k0s-controller" ]; then
 		pluginid=\"io.containerd.grpc.v1.cri\"
```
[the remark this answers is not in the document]
fidencio: I have force pushed a fix here, things will work as expected now.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -84,6 +85,12 @@ function get_container_runtime() {
 		else
 			echo "k3s"
 		fi
+	# Note: we assumed you used a conventional k0s setup and k0s will generate a systemd entry k0scontroller.service and k0sworker.service respectively    
+	# and it is impossible to run this script without a kubelet, so this k0s controller must also have worker mode enabled
```
cmaf, line 89: Possibly consider mentioning this in the README?

cmaf (approved):
thanks @stevefan1999-personal! Minor suggestion, otherwise approving

PR #7559, opened 2023-08-07
CC | agent: support to manage verity devices

ChengyuZhu6:
To share image with integrity on the host with CoCo as described in https://github.com/confidential-containers/confidential-containers/issues/137,  we need to manage verity devices in the kata-agent. We can reuse the `Storages` in the Sandbox to manage both mountpoints and verity devices.

Fixes: #7561

### DIFF
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -154,7 +154,7 @@ impl RefCountedObject<ObjectKind> {
     pub fn dec_ref(&mut self) -> Result<()> {
         self.ref_count -= 1;
         let name =  &self.name;
-        if self.is_zero() {
+        if self.get_count() == 0 {
             match self.ref_type {
                 ObjectKind::MountPoint => {
                     let mounts = vec![name.to_owned()];
@@ -170,10 +170,6 @@ impl RefCountedObject<ObjectKind> {
             }
         }
         Ok(())
-    }
-    #[inline]
-    pub fn is_zero(&self) -> bool {
-        self.ref_count == 0
     }
     #[inline]
     pub fn get_count(&self) -> u32 {
@@ -923,7 +919,7 @@ pub async fn add_storages(
 
         {
             let mut sb = sandbox.lock().await;
-            let new_storage = sb.set_sandbox_storage(&storage.mount_point);
+            let new_storage = sb.set_sandbox_storage(&storage.mount_point, ObjectKind::MountPoint);
             if !new_storage {
                 continue;
             }
--- a/src/agent/src/mount.rs
+++ b/src/agent/src/mount.rs
@@ -133,6 +133,54 @@ lazy_static! {
     ];
 }
 
+#[derive(Debug)]
+pub struct RefCountedObject<T> {
+    pub name: String,
+    pub ref_count: u32,
+    pub ref_type: T,
+}
+#[derive(Debug, Copy, Clone)]
+// Define an enum to represent the MountPoint and VerityDevice variants
+pub enum ObjectKind {
+    MountPoint,
+    VerityDevice,
+}
+// Implement the RefCounted trait for the generic struct
+impl RefCountedObject<ObjectKind> {
+    #[inline]
+    pub fn inc_ref(&mut self) {
+        self.ref_count += 1;
+    }
+    pub fn dec_ref(&mut self) -> Result<()> {
+        self.ref_count -= 1;
+        let name =  &self.name;
+        if self.is_zero() {
+            match self.ref_type {
+                ObjectKind::MountPoint => {
+                    let mounts = vec![name.to_owned()];
+                    remove_mounts(&mounts)?;
+                    // "remove_dir" will fail if the mount point is backed by a read-only filesystem.
+                    // This is the case with the device mapper snapshotter, where we mount the block device directly
+                    // at the underlying sandbox path which was provided from the base RO kataShared path from the host.
+                    fs::remove_dir(name.to_owned())?;
+                }
+                ObjectKind::VerityDevice => {
+                    image_rs::verity::destroy_verity_device(self.name.clone())?;
+                }
+            }
+        }
+        Ok(())
+    }
+    #[inline]
+    pub fn is_zero(&self) -> bool {
+        self.ref_count == 0
+    }
+    #[inline]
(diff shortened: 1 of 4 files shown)


src/agent/src/mount.rs
```diff
@@ -51,6 +52,75 @@ const RO_MASK: u32 = 0o440;
...
+}
+
+// Implement a custom dec_ref method for each variant
+impl RefCountedObject<ObjectKind> {
+    fn dec_ref(&self) -> Result<()> {
+        self.ref_count.fetch_sub(1, Ordering::Relaxed);
+        match self.ref_type {
+            ObjectKind::MountPoint => {
+                let mounts = vec![self.name.clone()];
+                remove_mounts(&mounts)?;
```
wedsonaf, line 108: Do you really want to remove mounts every time the refcount is decremented? Shouldn't you check that it's zero before removing?

src/agent/src/mount.rs
```diff
@@ -51,6 +52,75 @@ const RO_MASK: u32 = 0o440;
...
+// Define an enum to represent the MountPoint and VerityDevice variants
+pub enum ObjectKind {
+    MountPoint,
+    VerityDevice,
+}
+
+// Implement a custom dec_ref method for each variant
+impl RefCountedObject<ObjectKind> {
+    fn dec_ref(&self) -> Result<()> {
+        self.ref_count.fetch_sub(1, Ordering::Relaxed);
```
wedsonaf, line 104: If this is truly used concurrently, this should be a release operation.

And we should have an acquire fence after checking that the refcount is zero so that we can ensure that all usage of the object happens-before it is torn down.

src/agent/src/mount.rs
```diff
@@ -51,6 +52,75 @@ const RO_MASK: u32 = 0o440;
...
+                let mounts = vec![self.name.clone()];
+                remove_mounts(&mounts)?;
+                // "remove_dir" will fail if the mount point is backed by a read-only filesystem.
+                // This is the case with the device mapper snapshotter, where we mount the block device directly
+                // at the underlying sandbox path which was provided from the base RO kataShared path from the host.
+                fs::remove_dir(self.name.clone())?;
+            }
+            ObjectKind::VerityDevice => {
+                if self.ref_count.load(Ordering::Relaxed) == 0 {
+                    image_rs::verity::destroy_verity_device(self.name.clone())?;
```
wedsonaf, line 116: If you're going to perform destructive actions when the refcount reaches zero, and still keep the object alive even if the count is zero, you _must_ protect against the case when the refcount goes from zero back up to 1.

But I don't see such code in the `RefCounted` trait, which hints to me that this is either unsound and that at isn't truly used concurrently, in which case it doesn't really have to atomic.

src/agent/src/sandbox.rs
```diff
@@ -135,9 +143,9 @@ impl Sandbox {
     pub fn unset_sandbox_storage(&mut self, path: &str) -> Result<bool> {
         match self.storages.get_mut(path) {
             None => Err(anyhow!("Sandbox storage with path {} not found", path)),
-            Some(count) => {
-                *count -= 1;
-                if *count < 1 {
+            Some(obj) => {
+                obj.dec_ref()?;
+                if obj.is_zero() {
```
wedsonaf, line 148: `inc_ref` and `dec_ref` are only used here and in the block above. Both cases involve mutable references, so it is guaranteed that there is no aliasing.

Why do you need to make it atomic?

wedsonaf (commented):
I think there are two sets of changes here: converting to an atomic refcount and covering two different kinds of "objects". It would be useful to split them into separate commits.

One higher-level question: do you envision having dm-verity devices that aren't mounted?

I ask because when I implemented a similar functionality we never had this need. So it wasn't really necessary to manage them separately: every time the refcount of a mount point drops to zero, we remove the mount point, and if it's a dm-verity device, we remove the dm-verity device as well.

src/agent/src/mount.rs
ChengyuZhu6: You're right. I will change the codes

ChengyuZhu6:
> One higher-level question: do you envision having dm-verity devices that aren't mounted?

I thought about it. So I give` ObjectKind`. If it is not `VerityDevice`, you can add another type such as `Device` and implement the corresponding handler in `dec_ref`.

ChengyuZhu6:
> I ask because when I implemented a similar functionality we never had this need. So it wasn't really necessary to manage them separately: every time the refcount of a mount point drops to zero, we remove the mount point, and if it's a dm-verity device, we remove the dm-verity device as well.

But if multiple containers use the same image or layer, this can reduce the number of generated verity devices and reuse existing verity devices, thereby reducing the startup time of the container.

src/agent/src/sandbox.rs
ChengyuZhu6: I consider the situation that muti containers use the same verity device at the same time.

src/agent/src/mount.rs
ChengyuZhu6: Each caller is calling this method after acquiring a lock on sandbox. So I do not need to protect the case.

src/agent/src/mount.rs
```diff
@@ -50,7 +50,76 @@ const RW_MASK: u32 = 0o660;
 const RO_MASK: u32 = 0o440;
 const EXEC_MASK: u32 = 0o110;
 const MODE_SETGID: u32 = 0o2000;
+pub trait RefCounted {
```
wedsonaf, line 53: Do you intend to introduce other types that implement this trait?

If not, why are you defining it instead of just implementing these functions directly for the `RefCountedObject` type?

wedsonaf (commented):
Please ping me when you're done making changes and I'll do another pass.

src/agent/src/mount.rs
ChengyuZhu6: changed

ChengyuZhu6:
The PR https://github.com/kata-containers/kata-containers/pull/7582 has already contains this PR, so I close this PR.

ChengyuZhu6:
The PR https://github.com/kata-containers/kata-containers/pull/7582 has already contains this PR, so I close this PR.

Issue #7573, opened 2023-08-07
new Agent Policy feature


danmihai1:
**Is your feature request related to a problem? Please describe.**

See:

https://docs.google.com/presentation/d/1GraNlqLrP9z8ZMEF8WDdBHv0vXmQbE_d/edit?usp=sharing&ouid=117914183884789981574&rtpof=true&sd=true

https://github.com/confidential-containers/confidential-containers/issues/126

**Describe the solution you'd like**

See:

https://docs.google.com/presentation/d/1GraNlqLrP9z8ZMEF8WDdBHv0vXmQbE_d/edit?usp=sharing&ouid=117914183884789981574&rtpof=true&sd=true

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

Yes

PR #7574, opened 2023-08-07
agent: runtime: add Agent Policy feature

danmihai1:
Fixes: #7573

To enable this feature, build your rootfs using AGENT_POLICY=yes. The default is AGENT_POLICY=no.

Building rootfs using AGENT_POLICY=yes has the following effects:

1. The kata-opa service gets included in the Guest image.

2. The agent gets built using AGENT_POLICY=yes.

After this patch, the shim calls SetPolicy if and only if a Policy annotation is attached to the sandbox/pod. When creating a sandbox/pod that doesn't have an attached Policy annotation:

1. If the agent was built using AGENT_POLICY=yes, the new sandbox uses the default agent settings, that might include a default Policy too.

2. If the agent was built using AGENT_POLICY=no, the new sandbox is executed the same way as before this patch.

Any SetPolicy calls from the shim to the agent fail if the agent was built using AGENT_POLICY=no.

If the agent was built using AGENT_POLICY=yes:

1. The agent reads the contents of a default policy file during sandbox start-up.

2. The agent then connects to the OPA service on localhost and sends the default policy to OPA.

3. If the shim calls SetPolicy:

   a. The agent checks if SetPolicy is allowed by the current policy (the current policy is typically the default policy mentioned above).

    b. If SetPolicy is allowed, the agent deletes the current policy
       from OPA and replaces it with the new policy it received from
       the shim.

   A typical new policy from the shim doesn't allow any future SetPolicy calls.

4. For every agent rpc API call, the agent asks OPA if that call should be allowed. OPA allows or not a call based on the current policy, the name of the agent API, and the API call's inputs. The agent rejects any calls that are rejected by OPA.

When agent's debug level is debug or trace, additional Policy logging gets enabled in the agent. In particular, information about the inputs for agent rpc API calls is logged in /tmp/policy.txt, on the Guest VM. These inputs can be useful for investigating API calls that might have been rejected by the Policy. Examples:

1. Load a failing policy file test1.rego on a different machine:

opa run --server --addr 127.0.0.1:8181 test1.rego

2. Collect the API inputs from Guest's /tmp/policy.txt and test on the machine where the failing policy has been loaded:

curl -X POST http://localhost:8181/v1/data/agent_policy/CreateContainerRequest \ --data-binary @test1-inputs.json


src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        // OPA are expected to work without retrying, after connecting
+        // successfully for the first time.
+        for i in 0..50 {
+            if i > 0 {
+                sleep(Duration::from_millis(100)).await;
+                println!("policy initialize: PUT failed, retrying");
+            }
+
+            // Set-up the default policy.
+            if let Ok(_) = self
```
Xynnn007, line 97: it can be
```
if self
    .opa_client
    .put(&self.policy_path)
    .body(policy)
    .send()
    .await
    .is_ok() {
....
}
```

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+                .put(&self.policy_path)
+                .body(policy.clone())
+                .send()
+                .await
+            {
+                // Check if requests causing policy errors should actually
+                // be allowed. That is an insecure configuration but is
+                // useful for allowing insecure pods to start, then connect to
+                // them and inspect Guest logs for the root cause of a failure.
+                if let Ok(allow_failures) = self
```
Xynnn007, line 108: What about `pose_query` call returns an error?

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+            self.log_file = Some(tokio::fs::OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .create(true)
+                .open("/tmp/policy.txt")
+                .await
+                .unwrap());
+        }
+
+        let policy = tokio::fs::read_to_string("/etc/kata-opa/default-policy.rego")
```
Xynnn007, line 82: Currently we are using a lot of configurations/tempfiles/... in CoCo. It would be better that we put all these files under a same parent dir for better maintance, like `/run/confidential-containers/policy.txt` from https://github.com/kata-containers/kata-containers/pull/7570/files#diff-803c177f300cad05867e555f0b0080ee238df2ef218984786620f35d2198af33R10

However, if the policy file is assumed to be placed inside the guest image, `/etc` is a good choice.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+                }
+                return Ok(());
+            }
+        }
+        Err(anyhow!("failed to connect to OPA"))
+    }
+
+    /// Ask OPA to check if an API call should be allowed or not.
+    pub async fn is_allowed_endpoint(&mut self, ep: &str, request: &str) -> bool {
+        let post_input = "{\"input\":".to_string() + request + "}";
```
Xynnn007, line 122: it is better to use `format!` macro, s.t.
```rust
let post_input = format!("{{\"input\":\"{request}\"}}");
```

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /// Replace the Policy in OPA.
+    pub async fn set_policy(&mut self, policy: &str) -> Result<()> {
+        // Delete the old rules.
+        self.opa_client
+            .delete(&self.policy_path)
+            .send()
+            .await
+            .map_err(|e| anyhow!(e))?;
```
Xynnn007, line 137: if we are using `anyhow` crate, we can directly use `?` after `await`

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        // Put the new rules.
+        self.opa_client
+            .put(&self.policy_path)
+            .body(policy.to_string())
+            .send()
+            .await
+            .map_err(|e| anyhow!(e))?;
```
Xynnn007, line 145: same as mentioned.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+            .post_query("AllowRequestsFailingPolicy", EMPTY_JSON_INPUT)
+            .await?;
+        Ok(())
+    }
+
+    // Post query to OPA.
+    async fn post_query(&mut self, ep: &str, post_input: &str) -> Result<bool> {
+        debug!(sl!(), "policy check: {}", ep);
+
+        let uri = self.query_path.clone() + ep;
```
Xynnn007, line 161:
```suggestion
        let uri = format!("{}{ep}", self.query_path);
```

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        let response = self
+            .opa_client
+            .post(uri)
+            .body(post_input.to_string())
+            .send()
+            .await
+            .map_err(|e| anyhow!(e))?;
```
Xynnn007, line 168: if we are using anyhow crate, we can directly use ? after await

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        if response.status() != http::StatusCode::OK {
+            return Err(anyhow!(
+                "policy: post_query: POST response status {}",
+                response.status()
+            ));
```
Xynnn007, line 174:
```suggestion
            bail!(
                "policy: post_query: POST response status {}",
                response.status()
            );
```

Xynnn007: `bail` macro is short for `return Err(anyhow!(..))`

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        let http_response = response.text().await.unwrap();
```
Xynnn007, line 177: should we raise this error to upper caller?

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    #[cfg(feature = "agent-policy-debug")]
+    async fn log_opa_input(&mut self, ep: &str, opa_input: &str) {
+        if let Some(log_file) = &mut self.log_file {
+            match ep {
+                "StatsContainerRequest" | "ReadStreamRequest" | "SetPolicyRequest" => {}
+                _ => {
+                    let log_entry = "# ".to_string() + ep + "\n\n" + opa_input + "\n\n";
+                    log_file.write_all(log_entry.as_bytes()).await.unwrap();
+                    log_file.flush().await.unwrap();
```
Xynnn007, line 213: Please try to avoid using `unwrap()`. Change the function definition to `...-> Result<()>` is a choice. In this way all the errors occuring inside can be raised to upper caller.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+
+    #[cfg(feature = "agent-policy-debug")]
+    async fn log_opa_input(&mut self, ep: &str, opa_input: &str) {
+        if let Some(log_file) = &mut self.log_file {
+            match ep {
+                "StatsContainerRequest" | "ReadStreamRequest" | "SetPolicyRequest" => {}
+                _ => {
+                    let log_entry = "# ".to_string() + ep + "\n\n" + opa_input + "\n\n";
```
Xynnn007, line 211:
```suggestion
                    let log_entry = format!("# {ep}\n\n{opa_input}\n\n");
```

src/libs/protocols/protos/agent.proto
```diff
@@ -566,3 +567,7 @@ message ResizeVolumeRequest {
 	string volume_guest_path = 1;
 	uint64 size = 2;
 }
+
+message SetPolicyRequest {
```
Xynnn007, line 571: I'm not familiar with how kata-runtime runs, but does this mean that the policy is delivered from the host side. Is it covered by the tee evidence?

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -640,6 +649,55 @@ EOF
 		chmod g+rx,o+x "${ROOTFS_DIR}"
 	fi
 
+	if [ "${AGENT_POLICY}" == "yes" ]; then
```
Xynnn007, line 652: Seems that there are two ways to inject the policy.
1. Built-in guest image
2. from kata-runtime via annotations

right?

Xynnn007:
Another potential way to get the policy rathan embedding  into the guest image mentioned by  @fitzthum is from KBS, btw.

src/libs/protocols/protos/agent.proto
bergwolf: +1, from the referenced slides, I can clearly see the value of pre-populated policy and KBS-provided policy. Could you explain a bit about the use case and threat model of the host-provided policy? That is the one requiring a new agent proto API.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+    };
+}
+
+/// Example of HTTP response from OPA: {"result":true}
+#[derive(Debug, Serialize, Deserialize)]
+struct AllowResponse {
+    result: bool,
+}
+
+/// Singleton policy object.
```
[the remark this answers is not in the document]
danmihai1: No, I meant that there is just one instance of this object (https://en.wikipedia.org/wiki/Singleton_pattern)

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl AgentPolicy {
+    /// Create AgentPolicy object.
+    pub fn new(opa_uri: &str, coco_policy: &str) -> Result<Self> {
+        Ok(AgentPolicy {
+            allow_failures: false,
+            query_path: opa_uri.to_string() + OPA_DATA_PATH + coco_policy + "/",
+            policy_path: opa_uri.to_string() + OPA_POLICIES_PATH + coco_policy,
+            opa_client: Client::builder().http1_only().build().unwrap(),
```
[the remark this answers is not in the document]
danmihai1: My experience from debugging thousands of crashes at Microsoft shows that it's better to crash early than to propagate a fatal error to the caller and hope they will handle that error correctly. For this kind of error:
- The user wanted to support Policy, since they added Policy support by building using AGENT_POLICY=yes
- The Policy will not work at all, so this is a fatal error for the Agent.

So, I think unwrap is correct here. But, I know that most developers don't have the same debugging experience I have, so don't mind sending these errors to the caller if that's what Kata maintainers prefer.

src/agent/src/policy.rs
danmihai1: In the current design, an agent built using AGENT_POLICY=yes always starts with a default Policy that is baked in the Guest image. A typical default Policy is: https://github.com/microsoft/kata-containers/blob/cc-msft-prototypes/src/agent/samples/policy/set-policy-allowed/set-policy-allowed.rego

So, I think /etc is reasonable here. But, I wouldn't mind moving this file if other folks prefer it in a different location.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        let policy = tokio::fs::read_to_string("/etc/kata-opa/default-policy.rego")
+            .await
+            .unwrap();
+
+        // This loop is necessary to get the opa_client connected to the
+        // OPA service while that service is starting. Future requests to
+        // OPA are expected to work without retrying, after connecting
+        // successfully for the first time.
```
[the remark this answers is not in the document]
danmihai1: Yes:

1. systemd starts OPA, and later
2. systemd starts the Agent.

But, systemd doesn't wait for OPA to finish its initialization in between these two steps. In fact, I don't think systemd has a way to know when OPA finished its initialization. So, the Agent still has to wait until the OPA REST APIs become available.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        for i in 0..50 {
+            if i > 0 {
+                sleep(Duration::from_millis(100)).await;
+                println!("policy initialize: PUT failed, retrying");
```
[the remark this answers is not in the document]
danmihai1: This code runs very early during Agent initialization. As far as I know, Kata doesn't have a great way to collect the Rust log so early - because the Agent and the Shim are not connected yet here. Therefore, I used vanilla console output.

But it's possible that we can improve this. I will take another look.

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            // Set-up the default policy.
+            if let Ok(_) = self
+                .opa_client
+                .put(&self.policy_path)
+                .body(policy.clone())
```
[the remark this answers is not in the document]
danmihai1: I believe we need this clone due to the next iteration of this _for_ loop. Please let me know in case I missed something here.

src/agent/src/policy.rs
danmihai1: If post_query returns an error, the secure default from AgentPolicy::new gets used (allow_failures: false). I will add a comment here to explain that.

src/agent/src/policy.rs
danmihai1: Indeed, thanks for catching this!

src/agent/src/policy.rs
danmihai1: Thanks - I will change this.

src/agent/src/policy.rs
danmihai1: Sounds good!

src/agent/src/policy.rs
danmihai1: I will make this change.

src/agent/src/policy.rs
danmihai1: I think unwrap is correct - but I don't mind sending this error to the caller if other folks prefer it that way.

src/agent/src/policy.rs
danmihai1: Sounds good, thanks!

src/libs/protocols/protos/agent.proto
danmihai1: @Xynnn007 Sorry, we ran out of time in last week’s Kata meeting before I got a chance to talk about this. However:

1. I presented that in a CoCo meeting a while ago – at offset 44:00 from https://zoom.us/rec/share/MMDPlX90B9ouSn3weOHMqtCBh0RkJvutDn-KFGKUi4pCoUoyYKVnAVZk_FEYW1KL.SaW8DXu9gFI3JZDu

2. Similar information is present in slide 20 of my slides for the Kata group (https://docs.google.com/presentation/d/1GraNlqLrP9z8ZMEF8WDdBHv0vXmQbE_d/edit#slide=id.p20)

3. A longer discussion about this is in https://github.com/confidential-containers/confidential-containers/issues/126#issuecomment-1644402591

Please let me know if you have additional questions about this.


@bergwolf , the main advantages of the Host-provided Policy are:

a. Common, generic Guest image for multiple pods and multiple K8s customers - because we don't have to change the Guest image to provide the Policy for a Particular pod.

b. Immutable Policy - because the TEE doesn't allow anyone (even the trusted components inside the TEE!) to change the expected policy hash that gets attached to the TEE. For example, when using a SEV-SNP TEE, nobody can change the value of the HOST_DATA field after the TEE started its execution.

c. Ability to verify the expected Policy hash using a generic remote attestation service - as long as that service supports verifying the value of HOST_DATA or similar TEE field.

danmihai1:
> Another potential way to get the policy rathan embedding into the guest image mentioned by @fitzthum is from KBS, btw.

I agree. In that case a typical default Policy - included in the Guest image will be just:

package agent_policy

That default policy will not allow any agent rpc calls until the KBS provides a different Policy.

src/agent/src/policy.rs
danmihai1: Good point, thanks!

src/agent/src/policy.rs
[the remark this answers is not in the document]
danmihai1: Looks like debug! works well here - so I changed my println! to debug!

src/agent/src/policy.rs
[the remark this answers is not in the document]
fitzthum: In this case the caller is also part of this PR. It unwraps the output of this function. I don't see any reason to unwrap here when this function returns a result. 

More generally whether an error is fatal and whether an error is passed to the caller are a bit orthogonal. Passing the error up allows us to handle it at the appropriate level and to make the behavior of the callee more closely reflect its signature.

fitzthum (commented):
A few notes. Will review more tomorrow. 

To me having `unwrap()` all over the place is not a good look.

fitzthum:
Since this PR doesn't seem to include anything that allows us to validate a policy that is added via `SetPolicy`, I think that endpoint must be disabled in the default policy or moved to a future PR entirely. Otherwise anyone who uses the policy feature will allow the host to change the policy at will. Without SetPolicy, the policy will have to be added to the initrd, which is measured. This is a bit clunky but it seems like a good first step given that it is very similar to the agent-config design that we already have.

Speaking of the agent configuration file, I think the default policy should block Exec and ReseedRng like we do in the current config. These seem like reasonable defaults and can help us port the test of API restrictions. Are there any other tests we should add?

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -27,10 +27,14 @@ LIBC=${LIBC:-musl}
 # However, it is not enforced by default: you need to enable that in the main configuration file.
 SECCOMP=${SECCOMP:-"yes"}
 SELINUX=${SELINUX:-"no"}
+AGENT_POLICY=${AGENT_POLICY:-no}
+AGENT_POLICY_DEBUG=${AGENT_POLICY_DEBUG:-no}
```
fitzthum, line 31: Are we setting these variables in the GHA job?

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -640,6 +649,55 @@ EOF
...
+		fi
+
+		# Install default settings for the kata-opa service.
+		local kata_opa_in_dir="${script_dir}/../../../src/kata-opa"
+		local opa_settings_dir="/etc/kata-opa"
+		local policy_file="allow-all.rego"
+		local policy_dir="${ROOTFS_DIR}/${opa_settings_dir}"
+		mkdir -p "${policy_dir}"
+		install -D -o root -g root -m 0644 "${kata_opa_in_dir}/${policy_file}" -T "${policy_dir}/${policy_file}"
+		ln -sf "${policy_file}" "${policy_dir}/default-policy.rego"
```
fitzthum, line 683: It seems likely that different runtime classes will want different policies, at least for people who are using host data. What is best way to support that?

src/agent/src/rpc.rs
```diff
@@ -120,7 +124,7 @@ fn ttrpc_error(code: ttrpc::Code, err: impl std::fmt::Debug) -> ttrpc::Error {
     get_rpc_status(code, format!("{:?}", err))
 }
 
-fn is_allowed(req: &impl MessageDyn) -> ttrpc::Result<()> {
+fn config_allows(req: &impl MessageDyn) -> ttrpc::Result<()> {
```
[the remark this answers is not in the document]
danmihai1: I vote for replacing the agent config. But, as far as I know, there hasn't been a community decision to remove it yet.

tools/osbuilder/rootfs-builder/rootfs.sh
danmihai1: What is the GHA job?

If you are referring to adding test coverage for policy for non-default configurations: I plan to do that next.

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -315,6 +319,9 @@ check_env_variables()
 
 	[ "$AGENT_INIT" == "yes" -o "$AGENT_INIT" == "no" ] || die "AGENT_INIT($AGENT_INIT) is invalid (must be yes or no)"
 
+	[ "$AGENT_POLICY" == "yes" -o "$AGENT_POLICY" == "no" ] || die "AGENT_POLICY($AGENT_POLICY) is invalid (must be yes or no)"
+	[ "$AGENT_POLICY" == "no" -o "$AGENT_INIT" == "no" ] || die "AGENT_POLICY($AGENT_POLICY) and AGENT_INIT($AGENT_INIT) is an invalid combination (at least one must be no)"
```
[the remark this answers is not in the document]
danmihai1: As you said, -o is not a new problem here.

Without systemd, we just need to figure out how to execute OPA in an appropriate security context - that's the gap.

tools/osbuilder/rootfs-builder/rootfs.sh
danmihai1: I agree. Solution: TBD, in a future patch.

src/agent/src/policy.rs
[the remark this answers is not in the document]
danmihai1: I replaced the unwrap calls, but I added an abort call - please see https://github.com/kata-containers/kata-containers/pull/7574/commits/5f1bbe4a79d3d6d07ec25e0253c23b3f6dc758b0

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub async fn initialize(&mut self) -> Result<()> {
+        #[cfg(feature = "agent-policy-debug")]
+        {
+            self.log_file = Some(tokio::fs::OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .create(true)
+                .open("/tmp/policy.txt")
+                .await
+                .unwrap());
```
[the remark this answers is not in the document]
danmihai1: I replaced the unwrap calls, but I added an abort call - please see https://github.com/kata-containers/kata-containers/pull/7574/commits/5f1bbe4a79d3d6d07ec25e0253c23b3f6dc758b0

src/agent/src/policy.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        let policy = tokio::fs::read_to_string("/etc/kata-opa/default-policy.rego")
+            .await
+            .unwrap();
```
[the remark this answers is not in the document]
danmihai1: I replaced the unwrap calls, but I added an abort call - please see https://github.com/kata-containers/kata-containers/pull/7574/commits/5f1bbe4a79d3d6d07ec25e0253c23b3f6dc758b0

src/agent/src/policy.rs
danmihai1: I replaced the unwrap calls, but I added an abort call - please see https://github.com/kata-containers/kata-containers/pull/7574/commits/5f1bbe4a79d3d6d07ec25e0253c23b3f6dc758b0

src/agent/src/policy.rs
danmihai1: I replaced the unwrap calls, but I added an abort call - please see https://github.com/kata-containers/kata-containers/pull/7574/commits/5f1bbe4a79d3d6d07ec25e0253c23b3f6dc758b0

danmihai1:
> Nice PR. A few notes. Will review more tomorrow.
> 
> To me having `unwrap()` all over the place is not a good look.

I replaced the unwrap calls, but I added an abort call - please see https://github.com/kata-containers/kata-containers/pull/7574/commits/5f1bbe4a79d3d6d07ec25e0253c23b3f6dc758b0

danmihai1:
> Since this PR doesn't seem to include anything that allows us to validate a policy that is added via `SetPolicy`, I think that endpoint must be disabled in the default policy or moved to a future PR entirely. Otherwise anyone who uses the policy feature will allow the host to change the policy at will. Without SetPolicy, the policy will have to be added to the initrd, which is measured. This is a bit clunky but it seems like a good first step given that it is very similar to the agent-config design that we already have.
> 
> Speaking of the agent configuration file, I think the default policy should block Exec and ReseedRng like we do in the current config. These seem like reasonable defaults and can help us port the test of API restrictions. Are there any other tests we should add?

Yes, the additional tests are coming in a future PR.

My security goal for this PR is: "don't cause any regressions". Additional functionality will come later - including blocking some of the APIs by default, etc. Please let me know in case you see a regression (caused by the new SetPolicy, or something else).

Measuring the policy will come in a future PR.

fidencio:
@danmihai1,

"When building using AGENT_POLICY_DEBUG=yes ..."  ... this is something I'm not totally fond of, as it'd require a rebuild of the guest image to debug possible issues.  I wonder whether this could be done in a way that we'd:
1. DEBUG stuff would always be enabled
2. We log stuff or not based on the Kata Containers debug being enabled or not

With the suggested approach, if feasable, debugging issues would be a matter of switching a toggle in or off in the Kata Containers configuration file, which would save a reasonable amount of time.

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: We're not, I've discussed with @danmihai1 what's the way to approach with the tests, and we agreed to have the tests coming in a different PR.

This PR will ensure that no regressions are caused, the second PR will enable this for the guest images we care about, and add the tests for those.

fidencio:
> Are there any other tests we should add?

We should test basically every option we can block, but as mentioned by @danmihai1 and myself, this will come as a follow-up PR.

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: This is an orthogonal discussion, but kata-deploy can take care of specific policies for specific runtime classes.

We're already dealing (or about to deal) with specific node feature rules for specific runtime-classes, and the logic can just be followed based on what I'm doing for NFD enablement.

src/agent/src/rpc.rs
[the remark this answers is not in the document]
fidencio: I'd remove the agent config at the moment we have the tests in place for this one.

src/agent/src/main.rs
```diff
@@ -221,6 +229,24 @@ async fn real_main() -> std::result::Result<(), Box<dyn std::error::Error>> {
...
+            .initialize(
+                "http://localhost:8181/v1",
+                "/agent_policy",
+                "/etc/kata-opa/default-policy.rego",
+            )
+            .await
+        {
+            error!(logger, "Failed to initialize agent policy: {:?}", e);
+            // Continuing execution without a security policy could be dangerous.
+            std::process::abort();
```
Xynnn007, line 246: I think we can also use `?` to raise the error to upper level, s.t. `real_main()` https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/main.rs#L149.

The tokio runtime will spawn a task (like green thread in golang) https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/main.rs#L301. if any error is raised this sentence will return and finish executing `main`, finally exit the process.

src/agent/src/policy.rs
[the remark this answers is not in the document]
Xynnn007: I think we can also use `?` to raise the error to upper level, s.t. `real_main()` https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/main.rs#L149.

The tokio runtime will spawn a task (like green thread in golang) https://github.com/kata-containers/kata-containers/blob/CCv0/src/agent/src/main.rs#L301. if any error is raised this sentence will return and finish executing `main`, finally exit the process.

versions.yaml
```diff
@@ -284,6 +284,22 @@ externals:
...
+      # - If an OPA package is available for the Guest image distro, that package
+      #   is used instead of the binary below.
+      #
+      # - TODO: if an OPA package is not available for the Guest image distro,
+      #   Kata should cache the OPA source code, toolchain information, etc.
+      #   OPA should be built from the cached source code instead of downloading
+      #   this binary.
+      #
+      # yamllint disable-line rule:line-length
+      binary: "https://github.com/open-policy-agent/opa/releases/download/v0.55.0/opa_linux_amd64_static"
```
fidencio, line 301: So, it's nice that OPA has a binary for amd64 / arm64, but unfortunately it doesn't provide the same for other architectures that we're using (s390x, and maybe powerpc64).

Instead of adding the `binary` entry here, I'd recommend:
* Having a generic link `tarball_url` / `binary_url` which would point https://github.com/open-policy-agent/opa/releases/download/
  * This is similar to: https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L221C3-L221C78
* Have a script under https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/static-build, where we'd download the binary if we're building for x86_64 / arm64, build the binary if we're on s390x / powerpc64
  * Something similar is done here: https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/static-build/cloud-hypervisor/build-static-clh.sh
  * The build / download should be done inside a container
* Call the script, which always will give you the binary (or error out in case of build fails), and then we can use it
  * This will make things way friendlier for different arches.

tools/osbuilder/rootfs-builder/rootfs.sh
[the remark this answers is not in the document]
zvonkok: Agree on having support without systemd.

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -640,6 +649,55 @@ EOF
...
+		local opa_bin_dir="$(get_opa_bin_dir "${ROOTFS_DIR}")"
+		if [ -z "${opa_bin_dir}" ]; then
+			# OPA was not installed already, so download it here.
+			#
+			# TODO: if an OPA package is not available for the Guest image distro,
+			#   	Kata should cache the OPA source code, toolchain information, etc.
+			#   	OPA should be built from the cached source code instead of downloading
+			#   	this binary.
+			#
+			opa_bin_url="$(get_package_version_from_kata_yaml externals.open-policy-agent.meta.binary)"
```
fidencio, line 663: I left a few comments about this part in the versions.yaml changes.

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -456,6 +463,8 @@ build_rootfs_distro()
 			--env SELINUX="${SELINUX}" \
 			--env DEBUG="${DEBUG}" \
 			--env HOME="/root" \
+			--env AGENT_POLICY="${AGENT_POLICY}" \
+			--env AGENT_POLICY_DEBUG="${AGENT_POLICY_DEBUG}" \
```
fidencio, line 467: I've left a comment about the AGENT_POLICY_DEBUG as a separate comment, sorry.

fidencio (commented):
@danmihai1, overall this looks very good to me.

versions.yaml
fidencio: @BbolroC, @Amulyam24, can you help with the s390x / powerpc build side of the things?

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: BTW, I don't think we actually need to set anything in the GHA.
Adding an `export AGENT_POLICY=yes` to https://github.com/kata-containers/kata-containers/blob/7bf994827daf307b4cceea099cf8aff8a4d400e5/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L186-L189 was what we need to get it built by default for the TDX image.

This should **NOT** be done as part of this PR, though, and the message is a JFYI kind of comment,

versions.yaml
BbolroC: Yes, I can, but is it okay to go over this next Tuesday?

Amulyam24: Sure @fidencio, I can work on this next week. Will coordinate with @BbolroC if needed :)

tools/osbuilder/rootfs-builder/rootfs.sh
katexochen: > What is the GHA job?

GHA = GitHub Actions

fidencio:
BTW, none of my comments are blockers, apart from the debug one.
For the non-blockers comments, I'd like to see issues opened so we can track the effort.

IMHO, this is also valid for not using systemd.  This is not a blocker for this PR, as long as we have an issue opened so folks who're actively using that can jump in the boat and start to help as well.

src/agent/src/main.rs
danmihai1: I think we should shut down the process as soon as possible when the security policy is broken. So, I hope that calling abort here - from real_main - is acceptable.

fitzthum: This is not how other critical failures are handled and it seems unidiomatic to me, but maybe someone with more experience with the agent code can weight in.

src/agent/src/rpc.rs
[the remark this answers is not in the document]
fitzthum: Yeah I think it is acceptable to remove at agent config as long as doing so won't cause regressions. At the moment I guess it would given we require systemd.

fitzthum:
> My security goal for this PR is: "don't cause any regressions". Additional functionality will come later - including blocking some of the APIs by default, etc. Please let me know in case you see a regression (caused by the new SetPolicy, or something else).

Having `SetPolicy` exposed is a serious regression compared to the existing agent configuration file, which cannot be modified by the host. If we don't yet validate the new policy, it is inherently not secure and should not be enabled.

versions.yaml
BbolroC: FYI: I was able to generate a static binary `opa_linux_s390x_static` on a `amd64` machine with the following:

```
$ go version
go version go1.21.0 linux/amd64
$ git clone -b v0.55.0 https://github.com/open-policy-agent/opa.git
$ cd opa
$ export GOARCH=s390x
$ make clean generate
$ make ci-go-ci-build-linux-static
$ $ tree --du -h _release
_release
└── [ 39M]  0.55.0
    ├── [ 39M]  opa_linux_s390x_static
    └── [  89]  opa_linux_s390x_static.sha256
$ file _release/0.55.0/opa_linux_s390x_static
_release/0.55.0/opa_linux_s390x_static: ELF 64-bit MSB executable, IBM S/390, version 1 (SYSV), statically linked, Go BuildID=lgngaupMJCkzniJiPQae/zxxUQm2pjE0oRsFgizYe/dCaqFReGFIt0ZDCUd3xO/rLftgYdjD6B86PJjIXAR, with debug_info, not stripped
```

danmihai1:
> 

Great idea, thanks! I have just replaced AGENT_POLICY_DEBUG=yes with a run time check for agent's debug level, in https://github.com/kata-containers/kata-containers/pull/7574/commits/eb6f97685fb5afa965a9aba59875c30a0381faea

danmihai1:
> > My security goal for this PR is: "don't cause any regressions". Additional functionality will come later - including blocking some of the APIs by default, etc. Please let me know in case you see a regression (caused by the new SetPolicy, or something else).
> 
> Having `SetPolicy` exposed is a serious regression compared to the existing agent configuration file, which cannot be modified by the host. If we don't yet validate the new policy, it is inherently not secure and should not be enabled.

The agent configuration file is still supported and still rejects the same APIs it used to reject before these changes - so I don't see the regression. Please let me know in case I am missing something here.

tools/osbuilder/rootfs-builder/rootfs.sh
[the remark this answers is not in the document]
danmihai1: New issue to support Guest images that don't use systemd: https://github.com/kata-containers/kata-containers/issues/7615

fitzthum:
> The agent configuration file is still supported and still rejects the same APIs it used to reject before these changes - so I don't see the regression. Please let me know in case I am missing something here.

So the agent config file will take priority? It looks like the policy cannot enable something that the config blocks. Right now the config does not include `SetPolicy` in the allowed endpoints, so it would be blocked, which would be okay. Is that your understanding as well?

versions.yaml
danmihai1: Thanks everyone! New issue to tracked this work: https://github.com/kata-containers/kata-containers/issues/7615

tools/osbuilder/rootfs-builder/rootfs.sh
danmihai1: Thanks! New issue to track this work: https://github.com/kata-containers/kata-containers/issues/7616

danmihai1:
> > The agent configuration file is still supported and still rejects the same APIs it used to reject before these changes - so I don't see the regression. Please let me know in case I am missing something here.
> 
> So the agent config file will take priority? It looks like the policy cannot enable something that the config blocks. Right now the config does not include `SetPolicy` in the allowed endpoints, so it would be blocked, which would be okay. Is that your understanding as well?

The agent config takes precedence - see this code:

async fn is_allowed(req: &(impl MessageDyn + serde::Serialize)) -> ttrpc::Result<()> {
    **let res = config_allows(req);**

    #[cfg(feature = "agent-policy")]
    if res.is_ok() {
        **return policy_allows(req).await;**
    }

    res
}

fitzthum:
> The agent config takes precedence 

Ok. We'll just have to be careful in subsequent PRs.

stevenhorsman (commented):
I've had a look through this PR and there were already lots of comments already which I think covered more than what I was thinking, so I don't have much to add beyond then. I'm slightly concerned by the lack of testing and deferring some features to later, but I understand the argument that it's not enabled at the moment, so I can understand the pragmatic approach to incrementally get things in rather than blocking progress.

One comment maybe for the future it that it would be nice to see more separation in the commits by component and potentially features to help with the review and seeing the commit history.

danmihai1:
> I've had a look through this PR and there were already lots of comments already which I think covered more than what I was thinking, so I don't have much to add beyond then. I'm slightly concerned by the lack of testing and deferring some features to later, but I understand the argument that it's not enabled at the moment, so I can understand the pragmatic approach to incrementally get things in rather than blocking progress.
> 
> One comment maybe for the future it that it would be nice to see more separation in the commits by component and potentially features to help with the review and seeing the commit history.

@stevenhorsman , @fidencio 

I started with 3 commits:
1. OPA + rootfs
2. Runtime
3. Agent

These commits would have had to be merged in order, because for example (2) and (3) depended on the same agent protocol change. 

I eventually concluded that you guys would prefer these squashed together. Maybe my conclusions was wrong. I'd be happy to split the code that way again if you guys want - it's easy enough.

danmihai1:
> @danmihai1, thanks for the PR, lgtm!
> 
> For me, it can get merged as soon as we get a second approval. I'd squash the follow-up commits now that CI ran and we know they're good.
> 
> I also agree with @stevenhorsman's comment on splitting the next one per part touched, as that makes it easier for reviewing / debugging in the future. It's not a blocker, this one does not have to be re-written, but that's a good tip for the next ones.

I have just squashed the follow-up commits. Thanks!

fidencio:
> I'd be happy to split the code that way again if you guys want - it's easy enough.

It's fine as it's for this one, thanks for the offer though, @danmihai1!

fidencio:
@danmihai1, static checks are failing, and they can be fixed by:
```
⋊> kata-containers/src/runtime on danmihai1/policy ≡ ⨯ git diff HEAD
diff --git a/src/runtime/virtcontainers/pkg/mock/mock.go b/src/runtime/virtcontainers/pkg/mock/mock.go
index 562d4a7e3..2e4705126 100644
--- a/src/runtime/virtcontainers/pkg/mock/mock.go
+++ b/src/runtime/virtcontainers/pkg/mock/mock.go
@@ -256,3 +256,8 @@ func (p *HybridVSockTTRPCMockImp) GetIPTables(ctx context.Context, req *pb.GetIP
 func (p *HybridVSockTTRPCMockImp) SetIPTables(ctx context.Context, req *pb.SetIPTablesRequest) (*pb.SetIPTablesResponse, error) {
        return &pb.SetIPTablesResponse{}, nil
 }
+
+func (p *HybridVSockTTRPCMockImp) SetPolicy(ctx context.Context, req *pb.SetPolicyRequest) (*gpb.Empty, error) {
+       return &gpb.Empty{}, nil
+}
+
```

danmihai1:
> @danmihai1, static checks are failing, and they can be fixed by:
> 
> ```
> ⋊> kata-containers/src/runtime on danmihai1/policy ≡ ⨯ git diff HEAD
> diff --git a/src/runtime/virtcontainers/pkg/mock/mock.go b/src/runtime/virtcontainers/pkg/mock/mock.go
> index 562d4a7e3..2e4705126 100644
> --- a/src/runtime/virtcontainers/pkg/mock/mock.go
> +++ b/src/runtime/virtcontainers/pkg/mock/mock.go
> @@ -256,3 +256,8 @@ func (p *HybridVSockTTRPCMockImp) GetIPTables(ctx context.Context, req *pb.GetIP
>  func (p *HybridVSockTTRPCMockImp) SetIPTables(ctx context.Context, req *pb.SetIPTablesRequest) (*pb.SetIPTablesResponse, error) {
>         return &pb.SetIPTablesResponse{}, nil
>  }
> +
> +func (p *HybridVSockTTRPCMockImp) SetPolicy(ctx context.Context, req *pb.SetPolicyRequest) (*gpb.Empty, error) {
> +       return &gpb.Empty{}, nil
> +}
> +
> ```

Done, thanks!

jepio (approved):
looks good to me

Issue #7584, opened 2023-08-08
dragonball vsock support pass through pipe/fifo  fd


lifupan:
In the dragonball hybrid vsock we had support pass through a fd for as a vsock backend,
but this is not enough, since the users may pass a fifo/pipe fd instead of the unix stream socket
fd, thus in order to support all of the stream fd, we add a hybrid stream support to cover
the other type fd.

PR #7585, opened 2023-08-08
dragonball: vsock add fifo/pipe stream support for passed fd hybridSt…

lifupan:
dragonball: vsock add fifo/pipe stream support for passed fd hybridStream
    
    Since the passed fd through unix socket would be any
    stream fd such as pipe/fifo fd or any other socket
    fd, thus we should deal with it as a normal hybrid
    stream instead of a unix stream.
    
    Fixes:#7584
    
### DIFF
--- a/src/dragonball/src/dbs_virtio_devices/src/vsock/backend/hybrid_stream.rs
+++ b/src/dragonball/src/dbs_virtio_devices/src/vsock/backend/hybrid_stream.rs
@@ -1,0 +1,94 @@
+// Copyright 2023 Ant Group. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::Any;
+use std::io::{Error, Read, Write};
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::time::Duration;
+
+use log::error;
+use nix::errno::Errno;
+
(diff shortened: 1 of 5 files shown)


src/dragonball/src/dbs_virtio_devices/src/vsock/backend/hybrid_stream.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn backend_type(&self) -> VsockBackendType {
+        VsockBackendType::HybridStream
+    }
+
+    fn set_nonblocking(&mut self, nonblocking: bool) -> std::io::Result<()> {
+        let fd = self.hybrid_stream.as_raw_fd();
+        let mut flag = unsafe { libc::fcntl(fd, libc::F_GETFL) };
+
+        if nonblocking {
+            flag = flag & libc::O_NONBLOCK;
```
frezcirno, line 58: typo? should be `flag = flag | libc::O_NONBLOCK;`?

lifupan: Good catch, Fixed.

bergwolf (approved):

studychao (approved):
LGTM

Issue #7588, opened 2023-08-08
CC | runtime: Re-add remote hypervisor sandbox annotations


stevenhorsman:
Re-add (with improvement?) the logic to add sandbox annotation to the hypervisorconfig. This was accidentally reverted in a merge from main.

stevenhorsman:
Closed with #7589 merging

PR #7589, opened 2023-08-08
Re-added hypervisor annotations

stevenhorsman:
- Add support for setting the sandbox name and namespace in the hypervisor config, which is needed in the remote hypervisor implementation to get the pod name and namespace for the remote pod create request
- Also cherry-pick in the nydus version bump to attempt to stop the dragonball errors

Fixes: #7588

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -25,6 +25,8 @@ import (
 
 	v1 "github.com/containerd/cgroups/stats/v1"
 	v2 "github.com/containerd/cgroups/v2/stats"
+	cri "github.com/containerd/containerd/pkg/cri/annotations"
+	crio "github.com/containers/podman/v4/pkg/annotations"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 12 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman!

fidencio:
And sorry for mistakenly removing it in the first place.

stevenhorsman:
> And sorry for mistakenly removing it in the first place.

It happens to the best of us (and me!) Thanks to @genjuro214 for spotting the issue!

fidencio:
/test

And let's hope the non-triggered tests will be triggered this time

stevenhorsman:
The SEV tests is failing with `not ok 3 [cc][kubernetes][containerd][sev] Test SEV encrypted container launch success with NO measurement` which I think we know is flakey.
@fidencio, @arronwy - if you could check the TDX logs to see if they are real problems and/or re-queue that would be appreciated to try and get peer pods back working. Thanks!

src/runtime/virtcontainers/sandbox.go
```diff
@@ -638,6 +642,23 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
 	return s, nil
 }
 
+func setHypervisorConfigAnnotations(sandboxConfig *SandboxConfig) {
+	if len(sandboxConfig.Containers) > 0 {
+		// These values are required by remote hypervisor
+		for _, a := range []string{cri.SandboxName, crio.SandboxName} {
```
wainersm, line 648: It took me 5 min looking at this line to realize the "cri" and "crio" tokens! hehehe I was like "why the entries are equal?" /me getting old....

wainersm (approved):
Thanks @stevenhorsman ! This regression really hit me!

fidencio:
/test-tdx

So we can have the result of the QEMU one

fidencio:
Actually, I wouldn't delay it considering everything else's passing.

PR #7596, opened 2023-08-08
tests: Create image that will be used in the unencrypted confidential tests

fidencio:
This will help us to have control of the images we're using, and also make sure we can add to those images whatever we need to make guest detection easier.


stevenhorsman:
I think hadolint doesn't like the Dockerfile:
```
tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile:24 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile:28 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile:33 DL3025 warning: Use arguments JSON notation for CMD and ENTRYPOINT arguments
ERROR: failed to check Dockerfile 'tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile'
```

tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN ssh-keygen -t ed25519 -f /etc/ssh/ssh_host_ed25519_key -P ""
+
+# A password needs to be set for login to work. An empty password is
+# unproblematic as password-based login to root is not allowed.
+RUN passwd -d root
+
+# Generated with `ssh-keygen -t ed25519 -f unencrypted -P "" -C ""`
+COPY ssh/unencrypted.pub /root/.ssh/authorized_keys
+
+ENTRYPOINT /usr/sbin/sshd -D
```
stevenhorsman, line 33: I think best practice recommends the following
```suggestion
ENTRYPOINT ["/usr/sbin/sshd -D"]
```

tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2023 Intel Corporatiion
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ubuntu:22.04
```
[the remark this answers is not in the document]
ChengyuZhu6: I had issues the same as you @stevenhorsman in other projects. https://github.com/containerd/nydus-snapshotter/issues/474

tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    'ARCH=$(uname -m) && \
+    [[ "${ARCH}" == "x86_64" ]] && \
+    curl -LO https://github.com/klauspost/cpuid/releases/download/v2.2.5/cpuid-Linux_x86_64_2.2.5.tar.gz && \
+    tar -xvzf cpuid-Linux_x86_64_2.2.5.tar.gz  -C /usr/bin && \
+    rm -rf cpuid-Linux_x86_64_2.2.5.tar.gz && \
+    rm -f /usr/bin/LICENSE'
+
+# This is done just to avoid the following error starting sshd
+# `sshd: no hostkeys available -- exiting.`
+RUN ssh-keygen -t ed25519 -f /etc/ssh/ssh_host_ed25519_key -P ""
```
stevenhorsman, line 24: `apt-get -y install --no-install-recommends ca-certificates curl openssh-server` seems to generate `/etc/ssh/ssh_host_ed25519_key`, so we hit the following error:
```
 > [4/6] RUN ssh-keygen -t ed25519 -f /etc/ssh/ssh_host_ed25519_key -P "":
0.592 Generating public/private ed25519 key pair.
0.592 /etc/ssh/ssh_host_ed25519_key already exists.
0.592 Overwrite (y/n)?
```
so I'm not sure this is needed if we use the ubuntu base?

tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile
stevenhorsman, line 33: If I remove the shh-keygen to get it to build, I'm now hitting `Missing privilege separation directory: /run/sshd` when I try and run the container locally

tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/Dockerfile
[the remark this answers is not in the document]
fidencio: Oh, I was not aware that the alpine image was actually multi-arch and had support for all the architectures we need (for now and for the future).  Let me just change it.

stevenhorsman (approved):
LGTM. I've tried out the Dockerfile locally and as long as we remember to correctly set the private key to `600` then it works for me:
```
 sudo docker run -d ccv0-ssh
6a9521363068d19e14756f835501ce35382e51e8e7abe0a3b3146cae582d61a0
root@sh-pp-e2e1:~/kata-containers/tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted# ssh -i ssh/unencrypted root@172.17.0.2
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0644 for 'ssh/unencrypted' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "ssh/unencrypted": bad permissions
root@172.17.0.2's password:

root@sh-pp-e2e1:~/kata-containers/tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted# chmod 600 ssh/unencrypted
root@sh-pp-e2e1:~/kata-containers/tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted# ssh -i ssh/unencrypted root@172.17.0.2
Welcome to Alpine!

The Alpine Wiki contains a large amount of how-to guides and general
information about administrating Alpine systems.
See <https://wiki.alpinelinux.org/>.

You can setup the system with the command: setup-alpine

You may change this message by editing /etc/motd.

6a9521363068:~#
```

fidencio:
> and as long as we remember to correctly set the private key to `600`

It's important to remember something that Steven and I learnt Today, GitHub doesn't fully handle file modes.  So, for GitHub either files have the executable bit (and then they're 0755), or otherwise they'll be 0644, and we cannot force anything else.  That's the reason we cannot help the users here by setting this to 0600 on our side.

wainersm (approved):

tests/integration/kubernetes/runtimeclass_workloads/confidential/unencrypted/ssh/unencrypted
```diff
@@ -0,0 +1,7 @@
+-----BEGIN OPENSSH PRIVATE KEY-----
```
ryansavino, line 1: Sorry, I know this one was merged all ready. I was thinking the ssh folder could live two folder levels up so that the key (or anything else related) could be used by any test, not just the confidential unencrypted one. For that matter the Dockerfile above doesn't have anything confidential or unencrypted related either - although maybe it might later?

ryansavino: Another thought is that the key could just get generated on the fly in the script that generates the image, and then we wouldn't have a key under source control. I don't think it's a big deal, but someone may be curious as to why we have keys under source control in a security project. The understanding here of course is that the key is for testing purposes only.

Issue #7597, opened 2023-08-08
Upgrade the bats version


UnmeshDeodhar:
Currently we are using system's package manager to install bats in main branch. If we run this on Ubuntu 20.04, it installs older version of the bats which does not support functions such as `setup_file` and `teardown_file`.
This feature has been introduced in version 1.7: https://github.com/bats-core/bats-core/issues/39
We need to build this from source and install so that we can always get the functionality regardless of what the platform is supporting.
We already do this in the CCv0 branch: https://github.com/kata-containers/tests/blob/CCv0/.ci/install_bats.sh

PR #7598, opened 2023-08-08
tests: upgrade bats version

UnmeshDeodhar:
Instead of using package manager to install bats, building this from source. This gives us the updated version of bats which supports functions such as setup_file and
teardown_file.
We can use these functions into our current tests.

Fixes: #7597

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -47,8 +47,14 @@ function create_cluster() {
 }
 
 function install_bats() {
-    sudo apt-get update
-    sudo apt-get -y install bats
+    # Installing bats from the lunar repo.
+    # This installs newer version of the bats which supports setup_file and teardown_file functions.
+    # These functions are helpful when adding new tests that require one time setup.
+
+    sudo apt install -y software-properties-common
+    sudo add-apt-repository 'deb http://archive.ubuntu.com/ubuntu/ lunar universe'
+    sudo apt install -y bats
+    sudo add-apt-repository --remove 'deb http://archive.ubuntu.com/ubuntu/ lunar universe'
 }
 
 function install_kubectl() {


tests/integration/kubernetes/gha-run.sh
```diff
@@ -47,8 +47,14 @@ function create_cluster() {
 }
 
 function install_bats() {
-    sudo apt-get update
-    sudo apt-get -y install bats
+    BATS_REPO="https://github.com/bats-core/bats-core.git"
+
+    echo "Install BATS from sources"
+    pushd "${GOPATH}/src/"
+    git clone "${BATS_REPO}" || true
```
fidencio, line 54: What happens if `bats` has a bug that blocks us?  Who'll take care of debugging and downgrading this?

I'd like to fuly understand, preferrably with one example, what's exactly needed and how those new setup_file nd teardown_file will be used, to make sure taking this approach is worth the risk.

Moreover, we could add this to versions.yaml and pick it up from there, but again I'm looking at the maintenance side of things and I wonder who'll be responsible to:
1. Check that new releases of bats are available
2. Test that new releases can be updated here
3. Make sure to open PRs once new releases are released

UnmeshDeodhar:
@fidencio thanks for your comments. As I understand it, this will be a very low risk component. And also we can get away with using one version for a long time instead of upgrading it continuously.

tests/integration/kubernetes/gha-run.sh
fidencio: It seems we can do something like:
* Install `add-apt-repository`
  * `apt install -y software-properties-common`
* Enable lunar's universe repo
  * `add-apt-repository 'deb http://archive.ubuntu.com/ubuntu/ lunar universe'
* Install bats from there
  * `apt install -y bats`
* Remove the lunar's repo
  * `add-apt-repository --remove 'deb http://archive.ubuntu.com/ubuntu/ lunar universe'

This makes me way more confident, as the distro theoretically will test the package, than just installing one specific version, or counting on `main` to be always working.

@UnmeshDeodhar, WDYT?

UnmeshDeodhar: Sounds good to me. Let me verify the version it installs.

UnmeshDeodhar: This installs `Bats 1.8.2` so I am just going to change my PR and add this. Thanks @fidencio !

tests/integration/kubernetes/gha-run.sh
```diff
@@ -47,8 +47,10 @@ function create_cluster() {
 }
 
 function install_bats() {
-    sudo apt-get update
-    sudo apt-get -y install bats
+    sudo apt install -y software-properties-common
```
fidencio, line 50: @UnmeshDeodhar, just for the sake of helping my really bad memory, would you mind adding a comment saying we're installing `bats` from `lunar`  because we need a newer version that supports `setup_file` and `teardown_file`?

fidencio (approved):
lgtm, thanks @UnmeshDeodhar!

PR #7604, opened 2023-08-09
Follow up fixes for https://github.com/kata-containers/kata-containers/pull/7596

fidencio:
Please, take a look at each one of the commit messages.


stevenhorsman (approved):
LGTM - thanks. If only we could try out the workflows easily upfront!

Issue #7595, opened 2023-08-08
tests: Have full control of the images that will be used in the confidential tests


fidencio:
RIght now, the only proposed test is using an image that we don't know how it's been built, and we have simple way to modify it.

Let's make sure we will, from now on, version and build those images ourselves.

PR #7607, opened 2023-08-09
ci: create-confidential-image: Add dependent actions

fidencio:
Following the example on https://github.com/docker/build-push-action, it's clear that the actions to "Set up QEMU" and "Set up Docker Buildx" are missing.

Let's add them, and also take the advantage to bump the build-push-action to its v4, which, by the way, had a typo on its name (build-and-push-action does **NOT** exist, build-push-action does).

Fixes: #7595


jepio (approved):

stevenhorsman (approved):
LGTM, but we've seen what that is worth!

wainersm (approved):

Issue #7610, opened 2023-08-10
Upgrade to Firecracker v1.4.0


ManaSugi:
Kata Containers are still using an older version of Firecracker (v1.1.0) released on May 6 2022. Since then, newer versions of Firecracker have been released offering new features. 
We are interested on upgrading Kata Containers runtime to support newer Firecracker versions.

PR #7611, opened 2023-08-10
versions: Update firecracker version to 1.4.0

ManaSugi:
This patch upgrades Firecracker version from v1.1.0 to v1.4.0.

* docs: Specify the supported version of Firecracker using `versions.yaml` 
* Generate swagger models for v1.4.0 (from `firecracker.yaml`)
  - The version of go-swagger used is v0.30.0
* The firecracker v1.4.0 includes the following changes.
  - Added
    * Added support for custom CPU templates allowing users to adjust vCPU features exposed to the guest via CPUID, MSRs and ARM registers. 
    * Introduced V1N1 static CPU template for ARM to represent Neoverse V1 CPU as Neoverse N1. 
    * Added support for the virtio-rng entropy device. The device is optional. A single device can be enabled per VM using the /entropy endpoint. 
    * Added a cpu-template-helper tool for assisting with creating and managing custom CPU templates.
  - Changed
    * Set FDP_EXCPTN_ONLY bit (CPUID.7h.0:EBX[6]) and ZERO_FCS_FDS bit (CPUID.7h.0:EBX[13]) in Intel's CPUID normalization process.
  - Fixed 
    * Fixed feature flags in T2S CPU template on Intel Ice Lake. 
    * Fixed CPUID leaf 0xb to be exposed to guests running on AMD host. 
    * Fixed a performance regression in the jailer logic for closing open file descriptors. 
    * A race condition that has been identified between the API thread and the VMM thread due to a misconfiguration of the api_event_fd. * Fixed CPUID leaf 0x1 to disable perfmon and debug feature on x86 host. 
    * Fixed passing through cache information from host in CPUID leaf 0x80000006. 
    * Fixed the T2S CPU template to set the RRSBA bit of the IA32_ARCH_CAPABILITIES MSR to 1 in accordance with an Intel microcode update. 
    * Fixed the T2CL CPU template to pass through the RSBA and RRSBA bits of the IA32_ARCH_CAPABILITIES MSR from the host in accordance with an Intel microcode update. 
    * Fixed passing through cache information from host in CPUID leaf 0x80000005. 
    * Fixed the T2A CPU template to disable SVM (nested virtualization). 
    * Fixed the T2A CPU template to set EferLmsleUnsupported bit (CPUID.80000008h:EBX[20]), which indicates that EFER[LMSLE] is not supported.

Fixes: #7610


ManaSugi:
The firecracker version of `versions.yaml` used in CI and `firecracker.yaml` sometimes seem to be different.
I'm not sure the policy, but is it good to update both files at the same time?
(Sometimes only `versions.yaml` is updated).
@GabyCT @fidencio @amshinde

amshinde (approved):

amshinde:
@ManaSugi The policy is that both the `versions.yaml` file and swagger file generated `firecracker.yaml` need to be updated at the same time for any version change for firecracker. If this was not done in the past, then this policy has been overlooked. 
Perhaps, we should have this as a CI check to enforce this strictly in the future, perhaps as a github action.

bergwolf (approved):
lgtm!

Issue #6909, opened 2023-05-19
Resource allocation: Potential over-allocation of vCPUs in kata-shim leading to UVM bring-up failure


ms-mahuber:
**Disclaimer:**
  - This issue is one of the four issues forked based the discussion from [issue 6818, using static resource management for TEEs](https://github.com/kata-containers/kata-containers/issues/6818). All four issues are pre-fixed with "Resource allocation:"
  - I am aware I may be a bit deviating a bit from the style of how a bug/request should be opened, but this reflects an obvious issue observed by other peers, so I assume this is fine.

**Description:**
In some scenarios, the shim is prone to request too many vCPUs, leading to pod creation failures when static resource allocation is enabled.
  - Example: podOverhead of 0.25 CPU shares and a node with capacity of 3.86 CPU shares (a node with 4 CPUs and a bit of CPU reserved for the host kubernetes runtime itself, specified via e.g. --kube-reserved=cpu=140m,memory=2662Mi). If a user requests 3.25 CPU shares via a limit, kubernetes accepts the pod yaml (as 3.25+0.25<3.86). Under the plate, the shim tries to allocate 1+ceil(3.25) vCPUs = 5 vCPUs as part of the calculations via calculateVMResources() and as part of the sum being made in SandboxConfig() in src/runtime/pkg/oci/utils.go: SandboxConfig() (note, this example assumes a value of default_vcpus = 1). At least with CH this leads to a pod bring-up failure because CH basically has an assert that does not allow more vCPUs than present CPUs.
  - This will need improvement. While it is plausible to use at least 1 vCPU, adding the limit value this way is bad.
  - Note, that in the dynamic resource management case, this over-allocation can also happen. I am observing warnings where we try to hotplug 4 vCPUs to a 1 vCPU VM in this scenario (see the updateResources() function in src/runtime/virtcontainers/sandbox.go), but cloud-hypervisor rejects this (see in ResizeVCPUs(), only a warning, not an error is issued), and then kata-shim falls back to using 4 vCPUs, i.e., allocates the max number of CPUs available. Unsure what the specified behavior is (whether a warning and best-effort CPU hotplug is desired, or whether even an error should be thrown)

fidencio:
@ms-mahuber and I had a call Yesterday where we discussed this issue.

The approach we're going to take and propose to the community is:
* Use `default_vcpus` and `default_max_vcpus` as floats, which will allow users to specify a CPU share.
[...]

PR #7623, opened 2023-08-11
runtime: Improve vCPU allocation for the VMMs

fidencio:
First of all, this is a controversial piece, and I know that.

In this commit we're trying to make a less greedy approach regards the
amount of vCPUs we allocate for the VMM, which will be advantageous
mainly when using the `static_sandbox_resource_mgmt` feature, which is
used by the confidential guests.

The current approach we have basically does:
* Gets the amount of vCPUs set in the config (an integer)
* Gets the amount of vCPUs set as limit (an integer)
* Sum those up
* Starts / Updates the VMM to use that total amount of vCPUs

The fact we're dealing with integers is logical, as we cannot request
500m vCPUs to the VMMs.  However, it leads us to, in several cases, be
wasting one vCPU.

Let's take the example that we know the VMM requires 500m vCPUs to be
running, and the workload sets 250m vCPUs as a resource limit.

In that case, we'd do:
* Gets the amount of vCPUs set in the config: 1
* Gets the amount of vCPUs set as limit: ceil(0.25)
* 1 + ceil(0.25) = 1 + 1 = 2 vCPUs
* Starts / Updates the VMM to use 2 vCPUs

With the logic changed here, what we're doing is considering everything
as float till just before we start / update the VMM. So, the flow
describe above would be:
* Gets the amount of vCPUs set in the config: 0.5
* Gets the amount of vCPUs set as limit: 0.25
* ceil(0.5 + 0.25) = 1 vCPUs
* Starts / Updates the VMM to use 1 vCPUs

In the way I've written this patch we introduce zero regressions, as
the default values set are still the same, and those will only be
changed for the TEE use cases (although I can see firecracker, or any
other user of `static_sandbox_resource_mgmt=true` taking advantage of
this).

There's, though, an implicit assumption in this patch that we'd need to
make explicit, and that's that the default_vcpus / default_memory is the
amount of vcpus / memory required by the VMM, and absolutely nothing
else.  Also, the amount set there should be reflected in the
podOverhead for the specific runtime class.

One other possible approach, which I am not that much in favour of
taking as I think it's **less clear**, is that we could actually get the
podOverhead amount, subtract it from the default_vcpus (treating the
result as a float), then sum up what the user set as limit (as a float),
and finally ceil the result.  It could work, but IMHO this is **less
clear**, and **less explicit** on what we're actually doing, and how the
default_vcpus / default_memory should be used.

Fixes: #6909

### DIFF
--- a/src/runtime/pkg/katautils/config.go
+++ b/src/runtime/pkg/katautils/config.go
@@ -133,7 +133,7 @@ type hypervisor struct {
 	MemSlots                       uint32                    `toml:"memory_slots"`
 	DefaultBridges                 uint32                    `toml:"default_bridges"`
 	Msize9p                        uint32                    `toml:"msize_9p"`
-	NumVCPUs                       int32                     `toml:"default_vcpus"`
+	NumVCPUs                       float32                   `toml:"default_vcpus"`
 	BlockDeviceCacheSet            bool                      `toml:"block_device_cache_set"`
 	BlockDeviceCacheDirect         bool                      `toml:"block_device_cache_direct"`
 	BlockDeviceCacheNoflush        bool                      `toml:"block_device_cache_noflush"`
@@ -395,17 +395,17 @@ func getCurrentCpuNum() uint32 {
 	return cpu
 }
 
-func (h hypervisor) defaultVCPUs() uint32 {
-	numCPUs := getCurrentCpuNum()
+func (h hypervisor) defaultVCPUs() float32 {
+	numCPUs := float32(getCurrentCpuNum())
 
-	if h.NumVCPUs < 0 || h.NumVCPUs > int32(numCPUs) {
+	if h.NumVCPUs < 0 || h.NumVCPUs > numCPUs {
 		return numCPUs
 	}
 	if h.NumVCPUs == 0 { // or unspecified
-		return defaultVCPUCount
+		return float32(defaultVCPUCount)
 	}
 
-	return uint32(h.NumVCPUs)
+	return h.NumVCPUs
 }
 
 func (h hypervisor) defaultMaxVCPUs() uint32 {
@@ -723,7 +723,7 @@ func newFirecrackerHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		RootfsType:            rootfsType,
 		FirmwarePath:          firmware,
 		KernelParams:          vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
-		NumVCPUs:              h.defaultVCPUs(),
+		NumVCPUsF:             h.defaultVCPUs(),
 		DefaultMaxVCPUs:       h.defaultMaxVCPUs(),
 		MemorySize:            h.defaultMemSz(),
 		MemSlots:              h.defaultMemSlots(),
@@ -857,7 +857,7 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		CPUFeatures:             cpuFeatures,
 		KernelParams:            vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
 		HypervisorMachineType:   machineType,
-		NumVCPUs:                h.defaultVCPUs(),
+		NumVCPUsF:               h.defaultVCPUs(),
 		DefaultMaxVCPUs:         h.defaultMaxVCPUs(),
 		MemorySize:              h.defaultMemSz(),
 		MemSlots:                h.defaultMemSlots(),
@@ -968,7 +968,7 @@ func newAcrnHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		HypervisorCtlPathList: h.CtlPathList,
 		FirmwarePath:          firmware,
 		KernelParams:          vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
-		NumVCPUs:              h.defaultVCPUs(),
+		NumVCPUsF:             h.defaultVCPUs(),
 		DefaultMaxVCPUs:       h.defaultMaxVCPUs(),
 		MemorySize:            h.defaultMemSz(),
 		MemSlots:              h.defaultMemSlots(),
@@ -1059,7 +1059,7 @@ func newClhHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		MachineAccelerators:            machineAccelerators,
 		KernelParams:                   vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
 		HypervisorMachineType:          machineType,
-		NumVCPUs:                       h.defaultVCPUs(),
+		NumVCPUsF:                      h.defaultVCPUs(),
 		DefaultMaxVCPUs:                h.defaultMaxVCPUs(),
 		MemorySize:                     h.defaultMemSz(),
 		MemSlots:                       h.defaultMemSlots(),
@@ -1132,7 +1132,7 @@ func newDragonballHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		ImagePath:       image,
 		RootfsType:      rootfsType,
 		KernelParams:    vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
-		NumVCPUs:        h.defaultVCPUs(),
+		NumVCPUsF:       h.defaultVCPUs(),
 		DefaultMaxVCPUs: h.defaultMaxVCPUs(),
 		MemorySize:      h.defaultMemSz(),
 		MemSlots:        h.defaultMemSlots(),
@@ -1297,7 +1297,7 @@ func GetDefaultHypervisorConfig() vc.HypervisorConfig {
 		MachineAccelerators:      defaultMachineAccelerators,
 		CPUFeatures:              defaultCPUFeatures,
 		HypervisorMachineType:    defaultMachineType,
-		NumVCPUs:                 defaultVCPUCount,
+		NumVCPUsF:                float32(defaultVCPUCount),
 		DefaultMaxVCPUs:          defaultMaxVCPUCount,
 		MemorySize:               defaultMemSize,
 		MemOffset:                defaultMemOffset,
--- a/src/runtime/pkg/oci/utils.go
+++ b/src/runtime/pkg/oci/utils.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"path/filepath"
 	"regexp"
 	goruntime "runtime"
@@ -137,7 +138,7 @@ type RuntimeConfig struct {
 
 	// Sandbox sizing information which, if provided, indicates the size of
 	// the sandbox needed for the workload(s)
-	SandboxCPUs  uint32
+	SandboxCPUs  float32
 	SandboxMemMB uint32
 
 	// Determines if we should attempt to size the VM at boot time and skip
@@ -683,11 +684,11 @@ func addHypervisorMemoryOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig
 func addHypervisorCPUOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig) error {
 	numCPUs := goruntime.NumCPU()
 
-	if err := newAnnotationConfiguration(ocispec, vcAnnotations.DefaultVCPUs).setUintWithCheck(func(vcpus uint64) error {
-		if uint32(vcpus) > uint32(numCPUs) {
-			return fmt.Errorf("Number of cpus %d specified in annotation default_vcpus is greater than the number of CPUs %d on the system", vcpus, numCPUs)
+	if err := newAnnotationConfiguration(ocispec, vcAnnotations.DefaultVCPUs).setFloat32WithCheck(func(vcpus float32) error {
+		if vcpus > float32(numCPUs) {
+			return fmt.Errorf("Number of cpus %f specified in annotation default_vcpus is greater than the number of CPUs %d on the system", vcpus, numCPUs)
 		}
-		sbConfig.HypervisorConfig.NumVCPUs = uint32(vcpus)
+		sbConfig.HypervisorConfig.NumVCPUsF = float32(vcpus)
 		return nil
 	}); err != nil {
 		return err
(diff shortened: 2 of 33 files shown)


dcmiddle (approved):
LGTM.
For my review I searched for areas where there might be an integer to float conversion problem but couldn't find any.
For example, it looks like the toml parser[1] will convert automatically for our existing configs e.g. `default_vcpus = 1` [2].

[1] https://github.com/BurntSushi/toml/pull/325

[2] https://github.com/kata-containers/kata-containers/blob/main/src/runtime/config/configuration-qemu-tdx.toml.in#L104

ms-mahuber:
Thanks you Fabiano, this patch looks very good. One thought: There is a good amount of ``uint32(math.Ceil(...`` occurrences in the PR. How about turning this logic into a single function that is accessible from relevant files? That function pretty much takes a float of vCPU shares as a parameter and converts the input to an integer for the VMM to use. While the function is still super simple and nothing really changes, this makes the change/logic a bit more conceivable. This way we can also add a line of documentation that explains that we get CPU shares from the config and from limits (from kubernetes layer) and we turn those into an integer because the VMM/HV uses absolute numbers and not fractions of CPU shares like in the kubernetes world.

bergwolf (approved):
Makes sense, thanks!

bergwolf:
@fidencio Thanks for raising this! I wonder if it is possible to add an E2E test to validate the new allocation algorithm?

src/runtime/pkg/katautils/config.go
```diff
@@ -131,7 +131,7 @@ type hypervisor struct {
 	MemSlots                       uint32          `toml:"memory_slots"`
 	DefaultBridges                 uint32          `toml:"default_bridges"`
 	Msize9p                        uint32          `toml:"msize_9p"`
-	NumVCPUs                       int32           `toml:"default_vcpus"`
+	NumVCPUs                       float64         `toml:"default_vcpus"`
```
c3d, line 134: Can we use float32 here just to avoid using more memory? I can't see us exceeding 2^23 CPUs anytime soon.

fidencio: @c3d, thanks for the suggestion, and I do believe it makes sense.

I did opt for using `float64` instead of `float32` because the golang's `Ceil` function takes a `float64` as parament and returns a `float64`.  Meaning that we'd need to cast it to `float32` in a lot of places.

Do you think it's still worh doing the casts here and there, to avoid using `float64`?

fidencio: > For example, consider [the int cast in (MacvtapEndpoint) Attach](https://github.com/kata-containers/kata-containers/blob/708b0a305267d5ae0641122e55576067d107182a/src/runtime/virtcontainers/macvtap_endpoint.go#L80)

Very nice catch!

fidencio:
> This would allow to keep the int32 fields and avoid all the castings. Makes sense ?

One problem that I have with this approach is that we need to change the semantics of what's exposed in the configuration file, I think.

By this, I mean, a user would have to expose 500 as 0.5 CPUs, and the "default 1" would have to change.  While I'm not much against those changes, I think this is a big one that could get users confused.

gkurz:
> > This would allow to keep the int32 fields and avoid all the castings. Makes sense ?
> 
> One problem that I have with this approach is that we need to change the semantics of what's exposed in the configuration file, I think.
> 
> By this, I mean, a user would have to expose 500 as 0.5 CPUs, and the "default 1" would have to change. While I'm not much against those changes, I think this is a big one that could get users confused.

As discussed privately on slack, it seems that this PR is trying to address two separate issues :

1. allow the user to specify fractions of CPU in the config file
2. improve the accuracy of vCPU allocation

Turning the int32 into a float32 would allow to address (1) .

My proposal is to fix (2) by translating the float32 from the config into milli CPUs as an int32 and to only use the latter for all the rest of the computation. This would improve precision without having to rely on float64 and `Ceil()` everywhere.

src/runtime/pkg/katautils/config.go
c3d: @fidencio what do you think of [this approach](https://github.com/c3d/kata-containers/tree/topic/runtime-improve-vcpu-allocation-on-host-side). It looks like in practice, the casts are more from `uint32` than from `float64`, and most of the `float` casts replace `float64` casts with `float32`. So I believe in the end it is worth it. WDYT?

gkurz: LGTM @c3d. I've just commented about some oversights in your proposal at https://github.com/kata-containers/kata-containers/commit/285f4de51ff937141b077a1475964e787fe9dca1.

I'd recommend to squash the patches for the sake of bisect.

fidencio:
Folks, I'm taking some time Today to go through Greg's suggestion and hopefully have it closed / merged. :-)

fidencio:
I've rebased this atop of `main`, squashed @c3d's changes on my PR, added @c3d's `Signed-off-by`, and I'll write tests for it.
Please, do not review before I have the tests written.

fidencio:
When writing tests for this I noticed that we need to have something like https://github.com/kata-containers/kata-containers/pull/8404 present, so I just went ahead an opened a PR for that.

Once that one gets merged I will update this one with tests.

src/runtime/pkg/katautils/config.go
```diff
@@ -722,8 +722,8 @@ func newFirecrackerHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		ImagePath:             image,
 		RootfsType:            rootfsType,
 		FirmwarePath:          firmware,
-		KernelParams:          vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
-		NumVCPUs:              h.defaultVCPUs(),
+		KernelParams:          vc.DeserializeParams(strings.Fields(kernelParams)),
```
gkurz, line 725: This change is partially reverting 82ff2db460fc1fbc646e1915cd24b12a2da32b1a. I guess this is not intended and should be dropped. Maybe something went wrong when squashing @c3d 's patch since it still had the previous version.

src/runtime/pkg/katautils/config.go
```diff
@@ -855,9 +855,9 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		PFlash:                  pflashes,
 		MachineAccelerators:     machineAccelerators,
 		CPUFeatures:             cpuFeatures,
-		KernelParams:            vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
+		KernelParams:            vc.DeserializeParams(strings.Fields(kernelParams)),
```
gkurz, line 858: Ditto.

src/runtime/pkg/katautils/config.go
```diff
@@ -967,8 +967,8 @@ func newAcrnHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		HypervisorCtlPath:     hypervisorctl,
 		HypervisorCtlPathList: h.CtlPathList,
 		FirmwarePath:          firmware,
-		KernelParams:          vc.DeserializeParams(vc.KernelParamFields(kernelParams)),
-		NumVCPUs:              h.defaultVCPUs(),
+		KernelParams:          vc.DeserializeParams(strings.Fields(kernelParams)),
```
gkurz, line 970: Ditto.

src/runtime/pkg/oci/utils.go
```diff
@@ -1254,8 +1266,8 @@ func CalculateContainerSizing(spec *specs.Spec) (numCPU, memSizeMB uint32) {
 	return calculateVMResources(period, quota, memory)
 }
 
-func calculateVMResources(period uint64, quota int64, memory int64) (numCPU, memSizeMB uint32) {
-	numCPU = vcutils.CalculateVCpusFromMilliCpus(vcutils.CalculateMilliCPUs(quota, period))
+func calculateVMResources(period uint64, quota int64, memory int64) (numCPU float32, memSizeMB uint32) {
+	numCPU = vcutils.CalculateMilliCPUs(quota, period)
```
gkurz, line 1270: What happened to `CalculateVCpusFromMilliCpus()` ?

src/runtime/pkg/oci/utils.go
```diff
@@ -1183,9 +1183,21 @@ func (a *annotationConfiguration) setUintWithCheck(f func(uint64) error) error {
 	return nil
 }
 
+func (a *annotationConfiguration) setFloat32WithCheck(f func(float32) error) error {
+	if value, ok := a.ocispec.Annotations[a.key]; ok {
+		float64Value, err := strconv.ParseFloat(value, 32)
+		if err != nil || float64Value < 0 {
+			return fmt.Errorf(errAnnotationPositiveNumericKey, a.key)
+		}
+		float32Value := float32(float64Value)
```
c3d, line 1192: While you are checking, maybe add a check for some sane maximum value?

src/runtime/pkg/oci/utils_test.go
c3d, line 1: While you are at it, would you mind fixing the typo in `TestCaluclateVcpusFromMilliCpus`?

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -122,13 +122,12 @@ func WriteToFile(path string, data []byte) error {
...
 // CalculateMilliCPUs converts CPU quota and period to milli-CPUs
-func CalculateMilliCPUs(quota int64, period uint64) uint32 {
-
+func CalculateMilliCPUs(quota int64, period uint64) float32 {
 	// If quota is -1, it means the CPU resource request is
 	// unconstrained.  In that case, we don't currently assign
 	// additional CPUs.
 	if quota >= 0 && period != 0 {
-		return uint32((uint64(quota) * 1000) / period)
+		return float32(quota) / float32(period)
```
gkurz, line 130: With the 1000 factor removed, it seems that the returned value doesn't match the function's contract.

src/runtime/pkg/oci/utils.go
c3d: From what I gather, the return value is now a `float32` and it's propagated all the way up to `config.SandboxCPUs`. So I think that removing it here is legit.

src/runtime/virtcontainers/utils/utils_test.go
```diff
@@ -151,14 +151,19 @@ func TestCalculateMilliCPUs(t *testing.T) {
 	assert := assert.New(t)
 
 	n := CalculateMilliCPUs(1, 1)
-	expected := uint32(1000)
+	expected := float32(1)
```
gkurz, line 154: Ok, I get it now... `CalculateMilliCPUs()` is now returning CPUs in float. It should be renamed to something like  `CalculateCPUsF()` and I'd personally push this semantic change to a separate patch :wink: .

src/runtime/pkg/oci/utils.go
c3d: The actual rounding happens by calling `RoundUpNumVCPUs` from `NumVCPUs()`, which is now a function.

src/runtime/pkg/oci/utils_test.go
c3d: Actually, I think both the function and the test can be removed.

src/runtime/pkg/oci/utils.go
c3d: The function `CalculateVCpusFromMilliCpus` is not needed anymore. I think it should be removed (along with the associated tests, which has a typo in the name anyway).

c3d (changes requested):
Agree with @gkurz that there is a problem with the squashing. Other than that, looks good. Suggested a minor change.

src/runtime/pkg/katautils/config.go
fidencio: It went wrong as part of my rebase. :-)

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2420,10 +2421,10 @@ func (s *Sandbox) calculateSandboxCPUs() (uint32, error) {
 	//  1. BestEffort QoS: no proper support today in Kata.
 	//  2. We could be constrained only by CPUSets. Check for this:
 	if mCPU == 0 && cpusetCount > 0 {
-		return uint32(cpusetCount), nil
+		return float32(cpusetCount), nil
 	}
 
-	return utils.CalculateVCpusFromMilliCpus(mCPU), nil
+	return mCPU, nil
```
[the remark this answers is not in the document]
gkurz: Ok, now that I understand the disappearance of `CalculateVCpusFromMilliCpus()`, I'm back here :wink:

This function is obviously manipulating fractional CPUs, not milli cpus. The `mCPU` variable should be renamed  accordingly.

src/runtime/virtcontainers/utils/utils.go
```diff
@@ -122,13 +122,12 @@ func WriteToFile(path string, data []byte) error {
 }
 
 // CalculateMilliCPUs converts CPU quota and period to milli-CPUs
-func CalculateMilliCPUs(quota int64, period uint64) uint32 {
-
+func CalculateMilliCPUs(quota int64, period uint64) float32 {
```
gkurz, line 125: s/CalculateMilliCPUs/CalculateCPUsF/

src/runtime/virtcontainers/utils/utils.go
fidencio: I'm not sure if I follow you here, @gkurz.  Mind to expand?

src/runtime/pkg/oci/utils.go
fidencio: I think the sane maximm value here would the math.MaxFloat32.
Then whether it's bigger than the amount of CPUs present in the system is something already checked elsewhere.

src/runtime/virtcontainers/utils/utils.go
gkurz: Someone calling `CalculateMilliCPUs()` would legitimately expect to get milli cpus but they would get fractional cpus instead.

fidencio:
@bergwolf, I've removed your approval as I'd like you to, at least, take a look at the last commit, the one adding the tests as you wanted to have it added.

gkurz (approved):
Code changes LGTM. Thanks @fidencio !

I don't have cycles to review the tests commit though.

src/runtime/virtcontainers/hypervisor_config_linux.go
```diff
@@ -54,11 +54,6 @@ func validateHypervisorConfig(conf *HypervisorConfig) error {
 		conf.DefaultMaxVCPUs = defaultMaxVCPUs
 	}
 
-	if conf.ConfidentialGuest && conf.NumVCPUs != conf.DefaultMaxVCPUs {
```
c3d, line 57: Interesting addition to the series.

src/runtime/virtcontainers/utils/utils.go
c3d: @fidencio I had the same reaction reading the code, and I don't see a change here. But it's marked as 'resolved'.

c3d (approved):
Looks good to me. Thanks @fidencio !

tests/integration/kubernetes/k8s-sandbox-vcpus-allocation.bats
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (c) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../common.bash"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	[ "${KATA_HYPERVISOR}" == "dragonball" ] && \
```
bergwolf, line 12: we need something more than dragonball to identify runtime-rs as it now supports clh and qemu/fc are on the way.

bergwolf (approved):
lgtm!

tests/integration/kubernetes/k8s-sandbox-vcpus-allocation.bats
fidencio: Perfect, and this will end up covering more than just this test.
Thanks for the review, @bergwolf!

Issue #7624, opened 2023-08-11
agent: simplify error handling


wedsonaf:
**Which feature do you think can be improved?**

Error handling of the rpc layer.

**How can it be improved?**

It is more verbose than it needs to be because of conversions to ttrpc errors.

PR #7627, opened 2023-08-11
agent: simplify error handling

wedsonaf:
We extend the `Result` and `Option` types with associated types that allows converting a `Result<T, E>` and `Option<T>` into `ttrpc::Result<T>`.

This allows the elimination of many `match` statements in favor of calling the map function plus the `?` operator. This transformation simplifies the code.

Fixes: #7624


fidencio (approved):
lgtm, thanks @wedsonaf!

This is yet another nice cleanup!

jepio (approved):
lgtm

PR #7632, opened 2023-08-12
Fix a possible race condition in image service

jiangliu:
Currently we use a HashMap<image_name, container_id> to maintain relationship between images and containers. If an image is used for multiple containers, it will conflicts and lost information.

Also refine the image service code to prepare for coming work.


fidencio:
@jiangliu, mind to rebase this one?

PR #7643, opened 2023-08-14
gha: tests: Add kata-deploy functional tests -- Part 1

fidencio:
This first part does:
* Splits gha-run.sh from the `integration/kubernetes`  dir, so it can be reused for the `functional/kata-deploy`  tests that will come in the future
* Add the aks / tdx tests, as those are the ones I will be touching
  * sev / snp tests are also welcome but I'll leave this to AMD to play with
* Add a basic placeholder for the gha-run.sh, which will be used by the `functional/kata-deploy` tests.

### DIFF
--- a/.github/workflows/run-kata-deploy-tests-on-aks.yaml
+++ b/.github/workflows/run-kata-deploy-tests-on-aks.yaml
@@ -1,0 +1,77 @@
+name: CI | Run kata-deploy tests on AKS
+on:
+  workflow_call:
+    inputs:
+      registry:
+        required: true
+        type: string
+      repo:
+        required: true
+        type: string
+      tag:
+        required: true
+        type: string
+      pr-number:
+        required: true
+        type: string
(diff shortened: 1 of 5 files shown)


sprt (approved):
LGTM

.github/workflows/run-kata-deploy-tests-on-aks.yaml
```diff
@@ -0,0 +1,77 @@
+name: CI | Run kubernetes tests on AKS
```
wainersm, line 1: rename the job?

fidencio: Just did, thanks!

jepio (approved):

wainersm (approved):

fidencio:
I'm merging this one even with the power CI not running, as this is not related to jenkins at all.

PR #7644, opened 2023-08-14
metrics: Rename tensorflow scripts

GabyCT:
This PR renames the tensorflow scripts to include the data format that is being used as we will have multiple tests with different data and model formats for tensorflow so this will help us to distinguish them.


fidencio (approved):
lgtm, thanks @GabyCT!

dborquez (approved):

PR #7649, opened 2023-08-15
gha: k8s: kata-deploy: Move kata-deploy specific tests from integration/kubernetes to functional/kata-deploy

fidencio:
We're not expanding the tests in any way, shape, form, or fashion.  We're just moving a test that was already part of k8s to its own functional/kata-deploy directory.


tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -0,0 +1,103 @@
...
+		kubectl apply -f "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
+	fi
+	kubectl -n kube-system wait --timeout=10m --for=condition=Ready -l name=kata-deploy pod
+}
+
+@test "Test runtimeclasses are being properly created" {
+	# We filter `kata-mshv-vm-isolation` out as that's present on AKS clusters, but that's not coming from kata-deploy
+	current_runtime_classes=$(kubectl get runtimeclasses | grep -v "kata-mshv-vm-isolation" | grep "kata" | wc -l)
+	[[ ${current_runtime_classes} -eq ${expected_runtime_classes} ]]
+	[[ $(kubectl get runtimeclasses | grep -q "${KATA_HYPERVISOR}") -eq 0 ]]
```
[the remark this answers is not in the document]
fidencio: Thanks, I've fixed this one, rebased atop of the main branch, and we should see the tests running with this take. :-)

tests/functional/kata-deploy/gha-run.sh
```diff
@@ -13,11 +13,34 @@ kata_deploy_dir="$(dirname "$(readlink -f "$0")")"
...
 	pushd "${kata_deploy_dir}"
 	bash run-kata-deploy-tests.sh
 	popd
 }
 
+function cleanup_runtimeclasses() {
+	# Cleanup any runtime class that was left behind in the cluster, in
+	# case of a test failure, apart from the default one that comes from
+	# AKS
+	for rc in `kubectl get runtimeclass -o name | grep -v "kata-mshv-vm-isolation" | sed 's|runtimeclass.node.k8s.io/||'`
```
sprt, line 27: Do we also want to exclude the `runc` runtime class here? I think it's added by AKS, same as `kata-mshv-vm-isolation`.

tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -0,0 +1,119 @@
...
+	grep "${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}" "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" || die "Failed to setup the tests image"
+	echo "::endgroup::"
+	
+	kubectl apply -f "${repo_root_dir}/tools/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml"
+	if [ "${platform}" = "tdx" ]; then
+		kubectl apply -k "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/overlays/k3s"
+	else
+		kubectl apply -f "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
+	fi
+	kubectl -n kube-system wait --timeout=10m --for=condition=Ready -l name=kata-deploy pod
```
sprt, line 56: Any way we can deduplicate this with the code already present in gha-run.sh?

GabyCT (approved):

tests/functional/kata-deploy/gha-run.sh
fidencio: `runc` is the default one, so that's never actually explicitly shown.

tests/functional/kata-deploy/kata-deploy.bats
fidencio: I went back and forth on this one, and in a follow-up PR I will break this code into functions and then re-use parts.

The reason I didn't do this yet, is because we may want to exercise each single one of the env vars passed as part of a different test.

Issue #7650, opened 2023-08-15
runk: containerd integration tests (TestContainerKill*) failed 


ManaSugi:
# Description of problem

When [the containerd integration tests](https://github.com/containerd/containerd/tree/main/integration) are executed with runk,  the `TestContainerKill` and `TestContainerKillAll` failed.

```
=== FAIL: . TestContainerKill (0.10s)
=== PAUSE TestContainerKill
=== CONT  TestContainerKill
    log_hook.go:47: time="2023-08-08T19:30:52.070904682+05:30" level=debug msg="remote introspection plugin filters" func="introspection.(*introspectionRemote).Plugins" file="***/go/src/github.com/containerd/containerd/services/introspection/introspection.go:46" filters="[type==io.containerd.snapshotter.v1, id==overlayfs]" testcase=TestContainerKill
    container_test.go:780: expected error "not found" but received "unknown error after kill: runk did not terminate successfully: exit status 1: ERROR: container TestContainerKill can't be killed because it is Stopped\n: unknown"

=== FAIL: . TestContainerKillAll (0.06s)
=== PAUSE TestContainerKillAll
=== CONT  TestContainerKillAll
    log_hook.go:47: time="2023-08-08T19:31:53.330842914+05:30" level=debug msg="remote introspection plugin filters" func="introspection.(*introspectionRemote).Plugins" file="***/go/src/github.com/containerd/containerd/services/introspection/introspection.go:46" filters="[type==io.containerd.snapshotter.v1, id==overlayfs]" testcase=TestContainerKillAll
    container_linux_test.go:932: failed to create shim task: OCI runtime create failed: runk did not terminate successfully: exit status 1: unknown
```

# Expected result

The both tests will pass.

# Further information

The failure occurs because the error message runk returns is different from the message the tests expect.
The expected message is  [`"container not running"`](https://github.com/containerd/containerd/blob/release/1.7/pkg/process/utils.go#L135) when the kill command is executed with the container's state == Stopped, but runk returns  `"container *** can't be killed because it is Stopped"` and the tests fail.

PR #7651, opened 2023-08-15
runk: Modify kill command's error message for containerd tests

ManaSugi:
The error message when the kill command is executed with the container's 
state == Stopped should be "container not running" because the containerd
tests expect that OCI runtimes return the error message and compare it.  
If the error message is different from the expected one, the tests fail. 
                                                                         
Fixes: #7650                                                             
                                                                         
### DIFF
--- a/src/tools/runk/Cargo.lock
+++ b/src/tools/runk/Cargo.lock
@@ -439,10 +439,31 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @ManaSugi!

amshinde (approved):

Issue #7656, opened 2023-08-16
virtio-mem: virtio-mem doesn't work well for non-4k page size


jongwu:
When host page size is not 4k. kata may abort to run when virtio-mem enabled.

Test on 64k host page size.
error log:

> Unable to add memory device virtiomem: QMP command failed: backend memory size must be multiple of 0x20000000

PR #7657, opened 2023-08-16
runtime: qemu/virtio-mem: fix block size

jongwu:
virtio-mem block size is not a const value, but the current implementation set it to 4M. From [1] we can see that the block size may vary with page size. So, we should caculate blk size depends on host configuration.

Fixes: #7656
[1] https://github.com/qemu/qemu/blob/408af44d04476c633065bfb1eca6865ea93f2984/hw/virtio/virtio-mem.c#L119

### DIFF
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -860,10 +860,11 @@ func (q *qemu) stopVirtiofsDaemon(ctx context.Context) (err error) {
 	return nil
 }
 
-func (q *qemu) getMemArgs() (bool, string, string, error) {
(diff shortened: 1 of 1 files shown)


fidencio:
```
Error: qemu.go:17:2: SA1019: "io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
	"io/ioutil"
```

From the static checks.

jongwu:
Thanks @fidencio -,  fixed.

fidencio (approved):
lgtm, thanks @jongwu!

cmaf (approved):
thanks @jongwu

Issue #7658, opened 2023-08-16
CC | tools: Install remote snapshotter on Kata Containers installation 


ChengyuZhu6:
**Which feature do you think can be improved?**

For `image sharing on the host`  and `image pulling in the guest` without forked containerd, we need to install a remote snapshotter to run on a Kata Containers installation for CoCo .

fidencio:
> For `image sharing on the host` and `image pulling in the guest` without forked containerd, we need to install a remote snapshotter to run on a Kata Containers installation for CoCo .

This is an interesting one that will require some work on the kata-deploy daemonset.

Right now kata-deploy doesn't know anything about snapshotters, but it should and I had planned to work on that as part of adding tests to using devmapper snapshotter (partially orthogonal to the Confidential Containers work).

We need to:
* Start shipping the snapshotters as part of the kata-tarball
  *  cc @wedsonaf here, so we can also ship the tardev-snapshotter
* Adapt kata-deploy.yaml to receive an env-var for which snapshotter should be set per runtime used
* Adapt kata-deploy installation script itself to actually properly:
  * Start the snapshhotter, if needed
  * Configure containerd to set the proper snapshotter for the specific runtime

I **sincerely** don't think we can have this done as part of v0.8.0, but rather as part of v0.9.0.

cc @jiangliu for visibility as well.

fidencio:
So, I went back and forth on this Yesterday, and I do **NOT** think that Kata Containers should be the entity responsible for installing and setting up the snapshotter.

I think this should be done as part of the Confidential Containers operator pre-install hook, instead.  There we have the chance to properly setup everything that's needed for Kata Containers (and any other CC runtime that may exist, such as Enclave CC).

The approach that I'd take is:
* Ensure the snapshotter is installed, configured, and running as part of the pre-install hook on the operator
* Kata Containers' kata-deploy should have a way to be told to setup containerd with a specific snapshotter
  * If such snapshotter is not configured, kata-deploy will fail with a reasonable log message

/cc @bpradipt and @zvonkok for opinions as well.

ChengyuZhu6:
[...]

PR #7659, opened 2023-08-16
gha: nydus: Enable nydus tests

fidencio:
So far we've had a placeholder for the nydus tests, which was always returning "success¨.

Now, let's actually have those tests running as part of our GHA CI.

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -424,3 +424,14 @@ function install_nydus() {
 	sudo tar xfz "${tarball_name}" -C /usr/local/bin --strip-components=1
 	rm -f "${tarball_name}"
 }
+
+function install_nydus_snapshotter() {
+	version="${1}"
+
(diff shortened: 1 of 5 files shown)


ChengyuZhu6:
@fidencio,  Do you plan to add the installation of `nydus-snapshotter` to the kata installation as well ? For CoCo without forked containerd, we need to install a remote snapshotter. https://github.com/kata-containers/kata-containers/issues/7658

fidencio:
> @fidencio, Do you plan to add the installation of `nydus-snapshotter` to the kata installation as well ? For CoCo without forked containerd, we need to install a remote snapshotter. #7658

@ChengyuZhu6, that's ortoghonal to this PR, but we may need to start the discussion, on how do we actually package the snapshotters to be deployed as part of the kata-deploy daemonset (and consequently as part of the Operator).

Let me comment on the issue.

Issue #7615, opened 2023-08-10
Agent Policy should be supported on Guest images that don't use systemd


danmihai1:
**Which feature do you think can be improved?**

The new Agent Policy from https://github.com/kata-containers/kata-containers/issues/7573

**How can it be improved?**

Guest images that use systemd are supported. We currently don't have a good way to start OPA on images that don't use systemd as init.

**Additional Information**

N/A

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?
Yes

PR #7664, opened 2023-08-16
rootfs: agent: Policy support with AGENT_INIT=yes

danmihai1:
When building with AGENT_POLICY=yes and AGENT_INIT=yes:
1. Include OPA and the Policy settings in rootfs.
2. Start OPA from the kata agent.

Before these changes, building with both AGENT_POLICY=yes and AGENT_INIT=yes was unsupported.

Starting OPA from systemd (when AGENT_INIT=no) was already supported.

Fixes: #7615


fidencio (approved):
lgtm, thanks @danmihai1!

sprt (approved):

PR #7654, opened 2023-08-15
tests: k8s: Call ensure_yq() from setup.sh

sprt:

PR #7682, opened 2023-08-17
CC | kata-deploy: Properly create default runtime class

sprt:
The default `kata` runtime class would get created with the `kata`
handler instead of `kata-$KATA_HYPERVISOR`. This made Kata use the wrong
hypervisor and broke CI.

Fixes: https://github.com/kata-containers/kata-containers/issues/7681

Backport of #7654


fidencio (approved):
lgtm, thanks @sprt!

Issue #7667, opened 2023-08-17
Add Agent Policy tests - for cbl-mariner, tdx and sev images


danmihai1:
**Which feature do you think can be improved?**

Agent Policy (https://github.com/kata-containers/kata-containers/issues/7573)

**How can it be improved?**

Need automated tests.

**Additional Information**

N/A

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

Yes

PR #7683, opened 2023-08-17
tests: add policy to existing tests

danmihai1:
tests: add policy to existing tests

A very permissive policy is currently being used for all test pods.
More stringent policies, tailored to the properties of each test pod
will be provided in a future PR.

Fixes: #7667


fidencio:
@danmihai1, you also have to add the new `.bats` file to https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/run_kubernetes_tests.sh#L20

I'd advise to add it as the first test to run, as it'll be skipped anyways for non-TDX, and we can get its results earlier on TDX / TEEs (when enabled on other TEEs as well).

danmihai1:
> @danmihai1, you also have to add the new `.bats` file to https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/run_kubernetes_tests.sh#L20
> 
> I'd advise to add it as the first test to run, as it'll be skipped anyways for non-TDX, and we can get its results earlier on TDX / TEEs (when enabled on other TEEs as well).

Will do, thanks!

sprt:
Shouldn't we rather implement a scheme where we run (1) all tests *without* policy and (2) all tests *with* policy? Or do we know that some tests will fail with the default policy?

fidencio:
@danmihai1, the patches look good, but I'm wondering whether we could have a test that explicitly fails due to the / a specific policy.

danmihai1:
> @danmihai1, the patches look good, but I'm wondering whether we could have a test that explicitly fails due to the / a specific policy.

OK, I will add one such test. 

There are hundreds of times a reasonable policy could reject a pod creation for example, so it's difficult to pick just one. But, I will pick one to start with.

fidencio (approved):
lgtm, thanks @danmihai1!

tests/integration/kubernetes/k8s-exec-rejected.bats
```diff
@@ -0,0 +1,41 @@
...
+
+load "${BATS_TEST_DIRNAME}/../../common.bash"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	get_pod_config_dir
+	pod_name="busybox"
+	pod_yaml="${pod_config_dir}/busybox-pod.yaml"
+
+	# String generated using "base64 -w 0 kata-containers/src/kata-opa/allow-all-except-exec-process.rego"
```
wedsonaf, line 16: Wouldn't it be better to actually generate this string from the file? Or is the file or `base64` not available at runtime?

wedsonaf (approved):

tests/integration/kubernetes/k8s-exec-rejected.bats
danmihai1: I agree but I wasn't familiar enough with the test Host environment - didn't know how easy it is to find base64 on these Hosts. There is always room left for improvements in these tests.

fidencio:
We had all the GHa k8s tests passing, and finally we have the other tests passing, so I'll go ahead and merge this one!

Issue #7616, opened 2023-08-10
Agent Policy should be supported on s390x, powerpc64


danmihai1:
**Which feature do you think can be improved?**

The new Agent Policy from https://github.com/kata-containers/kata-containers/issues/7573

**How can it be improved?**

Add support for s390x, powerpc64. 

**Additional Information**

N/A

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?
Yes

Amulyam24:
Hi @BbolroC , was able to successfully build opa binary, Thanks for the instructions!
```
# file _release/0.55.0/opa_linux_ppc64le_static
_release/0.55.0/opa_linux_ppc64le_static: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), statically linked, with debug_info, not stripped
```
Next step would be to make changes for building it.

Also, they have workflows to publish the binaries for amd, not sure how open they are to supporting other architectures. I have seen previous PRs for arm & [ppc64le support](https://github.com/open-policy-agent/opa/pull/5360) were closed.

BbolroC:
FYI: for a native build:

```
$ git clone -b v0.55.0 https://github.com/open-policy-agent/opa.git && cd opa
$ export WASM_ENABLED=0
$ export DOCKER_RUNNING=0
$ make ci-go-ci-build-linux-static
```

BbolroC:
I will raise a PR for this. Thanks.

zvonkok:
@Amulyam24  @danmihai1  Arm64 support was added in `22 and is published https://github.com/open-policy-agent/opa/releases/download/v0.55.0/opa_linux_arm64_static but only "static" builds. Can we add `aarch64` support as well? I can do some tests on my side to verify.

zvonkok:
Looks like only an update to the `versions.yaml` would be needed and how those versions are deduced in the build scripts.

PR #7694, opened 2023-08-18
WIP: packaging: either download or build OPA

danmihai1:
Download the released binary or build the binary from source code, if the Guest distro didn't provide an OPA package.

Fixes: #7616


fidencio:
@danmihai1, sorry for the radio silence here in the past week(s), things have been busy.

When I suggested this work, I had in mind this to be done in a few steps, not necessarily to be done in the order I'm suggesting here.
* Allow to pass the binary directly to the rootfs builder.
  * Currently the scripts responsible for building the rootfs image / rootfs initrd, which are triggered by the `make rootfs-*-tarball` target do not pass that info down for the binary we currently install there (the agent).
  * I think this should be expanded to deal with the kata-agent binary, OPA binary, CDH binary, attestation-agent binary, etc, but this is out of the scope of this PR, okay?
  * Considering we have a way to pass the binary, we could split the rootfs build from the agent / CDH / attestation-agent build, as in ... have those as pre-reqs via the Makefile targets.

* Allow to build the binary in case there's no official release for this or that architecture
  * This is mostly what @BbolroC raised
  * However, this adds even more time spent on the CI than it already consumes, as we'd be building OPA pretty much for every PR that would change anything related to the rootfs.

* Allow to consume a pre-built binary of OPA, in case there's no official release for this or that architecture, in case there's no change in the OPA version
  * This PR is basically providing the ground work for that
  * We'd build OPA as part a jenkins job that caches the component
    * This has to be added
  * We'd then adapt the code provided by @BbolroC, on his PR, to:
    * Download the text file with the version present on jenkins
    * Check whether that version is the same used in versions.yaml
      * If yes, just download the tarball that's stored in Jenkins, and untar it in the correct location
      * If not, go ahead and build it.

As you can see, the second and third bullet are quite related to each other, and that's the reason I mentioned in a thread on Slack that the work should be combined.

To have an ideal solution, which is not part of this PR, but still part of my wishes, we'd need to also work on the first bullet, but that's not a blocker for this PR, and IIRC @wainersm had a preliminary work on that a long time ago that we could leverage.

With everything said, do you think the plan here is clear @danmihai1?  If not, which part is not clear?

Now, let me actually get some time and review the PR itself.

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -654,7 +654,7 @@ EOF
 			#   	OPA should be built from the cached source code instead of downloading
 			#   	this binary.
 			#
-			opa_bin_url="$(get_package_version_from_kata_yaml externals.open-policy-agent.meta.binary)"
+			opa_bin_url="$(get_package_version_from_kata_yaml externals.open-policy-agent.architecture.${ARCH}.binary)"
```
fidencio, line 657: In this part here, we'll actually need to point to the jenkins job URL, which we'll create.

Then we download the version from there, see whether it matches with the one coming from the versions.yaml file, and then decide whether we need to build it or not.

I'd leave this as a second PR so we can easily get the other parts merged first, and then we just do the switch here.

danmihai1:
Closing because other folk solved these problems better than me.

Issue #7248, opened 2023-07-06
dragonball: complete metrics in vmm and devices


ZizhengBian:
Currently, the metric in Dragonball is not perfect, so we need to improve the metric information for Dragonball's VMM and devices and output them to the runtime.

ZizhengBian:
Let's continue in kata-containers repo!

lisongqian:
The metrics in Dragonball will be upgraded to:

## DragonballMetrics Definition

The   `DragonballMetrics` is used to collect metrics from device and non-device, and output the information to runtime.

```rust
lazy_static! {
    pub static ref METRICS: RwLock<DragonballMetrics> = RwLock::new(DragonballMetrics::default());
}
pub struct DragonballMetrics {
    /// Metrics related to a vcpu's functioning.
    pub vcpu: HashMap<String, Arc<VcpuMetrics>>,
    /// Metrics related to seccomp filtering.
    pub seccomp: SeccompMetrics,
    /// Metrics related to signals.
    pub signals: SignalMetrics,
    // ...
}
```
When a device is created, the device manager add data(`device_id: Arc<deviceMetrics>`) to `DragonballMetrics`  at the same time.
When a device is dropped,  the device manager remove the `Arc` of the removed device in `DragonballMetrics`.

## Metric of Device
The metric of device is a field of device, such as:
```rust
pub struct Vcpu {
    ...
    metrics: Arc<VcpuMetrics>,
}
```
and update with:
```rust
self.metrics.xxx.inc();
```

We can collect the information of metrics with:
```rust
impl Vcpu {
    pub fn metrics(&self) -> Arc<VcpuMetrics> {
        self.metrics.clone()
    }
}
```

## Metric of Non-device

The metric of non-device is updated in `METRICS`, such as:
```rust
METRICS.read().unwrap().signals.sigbus.inc();
```

lisongqian:
the metric in Dragonball  needs to add include:

|  | type                  | link |
|--|-----------------------|-----|
| [] | vcpu                  |     |
| [] | legacy device         |     |
| [] | virtio net device     |     |
| [] | virtio balloon device |     |
| [] | virtio block device   |     |
| [] | virtio vsock  device  |     |
| [] | virtio fs device      |     |
| [] | virtio mmio device    |     |
| [] | vmm action request    |     |
| [] | upcall device         |     |

lisongqian:
the metric in Dragonball  needs to add include:

|              | type                  | link |
[...]

PR #7697, opened 2023-08-19
dragonball: add metrics support for balloon device

lisongqian:
This PR adds metrics support for balloon device and outputs the metrics to runtime.

Fixes: #7248

Signed-off-by: lisongqian <mail@lisongqian.cn>

### DIFF
--- a/src/dragonball/src/dbs_virtio_devices/src/balloon.rs
+++ b/src/dragonball/src/dbs_virtio_devices/src/balloon.rs
@@ -31,7 +31,9 @@ use dbs_interrupt::{InterruptNotifier, NoopNotifier};
 use dbs_utils::epoll_manager::{
     EpollManager, EventOps, EventSet, Events, MutEventSubscriber, SubscriberId,
 };
+use dbs_utils::metric::{IncMetric, SharedIncMetric, SharedStoreMetric, StoreMetric};
 use log::{debug, error, info, trace};
+use serde::Serialize;
 use virtio_bindings::bindings::virtio_blk::VIRTIO_F_VERSION_1;
 use virtio_queue::{QueueOwnedT, QueueSync, QueueT};
 use vm_memory::{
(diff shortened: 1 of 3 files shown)


src/dragonball/src/metric.rs
```diff
@@ -55,4 +58,6 @@ pub struct DragonballMetrics {
     pub seccomp: SeccompMetrics,
     /// Metrics related to signals.
     pub signals: SignalMetrics,
+    /// Metrics related to balloon device.
+    pub balloon: HashMap<String, Arc<BalloonDeviceMetrics>>,
```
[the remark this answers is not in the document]
lisongqian: There will be at least eleven `RwLock ` for the metric since we have various device types, which may increase overhead. And the write operation is only used when creating or removing devices, it has low competitive overhead. So I think one big RwLock may be able to handle it.

src/dragonball/src/device_manager/balloon_dev_mgr.rs
```diff
@@ -275,6 +286,13 @@ impl Default for BalloonDeviceMgr {
     }
 }
 
+impl Drop for BalloonDeviceMgr {
+    fn drop(&mut self) {
+        // todo move METIRCS oprations to remove_device?
```
ZizhengBian, line 291: finish your todo

lisongqian: The balloon device metric should be removed in device_manager since it's added in device_manager. But the device_manager doesn't have the `remove_device` function. So we could remove the metrics here.

src/dragonball/src/dbs_virtio_devices/src/balloon.rs
```diff
@@ -88,6 +90,25 @@ const PAGE_REPORTING_CAPACITY: u16 = 128;
...
+    /// Number of times when handling events on a balloon device.
+    pub event_count: SharedIncMetric,
+    /// Number of times when activate failed on a balloon device.
+    pub activate_fails: SharedIncMetric,
+    /// Number of balloon device inflations.
+    pub inflate_count: SharedIncMetric,
+    /// Number of balloon device deflations.
+    pub deflate_count: SharedIncMetric,
+    /// Memory size(mb) of balloon device.
+    pub balloon_size: SharedStoreMetric,
```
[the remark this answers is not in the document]
lisongqian: Fixed.

src/dragonball/src/device_manager/balloon_dev_mgr.rs
lisongqian: Done.

quanweiZhou (approved):

src/dragonball/src/metric.rs
```diff
@@ -60,6 +61,8 @@ pub struct DragonballMetrics {
...
```
studychao, line 65: Could you add the comment here that `String` refer to balloon id.

lisongqian: I added a comment to explain that the key of HashMap is `DeviceId`.

studychao (approved):

PR #7698, opened 2023-08-20
kata-types: introduce KataVirtualVolume to support nydus, direct volume and image pull

jiangliu:
Introduce a common structure KataVirtualVolume to support Nydus, direct volume and image pulling for Coco, so we can build a common infrastructure for them later.

### DIFF
--- a/src/libs/kata-types/src/mount.rs
+++ b/src/libs/kata-types/src/mount.rs
@@ -340,6 +340,32 @@ impl KataVirtualVolume {
 
         Ok(())
     }
+
+    /// Serialize the virtual volume object to json.
+    pub fn to_json(&self) -> Result<String> {
+        Ok(serde_json::to_string(self)?)
+    }
+
+    /// Deserialize a virtual volume object from json string.
+    pub fn from_json(value: &str) -> Result<Self> {
+        let volume: KataVirtualVolume = serde_json::from_str(value)?;
+        volume.validate()?;
+        Ok(volume)
+    }
+
+    /// Serialize the virtual volume object to json and encode the string with base64.
+    pub fn to_base64(&self) -> Result<String> {
+        let json = self.to_json()?;
+        Ok(base64::encode(json))
+    }
+
+    /// Decode and deserialize a virtual volume object from base64 encoded json string.
+    pub fn from_base64(value: &str) -> Result<Self> {
+        let json = base64::decode(value)?;
+        let volume: KataVirtualVolume = serde_json::from_slice(&json)?;
+        volume.validate()?;
+        Ok(volume)
+    }
 }
 
 impl TryFrom<&DirectVolumeMountInfo> for KataVirtualVolume {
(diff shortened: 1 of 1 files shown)


src/libs/kata-types/src/mount.rs
```diff
@@ -159,6 +136,282 @@ impl NydusExtraOptions {
...
+    pub source: String,
+    /// Filesystem type.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub fs_type: String,
+    /// Mount options.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub options: Vec<String>,
+
+    /// Volume configuration information.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
```
bergwolf, line 214: Can we have something like `pub nydus_volume: Option<NydusVolume>` and `pub guest_image: Option<GuestImage>` so that these fields are self-contained and optional when not needed?

jiangliu: Good point, improved.

bergwolf (approved):

src/libs/kata-types/src/mount.rs
```diff
@@ -159,6 +136,333 @@ impl NydusExtraOptions {
...
+}
+
+/// Information about directly assigned volume.
+#[derive(Debug, Clone, Eq, PartialEq, Default, Serialize, Deserialize)]
+pub struct DirectAssignedVolume {
+    /// Meta information for directly assigned volume.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub metadata: HashMap<String, String>,
+}
+
```
lifupan, line 163: Why we define two different struct with the same contents? Can we merge them into general one?

jiangliu: We define a dedicated structure for different usage scenarios, so we can easily extend/validate the message format.

fidencio:
Adding the `do-not-merge` so myself and @wedsonaf have time to go through this during Today's working day.

fidencio (approved):
@jiangliu, I like the idea of standardising the way of dealing wih nydus, direct volume, and what we'll be doing with the image pull.

It comes with an extra cost of serialising / deserialising the content, but I think we can live with that for the sake of having one standard.

One question that I have, though, is how do you see this common struct expanding and from which moment we offer backward compats (in case we ever do), and cosider this a stable thing.

I'd like to hear from @wedsonaf here as well, but you got my green tick.  Thanks for working on that.

jiangliu:
> One question that I have, though, is how do you see this common struct expanding and from which moment we offer backward compats (in case we ever do), and cosider this a stable thing.
> 

This structure may play two roles:
1) an internal structure to build virtual volume management subsystem for runtime-rs. And provide adaptors to convert existing direct volume config/nydus extra option/mount options to KataVirtualVolume.
2) a message format between snapshotters/CSI drivers and kata-runtime.
My plan is to get 1 done first. 2 is not mandatory, and may coexist with other message format/mount options for ever.

PR #7700, opened 2023-08-21
WIP: tests: really use initrd for testing on mariner

danmihai1:
The k8s tests are using kata-containers.img instead of 
kata-containers-initrd-mariner.img.


danmihai1:
Will experiment with https://github.com/kata-containers/kata-containers/pull/7683 instead.

PR #7702, opened 2023-08-21
[WIP] CC | tools: Install and configure remote snapshotter

ChengyuZhu6:
To avoid using forked containerd in confidential containers, we use remote snapshotter to support image pulling. Consequently, it becomes imperative to install and configure remote snapshotter.


ChengyuZhu6:
close as #7658 said.

Issue #6384, opened 2023-02-28
network: endpoint hotplug (post-vmm boot) breaks QEMU on aarch64 and Firecracker on x86_64 and aarch64   


ananos:
on stable-3.1, we are facing an issue on FC sandboxes regarding networking:

```
# uname -a
Linux hostname 5.10.104-nbfc #4 SMP PREEMPT Mon Feb 13 16:42:13 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
# containerd --version
containerd github.com/containerd/containerd v1.7.0-beta.3 b8962099e7ce533cc62e540ec25f6d1d93e35a0b
# /opt/kata/bin/containerd-shim-kata-v2 --version
Kata Containers containerd shim (Golang): id: "io.containerd.kata.v2", version: 3.1.0-rc0, commit: 4a35d5fa6ea083837670e295370b42d43452374f

### ctr does not add a network endpoint
# ctr run --snapshotter devmapper --runtime io.containerd.run.kata-fc.v2 -t --rm docker.io/library/ubuntu:latest ubuntu-kata-test uname -a
Linux localhost 5.15.26 #2 SMP Mon Feb 13 15:15:16 GMT 2023 aarch64 aarch64 aarch64 GNU/Linux

### nerdctl does add a network endpoint
# nerdctl run --snapshotter devmapper --runtime io.containerd.run.kata-fc.v2 -it --rm ubuntu:latest uname -a
[...]

PR #7703, opened 2023-08-21
runtime: run prestart hooks before starting VM for FC

bergwolf:
Add a new hypervisor capability to tell if it supports device hotplug. If not, we should run prestart hooks before starting new VMs as nerdctl is using the prestart hooks to set up netns. To make nerdctl + FC to work, we need to run the prestart hooks before starting new VMs.

Fixes: #6384

### DIFF
--- a/src/runtime/virtcontainers/acrn_arch_base.go
+++ b/src/runtime/virtcontainers/acrn_arch_base.go
@@ -366,6 +366,7 @@ func (a *acrnArchBase) capabilities(config HypervisorConfig) types.Capabilities 
 
 	caps.SetBlockDeviceSupport()
 	caps.SetBlockDeviceHotplugSupport()
+	caps.SetNetworkDeviceHotplugSupported()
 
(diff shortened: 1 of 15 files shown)


bergwolf:
@ananos Would you like to take a look?

ananos (approved):
LGTM! thanks Tao!

teawater (approved):
LGTM

zvonkok:
@bergwolf Haven't digged through the code, just for completeness, we're not changing the current behaviour? 
One thing that is still outstanding is how do we know which runtime is going to execute the hook? The inner runtime (kata-agent) or the outer runtime (shim)?

bergwolf:
@zvonkok No, the PR does not change the behavior w.r.t. which runtime is executing the hooks. As for your question, we are currently implementing the table outlined here https://github.com/kata-containers/kata-containers/issues/5787#issuecomment-1331748449

PR #7704, opened 2023-08-21
gha: vfio: Import test script

jepio:
This is based on the current vfio test script. Extensions (like cold_plug_vfio) can be added after this is merged so that we always have a passing baseline.


jepio:
This is now ready for review.

src/runtime/virtcontainers/clh.go
```diff
@@ -457,6 +457,11 @@ func (clh *cloudHypervisor) CreateVM(ctx context.Context, id string, network Net
 		return err
 	}
 
+	if !(clh.config.HotPlugVFIO == config.RootPort || clh.config.HotPlugVFIO == config.NoPort) {
+		return fmt.Errorf("Unsupported hotplug vfio value: %s. Must be %s or %s",
+			clh.config.HotPlugVFIO, config.NoPort, config.RootPort)
+	}
+
```
zvonkok, line 464: Since this is a configuration setting, there is `checkPCIeConfig` function that I added for the general use case. Do you think we should move this check there?

tests/common.bash
```diff
@@ -412,4 +412,4 @@ function install_cri_tools() {
 	download_github_project_tarball "${project}" "${version}" "${tarball_name}"
 	sudo tar -xvf "${tarball_name}" -C /usr/local/bin
 	rm -f "${tarball_name}"
-}
+}
```
zvonkok, line 415: No newline at EOF

zvonkok (approved):
LGTM, thanks Jeremi, let's now get the cold_plug and GPU working with this.

fidencio (approved):
lgtm, thanks @jepio!

tests/common.bash
jepio: fixed this

src/runtime/virtcontainers/clh.go
jepio: yes, i had stumbled across this function but didn't think to extend it. done

Issue #7705, opened 2023-08-21
"initrd" pod annotation is ignored if config file specifies an "image"


danmihai1:
Hopefully I understand the intended behavior. I am testing a fix.

When testing on the cbl-mariner distro:

1. A mariner initrd file gets built and installed (https://github.com/kata-containers/kata-containers/blob/fa34d618054f2cd32baf4cb68c5074c2bb6b010c/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L236)

2. A pod annotation is used to specify the path to the mariner initrd file (https://github.com/kata-containers/kata-containers/blob/fa34d618054f2cd32baf4cb68c5074c2bb6b010c/tools/packaging/kata-deploy/scripts/kata-deploy.sh#L120C18-L120C18)

3. However, **the initrd annotation gets ignored** and an unexpected Guest image gets used because of kata's config.toml specifying:

image = "/opt/kata/share/kata-containers/kata-containers.img"

PR #7706, opened 2023-08-21
runtime: use custom initrd, not configured image

danmihai1:
- Use a custom initrd asset instead of a configured image.
- Use a custom images asset instead of a configured initrd.

Fixes: #7705

### DIFF
--- a/src/runtime/virtcontainers/hypervisor.go
+++ b/src/runtime/virtcontainers/hypervisor.go
@@ -705,6 +705,20 @@ func (conf *HypervisorConfig) assetPath(t types.AssetType) (string, error) {
 		return a.Path(), nil
 	}
 
+	if t == types.ImageAsset {
+		_, ok = conf.customAssets[types.InitrdAsset]
+		if ok {
+			// Use the initrd custom asset instead of the configured image.
+			return "", nil
+		}
+	} else if t == types.InitrdAsset {
+		_, ok = conf.customAssets[types.ImageAsset]
+		if ok {
+			// Use the image custom asset instead of the configured initrd.
+			return "", nil
+		}
+	}
+
 	// We could not find a custom asset for the given type, let's
 	// fall back to the configured ones.
 	switch t {


fidencio:
@danmihai1, I wonder if your patch is fixing the issue in the correct place.

After taking a look at the code, which I must say that I'm at best as familiar with it as you're, I noticed that Hypervisor related annotations are being applied here: https://github.com/kata-containers/kata-containers/blob/700759232fd19121002015653a506d6e6c556a0a/src/runtime/pkg/oci/utils.go#L432

And I also noticed that image / initrd paths are not being overwritten there.

Wouldn't that place be more suitable for this fix?

danmihai1:
> @danmihai1, I wonder if your patch is fixing the issue in the correct place.
> 
> After taking a look at the code, which I must say that I'm at best as familiar with it as you're, I noticed that Hypervisor related annotations are being applied here:
> 
> https://github.com/kata-containers/kata-containers/blob/700759232fd19121002015653a506d6e6c556a0a/src/runtime/pkg/oci/utils.go#L432
> 
> And I also noticed that image / initrd paths are not being overwritten there.
> 
> Wouldn't that place be more suitable for this fix?

I have just looked again in both of these places (utils.go and hypervisor.go): 

1. Both deal with overriding config.toml values with annotation values. So, most likely we could fix this issus in either one of these two places.
2. I selfishly preferred to fix the code path related to createAssets() because it's easier for me to understand, compared with utils.go.
3. Let's see if @bergwolf thinks that changing the utils.go code path mentioned by Fabiano would be better.

src/runtime/virtcontainers/hypervisor.go
```diff
@@ -705,6 +705,20 @@ func (conf *HypervisorConfig) assetPath(t types.AssetType) (string, error) {
...
+			// Use the initrd custom asset instead of the configured image.
+			return "", nil
+		}
+	} else if t == types.InitrdAsset {
+		_, ok = conf.customAssets[types.ImageAsset]
+		if ok {
+			// Use the image custom asset instead of the configured initrd.
+			return "", nil
+		}
+	}
```
bergwolf, line 720: It is a bit confusing to return an initrd path for `ImageAsset`or vice versa.

bergwolf (commented):
It seems that you are fixing things in the wrong way. `assetPath` is implemented correctly FWICT. It is the callers that are not handling the annotation vs. config incorrectly. It seems that not just clh.go, qemu.go is not handling it well enough either, where if we configure an image annotation and have a config file setting initrd, the initrd config would override the image annotation as well.

bergwolf:
What about https://github.com/kata-containers/kata-containers/pull/7716? I think we should use a centralized helper to handle the image and initrd asset paths.

danmihai1:
> What about #7716? I think we should use a centralized helper to handle the image and initrd asset paths.

Thanks - I like https://github.com/kata-containers/kata-containers/pull/7716

src/runtime/virtcontainers/hypervisor.go
danmihai1: Just for the record, this code returns:

1. An empty path for ImageAsset when it wants to force the caller to use a custom InitrdAsset.
2. An empty path for InitrdAsset when it wants to force the caller to use a custom ImageAsset.

It doesn't return an initrd path for ImageAsset or vice versa.

With that said, I agree that https://github.com/kata-containers/kata-containers/pull/7716 is a cleaner solution.

Thanks!

danmihai1:
This issue has been fixed by https://github.com/kata-containers/kata-containers/pull/7716

Issue #7713, opened 2023-08-22
runtime-rs: Volumes and rootfs can't be cleaned up while containers exit on its own.


justxuewei:
Create a container by ctr and type `exit` on its console. Then, we can get an error from containerd's log as shown below.

```
Aug 22 10:01:44 devant containerd[1648054]: time="2023-08-22T10:01:44.265177309+08:00" level=error msg="failed to shutdown shim task" error="Others(\"failed to handler message handler request\\n\\nCaused by:\\n    0: do shutdown\\n    1: do the clean up\\n    2: resource clean up\\n    3: failed to cleanup host path\\n    4: failed to remove rw path\\n    5: Resource busy (os error 16)\"): unknown" id=hi
```

Issue #6393, opened 2023-03-01
Container is not cleaned up when it exits by itself


yipengyin:
When container exits by itself, `ProcessStatus` will be stopped. So, cleanup will be skipped. Has anyone fixed this?

https://github.com/kata-containers/kata-containers/blob/785310fe1850fdcc14a1e6892e152e82c50cb2e1/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs#L197-L232

PR #7714, opened 2023-08-22
runtime-rs: Fix volumes and rootfs cleanup issues

justxuewei:
There are several processes for container exit:

- Non-detach mode: `Wait` request is sent by containerd, then
  `stop_process()` will be called eventually.
- Detach mode: `Wait` request is not sent, the `wait_process()` won’t be
  called.
    - Killed by ctr: For example, a container runs `tail -f /dev/null`, and
      is killed by `sudo ctr t kill -a -s SIGTERM <CID>`. Kill request is
      sent, then `kill_process()` will be called. User executes `sudo ctr c
      rm <CID>`, `Delete` request is sent, then `delete_process()` will be
      called.
    - Exited on its own: For example, a container runs `sleep 1s`. The
      container’s state goes to `Stopped` after 1 second. User executes
      the delete command as below.

Where do we do container cleanup things?

- `wait_process()`: No, because it won’t be called in detach mode.
- `delete_process()`: No, because it depends on when the user executes the
  delete command.
- `run_io_wait()`: Yes. A container is considered exited once its IO ended.
  And this always be called once a container is launched.

Fixes: #7713

### DIFF
--- a/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
@@ -4,21 +4,19 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use agent::Agent;
 use anyhow::{Context, Result};
 use awaitgroup::{WaitGroup, Worker as WaitGroupWorker};
 use common::types::{ContainerProcess, ProcessExitStatus, ProcessStateInfo, ProcessStatus, PID};
-use tokio::{
-    io::{AsyncRead, AsyncWrite},
-    sync::{watch, RwLock},
-};
+use tokio::io::{AsyncRead, AsyncWrite};
+use tokio::sync::{watch, RwLock};
 
-use super::{
-    io::{ContainerIo, ShimIo},
-    logger_with_process,
-};
+use super::container::Container;
+use super::io::{ContainerIo, ShimIo};
+use super::logger_with_process;
 
 pub type ProcessWatcher = (
     Option<watch::Receiver<bool>>,
@@ -83,6 +81,7 @@ impl Process {
 
     pub async fn start_io_and_wait(
         &mut self,
+        containers: Arc<RwLock<HashMap<String, Container>>>,
         agent: Arc<dyn Agent>,
         container_io: ContainerIo,
     ) -> Result<()> {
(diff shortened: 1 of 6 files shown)


Ji-Xinyou (approved):
LGTM, but why `kill_process` is accepting a `signal` anyway?

justxuewei:
> LGTM, but why `kill_process` is accepting a `signal` anyway?

I think that there are many signals to kill a process, like `SIGKILL`, `SIGTERM`, etc.

justxuewei:
Hi @yipengyin. This PR reverts your commit 8030e469b22083e7ecf9615555e57a509807655a totally. Could you help test if cleanup funcionality works as expected on your local?

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/process.rs
```diff
@@ -177,12 +186,32 @@ impl Process {
...
+                None => {
+                    error!(
+                        logger,
+                        "Failed to stop process, since container {} not found", container_id
+                    );
+                    return;
+                }
+            };
+
+            if let Err(err) = c.stop_process(&process).await {
```
lifupan, line 202: If we had call stop_process here, then there's no need to call it in  wait_process ?

justxuewei: Yes. Moving here makes sure that the `stop_process()` is called once the container exited. Therefore, calling it in `wait_process()` is unneccessary.

YushuoEdge:
Hi @justxuewei , we add the `Exited` status because of this issue: https://github.com/kata-containers/kata-containers/issues/6393.

Could you please help check if this PR will break the original fixup?

justxuewei:
> Hi @justxuewei , we add the `Exited` status because of this issue: #6393.
> 
> Could you please help check if this PR will break the original fixup?

Yes, I've referenced this issue before. IMO, the `Exited` status could be totally removed, since this PR guarantees that `stop_process()`, which does some cleanup things, is called once container exit. And we've tested several cases to verify this could work.

jongwu:
Retest dragonball on Arm to see if it's stable.
/test-arm-dragonball

justxuewei:
@jongwu @YushuoEdge @lifupan @studychao Could you kindly take a look at this?

quanweiZhou (approved):
Thanks, @justxuewei, LGTM.

Tim-0731-Hzt (approved):

PR #7715, opened 2023-08-22
CC: kata-deploy: Set the snapshotter in the containerd runtime config

fidencio:
This is a patch that should **NOT** be forward ported to main, as there we want to take a cleaner approach on configuring specific snapshotters for specific runtime handlers.

However, for CC, for the v0.8.0 release of CC, this is good enough as it is, and it'll allow us to set one snapshotter for all the deployments done with the CoCo Operator.

This is the Kata Containers counterpart of the work, and there's still work to be done on the Confidential Containers in order to make it work as expected, as:
* Confidential Containers Operator has to expose to the users which snapshotter will be configured
* Confidential Containers Opereator, specifically the pre-install hook, will have to take care of actually installing and configuring the snapshotter, so it can be used.

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -308,7 +308,6 @@ function configure_containerd_runtime() {
 		cat <<EOF | tee -a "$containerd_conf_file"
 [$runtime_table]
   runtime_type = "${runtime_type}"
-  cri_handler = "cc"
   snapshotter = "${SNAPSHOTTER}"
   privileged_without_host_devices = true
   pod_annotations = ["io.katacontainers.*"]
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -309,6 +309,7 @@ function configure_containerd_runtime() {
 [$runtime_table]
   runtime_type = "${runtime_type}"
   cri_handler = "cc"
+  snapshotter = "${SNAPSHOTTER}"
   privileged_without_host_devices = true
   pod_annotations = ["io.katacontainers.*"]
 EOF
@@ -414,6 +415,7 @@ function main() {
 	echo "* DEFAULT_SHIM: ${DEFAULT_SHIM}"
 	echo "* CREATE_RUNTIMECLASSES: ${CREATE_RUNTIMECLASSES}"
 	echo "* CREATE_DEFAULT_RUNTIMECLASS: ${CREATE_DEFAULT_RUNTIMECLASS}"
+	echo "* SNAPSHOTTER: ${SNAPSHOTTER}"
 
 	# script requires that user is root
 	euid=$(id -u)


fidencio:
**NOTE** that this requires a new enough version of containerd, which allows us setting snapshotters per runtime handlers.

BbolroC (approved):
LGTM, thanks @fidencio !

stevenhorsman (commented):
LGTM

stevenhorsman:
Do we have any awareness of whether we are using forked containerd at this point via and environment variable that the operator sets? I wonder if we need to keep `cri_handler = "cc"` for enclave-cc and the other forked cases?

fidencio:
> Do we have any awareness of whether we are using forked containerd at this point via and environment variable that the operator sets? I wonder if we need to keep `cri_handler = "cc"` for enclave-cc and the other forked cases?

Enclave CC will always use the forked contaienrd, this is hardcoded on their side: https://github.com/confidential-containers/enclave-cc/blob/8e99817238a42ca2c9ff06c9b658eb4ab36b7f52/tools/packaging/deploy/enclave-cc-deploy.sh#L120-L122

Then what the Operator simply does is "installing the right containerd to be used" with specific cases.

"the other forked cases", I assume you're talking about peer-pods.
As the remote hypervisor is now deployed as part of the kata-deploy.sh as well, it'd just automatically be using the non-forked containerd.

Let me see if I'm reading your comment correctly.  You want to maintain the forked containerd working with Kata Containers, is it?

stevenhorsman:
> Enclave CC will always use the forked contaienrd, this is hardcoded on their side: https://github.com/confidential-containers/enclave-cc/blob/8e99817238a42ca2c9ff06c9b658eb4ab36b7f52/tools/packaging/deploy/enclave-cc-deploy.sh#L120-L122
 
Cool that answers most of my concerns

> Then what the Operator simply does is "installing the right containerd to be used" with specific cases.
> 
> "the other forked cases", I assume you're talking about peer-pods. As the remote hypervisor is now deployed as part of the kata-deploy.sh as well, it'd just automatically be using the non-forked containerd.

No - my assumption is that peer pods will be using the non-forked containerd for 0.8 release, but given some of the discussions yesterday I don't want to rule out any use case needing to fall back to containerd, so if it's an easy fix to support both (either by including both settings, or selecting the correct one given the `INSTALL_COCO_CONTAINERD` value) then I can't see a problem with doing so. If that's not the case then that's fine and we'll take the risk that we might have to do something different here.

> Let me see if I'm reading your comment correctly. You want to maintain the forked containerd working with Kata Containers, is it?

There is no 'maintain' as this 0.8 release is a dead codepath once it's gone out and for 0.9 based on main then `cri_handler = "cc"` would never be an option as the forked containerd won't go near it.

fidencio:
> No - my assumption is that peer pods will be using the non-forked containerd for 0.8 release, but given some of the discussions yesterday I don't want to rule out any use case needing to fall back to containerd, so if it's an easy fix to support both (either by including both settings, or selecting the correct one given the `INSTALL_COCO_CONTAINERD` value) then I can't see a problem with doing so. If that's not the case then that's fine and we'll take the risk that we might have to do something different here.

So, the concern that I have is not exactly adding a new env-var here, but rather having it properly tested on the operator side.

Let's assume we add a `INSTALL_COCO_CONTAINERD` option here.  Then, on the Operator side we'll have to deal with this and start doing checks like:
* install_coco_containerd is set - snapshotter is also set
* install_coco_containerd is set - snapshotter is not set
* install_coco_containerd is not set - snapshotter is set
* install_coco_containerd is not set - snapshotter is not set

Then we have to think also about adding the tests for the Operator side for those cases, which, as expected, absolutely nobody is taking care of migrating those to the CoCo subscription.

Last but not least, the amount of time spent on a code that will land and will already be dead code.

All in all, I'm not in favour to go the extra mile for this.

stevenhorsman:
> All in all, I'm not in favour to go the extra mile for this.

Ok, it's not a blocker, so as long as enclave-CC is not impacted by this then I don't have a problem with us non investing lots of time to get it working now.

Issue #7717, opened 2023-08-22
kata-agent: block devices are created with permissions set to 0


littlejawa:
# Description of problem

Using cri-o, when creating a container with a block device attached, the container can't access the device because the permissions are set to 0 for the device.
This is because the LinuxDevice structure provided to the runtime has the FileMode field set to 0, and the kata agent is applying it as is.

Looking at what runc does, it actually ignores a FileMode field set to 0 and uses a default value (0o666) instead. The FileMode from the LinuxDevice structure is taken into account only when non-zero.
(https://github.com/opencontainers/runc/blob/ff8c4c7b72ae369361ef76d12590c3165f46bca8/libcontainer/specconv/spec_linux.go#L917)

The kata agent should be modified in the same way in order to behave like runc.

PR #7718, opened 2023-08-22
kata-agent: use default filemode for block device when it is set to 0

littlejawa:
When the FileMode field for the device is unset (0), use a default value instead to allow the use of the device from the container.

This behaviour is seen from cri-o typically.

Fixes: #7717

### DIFF
--- a/src/agent/rustjail/src/lib.rs
+++ b/src/agent/rustjail/src/lib.rs
@@ -423,12 +423,18 @@ fn linux_grpc_to_oci(l: &grpc::Linux) -> oci::Linux {
         let mut r = Vec::new();
 
         for d in l.Devices.iter() {
+            // if the filemode for the device is 0 (unset), use a default value as runc does
+            let filemode = if d.FileMode != 0 {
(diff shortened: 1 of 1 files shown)


littlejawa:
@gkurz FYI

src/agent/rustjail/src/lib.rs
```diff
@@ -423,12 +423,18 @@ fn linux_grpc_to_oci(l: &grpc::Linux) -> oci::Linux {
         let mut r = Vec::new();
 
         for d in l.Devices.iter() {
+            // if the filemode for the device is 0 (unset), use a default value
```
gkurz, line 426: Maybe add `as runc does` ?

gkurz (approved):
LGTM. Thanks @littlejawa !

src/agent/rustjail/src/lib.rs
littlejawa: sure :)

fidencio (approved):
lgtm, thanks @gkurz.

Issue #7420, opened 2023-07-24
Transition from github.com/gogo/protobuf to google.golang.org/protobuf


yoheiueda:
`github.com/gogo/protobuf`has been deprecated.

Dependency to `gogo/protobuf` has been already removed from containred and cri-o.

https://github.com/containerd/containerd/issues/6564
https://github.com/containerd/ttrpc/pull/99
https://github.com/cri-o/cri-o/issues/6350#issuecomment-1309988197
https://github.com/cri-o/cri-o/pull/6382

Kata still has dependency to `gogo/protobuf`, which causes incompatibility with external modules that uses the latest `google.golang.org/protobuf` like this one.

https://github.com/confidential-containers/cloud-api-adaptor/issues/1212

One of the reason of the incompatibility is the fact that an old version of the `genproto` package needs to be used with `gogo/protobuf` by using the `replace` directive in go.mod.

https://github.com/kata-containers/kata-containers/blob/e1a4040a6c810b84df29ee71615b35b49fe0f46c/src/runtime/go.mod#L116

yoheiueda:
I did a quick research on how to migrate from `github.com/gogo/protobuf` to `google.golang.org/protobuf`, but it turns out to be complicated, and we need to update a lot of code and scripts.

I listed what I found so far. I don't think this list is exhaustive. It looks like Kata extensively uses protobuf features. I am not an export of this area, and I think we need some volunteers to fix this issue.

### 1. Remove the replace directive for the `genproto` package in go.mod

https://github.com/kata-containers/kata-containers/blob/e1a4040a6c810b84df29ee71615b35b49fe0f46c/src/runtime/go.mod#L116

### 2. Update the references to containred modules in go.mod
https://github.com/kata-containers/kata-containers/blob/e1a4040a6c810b84df29ee71615b35b49fe0f46c/src/runtime/go.mod#L12-L15

### 3. Update the script that generates protobuf go source files.

  [src/libs/protocols/hack/update-generated-proto.sh](https://github.com/kata-containers/kata-containers/blob/e1a4040a6c810b84df29ee71615b35b49fe0f46c/src/libs/protocols/hack/update-generated-proto.sh#L50-L62)

It may be a good time to migrate from `protoc` to [protobiuld](https://github.com/containerd/protobuild).

[...]

Issue #4150, opened 2022-04-26
TestIoCopy fails due to data race


jodh-intel:
```
12:21:57 === RUN   TestIoCopy
12:21:57 ==================
12:22:13 WARNING: DATA RACE
12:22:13 Write at 0x00c000c32000 by goroutine 199:
12:22:13   github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2.(*ttyIO).close()
12:22:13       /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2/stream.go:40 +0xa0
12:22:13   github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2.ioCopy()
12:22:13       /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2/stream.go:137 +0x278
12:22:13 
12:22:13 Previous read at 0x00c000c32000 by goroutine 205:
12:22:13   github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2.ioCopy.func2()
12:22:13       /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2/stream.go:118 +0x1bc
12:22:13 
12:22:13 Goroutine 199 (running) created at:
12:22:13   github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2.TestIoCopy.func1()
12:22:13       /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2/stream_test.go:179 +0x9b4
12:22:13   github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2.TestIoCopy()
12:22:13       /home/jenkins/workspace/kata-containers-2.0-ubuntu-ARM-PR/go/src/github.com/kata-containers/kata-containers/src/runtime/pkg/containerd-shim-v2/stream_test.go:292 +0x454
12:22:13   testing.tRunner()
12:22:13       /usr/local/go/src/testing/testing.go:1259 +0x198
12:22:13 
[...]

PR #7719, opened 2023-08-22
Remove gogoproto.nullable extension

beraldoleal:
This is the first attempt to remove gogoprotobuff from our code base. For now the idea is just to remove the nullable extension and check the community acceptance of this change.

If approved I will keep working to remove gogoprotobuff.

This is related to #7420


fidencio:
Sorry, clicked the wrong button. :-)

/test

beraldoleal:
@fidencio @wainersm:

```
Error: File was unable to be removed Error: EACCES: permission denied, rmdir '/home/github-runner/actions-runner/_work/kata-containers/kata-containers/pkg/mod/gopkg.in/yaml.v3@v3.0.1/.github'
```

Is this safe to ignore? It seems to be a CI issue, not sure.

src/libs/protocols/hack/update-generated-proto.sh
```diff
@@ -47,7 +47,7 @@ show_usage() {
 }
 
 generate_go_sources() {
-    local cmd="protoc -I$GOPATH/src:$GOPATH/src/github.com/kata-containers/kata-containers/src/libs/protocols/protos \
+    local cmd="protoc -I$GOPATH/src:$(realpath ../)/libs/protocols/protos \
```
wainersm, line 50: hi @beraldoleal !

This change still assume you are running from the `src/<component>`.  I suggest to use the path relative to this script's directory. Something like:
```
script_dir="$(dirname "$(readlink -f "$0")")"
local cmd="protoc -I$GOPATH/src:${script_dir}/../protos \
```

wainersm:
@beraldoleal , what's your overall strategy to https://github.com/kata-containers/kata-containers/issues/7420 ? I see that @yoheiueda listed many changes required. Here you are addressing step 5 `Remove uses of gogo specific types`. If yes, are there many other types that should be removed?

src/libs/protocols/hack/update-generated-proto.sh
beraldoleal: I didn't want to change the current logic here. There is an assumption the script must be executed from `src/agent`. There are other places talking about it.

Here I just would like to make sure it is fetching the source from workdir instead of GOPATH.

I agree with your proposal, but I hadn't time to investigate why this is a requirement.

.gitignore
```diff
@@ -12,6 +12,9 @@ pkg/logging/Cargo.lock
 src/agent/src/version.rs
 src/agent/kata-agent.service
 src/agent/protocols/src/*.rs
+src/runtime/virtcontainers/pkg/agent/protocols/grpc/*.pb.go
```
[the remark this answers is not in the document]
beraldoleal: There is a target called `generate-protocols`. rust builds it automatically but for go we need to call them. So yes, I think we need to call this target with make build, to make sure we are building every time.

beraldoleal:
> @beraldoleal , what's your overall strategy to #7420 ? I see that @yoheiueda listed many changes required. Here you are addressing step 5 `Remove uses of gogo specific types`. If yes, are there many other types that should be removed?

This seems to be the only gogo.protobuf extension and the most critical use here. The idea with this PR is to make sure the community is aware of this change. In case gets approved we can start working on the other tasks.

wainersm:
The jenkins-ci-ARM-ubuntu-20.04-main and jenkins-ci-dragonball-ARM-ubuntu-20.04 jobs failed due a network issue on the machine:

```
11:20:35 ERROR: failed to solve: docker.io/alpine:3.15: failed to authorize: failed to fetch anonymous token: Get "https://auth.ipv6.docker.com/token?scope=repository%3Alibrary%2Falpine%3Apull&service=registry.docker.io": dial tcp [2600:1f18:2148:bc00:e947:5a7f:a49f:9126]:443: connect: cannot assign requested address
11:20:35 make: *** [Makefile:96: /home/jenkins/workspace/kata-containers-2.0-ubuntu20.04-ARM-main/go/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/rootfs-initrd/builddir/initrd-image/.alpine_rootfs.done] Error 1
```

Let's try once again:

/test-arm

wainersm:
The `Static checks / static-checks (make test) (pull_request) `  hit this https://github.com/kata-containers/kata-containers/issues/4150 . I will manually re-run it, hope the fail goes away.

wainersm:
Hi @beraldoleal ! Allow me to invoke the @kata-containers/architecture-committee powers....

Dear AC,

This PR is just the begin of a major effort to get ride of the dependency to `gogo/protobuf`. That tool/lib was deprecated and removed out of projects like cri-o and containerd already; and it is a **major blocker** for implementing remote runtime support for GCP (Google Cloud Platform) on [Peer Pods ](https://github.com/confidential-containers/cloud-api-adaptor).

We will need your support and advisories for delivering that change. @beraldoleal  will be implementing the changes, I will be helping him with testing, CI and general concerns but as I'm not expert on this area of the project and so I'm afraid my approvals shouldn't count much. Therefore, we will need some experts to **help with reviews**.

CC'ing @fengwang666 @lifupan as you have already touched some of the files to be changed. Mind to help?

src/libs/protocols/hack/update-generated-proto.sh
```diff
@@ -47,19 +47,20 @@ show_usage() {
...
 --gogottrpc_out=plugins=ttrpc+fieldpath,\
 import_path=github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc,\
-\
-Mgithub.com/kata-containers/kata-containers/src/libs/protocols/protos/csi.proto=github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc,\
-\
-Mgithub.com/kata-containers/kata-containers/src/libs/protocols/protos/types.proto=github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols,\
-\
-Mgithub.com/kata-containers/kata-containers/src/libs/protocols/protos/oci.proto=github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/grpc,\
-\
```
jepio, line 59: why is this being removed?

jepio (commented):
Commit `8097e4c65cc67dafe20518ae98a29e5d8eb494b7 protocol: use workdir as source file` does not work as intended. If i don't have GOPATH defined then the command line looks like this:
```
protoc -I/src:/home/jeremi/git/confidential-containers/kata-containers/src/libs/protocols/protos ....
```
and fails with:
```
/src: warning: directory does not exist.
/src/: No such file or directory
====================================================================
====                compile protocols failed                    ====

Failed to generate golang file from agent.proto

====================================================================
make: *** [Makefile:224: generate-protocols] Error 1
```

If I do set GOPATH, then the script generates updated definitions in `$GOPATH/src/github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/`. These show changes compared to the checked in ones (not just different randomly generated suffixes but also arguments change from value to pointer).

So this needs checking again.

src/libs/protocols/hack/update-generated-proto.sh
beraldoleal: I tried to explain at the commit message:

>   The definitions are already specified in the .proto files using the
>     go_package option. Centralizing them in one location reduces the
>     potential for errors and simplifies the script.

Any reason we should keep it duplicated?

beraldoleal:
Hi @jepio,

Thank you for your review and testing this... my comments bellow:

> Commit `8097e4c65cc67dafe20518ae98a29e5d8eb494b7 protocol: use workdir as source file` does not work as intended. If i don't have GOPATH defined then the command line looks like this:
> 
> ```
> protoc -I/src:/home/jeremi/git/confidential-containers/kata-containers/src/libs/protocols/protos ....
> ```
> 
> and fails with:
> 
> ```
> /src: warning: directory does not exist.
> /src/: No such file or directory
> ====================================================================
> ====                compile protocols failed                    ====
> 
> Failed to generate golang file from agent.proto
> 
> ====================================================================
> make: *** [Makefile:224: generate-protocols] Error 1
> ```

This scenario is also valid for the current code base. GOPATH was used before and not introduced by this commit. The commit is not trying to fix that. We could, but I would prefer to fix it in a different PR, if you don´t mind. 

> If I do set GOPATH, then the script generates updated definitions in `$GOPATH/src/github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/`. These show changes compared to the checked in ones (not just different randomly generated suffixes but also arguments change from value to pointer).

Yes, this is the point of this PR.  It is a core change, indeed. However, to transition away from gogoprotobuff, it's essential that we remove 'nullable', which will impact our generated structs. I've reviewed the code and couldn't identify any other areas that require modification. With both the build and tests passing successfully, I'm submitting this PR to start the discussions about its implications and to ensure we're on the right trajectory.

src/libs/protocols/hack/update-generated-proto.sh
jepio: No, i didn't catch that

jepio:
> This scenario is also valid for the current code base. GOPATH was used before and not introduced by this commit. The commit is not trying to fix that. We could, but I would prefer to fix it in a different PR, if you don´t mind.
> 

You're changing from GOPATH to `readlink` in one spot. If it doesn't work outside GOPATH yet, then lets drop that change so that we confuse people.

> > If I do set GOPATH, then the script generates updated definitions in `$GOPATH/src/github.com/kata-containers/kata-containers/src/runtime/virtcontainers/pkg/agent/protocols/`. These show changes compared to the checked in ones (not just different randomly generated suffixes but also arguments change from value to pointer).
> 
> Yes, this is the point of this PR. It is a core change, indeed. However, to transition away from gogoprotobuff, it's essential that we remove 'nullable', which will impact our generated structs. I've reviewed the code and couldn't identify any other areas that require modification (maybe I'm missing something). With both the build and tests passing successfully, I'm submitting this PR to start the discussions about its implications and to ensure we're on the right trajectory.

Sure the build and tests pass here since the bindings are not regenerated. Maybe I'm doing something wrong but if i do:
```
export GOPATH=$PWD
mkdir -p src/github.com/kata-containers/
cd src/github.com/kata-containers/
git clone https://github.com/kata-containers/kata-containers
cd kata-containers/
git fetch origin refs/pull/7719/head
git checkout FETCH_HEAD
make generate-protocols
```
then I get a diff and code needs to be changed (haven't done that part): https://github.com/kata-containers/kata-containers/pull/7822.

beraldoleal:
> If it doesn't work outside GOPATH yet, then lets drop that change so that we don't confuse people.

You are right, it is not working. It was, but for some reason, I can't reproduce anymore.
I will give one more shot to incorporate this commit and try to fix it in this PR, since it is a good opportunity.
 
 > then I get a diff and code needs to be changed (haven't done that part): https://github.com/kata-containers/kata-containers/pull/7822.
 
Thank you. I finally could see some errors now. I guess with my local tests outside GOPATH, it was somehow reusing cached pb.go files from a different place I was expecting. :(

The errors are still less than expected, since I removed nullable from a many places. I will be fixing the ones I found so far, and send a force-push.

Besides that... and back to the GOPATH problem... I was planning to remove .pb.go files from this source tree, since those are auto generated and could lead to confusion, if someone update the definition and forgot to commit the output (specially for the cases outside GOPATH). The idea could be removing .pb.go files and adding it to the .gitignore.

What are your thoughts about this? If you agree, I can do more tests to stress it.

gkurz:
> > Also, I'm considering removing the .pb.go files from the source tree. These files are auto-generated and could create confusion if someone updates the definition but forgets to commit the resultant output.

I dislike adding generated files to repos as well. It adds noise and complexity but the benefit isn't generally obvious.

> > I'm updating this PR with a force-push to stress this scenario. Let me know what are your thoughts on this scenario.
> 
> I know nothing about protobuf, but my personal feeling right now is that I'd rather have the generated files checked in and have a static check in the CI that fails if they are not regenerated when the input files are changed.
> 

Having CI to check if people got this right is an example of both noise and complexity.

> They don't change often enough for it to make sense to regenerate them on every system building the repo,

Do you expect a substantial  gain in term of resource usage if these aren't regenerated every time ? Especially if a CI job has to run on every build.

>and it's nice to be able to grep for the code that is being compiled.

Sure, depending on what you're looking for it might be nice to be able to `git grep` but this doesn't look vital enough IMHO. Matter of taste. :wink:

gkurz:
Also you can still `make -C src/agent generate-protocols && git grep --untracked ...` .

jepio:
> > > Also, I'm considering removing the .pb.go files from the source tree. These files are auto-generated and could create confusion if someone updates the definition but forgets to commit the resultant output.
> 
> I dislike adding generated files to repos as well. It adds noise and complexity but the benefit isn't generally obvious.
> 

It's a rarely changing file that is compiled into the binary verbatim. We should track/version control source files. Imagine a PR that touches the proto file and adapts source code to make the repo compile against an invisible file. This makes it harder to follow git history of changes.
 
> > > I'm updating this PR with a force-push to stress this scenario. Let me know what are your thoughts on this scenario.
> > 
> > 
> > I know nothing about protobuf, but my personal feeling right now is that I'd rather have the generated files checked in and have a static check in the CI that fails if they are not regenerated when the input files are changed.
> 
> Having CI to check if people got this right is an example of both noise and complexity.
> 

The CI is there to automate such processes. This is a fairly standard CI check "if one file is modified another file needs to be checked too". 

I'd say regenerating a file on every system with a protoc version that is user supplied is more complexity and less reproducible. Imagine a new protoc version changes a default flag or changes its output or whatever. I'd rather not run that risk.

> > They don't change often enough for it to make sense to regenerate them on every system building the repo,
> 
> Do you expect a substantial gain in term of resource usage if these aren't regenerated every time ? Especially if a CI job has to run on every build.
> 

I don't worry about resource usage. Since the file does not change often, having the PR submitter modify proto files and regenerate go files does not feel like too much process overhead.

> > and it's nice to be able to grep for the code that is being compiled.
> 
> Sure, depending on what you're looking for it might be nice to be able to `git grep` but this doesn't look vital enough IMHO. Matter of taste. 😉

Sure, its a matter of taste. But i prefer the reproducibility and visibility aspects of committing these generated files.

beraldoleal:
> It's a rarely changing file that is compiled into the binary verbatim. We should track/version control source files. Imagine a PR that touches the proto file and adapts source code to make the repo compile against an invisible file. This makes it harder to follow git history of changes.

I see your point. Just another angle: some could agree that .proto files are our  "source files". Also, removing the compiled files, would help with merge conflicts and focused reviews on a single file instead of two. AFAICT, we are already doing that for rust protoc generated files (https://github.com/kata-containers/kata-containers/commit/7559382b15eca32d3d6573df997cd11b44dd7512).

If a .proto file gets touched in a PR, wouldn't that pique our curiosity to dig deeper into its implications? After all, they're kind of our blueprint. The .proto files are our source of truth, so any changes there should be scrutinized, making the 'invisible file' less of an issue.

> I'd say regenerating a file on every system with a protoc version that is user supplied is more complexity and less reproducible. Imagine a new protoc version changes a default flag or changes its output or whatever. I'd rather not run that risk.

Understand the risk with varying protoc versions. One idea could be to pin a specific protoc version in CI/CD for consistency. Haven't explored how doable that is yet, though.

And just to clarify, I'm not strongly against having it or not having it, even if it may appear that way. :) I'm simply interested in exploring this avenue and am enjoying the arguments.

I'm particularly keen to hear more insights on the core issue: the removal of the nullable field. I'd appreciate input from other developers experienced with Protoc. If necessary, I'm open to either dropping the commit related to versioning the .pb.go files or splitting it into a separate PR to facilitate focused discussion.

Appreciate again your comments, @jepio .

jepio:
> I'm particularly keen to hear more insights on the core issue: the removal of the nullable field. I'd appreciate input from other developers experienced with Protoc. If necessary, I'm open to either dropping the commit related to versioning the .pb.go files or splitting it into a separate PR to facilitate focused discussion.

Removing the nullable extension seems perfectly reasonable to me

beraldoleal:
@fidencio @gkurz @wainersm @jepio,

Tests that are failing are not required... By looking at the logs it seems to not be related to this PR. Are they safe to ignore?

jepio (approved):

jepio (approved):

bergwolf (approved):
lgtm!

Issue #7726, opened 2023-08-22
gha: Kata gha runners not working properly


GabyCT:
Seems like the gha runners are not working as I can see the following failure
```
home/runner/work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-binaries-in-docker.sh  --build=cloud-hypervisor
Error: The operation was canceled.
```
See 
https://github.com/kata-containers/kata-containers/actions/runs/5942673345/job/16118367996?pr=7577

fidencio:
This is happening due to the following failure:
```
Run make "${KATA_ASSET}-tarball"
make kernel-nvidia-gpu-snp-tarball-build
make[1]: Entering directory '/home/runner/work/kata-containers/kata-containers'
/home/runner/work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-copy-yq-installer.sh
/home/runner/work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-binaries-in-docker.sh  --build=kernel-nvidia-gpu-snp
Dockerfile:31
--------------------
  29 |     ARG HOST_DOCKER_GID
  30 |     
  31 | >>> RUN if [ ${IMG_USER} != "root" ]; then groupadd --gid=${GID} ${IMG_USER};fi
  32 |     RUN if [ ${IMG_USER} != "root" ]; then adduser ${IMG_USER} --uid=${UID} --gid=${GID};fi
  33 |     RUN if [ ${IMG_USER} != "root" ] && [ ! -z ${HOST_DOCKER_GID} ]; then groupadd --gid=${HOST_DOCKER_GID} docker_on_host;fi
--------------------
ERROR: failed to solve: process "/bin/bash -o pipefail -c if [ ${IMG_USER} != \"root\" ]; then groupadd --gid=${GID} ${IMG_USER};fi" did not complete successfully: exit code: 4
make[1]: *** [tools/packaging/kata-deploy/local-build/Makefile:53: kernel-nvidia-gpu-snp-tarball-build] Error 1
make[1]: Leaving directory '/home/runner/work/kata-containers/kata-containers'
make: *** [tools/packaging/kata-deploy/local-build/Makefile:71: kernel-nvidia-gpu-snp-tarball] Error 2
```

I'm getting the kernel-nvidia-gpu-snp hjere as an example, but it's happening with other targets as well.

fidencio:
It's important to note that:
[...]

PR #7737, opened 2023-08-23
local-build: Remove GID before creating group

jepio:
docker install now creates a group with gid 999 which happens to match what we need to get docker-in-docker to work. Remove the group first as we don't need it.

Fixes: #7726


fidencio (approved):
lgtm, thanks @jepio!

sprt (approved):

GabyCT (approved):

Issue #7732, opened 2023-08-23
runtime: Incorrect handling of non-empty []Endpoint parameter in RemoveEndpoints function


Xuanqing-Shi:
In the `RemoveEndpoints()` `(src/runtime/virtcontainers/network_linux.go)`,  when the `endpoints` slice is empty, it will result in the removal of all endpoints within `n.eps`. Otherwise, if the `endpoints` slice contains specific elements, only those endpoints will be removed. 
```go
func (n *LinuxNetwork) RemoveEndpoints(ctx context.Context, s *Sandbox, endpoints []Endpoint, hotplug bool) error {
	span, ctx := n.trace(ctx, "RemoveEndpoints")
	defer span.End()

	eps := n.eps
	if endpoints != nil {
		eps = endpoints
	}

	for idx, ep := range eps {
		if endpoints != nil {
			new_ep, _ := findEndpoint(ep, n.eps)
			if new_ep == nil {
				continue
			}
		}

		if err := n.removeSingleEndpoint(ctx, s, idx, hotplug); err != nil {
			return err
		}
	}

	networkLogger().Debug("Endpoints removed")

	if n.netNSCreated && endpoints == nil {
		networkLogger().Infof("Network namespace %q deleted", n.netNSPath)
		return deleteNetNS(n.netNSPath)
	}

	return nil
}
```
The operation relies on the use of the `idx` (index) parameter to target the endpoint for removal within the `removeSingleEndpoint()` `(src/runtime/virtcontainers/network_linux.go)`. However, a potential issue arises when using `removeSingleEndpoint()`. Since `removeSingleEndpoint()` removes one endpoint from `n.eps`, when the `endpoints` slice is not empty, the `idx` value refers to the index within the `endpoints` slice rather than `n.eps`. As a result, there is a risk of incorrect endpoint removal occurring.
```go
func (n *LinuxNetwork) removeSingleEndpoint(ctx context.Context, s *Sandbox, idx int, hotplug bool) error {
	if idx > len(n.eps)-1 {
		return fmt.Errorf("Endpoint index overflow")
	}

	endpoint := n.eps[idx]
	...
	n.eps = append(n.eps[:idx], n.eps[idx+1:]...)

	return nil
}
```
[...]

PR #7738, opened 2023-08-23
runtime: incorrect handling of non-empty []Endpoint parameter in Remo…

Xuanqing-Shi:
…veEndpoints function

In the RemoveEndpoints(), when the endpoints paramete isn't empty, using idx may result in wrong endpoint removals. To improve, directly passing the endpoint parameter helps locate the correct elements within n.eps.

Fixes: #7732

### DIFF
--- a/src/runtime/virtcontainers/network_linux.go
+++ b/src/runtime/virtcontainers/network_linux.go
@@ -211,12 +211,17 @@ func (n *LinuxNetwork) addSingleEndpoint(ctx context.Context, s *Sandbox, netInf
 	return endpoint, nil
 }
 
-func (n *LinuxNetwork) removeSingleEndpoint(ctx context.Context, s *Sandbox, idx int, hotplug bool) error {
-	if idx > len(n.eps)-1 {
-		return fmt.Errorf("Endpoint index overflow")
+func (n *LinuxNetwork) removeSingleEndpoint(ctx context.Context, s *Sandbox, endpoint Endpoint, hotplug bool) error {
+	var idx int = len(n.eps)
+	for i, val := range n.eps {
+		if val.HardwareAddr() == endpoint.HardwareAddr() {
+			idx = i
+			break
+		}
 	}
-
-	endpoint := n.eps[idx]
+	if idx == len(n.eps) {
+		return fmt.Errorf("Endpoint not found")
+	}
 
 	if endpoint.GetRxRateLimiter() {
 		networkLogger().WithField("endpoint-type", endpoint.Type()).Info("Deleting rx rate limiter")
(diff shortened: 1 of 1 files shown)


src/runtime/virtcontainers/network_linux.go
```diff
@@ -211,13 +211,18 @@ func (n *LinuxNetwork) addSingleEndpoint(ctx context.Context, s *Sandbox, netInf
...
+	var idx int = len(n.eps)
+	for i, val := range n.eps {
+		if val == endpoint {
+			idx = i
+			break
+		}
+	}
 	if idx > len(n.eps)-1 {
 		return fmt.Errorf("Endpoint index overflow")
 	}
```
justxuewei, line 224: To make it clear, I tend to not to use `idx > len(n.eps)-1`.

```suggestion
	if idx == len(n.eps) {
		return fmt.Errorf("Endpoint not found")
	}
```

src/runtime/virtcontainers/network_linux.go
```diff
@@ -211,13 +211,18 @@ func (n *LinuxNetwork) addSingleEndpoint(ctx context.Context, s *Sandbox, netInf
 	return endpoint, nil
 }
 
-func (n *LinuxNetwork) removeSingleEndpoint(ctx context.Context, s *Sandbox, idx int, hotplug bool) error {
+func (n *LinuxNetwork) removeSingleEndpoint(ctx context.Context, s *Sandbox, endpoint Endpoint, hotplug bool) error {
+	var idx int = len(n.eps)
+	for i, val := range n.eps {
+		if val == endpoint {
```
bergwolf, line 217: Looking at [`findEndpoint`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/endpoint.go#L232), we are relying on the endpoint hardware address to determine if two endpoints are the same. So here we should use
```
if val.HardwareAddr() == endpoint.HardwareAddr() {
```

src/runtime/virtcontainers/network_linux.go
Xuanqing-Shi: Thanks for your feedback! I agree with your suggestion to modify the original source code. I'll make the necessary changes in the PR.

src/runtime/virtcontainers/network_linux.go
Xuanqing-Shi: Thank you for your valuable feedback and suggestions. I have made the necessary modifications to address your concerns. You can review the changes in the updated.

bergwolf (approved):
lgtm, thanks!

bergwolf:
@Xuanqing-Shi Please squash your commits into a single one, and add commit message as explained in https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

teawater (approved):

PR #7745, opened 2023-08-24
gha: vfio: Run on Ubuntu 23.04 runner

jepio:
Move the test to run on Ubuntu 23.04, which already ships kernel v6.2. We need this to get L3 virtualization to work.

To be merged before #7704 so that the other PR can make use of it (thanks GHA...).


fidencio (approved):
lgtm, thanks @jepio!

stevenhorsman (approved):
LGTM. Thanks @jepio!

Issue #7746, opened 2023-08-24
kata-deploy: Don't try to remove /opt/kata


fidencio:
The directory is a host path mount and cannot be removed from within the
container.  What we actually want to remove is whatever is inside that
directory.

PR #7748, opened 2023-08-24
CC | backport | kata-deploy: Don't try to remove /opt/kata

fidencio:
The directory is a host path mount and cannot be removed from within the container.  What we actually want to remove is whatever is inside that directory.

This may raise errors like:
```
rm: cannot remove '/opt/kata/': Device or resource busy
```

Fixes: #7746


fidencio:
There's absolutely no reason to run CI on this PR, as CCv0 branch doesn't test kata-deploy, and it'll only be tested as part of the Operator.

stevenhorsman (approved):
LGTM

PR #7749, opened 2023-08-24
CC | Backport | local-build: Remove GID before creating group

fidencio:
docker install now creates a group with gid 999 which happens to match what we need to get docker-in-docker to work. Remove the group first as we don't need it.

Fixes: #7726
Signed-off-by: Jeremi Piotrowski <jpiotrowski@microsoft.com>
(cherry picked from commit 3b881fbc0edbe3c0979bff0d6f12bbce7215e3bb)

Another one that can simply bypass the CI


stevenhorsman (approved):
LGTM. Thanks @fidencio & @jepio

Issue #7753, opened 2023-08-25
k8s tests: k8s-kill-all-process-in-container.bats using the same test name as k8s-pid-ns.bats


danmihai1:
Probably a copy/paste bug. Tests should use unique names, for easier to understand test logs.

PR #7755, opened 2023-08-25
tests: use unique test name

danmihai1:
k8s-pid-ns.bats was already using the test name from k8s-kill-all-process-in-container.bats - probably a copy/paste bug.

Fixes: #7753


fidencio (approved):
lgtm, thanks @danmihai1!

GabyCT (approved):

PR #7756, opened 2023-08-25
3.2 backport | kata-agent: use default filemode for block device when it is set to 0

littlejawa:
This is a backport of #7718 

When the FileMode field for the device is unset (0), use a default value instead to allow the use of the device from the container. This behaviour is seen from cri-o typically.

Note: this is what runc is doing, which is why regular containers don't have an issue. This change makes sure kata behaves the same as runc.


fidencio (approved):
lgtm, thanks @littlejawa!

GabyCT (approved):

gkurz:
Superseded by #7914 .

PR #7761, opened 2023-08-25
runtime: Fix data race in ioCopy

jepio:
Move the wg.Add(1) of the stdout routine past the close so that we guarantee that the ioCopy goroutine does not run close before all the other goroutines are done.


c3d:
Nice catch. Do you have a reliable reproducer of #5031, or is it a once in a lifetime kind of scenario?

jepio:
> Nice catch. Do you have a reliable reproducer of #5031, or is it a once in a lifetime kind of scenario?

Unfortunately no reproducer. It happens every couple of PRs during test execution. I might just rerun it several times here to see if its fixed.

@bergwolf mentioned his previous PR (with more changes) hit some deadlocks, so I want to see if this approach is more safe.

fidencio (approved):
lgtm, thanks @jepio!

gkurz (approved):
Hi @jepio ! The fix by itself looks good to me but the second paragraph of the change log should be slighty amended : replace `wg.Add(1)` with `wg.Done()`  to match the code change and replace `increments` with `decrements` since that's what `wg.Done()` does.

Approving anyway.

jepio:
Thx, I've reworded the commit message according to your suggestion.

PR #7602, opened 2023-08-09
Refine storage device management for kata-agent

jiangliu:
Refine storage device management for kata-agent by:
- use common code from kata-types and kata-sys-util crate
- maintain a StorageDevice object instead of a reference count for each storage device.
- introduce a manager for storage device handlers.

PR #7763, opened 2023-08-25
CC | cherry-pick #7602 from main branch

jiangliu:
Cherry-pick #7602 from main branch. There's only a minor conflict in src/agent/src/rpc.rs, which should be pretty safe.


fidencio (approved):
lgtm, thanks @jiangliu!

PR #7768, opened 2023-08-28
tdx: Update the components needed for using the 6.2 kernel stack

fidencio:
There are changes on the components needed, based on the tdx-tools repo released / maintained by Intel, and there's also changes in the QEMU command line, which is needed for starting TDX with the 6.2 kernel stack.


jepio (approved):
LGTM.

fidencio:
As the static checks passed, I'm adding the `force-skip-ci` label, as the jenkins jobs do not test this PR in any way, shape, form, or fashion.

PR #7781, opened 2023-08-29
Revert "agent: use function from kata-sys-utils to reduce code"

stevenhorsman:
This reverts commits back to b6218beef69843745430a4159f7163abdebb7de6.
Try to see if CI back working


stevenhorsman:
Fabiano found the issue, so this can close

PR #7787, opened 2023-08-29
metrics: Fix memory inside limits for kata metrics

GabyCT:
This PR fixes the memory inside limit for clh for kata metrics due to the recent changes that we had in the script which impacted in the performance measurement.


jepio (approved):
thanks!

PR #7789, opened 2023-08-29
Fixes tests on AMD machines

beraldoleal:
Some tests are failing when running on AMD machines. I suspect GH CI is not running on amd and is not failing there. Vide individual commits for details.


wainersm (approved):
Good catch @beraldoleal ! LGTM!

jepio (approved):

jepio:
The failing `run-metrics` test is being fixed here https://github.com/kata-containers/kata-containers/pull/7787.

beraldoleal:
@jepio thank you. Can we run again only the failing tests?

jepio:
No need in this case

Issue #7384, opened 2023-07-19
runtime-rs: Update GuestProtection for newer TDX kernels


jodh-intel:
Now that #7236 has landed in `kata-sys-util`, we need to update the code for newer Intel TDX kernels. Specifically, we now only need to check for `/sys/firmware/tdx/` (no CPU flags need to be checked).

Issue #6448, opened 2023-03-10
runtime-rs: cs: Enable Intel TDX


jodh-intel:
#6349 adds the generic confidential guest implementation for Cloud Hypervisor. However, it does not enable TDX as the code needs to detect this feature.

See:

- https://github.com/kata-containers/kata-containers/issues/6383
- https://github.com/kata-containers/kata-containers/pull/6257
- https://github.com/kata-containers/kata-containers/pull/6439/files#diff-731fce1d7a2ca93b9a09456b7b22b3f504c74c973afd404828efc93a20fdf4a8R142

PR #7800, opened 2023-08-30
kata-sys-util: protection: Update TDX checks

jodh-intel:
Update the protection checking code to detect newer versions of Intel TDX (whose userland interface has now stabilised).

> **Note:** that we don't need to retain the existing behaviour since:
>
> - We haven't yet landed the TDX feature (#6448).
> - Systems wishing to use TDX will need to use the latest available
>   system components (such as firmware and host kernel).

Also added an explicit TDX unit test.

Fixes: #7384.

### DIFF
--- a/src/libs/kata-sys-util/src/protection.rs
+++ b/src/libs/kata-sys-util/src/protection.rs
@@ -3,8 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#[cfg(target_arch = "x86_64")]
-use anyhow::anyhow;
 #[cfg(any(target_arch = "s390x", target_arch = "x86_64", target_arch = "aarch64"))]
 use anyhow::Result;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


src/libs/kata-sys-util/src/protection.rs
```diff
@@ -72,42 +70,20 @@ pub fn available_guest_protection() -> Result<GuestProtection, ProtectionError>
...
 }
 
-#[cfg(target_arch = "x86_64")]
-fn retrieve_cpu_flags() -> Result<String> {
-    let cpu_info =
-        crate::cpu::get_single_cpu_info(crate::cpu::PROC_CPUINFO, crate::cpu::CPUINFO_DELIMITER)?;
-
-    let cpu_flags =
-        crate::cpu::get_cpu_flags(&cpu_info, crate::cpu::CPUINFO_FLAGS_TAG).map_err(|e| {
-            anyhow!(
```
fidencio, line 88: As you removed this one, the static checks are complaining about anyhow imported but not used.

fidencio (approved):
lgtm, assuming green tests.
Thanks, @jodh-intel!

jodh-intel:
alpinelinux.org seems to have gone down that caused the tests to fail.

/retest

jiangliu (approved):

PR #7819, opened 2023-09-01
Improve the way to clean up storage devices for sandbox

jiangliu:
Currently there's no way to return error when cleaning up storage devices for sandbox. So add a way to return error and refine the logic to clean up resources.

### DIFF
--- a/src/agent/src/storage/mod.rs
+++ b/src/agent/src/storage/mod.rs
@@ -5,7 +5,6 @@
 //
 
 use std::collections::HashMap;
-use std::fmt::Formatter;
 use std::fs;
 use std::os::unix::fs::{MetadataExt, PermissionsExt};
 use std::path::Path;
@@ -55,31 +54,26 @@ pub struct StorageContext<'a> {
 }
 
 /// An implementation of generic storage device.
+#[derive(Default, Debug)]
 pub struct StorageDeviceGeneric {
-    path: String,
-}
-
-impl std::fmt::Debug for StorageDeviceGeneric {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("StorageDeviceGeneric")
-            .field("path", &self.path)
-            .finish()
-    }
+    path: Option<String>,
 }
 
 impl StorageDeviceGeneric {
     /// Create a new instance of `StorageStateCommon`.
     pub fn new(path: String) -> Self {
-        StorageDeviceGeneric { path }
+        StorageDeviceGeneric { path: Some(path) }
     }
 }
 
 impl StorageDevice for StorageDeviceGeneric {
-    fn path(&self) -> &str {
-        &self.path
+    fn path(&self) -> Option<&str> {
+        self.path.as_ref().map(|v| v.as_str())
     }
 
-    fn cleanup(&self) {}
+    fn cleanup(&self) -> Result<()> {
+        Ok(())
+    }
 }
 
 /// Trait object to handle storage device.
@@ -130,9 +124,10 @@ pub async fn add_storages(
         let path = storage.mount_point.clone();
         let state = sandbox.lock().await.add_sandbox_storage(&path).await;
         if state.ref_count().await > 1 {
-            let path = state.path();
-            if !path.is_empty() {
-                mount_list.push(path.to_string());
+            if let Some(path) = state.path() {
+                if !path.is_empty() {
+                    mount_list.push(path.to_string());
+                }
             }
             // The device already exists.
             continue;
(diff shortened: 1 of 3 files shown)


lifupan (approved):

src/agent/src/storage/mod.rs
```diff
@@ -55,6 +54,86 @@ pub struct StorageContext<'a> {
...
+impl StorageDevice for StorageDeviceGeneric {
+    fn path(&self) -> &str {
+        &self.path
+    }
+
+    fn cleanup(&self) -> Result<()> {
+        if !self.initialized {
+            return Ok(());
+        } else if self.path.is_empty() {
+            return Err(anyhow!("mountpoint path is empty"));
```
quanweiZhou, line 99: If the path is empty, there is no need to clean it up, maybe return `ok()` is better.

src/agent/src/storage/mod.rs
```diff
@@ -55,6 +54,86 @@ pub struct StorageContext<'a> {
...
+/// An implementation of generic storage device.
+pub struct StorageDeviceGeneric {
+    path: String,
+    initialized: bool,
+}
+
+impl std::fmt::Debug for StorageDeviceGeneric {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("StorageDeviceGeneric")
+            .field("path", &self.path)
```
quanweiZhou, line 66: Lack of the field `initialized`.

src/agent/src/storage/mod.rs
```diff
@@ -55,6 +54,86 @@ pub struct StorageContext<'a> {
     sandbox: &'a Arc<Mutex<Sandbox>>,
 }
 
+/// An implementation of generic storage device.
+pub struct StorageDeviceGeneric {
+    path: String,
+    initialized: bool,
```
quanweiZhou, line 60: Can you give more detail for the purpose of the field of `initialized`, currently only `StorageDeviceGeneric::default()` sets the `initialized = false`, What circumstances do `StorageDeviceGeneric::default()` and `StorageDeviceGeneric::new("")` has a difference.

src/agent/src/storage/mod.rs
```diff
@@ -55,6 +54,86 @@ pub struct StorageContext<'a> {
...
+        }
+        if matches!(is_mounted(&self.path), Ok(true)) {
+            return Err(anyhow!("failed to umount mountpoint {}", self.path));
+        }
+
+        if path.is_dir() {
+            let is_empty = path.read_dir()?.next().is_none();
+            if !is_empty {
+                return Err(anyhow!("directory is not empty when clean up storage"));
+            }
```
quanweiZhou, line 119: Maybe the 116-119 is unnecessary because the `remove_dir` can ensure that the directory is empty.

quanweiZhou (commented):
Thanks, @jiangliu, few comments.

src/agent/src/storage/mod.rs
```diff
@@ -55,6 +54,86 @@ pub struct StorageContext<'a> {
...
+            // "remove_dir" will fail if the mount point is backed by a read-only filesystem.
+            // This is the case with the device mapper snapshotter, where we mount the block device
+            // directly at the underlying sandbox path which was provided from the base RO kataShared
+            // path from the host.
+            let _ = fs::remove_dir(path);
+        } else if !path.is_file() {
+            // TODO: should we remove the file for bind mount?
```
quanweiZhou, line 126: If the future such as the use of oss, maybe exist bind mount for files.

src/agent/src/storage/mod.rs
jiangliu: I have refactor the code to use `path: Option<String>` to encode three states:
- uninitialized: None
- empty path: Some("")
- non-empty path: Some(path)

src/agent/src/storage/mod.rs
jiangliu: Fixed

src/agent/src/storage/mod.rs
jiangliu: Fixed

src/agent/src/storage/mod.rs
jiangliu: There may be multiple filesystems mounted at the same destination directory, and umount() only unmounts one level, so hope to give a clear error message for such a case.

src/agent/src/storage/mod.rs
jiangliu: Need to work to ensure that "deleting a data file" is safe:)

quanweiZhou (approved):
Thanks, @jiangliu, LGTM.

Issue #7820, opened 2023-09-01
Possible leakage of storage device


jiangliu:
# Get your issue reviewed faster

To help us understand the problem more quickly, please do the following:

Found a possible leakage of storage device when reading code.

PR #7821, opened 2023-09-01
agent: avoid possible leakage of storage device

jiangliu:
When a storage device is used by more than one container, the second and forth instances will cause storage device reference count leakage, thus cause storage device leakage. The reason is:
add_storages() will increase reference count of existing storage device, but forget to add the device to the `mount_list` array, thus leak the reference count.

Fixes: #7820

### DIFF
--- a/src/agent/src/storage/mod.rs
+++ b/src/agent/src/storage/mod.rs
@@ -103,6 +103,10 @@ pub async fn add_storages(
         let path = storage.mount_point.clone();
         let state = sandbox.lock().await.add_sandbox_storage(&path).await;
         if state.ref_count().await > 1 {
+            let path = state.path();
+            if !path.is_empty() {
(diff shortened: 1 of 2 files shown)


src/agent/src/storage/mod.rs
```diff
@@ -103,6 +103,10 @@ pub async fn add_storages(
         let path = storage.mount_point.clone();
         let state = sandbox.lock().await.add_sandbox_storage(&path).await;
         if state.ref_count().await > 1 {
+            let path = state.path();
```
lifupan, line 106: how about changed it as below:
if let Some(path) = state.path() {
     mount_list.push(path.to_string());
}

lifupan (approved):

quanweiZhou (approved):

Issue #7828, opened 2023-09-03
osbuilder: Cannot create Debian guest rootfs


ManaSugi:
We cannot create a Debian guest rootfs because it hasn't been maintained even though it was restored once by the https://github.com/kata-containers/kata-containers/commit/3093f93a6f466a69bbf8a5eb15ce4b60489582d1.

Some packages for Docker are missing (e.g. `multistrap`, `makedev`) to create the rootfs, but even if they are added, we wouldn't create it properly because the current mechanism to create it has been re-using the [rootfs_lib.sh](https://github.com/kata-containers/kata-containers/blob/27dab249a0b434134ac0bee1595105d7ccbba363/tools/osbuilder/rootfs-builder/ubuntu/rootfs_lib.sh) for an Ubuntu guest rootfs and the script has supported for only the multistrap config for Ubuntu.  
https://github.com/kata-containers/kata-containers/blob/27dab249a0b434134ac0bee1595105d7ccbba363/tools/osbuilder/rootfs-builder/ubuntu/rootfs_lib.sh#L17-L26

As a result, an Ubuntu guest rootfs ends up being created despite our specifying `"debian"` with `rootfs.sh` when we create it.

Let's fix them and activate the Debian guest rootfs for Kata again.

PR #7829, opened 2023-09-03
osbuilder: Activate Debian guest rootfs

ManaSugi:
Activate Debian guest rootfs for Kata using the multistrap. The created rootfs is `bookworm` which is the latest stable version of Debian and the `config.sh` for Debian reuses the one for Ubuntu because the installed packages inside the guest will be same as Ubuntu and this way can improve it's maintainability.

```sh
$ sudo -E kata-runtime exec <sandbox-id>
// Inside Kata VM
root@localhost:/# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

Fixes: #7828

### DIFF
--- a/tools/osbuilder/rootfs-builder/debian/Dockerfile.in
+++ b/tools/osbuilder/rootfs-builder/debian/Dockerfile.in
@@ -1,5 +1,5 @@
-#
 # Copyright (c) 2018 SUSE
+# Copyright (c) 2023 Sony Group Corporation
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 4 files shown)


ManaSugi:
At the same time, we might as well re-enable the Debian job, https://github.com/kata-containers/ci/pull/431.

fidencio:
@ManaSugi, one of the issues we had with multiple rootfs support was the lack of maintenance on those.
I am taking, from this PR, that you'd be helping to maintain the debian one (and related jobs) from the moment we add them.  Is this the case?

ManaSugi:
@fidencio, I understand about the background. Yes, I'll help to maintain the Debian rootfs and the related CI, so I'll send another PR to re-enable https://github.com/kata-containers/ci/pull/431.

fidencio:
> @GabyCT I have a question. Has the CI been running the tests on only `ubuntu` and `centos` guest rootfs?
> QAT jobs seems to be running the `centos` guest rootfs.

CI has only been running tests using the `ubuntu` guest rootfs.  I think the QAT job was disabled while ago.

ManaSugi:
> CI has only been running tests using the `ubuntu` guest rootfs. I think the QAT job was disabled while ago.

Okay, thanks. 
The installed packages inside the Debian rootfs is same as Ubuntu ones because of sharing `config.sh` and `rootfs_lib.sh`. I think it's nice to check whether our all supported rootfs are built properly or not at least by adding the CI to https://github.com/kata-containers/kata-containers/tree/main/.github/workflows. Running the basic tests on those rootfs is better, but the static check is also good as the first approach. 

The checked opeartions wii be as follows:
https://github.com/kata-containers/kata-containers/tree/main/tools/osbuilder/rootfs-builder#build-a-rootfs-using-docker

Could you give me your thoughts?

Issue #7832, opened 2023-09-04
kata-deploy create files owned by a non-root user


jepio:
```
$ ls -la /opt/kata/VERSION
-rw-r--r--. 1 1001 123 1 Aug  2 18:22 /opt/kata/VERSION
```
affects other files as well.

We should ensure the ownership is root:root.

PR #7833, opened 2023-09-04
kata-deploy: Create kata-static.tar with correct ownership

jepio:
Pass --owner and --group to the tar invokation to prevent gihtub runner user from leaking into release artifacts.

Fixes: #7832


fidencio (approved):
lgtm, thanks @jepio!

stevenhorsman (approved):
LGTM. Thanks

Issue #7834, opened 2023-09-04
cached-components: Use a docker image to cache the components instead of Jenkins


fidencio:
As we want to stop relying on Jenkins, let's use a docker image to cache the components instead.

PR #7835, opened 2023-09-04
cached-components: Store the cache components in a docker image

fidencio:
Let's do this in order to get rid of the jenkins jobs used for caching the components.

The work is going to be done in two parts:
* Cache the components in docker images
* Use the cached components from the docker images

This PR addresses the first part only.

It must be said here that I've copied the script from tools/packaging/static-build/cache_components_main.sh, and modified it to become the cache-kata-deploy-binaries.sh one.

When we start using the cached files from the docker images, that script will also be removed.

Fixes: #7834

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -61,6 +61,14 @@ jobs:
           username: ${{ secrets.QUAY_DEPLOYER_USERNAME }}
           password: ${{ secrets.QUAY_DEPLOYER_PASSWORD }}
 
+      - name: Login to Kata Containers ghcr.io
+        if: ${{ inputs.push-to-registry == 'yes' }}
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
       - uses: actions/checkout@v3
         with:
           ref: ${{ inputs.commit-hash }}
@@ -84,6 +92,13 @@ jobs:
           path: kata-build/kata-static-${{ matrix.asset }}.tar.xz
           retention-days: 1
           if-no-files-found: error
+
+      - name: Push ${{ matrix.asset }} to cache
+        if: ${{ inputs.push-to-registry == 'yes' }}
+        run: |
+          make "cache-${KATA_ASSET}-tarball"
+        env:
+          BRANCH: $GITHUB_REF_NAME
 
   create-kata-tarball:
     runs-on: ubuntu-latest
--- a/tools/packaging/kata-deploy/local-build/cache-kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/cache-kata-deploy-binaries.sh
@@ -1,0 +1,285 @@
+#!/bin/bash
+# Copyright (c) 2022 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+source "${script_dir}/../../scripts/lib.sh"
+
+cache_target=""
+ARCH=$(uname -m)
+BRANCH=${BRANCH:-"main"}
+REGISTRY=${REGISTRY:=ghcr.io/kata-containers/cached-components}
+
+KERNEL_FLAVOUR="${KERNEL_FLAVOUR:-kernel}" # kernel | kernel-nvidia-gpu | kernel-experimental | kernel-arm-experimental | kernel-dragonball-experimental | kernel-tdx-experimental | kernel-nvidia-gpu-tdx-experimental | kernel-nvidia-gpu-snp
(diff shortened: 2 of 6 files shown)


.github/workflows/build-kata-static-tarball-amd64.yaml
stevenhorsman, line 1: Is there a technical reason to not cache the arm64 and s390x components?

tools/packaging/kata-deploy/local-build/cache-kata-deploy-binaries.sh
```diff
@@ -0,0 +1,280 @@
...
+	mkdir -p ${cache_dir}/${cache_target}
+	cp "${build_dir}/${component_name}" "${cache_dir}/${cache_target}/${component_name}"
+	sha256sum "${build_dir}/${component_name}" > "${cache_dir}/${cache_target}/sha256sum"
+	cat "${cache_dir}/${cache_target}/sha256sum"
+	echo "${component_version}" > "${cache_dir}/${cache_target}/latest"
+	cat "${cache_dir}/${cache_target}/latest"
+	echo "${component_image}" > "${cache_dir}/${cache_target}/latest_image"
+	cat "${cache_dir}/${cache_target}/latest_image"
+
+	docker build -t ${REGISTRY}:${cache_target}-${BRANCH}-${ARCH} \
```
stevenhorsman, line 167: I'm going to assume that you've ruled it out for good reason before, but in case not (and to let me understand the reason as it might help with caching in other projects): What about using the commit and/or sha256sum/component version as the tag for the artifact's container image (maybe in addition to the current approach). Then if different branches have the same images (which I appreciate it likely to be an uncommon scenario), or if we have to revert back to an older version of builds in same branches (e.g. stable), they could just be pulled directly without only keeping the latest version. Or is the concern about ending up with too many images then?

A more concrete scenario that I'm thinking of is that in the cloud-api-adaptor we pull in the kata-agent binary (which I know isn't currently cached). It's likely that we don't want to always be running ahead `main`, so we will probably pin versions and test them before bumping (on maybe a weekly cadence). If there was a kata-containers cached version of that sha that we could just pick up that would save us building and managing our own cache.

Sorry - maybe off topic and rambling here!

stevenhorsman (approved):
LGTM. Just a couple of questions from me (that might be off topic)

.github/workflows/build-kata-static-tarball-amd64.yaml
fidencio: Oh, I forgot to add this to the arm64 and s390x yaml file, it seems. :-)

zvonkok (approved):
LGTM, Que é um bom trabalho lá!

tools/packaging/kata-deploy/local-build/cache/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM busybox:musl
```
wainersm, line 5: why the musl built busybox? Asking to understand if consumers can face any kind of issue due glibc vs musl

tools/packaging/kata-deploy/local-build/cache-kata-deploy-binaries.sh
```diff
@@ -0,0 +1,285 @@
...
+	mkdir -p ${cache_dir}/${cache_target}
+	cp "${build_dir}/${component_name}" "${cache_dir}/${cache_target}/${component_name}"
+	sha256sum "${build_dir}/${component_name}" > "${cache_dir}/${cache_target}/sha256sum"
+	cat "${cache_dir}/${cache_target}/sha256sum"
+	echo "${component_version}" > "${cache_dir}/${cache_target}/latest"
+	cat "${cache_dir}/${cache_target}/latest"
+	echo "${component_image}" > "${cache_dir}/${cache_target}/latest_image"
+	cat "${cache_dir}/${cache_target}/latest_image"
+
+	docker build \
```
wainersm, line 167: Docker build uses cached layers too, so I am wondering if we shouldn't pass `--no-cache` to avoid silent problems.

For example, if I build `cache_target=kernel` for two different branches, BRANCH=`main` first then BRANCH=`ccv0`,  it will reuse the built layers to build the ccv0 image, won't it?  Because the `COMPONENT` is the same then no cache miss occurs, right?

tools/packaging/kata-deploy/local-build/cache-kata-deploy-binaries.sh
wainersm: Thanks for pointing me to this PR @stevenhorsman !

@fidencio peer pods currently rebuild the kata-agent and other dependencies (attestation-agent, for example). It consumes a significant amount of time that's we are looking forward ways to reduce. One approach is to implement a cache mechanism like in kata; another would be to fetch already built dependencies...you know what is easier :)

wainersm: continuing... kata will be caching the osbuilder entirely, meaning that peer pods will need to open the image to extract the `kata-agent` binary. Is there any problem if we try to cache kata-agent alone?

fidencio:
Closing this one as the ORAS work was merged.

Issue #7838, opened 2023-09-05
CI: switch static-checks-dragonball CI machiens to Azure


studychao:
Previously, static-checks-dragonball is using machines from Alibaba Cloud to run all the CI jobs.

Currently, we are going through an internal process to apply for the new machines for Dragonball CI. Before the internal process is over, we will temporarily use Azure VM to run static-checks-dragonball jobs.

PR #7839, opened 2023-09-05
CI: switch static-checks-dragonball CI machines to Azure

studychao:
Previously, static-checks-dragonball is using machines from Alibaba Cloud to run all the CI jobs.

Currently, we are going through an internal process to apply for the new machines for Dragonball CI. Before the internal process is over, we will temporarily use Azure VM to run static-checks-dragonball jobs.

fixes: #7838


fidencio (approved):
lgtm, thanks @studychao!

jepio:
@studychao you're going to need to install make.

jepio (approved):

Issue #6542, opened 2023-03-29
CI | GHA | Create a PoC for running kubernetes tests using an Azure VM


fidencio:
A proof-of-concept has to be created to run the equivalent of `make kubernetes` tests, but using GitHub actions, on an Azure VM

The expected outcome is:
* A PR with the PoC
* A list of hypervisors being tested
* A list of passing tests per hypervisor tested
* Issues created for the failing tests per hypervisor

This is targetting tests using CRI-O, and tests using devmapper.

PR #7841, opened 2023-09-05
 ci: k8s: Add k8s devmapper tests (part 0)

fidencio:
Let's enable the devmapper kubernetes tests to match exactly what's been
tested as part of the Jenkins CI.

Fixes: #6542


BbolroC (approved):
LGTM, thanks @fidencio

zvonkok (approved):
LGTM, the nice thing is about having tests for everything, it is implicitly documented how to do things the right way :)

Issue #7845, opened 2023-09-06
CC | Upgrade guest kernel for image sharing


ChengyuZhu6:
**Which feature do you think can be improved?**

To support image sharing on the host with erofs tarfs for Confidential Containers, we need to upgrade guest kernel to 6.2.16 with erofs patches.

**How can it be improved?**
This is the v6.2.16 linux kernel branch, featuring support for EROFS tarfs based on the 6.2.16 linux kernel from the upstream: 
https://github.com/erofs/erofs-backports/tree/linux-6.2.y-tarfs

PR #7847, opened 2023-09-06
CC | versions: Bump to 6.4.14 stable kernel

fidencio:
This kernel update is needed in order to get the latest and greatest commits related to EROFS, which will be used for allowing sharing the container images between the guest and host for Confidential Containers using the tarfs mode of EROFS.

We're removing a few options here, because:
* SECURITY_SELINUX_CHECKREQPROT_VALUE was deprecated as part of a7e4676e8e2c.
* CONFIG_IP_NF_TARGET_CLUSTERIP was removed as part of 9db5d918e2c0.
* CONFIG_NET_SCH_CBQ was removed as part of 051d44209842.

Fixes: #7845
Backports: #7846


jiangliu (approved):

GabyCT (approved):

amshinde (approved):

fidencio:
This one should not have been merged without the `main` one being merged first, and we hit issues with the one from main. :-/

Issue #7848, opened 2023-09-06
Hash Sum mismatch for docker-compose-plugin in ubuntu:20.04 for s390x


BbolroC:
# Description of problem

ATM, CI jobs for s390x are failing to the hash sum mismatch for `docker-compose-plugin` while building a `build-kata-deploy` like

```
INFO: apt-cache madison docker-ce-cli | grep '20.10' | head -1 | awk '{$1=$1};1' | cut -d' ' -f 3
[91m+ sh -c DEBIAN_FRONTEND=noninteractive apt-get install -y -qq docker-ce=5:20.10.24~3-0~ubuntu-focal docker-ce-cli=5:20.10.24~3-0~ubuntu-focal containerd.io docker-compose-plugin docker-ce-rootless-extras=5:20.10.24~3-0~ubuntu-focal >/dev/null
[0m[91mE: Failed to fetch https://download.docker.com/linux/ubuntu/dists/focal/pool/stable/s390x/docker-compose-plugin_2.19.1-1~ubuntu.20.04~focal_s390x.deb  Hash Sum mismatch
   Hashes of expected file:
    - SHA512:ada47efffbda879c50a15c7a64c8b382fe5f4afd985e86e4ee529694458abc0b68b4ab679271606bd69720bb7a845b6230d5fbc42a10acec1fd2f6607bf8a443
    - SHA256:5baf45c090caaca6e6b43b1258d654e397ebfd454e0228ee72cb485e60264065
    - SHA1:0d2e1a48c6b3b7b2d6a32223aa76725067f5d349 [weak]
    - MD5Sum:b2737448590743d57c74d51204f9e54a [weak]
    - Filesize:10734326 [weak]
   Hashes of received file:
    - SHA512:46883156f2b0176e5cda69c11cdefda2dc9272e86c2ec11d5b8f42103605c87368cb0a8cfed3982c2f2ff80c466446e118e51c1dc908e4a9a3f7d878af049af2
    - SHA256:99205c1c2bb723d7a2666a01fcd8b3f383f81d9ea808e5223a24d74f1025212b
    - SHA1:f9d39bb74b918c1c2a16a80c33da8521fab5baa8 [weak]
    - MD5Sum:03a46ad6d138e787b6c842f1265b99f9 [weak]
    - Filesize:10734326 [weak]
   Last modification reported: Fri, 25 Aug 2023 16:06:03 +0000
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
[...]

PR #7849, opened 2023-09-06
packaging: do not install docker-compose-plugin for s390x|ppc64le

BbolroC:
This PR is to skip installing docker-compose-plugin while buiding a `build-kata-deploy` image for s390x|ppc64le. It is a temporary solution to fix current CI failures for s390x regarding `hash sum mismatch`.

Fixes: #7848


stevenhorsman (approved):
Looks ok to me. Thanks!

Issue #7840, opened 2023-09-05
Support for direct-volume block devices backed by disk images


alex-matei:
**Which feature do you think can be improved?**

Loop device creation is racy by nature and can cause problems in high density scenarios. 
Excerpt from losetup man page:

```
The loop device setup is not an atomic operation when used with
--find, and losetup does not protect this operation by any lock.
The number of attempts is internally restricted to a maximum of
16.
```

Kata should have the ability to attach direct-volumes (or rootfs) backed by disks images directly 
without the need to first create a loop block device in kata-runtime/CSI drivers. 
Qemu and Cloud-hypervisor support qcow and raw image formats, support for both
would be useful for various scenarios, like fast volume creation using CoW semantics.

PR #7850, opened 2023-09-06
Support for direct-volume block devices backed by disk images (loop mounts)

alex-matei:
This patch adds support for attaching direct-volumes where the source is a disk image file. It does so by using the loop (preferred) or io.katacontainers.fs-opt.block_device=file mount option. If the loop option is found, the file/device is attached as a block device to vm and kata-agent mounts it (without the loop option). 

The new `blockfile` snapshotter introduced in containerd 1.7.4 can also be supported using the loop mount option. I'll address this feature in another PR.

I fixed a few issues along the way:

1. There is a regression in kata agent when trying to use regular files as block devices 
2. In case the host path for the device points to a regular file on host the sandbox cgroup controller will generate an error
3. There is a hack where the Format field in BlockDrive is used for fstype. This field should be used for the format of the actual file/device on disk, not for the contents inside the drive. I removed the hack and documented what Format is actually supposed to mean.

Fixes: #7840 #7858 #7859

### DIFF
--- a/src/libs/kata-sys-util/src/mount.rs
+++ b/src/libs/kata-sys-util/src/mount.rs
@@ -399,7 +399,9 @@ pub fn parse_mount_options<T: AsRef<str>>(options: &[T]) -> Result<(MsFlags, Str
     let mut data: Vec<String> = Vec::new();
 
     for opt in options.iter() {
-        if opt.as_ref() == "loop" {
+        if opt.as_ref().starts_with("io.katacontainers.") {
+            continue;
+        } else if opt.as_ref() == "loop" {
             return Err(Error::InvalidMountOption("loop".to_string()));
         } else if let Some(v) = parse_mount_flags(flags, opt.as_ref()) {
             flags = v;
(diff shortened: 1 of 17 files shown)


fidencio:
@wedsonaf, @jiangliu, can you take a look at this?

I think we need this one also backported to the CCv0 branch.

alex-matei:
I fixed the error regarding cyclomatic complexity in `(*qemu).hotplugAddBlockDevice`

alex-matei:
@wedsonaf @jiangliu Can you please take a quick look on this PR? It fixes an issue with tardev and enables usage of direct volumes backed by disk images.

alex-matei:
Can someone please take a look?

ChengyuZhu6:
Hi @alex-matei Could you please continue the excellent work? There have been a lot of patches merged into main recently. You might need to rebase this PR, and then reviewers can help with the review.

alex-matei:
Hi @ChengyuZhu6 . I did a rebase and resolved the conflicts. Please take a look when you have time. This PR https://github.com/kata-containers/kata-containers/pull/7997 is a follow-up of the current one, it was closed because of staleness. Can you please reopen it?

Issue #7853, opened 2023-09-06
runtime: Allow virtio_fs_extra_args annotation


fidencio:
Some use cases may just require passing extra arguments to virtiofsd,
and having this disabled by default makes it impossible to set when
using kata-deploy, as changes in the configuration file would be
overwritten by the daemon-set.

With this in mind, let's allow users to pass whatever thet need (and
here I'm specifically looking at `--xattr`) as a virtio_fs_extra_arg.

PR #7854, opened 2023-09-06
runtime: Allow virtio_fs_extra_args annotation

fidencio:
Some use cases may just require passing extra arguments to virtiofsd, and having this disabled by default makes it impossible to set when using kata-deploy, as changes in the configuration file would be overwritten by the daemon-set.

With this in mind, let's allow users to pass whatever thet need (and here I'm specifically looking at `--xattr`) as a virtio_fs_extra_arg.

Fixes: #7853


stevenhorsman (approved):
LGTM. Thanks!

PR #7865, opened 2023-09-07
runtime: Fix more virtiofs args

gkurz:
Update some more virtiofsd arguments from the legacy syntax of the C implementation to the new syntax used by the rust implementation.

### DIFF
--- a/src/runtime/virtcontainers/virtiofsd_test.go
+++ b/src/runtime/virtcontainers/virtiofsd_test.go
@@ -76,15 +76,15 @@ func TestVirtiofsdArgs(t *testing.T) {
 	v := &virtiofsd{
 		path:       "/usr/bin/virtiofsd",
 		sourcePath: "/run/kata-shared/foo",
-		cache:      "none",
+		cache:      "never",
 	}
 
-	expected := "--syslog --cache=none --shared-dir=/run/kata-shared/foo --fd=123"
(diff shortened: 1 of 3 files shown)


jepio (approved):

src/runtime/virtcontainers/virtiofsd_test.go
```diff
@@ -130,12 +130,7 @@ func TestValid(t *testing.T) {
 		{"source is not available", func(v *virtiofsd) {
 			v.sourcePath = "/foo/bar"
 		}, errVirtiofsdSourceNotAvailable, nil},
-		{"replace cache mode none by never", func(v *virtiofsd) {
-			v.cache = "none"
-		}, nil, func(name string, a *assert.Assertions, v *virtiofsd) {
-			a.Equal("never", v.cache, "test case %+s, cache mode none should be replaced by never", name)
-		}},
-		{"invald cache mode: replace none by never", func(v *virtiofsd) {
+		{"invald cache mode", func(v *virtiofsd) {
```
stevenhorsman, line 133: `invald` -> `invalid` (which I know was there before, but worth fixing in this.

stevenhorsman (approved):
LGTM. Thanks

src/runtime/virtcontainers/virtiofsd_test.go
gkurz: Good catch. Fixed. Thanks @stevenhorsman !

amshinde (approved):

PR #7868, opened 2023-09-07
ci: k8s: Second round of fix-ups with the devmapper CI

fidencio:
This time we'll actually be able to run the tests, hopefully, and this PR will include all the needed fixes to get them green.

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -118,7 +118,7 @@ function deploy_kata() {
     echo "::endgroup::"
 
     kubectl apply -f "${tools_dir}/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml"
-    if [ "${platform}" = "tdx" ]; then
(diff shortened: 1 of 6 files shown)


stevenhorsman (approved):
It would be nice to have some of the arbitrary sleeps replaced by checking if things are ready to avoid potential flakiness, but I think this is fine to get it up and working and we can potentially look to improve that later (as long as the tests pass). Thanks!

jepio (approved):

tests/integration/kubernetes/gha-run.sh
```diff
@@ -141,10 +144,31 @@ function deploy_kata() {
 }
 
 function deploy_k3s() {
-	curl -sfL https://get.k3s.io | sh -
+	curl -sfL https://get.k3s.io | sh -s - --write-kubeconfig-mode 644
 
 	# This is an arbitrary value that came up from local tests
-	wait 240s
+	sleep 120s
```
jepio, line 150: this one could be a loop trying to `kubectl get nodes`

Issue #7886, opened 2023-09-11
runtime-rs: hypervisor: Remove unecessary kernel boot options


jodh-intel:
Remove the `earlyprintk=` and `initcall_debug` from the runtime-rs's `hypervisor` crate default kernel boot options. They are not required by default and `earlyprintk=` can degrade the VM boot speed significantly.

PR #7887, opened 2023-09-11
runtime-rs: hypervisor: Remove debug kernel options

jodh-intel:
Removed the following kernel command line options:

- `earlyprintk=ttyS0`
- `initcall_debug`

Both these options are only useful when debugging a guest kernel failure
which is not a common occurrence.

Further, the `earlyprintk=` option can have a large negative performance
impact (it can increase the VM boot time significantly).

If the user wishes to use either of these options, they can add them to the
`kernel_params=` setting in the Kata configuration file's hypervisor
stanza.

Fixes: #7886.


fidencio:
@jodh-intel, this looks good to me, but we need to ensure that users will actually have a way to add this to the command-line when, for instance, using kata-deploy.

In order to do so, I'd like to ask you to also follow what I've done here: https://github.com/kata-containers/kata-containers/pull/7854/files

BTW, does this also need changes on the go runtime?  I do think so.

fidencio:
> BTW, does this also need changes on the go runtime? I do think so.

No, I'm wrong, that's not the case. :-)

jodh-intel:
@fidencio - I put details of how users can re-add those options (if required) in the commits.

fidencio:
> @fidencio - I put details of how users can re-add those options (if required) in the commits.

I saw that, thanks for doing so.
What I'm saying is that if a user deploys kata-containers using kata-deploy they simply cannot go there and edit the files, they should pass it via annotations, and that specific annotation is blocked right now.

Thus, I'm asking you to also add that annotation to the list of allowed annotations, as one for virtio_fs_extra_args (for the go runtime) by https://github.com/kata-containers/kata-containers/pull/7854/files.

And here I'm considering we do have such a list for the runtime-rs, if we don't, then just let me know and ignore the message.

fidencio (approved):
lgtm, thanks @jodh-intel!

src/runtime-rs/crates/runtimes/virt_container/Cargo.toml
```diff
@@ -20,15 +20,16 @@ serde_derive = "1.0.27"
...
+tokio = { version = "1.8.0" }
 toml = "0.4.2"
 url = "2.1.1"
 async-std = "1.12.0"
-tracing = "0.1.36"
 
 agent = { path = "../../agent" }
 common = { path = "../common" }
-hypervisor = { path = "../../hypervisor" }
+# BUG: FIXME: TESTING:
```
studychao, line 30: Why is this introduced ?

jodh-intel: Good catch @studychao - fixed! :smile:

Issue #7888, opened 2023-09-11
Dragonball builds aren't deterministic


alex-matei:
The dragonball binary is missing the `Cargo.lock` file. This leads to builds that aren't reproducible when a dependency is updated and can cause problems if that dependency doesn't correctly update its version based on its own dependencies.

Recently nydus-rafs, nydus-storage and nydus-utils and vm-memory got new releases:
```
nydus-rafs: 0.3.2
nydus-storage: 0.6.4
nydus-utils: 0.4.3
vm-memory: 0.10.0
```

dbs_virtio_devices has the following dependencies:
```
nydus-rafs: 0.3.1 -> upgradeable to 0.3.2, minor semver change
nydus-storage: 0.6.3 -> upgradeable to 0.6.4, minor semver change
vm-memory: 0.9.0 -> not upgradeable to 0.10.0 because of how dependency resolution works
```

The new version in `nydus-storage` changed its dependency for `vm-memory` from `0.9` to `0.10`. `0.9` semver is not compatible with `0.10`, that means that this was a breaking change. 

We got into a situation (for stable-3.2 branch) where vm-memory was present in the dependency graph with both versions: 0.9.0 because of `dbs_virtio_devices` and 0.10.0 because of `nydus-storage`. 

All the packages that depend on `vm-memory` should have incremented their version in a way to make it incompatible with their previous versions. For example for `nydus-storage`, that means going from 0.6.3 to 0.7.0.

The general solution is to add Cargo.lock so that we can ensure stable branches can still be compiled in the future, even if the cargo packages get updated without taking this issue into account.

PR #7889, opened 2023-09-11
dragonball: fix for non-deterministic builds

alex-matei:
Fix for non-deterministic builds

Fixes: #7888


fidencio (approved):
lgtm, thanks @alex-matei!

studychao:
Hi @alex-matei , in runtime-rs's built-in design, Dragonball is a Rust lib for runtime-rs and it doesn't have standalone binary. 
In Rust, the lib itself should not have Cargo.lock file, because it's up to the projects who use libs to define the Cargo lock. In this way, it is runtime-rs 's responsibility to keep the Cargo.lock.

So I'm curious why to introduce Cargo.lock in this PR.

src/dragonball/.gitignore
```diff
@@ -1,3 +1,2 @@
 target
-Cargo.lock
```
studychao, line 2: Ideally, we don't need lock file for a Rust lib.

studychao (changes requested):

alex-matei:
@studychao Dragonball is defined as a standalone component in the Makefile https://github.com/kata-containers/kata-containers/blob/46e73cf7a230f861feb4565d0d5d50130e25236f/Makefile#L11 . When running `make check`, which the `static-checks` job runs on each PR, the target `check` and its dependency, `build`, will run.

studychao:
@alex-matei @fidencio ok, that makes sense.

Ideally, Dragonball is a lib to runtime-rs and it should not have a Cargo.lock itself. But we surely need to keep Dragonball working when we only want to build Dragonball alone. Introduce a Cargo.lock file could help with that.

LGTM. Thanks!

studychao (approved):

GabyCT (approved):

PR #7811, opened 2023-08-31
versions: Bump rust version

stevenhorsman:
Bump rust to 1.72.0 to test what extra warnings/issues we get

Fixes: #7902

PR #7893, opened 2023-09-11
Bump rust 1.70

stevenhorsman:


stevenhorsman:
I decided to try going straight for 1.72 in #7811, so this isn't needed now

PR #7896, opened 2023-09-11
ci: kata-deploy: Enable all k8s flavours that we support

fidencio:
Basically, let's also test on rke2 and k0s as well.

### DIFF
--- a/.github/workflows/run-kata-deploy-tests-on-aks.yaml
+++ b/.github/workflows/run-kata-deploy-tests-on-aks.yaml
@@ -44,6 +44,7 @@ jobs:
       GH_PR_NUMBER: ${{ inputs.pr-number }}
       KATA_HOST_OS: ${{ matrix.host_os }}
       KATA_HYPERVISOR: ${{ matrix.vmm }}
+      KUBERNETES: "vanilla"
       USING_NFD: "false"
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/run-kata-deploy-tests-on-garm.yaml
+++ b/.github/workflows/run-kata-deploy-tests-on-garm.yaml
@@ -1,0 +1,65 @@
+name: CI | Run kata-deploy tests on GARM
+on:
+  workflow_call:
+    inputs:
+      registry:
+        required: true
+        type: string
+      repo:
+        required: true
+        type: string
+      tag:
+        required: true
+        type: string
+      pr-number:
+        required: true
+        type: string
+      commit-hash:
+        required: false
+        type: string
+      target-branch:
+        required: false
+        type: string
+        default: ""
+
+jobs:
+  run-kata-deploy-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        vmm:
+          - clh
+          - qemu
+        k8s:
+          - k0s
+          - k3s
+          - rke2
+    runs-on: garm-ubuntu-2004-small
+    env:
+      DOCKER_REGISTRY: ${{ inputs.registry }}
+      DOCKER_REPO: ${{ inputs.repo }}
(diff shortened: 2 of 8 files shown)


.github/workflows/run-kata-deploy-tests-on-aks.yaml
```diff
@@ -44,6 +44,7 @@ jobs:
       GH_PR_NUMBER: ${{ inputs.pr-number }}
       KATA_HOST_OS: ${{ matrix.host_os }}
       KATA_HYPERVISOR: ${{ matrix.vmm }}
+      KUBERNETES: "vanilla"
```
wainersm, line 47: silly question: why vanilla and not aks?

.github/workflows/run-kata-deploy-tests-on-garm.yaml
```diff
@@ -0,0 +1,64 @@
...
+      DOCKER_REGISTRY: ${{ inputs.registry }}
+      DOCKER_REPO: ${{ inputs.repo }}
+      DOCKER_TAG: ${{ inputs.tag }}
+      PR_NUMBER: ${{ inputs.pr-number }}
+      KATA_HYPERVISOR: ${{ matrix.vmm }}
+      KUBERNETES: ${{ matrix.k8s }}
+      USING_NFD: "false"
+    steps:
+      - uses: actions/checkout@v3
+        with:
```
wainersm, line 48: don't you need `fetch-depth: 0` to checkout on an arbitrary `ref`?

.github/workflows/run-kata-deploy-tests-on-garm.yaml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          ./tests/git-helper.sh "rebase-atop-of-the-latest-target-branch"
+        env:
+          TARGET_BRANCH: ${{ inputs.target-branch }}
+
+      - name: Deploy ${{ matrix.k8s }}
+        run:  bash tests/functional/kata-deploy/gha-run.sh deploy-k8s
+
+      - name: Install `bats`
+        run: bash bash tests/functional/kata-deploy/gha-run.sh install-bats
```
wainersm, line 61: s/bash bash/bash/

.github/workflows/run-kata-deploy-tests-on-aks.yaml
fidencio: Because this will be also used for self-deployed k8s on baremetal nodes.

tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -74,9 +78,15 @@ setup() {
 teardown() {
 	kubectl get runtimeclasses -o name | grep -v "kata-mshv-vm-isolation"
 
-	if [ "${KUBERNETES}" = "k3s" ]; then
+	if [ "${KUBERNETES}" = "k0s" ]; then
+		deploy_spec="-k "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/overlays/k0s""
```
gkurz, line 82: The quoting is bogus here. The script will break if `${repo_root_dir}` contains a space. The same error already exists in the other cases. Not necessarily asking to fix them all here but at least not to propagate the bogus construct.

gkurz (commented):
Apart from the quoting issue, the rest LGTM on the form.

tests/functional/kata-deploy/kata-deploy.bats
fidencio: Fixed, thanks!

gkurz (approved):
LGTM. Thanks @fidencio !

wainersm (approved):
Thanks again @fidencio !

Issue #7910, opened 2023-09-12
ci: Add a smoke test to ensure we have Kata Containers working with docker


fidencio:
We don't need anything sophisticated here, at least not to start with, but we do want to make sure that Kata Containers + docker can at least run a basic containerd.

Issue #7912, opened 2023-09-12
runtime-rs: Using Kata Containers + docker is resulting in the process hanging


fidencio:
On a Ubuntu 23.04 machine, I've installed docker according to the following steps:
```sh
info "Installing the dependencies needed for running the docker smoke test"

# Add Docker's official GPG key:
sudo apt-get update
sudo apt-get -y install ca-certificates curl gnupg
sudo install -m 0755 -d /etc/apt/keyrings
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
sudo chmod a+r /etc/apt/keyrings/docker.gpg

# Add the repository to Apt sources:
echo \
        "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
        "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
        sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
sudo apt-get update

sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
```

Then I've used kata-containers from `main`, specifically af29eeb8b15b185c54f3750a50f3d33477ece4ce.

The containerd-shim-kata-v2 binary and the configuration are pointing to the runtime-rs ones:
```
ubuntu@docker:~$ ls -lha /usr/local/bin/containerd-shim-kata-v2 /opt/kata/share/defaults/kata-containers/configuration.toml
lrwxrwxrwx 1 root root 70 Sep 12 07:13 /opt/kata/share/defaults/kata-containers/configuration.toml -> /opt/kata/share/defaults/kata-containers/configuration-dragonball.toml
lrwxrwxrwx 1 root root 48 Sep 12 07:13 /usr/local/bin/containerd-shim-kata-v2 -> /opt/kata/runtime-rs/bin/containerd-shim-kata-v2
```

And when I run a container the process hangs after showing the expected output:
```
ubuntu@docker:~$ sudo docker run --rm --runtime io.containerd.kata.v2 busybox date
Tue Sep 12 07:14:41 UTC 2023
```

When using QEMU / Cloud Hypervisor, with the go runtime, it works as expected

studychao:
Since this problem may relate to OCI hook, @YushuoEdge will help take a look

JaewonHur:
[...]

PR #7913, opened 2023-09-12
ci: Add a very basic docker sanity test

fidencio:
Let's add a very basic sanity test to check that we can spawn a containers using docker + Kata Containers.

This will ensure that, at least, we don't regress to the point where this feature doesn't work at all.

For now we're running this test against Cloud Hypervisor and QEMU only, due to an already reported issue with dragonball:
https://github.com/kata-containers/kata-containers/issues/7912

In the future, we should also test all the VMMs with devmapper, but that's for a follow-up PR after this test is working as expected.

Fixes: #7910


gkurz:
First naive question that comes to mind : why is this considered a `functional` test ? It seems that testing that kata works well with docker rather falls into the `integration` category, like the cri-containerd tests for example.

Cc CI expert @wainersm for insights :smile:

stevenhorsman (approved):
LGTM and it's great to have a basic docker test :) Thank you!

fidencio:
> It seems that testing that kata works well with docker rather falls into the `integration` category, like the cri-containerd tests for example.

I agree and I have changed that, thanks!

fidencio:
Adding the `force-skip-ci`  label as the jenkins tests will not test anything here.

/test

tests/integration/docker/gha-run.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+kata_tarball_dir="${2:-kata-artifacts}"
+docker_dir="$(dirname "$(readlink -f "$0")")"
```
gkurz, line 13: Cosmetic: trailing white space.

.github/workflows/run-docker-tests-on-garm.yaml
```diff
@@ -0,0 +1,56 @@
...
+      target-branch:
+        required: false
+        type: string
+        default: ""
+
+jobs:
+  run-docker-tests:
+    strategy:
+      # We can set this to true whenever we're 100% sure that
+      # the all the tests are not flaky, otherwise we'll fail
```
gkurz, line 20: This sentence has a syntax problem "the all the tests". Since this comment appears in other YAML, this can be fixed globally I guess.

gkurz (approved):
I'm approving as I cannot spot anything that looks obviously wrong. Thanks @fidencio !

.github/workflows/run-docker-tests-on-garm.yaml
fidencio: Fixed now, will fix globally later, thanks!

tests/integration/docker/gha-run.sh
fidencio: Fixes, thanks!

PR #7928, opened 2023-09-13
Some sanity check on the Option value before unwrap on them

lifupan:
There's some cases where the sanity check missed and caused the shim process panic, thus we should add the proper
sanity check for those cases such as wrap on the None.


fidencio (approved):
lgtm,thanks @lifupan!

As this PR is in a state where new tests were added which are not present when this PR was raised, but the older tests ran, I'd say *DO NOT* rebase on main.

Let's get a review here, and bypass the checks as an extraordinary situation.

quanweiZhou (approved):
LGTM

PR #7929, opened 2023-09-13
ci: docker: nerdctl: Switch to tcp port 80 ping

fidencio:
TIL that the Azure VMs we use are created without an explicit outbund
connectivity defined.

This leads us to issues using `ping ...` as part of our tests, and when
consulting Jeremi Piotrowski about the issue he pointed me out to two
interesting links:
* https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access
* https://learn.microsoft.com/en-us/archive/blogs/mast/use-port-pings-instead-of-icmp-to-test-azure-vm-connectivity

For your own sanity, do not read the comments, after all this is
internet. :-)

Anyways, the suggestion is to use nping instead, which is provided by
the nmap package, so we can explicitly switch to using the tcp port 80
for the ping.  With this in mind, I'm switching the image we use for the
test and using one that provided nping as a possible entry point, and
from now on (this part of) the tests should work.

Fixes: #7910 #7911


fidencio:
A third commit was added here:
```
[ci: nerdctl: Create the containerd config](https://github.com/kata-containers/kata-containers/pull/7929/commits/f6bc05d5055df46d71977fe871ce6751224f51fb) 

Otherwise we'll fail to configure kata-containers in the `install-kata`
step.

This is mostly needed because the nerdctl-full tarball doesn't provide a
contaienrd configuration, just the binary, as contaienrd does not
actually require a configuration file to run with the default config.
```

This is needed in order to get the nerdctl tests tor run.

fidencio:
Added the `force-skip-ci` label so we don't spend time running the jenkins CI, which is not relevant to this PR.

fidencio:
```
ci: docker: nerdtl: Use io.containerd.kata-${KATA_HYPERVISOR}.io

This will ensure that we're calling the correct binary for the
hypervisor.
```

A fourth patch was needed in order to properly use the hypervisor specific runtime class.

jepio (approved):

GabyCT (approved):

Issue #6264, opened 2023-02-10
runtime-rs: ch: Enable feature by default


jodh-intel:
Once the remaining tasks on #6263 have landed, we need to [enable the `cloud-hypervisor` feature](https://github.com/kata-containers/kata-containers/pull/6201#issuecomment-1413690515) by default.

jodh-intel:
We'll need to either resolve the runtime-rs config changes, or provide a new config file for CH with runtime-rs.

See #5954 for further details.

PR #7944, opened 2023-09-14
runtime-rs: ch: Enable Intel TDX

jodh-intel:
Allow Cloud Hypervisor to create a confidential guest (a TD or "Trust Domain") rather than a VM (Virtual Machine) on Intel systems that provide TDX functionality.

> **Note:**
>
> At least currently, when built with the `tdx` feature, Cloud Hypervisor
> cannot create a standard VM on a TDX capable system: it can only create
> a TD. This implies that on TDX capable systems, the Kata Configuration
> option `confidential_guest=` must be set to `true`. If it is not, Kata
> will detect this and display the following error:
>
> ```
> TDX guest protection available and must be used with Cloud Hypervisor (set 'confidential_guest=true')
> ```

Fixes: #6448.

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/ch.rs
+++ b/src/libs/kata-types/src/config/hypervisor/ch.rs
@@ -13,7 +13,6 @@ use super::{default, register_hypervisor_plugin};
 use crate::config::default::MAX_CH_VCPUS;
 use crate::config::default::MIN_CH_MEMORY_SIZE_MB;
 
-use crate::config::hypervisor::VIRTIO_BLK_MMIO;
 use crate::config::{ConfigPlugin, TomlConfig};
 use crate::{eother, resolve_path, validate_path};
 
@@ -101,12 +100,6 @@ impl ConfigPlugin for CloudHypervisorConfig {
             }
             if !ch.valid_jailer_paths.is_empty() {
                 return Err(eother!("Valid CH jailer path list should be empty"));
-            }
-
-            if !ch.blockdev_info.disable_block_device_use
-                && ch.blockdev_info.block_device_driver == VIRTIO_BLK_MMIO
-            {
-                return Err(eother!("CH doesn't support virtio-blk-mmio"));
             }
 
             if ch.boot_info.kernel.is_empty() {
--- a/src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
+++ b/src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
@@ -6,12 +6,17 @@ use crate::net_util::MAC_ADDR_LEN;
 use crate::NamedHypervisorConfig;
 use crate::VmConfig;
 use crate::{
-    ConsoleConfig, ConsoleOutputMode, CpuFeatures, CpuTopology, CpusConfig, DiskConfig, MacAddr,
-    MemoryConfig, PayloadConfig, PlatformConfig, PmemConfig, RngConfig, VsockConfig,
+    guest_protection_is_tdx, ConsoleConfig, ConsoleOutputMode, CpuFeatures, CpuTopology,
+    CpusConfig, DiskConfig, MacAddr, MemoryConfig, PayloadConfig, PlatformConfig, PmemConfig,
+    RngConfig, VsockConfig,
 };
 use anyhow::{anyhow, Context, Result};
+use kata_sys_util::protection::GuestProtection;
 use kata_types::config::default::DEFAULT_CH_ENTROPY_SOURCE;
-use kata_types::config::hypervisor::{CpuInfo, MachineInfo, MemoryInfo};
+use kata_types::config::hypervisor::Hypervisor as HypervisorConfig;
+use kata_types::config::hypervisor::{
+    CpuInfo, MachineInfo, MemoryInfo, VIRTIO_BLK_MMIO, VIRTIO_BLK_PCI,
+};
 use kata_types::config::BootInfo;
 use std::convert::TryFrom;
 use std::fmt::Display;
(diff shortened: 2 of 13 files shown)


jodh-intel:
Note that this PR adds new functionality to the currently disabled `cloud-hypervisor` runtime-rs build feature.

See #6264 for further details.

src/libs/kata-types/src/config/hypervisor/ch.rs
```diff
@@ -103,12 +102,6 @@ impl ConfigPlugin for CloudHypervisorConfig {
                 return Err(eother!("Valid CH jailer path list should be empty"));
             }
 
-            if !ch.blockdev_info.disable_block_device_use
-                && ch.blockdev_info.block_device_driver == VIRTIO_BLK_MMIO
-            {
-                return Err(eother!("CH doesn't support virtio-blk-mmio"));
```
cmaf, line 109: @jodh-intel was this support recently added in CH, or was there a reason it was not previously supported in runtime-rs?

jodh-intel: @cmaf - Adding that check originally was an oversight, hence I zapped it.

amshinde: Could this go in a separate commit providing an explanation for it in the commit message?

src/runtime-rs/crates/hypervisor/ch-config/src/ch_api.rs
```diff
@@ -122,3 +122,21 @@ pub async fn cloud_hypervisor_vm_fs_add(
     })
     .await?
 }
+
+pub async fn cloud_hypervisor_vm_disk_add(
```
amshinde, line 126: I dont see this function being called at all. Let me know if I missed it. I also see there is an existing function `cloud_hypervisor_vm_blockdev_add` that implements the same functionality. I think you ended up using that function later in this PR as well. Maybe this function is not required?

src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
```diff
@@ -481,10 +599,29 @@ where
     Ok(MacAddr { bytes })
 }
 
+// Determine if the specified Share Fs (virtiofs mount) can be safely ignored.
+//
+// Normally, the container rootfs is passed as a virtiofs device.
+// However, if the user requested the container rootfs be passed
+// as a block device (for example if using containerd with the
+// devmapper snapshotter), runtime-rs will pass the container
+// rootfs _twice_: once as a virtiofs device and once as a block
```
amshinde, line 612: If in case of devicemapper, the rootfs is being passed twice, this needs to be fixed at some point. I think this needs to be handled in the device manager code. Do we have an issue tracking this?

src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs
```diff
@@ -481,10 +599,29 @@ where
...
+// device. Since TDX does not permit the use of virtiofs, ignore
+// the virtiofs request as we'll actually use the block device
+// request instead.
+//
+// See: https://github.com/kata-containers/kata-containers/issues/7125
+//
+pub fn can_ignore_share_fs(mount_tag: &str, guest_protection_to_use: GuestProtection) -> bool {
+    mount_tag == MOUNT_GUEST_TAG && guest_protection_is_tdx(guest_protection_to_use)
```
amshinde, line 620: I am good with using the `guest_protection_is_tdx` check for now. But shouldnt this be more of a devicemapper check like `is_storge_driver_devicemapper` to make sure this works with devicemapper as well without the confidential case .
If so, maybe track this as well as an issue?

amshinde:
@jodh-intel I did an initial review, will come back to this later again. I have left some comments.
Also, do you think you can split this commit into separate commits so it is easier to digest and understand the reasoning behind?

jodh-intel:
@amshinde, @cmaf - branch updated! ;)

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -25,17 +25,44 @@ const DEFAULT_DISK_QUEUE_SIZE: u16 = 1024;
...
+            //   DeviceType::Block and this method is called *after* the VM
+            //   has started so the device does not need to be added to the
+            //   pending list.
+            //
+            // - The VM rootfs is handled without waiting for calls to this
+            //   method as the file in question (image= or initrd=) is available
+            //   from HypervisorConfig.BootInfo.{image,initrd}
+            //   (see 'convert.rs').
+            //
+            if let DeviceType::ShareFs(_) = device {
```
amshinde, line 49: Network device is also considered a sandbox resource and needs to be added before the VM starts running. This change may break that case?

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -25,17 +25,44 @@ const DEFAULT_DISK_QUEUE_SIZE: u16 = 1024;
...
+                let mut devices: Vec<DeviceType> =
+                    if let Some(devices) = self.pending_devices.take() {
+                        devices
+                    } else {
+                        vec![]
```
amshinde, line 54: Why are we returning an empty vector here. This will get rid of any pending devices. I am also considering a case we have with golang runtime that allows for certain device passthrough devices to be coldplugged instead of the usual hotplug route.

amshinde:
@jodh-intel Can you rebase this PR?

cmaf (approved):

src/libs/kata-types/src/config/hypervisor/ch.rs
jodh-intel: Done.

src/runtime-rs/crates/hypervisor/ch-config/src/ch_api.rs
jodh-intel: Fixed.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -524,7 +605,18 @@ impl CloudHypervisorInner {
...
         let mut caps = Capabilities::default();
-        caps.set(CapabilityBits::FsSharingSupport);
+
+        let flags = if guest_protection_is_tdx(self.guest_protection_to_use.clone()) {
+            // TDX does not permit the use of virtio-fs.
+            CapabilityBits::BlockDeviceSupport | CapabilityBits::BlockDeviceHotplugSupport
+        } else {
+            CapabilityBits::BlockDeviceSupport
+                | CapabilityBits::BlockDeviceHotplugSupport
+                | CapabilityBits::FsSharingSupport
```
cmaf, line 615: What effect does adding the additional `CapabilityBits` have for the non-TDX case - was something broken prior to this?

cmaf (approved):
@jodh-intel approving with one question.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -26,7 +26,37 @@ const DEFAULT_DISK_QUEUE_SIZE: u16 = 1024;
...
+            // If the VM is not running, add the device to the pending list to
+            // be handled later.
+            //
+            // Note that the only device types considered are DeviceType::ShareFs
+            // and DeviceType::Network since:
+            //
+            // - ShareFs (virtiofsd) is only needed in an non-DM and non-TDX scenario
+            //   for the container rootfs.
+            //
+            // - For all other scenarios, the container rootfs is handled by a
```
amshinde, line 40: I think we may need to handle rootfs as block device as well, as I am sure we are supporting that case with the golang runtime. This comment will need to be changed to reflect this, of course that could so in as a future change. 
But something to note here.

amshinde (approved):

jodh-intel:
Ugh - the TDX kata deploy test is failing with:

```
# error: must build at directory: not a valid directory: evalsymlink failure on '"/home/github-runner/actions-runner/_work/kata-containers/kata-containers/tests/functional/kata-deploy/../../..//tools/packaging/kata-deploy/kata-cleanup/overlays/k3s"' : lstat /home/github-runner/actions-runner/_work/kata-containers/kata-containers/tests/functional/kata-deploy/": no such file or directory
```

jodh-intel:
This PR has been tested twice by the CI. The only non-wobbly failing CI was the TDX kata deploy test. However, that seems unrelated since (a) this PR does not enable the TDX feature; and (b) the failure related to a missing sym-link which suggests a CI script issue.

Merging...

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
jodh-intel: Fixed.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
jodh-intel: We just need a vector to store as the `pending_devices` list so this code creates one if there are no current pending devices.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
jodh-intel: Not broken, but we only handled the default (non TDX, non-DM) cases.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
jodh-intel: Container rootfs block devices are handled "post VM boot" so that filters through to us just calling `self.handle_add_device(device)` immediately (aka nothing is added to the pending list).

For a VM rootfs block device, we already have all the info we need so although runtime-rs does call `add_device()` pre-boot, we don't actually need that entry as we just use `PmemConfig::try_from(&boot_info)` currently.

jodh-intel:
btw, wobbly kata-deploy TDX test failure recorded on https://github.com/kata-containers/kata-containers/issues/8064.

PR #7946, opened 2023-09-14
ci: install ORAS when needed

fidencio:
ORAS (OCI Registry As Storage) will be used to cache the artefacts we build, in a way to reduce the usage of jenkins and lower the costs on Azure, as this can be done from the GHA, using the free machines.

For now the only thing we're doing is installing ORAS whenever it's needed, mostly targetting the baremetal machines (s390x and arm64), but also doing the same for x86_64 for the sake of consistency.

Fixes: #7834 -- with the caveat that this is only the first part of the
                of this work.

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -76,6 +76,10 @@ jobs:
         env:
           TARGET_BRANCH: ${{ inputs.target-branch }}
 
+      - name: Install ORAS
+        run: |
+          ./tests/install_oras.sh
+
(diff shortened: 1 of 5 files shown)


fidencio:
With @BbolroC's help, we were able to verify that `./tests/install_oras.sh` works on s390x, arm64, and x86_64.

stevenhorsman (approved):
LGTM. Thanks!

zvonkok (approved):
LGTM, one step forward to have a great CI

fidencio:
I'm closing this PR as, in the end, we'll be doing this from inside a container, and we'll not need this installed in the host.

PR #6146, opened 2023-01-28
runtime: Add hmp for qemu

zhaojizhuang:
Fixes: #6092
Signed-off-by: zhaojizhuang <571130360@qq.com>

PR #7953, opened 2023-09-14
runtime/qemu: Rework QMP/HMP support

### LINKED ISSUE #7952: Shim v2 starts QEMU with an unwanted `-monitor` control point
PR #6146 added the possibility to control QEMU with an extra HMP socket as an aid for debugging. This is great for  development or bug chasing but this raises some concerns in production :
- the extra HMP is an actual backdoor that can be used to silently temper with the VM and cause heisenbug-grade issues
- it is currently piggybacked on the `hypervisor.qemu.enable_debug` configuration setting which was historically used to _just_ control logging

The enablement of this feature should be hardened so that users cannot end up using it without knowing it.

Even if this looks like an enhancement request, let's label it as a bug because it is a regression for pre-#6146 users of `enable_debug`.

gkurz:
PR #6146 added the possibility to control QEMU with an extra HMP socket as an aid for debugging. This is great for development or bug chasing but this raises some concerns in production.

The HMP monitor allows to temper with the VM state in a variety of ways. This could be intentionally or mistakenly used to inject subtle bugs in the VM that would be extremely hard if not even impossible to debug. We definitely don't want that to be enabled by default.

The feature is currently wired to the `enable_debug` setting in the `[hypervisor.qemu]` section of the configuration file. This setting has historically been used to control "debug output" and it is used as such by some downstream users (e.g. Openshift). Forcing people to have the extra HMP backdoor at the same time is abusive and dangerous.

A new `extra_monitor_socket` is added to `[hypervisor.qemu]` to give fine control on whether the HMP socket is wanted or not. This setting is still gated by `enable_debug = true` to make it clear it is for debug only. The default is to not have the HMP socket though. This isn't backward compatible with #6416 but it is for the sake of "better safe than sorry".

An extra monitor socket makes the QEMU instance untrusted. A warning is thus logged to the journal when one is requested.

While here, also allow the user to choose between HMP and QMP for the extra monitor socket. Motivation is that QMP offers way more options to control or introspect the VM than HMP does. Users can also ask for pretty json formatting well suited for human reading. This will improve the debugging experience.

This feature is only made visible in the base and GPU configurations of QEMU for now.

Fixes #7952

Cc @zhaojizhuang as author of #6146 
Cc @Amulyam24 and @zvonkok who showed interest when this was discussed on slack

### DIFF
--- a/src/runtime/pkg/govmm/qemu/qemu.go
+++ b/src/runtime/pkg/govmm/qemu/qemu.go
@@ -2471,14 +2471,28 @@ const (
 	Unix QMPSocketType = "unix"
 )
 
-// QMPSocket represents a qemu QMP socket configuration.
+// MonitorProtocol tells what protocol is used on a QMPSocket
+type MonitorProtocol string
+
+const (
+	// Socket using a human-friendly text-based protocol.
+	Hmp MonitorProtocol = "hmp"
+
+	// Socket using a richer json-based protocol.
+	Qmp MonitorProtocol = "qmp"
+
+	// Same as Qmp with pretty json formatting.
+	QmpPretty MonitorProtocol = "qmp-pretty"
+)
+
+// QMPSocket represents a qemu QMP or HMP socket configuration.
 // nolint: govet
 type QMPSocket struct {
 	// Type is the socket type (e.g. "unix").
 	Type QMPSocketType
 
-	// Human Monitor Interface (HMP) (true for HMP, false for QMP, default false)
-	IsHmp bool
+	// Protocol is the protocol to be used on the socket.
+	Protocol MonitorProtocol
 
 	// QMP listener file descriptor to be passed to qemu
 	FD *os.File
@@ -2501,6 +2515,10 @@ func (qmp QMPSocket) Valid() bool {
 	}
 
 	if qmp.Type != Unix {
+		return false
+	}
+
+	if qmp.Protocol != Hmp && qmp.Protocol != Qmp && qmp.Protocol != QmpPretty {
 		return false
 	}
 
@@ -2855,10 +2873,11 @@ func (config *Config) appendQMPSockets() {
 			}
 		}
 
-		if q.IsHmp {
+		switch q.Protocol {
+		case Hmp:
 			config.qemuParams = append(config.qemuParams, "-monitor")
-		} else {
-			config.qemuParams = append(config.qemuParams, "-qmp")
+		default:
+			config.qemuParams = append(config.qemuParams, fmt.Sprintf("-%s", q.Protocol))
 		}
 
 		config.qemuParams = append(config.qemuParams, strings.Join(qmpParams, ","))
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -119,11 +119,11 @@ type qemu struct {
 }
 
 const (
-	consoleSocket = "console.sock"
-	qmpSocket     = "qmp.sock"
-	hmpSocket     = "hmp.sock"
-	vhostFSSocket = "vhost-fs.sock"
-	nydusdAPISock = "nydusd-api.sock"
+	consoleSocket      = "console.sock"
+	qmpSocket          = "qmp.sock"
+	extraMonitorSocket = "extra-monitor.sock"
+	vhostFSSocket      = "vhost-fs.sock"
+	nydusdAPISock      = "nydusd-api.sock"
 
 	// memory dump format will be set to elf
 	memoryDumpFormat = "elf"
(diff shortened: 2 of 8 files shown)


src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -2471,14 +2471,26 @@ const (
...
 // nolint: govet
 type QMPSocket struct {
 	// Type is the socket type (e.g. "unix").
 	Type QMPSocketType
 
-	// Human Monitor Interface (HMP) (true for HMP, false for QMP, default false)
-	IsHmp bool
+	// Protocol is the protocol to be used on the socket (either Hmp, Qmp or
+	// QmpPretty).
+	Protocol string
```
pmores, line 2493: You could consider using the widely used pattern in govmm that's also used here for `QMPSocketType` - declare something like `type QMPSocketProtocol string` and make the protocol consts and this new `Protocol` member be of type `QMPSocketProtocol`.  This would also make the explanation in the comment unnecessary.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -2471,14 +2471,26 @@ const (
...
+	// Socket using a richer json-based protocol.
+	Qmp string = "qmp"
+
+	// Same as Qmp with pretty json formatting.
+	QmpPretty string = "qmp-pretty"
+)
+
+// QMPSocket represents a qemu QMP or HMP socket configuration.
 // nolint: govet
 type QMPSocket struct {
```
pmores, line 2487: Just as something to consider, now that `QMPSocket` is not necessarily QMP, a more general name, perhaps `ControlSocket` (?), might be more appropriate.

src/runtime/virtcontainers/qemu.go
```diff
@@ -119,11 +119,11 @@ type qemu struct {
...
 
 const (
-	consoleSocket = "console.sock"
-	qmpSocket     = "qmp.sock"
-	hmpSocket     = "hmp.sock"
-	vhostFSSocket = "vhost-fs.sock"
-	nydusdAPISock = "nydusd-api.sock"
+	consoleSocket      = "console.sock"
+	qmpSocket          = "qmp.sock"
+	extraMonitorSocket = "extra-monitor.sock"
```
pmores, line 124: Just as a thought, do we need both `qmpSocket` and `extraMonitorSocket`? I understand that the latter is just a rename of an existing `hmpSocket` constant, however if I understand correctly, there will only ever be at most one monitor socket per a kata VM. If so, then we could either want to bake the name of the protocol used into the name of the socket file (which I could see value in, for debugging), in which case we'd need separate constants for `hmp.sock` and `qmp-pretty.sock`, or we don't want to do that, in which case it would make sense to use a single name (maybe `monitor.sock`?) for all protocol variants.

The currently proposed solution kind of doesn't want to make up its mind, using an explicit `qmp.sock` name for QMP but subsuming the pretty variant of QMP and HMP under `extra-monitor.sock`.

src/runtime/pkg/govmm/qemu/qemu.go
gkurz: You're right. I thought about doing that but maybe not in this PR because the renaming will be quite noisy and I really want this fix to reach main and stable-3.2 as soon as possible.

src/runtime/pkg/govmm/qemu/qemu.go
gkurz: I'll give a try.

src/runtime/pkg/govmm/qemu/qemu.go
pmores: Sure, doing the rename in a subsequent PR would be perfectly fine.

src/runtime/virtcontainers/qemu.go
gkurz: This PR allows to add exactly one monitor socket. This socket must have a different _address_ than the primary socket and the primary socket is de facto a plain QMP session. We thus cannot `bake the name of the protocol used into the name of the socket file` since the extra socket could be plain QMP as well.
Not sure if the protocol variant really needs to be part of the extra socket name because the user already knows what he asked for in the configuration TOML... or do you have some other scenario in mind @pmores ?

pmores: Alright, so it seems I misunderstood, I apologize. If it's the case that there will (almost) always be a `qmpSocket` and independent of that there _might_ be an `extraMonitorSocket` if requested by `configuration.toml` then please ignore my previous comment.

src/runtime/pkg/govmm/qemu/qemu.go
gkurz: Done

pmores (approved):
lgtm thanks @gkurz !

fidencio:
The WIP Check and the Commit Message Check didn't start, but I will bypass that for the sake of avoid re-running the whole CI again.

I will merge this one after the required jobs pass, thanks @gkurz

fidencio (approved):

Amulyam24 (approved):
LGTM, Thanks @gkurz, this is definitely helpful!

PR #7954, opened 2023-09-14
ci: cache: Export env vars needed to use ORAS

fidencio:
We do the build of our artefacts inside a container image, and we need to expose some env vars to the container so ORAS can be used there to push the artefacts we want to cache to ghcr.io.

The env vars we're exposing are:
* ARTEFACT_REGISTRY: The registry where we're going to save the artefacts.
* ARTEFACT_REGISTRY_USERNAME: The username to log in to the registry, as ORAS does not use the same json file used by docker.
* ARTEFACT_REGISTRY_PASSWORD: The pasword to log in to the the registry, as the ORAS does not use the same json file used by docker.
* TARGET_BRANCH: The target branch, which will be part of the tag of the artefact, as we may end up caching the artefacts for both main and stable branches.

I'm opening this PR as "part 0" of the full work that can be seen in the following branch: https://github.com/fidencio/kata-containers/pull/new/topic/ci-cache-using-oras-part-1

The reason I'm doing so is that we first need to change the yaml file, which is not tested, and only then we'll be able to test the upcoming work.

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -86,6 +86,10 @@ jobs:
           KATA_ASSET: ${{ matrix.asset }}
           TAR_OUTPUT: ${{ matrix.asset }}.tar.gz
           PUSH_TO_REGISTRY: ${{ inputs.push-to-registry }}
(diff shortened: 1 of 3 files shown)


zvonkok (approved):
LGTM, thanks!!

Issue #7955, opened 2023-09-14
ci: Reduce the size of the AKS VMs


fidencio:
We do **not** need a very powerful machine for our tests, as we're not
building anything there.

The instance we switched to (Standard_D2s_v5) still has nested virt
available, as shown here[0], but has half of the amount of vCPUs /
Memory, which should be fine only for running the tests, costing us
basically half of the price[1].

[0]:
https://learn.microsoft.com/en-us/azure/virtual-machines/dv5-dsv5-series
[1]:
https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/#pricing

PR #7956, opened 2023-09-14
ci: Reduce the size of the AKS VMs

fidencio:
We do **not** need a very powerful machine for our tests, as we're not building anything there.

The instance we switched to (Standard_D2s_v5) still has nested virt available, as shown here[0], but has half of the amount of vCPUs / Memory, which should be fine only for running the tests, costing us basically half of the price[1].

[0]:
https://learn.microsoft.com/en-us/azure/virtual-machines/dv5-dsv5-series [1]:
https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/#pricing

Fixes: #7955

### DIFF
--- a/tests/gha-run-k8s-common.sh
+++ b/tests/gha-run-k8s-common.sh
@@ -43,7 +43,7 @@ function create_cluster() {
     az aks create \
         -g "${AZ_RG}" \
         -n "$(_print_cluster_name ${test_type})" \
-        -s "Standard_D4s_v5" \
+        -s "Standard_D2s_v5" \
         --node-count 1 \
         --generate-ssh-keys \
         $([ "${KATA_HOST_OS}" = "cbl-mariner" ] && echo "--os-sku AzureLinux --workload-runtime KataMshvVmIsolation")
--- a/tests/integration/kubernetes/run_kubernetes_tests.sh
+++ b/tests/integration/kubernetes/run_kubernetes_tests.sh
@@ -40,18 +40,15 @@ else
 		"k8s-liveness-probes.bats" \
 		"k8s-memory.bats" \
 		"k8s-nested-configmap-secret.bats" \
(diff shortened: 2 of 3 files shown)


sprt (approved):
Approved pending CI

sprt:
Is the configmap test failure new?

fidencio:
> Is the configmap test failure new?

It fails due to the size of the instance, which only has 2 vCPUS.
I'm working on getting exactly which and how many tests will need a bigger instance, and we could split which tests will run on which instance ... still saving quite a bunch on our side.

WDYT?

sprt:
@fidencio Great idea, thanks for this!

sprt:
> It fails due to the size of the instance, which only has 2 vCPUS.

Oh I just noticed this - not familiar with this test but is the request required? https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/runtimeclass_workloads/inotify-configmap-pod.yaml#L19

fidencio:
> > It fails due to the size of the instance, which only has 2 vCPUS.
> 
> Oh I just noticed this - not familiar with this test but is the request required? https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/runtimeclass_workloads/inotify-configmap-pod.yaml#L19

It is. :-)

fidencio:
> > > It fails due to the size of the instance, which only has 2 vCPUS.
> > 
> > 
> > Oh I just noticed this - not familiar with this test but is the request required? https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/runtimeclass_workloads/inotify-configmap-pod.yaml#L19
> 
> It is. :-)

No, it's not!!!

chavafg (approved):

GabyCT (approved):

Issue #6333, opened 2023-02-20
runtime-rs: Add Cloud Hypervisor network support


jodh-intel:
The original [CH PR](https://github.com/kata-containers/kata-containers/pull/6201) included a [hard-coded network example](https://github.com/kata-containers/kata-containers/pull/6201/files#diff-5733f4b6eceb4ea929963b18440426f00b5cb6ba5cf8d6cb0b876a6d84996b70R187-R199) with some utility functions, but the [config cleanup PR](https://github.com/kata-containers/kata-containers/pull/6332) removed the network configuration.

Add proper CH networking by setting `VmConfig.net` to a list of `NetConfig`.

Related:  #6263.

jodh-intel:
Hi @amshinde - this might be a good one for you to tal at?

jodh-intel:
Note that this issue is part of the device handling as [`Network` is a `Device`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/src/device/mod.rs#L25).

Related:

- https://github.com/kata-containers/kata-containers/issues/6690
- https://github.com/kata-containers/kata-containers/issues/6691
- https://github.com/kata-containers/kata-containers/issues/6692

PR #7966, opened 2023-09-15
runtime-rs: Add network support for cloud-hypervisor

amshinde:
This PR adds support for adding a network device before starting the cloud-hypervisor VM.

Support for adding and removing network devices is not really added to the resource manager, so supporting this for cloud-hypervisor is not scoped in this PR.

This also changes "pending_devices" for clh implementation from an Option of vector to simply a vector. This simplifies the structure a bit as we can simple iterate over the pending devices instead of having to check for a "Some" value as this is not really required.

Fixes: #6333

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
@@ -247,3 +247,44 @@ impl TryFrom<ShareFsSettings> for FsConfig {
         Ok(fs_cfg)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::Address;
+
+    #[test]
+    fn test_networkconfig_to_netconfig() {
+        let mut cfg = NetworkConfig {
+            host_dev_name: String::from("tap0"),
+            virt_iface_name: String::from("eth0"),
+            queue_size: 256,
+            queue_num: 2,
+            guest_mac: None,
+            index: 1,
+        };
+
+        let net = NetConfig::try_from(cfg.clone());
+        assert_eq!(
+            net.unwrap_err().to_string(),
+            "Missing mac address for network device"
+        );
+
+        let v: [u8; 6] = [10, 11, 128, 3, 4, 5];
+        let mac_address = Address(v);
+        cfg.guest_mac = Some(mac_address.clone());
+
+        let expected = NetConfig {
+            tap: Some(cfg.host_dev_name.clone()),
+            id: Some(cfg.virt_iface_name.clone()),
+            num_queues: cfg.queue_num,
+            queue_size: cfg.queue_size as u16,
+            mac: MacAddr { bytes: v },
+            ..Default::default()
+        };
+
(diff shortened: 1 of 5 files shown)


src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -161,32 +152,53 @@ impl CloudHypervisorInner {
...
+    pub(crate) async fn get_shared_devices(
+        &mut self,
+    ) -> Result<(Option<Vec<FsConfig>>, Option<Vec<NetConfig>>)> {
+        let mut shared_fs_devices = Vec::<FsConfig>::new();
+        let mut network_devices = Vec::<NetConfig>::new();
+
+        while let Some(dev) = self.pending_devices.pop() {
+            match dev {
+                DeviceType::ShareFs(dev) => {
+                    let settings = ShareFsSettings::new(dev.config, self.vm_path.clone());
```
fidencio, line 164: @amshinde, with the go runtime we're passing an FD to be used, not doing so will make SELinux integration way harder in the future.

Would you mind to follow what's done there?

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -161,32 +152,53 @@ impl CloudHypervisorInner {
...
+impl TryFrom<NetworkConfig> for NetConfig {
+    type Error = anyhow::Error;
 
-                        root_devices.push(fs_cfg);
-                    }
-                    _ => continue,
-                };
-            }
+    fn try_from(cfg: NetworkConfig) -> Result<Self, Self::Error> {
+        let mut net_config = NetConfig {
```
jodh-intel, line 186: - If set, should / can `NetworkConfig.virt_iface_name` be used for `net_config.id`?

amshinde: This is strictly not required, as the agent relies on the mac address solely and not on the interface name for updating the interfaces later within the VM. But no harm updating this.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -161,32 +152,53 @@ impl CloudHypervisorInner {
...
+            queue_size: cfg.queue_size as u16,
+            ..Default::default()
+        };
 
-            Ok(Some(root_devices))
+        if let Some(mac) = cfg.guest_mac {
+            net_config.mac = MacAddr { bytes: mac.0 };
         } else {
-            Ok(None)
+            return Err(anyhow!("Missing mac address for network device"));
```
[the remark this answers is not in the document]
amshinde: I have added test for checking this.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
amshinde: Sure, i'll follow up on this.

cmaf (approved):
thanks @amshinde

PR #7969, opened 2023-09-15
ci: cache: Check the sha256sum of the components & fix ovmf-sev cache usage

fidencio:
See the commits for more info, please.

This is the last bit of this series, although we'll be creating an issue to have the kernel-sev cache usage to be fixed (which is also making the sev-initrd cache to not be used).


stevenhorsman (approved):
LGTM. Thanks

Issue #7972, opened 2023-09-15
ci: Run some of the k8s tests in a bigger instance


fidencio:
Some of the k8s tests require the "normal" instance to run / succeed, usually because they use more vCPUs than what the smaller instances provide.

With this is mind, let's make sure that we run all the tests that can be run on a smaller instance using D2s_v5 machines, but then use D4s_v5 machines to run the tests which were disabled by #7956

PR #7973, opened 2023-09-15
ci: Use variable size of VMs depending on the tests running

fidencio:
Let me start with a fair warning that this commit is hard to split into different parts that could be easily tested (or not tested, just ignored) without breaking pieces.

Now, about the commit itself, as we're on the run to reduce costs related to our sponsorship on Azure, we can split the k8s tests we run in 2 simple groups:
* Tests that can be run in the smaller Azure instance (D2s_v5)
* Tests that required the normal Azure instance (D4s_v5)

With this in mind, we're now passing to the tests which type of host we're using, which allows us to select to run either one of the two types of tests, or even both in case of running the tests on a baremetal system.

Fixes: #7972

NOTE:
As GHA will **NOT** test this when running the tests for this PR, we 
need a very careful review, and people around to unbreak things (which
we know that will happen) once this commits gets merged.

PLEASE, DO NOT RUN THE CI TILL IT GETS REVIEWED.


.github/workflows/run-k8s-tests-on-garm.yaml
```diff
@@ -35,7 +35,15 @@ jobs:
...
-    runs-on: garm-ubuntu-2004-smaller
+        instance:
+          garm-ubuntu-2004
+          garm-ubuntu-2004-smaller
+        include:
+          - instance: garm-ubuntu-2004
+            instance-type: normal
+          - instance: garm-ubuntu-2004-smaller
+            instance-type: small
+    runs-on: {{ matrix.instance }}
```
jepio, line 46: are you sure we need the workflow runner to be big when the aks cluster uses a big node size? I would think that we could stick to a small runner even with bigger aks node instance, because the runner only executes remote commands.

fidencio: So, this is the "on-garm" case.  Here we're actually deploying a k8s cluster in a garm VM, as we need to setup things like the snapshotter being used.

In this case we're running the tests from inside the garm VM, which happens to also be the k8s cluster we setup, thus we need the correct size.  Does it make sense?

jepio: now i follow - ok

gkurz (approved):
With my limited knowledge, LGTM. Thanks  @fidencio !

PR #7976, opened 2023-09-15
ci: Rework static checks

fidencio:
This is a huge effort to reduce costs and to make the static checks more organised / faster / cheaper.

Please, see each commit for more detaisl.


.github/workflows/static-checks-dragonball.yaml
jepio, line 1: my understanding is that `static-checks-dragonball` are not actually static checks and require kvm

fidencio: I will get there, I promise I will get there. :-)

ChengyuZhu6:
I encountered this error :
```
INFO: Checking JSON file 'tests/integration/nydus/nydusd-config.json'
Command 'jq' not found in /home/runner/actions-runner/_work/kata-containers/kata-containers/bin, /home/runner/actions-runner/_work/_tool/go/1.19.3/x64/bin, /usr/local/sbin, /usr/local/bin, /usr/sbin, /usr/bin, /sbin, /bin, /snap/bin, /home/runner/.cargo/bin at /usr/bin/chronic line 69.
ERROR: failed to check JSON file 'tests/integration/nydus/nydusd-config.json'
```
 while executing `make static-checks` in ci tests. Should we install `jq` or just move the binary to PATH?

tests/common.bash
```diff
@@ -413,3 +413,42 @@ function install_cri_tools() {
 	sudo tar -xvf "${tarball_name}" -C /usr/local/bin
 	rm -f "${tarball_name}"
 }
+
+# Convert architecture to the name used by golang
+function arch_to_golang() {
```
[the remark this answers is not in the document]
fidencio: I'd prefer doing it in a follow-up PR, if you don't mind.

Please, mind to open an issue about this?  We can mark it as a `good-first-issue`.

fidencio:
As all the static checks are now passing, and they're passing under 10 minutes, with only one job (for testing dragonball) running on the smaller VM instance, I will add the `force-skip-ci` label as nothing here is related to jenkins.

.github/workflows/static-checks.yaml
```diff
@@ -12,84 +12,183 @@ concurrency:
...
+          ./tests/install_rust.sh
+      - name: Running `${{ matrix.command }}` for ${{ matrix.component }}
+        run: |
+          export PATH="$PATH:${HOME}/.cargo/bin"
+          cd ${{ matrix.component-path }}
+          ${{ matrix.command }}
+        env:
+          RUST_BACKTRACE: "1"
+
+  static-checks:
```
stevenhorsman, line 162: Have we lost the `force-skip-ci` functionality now? If so, was that deliberate and you think we don't need it any more?

stevenhorsman (commented):
Generally looks good, just one question to confirm

.github/workflows/static-checks.yaml
fidencio: That was deliberate.  I don't see any reason to have that, unless we have that for all the jobs, and that could be covered in a different PR, after we discuss whether this will really be needed.

stevenhorsman (approved):
LGTM. Thanks

jepio (approved):

PR #7983, opened 2023-09-15
ci: Create clusters in individual resource groups

sprt:
This makes it so that each AKS cluster is created in its own individual
resource group, rather than using the "kataCI" resource group for all
test clusters.

This is to accommodate a tool that we recently introduced in our Azure
subscription which automatically deletes resource groups after a set
amount of time, in order to keep spending under control.

The tool will automatically delete any resource group, unless it has a
tag SkipAutoDeleteTill = YYYY-MM-DD. When this tag is present, the
resource group will be retained until the specified date.

Note that I tagged all current resource groups in our subscription with
SkipAutoDeleteTill = 2043-01-01 so that we don't lose any existing
resources.

Fixes: https://github.com/kata-containers/kata-containers/issues/7982


tests/gha-run-k8s-common.sh
```diff
@@ -21,6 +20,12 @@ function _print_cluster_name() {
     echo "${test_type}-${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64"
 }
 
+function _print_rg_name() {
+    test_type="${1:-k8s}"
+
+    echo "kataCI-$(_print_cluster_name ${test_type})"
```
fidencio, line 26: As part of the cluster name, we'll have to add "small" "normal", otherwise I think we may have some clashes.

fidencio: Either in the rg name or in the cluster name, and I only noticed this after my PR got merged, sorry.

fidencio (approved):
lgtm, thanks @sprt!

stevenhorsman (approved):
Looks ok to me. Thanks!

fidencio:
As this PR is crucial for the reducing costs effort, I will *NOT* wait till the baremetal CIs finish, as everything done here is 100% specific to the GHA.

PR #7985, opened 2023-09-18
clh: arm: Use static_sandbox_resource_mgmt=true

fidencio:
Users have noticed that this is needed, as CLH does not yet implement a way to hotplug resources on aarh64.

With this patch, when building for x86_64, I can see the this is the resulting config:
```
$ ARCH=amd64 make
...

$ cat config/configuration-clh.toml | grep static_sandbox_resource_mgmt
static_sandbox_resource_mgmt=false

```

And when building for aarch64:
```
$ ARCH=arm64 make
...

$ cat config/configuration-clh.toml | grep static_sandbox_resource_mgmt
static_sandbox_resource_mgmt=true
```

Fixes: #7941


jepio (approved):

gkurz:
@fidencio shouldn't this backported to 3.2 ?

fidencio:
> @fidencio shouldn't this backported to 3.2 ?

It should!

fidencio:
I am bypassing the WIP check that got stuck and never started, instead of re-running the whole CI. :-/

Issue #7986, opened 2023-09-18
CI | GHA: create a test workflow for internal nightly jobs for s390x


BbolroC:
**Which feature do you think can be improved?**

In Jenkins, the following e2e test has been verified as a form of the nightly test. This is because there is limitation on making a test machine capable of IBM Z hardware specific features (e.g. Secure Execution, Crypto Express card and etc.) in public. 

- http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-20.04-s390x-VFIO-AP-daily/

Followed by https://github.com/kata-containers/kata-containers/issues/7930, these also have to be migrated to a GHA workflow.

PR #7987, opened 2023-09-18
tests|gha: add nightly tests for s390x

BbolroC:
This is to add a workflow for internal nightly tests for s390x in Jenkins.

This will replace the following CI job in Jenkins:

- http://jenkins.katacontainers.io/job/kata-containers-2.0-ubuntu-20.04-s390x-VFIO-AP-daily/

Fixes: #7986

### DIFF
--- a/.github/workflows/ci-nightly-s390x.yaml
+++ b/.github/workflows/ci-nightly-s390x.yaml
@@ -1,0 +1,47 @@
+on:
+  schedule:
+    - cron: '0 2 * * *'
+
+name: Nightly CI for s390x
+jobs:
+  check-internal-test-result:
+    runs-on: s390x
+    strategy:
+      fail-fast: false
+      matrix:
+        test_title:
+          - kata-vfio-ap-e2e-tests
+    steps:
+    - name: Fetch a test result for {{ matrix.test_title }}
+      run: |
+        file_name="${TEST_TITLE}-$(date +%Y-%m-%d).log"
+        /home/${USER}/script/handle_test_log.sh download $file_name
+      env:
+        TEST_TITLE: ${{ matrix.test_title }}
+
+  k8s-cri-containerd-rhel9-e2e-tests:
+    runs-on: s390x-rhel9
+    steps:
+    - name: Delete the existing files
+      run: |
+        sudo chown -R $USER:$USER $GITHUB_WORKSPACE
+        sudo rm -rf $GITHUB_WORKSPACE/*
+
+    - name: Take a pre-action for self-hosted runner
+      run: |
+        ${HOME}/script/pre_action.sh rhel9-nightly
+
+    - name: Run k8s/cri-containerd e2e tests on RHEL9
+      run: |
+        export WORKSPACE=$GITHUB_WORKSPACE
+        export GITHUB_ACTION=""
+        bash ci_crio_entry_point.sh
+      env:
+        BAREMETAL: "true"
+        REPO_OWNER: "cri-o"
+        REPO_NAME: "cri-o"
+    
+    - name: Take a post-action for self-hosted runner
+      if: always()
+      run: |
+        ${HOME}/script/post_action.sh rhel9-nightly


stevenhorsman:
Hey @BbolroC - It looks like the test code that is running here is based on `CCv0`, rather than `main`, so I wonder if this should go into the `CCv0` branch for now, so it doesn't get too confusing? 

Also, do these jobs get reflected in a badge/notification anywhere, or do we just need to remember to manually check them to see if there are problems?

BbolroC:
> Hey @BbolroC - It looks like the test code that is running here is based on `CCv0`, rather than `main`, so I wonder if this should go into the `CCv0` branch for now, so it doesn't get too confusing?

Ah, yes. You are right. `vfio-ap` is bound to `main` and the rest goes with `CCv0`. I will split the PR. Thanks.

> Also, do these jobs get reflected in a badge/notification anywhere, or do we just need to remember to manually check them to see if there are problems?

I've maintained them via the latter approach in Jenkins. But it would be great to have such a good feature for the notification in GHA. (I haven't experienced yet to configure those things for any repos.)

BbolroC:
I will suspend this PR with `WIP` until the CCv0 branch is merged into the main branch. Thanks!

BbolroC:
> Hey @BbolroC - It looks like the test code that is running here is based on `CCv0`, rather than `main`, so I wonder if this should go into the `CCv0` branch for now, so it doesn't get too confusing?

No, the test code is pulled on the internal test machines individually (main for `vfio-ap`, CCv0 for `se` and `operator`). This is only to fetch/evaluate a result.

> Also, do these jobs get reflected in a badge/notification anywhere, or do we just need to remember to manually check them to see if there are problems?

I've maintained them via the latter approach in Jenkins. But it would be great to have such a good feature for the notification in GHA. (I haven't experienced yet to configure those things for any repos.)

stevenhorsman:
> > Hey @BbolroC - It looks like the test code that is running here is based on CCv0, rather than main, so I wonder if this should go into the CCv0 branch for now, so it doesn't get too confusing?
> 
> No, the test code is pulled on the internal test machines individually (main for vfio-ap, CCv0 for se and operator). This is only to fetch/evaluate a result.

Yes, sorry if I wasn't clear - I was wondering if the reporting of `main` test results mixed with `CCv0` results would be confusing here, but if they are currently only manually reviewed and mostly by you, then hopefully it won't confuse people

stevenhorsman (approved):

BbolroC:
Ah, I got your point. I will raise another hand for the notification when the `CCv0` branch is merged onto the `main` branch. Thanks for the clarification.

BbolroC:
If @fidencio or @wainersm does not want to have the mixed test result. I will suspend this until the merging task is finished. Thanks.

GabyCT (approved):

fidencio:
I really would not like to have anything from CCv0 "leaked" into the upstream stuff.

In this part https://github.com/kata-containers/kata-containers/pull/7987/files#diff-0499c9781b08215f6d1179e9ee91c615bac834b22579bf1c7b2e3a6835904bd2R1-R3 you should be able to specify a branch where it's going to run, no?  If so, then I'd be fine on adding it and not having it reported anywhere.

But I'd like to avoid any possibility of having this running on a different branch than `CCv0`.

BbolroC:
@fidencio Thanks for the feedback. Then, I will split the PR into two for each branch, respectively like:

- vfio-ap: main
- the rest: CCv0

BbolroC:
I've removed 2 nightly jobs for CCv0. Thanks!

BbolroC:
@GabyCT @stevenhorsman I would like to update you on the latest change I have made:

- Add a nightly e2e test on RHEL 9

You can check the test result at https://github.com/BbolroC/kata-containers/actions/runs/6781260025 for the change. Thanks.

BbolroC:
@stevenhorsman We should register another self-hosted runner for RHEL before merging this PR. Thanks.

stevenhorsman:
> @stevenhorsman We should register another self-hosted runner for RHEL before merging this PR. Thanks.

Ok, let me know when you a ready and we can sync on slack to get that happening :)

Issue #7960, opened 2023-09-14
Update virtiofsd to v1.8.0


skaegi:
Release v1.8.0 of virtiofsd has a number of important fixes but one in particular allows its use on hosts that are running slightly older kernels. We are using this in production now on a 5.4.x kernel (with kata 3.1.x) and would suggest giving consideration to adopting this release for the *gulp* 3.2 release even though it's very late in the cycle.

It would help out anyone trying to use Kata and kata-deploy on the public cloud using providers managed Kubernetes solutions as they are all generally using 5.4.x as a default for the next year or so. This would also be worthwhile to consider for backport to 3.1.x to help current adopters who want to move to 3.x but are not ready to move to 3.2 right away. We've done this in our internal builds but other teams might not be in the same position

See https://github.com/kata-containers/kata-containers/issues/5157 for details

PR #7992, opened 2023-09-18
versions: Bump virtiofsd to v1.8.0

skaegi:
https://gitlab.com/virtio-fs/virtiofsd/-/releases/v1.8.0 was released two weeks ago. We have fully tested and are using this version.

Fixes: #7960

Signed-off-by: Simon Kaegi <simon.kaegi@gmail.com>

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -329,14 +329,14 @@ externals:
   virtiofsd:
     description: "vhost-user virtio-fs device backend written in Rust"
     url: "https://gitlab.com/virtio-fs/virtiofsd"
-    version: "v1.6.1"
-    toolchain: "1.66.0"
+    version: "v1.8.0"
+    toolchain: "1.72.0"
     meta:
-      # From https://gitlab.com/virtio-fs/virtiofsd/-/releases/v1.7.0,
-      # this is the link labelled virtiofsd-v1.7.0.zip
+      # From https://gitlab.com/virtio-fs/virtiofsd/-/releases/v1.8.0,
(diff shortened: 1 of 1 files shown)


versions.yaml
```diff
@@ -329,14 +329,14 @@ externals:
   virtiofsd:
     description: "vhost-user virtio-fs device backend written in Rust"
     url: "https://gitlab.com/virtio-fs/virtiofsd"
-    version: "v1.6.1"
+    version: "v1.8.0"
     toolchain: "1.66.0"
```
fidencio, line 333: @skaegi, mind to also bump the toolchain version to 1.72.0?

The reason for that is that virtiofsd has been using the alpine:rust image (see: https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/.gitlab-ci.yml?ref_type=heads#L6), which right now used the 1.72.0 version of the toolchain.

skaegi: Sure - done!

fidencio (approved):
lgtm, thanks @skaegi!

GabyCT (approved):

gkurz:
Don't backport enhancements to stable streams, only bug fixes.

gkurz:
> Don't backport enhancements to stable streams, only bug fixes.

After discussing this again with @fidencio , we've decided this is worth backporting to stable-3.2 since 3.2.0 isn't released yet, but definitely not to stable-3.1 as it could break existing production setups.

Issue #7996, opened 2023-09-19
Support for block file snapshotter/rootfs loop mounts


alex-matei:
**Which feature do you think can be improved?**

Recently containerd introduced a new snapshotter: `blockfile`. 
It uses disk images mounted via loop devices for storing snapshots. When using kata the images are currently mounted
on the host and exposed via virtiofs to the guest. This is not ideal performance wise,
the rootfs image should be directly attached to the VMM and kata should mount it directly inside the VM.

PR #7997, opened 2023-09-19
runtime: support for loop mounts for rootfs

alex-matei:
This PR is a follow-up of #7850, it is built on top of it. It adds support for rootfs loop mounts. 
Containerd recently introduced a new snapshotter, `blockfile`, which uses disk images for storing snapshots. It creates loop devices in order to mount the disk images. The normal flow when using `runc` would be: `create loop device for image -> mount the device -> ...`. For kata we replace this with the following flow: `attach file as disk to vm (on host) -> mount the disk device (guest) -> ...`

Fixes: #7996


alex-matei:
This haven't been merged, please reopen.

alex-matei:
@fidencio @amshinde @gkurz Can you please reopen this? It has not been merged and was automatically closed :(.

PR #7998, opened 2023-09-19
CC | version: Specify forked / vanilla entries of containerd

fidencio:
As we'll need to test with both the vanilla and the forked versions of containerd, let's make sure we'll specify both entries as part of our versions.yaml file, and we can read whatever we need accordingly as part of our tests jobs.

NOTE: This will never make into the `main` branch, but we need this right now in order to have our tests properly setup.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -300,12 +300,12 @@ externals:
   nydus:
     description: "Nydus image acceleration service"
     url: "https://github.com/dragonflyoss/image-service"
-    version: "v2.2.1"
+    version: "v2.2.3"
 
   nydus-snapshotter:
     description: "Snapshotter for Nydus image acceleration service"
     url: "https://github.com/containerd/nydus-snapshotter"
-    version: "v0.3.3"
+    version: "v0.12.0"
 
   ovmf:
     description: "Firmware, implementation of UEFI for virtual machines."
--- a/versions.yaml
+++ b/versions.yaml
@@ -221,14 +221,23 @@ externals:
   containerd:
     description: |
       Containerd for Kubernetes Container Runtime Interface.
-    # CCv0 is using our fork of containerd as the changes can't be merged yet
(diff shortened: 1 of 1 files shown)


fidencio:
Please, **DO NOT RUN `/test` HERE**, at least not yet.

versions.yaml
```diff
@@ -221,14 +221,23 @@ externals:
...
+      tarball_url: "https://github.com/confidential-containers/containerd/releases/download"
+      version: "v1.6.8.2"
+      # CCv0 needs to know about the specific branch for the integration tests
+      branch: "CC-main"
+    upstream:
+      # DO NOT TOUCH DURING MERGES FROM MAIN TO CCv0
+      url: "github.com/containerd/containerd"
+      # yamllint disable-line rule:line-length
+      tarball_url: "https://github.com/containerd/containerd/releases/download"
+      version: "v1.7.2"
```
stevenhorsman, line 238: Are you happy for us to be ahead of the `main` versions.yaml here? I'm not against this, but it means that we'd probably need to bump containerd in `main` before we merge this function to `main` if we aren't confident that it works on 1.6.8

fidencio: So, we're already testing on `main` with the `latest` containerd (at least the cri-containerd tests).  The k8s tests are relying on what k3s is using, which is also 1.7+

fidencio: It's very hard to compare this with `main`, as we have changed the way we consume those for the tests on main.

stevenhorsman: Cool - fair point, if AKS is using containerd 1.7+ and we can kill on Jenkins soon then this problem goes away a bit

stevenhorsman (approved):
LGTM assuming the tests are passing

fidencio:
TEEs issues are not related to this PR.

Issue #8009, opened 2023-09-20
CC | Reduce guest components binary size with strip


arronwy:
Guest rootfs will aligned to 128M,  we may exceed the rootfs with several megabytes but the rootfs will add 128M.

PR #8010, opened 2023-09-20
CC | osbuild: Reduce guest components binary size with strip

arronwy:
Guest rootfs will aligned to 128M, we may exceed the rootfs with several megabytes but the rootfs will add 128M.

Fixes: #8009


fidencio (approved):
lgtm, thanks @arronwy!

fidencio:
As this PR is not TDX / SEV specific, and the TDX / SEV CIs haven't been running okay for a few days, I'm going ahead and merging it as all other CIs passed.

PR #8021, opened 2023-09-20
CC | cache: Fix rootfs-image-tdx and rootfs-initrd-sev cached artefacts

fidencio:
The name of the tarballs changed on main, but we didn't follow up changing this on the CCv0 branch. :-/


fidencio:
As this code is **NOT RELATED** to anything but the cache.

PR #8023, opened 2023-09-20
runtime-rs: Update status for pause and resume

### LINKED ISSUE #6434: runtime-rs: ch: Implement pause/resume VM
Implement the following methods in https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs:

- `pause_vm()`
- `resume_vm()`

cmaf:
Pause and resume task do not currently update the status of the container to paused or running, so fix this. This is for specifically pausing the task and not the VM.

Fixes #6434

### DIFF
--- a/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
@@ -349,28 +349,42 @@ impl Container {
     }
 
     pub async fn pause(&self) -> Result<()> {
-        let inner = self.inner.read().await;
-        if inner.init_process.get_status().await == ProcessStatus::Paused {
-            warn!(self.logger, "container is paused no need to pause");
+        let mut inner = self.inner.write().await;
+        let status = inner.init_process.get_status().await;
+        if status != ProcessStatus::Running {
+            warn!(
+                self.logger,
+                "container is in {:?} state, will not pause", status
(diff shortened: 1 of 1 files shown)


GabyCT (approved):

cmaf:
Added WIP label since I need to make `pause` and `resume` atomic before merging

cmaf:
The locking is already covered by `.write()` which returns a RwLockWriteGuard which locks with exclusive write access.

jodh-intel (approved):
Thanks @cmaf - lgtm.

Comment: yes, this does initially feel slightly "sneaky" in that we don't set the intermediate "pausing". However, it does mean that we can _effectively_ be atomic and tbh if we were to use `ProcessStatus::Pausing`, there should strictly be a corresponding `ProcessStatus::Resuming`, but there isn't :smile:

If we find this feature gets used a lot _and_ if we encounter noticeable delays in pausing / resuming, we can revisit the issue at that point.

> **Note:** For bonus points, it should be possible to construct a test that uses a fake `Agent` inside the `ContainerManager` so that you could test the failure scenarios and assert the expected states (possibly on a follow-up PR?)

amshinde (approved):

amshinde:
Looks like there was an error on uploading the tarball in GH workflow. Rerunning jobs.

Issue #8031, opened 2023-09-21
ci: Trigger payload-after-push on workflow_dispatch


fidencio:
This will allow us to easily test failures and fixes on that workflows.

PR #8032, opened 2023-09-21
ci: Trigger payload-after-push on workflow_dispatch

fidencio:
This will allow us to easily test failures and fixes on that workflows.

Fixes: #8031

### DIFF
--- a/.github/workflows/payload-after-push.yaml
+++ b/.github/workflows/payload-after-push.yaml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - stable-*
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
LGTM as I like to add manual triggering on all my workflows (but maybe that's as I need more debug attempts than most!)

zvonkok (approved):
LGTM, @stevenhorsman That was my first thought as well; "I NEED MANUAL TRIGGERING"

fidencio:
Bypassing the CIs as, again, this will only aftect a workflow after it gets merged.

Issue #6715, opened 2023-04-25
Direct cross-container shared-mount support


bergwolf:
Some workloads need a sidecar to mount some storage and share it with the application container.

Currently, there are some drawbacks:
* mount operation inside the sidecar requires SYS_ADMIN privilege. We need to support mount w/o root access.
* to share a mount between containers, we have to set up shared bind mounts between containers and the host. We need a way to share mounts w/o polluting the rootns mount namespace.

To address these drawbacks, we can add userns support in the guest/kata-agent to enable mount w/o root access for containers, and use the new linux kernel mount API (`open_tree()`/`move_mount()`) to set up bind mounts between containers, w/o polluting the guest rootns mount namespace.

Key change:
* agent: support container userns inside the guest
* agent: use open_tree()/move_mount() to set up bind mounts between containers directly 
* runtime-rs: support declaration of shared mounts between containers

changweige:
This is a handy feature for serverless products since customers may pack their fuse storage clients as a sidecar container where the users want to mount file systems alone.  Expecting this 👍

h56983577:
I'm working on it. ^_^

PR #8033, opened 2023-09-21
agent: use open_tree()/move_mount() to set up bind mounts between containers directly.

h56983577:
Some workloads need a sidecar to mount some storage and share it with the application container. This feature supports sharing mount objects directly between containers through annotations. Thus, the application container can access the storage mounted by the sidecar.

The advantage of using a sidecar to mount devices is that the sidecar can act as a FUSE daemon, which can work with the user namespace to achieve non-privileged mounting.
shared mount conception

To enable this feature, simply declare it within the annotations like this. This is a JSON-formatted list containing some shared mount declarations. Each declaration must include the following fields:
- `name` to uniquely identify a pair of mount points.
- `src_ctr` specifying the source container's name.
- `src_path` indicating the source container's mount point path.
- `dst_ctr` specifying the destination container's name.
- `dst_path` indicating the destination container's mount point path.

```yaml
annotations:
    io.katacontainers.config.runtime.shared_mounts: |
    [{"name":"storage","src_ctr":"sidecar","src_path":"/mnt/storage","dst_ctr":"app","dst_path":"/mnt/storage"}]
```

Runtime can get the container's name through the `io.kubernetes.cri.container-name` annotation provided by containerd and dispatch mount tasks to the agent (For cri-o, this annotation is `io.kubernetes.cri-o.ContainerName`).

Since shared mount point is created by application within the source container, the destination container needs to ensure it is ready by polling `/proc/mounts` of the source container. Then, it can use `open_tree()/move_mount()` to clone this mount object to the destination container.

shared mount process

Fixes: #6715.

### DIFF
--- a/src/agent/src/sandbox.rs
+++ b/src/agent/src/sandbox.rs
@@ -7,19 +7,24 @@ use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
 use std::fs;
+use std::os::fd::FromRawFd;
 use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 use std::{thread, time};
 
 use anyhow::{anyhow, Context, Result};
 use kata_types::cpu::CpuSet;
 use kata_types::mount::StorageDevice;
-use libc::pid_t;
+use libc::{pid_t, syscall};
+use nix::fcntl::{self, OFlag};
+use nix::sched::{setns, unshare, CloneFlags};
+use nix::sys::stat::Mode;
 use oci::{Hook, Hooks};
-use protocols::agent::OnlineCPUMemRequest;
+use protocols::agent::{OnlineCPUMemRequest, SharedMount};
 use regex::Regex;
 use rustjail::cgroups as rustjail_cgroups;
 use rustjail::container::BaseContainer;
@@ -254,6 +259,12 @@ impl Sandbox {
         self.containers.get_mut(id)
     }
 
+    pub fn find_container_by_name(&self, name: &str) -> Option<&LinuxContainer> {
+        self.containers
+            .values()
+            .find(|&c| c.config.container_name == name)
+    }
+
     pub fn find_process(&mut self, pid: pid_t) -> Option<&mut Process> {
         for (_, c) in self.containers.iter_mut() {
             if let Some(p) = c.processes.get_mut(&pid) {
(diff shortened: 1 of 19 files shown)


src/agent/src/sandbox.rs
```diff
@@ -403,6 +407,107 @@ impl Sandbox {
...
+        let mut mnt_ns = Namespace::new(&self.logger).get_mount();
+        mnt_ns.path = format!("/proc/{}/ns/mnt", c.init_process_pid);
+    
+        let new_thread = std::thread::spawn(move || {
+            if let Err(err) = || -> Result<()> {
+                mnt_ns.switch_to()?;
+                block_on(async move {
+                    for m in mounts {
+                        if m.type_() == "dst" {
+                            let shared_mounts = shared_mounts.lock().await;
```
bergwolf, line 475: The thread may win the lock race before `add_shared_mounts` and find out `shared_mounts` to be empty. We need something more reliable to make sure `move_mount` happens **after** `open_tree`.

Actually, I think the current workflow is a bit confusing. Can we simplify it a little bit such that:
1. when creating the source mount point container, do nothing;
2. when creating the destination mount point container, if there are shared mount configured, do the following
  1. spawn a new thread, which
  2. waits for the source mount point in the source container to be ready, by polling the mount point directory to see if it becomes a mount point, with a short sleep between each poll of course
  3. when the source mount point is ready, create a bind mount from it to the destination mount point

src/libs/protocols/protos/agent.proto
```diff
@@ -444,6 +448,23 @@ message FSGroup {
 	types.FSGroupChangePolicy group_change_policy = 3;
 }
 
+// SharedMount declares a set of shared bind mount points that support
+// cross-container sharing of mount objects. Including the mount point's name,
+// path and type is necessary.
+message SharedMount {
```
bergwolf, line 454: Let's just pass `config::SharedMount` to the agent in `CreateContainerRequest` for the destination container.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container.rs
```diff
@@ -168,13 +169,36 @@ impl Container {
             linux.resources = resources;
         }
 
+        let ctr_name = k8s::container_name(&spec);
+        let mut shared_mounts = Vec::new();
+        for shared_mount in &toml_config.runtime.shared_mounts {
```
bergwolf, line 174: Here let's only set shared_mount for the dst container.

src/libs/kata-types/src/annotations/crio.rs
```diff
@@ -7,6 +7,7 @@
 #![allow(missing_docs)]
 
 pub const CONTAINER_TYPE_LABEL_KEY: &str = "io.kubernetes.cri.container-type";
```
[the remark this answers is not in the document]
h56983577: Sure, thanks! I'll also make those changes.

src/libs/kata-types/src/config/shared_mount.rs
```diff
@@ -0,0 +1,83 @@
+use std::io::Result;
+
+use regex::Regex;
+
+use crate::eother;
+
+#[derive(Debug, Deserialize, Serialize, Clone, Default)]
+pub struct SharedMount {
+    /// Name is used to declare a group of related mount points.
+	/// Mount points in different containers can find each other using the same name.
```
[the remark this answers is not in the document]
h56983577: Sorry, I'll check the code formatting before submitting in the future.

src/agent/src/sandbox.rs
h56983577: Thank you so much for your thoughtful review and valuable suggestions. I'll promptly make the necessary changes to the PR.

src/agent/src/sandbox.rs
```diff
@@ -403,6 +417,124 @@ impl Sandbox {
...
+            "/proc/self/ns/mnt",
+            OFlag::O_RDONLY | OFlag::O_CLOEXEC,
+            Mode::empty(),
+        )?;
+        let dst_mntns_path = format!("/proc/{}/ns/mnt", c.init_process_pid);
+        let dst_mntns = fcntl::open(
+            dst_mntns_path.as_str(),
+            OFlag::O_RDONLY | OFlag::O_CLOEXEC,
+            Mode::empty(),
+        )?;
```
bergwolf, line 451: Both init_mntns and dst_mntns are leaked here. You can use `fs::File::from_raw_fd` to wrap it up into a `File` so that it is closed properly.

src/agent/src/sandbox.rs
```diff
@@ -403,6 +417,124 @@ impl Sandbox {
...
+                        setns(src_mntns, CloneFlags::CLONE_NEWNS)?;
+                        let src = std::ffi::CString::new(m.src_path())?;
+                        let mount_fd = unsafe {
+                            syscall(
+                                libc::SYS_open_tree,
+                                libc::AT_FDCWD,
+                                src.as_ptr(),
+                                0x1 | 0x8000 | libc::O_CLOEXEC, // OPEN_TREE_CLONE | AT_RECURSIVE | OPEN_TREE_CLOEXEC
+                            ) as i32
+                        };
```
bergwolf, line 504: The same applies to `src_mntns` and `mount_fd` as well. Please make sure that they are closed properly.

src/agent/src/sandbox.rs
```diff
@@ -403,6 +417,124 @@ impl Sandbox {
...
+                                    is_ready = true;
+                                    break;
+                                }
+                            }
+
+                            if is_ready {
+                                break;
+                            }
+
+                            thread::sleep(Duration::from_millis(100));
```
bergwolf, line 485: Let's add a timeout (e.g., 10 seconds?) here to break the loop and fail the operation if the src dir cannot become a mountpoint for some time.

src/agent/src/sandbox.rs
```diff
@@ -403,6 +417,124 @@ impl Sandbox {
...
+        if src_ctrs.len() == 0 {
+            return Ok(());
+        }
+
+        let mounts = mounts.clone();
+        let init_mntns = fcntl::open(
+            "/proc/self/ns/mnt",
+            OFlag::O_RDONLY | OFlag::O_CLOEXEC,
+            Mode::empty(),
+        )?;
```
bergwolf, line 445: For all of these nix wrapped syscalls, please add some additional error texts otherwise it will be very difficult to debug related failures.

src/agent/src/sandbox.rs
h56983577: Sure, thanks! I didn't consider the issue of resource leaks, particularly file descriptors.

src/agent/src/sandbox.rs
h56983577: Agreed!

src/agent/src/sandbox.rs
h56983577: OK!

bergwolf (commented):
Generally lgtm. Let's see how the test goes.

bergwolf:
Please amend your commit subject and message.
image

h56983577:
> Please amend your commit subject and message. image

@bergwolf I'm sorry. The commit message is too long, and I forgot to run `make check` in runk. :cry:

bergwolf (approved):
lgtm, thanks!

liubin (approved):

h56983577:
> lgtm, thanks!

@bergwolf Thank you for your dedicated guidance over the past six months, patiently answering all my questions, and steering me from scratch to complete an open-source community PR. Moreover, learning alongside a top programmer during my student days has been a precious experience for me. Once again, thank you! :laughing:

PR #8036, opened 2023-09-21
runtime: fix reading cgroup stats of sandboxes

### LINKED ISSUE #8035: Incorrect pod overhead metrics
# Description of problem
We've encountered an issue with pod overhead metrics provided by the kata shim. In particular, the `kata_shim_pod_overhead_cpu` is always negative and `kata_shim_pod_overhead_memory_in_bytes` is always a value close to uint64 max (way above the memory installed on host machine).

The kata version we're using is 3.1.0 (golang runtime). The `sandbox_cgroup_only` is set to true.

# Expected result
Correct overhead metrics.

# Actual result
Incorrect values as described above.

```
$ sudo curl --silent --unix-socket "/run/vc/sbs/6788b39a7ed42f2b945e2e4fdf8e3ae69874958af20a9f7b3329f8fa1526c3fd/shim-monitor.sock" "http://shim/metrics" | grep -E '^kata_shim_pod_overhead'
kata_shim_pod_overhead_cpu -11.00378333139894
kata_shim_pod_overhead_memory_in_bytes 1.8446744070680211e+19
```

# Further information
I tracked the issue down to a [switch statement](https://github.com/kata-containers/kata-containers/blob/dd27912f3112d84170e80bd910a2b9e8861ef6f4/src/runtime/virtcontainers/sandbox.go#L1744-L1751) in sandbox.go. It's expecting two different kinds of Metrics structs (cgroup v1 or v2), but it's actually getting pointers to said structs. Therefore the host values are for cpu/mem are always 0, thus the wrong metric values.

I'll shortly open a PR with a fix for this issue.

rye-stripe:
The cgroup stats come from `resourcecontrol` package in the form of pointers to structs. The sandbox `Stat()` method incorrectly was expecting structs. This caused the cpu and memory stats to always be 0, which in turn caused incorrect pod overhead metrics.

Fixes #8035

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -1742,12 +1742,14 @@ func (s *Sandbox) Stats(ctx context.Context) (SandboxStats, error) {
 
 	// TODO Do we want to aggregate the overhead cgroup stats to the sandbox ones?
 	switch mt := metrics.(type) {
-	case v1.Metrics:
+	case *v1.Metrics:
 		stats.CgroupStats.CPUStats.CPUUsage.TotalUsage = mt.CPU.Usage.Total
 		stats.CgroupStats.MemoryStats.Usage.Usage = mt.Memory.Usage.Usage
-	case v2.Metrics:
+	case *v2.Metrics:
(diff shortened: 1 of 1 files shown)


gkurz (approved):
Good catch. Thanks @rye-stripe !

sprt (approved):

PR #8039, opened 2023-09-21
docs: Fix description of container without cpu constraint

garvinp-stripe:
docs: Update docs around containers without constraint

Fixes: https://github.com/kata-containers/kata-containers/issues/8088

Updating the doc to correct describe the actual behavior of containers when they are created without constraint. Specifically each container without constraint will still have their own cgroup however with a -1 cfs_quota_us (unrestrained). The original doc seems to indicate that containers without constraint would be limited to `default_vcpus` which isn't the case given what the actual cgroups look like.

Signed-off-by: Garvin Pang <garvinp@stripe.com>


stevenhorsman:
@garvinp-stripe Hey - thanks for the PR. Kata containers commits need a message body that matches the [patch format](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format) in the contribution guidelines and also need an issue to reference the fix against otherwise the checks will fail. Let us know if you need any help with that!

GabyCT (approved):

Issue #8027, opened 2023-09-21
ci: rootfs-image build-asset is failing


portersrc:
The github action to build rootfs-image is failing on main.

PR #8045, opened 2023-09-22
local-build: Fix .docker ownership before build-payload

jepio:
The permissions on .docker/buildx/activity/default are regularly broken by us passing docker.sock + $HOME/.docker to a container running as root and then using buildx inside. Fixup ownership before executing docker commands.

Fixes: #8027


BbolroC (approved):
LGTM, thanks @jepio !

sprt (approved):

GabyCT (approved):

PR #8046, opened 2023-09-22
runtime: remove unimplemented CoCo configurations

fitzthum:
At the very beginning of the CoCo project, we added some speculative options to the config to guide development. It turns out that these don't match the features we implemented and that we will soon be porting to main. Let's remove them.

Technically the service_offload section still doesn't do much in main, but it will soon.


fidencio:
> Technically the service_offload section still doesn't do much in main, but it will soon.

This is an interesting one, @fitzthum.  TIL that `service_offload` is irrelevant when using the nydus snapshotter, as offloading the image to the guest or to the guest is a snapshotter config, meaning that we most likely can drop it as well.

fidencio:
IOW, I think we can drop the whole `image` section.

fitzthum:
@ChengyuZhu6 Can you confirm that we won't need this `service_offload` flag?

ChengyuZhu6:
> @ChengyuZhu6 Can you confirm that we won't need this `service_offload` flag?

@fitzthum , I'm very sorry for the delay in replying to you. I had tested it and I'm sure that we won't need `service_offload`.

ChengyuZhu6:
@fitzthum One query here, but otherwise it looks good to me.

ChengyuZhu6:
Same question like @fidencio. Is `Image` section no longer necessary? Should we maintain the flags other than `service_offload `?

stevenhorsman (approved):
Sorry I've missed/ignored this for a long time. I think we've discussed it on other issues that probably lead to this PR being created and I agree that the image configuration is not necessary and in fact confusing now, so the rebased version of this PR LGTM!

fitzthum:
Ah yeah, I forgot about this two. Just gave it a rebase.

ChengyuZhu6 (approved):
LGTM thanks @fitzthum

src/runtime/pkg/katautils/config.go
```diff
@@ -70,8 +70,7 @@ type tomlConfig struct {
 }
 
 type image struct {
-	Provision      string `toml:"provision"`
-	ServiceOffload bool   `toml:"service_offload"`
+	Provision string `toml:"provision"`
 }
```
stevenhorsman, line 74: I think we can get rid of the whole `image` struct now we've removed that block from the configuration.toml files?

fitzthum: Good point. Removed.

stevenhorsman:
Sorry Tobin, it looks like the image timeout config has conflicted with these. If you are able to rebase today that would be great, otherwise let me know and I can try and take a look tomorrow

Issue #8057, opened 2023-09-25
Upgrade to Cloud Hypervisor v35.0


likebreath:
### Paravirtualised Panic Device Support

A new device has been added that can communicate when the guest kernel has panicked and share those details with the VMM. This is controlled with a new `--pvpanic` command line option and JSON API change equivalent. 

### Improvements to VM Core Dump

Requesting to dump the guest memory as core dump will now transparently pause the VM if required; returning to the original state after. 

### QCOW2 Support for Backing Files

The support for QCOW2 files has been enhanced to include support for using backing files. 

### Minimum Host Kernel Bump

The minimum supported host kernel is now 5.13 in order to incorporate a bug fix for `KVM_FEATURE_ASYNC_PF_INT` functionality. 

### Notable Bug Fixes

* The x86 emulator is only compiled in if MSHV is compiled in (the kernel carries out this job with KVM)
* A regression has been fixed in VFIO support for devices that use MSI rather  than MSI-X
* When triggering a VM shutdown or reset via I/O the vCPU thread will be  blocked until that asynchronous event has been received
* Pausing a VM is now a synchronous action with the request only completing  when all vCPUs are paused
* Event monitor support now correctly supports concurrent access
* Bug fixes for the QCOW2 file support 

### `virtio-vsock` Support for Linux Guest Kernel v6.3+

Since kernel v6.3, a vsock packet can be included in a single descriptor,
instead of being split over two descriptors. Our `virtio-vsock` implementation
now support both situations.

### User Specified Serial Number for `virtio-block`

A new option `serial` is added to the `--block` command that allows users to
specify a serial number for block devices which will be guest visible.

### vCPU TSC Frequency Included in Migration State

This ensures migration works correctly between hosts that have different TSC
frequencies if the guest is running with TSC as the source of timekeeping.

### Notable Bug Fixes

* Disallow concurrent CPU resizing
* Handle APIC EOI message for MSHV
* Use wrapping add for memory offset from instruction emulator
* Add global spell check via the 'typos' GitHub action 
* Ensure probing reads are block size aligned 
* Multiple bug fixes around the latency counter for block devices
* Replace unsound `static mut` with `once_cell`

skaegi:
Should be considered for back-porting for 3.2 --  https://github.com/advisories/GHSA-49hh-fprx-m68g is fixed in v35 -- yes minor but teams like mine still have to react to it.

likebreath:
[...]

PR #8058, opened 2023-09-25
Upgrade to Cloud Hypervisor v35.0

likebreath:
### Paravirtualised Panic Device Support

A new device has been added that can communicate when the guest kernel has panicked and share those details with the VMM. This is controlled with a new `--pvpanic` command line option and JSON API change equivalent. 

### Improvements to VM Core Dump

Requesting to dump the guest memory as core dump will now transparently pause the VM if required; returning to the original state after. 

### QCOW2 Support for Backing Files

The support for QCOW2 files has been enhanced to include support for using backing files. 

### Minimum Host Kernel Bump

The minimum supported host kernel is now 5.13 in order to incorporate a bug fix for `KVM_FEATURE_ASYNC_PF_INT` functionality. 

### Notable Bug Fixes

* The x86 emulator is only compiled in if MSHV is compiled in (the kernel carries out this job with KVM)
* A regression has been fixed in VFIO support for devices that use MSI rather  than MSI-X
* When triggering a VM shutdown or reset via I/O the vCPU thread will be  blocked until that asynchronous event has been received
* Pausing a VM is now a synchronous action with the request only completing  when all vCPUs are paused
* Event monitor support now correctly supports concurrent access
* Bug fixes for the QCOW2 file support 

### `virtio-vsock` Support for Linux Guest Kernel v6.3+

Since kernel v6.3, a vsock packet can be included in a single descriptor,
instead of being split over two descriptors. Our `virtio-vsock` implementation
now support both situations.

### User Specified Serial Number for `virtio-block`

A new option `serial` is added to the `--block` command that allows users to
specify a serial number for block devices which will be guest visible.

### vCPU TSC Frequency Included in Migration State

This ensures migration works correctly between hosts that have different TSC
frequencies if the guest is running with TSC as the source of timekeeping.

### Notable Bug Fixes

* Disallow concurrent CPU resizing
* Handle APIC EOI message for MSHV
* Use wrapping add for memory offset from instruction emulator
* Add global spell check via the 'typos' GitHub action 
* Ensure probing reads are block size aligned 
* Multiple bug fixes around the latency counter for block devices
* Replace unsound `static mut` with `once_cell`

Details can be found here: https://github.com/cloud-hypervisor/cloud-hypervisor/releases/

Fixes: #8057

### DIFF
--- a/src/runtime/virtcontainers/pkg/cloud-hypervisor/client/api/openapi.yaml
+++ b/src/runtime/virtcontainers/pkg/cloud-hypervisor/client/api/openapi.yaml
@@ -485,6 +485,7 @@ components:
             iommu: false
             queue_size: 5
             vhost_socket: vhost_socket
+            serial: serial
             vhost_user: false
             direct: false
             rate_limiter_config:
@@ -504,6 +505,7 @@ components:
             iommu: false
             queue_size: 5
             vhost_socket: vhost_socket
+            serial: serial
             vhost_user: false
             direct: false
             rate_limiter_config:
@@ -847,6 +849,7 @@ components:
           iommu: false
           queue_size: 5
           vhost_socket: vhost_socket
+          serial: serial
           vhost_user: false
           direct: false
           rate_limiter_config:
@@ -866,6 +869,7 @@ components:
           iommu: false
           queue_size: 5
           vhost_socket: vhost_socket
+          serial: serial
           vhost_user: false
           direct: false
           rate_limiter_config:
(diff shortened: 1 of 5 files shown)


fidencio (approved):
lgtm, thanks @likebreath!

amshinde (approved):

jodh-intel:
Better late than never...

FTR, I enabled the [`ch` feature](https://github.com/kata-containers/kata-containers/pull/6201#issuecomment-1413690515) and tested the rust runtime with CH v35 and it works in the major scenarios (TDX, default, DM).

PR #8061, opened 2023-09-26
Updating containerd to a GogoProtobuf free version

beraldoleal:
This is a substantial PR aimed at updating the containerd to a version that is free of gogoprotobuf, thereby removing this dependency from the entire project. To achieve this, several other libraries needed to be bumped along with containerd.

I welcome your feedback on this PR.


src/runtime/virtcontainers/factory/grpccache/grpccache.go
```diff
@@ -10,7 +10,8 @@ import (
 	"context"
 	"fmt"
 
-	types "github.com/gogo/protobuf/types"
+        emptypb "google.golang.org/protobuf/types/known/emptypb"
```
littlejawa, line 13: nit: not aligned? (`go fmt` found this)

littlejawa (approved):
I've reported 1 nit, but there is actually more in there if you run `go fmt` on the code you changed.
Basically: the import lines you added/modified are moved and/or realigned for consistency.
Not a big deal, but you may want to run it and fix as part of your commit.

Otherwise LGTM
Thanks @beraldoleal

gkurz:
BTW @beraldoleal you might squash patch 2 into patch 1 right away : it isn't a problem for review since GH doesn't expand big diffs  and it would fix the commit message check.

src/runtime/virtcontainers/factory/grpccache/grpccache.go
beraldoleal: Fixed. Thank you.

src/runtime/go.mod
```diff
@@ -4,114 +4,129 @@ go 1.19
...
 	code.cloudfoundry.org/bytefmt v0.0.0-20211005130812-5bb3c17173e5
-	github.com/BurntSushi/toml v1.2.0
+	github.com/BurntSushi/toml v1.2.1
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/blang/semver/v4 v4.0.0
-	github.com/containerd/cgroups v1.0.5-0.20220625035431-cf7417bca682
+	github.com/containerd/cgroups v1.1.0
 	github.com/containerd/console v1.0.3
-	github.com/containerd/containerd v1.6.8
+	github.com/containerd/containerd v1.7.2
```
wainersm, line 12: @beraldoleal I'd like to double-check this bump is ok.

As per [`versions.yaml`](https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L217) we support 3 versions of containerd: `version` (minimum?), `lts` and `active` , based on https://containerd.io/releases/#support-horizon

On commit 7319cff77a @fidencio introduced CI for `lts` and `active`.

Bumping runtime to v1.7.2 it seems will replace the `version` required.

src/libs/protocols/hack/update-generated-proto.sh
```diff
@@ -81,8 +83,8 @@ fi;
 which protoc
 [ $? -eq 0 ] || die "Please install protoc from github.com/protocolbuffers/protobuf"
 
-which protoc-gen-gogottrpc
-[ $? -eq 0 ] || die "Please install protoc-gen-gogottrpc from https://github.com/containerd/ttrpc"
+which protoc-gen-gottrpc
+[ $? -eq 0 ] || die "Please install protoc-gen-gottrpc from https://github.com/containerd/ttrpc"
```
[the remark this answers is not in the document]
beraldoleal: Hi @wainersm maybe is outdated, yes. They still have binaries for code generation for both... https://github.com/containerd/ttrpc/tree/main/cmd.

I choose to keep with containerd/ttrpc, because of a few things:

1. We depend on containerd, then aligning this project to use ttrpc would ensure compatibility;
2. Another alternative would be to adopt [protobiuld](https://github.com/containerd/protobuild). But I would like to make this change as simple as possible, since it is already a big change.
3. If future versions of containerd move away from ttrpc, then we might want to re-evaluate whether to continue using ttrpc or not.

src/runtime/go.mod
beraldoleal: Always good to be safe! Thanks @wainersm.

The issue is that if we are moving away from gogo, then 1.7 should be the minimum.

beraldoleal:
@wainersm We made some progress here... Are you aware of this error:

https://github.com/kata-containers/kata-containers/actions/runs/6732628337/job/18299691488?pr=8061

It seems to be not related to this PR.

Also, do you mind triggering again the ones has failed? They seem to be flaky.

src/runtime/virtcontainers/pkg/agent/protocols/grpc/utils.go
```diff
@@ -126,9 +126,9 @@ func copyStructSkipField(to, from reflect.Value) bool {
 	var ociWindows specs.Windows
 
 	toType := to.Type()
-	grpcSolarisType := reflect.TypeOf(grpcSolaris)
+	grpcSolarisType := reflect.TypeOf(&grpcSolaris)
 	ociSolarisType := reflect.TypeOf(ociSolaris)
-	grpcWindowsType := reflect.TypeOf(grpcWindows)
+	grpcWindowsType := reflect.TypeOf(&grpcWindows)
```
yoheiueda, line 131: One minor comment here. Are these changes related to GoGo? Or, did you fix a potential bug?

yoheiueda (approved):
Thank you very much for this hard work! 

LGTM overall. I checked files except for vendor directory, and I see consistent changes, although I am not an export of protobuf.

wainersm:
@beraldoleal the three non-required jobs failing are flaky. I've seen the kata-deploy failing on those jobs on other PRs. Even though I am giving it a last chance to pass so I just triggered them.

wainersm:
ok, so now we have a much sane result on CI. The ` run-monitor` is known to be failing, there is already an open issue to track it.

I will proceed with the merge ;)

hey @beraldoleal thank you so much for this change! You know became our protobuf black belt ;)

beraldoleal:
Thank you @wainersm, @gkurz and everyone for all the help here.

Issue #5268, opened 2022-09-28
Implement a runtime-rs hypervisor plugin for Firecracker


jodh-intel:
See: #4634.

PR #8070, opened 2023-09-26
runtime-rs: firecracker hypervisor backend 

Pyrromanis:
Add a basic runtime-rs `Hypervisor` trait implementation for Firecracker (FC).

- implement fc-api requests to setup and start an FC sandbox
- account for jailer directory hacks
- add separate config for runtime-rs (might need to change as there are only minor diffs with the go config)

Notes:
- only devmapper is currently supported
- no sharefs functionality, copy files in the sandbox
- nerdctl spawn is broken (TODO: #7703)
- pending FC-specific doc

Fixes: #5268


src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
```diff
@@ -0,0 +1,172 @@
...
+            jailed: false,
+            pending_devices: vec![],
+            capabilities,
+        }
+    }
+
+    pub(crate) async fn prepare_vmm(&mut self, netns: Option<String>) -> Result<()> {
+        let mut cmd: Command;
+        self.netns = netns.clone();
+        match !self.jailed {
```
pmores, line 65: Just playing a linter here ;-) but wouldn't this be a bit easier to read if it was `match self.jailed` and the arms were swapped?

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+        let dest = &[
+            &self.jailer_root,
+            "/",
+            HYPERVISOR_FIRECRACKER,
+            "/",
+            &self.id,
+            "/",
+            ROOT,
+        ]
+        .concat();
```
pmores, line 68: `jailed_location` right above in `jail_resource()` performs a very similar operation in a slightly different way.  It might be worth it to consider making a function to perform the path assembly up to the `ROOT` element.  Also, Rust's `PathBuf` type would seem to facilitate this kind of operation.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+            .await
+            .context(format!("failed to create directory {:?}", dest));
+        mount::bind_mount_unchecked(dest, dest, false, MsFlags::MS_SHARED)
+            .context("bind mount jailer root")?;
+
+        mount::bind_remount(dest, false).context("rebind mount jailer root")?;
+        Ok(())
+    }
+
+    pub(crate) async fn wait_api_socket(&mut self, id: &str) -> Result<()> {
```
pmores, line 79: I'm trying to understand the name of this function - why "wait"?

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+        let rootfs_driver = self.config.blockdev_info.block_device_driver.clone();
+
+        kernel_params.append(&mut KernelParams::new_rootfs_kernel_params(
+            &rootfs_driver,
+            &self.config.boot_info.rootfs_type,
+        )?);
+        kernel_params.append(&mut KernelParams::from_string(
+            &self.config.boot_info.kernel_params,
+        ));
+        let mut parameters: String = String::new().to_owned();
```
pmores, line 121: Would `let mut parameters = String::new();` do just the same, or is it just me still not knowing enough Rust?

src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
Pyrromanis: I mean sure, it was just that the jailer function came afterwards and that is why it is ordered like that, if the reviewers have a problem it will be swapped no problem.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
Pyrromanis: `PathBuf` would have been used if we had an extension or needed to push/pop a lot and we would need to mutate to string afterwards. `jailed_location` and `dest` are almost the same, but both are a single time concatenation and they already use a function for that exact purpose. I could see though ditching the `.concat()` here for the `.join()` used above.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
Pyrromanis: Yes you are correct, it will be fixed.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
Pyrromanis: We wait to see if the socket is ready or if something wrong has happened to it, as we cannot proceed if its not

jodh-intel (commented):
Thanks @Pyrromanis - just a few comments...

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+                    sl!(),
+                    "Request FAILED WITH STATUS: {:?} and BODY: {:?}", status, body
+                );
+                Some(body)
+            };
+        }
+
+        Err(anyhow::anyhow!(
+            "After {} attempts, it
+                            still doesn't work.",
```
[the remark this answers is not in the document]
Pyrromanis: Yes, it will be fixed

src/runtime-rs/crates/hypervisor/src/firecracker/inner_hypervisor.rs
```diff
@@ -0,0 +1,175 @@
...
+                    );
+                }
+            }
+        }
+        Ok(())
+    }
+
+    pub(crate) fn pause_vm(&self) -> Result<()> {
+        info!(sl!(), "Pausing Firecracker");
+        info!(sl!(), "FcInner: Currently not implemented");
```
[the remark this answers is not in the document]
Pyrromanis: will do

jodh-intel:
@Pyrromanis - please could you ping me on this PR once you've updated so I can re-review. Can't wait to see this land! ;)

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
pmores: Okay, `PathBuf` can be a judgment call, and `join()` does seem much better than `concat()` if you insist on representing paths as `String`s.

However, the path that's anonymously constructed here (and in multiple other places in this PR) has a proper name `chroot_dir` in `jailer`'s documentation.  Also, what's apparently called `jailer_root` in this code is called `chroot_base` by `jailer`.  I do believe that it would greatly benefit readability of this code if the original jailer terminology was used consistently by the firecracker driver.

This PR assembles _a lot_ of paths, mostly in an ad-hoc manner even though some of them are related by sharing a common prefix and/or semantics.  Even where `jailer` (or firecracker) doesn't have a name for them, it would be great if this code did.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+        self.request_with_retry(Method::PUT, "/boot-source", body_kernel)
+            .await?;
+        self.request_with_retry(Method::PUT, "/drives/rootfs", body_rootfs)
+            .await?;
+
+        let sb_path = get_sandbox_path(id)?;
+
+        let abs_dummy_path = match self.jailed {
+            false => [&sb_path, ROOT].join("/"),
+            true => [&self.jailer_root, HYPERVISOR_FIRECRACKER, &self.id, ROOT].join("/"),
```
pmores, line 157: Looks like yet another ad-hoc anonymous assembly of jailer's `chroot_dir`.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+        let body_kernel: String = json!({
+            "kernel_image_path": kernel,
+            "boot_args": parameters,
+        })
+        .to_string();
+
+        let body_rootfs: String = json!({
+            "drive_id": "rootfs",
+            "path_on_host": rootfs,
+            "is_root_device": false,
```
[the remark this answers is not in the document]
Pyrromanis: Yes it is correct even though it seems wrong. We add it as a regular block device so that we can use a partitioned root block device.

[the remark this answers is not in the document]
pmores: Fair enough, thanks! Maybe a comment to that effect so as to avoid confusing future readers would be nice.

src/runtime-rs/crates/hypervisor/src/firecracker/inner_device.rs
```diff
@@ -0,0 +1,56 @@
...
+                .await
+                .context("add block device"),
+            DeviceType::Network(network) => self
+                .add_net_device(&network.config, network.device_id)
+                .await
+                .context("add net device"),
+            _ => Err(anyhow!("unhandled device: {:?}", localdevice)),
+        };
+
+        Ok(localdevice)
```
pmores, line 40: Is all the device cloning necessary here?  `device` is cloned into `localdevice` right at the top so if the function always returns `device` then `localdevice` can safely be consumed in the course of function execution, right?

jodh-intel (approved):
Thanks @Pyrromanis.

lgtm

src/runtime-rs/crates/persist/src/lib.rs
```diff
@@ -29,6 +29,21 @@ pub fn to_disk<T: serde::Serialize>(value: &T, sid: &str) -> Result<()> {
     Err(anyhow!("invalid sid {}", sid))
 }
 
+pub fn to_disk_jailed<T: serde::Serialize>(value: &T, sid: &str, prefix: &str) -> Result<()> {
```
Tim-0731-Hzt, line 32: Maybe it's better to abstract `to_disk` and `to_disk_jailed` into one function, it seems there are some duplicate codes for these two functions.

Tim-0731-Hzt (commented):
Thanks @Pyrromanis,  generally LGTM,  Just one comment

src/runtime-rs/crates/persist/src/lib.rs
ananos: thanks @Tim-0731-Hzt! We would have been able to ditch this change altogether if that tiny bit of verify_id was a bit different (we could easily allow the '/' char by passing the prefix+'/'+sid to this function things would work as expected). 

However this is a combination of the FC jailer hardcoded jailer path (`<>/firecracker/<>`) and the fact that we didn't want to change the signature for the `to_disk` function. We'll refactor this logic and iterate over this to make sure we don't have any duplicate code in there. 
thanks!

src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
ananos: ack!

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
ananos: totally agree, we tried to refactor the paths and also "tried" to use as much of the terminology fit as possible (`jailer root` != `jailer chroot dir` etc.). We've come up to the following, and we plan to clarify some more in comments and in the follow-up doc. In short it is summarized early in the `inner_hypervisor.rs` file. 

jailed case:
`jailer_path`: jailer executable
`jailer_root`: usually `/run/kata`
`vm_path`: usually `/run/kata/{vmm_exec_filename}/{sid}`
`run_dir`: usually `/run/kata/{vmm_exec_filename}/{sid}/root/run`

non-jailed case:
`jailer_path`: empty
`jailer_root`: empty
`vm_path`: usually `/run/kata/{sid}`
`run_dir`: usually `/run/kata/{sid}/run`

```
        if !self.config.jailer_path.is_empty() {
            debug!(sl!(), "[Firecracker] Running jailed");
            self.jailed = true;
            self.jailer_root = KATA_PATH.to_string();
            debug!(sl!(), "[Firecracker] jailer_root: {:?}", self.jailer_root);
            self.vm_path = [
                self.jailer_root.clone(),
                HYPERVISOR_FIRECRACKER.to_string(),
                id.to_string(),
            ]
            .join("/");
            debug!(sl!(), "[Firecracker] VM Path: {:?}", self.vm_path);
            self.run_dir = [self.vm_path.clone(), "root".to_string(), "run".to_string()].join("/");
            debug!(sl!(), "[Firecracker] Rundir: {:?}", self.run_dir);
            let _ = self.remount_jailer_with_exec().await;
        } else {
            self.vm_path = [KATA_PATH.to_string(), id.to_string()].join("/");
            debug!(sl!(), "[Firecracker] VM Path: {:?}", self.vm_path);
            self.run_dir = [self.vm_path.clone(), "run".to_string()].join("/");
            debug!(sl!(), "[Firecracker] Rundir: {:?}", self.run_dir);
        }
```

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
ananos: ack. we used this for debug purposes (trying to understand whether FC was starting or not, so we were waiting until the socket appeared. Afterwards, we stopped waiting and just used this for stale sockets lying around). Not needed anymore, so we removed that.

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
ananos: ack

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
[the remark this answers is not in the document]
ananos: ack

src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
```diff
@@ -0,0 +1,172 @@
...
+                    sl!(),
+                    "Firecracker started successfully with id: {:?}, {:?}", id, cur_tid
+                );
+                self.pid = Some(id);
+            }
+            None => {
+                let exit_status = child.wait().await?;
+                error!(
+                    sl!(),
+                    "ERROR FC process exited Immediatelly {:?}", exit_status
```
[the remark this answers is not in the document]
ananos: ack

src/runtime-rs/crates/hypervisor/src/firecracker/inner_hypervisor.rs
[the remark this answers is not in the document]
ananos: thanks, we reworked most of the debug/warn/info messages to be consistent.

src/runtime-rs/crates/hypervisor/src/firecracker/inner_device.rs
ananos: ack, we have reworked the add_device function to capture more devices, even if this is needed quite early in the FC case.

src/runtime-rs/crates/persist/src/lib.rs
ananos: we have added a new parameter in the `to_disk` to account for this. However, we'd like some feedback about that, as we're not sure this captures the correct functionality for dragonball, and/or other hypervisors...

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -570,9 +572,31 @@ impl Persist for VirtSandbox {
         let sandbox_state = crate::sandbox_persist::SandboxState {
             sandbox_type: VIRTCONTAINER.to_string(),
             resource: Some(self.resource_manager.save().await?),
-            hypervisor: Some(self.hypervisor.save_state().await?),
+            //hypervisor: Some(self.hypervisor.save_state().await?),
```
pmores, line 575: This line could probably be removed, right?

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -570,9 +572,31 @@ impl Persist for VirtSandbox {
...
+            hypervisor: match self.hypervisor.save_state().await?.hypervisor_type.as_str() {
+                // TODO support other hypervisors
+                HYPERVISOR_DRAGONBALL => Ok(Some(self.hypervisor.save_state().await?)),
+                HYPERVISOR_FIRECRACKER => Ok(Some(self.hypervisor.save_state().await?)),
```
pmores, line 579: This looks odd to me, why is `save_state()` called again?  If the purpose of the first call (in the `match` expression above) is only to get the hypervisor type, isn't there a more direct way to do that?

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
ananos: ack, will clean up once we get the correct implementation on this snippet.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
ananos: exactly! maybe we should call save state earlier and keep it as a variable to pass it on there. Not sure, what do you think?

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
pmores: Right, I'm wondering though, why don't you just rename `jailer_root` to `chroot_base` in the code so that it directly matches jailer terminology?

ananos: well, no problem in doing that, however, there's no chroot_base in the kata code, only `jailer_root` (in both dragonball, CLH) so I think renaming it add another level of confusion maybe? not sure. What do the @kata-containers/rust people think? (as I'm not sure if the dragonball jailer code is in a working state... same stands for CLH I guess)

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
ananos: ACK, thanks!

src/runtime-rs/crates/hypervisor/src/firecracker/fc_api.rs
pmores: Okay, that's a fair point, thanks.  I'd still think that the relation between firecracker/jailer and its driver in runtime-rs is rather stronger than the one between different hypervisor drivers (and folks perusing this code seem more likely to be familiar with firecracker/jailer than other hypervisor drivers) but I'm not insisting. :-)

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -597,3 +601,18 @@ fn update_component_log_level(config: &TomlConfig) {
         updated_inner
     });
 }
+
+fn determine_jailed(config: &TomlConfig) -> Result<String> {
+    let Some(hypervisor_config) = config.hypervisor.get(&config.runtime.hypervisor_name) else {
```
pmores, line 606: Does this compile? It kind of looks like `if let` was meant but the following code seems to contradict that.

src/runtime-rs/crates/runtimes/src/manager.rs
```diff
@@ -597,3 +601,18 @@ fn update_component_log_level(config: &TomlConfig) {
...
+        todo!()
```
pmores, line 607: There seems to be uncertainty how to handle this, however (assuming I'm guessing the intent of the previous line correctly) this block will only execute if `hypervisor_name` points to an unconfigured hypervisor, right?  If so then that would clearly be a major misconfiguration, warranting a return with an `Err`.  Or am I misunderstanding completely?

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -567,12 +569,34 @@ impl Persist for VirtSandbox {
...
+            }?,
+        };
+        // FIXME: properly handle jailed case
+        // eg: Determine if we are running jailed:
+        // let h = sandbox_state.hypervisor.clone().unwrap_or_default();
+        // Figure out the jailed path:
+        // jailed_path = h.<>
+        // and somehow store the sandbox state into the jail:
+        // persist::to_disk(&sandbox_state, &self.sid, jailed_path)?;
+        // Issue is, how to handle restore.
```
pmores, line 593: Do I get the issue right that it's the chicken-egg problem where in order to restore a hypervisor you need to get the corresponding jailed path, for which you need the very same hypervisor you're trying to restore?

src/runtime-rs/crates/runtimes/src/manager.rs
ananos: correct -- AFAIU todo!() will raise an error. If not, we should definitely raise an error there.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
ananos: again, correct -- and we don't like the way we've implemented that (at all...) so we're open to suggestions about what to do here ...

src/runtime-rs/crates/runtimes/src/manager.rs
pmores: Oh I missed that addition to the language, thanks! :-)

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
pmores: Honestly I don't know if there can even be any other way apart from either 1) not saving the state to the jailed path, or 2) store the jailed path also somewhere outside of the state (which would require building something of a registry of sandbox state paths...).

src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub(crate) jailed: bool,
+    pub(crate) run_dir: String,
+    pub(crate) pending_devices: Vec<DeviceType>,
+    pub(crate) capabilities: Capabilities,
+}
+
+impl FcInner {
+    pub fn new() -> FcInner {
+        let mut capabilities = Capabilities::new();
+        capabilities.set(CapabilityBits::BlockDeviceSupport);
```
lifupan, line 47: Did the FC need to set HybridVsockSupport support specifically?

src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
```diff
@@ -0,0 +1,217 @@
...
+                    id,
+                    cur_tid
+                );
+                self.pid = Some(id);
+            }
+            None => {
+                let exit_status = child.wait().await?;
+                error!(
+                    sl!(),
+                    "[Firecracker] Process exited, status: {:?}", exit_status
```
lifupan, line 128: Maybe you can redefine a sl! macro including the Firecracker sub system instead of writing them in the log message every time?

lifupan:
Hi @Pyrromanis 

Thanks for the FC supporting effort, most of the patch LGTM except some small comments.

ananos:
> Hi @Pyrromanis
> 
> Thanks for the FC supporting effort, most of the patch LGTM except some small comments.

thanks @lifupan for the comments -- all look reasonable and prob. easy to address. We'll report back once we get to it!

src/libs/kata-types/src/config/hypervisor/firecracker.rs
```diff
@@ -0,0 +1,116 @@
...
+            validate_path!(
+                firecracker.path,
+                "FIRECRACKER binary path `{}` is invalid: {}"
+            )?;
+            if firecracker.boot_info.kernel.is_empty() {
+                return Err(eother!("Guest kernel image for firecracker is empty"));
+            }
+            if firecracker.boot_info.image.is_empty() {
+                return Err(eother!(
+                    "Both guest boot image and initrd for firecracker are empty"
```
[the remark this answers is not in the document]
ananos: actually [it does](https://github.com/firecracker-microvm/firecracker/blob/main/docs/initrd.md), although we haven't added that in kata just yet ;-)

src/libs/kata-types/src/config/hypervisor/firecracker.rs
```diff
@@ -0,0 +1,116 @@
...
+                ));
+            }
+
+            if (firecracker.cpu_info.default_vcpus > 0
+                && firecracker.cpu_info.default_vcpus as u32 > default::MAX_FIRECRACKER_VCPUS)
+                || firecracker.cpu_info.default_maxvcpus > default::MAX_FIRECRACKER_VCPUS
```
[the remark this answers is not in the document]
ananos: the rationale here was to make sure that no vCPU setting is above the maximum number of vCPUs supported by the hypervisor. But I agree, let's just ignore the default_maxvcpus setting, as it is not used.

src/runtime-rs/Makefile
```diff
@@ -164,8 +170,11 @@ DEFMSIZE9P := 8192
 DEFVFIOMODE := guest-kernel
 ##VAR DEFSANDBOXCGROUPONLY=<bool> Default cgroup model
 DEFSANDBOXCGROUPONLY ?= false
+DEFSANDBOXCGROUPONLY_DB ?= true
```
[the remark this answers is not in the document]
ananos: issue is that if we want a separate default option for hypervisors, we should have a per-hypervisor variable.

src/libs/shim-interface/src/lib.rs
```diff
@@ -49,14 +52,18 @@ fn get_uds_with_sid(short_id: &str, path: &str) -> Result<String> {
             "sandbox with the provided prefix {short_id:?} is not found"
         )),
         1 => {
+            // FIXME: jailed case seems to fail
             // One element and only one exists.
-            Ok(format!(
-                "unix://{}",
-                kata_run_path
+            let p = match jailed_path {
```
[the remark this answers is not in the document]
ananos: that's true -- we thought it would be better if all container-related files are inside the jail, but, indeed, we can keep the monitor shim at the generic (non-jailed) dir.

src/runtime-rs/crates/hypervisor/Cargo.toml
```diff
@@ -45,6 +45,9 @@ tempdir = "0.3.7"
 
 [target.'cfg(not(target_arch = "s390x"))'.dependencies]
 dragonball = { path = "../../../dragonball", features = ["atomic-guest-memory", "virtio-vsock", "hotplug", "virtio-blk", "virtio-net", "virtio-fs", "vhost-net", "dbs-upcall", "virtio-mem", "virtio-balloon", "vhost-user-net", "host-device"] }
+dbs-utils = { path = "../../../dragonball/src/dbs_utils" }
```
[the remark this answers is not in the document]
ananos: we only use the MacAddr from dbs_utils -- maybe we could use another crate for that.

src/runtime-rs/crates/hypervisor/src/firecracker/inner.rs
ananos: ack!

lifupan (approved):

PR #8072, opened 2023-09-26
gha: Create workflow to trigger CI with command 

sprt:
This PR makes it so that CI is triggered manually rather than
automatically in order to reduce costs in our Azure subscription.

With these changes, members of the Kata organization will be able to
trigger CI by submitting a review for the PR and including the command
/test (on its own line). Note that this must be part of a PR review
comment, as regular comments will NOT trigger this workflow because of a
GHA limitation: workflows triggered by the issue_comment workflow aren't
run in the PR's context and hence don't automatically add a check to the
PR.

Also note that this removes the need for the ok-to-test label.

Tested working here: https://github.com/sprt/kata-containers/pull/3


.github/workflows/ci-on-comment.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Check is PR
+        if: always()
+        run: |
+          echo '${{ github.event.issue.pull_request }}'
+
+      - name: Check ok-to-test
+        if: always()
+        run: |
+          echo '${{ contains(github.event.issue.pull_request.labels.*.name, 'ok-to-test') }}'
```
fidencio, line 33: I'd drop the need of the `ok-to-test` label.

fidencio (commented):
One thing that worries me a little bit is the commit-hash passed down to the workflows.
I think those will have to be adjusted, togteher with all the usages `github.$something`.

Did you manage to have it tested / triggered in some personal branch?

.github/workflows/ci-on-comment.yaml
sprt: Done

sprt:
@fidencio Tweaked this a bit, this will only work on PR review comments - see commit message.

fidencio (approved):
lgtm, thanks a whole lot @sprt!

sprt:
> IIUC I should comment "/test" my review comment. If a CI job fails then the submitter will need to fix. Then should I approve the PR again to trigger the tests? or they will be executed automatically from the first approval onward?

With this design you'd have to submit another review with `/test` to trigger CI again (doesn't have to be an approval BTW), no more automatic triggering (otherwise that'd be equivalent to the `ok-to-test` label we had).

sprt:
Note that in the future we want to have more granular commands (e.g. `/test-tdx`) and also only rerun failed jobs where possible.

.github/workflows/ci-on-pr-review.yaml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: Check for /test
+        id: test-command
+        run: |
+          comment_body=$(
+          cat <<'END_COMMENT'
+          ${{ github.event.review.body }}
+          END_COMMENT
+          )
+          if echo "$comment_body" | sed 's/\r$//' | grep -xq '^/test$'; then
```
wainersm, line 20: Tested this shell locally:
```
$ comment_body=$(
> cat <<'END_COMMENT'
> LGTM
> /test
> END_COMMENT
> )
$ echo "$comment_body" | sed 's/\r$//' | grep -xq '^/test$'
$ echo $?
0

```

sprt: So working as intended yes? 🙂

.github/workflows/ci-on-pr-review.yaml
```diff
@@ -0,0 +1,37 @@
...
+            echo "is_present=true" >> $GITHUB_OUTPUT
+          fi
+
+  kata-containers-ci-on-pr-review:
```
wainersm, line 24: Will "kata-containers-ci-on-pr-review" display on checkers table? If yes, could be make it shorter?

The jobs names are getting really long, even with big screen their names are truncated:

` Kata Containers CI / kata-containers-ci-on-push / run-metrics-tests / Kata Setup (...`

Perhaps even reduce this workflow name from `Kata Containers CI` to simply `CI`.

wainersm:
> > IIUC I should comment "/test" my review comment. If a CI job fails then the submitter will need to fix. Then should I approve the PR again to trigger the tests? or they will be executed automatically from the first approval onward?
> 
> With this design you'd have to submit another review with `/test` to trigger CI again (doesn't have to be an approval BTW), no more automatic triggering (otherwise that'd be equivalent to the `ok-to-test` label we had).

Cool! I forgot one can send a review comment without approval.

.github/workflows/ci-on-pr-review.yaml
sprt: Agreed. How about we make it look like `CI / on-review / run-metrics-tests /...` ? cc @fidencio 

Also gotta be mindful of what appears here: https://github.com/kata-containers/kata-containers/actions

fidencio: I'm fine with the suggestion!

sprt: Done

sprt (commented):
/test

sprt (commented):
/test

sprt (commented):
/test

wainersm (approved):
Thanks @sprt !

Issue #6691, opened 2023-04-20
runtime-rs: ch: Implement VFIO device handling


jodh-intel:
Update the cloud Hypervisor implementation of [`handle_add_device()`](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs#L41) to allow VFIO devices to be handled.

Also implement `remove_device()` for this device type.

PR #8073, opened 2023-09-27
runtime-rs: Add support for adding vfio device for cloud-hypervisor

amshinde:
This change adds support for adding vfio devices for cloud-hypervisor.

Fixes: #6691

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner.rs
@@ -13,6 +13,7 @@ use kata_types::capabilities::{Capabilities, CapabilityBits};
 use kata_types::config::hypervisor::Hypervisor as HypervisorConfig;
 use kata_types::config::hypervisor::HYPERVISOR_NAME_CH;
 use persist::sandbox_persist::Persist;
+use std::collections::HashMap;
 use std::os::unix::net::UnixStream;
 use tokio::process::Child;
 use tokio::sync::watch::{channel, Receiver, Sender};
@@ -59,6 +60,13 @@ pub struct CloudHypervisorInner {
     // For example, on Intel TDX capable systems with `confidential_guest=true`,
     // this will be set to "tdx".
     pub(crate) guest_protection_to_use: GuestProtection,
+
+    // Store mapping between device-ids created by runtime-rs device manager
+    // and device-ids returned by cloud-hypervisor when the device is added to the VM.
+    //
+    // The cloud-hypervisor device-id is later looked up and used while
+    // removing the device.
+    pub(crate) device_ids: HashMap<String, String>,
 }
 
 const CH_DEFAULT_TIMEOUT_SECS: u32 = 10;
@@ -90,6 +98,7 @@ impl CloudHypervisorInner {
             run_dir: String::default(),
             netns: None,
             pending_devices: vec![],
+            device_ids: HashMap::<String, String>::new(),
             _capabilities: capabilities,
             shutdown_tx: Some(tx),
             shutdown_rx: Some(rx),
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
@@ -6,16 +6,21 @@
 
 use super::inner::CloudHypervisorInner;
 use crate::device::DeviceType;
-use crate::BlockConfig;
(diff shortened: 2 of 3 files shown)


src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -121,6 +124,40 @@ impl CloudHypervisorInner {
         Ok(())
     }
 
+    async fn handle_vfio_device(&mut self, device: &VfioDevice) -> Result<()> {
+        // A device with multi-funtions, or a IOMMU group with one more
+        // devices, the Primary device is selected to be passed to VM.
+        // And the the first one is Primary device.
+        // safe here, devices is not empty.
+        let primary_device = device.devices.first().unwrap().clone();
```
jodh-intel, line 132: Could you use `?` rather than `.unwrap()` as the function returns a `Result`?

jodh-intel:
Better update that fixes issue number too ;)

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
amshinde: Fixed

amshinde:
@jodh-intel I have fixed teh issue number:) 
I have also added to the PR functionality to remove vfio device. Please take a look.

amshinde:
Also note, for vfio device removal to work, a change is required to the device manager, as currently thepredicted  pcie path for the devices inside in the VM does not seem to work for cloud-hypervisor. I am working on fixing this and have tested this change with the agent. Will propose that as part of another PR.

src/runtime-rs/crates/hypervisor/src/ch/inner.rs
```diff
@@ -59,6 +60,8 @@ pub struct CloudHypervisorInner {
     // For example, on Intel TDX capable systems with `confidential_guest=true`,
     // this will be set to "tdx".
     pub(crate) guest_protection_to_use: GuestProtection,
+
+    pub(crate) device_ids: HashMap<String, String>,
```
jodh-intel, line 64: Might be worth adding a comment here explaining what the two string values are.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -72,15 +77,37 @@ impl CloudHypervisorInner {
...
+            self.add_device(dev).await.context("add_device")?;
+        }
+
         Ok(())
     }
 
+    pub(crate) async fn remove_device(&mut self, device: DeviceType) -> Result<()> {
+        match device {
+            DeviceType::Vfio(vfiodev) => self.remove_vfio_device(&vfiodev).await,
+            _ => Ok(()),
```
jodh-intel, line 107: Shouldn't this generate an error, or at least a log warning?

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -137,7 +164,85 @@ impl CloudHypervisorInner {
...
-    async fn handle_block_device(&mut self, cfg: BlockConfig) -> Result<()> {
+    async fn handle_vfio_device(&mut self, device: &VfioDevice) -> Result<()> {
+        // A device with multi-funtions, or a IOMMU group with one more
+        // devices, the Primary device is selected to be passed to VM.
+        // And the the first one is Primary device.
+        // safe here, devices is not empty.
+        let primary_device = device
+            .devices
+            .first()
+            .ok_or(anyhow!("Primary device list empty for vfio device"))?;
```
jodh-intel, line 175: Is it worth adding `device` to the error here?

```suggestion
            .ok_or(anyhow!("Primary device list empty for vfio device {:?}", device))?;
```

jodh-intel (commented):
Thanks @amshinde - a few comments...

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
amshinde: I am returning an error here, as if device removal is not implemented, this should not error out at the moment. For certain devices, removal may not be even required, but this will have to be evaluated later on a case by case basis.

amshinde:
@jodh-intel Comments addressed. Please take a look.

cmaf (approved):
Thanks @amshinde LGTM!

Issue #8094, opened 2023-10-02
Configure kernel to enable "native overlay diff" by default


skaegi:
Generally speaking it's a good idea to have your kernel configured to support native overlay diff to make it faster to build images in kata. There is a trade-off here between metacopy (which can speed-up big copies) and native overlay diff but I think in this case we should err towards typical defaults as we're getting questions about why native overlay diff is disabled in our micro VMs.

Kata is currently defining both `CONFIG_OVERLAY_FS_METACOPY` and `CONFIG_OVERLAY_FS_REDIRECT_DIR` which is what prevent native overlay diff. I'm at least partly to blame here as I suggested adding metacopy in #1075 but we've now removed both config in our custom kernel and would suggest that is a better default as it matches what is the typical setting in Host OSes for managed K8s and well... performs better.

skaegi:
In retrospect @c3d is right -- https://github.com/kata-containers/kata-containers/issues/1075#issuecomment-789773023 -- you're better to pass `overlay=true` as a mount opt vs. changing the default behavior.

skaegi:
I'm going to close and withdraw this for now. I've done a ton of testing and in the tradeoff between metacopy and native diff am consistently seeing more benefit from metacopy ... and this in a ci system where native diff should really help.

Still makes sense to use mount options to allow native diff -- just not a good default.

Issue #1075, opened 2020-11-03
Update Kernel config for overlayfs


skaegi:
We package a recent kernel and use overlayfs inside of Kata and have found you really need to update the `fs` config to include...

`CONFIG_OVERLAY_FS_METACOPY=y`
... to have reasonable performance for recursive chmod and similar calls.

`CONFIG_OVERLAY_FS_XINO_AUTO=y`
... less critical but set primarily because it is recommended for POSIX compliance

See... https://github.com/torvalds/linux/blob/master/fs/overlayfs/Kconfig
and https://github.com/torvalds/linux/blob/master/Documentation/filesystems/overlayfs.rst

These settings have been present since 4.19 so it might be time to make them mainstream.

egernst:
Thanks @skaegi! Are you running with these options today?

Seems reasonable to me -- you interested in submitting a PR?

skaegi:
Yes and have been for some time with 5.8.x and now 5.9.x. Sure let me create a PR.

fidencio:
@skaegi, thanks for raising the issue.  I've opened a #1384 that should cover your recommendations.

skaegi:
Thanks @fidencio -- I never got back to this but yes you've covered it.

c3d:
@fidencio You fix look good, but I think it might be better to actually pass that as mount options to overlayfs in case you pick up a kernel that does not have the option set (e.g., at random, a Fedora kernel).

PR #8095, opened 2023-10-02
Configure kernel to allow native overlay diff

skaegi:
For native overlay diff to work both `CONFIG_OVERLAY_FS_REDIRECT_DIR` and `CONFIG_OVERLAY_FS_METACOPY` must not be configured. This is the typical default in Host OSes used to run K8s and noticeably improves the performance of docker build.

Fixes: #8094

Signed-off-by: Simon Kaegi <simon.kaegi@gmail.com>

### DIFF
--- a/tools/packaging/kernel/configs/fragments/common/fs.conf
+++ b/tools/packaging/kernel/configs/fragments/common/fs.conf
@@ -52,8 +52,6 @@ CONFIG_AIO=y
 # Docker in Docker support requires overlay
 CONFIG_OVERLAY_FS=y
 CONFIG_OVERLAY_FS_INDEX=y
-CONFIG_OVERLAY_FS_REDIRECT_DIR=y
-CONFIG_OVERLAY_FS_METACOPY=y
 CONFIG_OVERLAY_FS_XINO_AUTO=y
 
 # virtio-fs driver support:


tools/packaging/kernel/configs/fragments/common/fs.conf
```diff
@@ -52,8 +52,6 @@ CONFIG_AIO=y
 # Docker in Docker support requires overlay
```
fidencio, line 52: @skaegi, just reading the comment here makes me wonder ... does the change break docker in docker support?

tools/packaging/kernel/configs/fragments/common/fs.conf
```diff
@@ -52,8 +52,6 @@ CONFIG_AIO=y
...
 CONFIG_OVERLAY_FS=y
 CONFIG_OVERLAY_FS_INDEX=y
-CONFIG_OVERLAY_FS_REDIRECT_DIR=y
```
amshinde, line 55: @skaegi This was added to support docker-in-docker use case. According to the docs, you can still disable this with `"redirect_dir=off" ` module option. Maybe you can consider passing this on the kernel command line and see if that works out.

tools/packaging/kernel/configs/fragments/common/fs.conf
skaegi: no -- both ways work with docker-in-docker
The difference is that when you do a `docker info` for the Storage driver.
```
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
```  
you'll either see `Native Overlay Diff` or `Using metacopy` set to `true` but not both. Native Overlay Diff is "much" more valuable for performance in typical use.

tools/packaging/kernel/configs/fragments/common/fs.conf
skaegi: Looking back at #1075 I only suggested `CONFIG_OVERLAY_FS_METACOPY` be added. I'm not sure why `CONFIG_OVERLAY_FS_REDIRECT_DIR` would ever be required for dind. We certainly don't use it in our kernel and do literally millions of dind builds a month.

tools/packaging/kernel/configs/fragments/common/fs.conf
skaegi: ah ok -- rebuilt the kernel and saw Native Overlay Diff again disabled. More info here -- https://github.com/torvalds/linux/commit/d47748e5ae5af6572e520cc9767bbe70c22ea498#diff-315c61b86e39b9b47f4ab0cd9efc2467

So... both need to be removed to enable Native Overlay Diff.

tools/packaging/kernel/configs/fragments/common/fs.conf
skaegi: More info here -- https://dev.to/jarekprzygodzki/a-curious-case-of-slow-docker-image-builds-2o7k

fidencio:
@skaegi, your change looks good to me, but the tests will fail as you also need to bump the kernel config version here: https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kernel/kata_config_version

Please, do the bump, and we'll get the CI running and this PR merged.

tools/packaging/kernel/configs/fragments/common/fs.conf
amshinde: Thanks for the links @skaegi. 
This is not related to Kata itself or `dind` usecase, but I do recall running into an issue where files in renamed directories were not correctly propogated to the upper layer in overlay file system due to `CONFIG_OVERLAY_FS_REDIRECT_DIR` not being enabled in the host system on a particular linux distro. Its been a while though, and I am not aware of changes since then in this space to not require this option anymore nor dabbled in dinds recently.
If you feel it is safe to turn off this option, lets do that. 

I just happened to take a look at the kernel change related to this kernel option: https://github.com/torvalds/linux/commit/d47748e5ae5af6572e520cc9767bbe70c22ea498#diff-315c61b86e39b9b47f4ab0cd9efc2467
That commit suggests that for metacopy to be turned on for overlay file system, it depends on `redirect_dir` to be turned on. Referring to this line in the commit message:  `This patch makes metacopy=on imply redirect_dir=on.`

skaegi:
I'm going to withdraw this for now. After doing a ton of testing the current kernel config seems to perform better than what I'm proposing for more general cases. I can use mount opts to support native diff for my special case.

PR #8100, opened 2023-10-03
kata-deploy: Build kata-agent as we build all the other components

fidencio:
For now we're not using the result of the build, but we should do it for the rootfs-build, and also for the tracing tests that will be migrated to GHA.

In order to test this PR, please, do, from the topdir of the kata-containers repo:
* `make agent-tarball`
* `make agent-opa-tarball`


tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
```diff
@@ -106,6 +107,7 @@ docker run \
 	--env TARGET_BRANCH="${TARGET_BRANCH}" \
 	--env BUILDER_REGISTRY="${BUILDER_REGISTRY}" \
 	--env PUSH_TO_REGISTRY="${PUSH_TO_REGISTRY}" \
+	--env AGENT_CONTAINER_BUILDER="${INITRAMFS_CONTAINER_BUILDER}" \
```
stevenhorsman, line 111: I'm guessing this should be:
```
--env AGENT_CONTAINER_BUILDER="${AGENT_CONTAINER_BUILDER}" \
```

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -625,6 +628,32 @@ install_ovmf_sev() {
 	install_ovmf "sev" "edk2-sev.tar.gz"
 }
 
+install_agent_helper() {
+	agent_policy="${1:-no}"
+
+	latest_artefact="$(git log -1 --pretty=format:"%h" ${repo_root_dir}/src/agent)"
+	latest_builder_image="$(get_agent_image_name)"
+
+	install_cached_tarball_component \
```
stevenhorsman, line 637: I might have missed something, but does this `agent_policy` value need to change the `install_cached_tarball_component` to differentiate between the cached agent with and without policy?

tools/packaging/static-build/agent/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2023 Intel
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM alpine:3.18
+ARG GO_TOOLCHAIN
```
stevenhorsman, line 6: nit: Is `GO_TOOLCHAIN` needed?

stevenhorsman (commented):
Caching the agent will be super useful as it's the longest part of the kata build process in my experience, especially with the extra CoCo pieces. A few questions/comments from me:

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
fidencio: We should just use the `build_target` as the component name, I think.

tools/packaging/static-build/agent/Dockerfile
fidencio: Nops. :-)

stevenhorsman (approved):
Thanks for the fixes. Code LGTM and I ran both commands locally and they worked

Issue #8105, opened 2023-10-03
ci: k0s related tests are failing


fidencio:
The tests are failing when setting up k0s, and that happens because we download a kubectl binary matching the kubernetes version k0s is using, and we do that by:
```
sudo k0s kubectl version --short 2>/dev/null | grep "Client Version" | sed -e 's/Client Version: //'
```

With k8s 1.28, `kubectl version --short` has been removed, leading us to an empty stringm causing then the error in the CI.

PR #8106, opened 2023-10-03
gha: Fix k0s deployment

fidencio:
The tests are failing when setting up k0s, and that happens because we download a kubectl binary matching the kubernetes version k0s is using, and we do that by:
```
sudo k0s kubectl version --short 2>/dev/null | ...
```

With kubectl 1.28, which is now the default on k0s, `kubectl version --short` has been removed, leading us to an empty stringm causing then the error in the CI.

Fixes: #8105


stevenhorsman (approved):
LGTM

GabyCT (approved):

amshinde (approved):

Issue #8109, opened 2023-10-04
gha: arm64: Ensure the builder is arm64-builder


fidencio:
Otherwise we'll use any arm64 machine that's added as a runner, and
whenever new machines are added those may end up being only used for
running some specific set of the tests.

PR #8110, opened 2023-10-04
gha: arm64: Ensure the builder is arm64-builder

fidencio:
Otherwise we'll use any arm64 machine that's added as a runner, and whenever new machines are added those may end up being only used for running some specific set of the tests.

Fixes: #8109

image


zvonkok (approved):
LGTM, Muito obrigado!!

GabyCT (approved):

PR #7779, opened 2023-08-29
Backport 3.2 | local-build: Remove GID before creating group

alex-matei:
docker install now creates a group with gid 999 which happens to match what we need to get docker-in-docker to work. Remove the group first as we don't need it.

Fixes: #7726
Signed-off-by: Jeremi Piotrowski <jpiotrowski@microsoft.com>
(cherry picked from commit 3b881fb)

EDIT: I added a commit to fix the dragonball build that was failing.

Fixes: #7888

PR #8143, opened 2023-10-06
do not merge | check the current state of stable 3.2

fidencio:


fidencio:
From what I can see here, with the tests that run, we're basically running the checks from this branch, which means we **do** need to backport all the tests related stuff.

alex-matei:
The make check will fail without this pull request: #7779 . I added Cargo.lock in a commit in that PR to fix #7888

fidencio:
> The make check will fail without this pull request: #7779 . I added Cargo.lock in a commit in that PR to fix #7888

@alex-matei, thanks for pointing this out, but here we're simply testing whether we actually need to backport all the GitHub actions related stuff from the main branch, and the tests running here shows us that yes, we do.

I'm closing this one, going ahead with the tests, and we'll backport the needed fixes as the following step.

PR #8146, opened 2023-10-06
do-not-merge: check the current state of stable-3.2-test branch

fidencio:


fidencio:
From the static-checks:
```
Already on 'main'
Your branch is up to date with 'origin/main'.
```

Why does it think we should be on main?

Issue #8152, opened 2023-10-06
runtime-rs: ch: Add extra TDX check if Cloud Hypervisor advertises features


jodh-intel:
If you attempt to create a container (a TD) on a TDX system using a custom build of Cloud Hypervisor (CH) that was not built with the `tdx` CH feature, Kata will report the following CH error:

```
ApiError(VmBoot(InvalidPayload))
```

This is too cryptic! [Newer versions of CH now report their build-time features in the ping API response message](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/5820) so Kata can use that, if available, to detect this scenario and generate a user-friendly error message instead.

PR #8149, opened 2023-10-06
runtime-rs: ch: Detect Intel TDX version

jodh-intel:
Improve the `GuestProtection` handling to detect the version of Intel TDX available.

The TDX version is now logged by the Cloud Hypervisor driver.

Fixes: #8147.

PR #8155, opened 2023-10-06
runtime-rs: ch: Add TDX CH features check

jodh-intel:
If you attempt to create a container (a TD) on a TDX system using a
custom build of Cloud Hypervisor (CH) that was not built with the `tdx`
CH feature, Kata will report the following, somewhat cryptic, CH error:

```
ApiError(VmBoot(InvalidPayload))
```

Newer versions of CH now report their build-time features in the ping
API response message so we now use that, if available, to detect this
scenario and generate a user-friendly error message instead.

Also adds a commit to check for TDX as this is the only expected guest
protection for CH for now.

Fixes: #8152.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
@@ -23,6 +23,8 @@ use kata_sys_util::protection::{available_guest_protection, GuestProtection};
 use kata_types::capabilities::{Capabilities, CapabilityBits};
 use kata_types::config::default::DEFAULT_CH_ROOTFS_TYPE;
 use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
 use std::convert::TryFrom;
 use std::fs::create_dir_all;
 use std::os::unix::net::UnixStream;
@@ -41,6 +43,20 @@ const CH_NAME: &str = "cloud-hypervisor";
 
 /// Number of milliseconds to wait before retrying a CH operation.
 const CH_POLL_TIME_MS: u64 = 50;
+
+// The name of the CH JSON key for the build-time features list.
+const CH_FEATURES_KEY: &str = "features";
+
+// The name of the CH build-time feature for Intel TDX.
+const CH_FEATURE_TDX: &str = "tdx";
+
+#[derive(Clone, Deserialize, Serialize)]
+pub struct VmmPingResponse {
+    pub build_version: String,
+    pub version: String,
+    pub pid: i64,
+    pub features: Vec<String>,
+}
 
 #[derive(thiserror::Error, Debug, PartialEq)]
 pub enum GuestProtectionError {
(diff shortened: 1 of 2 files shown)


jodh-intel:
This will conflict with #8149.

Ji-Xinyou (approved):
Thanks! This PR LGTM

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -491,7 +495,13 @@ impl CloudHypervisorInner {
...
 
-            info!(sl!(), "guest protection available and requested"; "guest-protection" => protection.to_string());
+            if protection == GuestProtection::Tdx {
+                info!(sl!(), "guest protection available and requested"; "guest-protection" => protection.to_string());
+            } else {
+                return Err(anyhow!(GuestProtectionError::ExpectedTDXProtection(
+                    protection
+                )));
+            }
         } else if protection == GuestProtection::Tdx {
```
amshinde, line 560: We are checking for `GuestProtection::Tdx` here and on line 498 as well. Similarly the check for `confidential_guest` is made on line 490 and then 495. Its a bit confusing on how the checks are made. I think we can organize this better with checking for  confidential_guest first and then branching out from there:

```
if  confidential_guest {
         if protection == GuestProtection::NoProtection {
         } 
         else if protection == GuestProtection::Tdx {
          } 
         else { 
          } 
 } else {
     // We could omit this branch altogether as protection value does not really matter if confidential case if not enforced
}
```

jodh-intel: @amshinde - Agreed. Branch updated.

cmaf (approved):

jodh-intel:
@jepio - Any idea what's going on with the VFIO CI? It's failure is blocking a number of PRs from landing.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -492,6 +496,10 @@ impl CloudHypervisorInner {
                 return Err(anyhow!(GuestProtectionError::NoProtectionAvailable));
             } else if let GuestProtection::Tdx(_) = protection {
                 info!(sl!(), "guest protection available and requested"; "guest-protection" => protection.to_string());
+            } else {
+                return Err(anyhow!(GuestProtectionError::ExpectedTDXProtection(
+                    protection
+                )));
```
amshinde, line 557: Is the check on line 506 correct? Isnt the condition checking for TDX and then erroring out about TDX not being used. Let me know if I misunderstood. 
Also, I feel that when confidential_guest is not requested, we should not really error out on the protection value. We could just log the protection value. In fact, in case confidential_guest is not requested, it is really not required to compute the protection value at all. What do you think?

Also, do you think you can squash the first and second commit to make this easier to review?

jodh-intel: Hi @amshinde - Do you mean line 556? This part of the diff does look a little bit odd as a result of inverting the logic (at your request ;)

Here's a summary of the conditions the code and unit tests handle:

| `confidential_guest` | available protection | result | notes |
|-|-|-|-|
| `false` | none | Ok | Log a message that no protection available |
| `false` | TDX | Error | CH requires TDX protection be used if available |
| `false` | non-TDX | Ok | Log available but unused protection |
| `true` | none | Error | The user requested protection, but none available, so we must fail |
| `true` | TDX | Ok | Log a message that protection available and requested |
| `true` | non-TDX | Error | The CH driver is only currently tested with TDX |

jodh-intel: @amshinde - I've now combined the first two commits as requested :smile:

amshinde (approved):

PR #8158, opened 2023-10-06
gha: Trigger CI on `issue_comment`

sprt:
This is a rework of #8072 - since that one triggered CI via the `pull_request_review` event, it didn't have access to the repo secrets and thus couldn't create AKS clusters and so on.

This PR is similar with a few differences:

* The trigger event is `issue_comment`, meaning that the new `/test-gha` command works on regular PR comments, not PR review comments.
* I use the `gh workflow run` command to launch `ci.yaml`, rather than a `uses` block. Because `issue_comment` workflows also don't have access to the repo secrets, this launches the downstream jobs via the `workflow_dispatch` event, which gives access to secrets.
  * Upside: `ci.yaml` and downstream jobs are now executed from the base branch, making future YAML changes much easier to test.
  * **DOWNSIDE**: PR checks aren't created automatically for workflows run with `gh workflow run` - that's something we have to do manually via the API. Here I created a single "Kata Containers CI" check that accounts for all CI jobs, thus we lose granularity. In a future PR, we can create a wrapper workflow that takes a target workflow as argument and creates a check via the API.

Note that I'm keeping `ci-on-push.yaml` for now and will remove it in another PR once we've validated this works. One weird thing that I've seen in #8072 is that `github.event.issue.author_association` returns `CONTRIBUTOR` (instead of `MEMBER`) for me on this repo.

Tested working here: https://github.com/sprt/kata-containers/pull/5

### DIFF
--- a/.github/workflows/ci-on-comment.yaml
+++ b/.github/workflows/ci-on-comment.yaml
@@ -1,0 +1,124 @@
+name: CI
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  kata-containers-ci-on-comment:
+    if: github.event.issue.pull_request
+    permissions:
+      actions: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for /test-gha
+        id: test-command
+        run: |
+          comment_body=$(
+          cat <<'END_COMMENT'
+          ${{ github.event.comment.body }}
+          END_COMMENT
+          )
+          if echo "$comment_body" | sed 's/\r$//' | grep -xq '^/test-gha$'; then
+            echo "is_present=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Fetch PR info
+        id: pr-info
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          pr_data="$(gh api repos/$GITHUB_REPOSITORY/pulls/${{ github.event.issue.number }})"
+          echo "base_ref=$(jq -r '.base.ref' <<< "$pr_data")" >> $GITHUB_OUTPUT
+          echo "head_ref=$(jq -r '.head.ref' <<< "$pr_data")" >> $GITHUB_OUTPUT
+          echo "head_sha=$(jq -r '.head.sha' <<< "$pr_data")" >> $GITHUB_OUTPUT
+
+      - name: Debug
+        run: |
+          echo "Is member/owner: ${{ contains(fromJSON('["MEMBER", "OWNER"]'), github.event.comment.author_association) }}"
+          echo "${{ github.event.comment.author_association }}"
+          echo "Is against main/stable-*: ${{ steps.pr-info.outputs.base_ref == 'main' || startsWith(steps.pr-info.outputs.base_ref, 'stable-') }}"
+          echo "${{ steps.pr-info.outputs.base_ref }}"
+          echo "Head commit: ${{ steps.pr-info.outputs.head_sha }}"
+          echo "Has /test-gha: ${{ steps.test-command.outputs.is_present == 'true' }}"
+
+      - name: Start CI
+        id: ci
+        if: |
+          contains(fromJSON('["MEMBER", "OWNER"]'), github.event.issue.author_association) &&
+          (steps.pr-info.outputs.base_ref == 'main' ||
+            startsWith(steps.pr-info.outputs.base_ref, 'stable-')) &&
+          steps.test-command.outputs.is_present == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "started-at=$(date +"%Y-%m-%dT%H:%M:%S%z")" >> $GITHUB_OUTPUT
+
+          gh workflow run ci.yaml \
+            -R "$GITHUB_REPOSITORY" \
+            -r ${{ steps.pr-info.outputs.head_ref }} \
+            -f commit-hash=${{ steps.pr-info.outputs.head_sha }} \
+            -f pr-number=${{ github.event.issue.number }} \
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-on-comment.yaml
```diff
@@ -0,0 +1,123 @@
+name: CI
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  kata-containers-ci-on-comment:
```
wainersm, line 7: Disclaimer: my knowledge with gh actions is limited, I've often forgot some details and have to consult the docs.

According to the docs https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment an `issue_comment` is triggered for comments either in an issue or pull request.

Here we are interested on comments in pull requests. So could we conditionally run the `kata-containers-ci-on-comment` only and only if `github.event.issue.pull_request` to avoid unnecessary runs? If yes, then the `if` clause in `Start CI` should be refactored.

BTW, I think this will fix a bug...because the way it is implemented, IIUC, if I a send a comment to an issue then all the steps, except for `Start CI`, are going to run. And it does not make sense to run, say, `Fetch CI run ID` if `Start CI` is not running.

.github/workflows/ci-on-comment.yaml
```diff
@@ -0,0 +1,123 @@
...
+          END_COMMENT
+          )
+          if echo "$comment_body" | sed 's/\r$//' | grep -xq '^/test-gha$'; then
+            echo "is_present=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Fetch PR info
+        id: pr-info
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
wainersm, line 25: On this PR's cover letter (which I loved!) you wrote `Because issue_comment workflows also don't have access to the repo secrets (...)`. So how does it has access to `secrets` here?

I couldn't find nothing about secrets on issue_comment doc: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment . But it says `This event will only trigger a workflow run if the workflow file is on the default branch` and other events that run on default branch (e.g. pull_request_target) have access to secrets...so perhaps it does as well but it is not documented?

Another possibility is that it works in your fork only. I've seen cases where this "event X hasn't access to secrets" rule simply does not apply when I run the workflow in a forked repository. 

I will pause my review here because the answers of those questions can change the whole thing.

.github/workflows/ci-on-comment.yaml
sprt: You're 100% right, will put a condition on the entire job!

.github/workflows/ci-on-comment.yaml
sprt: Hmm you do bring up a good point. I did some more digging here and I'm confused now.

The easy stuff first: from [1] and [2], `GITHUB_TOKEN` is a special secret that's always available (but it might not always have full permissions, although for this step we only need read permissions so we're good).

So for this step we're good anyway, since we're only reading. But for the `gh workflow run` step things get murky.

I'm not sure anymore whether `issue_comment` has access to secrets: it makes sense that `pull_request_review_comment` doesn't as that runs in the PR context, and I thought since `issue_comment` runs in the default branch context that it does have access (and nothing in the docs seem to indicate it can't access secrets).

However from [2] `With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.` - so whether it has access to secrets depends on whether `issue_comment` counts as "triggered from a forked repository".

Even then, the above isn't a problem as long as the underlying `ci.yaml` workflow has access to secrets. However, from [1] ("Permissions for the GITHUB_TOKEN") the maximum access for PRs from public forks for the `actions` scope is `read`. So again, this depends on whether `issue_comment` counts as public fork PR.

With all this, I'm not sure this is at all implementable. I suggest to add a manual permission override in the YAML and then merge this and see how far we get.

[1] https://docs.github.com/en/actions/security-guides/automatic-token-authentication
[2] https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions

.github/workflows/ci-on-comment.yaml
sprt: Scoped the condition to entire job, PTAL

.github/workflows/ci-on-comment.yaml
sprt: Added a `permissions` block to the job YAML, PTAL

sprt: To conclude, as you mentioned, my hope is that `issue_comment` does allow us to call `gh workflow run`. Ideally even, `issue_comment` has access to all secrets and we can just use a `uses` block.

.github/workflows/ci-on-comment.yaml
wainersm: Cool, thanks!

.github/workflows/ci-on-comment.yaml
```diff
@@ -0,0 +1,124 @@
...
+            -r ${{ steps.pr-info.outputs.head_ref }} \
+            -f commit-hash=${{ steps.pr-info.outputs.head_sha }} \
+            -f pr-number=${{ github.event.issue.number }} \
+            -f tag=${{ github.event.issue.number }}-${{ steps.pr-info.outputs.head_sha }} \
+            -f target-branch=${{ steps.pr-info.outputs.base_ref }}
+
+      # NOTE: There's a potential race condition here if another run was started
+      # right after the previous step. In any case, the concurrency directive of
+      # ci.yaml ensures that only one run prevails.
+      - name: Fetch CI run ID
```
wainersm, line 68: The `Start CI` will conditionally not run, so `Fetch CI run ID` and following steps should not too.

Issue #8172, opened 2023-10-09
CCv0: GPU support not working with Kata remote hypervisor


bpradipt:
The guest hook enablement is missing in Kata remote hypervisor, thereby preventing (Nvidia) GPUs to work with Kata remote hypervisor.
Nvidia requires guest prestart hook to work.

PR #8173, opened 2023-10-09
config: Enable guestHook for remote hyp

bpradipt:
The guesthook config was missing which prevented handling of GPUs with remote hypervisor

Fixes: #8172


stevenhorsman:
The ubuntu tests are failing on stability checks that have been migrated to the kata-containers repo in `main`, so not supported in the test package anymore, so I think we are fine to ignore this failure

PR #8174, opened 2023-10-09
ci: Re-add tracing tests and move docker/nerdctl to the basic-ci-amd64.yaml file

fidencio:
We're just bringing back the tracing tests after the breakage was solved, and also taking the advantage to move the docker / nerdctl tests to the basic-ci-amd64.yaml file.


bergwolf (approved):
lgtm!

tests/common.bash
```diff
@@ -547,6 +547,24 @@ EOF
 	sudo systemctl enable --now crio
 }
 
+function install_docker() {
```
jodh-intel, line 550: Now that https://github.com/kata-containers/kata-containers/issues/8357 has landed, could we use `kata-manager` instead here?

fidencio: No, it doesn't make sense to change to using kata-manager as part of this series that's just re-arranging things.

We need, though, an issue to start using kata-manager to install docker / nerdctl etc, and then replace the containerd / ctr / docker / nerctl deployment throughout our CI.

Doing this here will just make things not consistent, and is out of the scope of this PR.

jodh-intel (approved):
Thanks @fidencio!

lgtm

PR #8195, opened 2023-10-10
versions: bump td-shim version

fidencio:
We're doing so in order to avoid errors like the ones from here: https://github.com/kata-containers/kata-containers/actions/runs/6468115049/job/17559485246

I was able to locally build it, but as we're not testing CLH + TDX lately, running CI on this will be useless.


stevenhorsman (approved):
LGTM. Thanks

Issue #8200, opened 2023-10-11
release: tag_repos: Stop tagging the `tests` repo


fidencio:
As we've moved all the tests to the `kata-containers` repo, the `tests`
repo will become a read-only repo.

PR #8201, opened 2023-10-11
release: tag_repos: Stop tagging the `tests` repo

fidencio:
As we've moved all the tests to the `kata-containers` repo, the `tests` repo will become a read-only repo.

Fixes: #8200


tools/packaging/scripts/lib.sh
```diff
@@ -29,19 +27,8 @@ ARCH=${ARCH:-$(uname -m)}
...
-		return
-	fi
-
-	go get -d -u "$tests_repo" || true
-}
-
 install_yq() {
-	clone_tests_repo
-	pushd "$tests_repo_dir"
+	pushd ${repo_root_dir}
```
gkurz, line 31:
```suggestion
	pushd "${repo_root_dir}"
```
in case there's a space in some path ?

fidencio: I don't think thers is, but better be safe than sorry.

gkurz (approved):
LGTM. Thanks @fidencio !

stevenhorsman (approved):
LGTM. Thanks!

PR #8205, opened 2023-10-11
tests: encode policy file during test

danmihai1:
Encode policy file during test - easier to understand than hard-coding the encoded file contents.

Fixes: #8214


jepio (approved):

fidencio (approved):
lgtm, thanks @danmihai1!

Issue #8216, opened 2023-10-12
Cannot run test twice because of permission denied


beraldoleal:
# Description of problem

Trying to run integration tests locally twice is broken.

1. Create a vm:
 $ kcli create vm -i ubuntu2304 -P disks=[60] -P numcpus=2 -P memory=8192 -P cpumodel=host-passthrough debug-PR8061
2. connect to the machine
3. clone kata-containers
4. Download kata-tarball from GH artifacts inside kata-containers/kata-artifacts
5. Install dependencies:
  - tests/integration/cri-containerd/gha-run.sh install-dependencies
6. Install kata:
  - tests/integration/cri-containerd/gha-run.sh install-kata kata-artifacts
7. Run the test twice:
  - tests/integration/cri-containerd/gha-run.sh run
  - tests/integration/cri-containerd/gha-run.sh run

# Expected result
No additional error, except the test failing itself.
 
# Actual result

```
bash tests/integration/cri-containerd/gha-run.sh run
[gha-run.sh:60] INFO: Running cri-containerd tests using qemu hypervisor
[integration-tests.sh:435] INFO: Stop crio service
[integration-tests.sh:438] INFO: Stop containerd service
~/kata-containers/containerd ~/kata-containers
+ clean
rm: cannot remove 'bin/ctr': Permission denied
rm: cannot remove 'bin/containerd': Permission denied
rm: cannot remove 'bin/containerd-stress': Permission denied
rm: cannot remove 'bin/containerd-shim': Permission denied
rm: cannot remove 'bin/containerd-shim-runc-v1': Permission denied
rm: cannot remove 'bin/containerd-shim-runc-v2': Permission denied
make: *** [Makefile:367: clean] Error 1

```
# Further information

(replace this text with any extra information you think might be useful)

PR #8217, opened 2023-10-12
tests: fixes permission denied when running test

beraldoleal:
After running cri-containerd/integration-tests twice we receive permission denied during containerd clean.

Fixes: #8216


wainersm (approved):
Good catch @beraldoleal !

amshinde (approved):

Issue #8219, opened 2023-10-13
CCv0: conflicting behavior of CCv0's agent-config.toml and main branch's default agent policy


danmihai1:
**Which feature do you think can be improved?**

rootfs.sh

**How can it be improved?**

The main branch doesn't have yet support for creating agent-config.toml from rootfs.sh. However, rootfs.sh from the main
branch:

- Creates a default agent policy file that unlike agent-config.toml allows access to all the kata agent endpoints, including SetPolicy
  and ExecProcess.

- Is expected to deprecate the endpoint restrictions implemented based on agent-config.toml.

This issue is a reminder to review the default policy file behavior when these two versions of rootfs.sh will eventually be merged.

PR #8220, opened 2023-10-13
rootfs: add default agent policy file

danmihai1:
rootfs: add default agent policy file

Restricting access to agent endpoints using agent-config.toml is expected to be deprecated in the main branch. Therefore, in
preparation of merging this script with its main branch version, install default settings for main branch's kata-opa service.

coco-default.rego blocks access to the same kata agent endpoints that are blocked by agent-config.toml. For additional information, search for "default-policy.rego" in main branch's rootfs.sh.

Fixes: #8219


tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -775,6 +777,15 @@ detect_host_distro()
 	esac
 }
 
+set_default_policy()
+{
+	# Use the policy file providing the same agent enpoint restrictions as agent-config.toml
+	default_policy="${ROOTFS_DIR}/etc/kata-opa/default-policy.rego"
```
stevenhorsman, line 783: Hey Dan, I know if won't get picked up until we move it to `main`, but do you think it's worth adding a `coco-default.rego` file in this PR that matches the current SEV agent-config set-up: https://github.com/kata-containers/kata-containers/blob/CCv0/tools/osbuilder/rootfs-builder/agent-config.toml.in - so execProcess, reseed and setpolicy requests would be blocked out of the box. Apologies if I'm misunderstood the intention behind this though!

danmihai1: We already have allow-all.rego and allow-all-except-exec-process.rego in the main branch. So, my intention was to add coco-default.rego at the same localtion, in the main branch. But your idea might be better than mine - I will think a bit more about it and let you know.

Thanks Steve!

danmihai1:
@fitzthum , please speak up in case you have better ideas to mitigate potential policy issues when merging CCv0 with main.

Thanks!

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -667,6 +667,20 @@ EOF
...
+	# kata-opa service. coco-default.rego blocks access to the same kata agent
+	# endpoints that are blocked by agent-config.toml. For additional information,
+	# search for "default-policy.rego" in main branch's rootfs.sh.
+	local kata_opa_in_dir="${script_dir}/../../../src/kata-opa"
+	local opa_settings_dir="/etc/kata-opa"
+	local policy_file="coco-default.rego"
+	local policy_dir="${ROOTFS_DIR}/${opa_settings_dir}"
+	mkdir -p "${policy_dir}"
+	install -D -o root -g root -m 0644 "${kata_opa_in_dir}/${policy_file}" -T "${policy_dir}/${policy_file}"
+	ln -sf "${policy_file}" "${policy_dir}/default-policy.rego"
```
stevenhorsman, line 682: So just to double check I understand this, as part of the merge to main for endpoint restriction, we'll want to take over the coco-deafult.rego file and then for the TEE_PLATFORM="SEV/SNP" (at least and probably SE & TDX too), we'll want to update the existing logic in root.sh to link from `coco-default.rego` instead of `allow-all.rego`?

danmihai1: Good to know about TEE_PLATFORM - I wasn't aware of it!

When we'll merge the two versions of rootfs.sh, I hope git will complain about a merge conflict - since these few lines of the script are almost identical but not quite across the two branches. That conflict will force us to think about which default policy we'll want on which platforms:

- If we were to merge today, I would enable the new coco-default.rego as default policy for any TEE_PLATFORM, and continue to use all-allowed.rego for all non-TEE platforms that support Policy. 

- But by the time of the merge it's possible that we'll come up with better default policies than coco-default.rego. The current coco-default.rego is safe enough for the short term but it's not flexible enough for the future.

stevenhorsman: Just to clarify - `TEE_PLATFORM` hasn't been written yet, but is the current best plan we have for getting the attestation-agent variants in for the different architectures.

> If we were to merge today, I would enable the new coco-default.rego as default policy for any TEE_PLATFORM, and continue to use all-allowed.rego for all non-TEE platforms that support Policy.
>
> But by the time of the merge it's possible that we'll come up with better default policies than coco-default.rego. The current coco-default.rego is safe enough for the short term but it's not flexible enough for the future.

Great, that matches my expectation!

stevenhorsman (approved):
LGTM. Thanks @danmihai1!

fitzthum:
> @fitzthum , please speak up in case you have better ideas to mitigate potential policy issues when merging CCv0 with main.

I think this is the right approach. This is a good chance to standardize on one default (unlike the current agent-config) and I think the one you have here is appropriate. We might be able to evolve the default later but this seems like a good start.

stevenhorsman:
All the SEV test failed and it looks like the SNP node has network issues. @ryansavino - can you take a look please

fitzthum:
SEV tests have `Error: failed to create containerd container: create instance 1215: object with key "1215"`

ryansavino:
@AdithyaKrishnan

ryansavino:
@fitzthum Do we know if someone is able to try to source the issue with the nydus snapshotter on this? Going to quickly become unsupportable for us to clean up the ctr images for every PR. In the interim, we can maybe add a line in the tests to cleanup these images.

fitzthum:
> @fitzthum Do we know if someone is able to try to source the issue with the nydus snapshotter on this? Going to quickly become unsupportable for us to clean up the ctr images for every PR. In the interim, we can maybe add a line in the tests to cleanup these images.

I'm not sure if anyone is really on it yet, but I see the same error coming up on the operator changes that @fidencio is working on. Seems like this issue is on the critical path. I will try to reproduce locally.

ryansavino:
Weird. For the sev test, I see that message in the log: object with key "1215"

However, the image isn't there like it has been before. Let me try just restarting the sev test.

stevenhorsman:
@ryansavino - given the SNP tests have been failing for over a week are you okay for us to ignore them and merge PRs that fail there, or have I missed an update about what the problem is and is being worked on?

fitzthum:
> given the SNP tests have been failing for over a week are you okay for us to ignore them and merge PRs that fail there, or have I missed an update about what the problem is and is being worked on?

These changes are in theory totally disconnected from the failures that we are seeing, but the error in the SEV CI is a serious one (and probably not SEV specific).

stevenhorsman:
> These changes are in theory totally disconnected from the failures that we are seeing, but the error in the SEV CI is a serious one (and probably not SEV specific).

If you are referring to the errors like:
```
Error: failed to create containerd container: create instance 1651: object with key "1651" already exists: unknown
```
we see them in the SEV daily baseline and I believe you and Ryan have both seen them locally as well, so I don't think it's related at all to the changes in this PR. Also the last SEV test run was green here.

fitzthum:
> we see them in the SEV daily baseline and I believe you and Ryan have both seen them locally as well, so I don't think it's related at all to the changes in this PR. Also the last SEV test run was green here.

Yes. It isn't related to this PR so there is a case to be made for merging this, but we shouldn't gloss over the fact that we have a fairly serious bug. We can get the CI to behave if Ryan manually prunes some of the old images, hence the occasional pass.

I think that Gerry's recent containerd PR might fix this. See [this thread](https://cloud-native.slack.com/archives/C039JSH0807/p1697467462470849). I haven't confirmed this yet.

fitzthum:
Btw, I think the only reason that we don't see this issue on the non-tee CI is that those tests run in a VM so the offending containerd state is wiped out automatically.

stevenhorsman:
> Yes. It isn't related to this PR so there is a case to be made for merging this, but we shouldn't gloss over the fact that we have a fairly serious bug. We can get the CI to behave if Ryan manually prunes some of the old images, hence the occasional pass.

There is no glossing over, but I'm not sure unrelated PRs are the best place to be tracking bugs. If you think we should have a global branch freeze until this is resolved then I'm not necessarily against that, but we need to discuss and announce it on the CNCF slack thread so everyone is aware as I've had multiple people ask me why their PRs aren't getting merge due to the SEV and SNP tests failing. I'm also not convinced that the bug you are referring to has anything to do with every SNP tests I've seen failing with:
```
03:17:30 [init.sh:106] INFO: Store custom stress image in registry
03:17:30 4d613fd01acdec87c3369851ceba38d867e76debffdc9b689526760119175546
03:17:30   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
03:17:30                                  Dload  Upload   Total   Spent    Left  Speed
03:17:30 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
03:17:30 curl: (56) Recv failure: Connection reset by peer
```
for at least the last 5/6 days?

fitzthum:
> I'm also not convinced that the bug you are referring to has anything to do with every SNP tests I've seen failing with:

This is a different issue; networking error.

In theory we are tracking the key issue separately, although I guess an earlier bug is being prioritized rather strongly.

ryansavino:
I noticed yesterday that there were some pushes on this PR, so I refrained from checking the SEV/SNP tests on this PR until the code changes were complete. I'll retrigger the SNP tests now. @AdithyaKrishnan and I will do some investigation into the SNP node network issues.

In regards to the "key already exists" error. I'm hoping that Gerry's PR fixes it. However, that error is not SEV specific so I don't think that referring to SEV failures due to that error is necessarily correct. We should prioritize finding a fix for it because the current model of manually deleting the images on the host for every PR isn't going to be sustainable. We're happy to assist for now, but this error probably should have gated the nydus snapshotter PR. Let me know if there's anything I can do to help here. I'm not too familiar with the nydus stuff - I could take a look, but would like some help from the experts if possible. Thanks.

danmihai1:
Thanks @ryansavino! Those pushes were just my way of re-triggering the tests after rebasing my commit onto the tip of the target branch.

danmihai1:
@ryansavino , @fitzthum , @stevenhorsman , do you still want to keep this apparently-harmless PR on hold?

ryansavino:
I'm okay for it to be merged. Still some networking issue with the SNP test, but I'm pretty sure it's unrelated to the PR contents. Just trying to keep the tests running and passing.

Issue #8221, opened 2023-10-13
CC | policy: Enable the image-rs feature to get signatures from a registry


mattarnoatibm:
**Which feature do you think can be improved?**

Image-rs can now get simple signing signature from a registry that supports the X-Registry-Supports-Signatures
API extension as well as from a local sigstore. This issue to to bump the version of image rs the kata agent uses and enable this feature.

**How can it be improved?**

Build the kata agent with a new version of image rs and enable the `signature-simple-xrss` feature.

**Additional Information**

N / A

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

PR #8222, opened 2023-10-13
CC | policy: Enable signature-simple-xrss

mattarnoatibm:
Bump version of image-rs.

Build image-rs with the signature-simple-xrss feature flag.

Fixes: #8221


stevenhorsman (approved):
LGTM. Thanks @mattarnoatibm

stevenhorsman:
Given the TDX and SNP tests are not stable at the moment, I think we can merge this on the condition that if we find any issues with this then we will back out the change. Thanks Matt.

PR #8096, opened 2023-10-03
tests/cri-containerd: split between cri-containerd and crictl suites

wainersm:
This split the cri-containerd in two suites of tests:

* cri-containerd:  keep running containerd's [cri-integration tests](https://github.com/containerd/containerd/blob/main/docs/cri/testing.md#cri-integration-test)
* crictl: new home for the crictl-based tests 

Then the crictl tests were convert to use the bats framework. I didn't  want to improve the tests, depict many opportunities, so I tried to keep the original code and logic intact as much as possible. In the future, when we start pulling tests from CCv0 I think it will be the right time for improvements.

A github workflow was created to run the crictl suite of tests. The workflow won't fail due to failures on tests, because I will be able to fully test the tests on CI after the workflow is merged (right?) and then I don't want a broken workflow blocking unrelated pull requests.

Fixes #8160

Issue #7415, opened 2023-07-21
 Proposal to always build the measured boot image feature


wainersm:
>Note: this issue is based on a discussion that @fidencio, @stevenhorsman and I had about the merge of CCv0 and main branches. This is basically a re-write of @stevenhorsman 's comment on https://github.com/kata-containers/kata-containers/pull/7231#issuecomment-1643944633 with a couple of extra information.

**Problem statement**

The [measured boot image/rootfs](https://github.com/confidential-containers/documentation/issues/40) feature was proposed by @arronwy on the context of Confidential Containers (CoCo) aiming to provide an efficient (performance wise) way to check the integrity of the guest image/rootfs.

In a nutshell it works like:

 * an [initramfs that contains the tooling](https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/static-build/initramfs) (e.g. `veritysetup`)  to verify guest image on boot time should be built
 * the initramfs is built into the guest kernel
 * the initramfs's init scripts mount the guest rootfs and perform checks
 * the measured boot is activated only and only if the `rootfs_verity.scheme` and `cc_rootfs_verity.hash` parameters are passed to the kernel (down to the initramfs's init) via Kata's `configuration.toml` (`kernel_params` option). 

That feature is currently enabled on build time by exporting the `MEASURED_BOOT=yes` variable. It has been enabled by default on "vanilla" and TDX builds of Kata Containers on the CCv0 branch. The SEV/SNP builds doesn't build on that because they have already another TEE-specific approach for boot measurement, thus, SEV/SNP build with `MEASURED_BOOT=no`. On the `main` branch, despite the measured boot image code already be there, currently it isn't tested by any CI job.

All that said, our concern is about how to handle that feature on the `main` branch. After some chatting, we agreed that the best approach is what is described on the next section.

**Which feature do you think can be improved?**

The measured boot rootfs is hardware-agnostic so you don't need an TEE (Trusted Execution Environment) like Intel TDX or AMD SEV to enable it. We believe that it can be useful outside of CoCo uses case, for example, an infra-owner can enable the measured boot image to check the kata's guest image is not tampered.

Our proposal comprises of two improvements:

1.  Adapt the build scripts to **always** build the initramfs into the kernel, even for the "vanilla" kernels
2.  Provides a mechanism to allow the user to easily enable the measured boot on a Kata Containers installation

**How can it be improved?**

[...]

PR #8227, opened 2023-10-13
tests: bring the CCv0 cri-containerd tests to main

wainersm:
## Motivation

As part of https://github.com/kata-containers/kata-containers/issues/7415 (Proposal to always build the measured boot image feature) I want to migrate the measured roofts tests that exist on CCv0 branch to main. Those tests share code with the agent API and image tests, actually the measured rootfs tests are hosted in the image tests file. To bring those tests to main I would need to bring the common code which would require bringing more tests by cherry-picking commits if we want to preserve the authorship (and I'd like to keep the authorship of the code!). In the end, unless I am proven wrong, the process of cherry-picking a couple of tests would result on almost the same work of pulling the entire cri-containerd suite of tests.

I'm very aware that currently in CCv0 we have cri-containerd and k8s tests that overlap and it seems the community want to avoid duplication and prioritize the k8s ones (which I am in favor too). However, one thing to consider is that with the new CI and the k8s tests running on AKS, tests that require changes on the system's files (e.g. some tweaks on containerd's configuration, or kata's configuration itself) become more complex because the test machine (k8's worker node) isn't the job machine. While with the cri-containerd based tests we run on same machine so easier to apply changes. That would be the case of the measured rootfs tests and so I suspect there are more cases.

Taking all that in consideration I thought the best approach would be to bring the entire cri-containerd suite of tests to main, and preserve the authorship of the code. Then we begin the process of dropping, adapting and adding new code.

Like making a marble statue, let's have a big and imperfect stone then we begin pitching it and finish sculpting a bust. :)

## The process 

How I ended up with this PR:

Step1) Created a work base

I cherry-picked (`git am`) an initial set of commits that touch `integration/containerd/confidential` from the tests repository with the script [1] . Some commits didn't apply because 1) the target file does not exist on kata-containers repository (e.g. `lib/common.bash`), 2) Git couldn't manage conflicts (mostly on `confidential/lib.sh`), so the script uses `exclude files` rules to apply the commits regardless of conflicts.

Notice that I stopped pulling commits from commit 4445f5fc on. It remained just a couple of recent changes that were introduced to the release 0.8 I think. Anyway, I should revisit those on a future PR.

Step 2) Resolved conflicts in `lib/common.bash`

`lib/common.bash` does not exist on kata-containers. Manually applied the changes from commits b306e7a7 and 5f813c46 that touch `lib/common.bash`. Additionally cherry-picked the commits:

 * 40d7f5d0a (integration/containerd: add handler for initrd image on cp_to_guest_img())
 * 976640051 (ci/lib: run kata-runtime as non-sudo in cp_to_guest_img())
 * 08d9610fe (ci: Add sudo's to kata-runtime commands in CC code)

Step 3) Resolved conflicts in `tests/integration/confidential/lib.sh`

Manually applied the changes from commits e62a1852, b6c09d45, d09395e1 and 1940a4db, respectively. Additionally cherry-picked the commits: 

 * aa873fac42 (confidential: Adjust switch_image_service_offload())

step 4) Resolved conflicts in `confidential/fixtures/cosign/`

Cherry-picked additional commits:

 * 32816c2b6 (fixtures/cosign: Use KBS URI in keyPath attributes of policy)
 * f9b180b58 (fixtures/cosign: add KBS URI resources to offline_fs_kbc.json)

Then handled commit e922c5bb.

Step 4) Removed the `Fixes` and `Depends-on` on the commits messages to avoid problems with the CI's static checks.

Note:

* In the end I didn't apply c4e7c226 (see on migration's script that the commit is skipped). This commit introduces the proxy stuff that we aren't sure will be using on main
* The commit 08d3693af (CC: Enable e2e tests for IBM Z secure execution (SE))  despite changing the confidential/lib.sh file, it was left behind too. It changes the `cp_to_guest_img()` such as the guest image build for s390x is triggered, which isn't the way we want on main (I guess). 

## To be done

 * I plan to merge those tests in the (new) crictl suite that I create in https://github.com/kata-containers/kata-containers/pull/8096; so that PR should be merged first
 * Apply some simple clean ups like remove the references to `confidential` and `cc` on names (e.g. pods names) as well as drop functions that we know won't be used (e.g. `switch_image_service_offload()`)
 * Change the tests to skip all initially

## References

* [1] Cherry-pick commits script:

```bash
#!/bin/bash
#
set -e

basedir="tests/integration"
tests_repo_dir="../../../../tests"

# Build a list of commits on revert order (i.e first to apply first)
pushd  "${tests_repo_dir}"
for c in $(git --no-pager log --abbrev-commit --pretty=oneline \
	integration/containerd/confidential/ \
	| awk '{ print $1 }'); do

	commits="$c ${commits}"
done

popd

# The tests were introduced on integration/confidential then moved to
# integration/containerd/confidential. So the first commit is not catch
# on the history list above.
commits="7a06c29d6 ${commits}"

cherry_pick() {
	local commit="$1"
	shift
	local exclude="$*"

	local exclude_patterns
	for p in $exclude; do
		exclude_patterns+=" --exclude=$p"
	done

	git --git-dir="${tests_repo_dir}/.git" \
		format-patch -k -1 --stdout "$commit" | \
	git am -3 -k -p2 \
		--directory "$basedir" ${exclude_patterns}
}

for c in $commits; do
	echo "Cherry-pick: $c"
	exclusion_patterns="${basedir}/setup_env_ubuntu.sh ${basedir}/kubernetes/**"

	if [ "$c" = "4445f5fc" ]; then
		#ci: containerd_nydus_setup
		echo "Stop here. More recent changes will be manually cherry-picked"
		break
	fi

	case "$c" in
		c4e7c226)
			# This introduces the problematic confidential/lib.sh
			echo "skip..."
			continue ;;
		b306e7a7|5f813c46)
			# common.bash is not in this repo
			exclusion_patterns+=" ${basedir}/common.bash"
			;;
		e62a1852|b6c09d45|d09395e1|1940a4db)
			exclusion_patterns+=" ${basedir}/confidential/lib.sh"
			;;
		e922c5bb)
			exclusion_patterns+=" ${basedir}/confidential/fixtures/cosign/offline-fs-kbc/aa-offline_fs_kbc-resources.json"
			exclusion_patterns+=" ${basedir}/confidential/lib.sh"
			;;
	esac

	cherry_pick "$c" $exclusion_patterns

	# Append the cherry-picked message to commit
	commit_msg=$(git log --format=%B -n1)
	git commit --amend -m "$commit_msg" -m "[cherry-picked from tests commit $c]" >/dev/null
done

echo "SUCCEED!"
```


wainersm:
@stevenhorsman @fidencio what do you think?

Apart from your opinion, I want to merge this and https://github.com/kata-containers/kata-containers/pull/8096 , so I labeled `wip`.

fidencio:
> @stevenhorsman @fidencio what do you think?

I don't like it, I really must say I don't like it at all.

Let's avoid bringing the whole chunk, including things like `cp_to_guest_img`, which is really not acceptable with the tests on main, and rather do this with the chunks that can be tests right now.

Note: `cp_to_guest_img` may be a good way to modify the rootfs and do a negative test on the measured boot, but that's it, we should not be relying on this for other things.

Last but not least, bring chunks, add the SoB the folks who touched those chunks, but there's absolutely no need to preserve the history from CCv0, as that branch is just something that should never been existed in the first place.

Issue #8228, opened 2023-10-13
remove endpoint blocking ability from agent-config.toml


danmihai1:
**Which feature do you think can be improved?**

Agent Policy (https://github.com/kata-containers/kata-containers/issues/7573)

**How can it be improved?**

Deprecate the endpoint blocking feature described by https://github.com/kata-containers/kata-containers/issues/1837, because the same functionality is implemented by the Policy too.

PR #8229, opened 2023-10-13
agent: no endpoint blocking from agent-config.toml

danmihai1:
Remove the ability to block access to kata agent endpoints by using agent-config.toml. That functionality is now implemented using the Agent Policy feature (#7573).

The CCv0 branch relied on blocking endpoints using agent-config.toml but will set-up an equivalent default policy file instead (#8219).

Fixes: #8228


fidencio (approved):
lgtm, thanks @danmihai1!

sprt (approved):

Issue #7673, opened 2023-08-17
add tool/application that generates CoCo policy based on user's Kubernetes YAML files


danmihai1:
**Is your feature request related to a problem? Please describe.**

The Policy discussed in https://github.com/confidential-containers/confidential-containers/issues/126 can be very complex. Users should be able to generate good Policy files using their K8s YAML files as inputs.

**Describe the solution you'd like**

Provide to users the app from https://github.com/microsoft/kata-containers/tree/cc-msft-prototypes/src/tools/genpolicy or a very similar app.

To see examples of those generated Policies, you can run "base64 -d" on the values of the Policy annotations from https://github.com/microsoft/kata-containers/tree/cc-msft-prototypes/src/agent/samples/policy/yaml 

**Describe alternatives you've considered**

Creating good Policy files manually is very complex and error prone. 

**Additional context**

The Policy gets enforced by the Agent in collaboration with user's remote attestation server(s).

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?
Yes

PR #8248, opened 2023-10-18
tools: add policy generation tool

danmihai1:
Add application that infers K8s user's intentions based on user's K8s YAML file, and generates a Rego/OPA based policy for that YAML.

Fixes: #7673


src/tools/genpolicy/README.md
```diff
@@ -0,0 +1,28 @@
+# Agent Policy generation tool
+
+The Kata Containers policy generation tool (`genpolicy`): 
+
+1. Reads user's Kubernetes YAML file.
+
+1. Infers user's intentions based on the contents of that file.
+
+1. Generates a Kata Containers Agent (`kata-agent`) policy file
```
stevenhorsman, line 9: I'd be interested in understanding/communicating some of the rules that genpolicy used based on input patterns, or at least some examples as the policy file that is output is pretty large (the basic pod yaml example you give it decodes to >1600 lines long) so is tricky to understand e.g. things like the container image name will be extracted and added to the list of images that are allowed in the CreateContainerRequest, or something like that?

Maybe explaining the genpolicy-settings.json file, which looks like it does some templating? - what it is and what is does would help with this?

stevenhorsman (commented):
Hey Dan, thanks so much for the submission. There is a lot of code here and I'm not an expert, so I've focussed on the documentation and through to get the tool build and running myself primarily, and left some comments about how we might be able to make this a bit more accessible as there were a few hurdles I bumped into and the tool has a lot of power, so letting people understand the processing and output more feels important. Thanks!

danmihai1:
> Hey Dan, thanks so much for the submission. There is a lot of code here and I'm not an expert, so I've focussed on the documentation and through to get the tool build and running myself primarily, and left some comments about how we might be able to make this a bit more accessible as there were a few hurdles I bumped into and the tool has a lot of power, so letting people understand the processing and output more feels important. Thanks!

Thanks for all the feedback @stevenhorsman. I have just pushed what are significant doc improvements in https://github.com/kata-containers/kata-containers/pull/8248/commits/2257a1691794e4d2ad0c7e035ce8a4435f94d7c4

src/tools/genpolicy/README.md
```diff
@@ -0,0 +1,28 @@
...
+corresponding to the input YAML, using the [Open Policy Agent format](https://www.openpolicyagent.org/docs/latest/policy-language/).
+
+1. Appends the policy as an annotation to user's YAML file.
```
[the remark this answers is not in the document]
danmihai1: Added that note.

src/tools/genpolicy/README.md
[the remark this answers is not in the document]
danmihai1: Added the build steps into the new version of README.md

src/tools/genpolicy/README.md
danmihai1: I've just added many policy details into the new genpolicy-auto-generated-policy-details.md, and advanced command line usage in genpolicy-advanced-command-line-parameters.md.

stevenhorsman (approved):
Thanks very much for the extra documentation Dan - it really helps. Again the code is too much for me to do a sensible review, but I've kicked the tyres with testing, so I think it's fine to go in as the initial drop of some tooling.

katexochen (changes requested):
I'm also if there are any requirements for the container images used. Not sure if that's written anywhere in the documentation and I've missed it. I've tested a bunch of images, some worked, some failed:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: test
      image: ${image}
```

`image="ubuntu:latest"`
```
Error: failed to generate policies: failed to generate policy for deployment/test.yml: genpolicy failed with exit code 101: thread 'main' panicked at src/registry.rs:111:17:
Failed to pull container image manifest and config - error: UnsupportedMediaTypeError(
    "application/vnd.oci.image.index.v1+json",
)
```
`image="nixery.dev/shell/socat"`
```
Error: failed to generate policies: failed to generate policy for deployment/test.yml: genpolicy failed with exit code 101: thread 'main' panicked at src/registry.rs:91:61:
called `Result::unwrap()` on an `Err` value: Error("missing field `Env`", line: 1, column: 969)
```

src/tools/genpolicy/src/policy.rs
```diff
@@ -0,0 +1,900 @@
...
+}
+
+impl AgentPolicy {
+    pub async fn from_files(config: &utils::Config) -> Result<AgentPolicy> {
+        let mut config_maps = Vec::new();
+        let mut secrets = Vec::new();
+        let mut resources = Vec::new();
+        let yaml_contents = yaml::get_input_yaml(&config.yaml_file)?;
+
+        for document in serde_yaml::Deserializer::from_str(&yaml_contents) {
```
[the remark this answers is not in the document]
danmihai1: @katexochen , thanks for the feedback! This tool doesn't solve all the Confidential K8s Policy problems (yet?). For example, if you search for TODO in its source code, you'll find a bunch of other problem that were not solved yet. I think we should start somewhere and keep improving. After working on this for several months, I think that waiting for the ideal tool before the first merge is impractical.

danmihai1:
> I'm also if there are any requirements for the container images used. Not sure if that's written anywhere in the documentation and I've missed it. I've tested a bunch of images, some worked, some failed:
> 
> ```yaml
> apiVersion: v1
> kind: Pod
> metadata:
>   name: test
> spec:
>   containers:
>     - name: test
>       image: ${image}
> ```
> 
> `image="ubuntu:latest"`
> 
> ```
> Error: failed to generate policies: failed to generate policy for deployment/test.yml: genpolicy failed with exit code 101: thread 'main' panicked at src/registry.rs:111:17:
> Failed to pull container image manifest and config - error: UnsupportedMediaTypeError(
>     "application/vnd.oci.image.index.v1+json",
> )
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> `image="nixery.dev/shell/socat"`
> 
> ```
> Error: failed to generate policies: failed to generate policy for deployment/test.yml: genpolicy failed with exit code 101: thread 'main' panicked at src/registry.rs:91:61:
> called `Result::unwrap()` on an `Err` value: Error("missing field `Env`", line: 1, column: 969)
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Please log github issues for YAML files that are not working and attach the problematic YAML files to those issues. For these particular two problems:

1. One of my colleagues at MSFT noticed the ubuntu:latest problem a few days ago, and we are working on a solution for it. I can share the gory details in case you are curious about the root cause. Meanwhile, creating your own container image on top of an Ubuntu base container image (but using a different manifest format) might be a good enough workaround.
2. Please attach your test.yaml in github - the Env problem you mentioned sounds more related to the YAML contents, not to the container image.

src/tools/genpolicy/genpolicy-advanced-command-line-parameters.md
```diff
@@ -0,0 +1,216 @@
...
+    - name: first-test-container
+      image: quay.io/prometheus/busybox:latest
+      command:
+        - sleep
+        - "120"
+```
+
+`genpolicy` downloads the `quay.io/prometheus/busybox:latest` container image. 
+
+Depending on the size of the container images and the speed of the network connection to the container registry, downloading these images might take several minutes. For testing scenarios where `genpolicy` gets executed several times, it can be useful to cache the container images after downloading them, in order to avoid most of the time needed to download the same container images multiple times. If a container image layer was already cached locally, `genpolicy` uses the local copy of that container layer. The application caches the image information under the `./layers_cache` directory.
```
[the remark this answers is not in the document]
danmihai1: I agree that caching needs future improvements, but I don't think the initial merge of this code should wait for these improvements to become available. Please create github issues for such ideas for improvements.

src/tools/genpolicy/genpolicy-advanced-command-line-parameters.md
```diff
@@ -0,0 +1,216 @@
...
+
+The default `genpolicy` settings file is `./genpolicy-settings.json`. Users can specify in the command line a different settings file by using the `-j` parameter - e.g.,
+
+```bash
+$ genpolicy -j my-settings.json -y test.yaml
+```
+
+# Use a custom path to `genpolicy` input files
+
+By default, the `genpolicy` input files [`rules.rego`](rules.rego) and [`genpolicy-settings.json`](genpolicy-settings.json) must be present in the current directory - otherwise `genpolicy` returns an error. Users can specify a different path to these two files, using the `-i` command line parameter - e.g.,
```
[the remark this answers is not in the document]
danmihai1: Please log an issue for the future.

src/tools/genpolicy/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+# In YAML, the value yes without quotes is reserved for boolean,
+# and confuses kubectl, that expects a string value.
+serde_yaml = "0.8"
+
+# Container repository.
+anyhow = "1.0.32"
+async-trait = "0.1.68"
+docker_credential = "1.2.0"
+flate2 = { version = "1.0.26", features = ["zlib-ng"], default-features = false }
+oci-distribution = { version = "0.9.4" }
```
ChengyuZhu6, line 37: > 1. One of my colleagues at MSFT noticed the ubuntu:latest problem a few days ago, and we are working on a solution for it. I can share the gory details in case you are curious about the root cause. Meanwhile, creating your own container image on top of an Ubuntu base container image (but using a different manifest format) might be a good enough workaround.

Hey @danmihai1, I noticed the problem with ubuntu:latest, which was caused by the manifest type. We should handle manifests with the mediatype “application/vnd.oci.image.manifest.v1+json” when pulling them from the registry. This issue had been solved in the [oci-distribution](https://github.com/krustlet/oci-distribution/pull/67). Could you please change the oci-distribution repo branch to [this commit](https://github.com/krustlet/oci-distribution/pull/67/commits/0036d93e8f6eb1df4e01e60c4f1c9c77ea43461c)?”

src/tools/genpolicy/src/policy.rs
```diff
@@ -0,0 +1,900 @@
...
+
+        add_missing_strings(&self.Process.Env, &mut process.Env);
+    }
+}
+
+fn get_image_layer_storages(
+    storages: &mut Vec<agent::Storage>,
+    image_layers: &Vec<registry::ImageLayer>,
+    root: &KataRoot,
+) {
```
ChengyuZhu6, line 626: Hey @danmihai1. Am I correct in assuming that this is an interface for sharing image layers? I am thrilled to see this feature, as I am working on integrating image sharing feature into the main branch (https://github.com/kata-containers/kata-containers/pull/7837). I would like to pass the dm-verity information of the image or layer to the guest via the policy to check the info from host is correct as expected, if possible. I would appreciate your assistance with this task.

danmihai1: Please ping me on Slack when you are ready to discuss this. 

Yes, the dm-verity information of each container layer is being verified already when using the Policy generated by the genpolicy tool. But, the current Policy is based on Wedson's snapshotter design/implementation that we discussed several months ago. Hopefully there aren't big differences between your version and Wedson's. If there are significant differences, we'll have to modify this tool to support both.

danmihai1:
No code or doc changes in my latest push. Just squashed the doc changes into a single commit and rebased on top of the latest main code.

src/tools/genpolicy/Cargo.toml
danmihai1: @ChengyuZhu6 @katexochen Christmas came early for me this year and brought a new release of the oci_distribution crate - a few days ago. Therefore, I have just added to this PR @Redent0r's commit that adds support for the ubuntu:latest image by using the new oci_distribution version.

huangyingting:
@danmihai1 I want to report an issue while I am testing confidential containers from AKS, there is no place from microsoft/kata-containers to let me open an issue, so I put it here

It seems we not able to correctly handle mountPropagation: Bidirectional or mountPropagation: HostToContainer on emptyDir volume, as a consequence, the container won't be created with error "Error: failed to create containerd task: failed to create shim task: "CreateContainerRequest is blocked by policy": unknown"

To work around the issue, for HostToContainer, we need to replace rprivate to rslave in the generated policy.

          {
            "destination": "/mnt/remote",
            "source": "^$(cpath)/$(sandbox-id)/local/remotemounts$",
            "type_": "local",
            "options": [
              "rbind",
              "rslave",
              "rw"
            ]
          }
For Bidirectional, we need to replace rprivate to rshared in the generated policy.

          {
            "destination": "/mnt/remote",
            "source": "^$(cpath)/$(sandbox-id)/local/remotemounts$",
            "type_": "local",
            "options": [
              "rbind",
              "rshared",
              "rw"
            ]
          }

To reproduce the issue, you can use https://github.com/huangyingting/sev-snp/blob/main/encfs.yaml (ignore io.katacontainers.config.agent.policy, it was tweaked already based on above findings to make it work)

danmihai1:
> @danmihai1 I want to report an issue while I am testing confidential containers from AKS, there is no place from microsoft/kata-containers to let me open an issue, so I put it here
> 
> It seems we not able to correctly handle mountPropagation: Bidirectional or mountPropagation: HostToContainer on emptyDir volume, as a consequence, the container won't be created with error "Error: failed to create containerd task: failed to create shim task: "CreateContainerRequest is blocked by policy": unknown"

Thanks! We are tracking other bugs applicable to this app under https://github.com/kata-containers/kata-containers/issues .

fidencio (approved):
@danmihai1, lgtm, thanks!

I think this one should get in sooner than later, as we have been tracking the improvements in a different issue.

katexochen (approved):

PR #8252, opened 2023-10-19
[WIP] runtime-rs: support nydus v6 as rootfs of container

Leospard:
Depend on kernel patch where nydus v6 could also be handled by the guest kernel.
With that patch, bootstrap and blobs are shared with guest VM via passthroughfs and blobfs so that they can be mounted inside guest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ src/agent/protocols/src/*.rs
 !src/agent/protocols/src/lib.rs
 build
 src/tools/log-parser/kata-log-parser
+.DS_Store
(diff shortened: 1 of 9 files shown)


imeoer:
@Leospard Can we give some usage docs?

.gitignore
```diff
@@ -15,3 +15,4 @@ src/agent/protocols/src/*.rs
 !src/agent/protocols/src/lib.rs
 build
 src/tools/log-parser/kata-log-parser
+.DS_Store
```
imeoer, line 18: Redundant file.

src/runtime-rs/crates/resource/src/rootfs/nydus_rootfs.rs
```diff
@@ -124,6 +130,72 @@ impl NydusRootfs {
                     rootfs_guest_path,
                 ))
             }
+            NYDUS_ROOTFS_V6 => {
+                // enable dax flag for bootstrap and blobfiles.
+                let dax_threshold_size_kb: u64 = 0;
```
Apokleos, line 135: Why `dax_threshold_size_kb` with hardcode `0` ?

One more question but important: does it need to check DAX whether it's enabled ?

Apokleos (commented):
Thx @Leospard some comments!

PR #8261, opened 2023-10-20
do-not-merge: agent: add example of querying data from policy

danmihai1:
This query method can be used after the policy has been initialized by the kata agent.

### DIFF
--- a/src/agent/src/policy.rs
+++ b/src/agent/src/policy.rs
@@ -29,6 +29,12 @@ struct AllowResponse {
     result: bool,
 }
 
+/// Example of HTTP response from OPA: {"result":"foo""}
+#[derive(Debug, Serialize, Deserialize)]
+struct StringResponse {
+    result: String,
+}
+
 /// Singleton policy object.
 #[derive(Debug, Default)]
 pub struct AgentPolicy {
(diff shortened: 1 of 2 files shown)


src/agent/src/policy.rs
```diff
@@ -243,6 +249,70 @@ impl AgentPolicy {
...
+
+    async fn get_string_value(&mut self, key: &str) -> Result<String> {
+        Ok(self.get_data(key).await?)
+    }
+
+    // TODO: remove this test
+    pub async fn test_query_policy_data(&mut self) {
+        // See kata-containers/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-set-keys.rego
+        // for an example of how to add this data into your policy.
+        let key = "my_test_data/default/key/ssh-demo";
```
fitzthum, line 303: This looks like a key, which I don't think we should be passing via policy. Might be better to do something more like a public key here.

danmihai1: It's possible that a better example would be more helpful. But the only goal for this code is to show how to read arbitrary data from Policy (the my_test_data/default/key/ssh-demo data in this case).

Here I just borrowed some example data from https://github.com/confidential-containers/confidential-containers/blob/b36a57e530aa83741fba2592574a7a4a4d8892b6/demos/ssh-demo/aa-offline_fs_kbc-keys.json or similar JSON file.

fitzthum: Yeah no big deal, but that is a symmetric key. Might help make your example clearer to use something else.

Issue #8283, opened 2023-10-21
runtime-rs: Update PCI path information for VFIO and virtio-blk devices


amshinde:
While implementing change to hotplug VFIO and virtio-block devices for cloud-hypervisor, it was observed that although these devices are added to the hypervisor sucessfully , these are not detected by the agent within the VM. This is because the PCI Path information for these devices is not provided correctly to the agent by runtime-rs.

jodh-intel:
Related: #7218, #7489.

PR #7932, opened 2023-09-13
runtime-rs: bringing virtio-fs device in device-manager

Apokleos:
Currently, virtio-fs is still outside of the device manager. This causes some management issues, such as the inability to unify PCI address management.
To do Enhancement of device manager, it will bring virtio-fs device in device-manager for unified management

Fixes: #7915

PR #8284, opened 2023-10-21
runtime-rs: update device pci info for vfio and virtio-blk devices

amshinde:
runtime-rs: change hypervisor add_device trait to return device copy
    
 Block(virtio-blk) and vfio devices are currently not handled correctly
 by the agent as the agent is not provided with correct PCI paths for
 these devices.
    
 The PCI paths for these devices can be inferred from the PCI information
 provided by the hypervisor when the device is added.
 Hence changing the add_device trait function to return a device copy
 with PCI info potentially provided by the hypervisor. This can then be
 provided to the agent to correctly detect devices within the VM.
    
This commit includes implementation for PCI info update for
cloud-hupervisor for virtio-blk devices with stubs provided for other
hypervisors.

This PR fixes virtio-block device addition bug. The following works now:

```
$ sudo ctr run --runtime "io.containerd.kata.v2" --snapshotter devmapper --rm -t quay.io/prometheus/busybox:latest foo-dm /bin/true;echo $?
0
```
 as well as 
```
sudo ctr run --runtime "io.containerd.kata.v2" --device=/dev/block1 --rm -t quay.io/prometheus/busybox:latest foo-dm /bin/true;echo $?
0
```
The PR also fixes bugs around vfio device addition. End to end vfio device add now works:
```
sudo ctr run --runtime "io.containerd.kata.v2" --device=/dev/vfio/{vfio_group} --rm -t quay.io/prometheus/busybox:latest foo-dm /bin/true;echo $?
0
```

    
 Fixes: #8283

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
@@ -169,7 +169,7 @@ impl CloudHypervisorInner {
     }
 
     async fn handle_vfio_device(&mut self, device: VfioDevice) -> Result<DeviceType> {
-        let vfio_device: VfioDevice = device.clone();
+        let mut vfio_device: VfioDevice = device.clone();
 
         // A device with multi-funtions, or a IOMMU group with one more
         // devices, the Primary device is selected to be passed to VM.
@@ -210,6 +210,12 @@ impl CloudHypervisorInner {
                 serde_json::from_str(detail.as_str()).map_err(|e| anyhow!(e))?;
             self.device_ids
                 .insert(device.device_id.clone(), dev_info.id);
+
+            // Update PCI path for the vfio host device. It is safe to directly access the slice element
+            // here as we have already checked if it exists.
+            // Todo: Handle vfio-ap mediated devices - return error for them.
+            vfio_device.devices[0].guest_pci_path =
+                Some(Self::clh_pci_info_to_path(&dev_info.bdf)?);
         }
 
         Ok(DeviceType::Vfio(vfio_device))
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
@@ -7,16 +7,18 @@
 use super::inner::CloudHypervisorInner;
 use crate::device::DeviceType;
 use crate::BlockDevice;
-use crate::HybridVsockConfig;
+use crate::HybridVsockDevice;
 use crate::NetworkConfig;
+use crate::PciPath;
+use crate::ShareFsDevice;
 use crate::ShareFsDeviceConfig;
 use crate::VfioDevice;
 use crate::VmmState;
 use anyhow::{anyhow, Context, Result};
+use ch_config::ch_api::cloud_hypervisor_vm_device_add;
 use ch_config::ch_api::{
-    cloud_hypervisor_vm_blockdev_add, cloud_hypervisor_vm_device_add,
-    cloud_hypervisor_vm_device_remove, cloud_hypervisor_vm_fs_add, PciDeviceInfo,
-    VmRemoveDeviceData,
+    cloud_hypervisor_vm_blockdev_add, cloud_hypervisor_vm_device_remove,
+    cloud_hypervisor_vm_fs_add, PciDeviceInfo, VmRemoveDeviceData,
 };
 use ch_config::convert::{DEFAULT_DISK_QUEUES, DEFAULT_DISK_QUEUE_SIZE, DEFAULT_NUM_PCI_SEGMENTS};
 use ch_config::DiskConfig;
@@ -31,7 +33,7 @@ pub const DEFAULT_FS_QUEUES: usize = 1;
 const DEFAULT_FS_QUEUE_SIZE: u16 = 1024;
 
 impl CloudHypervisorInner {
-    pub(crate) async fn add_device(&mut self, device: DeviceType) -> Result<()> {
+    pub(crate) async fn add_device(&mut self, device: DeviceType) -> Result<DeviceType> {
         if self.state != VmmState::VmRunning {
             // If the VM is not running, add the device to the pending list to
             // be handled later.
@@ -55,8 +57,8 @@ impl CloudHypervisorInner {
             // - Network details need to be saved for later application.
             //
             match device {
-                DeviceType::ShareFs(_) => self.pending_devices.insert(0, device),
-                DeviceType::Network(_) => self.pending_devices.insert(0, device),
+                DeviceType::ShareFs(_) => self.pending_devices.insert(0, device.clone()),
+                DeviceType::Network(_) => self.pending_devices.insert(0, device.clone()),
                 _ => {
                     debug!(
                         sl!(),
@@ -65,20 +67,18 @@ impl CloudHypervisorInner {
                 }
             }
 
-            return Ok(());
+            return Ok(device);
         }
 
-        self.handle_add_device(device).await?;
-
-        Ok(())
+        self.handle_add_device(device).await
     }
 
-    async fn handle_add_device(&mut self, device: DeviceType) -> Result<()> {
+    async fn handle_add_device(&mut self, device: DeviceType) -> Result<DeviceType> {
         match device {
-            DeviceType::ShareFs(sharefs) => self.handle_share_fs_device(sharefs.config).await,
-            DeviceType::HybridVsock(hvsock) => self.handle_hvsock_device(&hvsock.config).await,
+            DeviceType::ShareFs(sharefs) => self.handle_share_fs_device(sharefs).await,
+            DeviceType::HybridVsock(hvsock) => self.handle_hvsock_device(hvsock).await,
             DeviceType::Block(block) => self.handle_block_device(block).await,
-            DeviceType::Vfio(vfiodev) => self.handle_vfio_device(&vfiodev).await,
+            DeviceType::Vfio(vfiodev) => self.handle_vfio_device(vfiodev).await,
             _ => Err(anyhow!("unhandled device: {:?}", device)),
         }
     }
@@ -108,9 +108,13 @@ impl CloudHypervisorInner {
         }
     }
 
-    async fn handle_share_fs_device(&mut self, cfg: ShareFsDeviceConfig) -> Result<()> {
-        if cfg.fs_type != VIRTIO_FS {
-            return Err(anyhow!("cannot handle share fs type: {:?}", cfg.fs_type));
+    async fn handle_share_fs_device(&mut self, sharefs: ShareFsDevice) -> Result<DeviceType> {
+        let device: ShareFsDevice = sharefs.clone();
+        if device.config.fs_type != VIRTIO_FS {
+            return Err(anyhow!(
+                "cannot handle share fs type: {:?}",
+                device.config.fs_type
+            ));
         }
 
         let socket = self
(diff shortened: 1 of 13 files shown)


Apokleos:
Hey @amshinde 

Nice work on this PR! I'm also working on this feature, and I've submitted a [WIP PR](https://github.com/kata-containers/kata-containers/pull/7489) as well.

My PR mainly implements the following features:

(1) Due to the different ways that different VMMs handle PCI devices, we hope to provide a general PCIe topology 
   processing framework that is as compatible as possible with VMMs such as dragonball, cloud-hyperivor, qemu, and fc.
(2) Provide the correct PCI path for PCI devices, including VFIO-PCI devices and virtio-PCI(virtio-blk/vhost-user-blk...) devices.
(3) Fill in the correct correspondence between host path and guest path for kata-agent devices.

Would you be interested in discussing the difference and a merge? 
We can compare the pros and cons of both approaches and see if we can come up with a solution that works for everyone.

Let me know what you think.

Thx.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -7,16 +7,19 @@
 use super::inner::CloudHypervisorInner;
 use crate::device::DeviceType;
 use crate::BlockDevice;
-use crate::HybridVsockConfig;
+use crate::HybridVsockDevice;
+//use crate::HybridVsockConfig;
```
jodh-intel, line 11: Commented out line.

src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
```diff
@@ -65,20 +68,22 @@ impl CloudHypervisorInner {
...
 
-    async fn handle_add_device(&mut self, device: DeviceType) -> Result<()> {
+    async fn handle_add_device(&mut self, device: DeviceType) -> Result<DeviceType> {
         match device {
-            DeviceType::ShareFs(sharefs) => self.handle_share_fs_device(sharefs.config).await,
-            DeviceType::HybridVsock(hvsock) => self.handle_hvsock_device(&hvsock.config).await,
+            //DeviceType::ShareFs(sharefs) => self.handle_share_fs_device(sharefs.config).await,
+            //DeviceType::HybridVsock(hvsock) => self.handle_hvsock_device(&hvsock.config).await,
+            //DeviceType::Block(block) => self.handle_block_device(block.config).await,
+            //_ => Err(anyhow!("unhandled device: {:?}", device)),
```
jodh-intel, line 82: Can you remove these commented out lines.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_blk.rs
```diff
@@ -78,11 +82,20 @@ impl Device for BlockDevice {
...
+        match h.add_device(DeviceType::Block(self.clone())).await {
+            Ok(dev) => {
+                // Update device info with the one received from device attach
+                if let DeviceType::Block(blk) = dev {
+                    self.config = blk.config;
+                }
+                Ok(())
+            }
+            Err(e) => {
+                self.decrease_attach_count().await?;
```
jodh-intel, line 95: If this fails, we'll return the count error, not the original one (line below).

jodh-intel (commented):
Thanks @amshinde - a few comments.

Also, although this works with the default pmem driver, it doesn't wfm using DM:

```bash
$ sudo ctr run --runtime "io.containerd.kata.v2"  --rm -t quay.io/prometheus/busybox:latest foo /bin/true;echo $?
0

$ sudo ctr run --runtime "io.containerd.kata.v2" --snapshotter devmapper --rm -t quay.io/prometheus/busybox:latest foo-dm /bin/true;echo $?
ctr: failed to create shim task: Others("failed to handler message create container\n\nCaused by:\n    0: create\n    1: handler rootfs\n    2: new block rootfs\n    3: do handle device failed.\n    4: failed to add deivce\n    5: Unexpected PCI address \"0000:00:05.0\" for clh device add"): unknown
1
```

amshinde:
> Would you be interested in discussing the difference and a merge? We can compare the pros and cons of both approaches and see if we can come up with a solution that works for everyone.
> 
> Let me know what you think.

@Apokleos Yes, lets discuss this and figure out a common approach.

amshinde:
hey @Apokleos I just saw your PR https://github.com/kata-containers/kata-containers/pull/7932, where you introduce update-device trait with similar idea of the hypervisor providing pcie info. Wanted to discuss with you if we could arrive and agree on a way to do this.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_blk.rs
amshinde: This is the same as previous behaviour. If we want to change this, lets discuss. It could go as part of a separate change.

src/runtime-rs/crates/resource/src/share_fs/share_virtio_fs.rs
```diff
@@ -89,9 +89,16 @@ pub(crate) async fn setup_inline_virtiofs(id: &str, h: &dyn Hypervisor) -> Resul
...
-    h.add_device(DeviceType::ShareFsMount(virtio_fs))
+
+    let result = h
+        .add_device(DeviceType::ShareFsMount(virtio_fs))
         .await
-        .with_context(|| format!("fail to attach passthrough fs {:?}", source))
+        .with_context(|| format!("fail to attach passthrough fs {:?}", source));
+
+    match result {
+        Ok(__) => Ok(()),
```
jodh-intel, line 99: Nit:

```suggestion
        Ok(_) => Ok(()),
```

? :smile:

jodh-intel (approved):
Thanks @amshinde - Tested and wfm.

We still need input from @Apokleos about his PR and also comments on the very minor hypervisor trait tweak on this PR.

lgtm

src/runtime-rs/crates/resource/src/share_fs/share_virtio_fs.rs
amshinde: fixed :) It still compiled with that!

amshinde:
> Also, although this works with the default pmem driver, it doesn't wfm using DM:

This is now fixed with this PR. This PR fixes the device addition bug seen with:

```
$ sudo ctr run --runtime "io.containerd.kata.v2" --snapshotter devmapper --rm -t quay.io/prometheus/busybox:latest foo-dm /bin/true;echo $?
0
```
 as well as 
```
sudo ctr run --runtime "io.containerd.kata.v2" --device=/dev/block1 --rm -t quay.io/prometheus/busybox:latest foo-dm /bin/true;echo $?
0
```

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -47,29 +44,6 @@ lazy_static! {
         Arc::new(RwLock::new(HashMap::new()));
 }
 
-// map host/guest bdf and the mapping saved into `HOST_GUEST_MAP`,
-// and return PciPath.
-pub fn generate_guest_pci_path(bdf: String) -> Result<PciPath> {
```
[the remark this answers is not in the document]
amshinde: Ok, I guess this would be used by dragonball. I'll add it back.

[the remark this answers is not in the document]
amshinde: @Apokleos I have added  `generate_guest_pci_path` back in case it is usefull for dragonball. Plese take a look.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -493,13 +455,34 @@ impl Device for VfioDevice {
...
+                if let DeviceType::Vfio(vfio) = dev {
+                    self.config = vfio.config;
+                    self.devices = vfio.devices;
+                }
 
-            return Err(e);
-        }
+                if self.bus_mode == VfioBusMode::PCI {
+                    for hostdev in self.devices.iter() {
+                        if hostdev.guest_pci_path.is_some() {
```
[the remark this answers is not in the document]
amshinde: Done

amshinde:
@Apokleos I have addressed your comments.

src/runtime-rs/crates/hypervisor/src/device/mod.rs
```diff
@@ -40,7 +40,7 @@ pub enum DeviceType {
     ShareFs(ShareFsDevice),
     ShareFsMount(ShareFsMountDevice),
     HybridVsock(HybridVsockDevice),
-    Vsock(VsockDevice),
+    //Vsock(VsockDevice),
```
pmores, line 43: Why this change?

amshinde: I have now taken out the Vsock value entirely from the enum. I have added an explanation for it in the commit message. Please take a look at it and let me know if it makes sense.
To summarize, Vsock value was removed as the Vsock device is not really used anywhere nor can it be used in its current state as it does not implement the Device trait. This means that it cannot be attached or detached from the hypervisor. What I can deduce from the code is that one of the reasons is that Vsock does not really implement the Clone trait which looks like a prerequisite for the Device trait, as the async trait functions require Device copies to be passed between them.(Vsock struct currently has a File handle which does not implement Clone, so this would require some rework) This is also the reason that I decided to remove Vsock for now, as the change I am introducing here requires devices to implement Clone.

bergwolf: @amshinde Vsock is required for QEMU support. Could you add it back? Or @pmores would need to add it back in his PR. Vsock device is not fully implemented but it does not mean that it is not needed.

amshinde: @bergwolf The implementation is still present, I just removed it from the DeviceType enum. With its current implementation it neither supports attach or detach Device traits, meaning it cant really be hotplugged to a hypervisor. @pmores can add it back in the Device list once he adds support for qemu.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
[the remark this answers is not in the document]
Apokleos: Thx @amshinde

Apokleos (approved):
Thx @amshinde LGTM!

cmaf (approved):
lgtm @amshinde

src/runtime-rs/crates/hypervisor/src/device/mod.rs
pmores: @amshinde qemu support is added by PR #8185 which is broken by this change.

gkurz: @amshinde It would have been greatly appreciated to hold on a bit, so that @pmores had at least a chance to discuss some more before this change was merged.

amshinde: @gkurz Was not aware of this change drastiaclly breaking @pmores PR.  I was eager to get this landed to unblock folks to get started with implementing a CI for clh which is borked without this change.I had a look at @pmores PR for qemu support and here are my suggestions to handle the conflicts:
Like I mentioned I have removed VsockDevice from the DeviceType enum list. I think these enums should only specify devices that implement the `Device` trait (attach/detach) and are passed down by the DeviceManager. None of these two are true for VsockDevice. VsockDevice is a special case device that is useful internally for qemu for agent communication(currently) and will never be handled/created by the Device Manager. Nor will the device really need to implement the Device trait with attach and detach since it does not make sense to hotplug/unplug this device. What we really want is initializing code to get the guest cid. 

From what I have seen from a quick walkthrough of @pmores PR the breakage is due to the fact that he currently maintains a device vector of type DeviceType and stores the VsockDevice inside it which would not work due to change in my PR (and my argument above for the change). My suggestion would be since we only need Vsock for agent communication(for now), would be to store this as an expicit reference to VsockDevice in the QemuInner struct and not handle this in the device list that is semantically meant to handle devices that can be attached/detached. This can then be used in the `get_agent_socket` and `get_agent_vsock_dev` methods. Something like: 
```
#[derive(Debug)]
  pub struct QemuInner {
      /// sandbox id
      id: String,
    
      qemu_process: Option<Child>,
  
      config: HypervisorConfig,
      devices: Vec<DeviceType>,
      vsock: VsockDevice,
  }
```

pmores: Hi @amshinde ! I'm aware that the problem this PR creates in the qemu driver can be solved by special-casing, like just about any problem for that matter, I'm not sure though that this is warranted.

I'm not aware of the close semantic ties in the runtime-rs design and current practice that you describe between `DeviceType`, the `Device` trait and `DeviceManager`.  Also from Rust standpoint, `DeviceType` is an enum so no relations among the variant types are implied.  In short, I don't see that a `DeviceType` variant implies cloneability or attach/detach functionality.

Also, could you explain the specific problem that `DeviceType::Vsock` causes to you? If `add_device()` now takes and returns a `DeviceType` by value, I think you can do that with copyable and non-copyable types just as well.

What do other runtime-rs devs think, @bergwolf , @studychao , @Tim-0731-Hzt perhaps?

PR #6056, opened 2023-01-12
arm64/CI: fix unit test failure on arm64

jongwu:
There are lots of unit test failure on arm64. fix or skip them to unblock the CI.

PR #8289, opened 2023-10-23
agent: Skip flaky create_tmpfs on s390x

BbolroC:
This is to skip a flaky test `create_tmpfs()` on s390x until a root cause is identified and fixed.

It would be great to get this backported onto `CCv0` to prevent the hiccup.

Fixes: #4248

### DIFF
--- a/src/agent/src/watcher.rs
+++ b/src/agent/src/watcher.rs
@@ -1291,7 +1291,7 @@ mod tests {
 
     #[tokio::test]
     #[serial]
-    #[cfg(not(target_arch = "aarch64"))]
+    #[cfg(not(any(target_arch = "aarch64", target_arch = "s390x")))]
     async fn create_tmpfs() {
         skip_if_not_root!();
 


src/agent/src/watcher.rs
```diff
@@ -1291,7 +1291,7 @@ mod tests {
 
     #[tokio::test]
     #[serial]
-    #[cfg(not(target_arch = "aarch64"))]
+    #[cfg(not(any(target_arch = "aarch64", target_arch = "s390x")))]
```
gkurz, line 1294: Are the tests failing for the same reasons ? We have #4248 for s390x but I couldn't find anything for aarch64 (Cc @jongwu ). It would be good to have the details in a comment to avoid the need for `git blame`.

gkurz (approved):
Approving to unblock the release. Thanks @BbolroC !

src/agent/src/watcher.rs
BbolroC: Sorry. A root cause has not yet been identified for `s390x`. I have no idea regarding how the failure was triggered for `aarch64`.

BbolroC: Based on https://github.com/kata-containers/kata-containers/pull/6056#issuecomment-1382613020, it seems there is a dependency issue while running the tests concurrently.

amshinde (approved):

PR #8290, opened 2023-10-23
agent: rustjail's rust implementation of OCI runtime spec update

yawqi:
In this PR, the rustjail's rust implementation of OCI runtime spec has been updated to be consistent with the latest OCI runtime spec. This mainly consists of three parts:
- Update the rustjail lib according to OCI runtime spec's go implementation.
- Update the protocol lib's trans mod, which is used to convert between OCI implementation and grpc implementation.
- Relevant dependent modules and binaries.


src/agent/rustjail/src/cgroups/fs/mod.rs
```diff
@@ -23,7 +23,7 @@ use crate::container::DEFAULT_DEVICES;
 use anyhow::{anyhow, Context, Result};
 use libc::{self, pid_t};
 use oci::{
-    LinuxBlockIo, LinuxCpu, LinuxDevice, LinuxDeviceCgroup, LinuxHugepageLimit, LinuxMemory,
+    LinuxBlockIO, LinuxCPU, LinuxDevice, LinuxDeviceCgroup, LinuxHugepageLimit, LinuxMemory,
```
bergwolf, line 26: Could you split the renaming part of the patch into a separate commit? So that it is clear what is renamed and what is updated.

bergwolf (commented):
otherwise lgtm, thanks!

src/agent/rustjail/src/cgroups/fs/mod.rs
yawqi: Thanks, fixed. :)

Issue #8313, opened 2023-10-25
kata-runtime/kata-ctl: Add confidential guest to output


jodh-intel:
Add the hypervisor security details to the output of `kata-runtime env` and `kata-ctl env` to show the user/admin the value of the `confidential_guest` setting.

PR #8314, opened 2023-10-25
kata-runtime/kata-ctl: Add security details to output

jodh-intel:
Add the hypervisor security details to the output of the `kata-runtime env` and `kata-ctl env` commands so the user can see, amongst other things, the value of `confidential_guest`.

Fixes: #8313.


cmaf (approved):
thanks @jodh-intel

amshinde (approved):

Issue #8321, opened 2023-10-26
kata-manager: Fails to determine download URL for Kata Containers release


jodh-intel:
[`kata-manager.sh`](https://github.com/kata-containers/kata-containers/blob/main/utils/kata-manager.sh) currently fails to determine the latest release:

```
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh) -o -f";echo $?
INFO: Checking dependencies
INFO: Downloading Kata Containers release (latest version)
ERROR: Cannot determine download URL for version 3.2.0 (https://api.github.com/repos/kata-containers/kata-containers/releases)
ERROR: Cannot determine download URL for version 3.2.0
ERROR: Cannot determine Kata Containers release file
1
```

PR #8323, opened 2023-10-26
utils: kata manager: Fix version checks

jodh-intel:
The archive names for x86_64 [Kata releases](https://github.com/kata-containers/kata-containers/releases) used to include the tag `x86_64`, but that has now been changed to `amd64`, which unfortunately broke `kata-manager.sh`:

```
kata-static-3.1.3-x86_64.tar.xz
                  ~~~~~~
                  expected

kata-static-3.2.0-alpha3-x86_64.tar.xz
                         ~~~~~~
                         expected

kata-static-3.2.0-alpha4-amd64.tar.xz
                         ~~~~~
                         changed
```

PR also includes lint fixes and a follow-on fix for installing containerd.

Fixes: #8321.

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -145,6 +145,7 @@ github_get_release_file_url()
 
 	local arch
 	arch=$(uname -m)
+	[ "$arch" = "x86_64" ] && arch="amd64"
 
 	local regex=""
 
@@ -154,7 +155,6 @@ github_get_release_file_url()
 			;;
 
 		*containerd*)
-			[ "$arch" = "x86_64" ] && arch="amd64"
 			regex="containerd-.*-linux-${arch}.tar.gz"
 			;;
 
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -168,7 +168,7 @@ github_get_release_file_url()
 		-r '.[] | select(.tag_name == $version) | .assets[].browser_download_url' |\
 		grep "/${regex}$")
 
-        download_url=$(echo "$download_url" | awk '{print $1}')
+	download_url=$(echo "$download_url" | awk '{print $1}')
 
 	[ -z "$download_url" ] && die "Cannot determine download URL for version $version ($url)"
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
It would be great if folk could test this on a spare (non critical! ;) system that ideally doesn't already have Kata or containerd installed (if it does, use `-f` to blow them).

Minimally, once you've checked out this branch:

```bash
# Show options
$  utils/kata-manager.sh -h

# Install Kata + Containerd
$  utils/kata-manager.sh
```

See https://github.com/kata-containers/kata-containers/tree/main/utils for further details.

utils/kata-manager.sh
```diff
@@ -145,6 +145,7 @@ github_get_release_file_url()
 
 	local arch
 	arch=$(uname -m)
+	[ "$arch" = "x86_64" ] && arch="amd64"
```
fidencio, line 152: Out of curiosity, James, `kata-manager` is not aware of other arches, is it?
The reason I'm asking is because we most likely would have to do the same trick for aarch64 -> arm64.

fidencio:
lgtm, apart from my comment.

dborquez (approved):
LGTM, thank you @jodh-intel

fidencio (changes requested):
Let me add the request changes, as it will ensure the comment will be addressed.

utils/kata-manager.sh
jodh-intel: > Out of curiosity, James, `kata-manager` is not aware of other arches, is it?

Correct. It would be up to the other architecture maintainers to add the (very minor) tweaks necessary for other architectures. And to test on those architectures of course :smile:

fidencio: Hmm.  Okay.  I personally think that knowing it'd be needed for ARM adding the aarch64 -> arm64 switch is very cheap and could be done on this PR.  But I won't block it if you don't fancy addressing it.

fidencio (approved):
lgtm, thnks @jodh-intel!

amshinde (approved):

Issue #8328, opened 2023-10-27
runitme-rs: kata pod with multi-containers sharing one direct volume resulting in wrong behaviour


Apokleos:
# Get your issue reviewed faster
We discovered that two containers sharing the same direct volume in a pod can result in incorrect mount information in the guest.
```bash
# kata-ctl exec ff86fdec025e52a2f048e5b3c1ef4e91562857f1ae2374b3da5642d3d4f1080b
root@localhost:/# mount
/dev/vda1 on / type ext4 (ro,relatime,errors=remount-ro,data=ordered)
devtmpfs on /dev type devtmpfs (rw,relatime,size=1019040k,nr_inodes=254760,mode=755)
...
/dev/vdb on /run/kata-containers/shared/containers/passthrough/a14a83d6f4-683160fb-disk_xcdix0009 type ext4 (rw,relatime)
/dev/vdb on /run/kata-containers/shared/containers/passthrough/434e8c2967-f08e1cbd-disk_xcdix0009 type ext4 (rw,relatime)
...
```
When multiple containers in a kata pod share one direct volume, it's important to make sure that the corresponding block device is only mounted once in the guest. This means that there should be only one mount entry for the device in the mount information.

# Description of problem

pod.yaml
```bash
apiVersion: v1
kind: Pod
metadata:
  name: kata-direct-x0004
spec:
  runtimeClassName: kata
  containers:
  - name: kata-direct-container-01
    image: docker.io/library/ubuntu:22.04 
    volumeMounts:
    - name: directvol-test-volx03
      mountPath: /disk_xcdix0009
  - name: kata-direct-container-02
    image: docker.io/library/ubuntu:22.04
    volumeMounts:
    - name: directvol-test-volx03
      mountPath: /disk_xcdix0009

  volumes:
  - name: directvol-test-volx03 
    emptyDir: {}
```

```bash
# kubectl get po
NAME                    READY   STATUS    RESTARTS   AGE
kata-direct-cdi-x0004   2/2     Running   0          72m
```

```bash
# kata-ctl exec ff86fdec025e52a2f048e5b3c1ef4e91562857f1ae2374b3da5642d3d4f1080b
root@localhost:/# mount
/dev/vda1 on / type ext4 (ro,relatime,errors=remount-ro,data=ordered)
devtmpfs on /dev type devtmpfs (rw,relatime,size=1019040k,nr_inodes=254760,mode=755)
...
[...]

PR #8332, opened 2023-10-28
runitme-rs/bugfix: kata pod with multi-containers sharing one direct volume

Apokleos:
When multiple containers in a kata pod share one direct volume, it's important to make sure that the corresponding block device is only mounted once in the guest. This means that there should be only one mount entry for the device in the mount information.

Fixes: #8328


quanweiZhou (approved):
Thanks, @Apokleos, LGTM.

Tim-0731-Hzt (approved):

amshinde (approved):

Issue #8333, opened 2023-10-28
Ipvlan and  macvlan networks are not supported with nerdctl and docker since adding support with shimv2


amshinde:
When trying to use ipvlan network with nerdctl you get the following error:

```
$ sudo nerdctl network create ipvlan70 --driver ipvlan --subnet=10.54.74.0/24 -o parent=eno1
$ sudo nerdctl run -it --rm --network ipvlan70 --runtime io.containerd.kata-clh.v2 alpine
FATA[0000] failed to create shim task: IPVlanEndpoint does not support Hot attach: unknown
```

PR #8334, opened 2023-10-28
network: Fix network attach for ipvlan and macvlan

amshinde:
We used the approach of cold-plugging network interface for pre-shimv2 support for docker.Since the hotplug approach was not required, we never really got to implementing hotplug support for certain network endpoints, ipvlan and macvlan being among them.

Since moving to shimv2 interface as the default for runtime, we switched to hotplugging the network interface for supporting docker and nerdctl. This was done for veth endpoints only.

Implement the hot-attach apis for ipvlan and macvlan as well to support ipvlan and macvlan networks with docker and nerdctl.

Fixes: #8333

### DIFF
--- a/src/runtime/virtcontainers/ipvlan_endpoint.go
+++ b/src/runtime/virtcontainers/ipvlan_endpoint.go
@@ -125,14 +125,42 @@ func (endpoint *IPVlanEndpoint) Detach(ctx context.Context, netNsCreated bool, n
 	})
 }
 
-// HotAttach for ipvlan endpoint not supported yet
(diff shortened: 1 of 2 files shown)


amshinde:
Follow-up for this PR would be to add a test with nerdtcl for ipvlan and macvlan networking.

amshinde:
cc @fidencio

fidencio (approved):
lgtm, thanks @amshinde!

@Greatsamps has given it a try and it seems to work as expected on his setup!

dborquez (approved):
LGTM, thank you @amshinde

Issue #8336, opened 2023-10-30
ci-metrics: fix the helper function that removes all kata components before running a metrics test


dborquez:
Instead of using SIGTERM signal, use SIGKILL signal to successfully remove each kata component

PR #8338, opened 2023-10-30
metrics: Fix function that completely stops kata containers before running a test

dborquez:
This PR makes the change to using the SIGKILL signal instead of SIGTERM to force stop each kata component before start running any metric test.

Fixes: #8336

### DIFF
--- a/tests/metrics/lib/common.bash
+++ b/tests/metrics/lib/common.bash
@@ -176,13 +176,12 @@ function init_env()
 
 	# restart docker only if it is not masked by systemd
 	docker_masked="$(systemctl list-unit-files --state=masked | grep -c docker)" || true
-	if [ "${docker_masked}" -eq 0 ]; then
-		sudo systemctl restart docker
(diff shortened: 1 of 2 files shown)


tests/metrics/lib/common.bash
```diff
@@ -201,9 +197,12 @@ function kill_processes_before_start()
 	CTR_PROCS=$(sudo "${CTR_EXE}" t list -q)
 	[[ -n "${CTR_PROCS}" ]] && clean_env_ctr
 
-	kill_kata_components
+	restart_containerd_service
```
amshinde, line 200: Do need containerd restart twice? I see you are restarting on line 200 and later on line 205 as well?

dborquez: @amshinde containerd just need to be restarted once, thanks.

cmaf (approved):

amshinde (approved):

Issue #8340, opened 2023-10-31
"CPUShares is out of range" when start kata using docker


jongwu:
Start kata using docker get error on aarch64:

```
docker run --cpu-shares 0 --runtime io.containerd.run.kata.v2 -it --rm docker.io/library/ubuntu:latest bash
docker: Error response from daemon: failed to create task for container: failed to create shim task: failed to set unit docker-cd44679e55db4c83a93ffec3aad5dc39a5240fdc925040d12fd89fa2e35f626f.scope properties

Caused by:
    org.freedesktop.DBus.Error.InvalidArgs: Value specified in CPUShares is out of range: unknown.
```
This is arch agnostic issue.

PR #8341, opened 2023-10-31
agent: correct CPUShares and CPUWeight value

jongwu:
If cgroup driver is systemd, CPUShares, for cgroup v1, should be at least 2 [1] and CPUWeight for cgroup v2, should be at least 1 [2].

Fixes: #8340

[1] https://github.com/systemd/systemd/blob/d19434fbf81db04d03c8cffa87821f754a86635b/src/basic/cgroup-util.h#L122
[2] https://github.com/systemd/systemd/blob/d19434fbf81db04d03c8cffa87821f754a86635b/src/basic/cgroup-util.h#L91

### DIFF
--- a/src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
+++ b/src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
@@ -6,7 +6,7 @@
 use super::super::common::{CgroupHierarchy, Properties};
 use super::transformer::Transformer;
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use oci::{LinuxCpu, LinuxResources};
 use zbus::zvariant::Value;
 
@@ -14,6 +14,7 @@ const BASIC_SYSTEMD_VERSION: &str = "242";
 const DEFAULT_CPUQUOTAPERIOD: u64 = 100 * 1000;
 const SEC2MICROSEC: u64 = 1000 * 1000;
 const BASIC_INTERVAL: u64 = 10 * 1000;
+const CGROUP_CPU_SHARES_MAX: u64 = 262144;
 
 pub struct Cpu {}
 
@@ -50,6 +51,12 @@ impl Cpu {
         systemd_version: &str,
     ) -> Result<()> {
         if let Some(shares) = cpu_resources.shares {
+            // Minimum value of CPUShares should be 2, see https://github.com/systemd/systemd/blob/d19434fbf81db04d03c8cffa87821f754a86635b/src/basic/cgroup-util.h#L122
+            let shares = match shares {
+                0 => 1024,
+                2..=CGROUP_CPU_SHARES_MAX => shares,
+                _ => bail!("Invalid CpuShares"),
+            };
             properties.push(("CPUShares", Value::U64(shares)));
         }
 
@@ -80,7 +87,7 @@ impl Cpu {
         systemd_version: &str,
     ) -> Result<()> {
         if let Some(shares) = cpu_resources.shares {
-            let weight = shares_to_weight(shares);
+            let weight = shares_to_weight(shares).unwrap();
             properties.push(("CPUWeight", Value::U64(weight)));
         }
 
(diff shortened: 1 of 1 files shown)


cmaf (approved):
thanks @jongwu

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
```diff
@@ -81,7 +84,9 @@ impl Cpu {
     ) -> Result<()> {
         if let Some(shares) = cpu_resources.shares {
             let weight = shares_to_weight(shares);
-            properties.push(("CPUWeight", Value::U64(weight)));
```
bergwolf, line 84: Could you add a reference to https://github.com/systemd/systemd/blob/d19434fbf81db04d03c8cffa87821f754a86635b/src/basic/cgroup-util.h#L91 like the one for CPUShares? Thanks!

jongwu: Thanks @bergwolf -, fixed.

bergwolf (approved):
lgtm, thanks!

amshinde (approved):

jongwu:
Can we merge this? @bergwolf

jongwu:
Hello, can we get this merged?

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
```diff
@@ -50,7 +50,10 @@ impl Cpu {
         systemd_version: &str,
     ) -> Result<()> {
         if let Some(shares) = cpu_resources.shares {
-            properties.push(("CPUShares", Value::U64(shares)));
+            // Minimum value of CPUShares should be 2, see https://github.com/systemd/systemd/blob/d19434fbf81db04d03c8cffa87821f754a86635b/src/basic/cgroup-util.h#L122
+            if shares >= 2 {
+                properties.push(("CPUShares", Value::U64(shares)));
+            }
```
gkurz, line 56: So this is basically ignoring an invalid value received from the runtime... not correcting anything unlike the title seems to suggest. Is it appropriate to continue execution ? FWIW the memory subsystem bails out when receiving an invalid value. See https://github.com/kata-containers/kata-containers/blob/main/src/agent/rustjail/src/cgroups/systemd/subsystem/memory.rs#L42 .

gkurz (changes requested):
Blocking this PR because it has a lot of approvals already but I'd really like to understand why invalid values are ignored instead of being corrected or at least logged.

@jongwu @bergwolf @cmaf @amshinde ?

jongwu:
Thanks @gkurz -,  Updated.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
gkurz: On second thought, cgroupsv2 treats 0 as a hint to use the default value. Maybe better to do the same in cgroupsv1 for consistency ? WDYT @jongwu @cmaf ?

jongwu: What's the default value of cpuShares? 100?

gkurz: > What's the default value of cpuShares? 100?

See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/sched/sched.h#n409

```
/*
 * A weight of 0 or 1 can cause arithmetics problems.
 * A weight of a cfs_rq is the sum of weights of which entities
 * are queued on this cfs_rq, so a weight of a entity should not be
 * too large, so as the shares value of a task group.
 * (The default weight is 1024 - so there's no practical
 *  limitation from this.)
 */
#define MIN_SHARES		(1UL <<  1)
#define MAX_SHARES		(1UL << 18)
```
You can :
- convert runtime input of 0 to 1024
- bail out if value is not in the MIN_SHARES...MAX_SHARES range


 and the value should be in the 2..262144 range.

src/agent/rustjail/src/cgroups/systemd/subsystem/cpu.rs
```diff
@@ -81,7 +84,10 @@ impl Cpu {
     ) -> Result<()> {
         if let Some(shares) = cpu_resources.shares {
             let weight = shares_to_weight(shares);
-            properties.push(("CPUWeight", Value::U64(weight)));
+            // CPUWeigth can not be 0. See https://github.com/systemd/systemd/blob/d19434fbf81db04d03c8cffa87821f754a86635b/src/basic/cgroup-util.h#L91
```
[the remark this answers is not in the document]
gkurz: Similarly to cpu shares, cpu weight has a validity range.

See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/cgroup-v2.rst#n646
```
All weights are in the range [1, 10000] with the default at 100.  This
allows symmetric multiplicative biases in both directions at fine
enough granularity while staying in the intuitive range.
```

Maybe bail out if value is out of range, which happens if passed cpu shares is > 262144, for consistency.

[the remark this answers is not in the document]
jongwu: Great! Changed.

gkurz (approved):
LGTM. Thanks @jongwu !

Issue #8350, opened 2023-11-01
 The Cargo.lock for agent needs to be updated to include   "safe-path" dependency.


amshinde:
After running make in the agent repo, the Cargo.lock is not the same as we have in our repo. I am surprised that was not caught by the checks. I realised this when running VScode that compiles code in the background, and seeing the git branch I was workinf on never in a clean state.

PR #8351, opened 2023-11-01
cargo: Agent cargo.lock updated

amshinde:
The Cargo.lock for agent needs to be updated to include "safe-path" dependency.

Fixes: #8350


fidencio (approved):
lgtm, thanks @amshinde.

I wonder how we didn't catch this one as part of the CI, or whether we're missing a Ci to check for those updates.

bergwolf (approved):

Issue #8356, opened 2023-11-02
kata-manager | Restrict containerd installation options to either install "lts" or "active" versions of the project


fidencio:
Currently kata-manager provides the user a way to specify a version of containerd to be used, which is nice for debugging purposes but should **NOT** be the responsibility of kata-manager to be switching containerd versions left and right.

What we should do instead is:
* Allow the user to specify "lts" or "active", and we handle on our side which is the latest release from those to be pulled in.
* If user does **NOT** specify such a version, we should instead:
  * Rely on the containerd installed in the system, if it's installed, and just have kata-containers installed
  * Install the LTS version of containerd, in case containerd is not present in the system

By doing this we ensure a user will always have a system up and running, ready to use kata-containers, even if nothing is specified (as done Today).

PR #8365, opened 2023-11-02
kata-manager: Accept only "lts" or "active" as containerd versions

fidencio:
kata-manager is a very nice tool, but we shouldn't be trying to take
care of "everything" in "all possible scenarios", and we should focus on
installing Kata Containers dependencies that are supported.

With this in mind, let's limit a little bit the scope of which versions
of containerd can be installed, limitting to "active" and "lts", which
will then install the latest version of those "flavours".  The default
value will always be "lts" as that's supposed to be the stable one.

Fixes: #8356

While here, we're also adding a commit to ensure that `apt-get update` runs before `apt-get install`.

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -25,6 +25,7 @@ readonly containerd_project_url="https://github.com/${containerd_slug}"
 
 readonly kata_releases_url="https://api.github.com/repos/${kata_slug}/releases"
 readonly containerd_releases_url="https://api.github.com/repos/${containerd_slug}/releases"
+readonly containerd_io_releases_url="https://raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md"
 
 # Directory created when unpacking a binary release archive downloaded from
 # $kata_releases_url.
@@ -239,7 +240,9 @@ Description: Install $kata_project [1] (and optionally $containerd_project [2])
 
 Options:
 
- -c <version> : Specify containerd version.
+ -c <flavour> : Specify containerd flavour ("lts" | "active" - default: "lts").
+                Find more details on LTS and Active versions of containerd on
+                https://containerd.io/releases/#support-horizon
  -d           : Enable debug for all components.
  -f           : Force installation (use with care).
  -h           : Show this help statement.
@@ -669,9 +672,25 @@ handle_kata()
 	kata-runtime --version
 }
 
+containerd_version_number()
+{
+	local flavour="${1:-}"
+	[ -z "$flavour" ] && die "need containerd flavour"
+
+	base_version="$(curl -fsSL $containerd_io_releases_url | \
+		grep -iE "\[d+.d+\].*| $flavour" | \
+		grep -oE "\[[0-9]+.[0-9]+\]" | \
+		grep -oE "[0-9]+.[0-9]+")"
+
+	curl --silent ${containerd_releases_url} | \
+		jq -r .[].tag_name | \
+		grep "^v${base_version}.[0-9]*$" -m1
+}
+
 handle_containerd()
 {
-	local version="${1:-}"
+	local flavour="${1:-}"
+	[ -z "$flavour" ] && die "need containerd flavour"
 
 	local force="${2:-}"
 	[ -z "$force" ] && die "need force value"
@@ -679,6 +698,7 @@ handle_containerd()
 	local enable_debug="${3:-}"
 	[ -z "$enable_debug" ] && die "no enable debug value"
 
+	local version="$(containerd_version_number "$flavour")"
 	local ret
 
 	if [ "$force" = "true" ]
@@ -755,7 +775,7 @@ handle_installation()
 
 	# These params can be blank
 	local kata_version="${7:-}"
-	local containerd_version="${8:-}"
+	local containerd_flavour="${8:-}"
 
 	[ "$only_run_test" = "true" ] && test_installation && return 0
 
@@ -765,7 +785,7 @@ handle_installation()
 
 	[ "$skip_containerd" = "false" ] && \
 		handle_containerd \
-		"$containerd_version" \
+		"$containerd_flavour" \
 		"$force" \
 		"$enable_debug"
 
@@ -781,6 +801,14 @@ handle_installation()
 	echo -e "\n${warnings}\n"
 }
 
+validate_containerd_flavour()
+{
+	local flavour="${1:-}"
+	local flavours_regex='lts|active'
+
+	grep -qE "$flavours_regex" <<< "$flavour" || die "expected flavour to match '$flavours_regex', found '$flavour'"
+}
+
 handle_args()
 {
 	local cleanup="true"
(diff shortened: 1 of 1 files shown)


fidencio:
@jodh-intel, I went back and forth on using "lts" or "active" as the version to be installed when nothing is passed, and I'm fine either way.  So, please, let me know your preference.

jodh-intel:
> @jodh-intel, I went back and forth on using "lts" or "active" as the version to be installed when nothing is passed, and I'm fine either way. So, please, let me know your preference.

I'd vote for "lts" and ensure that we mention it in the usage and what it stands for.

utils/kata-manager.sh
```diff
@@ -669,16 +669,26 @@ handle_kata()
 	kata-runtime --version
 }
 
+containerd_version_number()
+{
+	local flavour="${1:-lts}"
+
+	base_version="$(curl -fsSL https://raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md | grep -iE "\[d+.d+\].*| $flavour" | grep -oE "\[[0-9]+.[0-9]+\]" | grep -oE "[0-9]+.[0-9]+")"
+
```
jodh-intel, line 679: Could you fold this long line for clarity?

```suggestion
	base_version="$(curl -fsSL https://raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md |\
	grep -iE "\[d+.d+\].*| $flavour" |\
	grep -oE "\[[0-9]+.[0-9]+\]" |\
	grep -oE "[0-9]+.[0-9]+")"

```

utils/kata-manager.sh
```diff
@@ -239,7 +239,7 @@ Description: Install $kata_project [1] (and optionally $containerd_project [2])
 
 Options:
 
- -c <version> : Specify containerd version.
+ -c <flavour> : Specify containerd flavour ("lts" | "active" - default: "lts")
```
jodh-intel: I'd consider defining "lts". Something like:

```suggestion
 -c <flavour> : Specify containerd flavour ("lts" | "active"
                Defaults to "lts" for the Long Term Support (LTS) version which is appropriate for most users.
```

utils/kata-manager.sh
```diff
@@ -669,16 +669,26 @@ handle_kata()
 	kata-runtime --version
 }
 
+containerd_version_number()
+{
+	local flavour="${1:-lts}"
```
jodh-intel, line 676: I think it would be better to set the default value of "lts" in `handle_args()` since:

- That avoids having to set the default twice - here and in `validate_containerd_flavour()`.
- I think the code would be clearer since flavour would always have a value rather than the current "magic empty string" behaviour.

jodh-intel (commented):
Thanks @fidencio ! A few comments...

fidencio:
> I'd vote for "lts" and ensure that we mention it in the usage and what it stands for.

I've updated the patch, pointing users to https://containerd.io/releases/#support-horizon for more details.
They can explain better what each one of the options mean for their project than us. :-)

utils/kata-manager.sh
fidencio: Oh, I've updated the patch while you're reviewing it, and decided to point users to: https://containerd.io/releases/#support-horizon

Is that fine by you?

utils/kata-manager.sh
fidencio: I totally agree, I've addressed that!

utils/kata-manager.sh
fidencio: Done, thanks for the suggestion!

fidencio:
> It might be worth stating that this is a breaking change in the commit.

I've added a NOTE about this in the commit.

utils/kata-manager.sh
```diff
@@ -669,16 +671,31 @@ handle_kata()
 	kata-runtime --version
 }
 
+containerd_version_number()
+{
+	local flavour="$1"
```
jodh-intel, line 676: Now that `flavour` has to be set, could you add a check on it. Something like:

```suggestion
	local flavour="${1:-}"
	[ -n "$flavour" ] || die "need containerd flavour"
```

utils/kata-manager.sh
```diff
@@ -669,16 +671,31 @@ handle_kata()
...
+
+	base_version="$(curl -fsSL https://raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md | \
```
jodh-intel, line 678: Nit: URL would be better as a global `readonly`.

utils/kata-manager.sh
```diff
@@ -815,7 +842,10 @@ handle_args()
 	shift $[$OPTIND-1]
 
 	[ -z "$kata_version" ] && kata_version="${1:-}" || true
-	[ -z "$containerd_version" ] && containerd_version="${2:-}" || true
+	[ -z "$containerd_flavour" ] && containerd_flavour="${2:-}" || true
+
+	[ -z "$containerd_flavour" ] && containerd_flavour="lts"
```
jodh-intel, line 847: Fine, or you could just set `local containerd_flavour='lts'` above to avoid the need for this line.

jodh-intel (commented):

stevenhorsman (approved):
Looks okay to me, but as a disclaimer I've not run it manually to check

utils/kata-manager.sh
```diff
@@ -25,6 +25,7 @@ readonly containerd_project_url="https://github.com/${containerd_slug}"
 
 readonly kata_releases_url="https://api.github.com/repos/${kata_slug}/releases"
 readonly containerd_releases_url="https://api.github.com/repos/${containerd_slug}/releases"
+readonly containerd_io_releases_url="//raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md"
```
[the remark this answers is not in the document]
fidencio: Ops, bad c&p. :-)

utils/kata-manager.sh
```diff
@@ -25,6 +25,7 @@ readonly containerd_project_url="https://github.com/${containerd_slug}"
 
 readonly kata_releases_url="https://api.github.com/repos/${kata_slug}/releases"
 readonly containerd_releases_url="https://api.github.com/repos/${containerd_slug}/releases"
+readonly containerd_io_releases_url="https//raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md"
```
jodh-intel, line 28: Missing colon in URL which breaks the download:

```suggestion
readonly containerd_io_releases_url="https://raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md"
```

Luckily manual testing caught this one ;)

```bash
+++ curl -fsSL https//raw.githubusercontent.com/containerd/containerd.io/main/content/releases.md
+++ grep -iE '\[d+.d+\].*| lts'
+++ grep -oE '\[[0-9]+.[0-9]+\]'
+++ grep -oE '[0-9]+.[0-9]+'
curl: (6) Could not resolve host: https
```

jodh-intel (commented):
Thanks @fidencio - Manual testing found a bug ;)

utils/kata-manager.sh
fidencio: Fixed, thanks!

jodh-intel (approved):
Thanks @fidencio - lgtm.

Issue #8371, opened 2023-11-03
arm64: make agent-opa-tarball 


zvonkok:
Building the agent-opa-tarball fails on aarch64 and Alpine base container image 
PKG musl-dev installs `/usr/bin/aarch64-alpine-linux-musl-gcc` but we need `"/usr/bin/$(uname -m)-linux-musl-gcc"`

```
  Compiling parking_lot_core v0.8.5
   Compiling libseccomp-sys v0.2.1
   Compiling libseccomp v0.3.0
   Compiling slog-async v2.7.0
   Compiling prometheus v0.13.1
error: linker `aarch64-linux-musl-gcc` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `log` (build script) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `futures-core` (build script) due to previous error
error: could not compile `futures-channel` (build script) due to previous error
error: could not compile `futures-task` (build script) due to previous error
error: could not compile `futures-util` (build script) due to previous error
error: could not compile `parking_lot_core` (build script) due to previous error
error: could not compile `memchr` (build script) due to previous error
```

PR #8372, opened 2023-11-03
arm64: make agent-opa-tarball fix

zvonkok:
Building the agent-opa-tarball fails on aarch64 and Alpine base container image PKG musl-dev installs /usr/bin/aarch64-alpine-linux-musl-gcc but we need "/usr/bin/$(uname -m)-linux-musl-gcc"

Fixes: #8371

### DIFF
--- a/tools/packaging/static-build/agent/Dockerfile
+++ b/tools/packaging/static-build/agent/Dockerfile
@@ -2,7 +2,9 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM alpine:3.18
+# BUILDPLATFORM - matches the current machine
+FROM --platform=$BUILDPLATFORM alpine:3.18 
+
 ARG RUST_TOOLCHAIN
 
 SHELL ["/bin/ash", "-o", "pipefail", "-c"]
@@ -19,3 +21,9 @@ RUN apk --no-cache add \
         openssl-libs-static \
         protoc && \
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${RUST_TOOLCHAIN}
+
+ARG TARGETARCH
+
+RUN if [ "${TARGETARCH}" = "arm64" ]; then \
+        ln -s /usr/bin/$(uname -m)-alpine-linux-musl-gcc /usr/bin/$(uname -m)-linux-musl-gcc; fi 
+
--- a/tools/packaging/static-build/agent/Dockerfile
+++ b/tools/packaging/static-build/agent/Dockerfile
@@ -25,5 +25,5 @@ RUN apk --no-cache add \
 ARG TARGETARCH
 
 RUN if [ "${TARGETARCH}" = "arm64" ]; then \
-        ln -s /usr/bin/$(uname -m)-alpine-linux-musl-gcc /usr/bin/$(uname -m)-linux-musl-gcc; fi 
+        ln -s "/usr/bin/$(uname -m)-alpine-linux-musl-gcc" "/usr/bin/$(uname -m)-linux-musl-gcc"; fi
 
--- a/tools/packaging/static-build/agent/build.sh
+++ b/tools/packaging/static-build/agent/build.sh
@@ -17,16 +17,16 @@ source "${script_dir}/../../scripts/lib.sh"
 container_image="${AGENT_CONTAINER_BUILDER:-$(get_agent_image_name)}"
 [ "${CROSS_BUILD}" == "true" ] && container_image="${container_image}-cross-build"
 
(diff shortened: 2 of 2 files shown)


fidencio (approved):
lgtm, thanks @zvonkok!

jepio (approved):

tools/packaging/static-build/agent/Dockerfile
```diff
@@ -19,3 +19,5 @@ RUN apk --no-cache add \
         openssl-libs-static \
         protoc && \
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${RUST_TOOLCHAIN}
+
+RUN ln -s /usr/bin/aarch64-alpine-linux-musl-gcc "/usr/bin/$(uname -m)-linux-musl-gcc"
```
jongwu, line 23: What happens when run this on non-aarch64? Should this be "RUN ln -s /usr/bin/$(uname -m)-alpine-linux-musl-gcc "/usr/bin/$(uname -m)-linux-musl-gcc"?

jongwu:
Thanks @zvonkok -,

I run "make agent-opa-tarball" on arm64. Something more serious is there is no muti-arch docker image for agent builder.  quay.io/kata-containers/builders:agent-4ad2cfe0c-148c565b2 in my case, is only for amd64. How to address this? @fidencio

tools/packaging/static-build/agent/Dockerfile
zvonkok: Ha good catch you're right, updating.

jongwu: Should we test this on s390x and powerpc?  As they may not use musl or has different name here.

zvonkok: What about only doing this on aarch64 and ignore other archs this way only needed symlinks are created?

jongwu: Great!

zvonkok:
@jongwu PTAL now it is proper multi-arch :)

zvonkok:
This will not work for multi-arch cross-builds.

jongwu:
OK, thanks @zvonkok .

jongwu (approved):

zvonkok:
Will this work with multi-arch cross-builds?

tools/packaging/static-build/agent/build.sh
```diff
@@ -16,16 +17,16 @@ source "${script_dir}/../../scripts/lib.sh"
 container_image="${AGENT_CONTAINER_BUILDER:-$(get_agent_image_name)}"
 [ "${CROSS_BUILD}" == "true" ] && container_image="${container_image}-cross-build"
 
-sudo docker pull ${container_image} || \
+sudo docker pull --platform=linux/"${target_arch}" "${container_image}" || \
```
zvonkok, line 20: @jongwu This fixes the problem with pulling an amd64 image to an arm64 machine. Once this is merged one of the CI jobs will create that arm64 build and we can reuse it.

danmihai1 (approved):

tools/packaging/static-build/agent/build.sh
```diff
@@ -16,13 +17,13 @@ source "${script_dir}/../../scripts/lib.sh"
...
 
-sudo docker pull ${container_image} || \
+sudo docker pull --platform=linux/${target_arch} ${container_image} || \
 	(sudo docker $BUILDX build $PLATFORM \
 	    	--build-arg RUST_TOOLCHAIN="$(get_from_kata_deps "languages.rust.meta.newest-version")" \
 		-t "${container_image}" "${script_dir}" && \
 	 # No-op unless PUSH_TO_REGISTRY is exported as "yes"
 	 push_to_registry "${container_image}")
-
+exit 1
```
jongwu: why 'exit 1' here?

tools/packaging/static-build/agent/build.sh
jongwu: Good, thanks @zvonkok

tools/packaging/static-build/agent/build.sh
zvonkok: This isn't there anymore.

jongwu: Yes. But it's better to squash "+ exit 1" here with "- exit 1" in next commit. WDYT? 😄

zvonkok: The exit is squashed to the second commit.

zvonkok:

zvonkok:
We can close this the latest changes made by @fidencio on agent caching made it work

PR #8374, opened 2023-11-03
utils: kata-manager: Ensure only one download URL

jodh-intel:
Add an extra sanity check to ensure that only a single download URL is
found for the specified release version.

Fixes: #8364.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #8376, opened 2023-11-03
kata-manager: Add support for Docker CLI installation

fidencio:
Add support for also installing the Docker CLI, giving users the chance
to try Kata Containers with docker in the same way we provide users the
chance to try Kata Containers with `ctr`.

Fixes: https://github.com/kata-containers/kata-containers/issues/8357

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -722,19 +722,22 @@ handle_containerd()
 
 test_installation()
 {
+	local tool="${1:-}"
+	[ -z "$tool" ] && die "The tool to test $kata_project with was not informed"
+
 	info "Testing $kata_project\n"
 
 	sudo kata-runtime check -v
 
 	local image="docker.io/library/busybox:latest"
-	sudo ctr image pull "$image"
+	sudo $tool image pull "$image"
 
 	local container_name="test-kata"
 
 	# Used to prove that the kernel in the container
 	# is different to the host kernel.
 	local container_kernel
-	container_kernel=$(sudo ctr run \
+	container_kernel=$(sudo $tool run \
 		--runtime "$kata_runtime_type" \
 		--rm \
 		"$image" \
@@ -777,7 +780,10 @@ handle_installation()
 	local kata_version="${7:-}"
 	local containerd_flavour="${8:-}"
 
-	[ "$only_run_test" = "true" ] && test_installation && return 0
+	# The tool to be testing the installation with
+	local tool="ctr"
+
+	[ "$only_run_test" = "true" ] && test_installation "$tool"  && return 0
 
 	setup "$cleanup" "$force" "$skip_containerd"
 
@@ -789,7 +795,7 @@ handle_installation()
 		"$force" \
 		"$enable_debug"
 
-	[ "$disable_test" = "false" ] && test_installation
+	[ "$disable_test" = "false" ] && test_installation "$tool"
 
 	if [ "$skip_containerd" = "true" ]
 	then
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -45,6 +45,9 @@ readonly kata_clh_configuration="configuration-clh"
 # Systemd unit name for containerd daemon
 readonly containerd_service_name="containerd.service"
 
+# Containerd configuration file
+readonly containerd_config="/etc/containerd/config.toml"
+
 # Directory in which to create symbolic links
 readonly link_dir=${link_dir:-/usr/bin}
 
@@ -455,8 +458,6 @@ configure_containerd()
 
 	info "Configuring $project"
 
-	local cfg="/etc/containerd/config.toml"
-
 	local systemd_unit_dir="/etc/systemd/system"
 	sudo mkdir -p "$systemd_unit_dir"
 
@@ -488,19 +489,19 @@ configure_containerd()
 	fi
 
 	# Backup the original containerd configuration:
-	sudo mkdir -p "$(dirname $cfg)"
+	sudo mkdir -p "$(dirname $containerd_config)"
 
-	sudo test -e "$cfg" || {
-		sudo touch "$cfg"
-		info "Created $cfg"
+	sudo test -e "$containerd_config" || {
+		sudo touch "$containerd_config"
+		info "Created $containerd_config"
 	}
 
 	local original
-	original="${cfg}-pre-kata-$(date -I)"
+	original="${containerd_config}-pre-kata-$(date -I)"
 
-	sudo grep -q "$kata_runtime_type" "$cfg" || {
-		sudo cp "$cfg" "${original}"
-		info "Backed up $cfg to $original"
+	sudo grep -q "$kata_runtime_type" "$containerd_config" || {
+		sudo cp "$containerd_config" "${original}"
+		info "Backed up $containerd_config to $original"
 	}
 
 	local modified="false"
@@ -512,8 +513,8 @@ configure_containerd()
 		"$(date -Iseconds)" \
 		"$script_name")
 
-	sudo grep -q "$kata_runtime_type" "$cfg" || {
-		cat <<-EOF | sudo tee -a "$cfg"
+	sudo grep -q "$kata_runtime_type" "$containerd_config" || {
+		cat <<-EOF | sudo tee -a "$containerd_config"
 		# $comment_text
 		[plugins]
 		  [plugins."io.containerd.grpc.v1.cri"]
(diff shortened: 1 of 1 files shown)


utils/kata-manager.sh
```diff
@@ -720,26 +721,64 @@ handle_containerd()
...
 
+handle_docker()
+{
+	{ containerd_installed; ret=$?; } || true
+	if [ "$ret" -eq 0 ]
+	then
+		info "Backing up previous containerd configuratinn"
+		local cfg="/etc/containerd/config.toml"
+
+		sudo mv $cfg $cfg.system
```
jodh-intel, line 732: How about appending the date to the backup file too, as we do in `configure_containerd()`? That would mean the script would create 1 file / run, but the timestamps could be useful to help debug issues and I don't think we would need to worry about `ENOSPC`.

jodh-intel (commented):
Thanks @fidencio - A few comments...

utils/kata-manager.sh
```diff
@@ -720,26 +721,64 @@ handle_containerd()
...
-		"$image" \
-		"$container_name" \
-		uname -r || true)
+	cmd="sudo $tool run --runtime "$kata_runtime_type" --rm"
+	case "$tool" in
+		docker)
+			# docker takes the container name as `--name
+			# $container_name`, passed to the run option.
+		       	cmd+=" --name $container_name" ;&
+	esac
```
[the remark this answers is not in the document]
fidencio: I'm adding a `case` here as we will need to expand it (kinda `docker|nerdctl`) in the future.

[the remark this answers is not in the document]
jodh-intel: Ok. Do you (will you) actually need the fall-through behaviour or can we just use the conventional `;;`?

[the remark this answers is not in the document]
fidencio: I don't actually need to fall through, I had the ctr into the case before but I think it's clear to leave it outside, and I forgot to switch back to the conventional behaviour.

utils/kata-manager.sh
fidencio: Done!

utils/kata-manager.sh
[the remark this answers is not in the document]
fidencio: Removed the `fall-through`

utils/kata-manager.sh
```diff
@@ -720,26 +721,74 @@ handle_containerd()
...
+		local cfg="/etc/containerd/config.toml"
+
+		sudo mv $cfg $cfg.system-$(date -Iseconds)
+	fi
+
+	containerd_installed
+
+	local tmpdir
+	tmpdir=$(mktemp -d)
+
```
jodh-intel, line 739: Can you delete these 3 lines as they cause the following error (since `tmpdir` is already defined globally):

```
utils/kata-manager.sh: line 737: local: tmpdir: readonly variable
```

utils/kata-manager.sh
```diff
@@ -720,26 +721,74 @@ handle_containerd()
 	containerd --version
 }
 
+handle_docker()
+{
+	{ containerd_installed; ret=$?; } || true
+	if [ "$ret" -eq 0 ]
+	then
+		info "Backing up previous containerd configuratinn"
```
jodh-intel, line 729:
```suggestion
		info "Backing up previous $containerd_project configuratinn"
```

utils/kata-manager.sh
```diff
@@ -720,26 +721,74 @@ handle_containerd()
...
+		local cfg="/etc/containerd/config.toml"
```
jodh-intel, line 730: Actually, could you make this a global read-only as it's also referenced in `configure_containerd()`.

utils/kata-manager.sh
```diff
@@ -244,6 +244,7 @@ Options:
                 Find more details on LTS and Active versions of containerd on
                 https://containerd.io/releases/#support-horizon
  -d           : Enable debug for all components.
+ -D           : Install Docker CLI.
```
jodh-intel, line 247: Can you explain here that the version installed is the packaged version, not the latest release from GH.

I think this distinction is important given that we install the latest Kata and containerd releases.

jodh-intel: Ah - wait. It's *not* the latest _distro packaged_ version, it's the latest _docker project packaged_ version.

utils/kata-manager.sh
```diff
@@ -720,26 +721,74 @@ handle_containerd()
...
 
+handle_docker()
+{
+	{ containerd_installed; ret=$?; } || true
+	if [ "$ret" -eq 0 ]
+	then
+		info "Backing up previous containerd configuratinn"
+		local cfg="/etc/containerd/config.toml"
+
+		sudo mv $cfg $cfg.system-$(date -Iseconds)
```
jodh-intel, line 732: This fails on my test box as the file doesn't exist, so please can you add a check:

```suggestion
		[ -e "$cfg" ] && sudo mv $cfg $cfg.system-$(date -Iseconds)
```

fidencio:
@jodh-intel, addressed all the comments, and have successfully tested it locally.

utils/kata-manager.sh
```diff
@@ -720,26 +722,69 @@ handle_containerd()
...
+	containerd_installed
+
+	local filename='get-docker.sh'
+	
+	local file
+	file="$tmpdir/$filename"
+	
+	curl -fsSL https://get.docker.com -o "$file"
+	sudo sh "$file"
+
```
jodh-intel, line 743: These lines have some whitespace "damage" (tabs on otherwise empty lines).

utils/kata-manager.sh
```diff
@@ -777,7 +822,21 @@ handle_installation()
...
-	[ "$only_run_test" = "true" ] && test_installation && return 0
+	local install_docker="${9:-}"
+	[ -z "$install_docker" ] && die "no install docker value"
+
+	# The tool to be testing the installation with
+	local tool="ctr"
+
+	if [ "$install_docker" = "true" ]
+	then
+		# The script provided by docker already takes care
```
jodh-intel, line 833: There's a small issue here: what if the user ran `-c lts -D`? The code will ignore the request to install containerd since `-D` takes priority. But that might be confusing to users so we could add a log message here. The secondary problem is that this function doesn't know how to deal with that scenario currently since `containerd_flavour` is always set.

The simplest solution might be to just document that Docker takes priority over containerd in `usage()`.

utils/kata-manager.sh
```diff
@@ -244,6 +247,7 @@ Options:
                 Find more details on LTS and Active versions of containerd on
                 https://containerd.io/releases/#support-horizon
  -d           : Enable debug for all components.
+ -D           : Install Docker CLI.
```
jodh-intel, line 250: This might be more informative?

```suggestion
 -D           : Install Docker server and CLI tooling (takes priority over '-c').
```

jodh-intel (commented):
Thanks @fidencio - Tests ok but a few final comments...

utils/kata-manager.sh
fidencio: I've added an info message.

utils/kata-manager.sh
fidencio: Fixed.

fidencio:
@jodh-intel, this is, yet again, ready for your review.

utils/kata-manager.sh
```diff
@@ -777,7 +822,25 @@ handle_installation()
...
+	local tool="ctr"
+
+	if [ "$install_docker" = "true" ]
+	then
+		if [ "$skip_containerd" = "false" ]
+		then
+			# The script provided by docker already takes care
+			# of properly installing containerd
+			skip_containerd="true"
+			info "Containerd will be instaslled during the Docker installation.\nThe '-c' option is being ignored."
```
jodh-intel, line 838: Nit: Typo fix + I'd keep this all on one line:

```suggestion
			info "Containerd will be installed during the Docker installation ('-c' option ignored)"
```

jodh-intel (approved):
Thanks @fidencio - just one non-blocking nit.

jodh-intel:
@fidencio  - Related: could you tal @ #8374 and #8383 when you get a chance?

fidencio:
> @fidencio - Related: could you tal @ #8374 and #8383 when you get a chance?

I will as soon as this one gets merged.

cmaf (approved):
thanks @fidencio

amshinde (approved):

PR #8389, opened 2023-11-07
runtime: Fix TestCheckHostIsVMContainerCapable unstablity issue

justxuewei:
`TestCheckHostIsVMContainerCapable` removes `sysModuleDir` to simulate a
case that the kernel modules are not loaded. However,
`checkKernelModules()` executes `modprobe <module>` if a module not
found in that directory. Loading those modules is required to be denied
temporarily.

Fixes: #8390

### DIFF
--- a/src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
+++ b/src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"testing"
@@ -17,6 +18,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
+
+const denylistModuleConf = "/etc/modprobe.d/denylist-kata-kernel-modules.conf"
 
 func setupCheckHostIsVMContainerCapable(assert *assert.Assertions, cpuInfoFile string, cpuData []testCPUData, moduleData []testModuleData) {
 	createModules(assert, cpuInfoFile, moduleData)
@@ -322,6 +325,36 @@ func TestCheckHostIsVMContainerCapable(t *testing.T) {
 
 	err = hostIsVMContainerCapable(details)
 	assert.Nil(err)
+
+	// Remove required kernel modules and add them to denylist
+	denylistFile, err := os.Create(denylistModuleConf)
+	assert.Nil(err)
+	succeedToRemoveOneModule := false
+	for mod := range archRequiredKernelModules {
+		cmd := exec.Command(modProbeCmd, "-r", mod)
+		if output, err := cmd.CombinedOutput(); err == nil {
+			succeedToRemoveOneModule = true
+		} else {
+			kataLog.WithField("output", string(output)).Warn("failed to remove module")
+		}
+		// Write the following into the denylist file
+		// blacklist <mod>
+		// install <mod> /bin/false
+		_, err = denylistFile.WriteString(fmt.Sprintf("blacklist %s\ninstall %s /bin/false\n", mod, mod))
+		assert.Nil(err)
+	}
+	denylistFile.Close()
+	assert.True(succeedToRemoveOneModule)
+
+	defer func() {
+		os.Remove(denylistModuleConf)
+	}()
+
+	// remove the modules to force a failure
+	err = os.RemoveAll(sysModuleDir)
+	assert.NoError(err)
+	err = hostIsVMContainerCapable(details)
+	assert.Error(err)
 }
 
 func TestArchKernelParamHandler(t *testing.T) {


src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
```diff
@@ -18,6 +19,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const blacklistModuleConf = "/etc/modprobe.d/blacklist-kata-kernel-modules.conf"
```
fidencio, line 22: s/blacklist/denylist/g

And also everywhere in this patch.

justxuewei: Fixed.

src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
```diff
@@ -322,6 +325,30 @@ func TestCheckHostIsVMContainerCapable(t *testing.T) {
...
+	succeedToRemoveOneModule := false
+	for mod := range archRequiredKernelModules {
+		cmd := exec.Command(modProbeCmd, "-r", mod)
+		if output, err := cmd.CombinedOutput(); err == nil {
+			succeedToRemoveOneModule = true
+		} else {
+			kataLog.WithField("output", string(output)).Warn("failed to remove module")
+		}
+		// Write the following into the denylist file
+		// blacklist <mod>
```
fidencio, line 340: s/blacklist/denylist here.

src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
```diff
@@ -322,6 +325,30 @@ func TestCheckHostIsVMContainerCapable(t *testing.T) {
...
+		// install <mod> /bin/false
+		_, err = denylistFile.WriteString(fmt.Sprintf("blacklist %s\ninstall %s /bin/false\n", mod, mod))
```
fidencio, line 342: And here as well. :-)

justxuewei: This can't be modified to `denylist` since it is a invalid syntax for libkmod.

```bash
$ cat /etc/modprobe.d/denylist-kata-kernel-modules.conf
denylist vhost-net
$ sudo modprobe vhost-net
libkmod: ERROR ../libkmod/libkmod-config.c:712 kmod_config_parse: /etc/modprobe.d/denylist-kata-kernel-modules.conf line 1: ignoring bad line starting with 'denylist'
```

src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
justxuewei: See below.

fidencio (approved):
lgtm, thanks @justxuewei!

studychao (approved):
LGTM

src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
```diff
@@ -322,6 +325,30 @@ func TestCheckHostIsVMContainerCapable(t *testing.T) {
 	err = hostIsVMContainerCapable(details)
 	assert.Nil(err)
 
+	// Remove required kernel modules and add them to denylist
```
gkurz, line 328: I have two remarks here :

1. the "failure" test already removes the files on disk. This means they won't be loadable anymore and that we don't need to bother with a deny list,
2. making sure that the modules are unloaded from memory is rather a requirement of the "failure" test. For clarity, I'd move this block under the `// remove the modules to force a failure` comment.

gkurz (changes requested):
Blocking this temporarily as it already has two approvals.

gkurz:
@justxuewei any idea how this test can even pass without your fix ?

justxuewei:
Things are getting complicated now. Commit dd530ba8ee279d620b308c5ece5e450f9c1090b0 removes the failure test. Runtime's CI tests, `make test` and `sudo -E PATH="$PATH" make test`, will pass at the expense of declining coverage. This patch is still valuable to be merged.

src/runtime/cmd/kata-runtime/kata-check_amd64_test.go
justxuewei: Removing module files is not enough for passing the "failure" test. Let me give a more details about that:

1. `sysModuleDir` is replaced with a temp dir, e.g. "/tmp/tmp.xxGmSePqj2".
2. Call `createModules()`. The function just creates **regular files** to simulate system modules. https://github.com/kata-containers/kata-containers/blob/3b2fb6a604a624f1ec0c21a85d6e4182344aac4c/src/runtime/cmd/kata-runtime/kata-check_test.go#L68-L71
3. Remove `sysModuleDir` under the "// remove the modules to force a failure" comment.
4. Call `haveKernelModule()`. This function checks existence of a kernel module. Three steps are involved:
    1. If a kernel module exists at `sysModuleDir`, return true. In this case, it won't return as the dir was removed. https://github.com/kata-containers/kata-containers/blob/3b2fb6a604a624f1ec0c21a85d6e4182344aac4c/src/runtime/cmd/kata-runtime/kata-check.go#L139-L142
    2. Check if the current user is root. (Our test is started with `sudo`, it won't return here.)
    3. Try to load the module. Returns false if it fails. Please note that if a module was loaded in the memory or wasn't added to blacklist, loading module will be succeed. https://github.com/kata-containers/kata-containers/blob/3b2fb6a604a624f1ec0c21a85d6e4182344aac4c/src/runtime/cmd/kata-runtime/kata-check.go#L152-L156
    4. Return true.
 
I admit adding the blacklist does more dirty work in the test, but I have no idea how to pass the test without this.

justxuewei: Plus, the blacklist file is restored after the function is completed. It won't poison environment.

fidencio:
Folks, right now we have **all** the PRs being blocked on this issue.
We have 2 options here, from my personal point of view.
1) Merge it, unblock other PRs, and figure out what's going on.
2) Mark the static-checks that are failing as non-required, and have this properly fixed on this PR.

I'm fine with both options, but I'd like to unblock `main` ASAP, so I need the opinion from the reviewers on what  to do.

gkurz:
> Folks, right now we have **all** the PRs being blocked on this issue. We have 2 options here, from my personal point of view.
> 
> 1. Merge it, unblock other PRs, and figure out what's going on.
> 2. Mark the static-checks that are failing as non-required, and have this properly fixed on this PR.
> 
> I'm fine with both options, but I'd like to unblock `main` ASAP, so I need the opinion from the reviewers on what to do.

Sorry for the delay. @justxuewei's clarifications are enough for now. Let's go with option 1 for the sake of unblocking everyone,  but I really want this to be reworked again afterwards.

gkurz (approved):
Thanks for unblocking the CI @justxuewei !

amshinde (approved):

Issue #8399, opened 2023-11-07
CC - agent: support crio annotations in pullimage code


littlejawa:
I'm working on supporting the pullimage in guest feature from cri-o. When I tried sending the relevant data through the "KataVirtualVolume" structure, the agent crashed.

I found that the agent is looking for annotations from containerd to change its behavior between sandbox and container. But the annotation name is different with cri-o, and the lookup caused the panic.

As part of this issue, we should not only avoid the panic, but also make sure cri-o's annotation is understood by the agent.

PR #8400, opened 2023-11-07
CC | agent: fix annotation name to allow cri-o support of CoCo pullimage in guest

littlejawa:
The pullimage handler behaves differently for the sandbox and container, and it uses the annotations to differentiate between the two. The annotation name used by crio for container-type being different than the containerd one, we need to check both.

Fixes: #8399

### DIFF
--- a/src/agent/src/image_rpc.rs
+++ b/src/agent/src/image_rpc.rs
@@ -207,7 +207,16 @@ impl ImageService {
     ) -> Result<String> {
         info!(sl(), "image metadata: {:?}", image_metadata);
         Self::set_proxy_env_vars();
-        if image_metadata["io.kubernetes.cri.container-type"] == "sandbox" {
+        let is_sandbox = if let Some(value) = image_metadata.get("io.kubernetes.cri.container-type")
+        {
+            value == "sandbox"
+        } else if let Some(value) = image_metadata.get("io.kubernetes.cri-o.ContainerType") {
+            value == "sandbox"
+        } else {
+            false
+        };
+
+        if is_sandbox {
             let mount_path = Self::unpack_pause_image(cid, "pause")?;
             self.add_image(String::from(image), String::from(cid)).await;
             return Ok(mount_path);


pmores (approved):
lgtm, thanks @littlejawa !

src/agent/src/image_rpc.rs
```diff
@@ -207,7 +207,15 @@ impl ImageService {
     ) -> Result<String> {
         info!(sl(), "image metadata: {:?}", image_metadata);
         Self::set_proxy_env_vars();
-        if image_metadata["io.kubernetes.cri.container-type"] == "sandbox" {
+        let is_sandbox = if let Some(value) = image_metadata.get("io.kubernetes.cri.container-type") {
+            value == "sandbox"
+        } else if let Some(value) = image_metadata.get("io.kubernetes.cri-o.ContainerType") {
```
fidencio, line 212: Personally, I'd prfer those two annotations being a const / enum, and looping into that to check whether value is sandbox.

OTOH, this branch is almost dead, so I wouldn't block the PR on this,

fidencio:
@littlejawa, please, check make sure the checks are happy.
@portersrc, @stevenhorsman, we need this one in before we officially cut v0.8.0

gkurz (approved):
@fidencio's suggestion can be addressed by a follow-up enhancement when this eventually hits the main branch.

Thanks @littlejawa !

stevenhorsman (approved):
LGTM. Thanks!

src/agent/src/image_rpc.rs
```diff
@@ -207,7 +207,15 @@ impl ImageService {
     ) -> Result<String> {
         info!(sl(), "image metadata: {:?}", image_metadata);
         Self::set_proxy_env_vars();
-        if image_metadata["io.kubernetes.cri.container-type"] == "sandbox" {
+        let is_sandbox = if let Some(value) = image_metadata.get("io.kubernetes.cri.container-type") {
```
stevenhorsman, line 210: It looks like the static check wants 
```suggestion
        let is_sandbox = if let Some(value) = image_metadata.get("io.kubernetes.cri.container-type")
        {
```
instead. I'm not sure why though as I think the current version is better and more consistent? I think my IDE would auto format that suggest to match what Julien added anyway, but maybe there is a refactor that would help it pass like:
```
let container_type = image_metadata.get("io.kubernetes.cri.container-type");
    let is_sandbox = if let Some(value) = container_type {
        value == "sandbox"
    } else if let Some(value) = image_metadata.get("io.kubernetes.cri-o.ContainerType") {
        value == "sandbox"
    } else {
        false
    };
```
?

littlejawa: I see there are other "if let" statements here and there in the code where the accolade is on a newline, so I guess that's actually consistent - even if it seems to be an exception rather than the norm. Maybe it's a matter of line length too? I don't know.
Anyway, I went with the suggestion from "make check".

Issue #8403, opened 2023-11-08
kata-deploy: Allow users to set allowed hypervisor annotations


fidencio:
Currently the only way one can specify allowed hypervisor annotations is
during build time, which is a big issue for users grabbing kata-deploy
as we provide.

PR #8404, opened 2023-11-08
kata-deploy: Allow users to set hypervisor annotations

fidencio:
Currently the only way one can specify allowed hypervisor annotations is
during build time, which is a big issue for users grabbing kata-deploy
as we provide.

Fixes: #8403

---

Although a formal test for this is not provided, we're testing this as part of all the deployments / tests using CBL mariner, as we're set the extra "initrd" and "kernel" to the allowed list of hypervisor annotations.

@sprt, I'd appreciate a lot a review and tests from your side to make sure this is working as expected with CBL mariner.

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -130,7 +130,6 @@ function install_artifacts() {
 	if [ "${HOST_OS:-}" == "cbl-mariner" ]; then
 		config_path="/opt/kata/share/defaults/kata-containers/configuration-clh.toml"
 		clh_path="/opt/kata/bin/cloud-hypervisor-glibc"
-		sed -i -E 's|(enable_annotations) = .+|\1 = ["enable_iommu", "initrd", "kernel"]|' "${config_path}"
 		sed -i -E "s|(valid_hypervisor_paths) = .+|\1 = [\"${clh_path}\"]|" "${config_path}"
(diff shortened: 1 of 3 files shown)


tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -105,16 +111,20 @@ function install_artifacts() {
...
 		for shim in "${shims[@]}"; do
 			sed -i -e 's/^#\(enable_debug\).*=.*$/\1 = true/g' "${config_path}/configuration-${shim}.toml"
 			sed -i -e 's/^#\(debug_console_enabled\).*=.*$/\1 = true/g' "${config_path}/configuration-${shim}.toml"
 			sed -i -e 's/^kernel_params = "\(.*\)"/kernel_params = "\1 agent.log=debug initcall_debug"/g' "${config_path}/configuration-${shim}.toml"
 		done
 	fi
 
+	if [ -n "${allowed_hypervisor_annotations}" ]; then
+		sed -i -e "s/^enable_annotations = \[\(.*\)\]/enable_annotations = [\1, $allowed_hypervisor_annotations]/" "${config_path}/configuration-${shim}.toml"
+	fi
```
stevenhorsman, line 126: Maybe a dumb question, but will `shim` be set here, or should this be running inside the `for shim in "${shims[@]}"; do` loop?

stevenhorsman (approved):
LGTM. I've run some of the env to `allowed_hypervisor_annotations` logic locally as I know from experience that it's easy to get wrong and it seems to work as expected to me. Thanks!

Issue #8410, opened 2023-11-09
gha: k3s CIs are failing to download kubectl


fidencio:
```sh
[INFO]  Finding release for channel stable
  [INFO]  Using v1.27.7+k3s2 as release
  [INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.27.7+k3s2/sha256sum-amd64.txt
  [INFO]  Downloading binary https://github.com/k3s-io/k3s/releases/download/v1.27.7+k3s2/k3s
  [INFO]  Verifying binary download
  [INFO]  Installing k3s to /usr/local/bin/k3s
  [INFO]  Skipping installation of SELinux RPM
  [INFO]  Creating /usr/local/bin/kubectl symlink to k3s
  [INFO]  Creating /usr/local/bin/crictl symlink to k3s
  [INFO]  Creating /usr/local/bin/ctr symlink to k3s
  [INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
  [INFO]  Creating uninstall script /usr/local/bin/k3s-uninstall.sh
  [INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
  [INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
  sh: 970: restorecon: not found
  sh: 971: restorecon: not found
  [INFO]  systemd: Enabling k3s unit
  Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service → /etc/systemd/system/k3s.service.
  [INFO]  systemd: Starting k3s
  curl: ([22](https://github.com/kata-containers/kata-containers/actions/runs/6807799372/job/18511651149?pr=8404#step:4:23)) The requested URL returned error: 404
```

This is happening as the regex used to grab the kubectl version from k3s is not working as expected, due to a bump on k3s version.

PR #8411, opened 2023-11-09
gha: Fix regex used to get kubectl version from the k3s version

fidencio:
It seems that with the new k3s release, they've bumped their kubectl version from x.y.z+k3s1 to x.y.z+k3s2.

Let's ensure our regexp is more generic and future proof for such changes.

Fixes: #8410

### DIFF
--- a/tests/gha-run-k8s-common.sh
+++ b/tests/gha-run-k8s-common.sh
@@ -161,7 +161,7 @@ function deploy_k3s() {
 	if [ "${ARCH}" = "x86_64" ]; then
 		ARCH=amd64
 	fi
-	kubectl_version=$(/usr/local/bin/k3s kubectl version --short 2>/dev/null | grep "Client Version" | sed -e 's/Client Version: //' -e 's/\+k3s[0-9]\+//')
+	kubectl_version=$(/usr/local/bin/k3s kubectl version --short 2>/dev/null | grep "Client Version" | sed -e 's/Client Version: //' -e 's/+k3s[0-9]\+//')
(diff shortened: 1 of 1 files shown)


tests/gha-run-k8s-common.sh
```diff
@@ -161,7 +161,7 @@ function deploy_k3s() {
 	if [ "${ARCH}" = "x86_64" ]; then
 		ARCH=amd64
 	fi
-	kubectl_version=$(/usr/local/bin/k3s kubectl version --short 2>/dev/null | grep "Client Version" | sed -e 's/Client Version: //' -e 's/\+k3s1//')
+	kubectl_version=$(/usr/local/bin/k3s kubectl version --short 2>/dev/null | grep "Client Version" | sed -e 's/Client Version: //' -e 's/\+k3s[0-9]\+//')
```
gkurz, line 164: It seems that we want to match several trailing digits, not a plus sign:

```suggestion
	kubectl_version=$(/usr/local/bin/k3s kubectl version --short 2>/dev/null | grep "Client Version" | sed -e 's/Client Version: //' -e 's/\+k3s[0-9]+//')
```

gkurz: Oops I was wrong, we need to escape the final plus sign. The one before k3s doesn't need to be escaped though.

gkurz (approved):
LGTM. Thanks @fidencio !

tests/gha-run-k8s-common.sh
fidencio: As discussed on Slack, we actually have to escape the latter +, but the former one is unnecessarily escaped right now.
I've added a commit to address that.

studychao (approved):
Gooood Catch

Issue #8413, opened 2023-11-09
PR #8284 breaks PR #8185 unnecessarily


pmores:

pmores:
A straight-forward fix was proposed in #8414, ultimately fixed by #8516 and #8672.

PR #8414, opened 2023-11-09
Fix PR #8284

pmores:
PR #8284 removed `DeviceType::Vsock` which broke PR #8185.  The removal (and thus breakage) doesn't seem necessary to achieve PR #8284's goals and this PR proposes a solution.  It is based on observation that while PR #8284 imposes the requirement of cloneability on all `DeviceType` variants, the requirement is arguably unnecessarily strong since fallible cloneability (try_clone() functionality) should be enough.

NB I couldn't test this change properly.  Since this is an emergency fix of another PR I take the liberty of relying on reviewers' help to verify the change.

NB This PR cannot and does not address potential technical design flaws in PR #8284. Most importantly, it's not clear whether adding a `DeviceType` return value to `add_device()` is the best way of achieving PR #8284's goals.  The resulting signature incorrectly suggests that `add_device()` could take a `DeviceType` and return a completely unrelated one.  Another solution, maybe passing a mutable reference to `add_device()`, could have communicated the `add_device()` contract better.

Fixes: #8413

@Apokleos @bergwolf @amshinde

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/mod.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/mod.rs
@@ -10,9 +10,9 @@ use crate::device::driver::vhost_user_blk::VhostUserBlkDevice;
 use crate::{
     BlockConfig, BlockDevice, HybridVsockConfig, HybridVsockDevice, Hypervisor as hypervisor,
     NetworkConfig, NetworkDevice, ShareFsDevice, ShareFsDeviceConfig, ShareFsMountConfig,
-    ShareFsMountDevice, VfioConfig, VfioDevice, VhostUserConfig, VsockConfig,
+    ShareFsMountDevice, VfioConfig, VfioDevice, VhostUserConfig, VsockConfig, VsockDevice,
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 
 pub mod device_manager;
@@ -31,7 +31,7 @@ pub enum DeviceConfig {
     HybridVsockCfg(HybridVsockConfig),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum DeviceType {
     Block(BlockDevice),
     VhostUserBlk(VhostUserBlkDevice),
@@ -40,6 +40,7 @@ pub enum DeviceType {
     ShareFs(ShareFsDevice),
     ShareFsMount(ShareFsMountDevice),
     HybridVsock(HybridVsockDevice),
+    Vsock(VsockDevice),
 }
 
 impl DeviceType {
@@ -56,6 +57,7 @@ impl DeviceType {
                 Ok(DeviceType::ShareFsMount(share_fs_mount.clone()))
             }
             DeviceType::HybridVsock(hvsock) => Ok(DeviceType::HybridVsock(hvsock.clone())),
+            DeviceType::Vsock(_) => Err(anyhow!("DeviceType::Vsock cannot be cloned")),
         }
     }
 }
--- a/src/runtime-rs/crates/hypervisor/src/device/mod.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/mod.rs
@@ -42,6 +42,24 @@ pub enum DeviceType {
     HybridVsock(HybridVsockDevice),
 }
 
+impl DeviceType {
+    pub fn try_clone(&self) -> Result<DeviceType> {
+        match self {
+            DeviceType::Block(block) => Ok(DeviceType::Block(block.clone())),
+            DeviceType::VhostUserBlk(vhost_user_blk) => {
+                Ok(DeviceType::VhostUserBlk(vhost_user_blk.clone()))
+            }
+            DeviceType::Vfio(vfio) => Ok(DeviceType::Vfio(vfio.clone())),
+            DeviceType::Network(network) => Ok(DeviceType::Network(network.clone())),
+            DeviceType::ShareFs(share_fs) => Ok(DeviceType::ShareFs(share_fs.clone())),
+            DeviceType::ShareFsMount(share_fs_mount) => {
+                Ok(DeviceType::ShareFsMount(share_fs_mount.clone()))
+            }
+            DeviceType::HybridVsock(hvsock) => Ok(DeviceType::HybridVsock(hvsock.clone())),
+        }
+    }
+}
+
 impl fmt::Display for DeviceType {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{:?}", self)
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs
@@ -57,8 +57,8 @@ impl CloudHypervisorInner {
             // - Network details need to be saved for later application.
             //
             match device {
-                DeviceType::ShareFs(_) => self.pending_devices.insert(0, device.clone()),
-                DeviceType::Network(_) => self.pending_devices.insert(0, device.clone()),
+                DeviceType::ShareFs(_) => self.pending_devices.insert(0, device.try_clone()?),
+                DeviceType::Network(_) => self.pending_devices.insert(0, device.try_clone()?),
                 _ => {
                     debug!(
                         sl!(),
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
@@ -34,23 +34,27 @@ pub(crate) fn drive_index_to_id(index: u64) -> String {
 }
 
 impl DragonballInner {
-    pub(crate) async fn add_device(&mut self, device: DeviceType) -> Result<()> {
+    pub(crate) async fn add_device(&mut self, device: DeviceType) -> Result<DeviceType> {
         if self.state == VmmState::NotReady {
             info!(sl!(), "VMM not ready, queueing device {}", device);
 
             // add the pending device by reverse order, thus the
             // start_vm would pop the devices in an right order
             // to add the devices.
-            self.pending_devices.insert(0, device);
-            return Ok(());
+            self.pending_devices.insert(0, device.try_clone()?);
+            return Ok(device);
         }
 
         info!(sl!(), "dragonball add device {:?}", &device);
-        match device {
+        match &device {
             DeviceType::Network(network) => self
                 .add_net_device(&network.config)
+                .map(|_ok| device)
                 .context("add net device"),
-            DeviceType::Vfio(hostdev) => self.add_vfio_device(&hostdev).context("add vfio device"),
+            DeviceType::Vfio(hostdev) => self
+                .add_vfio_device(hostdev)
+                .map(|_ok| device)
+                .context("add vfio device"),
             DeviceType::Block(block) => self
                 .add_block_device(
                     block.config.path_on_host.as_str(),
@@ -58,6 +62,7 @@ impl DragonballInner {
                     block.config.is_readonly,
                     block.config.no_drop,
                 )
+                .map(|_ok| device)
                 .context("add block device"),
             DeviceType::VhostUserBlk(block) => self
                 .add_block_device(
@@ -66,13 +71,19 @@ impl DragonballInner {
                     block.is_readonly,
                     block.no_drop,
                 )
+                .map(|_ok| device)
                 .context("add vhost user based block device"),
-            DeviceType::HybridVsock(hvsock) => self.add_hvsock(&hvsock.config).context("add vsock"),
+            DeviceType::HybridVsock(hvsock) => self
+                .add_hvsock(&hvsock.config)
+                .map(|_ok| device)
+                .context("add vsock"),
             DeviceType::ShareFs(sharefs) => self
                 .add_share_fs_device(&sharefs.config)
+                .map(|_ok| device)
                 .context("add share fs device"),
             DeviceType::ShareFsMount(sharefs_mount) => self
                 .add_share_fs_mount(&sharefs_mount.config)
+                .map(|_ok| device)
                 .context("add share fs mount"),
         }
     }
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
@@ -85,6 +85,9 @@ impl DragonballInner {
                 .add_share_fs_mount(&sharefs_mount.config)
                 .map(|_ok| device)
                 .context("add share fs mount"),
+            DeviceType::Vsock(_) => {
+                todo!()
+            }
         }
     }
 
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
@@ -94,10 +94,7 @@ impl Hypervisor for Dragonball {
 
     async fn add_device(&self, device: DeviceType) -> Result<DeviceType> {
         let mut inner = self.inner.write().await;
-        match inner.add_device(device.clone()).await {
-            Ok(_) => Ok(device),
-            Err(err) => Err(err),
-        }
+        inner.add_device(device).await
     }
 
     async fn remove_device(&self, device: DeviceType) -> Result<()> {


gkurz (approved):
This is definitely a smoother approach. We're obvisouly not at the stage where all DeviceType are expected to be passed to `add_device()`.

LGTM. Thanks @pmores !

src/runtime-rs/crates/hypervisor/src/device/mod.rs
```diff
@@ -56,6 +57,7 @@ impl DeviceType {
                 Ok(DeviceType::ShareFsMount(share_fs_mount.clone()))
             }
             DeviceType::HybridVsock(hvsock) => Ok(DeviceType::HybridVsock(hvsock.clone())),
+            DeviceType::Vsock(_) => Err(anyhow!("DeviceType::Vsock cannot be cloned")),
```
Apokleos, line 60: IMO, "DeviceType::Vsock cannot be cloned" maybe not accurate, what's about "unsupported Vsock device type"

pmores: Good point, thanks. I haven't really looked into what it would take to make `Vsock` cloneable but it might well be possible to add cloning support to `Vsock` after all even if it perhaps wouldn't be trivial.

src/runtime-rs/crates/hypervisor/src/device/mod.rs
```diff
@@ -40,6 +40,26 @@ pub enum DeviceType {
...
+                Ok(DeviceType::VhostUserBlk(vhost_user_blk.clone()))
+            }
+            DeviceType::Vfio(vfio) => Ok(DeviceType::Vfio(vfio.clone())),
+            DeviceType::Network(network) => Ok(DeviceType::Network(network.clone())),
+            DeviceType::ShareFs(share_fs) => Ok(DeviceType::ShareFs(share_fs.clone())),
+            DeviceType::ShareFsMount(share_fs_mount) => {
+                Ok(DeviceType::ShareFsMount(share_fs_mount.clone()))
+            }
+            DeviceType::HybridVsock(hvsock) => Ok(DeviceType::HybridVsock(hvsock.clone())),
+            DeviceType::Vsock(_) => Err(anyhow!("DeviceType::Vsock cannot be cloned")),
```
studychao, line 60: tokio::File has method `try_clone` (https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.try_clone), could you also implement try_clone for `DeviceType::Vsock`, so that we could use the `try_clone` method for `Vsock` device in the device manager process?

studychao (commented):
Thanks @pmores , one comment.

studychao:
Hi @pmores , after reading through your PR and also the vhost vsock implementation in go runtime, I have clearer view of this problem.

First, in go runtime, runtime will create vhostfd and then *pass that fd* to qemu process when launching the qemu child process, which means runtime doesn't need to manage the whole life cycle of the fd.

Currently, in rust runtime, the reason why we couldn't use `clone` on `VsockConfig` is that because we store the vhost fd (tokio::fs) in `VsockConfig` and implicitly manage the life cycle of the vhost vsock fd in the rust runtime side which is actually not rust runtime's responsibility.

So, I think the better way is to pass that fd to QEMU process like what is done in go runtime ( the actual code will be pasted)
image
and remove the vsock fd file in `VsockConfig` so that we could implement `Clone` again for `DeviceType`.

src/runtime-rs/crates/hypervisor/src/device/mod.rs
studychao: this is a outdated comment.
Instead of thinking about how to clone that fd, I think we could remove that in `VsockConfig` and directly pass the fd to QEMU process and let QEMU manage the life cycle of the fd.

pmores:
Hi @studychao , thanks for your analysis, I appreciate it! I'll need a little time to digest your suggestion and think through its consequences but a quick question that comes to mind right now - is [this comment](https://github.com/kata-containers/kata-containers/blob/b3318e59ebc8ba74c3745dc50d6d5ec281270e17/src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs#L87) mistaken then?

studychao:
> Hi @studychao , thanks for your analysis, I appreciate it! I'll need a little time to digest your suggestion and think through its consequences but a quick question that comes to mind right now - is [this comment](https://github.com/kata-containers/kata-containers/blob/b3318e59ebc8ba74c3745dc50d6d5ec281270e17/src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs#L87) mistaken then?

Hi @pmores , this comment is correct that the fd it takes control the cid and release the fd will let other vm take the same cid. 

But the problem is that who should take the responsibility to manage the life cycle of this cid, if we store this fd in vsockconfig struct, we let kata runtime to manage the life cycle which is not the behavior of the go runtime. In go runtime, runtime creates the fd and then pass it to the qemu process and let qemu manage the life cycle of the vsock fd.

So the comment of the usage of this file is correct, but store it in the runtime side is not correct AFAIK.

Apokleos:
> Hi @pmores , after reading through your PR and also the vhost vsock implementation in go runtime, I have clearer view of this problem.
> 
> First, in go runtime, runtime will create vhostfd and then _pass that fd_ to qemu process when launching the qemu child process, which means runtime doesn't need to manage the whole life cycle of the fd.
> 
> Currently, in rust runtime, the reason why we couldn't use `clone` on `VsockConfig` is that because we store the vhost fd (tokio::fs) in `VsockConfig` and implicitly manage the life cycle of the vhost vsock fd in the rust runtime side which is actually not rust runtime's responsibility.
> 
> So, I think the better way is to pass that fd to QEMU process like what is done in go runtime ( the related code will be pasted) image and remove the vsock fd file in `VsockConfig` so that we could implement `Clone` again for `DeviceType`.
>
@studychao  If the `vhost_fd` can be moved out of the `VsockConfig`, it could be a win-win solution. The todo for`vsock device` moved into device manager will also benefit from it.

pmores:
@studychao So, I understand your point and it makes sense to me. How would you like to go about it though?  If we don't store the fd, do we turn `VsockDevice::new()` into an ordinary member function that returns the `File`?

pmores:
@studychao Also, regarding what the golang runtime does: after having a closer look frankly I don't quite see what lesson runtime-rs can learn there as it seems to do the same already.

The go runtime opens the fd and stores it in a struct just like runtime-rs does with `VSockDevice` (it's just called `VSOCKDevice`).  The device struct is then stored in govmm's `Config::Devices` until it's used in `StartVM()` which is what at least the runtime-rs QEMU driver does as well.

The only difference is that in go runtime `StartVM()` closes all raw fds from under the `Device` structs that use them, including `VSOCKDevice`, silently invalidating them.  I'm sure we don't want to emulate _that_ step, and luckily it would probably be hard in Rust.

studychao:
hi @pmores , yes I agree with you, after reading carefully more in go runtime, I find out that go runtime also store that fd for a little while before `StartVM()` and then go runtime close the fd after it is passed to QEMU process

But since the fd need to be stored, this behavior will make `VsockConfig` hard to impl `Clone`...

Let me read more code and see if there are any solutions

studychao:
then, could you use tokio::File's try_clone method (https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.try_clone), and also implement try_clone for DeviceType::Vsock, so that we could use the try_clone method to actually clone Vsock device in the device manager process?

Also, go runtime will close the fd after `StartVM` may also be considered to implement in Rust runtime in order to avoid fd leak.

gkurz:
> hi @pmores , yes I agree with you, after reading carefully more in go runtime, I find out that go runtime also store that fd for a little while before `StartVM()` and then go runtime close the fd after it is passed to QEMU process
> 
> But since the fd need to be stored in somewhere , this behavior will make `VsockConfig` hard to impl `Clone`...
> 
> Let me read more code and see if there are any solutions

First thing that comes to mind is that the File and the CID have different life cycles. They have an overlap because the File is obviously needed to create the CID, but that's all. Then we just give the fd to QEMU and the File can be closed,
while the CID continues to exist until QEMU terminates. Something like below :

```
                     QEMU          QEMU
                     start     termination
                      ^             |
                      |             |
<------- File -------->             |
                  |                 |
      vhost_vsock_set_guest_cid()   |
                  |                 |
                  v                 v
                  <------ CID ------>
```

It thus looks they shouldn't live in the same place.

pmores:
Thanks for the great discussion guys!

I think we have several options to consider.  First, there's the `try_clone()` way.  I'd like to understand under which circumstances this function can fail.  This doesn't seem to be directly documented and it's not obvious to me from looking at the function's source code.  Does anybody have an idea?

We could also remove the `File` member from `VsockConfig` and have `VsockDevice::new()` do basically just default construction (set `VsockDevice::id` to the passed-in value and `VsockDevice::config::guest_cid` to an invalid value, perhaps `VMADDR_CID_ANY`).  Then add another function, perhaps `VsockDevice::init()` (I'm sure there would be a better name, I just can't come up with any at the moment ;-)) that would basically do what `VsockConfig::new()` does now *but* it wouldn't store the `File` and return it instead. Hypervisor driver could call this when it's ready and by virtue of the `File` returned to it, it would take its ownership, meaning that the return value would be dropped by default and the fd closed automatically.

There might be other ways but since `File` doesn't seem to support the concept of an empty or uninitialised instance (it has no `new()`, just `open()`, and no `close()`, it needs to be dropped instead) we'd probably have to store it in an `Option` - if we insisted to have it in `VsockConfig` in some way.  Then we could have an accessor that moved from the Option and passed the ownership to the caller.  However since we don't seem to insist on storing the fd in `VsockConfig` I'd consider the "`VsockDevice::init()`" option above more attractive.

What do you think?

gkurz:
> Thanks for the great discussion guys!
> 
> I think we have several options to consider. First, there's the `try_clone()` way. I'd like to understand under which circumstances this function can fail. This doesn't seem to be directly documented and it's not obvious to me from looking at the function's source code. Does anybody have an idea?
> 
> We could also remove the `File` member from `VsockConfig` and have `VsockDevice::new()` do basically just default construction (set `VsockDevice::id` to the passed-in value and `VsockDevice::config::guest_cid` to an invalid value, perhaps `VMADDR_CID_ANY`). Then add another function, perhaps `VsockDevice::init()` (I'm sure there would be a better name, I just can't come up with any at the moment ;-)) that would basically do what `VsockConfig::new()` does now _but_ it wouldn't store the `File` and return it instead. Hypervisor driver could call this when it's ready and by virtue of the `File` returned to it, it would take its ownership, meaning that the return value would be dropped by default and the fd closed automatically.
> 
> There might be other ways but since `File` doesn't seem to support the concept of an empty or uninitialised instance (it has no `new()`, just `open()`, and no `close()`, it needs to be dropped instead) we'd probably have to store it in an `Option` - if we insisted to have it in `VsockConfig` in some way. Then we could have an accessor that moved from the Option and passed the ownership to the caller. However since we don't seem to insist on storing the fd in `VsockConfig` I'd consider the "`VsockDevice::init()`" option above more attractive.
> 
> What do you think?

I was thinking about something just like your `VsockDevice::init()` option :-)

What about `VsockDevice::create_vhost_in_kernel()` since this is basically what this function is doing ?

studychao:
> Thanks for the great discussion guys!
> 
> I think we have several options to consider. First, there's the `try_clone()` way. I'd like to understand under which circumstances this function can fail. This doesn't seem to be directly documented and it's not obvious to me from looking at the function's source code. Does anybody have an idea?
> 
> We could also remove the `File` member from `VsockConfig` and have `VsockDevice::new()` do basically just default construction (set `VsockDevice::id` to the passed-in value and `VsockDevice::config::guest_cid` to an invalid value, perhaps `VMADDR_CID_ANY`). Then add another function, perhaps `VsockDevice::init()` (I'm sure there would be a better name, I just can't come up with any at the moment ;-)) that would basically do what `VsockConfig::new()` does now _but_ it wouldn't store the `File` and return it instead. Hypervisor driver could call this when it's ready and by virtue of the `File` returned to it, it would take its ownership, meaning that the return value would be dropped by default and the fd closed automatically.
> 
> There might be other ways but since `File` doesn't seem to support the concept of an empty or uninitialised instance (it has no `new()`, just `open()`, and no `close()`, it needs to be dropped instead) we'd probably have to store it in an `Option` - if we insisted to have it in `VsockConfig` in some way. Then we could have an accessor that moved from the Option and passed the ownership to the caller. However since we don't seem to insist on storing the fd in `VsockConfig` I'd consider the "`VsockDevice::init()`" option above more attractive.
> 
> What do you think?

+1 for something like `VsockDevice::init()` from me.

studychao:
@pmores Hi, pmores, could you update this PR based on our disscussion? 
I think maybe we can separate this PR into 2 PRs:
1. reintroduce vsock device type, delete vsock vhost fd in `VsockConfig`  and add `Clone` impl
2. add something like `VsockDevice::init()` method to actually enable Vsock in QEMU

Apokleos:
> Thanks for the great discussion guys!
> 
> I think we have several options to consider. First, there's the `try_clone()` way. I'd like to understand under which circumstances this function can fail. This doesn't seem to be directly documented and it's not obvious to me from looking at the function's source code. Does anybody have an idea?
> 
> We could also remove the `File` member from `VsockConfig` and have `VsockDevice::new()` do basically just default construction (set `VsockDevice::id` to the passed-in value and `VsockDevice::config::guest_cid` to an invalid value, perhaps `VMADDR_CID_ANY`). Then add another function, perhaps `VsockDevice::init()` (I'm sure there would be a better name, I just can't come up with any at the moment ;-)) that would basically do what `VsockConfig::new()` does now _but_ it wouldn't store the `File` and return it instead. Hypervisor driver could call this when it's ready and by virtue of the `File` returned to it, it would take its ownership, meaning that the return value would be dropped by default and the fd closed automatically.
> 
> There might be other ways but since `File` doesn't seem to support the concept of an empty or uninitialised instance (it has no `new()`, just `open()`, and no `close()`, it needs to be dropped instead) we'd probably have to store it in an `Option` - if we insisted to have it in `VsockConfig` in some way. Then we could have an accessor that moved from the Option and passed the ownership to the caller. However since we don't seem to insist on storing the fd in `VsockConfig` I'd consider the "`VsockDevice::init()`" option above more attractive.
> 
> What do you think?
>
Hi @pmores I agree with you about the idea moving the vhost_fd out of `VsockConfig` and adding one method VsockDevice::init().
And currently [the work on moving vsock device into device manager ](https://github.com/kata-containers/kata-containers/pull/8516)will  also benefit from it.   

Just FYI, my PR will stay in WIP until yours gets merged. I'll reimplement mine based on your implementation related to vsock device then.

pmores:
Thanks guys for your quick reactions @Apokleos @studychao @gkurz ! So it looks like we're converging to a solution, good! :-) 

@studychao I can do two PRs just fine but I want to make sure that we're OK with `VsockDevice` not fully functional after the first one and before the second one?  To explain, if I just remove `VsockConfig::vsock_fd` but don't introduce the "`init()`" function right away, the implementation will have to stay in `new()` which cannot return the fd and has to drop it, now that it has nowhere to store it.  Thus `VsockDevice` would have no way to make the fd available to a hypervisor, crippling its usefulness.  Is that OK, or am I overlooking something?

Also I'd like to pay a bit more attention to the proposed function name.  So far, the options I can see are
- `init()` which is OK I guess but doesn't seem great to me frankly
- `create_vhost_in_kernel()` which describes what the function does which I feel is both its strength but perhaps also a weakness since it might be focusing a bit too much on "how" instead of "what"
- `make()` comes to my mind as maybe a better alternative to `init()` since it reflects the fact better that the function is in fact a sort of a factory function which creates a new vsock setup each time its called (not that it would make much sense to call it multiple times I guess...)

Any other ideas? Which name do we prefer in the end?

studychao:
> Thanks guys for your quick reactions @Apokleos @studychao @gkurz ! So it looks like we're converging to a solution, good! :-)
> 
> @studychao I can do two PRs just fine but I want to make sure that we're OK with `VsockDevice` not fully functional after the first one and before the second one? To explain, if I just remove `VsockConfig::vsock_fd` but don't introduce the "`init()`" function right away, the implementation will have to stay in `new()` which cannot return the fd and has to drop it, now that it has nowhere to store it. Thus `VsockDevice` would have no way to make the fd available to a hypervisor, crippling its usefulness. Is that OK, or am I overlooking something?
> 

I think that is ok since QEMU is the only hypervisor that uses vhost vsock right now.
If you (stand as the QEMU perspective) think it is ok to be in a middle stage that `VsockDevice` temporarily disabled before your next PR, then I think it's not a big problem

studychao:
> Thanks guys for your quick reactions @Apokleos @studychao @gkurz ! So it looks like we're converging to a solution, good! :-)

Yeah, really enjoy the process :)

pmores:
@studychao Now I've took a look at @Apokleos 's PR linked to above and if I understand correctly it's actually quite exactly the first of the PRs that you propose.  So we might just as well head over there and give it a review. :-)

BbolroC:
I am closing this because #8516 and #8671 were merged, which should resolve the issue by #8284. Thanks.

PR #8415, opened 2023-11-09
cc: gha: kernel-snp-experimental is not a valid tag

fidencio:
We fixed this as part of the after-push PR, but we forgot to have this one fixed.


stevenhorsman (approved):
LGTM

fidencio:
I'm merging this without running CI, as it doesn't touch absolutely any code but the release one.

Issue #8418, opened 2023-11-10
runtime-rs: ch: Fix TDX


jodh-intel:
PR #8311 appears to have inadvertently broken TDX handling.

PR #8311, opened 2023-10-25
Improve logging control and reduce logger re-creation

TimePrinciple:
By modifying RuntimeLevelFilter drain to improve logging control, enabling isolation of change effect of the loggers between components.

PR #8419, opened 2023-11-10
runtime-rs: ch: Fix TDX

jodh-intel:
PR #8311 inadvertently broke the runtime-rs / Cloud Hypervisor TDX handling. It also introduced unrecoverable failure scenarios. Hence, replace slow, fallible regex matching in logging fast path with single pass non-failing multi-string log level matching.

Also, added a unit test for `parse_ch_log_level()`.

Fixes: #8418.


fidencio (approved):
lgtm, thanks @jodh-intel!

cmaf (approved):
thanks @jodh-intel

Issue #8416, opened 2023-11-10
runtime-rs occurred a panic when the 'enable-debug' option was activated with Cloud Hypervisor


brianwang12:
To help us understand the problem more quickly, please do the following:
$ sudo ctr run  --runtime io.containerd.run.kata.v2 -t --rm docker.io/library/busybox:latest hello7 sh 

# Expected result
It could print the guest kernel from Cloud-hypervisor.
# Actual result
[...]

PR #8422, opened 2023-11-13
runtime-rs on arm64: Fixes unable to Boot Container Image using Cloud…

brianwang12:
This patch addresses the issue of being unable to boot container images using Cloud-Hypervisor on arm64
Fitxes: https://github.com/kata-containers/kata-containers/issues/8381

PR #8427, opened 2023-11-13
runtime-rs: Fix panic issue with 'enable-debug' option on Cloud Hypervisor

brianwang12:
Fixes a problem where parse_ch_log_level() fails to work correctly when the "enable-debug" option is turned on in the configuration.toml.

Root Cause of Panic:
Parsing certain logs,
such as "[0.000000] Booting Linux on physical CPU 0x0000000000 [0x411fd0c0]"
        "cloud-hypervisor: 1.302s: <vcpu0> INFO:.."
can triggers the following panic:
{"msg":"A panic occurred at crates/hypervisor/src/ch/inner_hypervisor.rs:768: There should be a match for level\r\n"...

Fixes: #8416

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
@@ -761,13 +761,16 @@ async fn cloud_hypervisor_log_output(mut child: Child, mut shutdown: Receiver<bo
 }
 
 pub fn parse_ch_log_level(line: &str) -> &str {
-    let re = Regex::new(r"cloud-hypervisor: [0-9]*[.][0-9]+ms: <\w+> (?<level>\w+)").unwrap();
-    let level = re
-        .captures(line)
-        .expect("There should be a match for level")
-        .name("level")
-        .expect("Level should be found in record")
-        .as_str();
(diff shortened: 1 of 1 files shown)


TimePrinciple:
image
This snipet is from the `cloud-hypervisor`'s source code, which documents that there should be a `Level` field as long as this is a log record.

As to the panic, I suggest finding out which exactly the record is causing this, and modify the logic of mapping between `cloud-hypervisor level` and our `slog-level`. There might be modifications on `regex` for it to correctly locate the `level` field in record.

jodh-intel:
Hi @brianwang12 - thanks for raising but I think the problem you are seeing was fixed by #8419. Please could you verify that the latest runtime-rs code with CH doesn't panic?

brianwang12:
> Hi @brianwang12 - thanks for raising but I think the problem you are seeing was fixed by #8419. Please could you verify that the latest runtime-rs code with CH doesn't panic?



> Hi @brianwang12 - thanks for raising but I think the problem you are seeing was fixed by #8419. Please could you verify that the latest runtime-rs code with CH doesn't panic?

@jodh-intel  I got it , I have verified the latest runtime-rs code, and it could work well based on https://github.com/kata-containers/kata-containers/pull/8422 in my arm machine.

jodh-intel:
Thanks for confirming @brianwang12. Shall we close this PR then?

PR #8434, opened 2023-11-14
CC | image-builder: bugfix incorrect partition location

Lu-Biao:
In the `image_builder.sh`, `part_start` of `image` and `dax_image` should exactly specify the same location, according to the parted documentation (https://www.gnu.org/software/parted/manual/parted.html#IEC-binary-units), to exactly specify the location, the units of partition start and end should use MiB.

### DIFF
--- a/tools/osbuilder/image-builder/image_builder.sh
+++ b/tools/osbuilder/image-builder/image_builder.sh
@@ -411,15 +411,20 @@ create_disk() {
 	# The partition is the rootfs content
 	info "Creating partitions"
 
+	if [ "${rootfs_end}" == "-1" ]; then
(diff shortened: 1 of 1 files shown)


ChengyuZhu6:
@Lu-Biao You should create an associated issue to track the changes and include `Fixes #{issue_id}` in the commit message to link the pull request to the corresponding issue.

fidencio:
@Lu-Biao, as the v0.8.0 release was the last one based on this branch, I'd recommend you open this PR against the `main` branch.  This one will be made read-only soon enough.

Lu-Biao:
@ChengyuZhu6 @fidencio  thanks, I will create an issue, and against main branch.

Issue #8435, opened 2023-11-14
image-builder:  incorrect partition location


Lu-Biao:
According to the description of `tools/osbuilder/image-builder/image_builder.sh`:
```
117 |       .-----------.----------.---------------.-----------.
118 |       | 0 - 512 B | 4 - 8 Kb |  2M - 2M+512B |    3M     |
119 |       |-----------+----------+---------------+-----------+
120 |       |   MBR #1  |   DAX    |    MBR #2     |  Rootfs   |
121 |       '-----------'----------'---------------'-----------+
122 |             |          |      ^      |        ^
123 |             |          '-data-'      '--------'
124 |             |                                 |
125 |             '--------rootfs-partition---------'

```

the `Rootfs` location in `MBR#1` should be the exactly same  with `MBR#2`.
According to parted's documentation (https://www.gnu.org/software/parted/manual/parted.html#IEC-binary-units), when executing `parted`, in order to specify the exact location, in `image_builder.sh`, the unit of partition start and end should use `MiB` instead of `M`.

PR #8436, opened 2023-11-14
image-builder: bugfix incorrect partition location

Lu-Biao:
In the `image_builder.sh`, `part_start` of `image` and `dax_image` should exactly specify the same location, according to the parted documentation (https://www.gnu.org/software/parted/manual/parted.html#IEC-binary-units), to exactly specify the location, the units of partition start and end should use MiB.

Fixes: #8435

### DIFF
--- a/tools/osbuilder/image-builder/image_builder.sh
+++ b/tools/osbuilder/image-builder/image_builder.sh
@@ -406,15 +406,20 @@ create_disk() {
 	# The partition is the rootfs content
 	info "Creating partitions"
 
+	if [ "${rootfs_end}" == "-1" ]; then
+		rootfs_end_unit="s"
+	else
+		rootfs_end_unit="MiB"
+	fi
 	if [ "${MEASURED_ROOTFS}" == "yes" ]; then
 		info "Creating partitions with hash device"
 		# The hash data will take less than one percent disk space to store
 		hash_start=$(echo $img_size | awk '{print $1 * 0.99}' |cut -d $(locale decimal_point) -f 1)
-		partition_param="mkpart primary ${fs_type} ${part_start}M ${hash_start}M "
-		partition_param+="mkpart primary ${fs_type} ${hash_start}M ${rootfs_end}M "
+		partition_param="mkpart primary ${fs_type} ${part_start}MiB ${hash_start}MiB "
+		partition_param+="mkpart primary ${fs_type} ${hash_start}MiB ${rootfs_end}${rootfs_end_unit} "
 		partition_param+="set 1 boot on"
 	else
-		partition_param="mkpart primary ${fs_type} ${part_start}M ${rootfs_end}M"
+		partition_param="mkpart primary ${fs_type} ${part_start}MiB ${rootfs_end}${rootfs_end_unit}"
 	fi
 
 	parted -s -a optimal "${image}" -- \


tools/osbuilder/image-builder/image_builder.sh
```diff
@@ -406,15 +406,20 @@ create_disk() {
...
+		rootfs_end_unit="MiB"
+	fi
 	if [ "${MEASURED_ROOTFS}" == "yes" ]; then
 		info "Creating partitions with hash device"
 		# The hash data will take less than one percent disk space to store
 		hash_start=$(echo $img_size | awk '{print $1 * 0.99}' |cut -d $(locale decimal_point) -f 1)
-		partition_param="mkpart primary ${fs_type} ${part_start}M ${hash_start}M "
-		partition_param+="mkpart primary ${fs_type} ${hash_start}M ${rootfs_end}M "
+		partition_param="mkpart primary ${fs_type} ${part_start}MiB ${hash_start}MiB "
+		partition_param+="mkpart primary ${fs_type} ${hash_start}MiB ${rootfs_end}${rootfs_end_unit} "
```
fidencio, line 419: Should those "MiB" also be replaced by ${rootfs_end_unit}?

tools/osbuilder/image-builder/image_builder.sh
```diff
@@ -406,15 +406,20 @@ create_disk() {
...
 		partition_param+="set 1 boot on"
 	else
-		partition_param="mkpart primary ${fs_type} ${part_start}M ${rootfs_end}M"
+		partition_param="mkpart primary ${fs_type} ${part_start}MiB ${rootfs_end}${rootfs_end_unit}"
```
fidencio, line 422: Same comment about the "MiB" used here.

fidencio (commented):
@Lu-Biao, thanks for the contribution!

I have raised two questions, but in general this one looks good.

tools/osbuilder/image-builder/image_builder.sh
Lu-Biao: Because `part_start` and `hash_start` are calculated based on `MiB`, `MiB` is used directly.
But `rootfs_end` is a variable and it is set to `-1` by default, so add a variable, `rootfs_end_unit`. If it is `-1`, the `rootfs_end_unit`  can only use `s(sector)` to point to the last sector of the disk.

fidencio (approved):
lgtm, thanks @Lu-Biao!

gkurz:
> @Lu-Biao wrong source branch?

Likely yes. https://github.com/Lu-Biao/kata-containers/commits/main doesn't look like a valid branch for a PR.

danmihai1:
If I remember correctly, the original patch was https://github.com/Lu-Biao/kata-containers/commit/d95e1000ede8d1743e32dbb5f6532e8782ed5d65. I tried that with the incorrect image format problem that I experienced a while ago, and my image started working.

gkurz:
> If I remember correctly, the original patch was [Lu-Biao@d95e100](https://github.com/Lu-Biao/kata-containers/commit/d95e1000ede8d1743e32dbb5f6532e8782ed5d65). I tried that with the incorrect image format problem that I experienced a while ago, and my image started working.


Hi @Lu-Biao ! Can you rebase your fix on current main and force-push ?

Lu-Biao:
@ChengyuZhu6   @gkurz @danmihai1 
Thanks your comments.  I have updated.

Lu-Biao:
> @Lu-Biao Please add a `Fixes #8435` stanza in the commit changelog.

@gkurz  thanks, added

Issue #8438, opened 2023-11-14
kata-manager | Add an option to install any given kata-static-tarball


fidencio:
Right now kata-manager will download a kata-static-tarball from a release, but we should allow users to install any given tarball.

The reasons for that are basically:
* Allow users to test code coming from `main` in an easy way
* Allow users to test their own build of kata-containers
* Allow CI to use the tarball generated as part of the CI pipeline.

PR #8439, opened 2023-11-14
utils: kata-manager: Allow installing kata from a given tarball

fidencio:
With this change, we give the users the change to try kata-containers with their own pre-built tarball.

This will become very useful in the CI context, as we won't be downloading a specific version of kata-containers, but rather installing whatever was built in previous steps of the CI pipeline.

Fixes: #8438


utils/kata-manager.sh
```diff
@@ -957,8 +968,9 @@ handle_args()
 
 	local kata_version=""
 	local containerd_flavour="lts"
+	local kata_tarball=""
 
-	while getopts "c:dDfhk:lortT" opt "$@"
+	while getopts "c:dDfhkK:lortT" opt "$@"
```
jodh-intel, line 973: I think we lost a colon. Shouldn't this be...

```suggestion
	while getopts "c:dDfhk:K:lortT" opt "$@"
```

utils/kata-manager.sh
```diff
@@ -838,10 +847,12 @@ handle_installation()
 	# These params can be blank
 	local kata_version="${7:-}"
 	local containerd_flavour="${8:-}"
+	local kata_tarball="${10:-}"
 
 	local install_docker="${9:-}"
 	[ -z "$install_docker" ] && die "no install docker value"
 
+
```
jodh-intel, line 855: Can you move the new param variable down below the `$9` one:

```suggestion
	local install_docker="${9:-}"
	[ -z "$install_docker" ] && die "no install docker value"

	local kata_tarball="${10:-}"
```

utils/kata-manager.sh
```diff
@@ -680,11 +688,12 @@ configure_kata()
 handle_kata()
 {
 	local version="${1:-}"
+	local tarball="${2:-}"
 
 	local enable_debug="${2:-}"
```
jodh-intel, line 693:
```suggestion
	local tarball="${2:-}"

	local enable_debug="${3:-}"
```

utils/kata-manager.sh
```diff
@@ -264,6 +264,7 @@ Options:
  -f           : Force installation (use with care).
  -h           : Show this help statement.
  -k <version> : Specify Kata Containers version.
+ -K <tarball> : Specify Kata Containers tarball (takes priority over '-k').
```
jodh-intel, line 267:
```suggestion
 -K <tarball> : Specify local Kata Containers tarball to install (takes priority over '-k').
```

utils/kata-manager.sh
```diff
@@ -680,11 +688,12 @@ configure_kata()
 handle_kata()
 {
 	local version="${1:-}"
+	local tarball="${2:-}"
 
 	local enable_debug="${2:-}"
 	[ -z "$enable_debug" ] && die "no enable debug value"
 
-	install_kata "$version" "$enable_debug"
+	install_kata "$version" "$tarball" "$enable_debug"
```
jodh-intel, line 696: We seem to have removed the debug arg, so:

```suggestion
	install_kata "$version" "$tarball"
```

jodh-intel (commented):
Thanks @fidencio - A few comments...

utils/kata-manager.sh
fidencio: I've removed this one, but as part of a different commit, as it was not exactly related to this PR.

fidencio:
@jodh-intel, all the comments have been addressed, thanks!

utils/kata-manager.sh
```diff
@@ -583,29 +584,36 @@ configure_containerd()
...
+		[ -z "$results" ] && die "Cannot download $project release file"
+	
+		version=$(echo "$results"|cut -d: -f1)
+
+		[ -z "$version" ] && die "Cannot determine $project resolved version"
+
+		local file
+		file=$(echo "$results"|cut -d: -f2-)
+	else
+		file="$kata_tarball"
```
jodh-intel, line 614: Nit: This works fine but you end up with a message like this (with my hash comments below the output):

```
Installing Kata Containers release  from /home/james/data/kata/kata-static-3.2.0-amd64.tar.xz
#                                 ~~
#                                 |
#                            no version
```

How about we change the `info` call below to something like the following to resolve that:

```bash
if [ -n "$kata_tarball" ]
then
    info "Installing $project release from '$file'"
else
    info "Installing $project release $version from '$file'"
fi
```

jodh-intel (approved):
Thanks @fidencio - tests fine but one nit you might want to look at. Overall...

lgtm

beraldoleal (approved):
Thanks. Useful. lgtm

wainersm:
Interesting I never used the kata-manager.sh. Has it always been a first class citizen and I didn't know? or are you perhaps trying to bring it back to life?

fidencio:
> Interesting I never used the kata-manager.sh. Has it always been a first class citizen and I didn't know? or are you perhaps trying to bring it back to life?

It existed but I'd say it was never a first class citizen.  My plan, together with @jodh-intel, is to:
* Make this the way to install Kata Containers via `curl https://get.katacontainers.io | sh - ...`
* Use this as part of all the tests

But there's still work to be done on this front. :-)

jodh-intel:
@wainersm - It's been used for the "Install a complete Kata with a single command" use-case for a few years now. See https://github.com/kata-containers/kata-containers/tree/main/docs/install#automatic-installation.

Issue #8444, opened 2023-11-14
build/packaging: runtime-rs: Move rust runtime config files to runtime-rs directory


jodh-intel:
The configuration files for the rust runtime should be stored in a separate "`runtime-rs`" directory to allow users and tooling to distinguish the difference between golang and rust Kata config files.

## Rationale

- In some cases, the rust runtime configuration syntax is slightly different to the golang runtime configuration files so some hypervisors need two different configuration files, one for reach runtime type (for example Cloud Hypervisor which uses `clh` for the golang runtime and `cloud-hypervisor` for the rust runtime.

- Some hypervisors only currently work with the golang runtime.

- Some hypervisors only work with the rust runtime (dragonball).

Related: #8306, #6020.

Issue #6020, opened 2023-01-10
runtime-rs: kata config gap between runtime-rs and runtime-go


studychao:
Lately, several kata config gap between runtime-rs and runtime-go have been discussed, such as #6018 #5954 .

Besides fixing two problems, we need to take a deeper look at the config gap, and :
1. If there is any config missing, we need to add it back if needed.
2. If there is any config only used in runtime-rs, we need to write comments into runtime-rs config to declear that it is only needed in the runtime-rs config.
3. If there is any config missing but we choose not to add it back into runtime-rs, we need to find a place to declear why we choose not to make this change.

For the point 3, we may need a document somewhere to document the behavior / config difference between runtime-rs and runtime-go , in order to help users to understand runtime-rs more.

studychao:
cc @jodh-intel @liubin

jodh-intel:
Related: #5954, #5959.

PR #8445, opened 2023-11-14
build/kata-deploy: Move rust runtime config files to runtime-rs directory

jodh-intel:
Install the rust runtime configuration files to a `runtime-rs/`
directory to distinguish them from the golang config files (which may
have a different syntax).

The default values mean that the rust config files are now installed to
`/opt/kata/share/defaults/kata-containers/runtime-rs/` rather than
`/opt/kata/share/defaults/kata-containers/`.

See: https://github.com/kata-containers/kata-containers/issues/6020

Also:

- Updates `kata-deploy` to include the rust runtime CH config file and als to make use of the `runtime-rs/` directory for all rust runtime config files.
- Updates the rust runtime to look in the `runtime-rs/` directories before looking for config files in the "traditional" (golang runtime) locations.

Fixes: #8444.

### DIFF
--- a/src/libs/kata-types/src/config/default.rs
+++ b/src/libs/kata-types/src/config/default.rs
@@ -14,6 +14,12 @@ use lazy_static::lazy_static;
 lazy_static! {
     /// Default configuration file paths, vendor may extend the list
     pub static ref DEFAULT_RUNTIME_CONFIGURATIONS: Vec::<&'static str> = vec![
+        // The rust runtime specific paths
+        "/etc/kata-containers/runtime-rs/configuration.toml",
+        "/usr/share/defaults/kata-containers/runtime-rs/configuration.toml",
+        "/opt/kata/share/defaults/kata-containers/runtime-rs/configuration.toml",
+
+        // The traditional golang paths
         "/etc/kata-containers/configuration.toml",
         "/usr/share/defaults/kata-containers/configuration.toml",
         "/opt/kata/share/defaults/kata-containers/configuration.toml",
--- a/src/runtime-rs/Makefile
+++ b/src/runtime-rs/Makefile
@@ -189,7 +189,7 @@ KNOWN_HYPERVISORS =
 # List of hypervisors known for the current architecture
 KNOWN_HYPERVISORS =
 
-CONFDIR := $(DEFAULTSDIR)/$(PROJECT_DIR)
+CONFDIR := $(DEFAULTSDIR)/$(PROJECT_DIR)/runtime-rs
 SYSCONFDIR := $(SYSCONFDIR)/$(PROJECT_DIR)
 ##VAR CONFIG_PATH=<path> Main configuration file location for stateless systems
 CONFIG_PATH := $(abspath $(CONFDIR)/$(CONFIG_FILE))
--- a/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
+++ b/tools/packaging/kata-deploy/kata-cleanup/base/kata-cleanup.yaml
@@ -30,7 +30,7 @@ spec:
             - name: DEBUG
               value: "false"
             - name: SHIMS
-              value: "clh dragonball fc qemu-nvidia-gpu qemu-sev qemu-snp qemu-tdx qemu"
+              value: "clh cloud-hypervisor dragonball fc qemu-nvidia-gpu qemu-sev qemu-snp qemu-tdx qemu"
             - name: DEFAULT_SHIM
               value: "qemu"
             - name: CREATE_RUNTIMECLASSES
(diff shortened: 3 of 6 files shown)


fidencio:
@jodh-intel, this patchset is very clean and does exactly what it should be doing.  However, there are a few points that need to be addressed and I'd like to have a discussion on those to fully understand the path forward.

Thankfully to the change of `clh` to `cloud-hypervisor`, the runtime class names won't clash.  I assume `fc` could just switch to `firecracker` and hey, we're out of the clash there as well.  What about QEMU, though?  I'm not sure when QEMU will be added to runtime-rs, but I can see a clash in the runtime class name happening there.

Another part of kata-deploy that needs to be updated is https://github.com/kata-containers/kata-containers/pull/8445/files#diff-08a680425cfe93e69c4108dc8fd91a9aeaefbf016bd9941b03c66834bf87a341R237-R239, as right now we'd only install the runtime-rs when `dragonball` is used.  Now with your patch we want to also do that when `cloud-hypervisor`  will be used with runtime-rs, and that should be a simple change.

Looking at the long term, I still trying to understand how we'll differentiate QEMU, and maybe we'll have to leave that discussion / problem for our future selves to solve, but I still would like to hear what's your take on this.

All in all, thanks a lot for the PR!

fidencio:
@jodh-intel, I'm thinking here about having a `experimental_runtime` environment variable added, which could allow folks to simply select between the go and rust implementations.

This would make sure that whenever we deploy the rust one we will just deploy the rust artefacts, and also make sure that we can have a clean way to get folks to use either the runtime-rs or the go runtime.

If you agree with the idea, what I'd like to do is to provide you a PoC atop of your changes here.

jodh-intel:
Hi @fidencio - We could "sort of" solve this problem by resolving #6020, but I do feel that isn't the best solution here. It sounds sensible to make the golang and rust runtime config compatible which would _almost_ make this PR redundant, but given the existing complications (dragonball is rust only, CH has different formats and config names between golang and rust runtimes, QEMU and FC aren't fully enabled in rust yet, the runtimes provide slightly different feature sets), I do feel we should keep the assets as separate as possible for now to avoid any confusion.

> If you agree with the idea, what I'd like to do is to provide you a PoC atop of your changes here.

+1 - That sounds like a good idea.

> This would make sure that whenever we deploy the rust one we will just deploy the rust artefacts, 

This also sounds like a good idea forthe majority of users and will provide a smaller installation. Maybe we could also accept a magic value of `experimental_runtime=all` to install both runtimes as an option?

I guess if the user changes their mind, they can simply re-run with or without `experimental_runtime` to select the "other" runtime.

Again, to minimise confusion, I think it would be useful if `kata-deploy` displays an `info()` message showing which runtime was selected by the user (explicitly or implicitly) and how it was chosen (was `experimental_runtime` set?) To be even more clear, how about we add a call to `logger(1)` along the lines of:

```bash
logger -i -t "kata-deploy" "installing Kata Containers '$runtime_type' runtime (experimental_runtime variable set to: '$experimental_runtime')" 
```

fidencio:
> This also sounds like a good idea forthe majority of users and will provide a smaller installation. Maybe we could also accept a magic value of `experimental_runtime=all` to install both runtimes as an option?

This is more complicated, as we need to link `containerd-shim-kata-v2` to either one or another runtime.
Right now we have never supported both installations, although we never explicitly raised this limitation.

I will finish a POC during the day and send you a branch so you can take a look at it.

fidencio:
Just for transparency sake, I've poken @jodh-intel internally, and provided him a simple PoC adding the `RUST_RUNTIME` env var that can be passed to kata-deploy, and it's here: https://github.com/fidencio/kata-containers/commits/mv-rust-config-files-to-subdir

Note, this was not tested, just hacked together.

We'll also need to change runtimeclasses and things like that to match what I'm proposing, but I can take care of that together with James.

src/libs/kata-types/src/config/default.rs
```diff
@@ -14,6 +14,12 @@ use lazy_static::lazy_static;
 lazy_static! {
     /// Default configuration file paths, vendor may extend the list
     pub static ref DEFAULT_RUNTIME_CONFIGURATIONS: Vec::<&'static str> = vec![
+        // The rust runtime specific paths
```
amshinde, line 17: @jodh-intel Should the rust runtime looks for configs only under  dir "runtime-rs". I am not sure we need to keep both paths since the configs for golang and rust runtime are not compatible.

I know that the rust Makefile installs under `runtime-rs` directory, do we do the same for the configs as well?

amshinde:
> We'll also need to change runtimeclasses and things like that to match what I'm proposing, but I can take care of that together with James.

@fidencio @jodh-intel Are we planning to expose runtimeclasses in such a way that the end-user can use both the golang and rust runtimes at the same time, so that the runtimeclasses need to be named differently or based on a flag just one or the other runtime is installed and used. 
(Like you pointed out the point of contention would be handling the naming of the qemu config, runtime and runtimeclass)

jodh-intel:
Closing in favour of #8483.

Issue #8446, opened 2023-11-14
runtime-rs: make compilation for QEMU on s390x


BbolroC:
**Which feature do you think can be improved?**

`runtime-rs` has been compiled only for `x86_64` and `aarch64` for a while. With the recent addition of a basic QEMU implementation (https://github.com/kata-containers/kata-containers/pull/5798), it would be a good starting point for the s390x enablement if `runtime-rs` is compiled with QEMU on s390x.

PR #8454, opened 2023-11-15
runtime-rs: make compilation for QEMU on s390x

BbolroC:
This PR is to make `runtime-rs` compilable with QEMU on s390x. This is just a base work for the s390x enablement. There must be a few more steps (implementation for s390x-specific) to run a container with runtime-rs.

Fixes: #8446

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/lib.rs
+++ b/src/runtime-rs/crates/hypervisor/src/lib.rs
@@ -13,6 +13,7 @@ pub mod device;
 pub mod hypervisor_persist;
 pub use device::driver::*;
 use device::DeviceType;
+#[cfg(not(target_arch = "s390x"))]
 pub mod dragonball;
 mod kernel_param;
 pub mod qemu;
@@ -20,7 +21,7 @@ pub use kernel_param::Param;
 pub mod utils;
 use std::collections::HashMap;
 
-#[cfg(feature = "cloud-hypervisor")]
+#[cfg(all(feature = "cloud-hypervisor", not(target_arch = "s390x")))]
 pub mod ch;
 
 use anyhow::Result;
@@ -49,9 +50,12 @@ const VM_ROOTFS_FILESYSTEM_EROFS: &str = "erofs";
 // /dev/hugepages will be the mount point
 // mkdir -p /dev/hugepages
 // mount -t hugetlbfs none /dev/hugepages
+#[cfg(not(target_arch = "s390x"))]
 const DEV_HUGEPAGES: &str = "/dev/hugepages";
 pub const HUGETLBFS: &str = "hugetlbfs";
+#[cfg(not(target_arch = "s390x"))]
 const SHMEM: &str = "shmem";
+#[cfg(not(target_arch = "s390x"))]
 const HUGE_SHMEM: &str = "hugeshmem";
 
 pub const HYPERVISOR_DRAGONBALL: &str = "dragonball";
@@ -60,6 +64,7 @@ pub const HYPERVISOR_QEMU: &str = "qemu";
 pub const DEFAULT_HYBRID_VSOCK_NAME: &str = "kata.hvsock";
 pub const JAILER_ROOT: &str = "root";
 
+#[cfg(not(target_arch = "s390x"))]
 #[derive(PartialEq, Debug, Clone)]
 pub(crate) enum VmmState {
     NotReady,
--- a/src/runtime-rs/crates/runtimes/virt_container/src/lib.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/lib.rs
@@ -20,15 +20,17 @@ use agent::{kata::KataAgent, AGENT_KATA};
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use common::{message::Message, RuntimeHandler, RuntimeInstance};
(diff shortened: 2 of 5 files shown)


src/runtime-rs/crates/hypervisor/src/lib.rs
```diff
@@ -45,14 +46,7 @@ const VM_ROOTFS_FILESYSTEM_EXT4: &str = "ext4";
...
 const VM_ROOTFS_FILESYSTEM_EROFS: &str = "erofs";
 
-// before using hugepages for VM, we need to mount hugetlbfs
-// /dev/hugepages will be the mount point
-// mkdir -p /dev/hugepages
-// mount -t hugetlbfs none /dev/hugepages
-const DEV_HUGEPAGES: &str = "/dev/hugepages";
 pub const HUGETLBFS: &str = "hugetlbfs";
-const SHMEM: &str = "shmem";
-const HUGE_SHMEM: &str = "hugeshmem";
```
Apokleos, line 55: I want to confirm that the change is for compilation?

Apokleos (commented):
Hi @BbolroC, Thx for your work.
Can we break this PR down into smaller commits? The commit messages don't seem to match the code changes. For example, a lot of the code changes don't seem to be related to "make compilation for QEMU on s390x".

src/runtime-rs/crates/hypervisor/src/lib.rs
BbolroC: Yes, they are only being used by `Dragonball`. So I faced with the `never used` error unless I move them to the hypervisor using them.

BbolroC:
> Hi @BbolroC, Thx for your work. Can we break this PR down into smaller commits? The commit messages don't seem to match the code changes. For example, a lot of the code changes don't seem to be related to "make compilation for QEMU on s390x".

Yes, I will definitely. Thanks for the comment. I also got the similar comment from @gkurz. 

But at the moment this PR has lots of conflict with https://github.com/kata-containers/kata-containers/pull/8185 (upcoming QEMU implementation) which also depends on https://github.com/kata-containers/kata-containers/pull/8414. I will start re-working this PR after getting the confirmation for going ahead from @pmores . Thanks. 😉

cmaf (commented):
Hi @BbolroC, one initial question. I would also be grateful if you could break this into smaller commits.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -7,7 +7,7 @@
 use super::vmm_instance::VmmInstance;
 use crate::{
     device::DeviceType, hypervisor_persist::HypervisorState, kernel_param::KernelParams, VmmState,
-    DEV_HUGEPAGES, HUGETLBFS, HUGE_SHMEM, HYPERVISOR_DRAGONBALL, SHMEM,
```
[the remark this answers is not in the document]
BbolroC: If I understand your question correctly, they are not removed actually. They are just moved from `src/lib.rs` to `dragonball/inner.rs` other than `HUGETLBFS` The moved constants are only used by dragonball, which gave me the `never used` compile error when the hypervisor is excluded from the compilation on s390x. `HUGETLBFS` is left in `src/lib.rs` as it was because it is used by the `resource` crate. Any suggestion for leaving them as they were while not facing with the compile error would be appreciated. Thanks.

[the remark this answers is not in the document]
cmaf: @BbolroC thanks, I understand

src/runtime-rs/crates/runtimes/virt_container/src/lib.rs
```diff
@@ -20,15 +20,19 @@ use agent::{kata::KataAgent, AGENT_KATA};
...
 use kata_types::config::{
-    hypervisor::register_hypervisor_plugin, DragonballConfig, QemuConfig, TomlConfig,
+    hypervisor::register_hypervisor_plugin, QemuConfig, TomlConfig,
 };
 
-#[cfg(feature = "cloud-hypervisor")]
+#[cfg(all(feature = "cloud-hypervisor", not(target_arch = "s390x")))]
 use hypervisor::ch::CloudHypervisor;
-#[cfg(feature = "cloud-hypervisor")]
+#[cfg(all(feature = "cloud-hypervisor", not(target_arch = "s390x")))]
```
pmores, line 35: I would love for the resulting code to have fewer conditional compilation lines but frankly, I don't see how to achieve that at the moment, without a major refactor to group individual hypervisors' code more closely together.

pmores (approved):
LGTM, thanks @BbolroC !

gkurz (changes requested):
It is certain that you had to enable the compilation first in order to hit the build issues and fix them. Now that this code is going to be merged into main, please consider bisectability : in order for the enablement patch to not break the build, it should be applied **after** the fixes, i.e. last in the PR.

BbolroC:
> It is certain that you had to enable the compilation first in order to hit the build issues and fix them. Now that this code is going to be merged into main, please consider bisectability : in order for the enablement patch to not break the build, it should be applied **after** the fixes, i.e. last in the PR.

I have finished studying `git bisect` and got what it means and implies on finding a bad commit. The enablement path has been moved to the latest place. Thanks for the suggestion.

gkurz (approved):
LGTM. Thanks for understanding the `git bisect` related request @BbolroC ! :smiley:

Issue #8458, opened 2023-11-16
github: add workflows for building and publishing kata tarball for ppc64le


Amulyam24:
This is the first step in migrating from Jenkins to GHA.

PR #8459, opened 2023-11-16
github: add workflows for building and publishing kata artefacts on ppc64le

Amulyam24:
This is the first step of the migration of jobs from Jenkins to GHA. 

Fixes: #8458

### DIFF
--- a/.github/workflows/build-kata-static-tarball-ppc64le.yaml
+++ b/.github/workflows/build-kata-static-tarball-ppc64le.yaml
@@ -1,0 +1,116 @@
+name: CI | Build kata-static tarball for ppc64le
+on:
+  workflow_call:
+    inputs:
+      stage:
+        required: false
+        type: string
+        default: test
+      tarball-suffix:
+        required: false
+        type: string
+      push-to-registry:
+        required: false
+        type: string
+        default: no
+      commit-hash:
+        required: false
+        type: string
+      target-branch:
+        required: false
+        type: string
+        default: ""
+
+jobs:
+  build-asset:
+    runs-on: ppc64le
+    strategy:
+      matrix:
+        asset:
+          - kernel
+          - qemu
+          - rootfs-initrd
+          - shim-v2
+          - virtiofsd
+        stage:
+          - ${{ inputs.stage }}
+    steps:
+      - name: Adjust a permission for repo
+        run: |
+          sudo chown -R $USER:$USER $GITHUB_WORKSPACE
+
+      - name: Login to Kata Containers quay.io
+        if: ${{ inputs.push-to-registry == 'yes' }}
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_DEPLOYER_USERNAME }}
+          password: ${{ secrets.QUAY_DEPLOYER_PASSWORD }}
+
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.commit-hash }}
+          fetch-depth: 0 # This is needed in order to keep the commit ids history
+
+      - name: Rebase atop of the latest target branch
+        run: |
+          ./tests/git-helper.sh "rebase-atop-of-the-latest-target-branch"
+        env:
+          TARGET_BRANCH: ${{ inputs.target-branch }}
+
+      - name: Build ${{ matrix.asset }}
+        run: |
+          make "${KATA_ASSET}-tarball"
+          build_dir=$(readlink -f build)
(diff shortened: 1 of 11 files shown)


Amulyam24:
Hi @fidencio @stevenhorsman , can you please help me in adding a self hosted ppc64le runner?

stevenhorsman:
> Hi @fidencio @stevenhorsman , can you please help me in adding a self hosted ppc64le runner?

Sure, send me a slack when you are ready and I can send through the token and instructions

.github/workflows/build-kata-static-tarball-ppc64le.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Rebase atop of the latest target branch
+        run: |
+          ./tests/git-helper.sh "rebase-atop-of-the-latest-target-branch"
+        env:
+          TARGET_BRANCH: ${{ inputs.target-branch }}
+
+      - name: Build ${{ matrix.asset }}
+        run: |
+          CROSS_BUILD=true ARCH=ppc64le TARGET_ARCH=ppc64le make "${KATA_ASSET}-tarball"
```
BbolroC, line 63: Assuming that a runner is a ppc64le native machine, I was wondering if it is necessary.

tools/packaging/kata-deploy/local-build/dockerbuild/install_oras.sh
```diff
@@ -29,9 +29,21 @@ fi
 
 arch=$(uname -m)
 if [ "${arch}" = "ppc64le" ]; then
-	echo "An ORAS release for ppc64le is not available yet."
-	exit 0
-fi
+ 	echo "Building oras from source"
+	go_version="go1.21.3"
```
BbolroC, line 33: Why don't you reuse the existing [script](https://github.com/kata-containers/kata-containers/blob/main/ci/install_go.sh) for this?

BbolroC (commented):
Thanks for the request. I've left 2 comments. 

Additionally, I've noticed that we finally have workflows to build artifacts for 4 different architectures where no steps including a `if: always()` condition for a self-hosted runner (running 24/7) are engaged, but to produce lots of duplicated codes. 

I was wondering if this PR would be a right place to incorporate the workflows or we could do it in a separate PR. @fidencio and @stevenhorsman

tools/packaging/kata-deploy/local-build/dockerbuild/install_oras.sh
BbolroC: Hmm... it fetches the `test` repo which is going to be deprecated soon. I think we have to rewrite the script in a separate PR and refactor this there. Just ignore this comment.

BbolroC (approved):
Other than the comment regarding the deduplication, LGTM! Thanks @Amulyam24 !

tools/packaging/kata-deploy/local-build/dockerbuild/install_oras.sh
Amulyam24: @BbolroC, this is a temporary fix to build and use Oras, once the release it out for ppc64le, I will switch to using the artefacts from releases.

stevenhorsman (approved):
It would be good to see a run on this testing on your fork, but I think the code looks okay and expected.

I definitely understand @BbolroC's comments about the duplication:
- The build-kata-static-tarball workflow only differs in artifact name, runner and assets that are created
- The publish workflows only varies on the artifact name and self-host runner prepare in ppc
- The release workflows are pretty identical other than the build workflows they call
I'd like to hear @fidencio's thought, but I think we should consider refactoring these to try and get a single workflow with passing the arch in. We might have a couple of steps like:
```
 - name: Prepare the self-hosted runner
        if: ${{ arch == 'ppc64le' }}
        run: ${HOME}/scripts/prepare_runner.sh)
```
but I think it's worth tying to see if it's not too much of a mess and would potentially save on the maintenance cost of updating 4 versions in future.
I'm okay for this to be done as a separate PR though.

Amulyam24:
@stevenhorsman, testing done for both this and containerd PR, PTAL for reference - https://github.com/Amulyam24/kata-containers/actions/runs/6979218191/job/18992074251?pr=6

GabyCT (approved):

Amulyam24:
Hi @stevenhorsman, we have 3 approvals, can this PR be merged?

fidencio:
> I'm okay for this to be done as a separate PR though.

I am with Steve and Choi here, but I'd for sure prefer this addition coming on a different PR.

Issue #7668, opened 2023-08-17
support for setting and verifying the Agent Policy hash value in SEV-SNP Host Data 


danmihai1:
**Which feature do you think can be improved?**

Agent Policy (https://github.com/kata-containers/kata-containers/issues/7573)

**How can it be improved?**

- The Shim should be able to set the Policy hash value into the Host Data field of SEV-SNP Guest VMs.
- The Agent should be able to verify that the value of the Host Data field matches the Policy file it gets from the Shim.

**Additional Information**

CoCo remote attestation should also check the expected value of the Host Data field.

**Before raising this enhancement request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?
Yes

davidhadas:
@danmihai1 

Sorry for the last response

Depending on remote attestation to verify the policy can be poor from an operational standpoint. Most organizations using Kubernets will be using CI/CD automation and the Yaml files are just one more resource in such a CI/CD system. CoCo should align to the standard procedures used in Kubernetes. 

When a DevOp team adds some ENV for example, or changes a ConfigMap, related to one of the several pod types in one of the several workloads the team is responsible for, this change should streamline with CI/CD. From an operational point of view, it is perfectly ok to run some automated tool to deduct a policy (e.g. genpolicy ) or require the DevOp team to also make some change in a corresponding policy file. However, the use of remote attestation for verifying such a change means that now, with every change in every yaml file of every pod type of every workload in the organization using CoCo, a new measurement needs to be added.

I believe CoCo avoided measuring Pod Images for this same reason.
Coco could have chosen to let the Kata Agent hash the downloaded pod image and add it to the measurement. 
However, this creates an unreasonable operational mechanism as now, the measurement of every Guest VM change every time someone making a change to a pod image. 
Instead, CoCo measures the Guest VM and use signatures or encryption for pod images. 

I believe this is much a better approach for policy as well.  
The policy should be signed and/or encrypted and the Kata Agent can then verify the policy using keys received following attestation.

PR #8469, opened 2023-11-17
do-not-merge: runtime: agent: verify the agent policy hash

danmihai1:
Attach the policy hash to SEV-SNP VMs and verify the value of the hash from SetPolicy.

Fixes: #7668

### DIFF
--- a/src/agent/src/sev.rs
+++ b/src/agent/src/sev.rs
@@ -1,0 +1,19 @@
+// Copyright (c) 2023 Microsoft Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::Result;
+
+pub fn get_snp_host_data() -> Result<Vec<u8>> {
+    match sev::firmware::guest::Firmware::open() {
+        Ok(mut firmware) => {
+            let report_data: [u8; 64] = [0; 64];
+            match firmware.get_report(None, Some(report_data), Some(0)) {
+                Ok(report) => Ok(report.host_data.to_vec()),
+                Err(e) => Err(e.into()),
+            }
+        }
+        Err(e) => Err(e.into()),
+    }
+}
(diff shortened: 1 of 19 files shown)


src/agent/src/sev.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2023 Microsoft Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::Result;
+
+pub fn get_snp_host_data() -> Result<Vec<u8>> {
```
Xynnn007, line 8: Get hostdata field in kata-agent seems arch-specific. I'd suggest to do this in AA side. As it already has the ability to get hardware evidence/quote. TDX is the same.

danmihai1: I don't think we should force users to use the CoCo AA. Users should be able to perform remote attestation on their own if they prefer to - based on the arch specific docs.

However, sharing a rust crate between the Kata Agent and AA would sound reasonable to me - not sure if this is what you meant.

If you meant adding some protocol for Policy validation between the Kata Agent and AA, please describe the details when we'll meet. For example, we'd need that kind of protocol to succeed before executing many other Kata Agent API calls - otherwise a not-yet-validated Policy might allow dangerous actions.

Xynnn007: IMO we should try to make kata-agent arch-independent. Some reasons:

1. CoCoAA has already had the ability of remote attestation related things, so we should try to leverage them rather than re-invention. As you said, users should be able to perform remote attestation on their own if they prefer to. We are refactoring AA to support only have attestation related functionalities like GetAttestationToken and GetEvidence. This means that not only KBS/CoCo-AS can be accessed, but also MAA/Intel TA and other ASes/ways could implement the specific plugins. If we insist to implement the logic inside kata-agent, there will be two copies of the code.
2. If we want to check the hostdata field, it hints that we already enable the confidential computing feature. Furthermore confidential computing feature means that CoCo-AA will be built-in.

PR #8483, opened 2023-11-21
build/kata-deploy: Move rust runtime config files to runtime-rs directory -- based on #8445

fidencio:
Install the rust runtime configuration files to a `runtime-rs/`
directory to distinguish them from the golang config files (which may
have a different syntax).

The default values mean that the rust config files are now installed to
`/opt/kata/share/defaults/kata-containers/runtime-rs/` rather than
`/opt/kata/share/defaults/kata-containers/`.

See: https://github.com/kata-containers/kata-containers/issues/6020

Also:

- Updates `kata-deploy` to include the rust runtime CH config file and als to make use of the `runtime-rs/` directory for all rust runtime config files.
- Updates the rust runtime to look in the `runtime-rs/` directories before looking for config files in the "traditional" (golang runtime) locations.

Fixes: #8444.


fidencio:
@jodh-intel, I took the freedom to do some ground work needed for your #8445 PR and then have #8445 atop of mine.

I'd like to ask you for a review on the patches that are coming from me, as I'd like to know what's your take on this approach.

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -8,6 +8,14 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
+RUST_RUNTIME="${RUST_RUNTIME:-false}"
+supported_rust_runtimes="dragonball cloud-hypervisor"
+supported_go_runtimes="clh fc qemu qemu-nvidia-gpu qemu-sev qemu-snp qemu-tdx remote stratovirt"
+supported_runtimes="${allowed_go_runtimes}"
```
jodh-intel, line 14: Should this be:

```suggestion
supported_runtimes="${supported_go_runtimes}"
```

tools/packaging/kata-deploy/runtimeclasses/kata-rustRuntimeClasses.yaml
```diff
@@ -0,0 +1,26 @@
+---
+kind: RuntimeClass
+apiVersion: node.k8s.io/v1
+metadata:
+    name: kata-cloud-hypervisor
```
jodh-intel, line 5: Random thought: I know the class is specified in the filename, but since YAML supports comments, how about we add one stating explicitly that `cloud-hypervisor` only refers to CH with the rust runtime.

Alternatively, we could add a `metadata.description` that could do the same job but be programmatically queryable.

jodh-intel (commented):
Thanks @fidencio - a couple of comments...

tools/packaging/kata-deploy/runtimeclasses/kata-rustRuntimeClasses.yaml
fidencio: I like this idea a lot!  I will add that, @jodh-intel!

fidencio: I had to drop the metadata.description bits as those are not supported by k8s, causing the runtime-class to fail to be applied.

jodh-intel (approved):
Thanks @fidencio.

lgtm

chavafg (approved):

fidencio:
dragonball CIs are failing, and I'm debugging what has to be changed there.

fidencio:
So, dragonball CI will fail as it does require the same changes we're doing for GHA to be done on the tests repo.

As we're moving all the non-x86_64 CIs to GHA, I wonder what would be the best approach here.
/cc @studychao

studychao:
> So, dragonball CI will fail as it does require the same changes we're doing for GHA to be done on the tests repo.
> 
> As we're moving all the non-x86_64 CIs to GHA, I wonder what would be the best approach here. /cc @studychao

Hi there, will the arm tests also be transferred to GHA?

fidencio:
> Hi there, will the arm tests also be transferred to GHA?

They will, and @jongwu is already working on that.

fidencio:
> So there will a period time that Dragonball doesn't have arm CI support if we don't move the change to test repo?
> Do you know how long will this vacancy last? Or when would the arm CI ready in GHA?
> cc @jongwu

Today we've added the arm64 machine for non-k8s test to the GHA.
I'd say it should be okay in a week or so, following the work that's being done for other architectures as well.

studychao:
> > So there will a period time that Dragonball doesn't have arm CI support if we don't move the change to test repo?
> > Do you know how long will this vacancy last? Or when would the arm CI ready in GHA?
> > cc @jongwu
> 
> Today we've added the arm64 machine for non-k8s test to the GHA. I'd say it should be okay in a week or so, following the work that's being done for other architectures as well.

I think it's okay that arm64 Dragonball is not covered by CI for one week in the transition from Jenkins to GHA. WDYT @jongwu ?

fidencio:
Based on the conversation, I'm marking the dragonball CI as non-required and I will have this one merged.

Issue #8505, opened 2023-11-23
deployment: Create a Helm Chart for Kata


zvonkok:
To make it easier for consuming Kata let's build an Helm Chart for deploying the stack. 
Helm charts are supported in several GitOps pipelines, ArgoCD, Flux ...

Issue #8510, opened 2023-11-23
deployment: Create stable/latest overlays for k3s, k0s, rke2


zvonkok:
Create additional overlays for deploying k3s, k02 and rke2 with stable or latest

PR #8514, opened 2023-11-23
 deployment: Create a Helm Chart for Kata

zvonkok:
To make it easier for consuming Kata let's build an Helm Chart for deploying the stack.
Helm charts are supported in several GitOps pipelines, ArgoCD, Flux ...

Fixes:  #8505
DependsOn:  #8510


versions.yaml
```diff
@@ -87,9 +87,9 @@ assets:
 
     qemu:
       description: "VMM that uses KVM"
-      url: "https://github.com/qemu/qemu"
-      version: "v7.2.0"
-      tag: "v7.2.0"
+      url: "https://github.com/zvonkok/qemu"
+      version: "switch-port"
+      tag: "switch-port"
```
fidencio, line 92: This is not related to this PR. :-)

PR #8516, opened 2023-11-24
move vsock device into device manager

Apokleos:


src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
```diff
@@ -121,46 +118,85 @@ const CID_RETRY_COUNT: u32 = 50;
...
-                }
-                Err(nix::Error::EADDRINUSE) => {
-                    // The CID is already in use. Try another one.
-                }
-                Err(err) => {
-                    return Err(err).context("failed to set guest CID");
-                }
-            }
-        }
+            .context("add hybrid vsock device.")?;
```
pmores, line 135: This should read `"add vsock device"`, shouldn't it?

pmores:
Hi @Apokleos , thanks for your PR!  The first commit seems largely what we discussed in depth at PR #8414 so that looks good to me already.  I have a question about how `VsockDevice` would be used after its introduction into `DeviceManager`.  

So far, the QEMU driver (in PR #8185) knew it would need a `VsockDevice` so it created one.  With `VsockDevice` handled by `DeviceManager`, that wouldn't seem to make much sense anymore, but who and when will create it now?

gkurz (commented):
Hi @Apokleos ! Thanks for submitting this PR. It looks quite good and we should be able to merge it soon.

A few general comments :
- "[runtime-rs: move vosck device into device manager](https://github.com/kata-containers/kata-containers/pull/8516/commits/296911768954774fb62456370a07b62e96e14156) doesn't seem to do anything with the device manager but [runtime-rs: add vsock device in device manager](https://github.com/kata-containers/kata-containers/pull/8516/commits/c1f2bd787894e8880a75d6ddc78fad36f9d17cb6) does... I suspect the first commit should have another title like "Add vsock to the DeviceType enum".
- please split out the introduction of `generate_vhost_sock_cid()` to an initial commit
- please populate the commit logs :smile:

Cheers.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
Apokleos: Thx @pmores!
Fixed

Apokleos:
> Hi @Apokleos , thanks for your PR! The first commit seems largely what we discussed in depth at PR #8414 so that looks good to me already. I have a question about how `VsockDevice` would be used after its introduction into `DeviceManager`.
> 
> So far, the QEMU driver (in PR #8185) knew it would need a `VsockDevice` so it created one. With `VsockDevice` handled by `DeviceManager`, that wouldn't seem to make much sense anymore, but who and when will create it now?
>
@pmores Good point！I believe that the vsock config should be configured at the same time as the hvsock config, or in other words, they should both be in the prepare_for_start_sandbox function. 
However, the existing implementation does not take vsock into account. Therefore, I added the `prepare_vm_socket_config` function to call the hypervisor trait `get_agent_socket` to return socket address with coresponding URI scheme to determine whether it is vsock or hvsock, and then initialize the corresponding config in `prepare_for_start_sandbox`. And next, it walks the `Device Manager::do_handle_device` in `prepare_before_start_vm` and finally call the Qemu's add_device to add vsock device.

Apokleos:
> Hi @Apokleos ! Thanks for submitting this PR. It looks quite good and we should be able to merge it soon.
> 
> A few general comments :
> 
> * "[runtime-rs: move vosck device into device manager](https://github.com/kata-containers/kata-containers/pull/8516/commits/296911768954774fb62456370a07b62e96e14156) doesn't seem to do anything with the device manager but [runtime-rs: add vsock device in device manager](https://github.com/kata-containers/kata-containers/pull/8516/commits/c1f2bd787894e8880a75d6ddc78fad36f9d17cb6) does... I suspect the first commit should have another title like "Add vsock to the DeviceType enum".
> * please split out the introduction of `generate_vhost_sock_cid()` to an initial commit
> * please populate the commit logs 😄
> 
> Cheers.
>
Thx @gkurz you're right. Fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
```diff
@@ -121,46 +118,85 @@ const CID_RETRY_COUNT: u32 = 50;
 
 impl VsockDevice {
     pub async fn new(id: String) -> Result<Self> {
-        let vhost_fd = OpenOptions::new()
-            .read(true)
-            .write(true)
-            .open(VHOST_VSOCK_DEVICE)
+        let (guest_cid, _file) = generate_vhost_sock_cid().await?;
```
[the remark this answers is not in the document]
Apokleos: Thx, Fixed

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
```diff
@@ -121,46 +118,87 @@ const CID_RETRY_COUNT: u32 = 50;
...
+    }
+
+    async fn decrease_attach_count(&mut self) -> Result<bool> {
+        // vsock devices will not be detached multiple times, Just return Ok(false)
+
+        Ok(false)
+    }
+}
+
+pub async fn generate_vhost_sock_cid() -> Result<(u32, File)> {
```
studychao, line 169: should this be `generate_vhost_vsock_cid `

Apokleos: Thx @studychao Fixed!

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -223,6 +227,43 @@ impl VirtSandbox {
         })
     }
 
+    async fn prepare_vm_socket_config(&self) -> Result<ResourceConfig> {
```
gkurz, line 230: Similar comment as suggested in your initial proposal : please introduce this function with no functional change in a preliminary commit and only bring support for legacy vsock in a second commit. Separating refactoring changes from functional changes is a good practice that makes review and maintenance a lot easier.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -223,6 +227,43 @@ impl VirtSandbox {
...
+        // There're two kinds of vm sockets:
+        // - Hybrid Vsock: hvsock://<path>:<port>.
+        // -- Example: "hvsock:///x/y/z/kata.hvsock:port"
+        // - Vsock: vsock://<cid>:<port>
+        // -- Example: "vsock://31513974:1024"
+        let agent_socket = self
+            .hypervisor
+            .get_agent_socket()
+            .await
+            .context("get agent socket")?;
```
pmores, line 241: I have some thoughts here.  `get_agent_socket()` returns a `Result` which implies it is designed as fallible, it can fail.  However, called from this place, it will by definition not yet be able to return its full intended return value (`<scheme>://<cid>`).  It will be reduced to returning just the scheme (likely as `vsock://VMADDR_CID_ANY`).

Now, if returning just the scheme with an invalid authority part of the URL is valid then it's hard to imagine under which circumstances `get_agent_socket()` could ever fail, so returning a `Result` is becomes kind of pointless.  Note however that this change doesn't really introduce anything new in this respect, it just confirms existing practice where both Dragonball's and CH's `get_agent_socket()` implementations indeed never fail.

I would really like to hear from the original `Hypervisor` trait designers to learn under which circumstances `get_agent_socket()` was supposed to fail.  If anybody knows who to tag please do so as I'm not sure (@bergwolf ?).  Also, would there be other ways to consider how to query a `Hypervisor` for the vsock flavour it prefers?  Could e.g. `Capabilities` be used for this?  (I'm not familiar enough with `Capabilities` semantics to judge if vsock would be appropriate there.)

At any rate, I don't consider this a flaw of this PR and don't think it should block a potential merge. Just sayin'. ;-)  It's great that this functionality is in its own function `prepare_vm_socket_config()` so any future refactor would be easier.

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -223,6 +227,43 @@ impl VirtSandbox {
...
+
+        let uri_fields: Vec<&str> = agent_socket.split("://").collect();
+        if uri_fields.len() != 2 {
+            return Err(anyhow!("invalid URI"));
+        }
+
+        // Generate device configs for vm sockets depending on the URI scheme.
+        let uri_scheme = uri_fields[0].to_uppercase();
```
pmores, line 249: It might make sense to define the values of `SCHEME_VSOCK` and `SCHEME_HYBRID_VSOCK` in lowercase in the first place.  Then there would be no need to uppercase the scheme here, right?

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
pmores: In my view, introducing just a single caps bit seems better.  To my knowledge, there's no other vsock flavour in sight apart from plain vsock and hybrid vsock and even if another one appears in the future we could always add a bit for it when that happens.

As to whether that single caps bit should represent vsock or hvsock, I think that's kind of arbitrary since there doesn't seem to be any hierarchical ordering between vsock and hvsock in terms of superiority or support.  Based on existing practice I'd lean towards hvsock.

So I guess I'm basically in favour of the former variant that you suggest.

pmores (approved):
@Apokleos lgtm, thanks Alex! This PR was very pleasant to work with and review. :+1:

Issue #8525, opened 2023-11-29
gha: building kata artefacts fails on ppc64le


Amulyam24:
```
Run make "${KATA_ASSET}-tarball"
utils.mk:150: "WARNING: powerpc64le-unknown-linux-musl target is unavailable"
make kernel-tarball-build
make[1]: Entering directory '/home/ubuntu/runner/_layout/_work/kata-containers/kata-containers'
utils.mk:150: "WARNING: powerpc64le-unknown-linux-musl target is unavailable"
/home/ubuntu/runner/_layout/_work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-copy-yq-installer.sh
/home/ubuntu/runner/_layout/_work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-binaries-in-docker.sh  --build=kernel
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
956774821c9c09eaf21ad0c4866a6c98246eea3c8d23f801e61ca6067e9d25ed
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Build kata version 3.3.0-alpha0: kernel
mkdir: cannot create directory '/home/ubuntu/runner/_layout/_work/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build': Permission denied
make[1]: *** [tools/packaging/kata-deploy/local-build/Makefile:54: kernel-tarball-build] Error 1
make[1]: Leaving directory '/home/ubuntu/runner/_layout/_work/kata-containers/kata-containers'
make: *** [tools/packaging/kata-deploy/local-build/Makefile:90: kernel-tarball] Error 2
```

PR #8526, opened 2023-11-29
gha: fix artefacts build on ppc64le

Amulyam24:
Add step in the right place to prepare the runner for the builds/tests.

Fixes: #8525

### DIFF
--- a/.github/workflows/build-kata-static-tarball-ppc64le.yaml
+++ b/.github/workflows/build-kata-static-tarball-ppc64le.yaml
@@ -51,6 +51,9 @@ jobs:
         with:
           ref: ${{ inputs.commit-hash }}
           fetch-depth: 0 # This is needed in order to keep the commit ids history
+      
+      - name: Prepare the self-hosted runner
+        run: ${HOME}/scripts/prepare_runner.sh
 
       - name: Rebase atop of the latest target branch
         run: |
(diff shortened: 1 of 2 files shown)


stevenhorsman (approved):
LGTM

.github/workflows/build-kata-static-tarball-ppc64le.yaml
```diff
@@ -38,6 +38,9 @@ jobs:
       - name: Adjust a permission for repo
         run: |
           sudo chown -R $USER:$USER $GITHUB_WORKSPACE
+      
+      - name: Prepare the self-hosted runner
```
BbolroC, line 42: I would be better to place this step after the checkout, just opening up a room to make an urgent patch against the repo. What do you think?

BbolroC (approved):
Other than the suggestion for the location of the step, it looks good to me. Thanks @Amulyam24 !

.github/workflows/build-kata-static-tarball-ppc64le.yaml
Amulyam24: > just opening up a room to make an urgent patch against the repo.

Hi @BbolroC, I don't mind moving it to after checkout but can I ask what you mean by this? I didn't get you completely

GabyCT (approved):

.github/workflows/build-kata-static-tarball-ppc64le.yaml
BbolroC: Let's say what if you need to do something against the source code before the real steps start? (I know it is very unlikely to happen). I just wanted to inform you of this kind of situation in which I experienced before.

Amulyam24: Got it, makes sense. Changes applied, Thanks @BbolroC !

Issue #6797, opened 2023-05-10
Consider adding the rust log-parser to kata-ctl


jodh-intel:
Now that #6512 has landed, we could consider adding the code to [`kata-ctl`](https://github.com/kata-containers/kata-containers/tree/main/src/tools/kata-ctl) as a sub-command (`kata-ctl log-parser ...`).

/cc @gabevenberg in case he's interested.

gabevenberg:
Ill take this one.

gabevenberg:
reopening, its not closed till #6826 is merged

PR #6826, opened 2023-05-13
kata-ctl: Moved log-parser-rs into kata-ctl

gabevenberg:
Log-parser-rs was always intended to become a sub-functionality of kata-ctl, but it was useful to develop it and initaly merge it as a standalone program, and migrate it to a subcommand later.

Fixes #6797

PR #6282, opened 2023-02-14
# Kata Containers 3.0.2

amshinde:
- stable-3.0: Stable 3.0 backports
- stable-3.0 | docs: Fix missing critical steps in how-to-hotplug-memory-arm64.md
- Stable-3.0 | Upgrade to Cloud Hypervisor v28.2
- Qemu logs for stable 3.0
- Backport CI fixes for s390x and ppc64le to stable-3.0
- docs: Fix missing critical steps in how-to-hotplug-memory-arm64.md
- Stable-3.0 | Upgrade to Cloud Hypervisor v28.1

4ebeb51bb release: Adapt kata-deploy for 3.0.2
178ee3d7e agent: check command before do test_ip_tables
7461bcd76 runtime-rs: change cache mode
123c86717 SEV: Update ReducedPhysBits
98f60c100 clh: Enforce API timeout only for vm.boot request
960f089d3 virtiofsd: fix the build on ppc64le
92f3b11c9 runtime:all APIs are hang in the service.mu
92619c833 runtime: Drop QEMU log file support
4f3db7678 runtime: Collect QEMU's stderr
918c11e46 runtime: Start QEMU undaemonized
8c4507be2 runtime: Launch QEMU with cmd.Start()
a61fba6d4 runtime: Pre-establish the QMP connection
ad9cb0ba5 govmm: Optionally pass QMP listener to QEMU
d6dd99e98 govmm: Optionally start QMP with a pre-configured connection
0623f1fe6 virtiofsd: Not use "link-self-contained=yes" on s390x
5883dc1bd CI: Set docker version to v20.10 in ubuntu:20.04 for s390x|ppc64le
4a5877f45 docs: Fix missing critical steps in how-to-hotplug-memory-arm64.md
d3b57325e versions: Upgrade to Cloud Hypervisor v28.2
0d7bd066d docs: Fix missing critical steps in how-to-hotplug-memory-arm64.md
ac1ce2d30 docs: Fix missing critical steps in how-to-hotplug-memory-arm64.md
f4d71af45 docs: Fix missing critical steps in how-to-hotplug-memory-arm64.md
fcc120d49 versions: Upgrade to Cloud Hypervisor v28.1

PR #8528, opened 2023-11-29
tools: Stop building / shipping log-parser-rs

fidencio:
This is a commit that's a pre-req for #6826, as that PR will merge log-parser-rs into kata-ctl, but that will result in a CI breakage.

So, let's deal with the CI changes here, thanks to GHA and our favourite `pull_request_target` event, unblocking that PR to be merged.

Fixes: #6797 (not really, but related).


gkurz (approved):
Please reference PR #6826 instead of #6282 in the changelog. Apart from that, rest LGTM. Thanks @fidencio !

GabyCT (approved):

PR #8545, opened 2023-12-01
DON'T MERGE:GHA: enable static check on arm64

jongwu:


jongwu:
Great! Thanks for your kindness @BbolroC

Issue #8546, opened 2023-12-01
libs: logging: Fix logger


jodh-intel:
It appears that #8311 inadvertently broke the logging: with that PR, no messages below `Info` are logged.

PR #8547, opened 2023-12-01
libs:logging: Fix logger

jodh-intel:
PR #8311 inadvertently broke the logging since no log messages below the `Info` level are logged now, regardless of the requested log level.

Resolve the issue by storing the requested log level in the `RuntimeComponentLevelFilter` and using that level in the `log()` function, rather than hard-coding `Info` as the default where no entry is found in the `FILTER_RULE` hashmap.

Fixes: #8546.

### DIFF
--- a/src/libs/logging/src/lib.rs
+++ b/src/libs/logging/src/lib.rs
@@ -68,7 +68,7 @@ pub fn create_term_logger(level: slog::Level) -> (slog::Logger, slog_async::Asyn
     });
 
     // Allow runtime filtering of records by log level
-    let filter_drain = RuntimeComponentLevelFilter::new(unique_drain).fuse();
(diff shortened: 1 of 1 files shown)


jodh-intel:
It's worth noting that the logger tests were failing. We need to find a way to re-introduce the changes I added on #2970 in a reliable way. Part of the problem is that our build system doesn't currently distinguish between golang and rust components. If it did, we could make all rust components depend on running the logger test suite before building the component. But that's not reliable as developers can in some instances just run `cargo build` which bypasses that.

Another approach would be to have a rust "build script" that we add to all rust components to force-run the logger tests. That would involve duplication though.

jodh-intel:
This PR will need to go into the next release.

jodh-intel:
#8554 raised to address the issue of re-enabling the feature where we always run the logger tests (_always_, I say! ;)

Issue #8548, opened 2023-12-01
libs: x86_64: Don't require root to query GuestProtection


jodh-intel:
The current `available_guest_protection()` implementation for `x86_64` requires root privs. However, this is no longer true, so drop the requirement.

PR #8549, opened 2023-12-01
libs: protection: x86_64: drop root requirement for querying

jodh-intel:
It is no longer necessary to be `root` to query the guest protection
(TDX) on `x86_64` systems, so drop the requirement.

Also update `kata-ctl` to remove the duplicate, and now redundant, root check.

Fixes: #8548.

### DIFF
--- a/src/libs/kata-sys-util/src/protection.rs
+++ b/src/libs/kata-sys-util/src/protection.rs
@@ -13,7 +13,7 @@ use std::path::Path;
 use std::path::PathBuf;
 use thiserror::Error;
 
-#[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "s390x", target_arch = "powerpc64le"))]
 use nix::unistd::Uid;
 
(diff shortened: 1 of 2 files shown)


src/libs/kata-sys-util/src/protection.rs
```diff
@@ -13,7 +13,7 @@ use std::path::Path;
 use std::path::PathBuf;
 use thiserror::Error;
 
-#[cfg(any(target_arch = "s390x", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "s390x", target_arch = "powerpc64le"))]
```
fidencio, line 16: Out of curiosity, why powerpc64le was added here?

It seems not related to this commit.

jodh-intel: I've added a comment to the commit. Basically PPC64 needs this import fwics but it was previously missing.

amshinde (approved):

cmaf (approved):
@jodh-intel LGTM.

PR #8576, opened 2023-12-05
runtime, factory: fix resetHypervisorConfig

### LINKED ISSUE #8488: Commit 57d4dd8 breaks vmtemplate / vmcache
With commit 57d4dd8 new parameters in the hypervisor configs are added, namely `SanboxName` and `SandboxNamespace`. This makes the [config comparison between the cached VM config and the one requested by the pod](https://github.com/kata-containers/kata-containers/blob/45a41c34319456330e4f31f13a95dbfcb856776d/src/runtime/virtcontainers/factory/factory_linux.go#L95) to always return false, and so the cached VM is never used.

Simple fix would be to reset those new values before config comparison:

```
diff --git a/src/runtime/virtcontainers/factory/factory_linux.go b/src/runtime/virtcontainers/factory/factory_linux.go
index 4a0cfcfe3..e2d28696a 100644
--- a/src/runtime/virtcontainers/factory/factory_linux.go
+++ b/src/runtime/virtcontainers/factory/factory_linux.go
@@ -80,6 +80,8 @@ func resetHypervisorConfig(config *vc.VMConfig) {
        config.HypervisorConfig.SharedPath = ""
        config.HypervisorConfig.VMStorePath = ""
        config.HypervisorConfig.RunStorePath = ""
+       config.HypervisorConfig.SandboxName = ""
+       config.HypervisorConfig.SandboxNamespace = ""
 }
```

Note that VMcache is anyway broken until #6981 gets merged.

@stevenhorsman

ruocco:
With the support for the remote hypervisor type, new parameters in the hypervisor configs are added, namely SanboxName and SandboxNamespace. This makes the config comparison between the cached VM config and the one requested by the pod to always return false, and so the cached VM is never used.

This PR fixes this behavior resetting those parameter to an empty string for the comparison.

fixes #8488

### DIFF
--- a/src/runtime/virtcontainers/factory/factory_linux.go
+++ b/src/runtime/virtcontainers/factory/factory_linux.go
@@ -80,6 +80,8 @@ func resetHypervisorConfig(config *vc.VMConfig) {
 	config.HypervisorConfig.SharedPath = ""
 	config.HypervisorConfig.VMStorePath = ""
 	config.HypervisorConfig.RunStorePath = ""
+	config.HypervisorConfig.SandboxName = ""
+	config.HypervisorConfig.SandboxNamespace = ""
 }
 
 // It's important that baseConfig and newConfig are passed by value!


fidencio (approved):
lgtm, thanks @ruocco!

Issue #8579, opened 2023-12-05
Runtime-rs + clh does not work with nerdctl and docker


amshinde:
Nerdctl and docker run hooks for network setup after the sandbox is started. This requires runtime to implement network hotlug to support network attach for a sandbox. As network hotplug is not currently impemented with clh, runtime-rs + clh does not work with nerdctl/docker.

PR #8580, opened 2023-12-05
runtime-rs: add network hotplug for clh

amshinde:
This is required for clh to work with nerdtcl and docker.
This fixes the issues seen with nerdctl while starting a container.
 Hoewever, container exit with docker is still broken due to an unrelated
 issue.
    
 Fixes: #8579


amshinde:
@fidencio I have tested this PR with nerdctl and docker. This should fix the network issues seen earlier.
However I am seeing another issue with docker, the container is not able to exit while running both an interactive and non-interactive container. From the logs, I see the last log to be about executing the Poststop hooks, but nothing beyond that. 
In any case, this looks like a separate issue not related to network that needs further debugging.
I will open a separate issue for that. Meanwhile would appreciate it if you can take a look at this PR and do a quick test on your end as well.

cmaf (approved):
@amshinde LGTM

GabyCT (approved):

fidencio (approved):
@amshinde, I've tested this both with nerdctl and docker, and I could replicate exactly the same results as you did.

lgtm, thanks!

amshinde:
> @amshinde, I've tested this both with nerdctl and docker, and I could replicate exactly the same results as you did.

Thanks for testing this out.

Issue #8553, opened 2023-12-04
CI: Make URL checker use a list of user agents


jodh-intel:
See discussion on https://github.com/kata-containers/kata-containers/pull/6826, namely https://github.com/kata-containers/kata-containers/pull/6826#issuecomment-1836427181.

amshinde:
@jodh-intel  Are you taking a look at this one? Till this is implemented, can we consider adding 403 to the http_statuses to ignore in order to unblock PR 6826?

PR #8592, opened 2023-12-06
CI: static-checks: Try multiple user agents 

jodh-intel:
Make the URL checker cycle through a list of user agent values until we
hit one the remote server is happy with.

This is required since, unfortunately, we really, really want to check
these URLs, but some sites block clients based on their `User-Agent`
(UA) request header value. And of course, each site is different and can
change its behaviour at any time.

Our strategy therefore is to try various UA's until we find one the
server accepts:

- No explicit UA (use `curl`'s default)
- Explicitly no UA.
- A blank UA.
- Partial UA values for various CLI tools.
- Partial UA values for various console web browsers.
- Partial UA for Emacs's built-in browser.
- The existing UA which is used as a "last ditch" attempt where the UA implies multiple platforms and browser.

> **Notes:**
>
> - The "partial UA" values specify specify the UA "product" but not the
>   UA "product version": we specify `foo` and not `foo/1.2.3`). We do
>   this since most sites tested appear to not care about the version.
>   This is as expected given that the version is strictly optional (see `[*]`).
>
> - We now treat URLs that the server reports as HTTP 401, HTTP 402 or
>   HTTP 403 as *valid*. See the comments in the code.
>
> - We now log all errors and display a summary on error in addition to
>   the simple list of the URLs we believe to be invalid. This should make
>   future debugging simpler.

`[*]` - https://www.rfc-editor.org/rfc/rfc9110#section-10.1.5

Fixes: #8553.

### DIFF
--- a/tests/static-checks.sh
+++ b/tests/static-checks.sh
@@ -500,13 +500,16 @@ check_url()
 	local invalid_urls_dir="${2:-}"
 	[ -n "$invalid_urls_dir" ] || die "need invalid URLs directory"
 
-	local curl_out=$(mktemp)
+	local curl_out
+	curl_out=$(mktemp)
+
 	files_to_remove+=("${curl_out}")
 
 	info "Checking URL $url"
 
 	# Process specific file to avoid out-of-order writes
-	local invalid_file=$(printf "%s/%d" "$invalid_urls_dir" "$$")
+	local invalid_file
+	invalid_file=$(printf "%s/%d" "$invalid_urls_dir" "$$")
 
 	local ret
 	local user_agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36"
--- a/tests/static-checks.sh
+++ b/tests/static-checks.sh
@@ -811,7 +811,10 @@ static_check_docs()
 
 	popd
 
-	[ $docs_failed -eq 0 ] || die "spell check failed, See https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#spelling for more information."
+	[ $docs_failed -eq 0 ] || {
+        url='https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#spelling'
+        die "spell check failed, See $url for more information."
+    }
 }
 
 static_check_eof()
--- a/tests/static-checks.sh
+++ b/tests/static-checks.sh
@@ -539,58 +539,117 @@ check_url()
 
 	files_to_remove+=("${curl_out}")
 
-	info "Checking URL $url"
-
 	# Process specific file to avoid out-of-order writes
 	local invalid_file
 	invalid_file=$(printf "%s/%d" "$invalid_urls_dir" "$$")
 
 	local ret
-	local user_agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36"
 
-	local curl_ua_args
-	curl_ua_args="-A '$user_agent'"
+	local -a errors=()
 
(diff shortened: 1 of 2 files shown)


cmaf (approved):
@jodh-intel thanks for fixing this! I confirmed it works locally with the exception of https://developer.amd.com/sev/. 

I am going to port 2 commits to the tests repo, the curl and multiple user agents one (https://github.com/kata-containers/kata-containers/pull/8592/commits/d20822c6e2b28809b8b51efaff50ceccb89e630b and https://github.com/kata-containers/kata-containers/pull/8592/commits/404162c7aa7f5bf2f538be1110da79926724dcea). 

**Note**: We will not be maintaining 2 separate static checks scripts beyond these changes because it will be running from GHA after #8595. However, I am porting them in favor of getting https://github.com/kata-containers/kata-containers/pull/6826 merged and in case testing takes longer than expected.

tests/static-checks.sh
```diff
@@ -500,13 +500,16 @@ check_url()
 	local invalid_urls_dir="${2:-}"
 	[ -n "$invalid_urls_dir" ] || die "need invalid URLs directory"
 
-	local curl_out=$(mktemp)
+	local curl_out
```
amshinde, line 537: @jodh-intel Is there a benefit for doing it this way..declaring and then defining later?

tests/static-checks.sh
```diff
@@ -546,51 +546,133 @@ check_url()
...
+			#   which get upset by "curl -L" when the
+			#   redirection is not required.
+			#
+			# Anything else is considered an error.
+			#
+			# See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
+
+			{ grep -qE "^(1[0-9][0-9]|2[0-9][0-9]|3[0-9][0-9]|405)" <<< "$status"; ret=$?; } || true
+
+			[ "$ret" -eq 0 ] && continue
```
amshinde, line 637: Why do we continue here? We have got a sucessful code, shouldnt we just break or return success here?

tests/static-checks.sh
jodh-intel: Yep - see https://www.shellcheck.net/wiki/SC2155.

tests/static-checks.sh
jodh-intel: No, we shouldn't break/return here since getting a URL can result in a list of status codes. For example, we might get these:

```
HTTP/2 301    # (Moved Permanently): No problem here - the server is telling us we're being redirected
HTTP/2 200    # (OK): Great, we got the doc
```

But we might instead get this:

```
HTTP/2 301    # (Moved Permanently): No problem here - the server is telling us we're being redirected
HTTP/2 404    # (Not Found): URL is invalid
```

So if we accept the first non-error HTTP code only, we won't detect invalid URLs so we must iterate through all of them.

tests/static-checks.sh
```diff
@@ -546,51 +546,133 @@ check_url()
...
+					#
+					# 401: Unauthorized (need to login to the site).
+					# 402: Payment Required.
+					# 403: Forbidden (possibley the same as 401).
+					#
+					# If they did not exist, we'd get a 404, so since
+					# they have been "semi validated" by the server by
+					# simply returning these codes, we'll assume they
+					# are valid.
+					401|402|403) success='true' ;;
```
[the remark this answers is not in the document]
jodh-intel: The assumption here is that if the document wasn't valid, the server would return 404 (Not Found). Instead, it's basically saying `EPERM` so there doesn't seem much point in trying the other UAs as that would imply they might be able to bypass the site security.

Looking at is another way, the static checker needs to do something with this awkward URL. We can't verify it fully as we don't have the credentials to do so. Hence, we either error (which would be annoying as any PR that modified a doc mentioning these protected URLs would be guaranteed to fail the CI checks, or we allow it. I guess we could add a warning here as that would at least allow someone to check the URL manually?

gkurz (approved):
> branch updated (and simplified) - ptal! ;)
> 

Simpler indeed. :-) Thanks @jodh-intel !

Issue #8598, opened 2023-12-07
Dragonball: cargo fmt not enabling with "--all" options


studychao:
Right now, cargo fmt check in Dragonball only test with the default features but not all features. This will cause some code being untested by the fmt tool.

PR #8599, opened 2023-12-07
dragonball: add --all for fmt ci

studychao:
Right now, cargo fmt check in Dragonball only test with the default features but not all features. This will cause some code being untested by the fmt tool.

This PR adds --all option for the Dragonball CI and also fix some code that forgets to do cargo fmt --all.

fixes: #8598


liubogithub (approved):

justxuewei (approved):
Lgtm, thanks!

Issue #8600, opened 2023-12-07
GHA: Rebasing error affects subsequent workflows on a self-hosted runner


BbolroC:
# Description of problem

When a pull request workflow fails due to a [merge conflict]( https://github.com/kata-containers/kata-containers/blob/c14e3096c83615e8368fd6d8bd10e251a1da00dd/tests/git-helper.sh#L24C15-L24C15) like:

```
Run ./tests/git-helper.sh "rebase-atop-of-the-latest-target-branch"
Adding user name and email to the local git repo
Rebasing atop of the latest main
Auto-merging src/libs/kata-types/src/capabilities.rs
CONFLICT (content): Merge conflict in src/libs/kata-types/src/capabilities.rs
Auto-merging src/runtime-rs/Cargo.lock
Auto-merging src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
Auto-merging src/runtime-rs/crates/hypervisor/src/ch/mod.rs
Auto-merging src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
Auto-merging src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
CONFLICT (content): Merge conflict in src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
Auto-merging src/runtime-rs/crates/hypervisor/src/lib.rs
Auto-merging src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
Auto-merging src/runtime-rs/crates/hypervisor/src/qemu/mod.rs
Auto-merging src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
Rebasing (1/5)
error: could not apply 20d[6](https://github.com/kata-containers/kata-containers/actions/runs/7110651490/job/19357439752?pr=6876#step:4:7)0[8](https://github.com/kata-containers/kata-containers/actions/runs/7110651490/job/19357439752?pr=6876#step:4:9)e33... runtim-rs: get guest memory details
hint: Resolve all conflicts manually, mark them as resolved with
hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
hint: You can instead skip this commit: run "git rebase --skip".
hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply [20](https://github.com/kata-containers/kata-containers/actions/runs/7110651490/job/19357439752?pr=6876#step:4:21)d608e33... runtim-rs: get guest memory details
[...]

PR #8601, opened 2023-12-07
GHA: remove GITHUB_WORKSPACE when workflow fails due to merge conflict

BbolroC:
It is to remove a GITHUB_WORKSPACE directory for self-hosted runners when a workflow fails due to the merge conflict. This will prevent the subsequent workflows from getting stuck in the same situation.

Fixes: #8600


Amulyam24 (approved):
LGTM, Thanks @BbolroC!

Issue #8608, opened 2023-12-08
dragonball: Use vhost-net device by default


justxuewei:
Dragonball config has an item named `disable_vhost_net`. But it is never used since vhost-net is not available until https://github.com/kata-containers/kata-containers/pull/7675 merged.

PR #8609, opened 2023-12-08
dragonball: Use vhost-net device by default

justxuewei:
This patch set vhost-net as default backend of networking. It allows users
to set `disable_vhost_net` to `true` to reenable virtio-net backend.
Plus, which backend to use is a matter of hypervisor, runtime-rs will no
longer need to know that.

Fixes: #8608

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
@@ -12,7 +12,7 @@ use dragonball::api::v1::{
 };
 use dragonball::device_manager::blk_dev_mgr::BlockDeviceType;
 
-use super::DragonballInner;
+use super::{build_dragonball_network_config, DragonballInner};
 use crate::{
     device::DeviceType, HybridVsockConfig, NetworkConfig, ShareFsConfig, ShareFsMountConfig,
     ShareFsMountOperation, ShareFsMountType, VfioBusMode, VfioDevice, VmmState, JAILER_ROOT,
@@ -210,8 +210,9 @@ impl DragonballInner {
     }
 
     fn add_net_device(&mut self, config: &NetworkConfig) -> Result<()> {
+        let net_cfg = build_dragonball_network_config(&self.config, config);
         self.vmm_instance
-            .insert_network_device(config.into())
+            .insert_network_device(net_cfg)
             .context("insert network device")
     }
 
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
@@ -26,7 +26,7 @@ use kata_types::config::hypervisor::Hypervisor as HypervisorConfig;
 use tokio::sync::RwLock;
 use tracing::instrument;
 
-use crate::{Backend, DeviceType, Hypervisor, NetworkConfig, VcpuThreadIds};
+use crate::{DeviceType, Hypervisor, NetworkConfig, VcpuThreadIds};
 
 pub struct Dragonball {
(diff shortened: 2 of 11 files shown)


Apokleos (commented):
Thx @justxuewei for your work!
But IMO you should split the PR into at least two commits, one for dragonball and one for runtime-rs, rather than having one PR which has only one commit with several changes at different levels.

src/dragonball/src/dbs_virtio_devices/src/vhost/vhost_kern/net.rs
```diff
@@ -290,7 +290,9 @@ where
                 "{}: Invalid virtio queue pairs, expected a value greater than 0, but got {}",
                 NET_DRIVER_NAME, self.vq_pairs
             );
-            return Err(VirtioError::ActivateError(Box::new(ActivateError::InvalidParam)));
+            return Err(VirtioError::ActivateError(Box::new(
```
studychao, line 293: need a rebase here when the format commit from Dragonball is merged.

justxuewei:
> Thx @justxuewei for your work! But IMO you should split the PR into at least two commits, one for dragonball and one for runtime-rs, rather than having one PR which has only one commit with several changes at different levels.

@Apokleos Thanks for your kindly review. I've split this patch into two commits, one for runtime-rs, another one for dragonball. PTAL.

studychao (approved):

src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
```diff
@@ -209,9 +210,14 @@ impl DragonballInner {
         Ok(())
     }
 
-    fn add_net_device(&mut self, config: &NetworkConfig) -> Result<()> {
+    fn add_net_device(
+        &mut self,
+        hconfig: &HypervisorConfig,
```
Apokleos, line 215: IMO, there's no need introduce a hconfig here, you can just get the hypervisor config in add_net_device before you do dragonball_network_device

Apokleos, line 215: IMO, there's no need introduce a hconfig here, you can just get the hypervisor config in add_net_device before you do dragonball_network_device

src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
```diff
@@ -201,41 +201,39 @@ impl Persist for Dragonball {
...
-            guest_mac: value.guest_mac.clone().map(|mac| {
-                // We are safety since mac address is checked by endpoints.
-                DragonballMacAddr::from_bytes(&mac.0).unwrap()
-            }),
-            use_shared_irq: value.use_shared_irq,
-            use_generic_irq: value.use_generic_irq,
-        }
+/// Generate Dragonball network config according to hypervisor config and
+/// runtime network config.
+pub(crate) fn dragonball_network_config(
```
Apokleos, line 206: The function name ` dragonball_network_config` is not clear. I think it'd better to change it to something that is more specific and easier to understand, such as "build_xxx", "setup_xxx", "init_xxx"...

Apokleos (commented):
Thx @justxuewei for your work! I have some comments.

src/runtime-rs/crates/hypervisor/src/dragonball/inner_device.rs
justxuewei: Fixed.

src/runtime-rs/crates/hypervisor/src/dragonball/mod.rs
justxuewei: Fixed.

Apokleos (approved):
Thx @justxuewei Nice work!
LGTM !

Issue #8619, opened 2023-12-10
runtime-rs: enhancement of DirectVolume when using CSI


Apokleos:
**Which feature do you think can be improved?**
DirectVolume structure in runtime-rs is different from it in kata-runtime, which causes they has no unified handling method for [DirectVolumeMountInfo](https://github.com/kata-containers/kata-containers/blob/1c42d945503f2d98d30e784bf8d2f0b7555618b9/src/libs/kata-types/src/mount.rs#L95) and [MountInfo](https://github.com/kata-containers/kata-containers/blob/1c42d945503f2d98d30e784bf8d2f0b7555618b9/src/runtime/pkg/direct-volume/utils.go#L43).

In other words, the two direct volume structs has different json description for `fs Type` and `volume Type`

- direct volume in Kata runtime

```json
{
  "volume-type": "directvol",
  "device": "rawdisk.1g",
  "fstype": "ext4",
  "metadata": {},
  "options": []
}
```
- direct volume in runtime-rs

```json
{
  "volume_type": "directvol",
  "device": "rawdisk.1g",
  "fs_type": "ext4",
  "metadata": {},
  "options": []
}
```

We should align the two by simply adding the attribute #[serde(rename = “xx”) to each field in DirectVolumeMountInfo as below:

```
pub struct DirectVolumeMountInfo {
    /// The type of the volume (ie. block)
    #[serde(rename = "volume-type")]
    pub volume_type: String,
    /// The device backing the volume.
    #[serde(rename = "device")]
    pub device: String,
    /// The filesystem type to be mounted on the volume.
    #[serde(rename = "fstype")]
    pub fs_type: String,
    /// Additional metadata to pass to the agent regarding this volume.
    #[serde(rename = "metadata", skip_serializing_if = "HashMap::is_empty")]
    pub metadata: HashMap<String, String>,
    /// Additional mount options.
    #[serde(rename = "options", skip_serializing_if = "Vec::is_empty")]
    pub options: Vec<String>,
}
```

PR #8620, opened 2023-12-10
runtime-rs: Enhancement of DirectVolume when using a dedicated CSI

Apokleos:
(1) Enhancement of DirectVolume when using CSI.
  We use a matching direct-volume path to determine whether an OCI mount
  is a DirectVolume. However, we should handle the case where no match is
  found appropriately.
  This error will be defined as a non-DirectVolume type when judging the
  OCI mount but not failed.

(2) Add attribe serde rename to each field of DirectVolume.
  DirectVolume structure in runtime-rs is different from it in kata-runtime,
  which causes they has no unified handling method for DirectVolumeMountInfo
  and MountInfo.
  
  We should align the two by simply adding the attribute #[serde(rename="x")
  to each field in DirectVolumeMountInfo
  
  Fixes: #8619
    
### DIFF
--- a/src/libs/kata-types/src/mount.rs
+++ b/src/libs/kata-types/src/mount.rs
@@ -94,14 +94,18 @@ impl Mount {
 #[derive(Debug, Clone, Eq, PartialEq, Default, Serialize, Deserialize)]
 pub struct DirectVolumeMountInfo {
     /// The type of the volume (ie. block)
+    #[serde(rename = "volume-type")]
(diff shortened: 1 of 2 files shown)


Tim-0731-Hzt (approved):

Tim-Zhang (approved):
LGTM, thanks @Apokleos

quanweiZhou (approved):
Thanks @Apokleos, LGTM.

Issue #8621, opened 2023-12-11
GHA: k3s kubectl version --short broken


BbolroC:
# Description of problem

At the moment, all `run-k8s-tests` tests using `k3s` are failing (e.g. https://github.com/kata-containers/kata-containers/actions/runs/7162079316/job/19498931906?pr=8609) due to the following:

```
$ /usr/local/bin/k3s kubectl version --short
error: unknown flag: --short
See 'kubectl version --help' for usage.
```

Looking at the usage:

```
$ /usr/local/bin/k3s kubectl version --help
Print the client and server version information for the current context.

Examples:
  # Print the client and server versions for the current context
  kubectl version

Options:
    --client=false:
        If true, shows client version only (no server required).

    -o, --output='':
        One of 'yaml' or 'json'.

Usage:
  kubectl version [flags] [options]

Use "kubectl options" for a list of global command-line options (applies to all commands).
```

The option `--short` should be updated to `--client=true`.

PR #8622, opened 2023-12-11
GHA: Use --client=true for k3s kubectl version

BbolroC:
This is to fix a broken usage for `k3s kubectl version` by switching an option `--short` to `--client=true`.

Fixes: #8621

### DIFF
--- a/tests/gha-run-k8s-common.sh
+++ b/tests/gha-run-k8s-common.sh
@@ -161,7 +161,7 @@ function deploy_k3s() {
 	if [ "${ARCH}" = "x86_64" ]; then
 		ARCH=amd64
 	fi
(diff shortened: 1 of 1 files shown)


studychao (approved):
thanks for the quick fix!

justxuewei (approved):
Lgtm, thanks!

gkurz (approved):
>  GHA: Use --client=true for k3s kubectl version
This is to fix a broken usage for `k3s kubectl version` by switching
an option `--short` to `--client=true`.

The title and comment barely paraphrase the code. They even give the impression that `--client=true` is the actual fix, while it is not : this is _just_ to skip and uneeded connection to the server since we're only interested in the client version.

The real fix is the removal of `--short`, which was dropped in kubernetes 1.28 : short format is now the default output of `kubectl version`.

Not worth holding the PR though, so I'm approving.

Issue #8629, opened 2023-12-11
runtime-rs: ch: Change state when VM stopped


jodh-intel:
The Cloud Hypervisor driver is not changing the state of the VM on VM shutdown.

Currently, although the container stops, the `stop_vm()` method gets called twice if the workload exits. The first time the call succeeds, but the second time it fails with `EPIPE` ("`Error writing to or reading from HTTP socket: Broken pipe (os error 32)`").

PR #8630, opened 2023-12-11
runtime-rs: ch: Change state when VM stopped

jodh-intel:
Make the CH (Cloud Hypervisor) `stop_vm()` method check the VM state before attempting to stop the VM, and update the state once the VM has stopped.

This avoids the method failing if called multiple times which will happen if the workload exits before the container manager requests that the container stop.

This change ensures the CH driver finishes cleanly.

Fixes: #8629.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
@@ -110,8 +110,6 @@ impl CloudHypervisorInner {
             }
         }
 
-        self.state = VmmState::VmmServerReady;
-
         Ok(())
     }
 
@@ -246,8 +244,6 @@ impl CloudHypervisorInner {
         if let Some(detail) = response {
             debug!(sl!(), "vm start response: {:?}", detail);
         }
-
-        self.state = VmmState::VmRunning;
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


dborquez (approved):
LGTM, thanks @jodh-intel

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -631,8 +631,20 @@ impl CloudHypervisorInner {
     }
 
     pub(crate) fn stop_vm(&mut self) -> Result<()> {
+        // If the container workload exits, this method gets called. However,
+        // the container manager always makes a ShutdownContainer request,
+        // which results in this method being called potentially a second
+        // time. Without this check, we'll return an error representing EPIPE
+        // since the CH API socket is at that point invalid.
+        if self.state != VmmState::VmRunning {
```
amshinde, line 639: I was going to suggest to explicitly check for Stopped State here, but looks like there are only 3 states defined. I would eventually like to explicitly distinguish between NotReady and Stopped, but this would work for now.

jodh-intel:
@jongwu - Is there a known problem with the ARM jenkins CI as I'm seeing lots of errors like this:

```
host system doesn't support vsock: stat /dev/vhost-vsock: no such file or directory
```

jodh-intel:
> is there a reason we don't update the state within cloud_hypervisor_shutdown

@cmaf - I've moved the state handling to the Hypervisor API calls now for clarity and consistency (see new commit).

> Also I wanted to check I'm understanding this correctly, if cloud_hypervisor_shutdown returns an error the state will remain VmRunning?

@cmaf - Good catch. I've updated the branch to set the state to `NotReady` before that call as, conveniently, that's the state we want on success and failure ;)

Issue #8633, opened 2023-12-12
dragonball: Disable packed virtqueue explicitly


justxuewei:
Vhost-user negotitation didn't support the layout of packed virtqueue. If a device and a guest kernel support packed virtqueue at the same time, the address of vring couldn't be acquired correctly.

PR #8634, opened 2023-12-12
dragonball: Disable packed virtqueue for vhost-user devices

justxuewei:
The layout of packed virtqueue isn't supported by `Endpoint::negotiate()`.
Communication between device and driver will be failed due to the failure
of parsing virtqueue if we don't disable the packed feature. This patch
fixes this issue.

Fixes: #8633


studychao (approved):
thanks for the deep research into this bug!

lifupan (approved):

jiangliu (approved):

jiangliu (approved):

PR #8637, opened 2023-12-12
deployment: kata-cleanup overlay, stable and latest components

zvonkok:
kata-cleanup relies on a stable and latest SHIMS env var. We need to add components for stable and latest that we can use for any manifest.
 We may also need to rename the manifest to match kata-deploy naming

### DIFF
--- a/tools/packaging/kata-deploy/kata-deploy/components/stable/env_stable_conf.yaml
+++ b/tools/packaging/kata-deploy/kata-deploy/components/stable/env_stable_conf.yaml
@@ -1,8 +1,8 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata: 
-  name: kata-deploy
-  namespace: kube-system
+  name: not-important
+  namespace: not-important
 spec:
   template:
     spec:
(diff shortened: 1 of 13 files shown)


tools/packaging/kata-deploy/kata-deploy/components/stable/env_stable_conf.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata: 
-  name: kata-deploy
-  namespace: kube-system
+  name: not-important
+  namespace: not-important
```
fidencio, line 5: Why? :-)

zvonkok: If you're using overlays, Kustomize may ignore changes in name or namespace specified in the overlay if they are not "properly" configured. The base configuration often takes precedence unless the overlay explicitly modifies these fields.

zvonkok: By modifying adding a prefix or suffix explicitly but since this is a selector nothing gets changed.

zvonkok:

zvonkok:
We have `components/stable` and `components/latest` that are used for `kata-deploy` and `kata-cleanup`, single source of truth for `stable` and `latest`.

Issue #8640, opened 2023-12-12
kata-ctl: Add "--show-default-config-paths" CLI option


jodh-intel:
`kata-ctl` (rust tooling) needs an option to dump out the config file locations that the runtime will try (in order).

This is for parity with `kata-runtime --show-default-config-paths` (golang tooling).

PR #8641, opened 2023-12-12
kata-ctl: Add option to dump config files

jodh-intel:
Add a `--show-default-config-paths` command line option for parity with `kata-runtime`.

Note that this requires the `KataCtlCli.command` to be optional so that the user can run simply:

```bash
$ kata-ctl --show-default-config-paths
```

... without also specifying a (sub-)command.

Fixes: #8640.


cmaf (approved):
thanks @jodh-intel

Issue #8646, opened 2023-12-13
Physical network interfaces passed in the network interface are not returned to original driver in case of failure.


amshinde:
I am currently testing with SRIOV CNI + device plugin. I am seeing some failures that need to be debugged. But I am seeing that in case the runtime errors out the sriov virtual function passed in the network namespace by the CNI remains bound to vfio-pci. 
Since the runtime binds the network interface to vfio-pci from host driver, in case of error the interface needs to be reassigned to the host driver during network cleanup, which is not happening currently.

PR #8647, opened 2023-12-13
Cleanup network to make sure physical interfaces are restores back to original host driver.

amshinde:
network: Move up defer block to cleanup network
    
 Move the defer for cleaning up network before the call to add network.
 This way if any change made by add network is reverted by in case of
 failure. This is particulary important for physical network interfaces
 as with this step we make sure that driver for the physical interface is
 reverted back to the original host driver. Without this the physical
 network iterface will remain bound to vfio.
    
 Fixes: #8646

### DIFF
--- a/src/runtime/virtcontainers/api.go
+++ b/src/runtime/virtcontainers/api.go
@@ -70,17 +70,18 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		}
 	}()
 
+	// network rollback
+	defer func() {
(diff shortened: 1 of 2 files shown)


amshinde:
Adding dnm label for now as I need to some more testing in case errors are encountered in other code paths such as error in CreateContainer. 
Would appreciate some feedback till then.

jodh-intel (approved):
Thanks @amshinde.

lgtm

dborquez (approved):
lgtm

Issue #8615, opened 2023-12-08
kata-deploy: Provide a way for the user to select which snapshotter will be set for a specific runtime class


fidencio:
Since containerd 1.7+, we have support for setting up snapshotters per runtime hanlders, and we should take advantage of that, as this would allow us to properly deploy Firecracker and Confidential Containers workloads, which require specific snapshotters.

I'd like to make it clear here that Kata Containers has **absolutely no business** in setting the snapshotter up, that's the responsibility of the cluster admin.  However, it's of our business to be able to properly set the runtime handler to use a specific snapshotter whenever it's desired **and** supported.

This, of course, comes with some drawbacks that will need to have a explicit documentation about, such as explicitly adding to the pod yaml of the workloads something like:
```
metadata:
  ...
  annotations:
    io.containerd.cri.runtime-handler: $runtimeclass_name
```

Ideally, we'd have this done as part of https://github.com/kata-containers/kata-containers/pull/7304, so the users won't need to do that themselves.

Also, this is **very specific** to containerd 1.7+.

We'd go with **no support**, at least for now, for containerd 1.6 and cri-o.

The way I'm thinking about having this set / configured in the daemonset is via an environment variable, which would have the following format: `"handler:snapshotter,handler:snapshotter,..."`.  Anything that's not empty and is not in this format would result in an error.

Once https://github.com/kata-containers/kata-containers/issues/8614 gets implemented, we should take advantage of that, but for now we'll not be checking whether the snapshotter is actually configured in the node.

fidencio:
So, another thing that I didn't take into consideration here, and I may need some help from other folks involved in the projects, is whether we can easily configure a snapshoitter for the k8s flavours we currently support.

We currently support:
* k0s
* k3s
* rke2
* vanilla

For sure we can set this up as part of vanilla and k3s.  Does someone happen to know how easily we can get this to work with rke2 and k0s?

PR #8655, opened 2023-12-13
kata-deploy: Allow setting up snapshotters per runtime handler

fidencio:
Since containerd 1.7.0 we can easily set a specific snapshotter to be used with a runtime handler, and we should take advantage of this, mostly as it'll help setting up any runtime using devmapper or nydus snapshotters.

This implementation here has a few caveats:
* The format expected for the SNAPSHOTTER_HANDLER_MAPPING is: `shim:snapshotter,shim:snapshotter,...`
* It only works with containerd 1.7 or newer
* We **never** change the default containerd snapshotter
* We don't do any check on our side to verify whether the snapshotter required is properly deployed
* Users will have to add an annotation to their pods, in order to use the snapshotter set up per runtime handler
  * Example: ``` metadata: ... annotations: io.containerd.cri.runtime-handler: kata-fc ```

Fixes: #8615

**NOTE**: Due to the lack of time from my side, as I'm vanishing for a few weeks, I've opted for not yet change the tests related to devmapper to use this new "feature".  Together with this decision comes a head's up that this is not tested anywhere yet, but it'll help us to unblock @ChengyuZhu6's series.

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -25,6 +25,10 @@ for allowed_hypervisor_annotation in "${non_formatted_allowed_hypervisor_annotat
 done
 allowed_hypervisor_annotations=$(echo $allowed_hypervisor_annotations | sed 's/,$//')
 
+SNAPSHOTTER_HANDLER_MAPPING="${SNAPSHOTTER_HANDLER_MAPPING:-}"
+IFS=',' read -a snapshotters <<< "$SNAPSHOTTER_HANDLER_MAPPING"
+snapshotters_delimiter=':'
+
 # If we fail for any reason a message will be displayed
 die() {
         msg="$*"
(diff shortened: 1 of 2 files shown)


ChengyuZhu6:
Thanks very much @fidencio !

tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -25,6 +25,10 @@ for allowed_hypervisor_annotation in "${non_formatted_allowed_hypervisor_annotat
 done
 allowed_hypervisor_annotations=$(echo $allowed_hypervisor_annotations | sed 's/,$//')
 
+SNAPSHOTTER_HANDLER_MAPPING="${SNAPSHOTTER_HANDLER_MAPPING:=}"
```
stevenhorsman, line 28: Based on https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html I think this is supposed to be `{SNAPSHOTTER_HANDLER_MAPPING:-}` as we are effectively double assigning here? I guess it will still work though?

stevenhorsman (commented):
I've tested lots of individual bits of this PR and the processing logic seems correct, apart from the case that I'll mentioned

stevenhorsman:
Add `do-not-merge` on this whilst we investigate issues with tomlq in the previous PR as mentioned in https://github.com/confidential-containers/operator/issues/305

beraldoleal (approved):
A part of @stevenhorsman comment, LGTM. Thanks @fidencio .

fidencio:
> Add `do-not-merge` on this whilst we investigate issues with tomlq in the previous PR as mentioned in [confidential-containers/operator#305](https://github.com/confidential-containers/operator/issues/305)

I'm removing the do-not-merge as the issue was solved.

stevenhorsman (approved):
LGTM. Thanks!

amshinde (approved):

Issue #8656, opened 2023-12-13
gha: Fix the StratoVirt failure on gha metrics


WenyuanLau:
Update the gha metrics test baseline for StratoVirt, and re-enable StratoVirt in gha metrics.

WenyuanLau:
Related: #8497

PR #8657, opened 2023-12-13
gha: Fix the failure of gha metrics for StratoVirt

WenyuanLau:
Update the metrics test baseline for StratoVirt and re-enable.

Fixes: #8656 

Related: #8496

### DIFF
--- a/.github/workflows/run-metrics.yaml
+++ b/.github/workflows/run-metrics.yaml
@@ -48,7 +48,7 @@ jobs:
       # all the tests due to a single flaky instance.
       fail-fast: false
       matrix:
-        vmm: ['clh', 'qemu']
+        vmm: ['clh', 'qemu', 'stratovirt']
       max-parallel: 1
     runs-on: metrics
     env:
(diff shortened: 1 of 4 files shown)


GabyCT:
If you enable but do not run each tests it will be safer and easier to verify which test has issues and avoid the random failures with other hypervisors

WenyuanLau:
@GabyCT I record the data for 5 times when StratoVirt running in GHA but failure, and I get the `fio` limit just by compare the performance between StratoVirt and other VMMs in the same baremetal. I did run the metrics in the baremetal for many times and get the stable data.

How can I repeat and get the inproper status you said? :)

WenyuanLau:
I've done some tests, how about we enable Speed & Density metric tests and skip other tests for StratoVirt first? After things are stable, we can enable other tests step by step. @GabyCT 

My plan: 
1. Speed & Density (launchtimes, memory_usage, memory_usage_inside_container) 
2. iperf
3. latency 
4. blogbench 
5. fio 
6. tensorflow

GabyCT (approved):

fidencio (approved):
lgtm, thanks @WenyuanLau!

PR #8659, opened 2023-12-13
metrics: Improve latency network cleanup

### LINKED ISSUE #8658: metrics: Improve latency network cleanup

GabyCT:
This PR improves the latency network cleanup by removing the pods even if the test fails.

Fixes #8658

### DIFF
--- a/tests/metrics/network/latency_kubernetes/latency-network.sh
+++ b/tests/metrics/network/latency_kubernetes/latency-network.sh
@@ -18,6 +18,12 @@ function remove_tmp_file() {
 
 trap remove_tmp_file EXIT
 
+function latency_cleanup() {
+	info "Latency test cleanup"
+	kubectl delete -f "${SCRIPT_PATH}/latency-server.yaml"
+	kubectl delete -f "${SCRIPT_PATH}/latency-client.yaml"
+}
+
 function main() {
 	init_env
 	cmds=("bc" "jq")
@@ -43,6 +49,8 @@ function main() {
 
 	# Create client
 	kubectl create -f "${SCRIPT_PATH}/latency-client.yaml"
+
+	trap latency_cleanup EXIT
 
 	# Get the names of the client pod
 	export client_pod_name="latency-client"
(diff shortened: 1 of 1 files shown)


tests/metrics/network/latency_kubernetes/latency-network.sh
```diff
@@ -18,6 +18,14 @@ function remove_tmp_file() {
 
 trap remove_tmp_file EXIT
 
+function latency_cleanup() {
+	info "Latency test cleanup"
+	kubectl delete -f latency-server.yaml
```
dborquez, line 23: I think you need to pass the full path to the yaml file.

```suggestion
	kubectl delete -f "${SCRIPT_PATH}/latency-server.yaml"
```

tests/metrics/network/latency_kubernetes/latency-network.sh
```diff
@@ -18,6 +18,14 @@ function remove_tmp_file() {
...
+	kubectl delete -f latency-client.yaml
```
dborquez, line 24: Same suggestion as before:

```suggestion
	kubectl delete -f "${SCRIPT_PATH}/latency-client.yaml
```

tests/metrics/network/latency_kubernetes/latency-network.sh
```diff
@@ -18,6 +18,14 @@ function remove_tmp_file() {
...
+}
+
+trap latency_cleanup EXIT
```
dborquez, line 27: If the script ends before the server and clients pods are created, call to `latency_cleanup` could report error.
I can think of two options, :
- You can make that the lines that deletes the pods return always true
- or move the `trap` line just after the creation of the server and client pods.

dborquez (changes requested):
good job, just a couple of comments inline

tests/metrics/network/latency_kubernetes/latency-network.sh
GabyCT: thanks @dborquez for the feedback changes applied

dborquez (approved):
thank you @GabyCT LGTM

chavafg (approved):

Issue #8663, opened 2023-12-14
Pull requests containing changes to *.md files have github actions that are stalled.


amshinde:
An example of this kind of PR is 
https://github.com/kata-containers/kata-containers/pull/8588

The above PR contains changes to *.md file.  Two github actions `darwin-tests.yaml` and `cargo-deny-runner.yaml` have `.*md`  files in their ignore path:
https://github.com/kata-containers/kata-containers/blob/c5cfad7023bb2df8bbe0e3cc3bc1cac666e57a85/.github/workflows/darwin-tests.yaml#L8
https://github.com/kata-containers/kata-containers/blob/c5cfad7023bb2df8bbe0e3cc3bc1cac666e57a85/.github/workflows/cargo-deny-runner.yaml#L9

This means the above two tests will be skipped for PRs containing changes in the `ignore-paths` secified. These two checks are currently marked as required in our CI. When github skips the above actions, it is as if the action was never run and CI for the PR is stalled.

PR #8664, opened 2023-12-14
github-actions: Remove ignore paths for required CI checks

amshinde:
If a PR contains files from the ignore-paths, these actions do not run as intended. However, the actions are make as required. And there does not seem to be a way to mark these as non-required in that case. As a result a PR containing the files from the ignore-paths remains stalled.
Hence remove the ignore-paths until github provides a way to mark actions that are skipped due to ignore-paths as non-required/passed.

Fixes: #8663

### DIFF
--- a/.github/workflows/cargo-deny-runner.yaml
+++ b/.github/workflows/cargo-deny-runner.yaml
@@ -6,7 +6,6 @@ on:
       - edited
       - reopened
       - synchronize
-    paths-ignore: [ '**.md', '**.png', '**.jpg', '**.jpeg', '**.svg', '/docs/**' ]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/ci-on-push.yaml
+++ b/.github/workflows/ci-on-push.yaml
@@ -13,8 +13,6 @@ on:
       - synchronize
       - reopened
       - labeled
-    paths-ignore:
-      - 'docs/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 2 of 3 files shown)


amshinde:
@chavafg  I think we faced a similar issue on one of Chelsea's PR with doc changes that was stalled as the github actions are skipped. I think its best to remove the ignore-paths until github supports marking the checks as successful when they are skipped.
Wdyt?

cmaf (approved):
@amshinde this LGTM. The downside is that these will run when they are not required? Or is there something I'm missing?

amshinde:
> The downside is that these will run when they are not required?

Yes, thats correct, they will run for doc chnages as well now. But that would mean we no longer have to force merge a doc PR using admin privs. 
We could possibly check for doc changes ourselves in the action and return success without running the actual tests.

jodh-intel (approved):
Thanks @amshinde.

lgtm

stevenhorsman:
Hey @amshinde - I wonder if it's worth also considering the doc exclusions on ci-on-push:
https://github.com/kata-containers/kata-containers/blob/0f80dc636c3a4df6eb5cc3034e5702c87c17778e/.github/workflows/ci-on-push.yaml#L16-L17
in this PR as well which causes the same issue of required actions being ignored?

amshinde:
> Hey @amshinde - I wonder if it's worth also considering the doc exclusions on ci-on-push:
> 
> https://github.com/kata-containers/kata-containers/blob/0f80dc636c3a4df6eb5cc3034e5702c87c17778e/.github/workflows/ci-on-push.yaml#L16-L17
> 
> 
> in this PR as well which causes the same issue of required actions being ignored?

Thanks @stevenhorsman for pointing that out. I have included  doc exclusions on ci-on-push as well.

stevenhorsman (approved):
LGTM. Thanks!

Issue #8665, opened 2023-12-14
runtime-rs: Refactor the code related to PCI paths and VFIO device driver initialize in DM.


Apokleos:
**Which feature do you think can be improved?**

Do enhancement of vfio/vhost-user block, this work is the preparation of PCIe Topology:
- make vhost-user pci_address more specific;
-  move pcipath code into a dedicated file;
- remove the wrapper(Option) of guest_pci_path, Option is not meaningful for vfio device
- refactor vfio driver implement

PR #8668, opened 2023-12-14
runtime-rs: Refactor the code related to PCI paths and VFIO device driver initialize in DM.

Apokleos:
Do enhancement of vfio/vhost-user block, this work is the preparation of PCIe Topology:

-  make vhost-user pci_address more specific;
-  move pcipath code into a dedicated file;
-  refactor vfio driver implement

Fixes: #8665


src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,216 @@
...
+// This encapsulates the PCI slot number a.k.a device number, which is
+// limited to a 5 bit value [0x00..0x1f] by the PCI specification
+//
+// To support multifunction device's, It's needed to extend
+// this to include the PCI 3-bit function number as well.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct PciSlot(pub u8);
+
+impl PciSlot {
+    pub fn convert_from_string(s: &str) -> Result<PciSlot> {
```
jodh-intel, line 36: It would be more idiomatic to implement `TryFrom` for `PciSlot`:

```rust
use std::convert::TryFrom;

impl TryFrom<&str> for PciSlot {
    type Error = anyhow::Error;

    fn try_from(s: &str) -> Result<Self> {
      // ...
    }
}

impl TryFrom<u32> for PciSlot {
    type Error = anyhow::Error;

    fn try_from(slot: u32) -> Result<Self> {
      // ...
    }
}
```

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    pub fn convert_from_u32(v: u32) -> Result<PciSlot> {
+        if v > MAX_PCI_SLOTS {
+            return Err(anyhow!("value {:?} exceeds MAX: {:?}", v, MAX_PCI_SLOTS));
+        }
+
+        Ok(PciSlot(v as u8))
+    }
+
+    pub fn convert_to_string(&self) -> String {
```
jodh-intel, line 62:
```
impl ToString for PciSlot {
    fn to_string(&self) -> String {
        // ...
    }
}
```

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,216 @@
...
+// If a device is directly connected to the root bus, which used in
+// lightweight hypervisors, such as dragonball/firecracker/clh, and
+// its PciPath.slots will contains only one PciSlot.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct PciPath {
+    // list of PCI slots
+    pub slots: Vec<PciSlot>,
+}
+
+impl PciPath {
```
jodh-intel, line 87: Again, could you implement the standard traits to make this more "rusty" ;)

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,216 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::{anyhow, Context, Result};
+use lazy_static::lazy_static;
+use regex::Regex;
+
+// get pci path from bdf, and the bdf will be start with "0000:00"
+lazy_static! {
+    static ref RE: Regex = Regex::new(r"^0000:00:([[:xdigit:]]{2}).[[:xdigit:]]$").unwrap();
```
jodh-intel, line 13: - Could you give this a clearer name, maybe `PCI_BDF_REGEX` or `BDF_REGEX`?
- Could you add an example to the comment showing the fully expanded BDF format.

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,216 @@
...
+pub struct PciSlot(pub u8);
+
+impl PciSlot {
+    pub fn convert_from_string(s: &str) -> Result<PciSlot> {
+        if s.is_empty() || s.len() > 2 {
+            return Err(anyhow!("string given is invalid."));
+        }
+
+        let base = 16;
+        let n = u64::from_str_radix(s, base).context("convert string to number failed")?;
```
jodh-intel, line 42: It might be worth adding `base` to the error for more clarity.

jodh-intel (commented):
Thanks @Apokleos - a few initial comments...

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -493,21 +493,14 @@ impl Device for VfioDevice {
...
-                            hostdev.guest_pci_path = Some(
-                                generate_guest_pci_path(hostdev.bus_slot_func.clone())
-                                    .map_err(|e| anyhow!("generate pci path failed: {:?}", e))?,
-                            );
-                        }
-
-                        // Safe to call unwrap here because of previous assignment.
-                        let pci_path = hostdev.guest_pci_path.clone().unwrap();
+                        hostdev.guest_pci_path =
+                            generate_guest_pci_path(hostdev.bus_slot_func.clone())
```
amshinde, line 506: This would break for cloud-hypervisor and other hypervisors. We assign the `guest_pci_path` in CH when the device is attached:
https://github.com/kata-containers/kata-containers/blob/04de6eb4fd9bbd5440d8a2f48e26806ba3756eb9/src/runtime-rs/crates/hypervisor/src/ch/inner_device.rs#L220

The same should be done for qemu and others. 
The guest_pci_path is empty only for dragonball, since it doesnt use PCIe at the moment. 
So overwriting this would break for CH and others. I had added back the above conidtional assignment for drangonball based on your suggestion in my other PR.

amshinde: If the assignment to guest_pci_path is necessary for dragonball, my suggestion would be to move this to dragonball specific `inner_device.rs` similar to how it is being done currently for CH.

amshinde (changes requested):
Can you address the comments I left on the third commit in this PR. Marking this as "needs-changes" till then as it breaks the CH driver.

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: Thx @jodh-intel for your suggestion! Yeah! make it more rusty!

Apokleos (commented):
Thx @jodh-intel fixed. please check it

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: fixed

quanweiZhou (approved):
Thanks, @Apokleos, LGTM.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -207,7 +207,7 @@ pub struct HostDevice {
     pub vfio_type: VfioDeviceType,
 
     /// guest PCI path of device
-    pub guest_pci_path: Option<PciPath>,
+    pub guest_pci_path: PciPath,
```
amshinde, line 210: I was giving this more thought..do we really need to remove the Option here. The `guest_pci_path` may be unitialised in some cases, for example qemu currently does not set `guest_pci_path` as that code is not implemented yet. In general I think since `guest_pci_path` may be unitialised/unassigned by the hypervisor I think it may be a good idea to keep this as an option.
Would like to hear what @jodh-intel and others think about this as well.

Apokleos: @amshinde I don't think so. 
It's an incontestable fact that `guest_pci_path` exists in Qemu, CLH, and Dragonball, regardless of VMM support， right? 
For CLH/dragonball vfio device, what does it mean if `guest_pci_path` uses Option? Is there a case where pci path does not exist? IMO, Option is not set for whether some VMM exists. It is set for whether `guest_pci_path` exists in the VMM.
I want this field `guest_pci_path` to have a definite meaning. The word "Option" gives a misleading impression, so I want to try to remove it.

The `guest_pci_path` is initialized with default value, and Rust does not allow uninitialized or unassigned variables to be used. This is guaranteed by its memory safety.

Apokleos: CC @jodh-intel

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: > Rustifying the code is definitely a good idea and I think it could have been put in separate patches on top of the original one which was only doing code movement. For both easier review (less changes per patch) and for educational purpose (highlight the improvement in git log).
>
Ok, Thx. I'll fix it.

src/runtime-rs/crates/hypervisor/src/device/driver/vhost_user.rs
```diff
@@ -50,8 +52,8 @@ pub struct VhostUserConfig {
     pub device_type: VhostUserType,
     /// guest block driver
     pub driver_option: String,
-    /// pci_addr is the PCI address used to identify the slot at which the drive is attached.
-    pub pci_addr: Option<String>,
+    /// pci_path is the PCI path used to identify the slot at which the drive is attached.
```
studychao, line 55: `drive` -> `driver`

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,259 @@
...
+            }
+        }
+
+        Ok(PciPath { slots: pci_slots })
+    }
+}
+
+// get pci path from bdf
+pub fn get_pcipath_from_bdf(bdf: &str) -> Option<u32> {
+    if !bdf.starts_with("0000:00") {
```
studychao, line 175: VMM that supports complex PCI hierarchy may not start with "0000:00" in BDF. 
In that case, they could not get pci path through this function, is that expected?

studychao (commented):
hi @Apokleos , a few comments.

src/runtime-rs/crates/hypervisor/src/device/driver/mod.rs
```diff
@@ -29,189 +29,3 @@ pub use virtio_vsock::{
...
-        assert!(dev_slot.is_some());
-        assert_eq!(dev_slot.unwrap().0, 5);
-
-        let root_slot = pci_path.get_root_slot();
-        assert!(root_slot.is_some());
-        assert_eq!(root_slot.unwrap().0, 1);
-    }
-
-    #[test]
-    fn test_get_host_guest_map() {
```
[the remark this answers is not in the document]
Apokleos: Thx @gkurz I split it into one dedicated commit to address the dropping case

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: Thx @gkurz Fixed, the following commit to refactor and re-implement the pci path CC @jodh-intel

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.slots.first().cloned()
+    }
+}
+
+impl ToString for PciPath {
+    // method to format the PciPath into a string
+    fn to_string(&self) -> String {
+        self.slots
+            .iter()
+            .map(|pci_slot| format!("{:02x}", pci_slot.0))
```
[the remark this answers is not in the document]
Apokleos: Fixed. keep it as it was, better.

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
```diff
@@ -0,0 +1,259 @@
...
+                Err(e) => return Err(anyhow!("slot is invalid with: {:?}", e)),
+            }
+        }
+
+        Ok(PciPath { slots: pci_slots })
+    }
+}
+
+// get pci path from bdf
+pub fn get_pcipath_from_bdf(bdf: &str) -> Option<u32> {
```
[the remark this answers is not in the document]
Apokleos: fixed, removed from this PR

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -391,11 +391,17 @@ impl VfioDevice {
             .get_vfio_device_vendor(&dev_bdf)
             .context("get property device and vendor failed")?;
 
+        // As some Vmm (Dragonball) only supports pre-allocated PCI Path, we must set the guest pcipath
```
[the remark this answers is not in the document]
Apokleos: Thx @studychao fixed

src/runtime-rs/crates/hypervisor/src/device/driver/vhost_user.rs
Apokleos: Fixed

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
Apokleos: Fixed. It was useless now, remove it.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
Apokleos: fixed

Apokleos:
> @Apokleos Can you address the comments I left on the third commit in this PR. Marking this as "needs-changes" till then as it breaks the CH driver.
>
Hi @amshinde I have addressed the comments about the codes which broke the CH, and please review it again. thx

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
jodh-intel: I agree with @amshinde  - if there is any chance `guest_pci_path` is not set by a hypervisor (unrelated to rust variable initialisation), then we should keep the `Option` imho as that could be a nasty source of bugs. Maybe @pmores can comment from the QEMU driver perspective?

src/runtime-rs/crates/hypervisor/src/device/pci_path.rs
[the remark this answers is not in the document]
Apokleos: Less unnecessary changes, easier for reviewers to read！

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -262,15 +262,21 @@ impl VfioDevice {
...
             config: dev_info.clone(),
             devices,
             device_options,
-        }
+        };
+
+        vfio_device
+            .initialize_vfio_device()
+            .context("initialize vfio device failed.")?;
+
```
gkurz, line 278: Not only the vfio device setup code is moved to a separate function, but it is now called at device creation time, not at device attach time. Can this change behavior or expectations of some other code ? In any case, this should be highlighted in the commit change log.

Apokleos: Thx @gkurz I will add info to the commit message to highlight that this change does not change the behavior of other code and is in line with expectations.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
Apokleos: @jodh-intel Yes I agree with the point that the `guest_pci_path` is unrelated to rust variable initialisation.

Apokleos: @jodh-intel @amshinde But I’m curious if Qemu can do without `guest_pci_path`. Or is there a new way to work around ? 
AFAIK, regardless of the VMM, the same kata-agent code will be used, right?
For VFIO devices, we also need to do the mapping between the host device BDF and the PCI path that will be generated after it is inserted into the VM on the runtime side. Then, when building the kata agent device, we assign the mapping relationship to `Device.options`, and then pass it to the kata agent and parse the mapping relationship.

If Qemu can avoid do this,   or `guest_pci_path` is no point for it. Ok, I will agree with Option<PciPath>

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
Apokleos: fixed

amshinde (approved):

amshinde:
@Apokleos Changes look good to me, however can you hold off merging, I just want to test the PR with cloud-hypervisor once to make sure things are working for vfio.

Apokleos:
> @Apokleos Changes look good to me, however can you hold off merging, I just want to test the PR with cloud-hypervisor once to make sure things are working for vfio.
>
Thx @amshinde  Ok, I'll wait for your feedback with clh.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
gkurz: > @jodh-intel @amshinde But I’m curious if Qemu can do without `guest_pci_path`. Or is there a new way to work around ? And if we're all concern about the Option and have difference points about it, I will remove the commit about removing Option. We can disscus it when Qemu does implement the vfio feature.

We're not there yet but I agree with @Apokleos. The QEMU driver in the go runtime does set it :

```
                device.GuestPciPath, err = q.arch.qomGetPciPath(device.ID, &q.qmpMonitorCh)
```

I see no reason for the runtime-rs driver to behave differently.

gkurz (approved):
LGTM. Thanks @Apokleos !

gkurz:
> > @Apokleos Changes look good to me, however can you hold off merging, I just want to test the PR with cloud-hypervisor once to make sure things are working for vfio.
> 
> Thx @amshinde Ok, I'll wait for your feedback with clh.

Let's then add the `do-not-merge` label for now.

amshinde:
@Apokleos Please feel free to merge this PR.

Apokleos:
> > > @Apokleos Changes look good to me, however can you hold off merging, I just want to test the PR with cloud-hypervisor once to make sure things are working for vfio.
> > 
> > 
> > Thx @amshinde Ok, I'll wait for your feedback with clh.
> 
> Let's then add the `do-not-merge` label for now.
>
Thx @gkurz

Apokleos:
> @Apokleos Please feel free to merge this PR.
>
Ok, Thx @amshinde

Issue #8671, opened 2023-12-14
runtime-rs: guest_id and vhost_fd for VsockDevice should be initialised when necessary


BbolroC:
**Which feature do you think can be improved?**

As a follow-up for #8516 to realise the 2nd step in https://github.com/kata-containers/kata-containers/pull/8414#issuecomment-1825104380, we should separate an initialisation part from an instantiation part for a struct `VsockDevice`.

PR #8672, opened 2023-12-14
runtime-rs: Separate init_config() from new() for struct VsockDevice

BbolroC:
As a follow-up for #8516, guest_cid and vhost_fd are not necessarily initialised via new(). Instead, the fields should be initialised later when they are really used to construct hypervisor's parameters.
This commit is to separate init_config() from new() to initialise guest_cid and vhost_fd and leave only the assignment of id for the existing function.

Fixes: #8671

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
@@ -330,9 +330,11 @@ impl DeviceManager {
                 // No need to do find device for hybrid vsock device.
                 Arc::new(Mutex::new(HybridVsockDevice::new(&device_id, hvconfig)))
             }
-            DeviceConfig::VsockCfg(_vconfig) => {
+            DeviceConfig::VsockCfg(vconfig) => {
                 // No need to do find device for vsock device.
-                Arc::new(Mutex::new(VsockDevice::new(device_id.clone()).await?))
+                Arc::new(Mutex::new(
+                    VsockDevice::new(device_id.clone(), vconfig).await?,
+                ))
             }
             DeviceConfig::ShareFsCfg(config) => {
                 // Try to find the sharefs device. If found, just return matched device id.
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
@@ -117,15 +117,19 @@ nix::ioctl_write_ptr!(
 const CID_RETRY_COUNT: u32 = 50;
 
 impl VsockDevice {
-    pub async fn new(id: String) -> Result<Self> {
-        let (guest_cid, _vhost_fd) = generate_vhost_vsock_cid()
-            .await
-            .context("generate vhost vsock cid failed")?;
-
+    pub async fn new(id: String, config: &VsockConfig) -> Result<Self> {
         Ok(Self {
             id,
-            config: VsockConfig { guest_cid },
+            config: config.clone(),
         })
+    }
+
+    pub async fn init_config(&mut self) -> Result<File> {
+        let (guest_cid, vhost_fd) = generate_vhost_vsock_cid()
+            .await
+            .context("generate vhost vsock cid failed")?;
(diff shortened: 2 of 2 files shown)


src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
```diff
@@ -118,15 +118,21 @@ const CID_RETRY_COUNT: u32 = 50;
...
         Ok(Self {
             id,
-            config: VsockConfig { guest_cid },
+            config: VsockConfig {
+                guest_cid: libc::VMADDR_CID_ANY,
+            },
         })
     }
+
+    pub async fn init_config(&mut self) -> Result<File> {
```
studychao, line 129: seems like this function is not used anywhere, will the actual usage of this will be introduced later?

studychao (commented):
LGTM, but a little question

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
BbolroC: yeah, coming soon from https://github.com/kata-containers/kata-containers/pull/8185 😉

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
```diff
@@ -118,15 +118,21 @@ const CID_RETRY_COUNT: u32 = 50;
...
-        let (guest_cid, _vhost_fd) = generate_vhost_vsock_cid()
-            .await
-            .context("generate vhost vsock cid failed")?;
-
         Ok(Self {
             id,
-            config: VsockConfig { guest_cid },
+            config: VsockConfig {
+                guest_cid: libc::VMADDR_CID_ANY,
+            },
```
Apokleos, line 125: IMO, the vsock config is first initialized in [prepare_vm_socket_config](https://github.com/kata-containers/kata-containers/blob/0f80dc636c3a4df6eb5cc3034e5702c87c17778e/src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs#L264), but it's not passed down as an argument in VsockDevice::new(). What's about changing the `VsockDevice::new(id)` to `VsockDevice::new(id, config) `and in the body is `{
    Self {id, config}
}`

BbolroC: It sounds better. I will reflect your suggestion on the code. Thanks! 😉

Apokleos (commented):
Thx @BbolroC for your work! one comment!

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_vsock.rs
BbolroC: Done, I hope the change would meet your expectation. Thanks!

Apokleos (approved):
Thx @BbolroC for your work. LGTM!

gkurz (approved):
LGTM. Thanks @BbolroC !

src/runtime-rs/crates/hypervisor/src/device/device_manager.rs
```diff
@@ -330,9 +330,9 @@ impl DeviceManager {
                 // No need to do find device for hybrid vsock device.
                 Arc::new(Mutex::new(HybridVsockDevice::new(&device_id, hvconfig)))
             }
-            DeviceConfig::VsockCfg(_vconfig) => {
+            DeviceConfig::VsockCfg(vconfig) => {
                 // No need to do find device for vsock device.
-                Arc::new(Mutex::new(VsockDevice::new(device_id.clone()).await?))
+                Arc::new(Mutex::new(VsockDevice::new(device_id.clone(), vconfig).await?))
```
gkurz, line 335:
```suggestion
                Arc::new(Mutex::new(
                    VsockDevice::new(device_id.clone(), vconfig).await?,
                ))
```

in order to make `cargo fmt -- --check` happy and fix https://github.com/kata-containers/kata-containers/actions/runs/7220732032/job/19674258988?pr=8672 .

Issue #8678, opened 2023-12-15
kata-deploy: tomlq breaks containerd config


stevenhorsman:
In https://github.com/kata-containers/kata-containers/pull/8639 a change was made to use `tomlq` to update the containerd config.toml, but in the operator (https://github.com/confidential-containers/operator/issues/305) we've found that this changes breaks the payload. I've been able to isolate the reason for this - and tomql is converting a float to an int, which breaks the containerd parsing. The cut down reproduce of this issue can be seen:
```bash
$ containerd config default > config.toml
$ cp config.toml config.toml.bak
$ tomlq -i -t '.debug.level = "debug"' config.toml
$ diff -w config.toml config.toml.bak
18c18
< level = "debug"
---
>   level = ""
36a37
>
98a100,101
>       [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
>
163c166
< sampling_ratio = 1
---
>     sampling_ratio = 1.0
223a227
>
```
showing that sample_ratio is switched from `1.0` to `1`.

After this change and restarting container we see the error:
```
containerd: (204, 1): Can't convert 1(int64) to float64
```

There is still an outstanding question of how this passes in the kata CI?

We have a few options:
1. Horrible option - create an issue in the tomlq project add some post processing to convert the sampling_ratio back to a float with sed. I don't think this is viable as the point of using tomlq was to allow us to be less brittle in case of other containerd configs that we haven't seen and this just moves that problem along
2. Try and find another toml tool that works better and switch it in place of tomlq. I think Fabiano tried multiple options and found issues with them
3. Write out own program to do the editting - Magnus has suggested toml-edit: https://github.com/confidential-containers/operator/issues/304 might help for rust
4. Back out the PR and use sed

I think that 4 is the solution we need to do now as this close to the holidays for part of the community I don't want to leave kata-deploy broken and I think 2 and 3 might take more time to resolve. I also would like to understand why the CI didn't catch this before we re-introduce changes.

stevenhorsman:
Ok, one mystery down - I ran the kata-deploy test locally:
```
DOCKER_REGISTRY=" quay.io"
export DOCKER_REGISTRY="quay.io"
export DOCKER_REPO="kata-containers/kata-deploy-ci"
export DOCKER_TAG="kata-containers-latest"
bash tests/functional/kata-deploy/gha-run.sh run-tests
```
and they passed:
```
[run-kata-deploy-tests.sh:21] INFO: Run tests
[...]

PR #8679, opened 2023-12-15
gha: kata-deploy: Revert containerd config break

stevenhorsman:
After kata-deploy has installed, check containerd is still in active state, to ensure that we didn't corrupt the containerd config during kata-deploy's edits

Fixes: #8678

### DIFF
--- a/tests/functional/kata-deploy/kata-deploy.bats
+++ b/tests/functional/kata-deploy/kata-deploy.bats
@@ -36,17 +36,17 @@ setup() {
 	yq write -i "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[4].value' --tag '!!str' "true"
 	# Let the `kata-deploy` create the default `kata` runtime class
 	yq write -i "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[5].value' --tag '!!str' "true"
-	
+
 	if [ "${KATA_HOST_OS}" = "cbl-mariner" ]; then
 		yq write -i "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[+].name' "HOST_OS"
 		yq write -i "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[-1].value' "${KATA_HOST_OS}"
 	fi
-	
+
 	echo "::group::Final kata-deploy.yaml that is used in the test"
 	cat "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
 	grep "${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}" "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" || die "Failed to setup the tests image"
 	echo "::endgroup::"
-	
+
 	kubectl apply -f "${repo_root_dir}/tools/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml"
 	if [ "${KUBERNETES}" = "k0s" ]; then
 		kubectl apply -k "${repo_root_dir}/tools/packaging/kata-deploy/kata-deploy/overlays/k0s"
@@ -64,7 +64,7 @@ setup() {
 	sleep 30s
 }
 
-@test "Test runtimeclasses are being properly created" {
+@test "Test runtimeclasses are being properly created and container runtime not broken" {
 	# We filter `kata-mshv-vm-isolation` out as that's present on AKS clusters, but that's not coming from kata-deploy
 	current_runtime_classes=$(kubectl get runtimeclasses | grep -v "kata-mshv-vm-isolation" | grep "kata" | wc -l)
 	[[ ${current_runtime_classes} -eq ${expected_runtime_classes} ]]
@@ -73,6 +73,20 @@ setup() {
 	do
 		kubectl get runtimeclass | grep -E "${handler_re}"
 	done
+
+	# Ensure that kata-deploy didn't corrupt containerd config, by trying to get the container runtime and node status
+	echo "::group::kubectl node debug"
+	kubectl get node -o wide
+	kubectl describe nodes
+	echo "::endgroup::"
+
+	# Wait to see if the nodes get back into Ready state - if not then containerd might be having issues
+	kubectl wait nodes --timeout=60s --all --for condition=Ready=True
+
+	# Check that the container runtime verison doesn't have unknown, which happens when containerd can't start properly
+	container_runtime_version=$(kubectl get nodes --no-headers -o custom-columns=CONTAINER_RUNTIME:.status.nodeInfo.containerRuntimeVersion)
+	[[ ${container_runtime_version} != *"containerd://Unknown"* ]]
+
 }
 
 teardown() {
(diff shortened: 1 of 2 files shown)


stevenhorsman:
@Amulyam24 - FYI the ppc64le checks are failing with:
```
Step 1/23 : FROM ubuntu:20.04
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
make[1]: *** [tools/packaging/kata-deploy/local-build/Makefile:75: kernel-tarball-build] Error 1
make: *** [tools/packaging/kata-deploy/local-build/Makefile:111: kernel-tarball] Error 2
```
I'm guessing that we need to login to docker hub on the self-hosted node as the anonymous rate limit is pretty low?

beraldoleal (approved):
LGTM, thanks @stevenhorsman. My only question is if this is indeed fixing #8678 as described at the commit message. Unless I'm missing something, this is just adding a test, not fixing the actual problem, right?

stevenhorsman:
> LGTM, thanks @stevenhorsman. My only question is if this is indeed fixing #8678 as described at the commit message. Unless I'm missing something, this is just adding a test, not fixing the actual problem, right?

Correct - this is just the first stage of the PR, assuming it works (and the test fails), we'll have a way to test the issue and I plan to revert part of the tomlq PR and check that fixes things. I'll add `do-not-merge` to this to show it's not completed.

GabyCT (approved):

stevenhorsman:
Ok, I think this containerd test approach works for baremetal systems, but not a other clusters as we can't necessarily get in an access their containerd. I can try using kata-debug for this process though...

stevenhorsman:
I didn't get kata-debug working reliably, so just using kubectl with custom columns to extract status and containerruntimeversion. I'll also temporarily dropped the fix, so I can check that this test is correctly failing.

stevenhorsman:
In the previous run, I got some of the deploy tests to fail as expected, but more of them were still passing, which doesn't really make sense, so I've upped the sleep to see if that gives enough time for the containerd config changes to take effect.

stevenhorsman:
As I shared in https://github.com/confidential-containers/operator/issues/305#issuecomment-1862430051 - I'm not super happy with this change and think the testing could do with some wider input, but the revert of using tomlq resolves the problem of kata-deploy corrupting containerd's config, so I think it's good enough to merge and unblock people until January when more people are back to discuss this.

danmihai1 (approved):

tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -73,6 +73,19 @@ setup() {
...
 		kubectl get runtimeclass | grep -E "${handler_re}"
 	done
+
+	# Ensure that kata-deploy didn't corrupt containerd config, by trying to get the container runtime and node status
+	echo "::group::kubectl pods"
+	kubectl get node -o wide
+	kubectl describe node --selector='!node-role.kubernetes.io/control-plane'
+	echo "::endgroup::"
+
+	ready_status=$(kubectl get node --no-headers --selector='!node-role.kubernetes.io/control-plane' -o custom-columns=STAUS:.status.conditions\[\?\(\@.type==\'Ready\'\)\].status)
```
gkurz, line 83: Even if it isn't printed because of `--no-headers`, `s/STAUS/STATUS` :smile:

Could there be a situation where `kata-deploy` would cause the node to be temporary not ready ?

If so, maybe do `kubectl wait node --selector='!node-role.kubernetes.io/control-plane' --for condition=Ready=True` ?

tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -73,6 +73,19 @@ setup() {
 	do
 		kubectl get runtimeclass | grep -E "${handler_re}"
 	done
+
+	# Ensure that kata-deploy didn't corrupt containerd config, by trying to get the container runtime and node status
+	echo "::group::kubectl pods"
```
gkurz, line 78: `s/pods/node` ?

tests/functional/kata-deploy/kata-deploy.bats
stevenhorsman: Good suggestion - I think we also have the other way around, where it takes kata-deploy a little while to break the containerd config, so I'm not sure how best to balance them, but doing something like `kubectl wait node --selector='!node-role.kubernetes.io/control-plane' --timeout=60s --for condition=Ready=True` would give it a fair chance of succeeding, whether we want to bias in favour of success, or not in this test is a good question though!

stevenhorsman: I've pushed this change - if nothing else it's easier to read and understand.

stevenhorsman: Ah, slight issue that a bunch of the test clusters are only control-plane clusters without workers it seems so `kubectl wait node --selector='!node-role.kubernetes.io/control-plane' --timeout=60s --for condition=Ready=True` fails with `error: no matching resources found`, so we might need to re-think this approach

stevenhorsman: Maybe `kubectl wait nodes --timeout=60s --all --for condition=Ready=True` is the play to check that all the node are ready?

tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -73,6 +73,20 @@ setup() {
 	do
 		kubectl get runtimeclass | grep -E "${handler_re}"
 	done
+
+	# Ensure that kata-deploy didn't corrupt containerd config, by trying to get the container runtime and node status
+	echo "::group::kubectl node debug"
+	kubectl get node -o wide
+	kubectl describe node --selector='!node-role.kubernetes.io/control-plane'
```
gkurz, line 80: Maybe drop the selector since all nodes in the cluster can be control-plane ?

tests/functional/kata-deploy/kata-deploy.bats
```diff
@@ -73,6 +73,20 @@ setup() {
...
+	echo "::endgroup::"
+
+	# Wait to see if the nodes get back into Ready state - if not then containerd might be having issues
+	kubectl wait nodes --timeout=60s --all --for condition=Ready=True
+
+	# Check that the container runtime verison doesn't have unknown, which happens when containerd can't start properly
+	container_runtime_version=$(kubectl get node --no-headers --selector='!node-role.kubernetes.io/control-plane' -o custom-columns=CONTAINER_RUNTIME:.status.nodeInfo.containerRuntimeVersion)
```
gkurz, line 87: ditto

tests/functional/kata-deploy/kata-deploy.bats
stevenhorsman: Good point - sorry for my lack of consistency after I made the change in the wait. Will update now!

gkurz (approved):
LGTM. Thanks @stevenhorsman !

fidencio:
Folks, I don't like the commit being reverted, but I must say it's what it's at this point.

As I'm working Today, I've decided to poke steve and get to the root cause of the issue, and here's the full explanation.
`tomlq` relies on `jq` in order to do the parsing / editing of the file, and it's done in this specific [code path](https://github.com/kislyuk/yq/blob/0243832710bf1e6a7c68d75e83957fe410c8064b/yq/__init__.py#L346-L351C57)
```python
            elif input_format == "toml":
                import tomlkit

                for input_stream in input_streams:
                    json.dump(tomlkit.load(input_stream), jq.stdin, cls=JSONDateTimeEncoder)  # type: ignore
                    jq.stdin.write("\n")  # type: ignore
```

The interesting part to understand here is that jq is an invocation of the `jq` command, as shown [here](https://github.com/kislyuk/yq/blob/0243832710bf1e6a7c68d75e83957fe410c8064b/yq/__init__.py#L209-L215)
```python
        jq = subprocess.Popen(
            ["jq"] + list(jq_args),
            stdin=subprocess.PIPE,
            stdout=subprocess.PIPE if converting_output else None,
            close_fds=False,
            universal_newlines=True,
        )
```

Okay, with this in mind, we can simplify a whole lot the test case.
Let's test with ` echo '{"foo": 1.0}' | jq .foo`.
```bash
/ # jq --version
jq-1.6
/ # echo '{"foo": 1.0}' | jq .foo
1
```

Bingo, we can reproduce the issue we've faced and we know that the version of jq being used is 1.6  When taking a look at `jq`'s GitHub page, I've noticed a newer release, 1.7, and decided to give it a try before hunting this issue down.

```bash
⋊> Downloads ./jq-linux-amd64 --version
jq-1.7
⋊> Downloads echo '{"foo": 1.0}' | jq .foo
1.0
```

And, hey, it works as expected. 

So, the actual solution for the issue should be:
* Re-revert the reverted commit
* Bump the jq version used in our daemonse
* Profit

I will open a PR with this soon enough, and this will allow us to move on and get the snapshotter work merged, unblocking others depending on this one.

PR #8682, opened 2023-12-16
ci: Use static checks from kata repo for lib functions

### LINKED ISSUE #8681: Change all instances of static-checks.sh to main repo
GHA was changed to use static-checks.sh in the main repo and not the test repo. There are still a few instances of using the old repo, so we should switch these to the new one. This will make using the tests repo for static checks obsolete.

cmaf:
Change the two functions in lib.sh to use the static checks script from
the kata containers repo instead of tests. Remove cloning the repo from
these functions since we don't need it anymore. Leave these two
functions because the document checking one may be used locally and the
static checks one is called from the virtcontainers Makefile.
    
Fixes #8681
  
### DIFF
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -5,6 +5,8 @@
 
 set -o nounset
 
+export kata_repo="github.com/kata-containers/kata-containers"
+export kata_repo_dir="$GOPATH/src/$kata_repo"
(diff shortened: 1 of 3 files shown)


gkurz (approved):
LGTM. Thanks @cmaf !

GabyCT (approved):

Issue #8683, opened 2023-12-16
kata-runtime fails with runtime-rs when exec


liubogithub:
# Description of problem

kata-runtime fails with runtime-rs when exec

# Expected result

successfully enter debug console of sandbox.

# Actual result

```
file /opt/kata/bin/qemu-system-x86_64 does not exist
```

Apokleos:
Hi @liubogithub sorry for the trouble with you.
But I want to clarify that kata-runtime is for kata-containers with go, and kata-ctl is for kata3.x(runtime-rs). 
Next, some documents about the kata-runtime/kata-ctl will be added to enhance the user experience.
Thx

PR #8684, opened 2023-12-16
configuration: update comments for debug_console_enabled

liubogithub:
In terms of entering debug console, kata-runtime only works for runtime(golang), for runtime-rs, it's supposed to use kata-ctl.

fixes:  #8683

v2: 
* updated commit log.
* updated config for clh as well.

### DIFF
--- a/src/runtime-rs/config/configuration-cloud-hypervisor.toml.in
+++ b/src/runtime-rs/config/configuration-cloud-hypervisor.toml.in
@@ -222,7 +222,7 @@ container_pipe_size=@PIPESIZE@
 # Enable debug console.
 
 # If enabled, user can connect guest OS running inside hypervisor
-# through "kata-runtime exec <sandbox-id>" command
+# through "kata-ctl exec <sandbox-id>" command
 
 #debug_console_enabled = true
 
--- a/src/runtime-rs/config/configuration-dragonball.toml.in
+++ b/src/runtime-rs/config/configuration-dragonball.toml.in
@@ -235,7 +235,7 @@ container_pipe_size=@PIPESIZE@
 # Enable debug console.
 
 # If enabled, user can connect guest OS running inside hypervisor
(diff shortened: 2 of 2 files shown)


Apokleos:
> kata-runtime only works for qemu, for dragonball, it's supposed to use kata-ctl.
> 
> fixes: #8683
>
Thx @liubogithub 
Could you please edit your commit message? IMO, its description is not OK. kata-runtime is for kata-containers with go, not only for qemu, kata-ctl is for runtime-rs, not only for  dragonball.

amshinde:
> kata-runtime is for kata-containers with go, not only for qemu, kata-ctl is for runtime-rs, not only for dragonball.

Agree. Please make the change as @Apokleos pointed out. Also similar change is required for `configuration-cloud-hypervisor.toml.in` file as well. While you are at it, can you make similar change in that file as well?

gkurz:
@liubogithub please add a `Fixes: #8683` line in the commit change log. This is required by the commit message check.

liubogithub:
@gkurz Sorry for the late reply. Now the commit is with the Fixes tag.

liubogithub:
@studychao can you please review?

fidencio (approved):
lgtm, thanks @liubogithub!

studychao (approved):
LGTM! thanks!

Apokleos:
Thx @liubogithub update your commit message. It can pass CI now. But I think it's still different from the suggestion in [patch-format](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#main-patch). Could you please correct it again like this:
```
pod: Remove token from Cmd structure

The token and pid data will be hold by the new Process structure and
they are related to a container.

Fixes: #{Reference Issue ID}

Signed-off-by: ABC<abc@xxx.com>
```
The `Fixes #xxx` or `Fixes: #xxx` should be put between the main message and Signed-off-by. 

Everything else seems fine to me ！

PR #8685, opened 2023-12-16
runtime-rs: Add a PciSlot fuzz-harness

nathaniel-brough:


silvergasp:
Not sure how to fix the backport label CI error. But I don't think this needs/should be back ported.

Apokleos:
Thx @silvergasp for your work bringing the fuzz-harness to improve the security and stability. 
I suggest that add more document to explain it and split the README from the PR into one commit.

Apokleos:
I want to confirm that is the subdir fuzz placed at x/runtime-rs/ ok? CC @lifupan @bergwolf @gkurz

src/runtime-rs/fuzz/README.md
```diff
@@ -0,0 +1,25 @@
+# Fuzz testing
+To get started with fuzz testing you'll need to install cargo-fuzz.
+
+`cargo install cargo-fuzz`
+
+You can list the available fuzz harnesses using cargo fuzz. e.g.
+```
+$ cd src/runtime-rs/fuzz
+$ cargo +nightly fuzz list
```
Apokleos, line 9: I'm wondering if any Rust stable versions can be used instead of nightly versions to run cargo fuzz list ？
I am concerned that we should minimize the use of nightly versions as much as possible.

src/runtime-rs/fuzz/README.md
```diff
@@ -0,0 +1,25 @@
...
+fuzz_hypervisor_device_roundtrip
+```
+
+To run a fuzz harness simply use the following command.
+
+```
+$ cargo +nightly run fuzz_hypervisor_device_roundtrip
```
Apokleos, line 16: The same question as the previous one.

Apokleos (commented):
Thx @silvergasp for your work. Some initial comments.

Apokleos:
One more question: Do you plan to support kata-runtime(golang) with fuzz ?

src/runtime-rs/fuzz/README.md
amshinde: +1 For using rust stable versions.

jiangliu (approved):

jiangliu:
Thanks for introducing the Fuzz framework:)

src/runtime-rs/fuzz/README.md
silvergasp: It turns out you can use stable for `cargo fuzz list`, but nightly is [still required for building the fuzzer's](https://rust-fuzz.github.io/book/cargo-fuzz/setup.html?highlight=nightly#requirements). The reason nightly should be used is because it allows for certain instrumentation flags that are not available on stable. It's worth pointing out however that the fuzzer's don't use nightly API's and as a result *can* be built on stable. It's just that without the instrumentation the fuzzer ends up being next to useless, as the fuzzing engine will use the [coverage instrumentation to guide it's inputs](https://llvm.org/docs/LibFuzzer.html#id12), as well as various other things.

src/runtime-rs/fuzz/README.md
silvergasp: See response above. Nightly is required for instrumentation. But no nightly API's are used to write the fuzzer's themselves.

silvergasp:
> One more question: Do you plan to support kata-runtime(golang) with fuzz ?

I'd be happy to. Though I'll probably start with rust first. A more general question here, about the state of these runtimes, as I couldn't find this in the docs. Is the plan to support both rust/golang runtimes indefinitely, or is the runtime being ported between languages e.g. rust-> golang or golang -> rust. I ask as I'd like to focus my efforts on API's that are going to be actively developed in the future rather than API's that are actively being replaced.

silvergasp:
Just looking at the [build failure](https://github.com/kata-containers/kata-containers/actions/runs/7234404921/job/19776405409?pr=8685#step:6:33936). It mentions that the README.md doc I wrote isn't referenced. Should this be referenced somewhere in the doc directory?

silvergasp:
Latest change attempts to fix the doc linking static check.

silvergasp:
Friendly ping is there anything more I need to do here?

Issue #6222, opened 2023-02-06
agent-ctl: Rework client code to use a trait


jodh-intel:
Rework the code in [`src/client.rs`](https://github.com/kata-containers/kata-containers/blob/main/src/tools/agent-ctl/src/client.rs) to use rust traits.

The current code uses a very "C-like" approach to build the list of agent commands (`AGENT_CMDS`): essentially each element is a structure containing a function pointer (type `AgentCmdFp`). A more idiomatic / "rusty" approach would be to define a trait and then define separate structures (which could be empty), one for each command. Each struct can then implement the trait.

jodh-intel:
Related: #5348.

joannejchen:
Hi! @scotran and I are students from UT Austin that are taking a Virtualization course where one of the projects is attempting to contribute to repositories -- are we allowed to take on this issue (is the outreach tag for a specific audience)?

kalil-pelissier:
Hello @jodh-intel  👋,

I'm new to the project and I would like to contribute. This issue seems like a great place to begin. Would it be possible to take it on?

Thank you!

jodh-intel:
Hi @kalil-pelissier - Welcome to the project! Sure - that would be great, thanks!  I'll assign this one to you now... Let us know if you hit any issues (here or Slack).

PR #8689, opened 2023-12-17
agent-ctl: Refactor agent-ctl command handling with idiomatic rust trait structure.

kalil-pelissier:
The current code uses a very "C-like" approach to build the list of agent commands (AGENT_CMDS): essentially each element is a structure containing a function pointer (type AgentCmdFp). A more idiomatic / "rusty" approach would be to define a trait and then define separate structures (which could be empty), one for each command. Each struct can then implement the trait.

Fixes: #6222 

Signed-off-by: kalil kalil.pelissier@gmail.com


kalil-pelissier:
Hi @jodh-intel,
I have almost done but I'm hesitating on the way to getting the list of commands with the `-c list` option.

> For now I have implemented a macro that just prints a String that lists all the commands.

What I would like to do is to add a `description` attribute to each command struct, but there is a problem with accessing all the commands in one point as they are only instantiated when they are executed. A workaround will be to create a struct that gets a vec attribute with the list of all the commands (agent and builtin) but it's looking exaggerated only for this.

jodh-intel:
Thanks @kalil-pelissier !

> What I would like to do is to add a description attribute to each command struct, but there is a problem with accessing all the commands in one point as they are only instantiated when they are executed. A workaround will be to create a struct that gets a vec attribute with the list of all the commands (agent and builtin) but it's looking exaggerated only for this.

I think https://docs.rs/lazy_static/latest/lazy_static will help here. You could then add a new trait method. Something like:

```rust
pub trait AgentCmd {
    fn exec(
        &self,
        ctx: &Context,
        client: &AgentServiceClient,
        health: &HealthClient,
        options: &mut Options,
        args: &str,
    ) -> (Result<()>, bool);

    fn descr(&self) -> &'static str;
}

#[derive(Debug, Clone, Default)]
struct FooCmd;

impl AgentCmd for FooCmd {
    // ...

    fn descr(&self) -> &'static str {
         "Foo command"
    }
}

lazy_static! {
    static ref COMMANDS: RwLock<Vec<Box<dyn AgentCmd + Send + Sync>>> = RwLock::new(vec![
        Box::new(FooCmd::default()),
        Box::new(BarCmd::default()),
        Box::new(BazCmd::default())

        // ...
      ]);
  }

fn test() -> Result<()> {
      // XXX: This **has** to be an upwrap here unfortunately ;(
      let commands = COMMANDS.read().unwrap();

      for cmd in commands.iter() {
          println!("command: {:?}", cmd.descr());
      }
}
```

You could then interate through COMMANDS and call `.descr()` for each object. That would also allow you to remove the hard-coded list of commands in `print_cmd_list!()` macro.

kalil-pelissier:
Hi @jodh-intel, looking for review.

jodh-intel (approved):
Thanks @kalil-pelissier!

lgtm :smile: 

btw, entirely optional but you could add a commit to fix the examples subcommand output as I spotted a mistake: it shows `--repeat 7` and `--repeat -1` but that's the old syntax. It should now display `-c "repeat 7"` and `-c "repeat -1"` respectively (note that both of those could also be specified using `--cmd` rather than `-c`).

kalil-pelissier:
Ok, I wanted to open another issue to fix it but if it's okay for you I'll fix it here.

Also, I wanted to propose moving the `list` command up (e.g. `kata-agent cmd-list`).
This way we can list commands without connecting to the agent and remove the `list` command logic from builtin commands.

jodh-intel:
Hi @kalil-pelissier - Thanks. Feel free to add another commit to move the `list` command up on this PR or raise a new PR for that (I'm happy with either ;)

kalil-pelissier:
Hi @jodh-intel, All done.

jodh-intel:
Thanks @kalil-pelissier ! still lgtm so let's see if we can get another review ack so we can land this...

Apokleos:
Hi @kalil-pelissier Thx for your contribution!
Please follow the [CONTRIBUTING/patch-format](https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format) and add the `Fixes: #XXXX` line in the changelog of the commit as it is a mandatory CI test. CI error as below:
```...
Error: 

  dc762[20](https://github.com/kata-containers/kata-containers/actions/runs/7449814456/job/20268849225?pr=8689#step:7:21)753b39afa47890d1b1dd8b069eb3b2e36    No "Fixes" found
  24fcaea95e427d702975152e986513feec956197    No "Fixes" found
  c12bf7523ca00c61d3926c17488f2717f61da9f0    No "Fixes" found
  8feca72924fb9d761e51d8eded135b6c8113058b    No "Fixes" found
  fecf3d2494c2c0599133fd62f948e411ada6bb62    No "Fixes" found
  95c795d4992a4f336099f29d9c5ca8e560484be5    No "Fixes" found
  81848e30cabfef76961e5dda81489e3f14f56864    No "Fixes" found
  4ef1c8d0472fcbd8f32c8b074ccd745e9[22](https://github.com/kata-containers/kata-containers/actions/runs/7449814456/job/20268849225?pr=8689#step:7:23)ca2ad    No "Fixes" found
```

src/tools/agent-ctl/src/client/command.rs
```diff
@@ -1,520 +1,1710 @@
...
+    ) -> (Result<()>, bool) {
+        let req: CheckRequest = match utils::make_request(args) {
+            Ok(res) => res,
+            Err(e) => {
+                return (
+                    Err(anyhow!("{:?}", e).context(REQUEST_BUILD_FAIL_MESSAGE)),
+                    false,
+                )
+            }
+        };
```
Apokleos, line 301: lots of duplicated codes found in the commits. could you refactor it ?

kalil-pelissier:
Hi @Apokleos thanks for the feedback.
I tried to refactor the `make_request()` function a bit so the anywho Error creation is no longer in the `command` submodule.
I didn't want to change the return type of `make_request()` otherwise it would strongly couple the function to the `command` submodule.

What do you think?

src/tools/agent-ctl/src/client/command.rs
kalil-pelissier: Update!
Were you only talking about the `make_request()` function call?

Apokleos: No, IMO, maybe something like `utils::make_request(args).context(REQUEST_BUILD_FAIL_MESSAGE)?` is enough.

src/tools/agent-ctl/src/client.rs
```diff
@@ -777,39 +765,39 @@ fn handle_cmd(
...
+
+//     let result = f(ctx, client, health, options, args);
+//     if result.is_err() {
+//         return (result, false);
+//     }
+
+//     let shutdown = cmd.eq(SHUTDOWN_CMD);
+
+//     (Ok(()), shutdown)
+// }
```
Apokleos: So many lines like these in the file client.rs commented out, why? draft？

Apokleos (commented):
Thx some comments !

kalil-pelissier:
Hi @Apokleos,
Do you have some time to review this PR?

Issue #6202, opened 2023-02-02
runtime-rs: ch: Implement Persist trait for Cloud Hypervisor


jodh-intel:
The [`Persist` trait](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/persist/src/sandbox_persist.rs) needs to be implemented for CH.

Related: #6263.

jodh-intel:
This might be a good task for the outreach project. Does depend on #6201 landing though.

jodh-intel:
We also need to implement the `save_vm()` method in https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs.

PR #8704, opened 2023-12-20
runtime-rs: implement persist api for cloud-hypervisor

amshinde:
Store/Restore additional fields to/from hypervisor state for supporting Persist API.

Fixes: #6202


amshinde:
cc @jodh-intel

src/runtime-rs/crates/hypervisor/src/ch/inner.rs
```diff
@@ -156,12 +157,34 @@ impl Persist for CloudHypervisorInner {
         _hypervisor_args: Self::ConstructorArgs,
         hypervisor_state: Self::State,
     ) -> Result<Self> {
+        let mut capabilities = Capabilities::new();
+        capabilities.set(
+            CapabilityBits::BlockDeviceSupport
+                | CapabilityBits::BlockDeviceHotplugSupport
+                | CapabilityBits::FsSharingSupport
+                | CapabilityBits::HybridVsockSupport,
+        );
```
jodh-intel, line 166: This duplicates the code we have in `new()` so can we either create a new caps setup method to be called from `new()` and `restore()`, or somehow extract it from the args?

jodh-intel (commented):
Thanks @amshinde - I've added a comment. Also, could you add some unit test for `save()` and `restore()` now they are returning the full objects?

src/runtime-rs/crates/hypervisor/src/ch/inner.rs
amshinde: @jodh-intel,  there is an existing capabilities() function for CloudHypervisorInner. I made use of that instead. Which also makes me wonder if we really want to have a field called _capabilites at all. Would be just better to make use of the function instead. Wdyt? Could be addressed in a separate PR of course as the change would be unrelated.

jodh-intel: > @jodh-intel, there is an existing capabilities() function for CloudHypervisorInner. I made use of that instead.

+1.

> Which also makes me wonder if we really want to have a field called _capabilites at all. Would be just better to make use of the function instead. Wdyt? Could be addressed in a separate PR of course as the change would be unrelated.

I agree. And I also spotted that we're setting caps in `CloudHypervisorInner::new()`, but that code could also just call `self.capabilities()` to avoid the duplication.

amshinde:
> Also, could you add some unit test for `save()` and `restore()`

@jodh-intel Done

jodh-intel (approved):
Thanks @amshinde.

amshinde:
@Apokleos Can you take a look at this PR?

amshinde:
@cmaf Can you take a look?

src/libs/kata-sys-util/src/protection.rs
```diff
@@ -12,21 +12,22 @@ use std::fmt;
 use std::path::Path;
 use std::path::PathBuf;
 use thiserror::Error;
+use serde::{Deserialize, Serialize};
```
mkbhanda, line 15: Could lines 12 and 13 be combined as "use std::path::{Path, PathBuf}?

src/runtime-rs/crates/hypervisor/src/ch/inner.rs
```diff
@@ -143,10 +143,11 @@ impl Persist for CloudHypervisorInner {
             vm_path: self.vm_path.clone(),
             jailed: false,
             jailer_root: String::default(),
-            netns: None,
+            netns: self.netns.clone(),
             config: self.hypervisor_config(),
             run_dir: self.run_dir.clone(),
-            cached_block_devices: Default::default(),
```
cmaf, line 149: @amshinde I am assuming this variable is unrelated to this PR and not used?

dborquez (approved):
LGTM, thank you @amshinde

src/libs/kata-sys-util/src/protection.rs
amshinde: @mkbhanda Yes it definitely could. But since that is an unrelated change I prefer not to include it here, maybe in a separate change.

jodh-intel:
@amshinde - could you rebase so we can get this one landed?

amshinde:
> @amshinde - could you rebase so we can get this one landed?

Done

PR #8709, opened 2023-12-20
kata-deploy: Update `jq` as part of the kata-deploy daemonset

fidencio:
This is a follow-up from https://github.com/kata-containers/kata-containers/pull/8679#issuecomment-1864556280, where we've narrowed down the correct component that was causing the issue, and we can go ahead and unblock the snapshotter related work (and, consequently, the merge to main work).


stevenhorsman (approved):
LGTM. Thanks for digging in and finding the 'proper' fix!

danmihai1 (approved):

gkurz (approved):
Much better indeed. Thanks @fidencio !

Issue #6438, opened 2023-03-09
runtime-rs: ch: Implement minimal thread/tid/pid handling


jodh-intel:
Implement the following methods in https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs:

- `get_thread_ids()`
- `get_pids()`
- ~`get_vmm_master_tid()`~ (already done)

Seemingly related: #6429

joannejchen:
Hi @jodh-intel! @scotran and I are part of a Virtualization course at UT Austin, and as part of a project we have to contribute to open source repositories. Can we take up this issue? If so, where can we look for resources/more information on intended behavior?

jodh-intel:
Hi @joannejchen - That would be great! I am actually out for a week now, but what you'll need to do is:

- [Install Kata](https://github.com/kata-containers/kata-containers/tree/main/docs/install)
  For development/test purposes only, the simplest way to do that is using the "automatic" method as it's a "one liner":
  
  - https://github.com/kata-containers/kata-containers/blob/main/utils/README.md#to-install-kata-containers-and-containerd
 
- Change the default [configuration](https://github.com/kata-containers/kata-containers/#configuration) to use Cloud Hypervisor (rather than the default QEMU [hypervisor](https://github.com/kata-containers/kata-containers/blob/main/docs/hypervisors.md)):
  ```bash
  $ sudo mkdir -p /etc/kata-containers/
  $ sudo cp /opt/kata/share/defaults/kata-containers/configuration-clh.toml /etc/kata-containers/configuration.toml
  ```
- Enable the [Cloud Hypervisor feature](https://github.com/kata-containers/kata-containers/pull/6201#issuecomment-1413690515) - you just need to change one `Cargo.toml` file as described in the link.

- Build the new rust runtime with that feature enabled
  ```bash
  $ git clone https://github.com/kata-containers/kata-containers
  $ cd kata-containers/src/runtime-rs
  $ make && sudo make install
  ```

- Ensure the version of the runtime you just built is found in the `$PATH` variable first:
  ```bash
  $ which containerd-shim-kata-v2
  /usr/local/bin/containerd-shim-kata-v2
  ```
- Restart containerd to pick up the new runtime version:
  ```bash
  $ sudo systemctl restart containerd
  ```
- Create a new Kata container and start hacking ;)

[...]

PR #8710, opened 2023-12-20
runtime-rs: ch: Implement minimal implementation for missing thread/pid APIs

jodh-intel:
Add implementations for the following `Hypervisor` trait methods which simply return the same details as the `get_vmm_master_tid()` method:

- `get_thread_ids()`
- `get_pids()`

Fixes: #6438.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
@@ -28,6 +28,7 @@ use lazy_static::lazy_static;
 use nix::sched::{setns, CloneFlags};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fs::create_dir_all;
 use std::os::unix::io::AsRawFd;
@@ -676,7 +677,16 @@ impl CloudHypervisorInner {
     }
 
     pub(crate) async fn get_thread_ids(&self) -> Result<VcpuThreadIds> {
-        Ok(VcpuThreadIds::default())
+        let mut vcpus = HashMap::new();
+
+        let vcpu = 0;
+        let thread_id = self.get_vmm_master_tid().await?;
+
+        vcpus.insert(vcpu, thread_id);
+
+        let vcpu_thread_ids = VcpuThreadIds { vcpus };
+
+        Ok(vcpu_thread_ids)
     }
 
     pub(crate) async fn cleanup(&self) -> Result<()> {
@@ -688,7 +698,9 @@ impl CloudHypervisorInner {
     }
 
     pub(crate) async fn get_pids(&self) -> Result<Vec<u32>> {
-        Ok(Vec::<u32>::new())
+        let pid = self.get_vmm_master_tid().await?;
+
+        Ok(vec![pid])
     }
 
     pub(crate) async fn get_vmm_master_tid(&self) -> Result<u32> {


src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -676,7 +677,16 @@ impl CloudHypervisorInner {
     }
 
     pub(crate) async fn get_thread_ids(&self) -> Result<VcpuThreadIds> {
-        Ok(VcpuThreadIds::default())
+        let mut vcpus = HashMap::new();
+
+        let vcpu = 0;
```
cmaf, line 682: @jodh-intel is this always 0?

jodh-intel: @cmaf - well, I'm not sure what else we could do here. Thoughts?

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -688,7 +698,9 @@ impl CloudHypervisorInner {
     }
 
     pub(crate) async fn get_pids(&self) -> Result<Vec<u32>> {
-        Ok(Vec::<u32>::new())
+        let pid = self.get_vmm_master_tid().await?;
```
amshinde, line 701: @jodh-intel We should probably return the virtio-fs deamon pid here as well.

amshinde: I just checked the golang implementation for the `GetPids` API. 
Its interesting that the qemu implementation returns both hypervisor+virtiofs pids, whereas clh just returns the hypervisor pid.
Point to note is that golang has a separate API for the virtiofs deamon:
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor.go#L1108

This is missing from current rust hypervisor plugin API. It would not apply to dragonball anyway as it does not run a separate virtio-fs deamon.

Also, dragonball returns all the child pids in its implementation:
https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/src/dragonball/inner_hypervisor.rs#L110

This should also include virtio-fs thread id.

So in conclusion, I feel that with runtime-rs+clh, we should return hypervisor_pid + virtiofs_pid as well.

FYI, the only place I see the golang GetPids api being called is for metrics:
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/sandbox_metrics.go#L135

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
amshinde: @jodh-intel This api is meant to return all the tids corresponding to all the vcpus that the VM uses. 

qemu has an qmp api `query-cpus-fast` that returns the details about the vcpus: https://www.qemu.org/docs/master/system/cpu-hotplug.html

I am not sure if such an api exists or has been added to clh since the go implementation. 
cc @likebreath  Can you let us know about this?

Else we may have to follow the same method that we used for the go-implementation i.e walk through the /proc/pid  tree:
https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/clh.go#L760

likebreath: > qemu has an qmp api `query-cpus-fast` that returns the details about the vcpus: https://www.qemu.org/docs/master/system/cpu-hotplug.html
> 
> I am not sure if such an api exists or has been added to clh since the go implementation. cc @likebreath Can you let us know about this?

@amshinde is right. Cloud Hypervisor does not provide such pid information at this time. Of course, please feel free to send feature requests to the CLH community.

One clarification questions: what kind of PID information does kata want from VMM?

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
jodh-intel: Hi @amshinde - I'm not sure that's needed since with the golang runtime, the CH driver spawns the virtiofsd but with runtime-rs, the runtime itself does that, not the CH driver.

amshinde: Ok, it makes sense to not include virtio-fs then

jodh-intel:
This PR is a minimal implementation. We can't return the full details since CH doesn't provide that information currently. I've raised https://github.com/cloud-hypervisor/cloud-hypervisor/issues/6103 and https://github.com/kata-containers/kata-containers/issues/8799 for the "phase 2" implementation :smile: 

/cc @amshinde, @cmaf

amshinde (approved):
@jodh-intel I am approving this, lets open a separate PR to handle multiple VCPUs in a separate PR similar to the golang implementation.

cmaf (approved):
thanks @jodh-intel

PR #8714, opened 2023-12-21
tests: Use function from Kata repo

### LINKED ISSUE #8713: Change get_pr_changed_file_details_full to use function from kata containers repo
Switch that function to use one from the kata containers repo instead of tests to reduce dependency on tests repo.

cmaf:
Switch to use function from Kata repo in common.bash to reduce dependency on the tests repo.

Fixes #8713

### DIFF
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -57,11 +57,10 @@ run_docs_url_alive_check()
 
 run_get_pr_changed_file_details()
 {
-	clone_tests_repo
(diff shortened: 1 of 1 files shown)


gkurz (approved):
LGTM. Thanks @cmaf !

clone_tests_repo-- :heart:

fidencio (approved):
lgtm, thanks @cmaf

Issue #8717, opened 2023-12-22
tests: Vhost modules aren't loaded


justxuewei:
[Nydus tests](https://github.com/kata-containers/kata-containers/actions/runs/7287159798/job/19881687409?pr=8503) complain that vhost* modules weren't loaded. 

```
 Dec 22 02:14:42.617 INFO check CPU: x86_64
 Dec 22 02:14:42.617 INFO check kernel modules for: x86_64
 Dec 22 02:14:42.622 INFO kvm Ok
 Dec 22 02:14:42.626 INFO kvm_intel Ok
 Dec 22 02:14:42.630 WARN The kernel module `vhost` exist but is not already loaded. Try reloading it using 'modprobe vhost'
 Dec 22 02:14:42.633 WARN The kernel module `vhost_net` exist but is not already loaded. Try reloading it using 'modprobe vhost_net'
 Dec 22 02:14:42.635 WARN The kernel module `vhost_vsock` exist but is not already loaded. Try reloading it using 'modprobe vhost_vsock'
```

The default network backend of runtime-rs with Dragonball is vhost-net after https://github.com/kata-containers/kata-containers/pull/8609 merged. Loading the modules explicitly is required.

PR #8718, opened 2023-12-22
tests: Load vhost modules explicitly while Kata installing

justxuewei:
The default network backend of runtime-rs with Dragonball is vhost-net
after #8609 merged. The tests might be failed if vhost modules are not
loaded.

Fixes: #8717


studychao (approved):
thanks, lgtm

lifupan (approved):

PR #8731, opened 2023-12-26
runtime-rs:enable static_sandbox_resource_mgmt

Tim-0731-Hzt:
enable static_sandbox_resource_mgmt

Fixes:#8730


src/runtime-rs/Makefile
```diff
@@ -167,7 +167,7 @@ DEFVFIOMODE := guest-kernel
 ##VAR DEFSANDBOXCGROUPONLY=<bool> Default cgroup model
 DEFSANDBOXCGROUPONLY ?= false
 DEFSTATICRESOURCEMGMT ?= false
```
Apokleos, line 169: CLH depends on the `DEFSTATICRESOURCEMGMT`. 
IMO, I'm afraid that CLH with the configuration will not work well, if it's set with false.

Tim-0731-Hzt: fixed

fidencio: Just for the sake of making things clear.

Cloud Hypervisor itself could be set to `false` without any issues, as Cloud Hypervisor supports memory / cpu hotplug.  However, when we deal with TEEs (TDX, in our case), we need to have it set to `true` as TDX doesn't support memory / cpu hotplug.

For now, TDX is not supported on Cloud Hypervisor, and even when TDX will be supported, we'll need a new configuration file for it.  So, it can be set to `false`  here without issues.

Tim-0731-Hzt: fixed

Apokleos (approved):
Thx @Tim-0731-Hzt  LGTM!

src/runtime-rs/Makefile
amshinde: cc @jodh-intel

jodh-intel: @Tim-0731-Hzt - To avoid breaking Cloud Hypervisor, how about adding a `CLH_DEFSTATICRESOURCEMGMT` option?

Apokleos: Hi @jodh-intel @amshinde Could you please check if the kata pod running ok with CLH(runtime-rs) when the `static_resource_mgmt` is set the default `false`? I'm afraid that it won't work well with the un-implement of `resize_memory` in CLH.

Apokleos: CC @Tim-0731-Hzt [related issue](https://github.com/kata-containers/kata-containers/issues/8784)

Ji-Xinyou:
For VMM that supports hotplug, setting this true will prevent the VMM from hotplugging the CPU after the VM boots.

Issue #8732, opened 2023-12-26
kata-deploy: Fix shim check for snapshotter configuration


fidencio:
We want to check whether the shim is part of the "plain text" shims
passed to the daemonset (meaning, checking against `$SHIMS`).  Before
this fix we were checking against `$shims`, which is an array of shims
instead of a string, resulting on a broken check.

PR #8733, opened 2023-12-26
kata-deploy: Fix shim check for snapshotter configuration

fidencio:
We want to check whether the shim is part of the "plain text" shims passed to the daemonset (meaning, checking against `$SHIMS`).  Before this fix we were checking against `$shims`, which is an array of shims instead of a string, resulting on a broken check.

Fixes: #8732


danmihai1 (approved):

ChengyuZhu6:
LGTM, thanks @fidencio

ChengyuZhu6 (approved):
LGTM, thanks @fidencio

justxuewei (approved):
Lgtm, thanks!

PR #8753, opened 2023-12-29
TEEs: Introduce kernel-confidential

fidencio:
In order to reduce the amount of artefacts we're shipping, let's try to introduce a single kernel and QEMU that would be capable of supporting SEV / SNP / TDX.

This series only introduce such artefacts, but those are not used yet.

We'll first have those ones built, and after that start testing it as part of the CI.  This gives us time to solve the redness of the AMD CIs, and the switch to a new host side stack on the Intel CI,  after folks are back from vacation.


fitzthum:
A few high-level notes. 

First, SNP support is not upstream in QEMU yet. I think the most recent patches are [here](https://github.com/AMDESE/qemu/tree/snp-latest). So we will probably need to continue shipping SNP QEMU for a while.

Second, for SEV(-ES) to support attestation. we need a kernel module to be built into the rootfs. This is sort of related to this work because the kernel module has to be compatible with the guest kernel. In CCv0 we have [this code](https://github.com/kata-containers/kata-containers/blob/CCv0/tools/packaging/guest-image/build_image.sh#L41) to tell the rootfs builder about the kernel module. We assume that the module has already been built when we build the SEV kernel. In theory we can do something very similar with the generic kernel. I don't think this code for the module has made it into main yet. It might need to be added as part of @ryansavino's [shim patch](https://github.com/kata-containers/kata-containers/pull/7517).

I can take a closer look at this shortly and test out the SEV components at some point.

fidencio:
> So we will probably need to continue shipping SNP QEMU for a while.

I'll drop the QEMU part of this patch then.  Thanks for letting me know!

fidencio:
@fitzthum, could you give it a try on this latest update?

In order to generate a rootfs in the same way you're used to do for SEV, please, apply also this patch (which is out of context of this series, but will come in a different one after this one gets merged).

```
⋊> kata-containers on topic/add-confidential-artefacts ≡ ⨯ git diff
diff --git a/tools/packaging/kata-deploy/local-build/Makefile b/tools/packaging/kata-deploy/local-build/Makefile
index 21ad3626d..5a863bc86 100644
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -153,7 +153,7 @@ rootfs-image-tdx-tarball: kernel-tdx-experimental-tarball
 rootfs-initrd-mariner-tarball:
        ${MAKE} $@-build
 
-rootfs-initrd-sev-tarball: kernel-sev-tarball
+rootfs-initrd-sev-tarball: kernel-confidential-tarball
        ${MAKE} $@-build
 
 rootfs-initrd-tarball:
```

tools/packaging/kernel/configs/fragments/common/fs.conf
```diff
@@ -31,7 +31,6 @@ CONFIG_FSNOTIFY=y
 CONFIG_DNOTIFY=y
 CONFIG_INOTIFY_USER=y
 CONFIG_FANOTIFY=y
-CONFIG_AUTOFS4_FS=y
```
gkurz, line 34: Nit: it seems CONFIG_AUTOFS4_FS should be added to the commit changelog along with the other ones for consistency.

tools/packaging/kernel/configs/fragments/common/base.conf
```diff
@@ -4,13 +4,6 @@ CONFIG_SECTION_MISMATCH_WARN_ONLY=y
 CONFIG_SMP=y
 # Note, no nested VM support enabled here
 
-# Turn off embedded mode, as it disabled 'too much', and we
-# no longer pass all the tests. We should refine this, and
-# work out which of the ~66 items it enables are really needed.
-# I believe this is the actual syntax we need for a fragment to
-# disable an item...
-# CONFIG_EMBEDDED is not set
```
gkurz, line 12: I guess it is fine to not mention CONFIG_EMBEDDED in the changelog since it wasn't set.

gkurz: Note that this was removed in kernel 6.6, not 6.5.

jepio (approved):
i like where this is headed. lgtm

tools/packaging/kernel/build-kernel.sh
```diff
@@ -110,7 +110,7 @@ Options:
 	-t <hypervisor>	: Hypervisor_target.
 	-u <url>	: Kernel URL to be used to download the kernel tarball.
 	-v <version>	: Kernel version to use if kernel path not provided.
-	-x <type>	: Confidential guest protection type, such as sev, snp and tdx
+	-x <type>	: Confidential guest protection type, such as sev, snp, tdx. or "confidential" (for all of those).
```
stevenhorsman, line 113: minor nit: switch full-stop for comma (or just remove entirely if you aren't a fan of the Oxford comma)? I'm also very slightly triggered by the change in alignment for url and type, but that wasn't added in this PR, so I can get over it easily!
```suggestion
	-x <type>	: Confidential guest protection type, such as sev, snp, tdx, or "confidential" (for all of those).
```

tools/packaging/kernel/build-kernel.sh
```diff
@@ -152,8 +152,8 @@ get_tee_kernel() {
 	local kernel_tarball="${version}.tar.gz"
 
 	# Depending on where we're getting the terball from it may have a
```
stevenhorsman, line 154: Minor nit: You could fix `terball` typo whilst in here?

tools/packaging/kernel/build-kernel.sh
```diff
@@ -152,8 +152,8 @@ get_tee_kernel() {
...
-	# different name, such as linux-${version}.tar.gz or simply
-	# ${version}.tar.gz.  Let's try both before failing.
+	# different name, such as linux-${version}.tar.gz, ${version}.tar.gz,
+	# or even v${version}.tar.gz.  Let's try all of them before failing.
 	curl --fail -L "${kernel_url}/linux-${kernel_tarball}" -o ${kernel_tarball} || curl --fail -OL "${kernel_url}/${kernel_tarball}"
```
stevenhorsman, line 157: Should this have `|| curl --fail -OL "${kernel_url}/v${kernel_tarball}"` added, to try the or `v${version}.tar.gz` approach, or is the comment wrong?

stevenhorsman (approved):
Just left some minor comments/suggestions, but otherwise it looks good. Thanks!

gkurz (approved):
LGTM. Thanks @fidencio !

fidencio:
I've got all your comments addressed, thanks a lot for your review!

Issue #8758, opened 2024-01-02
mounts: Too many mount points in k8s high-density deployment with Kata go runtime


fadecoder:
Recently, we notice one problem when deploying Kata Containers go runtime through k8s. When we deploy over 400 Pods on one node, the number of mount points on the system will exceed 10,000, no matter what VMM(qemu, clh, stratovirt) we use. 

A large number of mount points will cause the `systemd` process in the `host` to be very busy (100% CPU load), executing system calls such as `open`/`read` the mounts, due to some kind of container health check mechanism. The high load of the `systemd` process will cause the deployment of more Pods to become extremely slow, making it difficult to meet the requirements for fast and high-density deployment.

Inspired by rust runtime, we add the mount namespace isolation to go runtime, and it can prevent the `systemd` process from seeing more mounts, thereby reducing the burden on the `systemd`, making it faster to deploy more Pods.

Considering the runtime-rs VMMs are under heavy developments, adding such isolation is still useful for many user-cases.

WenyuanLau:
Thanks! @fadecoder

DeciderWill:
thanks @fadecoder this has been affecting us for a while 🎉

PR #8760, opened 2024-01-02
runtime: Reduce the mount points with namespace isolation

fadecoder:
We notice a 100% CPU load `systemd` process in the `host` when deploying more than 400 Kata Containers Pods (using go runtime) through k8s in one node, as described in issue #8758 .

Inspired by rust runtime, we add the mount namespace isolation to go runtime, and it can prevent the `systemd` process from seeing more mounts, thereby reducing the burden on the `systemd`, making it faster to deploy more Pods.

Considering the runtime-rs VMMs are under heavy developments, adding such isolation is still useful for many user-cases.

Fixes: #8758

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/service.go
+++ b/src/runtime/pkg/containerd-shim-v2/service.go
@@ -191,6 +191,27 @@ func newCommand(ctx context.Context, id, containerdBinary, containerdAddress str
 	return cmd, nil
 }
 
+func setupMntNs() error {
+	err := unix.Unshare(unix.CLONE_NEWNS)
+	if err != nil {
+		return err
+	}
+
+	err = unix.Mount("", "/", "", unix.MS_REC|unix.MS_SLAVE, "")
+	if err != nil {
+		err = fmt.Errorf("failed to mount with slave: %v", err)
+		return err
+	}
+
+	err = unix.Mount("", "/", "", unix.MS_REC|unix.MS_SHARED, "")
+	if err != nil {
+		err = fmt.Errorf("failed to mount with shared: %v", err)
+		return err
+	}
+
+	return nil
+}
+
 // StartShim is a binary call that starts a kata shimv2 service which will
 // implement the ShimV2 APIs such as create/start/update etc containers.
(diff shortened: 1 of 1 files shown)


WenyuanLau:
Thanks! @fadecoder

WenyuanLau (approved):
LGTM!

zvonkok:
Interesting, I would like to see how OpenRC behaves here ... and if we can save some CPU cycles using OpenRC vs. systemd. @fidencio FYI https://github.com/kata-containers/kata-containers/pull/8760

WenyuanLau:
> Interesting, I would like to see how OpenRC behaves here ... and if we can save some CPU cycles using OpenRC vs. systemd. @fidencio FYI #8760

The `systemd` process mentioned in the PR is in the `host machine`, not in the `VM`. I'm not quite sure that we are talking about the same process. :)

gkurz (approved):
Allowing untrusted users to hog the host CPU and add measurable latency in a critical component like the host systemd is a security issue. Maybe not critical but severe enough to waste precious cycles and $$ on a big system.

Adding the missing isolation not only helps with the k8s deployment density, but it
is true to kata's `provide the workload isolation and security advantages of VMs.` motto that all users legitimately expect.

Thanks @fadecoder !

fidencio:
@fadecoder, bear with me here in case I'm wrong, but I think I understand why CRI-O tests are failing.

Creating a new mount namespace requires that the process creating it has `CAP_SYS_ADMIN` capabilities.  On the CRI-O side, that's not part of the default capabilites, and you'd have to expand that by adding `CAP_SYS_ADMIN` to https://github.com/kata-containers/kata-containers/blob/0dc00ae373de03169c18baf0069e4f8370891a9c/tests/common.bash#L590-L601

That will make the CI pass, most likely.

Whether or not giving this set of permissions is a good thing, is not exactly up for discussion here (mainly as we're already allowing on `DAC_OVERRIDE`). But, again, this discussion is out of the scope of this PR.

lifupan (approved):

gkurz:
This PR needs to be rebased to main in order to get rid of the bogus pending build checks in CI.

WenyuanLau:
CI have passed, I will get it merged. Thanks! @fadecoder

Issue #8125, opened 2023-10-05
gha: Add backport label is failing


GabyCT:
Seems like the add backport label is failing on multiple PRs with the next message
```
Run pr=8119
need command: hub
Error: No linked issues for PR 8119
Error: Process completed with exit code 1.
```

stevenhorsman:
I'll take a look at this and see if I can port the functionality over to use the github cli now that hub is dropped.

PR #8763, opened 2024-01-03
Fix backport check hub

stevenhorsman:
Add gh-util.sh script
    
- The hub tool is now deprecated, so introduce a new alternative to `hub-util.sh`
    https://github.com/kata-containers/.github/blob/main/scripts/hub-util.sh
    that works with it.
    Initially I've only started with the couple of commands that we use regularly, but we can extend it in future.
- Expects jq to be installed and `gh` to be installed an setup (see https://docs.github.com/en/github-cli/github-cli/quickstart#prerequisites    workflows: Update backport-label to use gh-utils.sh)
- Now we don't have lots of repos, I've moved it into `kata-containers` rather than `.github`,
   so it is more visible.
- Update workflow to use new script
    
Fixes: #8125

### DIFF
--- a/.github/workflows/add-backport-label.yaml
+++ b/.github/workflows/add-backport-label.yaml
@@ -19,18 +19,11 @@ jobs:
     if: ${{ github.event.label.name != 'auto-backport' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code to allow hub to communicate with the project
+      - name: Checkout code to get gh-utils script
         if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') }}
         uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


GabyCT:
seems like a fix needs to be done
```
[static-checks.sh:1153] INFO: Checking script file 'ci/gh-util.sh'
[static-checks.sh:935] ERROR: Expected 'EOF' here anchor, in ci/gh-util.sh found: EOT
```

stevenhorsman:
> seems like a fix needs to be done

Thanks for flagging that up. Sorry I missed it, but should be okay now...

fidencio (approved):
lgtm, thanks @stevenhorsman!

GabyCT (approved):

Issue #8770, opened 2024-01-04
Dragonball: fix unit test problems in dbs_pci / dbs_util / dbs_virtio_devices


studychao:
As reported in https://github.com/kata-containers/kata-containers/issues/8767, we have found that the root cause is that rust-vmm's vmm-sys-utils introduce a new release `0.12.1` and dbs-pci rely on rust-vmm's `vfio-ioctls` which uses `>=` to declare `vmm-sys-utils` so it automatically upgrade `vmm-sys-utils` to `0.12.1`. That's how two different versions of `vmm-sys-utils` is introduced and this breaks the compilation.

In order to fix this and also avoid future problems, we introduce Cargo.lock file to dbs crates.

Issue #8767, opened 2024-01-03
new src/dragonball/src/dbs_pci CI/build breaks


danmihai1:
Looks like it ends up trying to use two versions of vmm-sys-util, that are incompatible with each other.

Example:

https://github.com/kata-containers/kata-containers/actions/runs/7394173738/job/20127819265?pr=8759

```
Compiling dbs-pci v0.1.0 (/home/runner/actions-runner/_work/kata-containers/kata-containers/src/dragonball/src/dbs_pci)
error[E0308]: mismatched types
   --> src/vfio.rs:499:36
    |
492 |                     irqfds.push(fd)
    |                     ------      -- this argument has type `&vmm_sys_util::linux::eventfd::EventFd`...
    |                     |
    |                     ... which causes `irqfds` to have type `Vec<&vmm_sys_util::linux::eventfd::EventFd>`
...
499 |                 .enable_irq(index, irqfds)
    |                  ----------        ^^^^^^ expected `EventFd`, found a different `EventFd`
    |                  |
    |                  arguments to this method are incorrect
    |
    = note: `EventFd` and `EventFd` have similar names, but are actually distinct types
note: `EventFd` is defined in crate `vmm_sys_util`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/vmm-sys-util-0.11.2/src/linux/eventfd.rs:22:1
    |
22  | pub struct EventFd {
    | ^^^^^^^^^^^^^^^^^^
note: `EventFd` is defined in crate `vmm_sys_util`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/vmm-sys-util-0.12.1/src/linux/eventfd.rs:22:1
    |
22  | pub struct EventFd {
    | ^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `vmm_sys_util` are being used?
note: method defined here
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/vfio-ioctls-0.1.0/src/vfio_device.rs:818:12
    |
818 |     pub fn enable_irq(&self, irq_index: u32, event_fds: Vec<&EventFd>) -> Result<()> {
    |            ^^^^^^^^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `dbs-pci` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `dbs-pci` (lib test) due to previous error
[...]

PR #8771, opened 2024-01-04
dbs-pci: introduce Cargo.lock to prevent the influence from upstream

studychao:
As reported in #8767, we have found that the root cause is that rust-vmm's vmm-sys-utils introduce a new release 0.12.1 and dbs-pci rely on rust-vmm's vfio-ioctls which uses >= to declare vmm-sys-utils so it automatically upgrade vmm-sys-utils to 0.12.1. That's how two different versions of vmm-sys-utils is introduced and this breaks the compilation.

In order to fix this and also avoid future problems, we introduce Cargo.lock file to dbs crates.

fixes: #8770


justxuewei (approved):
Lgtm, thanks!

danmihai1 (approved):

danmihai1:
I agree, let's try https://github.com/kata-containers/kata-containers/pull/8771 instead.

danmihai1:
Sorry, I closed this one by mistake. Thanks for re-opening it.

Issue #8769, opened 2024-01-04
sporadic kata-containers-ci-on-push / run-k8s-tests-on-aks / run-k8s-tests (cbl-mariner, clh) failures


danmihai1:
I have seen these tests failing every once in a while - apparently running out of memory. Most recently in https://github.com/kata-containers/kata-containers/actions/runs/7402656467/job/20145920312?pr=8768 👍 

```
[run_kubernetes_tests.sh:138] INFO: Test actions that must be successful
1..1
ok 1 Test unencrypted confidential container launch success and verify that we are running in a secure enclave. # skip Test not supported for clh.
1..1
ok 1 Running with postStart and preStop handlers
1..1
ok 1 Check capabilities of pod
1..1
ok 1 ConfigMap for a pod
1..2
ok 1 Copy file in a pod
ok 2 Copy from pod to host
1..1
ok 1 Check CPU constraints
1..1
ok 1 Credentials using secrets
1..1
ok 1 Check custom dns
1..2
ok 1 Empty dir volumes
ok 2 Empty dir volume when FSGroup is specified with non-root container
1..1
ok 1 Environment variables
1..1
ok 1 Kubectl exec
1..1
ok 1 Test readonly volume for pods
1..1
not ok 1 configmap update works, and preserves symlinks
# (in test file k8s-inotify.bats, line 25)
#   `kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"' failed

...

# Events:
#   Type     Reason     Age   From               Message
#   ----     ------     ----  ----               -------
#   Normal   Scheduled  91s   default-scheduler  Successfully assigned kata-containers-k8s-tests/inotify-configmap-testing to aks-nodepool1-32907564-vmss000000
#   Normal   Pulling    82s   kubelet            Pulling image "quay.io/kata-containers/fsnotify:latest"
#   Normal   Pulled     79s   kubelet            Successfully pulled image "quay.io/kata-containers/fsnotify:latest" in 3.391115011s (3.391140111s including waiting)
#   Normal   Created    79s   kubelet            Created container c1
#   Warning  Failed     72s   kubelet            Error: failed to create containerd task: failed to create shim task: error: Failed to resize memory from 2147483648 to 3221225472: error: Put "http://localhost/api/v1/vm.resize": context deadline exceeded reason:  reason: : unknown
```

PR #8772, opened 2024-01-04
tests: list the current k8s pods

danmihai1:
Log the list of the current pods between tests because these pods might be related to cluster nodes occasionally running out of memory.

Fixes: #8769

### DIFF
--- a/tests/integration/kubernetes/run_kubernetes_tests.sh
+++ b/tests/integration/kubernetes/run_kubernetes_tests.sh
@@ -138,12 +138,15 @@ test_successful_actions() {
 	info "Test actions that must be successful"
 	for K8S_TEST_ENTRY in ${K8S_TEST_UNION[@]}
 	do
+		info "$(kubectl get pods --all-namespaces 2>&1)"
 		bats "${K8S_TEST_ENTRY}"
 	done
 }
 
 run_policy_specific_tests() {
+	info "$(kubectl get pods --all-namespaces 2>&1)"
 	bats k8s-exec-rejected.bats
+	info "$(kubectl get pods --all-namespaces 2>&1)"
 	bats k8s-policy-set-keys.bats
 }
 


stevenhorsman:
Hey @danmihai1 - I might have misunderstood something, but it looks like wait=true the default value from https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#delete ?

stevenhorsman:
> It looks like "delete" returns without waiting for some of the K8s YAML files that I tested. Unfortunately --wait=true doesn't change that behavior. For example, my 3 VMs below are still around after deleting my Deployment.

Thanks a lot for the recreate steps that lets me have a play. I think what I've seen so far is that the problem here is that `kubectl delete` waits for the deployment resource to be deleted, which seems to work, but not the pods:
```
# kubectl delete -f deploy.yaml && kubectl get deployment &&  kubectl get pods
deployment.apps "expose-test-deployment" deleted
No resources found in default namespace.
NAME                                      READY   STATUS        RESTARTS   AGE
expose-test-deployment-5d5f68bb97-2kbkz   1/1     Terminating   0          109s
expose-test-deployment-5d5f68bb97-glpq6   1/1     Terminating   0          109s
expose-test-deployment-5d5f68bb97-mmf5r   1/1     Terminating   0          109s
```

I'll have a play and see if I find a nicer way to wait for the pods as well and if I do, I'll let you know.

stevenhorsman:
So, at first attempt, the only way I can see to wait for the deletion of the pods in a deployment is using the deployment's selector. e.g. the above example has a selector `name: nginx`, so I can do the following:
```
kubectl delete -f deploy.yaml && kubectl wait --for delete pod --selector=name=nginx && kubectl get pods
deployment.apps "expose-test-deployment" deleted
pod/expose-test-deployment-5d5f68bb97-bp6z9 condition met
pod/expose-test-deployment-5d5f68bb97-dlm5f condition met
pod/expose-test-deployment-5d5f68bb97-fscjk condition met
No resources found in default namespace.
```
That's admittedly not super helpful for the generic case, where we don't necessarily know the labels, but from a quick scan through the PR, it looks like most of the cases are directly deleting the pod, which should just wait ok IIUC?

danmihai1:
That sounds very promising, thanks! I will give this a try when I can come back to it - hopefully in the next couple of days.

danmihai1:
It would be nice to see some evidence that Kata CI is dealing with leaked (for short term or longer term) pods, before making relatively intrusive changes to wait for pod termination. I didn't get such evidence here, so maybe my guess about zombie pods running the node out of memory was incorrect.

So, I think we could proceed in one of the following ways:

1. Merge this PR and review the new log output when the K8s tests will run out of memory again in the future.
2. Abandon this PR. (maybe it does more harm than good?!?)

I vote for #1, but I wouldn't mind #2 either.

stevenhorsman:
> So, I think we could proceed in one of the following ways:
> 
> 1. Merge this PR and review the new log output when the K8s tests will run out of memory again in the future.
> 2. Abandon this PR. (maybe it does more harm than good?!?)
> 
> I vote for (1), but I wouldn't mind (2) either.

I feel like getting more information is a good plan if we don't quite understand what's going wrong. I don't think that it seems like an excessive amount and harmful, so I'm okay with it.

stevenhorsman (approved):
LGTM - as discussed in the comments. Thanks

GabyCT (approved):

Issue #8715, opened 2023-12-21
tests: An empty-dir test for Dragonball is fragile


justxuewei:
I've suffered from some weird CI issues recently. I raised a pull request https://github.com/kata-containers/kata-containers/pull/8503 to add support for vhost-user-net devices on Dragonball.

A test from [kata-containers-ci-on-push / run-k8s-tests-on-aks / run-k8s-tests (ubuntu, dragonball, small)](https://github.com/kata-containers/kata-containers/actions/runs/7257151387/job/19772135479?pr=8503#logs), named "Empty dir volume when FSGroup is specified with non-root container", was failed due to OOM. It could be reproduced in everytime it is re-run.

The weird things are:

- [See UPDATE1] This might happen with any PRs. ~~A pull request without this patch is not affected, see: https://github.com/kata-containers/kata-containers/pull/8706.~~
- My patch doesn't modify the code associated with empty dir or filesystem.
- The test passed on my local environment and on AKS with a normal spec.

UPDATE1:

- The test was failed with the newer version of agnhost image, see: https://github.com/kata-containers/kata-containers/pull/8716.

/cc @studychao 

Logs

```
# (from function `waitForProcess' in file ../../common.bash, line 65,
#  in test file k8s-empty-dirs.bats, line 53)
#   `waitForProcess "${wait_time}" "${sleep_time}" "${cmd}"' failed
# [bats-exec-test:55] INFO: k8s configured to use runtimeclass
# pod/sharevol-kata created
# Name:                sharevol-kata
# Namespace:           kata-containers-k8s-tests
# Priority:            0
# Runtime Class Name:  kata
# Service Account:     default
# Node:                aks-nodepool1-20954695-vmss000000/10.224.0.4
# Start Time:          Tue, 19 Dec 2023 05:[41](https://github.com/kata-containers/kata-containers/actions/runs/7257151387/job/19772135479?pr=8503#step:11:42):27 +0000
# Labels:              <none>
# Annotations:         <none>
# Status:              Succeeded
# IP:                  10.2[44](https://github.com/kata-containers/kata-containers/actions/runs/7257151387/job/19772135479?pr=8503#step:11:45).0.21
# IPs:
#   IP:  10.244.0.21
# Containers:
#   mounttest-container:
#     Container ID:  containerd://21f08e371c0846c10205035c94795896ab5f959f9d80eedb2e1ac33ae8812309
#     Image:         registry.k8s.io/e2e-test-images/agnhost:2.21
[...]

PR #8773, opened 2024-01-04
agent: Fix an issue reporting OOM events by mistake

justxuewei:
The agent registers an event fd in `memory.oom_control`. An OOM event is
forwarded to containerd when the event is emitted, regardless of the
content in that file.

I observed content indicating that events should not be forwarded, as shown
below. When `oom_kill` is set to 0, it means no OOM has occurred. Therefore,
it is important to check the content to avoid mistakenly forwarding OOM
events.

```
oom_kill_disable 0
under_oom 0
oom_kill 0
```

Fixes: #8715

### DIFF
--- a/src/agent/rustjail/src/cgroups/notifier.rs
+++ b/src/agent/rustjail/src/cgroups/notifier.rs
@@ -3,18 +3,19 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use anyhow::{anyhow, Context, Result};
-use eventfd::{eventfd, EfdFlags};
-use nix::sys::eventfd;
 use std::fs::{self, File};
 use std::os::unix::io::{AsRawFd, FromRawFd};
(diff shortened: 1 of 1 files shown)


src/agent/rustjail/src/cgroups/notifier.rs
```diff
@@ -165,22 +166,17 @@ async fn register_memory_event(
 
     tokio::spawn(async move {
         loop {
-            let sender = sender.clone();
```
Apokleos, line 168: why the line deleted ?

justxuewei: Cuz `clone()` is unnecessary here.

Apokleos (approved):
Thx @justxuewei

Tim-Zhang (approved):

Issue #7881, opened 2023-09-11
runtime-rs: Support event mechanism


justxuewei:
The Runtime-rs doesn't support event machanism so far. The events include OOM event, task exit event, etc. See: https://github.com/kata-containers/kata-containers/blob/6cd5d83a3712020dacab53c17cb4a4c9466c3fbb/src/runtime/pkg/containerd-shim-v2/service.go#L294-L320

PR #8780, opened 2024-01-06
runtime-rs: Forward events to containerd via ttrpc

justxuewei:
It is a little bit heavy for the runtime-rs to forward events via
containerd CLI, contrast to the ttrpc way. Plus, for runtimes that haven't
this mechanism, e.g. CRI-O, we can't get those events anywhere.

This patch introduces two types of forwarders:

- `ContainerdForwarder`: Acquire ttrpc address from environment variables
  and forward events via ttrpc connection.
- `LogForwarder`: Write event info into logs.

Fixes: #7881

### DIFF
--- a/src/runtime-rs/crates/service/src/event.rs
+++ b/src/runtime-rs/crates/service/src/event.rs
@@ -1,0 +1,156 @@
+// Copyright (c) 2019-2024 Ant Group
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use std::collections::HashMap;
+use std::env;
+use std::sync::Arc;
+
+use anyhow::{anyhow, Context, Result};
+use async_trait::async_trait;
+use common::message::Event;
+use containerd_shim::publisher::RemotePublisher;
+use containerd_shim::util::timestamp;
+use containerd_shim::TtrpcContext;
+use containerd_shim_protos::protobuf::well_known_types::any::Any;
+use containerd_shim_protos::shim::events;
+use containerd_shim_protos::shim_async::Events;
+use ttrpc::MessageHeader;
+
+const REMOTE_FORWARDER: &str = "remote";
+const LOG_FORWARDER: &str = "log";
+
+// Ttrpc address passed from container runtime.
+// For now containerd will pass the address, and CRI-O will not.
+const TTRPC_ADDRESS_ENV: &str = "TTRPC_ADDRESS";
+
+/// Forwarder forwards events to upper runtime.
+#[async_trait]
+pub(crate) trait Forwarder {
+    /// Forward an event to publisher
+    async fn forward(&self, event: Arc<dyn Event + Send + Sync>) -> Result<()>;
+    /// Get forwarder type
+    async fn r#type(&self) -> String;
+}
+
+/// Returns an instance of `ContainerdForwarder` in the case of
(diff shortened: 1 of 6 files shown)


src/runtime-rs/crates/service/src/event.rs
```diff
@@ -0,0 +1,135 @@
...
+use containerd_shim::TtrpcContext;
+use containerd_shim_protos::protobuf::well_known_types::any::Any;
+use containerd_shim_protos::shim::events;
+use containerd_shim_protos::shim_async::Events;
+use ttrpc::MessageHeader;
+
+const REMOTE_FORWARDER: &str = "remote";
+const LOG_FORWARDER: &str = "log";
+
+// Rtrpc address passed from container runtime.
```
quanweiZhou, line 23: typo `Rtrpc`？

src/runtime-rs/crates/service/src/manager.rs
```diff
@@ -159,39 +159,20 @@ impl ServiceManager {
...
-            ..Default::default()
+    async fn new_event_publisher(namespace: &str) -> Result<Box<dyn Forwarder>> {
+        let fwd: Box<dyn Forwarder> = match env::var(TTRPC_ADDRESS_ENV) {
+            Ok(address) => Box::new(
+                ContainerdForwarder::new(namespace, &address)
+                    .await
+                    .context("new containerd forwarder")?,
+            ),
+            Err(_) => Box::new(
+                LogForwarder::new(namespace)
```
quanweiZhou, line 170: I want to discuss if there is no `TTRPC_ADDRESS_ENV` should we use log or rollback to use the containerd binary to send the event?

justxuewei: I think it's unnecessary to do fallback. The reasons are:

1. [Containerd doc](https://github.com/containerd/containerd/blob/main/runtime/v2/README.md#start) implies that this env will be available regardless of its version.
2. Runtime written in Go don't have that fallback either.

src/runtime-rs/crates/service/src/event.rs
justxuewei: Fixed.

src/runtime-rs/crates/service/src/manager.rs
```diff
@@ -159,39 +159,20 @@ impl ServiceManager {
         Ok(())
     }
 
-    async fn send_event(&self, event: Arc<dyn Event>) -> Result<()> {
-        let any = Any {
-            type_url: event.type_url(),
-            value: event.value().context("get event value")?,
-            ..Default::default()
+    async fn new_event_publisher(namespace: &str) -> Result<Box<dyn Forwarder>> {
```
quanweiZhou, line 162: I don't think it's better to move the `new_event_publisher`  to `event.rs` and the `manager.rs` doesn't need to be aware of `ContainerdForwarder` or `LogForwarder`.

justxuewei: Thanks for your suggestion. It has been moved to `event.rs`. PTAL.

quanweiZhou (approved):
Thanks, @justxuewei, LGTM.

Tim-0731-Hzt (approved):

Issue #8738, opened 2023-12-27
runtime-rs: optimize `increase_attach_count` function


studychao:
When the device already has attach point larger than 1, if it more suitable to directly return Err instead of return `true`. 
Also, there is no comment here to describe what `true` or `false` mean in this function.

studychao:
also, when there are multiple containers in the pod, there may be case that attach point larger than 2, we should not return `Err` in that case when we are doing `attach` ops.

Apokleos:
Thx @studychao, the PR to do bugfix is coming!
cc @Tim-0731-Hzt

PR #8782, opened 2024-01-08
bugfix and refactor device increate count

Apokleos:
runtime-rs: bugfix incorrect use of refcount before vfio attach
    
    When there's a pod with multiple containers, there may be case that
    attach point more than 2, we should not return Err in that case when
    we are doing attach ops, but just return Ok.

    runtime-rs: introduce dedicated function do_increase_count/do_decrease_count
    
    Since there are many implementations of reference counting in the
    drivers, all of which have the same implementation, we should try
    to reduce such duplicated code as much as possible. Therefore, a
    new function is introduced to solve the problem of duplicated code.

    runtime-rs: refactor increase_attach_count with do_increase_count/do_decrease_count
    
    Try to reduce duplicated code in increase_attach_count with public
    new function do_increase_count/do_decrease_count.

    Fixes: #8738

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
@@ -20,6 +20,7 @@ use crate::{
     device::{
         pci_path::PciPath,
         topology::{do_add_pcie_endpoint, PCIeTopology},
+        util::do_increase_count,
         Device, DeviceType, PCIeDevice,
     },
     register_pcie_device, unregister_pcie_device, update_pcie_device, Hypervisor as hypervisor,
@@ -522,18 +523,7 @@ impl Device for VfioDevice {
     }
 
     async fn increase_attach_count(&mut self) -> Result<bool> {
-        match self.attach_count {
-            0 => {
-                // do real attach
-                self.attach_count += 1;
-                Ok(false)
-            }
-            std::u64::MAX => Err(anyhow!("device was attached too many times")),
-            _ => {
-                self.attach_count += 1;
-                Ok(true)
-            }
-        }
+        do_increase_count(&mut self.attach_count)
     }
 
     async fn decrease_attach_count(&mut self) -> Result<bool> {
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
@@ -20,7 +20,7 @@ use crate::{
     device::{
         pci_path::PciPath,
         topology::{do_add_pcie_endpoint, PCIeTopology},
-        util::do_increase_count,
+        util::{do_decrease_count, do_increase_count},
         Device, DeviceType, PCIeDevice,
     },
     register_pcie_device, unregister_pcie_device, update_pcie_device, Hypervisor as hypervisor,
@@ -527,18 +527,7 @@ impl Device for VfioDevice {
     }
 
     async fn decrease_attach_count(&mut self) -> Result<bool> {
-        match self.attach_count {
-            0 => Err(anyhow!("detaching a device that wasn't attached")),
-            1 => {
-                // do real wrok
-                self.attach_count -= 1;
-                Ok(false)
-            }
-            _ => {
-                self.attach_count -= 1;
-                Ok(true)
-            }
-        }
+        do_decrease_count(&mut self.attach_count)
     }
 
     async fn get_device_info(&self) -> DeviceType {
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
@@ -456,7 +456,13 @@ impl Device for VfioDevice {
             .await
             .context("failed to increase attach count")?
         {
-            return Err(anyhow!("attach count increased failed as some reason."));
+            warn!(
+                sl!(),
+                "The device {:?} is not allowed to be attached more than one times.",
+                self.device_id
+            );
+
+            return Ok(());
         }
 
         // do add device for vfio deivce
--- a/src/runtime-rs/crates/hypervisor/src/device/util.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/util.rs
@@ -84,10 +84,28 @@ pub fn do_increase_count(ref_count: &mut u64) -> Result<bool> {
     Ok((*ref_count) != 1)
 }
 
+// The return value of do_decrease_count can be used to indicate whether the device is still in use.
+// Specifically, the reference count can be decremented by 1 first, then check the decremented ref_count:
+// If the decremented reference count is not equal to 0, it indicates that the device is still in use by
+// the guest and cannot be detached. Return true.
+// If the reference count is equal to 0, it indicates that the device will not be used and can be unplugged
+// from the guest. Return false.
+pub fn do_decrease_count(ref_count: &mut u64) -> Result<bool> {
+    // ref_count = 0: Device not inserted (cannot decrease further).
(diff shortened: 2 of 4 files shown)


src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -456,7 +457,12 @@ impl Device for VfioDevice {
             .await
             .context("failed to increase attach count")?
         {
-            return Err(anyhow!("attach count increased failed as some reason."));
+            warn!(
+                sl!(),
+                "It's not allowed that device {:?} is attached many times,", self.device_id
```
justxuewei, line 462: I think it is more concise.

```suggestion
                "The device {:?} is not allowed to be attached multiple times.", self.device_id
```

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -69,9 +69,55 @@ pub(crate) fn get_virt_drive_name(mut index: i32) -> Result<String> {
...
+        // this maximum value, the device is theoretically allowed to do so.
+        // However, we will not allow the device to actually be inserted into the Guest more than once.
+        std::u64::MAX => Err(anyhow!("device was attached too many times")),
+        _ => {
+            // When ref_count is greater than 0, it indicates that how many times the device has been attempted to
+            // be inserted into the Guest.
+            // In this case, we know the device has been attached into the Guest.
+            *ref_count += 1;
+            Ok(true)
+        }
```
justxuewei, line 91: How about using `checked_add()`? FYI, https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add.

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -69,9 +69,55 @@ pub(crate) fn get_virt_drive_name(mut index: i32) -> Result<String> {
...
+    }
+}
+
+// do_decrease_count
+pub fn do_decrease_count(ref_count: &mut u64) -> Result<bool> {
```
justxuewei, line 96: Ditto.

justxuewei (commented):
Thanks! A few comments here.

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -69,9 +69,55 @@ pub(crate) fn get_virt_drive_name(mut index: i32) -> Result<String> {
     Ok(String::from(PREFIX) + std::str::from_utf8(&disk_letters)?)
 }
 
+// do_increase_count
```
justxuewei, line 72: Remove it or expand on this, as this comment doesn't provide any valuable information.

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -69,9 +69,55 @@ pub(crate) fn get_virt_drive_name(mut index: i32) -> Result<String> {
...
+            // When ref_count is greater than 0, it indicates that how many times the device has been attempted to
+            // be inserted into the Guest.
+            // In this case, we know the device has been attached into the Guest.
+            *ref_count += 1;
+            Ok(true)
+        }
+    }
+}
+
+// do_decrease_count
```
justxuewei, line 95: Ditto.

src/runtime-rs/crates/hypervisor/src/device/util.rs
Apokleos: Thx @justxuewei Although I know that the `checked_add()` method is a good choice in some cases, I don't think it's appropriate here. And I don't want to introduce too much judgment logic because of this method, so I choose to stick to the original approach.

Another reason is, the code are just moved here from its original places, and the key point is, simple and it works well.

Apokleos: @justxuewei  Maybe saturating_add(1)/saturating_sub(1) is a good choice !

Apokleos: However, It's indeed better to refactor it with check_add as @justxuewei said:
```
checked_add(1).ok_or(anyhow!("error msg")?;
return Ok(*ref_count != 1))
```

justxuewei (approved):
Lgtm, thanks! @Apokleos

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -69,9 +69,31 @@ pub(crate) fn get_virt_drive_name(mut index: i32) -> Result<String> {
     Ok(String::from(PREFIX) + std::str::from_utf8(&disk_letters)?)
 }
 
+pub fn do_increase_count(ref_count: &mut u64) -> Result<bool> {
```
studychao, line 72: maybe add some document here to explain what does `true` or `false` returns means. 
Right now, it seems vague to me.

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -69,9 +69,31 @@ pub(crate) fn get_virt_drive_name(mut index: i32) -> Result<String> {
...
+    // ref_count = 0: Device is new and not attached.
+    // ref_count > 0: Device has been attempted to be attached many times.
+    *ref_count = (*ref_count)
+        .checked_add(1)
+        .ok_or(anyhow!("device reference count overflow"))?;
+
+    Ok((*ref_count) != 1)
+}
+
+pub fn do_decrease_count(ref_count: &mut u64) -> Result<bool> {
```
studychao, line 82: ditto.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
```diff
@@ -456,7 +457,13 @@ impl Device for VfioDevice {
             .await
             .context("failed to increase attach count")?
         {
-            return Err(anyhow!("attach count increased failed as some reason."));
+            warn!(
```
studychao, line 460: Could we detect the case that the device is attached more than once due to multiple containers?
Right now this warn doesn't tell users why multiple time attachment happen and this seems like a bug warn to users.

studychao (commented):
Thanks, a few comments.

src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
Apokleos: First of all, we need ensure the device can be attached only once, right ? No matter which case is.
And, the warning is appropriate for such cases,  multi container or other cases trying to attach devices more than once error!

src/runtime-rs/crates/hypervisor/src/device/util.rs
```diff
@@ -88,4 +110,61 @@ mod tests {
...
+
+        // Second, ref_count > 0
+        let ref_count_3 = &mut 3_u64;
+        assert!(do_increase_count(ref_count_3).unwrap());
+        assert!(do_decrease_count(ref_count_3).unwrap());
+
+        // Third, ref_count is MAX
+        let mut max_count = std::u64::MAX;
+        let ref_count_max: &mut u64 = &mut max_count;
+        let _ = do_increase_count(ref_count_max).is_err();
```
studychao, line 131: no assert here?

Apokleos: xxx`.is_err()` is doing the assertion work. test cast pass !

src/runtime-rs/crates/hypervisor/src/device/util.rs
Apokleos: fixed

src/runtime-rs/crates/hypervisor/src/device/util.rs
Apokleos: fixed

studychao (approved):
LGTM

Issue #8802, opened 2024-01-11
runtime-rs: ch: Implement full guest memory block size API


jodh-intel:
Implement full versions of the following two trait methods:

- `set_guest_memory_block_size()`
- `guest_memory_block_size_mb()`

See: #8784.

Issue #8801, opened 2024-01-11
runtime-rs: ch: Implement full resize memory API


jodh-intel:
#8784 implemented a minimal set of APIs to unbreak the CH runtime-rs driver. This issue is to provide the full version which will be similar to the golang `cloudHypervisor.ResizeMemory()` implementation:

https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/clh.go#L986..L1061

We will also need to add in full unit tests for this API.

> **Notes:**
> 
> - We will probably need to leverage / refactor `MemoryConfig::try_from()` in `src/runtime-rs/crates/hypervisor/ch-config/src/convert.rs`.
> - ACPI-based memory hotplug does not allow shrinking memory, while virtio-mem based supports it.
> - `virtio-balloon` is another way of allowing memory shrink at runtime.
> - TDX may require a different codepath.

Issue #8800, opened 2024-01-11
runtime-rs: ch: implement full hypervisor metrics method


jodh-intel:
This issue will be used to implement a full/fuller version of the `get_hypervisor_metrics()` API provided by #8785.

CH provides a `VmCounters` API which we can use for this.

kalil-pelissier:
Hi @jodh-intel!
I want to start working on the Rust runtime, can I take this issue?

jodh-intel:
Hi @kalil-pelissier - Sure! Just a warning though that you will probably need to do a bit of research on this one. Before you start, let's ask a few folk who might be able to give us some answers...

@Yuan-Zhuo and @Tim-0731-Hzt - Do you think it makes sense for us to implement this API for an external hypervisor like Cloud Hypervisor (CH)? CH doesn't provide all the metrics that dragonball does and in fact they metrics will be different anyway compared to https://github.com/kata-containers/kata-containers/blob/main/docs/design/kata-metrics-in-runtime-rs.md#kata-hypervisor. tbh, I wonder if we need to change this API so that a hypervisor can register the metrics it can provide since if CH returns different metrics, that could confused the consumer! wdyt @Yuan-Zhuo and @Tim-0731-Hzt?

*Iff* it makes sense to implement this, CH does have a [`/vm.counters` API](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/7c4ca2bcf97c792936abe2a5e2fb64ca6b5c0ff1/vmm/src/api/openapi/cloud-hypervisor.yaml#L44), so you could add a new function to https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/hypervisor/ch-config/src/ch_api.rs to allow the CH driver to call this CH HTTP GET API (see  `cloud_hypervisor_vmm_ping()` in that file for an example).

fwics, CH provides the following counters:

- net device:
  - `rx_bytes`
  - `rx_frames`
  - `tx_bytes`
  - `tx_frames`
- block device:
  - `read_bytes`
  - `read_latency_avg`
  - `read_latency_max`
  - `read_latency_min`
  - `read_ops`
  - `write_bytes`
  - `write_latency_avg`
  - `write_latency_max`
  - `write_latency_min`
  - `write_ops`
 
/cc @likebreath

Related:

- #5797.
[...]

PR #8803, opened 2024-01-11
runtime-rs: ch: Unbreak CH driver

jodh-intel:
Unfortunately, PR #6876 broke running runtime-rs with Cloud Hypervisor so add minimal implementations of the new APIs to allow CH to be used with runtime-rs.

Full implementations will be added later on issues #8800, #8801, and #8802.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
@@ -744,20 +744,24 @@ impl CloudHypervisorInner {
         Err(anyhow!("CH hypervisor metrics not implemented - see https://github.com/kata-containers/kata-containers/issues/8800"))
     }
 
-    pub(crate) fn set_capabilities(&mut self, _flag: CapabilityBits) {
-        todo!()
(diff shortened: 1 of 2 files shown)


src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -741,23 +741,29 @@ impl CloudHypervisorInner {
...
     }
 
     pub(crate) fn guest_memory_block_size_mb(&self) -> u32 {
-        todo!()
+        self._guest_memory_block_size_mb
     }
 
     pub(crate) fn resize_memory(&self, _new_mem_mb: u32) -> Result<(u32, MemoryConfig)> {
-        todo!()
+        // No resize currently - see https://github.com/kata-containers/kata-containers/issues/8801
```
fidencio, line 764: @jodh-intel, I think it's worth adding a message here saying that the resize is not currently implemented.

fidencio (approved):
lgtm, thanks @jodh-intel!

cmaf (approved):
thanks @jodh-intel

Issue #8805, opened 2024-01-11
gha: Limit jobs to run after "ok-to-test¨ is set


fidencio:
Let's limit a variety of tests that are either using the
`pull_request_target` event, or running on Azure instances, to only be
triggered after an `okay-to-test` label is set.

This helps us to avoid leaking the GITHUB_TOKEN in case a malicious
submitter tries to do so.

fidencio:
This is already done.

PR #8806, opened 2024-01-11
gha: Limit jobs to run after "ok-to-test¨ is set

fidencio:
Let's limit a variety of tests that are either using the `pull_request_target` event, or running on Azure instances, to only be triggered after an `okay-to-test` label is set.

This helps us to avoid leaking the GITHUB_TOKEN in case a malicious submitter tries to do so.

Fixes: #8805

Reported-by: Lukas Jokubauskas <sakul1234@proton.me>

### DIFF
--- a/.github/workflows/require-pr-porting-labels.yaml
+++ b/.github/workflows/require-pr-porting-labels.yaml
@@ -11,7 +11,6 @@ on:
       - opened
       - reopened
       - labeled
-      - unlabeled
     branches:
       - main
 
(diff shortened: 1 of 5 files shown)


sprt (commented):
You'll also want to add `labeled` to `on.pull_request_target.types` if you want to trigger these when a PR is labeled.

fidencio:
> You'll also want to add `labeled` to `on.pull_request_target.types` if you want to trigger these when a PR is labeled.

Why? I don't see this present in any of the current actions.

fidencio:
> Why? I don't see this present in any of the current actions.

Nevermind, now I understand the comment, and I've addressed it accordingly. :-)

beraldoleal (approved):
LGTM, thanks.

danmihai1 (approved):

.github/workflows/require-pr-porting-labels.yaml
```diff
@@ -11,7 +11,6 @@ on:
       - opened
       - reopened
       - labeled
-      - unlabeled
```
sprt, line 14: I think we want to keep this as some of the steps look for the *absence* of the `force-skip-ci` label.

Issue #8824, opened 2024-01-14
tests: cbl-mariner: disable k8s-oom.bats


danmihai1:
This test failed for my unrelated PRs at least 3 times in two days: 

https://github.com/kata-containers/kata-containers/issues/8821

Let's disable the test on cbl-mariner until it becomes more stable.

PR #8827, opened 2024-01-14
tests: cbl-mariner: disable k8s-oom.bats

danmihai1:
Disable k8s-oom.bats on cbl-mariner until it passes more often.

Fixes: #8824


justxuewei:
How about setting this as non-required? /cc @fidencio

fidencio:
> How about setting this as non-required? /cc @fidencio

I'm not the main person to decide what should or should not be required.
But if we can skip one test, that's better than making the whole CI as non-required, as it avoids things to get in that will end up breaking even more that specific CI.

fidencio (approved):
lgtm, thanks @danmihai1!

Issue #8835, opened 2024-01-15
tools: genpolicy: cbl-mariner: container root path is incorrect


danmihai1:
As a step towards testing genpolicy for CoCo, we should start by testing policies generated by this tool on cbl-mariner Guest VMs. Testing cbl-mariner on AKS might be easier than testing using a different Guest that supports AGENT_POLICY=yes, because my team at Microsoft is already testing on cbl-mariner.

genpolicy expects a container Rootfs path similar to:

/run/kata-containers/shared/containers/d5dff9a64f2e096c0ebbfbc2daab5da2a6975c83e231cf03f33c4561a8cf5e85

That path is incorrect for Kata main branch + cbl-mariner Guest VMs, and that blocks the testing described above. The actual path is similar to:

/run/kata-containers/shared/containers/d5dff9a64f2e096c0ebbfbc2daab5da2a6975c83e231cf03f33c4561a8cf5e85/rootfs

PR #8836, opened 2024-01-15
genpolicy: use root path from cbl-mariner Guest VM

danmihai1:
Adjust genpolicy-settings.json to match the container root path from the main branch + cbl-mariner Guest VMs.

This configuration might have to be adjusted again when other types of Guest VMs will be tested during CI using genpolicy, in the future.

Also, improve logging from allow_root_path(), to easier debug these issues in the future.

Fixes: #8835

### DIFF
--- a/src/tools/genpolicy/genpolicy-settings.json
+++ b/src/tools/genpolicy/genpolicy-settings.json
@@ -1,7 +1,7 @@
 {
     "pause_container": {
         "Root": {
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, in order to allow you to proceed with the tests.

We need to think, though, what's the best way to have this one sorted out for the generic case.

stevenhorsman (approved):
Looks okay to me in the short-term. Thanks

PR #8840, opened 2024-01-16
tdx: Update TDX artefacts to be used with the Ubuntu 23.10 / CentOS 9 stream OSVs.

fidencio:
As host side patches of TDX are being shipped by distros like Ubuntu, CentOS Stream, and SuSE, let's take advantage of this and switch to using those, while on the guest side we rely on the upstream content (kernel 6.7, which has everything we need).

Please, take a look at each commit in the series for further details.


danmihai1 (approved):

cmaf (approved):
@fidencio LGTM

Issue #8851, opened 2024-01-18
tests: start testing auto-generated Agent Policy for CoCo


danmihai1:
Kata CI already uses a simple "allow all" policy for many tests. Start converting these tests to use a more sophisticated policy, generated by the genpolicy tool.

Tests will gradually be moved in the future from the old "allow all policy bucket" to the new "auto-generated policy bucket" - as these tests start working with auto-generated policy.

Examples of tests that don't currently work with auto-generated policy: https://github.com/kata-containers/kata-containers/issues/8850.

PR #8852, opened 2024-01-18
tests: start auto-generating the policy

danmihai1:
Start auto-generating CoCo policy for kubernetes integration tests.

Policy gets auto-generated just for cbl-mariner Hosts for now. Auto-generated policy will be enabled for other Hosts too, after
testing these Hosts. For now, the other Hosts continue to use the allow-all.rego policy.

Some of the tests are still using the allow-all policy even on cbl-mariner Hosts, until the auto-generated policy will start working for these tests too.

Fixes: #8851

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -371,11 +371,10 @@ log_level = "debug"
 EOF
 }
 
-function install_kata() {
-	local kata_tarball="kata-static.tar.xz"
-	declare -r katadir="/opt/kata"
+function install_kata_common() {
+	local katadir="$1"
 	declare -r destdir="/"
-	declare -r local_bin_dir="/usr/local/bin/"
+	declare -r kata_tarball="kata-static.tar.xz"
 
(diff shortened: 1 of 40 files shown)


.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -63,6 +63,15 @@ jobs:
...
           TARGET_BRANCH: ${{ inputs.target-branch }}
 
+      - name: get-kata-tarball
+        uses: actions/download-artifact@v3
+        with:
+          name: kata-static-tarball-amd64${{ inputs.tarball-suffix }}
+          path: kata-artifacts
+
+      - name: Install kata
+        run: bash tests/integration/kubernetes/gha-run.sh install-kata kata-artifacts
```
sprt, line 73: I believe the right approach here is to add genpolicy to kata-deploy, otherwise this step would conflict with the Deploy Kata step.

Issue #8832, opened 2024-01-15
tools/genpolicy: unable to parse 'spec.dnsPolicy'


wainersm:
The genpolicy tool fails to parse a pod yaml that contains the `dnsPolicy` configuration.

How to reproduce the problem:

1. Create the pod yaml like below:
```
---
apiVersion: v1
kind: Pod
metadata:
  name: http-server
  labels:
    app: http-server-app
spec:
  runtimeClassName: kata-remote
  containers:
    - name: http-server
      image: registry.fedoraproject.org/fedora
      ports:
        - containerPort: 8080
      command: ["python3"]
      args: [ "-m", "http.server", "8080"]
  dnsPolicy: ClusterFirst
  restartPolicy: Never
```

2. Try to generate the Policy rego as:

```
$ ./genpolicy -r -y pod.yaml 
thread 'main' panicked at 'Unsupported field: spec.dnsPolicy', src/yaml.rs:303:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Removing `dnsPolicy: ClusterFirst` from the Pod yaml it should work fine. Found this issue while reviewing https://github.com/kata-containers/kata-containers/pull/8814

PR #8862, opened 2024-01-19
genpolicy: ignore pod DNS settings

danmihai1:
Ignore pod DNS settings because policing the network traffic is currently outside the scope of the Agent Policy.

Example from Kata CI: pod-custom-dns.yaml

Fixes: #8832


sprt (approved):

fidencio (approved):
lgtm, thanks @danmihai1!

Issue #8865, opened 2024-01-20
runtime-rs(Qemu): add netdev parameters to command line for starting up 


Apokleos:
add network related params to command line when qemu-rs startup.

PR #8866, opened 2024-01-20
runtime-rs: add netdev params to cmdline for qemu-rs.

Apokleos:
add network related params to command line when qemu-rs startup.

Fixes: #8865


Apokleos:
cc @pmores @gkurz @lifupan @Tim-0731-Hzt

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+//
+
+use std::{convert::TryFrom, fs::File};
+
+use crate::{Address, NetworkConfig};
+use anyhow::{anyhow, Context, Result};
+
+#[derive(Debug, Default, PartialEq)]
+enum NetDeviceWithTransport {
+    // VirtioNetPCI is the virt-io pci networking device driver.
```
gkurz, line 13: Official wording is `virtio`, not `virt-io`.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+            _ => return Err(anyhow!("unsupported transport")),
+        };
+
+        Ok(transport)
+    }
+}
+
+#[derive(Debug, Default, PartialEq)]
+enum NetDeviceDriver {
+    #[default]
```
gkurz, line 68: I don't understand this enum. It is mixing guest visible entities (`VirtioNet`, `Vfio`) with host backends (`Vhostuser`)... What is it used for ?

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// MACVTAP NetDeviceType = "macvtap"
+    MacvTap,
+    /// IpTap is a ipvtap virtual networking device type.
+    /// IPVTAP NetDeviceType = "ipvtap"
+    IpvTap,
+    /// VethTap is a veth-tap virtual networking device type.
+    /// VETHTAP NetDeviceType = "vethtap"
+    VethTap,
+    /// VhuNet is a vhost-user port (socket)
+    /// VhuNet NetDeviceType = "vhostuser"
```
gkurz, line 100: and `vhost-user`

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+        match self {
+            NetDeviceDriver::VirtioNet => "virtio-net".to_owned(),
+            NetDeviceDriver::Vhostuser => "vhost-user".to_owned(),
+            NetDeviceDriver::Vfio => "vfio".to_owned(),
+        }
+    }
+}
+
+#[derive(Debug, Default, PartialEq)]
+enum NetDeviceType {
```
gkurz, line 85: This tells what backend will forward I/O to/from the host network. This is what gets passed to `-netdev` in the end. Maybe call the enum `Netdev` ?

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+    IpvTap,
+    /// VethTap is a veth-tap virtual networking device type.
+    /// VETHTAP NetDeviceType = "vethtap"
+    VethTap,
+    /// VhuNet is a vhost-user port (socket)
+    /// VhuNet NetDeviceType = "vhostuser"
+    VhuNet,
+    /// Physical is a direct assigned PCI device or PCI VF
+    /// Physical NetDeviceType = "VFIO"
+    Physical,
```
gkurz, line 104: When assigning a device from the host using vfio, all network traffic goes directly to the network the device is plugged into, with no `-netdev` on the command line. You likely don't need this.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+impl NetDeviceType {
+    // QemuNetdevParam converts to the QEMU -netdev parameter notation
+    fn qemu_netdev_param(&self, transport: VirtioTransport) -> Option<&str> {
+        match self {
+            // -netdev type=tap -device virtio-net-pci
+            NetDeviceType::VethTap
+            | NetDeviceType::Tap
+            | NetDeviceType::MacvTap
+            | NetDeviceType::IpvTap => Some("tap"),
+            // -netdev type=vhost-user (no device)
```
gkurz, line 116: This comment you borrowed from govmm looks wrong. It certainly doesn't make sense to pass a  `-netdev` host backend without a companion guest visible `-device`...

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+            NetDeviceType::VhuNet => {
+                if transport == VirtioTransport::Ccw {
```
gkurz, line 118: Checking the transport that will be used inside the guest in this code that deals with the host backend is violating software boundaries. It should be performed elsewhere.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+            // -device vfio-pci (no netdev)
+            NetDeviceType::Physical => {
+                if transport == VirtioTransport::Mmio {
+                    warn!(
+                        sl!(),
+                        "vfio devices are not support with the MMIO transport"
+                    );
+                    return None;
+                }
+                Some("")
```
gkurz, line 133: As mentioned above, `-device vfio` comes with no `-netdev`. It seems that this function should not even be called in the vfio case.

gkurz (commented):
Hi @Apokleos ! Great thanks for starting working on this !

As the patch is fairly big, I will likely need several passes to review. Here's a first round of comments.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// VethTap is a veth-tap virtual networking device type.
+    /// VETHTAP NetDeviceType = "vethtap"
+    VethTap,
+    /// VhuNet is a vhost-user port (socket)
+    /// VhuNet NetDeviceType = "vhostuser"
+    VhuNet,
+    /// Physical is a direct assigned PCI device or PCI VF
+    /// Physical NetDeviceType = "VFIO"
+    Physical,
+}
```
pmores, line 105: The govmm original [NetDevice::Valid()](https://github.com/kata-containers/kata-containers/blob/295494c7dc77c563435734d74f3bcd2f88fc19b7/src/runtime/pkg/govmm/qemu/qemu.go#L874) function (not adopted by this PR) seems to suggest that no `NetDeviceType` is ever valid apart from `Tap` and `MacvTap`.

Also [networkModelToQemuType()](https://github.com/kata-containers/kata-containers/blob/295494c7dc77c563435734d74f3bcd2f88fc19b7/src/runtime/virtcontainers/qemu_arch_base.go#L569) doesn't ever return anything but `Tap` and `MacvTap`.

I suggest that we get rid of unused stuff in govmm.  I appears to me that we should get rid of all netdevice types & their handling except for `Tap` and `MacvTap`.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+// -netdev tap,fd=37,id=hostnet0,vhost=on,vhostfd=38
+// -device virtio-net-pci,netdev=hostnet0,id=net0,mac=24:42:54:20:50:46,bus=pci.0,addr=0x7
+// NetDevice represents a guest networking device
+#[derive(Debug, Default)]
+pub struct NetDevice {
+    // Type is the netdev type (e.g. tap).
+    net_type: NetDeviceType,
+
+    // Driver is the qemu device driver
+    driver: NetDeviceDriver,
```
pmores, line 210: This field doesn't ever seem to be read (i.e. used) so I believe we should drop it.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+    bus: String,
+
+    // Addr is the address offset of a PCI device.
+    pci_addr: String,
+
+    // DownScript is the tap interface deconfiguration script.
+    down_script: Option<String>,
+
+    // Script is the tap interface configuration script.
+    script: Option<String>,
```
pmores, line 228: `configuration.toml` doesn't seem to let users set the scripts so I think both fields can be dropped.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    // FIXME: to support mq
+    // // FDs represents the list of already existing file descriptors to be used.
+    // // This is mostly useful for mq support.
+    // fds:      Vec<File>,
+    // vhost_fds: Vec<File>,
+
+    // VHost enables virtio device emulation from the host kernel instead of from qemu.
+    vhost_enabled: bool,
```
pmores, line 237: `configuration.toml` doesn't allow users to set this per device so I think we should simplify by dropping this and referring to the master copy of this piece of information in `HypervisorConfig::network_info::disable_vhost_net` instead.

Apokleos:
Thx @gkurz @pmores  I will spare some time to correct the code as your suggestion.
As the PR is draft and many warnings and incorrect expression exist in this version, I know it.
But code review is welcome in this stage

pmores:
@Apokleos no worries, I understand this is WIP, thanks for kicking this off, I appreciate that!

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -43,6 +44,7 @@ impl QemuInner {
             qemu_process: None,
             config: Default::default(),
             devices: Vec::new(),
+            netns: None,
```
pmores, line 47: As discussed on #runtime-rs, this commit should also assign to `QemuInner::netns` in `prepare_vm()`, otherwise this won't work.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -61,6 +61,11 @@ impl QemuInner {
     pub(crate) async fn start_vm(&mut self, _timeout: i32) -> Result<()> {
         info!(sl!(), "Starting QEMU VM");
 
+        let netns = self.netns.clone();
+        if let Err(e) = do_add_netns(&netns) {
+            return Err(anyhow!("add netns failed with error {:?}", e));
+        };
+
```
pmores, line 68: This works, however it puts the executing thread permanently into the custom netns.  This is a departure from existing practice of `resource` crate code which rigorously ensures that the netns is only entered for operations that require it and is left afterwards.  I believe we should stick to that in the qemu driver as well.

The question is how to do it.  I'd very much like to reuse the `resource` crate's `NetnsGuard` for this, however we apparently cannot just `use` it since the `resource` crate already depends on the `hypervisor` crate so that would create a cyclic dependency.  One solution could be to move `NetnsGuard` to a lower-level utitility crate which could be imported by both `resource` and `hypervisor`, what do you think?

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -113,6 +118,15 @@ impl QemuInner {
...
 
+        // The qemu process needs be moved into Network Namespace.
+        unsafe {
+            let _pre_exec = command.pre_exec(move || {
+                let _ = do_add_netns(&netns);
+
+                Ok(())
+            });
+        }
+
```
pmores, line 138: Given the code as it is now, this is actually superfluous since the previous hunk already makes the executing thread enter the netns.

However, if we can manage to import `NetnsGuard` as proposed above, I would like to use it to enter the netns _just_ for the `add_network_device()` invocation.  In that case, this hunk would become useful.

src/runtime-rs/crates/hypervisor/src/utils.rs
```diff
@@ -47,3 +50,15 @@ pub fn get_jailer_root(sid: &str) -> String {
 
     [&sandbox_path, JAILER_ROOT].join("/")
 }
+
+pub fn do_add_netns(netns: &Option<String>) -> Result<()> {
```
pmores, line 74: I propose calling this function just `enter_netns()` since that's precisely what it does.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
Apokleos: Thx @pmores I re-checked my draft code in my local repo and I find that `self.netns = netns` indeed exists. 
I accidentally missed this when pushing the code. I will fix it.

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
```diff
@@ -54,7 +54,8 @@ pub struct NetworkConfig {
 pub struct NetworkDevice {
     /// Unique identifier of the device
     pub device_id: String,
-
+    /// network device driver: virtio-net-pci,
+    pub device_driver: String,
```
pmores, line 58: This is code common to all hypervisor drivers.  Is this field appropriate here, given that it's only set to what appears to be a qemu-specific value?

It seems to me that the qemu driver should do without this field and figure out the network device driver by itself, from other information.  What do you think?

pmores:
I'm really not sure about the `NetworkDevice::device_driver` addition (as pointed out in a review comment).  Seeing as it's not even really used in any substantial logic (just set to a literal and read once), it seems that it hits shared code for no good reason and could be removed.

Other than that, I'm fine merging this PR as-is.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -5,12 +5,15 @@
...
 use std::collections::HashMap;
 use std::convert::TryFrom;
+use std::fs::{File, OpenOptions};
 use std::os::fd::RawFd;
 
 use crate::utils::clear_fd_flags;
 use crate::{Address, NetworkConfig};
 use anyhow::{anyhow, Context, Result};
 
+use dbs_utils::net::Tap;
```
gkurz, line 15: This breaks build on s390x as reported by https://github.com/kata-containers/kata-containers/actions/runs/8188007403/job/22442402744?pr=8866 :

```
error[E0433]: failed to resolve: use of undeclared crate or module `dbs_utils`
  --> crates/hypervisor/src/qemu/network.rs:15:5
   |
15 | use dbs_utils::net::Tap;
   |     ^^^^^^^^^ use of undeclared crate or module `dbs_utils`
```

pmores: Yes, it's actually questionable to pull in an additional dependency when the same functionality seems to be available already in the `resource` crate.  I meant to address this in a subsequent PR but if it's breaking a build then it should perhaps be addressed earlier. (?)

gkurz: > Yes, it's actually questionable to pull in an additional dependency when the same functionality seems to be available already in the `resource` crate. I meant to address this in a subsequent PR but if it's breaking a build then it should perhaps be addressed earlier. (?)

Required checks depend on this build to succeed. The build break must be addressed in this PR.

gkurz:
> With my maintainer cap, I'd prefer the fixes to be addressed _of course_ but I won't make it mandatory since the PR is overall good. Up to you @Apokleos 😉

With required checks not passing, you'll have to fix this PR anyway so please address all comments. Thanks !

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: > This is code common to all hypervisor drivers. Is this field appropriate here, given that it's only set to what appears to be a qemu-specific value?
> 
> It seems to me that the qemu driver should do without this field and figure out the network device driver by itself, from other information. What do you think?
>
Fixed

Apokleos:
> Other than that, I'm fine merging this PR as-is.
>
Thx Pavel @pmores I have removed it from the `NetworkDevice in virtio_net.rs`

src/runtime-rs/crates/hypervisor/src/device/driver/virtio_net.rs
Apokleos: > > This is code common to all hypervisor drivers. Is this field appropriate here, given that it's only set to what appears to be a qemu-specific value?
> > It seems to me that the qemu driver should do without this field and figure out the network device driver by itself, from other information. What do you think?
> 
> I fully agree with you. This should be hidden in the QEMU driver for now.
> 
Thx Greg @gkurz I agree with you that the driver is needed to hidden in Qemu driver deduced with more info such as transport and net device type  or others.

Apokleos:
> > I'm really not sure about the `NetworkDevice::device_driver` addition (as pointed out in a review comment). Seeing as it's not even really used in any substantial logic (just set to a literal and read once), it seems that it hits shared code for no good reason and could be removed.
> 
> Yes for now I rather hard-code "virtio-net-pci" at the line where `device_driver` is "read once" in `src/runtime-rs/crates/hypervisor/src/qemu/inner.rs`.
> 
> > Other than that, I'm fine merging this PR as-is.
> 
> With my maintainer cap, I'd prefer the fixes to be addressed _of course_ but I won't make it mandatory since the PR is overall good. Up to you @Apokleos 😉
>
`NetworkDevice::device_driver` removed now.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
Apokleos: @pmores @gkurz I have checked the code and found that the dbs_utils dependencies `dbs-utils = { path = "../../../dragonball/src/dbs_utils" }` was moved to [`target.'cfg(not(target_arch = "s390x"))'.dependencies]` and hidden which causes the error of `undeclared crate or module` in s390x platform. And IMO, it should not be hidden now because `dbs_utils` is just a public utils for all vmms

Apokleos:
> > With my maintainer cap, I'd prefer the fixes to be addressed _of course_ but I won't make it mandatory since the PR is overall good. Up to you @Apokleos 😉
> 
> With required checks not passing, you'll have to fix this PR anyway so please address all comments. Thanks !
>
Thx @gkurz I have fixed the PR  and the required CI for s390x passed

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -0,0 +1,272 @@
...
+        //      "fds": vec![fd1, fd2,...],
+        //      "vhostfds": vec![fd3, fd4,...],
+        // }
+        let mut fds: HashMap<String, Vec<RawFd>> = HashMap::with_capacity(2);
+        fds.insert("fds".to_owned(), tun_fds);
+        fds.insert("vhostfds".to_owned(), vhost_fds);
+
+        // FIXME(Hard Code): It's safe to unwrap here because of the valid input.
+        let device_driver = DeviceDriver::try_from("virtio-net-pci").unwrap();
+        let transport = VirtioTransport::try_from("pci").unwrap();
```
gkurz, line 154: Ideally `device_driver` should be derived from `transport` to avoid code duplication.

At some point in time we might want to support other transports, e.g. s390x's CCW, but I guess it is acceptable to get started with PCI only.

src/runtime-rs/crates/hypervisor/src/qemu/network.rs
```diff
@@ -270,3 +271,66 @@ impl ToString for Address {
...
+
+// /dev/tap$(cat /sys/class/net/macvtap1/ifindex)
+// for example: /dev/tap2381
+#[allow(dead_code)]
+pub fn create_macvtap_fds(ifindex: u32, queues: u32) -> Result<Vec<File>> {
+    let macvtap = format!("/dev/tap{}", ifindex);
+    create_fds(macvtap.as_str(), queues as usize)
+}
+
+pub fn open_vhost_net(queues: u32) -> Result<Vec<File>> {
```
gkurz, line 283: Nit: maybe stay consistent and name it `create_vhost_net_fds` ?

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -68,6 +68,8 @@ impl QemuInner {
         // descriptor needs to stay open until the qemu process launches.
         // This is why we need to store it in a variable at this scope.
         let mut _vhost_fd = None;
+        let mut _tun_fds: Vec<std::fs::File> = Vec::new();
+        let mut _vhost_fds: Vec<std::fs::File> = Vec::new();
```
gkurz, line 74: We don't need (and probably don't want) to know about the underlying type of fd. We just want them to stay open in the current scope. I'd prefer to have a single `_fds_for_qemu`  array to hold them all, for the sake of clarity. The comment above needs to be adapted as well.

gkurz (approved):
I left some comments but LGTM overall and it passed CI so I'm approving.

Thanks @Apokleos !

pmores (approved):

Issue #8873, opened 2024-01-22
gpu: Remove the old SNP, TDX targets


zvonkok:
Remove the old SNP, TDX GPU related targets we no rely on https://github.com/kata-containers/kata-containers/pull/8872

PR #8872, opened 2024-01-22
gpu: Add NVIDIA GPU Confidential kernel target

zvonkok:
This is a follow up to the work of minimizing targets, unifying TDX,SNP builds for NVIDIA GPUs

Fixes: #8828

Follow-up PR will remove the old targets.

PR #8874, opened 2024-01-22
gpu: Remove the old SNP, TDX targets

zvonkok:
Remove the old SNP, TDX GPU related targets we no rely on #8872

Fixes:  #8873


fidencio:
So, just to make it clear for the maintainers, all the tests are going to fail as we're removing a build target.
This is expected, and someone will have to mark the checks as non-required and then as required again.

fidencio (approved):
lgtm, thanks @zvonkok!

jepio (approved):

zvonkok:
Closing in favour #9026

Issue #8877, opened 2024-01-22
gpu: cache kernel-headers for module builds for GPUs 


zvonkok:
We need to provide new targets for kernel-header packaging so it can be cached.

PR #8878, opened 2024-01-22
gpu: cache kernel-headers for module builds for GPUs

zvonkok:
We need to provide new targets for kernel-header packaging so it can be cached.

Fixes: #8877

### DIFF
--- a/tools/packaging/kata-deploy/local-build/Makefile
+++ b/tools/packaging/kata-deploy/local-build/Makefile
@@ -24,6 +24,8 @@ BASE_TARBALLS = serial-targets \
 	kernel-dragonball-experimental-tarball \
 	kernel-nvidia-gpu-tarball \
 	kernel-nvidia-gpu-confidential-tarball \
+	kernel-headers-nvidia-gpu-tarball \
+	kernel-headers-nvidia-gpu-confidential-tarball \
 	kernel-nvidia-gpu-snp-tarball \
 	kernel-nvidia-gpu-tdx-experimental-tarball \
 	kernel-tarball \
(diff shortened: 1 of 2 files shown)


zvonkok:

fidencio:
@zvonkok, GHA will have to be changed, as the cache is done when they run.

tools/packaging/kata-deploy/local-build/Makefile
```diff
@@ -24,6 +24,8 @@ BASE_TARBALLS = serial-targets \
 	kernel-dragonball-experimental-tarball \
 	kernel-nvidia-gpu-tarball \
 	kernel-nvidia-gpu-confidential-tarball \
+	kernel-headers-nvidia-gpu-tarball \
+	kernel-headers-nvidia-gpu-confidential-tarball \
```
fidencio, line 28: Are there situations where you want only the headers, but not the kernel itself?

I'm asking here because, IMHO, we could *always* make sure that building the nvidia-gpu kernels we'd generate the deb and headers, and *always* add them to the ORAS container related to the kernel itself.  No need to add new extra targets.

WDYT?

fidencio (changes requested):
@zvonkok, I think this could be simpler, please, take a look at my comment.

tools/packaging/kata-deploy/local-build/Makefile
fidencio: https://github.com/kata-containers/kata-containers/blob/e5e0983b56534f007cf06ab39f323961b17ff642/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L185-L187

Here's what you can use, the extra_tarballs.

PR #8904, opened 2024-01-24
tools: Use defined variable in build base qemu script

### LINKED ISSUE #8903: tools: Use defined variable in build base qemu script

GabyCT:
This PR uses a variable that is already defined in the build base qemu script to have uniformity across the script as this variable is already used in the script.

Fixes #8903


cmaf (approved):
thanks @GabyCT

PR #8910, opened 2024-01-25
dragonball: add dbs_tdshim crate to support TD-Shim

NK-fengsf:
dbs_tdshim crate has achieved the ability to parse TD metadata in TD-Shim binary file.

TD-Shim is a shim firmware to support intel TDX.
For more detail: https://github.com/confidential-containers/td-shim

#8911


NK-fengsf:
@studychao  Can you help me review this code？

fidencio:
@jodh-intel, mind to take a look at this one?

My interest here is mostly whether this code could be reused outside of dragonball, as Cloud Hypervisor will also rely on td-shim.

NK-fengsf:
> @jodh-intel, mind to take a look at this one?
> 
> My interest here is mostly whether this code could be reused outside of dragonball, as Cloud Hypervisor will also rely on td-shim.

Yes, it could be reused outside of dragonball. 
Actually most of the code are derived from the [td-shim](https://github.com/confidential-containers/td-shim) project.  But  it will introduce many unnecessary dependencies， If we use td-shim  cartes directly. As the code is relatively scattered in td-shim. 
Cloud hypervisor has also done similar things.

Apokleos:
cc @lifupan @bergwolf @Apokleos

jyao1:
https://crates.io/crates/td-shim-interface is submitted.

studychao:
@jyao1 thank you so much for your help!

Issue #8913, opened 2024-01-25
Add confidential containers test for IBM Secure Execution


BbolroC:
**Which feature do you think can be improved?**

It would be great to add an additional hypervisor `qemu-se` on top of the existing ones in a test titled `Test unencrypted confidential container launch success and verify that we are running in a secure enclave`: https://github.com/kata-containers/kata-containers/blob/24b33ae35bacf22d70ec9fcfa0f9aa211dca0ab0/tests/integration/kubernetes/confidential_common.sh#L24-L34

To ensure the inclusion of e2e tests for a new confidential containers feature, it is essential that they are specifically designed to function with IBM Secure Execution. Otherwise, there is a risk of oversight, and the motivation to incorporate them later may be lacking.

Issue #8906, opened 2024-01-24
Pod stuck in `StartError` for k8s-inotify.bats on IBM Z16 LPAR


BbolroC:
# Description of problem

While a test titled `configmap update works, and preserves symlinks` passes on a zVSI instance in IBM Cloud (self-hosted runners in a GHA workflow) https://github.com/kata-containers/kata-containers/blob/f4290688bb7b36e7a65b184ba750d21e61004100/.github/workflows/run-k8s-tests-on-zvsi.yaml#L72-L74, it has been observed that the test fails on an internal Z16 LPAR bare-metal machine with the following error:

```

Containers:
  c1:
    Container ID:  containerd://22a9667f7277e4b2aca2439a656f781a201777af2e0f709b1dc34170c083f0e0
    Image:         quay.io/kata-containers/fsnotify:latest
    Image ID:      quay.io/kata-containers/fsnotify@sha256:244e5624ddfcb2cc0c4c327f26b3bdc8aeca6e29ab4cc492c8ba315905900765
    Port:          <none>
    Host Port:     <none>
    Command:
      bash
    Args:
      -c
      inotifywait --timeout 120 -r /config/ && [[ -L /config/config.toml ]] && echo success
    State:          Terminated
      Reason:       StartError
      Message:      failed to create containerd task: failed to create shim task: process: 1 failed to receive async message from peer: got msg length: 0, expected: 4: unknown
      Exit Code:    128
      Started:      Thu, 01 Jan 1970 00:00:00 +0000
      Finished:     Wed, 24 Jan 2024 17:48:01 +0000
    Ready:          False
    Restart Count:  0
    Limits:
      memory:  1Gi
    Requests:
      memory:     50Mi
    Environment:  <none>
    Mounts:
      /config from config (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-lj99n (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             False
  ContainersReady   False
  PodScheduled      True
Volumes:
  config:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      cm
    Optional:  false
  kube-api-access-lj99n:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   Burstable
[...]

PR #8914, opened 2024-01-25
tests: Add IBM SE to the basic confidential test

BbolroC:
The existing confidential basic test titled `Test unencrypted confidential container launch success and verify that we are running in a secure enclave` has been updated to incorporate IBM Secure Execution (`qemu-se`).
Previously, a secure image was absent from kata-deploy, hindering the inclusion of IBM SE in the test.
Thanks to the #6755 update, it is now possible to test the TEE.

This modification extends the existing test by introducing `qemu-se`. The specific changes are outlined below:

- Add an additional test `cc-se-e2e-tests` to s390x nightly
- Expansion of `REMOTE_COMMAND_PER_HYPERVISOR` for `qemu-se`
- Temporary exclusion of two test cases currently incompatible with IBM SE (`cpu-ns` is a common issue across all TEEs, while `inotify` will be addressed in a subsequent pull request for #8906).

Reviewers, please note that the test has already been verified on Jenkins at http://jenkins.katacontainers.io/job/kata-containers-CCv0-ubuntu-20.04-s390x-SE-daily/374/. It can only be verified internally due to limitations in computing resources (SE capable). To address this, the test is triggered by the nightly GHA workflow. The test log fetched during the workflow will be the same as the one verified on Jenkins.

Fixes: #8913

### DIFF
--- a/tests/integration/kubernetes/k8s-confidential.bats
+++ b/tests/integration/kubernetes/k8s-confidential.bats
@@ -10,7 +10,7 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 load "${BATS_TEST_DIRNAME}/tests_common.sh"
 
 setup() {
-	SUPPORTED_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx")
+	SUPPORTED_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx" "qemu-se")
 
 	# This check must be done with "<SPACE>${KATA_HYPERVISOR}<SPACE>" to avoid
 	# having substrings, like qemu, being matched with qemu-$something.
@@ -36,7 +36,7 @@ setup() {
 			warn "Failed to get pod IP address."
 		else
 			info "Pod IP address: ${pod_ip}"
-			coco_enabled=$(ssh -i ${SSH_KEY_FILE} -o "StrictHostKeyChecking no" -o "PasswordAuthentication=no" root@${pod_ip} /bin/sh -c "$(get_remote_command_per_hypervisor)") && break
+			coco_enabled=$(ssh -i ${SSH_KEY_FILE} -o "StrictHostKeyChecking no" -o "PasswordAuthentication=no" root@${pod_ip} "$(get_remote_command_per_hypervisor)" 2> /dev/null) && break
 			warn "Failed to connect to pod."
 		fi
 		sleep 5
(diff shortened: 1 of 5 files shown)


tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -36,7 +36,9 @@ setup() {
 			warn "Failed to get pod IP address."
 		else
 			info "Pod IP address: ${pod_ip}"
-			coco_enabled=$(ssh -i ${SSH_KEY_FILE} -o "StrictHostKeyChecking no" -o "PasswordAuthentication=no" root@${pod_ip} /bin/sh -c "$(get_remote_command_per_hypervisor)") && break
+			base_command="/bin/sh -c"
+			[[ "${KATA_HYPERVISOR}" == "qemu-se" ]] && base_command=""
```
gkurz, line 40: Why dropping `"/bin/sh -c"` ?

BbolroC: The command for `qemu-se` only worked without `/bin/sh -c` (a default shell was `/bin/ash` as checked). 

This is a solution I ended up with from lots of different combinations of commands to get the command `kubectl exec ...` to return 1 for the entitled guest on the platform. 

So this is the best I can do to achieve my goal. I would be grateful if you suggest any clean and better solution.

gkurz: > So this is the best I can do to achieve my goal. I would be grateful if you suggest any clean and better solution.

From the `SSH(1)` manual page :
```
     If a command is specified, it will be executed on the remote host instead
     of a login shell.  A complete command line may be specified as command,
     or it may have additional arguments.  If supplied, the arguments will be
     appended to the command, separated by spaces, before it is sent to the
     server to be executed.
```

This tends to indicate that `/bin/sh -c` isn't needed at all. I'd rather drop it completely then.

GabyCT (approved):

hbrueckner:
Hi @stevenhorsman 

> Do you know if there are any plans for Marist to have an LPAR with SE available, or any other options for us to add an SE self-hosted runner in the future?

I have to ask the Marist team on their plans. For other options, I will contact you offline and then let's circle back if there are options.

cc: @magowan

tests/integration/kubernetes/k8s-confidential.bats
BbolroC: tdx is passed, but sev and snp using `dmesg | grep ...` are not picked up by a runner.

gkurz (approved):
LGTM. Thanks @BbolroC !

PR #8916, opened 2024-01-25
packaging:  Don't always build the kata-agent

fidencio:
There's absolutely no need to always build the Kata Containers agent when building the rootfs. We already have the agent being built / packaged on every PR, and we can rely on that to simply be unpacked into the rootfs, unless there's a change in the agent.

Please, see each commit for more info.

### DIFF
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -680,7 +680,6 @@ EOF
 		make clean
 		make LIBC=${LIBC} INIT=${AGENT_INIT} SECCOMP=${SECCOMP} AGENT_POLICY=${AGENT_POLICY}
 		make install DESTDIR="${ROOTFS_DIR}" LIBC=${LIBC} INIT=${AGENT_INIT}
-		${stripping_tool} ${ROOTFS_DIR}/usr/bin/kata-agent
 		if [ "${SECCOMP}" == "yes" ]; then
 			rm -rf "${libseccomp_install_dir}" "${gperf_install_dir}"
 		fi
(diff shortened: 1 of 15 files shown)


stevenhorsman (approved):
LGTM. Thanks for the update - I hope this will save a lots of CI time and money!

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -156,6 +156,24 @@ install_cached_tarball_component() {
 	mv "${component_tarball_name}" "${component_tarball_path}"
 }
 
+get_agent_tarball_path()
+{
```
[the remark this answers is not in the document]
fidencio: I will add it back :-)

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -693,6 +692,8 @@ EOF
 		tar xvJpf ${AGENT_TARBALL} -C ${ROOTFS_DIR}
 	fi
 
+	${stripping_tool} ${ROOTFS_DIR}/usr/bin/kata-agent
```
danmihai1, line 695: We can do this later, but perhaps we should have a way to bypass this new behavior.

danmihai1:
A lot of goodness here - thanks @fidencio !

danmihai1 (approved):

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: I'd prefer doing this later, but just to confirm.

The idea here is that we can bypass the behaviour and actually **not** have the component installed, correct?
If so, I think we can enforce this for the attestation-agent / confiential-data-hub / opa, are those are extra components.  The kata agent itself must always be there, though.

stevenhorsman (approved):
LGTM. Thanks!

tools/osbuilder/rootfs-builder/rootfs.sh
danmihai1: Never mind. Yesterday I thought this rootfs.sh change was a significant change in symbol strip behavior - but I see now that it's consistent with the previous behavior.

Issue #8891, opened 2024-01-23
genpolicy: PodTemplateSpec metadata is an optional field


danmihai1:
genpolicy treats PodTemplateSpec's metadata field as mandatory, and therefore rejects tests such as k8s-job.bats from Kata CI.

PR #8918, opened 2024-01-25
genpolicy: optional PodTemplateSpec metadata field

danmihai1:
Add metadata containing the Policy annotation if the user didn't provide any metadata in the input yaml file.

For a simple sanity test using a Kata CI YAML file:

genpolicy -u -y job.yaml

kubectl apply -f job.yaml

kubectl get pods | grep job
job-pi-test-64dxs 0/1     Completed   0          14s

Fixes: #8891

### DIFF
--- a/src/tools/genpolicy/src/daemon_set.rs
+++ b/src/tools/genpolicy/src/daemon_set.rs
@@ -111,7 +111,7 @@ impl yaml::K8sResource for DaemonSet {
     }
 
     fn serialize(&mut self, policy: &str) -> String {
-        yaml::add_policy_annotation(&mut self.doc_mapping, "spec.template.metadata", policy);
+        yaml::add_policy_annotation(&mut self.doc_mapping, "spec.template", policy);
         serde_yaml::to_string(&self.doc_mapping).unwrap()
     }
 
@@ -120,7 +120,10 @@ impl yaml::K8sResource for DaemonSet {
     }
 
     fn get_annotations(&self) -> &Option<BTreeMap<String, String>> {
-        &self.spec.template.metadata.annotations
+        if let Some(metadata) = &self.spec.template.metadata {
+            return &metadata.annotations;
+        }
+        &None
     }
 
     fn use_host_network(&self) -> bool {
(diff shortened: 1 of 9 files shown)


sprt (approved):

src/tools/genpolicy/src/yaml.rs
```diff
@@ -264,10 +264,23 @@ pub fn add_policy_annotation(
     let policy_key = serde_yaml::Value::String("io.katacontainers.config.agent.policy".to_string());
     let policy_value = serde_yaml::Value::String(policy.to_string());
 
-    let path_components = metadata_path.split('.');
-    for name in path_components {
-        ancestor = ancestor.get_mut(name).unwrap();
+    if !metadata_path.is_empty() {
```
wainersm, line 267: It seems to not make sense `metadata_path` being `empty`, if so, then it is an error?

src/tools/genpolicy/src/daemon_set.rs
```diff
@@ -120,7 +120,10 @@ impl yaml::K8sResource for DaemonSet {
     }
 
     fn get_annotations(&self) -> &Option<BTreeMap<String, String>> {
```
wainersm, line 122: <not_rust_expert>Not related with this change, but is there a way for the `K8sResource` trait to implement a default `get_annotations()` so that it doesn't need to be define on every and each implementer?</not_rust_expert>

src/tools/genpolicy/src/yaml.rs
danmihai1: It's empty for the caller from pod.rs.

wainersm: sorry, missed that. Yeah, for pod spec you need to include `metadata` in the yaml's root. ok.

wainersm (approved):
LGTM, thanks @danmihai1 !

src/tools/genpolicy/src/daemon_set.rs
danmihai1: Yes - I realized that too recently, but planning to address all of the these in a separate PR. Please create a github issue as a reminder if you'd like.

I am not a rust expert either and earlier I was looking for a while to inherit across 2 or more *structs* - similar to C++ - and I couldn't find that. But, more recently I've learned that a default implementation can be provided in the *trait* instead of a struct.

GabyCT (approved):

Issue #8921, opened 2024-01-25
tests: auto-generated policy for k8s-attach-handlers.bats


danmihai1:
Policy will get auto-generated just for cbl-mariner Hosts for now, while experimenting with genpolicy during CI. Auto-generated policy will be enabled for other Hosts too, after testing these other Hosts. For now, these other Hosts will continue to use the allow-all.rego policy.

The older allow-all policy will also be used on cbl-mariner Hosts for tests where auto-generated policy has not been enabled yet. Auto-generated policy will be enabled for these other tests using future PRs.

Issue #8864, opened 2024-01-19
tests: need better solution for installing the Kata tools without downloading the entire Kata tarball too


danmihai1:
https://github.com/kata-containers/kata-containers/pull/8857 uses the entire Kata tarball to install the genpolicy tool. Tests should be able to install just the tools, or just the Kata genpolicy tool.

Not sure if these other tests would benefit from this test feature too:

workflows/run-metrics.yaml:        run: bash tests/metrics/gha-run.sh install-kata kata-artifacts
workflows/run-kata-monitor-tests.yaml:        run: bash tests/functional/kata-monitor/gha-run.sh install-kata kata-artifacts
workflows/run-runk-tests.yaml:        run: bash tests/integration/runk/gha-run.sh install-kata kata-artifacts
workflows/run-cri-containerd-tests-ppc64le.yaml:        run: bash tests/integration/cri-containerd/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/integration/cri-containerd/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/stability/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/integration/nydus/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/integration/runk/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/functional/tracing/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/integration/docker/gha-run.sh install-kata kata-artifacts
workflows/basic-ci-amd64.yaml:        run: bash tests/integration/nerdctl/gha-run.sh install-kata kata-artifacts
workflows/run-cri-containerd-tests-s390x.yaml:        run: bash tests/integration/cri-containerd/gha-run.sh install-kata kata-artifacts

wainersm:
When I looked at #8857 I had this same idea, have a separated tarball for the tools that perhaps shouldn't be installed in the cluster so decreasing the size of the kata-deploy payload.

[...]

PR #8922, opened 2024-01-25
tests: k8s-attach-handlers auto-generated policy

danmihai1:
Automatically generate the test policy for k8s-attach-handlers.bats, if AUTO_GENERATE_POLICY is enabled.

A few commits are part of this PR - some of them not directly related to k8s-attach-handlers.bats but helpful for easier testing of the k8s-attach-handlers.bats changes.

Fixes: #8921


tests/integration/kubernetes/gha-run.sh
```diff
@@ -20,6 +20,7 @@ KATA_DEPLOY_WAIT_TIMEOUT=${KATA_DEPLOY_WAIT_TIMEOUT:-10m}
 KATA_HYPERVISOR=${KATA_HYPERVISOR:-qemu}
 KUBERNETES="${KUBERNETES:-}"
 SNAPSHOTTER="${SNAPSHOTTER:-}"
+export TEST_CLUSTER_NAMESPACE="${TEST_CLUSTER_NAMESPACE:-kata-containers-k8s-tests}"
```
wainersm, line 27: hi @danmihai1 !

Could you define `TEST_CLUSTER_NAMESPACE` in tests/gha-run-k8s-common.sh along-side the *_namespace() functions so that the callers don't need to export it?

fidencio (approved):
@danmihai1, series looks good to me.
I'd like to have @wainersm taking a look at this one as well, if possible.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -8,6 +8,9 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+DEBUG="${DEBUG:-}"
+[ -n "$DEBUG" ] && set -x
```
wainersm, line 12: I was about to suggest the use of true/false instead of empty/not empty value, but I grep'ed the other tests scripts and most of them are using this same approach.

danmihai1: I agree - but I decided to keep DEBUG consistent with most of the other scripts.

tests/common.bash
```diff
@@ -371,18 +371,29 @@ log_level = "debug"
 EOF
 }
 
-function install_kata() {
-	local kata_tarball="kata-static.tar.xz"
-	declare -r katadir="/opt/kata"
+function install_kata_common() {
```
wainersm, line 374: IMHO `install_kata_core` would be a better name.

tests/common.bash
```diff
@@ -371,18 +371,29 @@ log_level = "debug"
...
 	sudo rm -rf "${katadir}"
 
 	pushd "${kata_tarball_dir}"
 	sudo tar -xvf "${kata_tarball}" -C "${destdir}"
 	popd
+}
+
+function install_kata_tools() {
+	declare -r katadir="/opt/kata"
+	install_kata_common "${katadir}"
```
wainersm, line 389: I suppose you will change this when tools are split out of the kata tarball (https://github.com/kata-containers/kata-containers/issues/8864 ). Right? Worth a `TODO` comment?

tests/integration/kubernetes/gha-run.sh
```diff
@@ -169,6 +169,11 @@ function run_tests() {
 	[ "$platform" = "kcli" ] && \
 		export KUBECONFIG="$HOME/.kcli/clusters/${CLUSTER_NAME:-kata-k8s}/auth/kubeconfig"
 
+	# Enable auto-generated policy for CI images that support policy.
+	#
+	# TODO: enable testing auto-generated policy for other types of hosts too.
+	[ "${KATA_HOST_OS}" = "cbl-mariner" ] && export AUTO_GENERATE_POLICY="yes"
```
wainersm, line 174: I don't have a firm opinion about that but I think I would export `AUTO_GENERATE_POLICY="yes"` at the Github's workflow level. Other job configuration rules will be passed to the scripts via environment variables from the workflows, for example, https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-kata-coco-tests.yaml#L45

I understand that moving the rules to the workflows it will be harder to test on pull request time. So perhaps keep it as is, running some days then switching?

tests/integration/kubernetes/k8s-attach-handlers.bats
```diff
@@ -15,28 +15,39 @@ setup() {
...
-}
+	yaml_file="${pod_config_dir}/test-lifecycle-events.yaml"
 
-@test "Running with postStart and preStop handlers" {
 	# Create yaml
 	sed -e "s/\${nginx_version}/${nginx_image}/" \
-		"${pod_config_dir}/lifecycle-events.yaml" > "${pod_config_dir}/test-lifecycle-events.yaml"
+		"${pod_config_dir}/lifecycle-events.yaml" > "${yaml_file}"
+
+	# Add policy to yaml
```
wainersm, line 24: This block could be in a `setup_file()` which is executed once. Unless you really need the policy files re-generate for every and each test case.

Another question is: are you planning to copy this boilerpate over all the k8s tests as you go enabling them? or maybe run 1x on https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/setup.sh ?

wainersm:
hi @danmihai1 , finish my 1st round of review. Most nits & questions to you :)

tests/integration/kubernetes/k8s-attach-handlers.bats
danmihai1: The steps to generate Policy are specific to each test case, so my plan is to add similar steps to every relevant bats file. 

Examples why the steps are specific to each test case:
- Some of the tests use an already existing YAML file vs. generating one with sed
- Some of the tests want to "kubectl exec" one or more commands, others don't exec anything
- Some use config maps, while others don't 
- etc.

So, I don't think the centralized setup.sh approach works for these tests.

danmihai1:
> hi @danmihai1 , finish my 1st round of review. Most nits & questions to you :)

Thanks for the feedback, @wainersm! I believe I addressed it with my latest push.

tests/integration/kubernetes/gha-run.sh
wainersm: sounds reasonable.

tests/integration/kubernetes/gha-run.sh
danmihai1: @wainersm, I tried, but looks like we have to go back to the exported TEST_CLUSTER_NAMESPACE, because tests_common.sh is using that variable too (https://github.com/kata-containers/kata-containers/actions/runs/7809105784/job/21320896440?pr=8922). Unless you think it's worth doing a different type of refactoring here.

GabyCT (approved):

PR #8923, opened 2024-01-25
[kata-deploy]remove stale kata shims

tariq1890:
The kata-cloud-hypervisor and kata-stratovirt runtime classes have been removed from the stable branches.

Since the stable docker image tag is used in the kata-deploy manifests in the main branch, the `SHIMS` env var should be updated to be in sync with the latest stable kata-deploy docker images


tariq1890:
cc @zvonkok @shivamerla @cdesiniotis

fidencio:
> The kata-cloud-hypervisor and kata-stratovirt runtime classes have been removed from the stable branches.

It's actually the opposite, those were never added to the stable branches.
We have a current problem, IMHO, that kata-deploy is targetting the "latest" release, rather than the latest development branch, causing us this kind of issues.

I'm open to hear suggestions on how to address this.

tariq1890:
I am new to this repo, so I was wondering - Why do we have the divergence? 

Ideal `latest` should represent whatever is pushed to the `main` branch

Issue #8892, opened 2024-01-23
genpolicy: add support for Kata CI's pod-nested-configmap-secret.yaml


danmihai1:
genpolicy fails when encountering type: Opaque in the k8s Secret from pod-nested-configmap-secret.yaml , in Kata CI.

PR #8924, opened 2024-01-26
genpolicy: fix ConfigMap volume mount paths

danmihai1:
Allow Kata CI's pod-nested-configmap-secret.yaml to work with
genpolicy and current cbl-mariner images:

1. Ignore the optional type field of Secret input YAML files.

    It's possible that CoCo will need a more sophisticated Policy
    for Secrets, but this change at least unblocks CI testing for
    already-existing genpolicy features.

2. Adapt the value of the settings field below to fit current CI
    images for testing on cbl-mariner Hosts:

    "kata_config": {
        "confidential_guest": false
    },

    Switching this value from true to false instructs genpolicy to
    expect ConfigMap volume mounts similar to:

        "configMap": {
            "mount_type": "bind",
            "mount_source": "$(sfprefix)",
            "mount_point": "^$(cpath)/watchable/$(bundle-id)-[a-z0-9]{16}-",
            "driver": "watchable-bind",
            "fstype": "bind",
            "options": [
                "rbind",
                "rprivate",
                "ro"
            ]
        },

    instead of:

        "confidential_configMap": {
            "mount_type": "bind",
            "mount_source": "$(sfprefix)",
            "mount_point": "$(sfprefix)",
            "driver": "local",
            "fstype": "bind",
            "options": [
                "rbind",
                "rprivate",
                "ro"
            ]
        }
    },

    This settings change unblocks CI testing for ConfigMaps.

Simple sanity testing for these changes:

genpolicy -u -y pod-nested-configmap-secret.yaml

kubectl apply -f pod-nested-configmap-secret.yaml

kubectl get pods | grep config
nested-configmap-secret-pod 1/1     Running   0          26s

Fixes: #8892

### DIFF
--- a/src/tools/genpolicy/src/secret.rs
+++ b/src/tools/genpolicy/src/secret.rs
@@ -33,6 +33,9 @@ pub struct Secret {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     immutable: Option<bool>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    r#type: Option<String>,
     // TODO: additional fields.
 }
 
--- a/src/tools/genpolicy/genpolicy-settings.json
+++ b/src/tools/genpolicy/genpolicy-settings.json
@@ -258,7 +258,7 @@
         ]
     },
     "kata_config": {
-        "confidential_guest": true
+        "confidential_guest": false
     },
     "request_defaults": {
         "CreateContainerRequest": {


src/tools/genpolicy/src/secret.rs
```diff
@@ -33,6 +33,9 @@ pub struct Secret {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     immutable: Option<bool>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    r#type: Option<String>,
```
wainersm, line 38: Sorry to keep asking you stupid questions, but what happens if not a raw type `r#type` like `data` or `immutable`?

IIUC it will ignore the `type: Opaque` on pod-nested-configmap-secret.yaml's Secret:


```
<snip>
---
apiVersion: v1
data:
  secret_key: bXlzZWNyZXQ= #mysecret
kind: Secret
metadata:
  name: secret
type: Opaque
---
<snip>
```

danmihai1: Please never worry about "stupid" questions with me - I am not a rust expert either.

Here I believe "r#type" is needed here because "type" is a reserved keyword in Rust. We're forcing the compiler to use "type" as a struct field name here.

wainersm:
@danmihai1 mind if I open an issue to document `confidential_guest` setting from `genpolicy-settings.json`?

src/tools/genpolicy/src/secret.rs
danmihai1: When I said "Ignore the optional type field" I meant:

- If serde_yaml finds the type field in an intput YAML file, it gives that field value to genpolicy, and *genpolicy will ignore it*
- If serde_yaml doesn't find the type field in an intput YAML file, it gives None as the field value to genpolicy, and *genpolicy will ignore None*

In other words, this new r#type behaves the same way as the new nodeName field from https://github.com/kata-containers/kata-containers/commit/076869aa39d96b548aa1ab4c12c999239392e823.

danmihai1:
> @danmihai1 mind if I open an issue to document `confidential_guest` setting from `genpolicy-settings.json`?

That sounds great, thanks!

wainersm (approved):
Thanks @danmihai1 !

Issue #8917, opened 2024-01-25
cache: Cache the agent for non-x86_64 architectures


fidencio:
Those are not being cached at the moment.

fidencio:
The relevant parts for this work are:
* Having the agent being built as part of our GHA:
  * ARM64: https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/build-kata-static-tarball-arm64.yaml#L29
  * PowerPC64: https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/build-kata-static-tarball-ppc64le.yaml#L29
  * s390x: https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/build-kata-static-tarball-s390x.yaml
  * Reference of what we do on x86_64: https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/build-kata-static-tarball-amd64.yaml#L30

Now, I do believe that for building the agent for s390x and ppc64le we'd need to also expandhttps://github.com/kata-containers/kata-containers/tree/main/tools/packaging/static-build/agent as that's currnetly using alpine and relying on musl.

Ideally I'd like to see something as we've done for virtiofsd: https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/static-build/virtiofsd, keeping gnu and musl libc with their own dockerfiles as the alpine one is *super* simple.

PR #8926, opened 2024-01-26
gha: Cache the agent for non-x86_64 arches

fidencio:
Those are not yet being cached for no reason, and they better be as it'll allow us to save a considerable amount of time building the rootfs.

Fixes: #8917

NOTE: The agent-opa is not in the scope for this, and I will leave it for the maintainers of each specific architecture to do so, and to also add the dependency to their specific rootfs as done for this one in a previous commit.

### DIFF
--- a/.github/workflows/build-kata-static-tarball-arm64.yaml
+++ b/.github/workflows/build-kata-static-tarball-arm64.yaml
@@ -27,6 +27,7 @@ jobs:
     strategy:
       matrix:
         asset:
+          - agent
           - cloud-hypervisor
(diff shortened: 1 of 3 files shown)


gkurz (approved):
Good catch ! Thanks @fidencio !

stevenhorsman (approved):
LGTM. Thanks

BbolroC:
Why don't we merge this? `Static checks / build-checks (agent, sudo -E PATH="$PATH" make test, ppc64le)` has not been picked up quite a while.

Issue #8930, opened 2024-01-26
runtime: merged ValidCgroupPath method


yaoyinnan:
**Is your feature request related to a problem? Please describe.**

ValidCgroupPath is currently implemented separately by cgroupv1 and cgroupv2 and should be merged.

**Describe the solution you'd like**

Merged the ValidCgroupPath method of resource control in the go runtime and modified unit tests.

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

yes

PR #8931, opened 2024-01-26
runtime: merged ValidCgroupPath method

yaoyinnan:
Merged ValidCgroupPath method and modified unit tests.

Fixes: #8930

### DIFF
--- a/src/runtime/pkg/resourcecontrol/cgroups.go
+++ b/src/runtime/pkg/resourcecontrol/cgroups.go
@@ -136,7 +136,7 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 	var cgroupPath string
 
 	if cgroups.Mode() == cgroups.Legacy || cgroups.Mode() == cgroups.Hybrid {
-		cgroupPath, err = ValidCgroupPathV1(path, IsSystemdCgroup(path))
+		cgroupPath, err = ValidCgroupPath(path, false, IsSystemdCgroup(path))
 		if err != nil {
 			return nil, err
 		}
@@ -145,7 +145,7 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 			return nil, err
 		}
 	} else if cgroups.Mode() == cgroups.Unified {
-		cgroupPath, err = ValidCgroupPathV2(path, IsSystemdCgroup(path))
+		cgroupPath, err = ValidCgroupPath(path, true, IsSystemdCgroup(path))
 		if err != nil {
 			return nil, err
 		}
--- a/src/runtime/pkg/resourcecontrol/utils_linux.go
+++ b/src/runtime/pkg/resourcecontrol/utils_linux.go
@@ -21,35 +21,15 @@ import (
 // DefaultResourceControllerID runtime-determined location in the cgroups hierarchy.
 const DefaultResourceControllerID = "/vc"
 
-// ValidCgroupPathV1 returns a valid cgroup path for cgroup v1.
+// ValidCgroupPath returns a valid cgroup path.
 // see https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
-func ValidCgroupPathV1(path string, systemdCgroup bool) (string, error) {
+func ValidCgroupPath(path string, isCgroupV2 bool, systemdCgroup bool) (string, error) {
 	if IsSystemdCgroup(path) {
-		return path, nil
-	}
-
-	if systemdCgroup {
-		return "", fmt.Errorf("malformed systemd path '%v': expected to be of form 'slice:prefix:name'", path)
-	}
-
-	// In the case of an absolute path (starting with /), the runtime MUST
-	// take the path to be relative to the cgroups mount point.
-	if filepath.IsAbs(path) {
-		return filepath.Clean(path), nil
-	}
-
-	// In the case of a relative path (not starting with /), the runtime MAY
-	// interpret the path relative to a runtime-determined location in the cgroups hierarchy.
-	// clean up path and return a new path relative to DefaultResourceControllerID
-	return filepath.Join(DefaultResourceControllerID, filepath.Clean("/"+path)), nil
-}
-
-// ValidCgroupPathV2 returns a valid cgroup path for cgroup v2.
-// see https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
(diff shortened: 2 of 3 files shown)


src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -21,39 +21,15 @@ import (
...
-	// In the case of an absolute path (starting with /), the runtime MUST
-	// take the path to be relative to the cgroups mount point.
-	if filepath.IsAbs(path) {
-		return filepath.Clean(path), nil
-	}
-
-	// In the case of a relative path (not starting with /), the runtime MAY
-	// interpret the path relative to a runtime-determined location in the cgroups hierarchy.
-	// clean up path and return a new path relative to DefaultResourceControllerID
-	return filepath.Join(DefaultResourceControllerID, filepath.Clean("/"+path)), nil
```
justxuewei, line 44: Why are these removed? It gives a conversion from non-systemd-cgroup path.

justxuewei (commented):
Thanks @yaoyinnan! One comment here.

src/runtime/pkg/resourcecontrol/utils_linux.go
yaoyinnan: ValidCgroupPathV1 and ValidCgroupPathV2 are treated the same on non-systemd-cgroup paths. I didn't remove the handling of the non-systemd-cgroup path, just the duplicate part.

justxuewei (approved):
Lgtm, thanks @yaoyinnan!

Apokleos (commented):
Thx @yaoyinnan could you please split the PR into more one commits to make it clearer ? IMO, the code related to test should be moved into one commit ?

src/runtime/pkg/resourcecontrol/utils_linux.go
```diff
@@ -21,39 +21,15 @@ import (
 // DefaultResourceControllerID runtime-determined location in the cgroups hierarchy.
 const DefaultResourceControllerID = "/vc"
 
-// ValidCgroupPathV1 returns a valid cgroup path for cgroup v1.
+// ValidCgroupPath returns a valid cgroup path.
 // see https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
-func ValidCgroupPathV1(path string, systemdCgroup bool) (string, error) {
+func ValidCgroupPath(path string, v2 bool) (string, error) {
```
justxuewei, line 26: Looks like the logic was changed due to `systemdCgroup` removal

- before: the function **throws an error** if `systemdCgroup == true` and the path is not in the format of systemd cgroup.
- now: giving a new path without any errors.

justxuewei (changes requested):
Block this PR mergence until the above comment is addressed.

justxuewei:
Btw, your email in `sign-off-by` is incorrect. You should take a look at this as well.

src/runtime/pkg/resourcecontrol/utils_linux.go
yaoyinnan: @justxuewei Thank you for your review, this issue is indeed worthy of discussion.

The IsSystemdCgroup() is repeated. When calling the ValidCgroupPath() method, the incoming systemdCgroup is obtained through IsSystemdCgroup(path).
This logic seems to be intended to handle the following situation: when calling the `ValidCgroupPath()`, the parameter `systemdCgroup==true` is actively passed in, but in fact, `IsSystemdCgroup(path)` is judged to be `false` inside the function.

I'm not sure if I want to keep this design from before, please give advice, thanks.

justxuewei: I still think they are different. Let us list all possible cases here.

| index | path        | systemdCgroup | result             |
|-------|-------------|---------------|--------------------|
| 1     | systemd     | true          | (origin path, nil) |
| 2     | systemd     | false         | (origin path, nil) |
| 3     | non-systemd | true          | ("", error)        |
| 4     | non-systemd | false         | (new path, nil)    |

Case 3 means I expect the path is in the format of systemd, but it isn't actually. Therefore, an error should be thrown. The case is unreachable if `systemdCgroup` is removed. Wdyt?

yaoyinnan: You are right

yaoyinnan: I fixed this question and the email issue.

justxuewei (approved):
Lgtm, thanks!

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -136,7 +136,8 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 	var cgroupPath string
 
 	if cgroups.Mode() == cgroups.Legacy || cgroups.Mode() == cgroups.Hybrid {
-		cgroupPath, err = ValidCgroupPathV1(path, IsSystemdCgroup(path))
+		isCgroupV2 := false
+		cgroupPath, err = ValidCgroupPath(path, isCgroupV2, IsSystemdCgroup(path))
```
justxuewei, line 140: Could you remove this line of code?

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -145,7 +146,8 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 			return nil, err
 		}
 	} else if cgroups.Mode() == cgroups.Unified {
-		cgroupPath, err = ValidCgroupPathV2(path, IsSystemdCgroup(path))
+		isCgroupV2 := true
```
justxuewei, line 149: Ditto.

src/runtime/pkg/resourcecontrol/cgroups.go
```diff
@@ -136,7 +136,8 @@ func NewResourceController(path string, resources *specs.LinuxResources) (Resour
 	var cgroupPath string
 
 	if cgroups.Mode() == cgroups.Legacy || cgroups.Mode() == cgroups.Hybrid {
-		cgroupPath, err = ValidCgroupPathV1(path, IsSystemdCgroup(path))
+		isCgroupV2 := false
```
justxuewei, line 139: Could you remove this line of code?

yaoyinnan: removed

Apokleos (approved):
Thx @yaoyinnan LGTM

PR #8938, opened 2024-01-27
runtime-rs: Log qemu's stderr in shim log

pmores:
Having a ready access to qemu's stderr is immensely helpful in debugging qemu's launch failures.  Qemu's stderr output is now included in regular shim's log (look for lines prepended with "qemu stderr: ").

An unrelated small fix promised in review of PR #8185 is also included in this PR's initial commit.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -8,7 +8,7 @@ use crate::{
     hypervisor_persist::HypervisorState, HypervisorConfig, MemoryConfig, VcpuThreadIds,
     VsockDevice, HYPERVISOR_QEMU,
 };
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use kata_types::{
     capabilities::{Capabilities, CapabilityBits},
@@ -17,7 +17,11 @@ use kata_types::{
 use persist::sandbox_persist::Persist;
 use std::collections::HashMap;
 use std::os::unix::io::AsRawFd;
-use tokio::process::{Child, Command};
+use std::process::Stdio;
+use tokio::{
+    io::{AsyncBufReadExt, BufReader},
+    process::{Child, ChildStderr, Command},
+};
 
 const VSOCK_SCHEME: &str = "vsock";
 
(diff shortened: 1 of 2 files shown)


src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -270,6 +276,24 @@ impl QemuInner {
...
+
+    let stderr_reader = BufReader::new(stderr);
+    let mut stderr_lines = stderr_reader.lines();
+
+    while let Some(buffer) = stderr_lines
+        .next_line()
+        .await
+        .context("next_line() failed on qemu stderr")?
+    {
+        info!(sl!(), "qemu stderr: {:?}", buffer);
```
gkurz, line 290: The equivalent code for `virtiofsd` also filters out lines that would only contain _whitespaces_. I'm not sure how this could happen but I guess it wouldn't harm to do the same here.

zvonkok:
Just a nit, can we prefix PRs that target runtime-rs with `runtime-rs: ... whatever` I was like, huh, didn't (at)gkurz already added that to go-runtime QEMU ..

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
pmores: I've never seen qemu log an empty line (although I haven't seen virtiofsd to do that either :-)).  I also believe that logging empty lines would better be fixed in the program that does that. ;-)

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -17,7 +17,10 @@ use kata_types::{
 use persist::sandbox_persist::Persist;
 use std::collections::HashMap;
 use std::os::unix::io::AsRawFd;
-use std::process::Child;
+use tokio::{
+    io::{AsyncBufReadExt, BufReader},
+    process::{Child, ChildStderr, Command},
```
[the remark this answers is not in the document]
pmores: I missed that somehow, sorry, I'll fix it in my next push.

gkurz:
> I'm sorry, I thought the prefix only applies to commit messages and didn't realise that it applies to PR titles as well. Next time around (I don't think I can fix it for this one)!

Yes you can. There's an `Edit` button at the far right of the PR title. I've just done it.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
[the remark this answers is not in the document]
pmores: fixed

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
gkurz: > I've never seen qemu log an empty line (although I haven't seen virtiofsd to do that either :-)). I also believe that logging empty lines would better be fixed in the program that does that. ;-)

As any other command, QEMU is allowed to format the error message it prints on stderr as it wants.
These error messages are meant for a human user and can legitimately contain multiple lines and empty lines.

The code in this PR is collecting the QEMU stderr on a per-line basis and turns each line into a full shim log entry, i.e. a given QEMU error can span over multiple shim logs. We cannot do much about that but we could maybe avoid generating a log entry with no information in it... No big deal. We can still revisit this later, based on experience.

gkurz (approved):
LGTM. Thanks @pmores !

GabyCT (approved):

PR #8947, opened 2024-01-29
gha: azure: Set the correct subscription to the account

fidencio:
Due to the changes done in the CI, we need to set the correct subscription to be used with the account from now on, otherwise we'd end up using CoCo subscription.


jepio (approved):

stevenhorsman (approved):
LGTM. Thanks

fidencio:
NOTE: This PR will be merged without running the CI as it contains a change in the yaml file, and pull_request_target will not pick this up.

PR #8974, opened 2024-01-31
runtime: add SingleContainer when obtaining OCI Spec

yaoyinnan:
When creating a cgroup, add a SingleContainer when obtaining the OCI Spec to apply to ctr, podman, etc.

Fixes: #5240

### DIFF
--- a/src/runtime/virtcontainers/sandbox.go
+++ b/src/runtime/virtcontainers/sandbox.go
@@ -2576,11 +2576,12 @@ func (s *Sandbox) GetPatchedOCISpec() *specs.Spec {
 		return nil
 	}
 
-	// get the container associated with the PodSandbox annotation. In Kubernetes, this
-	// represents the pause container. In Docker, this is the container.
-	// On Linux, we derive the group path from this container.
+	// Get the container associated with the PodSandbox annotation.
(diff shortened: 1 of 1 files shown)


justxuewei (approved):
Lgtm, thanks!

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2576,11 +2576,11 @@ func (s *Sandbox) GetPatchedOCISpec() *specs.Spec {
 		return nil
 	}
 
-	// get the container associated with the PodSandbox annotation. In Kubernetes, this
-	// represents the pause container. In Docker, this is the container.
-	// On Linux, we derive the group path from this container.
+	// Get the container associated with the PodSandbox annotation.
+	// In Kubernetes, this represents the pause container. In Docker, this is the container.
```
Apokleos, line 2580: It seems that "In Docker"  is not appropriate today. Should we make it more generic?

yaoyinnan: Yes, I changed it to "CRI-compatible runtime, such as Containerd"

Apokleos (approved):
Thx @yaoyinnan LGTM!

Issue #8982, opened 2024-01-31
packaging: Build the a single confidential rootfs image / initrd


fidencio:
Let's use a single rootfs image / initrd for confidential workloads,
instead of having those split for different TEEs.

We can easily do this now as the soon-to-be-added guest-components can
be built in a generic way.

Issue #8183, opened 2023-10-09
WIP: CC: kata-deploy packaging of CC versions


stevenhorsman:
This will need input from @fidencio to get well spec'd but there are a few questions about kata-deploy and how the kata-cc stuff is packaged.

Here are some thoughts/comments I have so far:
- I think we've decided that we will add all the kata CC artefacts into the existing kata-deploy payload, which will inflate it quite a lot, but then aim to us NFD to reduce the amount that gets installed on user's machines, so they are not overwhelmed with new runtime classes.
- How does the coco operator consume kata-deploy? I think the gaps between kata-deploy and the kata-runtime-payload have got a lot smaller in the last release, so can we just get the operator to consume the kata-deploy image directly? or are changes needs to switch over? One of the differences was/is (for non default case) in the pre-install and post-install hooks due to where the containerd fork was installed and how kata-cleanup removes stuff, so it depends if we can get away with removing these once we switch to main, or we'll still need them for enclave-cc, gpu support?
- We might have CC specific guest images, with their own Attestation-agent KBC implementation in (see https://github.com/kata-containers/kata-containers/issues/8111#issuecomment-1751585623 for some discussion), so will we end up having duplicated runtime classes for these images, or aim to add the CC features into the main runtime class for each platform

fidencio:
> * We need to port support to kata-deploy for the snapshotter environment into kata-deploy

Let me add some notes to this one here.

Supporting snapshotter for kata-deploy requires:
* Dealing with containerd 1.6.x and higher
  * 1.6.x: setting one snapshotter globally
  * 1.7.x or newer: setting snapshotters per runtime handlers
    * This requires a beter mapping on how we proporly set things up
  
I'm doing a prototype of this, and will open a PR soon.

ariel-adam:
@stevenhorsman, after @fidencio provides a prototype, what should be the next steps?

stevenhorsman:
Fabiano provided the snapshotter support, which is why it's ticked off in the description. There is a question of if we want to support the TEE platform parameter like the attestation-agent does, or just flow it through, but that's probably related to the issue that will integrate the attestation-agent and CDH into the rootfs.

fidencio:
> I'm not sure about the NFD next steps. I don't think they are super priority for 0.9 release though

They're not a priority for 0.9!

PR #8978, opened 2024-01-31
runtime: packaging: Use confidential kernel instead of the TDX one

fidencio:
We've been building a 6.7 kernel for some time and it can be used for all the TEEs.  For now I'm just enabling this for TDX, at least till the SEV / SNP CIs are up and running again.

PR #8983, opened 2024-01-31
packaging: Add confidential image / initrd

fidencio:
Let's use a single rootfs image / initrd for confidential workloads,
instead of having those split for different TEEs.

We can easily do this now as the soon-to-be-added guest-components can
be built in a generic way.

Fixes: #8982

NOTE: This is rebased atop of #8978

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -49,8 +49,10 @@ jobs:
           - qemu-tdx-experimental
           - stratovirt
           - rootfs-image
+          - rootfs-image-confidential
           - rootfs-image-tdx
           - rootfs-initrd
+          - rootfs-initrd-confidential
           - rootfs-initrd-mariner
           - rootfs-initrd-sev
           - runk
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -112,8 +112,10 @@ options:
 	qemu-tdx-experimental
 	stratovirt
 	rootfs-image
+	rootfs-image-confidential
 	rootfs-image-tdx
 	rootfs-initrd
+	rootfs-initrd-confidential
 	rootfs-initrd-mariner
 	rootfs-initrd-sev
 	runk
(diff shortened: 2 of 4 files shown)


.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -38,9 +38,7 @@ jobs:
           - kata-ctl
           - kernel
           - kernel-confidential
-          - kernel-sev
```
beraldoleal, line 41: `tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh` still has some references to them. Should we remove there too?

fidencio: Very good point, @beraldoleal, and I should have written this as part of a note in the PR.
If we remove those right now, we'll break the build as:
* removing from the yaml file will only take effect after the PR is merged
* removong from the kata-deploy-binaries.sh will take effect during the tests

With this in mind, I'd like to remove those bits in a follow-up PR.

Does it sound reasonable?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -268,6 +276,12 @@ install_initrd() {
 	"${rootfs_builder}" --osname="${os_name}" --osversion="${os_version}" --imagetype=initrd --prefix="${prefix}" --destdir="${destdir}" --image_initrd_suffix="${variant}"
 }
 
+#Install guest initrd for sev
```
portersrc, line 279: "sev" should be "confidential guests" maybe?

portersrc (approved):
Looks good to me. I had the same comment as @beraldoleal but your point about a follow-up PR sounds good.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
fidencio: Fixed, thanks!

.github/workflows/build-kata-static-tarball-amd64.yaml
beraldoleal: Sounds acceptable! :) Thanks.

beraldoleal (approved):
LGTM.

fidencio:
I've moved this to a draft to ensure it doesn't get merged before its dependency.

fidencio:
I've updated this one dropping the AMD changes.

stevenhorsman (approved):
This LGTM.

stevenhorsman:
One thing that occurred to me whilst reviewing this (that is out of scope for this PR, but I thought might be needed to think about before we close #8183), is what are we going to do about SE image. It's not being made through this as it's only creating the confidential image in the amd64 payload, which makes sense, as the SE initrd needs to go through genprotimg tool and requires a host key document of the machine it will run on, so I'm not too sure how we want to integrate that into CI pipeline? In `CCv0` I think we just built the image from scratch during the CI phase, which I guess we could do, but I'm not sure if there is a better idea?

Anyway, that's not for this PR, so I'll try not to derail it

PR #8986, opened 2024-02-01
runtime-rs: fix interoperability issues between runtime-rs and cri-o

pmores:
runtime-rs cannot currently be launched by cri-o since cri-o passes an empty value for -address which runtime-rs incorrectly rejects due to an excessive -address value validation.  This PR fixes this along with a different but related problem of incorrect handling of the shim v2 protocol `TTRPC_ADDRESS` environment variable.

This also makes runtime-rs consistent with the go runtime which correctly tolerates an empty -address value passed by cri-o.

Thanks to @littlejawa for insights in cri-o implementation!


pmores:
@littlejawa @gkurz @fidencio please take a look

src/runtime-rs/crates/shim/src/args.rs
```diff
@@ -52,21 +51,6 @@ impl Args {
...
-            .metadata()
-            .context(Error::FileGetMetadata(format!("{:?}", path)))?;
-        if !md.file_type().is_socket() {
-            return Err(Error::InvalidArgument).context("address is not socket");
-        }
-        self.address = path
-            .to_str()
-            .map(|v| v.to_owned())
-            .ok_or(Error::InvalidArgument)?;
-
```
fidencio, line 69: @pmores, just checking if this is possible or not ... but can we warn those instead of erroring out?
By doing this we can make sure at least people will have enough info to debug in case something goes awry in the containerd case.

pmores: It can definitely be done, I'm just not sure though how useful it would be.  If I get this right, the value is fully owned by the corresponding runtime and its semantics are opaque to a shim.  If we don't know what the value can or cannot be we can't really check it, right?

In particular, cri-o intentionally sends an empty value and is just fine getting it back from the shim.  Warning about the `-address` argument not being a socket would then mean cri-o users being warned of something that's meant to be that way.

How the go shim handles this could be a useful guideline here. I haven't been able to figure that out with any certainty although I didn't see a warning looking into what I _think_ was the pertinent code.  One thing I know is that it doesn't seem to error out.

littlejawa (approved):
LGTM. Thanks @pmores !

src/runtime-rs/crates/shim/src/args.rs
```diff
@@ -52,21 +51,6 @@ impl Args {
...
         validate::verify_id(&self.namespace).context("verify namespace")?;
 
-        // Ensure `address` is a valid path.
-        let path = PathBuf::from(self.address.clone())
-            .canonicalize()
-            .context(Error::InvalidPath(self.address.clone()))?;
-        let md = path
-            .metadata()
-            .context(Error::FileGetMetadata(format!("{:?}", path)))?;
-        if !md.file_type().is_socket() {
```
gkurz, line 62: This is the last user of `os::unix::fs::FileTypeExt` as detected by CI :
```
error: unused import: `os::unix::fs::FileTypeExt`
 --> crates/shim/src/args.rs:7:11
  |
7 | use std::{os::unix::fs::FileTypeExt, path::PathBuf};
  |           ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
```

src/runtime-rs/crates/shim/src/args.rs
```diff
@@ -40,7 +40,6 @@ impl Args {
     pub fn validate(&mut self, should_check_bundle: bool) -> Result<()> {
         if self.id.is_empty()
             || self.namespace.is_empty()
-            || self.address.is_empty()
             || self.publish_binary.is_empty()
```
gkurz, line 43: rustfmt suggests the following change :

```
-        if self.id.is_empty()
-            || self.namespace.is_empty()
-            || self.publish_binary.is_empty()
-        {
+        if self.id.is_empty() || self.namespace.is_empty() || self.publish_binary.is_empty() {
```

src/runtime-rs/crates/shim/src/args.rs
gkurz: I fully agree with @pmores and @littlejawa. This address is obviously something that should be considered as an opaque blob by the shim. Dropping any code that makes assumption on its content is definitely a good thing.

gkurz: > I fully agree with @pmores and @littlejawa. This address is obviously something that should be considered as an opaque blob by the shim. Dropping any code that makes assumption on its content is definitely a good thing.

... which make me think that the test code in this file should be adapted to your change. I'll submit a patch based on your PR.

littlejawa:
nit: in the commit message of your [second commit](https://github.com/kata-containers/kata-containers/pull/8986/commits/cd5f9609e0d735482d337d30e790304437b12eeb):
"s/exactl/exact/"

src/runtime-rs/crates/shim/src/args.rs
pmores: I've never understood how an idea such as treating warnings as errors managed not to die immediately... _sigh_ I'll fix it.

src/runtime-rs/crates/shim/src/args.rs
fidencio: That's fine by me.

gkurz:
@studychao @amshinde @justxuewei this PR already has 2  approvals and CI passed but I'd still like to hear from you and make sure this won't break the runtime-rs experience with containerd. My current goal is to merge this tomorrow. Please provide feeback ASAP.

src/runtime-rs/crates/service/src/event.rs
```diff
@@ -38,12 +38,12 @@ pub(crate) trait Forwarder {
...
 pub(crate) async fn new_event_publisher(namespace: &str) -> Result<Box<dyn Forwarder>> {
     let fwd: Box<dyn Forwarder> = match env::var(TTRPC_ADDRESS_ENV) {
-        Ok(address) => Box::new(
+        Ok(address) if !address.is_empty() => Box::new(
             ContainerdForwarder::new(namespace, &address)
                 .await
                 .context("new containerd forwarder")?,
         ),
-        Err(_) => Box::new(
+        Ok(_) | Err(_) => Box::new(
```
studychao, line 46: maybe some comment here would be better to tell why do you take `Ok(_)` here after checking `Ok(address)`

studychao (approved):
LGTM but one small nit

gkurz:
> > @studychao @amshinde @justxuewei this PR already has 2 approvals and CI passed but I'd still like to hear from you and make sure this won't break the runtime-rs experience with containerd.
> 
> For what it's worth, I launch runtime-rs both from cri-o and containerd and haven't run into any problems as far as I can tell.

Thanks @pmores.

The motivations behind my question are:
# stay on the safe side as our RH experience is mostly about using cri-o
# politely inform regular users of containerd that we are doing this change

justxuewei (approved):
Lgtm, thanks!

amshinde (approved):

PR #8990, opened 2024-02-01
gha: wait az resource group be created

### LINKED ISSUE #8989: CI: azure aks deploy sometimes fail to find the resources group
Recently I've noticed some CI jobs that relies on AKS provisioning failing because the Azure resource group (which is created on-demand before creating the cluster) is not found:

For example, https://github.com/kata-containers/kata-containers/actions/runs/7724334694/job/21074189970?pr=8839 and https://github.com/kata-containers/kata-containers/actions/runs/7722583170/job/21076489139?pr=8974, you will see an error similar to:

```
ERROR: (ResourceGroupNotFound) Resource group 'kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found.
Code: ResourceGroupNotFound
Message: Resource group 'kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found.
{
  "id": "/subscriptions/***/resourceGroups/kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s",
  "location": "eastus2",
  "managedBy": null,
  "name": "kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s",
  "properties": {
  },
  "tags": null,
}
WARNING: The behavior of this command has been altered by the following extension: aks-preview
WARNING: SSH key files '/home/runner/.ssh/id_rsa' and '/home/runner/.ssh/id_rsa.pub' have been generated under ~/.ssh to allow SSH access to the VM. If using machines without permanent storage like Azure Cloud Shell without an attached file share, back up your keys to a safe location
ERROR: (ResourceGroupNotFound) Create or update resource group node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s failed. autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found."
Code: ResourceGroupNotFound
Message: Create or update resource group node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s failed. autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found."
Exception Details:	(ResourceGroupNotFound) Resource group 'node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found.
	Code: ResourceGroupNotFound
	Message: Resource group 'node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found.
Error: Process completed with exit code 1.
```

On https://github.com/kata-containers/kata-containers/blob/main/tests/gha-run-k8s-common.sh#L68 it is **requested** the creation the aforementioned resource group but it doesn't wait the resource to effectively exist. That might be the problem.

wainersm:
To provision k8s on azure (AKS) there should be created a temporary resources group before. The script sends the request to get it created but doesn't wait the operation to finish, so sometimes it tries to use a resources group what doesn't exist and then bail out.

This added some `az group wait` check points. Even on deletion we want to ensure there won't be dangling resources groups.

Fixes #8989

---

By default it will wait 30 seconds (polling every 10 seconds) for the resources group to be created/deleted. **I don't know if it enough or too much**, please advise.

### DIFF
--- a/tests/gha-run-k8s-common.sh
+++ b/tests/gha-run-k8s-common.sh
@@ -11,6 +11,10 @@ set -o pipefail
 tests_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "${tests_dir}/common.bash"
 
+# The next two variables define, in seconds, the polling interval and timeout
+# for waiting the az group creation and deletion.
+AZ_GROUP_INTERVAL=${AZ_GROUP_INTERVAL:-10}
+AZ_GROUP_TIMEOUT=${AZ_GROUP_TIMEOUT:-30}
 K8S_TEST_HOST_TYPE="${K8S_TEST_HOST_TYPE:-small}"
 
 function _print_instance_type() {
@@ -69,6 +73,11 @@ function create_cluster() {
         -l eastus2 \
         -n "${rg}"
 
+    az group wait --exists \
+        -g "${rg}" \
+        --interval "${AZ_GROUP_INTERVAL}" \
+        --timeout "${AZ_GROUP_TIMEOUT}"
+
     az aks create \
         -g "${rg}" \
 	--node-resource-group "node-${rg}" \
@@ -104,10 +113,17 @@ function get_cluster_credentials() {
 
 function delete_cluster() {
     test_type="${1:-k8s}"
+	local rg
+	rg="$(_print_rg_name ${test_type})"
 
     az group delete \
-        -g "$(_print_rg_name ${test_type})" \
+        -g "${rg}" \
         --yes
+
+    az group wait --deleted \
+        -g "${rg}" \
+        --interval "${AZ_GROUP_INTERVAL}" \
+        --timeout "${AZ_GROUP_TIMEOUT}"
 }
 
 function delete_cluster_kcli() {


beraldoleal (approved):
Hi @wainersm this LGTM, thanks.

How those errors are treated by the CI/tests?

I mean.. are tests flagged as FAIL or ERROR? I'm not sure if this is the case here but I remember seeing before some errors during the bootstrap/teardown that flags the tests as failed and it was hard to say if it was the test or the bootstrap (without debugging).

But it would be nice if we had the ability to differentiate boostrap/teardown from the test itself. Usually we differ that by using ERROR vs FAIL. Do you know if we are using that already?

Isn't a requirement for this PR, of course, I'm just taking the hook here.

wainersm:
> Hi @wainersm this LGTM, thanks.
> 
> How those errors are treated by the CI/tests?
> 
> I mean.. are tests flagged as FAIL or ERROR? I'm not sure if this is the case here but I remember seeing before some errors during the bootstrap/teardown that flags the tests as failed and it was hard to say if it was the test or the bootstrap (without debugging).
>

That's still  the case either the CI job ends with `success` or `failure`, there isn't any intermediate status like `error` on Github workflows.

> But it would be nice if we had the ability to differentiate boostrap/teardown from the test itself. Usually we differ that by using ERROR vs FAIL. Do you know if we are using that already?
>

With the CI in Github workflows (vs Jenkins), the separation between bootstrap/teardown and tests run is better. Most of the jobs encapsulate the bootstrap/teardown of the test environment on [job steps](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsteps) that on fail will result on the entire job failure (unless the step is `continue-on-error`). Job steps are reported individually on the UI so the logs of the failed step won't mix with the previous executed steps, which is good.

Our "test harness" currently is a mix of scripts, bats and github workflows.  Unless we use a truly test harness (e.g. Avocado), I don't think we will be able to implement this `error` vs `fail` mechanism you are thinking of.

> Isn't a requirement for this PR, of course, I'm just taking the hook here.

It's a valid discussion indeed. Please don't hesitate to ask and/or share your ideas. ;)

tests/gha-run-k8s-common.sh
```diff
@@ -69,6 +73,11 @@ function create_cluster() {
         -l eastus2 \
         -n "${rg}"
 
+    az group wait --exists \
+        -g "${rg}" \
+        --interval "${AZ_GROUP_INTERVAL}" \
+        --timeout "${AZ_GROUP_TIMEOUT}"
+
```
sprt, line 80: This isn't the problematic RG - looking at the [failure you linked](https://github.com/kata-containers/kata-containers/actions/runs/7724334694/job/21074189970?pr=8839) in the associated issue, there are 2 errors logged but only one is fatal:

1. `ERROR: (ResourceGroupNotFound) Resource group 'kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found.` This is logged by the cleanup step below before we actually try to create the cluster and that is fine: https://github.com/kata-containers/kata-containers/blob/7ec99ce2bbb2fc1de385d48d01b21ebd55b0c29a/tests/gha-run-k8s-common.sh#L64-L68
2. `ERROR: (ResourceGroupNotFound) Create or update resource group node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s failed. autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found."` Note the RG name is `node-...`, referring to the RG passed via `--node-resource-group`. This RG contains the actual cluster nodes, rather than the AKS instance. This is the fatal error.

I've looked internally and it seems this a known new issue and a hotfix is rolling out. I've also confirmed that the client is not supposed to create this node RG. I propose to either:

1. Do nothing and wait till the hotfix is released. I will try to find out when that is.
2. Refactor this PR to create the node RG manually ourselves regardless, but I'm not sure this is supported, and even if it is we could run into the same AKS is running into and have to setup retry logic. I prefer option 1.

sprt (changes requested):
Think we should off on this right now - see my comments. 🙂

tests/gha-run-k8s-common.sh
wainersm: Hi @sprt !

> This isn't the problematic RG - looking at the [failure you linked](https://github.com/kata-containers/kata-containers/actions/runs/7724334694/job/21074189970?pr=8839) in the associated issue, there are 2 errors logged but only one is fatal:
> 
>     1. `ERROR: (ResourceGroupNotFound) Resource group 'kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found.` This is logged by the cleanup step below before we actually try to create the cluster and that is fine: https://github.com/kata-containers/kata-containers/blob/7ec99ce2bbb2fc1de385d48d01b21ebd55b0c29a/tests/gha-run-k8s-common.sh#L64-L68
> 

oh, I saw the first error and concluded it was the fatal (next being like a cascaded error message). I was mislead by the `delete_cluster || true`. I will add some guards within the `delete_cluster()` to only delete resources that exist, so do not mess with the logs.

Still regarding this PR, do you think it is worthless wait for the RG be created even though it's not causing problems now? It might cause future problems... unless the `az group create` is guaranteed to return to the caller after the group is effectively created. Is it the case?
 
>     2. `ERROR: (ResourceGroupNotFound) Create or update resource group node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s failed. autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'node-kataCI-k8s-8839-27c908f10757-qemu-ubuntu-amd64-s' could not be found."` Note the RG name is `node-...`, referring to the RG passed via `--node-resource-group`. This RG contains the actual cluster nodes, rather than the AKS instance. This is the fatal error.
> 
> 
> I've looked internally and it seems this a known new issue and a hotfix is rolling out. I've also confirmed that the client is not supposed to create this node RG. I propose to either:
> 
>     1. Do nothing and wait till the hotfix is released. I will try to find out when that is.
> 
>     2. Refactor this PR to create the node RG manually ourselves regardless, but I'm not sure this is supported, and even if it is we could run into the same AKS is running into and have to setup retry logic.

Since that the client is not supposed to create the node RG, I suggest to not adopt 2).

sprt: > oh, I saw the first error and concluded it was the fatal (next being like a cascaded error message). I was mislead by the `delete_cluster || true`. I will add some guards within the `delete_cluster()` to only delete resources that exist, so do not mess with the logs.

Gotcha. We could also redirect the output to /dev/null - your call.

> Still regarding this PR, do you think it is worthless wait for the RG be created even though it's not causing problems now? It might cause future problems... unless the `az group create` is guaranteed to return to the caller after the group is effectively created. Is it the case?

Personally I wouldn't add more complexity to this function for that purpose. I'm no `az` expert but it seems those commands are meant to be synchronous and `az group wait` would be for concurrent scripts - historically we haven't had issues. Maybe the folks over at [Azure/azure-cli](https://github.com/Azure/azure-cli) would be interested in adding more docs here.

beraldoleal:
> > Hi @wainersm this LGTM, thanks.
> > How those errors are treated by the CI/tests?
> > I mean.. are tests flagged as FAIL or ERROR? I'm not sure if this is the case here but I remember seeing before some errors during the bootstrap/teardown that flags the tests as failed and it was hard to say if it was the test or the bootstrap (without debugging).
> 
> That's still the case either the CI job ends with `success` or `failure`, there isn't any intermediate status like `error` on Github workflows.

Yes, I understand the GH limitations. I was not expecting this to be solved on the GH, but a simple junit (or another artifact) could solve this. Either to parse it locally or to parse it as part of an additional step to provide this differentiation.

> Our "test harness" currently is a mix of scripts, bats and github workflows. Unless we use a truly test harness (e.g. Avocado), I don't think we will be able to implement this `error` vs `fail` mechanism you are thinking of.

Completely understand the complexity due to our current situation. No worries.
I think we might be able to archive something better, without a full test harness.
For instance bats supports junit output, and if the setup has skips, maybe we could differentiate those at the artifact level, at least.

Also, I know we GH has some extensions to parse those junit and present them nicely. But I haven't tested them.

Again I'm hijacking the PR with this discussion. So no worries, not expecting a reply here, we can discuss this in a different place.

tests/gha-run-k8s-common.sh
wainersm: > > oh, I saw the first error and concluded it was the fatal (next being like a cascaded error message). I was mislead by the `delete_cluster || true`. I will add some guards within the `delete_cluster()` to only delete resources that exist, so do not mess with the logs.
> 
> Gotcha. We could also redirect the output to /dev/null - your call.
>

Gonna send in another PR.
 
> > Still regarding this PR, do you think it is worthless wait for the RG be created even though it's not causing problems now? It might cause future problems... unless the `az group create` is guaranteed to return to the caller after the group is effectively created. Is it the case?
> 
> Personally I wouldn't add more complexity to this function for that purpose. I'm no `az` expert but it seems those commands are meant to be synchronous and `az group wait` would be for concurrent scripts - historically we haven't had issues. Maybe the folks over at [Azure/azure-cli](https://github.com/Azure/azure-cli) would be interested in adding more docs here.

ok, I will drop this PR entirely. Much appreciated your comments here @sprt , thanks!

wainersm:
hi @beraldoleal !

> > > Hi @wainersm this LGTM, thanks.
> > > How those errors are treated by the CI/tests?
> > > I mean.. are tests flagged as FAIL or ERROR? I'm not sure if this is the case here but I remember seeing before some errors during the bootstrap/teardown that flags the tests as failed and it was hard to say if it was the test or the bootstrap (without debugging).
> > 
> > 
> > That's still the case either the CI job ends with `success` or `failure`, there isn't any intermediate status like `error` on Github workflows.
> 
> Yes, I understand the GH limitations. I was not expecting this to be solved on the GH, but a simple junit (or another artifact) could solve this. Either to parse it locally or to parse it as part of an additional step to provide this differentiation.
> 
> > Our "test harness" currently is a mix of scripts, bats and github workflows. Unless we use a truly test harness (e.g. Avocado), I don't think we will be able to implement this `error` vs `fail` mechanism you are thinking of.
> 
> Completely understand the complexity due to our current situation. No worries. I think we might be able to archive something better, without a full test harness. For instance bats supports junit output, and if the setup has skips, maybe we could differentiate those at the artifact level, at least.
> 
> Also, I know we GH has some extensions to parse those junit and present them nicely. But I haven't tested them.
> 
> Again I'm hijacking the PR with this discussion. So no worries, not expecting a reply here, we can discuss this in a different place.

Could you please open issues for those ideas so that we can move the discussion there?

PR #8994, opened 2024-02-01
ci: aks: switch from eastus2 to eastus region

sprt:
This addresses an internal AKS issue that intermittently prevents clusters from getting created. The fix has been rolled out to eastus but not yet eastus2, so we unblock the CI by switching. No downsides in general.

This supersedes #8990.

Fixes: #8989


fidencio (approved):
lgtm, thanks @sprt!

PR #8996, opened 2024-02-01
gha: k8s: Add cloud-hypervisor (runtime-rs) support

### LINKED ISSUE #8995: gha: k8s: Add cloud-hypervisor (runtime-rs) support without devmapper

GabyCT:
This PR adds the Cloud Hypervisor driver, integrated with the runtime-rs, as part of the kubernetes tests different without devmapper.

Fixes #8995

### DIFF
--- a/.github/workflows/run-k8s-tests-on-aks.yaml
+++ b/.github/workflows/run-k8s-tests-on-aks.yaml
@@ -37,6 +37,7 @@ jobs:
           - dragonball
           - qemu
           - stratovirt
+          - cloud-hypervisor
         instance-type:
           - small
           - normal


justxuewei (approved):
Lgtm, thanks!

fidencio (approved):
lgtm, thanks @GabyCT!

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
           - dragonball
           - qemu
           - stratovirt
+          - cloud-hypervisor
```
sprt, line 40: @GabyCT What's the difference between cloud-hypervisor and clh in this matrix?

jodh-intel: @sprt - `clh` refers to the golang runtime (the current default) configuration for Cloud Hypervisor, whereas `cloud-hypervisor` refers to the rust runtime (runtime-rs) configuration for Cloud Hypervisor. The names are different since they refer to different config files for each runtime since the runtimes are  very slightly different externally.

We could try to make both config formats the same, but that could end up causing a lot of confusion, particularly when we start thinking about changing the default Kata runtime, so the current view is that having separate configuration files for each runtime will actually help minimise confusion (but note this is still open to discussion).

For further information, see #5954 and #6020, and #8483 where we force the separation (for clarity) between the golang and rust runtime config files by requiring them to be in different directories.

Issue #8999, opened 2024-02-02
`kata-runtime exec` error message isn't correct


WeiZhang555:
# Description of problem

Following instructions from https://github.com/kata-containers/kata-containers/blob/main/docs/Developer-Guide.md#simple-debug-console-setup 

I'm trying to connect to debug console of one VM with modified rootfs, it reports:

```
#kata-runtime exec c649a44d3ff5fb8513c6b6814fe078f9232f1762374ca56a48e2a015c5d2029c
Get "http://shim/agent-url": dial unix /run/kata/c649a44d3ff5fb8513c6b6814fe078f9232f1762374ca56a48e2a015c5d2029c/shim-monitor.sock: connect: no such file or directory
```

/run/kata directory doesn't exist at all, even though I switch the rootfs to correct one and successfully exec into the VM.

After some search, I found that the `shim-monitor.sock` resides in 
```
#find /run/ -name shim-monitor.sock
/run/vc/sbs/64f59bdbb91b1ad892e1d3988ebc570dd03d42361812c550bb2263313833a929/shim-monitor.sock
```

The correct dir should be `/run/vc/sbs/`

# Expected result

The error message should be 
```
#kata-runtime exec c649a44d3ff5fb8513c6b6814fe078f9232f1762374ca56a48e2a015c5d2029c
Get "http://shim/agent-url": dial unix /run/vc/sbs/c649a44d3ff5fb8513c6b6814fe078f9232f1762374ca56a48e2a015c5d2029c/shim-monitor.sock: connect: no such file or directory
```

Apokleos:
related issue [#6622](https://github.com/kata-containers/kata-containers/issues/6622), [#6148](https://github.com/kata-containers/kata-containers/issues/6148)

PR #9015, opened 2024-02-04
runtime: display accurate error msg to avoid misleading users.

Apokleos:
runtime: display accurate error msg to avoid misleading users.
    
    The original handling method does not reach user expectations.
    When the ClientSocketAddress method stats the corresponding
    path of runtime-rs and has not found it yet, we should return
    an error message here that includes the reason for the failure
    (which should be an error display indicating that both runtime-go
    and runtime-rs were not found). Instead of simply displaying the
    corresponding path of runtime-rs as the final error message to
    users.
    It is also necessary to return the error promptly to the caller
    for further error handling.
    
    Fixes: #8999

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/shim_management.go
+++ b/src/runtime/pkg/containerd-shim-v2/shim_management.go
@@ -338,12 +338,16 @@ func ServerSocketAddress(id string) string {
 // shim management endpoint
 // NOTE: this code allows various go clients, e.g. kata-runtime or kata-monitor commands, to
 // connect to the rust shim management implementation.
-func ClientSocketAddress(id string) string {
+func ClientSocketAddress(id string) (string, error) {
 	// get the go runtime uds path
 	socketPath := SocketPathGo(id)
 	// if the path not exist, use the rust runtime uds path instead
 	if _, err := os.Stat(socketPath); err != nil {
 		socketPath = SocketPathRust(id)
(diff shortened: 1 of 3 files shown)


lifupan (approved):

justxuewei (approved):
Lgtm, thanks!

src/runtime/pkg/containerd-shim-v2/shim_management.go
```diff
@@ -338,12 +338,16 @@ func ServerSocketAddress(id string) string {
...
 // connect to the rust shim management implementation.
-func ClientSocketAddress(id string) string {
+func ClientSocketAddress(id string) (string, error) {
 	// get the go runtime uds path
 	socketPath := SocketPathGo(id)
 	// if the path not exist, use the rust runtime uds path instead
 	if _, err := os.Stat(socketPath); err != nil {
 		socketPath = SocketPathRust(id)
+		if _, err := os.Stat(socketPath); err != nil {
+			return "", fmt.Errorf("It fails to stat both %s and %s with error %v.", SocketPathGo(id), SocketPathRust(id), err)
```
WeiZhang555, line 348: Error message starts with lower case generally:

```
return "", fmt.Errorf("it fails...
```

WeiZhang555 (commented):
One comment, otherwise LGTM

PR #9019, opened 2024-02-05
gha: add GOPATH env var to the ppc64le k8s workflow

Amulyam24:
The filtering of testing cases installs/uses yq and expects GOPATH to be present. Hence, add it to the workflow.

Fixes: #9018

### DIFF
--- a/.github/workflows/run-k8s-tests-on-ppc64le.yaml
+++ b/.github/workflows/run-k8s-tests-on-ppc64le.yaml
@@ -37,6 +37,7 @@ jobs:
       DOCKER_REPO: ${{ inputs.repo }}
       DOCKER_TAG: ${{ inputs.tag }}
       PR_NUMBER: ${{ inputs.pr-number }}
+      GOPATH: ${{ github.workspace }}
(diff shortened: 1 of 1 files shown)


.github/workflows/run-k8s-tests-on-ppc64le.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
       DOCKER_REPO: ${{ inputs.repo }}
       DOCKER_TAG: ${{ inputs.tag }}
       PR_NUMBER: ${{ inputs.pr-number }}
+      GOPATH: ${{ github.workspace }}
```
fidencio, line 40: out of curiosity, why?
do we build something as part of this workflow?

Amulyam24: @fidencio, check this out - https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/filter_k8s_test.sh#L25-L31

This is mainly used on the non x86 arches to build/use `yq` which is used to filter out certain test cases.

fidencio (approved):
lgtm, thanks @Amulyam24!

Issue #8998, opened 2024-02-02
static-checks: CI slowdown from unconditional job triggers on self-hosted runner


BbolroC:
**Which feature do you think can be improved?**

The event `pull_request` for `Static checks / build-checks` triggers the workflow whenever a PR is updated. A job `build-checks` is followed by 36 sub-jobs. If multiple PRs are actively updated, an event for each PR would add up another set of CI jobs (after cancelling the previous set). The cumulative CI jobs would be:

- 36 jobs x # of PRs actively updated

While this has been tolerated with GitHub-hosted `ubuntu-20.04` runners, it poses challenges for self-hosted runners (e.g., `arm-no-k8s`, `s390x` and `ppc64le`) due to restrictions on the number of instance provisioning.

This inevitably creates a bottleneck for the entire CI process, significantly impacting the speed of the `Kata Containers CI`. We could easily observe that some workflows take longer than 5 hours to complete, which is not the desired outcome of migrating CI from Jenkins to GHA.

**How can it be improved?**

To alleviate the bottleneck, we could refactor the workflow so that the `static-checks` runs on the self-hosted runner only when triggered by the `ok-to-test` label. It would be sensible to maintain the jobs running on `ubuntu-20.04` as is, as it can serve as a preemptive measure for the others without causing a slowdown in the CI process.

CC: @wainersm @fidencio @stevenhorsman

BbolroC:
FYI: discussion on Slack https://katacontainers.slack.com/archives/C01HUGDR0CE/p1706706952999819

BbolroC:
FYI: This issue has been identified after https://github.com/kata-containers/kata-containers/pull/8485 was merged.

PR #9020, opened 2024-02-05
gha: Run static-checks on self-hosted runners conditionally

BbolroC:
Due to the restrictions on instance provisioning for self-hosted runners, performing static checks (36 jobs at the time of writing) on them each time a PR is updated could significantly burden them, consequently slowing down the entire CI system. To address this, the decision is to trigger these checks only when an 'ok-to-test' label is added. Meanwhile, the checks for x86_64, which are supported by GitHub-hosted runners, will remain unchanged.

Fixes: #8998


.github/workflows/static-checks-self-hosted.yaml
```diff
@@ -0,0 +1,130 @@
+on:
+  pull_request_target:
```
wainersm, line 2: hi @BbolroC , first of all, this is a critical problem so thanks for raising it (and fixing it!).

The static-check jobs don't rely on secrets so there isn't need to run on `pull_request_target`, we can run on `pull_request` and avoid all the nightmares bought by the former event.

.github/workflows/static-checks-self-hosted.yaml
```diff
@@ -0,0 +1,130 @@
...
+      - reopened
+      - labeled # a workflow runs only when the 'ok-to-test' label is added
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+name: Static checks
+jobs:
+  build-checks:
```
wainersm, line 18: You will need to guard this job with `if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}` (e.g. in https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/ci-on-push.yaml#L23) otherwise it will be triggered to any label.

Related with the label, does it make sense to use the same `ok-to-test` label to trigger the static-checks? Because that same label triggers the e2e jobs, all will race for the same bare-metal machines at same time. What's your opinion?

.github/workflows/static-checks-self-hosted.yaml
```diff
@@ -0,0 +1,130 @@
...
+    strategy:
```
wainersm, line 19: The `build-checks` workflow is almost a complete copy of  https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/static-checks.yaml . Changes on static-checks will be carried out on both files from now onward, which isn't good.

How feasible it would be to convert https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/static-checks.yaml into a callable workflow (`workflow_call`)? Then both static-checks and static-checks-self-hosted would call it passing the `instance` as an input.

BbolroC: Sounds much better. I will update the PR soon. Thanks for the feedback. 😉

.github/workflows/static-checks-self-hosted.yaml
BbolroC: Oh, I didn't know that. Thanks for the info. I will change it back to `pull_request` again.

.github/workflows/static-checks-self-hosted.yaml
BbolroC: I think adding another `label` would make reviewers confused in term of:

- Which label should be added first and under which circumstance (precedence issue)

And once both labels are added, two sets of tests will react to the same set of events, which race for the same resources in the end.

I will keep the config for `ok-to-test`. But I won't make any objection against your suggestion. Thanks.

.github/workflows/static-checks-self-hosted.yaml
```diff
@@ -0,0 +1,130 @@
...
+          sudo chown -R $USER:$USER $GITHUB_WORKSPACE $HOME
+          sudo rm -rf $GITHUB_WORKSPACE/*
+          sudo rm -f /tmp/kata_hybrid*  # Sometime we got leftover from test_setup_hvsock_failed()
+
+      - name: Checkout the code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Rebase atop of the latest target branch
```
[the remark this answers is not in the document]
BbolroC: Yes, I've removed it. Thanks!

.github/workflows/static-checks-self-hosted.yaml
wainersm: Good point @BbolroC . An extra label means extra complex == more confusion :)

Let's stick with 'ok-to-test` then.

BbolroC:
All checks are passed as planned.

stevenhorsman (approved):
LGTM. Thanks @BbolroC & @wainersm!

wainersm (approved):
Good job @BbolroC !

wainersm:
Hi @stevenhorsman ! We will need to re-do the list of required jobs as the actual `build-checks` changed of name.

stevenhorsman:
> Hi @stevenhorsman ! We will need to re-do the list of required jobs as the actual `build-checks` changed of name.

I can disable them now and then in ~10 days we'll re-enable them to give people grace period to re-base?

stevenhorsman:
> > Hi @stevenhorsman ! We will need to re-do the list of required jobs as the actual `build-checks` changed of name.
> 
> I can disable them now and then in ~10 days we'll re-enable them to give people grace period to re-base?

Hmm, given there are so many, I might raise this on the AC meeting first.

wainersm:
> > > Hi @stevenhorsman ! We will need to re-do the list of required jobs as the actual `build-checks` changed of name.
> > 
> > 
> > I can disable them now and then in ~10 days we'll re-enable them to give people grace period to re-base?
> 
> Hmm, given there are so many, I might raise this on the AC meeting first.

Yeah, it would be better to discuss on AC. I completely overlooked the actual PRs, they will need all to be rebased :(

BbolroC:
Yeah, again, this is something @gkurz also wanted to discuss on the AC meeting. 😭

PR #9026, opened 2024-02-05
packaging: Remove leftovers from the transition from TEE specific kernel / initrd / image to the "confidential" ones

fidencio:
Please, see each commit message for more details.


ChengyuZhu6 (approved):
LGTM, thanks @fidencio. One minor comment:  
You may need to update `kata_config_version`.

zvonkok (approved):
LGTM, thanks

niteeshkd:
What about removing the following sev/snp entries under kernel in versions.yaml? 
```
   sev:
      description: "Linux kernel that supports SEV and SNP"
      url: "https://cdn.kernel.org/pub/linux/kernel/v5.x/"
      version: "v5.19.2"
    snp:
      description: "Linux kernel that supports AMD SEV-SNP for VMs"
      url: "https://cdn.kernel.org/pub/linux/kernel/v5.x/"
      version: "v5.19.2"
```

fidencio:
> What about removing the following sev/snp entries under kernel in versions.yaml?

Very good call, and while doing this I also removed:
* kernel-tdx-experimental, as the confidential one is used instead
* image / initrd for tdx / sev, as the confidential one is used instead

GabyCT (approved):

tools/packaging/kernel/build-kernel.sh
```diff
@@ -110,7 +110,7 @@ Options:
 	-t <hypervisor>	: Hypervisor_target.
 	-u <url>	: Kernel URL to be used to download the kernel tarball.
 	-v <version>	: Kernel version to use if kernel path not provided.
-	-x <type>	: Confidential guest protection type, such as sev, snp, tdx, or "confidential" (for all of those).
+	-x <type>	: All the confidential guest protection type for a specific architecture.
```
jodh-intel, line 113: Nit: This looks a bit odd as the only value currently is "confidential" so could we either remove the `<type>` arg here, or specify in the usage that the value must be "confidential"?

jodh-intel (approved):
Hi, @fidencio - just one nit, but overall...

lgtm

tools/packaging/kernel/build-kernel.sh
fidencio: I've removed the type, thanks for the suggestion!

Issue #8887, opened 2024-01-23
tests: genpolicy: k8s-empty-dirs.bats doesn't work with automatically-generated Policy


danmihai1:
I don't know yet what the root cause is, but k8s-empty-dirs.bats is not working well when auto-generating a CoCo Policy for it.

danmihai1:
Also, this part of the same test script needs some work - to auto-generate policy before creating the pod.

	sed -e "s#\${agnhost_image}#${image}#" "$pod_file" |\
		kubectl create -f -

PR #9029, opened 2024-02-06
genpolicy: mount source for non-confidential guest

danmihai1:
The emergent Kata CI tests for Policy use confidential_guest = false in genpolicy-settings.json. That value is inconsistent with the following mount settings:

        "emptyDir": {
            "mount_type": "local",
            "mount_source": "^$(cpath)/$(sandbox-id)/local/",
            "mount_point": "^$(cpath)/$(sandbox-id)/local/",
            "driver": "local",
            "source": "local",
            "fstype": "local",
            "options": [
                "mode=0777"
            ]
        },

We need to keep those settings for confidential_guest = true, and change confidential_guest = false to use:

        "emptyDir": {
            "mount_type": "local",
            "mount_source": "^$(cpath)/$(sandbox-id)/rootfs/local/",
            "mount_point": "^$(cpath)/$(sandbox-id)/local/",
            "driver": "local",
            "source": "local",
            "fstype": "local",
            "options": [
                "mode=0777"
            ]
        },

The value of the mount_source field is different.

This change unblocks testing using Kata CI's pod-empty-dir.yaml:

genpolicy -u -y pod-empty-dir.yaml

kubectl apply -f pod-empty-dir.yaml

k get pod sharevol-kata
NAME            READY   STATUS    RESTARTS   AGE
sharevol-kata   1/1     Running   0          53s

Fixes: #8887

### DIFF
--- a/src/tools/genpolicy/genpolicy-settings.json
+++ b/src/tools/genpolicy/genpolicy-settings.json
@@ -137,6 +137,17 @@
     "volumes": {
         "emptyDir": {
             "mount_type": "local",
+            "mount_source": "^$(cpath)/$(sandbox-id)/rootfs/local/",
+            "mount_point": "^$(cpath)/$(sandbox-id)/local/",
+            "driver": "local",
+            "source": "local",
+            "fstype": "local",
+            "options": [
+                "mode=0777"
+            ]
+        },
+        "confidential_emptyDir": {
+            "mount_type": "local",
             "mount_source": "^$(cpath)/$(sandbox-id)/local/",
             "mount_point": "^$(cpath)/$(sandbox-id)/local/",
             "driver": "local",
--- a/src/tools/genpolicy/src/mount_and_storage.rs
+++ b/src/tools/genpolicy/src/mount_and_storage.rs
@@ -99,12 +99,24 @@ pub fn get_mount_and_storage(
     yaml_mount: &pod::VolumeMount,
 ) {
     if let Some(emptyDir) = &yaml_volume.emptyDir {
-        let memory_medium = if let Some(medium) = &emptyDir.medium {
-            medium == "Memory"
-        } else {
-            false
-        };
-        get_empty_dir_mount_and_storage(settings, p_mounts, storages, yaml_mount, memory_medium);
+        let settings_volumes = &settings.volumes;
+        let mut volume: Option<&settings::EmptyDirVolume> = None;
+
+        if let Some(medium) = &emptyDir.medium {
+            if medium == "Memory" {
+                volume = Some(&settings_volumes.emptyDir_memory);
+            }
+        }
+
+        if volume.is_none() {
+            volume = if settings.kata_config.confidential_guest {
+                Some(&settings_volumes.confidential_emptyDir)
+            } else {
+                Some(&settings_volumes.emptyDir)
+            }
+        }
+
+        get_empty_dir_mount_and_storage(settings, p_mounts, storages, yaml_mount, volume.unwrap());
     } else if yaml_volume.persistentVolumeClaim.is_some() || yaml_volume.azureFile.is_some() {
         get_shared_bind_mount(yaml_mount, p_mounts, "rprivate", "rw");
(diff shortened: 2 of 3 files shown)


wainersm:
Hi @danmihai1 !

I didn't follow the initial CI implementation for testing the agent policy so I'm trying to understand it now.

If I got it right, the [yamls for the k8s tests](https://github.com/kata-containers/kata-containers/tree/main/tests/integration/kubernetes/runtimeclass_workloads) are transformed (i.e. added annotations) for certain `$KATA_HYPERVISOR` (i.e. qemu-sev, qemu-tdx, qemu-snp) or `$KATA_HOST_OS` is `cb-lmariner` (see [code](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/run_kubernetes_tests.sh#L86)).

The following jobs met those conditions:
 * https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-kata-coco-tests.yaml
 * https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-k8s-tests-on-aks.yaml#L45

The aforementioned "yamls transformation" don't use the `genpolicy` at the moment. Instead the [default kata's repo policies](https://github.com/kata-containers/kata-containers/tree/main/src/kata-opa) are applied "manually". So here is what you doing: improving genpolicy and configuration files so that it could be used instead?

wainersm:
> > The aforementioned "yamls transformation" don't use the `genpolicy` at the moment. Instead the [default kata's repo policies](https://github.com/kata-containers/kata-containers/tree/main/src/kata-opa) are applied "manually". So here is what you doing: improving genpolicy and configuration files so that it could be used instead?
> 
> Yes:
> 
>     1. The older way of testing Policy ( https://github.com/kata-containers/kata-containers/blob/f1ca5d1563be3e704acd10ccf2c44a975f26bf8b/tests/integration/kubernetes/run_kubernetes_tests.sh#L133
>         ) uses a very basic, hand-crafted policy - insufficient for real world use cases.
> 
>     2. A proper CoCo Policy would be significantly more complex - therefore typical customers will need to use an automatically generated Policy, such as a Policy generated by the _genpolicy_ tool. So, the next goal is to test using genpolicy, as shown for one test in [tests: k8s-attach-handlers auto-generated policy #8922](https://github.com/kata-containers/kata-containers/pull/8922)

Perfect. Got it!

wainersm:
Ok, I think I understood how the `confidential_guest` setting means. If set to `true` then `genpolicy` will pick up the confidential rules `confidential_emptyDir`, `confidential_configMap`, etc...

I was thinking on how to remove the conditionals `if confidential then do this` from the code (if that´s desirable):
* having two settings files (e.g. `genpolicy-settings.json` and `coco-genpolicy-settings.json`), user specify which he/she wants to use. Problem: changes on common rules should be applied on both files (i.e. maintenance gets worse)
*  still haveing two settings but the specialized one (i.e. `coco-genpolicy-settings.json`) points to the `base_settings_file`, overwriting existing rules from the base and/or adding new rules. User still need to switch the settings file with the `-j` option.

wainersm (approved):
Based on my limited knowledge of genpolicy: LGTM!

danmihai1:
> Ok, I think I understood how the `confidential_guest` setting means. If set to `true` then `genpolicy` will pick up the confidential rules `confidential_emptyDir`, `confidential_configMap`, etc...
> 
> I was thinking on how to remove the conditionals `if confidential then do this` from the code (if that´s desirable):
> 
> * having two settings files (e.g. `genpolicy-settings.json` and `coco-genpolicy-settings.json`), user specify which he/she wants to use. Problem: changes on common rules should be applied on both files (i.e. maintenance gets worse)
> * still haveing two settings but the specialized one (i.e. `coco-genpolicy-settings.json`) points to the `base_settings_file`, overwriting existing rules from the base and/or adding new rules. User still need to switch the settings file with the `-j` option.

That is correct. Sorry these settings are confusing - I agree we need to clean them up.

GabyCT (approved):

Issue #9036, opened 2024-02-06
runtime: it does not allow to create SEV confidential container on SNP enabled host.


niteeshkd:
When I try to create SEV confidential container on SNP enabled host using the runtimeclass  `kata-qemu-sev` it creates SNP container instead of SEV container. 

The qemu process for the container shows the following qemu options.
```
-machine q35,accel=kvm,kernel_irqchip=split,confidential-guest-support=snp -cpu EPYC-v4,pmu=off 
-object sev-snp-guest,id=snp,cbitpos=51,reduced-phys-bits=1

```
The output of 'dmesg | grep -i -e sev -e snp' inside the container shows the following.
```
[    0.060867] Memory Encryption Features active: AMD SEV SEV-ES SEV-SNP
[    0.174847] SEV: APIC: wakeup_secondary_cpu() replaced with wakeup_cpu_via_vmgexit()
[    0.181378] SEV: Using SNP CPUID table, 31 entries present.
[    0.300942] SEV: SNP guest platform device initialized.
[    1.079915] sev-guest sev-guest: Initialized SEV guest driver (using vmpck_id 0)
```

PR #9037, opened 2024-02-06
runtime: fix creation of SEV confidential container on SNP enabled host.

niteeshkd:
Fixes the issue #9036 .

### DIFF
--- a/src/runtime/virtcontainers/qemu_amd64.go
+++ b/src/runtime/virtcontainers/qemu_amd64.go
@@ -124,7 +124,7 @@ func newQemuArch(config HypervisorConfig) (qemuArch, error) {
 			legacySerial:         config.LegacySerial,
 		},
 		vmFactory: factory,
-		snpGuest:  config.ConfidentialGuest,
+		snpGuest:  config.SevSnpGuest,
 	}
 
 	if config.ConfidentialGuest {


niteeshkd:
@ryansavino @fitzthum @zvonkok @jpecholt

ryansavino (approved):
Ah, thanks Niteesh for fixing this.

fitzthum (approved):
Great catch. We're changing a lot of stuff to `ConfidentialGuest`, but the `snpGuest` flag should be `SevSnpGuest`, because `ConfidentialGuest` is true for both SEV and SNP.

fidencio (changes requested):
@niteeshkd, would you mind to add a commiit message explaining why this is needed?
If this is a regression, mind to mention when we regressed?

This kind of information is crucial for maintainers to navigate through the history of the projects (and bugs, and fixes).

fidencio:
Answering some of my own questions here, it seems that the regression was introduced by de39fb7d3897bbb318c98ddd97805e54fa2b2625.  Let's also loop in @zvonkok to the loop to make sure it doesn't break his GPU related work, and if it does break, figure out the best way to solve.

fidencio (approved):
lgtm, thanks @niteeshkd!

niteeshkd:
Created #9040 to add a test in future.

ryansavino:
All required checks are passing on this one. @stevenhorsman @danmihai1 @fidencio We good to merge this?

danmihai1 (approved):

stevenhorsman (approved):
LGTM

Issue #9044, opened 2024-02-07
nydus: Bump nydus snapshotter version to v0.13.7


ChengyuZhu6:
The latest release of nydus snapshotter is v0.13.7, which is required for confidential containers to enable guest pull feature. let's switch to this release.

PR #9045, opened 2024-02-07
nydus: Bump nydus snapshotter version to v0.13.7

ChengyuZhu6:
Bump nydus snapshotter version to v0.13.7.

The new release name of nydus snapshotter is `nydus-snapshotter-v0.13.7-linux-amd64.tar.gz`, which differs from the version used by kata (`nydus-snapshotter-v0.12.0-x86_64.tgz`). Therefore, we need to update the script to obtain the correct nydus snapshotter name.

Fixes: #9044

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -551,7 +551,7 @@ function install_nydus_snapshotter() {
 	version="${1}"
 
 	project="containerd/nydus-snapshotter"
-	tarball_name="nydus-snapshotter-${version}-$(${repo_root_dir}/tests/kata-arch.sh).tgz"
+	tarball_name="nydus-snapshotter-${version}-$(uname -s| tr A-Z a-z)-$(${repo_root_dir}/tests/kata-arch.sh -g).tar.gz"
 
 	download_github_project_tarball "${project}" "${version}" "${tarball_name}"
 	sudo tar xfz "${tarball_name}" -C /usr/local/bin --strip-components=1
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @ChengyuZhu6!

tests/common.bash
```diff
@@ -551,7 +551,7 @@ function install_nydus_snapshotter() {
 	version="${1}"
 
 	project="containerd/nydus-snapshotter"
-	tarball_name="nydus-snapshotter-${version}-$(${repo_root_dir}/tests/kata-arch.sh).tgz"
+	tarball_name="nydus-snapshotter-${version}-$(uname -s| tr A-Z a-z)-$(${repo_root_dir}/tests/kata-arch.sh).tar.gz"
```
stevenhorsman, line 554: Do we need to use something like `kata-arch.sh -g` now we want `amd64 rather than `x86_64` in the tarball name?

stevenhorsman, line 554: Do we need to use something like kata-arch.sh -g now we want amd64 rather than x86_64` in the tarball name?
```suggestion
	tarball_name="nydus-snapshotter-${version}-$(uname -s| tr A-Z a-z)-$(${repo_root_dir}/tests/kata-arch.sh -g).tar.gz"
```

wainersm: I second @stevenhorsman 's suggestion.

wainersm (approved):
With @stevenhorsman 's suggestion

tests/common.bash
ChengyuZhu6: Thanks @stevenhorsman. I have fixed it.

ChengyuZhu6:
> With @stevenhorsman 's suggestion

Thanks @wainersm. I have fixed it.

GabyCT (approved):

beraldoleal (approved):
LGTM, thanks.

Issue #8305, opened 2023-10-24
utils: kata-manager: Add option to switch hypervisor (config file)


jodh-intel:
Update the [`kata-manager.sh`](https://github.com/kata-containers/kata-containers/blob/main/utils/kata-manager.sh) tool to allow the user to switch config file.

Related: #8250, #8281.

jodh-intel:
I've started this but I think it makes sense to block on #8306 as it simplifies the task :smile:

jodh-intel:
Rescoping this issue to only switch hypervisor / config file. See #8677 for switching the runtime type.

> **Note:** The logic in the script can be simplified when #8641 lands.

Issue #8677, opened 2023-12-14
utils: kata-manager: Add option to switch runtime type


jodh-intel:
Update the [kata-manager.sh](https://github.com/kata-containers/kata-containers/blob/main/utils/kata-manager.sh) tool to allow the user to switch runtime type (golang or rust).

Related: #8250, #8281, #8305.

jodh-intel:
See https://github.com/kata-containers/kata-containers/issues/9060#issuecomment-1938694127 for thoughts on handling the set of config files for each type of runtime.

PR #9059, opened 2024-02-08
kata-manager: Allow hypervisor to be changed

jodh-intel:
kata-manager: Allow hypervisor to be changed

Add new options to allow the configured hypervisor to be changed:

- `-L`: List available _packaged_ hypervisor config names (filename fragments).
- `-e`: List available _local_ hypervisor config names (filename fragments).
- `-H <hypervisor>`: Install Kata then switch to the specified hypervisor (by filename fragment).
- `-S <hypervisor>`: Switch to the specified hypervisor (by filename fragment [Errors if Kata not installed]).

For example, to install Kata and configure it to use Cloud Hypervisor
with the golang Kata runtime:

```bash
$ kata-manager.sh -H clh
```

To switch back to the default hypervisor:

```bash
$ kata-manager.sh -S default
```

To show details of the available packaged configs:

```bash
$ kata-manager.sh -L
```

To show details of the local configs:

```bash
$ kata-manager.sh -e
```

> **Notes:**
>
> - This change **only** applies to the current default (golang) Kata runtime.
>
> - Although this is mainly for users wishing to switch hypervisor (by
>   changing the Kata config file to another of the packaged config files
>   provided for specific hypervisors), strictly it allows users to change
>   to _any_ config file. For example, if the user has a config file called
>   `/etc/kata-containers/configuration-my-custom-config.toml`, they could
>   switch to this by running:
>
>   ```bash
>   $ kata-manager.sh -S my-custom-config
>   ```

Fixes: #8305.

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -2,15 +2,18 @@
 
 # Kata Manager
 
-> **Note:**
+> **Notes:**
 >
-> We recommend users install Kata Containers using
-> [official distribution packages](../docs/install/README.md#official-packages), where available.
+> - We recommend users install Kata Containers using
+>   [official distribution packages](../docs/install/README.md#official-packages), where available.
+>
+> - These instructions only apply to the current default (golang) Kata runtime.
+>   See https://github.com/kata-containers/kata-containers/issues/9060 for further details.
 
 The [`kata-manager.sh`](kata-manager.sh) script automatically installs and
-configures Kata Containers and containerd.
+configures Kata Containers and a container manager (such as containerd, Docker and `nerdctl`).
 
-This scripted method installs the latest versions of Kata Containers and
+By default, the script installs the latest versions of Kata Containers and
 containerd. However, be aware of the following before proceeding:
 
 - Packages will **not** be automatically updated
@@ -35,10 +38,22 @@ containerd, and then configure containerd to use Kata Containers. However,
 the script provides a number of options to allow you to change its
 behaviour.
 
-> **Note:**
+> **Notes:**
 >
-> Before running the script to install Kata Containers, we recommend
-> that you [review the available options](#show-available-options).
+> - Before running the script to install Kata Containers, we recommend
+>   that you [review the available options](#show-available-options).
+>
+> - The `kata-manager.sh` script is
+>   [now packaged](https://github.com/kata-containers/kata-containers/pull/9091)
+>   as part of the Kata Containers release and can be called as either
+>   `kata-manager.sh` or simply `kata-manager`. Some of the sections
+>   below give two examples of how to run the script: running the
+>   local version, and by downloading and running the latest version
+>   in the Kata Containers GitHub repository. If your version of Kata
+>   Containers includes the `kata-manager.sh` script, you can run
+>   either version although we would suggest you use the local version
+>   since: (a) it has been tested with the Kata Containers release you
+>   are using; and (b) it is simpler to run the command directly.
 
 ### Show available options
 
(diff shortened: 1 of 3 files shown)


jodh-intel:
Note that running with `-L` won't work unless you are using a very new installation of Kata (newer than the official `3.3.0-alpha0` release: https://github.com/kata-containers/kata-containers/releases. The reason being that #8483 isn't yet in any official releases.

/cc @fidencio, @amshinde, @cmaf, @dborquez.

jodh-intel:
Two thoughts:

- I don't particularly like the option "names": `-L` and `-e` are related but you'd never know from the option letters. Maybe we should switch to using long options with `getopt(1)` for greater clarity?
- We could drop one of `-H` or `-S` and just have a single option that will dtrt. Thoughts?

jodh-intel:
I've updated the branch today to add in `-e` (show local config details). But the change also removes the hard-coded default hypervisor from the script which now derives it from the packaged config file directory.

jodh-intel:
As demonstrated by the doc changes on this PR, using the `kata-manager` script from GitHub is unwieldy. I've raised #9066 so we can consider adding it to the release.

cmaf (approved):
@jodh-intel this LGTM although I have not tested it locally

jodh-intel:
Thanks, @cmaf!

> @jodh-intel this LGTM although I have not tested it locally

What you can do is run the code using `curl`. For example, to see the help statement:

```bash
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/3dcd5341637fd60d7b9f86602acb0312dcf005d0/utils/kata-manager.sh) -h"
```

> **Note:** The commit in the `curl` URL is the latest commit currently on this PR (3dcd5341637fd60d7b9f86602acb0312dcf005d0).

See https://github.com/kata-containers/kata-containers/blob/main/utils/README.md for other options not added on this PR.

utils/README.md
```diff
@@ -63,3 +66,87 @@ To install and configure a system with Kata Containers and containerd, run:
 ```bash
 $ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh)"
 ```
+
+### Choose a Hypervisor
+
+Kata works with different [hypervisors](../docs/hypervisors.md). When you install a Kata system, the default hypervisor
```
fidencio, line 72: I understand the motivation on linking to the hypervisors page, but the user will not know which name to use to select, for instance, "Cloud Hypervisor".

utils/README.md
```diff
@@ -63,3 +66,87 @@ To install and configure a system with Kata Containers and containerd, run:
...
+##### To install Kata Containers and containerd and configure it for a specific hypervisor
+
+To specify that Kata be installed and configured to use a specific
+hypervisor, use the `-H` option. For example, to select Cloud Hypervisor, run:
+
+```bash
+$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh) -H clh"
+```
+
+> **Note:** See the [List available hypervisors](#list-available-hypervisors) section
```
fidencio, line 129: This note is not that much useful as we don't have a way to actually list the hypervisors before we install Kata Containers. :-/

utils/README.md
```diff
@@ -63,3 +66,87 @@ To install and configure a system with Kata Containers and containerd, run:
...
+installed system, run:
+
+```bash
+$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/kata-containers/main/utils/kata-manager.sh) -e"
+```
+
+> **Note:** This command can only be run once Kata has been installed.
+
+> See the [configuration documentation](https://github.com/kata-containers/kata-containers#configuration)
+> for further information.
```
fidencio, line 108: I wonder if those 3 options could be combined into a single one which would output something like:
```
HYPERVISOR | ARGUMENT NAME | DEFAULT
Cloud Hypervisor | clh | yes
QEMU | qemu | no
,,,,
```
Or something towards that line ...
NOTE: After the next release e should remove all the mentions to `$ bash -c ...` and just point to the script directly.

fidencio (commented):
@jodh-intel, I have a few suggestions, none of those are code related so far.
Please, let me know your thoights.

utils/README.md
fidencio: We need a way to have the list of supported hypervisors being displayed without installing kata.

fidencio:
A few more notes here:
```
ubuntu@kata-manager:~/kata-containers/utils$ ./kata-manager.sh -L
acrn;;packaged;golang
clh;;packaged;golang
dragonball;;packaged;golang
fc;;packaged;golang
qemu-nvidia-gpu;;packaged;golang
qemu-sev;;packaged;golang
qemu-snp;;packaged;golang
qemu-tdx;;packaged;golang
qemu;default;packaged;golang
```

This output is not user friendly, and dragonball is not golang. :-)
It's hard for people to know what each column means.

fidencio:
I noticed you'be been changing the default configuration on `/etc/kata-containers`, wouldn't be easier to just set the default in `/opt/kata/...`?  That's what we already do with kata-deploy.

utils/README.md
jodh-intel: It's not explained in the commits, but I tried to minimise what we hard-code in the script as it's better to rely on the structure of the packaged config files to determine the logic. Further, it will simplify our lives when we tackle #8677 and #9060.

 I've now updated the branch so that the `hypervisors.md` page shows the short names, which allows the users to determine the names used by the script.

utils/README.md
jodh-intel: > I wonder if those 3 options could be combined into a single one

I'm not totally clear what you mean by this comment but I have updated the hypervisors doc to include more hypervisor config details including the "short name" used by `kata-manager`.

> NOTE: After the next release e should remove all the mentions to $ bash -c ... and just point to the script directly.

Yep - see #9091.

utils/README.md
jodh-intel: The doc now lists the "short names".

However, I am wondering about removing the `-H` option since, effectively, that option just installs Kata, then calls `-S`. There isn't much benefit to providing `-H` since we still need to install Kata. Then the `-S` just does the config switch. wdyt to removing `-H`?

jodh-intel:
> I noticed you'be been changing the default configuration on /etc/kata-containers, wouldn't be easier to just set the default in /opt/kata/...? That's what we already do with kata-deploy.

It is admittedly more "fiddly" to deal with, but `kata-manager` isn't just for `kata-deploy` environments, so it's safer and cleaner to modify the local config files in `/etc/kata-containers/` as that also works for stateless environments, read-only disk environments, _etc_. So, once `kata-manager` installs Kata, it never touches the pristine files below `/opt/kata/`. Plus, I don't think we want to drop the concept that if you futz up your local config somehow, you can either move / rename your local config files, or even `sudo rm -rf /etc/kata-containers/`, and the system will reset to "factory defaults" :smile:

jodh-intel:
> A few more notes here:
> ```
> ubuntu@kata-manager:~/kata-containers/utils$ ./kata-manager.sh -L
> acrn;;packaged;golang
> clh;;packaged;golang
> dragonball;;packaged;golang
> fc;;packaged;golang
> qemu-nvidia-gpu;;packaged;golang
> qemu-sev;;packaged;golang
> qemu-snp;;packaged;golang
> qemu-tdx;;packaged;golang
> qemu;default;packaged;golang
> ```
>
> This output is not user friendly

It's designed to be parseable, but we could add in a tab separated format too.

> , and dragonball is not golang. :-)

Correct. See my note at https://github.com/kata-containers/kata-containers/pull/9059#issuecomment-1934490946. That issue is caused by the fact that the Kata project hasn't produced a new release in four months. Related: #9064.

> It's hard for people to know what each column means.

I can update the usage.

>
> Also, what's the value of having `packaged` there?

I could drop that. It's there to show these are packaged config short names. We're not showing the full paths (since they may differ, are quite long and would make the output even worse unless you're happy with line wrapping or happen to have a wide terminal), so need an indicator as to whether these are the pristine packaged config files, or if they are the local config file copies.

utils/kata-manager.sh
```diff
@@ -737,6 +737,7 @@ configure_kata()
 	sudo install -o root -g root -m 0644 "$cfg_from" "$cfg_to"
 
 	sudo sed -i \
+		--follow-symlinks \
```
amshinde, line 813: Great catch @jodh-intel !
I wasnt aware myself about this behaviour myself. Reading about this, it shows that GNU sed has --follow-symlinks, while  BSD sed does not. It will be useful to atleast document this here, in case someone tries  to use this/port this to MacOS. We afterall do support macos to some extent.

We could mention something like "If you're on macOS or another BSD variant, you'll need to install GNU sed via your package manager (brew, pkg, etc)"

utils/kata-manager.sh
```diff
@@ -265,31 +293,54 @@ Description: Install $kata_project [1] (and optionally $containerd_project [2])
...
 Notes:
 
 - The version strings must refer to official GitHub releases for each project.
   If not specified or set to "", install the latest available version.
 
+- The '-L' option requires an installed system.
+
+  > **Note:** For details of each hypervisor, see [3].
+
+- If an invalid hypervisor name is specified with the '-H' option, an
```
amshinde, line 325: Personally I do not like the concept of designating any partcular hypervisor as default. I feel all the hypervisors should be treated the same with the user choosing one based on his/her needs.
(While I do understand that configuration.toml points to qemu conf by default, we should limit the usage of default and not designate any one hypervisor as default)

If there is a failure on configuration here, could we just go back to previous configuration instead of default?

utils/kata-manager.sh
```diff
@@ -23,6 +23,8 @@ readonly containerd_slug="containerd/containerd"
 readonly kata_project_url="https://github.com/${kata_slug}"
```
amshinde, line 23: Adding my comment here, but this is for the commiot message. Just reading the commit message, the usage of filename fragments is not really clear to me. Could we reword it?

utils/kata-manager.sh
jodh-intel: Hi, @amshinde - I don't disagree with what you've said, but we need some way to designate "the currently configured hypervisor". How about just `configured`?

amshinde: @jodh-intel Yes that wording works for me.

jodh-intel: @amshinde - Switching everything to `configured` unfortunately results in a less clear approach. For example, how does the user "undo all changes" and revert to the pristine default configured hypervisor?

The current approach is:

```bash
$ kata-manager.sh -S default
```

But that just looks odd if we change the name to `configured` as it just doesn't express the same intent:

```bash
$ kata-manager.sh -S configured
```

That's confusing and ambiguous imho as it doesn't suggest the value specified to the `-S` option means "revert to the original configured hypervisor".

The best I can come up with as a compromise is:


```bash
$ kata-manager.sh -S pristine
```

wdyt?

/cc @fidencio

amshinde (approved):

utils/kata-manager.sh
jodh-intel: Fixed.

jodh-intel:
I've also changed the format so that it now uses tabs rather than semi-colons so that it's still parseable but looks a bit less "busy" :smile:

jodh-intel:
The ARM CI seem to be failing:

```
4:21:11 Reading package lists...
14:21:11 E: The repository 'http://packages.cloud.google.com/apt kubernetes-xenial-unstable Release' no longer has a Release file.
14:21:11 W: Target Packages (stable/binary-arm64/Packages) is configured multiple times in /etc/apt/sources.list:5 and /etc/apt/sources.list.d/docker.list:1
14:21:11 W: Target Packages (stable/binary-all/Packages) is configured multiple times in /etc/apt/sources.list:5 and /etc/apt/sources.list.d/docker.list:1
14:21:11 W: Target Packages (stable/binary-arm64/Packages) is configured multiple times in /etc/apt/sources.list:5 and /etc/apt/sources.list.d/docker.list:1
14:21:11 W: Target Packages (stable/binary-all/Packages) is configured multiple times in /etc/apt/sources.list:5 and /etc/apt/sources.list.d/docker.list:1
14:21:11 [setup_env_ubuntu.sh:15] ERROR: sudo -E apt update
14:21:11 [jenkins_job_build.sh:248] ERROR: ci/setup.sh
14:21:13 Build step 'Execute shell' marked build as failure
```

/cc @jongwu.

PR #9062, opened 2024-02-08
WIP tests: Add ability to run non-TEE environments

### LINKED ISSUE #9055: Add the ability to run (some of the) confidential tests in non-TEE environments
Currently, the only confidential test we have is designed around TEE environments and it won't work for non-TEEs, as expected.
However, in the future, we'll want to run some of the tests as part of the non-TEEs and it requires some changes in the logic so the tests can be expanded to support this case as well.

The changes that I have in mind to adapt what we have are:
* Create a list of SUPPORTED_NON_TEE_HYPERVISORS, similar to https://github.com/kata-containers/kata-containers/blob/6ead48ec062fcfa9bcd95d12a20a266e07f5e95a/tests/integration/kubernetes/k8s-confidential.bats#L13
* Ensure that QEMU will be part of that list, and do not bail during the setup
* If QEMU is being use (so, the a non-tee case) then we need to:
  * Skip the tests that do not make sense to run on non-TEE environments
  * Ensure that we can set the image as an annotation
    * This should be done in a very similar way as it'd been done for mariner, see: https://github.com/kata-containers/kata-containers/blob/b99f5745222536936ccf8a0fb23653d4f71fc0de/tests/integration/kubernetes/gha-run.sh#L129-L133
      * It can be done in the very same file, in the very same way
  * Add the annotations to the yaml files we'll use
    *  We need to image to be used to be the confidential one, as it has the guest components inside it
    * This will be done using the resulting work of https://github.com/kata-containers/kata-containers/issues/9054

GabyCT:
This PR adds the ability to run k8s confidential tests in a non-TEE environment.

Fixes #9055


tests/integration/kubernetes/gha-run.sh
```diff
@@ -126,7 +126,7 @@ function deploy_kata() {
 	# Enable 'default_vcpus' hypervisor annotation
 	yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[6].value' "default_vcpus"
 
-	if [ "${KATA_HOST_OS}" = "cbl-mariner" ]; then
+	if [ "${KATA_HOST_OS}" = "cbl-mariner" ] || [ "${KATA_HOST_OS}" = "ubuntu" ]; then
```
fidencio, line 129: @GabyCT, I don't think we want to do it here.

What I think we should do is creating a new check, after this one, like:
```sh
if [ "${KATA_HYPERVISOR}" = "qemu" ]; then
    yq write -i "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" 'spec.template.spec.containers[0].env[6].value' "image initrd kernel default_vcpus"
fi
```

tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -10,17 +10,22 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
...
 
 setup() {
-	SUPPORTED_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx" "qemu-se")
+	SUPPORTED_TEE_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx" "qemu-se")
+	SUPPORTED_NON_TEE_HYPERVISORS=("qemu")
 
 	# This check must be done with "<SPACE>${KATA_HYPERVISOR}<SPACE>" to avoid
 	# having substrings, like qemu, being matched with qemu-$something.
-	[[ " ${SUPPORTED_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]] ||  skip "Test not supported for ${KATA_HYPERVISOR}."
+	if [[ " ${SUPPORTED_TEE_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]] || [[ " ${SUPPORTED_NON_TEE_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]]; then
```
fidencio, line 18: I think this check is wrong.  Reading it it says:
* If kata-hypervisor is in the supported list of TEEs, or if kata-hypervisor is in the supported list of non-TEEs
  * then skip the test

What we would actually want to check is:
* If kata-hypervisor is **NOT** in the supported list of tees **AND** kata-hypervisor is **NOT** in the supported list of non-tees
  *  then skip the test.

tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -10,17 +10,22 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
...
+		skip "Test not supported for ${KATA_HYPERVISOR}."
+	fi
 
 	get_pod_config_dir
```
fidencio, line 22: I'd add an if-else here to already call the method to set-up annotations, as that just got merged this morning

tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -46,7 +51,9 @@ setup() {
 }
 
 teardown() {
-	[[ " ${SUPPORTED_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]] ||  skip "Test not supported for ${KATA_HYPERVISOR}."
+	if [[ " ${SUPPORTED_TEE_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]] || [[ " ${SUPPORTED_NON_TEE_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]]; then
```
fidencio, line 54: This check also seems incorrect.

fidencio (commented):
@GabyCT, I've left a few comments.

tests/integration/kubernetes/k8s-confidential.bats
GabyCT: @fidencio thanks for the feedback so I just have a question, for the supported_tee_hypervisors I need to call https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/setup.sh#L100 but in the case of qemu that does not apply

fidencio: Actually the opposite, @GabyCT.

For the **TEEs** we don't need to set the annotation, because those will be using the correct image (the confidential image)
For the **non-TEEs** we need to set the annotation, because for those we want to use the confidential image instead of the normal one).

We can even split that function out of the setup.sh and put it somewhere easier to be accessed.

Does the explanation help?

tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -10,12 +10,17 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
...
 
 setup() {
-	SUPPORTED_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx" "qemu-se")
+	SUPPORTED_TEE_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx" "qemu-se")
+	SUPPORTED_NON_TEE_HYPERVISORS=("qemu")
 
 	# This check must be done with "<SPACE>${KATA_HYPERVISOR}<SPACE>" to avoid
 	# having substrings, like qemu, being matched with qemu-$something.
-	[[ " ${SUPPORTED_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]] ||  skip "Test not supported for ${KATA_HYPERVISOR}."
+	if ! [[ "${SUPPORTED_TEE_HYPERVISORS[@]}" =~ "${KATA_HYPERVISOR}" ]] && [[ "${SUPPORTED_NON_TEE_HYPERVISORS}" =~ "${KATA_HYPERVISOR}" ]]; then
```
fidencio, line 18:
```suggestion
	if ! [[ "${SUPPORTED_TEE_HYPERVISORS[@]}" =~ "${KATA_HYPERVISOR}" ]] && ! [[ "${SUPPORTED_NON_TEE_HYPERVISORS}" =~ "${KATA_HYPERVISOR}" ]]; then
```

Otherwise the logic is not correct.

tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -46,7 +51,9 @@ setup() {
 }
 
 teardown() {
-	[[ " ${SUPPORTED_HYPERVISORS[*]} " =~  " ${KATA_HYPERVISOR} " ]] ||  skip "Test not supported for ${KATA_HYPERVISOR}."
+	if ! [[ "${SUPPORTED_TEE_HYPERVISORS[@]}" =~ "${KATA_HYPERVISOR}" ]] && [[ "${SUPPORTED_NON_TEE_HYPERVISORS}" =~ "${KATA_HYPERVISOR}" ]]; then
```
fidencio, line 54:
```suggestion
	if ! [[ "${SUPPORTED_TEE_HYPERVISORS[@]}" =~ "${KATA_HYPERVISOR}" ]] && ! [[ "${SUPPORTED_NON_TEE_HYPERVISORS}" =~ "${KATA_HYPERVISOR}" ]]; then
```

Otherwise the logic is not correct.

fidencio (commented):
A few more comments, @GabyCT!

dborquez (approved):

GabyCT:
@fidencio seems like we are having some issues 
```
not ok 1 Test unencrypted confidential container launch success and verify that we are running in a secure enclave.
# (from function `die' in file ../../common.bash, line 32,
#  in test file k8s-confidential.bats, line 51)
#   `[ -z "$coco_enabled" ] && die "Confidential compute is expected but not enabled."' failed
```

GabyCT:
that is related with https://github.com/kata-containers/kata-containers/blob/00be9ae8725f25a9b32e48a9d11fa901fc997b43/tests/integration/kubernetes/confidential_common.sh#L10
Do we need this for non-tee?

Issue #9069, opened 2024-02-09
tests: k8s: exec_host() deletes pods that were not created by the test


danmihai1:
Sometimes it's helpful for me to run the Kata CI tests while having a debugger pod running on my cluster. But running k8s-file-volume.bats and/or similar tests kills my already-existing debugger pods. :-(

PR #9070, opened 2024-02-09
tests: k8s: avoid deleting unrelated pods

danmihai1:
Delete the debugger pod created during the test, rather than already existing debugger pods.

Fixes: #9069


wainersm (approved):
LGTM. Thanks @danmihai1 !

Issue #9032, opened 2024-02-06
CC: Integrate pause image into rootfs


ChengyuZhu6:
**Which feature do you think can be improved?**

For Confidential containers stack, the pause image is managed by host side, then it may configure a malicious pause image, we need package a pause image inside the rootfs and don't the pause image from host. So the pause image will be part of the Kata Containers rootfs that's used by the Confidential Containers usecase.

danmihai1:
I hope there is an option to keep some of these components out of the Guest rootfs. 

Updating the Guest rootfs is expensive in a Production environment so ideally the pause container, the Attestation client, etc. should be pulled from a container image repository rather than cramming all of them into the Guest rootfs.

PR #9080, opened 2024-02-13
release: Don't ship the pause-image / coco-guest-components as part of the release artefacts

fidencio:
It doesn't make sense to ship the pause-image / coco-guest-components themselves as an release artefact.

The reason we build and cache those is in order to use them inside the rootfs, and that's it, there's not need to ship them as part of the release, at all.

Fixes: #9032 -- part II


BbolroC (approved):
LGTM, thanks @fidencio !

zvonkok (approved):
LGTM, thanks!

Issue #9081, opened 2024-02-13
ci: tdx: clean up kata-deploy before running the tests


fidencio:
As we've seen some leftovers and those are making the tests fail more than often.

PR #9082, opened 2024-02-13
tests: Remove kata-deploy-tdx test and ensure kata-deploy is always cleaned up before starting the tests

fidencio:
---

ci: Clean up kata-deploy ds before starting the tests

This will ensure no leftovers are in the node, which has been cause the
TDX CI to fail every now and then.

Fixes: #9081

---

gha: tdx: Stop running kata-deploy tests on TDX

We only have one TDX machine, let's not make it busier than needed.

---

### DIFF
--- a/.github/workflows/run-kata-coco-tests.yaml
+++ b/.github/workflows/run-kata-coco-tests.yaml
@@ -23,42 +23,6 @@ on:
         default: ""
 
 jobs:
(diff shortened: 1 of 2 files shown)


GabyCT (approved):

zvonkok (approved):
LGTM, thanks

PR #9086, opened 2024-02-13
packaging: qemu-snp-experimental: support host kernel with gmem

niteeshkd:
This is required to allow creation of SNP coco on host kernel (e.g. https://github.com/AMDESE/linux ,branch:snp-host-latest) supporting guest private memory for SNP using gmem.

### DIFF
--- a/tools/packaging/static-build/qemu/Dockerfile
+++ b/tools/packaging/static-build/qemu/Dockerfile
@@ -60,6 +60,7 @@ RUN apt-get update && apt-get upgrade -y && \
 	    patch \
 	    python \
 	    python-dev \
+	    python3-venv \
 	    rsync \
 	    zlib1g-dev${DPKG_ARCH} && \
     if [ "${ARCH}" != s390x ]; then apt-get install -y --no-install-recommends libpmem-dev${DPKG_ARCH}; fi && \
--- a/tools/packaging/scripts/configure-hypervisor.sh
+++ b/tools/packaging/scripts/configure-hypervisor.sh
@@ -328,7 +328,11 @@ generate_qemu_options() {
 	# From Kata Containers 2.5.0-alpha2 all arches but powerpc have been
 	# using the new implementation of virtiofs daemon, which is not part
 	# of QEMU.
-	qemu_options+=(functionality:--disable-virtiofsd)
+	#
+	# qemu configure does not support virtiofsd if qemu version >= 8.0.0.
+        if ! gt_eq "${qemu_version}" "8.0.0" ; then
+		qemu_options+=(functionality:--disable-virtiofsd)
+        fi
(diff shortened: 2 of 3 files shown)


niteeshkd:
@ryansavino @fitzthum @fidencio

fidencio:
@niteeshkd, does it require any change in the QEMU command line used to create the VM?

niteeshkd:
> does it require any change in the QEMU command line used to create the VM?

It does not require change in the kata shim to create the SNP coco. If we need to handle the attestation, we will have to modify the shim which we will handle later.

fidencio:
I'll leave the review for those more familiar with those changes, and if an ack is needed from my side, I'll add after they do the review.
Thanks @niteeshkd!

fitzthum (approved):
Looks good. We'll want to make sure to run to upgrade the CI nodes.

Btw @niteeshkd will this new QEMU still function on the old non-UPM SNP Kernel? If not we should probably also add a note to the coco docs.

Also @danmihai1 might be interested.

niteeshkd:
> Btw @niteeshkd will this new QEMU still function on the old non-UPM SNP Kernel? If not we should probably also add a note to the coco docs.

No, this qemu on non-UPM kernel does not work.  Here is the error it reports.
```
$ sudo journalctl -xeu containerd
...
 Feb 13 18:41:04 amdmilan1 kata[4163390]: time="2024-02-13T18:41:04.079527912Z" level=error msg="qemu-system-x86_64: vm-type snp not supported by KVM"
```

fitzthum:
> No, this qemu on non-UPM kernel does not work. Here is the error it reports.

Ok, we'll just have to be clear that we have a change of host requirements for the next release.

fitzthum:
Also @zvonkok might be interested in this?

fidencio:
> No, this qemu on non-UPM kernel does not work. Here is the error it reports.

Please, add this to the commit message. :-)

niteeshkd:
Fixed building of qemu-snp-experimental using tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh .

tools/packaging/static-build/qemu/Dockerfile
```diff
@@ -60,7 +60,8 @@ RUN apt-get update && apt-get upgrade -y && \
 	    patch \
 	    python \
 	    python-dev \
-	    rsync \
+	    python3-venv\
+	    rsync\
```
ryansavino, line 64: Can we re-add the space here?

ryansavino (changes requested):

ryansavino (approved):

fidencio (approved):
lgtm, thanks @niteeshkd!

Issue #9094, opened 2024-02-14
Local build for nydus snapshotter missing


BbolroC:
**Which feature do you think can be improved?**

As #8953 and #8585 were merged, we could use the nydus snapshotter for confidential containers e2e test. But it turns out that a [container image](https://github.com/containerd/nydus-snapshotter/pkgs/container/nydus-snapshotter/174017689?tag=latest) `ghcr.io/containerd/nydus-snapshotter` only supports `x86_64`. This image is required to deploy the snapshotter during the coco e2e test at:

https://github.com/kata-containers/kata-containers/blob/c95c37d2ab5f2deca2bb3ff359edbe7cfc0d5efd/.github/workflows/run-kata-coco-tests.yaml#L60-L62

It would be great to build the image locally and make it available via a local registry for other platforms (e.g. internal nightly coco test for IBM SE https://github.com/kata-containers/kata-containers/actions/runs/7896678751/job/21551051487).

BbolroC:
I am closing this as https://github.com/containerd/nydus-snapshotter/pull/584 and https://github.com/containerd/nydus-snapshotter/pull/585 resolved the issue.

Issue #9093, opened 2024-02-14
vcpu allocation k8s test showing different behaviour on s390x


BbolroC:
**Which feature do you think can be improved?**

An actual test result for `k8s-sandbox-vcpus-allocation` on s390x is different from the expected result as follows:

```
k8s-sandbox-vcpus-allocation.bats
 ✗ Check the number vcpus are correctly allocated to the sandbox
   (in test file k8s-sandbox-vcpus-allocation.bats, line 27)
     `[ `kubectl logs ${pods[$i]}` -eq ${expected_vcpus[$i]} ]' failed
   [bats-exec-test:55] INFO: k8s configured to use runtimeclass
   pod/vcpus-less-than-one-with-no-limits created
   pod/vcpus-less-than-one-with-limits created
   pod/vcpus-more-than-one-with-limits created
   pod/vcpus-less-than-one-with-no-limits condition met
   pod/vcpus-less-than-one-with-limits condition met
   pod/vcpus-more-than-one-with-limits condition met
   1
   1
   1
   pod "vcpus-less-than-one-with-no-limits" deleted
   pod "vcpus-less-than-one-with-limits" deleted
   pod "vcpus-more-than-one-with-limits" deleted

1 test, 1 failure
```

The expected result for the last test `vcpus-more-than-one-with-limits` is 2. It is observed that the number of allocated vcpus on the platform only follows the rounding of `default_vcpus` like:

| default vcpus   |      cpu limits      |  # of allocated vcpus |
|----------:|-------------:|------:|
| 0.75 |  1.2 | 1 |
| 0.75 |    4   |   1 |
| 1 | 4 |    1 | 
| 1.1 | 0.5 | 2 |
| 1.1 | 4.1 | 2 |
| 4.5 | 1.2 | 5 |

We need an investigation for this. The test should be skipped until a root cause is figured out and the issue is resolved.

PR #9095, opened 2024-02-14
gha: Build nydus snapshotter image locally for unsupported platforms

BbolroC:
While deploying the nydus snapshotter, a container image `ghcr.io/containerd/nydus-snapshotter` is used, which only supports `x86_64`. This PR fulfils the following

- To enable to build the image locally and use it for the installation via a local registry for other architectures
- To make a test `k8s-sandbox-vcpus-allocation` skipped until #9093 is resolved for s390x

Once the PR is merged, the test result will be reported like:

https://github.com/kata-containers/kata-containers/actions/runs/7896678751/job/21574752063

The test results provided above were obtained using this PR.

Fixes: #9094 


fidencio:
@BbolroC, I'd rather have this one fixed on the nydus-snapshotter side, as they're already releasing the binaries for different architectures.

fidencio:
> > @BbolroC, I'd rather have this one fixed on the nydus-snapshotter side, as they're already releasing the binaries for different architectures.
> > Here's a pointer to where they do the release: https://github.com/containerd/nydus-snapshotter/blob/main/.github/workflows/release.yml#L98
> 
> Thanks for the feedback. I will raise a PR to enable a multi-arch image on the nydus snapshotter side then. 👍

Thanks, @BbolroC!

Issue #9097, opened 2024-02-14
runtime: Update runc to v1.1.12


fidencio:
Although we don't seem to be affected by
https://nvd.nist.gov/vuln/detail/CVE-2024-21626, we vendor and use the
runc package in a few different places of our code, and we better update
the package to its latest release.

PR #9098, opened 2024-02-14
runtime: Update runc to v1.1.12

fidencio:
Although we don't seem to be affected by
https://nvd.nist.gov/vuln/detail/CVE-2024-21626, we vendor and use the runc package in a few different places of our code, and we better update the package to its latest release.

Fixes: #9097


egernst (approved):
Thanks!

danmihai1 (approved):

Issue #5738, opened 2022-11-24
kata-shim memory leak


yiya1989:
The metrics function has collect all the kata-shim's net-ns netdev metrics, but the kata-shim net-ns is the hostnetwork.

`updateShimMetrics` and `UpdateRuntimeMetrics` both has this code as bellow, but `proc.NetDev()` equals to `/proc/self/net/dev` and due to the kata-shim belongs to the hostnetwork, then the `mutils.SetGaugeVecNetDev` will record all interface's metrics in the host network namespace.   The `kata_shim_netdev` only be set and update, but without reset, so when host network interface changed, then the `kata_shim_netdev` will continue increase which will result in memory continue inscrease. 
```
	proc, err := procfs.Self()
	if err != nil {
		return err
	}
        . . .
        // network device metrics
        if netdev, err := proc.NetDev(); err == nil {
	        // netdev: map[string]NetDevLine
	        for _, v := range netdev {
		        mutils.SetGaugeVecNetDev(katashimNetdev, v)
	        }
        }
```

Here is a metrics sample:
```
root@devbox-gaea:~# curl -fsS http://./metrics --unix-sock /run/vc/sbs/d29b349e5b24fb1da589942479aa6ce0e4b717f5c8020a252bd6d8c33a7e22bc/shim-monitor.sock |grep kata_shim_netdev |grep recv_compressed
kata_shim_netdev{interface="br-int",item="recv_compressed"} 0
kata_shim_netdev{interface="cni0",item="recv_compressed"} 0
kata_shim_netdev{interface="docker0",item="recv_compressed"} 0
kata_shim_netdev{interface="eth0",item="recv_compressed"} 0
kata_shim_netdev{interface="flannel.1",item="recv_compressed"} 0
kata_shim_netdev{interface="lo",item="recv_compressed"} 0
kata_shim_netdev{interface="lxcbr0",item="recv_compressed"} 0
kata_shim_netdev{interface="ovs-system",item="recv_compressed"} 0
kata_shim_netdev{interface="veth01e7bc92",item="recv_compressed"} 0
kata_shim_netdev{interface="veth10db999e",item="recv_compressed"} 0
kata_shim_netdev{interface="veth140df615",item="recv_compressed"} 0
kata_shim_netdev{interface="veth3db8b3e8",item="recv_compressed"} 0
kata_shim_netdev{interface="veth5154c3d0",item="recv_compressed"} 0
[...]

PR #9100, opened 2024-02-15
runtime: remove kata_shim_netdev metric

littlejawa:
Following discussion on #5738 - I think the `kata_shim_netdev` metrics is unneeded.
It is not only using too much memory (causing Prometheus pods to be OOM-killed in some installations), but it also reports data from interfaces that are unrelated to kata (no namespace isolation allowing the reporting of host-wide metrics).

The shim network usage is not giving any useful insight anyway - we already have hypervisor and agent metrics for that.
So rather than trying to fix the memory/namespace question, I'm suggesting to just remove the shim netdev metrics.

I'm also doing it on the rust implementation, even if it's not impacted by the initial bug as far as I can tell: In the rutime-rs case, the metrics are isolated (not verified, but that's my understanding at this point). Also it seems we don't have hypervisor netdev metrics coming from runtime-rs, but we still have the agent's network metrics.
I feel it's worth removing anyway, if only for consistency, but I'm making it in a separate commit in case reviewers ask to keep it as it is.

### DIFF
--- a/src/runtime-rs/crates/runtimes/src/shim_metrics.rs
+++ b/src/runtime-rs/crates/runtimes/src/shim_metrics.rs
@@ -33,8 +33,6 @@ lazy_static! {
 
     static ref SHIM_PROC_STAT: GaugeVec = GaugeVec::new(Opts::new(format!("{}_{}",NAMESPACE_KATA_SHIM,"proc_stat"), "Kata containerd shim v2 process statistics."), &["item"]).unwrap();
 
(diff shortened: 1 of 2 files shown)


gkurz (approved):
LGTM. Thanks @littlejawa !

gkurz:
```
go: downloading github.com/safchain/ethtool v0.2.0
verifying github.com/safchain/ethtool@v0.2.0: checksum mismatch
	downloaded: h1:dILxMBqDnQfX192cCAPjZr9v2IgVXeElHPy435Z/IdE=
	go.sum:     h1:tjsEsesUSlGdnUAAiIaEvk/YEycwk0k3Q6/q77qGpBI=
```

@littlejawa you need to rebase to get the fix from #9112 .

justxuewei:
Please modify the title to "runtime-rs: xxxx".

justxuewei (approved):
Lgtm, thanks!

PR #9108, opened 2024-02-18
ci: k8s: Fix checks used to skip confidential tests

fidencio:
This has been introduced by 53bc4a432bc9782272fb7fd1fe3113e006172294, where the condition was changed.

The correct condition is:
* If the list of supported tees does not contain the kata hypervisor and the list of supported non tees does not contain the kata hypervisor.

The error is that we were checking whether kata-hypervisor would contain the list of supported tees, and that would almost always be false (unless in the case where the list had an one and only one element).

Fixes: #9055 -- part II

### DIFF
--- a/tests/integration/kubernetes/k8s-confidential.bats
+++ b/tests/integration/kubernetes/k8s-confidential.bats
@@ -15,11 +15,11 @@ setup() {
 
 	# This check must be done with "<SPACE>${KATA_HYPERVISOR}<SPACE>" to avoid
 	# having substrings, like qemu, being matched with qemu-$something.
-	if ! [[ " ${KATA_HYPERVISOR} " =~ " ${SUPPORTED_TEE_HYPERVISORS[@]} " ]] && ! [[ " ${KATA_HYPERVISOR} " =~ " ${SUPPORTED_NON_TEE_HYPERVISORS} " ]]; then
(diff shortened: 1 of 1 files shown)


fidencio:
> I spent some time trying to debug the snp failure. I will continue to investigate this, but may be a good idea to get this in so that the confidential test runs for now.

My understanding is that SNP shouldn't work, as the patch bringing in gmem support was not merged yet.  Is this assumption valid?

stevenhorsman (approved):
My shell script interpretation logic hasn't booted yet, so I copied the if logic into a little script and tried it for all the hypervisors (and some random ones) and they all worked as I expected:
```
# ./test.sh qemu
KATA_HYPERVISOR: qemu
Setup proceeded
Non-tee - Need to apply image annotations
Confidential Launch Test not supported for qemu.
# ./test.sh qemu-sev
KATA_HYPERVISOR: qemu-sev
Setup proceeded
# ./test.sh qemu-se
KATA_HYPERVISOR: qemu-se
Setup proceeded
# ./test.sh qemu-snp
KATA_HYPERVISOR: qemu-snp
Setup proceeded
# ./test.sh qemu-tdx
KATA_HYPERVISOR: qemu-tdx
Setup proceeded
# ./test.sh qemu-fc
KATA_HYPERVISOR: qemu-fc
Test not supported for qemu-fc.
# ./test.sh qemu-junk
KATA_HYPERVISOR: qemu-junk
Test not supported for qemu-junk.
```
Thanks for the fix!

ryansavino:
> > I spent some time trying to debug the snp failure. I will continue to investigate this, but may be a good idea to get this in so that the confidential test runs for now.
> 
> My understanding is that SNP shouldn't work, as the patch bringing in gmem support was not merged yet. Is this assumption valid?

As the node kernel version has been updated, yes it shouldn't work. However, the job seemed to be failing on an earlier test.

Issue #9121, opened 2024-02-21
gha: nydus snapshotter could not be deployed without prior cleanup


ChengyuZhu6:
**Which feature do you think can be improved?**

Failed to deploy nydus snapshotter because it was not cleaned up last time.
https://github.com/kata-containers/kata-containers/actions/runs/7987544130/job/21811018961?pr=9119

PR #9122, opened 2024-02-21
gha: try to cleanup nydus snapshotter before deploying it

ChengyuZhu6:
CI failed to deploy nydus snapshotter because it was not cleaned up last time. So we can try to cleanup nydus snapshotter before deploying it.

Fixes: #9121

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -21,7 +21,7 @@ DOCKER_REGISTRY=${DOCKER_REGISTRY:-quay.io}
 DOCKER_REPO=${DOCKER_REPO:-kata-containers/kata-deploy-ci}
 DOCKER_TAG=${DOCKER_TAG:-kata-containers-latest}
 KATA_DEPLOY_WAIT_TIMEOUT=${KATA_DEPLOY_WAIT_TIMEOUT:-10m}
-SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT=${SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT:-5m}
+SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT=${SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT:-8m}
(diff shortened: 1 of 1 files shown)


stevenhorsman (approved):
LGTM. Thanks!

ryansavino (approved):
LGTM

tests/integration/kubernetes/gha-run.sh
```diff
@@ -284,6 +284,8 @@ function deploy_nydus_snapshotter() {
 	git clone -b "${nydus_snapshotter_version}" "${nydus_snapshotter_url}" "${nydus_snapshotter_install_dir}"
 
 	pushd "$nydus_snapshotter_install_dir"
+	kubectl delete -f "misc/snapshotter/nydus-snapshotter-rbac.yaml" || true
```
fidencio, line 287: @ChengyuZhu6, please, instead of just deleting the RBAC here, I'd strongly suggest you call the cleaup method so we make sure that the daemon-set is also deleted.  Deleting just the RBAC doesn't solve the issue.

fidencio (changes requested):
I've left a comment.

tests/integration/kubernetes/gha-run.sh
fidencio: @ChengyuZhu6, on a local cluster I noticed something quite interesting, if the namespace is empty, every now and then it'd be in a "Terminatting" state forever, not actually deleting it, neither being able to deploy the nydus-snapshotter without an error (as the namespace already exists at the deployment time).

I "solved" this by doing:
```sh
kubectl get ns nydus-system -o json | jq '.spec.finalizers = []' | kubectl replace --raw "/api/v1/namespaces/nydus-system/finalize" -f 
```

So, I think this is worth adding to ensure it's actually gone.

ChengyuZhu6: Fixed. Thanks for your help @fidencio !

fidencio (approved):
lgtm, thanks a lot @ChengyuZhu6!

stevenhorsman (approved):
Thanks @ChengyuZhu6 & @fidencio!

PR #9126, opened 2024-02-21
gha: Storing artifacts for logs of k8s tests garm

### LINKED ISSUE #9103: tests: Collect test artifacts
Since we switched to using GitHub actions, at least some of the tests do not capture the "test artifacts" (like the kernel log, containerd log, _etc_).

These files are critically important for debugging CI failures, so we should add the feature back.

Specifically, we used to call `collect_logs()` from https://github.com/kata-containers/tests/blob/main/.ci/teardown.sh to attach the interesting files to the jenkins job.

GabyCT:
This PR helps to store the artifacts for different logs for k8s tests on garm. This includes changes in the workflow to install the dependencies required as well as saving the artifacts files, the introduction of the teardown script which will help to collect kata env details as well as logs from the different components.

Fixes #9103

### DIFF
--- a/.github/workflows/run-k8s-tests-on-garm.yaml
+++ b/.github/workflows/run-k8s-tests-on-garm.yaml
@@ -94,7 +94,6 @@ jobs:
           name: k8s-tests-garm-${{ matrix.vmm }}
           path: /tmp/artifacts
           retention-days: 1
-          if-no-files-found: error
 
       - name: Delete kata-deploy
         if: always()
--- a/.github/workflows/run-k8s-tests-on-garm.yaml
+++ b/.github/workflows/run-k8s-tests-on-garm.yaml
@@ -85,6 +85,17 @@ jobs:
         timeout-minutes: 30
         run: bash tests/integration/kubernetes/gha-run.sh run-tests
   
+      - name: Make artifacts tarball ${{ matrix.vmm }}
+        run: bash tests/integration/kubernetes/gha-run.sh make-tarball-artifacts
+
+      - name: Archive artifacts ${{ matrix.vmm }}
+        uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


GabyCT:
@fidencio could you please help me to test this? as I am changing the workflows I want to verify that the changes are applied correctly, thanks

jodh-intel (approved):
Thanks, @GabyCT !

lgtm

dborquez:
@GabyCT nice job, but I have some comments:

For example when running a simple workload using ctr client, I've observed empty results for some paramters as for example `journalctl -t kata`, `journalctl -t kata-runtme`, etc.

@jodh-intel , @GabyCT:

How about if we record the entire journal log for the time period that the workflow is running, to have a complete picture of what happens not only at the kata level, but also in the messages of the host itself?

```
t0=$SECONDS
# workflow to execute
elapsedseconds=$(( SECONDS - t0 ))
elapsedminutes=$(( elapsedseconds / 60 ))
minutes=$(( elapsedminutes + 1 ))
journalctl --since "$minutes minutes ago"
```

jodh-intel:
@dborquez - that would indeed simplify the script as we'd just be storing a single journal file, so that wfm. Note that it's probably simpler to use a human-readable timestamp and let `journalctl` work out the delta itself:

```bash
# Format: "YYYY-MM-DD HH:MM:SS"
start_time=$(date '+%Y-%m-%d %H:%M:%S')

# run the tests

journalctl --since="$start_time" ...
```

GabyCT:
@dborquez changes applied

dborquez (approved):
LGTM, thank you @GabyCT

GabyCT:
this change will save the logs when we are running the k8s tests only

Issue #9129, opened 2024-02-22
runtime-rs: (Drgaonball) GPU passthrough failed with error InvalidOperation


Apokleos:
# Description of problem

As the result of Dragonball `VmConfigInfo` initialize a default `pci_hotplug_enabled` which is false, causing the InvalidOperation error.

```
ctr: failed to create shim task: Others("failed to handle message create container\n\nCaused by:\n    0: create\n    1: do handle device\n    2: failed to add deivce\n    3: add vfio device\n    4: insert host device failed\n    5: Failed to insert host device HostDeviceConfig { hostdev_id: \"vfio_device_0\", sysfs_path: \"/sys/bus/pci/devices/0000:45:00.0\", dev_config: VfioPciDeviceConfig { bus_slot_func: \"45:00.0\", vendor_device_id: 368578782, guest_dev_id: Some(1), clique_id: None } }\n    6: failed to add VFIO passthrough device: VfioDeviceMgr(InvalidOperation)\n    7: failure in device manager while managing VFIO device, InvalidOperation\n    8: invalid device manager operation"): unknown)
```
# Expected result

Passthrough GPU Device successfully 

# Actual result

Failed with error "failure in device manager while managing VFIO device, InvalidOperation"

# Further information
runtime-rs/dragonball with the upstream code

PR #9130, opened 2024-02-22
runtime-rs: bugfix for GPU passthrough failed with InvalidOperation.

Apokleos:
runtime-rs: bugfix for GPU passthrough failed with InvalidOperation.
    
    We need initailize the pci_hotplug_enabled with true before we do GPU
    passthrough with runtime-rs/dragonball. Otherwise it fails with error
    `InvalidOperation`.

Fixes: #9129

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
@@ -220,6 +220,7 @@ impl DragonballInner {
             max_vcpu_count: self.config.cpu_info.default_maxvcpus as u8,
             mem_type,
             mem_file_path,
+            pci_hotplug_enabled: true,
             ..Default::default()
         };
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -220,6 +220,7 @@ impl DragonballInner {
             max_vcpu_count: self.config.cpu_info.default_maxvcpus as u8,
             mem_type,
             mem_file_path,
+            pci_hotplug_enabled: true,
```
justxuewei, line 223: Do we need to read from config?

Apokleos: There's no need to get the value from the config, as it's just dragonball's behaviour. And we can do it just initialize it with `true`.

justxuewei (approved):
Lgtm, thanks!

Issue #9152, opened 2024-02-24
CLH: consider relaxing the 1 second timeout for calling CreateVM followed by BootVM


danmihai1:
Creating followed by booting up a Guest VM are more complex steps compared with executing other CLH APIs that are using the same 1 second timeout. When hitting this timeout during CI, the creation of the sandbox gets retried *on the same Host* and:

1. Sometimes we get lucky and we get a working sandbox after just a few seconds.
2. Other times the processor and/or I/O shortage continues for a while.

Either way, the test workload might eventually time out and fail. This results in unstable/unpredictable CI behavior on these slower Hosts.

When comparing:

a. Waiting a few more seconds for the initial CreateVM + BootVM to succeed on a slower Host, and 
b. Killing a half-booted Guest, retrying the same boot sequence, and hoping for the best on retry.

( a ) seems better to me.

I don't know if the 1 sec for CreateVM + BootVM is helpful for anyone. If users depend on that short timeout, maybe we should just allow this value to be configured by users. I hope changing the default value is acceptable though, to avoid the complexity of that additional configuration.

All or most examples of these timeouts that I noticed so far were on Standard_D2s_v5 running CBL-Mariner Host with MSHV as the hypervisor. Example:

https://productionresultssa2.blob.core.windows.net/actions-results/71fc19d5-224c-4f22-85c4-5ba6373872c6/workflow-job-run-90407ba3-73b7-510c-2d0c-ae58a7f23e96/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-02-24T18%3A32%3A06Z&sig=biDEcIENuljt%2FFzPX7Vm82nXnWpAcPBLXSqD6PKzvAA%3D&sp=r&spr=https&sr=b&st=2024-02-24T18%3A22%3A01Z&sv=2021-12-02

7 timeouts similar to:

```
2024-02-23T22:48:10.2684293Z # Events:
2024-02-23T22:48:10.2687074Z #   Type     Reason                  Age   From               Message
2024-02-23T22:48:10.2689046Z #   ----     ------                  ----  ----               -------
2024-02-23T22:48:10.2693729Z #   Normal   Scheduled               32s   default-scheduler  Successfully assigned kata-containers-k8s-tests/security-context-test to aks-nodepool1-76538356-vmss000000
[...]

PR #9153, opened 2024-02-24
runtime: clh: minimum 10s timeout for CreateVM + BootVM

danmihai1:
Relax the timeout for calling CLH's CreateVM + BootVM APIs. When hitting the older 1s timeout, killing a half-booted Guest and retrying the same boot sequence could have been wasteful and resulting in unstable CI testing on slower Hosts.

Fixes: #9152

### DIFF
--- a/src/runtime/virtcontainers/clh.go
+++ b/src/runtime/virtcontainers/clh.go
@@ -75,6 +75,9 @@ const (
 	clhTimeout                     = 10
 	clhAPITimeout                  = 1
 	clhAPITimeoutConfidentialGuest = 20
+	// Minimum timout for calling CreateVM followed by BootVM. Executing these two APIs
+	// might take longer than the value returned by getClhAPITimeout().
+	clhCreateAndBootVMMinimumTimeout = 10
 	// Timeout for hot-plug - hotplug devices can take more time, than usual API calls
 	// Use longer time timeout for it.
 	clhHotPlugAPITimeout                   = 5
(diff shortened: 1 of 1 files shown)


src/runtime/virtcontainers/clh.go
```diff
@@ -711,7 +714,11 @@ func (clh *cloudHypervisor) StartVM(ctx context.Context, timeout int) error {
 	}
 	clh.state.PID = pid
 
-	ctx, cancel := context.WithTimeout(ctx, clh.getClhAPITimeout()*time.Second)
+	bootTimeout := clh.getClhAPITimeout()
+	if bootTimeout < clhCreateAndBootVMMinimumTimeout {
+		bootTimeout = clhCreateAndBootVMMinimumTimeout
+	}
+	ctx, cancel := context.WithTimeout(ctx, bootTimeout*time.Second)
```
fidencio, line 721: @danmihai1, conceptually this is correct.  I'm wondering only on whether we could simply use the clhTimeout instead though.  Let me know your thoughts on this and why you decided to add a specific new var.

danmihai1: @fidencio, my preference as a Kata relative newcomer would be to replace all of the timeouts below with a single number between 10 and 20. I didn't propose that just because I didn't want to rock the boat more than I had to - given that I am not familiar with a lot of the Kata history. Please let me know in case you think simplifying the code by reducing the number of Timeout variables would acceptable.

clhTimeout is a harder to understand name compared with clhCreateAndBootVMMinimumTimeout. Also, clhTimeout seems to be intended strictly for the CLH API to become available - not intended as an API call timeout. But again, I don't know the history here, so reusing clhTimeout would be OK with me too.

	clhTimeout                     = 10
	clhAPITimeout                  = 1
	clhAPITimeoutConfidentialGuest = 20
	// Timeout for hot-plug - hotplug devices can take more time, than usual API calls
	// Use longer time timeout for it.
	clhHotPlugAPITimeout                   = 5
	clhStopSandboxTimeout                  = 3
	clhStopSandboxTimeoutConfidentialGuest = 10

Thanks!

fidencio (approved):
lgtm, thanks @danmihai1!

sprt (approved):
Thanks Dan!

Issue #1834, opened 2021-05-12
RFC [Do not read yet]: Separate trust domains for tenant and host


c3d:
[Unfinished business - No need to read until I remove this label]

# Summary

In the context of Confidential Computing Enablement (#1332), we break the assumption that the containers trust the host. It makes clear that the trust domains for the container workload and for the pod running it do not belong to the same owner. This RFC explores the consequences in terms of API and usage model.

# Motivation

Hardware-level enablement of features such as memory encryption, CPU state encryption and memory integrity protection are not sufficient to ensure confidentiality. Exploring this question has shown that achieving real confidentiality raises serious issues regarding system administration, and may not be achievable without deep changes in the existing APIs, notably with respect to the data flow between components of Kata Containers (in particular between runtime and agent).

## Objective

The objective of this document issue is to explore the implications of confidentiality with respect to the architecture of Kata. The following aspects seem particularly important:

1. _Confidentiality_: We don't want a design where the door is locked but the window is open. So we need to ensure that the benefits of technologies such as memory encryption are not defeated by Kata communication channels.
2. _Flexibility_: Very early in the platform support for confidentiality, we already have several models that behave somewhat differently, notably with respect to attestation or integrity protection. We want to accommodate various platforms, and to not settle for a lowest common denominator.
3. _Compatibility_: To the maximum extent, we want to preserve the ability to run existing workloads, or alternatively, to provide a clear upgrade path.
4. _Enforcement_: We want to be able to enforce confidentiality when it is requested, not to rely on the user not making mistakes.

## Non-objective

1. _All features with confidentiality_: Some features will be incompatible with confidentiality, for example VGPU (at least until GPU vendors themselves provide support for confidentiality). The "enforcement" objective means it is not an objective to support confidentiality with these features, but to detect such incompatibilities as much as possible.
2. _Fork_: As much as possible, we want to avoid a confidential-only fork of Kata, knowing that some other projects are already exploring other ways to deliver confidential containers.

# Proposal

This RFC makes the following recommendations to address the problem:

1. Introduce a new terminology, _hosts_ vs. _tenants_, and the associated _security domains_.
2. Introduce the notion of _immutable pod_ and the implications regarding notably the OCI specification, one idea being to use annotations to pass a complete, non-incremental description of the pod and its containers.
[...]

PR #9159, opened 2024-02-26
agent: Add grpctls proxy API endpoints

ray-valdez:
This commit is an initial step towards mitigating the risk of untrustworthy host systems when running Kata Containers in the context of confidential computing. It serves to safeguard against malicious, privileged users gaining access to the vulnerable Kata control plane. This protective measure becomes crucial in scenarios where a malicious cloud service provider or administrator might intercept or compromise commands from the Kata control plane, tamper with container configuration files, execute processes within the container, retrieve workload statistics, or obtain sensitive container workload information.

## Problem statement

This commit addresses the following open issues:
[Securing the Kata Control Plane](https://github.com/confidential-containers/community/issues/53) and RFC: Separate trust realms for tenant and host #1834. A detailed history can be found in:
https://github.com/ray-valdez/kata-containers/tree/split-api-feature.

## Architectural changes

The commit introduces a new split API mode in the kata-agent, which partitions the kata-agent’s API endpoints between **host-side** and **owner-side** controllers. When this mode is enabled, the host-side controller is restricted to manage resource allocation during startup and resource recycling at termination. In contrast, the owner-side controller allows workload owners to directly manage their deployment of pods and containers. This partitioning implicitly labels kata-agent’s endpoint APIs as _host-exclusive_, _owner-exclusive_, or _shared_.

Host-exclusive and owner-exclusive APIs are assigned specifically to either the host-side or owner-side. For instance, `CreateSandbox` and `DestroySandbox` are examples of host-exclusive APIs, while `CopyFile` and `ExecProcess` are examples of owner-exclusive APIs. Shared APIs include those that must be shared to some extent between the control planes, such as `GetOOMEvent` and `GetGuestDetails`.

## Content of this commit

This commit focuses on providing a secure channel for the owner-side to access owner-exlusive and shared APIs. Future commit(s) will restrict the host-side access to owner-exclusive APIs when split mode is enabled on the kata-agent and will address the sharing of APIs between host-side and owner-side.

This commit implements the following changes:
- Introduces the split mode to the kata-agent.
- Integrates a gRPC TLS server to handle API requests from the owner-side.  We refer to this as the kata-agent’s API proxy server which ensures that workload owners can establish a secure end-to-end communication channel with the kata-agent for invoking API endpoint commands.
- Utilizes the Key Broker Service (KBS) to provision secrets, i.e., cryptographic public and private key pairs. These secrets are crucial for establishing a secure communication channel between the owner-side and the API proxy server.

## Testing

To enable split mode functionality, the following steps are required:

1. Configuration: Modify the `kernel_params` option in the Kata's configuration.toml file to enable split mode and specify the IP address of the KBS.
- Add following settings to the `kernel_params` option: `agent.split_api=true` and `agent.aa_kbc_params=cc_kbc::http://[IP_ADDRESS]:[PORT]`.

2. Dependency on KBS: The kata-agent relies on the KBS to provision cryptographic keys to the split API proxy server, facilitating the establishment of a secure channel.

- Generate TLS keys and certificates for kata-agent’s API proxy server and client (owner-side)

```
$ KATA_DIR=”<PATH to cloned repo>”
$ pushd ${KATA_DIR}/src/agent/grpc_tls_keys
$ ./gen_key_cert.sh
```

 - Create a zip file named 'tls-keys.zip' containing the CA public key and the server’s public and private key pair

` $ zip tls-keys.zip server.pem server.key ca.pem`

  - Place this zip file in the KBS resource path '/default/tenant-keys/'. During sandbox creation, the kata-agent retrieves this file using the KBS 'get resource' API.  It's important to note that the KBS conducts a background check on the key request, verifying evidence provided by the Trusted Execution Environment (TEE). Future extensions to the KBS will automate the creation of the server’s public and private key pair for each sandbox.

` $ popd`

## External tools required for testing 

To  exercise the API proxy server, we provide the Kata Containers agent TLS control tool (kata-agent-tls-ctl), derived from the `kata-agent-ctl` tool in another commit (see `split-api-feature` branch referenced above).

This tool communicates over a gRPC TLS channel with the kata-agent. Similar to the kata-agent-ctl, this is a low level tool that is intended for advanced users. Future commit(s) will introduce a more user-friendly tool that maintains state, designed to function as a kubectl plugin for managing owners’ workloads.

Examples of creating and starting a container using `kata-agent-tls-ctl`:

Setup environment

```
$ export guest_addr=10.89.0.28 # IP address associated with the confidential VM
$ export guest_port=50090         # API proxy server’s port (listens on)
$ export ctl=./target/x86_64-unknown-linux-musl/release/kata-agent-tls-ctl
$ export key_dir=${KATA_DIR}/src/agent/grpc_tls_key
```

**Display the status of containers in the sandbox environment**

```
$ ${ctl} -l trace connect --key-dir "${key_dir}"  --bundle-dir "${bundle_dir}"  \
--server-address  "ipaddr://${guest_addr}:${guest_port}" \
 -c  "ListContainers"
```

Set a container ID and specify an OCI spec:

```
$ container_id=9e3d1d4750e4e20945d22c358e13c85c6b88922513bce2832c0cf403f065dc6
$ OCI_SPEC_CONFIG=${KATA_DIR}/src/tools/agent-tls-ctl/config.json
```

_Note: the next two commands require pull_image support in the guest!_ 

**Create container request**
```

$ ${ctl} -l trace connect --key-dir "${key_dir}" --bundle-dir "${bundle_dir}"  \
--server-address  "ipaddr://${guest_addr}:${guest_port}" \
 -c "CreateContainer cid=${container_id} spec=file:///${OCI_SPEC_CONFIG}"
```

**Start container request**

```
$ ${ctl} -l trace connect --no-auto-values --key-dir "${key_dir}" --bundle-dir "${bundle_dir}"  \
--server-address "ipaddr://${guest_addr}:${guest_port}" \
-c "StartContainer json://{\"container_id\": \"${container_id}\"}"
```

Fixes: #1834


c3d:
I addressed issues with `cargo fmt`, but there are still build failures in the `make check` tests, which as far as I can tell are also present on `main`.

src/tools/agent-ctl/src/client.rs
```diff
@@ -1956,6 +1961,35 @@ fn agent_cmd_sandbox_set_ip_tables(
...
+    options: &mut Options,
+    args: &str,
+) -> Result<()> {
+    let mut req: SetPolicyRequest = utils::make_request(args)?;
+
+    let ctx = clone_context(ctx);
+
+    debug!(sl!(), "sending request"; "request" => format!("{:?}", req));
+
+     run_if_auto_values!(ctx, || -> Result<()> {
```
ChengyuZhu6, line 1977:
```suggestion
    run_if_auto_values!(ctx, || -> Result<()> {
```

davidhadas:
This PR interwind the addition of split control with measures to protect the communication of the trusted controller in split control. 

Split control is a Kata Agent feature and require additions to Kata Agent  source. At the same time, protecting the communication, in one way or another, is not a required  Kata Agent  measure and can be achieved outside of the Kata Agent. Specifically under Peer Pods there are other ways to protect this communication. Further, depending on the use case, more than a single communication tunnel may be required between a trusted environment and the TEE, securing each such tunnel separately results in overhead and complexity. 

The Kata-Agent dependency on KBS and Attestation suggested by this PR is non-generic. The PR suggest protecting the communication using TLS and performing dedicated key management using CoCo KBS and Attestation. It assumes a very specific architecture and does not fit other potential solutions and architectures for protecting communication to the Kata Agent. 

Suggested alternatives:
1. Remove TLS and Key management from the code completely (preferred in my view). 
2. Keep TLS as an option only (allow TLS to be flagged on/off) and remove Key management from the code.

ray-valdez:
Peer-pods and split API are orthogonal solutions. Peer-pods addresses the nested virtualization issue, while the latter addresses the untrustworthy control plane. I agree that both solutions can leverage mechanisms and techniques from each other. This will require further exploration.

> 
> To summarize, the suggested Kata-Agent dependency on KBS and Attestation and the use of TLS are non-generic and do not fit other potential solutions and architectures for protecting communication to the TEE.

Although the `kata-agent` doesn’t address communication security, it is responsible for setting up the communication channel with the host’s container runtime stack. This PR depends on the KBS and its support for different attestation architectures. Thus, we delegate most key management operations, such as key generation, key distribution, and key rotation, to the KBS. In our design, the KBS provisions the channel keys once it successfully verifies the evidence from the `kata-agent`. Similar to setting up the ttrpc server, the kata-agent initializes a gRPC TLS server once it obtains the keys from the KBS.

davidhadas:
@ray-valdez 

>This PR depends on the KBS and its support for different attestation architectures. Thus, we delegate most key management operations, such as key generation, key distribution, and key rotation, to the KBS. In our design, the KBS provisions the channel keys once it successfully verifies the evidence from the kata-agent. Similar to setting up the ttrpc server, the kata-agent initializes a gRPC TLS server once it obtains the keys from the KBS.


Split-control design can use KBS for keys and TLS. 
However, the point I am making is that it should not be part of `kata-agent` code base. 
Instead it should be separated from the `kata-agent` code - implemented externally.

Integrating this into `kata-agent` code limits what we can do with Split-control from one hand (since we will only be able to use Split-control with these specific communications security measures and not with others) and adds more complexity to  `kata-agent` code which is not required.

ray-valdez:
@davidhadas I agree that your approach could potentially broaden the focus of the split API control. However, the current split API implementation extends the kata-agent with a new proxy API server, which has minimal impact on the existing kata-agent codebase due to being implemented in an isolated module. I think the crucial issue is the implementation of the split API controller, which may have consequences for how we provide access to the kata-agent APIs. I believe that, at this point, elaborating on the split API controller would be a more beneficial step than reworking the lower-layer split API mechanism, which may change once we gain more clarity on the matter.

c3d:
Still not reviewed, reopening.

ray-valdez:
Leaving this PR open for review. Reviews pending.

burgerdev:
Hey @ray-valdez,

I see that you are still looking for reviews. Personally, I feel that the split-api proposal has too many open questions to even consider taking any steps in that direction. 

The elephant in the room: Kata's main focus is integration into Kubernetes, and you can't really tell the Kubelet to _not_ call some CRI endpoints it wants to call. 

So, what I'd like to see before considering this any further is a detailed breakdown of how this split would work in practice, on the orchestrator side.

1. Assuming an unmodified untrusted Kubelet, how do you prevent it from creating/starting/deleting containers and running probes, _without making the Kubelet assume the pod to be broken_?
2. If the goal is to stub out things from the Kubelet, which Kubernetes distro would that be happening in?
3. If the goal is to run the Kubelet in some trusted place, why is an agent-forwarder not enough?

Cheers, Markus

ray-valdez:
@burgerdev I appreciate the feedback. These are fair concerns, and I agree that the orchestrator story is the key missing piece.

To set expectations up front: this PR is not attempting to fully define or validate the orchestrator-side design yet. The intent is to keep the split-API work available while we explore and validate that direction.

On your questions:

1. The current assumption is that the kubelet remains unmodified and continues to drive initial sandbox and container creation. After that point, sensitive lifecycle APIs can be disabled via a kata-agent policy. We recognize that if the kubelet issues further lifecycle calls and receives errors, it may interpret the pod as "broken". This is a known risk and one of the areas the orchestrator design needs to address.

2. The intent is not to depend on a specific Kubernetes distro, but rather to support environments where a trusted control plane can manage in-sandbox workloads while the existing Kubernetes control plane remains responsible for scheduling and resource allocation. Clarifying the supported deployment models would be required before this could move forward.

3. Conceptually, this resembles running a minimal container management layer in a trusted location, which then manages workloads inside the sandbox. This goes beyond an agent-forwarder model by shifting container lifecycle authority away from the kubelet after initial setup.

burgerdev:
> We recognize that if the kubelet issues further lifecycle calls and receives errors, it may interpret the pod as "broken". This is a known risk and one of the areas the orchestrator design needs to address.

Is it realistic to expect these orchestrator changes? If we're talking about Kubernetes, I imagine this to be a long process with a large likelihood of being rejected.

> The intent is to keep the split-API work available while we explore and validate that direction.

I remember @zvonkok pointing out on another PR that we should not merge code that is not actually being used. In this case I second that opinion, because I just don't see where and how we're going to use it. 

Maybe other maintainers have a different point of view regarding this piece of work? We could try to align in an AC meeting.

ray-valdez:
Acknowledge the challenges with modifying Kubernetes and agree we shouldn’t merge unused code.
Closing this PR for now while exploring alternatives for a trusted orchestrator. Will revisit with a clearer plan when ready.

Issue #9064, opened 2024-02-09
RFE: Re-think Kata Containers releases


fidencio:
Folks,

I'm taking the time to write down something that's been going on on my mind for a while, and I'd really like to gather feedback from the @kata-containers/architecture-committee, @kata-containers/kata-containers-maintainer, and users of our project.

Kata Containers has had, in the past, the commitment to do one release per month on each maintained branch, be this release an alpha one or a stable one.  In the past we used to maintain 2 stable branches, currently we've been maintaining only one.

However, the last release we had was 5 months ago, and the last backported commit to the stable branch is from 3 months ago.  This is not because we haven't had content to be released, quite the opposite, we had content, but we didn't have people's power to actually commit to backport patches and cut a release.

With this said, and knowing that historically we've struggled to get our releases out of the door, I'd like to propose a few things here:
* Drop the stable branch
  * We're not able to maintain one, and this is basically just increase the burden whenever we have to cut a release
  * Downstreams are already maintaining their own stable branches, based on either one fixed release of Kata Containers (be it a stable or a non-stable one), and applying patches atop
    * We cannot fit the needs of all downstream customers with our stable branch, each downstream customer has its own need, and they know better what they need
* Considering we trust our CI ...:
  * Cut one minor release per month, cut one major release per year
    * This meanas that we'd cut:
      * 2024.02.0 -> in February
      * 2024.03.0 -> in March
      * ...
      * 2024.11.0 -> in November
      * Do not cut a release in December, as part of the community is off!
      * Then next year we'd have:
        * 2025.01.0 -> January
        * 2025.02.0 -> February
        * 2025.03.0 -> March
        * ....
        * 2025.11.0 -> November

It's importanta to say that if a CVE needs to be fixed, we'd be able to cut a YYYY.MM.n+1 release at any moment.
For instance, if a CVE shows up in January 2025, after our release, we'd be able to cut a new release, `2025.01.1` as soon as the patches land on the main branch.

Theoretically, this can help us to focus on running one simple script, which would get what we have passing CI, and consider this a release.  This would also eliminate the burden to do the maintenance for stable releases (which we currently struggle quite hard to do anyways), and it'd also simplify a lot all the tools around the project (such as kata-deploy and other installers).

Please, I'd like to hear from everyone who's interested in the topic, and we can have a final discussion about this during the vPTG, but the sooner we start discussing this the better.

[...]

PR #9170, opened 2024-02-27
release: Add the needed fixes for the release process

fidencio:
Please, check each commit for its explanation.

Fixes: #9064 -- part II


tools/packaging/release/release.sh
```diff
@@ -34,6 +34,10 @@ function _check_required_env_var()
 	local env_var
 
 	case ${1} in
+		RELEASE_VERSION) env_var="${RELEASE_VERSION}" ;;
+		GH_TOKEN) env_var="${GH_TOKEN}" ;;
+		ARCHITECURE) env_var="${ARCHITECURE}" ;;
```
sprt, line 39: `s/ARCHITECURE/ARCHITECTURE`

Unless we're passed a var with a typo?

sprt (changes requested):

tools/packaging/release/release.sh
fidencio: Fixed in this not so small new series.

.github/workflows/release-amd64.yaml
```diff
@@ -41,8 +41,10 @@ jobs:
           # We need to do such trick here as the format of the $GITHUB_REF 
           # is "refs/tags/<tag>"
           tag=$(echo $GITHUB_REF | cut -d/ -f3-)
-          tags=($tag)
-          tags+=($([[ "$tag" =~ "alpha"|"rc" ]] && echo "latest" || echo "stable"))
+          if [ "${tag}" = "main" ]; then
+              tag=$(./tools/packaging/release/release.sh next-release-version)
+          fi
+          tags=(${tag} "latest")
```
stevenhorsman, line 47: Sorry for the dumb question, but is the case when we'd expect to trigger the release workflow_dispatch from a branch other than main just if are a re-creating an old release (e.g. after someone accidentally deleted it 😱). Under that circumstance do we still want the latest tag? I guess I'm trying to work out if there is a scenario that we have 3.n released and want to re-build 3.<n-1>?

stevenhorsman (approved):
LGTM. Just one question

.github/workflows/release-amd64.yaml
fidencio: That's not a dumb question, it's a very interesting one.
Releases can be created from branches, and we won't have any branch that's used apart from main.  That was my rationale when adding this logic here.

fidencio: it should only occur if we're running from main, let me fix it.

fidencio: Updated, thanks!

danmihai1 (approved):

GabyCT (approved):

sprt (approved):
Very nice, thank you Fabiano! Love the release notes as well.

Issue #9186, opened 2024-02-29
agent: kata-agent loses container io occasionally


gaohuatao-1:
# Get your issue reviewed faster

To help us understand the problem more quickly, please do the following:

1. Run the `kata-collect-data.sh` script, which is installed as part of Kata Containers
   or `kata-containers.collect-data`, which is installed as part of the Kata Containers
   snapcraft package.
   ```sh
   $ sudo kata-collect-data.sh > /tmp/kata.log
   ```
   or
   ```sh
   $ sudo kata-containers.collect-data > /tmp/kata.log
   ```
1. Review the output file (`/tmp/kata.log`) to ensure it doesn't
   contain any private / sensitive information
1. Paste the *entire* contents of the file into this issue as a comment
   (the script generates markdown format output).

The information provided will help us to understand the problem more quickly
so saves time for both of us! :smile:

# Description of problem
We create multi-copy pods through job with yaml as follows. Then we check the container log file, the weird thing is that there is no container logs. That is to say, the last container io msg is lost.

```
apiVersion: batch/v1
kind: Job
metadata:
  name: gtest-container-io
  namespace: default
spec:
  activeDeadlineSeconds: 300
  backoffLimit: 6
  completionMode: NonIndexed
  completions: 20
  parallelism: 20
  suspend: false
  template:
    metadata:
      annotations:
    spec:
      containers:
      - args:
        - -c
        - sleep 15;str=stdout+`date`;echo $str
        command:
        - /bin/sh
        image: xxxx:v1
        imagePullPolicy: IfNotPresent
        name: ght-exec
        resources:
          limits:
            cpu: 500m
            memory: 500Mi
          requests:
            cpu: 500m
            memory: 500Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Never
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
```

# Expected result
We can get the complete container io output.
![image](https://github.com/kata-containers/kata-containers/assets/22569211/249d9099-8907-4a43-9feb-0b1d5928edd0)

# Actual result
![image](https://github.com/kata-containers/kata-containers/assets/22569211/dcf2c49d-009e-44e7-9794-3c13798d0d0c)
The 0.log file for container ght-exec is empty.

# Further information

sprt:
Potentially related: #1160 #4747 #9071

[...]

Issue #9071, opened 2024-02-09
agent,runtime: host sometimes not able to read the standard output of a remote command execution


sprt:
We noticed an issue when upgrading from v3.1.0 to v3.2.0 where the shim is sometimes not able to read the standard output of a command execution. We detected this when executing commands through the k8s control plane WebSocket API as part of the k8s conformance test suite: https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/test/e2e/common/node/pods.go#L537

The test is simple:
 * Create a long running pod
 * Open a WebSocket to the K8s control plane against `/api/v1/namespaces/default/pods/$POD_NAME/exec?command=echo&command=remote+execution+test&stderr=1&stdout=1`
 * Print messages coming from the WebSocket
 * Success: prints a message containing "remote execution test"
 * Failure observed: no message is printed
 
<details>
<summary>Simple repro script</summary>

```js
// Usage: node ws.js

// Prereqs:
// $ sudo apt install nodejs
// $ npm install ws
// $ kubectl apply -f busybox.yaml

CLUSTER_URL = "ws://k8s-control-plane-url:443";
POD_NAME = "busybox";

const fs = require('fs');
const WebSocket = require('ws');

let url = `${CLUSTER_URL}/api/v1/namespaces/default/pods/${POD_NAME}/exec?command=echo&command=foo&stderr=true&stdout=true`;

let sock = new WebSocket(url, 'channel.k8s.io', {
    // Cert only needed when using the wss:// scheme.  
    // ca: fs.readFileSync(`./ca.crt`),
    // cert: fs.readFileSync(`./client.crt`),
    // key: fs.readFileSync(`./client.key`),
});

sock.on('open', () => console.log('open'));
sock.on('message', x => console.log('message', JSON.stringify(x.toString())));
sock.on('close', () => console.log('close'));

```
</details>

We've narrowed down the issue to 6c91af0a26cf0b811487288a27b97a46d8101a86. Reverting that commit gets rid of the issue, however at the expense of another one, so this is not a proper fix.

sprt:
Okay after looking through the shim and agent code I think I've figured out what the WebSocket API does differently from the CLI - more info below.

[...]

PR #7352, opened 2023-07-16
agent: Fix exec hang issues with a backgroud process

justxuewei:
Issue #4747 and pull request #4748 fix exec hang issues where the exec command hangs when a process's stdout is not closed. However, the PR might cause the exec command not to work as expected, leading to CI failure. The PR was reverted in #7042. This PR resolves the exec hang issues and has undergone 1000 rounds of testing to verify that it would not cause any CI failures.

Fixes: #4747

Signed-off-by: Fupan Li <fupan.lfp@antgroup.com>
Signed-off-by: Xuewei Niu <niuxuewei.nxw@antgroup.com>

Issue #1160, opened 2020-12-01
2.0: exit from exec hangs if background process is present


egernst:
# Get your issue reviewed faster

(with 2.0.0 release binaries in place) 

# Description of problem

You cannot exit a exec shell if there are background processes associated with that session 

# Expected result

```
eernst@ubuntu:~$ kubectl exec -it kata-bb -- sh
/ # md5sum /dev/urandom & 
/ # exit
```

You can exit and background processes persist 

# Actual result

kubectl exec is still running. 
# Further information

test yaml:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: kata-bb
spec:
  containers:
  - name: bb
    image: busybox
    command: ["sh"]
    args: ["-c", "sleep infinity"]
    resources:
      requests:
        cpu: 4
        memory: 1Gi
      limits:
        cpu: 4
        memory: 2Gi
  runtimeClassName: kata-qemu
```

fidencio:
@egernst, I'd recommend bisecting this in order to figure out whether it ever worked before and which commit broke it.

I know you're using containerd.  I've tested and faced the same issue with cri-o (which makes me relieved as the issue seems to be on our side).

egernst:
I'm not sure this ever worked on Kata 2.0.. I believe this is likely rust-agent behavior.

liubin:
This may fix this: https://github.com/kata-containers/agent/pull/371/files

PR #9187, opened 2024-02-29
agent: do not close container io when init process exits

gaohuatao-1:
For container io, do_read_stream() can not be interrupted to prevent
loss container io when init process exists. 

Fixes: https://github.com/kata-containers/kata-containers/issues/9186

### DIFF
--- a/src/agent/src/signal.rs
+++ b/src/agent/src/signal.rs
@@ -77,10 +77,12 @@ async fn handle_sigchild(logger: Logger, sandbox: Arc<Mutex<Sandbox>>) -> Result
             p.exit_code = ret;
             let _ = p.exit_tx.take();
 
-            info!(logger, "notify term to close");
-            // close the socket file to notify readStdio to close terminal specifically
-            // in case this process's terminal has been inherited by its children.
-            p.notify_term_close();
+            if !p.init {
+                info!(logger, "notify term to close");
+                // close the socket file to notify readStdio to close terminal specifically
+                // in case this process's terminal has been inherited by its children.
+                p.notify_term_close();
+            }
         }
     }
 }


src/agent/src/signal.rs
```diff
@@ -76,11 +76,6 @@ async fn handle_sigchild(logger: Logger, sandbox: Arc<Mutex<Sandbox>>) -> Result
 
             p.exit_code = ret;
             let _ = p.exit_tx.take();
-
-            info!(logger, "notify term to close");
-            // close the socket file to notify readStdio to close terminal specifically
-            // in case this process's terminal has been inherited by its children.
-            p.notify_term_close();
```
sprt, line 83: IIRC I tried this exact same change a couple weeks ago to fix #9071, however in my testing that regressed #1160. Could you add an integration test to `k8s-exec.bats` for #1160 and for your use case?

Also have you tested this to see if it might fix your issue? 5536743361b6a3d789826cad1000d84007bb41b0

sprt (changes requested):

src/agent/src/signal.rs
gaohuatao-1: Thanks for your reply. I have updated the PR. PTAL again.  For container io, kata-agent must not terminate io to avoid loss io msg when the exit code of container init process is watched . When init process of container exited,  the children processes will be reaped by os kernel if the value of shareProcessNamespace  param is false.  That is different from exec io.

gaohuatao-1: Ok, I will add integration test in the future. Now, I find that the tokio::sync::Notify is not suitable to handle #1160 . if p. notify_term_close() is called before term_exit_notifier.notified() in function do_read_stream(), term_exit_notifier.notified() will be blocked until notify_term_close() called again. That is to say, #1160 is still exists. Of course, the probability of occurrence is lower.

gaohuatao-1:
@sprt  I have added a test to k8s-exec.bats and it's my first time to modify it.  PTAL!

sprt:
@gaohuatao-1 Could you rebase your PR on main? It seems like CI in broken on your branch

gaohuatao-1:
> @gaohuatao-1 Could you rebase your PR on main? It seems like CI in broken on your branch

Completed.

sprt:
My apologies @gaohuatao-1, the CI was still broken when you made that rebase but we should be good now. Could you rebase on 83e39a206c91aeedc7f9e83dc150bbe9a8613c6b from main? I know that that commit is clean.

gkurz (commented):
Hi @gaohuatao-1 !

Please provide a commit log that explains what this does and how it fixes #9186 .

gaohuatao-1:
> Hi @gaohuatao-1 !
> 
> Please provide a commit log that explains what this does and how it fixes #9186 .

Thanks！ Completed.

src/agent/rustjail/src/process.rs
```diff
@@ -204,11 +206,6 @@ impl Process {
         Ok(p)
     }
 
-    pub fn notify_term_close(&mut self) {
-        let notify = self.term_exit_notifier.clone();
-        notify.notify_waiters();
```
lifupan, line 209: Hi @gaohuatao-1 

I think it's ok to only notify the term close when the process wasn't the init one, but 
can you have a try with replacing the notify.notify_waiters() with notify.notify_one() ?
I think it would be work.

gaohuatao-1: Thanks for you reply.
The function notify.notify_one() will regress https://github.com/kata-containers/kata-containers/pull/7352. 
The term_exit_notifier.notified() may be called in multiple futures, such as read_stdout() and read_stderr(), if tty of pod is false.  The first call to notified() will complete immediately, but the one after that will wait.  Pls note:  in the above scenario of tty closed, there is always one stream blocked if the exec cmd has a background process.
Of course, it will be ok if tty is true, because notified() and notify_one() are in one-to-one correspondence.

gaohuatao-1:
I have a question. #1160 solves the problem of "exit from exec hangs if background process is present", but it will cause exec io to be lost. The two questions of exec hang and exec io lost are mutually exclusive. What do you think about it? @sprt

gaohuatao-1:
Hi @sprt, PTAL!

sprt:
I see a few failing tests, rerunning in case they're flakes

lifupan:
> Hi @sprt @lifupan I have a question. #1160 solves the problem of "exit from exec hangs if background process is present", but it will cause exec io to be lost occasionally. The two questions of exec hang and exec io lost are mutually exclusive. What do you think about it?

Hi @gaohuatao-1 

That's why I'm curious about this issue you met. In #1160, we had considered the io stream of the exited process, and wait until there's no io stream before we could terminated ths io copy. What I'm curious about is why when the process exit signal has been received and there is no data output in its stdout/stderr, why is there still data left in stdout/stderr?

Therefore, I still think we should understand this problem first and solve it, then there will be no problem of waiting for the background process and occasionally losing io data, including the init container process and the exec process.

gaohuatao-1:
> > Hi @sprt @lifupan I have a question. #1160 solves the problem of "exit from exec hangs if background process is present", but it will cause exec io to be lost occasionally. The two questions of exec hang and exec io lost are mutually exclusive. What do you think about it?
> 
> Hi @gaohuatao-1
> 
> That's why I'm curious about this issue you met. In #1160, we had considered the io stream of the exited process, and wait until there's no io stream before we could terminated ths io copy. What I'm curious about is why when the process exit signal has been received and there is no data output in its stdout/stderr, why is there still data left in stdout/stderr?
> 
> Therefore, I still think we should understand this problem first and solve it, then there will be no problem of waiting for the background process and occasionally losing io data, including the init container process and the exec process.



> > Hi @sprt @lifupan I have a question. #1160 solves the problem of "exit from exec hangs if background process is present", but it will cause exec io to be lost occasionally. The two questions of exec hang and exec io lost are mutually exclusive. What do you think about it?
> 
> Hi @gaohuatao-1
> 
> That's why I'm curious about this issue you met. In #1160, we had considered the io stream of the exited process, and wait until there's no io stream before we could terminated ths io copy. What I'm curious about is why when the process exit signal has been received and there is no data output in its stdout/stderr, why is there still data left in stdout/stderr?
> 
> Therefore, I still think we should understand this problem first and solve it, then there will be no problem of waiting for the background process and occasionally losing io data, including the init container process and the exec process.

The test cmd: crictl exec -ti cid /bin/sh -c "sleep 10;echo hello" .Execute the above command repeatedly, and occasionally io is lost.

gaohuatao-1:
I resubmitted. PTAL again! @sprt @lifupan

gaohuatao-1:
> Hi @sprt @lifupan I have a question. #1160 solves the problem of "exit from exec hangs if background process is present", but it will cause exec io to be lost occasionally. The two questions of exec hang and exec io lost are mutually exclusive. What do you think about it?
I did test for different runtime and the  phenomenon is as follows
For runc, the exec cmd do not hang and io may be lost occasionally when tty is true.The exec cmd will hang but io is not lost when tty is false.
For kata3, the exec cmd do not hang and io may be lost occasionally no matter what the value of tty is.
The test comand is "crictl exec -ti cid /bin/sh -c "sleep 10;echo hello" or "crictl exec cid /bin/sh -c "sleep 10;echo hello".

sprt:
Hi @gaohuatao-1, this PR has gone through many *significantly* different iterations and it hasn't been clear to me what changed between each iteration or why. Could you please provide the following to help us review your PR and make sure we're not introducing regressions:

 1. A *detailed* explanation of your current change and how exactly it fixes #9186 while still avoiding #1160. Please assume minimal knowledge from reviewers and discuss in which cases each code path is triggered.
 2. New integration tests that cover both #9186 and #1160 to ensure we don't regress now or in the future.

gaohuatao-1:
> Hi @gaohuatao-1, this PR has gone through many _significantly_ different iterations and it hasn't been clear to me what changed between each iteration or why. Could you please provide the following to help us review your PR and make sure we're not introducing regressions:
> 
> 1. A _detailed_ explanation of your current change and how exactly it fixes [agent: kata-agent loses container io occasionally #9186](https://github.com/kata-containers/kata-containers/issues/9186) while still avoiding [2.0: exit from exec hangs if background process is present #1160](https://github.com/kata-containers/kata-containers/issues/1160). Please assume minimal knowledge from reviewers and discuss in which cases each code path is triggered.
> 2. New integration tests that cover both [agent: kata-agent loses container io occasionally #9186](https://github.com/kata-containers/kata-containers/issues/9186) and [2.0: exit from exec hangs if background process is present #1160](https://github.com/kata-containers/kata-containers/issues/1160) to ensure we don't regress now or in the future.

@sprt  PTAL again. Now 

> Hi @gaohuatao-1, this PR has gone through many _significantly_ different iterations and it hasn't been clear to me what changed between each iteration or why. Could you please provide the following to help us review your PR and make sure we're not introducing regressions:
> 
> 1. A _detailed_ explanation of your current change and how exactly it fixes [agent: kata-agent loses container io occasionally #9186](https://github.com/kata-containers/kata-containers/issues/9186) while still avoiding [2.0: exit from exec hangs if background process is present #1160](https://github.com/kata-containers/kata-containers/issues/1160). Please assume minimal knowledge from reviewers and discuss in which cases each code path is triggered.
> 2. New integration tests that cover both [agent: kata-agent loses container io occasionally #9186](https://github.com/kata-containers/kata-containers/issues/9186) and [2.0: exit from exec hangs if background process is present #1160](https://github.com/kata-containers/kata-containers/issues/1160) to ensure we don't regress now or in the future.

@sprt  Hi, I'm making this MR simpler, addressing only a single issue:  "container io loss occasionally".  PTAL again! Thanks! I updated this PR just now.  Just only do one thing that agent does not close container io when init process exits.

Issue #9194, opened 2024-02-29
runtime: Set maxvcpus to the number of vCPUs used for static resources case


alex-matei:
**Which feature do you think can be improved?**

The number of max vCPUs can be much higher then the actual number of vCPUs used. The value can safely be set to the actual number of vCPUs that the VM will use since CPU hotplugging will be no op when static_sandbox_resources_mgmt is set to true.
Booting with a high number of max vCPUs is a bit slower compared to a lower number.

**How can it be improved?**

default_maxcpus = num_vcpus

PR #9195, opened 2024-02-29
runtime: Set maxvcpus equal to vcpus for the static resources case

alex-matei:
This patch sets maxvcpus equal to vcpus for the static sandbox resources management case. The value can safely be set to the actual number of vCPUs that the VM will use since CPU hotplugging will be no op when static_sandbox_resources_mgmt is set to true. The value gets passed down to the kernel when using clh/qemu. Booting with a high number of max vCPUs is a bit slower compared to a lower number.

Fixes: #9194


lifupan (approved):
LGTM, thanks.

Apokleos (approved):
Thx @alex-matei LGTM!

pmores:
This PR has enough acks to be merged (and it looks good to me, too) but CI is failing for reasons that look unrelated to me (though admittedly I don't really understand the cause of failure in case of the kata-deploy test).  I'm not familiar enough with CI unfortunately, could someone take a look and make it pass?

alex-matei:
@BbolroC I rebased on top of main.

Issue #9102, opened 2024-02-16
containerd-shim-kata-v2 ran into the below panic when "enable_tracing = true"


liubocflt:
Hi devs,

# Description of problem

While running kata release 3.1, I followed the tracing doc (jaeger and kata) and enabled tracing capability of "runtime" only.
However, on container exit `containerd-shim-kata-v2` ran into the below panic, although jeager collector did receive some time span before panic.
```
SAGE" : "time=\"2024-02-15T23:55:03.947563413Z\" level=warning msg=\"clh.VmmPingGet API call failed\" error=\"Get \\\"http://localhost/api/v1/vmm.ping\\\": dial unix /run/user/1479/run/vc/vm/65e18f8a88fc48e14af0a35853a0a0faa8b5924f9185cf611a1babf153de5308/clh-api.sock: connect: no such file or directory\" name=containerd-shim-v2 pid=728215 sandbox=65e18f8a88fc48e14af0a3
SAGE" : "time=\"2024-02-15T23:55:04.336415771Z\" level=info msg=\"RunPodSandbox for &PodSandboxMetadata{Name:liubo-pod-fb59dc8db-v58pp,Uid:a9002547-b2d8-4fe4-9080-b2d31d95c56e,Namespace:liubo-parallel-128,Attempt:9,} returns sandbox id \\\"65e18f8a88fc48e14af0a35853a0a0faa8b5924f9185cf611a1babf153de5308\\\"\"" }
SAGE" : "time=\"2024-02-15T23:55:04.903086619Z\" level=info msg=\"PullImage \\\"alpine:latest\\\"\"" }
SAGE" : "time=\"2024-02-15T23:55:05.038197216Z\" level=info msg=\"ImageUpdate event &ImageUpdate{Name:docker.io/library/alpine:latest,Labels:map[string]string{io.cri-containerd.image: managed,},XXX_unrecognized:[],}\"" }
SAGE" : "time=\"2024-02-15T23:55:05.042279370Z\" level=info msg=\"ImageUpdate event &ImageUpdate{Name:sha256:05455a08881ea9cf0e752bc48e61bbd71a34c029bb13df01e40e3e70e0d007bd,Labels:map[string]string{io.cri-containerd.image: managed,},XXX_unrecognized:[],}\"" }
[...]

PR #9201, opened 2024-03-01
katautils: fix panic on tracing.

liubogithub:
This fixes a panic on tracing on container exit.

The root cause is that global var needs to be set by "=" instead of ":=".

Fixes: #9102


Apokleos (approved):
Thx @liubogithub LGTM

justxuewei (approved):
Lgtm, thanks!

liubogithub:
FYI. 

This should at least be backported to prior to 3.1.

justxuewei:
> This should at least be backported to prior to 3.1.

The stable branches will no longer be maintained. Please refer to https://github.com/kata-containers/kata-containers/issues/9064.

studychao (approved):
nice catch!!

PR #9203, opened 2024-03-01
release: Ensure the release-type is passed to workflows

fidencio:
We need to ensure the release type is passed down to workflows, otherwise we'll fail to get the correct release version for tagging the daemonset images.

Fixes: #9064 - part III

### DIFF
--- a/.github/workflows/release-amd64.yaml
+++ b/.github/workflows/release-amd64.yaml
@@ -5,6 +5,9 @@ on:
       target-arch:
         required: true
         type: string
+      release-type:
+        required: true
+        type: string
 
 jobs:
   build-kata-static-tarball-amd64:
@@ -42,7 +45,7 @@ jobs:
           # is "refs/tags/<tag>"
           tag=$(echo $GITHUB_REF | cut -d/ -f3-)
           if [ "${tag}" = "main" ]; then
-              tag=$(./tools/packaging/release/release.sh next-release-version)
+              tag=$(RELEASE_TYPE=${{ inputs.release-type }} ./tools/packaging/release/release.sh next-release-version)
               tags=(${tag} "latest")
           else
               tags=(${tag})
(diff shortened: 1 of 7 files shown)


.github/workflows/release-amd64.yaml
```diff
@@ -5,6 +5,10 @@ on:
       target-arch:
         required: true
         type: string
+      release-type:
+        required: false
```
stevenhorsman, line 9: If release-type isn't given won't we just error in `next-release-version` function now that we are not defaulting to minor? If there another reason I'm missing that it's not required?

gkurz: > If release-type isn't given won't we just error in `next-release-version` function now that we are not defaulting to minor? If there another reason I'm missing that it's not required?

Does `workflow_call` actually do something with `required`  ? I don't find it in https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions .

fidencio: This was a c&p error, fixed.

Also, I've tried to reduce the amount of info we add that's not that much useful for the ones consuming the release.  Hopefully with this we can have the new contributors for the next release (which is basically 5~6 releases into a single one).

stevenhorsman (approved):
LGTM. Thanks for making some more space for the first-time contributors celebration, even if the first release on this system will hopefully be the only one with such a long list of commits!

lifupan (approved):

gkurz (approved):
LGTM. Thanks @fidencio !

We cannot expect CI to pass because the current payload-after-push workflow won't pass the required tag to `tools/packaging/release/release.sh`. Let's merge this PR right away.

stevenhorsman:
The mariner tests are failing with: https://github.com/kata-containers/kata-containers/issues/9239 on this PR. I think Dan has fixed this with https://github.com/kata-containers/kata-containers/pull/9243, so in order to get this going I'll rebase this PR to pick that up...

Issue #9028, opened 2024-02-05
CI: "run-monitor (qemu, crio)" job failing on CreateContainer


wainersm:
The kata-monitor tests for QEMU hypervisor and CRI-O has failed for a while now.

For example, on https://github.com/kata-containers/kata-containers/actions/runs/7754650968/job/21149196848?pr=9003, it failed on the `CreateContainer` operation:

```
[gha-run.sh:62] INFO: Running cri-containerd tests using qemu hypervisor

* STEP: pre-checks
OK: connect to the container engine

* STEP: pull the image to be used
Image is up to date for docker.io/library/busybox@sha256:538721340ded10875f4710cad688c70e5d0ecb4dcd5e7d0c161f301f36f79414

* STEP: create workloads
OK: start workload (runc) - POD ID:c062fff4593cebb93a4a7a45f77d922d46e827058bef6c0bb0744a2720aa7c81, CID:312ad5b30e2ea970e1f5426846bb6e6b36ed161fdbad781ac4fa050a7cb3c050
E0202 10:58:52.305103    5339 remote_runtime.go:302] "CreateContainer in sandbox from runtime service failed" err="rpc error: code = Unknown desc = CreateContainer failed: the file top was not found: unknown" podSandboxID="9cba4c73f4e9386a37bd6756b1eca1fcba7433b82048cd104182e9981642a19b"
time="2024-02-02T10:58:52Z" level=fatal msg="creating container: rpc error: code = Unknown desc = CreateContainer failed: the file top was not found: unknown"
E0202 10:58:53.463181    5451 remote_runtime.go:349] "StopContainer from runtime service failed" err="rpc error: code = NotFound desc = could not find container \"312ad5b30e2ea970e1f5426846bb6e6b36ed161fdbad781ac4fa050a7cb3c050\": container with ID starting with 312ad5b30e2ea970e1f5426846bb6e6b36ed161fdbad781ac4fa050a7cb3c050 not found: ID does not exist" containerID="312ad5b30e2ea970e1f5426846bb6e6b36ed161fdbad781ac4fa050a7cb3c050"
[...]

PR #9206, opened 2024-03-03
CI: fix the issue of ci failure on crio

lifupan:
PR #8760 tentatively tried to have the shim to run in its own mount namespace for the sake of improving isolation between the sandbox and the host. Thus crio storage drivers shouldn't create a PRIVATE bind mount on their home directory. Otherwise, the container's rootfs mount wouldn't be propagated to kata runtime's mount namespace, and kata runtime couldn't access the container's rootfs files.

So, when kata cooperated with crio, crio should set skip_mount_home=true for its storage overlay.

Fixes: #9028

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -620,6 +620,10 @@ function install_crio() {
 	# See: https://github.com/kata-containers/kata-containers/issues/8034
 	sudo mkdir -p /etc/crio/crio.conf.d/
 	cat <<EOF | sudo tee /etc/crio/crio.conf.d/00-default-capabilities
+[crio]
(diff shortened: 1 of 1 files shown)


BbolroC (approved):
LGTM, it makes sense. Thanks @lifupan !

littlejawa:
I have a concern on this: this configuration on the crio side is applied to all containers, not only kata containers.
This will solve the problem on our CI, but it's not something we can push in production because we can't change the behavior of all containers for the cluster just because of kata containers.
In other words: our CI may pass, but our users may still see a problem.

Issue #9207, opened 2024-03-03
Dragonball: fix several bugs in Dragonball unit test when switch to github virt machine


studychao:
1. test_signal_handler: a) There is some unknown syscalls triggered in new github virt machine that would break the make test process with `SIGSYS` after applying `SeccompFilter`. In order to fix this, we change the allowlist in this unit test for seccompfileter into a blocklist to avoid meeting the unknown syscalls. b) lazy static `METRICS` is not fully initialize in the unit test and may lead to unstable result for this UT.

2. test_handler_insert_region: the `mmap region start guest addr `hard-code a value and later there would be check whether the mentioned addr is larger than or equal to `mem_end` (default to `host_phy_mem` >> 1) in order to satisfy the requirement for DaxMemory. Since github virt machine phy_mem is larger than previous CI machine we use, the hard-code value could no longer be worked. To fix this, we change the address to `mem_end` in unit test to avoid the influence of host machine change.

studychao:
The PR that pass the unit tests with those fixes : #9205

PR #9205, opened 2024-03-03
do-not-merge: test virt machine

studychao:

PR #9208, opened 2024-03-03
Dragonball: fix unit test problems when switching to new virt github machine

studychao:
fixes: #9207 

test_signal_handler: a) There is some unknown syscalls triggered in new github virt machine that would break the make test process with SIGSYS after applying SeccompFilter. In order to fix this, we change the allowlist in this unit test for seccompfileter into a blocklist to avoid meeting the unknown syscalls. b) lazy static METRICS is not fully initialize in the unit test and may lead to unstable result for this UT.

test_handler_insert_region: the mmap region start guest addr hard-code a value and later there would be check whether the mentioned addr is larger than or equal to mem_end (mem_end default to host_phy_mem >> 1) in order to satisfy the requirement for DaxMemory. Since github virt machine phy_mem is larger than previous CI machine we use, the hard-code value could no longer be worked. To fix this, we change the address to mem_end in unit test to avoid the influence of host machine change.

### DIFF
--- a/src/dragonball/src/device_manager/mod.rs
+++ b/src/dragonball/src/device_manager/mod.rs
@@ -1429,11 +1429,7 @@ mod tests {
             Some(vm.vm_config().clone()),
             vm.shared_info().clone(),
         );
-        #[cfg(target_arch = "x86_64")]
-        let guest_addr = GuestAddress(0x200000000000);
-        // TODO: #7290 - https://github.com/kata-containers/kata-containers/issues/7290
-        #[cfg(target_arch = "aarch64")]
-        let guest_addr = GuestAddress(0xF800000000);
+        let guest_addr = GuestAddress(*dbs_boot::layout::GUEST_MEM_END);
 
         let cache_len = 1024 * 1024 * 1024;
         let mmap_region = MmapRegion::build(
--- a/src/dragonball/src/signal_handler.rs
+++ b/src/dragonball/src/signal_handler.rs
@@ -156,28 +156,17 @@ mod tests {
 
     #[test]
     fn test_signal_handler() {
+        // When METRICS initializes lazy, it will call the call_once to add locks.
+        // If the signal interrupts the initialization process, initializing again the
+        // metrics in the signal interrupt handler will cause a deadlock.
+        lazy_static::initialize(&METRICS);
         let child = thread::spawn(move || {
             assert!(register_signal_handlers().is_ok());
 
             let filter = SeccompFilter::new(
-                vec![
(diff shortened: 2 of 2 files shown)


studychao:
The PR that passes the unit tests with those fixes : https://github.com/kata-containers/kata-containers/pull/9205

src/dragonball/src/signal_handler.rs
```diff
@@ -156,28 +156,17 @@ mod tests {
...
-                    (libc::SYS_rt_sigreturn, vec![]),
-                    (libc::SYS_sched_getaffinity, vec![]),
-                    (libc::SYS_set_tid_address, vec![]),
-                    (libc::SYS_sigaltstack, vec![]),
-                    (libc::SYS_write, vec![]),
-                ]
-                .into_iter()
-                .collect(),
-                SeccompAction::Trap,
+                vec![(libc::SYS_mkdirat, vec![])].into_iter().collect(),
```
justxuewei, line 167: Why block only the `SYS_mkdirat` system call?

src/dragonball/src/device_manager/mod.rs
```diff
@@ -1430,7 +1430,7 @@ mod tests {
             vm.shared_info().clone(),
         );
         #[cfg(target_arch = "x86_64")]
-        let guest_addr = GuestAddress(0x200000000000);
+        let guest_addr = GuestAddress(*dbs_boot::layout::GUEST_MEM_END);
         // TODO: #7290 - https://github.com/kata-containers/kata-containers/issues/7290
         #[cfg(target_arch = "aarch64")]
         let guest_addr = GuestAddress(0xF800000000);
```
justxuewei, line 1436: Should we replace `0xF800000000` with `GUEST_MEM_END` for aarch64 as well?

justxuewei (commented):
Thanks @studychao! A few comments...

src/dragonball/src/signal_handler.rs
studychao: cuase this UT mainly test using `SYS_mkdirat` when it is blocked, you could read through this UT to get the hint.

justxuewei: I see.

src/dragonball/src/device_manager/mod.rs
studychao: replaced.

justxuewei (approved):
Lgtm, thanks!

Apokleos (commented):
Thx @studychao  LGTM！

Apokleos (approved):

PR #9217, opened 2024-03-05
gha: export start_time to collect artifacts properly 

### LINKED ISSUE #9216: CI: garm jobs broken on collect artifacts
It seems the commit xxxx has broken the collect actifacts step on the garm jobs:

```
Run bash tests/integration/kubernetes/gha-run.sh collect-artifacts
[gha-run.sh:265] INFO: Collecting artifacts using clh hypervisor
tests/integration/kubernetes/gha-run.sh: line…
```

The problem is that start_time  is exported on run_tests() - https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/gha-run.sh#L212 . That variable is referenced  on collect_artifacts() - https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/gha-run.sh#L231. run_tests() and collect_artifacts() are executed on different steps of the workflow but github actions does not preserve the env variables between steps. So when collect_artifacts() is called, start_time is undefined

wainersm:
~~This reverts commit 7299dbdb43b62a4a23783e6eeda714d4d465e782.~~

~~The collect artifacts step on garm jobs are failing due to the unset `start_time` variable.~~

Reverting didn't work.

---

The jobs running on garm will collect journal information. The data gathered
is based on the time the tests started running. The $start_time is
exported on run_tests() and used in collect_artifacts(). It happens that
run_tests() and collect_artifacts() are called on different steps of the
workflow and the environment variables aren't preserved between them,
i.e, $start_time exported on the first step is not available on the
subsequents.

To solve that issue, let's save $start_time in the file pointed out by
$GITHUB_ENV that Github actions uses to export variables. In case $GITHUB_ENV is
empty then probably it is running locally outside of Github, so it won't
save the start time value.

Fixes #9216

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -209,7 +209,16 @@ function run_tests() {
 
 	pushd "${kubernetes_dir}"
 	bash setup.sh
-	export start_time=$(date '+%Y-%m-%d %H:%M:%S')
+
+	# In case of running on Github workflow it needs to save the start time
+	# on the environment variables file so that the variable is exported on
+	# next workflow steps.
+	if [ -n "$GITHUB_ENV" ]; then
+		start_time=$(date '+%Y-%m-%d %H:%M:%S')
+		export start_time
+		echo "start_time=${start_time}" >> "$GITHUB_ENV"
+	fi
+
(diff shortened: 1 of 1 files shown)


GabyCT:
@wainersm this is giving an error about a file that it does not exist

wainersm:
> @wainersm this is giving an error about a file that it does not exist

yeah, didn't work. I updated with a different solution. Let me see what you think about it.

GabyCT (approved):

Issue #9104, opened 2024-02-16
tests: Add k3s containerd config and log


jodh-intel:
k3s uses an embedded version of containerd and configures it to log to a file, not the journal.

So, once #9103 is re-implemented, add the containerd config, config template, and log files if they exist:

- `/var/lib/rancher/k3s/agent/containerd/containerd.log`
- `/var/lib/rancher/k3s/agent/etc/containerd/config.toml`
- `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl`

PR #9223, opened 2024-03-05
tests: Add k3s artifacts

jodh-intel:
The k3s distribution of k8s uses an embedded version of containerd and configures it to log to a file, not the journal. Hence, although we collect the journal as a test artifact, we also need to collect the actual log files for containerd.

Also collect the k3s containerd config files to help with debugging.

Fixes: #9104.


GabyCT (approved):

dborquez (approved):
lgtm, thanks @jodh-intel

amshinde (approved):

Issue #9220, opened 2024-03-05
genpolicy: support sidecar containers


3u13r:
**Which feature do you think can be improved?**

Sidecar containers have been introduced in Alpha state in 1.28 and are now in beta with 1.29. This also means that this feature is enabled inside AKS by default. See: https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

This feature adds the optional `restartPolicy` field to the `Container` definition or more precise for entries in the `initContainers` section.
This field is missing in https://github.com/kata-containers/kata-containers/blob/main/src/tools/genpolicy/src/pod.rs#L95.

**How can it be improved?**

Add support to generate policies for deployments with sidecar containers.

**Additional Information**

I'm open to contributing that feature as it should just be as simple as adding the new field (as far as I can see).

PR #9224, opened 2024-03-05
genpolicy: add restartPolicy to container struct

3u13r:
This adds support for sidecar container introduced in Kubernetes 1.28.

Fixes: #9220

I tested this for our use-case and it worked well.

### DIFF
--- a/src/tools/genpolicy/src/pod.rs
+++ b/src/tools/genpolicy/src/pod.rs
@@ -137,6 +137,9 @@ pub struct Container {
     startupProbe: Option<Probe>,
 
     #[serde(skip_serializing_if = "Option::is_none")]
+    restartPolicy: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub serviceAccountName: Option<String>,
 
     #[serde(skip_serializing_if = "Option::is_none")]


danmihai1:
Thanks @3u13r !

danmihai1 (approved):

lifupan (approved):

wainersm:
hi @3u13r, non expert on this area speaking... but technically the `container` spec doesn't have a `restartPolicy` field, it might be in an initContainer only.  The genpolicy maps `initContainers` to an array of `container` struct (https://github.com/kata-containers/kata-containers/pull/9224/commits/4b6a0294f3e4f28737d547bcc73959338b5c1fcc#diff-742867f26fd08a8659faf8e8d6a4819cba63db91a7f7b8848f65cd56e1a2fc32R51) rather than defining a `initContainer` struct. Is there any bad implication by threading `container` and `initContainer` with same `container` struct?

wainersm:
Ah, it should be nice to have a sidecar test case in our k8s suite. I don't think we have such as case, nor how difficult it would be to implement and clue to CI (as it requires k8s 1.28 or greater).

danmihai1:
@3u13r , when you get a chance please rebase to the latest commit from the main branch - to pick up the fix for the cbl-mariner error.

3u13r:
> but technically the container spec doesn't have a restartPolicy field

According to https://github.com/kubernetes/kubernetes/blob/a31a465405e52bf0ca1d29f17e2d534646074072/pkg/apis/core/types.go#L2457-L2474 and the docs https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#Container the container "spec" has a `restartPolicy` field. 
Also see `For non-init containers or when this field is not specified, the restart behavior is defined by the Pod's restart policy and the container type.`

Sp I think aligning with the upstream struct is fine, right?

wainersm:
Hi @3u13r !

> > but technically the container spec doesn't have a restartPolicy field
> 
> According to https://github.com/kubernetes/kubernetes/blob/a31a465405e52bf0ca1d29f17e2d534646074072/pkg/apis/core/types.go#L2457-L2474 and the docs https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#Container the container "spec" has a `restartPolicy` field. Also see `For non-init containers or when this field is not specified, the restart behavior is defined by the Pod's restart policy and the container type.`
>

Interesting!
 
> So I think aligning with the upstream struct is fine, right?

Yes, it's how it should be. Please ignore my comment.

wainersm (approved):

PR #9242, opened 2024-03-06
gha: Add collect artifacts step to nerdctl workflow

### LINKED ISSUE #9241: gha: Add collect artifacts step to nerdctl workflow

GabyCT:
This PR adds the collect artifacts step to nerdctl workflow.

Fixes #9241

### DIFF
--- a/tests/integration/nerdctl/gha-run.sh
+++ b/tests/integration/nerdctl/gha-run.sh
@@ -58,6 +58,23 @@ function install_dependencies() {
 	sudo systemctl restart containerd
 }
 
+function collect_artifacts() {
+	if [ -z "${start_time:-}" ]; then
+		warn "tests start time is not defined. Cannot gather journal information"
+		return
+	fi
+
+	local artifacts_dir="/tmp/artifacts"
+	if [ -d "${artifacts_dir}" ]; then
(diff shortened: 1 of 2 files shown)


amshinde (approved):

tests/integration/nerdctl/gha-run.sh
```diff
@@ -58,6 +58,23 @@ function install_dependencies() {
 	sudo systemctl restart containerd
 }
 
+function collect_artifacts() {
+	if [ -z "${start_time:-}" ]; then
+		warn "tests start time is not defined. Cannot gather journal information"
+		return
+	fi
+
+	local artifacts_dir="/tmp/artifacts"
```
danmihai1, line 67: @GabyCT , should a unique directory name be generated here, just in case we'll have two or more of these scripts running in parallel someday?

danmihai1 (approved):

tests/integration/nerdctl/gha-run.sh
GabyCT: @danmihai1 yes that is a good suggestion, let me think about it how we can do that and meanwhile I will merge this PR and submitted another to improve this situation as we need to gathered some logs at the moment

Issue #9240, opened 2024-03-06
genpolicy: unblock CI related to #9239


danmihai1:
https://github.com/kata-containers/kata-containers/issues/9239 might take some time to fix, so for now we should unblock CI quickly.

Issue #9239, opened 2024-03-06
genpolicy: handle container images that don't specify the Env field


danmihai1:
Looks like this busybox image has just changed to not include the Env field, and as a result Kata CI started hitting this error:

https://github.com/kata-containers/kata-containers/actions/runs/8168001754/job/22353019227?pr=8798

```
1..1
not ok 1 Test readonly volume for pods
# (from function `auto_generate_policy' in file tests_common.sh, line 179,
#  from function `setup' in test file k8s-file-volume.bats, line 40)
#   `auto_generate_policy "${policy_settings_dir}" "${test_yaml}"' failed with status 101
# pod "node-debugger-aks-nodepool1-21914191-vmss000000-n47fm" deleted
# [bats-exec-test:55] INFO: k8s configured to use runtimeclass
# pod "node-debugger-aks-nodepool1-21914191-vmss000000-bfwgq" deleted
#
# [bats-exec-test:191] INFO: /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/genpolicy.LB1pw8ZnYK/genpolicy-settings.json: allowing exec: cat /tmp/foo.txt
# [bats-exec-test:210] INFO: /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/genpolicy.LB1pw8ZnYK/genpolicy-settings.json: allowing ReadStreamRequest
# [bats-exec-test:178] INFO: Executing: RUST_LOG=info /opt/kata/bin/genpolicy -u -y /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/test-pod-file-volume.yaml -p /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/genpolicy.LB1pw8ZnYK/rules.rego -j /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/genpolicy.LB1pw8ZnYK/genpolicy-settings.json
# [2024-03-06T17:26:31Z INFO  genpolicy::registry] ============================================
# [2024-03-06T17:26:31Z INFO  genpolicy::registry] Pulling manifest and config for "busybox"
# thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `Env`", line: 6, column: 2)', src/registry.rs:92:61
[...]

PR #9243, opened 2024-03-06
genpolicy: disable env variable verification

danmihai1:
Disable env variable verification to unblock CI, until container images that don't specify the Env variables will be handled correctly (see #9239).

Also, mark the image config Env field as optional, thus allowing policy generation for these container images.

Fixes: #9240

### DIFF
--- a/src/tools/genpolicy/rules.rego
+++ b/src/tools/genpolicy/rules.rego
@@ -550,9 +550,10 @@ allow_env(p_process, i_process, s_name) {
     print("allow_env: p env =", p_process.Env)
     print("allow_env: i env =", i_process.Env)
 
-    every i_var in i_process.Env {
-        allow_var(p_process, i_process, i_var, s_name)
-    }
+    # TODO: re-enable after fixing https://github.com/kata-containers/kata-containers/issues/9239.
+    # every i_var in i_process.Env {
+    #    allow_var(p_process, i_process, i_var, s_name)
+    # }
 
     print("allow_env: true")
 }
--- a/src/tools/genpolicy/src/registry.rs
+++ b/src/tools/genpolicy/src/registry.rs
@@ -41,7 +41,7 @@ struct DockerConfigLayer {
 struct DockerImageConfig {
     User: Option<String>,
     Tty: Option<bool>,
-    Env: Vec<String>,
+    Env: Option<Vec<String>>,
     Cmd: Option<Vec<String>>,
     WorkingDir: Option<String>,
     Entrypoint: Option<Vec<String>>,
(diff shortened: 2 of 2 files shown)


wainersm (approved):
Hi @danmihai1 ,

As a quick fix for CI, it looks good to me. I'm also assuming that the changes on [rc/tools/genpolicy/src/registry.rs](https://github.com/kata-containers/kata-containers/pull/9243/files#diff-754926dd5feb413b73f033fd1484312752a69e1b24d71fc57e422316b90c587b) are only cosmetics.

wainersm:
Out-of-curiosity, why the genpolicy need to parse the images manifests and configs?

danmihai1:
> Hi @danmihai1 ,
> 
> As a quick fix for CI, it looks good to me. I'm also assuming that the changes on [rc/tools/genpolicy/src/registry.rs](https://github.com/kata-containers/kata-containers/pull/9243/files#diff-754926dd5feb413b73f033fd1484312752a69e1b24d71fc57e422316b90c587b) are only cosmetics.

Allowing empty Env but not doing anything useful yet when Env is empty. Handling better an empty Env is tracked by separate issue https://github.com/kata-containers/kata-containers/issues/9239.

danmihai1:
> Out-of-curiosity, why the genpolicy need to parse the images manifests and configs?

A CoCo Policy tries to reject hypothetical environment variables that were injected/changed by the untrusted K8s Control Plane. Typically, the expected environment lives in the Env field of the container image config. So, genpolicy reads that environment configuration, to determine what is the expected environment.

All of the images we've been tested so far with genpolicy had a non-empty Env field - but it appears that someone removed that field from the latest version of this busybox image (that kind of change seems reasonable, but uncommon).

sprt (approved):

ChengyuZhu6 (approved):
LGTM thanks @danmihai1

danmihai1:
The behavior of metrics(qemu) was bizarre - producing very different output from metrics(clh). Trying rebase - just in case someone has already fixed that problem.

danmihai1:
> > The behavior of metrics(qemu) was bizarre - producing very different output from metrics(clh). Trying rebase - just in case someone has already fixed that problem.
> 
> @GabyCT ^ is it something we should be fixing?

In metrics(qemu) this step:

Run bash tests/metrics/gha-run.sh make-tarball-results
  bash tests/metrics/gha-run.sh make-tarball-results

refused to create the artifact for a reason that I wasn't able to understand, and although all the earlier test steps succeeded. The same step was working fine for metrics(clh).

This problem went away after I picked up a couple of additional recent PR merges.

PR #9246, opened 2024-03-07
genpolicy: default env if image doesn't have env

danmihai1:
Use containerd's default environment for container images that don't specify the Env field.

Also, re-enable policy env variable verification, now that these uncommon images are supported too.

Fixes: #9239


wainersm (approved):
Hi @danmihai1 !

I see `ok 1 Test readonly volume for pods` passing in https://github.com/kata-containers/kata-containers/actions/runs/8208649162/job/22563684338?pr=9246

which was failing before this fix.

GabyCT (approved):

Issue #8186, opened 2023-10-09
WIP: CC Adding image pull on host with snapshotter


stevenhorsman:
This feature builds upon the initial basic image pull work in https://github.com/kata-containers/kata-containers/issues/8103 as one of the possible next steps to extend it to support more of the current `CCv0` feature set.

The scope of this is to add the ability to pull the image on the host via the nydus snapshotter using tarfs to share data from the host to the guest. There are two modes for this - `image_block` and `image_block_with_verity` which also adds integrity protection to the images that are shared to ensure that anyone with host access cannot modify them

## Things to note/include with this work
- We want to port over the code from:
  - https://github.com/kata-containers/kata-containers/pull/7688 (although some of it might have been merged over as dependencies on the pull on guest code in #8103)
  - https://github.com/kata-containers/kata-containers/pull/7601
  - https://github.com/kata-containers/kata-containers/pull/7880
- New tests
  - I think we are primarily focussed on regression tests to ensure that the things that work for pull on guest, also work for pull on host, but we might want to/be able to add some dm-verity integrity specific tests later?

## Questions
- Should we split this up into two issues and address pull on host with tarfs and snapshotter first and then the integrity protection? @ChengyuZhu6

ChengyuZhu6:
>Should we split this up into two issues and address pull on host with tarfs and snapshotter first and then the integrity protection? @ChengyuZhu6

I like this idea @stevenhorsman. But the main issue is that the guest kernel used in the CoCo is blocking the tests of pulling on host, which I had created an issue to track(https://github.com/kata-containers/kata-containers/issues/8083). So I think we need to fix it first.

ChengyuZhu6:
Now the following tasks need to be completed:: 

- [ ] prepare the workflow to deploy nydus snapshotter enabling `host-share-with-block` feature.
- [ ] Support for sharing images to guest with integrity check using virtio-blk.

ChengyuZhu6:
I will outline various scenarios that this feature currently supports, as well as some challenges that need addressing. Thanks @wainersm  for the nice summary table.

ChengyuZhu6:
|Case #|pull-type|fs_driver|export_mode| CoCo supported |
|----------|-------------|------------|--------------------| ----------------------- |
|1|guest-pull |proxy| N/A| Y |
|2|host-share-image-block |blockdev| image_block_with_verity | Y |
|3|host-share-image-block |blockdev | image_block | Y |
[...]

PR #9260, opened 2024-03-11
gha: Prepare workflow to deploy nydus snapshotter with host-share-image-block feature

ChengyuZhu6:
Update `run k8s tests on coco` workflow to deploy nydus snapshotter enabling `host-share-image-block` feature,
which allows us to execute the tests with this feature active.

And here I skip running the tests for the pull type `host-share-image-block` until the PR https://github.com/kata-containers/kata-containers/pull/7837 is prepared for testing. I think this can help save CI resources.

Fixes: #8186 -- part I


tests/integration/kubernetes/gha-run.sh
```diff
@@ -452,15 +452,31 @@ function deploy_nydus_snapshotter() {
 		  misc/snapshotter/base/nydus-snapshotter.yaml \
 		  'data.FS_DRIVER' \
 		  "proxy" --style=double
+		# Disable to read snapshotter config from configmap
+		yq write -i \
+		  misc/snapshotter/base/nydus-snapshotter.yaml \
+		  'data.ENABLE_CONFIG_FROM_VOLUME' \
+		  "false" --style=double
+	elif [ "${PULL_TYPE}" == "host-share-image-block" ]; then
```
wainersm, line 465: This block cannot be tested on this PR. So I suggest to move these changes to https://github.com/kata-containers/kata-containers/pull/7837

Here we leave only a placeholder like:

```
elif [ "${PULL_TYPE}" == "host-share-image-block" ]; then
 echo "Skip deploying nydus for the pull type `host-share-image-block` until the PR https://github.com/kata-containers/kata-containers/pull/7837 is prepared"
 return
else
..
..
```

tests/integration/kubernetes/gha-run.sh
```diff
@@ -482,7 +498,17 @@ function deploy_nydus_snapshotter() {
 	popd
 
 	kubectl rollout status daemonset nydus-snapshotter -n nydus-system --timeout ${SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT}
-
+	if [ "${PULL_TYPE}" == "host-share-image-block" ]; then
```
wainersm, line 506: Likewise move this change to https://github.com/kata-containers/kata-containers/pull/7837

.github/workflows/run-kata-coco-tests.yaml
```diff
@@ -181,6 +181,8 @@ jobs:
       - name: Run tests
         timeout-minutes: 30
         run: bash tests/integration/kubernetes/gha-run.sh run-tests
+        env:
+          PULL_TYPE: ${{ matrix.pull-type }}
```
wainersm, line 185: This isn't needed because `PULL_TYPE` is already exported on all steps of the job.

tests/integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -167,6 +167,24 @@ if [ -f "${arch_config_file}" ]; then
 fi
 
 if policy_tests_enabled; then
+	# we are encountering the issue (https://github.com/kata-containers/kata-containers/issues/8407)
```
wainersm, line 170: Wouldn't this clean up block be better fitted in [cleanup_nydus_snapshotter()](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/gha-run.sh#L494)? That function will be execute every time the snapshotter is deployed in https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/gha-run.sh#L434

The `cleanup_nydus_snapshotter()` will also be executed when the CI job finishes on https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-kata-coco-tests.yaml

.github/workflows/run-kata-coco-tests.yaml
```diff
@@ -87,6 +88,7 @@ jobs:
           - nydus
         pull-type:
           - guest-pull
+          - host-share-image-block
```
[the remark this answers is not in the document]
wainersm: Looking at https://github.com/kata-containers/kata-containers/pull/7837 , it seems that this PR is in preparation to host share image-block with verity; so shouldn't the `pull-type` be something like `host-share-image-block-with-verity`?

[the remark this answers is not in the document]
ChengyuZhu6: @wainersm Yes, #7837 is set up for host sharing, and the test covers only the verification of integrity support. It’s important to note, however, that this PR also support sharing without integrity checks. The presence of integrity verification in the block device is depended on the configuration settings of the nydus snapshotter. For example, configuring `export_mode` in the nydus snapshotter to `image_block_with_verity` enables integrity verification. Conversely, setting `export_mode` to `image_block` indicates that device integrity checks will not be performed.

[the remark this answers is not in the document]
ChengyuZhu6: So I think `host-share-image-block` includes both `host-share-image-block-with-verity` and `host-share-image-block-without-verity`.

tests/integration/kubernetes/gha-run.sh
ChengyuZhu6: Done.

tests/integration/kubernetes/gha-run.sh
ChengyuZhu6: Done.

.github/workflows/run-kata-coco-tests.yaml
ChengyuZhu6: Fixed.

tests/integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -167,6 +167,24 @@ if [ -f "${arch_config_file}" ]; then
 fi
 
 if policy_tests_enabled; then
```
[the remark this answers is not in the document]
ChengyuZhu6: The issue occurs exclusively on the TEE platforms, and `policy_tests_enabled` function serves to check the platforms. I think we should implement a new function to tackle the issue, rather than relying on the existing one.
```bash
policy_tests_enabled() {
	# The Guest images for these platforms have been built using AGENT_POLICY=yes -
	# see kata-deploy-binaries.sh.
	[ "${KATA_HYPERVISOR}" == "qemu-sev" ] || [ "${KATA_HYPERVISOR}" == "qemu-snp" ] || \
		[ "${KATA_HYPERVISOR}" == "qemu-tdx" ] || [ "${KATA_HOST_OS}" == "cbl-mariner" ]
}
```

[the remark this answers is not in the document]
ChengyuZhu6: But now I move the code to `reset_k8s_images` and call it in `deploy_nydus_snapshotter`.

.github/workflows/run-kata-coco-tests.yaml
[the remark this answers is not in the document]
ChengyuZhu6: Based on my understanding, the back-end is the same as guest-pull.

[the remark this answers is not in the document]
ChengyuZhu6: >In general I'm concerned about these variations of configurations and if a single variable pull-type is enough to represent it in a scalable way.

Understanding your concern, could you tell me more about it?

tests/integration/kubernetes/run_kubernetes_tests.sh
ChengyuZhu6: now I move the code to reset_k8s_images and call it in deploy_nydus_snapshotter.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -431,6 +436,33 @@ function cleanup_snapshotter() {
...
+		ctr_args="--address /run/k3s/containerd/containerd.sock "
+	fi
+	ctr_args+="--namespace k8s.io"
+	ctr_command="sudo -E ctr ${ctr_args}"
+	pause_images=$(${ctr_command} i ls | grep 'pause' | awk '{print $1}')
+	for image in $pause_images; do
+		${ctr_command} content fetch "$image"
+	done
+	# remove related images in k8s.io namespace
+	images_to_remove=(
```
portersrc, line 455: I guess these 3 images are test images that you want cleaned up? This seems error prone (what if another test image gets added later and needs this cleanup), but I don't have a better suggestion. Something to consider in case you know of a better way to do this.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -431,6 +436,33 @@ function cleanup_snapshotter() {
...
+		"quay.io/sjenning/nginx:1.15-alpine"
+		"quay.io/prometheus/busybox:latest"
+		"quay.io/confidential-containers/test-images:largeimage"
+	)
+
+	for image in "${images_to_remove[@]}"; do
+		${ctr_command} i remove "$image" || true
```
portersrc, line 462: I'm not a `ctr` expert, but I suppose this `remove` could fail for a valid reason, but the `|| true` would hide it. I guess you have the short-circuit logic so that if the image doesn't exist, you'll still continue normally? Maybe it's better to do something like (pseudocode):
```
for image in images; do
  if image exists; then
    ctr remove image # do not use || true here
  fi
done
```

portersrc (approved):
Looks OK. Minor code you could observe or ignore. Thanks @ChengyuZhu6 !

tests/integration/kubernetes/gha-run.sh
ChengyuZhu6: Yeah, I agree with your suggestion. But currently this issue is likely to arise only with images pulled using a snapshotter other than overlayfs. In this context, I have specifically processed the images used in guest-pull. If there are better suggestions, I would like to make changes.

tests/integration/kubernetes/gha-run.sh
ChengyuZhu6: Good point. I have fixed it. Thanks.

.github/workflows/run-kata-coco-tests.yaml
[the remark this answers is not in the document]
wainersm: Hi @ChengyuZhu6 !

Please, take my words with caution because I'm still learning so I might say something that doesn't make sense.

> @wainersm Yes, #7837 is set up for host sharing, and the test covers only the verification of integrity support. It’s important to note, however, that this PR also support sharing without integrity checks. The presence of integrity verification in the block device is depended on the configuration settings of the nydus snapshotter. For example, configuring `export_mode` in the nydus snapshotter to `image_block_with_verity` enables integrity verification. Conversely, setting `export_mode` to `image_block` indicates that device integrity checks will not be performed.

ok. So setting `pull_type="host-share-image-block"` implies nydus-snapshotter's `fs_driver = "blockdev"` that uses https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/config-blockdev.toml that has `export_mode = "image_block_with_verity"`,i.e., by default it will host-share blockdev with dm-verity. Now, to set the case of host-share blockdev **without** dm-verity: what's the plan?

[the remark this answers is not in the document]
ChengyuZhu6: >First and foremost: is the case 3 supported by CoCo?
Yes, the case three is supported by both CoCo and Kata.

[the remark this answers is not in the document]
ChengyuZhu6: >If case 3 and 4 are supported and pull-type is the same as 2 then how to deploy nydus-snapshotter for these cases? I also realized that [snapshotter.sh](https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/snapshotter.sh) doesn't handle that cases

It depends on the value of `export_mode`. When set to `image_block`, case 3 is operational; conversely, setting `export_mode` to `layer_block_with_verity` enables case 4.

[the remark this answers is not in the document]
ChengyuZhu6: Unfortunately, currently the snapshotter is limited to handling a single `export_mode` at any given time. Consequently, setting `export_mode` to `image_block_with_verity` precludes support for case 3 and case 4.  😢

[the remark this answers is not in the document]
ChengyuZhu6: >Is there any use case for cases 5 and 6 in CoCo?

Yes. In CoCo, we now support on-demand pulling nydus images (including encrypted images) into the guest with `fusedev` driver, a feature provided by image-rs.
feature code: https://github.com/confidential-containers/guest-components/pull/117
test: https://github.com/confidential-containers/guest-components/pull/469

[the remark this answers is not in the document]
ChengyuZhu6: @wainersm It’s important to note that the processes of `image_block_with_verity` and `image_block` are fundamentally similar https://github.com/kata-containers/kata-containers/pull/7837/commits/572ea7f8f6da4aa8fb67c5be208e587fad1d8b8c. The key difference is that `image_block_with_verity` calls image-rs to mount the dm-device, whereas `image_block` doesn't require this step. Based on this, it’s reasonable to assume that if case 2 is working, case 3 should also work as expected.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -439,6 +439,35 @@ function cleanup_snapshotter() {
...
+	# which can happen when switching between different snapshotters.
+	# To help sort it out, we now clean up or fetch related images in k8s.io namespace.
+	ctr_args=""
+	if [ "${KUBERNETES}" = "k3s" ]; then
+		ctr_args="--address /run/k3s/containerd/containerd.sock "
+	fi
+	ctr_args+="--namespace k8s.io"
+	ctr_command="sudo -E ctr ${ctr_args}"
+	pause_images=$(${ctr_command} i ls | grep 'pause' | awk '{print $1}')
+	for image in $pause_images; do
```
stevenhorsman, line 454: minor nit:
For consistency (and extra safety) could we add the braces to `pause_image`? e.g. `for image in ${pause_images}; do`

tests/integration/kubernetes/gha-run.sh
```diff
@@ -439,6 +439,35 @@ function cleanup_snapshotter() {
...
+	# remove related images in k8s.io namespace
+	images_to_remove=(
+		"quay.io/sjenning/nginx:1.15-alpine"
+		"quay.io/prometheus/busybox:latest"
+		"quay.io/confidential-containers/test-images:largeimage"
+	)
+
+	for image in "${images_to_remove[@]}"; do
+		if ${ctr_command} i ls | grep -q "$image"; then
+			${ctr_command} i remove "$image"
```
stevenhorsman, line 466: Minor nit:
Similar comment as above for for `${image}` references

tests/integration/kubernetes/gha-run.sh
```diff
@@ -439,6 +439,35 @@ function cleanup_snapshotter() {
 	echo "::endgroup::"
 }
 
+function reset_k8s_images(){
```
stevenhorsman, line 442: minor nit: Can we add a space between the `)` and `{` please to match the formatting of the rest of the file.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -439,6 +439,35 @@ function cleanup_snapshotter() {
...
+	if [ "${KUBERNETES}" = "k3s" ]; then
+		ctr_args="--address /run/k3s/containerd/containerd.sock "
+	fi
+	ctr_args+="--namespace k8s.io"
+	ctr_command="sudo -E ctr ${ctr_args}"
+	pause_images=$(${ctr_command} i ls | grep 'pause' | awk '{print $1}')
+	for image in $pause_images; do
+		${ctr_command} content fetch "$image"
+	done
+	# remove related images in k8s.io namespace
```
stevenhorsman, line 457: I'm sympathetic to Chris' comments, but don't have a great solution. I wonder if a small improvement might be to add the error we get if this isn't cleaned up (I think it's something like `failed to get reader from content store`) in this comment just so if someone hits this problem they have a better chance of finding this code?

.github/workflows/build-checks.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
         if: ${{ matrix.component != 'runtime' }}
         run: sudo apt-get -y install musl-tools
       - name: Install devicemapper
-        if: ${{ matrix.command == 'make check' && matrix.component == 'agent' }}
+        if: ${{ (matrix.command == 'make check' || matrix.command == 'make test') && matrix.component == 'agent' }}
```
stevenhorsman, line 86: Optional suggestion:
I wonder if we could make the commit message be a bit cleaner for future references and explain why it's needed.

e.g. something like (but probably more accurate than my understanding):

> checks: Install dependency of agent unit tests
> 
> With the requirement of devicemapper in the agent as part of the
> work to add integrity checks of shared images in the agent
> we now require `libdevmapper-dev` to be installed in order
> to run the agent unit tests

?

.github/workflows/run-kata-coco-tests.yaml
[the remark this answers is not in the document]
wainersm: Talking in private with @ChengyuZhu6 I finally understood how to set a completely custom configuration file so that [snapshotter.sh](https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/snapshotter.sh) doesn´t need to change. Here is the doc that I missed: https://github.com/containerd/nydus-snapshotter/blob/main/docs/setup_snapshotter_by_daemonset.md#same-filesystem-with-different-snapshotter-configuration-and-different-nydusd-configuration

.github/workflows/run-kata-coco-tests.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
...
```
wainersm, line 36: Hi @ChengyuZhu6 ,

On the CoCo meeting today it was raised (again) the concerns with the guest-pull running on very limited scenario (and being unrealistically on production). @danmihai1 will be looking into the solution they provided to AKS.

Then there is this PR to enable testing of host-share within TEE (bare-metal) environments; which will require the commit  281ec8d5d859fef493fe2c2e43d893209feef1e1 to clean up containerd's image cache when switching between snapshotters  (a problem on CI on bare-metal only). So I'm going to suggest to enable the host-share tests be running on non-TEE (based on #9424 ) first, then drop 281ec8d5d859fef493fe2c2e43d893209feef1e1 from this PR. This way we avoid mixing up two problems that are already hard to solve in isolation :)

Or, alternatively, de-prioritize host-share for the time being...not sure it is the best course of action.

Cc @fidencio @stevenhorsman

ChengyuZhu6:
Keep open it.

Issue #9262, opened 2024-03-12
gha: building kata payload intermittently fails on ppc64le


Amulyam24:
```
Run ./tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh \
Copying /home/ubuntu/runner/_layout/_work/kata-containers/kata-containers/kata-static.tar.xz to ./tools/packaging/kata-deploy/local-build/../../kata-deploy
~/runner/_layout/_work/kata-containers/kata-containers/tools/packaging/kata-deploy ~/runner/_layout/_work/kata-containers/kata-containers
Building the image
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  107.4MB

Step 1/9 : ARG BASE_IMAGE_NAME=alpine
Step 2/9 : ARG BASE_IMAGE_TAG=3.18
error creating temporary lease: connection error: desc = "transport: Error while dialing dial unix:///run/containerd/containerd.sock: timeout": unavailable
Step 3/9 : FROM $BASE_IMAGE_NAME:$BASE_IMAGE_TAG
```

Reference:  https://github.com/kata-containers/kata-containers/actions/runs/8245434576/job/22550378236

PR #9263, opened 2024-03-12
gha: ensure that the self hosted runner is in desired state before running the workflow 

Amulyam24:
This PR adds a step to ensure that containerd is up and running before building the payload image on Power.

Fixes: #9262

### DIFF
--- a/.github/workflows/build-kata-static-tarball-ppc64le.yaml
+++ b/.github/workflows/build-kata-static-tarball-ppc64le.yaml
@@ -39,6 +39,12 @@ jobs:
       - name: Adjust a permission for repo
         run: |
           sudo chown -R $USER:$USER $GITHUB_WORKSPACE
+      
+      - name: Prepare the self-hosted runner
+        run: |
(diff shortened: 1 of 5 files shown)


Amulyam24:
> Do we also need this step in https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/release-ppc64le.yaml ?

Hi @stevenhorsman, added the step wherever necessary. PTAL!

BbolroC:
Hi, I think it would be wonderful to make the commit message more concrete (what kind of things should be proceeded in order to reach a desired state) so that maintainers for a self-hosted runner could have a look when they are stuck in a similar situation. 😉

Amulyam24:
> Hi, I think it would be wonderful to make the commit message more concrete (what kind of things should be proceeded in order to reach a desired state) so that maintainers for a self-hosted runner could have a look when they are stuck in a similar situation. 😉

Hi @BbolroC , addressed it, PTAL.

BbolroC (approved):
LGTM, thanks @Amulyam24 !

Issue #9267, opened 2024-03-13
kata-agent: missing createcontainer hook


zvonkok:
The doc states we have support for `all` lifecycle hooks. There are still some missing.
This is the first issue regarding the CreateContainer hook which is run before pivot_root but after prestart and createruntime

PR #9268, opened 2024-03-13
kata-agent: CreateContainer Hook

zvonkok:
Fixes: #9267

The doc states we have support for all lifecycle hooks. There are still some missing. This is the first issue regarding the CreateContainer hook which is run before pivot_root but after prestart and createruntime

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -601,6 +601,22 @@ fn do_init_child(cwfd: RawFd) -> Result<()> {
         unistd::close(mount_fd)?;
     }
 
+    if init {
+        // CreateContainer Hooks:
+        // before pivot_root after prestart, createruntime
+        state.pid = std::process::id() as i32;
+        state.status = oci::ContainerState::Created;
+        if let Some(hooks) = spec.hooks.as_ref() {
+            log_child!(
+                cfd_log,
+                "create_container hooks {:?}",
+                hooks.create_container
+            );
+            let mut create_container_states = HookStates::new();
+            create_container_states.execute_hooks(&hooks.create_container, Some(state.clone()))?;
+        }
+    }
+
     if to_new.contains(CloneFlags::CLONE_NEWNS) {
         // unistd::chroot(rootfs)?;
         if no_pivot {


danmihai1:
@zvonkok , I just wanted to point out that in a Confidential Containers world the Hooks are one way that an untrusted Control Plane could use to attack the trusted code inside the TEE. That's the reason why AKS Confidential Containers is currently rejecting all such Hooks, and I was planning to propose rejecting them in upstream CoCo as well. So, we might need a bunch of additional work if you think Hooks should be used in the CoCo context too.

Apokleos:
> @zvonkok , I just wanted to point out that in a Confidential Containers world the Hooks are one way that an untrusted Control Plane could use to attack the trusted code inside the TEE. That's the reason why AKS Confidential Containers is currently rejecting all such Hooks, and I was planning to propose rejecting them in upstream CoCo as well. So, we might need a bunch of additional work if you think Hooks should be used in the CoCo context too.
>
Hi @danmihai1 AFAIK, if we want agent to execute hooks, first, we need enable `guest_hook_path` in `configuration.toml` or  in other word, we need set `guest_hook_path` within custom guest path. And, in kata-agent, it's  judged the `req.guest_hook_path` empty or not.
IMO, with such method, just hiding the `guest_hook_path` in `configuration.toml`, can we reach the goal rejecting all such hooks ?

zvonkok:
@danmihai1 I think we need to discuss this and also keep in mind that we have two use cases here: Kata and CoCo. 
Please define "untrusted Control Plane", which components are untrusted? 
The CreateContainer hook executes a binary from the VM namespace in the container namespace. The host is not involved. 
I also think this should be configurable either to allow or not to allow depending on the use-case.

danmihai1:
> > @zvonkok , I just wanted to point out that in a Confidential Containers world the Hooks are one way that an untrusted Control Plane could use to attack the trusted code inside the TEE. That's the reason why AKS Confidential Containers is currently rejecting all such Hooks, and I was planning to propose rejecting them in upstream CoCo as well. So, we might need a bunch of additional work if you think Hooks should be used in the CoCo context too.
> 
> Hi @danmihai1 AFAIK, if we want agent to execute hooks, first, we need enable `guest_hook_path` in `configuration.toml` or in other word, we need set `guest_hook_path` within custom guest path. And, in kata-agent, it's judged the `req.guest_hook_path` empty or not. IMO, with such method, just hiding the `guest_hook_path` in `configuration.toml`, can we reach the goal rejecting all such hooks ?

@Apokleos , @zvonkok , AKS Confidential Containers are currently using the following brute-force blocking using Policy:

https://github.com/microsoft/kata-containers/commit/f84ac0e5a995c3e7431a13f3ec5bc462791a03cf
https://github.com/microsoft/kata-containers/commit/32c21f2298c99437be65d278fe47996386329eb9

I agree that we might be able to come up with a more sophisticated model to filter safe Hooks - I just wanted to place these issues on your radar for now, in case you didn't consider them.

danmihai1:
> @danmihai1 I think we need to discuss this and also keep in mind that we have two use cases here: Kata and CoCo. Please define "untrusted Control Plane", which components are untrusted? The CreateContainer hook executes a binary from the VM namespace in the container namespace. The host is not involved. Are you saying we do not trust the guest?
> 
> I do no think that hooks are bad per-se maybe we should guard them as well behind a policy to allow only specific actions. Hooks are currently an essential part of the GPU enablement. If we want to support GPUs with CoCo in a proper way we need to find a way to do this "correctly" without breaking the premise: "No container/code modifications needed" to run on Kata/CoCo.

I don't think Hooks are a problem in the vanilla Kata threat model - just potentially dangerous in the CoCo threat model.

I am referring to not trusting the Kata Shim and other components outside of the Guest. These untrusted components could make changes to the fields related to Hooks when calling CreateContainer and/or CreateSandbox.

zvonkok:
@danmihai1 In the case of the GPU those hooks are added in the VM namespace so even with your policies it would work. Will tag you on some more PRs coming soon regarding this, so we can discuss best approach.

rootxrishabh:
Sorry, but this is a general question. In the context of kata container there are 2 hooks - 

1. Kata Hooks which are defined inside the guestVM by enabling #guest_hook_path in the configuration.toml.
2. OCI hooks which reside on the hostOS which runs the guestVM.

Is the above information correct? If yes, then what all hooks does kata supports and do we have documentation on how to run these?

Thank you

zvonkok:
@rootxrishabh https://github.com/kata-containers/kata-containers/blob/main/docs/design/hooks-handling.md

BbolroC (approved):
LGTM, thanks @zvonkok !

lifupan (approved):
LGTM

Issue #9269, opened 2024-03-13
kata-agent: optional bind flag


zvonkok:
From https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts
type (string, OPTIONAL) The type of the filesystem to be mounted. 
bind may be only specified in the oci spec options -> flags update r#type 

The agent will ignore bind mounts if they are only specified in the OCI spec options and not in the flags.

PR #9270, opened 2024-03-13
kata-agent: optional bind flag

zvonkok:
Fixes: #9269

From https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts type (string, OPTIONAL) The type of the filesystem to be mounted. bind may be only specified in the oci spec options -> flags update r#type The agent will ignore bind mounts if they are only specified in the OCI spec options and not in the flags.

### DIFF
--- a/src/agent/rustjail/src/mount.rs
+++ b/src/agent/rustjail/src/mount.rs
@@ -218,6 +218,17 @@ pub fn init_rootfs(
             ));
         }
 
+        // From https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts
+        // type (string, OPTIONAL) The type of the filesystem to be mounted.
+        // bind may be only specified in the oci spec options -> flags update r#type
+        let m = &{
+            let mut mbind = m.clone();
+            if mbind.r#type.is_empty() && flags & MsFlags::MS_BIND == MsFlags::MS_BIND {
+                mbind.r#type = "bind".to_string();
+            }
+            mbind
+        };
+
         if m.r#type == "cgroup" {
             mount_cgroups(cfd_log, m, rootfs, flags, &data, cpath, mounts)?;
         } else {


src/agent/rustjail/src/mount.rs
```diff
@@ -218,8 +218,19 @@ pub fn init_rootfs(
...
+        // From https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts
+        // type (string, OPTIONAL) The type of the filesystem to be mounted.
+        // bind may be only specified in the oci spec options -> flags update r#type
+        let m = {
+            let mut mbind = m.clone();
+            if mbind.r#type.is_empty() && flags & MsFlags::MS_BIND == MsFlags::MS_BIND {
+                mbind.r#type = "bind".to_string();
+            }
+            mbind
+        };
```
Apokleos, line 230:
```suggestion
let m = if m.r#type.is_empty() && flags & MsFlags::MS_BIND == MsFlags::MS_BIND {
    let mut mbind = m.clone();
    mbind.r#type = "bind".into();
    mbind
} else {
    m
};
```
We need avoid using .clone() for `m` when `m.r#type` is not empty. I suggest move mbind = m.clone() into `if xxx {let mut mbind = m.clone() ... }`

Apokleos (commented):
Thx @zvonkok one comment!

src/agent/rustjail/src/mount.rs
zvonkok: This will not work because `m` is a reference borrowed from spec.Mounts, and mbind is not a reference, hence mismatched types if / else.

zvonkok:
```
        // From https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts
        // type (string, OPTIONAL) The type of the filesystem to be mounted. 
        // bind may be only specified in the oci spec options -> flags update r#type 
        let mut mbind = m.clone();
        let m = if m.r#type.is_empty() && flags & MsFlags::MS_BIND == MsFlags::MS_BIND {
            mbind.r#type = "bind".into();
            &mbind
        } else {
            m
        };
```

What about this?

zvonkok: Otherwise we would need to clone `m` in the else branch.

Apokleos: @zvonkok Apologies for the confusion. IMO, it seems that using clone is unavoidable. Hope somebody else can give us a new one if there is.
In this case, mybe it's better to keep the original code structure as much as possible. However, we need to ensure that the type of m remains &Mount. 
```suggestion
 let m = &{
      let mut mbind = m.clone();
      if mbind.r#type.is_empty() && flags & MsFlags::MS_BIND == MsFlags::MS_BIND {
          mbind.r#type = "bind".to_string();
      }
      mbind
  };
```

zvonkok: `mbind` does not live long enough

zvonkok:
@Apokleos Fixed PTAL

amshinde (approved):

justxuewei (approved):

PR #9282, opened 2024-03-14
runtime-rs: Consolidate the handling of fds passed to QEMU

### LINKED ISSUE #9281: runtime-rs: Handling of fds passed to QEMU
Some special care is needed when it comes to pass file descriptors to QEMU : they must stay open until the QEMU process is forked and closed afterwards. The current approach to handle that is a bit fragile and it doesn't scale well with the addition of new kind of fds.

Improvements could be :
- collect the fds in a single variable
- make the lifetime of fds more explicit

gkurz:
Small improvement on the passed file descriptors.

Fixes #9281

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -66,13 +66,11 @@ impl QemuInner {
 
         let mut cmdline = QemuCmdLine::new(&self.id, &self.config)?;
 
-        // If there's a Vsock Device in `self.devices` the vhost-vsock file
-        // descriptor needs to stay open until the qemu process launches.
-        // This is why we need to store it in a variable at this scope.
-        let mut _vhost_fd = None;
-        // We need to keep the vhost-net/tuntap file descriptor open until the QEMU process launches.
-        // However, we're likely not interested in the specific type of file descriptor itself. We just
-        // want to ensure any fds associated with network devices remain open within the current scope.
+        // CAUTION: File descriptors that are passed to QEMU must stay open until the QEMU process
+        // is started and closed afterwards. This is achieved by collecting them in _fds_for_qemu.
+        // It is mandatory for _fds_for_qemu to last until the QEMU process is forked. Leave it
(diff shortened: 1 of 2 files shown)


Apokleos (approved):
Nice! Thx @gkurz I have tested it in my local Dev and it works well. LGTM !

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -140,6 +137,9 @@ impl QemuInner {
         }
 
         self.qemu_process = Some(command.stderr(Stdio::piped()).spawn()?);
+        // Close the file descriptors now that QEMU was forked. This must
+        // be done here.
+        fds_for_qemu.clear();
```
justxuewei, line 142: `fds_for_qemu` is no longer used, I think just dropping it, e.g. `drop(fds_for_qemu)`.

pmores: This is actually not necessary, the `File`s will be closed implicitly when this function exits.

gkurz: > `fds_for_qemu` is no longer used, I think just dropping it, e.g. `drop(fds_for_qemu)`.

Yeah even better.

> This is actually not necessary, the `File`s will be closed implicitly when this function exits.

Yes I know but the idea is to make this explicit.

pmores: Why would we want to do so?  The whole idea of destructors is to make this implicit.  You should only `drop()` explicitly when there's a good reason to do so, otherwise it's confusing to the reader.

gkurz: Because I don't really know when the destructor will be eventually called but I do know that we want the fds to be closed as early as possible... ideally, we'd even close them before the child process execs into QEMU.

gkurz: Finally dropped this line after discussion with @pmores on slack.

justxuewei (approved):
Lgtm, thanks!

Issue #9285, opened 2024-03-15
agent: Show features enabled at build time


jodh-intel:
The agent now has a number of optional build-time features that can be enabled.

Add details of these features to the following areas:

- Version output (`kata-agent --version`)
- Announce message (so that the details are always added to the journal at agent startup).
- The response message returned by the ttRPC `GetGuestDetails()` API.

PR #9286, opened 2024-03-15
agent: Show features enabled at build time

jodh-intel:
The agent now has a number of optional build-time features that can be enabled.

Add details of these features to the following areas:

- Version output (`kata-agent --version`)
- Announce message (so that the details are always added to the journal at agent startup).
- The response message returned by the ttRPC `GetGuestDetails()` API.

Fixes: #9285.

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -1357,6 +1357,7 @@ dependencies = [
  "chrono",
  "common-path",
  "fail",
+ "hex",
  "kata-types",
  "lazy_static",
  "libc",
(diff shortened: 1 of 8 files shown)


jodh-intel:
/cc @danmihai1 as this might need to be reviewed for the `kata-agent`'s OPA policy [*].

---

[*] - Note that fwics even if `GetGuestDetails()` is disabled in the repo policy file, we currently still log a subset of the details that would be returned by that API in [`announce()`](https://github.com/kata-containers/kata-containers/blob/2c50d3c/src/agent/src/main.rs#L123).

ChengyuZhu6 (commented):
Hi @jodh-intel. Why we need update `Cargo.lock`?

jodh-intel:
@ChengyuZhu6 - the build automatically updates it in this case.

amshinde (approved):

GabyCT (approved):

cmaf (approved):
thanks @jodh-intel

PR #9288, opened 2024-03-15
Revert "tests: k8s: k8s-exec.bats auto-generated policy"

sprt:
Reverts kata-containers/kata-containers#9248


danmihai1 (changes requested):
Hopefully https://github.com/kata-containers/kata-containers/pull/9347 will bring perf back to an acceptable level.

PR #9289, opened 2024-03-16
releases: Simply the release in order to avoid pushing a commit updating the VERSION file

fidencio:
This puts on our back the resposibility to bump the VERSION file as soon as a release is cut out, in favour of allowing a fully automated release process.

@stevenhorsman, this PR is untested, and you have the rights to force-push to this branch.


docs/Release-Process.md
```diff
@@ -10,14 +10,11 @@ This document lists the tasks required to create a Kata Release.
...
-and
-[major](https://github.com/kata-containers/kata-containers/actions/workflows/release-major.yaml)
-files from the
-`kata-containers/kata-containers` repository to build and upload release
-artifacts.
-
-Those actions are manually triggered and are responsible for generating a new
+[release](https://github.com/kata-containers/kata-containers/actions/workflows/release.yaml)
+file from the `kata-containers/kata-containers` repository to build and upload
+release artifacts.
```
gkurz, line 15: It seems this would be a good place to document the manual version bump.

gkurz (dismissed):
LGTM. Please document the manual version bump and we should be good. Thanks @fidencio and welcome back ! :smile:

tools/packaging/release/release.sh
```diff
@@ -51,56 +49,9 @@ function _check_required_env_var()
...
-
-	git config user.email "katacontainersbot@gmail.com"
-	git config user.name "Kata Containers Bot"
-
-	echo "${RELEASE_VERSION}" > "${repo_root_dir}/VERSION"
-	git diff
-	git add "${repo_root_dir}/VERSION"
-	git commit -s -m "release: Kata Containers ${RELEASE_VERSION}"
-	git push
+	cat "${repo_root_dir}/VERSION"
```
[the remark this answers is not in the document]
stevenhorsman: FYI - the failure in my command is that it seems to ignore pre-releases, so at the moment we'd successful pass when it should fail. I'm not sure with the lack of alpha is we are likely to use pre-releases much going forwards though?

docs/Release-Process.md
fidencio: Please, do it and add to the PR if possible, @gkurz.
I don't want to block it based on that, and coding time is very reduced on my end.

tools/packaging/release/release.sh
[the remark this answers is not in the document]
fidencio: The release would fail straight away if we're trying to push a tag that already exists.  With this in mind, I sincerely think we can just ignore the check.

About pre-releases, I don't think we'll use them at all from now on.

.github/workflows/release.yaml
stevenhorsman, line 1: We still have a bunch of `RELEASE_TYPE` envs set here. I don't think we need them any more? If so I'm happy to edit and remove them

fidencio: Yeah, we can remove them. :-)

Issue #9292, opened 2024-03-18
kata-manager: Fix docker install when containerd is not already installed


jodh-intel:
The [`kata-manager`](https://github.com/kata-containers/kata-containers/blob/main/utils/kata-manager.sh) script will fail if called with `-D` and containerd is not installed.

PR #9293, opened 2024-03-18
kata-manager: Fix Docker install

jodh-intel:
Fix the Docker install by removing the second (erroneous) call to `containerd_installed()` in `handle_docker()`.

Without this fix, installing using Docker (`-D`) will work *iff* you already have containerd installed. However, if you do not have containerd installed, the `containerd_installed()` function returns 1, which exits the script as we're running with `set -e`, leaving a broken Docker installation.

> **Note:** containerd is installed via Docker's `get-docker.sh` script.

Fixes: #9292.


amshinde (approved):

fidencio (approved):
lgtm, thanks @jodh-intel!

Issue #9008, opened 2024-02-02
genpolicy: support insecure registries and hermetic environments


burgerdev:
**Which feature do you think can be improved?**

Policy generation with `genpolicy`. Specifically, image pulling for dm-verity calculation.

**How can it be improved?**

Add support for insecure registries, at least local ones. Many build systems restrict network access for security and reproducibility, and hermetic build rules are encouraged (e.g. Bazel, Nix). Local insecure registries would be an option for this use case, and they are supported by OCI tooling like docker, podman, crane, skopeo. 

**Additional Information**

What I'm trying to do:

```sh
crane registry serve --address localhost:8080 &

curl -sLO https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/pods/simple-pod.yaml
curl -sLO https://raw.githubusercontent.com/kata-containers/kata-containers/main/src/tools/genpolicy/rules.rego
curl -sLO https://raw.githubusercontent.com/kata-containers/kata-containers/main/src/tools/genpolicy/genpolicy-settings.json

genpolicy -y simple-pod.yaml

sed -i 's|nginx:|localhost:8080/nginx:|' simple-pod.yaml

genpolicy -y simple-pod.yaml. # <--- FAILS
```

Error message:

```
thread 'main' panicked at src/registry.rs:112:17:
Failed to pull container image manifest and config - error: RequestError(
    reqwest::Error {
        kind: Request,
        url: Url {
            scheme: "https",
            cannot_be_a_base: false,
            username: "",
            password: None,
            host: Some(
                Domain(
                    "localhost",
                ),
            ),
            port: Some(
                8080,
            ),
            path: "/v2/",
            query: None,
            fragment: None,
        },
        source: hyper::Error(
            Connect,
            Ssl(
                Error {
                    code: ErrorCode(
                        1,
                    ),
                    cause: Some(
                        Ssl(
                            ErrorStack(
                                [
                                    Error {
                                        code: 167772427,
                                        library: "SSL routines",
                                        function: "ssl3_get_record",
                                        reason: "wrong version number",
                                        file: "ssl/record/ssl3_record.c",
                                        line: 354,
                                    },
                                ],
                            ),
                        ),
                    ),
                },
                X509VerifyResult {
                    code: 0,
                    error: "ok",
                },
            ),
        ),
    },
)
```

**Before raising this enhancement request**

[...]

PR #9185, opened 2024-02-28
genpolicy: Add optional toggle to pull images using containerd

Redent0r:
Fixes: https://github.com/kata-containers/kata-containers/issues/9144

Add an optional flag that allows you to use existing containerd installation to pull and manage images.

This improvement will:

- Add support to images with v1 [manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md), such as `bprashanth/nginxhttps:1.0`
- Support images with any layer media type that is supported by containerd, such as `application/vnd.docker.distribution.manifest.v1+prettyjws` from image `registry.k8s.io/galera-install:0.1`
- Fix issue with current method where sometimes you get API throttled when pulling layers individually

PR #9294, opened 2024-03-18
genpolicy: support insecure registries and custom pause containers

burgerdev:
This PR adds support for a few registry options that allow running genpolicy in hermetic environments (e.g. build systems). Instead of adding more arguments to the individual function calls, I decided to pass around an options struct that holds the currently required options: `use_cache`, `pause_container_image` and `insecure_registries`. 

I think the right place to obtain the pause container image from is the settings file, because this is where other implementation details of the CRI currently live (such as default env vars, capabilities, etc.). The `cluster_config` section seemed like a good idea, but I'm open to alternatives. Although there already is a `pause_container` section, that one is parsed as a `KataSpec`, where putting an image reference does not make sense. 

Insecure registries are something specific to the environment where genpolicy  is run, not the container runtime environment. Since the settings file deals with settings for the latter, I decided not to put them there but rather make them arguments to the CLI.

Fixes: #9008

### DIFF
--- a/src/tools/genpolicy/genpolicy-settings.json
+++ b/src/tools/genpolicy/genpolicy-settings.json
@@ -272,7 +272,8 @@
         "confidential_guest": false
     },
     "cluster_config": {
-        "default_namespace": "default"
+        "default_namespace": "default",
+        "pause_container_image": "mcr.microsoft.com/oss/kubernetes/pause:3.6"
     },
     "request_defaults": {
         "CreateContainerRequest": {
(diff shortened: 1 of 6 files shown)


burgerdev:
cc @danmihai1

sprt:
Thanks @burgerdev! Dan will be back from vacation next week and will be able to give this a proper look then. 🙂

Redent0r (approved):
LGTM

danmihai1:
Thanks for helping out, @burgerdev !

danmihai1:
@burgerdev , when you get a chance, please fix the new "make check" issues - from https://github.com/kata-containers/kata-containers/actions/runs/8344078088/job/23067991633?pr=9294

burgerdev:
I fixed the linting issues, but still need to resolve the merge conflict with #9185.

burgerdev:
Rebased, ptal @danmihai1 @Redent0r.

danmihai1 (approved):

burgerdev:
CI seems to be having a bad day, but I don't think this change here is to blame (lots of test setup issues). Any advice on how to proceed?

sprt (approved):

Issue #9254, opened 2024-03-08
runtime-rs: ch: runtime crashes with Docker when creating network tap device with newer Cloud Hypervisor versions


jodh-intel:
```bash
$ ~/.cargo/bin/cloud-hypervisor --version
cloud-hypervisor v38.0-84-gdccc00f2

$ sudo docker --debug run -ti --runtime "io.containerd.kata-rust-johunt-ch.v2" quay.io/prometheus/busybox:latest
DEBU[0000] [hijack] End of stdout
docker: Error response from daemon: failed to create task for container: failed to create shim task: Others("failed to handle message try init runtime instance\n\nCaused by:\n    0: init runtime handler\n    1: start sandbox\n    2: set up device after start vm\n    3: failed to set up network\n    4: setup network\n    5: attach\n    6: do handle network Veth endpoint device failed.\n    7: failed to add deivce\n    8: add network device.\n    9: Server responded with an error: InternalServerError: Error from device manager: CreateVirtioNet(OpenTap(TapSetNetmask(IoctlError(35100, Os { code: 99, kind: AddrNotAvailable, message: \"Cannot assign requested address\" }))))"): unknown.
ERRO[0000] error waiting for container: context canceled
```

The journal shows:

```
time="2024-03-08T11:39:31.370784538Z" level=error msg="Handler for POST /v1.44/containers/7c644c92650df3d385ef1bf93d30e441b3435a9cc158eeaff43fcb4098957540/start returned er  ror: failed to create task for container: failed to create shim task: Others(\"failed to handle message try init runtime instance\\n\\nCaused by:\\n    0: init runtime hand  ler\\n    1: start sandbox\\n    2: set up device after start vm\\n    3: failed to set up network\\n    4: setup network\\n    5: attach\\n    6: do handle network Veth en  dpoint device failed.\\n    7: failed to add deivce\\n    8: add network device.\\n    9: Server responded with an error: InternalServerError: Error from device manager: Cr  eateVirtioNet(OpenTap(TapSetNetmask(IoctlError(35100, Os { code: 99, kind: AddrNotAvailable, message: \\\"Cannot assign requested address\\\" }))))\"): unknown"
```

More context from the journal:

```
[...]

PR #9295, opened 2024-03-18
runtime-rs: ch: Provide valid default value for NetConfig

likebreath:
The current default value of IP `0.0.0.0` with mask `0.0.0.0` will cause ioctl error when being used to create and configure TAP device, with newer version of Cloud Hypervisor [1]. This patch replaces them with valid value that are the same as the Go-lang runtime [2].

[1] https://github.com/cloud-hypervisor/cloud-hypervisor/pull/5924
[2] https://github.com/kata-containers/kata-containers/blob/e3f78527388a22d0c1a4cfea66fe5c43ec041fea/src/runtime/virtcontainers/pkg/cloud-hypervisor/client/model_net_config.go#L40-L57

Fixes: #9254

### DIFF
--- a/src/runtime-rs/crates/hypervisor/ch-config/src/lib.rs
+++ b/src/runtime-rs/crates/hypervisor/ch-config/src/lib.rs
@@ -267,8 +267,8 @@ impl Default for NetConfig {
     fn default() -> Self {
         NetConfig {
             tap: None,
-            ip: Ipv4Addr::new(0, 0, 0, 0),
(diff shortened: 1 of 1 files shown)


lifupan (approved):

justxuewei (approved):
Lgtm, thanks!

jodh-intel:
@likebreath - Thanks. These values do look rather odd, but I can confirm this fixes the issue with the rust CH driver. I guess we could potentially stop serialising those fields if not explicitly, set which would cause CH to use it's own defaults (the values you've added here).

PR #9303, opened 2024-03-19
runtime: logging with file names instead of pointers

markyangcc:
Loging with pointers make no sense to users.
Because `String()` method is not implemented in `os.File`. So expand `fds` to name list

Logging output(before),
`"Adding extra file [0xc000128f58 0xc000128d40]"`

Logging output(now),
`"Adding extra files [unix:/run/vc/vm/bash1/qmp.sock-> /dev/vhost-vsock]"`

Fixes: #9333

### DIFF
--- a/src/runtime/pkg/govmm/qemu/qemu.go
+++ b/src/runtime/pkg/govmm/qemu/qemu.go
@@ -24,6 +24,8 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/utils"
 )
 
 // Machine describes the machine type qemu will emulate.
@@ -3216,7 +3218,7 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
 	/* #nosec */
 	cmd := exec.CommandContext(ctx, path, params...)
 	if len(fds) > 0 {
-		logger.Infof("Adding extra file %v", fds)
+		logger.Infof("Adding extra files %v", utils.ExpandFdsNames(fds))
 		cmd.ExtraFiles = fds
 	}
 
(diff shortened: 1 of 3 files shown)


Apokleos:
Hi @markyangcc Thx for your contribution!
This PR doesn't have any linked issues, could you please create an issue first and then link it in this PR commit message.?

markyangcc:
Thanks reviewing~
An issue has been created and linked in the commit message.

beraldoleal (commented):
Hi @markyangcc, unless I'm missing something, this makes sense and lgtm.

The only, concern I have is that we are duplicating this func every time I need to log a file name. So my suggestion is to move this into its own func so we can reuse (if possible).

markyangcc:
Thanks reviewing，a new commit has been pushed with the aim of preventing code duplication.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -3216,7 +3216,13 @@ func LaunchCustomQemu(ctx context.Context, path string, params []string, fds []*
...
 	cmd := exec.CommandContext(ctx, path, params...)
 	if len(fds) > 0 {
-		logger.Infof("Adding extra file %v", fds)
+		logger.Infof("Adding extra files %v", func() []string {
+			names := make([]string, len(fds))
+			for i, fd := range fds {
+				names[i] = fd.Name()
+			}
+			return names
+		}())
```
Apokleos: Duplicated codes found in both `qemu.go` and `stratovirt.go` in this commit.
Though I know you add a new method to reduce the dup code in other commit, which looks enough.
We should keep commits clean and reduce unnecessary changes, so I suggest removing this commit.

Apokleos (commented):
Thx @markyangcc. comments left.

beraldoleal (approved):
LGTM, thanks

src/runtime/pkg/govmm/qemu/qemu.go
markyangcc: Thanks, I squashed two commits into a single commit.

Apokleos (approved):
Thx @markyangcc LGTM!

Issue #9304, opened 2024-03-19
CI|k8s: test including a comment not properly handled in filter_k8s_test.sh


BbolroC:
**Which feature do you think can be improved?**

Currently, we're excluding tests with YAML files located at [tests/integration/kubernetes/filter_out_per_arch](https://github.com/kata-containers/kata-containers/tree/main/tests/integration/kubernetes/filter_out_per_arch) for all platforms except `x86_64`.

However, it appears that the script `filter_k8s_test.sh` doesn't handle cases where a comment is provided alongside a test item (e.g., `k8s-sandbox-vcpus-allocation` for s390x). https://github.com/kata-containers/kata-containers/blob/46aec0f15a28cb9ff92f867c33e16f6b26f8431d/tests/integration/kubernetes/filter_k8s_test.sh#L38

Including comments next to skipped test items is beneficial for understanding why they're excluded on a particular platform. Therefore, it would be beneficial to ensure the script properly handles test items that include comments.

PR #9305, opened 2024-03-19
CI|k8s: Handle skipped tests with a comment for filter_out_per_arch

BbolroC:
This commit updates `filter_k8s_test.sh` to handle skipped tests that include comments. In addition to the existing parameter expansion, the following expansions have been added:

- Removal of a comment
- Stripping of trailing spaces

Fixes: #9304

### DIFF
--- a/tests/integration/kubernetes/filter_k8s_test.sh
+++ b/tests/integration/kubernetes/filter_k8s_test.sh
@@ -35,7 +35,19 @@ main()
 		local flag="false"
 		for SKIP_ENTRY in "${_K8S_SKIP_UNION[@]}"
 		do
-			SKIP_ENTRY="${SKIP_ENTRY#- }.bats"
+			# Remove '- ' from the beginning of the string
+			# Example: "- test.bats" -> "test.bats"
+			SKIP_ENTRY="${SKIP_ENTRY#- }"
+			# Remove a comment if it exists
+			# Example: "test.bats # comment" -> "test.bats "
+			SKIP_ENTRY="${SKIP_ENTRY%#*}"
+			# Strip trailing spaces if it exists
+			# Example: "test.bats " -> "test.bats"
+			# Explanation for parameter expansion:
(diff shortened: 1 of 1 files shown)


tests/integration/kubernetes/filter_k8s_test.sh
```diff
@@ -35,7 +35,19 @@ main()
 		local flag="false"
 		for SKIP_ENTRY in "${_K8S_SKIP_UNION[@]}"
 		do
-			SKIP_ENTRY="${SKIP_ENTRY#- }.bats"
+			# Remove all characaters including and before the matching pattern '- '
+			# Example: " - test.bats" -> "test.bats"
+			SKIP_ENTRY="${SKIP_ENTRY#- }"
```
stevenhorsman, line 40: I think this comment is wrong as this only removes a starting `- `, not characters before it?:
```
SKIP_ENTRY=" - test.bats"
echo "${SKIP_ENTRY#- }"
 - test.bats
```
I'm not sure it makes the code wrong as we get the SKIP_ENTRY from yaml, so it shouldn't be mismatching?

stevenhorsman (changes requested):

tests/integration/kubernetes/filter_k8s_test.sh
BbolroC: Yes, you are right. The expansion only removes the matching characters. The string parsed by yq should start with `-`.  I will update the comment. Thanks!

stevenhorsman (approved):
LGTM. For reference, I ran the changes locally in a separate test script and it did the conversions I expected:
```
'- k8s-caps' -> 'k8s-caps.bats'
'- k8s-inotify ' -> 'k8s-inotify.bats'
'- k8s-sandbox-vcpus-allocation # see https://github.com/kata-containers/kata-containers/issues/9093' -> 'k8s-sandbox-vcpus-allocation.bats'
```
Thanks @BbolroC

GabyCT (approved):

PR #9313, opened 2024-03-20
agent: Refactor unit tests to leverage rstest for parameterization

ChengyuZhu6:
Refactor unit tests to leverage rstest for parameterization. 

Fixes: #9314

### DIFF
--- a/src/agent/src/namespace.rs
+++ b/src/agent/src/namespace.rs
@@ -182,6 +182,7 @@ mod tests {
     use super::{Namespace, NamespaceType};
     use crate::mount::remove_mounts;
     use nix::sched::CloneFlags;
+    use rstest::rstest;
     use tempfile::Builder;
     use test_utils::skip_if_not_root;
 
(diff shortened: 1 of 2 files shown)


src/agent/src/namespace.rs
```diff
@@ -301,65 +299,20 @@ mod tests {
...
-            TestData {
-                ns_type: NamespaceType::Pid,
-                ns_flag: CloneFlags::CLONE_NEWPID,
-            },
-        ];
-
+    #[rstest]
+    #[case(NamespaceType::Ipc, CloneFlags::CLONE_NEWIPC)]
+    #[case(NamespaceType::Uts, CloneFlags::CLONE_NEWUTS)]
+    #[case(NamespaceType::Pid, CloneFlags::CLONE_NEWPID)]
```
stevenhorsman, line 313: Optional suggestions: Maybe we could add a description to the cases to help identify which one might go wrong:
```
    #[case::Ipc(NamespaceType::Ipc, CloneFlags::CLONE_NEWIPC)]
    #[case::Uts(NamespaceType::Uts, CloneFlags::CLONE_NEWUTS)]
    #[case::Pid(NamespaceType::Pid, CloneFlags::CLONE_NEWPID)]
```
?

src/agent/src/namespace.rs
```diff
@@ -226,31 +227,28 @@ mod tests {
...
+        #[case] ns_type: NamespaceType,
+        #[case] ns_name: &str,
+        #[case] ns_flag: CloneFlags,
+    ) {
+        assert_eq!(ns_name, ns_type.get());
+        assert_eq!(ns_flag, ns_type.get_flags());
     }
 
-    #[test]
+    #[rstest]
```
stevenhorsman, line 243: Hi Chengyu - this is a genuine question, I'm not saying that you have done anything wrong, but in my own projects I only use `#[rstest]` if I'm using a feature from it e.g. `[case...]` and it doesn't then I leave it as `#[test]` and have a mix of them in my unit tests. I see in this commit you've switched all the test cases to use `#[rstest]` - was there a particular reason for this?

src/agent/src/util.rs
```diff
@@ -172,90 +173,58 @@ mod tests {
...
+        assert_eq!(byte_count, reader_value.len());
 
-            let value = writer.to_string();
-            assert_eq!(value, d.reader_value, "{}", msg);
-        }
+        let value = writer.to_string();
+        assert_eq!(value, reader_value);
     }
 
+    #[rstest]
```
stevenhorsman, line 227: Similar question on here as the other commit, whether the non-modified tests need to be switched to `rstest`

stevenhorsman (commented):
I like rstest, so might be biased, but this looks good to me - I just have a couple of questions and a possibly suggestion.

src/agent/src/namespace.rs
ChengyuZhu6: Done.

src/agent/src/namespace.rs
ChengyuZhu6: Hey Steve. The reason for switching all the test cases to `#[rstest]` in this commit was to maintain consistency across the test suite.

ChengyuZhu6: But I think it's acceptable to have a mix of unit tests. Therefore, I have reverted the unit tests for which rstest is not necessary.

stevenhorsman: Great - thanks for making the changes

stevenhorsman (approved):
LGTM. Thanks @ChengyuZhu6!

arronwy (approved):
Thanks @ChengyuZhu6 LGTM!

PR #9316, opened 2024-03-20
release: Skip --generate-notes for this release

fidencio:
This release is a special case, as we've slacked for 6 months and the release content is way too long ... long enough to exceed the allowed limit for the release notes.

With this in mind we'll just remove the `--generate-notes` for now, and then revert this commit as soon as the release is out, as releases should be happening every month and, ideally, we won't reach this situation never ever again.

Fixes: #9064 - part V


fidencio:
We don't need to run CI on this one, as CI won't cover anything related to this PR.

stevenhorsman (approved):
LGTM. Thanks

Issue #9225, opened 2024-03-06
Optimize guest pull feature 


ChengyuZhu6:
~~I’m basing this on the assumption that `guest-pull` feature will be implemented by PR #8484.~~ I will focus on completing the following tasks:

- [x] Add documents for kata guest image management design, covering both design and usage. (Appreciate @wainersm’s suggestion )
- [x] Improve the memory efficiency of image service in the guest (Appreciate @Xynnn007’s suggestion )
- [ ] Support different pause image in the guest for `guest-pull` (Appreciate @littlejawa’s suggestion )
- [x] Add `guest-pull` to the list of agent features in `announce()` (Appreciate @jodh-intel’s suggestion )

Please share any relevant details or issues that you may have.

BTW, that's not the part of tasks for merging code from CCv0 into main.

cc @wainersm @Xynnn007 @littlejawa @jodh-intel @fidencio @stevenhorsman

PR #9325, opened 2024-03-21
agent:image: Refactor code to improve memory efficiency of image service

ChengyuZhu6:
Currently, `.lock().await.clone()` results in `Option<ImageService>` being duplicated in memory with each call to `singleton()`. Consequently, if kata-agent receives numerous image pulling requests simultaneously, it will lead to the allocation of multiple `Option<ImageService>` instances in memory, thereby consuming additional memory resources.

In image.rs, we introduce two public functions: `merge_bundle_oci()` and `init_image_service()`. These functions will encapsulate the operations on `IMAGE_SERVICE`, ensuring that its internal details remain hidden from external modules such as `rpc.rs`.

Fixes: #9225 -- part II

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -2282,9 +2282,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.151"
+version = "0.2.153"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "302d7ab3130588088d277783b1e2d2e10c9e9e4a16dd9050e6ec93fb3e7048f4"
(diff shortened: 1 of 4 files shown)


ChengyuZhu6:
cc @Xynnn007

Xynnn007 (approved):
Thanks. lgtm

stevenhorsman (approved):
I'm not a rust expert, but my understand of this PR is that rather than have the lock().await.clone() on the singleton method which as you say will result in memory duplication, you've effectively moving the clone and lock down only when it's needed in functions, which should reduce the resources. Assuming that I've not missed the point - this makes sense to me. Thanks!

arronwy (approved):
Thanks @ChengyuZhu6 LGTM!

Issue #9327, opened 2024-03-21
Default max CPU handling assesses max CPUs while still in upper-layer runtime's cgroup


egernst:
# Description of problem

I observed unexpected behavior when moving the upper-layer runtime, containerd in my case, into a cgroup that is bound by a cpuset. When doing so, the maximum pod I can launch with kata-containers is dictated by the size of that containerd specific cgroup, rather than the separate pool for kubepods, which could be (and should be) much larger.

As an example, say I have containerd in a cgroup with cpuset[0,1] as part of system reserved, and intend to run Kubernetes pods on the rest of the host, say [2-63].
 
# Expected result

I would expect I can run a pod with req=limits=4CPU.
 
# Actual result

We'll receive an error stating that MaxCPU is 2, and that 4 exceeds 2.

# Analysis

The issue we have is that we are setting up the runtime config earlier in the sandbox creation, and are validating that the maximum CPU provided in the configuration file "makes sense" based on the system. At this point in time, however, [1] will show what is available when we are in context of containerd cgroup and haven't yet moved to the proper resource controller space - this occurs later @ [2]. 

From a first glance, I think we could do one of a couple of things here:
 1. If a user specifies a value in the toml for max CPUs, just trust it without validating. By default we don't set anything, and in that case the existing logic is not great, but at least the administrator of Kata has an option to make it "do the right thing."
 2. Set the default max CPUs only after we have called 'setupResourceController'

[update]
3. Check number of logical CPUs on the host system and don't rely on goruntime's method of what's visible to the process.

 
I think (3) is best - Hopefully folks have even easier ideas to suggest.

[1] - https://github.pie.apple.com/kata-containers/kata-containers/blob/apple-main/src/runtime/pkg/katautils/config.go#L351

[2] - https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/api.go#L87

egernst:
The max handling is a bit of a mess to untangle given that we are checking at LoadConfiguration time.  Rather than reorder or change too much, I think it makes sense to do what we have in the confirmation toml comments: Validate only against the number of actual CPUs on the host system, not what is visible to the particular process via go runtime.  I'm working to just update checking procfs to see number of logical CPUs instead of calling the go runtime to determine.

PR #9330, opened 2024-03-21
vc: adjust default and max CPUs after we move shim

egernst:
We are setting up the runtime config earlier in configuration load and are validating the maximum CPU provided in the configuration file "makes sense" based on how many CPUs are available to the shim process. At this point in time, however, the shim is bound by how many CPUs its parent have.

In the event that the upper layer runtime (say containerd) is running within a constrained cpuset cgroup, this check will return however many CPUs the upper layer has available. Once the actual pod/sandbox is started, however, we'll no longer be in this cgroup and may have a different limitation (or no limitation).

Based on this unexpected behavior, let's make sure that we adjust the default and maximum number of vCPUs *after* we move into our own cgroup hierarchy.

Fixes: #9327

### DIFF
--- a/src/runtime/virtcontainers/api.go
+++ b/src/runtime/virtcontainers/api.go
@@ -88,6 +88,14 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		return nil, err
 	}
 
+	// Now that we're in our 'final host cgroup' - let's make sure that the maximum number of
+	// CPUs reflects what is available to the shim.
+	currentMax := s.config.HypervisorConfig.DefaultMaxVCPUs
+	currentDefaultCPU := s.config.HypervisorConfig.NumVCPUs
+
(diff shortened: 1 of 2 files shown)


src/runtime/virtcontainers/api.go
```diff
@@ -88,6 +88,14 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		return nil, err
 	}
 
+	// Now that we're in our 'final host cgroup' - let's make sure that the maximum number of
+	// CPUs reflects what is available to the shim.
+	currentMax := s.config.HypervisorConfig.DefaultMaxVCPUs
+	currentDefaultCPU := s.config.HypervisorConfig.NumVCPUs
```
egernst, line 94: this'll be broken b/c the config will have applied a default already at LoadConfiguration time.  Need to revisit.  maybe there's not a benefit of checking max at all.

PR #9331, opened 2024-03-22
katautils: check number of cores on the system intead of go runtime

egernst:
We used go runtime's "NumCPUs()" in the past, which will give the value available to the runtime, which may be a subset of physical cores if the shim is started from within a cpuset.

Since the shim will be moved into its own cgroup that may have a different cpuset, let's stick with checking physical cores, since that's what we have documented.

In the event we fail to read /proc/cpuinfo, let's use the goruntime.

Fixes: #9327


src/runtime/pkg/katautils/config.go
```diff
@@ -386,12 +386,32 @@ func (h hypervisor) GetEntropySource() string {
 	return h.EntropySource
 }
 
+var procCPUInfo = "/proc/cpuinfo"
+
+func getHostCPUs() uint32 {
+	cpuInfo, err := os.ReadFile(procCPUInfo)
+	if err != nil {
+		kataUtilsLogger.Warn("unable to read /proc/cpuinfo to determine cpu count - using go runtime value instead")
+		return uint32(goruntime.NumCPU())
```
gkurz, line 395: Not being able to read `/proc/cpuinfo` looks like the environment where the shim is running is in very bad shape, doesn't it ? Also it feels strange to fall back on the buggy value that this PR is going away from.

What about collecting `/proc/cpuinfo` at shim startup and either error out if it fails or cache it for `getHostCPus` if it succeeds ?

src/runtime/pkg/katautils/config_test.go
```diff
@@ -1802,3 +1802,79 @@ func TestUpdateRuntimeConfigHypervisor(t *testing.T) {
...
+		if err := file.Close(); err != nil {
+			t.Fatalf("Error closing temp file: %v", err)
+		}
+
+		parsed := getHostCPUs()
+
+		if parsed != tc.expected {
+			t.Errorf("Expected number of cores: %d, got: %d", tc.expected, parsed)
+		}
+	}
```
gkurz, line 1879: Maybe add a test for the case where `/proc/cpuinfo` cannot be read ?

src/runtime/pkg/katautils/config.go
```diff
@@ -386,12 +386,32 @@ func (h hypervisor) GetEntropySource() string {
 	return h.EntropySource
 }
 
+var procCPUInfo = "/proc/cpuinfo"
+
+func getHostCPUs() uint32 {
```
ChengyuZhu6, line 391: Would it be better to return the value as the lesser of the two numbers: one from `/proc/cpuinfo` and the other from `NumCPU()`?

egernst: No, I don’t think that’s what we want. Pls read commit message which explains rationale for current pattern.

src/runtime/pkg/katautils/config.go
egernst: Well, in case you aren’t in linux, this wouldn’t exist. We can certainly fail hard here; would need to test the underlying error flow and messages passed to logs in the system. 

cacheing is an optimization to not read for next container, or? AFAICT we’d call it for sandbox and each container add (so maybe two or three times)?

gkurz: > Well, in case you aren’t in linux, this wouldn’t exist. We can certainly fail hard here; would need to test the underlying error flow and messages passed to logs in the system.
> 

Ahh yes, I tend to forget. Sorry for that.

Then all non-linux OSes that officially support kata will then get this warning systematically for no good reason.

Either they don't have a way to start the shim in a restricted cpuset and they can go on with the current implementation, or they likely(?) provide a way to get the number of physical cores and we should use that rather that providing a potentially wrong value IMHO.

> cacheing is an optimization to not read for next container, or? AFAICT we’d call it for sandbox and each container add (so maybe two or three times)?

Not an optimization, it was just a trick to handle the lack of `/proc/cpuinfo` outside of `getHostCPUs()` and avoid all the noise of returning an error if fallback is dropped.

egernst: Thinking through this a bit more:
 - we don't really have an error path in this part of the code, and I'm not comfortable with not handling the error if we can't read the filesystem for "whatever reason"; I was  looking for minimal changes that'll fix the current bug. By falling back, we can continue while logging the warning.
 -  For "non linux", at this point upstream doesn't have a non-linux runtime, just build and unit tests to help ensure that we don't leak Linux'isms that'll break Darwin builds. In the event we add proper Darwin support (or other nix), we'll need them to have an architecture specific check here anyway. Imo we can leave that to the person bringing up the new OS at that time?

src/runtime/pkg/katautils/config_test.go
egernst: Good suggestion - done!

src/runtime/pkg/katautils/config.go
egernst: > Either they don't have a way to start the shim in a restricted cpuset and they can go on with the current implementation, or they likely(?) provide a way to get the number of physical cores and we should use that rather that providing a potentially wrong value IMHO.

Exactly - they'll have their own implementation of this function at the time we _add_ that new OS. For now, this fallback is more of handling the local error that I don't expect we'll run into.

fidencio (approved):
@egernst, welcome back! :-)
This seems reasonable to me, assuming all the tests pass.

I've restarted the failing ones.

ChengyuZhu6 (approved):
LGTM thanks @egernst !

Issue #9363, opened 2024-03-27
tests: k8s: incomplete memory clean-up before starting the next test


danmihai1:
The Kata CI tests for k8s are using relatively small k8s nodes. Therefore, it's important to release the memory used by the current test case before starting the next test case. Each test case waits for its pods to be deleted before moving on to the next test case. However, every once in a while we are seeing test pods that either don't start or get evicted because k8s detects memory pressure. When that happens, we end up with CI failures similar to:

```
[run_kubernetes_tests.sh:147] INFO: Executing k8s-credentials-secrets.bats
...
     `kubectl exec $second_pod_name -- sh -c "$second_cmd" | grep -w "SECRET_USERNAME"' failed
...
   Status:              Failed
   Reason:              Evicted
   Message:             The node was low on resource: memory. Threshold quantity: 750Mi, available: 737488Ki.
```

In my test cluster, I see that every once in a while the Shim + VMM + probably the Guest VM sometimes stick around for several seconds *after* k8s reported the pod as deleted. Also, ```kubectl top nodes``` reports ~80% memory usage after deleting a test pod, but ~40% memory usage a few seconds later. CI should wait for the memory usage to go back under ~40% (or whatever other acceptable percentage value) before starting the next test, to increase our changes of avoiding spurious test failures.

danmihai1:
It appears that the information returned by "kubectl top nodes" is sometimes not up to date, so it's difficult to make CI decisions based on that information.

Giving up on this - at least for now.

PR #9366, opened 2024-03-27
tests: k8s: wait for memory clean-up

danmihai1:
Wait for k8s to reclaim the current test pod's memory before starting the next test pod.

Sometimes it takes several seconds between k8s reporting that a test pod has been deleted and k8s reporting that the pod's memory has been reclaimed. Without waiting for the memory to be reclaimed we can hit spurious CI failures similar to:

[run_kubernetes_tests.sh:147] INFO: Executing k8s-credentials-secrets.bats

...

`kubectl exec $second_pod_name -- sh -c "$second_cmd" | grep -w "SECRET_USERNAME"' failed

...

Status:  Failed
Reason:  Evicted
Message: The node was low on resource: memory. Threshold quantity: 750Mi, available: 737488Ki.

Fixes: #9363


Redent0r (approved):

danmihai1:
It appears that the information returned by "kubectl top nodes" is sometimes not up to date, so it's difficult to make CI decisions based on that information.

Giving up on this - at least for now.

PR #9367, opened 2024-03-28
runtime: remove stream copy infinite loop

danmihai1:
This reverts commit 1c5693be86be4dcfee83d2d5ff2ec68dcfd18df4.

Avoid apparent infinite loop when ReadStreamRequest is blocked by policy - for some of the pods.

When running the k8s-limit-range.bats test with Policy enabled, the Shim + VMM never get terminated on my cluster. Not sure why the sandbox clean-up works better for other tests, but the k8s-limit-range test pod gets stuck in an infinite loop:

```
stdout io stream copy error happens: error = %wrpc error: code =
PermissionDenied desc = \"ReadStreamRequest is blocked by policy

...

policy check: ReadStreamRequest

...

stdout io stream copy error happens: error = %wrpc error: code =
PermissionDenied desc = \"ReadStreamRequest is blocked by policy

...

policy check: ReadStreamRequest

...
```

Fixes: #9380

### DIFF
--- a/src/runtime/pkg/containerd-shim-v2/stream.go
+++ b/src/runtime/pkg/containerd-shim-v2/stream.go
@@ -10,9 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
-	"strings"
 	"sync"
-	"time"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @danmihai1!

ChengyuZhu6 (approved):
LGTM thanks @danmihai1 !

PR #9369, opened 2024-03-28
agent:image: Support different pause image in the guest for guest pull

ChengyuZhu6:
Support different pause images in the guest for guest-pull, such as k8s pause image (registry.k8s.io/pause) and openshift pause image (quay.io/bpradipt/okd-pause).

The primary distinction between the k8s pause image and the OKD pause image lies in the arguments specified in the `config.json` file:

For the k8s pause image, the arguments are set as follows:
```json
"args": [
    "/pause"
],
```

Conversely, for the OKD pause image, the arguments are different:
```json
"args": [
    "/usr/bin/pod"
],
```
Fixes: #9225 -- part III

### DIFF
--- a/src/agent/src/image.rs
+++ b/src/agent/src/image.rs
@@ -41,6 +41,15 @@ fn sl() -> slog::Logger {
     slog_scope::logger().new(o!("subsystem" => "image"))
 }
 
+// Function to copy a file if it does not exist at the destination
+fn copy_if_not_exists(src: &Path, dst: &Path) -> Result<()> {
+    if let Some(dst_dir) = dst.parent() {
+        fs::create_dir_all(dst_dir)?;
+    }
+    fs::copy(src, dst)?;
+    Ok(())
+}
+
 pub struct ImageService {
     image_client: ImageClient,
     images: HashMap<String, String>,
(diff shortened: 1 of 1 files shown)


ChengyuZhu6:
@littlejawa Could you please have a test using the okd-pause-image? Unfortunately, I don’t have the setup to test okd-pause-image myself. However, I’ve verified the code with the k8s pause image (registry.k8s.io/pause:3.9), and it works correctly.

src/agent/src/image.rs
```diff
@@ -66,26 +75,41 @@ impl ImageService {
         if !guest_pause_bundle.exists() {
             bail!("Pause image not present in rootfs");
         }
+        let guest_pause_config = guest_pause_bundle.join(CONFIG_JSON);
```
wainersm, line 78: Hi @ChengyuZhu6 ! Not a Rust expert speaking but... I noticed that some path operations in `image.rs` have used [safe_path](https://docs.rs/safe-path/latest/safe_path/index.html)::[scoped_join](https://docs.rs/safe-path/latest/safe_path/fn.scoped_join.html#) instead of `join`. Perhaps we should use it in this function as well to make it safer and consistent?

src/agent/src/image.rs
```diff
@@ -66,26 +75,41 @@ impl ImageService {
...
 
         info!(sl(), "use guest pause image cid {:?}", cid);
```
wainersm, line 80: One suggestion is to move the `let image_oci = oci::Spec::load( ... )`  call and any parsing of `image_oci` to this line. Because if there is something wrong then it doesn't even need to create the pause_bundle and Pause_rootfs directory;

src/agent/src/image.rs
```diff
@@ -66,26 +75,41 @@ impl ImageService {
...
-            Ok(())
-        };
-        copy_if_not_exists(
-            &guest_pause_bundle.join(CONFIG_JSON),
-            &pause_bundle.join(CONFIG_JSON),
-        )?;
+        info!(sl(), "pause_rootfs {:?}", pause_rootfs);
+        let image_oci = oci::Spec::load(guest_pause_config.to_str().ok_or_else(|| {
+            anyhow!(
+                "Invalid container image OCI config path {:?}",
```
wainersm, line 87: I suggest re-phrase this message like: "Failed to load the guest pause image config from {:?}"

src/agent/src/image.rs
```diff
@@ -66,26 +75,41 @@ impl ImageService {
...
+                guest_pause_config
+            )
+        })?)
+        .context("load image config file")?;
+
+        let image_oci_process = image_oci.process.ok_or_else(|| {
+            anyhow!("The image config file does not contain a process specification.")
```
wainersm, line 94: And here: "The guest pause image config does not contain a process specification. Please check the pause image" to be consistent with the message below when `args.len()` checker fails.

src/agent/src/image.rs
```diff
@@ -66,26 +75,41 @@ impl ImageService {
...
+        info!(
+            sl(),
+            "pause image oci process {:?}",
+            image_oci_process.clone()
+        );
+
+        // Ensure that the args vector is not empty before accessing its elements.
+        let args = image_oci_process.args;
+        // Check the number of arguments.
+        if args.len() != 1 {
```
wainersm, line 105: For the `registry.k8s.io/pause` and `quay.io/bpradipt/okd-pause` images there is only one `args` (the pause executable file). However, is it always the case? It seems not unlikely to pass additional arguments to the pause executable so that `args.len() >= 1`.

wainersm:
Hi @ChengyuZhu6 , I'm not exactly an expert on the subject but I think I understood this change, so I left just a couple of comments but some might not make sense (please ignore if the case).

src/agent/src/image.rs
ChengyuZhu6: Fixed.

src/agent/src/image.rs
ChengyuZhu6: Fixed.

src/agent/src/image.rs
ChengyuZhu6: Good point. I have fixed it.

src/agent/src/image.rs
ChengyuZhu6: Done.

ChengyuZhu6:
> Hi @ChengyuZhu6 , I'm not exactly an expert on the subject but I think I understood this change, so I left just a couple of comments but some might not make sense (please ignore if the case).

@wainersm thanks for your comments! I've actions your comments. Please double check when you get a chance.

fidencio:
@ChengyuZhu6, I was able to verify your patches on a single node OCP cluster, that used the OCP's pause image by default, and it works as expected.  Nice work!

src/agent/src/image.rs
```diff
@@ -66,28 +75,49 @@ impl ImageService {
...
+        let container_bundle = scoped_join(CONTAINER_BASE, cid)?;
+        fs::create_dir_all(&container_bundle)?;
+        let pause_bundle = scoped_join(&container_bundle, target_subpath)?;
+        fs::create_dir_all(&pause_bundle)?;
+        let pause_rootfs = scoped_join(&pause_bundle, "rootfs")?;
+        fs::create_dir_all(&pause_rootfs)?;
+        info!(sl(), "pause_rootfs {:?}", pause_rootfs);
+
+        copy_if_not_exists(&guest_pause_config, &pause_bundle.join(CONFIG_JSON))?;
+        for pause_arg in args {
```
wainersm, line 114: Hi @ChengyuZhu6 !

Thanks for addressing my comments but I'm afraid I confused you, sorry.

It should not raise if `len(args)` > 1 as it you did before, however, we still want to copy only `args[0]` (which is the pause binary file).

Let's use a hypothetical `config.json` below, where `len(args) > 1`:
```
{
	"ociVersion": "1.0.0",
	"process": {
		"terminal": true,
		"user": {
			"uid": 65535,
			"gid": 65535
		},
		"args": [
			"/usr/bin/pod",
                        "arg1",
                        "arg2"
		],
		"env": [
			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
			"TERM=xterm"
		],
```

`arg1` and `arg2` are really arguments to `/usr/bin/pod` binary, it will fail if it tries to copy these arguments :D

ChengyuZhu6: Oh, apologies for the misunderstanding. Thanks for bringing this.

ChengyuZhu6:
@wainersm @fidencio Would you mind providing a re-review on this PR ?

littlejawa (approved):
lgtm
Thanks @ChengyuZhu6

fidencio:
LGTM, I will wait for @wainersm approval!

fidencio (approved):

Issue #6070, opened 2023-01-13
runk: print log to stdout


liubin:
Seems runk can only write logs to files if the `--log` option is specified, otherwise the log will be discarded.
https://github.com/kata-containers/kata-containers/blob/2b4b825228716c08e95079b831ad66341b2b9b72/src/tools/runk/src/main.rs#L102-L117

My advice is to
- add an option named `log-to-stdout`, if this option is set, then print the logs to stdout.
- print all logs to stdout if the `--log` file is not specified, but set the default log level to a high value, for example, ERROR, that will reduce logs printed by rustjail, which mainly uses INFO log level.
- print logs to stdout if `--debug` is set

Because runk is using https://docs.rs/liboci-cli/0.0.4/liboci_cli/struct.GlobalOpts.html, it's hard to modify the command line options, so maybe the 2nd/3rd are the simplest ones to implement.

/cc @ManaSugi @cyyzero

PR #9370, opened 2024-03-28
runk: print log to stdout

mayamrinal:
- Print log to stdout when log file is not provided.
	- Log level: Error
- Print log to stdout when --debug option is used.
	- Log level: Debug

Fixes: #6070


mayamrinal:
@ManaSugi can you please review when you have time?

PR #8870, opened 2024-01-22
port attestation agent from CCv0 branch to main branch

LindaYu17:
build attestation-agent, confidential-data-hub, api-server-rest for rootfs image.
and launch them by kata-agent in guest

Issue #9379, opened 2024-03-28
tests: strengthen kbs_k8s_delete() for baremetal


wainersm:
In PR  https://github.com/kata-containers/kata-containers/pull/9065 we introduced the code to install and delete the CoCo KBS for the attestation tests. Initially those tests were meant for running on AKS (non-TEE) which is an ephemeral environment, so that the  [kbs_k8s_delete()](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/confidential_kbs.sh#L162) function wasn't implemented to run within baremetal.

Now that we are planning to expand attestation tests for TEE environments (https://github.com/kata-containers/kata-containers/issues/9376, https://github.com/kata-containers/kata-containers/issues/9378), we will need to strengthen that function to baremetal requirements:
 * should wait all resources be effectively deleted
 * should not fail if resources weren't created

Issue #9378, opened 2024-03-28
CC: Enable guest-components tests for AMD SNP


wainersm:
### Context

On [port attestation agent from CCv0 branch to main branch](https://github.com/kata-containers/kata-containers/pull/8870) we introduce the last bits of attestation on non-TEE environments, including tests. The follow on should be to add support on TEE environments (SEV, SNP, TDX).

### Details

Testing for AMD SNP should be pretty similar to non-TEE. It requires same CoCo's KBS and `cc_kbc` KBC.

Likewise [SEV](https://github.com/kata-containers/kata-containers/issues/9376) it will be needed some features not yet implemented on runtime. 

This issue tracks the work required to enable tests on CI for attestation.

Depends on:

- [x] [port attestation agent from CCv0 branch to main branch](https://github.com/kata-containers/kata-containers/pull/8870)
- [ ] [Fix SNP extended attestation with latest AMD kernel/qemu patches](https://github.com/kata-containers/kata-containers/issues/9101)
- [x] [runtime: enable kernel-hashes for SNP confidential container](https://github.com/kata-containers/kata-containers/pull/9151)
- [x] [tests: strengthen kbs_k8s_delete() for baremetal](https://github.com/kata-containers/kata-containers/issues/9379) 

### Work breakdown

- [x] [tests: Support for kbs setup on kcli](https://github.com/kata-containers/kata-containers/pull/9273) - This same KBS setup should work on SNP with kubeadm 
- [x] Change the [run-k8s-tests-on-snp](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-kata-coco-tests.yaml#L135) workflow to install KBS
- [x] Enable the attestation test to run on jobs with `qemu-snp` hypervisor.

Cc @ryansavino @fitzthum @AdithyaKrishnan

wainersm:
https://github.com/kata-containers/kata-containers/pull/9273 was merged but it wasn't tested with kubeadm. I really believe it will work out; we will see when try to enable guest-components tests on SNP.

ariel-adam:
@wainersm should we leave this in 0.9.0 or move it to 0.10.0?

wainersm:
> @wainersm should we leave this in 0.9.0 or move it to 0.10.0?

@ariel-adam the runtime is still not ready for SNP attestation. I will let @ryansavino @fitzthum answers if they want it for 0.9 or 0.10.

[...]

Issue #9376, opened 2024-03-28
CC: Enable guest-components tests for AMD SEV


wainersm:
### Context

On [port attestation agent from CCv0 branch to main branch](https://github.com/kata-containers/kata-containers/pull/8870) we introduce the last bits of attestation on non-TEE environments, including tests. The follow on should be to add support on TEE environments (SEV, SNP, TDX).

### Details

AMD SEV is a particular case because it should still be implemented some features on runtime, for example, [sev: add sev support to kata-shim (main)](https://github.com/kata-containers/kata-containers/issues/6571). Apart from that, testing will be slightly different because on SEV it uses [simple-kbs](https://github.com/confidential-containers/simple-kbs) (and the `online_sev_kbc` KBC).

This issue tracks the work required to enable tests on CI for attestation.

Depends on:
- [x] [port attestation agent from CCv0 branch to main branch](https://github.com/kata-containers/kata-containers/pull/8870)
- [ ] [sev: add sev support to kata-shim (main)](https://github.com/kata-containers/kata-containers/issues/6571) 

### Work breakdown

- [ ] Add support for simple-kbs in [confidential_kbs.sh](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/confidential_kbs.sh) lib
- [ ] Change the [run-k8s-tests-on-sev](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-kata-coco-tests.yaml#L80) workflow to install simple-kbs
- [ ] Change the attestation test to proper configure simple-kbs and use `offline_sev_kbc`

### Additional resources

[1] [simple-kbs utility functions on CCv0](https://github.com/kata-containers/tests/blob/CCv0/integration/confidential/lib.sh#L360)

Cc @ryansavino @fitzthum @AdithyaKrishnan

stevenhorsman:
SEV has been removed now

PR #9382, opened 2024-03-29
Merge to main: supporting pull encrypted images

### LINKED ISSUE #8111: CC: Pull image and unpack of encrypted image on guest
This feature builds upon the initial basic image pull work in https://github.com/kata-containers/kata-containers/issues/8103 as one of the possible next steps to extend it to support more of the current `CCv0` feature set.

The scope of this is to add the ability to pull an encrypted container image on the guest and using a provided decryption key decrypt it and unpack it, so we can run an pod from it.

Initially I/we'd hoped to be able to cherry-pick over commits from `CCv0`, but given how much iteration there has been we're going to need to copy over chunks of the code and tests, improving as we go. Ideally we'll add `co-authored` tags to reflect the original authors though.

## Things to note/include with this work
- With this feature (unless already added by something like signature validation if that is picked up before this) we'll need to make the attestation-agent available to image-rs, but I'm not sure we need the confidential-data-hub, or api-server-rest at this point. This involves the follow code changes (plus others I may have missed):
  -  Add attestation-agent to the [versions.yaml](https://github.com/kata-containers/kata-containers/blob/1ad87faaf4d3068569d99c270df640e41465c410/versions.yaml#L200-L203)
  - `rootfs.sh` changes to add the `AA_KBC` parameter (assuming we want to keep this?) and then to [install the attestation-agent](https://github.com/kata-containers/kata-containers/blob/1ad87faaf4d3068569d99c270df640e41465c410/tools/osbuilder/rootfs-builder/rootfs.sh#L683-L697) into the rootfs. See questions
  - code changes in `main.rs` related to [initialising the attestation-agent](https://github.com/kata-containers/kata-containers/blob/1ad87faaf4d3068569d99c270df640e41465c410/src/agent/src/main.rs#L389). This should be toggled on the same feature we add for image_pull support in #8103
  - Adding support for `aa_kbc_params` in agent config: https://github.com/kata-containers/kata-containers/pull/5533/commits/112a3d2baef3709a50655b23d6bd80ef121d50c0
  - The keyprovider pieces of [`get_security_config` in image_rpc.rs](https://github.com/kata-containers/kata-containers/blob/1ad87faaf4d3068569d99c270df640e41465c410/src/agent/src/image_rpc.rs#L143-L152)
- New tests
  - There are three new test cases I'd suggest, that are detailed in the acceptance criteria. The code for this code be based on the [existing `CCv0` tests](https://github.com/kata-containers/tests/blob/CCv0/integration/kubernetes/confidential/agent_image_encrypted.bats), but we want to work to make them easier to apply to different KBCs/TEE by extracting out the KBC/TEE logic into library functions, so the test cases are cleaner, similar to the approach done in https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-confidential.bats rather than a pure cherry-pick
  - For the offline_fs_kbc scenario which is used in non-TEE hardware and SE, we need to have a way to modify the guest image to add the `/etc/aa-offline_fs_kbc-keys.json` file that contains the decryption key. This is [currently done](https://github.com/kata-containers/tests/blob/6cf430642335c010014b5d9355a861328c4213ed/integration/kubernetes/confidential/lib.sh#L129) using the [`cp_to_guest_img`](https://github.com/kata-containers/tests/blob/6cf430642335c010014b5d9355a861328c4213ed/lib/common.bash#L295), which ideally we said we'd prefer not to move across into `main`, but rather use policies to pass through this sort of configuration, but I don't think that is available, so I guess we just need to temporarily port over the cp_to_guest_img and associated lib functions over, if it hasn't already been moved over as part of another issue (like #8101)

## Questions
- How do we control and create the attestation-agent integration in the rootfs? If it is small, should we always include it, or do we need separate versions of the image with AA installed? How about support for mutliple KBCs? Currently I don't think there are any plans to support a single attestation-agent binary with multiple KBC modules built, so does this mean that we need different images build for each KBC we support, which will make kata-deploy larger and potential more painful with extra runtime classes? @Xynnn007 @fidencio 
- Does decryption need the CDH & api-server-rest? or are they still just needed for sealed-secrets? @arronwy 
- In `CCv0` we only have integration tests for this feature on Kubernetes, so do we want and can we test this on `crictl` as well?
- What container image do we want to use? In `CCv0` we use, `[docker.io/katadocker/ccv0-ssh](https://hub.docker.com/r/katadocker/ccv0-ssh)` built by https://github.com/confidential-containers/confidential-containers/tree/main/demos/ssh-demo, but should we take ownership of the image and build it ourselves like we do with the unencrypted test image: https://github.com/kata-containers/kata-containers/tree/main/tests/integration/kubernetes/runtimeclass_workloads/confidential? 

## Acceptance Criteria

For each supported platform (different TEEs):

_**Scenario: Pull encrypted image on guest works**_
**Given** I have a version of kata deployed with a guest image that has an agent with `guest_pull` feature enabled and nydus-snapshotter installed and configured for [guest-pulling](https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/config-coco-guest-pulling.toml)
  **And** A public encrypted container image *i* with a decryption key *k* that is configured in a KBS, so that image-rs on the guest can connect to it
**When** I try and create a pod from *i*
**Then** The pod is successfully created and runs

_**Scenario: Cannot pull encrypted image with no decryption key**_
**Given** I have a version of kata deployed with a guest image that has an agent with `guest_pull` feature enabled and nydus-snapshotter installed and configured for [guest-pulling](https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/config-coco-guest-pulling.toml)
  **And** A public encrypted container image *i* with a decryption key *k* that is **not** configured in a KBS, so that image-rs on the guest can connect to it
**When** I try and create a pod from *i*
**Then** The pod is not created an errors (with a helpful message hopefully)

_**Scenario: Cannot pull encrypted image with wrong decryption key**_
**Given** I have a version of kata deployed with a guest image that has an agent with `guest_pull` feature enabled and nydus-snapshotter installed and configured for [guest-pulling](https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/config-coco-guest-pulling.toml)
  **And** A public encrypted container image *i* with a decryption key *k* and a different key *k'* that is configured in a KBS, so that image-rs on the guest can connect to it
**When** I try and create a pod from *i*
**Then** The pod is not created an errors (with a helpful message hopefully)

Xynnn007:
This PR mainly fixes #8111. Le's temporarily make it draft as it depends the following PRs
- #8870.  Which brings CDH, the component inside guest to decrypt image layer
- https://github.com/confidential-containers/guest-components/pull/524 and https://github.com/confidential-containers/guest-components/pull/523: let kata-agent get rid of reading aa_kbc_params from kernel cmdline
- https://github.com/confidential-containers/guest-components/pull/522 makes kata-agent built-able

Much thanks to @ChengyuZhu6 .

hi @wainersm , I use the following yaml on tdx and it succeed
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox
  annotations:
    io.containerd.cri.runtime-handler: kata-qemu-tdx
spec:
  runtimeClassName: kata-qemu-tdx
  containers:
  - name: busybox
    image:  ghcr.io/confidential-containers/test-container:encrypted
```

We should set up a KBS resource recorded [in here](https://github.com/confidential-containers/confidential-containers/blob/main/demos/ssh-demo/aa-offline_fs_kbc-keys.json).

Also, kernel cmdline should add an item `agent.aa_kbc_params=cc_kbc::http://<kbs-addr-and-port>`

I must miss something so lmk : )

BTW, the dependency I used `cargo update` because there is conflict when I enable feature `image-rs/kata-cc-rustls-tls`. So the changed lines are huge. Hi @ChengyuZhu6 please help to check if the feature flag is as expected (actually pull encrypted image and normal image in guest we just need change the feature of image-rs. I enable the "bigger" feature by default)

cc @fitzthum  @stevenhorsman

### DIFF
--- a/src/agent/Makefile
+++ b/src/agent/Makefile
@@ -159,7 +159,7 @@ vendor:
 
 #TARGET test: run cargo tests
 test: $(GENERATED_FILES)
-	@cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
+	@RUST_LIB_BACKTRACE=0 cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
 
 ##TARGET check: run test
 check: $(GENERATED_FILES) standard_rust_check
--- a/src/agent/src/main.rs
+++ b/src/agent/src/main.rs
@@ -110,6 +110,9 @@ const CDH_SOCKET_URI: &str = concatcp!(UNIX_SOCKET_PREFIX, CDH_SOCKET);
 
 const API_SERVER_PATH: &str = "/usr/local/bin/api-server-rest";
 
+/// Path of ocicrypt config file. This is used by image-rs when decrypting image.
+const OCICRYPT_CONFIG_PATH: &str = "/tmp/ocicrypt_config.json";
+
 const DEFAULT_LAUNCH_PROCESS_TIMEOUT: i32 = 6;
 
 lazy_static! {
@@ -478,10 +481,22 @@ fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<
         return Ok(None);
     }
 
+    let ocicrypt_config = serde_json::json!({
+        "key-providers": {
+            "attestation-agent":{
+                "ttrpc":CDH_SOCKET_URI
+            }
+        }
+    });
+
+    fs::write(OCICRYPT_CONFIG_PATH, ocicrypt_config.to_string().as_bytes())?;
+    env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
+
     debug!(
         logger,
         "spawning confidential-data-hub process {}", CDH_PATH
     );
+
     launch_process(
         logger,
         CDH_PATH,
--- a/tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
@@ -1,0 +1,138 @@
+#!/usr/bin/env bats
+# Copyright (c) 2024 IBM Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/lib.sh"
+load "${BATS_TEST_DIRNAME}/confidential_common.sh"
+load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
+
+export KBS="${KBS:-false}"
+
+setup() {
+    if is_confidential_hardware; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    fi
+
+    if ! is_confidential_runtime_class; then
+        skip "Test not supported for ${KATA_HYPERVISOR}."
+    fi
+
+    if [ "${KBS}" = "false" ]; then
+        skip "Test skipped as KBS not setup"
+    fi
+
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
+    setup_common
+    ENCRYPTED_IMAGE="${ENCRYPTED_IMAGE:-ghcr.io/confidential-containers/test-container:multi-arch-encrypted}"
+    DECRYPTION_KEY="${DECRYPTION_KEY:-HUlOu8NWz8si11OZUzUJMnjiq/iZyHBJZMSD3BaqgMc=}"
+    DECRYPTION_KEY_ID="${DECRYPTION_KEY_ID:-ssh-demo}"
+}
+
+function setup_kbs_decryption_key() {
+    decryption_key=$1
+    decryption_key_id=$2
+
+    if ! is_confidential_hardware; then
+        kbs_set_allow_all_resources
+    fi
+
+    # Note: the ssh-demo
+    kbs_set_resource_base64 "default" "key" "${decryption_key_id}" "${decryption_key}"
+}
+
+function create_pod_yaml_with_encrypted_image() {
+    image=$1
+
+    # Note: this is not local as we use it in the caller test
+    kata_pod_with_encrypted_image="$(new_pod_config "$image" "kata-${KATA_HYPERVISOR}")"
+    set_node "${kata_pod_with_encrypted_image}" "$node"
+    set_container_command "${kata_pod_with_encrypted_image}" "0" "sleep" "30"
+
+    local CC_KBS_ADDR
+    export CC_KBS_ADDR=$(kbs_k8s_svc_http_addr)
+    kernel_params_annotation="io.katacontainers.config.hypervisor.kernel_params"
+    kernel_params_value+=" agent.guest_components_procs=confidential-data-hub"
+    kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"
+
+    set_metadata_annotation "${kata_pod_with_encrypted_image}" \
+        "${kernel_params_annotation}" \
+        "${kernel_params_value}"
+
+    # Set annotation to pull image in guest
+    set_metadata_annotation "${kata_pod_with_encrypted_image}" \
+        "io.containerd.cri.runtime-handler" \
+        "kata-${KATA_HYPERVISOR}"
+
+    add_allow_all_policy_to_yaml "${kata_pod_with_encrypted_image}"
+}
+
+@test "Test that creating a container from an encrypted image, with no decryption key fails" {
+
+    # TODO - there is now delete KBS resource to ensure there is no key, so we need to keep
+    # this test running first to ensure that the KBS doesn't have the resource. An alternative
+    # is to run kbs_set_deny_all_resources, but we don't have a way to reset to the default
+    # policy, so for TEE tests we'd stay remaining with reject all, which could cause other
+    # subsequent tests to fail
+
+    create_pod_yaml_with_encrypted_image "${ENCRYPTED_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_encrypted_image}: $(cat ${kata_pod_with_encrypted_image})"
+
+    assert_pod_fail "${kata_pod_with_encrypted_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" 'failed to get decrypt key missing private key needed for decryption'
+}
+
+
+@test "Test that creating a container from an encrypted image, with correct decryption key works" {
+
+    setup_kbs_decryption_key "${DECRYPTION_KEY}" "${DECRYPTION_KEY_ID}"
+
+    create_pod_yaml_with_encrypted_image "${ENCRYPTED_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_encrypted_image}: $(cat ${kata_pod_with_encrypted_image})"
+
+    k8s_create_pod "${kata_pod_with_encrypted_image}"
+    echo "Kata pod test-e2e from encrypted image is running"
+}
+
+@test "Test that creating a container from an encrypted image, with incorrect decryption key fails" {
+
+    setup_kbs_decryption_key "anVua19rZXk=" "${DECRYPTION_KEY_ID}"
+
+    create_pod_yaml_with_encrypted_image "${ENCRYPTED_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_encrypted_image}: $(cat ${kata_pod_with_encrypted_image})"
+
+    assert_pod_fail "${kata_pod_with_encrypted_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" 'failed to get decrypt key missing private key needed for decryption'
+}
+
+teardown() {
+    if is_confidential_hardware; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    fi
+
+    if ! is_confidential_runtime_class; then
+        skip "Test not supported for ${KATA_HYPERVISOR}."
+    fi
+
+    if [ "${KBS}" = "false" ]; then
+        skip "Test skipped as KBS not setup"
+    fi
+
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
+    kubectl describe pods
+    k8s_delete_all_pods_if_any_exists || true
+
+    if [[ -n "${node_start_time}:-}" && -z "$BATS_TEST_COMPLETED" ]]; then
+        echo "DEBUG: system logs of node '$node' since test start time ($node_start_time)"
+        print_node_journal "$node" "kata" --since "$node_start_time" || true
+    fi
+}
--- a/tests/integration/kubernetes/lib.sh
+++ b/tests/integration/kubernetes/lib.sh
@@ -78,7 +78,7 @@ assert_logs_contain() {
 
 	# Note: with image-rs we get more than the default 1000 lines of logs
 	print_node_journal "$node" "$log_id" --since "$datetime" -n 100000 \
-		grep "$message"
+		| grep "$message"
 }
 
 # Create a pod then assert it fails to run. Use in tests that you expect the
--- a/tools/packaging/static-build/agent/Dockerfile
+++ b/tools/packaging/static-build/agent/Dockerfile
@@ -26,16 +26,17 @@ RUN mkdir ${RUSTUP_HOME} ${CARGO_HOME} && chmod -R a+rwX /opt
 RUN apt-get update && \
 	apt-get --no-install-recommends -y install \
 		ca-certificates \
+		clang \
 		curl \
 		g++ \
 		gcc \
+		libprotobuf-dev \
 		libssl-dev \
 		make \
 		musl-tools \
 		openssl \
 		perl \
-		protobuf-compiler \
-		clang && \
+		protobuf-compiler && \
 	apt-get clean && rm -rf /var/lib/apt/lists/ && \
 	curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${RUST_TOOLCHAIN}
 
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -24,7 +24,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "crypto-common",
- "generic-array",
+ "generic-array 0.14.7",
 ]
 
 [[package]]
@@ -36,13 +36,28 @@ dependencies = [
  "cfg-if 1.0.0",
  "cipher",
  "cpufeatures",
+ "zeroize",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
 ]
 
 [[package]]
 name = "ahash"
-version = "0.7.7"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
  "getrandom",
  "once_cell",
@@ -51,9 +66,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.5"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c378d78423fdad8089616f827526ee33c19f2fddbd5de1629152c9593ba4783"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
@@ -84,21 +99,36 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.58"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
+checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
 name = "arc-swap"
-version = "1.5.0"
+version = "1.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5d78ce20460b82d3fa150275ed9d55e21064fc7951177baacf86a145c4a4b1f"
+checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+
+[[package]]
+name = "arrayref"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
 
 [[package]]
 name = "arrayvec"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
+name = "ascii-canvas"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8824ecca2e851cec16968d54a01dd372ef8f95b244fb84b84e70128be347c3c6"
+dependencies = [
+ "term",
+]
 
 [[package]]
 name = "async-attributes"
@@ -116,19 +146,31 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c48ccdbf6ca6b121e0f586cbc0e73ae440e56c67c30fa0873b4e110d9c26d2b"
 dependencies = [
- "event-listener",
+ "event-listener 2.5.3",
  "futures-core",
 ]
 
 [[package]]
 name = "async-channel"
-version = "1.7.1"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e14485364214912d3b19cc3435dde4df66065127f05fa0d75c712f36f12c2f28"
+checksum = "81953c529336010edd6d8e358f886d9581267795c61b19475b71314bffa46d35"
 dependencies = [
- "concurrent-queue 1.2.4",
- "event-listener",
+ "concurrent-queue",
+ "event-listener 2.5.3",
  "futures-core",
+]
+
+[[package]]
+name = "async-channel"
+version = "2.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "89b47800b0be77592da0afd425cc03468052844aff33b84e33cc696f64e77b6a"
+dependencies = [
+ "concurrent-queue",
+ "event-listener-strategy",
+ "futures-core",
+ "pin-project-lite",
 ]
 
 [[package]]
@@ -143,21 +185,20 @@ dependencies = [
  "memchr",
  "pin-project-lite",
  "tokio",
- "zstd 0.13.0",
- "zstd-safe 7.0.0",
+ "zstd 0.13.2",
+ "zstd-safe 7.2.0",
 ]
 
 [[package]]
 name = "async-executor"
-version = "1.5.1"
+version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6fa3dc5f2a8564f07759c008b9109dc0d39de92a88d5588b8a5036d286383afb"
+checksum = "c8828ec6e544c02b0d6691d21ed9f9218d0384a82542855073c2a3f58304aaf0"
 dependencies = [
- "async-lock",
  "async-task",
- "concurrent-queue 2.2.0",
- "fastrand",
- "futures-lite",
+ "concurrent-queue",
+ "fastrand 2.1.0",
+ "futures-lite 2.3.0",
  "slab",
 ]
 
@@ -167,24 +208,24 @@ version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "279cf904654eeebfa37ac9bb1598880884924aab82e290aa65c9e77a0e142e06"
 dependencies = [
- "async-lock",
+ "async-lock 2.8.0",
  "autocfg",
  "blocking",
- "futures-lite",
+ "futures-lite 1.13.0",
 ]
 
 [[package]]
 name = "async-global-executor"
-version = "2.3.1"
+version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1b6f5d7df27bd294849f8eec66ecfc63d11814df7a4f5d74168a2394467b776"
+checksum = "05b1b633a2115cd122d73b955eadd9916c18c8f510ec9cd1686404c60ad1c29c"
 dependencies = [
- "async-channel",
+ "async-channel 2.3.1",
  "async-executor",
- "async-io",
- "async-lock",
+ "async-io 2.3.3",
+ "async-lock 3.4.0",
  "blocking",
- "futures-lite",
+ "futures-lite 2.3.0",
  "once_cell",
 ]
 
@@ -194,18 +235,37 @@ version = "1.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fc5b45d93ef0529756f812ca52e44c221b35341892d3dcc34132ac02f3dd2af"
 dependencies = [
- "async-lock",
+ "async-lock 2.8.0",
  "autocfg",
  "cfg-if 1.0.0",
- "concurrent-queue 2.2.0",
- "futures-lite",
+ "concurrent-queue",
+ "futures-lite 1.13.0",
  "log",
  "parking",
- "polling",
- "rustix 0.37.3",
+ "polling 2.8.0",
+ "rustix 0.37.27",
  "slab",
- "socket2 0.4.9",
+ "socket2 0.4.10",
  "waker-fn",
+]
+
+[[package]]
+name = "async-io"
+version = "2.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d6baa8f0178795da0e71bc42c9e5d13261aac7ee549853162e66a241ba17964"
+dependencies = [
+ "async-lock 3.4.0",
+ "cfg-if 1.0.0",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite 2.3.0",
+ "parking",
+ "polling 3.7.2",
+ "rustix 0.38.34",
+ "slab",
+ "tracing",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -214,24 +274,34 @@ version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "287272293e9d8c41773cec55e365490fe034813a2f172f502d6ddcf75b2f582b"
 dependencies = [
- "event-listener",
+ "event-listener 2.5.3",
+]
+
+[[package]]
+name = "async-lock"
+version = "3.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ff6e472cdea888a4bd64f342f09b3f50e1886d32afe8df3d663c01140b811b18"
+dependencies = [
+ "event-listener 5.3.1",
+ "event-listener-strategy",
+ "pin-project-lite",
 ]
 
 [[package]]
 name = "async-process"
-version = "1.7.0"
+version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a9d28b1d97e08915212e2e45310d47854eafa69600756fc735fb788f75199c9"
+checksum = "ea6438ba0a08d81529c69b36700fa2f95837bfe3e776ab39cde9c14d9149da88"
 dependencies = [
- "async-io",
- "async-lock",
- "autocfg",
+ "async-io 1.13.0",
+ "async-lock 2.8.0",
+ "async-signal",
  "blocking",
  "cfg-if 1.0.0",
- "event-listener",
- "futures-lite",
- "rustix 0.37.3",
- "signal-hook",
+ "event-listener 3.1.0",
+ "futures-lite 1.13.0",
+ "rustix 0.38.34",
  "windows-sys 0.48.0",
 ]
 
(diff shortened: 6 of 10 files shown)


Xynnn007:
BTW, I have a commit to make kata-agent fully get rid of `openssl`. I am not sure if it is ok to include that also in this PR or a separate one.

ChengyuZhu6:
@Xynnn007  I have successfully tested the code on my local machine, and it works to pull an unencrypted and unsigned image in the guest.

Xynnn007:
Changes upon upstream `guest-components/image-rs` are finished. I think the left things are
 - waiting for #8870 to be merged and take a rebase
 - Add CI test cases declared by in [Acceptance Criteria](https://github.com/kata-containers/kata-containers/issues/8111#issue-1925852036)
 - Let's review this!

Xynnn007:
@wainersm Thanks for the information. Seems like the current situation
| Case | install KBS | nydus-snapshotter|
|---|---|---|
|non-TEE|Yes|No|
|TEE(prefered)|#9379|Yes|

I am ok with your ideas. If anything I could help, lmk.

wainersm:
Hi @Xynnn007 ,

I liked your table pretty much so let me copy it here with added information:

| Case | install KBS | nydus-snapshotter|
|---|---|---|
|non-TEE|Yes|No|
|TEE(prefered)|#9379 #9273 |Yes|

Regardless whether TEE or non-TEE, the following is in the plan too:
 * [CC: build and publish encrypted image(s) for testing](https://github.com/kata-containers/kata-containers/issues/9360)

Related to:
 * [CC: Enable guest-components tests for AMD SNP](https://github.com/kata-containers/kata-containers/issues/9378)
 * [CC: Enable guest-components tests for AMD SEV](https://github.com/kata-containers/kata-containers/issues/9376)
 * [CC: Enable guest-components tests for Intel TDX](https://github.com/kata-containers/kata-containers/issues/9394)

Xynnn007:
Rebased to solve the conflicts.

src/agent/src/main.rs
```diff
@@ -427,6 +431,17 @@ fn init_attestation_components(logger: &Logger, _config: &AgentConfig) -> Result
...
+    let ocicrypt_config = serde_json::json!({
+        "key-providers": {
+            "attestation-agent":{
+                "ttrpc":CDH_URI
+            }
+        }
+    });
+
+    fs::write(OCICRYPT_CONFIG_PATH, ocicrypt_config.to_string().as_bytes())?;
+    env::set_var("OCICRYPT_KEYPROVIDER_CONFIG", OCICRYPT_CONFIG_PATH);
```
mkulke, line 443: would it be an option to move this (writing the file and setting the env) down the stack to image-rs?

Xynnn007: This might be a choice. But as a crate, `image-rs` wants the config env be passed via higher binary rather than set by itself.

I get your point setting such an env in kata-agent would make some potential trouble. A better solution to me is we can have CDH set this after moving image pull stuff to CDH fully per https://github.com/kata-containers/kata-containers/issues/9266

For now, as this way is used in CCv0, probably it makes sense to keep the workaround still until better solution is brought?

mkulke: > This might be a choice. But as a crate, image-rs wants the config env be passed via higher binary rather than set by itself.

Can you elaborate a bit? not sure I understand. Since ocicrypt-rs is an implementation detail of image-rs, wouldn't it be possible to set the env locally in image-rs like the [units tests](https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/decrypt.rs#L301-L304)?

This way we don't have to leak image-rs specifics to the kata codebase.

Xynnn007: `ocicrypt-rs` is a rust version of [ocicrypt](https://github.com/containers/ocicrypt), and `image-rs` is a rust version of [image](https://github.com/containers/image). The two are both libs and I think the setting thing should not also be added to image-rs.

Another reason is for [enclave-cc](https://github.com/confidential-containers/enclave-cc/blob/main/src/enclave-agent/src/config.rs#L39), the config path is set inside `image-rs`'s caller. If we set env inside image-rs, the original one would be overwritten.

I get your point that we'd better keep kata codebase away from image stuff. So my idea is to leave it as-is in this PR. Later, when we decided to move the image-pulling things fully to confidential data hub, kata-agent could have a very clean codebase without any image pull stuff even `image-rs`. Those things would all be implemented inside CDH. 

wdyt?

mkulke: If it's a requirement of enclave-cc, then this would be a reason to do this here. maybe there's an alternative that doesn't taint kata code directly, like the env could be set outside of kata the process `OCICRYPT_...=/etc/ocicrypt... kata-agent` and the config file added statically to the rootfs, not sure how/whether this would work in the current rootfs build process, though.

In general, I still think that ocicrypt-rs is a dep of image-rs (i.e. an impl detain that could change in the future) and hence kata has no business configuring it directly, but maybe i'm overly neurotic and others can chime in. cc @fidencio

Xynnn007: Yea I think set outside kata process is a right way to go for me. As the CDH socket is fixed, we could do this. If in non-coco scenario, the env actually will not be used. Technically it works for systemd, I am not sure whether init process works. Also I want to hear more voices upon this.

portersrc: >  mkulke: ...like the env could be set outside of kata the process OCICRYPT_...=/etc/ocicrypt... kata-agent and the config file added statically to the rootfs, not sure how/whether this would work in the current rootfs build process, though.

I agree with both these points: (1) make the config file statically part of the rootfs, assuming the build process isn't difficult, and (2) set the environment variable outside of kata-agent. And yet...

> Xynnn007: ... A better solution to me is we can have CDH set this after moving image pull stuff to CDH fully per https://github.com/kata-containers/kata-containers/issues/9266

--so if this change to `init_attestation_components` is temporary, it seems OK.

Xynnn007: Set env apart from both kata-agent and CDH might be the finest way for decoupling. some thoughts:
1. For guest images where `AGENT_INIT=false`, we can set the env in systemd-service config
2. For initrds where `AGENT_INIT=true`. How can we set the env for `/sbin/init`?

cc @fidencio

Xynnn007:
I temporarily mark this PR draft because we are not converged still on how to handle `ocicrypt.config` file. In the current commit I pushed I embed this file inside rootfs and mark the env in systemd of kata-agent. But for initrd whose init process is kata-agent, it is hard to set env. kindly ping @fidencio 

Thus my preference is to keep the config file setting inside kata-agent as a workaround. Once initrd supports systemd, we can move this file a part of rootfs -- both guest-image and initrd, and set env for kata-agent in systemd defining file. wdyt?

@mkulke

portersrc:
> Thus my preference is to keep the config file setting inside kata-agent as a workaround. 

Yes, as a workaround, I think this is also my preference.

Xynnn007:
Let's follow the old way to set ocicrypt config, s.t. putting it by kata-agent and setting env by kata-agent itself. In future let's use a better way.

mkulke:
Can you confirm this env does not need to be propagated to a kata-child processes (like CDH) but will only be consumed in kata-agent (via image-rs -> ocicrypt-rs dependencies)?

I'm asking because we would need a different workaround for systemd-managed guest-components

Xynnn007:
@mkulke Yes. Currently the env is consumed by `ocicrypt-rs`. The only parent dep of `ocicrypt-rs` now is `image-rs`. What's more, the only parent dep of `image-rs` now is `kata-agent`. Thus both AA and CDH will not care about the env.

src/agent/src/main.rs
```diff
@@ -478,10 +481,22 @@ fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<
         return Ok(None);
     }
 
+    let ocicrypt_config = serde_json::json!({
+        "key-providers": {
+            "attestation-agent":{
+                "ttrpc":CDH_SOCKET_URI
+            }
```
mkulke, line 488: it's a bit confusing that we specify the CDH socket in a `attestation-agent` field, not sure if it's worth fixing that in ocicrypt?

```suggestion
            "attestation-agent": {
                "ttrpc": CDH_SOCKET_URI
            }
```

mkulke (approved):
small nit

I understand there's no alternative to configure ocicrypt-rs in kata-agent, currently. so, I'm fine w/ the change, but maybe someone with a more-than-coco angle should ack this.

src/agent/src/main.rs
Xynnn007: yea, this is a historical problem. in old times unwrapkey api was provided by AA thus we have this name. Even we moved ability to CDH, for backward compability we kept this name.

stevenhorsman:
OK, I've added e2e tests and a few tweaks to the agent build to this PR, so will run the tests now and see what happens 🤞

fitzthum (commented):
This looks good. A few small comments.

In the past I have talked about adding something to the encrypted image that distinguishes it from a generic image and having the test check to make sure we aren't somehow pulling an unencrypted image. We might still think about doing that, but we can add it later. I think the negative tests here should rule that out anyway.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,130 @@
...
+    set_metadata_annotation "${kata_pod_with_encrypted_image}" \
+        "${kernel_params_annotation}" \
+        "${kernel_params_value}"
+
+    add_allow_all_policy_to_yaml "${kata_pod_with_encrypted_image}"
+}
+
+@test "Test that creating a container from an encrypted image, with no decryption key fails" {
+
+    # TODO - ideally in future we'd have a delete kbs resource to ensure there is no key here
```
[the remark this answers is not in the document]
stevenhorsman: Would that mean it would be release for TEE environments though?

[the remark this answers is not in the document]
fitzthum: Yup it should block everything regardless of platform. The only thing that might be slightly weird is that when you go to then re-enable things, if you use the allow all, that would allow sample evidence on TEE platforms. I guess you could move the test to the end or add a function to switch back to the default policy which blocks the sample evidence. Maybe it's not really worth it idk.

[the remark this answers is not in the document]
stevenhorsman: Hmm, I'm not sure if we have a default policy to switch to, so I might back this out for now as even if I move it to the end of this test file it might end up running before another KBS at some point and this would be a non obvious trap causing failures elsewhere (though we are sort of in the same position running it first without this). We can think about it a bit more and whether we add delete resource.

stevenhorsman:
> @stevenhorsman Are these logs from kata-agent's unit tests? The `anyhow` upgration is for builtable for image, and [commit](https://github.com/kata-containers/kata-containers/pull/9904/commits/4cc672a2eb5a41c5c11f9a68bdf7aa96bc76456a) only influences the unit tests of kata-agent.
> 
> If for e2e test, maybe we can try @wainersm 's comment [#9904 (comment)](https://github.com/kata-containers/kata-containers/pull/9904#issuecomment-2221414256) and see if anything wrong with resource retrievement.

Sorry @Xynnn007 - I posted this on the wrong issue - it should have been on this PR. These logs are from both the authenticated and encrypted image e2e tests that both pick up the latest image-rs and anyhow bump. Encrypted image support is working, but we log the errors in case of failures and we have tests that deliberately cause failures e.g. you can see it in the `ok 1 Test that creating a container from an encrypted image, with no decryption key fails` test of https://github.com/kata-containers/kata-containers/actions/runs/9881657476/job/27309915953?pr=9382

stevenhorsman:
The tests are all passing now. There is a little bit of funkiness with the guest-component logging, but I think that is a separate issue to look at. Based on this I think it's ready for review.
```
[run_kubernetes_tests.sh:116] INFO: Executing k8s-guest-pull-image-encrypted.bats
1..3
ok 1 Test that creating a container from an encrypted image, with no decryption key fails
...
# Attempt to create the container but it should fail
# pod/test-e2e created
...
#     Last State:  Terminated
#       Reason:    StartError
#       Message:   failed to create containerd task: failed to create shim task: failed to handle layer: failed to get decrypt key missing private key needed for decryption
...
ok 2 Test that creating a container from an encrypted image, with correct decryption key works
...
# apiVersion: v1
# kind: Pod
# metadata:
#   name: test-e2e
#   annotations:
#     io.containerd.cri.runtime-handler: kata-qemu-coco-dev
#     io.katacontainers.config.hypervisor.kernel_params: ' agent.guest_components_procs=confidential-data-hub agent.aa_kbc_params=cc_kbc::http://kbs.2d5bb7f688f849439b6d.eastus.aksapp.io:80'
#     io.katacontainers.config.agent.policy: cGFja2FnZSBhZ2VudF9wb2xpY3kKCmRlZmF1bHQgQWRkQVJQTmVpZ2hib3JzUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgQWRkU3dhcFJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IENsb3NlU3RkaW5SZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBDb3B5RmlsZVJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IENyZWF0ZUNvbnRhaW5lclJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IENyZWF0ZVNhbmRib3hSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBEZXN0cm95U2FuZGJveFJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IEV4ZWNQcm9jZXNzUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgR2V0TWV0cmljc1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IEdldE9PTUV2ZW50UmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgR3Vlc3REZXRhaWxzUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgTGlzdEludGVyZmFjZXNSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBMaXN0Um91dGVzUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgTWVtSG90cGx1Z0J5UHJvYmVSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBPbmxpbmVDUFVNZW1SZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBQYXVzZUNvbnRhaW5lclJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFB1bGxJbWFnZVJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFJlYWRTdHJlYW1SZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBSZW1vdmVDb250YWluZXJSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBSZW1vdmVTdGFsZVZpcnRpb2ZzU2hhcmVNb3VudHNSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBSZXNlZWRSYW5kb21EZXZSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBSZXN1bWVDb250YWluZXJSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBTZXRHdWVzdERhdGVUaW1lUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgU2V0UG9saWN5UmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgU2lnbmFsUHJvY2Vzc1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFN0YXJ0Q29udGFpbmVyUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgU3RhcnRUcmFjaW5nUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgU3RhdHNDb250YWluZXJSZXF1ZXN0IDo9IHRydWUKZGVmYXVsdCBTdG9wVHJhY2luZ1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFR0eVdpblJlc2l6ZVJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFVwZGF0ZUNvbnRhaW5lclJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFVwZGF0ZUVwaGVtZXJhbE1vdW50c1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFVwZGF0ZUludGVyZmFjZVJlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFVwZGF0ZVJvdXRlc1JlcXVlc3QgOj0gdHJ1ZQpkZWZhdWx0IFdhaXRQcm9jZXNzUmVxdWVzdCA6PSB0cnVlCmRlZmF1bHQgV3JpdGVTdHJlYW1SZXF1ZXN0IDo9IHRydWUK
# spec:
#   runtimeClassName: kata-qemu-coco-dev
#   containers:
#     - name: test-container
#       image: ghcr.io/confidential-containers/test-container:multi-arch-encrypted
#       imagePullPolicy: Always
#       command:
#         - sleep
#         - "30"
#   nodeName: aks-nodepool1-11709097-vmss000000
# pod/test-e2e created
# pod/test-e2e condition met
# Kata pod test-e2e from encrypted image is running
...
# Events:
#   Type    Reason   Age   From     Message
#   ----    ------   ----  ----     -------
#   Normal  Pulling  2s    kubelet  Pulling image "ghcr.io/confidential-containers/test-container:multi-arch-encrypted"
#   Normal  Pulled   1s    kubelet  Successfully pulled image "ghcr.io/confidential-containers/test-container:multi-arch-encrypted" in 131ms (131ms including waiting)
#   Normal  Created  1s    kubelet  Created container test-container
#   Normal  Started  1s    kubelet  Started container test-container
# pod "test-e2e" deleted
ok 3 Test that creating a container from an encrypted image, with incorrect decryption key fails
...
# Attempt to create the container but it should fail
...
#     Last State:  Terminated
#       Reason:    StartError
#       Message:   failed to create containerd task: failed to create shim task: failed to handle layer: failed to get decrypt key missing private key needed for decryption
...
```

Xynnn007:
@stevenhorsman The logs seem normal to me. However, `...missing private key needed for decryption` this log hints that getting key failed because (mostly from KBS), thus a little confused. I think you are referring to this?

stevenhorsman:
> @stevenhorsman The logs seem normal to me. However, `...missing private key needed for decryption` this log hints that getting key failed because (mostly from KBS), thus a little confused. I think you are referring to this?

Sorry for the confusion, in the logs I printed I cut out a lot of the content to make it more easy to read. The logging "issue" is the empty stacetrace that shows up as we've discussed. The full error is:
```
#       Message:   failed to create containerd task: failed to create shim task: failed to handle layer: failed to get decrypt key missing private key needed for decryption
#
# Stack backtrace:
#    0: <unknown>
#    1: <unknown>
#    2: <unknown>
#    3: <unknown>
#    4: <unknown>
#    5: <unknown>
#    6: <unknown>
#    7: <unknown>
#    8: <unknown>
#    9: <unknown>
#   10: <unknown>
#   11: <unknown>
#   12: <unknown>
#   13: <unknown>
#   14: <unknown>
#   15: <unknown>
#   16: <unknown>
#
# Stack backtrace:
#    0: <unknown>
#    1: <unknown>
#    2: <unknown>
#    3: <unknown>
#    4: <unknown>
#    5: <unknown>
#    6: <unknown>
#    7: <unknown>
#    8: <unknown>
#    9: <unknown>
#   10: <unknown>
#   11: <unknown>
#   12: <unknown>
#   13: <unknown>
#   14: <unknown>
#   15: <unknown>
#   16: <unknown>: unknown
```

tools/packaging/static-build/agent/Dockerfile
```diff
@@ -26,16 +26,17 @@ RUN mkdir ${RUSTUP_HOME} ${CARGO_HOME} && chmod -R a+rwX /opt
 RUN apt-get update && \
 	apt-get --no-install-recommends -y install \
 		ca-certificates \
+		clang \
```
gkurz, line 29: IIUC this fixes build of the agent caused by the previous commit. Please squash the fix there for the sake of bisect.

src/agent/Makefile
```diff
@@ -159,7 +159,7 @@ vendor:
 
 #TARGET test: run cargo tests
 test: $(GENERATED_FILES)
-	@cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
+	@RUST_LIB_BACKTRACE=0 cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
```
gkurz, line 162: Same remark as with the previous commit, let's squash this fix along with the image-rs bump.

tests/integration/kubernetes/lib.sh
```diff
@@ -78,7 +78,7 @@ assert_logs_contain() {
 
 	# Note: with image-rs we get more than the default 1000 lines of logs
 	print_node_journal "$node" "$log_id" --since "$datetime" -n 100000 \
-		grep "$message"
+		| grep "$message"
```
gkurz, line 81: Ohhh what a nasty bug ! Good catch Steve !

Does it fix something introduced by this PR ? If yes, I suggest you make this the first commit, for the sake of bisect again. :smiley:

stevenhorsman: It took a long while to spot. The complexity is that without this change, doesn't actually break any tests. We use that function in a few different tests, including this, but the behaviour is that the grep fails, but test still passes, which is worrying, but explains why we missed this before.

I don't think it *needs* to go first as the tests pass without it (they just don't actually test the error message). I can move it up though if you would like.

stevenhorsman: Done

gkurz (approved):
/lgtm

Thanks Ding and Steve !

fitzthum (approved):
\o/

tests/integration/kubernetes/lib.sh
wainersm: Hi @stevenhorsman @gkurz !

hmm... one test that uses `assert_logs_contain` is `k8s-measured-rootfs.bats` that run only if qemu-tdx. Guess what? With the fix now that test fails...

https://github.com/kata-containers/kata-containers/actions/runs/9893453384/job/27330646200?pr=9382
```
2024-07-11T15:53:36.7466161Z not ok 1 Test cannnot launch pod with measured boot enabled and incorrect hash
2024-07-11T15:53:36.7504034Z # (from function `assert_logs_contain' in file lib.sh, line 81,
2024-07-11T15:53:36.7505839Z #  in test file k8s-measured-rootfs.bats, line 55)
2024-07-11T15:53:36.7536942Z #   `assert_logs_contain "$node" kata "$node_start_time" \' failed
```

The `assert_logs_contain` is also used in k8s-guest-pull-image.bats which is only executed if qemu-coco-dev. However, the only test case that call the function (`Test we can pull a large image inside the guest`) is actually skipped in qemu-coco-dev... that's why we never noticed the problem.

` ok 2 Test we can pull a large image inside the guest # skip Test not supported for qemu-coco-dev`

stevenhorsman: Ok, I'll add a commit to skip that!

stevenhorsman: Hmm, so do we understand if measured root is just broken? I don't really want to scope creep this PR into trying to debug and fix that, so we could skip it and create an issue to track and enable it on all confidential systems?

stevenhorsman: I've pushed a commit to print the logs for this test, in case it's just a log format issue and simple to fix

wainersm: Hi @stevenhorsman , I was looking at this problem and the best we can do is to follow your idea: disable the measured rootfs test; later  we decide what to do.

Longer discussion... the only kernel built for measured rootfs was the kernel-tdx-experimental, so that only the qemu-tdx job runs the test. The test check it cannot boot the VM and checks for a given message on the logs (using the `assert_logs_contain`). On commit 6cbdba72687683c2820d941905d934b8855baf62 we switched all TEE configurations to use the same kernel-confidential, so rootfs measured is disabled for qemu-tdx too now. The VM still fail to boot (because of a different reason...) but the bug on the `assert_logs_contain` is masking the checks on the logs.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,133 @@
...
+
+load "${BATS_TEST_DIRNAME}/lib.sh"
+load "${BATS_TEST_DIRNAME}/confidential_common.sh"
+load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
+
+export KBS="${KBS:-false}"
+
+setup() {
+    if is_confidential_hardware; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
```
wainersm, line 15: Hi @stevenhorsman !

Is this skip intentional or just inherited from non-encrypted tests (k8s-guest-pull-image.bats)?

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,133 @@
...
+    ENCRYPTED_IMAGE="${ENCRYPTED_IMAGE:-ghcr.io/confidential-containers/test-container:multi-arch-encrypted}"
+    DECRYPTION_KEY="${DECRYPTION_KEY:-HUlOu8NWz8si11OZUzUJMnjiq/iZyHBJZMSD3BaqgMc=}"
+    DECRYPTION_KEY_ID="${DECRYPTION_KEY_ID:-ssh-demo}"
+}
+
+function setup_kbs_decryption_key() {
+    decryption_key=$1
+    decryption_key_id=$2
+
+    if [ "${KBS}" = "false" ]; then
```
wainersm, line 34: All the tests on this file require KBS so this should be on `setup()`.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,133 @@
...
+        skip "Test skipped as KBS not setup"
+    fi
+
+    if ! is_confidential_hardware; then
```
wainersm, line 38: Likewise?

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,133 @@
...
+    # For debug sake
+    echo "Pod ${kata_pod_with_encrypted_image}: $(cat ${kata_pod_with_encrypted_image})"
+
+    k8s_create_pod "${kata_pod_with_encrypted_image}"
+    echo "Kata pod test-e2e from encrypted image is running"
+}
+
+@test "Test that creating a container from an encrypted image, with incorrect decryption key fails" {
+
+    setup_kbs_decryption_key "anVua19rZXk=" "${DECRYPTION_KEY_ID}"
```
wainersm, line 104: I decoded `anVua19rZXk=` because I was absolutely sure you made Easter Egg :)

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,133 @@
...
+
+    create_pod_yaml_with_encrypted_image "${ENCRYPTED_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_encrypted_image}: $(cat ${kata_pod_with_encrypted_image})"
+
+    assert_pod_fail "${kata_pod_with_encrypted_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" 'failed to get decrypt key missing private key needed for decryption'
```
wainersm, line 112: This log message is weird, isn't it? Because I would expect it to actually get the decrypt key but failing to actually decrypt the image.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
```diff
@@ -0,0 +1,133 @@
...
+        "${kernel_params_value}"
+
+    add_allow_all_policy_to_yaml "${kata_pod_with_encrypted_image}"
+}
+
+@test "Test that creating a container from an encrypted image, with no decryption key fails" {
+
+    # TODO - there is now delete KBS resource to ensure there is no key, so we need to keep
+    # this test running first to ensure that the KBS doesn't have the resource. An alternative
+    # is to run kbs_set_deny_all_resources, but we don't have a way to reset to the default
```
wainersm, line 75: About "we don't have a way to reset to the default policy" made me realize a problem on k8s-confidential-attestation.bats because it calls `kbs_set_deny_all_resources`.  We are lucky (or @stevenhorsman made with intent?) that on `K8S_TEST_SMALL_HOST_UNION` k8s-guest-pull-image-encrypted.bats is listed before k8s-confidential-attestation.bats so the former run first. Certainly we should solve that problem, but not on this PR I suggest.

tests/integration/kubernetes/lib.sh
stevenhorsman: Ok, thanks for clarifying the kernel situation, so we don't actually expect this to pass now on any TEE, so skipping makes even more sense

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: A bit of both. It seems like the pattern we are taking (see guest-pull, attestation and sealed-secrets tests) is to first enabled it on non-TEE and then to test and roll it out to the other runtimeclasses, in order to not be delayed but issues with just one platform. I'm happy to test it on all now in case though

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: A bit of luck and a bit of judgement. I realised in the discussion with Tobin above that it could be a problem to have multiple of them and the ordering is important, which is a no-no for testing for me, hence the TODO warning, but I agree it isn't very sustainable and we need to ensure that the attestation tests are closer to the bottom.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: That is true. @Xynnn007 - do you have any thoughts about this? Anything I might be doing wrong in the test, or could the log message be better in the guest-components?

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: I was tempted by a "CONMEBOL sucks" key, but we have UEFA, so it would be a bit harsh!

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: So I was interested in what would happen in the "no decryption key scenario" if there was no KBS, but I do set `    kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"` in the pod yaml, so it's probably not a very fair test.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: I don't follow this one - this check is to ensure that we only enable sample kbs in the non-TEE hardware, so we don't get false positives in case of a verifier problem on the real TEE

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
Xynnn007: Yes. I think it should  be promoted by guest components. An ancient issue for this https://github.com/confidential-containers/guest-components/issues/92 ; (

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
stevenhorsman: So the SNP tests passed for encrypted image, but other failed and I'm not sure if it was related or now.
TDX failed, but I'm not sure why as the behaviour lots correct

stevenhorsman:
I need to close and re-open this PR to refresh the expected tests as it is trying to run some on garm

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
wainersm: hey, I created the issue and didn't even remember! I should find a doctor...

ok, so it is indeed failing to decrypt but just the message a little bit misleading.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
wainersm: Got it. I only wanted to check it was intentional. IMHO we should stick with the tradition here: enable for no-tee then roll it out to the TEEs as follow ups.

tests/integration/kubernetes/k8s-guest-pull-image-encrypted.bats
wainersm: Sorry, I meant to get it moved to setup() likewise the previous comment. But let's not bother about it right now... the confidential tests have too many boilerplates that I plan to refactor, so I can handle this case too.

wainersm:
Hi @stevenhorsman @Xynnn007 ! Finished my review. Dropping https://github.com/kata-containers/kata-containers/pull/9382/commits/bee37096536d6171c4bf7de6a989df9607af0d97 and we are ready to merge this IMO.

wainersm (approved):
Thanks @Xynnn007 @stevenhorsman !

stevenhorsman:
@BbolroC for tomorrow - the s390x rootfs build is failing:
```
#4 [2/2] RUN ([ -n "$http_proxy" ] &&     sed -i '$ a proxy='$http_proxy /etc/dnf/dnf.conf ; true) &&     dnf install -y         e2fsprogs         findutils         gcc         gdisk         parted         qemu-img         veritysetup         xfsprogs &&     dnf clean all
#4 sha256:3769c770d2f3de884984bddd56c7a0362a309c5a7216749a7b1aa0a742197da0
#4 2.804 Fedora 38 - s390x                                31 MB/s |  76 MB     00:02    
#4 21.06 Fedora 38 openh264 (From Cisco) - s390x         6.3 kB/s | 2.6 kB     00:00    
#4 21.96 Fedora Modular 38 - s390x                       4.4 MB/s | 2.7 MB     00:00    
#4 23.24 Fedora 38 - s390x - Updates                     316  B/s | 196  B     00:00    
#4 23.24 Errors during downloading metadata for repository 'updates':
#4 23.24   - Status code: 404 for http://dl.fedoraproject.org/pub/fedora-secondary/updates/38/Everything/s390x/repodata/repomd.xml (IP: 38.145.60.23)
#4 23.24   - Status code: 404 for https://dl.fedoraproject.org/pub/fedora-secondary/updates/38/Everything/s390x/repodata/repomd.xml (IP: 38.145.60.23)
#4 23.24   - Status code: 404 for https://d2lzkl7pfhq30w.cloudfront.net/pub/fedora-secondary/updates/38/Everything/s390x/repodata/repomd.xml (IP: 18.164.93.81)
#4 23.25 Error: Failed to download metadata for repo 'updates': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
#4 ERROR: process "/bin/sh -c ([ -n \"$http_proxy\" ] &&     sed -i '$ a proxy='$http_proxy /etc/dnf/dnf.conf ; true) &&     dnf install -y         e2fsprogs         findutils         gcc         gdisk         parted         qemu-img         veritysetup         xfsprogs &&     dnf clean all" did not complete successfully: exit code: 1
------
 > [2/2] RUN ([ -n "$http_proxy" ] &&     sed -i '$ a proxy='$http_proxy /etc/dnf/dnf.conf ; true) &&     dnf install -y         e2fsprogs         findutils         gcc         gdisk         parted         qemu-img         veritysetup         xfsprogs &&     dnf clean all:
------
process "/bin/sh -c ([ -n \"$http_proxy\" ] &&     sed -i '$ a proxy='$http_proxy /etc/dnf/dnf.conf ; true) &&     dnf install -y         e2fsprogs         findutils         gcc         gdisk         parted         qemu-img         veritysetup         xfsprogs &&     dnf clean all" did not complete successfully: exit code: 1
make: *** [Makefile:132: /home/ansible/gha-runner/_layout/_work/kata-containers/kata-containers/tools/osbuilder/kata-containers.img] Error 1
make[1]: *** [tools/packaging/kata-deploy/local-build/Makefile:72: rootfs-image-tarball-build] Error 2
make[1]: Leaving directory '/home/ansible/gha-runner/_layout/_work/kata-containers/kata-containers'
make: *** [tools/packaging/kata-deploy/local-build/Makefile:141: rootfs-image-tarball] Error 2
```
I've tried it a few times over the weekend, so I don't think it's a temporary blip. Fedora 38 has been EoL for a couple of months, so I don't know if it is related to that? Any ideas?

BbolroC:
> I've tried it a few times over the weekend, so I don't think it's a temporary blip. Fedora 38 has been EoL for a couple of months, so I don't know if it is related to that? Any ideas?

Thanks for the notice. I will look into it now and share what I figure out here.

BbolroC:
Let's see how the CI reacts to https://github.com/kata-containers/kata-containers/pull/10010. Thanks!

stevenhorsman:
The TDX and SNP failures (`k8s-confidential.bats`) are due to `create instance 14310: object with key "14310" already exists: unknown`

stevenhorsman:
> The TDX and SNP failures (`k8s-confidential.bats`) are due to `create instance 14310: object with key "14310" already exists: unknown`. The SNP tests are failing on the last few nightlies, so it's probably not worth re-running them.

TDX passing after re-run!

Issue #9336, opened 2024-03-22
It fails to run a kata containers on host with cgroup v2. 


Apokleos:
# Get your issue reviewed faster

Inspired by [Eric's issue](https://github.com/kata-containers/kata-containers/issues/9327), I did some tests on the `defaultMaxCPU` setting in runtime-rs. I wanted to see if I could simulate the user-specified number of CPUs for Kata when `defaultMaxCpu` is set to `0`.  For example, if I specify 4 vCPUs for Kata, the expected result is that 4 CPUs should be visible in Kata. This test was successful on my development machine (which cgroup v1 is used), but I encountered problems on another node (cgroup v2).

I tested **runtime-rs** with **cloud-hypervisor** and **dragonball**, both of them have such problems.

# Description of problem

```
# nerdctl run --cpus 4 -ti --runtime io.containerd.kata.v2 --net=bridge --rm fedora:35
WARN[0000] cannot set cgroup manager to "systemd" for runtime "io.containerd.kata.v2" 
FATA[0002] failed to create shim task: Others("failed to handle message create container\n\nCaused by:\n    0: create\n    1: agent create container\n    2: rpc status: Status { code: INTERNAL, message: \"failed to set unit nerdctl-0ed3411a368b8a02d57882d4d0622bbca4a3aaac4758efc0dee6260d71ef3b91.scope properties\\n\\nCaused by:\\n    org.freedesktop.DBus.Error.InvalidArgs: Value specified in CPUShares is out of range\", details: [], special_fields: SpecialFields { unknown_fields: UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }"): unknown 
```

# Expected result

It will run successully

# Actual result

Failed with error 
```
FATA[0002] failed to create shim task: Others("failed to handle message create container\n\nCaused by:\n    0: create\n    1: agent create container\n    2: rpc status: Status { code: INTERNAL, message: \"failed to set unit nerdctl-0ed3411a368b8a02d57882d4d0622bbca4a3aaac4758efc0dee6260d71ef3b91.scope properties\\n\\nCaused by:\\n    org.freedesktop.DBus.Error.InvalidArgs: Value specified in CPUShares is out of range\", details: [], special_fields: SpecialFields { unknown_fields: UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }"): unknown 
```

# Further information

[...]

PR #9383, opened 2024-03-29
kata-agent: enabling cgroups-v2 by systemd.unified_cgroup_hierarchy

Apokleos:
Configure the system to mount cgroups-v2 by default during system boot by the systemd system, We must add the systemd.unified_cgroup_hierarchy=1 parameter to kernel cmdline, which will be passed by kernel_params in configuration.toml. To enable cgroup-v2, just add systemd.unified_cgroup_hierarchy=true[1] to kernel_params.

Fixes: #9336


justxuewei (approved):
Lgtm, thanks!

Tim-Zhang (approved):
LGTM

Issue #9384, opened 2024-03-31
Update`run-k8s-tests-on-aks.yaml` to support new paramters: `use-policy` and `genpolicy-pull-method`


Redent0r:
**Which feature do you think can be improved?**
Yaml parameters on run-k8s-tests-on-aks.yaml

Add `use-policy` and `genpolicy-pull-method` as matrix params. This will add an entry to test matrix and provide test coverage for https://github.com/kata-containers/kata-containers/issues/9144 / https://github.com/kata-containers/kata-containers/pull/9185

Redent0r:
Fixed by https://github.com/kata-containers/kata-containers/pull/9385

PR #9385, opened 2024-03-31
tests: pass AUTO_GENERATE_POLICY and GENPOLICY_PULL_METHOD from workf…

Redent0r:
…low yamls

This will improve policy parameters for testing

Fixes: #9384


.github/workflows/run-kata-coco-tests.yaml
```diff
@@ -45,6 +45,7 @@ jobs:
       K8S_TEST_HOST_TYPE: "baremetal"
       SNAPSHOTTER: ${{ matrix.snapshotter }}
       PULL_TYPE: ${{ matrix.pull-type }}
+      AUTO_GENERATE_POLICY: "true"
```
Redent0r, line 48: Changes in this file are for trying to maintain behavior from https://github.com/kata-containers/kata-containers/pull/9385/files#diff-55a1b92d4c9735afcdbe6ef38905599f0f58dbdd7cd5f8db16333529d57a47c3L94-L95

tests/integration/kubernetes/gha-run.sh
```diff
@@ -253,8 +254,6 @@ function run_tests() {
 
 	# Enable auto-generated policy for CI images that support policy.
 	#
-	# TODO: enable testing auto-generated policy for other types of hosts too.
-	[ "${KATA_HOST_OS}" = "cbl-mariner" ] && export AUTO_GENERATE_POLICY="yes"
```
Redent0r, line 257: AUTO_GENERATE_POLICY is already used in a few places. Moving this so it gets set in the yaml by each hosts that wants to autogenerate

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -45,9 +45,18 @@ jobs:
           - host_os: cbl-mariner
             vmm: clh
             instance-type: small
+            use-policy: true
+            genpolicy-pull-method: oci
```
sprt, line 49: @danmihai1 Does `oci` make sense? I remember you mentioned `oci-distribution` before.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -33,6 +33,7 @@ HTTPS_PROXY="${HTTPS_PROXY:-${https_proxy:-}}"
 NO_PROXY="${NO_PROXY:-${no_proxy:-}}"
 export AUTO_GENERATE_POLICY="${AUTO_GENERATE_POLICY:-no}"
```
sprt, line 34: The workflow YAML uses `true/false` instead of `yes/no` so let's be consistent and default this to `false`

tests/integration/kubernetes/run_kubernetes_tests.sh
```diff
@@ -91,8 +91,7 @@ fi
 policy_tests_enabled() {
 	# The Guest images for these platforms have been built using AGENT_POLICY=yes -
 	# see kata-deploy-binaries.sh.
-	[ "${KATA_HYPERVISOR}" == "qemu-sev" ] || [ "${KATA_HYPERVISOR}" == "qemu-snp" ] || \
-		[ "${KATA_HYPERVISOR}" == "qemu-tdx" ] || [ "${KATA_HOST_OS}" == "cbl-mariner" ]
+	[ "${AUTO_GENERATE_POLICY}" == "yes" ]
```
sprt, line 94: In this PR this variable will never be set (because the YAML changes won't be reflected). So we need to find a way to keep the CI working for both these cases:

1. With only script changes taken into account (for this PR's checks).
2. With both script changes and YAML changes taken into account (so that other PRs aren't broken right after you merge this one).

sprt (changes requested):

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -45,9 +45,18 @@ jobs:
           - host_os: cbl-mariner
             vmm: clh
             instance-type: small
+            use-policy: true
```
danmihai1, line 48: USE_POLICY should be enabled for SEV, SNP, TDX and Mariner - because these are the 4 Guest images that support Policy.

AUTO_GENERATE_POLICY should be enabled for now just on Mariner. Hopefully it we will get AUTO_GENERATE_POLICY enabled for other platforms too, soon.

danmihai1 (changes requested):

.github/workflows/run-k8s-tests-on-aks.yaml
Redent0r: Thanks Dan. I had this thought that USE_POLICY was equivalent to AUTO_GENERATE_POLICY. But I see now USE_POLICY is more meant specifically for deciding when to [run_policy_specific_tests](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/run_kubernetes_tests.sh#L171)

danmihai1:
@Redent0r , I would prefer a more concise commit and PR format - e.g.,

```
gha: add GENPOLICY_PULL_METHOD
```

```
Add GENPOLICY_PULL_METHOD that will be used to test pulling container
images in genpolicy using the oci-distribution crate and/or the
containerd interface.

GENPOLICY_PULL_METHOD will start being used in a future PR.

Fixes: #9384
```

tests/integration/kubernetes/gha-run.sh
```diff
@@ -33,6 +33,7 @@ HTTPS_PROXY="${HTTPS_PROXY:-${https_proxy:-}}"
...
 export TEST_CLUSTER_NAMESPACE="${TEST_CLUSTER_NAMESPACE:-kata-containers-k8s-tests}"
+export GENPOLICY_PULL_METHOD="${GENPOLICY_PULL_METHOD:-oci-distribution}"
```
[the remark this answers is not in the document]
Redent0r: Oop, I missed this. Why? :) It doesn't make much difference to me since nothing is using that variable yet. Maybe better to leave it and "reserve" the env variable for us since we'll use it soon?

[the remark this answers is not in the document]
danmihai1: The goal of this PR is just to add a variable that cannot be added in the PR that will use it.

If it's not used here yet, then it's not tested yet, so it's better to avoid adding it untested.

But I agree this is not a big deal.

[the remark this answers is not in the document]
sprt: Agree with Dan on principle but also that it's not a big deal for this time. Just always good to be mindful of your reviewers' time (e.g. I had to read through this thread for a change that has no consequence) 🙂

.github/workflows/run-k8s-tests-on-aks.yaml
```diff
@@ -45,9 +45,15 @@ jobs:
...
             instance-type: small
+            genpolicy-pull-method: oci-distribution
+          - host_os: cbl-mariner
+            vmm: clh
+            instance-type: small
+            genpolicy-pull-method: containerd
           - host_os: cbl-mariner
             vmm: clh
             instance-type: normal
+            genpolicy-pull-method: oci-distribution
```
sprt, line 56: Do we not need the equivalent of this entry with the small instance type?

sprt (commented):
Just a question before I approve

.github/workflows/run-k8s-tests-on-aks.yaml
Redent0r: Not needed since no test for "normal" instance-type autogenerate policy. Actually I just removed this line as well

tests/integration/kubernetes/gha-run.sh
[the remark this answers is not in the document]
Redent0r: Removed :)

sprt (approved):
LGTM!

Issue #9392, opened 2024-04-02
qemu: vcpu id is not continuous after hot plugged


deagon:
# Description of problem

When using PodOverhead with `default_vcpus=2` and resource limit cpu with `cpu: 6`,  the function `qemu.hotplugAddCPUs(6)` is called as we want. However, the cpu physical id in GuestOS is not continuous.

# Expected result

```
root@localhost:/# cat /proc/cpuinfo  | grep 'physical id'
physical id	: 0
physical id	: 1
physical id	: 2
physical id	: 3
physical id	: 4
physical id	: 5
physical id	: 6
physical id	: 7
```

# Actual result

```
root@localhost:/# cat /proc/cpuinfo  | grep 'physical id'
physical id	: 0
physical id	: 1
physical id	: 95
physical id	: 94
physical id	: 93
physical id	: 92
physical id	: 91
physical id	: 90
```

# Further information

deagon:
https://github.com/kata-containers/kata-containers/blob/7795f9c016d3c97a5d144da36f9c439db995edc5/src/runtime/virtcontainers/qemu.go#L2064-L2072

The result of `query-hotpluggable-cpus` call is reversed.

```log
(QEMU) query-hotpluggable-cpus
[...]

PR #9393, opened 2024-04-02
qemu: make the continuous id for hotpluggableVCPUs

deagon:
Reverse loop over hotpluggableVCPUs to find the continuous vcpu id.

Fixes: #9392

### DIFF
--- a/src/runtime/virtcontainers/qemu.go
+++ b/src/runtime/virtcontainers/qemu.go
@@ -2069,7 +2069,9 @@ func (q *qemu) hotplugAddCPUs(amount uint32) (uint32, error) {
 	machine := q.arch.machine()
 
 	var hotpluggedVCPUs uint32
-	for _, hc := range hotpluggableVCPUs {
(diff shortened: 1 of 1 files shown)


pmores:
@deagon This change looks reasonable in what it does but could you add some explanation why this is worth doing? Like, what's the state now, how it changes with this PR and why the change is desirable?

deagon:
> @deagon This change looks reasonable in what it does but could you add some explanation why this is worth doing? Like, what's the state now, how it changes with this PR and why the change is desirable?

@pmores  

In my case, I need to check the debug-threads in qemu sometimes.
Obviously, it's graceful to see the continuous vcpu id list and easier to observe it.
What's more, I think there is no side effect on this patch because I have tested it in my prod environment several months.

deagon:
@Apokleos @justxuewei please take a review, thanks.

pmores (approved):
Sure, makes sense, thanks @deagon !

Issue #9375, opened 2024-03-28
runtime-rs: test errors for cri-containerd for QEMU


BbolroC:
# Description of problem

**IMPORTANT NOTE**: this issue should be handled after #9374 is resolved. 

While resolving #9371, the following tests failed for the `cri-containered` test with runtime-rs (qemu) on x86_64:

- TestContainerMemoryUpdate
- TestKilledVmmCleanup
- TestDeviceCgroup

These tests should pass, as they did with the runtime.

Attached you can find the log for each test:

[TestContainerMemoryUpdate.log](https://github.com/kata-containers/kata-containers/files/14793181/TestContainerMemoryUpdate.log)
[TestKilledVmmCleanup.log](https://github.com/kata-containers/kata-containers/files/14793184/TestKilledVmmCleanup.log)
[TestDeviceCgroup.log](https://github.com/kata-containers/kata-containers/files/14793188/TestDeviceCgroup.log)

BbolroC:
`TestContainerMemoryUpdate` has not been tested on s390x due to lack of memory hotplug on the platform.  But, for go-runtime, this will be fixed by https://github.com/kata-containers/kata-containers/pull/12573. The implementation for rust-runtimes will be followed.

PR #9403, opened 2024-04-03
CI: Enable GHA cri-containerd workflow for runtime-rs with QEMU

BbolroC:
This PR is to extend the existing e2e test `run-cri-containerd` to cover runtime-rs with QEMU for `x86_64` and `s390x` platforms.

A new value `qemu-runtime-rs` for `KATA_HYPERVISOR` will be utilized to handle a test scenario where
the VMM is QEMU and runtime-rs is employed.

Note: Some of the tests are skipped. They are going to be reintegrated in the follow-up PR (Check out #9375). 

The workflows are verified at: https://github.com/BbolroC/kata-containers/actions/runs/8636974281

Signed-off-by: Hyounggyu Choi Hyounggyu.Choi@ibm.com


.github/workflows/basic-ci-amd64.yaml
```diff
@@ -23,11 +23,26 @@ jobs:
       matrix:
         containerd_version: ['lts', 'active']
         vmm: ['clh', 'dragonball', 'qemu', 'stratovirt', 'cloud-hypervisor']
+        use-runtime-rs: ['no']
```
gkurz, line 26: This might be confusing for the casual CI user. This `use-runtime-rs` parameter is ignored when VMM is `dragonball` or `cloud-hypervisor`. Also CLH folks had a different approach : a new VMM value was created for runtime-rs. Should we do the same for QEMU ?

Cc @pmores

BbolroC: Thanks for the feedback. This means that CLH are testing both runtime and runtime-rs atm, right? Could you share a link for their approach? I will do the same thing.

tests/common.bash
```diff
@@ -455,8 +457,13 @@ function enabling_hypervisor() {
 			declare -r CONFIG_DIR="${KATA_DIR}/share/defaults/kata-containers/runtime-rs"
 			;;
 		*)
-			sudo ln -sf "${KATA_DIR}/bin/containerd-shim-kata-v2" "${CONTAINERD_SHIM_KATA}"
-			declare -r CONFIG_DIR="${KATA_DIR}/share/defaults/kata-containers"
+			if [ "${USE_RUNTIME_RS}" = "yes" ]; then
+				sudo ln -sf "${KATA_DIR}/runtime-rs/bin/containerd-shim-kata-v2" "${CONTAINERD_SHIM_KATA}"
+				declare -r CONFIG_DIR="${KATA_DIR}/share/defaults/kata-containers/runtime-rs"
```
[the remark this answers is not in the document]
BbolroC: I like your suggestion, thanks!

.github/workflows/basic-ci-amd64.yaml
BbolroC: ah.... you meant `clh` and `cloud-hypervisor`? I will have a look at it tomorrow. Thanks! 😉

gkurz: Yes I meant exactly that. Maybe @jodh-intel or @amshinde can provide insights ?

tests/common.bash
```diff
@@ -450,16 +456,12 @@ function enabling_hypervisor() {
...
 	case "${KATA_HYPERVISOR}" in
-		dragonball | cloud-hypervisor)
-			sudo ln -sf "${KATA_DIR}/runtime-rs/bin/containerd-shim-kata-v2" "${CONTAINERD_SHIM_KATA}"
-			declare -r CONFIG_DIR="${KATA_DIR}/share/defaults/kata-containers/runtime-rs"
-			;;
-		*)
-			sudo ln -sf "${KATA_DIR}/bin/containerd-shim-kata-v2" "${CONTAINERD_SHIM_KATA}"
-			declare -r CONFIG_DIR="${KATA_DIR}/share/defaults/kata-containers"
-			;;
+		dragonball|cloud-hypervisor) INT_RUNTIME_PATH="/runtime-rs" ;;
```
gkurz, line 459: ... `qemu-rs` would be treated like other `runtime-rs` citizens :smile:

tests/integration/cri-containerd/integration-tests.sh
```diff
@@ -257,6 +258,13 @@ function TestContainerMemoryUpdate() {
 
 	if [[ "${KATA_HYPERVISOR}" != "qemu" ]] || [[ "${ARCH}" == "ppc64le" ]] || [[ "${ARCH}" == "s390x" ]]; then
 		return
+	elif [[ "${KATA_HYPERVISOR}" == "qemu" ]]; then
+		if [[ -n "${INT_RUNTIME_PATH}" ]]; then
```
gkurz, line 262: `[[ "${KATA_HYPERVISOR}" == "qemu-rs" ]]` would fit much better here. :wink:

tests/integration/cri-containerd/integration-tests.sh
```diff
@@ -657,8 +665,14 @@ function main() {
 	TestContainerMemoryUpdate
 
 	if [[ "${ARCH}" != "ppc64le" ]]; then
-		TestKilledVmmCleanup
-		TestDeviceCgroup
+		if [[ "${KATA_HYPERVISOR}" == "qemu" ]] && [[ -n "${INT_RUNTIME_PATH}" ]]; then
```
gkurz, line 668: ditto

BbolroC:
Rationale behind the recent change:

- I love to introduce a new item (e.g. `qemu-rs`) for the `vmm` matrix because it would not raise the known issue for renaming a workflow.
- I do not want to go with using `qemu-rs` for `KATA_HYPERVISOR` because `qemu-rs` is not a new hypervisor. So `qemu-rs` should be re-assigned to `qemu` and exported again at the very early phase of the test.

But I have spotted https://github.com/kata-containers/kata-containers/pull/8996/files#r1477981658 for why `clh` and `cloud-hypervisor` are handled differently. Let's hold this PR until we figure out that the separation would also be helpful for the upcoming implementation for QEMU and, if that is the case, what a proper name would be rather than `qemu-rs` (`quick-emulator`??).

I am looking forward to your opinions for this @gkurz @pmores @GabyCT @jodh-intel

BbolroC:
@gkurz Could you review the latest change again? Test results look good: https://github.com/BbolroC/kata-containers/actions/runs/8631193142

pmores (approved):
thanks @BbolroC , I'm not knowledgeable about CI but the parts I can evaluate do look good to me!

tests/integration/cri-containerd/integration-tests.sh
```diff
@@ -257,6 +257,11 @@ function TestContainerMemoryUpdate() {
 
 	if [[ "${KATA_HYPERVISOR}" != "qemu" ]] || [[ "${ARCH}" == "ppc64le" ]] || [[ "${ARCH}" == "s390x" ]]; then
 		return
+	elif [[ "${KATA_HYPERVISOR}" == "qemu-runtime-rs" ]]; then
```
gkurz, line 260: This arm cannot be reached since `"${KATA_HYPERVISOR}" != "qemu"` obviously :wink:

If you really want to print the info message you should do this check first. Otherwise you can just drop that hunk.

gkurz (commented):
@BbolroC what about existing users of the "qemu" string in the tests directory ? It seems they should be audited and maybe changed to handle "qemu-runtime-rs" as well.

tests/integration/cri-containerd/integration-tests.sh
BbolroC: Yeah, you are right. I will update the PR. Thanks!

BbolroC:
> @BbolroC what about existing users of the `"qemu"` string in the tests directory ? It seems they should be audited and maybe changed to handle `"qemu-runtime-rs"` as well.

This PR only covers the tests under `tests/integration/cri-containerd`, not the whole tests under `tests`. We could gradually configure the CI for `qemu-runtime-rs` as more tests for the new `KATA_HYPERVISOR` come in.

gkurz (approved):
LGTM. Thanks @BbolroC !

Issue #9317, opened 2024-03-20
GHA: rm: cannot remove: Directory not empty on self-hosted runners for s390x


BbolroC:
# Description of problem

Occasionally, it is observed that an error occurs at a GHA workflow step `Adjust a permission for repo` like:

```
Run sudo chown -R $USER:$USER $GITHUB_WORKSPACE $HOME
  sudo chown -R $USER:$USER $GITHUB_WORKSPACE $HOME
  sudo rm -rf $GITHUB_WORKSPACE/*
  sudo rm -f /tmp/kata_hybrid*  # Sometime we got leftover from test_setup_hvsock_failed()
  shell: /usr/bin/bash -e {0}
rm: cannot remove '/home/ansible/gha-runner/_layout/_work/kata-containers/kata-containers/src/agent/target/s390x-unknown-linux-gnu/debug/build/openssl-sys-b15451[2](https://github.com/kata-containers/kata-containers/actions/runs/8357896317/job/22878038981?pr=8870#step:2:2)81ded02ef/out/openssl-build/build/src/crypto': Directory not empty
```

https://github.com/kata-containers/kata-containers/actions/runs/8357896317/job/22878038981?pr=8870

It appears that a file was being held by another process while the `rm` command was attempting to remove it. It's worth noting that `rm` itself isn't atomic. In such cases, a second attempt should typically succeed. However, to prevent this error from occurring repeatedly, we need to implement a more robust solution.

PR #9415, opened 2024-04-04
GHA: Implement secondary GITHUB_WORKSPACE cleanup on 1st failure

BbolroC:
Occasionally, the removal of GITHUB_WORKSPACE fails for self-hosted runners because one of the subdirectories is not empty. This is likely due to another process occupying the directory at the time.
Implementing a secondary cleanup resolves this issue. This commit focuses on the implementation for the secondary cleanup.

Fixes: #9317

### DIFF
--- a/.github/workflows/build-checks.yaml
+++ b/.github/workflows/build-checks.yaml
@@ -55,7 +55,7 @@ jobs:
       - name: Adjust a permission for repo
         run: |
           sudo chown -R $USER:$USER $GITHUB_WORKSPACE $HOME
-          sudo rm -rf $GITHUB_WORKSPACE/*
+          sudo rm -rf $GITHUB_WORKSPACE/* && echo "GITHUB_WORKSPACE removed" || { sleep 10 && sudo rm -rf $GITHUB_WORKSPACE/*; }
           sudo rm -f /tmp/kata_hybrid*  # Sometime we got leftover from test_setup_hvsock_failed()
         if: ${{ inputs.instance != 'ubuntu-20.04' }}
 


.github/workflows/build-checks.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Adjust a permission for repo
         run: |
           sudo chown -R $USER:$USER $GITHUB_WORKSPACE $HOME
-          sudo rm -rf $GITHUB_WORKSPACE/*
+          sudo rm -rf $GITHUB_WORKSPACE/* && echo "GITHUB_WORKSPACE removed" || { sleep 3 && sudo rm -rf $GITHUB_WORKSPACE/*; }
```
fidencio, line 58: How risky it'd be to actually loop till we can clean up the directory, preferably killing any process trying to access it during the iteration?

BbolroC: Being occupied by another processor is just an assumption, which has not been verified. Unfortunately, the failure is not reproducible and seems an issue from infra (IBM Cloud). Even I am not sure if atomicity of figuring out a process ID and killing it is shorter than a period of time for the occupancy. I think the second try of the removal would be the best option we have without reproducing/verifying the issue.

BbolroC: If your preference still goes with a loop, I will refactor the code. But the other day I luckily had a chance to face with this `directory not empty` error before. The immediate secondary removal was successful at the time.

fidencio: I'd say, just increase the sleep time to 10, for instance, to give us a little bit more leeway for now.
If this happens again, we should go for the loop approach, but I don't necessarily feel strong about taking this approach right now.

WDYT? Is it a fair compromise?

fidencio (approved):
lgtm, thanks @BbolroC!

stevenhorsman (approved):
LGTM. Thanks

Issue #9410, opened 2024-04-03
CC: run guest-pull image tests on non-TEE CI jobs


wainersm:
## Context

The [guest-pull image tests](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-guest-pull-image.bats) were merged in PR #8484 , and they are currently executed only in [TEE (SNP, SEV, TDX) jobs](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-kata-coco-tests.yaml).

The [CoCo release v0.9.0](https://github.com/confidential-containers/confidential-containers/issues/200) aim to provide the very first CoCo release out of Kata's main branch. Compared to the previous releases we don't plan to have feature parity, nor full support to TEEs. Instead the dev/test team will focus on putting together features that can be emulated/mocked on non-TEE environments.

Therefore, it will be required to run the guest-pull image tests on non-TEE CI jobs for the 0.9.0 release. 

## Work breakdown

TBD

Cc @ChengyuZhu6

wainersm:
To fill out the "work breakdown" section, I'd like to raise a discussion: what non-TEE job(s) it should be running the guest-pull image tests?

Before going any further, let's remember that the guest-pull image feature depends on nydus-snapshotter installed in environment. Ok?

The [port attestation agent from CCv0 branch to main branch](https://github.com/kata-containers/kata-containers/pull/8870) will have its tests running on the non-TEE [run-k8s-tests-on-aks](https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-k8s-tests-on-aks.yaml) for `qemu` hypervisor configuration which execute the kata-qemu "vanilla" tests on k8s.

Suppose that we choose to onboard on the existing non-TEE alike the attestation agent tests. There will be needed to install nydus-snapshotter and that's going to change the behavior of existing tests (I believe the tests will even break). Alternatively we could create a new parameter on the matrix (e.g. 'snapshotter') where only one generated job will be set to run the non-TEE tests. For example,

(from .github/workflows/run-k8s-tests-on-aks.yaml)
```yaml
  run-k8s-tests:
    strategy:
      fail-fast: false
      matrix:
        host_os:
          - ubuntu
        vmm:
          - clh
          - dragonball
          - qemu
          - stratovirt
          - cloud-hypervisor
        instance-type:
          - small
          - normal
        include:
          - host_os: cbl-mariner
            vmm: clh
            instance-type: small
[...]

PR #9424, opened 2024-04-05
CC: run guest-pull tests on non-TEE jobs

wainersm:
This is for running the existing guest-pull tests on non-TEE environment for the CoCo 0.9.0 release.

I could have introduced a new column to the current k8s jobs matrix (`.github/workflows/run-k8s-tests-on-aks.yaml`) but this would have created some issues as I described in https://github.com/kata-containers/kata-containers/issues/9410. So instead this is going to create a new matrix of jobs (`run-k8s-tests-coco-nontee`) alongside the TEE jobs in `.github/workflows/run-kata-coco-tests.yaml` for running the CoCo tests on non-TEE.

Initially it is going to run only the guest-pull tests but in the future I envision it being used for all CoCo's non-TEE tests, that's why I made it generic.

Talking about the guest-pull tests, because I  they are now running on non-CoCo runtimeclasses, i.e., when `KATA_HYPERVISOR=qemu`, I had to add a new guard to ensure the tests are executed only on environments that have a snapshotter installed. This way the tests won't get executed on the `run-k8s-tests-on-aks` jobs, for example.

"Simulated" these changes with the following script running from `./tests/integration/kubernetes`:
```bash
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

export AZ_RG="wmoschet-ci-tests"
export AKS_NAME="wmoschet-aks-ci-test"

#export DOCKER_REGISTRY=
#export DOCKER_REPO=
#export DOCKER_TAG=
#export GH_PR_NUMBER=
export KATA_HOST_OS=ubuntu
export KATA_HYPERVISOR=qemu
export KUBERNETES=vanilla
export K8S_TEST_HOST_TYPE=small
#export K8S_TEST_UNION="k8s-guest-pull-image.bats"
export PULL_TYPE=guest-pull
export SNAPSHOTTER=nydus
export USING_NFD="false"

prep_env() {
    echo "== CREATE CLUSTER =="
    ./gha-run.sh create-cluster
    echo "== GET CREDENTIALS =="
    ./gha-run.sh get-cluster-credentials
    echo "== DEPLOY SNAPSHOTTER =="
    ./gha-run.sh deploy-snapshotter
    echo "== DEPLOY KATA =="
    ./gha-run.sh deploy-kata-aks
}

teardown_env() {
    echo "== DESTROY CLUSTER =="
    ./gha-run.sh delete-cluster
}

run_tests() {
    echo "== RUN TESTS =="
    ./gha-run.sh run-tests
}

prep_env
run_tests
teardown_env
```

### DIFF
--- a/tests/integration/kubernetes/k8s-guest-pull-image.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image.bats
@@ -10,6 +10,8 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
     confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
     setup_common 
     unencrypted_image_1="quay.io/sjenning/nginx:1.15-alpine"
     unencrypted_image_2="quay.io/prometheus/busybox:latest"
@@ -17,7 +19,6 @@ setup() {
 }
 
 @test "Test we can pull an unencrypted image outside the guest with runc and then inside the guest successfully" {
-    [[ " ${SUPPORTED_NON_TEE_HYPERVISORS} " =~ " ${KATA_HYPERVISOR} " ]] && skip "Test not supported for ${KATA_HYPERVISOR}."
     # 1. Create one runc pod with the $unencrypted_image_1 image
     # We want to have one runc pod, so we pass a fake runtimeclass "runc" and then delete the runtimeClassName,
     # because the runtimeclass is not optional in new_pod_config function.
@@ -100,7 +101,6 @@ setup() {
 }
 
 @test "Test we can pull an unencrypted image inside the guest twice in a row and then outside the guest successfully" {
-    [[ " ${SUPPORTED_NON_TEE_HYPERVISORS} " =~ " ${KATA_HYPERVISOR} " ]] && skip "Test not supported for ${KATA_HYPERVISOR}."
     skip "Skip this test until we use containerd 2.0 with 'image pull per runtime class' feature: https://github.com/containerd/containerd/issues/9377"
     # 1. Create one kata pod with the $unencrypted_image_1 image and nydus annotation twice
     kata_pod_with_nydus_config="$(new_pod_config "$unencrypted_image_1" "kata-${KATA_HYPERVISOR}")"
@@ -157,7 +157,6 @@ setup() {
 }
 
 @test "Test we can pull an other unencrypted image outside the guest and then inside the guest successfully" {
-    [[ " ${SUPPORTED_NON_TEE_HYPERVISORS} " =~ " ${KATA_HYPERVISOR} " ]] && skip "Test not supported for ${KATA_HYPERVISOR}."
     skip "Skip this test until we use containerd 2.0 with 'image pull per runtime class' feature: https://github.com/containerd/containerd/issues/9377"
     # 1. Create one kata pod with the $unencrypted_image_2 image and without nydus annotation
     kata_pod_without_nydus_config="$(new_pod_config "$unencrypted_image_2" "kata-${KATA_HYPERVISOR}")"
@@ -215,6 +214,8 @@ setup() {
 
 teardown() {
     check_hypervisor_for_confidential_tests ${KATA_HYPERVISOR} || skip "Test not supported for ${KATA_HYPERVISOR}."
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
     kubectl describe pod "$pod_name"
     k8s_delete_all_pods_if_any_exists || true
 }
--- a/tests/integration/kubernetes/k8s-guest-pull-image.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image.bats
@@ -47,6 +47,11 @@ setup() {
         "io.containerd.cri.runtime-handler" \
         "kata-${KATA_HYPERVISOR}"
 
+    [[ " ${SUPPORTED_NON_TEE_HYPERVISORS} " =~ " ${KATA_HYPERVISOR} " ]] && \
+        set_metadata_annotation "$kata_pod_with_nydus_config" \
+            "io.katacontainers.config.hypervisor.image" \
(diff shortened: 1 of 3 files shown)


ChengyuZhu6 (approved):
Good job! One minor comment. Thanks @wainersm

stevenhorsman (commented):
A couple of initial comments - optional suggestion. Sorry I'm not sure why my comments didn't show up here and now I can't delete it 😿

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -169,7 +168,9 @@ setup() {
 }
 
 teardown() {
-    check_hypervisor_for_confidential_tests ${KATA_HYPERVISOR} || skip "Test not supported for ${KATA_HYPERVISOR}."
+    confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
```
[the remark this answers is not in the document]
stevenhorsman: If you make the suggested change above then I guess this needs to switch back?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -44,6 +45,11 @@ setup() {
         "io.containerd.cri.runtime-handler" \
         "kata-${KATA_HYPERVISOR}"
 
+    [[ " ${SUPPORTED_NON_TEE_HYPERVISORS} " =~ " ${KATA_HYPERVISOR} " ]] && \
+        set_metadata_annotation "$kata_pod_with_nydus_config" \
+            "io.katacontainers.config.hypervisor.image" \
+            "/opt/kata/share/kata-containers/kata-containers-confidential.img"
```
stevenhorsman, line 51: Should we move this block into `confidential_setup` replacing the `info "Need to apply image annotations"` line?

stevenhorsman (approved):
Apart from the optional suggestions I made below this looks good to me as long as the tests all pass on the CI. Thanks!

tests/integration/kubernetes/k8s-guest-pull-image.bats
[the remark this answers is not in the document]
wainersm: Good point @ChengyuZhu6 . It's a mistake, I will revert that line.

tests/integration/kubernetes/k8s-guest-pull-image.bats
wainersm: I dislike that boilerplate all over the tests, but it cannot be moved to `confidential_setup`. I explain: at the time `confidential_setup()` is called (in `setup()`) the `"$kata_pod_with_nydus_config"` file wasn't generated yet.  The config file is created with a call to `new_pod_config()` from the test method.

We should revisit and refactor the tests though.

@fidencio as part of his efforts to disable `shared_fs`, hacked the [setup.sh](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/setup.sh) to set the `cri-handler` on all [runtimeclass_workloads](https://github.com/kata-containers/kata-containers/tree/main/tests/integration/kubernetes/runtimeclass_workloads). I think this is the way to go for setting the image annotation when running on non-TEE. However, k8s-guest-pull-image.bats and k8s-measured-rootfs.bats use a different approach from the rest of the bats, which is to create a simple configuration file out of a template by calling `new_pod_config()`. If we ensure that template is proper instrumented by `setup.sh` then I think we are fine. Anyway, waiting @fidencio to publish his code to then think more deeply how to conciliate approaches.

stevenhorsman: Fair enough, I wonder if we could move the config creation as part of the setup, and remove this issue, but we can worry about that once it's more settled.

Thanks for the explaination.

fitzthum: Maybe a nontee runtime class would avoid having to make these kind of switches.

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -10,14 +10,15 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
     confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
```
fitzthum, line 13: We need this because the normal aks tests are still calling this and might run with a supported non-tee HV but without the snapshotter, right?

fitzthum (approved):
This looks fine. i think there might be some more optimizations/simplifications possible in the future.

tests/integration/kubernetes/k8s-guest-pull-image.bats
BbolroC: Yes, I would say, instead of Wainer, with an example for a workflow `run-k8s-tests-on-zvsi` which uses `devmapper`:

```
1..4
ok 1 Test we can pull an unencrypted image outside the guest with runc and then inside the guest successfully # skip None snapshotter was found but this test requires one
# [bats-exec-test:61] INFO: Need to apply image annotations
ok 2 Test we can pull a large image inside the guest # skip None snapshotter was found but this test requires one
# [bats-exec-test:61] INFO: Need to apply image annotations
ok 3 Test we can pull an unencrypted image inside the guest twice in a row and then outside the guest successfully # skip None snapshotter was found but this test requires one
# [bats-exec-test:61] INFO: Need to apply image annotations
ok 4 Test we can pull an other unencrypted image outside the guest and then inside the guest successfully # skip None snapshotter was found but this test requires one
```

BbolroC (approved):
Overall what I would like to ask about the PR has been already covered by other reviewers. It looks good to me and I could start my task for enabling the test for s390x right now. Thanks for the work. ❤️

tests/integration/kubernetes/k8s-guest-pull-image.bats
wainersm: yes, a nontee runtime class solve that. But probably introduce other problems. It's a trade-off that we should be discussion at some point

wainersm:
This PR is blocked by a bug on the runk test that's hopefully fixed in https://github.com/kata-containers/kata-containers/pull/9457

wainersm:
Only rebased to pick up #9457

Issue #8960, opened 2024-01-30
genpolicy: provide default implementation to K8sResource methods


wainersm:
**Which feature do you think can be improved?**

The K8sResource trait define methods (e.g. `get_annotations()`) that are implemented by all the implementers. If one need to change a given method implementation, it should apply to all the implementers (== maintenance burden).

See an example (with further discussion) here: https://github.com/kata-containers/kata-containers/pull/8918#discussion_r1470134982

**How can it be improved?**

Provides default implementations to the K8sResource trait that are common to all implementers. 

Cc @danmihai1

arc9693:
I would like to work on this. @danmihai1

PR #9428, opened 2024-04-08
genpolicy: implement default methods for K8sResource trait

arc9693:
This PR adds default implementations for the following methods of K8sResource trait:
- generate_policy
- serialize
- use_sandbox_pidns
- use_host_network
- get_containers
- get_container_mounts_and_storages
- get_sandbox_name
- get_annotations

Fixes: #8960

### DIFF
--- a/src/tools/genpolicy/src/config_map.rs
+++ b/src/tools/genpolicy/src/config_map.rs
@@ -9,13 +9,11 @@
 use crate::obj_meta;
 use crate::pod;
 use crate::policy;
-use crate::settings;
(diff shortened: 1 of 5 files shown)


danmihai1:
@arc9693 this looks good to me. However, do you volunteer to solve the other, similar methods too?

Thanks!

danmihai1 (approved):

sprt (approved):
I think ideally we want to split the trait into multiple ones so that the panics become compile-time errors.

Issue #9431, opened 2024-04-08
clh: isClhRunning always waits 10 seconds when clh exits


alex-matei:
# Description of problem

If clh exits, isClhRunning waits 10 seconds before returning that the process is dead.

PR #9432, opened 2024-04-08
clh: isClhRunning waits for full timeout when clh exits

alex-matei:
isClhRunning uses signal 0 to test whether the process is still alive or not. This doesn't work because the process is a direct child of the shim. Once it is dead the process becomes zombie.
Since no one waits for it the process lingers until its parent dies and init reaps it. Hence sending signal 0 in isClhRunning will always return success whether the process is dead or not.
This patch calls wait to remove the zombie from the process table before trying to send signal 0.

Fixes: #9431


danmihai1 (approved):

cmaf (approved):
@alex-matei LGTM

PR #9434, opened 2024-04-08
kata-deploy: Use ubuntu 22.04 for builds

fidencio:
Let's update the Ubuntu version, as we're seeing a whole bunch of errors both on our nightly and on the gha for different PRs.


fidencio:
Bad idea, for the components which are statically linking with GLIBC, this simply won't fly as we're using Ubuntu 20.04 as the base image. :-/

Issue #9463, opened 2024-04-11
tests: k8s: inject Agent Policy failures during CI - part 2


danmihai1:
The relevant description is in the "part 1" issue:

https://github.com/kata-containers/kata-containers/issues/9406

PR #9464, opened 2024-04-11
tests: k8s: inject agent policy failures (part2)

danmihai1:
Auto-generate the policy and then simulate attacks from the K8s control plane by modifying the test yaml files. The policy then detects and blocks those changes.

These test cases are using K8s Replication Controllers. Additional policy failures will be injected using other types of K8s resources
, e.g., using Pods and/or Jobs, in separate PRs.

Fixes: #9463


fidencio (approved):
lgtm, thanks @danmihai1!
/cc @wainersm

sprt (approved):

GabyCT (approved):

Issue #9465, opened 2024-04-12
Kernel 6.7 for confidential containers not compiled for s390x


BbolroC:
# Description of problem

The following config flags does not exist any more in kernel 6.7 which is used for confidential-kernel:

- CONFIG_TN3270_TTY (https://github.com/torvalds/linux/commit/9603cb334a7dc30fb544a3579bcbf23ae5b04f8f)
- CONFIG_S390_AP_IOMMU (https://github.com/torvalds/linux/commit/ac92c27935c759ff8b3cc9f761b086b15145c901)

We need to get the kernel compiled with a different set of flags for a different kernel version.

PR #9469, opened 2024-04-15
kernel: Adjust s390x config for confidential containers

BbolroC:
`CONFIG_TN3270_TTY` and `CONFIG_S390_AP_IOMMU` are dropped for s390x in 6.7.x which is used for a confidential kernel.
But they are still used for a vanilla kernel. So we need to add them to the whitelist.

Fixes: #9465


zvonkok (approved):
LGTM thanks!

stevenhorsman (approved):
LGTM. Thanks

Issue #9470, opened 2024-04-15
version: name and version missing for CoCo {image,initrd} for s390x


BbolroC:
# Description of problem

There is no name and version specified for `assets.{image,initrd}.s390x.confidential` in versions.yaml. This breaks the confidential build for the platform. We need to configure the spec when #9469 is merged.

PR #9471, opened 2024-04-15
version: Add coco name and version for {image,initrd} for s390x

BbolroC:
In order to build a coco {image,initrd}, it is required to specify its name and version in versions.yaml. This commit is to add the configuration for them, respectively.

Fixes: #9470


stevenhorsman (approved):
This looks consistent with the other platforms. Thanks @BbolroC

Issue #9472, opened 2024-04-15
version: add initrd, image NVIDIA sections


zvonkok:
For initrd and image, the related NVIDIA will not use the default targets and we will pin them to a specific release.

PR #9473, opened 2024-04-15
version: add initrd, image NVIDIA sections

zvonkok:
Fixes: #9472

For initrd and image, the related NVIDIA will not use the default targets and we will pin them to a specific release.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -123,6 +123,12 @@ assets:
       aarch64:
         name: &default-image-name "ubuntu"
         version: &default-image-version "latest"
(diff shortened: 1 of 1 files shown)


BbolroC (approved):
LGTM, thanks @zvonkok !

stevenhorsman (commented):
Ignorant question here - I don't think we are building an nvidia-gpu variant of the cc image at the moment? Is this the foundation for later adding this, or am I missing something?

stevenhorsman (approved):
LGTM. Thanks

Issue #9478, opened 2024-04-15
image: Add suffix to image or initrd depending on the NVIDIA driver version


zvonkok:
We want to keep track of the driver versions build during initrd/image build so update the artifact_name after the fact.

PR #9480, opened 2024-04-15
image: Add suffix to image or initrd depending on the NVIDIA driver version

zvonkok:
Add suffix to image or initrd depending on the NVIDIA driver version 

Fixes: #9478

We want to keep track of the driver versions build during initrd/image build so update the artifact_name after the fact.

### DIFF
--- a/tools/packaging/guest-image/build_image.sh
+++ b/tools/packaging/guest-image/build_image.sh
@@ -37,6 +37,7 @@ build_initrd() {
 	info "initrd os: $os_name"
 	info "initrd os version: $os_version"
 	make initrd \
+		VARIANT="${image_initrd_suffix}" \
 		DISTRO="$os_name" \
 		DEBUG="${DEBUG:-}" \
 		OS_VERSION="${os_version}" \
@@ -49,7 +50,7 @@ build_initrd() {
 		COCO_GUEST_COMPONENTS_TARBALL="${COCO_GUEST_COMPONENTS_TARBALL:-}" \
 		PAUSE_IMAGE_TARBALL="${PAUSE_IMAGE_TARBALL:-}"
 
-	if [[ "$artifact_name" == *"nvidia-gpu"* ]]; then
+	if [[ "${image_initrd_suffix}" == "nvidia-gpu"* ]]; then
 		nvidia_driver_version=$(cat "${builddir}"/initrd-image/*/nvidia_driver_version)
 		artifact_name=${artifact_name/.initrd/"-${nvidia_driver_version}".initrd}
 	fi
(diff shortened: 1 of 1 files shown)


tools/packaging/guest-image/build_image.sh
```diff
@@ -48,7 +48,13 @@ build_initrd() {
 		PULL_TYPE="${PULL_TYPE:-default}" \
 		COCO_GUEST_COMPONENTS_TARBALL="${COCO_GUEST_COMPONENTS_TARBALL:-}" \
 		PAUSE_IMAGE_TARBALL="${PAUSE_IMAGE_TARBALL:-}"
-	mv "kata-containers-initrd.img" "${install_dir}/${artifact_name}"
+
+	if [[ "$artifact_name" == *"nvidia-gpu"* ]]; then
+		nvidia_driver_version=$(cat "${builddir}"/initrd-image/*/nvidia_driver_version)
```
BbolroC, line 53: I was wondering where `nvidia_driver_version` is created. Thanks!

zvonkok: In the rootfs build, I need to follow up with another ton of PRs so all of this makes sense. I hadn't had any reviews ... :)

wainersm: Hi @zvonkok !

Why don't you have the caller to pass the driver version via `--image-initrd-suffix` option? This way we avoid special treatments, keeping this script generic.

zvonkok: Because we do not know upfront what the latest driver version is in the Ubuntu repos.

tools/packaging/guest-image/build_image.sh
```diff
@@ -37,6 +37,7 @@ build_initrd() {
 	info "initrd os: $os_name"
 	info "initrd os version: $os_version"
 	sudo -E PATH="$PATH" make initrd \
+		VARIANT="${image_initrd_suffix}" \
```
wainersm, line 40: Hi @zvonkok !

The other two commits I could understand and they don't need more context. This one I didn't get. Perhaps it will make sense with the following PRs? I'm afraid of getting it merged, devs start using this VARIANT thing in a different context then clashes with your original intention. So maybe either put this in you following PRs or simply stuff this PR with more commits...

tools/packaging/guest-image/build_image.sh
wainersm: Got it!

fidencio:
Just rebased.

fidencio (approved):
lgtm, thanks @zvonkok!

BbolroC (approved):
LGTM, thanks @zvonkok !

Apokleos (approved):
Thx @zvonkok LGTM

Issue #9467, opened 2024-04-12
Enable guest-pull tests for non-TEE for s390x


BbolroC:
**Which feature do you think can be improved?**

In line with #9424, it is time to enable the guest-pull test for non-TEE for s390x. We do have confidential artifacts (kernel, image and initrd) for the platform. This issue is to summarise some of the environmental adjustment required for the enablement. They were identified during the testing trials as follows:

- [x] Kernel 6.7 should be compiled: https://github.com/kata-containers/kata-containers/issues/9465
- [x] Configure a coco name and version for {image,initrd} https://github.com/kata-containers/kata-containers/issues/9470
- [x] OPA should be built from the source: https://github.com/kata-containers/kata-containers/issues/9466

Further issues could be added to this issue as the enablement goes on.

PR #9489, opened 2024-04-16
rootfs: Make OPA build working in docker for s390x and pp…

BbolroC:
The PR is to make the OPA build from source working in a docker builder `ubuntu-rootfs-osbuilder`. 
To achieve the goal, the configuration is changed as follows:

- Switch the make target to `ci-build-linux-static` not triggering docker-in-docker build
- Install go in the builder image for s390x and ppc64le

Fixes: #9466

Signed-off-by: Hyounggyu Choi <Hyounggyu.Choi@ibm.com>

PR #9494, opened 2024-04-17
CC: Enable guest-pull tests on non-TEE for s390x

BbolroC:
This commit is to add a new CI job to run-k8s-tests-on-zvsi.yaml.
Why the job is not configured in run-kata-coco-tests.yaml by having it integrated with `run-k8s-tests-coco-nontee` is:

- It uses k3s instead of AKS
- It runs on a self-hosted runner

These differences make the integrated job not easy to read and maintain when it comes to incorporating other platforms in the near future.

The verification for non-TEE has been done via a private self-hosted runner at https://github.com/BbolroC/kata-containers/actions/runs/8782763328

This PR also includes an adjustment of the build procedure for `boot-image-se` and updates the document accordingly. The verification for TEE (Secure Execution) from the internal test is at https://github.com/kata-containers/kata-containers/actions/runs/8795055047/job/24144779614

Fixes: #9467

**IMPORTANT NOTE**: This PR should not be merged before #9489.

### DIFF
--- a/.github/workflows/run-k8s-tests-on-zvsi.yaml
+++ b/.github/workflows/run-k8s-tests-on-zvsi.yaml
@@ -31,18 +31,31 @@ jobs:
           - qemu
         snapshotter:
           - devmapper
+          - nydus
         k8s:
           - k3s
+        include:
+          - snapshotter: devmapper
+            pull-type: default
+            using-nfd: true
+            deploy-cmd: configure-snapshotter
+          - snapshotter: nydus
+            pull-type: guest-pull
+            using-nfd: false
+            deploy-cmd: deploy-snapshotter
     runs-on: s390x-large
     env:
       DOCKER_REGISTRY: ${{ inputs.registry }}
       DOCKER_REPO: ${{ inputs.repo }}
       DOCKER_TAG: ${{ inputs.tag }}
       PR_NUMBER: ${{ inputs.pr-number }}
+      GH_PR_NUMBER: ${{ inputs.pr-number }}
+      KATA_HOST_OS: "ubuntu"
       KATA_HYPERVISOR: ${{ matrix.vmm }}
       KUBERNETES: "k3s"
+      PULL_TYPE: ${{ matrix.pull-type }}
       SNAPSHOTTER: ${{ matrix.snapshotter }}
-      USING_NFD: "true"
+      USING_NFD: ${{ matrix.using-nfd }}
       TARGET_ARCH: "s390x"
     steps:
       - name: Take a pre-action for self-hosted runner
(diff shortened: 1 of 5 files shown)


.github/workflows/run-k8s-tests-on-zvsi.yaml
```diff
@@ -78,3 +78,62 @@ jobs:
...
+          - nydus
+        pull-type:
+          - guest-pull
+    runs-on: s390x-large
+    env:
+      DOCKER_REGISTRY: ${{ inputs.registry }}
+      DOCKER_REPO: ${{ inputs.repo }}
+      DOCKER_TAG: ${{ inputs.tag }}
+      GH_PR_NUMBER: ${{ inputs.pr-number }}
+      KATA_HOST_OS: ${{ matrix.host_os }}
```
stevenhorsman, line 98: There doesn't seem to be a `host_os` matrix parameter, so should this `env` be set to `ubuntu` (I guess?) directly?

stevenhorsman (commented):
The workflow for `run-k8s-tests-coco-nontee` and `run-k8s-tests` looks pretty much identical apart from the snapshotter and pull type, so I wonder if you could add the nydus snapshotter to the original snapshotter matrix and then add the pull-type to be guest-pull there to save on the duplication?

.github/workflows/run-k8s-tests-on-zvsi.yaml
BbolroC: oh, good catch. I will update the PR. Thanks!

BbolroC:
> The workflow for `run-k8s-tests-coco-nontee` and `run-k8s-tests` looks pretty much identical apart from the snapshotter and pull type, so I wonder if you could add the nydus snapshotter to the original snapshotter matrix and then add the pull-type to be guest-pull there to save on the duplication?

Yeah, I was thinking about that too. But it would be a matter of tee or non-tee. In the same sense, `run-k8s-tests-coco-nontee` and `run-k8s-tests` for `x86_64` are almost the same, but duplicated. I am sure if the deduplication is a way to go. What do you think @wainersm ?

.github/workflows/run-k8s-tests-on-zvsi.yaml
```diff
@@ -31,18 +31,31 @@ jobs:
...
       DOCKER_REPO: ${{ inputs.repo }}
       DOCKER_TAG: ${{ inputs.tag }}
       PR_NUMBER: ${{ inputs.pr-number }}
+      GH_PR_NUMBER: ${{ inputs.pr-number }}
+      KATA_HOST_OS: "ubuntu"
       KATA_HYPERVISOR: ${{ matrix.vmm }}
       KUBERNETES: "k3s"
+      PULL_TYPE: ${{ matrix.pull-type }}
       SNAPSHOTTER: ${{ matrix.snapshotter }}
-      USING_NFD: "true"
```
wainersm, line 45: I suspect this `USING_NFD` is bogus. IIRC @fidencio was preparing the workflows for the NFD (Node Feature Discovery) but it's unfinished. Please correct if I am wrong.

wainersm (approved):
Despite the USING_NFD comment that in reality is spread in many workflows; this looks good to me.

.github/workflows/run-k8s-tests-on-zvsi.yaml
BbolroC: Yeah, we could deal with it all at once later when it is finished. Thanks!

Issue #9500, opened 2024-04-17
Fix mandatory Issue - PR dependency


zvonkok:
We've discussed this over and over. Let's try to get to an agreement here.

I will use this issue to remove the mandatory Issue - PR dependency.

PR #9501, opened 2024-04-17
kata: Remove check for "Fixes" in PR

zvonkok:
We've discussed this over and over. Let's try to get to an agreement here. I will use this issue to remove the mandatory Issue - PR dependency.

Fixes: #9500

### DIFF
--- a/.github/workflows/commit-message-check.yaml
+++ b/.github/workflows/commit-message-check.yaml
@@ -86,17 +86,6 @@ jobs:
         error: 'Body line too long (max 150)'
         post_error: ${{ env.error_msg }}
 
-    - name: Check Fixes
-      if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') && ( success() || failure() ) }}
-      uses: tim-actions/commit-message-checker-with-regex@v0.3.1
-      with:
-        commits: ${{ steps.get-pr-commits.outputs.commits }}
-        pattern: '\s*Fixes\s*:?\s*(#\d+|github\.com\/kata-containers\/[a-z-.]*#\d+)|^\s*release\s*:'
-        flags: 'i'
-        error: 'No "Fixes" found'
-        post_error: ${{ env.error_msg }}
-        one_pass_all_pass: 'true'
-
     - name: Check Subsystem
       if: ${{ !contains(github.event.pull_request.labels.*.name, 'force-skip-ci') && ( success() || failure() ) }}
       uses: tim-actions/commit-message-checker-with-regex@v0.3.1
--- a/.github/workflows/move-issues-to-in-progress.yaml
+++ b/.github/workflows/move-issues-to-in-progress.yaml
@@ -62,11 +62,10 @@ jobs:
             grep -v "^\#"  |\
             cut -d';' -f3 || true)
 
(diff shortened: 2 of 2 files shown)


stevenhorsman:
I think there are a fair few people who have shown interest in the past in this, so I'm wary of turning it into a bunfight, but @kata-containers/architecture-committee @fidencio @pmores @beraldoleal @jodh-intel should all have a chance to have their say

fidencio:
> /ok-to-test

You have to manually add the label :-/

beraldoleal (approved):
Unless I'm missing something, that is easy yes to me. Thanks @zvonkok

fidencio:
> I thought that the problem @fidencio raised on Slack was about dependencies. Aren't you also removing the check for our own PRs here? Is _that_ what we want?

The full removal was discussed as part of the AC meeting, on April 16th, and I think Zvonko is opening the conversation about it here.

pmores (approved):
The existing practice (that I adopted as well) seems to be mostly just to create an issue that's largely or completely empty, just to placate the Fixes check.  Considering that, I can't see what useful purpose the check could serve and what we'd lose if we did away with it.  Creating the mandatory issue becomes just a useless annoying step in PR submission process which is kind of involved already.

That's my view anyway.

gkurz (approved):
I share @pmores's view on this.

zvonkok:
We're keeping the other tooling around, so if a customer creates an issue, we can still link it to a PR, and I can be automatically set to in progress and closed.

sprt (approved):
Echoing @pmores's opinion. For quick fixes, this check is only adding overhead. Now of course this should not preclude opening issues to discuss more involved problems and their resolution, and we should make sure that such problems have an issue assigned to them.

Food for thought: should we have a bot that reminds folks to open an issue in case a PR is closed without being merged? I suspect such cases potentially denote an issue that was ultimately not addressed, so worth a note?

zvonkok:
@sprt Or someone just closed the PR due to complexity, irrelevance, or other circumstances that a bot can not deduce. If it is important one will create an issue. I have closed >10 PRs in the past, and I want to avoid seeing an issue appear for them automatically. People are often creating PRs just for testing the CI or other contexts' without needing an issue per se.

pmores:
> Now of course this should not preclude opening issues to discuss more involved problems and their resolution, and we should make sure that such problems have an issue assigned to them.

Discussing problems in issues ahead of time doesn't seem to be a huge part of kata project workflow/culture, at least from what I've seen.  Few issues created to discuss possible solutions to a problem get enough participation to be relevant, that's my experience at least, and I don't think I can actually remember any that was followed up by a PR implementing an agreed-upon solution.

This means (I mean, if my experience is not completely off) that we don't really have too much of a problem with PR's not linking back to genuine issues that contain discussion which spawned the PR in the first place.

All this might be considered unfortunate, and I for one would be very much in favour of using issues to discuss before a PR is made, but how to change it is another discussion and not related to the present PR.

fidencio (approved):
lgtm, thanks!

I've also spoken with @jodh-intel Today and he mentioned although he prefers keeping it, he has no strong motivation to avoid this being removed either,.

With that said, I think we can go ahead with this.

Issue #9502, opened 2024-04-17
gha: Update the start date of the stale bot


stevenhorsman:
In [November](https://github.com/kata-containers/kata-containers/pull/8348) we added stalebot workflow to auto close PRs that had not action for 180 days (with a 7 day warning). That seems to have helped the PR backlog stop growing (IIRC we had just over 200 then and are at 184 as I type). However it looks like there was an incorrect assumption in the workflow, so we started the clean up in May 2023, so if we removed that then an extra ~63 PRs would be warned about closing.

Let address this now, as we've seen success with the stale bot on PRs

PR #9503, opened 2024-04-17
gha: stale: Remove the start-date

stevenhorsman:
As documented in https://github.com/actions/stale?tab=readme-ov-file#start-date
> The start date is used to ignore the issues and pull requests created before the start date.
> Particularly useful when you wish to add this stale workflow on an existing repository
> and only wish to stale the new issues and pull requests.

As we don't want need to treat PRs older than May 2023 as a special case, then remove this option.

Fixes: #9502

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -10,7 +10,6 @@ jobs:
     steps:
       - uses: actions/stale@v8
         with:
-          start-date: '2023-05-01T00:00:00Z'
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman!

At first I misinterpreted the `start-date` as "anything older than this date will be affected", but that was a clear overlook from my side.

danmihai1 (approved):

sprt (approved):

Issue #9507, opened 2024-04-17
GHA: newly added runners for s390x failing for run-cri-containerd and run-k8s-tests


BbolroC:
# Description of problem

New 2 runners for s390x have been recently registered to the project, but it seems CI jobs `run-k8s-tests` and `run-cri-containerd` do not work on the runners due to the limitation on resources (4 vCPUs). 

For `run-k8s-tests`, a runtime class does not appear in a given time (even in 240s as AKS) and the test complains as follows:

```
1..1
not ok 1 Running with postStart and preStop handlers
# (in test file k8s-attach-handlers.bats, line 37)
#   `kubectl create -f "${yaml_file}"' failed
# [bats-exec-test:59] INFO: k8s configured to use runtimeclass
# Error from server (Forbidden): error when creating "/home/ansible/gha-runner/_layout/_work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/test-lifecycle-events.yaml": 
pods "handlers" is forbidden: pod rejected: RuntimeClass "kata" not found
# Error from server (NotFound): pods "handlers" not found
# Error from server (NotFound): pods "handlers" not found
```

and for `run-cri-containerd`, got the timeout:

```
[integration-tests.sh:556] INFO: Two loop devices, /dev/loop4 and /dev/loop5, are created.
[integration-tests.sh:282] INFO: restart containerd service
Image is up to date for sha256:6fed366a7bcbacc0ef09ca2d17065e5859d416094acae6bfeea186d67c37b208
E0418 11:25:57.924184  657811 remote_runtime.go:193] "RunPodSandbox from runtime service failed" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded"
time="2024-04-18T11:25:57Z" level=fatal msg="run pod sandbox: rpc error: code = DeadlineExceeded desc = context deadline exceeded"
```

We need to run the failing tests on the existing 3 runners (8 vCPUs) only by giving them a special label `s390x-large`.

PR #9508, opened 2024-04-17
gha: Make integration tests for s390x run on s390x-large runners

BbolroC:
This is to make a workflow `run-k8s-tests` and `run-cri-containerd` (s390x and zvsi) run only on the runners labeled by `s390x-large`.

FYI: here are the registered runners for s390x:

image

Fixes: #9507


justxuewei (approved):
Lgtm, thanks!

Issue #9388, opened 2024-04-01
agent: use the regorus crate instead of opa


danmihai1:
regorus is written in Rust and therefore it can run in-proc - faster - without the need to use the OPA REST APIs.

PR #9510, opened 2024-04-18
agent: use regorus instead of opa

danmihai1:
Implement Agent Policy using the regorus crate instead of the OPA daemon.

The OPA daemon will be removed from the Guest rootfs in a future PR.

The first commit from this PR is locking the anyhow crate version to 1.0.58 because:
- Versions between 1.0.59 - 1.0.76 have not been tested yet using Kata CI. (However, "make test" works fine for the
  Kata Agent when using these versions)
- Versions 1.0.77 or newer fail during "make test" - see https://github.com/kata-containers/kata-containers/issues/9538.
So. this version lock avoids an undesirable anyhow version update when making the Regorus changes.

Fixes: #9388


src/agent/Cargo.toml
```diff
@@ -73,14 +73,12 @@ clap = { version = "3.0.1", features = ["derive"] }
...
-reqwest = { version = "0.11.14", optional = true }
-# The "vendored" feature for openssl is required for musl build
-openssl = { version = "0.10.54", features = ["vendored"], optional = true }
-
 # Image pull/decrypt
 image-rs = { git = "https://github.com/confidential-containers/guest-components", rev = "ca6b438", default-features = true, optional = true }
+openssl = { version = "0.10.54", features = ["vendored"], optional = true }
+
+# Agent Policy
+regorus = { git = "https://github.com/danmihai1/regorus", rev = "5f5bb3e", default-features = false, features = ["arc", "regex"], optional = true }
```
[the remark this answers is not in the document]
danmihai1: Yes, that's the main reason why this PR is marked as work in progress. Please stay tuned.

src/agent/Cargo.toml
```diff
@@ -73,14 +73,12 @@ clap = { version = "3.0.1", features = ["derive"] }
...
-reqwest = { version = "0.11.14", optional = true }
-# The "vendored" feature for openssl is required for musl build
-openssl = { version = "0.10.54", features = ["vendored"], optional = true }
-
 # Image pull/decrypt
 image-rs = { git = "https://github.com/confidential-containers/guest-components", rev = "ca6b438", default-features = true, optional = true }
+openssl = { version = "0.10.54", features = ["vendored"], optional = true }
+
+# Agent Policy
+regorus = { git = "https://github.com/danmihai1/regorus", rev = "a615f40", default-features = false, features = ["arc", "regex"], optional = true }
```
anakrish, line 81: It is better to fetch regorus from crates.io once I push a release that meets this PR's requirements. Similar to what `trustee` does
https://github.com/confidential-containers/trustee/blob/dc01f454264fb4350e5f69eba05683a9a1882c41/Cargo.toml#L37

danmihai1: @anakrish , CI tests are now passing using:

git = "https://github.com/microsoft/regorus", rev = "316f3a76"

So, I think this would be a good time to publish that into crates.io.

src/agent/Cargo.toml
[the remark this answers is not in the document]
danmihai1: @zvonkok , we are now using here https://github.com/microsoft/regorus. 

Also, Anand is working on publishing that to crates.io.

stevenhorsman:
Hi @danmihai1 - Thanks for the PR. With this change can we remove the OPA logic from rootfs.sh as well that installed the OPA binary: https://github.com/kata-containers/kata-containers/blob/42a79801f3caaaf08a68ec69b46f506660638249/tools/osbuilder/rootfs-builder/rootfs.sh#L740-L806

We have also used the distinction between the agent with and without policy as `agent` and `agent_opa` which feels wrong now and that `agent_policy` might be better, but I'm not sure if we want to attack that in this PR yet :)

danmihai1:
> ~Hi @danmihai1 - Thanks for the PR. With this change can we remove the OPA related code from the versions.yaml, rootfs.sh that installs the OPA binary~:
> 
> https://github.com/kata-containers/kata-containers/blob/42a79801f3caaaf08a68ec69b46f506660638249/tools/osbuilder/rootfs-builder/rootfs.sh#L740-L806
> 
> Sorry, I just saw the PR comment
> 
> > The OPA daemon will be removed from the Guest rootfs in a future PR.
> 
> Is this in case we want to backtrack, or just to keep this PR smaller?
> 
> I think we will still need to keep the default rego policy written to the rootfs, but I'm trying to also work out if, given that it is a single file, there is a case to remove the `"${AGENT_POLICY}" == "yes"` guard and have it installed in all rootfs', but am not sure if that are many points for, or against that.
> 
> We have also used the distinction between the agent with and without policy as `agent` and `agent_opa` which feels wrong now and that `agent_policy`/`agent_rego` might be better, but I'm not sure if we want to attack that in this PR yet :)

Indeed, I would be happy to remove and/or replace all things related to OPA - but I think we can address them in separate PRs, to sharpen the focus of each PR.

Related to the agent vs. agent-opa distinction you mentioned: we should also consider adding Rego policy support into the vanilla agent and killing agent-opa. Users could still build their agent without Rego support, but CI might not want to worry about that distinction. 

Those seem to me like good things to debate in separate PRs.

fidencio:
@danmihai1, on TDX we're getting the following error:
```
# [bats-exec-test:190] INFO: Deleting /home/github-runner/actions-runner/_work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/k8s-policy-job-incorrect.yaml
not ok 6 Policy failure: unexpected readOnlyRootFilesystem
```

danmihai1:
> @danmihai1, on TDX we're getting the following error:
> 
> ```
> # [bats-exec-test:190] INFO: Deleting /home/github-runner/actions-runner/_work/kata-containers/kata-containers/tests/integration/kubernetes/runtimeclass_workloads_work/k8s-policy-job-incorrect.yaml
> not ok 6 Policy failure: unexpected readOnlyRootFilesystem
> ```

Thanks! It appears that I broke this a while ago. TDX was probably failing before running these genpolicy tests, so we didn't notice my older mistake until now. I will go ahead and clean this up.

stevenhorsman:
> Related to the agent vs. agent-opa distinction you mentioned: we should also consider adding Rego policy support into the vanilla agent and killing agent-opa. Users could still build their agent without Rego support, but CI might not want to worry about that distinction.

I floated this on AC meeting and there wasn't big push back, but maybe we can add it to the agenda and properly discuss it next week?

danmihai1:
> > Related to the agent vs. agent-opa distinction you mentioned: we should also consider adding Rego policy support into the vanilla agent and killing agent-opa. Users could still build their agent without Rego support, but CI might not want to worry about that distinction.
> 
> I floated this on AC meeting and there wasn't big push back, but maybe we can add it to the agenda and properly discuss it next week?
> 
> And to clarify this can also come after this PR too!

Sounds good - I will join the next meeting and bring information about advantages vs. disadvantages. Spoiler alert: I think that's mostly about reducing CI complexity vs. a larger CI Agent binary.

tests/integration/kubernetes/tests_common.sh
```diff
@@ -314,5 +314,5 @@ wait_for_blocked_request() {
 
 	command="kubectl describe pod ${pod} | grep \"${endpoint} is blocked by policy\""
 	info "Waiting ${wait_time} seconds for: ${command}"
-	waitForProcess "${wait_time}" "$sleep_time" "${command}" || return 1
+	waitForProcess "${wait_time}" "$sleep_time" "${command}" >/dev/null 2>/dev/null
```
fidencio, line 317: I don't fully get this change here, @danmihai1

fidencio (approved):
lgtm, thanks @danmihai1!

This simplifies things a lot and will also reduce significantly the size of the confidential image / initrd (with your planned follow-ups).

Also, log messages on regorus seem to be waaay better than the ones coming from OPA. \o/

tests/integration/kubernetes/tests_common.sh
danmihai1: > I don't fully get this change here, @danmihai1

The improved diagnosability when using Regorus this way comes at the expense of messier kubectl output. For examples of messy "kubectl describe <pod>", search for "CreateContainerRequest is blocked by policy" in https://productionresultssa10.blob.core.windows.net/actions-results/834f1908-e474-4be2-a0cf-8c3b1382901d/workflow-job-run-2b16350a-daeb-5112-dd9c-126ea1f33816/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-04-23T18%3A41%3A08Z&sig=rX2I1EZqLm%2BevdBFYESD3yyFq%2BuswZbpu9SeOJ%2BLCxQ%3D&sp=r&spr=https&sr=b&st=2024-04-23T18%3A31%3A03Z&sv=2021-12-02

Before this change, we were displaying the same "CreateContainerRequest is blocked by policy" messy output at least two times:
- First time from waitForProcess
- A second time from teardown() of each test

So, I decided to reduce some of that noise by keeping just of teardown() output and hiding here the redundant waitForProcess output.

fidencio: Makes sense, thanks for the explanation, the approval is still valid. :-)

src/agent/src/main.rs
```diff
@@ -379,14 +379,9 @@ async fn start_sandbox(
     #[cfg(feature = "guest-pull")]
     image::set_proxy_env_vars().await;
 
-    // - When init_mode is true, enabling the localhost link during the
-    //   handle_localhost call above is required before starting OPA with the
-    //   initialize_policy call below.
-    // - When init_mode is false, the Policy could be initialized earlier,
-    //   because initialize_policy doesn't start OPA. OPA is started by
-    //   systemd after localhost has been enabled.
+    // TODO: initialize earlier.
```
stevenhorsman, line 382: Is this initialise earlier comment for a future PR and related to the init data discussions?

tests/integration/kubernetes/k8s-policy-job.bats
```diff
@@ -9,7 +9,7 @@ load "${BATS_TEST_DIRNAME}/../../common.bash"
 load "${BATS_TEST_DIRNAME}/tests_common.sh"
 
 setup() {
-    policy_tests_enabled || skip "Policy tests are disabled."
+    auto_generate_policy_enabled || skip "Auto-generated policy tests are disabled."
```
stevenhorsman, line 12: Do these changes mean that the tests with only run against mariner now due to:
https://github.com/kata-containers/kata-containers/blob/89c85dfe846d661dba0f610260a8f69c61aca967/tests/integration/kubernetes/gha-run.sh#L261-L269

whereas before they were running on snp, sev & tdx as well due to
https://github.com/kata-containers/kata-containers/blob/89c85dfe846d661dba0f610260a8f69c61aca967/tests/integration/kubernetes/tests_common.sh#L266-L271
?

Obviously I'd like to see them running on all platforms, but that can come as part of the future discussion about enabling policy for the vanilla agent :)

stevenhorsman (approved):
A couple of questions from me, and a few future points that we've agreed can come as follow-on tasks, but apart from that this looks good to me and I think switching to a rust crate with direct API call over http requests to a separate binary is a big win in terms of code and then the size saving on top means I've very happy to see this change. Thanks to Dan and the other regorus devs!

tests/integration/kubernetes/k8s-policy-job.bats
danmihai1: @stevenhorsman 

- Mariner, SEV, SNP and TDX Hosts have been testing using just simple, hard-coded policies from https://github.com/kata-containers/kata-containers/tree/main/src/kata-opa since September (see https://github.com/kata-containers/kata-containers/commit/bf21411e90327b4c3192f1c1debe9cb0b1c4dcb5).

- Mariner started testing a more sophisticated, auto-generated, policy since February (https://github.com/kata-containers/kata-containers/commit/0de407f8b73ed2caf69205b6b859fbfa45a084f2). This testing took place just on Mariner because:
  - It's easier for me to connect and debug on Mariner Hosts when the tests fail (see below for the diagnosability improvements needed to address this).
  - Even while testing a single type of Host, developing tests and the genpolicy tool required by these tests, and passing CI has been challenging. *The policy tests are finally in pretty good shape just since last week*.

- Starting 2 weeks ago, I *accidentally* enabled new genpolicy tests for SEV, SNP and TDX Hosts too (https://github.com/kata-containers/kata-containers/commit/2252490a9688814c2f73b7bba1beb8d59973b33e)
  - SEV, SNP and TDX Hosts were failing CI for unrelated reasons, so none of us noticed this accident until Fabiano noticed yesterday.
  - Here, I am just fixing this accident: https://github.com/kata-containers/kata-containers/pull/9510/commits/e5c3f5fa9b4017b4892c3855701347ebbb8da9f5
  - So, these tests were enabled for a couple of weeks on SEV, SNP and TDX Hosts but never passed on these unexpected Hosts.

- Support for testing SEV, SNP and TDX and any other Hosts that might be interested in policy is coming up very soon. There were two requirements before enabling these tests:
  - Tests that are stable and have pretty good coverage. We satisfied this requirement just last week (https://github.com/kata-containers/kata-containers/commit/2c4d1ef76bf640d644dba26c97642afe49ddec05).
  - Policy logs that we can use to investigate failures, without me having to connect and debug on the Hosts. We'll finally get those logs from the current PR (https://github.com/kata-containers/kata-containers/pull/9510)

So, proper testing for SEV, SNP and TDX Hosts is coming up soon, but it depends on the current PR (https://github.com/kata-containers/kata-containers/pull/9510)

src/agent/src/main.rs
danmihai1: This was just a reminder for me that we *could* initialize the Policy module earlier, if we wanted. For example, we could use Policy data instead of some of the config.rs data.

But at this point I am not aware of anyone planning to use Policy data instead of config.rs data, so I will remove this comment/reminder.

tests/integration/kubernetes/k8s-policy-job.bats
stevenhorsman: Sounds good to me :)

Issue #9034, opened 2024-02-06
cloud-hypervisor runtime-rs: k8s tests not working with devicemapper


GabyCT:
Seems like the k8s tests are not working with cloud hypervisor runtime-rs with devicemapper as all of them are failing in the GHA CI.
Here are some examples of the failures that we can see
```
not ok 1 Check capabilities of pod
# (in test file k8s-caps.bats, line 34)
#   `kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"' failed
# [bats-exec-test:[55](https://github.com/kata-containers/kata-containers/actions/runs/7717165995/job/21036574048?pr=8628#step:8:56)] INFO: k8s configured to use runtimeclass
# pod/pod-caps created
# error: timed out waiting for the condition on pods/pod-caps
# expected capability mask:
# CapPrm.*..0..5fb$
# observed:
# Error from server (BadRequest): container "test-container" in pod "pod-caps" is waiting to start: ContainerCreating
# error: unable to upgrade connection: container not found ("test-container")
# pod "pod-caps" deleted
Error: Process completed with exit code 1.
```
see https://github.com/kata-containers/kata-containers/actions/runs/7717165995/job/21036574048?pr=8628
```
not ok 1 Scale nginx deployment
# (in test file k8s-scale-nginx.bats, line 25)
#   `kubectl wait --for=condition=Available --timeout=$timeout deployment/${deployment}' failed
# [bats-exec-test:55] INFO: k8s configured to use runtimeclass
# deployment.apps/nginx-deployment created
# error: timed out waiting for the condition on deployments/nginx-deployment
# deployment.apps "nginx-deployment" deleted
# Error from server (NotFound): services "nginx-deployment" not found****
```
see https://github.com/kata-containers/kata-containers/actions/runs/7718625512/job/21041124800?pr=8628
```
not ok 1 Parallel jobs
# (in test file k8s-parallel.bats, line 32)
#   `kubectl wait --for=condition=Ready --timeout=$timeout pod -l jobgroup=${job_name}' failed
# [bats-exec-test:[55](https://github.com/kata-containers/kata-containers/actions/runs/7717165995/job/21036573742?pr=8628#step:8:56)] INFO: k8s configured to use runtimeclass
# job.batch/process-item-test1 created
# job.batch/process-item-test2 created
# job.batch/process-item-test3 created
# NAME                 COMPLETIONS   DURATION   AGE
# process-item-test1   0/1           0s         0s
[...]

PR #9517, opened 2024-04-18
runtime-rs: Update storage source for pci block devices

amshinde:
In case of block devices using virtio-block, we need to pass the pci-path as the storage source field to the agent. Current the virt-path is being passed which works just for mmio block devices.
In the future when support is added for scsi, block-ccw and pmem devices, the storage source would need to be handled accordingly.

Fixes: #9034

### DIFF
--- a/src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
+++ b/src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
@@ -16,6 +16,9 @@ use hypervisor::{
     },
     BlockConfig,
 };
+use kata_types::config::hypervisor::{
+    VIRTIO_BLK_CCW, VIRTIO_BLK_MMIO, VIRTIO_BLK_PCI, VIRTIO_PMEM, VIRTIO_SCSI,
+};
 use kata_types::mount::Mount;
 use nix::sys::stat::{self, SFlag};
 use std::fs;
@@ -48,7 +51,7 @@ impl BlockRootfs {
         let block_device_config = &mut BlockConfig {
             major: stat::major(dev_id) as i64,
             minor: stat::minor(dev_id) as i64,
-            driver_option: block_driver,
+            driver_option: block_driver.clone(),
             ..Default::default()
         };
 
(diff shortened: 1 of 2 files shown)


jodh-intel (approved):
Thanks, @amshinde!

lgtm :smile:

src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
```diff
@@ -67,8 +68,22 @@ impl BlockRootfs {
         let mut device_id: String = "".to_owned();
         if let DeviceType::Block(device) = device_info {
             storage.driver = device.config.driver_option;
-            storage.source = device.config.virt_path;
             device_id = device.device_id;
+
+            // TODO: The following would work for blk-pci and mmio drivers.
+            // The source for storage drivers needs to be updates for other drivers like
+            // VIRTIO_SCSI, VIRTIO_BLK_CCW and VIRTIO_PMEM once support for them is added.
+            if block_driver == VIRTIO_BLK_PCI {
```
Apokleos, line 76: There will be more drivers to be added here, right ?
if so, IMO, we should take the future possible driver into account.
And I prefer match xx {}, instead if else here.

Apokleos (commented):
Thx @amshinde I left a comment here  talking about how to the code easier to extend for other drivers.
Anyway, but the comment, else LGTM !

Apokleos (changes requested):
As two approved, I think we should discuss this code, So I have to do such request changes here

src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
jodh-intel: @Apokleos - I understand your view. However, @amshinde did mention that there will need to be future changes at the top of the PR. A bit more detail...

- This change fixes the specific (and fairly urgent - see below) issue of handling devicemapper block (DM) devices, which currently does not work in runtime-rs.
- That bug is stopping us from enabling the Cloud Hypervisor CI (see #9034).
- We need to have the runtime-rs CH CI enabled as soon as possible to:
  - protect the code in that driver.
  - allow us to demonstrate that the runtime-rs CH driver is on parity with the golang runtime's CH driver.
- Hence, I don't think we should stop this PR landing until every possible block device type (some of which are even architecture specific) is also added to it.

How about we raise issues for the remaining device types to make it clear to the community what further work needs to be done in this area?

amshinde:
@Apokleos I have addressed some of your comments. Have used a `match` construct to handle all cases. Please take a look.

src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
```diff
@@ -67,8 +70,30 @@ impl BlockRootfs {
         let mut device_id: String = "".to_owned();
         if let DeviceType::Block(device) = device_info {
             storage.driver = device.config.driver_option;
-            storage.source = device.config.virt_path;
             device_id = device.device_id;
+
+            match block_driver.clone().as_str() {
```
Apokleos, line 75: I think the `.clone()` is not required here and directly use `block_driver.as_str()` is fine.

src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
Apokleos: @jodh-intel I agree to push the PR merged ASAP.

src/runtime-rs/crates/resource/src/rootfs/block_rootfs.rs
Apokleos: Hi, @amshinde @jodh-intel I have no other required changes request but this one.

Apokleos (approved):
Thx @amshinde LGTM

Issue #9523, opened 2024-04-19
runtime: the go runtime does not execute CreateRuntime hooks on container creation


littlejawa:
# Description of problem

As part of the CI running on upstream cri-o, a PR was failing the kata test on the hooks test.
https://github.com/cri-o/cri-o/pull/7852

The PR is updating dependencies on various libraries, including containers/common, where the deprecation of the "prestart" hooks is taken into account. As part of it, "prestart" hooks are registered on the "CreateRuntime" phase. But on kata, this phase seems to be skipped.

# Expected result

CreateRuntime hooks should be run

# Actual result

CreateRuntime hooks are not run

# Further information

Looking further, I find that on "CreateSandbox",[ we run both "Prestart" and "CreateRuntime" hooks](https://github.com/kata-containers/kata-containers/blob/3456483df9f3f39c006fb1d7c6bece34d3fca282/src/runtime/pkg/katautils/create.go#L172).
 But on "CreateContainer", [we run only "Prestart"](https://github.com/kata-containers/kata-containers/blob/3456483df9f3f39c006fb1d7c6bece34d3fca282/src/runtime/pkg/katautils/create.go#L276).

Verified on the runtime-rs side: it seems that prestart and createruntime hooks are always run together.

PR #9524, opened 2024-04-19
runtime: Call CreateRuntime hooks at container creation time

littlejawa:
CreateRuntime hooks are called at the CreateSandbox time, but not after CreateContainer.

Fixes: #9523


gkurz (approved):
LGTM. Thanks @littlejawa !

Issue #9526, opened 2024-04-19
runtime-rs: add NVIDIA GPUDirect clique_id support


Apokleos:
**Which feature do you think can be improved?**

In GPUDirect, NVIDIA defines "cliques" to enable p2p DMA. Devices within a clique (sharing the same ID) can directly access each other's memory. we need support such feature to allow more than GPU devices passed throug in guest within the same clique ID are capable of direct P2P.

PR #9527, opened 2024-04-19
runtime-rs: add NVIDIA GPUDirect clique_id support

Apokleos:
In GPUDirect, NVIDIA defines "cliques" to enable p2p DMA. Devices within a clique (sharing the same ID) can directly access each other's memory. we need support such feature to allow more than GPU devices passed throug in guest within the same clique ID are capable of direct P2P.

Fixes: #9526

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
+++ b/src/runtime-rs/crates/hypervisor/src/device/driver/vfio.rs
@@ -198,6 +198,9 @@ pub struct VfioConfig {
     /// outside of the container mount namespace
     /// virt_path: Option<(index, virt_path_name)>
     pub virt_path: Option<(u64, String)>,
+
+    /// NVIDIA GPUs within the same clique ID are capable of direct P2P
+    pub clique_id: Option<u8>,
 }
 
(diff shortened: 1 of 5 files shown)


zvonkok:
The clique-id is not something that is generated. This is a thing that is supplied to the HV through the runtime and supplied as meta information via CDI.

zvonkok:
I will do a POC in the go-runtime soon and then we can see how to apply it for runtime-rs as well.

Apokleos:
> The clique-id is not something that is generated. This is a thing that is supplied to the HV through the runtime
>
Yeah, I am aware of that ! 
>
 and supplied as meta information via CDI.
>
How to pass the clique id by CDI ? annotation in CDI ?

Apokleos:
> The clique-id is not something that is generated. This is a thing that is supplied to the HV through the runtime
> 
 Yeah, I am aware of that !
> 

> 
> and supplied as meta information via CDI.
> 
How to pass the clique id by CDI ? annotation in CDI ?

Apokleos:
> I will do a POC in the go-runtime soon and then we can see how to apply it for runtime-rs as well.
>
sure, good idea !

PR #9530, opened 2024-04-19
genpolicy: changing caching so the tool can run concurrently with itself

Redent0r:
Based on 3a1461b0a5186a92afedaaea33ff2bd120d1cea0

Previously the tool would use the layers_cache folder for all instances and hence delete the cache when it was done, interfereing with other instances. This change makes it so that each instance of the tool will have its own temp folder to use.

### DIFF
--- a/src/tools/genpolicy/src/registry.rs
+++ b/src/tools/genpolicy/src/registry.rs
@@ -8,20 +8,23 @@
 
 use crate::containerd;
 use crate::policy;
+use crate::utils::Config;
 use crate::verity;
 
-use crate::utils::Config;
-use anyhow::{anyhow, bail, Result};
+use anyhow::{anyhow, Result};
 use docker_credential::{CredentialRetrievalError, DockerCredential};
-use log::warn;
-use log::{debug, info, LevelFilter};
-use oci_distribution::client::{linux_amd64_resolver, ClientConfig, ClientProtocol};
-use oci_distribution::{manifest, secrets::RegistryAuth, Client, Reference};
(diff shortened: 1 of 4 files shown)


danmihai1 (approved):

sprt (approved):

src/tools/genpolicy/src/registry.rs
```diff
@@ -57,21 +60,20 @@ pub struct DockerRootfs {
...
 impl Container {
-    pub async fn new(config: &Config, image: &str) -> Result<Self> {
+    pub async fn new(use_cached_files: bool, image: &str) -> Result<Self> {
         info!("============================================");
         info!("Pulling manifest and config for {:?}", image);
         let reference: Reference = image.to_string().parse().unwrap();
         let auth = build_auth(&reference);
 
         let mut client = Client::new(ClientConfig {
-            protocol: ClientProtocol::HttpsExcept(config.insecure_registries.clone()),
```
burgerdev, line 74: @Redent0r @danmihai1 I just noticed that you removed [support for insecure registries](https://github.com/kata-containers/kata-containers/pull/9294) here. Is this not compatible with the changed caching? Can we please reintroduce it?

danmihai1: @burgerdev , sorry, that was a mistake. We'll look at adding this back.

Redent0r: Adding back in https://github.com/kata-containers/kata-containers/pull/9786. Sorry about that @burgerdev !

PR #9534, opened 2024-04-22
Fix ctr exec stuck problem

Tim-Zhang:
agent: 
- Close stdin when write_stdin receives data of length 0.
- Stop call notify_term_close() in close_stdin, because it could discard stdout unexpectedly.

runtime-rs:
- Instead of call agent.close_stdin in close_io, we call agent.write_stdin with 0 len data when the stdin pipe ends.

### DIFF
--- a/src/agent/rustjail/src/process.rs
+++ b/src/agent/rustjail/src/process.rs
@@ -200,15 +200,8 @@ impl Process {
     }
 
     pub async fn close_stdin(&mut self) {
-        // stdin will be closed automatically in passfd-io senario
-        if self.proc_io.is_some() {
-            return;
-        }
-
         close_process_stream!(self, term_master, TermMaster);
         close_process_stream!(self, parent_stdin, ParentStdin);
-
-        self.notify_term_close();
     }
 
     pub fn cleanup_process_stream(&mut self) {
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -584,24 +584,31 @@ impl AgentService {
         let cid = req.container_id;
         let eid = req.exec_id;
 
-        let writer = {
-            let mut sandbox = self.sandbox.lock().await;
-            let p = sandbox.find_container_process(cid.as_str(), eid.as_str())?;
-
-            // use ptmx io
-            if p.term_master.is_some() {
-                p.get_writer(StreamType::TermMaster)
-            } else {
-                // use piped io
-                p.get_writer(StreamType::ParentStdin)
-            }
-        };
-
-        let writer = writer.ok_or_else(|| anyhow!(ERR_CANNOT_GET_WRITER))?;
-        writer.lock().await.write_all(req.data.as_slice()).await?;
-
         let mut resp = WriteStreamResponse::new();
         resp.set_len(req.data.len() as u32);
+
+        // EOF of stdin
+        if req.data.is_empty() {
+            let mut sandbox = self.sandbox.lock().await;
+            let p = sandbox.find_container_process(cid.as_str(), eid.as_str())?;
+            p.close_stdin().await;
+        } else {
+            let writer = {
+                let mut sandbox = self.sandbox.lock().await;
+                let p = sandbox.find_container_process(cid.as_str(), eid.as_str())?;
+
+                // use ptmx io
+                if p.term_master.is_some() {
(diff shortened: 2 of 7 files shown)


src/agent/rustjail/src/process.rs
```diff
@@ -210,17 +210,8 @@ impl Process {
...
 
     pub async fn close_stdin(&mut self) {
-        if let Some(proc_io) = &mut self.proc_io {
-            // notify io copy task to close stdin stream
-            let _ = proc_io.close_stdin_tx.send(true);
-            // wait for io copy task to finish
-            proc_io.wg_input.wait().await;
-        }
-
         close_process_stream!(self, term_master, TermMaster);
```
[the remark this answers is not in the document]
Tim-Zhang: Yes, they could be closed. In passfd-io, we don't keep term_master and parent_stdin in Process struct so close_process_stream! is meaningless for passfd-io.

src/runtime-rs/crates/runtimes/virt_container/src/container_manager/io/container_io.rs
```diff
@@ -80,6 +83,29 @@ impl<'inner> ContainerIoWrite<'inner> {
             }
         }
     }
+
+    fn poll_shutdown_inner(&'inner mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        let mut shutdown_future = self.shutdown_future.take();
+        if shutdown_future.is_none() {
```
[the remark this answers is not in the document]
Tim-Zhang: done

sprt (commented):
Related to #9187. This is a tricky code path as we keep regressing here. We should have integration tests to address #1160 and #9186 to have more confidence in our changes.

Tim-Zhang:
> Related to #9187. This is a tricky code path as we keep regressing here. We should have integration tests to address #1160 and #9186 to have more confidence in our changes.

@sprt I have added a stdio CI in this pr, but unfortunately I can't see it in the gh list, maybe it only can be seen after the merge?

sprt:
@Tim-Zhang We already have the infra to support such tests, you don't have to modify any workflow YAML 🙂 I recommend simply adding your test cases to this file: https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-exec.bats Please do reach out on Slack if you need any more guidance here, I'm happy to help.

Tim-Zhang:
@sprt Since the stdio CI tests is aimed to test `ctr` not `kubectl`, I'm not sure if it's appropriate to put it in the k8s-exec.bats file.

sprt:
@Tim-Zhang I would reimplement your tests to use `kubectl` instead of `ctr` as those both repro the issues in the same way in my experience - referring to e.g. #1160 which repros using `kubectl exec`.

Tim-Zhang:
@sprt There are subtle differences between `ctr` and `kubectl`, and `kubectl exec` would not stuck on this issue.

Now that you want to reimplement the tests later how about letting this PR in first?

src/agent/src/rpc.rs
```diff
@@ -584,24 +584,31 @@ impl AgentService {
         let cid = req.container_id;
         let eid = req.exec_id;
 
-        let writer = {
+        let mut resp = WriteStreamResponse::new();
+        resp.set_len(req.data.len() as u32);
+
+        // EOF of stdin
+        if req.data.len() == 0 {
```
Apokleos, line 591: Some CIs report error that `req.data.len()== 0` should be changed to `req.is_empty()` to make CIs happy.

Apokleos: sorry, req.is_empty() -> req.data.is_empty()

Tim-Zhang: done, thanks

Apokleos (approved):
Thx Tim @Tim-Zhang LGTM

lifupan (approved):

stevenhorsman:
Hi @Tim-Zhang, I'm a bit confused by this PR, it's trying to do:
```
function run() {
	info "Running stdio tests"

	bats "${runk_dir}/stdio-tests.bats"
}
```
but we don't have any stdio-tests.bats file, so this is failing every time. Was there supposed to be a follow up to this, or should we revert the tests part of this PR?

PR #9535, opened 2024-04-22
kata-deploy: Stop append `log_level = "debug"` for CRI-O

fidencio:
This should only be done once, and if CRI-O restarts, there's a big chance kata-deploy will also restart and the user would end up with a file that looks like:
```
[crio]
log_level = "debug"
[crio]
log_level = "debug"
[crio]
log_level = "debug"
...
```

And that would simply cause CRI-O to not start.

### DIFF
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -341,6 +341,8 @@ function configure_crio() {
 	mkdir -p "$crio_drop_in_conf_dir"
 	rm -f "$crio_drop_in_conf_file"
 	touch "$crio_drop_in_conf_file"
+	rm -f "$crio_drop_in_conf_file_debug"
+	touch "$crio_drop_in_conf_file_debug"
 
 	for shim in "${shims[@]}"; do
 		configure_crio_runtime $shim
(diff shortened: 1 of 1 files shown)


tools/packaging/kata-deploy/scripts/kata-deploy.sh
```diff
@@ -348,7 +348,7 @@ function configure_crio() {
 
 
 	if [ "${DEBUG}" == "true" ]; then
-		cat <<EOF | tee -a $crio_drop_in_conf_file_debug
+		cat <<EOF | tee $crio_drop_in_conf_file_debug
```
gkurz, line 351: It seems this falls into the `just ensure we remove any exist configuration` category. What about making this fix even more explicit with an `rm -f "op_in_conf_file_debug"` at https://github.com/kata-containers/kata-containers/pull/9535/commits/bde3773c1473bde771d4fb4da9e440be284cfac7#diff-08a680425cfe93e69c4108dc8fd91a9aeaefbf016bd9941b03c66834bf87a341R344 ?

gkurz (approved):
Approving anyway. Thanks @fidencio !

beraldoleal (approved):
LGTM, thanks

PR #9537, opened 2024-04-22
kata-deploy: configuring CRI-O for guest-pull image pulling

### LINKED ISSUE #9474: CC: run on-TEE jobs configured with CRI-O
@bpradipt asked if we could have a CI job to run the non-TEE CoCo tests with CRI-O for the sake of CoCo's 0.9.0 release.

After a brief chat (see https://cloud-native.slack.com/archives/C06UCUCG744) we agreed on creating a new job, using k0s and the default ubuntu based guest image, configured with CRI-O.

Cc @bpradipt @stevenhorsman @ldoktor @fidencio

wainersm:
This introduces a handler for PULL_TYPE variable to allow configuring CRI-O for guest-pull image pulling.

kata-deploy is going to append `runtime_pull_image = true` to each kata runtimeClass configuration created at `/etc/crio/crio.conf.d/99-kata-deploy` , when `PULL_TYPE=guest-pull`. For example:

```toml
[crio.runtime.runtimes.kata-qemu]
        runtime_path = "/usr/local/bin/containerd-shim-kata-qemu-v2"
        runtime_type = "vm"
        runtime_root = "/run/vc"
        runtime_config_path = "/opt/kata/share/defaults/kata-containers//configuration-qemu.toml"
        privileged_without_host_devices = true
        runtime_pull_image = true
```

Currently there aren't any CI jobs exercising this feature. I tested it in a fresh Ubuntu 22.04 VM by running the following script and checking  was as expected:

 ```bash
#!/bin/bash

set -o errexit
set -o pipefail
set -o nounset

export DOCKER_REGISTRY="quay.io"
export DOCKER_REPO="wainersm/kata-deploy"
export DOCKER_TAG="kata-containers-e15858616b0496fd5ca8659d4c9f46512e68cc51-amd64"
#PR_NUMBER: ${{ inputs.pr-number }}
export KATA_HYPERVISOR=qemu
export KUBERNETES=k0s
export KUBERNETES_EXTRA_PARAMS='--cri-socket remote:unix:///var/run/crio/crio.sock --kubelet-extra-args --cgroup-driver="systemd"'
#USING_NFD: "false"
export K8S_TEST_HOST_TYPE=small
export PULL_TYPE=guest-pull

echo "== SETUP CRIO"
./gha-run.sh setup-crio
echo "== DEPLOY K8S"
./gha-run.sh deploy-k8s

echo "== DEPLOY KATA"
./gha-run.sh deploy-kata-garm
#./gha-run.sh install-bats
#./gha-run.sh run-tests
 ```

ps: I didn't actually run a workload because I've faced issues with nested virt on my workstation.

(Partially) Fixes #9474

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -215,6 +215,13 @@ function deploy_kata() {
 		  "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
 	fi
 
+	# Set the PULL_TYPE_MAPPING
+	if [ "${PULL_TYPE}" != "default" ]; then
+		yq -i \
+		  ".spec.template.spec.containers[0].env[10].value = \"${KATA_HYPERVISOR}:${PULL_TYPE}\"" \
+		  "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
+	fi
+
 	echo "::group::Final kata-deploy.yaml that is used in the test"
 	cat "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
 	grep "${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" || die "Failed to setup the tests image"
--- a/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
+++ b/tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
@@ -24,6 +24,7 @@ spec:
               exec:
                 command: ["bash", "-c", "/opt/kata-artifacts/scripts/kata-deploy.sh cleanup"]
           command: ["bash", "-c", "/opt/kata-artifacts/scripts/kata-deploy.sh install"]
+          # NOTE: Please don't change the order of the environment variables below.
           env:
             - name: NODE_NAME
               valueFrom:
@@ -46,6 +47,8 @@ spec:
             - name: AGENT_HTTPS_PROXY
               value: ""
             - name: AGENT_NO_PROXY
+              value: ""
+            - name: PULL_TYPE_MAPPING
               value: ""
           securityContext:
             privileged: true
--- a/tools/packaging/kata-deploy/scripts/kata-deploy.sh
+++ b/tools/packaging/kata-deploy/scripts/kata-deploy.sh
@@ -32,6 +32,9 @@ snapshotters_delimiter=':'
 AGENT_HTTPS_PROXY="${AGENT_HTTPS_PROXY:-}"
 AGENT_NO_PROXY="${AGENT_NO_PROXY:-}"
 
+PULL_TYPE_MAPPING="${PULL_TYPE_MAPPING:-}"
+IFS=',' read -a pull_types <<< "$PULL_TYPE_MAPPING"
+
 # If we fail for any reason a message will be displayed
 die() {
         msg="$*"
@@ -426,6 +429,27 @@ function configure_crio_runtime() {
 	runtime_config_path = "${kata_config_path}"
 	privileged_without_host_devices = true
 EOF
+
+	local key
+	local value
+	if [ -n "${PULL_TYPE_MAPPING}" ]; then
+		for m in "${pull_types[@]}"; do
+			key="${m%"$snapshotters_delimiter"*}"
+			value="${m#*"$snapshotters_delimiter"}"
+
+			if [[ "${value}" = "default" || "${key}" != "${shim}" ]]; then
+				continue
+			fi
+
+			if [ "${value}" == "guest-pull" ]; then
+				echo -e "\truntime_pull_image = true" | \
+					tee -a "$crio_drop_in_conf_file"
+			else
+				die "Unsupported pull type '$value' for ${shim}"
+			fi
+			break
+		done
+	fi
 }
 
 function configure_crio() {
@@ -616,6 +640,7 @@ function main() {
 	echo "* SNAPSHOTTER_HANDLER_MAPPING: ${SNAPSHOTTER_HANDLER_MAPPING}"
 	echo "* AGENT_HTTPS_PROXY: ${AGENT_HTTPS_PROXY}"
 	echo "* AGENT_NO_PROXY: ${AGENT_NO_PROXY}"
+	echo "* PULL_TYPE_MAPPING: ${PULL_TYPE_MAPPING}"
 
 	# script requires that user is root
 	euid=$(id -u)


tools/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml
```diff
@@ -24,6 +24,7 @@ spec:
               exec:
                 command: ["bash", "-c", "/opt/kata-artifacts/scripts/kata-deploy.sh cleanup"]
           command: ["bash", "-c", "/opt/kata-artifacts/scripts/kata-deploy.sh install"]
+          # NOTE: Please don't change the order of the environment variables below.
```
beraldoleal, line 27: Appreciated! :)

tests/integration/kubernetes/gha-run.sh
```diff
@@ -210,6 +210,13 @@ function deploy_kata() {
 		  "${NO_PROXY}"
 	fi
 
+	if [[ "${PULL_TYPE}" != "default" && -n "${PULL_TYPE}" ]]; then
```
beraldoleal, line 213: Do we need the first check? I'm thinking about the case we have a modified `kata-deploy.yaml` with something else, and I would like to bring back to default? Maybe just checking if `${PULL_TYPE}` is set to something?

beraldoleal (dismissed):
LGTM, thanks. Left just a nit comment/suggestion.

wainersm:
Hi @beraldoleal !

> @wainersm the option is working fine from the configuration PoV. However during some tests with a basic workflow, I found this message:
>

Thanks for giving it a try!
 
> ```
> $ kubect describe pods/foo
> ...
>   Normal   Scheduled               4m22s                   default-scheduler  Successfully assigned default/nginx-test-with-kata2 to crio-cluster-worker-0
>   Normal   AddedInterface          4m21s                   multus             Add eth0 [10.85.0.6/16] from crio
>   Normal   AddedInterface          4m20s                   multus             Add eth0 [10.85.0.7/16] from crio
>   Normal   AddedInterface          4m18s                   multus             Add eth0 [10.85.0.8/16] from crio
>   Normal   AddedInterface          4m16s                   multus             Add eth0 [10.85.0.9/16] from crio
>   Normal   AddedInterface          4m14s                   multus             Add eth0 [10.85.0.10/16] from crio
>   Normal   AddedInterface          4m12s                   multus             Add eth0 [10.85.0.11/16] from crio
>   Normal   AddedInterface          4m10s                   multus             Add eth0 [10.85.0.12/16] from crio
>   Normal   AddedInterface          4m8s                    multus             Add eth0 [10.85.0.13/16] from crio
>   Normal   AddedInterface          4m6s                    multus             Add eth0 [10.85.0.14/16] from crio
>   Normal   AddedInterface          3m34s (x16 over 4m4s)   multus             (combined from similar events): Add eth0 [10.85.0.30/16] from crio
>   Warning  FailedCreatePodSandBox  3m33s (x25 over 4m20s)  kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = CreateContainer failed: Pause image not present in rootfs: unknown
> ```
> 
> Logs seems to be in a loop trying to mount the image:
> 
> https://paste.centos.org/view/1bab1d13
> 
> (Not sure tho if is something with my env, created with kcli + crio)

You will need to add the following metadata annotation to the pod yaml: `io.katacontainers.config.hypervisor.image: /opt/kata/share/kata-containers/kata-containers-confidential.img`. The reason being that the "vanilla" guest image hasn't the kata-agent compile for guest-pull, unlike the CoCo's image.

@littlejawa does it need an annotation similar to `io.containerd.cri.runtime-handler` but for cri-o? I couldn't find that information on the docs.

fidencio:
@wainersm, I am not big fond of this approach, to be honest.

While I understand that we can pull the image inside the guest with non-tees as well, we usually don't want to do that for several reasons, and this option would be mostly restricted to TEEs / the QEMU confidential non-TEE configuration that you proposed.

With this in mind, I'd **ONLY** set, **and set by default**, the parameter for those configurations.

fidencio:
> @littlejawa does it need an annotation similar to `io.containerd.cri.runtime-handler` but for cri-o? I couldn't find that information on the docs.

That's not needed for CRI-O.

wainersm:
Hi @fidencio !
> @wainersm, I am not big fond of this approach, to be honest.
> 

I'm not sure I understood your point.

Are you talking about this PR (i.e. kata-deploy configure CRI-O for guest-pull) or the way we are testing on non-TEE environments?

> While I understand that we can pull the image inside the guest with non-tees as well, we usually don't want to do that for several reasons, and this option would be mostly restricted to TEEs / the QEMU confidential non-TEE configuration that you proposed.
> 
> With this in mind, I'd **ONLY** set, **and set by default**, the parameter for those configurations.

beraldoleal (commented):
@wainersm just tested with the annotation is working fine. Thanks.

fidencio:
> I'm not sure I understood your point.
> 
> Are you talking about this PR (i.e. kata-deploy configure CRI-O for guest-pull) or the way we are testing on non-TEE environments?

We shouldn't be adding the option to pull in the guest for **all runtime handlers**, this is my point.
We don't do that for containerd, we should be consistent with CRI-O.

So, IMHO, we have two options here:
* Only add this option for the TEEs (and also for the non-TEE that will be used in our tests)
* Take advantge of the `runtime-handler:snapshotter` configuration, maybe expanding it, to be used with CRI-O, so that would allow the user to set that only for the runtime handlers it should be set for.

Does it make sense?

wainersm:
> > I'm not sure I understood your point.
> > Are you talking about this PR (i.e. kata-deploy configure CRI-O for guest-pull) or the way we are testing on non-TEE environments?
> 
> We shouldn't be adding the option to pull in the guest for **all runtime handlers**, this is my point. We don't do that for containerd, we should be consistent with CRI-O.
>

Alright, got it. And it makes sense to not apply the same pull option to all runtime handler.

For containerd, kata-deploy uses the `SNAPSHOTTER_HANDLER_MAPPING` variable to set the `snapshotter` property per runtime handler. Each snapshotter will be configured to a given pull type, being out of kata-deploy's scope to configure the snapshotter. 
 
> So, IMHO, we have two options here:
> 
>     * Only add this option for the TEEs (and also for the non-TEE that will be used in our tests)
>

Since the begin I avoided hard-coding on kata-deploy's code a mapping per runtimeclass. I don't like that option tbh.

>     * Take advantge of the `runtime-handler:snapshotter` configuration, maybe expanding it, to be used with CRI-O, so that would allow the user to set that only for the runtime handlers it should be set for.
> 

I like that idea, although I've some doubts on how to implement it.

Starting by the fact that snapshotter is a containerd concept, afaik. So using `SNAPSHOTTER_HANDLER_MAPPING` with cri-o sounds wrong. Suppose we rename it; is it okay that breakage?

> 
> Does it make sense?

fidencio:
> Starting by the fact that snapshotter is a containerd concept, afaik. So using `SNAPSHOTTER_HANDLER_MAPPING` with cri-o sounds wrong. Suppose we rename it; is it okay that breakage?

I think breaking is not the problem, but I'd take the simpler approach first, and expand if we need the second.
So, just add that, unconditionally, for qemu-tdx, qemu-sev, qemu-snp, and the soon-to-be-added qemu-coco. :-)

wainersm:
Hi @fidencio !

> > Starting by the fact that snapshotter is a containerd concept, afaik. So using `SNAPSHOTTER_HANDLER_MAPPING` with cri-o sounds wrong. Suppose we rename it; is it okay that breakage?
> 
> I think breaking is not the problem, but I'd take the simpler approach first, and expand if we need the second. So, just add that, unconditionally, for qemu-tdx, qemu-sev, qemu-snp, and the soon-to-be-added qemu-coco. :-)

Just updated with a fixup commit that converts from `PULL_TYPE` to `PULL_TYPE_MAPPING`, so following the same logic of `SNAPSHOTTER_HANDLER_MAPPING` but using a different variable. All in all, I think this new variable will make kata-deploy more flexible for, if needed, handle pull_type <-> snapshotter configurations.

Re-tested, it seems to work fine:
```
Environment variables passed to this script
* NODE_NAME: kata-dev
* DEBUG: true
* SHIMS: qemu
* DEFAULT_SHIM: qemu
* CREATE_RUNTIMECLASSES: true
* CREATE_DEFAULT_RUNTIMECLASS: true
* ALLOWED_HYPERVISOR_ANNOTATIONS: image initrd kernel default_vcpus
* SNAPSHOTTER_HANDLER_MAPPING: 
* AGENT_HTTPS_PROXY: 
* AGENT_NO_PROXY: 
* PULL_TYPE_MAPPING: qemu:guest-pull
Creating the runtime classes
Creating the kata-qemu runtime class
runtimeclass.node.k8s.io/kata-qemu created
Creating the kata runtime class for the default shim (an alias for kata-qemu)
runtimeclass.node.k8s.io/kata created
Creating the default shim-v2 binary
Add Kata Containers as a supported runtime for CRIO:

[crio.runtime.runtimes.kata-qemu]
        runtime_path = "/usr/local/bin/containerd-shim-kata-qemu-v2"
        runtime_type = "vm"
        runtime_root = "/run/vc"
        runtime_config_path = "/opt/kata/share/defaults/kata-containers//configuration-qemu.toml"
        privileged_without_host_devices = true
        runtime_pull_image = true
[crio]
log_level = "debug"
node/kata-dev labeled
::endgroup::
::group::Runtime classes
NAME        HANDLER     AGE
kata-qemu   kata-qemu   66s
kata        kata-qemu   66s
::endgroup::
```

wainersm:
Updated this PR to:
 * address @stevenhorsman 's comment
 * rebase to main. Only needed to re-write a call to `yq` due recent bump on its version required

Run the same tests locally, it works.

fidencio (approved):
It sucks to have to add container engine specifics to the kata-deploy env vars, but at the end of the day it is what it is.

looks reasonable to me, thank @wainersm.

stevenhorsman (approved):
Looks okay to me

beraldoleal (approved):
LGTM, thanks

wainersm:
Hi @fidencio !

> It sucks to have to add container engine specifics to the kata-deploy env vars, but at the end of the day it is what it is.
>

I thought in unify in a single mapping but I was afraid of getting it too complex.

Also I believed that the new mapping (`PULL_TYPE_MAPPING`) might be useful for containerd in the future too. For example, host-sharing pull + nydus, kata-deploy may use `SNAPSHOTTER_HANDLER_MAPPING=qemu:nydus` and `PULL_TYPE_MAPPING=qemu:host-share` to proper configure the environment. For this example, if it didn't have `PULL_TYPE_MAPPING` then it would need to use something like `SNAPSHOTTER_HANDLER_MAPPING=qemu:nydus_with_host_sharing`.
 

> looks reasonable to me, thank @wainersm.

I'm okay to use a single variable. However, `SNAPSHOTTER_HANDLER_MAPPING` makes no sense to cri-o as the `snapshotter` concept is containerd's specific. We would need to use something else.

Let me know what you think, I still can re-work this.

PR #9547, opened 2024-04-23
tdx: Update QEMU version

fidencio:
Let's update QEMU to the -next-2024.03.27-v9.0.0-rc0, which has support for talking to the quote-generation-socket.


fidencio:
I need to update the host kernel and on the server side, but I will hold on on that till we have an official CentOS 9 Stream package in a few days.

fidencio:
Let me close this one, as we're currently using the QEMU / OVMF coming from the distros.

PR #9549, opened 2024-04-24
build: Fix tarball not building correctly in docker

JakubLedworowski:
When docker is installed on the host system using script from https://get.docker.com/ it automatically creates a docker group with _gid=999_. 
Then during docker build process of tarball, eg. **make qemu-tdx-experimental-tarball** docker is also installed inside the image with the same script, which also automatically adds docker group with _gid=999_.
Then, the build tries to add a new group _docker_on_host_ with _gid=999_, which already exists, which breaks the build.


fidencio (approved):
lgtm, thanks @JakubLedworowski!

dborquez (approved):
Lgtm, thanks @JakubLedworowski

PR #9550, opened 2024-04-24
Tag component caches

stevenhorsman:
Add the ability to tag certain components in our ghcr cache, rather than just storing them all by digest. In particular the cloud-api-adaptor implementation of the remote-hypervisor wants to use the agent, agent-opa and guest-components

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -102,6 +102,7 @@ jobs:
           ARTEFACT_REGISTRY_USERNAME: ${{ github.actor }}
           ARTEFACT_REGISTRY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
           TARGET_BRANCH: ${{ inputs.target-branch }}
+          RELEASE: ${{ if inputs.stage == 'release' && 'yes' || 'no' }}
 
       - name: store-artifact ${{ matrix.asset }}
         uses: actions/upload-artifact@v4
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -60,13 +60,7 @@ jobs:
         stage:
           - ${{ inputs.stage }}
         exclude:
-          - asset: agent
-            stage: release
           - asset: cloud-hypervisor-glibc
-            stage: release
-          - asset: pause-image
-            stage: release
-          - asset: coco-guest-components
             stage: release
     steps:
       - name: Login to Kata Containers quay.io
@@ -105,6 +99,7 @@ jobs:
           RELEASE: ${{ if inputs.stage == 'release' && 'yes' || 'no' }}
 
       - name: store-artifact ${{ matrix.asset }}
+        if: ${{ matrix.stage != 'release' || (matrix.component != 'agent' && matrix.component != 'coco-guest-components' && matrix.component != 'pause-image') }}
         uses: actions/upload-artifact@v4
         with:
           name: kata-artifacts-amd64-${{ matrix.asset }}${{ inputs.tarball-suffix }}
--- a/.github/workflows/release-amd64.yaml
+++ b/.github/workflows/release-amd64.yaml
@@ -10,6 +10,7 @@ jobs:
   build-kata-static-tarball-amd64:
     uses: ./.github/workflows/build-kata-static-tarball-amd64.yaml
     with:
+      push-to-registry: yes
       stage: release
 
   kata-deploy:
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -49,6 +49,7 @@ ARTEFACT_REGISTRY_PASSWORD="${ARTEFACT_REGISTRY_PASSWORD:-}"
 TARGET_BRANCH="${TARGET_BRANCH:-main}"
 PUSH_TO_REGISTRY="${PUSH_TO_REGISTRY:-}"
 KERNEL_HEADERS_PKG_TYPE="${KERNEL_HEADERS_PKG_TYPE:-deb}"
+RELEASE="${RELEASE:-"no"}"
 
 workdir="${WORKDIR:-$PWD}"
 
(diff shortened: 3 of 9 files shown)


tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1106,25 +1108,33 @@ handle_build() {
...
-					${build_target}-version \
-					${build_target}-builder-image-version \
-					${build_target}-sha256sum
-				;;
-		esac
+		tags=(latest-${TARGET_BRANCH}-$(uname -m))
+		if [ -n "${artefact_tag}" ]; then
+			tags+=("${artefact_tag}")
+		fi
+		# TODO - Ideally we would also tag release version if we are doing a release. How can we tell here?
```
fidencio, line 1115: We bump the `${repo_root_dir}/VERSION` file before the release is tagged, meaning that if we know that we're doing a release, heeeey, we easily can get the version.

Now, how can we know that we're in a release?
I'd simply export an env var as part of the RELEASE workflow, and check for this env var here, adding then the version coming from `${repo_root_dir}/VERSION` as part of the tags.

WDYT?

stevenhorsman: Yeah, that's worth a try. Thanks

stevenhorsman: An attempt was made

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1106,25 +1109,35 @@ handle_build() {
...
-					${build_target}-version \
-					${build_target}-builder-image-version \
-					${build_target}-sha256sum
-				;;
-		esac
+		tags=(latest-${TARGET_BRANCH}-$(uname -m))
+		if [ -n "${artefact_tag}" ]; then
+			tags+=("${artefact_tag}")
+		fi
+		if [ "${RELEASE}" == "yes" ]; then
```
fidencio, line 1116: Can you also add RELEASE=${RELEASE:-"no"} to the beginning of the script?
Just to make sure we don't run into "unbound variable" when not running from the CI.

.github/workflows/release-amd64.yaml
```diff
@@ -10,6 +10,7 @@ jobs:
   build-kata-static-tarball-amd64:
     uses: ./.github/workflows/build-kata-static-tarball-amd64.yaml
     with:
+      push-to-registry: yes
```
wainersm, line 13: Hi @stevenhorsman !

In `./.github/workflows/build-kata-static-tarball-amd64.yaml` with `push-to-registry=yes` it will log-in quay.io but push the image to ghcr.io.  Unless I missed something, we should log-in the right registry :D (applies to the arm64, ppc64le and s390x workflows).

wainersm: hmm... just realized that it's passing  ARTEFACT_REGISTRY_USERNAME and  ARTEFACT_REGISTRY_PASSWORD variables. So likely internally, somewhere, it's loging to ghcr.io

stevenhorsman: Yeah - I was puzzled by the ghcr.io login, but assumed it must been done auto-magically by github

.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -105,6 +99,7 @@ jobs:
           RELEASE: ${{ if inputs.stage == 'release' && 'yes' || 'no' }}
 
       - name: store-artifact ${{ matrix.asset }}
+        if: ${{ matrix.stage != 'release' && (matrix.component == 'agent' || matrix.component == 'coco-guest-components' || matrix.component == 'pause-image') }}
```
wainersm, line 102: This bug'ed me!

I think you meant:

`matrix.stage != 'release' || (matrix.component != 'agent' && matrix.component != 'coco-guest-components' && matrix.component != 'pause-image')`

.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -60,14 +60,8 @@ jobs:
         stage:
```
wainersm, line 60: Remove the `stage` parameter as you did for [.github/workflows/build-kata-static-tarball-s390x.yaml](https://github.com/kata-containers/kata-containers/pull/9550/commits/9b8805432b3a911bc91056b542f954a17a2863f3#diff-75d5ecb9b253be81b84649635e2d463572d82f18fe2fee5a51ed96225e237fed) ?

.github/workflows/build-kata-static-tarball-amd64.yaml
wainersm: If I'm correct here then it needs to adjust the next store-artifacts's ifs.

stevenhorsman: Yes - you are correct, I think I should have had a `!` before `(matrix.component == 'agent' || matrix.component == 'coco-guest-components' || matrix.component == 'pause-image')`, but your code works too. Thanks for the good spot!

.github/workflows/build-kata-static-tarball-amd64.yaml
stevenhorsman: We still need it to filter out:
```
          - asset: cloud-hypervisor-glibc
            stage: release
```
which the s390x build doesn't need?

wainersm: ok, missed that.

.github/workflows/build-kata-static-tarball-amd64.yaml
stevenhorsman: Ok, I've updated the logic on all the build jobs to actually work now. Thanks!

.github/workflows/release-amd64.yaml
wainersm: Found it: https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L1092

wainersm (approved):
This will be very useful on peer-pods indeed! Thanks @stevenhorsman !

PR #9551, opened 2024-04-24
runtime-rs: support IOMMU in qemu VMs

pmores:
This PR adds support for the IOMMU device and `iommu_platforms` device param to runtime-rs qemu driver.  It can be also viewed as implementing support for `configuration.toml`'s `enable_iommu` and `enable_iommu_platform` settings.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
@@ -823,6 +823,7 @@ struct VhostVsock {
     vhostfd: tokio::fs::File,
     guest_cid: u32,
     disable_modern: bool,
+    iommu_platform: bool,
 }
 
 impl VhostVsock {
@@ -832,11 +833,17 @@ impl VhostVsock {
             vhostfd,
             guest_cid,
             disable_modern: false,
+            iommu_platform: false,
         }
     }
 
     fn set_disable_modern(&mut self, disable_modern: bool) -> &mut Self {
         self.disable_modern = disable_modern;
+        self
+    }
+
+    fn set_iommu_platform(&mut self, iommu_platform: bool) -> &mut Self {
+        self.iommu_platform = iommu_platform;
         self
     }
 }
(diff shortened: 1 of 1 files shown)


pmores:
@BbolroC I took the liberty in this PR at a couple of places to provisionally support CCW-based systems.  While I believe it's reasonable since the idea is just to adopt what the golang shim does  I would appreciate if you could verify the PR on s390x.

lifupan (approved):

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1084,6 +1084,36 @@ impl ToQemuParams for DeviceVirtconsole {
...
+        }
+    }
+}
+
+#[async_trait]
+impl ToQemuParams for DeviceIntelIommu {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        let mut params = Vec::new();
+        params.push("intel-iommu".to_owned());
+        let to_onoff = |b| if b { "on" } else { "off" };
```
Apokleos, line 1139: It seems that the `else` branch of `to_onoff` is never reached ? as the result of all fields of `DeviceIntelIommu` initialized with `true`.

BbolroC (approved):
LGTM, thanks @pmores !

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
pmores: Yes, you're right.  As you can see, the whole `DeviceIntelIommu` struct is a bit make-believe since the initialisation is basically hard-coded, just like in the go runtime.  Here I just preferred to limit the hard-coded bit to the constructor and write the rest of the class _as if_ the values weren't hard-coded.

It might help just a little bit if we stop hard-coding the initialisation at some point in the future...

Apokleos: Ok SGTM !

Apokleos (approved):
Thx @pmores LGTM ! Let's push it merged !

PR #9557, opened 2024-04-25
CI: describe pod on k8s-create-pod wait failure

portersrc:
Currently attempting to debug a nydus pod launch, and this should give more insight.

Temporary PR without an issue yet. NOT a full fix for #9378, though we need to debug this to do that kind of SNP work.


ryansavino (approved):

danmihai1 (approved):

lifupan (approved):

Apokleos:
Hi @portersrc Thx for your contribution. 
the required CI reports error message "error_msg: See the document below for help on formatting commits for the project." I checked your commit`CI: extra cleanup step for nydus` myabe, you should add commit message to pass this CI.

portersrc:
Hey @Apokleos, thanks for catching this! I'll correct the commit message as you describe when we finalize it. In fact, We're not 100% sure on this second commit now anyway. Part of that discussion will probably be tracked in a slack thread [here](https://cloud-native.slack.com/archives/C066XLCU95Z/p1714086813327869).

wainersm:
Hi @portersrc , do you need this PR merged or using just to debug the errors?

tests/integration/kubernetes/gha-run.sh
```diff
@@ -548,6 +548,10 @@ function cleanup_nydus_snapshotter() {
 	kubectl delete -f "misc/snapshotter/nydus-snapshotter-rbac.yaml"
 	kubectl get namespace nydus-system -o json | jq 'del(.spec.finalizers)' | kubectl replace --raw "/api/v1/namespaces/nydus-system/finalize" -f - || true
 	popd
+	for i in `sudo ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1 || true`; do
+		sudo ctr -n k8s.io snapshot --snapshotter nydus rm $i
+	done
+	sudo rm -rf "/var/lib/containerd-nydus" || true
```
fidencio, line 554: This should be done by the nydus-snapshotter.
Any specific reason to enforce it here?

In fact, please, take a look at: https://github.com/containerd/nydus-snapshotter/blob/ce4848e4439b79723b0635e994b5d9991aa88819/misc/snapshotter/snapshotter.sh#L195

So, isn't the snapshotter doing exactly the same thing on its own side?

fidencio (changes requested):
Please, take a look at the nydus snapshotter's cleanup script.

tests/integration/kubernetes/gha-run.sh
fitzthum: I think these guys ran into the `Error: failed to create containerd container: create instance 6: object with key "6" already exists: unknown` that you also saw on the TDX node. Did you end up doing anything to fix that besides local cleanup?

fidencio: Regardless of the error they hit, this should be fixed on the nydus snapshotter side, *not* on the Kata Containers side.

I didn't hit the issue on TDX, but that could happen there as well.  I hit the issue on a local VM when trying the snapshotter before the release.  That cleanup was enough to get me moving from the error.

fidencio: @wainersm pointed out in the slack thread a snippet to clean up the `content`, that's not relevant for the error that Tobin mentioned to be happening, but rather to remove the image from overlayfs snapshotter and allowing to pull the image with nydus, which is an approach **we do not want to take here**, as that's just papering over real issues.

portersrc:
We've moved past this debug PR at this point. However, the first commit adds `kubectl describe pod "$pod_name"` on a failure case, and that could still be useful. We should take that change or close this.

portersrc:
This PR came up in the coco CI community meeting today. I've rebased, improved the commit message, and removed the do-not-merge label.

If folks are OK with this PR, it's ready for final reviews.

stevenhorsman (approved):
Looks fine to me. Thanks!

fidencio (approved):
lgtm, thanks @portersrc!

ryansavino:
@danmihai1 looks like we need your approval to get this one merged.

PR #9563, opened 2024-04-26
build: Build the shipped agent with policy enabled

fidencio:
Now that the OPA binary is not required anymore, let's start shipping
the agent with the policy enabled by default.

The agent *without* policy enabled has 30MB, while it's 34MB *with* the
policy enabled.

This 4MB (~10%) increase is, IMHO, worth it in order to reduce the
amount of components we have to maintain and test, including the
possibility to also reduce the amount of possible rootfs / initrd
images.

Whoever wants to use the agent without policy enabled can simply do that
by building their own agent. :-)

/cc @stevenhorsman @danmihai1

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -28,7 +28,6 @@ jobs:
       matrix:
         asset:
           - agent
-          - agent-opa
           - agent-ctl
           - cloud-hypervisor
           - cloud-hypervisor-glibc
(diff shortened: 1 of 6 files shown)


stevenhorsman (commented):
I'm happy with this and have added this PR link to the AC agenda as we told them we'd have the discussion/decision there tomorrow.

stevenhorsman (approved):
There were not complaints in the AC meeting, so it's a 👍 from me! Thanks!

danmihai1 (approved):
Thanks for helping out, @fidencio !

sprt (approved):

fidencio:
I'm merging this one without the CI passing, as it's removing some targets used by the CI.

PR #9564, opened 2024-04-26
runtime-rs: Implement resize memory for CH

### LINKED ISSUE #8801: runtime-rs: ch: Implement full resize memory API

cmaf:
Add resize_memory for Cloud Hypervisor and cloud_hypervisor_vm_resize for the call to the VM. Write unit tests checking failure cases.

Fixes #8801

### DIFF
--- a/src/runtime-rs/crates/hypervisor/ch-config/src/ch_api.rs
+++ b/src/runtime-rs/crates/hypervisor/ch-config/src/ch_api.rs
@@ -100,6 +100,31 @@ pub async fn cloud_hypervisor_vm_blockdev_add(
     .await?
 }
 
+#[derive(Deserialize, Serialize, Default, Debug)]
+pub struct VmResizeData {
+    pub desired_vcpus: Option<u8>,
+    pub desired_ram: Option<u64>,
+    pub desired_balloon: Option<u64>,
+}
+
+pub async fn cloud_hypervisor_vm_resize(
+    mut socket: UnixStream,
+    resize_data: VmResizeData,
+) -> Result<Option<String>> {
+    task::spawn_blocking(move || -> Result<Option<String>> {
+        let response = simple_api_full_command_and_response(
+            &mut socket,
+            "PUT",
+            "vm.resize",
+            Some(&serde_json::to_string(&resize_data)?),
+        )
+        .map_err(|e| anyhow!(e))?;
+
+        Ok(response)
+    })
+    .await?
+}
+
 pub async fn cloud_hypervisor_vm_netdev_add(
     mut socket: UnixStream,
     net_config: NetConfig,
--- a/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
+++ b/src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
@@ -14,9 +14,10 @@ use crate::VM_ROOTFS_DRIVER_PMEM;
 use crate::{VcpuThreadIds, VmmState};
 use anyhow::{anyhow, Context, Result};
 use ch_config::ch_api::{
-    cloud_hypervisor_vm_create, cloud_hypervisor_vm_start, cloud_hypervisor_vmm_ping,
-    cloud_hypervisor_vmm_shutdown,
+    cloud_hypervisor_vm_create, cloud_hypervisor_vm_resize, cloud_hypervisor_vm_start,
+    cloud_hypervisor_vmm_ping, cloud_hypervisor_vmm_shutdown, VmResizeData,
 };
+use ch_config::convert::checked_next_multiple_of;
 use ch_config::{guest_protection_is_tdx, NamedHypervisorConfig, VmConfig};
 use core::future::poll_fn;
 use futures::executor::block_on;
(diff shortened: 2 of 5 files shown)


src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,118 @@ impl CloudHypervisorInner {
...
+
+        let block_size_mb = self.guest_memory_block_size_mb;
+
+        let new_hotplug_size_mb = new_mem_mb - current_mem_size_mb;
+        let aligned_new_hotplug_size_mb =
+            checked_next_multiple_of(new_hotplug_size_mb.into(), block_size_mb.into());
+
+        if let Some(aligned_value_mb) = aligned_new_hotplug_size_mb {
+            let aligned_request = current_mem_size_mb + aligned_value_mb as u32;
+            if new_mem_mb != aligned_request {
```
pmores, line 820: Is this testing only done for the log message?  Because otherwise ~10 lines of code could be reduced to one.  (`new_mem_mb` will end up equal to `aligned_request` no matter what.)

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,118 @@ impl CloudHypervisorInner {
...
+                info!(
+                    sl!(),
+                    "Aligning VM memory request {} to {}", new_mem_mb, aligned_request
```
pmores, line 823: Provided we have to have this log message, I'd propose including `block_size_mg` as well so that complete information related to aligning is available to anyone trying to track down a problem.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,118 @@ impl CloudHypervisorInner {
...
+                new_mem_mb = aligned_request;
+            }
+        }
+
+        // Check if memory is the same as requested, a second Check is done
+        // to consider the memory request now that is updated to be memory aligned
+        if new_mem_mb == current_mem_size_mb {
+            info!(
+                sl!(),
+                "VM already has requested memory after alignment {}", new_mem_mb
```
pmores, line 834: It's a matter of taste but frankly, this wording seem a bit cryptic to me, considering that we're just trying to say that nothing needed to be done since the VM has had the requested amount of memory already.

src/runtime-rs/crates/hypervisor/ch-config/src/ch_api.rs
```diff
@@ -100,6 +100,31 @@ pub async fn cloud_hypervisor_vm_blockdev_add(
...
+
+pub async fn cloud_hypervisor_vm_resize(
+    mut socket: UnixStream,
+    resize_data: VmResizeData,
+) -> Result<Option<String>> {
+    task::spawn_blocking(move || -> Result<Option<String>> {
+        let response = simple_api_full_command_and_response(
+            &mut socket,
+            "PUT",
+            "vm.add-disk",
```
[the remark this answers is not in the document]
cmaf: Yes. this was a copy/paste error, thank you for catching it!

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
cmaf: Fixed

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
cmaf: I agree, it may not be necessary to log this.

cmaf:
@pmores thanks for the review. I updated a few things based on your comments and also modified some of the code comments for clarity. For logging, I chose to log the same information that is logged in the go runtime for cloud hypervisor. We could consider reducing it or switching it to debug.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,116 @@ impl CloudHypervisorInner {
...
+            if new_mem_mb != aligned_request {
+                info!(
+                    sl!(),
+                    "Aligning VM memory request {} to {} with block size {}", new_mem_mb, aligned_request, block_size_mb;
+                );
+                new_mem_mb = aligned_request;
+            }
+        }
+
+        // Additional checks after memory alignment
```
lifupan, line 829: Hi @cmaf 

Is the init default memory  block size aligned? If the initial values are aligned to the block size, then it is enough to do the alignment first and then do these size checks. Is there no need to check before the alignment?

pmores: That's a very interesting question.  Correct me if I'm wrong but it would be hard to rely on `memory_info.default_memory` being aligned since 1) it comes from the config file, ie. the user, and 2) the memory allocation granularity (the block size) is only determined at runtime via the agent.

So I _think_ (and I might well be off here) that at least on the initial invocation of this function when `current_hotplug_size_mb` is likely zero and `current_mem_size_mb` thus consists purely of `memory_info.default_memory` alignment of the initial values can hardly be guaranteed.

That said, I still think you're right suggesting that the pre-alignment checks could be left out.  They seem to serve as an early-out to avoid the alignment computations if it's clear that they are not needed.  This would make a lot of sense if they were considerably resource-heavy.  However, since that doesn't seem the case I'd consider leaving the pre-alignment checks out, simplifying the function at the risk that a bunch of simple arithmetics might be performed unnecessarily at times.

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
pmores: I don't mind logging this at all, I just wanted to make sure this is a conscious decision. Thanks!

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
cmaf: @lifupan @pmores I switched the checks to only occur after alignment

pmores (approved):
lgtm, thanks @cmaf !

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,109 @@ impl CloudHypervisorInner {
...
 
-    pub(crate) fn resize_memory(&self, _new_mem_mb: u32) -> Result<(u32, MemoryConfig)> {
-        warn!(sl!(), "CH memory resize not implemented - see https://github.com/kata-containers/kata-containers/issues/8801");
+    pub(crate) async fn resize_memory(
+        &mut self,
+        mut new_mem_mb: u32,
+    ) -> Result<(u32, MemoryConfig)> {
+        // check for probe in memory_config
+        let caps = &self.capabilities;
+        if caps.is_mem_hotplug_probe_supported() {
```
amshinde, line 778: Since we know that cloud-hypervisor supports memory hotplug, should be perform the check here?
I feel that this check needs to be done at a higher abstraction level (hypervisor level ) before this gets called at all.

amshinde: Also since GuestMemoryProbe has not been added to the capability set yet, is this check passing currently? Or am I missing something?

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,109 @@ impl CloudHypervisorInner {
...
+        if let Some(aligned_value_mb) = aligned_new_hotplug_size_mb {
+            let aligned_request = current_mem_size_mb + aligned_value_mb as u32;
+            if new_mem_mb < aligned_request {
+                info!(
+                    sl!(),
+                    "Aligning VM memory request {} to {} with block size {}", new_mem_mb, aligned_request, block_size_mb;
+                );
+                new_mem_mb = aligned_request;
+            }
+        }
```
pmores, line 823: What happens if `aligned_new_hotplug_size_mb` is `None`, i.e. if aligning fails?  Do we want to continue with the old unaligned size?

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,109 @@ impl CloudHypervisorInner {
     }
 
     pub(crate) fn set_guest_memory_block_size(&mut self, size: u32) {
-        self._guest_memory_block_size_mb = size;
+        self.guest_memory_block_size_mb = size;
```
pmores, line 765: From my tests it looks like the value passed to this function is actually in bytes, not megabytes - you might want to double-check that...

src/runtime-rs/crates/hypervisor/src/ch/inner_hypervisor.rs
```diff
@@ -761,17 +762,109 @@ impl CloudHypervisorInner {
...
+            return Ok((current_mem_size_mb, MemoryConfig::default()));
+        }
+
+        let block_size_mb = self.guest_memory_block_size_mb;
+
+        let new_hotplug_size_mb = new_mem_mb - current_mem_size_mb;
 
-        Ok((0, MemoryConfig::default()))
+        // checked_next multiple_of() will align 0 to the block size, which we don't want here,
+        // so run check to avoid that
```
pmores, line 807: It seems that the function will do that not just for zero but for any input value that is aligned already.

PR #9570, opened 2024-04-29
workflow: static-checks: Skip commit checks for dependabout

stevenhorsman:
Dependabot doesn't follow all our commit format guidelines, so add a check and skip these if the author is `dependabot[bot]`


zvonkok (approved):
LGTM, thanks a ton.

fidencio (approved):
lgtm, thanks @stevenhorsman!

Issue #8245, opened 2023-10-17
runtime-rs sandbox_cgroup_only is broken


sepich:
# Description of problem
Installed latest v3.2.0-rc0 and starting runtime-rs with default `/opt/kata/share/defaults/kata-containers/configuration-dragonball.toml`

# Expected result

cgroup would be created in the correct place (and inherit container mem limit)

# Actual result

1. `sandbox_cgroup_only=false` container does not start:
```
E1017 14:20:17.319380  136650 remote_runtime.go:193] "RunPodSandbox from runtime service failed" err="rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: Others(\"failed to handler message create container\\n\\nCaused by:\\n    0: create\\n    1: using method in wrong cgroup mode.\"): unknown"
```

2. with default value `sandbox_cgroup_only=true` cgroup is created without memory limit, and in a wrong place:
```
# systemd-cgls memory

Control group /:
├─kubepods-burstable-pod7432e27c_9134_4c82_9320_a13e0cf65f62.slice:cri-containerd:b3c5a2685e567bb3c9765138079f9e924540a63695a280e7dd6fc0d166f6ef83
│ └─163071 /opt/kata/runtime-rs/bin/containerd-shim-kata-v2 -id b3c5a2685e567bb…
└─kubepods.slice
  ├─kubepods-burstable.slice
  │ ├─kubepods-burstable-podb9f10f16090c87895b1c288d42df1915.slice
  │ │ ├─cri-containerd-220a5de21396daaeab31eecb8acb0ee7025845eb48df5f373f1b3af9f78e1f0d.scope …
  │ │ │ └─2404 /pause
...
  
# cat /sys/fs/cgroup/kubepods-burstable-pod7432e27c_9134_4c82_9320_a13e0cf65f62.slice\:cri-containerd\:b3c5a2685e567bb3c9765138079f9e924540a63695a280e7dd6fc0d166f6ef83/memory.high
max
```
Which contradicts docs (and how it works for СLH, QEMU):
https://github.com/kata-containers/kata-containers/blob/main/docs/design/host-cgroups.md#sandbox_cgroup_only--true
So, `sandbox_cgroup_only` is broken for both values.

Maybe because you are trying to work with cgroup v1, and it is cgroup v2 already from kernel v4:
```
[...]

PR #9572, opened 2024-04-30
runtime-rs: Use add_task_by_tgid for constrain_hypervisor

Champ-Goblem:
runtime-rs `CgroupsResource.constrain_hypervisor` used `add_task` instead of `add_task_by_tgid` when adding the hypervisor threads to the cgroup.
This causes an issue due to the default type of the cgroup being `domain` instead of `threaded` or `domain threaded` and throws the error: `using method in wrong cgroup mode`.
CgroupsResource.new already uses `add_task_by_tgid` so this change should be fine

Fixes: #8245


Champ-Goblem:
@studychao if not seen

PR #9577, opened 2024-04-30
genpolicy: add topologySpreadConstraints support

Redent0r:
Allow genpolicy to process Pod YAML files including topologySpreadConstraints.

### DIFF
--- a/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod.yaml
+++ b/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod.yaml
@@ -27,3 +27,7 @@ spec:
       securityContext:
         seccompProfile:
           type: RuntimeDefault
+  topologySpreadConstraints:
+    - maxSkew: 2
+      topologyKey: kubernetes.io/hostname
+      whenUnsatisfiable: ScheduleAnyway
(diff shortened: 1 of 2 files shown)


tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod.yaml
```diff
@@ -27,3 +27,7 @@ spec:
       securityContext:
         seccompProfile:
           type: RuntimeDefault
+  topologySpreadConstraints:
+        - maxSkew: 2
```
wainersm, line 31: could you please keep it 2-spaces indentation?

wainersm: That's the only comment I have about this PR. I check the pod spec in https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/ and this is handling all the fields (optional and required).

Redent0r: Updated. Thanks @wainersm !

wainersm (approved):
Thanks @Redent0r !

stevenhorsman (approved):
I don't necessarily understand the option we are enabling policy for, but the code looks okay and the tests pass, so I think it's ok to go in.

PR #9581, opened 2024-05-01
build: coco-guest-components: Fix TEE_PLATFORM support

fidencio:
We want to build with support for *all* the TEE_PLATFORM, however, we've been passing the wrong value to the TEE_PLATFORM env var, as using ${TEE_PLATFORM:+"all"} results in an empty string, and what we really want to use is ${TEE_PLATFORM:-"all"} instead.

/cc @bpradipt @stevenhorsman @mythi


stevenhorsman:
So I deliberately changed this as `TEE_PLATFORM=all` doesn't build properly and there were plans to fix that on the guest side, so I think I set it to only specify if not set, not if not empty so that in future we could back it off easier once it's fixed upstream. Sorry for the confusion

PR #9583, opened 2024-05-02
runtime: make kata-runtime check error more understandable when /dev/kvm doesn't exist

cncal:
If device /dev/kvm does not exist, kata-runtime check would fail with an ambiguous error messae 'no such file or directory'. I added a little more details to make it understandable and it will belike:

```
ERRO[0000] cannot open kvm device: no such file or directory  arch=arm64 check-type=full device=/dev/kvm name=kata-runtime pid=2849085 source=runtime
ERRO[0000] no such file or directory                          arch=arm64 name=kata-runtime pid=2849085 source=runtime
no such file or directory
```


stevenhorsman (commented):
Hi @cncal - you PRs are both failing the commit check. Just to remind you the required commit format is documented here: https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

For both your PRs I think `runtime` is the correct subsystem to use. Thanks

stevenhorsman (approved):

Issue #39, opened 2019-04-15
tracking issue for github label improvements


jodh-intel:
We need to make better use of GitHub labels. This issue is for tracking required repo, infrastructure and process changes.

See the main PR at: https://github.com/kata-containers/tests/pull/1466.

# TODO

- [x] Land [labels database and tooling](https://github.com/kata-containers/tests/pull/1466).
- [ ] Update `github-labels.sh` to add support for setting the labels on a github repo.
  That work *might* be using a tool like [`labeler`](https://github.com/tonglil/labeler) (which supports the format of the `labels.yaml.in` on this PR). However, the `labeler` tool hasn't been updated for a while and doesn't currently support handling Label descriptions (essential), although there is an [old unmerged PR that add this support](https://github.com/tonglil/labeler/pull/37).
- [x] Apply the labels database to the least used repo (https://github.com/kata-containers/www.katacontainers.io) for testing.
  - [x] Create a new `.github` **repo** to store common GitHub metadata files (currently stored in each repos `.github/` directory): https://help.github.com/en/articles/creating-a-default-community-health-file-for-your-organization
  - [x] Create a new `ISSUE_TEMPLATE.md` in the `.github` **repo** including metadata for auto-setting labels.
  - [x] Check new template works for https://github.com/kata-containers/www.katacontainers.io.
- [x] Add `needs-review`:
  - [x] To the labels db: https://github.com/kata-containers/tests/pull/1653.
  - [x] To all the issue templates in https://github.com/kata-containers/.github/tree/master/.github/ISSUE_TEMPLATE.
- [x] Get consensus that "all looks good" (@egernst and @gnawux happy with labels).
- [x] Switch over to new labels:
  - [x] Apply the labels to all other repos.
  - [ ] Remove all the `.github/*` files from the other repos:
    - [ ] agent
    - [ ] ksm-throttler
    - [ ] kata-containers
    - [ ] packaging
    - [ ] proxy
    - [ ] runtime
    - [ ] shim
  - [ ] Create issue templates for the following (including labels in templates):
    - [x] bug (https://github.com/kata-containers/.github/blob/master/.github/ISSUE_TEMPLATE/bug_report.md).
    - [x] feature (https://github.com/kata-containers/.github/blob/master/.github/ISSUE_TEMPLATE/feature_request.md)
    - [x] enhancement (https://github.com/kata-containers/.github/blob/master/.github/ISSUE_TEMPLATE/enhancement-request.md)
[...]

PR #9584, opened 2024-05-02
kata-agent: Add CDI support

zvonkok:
For proper device handling add CDI support
Depends on: ~~https://github.com/kata-containers/kata-containers/pull/9944~~ ~~https://github.com/kata-containers/kata-containers/issues/10059~~

### DIFF
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -84,8 +84,7 @@ regorus = { version = "0.1.4", default-features = false, features = [
     "arc",
     "regex",
 ], optional = true }
-#cdi = { git = "https://github.com/cncf-tags/container-device-interface-rs", rev = "19763bcf0de62599ce8b88558418b4dac1154745" }
-cdi = { git = "https://github.com/zvonkok/container-device-interface-rs", branch = "rename-with-auto-refresh" }
+cdi = { git = "https://github.com/cncf-tags/container-device-interface-rs", rev = "fba5677a8e7cc962fc6e495fcec98d7d765e332a" }
 
 [dev-dependencies]
 tempfile = "3.1.0"
--- a/src/agent/Cargo.toml
+++ b/src/agent/Cargo.toml
@@ -84,6 +84,8 @@ regorus = { version = "0.1.4", default-features = false, features = [
     "arc",
     "regex",
 ], optional = true }
+#cdi = { git = "https://github.com/cncf-tags/container-device-interface-rs", rev = "19763bcf0de62599ce8b88558418b4dac1154745" }
+cdi = { git = "https://github.com/zvonkok/container-device-interface-rs", branch = "rename-with-auto-refresh" }
 
 [dev-dependencies]
 tempfile = "3.1.0"
--- a/src/agent/src/device/mod.rs
+++ b/src/agent/src/device/mod.rs
@@ -11,6 +11,9 @@ use self::vfio_device_handler::{VfioApDeviceHandler, VfioPciDeviceHandler};
 use crate::pci;
 use crate::sandbox::Sandbox;
 use anyhow::{anyhow, Context, Result};
+use cdi::annotations::parse_annotations;
+use cdi::cache::{new_cache, CdiOption, with_auto_refresh};
+use cdi::spec_dirs::with_spec_dirs;
 use kata_types::device::DeviceHandlerManager;
 use nix::sys::stat;
 use oci::{LinuxDeviceCgroup, Spec};
@@ -25,6 +28,8 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
 use tokio::sync::Mutex;
+use tokio::time;
+use tokio::time::Duration;
 use tracing::instrument;
 
 pub mod block_device_handler;
@@ -34,6 +39,8 @@ pub mod scsi_device_handler;
 pub mod vfio_device_handler;
 
 pub const BLOCK: &str = "block";
+
+
 
 #[derive(Debug, Clone)]
 pub struct DeviceInfo {
@@ -236,6 +243,61 @@ pub async fn add_devices(
         update_env_pci(env_vec, &sandbox.lock().await.pcimap)?
     }
     update_spec_devices(logger, spec, dev_updates)
+}
+
+#[instrument]
+pub async fn handle_cdi_devices(logger: &Logger, spec: &mut Spec, spec_dir: &str, cdi_timeout: u64) -> Result<()> {
+    if let Some(container_type) = spec
+        .annotations()
+        .as_ref()
+        .and_then(|a| a.get("io.katacontainers.pkg.oci.container_type"))
+    {
+        if container_type == "pod_sandbox" {
+            return Ok(());
+        }
+    }
+
+    let (_, devices) = parse_annotations(spec.annotations().as_ref().unwrap())?;
+
+    if devices.is_empty() {
+        info!(logger, "no CDI annotations, no devices to inject");
+        return Ok(());
+    }
+    // Explicitly set the cache options to disable auto-refresh and
+    // to use the single spec dir "/var/run/cdi" for tests it can be overridden
+    let options: Vec<CdiOption> = vec![
+        with_auto_refresh(true),
+        with_spec_dirs(&[spec_dir]),
+    ];
+
+    let cache: Arc<std::sync::Mutex<cdi::cache::Cache>> = new_cache(options);
+
+    for _ in 0..=cdi_timeout {
+        let inject_result = {
+            // Lock cache within this scope, std::sync::Mutex has no Send
+            // and await will not work with time::sleep
+            let mut cache = cache.lock().unwrap();
+            cache.inject_devices(Some(spec), devices.clone())
+        };
+
+        match inject_result {
+            Ok(_) => {
+                info!(
+                    logger,
+                    "all devices injected successfully, modified CDI container spec: {:?}", &spec
+                );
+                return Ok(());
+            }
+            Err(e) => {
+                info!(logger, "error injecting devices: {:?}", e);
+            }
+        }
+        time::sleep(Duration::from_millis(1000)).await;
+    }
+    Err(anyhow!(
+        "failed to inject devices after CDI timeout of {} seconds",
+        cdi_timeout
+    ))
 }
 
 #[instrument]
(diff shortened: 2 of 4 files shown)


src/agent/src/device.rs
```diff
@@ -911,6 +912,76 @@ async fn vfio_ap_device_handler(_: &Device, _: &Arc<Mutex<Sandbox>>) -> Result<S
...
+        let cdi_oci_config = Path::new("/tmp/cdi-oci-config.json");
+
+        spec.save(raw_oci_config.to_str().unwrap())?;         
+ 
+        if !which("/usr/bin/cdi").is_ok() {
+            error!(sl(), "cdi not found in path");
+            return Ok(());
+        }
+
+        match  tokio::process::Command::new("/usr/bin/cdi")
```
zvonkok, line 939: This needs to be replaced with proper Rust implementation of CDI. For now using a go binary to do the proper injection of devices. Until then this PR remains a draft.

Apokleos: Hi Zvonko, does just execute cmd `cdi raw_oci_config cdi_oci_config` enough ? there's no such usage of cdi when I try to execute /usr/bin/cdi. Is it some specified version of cdi ?

zvonkok: This is a stripped down version of cdi but we could use the official one.

zvonkok:
@Apokleos Should we make the timeout configurable via kernel cmdline like we have for the `hotplug_timeout` ?

Apokleos (commented):
Hi @zvonkok great work! I have questions about this PR:
1. This patch now is limited to scenarios where containerd/CDI relies on CDI Annotations for configuration, right ?
2. How about the other one which containerd/CDI enabled and we can directly use `CDIDevices` instead of `Annotations`? I think we need do support it.

zvonkok:
This is only for the inner runtime aka kata-agent, this is independent of containerd. 
The flow is as follows: 
```
Pod allocates a CDI device  [nvidia.com/vfio=18]  # At this stage we have only VFIO devices 
    |
    v
Kubernetes sends the CDI device over CRI 
    |
    v
containerd reads the CDI device spec and creates OCI spec with LinuxDevice: /dev/vfio/18
    |
    v
go-runtime (outer runtime) receives OCI spec with /dev/vfio/18
    |
    v
hot-plug of /dev/vfio/18 to the VM # depending on the vfio_mode it becomes a device or vfio device inside the VM
                                   # we're assuming here guest_kernel mode hence the VFIO device becomes a GPU
    |
    v
go runtime: create annotation from actual GPU device # CDI spec with /dev/nvidia; /usr/bin/nvidai-smi; /ust/lib/libcuda.so .... 
# at this stage inside the VM we do not have a VFIO device but an actual GPU we need the CDI spec created inside the VM not outside of it 
    |
    v
go runtime adds nvidia.com/gpu=0 the index depends on the BDF lowest BDF == index 0, next higher BDF index 1, etc
    |
    v
go runtime send modified OCI spec to kata-agent
    |
    v
kata-agent reads the OCI spec and modifies the OCI spec according to the CDI spec created inside the VM 
    |
    v
kata-agent creates OCI spec with all the GPU containerEdits  /dev/nvidia; /usr/bin/nvidai-smi; /ust/lib/libcuda.so .... 
```

zvonkok:
containerd will only create a LinuxDevice from a CDI device in the CRI or passthrough the CDI device as an annotation for the Kubernetes KEP

zvonkok:
See also the presentation I did in the AC meeting about the flow.

Apokleos:
> This is only for the inner runtime aka kata-agent, this is independent of containerd. The flow is as follows:
> 
> ```
> Pod allocates a CDI device  [nvidia.com/vfio=18]  # At this stage we have only VFIO devices 
>     |
>     v
> Kubernetes sends the CDI device over CRI 
>     |
>     v
> containerd reads the CDI device spec and creates OCI spec with LinuxDevice: /dev/vfio/18
>     |
>     v
> go-runtime (outer runtime) receives OCI spec with /dev/vfio/18
>     |
>     v
> hot-plug of /dev/vfio/18 to the VM # depending on the vfio_mode it becomes a device or vfio device inside the VM
>                                    # we're assuming here guest_kernel mode hence the VFIO device becomes a GPU
>     |
>     v
> go runtime: create annotation from actual GPU device # CDI spec with /dev/nvidia; /usr/bin/nvidai-smi; /ust/lib/libcuda.so .... 
> # at this stage inside the VM we do not have a VFIO device but an actual GPU we need the CDI spec created inside the VM not outside of it 
>     |
>     v
> go runtime adds nvidia.com/gpu=0 the index depends on the BDF lowest BDF == index 0, next higher BDF index 1, etc
>     |
>     v
> go runtime send modified OCI spec to kata-agent
>     |
>     v
> kata-agent reads the OCI spec and modifies the OCI spec according to the CDI spec created inside the VM 
>     |
>     v
> kata-agent creates OCI spec with all the GPU containerEdits  /dev/nvidia; /usr/bin/nvidai-smi; /ust/lib/libcuda.so .... 
> ```
>
Great. Epiphany! the flow gives me a completely new perspective on this issue.

src/agent/Cargo.toml
```diff
@@ -84,6 +84,7 @@ regorus = { version = "0.1.4", default-features = false, features = [
     "arc",
     "regex",
 ], optional = true }
+cdi = { git = "https://github.com/zvonkok/container-device-interface-rs", branch = "main" }
```
Apokleos, line 87: As cdi-rs has been transferred to CNCF-Tags, IMO, it should replaced now !
```suggestion
cdi = { git = "https://github.com/cncf-tags/container-device-interface-rs", branch = "main" }
```

zvonkok: Definitely!

Apokleos:
> @Apokleos Should we make the timeout configurable via kernel cmdline like we have for the `hotplug_timeout` ?
>
Yes, agree with configurable `TIMEOUT_LIMIT` via kernel cmdline.

zvonkok:
@Apokleos PTAL, lets introduce the timeout via kernel in the next PR. I want to have a stable set first working and for now all use-cases work with the default timeout.

Apokleos:
It seems that you should do "make check" to make CI happy

Apokleos (approved):
Thx Zvonko @zvonkok , this PR almost looks good to me and I have tested it. But the CIs required failed with `make check` failure, which you should correct.

src/agent/src/device/mod.rs
```diff
@@ -238,6 +244,49 @@ pub async fn add_devices(
...
+    }
+
+    for _ in 0..=CDI_TIMEOUT_LIMIT {
+        match inject_devices(spec, devices.clone()) {
+            Ok(_) => {
+                info!(logger, "all devices injected successfully");
+                return Ok(());
+            }
+            Err(e) => {
+                info!(logger, "error injecting devices: {:?}", e);
```
ChengyuZhu6, line 279: Do you think we can skip the loop if we get an error and return the result?

src/agent/src/device/mod.rs
```diff
@@ -238,6 +244,49 @@ pub async fn add_devices(
...
+        .as_ref()
+        .unwrap()
+        .get("io.katacontainers.pkg.oci.container_type")
+    {
+        if container_type == "pod_sandbox" {
+            return Ok(());
+        }
+    }
+
+    let (_keys, devices) = parse_annotations(&spec.annotations().as_ref().unwrap())?;
```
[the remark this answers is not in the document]
ChengyuZhu6: I think we can ignore `_keys` if we would not use it.

ChengyuZhu6 (commented):
I’m not an expert on this, but I have a few minor comments and questions. Please feel free to correct me if I’m wrong.

ChengyuZhu6:
Could we add some tests to ensure that this feature remains stable and isn’t affected by other patches?

src/agent/src/device/mod.rs
Apokleos: There's a timeout CDI_TIMEOUT_LIMIT, it will do its best to inject devices till it exceeds CDI_TIMEOUT_LIMIT.
In this aspect, I think it's fine to do so.

src/agent/Cargo.toml
```diff
@@ -84,6 +84,7 @@ regorus = { version = "0.1.4", default-features = false, features = [
     "arc",
     "regex",
 ], optional = true }
+cdi = { git = "https://github.com/cncf-tags/container-device-interface-rs", branch = "main" }
```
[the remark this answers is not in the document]
zvonkok: Indeed, but we just moved to the new repository and just got admin rights. We need to build a release pipeline for this.

src/agent/src/device/mod.rs
zvonkok: We cannot just return since we do not know when the hot-plug event and CDI spec creation are complete.

src/agent/src/device/mod.rs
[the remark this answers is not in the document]
zvonkok: Ignoring keys.

src/agent/Cargo.toml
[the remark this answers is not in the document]
zvonkok: Updated to rev="commit-hash" until we have a proper release pipeline.

src/agent/src/device/mod.rs
```diff
@@ -238,6 +244,49 @@ pub async fn add_devices(
...
+pub async fn handle_cdi_devices(
+    logger: &Logger,
+    devices: &[Device],
+    spec: &mut Spec,
+    sandbox: &Arc<Mutex<Sandbox>>,
+) -> Result<()> {
+    if let Some(container_type) = spec
+        .annotations()
+        .as_ref()
+        .unwrap()
```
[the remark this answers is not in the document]
zvonkok: Done, using `and_then`

zvonkok:
@ChengyuZhu6 PTAL updated.

ChengyuZhu6:
> @ChengyuZhu6 PTAL updated.

@zvonkok  Thanks for addressing my comments.

src/agent/src/rpc.rs
```diff
@@ -224,6 +224,14 @@ impl AgentService {
         // cannot predict everything from the caller.
         add_devices(&sl(), &req.devices, &mut oci, &self.sandbox).await?;
 
+        // In guest-kernel mode some devices need extra handling. Taking the
+        // GPU as an example the shim will inject CDI annotations that will
+        // be used by the kata-agent to do containerEdits according to the
+        // CDI spec coming from a registry that is created on the fly by UDEV
+        // or other entities for a specifc device.
+        handle_cdi_devices(&sl(), &req.devices, &mut oci, &self.sandbox).await?;
+        info!(sl(), "modified CDI container spec: {:?}", &oci);
```
ChengyuZhu6, line 233: After handling cdi devices, it would be better if we can first verify whether the container is cdi container. If it is, we can then print the modified spec. Currently, the code prints 'modified CDI container spec' by default, regardless of whether it is a CDI container or not, which can cause confusion when checking the logs.

ChengyuZhu6 (commented):
@zvonkok

src/agent/src/rpc.rs
zvonkok: Good catch. Will update.

zvonkok:
> Could we add some tests to ensure that this feature remains stable and isn’t affected by other patches?
The problem here is that we need root write permissions since the default folder are `/etc/cdi` and `/var/run/cdi` both not writable by user. 
I think we need to update the CDI spec first to allow user writable directories as default ones. Will bring this up with the CDI WG.

src/agent/src/rpc.rs
zvonkok: @ChengyuZhu6 Updated, only if we inject devices which means we have done containeredits we will print out the new spec.

ChengyuZhu6:
> > Could we add some tests to ensure that this feature remains stable and isn’t affected by other patches?
> 
> @ChengyuZhu6 @Apokleos The problem is that we need root write permissions since the default folders are `/etc/cdi` and `/var/run/cdi`, both of which are not writable by the user. I think we need to update the CDI spec first to allow user-writable directories as the default ones. I will bring this up with the CDI WG.
> 
> Or I rewrite handle_cdi_devices to allow overriding the default directory so we can easily add a test-case without needing root privileges.

@zvonkok Thanks for pointing out. I prefer the second option. We can pass the annotation for the cdi spec directory to the vm, such as `cdi_spec_directory`, and override the directory if we detect this annotation when parsing spec annotations.

zvonkok:
@ChengyuZhu6 We do not need to pass anything to the VM. CDI has a default set of directories, and we should stick to it. The only thing needed here is the ability to override the `inject_devices` function to accept an additional dirs parameter. 
If supplied update the registry with new dir if not ignore and use the defaults.

zvonkok:
@ChengyuZhu6 @Apokleos Added test case we need the changes from https://github.com/cncf-tags/container-device-interface-rs/pull/39 first and then update the 'rev' entry in Config.toml. Nevertheless PTAL at both

ChengyuZhu6:
> @ChengyuZhu6 We do not need to pass anything to the VM. CDI has a default set of directories, and we should stick to it. The only thing needed here is the ability to override the `inject_devices` function to accept an additional dirs parameter. If supplied update the registry with new dir if not ignore and use the defaults.

Makes sense to me.

ChengyuZhu6:
> @ChengyuZhu6 @Apokleos Added test case we need the changes from [cncf-tags/container-device-interface-rs#39](https://github.com/cncf-tags/container-device-interface-rs/pull/39) first and then update the 'rev' entry in Config.toml. Nevertheless PTAL at both

Looks good. Let's wait for the PR to be merged.

Apokleos:
> @ChengyuZhu6 @Apokleos Added test case we need the changes from [cncf-tags/container-device-interface-rs#39](https://github.com/cncf-tags/container-device-interface-rs/pull/39) first and then update the 'rev' entry in Config.toml. Nevertheless PTAL at both
>
sure, the PR #39 is merged now.

ChengyuZhu6 (approved):
LGTM, thanks!

PR #9585, opened 2024-05-02
debugging: adding a script and instructions for debugging the GO shim

littlejawa:
This PR provides some developer's documentation and a helper script that makes it easier to use a debugger on the go shim.

NOTE: This PR doesn't address runtime-rs, it will come with a future commit.
Now the script provided here will probably be useful for runtime-rs too, as most of it is related to the shim's behavior, dictated by the shimv2 interface.
My expectation at this point is that the same script should be usable for both shims.

### DIFF
--- a/docs/Debug-shim-guide.md
+++ b/docs/Debug-shim-guide.md
@@ -1,0 +1,185 @@
+# Using a debugger with the runtime
+
+Setting up a debugger for the runtime is pretty complex: the shim is a server
+process that is run by the runtime manager (containerd/CRI-O), and controlled by
+sending gRPC requests to it.
+Starting the shim with a debugger then just gives you a process that waits for
+commands on its socket, and if the runtime manager doesn't start it, it won't
+send request to it.
+
+A first method is to attach a debugger to the process that was started by the
+runtime manager.
+If the issue you're trying to debug is not located at container creation, this
+is probably the easiest method.
+
+The other method involves a script that is placed in between the runtime manager
+and the actual shim binary. This allows to start the shim with a debugger, and
+wait for a client debugger connection before execution, allowing debugging of the
+kata runtime from the very beginning.
+
+## Prerequisite
+
+At the time of writing, a debugger was used only with the go shim, but a similar
+process should be doable with runtime-rs. This documentation will be enhanced
+with rust-specific instructions later on.
+
+In order to debug the go runtime, you need to use the [Delve debugger](https://github.com/go-delve/delve).
+
+You will also need to build the shim binary with debug flags to make sure symbols
+are available to the debugger.
+Typically, the flags should be: `-gcflags=all=-N -l`
+
+## Attach to the running process
+
+To attach the debugger to the running process, all you need is to let the container
+start as usual, then use the following command with `dlv`:
+
+`$ dlv attach [pid of your kata shim]`
+
+If you need to use your debugger remotely, you can use the following on your target
+machine:
+
+`$ dlv attach [pid of your kata shim] --headless --listen=[IP:port]`
+
+then from your client computer:
+
+`$ dlv connect [IP:port]`
+
+## Make CRI-O/containerd start the shim with the debugger
+
+You can use the [this script](../tools/containerd-shim-katadbg-v2) to make the
+shim binary executed through a debugger, and make the debugger wait for a client
+connection before running the shim.
+This allows starting your container, connecting your debugger, and controlling the
+shim execution from the beginning.
+
+### Adapt the script to your setup
+
+You need to edit the script itself to give it the actual binary
+to execute.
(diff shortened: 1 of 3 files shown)


littlejawa:
Updated instructions. Now ready for review.

docs/Developer-Guide.md
```diff
@@ -771,6 +771,189 @@ $ sudo su -c 'cd /var/run/vc/vm/${sandbox_id} && socat "stdin,raw,echo=0,escape=
...
+This is because the first thing that the shim will do is to daemonize itself,
+i.e: start itself as a subprocess, and exit. So you really want the debugger
+to attach to the child process.
+
+```bash
+(dlv) target follow-exec -on .*/__debug_bin
+```
+
+Note that we are providing a regular expression to filter the name of the binary.
+This is to make sure that the deubgger attaches to the runtime shim, and not
```
gkurz, line 910: s/deubgger/debugger

beraldoleal (approved):
Hi @littlejawa this is really welcome. Thanks!

My only suggestion is to move it to its own file, something like DEBUG-GOSHIM.md and make reference to it in the Developer Guide. Otherwise the guide it will be really extensive.

gkurz:
> Hi @littlejawa this is really welcome. Thanks!
> 
> My only suggestion is to move it to its own file, something like DEBUG-GOSHIM.md and make reference to it in the Developer Guide. Otherwise the guide it will be really extensive.

I like the idea as `docs/Developer-Guide.md` is big enough already.

@littlejawa can you do that please ?

littlejawa:
> > Hi @littlejawa this is really welcome. Thanks!
> > My only suggestion is to move it to its own file, something like DEBUG-GOSHIM.md and make reference to it in the Developer Guide. Otherwise the guide it will be really extensive.
> 
> I like the idea as `docs/Developer-Guide.md` is big enough already.
> 
> @littlejawa can you do that please ?

Sure. I actually hesitated when doing it, so I'm fine with it.
I think we shouldn't name it "*goshim*" though, as I expect the rust debugging setup to be pretty similar. Both will probably end up in the same doc, if and when we document the rust side.

docs/Debug-shim-guide.md
```diff
@@ -0,0 +1,183 @@
...
+### Start your container and connect to the debugger
+
+Once the above configuration is in place, you can start your container, using
+your "katadbg" runtime class.
+
+E.g: `$ crictl runp --runtime=katadbg sandbox.json`
+
+The command will hang, and you can see that a `dlv` process is started
+
+```bash
```
gkurz, line 115: `./tests/kata-doc-to-script.sh` assumes such code blocks to only contain code. Here this is rather the output of `ps aux | grep dlv`. Please make it an simple ` ``` `, this should fix the static-checks CI.

docs/Debug-shim-guide.md
```diff
@@ -0,0 +1,183 @@
...
+$ ps aux | grep dlv
+root        9137  1.4  6.8 6231104 273980 pts/10 Sl   15:04   0:02 dlv exec /go/src/github.com/kata-containers/kata-containers/src/runtime/__debug_bin --headless --listen=:12345 --accept-multiclient -r stdout:/tmp/shim_output_oMC6Jo -r stderr:/tmp/shim_output_oMC6Jo -- -namespace default -address  -publish-binary /usr/local/bin/crio -id 0bc23d2208d4ff8c407a80cd5635610e772cae36c73d512824490ef671be9293 -debug start
+```
+
+Then you can use the dlv debugger to connect to it:
+
+```bash
```
gkurz, line 122: ditto

docs/Debug-shim-guide.md
```diff
@@ -0,0 +1,183 @@
...
+(dlv)
+```
+
+Before doing anything else, you need to to enable `follow-exec` mode in delve.
+This is because the first thing that the shim will do is to daemonize itself,
+i.e: start itself as a subprocess, and exit. So you really want the debugger
+to attach to the child process.
+
+```bash
+(dlv) target follow-exec -on .*/__debug_bin
```
gkurz, line 134: ditto

docs/Debug-shim-guide.md
```diff
@@ -0,0 +1,183 @@
...
+
+Note that we are providing a regular expression to filter the name of the binary.
+This is to make sure that the debugger attaches to the runtime shim, and not
+to other subprocesses (hypervisor typically).
+
+To ease this process, we recommand the use of an init file containing the above
+command.
+
+```bash
+$ cat dlv.ini
```
gkurz, line 145: ditto

littlejawa:
Ok... Spell check is happy now :)

gkurz (approved):
LGTM. Thanks @littlejawa !

PR #9592, opened 2024-05-03
tests: adapt Mariner CI to unblock CH v39 upgrade

sprt:
```
The CH v39 upgrade in #9575 is currently blocked because of a bug in the
Mariner host kernel. To address this, we temporarily tweak the Mariner
CI to use an Ubuntu host and the Kata guest kernel, while retaining the
Mariner initrd. This is tracked in #9594.

Importantly, this allows us to preserve CI for genpolicy. We had to
tweak the default rules.rego however, as the OCI version is now
different in the Ubuntu host. This is tracked in #9593.

This change has been tested together with CH v39 in #9588.
```

### DIFF
--- a/src/tools/genpolicy/rules.rego
+++ b/src/tools/genpolicy/rules.rego
@@ -66,7 +66,8 @@ CreateContainerRequest {
     p_oci := p_container.OCI
 
     print("CreateContainerRequest: p Version =", p_oci.Version, "i Version =", i_oci.Version)
-    p_oci.Version == i_oci.Version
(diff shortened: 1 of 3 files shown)


ananos (approved):
go ahead and unblock the clh upgrade ;-)

danmihai1 (approved):

stevenhorsman (approved):
Looks okay to me. Thanks!

likebreath (approved):
Thank you for fast turn around.

GabyCT (approved):

Issue #9597, opened 2024-05-06
runtime-rs: when start sandbox failed, the dead shim process left 


lifupan:
Once the dragonball vm start failed, the pod would start failed, and the dead shim process left which should be killed once the 
pod start failed.

PR #9598, opened 2024-05-06
runtime-rs: fix the issue of the leak of dead shim

lifupan:
We should init and asign the runtime instance to runtime handler, otherwise, if the pause container failed to start, which means the runtime instance failed to start, then the following delete & shutdown request wouldn't be run, thus the dead shim would be left.

Fixes: #9597


Apokleos (approved):
Thx bro @lifupan LGTM!

justxuewei (approved):
Lgtm, thanks!

GabyCT (approved):

PR #9601, opened 2024-05-07
qemu: the error is logged only when it occurs

cncal:
Everytime I create contianer on arm64 machine, containerd/kata logs a redundant warning as follows:

``` shell
time="2024-05-07" level=warning msg="<nil>" arch=arm64 name=containerd-shim-v2 pid=xxx sandbox=fdd1f05 source=virtcontainers/hypervisor
```

I added an error statement so that the error would be logged when it occurs.


lifupan (approved):

stevenhorsman (approved):
LGTM. Thanks!

Issue #9603, opened 2024-05-07
runtime-rs: add QMP support for Qemu


Apokleos:
The whole feature contains two parts:

- [ ] add support QMP socket in cmdline (CR)
- [ ] add support QMP channel for managing many devices (WIP)

PR #9571, opened 2024-04-29
runtime-rs: Fix constructing the RTC struct

emanuellima1:
RTC was being built in a wrong fashion on PR 9519

PR #9604, opened 2024-05-07
runtime-rs: add QMP support for Qemu(part I)

Apokleos:
add support QMP socket in cmdline

Fixes: #9603

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/mod.rs
+++ b/src/libs/kata-types/src/config/hypervisor/mod.rs
@@ -411,6 +411,20 @@ pub struct DebugInfo {
     /// much disk space.
     #[serde(default)]
     pub guest_memory_dump_path: String,
+
+    /// This option allows to add a debug monitor socket when `enable_debug = true`
+    /// WARNING: Anyone with access to the monitor socket can take full control of
+    /// Qemu. This is for debugging purpose only and must *NEVER* be used in
+    /// production.
+    /// Valid values are :
+    /// - "hmp"
+    /// - "qmp"
+    /// - "qmp-pretty" (same as "qmp" with pretty json formatting)
+    /// If set to the empty string "", no debug monitor socket is added. This is
+    /// the default.
+    /// dbg_monitor_socket = "hmp"
+    #[serde(default)]
+    pub dbg_monitor_socket: String,
 }
 
 impl DebugInfo {
--- a/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
@@ -11,13 +11,18 @@ use async_trait::async_trait;
 use std::collections::HashMap;
 use std::fmt::Display;
 use std::fs::{read_to_string, File};
-use std::os::fd::AsRawFd;
+use std::os::fd::{AsRawFd, FromRawFd, IntoRawFd};
+use std::os::unix::net::UnixListener;
+use std::path::PathBuf;
 use tokio;
 
 // These should have been called MiB and GiB for better readability but the
 // more fitting names unfortunately generate linter warnings.
 const MI_B: u64 = 1024 * 1024;
 const GI_B: u64 = 1024 * MI_B;
+
+const QMP_SOCKET_FILE: &str = "qmp.sock";
+const DEBUG_MONITOR_SOCKET: &str = "debug-monitor.sock";
 
 // The approach taken here is inspired by govmm.  We build structs, each
 // corresponding to a qemu command line parameter, like Kernel, or a device,
@@ -1162,6 +1167,121 @@ impl ToQemuParams for DeviceIntelIommu {
     }
 }
 
+// Qemu provides methods and types for managing QEMU instances.
+// To manage a qemu instance after it has been launched you need
+// to pass the -qmp option during launch requesting the qemu instance
+// to create a QMP unix domain manageent socket, e.g.,
+// -qmp unix:fd=SOCK_FD,server=on,wait=off.
+// -monitor unix:path=SOCK_PATH,server=on,wait=off.
+#[derive(Debug, Default, PartialEq)]
+pub enum MonitorProtocol {
+    // Socket using a human-friendly text-based protocol.
+    Hmp,
+
+    // Socket using a richer json-based protocol.
+    #[default]
+    Qmp,
+
+    // Same as Qmp with pretty json formatting.
+    QmpPretty,
+}
+
+impl MonitorProtocol {
+    pub fn new(proto: &str) -> Self {
+        match proto {
+            "hmp" => MonitorProtocol::Hmp,
+            "qmp-pretty" => MonitorProtocol::QmpPretty,
+            _ => MonitorProtocol::Qmp,
+        }
+    }
+}
+
+impl ToString for MonitorProtocol {
+    fn to_string(&self) -> String {
+        match *self {
+            MonitorProtocol::Hmp => "monitor".to_string(),
+            MonitorProtocol::QmpPretty => "qmp-pretty".to_string(),
+            _ => "qmp".to_string(),
+        }
+    }
+}
+
+#[derive(Debug)]
+enum QmpSockType {
+    Fd(File),
+    Path(PathBuf),
+}
+
+#[derive(Debug)]
+pub struct QmpSocket {
+    // protocol to be used on the socket.
+    protocol: MonitorProtocol,
+    // QMP unix socket to be passed to qemu
+    address: QmpSockType,
+    // server tells if this is a server socket.
+    server: bool,
+    // nowait tells if qemu should block waiting for a client to connect.
+    nowait: bool,
+}
+
+impl QmpSocket {
+    fn new(proto: MonitorProtocol) -> Result<Self> {
+        let qmp_socket = match proto {
+            MonitorProtocol::Qmp | MonitorProtocol::QmpPretty => {
+                // let sock_path = root_path.join(QMP_SOCKET_FILE);
+                let listener =
+                    UnixListener::bind(QMP_SOCKET_FILE).context("unix listener bind failed.")?;
+                let raw_fd = listener.into_raw_fd();
+                clear_cloexec(raw_fd).context("clearing unix listenser O_CLOEXEC failed")?;
+                let sock_file = unsafe { File::from_raw_fd(raw_fd) };
+                // The default QMP socket or called base socket is qmp.sock.
+                QmpSocket {
+                    protocol: MonitorProtocol::new("qmp"),
+                    address: QmpSockType::Fd(sock_file),
+                    server: true,
+                    nowait: true,
+                }
+            }
+            MonitorProtocol::Hmp => {
+                // If extra monitor needed, HMP socket with qmp-extra.sock will be added.
+                QmpSocket {
+                    protocol: MonitorProtocol::new("hmp"),
+                    address: QmpSockType::Path(PathBuf::from(DEBUG_MONITOR_SOCKET)),
+                    server: true,
+                    nowait: true,
+                }
+            }
+        };
+
+        Ok(qmp_socket)
+    }
+}
+
+#[async_trait]
+impl ToQemuParams for QmpSocket {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        let param_qmp = format!("-{}", self.protocol.to_string());
+
+        let mut params: Vec<String> = Vec::new();
+
+        match &self.address {
+            // -qmp unix:fd=SOCK_FD,server=on,wait=off
+            QmpSockType::Fd(f) => params.push(format!("unix:fd={}", f.as_raw_fd())),
+            // -monitor unix:path=SOCK_PATH,server=on,wait=off
+            QmpSockType::Path(p) => params.push(format!("unix:path={}", p.display())),
+        }
+
+        if self.server {
+            params.push("server=on".to_owned());
+            if self.nowait {
+                params.push("wait=off".to_owned());
+            }
+        }
+
+        Ok(vec![param_qmp, params.join(",")])
+    }
+}
+
 fn is_running_in_vm() -> Result<bool> {
     let res = read_to_string("/proc/cpuinfo")?
         .lines()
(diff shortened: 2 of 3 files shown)


Apokleos:
This patch can make kata/qemu running with qmp socket set well ! and the dumped cmdline as below:
```
/opt/kata/bin/qemu-system-x86_64 -name sandbox-1e28630de2b9b2e8bf22714561899249f6254fc8a9d167c454f3112f088a0710 -kernel /opt/kata/share/kata-containers/vmlinux-6.1.52-116 -append reboot=k panic=1 systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro ro rootfstype=ext4 -qmp unix:fd=14,server=on,wait=off -smp 1,maxcpus=6 -machine q35,accel=kvm,nvdimm=on -cpu host,pmu=off -m 2G,slots=128,maxmem=15668M -object memory-backend-file,id=entire-guest-memory-share,mem-path=/dev/shm,size=2G,share=on,readonly=off -rtc base=utc,clock=host,driftfix=slew -device vhost-vsock-pci,vhostfd=15,guest-cid=3259189863 -netdev tap,id=network-0,vhost=on,vhostfds=18,fds=17 -device virtio-net-pci,netdev=network-0,mac=9a:6a:d4:5f:60:60,mq=on,vectors=4 -chardev socket,id=virtiofsd-chardev,path=virtiofsd.sock -device vhost-user-fs-pci,chardev=virtiofsd-chardev,tag=kataShared -numa node,memdev=entire-guest-memory-share -object memory-backend-file,id=TODO,mem-path=/opt/kata/share/kata-containers/kata-ubuntu-latest.image,size=128M,share=off,readonly=on -device nvdimm,memdev=TODO,unarmed=on -device virtio-serial-pci,id=serial0 -device virtconsole,id=console0,chardev=charconsole0 -chardev socket,id=charconsole0,server=on,wait=off,path=console.sock -vga none -no-user-config -nodefaults -nographic -no-reboot
```

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1178,6 +1183,133 @@ impl ToQemuParams for DeviceIntelIommu {
...
+        })
+    }
+
+    // validate returns true if the QmpSocket structure is valid and complete.
+    fn validate(&self) -> bool {
+        if self.socket_type.as_str() != QMP_SOCKET_TYPE {
+            return false;
+        }
+
+        if self.name.is_empty() && self.fd.is_none() {
```
BbolroC, line 1265: I was wondering if `||` would be more to catch invalid cases.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1238,16 +1240,44 @@ pub struct QmpSocket {
...
+            fd: Some(sock_file),
             name: QMP_SOCKET_FILE.to_owned(),
             server: true,
             nowait: true,
         })
     }
+
+    // validate returns true if the QmpSocket structure is valid and complete.
+    #[allow(dead_code)]
+    fn validate(&self) -> bool {
```
pmores, line 1260: We don't use `validate()` functions in qemu-rs.  They are used in govmm since its ambition was (initially) much broader so it made sense for its code to be more general.  Where an object set-up depends on outside information like user input it makes sense to validate the result.

In qemu-rs however we're generally in control of our objects' set-up.  (Even where we aren't we prefer detecting problems earlier.)  That's exactly the case for `QmpSocket` - `validate()` is reduced just to restate what the constructor guarantees already and thus seems superfluous.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1178,6 +1183,133 @@ impl ToQemuParams for DeviceIntelIommu {
...
+
+        true
+    }
+}
+
+#[async_trait]
+impl ToQemuParams for QmpSocket {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        if !self.validate() {
+            return Err(anyhow!("validate qmp socket failed"));
```
BbolroC, line 1284: It would be more helpful to have detailed info about the specific condition under which the parameters fail.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
BbolroC: My bad. Sorry for lack of knowledge. `name` and `fd` should not exist together. Please ignore my comment.

BbolroC (approved):
Other than an ignorant question regarding a log message for validation. It looks good to me. Thanks @Apokleos !

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1178,6 +1183,133 @@ impl ToQemuParams for DeviceIntelIommu {
...
+            MonitorProtocol::QmpPretty => "qmp-pretty".to_string(),
+            _ => "qmp".to_string(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct QmpSocket {
+    // type is the socket type (e.g. "unix").
+    pub socket_type: String,
```
gkurz, line 1227: We're only going to use `"unix"`. Please hard-code that.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1178,6 +1181,82 @@ impl ToQemuParams for DeviceIntelIommu {
...
+    pub protocol: MonitorProtocol,
+    // QMP listener file descriptor to be passed to qemu
+    pub fd: Option<File>,
+    // name is the socket name.
+    pub name: String,
+    // server tells if this is a server socket.
+    pub server: bool,
+    // nowait tells if qemu should block waiting for a client to connect.
+    pub nowait: bool,
+}
```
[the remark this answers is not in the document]
pmores: This is of course a matter of taste but qemu-rs follows a convention where hard-coding is limited to constructors and the rest of the objects' implementation is written generically.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
pmores: I'm in favour of leaving out `validate()` entirely as I explain in another comment.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1178,6 +1181,82 @@ impl ToQemuParams for DeviceIntelIommu {
...
+#[derive(Debug)]
+pub struct QmpSocket {
+    // type is the socket type (e.g. "unix").
+    pub socket_type: String,
+    // protocol to be used on the socket.
+    pub protocol: MonitorProtocol,
+    // QMP listener file descriptor to be passed to qemu
+    pub fd: Option<File>,
+    // name is the socket name.
+    pub name: String,
```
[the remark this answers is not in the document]
pmores: Yes, I don't think we'll ever use both of these options, will we?  We should leave out the one that we don't use.  The rationale is again, unlike (the original) govmm our ambition is not a generic qemu support that can do whatever qemu can do - for the sake of simplicity, we only support what's actually useful for runtime-rs.

pmores:
@Apokleos Have you considered using an existing crate to take care of the low-level details of QMP?  Maybe something like [qapi-qmp](https://crates.io/crates/qapi-qmp) could help us focus on what's important to us?

Apokleos:
> @Apokleos Have you considered using an existing crate to take care of the low-level details of QMP? Maybe something like [qapi-qmp](https://crates.io/crates/qapi-qmp) could help us focus on what's important to us?
>
Yes, you're right ! it seems fine to be the basis of QMP implementation for us and we don't need to reinvent the wheel with it.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
Apokleos: agree

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
Apokleos: sure!

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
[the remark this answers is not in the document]
gkurz: > IMO, we can just use the `unix:fd=xxx..` and no need to keep the `name` for `unix:path=xx...`. And I prepare to remove the field `name` in the next version.

Hold on. It is true that the primary QMP control point will only be `unix:fd=XXX` but the go runtime also offers the `extra_monitor_socket` configuration that allows to add an extra QMP/HMP control point for debugging purpose. The latter can only be `unix:path=xx...`.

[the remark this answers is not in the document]
pmores: Right, that's a good point.  We need to decide now whether the base and extra sockets share enough configuration to justify both being represented by the same struct - otherwise the extra socket might be better represented by a separate struct.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1212,6 +1320,8 @@ pub struct QemuCmdLine<'a> {
     smp: Smp,
     machine: Machine,
     cpu: Cpu,
+    rtc: Rtc,
```
BbolroC, line 1323: I was wondering why this item is added here even if it is add to `devices` below.

gkurz: This is certainly a typo because `rtc` was just removed from there by #9571 :wink:

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1228,6 +1296,9 @@ impl<'a> QemuCmdLine<'a> {
             smp: Smp::new(config),
             machine: Machine::new(config),
             cpu: Cpu::new(config),
+            rtc: Rtc::new(),
```
gkurz, line 1341: ditto

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
Apokleos: sorry, I will fix it.

Apokleos:
> @Apokleos , as explained in [#9604 (comment)](https://github.com/kata-containers/kata-containers/pull/9604#discussion_r1597958093) , the extra monitor socket **must** be created with `unix:path=`. The reason is that this extra control socket is to be used directly by the user with an external tool, e.g. `nc`. The shim won't need to pre-connect to the extra socket like it does with the primary one.
>
Ok, Thx Greg!

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1232,19 +1238,66 @@ pub struct QmpSocket {
 }
 
 impl QmpSocket {
-    fn new() -> Result<Vec<Self>> {
-        let qmp_sockets: Vec<QmpSocket> = Vec::new();
-        let _proto = MonitorProtocol::new("qmp");
+    fn new(sid: &str, extra_monitor: bool) -> Result<Vec<Self>> {
```
pmores, line 1241: I guess this works, however we now have a constructor that returns a container of instances instead of an instance - which is kind of unexpected.  :-)  Honestly I'd prefer if the `QmpSocket::new()` took the monitor protocol as a parameter and then returned a single instance as usual.

Then `QmpSocket::fd` and `QmpSocket::name` had probably better become a single member of an enum type since only one of them is ever needed.

Also, `QemuCmdLine::new()` would unconditionally create a `QmpSocket` instance for the Qmp socket and optionally, based on hypervisor config, another instance for Hmp, similar to how MMIO and RTC are already handled.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1299,7 +1352,7 @@ impl<'a> QemuCmdLine<'a> {
             smp: Smp::new(config),
             machine: Machine::new(config),
             cpu: Cpu::new(config),
-            qmp_sockets: QmpSocket::new()?,
+            qmp_sockets: QmpSocket::new(id, false)?,
```
pmores, line 1355: I believe that the Qmp sockets should go to the anonymous `devices` container since QemuCmdLine doesn't need to interact with them anymore once they are created (@gkurz ?) so there's no need to hold them by name for future reference.

pmores:
@Apokleos I proposed a bunch of structural changes - if you're fine with them but perhaps lack time to rework the PR to implement them we can merge as-is and I can make the changes subsequently.  What do you think?

pmores (approved):
Looks good to me, thanks @Apokleos !

(As a nit, I suspect the last two commits could be squashed however I leave that to your discretion.)

src/libs/kata-types/src/config/hypervisor/mod.rs
```diff
@@ -411,6 +411,19 @@ pub struct DebugInfo {
...
+    /// WARNING: Anyone with access to the monitor socket can take full control of
+    /// Qemu. This is for debugging purpose only and must *NEVER* be used in
+    /// production.
+    /// Valid values are :
+    /// - "hmp"
+    /// - "qmp"
+    /// - "qmp-pretty" (same as "qmp" with pretty json formatting)
+    /// If set to the empty string "", no debug monitor socket is added. This is
+    /// the default.
+    /// dbg_monitor_socket = hmp
```
gkurz, line 425:
```suggestion
    /// dbg_monitor_socket = "hmp"
```

Original config in go runtime also has this nit. No big deal.

Apokleos: Thx Greg @gkurz fixed

PR #9605, opened 2024-05-07
kata-agent: update env PCIDEVICE_<prefix>_<resource-name>_INFO

l8huang:
The new version of sriov-network-device-plugin adds an env `PCIDEVICE_<prefix>_<resource-name>_INFO`, which has a json value; kata-agent can't parse it as env `PCIDEVICE_<prefix>_<resource-name>` which has value in format "DDDD:BB:SS.F".

This change updates env `PCIDEVICE_<prefix>_<resource-name>_INFO`.

### DIFF
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -710,7 +710,14 @@ pub fn update_env_pci(
     env: &mut [String],
     pcimap: &HashMap<pci::Address, pci::Address>,
 ) -> Result<()> {
-    for envvar in env {
+    // SR-IOV device plugin may add two environment variables for one resource:
+    // - PCIDEVICE_<prefix>_<resource-name>: a list of PCI device ids separated by comma
+    // - PCIDEVICE_<prefix>_<resource-name>_INFO: detailed info in JSON for above PCI devices
+    // Both environment variables hold information about the same set of PCI devices.
+    // Below code updates both of them in two passes:
+    // - 1st pass updates PCIDEVICE_<prefix>_<resource-name> and collects host to guest PCI address mapping
+    let mut pci_dev_map: HashMap<String, HashMap<String, String>> = HashMap::new();
+    for envvar in env.iter_mut() {
         let eqpos = envvar
             .find('=')
             .ok_or_else(|| anyhow!("Malformed OCI env entry {:?}", envvar))?;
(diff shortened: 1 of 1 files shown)


src/agent/src/device.rs
```diff
@@ -718,7 +718,7 @@ pub fn update_env_pci(
         let (name, eqval) = envvar.split_at(eqpos);
         let val = &eqval[1..];
 
-        if !name.starts_with("PCIDEVICE_") {
+        if !name.starts_with("PCIDEVICE_") || name.ends_with("_INFO") {
```
lifupan, line 721: Does  this env still needed to be parsed? If  yes, can we parse it correctly instead of skipping it?

zvonkok: The sriov-device-pluign needs this env variable to make the correct connection of allocated extended resource to the actual BDF that this extended resource is belonging to. As lifupan said, parse it correctly rather than skipping it.

l8huang: In our use case, the sriov-network-device-plugin only adds `PCIDEVICE_<prefix>_<resource-name>_INFO` like below:
```
PCIDEVICE_NVIDIA_COM_BF2_INFO={\"0000:84:01.6\":{\"generic\":{\"deviceID\":\"0000:84:01.6\"}}} 
```
without takes additional action on that.

I can add code to process it instead of skipping it, just want to clarify, in above case, do I only need to update the key, or both of the key and the value of the `.generic.deviceID`? -- I understand the `generic` is provider name, it can be anything else.

l8huang: @lifupan @zvonkok if it's not clear about how to updating `PCIDEVICE_<prefix>_<resource-name>_INFO` right now, how about merge this PR firstly then come back to this issue later?

l8huang: kindly ping @lifupan @zvonkok

lifupan: > @lifupan @zvonkok if it's not clear about how to updating `PCIDEVICE_<prefix>_<resource-name>_INFO` right now, how about merge this PR firstly then come back to this issue later?

Maybe you can parse the value with finding the host BFD in the value string and  replacing the host BDF with the guest  BDF, and then save it into the env, thus the container can get the right BDF info from the env.

l8huang: `PCIDEVICE_<prefix>_<resource-name>` is a list of host BFD separated by comma, it easy to parse and replace them one by one; but `PCIDEVICE_<prefix>_<resource-name>_INFO` is formatted as JSON, parsing it directly as plain text to identify all host BFDs and replace them with guest BFD is challenging.

If `PCIDEVICE_<prefix>_<resource-name>_INFO` presents,I will utilize the existing code to perform a substring replacement, changing 'host BFD' to 'guest BFD' in the value of PCIDEVICE_INFO.
```
        for host_addr in val.split(',') {
            let host_addr = pci::Address::from_str(host_addr)
                .with_context(|| format!("Can't parse {} environment variable", name))?;
            let guest_addr = pcimap
                .get(&host_addr)
                .ok_or_else(|| anyhow!("Unable to translate host PCI address {}", host_addr))?;
 
            // if `PCIDEVICE_<prefix>_<resource-name>_INFO` presents, do host BFD -> guest BFD replacement
            guest_addrs.push(format!("{}", guest_addr));
        }

```

lifupan: Hi @l8huang  May be you can try to use regular expressions to match and extract the host BDF?

l8huang: Hi @lifupan, I pushed a change as I mentioned in my previous comment, PTAL.

lifupan (approved):
LGTM

zvonkok (approved):
LGTM, thanks!

PR #9608, opened 2024-05-08
tdx: Adapt kata-deploy to use QEMU / OVMF from the distros

fidencio:
This is the second part of the work related to using the host stack from the distros that support TDX, instead of building our own.

The changes proposed here are tied to the kata-deploy script, so they can go in without any issues as long as the required CIs are passing.

Once we do changes in the runtime itself, those will require that the TDX CI passes before we get them merged.

PR #9609, opened 2024-05-08
tests: pull-image: Don't run on TEEs

fidencio:
Let's skip those tests on TEEs as we've been facing a reasonable amount of issues, most likely on the containerd side, related to pulling the image on the guest.

Once we're able to fix the issues on containerd, we can get back and re-enable those by reverting this commit.

The decision of disabling the tests for TEEs is because the machines may end up in a state where human intervention is necessary to get them back to a functional state, and that's really not optimal for our CI.

### DIFF
--- a/tests/integration/kubernetes/k8s-guest-pull-image.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image.bats
@@ -9,7 +9,8 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
-    confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
+    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+
     [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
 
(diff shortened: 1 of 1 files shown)


fidencio:
TDX is supposed to fail, as we require https://github.com/kata-containers/kata-containers/pull/9608 to be merged before the CI gets fully back.

GabyCT (approved):

danmihai1 (approved):

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -9,7 +9,8 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
-    confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
+    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
```
stevenhorsman, line 12: This skips the non-TEE tests as well, which I don't think we want to do?

fidencio: Sorry, that was missed from my side, as I didn't realise that the `confidential_setup` check would actually check for non-TEEs as well. :-/

I will raise a follow-up.

PR #9610, opened 2024-05-08
metrics: Fix random write value for FIO

GabyCT:
This PR fixes the random write value for FIO for qemu and clh by decreasing it to avoid the random failures of the GHA CI.


dborquez (approved):
LGTM, thank you @GabyCT

cmaf (approved):
thanks @GabyCT

PR #9613, opened 2024-05-09
tests: pull-image: Only skip tests for TEEs

fidencio:
On 142342012c01debcd8b4c8812e6cc2b62daf698e, I've mistakenly disabled the tests entirely, for both non-TEEs and TEEs.

This happened as I didn't realise that `confidential_setup` would take non-TEEs into consideration. :-/

Now, let me follow-up on that and make sure that the tests will be running on non-TEEs.

### DIFF
--- a/tests/integration/kubernetes/confidential_common.sh
+++ b/tests/integration/kubernetes/confidential_common.sh
@@ -52,10 +52,12 @@ function check_hypervisor_for_confidential_tests() {
     fi
 }
 
-# Common setup for confidential tests.
-function confidential_setup() {
+# Common check for confidential tests.
+function is_confidential_runtime_class() {
 	ensure_yq
-	if ! check_hypervisor_for_confidential_tests "${KATA_HYPERVISOR}"; then
-        return 1
-    fi
+	if check_hypervisor_for_confidential_tests "${KATA_HYPERVISOR}"; then
+		return 0
+	fi
+
+	return 1
 }
--- a/tests/integration/kubernetes/confidential_common.sh
+++ b/tests/integration/kubernetes/confidential_common.sh
@@ -67,7 +67,6 @@ function check_hypervisor_for_confidential_tests_tee_only() {
 
 # Common check for confidential tests.
 function is_confidential_runtime_class() {
-	ensure_yq
 	if check_hypervisor_for_confidential_tests "${KATA_HYPERVISOR}"; then
 		return 0
 	fi
--- a/tests/integration/kubernetes/confidential_common.sh
+++ b/tests/integration/kubernetes/confidential_common.sh
@@ -44,7 +44,7 @@ function check_hypervisor_for_confidential_tests() {
 	local kata_hypervisor="${1}"
 	# This check must be done with "<SPACE>${KATA_HYPERVISOR}<SPACE>" to avoid
 	# having substrings, like qemu, being matched with qemu-$something.
-    if [[ " ${SUPPORTED_TEE_HYPERVISORS[*]} " =~ " ${kata_hypervisor} " ]] ||\
+    if check_hypervisor_for_confidential_tests_tee_only "${kata_hypervisor}" ||\
        [[ " ${SUPPORTED_NON_TEE_HYPERVISORS[*]} " =~ " ${kata_hypervisor} " ]]; then        
         return 0
     else
@@ -52,10 +52,32 @@ function check_hypervisor_for_confidential_tests() {
     fi
 }
 
+# This function verifies whether the input hypervisor supports confidential tests and 
+# relies on `KATA_HYPERVISOR` being an environment variable
+function check_hypervisor_for_confidential_tests_tee_only() {
+	local kata_hypervisor="${1}"
+	# This check must be done with "<SPACE>${KATA_HYPERVISOR}<SPACE>" to avoid
+	# having substrings, like qemu, being matched with qemu-$something.
+	if [[ " ${SUPPORTED_TEE_HYPERVISORS[*]} " =~ " ${kata_hypervisor} " ]]; then
+		return 0
+	fi
+
+	return 1
+}
+
 # Common check for confidential tests.
 function is_confidential_runtime_class() {
 	ensure_yq
 	if check_hypervisor_for_confidential_tests "${KATA_HYPERVISOR}"; then
+		return 0
+	fi
+
+	return 1
+}
+
+# Common check for confidential hardware tests.
+function is_confidential_hardware() {
+	if check_hypervisor_for_confidential_tests_tee_only "${KATA_HYPERVISOR}"; then
 		return 0
 	fi
 
--- a/tests/integration/kubernetes/k8s-confidential.bats
+++ b/tests/integration/kubernetes/k8s-confidential.bats
@@ -10,7 +10,9 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 load "${BATS_TEST_DIRNAME}/tests_common.sh"
 
 setup() {
-	confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
+	if ! is_confidential_runtime_class; then
+		skip "Test not supported for ${KATA_HYPERVISOR}."
+	fi
 	setup_unencrypted_confidential_pod
 }
 
@@ -41,7 +43,9 @@ setup() {
 }
 
 teardown() {
-	check_hypervisor_for_confidential_tests ${KATA_HYPERVISOR} || skip "Test not supported for ${KATA_HYPERVISOR}."
+	if ! is_confidential_runtime_class; then
+		skip "Test not supported for ${KATA_HYPERVISOR}."
+	fi
 	
 	kubectl describe "pod/${pod_name}" || true
 	kubectl delete -f "${pod_config_dir}/pod-confidential-unencrypted.yaml" || true
--- a/tests/integration/kubernetes/k8s-confidential.bats
+++ b/tests/integration/kubernetes/k8s-confidential.bats
@@ -10,8 +10,8 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 load "${BATS_TEST_DIRNAME}/tests_common.sh"
 
 setup() {
-	if ! is_confidential_runtime_class; then
-		skip "Test not supported for ${KATA_HYPERVISOR}."
+	if ! is_confidential_hardware; then
+		skip "Test is supported only on confidential hardware (which ${KATA_HYPERVISOR} is not)"
 	fi
 	setup_unencrypted_confidential_pod
 }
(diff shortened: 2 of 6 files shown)


tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -9,7 +9,7 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
-    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    confidential_setup_non_tees || skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
```
stevenhorsman, line 12: This is maybe slightly outside of the scope of this PR, but it feels like the `confidential_setup_non_tees`/ `confidential_setup` logic is pretty confusing with various negation and requiring `||` vs. `&&`. This lead to the original mistake that at least three people overlooked.

This might be controversial, but I wonder if we would have better clarity (and avoid the issue we had here) with some renaming and separation and a bit more code. So this code would look something like:
```
if [ ! is_confidential_runtime_class ] then;
    skip "Test not supported for ${KATA_HYPERVISOR}."
fi
if [ is_confidential_hardware ]; then
     skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
fi
```
It isn't as succinct coding wise, but I'd argue that the best code is the easier to understand rather than the shortest.

We could create a confidential_common function that wraps this and was called something like `skip_for_non_confidential_and_tee_hardware` that would add clarity as well, to avoid the duplication of this code in the bats files?

What do others thinks?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -9,7 +9,9 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
-    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    if [ is_confidential_runtime_class ]; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
```
stevenhorsman, line 13: I don't think this is correct - we need the 
```
if [ ! is_confidential_runtime_class ]; then
		skip "Test not supported for ${KATA_HYPERVISOR}."
	fi
```
here to stop the e.g. clh tests from running?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -20,7 +22,9 @@ setup() {
 }
 
 @test "Test we can pull an unencrypted image outside the guest with runc and then inside the guest successfully" {
-    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    if [ is_confidential_runtime_class ]; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    fi
```
stevenhorsman: Same comment as above this file

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -216,7 +220,9 @@ setup() {
 }
 
 teardown() {
-    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    if [ is_confidential_runtime_class ]; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    fi
```
stevenhorsman: And again

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -52,6 +52,19 @@ function check_hypervisor_for_confidential_tests() {
     fi
 }
 
+# This function verifies whether the input hypervisor supports confidential tests and 
+# relies on `KATA_HYPERVISOR` being an environment variable
+function check_hypervisor_for_confidential_tests_tee_only() {
```
stevenhorsman, line 57: Optional suggestion to refactor `check_hypervisor_for_confidential_tests` to call this method for the tee stuff.

tests/integration/kubernetes/k8s-guest-pull-image.bats
stevenhorsman: Ok, it was fixed in the later commit, so just this commit is wrong

stevenhorsman:
Looking much clearer - Thank you!

tests/integration/kubernetes/k8s-guest-pull-image.bats
fidencio: Yeah, the first commit is just changing the name, thus it'd be wrong (as the current logic is wrong).
The logic itself is fixed in the 3rd commit.

tests/integration/kubernetes/confidential_common.sh
fidencio: Done.

stevenhorsman (approved):
Looking much clearer. Thanks for the updates!

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -44,18 +44,43 @@ function check_hypervisor_for_confidential_tests() {
...
+	if [[ " ${SUPPORTED_TEE_HYPERVISORS[*]} " =~ " ${kata_hypervisor} " ]]; then
+		return 0
+	else
+		return 1
+	fi
+}
+
+# Common check for confidential tests.
+function is_confidential_runtime_class() {
 	ensure_yq
```
stevenhorsman, line 70: I appreciate this wasn't your change, but why do we need yq for `is_confidential_runtime_class` and `is_confidential_hardware`?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -9,7 +9,13 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
-    confidential_setup && skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    if is_confidential_hardware; then
+        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    fi
+
+    if is_confidential_runtime_class; then
```
stevenhorsman, line 16: I think this check is wrong?
```suggestion
    if ! is_confidential_runtime_class; then
```

Same for @test and teardown versions in this file?

stevenhorsman (changes requested):

tests/integration/kubernetes/k8s-guest-pull-image.bats
wainersm: `@test` is an user in github and its pic is a guitar players :D

stevenhorsman: Sorry to disturb you Mr Test!

tests/integration/kubernetes/confidential_common.sh
wainersm: IIRC originally `confidential_setup()` was conceived as a common setup function (not only to skip tests). So in commit 2c0bc8855b we added this line to ensure_yq. The commit doesn't say why and I found yq being only used for yaml instrumentation for policy tests. Besides, `tests/integration/kubernetes/run_kubernetes_tests.sh` also call `ensure_yq` so I suspect we are safe to remove that call from here.

tests/integration/kubernetes/k8s-confidential.bats
```diff
@@ -10,7 +10,9 @@ load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 load "${BATS_TEST_DIRNAME}/tests_common.sh"
 
 setup() {
-	confidential_setup || skip "Test not supported for ${KATA_HYPERVISOR}."
+	if ! is_confidential_runtime_class; then
```
wainersm, line 13: Hi @stevenhorsman @fidencio !

`k8s-confidential.bats` should run only on confidential hardware, so we need something like:

`is_confidential_hardware || skip "Test is supported only on confidential hardware (which ${KATA_HYPERVISOR} is not)"`

fidencio: Updated!

tests/integration/kubernetes/k8s-guest-pull-image.bats
fidencio: Updated!

wainersm (approved):
The code is much better now! Thanks @fidencio @stevenhorsman !

stevenhorsman (approved):
LGTM. Thanks!

stevenhorsman:
We are getting a failure here where the pause image rootfs which we still expect on the host isn't found in the standard place
```
 count of container rootfs in host is: 0, expect count is: 1
```
We have seen this before in https://github.com/kata-containers/tests/issues/5781, but we thought that was due to stale processes on bare-metal machines. 

With the impending move to `shared_fs=none` we shouldn't rely on this any more, so I'm tempted to relax the check to ensure that the rootfs count is no higher than we expect, but I've rebased this to pick up the latest nydus-snapshotter version changes first, and will see if that makes a difference.

fidencio:
> We are getting a failure here where the pause image rootfs which we still expect on the host isn't found in the standard place
> 
> ```
>  count of container rootfs in host is: 0, expect count is: 1
> ```
> 
> We have seen this before in [kata-containers/tests#5781](https://github.com/kata-containers/tests/issues/5781), but we thought that was due to stale processes on bare-metal machines.
> 
> With the impending move to `shared_fs=none` we shouldn't rely on this any more, so I'm tempted to relax the check to ensure that the rootfs count is no higher than we expect, but I've rebased this to pick up the latest nydus-snapshotter version changes first, and will see if that makes a difference.

Let's do this, as this whole test will basically be dropped in the very short future.

stevenhorsman:
> Let's do this, as this whole test will basically be dropped in the very short future.

Already done!

wainersm:
> Executed 0 instead of expected 2 tests

This kind of issue usually means unset variable being used. Let me look at the test's code.

wainersm:
> > Executed 0 instead of expected 2 tests
> 
> This kind of issue usually means unset variable being used. Let me look at the test's code.


Ok, k8s-confidential-attestation.bats  calls `confidential_setup()` that no longer exist. We should adjust all tests that are still using that function.

stevenhorsman:
> > > Executed 0 instead of expected 2 tests
> > 
> > 
> > This kind of issue usually means unset variable being used. Let me look at the test's code.
> 
> Ok, k8s-confidential-attestation.bats calls `confidential_setup()` that no longer exist. We should adjust all tests that are still using that function.

Thanks, I'll work on that tomorrow morning.

PR #9618, opened 2024-05-10
gpu: Add build targets for GPU rootfs initrd/image

zvonkok:
Preparation for complete GPU rootfs build step num 1 of N

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -414,6 +414,39 @@ install_initrd_mariner() {
 	install_initrd "mariner"
 }
 
+#Instal NVIDIA GPU image
+install_image_nvidia_gpu() {
+	export AGENT_POLICY="yes"
+	export AGENT_INIT="yes"
+	export EXTRA_PKGS="apt udev"
+	install_image "nvidia-gpu"
+}
+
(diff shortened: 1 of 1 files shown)


tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -414,6 +414,39 @@ install_initrd_mariner() {
 	install_initrd "mariner"
 }
 
+#Instal NVIDIA GPU image
+install_image_nvidia_gpu() {
+	export AGENT_POLICY="yes"
+	export AGENT_INIT="yes"
```
stevenhorsman, line 420: @fidencio and/or @danmihai1 can probably confirm, but with the regorus crate and follow on changes I'm not sure if `AGENT_POLICY="yes"` (as it defaults to `yes` now) and `AGENT_INIT="yes"` are strictly required to use policy, but equally I don't think it's and issue to have them set as a double measure.

zvonkok: It is more to amplify and show that be it image or initrd we're always using AGENT_INIT and AGENT_POLICY I think it does not "hurt".

stevenhorsman (approved):
Looks okay to me.

beraldoleal (approved):
lgtm, thanks

PR #9622, opened 2024-05-11
build: nvidia-gpu: Fix cache usage of the headers tarball

fidencio:
Whenever we count on having the headers tarball, we must unpack the cached content into the expected directory, otherwise we'd simply fail, as we've been failing in our CI, at the end of the process where we generate the tarball from the cached components.

It's weird to me, sincerely, that the headers tarball end up in such weird place (build/kernel-nvidia-gpu/builddir/), but I'll leave that to Zvonko to figure out whether something better can be done, as the intuit of this PR is simply unblock Kata Containers CI.


zvonkok:
At the kernel build time we have no rootfs. So we need unpack it somewhere. Usually the tar archive is unpacked and into the rootfs. So this fix is just to make the CI happy.

zvonkok (approved):

ananos (approved):
LGTM! Let’s unblock the CI!

PR #9633, opened 2024-05-14
ci: rootless ci cd, sudo less github-runner

zvonkok:
We need updates on the github actions as well. We need/can eliminate most of the `sudo` commands. 

The main issue is that all targets are built with `sudo` privileges aka `root` and `builddir` and `destdir` may have a mix of `$USER` and `root` owned files. That's why we have `sudo` sprinkled all around. 

Additionally the `kata-static-$(target-build).xz` are also owned by `root` and that's why we're doing `sudo oras`. 

With these patches all build files in `builddir` and `destdir` are owned by the `$USER` and not root anymore. 

I have a github-runner without `sudo` privileges running.  Here is an example of a simplified run: https://github.com/zvonkok/kata-containers/actions/runs/9082982416/job/24961219629?pr=14


tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1148,7 +1148,9 @@ silent_mode_error_trap() {
 }
 
 main() {
-	git config --global --add safe.directory ${repo_root_dir}
+	#git config --global --add safe.directory ${repo_root_dir}
```
fidencio, line 1151: leftover? :-)

tools/packaging/kata-deploy/local-build/kata-deploy-build-and-upload-payload.sh
```diff
@@ -25,7 +25,7 @@ arch=$(uname -m)
 [ "$arch" = "x86_64" ] && arch="amd64"
 IMAGE_TAG="${REGISTRY}:kata-containers-$(git rev-parse HEAD)-${arch}"
 
-sudo chown -R $USER $HOME/.docker
+#chown -R $USER $HOME/.docker
```
fidencio, line 28: leftover?

Issue #8407, opened 2023-11-09
CC | Issues about pulling image with vanilla containerd and remote-snapshotter


ChengyuZhu6:
This issue serves as a place to collect and discuss the current challenges of using vanilla containerd and remote-snapshotter in confidential-containers. If you have any related questions or are interested in this topic, please feel free to share them in this issue.

/cc @jiangliu  @fidencio @stevenhorsman @fitzthum @BbolroC @huoqifeng

ChengyuZhu6:
# Issue 1

## Description of problem
When we use the default snapshotter (overlayfs) to pull an image (such as a pause image) and create a container, and then we switch to another snapshotter (nydus-snapshotter) to pull the same image, we encounter an error:: 
```
Events:
  Type     Reason                  Age   From               Message
  ----     ------                  ----  ----               -------
  Normal   Scheduled               1s    default-scheduler  Successfully assigned default/coco-kata-1 to zcyubuntu22
  Warning  FailedCreatePodSandBox  1s    kubelet            Failed to create pod sandbox: rpc error: code = NotFound desc = failed to start sandbox "10e7b3ad6400414b393eb8b7e7dc84cfbdd2875bad6733b0db8081a8b99d61ae": failed to create containerd container: error unpacking image: failed to extract layer sha256:1021ef88c7974bfff89c5a0ec4fd3160daac6c48a075f74cff721f85dd104e68: failed to get reader from content store: content digest sha256:fbe1a72f5dcd08ba4ca3ce3468c742786c1f6578c1f6bb401be1c4620d6ff705: not found
```

and I checked the image (I replaced registry.cn-hangzhou.aliyuncs.com/google_containers/pause:3.6 with registry.k8s.io/pause:3.6 because of the network issue) : 
```
$ ctr -n k8s.io image check| grep pause
registry.cn-hangzhou.aliyuncs.com/google_containers/pause:3.6                                                                           application/vnd.docker.distribution.manifest.list.v2+json sha256:3d380ca8864549e74af4b29c10f9cb0956236dfb01c40ca076fb6c37253234db incomplete (1/2) 901.0 B/290.4 KiB   true
[...]

PR #9636, opened 2024-05-15
version: Bump nydus snapshotter to v0.13.13

ChengyuZhu6:
Bump nydus snapshotter to v0.13.13 to fix the gap when switching different snapshotters in guest pull.

Fixes: #8407

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -317,7 +317,7 @@ externals:
   nydus-snapshotter:
     description: "Snapshotter for Nydus image acceleration service"
     url: "https://github.com/containerd/nydus-snapshotter"
-    version: "v0.13.11"
+    version: "v0.13.13"
 
   ovmf:
     description: "Firmware, implementation of UEFI for virtual machines."


stevenhorsman (approved):
LGTM. Thanks!

GabyCT (approved):

wainersm:
Hi all!

Not related exactly with this PR but saying here because I got your audience ;)

Looking at the non-tee job to see if this change deployed the v0.13.13 I realized that it's actually fetching the `latest` image (`ghcr.io/containerd/nydus-snapshotter:latest`):

```
 Normal  Scheduled  5s    default-scheduler  Successfully assigned nydus-system/nydus-snapshotter-j2vmp to aks-nodepool1-98563127-vmss000000
    Normal  Pulling    5s    kubelet            Pulling image "ghcr.io/containerd/nydus-snapshotter:latest"
    Normal  Pulled     1s    kubelet            Successfully pulled image "ghcr.io/containerd/nydus-snapshotter:latest" in 3.55s (3.55s including waiting)
    Normal  Created    1s    kubelet            Created container nydus-snapshotter
    Normal  Started    1s    kubelet            Started container nydus-snapshotter
```

Looking at https://github.com/containerd/nydus-snapshotter/blob/v0.13.13/misc/snapshotter/base/nydus-snapshotter.yaml#L42 .... yeah... using latest.

You know I'm a big fan of pinning versions, so I think we should pin this properly as well. I see two options here:

 1. On our CI, change the image tag before apply. Somewhere here: https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/gha-run.sh#L472
 2. Ask nydus-snapshotter community to pin the version themselves at release time (IMHO they should). In our side we implement some safe-guard, e.g., check the deployed image is the version expected

stevenhorsman:
> Looking at the non-tee job to see if this change deployed the v0.13.13 I realized that it's actually fetching the `latest` image (`ghcr.io/containerd/nydus-snapshotter:latest`):

Yeah - @mkulke and I were chatting about this on slack earlier and thought that we'd either user kustomize to update the image, or ideal request the nydus-snapshotter community to pin the version, which we felt was the better solution (and then we follow that by pinning our operator image too!). So to repeat your words - we are also in agreement that the best way is to ask nydus-snapshotter for the update as part of their release process I guess?

fidencio:
@stevenhorsman, @wainersm, I fully agree, and I'm assuming we will be able to work on the items as follow-ups.
For noiw, merging this one.

ChengyuZhu6:
>Yeah - @mkulke and I were chatting about this on slack earlier and thought that we'd either user kustomize to update the image, or ideal request the nydus-snapshotter community to pin the version, which we felt was the better solution (and then we follow that by pinning our operator image too!). So to repeat your words - we are also in agreement that the best way is to ask nydus-snapshotter for the update as part of their release process I guess?

I have created a pr to fix it. https://github.com/containerd/nydus-snapshotter/pull/596

PR #9638, opened 2024-05-15
CI: Use `--abbrev=9` explicitly for abbreviated commit hash

BbolroC:
A length of the result of `git log -1 --pretty=format:%h` could vary over different CI systems, highly likely messing up their caching mechanisms.

This commit is to use an option `--abbrev=9` to standardize the length to 9 characters for CI.

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -186,7 +186,7 @@ install_cached_tarball_component() {
 	fi
 
 	local component="${1}"
-	local current_version="${2}-$(git log -1 --pretty=format:"%h" ${repo_root_dir}/tools/packaging/kata-deploy/local-build)"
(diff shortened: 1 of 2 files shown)


src/runtime/vendor/go.opentelemetry.io/otel/verify_examples.sh
```diff
@@ -36,7 +36,7 @@ if ! git diff --quiet; then \
 	exit -1
 fi
 
-if [ "$(git tag --contains $(git log -1 --pretty=format:"%H"))" = "" ] ; then
```
stevenhorsman, line 39: I don't think we should update vendor code

stevenhorsman (changes requested):

src/runtime/vendor/go.opentelemetry.io/otel/verify_examples.sh
BbolroC: ah... my bad. I will update the PR.

stevenhorsman (approved):
LGTM. Thanks!

beraldoleal (approved):
LGTM, thanks

sprt (approved):

PR #9639, opened 2024-05-15
runtime-rs: Add RNG to QEMU cmdline

emanuellima1:
It creates this line, as the Golang runtime does:
-object rng-random,id=rng0,filename=/dev/urandom -device virtio-rng-pci,rng=rng0,max-bytes=1024,period=1000

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
@@ -1149,6 +1149,63 @@ impl ToQemuParams for Rtc {
 }
 
 #[derive(Debug)]
+struct ObjectRngRandom {
+    // id is the device ID
+    id: String,
+
+    // filename is the entropy source on the host
+    filename: String,
+}
+
+impl ObjectRngRandom {
+    fn new() -> ObjectRngRandom {
+        ObjectRngRandom {
+            id: "rng0".to_owned(),
+            filename: "/dev/urandom".to_owned(),
+        }
+    }
+}
+
+#[async_trait]
+impl ToQemuParams for ObjectRngRandom {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        let mut object_params = Vec::new();
+
+        object_params.push("rng-random".to_owned());
+        object_params.push(format!("id={}", self.id));
+        object_params.push(format!("filename={}", self.filename));
+
+        Ok(vec!["-object".to_owned(), object_params.join(",")])
+    }
+}
+
+#[derive(Debug)]
+struct DeviceRng {
+    // transport is the virtio transport for this device.
+    transport: String,
+}
+
+impl DeviceRng {
+    fn new() -> DeviceRng {
+        DeviceRng {
+            transport: "virtio-rng-pci".to_owned(),
+        }
+    }
+}
+
+#[async_trait]
+impl ToQemuParams for DeviceRng {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        let mut device_params = Vec::new();
+
+        device_params.push(self.transport.clone());
+        device_params.push(format!("rng={}", "rng0".to_owned()));
+
+        Ok(vec!["-device".to_owned(), device_params.join(",")])
+    }
+}
+
+#[derive(Debug)]
 struct DeviceIntelIommu {
     intremap: bool,
     device_iotlb: bool,
@@ -1238,12 +1295,24 @@ impl<'a> QemuCmdLine<'a> {
 
         qemu_cmd_line.add_rtc();
 
+        if qemu_cmd_line.bus_type() != VirtioBusType::Ccw {
+            qemu_cmd_line.add_rng();
+        }
+
         Ok(qemu_cmd_line)
     }
 
     fn add_rtc(&mut self) {
         let rtc = Rtc::new();
         self.devices.push(Box::new(rtc));
+    }
+
+    fn add_rng(&mut self) {
+        let rng_object = ObjectRngRandom::new();
+        let rng_device = DeviceRng::new();
+
+        self.devices.push(Box::new(rng_object));
+        self.devices.push(Box::new(rng_device));
     }
 
     fn bus_type(&self) -> VirtioBusType {


emanuellima1:
@pmores @gkurz @beraldoleal please review this PR

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1148,6 +1148,61 @@ impl ToQemuParams for Rtc {
...
+    transport: String,
+}
+
+impl RngDevice {
+    fn new() -> RngDevice {
+        RngDevice {
+            id: "rng0".to_owned(),
+            filename: "/dev/urandom".to_owned(),
+            max_bytes: 1024,
+            period: 1000,
```
gkurz, line 1175: If you look carefully at the go runtime, you'll see that `virtcontainers` doesn't set values for `MaxBytes` and `Period`. This causes `govmm` to ignore these settings and to generate a simple `-device virtio-rng-pci,rng=rng0`.

At this stage, we don't have any reason to take control on the default configuration of the device. Please drop `max_bytes` and `period` for now. We can still introduce them when we have a actual case for it.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1148,6 +1148,61 @@ impl ToQemuParams for Rtc {
...
+            transport: "virtio-rng-pci".to_owned(),
```
gkurz, line 1176: Should this be `ccw` for s390x ? @BbolroC ?

beraldoleal (commented):
Hi @emanuellima1, just a small note for the future PRs:

It's best to avoid creating new PRs to replace old ones, as we might "lose" some comments and suggestions from previous reviewers. Pushing to the same PR helps track the history and progress of changes more effectively and centralizes discussions. 

If a new PR is *really* necessary, it's helpful to tag the original reviewers to maintain continuity and ensure everyone's input was considered. 

Other than @gkurz's comments, looks good to me, thanks.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
BbolroC: Yes, please. Thanks for the notice. 😉

BbolroC: But, I've noticed that it is not instantiated for s390x. https://github.com/kata-containers/kata-containers/blob/64b915b86e5be1a2becd950644ec748123e50b1e/src/runtime/virtcontainers/qemu.go#L690-L701

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1148,6 +1148,61 @@ impl ToQemuParams for Rtc {
...
+        }
+    }
+}
+
+#[async_trait]
+impl ToQemuParams for RngDevice {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        let mut object_params = Vec::new();
+        let mut device_params = Vec::new();
```
pmores, line 1185: qemu-rs uses a struct per command line switch for code readability.  `RngDevice` should be split into a `DeviceRng` part to generate `-device virtio-rng...` and an `ObjectRngRandom` to generate the `-object rng-random` part.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
pmores: I think it should use `VirtioBusType` like other virtio devices.

pmores: @BbolroC Ah okay, good catch, thanks!  @emanuellima1 So this should be reflected when the RNG is added (ie. only add it if not on a CCW system).

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1238,6 +1295,11 @@ impl<'a> QemuCmdLine<'a> {
 
         qemu_cmd_line.add_rtc();
 
+        if !config.machine_info.machine_type.contains("-ccw-") {
```
pmores, line 1298: Please use the existing `bus_type()` function to avoid code duplication.

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1246,6 +1308,16 @@ impl<'a> QemuCmdLine<'a> {
...
 
+    fn add_rng_object(&mut self) {
+        let rng_object = ObjectRngRandom::new();
+        self.devices.push(Box::new(rng_object));
+    }
+
+    fn add_rng_device(&mut self) {
+        let rng_device = DeviceRng::new();
+        self.devices.push(Box::new(rng_device));
+    }
```
pmores, line 1319: This should be a single `add_rng()` function that instantiates and stores all required objects (see other `QemuCmdLine::add_*()` functions for an example).

src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
```diff
@@ -1148,6 +1148,63 @@ impl ToQemuParams for Rtc {
...
+}
+
+#[async_trait]
+impl ToQemuParams for ObjectRngRandom {
+    async fn qemu_params(&self) -> Result<Vec<String>> {
+        let mut object_params = Vec::new();
+
+        object_params.push("rng-random".to_owned());
+        object_params.push(format!("id={}", self.id));
+        object_params.push(format!("filename={} ", self.filename));
```
gkurz, line 1176:
```suggestion
        object_params.push(format!("filename={}", self.filename));
```

gkurz: This should address https://github.com/kata-containers/kata-containers/pull/9639#issuecomment-2117562446.

gkurz (approved):
LGTM. Thanks @emanuellima1 !

pmores (approved):
Looks good to me, thanks @emanuellima1 !

PR #9642, opened 2024-05-15
runtime-rs: Drop some useless QEMU arguments

gkurz:
All these settings are hardcoded as `false` and result in no extra options on the QEMU command line, like the go runtime does. There actually not needed :
- we're never going to ask QEMU to survive a guest shutdown
- we're never going to run QEMU daemonized since it prevents log collection
- we're never going to ask QEMU to start with the guest stopped

No need to keep this code around then.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
@@ -363,10 +363,6 @@ struct Knobs {
     nodefaults: bool,
     nographic: bool,
     no_reboot: bool,
-    no_shutdown: bool,
-    daemonize: bool,
-    stopped: bool,
-
     vga: String,
 }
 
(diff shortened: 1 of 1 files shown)


fidencio:
> * we're never going to ask QEMU to start with the guest stopped

This is a case that I'm not exactly sure whether it's true or not.
How does VM Caching / VM Templating work? That's a sincere question, as I never ever used it.

c3d:
> > * we're never going to ask QEMU to start with the guest stopped
> 
> This is a case that I'm not exactly sure whether it's true or not.

Likewise. After a quick scan of the code, I believe that @gkurz is correct that the govmm knob `Stopped` is never set, so from that point of view, this is safe. However, I also know that original AMD SEV required VM to be started in stopped mode. Same thing if we ever want to support migration.

So probably irrelevant today, but might be needed someday.

> How does VM Caching / VM Templating work? That's a sincere question, as I never ever used it.

Templating does not appear to be using the `Stopped` knob, nor the `appendIncoming` method (related to migration).

lifupan (approved):

Apokleos:
Tend to remove/drop useless arguments when we develop qemu-rs. Actually, we are doing so, for example, add network and qmp and others support in cmdline. And there're indeed some useless arguments still existing runtime-go which are never used.

amshinde (approved):

PR #9643, opened 2024-05-15
gh: reorg on ci files

beraldoleal:
Lets move specific architecture workflows to their own folders for better organization.

If this is well-received, more changes could follow.


beraldoleal:
@fidencio @wainersm @gkurz let me know if this type of reorg is welcome and makes sense.

c3d (commented):
For what it's worth, I like it @beraldoleal

.github/workflows/amd64/basic-ci.yaml
gkurz, line 1: Hi @beraldoleal !

First, I definitely like the idea of structuring these files a bit. It seems right to move all the files with an `${arch}` suffix to an `${arch}` subdir...

.github/workflows/amd64/run-k8s-tests-on-aks.yaml
gkurz, line 1: This one doesn't have an architecture suffix but it requires the amd64 static tarball :
```
.github/workflows/run-k8s-tests-on-aks.yaml:          name: kata-static-tarball-amd64${{ inputs.tarball-suffix }}
```
so maybe it is acceptable to move it under the `amd64`.

.github/workflows/amd64/run-k8s-tests-on-garm.yaml
gkurz, line 1: This one doesn't seem to be arch specific. Why moving it under amd64 ?

gkurz: Same question for all the other ones without an architecture suffix actually.

beraldoleal: I moved because is depending on amd64 binaries.

.github/workflows/ci.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
 
   run-kata-deploy-tests-on-garm:
     needs: publish-kata-deploy-payload-amd64
-    uses: ./.github/workflows/run-kata-deploy-tests-on-garm.yaml
+    uses: ./.github/workflows/amd64/run-kata-deploy-tests-on-garm.yaml
```
beraldoleal, line 129: @gkurz here, see line 128. needs amd64 payload.

wainersm:
Hi @beraldoleal ! Nice idea but... :(

... last time I tried to organize action files in directories I realized github expect all them flat in `./.github/workflows`. I even found an issue with people suggesting allowing to have hierarchies of directories. Have it changed?

wainersm:
> Hi @beraldoleal ! Nice idea but... :(
> 
> ... last time I tried to organize action files in directories I realized github expect all them flat in `./.github/workflows`. I even found an issue with people suggesting allowing to have hierarchies of directories. Have it changed?

I just found this: https://github.com/orgs/community/discussions/18055

beraldoleal:
> > Hi @beraldoleal ! Nice idea but... :(
> > ... last time I tried to organize action files in directories I realized github expect all them flat in `./.github/workflows`. I even found an issue with people suggesting allowing to have hierarchies of directories. Have it changed?
> 
> I just found this: [github.com/orgs/community/discussions/18055](https://github.com/orgs/community/discussions/18055)

Yep, just realized that! :( `.github/actions` is reserved for something else and cannot be used either. I had hopes we could use outside `.github/workflows`. We would need to convert that to actions or services. :(

PR #9644, opened 2024-05-16
workflow: Remove if from env conditional

stevenhorsman:
- It appears like the `if` isn't required when setting env as a conditional :( Sorry!

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -96,10 +96,10 @@ jobs:
           ARTEFACT_REGISTRY_USERNAME: ${{ github.actor }}
           ARTEFACT_REGISTRY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
           TARGET_BRANCH: ${{ inputs.target-branch }}
-          RELEASE: ${{ if inputs.stage == 'release' && 'yes' || 'no' }}
+          RELEASE: ${{ inputs.stage == 'release' && 'yes' || 'no' }}
 
       - name: store-artifact ${{ matrix.asset }}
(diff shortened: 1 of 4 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman, and welcome to GHA whac-a-mole session! :-)

stevenhorsman:
As this is fixing a broken workflow on `main` and we run the workflow on `main` in PRs, I'll need to use admin powers to push this PR, but will wait for two approvals first.

.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -96,10 +96,10 @@ jobs:
           ARTEFACT_REGISTRY_USERNAME: ${{ github.actor }}
           ARTEFACT_REGISTRY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
           TARGET_BRANCH: ${{ inputs.target-branch }}
-          RELEASE: ${{ if inputs.stage == 'release' && 'yes' || 'no' }}
+          RELEASE: ${{ inputs.stage == 'release' && 'yes' || 'no' }}
```
wainersm, line 99: Hi @stevenhorsman !

Sorry, I reviewed this and completely overlooked all those mistakes! I should take some vacation...oh wait...

Seriously now, let's enable this https://github.com/marketplace/actions/actionlint action as part of the static-checks to lint the workflow files. Probably some on the mistakes here we would have caught.

wainersm (approved):

.github/workflows/build-kata-static-tarball-amd64.yaml
stevenhorsman: I tried running actionlint locally and (probably due to user error) it didn't detect this issues. I think it's worth a try anyway

wainersm: hmmm Interesting

PR #9647, opened 2024-05-16
ci: cache: Fix unbound variable

stevenhorsman:
Now we have the workflow updated and can test the changes in caching we've hit an error:
```
line 1180: artefact_tag: unbound variable
```
so we need to fix that up. Sorry for missing this before.


cmaf (approved):
thanks @stevenhorsman

GabyCT (approved):

PR #9649, opened 2024-05-16
k8s: Update negative confidential attestation k8s tests

GabyCT:
This PR updates the negative confidential attestation k8s test to look for a message when we are trying to retrieve the logs from the pod.

### DIFF
--- a/tests/integration/kubernetes/k8s-confidential-attestation.bats
+++ b/tests/integration/kubernetes/k8s-confidential-attestation.bats
@@ -76,9 +76,7 @@ setup() {
 	sleep 5
 
 	kubectl logs aa-test-cc
-	cmd="kubectl logs aa-test-cc | grep -q aatest"
-	run $cmd
-	[ "$status" -eq 1 ]
+	kubectl logs aa-test-cc | grep -q "Failed to connect"
 }
 
 teardown() {


tests/integration/kubernetes/k8s-confidential-attestation.bats
```diff
@@ -76,9 +76,7 @@ setup() {
 	sleep 5
 
 	kubectl logs aa-test-cc
-	cmd="kubectl logs aa-test-cc | grep -q aatest"
-	run $cmd
-	[ "$status" -eq 1 ]
+	kubectl logs aa-test-cc | grep -q "Failed to connect"
```
fitzthum, line 79: Do you happen to have the full message that we get from the ASR? This isn't quite what I would have expected.

GabyCT: mmm seems like there is still not message as I am seeing at the GHA that at the logs we still get
```
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   319  100   319    0     0    378      0 --:--:-- --:--:-- --:--:--   379
```

fitzthum: I think you might have to bump the version of guest components in `versions.yaml` to pick up the error handling change.

PR #9650, opened 2024-05-16
Caching tagging update part iii

stevenhorsman:
- Merge oras push logic together
- Container image tags can only contain alphanumeric, period,
hyphen and underscore characters, so convert characters outside
of these to be underscores, to avoid having invalid tag failures

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -1148,6 +1148,8 @@ handle_build() {
 		echo "Pushing ${build_target} with tags: ${tags[*]}"
 
 		for tag in "${tags[@]}"; do
+			# tags can only contain lowercase and uppercase letters, digits, underscores, periods, and hyphens
+			tag=("$(echo ${tag} | tr -c '[a-zA-Z0-9\_\.\-]' _)")
 			case ${build_target} in
 				kernel-nvidia-gpu)
 					sudo oras push \
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -1137,45 +1137,6 @@ handle_build() {
 
 		echo "${ARTEFACT_REGISTRY_PASSWORD}" | sudo oras login "${ARTEFACT_REGISTRY}" -u "${ARTEFACT_REGISTRY_USERNAME}" --password-stdin
 
-		case ${build_target} in
-			kernel-nvidia-gpu)
-				sudo oras push \
-					${ARTEFACT_REGISTRY}/kata-containers/cached-artefacts/${build_target}:latest-${TARGET_BRANCH}-$(uname -m) \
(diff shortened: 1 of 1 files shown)


GabyCT (approved):

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1184,11 +1145,32 @@ handle_build() {
 			tags+=("$(cat "${version_file}")")
 		fi
 
+		echo "Pushing ${build_target} with tags: ${tags[*]}"
+
 		for tag in "${tags[@]}"; do
 			case ${build_target} in
+				kernel-nvidia-gpu)
+					sudo oras push \
+						${ARTEFACT_REGISTRY}/kata-containers/cached-artefacts/${build_target}:latest-${tag} \
```
wainersm, line 1156: Hi @stevenhorsman !

This will tag like `latest-latest-${TARGET_BRANCH}-$(uname -m)`  (prefixed `latest` twice).

Could you give an example of `artefact_tag`?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1148,6 +1148,8 @@ handle_build() {
 		echo "Pushing ${build_target} with tags: ${tags[*]}"
 
 		for tag in "${tags[@]}"; do
+			# tags can only contain lowercase and uppercase letters, digits, underscores, periods, and hyphens
+			tag=("$(echo ${tag} | tr -c '[a-zA-Z0-9\_\.\-]' _)")
```
wainersm, line 1152: Did it come from a real problem? or are you just making it fail-proof?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
stevenhorsman: Ah goid spot. I'll fix that. The example `artefact_tag`s we have are:
- agent - this has the short hash of the kata-containers repo
- pause image - this is the <repo>-<tag> of the pause image. e.g. `docker___registry.k8s.io_pause-3.9` once we replaced the not allow characters
- guest-components - the version.yaml for guest components e.g. `adca2f94091d73c0b0c96a7789322a801c15811b` or `v0.9.0`

Does that help?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
stevenhorsman: Real problem - the pause image is `docker://registry.k8s.io/pause-3.9` at the moment, which isn't valid, but I figured it might occur for versions tags & branch names that are less restrictive than container image tags.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
wainersm: Yes, it does help. Thanks Steve!

wainersm (approved):
Hey @stevenhorsman ! LGTM!

Issue #9497, opened 2024-04-17
TDX: Ensure quote-generation-service is added to the QEMU command line


fidencio:
In order to use QEMU + TDX, and get the quote for the attestation, we need to make sure that QEMU is correctly started using the following command line `-object tdx-guest,id=tdx,quote-generation-service=vsock:2:$qgs_port` (default qgs_port is 4050).

From a quick look at the code, this is something that should be passed to https://github.com/kata-containers/kata-containers/blob/af3b19ed187e4412dee6ed96674714dd4d928422/src/runtime/virtcontainers/qemu_amd64.go#L278-L290, and then https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/govmm/qemu/qemu.go should be adjusted accordingly to be able to build the correct command line for QEMU.

The port, unfortunately, will have to be exposed to the user, and I see this as an user experience problem, as the user deploying Kata Containers may not be the same user who'd set up QGS, but the best we can do is hope for some clear documentation on the CSPs side, to make sure that this information is available.

For this, we'd also need to:
* Add a new `tdx_quote_generation_service_port` option as part of the config file, and this should only be used by the qemu-tdx configuration
* We need to test that passing a `quote-generation-service` won't break QEMU to start when there's no QGS installed in the host, but rather only the attestation would fail

Jakub will be done working on this, so we may need a few more discussions to happen here before actually seeing a patch.

PR #9653, opened 2024-05-17
runtime: Enable connection to Quote Generation Service (QGS)

JakubLedworowski:
For the TD attestation to work the connection to QGS on the host is needed. By default QGS runs on vsock port 4050, but can be modified by the host owner. Format of the qemu object follows the SocketAddress structure, so it needs to be provided in the JSON format, as in the example below: -object '{"qom-type":"tdx-guest","id":"tdx","quote-generation-socket":{"type":"vsock","cid":"2","port":"4050"}}'

Fixes: #9497


fidencio:
I am setting this to "Ready to Review"  as I was able to check that both Ubuntu and CentOS are using exactly the very same command line for QEMU, which is great, and avoids us to do any kind of distro check on the runtime side.

src/runtime/pkg/govmm/qemu/qemu.go
```diff
@@ -15,7 +15,9 @@ package qemu
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"github.com/kata-containers/kata-containers/src/runtime/pkg/device/api"
```
fidencio, line 20: I noticed that you're only using this for logging purposes.
Please, use `log`, which is already imported here, instead of importing a new package.

fidencio (changes requested):
I've left one comment, apart from the comment, everything looks good, thanks @JakubLedworowski!

fidencio (approved):
@JakubLedworowski, very nice work, thanks, and lgtm!

chavafg (approved):

PR #9656, opened 2024-05-17
runtime-rs: document architecture & implementation conventions in qem…

pmores:
As the number of contributors to qemu-rs gradually grows I've been annoying people with review requests to comply with conventions that were never explicitly stated in the first place.  My bad, I apologise. This PR tries to alleviate the problem by describing the conventions explicitly.

Admittedly, it's a wall of text, however the alternative would have been just to state some rules with no rationale which I considered worse.  Hopefully folks can spare some 10 minutes to read the description and avoid wasting time on force-pushes later.

Also note that existing code violates the conventions in several places due to oversight or historical reasons.  It will be fixed shortly.


Apokleos (commented):
Hi Pavel @pmores. I think it's a very good guideline doc about how to devel qemu-rs.
Going a step further, could we categorize the features to be added in the future and provide clearer instructions on where each category should be placed? This explanation could be concise, perhaps requiring only a simple comment to clarify the matter.

gkurz (approved):
LGTM. Thanks @pmores !

GabyCT (approved):

Apokleos (approved):
Thx @pmores LGTM

pmores:
> Hi Pavel @pmores. I think it's a very good guideline doc about how to devel qemu-rs. Going a step further, could we categorize the features to be added in the future and provide clearer instructions on where each category should be placed? This explanation could be concise, perhaps requiring only a simple comment to clarify the matter.

Makes sense though frankly I'm not quite clear on that myself...

Issue #9238, opened 2024-03-06
Annotations handling for remote hypervisor is limited due to relying on qemu verification


bpradipt:
Currently for remote-hypervisor, the verification of hypervisor annotations like default_memory, default_cpu, machine_type etc. is using the Kata/Qemu processing framework. This limits the capabilities of remote hypervisor.

For example when specifying an instance size which is higher than the worker node, the following error is thrown
```
Number of cpus 8 specified in annotation default_vcpus is greater than the number of CPUs 4 on the system: unknown"
```

PR #9657, opened 2024-05-17
runtime: Disable number of cpu comparison on remote hypervisor scenario 

ajaypvictor:
Currently for remote-hypervisor, the verification of hypervisor annotations like default_memory, default_cpu, machine_type etc. is using the Kata/Qemu processing framework. This is limiting the capabilities of remote hypervisor. This PR disables the check for number of cpu comparison on a remote hypervisor scenario.

Fixes https://github.com/kata-containers/kata-containers/issues/9238

### DIFF
--- a/src/runtime/pkg/oci/utils.go
+++ b/src/runtime/pkg/oci/utils.go
@@ -608,7 +608,7 @@ func addHypervisorHotColdPlugVfioOverrides(ocispec specs.Spec, sbConfig *vc.Sand
 func addHypervisorMemoryOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig, runtime RuntimeConfig) error {
 
 	if err := newAnnotationConfiguration(ocispec, vcAnnotations.DefaultMemory).setUintWithCheck(func(memorySz uint64) error {
-		if memorySz < vc.MinHypervisorMemory {
+		if memorySz < vc.MinHypervisorMemory && sbConfig.HypervisorType != vc.RemoteHypervisor {
 			return fmt.Errorf("Memory specified in annotation %s is less than minimum required %d, please specify a larger value", vcAnnotations.DefaultMemory, vc.MinHypervisorMemory)
 		}
 		sbConfig.HypervisorConfig.MemorySize = uint32(memorySz)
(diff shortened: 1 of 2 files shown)


docs/how-to/how-to-set-sandbox-config-kata.md
```diff
@@ -80,6 +82,7 @@ There are several kinds of Kata configurations and they are listed below.
 | `io.katacontainers.config.hypervisor.kernel` | string | the kernel used to boot the container VM |
 | `io.katacontainers.config.hypervisor.machine_accelerators` | string | machine specific accelerators for the hypervisor |
 | `io.katacontainers.config.hypervisor.machine_type` | string | the type of machine being emulated by the hypervisor |
+| `io.katacontainers.config.hypervisor.remote_machine_type` | string | the type of machine to be created by the remote hypervisor |
```
stevenhorsman, line 85: I think we probably should rename this to something like remote_instance_type or similar. Machine type in kata containers means something different, so now we've decoupled I think we should be clearer for the remote hypervisor annotations

stevenhorsman: Also - can you add some line break to you commit message - we need have max line length of 150 chars and more like 80ish is more common. Thanks!

ajayvic: Thanks Steve, will incorporate them..

ajaypvictor: Thanks Steve for the suggestions, will incorporate these..

src/runtime/pkg/oci/utils.go
```diff
@@ -689,7 +689,7 @@ func addHypervisorCPUOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig) e
 	numCPUs := goruntime.NumCPU()
 
 	if err := newAnnotationConfiguration(ocispec, vcAnnotations.DefaultVCPUs).setFloat32WithCheck(func(vcpus float32) error {
-		if vcpus > float32(numCPUs) {
+		if vcpus > float32(numCPUs) && sbConfig.HypervisorType != vc.RemoteHypervisor {
```
stevenhorsman, line 692: Sorry Ajay - I'm not sure if this code will be enough - I just gave it as an example. I think it is still executing `newAnnotationConfiguration` which has the same error check. I think it is worth seeing if there is a unit test that can be added to check the logic if possible?

We also need the same logic for the memory annotation

stevenhorsman (changes requested):

docs/how-to/how-to-set-sandbox-config-kata.md
ajaypvictor: Marking this comment as resolved as we are no more pursuing the solution via this route.

lifupan (approved):

bpradipt (approved):
/lgtm
Thanks @ajaypvictor

stevenhorsman (approved):
The code LGTM. Thanks for adding some new unit tests for this as well!

PR #9659, opened 2024-05-17
ci: cache: Filter out non-printable characters from tag

stevenhorsman:
- The tags have a trailing non-printable character, which results in our cache tags having a trailing underscore e.g. `ghcr.io/kata-containers/cached-artefacts/agent:ce24e9835_` For ease of use of these cached components, we should strip off the trailing underscore.


fidencio (approved):
lgtm, thanks @stevenhorsman!

GabyCT (approved):

PR #9660, opened 2024-05-17
genpolicy: detect empty string in ns as default

malt3:
In Kubernetes, the following values for namespace are equivalent and all refer to the default namespace:

- ` ` (namespace field missing)
- `namespace: ""` (namespace field is the empty string)
- `namespace: "default"`(namespace field has the explicit value `default`)

Genpolicy currently does not handle the empty string case correctly.

### DIFF
--- a/src/tools/genpolicy/src/policy.rs
+++ b/src/tools/genpolicy/src/policy.rs
@@ -481,14 +481,14 @@ impl AgentPolicy {
         let mut root = c_settings.Root.clone();
         root.Readonly = yaml_container.read_only_root_filesystem();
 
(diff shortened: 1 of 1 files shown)


fidencio:
@malt3, please add your Signed-off-by in the git commit message.  Also, please, add the same message that you added to the PR, to the commit message itself.  It'll help a lot folks navigating the commits but not actually looking at the github.

Thanks!

fidencio (approved):
lgtm, thanks @malt3!

danmihai1 (approved):

Redent0r (approved):

PR #9675, opened 2024-05-20
gha: release: Set inherit secrets on tarball builds

stevenhorsman:
Now we have updated the release builds to push artefacts to
our registry for the release, so we can cache the images, we need to set
`secrets: inherit` for all architecture's tarball builds so that we can log into quay.io and ghcr in those steps

This change was tested on release workflow https://github.com/kata-containers/kata-containers/actions/runs/9158766910/job/25179503476 which was green

### DIFF
--- a/.github/workflows/release-amd64.yaml
+++ b/.github/workflows/release-amd64.yaml
@@ -12,6 +12,7 @@ jobs:
     with:
       push-to-registry: yes
       stage: release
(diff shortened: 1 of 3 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman!

GabyCT (approved):

PR #9315, opened 2024-03-20
TEEs: Use `shared_fs=none` for TDX

fidencio:
See each commit message for a better explanation.

At the moment this PR was opened, it's been only done for TDX as that's the machine I easily have access to in case something breaks horribly and I need to do any recovery process. :-)

Issue #9667, opened 2024-05-19
k8s: guest-pull: "Test readonly volume for pods" fails


fidencio:
```
# Events:
#   Type     Reason   Age   From     Message
#   ----     ------   ----  ----     -------
#   Normal   Pulling  82s   kubelet  Pulling image "busybox"
#   Normal   Pulled   81s   kubelet  Successfully pulled image "busybox" in 739ms (739ms including waiting)
#   Normal   Created  81s   kubelet  Created container busybox-file-volume-container
#   Warning  Failed   79s   kubelet  Error: failed to create containerd task: failed to create shim task: failed to handle layer: hasher sha256: channel: send failed SendError { .. }: unknown
# pod "test-file-volume" deleted
```

Well, the volume has to be exposed to the VM somehow ... and this is yet not supported.
This will be the very same case for all the tests related to volumes, in the same way we have the limitations for firecracker.

The tests affected by this limitation are:
* k8s-credentials-secrets.bats
* k8s-file-volume.bats
* k8s-inotify.bats
* k8s-nested-configmap-secret.bats
* k8s-projected-volume.bats
* k8s-volume.bats

wainersm:
* k8s-file-volume.bats

Test's generate yaml:
```yaml
$ cat runtimeclass_workloads_work/test-pod-file-volume.yaml 
#
# Copyright (c) 2022 Ant Group
#
# SPDX-License-Identifier: Apache-2.0
#
apiVersion: v1
kind: Pod
metadata:
  name: test-file-volume
  annotations:
    io.containerd.cri.runtime-handler: kata-qemu-coco-dev
spec:
  terminationGracePeriodSeconds: 0
  runtimeClassName: kata
  restartPolicy: Never
  nodeName: aks-nodepool1-26043497-vmss000000
  volumes:
    - name: shared-file
      hostPath:
        path: /tmp/file-volume-test-foo.Nfy75
        type: File
  containers:
    - name: busybox-file-volume-container
      image: quay.io/prometheus/busybox:latest
      volumeMounts:
        - name: shared-file
          mountPath: /tmp/foo.txt
      command: ["/bin/sh"]
      args: ["-c", "tail -f /dev/null"]
```
It's is spec'ed `runtimeClassName: kata` and  `kata` is handled by `qemu-coco-dev`:
```
$ kubectl get runtimeclass
NAME                 HANDLER              AGE
kata                 kata-qemu-coco-dev   3h50m
[...]

PR #9678, opened 2024-05-20
TEEs: Skip a few CI tests for SEV/SNP

AdithyaKrishnan:
Replicating PR #9315 for AMD SEV/SNP CIs.


fidencio (changes requested):
Just dropping some comments here:
* Please, split the PRs into logical pieces, instead of having one full blob copying what was done for TDX
* Please, verify **each** test that was skipped for both SEV / SNP, instead of just blindly disabling what was disabled on the TDX side.
  * Maybe I disabled something by mistake that could have been re-enabled?
  * Maybe the tests fail on TDX and pass on SEV / SNP?
  * Maybe the tests pass on SEV but not on SNP?
* Instead of copying a whole switch statement several times, add cases to the current statement

Keep in mind, there's no need to rush, this is not a sprint run, but rather a marathon.

Once those are addressed it'll be easier for folks to do a more accurate review.

stevenhorsman:
Hi @AdithyaKrishnan - could you clarify is this PR is supposed to be ready for review. If it is not maybe set it to be draft to indicate this and if it is then I think you need to revisit things:
- Commit e4243e80e9f387899cd833d64f14b88fc825c80e still seems to implement the whole PR and then 835ec2a3d15f0521960c81fbf43c130d37613849 reverts much of it?
- The commit headers and bodies need clarifying and tidying up
- Some of the issues that you are linking to with skips are tdx specific, so not related to SEV/SNP

Once it is ready let us know and we can re-review.

gkurz:
Hi @AdithyaKrishnan !

I fully second @fidencio's and @stevenhorsman's statements. Make sure that all commits have a distinct and informative title and provide details in the commit descriptions that really explain why a specific CI is skipped.

ryansavino (approved):

stevenhorsman (approved):
This is looking a lot better now. Thanks

wainersm:
Hi @ryansavino !

I went through these changes and they seem correct. However, I'm wondering why there are some tests passing on SEV that we'd expect to fail. One example is `k8s-file-volume.bats` which is running and passing. The complete list of tests that we expect to fail due the `shared_fs=none` is in https://github.com/kata-containers/kata-containers/issues/9667

tests/integration/kubernetes/gha-run.sh
```diff
@@ -464,7 +464,7 @@ function cleanup() {
 	fi
 
 	# Switch back to the default namespace and delete the tests one
-	delete_test_cluster_namespace
+	delete_test_cluster_namespace || true
```
portersrc, line 467: Is it OK to bury an error here with `|| true`? Maybe it prints a warning or something, but you could also catch that error case? (Also, unrelated-ish: Why does the comment above say "switch back to the default namespace"; it's not doing that, is it?)

portersrc (approved):
One trivial comment; looks OK from a high level from me

tests/integration/kubernetes/gha-run.sh
fidencio: So, we need to fully understand why `delete_test_cluster_namespace` would fail.
@ryansavino, do you know?  Or, IOW, what issue are you trying to fix here?

ryansavino: This is in the cleanup, and ensuring that the kata deploy cleanup runs. I was running into this scenario when the tests didn't run (namespace wasn't created) or the tests exited in the very beginning.

Not a huge issue, as kata-deploy will try to cleanup before it runs anyways.

fidencio: The cleanup function:
```
function delete_test_cluster_namespace() {
        set_default_cluster_namespace
        kubectl delete namespace "${TEST_CLUSTER_NAMESPACE}"
}
```

I think we can live with this commit in to avoid retriggering CI and yatta-yatta.

Ideally, we should check whether the namespace exists before trying to delete it.  If possible, **please**,  address this in a follow-up PR.

I'm approving this one as it is.

PR #9679, opened 2024-05-20
kata-manager: Copy cni files under /opt/cni

amshinde:
nerdctl requires cni plugins to be installed in /opt/cni/bin
Without bridge plugin installed, it is not possible to run a
container with nerdctl.
The downloaded nerdctl tarball contains cni plugin files, but are
extracted under /usr/local/libexec.
Copy extracted tarball cni files under /usr/local/libexec
to /opt/cni/bin

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -584,6 +584,14 @@ install_nerdctl()
 		done
 
 	info "$project installed\n"
+
+	sudo mkdir -p /opt/cni/bin
+
+	# nerdctl requires cni plugins to be installed in /opt/cni/bin
+	# Copy extracted tarball cni files under /usr/local/libexec
+	sudo cp /usr/local/libexec/cni/* /opt/cni/bin/
+
+	info "cni plugins installed under /opt/cni/bin"
 }
 
 configure_containerd()
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -38,6 +38,10 @@ readonly nerdctl_project="nerdctl"
 readonly nerdctl_releases_url="https://api.github.com/repos/${nerdctl_slug}/releases"
 readonly nerdctl_supported_arches="x86_64 aarch64"
 
+readonly cni_project="cni-plugins"
+readonly cni_slug="containernetworking/plugins"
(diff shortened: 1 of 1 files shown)


utils/kata-manager.sh
```diff
@@ -584,6 +594,46 @@ install_nerdctl()
 		done
 
 	info "$project installed\n"
+
+	sudo mkdir -p /opt/cni/bin
```
Apokleos, line 598: I am not sure this step is required. As at this step, the path of `/opt/cni/bin` does actually exist, right ?

amshinde: This path needs to be created for a fresh install.

Apokleos: I find that there's two different place doing the same operation" mkdir -p /opt/cni/bin", why ?

amshinde: @Apokleos We do this separately for containerd install and nerdctl install.
When nerdctl is being installed, the nerdctl tarball includes containerd and cni files as well, we skip install_containerd in that case.
While containerd files are installed in correct location, the cni files needs to be installed under /opt/cni/bin which is what is happening here.

When only containerd is installed instead of nerdctl, the containerd tarball does not contain cni plugin files. I explicitly install cni in that case which is why you see ` mkdir -p /opt/cni/bin` operation there.

As you can see, the two cases are exclusive.

utils/kata-manager.sh
```diff
@@ -584,6 +594,46 @@ install_nerdctl()
...
+
+	# nerdctl requires cni plugins to be installed in /opt/cni/bin
+	# Copy extracted tarball cni files under /usr/local/libexec
+	sudo cp /usr/local/libexec/cni/* /opt/cni/bin/
```
jodh-intel, line 602: I wonder if we could sym-linke these instead?

jodh-intel (approved):
Thanks, @amshinde - tested and wfm :)

utils/kata-manager.sh
amshinde: I was thinking of sym-linking these first, but then decided against it to keeps things same in case of containerd install where the files are copied.

GabyCT (approved):

PR #9682, opened 2024-05-21
runtime: tdx: Allow default_{cpu,memory} annotations

fidencio:
For now, let's allow the users to set the default_cpu and default_memory when using TDX, as they may hit issues related to the size of the container image that must be pulled and unpacked inside the guest,


bpradipt (approved):
/lgtm
Thanks @fidencio

stevenhorsman (approved):

PR #9684, opened 2024-05-21
ci: cache: Add arch suffix to all cache tags

stevenhorsman:
As we have multi-arch builds for nearly all components, we want to ensure that all the cache tags we set have the architecture suffix, not just the `TARGET_BRANCH` one.

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -1148,9 +1148,11 @@ handle_build() {
 		echo "Pushing ${build_target} with tags: ${tags[*]}"
 
 		for tag in "${tags[@]}"; do
-			# tags can only contain lowercase and uppercase letters, digits, underscores, periods, and hyphens, so
-			# filter out non-printable characers and then replace invalid printable characters with underscode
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @stevenhorsman!

GabyCT (approved):

fidencio:
> Sorry, I changed my mind on the cache implementation _again_ 😢, so have added a new commit to this PR:
> 
> ```
> ci: cache: Update agent cache to use the full commit hash
> 
> - Previously I copied the logic that abbreviated the commit hash
> from the versioning, but looking at our versions.yaml the clear pattern
> is that when pointing at commits of dependencies, we use the full
> commit hash, not the abbreviated one, so for consistency I think we should
> do the same with the components that we make available
> ```
> 
> Please feel free to challenge/re-review this change.

Keep in mind this may result in the naming being way too long, actually going over the expected names for a pod.
Could you double check this won't happen?

stevenhorsman:
> Keep in mind this may result in the naming being way too long, actually going over the expected names for a pod.
Could you double check this won't happen?

When I was looking up the tag valid characters I found that the tag length is limited to 128 chars. I'm not expecting up to easily hit that, but can add a check. The only doc I've found on image name max length is 4096, which seem too long to need to worry about, but I'll try and test if this is reality on ghcr...

GabyCT (approved):

Issue #9688, opened 2024-05-22
Runtime-rs: `dial_timeout` was renamed to `dial_timeout_ms`


justxuewei:
https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/config/configuration-cloud-hypervisor.toml.in#L231

PR #9690, opened 2024-05-22
runtime-rs: Remove obsoleted dial_timeout config

justxuewei:
The `dial_timeout` works fine for Runtime-go, but is obsoleted in
Runtime-rs.

When the pod cannot connect to the Agent upon starting, we need to adjust
the `reconnect_timeout_ms` to increase the number of connection attempts to
the Agent.

Fixes: #9688


Apokleos (approved):
Thx @justxuewei LGTM

lifupan (approved):

Issue #9663, opened 2024-05-19
k8s: Check custom dns test is consistently failing on confidential tests


fidencio:
The test has been failing on TDX for a while:
```
#   Normal   Created    18s (x4 over 78s)  kubelet            Created container test
#   Warning  Failed     13s (x4 over 73s)  kubelet            Error: failed to create containerd task: failed to create shim task: failed to pull manifest error sending request for url (https://quay.io/v2/): error trying to connect: dns error: failed to lookup address information: Try again: unknown
#   Warning  BackOff    13s (x5 over 66s)  kubelet            Back-off restarting failed container test in pod custom-dns-test_kata-containers-k8s-tests(eb8c4fec-5dae-4c58-8b29-89e9f5fadd95)
```

/cc @GabyCT

fidencio:
Note, I also see this one failing on with the non-TEE confidential one, and with SEV.

wainersm:
Hi @fidencio , thanks for raising this!

I sent a PR to disable in all confidential configurations -> https://github.com/kata-containers/kata-containers/pull/9696
And @BbolroC has worked on a fix. We discussed this issue on the meeting today, we are betting this is a regression introduced on confidential image around May 15, or it was already there and only noticed after the changes @stevenhorsman has made on caching (perhaps fixing some hidden bug).

GabyCT:
@wainersm @fidencio I tested the `k8s-custom-dns.bats` at least 15 times using the nydus version `v0.13.12` and kata hypervisor `qemu-coco-dev` and the issue seems gone

PR #9696, opened 2024-05-22
tests/k8s: skip custom DNS tests on confidential jobs

wainersm:
This test has failed in confidential runtime jobs. Skip it until we don't have a fix.

Fixes: #9663


wainersm:
@ryansavino @fidencio @BbolroC could you please review it? The `is_confidential_runtime_class` should return true to qemu-tdx, qemu-sev, qemu-snp and qemu-se. In particular I'm not sure if filtering qemu-se is enough if I need to use arch (s390x).

fitzthum (approved):
cc @AdithyaKrishnan

BbolroC (approved):
LGTM, thanks @wainersm !

fidencio (approved):
lgtm, thanks @wainersm!

Issue #9697, opened 2024-05-22
tests/k8s: disable "fail-fast" behavior by default


wainersm:
This proposal is to disable the fail-fast behavior of our k8s tests because:
 * Often you want to run the entire test suite to check what tests failed, what passed. Usually on big/important changes
 * When running on Jenkins, we had this fail-fast mode disabled so developers got used to it

PR #9698, opened 2024-05-22
tests/k8s: disable "fail-fast" behavior by default

wainersm:
The k8s test suite halts on the first failure, i.e., failing-fast. This isn't the behavior that we used to see when running tests on Jenkins and it seems that running the entire test suite is still the most productive way. So this disable fail-fast by default.

However, if you still wish to run on fail-fast mode then just export K8S_TEST_FAIL_FAST=yes in your environment.

Fixes: #9697

### DIFF
--- a/tests/integration/kubernetes/run_kubernetes_tests.sh
+++ b/tests/integration/kubernetes/run_kubernetes_tests.sh
@@ -14,6 +14,8 @@ TARGET_ARCH="${TARGET_ARCH:-x86_64}"
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 K8S_TEST_DEBUG="${K8S_TEST_DEBUG:-false}"
 K8S_TEST_HOST_TYPE="${K8S_TEST_HOST_TYPE:-small}"
+# Setting to "yes" enables fail fast, stopping execution at the first failed test.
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @wainersm!

sprt:
I thought about doing this a while ago but for some reason that I don't remember, I decided against it - maybe in my testing some failing tests would not clean up properly and thus cause further failures down the chain. Any worries re: this nowadays?

GabyCT (approved):

fidencio:
> I thought about doing this a while ago but for some reason that I don't remember, I decided against it - maybe in my testing some failing tests would not clean up properly and thus cause further failures down the chain. Any worries re: this nowadays?

This may actually happen.  Let me add a do-not-merge label so we can discuss this properly.
Thanks for bringing this up, @sprt!

danmihai1:
I think we'd like to know if some of the tests are not cleaning up properly after a failure. So, unless you guys are already aware of such tests, I think it's worth making this change and exposing these tests.

fidencio:
based on @danmihai1's comment, and on the approbals from other members of the community, I'm removing the `do-not-merge` label.

We can experiment with this, and fix things up as we go if / when we hit the issues.

sprt (approved):

PR #9699, opened 2024-05-23
runtime: Improve error handling and add explicit ignore

JakubLedworowski:
Please refer to individual commits for detailed explanation.

### DIFF
--- a/src/runtime/virtcontainers/fs_share_linux.go
+++ b/src/runtime/virtcontainers/fs_share_linux.go
@@ -407,10 +407,14 @@ func (f *FilesystemShare) UnshareFile(ctx context.Context, c *Container, m *Moun
 		}
 		// Remove the empty file or directory
 		if s.Mode().IsRegular() && s.Size() == 0 {
-			os.Remove(m.HostPath)
+			if err := os.Remove(m.HostPath); err != nil {
+				return errors.Wrapf(err, "Could not remove file from host %v", m.HostPath)
+			}
 		}
 		if s.Mode().IsDir() {
-			syscall.Rmdir(m.HostPath)
+			if err := syscall.Rmdir(m.HostPath); err != nil {
+				return errors.Wrapf(err, "Could not remove directory from host %v", m.HostPath)
+			}
 		}
 	}
 
--- a/src/runtime/virtcontainers/monitor.go
+++ b/src/runtime/virtcontainers/monitor.go
@@ -66,7 +66,7 @@ func (m *monitor) newWatcher(ctx context.Context) (chan error, error) {
 					m.wg.Done()
 					return
 				case <-tick.C:
-					m.watchHypervisor(ctx)
+					_ = m.watchHypervisor(ctx)
 					m.watchAgent(ctx)
 				}
 			}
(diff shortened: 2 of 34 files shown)


JakubLedworowski:
Take a look at those explicitly ignored errors - maybe some of them need attention and handling.

src/runtime/virtcontainers/mount_linux.go
```diff
@@ -180,7 +180,7 @@ func bindUnmountAllRootfs(ctx context.Context, sharedDir string, sandbox *Sandbo
 			mountLogger().WithField("container", c.id).Warnf("container dir is a symlink, malicious guest?")
 			continue
 		}
-		c.unmountHostMounts(ctx)
+		_ = c.unmountHostMounts(ctx)
```
JakubLedworowski, line 183: Does this require handling and propagation to the caller?

src/runtime/virtcontainers/monitor.go
```diff
@@ -140,17 +140,14 @@ func (m *monitor) stop() {
...
-	err := m.sandbox.agent.check(ctx)
-	if err != nil {
+	if err := m.sandbox.agent.check(ctx); err != nil {
 		// TODO: define and export error types
 		m.notify(ctx, errors.Wrapf(err, "failed to ping agent"))
 	}
 }
 
-func (m *monitor) watchHypervisor(ctx context.Context) error {
+func (m *monitor) watchHypervisor(ctx context.Context) {
```
Apokleos, line 149: why remove this method's return value error ?

Apokleos (commented):
Thx @JakubLedworowski one comment

src/runtime/virtcontainers/monitor.go
JakubLedworowski: Thanks for catching that, I'll bring it back. The error in the caller is ignored, though.

JakubLedworowski:
> Thx @JakubLedworowski one comment

@Apokleos I reverted this one change, can we proceed to testing?

src/runtime/virtcontainers/fs_share_linux.go
```diff
@@ -407,10 +407,14 @@ func (f *FilesystemShare) UnshareFile(ctx context.Context, c *Container, m *Moun
 		}
 		// Remove the empty file or directory
 		if s.Mode().IsRegular() && s.Size() == 0 {
-			os.Remove(m.HostPath)
+			if err := os.Remove(m.HostPath); err != nil {
+				return errors.Wrapf(err, "Could not remove file from host %v", m.HostPath)
```
danmihai1, line 411: Returning errors from a clean-up path is tricky. I am a relative beginner when it comes to the Kata runtime code, but the authors' intention may have been to continue the rollbackFailingContainerCreation() clean-up after such error. Logging these errors (if practical) and continuing the clean-up seems reasonable to me.

src/runtime/virtcontainers/mount_linux.go
danmihai1: Agreed, good question.

src/runtime/virtcontainers/fs_share_linux.go
JakubLedworowski: If the intention is to ignore the error because something is not properly implemented and errors occur frequently then it is not clear to me in this part.

There is already an error returned in this function if no such HostPath exist (line 406). The error is propagated in the [call to this function](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/container.go#L550), so I don't understand why some errors are ignored and some are propagated. To me it just looks like this error check was forgotten.

JakubLedworowski: Adding one more thought - the UnshareFile function (and whatever other function) should honestly signal that something bad happened. It should be the function caller's responsibility to handle/propagate/ignore the error.

JakubLedworowski:
Closing as I am not able to finish work on this at the moment.

PR #9706, opened 2024-05-24
genpolicy: add support for devices

### LINKED ISSUE #9651: genpolicy: support raw block devices

The `genpolicy` tool.

**How can it be improved?**

`genpolicy` should accept Kubernetes resources that use `volumeDevice`  and create an appropriate policy for these devices. 

**Additional Information**

This should work:

```
cargo run -- -p rules.rego -j genpolicy-settings.json -u <<EOF
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: voldev
spec:
  replicas: 1
  serviceName: voldev
  selector:
    matchLabels:
      app: voldev
  template:
    metadata:
      labels:
        app: voldev
    spec:
      containers:
      - command:
        - /bin/sh
        - -c
        - |
          stat /dev/csi0
          tail -f /dev/null
        image: alpine:3.19.1
        name: voldev
        volumeDevices:
        - devicePath: /dev/csi0
          name: state
      runtimeClassName: kata-cc-isolation
  volumeClaimTemplates:
  - metadata:
      name: state
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
      storageClassName: default
      volumeMode: Block
EOF
```

Right now, it produces an error because some of the fields are not defined:

```
thread 'main' panicked at src/yaml.rs:341:9:
Unsupported field: spec.template.spec.containers.0.volumeDevices
```

The structs need to be modified to understand volumeDevices and the policy needs to be adjusted to allow the defined devices.

cc @danmihai1

burgerdev:
This PR allows `genpolicy` to parse Kubernetes objects using [raw block volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support), and derives rules for creating these volumes in the guest VM container.

Fixes #9651.

cc @danmihai1

### DIFF
--- a/src/tools/genpolicy/rules.rego
+++ b/src/tools/genpolicy/rules.rego
@@ -54,6 +54,7 @@ default AllowRequestsFailingPolicy := false
 CreateContainerRequest {
     i_oci := input.OCI
     i_storages := input.storages
+    i_devices := input.devices
 
     some p_container in policy_data.containers
     print("======== CreateContainerRequest: trying next policy container")
@@ -76,6 +77,9 @@ CreateContainerRequest {
 
     p_storages := p_container.storages
     allow_by_anno(p_oci, i_oci, p_storages, i_storages)
+
+    p_devices := p_container.devices
+    allow_devices(p_devices, i_devices)
 
     allow_linux(p_oci, i_oci)
 
@@ -328,6 +332,16 @@ allow_log_directory(p_oci, i_oci) {
     print("allow_log_directory: true")
 }
 
+allow_devices(p_devices, i_devices) {
+    print("allow_devices: start")
+    every i_device in i_devices {
+        print("allow_devices: i_device =", i_device)
+        some p_device in p_devices
+        p_device.container_path == i_device.container_path
+    }
+    print("allow_devices: true")
+}
+
 allow_linux(p_oci, i_oci) {
     p_namespaces := p_oci.Linux.Namespaces
     print("allow_linux: p namespaces =", p_namespaces)
@@ -339,6 +353,7 @@ allow_linux(p_oci, i_oci) {
 
     allow_masked_paths(p_oci, i_oci)
     allow_readonly_paths(p_oci, i_oci)
+    allow_linux_devices(p_oci.Linux.Devices, i_oci.Linux.Devices)
 
     print("allow_linux: true")
 }
@@ -425,6 +440,16 @@ allow_readonly_path(p_elem, i_array, masked_paths) {
     p_elem == i_masked
 
     print("allow_readonly_path 2: true")
+}
+
+allow_linux_devices(p_devices, i_devices) {
+    print("allow_linux_devices: start")
+    every i_device in i_devices {
+        print("allow_linux_devices: i_device =", i_device)
+        some p_device in p_devices
+        i_device.Path == p_device.Path
+    }
+    print("allow_linux_devices: true")
 }
 
 # Check the consistency of the input "io.katacontainers.pkg.oci.bundle_path"
--- a/tests/integration/kubernetes/k8s-policy-pvc.bats
+++ b/tests/integration/kubernetes/k8s-policy-pvc.bats
@@ -1,0 +1,66 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2024 Edgeless Systems GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/../../common.bash"
+load "${BATS_TEST_DIRNAME}/tests_common.sh"
+
+setup() {
+	auto_generate_policy_enabled || skip "Auto-generated policy tests are disabled."
+
+	pod_name="policy-pod-pvc"
+	pvc_name="policy-dev"
+
+	get_pod_config_dir
+
+	correct_pod_yaml="${pod_config_dir}/k8s-policy-pod-pvc.yaml"
+	incorrect_pod_yaml="${pod_config_dir}/k8s-policy-pod-pvc-incorrect.yaml"
+	pvc_yaml="${pod_config_dir}/k8s-policy-pvc.yaml"
+
+	# Save some time by executing genpolicy a single time.
+	if [ "${BATS_TEST_NUMBER}" == "1" ]; then
+		# Add policy to the correct pod yaml file
+		auto_generate_policy "${pod_config_dir}" "${correct_pod_yaml}"
+	fi
+
+    # Start each test case with a copy of the correct yaml files.
+	cp "${correct_pod_yaml}" "${incorrect_pod_yaml}"
+}
+
+@test "Successful pod with auto-generated policy" {
+	kubectl create -f "${correct_pod_yaml}"
+	kubectl create -f "${pvc_yaml}"
+	kubectl wait --for=condition=Ready "--timeout=${timeout}" pod "${pod_name}"
+}
+
+# Common function for several test cases from this bats script.
+test_pod_policy_error() {
+	kubectl create -f "${incorrect_pod_yaml}"
+	kubectl create -f "${pvc_yaml}"
+	wait_for_blocked_request "CreateContainerRequest" "${pod_name}"
+}
+
+@test "Policy failure: unexpected device mount" {
+	# Changing the location of a mounted device after policy generation should fail the policy check.
+	yq write -i \
+		"${incorrect_pod_yaml}" \
+		"spec.containers[0].volumeDevices.[0].devicePath" \
+		"/dev/unexpected"
+
+	test_pod_policy_error
+}
+
+teardown() {
+	auto_generate_policy_enabled || skip "Auto-generated policy tests are disabled."
+
+	# Debugging information. Don't print the "Message:" line because it contains a truncated policy log.
+	kubectl describe pod "${pod_name}" | grep -v "Message:"
+
+	# Clean-up
+	kubectl delete -f "${correct_pod_yaml}"
+	kubectl delete -f "${pvc_yaml}"
+	rm -f "${incorrect_pod_yaml}"
+}
--- a/src/tools/genpolicy/src/containerd.rs
+++ b/src/tools/genpolicy/src/containerd.rs
@@ -152,12 +152,14 @@ pub fn get_linux(privileged_container: bool) -> policy::KataLinux {
                 "/proc/sys".to_string(),
                 "/proc/sysrq-trigger".to_string(),
             ],
+            Devices: vec![],
         }
     } else {
         policy::KataLinux {
             Namespaces: vec![],
             MaskedPaths: vec![],
             ReadonlyPaths: vec![],
+            Devices: vec![],
         }
     }
 }
--- a/src/tools/genpolicy/src/persistent_volume_claim.rs
+++ b/src/tools/genpolicy/src/persistent_volume_claim.rs
@@ -34,6 +34,9 @@ struct PersistentVolumeClaimSpec {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     storageClassName: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    volumeMode: Option<String>,
     // TODO: additional fields.
 }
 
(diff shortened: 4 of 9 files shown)


burgerdev:
Note: in its current form, this change breaks integration with Kubernetes device plugins. I'm looking for ways to allow those, probably via the settings, but would appreciate feedback on the overall approach nonetheless.

burgerdev:
The settings file is now considered when allowlisting devices.

danmihai1:
Those changes seem consistent with the design of this tool, and very useful at a quick look - thanks a lot @burgerdev!

However, I am not familiar with the use cases for some of these fields. For example, I haven't seen oci.Linux.Devices being used yet, and it would be helpful for me to understand how that field works.

Therefore, I hope that between you and me we can add some test cases for these changes:
- If you want, I can add some skeleton test where you can contribute additional test cases if you want/can.
- In case you decide to add some test cases without my initial help: the closest examples we currently have are in:

https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-policy-job.bats
https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-policy-pod.bats
https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-policy-rc.bats

- Also, either I or someone else in my team can try harder to do all of the test work if you don't have bandwidth for it.

I would be happy to help with anything I can - please don't hesitate to ping me.

burgerdev:
> However, I am not familiar with the use cases for some of these fields. For example, I haven't seen oci.Linux.Devices being used yet, and it would be helpful for me to understand how that field works.

Afaik containerd translates the [high-level device definition of CRI](https://github.com/kubernetes/cri-api/blob/c75ef5b473bbe2d0a4fc92f82235efd665ea8e9f/pkg/apis/runtime/v1/api.proto#L887-L897) into [their low-level OCI Runtime spec counterparts](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices).

The field is then interpreted by https://github.com/kata-containers/kata-containers/blob/3276bb52b63f865d91fdbac0d5f48f8ef8d74686/src/agent/rustjail/src/mount.rs#L939-L958, which is basically a wrapper around `mknod` (the bind mount implementation seems only used for user namespaces,  don't know why).

I know of two potential use cases for devices: [raw block devices](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support) and [device plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/).

> 
> Therefore, I hope that between you and me we can add some test cases for these changes:
> 
> * If you want, I can add some skeleton test where you can contribute additional test cases if you want/can.
> * In case you decide to add some test cases without my initial help: the closest examples we currently have are in:
> 
> https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-policy-job.bats https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-policy-pod.bats https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-policy-rc.bats
> 
> * Also, either I or someone else in my team can try harder to do all of the test work if you don't have bandwidth for it.
> 
> I would be happy to help with anything I can - please don't hesitate to ping me.

Thanks for the pointers! I should be able to add some simple tests already, but there are scenarios that are hard to build with only k8s resources, but easy if you have access to an agent socket - not sure how to test those... Could we maybe add unit tests that check policy evaluation for a set of canned requests?

danmihai1:
> Thanks for the pointers! I should be able to add some simple tests already, but there are scenarios that are hard to build with only k8s resources, but easy if you have access to an agent socket - not sure how to test those... Could we maybe add unit tests that check policy evaluation for a set of canned requests?

Adding agent socket type of tests is part of our new Project from https://github.com/orgs/kata-containers/projects/44. I suspect it will take a few weeks to get to pretty good testing using this other method though. Let's add any tests that are not too difficult to implement using K8s, and we'll add the others when we get to testing using the agent socket.

Thanks @burgerdev!

danmihai1:
Looks like I created Project 44 with visibility == private by mistake. I will look tomorrow for an org owner to change that visibility.

src/tools/genpolicy/rules.rego
```diff
@@ -328,6 +332,16 @@ allow_log_directory(p_oci, i_oci) {
     print("allow_log_directory: true")
 }
 
+allow_devices(p_devices, i_devices) {
+    print("allow_devices: start")
+    every i_device in i_devices {
+        print("allow_devices: i_device =", i_device)
+        some p_device in p_devices
+        p_device.container_path == i_device.container_path
```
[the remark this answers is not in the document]
danmihai1: I would also consider:

contains(i_device.container_path, "../") == false
endswith(i_device.container_path, "/..") == false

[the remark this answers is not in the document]
burgerdev: If a user specifies a volumeDevice at `/mydev/x`, I don't think we should artificially block that. Instead, my idea here is to ensure that any device asked for by the runtime was explicitly specified in the pod yaml.

Limiting to block devices might make sense for volumes, but I want to cover device plugins, too. GPUs, for example, are usually character devices if I'm not mistaken. We could try to pull these two cases apart, but I'm not sure if it's worth it, given that the host is in full control of the devices anyway and users need to treat these with care.

burgerdev:
I added a test case in the last commit that hopefully illustrates what I'm trying to accomplish. Unfortunately, I don't have access to a test setup to run these: your help in verifying these would be appreciated!

danmihai1:
@burgerdev, please add the new test to the CI list:

```
diff --git a/tests/integration/kubernetes/run_kubernetes_tests.sh b/tests/integration/kubernetes/run_kubernetes_tests.sh
index 7aae576a8..4296cf88f 100755
--- a/tests/integration/kubernetes/run_kubernetes_tests.sh
+++ b/tests/integration/kubernetes/run_kubernetes_tests.sh
@@ -53,6 +53,7 @@ else
                "k8s-pod-quota.bats" \
                "k8s-policy-job.bats" \
                "k8s-policy-pod.bats" \
+               "k8s-policy-pvc.bats" \
                "k8s-policy-rc.bats" \
                "k8s-port-forward.bats" \
                "k8s-projected-volume.bats" \
```

Once you do that, you'll get test feedback from CI.

I had to make the following change to make your first test case pass on my cluster:

```
diff --git a/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod-pvc.yaml b/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod-pvc.yaml
index a7645f22e..03f1bde2c 100644
--- a/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod-pvc.yaml
+++ b/tests/integration/kubernetes/runtimeclass_workloads/k8s-policy-pod-pvc.yaml
@@ -10,7 +10,7 @@ spec:
       image: "quay.io/prometheus/busybox:latest"
       volumeDevices:
         - name: dev
-          devicePath: /mnt/dev0
+          devicePath: /dev/dev0
   volumes:
     - name: dev
       persistentVolumeClaim:
```

Without that YAML change, I was getting:

```
Normal   Pulled                  74s                kubelet                  Successfully pulled image "quay.io/prometheus/busybox:latest" in 271ms (271ms including waiting)
Normal   Pulled                  72s                kubelet                  Successfully pulled image "quay.io/prometheus/busybox:latest" in 124ms (124ms including waiting)
Normal   Pulled                  59s                kubelet                  Successfully pulled image "quay.io/prometheus/busybox:latest" in 167ms (167ms including waiting)
Normal   Pulling                 34s (x4 over 74s)  kubelet                  Pulling image "quay.io/prometheus/busybox:latest"
Normal   Created                 34s (x4 over 74s)  kubelet                  Created container busybox
Normal   Pulled                  34s                kubelet                  Successfully pulled image "quay.io/prometheus/busybox:latest" in 133ms (133ms including waiting)
Warning  Failed                  33s (x4 over 73s)  kubelet                  Error: failed to create containerd task: failed to create shim task: /mnt/dev0 is not a valid device path: unknown
Warning  BackOff                 7s (x7 over 72s)   kubelet                  Back-off restarting failed container busybox in pod policy-pod-pvc_kata-containers-k8s-tests(81a3eab4-7cb8-4675-99e2-b2370a401a07)
```

As far as I know, there is nothing special on my test cluster that would enforce devicePath = /mnt/path. It is just a coincidence that I suggested earlier enforcing devicePath = /mnt/path using the Policy, but other paths don't seem to work anyway according to my experiment. You should be able to get the same error from CI if you keep the current test YAML.

tests/integration/kubernetes/k8s-policy-pvc.bats
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bats
+#
+# Copyright (c) 2024 Microsoft.
```
danmihai1, line 3: I vaguely remember that reviewers directed me to change Intel -> Microsoft for a similar change. So, I suggest changing this line too.

danmihai1 (changes requested):
Waiting for those test changes, and to sort out the /mnt device path issue.

tests/integration/kubernetes/k8s-policy-pvc.bats
burgerdev: Changed to Intel. The other policy tests still say Microsoft, though.

burgerdev:
That's interesting! I think this is a restriction added by Kata, creating a pod with non-/dev mount point works with the default runtime on AKS. Changed the test setup accordingly, but either way I don't think we need to check the prefix.

danmihai1:
@burgerdev , when you add a new file, my understanding is that you should use Copyright Edgeless Systems.

burgerdev:
Thanks for the pointer :)

src/tools/genpolicy/rules.rego
```diff
@@ -339,6 +353,7 @@ allow_linux(p_oci, i_oci) {
 
     allow_masked_paths(p_oci, i_oci)
     allow_readonly_paths(p_oci, i_oci)
+    allow_linux_devices(p_oci.Linux.Devices, i_oci.Linux.Devices)
```
danmihai1, line 356: @burgerdev , without adding tests for this field, I would be more comfortable initially with just:

is_null(p_oci.Linux.Devices)

But, if you think we should add such tests later, please log an issue as a reminder for that future work.

danmihai1 (approved):

src/tools/genpolicy/rules.rego
```diff
@@ -339,6 +353,7 @@ allow_linux(p_oci, i_oci) {
...
 
     print("allow_linux: true")
```
danmihai1, line 358: I would also add:

count(i_oci.Linux.Resources.Devices) == 0

But we could try this in a separate PR.

src/tools/genpolicy/rules.rego
burgerdev: When mounting a block device, both `.devices` and `.OCI.Linux.Devices` are set by the runtime (example request from AKS-CoCo: https://gist.github.com/burgerdev/3a1ae37870e8ee53d603bb9f5d3a2069). I don't know why that's the case, but the test won't work if we don't allow the Linux device here.

fidencio:
I'll go through this one Tomorrow.

fidencio (approved):
lgtm, thanks @burgerdev.

Please, let's try to address Dan's comments on a follow-up.

burgerdev:
Thanks for the merge!

As per https://github.com/kata-containers/kata-containers/pull/9706#discussion_r1624683480, I don't think we can make the policy any tighter, so maybe there's no need for follow-up. Thoughts, @danmihai1?

danmihai1:
> Thanks for the merge!
> 
> As per [#9706 (comment)](https://github.com/kata-containers/kata-containers/pull/9706#discussion_r1624683480), I don't think we can make the policy any tighter, so maybe there's no need for follow-up. Thoughts, @danmihai1?

@burgerdev @fidencio , I will send out a separate PR with additional tightening - e.g., for the Linux.Resources.Devices field. So, somewhat related to this PR. The general idea will be:

1. If we can't test an input field, and
2. The field could be significant for the confidentiality of the Guest 

then the auto-generated policy should typically block untested values of that field. When/if we'll find reasonable use cases that require these fields, we might decide to relax these new rules.

Please stay tuned.

PR #9719, opened 2024-05-25
Support Confidential Sealed Secrets (as env vars)

fitzthum:
This PR brings sealed secret functionality to main. This functionality was originally implemented by @LindaYu17 and @Lu-Biao.

I have made some improvements and converted the [old sealed secrets test](https://github.com/kata-containers/tests/blob/CCv0/integration/kubernetes/confidential/sealed_secret.bats) from the defunct CCv0 branch of the tests repo into a test that is similar to the existing CoCo attestation test.

There are still refinements to be made (especially in the last PR). I have done some local testing, but let's see what the CI makes of it.

### DIFF
--- a/src/agent/src/cdh.rs
+++ b/src/agent/src/cdh.rs
@@ -1,0 +1,62 @@
+// Copyright (c) 2023 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Confidential Data Hub client wrapper.
+// Confidential Data Hub is a service running inside guest to provide resource related APIs.
+// https://github.com/confidential-containers/guest-components/tree/main/confidential-data-hub
+
+use anyhow::Result;
+use derivative::Derivative;
+use protocols::{
+    sealed_secret, sealed_secret_ttrpc_async, sealed_secret_ttrpc_async::SealedSecretServiceClient,
+};
+
+use crate::CDH_SOCKET_URI;
+
+// Nanoseconds
+const CDH_UNSEAL_TIMEOUT: i64 = 50 * 1000 * 1000 * 1000;
+const SEALED_SECRET_PREFIX: &str = "sealed.";
+
+#[derive(Derivative)]
+#[derivative(Clone, Debug)]
+pub struct CDHClient {
+    #[derivative(Debug = "ignore")]
+    sealed_secret_client: SealedSecretServiceClient,
+}
+
+impl CDHClient {
+    pub fn new() -> Result<Self> {
+        let client = ttrpc::asynchronous::Client::connect(CDH_SOCKET_URI)?;
+        let sealed_secret_client =
+            sealed_secret_ttrpc_async::SealedSecretServiceClient::new(client);
+
+        Ok(CDHClient {
+            sealed_secret_client,
+        })
+    }
+
+    pub async fn unseal_secret_async(&self, sealed_secret: &str) -> Result<Vec<u8>> {
+        let mut input = sealed_secret::UnsealSecretInput::new();
+        input.set_secret(sealed_secret.into());
+
+        let unsealed_secret = self
+            .sealed_secret_client
+            .unseal_secret(ttrpc::context::with_timeout(CDH_UNSEAL_TIMEOUT), &input)
+            .await?;
+        Ok(unsealed_secret.plaintext)
+    }
+
+    pub async fn unseal_env(&self, env: &str) -> Result<String> {
+        if let Some((key, value)) = env.split_once('=') {
+            if value.starts_with(SEALED_SECRET_PREFIX) {
+                let unsealed_value = self.unseal_secret_async(value).await?;
+                let unsealed_env = format!("{}={}", key, std::str::from_utf8(&unsealed_value)?);
+
+                return Ok(unsealed_env);
+            }
+        }
+        Ok((*env.to_owned()).to_string())
+    }
+}
--- a/src/agent/src/cdh.rs
+++ b/src/agent/src/cdh.rs
@@ -57,6 +57,94 @@ impl CDHClient {
                 return Ok(unsealed_env);
             }
         }
+
         Ok((*env.to_owned()).to_string())
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "sealed-secret")]
+mod tests {
+    use crate::cdh::CDHClient;
+    use crate::cdh::CDH_ADDR;
+    use anyhow::anyhow;
+    use async_trait::async_trait;
+    use protocols::{sealed_secret, sealed_secret_ttrpc_async};
+    use std::sync::Arc;
+    use test_utils::skip_if_not_root;
+    use tokio::signal::unix::{signal, SignalKind};
+
+    struct TestService;
+
+    #[async_trait]
+    impl sealed_secret_ttrpc_async::SealedSecretService for TestService {
+        async fn unseal_secret(
+            &self,
+            _ctx: &::ttrpc::asynchronous::TtrpcContext,
+            _req: sealed_secret::UnsealSecretInput,
+        ) -> ttrpc::error::Result<sealed_secret::UnsealSecretOutput> {
+            let mut output = sealed_secret::UnsealSecretOutput::new();
+            output.set_plaintext("unsealed".into());
+            Ok(output)
+        }
+    }
+
+    fn remove_if_sock_exist(sock_addr: &str) -> std::io::Result<()> {
+        let path = sock_addr
+            .strip_prefix("unix://")
+            .expect("socket address does not have the expected format.");
+
+        if std::path::Path::new(path).exists() {
+            std::fs::remove_file(path)?;
+        }
+
+        Ok(())
+    }
+
+    fn start_ttrpc_server() {
+        tokio::spawn(async move {
+            let ss = Box::new(TestService {})
+                as Box<dyn sealed_secret_ttrpc_async::SealedSecretService + Send + Sync>;
+            let ss = Arc::new(ss);
+            let ss_service = sealed_secret_ttrpc_async::create_sealed_secret_service(ss);
+
+            remove_if_sock_exist(CDH_ADDR).unwrap();
+
+            let mut server = ttrpc::asynchronous::Server::new()
+                .bind(CDH_ADDR)
+                .unwrap()
+                .register_service(ss_service);
+
+            server.start().await.unwrap();
+
+            let mut interrupt = signal(SignalKind::interrupt()).unwrap();
+            tokio::select! {
+                _ = interrupt.recv() => {
+                    server.shutdown().await.unwrap();
+                }
+            };
+        });
+    }
+
+    #[tokio::test]
+    async fn test_unseal_env() {
+        skip_if_not_root!();
+
+        let rt = tokio::runtime::Runtime::new().unwrap();
+        let _guard = rt.enter();
+        start_ttrpc_server();
+        std::thread::sleep(std::time::Duration::from_secs(2));
+
+        let cc = Some(CDHClient::new().unwrap());
+        let cdh_client = cc.as_ref().ok_or(anyhow!("get cdh_client failed")).unwrap();
+        let sealed_env = String::from("key=sealed.testdata");
+        let unsealed_env = cdh_client.unseal_env(&sealed_env).await.unwrap();
+        assert_eq!(unsealed_env, String::from("key=unsealed"));
+        let normal_env = String::from("key=testdata");
+        let unchanged_env = cdh_client.unseal_env(&normal_env).await.unwrap();
+        assert_eq!(unchanged_env, String::from("key=testdata"));
+
+        rt.shutdown_background();
+        std::thread::sleep(std::time::Duration::from_secs(2));
     }
 }
--- a/src/agent/src/main.rs
+++ b/src/agent/src/main.rs
@@ -38,6 +38,7 @@ use std::process::Command;
 use std::sync::Arc;
 use tracing::{instrument, span};
 
+mod cdh;
 mod config;
 mod console;
 mod device;
@@ -59,6 +60,7 @@ mod util;
 mod version;
 mod watcher;
 
+use cdh::CDHClient;
 use config::GuestComponentsProcs;
 use mount::{cgroups_mount, general_mount};
 use sandbox::Sandbox;
@@ -104,6 +106,7 @@ const AA_ATTESTATION_URI: &str = concatcp!(UNIX_SOCKET_PREFIX, AA_ATTESTATION_SO
 
 const CDH_PATH: &str = "/usr/local/bin/confidential-data-hub";
 const CDH_SOCKET: &str = "/run/confidential-containers/cdh.sock";
+const CDH_SOCKET_URI: &str = concatcp!(UNIX_SOCKET_PREFIX, CDH_SOCKET);
 
 const API_SERVER_PATH: &str = "/usr/local/bin/api-server-rest";
 
@@ -403,6 +406,7 @@ async fn start_sandbox(
     let (tx, rx) = tokio::sync::oneshot::channel();
     sandbox.lock().await.sender = Some(tx);
 
+    let mut cdh_client = None;
     let gc_procs = config.guest_components_procs;
     if gc_procs != GuestComponentsProcs::None {
         if !attestation_binaries_available(logger, &gc_procs) {
@@ -411,12 +415,19 @@ async fn start_sandbox(
                 "attestation binaries requested for launch not available"
             );
         } else {
-            init_attestation_components(logger, config)?;
+            cdh_client = init_attestation_components(logger, config)?;
         }
     }
 
     // vsock:///dev/vsock, port
-    let mut server = rpc::start(sandbox.clone(), config.server_addr.as_str(), init_mode).await?;
+    let mut server = rpc::start(
+        sandbox.clone(),
+        config.server_addr.as_str(),
+        init_mode,
+        cdh_client,
+    )
+    .await?;
+
     server.start().await?;
 
     rx.await?;
@@ -445,10 +456,11 @@ fn attestation_binaries_available(logger: &Logger, procs: &GuestComponentsProcs)
 // Start-up attestation-agent, CDH and api-server-rest if they are packaged in the rootfs
 // and the corresponding procs are enabled in the agent configuration. the process will be
 // launched in the background and the function will return immediately.
-fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<()> {
+// If the CDH is started, a CDH client will be instantiated and returned.
+fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<Option<CDHClient>> {
     // skip launch of any guest-component
     if config.guest_components_procs == GuestComponentsProcs::None {
-        return Ok(());
+        return Ok(None);
     }
 
     debug!(logger, "spawning attestation-agent process {}", AA_PATH);
@@ -463,7 +475,7 @@ fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<
 
     // skip launch of confidential-data-hub and api-server-rest
     if config.guest_components_procs == GuestComponentsProcs::AttestationAgent {
-        return Ok(());
+        return Ok(None);
     }
 
     debug!(
@@ -479,9 +491,11 @@ fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<
     )
     .map_err(|e| anyhow!("launch_process {} failed: {:?}", CDH_PATH, e))?;
 
+    let cdh_client = CDHClient::new().context("Failed to create CDH Client")?;
+
     // skip launch of api-server-rest
     if config.guest_components_procs == GuestComponentsProcs::ConfidentialDataHub {
-        return Ok(());
+        return Ok(Some(cdh_client));
     }
 
     let features = config.guest_components_rest_api;
@@ -498,7 +512,7 @@ fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<
     )
     .map_err(|e| anyhow!("launch_process {} failed: {:?}", API_SERVER_PATH, e))?;
 
-    Ok(())
+    Ok(Some(cdh_client))
 }
 
 fn wait_for_path_to_exist(logger: &Logger, path: &str, timeout_secs: i32) -> Result<()> {
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -76,6 +76,8 @@ use crate::policy::{do_set_policy, is_allowed};
 #[cfg(feature = "guest-pull")]
 use crate::image;
 
+use crate::cdh::CDHClient;
+
 use opentelemetry::global;
 use tracing::span;
 use tracing_opentelemetry::OpenTelemetrySpanExt;
@@ -171,6 +173,7 @@ impl<T> OptionToTtrpcResult<T> for Option<T> {
 pub struct AgentService {
     sandbox: Arc<Mutex<Sandbox>>,
     init_mode: bool,
+    cdh_client: Option<CDHClient>,
 }
 
 impl AgentService {
@@ -221,6 +224,22 @@ impl AgentService {
         // match real devices inside the VM. This step is necessary since we
         // cannot predict everything from the caller.
         add_devices(&req.devices, &mut oci, &self.sandbox).await?;
+
+        if let Some(cdh) = self.cdh_client.as_ref() {
+            let process = oci
+                .process
+                .as_mut()
+                .ok_or_else(|| anyhow!("Spec didn't contain process field"))?;
+
+            for env in process.env.iter_mut() {
+                match cdh.unseal_env(env).await {
+                    Ok(unsealed_env) => *env = unsealed_env.to_string(),
+                    Err(e) => {
+                        warn!(sl(), "Failed to unseal secret: {}", e)
+                    }
+                }
+            }
+        }
 
         // Both rootfs and volumes (invoked with --volume for instance) will
         // be processed the same way. The idea is to always mount any provided
@@ -1601,10 +1620,12 @@ pub async fn start(
     s: Arc<Mutex<Sandbox>>,
     server_address: &str,
     init_mode: bool,
+    cdh_client: Option<CDHClient>,
 ) -> Result<TtrpcServer> {
     let agent_service = Box::new(AgentService {
         sandbox: s,
         init_mode,
+        cdh_client,
     }) as Box<dyn agent_ttrpc::AgentService + Send + Sync>;
     let aservice = agent_ttrpc::create_agent_service(Arc::new(agent_service));
 
@@ -2148,6 +2169,7 @@ mod tests {
         let agent_service = Box::new(AgentService {
             sandbox: Arc::new(Mutex::new(sandbox)),
             init_mode: true,
+            cdh_client: None,
         });
 
         let req = protocols::agent::UpdateInterfaceRequest::default();
@@ -2162,10 +2184,10 @@ mod tests {
     async fn test_update_routes() {
         let logger = slog::Logger::root(slog::Discard, o!());
         let sandbox = Sandbox::new(&logger).unwrap();
-
         let agent_service = Box::new(AgentService {
             sandbox: Arc::new(Mutex::new(sandbox)),
             init_mode: true,
+            cdh_client: None,
         });
 
         let req = protocols::agent::UpdateRoutesRequest::default();
@@ -2180,10 +2202,10 @@ mod tests {
     async fn test_add_arp_neighbors() {
         let logger = slog::Logger::root(slog::Discard, o!());
         let sandbox = Sandbox::new(&logger).unwrap();
-
         let agent_service = Box::new(AgentService {
             sandbox: Arc::new(Mutex::new(sandbox)),
             init_mode: true,
+            cdh_client: None,
         });
 
         let req = protocols::agent::AddARPNeighborsRequest::default();
@@ -2322,6 +2344,7 @@ mod tests {
             let agent_service = Box::new(AgentService {
                 sandbox: Arc::new(Mutex::new(sandbox)),
                 init_mode: true,
+                cdh_client: None,
             });
 
             let result = agent_service
@@ -2811,6 +2834,7 @@ OtherField:other
         let agent_service = Box::new(AgentService {
             sandbox: Arc::new(Mutex::new(sandbox)),
             init_mode: true,
+            cdh_client: None,
         });
 
         let ctx = mk_ttrpc_context();
--- a/tests/integration/kubernetes/k8s-sealed-secret.bats
+++ b/tests/integration/kubernetes/k8s-sealed-secret.bats
@@ -1,0 +1,121 @@
+#!/usr/bin/env bats
+# Copyright 2024 IBM Corporation
+# Copyright 2024 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Test for Sealed Secret feature of CoCo
+#
+
+load "${BATS_TEST_DIRNAME}/lib.sh"
+load "${BATS_TEST_DIRNAME}/confidential_common.sh"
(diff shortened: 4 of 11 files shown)


fidencio:
Hey @fitzthum!  First of all, thanks for taking over this work and opening the PR, very much appreciated.

I see this touches mainly 2 parts of the code:
* runtime, which can be enabled / disabled in each specific configuration file
* agent, which can be enabled / disabled only during build time

When we have something enabled / disabled only during build time on the agent side, it basically means that in order to properly test it we'd have to create a new agent build for that specific case, and then add this new specific agent into the rootfs / initrd, which is a pain as it leads to a way too huge amount of things to test.

With this in mind, I´d like to ask you, is there the possibility to have this still as an option, still defaulting to not using it in the agent's Makefile, but enabled by default in the agent build (here: https://github.com/kata-containers/kata-containers/blob/b89c3e35dd40443349538bcf5137ce7d0fe09cb9/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L821)?  This would simplify things a lot, if doable.  If you consider this not doable, mind to explain what are the issues you foresee with this approach?

fidencio:
The way we're currently building the agent is different than at the time this code was first written.

When the code was originally written, the rootfs build would trigger the agent build, meaning that the `SEALED_SECRET` option must be passed to the rootfs, even if it has nothing to do with the rootfs itself, but rather with the agent build.

Nowadays, we build the agent separetely, and then just unpacl the agent inside the rootfs, see the code to build the agent here: https://github.com/kata-containers/kata-containers/blob/028b10ce7a30bec005f850f2f27de2d0bb17aeca/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L802-L821

So, instead of passing `SEALED_SECRET` to the rootfs, we need to actually pass it to the agent builder, and in my previous comment I mentioned about setting it always to yes / true.

Anyways, this wall of text was basically to explain that the commit https://github.com/kata-containers/kata-containers/pull/9719/commits/26ef03b386f2c2bca79ba23265c18e03c0d485cf can be dropped from the series.

fitzthum:
@fidencio Currently the sealed-secret stuff in the agent is behind the `sealed-secret` feature, which is disabled by default in the Makefile. I add the `SEALED_SECRET=yes` to `install_initrd_confidential` and `install_image_confidential`. I'm not sure if that's the best place to put it, but the idea is to only enable this for CoCo builds, since the feature depends on the CDH.

You are suggesting setting `SEALED_SECRET` in `install_agent`. Would this enable it for non-coco builds as well? I could potentially rework the code a bit fail gracefully if the CDH is not there, but the feature does increase the binary size.

fidencio:
> You are suggesting setting `SEALED_SECRET` in `install_agent`. Would this enable it for non-coco builds as well? I could potentially rework the code a bit fail gracefully if the CDH is not there, but the feature does increase the binary size

I'd prefer that.  I understand that it'll increase the binary size, but what we ship here is an example of what people should use, and I sincerely hope people end up building their own content when they decide to get kata-containers into their production environment.  With that said, as long as we leave a way for folks to build the agent without sealed secrets support, we're good.  That's better than having to do deal with different agent builds.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -921,12 +921,18 @@ func (k *kataAgent) stopSandbox(ctx context.Context, sandbox *Sandbox) error {
...
-func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts map[string]Mount) error {
+func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts map[string]Mount, sealedSecretEnabled bool) error {
 	ociMounts := spec.Mounts
 
 	for index, m := range ociMounts {
 		if guestMount, ok := guestMounts[m.Destination]; ok {
 			k.Logger().Debugf("Replacing OCI mount (%s) source %s with %s", m.Destination, m.Source, guestMount.Source)
+			if sealedSecretEnabled {
+				if strings.Contains(m.Source, "kubernetes.io~secret") {
+					ociMounts[index].Destination = "/sealed" + m.Destination
```
fitzthum, line 932: I'm not sure this approach will allow sealed secrets and non-sealed secrets to coexist. Any ideas how to improve? cc: @Xynnn007 maybe

Ideally I'd like to remove the shim support entirely.

fitzthum:
Fixed up static checks.

I think I will wait for https://github.com/kata-containers/kata-containers/pull/9749 before reshuffling how the feature is enabled.

src/runtime/virtcontainers/kata_agent.go
Xynnn007: I am not expert upon this. But at the firt glance, this logic will add all secrets (sealed/unsealed) with "/sealed" prefix. Togetherly, at kata-agent side all prefixs will be striped and put all files under `/run/secrets/`. Whether to unseal the file is determined by the content format of the secret file. Thus I am afraid we might not need both add-prefix on runtime side and delete-prefix part on agent side.

I am not sure, we might need to do some test for this. With both sealed and non-sealed secret binding to a container and check if it works.

Xynnn007 (commented):
There are some comments. Most are about the code architecture.

src/runtime/virtcontainers/kata_agent.go
fitzthum: Yeah i suspect there are a couple ways to get rid of the runtime stuff.

fitzthum:
I have some significant changes planned here so I will make this a draft.

fitzthum:
I have found a few issues with the implementation of sealed secrets as files. I have a plan for fixing this, but I think the smart thing to do is to break this PR apart and start only with secrets in environment variables, which are simpler. To that end, I have dropped a couple commits from this PR. I will re-introduce them in a follow-up. Notice that we don't actually need any shim support for sealed secrets to work with environment variables. I am hoping for the same thing when I do secrets in files.

Tomorrow I will finally rework the way the feature is loaded and address other comments.

src/agent/src/cdh.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        Ok(CDHClient {
+            sealed_secret_client,
+        })
+    }
+
+    pub async fn unseal_secret_async(&self, sealed_secret: &str) -> Result<Vec<u8>> {
+        let sealed_secret = sealed_secret
+            .strip_prefix("sealed.")
+            .ok_or(anyhow!("secret prefix not found"))?;
```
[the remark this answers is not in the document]
fitzthum: I removed this and bumped guest-components. Hopefully this won't cause any other issues.

src/agent/src/cdh.rs
```diff
@@ -0,0 +1,58 @@
...
+        let (key, value) = env.split_once('=').unwrap();
+        if value.starts_with("sealed.") {
+            let unsealed_value = self.unseal_secret_async(value).await?;
+            let unsealed_env = format!("{}={}", key, std::str::from_utf8(&unsealed_value).unwrap());
+
+            return Ok(unsealed_env);
+        }
+        Ok((*env.to_owned()).to_string())
+    }
+}
```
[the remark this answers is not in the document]
fitzthum: I am a bit split on this. I would prefer to keep the handling of the env outside of `cdh.rs`, but I would also like to hide as much complexity as possible from `rpc.rs`. Maybe other reviewers have a viewpoint on this?

src/agent/src/rpc.rs
```diff
@@ -167,10 +171,14 @@ impl<T> OptionToTtrpcResult<T> for Option<T> {
...
 
-#[derive(Clone, Debug)]
+#[derive(Derivative)]
+#[derivative(Clone, Debug)]
 pub struct AgentService {
     sandbox: Arc<Mutex<Sandbox>>,
     init_mode: bool,
+    #[derivative(Debug = "ignore")]
+    #[cfg(feature = "sealed-secret")]
+    cdh_client: Option<CDHClient>,
```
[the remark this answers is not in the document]
fitzthum: This is a good suggestion, but I have refactored the setup of the client to work a bit differently. Now it is instantiated based on what is specified in the new `gc_procs` setting. Thus, I am continuing to expose the entire cdh_client to rpc.rs. I think this will also be helpful when we start to introduce other capabilities that depend on cdh.rs.

src/agent/src/rpc.rs
```diff
@@ -222,6 +226,21 @@ impl AgentService {
...
+            let process = oci
+                .process
+                .as_mut()
+                .ok_or_else(|| anyhow!("Spec didn't contain process field"))?;
+
+            for env in process.env.iter_mut() {
+                let unsealed_env = cdh
+                    .unseal_env(env)
+                    .await
+                    .map_err(|e| anyhow!("unseal env failed: {:?}", e))?;
```
fitzthum, line 239: Maybe this should be non-fatal?

fitzthum:
I have reshuffled how this feature is enabled. There is no longer any feature required for sealed-secret support. Instead, it can be enabled at runtime simply be setting `guest-components-procs` to either `confidential-data-hub` or `api-server-rest`.

src/agent/src/rpc.rs
```diff
@@ -1601,10 +1620,19 @@ pub async fn start(
     s: Arc<Mutex<Sandbox>>,
     server_address: &str,
     init_mode: bool,
+    gc_procs: GuestComponentsProcs,
 ) -> Result<TtrpcServer> {
+    let cdh_client = match gc_procs {
+        GuestComponentsProcs::ConfidentialDataHub | GuestComponentsProcs::ApiServerRest => {
```
fitzthum, line 1626: Is there also an `all` variant?

fitzthum: (there isn't)

src/agent/src/cdh.rs
```diff
@@ -0,0 +1,57 @@
...
+// Confidential Data Hub client wrapper.
+// Confidential Data Hub is a service running inside guest to provide resource related APIs.
+// https://github.com/confidential-containers/guest-components/tree/main/confidential-data-hub
+
+use anyhow::Result;
+use derivative::Derivative;
+use protocols::{
+    sealed_secret, sealed_secret_ttrpc_async, sealed_secret_ttrpc_async::SealedSecretServiceClient,
+};
+const CDH_ADDR: &str = "unix:///run/confidential-containers/cdh.sock";
```
wainersm, line 15: !!!NIT ALERT!!! I would call it `CDH_SOCKET` as ADDR can be misleading (think the cdh's http address).

BTW, `src/main.rs` already defines `CDH_SOCKET`. I think it would be good to refactor the main's code to re-use this `CDH_SOCKET`. You may want to actually split in those variables, `CDH_SOCKET` and `CDH_SOCKET_URI` just like:

```
const AA_ATTESTATION_SOCKET: &str =
    "/run/confidential-containers/attestation-agent/attestation-agent.sock";
const AA_ATTESTATION_URI: &str = concatcp!(UNIX_SOCKET_PREFIX, AA_ATTESTATION_SOCKET);
```

src/agent/src/cdh.rs
```diff
@@ -0,0 +1,57 @@
...
+const CDH_TIMEOUT: i64 = 50 * 1000 * 1000 * 1000;
```
wainersm, line 16: IIUC from the (a little poor) documentation https://docs.rs/ttrpc/latest/ttrpc/context/struct.Context.html, this should be in nanoseconds. So leave a comment please.

Also rename it to something like `CDH_UNSEAL_SECRET_TIMEOUT`, more meaningful.

src/agent/src/cdh.rs
```diff
@@ -0,0 +1,57 @@
...
+        let unsealed_secret = self
+            .sealed_secret_client
+            .unseal_secret(ttrpc::context::with_timeout(CDH_TIMEOUT), &input)
+            .await?;
+        Ok(unsealed_secret.plaintext)
+    }
+
+    pub async fn unseal_env(&self, env: &str) -> Result<String> {
+        let (key, value) = env.split_once('=').unwrap();
+        if value.starts_with("sealed.") {
```
wainersm, line 49: We could make this "sealed." as a constant (eg. `SEALED_PREFIX`) and document the expected format of a sealed secret. We also could have a simple parser.

I'm also wondering if we shouldn't sanitize the input to avoid kind of injection attack (really don't know if a attack here is doable).

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,97 @@
...
+	# "type": "vault",
+	# "name": "kbs:///default/sealed-secret/test",
+	# "provider": "kbs",
+	# "provider_settings": {},
+	# "annotations": {}
+	# }
+	kubectl create secret generic sealed-secret --from-literal='password=sealed.fakejwsheader.ewogICAgInZlcnNpb24iOiAiMC4xLjAiLAogICAgInR5cGUiOiAidmF1bHQiLAogICAgIm5hbWUiOiAia2JzOi8vL2RlZmF1bHQvc2VhbGVkLXNlY3JldC90ZXN0IiwKICAgICJwcm92aWRlciI6ICJrYnMiLAogICAgInByb3ZpZGVyX3NldHRpbmdzIjoge30sCiAgICAiYW5ub3RhdGlvbnMiOiB7fQp9Cg==.fakesignature'
+}
+
+@test "Unseal Secrets with CDH" {
```
wainersm, line 66: Idea of new test: mix sealed and non-sealed secrects in a single deployment to check it can handle well (if possible make a unit test instead).

src/agent/src/cdh.rs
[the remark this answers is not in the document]
wainersm: Brainstorming: I read this comment after I began to review the code. I confess that I was wondering if those unseal methods wouldn't fit better somewhere else. Sealed secrets seems to not belong to the "cdh" concept, but it seems rather a simple user of cdh (to fetch the secret for unsealing the secret).

src/agent/src/cdh.rs
fitzthum: I can't think of any issues with injection, but I'll think about it a bit more.

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: Do you think we should add a new test for this or just extend the existing one? Can this be a follow-up PR?

fitzthum:
Non-tee tests passing (including with the sealed-secret test). Other tests looking ok too. Good time for some more review.

src/agent/src/main.rs
```diff
@@ -412,11 +415,19 @@ async fn start_sandbox(
...
         }
     }
 
     // vsock:///dev/vsock, port
-    let mut server = rpc::start(sandbox.clone(), config.server_addr.as_str(), init_mode).await?;
+    let mut server = rpc::start(
+        sandbox.clone(),
+        config.server_addr.as_str(),
+        init_mode,
+        gc_enabled,
```
Xynnn007, line 427: Maybe we could directly pass `Some(CDHClient)` as parameter here? which would help to avoid splited condition judgement on different places.

Xynnn007 (approved):
Cool. The pr looks good. Only a nit

src/agent/src/main.rs
fitzthum: Good idea. I moved the initialization of the CDHClient to `init_attestation_components` so we can create the CDHClient right after starting the CDIH.

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,98 @@
...
+load "${BATS_TEST_DIRNAME}/lib.sh"
+load "${BATS_TEST_DIRNAME}/confidential_common.sh"
+load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
+
+export KBS="${KBS:-false}"
+export KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
+export AA_KBC="${AA_KBC:-cc_kbc}"
+
+setup() {
+	[ "${KATA_HYPERVISOR}" = "qemu-coco-dev" ] || skip "Test not ready yet for ${KATA_HYPERVISOR}"
```
stevenhorsman, line 19: Could this work on TDX/SE (who have attestation support working) now? If so we could swap this to 
`is_confidential_runtime_class || skip "Test not supported for ${KATA_HYPERVISOR}."` like the attestation-agent and let the KBS check rule out platforms that don't support that.
If you have already checked and that they are failing currently, that I understand getting it working for just the non-TEE first, so this is fine.

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,98 @@
...
+	export K8S_TEST_YAML="${pod_config_dir}/pod-sealed-secret.yaml"
+	# Schedule on a known node so that later it can print the system's logs for
+	# debugging.
+	set_node "$K8S_TEST_YAML" "$node"
+
+	kbs_set_resource "default" "sealed-secret" "test" "unsealed_secret"
+	local CC_KBS_ADDR
+	export CC_KBS_ADDR=$(kbs_k8s_svc_http_addr)
+	kernel_params_annotation="io.katacontainers.config.hypervisor.kernel_params"
+	kernel_params_value="agent.guest_components_procs=confidential-data-hub"
```
stevenhorsman, line 37: Completely optional idea: Should we consider adding `agent.guest_components_rest_api=resource` annotation here? It's currently default, so not strictly required, but I guess the same is true for `guest_components_procs` defaulting to ASR that includes CDH?

tests/integration/kubernetes/k8s-sealed-secret.bats
stevenhorsman: I like the idea of checking this mixed scenario. I think a new test might be clearer

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,98 @@
...
+	# Retrieve pod name, wait for it to come up, retrieve pod ip
+	export pod_name=$(kubectl get pod -o wide | grep "secret-test-pod-cc" | awk '{print $1;}')
+
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout="$timeout" pod "${pod_name}"
+
+	kubectl logs secret-test-pod-cc
+	kubectl logs secret-test-pod-cc | grep -q "unsealed environment as expected"
+}
+
```
stevenhorsman, line 81: In the issues I had the acceptance criteria for a negative test:

> _**Scenario: Sealed secrets envar fails with incorrect key**_
> **Given** I have a version of kata deployed that is configured to have a decryption key in its KBS under the `"default/sealed-secret/test` resource
>   **And** I've created a kubernetes secret called `sealed secret` from a secret that has been 'sealed' with a different key
> **When** I try and create a pod/job from a config with:
> ```
> env:
>     - name: SECRET_PASSWORD
>       valueFrom:
>         secretKeyRef:
>           name: sealed-secret
>           key: password
> ```
> that prints the value of the `SECRET_PASSWORD` envar
> **Then** We *can't* see the unsealed secret in `kubectl logs` and ideally see some sort of error in the logs to indicate that that the secret wasn't valid

Do you think that is fair to add?

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,98 @@
...
+	# "type": "vault",
+	# "name": "kbs:///default/sealed-secret/test",
+	# "provider": "kbs",
+	# "provider_settings": {},
+	# "annotations": {}
+	# }
+	kubectl create secret generic sealed-secret --from-literal='password=sealed.fakejwsheader.ewogICAgInZlcnNpb24iOiAiMC4xLjAiLAogICAgInR5cGUiOiAidmF1bHQiLAogICAgIm5hbWUiOiAia2JzOi8vL2RlZmF1bHQvc2VhbGVkLXNlY3JldC90ZXN0IiwKICAgICJwcm92aWRlciI6ICJrYnMiLAogICAgInByb3ZpZGVyX3NldHRpbmdzIjoge30sCiAgICAiYW5ub3RhdGlvbnMiOiB7fQp9Cg==.fakesignature'
+}
+
+@test "Unseal Secrets with CDH" {
```
stevenhorsman, line 66: Should we specify that it's env, not file specific for clarity when review test results?

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,98 @@
...
+		kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"
+	fi
+	set_metadata_annotation "${K8S_TEST_YAML}" \
+		"${kernel_params_annotation}" \
+		"${kernel_params_value}"
+
+	# Setup k8s secret
+	kubectl delete secret sealed-secret --ignore-not-found
+
+	# Sealed secret format is defined at: https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/docs/SEALED_SECRET.md#vault
```
stevenhorsman, line 50: I think it might be nice for test clarity and customisability (or a real word that means that), to have the secret generated in the test code rather than hard coded, but I'm fine for that to be part of a later PR.

tests/integration/kubernetes/runtimeclass_workloads/pod-sealed-secret.yaml
```diff
@@ -0,0 +1,32 @@
...
+    image: quay.io/prometheus/busybox:latest
+    imagePullPolicy: Always
+    command:
+    - sh
+    - -c
+    - |
+      env
+      if [ "$SECRET_PASSWORD" == "unsealed_secret" ]; then
+        echo "unsealed environment as expected"
+      fi
```
stevenhorsman, line 22: Idea: I'm wondering whether having something like `echo "Unsealed env secret: $SECRET_PASSWORD"` and then including the secret content in the grep check in the test might be a bit more customisable?

stevenhorsman (commented):
A few comments - generally I think it looks great. Maybe updating the PR name to be clearer that it's just sealed secrets via envs, not files would be helpful?

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: The ASR isn't needed for sealed secrets hence `guest_components_procs` being set to `confidential-data-hub` and the rest api option not being set at all.

stevenhorsman: ah ok, so you are deliberately not running ASR for this testing 👍

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: I've added a negative test. It tries to unseal the secret without ever having provisioned the key to the KBS. Hopefully it (won't) work.

tests/integration/kubernetes/runtimeclass_workloads/pod-sealed-secret.yaml
fitzthum: Good idea. Ideally we could generate a random key each time and then package it into a secret. I haven't gone that far, but I've move the check from the yaml file to the bash script.

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: I ended up adding a not-sealed secret to the existing tests to avoid making a second yaml file.

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: I was thinking we would enable the TEEs separately. in theory we should run this test on all the confidential runtime classes as long as KBS is true (because it depends on attestation), but seeing the current state of the CI I'm not sure we're ready yet.

portersrc (approved):

src/agent/src/cdh.rs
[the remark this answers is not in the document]
fitzthum: Yeah cdh.rs currently contains the direct cdh code (which is very minimal) and also helper functions for some of the operations that use the cdh. I think this is reasonable but I am open to either renaming the file to be more clear or moving some of the helper code somewhere else if anyone has any idea where would be good.

stevenhorsman (approved):
I think all my suggestions are either for future PRs, or have been address here, so this looks good to me. Thanks!

src/agent/src/cdh.rs
```diff
@@ -0,0 +1,61 @@
...
+            .unseal_secret(ttrpc::context::with_timeout(CDH_UNSEAL_TIMEOUT), &input)
+            .await?;
+        Ok(unsealed_secret.plaintext)
+    }
+
+    pub async fn unseal_env(&self, env: &str) -> Result<String> {
+        let (key, value) = env.split_once('=').unwrap();
+        if value.starts_with(SEALED_SECRET_PREFIX) {
+            let unsealed_value = self.unseal_secret_async(value).await?;
+            let unsealed_env = format!("{}={}", key, std::str::from_utf8(&unsealed_value)?);
```
wainersm, line 55: @fitzthum Does k8s spec envvar secrets' value to be UTF-8? or it's our convention?

src/agent/src/cdh.rs
```diff
@@ -49,13 +49,101 @@ impl CDHClient {
     }
 
     pub async fn unseal_env(&self, env: &str) -> Result<String> {
-        let (key, value) = env.split_once('=').unwrap();
+        let (key, value) = env.split_once('=').unwrap_or(("", ""));
```
wainersm, line 52: I think it's safer to catch the error, log a warning message and return. Otherwise it will silently change the envvars, can cause bugs hard to find.

tests/integration/kubernetes/k8s-sealed-secret.bats
wainersm: Good idea.

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,127 @@
...
+	# "name": "kbs:///default/sealed-secret/test",
+	# "provider": "kbs",
+	# "provider_settings": {},
+	# "annotations": {}
+	# }
+	kubectl create secret generic sealed-secret --from-literal='secret=sealed.fakejwsheader.ewogICAgInZlcnNpb24iOiAiMC4xLjAiLAogICAgInR5cGUiOiAidmF1bHQiLAogICAgIm5hbWUiOiAia2JzOi8vL2RlZmF1bHQvc2VhbGVkLXNlY3JldC90ZXN0IiwKICAgICJwcm92aWRlciI6ICJrYnMiLAogICAgInByb3ZpZGVyX3NldHRpbmdzIjoge30sCiAgICAiYW5ub3RhdGlvbnMiOiB7fQp9Cg==.fakesignature'
+
+	kubectl create secret generic not-sealed-secret --from-literal='secret=not_sealed_secret'
+
+	kbs_set_allow_all_resources
```
wainersm, line 67: Let's not forget:
```
if ! is_confidential_hardware; then
        kbs_set_allow_all_resources
fi
```

src/agent/src/cdh.rs
fitzthum: Great question. It's our convention. k8s just sees the secret as a string, but we know that it's actually a wrapped secret. utf-8 tells us how to get back to a string after we decrypt the secret. tbh we haven't documented our convention very well. In theory this should be handled by the [secret-cli](https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/secret/src/bin/secret_cli.rs#L64) but we don't support vault secrets there yet. I will probably add that support soon. It also might be worth reworking the API of the CDH to return a string rather than bytes, but that is a different discussion.

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: ohhh good point. i will add that

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,127 @@
...
+}
+
+@test "Cannot Unseal Env Secrets with CDH without key" {
+	[ "${KATA_HYPERVISOR}" = "qemu-coco-dev" ] || skip "Test not ready yet for ${KATA_HYPERVISOR}"
+
+	kubectl apply -f "${K8S_TEST_YAML}"
```
wainersm, line 73: [k8s_create_pod()](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/lib.sh#L41) will save 3 lines and it will print debugging info if the pod fails to create.

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,127 @@
...
+	# Check pod creation
+	kubectl wait --for=condition=Ready --timeout="$timeout" pod "${pod_name}"
+
+	kubectl logs secret-test-pod-cc
+	kubectl logs secret-test-pod-cc | grep -q "UNPROTECTED_SECRET = not_sealed_secret"
+	cmd="kubectl logs secret-test-pod-cc | grep -q \"PROTECTED_SECRET = unsealed_secret\""
+	run $cmd
+	[ "$status" -eq 1 ]
+
+	kubectl delete -f "${K8S_TEST_YAML}"
```
wainersm, line 87: Delete it on teardown()

tests/integration/kubernetes/k8s-sealed-secret.bats
```diff
@@ -0,0 +1,127 @@
...
+	[ "${KATA_HYPERVISOR}" = "qemu-coco-dev" ] || skip "Test not ready yet for ${KATA_HYPERVISOR}"
+
+	if [ "${KBS}" = "false" ]; then
+		skip "Test skipped as KBS not setup"
+	fi
+
+	[ -n "${pod_name:-}" ] && kubectl describe "pod/${pod_name}" || true
+	[ -n "${pod_config_dir:-}" ] && kubectl delete -f "${K8S_TEST_YAML}" || true
+
+	if [ -n "${node_start_time}:-}" ]; then
```
wainersm, line 120: While ago I added a new conditional to print the logs only when the test failed to avoid a exceeded verbose output -> https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-confidential-attestation.bats#L97

src/agent/src/cdh.rs
wainersm: Got it. Thanks for the explanation!

wainersm:
Hi @fitzthum !

These are the last comment I had for this PR.

src/agent/src/cdh.rs
fitzthum: If the secret isn't of the form `key=value`, then the next conditional where we check for the secret prefix will fail and the secret won't be modified. The only case we should be modifying the env is if it is a sealed secret. I'm not sure if it's more clear to return immediately in that case or fall through to the end of the function.

tests/integration/kubernetes/k8s-sealed-secret.bats
fitzthum: The next test uses the same pod.

fitzthum: wait nvm i forgot how bats works

src/agent/src/cdh.rs
wainersm: oh, never mind, for some reason I thought that `*env.to_owned()).to_string()` was actually returning (the modified) `key=value`.

Return immediately will save one conditional.Maybe the compiler will optimize that....

fitzthum: I changed this to be an `if let`, which seems a lot clearer.

src/agent/src/cdh.rs
fitzthum: btw this is why this PR has the secret hard-coded. once we have the secret-cli updated, I will use that in the test instead.

wainersm: Cool!

wainersm (approved):
Hey @fitzthum ! Thanks for caring about this port to main!

Looks good. Hopefully it will pass on qemu-coco-dev CI.

ChengyuZhu6:
Tests have passed on tdx: https://github.com/kata-containers/kata-containers/actions/runs/9847939563/job/27200021621?pr=9719

ChengyuZhu6 (approved):
LGTM Thanks, @fitzthum

PR #9722, opened 2024-05-27
deploy: Add busybox target

zvonkok:
For a minimal initrd/image build we may want to leverage busybox. This is part number two of the NVIDIA initrd/image build


fidencio:
@zvonkok, could you add more information on how busybox is needed / used in your setup?

It's not totally clear to me whether you need busybox on the host, or inside the guest image, and if it's on the host side, what's the problem on relying on the busybox from the distro.

zvonkok:
@fidencio It is for the guest, I am running the bare minimum. busybox is compiled only with 5 commands that I need for the NVIDIA GPU initrd/image initialization and management. 
Additionally busybox ships a ntpd version where I can disable the server part, meaning no sockets are enabled and the ntpd does  not neet to inspect any messages before dropping them. A GPU workload would never act as a ntpd server.

zvonkok:
Some initial tests have shown that I can run a full blown GPU workload with an initrd with ~40-50MB size.

fidencio:
> Some initial tests have shown that I can run a full blown GPU workload/container with an initrd with ~40-50MB size. Additionally also getting rid of bash and only using POSIX shell (ash) to even more reduce any attack vectors.

This can be very useful for the non-TEE / non-confidential / non-GPU use case.

zvonkok:
Indeed for the CPU only use-case you can essentially go down to kata-agent + some helper libs and tools ~30MB.

zvonkok:
@fidencio I will add examples later on for CPU only, and GPU initrd/rootfs builds that showcase the minimal set.

tools/packaging/static-build/busybox/build-static-busybox.sh
```diff
@@ -0,0 +1,58 @@
...
+
+	curl -O "${URL_TARBZ2}"
+	curl -O "${URL_SHA}"
+	curl -O "${URL_SIG}"
+
+	echo "Verifying SHA256 checksum..."
+	sha256_file="$(basename "${URL_SHA}")"
+	sha256sum -c "${sha256_file}" 
+
+	#gpg --keyserver keys.openpgp.org --recv-keys C9E9416F76E610DBD09D040F47B70C55ACC9965B
```
zvonkok, line 34: Oops need to add that back in :)

zvonkok: Fixed

stevenhorsman:
Hey Zvonko, sorry for being slow on this. Just to check that I'm understanding correctly:
- This PR is to enable the option to build the busybox binary into the guest. You want this to create a small environment that can run workloads with the extra busybox commands? 
- How does this interact with us building the rootfs on top of ubuntu at the moment, does that not provide the commands that are in busybox?
- What is the next steps for post this PR - are you going to create a new variant of rootfs that installs busybox, or want it integrated to all rootfs's?

Sorry if I'm missing something obvious!

zvonkok:
@stevenhorsman Take a container multistage build as the mental model. 
In stage one, Ubuntu/Fedora/Alpine is used to build all the needed artifacts (kernel modules, helper tools, etc), which have many dependencies and need to be installed inside the rootfs for proper building. 

In stage two (the actual runtime) does not need all the build dependencies and we can build a rootfs that is minimal. Technically we do not even need a shell if `kata-agent` is the init process. 

I can create the PR for the GPU rootfs if this helps and we can use this as reference and add all the needed pieces step by step.

zvonkok:
The idea is to add a multi-stage rootfs build.

stevenhorsman:
> @stevenhorsman Take a container multistage build as the mental model.

Cool that helps and I think I can see more where this is going now.

stevenhorsman (approved):
This looks fine to me as a step towards a goal. Hopefully there aren't any issues, but we will test it fully once the workflows are merged 😢

wainersm:
Hi @zvonkok !

I'm unable to build busybox-tarball. What I'm doing wrong?

```
$ make busybox-tarball
make busybox-tarball-build
make[1]: Entering directory '/home/wmoschet/src/github.com/kata-containers/kata-containers'
/home/wmoschet/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build//kata-deploy-binaries-in-docker.sh  --build=busybox
sha256:d2a93ffb9e4ac781f3738194fdc60e34223ce839e54d696025eb2c61d179c317
Build kata version 3.5.0: busybox
INFO: DESTDIR /home/wmoschet/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/busybox/destdir
Error: 1:1: invalid input text "externals.busybo..."
Error response from registry: failed to resolve latest-main-x86_64: GET "https://ghcr.io/v2/kata-containers/cached-artefacts/busybox/manifests/latest-main-x86_64": denied: requested access to the resource is denied
INFO: build static busybox
/home/wmoschet/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh: line 817: BUSYBOX_CONF_FILE: parameter null or not set
make[1]: *** [tools/packaging/kata-deploy/local-build/Makefile:75: busybox-tarball-build] Error 1
make[1]: Leaving directory '/home/wmoschet/src/github.com/kata-containers/kata-containers'
make: *** [tools/packaging/kata-deploy/local-build/Makefile:84: busybox-tarball] Error 2
```

`Error: 1:1: invalid input text "externals.busybo..."`  --- caught my attention.

zvonkok:
@wainersm You need to provide a BUSYBOX_CONF_FILE otherwise we do not now what to build. I do not know if it make sense to add a default config but there is one that I need. 

```
 BUSYBOX_CONF_FILE="busybox.nvidia.conf" make busybox-tarball
```

The other error is from the updated `yq` pushing a new version....

fidencio (approved):

PR #9727, opened 2024-05-28
ci: ovmf without sudo

zvonkok:
Build ovmf without sudo docker this is not needed. This is part 1 of N

### DIFF
--- a/tools/packaging/static-build/ovmf/build.sh
+++ b/tools/packaging/static-build/ovmf/build.sh
@@ -47,12 +47,12 @@ fi
 [ -n "$ovmf_package" ] || die "failed to get ovmf package or commit"
 [ -n "$package_output_dir" ] || die "failed to get ovmf package or commit"
 
-sudo docker pull ${container_image} || \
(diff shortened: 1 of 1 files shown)


zvonkok:
```
❯ ls -l build/ovmf/{builddir,destdir}
build/ovmf/builddir:
total 1880
-rw-r--r-- 1 zvonkok zvonkok 962132 May 28 08:21 edk2-x86_64.tar.gz
-rw-r--r-- 1 root    root    955512 May 28 08:19 kata-static-ovmf.tar.xz
-rw-r--r-- 1 root    root        90 May 28 08:19 ovmf-sha256sum

build/ovmf/destdir:
total 4
drwxr-xr-x 3 zvonkok zvonkok 4096 May 28 08:21 opt
```

The final tarball is not owned by the user since we're doing a `sudo tar` in `kata-deploy-binaries.sh`. Once we have all parts build without sudo the final tarball should be owned by the user as well.

fidencio (approved):
lgtm, thanks @zvonkok!

BbolroC (approved):
LGTM, thanks @zvonkok !

PR #9728, opened 2024-05-28
ci: guest-components without sudo

zvonkok:
Build guest-components without sudo docker this is not needed. This is part 2 of N

### DIFF
--- a/tools/packaging/static-build/coco-guest-components/Dockerfile
+++ b/tools/packaging/static-build/coco-guest-components/Dockerfile
@@ -8,8 +8,15 @@ ARG RUST_TOOLCHAIN
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+ENV RUSTUP_HOME="/opt/rustup"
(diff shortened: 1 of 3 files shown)


zvonkok:
```
❯ ls -l build/coco-guest-components/{builddir,destdir} 
build/coco-guest-components/builddir:
total 6388
-rw-r--r--  1 root    root        107 May 28 08:56 coco-guest-components-sha256sum
drwxr-xr-x 10 zvonkok zvonkok    4096 May 28 08:56 guest-components
-rw-r--r--  1 root    root    6533060 May 28 08:56 kata-static-coco-guest-components.tar.xz
```
The final tarball and hashsum is not owned by the user since we're doing it with `sudo` in kata-deploy-binaries.sh. Once we have all parts build without sudo the final tarball should be owned by the user as well.

fidencio (approved):
lgtm, thanks @zvonkok!

BbolroC (approved):
LGTM, thanks @zvonkok !

PR #9731, opened 2024-05-28
ci: pause-image no sudo

zvonkok:
Build pause-image without sudo docker this is not needed. This is part 5 of N

### DIFF
--- a/tools/packaging/static-build/pause-image/build-static-pause-image.sh
+++ b/tools/packaging/static-build/pause-image/build-static-pause-image.sh
@@ -19,8 +19,8 @@ pull_pause_image_from_remote() {
 	echo "pull pause image from remote"
 
 	skopeo copy "${pause_image_repo}":"${pause_image_version}" oci:pause:"${pause_image_version}"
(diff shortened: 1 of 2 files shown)


zvonkok:
```
❯ ls -l build/pause-image/{builddir,destdir}
build/pause-image/builddir:
total 252
-rw-r--r-- 1 root    root    247596 May 28 11:31 kata-static-pause-image.tar.xz
drwxr-xr-x 3 zvonkok zvonkok   4096 May 28 11:31 pause
-rw-r--r-- 1 root    root        97 May 28 11:31 pause-image-sha256sum

build/pause-image/destdir:
total 4
drwx------ 3 zvonkok zvonkok 4096 May 28 11:31 pause_bundle
```

The final tarball and hashsum is not owned by the user since we're doing it with sudo in kata-deploy-binaries.sh. Once we have all parts build without sudo the final tarball and hashsum should be owned by the user as well.

fidencio (approved):
lgtm, thanks @zvonkok!

GabyCT (approved):

PR #9736, opened 2024-05-28
ci: qemu no sudo

zvonkok:
Build qemu without sudo docker this is not needed. This is part 9 of N


zvonkok:
```
 ❯ ls -l build/qemu/{builddir,destdir}
build/qemu/builddir:
total 62056
-rw-r--r-- 1 root root 37437160 May 28 16:08 kata-static-qemu.tar.gz
-rw-r--r-- 1 root root 26099944 May 28 16:06 kata-static-qemu.tar.xz
-rw-r--r-- 1 root root       90 May 28 16:06 qemu-sha256sum

build/qemu/destdir:
total 4
drwxr-xr-x 3 zvonkok zvonkok 4096 May 28 16:08 opt
```
The final tarball is not owned by the user since we're doing a sudo tar in kata-deploy-binaries.sh. Once we have all parts build without sudo the final tarball should be owned by the user as well.

fidencio:
lgtm, thanks @zvonkok!

GabyCT (approved):

PR #9737, opened 2024-05-28
ci: kata-deploy no sudo

zvonkok:
build/push/manage aritfacts without sudo

### DIFF
--- a/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
+++ b/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
@@ -196,7 +196,7 @@ install_cached_tarball_component() {
 	# "tarball1_name:tarball1_path tarball2_name:tarball2_path ... tarballN_name:tarballN_path"
 	local extra_tarballs="${6:-}"
 
-	sudo oras pull ${ARTEFACT_REGISTRY}/${ARTEFACT_REPOSITORY}/cached-artefacts/${build_target}:latest-${TARGET_BRANCH}-$(uname -m) || return 1
+	oras pull ${ARTEFACT_REGISTRY}/${ARTEFACT_REPOSITORY}/cached-artefacts/${build_target}:latest-${TARGET_BRANCH}-$(uname -m) || return 1
 
 	cached_version="$(cat ${component}-version)"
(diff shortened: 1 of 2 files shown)


zvonkok:
The `merge-builds` target will set all the files to the right owner for installation. 
```
echo "create ${tar_path}"
(cd "${tarball_content_dir}"; tar cvfJ "${tar_path}" --owner=0 --group=0 .)
popd
```

fidencio (approved):
lgtm, thanks @zvonkok!

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1211,8 +1209,6 @@ silent_mode_error_trap() {
 }
 
 main() {
-	git config --global --add safe.directory ${repo_root_dir}
```
gkurz, line 1214: There are other places that do that, e.g. `get_last_modification()` at https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/scripts/lib.sh#L111 that has a lot of callers in this script.

Please address the removal of this workaround globally in a distinct PR.

zvonkok: Right, AFAIR we have them also  sprinkled in the gh actions as well.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh
```diff
@@ -1297,5 +1293,5 @@ main() {
 }
 
 if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
-	main $@
+	main "$@"
```
gkurz, line 1296: This fix is theoretically correct in case some argument to the script contains a space. I'm not sure this happens in practice. Did you see an actual error ?

Also this script gets its arguments from `kata-deploy-binaries-in-docker.sh` which has the same _issue_ :

https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh#L139

Please fix this globally in a separate PR.

gkurz (approved):
LGTM. Thanks @zvonkok !

PR #9738, opened 2024-05-29
agent: Make /dev/sev-guest available to containers

aep:
This makes it so that any container has access to /dev/sev-guest.

tested on 6.9.0-rc7-snp-host-05b10142ac6a

### DIFF
--- a/src/agent/rustjail/src/container.rs
+++ b/src/agent/rustjail/src/container.rs
@@ -151,7 +151,7 @@ lazy_static! {
     };
 
     pub static ref DEFAULT_DEVICES: Vec<LinuxDevice> = {
-        vec![
+        let mut devices = vec![
             LinuxDevice {
                 path: "/dev/null".to_string(),
                 r#type: "c".to_string(),
@@ -206,7 +206,23 @@ lazy_static! {
                 uid: Some(0xffffffff),
                 gid: Some(0xffffffff),
             },
-        ]
+        ];
+        
+        let sev_guest_path = "/dev/sev-guest";
+        if let Ok(sev_guest_attr) = fs::metadata(sev_guest_path) {
+            let sev_guest_devid = sev_guest_attr.rdev();
+            devices.push(LinuxDevice {
+                path: sev_guest_path.to_string(),
+                r#type: "c".to_string(),
+                major: stat::major(sev_guest_devid) as i64,
+                minor: stat::minor(sev_guest_devid) as i64,
+                file_mode: Some(0o666),
+                uid: Some(sev_guest_attr.uid()),
+                gid: Some(sev_guest_attr.gid()),
+            });
+        };
+
+        devices
     };
 
     pub static ref SYSTEMD_CGROUP_PATH_FORMAT:Regex = Regex::new(r"^[\w\-.]*:[\w\-.]*:[\w\-.]*$").unwrap();


zvonkok:
What is the use case? 
Why would we expose this per default if the device is available? 

@fitzthum FYI

aep:
we're required to run attestation inside the pod.
without this change you cannot actually access /dev/dev-guest inside the pod at all.

since you get permission denied even with privileged. i think due to privileged_without_devices causing host bind mounts to actually not be to the real devices.

i believe its safe to expose /dev/dev-guest by default. It's the default in confidential guest vms anyway

danmihai1:
> should only be used in certain configurations where the measurement of the containers is bound to the hw measurement

I agree with @fitzthum that binding the integrity of containers to the TEE attestation evidence is needed when performing attestation from a container. I believe the Kata main branch didn't implement yet this binding, so this guest device change alone is indeed insufficient.


> It is also subpar because it introduces a platform-specific API into a container

But the alternative is subpar too because it gives up on the advantages why many applications are containerized nowadays.


> What is the use case?

@zvonkok , I described some of these use cases in https://github.com/confidential-containers/confidential-containers/issues/126#issuecomment-1644402591

zvonkok:
@danmihai1 My point is, why are we doing this per default if it is available?

-  some customers want to do attestation from the container, 
- some customers rely on Trustee and use higher level constructs than containers
- some customers want a hybrid mode where they want e.g. attestation but also an attestation token to do a secret release from a different KBS (multiple personas per TVM)

I also remember some discussions around workloads attacking the KBS if they have access to the attestation report ... 

Independent of the use case, we shouldn't expose this in the container per default. Just being there does not mean we need it in every container.

danmihai1:
> @danmihai1 My point is, why are we doing this per default if it is available?
> 
> * some customers want to do attestation from the container,
> * some customers rely on Trustee and use higher level constructs than containers
> * some customers want a hybrid mode where they want e.g. attestation but also do a secret release from a different KBS (multiple personas per TVM)
> 
> I also remember some discussions around workloads attacking the KBS if they have access to the attestation report ...
> 
> Independent of the use case, we shouldn't expose this in the container per default. Just being there does not mean we need it in every container.
> 
> We should at least make this configurable

Ah, I get your point now that exposing this device to containers doesn't have to take place *by default*. Thanks for explaining that, @zvonkok!

pegahnikbakht:
Hi,

I have setup a SNP based kata container and I want to attest the container using snpguest tool, I get the following error:
`ERROR: unable to open /dev/sev
because: Operation not permitted (os error 1)
Error: unable to open /dev/sev`
I tried to configure /dev/sev-guest as follows:
`/ # SNP_MAJOR=$(cat /sys/devices/virtual/misc/sev-guest/dev | awk -F: '{print $1}')
/ # SNP_MINOR=$(cat /sys/devices/virtual/misc/sev-guest/dev | awk -F: '{print $2}')
/ # mknod -m 600 /dev/sev-guest c "${SNP_MAJOR}" "${SNP_MINOR}"
`
But still I get the same error, any idea on how to add /dev/sev inside the container? and obtain an attestation report with snpguest? 
In the container I have this kernel 6.7.0.

fitzthum:
@pegahnikbakht

Kata (without CoCo) supports starting confidential guests, but if you want to do more advanced things (like getting the attestation report), you should think about using CoCo. As mentioned, it's dangerous to expose attestation reports directly to Kata guests.

pegahnikbakht:
@fitzthum Ok still if I want to do the attestation how I can solve operation not permitted issue with /dev/sev-guest?

pegahnikbakht:
@ryansavino @fitzthum 
Thanks for the help, I manged to get the report from an AMD SEV-SNP container. 
Tobin can you elaborate more what is the security issue of attesting the container directly (and what is the threat model you assume)? 
Is it only about “attestation factory attack” (when a container can produce arbitrary reports that other malicious containers can use), or anything else? 
Would that be still “not secure” if there is an unusual use case with one hardware platform with 1 vm and with 1 kata container?

fitzthum:
> can you elaborate more what is the security issue of attesting the container directly

In this case you aren't attesting the container at all. Due partly to implementation gaps and partly to design choices, the hardware measurement does not cover the container itself. Generally you should use signed images in conjunction with remote attestation and the CDH. 

In your case you are subject to evidence factory attacks in this case, which shouldn't be underestimated. The issue is that the evidence generated by your guest will look the same as the evidence generated by any other guest with the same basic configuration. Any secrets or privileged operations that are gated on this evidence, will be accessible to anyone else who has the same base configuration. 

This is true basically regardless of your deployment model. Perhaps if you only have one machine, you could make your verifier use the VCEK of the guest to make sure that the evidence comes exclusively from that machine, but this wouldn't support any kind of multi-tenancy securely.

katexochen:
> the hardware measurement does not cover the container itself.

@fitzthum with the execution policy feature in place, shouldn't this make the container image part of the attestation report and mitigate this kind of attack?

fitzthum:
> with the execution policy feature in place, shouldn't this make the container image part of the attestation report and mitigate this kind of attack?

Currently I don't think the policy is measured. Once that gap is closed, that will be a secure option, but it is just an option. 
For users that measure their workloads via signatures, the hw measurement won't cover the workload. That is a feature rather than a bug. Either way, exposing the hw evidence must be optional and enabling it must change the launch measurement. This is exactly what we have implemented already via the ASR.

fidencio:
> Kata (without CoCo) supports starting confidential guests, but if you want to do more advanced things (like getting the attestation report), you should think about using CoCo. As mentioned, it's dangerous to expose attestation reports directly to Kata workloads.

@fitzthum , what's the difference on exposing the device and exposing the rest API which also will generate an evidence for the workload?

Also, what do you mean by Kata and CoCo here? CoCo deploys Kata Containers with TEE flavours, exactly the same payload.

fitzthum:
> what's the difference on exposing the device and exposing the rest API which also will generate an evidence for the workload?

Turning on the API changes the measurement. It also exposes a higher-level API that abstracts away the platform details, although this is becoming a bit less of a problem with the new configfs support.

> Also, what do you mean by Kata and CoCo here? CoCo deploys Kata Containers with TEE flavours, exactly the same payload.

When I'm talking about plain kata, I mean using the standard runtime classes (i.e. `kata-qemu`) and then turning on some confidential features. I think this is what some of the people in this thread were doing.

mythi:
(I found this while searching for something cdh related)

I think this one can be closed. At the time we think the use-case can be enabled, it'd follow #9554 which is the vendor agnostic way of doing this. The vendor specific ioctl's are "lazily deprecated" anyways.

PR #9747, opened 2024-05-30
ci: Remove all git config safe.directory

zvonkok:
Now with the sudo less build we should be good
to remove those hacks.


dborquez (approved):
lgtm, thank you @zvonkok

gkurz (approved):
LGTM. Thanks @zvonkok !

PR #9749, opened 2024-05-30
CoCo: introduce config for guest-components procs

### LINKED ISSUE #9748: CoCo guest-components process spawning behaviour should be configurable

At the moment the agent actively spawns three CoCo-related child processes, depending on a mix of heuristics (are binaries present in the rootfs) and a configuration option (is `guest_components_rest_api=none`).

In certain situation this leads to problems, because we want to specify custom configuration flags or launch said processes in certain environments (network namespace) and we can only do this by changing agent source code.

Example 1: in peerpod, container sandboxes are created in a dedicated network namespace, isolated from the rest of the guest fs, the `api-server-rest` process needs to be spawned in the same ns to expose its http api to a pod workload. Peerpod guest vm's leverage systemd to manage the lifecycle of agent + guest-components. At the moment there are 2 processes of `api-server-rest` (1x root ns, 1x pod ns) which is unfortunate.

Example 2: In some cases binaries might be present on the rootfs, but we do explicitly want to avoid the agent from spawning those processes.

**How can it be improved?**

We should have an option to toggle the spawning of CoCo guest-components, by the agent. As there is a dependency relationship this can be flat configuration option `guest_components_procs=none|attestation-agent|confidential-data-hub|api-server-rest`. the latter two options would imply the previous option.

mkulke:
fixes #9748

A configuration option `guest_component_procs` has been introduced that indicates which guest component processes are supposed to be spawned by the agent. The default behaviour remains that all of those processes are actively spawned by the agent. At the moment this is based on presence of binaries in the rootfs and the guest_component_api_rest option.

The new option is incremental:

none -> attestation-agent -> confidential-data-hub -> api-server-rest

e.g. api-server-rest implies attestation-agent and confidential-data-hub

the `none` option has been removed from guest_component_api_rest, since this is addresses by the introduced option.

### DIFF
--- a/src/agent/README.md
+++ b/src/agent/README.md
@@ -126,7 +126,8 @@ The kata agent has the ability to configure agent options in guest kernel comman
 | `agent.debug_console_vport` | Debug console port | Allow to specify the `vsock` port to connect the debugging console | integer | `0` |
 | `agent.devmode` | Developer mode | Allow the agent process to coredump | boolean | `false` |
 | `agent.hotplug_timeout` | Hotplug timeout | Allow to configure hotplug timeout(seconds) of block devices | integer | `3` |
-| `agent.guest_components_rest_api` | `api-server-rest` configuration | Select the features that the API Server Rest attestation component will run with. Valid values are `all`, `attestation`, `resource`, or `none` to not launch the `api-server-rest` component | string | `resource` |
+| `agent.guest_components_rest_api` | `api-server-rest` configuration | Select the features that the API Server Rest attestation component will run with. Valid values are `all`, `attestation`, `resource` | string | `resource` |
+| `agent.guest_components_procs` | guest-components processes | Attestation-related processes that should be spawned as children of the guest. Valid values are `none`, `attestation-agent`, `confidential-data-hub` (implies `attestation-agent`), `api-server-rest` (implies `attestation-agent` and `confidential-data-hub`) | string | `api-server-rest` |
 | `agent.https_proxy` | HTTPS proxy | Allow to configure `https_proxy` in the guest | string | `""` |
 | `agent.log` | Log level | Allow the agent log level to be changed (produces more or less output) | string | `"info"` |
 | `agent.log_vport` | Log port | Allow to specify the `vsock` port to read logs | integer | `0` |
--- a/src/agent/src/main.rs
+++ b/src/agent/src/main.rs
@@ -59,11 +59,11 @@ mod util;
 mod version;
 mod watcher;
 
-use config::GuestComponentsFeatures;
+use config::GuestComponentsProcs;
 use mount::{cgroups_mount, general_mount};
 use sandbox::Sandbox;
 use signal::setup_signal_handler;
-use slog::{error, info, o, warn, Logger};
+use slog::{debug, error, info, o, warn, Logger};
 use uevent::watch_uevents;
 
 use futures::future::join_all;
@@ -403,8 +403,16 @@ async fn start_sandbox(
     let (tx, rx) = tokio::sync::oneshot::channel();
     sandbox.lock().await.sender = Some(tx);
 
-    if Path::new(CDH_PATH).exists() && Path::new(AA_PATH).exists() {
-        init_attestation_components(logger, config)?;
+    let gc_procs = config.guest_components_procs;
+    if gc_procs != GuestComponentsProcs::None {
+        if !attestation_binaries_available(logger, &gc_procs) {
+            warn!(
+                logger,
+                "attestation binaries requested for launch not available"
+            );
+        } else {
+            init_attestation_components(logger, config)?;
+        }
     }
 
     // vsock:///dev/vsock, port
@@ -417,9 +425,33 @@ async fn start_sandbox(
     Ok(())
 }
 
+// Check if required attestation binaries are available on the rootfs.
+fn attestation_binaries_available(logger: &Logger, procs: &GuestComponentsProcs) -> bool {
+    let binaries = match procs {
+        GuestComponentsProcs::AttestationAgent => vec![AA_PATH],
+        GuestComponentsProcs::ConfidentialDataHub => vec![AA_PATH, CDH_PATH],
+        GuestComponentsProcs::ApiServerRest => vec![AA_PATH, CDH_PATH, API_SERVER_PATH],
+        _ => vec![],
+    };
+    for binary in binaries.iter() {
+        if !Path::new(binary).exists() {
+            warn!(logger, "{} not found", binary);
+            return false;
+        }
+    }
+    true
+}
+
 // Start-up attestation-agent, CDH and api-server-rest if they are packaged in the rootfs
-fn init_attestation_components(logger: &Logger, _config: &AgentConfig) -> Result<()> {
-    // The Attestation Agent will run for the duration of the guest.
+// and the corresponding procs are enabled in the agent configuration. the process will be
+// launched in the background and the function will return immediately.
+fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<()> {
+    // skip launch of any guest-component
+    if config.guest_components_procs == GuestComponentsProcs::None {
+        return Ok(());
+    }
+
+    debug!(logger, "spawning attestation-agent process {}", AA_PATH);
     launch_process(
         logger,
         AA_PATH,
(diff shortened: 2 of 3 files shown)


src/agent/README.md
```diff
@@ -126,7 +126,8 @@ The kata agent has the ability to configure agent options in guest kernel comman
...
 | `agent.devmode` | Developer mode | Allow the agent process to coredump | boolean | `false` |
 | `agent.hotplug_timeout` | Hotplug timeout | Allow to configure hotplug timeout(seconds) of block devices | integer | `3` |
-| `agent.guest_components_rest_api` | `api-server-rest` configuration | Select the features that the API Server Rest attestation component will run with. Valid values are `all`, `attestation`, `resource`, or `none` to not launch the `api-server-rest` component | string | `resource` |
+| `agent.guest_components_rest_api` | `api-server-rest` configuration | Select the features that the API Server Rest attestation component will run with. Valid values are `all`, `attestation`, `resource` | string | `resource` |
+| `agent.guest_components_procs` | guest-components processes | Attestation-related processes that should be spawned as children of the guest. Valid values are `attestation-agent`, `confidential-data-hub` (implies `attestation-agent`), `api-server-rest` (implies `attestation-agent` and `confidential-data-hub`) | string | `api-server-rest` |
```
stevenhorsman, line 130: I think we can add `none` as a valid value here?

src/agent/src/main.rs
```diff
@@ -429,32 +436,43 @@ fn init_attestation_components(logger: &Logger, _config: &AgentConfig) -> Result
...
-                    &vec!["--features", &features.to_string()],
-                    "",
-                    0,
-                ) {
-                    error!(logger, "launch_process {} failed: {:?}", API_SERVER_PATH, e);
-                }
-            }
-        }
+    )
+    .map_err(|e| anyhow!("launch_process {} failed: {:?}", CDH_PATH, e))?;
```
stevenhorsman, line 455: The previous logic was that if cdh/asr failed to launch, we would log the error, but then continue, but now as the config is explicitly staring which of them to run we throw an error to better match the user's expectations?

src/agent/src/main.rs
```diff
@@ -418,8 +418,15 @@ async fn start_sandbox(
 }
 
 // Start-up attestation-agent, CDH and api-server-rest if they are packaged in the rootfs
-fn init_attestation_components(logger: &Logger, _config: &AgentConfig) -> Result<()> {
-    // The Attestation Agent will run for the duration of the guest.
+// and the corresponding procs are enabled in the agent configuration. the process will be
+// launched in the background and the function will return immediately.
+fn init_attestation_components(logger: &Logger, config: &AgentConfig) -> Result<()> {
```
stevenhorsman, line 423: I think we also need to review L406-408 (sorry - github still won't let us comment on unchanged lines) as that checks both CDH & AA paths exist before launching this code, whereas I guess now a user sets `agent.guest_components_procs=attestation-agent` then it would be valid to have a rootfs without CDH binary it and still launch it? I'm wondering if the best approach is to pull the path check(s) into this function?

stevenhorsman (commented):
General looks good, but I've left a couple of questions. Also if you can update the commit header subsystem to `agent:` rather than coco for consistency, that would be good. Thanks

src/agent/src/main.rs
mkulke: oh, right. that's indeed a change of behaviour. I'm not sure if this was intentional, because we return an err if we fail to launch the AA process. Is there a reason why we would deem AA to be critical but tolerate a CDH failure?

src/agent/src/main.rs
mkulke: good point, forgot those. curiously only the CDH and AA bins presence was checked, ASR isn't.

It's tricky, because I assume the intention behind this clause is to make sure attestation init is only called on CoCo guest vms. There doesn't seem to be a coco-specific cargo build feature, the build-time toggle is probably a make target like `rootfs-initrd-confidential-tarball` that will add a coco-layer on the rootfs, but the agent is generic.

So, to keep the behaviour as-is for non-coco guests, we probably need to keep a check like this, as we default to `_procs=api-server-rest`. It's not great because this check is redundant, but I wouldn't know a better way.

stevenhorsman: Yeah - I agree - In order to not have to have a separate agent build we effectively use that as a switch for confidential vs non-confidential VMs. Maybe we could just check the attestation-agent binary before calling init-attestation as that would allow the `agent.guest_components_procs=attestation-agent` to still work if only the attestation-agent is in the rootfs?

src/agent/src/main.rs
stevenhorsman: Yeah - I suspect that it's a bit of legacy build up in that the AA was there first and then CDH and ASR added. I think we had a discussion when we moved it over to `main` and the feeling was that the AA was a show stopper, but if the CDH didn't work then we could try and carry on. I'll see if I can find the thread and link it.

src/agent/src/main.rs
mkulke: I added a fn that checks for the presence of all the bins as a pre-requisite. is there a scenario in which we expect the AA bin to be present, but not CDH and ASR?

src/agent/src/main.rs
stevenhorsman: https://github.com/kata-containers/kata-containers/pull/8870#discussion_r1550045052 might have been what I was remembering, but doesn't seem like a particularly concrete decision, so I'm fine with your implementation.

mkulke: hmm, yeah, I guess you can still use AA standalone and it seems a bit odd to attempt but tolerate a failed startup (it helps with debugging, but it's also a source of confusion).

I understood that image-rs will rely on CDH for layer decryption eventually, but it doesn't hurt to keep the prior behaviour, i.e. only abort when AA launch fails, and tolerate the others. WDYT?

src/agent/src/main.rs
stevenhorsman: Good question - with our rootfs builds no, but as we provide the option `agent.guest_components_procs=attestation-agent` to only launch the attestation I'm guessing there is a scenario that makes sense to only use the attestation-agent (though this might change if we flip the image-rs guest pull to happen through CDH instead) and based on that a downstream rootfs builder might optimise to only build in the AA? However you know the attestation systems better than I do, so I won't block this PR based on this issue.

src/agent/src/main.rs
stevenhorsman: I guess my concern is that if a use specifies `agent.guest_components_procs=api-server-rest`, but then we don't launch CDH/ASR due to a failure in that launch, then I think the "obvious" user experience would be to fail, rather than log and proceed. The complication is that `agent.guest_components_procs=api-server-rest` is the default, so the user might not actually care about it in which case proceeding is fine! 😿

I guess if we might flip the CDH interaction then it's not super important to get this perfect now.

@fitzthum might be interested and have some thoughts on this and the other comment thread too?

src/agent/src/main.rs
mkulke: I see, I extended the verify logic to be a bit more forgiving and check only the bins that have been requested for launch

stevenhorsman: Ok, that is obviously (but not to me!) the best way to do this. Thanks!

src/agent/src/main.rs
mkulke: > I think the "obvious" user experience would be to fail, rather than log and proceed. The complication is that agent.guest_components_procs=api-server-rest is the default

Indeed, that's a bit messed up and hard to reconcile. I assume the intuitive green-field ux would be: A user requests the agent to launch a process (explicitly or implicity) and the agent itself fails hard if it cannot.

It is tough to debug in cases of failure, but I'm not sure this should be a guiding principle. We probably want to fail fast if anything attestation-related doesn't check out. And if we're confident about guest-components as a part of the CoCo stack, we shouldn't ignore breakage (the peerpod and kata CI is covering CDH + ASR now, afaik).

For debugging a developer could set `gc_procs=none` and attempt to launch the procs manually.

All that considered, I would recommend to fail in all cases.

stevenhorsman: > All that considered, I would recommend to fail in all cases.

Yeah - I think that's fine

stevenhorsman (approved):
After the discussions and updates I'm happy to go forward with this approach (until we can use a proper system/service manager!). Thanks Magnus!

src/agent/src/main.rs
fitzthum: Yeah this is an interesting one. Keep in mind that we already have runtime configurations that enable various feature that depend on these components. For instance if you enable the spi-server-rest, you need the whole stack. Other features (will) only require the CDH/AA. I am not against adding a config to explicitly set which processes are running, but it makes the user have to do a little bit more work.

Generally my instinct with this sort of thing is to fail fast and fatally, but let's say that the user enables the whole stack but doesn't use any config that depends on the guest components. In theory that could/should be non-fatal. My preference was actually to use rust features for all this because we can easily express dependency relationships, but we've moved away from build-time configuration.

Of course most of the time all the components will be there so idk how much this matters. In the other thread I suggested just logging the errors on startup and then failing later if someone tries to use a feature that is missing a dependency. I don't really have a preference, but either way it might be good to settle on some pattern that we can follow more generally.

mkulke: > For instance if you enable the spi-server-rest, you need the whole stack. Other features (will) only require the CDH/AA. I am not against adding a config to explicitly set which processes are running, but it makes the user have to do a little bit more work.

to simplify, the configuration here is incremental, so ASR implies CDH and AA, and CDH implies AA. in theory you could have ASR just to get evidence from AA without CDH, but I don't think we need to cover that. 

> but let's say that the user enables the whole stack but doesn't use any config that depends on the guest components. In theory that could/should be non-fatal

I think that's fair, but I wouldn't let processes fail to launch in cases of missing configuration but do the legwork and make sure that we're not launching them when we miss the prerequisites

fitzthum: yeah, if we're introducing a config to explicitly specify which components are launched it probably is best to fail if we can't launch them. I guess the only drawback here is that the client may have to fiddle with this config in some cases.

fidencio (approved):
lgtm, thanks @mkulke!

Issue #9750, opened 2024-05-30
tests/k8s: making confidential attestation tests less verbose 


wainersm:
**Which feature do you think can be improved?**

The  k8s-confidential-attestation.bats tests run `print_node_journal` at the `teardown()` which causes the worker node's kata's journal messaged be printed. `bats` by default only print the output of tests that fail, but we run with `--show-output-of-passing-tests` (for a good reason) which makes all output printed. Then these syslogs messages (usually a bunch of lines) gets interleaved with tests output, making it very hard to read the results.   

**How can it be improved?**

Print the worker node's kata's journal messaged only when the test fail.

**Additional Information**

I talked with the bats developers at their channel in gitter and I was told that currently there is only a workaround to guess if the test failed: https://github.com/bats-core/bats-core/issues/746#issuecomment-1629686296

But there's a feature request to introduce a `on_failure` callback (see https://github.com/bats-core/bats-core/issues/756) which should be implemented soon.

For now we can live with the workaround.

PR #9751, opened 2024-05-30
tests/k8s: print logs on fail only (k8s-confidential-attestation.bats)

wainersm:
Use the variable BATS_TEST_COMPLETED which is defined by the bats framework when the test finishes. `BATS_TEST_COMPLETED=` (empty) means the test failed, so the node syslogs will be printed only at that condition.

Fixes: #9750

---

First I created a function `has_test_failed()` (see below) that would be called from `teardown()`. It didn't work because `BATS_TEST_COMPLETED` is a local variable not visible from `has_test_failed()`, and I still don't know why. Any bash guru could help here? Then I decided to implement like in this PR.

```
--- a/tests/integration/kubernetes/tests_common.sh
+++ b/tests/integration/kubernetes/tests_common.sh
@@ -118,6 +118,16 @@ exec_host() {
        return ${exit_code}
 }
 
+# Return true if current test failed.
+# Use in teardown() to conditionally take actions.
+has_test_failed() {
+       if [ -z "$BATS_TEST_COMPLETED" ]; then
+               return 0
+       fi
+
+       return 1
+}
+
```

### DIFF
--- a/tests/integration/kubernetes/k8s-confidential-attestation.bats
+++ b/tests/integration/kubernetes/k8s-confidential-attestation.bats
@@ -93,7 +93,7 @@ teardown() {
 	[ -n "${pod_name:-}" ] && kubectl describe "pod/${pod_name}" || true
 	[ -n "${pod_config_dir:-}" ] && kubectl delete -f "${K8S_TEST_YAML}" || true
 
-	if [ -n "${node_start_time}:-}" ]; then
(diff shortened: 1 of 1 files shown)


fidencio (approved):
lgtm, thanks @wainersm!

gkurz (approved):
LGTM. Thanks @wainersm !

PR #9756, opened 2024-05-31
ci: Add NVIDIA GPU GHA

zvonkok:
Initial set of GH action for building GPU related artifacts


zvonkok:
Two runners were added `arm64-nvidia-gpu` and `amd64-nvidia-gpu`

zvonkok:
First round of additions. Next step is to get all the needed PRs for the rootfs build.

stevenhorsman (commented):
Hi @zvonkok - it feels like there is a lot of overlap with the existing arm and amd64 build/publish/release/test workflows. Do you think there is an opportunity to use some like this matrix, or inputs to refactor them to avoid this?

I appreciate that this is the case for the arm64, amd64, s390x and ppc64le workflows at the moment (and I raised it when they were added and was told that refactor would come later), but I would great to see things not get worse, without at least an issue to track the refactor effort (if feasible). Thanks

PR #9757, opened 2024-05-31
ci: k8s: Skip empty dir tests also for TDX

fidencio:
Wainer noticed this is failing for the coco-qemu-dev case, and decided to skip it, notifying me that he didn't fully understand why it was not failing on TDX.

Turns out, though, this is also failing on TDX, and we need to skip it there as well.

### DIFF
--- a/tests/integration/kubernetes/k8s-empty-dirs.bats
+++ b/tests/integration/kubernetes/k8s-empty-dirs.bats
@@ -18,6 +18,8 @@ assert_equal() {
 }
 
 setup() {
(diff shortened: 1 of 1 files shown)


zvonkok (approved):
LGTM, thanks !

fidencio:
I'm converting this to a draft, not because the PR is not ready, but because I want to make sure we have the TDX tests passing before it gets merged.

Thanks a lot for the quick reviews @gkurz and @zvonkok!

PR #9759, opened 2024-05-31
ci: k8s: Use kbs-client from inside a container

fidencio:
There's absolutely no need to build and use it from the host, if we can just build it and run it from a privileged container instead.

### DIFF
--- a/tests/integration/kubernetes/confidential_kbs.sh
+++ b/tests/integration/kubernetes/confidential_kbs.sh
@@ -24,6 +24,8 @@ readonly KBS_NS="coco-tenant"
 readonly KBS_PRIVATE_KEY="${COCO_KBS_DIR}/config/kubernetes/base/kbs.key"
 # The kbs service name
 readonly KBS_SVC_NAME="kbs"
+# The features to enable in the kbs client
+KBS_CLI_FEATURES="sample_only"
 
 # Set "allow all" policy to resources.
 #
@@ -52,9 +54,14 @@ kbs_set_resources_policy() {
 		return 1
 	fi
 
-	kbs-client --url "$(kbs_k8s_svc_http_addr)" config \
-		--auth-private-key "$KBS_PRIVATE_KEY" set-resource-policy \
-		--policy-file "$file"
+	docker run \
+		-v "/tmp:/tmp" \
+		--privileged \
+		--net=host \
+		$(kbs_get_cli_container_image_name) \
+			kbs-client --url "$(kbs_k8s_svc_http_addr)" config \
+			--auth-private-key "$KBS_PRIVATE_KEY" set-resource-policy \
+			--policy-file "$file"
 }
 
 # Set resource data.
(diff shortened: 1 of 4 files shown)


fidencio:
btw, let me expand here a little bit.  I'm aware that trustee builds and pushes its static binary, built with CLI_FEATURES="sample_only" to its repo as an ORAS image.

However, I will need sooner or later (in this case, sooner, as soon as this PR get merged), to build kbs-client with CLI_FEATURES="default", which means linking against libtdx-attest, which has no static library shipped, thus my decision to run everything from a container.

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -115,61 +122,57 @@ kbs_set_resource_from_file() {
...
+		-f Dockerfile.kbs-client .
+
+	docker push ${coco_kbs_client_container_image}
+}
+
+# If a docker image is not existent for the kbs-client, build it
+kbs_install_cli() {
+	# The name of this function is misleading ...
+
+	# but it's left as it is for backward compatibility with the CI
```
stevenhorsman, line 169: There looks to be indirection with the CI workflow via https://github.com/kata-containers/kata-containers/blob/7886ed66703221d082aa4b03a6f73ccf0da39458/tests/integration/kubernetes/gha-run.sh#L268-L270, so I think we could rename this function and update it in gha-run without breaking anything?

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -115,61 +122,57 @@ kbs_set_resource_from_file() {
...
 
 kbs_uninstall_cli() {
-	if [ -d "${COCO_KBS_DIR}" ]; then
-		pushd "${COCO_KBS_DIR}"
-		sudo make uninstall
-		popd
-	else
-		echo "${COCO_KBS_DIR} does not exist in the machine, skip uninstalling the kbs cli"
-	fi
+	# no-op, just left behind for backward compatibility with the CI
```
stevenhorsman, line 175: Again I guess we could remove this and update https://github.com/kata-containers/kata-containers/blob/7886ed66703221d082aa4b03a6f73ccf0da39458/tests/integration/kubernetes/gha-run.sh#L272-L274 to be the no-op, but maybe this is clearer

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -115,61 +122,57 @@ kbs_set_resource_from_file() {
...
 
-	kbs-client --url "$(kbs_k8s_svc_http_addr)" config \
-		--auth-private-key "$KBS_PRIVATE_KEY" set-resource \
-		--path "$path" --resource-file "$file"
+	docker run \
+		-v "$path:$path" \
+		-v "$file:$file" \
+		--privileged \
+		--net=host \
+			ghcr.io/kata-containers/kata-containers/coco-kbs-client:${cli_features}-${rust_version}-${trustee_version}-${uname -m} -- \
```
stevenhorsman, line 130: Can we use `$(kbs_get_cli_container_image_name)` here, so that if we change the name of the container we only have to do it in one place?

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -52,9 +54,14 @@ kbs_set_resources_policy() {
...
 	fi
 
-	kbs-client --url "$(kbs_k8s_svc_http_addr)" config \
-		--auth-private-key "$KBS_PRIVATE_KEY" set-resource-policy \
-		--policy-file "$file"
+	docker run \
+		-v "$file:$file" \
+		--privileged \
+		--net=host \
+			ghcr.io/kata-containers/kata-containers/coco-kbs-client:${cli_features}-${rust_version}-${trustee_version}-${uname -m} -- \
```
stevenhorsman, line 61: Same point about `$(kbs_get_cli_container_image_name)` as below

stevenhorsman (approved):
In general this LGTM and makes sense to do things in a container so we don't risk leaving things hanging around on the bare-metal systems. I've left a few questions/comments for consideration.

tests/integration/kubernetes/confidential_kbs.sh
fidencio: Not really, the full rename will come in a separate PR, as the function is called from run-kata-coco-tests.yaml

tests/integration/kubernetes/confidential_kbs.sh
fidencio: Done!

tests/integration/kubernetes/confidential_kbs.sh
fidencio: Done!

tests/integration/kubernetes/confidential_kbs.sh
fidencio: Done!

tests/integration/kubernetes/confidential_kbs.sh
stevenhorsman: I don't understand. `run-kata-coco-tests.yaml` calls `tests/integration/kubernetes/gha-run.sh install-kbs-client`, which calls `install_kbs_client` as I linked, so renaming `kbs_install_cli` here and then updating `install_kbs_client` won't break anything?

fitzthum:
Building the client in a container seems like a good idea. I don't think we will ever need to enable more features for the client, though (I may have said the opposite in the past, though :grimacing:). The CI only uses the client to provision resources/policies, which doesn't need any of the attesters to be enabled.

fidencio:
> The CI only uses the client to provision resources/policies, which doesn't need any of the attesters to be enabled.

This is exactly what I'm debugging and may challenge in the very near future. :-)
And I hope I'm wrong.

tests/integration/kubernetes/confidential_kbs.sh
fidencio: Ah, that won't break anything, indeed.  However, install-kbs-client is still a misleading name (as kbs_install_cli is as well).

I was planning to update those all together as part of a follow-up PR, as those changes mostly won't require CI running.

fidencio: I changed the name as part of this PR, leaving only the yaml change for a follow-up one.

wainersm:
> > The CI only uses the client to provision resources/policies, which doesn't need any of the attesters to be enabled.
> 
> This is exactly what I'm debugging and may challenge in the very near future. :-) And I hope I'm wrong.

Good point @fitzthum . @fidencio will you really need the CLI with TDX stuffs only for the sake of adding data to KBS? As I mentioned in private, the oras image built by trustee isn't enough?

fidencio:
> Good point @fitzthum . @fidencio will you really need the CLI with TDX stuffs only for the sake of adding data to KBS? As I mentioned in private, the oras image built by trustee isn't enough?

Most likely not.  If you prefer, I can do more concrete tests and leave this PR on-hold till proven that we may need to build kbs-client with the default CLI_FEATURES.

wainersm:
> > Good point @fitzthum . @fidencio will you really need the CLI with TDX stuffs only for the sake of adding data to KBS? As I mentioned in private, the oras image built by trustee isn't enough?
> 
> Most likely not. If you prefer, I can do more concrete tests and leave this PR on-hold till proven that we may need to build kbs-client with the default CLI_FEATURES.

I really prefer that :)

Because once we introduce this code here, it will be more build stuffs to maintain. If we can just consume the oras image from trustee, as in the "original plan", then it would be better.

tests/integration/kubernetes/confidential_kbs.sh
stevenhorsman: Cool. Thanks!

fidencio:
In the end I was able to get the CDH test to work with kbs-client as it is, as Tobin mentioned. :-)
Turns out we'll need to "update" our TDX machine, but that's already in process, closing this one for now.

PR #9760, opened 2024-05-31
kata-manager: Add symlinks for runc and slirp4netns

amshinde:
For nerdctl install, add symlinks for runc and slirp4netns in the binary install path.
runc link comes in handy for running runc containers with nerdctl fir quick tests.
slirp4netns allows for running containers with user mode networking useful in case of rootless containers.


Apokleos (approved):
Thx @amshinde LGTM

dborquez (commented):
Lgtm, thank you @amshinde

dborquez (approved):

PR #9770, opened 2024-06-03
tests: Use variable definition for better uniformity

GabyCT:
This PR replaces the name to use a variable that is already defined to have a better uniformity across the general script.

### DIFF
--- a/tests/integration/kubernetes/confidential_kbs.sh
+++ b/tests/integration/kubernetes/confidential_kbs.sh
@@ -179,8 +179,8 @@ kbs_uninstall_cli() {
 function kbs_k8s_delete() {
 	pushd "$COCO_KBS_DIR"
 	kubectl delete -k config/kubernetes/overlays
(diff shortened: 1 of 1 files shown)


tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -180,7 +180,7 @@ function kbs_k8s_delete() {
 	pushd "$COCO_KBS_DIR"
 	kubectl delete -k config/kubernetes/overlays
 	# Verify that coco-tenant resources were properly deleted
```
stevenhorsman, line 182: Is it worth updating the comment to something like:
`# Verify that KBS namespace resources were properly deleted`
in case we end up changing the namespace and for clarity?

GabyCT: @stevenhorsman changes applied

stevenhorsman (approved):
LGTM. Thanks!

dborquez (approved):

PR #9772, opened 2024-06-04
runtime-rs: add base qmp framework

pmores:
This PR creates a base framework for QMP functionality.  The idea of the new `Qmp` object is similar to the one of `QemuCmdLine` - encapsulate implementation details of QMP handling while offering `QemuInner` an interface that matches its needs.

The `qapi-rs` library is used to handle the QMP low-level details.  Its documentation isn't great but the library itself is close enough to QMP to be actually quite easy to use (so far at least ;-)).

We also add support for hotplugging vCPUs, basically as a test case for the newly added infrastructure. :-)

### DIFF
--- a/src/runtime-rs/crates/hypervisor/Cargo.toml
+++ b/src/runtime-rs/crates/hypervisor/Cargo.toml
@@ -43,6 +43,10 @@ safe-path = "0.1.0"
 crossbeam-channel = "0.5.6"
 tempdir = "0.3.7"
 
+qapi = { version = "0.14", features = [ "qmp", "async-tokio-all" ] }
+qapi-spec = "0.3.1"
+qapi-qmp = "0.14.0"
+
 [target.'cfg(not(target_arch = "s390x"))'.dependencies]
 dragonball = { path = "../../../dragonball", features = ["atomic-guest-memory", "virtio-vsock", "hotplug", "virtio-blk", "virtio-net", "virtio-fs", "vhost-net", "dbs-upcall", "virtio-mem", "virtio-balloon", "vhost-user-net", "host-device"] }
 
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -3,7 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use super::cmdline_generator::QemuCmdLine;
+use super::cmdline_generator::{QemuCmdLine, QMP_SOCKET_FILE};
+use super::qmp::Qmp;
 use crate::{
     hypervisor_persist::HypervisorState, utils::enter_netns, HypervisorConfig, MemoryConfig,
     VcpuThreadIds, VsockDevice, HYPERVISOR_QEMU,
@@ -32,6 +33,7 @@ pub struct QemuInner {
     id: String,
 
     qemu_process: Option<Child>,
+    qmp: Option<Qmp>,
 
     config: HypervisorConfig,
     devices: Vec<DeviceType>,
@@ -43,6 +45,7 @@ impl QemuInner {
         QemuInner {
             id: "".to_string(),
             qemu_process: None,
+            qmp: None,
             config: Default::default(),
             devices: Vec::new(),
             netns: None,
@@ -147,6 +150,14 @@ impl QemuInner {
 
         if let Some(ref mut qemu_process) = &mut self.qemu_process {
             tokio::spawn(log_qemu_stderr(qemu_process.stderr.take().unwrap()));
+        }
+
+        match Qmp::new(QMP_SOCKET_FILE) {
+            Ok(qmp) => self.qmp = Some(qmp),
+            Err(e) => {
+                error!(sl!(), "couldn't initialise QMP: {:?}", e);
+                return Err(e);
+            }
         }
 
         Ok(())
@@ -385,6 +396,7 @@ impl Persist for QemuInner {
         Ok(QemuInner {
             id: hypervisor_state.id,
             qemu_process: None,
+            qmp: None,
             config: hypervisor_state.config,
             devices: Vec::new(),
             netns: None,
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -17,6 +17,7 @@ use kata_types::{
     config::KATA_PATH,
 };
 use persist::sandbox_persist::Persist;
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::path::Path;
 use std::process::Stdio;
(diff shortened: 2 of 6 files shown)


src/runtime-rs/crates/hypervisor/Cargo.toml
```diff
@@ -43,6 +43,10 @@ safe-path = "0.1.0"
 crossbeam-channel = "0.5.6"
 tempdir = "0.3.7"
 
+qapi = { version = "0.14", features = [ "qmp", "qga", "async-tokio-all" ] }
+qapi-spec = "0.3.1"
+qapi-qmp = "0.14.0"
+
```
gkurz, line 49: Corresponding changes in `Cargo.lock` should be added to this commit as well.

pmores: You're right, thanks, I'll fix that.

lifupan (approved):
Thanks @pmores , LGTM！

Apokleos (commented):
Great! Pavel @pmores

Apokleos:
Discussion: A notable difference compared to the kata runtime with golang is that re-executing `qmpSetup()` before each qmp command, [such looks like called in `setupVirtioMem`](https://github.com/kata-containers/kata-containers/blob/138ef2c55f5127bc8944f347a7cc566961e9c227/src/runtime/virtcontainers/qemu.go#L920)
Why kata-runtime  do repeated `qmpSetup`, if we don't do such action in qemu-rs, is that OK ?
@gkurz @pmores FYI !

src/runtime-rs/crates/hypervisor/Cargo.toml
pmores: ... fixed.

src/runtime-rs/crates/hypervisor/Cargo.toml
```diff
@@ -43,6 +43,10 @@ safe-path = "0.1.0"
 crossbeam-channel = "0.5.6"
 tempdir = "0.3.7"
 
+qapi = { version = "0.14", features = [ "qmp", "qga", "async-tokio-all" ] }
```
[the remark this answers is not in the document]
pmores: True, good catch, yes, it should be fixed now.

pmores:
> Discussion: A notable difference compared to the kata runtime with golang is that re-executing `qmpSetup()` before each qmp command, [such looks like called in `setupVirtioMem`](https://github.com/kata-containers/kata-containers/blob/138ef2c55f5127bc8944f347a7cc566961e9c227/src/runtime/virtcontainers/qemu.go#L920) Why kata-runtime do repeated `qmpSetup`, if we don't do such action in qemu-rs, is that OK ? @gkurz @pmores FYI !

Yes, I noticed that as well.  My explanation is that in govmm, functions implementing QMP functionality are members of the `qemu` struct so they can be called anytime once `qemu` is constructed, possibly even before qemu is even launched.  Thus govmm lacks a single code path that would guarantee proper operation sequencing (i.e. first initialisation then usage) therefore each QMP function has to ensure individually that QMP is up and ready to go.  This is also hinted at by the fact that `qmpSetup()` is idempotent - it only does anything on its first invocation, subsequent invocations do nothing.

The structure of qemu-rs is different.  `Qmp` is only even instantiated after qemu has been launched, after your last PR ensured that a QMP socket is properly set up on qemu launch.  `Qmp`'s constructor ensures that QMP connection is set up properly, otherwise it fails, `Qmp` is never even constructed, an error is returned to runtime-rs proper which aborts the launch.  So by the time a `Qmp` member function that uses the QMP connection can even be called, it is guaranteed that QMP is good to go, therefore `Qmp` members don't need to check individually.

Apokleos:
> > Discussion: A notable difference compared to the kata runtime with golang is that re-executing `qmpSetup()` before each qmp command, [such looks like called in `setupVirtioMem`](https://github.com/kata-containers/kata-containers/blob/138ef2c55f5127bc8944f347a7cc566961e9c227/src/runtime/virtcontainers/qemu.go#L920) Why kata-runtime do repeated `qmpSetup`, if we don't do such action in qemu-rs, is that OK ? @gkurz @pmores FYI !
> 
> Yes, I noticed that as well. My explanation is that in govmm, functions implementing QMP functionality are members of the `qemu` struct so they can be called anytime once `qemu` is constructed, possibly even before qemu is even launched. Thus govmm lacks a single code path that would guarantee proper operation sequencing (i.e. first initialisation then usage) therefore each QMP function has to ensure individually that QMP is up and ready to go. This is also hinted at by the fact that `qmpSetup()` is idempotent - it only does anything on its first invocation, subsequent invocations do nothing.
> 
> The structure of qemu-rs is different. `Qmp` is only even instantiated after qemu has been launched, after your last PR ensured that a QMP socket is properly set up on qemu launch. `Qmp`'s constructor ensures that QMP connection is set up properly, otherwise it fails, `Qmp` is never even constructed, an error is returned to runtime-rs proper which aborts the launch. So by the time a `Qmp` member function that uses the QMP connection can even be called, it is guaranteed that QMP is good to go, therefore `Qmp` members don't need to check individually.
>
Yeah, Cool. Nice and clear explanation for me. I re-check the `qmpSetup` and yes  it's as you said.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -149,6 +152,14 @@ impl QemuInner {
             tokio::spawn(log_qemu_stderr(qemu_process.stderr.take().unwrap()));
         }
 
+        match Qmp::new(QMP_SOCKET_FILE) {
```
gkurz, line 156: This can legitimately fail because you have no guarantee that QEMU has called `listen()` yet.

Two options to address that :

1. pre-connect as the go runtime does (see https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/qemu.go#L964)
2. loop for some time until connection succeeds

pmores: I think just 2 is an option then since we don't do the connecting ourselves (qapi-rs does).

BbolroC (commented):
Other than @gkurz's comment, I am fine with it. I will defer to him for approval. Thanks @pmores !

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
gkurz: AFAICT qapi-rs does the QMP handshake but the connection is done by qemu-rs at https://github.com/kata-containers/kata-containers/pull/9772/commits/e26028c34cdcde085b866cc7d764284c3c46824a#diff-d564620b2876afa7b1db8ade06afda6d28c187cb887d27880af3fd46866d1362R16. If this gets moved out to the caller, it is possible to implement 1.

pmores: Right, I might be misunderstanding then what exactly it takes to implement the pre-connect option.

pmores: As it turns out this PR already does the pre-connect like the go runtime linked above.  The go code basically 1) `Listen()`s on the qmp socket path, 2) duplicates that socket, 3) `Dial()`s the qmp socket path, and 4) returns the connection from 3, stores the duplicate from 2 and lets the original from 1 die.

To see that we're doing all that already, please note that `UnixListener::bind()` (executed in `QmpSocket::new()`) does include a `listen()` call so we have an equivalent of 1.  Then 2 is not essential for the desired functionality and is apparently just a side effect of golang's `UnixListener::File()` implementation, we do 3 in this PR in `Qmp::new()` and we do our equivalents of 4, too.

So while I'll readily admit that this just by pure luck, I submit that this PR complies already. :-)

pmores:
Pushed a new version that sets a read timeout on the QMP connection.   This is to protect runtime-rs from blocking indefinitely on QMP in case qemu doesn't launch, and is done by runtime-go as well.

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
```diff
@@ -0,0 +1,127 @@
...
+impl Debug for Qmp {
+    fn fmt(&self, _f: &mut Formatter<'_>) -> Result<(), Error> {
+        Ok(())
+    }
+}
+
+impl Qmp {
+    pub fn new(qmp_sock_path: &str) -> Result<Self> {
+        let stream = UnixStream::connect(qmp_sock_path)?;
+        stream.set_read_timeout(Some(Duration::from_millis(250)))?;
```
Apokleos, line 31: why is timeout hardcode 250 ?

pmores: Good question!  The value is admittedly a matter of judgment but there's some thought behind it.  Setting the timeout is to protect runtime-rs from blocking forever trying to set up QMP connection if qemu fails to launch.  

The upper bound is the value of container runtime's task creation timeout (2 seconds for containerd by default) after which the container runtime aborts container launch and terminates the shim anyway.

The lower bound is the maximum delay that can reasonably be expected when things run fine.  I don't know what that could be but since this is a local unix-domain connection I suspect it should be fairly low.  However, it could be noticeable on a system under heavy load, and we don't want to interfere with a slightly delayed but otherwise successful hypervisor launch.

I thought a value between between 100 ms and 250 ms could be good.  I briefly mentioned that to @gkurz and we agreed to put 250 ms.

I'm not insisting that it must be 250 ms, if you have a reason to think that another value would be better I'm open to changing it.

Apokleos: Ok, It would be better add a brief comments about it to explain why setting 250 as the value of timeout for developers to check when they're confused.
Anyway, I think the PR reaches maturity. I will approve it.
Thx Pavel @pmores

pmores: I added a comment with explanation as per your request - please take a look.  Thanks @Apokleos !

Apokleos (approved):
Thx Pavel @pmores LGTM

PR #9784, opened 2024-06-06
kata-types: fix bug in kata-types several test cases

### LINKED ISSUE #9783: kata-types: do cargo test failed with some test cases
# Get your issue reviewed faster

kata-types: do cargo test failed with some test cases

Apokleos:
Fixes #9783

### DIFF
--- a/src/libs/kata-types/src/capabilities.rs
+++ b/src/libs/kata-types/src/capabilities.rs
@@ -127,7 +127,7 @@ mod tests {
         assert!(cap.is_fs_sharing_supported());
 
         // test set hybrid-vsock support
-        cap.set(CapabilityBits::HybridVsockSupport);
+        cap.add(CapabilityBits::HybridVsockSupport);
         assert!(cap.is_hybrid_vsock_supported());
         // test append capabilities
         cap.add(CapabilityBits::GuestMemoryProbe);
--- a/src/libs/kata-types/tests/texture/configuration-anno-0.toml
+++ b/src/libs/kata-types/tests/texture/configuration-anno-0.toml
@@ -84,6 +84,7 @@ sandbox_bind_mounts=["/proc/self"]
 vfio_mode="vfio"
 experimental=["a", "b"]
 enable_pprof = true
+name="virt-container"
 hypervisor_name = "qemu"
 agent_name = "agent0"
 
--- a/src/libs/kata-types/tests/texture/configuration-anno-1.toml
+++ b/src/libs/kata-types/tests/texture/configuration-anno-1.toml
@@ -83,6 +83,7 @@ sandbox_bind_mounts=["/proc/self"]
 vfio_mode="vfio"
 experimental=["a", "b"]
 enable_pprof = true
+name="virt-container"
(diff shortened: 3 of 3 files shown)


justxuewei (approved):
Lgtm, thanks!

lifupan (approved):

GabyCT (approved):

src/libs/kata-types/src/capabilities.rs
```diff
@@ -132,6 +132,6 @@ mod tests {
         // test append capabilities
         cap.add(CapabilityBits::GuestMemoryProbe);
         assert!(cap.is_mem_hotplug_probe_supported());
-        assert!(cap.is_fs_sharing_supported());
+        assert!(!cap.is_fs_sharing_supported());
```
amshinde, line 135: @Apokleos Why this change here?  On line 125 `FsSharingSupport` is being set and dont see it being unset anywhere. 
Why are you checking for false value here?

src/libs/kata-types/tests/texture/configuration-anno-0.toml
```diff
@@ -84,6 +84,7 @@ sandbox_bind_mounts=["/proc/self"]
 vfio_mode="vfio"
 experimental=["a", "b"]
 enable_pprof = true
+name="virt-container"
```
amshinde, line 87: Why is this change needed? Can you add some explanation in the commit message?

src/libs/kata-types/src/capabilities.rs
Apokleos: Thx @amshinde good catch !
As L130 cap does set causing previous Caps replaced by `CapabilityBits::HybridVsockSupport` and then there's no other Caps existing. with such reason, the  `assert!(cap.is_fs_sharing_supported());` failed. 
I rechecked the test case, the root cause is L130. It should be changed to ` cap.add(CapabilityBits::HybridVsockSupport);` and the L135 keep it was.

src/libs/kata-types/tests/texture/configuration-anno-0.toml
Apokleos: Ok, comments to be added. The explanation is , it will return error if no such name setting when do 
``` code
update_config_by_annotation {
    ...
if config.runtime.name.is_empty() {
            return Err(io::Error::new(
                io::ErrorKind::InvalidData,
                "Runtime name is missing in the configuration",
            ));
        }
...
}
```

But I am curious about our CIs no-run with UT ?

amshinde: Ok, thanks for the explanation and updating the commit message. Yes, I am a bit surprised why the CI did not catch this. Can you look into it?

amshinde (approved):

src/libs/kata-types/tests/texture/configuration-anno-0.toml
Apokleos: Yes, I will spend some time to check why CI doesn't run into these parts. And I will sync with you.

PR #9785, opened 2024-06-06
CI: Introduce retry mechanism for kubectl in gha-run.sh

BbolroC:
Frequent errors have been observed during k8s e2e tests:

- The connection to the server 127.0.0.1:6443 was refused - did you specify the right host or port?
- Error from server (ServiceUnavailable): the server is currently unable to handle the request
- Error from server (NotFound): the server could not find the requested resource

These errors can be resolved by retrying the kubectl command.

This commit introduces a wrapper function in common.sh that runs kubectl up to 3 times with a 5-second interval. Initially, this change only covers gha-run.sh for Kubernetes.

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -80,6 +80,22 @@ function handle_error() {
 }
 trap 'handle_error $LINENO' ERR
 
+# A wrapper function for kubectl with retry logic
+# runs the command up to 3 times with a 5-second interval
+# to ensure successful execution
+function kubectl_retry() {
+	local max_tries=3
+	local interval=5
+	local i=0
+	while true; do
+		kubectl $@ && return 0 || true
+		i=$((i + 1))
+		[ $i -lt $max_tries ] && echo "'kubectl $@' failed, retrying in $interval seconds" 1>&2 || break
+		sleep $interval
+	done
+	echo "'kubectl $@' failed after $max_tries tries" 1>&2 && return 1
+}
+
 function waitForProcess() {
 	wait_time="$1"
 	sleep_time="$2"
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -220,12 +220,12 @@ function deploy_kata() {
 	grep "${DOCKER_REGISTRY}/${DOCKER_REPO}:${DOCKER_TAG}" "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml" || die "Failed to setup the tests image"
 	echo "::endgroup::"
 
-	kubectl apply -f "${tools_dir}/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml"
+	kubectl_retry apply -f "${tools_dir}/packaging/kata-deploy/kata-rbac/base/kata-rbac.yaml"
 	case "${KUBERNETES}" in
-		k0s) kubectl apply -k "${tools_dir}/packaging/kata-deploy/kata-deploy/overlays/k0s" ;;
-		k3s) kubectl apply -k "${tools_dir}/packaging/kata-deploy/kata-deploy/overlays/k3s" ;;
-		rke2) kubectl apply -k "${tools_dir}/packaging/kata-deploy/kata-deploy/overlays/rke2" ;;
-		*) kubectl apply -f "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
+		k0s) kubectl_retry apply -k "${tools_dir}/packaging/kata-deploy/kata-deploy/overlays/k0s" ;;
+		k3s) kubectl_retry apply -k "${tools_dir}/packaging/kata-deploy/kata-deploy/overlays/k3s" ;;
+		rke2) kubectl_retry apply -k "${tools_dir}/packaging/kata-deploy/kata-deploy/overlays/rke2" ;;
+		*) kubectl_retry apply -f "${tools_dir}/packaging/kata-deploy/kata-deploy/base/kata-deploy.yaml"
 	esac
 
 	local cmd="kubectl -n kube-system get -l name=kata-deploy pod 2>/dev/null | grep '\<Running\>'"
@@ -241,11 +241,11 @@ function deploy_kata() {
 	fi
 
 	echo "::group::kata-deploy logs"
-	kubectl -n kube-system logs --tail=100 -l name=kata-deploy
+	kubectl_retry -n kube-system logs --tail=100 -l name=kata-deploy
 	echo "::endgroup::"
 
 	echo "::group::Runtime classes"
-	kubectl get runtimeclass
+	kubectl_retry get runtimeclass
 	echo "::endgroup::"
 }
 
@@ -395,7 +395,7 @@ function cleanup_kata_deploy() {
 	esac
 
 	# shellcheck disable=2086
-	kubectl delete ${deploy_spec}
+	kubectl_retry delete --ignore-not-found ${deploy_spec}
 	kubectl -n kube-system wait --timeout=10m --for=delete -l name=kata-deploy pod
 
 	# Let the `kata-deploy` script take care of the runtime class creation / removal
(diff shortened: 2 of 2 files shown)


BbolroC:
The following log shows that the mechanism works as expected:

```
Error from server (NotFound): the server could not find the requested resource ( pods/log nydus-snapshotter-gxth2)
  kubectl logs nydus-snapshotter-gxth2 -n nydus-system failed, retrying in 3 seconds
  Error from server (NotFound): the server could not find the requested resource ( pods/log nydus-snapshotter-gxth2)
  kubectl logs nydus-snapshotter-gxth2 -n nydus-system failed, retrying in 3 seconds
  Error from server (NotFound): the server could not find the requested resource ( pods/log nydus-snapshotter-gxth2)
  kubectl logs nydus-snapshotter-gxth2 -n nydus-system failed, retrying in 3 seconds
  Name:             nydus-snapshotter-gxth2   <--- continuing the workflow after 3 retries
  Namespace:        nydus-system
  Priority:         0
```

The log had been captured before increasing the interval from 3 secs to 5 secs (https://github.com/kata-containers/kata-containers/actions/runs/9404874866/job/25909282550).

justxuewei (approved):
Great work, thanks!

BbolroC:
One of the job logs (https://github.com/kata-containers/kata-containers/actions/runs/9409787120/job/25926473801) shows that the logic works as expected:

```
> Run bash tests/integration/kubernetes/gha-run.sh deploy-kata-zvsi
Context "default" modified.
> Final kata-deploy.yaml that is used in the test
The connection to the server 127.0.0.1:6443 was refused - did you specify the right host or port?
kubectl apply -f /home/ansible/gha-runner/_layout/_work/kata-containers/kata-containers/tools/packaging/
kata-deploy/kata-rbac/base/kata-rbac.yaml failed, retrying in 5 seconds   <<<--- HERE!!
serviceaccount/kata-deploy-sa created
clusterrole.rbac.authorization.k8s.io/kata-deploy-role created
clusterrolebinding.rbac.authorization.k8s.io/kata-deploy-rb created
daemonset.apps/kata-deploy created
kata-deploy-hwhnb   1/1     Running   0          34s
```

stevenhorsman (approved):
This looks cool to me and hopefully helps out with the timing issues. Thanks @BbolroC!

wainersm:
Hi @BbolroC ! This is indeed a cool idea and work!

I've only a few comments regarding the implementation.

tests/common.bash
```diff
@@ -80,10 +80,29 @@ function handle_error() {
...
+		sleep $interval
+	done
+	echo "kubectl $@ failed after $max_tries tries" && return 1
+}
+
 function waitForProcess() {
 	wait_time="$1"
 	sleep_time="$2"
 	cmd="$3"
+	if [ "${cmd}" == "kubectl" ]; then
```
wainersm, line 103: I'm not sure we should enable it on `waitForProcess` as this function in practice is an auto-retry. Also `$cmd` is a complete shell line, `[ "${cmd}" == "kubectl" ]` will always return false.

For example:
```
$ cmd="kubectl apply foobar"
$ [ "${cmd}" == "kubectl" ]
$ echo $?
1
```
or am I missing something?

tests/common.bash
```diff
@@ -80,10 +80,29 @@ function handle_error() {
...
+# runs the command up to 3 times with a 5-second interval
+# to ensure successful execution
+function kubectl_retry() {
+	local max_tries=3
+	local interval=5
+	local i=0
+	while [ $i -lt $max_tries ]; do
+		kubectl $@ && return 0 || true
+		i=$((i + 1))
+		[ $i -lt $max_tries ] && echo "kubectl $@ failed, retrying in $interval seconds" || break
```
wainersm, line 93: With this `[ $i -lt $max_tries ]` here (which is a good idea, saving the last sleep!) I think the loop can be simply `while true; do`

While in here, could you please single-quote `kube $@`? Like `echo "'kubectl $@' failed, retrying in $interval seconds"

tests/common.bash
```diff
@@ -80,10 +80,29 @@ function handle_error() {
...
+		sleep $interval
+	done
+	echo "kubectl $@ failed after $max_tries tries" && return 1
```
wainersm, line 96: Likewise

tests/common.bash
```diff
@@ -80,10 +80,29 @@ function handle_error() {
 }
 trap 'handle_error $LINENO' ERR
 
+# A wrapper function for kubectl with retry logic
```
wainersm, line 83: I feel so happy when I see functions documented \o/

tests/integration/kubernetes/gha-run.sh
```diff
@@ -395,8 +395,8 @@ function cleanup_kata_deploy() {
 	esac
 
 	# shellcheck disable=2086
-	kubectl delete ${deploy_spec}
-	kubectl -n kube-system wait --timeout=10m --for=delete -l name=kata-deploy pod
+	kubectl_retry delete ${deploy_spec}
+	kubectl_retry -n kube-system wait --timeout=10m --for=delete -l name=kata-deploy pod
```
wainersm, line 399: In case of a legit fail, e.g., due a bug the kata-deploy pod didn't get deleted this line will take 30"15' to return (vs the original 10"). Maybe don't retry here? or if retry then decrease the original timeout?

tests/integration/kubernetes/gha-run.sh
```diff
@@ -513,21 +513,21 @@ function deploy_nydus_snapshotter() {
...
-		kubectl apply -k "misc/snapshotter/overlays/k3s"
+		kubectl_retry apply -k "misc/snapshotter/overlays/k3s"
 	else
-		kubectl apply -f "misc/snapshotter/base/nydus-snapshotter.yaml"
+		kubectl_retry apply -f "misc/snapshotter/base/nydus-snapshotter.yaml"
 	fi
 	popd
 
-	kubectl rollout status daemonset nydus-snapshotter -n nydus-system --timeout ${SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT}
+	kubectl_retry rollout status daemonset nydus-snapshotter -n nydus-system --timeout ${SNAPSHOTTER_DEPLOY_WAIT_TIMEOUT}
```
wainersm, line 524: Likewise

tests/common.bash
beraldoleal: I had the same impression as you @wainersm .

Here, iiuc we are mixing a retry with a wait, it might confuse users of this method. Maybe use different methods for different situations (if possible)?

Also we could be creating precedence to include other exceptions.

beraldoleal (commented):
Super welcome improvement, @BbolroC thanks. Left just a comment for you, but on the same line as @wainersm .

tests/common.bash
BbolroC: Yep, let's not use `kubectl_retry` for `waitForProcess`. Thanks

tests/common.bash
BbolroC: Could you more elaborate on this? I've already explained about the function over 3 lines of comment. What does `functions documented` mean?

tests/integration/kubernetes/gha-run.sh
BbolroC: Oh, I missed it. Let's not use `kubectl_retry` for `kubectl` which  uses an `--timeout` option. Thanks.

tests/integration/kubernetes/gha-run.sh
BbolroC: Agreed.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -513,21 +513,21 @@ function deploy_nydus_snapshotter() {
...
 
 	echo "::endgroup::"
 	echo "::group::nydus snapshotter logs"
-	pods_name=$(kubectl get pods --selector=app=nydus-snapshotter -n nydus-system -o=jsonpath='{.items[*].metadata.name}')
-	kubectl logs "${pods_name}" -n nydus-system
-	kubectl describe pod "${pods_name}" -n nydus-system
+	pods_name=$(kubectl_retry get pods --selector=app=nydus-snapshotter -n nydus-system -o=jsonpath='{.items[*].metadata.name}')
```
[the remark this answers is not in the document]
BbolroC: Good idea!! I missed it. Thanks!

tests/common.bash
BbolroC: Yep!

tests/common.bash
BbolroC: Yep, I will follow your advice. Thanks!

GabyCT (approved):

wainersm (approved):
Great work @BbolroC ! Thanks!

beraldoleal (approved):
LGTM, thanks

Issue #8122, opened 2023-10-05
CC: Support pulling images on guest from private registries


stevenhorsman:
This feature builds upon the initial basic image pull work in https://github.com/kata-containers/kata-containers/issues/8103 as one of the possible next steps to extend it to support more of the current `CCv0` feature set.

The scope of this is to add the ability to pull a container image on the guest from an private credential using authentication information passed to the guest in a 'secure' way.

Initially I/we'd hoped to be able to cherry-pick over commits from `CCv0`, but given how much iteration there has been we're going to need to copy over chunks of the code and tests, improving as we go. Ideally we'll add `co-authored` tags to reflect the original authors though.

## Things to note/include with this work
- Add the [code in image_rpc.rs](https://github.com/kata-containers/kata-containers/blob/1ad87faaf4d3068569d99c270df640e41465c410/src/agent/src/image_rpc.rs#L144-L151) that enables credentials checking by the image client
- Add the `agent.image_registry_auth` part of https://github.com/kata-containers/kata-containers/pull/6707/files to enable the option of using image credentials without the need for a KBS. 
  > **Note:** we don't have a good way to edit the rootfs image and test this out in kata-containers, so will probably need to rely on other consumers of this feature, like peer-pods to test this (which has some existing tests from `CCv0`-based days https://github.com/confidential-containers/cloud-api-adaptor/blob/86e0f6d1da7b13c6a292948cfcc670c35f32b7cd/src/cloud-api-adaptor/test/e2e/common_suite.go#L235-L281 that we can re-use)
- New tests
  - There are a few new test cases I'd suggest, that are detailed in the acceptance criteria. The code for this code be based on the existing `CCv0` tests for [Kubernetes](https://github.com/stevenhorsman/tests/blob/31ea93fed12338f746b3c67642ef082f43471d38/integration/kubernetes/confidential/agent_image.bats#L185-L220), and [crictl](https://github.com/stevenhorsman/tests/blob/31ea93fed12338f746b3c67642ef082f43471d38/integration/containerd/confidential/agent_image.bats#L121-L139) but we want to work to make them easier to apply to different KBCs/TEE by extracting out the KBC/TEE logic into library functions, so the test cases are cleaner, similar to the approach done in https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-confidential.bats rather than a pure cherry-pick
[...]

PR #9803, opened 2024-06-10
agent: added param to provide auth for guest-pull

### LINKED ISSUE #9802: guest-pull: need a way to provide registry credentials to the agent

When using guest-pull via image-rs an image's metadata is pulled on the host, while the image layers are being pulled from a registry within the guest by image-rs. For authenticated registries we need a way to provide auth credentials to the guest. image-rs' image client can be initialized with a config.json auth file for OCI registries.

**How can it be improved?**

Add a kata param, gated behind the `guest-pull` feature flag, to point the agent to an config.json auth file it can use to pull image layers.

mkulke:
fixes #9802

The parameter is required for the `guest-pull` feature. While the metadata is pulled on the hosting node, the image layers are being pulled from the registry within the guest by image-rs. The image client can be initialized with a config.json auth file for OCI registries.

The change adds a param, gated behind `guest-pull` feature, to provide image-rs with an auth.config file. A simple sanity check for existence of this provided file is being performed at initialization.

### DIFF
--- a/src/agent/src/image.rs
+++ b/src/agent/src/image.rs
@@ -57,9 +57,15 @@ pub struct ImageService {
 
 impl ImageService {
     pub fn new() -> Self {
+        let mut image_client = ImageClient::new(PathBuf::from(KATA_IMAGE_WORK_DIR));
+        let images = HashMap::new();
+        if let Some(registry_auth) = &AGENT_CONFIG.image_registry_auth {
+            debug!(sl(), "Set registry auth file {:?}", registry_auth);
+            image_client.config.file_paths.auth_file = registry_auth.to_string_lossy().into();
+        }
         Self {
-            image_client: ImageClient::new(PathBuf::from(KATA_IMAGE_WORK_DIR)),
-            images: HashMap::new(),
+            image_client,
+            images,
         }
     }
 
@@ -287,9 +293,16 @@ pub async fn merge_bundle_oci(container_oci: &mut oci::Spec) -> Result<()> {
 }
 
 /// Init the image service
-pub async fn init_image_service() {
+pub async fn init_image_service() -> Result<()> {
+    if let Some(file) = &AGENT_CONFIG.image_registry_auth {
+        if !file.exists() {
+            bail!("The image registry auth file {:?} does not exist", file);
+        }
+    }
+
     let image_service = ImageService::new();
     *IMAGE_SERVICE.lock().await = Some(image_service);
+    Ok(())
 }
 
 pub async fn pull_image(
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -7,6 +7,7 @@ use anyhow::{anyhow, bail, ensure, Context, Result};
 use serde::Deserialize;
 use std::env;
 use std::fs;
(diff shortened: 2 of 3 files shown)


stevenhorsman:
At a quick read through I think this looks good. My concern with the PR is that it's missing the integration tests that we've enforced on the other CoCo features dropping into `main` (with some of the acceptance tests suggested in https://github.com/kata-containers/kata-containers/issues/8122), but I appreciate that adds a lot more work and complexity to get this feature in.

mkulke:
> At a quick read through I think this looks good. My concern with the PR is that it's missing the integration tests that we've enforced on the other CoCo features dropping into `main` (with some of the acceptance tests suggested in #8122), but I appreciate that adds a lot more work and complexity to get this feature in.

I think in general that's a fair requirement. If we _would_ want to skimp that, there's a point to be made that those integration tests w/ real registries are being performed in guest-components/image-rs, while this option merely passes a value to the image-rs api.

I'll look into how/where a reasonable integration test could be added before un-drafting it.

fidencio:
@mkulke, is the version of the guest-components we depend on the agent side good enough for this? Or would we also need to bump that?

mkulke:
> @mkulke, is the version of the guest-components we depend on the agent side good enough for this? Or would we also need to bump that?

I think there are no changes required. accessing the clients's config directly is not the prettiest of APIs and prone to breakage in a future refactoring of image-rs (since guest-components do not follow a versioning scheme yet) but that would be a compile error.

src/agent/src/image.rs
```diff
@@ -57,9 +57,15 @@ pub struct ImageService {
 
 impl ImageService {
     pub fn new() -> Self {
+        let mut image_client = ImageClient::new(PathBuf::from(KATA_IMAGE_WORK_DIR));
+        let images = HashMap::new();
+        if let Some(registry_auth) = &AGENT_CONFIG.image_registry_auth {
+            debug!(sl(), "Set registry auth file {:?}", registry_auth);
+            image_client.config.file_paths.auth_file = registry_auth.to_string_lossy().into();
```
wainersm, line 64: Hi @mkulke !

Reading [this line](https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/image.rs#L222) in image-rs' `pull_image()` and image-rs' credential test https://github.com/confidential-containers/guest-components/blob/main/image-rs/tests/credential.rs#L38

stevenhorsman:
@mkulke - FYI Wainer and I have been discussing tests for this, so we might end up creating a new PR with your commit cherry-picked in.

src/agent/src/image.rs
stevenhorsman: I've cherry-picked this and updated it in my version.

mkulke:
> @mkulke - FYI Wainer and I have been discussing tests for this, so we might end up creating a new PR with your commit cherry-picked in.

hey, yup, please go ahead. I was focussing on the AA/CDH config files and shelved this for the moment.

wainersm:
Superseded by https://github.com/kata-containers/kata-containers/pull/9904

Issue #9593, opened 2024-05-03
genpolicy: reenable checking the OCI version


sprt:
See #9592.

PR #9808, opened 2024-06-10
genpolicy: load OCI version from settings

Redent0r:
Load OCI version from genpolicy-settings.json and validate it in rules.rego

Fixes: #9593


fidencio (approved):
lgtm, thanks @Redent0r!

danmihai1 (approved):

PR #9811, opened 2024-06-10
agent: convert enum vals to kebab-case in cfg file

### LINKED ISSUE #9810: agent: parameter enum values in agent config are not kebab-cased
# Description of problem

In the agent file enum values cannot be specified in kebab-case

# Expected result

A config file entry `guest_components_rest_api = "all"` should be valid

# Actual result

We need to specify the value as `guest_components_rest_api = "All"`.

# Further information

The agent config file is parsed using the `AgentConfigBuilder` using `serde`, the params that have an enum as value are deserialized as literals without any conversions. There is a `strum` annotation on the enums to rename the entries to kebab-case, but that doesn't apply to the parsing logic.

mkulke:
fixes #9810

Add an annotation to the enum values in the agent config that will deserialize them using a kebab-case conversion, aligning the behaviour to parsing of params specified via kernel cmdline.

drive-by fix: add config override for guest_component_procs variable


fidencio (approved):
lgtm, thanks @mkulke!

stevenhorsman (approved):
LGTM

PR #9819, opened 2024-06-11
agent: config: Ensure envs take precedence

stevenhorsman:
- Update the config parsing logic so that when reading from the agent-config.toml file any envs are still processed
- Add units tests to formalise that the envs take precedence over values from the command line and the config file


Apokleos (approved):
Thx @stevenhorsman LGTM!

PR #9825, opened 2024-06-11
osbuilder: allow rootfs builds w/o git or version file deps

ms-mahuber:
This allows to call tooling like the rootfs builder outside of a git environment. Calculating the VERSION_COMMIT field without warnings/errors depends on the presence of a VERSION file and on a git environment. We set the VERSION_COMMIT field to 'unknown' if the version cannot be properly calculated. This not only prevents errors in the Makefile but also prevents from rootfs.sh being called with an empty -o parameter, thus the distro parameter from being parsed incorrectly.
The change also makes a rust version check optional for cases where we want to build the rootfs w/o building the agent and thus w/o the rust dependency

Fixes: 9824

### DIFF
--- a/tools/osbuilder/Makefile
+++ b/tools/osbuilder/Makefile
@@ -23,9 +23,9 @@ TARGET_IMAGE          := $(IMAGES_BUILD_DEST)/kata-containers.img
 TARGET_INITRD         := $(IMAGES_BUILD_DEST)/kata-containers-initrd.img
 
 VERSION_FILE   := ./VERSION
-VERSION        := $(shell grep -v ^\# $(VERSION_FILE))
-COMMIT_NO      := $(shell git rev-parse HEAD 2> /dev/null || true)
-COMMIT         := $(if $(shell git status --porcelain --untracked-files=no),${COMMIT_NO}-dirty,${COMMIT_NO})
+VERSION        := $(shell grep -v ^\# $(VERSION_FILE) 2>/dev/null || echo "unknown")
+COMMIT_NO      := $(shell git rev-parse HEAD 2>/dev/null || true)
+COMMIT         := $(if $(shell git status --porcelain --untracked-files=no 2>/dev/null || true),${COMMIT_NO}-dirty,${COMMIT_NO})
 VERSION_COMMIT := $(if $(COMMIT),$(VERSION)-$(COMMIT),$(VERSION))
 
 ifeq ($(filter $(BUILD_METHOD),$(BUILD_METHOD_LIST)),)
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -385,12 +385,6 @@ build_rootfs_distro()
 		mkdir -p ${ROOTFS_DIR}
 	fi
 
-	# need to detect rustc's version too?
-	detect_rust_version ||
-		die "Could not detect the required rust version for AGENT_VERSION='${AGENT_VERSION:-main}'."
-
-	echo "Required rust version: $RUST_VERSION"
-
 	if [ "${SELINUX}" == "yes" ]; then
 		if [ "${AGENT_INIT}" == "yes" ]; then
 			die "Guest SELinux with the agent init is not supported yet"
--- a/tools/osbuilder/scripts/lib.sh
+++ b/tools/osbuilder/scripts/lib.sh
@@ -225,14 +225,23 @@ generate_dockerfile()
 
 	[ -n "${http_proxy:-}" ] && readonly set_proxy="RUN sed -i '$ a proxy="${http_proxy:-}"' /etc/dnf/dnf.conf /etc/yum.conf; true"
 
-	# Rust agent
-	readonly install_rust="
+	# Only install Rust if agent needs to be built
+	local install_rust=""
+
+	if [ ! -z "${AGENT_SOURCE_BIN}" ] ; then
+		if [ "$RUST_VERSION" == "null" ]; then
+			detect_rust_version || \
+				die "Could not detect the required rust version for AGENT_VERSION='${AGENT_VERSION:-main}'."
+		fi
+		install_rust="
 ENV http_proxy=${http_proxy:-}
 ENV https_proxy=${http_proxy:-}
 RUN curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSLf | \
     sh -s -- -y --default-toolchain ${RUST_VERSION} -t ${rustarch}-unknown-linux-${LIBC}
 RUN . /root/.cargo/env; cargo install cargo-when
 "
+	fi
+
 	pushd "${dir}"
 
 	sed \
--- a/tools/osbuilder/scripts/lib.sh
+++ b/tools/osbuilder/scripts/lib.sh
@@ -181,7 +181,8 @@ create_summary_file()
 	[ "$AGENT_INIT" = yes ] && agent="${init}"
 
 	local -r agentdir="${script_dir}/../../../"
-	local -r agent_version=$(cat ${agentdir}/VERSION)
+	local agent_version=$(cat ${agentdir}/VERSION 2> /dev/null)
+	[ -z "$agent_version" ] && agent_version="unknown"
 
 	cat >"$file"<<-EOF
 	---
--- a/src/agent/Makefile
+++ b/src/agent/Makefile
@@ -15,7 +15,7 @@ PROJECT_COMPONENT = kata-agent
 TARGET = $(PROJECT_COMPONENT)
 
 VERSION_FILE := ./VERSION
-VERSION := $(shell grep -v ^\# $(VERSION_FILE))
+VERSION := $(shell grep -v ^\# $(VERSION_FILE) 2>/dev/null || echo "unknown")
 COMMIT_NO := $(shell git rev-parse HEAD 2>/dev/null || true)
 COMMIT := $(if $(shell git status --porcelain --untracked-files=no 2>/dev/null || true),${COMMIT_NO}-dirty,${COMMIT_NO})
 COMMIT_MSG = $(if $(COMMIT),$(COMMIT),unknown)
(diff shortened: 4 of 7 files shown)


tools/osbuilder/Makefile
```diff
@@ -23,9 +23,9 @@ TARGET_IMAGE          := $(IMAGES_BUILD_DEST)/kata-containers.img
 TARGET_INITRD         := $(IMAGES_BUILD_DEST)/kata-containers-initrd.img
 
 VERSION_FILE   := ./VERSION
-VERSION        := $(shell grep -v ^\# $(VERSION_FILE))
+VERSION        := $(if $(shell test -f $(VERSION_FILE)),$(shell grep -v ^\# $(VERSION_FILE)),"unknown")
```
fidencio, line 26: What would be the case where the `VERSION` file is not existent / not present?

Our releases do include such file, and any tarball generated on GitHub would as well.

tools/osbuilder/Makefile
```diff
@@ -23,9 +23,9 @@ TARGET_IMAGE          := $(IMAGES_BUILD_DEST)/kata-containers.img
...
 COMMIT_NO      := $(shell git rev-parse HEAD 2> /dev/null || true)
-COMMIT         := $(if $(shell git status --porcelain --untracked-files=no),${COMMIT_NO}-dirty,${COMMIT_NO})
+COMMIT         := $(if $(shell git status --porcelain --untracked-files=no 2> /dev/null),${COMMIT_NO}-dirty,${COMMIT_NO})
```
fidencio, line 28: For those 2, could we actually check whether it's a git repo as the first thing?  Even if we end up calling a helper shell script for doing that.

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -385,11 +385,15 @@ build_rootfs_distro()
...
-	echo "Required rust version: $RUST_VERSION"
+	# this check seems duplicate to the check in setup_rootfs, and we should
+	# also only need this check when we need to build the agent (see setup_rootfs)
+	if [ -z "${AGENT_SOURCE_BIN}" ] ; then
+		# need to detect rustc's version too?
+		detect_rust_version ||
+			die "Could not detect the required rust version for AGENT_VERSION='${AGENT_VERSION:-main}'."
+
+		echo "Required rust version: $RUST_VERSION"
+	fi
```
fidencio, line 396: I'd like to see this in a different commit, but still as part of the same PR.

But ... :-) ... can we try to simply remove this whole block? I think it's not needed.  And if it's needed, we should also check for AGENT_TARBALL.  But I truly believe it can be removed.

tools/osbuilder/scripts/lib.sh
```diff
@@ -181,7 +181,8 @@ create_summary_file()
 	[ "$AGENT_INIT" = yes ] && agent="${init}"
 
 	local -r agentdir="${script_dir}/../../../"
-	local -r agent_version=$(cat ${agentdir}/VERSION)
+	local agent_version=$(cat ${agentdir}/VERSION 2> /dev/null)
+	[ -z "$agent_version" ] && agent_version="unknown"
```
fidencio, line 185: Okay, but then again, when the VERSION file wouldn't be present?

fidencio (commented):
A few comments, @ms-mahuber.

tools/osbuilder/Makefile
ms-mahuber: I am not opinionated here. Once scenario *can* be where such release is being used to create packages (RPMs). For instance, a package that builds and ships the Kata host side components, or a minimal package that ships the relevant tooling/scripting (from osbuilder) to compose the guest components (rootfs, ...).

When creating such packages, my thinking is that one would merely want to include the minimal set of required files to compose the UVM.

Now, the VERSION file (which is symlinked to various places) seems to be less useful for some scenarios - for instance in the case where the UVM/rootfs is being assembled based on pre-built components (no build steps are executed, merely composing the UVM).

Looking at the code, the only real use of the VERSION file is within `create_summary_file` when we want to compose the rootfs based on pre-built components. Not everyone using kata-containers may be really interested in this summary file.

This is why I would propose to not 'hard depend' on this VERSION file and allow it to be inaccessible.

No emotional attachment here. If people think otherwise that's fair. :)

Note: Another topic may be in the future whether to include a summary file in the UVM at all (buzzword build reproducibility). :) Maybe an idea could be to produce this file alongside the rootfs artifact or make it an optional step.

tools/osbuilder/scripts/lib.sh
ms-mahuber: See answer above.

tools/osbuilder/rootfs-builder/rootfs.sh
ms-mahuber: To the best of my knowledge, this should not be needed. Happy to remove if nobody knows otherwise and if testing passes.

tools/osbuilder/Makefile
ms-mahuber: I believe there is even two aspects here when we think of more production-like environments that would build the UVM components:
- the source code may not be a git repository
- git may not even be available

So, I guess one may want to check for both and that would be in `utils.mk`?

tools/osbuilder/Makefile
fidencio: > For instance, a package that builds and ships the Kata host side components, or a minimal package that ships the relevant tooling/scripting (from osbuilder) to compose the guest components (rootfs, ...).

Oh, I see.  In this case we already installed such a tool in the system, and it'd still depend on the VERSION file to be placed somewhere, otherwise it'd bail.

I am not against having an "unknown" version here, as long as you're fine on having this on your side as well.

Ideally the release process should do better, and actually just have a version set there that the tool could use, without depending on the VERSION file, but this is orthogonal to your PR.

tools/osbuilder/Makefile
fidencio: So, let's break this down.

* Source code **is** a git repository
  * git **is** available, as a `git clone ...` happened
* Source code **is NOT** a git repository
  * git is available, but cannot be used to get the commit as this is not a git repo
  * git is not available

So, IMHO, it having git installed or not is really relevant if the source code is not a git repo.  And if it's a git repo, we know git is there as it was cloned from somewhere.

With that said, I'd just check whether the source code is a git repo or not.   And yes, I think the check could be done on `utils.mk`.

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: Yay, let's remove it.

ms-mahuber: I think we should remove this check for now indeed as there is no reason to check twice.

But: an early check in the case for when we need to build the agent may actually be reasonable. If you look at the relevant part in setup_rootfs(), the check looks a bit different, plus for instance a check for libseccomp is being executed (also, note that the .cargo related line is executed twice, unsure if that is required). Looks like that code can generally be improved.

Maybe in a subsequent PR one can think of adding a function that checks (and maybe installs) all the requirements early in order to fail early resp. more gracefully? Not sure what the intentions here are and whether it's worth the time. Thoughts? :)

tools/osbuilder/Makefile
ms-mahuber: I have aligned the read-in of the VERSION variable across all Makefiles

tools/osbuilder/Makefile
ms-mahuber: I have realized the same pattern exists in various other Makefiles and aligned the calls consistently. In a subsequent change one could de-duplicate those calls across makefiles.

ms-mahuber:
> A few comments, @ms-mahuber.

I've left a few new comments and split the change into two. I hope for the scope/intention of this PR/issue my changes make sense.

fidencio (approved):
lgtm, assuming green tests!

Thanks @ms-mahuber!

ms-mahuber:
> lgtm, assuming green tests!
> 
> Thanks @ms-mahuber!

Thanks @fidencio - looks like there is a dependency to this `detect_rust_version` call according to CI. Will be looking where this comes from. At a quick glance, it seems the agent is being built in one CI scenario but its rust dependencies are missing.

ms-mahuber:
> > lgtm, assuming green tests!
> > Thanks @ms-mahuber!
> 
> Thanks @fidencio - looks like there is a dependency to this `detect_rust_version` call according to CI. Will be looking where this comes from. At a quick glance, it seems the agent is being built in one CI scenario but its rust dependencies are missing.

@fidencio - in the failing CI case, the rootfs is being built using Docker. When running the Docker container, `INSTALL_RUST` is being executed (see `tools\osbuilder\rootfs-builder\ubuntu\Dockerfile.in` and `generate_dockerfile` in `tools\osbuilder\scripts\lib.sh`). This path relies on the `RUST_VERSION` variable being populated. Since my previous commit removed the `detect_rust_version` call, this path is now failing.

There is now two ways to address this:
- instead of removing the `detect_rust_version` call in `build_rootfs_distro`, move the `detect_rust_version` call into the code path where docker is being used (i.e., where `generate_dockerfile` is being called). This allows installing rust properly when the docker container is run as the `RUST_VERSION` variable is being set.
- only conditionally execute the `INSTALL_RUST` logic in the Dockerfile. It should only be required to build the agent. So, what's the purpose of unconditionally installing it. This can for instance be done by modifying `generate_dockerfile` by populating `install_rust` only if we need to build the agent.

fidencio:
> * only conditionally execute the `INSTALL_RUST` logic in the Dockerfile. It should only be required to build the agent. So, what's the purpose of unconditionally installing it. This can for instance be done by modifying `generate_dockerfile` by populating `install_rust` only if we need to build the agent.

I do prefer this approach.

Now, if you only need the changes related to the VERSION, I'd suggest to drop the patch from this PR and have that merged ASAP.  Then you open a PR with this second patch.

If you have the time, and taking the second approach doesn't prove to be too complicated, then we can do this in this very same PR.

danmihai1 (approved):

Issue #9826, opened 2024-06-12
runtime: panic when gathering metrics when running with a remote hypervisor


littlejawa:
# Description of problem

When running with a remote hypervisor, the shim makes a panic whenever kata-monitor tries to gather metrics.

# Expected result

kata-monitor should be able to gather metrics from the shim, for all hypervisors

# Actual result

No metrics can be gathered, and a panic stack dump is found in the kata logs

# Further information

The "GetVirtioFsPid()" function is called when the shim wants to gather metrics. This function is not implemented for remote hypervisor, and a call to panic() is made instead.

PR #9827, opened 2024-06-12
runtime: avoid panic on metrics gathering

littlejawa:
While running with a remote hypervisor, whenever kata-monitor tries to access metrics from the shim, the shim does a "panic" and no metric can be gathered.

The function GetVirtioFsPid() is called on metrics gathering, and had a call to "panic()". Since there is no virtiofs process for remote hypervisor, the right implementation is to return nil. The caller expects that, and will skip metrics gathering for virtiofs.

Fixes: #9826

### DIFF
--- a/src/runtime/virtcontainers/remote.go
+++ b/src/runtime/virtcontainers/remote.go
@@ -267,7 +267,7 @@ func (rh *remoteHypervisor) GetPids() []int {
 }
 
 func (rh *remoteHypervisor) GetVirtioFsPid() *int {
-	panic(notImplemented("GetVirtioFsPid"))
+	return nil
 }
 
 func (rh *remoteHypervisor) fromGrpc(ctx context.Context, hypervisorConfig *HypervisorConfig, j []byte) error {


Apokleos (approved):
Thx @littlejawa LGTM

danmihai1 (approved):

wainersm:
Hi @littlejawa !

We don't cover remote hypervisor on Kata tests (possible metrics are not even tests on peer pods, @stevenhorsman), so we didn't catch that before . Do you know what other component(s) might this GetVirtioFsPid() ? I'm just wondering what else might be broken.

littlejawa:
> Hi @littlejawa !
> 
> We don't cover remote hypervisor on Kata tests (possible metrics are not even tests on peer pods, @stevenhorsman), so we didn't catch that before . Do you know what other component(s) might this GetVirtioFsPid() ? I'm just wondering what else might be broken.

Hey @wainersm,

I've looked through the code, and the only place where this function is called is [here](https://github.com/kata-containers/kata-containers/blob/824287d64a9f030c7d796b34e330aba341a3feca/src/runtime/virtcontainers/sandbox_metrics.go#L186)

All other references to it are either implementation of the function by hypervisor management code, or test code. So we should be safe now.

Some additional notes:
- There are other "panic()" calls in the remote.go file, in the "fromGrpc()" and "toGrpc()" functions. It seems to be called when the "factory" code is exercised, but I think this is not the case today: looking up the chain of calls that leads to it, it seems we never use a factory object (it is explicitely set to "nil" when we create the container).
In any case, if it was exercised, we would get a panic on VM creation, so I think it would be obvious.

- There is also a bunch of "notimplemented()" error returns in the remote.go file, but they are on features that we actually don't support (yet?) with remote runtimes, so returning an error is fine.

stevenhorsman (approved):
LGTM and the reasoning makes sense. Thanks Julien

wainersm:
> > Hi @littlejawa !
> > We don't cover remote hypervisor on Kata tests (possible metrics are not even tests on peer pods, @stevenhorsman), so we didn't catch that before . Do you know what other component(s) might this GetVirtioFsPid() ? I'm just wondering what else might be broken.
> 
> Hey @wainersm,
> 
> I've looked through the code, and the only place where this function is called is [here](https://github.com/kata-containers/kata-containers/blob/824287d64a9f030c7d796b34e330aba341a3feca/src/runtime/virtcontainers/sandbox_metrics.go#L186)
> 
> All other references to it are either implementation of the function by hypervisor management code, or test code. So we should be safe now.
> 
> Some additional notes:
> 
>     * There are other "panic()" calls in the remote.go file, in the "fromGrpc()" and "toGrpc()" functions. It seems to be called when the "factory" code is exercised.
>       In any case, if it was exercised, we would get a panic on VM creation, so I think it would be obvious.
> 
>     * There is also a bunch of "notimplemented()" error returns in the remote.go file, but they are on features that we actually don't support (yet?) with remote runtimes, so returning an error is fine.

Glad to hear that we are good now. Thanks @littlejawa !

PR #9828, opened 2024-06-12
Image rs bump to latest main

stevenhorsman:
- Try bumping image-rs to latest main to pull in the latest security fixes and updates

### DIFF
--- a/src/agent/Cargo.lock
+++ b/src/agent/Cargo.lock
@@ -3,16 +3,6 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
-dependencies = [
- "lazy_static",
- "regex",
-]
-
-[[package]]
 name = "addr2line"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -143,9 +133,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.6"
+version = "0.4.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a116f46a969224200a0a97f29cfd4c50e7534e4b4826bd23ea2c3c533039c82c"
+checksum = "cd066d0b4ef8ecb03a55319dc13aa6910616d0f44008a045bb1835af830abff5"
 dependencies = [
  "flate2",
  "futures-core",
@@ -302,9 +292,9 @@ checksum = "7a40729d2133846d9ed0ea60a8b9541bccddab49cd30f0715a1da672fe9a2524"
 
 [[package]]
 name = "async-trait"
-version = "0.1.77"
+version = "0.1.80"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c980ee35e870bd1a4d2c8294d4c04d0499e67bca1e4b5cefcc693c2fa00caea9"
+checksum = "c6fa2087f2753a7da8cc1c0dbfcf89579dd57458e36769de5ac750b4671737ca"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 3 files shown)


stevenhorsman:
FYI I think there is possibly a bug in image-rs, maybe about layer ordering? We are seeing some failures like:
```
failed to create containerd task: failed to create shim task: failed to handle layer: hasher sha256: Operation not supported (os error 95): unknown
```
e.g. https://github.com/kata-containers/kata-containers/actions/runs/9486782210/job/26146223525?pr=9828 .

It doesn't happen every time, but pulling `registry.k8s.io/e2e-test-images/agnhost:2.21` seems to be the most consistent trigger of it

cc @arronwy @Xynnn007 @fidencio

stevenhorsman:
FYI - I've tried using `registry.k8s.io/e2e-test-images/agnhost:2.21` in the image-rs pull tests and can't reproduce the error, so there might be something specific to the confidential image?

stevenhorsman:
FYI @Xynnn007 - we are still hitting the same error
```
#   Warning  Failed     35s (x4 over 85s)  kubelet            Error: failed to create containerd task: failed to create shim task: failed to handle layer: hasher sha256: channel: send failed SendError { .. }: unknown
```
particularly with the `registry.k8s.io/e2e-test-images/agnhost:2.21` image. If we bump that image to the `2.41` it seems to be better, but I don't know if that is just hiding the problem?

Xynnn007:
> FYI @Xynnn007 - we are still hitting the same error
> 
> ```
> #   Warning  Failed     35s (x4 over 85s)  kubelet            Error: failed to create containerd task: failed to create shim task: failed to handle layer: hasher sha256: channel: send failed SendError { .. }: unknown
> ```
> 
> particularly with the `registry.k8s.io/e2e-test-images/agnhost:2.21` image. If we bump that image to the `2.41` it seems to be better, but I don't know if that is just hiding the problem?

@arronwy once talked that this is relavant to the image size. As the pulled image will be stored inside the `tmpfs` which is only a part of guest memory. Usually 1/10 of guest memory size.

However, I pulled the two images locally and `2.21` is 114MB, and `2.41` is 128 MB. Thus this might not be relavant to the image size?

Any ideas?

stevenhorsman:
> @arronwy once talked that this is relavant to the image size. As the pulled image will be stored inside the `tmpfs` which is only a part of guest memory. Usually 1/10 of guest memory size.
> 
> However, I pulled the two images locally and `2.21` is 114MB, and `2.41` is 128 MB. The bigger one is `2.41`, which is weird.
> 
> Any ideas?

That is a very interesting thought. IIRC the default memory size of the VMs we use is 2GB, so I guess it depends on how image-rs pull things and if there is a time that the layers are all pulled at the same time as the rootfs bundle is created from them so it could end up using more memory, or and maybe `agnhost:2.21` has a bigger max layer size, or bigger unpacked size even though `2.41` is bigger overall?

Just guessing here BTW!

I guess the other confusion is that I don't recall us ever seeing this error with previous versions of image-rs (e.g. commit `ca6b438`), so I wonder if any of the commits in image-rs since 12th March: https://github.com/confidential-containers/guest-components/commits/main/image-rs might be a cause of memory/size/layer handling changes?

We could try and do a bisect, but due to limitation in the kata ci it will be very slow and use a lot of compute power that we are lacking at the moment.

stevenhorsman:
I guess this also demonstrates again that we could do with better logging/error handling in image-rs. If we did run out of tmpfs here, then I don't think there is a security concern with throwing an error that hints at that?

stevenhorsman:
To add to the explanation of issues here. Ding added some debug code to image-rs and with it we were able to see that with the v0.9.0 version of image-rs the failure was whiteout related: `'failed to unpack: convert whiteout`. We believe this is caused by https://github.com/confidential-containers/guest-components/pull/563, so in the short-term are bumping to the commit before this change.

stevenhorsman:
With the latest image-rs (with the whiteout change reverted) all the tests using pull-image are now passing, so I think this is ready to merge.

Xynnn007 (approved):

stevenhorsman:
rebased on top of #9695

fidencio (approved):
lgtm, thanks @stevenhorsman!

ChengyuZhu6 (approved):
LGTM thanks @stevenhorsman

ChengyuZhu6:
Tests have passed on tdx https://github.com/kata-containers/kata-containers/actions/runs/9809346511/job/27153440236?pr=9828

Issue #9829, opened 2024-06-12
runtime-rs: fix the bug of func count_files


gaohuatao-1:
Runtime has fixed this bug, it should be ported to runtime-rs.

See #9780

PR #9830, opened 2024-06-12
runtime-rs: fix the bug of func count_files

gaohuatao-1:
When the total number of files observed is greater than limit, return -1 directly. runtime has fixed this bug, it should b ported to runtime-rs.

Fixes:#9829

### DIFF
--- a/src/libs/kata-types/src/k8s.rs
+++ b/src/libs/kata-types/src/k8s.rs
@@ -130,7 +130,11 @@ fn count_files<P: AsRef<Path>>(path: P, limit: i32) -> std::io::Result<i32> {
         let file = entry?;
         let p = file.path();
         if p.is_dir() {
-            num_files += count_files(&p, limit)?;
+            let inc = count_files(&p, limit - num_files)?;
+            if inc == -1 {
+                return Ok(-1);
+            }
+            num_files += inc;
         } else {
             num_files += 1;
         }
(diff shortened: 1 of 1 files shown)


Apokleos (commented):
Thx @gaohuatao-1 It looks good for me but lacks the UT of such general function.

justxuewei (approved):
Lgtm, thanks!

gaohuatao-1:
@justxuewei @Apokleos I have added a ut to the method. PTAL again, thanks.

gaohuatao-1:
@Apokleos  Hi, PTAL, Thanks.

Apokleos (commented):
Thx @gaohuatao-1 Compared with `CountFiles in kata runtime with golang`, this commit looks good to me. 
But I try to set the `limit < 0` (limit declared as i32), the UT fails. I think the argument  `limit: i32` in count_files()  shoud be change as `limit: u32`, maybe it's another bug, what do you think ?

gaohuatao-1:
> Thx @gaohuatao-1 Compared with `CountFiles in kata runtime with golang`, this commit looks good to me. But I try to set the `limit < 0` (limit declared as i32), the UT fails. I think the argument `limit: i32` in count_files() shoud be change as `limit: u32`, maybe it's another bug, what do you think ?

If the input param limit of func count_files is smaller than zero, test_count_files is still passed.  Of course, you cannot set the   local var limit to -1.  Could you please post your test code?

Apokleos:
> > Thx @gaohuatao-1 Compared with `CountFiles in kata runtime with golang`, this commit looks good to me. But I try to set the `limit < 0` (limit declared as i32), the UT fails. I think the argument `limit: i32` in count_files() shoud be change as `limit: u32`, maybe it's another bug, what do you think ?
> > @Apokleos
> > Thanks for your reply.
> > I  did the test:  the input param limit of func count_files is smaller than zero and test_count_files is still passed.  such as:
> > image
> 
> Could you please take a look again? Thanks a lot.
>
what's about -2 ? IMO, if you set -1 the return value is -1 which looks same with the Ok(-1), but they are in fact not same, I think.

gaohuatao-1:
> is



> > > Thx @gaohuatao-1 Compared with `CountFiles in kata runtime with golang`, this commit looks good to me. But I try to set the `limit < 0` (limit declared as i32), the UT fails. I think the argument `limit: i32` in count_files() shoud be change as `limit: u32`, maybe it's another bug, what do you think ?
> > > @Apokleos
> > > Thanks for your reply.
> > > I  did the test:  the input param limit of func count_files is smaller than zero and test_count_files is still passed.  such as:
> > > image
> > 
> > 
> > Could you please take a look again? Thanks a lot.
> 
> what's about -2 ? IMO, if you set -1 the return value is -1 which looks same with the Ok(-1), but they are in fact not same, I think.
-2 is ok, too. such as:
image

Apokleos (approved):
Thx @gaohuatao-1 LGTM

PR #9831, opened 2024-06-12
kernel: Enable 32bits emulation support (again)

stevefan1999-personal:
Revival of #3760, this should enable 32-bit emulation support. Here's the quote of the original justification:
> @fidencio Is it possible to reconsider this PR? I have a bunch of game servers that needs Steam access, but it currently lacks 32-bit compatibility due to Steam native library is still 32-bit based, rendering Kata hard to use in this case, because even if you have 64-bit game server, you will still fail to initialize Steam and the whole server just crashed at that point. Also, older HLDS and SRCDS servers like Counter Strike 1.6 and Garry's Mod dedicated servers are also 32-bit only at this moment. It is not certain for CS2 but I think the 32-bit Steam library would still be a stopper.
> 
> It takes quite a huge effort to build a kata-deploy for such use. Wonder why we can't just add CONFIG_IA32_EMULATION=y to the config matrix?

Shorter, tl;dr version:
> Please reconsider this PR. I have game servers needing Steam access, but the Steam native library is still 32-bit, making it difficult to use with Kata due to 64-bit game servers failing to initialize Steam. Older servers like Counter Strike 1.6 and Garry's Mod are also 32-bit only. Why not add CONFIG_IA32_EMULATION=y to the config matrix?

There is a whole fleet of game servers and all of them often requires 32-bit support due to Steam API library still being used as 32-bit at large (despite 64-bit library is available). One example I reckon is the [dedicated server for Insurgency Sandstorm](https://github.com/AndrewMarchukov/insurgency-sandstorm-server-dockerize).

Some dedicated server for old games (such as [Sven Coop](https://github.com/LacledesLAN/gamesvr-svencoop) and [TF2](https://github.com/melkortf/tf2-servers/blob/21c9580d003de57ab152bcc3fea567773db9ddc6/packages/tf2-base/Dockerfile#L29-L33)) are also 32-bit only.

[This is not a "corner case"](https://github.com/kata-containers/kata-containers/pull/3760#issuecomment-1063363194) and it is a typical problem with Kata Container's adoption with popular game server in real world.


fidencio:
@stevefan1999-personal, there are several things on your PR that are worth covering.

Whether or not having this enabled by default is out of my scope, but what's the main issue on building this on your side?  This will be used to evaluated and have an educated decision on the topic.

Another question is, would it be okay if we release a kernel with such support, but do not use this kernel by default? If this would help, I'd say this is worth considering, and then the work on adapting the configuration file to point to this kernel would still be on whoever is using it.  At least this would be simpler than forcing users to build their own kernel.

stevefan1999-personal:
> Whether or not having this enabled by default is out of my scope, but what's the main issue on building this on your side? This will be used to evaluated and have an educated decision on the topic.

I used [Talos Linux](https://www.talos.dev/) as a foundation for my Kubernetes cluster, and although their extension is available [here](https://github.com/siderolabs/extensions/tree/main/container-runtime/kata-containers), it is not trivial to track and build a custom kernel. But it is still possible.
 
> Another question is, would it be okay if we release a kernel with such support, but do not use this kernel by default? If this would help, I'd say this is worth considering, and then the work on adapting the configuration file to point to this kernel would still be on whoever is using it. At least this would be simpler than forcing users to build their own kernel.

Interesting take, but how do we make sure that it is easy for the end user to use that kernel? I'm using Kata Container on Kubernetes and the configuration is immutable on Talos Linux, so it is not easy to change the manifest to point to that kernel either. If you have auto Kata deployment on K8S it is also not easy to either since every time you hit `kata-deploy` your configuration file would reset.

zvonkok:
The 32-bit support was the base for this attack: https://ahoi-attacks.github.io/heckler/ against confidential VMs. Having this per default may have unseen consequences for all our use-cases.

stevefan1999-personal:
> The 32-bit support was the base for this attack: https://ahoi-attacks.github.io/heckler/ against confidential VMs. Having this per default may have unseen consequences for all our use-cases.

Per @fidencio's suggestion, is it possible to provide an alternative kernel that non-confidential VM users can easily enable? As a game server operator confidentiality is often not required, rather strong CPU and memory isolation is more than enough. Valve's model towards game server such as TF2 and CS2 is to use KVM and implemented a fast hypervisor tailored for game servers, and I believe Kata is the way to go even further.

stevefan1999-personal:
Another real world issue: SteamCMD 32-bit requirement. According to Valve's wiki:

> Since SteamCMD is a 32-bit binary, 32-bit libraries are required.

This is a hard requirement and I do not foresee any change in the future. Almost all public dedicated servers today need to use steamcmd for administration such as downloading and validating server assets, making this a hard requirement for a proper game server.

fidencio:
> I used [Talos Linux](https://www.talos.dev/) as a foundation for my Kubernetes cluster, and although their extension is available [here](https://github.com/siderolabs/extensions/tree/main/container-runtime/kata-containers), it is not trivial to track and build a custom kernel. But it is still possible.

Talos is using official releases, AFAIK.
We could, though, work with this on the Talos side and ensure they're fine providing a kernel with this change. WDYT?

stevefan1999-personal:
I still need this because I ran game servers that requires 32-bit support on Kata, not just SteamCMD but hlds too. Also, I remember UT2004 and UT99 dedicated server is also 32-bit only.

fidencio:
@stevefan1999-personal, I'm really sorry but I don't think we'll get this merged upstream.

We still provide all the infra for the users to build their own kernel on their side, but us building a kernel with 32-bit support and shipping it is a reasonable amount of cost (not only the build, but different CIs and etc as it cannot go as part of the default kernel).

I'm closing this for now.

Issue #9777, opened 2024-06-05
runtime-rs: possible problem in WaitProcess and DeleteProcess handling during shutdown


pmores:
QEMU driver in runtime-rs fails to shutdown cleanly.  The immediate cause is that its `Hypervisor::get_vmm_master_tid()` function is called after its `Hypervisor::stop_vm()` which terminates the qemu child process.  Thus `get_vmm_master_tid()` has no valid PID to return anymore and has to return an `Err`.

It seems odd for the runtime-rs proper to request shutdown of the hypervisor process/thread and then ask its PID/TID.  It happens because during shutdown, `Request::WaitProcess` is handled first(*) which stops the sandbox (and ultimately calls `Hypervisor::stop_vm()`).  Then `Request::DeleteProcess` is handled(**) whose processing calls `get_vmm_master_tid()` on a vmm that no longer runs.

`Hypervisor` implementations for CH and dragonball don't seem affected, apparently since both store the pertinent PID/TID in an own variable which enables them to return the PID/TID even after the corresponding process/thread has exited whereas qemu's implementation can only return a PID while it's actually valid.  (Note though that I might be wrong about the CH and dragonball drivers, I'm not overly familiar with them.)

I'm filing this issue to discuss how to do deal with this problem properly as I lack the deeper knowledge to make that call.  Fundamentally it seems that either sandbox is stopped too early in the shutdown process, or `DeleteProcess` handling should be able to make do without the vmm PID/TID (it does seem to run fine even with PID/TIDs that belong to an already stopped VM).

(*) The relevant part of the call stack is
```
   5: hypervisor::qemu::inner::QemuInner::stop_vm::{{closure}}
             at /home/pvl/src/work/kata-pmores/kata-containers/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs:169:9
   6: <hypervisor::qemu::Qemu as hypervisor::Hypervisor>::stop_vm::{{closure}}
             at /home/pvl/src/work/kata-pmores/kata-containers/src/runtime-rs/crates/hypervisor/src/qemu/mod.rs:62:25
   7: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/future/future.rs:125:9
   8: <virt_container::sandbox::VirtSandbox as common::sandbox::Sandbox>::stop::{{closure}}
[...]

Issue #9374, opened 2024-03-28
runtime-rs: ctr failed to delete a task when stopping a VM for QEMU


BbolroC:
# Description of problem

For runtime-rs with QEMU, we got the following error message when stopping a VM:

```
ctr: failed to delete task: Others("failed to handle message handler request\n\nCaused by:\n    0: do delete\n    1: cannot get qemu pid (though it seems running)"): unknown
```

This should be fixed.

PR #9834, opened 2024-06-13
sandbox: fix the issue of failed to get the vmm master tid

lifupan:
For kata container, the container's pid is meaning less to containerd/crio since the container's pid is belonged to VM, and containerd/crio couldn't use it. Thus we just return any tid of kata shim or hypervisor. But since the hypervisor had been stopped before deleting the container, and it wouldn't get the hypervisor's tid for some supported hypervisor, thus we'd better to return the kata shim's pid instead of hypervisor's tid.

Fixes: #9777
Fixes: #9374


pmores (approved):
lgtm, thanks a lot @lifupan !

Just a technical remark, since this PR fixes also issue #9374 it would be great to add a Fixes line for that issue as well.  Or we can handle #9374 manually afterwards, whatever is easier, let's just make sure that we don't leave the issue pollute the backlog.

Apokleos (approved):
Thx @lifupan LGTM

Issue #9778, opened 2024-06-05
runtime-rs: running a container with ctr and --memory-limit


pmores:
# Description of problem

I run a single kata container with a 5G memory limit (`ctr run --memory-limit 5368709120`) while having
```
[hypervisor.qemu]
default_memory = 2048
```
in my `configuration.toml`.

# Expected result

I expected to get a qemu VM with 2G of coldplugged memory and a call to `Qemu::resize_memory()` to handle the memory limit.

# Actual result

I got a qemu VM with 12G of coldplugged memory and a `resize_memory()` call to resize memory to 12288 (12G).

# Further information

The coldplugged memory size seems to come from [`InitialSizeManager` implementation](https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs) which, at least in a single container case, adds the memory limit size to `default_memory` and stores the result back in the config.  This also happens twice since runtime-rs appears to be processing a Create request twice(*).

The double call along with the non-idempotency of `InitialSizeManager` seems to explain the 12G of coldplugged memory (qemu sees 2G + 5G + 5G = 12G in `default_memory` in hypervisor config).

I have no idea about the `resize_memory()` to resize memory to the amount it already has.

(*) Relevant part of the callstack (same in both calls):
```
 { fn: "resource::cpu_mem::initial_size::InitialSizeManager::setup_config", file: "/home/pvl/src/work/kata-pmores/kata-containers/src/runtime-rs/crates/resource/src/cpu_mem/initial_size.rs", line: 102 },
 { fn: "runtimes::manager::RuntimeHandlerManagerInner::try_init::{{closure}}::{{closure}}", file: "/home/pvl/src/work/kata-pmores/kata-containers/src/runtime-rs/crates/runtimes/src/manager.rs", line: 184 },
 { fn: "runtimes::manager::RuntimeHandlerManagerInner::try_init::{{closure}}", file: "/home/pvl/src/work/kata-pmores/kata-containers/src/runtime-rs/crates/runtimes/src/manager.rs", line: 144 },
 { fn: "runtimes::manager::RuntimeHandlerManager::try_init_runtime_instance::{{closure}}::{{closure}}", file: "/home/pvl/src/work/kata-pmores/kata-containers/src/runtime-rs/crates/runtimes/src/manager.rs", line: 353 },
[...]

PR #9836, opened 2024-06-13
sandbox: fix the issue of double initial_size_manager config

lifupan:
It shouldn't call the initial_size_manager's setup_config in the load_config since it had been called in the sandbox's try_init function.

Fixes: #9778

### DIFF
--- a/src/runtime-rs/crates/runtimes/src/manager.rs
+++ b/src/runtime-rs/crates/runtimes/src/manager.rs
@@ -528,20 +528,6 @@ fn load_config(spec: &oci::Spec, option: &Option<Vec<u8>>) -> Result<TomlConfig>
     // validate configuration and return the error
     toml_config.validate()?;
 
-    // Sandbox sizing information *may* be provided in two scenarios:
-    //   1. The upper layer runtime (ie, containerd or crio) provide sandbox sizing information as an annotation
-    //	in the 'sandbox container's' spec. This would typically be a scenario where as part of a create sandbox
-    //	request the upper layer runtime receives this information as part of a pod, and makes it available to us
-    //	for sizing purposes.
-    //   2. If this is not a sandbox infrastructure container, but instead a standalone single container (analogous to "docker run..."),
-    //	then the container spec itself will contain appropriate sizing information for the entire sandbox (since it is
-    //	a single container.
-    let mut initial_size_manager =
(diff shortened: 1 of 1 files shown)


Apokleos (approved):
Thx fupan bro @lifupan It's indeed a dup code coming from different commits.
LGTM !

pmores (approved):
lgtm thanks a lot @lifupan !

As for [your question](https://github.com/kata-containers/kata-containers/issues/9778#issuecomment-2164878149) on the issue #9778: yes it does remove the double-add of memory limit so that's definitely a step ahead.

Now the VM is booted with `default_memory` + memory-limit (ie. 2G + 5G = 7G with reference to the concrete numbers I mentioned in the issue) and doesn't hotplug anything.  I'd like to confirm that this is the expected and correct behaviour. 

Originally I kind of expected memory-limit to be hot-plugged though thinking about it now, I can see how it makes sense to cold-plug it.  I'm wondering under which circumstances memory is truly hot-plugged in runtime-rs ("truly" as in "`resize_memory()` is called with an argument that _differs_ from the current VM memory").  Thanks in advance for any insights!

lifupan:
> lgtm thanks a lot @lifupan !
> 
> As for [your question](https://github.com/kata-containers/kata-containers/issues/9778#issuecomment-2164878149) on the issue #9778: yes it does remove the double-add of memory limit so that's definitely a step ahead.
> 
> Now the VM is booted with `default_memory` + memory-limit (ie. 2G + 5G = 7G with reference to the concrete numbers I mentioned in the issue) and doesn't hotplug anything. I'd like to confirm that this is the expected and correct behaviour.
> 
> Originally I kind of expected memory-limit to be hot-plugged though thinking about it now, I can see how it makes sense to cold-plug it. I'm wondering under which circumstances memory is truly hot-plugged in runtime-rs ("truly" as in "`resize_memory()` is called with an argument that _differs_ from the current VM memory"). Thanks in advance for any insights!

Hi  @pmores 

Generally speaking, the container's memory resource limit would be hotplugged since most of the time when creating the container, the VM had been launched. The containers created using the k8s/crictl are most of this case.  But for case of created with ctr/nerdctl client, the container and the VM would be created/launch at the same stage, thus there's an optimization and calculated the total memory and did a cold plug.

pmores:
Thanks @lifupan , that make a lot of sense and I do see container resources hotplugged.

PR #9838, opened 2024-06-13
CI: remove sudo from GHA

zvonkok:
Now that all artifacts are owned by $USER we can start to remove sudo from our GHA

### DIFF
--- a/.github/workflows/build-kata-static-tarball-amd64.yaml
+++ b/.github/workflows/build-kata-static-tarball-amd64.yaml
@@ -87,7 +87,7 @@ jobs:
           make "${KATA_ASSET}-tarball"
           build_dir=$(readlink -f build)
           # store-artifact does not work with symlink
-          sudo cp -r "${build_dir}" "kata-build"
+          mkdir -p kata-build && cp "${build_dir}"/kata-static-${KATA_ASSET}*.tar.* kata-build/.
         env:
           KATA_ASSET: ${{ matrix.asset }}
           TAR_OUTPUT: ${{ matrix.asset }}.tar.gz


stevenhorsman:
@zvonkok - Are you thinking of making the same change to the arm, ppc and s390x workflows as a separate PR?

BbolroC:
### s390x

I can see that files under the following directories are owned by root:

- `tools/packaging/kata-deploy/local-build/build/rootfs-image-confidential`

This is what I have just checked instantly on one of the runners.  I will update the list whenever I notice the ownership. Thanks.

zvonkok:
@stevenhorsman This is for testing if this works extend to "ALL".

stevenhorsman:
> @stevenhorsman This is for testing if this works extend to "ALL".

Thanks, sounds like a plan to me.

.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
           make "${KATA_ASSET}-tarball"
           build_dir=$(readlink -f build)
           # store-artifact does not work with symlink
-          sudo cp -r "${build_dir}" "kata-build"
+          mkdir kata-build && cp "${build_dir}/kata-static-*.tar.xz" kata-build/.
```
wainersm, line 90: Hi @zvonkok !

Let's make CI (almost) sudo-less! \o/

`kata-build` shouldn't exist but if for some miracle it does then `mkdir` fails. It is guaranteed  `mkdir -p` will be succeeded.

Apart from that, the code fails:
```shell
$ build_dir=$(readlink -f build)
$ mkdir kata-build && cp "${build_dir}/kata-static-*.tar.xz" kata-build/.
cp: cannot stat '/home/wmoschet/src/github.com/kata-containers/kata-containers/tools/packaging/kata-deploy/local-build/build/kata-static-*.tar.xz': No such file or directory
$ ls build
agent  agent-builder-image-version  agent-sha256sum  agent-version  kata-static-agent.tar.xz
$ ls $build_dir 
agent  agent-builder-image-version  agent-sha256sum  agent-version  kata-static-agent.tar.xz
$ mkdir kata-build && cp "${build_dir}"/kata-static-*.tar.xz kata-build/.
mkdir: cannot create directory ‘kata-build’: File exists
$ mkdir -p kata-build && cp "${build_dir}"/kata-static-*.tar.xz kata-build/.
$ ls kata-build/
kata-static-agent.tar.xz
```
It works with `mkdir -p kata-build && cp "${build_dir}"/kata-static-*.tar.xz kata-build/.`  (notice the `"${build_dir}"/kata-static-*.tar.xz`)

GabyCT (approved):

.github/workflows/build-kata-static-tarball-amd64.yaml
zvonkok: Ha it is not a miracle, for each asset we do this hence if we build a couple of assets (matrix.asset) and we do it will fail let me make this better good catch.

zvonkok:
@wainersm @stevenhorsman @fidencio updated. Let's only copy only the current build asset and not every asset again and again on each `matrix.asset` iteration. Also leave the compression appendix open if we might change compression algorithms this will still work.

.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
           make "${KATA_ASSET}-tarball"
           build_dir=$(readlink -f build)
           # store-artifact does not work with symlink
-          sudo cp -r "${build_dir}" "kata-build"
+          mkdir -p kata-build && cp "${build_dir}/kata-static-${KATA_ASSET}.tar.*" kata-build/.
```
stevenhorsman, line 90: Will this skip the extra modules/headers in the kernel build now? Maybe `"${build_dir}"/kata-static-${KATA_ASSET}*.tar.*` would be better (note the moving of double quotes too that I think Wainer said was needed for the wildcard matches to work

stevenhorsman (commented):
This looks okay to me, but it would be great to get Wainer's check of it.

zvonkok:
Waiting for @wainersm to give us the blessing ...

.github/workflows/build-kata-static-tarball-amd64.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
           make "${KATA_ASSET}-tarball"
           build_dir=$(readlink -f build)
           # store-artifact does not work with symlink
-          sudo cp -r "${build_dir}" "kata-build"
+          mkdir -p kata-build && cp "${build_dir}"/kata-static-${KATA_ASSET}*.tar.* kata-build/.
```
wainersm, line 90: Hi @zvonkok @stevenhorsman !

Yes, it does work. Thanks!

There is a situation where it might copy the tarball more than one time, but I don't think it is a big deal. When the asset's name is sub-string of another asset's name, for example, "kernel" and "kernel-confidential":
```
$ rm -rf kata-build/
$ export KATA_ASSET=kernel
$ build_dir=$(pwd)/build ; mkdir -p kata-build && cp "${build_dir}"/kata-static-${KATA_ASSET}*.tar.* kata-build/.
$ ls kata-build/
kata-static-kernel-confidential-modules.tar.xz  kata-static-kernel-confidential.tar.xz  kata-static-kernel.tar.xz
$ rm -rf kata-build/
$ export KATA_ASSET=kernel-confidential
$ build_dir=$(pwd)/build ; mkdir -p kata-build && cp "${build_dir}"/kata-static-${KATA_ASSET}*.tar.* kata-build/.
$ ls kata-build/
kata-static-kernel-confidential-modules.tar.xz  kata-static-kernel-confidential.tar.xz
```

We were already copying everything over and over again, this is just a corner case, again, not a big deal IMO.

wainersm (approved):
Thanks @zvonkok !

stevenhorsman (approved):
Thanks Zvonko & Wainer

stevenhorsman:
A bunch of the static checks are stuck and as this changes a workflow, the tests aren't valid, so I will bypass the checks to get it merged.

PR #9845, opened 2024-06-13
gha: Do not fail when collecting artifacts

GabyCT:
This PR will avoid the failures when collecting artifacts for the gha.


.github/workflows/run-k8s-tests-on-garm.yaml
```diff
@@ -86,7 +86,7 @@ jobs:
         run: bash tests/integration/kubernetes/gha-run.sh run-tests
   
       - name: Collect artifacts ${{ matrix.vmm }}
-        run: bash tests/integration/kubernetes/gha-run.sh collect-artifacts
+        run: bash tests/integration/kubernetes/gha-run.sh collect-artifacts || true
```
wainersm, line 89: Hi @GabyCT ! Thanks for promptly working on this one!

One question that I have and I forgot to mention on the issue is whether the next step `Archive artifacts ${{ matrix.vmm }}` is going to fail in case /tmp/artifacts doesn't exist.

ah, also, while here, I think we want to have this step ` if: always()` so it runs in case the test fails.

GabyCT: @wainersm so I added the if always step

wainersm (approved):
LGTM. Thanks @GabyCT !

.github/workflows/basic-ci-amd64.yaml
```diff
@@ -333,7 +333,8 @@ jobs:
         run: bash tests/integration/nerdctl/gha-run.sh run
 
       - name: Collect artifacts ${{ matrix.vmm }}
-        run: bash tests/integration/nerdctl/gha-run.sh collect-artifacts
+        if: always()
+        run: bash tests/integration/nerdctl/gha-run.sh collect-artifacts || true
```
beraldoleal, line 337: Hi @GabyCT, do you mind explaining better (at the commit level) why this is necessary?

Probably I'm missing things, but won't  this `|| true` make it difficult to debug the job? since it will mask the failed jobs?

GabyCT: @beraldoleal changes applied

beraldoleal: Hi @GabyCT sorry if is too noise, but IIUC, this might generate errors to the output when the script fails for some reason, but it will tell GH actions the step was successfully. As a developer debugging issues in a run, I would except to see the steps failing. Any reason for not using `continue-on-error: true`?

Also, "This will ensure that we collect and archive system's data for the purpose of debugging." not sure if this is true since it is not fixing the "collect and archive"  aspect, but just making sure the GH will not fail.

GabyCT: @beraldoleal changes applied

beraldoleal (approved):
LGTM, thanks

PR #9856, opened 2024-06-14
genpolicy: reject untested CreateContainer field values

danmihai1:
Reject CreateContainerRequest field values that are not tested by Kata CI and that might impact the confidentiality of CoCo Guests.

This change uses a "better safe than sorry" approach to untested fields. It is very possible that in the future we'll encounter
reasonable use cases that will either:

- Show that some of these fields are benign and don't have to be verified by Policy, or
- Show that Policy should verify legitimate values of these fields

These are the new CreateContainerRequest Policy rules:

    count(input.shared_mounts) == 0
    is_null(input.string_user)

    i_oci := input.OCI
    is_null(i_oci.Hooks)
    is_null(i_oci.Linux.Seccomp)
    is_null(i_oci.Solaris)
    is_null(i_oci.Windows)

    i_linux := i_oci.Linux
    count(i_linux.GIDMappings) == 0
    count(i_linux.MountLabel) == 0
    count(i_linux.Resources.Devices) == 0
    count(i_linux.RootfsPropagation) == 0
    count(i_linux.UIDMappings) == 0
    is_null(i_linux.IntelRdt)
    is_null(i_linux.Resources.BlockIO)
    is_null(i_linux.Resources.Network)
    is_null(i_linux.Resources.Pids)
    is_null(i_linux.Seccomp)
    i_linux.Sysctl == {}

    i_process := i_oci.Process
    count(i_process.SelinuxLabel) == 0
    count(i_process.User.Username) == 0

### DIFF
--- a/src/tools/genpolicy/rules.rego
+++ b/src/tools/genpolicy/rules.rego
@@ -52,10 +52,15 @@ default WriteStreamRequest := false
 default AllowRequestsFailingPolicy := false
 
 CreateContainerRequest {
+    # Check if the input request should be rejected even before checking the
+    # policy_data.containers information.
+    allow_create_container_input
+
     i_oci := input.OCI
     i_storages := input.storages
     i_devices := input.devices
 
+    # Check if any element from the policy_data.containers array allows the input request.
     some p_container in policy_data.containers
     print("======== CreateContainerRequest: trying next policy container")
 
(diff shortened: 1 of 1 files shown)


src/tools/genpolicy/rules.rego
```diff
@@ -85,6 +90,38 @@ CreateContainerRequest {
...
+    i_oci := input.OCI
+    is_null(i_oci.Hooks)
+    is_null(i_oci.Linux.Seccomp)
+    is_null(i_oci.Solaris)
+    is_null(i_oci.Windows)
+
+    i_linux := i_oci.Linux
+    count(i_linux.GIDMappings) == 0
+    count(i_linux.MountLabel) == 0
+    count(i_linux.Resources.Devices) == 0
```
burgerdev, line 108: As mentioned in https://github.com/kata-containers/kata-containers/pull/9706#discussion_r1624683480, I think this is going to break device plugins. I'm not too concerned about that yet, but that may change when people want to use GPUs or something :)

 I have some code with which I can verify that on Monday.

burgerdev (commented):
Thanks, this is a great addition!

src/tools/genpolicy/rules.rego
danmihai1: Thanks! Sure, if we have broken use cases, let's fix them.

danmihai1: @burgerdev, do you think we should merge this PR? Or should we wait for additional experiments?

Thanks!

burgerdev: Sorry for the delay. Things I verified to keep working:

* Adding a character device via device plugin
* Adding a block device via CSI

None of these sets `i_linux.Resources.Devices`, no idea how I arrived at that expectation. Anyway, please go ahead with the merge!

danmihai1: No worries, thanks for helping out!

sprt (approved):

GabyCT (approved):

PR #9870, opened 2024-06-18
runtime-rs: adjust qemu vm shutdown behaviour

pmores:
This is mostly just a small batch of logging and consistency improvements.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -167,10 +167,19 @@ impl QemuInner {
     pub(crate) async fn stop_vm(&mut self) -> Result<()> {
         info!(sl!(), "Stopping QEMU VM");
         if let Some(ref mut qemu_process) = &mut self.qemu_process {
-            info!(sl!(), "QemuInner::stop_vm(): kill()'ing qemu");
-            qemu_process.kill().await.map_err(anyhow::Error::from)
+            let is_qemu_running = qemu_process.id().is_some();
+            if is_qemu_running {
+                info!(sl!(), "QemuInner::stop_vm(): kill()'ing qemu");
+                qemu_process.kill().await.map_err(anyhow::Error::from)
+            } else {
+                info!(
+                    sl!(),
+                    "QemuInner::stop_vm(): qemu process isn't running (likely stopped already)"
+                );
+                Ok(())
+            }
         } else {
-            Err(anyhow!("qemu process not running"))
+            Err(anyhow!("qemu process has not been started yet"))
         }
     }
 
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -231,7 +231,7 @@ impl QemuInner {
                 );
                 Ok(qemu_pid)
             } else {
-                Err(anyhow!("cannot get qemu pid (though it seems running)"))
+                Err(anyhow!("QemuInner::get_vmm_master_tid(): qemu process isn't running (likely stopped already)"))
             }
         } else {
             Err(anyhow!("qemu process not running"))


pmores:
@lifupan I'm still kind of doubtful whether it's correct to shutdown the VM on WaitProcess handling.  This PR makes the qemu driver tolerate multiple attempts to stop the VM but I'm wondering if we couldn't prevent the repeated `stop_vm()` calls in the first place.  What do you think?

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -167,10 +167,19 @@ impl QemuInner {
...
-            info!(sl!(), "QemuInner::stop_vm(): kill()'ing qemu");
-            qemu_process.kill().await.map_err(anyhow::Error::from)
+            let is_qemu_running = qemu_process.id().is_some();
+            if is_qemu_running {
+                info!(sl!(), "QemuInner::stop_vm(): kill()'ing qemu");
+                qemu_process.kill().await.map_err(anyhow::Error::from)
+            } else {
+                info!(
+                    sl!(),
+                    "QemuInner::stop_vm(): qemu process isn't running (likely stopped already)"
```
gkurz, line 177: It is really unfortunate for that condition to be handled here but it is certainly the best option for now.
If you happen to re-post, please add some `// FIXME` to make it clear that this is a workaround that will need to be revisited (and not blindly copied into a future hypervisor driver).

BbolroC (approved):
LGTM, thanks @pmores

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -167,10 +167,19 @@ impl QemuInner {
...
         info!(sl!(), "Stopping QEMU VM");
         if let Some(ref mut qemu_process) = &mut self.qemu_process {
-            info!(sl!(), "QemuInner::stop_vm(): kill()'ing qemu");
-            qemu_process.kill().await.map_err(anyhow::Error::from)
+            let is_qemu_running = qemu_process.id().is_some();
+            if is_qemu_running {
+                info!(sl!(), "QemuInner::stop_vm(): kill()'ing qemu");
+                qemu_process.kill().await.map_err(anyhow::Error::from)
+            } else {
+                info!(
```
Apokleos, line 175: Nit: warn!() or info!(), which one is better ?

Apokleos (approved):
Thx Pavel @pmores LGTM

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
pmores: Note that I'm unfortunately _not_ sure that this is a work-around. :-( It could be that the shim v2 protocol just doesn't offer a better opportunity than the WaitProcess handler to stop the sandbox.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
pmores: That's a good question... I stayed with `info!()` in the end on the base that this PR makes the qemu driver accept multiple `stop_vm()` calls as normal, business as usual, so from that standpoint there's kind of nothing to warn about. :-)

BbolroC:
@gkurz please merge this PR when your question is answered. 😉

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
gkurz: Not being sure about how this is expected to work is even a better reason for a comment :wink:. At this stage we should know if shim v2 allows that and if it doesn't we should put a comment to explain the flow.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
gkurz: Warnings are usually devoted to something that we want the user to be aware of so that they can _maybe_ take action. In this case, it seems that the warning would popup every time and the user cannot really do anything about it. I'd keep `info!()`.

gkurz:
> @gkurz please merge this PR when your question is answered. 😉

I've asked for an extra comment and I don't have the feedback I'm expecting yet, so I won't merge now :wink:

This being said, the PR is already mergeable. Anyone in a hurry can merge it right away :wink:

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
pmores: The purpose of this PR is simply to make the qemu driver's behaviour consistent with other drivers whose `stop_vm()` implementations are (silently) idempotent.

I don't feel comfortable to comment on the call pattern of the function since I don't have the necessary understanding, just suspicions.  I raised a question above to clear this up but unless someone who does know chimes in we simply have to accept that we currently don't know what the right solution is.  I will continue investigating the issue but that's far beyond the scope of this PR.

Issue #9850, opened 2024-06-14
CI: nightly `kata-containers-ci-on-push / run-kata-deploy-tests-on-aks / run-kata-deploy-tests (ubuntu, qemu-runtime-rs)`  failing


wainersm:
Nightly `kata-containers-ci-on-push / run-kata-deploy-tests-on-aks / run-kata-deploy-tests (ubuntu, qemu-runtime-rs)` has always failed because the cluster name is too big, exceeding the limit of chars of AKS...

```
{
  "id": "/subscriptions/***/resourceGroups/kataCI-kata-deploy-nightly-f66a5b628725-qemu-runtime-rs-ubuntu-amd64-s-o",
  "location": "eastus",
  "managedBy": null,
  "name": "kataCI-kata-deploy-nightly-f66a5b628725-qemu-runtime-rs-ubuntu-amd64-s-o",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": null,
  "type": "Microsoft.Resources/resourceGroups"
}
WARNING: The behavior of this command has been altered by the following extension: aks-preview
ERROR: --name cannot exceed 63 characters and can only contain letters, numbers, or dashes (-).
```

wainersm:
One thing that I didn't understand is why it provisioned a cluster with:

`kataCI-kata-deploy-nightly-56423cbbfe66-dragonball-ubuntu-amd64-s-o`

and not

`kataCI-kata-deploy-nightly-56423cbbfe66-qemu-runtime-rs-ubuntu-amd64-s-o`

Because both characters count exceed 63:
```
$ echo -n kataCI-kata-deploy-nightly-56423cbbfe66-dragonball-ubuntu-amd64-s-o | wc -m
67
$ echo -n kataCI-kata-deploy-nightly-56423cbbfe66-qemu-runtime-rs-ubuntu-amd64-s-o | wc -m
72
```

Now, if one assume that special characters aren't counted (e.g. dash) then both are below the 63 chars limit:
```
$ echo -n kataCI-kata-deploy-nightly-56423cbbfe66-dragonball-ubuntu-amd64-s-o | tr -d '-' | wc -m
58
$ echo -n kataCI-kata-deploy-nightly-56423cbbfe66-qemu-runtime-rs-ubuntu-amd64-s-o | tr -d '-' | wc -m
61
```

[...]

PR #9871, opened 2024-06-18
tests/gha-run-k8s-common: shorten AKS cluster name

wainersm:
Because az client restricts the name to be less than 64 characters. In
    some cases (e.g. KATA_HYPERVISOR=qemu-runtime-rs) the generated name
    will exceed the limit. This changed the function to shorten the name:
    
    * SHA1 is computed from metadata then compound the cluster's name
    * metadata as plain-text are passed as --tags

Fixes: #9850

### DIFF
--- a/tests/gha-run-k8s-common.sh
+++ b/tests/gha-run-k8s-common.sh
@@ -31,12 +31,16 @@ function _print_instance_type() {
 function _print_cluster_name() {
 	local test_type="${1:-k8s}"
 	local short_sha
+	local metadata
 
 	if [ -n "${AKS_NAME:-}" ]; then
 		echo "$AKS_NAME"
 	else
 		short_sha="$(git rev-parse --short=12 HEAD)"
-		echo "${test_type}-${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1}"
+		metadata="${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1}"
+		# Compute the SHA1 digest of the metadata part to keep the name less
+		# than the limit of 63 chars of AKS
+		echo "${test_type}-$(sha1sum <<< "$metadata" | cut -d' ' -f1)"
 	fi
 }
 
@@ -80,11 +84,21 @@ function login_azure() {
 
 function create_cluster() {
 	test_type="${1:-k8s}"
+	local short_sha
+	local tags
 
 	# First ensure it didn't fail to get cleaned up from a previous run.
 	delete_cluster "${test_type}" || true
 
 	local rg="$(_print_rg_name ${test_type})"
+
+	short_sha="$(git rev-parse --short=12 HEAD)"
+	tags=("GH_PR_NUMBER=${GH_PR_NUMBER:-}" \
+		"SHORT_SHA=${short_sha}" \
+		"KATA_HYPERVISOR=${KATA_HYPERVISOR}"\
+		"KATA_HOST_OS=${KATA_HOST_OS:-}" \
+		"K8S_TEST_HOST_TYPE=${K8S_TEST_HOST_TYPE:0:1}" \
+		"GENPOLICY_PULL_METHOD=${GENPOLICY_PULL_METHOD:0:1}")
 
 	az group create \
 		-l eastus \
@@ -97,6 +111,7 @@ function create_cluster() {
 		-s "$(_print_instance_type)" \
 		--node-count 1 \
 		--generate-ssh-keys \
+		--tags "${tags[@]}" \
 		$([ "${KATA_HOST_OS}" = "cbl-mariner" ] && echo "--os-sku AzureLinux --workload-runtime KataMshvVmIsolation")
 }
 


wainersm:
Tested as:

```
$ echo $(GH_PR_NUMBER=kata-deploy-nightly; source ../../gha-run-k8s-common.sh; KATA_HYPERVISOR=qemu-runtime-rs; KATA_HOST_OS=ubuntu; echo $(_print_cluster_name KataCI))
KataCIkatadeploynightly486672qemuruntimersubuntuamd64so
$ echo $(GH_PR_NUMBER=kata-deploy-nightly; source ../../gha-run-k8s-common.sh; KATA_HYPERVISOR=qemu; KATA_HOST_OS=ubuntu; echo $(_print_cluster_name KataCI))
KataCI-kata-deploy-nightly-486672-qemu-ubuntu-amd64-s-o
```

wainersm:
@jepio does it break your code that clean up the instantiated AKS clusters?

tests/gha-run-k8s-common.sh
```diff
@@ -31,12 +31,16 @@ function _print_instance_type() {
...
 	if [ -n "${AKS_NAME:-}" ]; then
 		echo "$AKS_NAME"
 	else
-		short_sha="$(git rev-parse --short=12 HEAD)"
-		echo "${test_type}-${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1}"
+		short_sha="$(git rev-parse --short=6 HEAD)"
+		name="${test_type}-${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1}"
+		# An AKS cluster must have less than 64 characters. Attempt to shorten the name by removing dashes.
+		[ ${#name} -lt 64 ] || name=$(echo -n "$name" |  tr -d '-')
+		echo "$name"
```
mkulke, line 43: that might be surprising to people who assume that cluster name follow a common naming scheme, I wouldn't do that pro-actively. If we still reach > 64 chars, we have to consider moving metadata to tags

jepio:
> @jepio does it break your code that clean up the instantiated AKS clusters?

It shouldn't - the automation scans for any k8s cluster in that subscription.

jepio:
> > @jepio does it break your code that clean up the instantiated AKS clusters?
> 
> It shouldn't - the automation scans for any k8s cluster in that subscription.

Pushed the deployed code here: https://github.com/jepio/kata-azure-automation

tests/gha-run-k8s-common.sh
sprt: We can't really move metadata to tags since we're using those to ensure the cluster name is unique to a given test run. Maybe we should hash it instead? Like: `${test_type}-${GH_PR_NUMBER}-HASH(${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1})`

wainersm: I thought in using hash as @sprt suggested but then the cluster name will be intelligible. Having a meaningful name is important for when there are dangling clusters and you want to back-track to the original pipeline; but given that we are already clean up all aging k8s clusters we probably won't need to inspect the clusters anymore....meaningful names become almost useless. All in all, it seems that using hashes ought to work out.

jepio: could we atleast add the metadata to tags then? there are cases where clusters were stuck undeletable, and it helps correlated back to a GHA run.

wainersm: yes, it seems the best approach @jepio . I will re-do this PR. Thanks guys!

wainersm:
Hi @mkulke @sprt , re-implemented as we discussed. Mind to give another review?

tests/gha-run-k8s-common.sh
```diff
@@ -31,12 +31,16 @@ function _print_instance_type() {
...
 
 	if [ -n "${AKS_NAME:-}" ]; then
 		echo "$AKS_NAME"
 	else
 		short_sha="$(git rev-parse --short=12 HEAD)"
-		echo "${test_type}-${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1}"
+		metadata="${GH_PR_NUMBER}-${short_sha}-${KATA_HYPERVISOR}-${KATA_HOST_OS}-amd64-${K8S_TEST_HOST_TYPE:0:1}-${GENPOLICY_PULL_METHOD:0:1}"
+		# Compute the SHA1 digest of the metadata part to keep the name less
+		# than the limit of 63 chars of AKS
+		echo "${test_type}-$(sha1sum <<< "$metadata" | cut -d'-' -f1)"
```
sprt, line 43: Nit: I would still include at least the PR number in cleartext so it's directly obvious when looking at the list of clusters in the UI and such (may want to truncate the SHA1).

mkulke (commented):
just to make sure: aren't we good with the short git sha for the time being? So maybe we can leave the current naming scheme in place (w/ a shortened git sha), keep the tags (they don't hurt) but only change the naming scheme, once we run into the issue again?

wainersm:
> just to make sure: aren't we good with the short git sha for the time being? So maybe we can leave the current naming scheme in place (w/ a shortened git sha), keep the tags (they don't hurt) but only change the naming scheme, once we run into the issue again?

Even if we shrink the git sha to 6 chars, we still end up with 67 chars long names (`kataCI-kata-deploy-nightly-81d23a-qemu-runtime-rs-ubuntu-amd64-s-o`).

tests/gha-run-k8s-common.sh
wainersm: I thought about that too, @sprt. Then it will need to truncate the sha1 to keep it within the limit.

wainersm:
Hi @mkulke @sprt !

I'd like to revisit that PR. are you ok with the fix?

danmihai1 (approved):

mkulke (approved):

tests/gha-run-k8s-common.sh
sprt: @wainersm Any issue with addressing this?

sprt: Ping here :)

wainersm:
Hi all!

Cc @stevenhorsman @fidencio 

Just rebased and fixed a bug on the function that returns the cluster name.

I tested it locally and it did work:
```shell
$ export KATA_HOST_OS=ubuntu
$ export KATA_HYPERVISOR=qemu-runtime-rs
$ export GH_PR_NUMBER=nightly
$ cd tests/integration/kubernetes
$ ./gha-run.sh create-cluster
{
  "id": "xxxxxxxxxxxxxxxx",
  "location": "eastus",
  "managedBy": null,
  "name": "kataCI-k8s-b6c0971da3d55906d768bd12c22dd4e855f7fced",
  "properties": {
  },
  "tags": null,
}
The behavior of this command has been altered by the following extension: aks-preview
The new node pool will enable SSH access, recommended to use '--ssh-access disabled' option to disable SSH access for the node pool to make it more secure.
{
  "aadProfile": null,
  "addonProfiles": null,
  "agentPoolProfiles": [
    {
      "artifactStreamingProfile": null,
      "availabilityZones": null,
      "capacityReservationGroupId": null,
      "count": 1,
      "creationData": null,
      "currentOrchestratorVersion": "1.30.6",
<snip>
  "name": "k8s-b6c0971da3d55906d768bd12c22dd4e855f7fced",
  "networkProfile": {
    "advancedNetworking": null,
    "dnsServiceIp": "10.0.0.10",
    "ipFamilies": [
      "IPv4"
    ],
    "kubeProxyConfig": null,
    "loadBalancerProfile": {
      "allocatedOutboundPorts": null,
      "backendPoolType": "nodeIPConfiguration",
      "clusterServiceLoadBalancerHealthProbeMode": null,
      "effectiveOutboundIPs": [
        {
          "id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
          "resourceGroup": "node-kataCI-k8s-b6c0971da3d55906d768bd12c22dd4e855f7fced"
        }
      ],
<snip>
  "supportPlan": "KubernetesOfficial",
  "systemData": null,
  "tags": {
    "GENPOLICY_PULL_METHOD": "o",
    "GH_PR_NUMBER": "nightly",
    "K8S_TEST_HOST_TYPE": "s",
    "KATA_HOST_OS": "ubuntu",
    "KATA_HYPERVISOR": "qemu-runtime-rs",
    "SHORT_SHA": "955a5370e806"
  },
  "type": "Microsoft.ContainerService/ManagedClusters",
  "upgradeSettings": null,
  "windowsProfile": null,
  "workloadAutoScalerProfile": {
    "keda": null,
    "verticalPodAutoscaler": null
  }
}
```

stevenhorsman (approved):
I guess with the name switching to a sha of the metadata there is slightly less transparency in where the cluster comes from, but the tags should make up for that, so LGTM. Thanks!

fidencio (approved):
lgtm, thanks @wainersm!

PR #9876, opened 2024-06-19
ci: remove sudo from s390x build

zvonkok:
We can now do the same for s390x that we did for amd64 and remove the sudo cp.


BbolroC (approved):
LGTM, thanks @zvonkok !

stevenhorsman (approved):

GabyCT (approved):

PR #9879, opened 2024-06-19
runtime-rs: remove attempt to access sandbox bundle from container bu…

### LINKED ISSUE #9847: runtime-rs: container start fails under cri-o
When cri-o tries to launch a runtime-rs pod it fails with
```
ERROR: containerd-shim-kata-v2 failed: shim start

Caused by:
   0: do start
   1: failed to get bundle path
   2: No such file or directory (os error 2): connect: invalid argument
```
This happens when the shim is launched with the "start" argument and `container_type` is `ContainerType::PodContainer` in [`ShimExecutor::do_start()`](https://github.com/kata-containers/kata-containers/blob/824287d64a9f030c7d796b34e330aba341a3feca/src/runtime-rs/crates/shim/src/shim_start.rs#L35).  [`ShimExecutor::get_shim_info_from_sandbox()`](https://github.com/kata-containers/kata-containers/blob/824287d64a9f030c7d796b34e330aba341a3feca/src/runtime-rs/crates/shim/src/shim_start.rs#L109) is then called which tries to derive the path to the sandbox bundle from the container bundle the shim is currently running in.

This fails since the bundle directory structure is different between containerd and cri-o (containerd stores the bundles in `<container_root>/<sandbox_id>` and `<container_root>/<container_id>` while cri-o in `<container_root>/<sandbox_id>/userdata` and `<container_root>/<container_id>/userdata`, respectively).

Having discussed this with @littlejawa we believe that the shim shouldn't try to access the sandbox bundle in the first place while starting a container.  For a smoke test, I've removed the offending code locally and have been running with it for a bit, having observed no ill effects so far.

Do you folks agree that this is the right solution?  Admittedly I'm not clear what the reason for the current shim behaviour was in the first place so I might be missing something.

pmores:
…ndle

Since no objections were raised in the linked issue (#9847) this commit removes the attempt to derive sandbox bundle path from container bundle path.  As described in more detail in the linked issue, this is container runtime specific and doesn't seem to serve any purpose.

Fixes #9847

### DIFF
--- a/src/runtime-rs/crates/shim/src/shim_start.rs
+++ b/src/runtime-rs/crates/shim/src/shim_start.rs
@@ -50,8 +50,7 @@ impl ShimExecutor {
                 let sid = id
                     .ok_or(Error::InvalidArgument)
                     .context("get sid for container")?;
-                let (address, pid) = self.get_shim_info_from_sandbox(&sid)?;
-                self.write_pid_file(&bundle_path, pid)?;
+                let address = self.socket_address(&sid).context("socket address")?;
                 self.write_address(&bundle_path, &address)?;
                 Ok(address)
             }
@@ -105,22 +104,6 @@ impl ShimExecutor {
 
         Ok(child.id())
     }
-
-    fn get_shim_info_from_sandbox(&self, sandbox_id: &str) -> Result<(PathBuf, u32)> {
-        // All containers of a pod share the same pod socket address.
-        let address = self.socket_address(sandbox_id).context("socket address")?;
-        let bundle_path = get_bundle_path().context("get bundle path")?;
-        let parent_bundle_path = Path::new(&bundle_path)
-            .parent()
-            .unwrap_or_else(|| Path::new(""));
-        let sandbox_bundle_path = parent_bundle_path
-            .join(sandbox_id)
-            .canonicalize()
-            .context(Error::GetBundlePath)?;
-        let pid = self.read_pid_file(&sandbox_bundle_path)?;
-
-        Ok((address, pid))
-    }
 }
 
 fn new_listener(address: &Path) -> Result<UnixListener> {
@@ -139,7 +122,6 @@ mod tests {
     use std::path::Path;
 
     use serial_test::serial;
-    use tests_utils::gen_id;
 
     use super::*;
     use crate::Args;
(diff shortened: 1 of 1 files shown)


src/runtime-rs/crates/shim/src/shim_start.rs
```diff
@@ -105,22 +104,6 @@ impl ShimExecutor {
 
         Ok(child.id())
     }
-
-    fn get_shim_info_from_sandbox(&self, sandbox_id: &str) -> Result<(PathBuf, u32)> {
```
gkurz, line 109: One user left at https://github.com/kata-containers/kata-containers/blob/e57e1122c9850816999efdc26ebdf2bc344e64fe/src/runtime-rs/crates/shim/src/shim_start.rs#L197.

`test_get_info_from_sandbox()` should be removed as well.

pmores: Oh I missed that, good catch, thanks!

littlejawa:
This is similar to what the go shim does: write the address, but not the PID.
Neither containerd nor crio have issue with the missing PID file when running the go shim, so I think this is OK.
Now I'd really like to hear from other runtime-rs developers to confirm that we're not missing something.

src/runtime-rs/crates/shim/src/shim_start.rs
pmores: This should be fixed now.

pmores:
Force-pushed to placate clippy.  A bunch of required tests were failing but due to a k8s overconstrained allocation so that doesn't seem related to this change.

gkurz:
> Force-pushed to placate clippy. A bunch of required tests were failing but due to a k8s overconstrained allocation so that doesn't seem related to this change.

Remaining failures seem to point to some instability in AKS. Hopefully this will auto-fix at some point.

gkurz (approved):
/lgtm

Thanks @pmores !

jepio (approved):
Not a runtime-rs developer but as far as i can tell this shouldn't cause any issues.

Do we have any CRI-O tests in the CI? If not - what's the plan to add some?

gkurz:
> Not a runtime-rs developer but as far as i can tell this shouldn't cause any issues.
> 
> Do we have any CRI-O tests in the CI? If not - what's the plan to add some?

We have CRI-O tests in the CI but they are only run against the go runtime at this time. This fix might allow us to run them against runtime-rs as well.

pmores:
> Do we have any CRI-O tests in the CI? If not - what's the plan to add some?

I'm quite sure there are CRI-O tests of the go shim but none of runtime-rs (quite obviously since doesn't even launch :-)).

The plan is to add some right now and it's being worked on actually (@wainersm @ldoktor ).  This is also why this fix is important - with cri-o being unable to launch runtime-rs no cri-o/runtime-rs CI testing is possible and no development and testing on OpenShift is possible either.

lifupan (approved):
LGTM

PR #9881, opened 2024-06-19
runtime: Support policy in remote hypervisor

stevenhorsman:
Move the `sandbox.agent.setPolicy` call out of the remoteHypervisor if, block, so we can use the policy implementation on peer pods

### DIFF
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -766,19 +766,21 @@ func (k *kataAgent) startSandbox(ctx context.Context, sandbox *Sandbox) error {
 
 	if sandbox.config.HypervisorType == RemoteHypervisor {
 		ctx = context.WithValue(ctx, customRequestTimeoutKey, remoteRequestTimeout)
-	} else {
-		// Check grpc server is serving
-		if err = k.check(ctx); err != nil {
+	}
+
+	// Check grpc server is serving
+	if err = k.check(ctx); err != nil {
+		return err
+	}
+
+	// If a Policy has been specified, send it to the agent.
+	if len(sandbox.config.AgentConfig.Policy) > 0 {
+		if err := sandbox.agent.setPolicy(ctx, sandbox.config.AgentConfig.Policy); err != nil {
 			return err
 		}
+	}
 
-		// If a Policy has been specified, send it to the agent.
-		if len(sandbox.config.AgentConfig.Policy) > 0 {
-			if err := sandbox.agent.setPolicy(ctx, sandbox.config.AgentConfig.Policy); err != nil {
-				return err
-			}
-		}
-
+	if sandbox.config.HypervisorType != RemoteHypervisor {
 		// Setup network interfaces and routes
 		interfaces, routes, neighs, err := generateVCNetworkStructures(ctx, sandbox.network)
 		if err != nil {


bpradipt (approved):
/lgtm
Thanks @stevenhorsman

snir911:
LGTM, thanks Steve!

snir911 (approved):

wainersm (approved):
Great! Thanks @stevenhorsman !

wainersm:
A few jobs failed to deploy AKS (see below) and I am going to re-run them with hope in my heart that it's intermittent.
```
{
  "id": "/subscriptions/***/resourceGroups/kataCI-k8s-9881-440365b0afbd-dragonball-ubuntu-amd64-s-o",
  "location": "eastus",
  "managedBy": null,
  "name": "kataCI-k8s-9881-440365b0afbd-dragonball-ubuntu-amd64-s-o",
  "properties": {
  },
  "tags": null,
}
WARNING: The behavior of this command has been altered by the following extension: aks-preview
WARNING: SSH key files '/home/runner/.ssh/id_rsa' and '/home/runner/.ssh/id_rsa.pub' have been generated under ~/.ssh to allow SSH access to the VM. If using machines without permanent storage like Azure Cloud Shell without an attached file share, back up your keys to a safe location
ERROR: (OverconstrainedAllocationRequest) Code="OverconstrainedAllocationRequest" Message="Allocation failed. VM(s) with the following constraints cannot be allocated, because the condition is too restrictive. Please remove some constraints and try again. Constraints applied are:\n  - Networking Constraints (such as Accelerated Networking or IPv6)\n  - VM Size\n" Target="0"
Code: OverconstrainedAllocationRequest
Message: Code="OverconstrainedAllocationRequest" Message="Allocation failed. VM(s) with the following constraints cannot be allocated, because the condition is too restrictive. Please remove some constraints and try again. Constraints applied are:\n  - Networking Constraints (such as Accelerated Networking or IPv6)\n  - VM Size\n" Target="0"
Exception Details:	(OverconstrainedAllocationRequest) Allocation failed. VM(s) with the following constraints cannot be allocated, because the condition is too restrictive. Please remove some constraints and try again. Constraints applied are:
	  - Networking Constraints (such as Accelerated Networking or IPv6)
	  - VM Size
	
	Code: OverconstrainedAllocationRequest
	Message: Allocation failed. VM(s) with the following constraints cannot be allocated, because the condition is too restrictive. Please remove some constraints and try again. Constraints applied are:
	  - Networking Constraints (such as Accelerated Networking or IPv6)
	  - VM Size
	
	Target: 0
Error: Process completed with exit code 1.
```

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -797,6 +789,13 @@ func (k *kataAgent) startSandbox(ctx context.Context, sandbox *Sandbox) error {
 		kmodules = setupKernelModules(k.kmodules)
 	}
 
+	// If a Policy has been specified, send it to the agent.
```
danmihai1, line 792: I vote for keeping the SetPolicy call above the updateInterfaces, updateRoutes, and addARPNeighbors calls. That wouldn't result in a lot of duplication and would maintain the ability to possibly enforce Policy rules for the updateInterfaces, updateRoutes, and addARPNeighbors calls in the non-remote hypervisor case.

danmihai1 (changes requested):

src/runtime/virtcontainers/kata_agent.go
stevenhorsman: Sure, I can move it above the if block instead of below it

stevenhorsman: I decided to keep the separate remote hypervisor context just in case it takes a little longer to set the policy. It's added an extra if, but it's not too bad. Please take a look Dan and see if this works okay for you. Thanks!

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -766,19 +766,20 @@ func (k *kataAgent) startSandbox(ctx context.Context, sandbox *Sandbox) error {
...
+	}
 
-		// If a Policy has been specified, send it to the agent.
-		if len(sandbox.config.AgentConfig.Policy) > 0 {
-			if err := sandbox.agent.setPolicy(ctx, sandbox.config.AgentConfig.Policy); err != nil {
-				return err
-			}
+	if sandbox.config.HypervisorType != RemoteHypervisor { {
+		// Check grpc server is serving
+		if err = k.check(ctx); err != nil {
```
danmihai1, line 780: At least at a quick look, it's odd to move the "Check grpc server is serving" code after SetPolicy, because SetPolicy needs the grpc server to be working.

This code suggests to me two possibilities:
1. The k.check() call is not needed. Maybe calling SetPolicy and dealing with a possible error is good enough? If that's the case, perhaps we should remove the k.check() call.
2. The k.check() is needed before SetPolicy, so we should keep the original order of calls.

Would the k.check() call benefit the remote hypervisor code path too, before calling SetPolicy?

stevenhorsman: Good point, let me have a try in a dev environment and see if there are any issues with calling `k.check()` for remote hypervisor. Assuming not then I'm leaning towards running it for both before calling `SetPolicy` for safety.

stevenhorsman: Just to circle back on this - the check code doesn't seem to have any issues in my development environment, so I think it's work keeping.

danmihai1 (approved):
Thanks, Steve!

src/runtime/virtcontainers/kata_agent.go
stevenhorsman: Thanks so much for the valuable input in the review @danmihai1!

PR #9884, opened 2024-06-19
ci: tdx: Re-enable TDX CI

fidencio:
Now, using vanilla kubernetes, let's re-enable the TDX CI and hope it becomes more stable than it used to be.

The cleanup-snapshotter is now taking ~4 minutes, and that matches with the other platforms, mainly considering there's a sum of 210 seconds sleep in the process.


fidencio:
This one can also be merged without CI, as it changes the yaml file.

GabyCT (approved):

danmihai1 (approved):

Issue #9889, opened 2024-06-21
agent: the func scan_path may skip counting empty subdirs


gaohuatao-1:
# Description of problem
I find a bug when I test the func [scan_path](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/watcher.rs#L233).  If the input path is full of many empty subdirectories, no error will be reported even if the number of the subdirectories exceeds the limit value MAX_ENTRIES_PER_STORAGE. That is not as expected.

# Expected result
If there are only empty subdirectories under this path, no regular files, and the number of subdirectories exceeds the limit,
The func scan_path should return MAX_ENTRIES_PER_STORAGE error.

# Actual result
No MAX_ENTRIES_PER_STORAGE error returned.

# Further information
It is obvious that [scan_path](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/watcher.rs#L233) have a  
 bug.

PR #9890, opened 2024-06-21
agent: scan_path should not skip counting empty dirs

gaohuatao-1:
Fixes:#9889


gaohuatao-1:
@justxuewei @Apokleos  Hi, Could you please give me a test?

Issue #9758, opened 2024-05-31
runtime: Add support for directly attachable network


justxuewei:
The runtime-rs has already implemented a directly attachable network by https://github.com/kata-containers/kata-containers/pull/7383 to fix https://github.com/kata-containers/kata-containers/issues/1922. It could be ported to the runtime for customizing container networks.

l8huang:
please assign it to me

l8huang:
@justxuewei @lifupan  could you please review https://github.com/kata-containers/kata-containers/pull/9868?

l8huang:
Considering VFIO network device need a driver to be claimed in guest VM, should I raise a PR(like https://github.com/kata-containers/kata-containers/pull/6595) to add a pre-built kernel which includes the netdev driver(e.g. https://github.com/kata-containers/kata-containers/pull/9620) into kata-deploy image? I'm not sure if there are concerns regarding the size of the kata-deploy image.

CC @zvonkok

PR #9895, opened 2024-06-24
agent: add InterfaceMatcher

l8huang:
When a VFIO network device is plugged into a guest VM, it takes time for the driver to recognize the device and create its network interface. The InterfaceMatcher is required to ensure that the network interface is present before applying network configuration.

Fixes: #9758

### DIFF
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -32,7 +32,10 @@ fn sl() -> slog::Logger {
     slog_scope::logger().new(o!("subsystem" => "device"))
 }
 
+// uevent subsystems
 const BLOCK: &str = "block";
+const NET: &str = "net";
+
 pub const DRIVER_9P_TYPE: &str = "9p";
 pub const DRIVER_VIRTIOFS_TYPE: &str = "virtio-fs";
 pub const DRIVER_BLK_PCI_TYPE: &str = "blk";
@@ -377,6 +380,32 @@ pub async fn wait_for_pci_device(
         .ok_or_else(|| anyhow!("Bad device path {:?} in uevent", &uev.devpath))?;
     let addr = pci::Address::from_str(addr)?;
     Ok(addr)
+}
+
+#[derive(Debug)]
+struct InterfaceMatcher {
+    interface: String,
+}
+
+impl InterfaceMatcher {
+    fn new(name: &str) -> InterfaceMatcher {
+        InterfaceMatcher {
+            interface: name.into(),
+        }
+    }
+}
+
+impl UeventMatcher for InterfaceMatcher {
+    fn is_match(&self, uev: &Uevent) -> bool {
+        uev.action == "add" && uev.subsystem == NET && uev.interface == self.interface
+    }
+}
+
+pub async fn wait_for_interface(sandbox: &Arc<Mutex<Sandbox>>, name: &str) -> Result<()> {
+    let matcher = InterfaceMatcher::new(name);
+    wait_for_uevent(sandbox, matcher).await?;
+
+    Ok(())
 }
 
 #[derive(Debug)]
@@ -1830,5 +1859,30 @@ mod tests {
         let mut uev_other_device = uev.clone();
         uev_other_device.devpath = format!("{}/card{}/{}.0002", AP_ROOT_BUS_PATH, card, card);
         assert!(!matcher.is_match(&uev_other_device));
+    }
+
+    #[tokio::test]
+    #[allow(clippy::redundant_clone)]
+    async fn test_interface_matcher() {
+        let iface1 = "eth0";
+        let mut uev_a = crate::uevent::Uevent::default();
+        uev_a.action = crate::linux_abi::U_EVENT_ACTION_ADD.to_string();
+        uev_a.devpath = format!(
+            "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/{}",
+            iface1
+        );
+        uev_a.subsystem = NET.into();
+        uev_a.interface = iface1.into();
+        let matcher_a = InterfaceMatcher::new(iface1);
+
+        let iface2 = "net1";
+        let mut uev_b = uev_a.clone();
+        uev_b.interface = iface2.into();
+        let matcher_b = InterfaceMatcher::new(iface2);
+
+        assert!(matcher_a.is_match(&uev_a));
+        assert!(matcher_b.is_match(&uev_b));
+        assert!(!matcher_b.is_match(&uev_a));
+        assert!(!matcher_a.is_match(&uev_b));
     }
 }
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -52,7 +52,9 @@ use nix::sys::{stat, statfs};
 use nix::unistd::{self, Pid};
 use rustjail::process::ProcessOperations;
 
-use crate::device::{add_devices, get_virtio_blk_pci_device_name, update_env_pci};
+use crate::device::{
+    add_devices, get_virtio_blk_pci_device_name, update_env_pci, wait_for_interface,
+};
 use crate::features::get_build_features;
 use crate::linux_abi::*;
 use crate::metrics::get_metrics;
@@ -121,6 +123,8 @@ const ERR_NO_SANDBOX_PIDNS: &str = "Sandbox does not have sandbox_pidns";
 // filesystem lock. Based on this, 5 seconds seems a resonable timeout period in case the lock is
 // not available.
 const IPTABLES_RESTORE_WAIT_SEC: u64 = 5;
+
+const IFACE_TYPE_VFIO: &str = "vfio";
 
 // Convenience function to obtain the scope logger.
 fn sl() -> slog::Logger {
@@ -914,6 +918,16 @@ impl agent_ttrpc::AgentService for AgentService {
             ttrpc::Code::INVALID_ARGUMENT,
             "empty update interface request",
         )?;
+
+        // It may be necessary to wait for the network interface to become available
+        // before configuring it. This is particularly true for VFIO devices, as
+        // the guest kernel may need additional time to recognize and initialize
+        // them as network interfaces.
+        if interface.type_() == IFACE_TYPE_VFIO {
+            wait_for_interface(&self.sandbox, interface.name())
+                .await
+                .map_ttrpc_err(|e| format!("interface not available: {:?}", e))?;
+        }
 
         self.sandbox
             .lock()


src/agent/src/device.rs
```diff
@@ -51,6 +51,7 @@ pub const DRIVER_VFIO_PCI_TYPE: &str = "vfio-pci";
 pub const DRIVER_VFIO_AP_TYPE: &str = "vfio-ap";
 pub const DRIVER_OVERLAYFS_TYPE: &str = "overlayfs";
 pub const FS_TYPE_HUGETLB: &str = "hugetlbfs";
+const NET: &str = "net";
```
Apokleos, line 54: It needs move this L54 to the place following the L35 BLOCK subsystem. And it will looks better if some comment about it. I don't know what's the meaning of the `NET` in first sight.

Apokleos (commented):
Thx @l8huang. I want to make it more clear that, is there failure physical net device cases which use vfio devices not coming up in time causing timeout event? I think you now add a dedicated method  only for network with VFIO passthrough.

l8huang:
@Apokleos 
The VFIO network device can be correctly recognized by its driver, and the corresponding eth0 interface can also be created as expected without any failures. However, this process takes some time to complete, which results in when update_interface() being called,  it fails due to eth0 interface doesn't exist.

src/agent/src/device.rs
l8huang: will do

amshinde:
Hey @l8huang, I was debugging a couple of network related issues when running kata with nerdctl.
In case of nerdctl network is hotplugged, and we have seen in the CI that the network interface is not found in the update_interface call. While I have not been able to reproduce this locally, I have seen this happening quite frequently in the CI.
My guess was that the network hotplug was taking longer on the systems causing the interface to be not found. 

I just raised a PR that waits for the interface and just realised my PR is quite similar to yours, just resolves the issue for the nerdctl case. Please take a look at this commit:
https://github.com/kata-containers/kata-containers/pull/9899/commits/7f7a2ce27fd31e2bb210d5ca637690a51018f998

I had a good look at your PR and I see that you are waiting on the interface based on the name. While this works in most cases (esp after we have disabled systemd from renaming the interfaces), relying on the interface name in the VM is not really reliable, especially when multiple network interfaces are passed to Kata, the order in which the interfaces are scanned on the host and  then added to the VM may differ each time causing the name to change in the VM. 
I have based my wait method on the PCI path of the interface instead. Can you please take a look and let me know if the interface matcher would work in your scenario?

l8huang:
@amshinde thanks for reviewing this PR.

I looked at your commit, it doesn't work in my scenario, the Uevents observed on my env are:

```
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0", devname: "", subsystem: "pci", seqnum: "788", interface: "" }"
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/ptp/ptp1", devname: "ptp1", subsystem: "ptp", seqnum: "789", interface: "" }"
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/mlx5_core.eth.0", devname: "", subsystem: "auxiliary", seqnum: "790", interface: "" }"
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0", devname: "", subsystem: "net", seqnum: "791", interface: "eth0" }"
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0/queues/rx-0", devname: "", subsystem: "queues", seqnum: "792", interface: "" }"
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0/queues/tx-0", devname: "", subsystem: "queues", seqnum: "793", interface: "" }"
"event":"Uevent { action: "bind", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/mlx5_core.eth.0", devname: "", subsystem: "auxiliary", seqnum: "794", interface: "" }"
```

The Uevent this PR relies on is:
```
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0", devname: "", subsystem: "net", seqnum: "791", interface: "eth0" }"
```

This PR is part of supporting VFIO network interface via DAN. The DAN config is from ovn-k8s CNI, it defines the interface name(primary is eth0, secondary interfaces are named net1, net2, ...)  and network config, but it doesn't know the network interface's guest PCI address and the devpath format `/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0`. 

For a single network interface case this change works, as you mentioned, it may not reliable for multiple interfaces case, therefor, an additional interface renaming step could be required -- I know a KubeVirt VM use case which have udev rules to naming interface based on MAC address. The multiple interface support can be added by follow up PRs.

Apokleos:
Hi @l8huang @amshinde Thx for all your PRs, And IMO, to address the problem of `update_inteface` call failure, we need take several aspect into account:  it can cover scenarios, such as physical net device with vfio passthrough, virtio-net-pci and virtio-net-mmio, and it's a more generic solution, and it also can address both single interface and multiple interfaces cases.

Apokleos:
> I know a KubeVirt VM use case which have udev rules to naming interface based on MAC address. The multiple interface support can be added by follow up PRs.
>
Could you please provide more detailed info or link about it to help us to check it ? Thx

Apokleos:
> @Apokleos
> 
> > is there failure physical net device cases which use vfio devices not coming up in time causing timeout event?
> 
> Nope. The VFIO network device can be correctly recognized by its driver, and the corresponding eth0 interface can also be created as expected without any failures. However, this process takes some time to complete, which results in when update_interface() being called, it fails due to eth0 interface doesn't exist.
> 
> > I think you now add a dedicated method only for network with VFIO passthrough.
> 
> It's not specific to VFIO, it's generic for network interface which may take time to be created -- currently only VFIO device has this issue.
>
Yeah, expect to make it more generic for virtio-net-pci, virtio-net-mmio (even for virtio-net-ccw but not sure) network device but not only for vfio net device

l8huang:
> > I know a KubeVirt VM use case which have udev rules to naming interface based on MAC address. The multiple interface support can be added by follow up PRs.
> 
> Could you please provide more detailed info or link about it to help us to check it ? Thx

In KubeVirt use case, it relies on cloud-init to generate `/etc/udev/rules.d/70-persistent-net.rules`, e.g.:
```
# cat  /etc/udev/rules.d/70-persistent-net.rules 
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="0a:58:0a:00:00:01", NAME="eth0"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="0a:58:0a:00:00:02", NAME="net1"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="0a:58:0a:00:00:03", NAME="net2"
```

 related code is at https://github.com/canonical/cloud-init/blob/main/cloudinit/net/udev.py#L37-L45

amshinde:
@l8huang The uevent that you mentioned:
```
"event":"Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0", devname: "", subsystem: "net", seqnum: "791", interface: "eth0" }"
```
is precisely the one that my PR should catch.
Looks like for your DAN case, the device is attached on the PCI bus, so you should be able to figure our what the pci path on the guest would be.
In fact today with Kata, any pci device attached on the pci bus/bridge should get a predictable PCI path. If this is missing, then maybe that needs to be added. 

In any case I will special case my wait_for_interface method to only wait when pci-path for the device is present.

l8huang:
Using pci address requires an host-to-guest pci address translation, which complicates the overall logic.

The cloud-init uses MAC addresses to verify the presence of interfaces, which appears to be a more elegant solution, more details are in https://github.com/kata-containers/kata-containers/issues/9931.

l8huang:
@amshinde I tested your PR https://github.com/kata-containers/kata-containers/pull/9899, added some findings to there, PTAL.

I tested a Pod with two VFIO devices(expected name are eth0 and net1), the links were renamed as below: 
- rename link 2 from eth0 to eth0_temp
- rename link 3 from eth1 to eth0
- rename link 2 name to eth1
- rename link 2 from eth1 to net1

The related Uevents are:
```
"Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0_temp", devname: "", subsystem: "net", seqnum: "790", interface: "eth0_temp" }"}"
"Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0_temp", devname: "", subsystem: "net", seqnum: "790", interface: "eth0_temp" }"}"
"Uevent { action: "bind", devpath: "/devices/pci0000:00/0000:00:06.0/0000:03:00.0", devname: "", subsystem: "pci", seqnum: "791", interface: "" }"}"
"Uevent { action: "bind", devpath: "/devices/pci0000:00/0000:00:06.0/0000:03:00.0", devname: "", subsystem: "pci", seqnum: "791", interface: "" }"}"
"Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:06.0/0000:03:00.0/net/eth0", devname: "", subsystem: "net", seqnum: "792", interface: "eth0" }"}"
"Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:06.0/0000:03:00.0/net/eth0", devname: "", subsystem: "net", seqnum: "792", interface: "eth0" }"}"

"Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth1", devname: "", subsystem: "net", seqnum: "793", interface: "eth1" }"}"
"Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth1", devname: "", subsystem: "net", seqnum: "793", interface: "eth1" }"}"

Uevent { action: "move", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/net1", devname: "", subsystem: "net", seqnum: "794", interface: "net1" }"}"
```

Considering link renaming is an async call, I think the `InterfaceMatcher` should be added to ensure that renaming being completed before proceed to update_routes() -- I can change the match condition of action to "move" to match the renaming Uevent. @Apokleos what do you think?

PR #9898, opened 2024-06-24
ci: Add scheduled job to cleanup resources

sprt:
```
This is the first part of adding a job to clean up potentially dangling
Azure resources. This will be based on Jeremi's tool from
https://github.com/jepio/kata-azure-automation.

At first, we'll only clean up AKS clusters, as this is what has been
causing us problems lately, but this could very well be extended to
cleaning up entire resource groups, which is why I left the different
names pretty generic (i.e. "resources" instead of "clusters").
```


.github/workflows/cleanup-resources.yaml
```diff
@@ -0,0 +1,13 @@
+name: Cleanup dangling Azure resources
+on:
+  schedule:
+    - cron: "*/15 * * * *"
```
stevenhorsman, line 4: Do you think every 15min is the right cadence (assuming my cron syntax is correct). It feels like it might be unnecessarily frequent for a cleanup job that hopefully won't have much to do too often

stevenhorsman:
Given that this PR doesn't actually do anything, should we risk just merging it without running the tests against it?

.github/workflows/cleanup-resources.yaml
sprt: The job would be triggered every 15 mins but would only actually check resources that were created something like >=2 hours ago, as done below. That way resources stay alive at most 2h15min.

https://github.com/jepio/kata-azure-automation/blob/a4113f68ca31da5f36afda0ddac85db56810cfd4/autodelete-aks/function_app.py#L36

sprt:
> Given that this PR doesn't actually do anything, should we risk just merging it without running the tests against it?

Yes please! I had forgotten about the `force-skip-ci` flag, but AFAIK this doesn't affect the actual tests, does it? Not sure I have the permission to bypass those.

stevenhorsman:
> > Given that this PR doesn't actually do anything, should we risk just merging it without running the tests against it?
> 
> Yes please! I had forgotten about the `force-skip-ci` flag, but AFAIK this doesn't affect the actual tests, does it? Not sure I have the permission to bypass those.

Yeah, once we have approval I can do that (and I've raised the point on the AC meeting)

.github/workflows/cleanup-resources.yaml
fidencio: Please, make this running every 2 hours, if you want to be really cautious about the resources leaking.  But every 15 minutes is a bit excessive, IMHO.

sprt: Changed it to every 6 hours so that the GHA UI doesn't accumulate more than 4 runs per day.

stevenhorsman: Thanks!

stevenhorsman (approved):
LGTM. Thanks @sprt

fidencio (approved):
lgtm, thanks @sprt!

PR #9899, opened 2024-06-25
Fix issue while adding multiple networks with nerdctl

amshinde:
agent: Bring interface down before renaming it.
    
In case we are dealing with multiple interfaces and there exists a
network interface with a conflicting name, we temporarily rename it to
avoid name conflicts.
Before doing this, we need to rename bring the interface down.
Failure to do so results in netlink returning Resource busy errors.
    
The resource needs to be down for subsequent operation when the name is
swapped back as well.
    
This solves the issue of passing multiple networks in case of nerdctl
as:
nerdctl run --rm  --net foo --net bar docker.io/library/busybox:latest ip a

Fixes: #9900    

### DIFF
--- a/src/agent/src/device.rs
+++ b/src/agent/src/device.rs
@@ -380,6 +380,65 @@ pub async fn wait_for_pci_device(
 }
 
 #[derive(Debug)]
+struct NetPciMatcher {
+    devpath: String,
+}
+
+impl NetPciMatcher {
+    fn new(relpath: &str) -> NetPciMatcher {
+        let root_bus = create_pci_root_bus_path();
+
+        NetPciMatcher {
+            devpath: format!("{}{}", root_bus, relpath),
+        }
+    }
+}
+
+impl UeventMatcher for NetPciMatcher {
+    fn is_match(&self, uev: &Uevent) -> bool {
+        uev.devpath.starts_with(self.devpath.as_str())
+            && uev.subsystem == "net"
+            && !uev.interface.is_empty()
+            && uev.action == "add"
+    }
+}
+
+pub async fn wait_for_net_interface(
+    sandbox: &Arc<Mutex<Sandbox>>,
+    pcipath: &pci::Path,
+) -> Result<()> {
+    let root_bus_sysfs = format!("{}{}", SYSFS_DIR, create_pci_root_bus_path());
+    let sysfs_rel_path = pcipath_to_sysfs(&root_bus_sysfs, pcipath)?;
+
+    let matcher = NetPciMatcher::new(&sysfs_rel_path);
+
+    // Check if the interface is already added in case network is cold-plugged
+    // or the uevent loop is started before network is added.
+    // We check for the pci deive in the sysfs directory for network devices.
+    let pattern = format!(
+        r"[./]+{}/[a-z0-9/]*net/[a-z0-9/]*",
+        matcher.devpath.as_str()
+    );
+    let re = Regex::new(&pattern).expect("BUG: Failed to compile regex for NetPciMatcher");
+
+    for entry in fs::read_dir(SYSFS_NET_PATH)? {
+        let entry = entry?;
+        let path = entry.path();
+        let target_path = fs::read_link(path)?;
+        let target_path_str = target_path
+            .to_str()
+            .ok_or_else(|| anyhow!("Expected symlink in dir {}", SYSFS_NET_PATH))?;
+
+        if re.is_match(target_path_str) {
+            return Ok(());
+        }
+    }
+    let _uev = wait_for_uevent(sandbox, matcher).await?;
+
+    Ok(())
+}
+
+#[derive(Debug)]
 struct VfioMatcher {
     syspath: String,
 }
@@ -1689,6 +1748,41 @@ mod tests {
         assert!(matcher_b.is_match(&uev_b));
         assert!(!matcher_b.is_match(&uev_a));
         assert!(!matcher_a.is_match(&uev_b));
+    }
+
+    #[tokio::test]
+    #[allow(clippy::redundant_clone)]
+    async fn test_net_pci_matcher() {
+        let root_bus = create_pci_root_bus_path();
+        let relpath_a = "/0000:00:02.0/0000:01:01.0";
+
+        let mut uev_a = crate::uevent::Uevent::default();
+        uev_a.action = crate::linux_abi::U_EVENT_ACTION_ADD.to_string();
+        uev_a.devpath = format!("{}{}", root_bus, relpath_a);
+        uev_a.subsystem = String::from("net");
+        uev_a.interface = String::from("eth0");
+        let matcher_a = NetPciMatcher::new(relpath_a);
+        println!("Matcher a : {}", matcher_a.devpath);
+
+        let relpath_b = "/0000:00:02.0/0000:01:02.0";
+        let mut uev_b = uev_a.clone();
+        uev_b.devpath = format!("{}{}", root_bus, relpath_b);
+        let matcher_b = NetPciMatcher::new(relpath_b);
+
+        assert!(matcher_a.is_match(&uev_a));
+        assert!(matcher_b.is_match(&uev_b));
+        assert!(!matcher_b.is_match(&uev_a));
+        assert!(!matcher_a.is_match(&uev_b));
+
+        let relpath_c = "/0000:00:02.0/0000:01:03.0";
+        let net_substr = "/net/eth0";
+        let mut uev_c = uev_a.clone();
+        uev_c.devpath = format!("{}{}{}", root_bus, relpath_c, net_substr);
+        let matcher_c = NetPciMatcher::new(relpath_c);
+
+        assert!(matcher_c.is_match(&uev_c));
+        assert!(!matcher_a.is_match(&uev_c));
+        assert!(!matcher_b.is_match(&uev_c));
     }
 
     #[tokio::test]
--- a/src/agent/src/netlink.rs
+++ b/src/agent/src/netlink.rs
@@ -134,14 +134,33 @@ impl Handle {
             .arp(iface.raw_flags & libc::IFF_NOARP as u32 == 0)
             .up()
             .execute()
-            .await?;
+            .await
+            .map_err(|err| {
+                anyhow!(
+                    "Failure in LinkSetRequest for interface {}: {}",
+                    iface.name.as_str(),
+                    err
+                )
+            })?;
 
         // swap the updated iface's name.
         if let Some(nlink) = new_link {
             let mut request = self.handle.link().set(nlink.index());
             request.message_mut().header = nlink.header.clone();
 
-            request.name(link.name()).up().execute().await?;
+            request
+                .name(link.name())
+                .up()
+                .execute()
+                .await
+                .map_err(|err| {
+                    anyhow!(
+                        "Error swapping back interface name {} to {}: {}",
+                        nlink.name().as_str(),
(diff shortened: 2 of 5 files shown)


amshinde:
Todo: I need to add an integration test for this with nerdctl, will add it to this PR itself.

src/agent/src/rpc.rs
```diff
@@ -915,6 +918,13 @@ impl agent_ttrpc::AgentService for AgentService {
             "empty update interface request",
         )?;
 
+        let pcipath = pci::Path::from_str(&interface.pciPath)
+            .map_ttrpc_err(|e| format!("Unexpected pci-path for network interface: {:?}", e))?;
+
+        wait_for_net_interface(&self.sandbox, &pcipath)
+            .await
+            .map_ttrpc_err(|e| format!("interface not available: {:?}", e))?;
+
```
Apokleos, line 927: @amshinde AFAIK, currently we support both virtio-pci and virtio-mmio bus in kata. this is for virtio-pci but there's no pci path existing in virtio-mmio bus, we need take this into account, right ?

Apokleos: And as [the comment](https://github.com/kata-containers/kata-containers/pull/9895#issuecomment-2190005113) said, it seems not work for net device with vfio passthrough.

src/agent/src/netlink.rs
```diff
@@ -95,16 +95,29 @@ impl Handle {
...
+                }
+
                 // update the existing interface name with a temporary name, otherwise
                 // it would failed to udpate this interface with an existing name.
                 let mut request = self.handle.link().set(link.index());
                 request.message_mut().header = link.header.clone();
+                let link_name = link.name();
 
                 request
                     .name(format!("{}_temp", link.name()))
```
pmores, line 110: Total nit: now that we have `link_name` we could use it here as well.

pmores (approved):
lgtm, thanks @amshinde ! (Full disclosure, admittedly I can't really vet the third commit of this series.)

src/agent/src/device.rs
```diff
@@ -379,6 +379,48 @@ pub async fn wait_for_pci_device(
...
+    }
+}
+
+impl UeventMatcher for NetPciMatcher {
+    fn is_match(&self, uev: &Uevent) -> bool {
+        uev.devpath == self.devpath && uev.subsystem == "net" && !uev.interface.is_empty()
+    }
+}
+
+pub async fn wait_for_net_interface(
```
jodh-intel, line 402: Could you add a few unit tests for this function? The final call to ` wait_for_uevent()` would probably be too awkward to test, but the remaining parts should be possible, I think?

jodh-intel (commented):
Thanks, @amshinde. A few comments...

src/agent/src/rpc.rs
amshinde: @Apokleos Ack, I tested this with qemu and clh where pci bus is in use. I will special case this to pci case only. Will have to deal with virtio-mmio separately, maybe I can address that in a follow-up PR.

src/agent/src/device.rs
```diff
@@ -379,6 +379,48 @@ pub async fn wait_for_pci_device(
     Ok(addr)
 }
 
+#[derive(Debug)]
+struct NetPciMatcher {
+    devpath: String,
```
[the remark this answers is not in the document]
amshinde: @jodh-intel This is actually a relative path and I thought of making this a PathBuf initially but decided to keep this as String to align with other declarations/usages in the same file.  Maybe I can have a separate PR to change all usages but decided to go with String to keep changes minimal for now.

src/agent/src/device.rs
```diff
@@ -379,6 +379,48 @@ pub async fn wait_for_pci_device(
...
+        let root_bus = create_pci_root_bus_path();
+        Ok(NetPciMatcher {
+            devpath: format!("{}{}", root_bus, relpath),
+        })
+    }
+}
+
+impl UeventMatcher for NetPciMatcher {
+    fn is_match(&self, uev: &Uevent) -> bool {
+        uev.devpath == self.devpath && uev.subsystem == "net" && !uev.interface.is_empty()
```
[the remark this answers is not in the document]
amshinde: `uev.devpath` is received from kernel netlink messages and devpath is the one we construct. So yes both are guaranteed to be absolute with no multile slashes. Also `self.devpath`  will not be emppty for a pci device which is the case here.

src/agent/src/device.rs
amshinde: Done

src/agent/src/device.rs
```diff
@@ -379,6 +379,48 @@ pub async fn wait_for_pci_device(
...
+#[derive(Debug)]
+struct NetPciMatcher {
+    devpath: String,
+}
+
+impl NetPciMatcher {
+    fn new(relpath: &str) -> Result<NetPciMatcher> {
+        let root_bus = create_pci_root_bus_path();
+        Ok(NetPciMatcher {
+            devpath: format!("{}{}", root_bus, relpath),
```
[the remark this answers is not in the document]
amshinde: Fixed

src/agent/src/netlink.rs
```diff
@@ -95,16 +95,29 @@ impl Handle {
...
-                    .up()
                     .execute()
-                    .await?;
+                    .await
+                    .map_err(|err| {
+                        anyhow!(
+                            "Failed to rename interface {} to {}_temp with error: {}",
```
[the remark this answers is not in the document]
amshinde: Done

amshinde:
@jodh-intel I have addressed your comments. Please take a look.

amshinde:
@Apokleos Can you take another look at this?

src/agent/src/device.rs
```diff
@@ -379,6 +379,51 @@ pub async fn wait_for_pci_device(
...
+    pcipath: &pci::Path,
+) -> Result<()> {
+    let root_bus_sysfs = format!("{}{}", SYSFS_DIR, create_pci_root_bus_path());
+    let sysfs_rel_path = pcipath_to_sysfs(&root_bus_sysfs, pcipath)?;
+
+    let matcher = NetPciMatcher::new(&sysfs_rel_path);
+
+    // Check if the interface is already added in case network is cold-plugged
+    // or the uevent loop is started before network is added.
+    let sys_path = Path::new("/sys").join(matcher.devpath.as_str()).canonicalize()?;
```
l8huang, line 416: `mather.devpath` is an absolute path like "/devices/pci0000:00/...", so `join()` will use it to replace `/sys`.

l8huang (commented):
cherry-picked and tested this PR, found some bugs which caused by `join()` function.

src/agent/src/device.rs
```diff
@@ -379,6 +379,51 @@ pub async fn wait_for_pci_device(
...
 
+#[derive(Debug)]
+struct NetPciMatcher {
+    devpath: String,
+}
+
+impl NetPciMatcher {
+    fn new(relpath: &str) -> NetPciMatcher {
+        let root_bus = create_pci_root_bus_path();
+        let path = Path::new(root_bus.as_str()).join(relpath);
```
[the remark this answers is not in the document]
amshinde: @l8huang Thanks for pointing this out, I did need to test the code a bit after making the changes. 
Looks like the `join()` method on paths in rust is  different from the golang one in this aspect when the second path is absolute.
Just came accross this issue railsed a while ago:
https://github.com/rust-lang/rust/issues/16507

In light of this, I have got rid of the `join()` calls and resorted to string concatenation instead.

@jodh-intel Please note, I need to to go back to string concatenation as the existing method `pcipath_to_sysfs` that I used to get the syspath returns an abosolute path. We will need to change it to return relative path and change the calls in a number of places which warrants a huge change worthy of a separate PR itself.

src/agent/src/device.rs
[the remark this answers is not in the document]
amshinde: @l8huang I have made this change. Using `starts_with` now. This should satisfy the VFIO case as well.

amshinde:
@l8huang I have made changes to the PR based on your suggestions. The NetPciMatcher should now work in vfio case as well. I havent had chance to test with vfio yet as I dont have access to a net device that I can passthrough, trying to get access to one. Meanwhile would appreciate it if you could give this PR a try as well.

src/agent/src/device.rs
```diff
@@ -379,6 +379,53 @@ pub async fn wait_for_pci_device(
...
+
+    let matcher = NetPciMatcher::new(&sysfs_rel_path);
+
+    // Check if the interface is already added in case network is cold-plugged
+    // or the uevent loop is started before network is added.
+    let sys_path_str = format!("{}{}", SYSFS_DIR, matcher.devpath.as_str());
+    let path = Path::new(sys_path_str.as_str());
+
+    if path.exists() {
+        return Ok(());
```
l8huang, line 421: For VFIO network device, the device is hot-plugged earlier, so the path exists at this point, this function returns OK here and `wait_for_uevent()` will not be called.  This caused update_interfaces() started to configure interface before 
```
Uevent { action: "add", devpath: "/devices/pci0000:00/0000:00:05.0/0000:02:00.0/net/eth0", devname: "", subsystem: "net", seqnum: "778", interface: "eth0" }
```
and failed with
```
error="rpc error: code = Internal desc = \"update interface: Link not found (Address: 0a:58:0a:c0:02:39)\""    
```

After comment out  `if path.exists() {...}` block, it works for VFIO device. Is this check added as a shortcut to avoid unnecessarily calling `wait_for_uevent()` in certain cases? If so, it can be removed.

l8huang (commented):
Thanks for addressing my comments. I tested your change and found a issue, please take a look at my comment.

src/agent/src/device.rs
amshinde: @l8huang This check is needed for virtio devices that may be added much before or are coldplugged instead of hotplugged. 
I am checking for "/sys/devices/pci0000:00/0000:00:05.0/0000:02:00.0" here, but ideally should check for path that matches something like:
"devices/pci0000:00/0000:00:05.0/0000:02:00.0/{virtio*}*/net/eth*}".
I need to test the vfio case on my side a bit more.  
Wanted to know a bit about your setup so I can replicate this on my end. The above error occurred when you passed a vfio device with sriov device plugin+k8s?

l8huang: > This check is needed for virtio devices that may be added much before or are coldplugged instead of hotplugged.

VFIO device can only be hot-plugged now. Probably this can be resolved by differentiating the check based on the interface type.

>  The above error occurred when you passed a vfio device with sriov device plugin+k8s?

yes

We have an in-house ovn-k8s which includes change for generation DAN config, that change has not been upstreamed yet. Basically I'm using a Nvidia Connect-X6 network adapter, VFIO devices are allocated from it, [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin) and ovn-k8s are used to setup pod network.

amshinde: Pushed a potential fix. Need to test this more with sriov plugin, may have to come back to this in a couple of days.

l8huang: I tested it in my environment, and it works for VFIO -- the wait_for_uevent() function was called with NetPciMatcher. 

This PR looks good to me, thanks for addressing this issue 👍

amshinde: @l8huang Thanks for testing this out.

jodh-intel (approved):
Thanks, @amshinde.

PR #9906, opened 2024-06-25
tests: attestation: Restrict sample policy use

stevenhorsman:
- We only want to enable the sample verifier in the KBS for non-TEE tests, so prevent an edge case where the TEE platform isn't set up correctly and we might fall back to the sample and get false positives. To prevent this we add guards around the sample policy enablement and only run it for non confidential hardware

### DIFF
--- a/tests/integration/kubernetes/k8s-confidential-attestation.bats
+++ b/tests/integration/kubernetes/k8s-confidential-attestation.bats
@@ -47,7 +47,10 @@ setup() {
 }
 
 @test "Get CDH resource" {
-	kbs_set_allow_all_resources
+	if ! is_confidential_hardware; then
+		kbs_set_allow_all_resources
+	fi
+
 	kubectl apply -f "${K8S_TEST_YAML}"
 
 	# Retrieve pod name, wait for it to come up, retrieve pod ip


fidencio (approved):
lgtm, thanks @stevenhorsman!

fidencio:
Befoire merging, I'd like to run this locally later Today, just to avoid burning CI cycles.

tests/integration/kubernetes/k8s-confidential-attestation.bats
```diff
@@ -64,7 +67,10 @@ setup() {
 }
 
 @test "Cannot get CDH resource when deny-all policy is set" {
-	kbs_set_deny_all_resources
+	if ! is_confidential_hardware; then
+		kbs_set_deny_all_resources
+	fi
```
fitzthum, line 72: I think `kbs_set_deny_all_resources` should still be called for all platforms. Otherwise I don't think this test will pass.

fidencio: Very good catch!

fitzthum (approved):
LGTM

fidencio:
TDX failure:
```
#   Warning  Failed   2s (x8 over 82s)  kubelet  Error: failed to create containerd container: create instance 4879: object with key "4879" already exists: unknown`
```

I will re-run after the other tests finish, but we're hitting this more often than I'd like to see it happening.

wainersm (approved):
LGTM!

wainersm:
I realized our confidential tests have too many boilerplates. I'm planning to refactor the tests to reduce that, but first wait the PR for registry authentication and sealed secrets land.

sprt (approved):

fidencio:
TDX CDH test passed, I will merge it after the TDX CI finishes.

PR #9909, opened 2024-06-25
ci: Add scheduled job to cleanup resources, pt. II 

sprt:
```
Follow-up to #9898 and final PR of this set. This implements the actual
deletion logic.
```

Tested here as a dry run (nothing was actually deleted): https://github.com/kata-containers/kata-containers/actions/runs/9670071327/job/26678039258 Interestingly, that shows 12 dangling clusters.

### DIFF
--- a/.github/workflows/cleanup-resources.yaml
+++ b/.github/workflows/cleanup-resources.yaml
@@ -7,7 +7,25 @@ on:
 jobs:
   cleanup-resources:
     runs-on: ubuntu-latest
-    secrets: inherit
     steps:
+      - uses: actions/checkout@v4
+
+      - name: Log into Azure
+        env:
+          AZ_APPID: ${{ secrets.AZ_APPID }}
+          AZ_PASSWORD: ${{ secrets.AZ_PASSWORD }}
+          AZ_TENANT_ID: ${{ secrets.AZ_TENANT_ID }}
+          AZ_SUBSCRIPTION_ID: ${{ secrets.AZ_SUBSCRIPTION_ID }}
+        run: bash tests/integration/kubernetes/gha-run.sh login-azure
+
+      - name: Install Python dependencies
+        run: |
+          pip3 install --user --upgrade \
+            azure-identity==1.16.0 \
+            azure-mgmt-resource==23.0.1
+
       - name: Cleanup resources
+        env:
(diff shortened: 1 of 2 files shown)


fidencio (approved):
lgtm, thanks @sprt!

tests/cleanup_resources.py
```diff
@@ -0,0 +1,49 @@
...
+    # Ignore resources created after the cutoff time (i.e. less than
+    # `cleanup_after` time ago).
+    if resource.created_time > cleanup_cutoff_time:
+        print(f"{resource.name}: created at {resource.created_time}, after delta cutoff, ignored")
+        continue
+
+    print(f"{resource.name}: created at {resource.created_time}, before delta cutoff, deleting...")
+
+    # XXX DANGER ZONE: Delete the resource.
+    client.resources.begin_delete_by_id(resource.id)
```
jepio, line 44: We create a unique resource group per cluster - should we be cleaning that up as well?

.github/workflows/cleanup-resources.yaml
```diff
@@ -7,7 +7,25 @@ on:
...
     steps:
+      - uses: actions/checkout@v4
+
+      - name: Log into Azure
+        env:
+          AZ_APPID: ${{ secrets.AZ_APPID }}
+          AZ_PASSWORD: ${{ secrets.AZ_PASSWORD }}
+          AZ_TENANT_ID: ${{ secrets.AZ_TENANT_ID }}
+          AZ_SUBSCRIPTION_ID: ${{ secrets.AZ_SUBSCRIPTION_ID }}
+        run: bash tests/integration/kubernetes/gha-run.sh login-azure
```
jepio, line 19: i would love to see us use federated credentials some day

tests/cleanup_resources.py
sprt: Good point, just added code to do this as long as it's the only resource in the resource group.

stevenhorsman (approved):
I'm not and expert in the Azure python APIs, but the code LGTM. Thanks @sprt

fidencio (approved):
lgtm, thanks @sprt!

stevenhorsman:
This is a change to a workflow that isn't tested by the PR checks as it is not triggered with a on/pull_request. It also has at least two reviews with the force-skip-ci label, therefore we can skip the CI and use admin rights to merge this.

PR #9911, opened 2024-06-25
genpolicy: deny UpdateEphemeralMountsRequest

Redent0r:
* genpolicy: deny UpdateEphemeralMountsRequest

Deny UpdateEphemeralMountsRequest by default, because paths to critical Guest components can be redirected using such request.

Upstreaming from https://github.com/microsoft/kata-containers/pull/126

### DIFF
--- a/src/tools/genpolicy/rules.rego
+++ b/src/tools/genpolicy/rules.rego
@@ -39,7 +39,7 @@ default StatsContainerRequest := true
 default StopTracingRequest := false
 default TtyWinResizeRequest := true
 default UpdateContainerRequest := false
-default UpdateEphemeralMountsRequest := true
+default UpdateEphemeralMountsRequest := false
 default UpdateInterfaceRequest := true
 default UpdateRoutesRequest := true
 default WaitProcessRequest := true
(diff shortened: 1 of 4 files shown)


src/tools/genpolicy/rules.rego
```diff
@@ -38,8 +38,7 @@ default StartTracingRequest := false
 default StatsContainerRequest := true
 default StopTracingRequest := false
 default TtyWinResizeRequest := true
-default UpdateContainerRequest := false
```
danmihai1, line 41: UpdateContainerRequest seems unrelated to this PR - let's keep it here.

danmihai1 (changes requested):

src/tools/genpolicy/rules.rego
Redent0r: Good catch! updated

danmihai1:
Let's test these changes on TDX, SEV, SNP too. It's possible that they still need to call UpdateEphemeralMounts.

danmihai1 (approved):

GabyCT (approved):

PR #9920, opened 2024-06-26
gpu: rootfs/initrd build init

zvonkok:
Initramfs expects /init, create symlink only if ${ROOTFS}/init does not exist Init may be provided by other packages, e.g. systemd or GPU initrd/rootfs


fidencio (approved):
lgtm, thanks @zvonkok!

stevenhorsman (approved):
Looks okay to me.

PR #9926, opened 2024-06-27
ci: remove sudo from kata-static-arm64 build

zvonkok:
We do not need it in the amd64 case and we've made all artifacts owned by USER this is not needed anymore.


zvonkok:
We need to distinguish between self-hosted runners and the github runners. self-hosted runners are bare-metal machines that are not destroyed after each run.

fidencio (approved):
lgtm, thanks @zvonkok!

beraldoleal (approved):
LGTM, thanks.

PR #9937, opened 2024-06-28
CI: Correct rootfs ubuntu version and fix local (re)build

koct9i:
Do osbuilder rootfs cleanup inside docker when it have root permissions.

Add osbuilder "invalidate-rootfs" makefile target to rebuild image from scratch.

Fix passing "OS_VERSION" and fix default in "versions.yaml".
It actually is "focal". Before ce82b5e3f5b9a0b26bb127ff4490233cafc67ba0
it was hardcoded in "tools/osbuilder/rootfs-builder/ubuntu/config.sh",
now in "tools/packaging/kata-deploy/local-build/dockerbuild/Dockerfile".

Actual "latest" ubuntu (noble) does not work - "init" is not installed in rootfs.

NB: update from "focal" to "jammy" changes default cgroup v1 to v2 in in guest.

I've tested this via "make rootfs-image-tarball".

### DIFF
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -398,12 +398,6 @@ build_rootfs_distro()
 		trap error_handler ERR
 	fi
 
-	if [ -d "${ROOTFS_DIR}" ] && [ "${ROOTFS_DIR}" != "/" ]; then
-		rm -rf "${ROOTFS_DIR}"/*
-	else
-		mkdir -p ${ROOTFS_DIR}
-	fi
-
 	if [ "${SELINUX}" == "yes" ]; then
 		if [ "${AGENT_INIT}" == "yes" ]; then
 			die "Guest SELinux with the agent init is not supported yet"
@@ -414,6 +408,13 @@ build_rootfs_distro()
 	fi
 
 	if [ -z "${USE_DOCKER}" ] && [ -z "${USE_PODMAN}" ]; then
+
+		if [ -d "${ROOTFS_DIR}" ] && [ "${ROOTFS_DIR}" != "/" ]; then
+			rm -rf "${ROOTFS_DIR}"/*
+		else
+			mkdir -p ${ROOTFS_DIR}
+		fi
+
 		info "build directly"
 		build_rootfs ${ROOTFS_DIR}
 	else
--- a/tools/packaging/guest-image/build_image.sh
+++ b/tools/packaging/guest-image/build_image.sh
@@ -36,7 +36,7 @@ build_initrd() {
 	info "Build initrd"
 	info "initrd os: $os_name"
 	info "initrd os version: $os_version"
-	make initrd \
+	make invalidate-rootfs initrd \
 		DISTRO="$os_name" \
 		DEBUG="${DEBUG:-}" \
 		OS_VERSION="${os_version}" \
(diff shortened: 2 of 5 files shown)


tools/packaging/guest-image/build_image.sh
```diff
@@ -59,11 +59,11 @@ build_image() {
 	info "Build image"
 	info "image os: $os_name"
 	info "image os version: $os_version"
-	make image \
+	make invalidate-rootfs image \
```
sprt, line 62: I like the new `invalidate-rootfs` target but I'm not sure we should do this: this means we're changing the behavior of `make rootfs-image-tarball` and it's not idempotent anymore. Curious to hear what others think.

sprt (changes requested):
Thanks @koct9i! You'll want to fix the [commit message check](https://github.com/kata-containers/kata-containers/actions/runs/9712763301/job/26820430878?pr=9937) and specify a subsystem in your commit message as described here: https://github.com/kata-containers/community/blob/main/CONTRIBUTING.md#patch-format

I would also recommend splitting this into 2 commits: one for the rebuild part and another one for the version fix.

tools/packaging/guest-image/build_image.sh
koct9i: Ok. I've addressed this with "artifact-version" saved in rootfs mark file.

tools/packaging/guest-image/build_image.sh
```diff
@@ -59,15 +60,16 @@ build_image() {
...
 		DISTRO="${os_name}" \
 		DEBUG="${DEBUG:-}" \
 		USE_DOCKER="1" \
-		IMG_OS_VERSION="${os_version}" \
+		OS_VERSION="${os_version}" \
 		ROOTFS_BUILD_DEST="${builddir}/rootfs-image" \
 		AGENT_TARBALL="${AGENT_TARBALL}" \
 		AGENT_POLICY="${AGENT_POLICY:-}" \
 		PULL_TYPE="${PULL_TYPE:-default}" \
+		ARTEFACT_VERSION="${artefact_version}" \
```
sprt, line 72: Does this also pass the artifact version to the `invalidate-rootfs` target?

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -405,6 +399,13 @@ build_rootfs_distro()
 	fi
 
 	if [ -z "${USE_DOCKER}" ] && [ -z "${USE_PODMAN}" ]; then
+
+		if [ -d "${ROOTFS_DIR}" ] && [ "${ROOTFS_DIR}" != "/" ]; then
+			rm -rf "${ROOTFS_DIR}"/*
+		else
+			mkdir -p ${ROOTFS_DIR}
+		fi
```
sprt, line 407: Can you explain why we're only doing this in the Docker case now?

tools/packaging/guest-image/build_image.sh
koct9i: Yep. Same variables set for all targets called together.
make runs them sequentially.

koct9i:
replacement "touch maker-file" with "echo version > marker-file"
highlighted yet another problem with directory permissions:
"tools/packaging/kata-deploy/local-build/build/rootfs-image/builddir/rootfs-image"
was created inside rootfs builded and owned by root.
should be created outside by normal user to let changes from makefile.

I've added separate patch for that.

koct9i:
fixed long line in patch comment (>150)

PR #9941, opened 2024-06-28
ci: Temporarily disable kata-deploy and GARM tests

sprt:
```
Per the decision taken in the 6/27 AC meeting, this PR temporarily
disables kata-deploy and GARM tests until we secure further Azure CI
funding.

In the meantime, I'll transition the GARM tests to free runners and
reenable them to regain that coverage without affecting spending (see
#9940). If it turns out the free runners are too slow, we'll switch back
to GARM.

After funding is secured, we'll reenable the kata-deploy tests (see
#9939).
```

### DIFF
--- a/.github/workflows/basic-ci-amd64.yaml
+++ b/.github/workflows/basic-ci-amd64.yaml
@@ -15,6 +15,8 @@ on:
 
 jobs:
   run-cri-containerd:
+    # TODO: Transition to free runner (see #9940).
+    if: false
     strategy:
       # We can set this to true whenever we're 100% sure that
       # the all the tests are not flaky, otherwise we'll fail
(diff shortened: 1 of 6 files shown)


stevenhorsman (approved):
It's a bit sad to see and I really hope we can get some of these back on free runners soon, but I understand the need. We just need to sync this up with removing required tests as I suspect quite a lot of these might be.

wainersm:
> It's a bit sad to see and I really hope we can get some of these back on free runners soon, but I understand the need. We just need to sync this up with removing required tests as I suspect quite a lot of these might be.

One step back, two forward! \o/

wainersm (approved):
Thanks @sprt !

GabyCT (approved):

Issue #9942, opened 2024-06-28
The TEE CI tests are failing due to the deploy snapshotter timeout


AdithyaKrishnan:
Here's the log from the `kata-containers-ci-on-push / run-kata-coco-tests / run-k8s-tests-on-sev`:

```
/tmp/nydus-snapshotter ~/actions-runner/_work/kata-containers/kata-containers
  cleanup_nydus_snapshotter
  /tmp/nydus-snapshotter /tmp/nydus-snapshotter ~/actions-runner/_work/kata-containers/kata-containers
  Error from server (NotFound): namespaces "nydus-system" not found
  'kubectl get namespace nydus-system -o json' failed, retrying in 15 seconds
  Error from server (NotFound): namespaces "nydus-system" not found
  'kubectl get namespace nydus-system -o json' failed, retrying in 15 seconds
  Error from server (NotFound): namespaces "nydus-system" not found
  'kubectl get namespace nydus-system -o json' failed, retrying in 15 seconds
  Error from server (NotFound): namespaces "nydus-system" not found
  'kubectl get namespace nydus-system -o json' failed, retrying in 15 seconds
  Error from server (NotFound): namespaces "nydus-system" not found
  'kubectl get namespace nydus-system -o json' failed after 5 tries
  Error from server (Timeout): the server was unable to return a response in the time allotted, but may still be processing the request
  'kubectl replace --raw /api/v1/namespaces/nydus-system/finalize -f -' failed, retrying in 15 seconds
  Error from server (BadRequest): the object provided is unrecognized (must be of type Namespace): couldn't get version/kind; json parse error: unexpected end of JSON input (<empty>)
  'kubectl replace --raw /api/v1/namespaces/nydus-system/finalize -f -' failed, retrying in 15 seconds
  Error from server (BadRequest): the object provided is unrecognized (must be of type Namespace): couldn't get version/kind; json parse error: unexpected end of JSON input (<empty>)
  'kubectl replace --raw /api/v1/namespaces/nydus-system/finalize -f -' failed, retrying in 15 seconds
  Error from server (BadRequest): the object provided is unrecognized (must be of type Namespace): couldn't get version/kind; json parse error: unexpected end of JSON input (<empty>)
  'kubectl replace --raw /api/v1/namespaces/nydus-system/finalize -f -' failed, retrying in 15 seconds
[...]

PR #9923, opened 2024-06-27
tests: Increase interval and max_tries for kubectl_retry

BbolroC:
Observed instability in the API server after deploying kata-deploy caused test failures. (see: https://github.com/kata-containers/kata-containers/actions/runs/9681494440/job/26743286861) Specifically, `kubectl_retry logs` failed before the API server could respond properly.

This commit increases the interval and max_tries for kubectl_retry(), allowing sufficient time to handle this situation.

Signed-off-by: Hyounggyu Choi <Hyounggyu.Choi@ibm.com>

PR #9943, opened 2024-06-28
tests: Fixes TEE timeout issue

AdithyaKrishnan:
tests: remove unnecessary sleeps that slow down tests
The modifications to kubectl_retry from #9923 in combination with extensive sleep times in the nydus cleanup is causing various TEE environment timeouts and test failures.
Fixes: #9942

tests: delete the nydus directory at the end of cleanup
The nydus cleanup is skipped if the directory does not exist.
The directory was not being deleted.

Signed-Off-By: Adithya Krishnan Kannan <AdithyaKrishnan.Kannan@amd.com>

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -556,7 +556,6 @@ function cleanup_nydus_snapshotter() {
 	fi
 	sleep 180s
 	kubectl_retry delete --ignore-not-found -f "misc/snapshotter/nydus-snapshotter-rbac.yaml"
-	kubectl_retry get namespace nydus-system -o json | jq 'del(.spec.finalizers)' | kubectl_retry replace --raw "/api/v1/namespaces/nydus-system/finalize" -f - || true
 	popd
 	sleep 30s
 	echo "::endgroup::"


tests/integration/kubernetes/gha-run.sh
```diff
@@ -554,11 +554,24 @@ function cleanup_nydus_snapshotter() {
...
 	fi
-	sleep 180s
+
+	# The extensive sleep durations lead to certain tests timing out. 
+	# Commenting the sleep for archival purposes for now.
+	#sleep 180s
+
 	kubectl_retry delete --ignore-not-found -f "misc/snapshotter/nydus-snapshotter-rbac.yaml"
-	kubectl_retry get namespace nydus-system -o json | jq 'del(.spec.finalizers)' | kubectl_retry replace --raw "/api/v1/namespaces/nydus-system/finalize" -f - || true
+	kubectl get namespace nydus-system -o json | jq 'del(.spec.finalizers)' | kubectl replace --raw "/api/v1/namespaces/nydus-system/finalize" -f - || true
```
fidencio, line 563: Please, also drop this part, as we haven't seen it hanging anymore.

tests/integration/kubernetes/gha-run.sh
```diff
@@ -554,11 +554,24 @@ function cleanup_nydus_snapshotter() {
...
-	sleep 30s
+
+	# The extensive sleep durations lead to certain tests timing out. 
+	# Commenting the sleep for archival purposes for now.
+	#sleep 30s
+
+	# Deleting the nydus snapshotter directory
+	if [ -d "${nydus_snapshotter_install_dir}" ]; then
+		rm -rf "${nydus_snapshotter_install_dir}"
+	fi
```
fidencio, line 573: Don't do this, as the cleanup is also called during install, to remove any leftover installation.

BbolroC:
The TEE CI jobs are still failing due to:

```
  cleanup_nydus_snapshotter
  /tmp/nydus-snapshotter /tmp/nydus-snapshotter ~/actions-runner/_work/kata-containers/kata-containers
  Error from server (NotFound): namespaces "nydus-system" not found
  Error from server (BadRequest): the object provided is unrecognized (must be of type Namespace): couldn't get version/kind; json parse error: unexpected end of JSON input (<empty>)
  /tmp/nydus-snapshotter ~/actions-runner/_work/kata-containers/kata-containers
  
Error: stat misc/snapshotter/base/nydus-snapshotter.yaml: no such file or directory
```

fidencio:
I have force-pushed to @AdithyaKrishnan's  branch

fidencio:
TDX CDH test is failing with:
```
Events:
  Type     Reason   Age              From     Message
  ----     ------   ----             ----     -------
  Normal   Pulling  6s (x2 over 7s)  kubelet  Pulling image "quay.io/kata-containers/alpine-bash-curl:latest"
  Normal   Pulled   6s               kubelet  Successfully pulled image "quay.io/kata-containers/alpine-bash-curl:latest" in 403ms (404ms including waiting). Image size: 6924490 bytes.
  Warning  Failed   5s (x2 over 6s)  kubelet  Error: failed to create containerd container: create instance 8691: object with key "8691" already exists: unknown
  Normal   Pulled   5s               kubelet  Successfully pulled image "quay.io/kata-containers/alpine-bash-curl:latest" in 370ms (370ms including waiting). Image size: 6924490 bytes.
```

mkulke (approved):

tests/integration/kubernetes/gha-run.sh
```diff
@@ -554,11 +554,18 @@ function cleanup_nydus_snapshotter() {
 	else
 		kubectl_retry delete --ignore-not-found -f "misc/snapshotter/base/nydus-snapshotter.yaml"
 	fi
-	sleep 180s
+
+	# The extensive sleep durations lead to certain tests timing out. 
+	# Commenting the sleep for archival purposes for now.
+	#sleep 180s
+
```
stevenhorsman, line 561: I think we can delete the comments - if the sleeps aren't needed? Thanks

fidencio: done!

stevenhorsman (approved):
LGTM. Thanks!

stevenhorsman:
It looks like most of the tests are failing with a 90s delay:
```
Mon, 01 Jul 2024 11:30:26 GMT
1..1
Mon, 01 Jul 2024 11:31:56 GMT
not ok 1 Check capabilities of pod
Mon, 01 Jul 2024 11:31:56 GMT
# (in test file k8s-caps.bats, line 45)
Mon, 01 Jul 2024 11:31:56 GMT
Mon, 01 Jul 2024 11:31:56 GMT
# [bats-exec-test:67] INFO: k8s configured to use runtimeclass
```

stevenhorsman:
Oh - they are all failing with:
```
Error: failed to create containerd container: create instance 28619: object with key "28619" already exists: unknown
```
so I guess there is something in the cleanup not working?

tests/integration/kubernetes/gha-run.sh
ryansavino: We were discussing this in the slack thread, but the `kubectl_retry` call to patch the finalizer in the cleanup was causing an extended delay because it wasn't finding anything in the `nydus-system` namespace (already deleted) and was retrying the check 5 times 15 secs each time. This command is removed now in your revised commit, and I'm guessing the other `kubectl_retry` delete calls that have the `--ignore-not-found` flag won't have this issue, so maybe no hanging in the cleanup anymore. However, everything in the cleanup is skipped based on the initial check for the directory, and it was cleaned up. So my thought was that it would be nice to keep this here to streamline the deploy in case the cleanup call takes some time.

Wouldn't it be best to just completely clean it up and remove the directory in the cleanup method?

fidencio: So, let's think about what's happening here.
1. We call the method to deploy nydus snapshotter
2. The method itself will download nydus
3. And call the clean up function
4. Which will then delete the directory if this is added where it was added
5. Snapshot deployment will fail as the directory was removed

This is for sure **not** what we want.

ryansavino: My bad, I see your point. This is the wrong place. I still think we should remove it as part of a final cleanup step (part not to be run during deployment). Will look into the right place to put it.

stevenhorsman:
Are we okay to merge this now, or is there anything outstanding? It looks like the snapshotter deploy is passing in the PR and we've been blocked with TEE testing for about a week already?

BbolroC (approved):
I am fine with merging this because it looks like the discussion is complete. Thanks @AdithyaKrishnan !

wainersm (approved):
Thanks @AdithyaKrishnan @fidencio !

fidencio:
Sorry, I am using the superpowers here.
The only 2 failures are on the metrics CI, and this PR has absolutely nothing to do with those.

I'm merging it as we need to unblock the TEE tests.

PR #9947, opened 2024-07-01
gha: ci: Remove incorrect secrets line

stevenhorsman:
The CI is failing in https://github.com/kata-containers/kata-containers/actions/runs/9723935611 with:
```
Invalid workflow file: .github/workflows/cleanup-resources.yaml#L10
The workflow is not valid. .github/workflows/cleanup-resources.yaml (Line: 10, Col: 5): Unexpected value 'secrets'
```
I think this is because `secrets: inherit` is only applicable when re-using a workflow, not for a standalone job like we have here.

### DIFF
--- a/.github/workflows/cleanup-resources.yaml
+++ b/.github/workflows/cleanup-resources.yaml
@@ -7,7 +7,6 @@ on:
 jobs:
   cleanup-resources:
     runs-on: ubuntu-latest
-    secrets: inherit
     steps:
       - name: Cleanup resources
         run: python3 tests/cleanup_resources.py


mkulke (approved):
yes:

```bash
$ actionlint .github/workflows/cleanup-resources.yaml
.github/workflows/cleanup-resources.yaml:10:5: "secrets" is only available for a reusable workflow call with "uses" but "uses" is not found in job "cleanup-resources" [syntax-check]
   |
10 |     secrets: inherit
   |     ^~~~~~~~
```

GabyCT (approved):

stevenhorsman:
This is a workflow only change that isn't tested by the PR checks as it is not triggered with a `on/pull_request`. It also has been reviewed with the `force-skip-ci` Therefore we can skip the CI and use admin right to merge this.

Issue #9933, opened 2024-06-28
Enable the k8s attestation tests for IBM Secure Execution (SE)


BbolroC:
As https://github.com/confidential-containers/guest-components/pull/590 and https://github.com/confidential-containers/trustee/pull/345 are merged, we are ready to enable [the k8s attestation test](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-confidential-attestation.bats) for IBM Secure Execution (SE).

The following items are identified during the internal verification to achieve the goal:

- [x] Load a kernel module `uvdevice` to make a ultravisor present in a guest VM: https://github.com/kata-containers/kata-containers/pull/9886
- [x] Enable a test image `storytel/alpine-bash-curl:latest` for s390x: https://github.com/kata-containers/kata-containers/issues/9935
- [x] Enable kustomize config for [kbs deployment](https://github.com/confidential-containers/trustee/tree/main/kbs/config/kubernetes) (including making [files required by SE verifier](https://github.com/confidential-containers/trustee/tree/main/attestation-service/verifier/src/se#option-1-launch-kbs-as-a-program) available in k8s via a persistent volume/claim)https://github.com/confidential-containers/trustee/pull/436
- [ ] Prepare the required files above before deploying a KBS and rebuild a secure boot image when a KBS service is identified in order to pass additional kernel parameters (e.g. `agent.guest_components_rest_api=resource agent.aa_kbc_params=cc_kbc::$SVC_ADDR`) https://github.com/kata-containers/kata-containers/pull/9951

Due to limitation on availability of the test machine, the test will be run/reported via the [s390x nightly test](https://github.com/kata-containers/kata-containers/actions/workflows/ci-nightly-s390x.yaml).

PR #9951, opened 2024-07-01
tests: Enable attestation e2e tests for IBM SE

BbolroC:
This PR enables [k8s-confidential-attestation.bats](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-confidential-attestation.bats) for s390x. Detailed explanation can be found in each commit message.

Fixes: #9933

For reviewers: the tests have been passing since the 27th of June at https://github.com/kata-containers/kata-containers/actions/workflows/ci-nightly-s390x.yaml

### DIFF
--- a/tests/integration/kubernetes/confidential_kbs.sh
+++ b/tests/integration/kubernetes/confidential_kbs.sh
@@ -180,7 +180,7 @@ kbs_uninstall_cli() {
 #
 function kbs_k8s_delete() {
 	pushd "$COCO_KBS_DIR"
-	kubectl delete -k config/kubernetes/overlays
+	kubectl delete -k config/kubernetes/overlays/$(uname -m)
 	# Verify that KBS namespace resources were properly deleted
 	cmd="kubectl get all -n $KBS_NS 2>&1 | grep 'No resources found'"
 	waitForProcess "120" "30" "$cmd"
@@ -237,7 +237,7 @@ function kbs_k8s_deploy() {
 
 	# Tests should fill kbs resources later, however, the deployment
 	# expects at least one secret served at install time.
-	echo "somesecret" > overlays/key.bin
+	echo "somesecret" > overlays/$(uname -m)/key.bin
 
 	echo "::group::Update the kbs container image"
 	install_kustomize
@@ -437,7 +437,7 @@ _handle_ingress_aks() {
 		return 1
 	fi
 
-	pushd "${COCO_KBS_DIR}/config/kubernetes/overlays"
+	pushd "${COCO_KBS_DIR}/config/kubernetes/overlays/common"
 
 	echo "::group::$(pwd)/ingress.yaml"
 	KBS_INGRESS_CLASS="addon-http-application-routing" \
--- a/tests/integration/kubernetes/confidential_kbs.sh
+++ b/tests/integration/kubernetes/confidential_kbs.sh
@@ -12,6 +12,9 @@ kubernetes_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "${kubernetes_dir}/../../gha-run-k8s-common.sh"
 # shellcheck disable=1091
 source "${kubernetes_dir}/../../../tests/common.bash"
+source "${kubernetes_dir}/../../../tools/packaging/guest-image/lib_se.sh"
+# For kata-runtime
+export PATH="${PATH}:/opt/kata/bin"
 
 KATA_HYPERVISOR="${KATA_HYPERVISOR:-qemu}"
 # Where the trustee (includes kbs) sources will be cloned
@@ -239,6 +242,14 @@ function kbs_k8s_deploy() {
 	# expects at least one secret served at install time.
 	echo "somesecret" > overlays/$(uname -m)/key.bin
 
+	# For qemu-se runtime, prepare the necessary resources
+	if [ "${KATA_HYPERVISOR}" == "qemu-se" ]; then
+		prepare_credentials_for_qemu_se
+		# SE_SKIP_CERTS_VERIFICATION should be set to true
+		# to skip the verification of the certificates
+		sed -i "s/false/true/g" overlays/s390x/patch.yaml
+	fi
+
 	echo "::group::Update the kbs container image"
 	install_kustomize
 	pushd base
@@ -455,4 +466,33 @@ _handle_ingress_aks() {
 _handle_ingress_nodeport() {
 	# By exporting this variable the kbs deploy script will install the nodeport service
 	export DEPLOYMENT_DIR=nodeport
+}
+
+
+# Prepare necessary resources for qemu-se runtime
+# Documentation: https://github.com/confidential-containers/trustee/tree/main/attestation-service/verifier/src/se
+prepare_credentials_for_qemu_se() {
+	echo "::group::Prepare credentials for qemu-se runtime"
+	if [ -z "${IBM_SE_CREDS_DIR:-}" ]; then
+		>&2 echo "ERROR: IBM_SE_CREDS_DIR is empty"
+		return 1
+	fi
+	config_file_path="/opt/kata/share/defaults/kata-containers/configuration-qemu-se.toml"
+	kata_base_dir=$(dirname $(kata-runtime --config ${config_file_path} env --json | jq -r '.Kernel.Path'))
+	if [ ! -d ${HKD_PATH} ]; then
+		>&2 echo "ERROR: HKD_PATH is not set"
+		return 1
+	fi
+	pushd "${IBM_SE_CREDS_DIR}"
+	mkdir {certs,crls,hdr,hkds,rsa}
+	openssl genrsa -aes256 -passout pass:test1234 -out encrypt_key-psw.pem 4096
+	openssl rsa -in encrypt_key-psw.pem -passin pass:test1234 -pubout -out rsa/encrypt_key.pub
+	openssl rsa -in encrypt_key-psw.pem -passin pass:test1234 -out rsa/encrypt_key.pem
+	cp ${kata_base_dir}/kata-containers-se.img hdr/hdr.bin
+	cp ${HKD_PATH}/HKD-*.crt hkds/
+	cp ${HKD_PATH}/ibm-z-host-key-gen2.crl crls/
+	cp ${HKD_PATH}/DigiCertCA.crt ${HKD_PATH}/ibm-z-host-key-signing-gen2.crt certs/
+	popd
+	ls -R ${IBM_SE_CREDS_DIR}
+	echo "::endgroup::"
 }
--- a/tools/packaging/guest-image/build_se_image.sh
+++ b/tools/packaging/guest-image/build_se_image.sh
@@ -9,27 +9,18 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly script_name="$(basename "${BASH_SOURCE[0]}")"
-readonly script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-readonly packaging_root_dir="$(cd "${script_dir}/../" && pwd)"
-readonly kata_root_dir="$(cd "${packaging_root_dir}/../../" && pwd)"
+script_name="$(basename "${BASH_SOURCE[0]}")"
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+packaging_root_dir="$(cd "${script_dir}/../" && pwd)"
+kata_root_dir="$(cd "${packaging_root_dir}/../../" && pwd)"
 
-source "$kata_root_dir/tests/common.bash"
 source "${packaging_root_dir}/scripts/lib.sh"
+source "${script_dir}/lib_se.sh"
 
 ARCH=${ARCH:-$(uname -m)}
 if [ $(uname -m) == "${ARCH}" ]; then
 	[ "${ARCH}" == "s390x" ] || die "Building a Secure Execution image is currently only supported on s390x."
 fi
-
-finish() {
-	if [ -e "${parmfile}" ]; then
-		rm -f "${parmfile}"
-	fi
-}
-
-trap finish EXIT
-
 usage() {
 	cat >&2 << EOF
 Usage:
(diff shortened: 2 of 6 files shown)


tools/packaging/static-build/coco-guest-components/build.sh
stevenhorsman, line 1: Based on the logic in the guest-components `Makefile`, doesn't setting `TEE_PLATFORM` effectively overide what is set in `ATTESTER`?

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
```diff
@@ -114,6 +114,7 @@ docker run \
 	--env ARTEFACT_REGISTRY_PASSWORD="${ARTEFACT_REGISTRY_PASSWORD}" \
 	--env TARGET_BRANCH="${TARGET_BRANCH}" \
 	--env BUILDER_REGISTRY="${BUILDER_REGISTRY}" \
+	--env TEE_PLATFORM="${TEE_PLATFORM:-}" \
```
stevenhorsman, line 117: Shouldn't we extract out `TEE_PLATFORM="${TEE_PLATFORM:-}"` like all the other envs?

tools/packaging/static-build/coco-guest-components/build.sh
```diff
@@ -39,17 +39,17 @@ docker pull ${container_image} || \
 	 push_to_registry "${container_image}")
 
 # Temp settings until we have a matching TEE_PLATFORM
-TEE_PLATFORM=""
-RESOURCE_PROVIDER="kbs,sev"
+TEE_PLATFORM="${TEE_PLATFORM:-all}"
```
stevenhorsman, line 42: If we default to `all` here, are we not going to end up trying to build the tpm attesters in `s390x` and fail here, or is there an override that sets `TEE_PLATFORM=se` if the arch it s390x, that I'm missing?

BbolroC: You are right. That should make the s390x build failing. There is no override setting `TEE_PLATFORM=se` in the s390x pipeline. It must be `TEE_PLATFORM="${TEE_PLATFORM:-}"`.  Thanks for pointing it out.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
BbolroC: If we don't have this line, there is no way of building the guest components where `se-attester` is solely built with `make coco-guest-components-tarball`. With this line, we can achieve to have those guest components by running `make coco-guest-components-tarball TEE_PLATFORM=se`. 

If there is another way of achieving this goal, please let me know. I will following that way.

tools/packaging/static-build/coco-guest-components/build.sh
BbolroC: yeah, setting `TEE_PLATFORM` effectively overrides `ATTESTER`.

tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh
stevenhorsman: Sorry for not being clear. The pattern in that file seems to be to set `TEE_PLATFORM="${TEE_PLATFORM:-}"` separately (e.g. https://github.com/kata-containers/kata-containers/blob/2396968d92f00636caa216675a5b2e0f85844742/tools/packaging/kata-deploy/local-build/kata-deploy-binaries-in-docker.sh#L82-L103), then in the `docker` call, just use `--env TEE_PLATFORM="${TEE_PLATFORM}"`. I don't think the effect will be different and don't necessarily even like that convention, but I think it's best to be consistent.

BbolroC: yes, you are right. It does not look consistent with others. Thanks!

stevenhorsman (approved):
The code looks good to me now, lets hope the CI is okay 🤞. Thanks for the updates!

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -237,7 +237,13 @@ function kbs_k8s_deploy() {
 
 	# Tests should fill kbs resources later, however, the deployment
 	# expects at least one secret served at install time.
-	echo "somesecret" > overlays/key.bin
+	if [ -d "overlays/$(uname -m)" ]; then
```
wainersm, line 243: Hi @BbolroC !

Kata uses trustee v0.9.0 which doesn't contain the multi-arch support. So once we bump to newest trustee we will re-do this change? i.e. to always `echo "somesecret" > overlays/$(uname -m)/key.bin` ? 

If yes, could we mark with a `TODO` comment?

Thinking aloud... perhaps trustee's `deploy-kbs.sh` could deal with that secret. Consumers should simply export the secret in a variable, this way any break on trustee side doesn't affect us.

wainersm: ah, also any special reason to not bump trustee on this PR?

tools/packaging/guest-image/build_se_image.sh
```diff
@@ -9,27 +9,18 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly script_name="$(basename "${BASH_SOURCE[0]}")"
-readonly script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-readonly packaging_root_dir="$(cd "${script_dir}/../" && pwd)"
-readonly kata_root_dir="$(cd "${packaging_root_dir}/../../" && pwd)"
+script_name="$(basename "${BASH_SOURCE[0]}")"
```
wainersm, line 12: I'm curious why it needed the removal of `readonly`

wainersm: hmm maybe because same variables are set in [tools/packaging/guest-image/lib_se.sh](https://github.com/kata-containers/kata-containers/pull/9951/commits/15b16041850acd7cd744a5a758c5212048820e30#diff-3513f55f212771374656907cdec9a7f2335c41d9903426eb6687dacc307f1510) so it complains they are readonly?

tools/packaging/guest-image/lib_se.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
```
wainersm, line 1: didn't review this file. I assume it's just a copy without modification ;)

tests/integration/kubernetes/confidential_kbs.sh
stevenhorsman: I guess that the trustee team cut the `v0.9.0` as they felt it was stable. That doesn't mean we can't bump it here, but we probably want to get some input/agreement with them first.

BbolroC: Thanks for the comment. I think the change covers both cases (a version with/without the multi-arch support), which means x86_64 is not affected as the version changes. However, we could refactor the code if `echo "somesecret" > overlays/key.bin` is no longer used. 

> Thinking aloud... perhaps trustee's deploy-kbs.sh could deal with that secret. Consumers should simply export the secret in a variable, this way any break on trustee side doesn't affect us.

I was thinking the same. But even in this case, we will face the similar issue once when bumping the version between two repos.

ATM, the s390x nightly test is using the main branch for trustee like:

```diff
diff --git a/tests/integration/kubernetes/confidential_kbs.sh b/tests/integration/kubernetes/confidential_kbs.sh
index 36f4244b5..1e297eaf1 100644
--- a/tests/integration/kubernetes/confidential_kbs.sh
+++ b/tests/integration/kubernetes/confidential_kbs.sh
@@ -230,10 +230,10 @@ function kbs_k8s_deploy() {

        echo "::group::Clone the kbs sources"
        git clone --depth 1 "${repo}" "$COCO_TRUSTEE_DIR"
-       pushd "$COCO_TRUSTEE_DIR"
-       git fetch --depth=1 origin "${version}"
-       git checkout FETCH_HEAD -b kbs_$$
-       popd
+       #pushd "$COCO_TRUSTEE_DIR"
+       #git fetch --depth=1 origin "${version}"
+       #git checkout FETCH_HEAD -b kbs_$$
+       #popd
        echo "::endgroup::"

        pushd "${COCO_KBS_DIR}/config/kubernetes/"
```

I've opened a thread for the update of the trustee for kata containers on Slack: https://cloud-native.slack.com/archives/C039JSH0807/p1720080885154259. If the trustee team is okay to bump it, I will include it here. Thanks.

tools/packaging/guest-image/build_se_image.sh
BbolroC: Exactly! 😉 They are declared as readonly before sourcing the library. Fortunately, the two files are located in the same directory. We could differentiate those names to prevent this kind of conflict if the library is used in various places, which doesn't seem to be the case here. I need your advise on this. Thanks!

tools/packaging/guest-image/lib_se.sh
BbolroC: The only change I made was to relocate `rm -f "${parmfile}"` from a trap callback to the end of the `build_secure_image()` function. Thanks!

BbolroC:
@wainersm  As confirmed by the trustee team, I've added a commit to bump trustee and updated the commit for changing the location where a `key.bin` is created. Please take a look. Thanks!

BbolroC:
The following issue have been observed:

```
Setting up custom PCCS for TDX
  error: accumulating resources: accumulation err='accumulating resources from '../nodeport': '/tmp/trustee/kbs/config/kubernetes/nodeport' must resolve to a file': couldn't make target for path '/tmp/trustee/kbs/config/kubernetes/nodeport': unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/tmp/trustee/kbs/config/kubernetes/nodeport'
```

This should be resolved from trustee. This PR should be held until then. Thanks.

fitzthum:
Ok, looks like there is a little bug in the most recent version of Trustee. We will break TDX if we bump it. I should be able to post a fix soon tho.

fitzthum:
See https://github.com/confidential-containers/trustee/pull/439

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -310,6 +321,29 @@ EOF
...
 		fi
 
+		# A secure boot image for IBM SE should be rebuilt according to the KBS configuration.
+		if [ "${KATA_HYPERVISOR}" == "qemu-se" ]; then
+			if [ -z "${IBM_SE_CREDS_DIR:-}" ]; then
+				>&2 echo "ERROR: IBM_SE_CREDS_DIR is empty"
+				return 1
+			fi
+			creds_dir="${IBM_SE_CREDS_DIR}"
+			kernel_params_value="agent.guest_components_rest_api=resource agent.aa_kbc_params=cc_kbc::${svc_host}"
```
wainersm, line 331: Hi @BbolroC !

I started reading this commit with a feeling of "it seems to belong to somewhere else". Then I hit that line, which made me more inclined towards suggesting to wrap in a function to be called from the test. Why? Because if this `kernel_params_value` should match the one it's passed via Pod annotation, then it might be different from test to test.

For example, the recent added [test for sealed secrets](https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-sealed-secret.bats#L36) will pass the following kernel parameters instead: `agent.guest_components_procs=confidential-data-hub  agent.aa_kbc_params=cc_kbc::${svc_host}`

or am I missing something?

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -310,6 +321,29 @@ EOF
...
+			config_file_path="/opt/kata/share/defaults/kata-containers/configuration-qemu-se.toml"
+			if [ ! -f "${config_file_path}" ]; then
+				>&2 echo "ERROR: config file not found: ${config_file_path}"
+				return 1
+			fi
+			kata_base_dir=$(dirname $(kata-runtime --config ${config_file_path} env --json | jq -r '.Kernel.Path'))
```
wainersm, line 337: `kernel_base_dir`?

I was going to suggest you to use `extract_kata_env()` from `tests/common.bash`, but it would need to add support to export the kernel path. I don't know if that function will get the right config file path too. It's up to you fight that battle or not :)

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -456,3 +490,30 @@ _handle_ingress_nodeport() {
...
+# Documentation: https://github.com/confidential-containers/trustee/tree/main/attestation-service/verifier/src/se
+prepare_credentials_for_qemu_se() {
+	echo "::group::Prepare credentials for qemu-se runtime"
+	config_file_path="/opt/kata/share/defaults/kata-containers/configuration-qemu-se.toml"
+	kata_base_dir=$(dirname $(kata-runtime --config ${config_file_path} env --json | jq -r '.Kernel.Path'))
+	if [ ! -d ${HKD_PATH} ]; then
+		>&2 echo "ERROR: HKD_PATH is not set"
+		return 1
+	fi
+	ibmse_creds_dir=$(mktemp -d -t ibmse.creds.XXXXXXXXXX)
```
wainersm, line 505: if you end up moving this code close to the tests then you will be able to use bats' temporary directory and not care about cleaning that directory later.

tests/integration/kubernetes/confidential_kbs.sh
```diff
@@ -456,3 +490,30 @@ _handle_ingress_nodeport() {
...
+	export IBM_SE_CREDS_DIR="${ibmse_creds_dir}"
+	pushd "${ibmse_creds_dir}"
+	mkdir {certs,crls,hdr,hkds,rsa}
+	openssl genrsa -aes256 -passout pass:test1234 -out encrypt_key-psw.pem 4096
```
wainersm, line 509: I hate that many websites don't allow me to set such as amazing test1234 password!

wainersm:
Hi @BbolroC ! I think I finished my review. Left just a few comments, overall looks great!

tests/integration/kubernetes/confidential_kbs.sh
BbolroC: Yes, from the beginning, this was wrapped as a function and called from k8s-confidential-attestation.bats. But, at the time of refactoring, I realised that file is the only one calling the function. That is why I had this snippet called once when the KBS is deployed. 

I totally agree with your opinion. I will change the code.

tests/integration/kubernetes/confidential_kbs.sh
BbolroC: Thanks for the suggestion. I think renaming it to `kernel_base_dir` would be enough in this case. Thanks.

tests/integration/kubernetes/confidential_kbs.sh
BbolroC: A lifecycle of this directory should match with that of KBS rather than the tests. It is mounted to the KBS as a persistent volume and cleaned up by the post-action script. If it is removed at a task `running tests`, this would be problematic for uninstalling a cluster. I would like to leave it as it is. Thanks.

BbolroC: In line with my argument, I took out the declaration of `IBM_SE_CREDS_DIR`. The env variable should be defined/cleaned up in the workflow for the internal SE k8s tests.

BbolroC: I made a new function `repack_secure_image()` general to cover a case where a secure image only needs to be built without KBS. Thanks!

BbolroC:
I've confirmed the e2e tests for the latest update are passed.

fitzthum:
TDX and SEV/SNP CI failing to rebase :'(

GabyCT:
I just re-run the TDX CI job

BbolroC:
> I just re-run the TDX CI job

The job has not been picked up by the runner quite a while. Is the runner being occupied too much?

GabyCT:
> > I just re-run the TDX CI job
> 
> The job has not been picked up by the runner quite a while. Is the runner being occupied too much?

yes, the job is in queue I re-triggered other PRs for TDX and seems like now is running another PR

BbolroC:
@GabyCT The error says:

```
The pkg-config command could not be found.

  Most likely, you need to install a pkg-config package for your OS.
  Try `apt install pkg-config`, or `yum install pkg-config`,
  or `pkg install pkg-config`, or `apk add pkgconfig` depending on your distribution.
```

The package seems not to be installed. Or am I missing something?

GabyCT:
> @GabyCT The error says:
> 
> ```
> The pkg-config command could not be found.
> 
>   Most likely, you need to install a pkg-config package for your OS.
>   Try `apt install pkg-config`, or `yum install pkg-config`,
>   or `pkg install pkg-config`, or `apk add pkgconfig` depending on your distribution.
> ```
> 
> The package seems not to be installed. Or am I missing something?

@BbolroC I just installed the package and re-triggered the job

GabyCT:
@BbolroC I installed the required packages to avoid the failures in installing KBS client step, however, I re-ran the job, and I got twice the same failure on TDX on the following test
```
[run_kubernetes_tests.sh:122] ERROR: Tests FAILED from suites: k8s-confidential.bats
```
```
#   Warning  Failed     82s               kubelet            Error: failed to create containerd container: error unpacking image: failed to prepare extraction snapshot "extract-304855051-hXUq sha256:414cb028a051f35dcb03e7e474dabfa2d04492a94e30b4862d819cab4917cc01": create instance 6186: object with key "6186" already exists: unknown
```

fitzthum:
> seems like this failure is not related with this PR. @wainersm @fitzthum @stevenhorsman if you are familiar with this kind of failure?

pretty common snapshotter gremlin. @ChengyuZhu6 probably knows how to reset things to fix this

BbolroC:
@wainersm all TEE test are passing. It is time to merge this, right? Please. 😢

wainersm:
Hi @BbolroC !

> @wainersm all TEE tests are passing. It is time to merge this, right? Please. 😢

I'd like to propose a change and it will trigger all jobs again....

... just kidding! Looks good! :)

wainersm (approved):

PR #9954, opened 2024-07-02
metrics: Remove variable in sysbench that is not being used

GabyCT:
This PR removes the CI_JOB variable which previously was used but not longer being supported of the metrics sysbench test.


stevenhorsman (approved):
LGTM. Thanks @GabyCT!

dborquez (approved):
LGTM, thank you @GabyCT

PR #9955, opened 2024-07-02
genpolicy: support named user mapping

### LINKED ISSUE #9928: genpolicy: support user names in image config

`genpolicy`

**How can it be improved?**

Generated policies for images that configure the runtime user by name should work.

```Dockerfile
FROM alpine:latest

USER postmaster
```

<details>
<summary>Manifest</summary>

```json
{
  "architecture": "amd64",
  "config": {
    "User": "postmaster",
    "Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ],
    "Cmd": [
      "/bin/sh"
    ]
  },
  "created": "2023-12-08T01:20:49.650406179Z",
  "history": [
    {
      "created": "2023-12-08T01:20:49.493752696Z",
      "created_by": "/bin/sh -c #(nop) ADD file:1f4eb46669b5b6275af19eb7471a6899a61c276aa7d925b8ae99310b14b75b92 in / "
    },
    {
      "created": "2023-12-08T01:20:49.650406179Z",
      "created_by": "/bin/sh -c #(nop)  CMD [\"/bin/sh\"]",
      "empty_layer": true
    },
    {
      "created": "2023-12-08T01:20:49.650406179Z",
      "created_by": "USER postmaster",
      "comment": "buildkit.dockerfile.v0",
      "empty_layer": true
    }
  ],
  "os": "linux",
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:5af4f8f59b764c64c6def53f52ada809fe38d528441d08d01c206dfb3fc3b691"
    ]
  }
}
```

</details>

A pod using this image makes genpolicy log a warning (if the log level is set correctly):

```txt
[2024-06-27T12:52:52Z WARN  genpolicy::registry] Failed to parse postmaster as u32, using uid = 0 - error invalid digit found in string
```

https://github.com/kata-containers/kata-containers/blob/29bb9de8646c3899a60f4f92c080b85f2d3d18ae/src/tools/genpolicy/src/registry.rs#L136-L158

The generated policy for this image never passes, because the runtime uid 14 does not match the defaulted 0.

**Additional Information**

It's definitely not straightforward to implement, but also not impossible - we're already unpacking the layers to generate the dm-verity hash, so we _could_ also look for `/etc/passwd` while we're doing that.

As a side note, the OCI image spec regrettably allows all kinds of combinations of UID and names, probably causing the group `.unwrap()` to panic.

> For Linux based systems, all of the following are valid: `user`, `uid`, `user:group`, `uid:gid`, `uid:group`, `user:gid`.
>
> Source: [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/main/config.md#properties)

:heavy_check_mark:  Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)? 

**Kata Containers survey**

:heavy_check_mark:  Please consider taking the survey to help us help you: https://openinfrafoundation.formstack.com/forms/kata_containers_user_survey

HernanGatta:
Some containers specify their user via a name (e.g., Spark). As a result, instead of there being a UID:GID pair in the user field, there is instead a string (e.g., 'spark') that only resolves to a UID:GID pair at runtime.

Add a field to the `genpolicy` configuration file that allows specifying what UID:GID pair a given named user corresponds to. That way, the generated policy enforces the correct pair instead of rejecting the creation of the container, since the tool defaults to 0:0.

Fixes #9928

### DIFF
--- a/src/tools/genpolicy/src/registry.rs
+++ b/src/tools/genpolicy/src/registry.rs
@@ -23,6 +23,7 @@ use oci_distribution::{
 };
 use serde::{Deserialize, Serialize};
 use sha2::{digest::typenum::Unsigned, digest::OutputSizeUser, Sha256};
+use std::io::Read;
 use std::{fs::OpenOptions, io, io::BufWriter, io::Seek, io::Write, path::Path};
 use tokio::io::AsyncWriteExt;
 
@@ -64,7 +65,36 @@ pub struct DockerRootfs {
 pub struct ImageLayer {
     pub diff_id: String,
     pub verity_hash: String,
+    pub passwd: String,
 }
+
+/// A single record in a Unix passwd file.
+#[derive(Debug)]
+struct PasswdRecord {
+    pub user: String,
+    #[allow(dead_code)]
+    pub validate: bool,
+    pub uid: u32,
+    pub gid: u32,
+    #[allow(dead_code)]
+    pub gecos: String,
+    #[allow(dead_code)]
+    pub home: String,
+    #[allow(dead_code)]
+    pub shell: String,
+}
+
+/// Path to /etc/passwd in a container layer's tar file.
+const PASSWD_FILE_TAR_PATH: &str = "etc/passwd";
+
+/// Path to a file indicating a whiteout of the /etc/passwd file in a container
+/// layer's tar file (i.e., /etc/passwd was deleted in the layer).
+const PASSWD_FILE_WHITEOUT_TAR_PATH: &str = "etc/.wh.passwd";
+
+/// A marker used to track whether a particular container layer has had its
+/// /etc/passwd file deleted, and thus any such files read from previous, lower
+/// layers should be discarded.
+const WHITEOUT_MARKER: &str = "WHITEOUT";
 
 impl Container {
     pub async fn new(config: &Config, image: &str) -> Result<Self> {
@@ -133,28 +163,87 @@ impl Container {
         debug!("Getting process field from docker config layer...");
         let docker_config = &self.config_layer.config;
 
+        // Defaults to start with if we can't work out the correct values.
+        process.User.UID = 0;
+        process.User.GID = 0;
+
+        /*
+         * The user field may:
+         *
+         * 1. Be empty
+         * 2. Contain only a UID
+         * 3. Contain a UID:GID pair, in that format
+         * 4. Contain a user name, which we need to translate into a UID/GID pair
+         * 5. Be erroneus, somehow
+         */
         if let Some(image_user) = &docker_config.User {
             if !image_user.is_empty() {
-                debug!("Splitting Docker config user = {:?}", image_user);
-                let user: Vec<&str> = image_user.split(':').collect();
-                if !user.is_empty() {
-                    debug!("Parsing uid from user[0] = {}", &user[0]);
-                    match user[0].parse() {
-                        Ok(id) => process.User.UID = id,
-                        Err(e) => {
-                            // "image: prom/prometheus" has user = "nobody", but
-                            // process.User.UID is an u32 value.
-                            warn!(
-                                "Failed to parse {} as u32, using uid = 0 - error {e}",
-                                &user[0]
-                            );
-                            process.User.UID = 0;
+                if image_user.contains(':') {
+                    debug!("Splitting Docker config user = {:?}", image_user);
+                    let user: Vec<&str> = image_user.split(':').collect();
+                    let parts_count = user.len();
+                    if parts_count != 2 {
+                        warn!(
+                            "Failed to split user, expected two parts, got {}, using uid = gid = 0",
+                            parts_count
+                        );
+                    } else {
+                        debug!("Parsing uid from user[0] = {}", &user[0]);
+                        match user[0].parse() {
+                            Ok(id) => process.User.UID = id,
+                            Err(e) => {
+                                warn!(
+                                    "Failed to parse {} as u32, using uid = 0 - error {e}",
+                                    &user[0]
+                                );
+                            }
+                        }
+
+                        debug!("Parsing gid from user[1] = {:?}", user[1]);
+                        match user[1].parse() {
+                            Ok(id) => process.User.GID = id,
+                            Err(e) => {
+                                warn!(
+                                    "Failed to parse {} as u32, using gid = 0 - error {e}",
+                                    &user[0]
+                                );
+                            }
                         }
                     }
-                }
-                if user.len() > 1 {
-                    debug!("Parsing gid from user[1] = {:?}", user[1]);
-                    process.User.GID = user[1].parse().unwrap();
+                } else {
+                    match image_user.parse::<u32>() {
+                        Ok(uid) => process.User.UID = uid,
+                        Err(outer_e) => {
+                            // Find the last layer with an /etc/passwd file,
+                            // respecting whiteouts.
+                            let mut passwd = "".to_string();
+                            for layer in self.get_image_layers() {
+                                if !layer.passwd.is_empty() {
+                                    passwd = layer.passwd
+                                } else if layer.passwd == WHITEOUT_MARKER {
+                                    passwd = "".to_string();
+                                }
+                            }
+
+                            if passwd.is_empty() {
+                                warn!("Failed to parse {} as u32 - error {outer_e} - and no /etc/passwd file is available, using uid = gid = 0", image_user);
+                            } else {
+                                match parse_passwd_file(passwd) {
+                                    Ok(records) => {
+                                        if let Some(record) =
+                                            records.iter().find(|&r| r.user == *image_user)
+                                        {
+                                            process.User.UID = record.uid;
+                                            process.User.GID = record.gid;
+                                        }
+                                    }
+                                    Err(inner_e) => {
+                                        warn!("Failed to parse {} as u32 - error {outer_e} - and failed to parse /etc/passwd - error {inner_e}, using uid = gid = 0", image_user);
+                                    }
+                                }
+                            }
+                        }
+                    }
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


danmihai1:
@HernanGatta , does https://github.com/kata-containers/kata-containers/issues/9928 sound like the same issue you are trying to address here?

If yes, I still hope that we'll be able to overlay the image layers and figure out the UID:GID from the overlay someday.

HernanGatta:
> @HernanGatta , does #9928 sound like the same issue you are trying to address here?
> 
> If yes, I still hope that we'll be able to overlay the image layers and figure out the UID:GID from the overlay someday.

@danmihai1 Yes, I included that issue in the commit's "Fixes" line.

I would indeed prefer to figure the UID:GID from the container image itself, though I'm not familiar enough (yet) with the tool's internals to understand how much work it would be to mount all the overlays, then go parse `/etc/passwd`.

danmihai1:
> > @HernanGatta , does #9928 sound like the same issue you are trying to address here?
> > If yes, I still hope that we'll be able to overlay the image layers and figure out the UID:GID from the overlay someday.
> 
> @danmihai1 Yes, I included that issue in the commit's "Fixes" line.
> 
> I would indeed prefer to figure the UID:GID from the container image itself, though I'm not familiar enough (yet) with the tool's internals to understand how much work it would be to mount all the overlays, then go parse `/etc/passwd`.

The relevant code is around get_image_layers(). 

CC: @Redent0r , @burgerdev. Hopefully one of us can figure out a good way to overlay these files, and parse /etc/passwd.

burgerdev:
We would not even need to overlay them - it should be enough to find the last layer containing an `/etc/passwd` file (respecting whiteouts).

HernanGatta:
By way of update, I'm actively working on this. I've part of it working, and I'm currently debugging an issue that I presume is due to my unfamiliarity with Rust. I'll push an update once I figure it out.

danmihai1:
> By way of update, I'm actively working on this. I've part of it working, and I'm currently debugging an issue that I presume is due to my unfamiliarity with Rust. I'll push an update once I figure it out.

That sounds great, thank you @HernanGatta !

HernanGatta:
@danmihai1 I've updated the change to use `/etc/passwd` as requested.

With this change, the tool generates policies for our workloads that have the correct `uid` and `gid` values.

I have some thoughts about what I've done, and ways of doing things differently, but knowing that I tend to overcomplicate things, I'll let you be the judge of how horrible this is, and do please pardon my Rust.

src/tools/genpolicy/src/registry.rs
```diff
@@ -453,7 +557,22 @@ pub fn get_verity_hash_value(path: &Path) -> Result<String> {
...
+    for entry_wrap in tar::Archive::new(file).entries()? {
+        let mut entry = entry_wrap?;
+        let entry_path = entry.header().path()?;
+        let path_str = entry_path.to_str().unwrap();
+        if path_str == PASSWD_FILE_TAR_PATH {
+            entry.read_to_string(&mut passwd)?;
+        } else if path_str == PASSWD_FILE_WHITEOUT_TAR_PATH {
+            passwd = WHITEOUT_MARKER.to_owned();
+        }
+        break;
```
HernanGatta, line 572: Mistake in latest update, will fix.

src/tools/genpolicy/src/registry.rs
```diff
@@ -133,28 +163,87 @@ impl Container {
...
+        process.User.UID = 0;
+        process.User.GID = 0;
+
+        /*
+         * The user field may:
+         *
+         * 1. Be empty
+         * 2. Contain only a UID
+         * 3. Contain a UID:GID pair, in that format
+         * 4. Contain a user name, which we need to translate into a UID/GID pair
```
burgerdev, line 176: This list is not exhaustive[^1]: we might also get a `username:gid`, `uid:groupname` etc. I think it might be better to split at `:` and then treat the first item as in the `else` block below (213ff.). If there is a second item, we can either leave the int parsing (203ff.) for now , or you could add similar functionality for `/etc/group`.



[^1]: https://github.com/opencontainers/image-spec/blob/8797c3fedb77200c102c816ddcd295bb1e19d3e3/config.md?plain=1#L140

src/tools/genpolicy/src/registry.rs
```diff
@@ -133,28 +163,87 @@ impl Container {
...
+                            if passwd.is_empty() {
+                                warn!("Failed to parse {} as u32 - error {outer_e} - and no /etc/passwd file is available, using uid = gid = 0", image_user);
+                            } else {
+                                match parse_passwd_file(passwd) {
+                                    Ok(records) => {
+                                        if let Some(record) =
+                                            records.iter().find(|&r| r.user == *image_user)
+                                        {
+                                            process.User.UID = record.uid;
+                                            process.User.GID = record.gid;
```
burgerdev, line 237: I just remembered the big warning sign here: https://docs.docker.com/reference/dockerfile/#user. It's not only correct to use the gid from the passwd file here, but I think we should even do that lookup if the field is a uid (without a gid).

src/tools/genpolicy/src/registry.rs
```diff
@@ -453,7 +557,23 @@ pub fn get_verity_hash_value(path: &Path) -> Result<String> {
     let hash = verity::traverse_file(&mut file, 0, false, v, &mut verity::no_write)?;
     let result = format!("{:x}", hash);
 
-    Ok(result)
+    file.seek(std::io::SeekFrom::Start(0))?;
+
+    let mut passwd = String::new();
+    for entry_wrap in tar::Archive::new(file).entries()? {
+        let mut entry = entry_wrap?;
+        let entry_path = entry.header().path()?;
```
burgerdev, line 565: Do we really want to return early if these fail, or rather continue the loop?

src/tools/genpolicy/src/registry.rs
```diff
@@ -453,7 +557,23 @@ pub fn get_verity_hash_value(path: &Path) -> Result<String> {
...
+        let path_str = entry_path.to_str().unwrap();
+        if path_str == PASSWD_FILE_TAR_PATH {
+            entry.read_to_string(&mut passwd)?;
+            break;
+        } else if path_str == PASSWD_FILE_WHITEOUT_TAR_PATH {
+            passwd = WHITEOUT_MARKER.to_owned();
+            break;
+        }
```
burgerdev, line 573: Now I genuinely wonder whether the spec deals with both the file and a whiteout marker in the same layer...

src/tools/genpolicy/src/registry.rs
```diff
@@ -64,8 +65,37 @@ pub struct DockerRootfs {
...
     pub verity_hash: String,
+    pub passwd: String,
 }
 
+/// A single record in a Unix passwd file.
+#[derive(Debug)]
+struct PasswdRecord {
+    pub user: String,
+    #[allow(dead_code)]
+    pub validate: bool,
```
burgerdev, line 76: Nit: I believe the second field has three interpretations: login disabled, password login, passwordless. Doesn't matter much for this implementation, though.

src/tools/genpolicy/src/registry.rs
```diff
@@ -508,3 +628,31 @@ fn build_auth(reference: &Reference) -> RegistryAuth {
...
+                "Incorrect passwd record, expected 7 fields, got {}",
+                field_count
+            ));
+        }
+
+        records.push(PasswdRecord {
+            user: fields[0].to_string(),
+            validate: fields[1] == "x",
+            uid: fields[2].parse().unwrap(),
+            gid: fields[3].parse().unwrap(),
```
burgerdev, line 650: Did you mean to use the question-mark operator?

src/tools/genpolicy/src/registry.rs
```diff
@@ -508,3 +628,31 @@ fn build_auth(reference: &Reference) -> RegistryAuth {
 
     RegistryAuth::Anonymous
 }
+
+fn parse_passwd_file(passwd: String) -> Result<Vec<PasswdRecord>> {
```
burgerdev, line 632: I just enabled tests for genpolicy the other day - how about adding one for this parser?

burgerdev (commented):
Thanks a lot, this is a nice addition and matches what I had in mind when thinking about the problem.

src/tools/genpolicy/src/registry.rs
```diff
@@ -133,28 +163,87 @@ impl Container {
         debug!("Getting process field from docker config layer...");
         let docker_config = &self.config_layer.config;
 
+        // Defaults to start with if we can't work out the correct values.
```
danmihai1, line 166: Unless you are seeing test failures, let's remove these defaults - because the intended design was that the caller provides the default values.

src/tools/genpolicy/src/registry.rs
danmihai1: @burgerdev @HernanGatta , I think it's OK to file a github issue for future improvements and make those improvements later.

src/tools/genpolicy/src/registry.rs
danmihai1: This kind of error suggests to me that we're dealing with a corrupted tar file. So, returning early seems reasonable to me - at least until/unless we'll find examples that need workarounds.

src/tools/genpolicy/src/registry.rs
danmihai1: genpolicy panics on various paths when it deals with unexpected input format. So, the unwrap is consistent with the current design.

danmihai1 (commented):
@HernanGatta @burgerdev, the genpolicy code changes looks to me good enough to be merged. Thanks for the contribution, Hernan! I agree with @burgerdev that future improvements might be useful.

@HernanGatta, are you available to start looking at adding a simple test - maybe based on one of the images from https://quay.io/search?q=spark? A recent example of similar test: https://github.com/kata-containers/kata-containers/pull/10153.

If you are not available to work on that test soon, we could file an issue for it and either you or someone from MSFT can look at it a bit later.

We should also file issues as reminders to look into the other ideas for improvements from @burgerdev. 

@burgerdev, please speak up in case you think we need to delay the merge of this PR.

src/tools/genpolicy/src/registry.rs
burgerdev: I checked the [docs](https://docs.rs/tar/latest/tar/struct.Header.html#method.path), should be fine to fail here:

> This method may fail if the pathname is not valid Unicode and this is called on a Windows platform.

PR #9957, opened 2024-07-02
genpolicy: add support for ignoring sandbox name

### LINKED ISSUE #9956: genpolicy: ignore sandbox name in policy

The `genpolicy` tool should allow for generated sandbox names when given a YAML that specifies a `PodSpec`.

**Note:** I'll be submitting a PR for this shortly.

**How can it be improved?**

In some use cases, a user may feed a `PodSpec` YAML into `genpolicy` that is a representation of a pod that was automatically generated by an operator.

For example, the Spark Operator generates `PodSpec`s for Spark Driver pods whose name is generated. In other words, the `PodSpec` that will be created at runtime is not exactly the same as the `PodSpec` fed into `genpolicy`, differing only in its sandbox name.

**Additional Information**

The current ruleset allows for the string `$(generated-name)` in the sandbox name. So, the tool could accept a configuration that instructs it to ignore the name of a given `PodSpec` and put `$(generated-name)` in the resulting JSON data object instead.

✔️ Have you looked at the [limitations document](https://github.com/kata-containers/kata-containers/blob/main/docs/Limitations.md)?

**Kata Containers survey**

✔️ Please consider taking the survey to help us help you: https://openinfrafoundation.formstack.com/forms/kata_containers_user_survey

HernanGatta:
In some usecases, the workload that the policy is expected to apply to is generated dynamically. In other words, it may be impossible to determine, ahead of time, what the name of a pod will be. As a result, it is in turn also not possible to figure what the sandbox name will be.

Add an option that enables setting the expected sandbox name to be a generated name, so that any name is allowed.

Fixes #9956

### DIFF
--- a/src/tools/genpolicy/genpolicy-settings.json
+++ b/src/tools/genpolicy/genpolicy-settings.json
@@ -266,7 +266,8 @@
             "CAP_PERFMON",
             "CAP_BPF",
             "CAP_CHECKPOINT_RESTORE"
-        ]
+        ],
+        "ignore_sandbox_name": false
     },
     "kata_config": {
         "confidential_guest": false,
--- a/src/tools/genpolicy/src/policy.rs
+++ b/src/tools/genpolicy/src/policy.rs
@@ -369,6 +369,9 @@ pub struct CommonData {
 
     /// Default capabilities for a privileged container.
     pub privileged_caps: Vec<String>,
+
+    /// Do not check sandbox name (e.g., if they are generated).
+    pub ignore_sandbox_name: bool,
 }
 
 /// Configuration from "kubectl config".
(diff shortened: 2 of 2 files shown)


danmihai1:
@HernanGatta I will think more about this, and then let you know. My current thinking is that we should:

1. Introduce the concept of "Policy State".
2. The sandbox-name is empty in the State when starting a Pod.
3. When creating the first container, accept any input sandbox-name, and store that sandbox-name in the State.
4. When creating containers index 2+, verify that the input sandbox-name is the same as the value from the State.
5. Use the common sandbox-name for Policy rules the same way as before.
6. Remove the generation of the sandbox-state field from the genpolicy Rust code.

Stateful policy evaluation might help us for other future rules too.

PR #9959, opened 2024-07-02
ci: cleanup: Ignore nonexisting resources

sprt:
```
Some resource names seem to be lingering in Azure limbo but do not map
to any actual resources, so we ignore those.
```

### DIFF
--- a/tests/cleanup_resources.py
+++ b/tests/cleanup_resources.py
@@ -3,6 +3,7 @@
 from datetime import datetime, timedelta, timezone
 import os
 
+from azure.core.exceptions import ResourceNotFoundError
 from azure.identity import AzureCliCredential
 from azure.mgmt.resource import ResourceManagementClient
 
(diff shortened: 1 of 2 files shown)


sprt:
The workflow has been failing: https://github.com/kata-containers/kata-containers/actions/workflows/cleanup-resources.yaml

tests/cleanup_resources.py
```diff
@@ -44,12 +45,19 @@
     # /subscriptions/(subscriptionId)/resourceGroups/(resourceGroupName)/providers/(resourceProviderNamespace)/(resourceType)/(resourceName)
     rg_id, _, _ = resource.id.partition("/providers/")
     _, _, rg_name = rg_id.partition("/resourceGroups/")
-    rg_resources = client.resources.list_by_resource_group(rg_name)
+
+    try:
+        num_rg_resources = len(list(client.resources.list_by_resource_group(rg_name)))
+    except ResourceNotFoundError:
+        # Some resource names seem to be lingering in Azure limbo but do
+        # not map to any actual resources, so we ignore those.
```
stevenhorsman, line 53: I guess this is the reason for the error:
```
azure.core.exceptions.ResourceNotFoundError: (ResourceGroupNotFound) Resource group 'kataCI-k8s-nightly-9445a967b65d-clh-cbl-mariner-amd64-s' could not be found.
```
we are seeing? Any way to manually clean those up?

stevenhorsman (approved):
LGTM. Thanks @sprt

tests/cleanup_resources.py
sprt: So my understanding from @jepio is that some of these queries returned stale cache data. If you try to query the resource directly using a fully qualified name, you can't find the resource, so I'm pretty sure it doesn't exist anymore. Now I'm unsure how to actually invalidate the cache so I'm using this as a workaround. Might have to reach out to Azure Support here.

stevenhorsman: Cool - I have a feeling I've had the same in IBM Cloud before too :)

GabyCT (approved):

PR #9960, opened 2024-07-02
ci: Transition GARM tests to free runners, pt. I

sprt:
Transitioning the first batch of tests in `basic-ci-amd64.yaml`.

This is tracked in #9940.

### DIFF
--- a/.github/workflows/basic-ci-amd64.yaml
+++ b/.github/workflows/basic-ci-amd64.yaml
@@ -57,14 +57,12 @@ jobs:
         run: bash tests/integration/cri-containerd/gha-run.sh run
 
   run-containerd-stability:
-    # TODO: Transition to free runner (see #9940).
-    if: false
     strategy:
       fail-fast: false
       matrix:
         containerd_version: ['lts', 'active']
         vmm: ['clh', 'cloud-hypervisor', 'dragonball', 'qemu', 'stratovirt']
-    runs-on: garm-ubuntu-2204-smaller
+    runs-on: ubuntu-22.04
     env:
       CONTAINERD_VERSION: ${{ matrix.containerd_version }}
       GOPATH: ${{ github.workspace }}
--- a/.github/workflows/basic-ci-amd64.yaml
+++ b/.github/workflows/basic-ci-amd64.yaml
@@ -15,8 +15,6 @@ on:
 
 jobs:
   run-cri-containerd:
-    # TODO: Transition to free runner (see #9940).
-    if: false
     strategy:
       # We can set this to true whenever we're 100% sure that
       # the all the tests are not flaky, otherwise we'll fail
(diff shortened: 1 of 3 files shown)


.github/workflows/basic-ci-amd64.yaml
```diff
@@ -270,7 +260,7 @@ jobs:
         vmm:
           - clh
           - qemu
-    runs-on: garm-ubuntu-2304-smaller
+    runs-on: ubuntu-latest
```
sprt, line 263: @fidencio Any reason we used to use 23.04 here? Should we use the 24.04 runner?

.github/workflows/basic-ci-amd64.yaml
```diff
@@ -25,7 +23,7 @@ jobs:
       matrix:
         containerd_version: ['lts', 'active']
         vmm: ['clh', 'dragonball', 'qemu', 'stratovirt', 'cloud-hypervisor', 'qemu-runtime-rs']
-    runs-on: garm-ubuntu-2204-smaller
+    runs-on: ubuntu-latest
```
stevenhorsman, line 26: I wonder if we'd bee better off picking a pinned version (so `ubuntu-22.04` in this case) that would insulate us from potentially hitting a breakage that is very hard to understand when `ubuntu-24.04` comes out of beta? I know we haven't necessarily followed that approach elsewhere and it might need some effort to try bumping these, but that only happens every couple of years?

stevenhorsman (commented):
Just a general comment about pinning runners for stability, but other than that it look fine and we'll have to see how it works in practice once merged I guess

.github/workflows/basic-ci-amd64.yaml
wainersm: I agree with @stevenhorsman 's rationale and my preference is to pin to ubuntu-22.04

sprt: Agreed, I've run into that in other projects. Done!

stevenhorsman (approved):
LGTM. Thanks @sprt!

GabyCT (approved):

stevenhorsman:
This is a workflow only change that isn't tested by the PR checks as it is not triggered with a on/pull_request. It also has at least two reviews with the force-skip-ci label, therefore we can skip the CI and use admin rights to merge this.

stevenhorsman:
Hey @sprt - I guess you are aware, but the free runners are failing on the dependency install:
```
Unpacking podman-docker (3.4.4+ds1-1ubuntu1.22.04.2) ...
dpkg: error processing archive /tmp/apt-dpkg-install-UZ2jFw/7-podman-docker_3.4.4+ds1-1ubuntu1.22.04.2_amd64.deb (--unpack):
 trying to overwrite '/usr/bin/docker', which is also in package docker-ce-cli 5:26.1.3-1~ubuntu.22.04~jammy
Errors were encountered while processing:
 /tmp/apt-dpkg-install-UZ2jFw/7-podman-docker_3.4.4+ds1-1ubuntu1.22.04.2_amd64.deb
needrestart is being skipped since dpkg has failed
E: Sub-process /usr/bin/dpkg returned an error code (1)
Error: Process completed with exit code 100.
```

Issue #9934, opened 2024-06-28
Leverage ACTIONS_RUNNER_HOOK_JOB_{STARTED,COMPLETED}


BbolroC:
**Which feature do you think can be improved?**

For self-hosted runners, especially those on s390x and ppc64le, the following steps are used to manage the runners:

- [x] s390x: {pre,post}_action.sh
- [ ] ppc64le: {prepare,cleanup}_runner.sh

A GitHub introduced hook, [`ACTIONS_RUNNER_HOOK_JOB_{STARTED,COMPLETED}`](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job#triggering-the-scripts), can not only achieve the same goal but also hide steps that are not relevant to the workflow itself.

CC: @Amulyam24

BbolroC:
As a separate topic, but worth thinking for implementation, we could share and track parts of the scripts that are commonly useful while hiding other parts through a function call in the hook.

wainersm:
Hi @BbolroC !

> **Which feature do you think can be improved?**
> 
> For self-hosted runners, especially those on s390x and ppc64le, the following steps are used to manage the runners:
> 
>     * s390x: {pre,post}_action.sh
> 
>     * ppc64le: {prepare,cleanup}_runner.sh
> 
> 
> A GitHub introduced hook, [`ACTIONS_RUNNER_HOOK_JOB_{STARTED,COMPLETED}`](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job#triggering-the-scripts), can not only achieve the same goal but also hide steps that are not relevant to the workflow itself.
> 
> CC: @Amulyam24

That's a good idea. It will reduce the size (one step less) of some workflows. Also if you need to add pre/post steps for other workflows you won't need to change the workflow yaml then pass the painful PR experience.

One thing that just occurred me is that those hooks are called for every workflow. Let's suppose you got workflow `A` and `B`, you got a `cleanup.sh` script that should be executed before `A`. Now suppose that running `cleanup.sh` on `B` will cause problems (try to clean up something that wasn't created, for example). Do you know if github is going to export the normal workflow environment variables (i.e. https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) in same environment as `cleanup.sh` executes, so that you can selectively run `cleanup.sh`'s procedures when `A` and not `B`?

[...]

PR #9964, opened 2024-07-04
ci: Cleanup working dir when using self-hosted runners

ananos:
We get permission issues when the current workspace directory is not clean. This is the case for self-hosted runners. This patch cleans the workspace dir before the checkout. An alternative would be to make sure we clean it as a last step of each job.


GabyCT (approved):

zvonkok:
@fidencio This is in the works by https://github.com/kata-containers/kata-containers/issues/9934  this is to unblock the CI and once @BbolroC is done with the POC we will create something as you already said for "ALL" selfhosted-runners

BbolroC:
Please have a look at https://github.com/kata-containers/kata-containers/issues/9934#issuecomment-2258094441 if you follow Fabi's suggestion. Thanks!

PR #9965, opened 2024-07-04
runtime-rs: add memory hotplugging support to qemu-rs

pmores:
A rather standard memory hotplugging implementation, the only slightly unusual thing about it is that it makes a best-effort attempt at hotunplugging, too.  While hotunplugging memory seems to fail generally I'd like to have it in the code base at least for a while for observation.  It does no harm either.

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -334,19 +334,31 @@ impl QemuInner {
         todo!()
     }
 
-    pub(crate) fn set_guest_memory_block_size(&mut self, _size: u32) {
-        warn!(
-            sl!(),
-            "QemuInner::set_guest_memory_block_size(): NOT YET IMPLEMENTED"
-        );
+    pub(crate) fn set_guest_memory_block_size(&mut self, size: u32) {
+        if let Some(ref mut qmp) = self.qmp {
+            info!(
+                sl!(),
+                "QemuInner::set_guest_memory_block_size(): block size set to {}", size
+            );
+            qmp.set_guest_memory_block_size(size.into());
+        } else {
+            warn!(
+                sl!(),
+                "QemuInner::set_guest_memory_block_size(): QMP not initialized"
+            );
+        }
     }
 
-    pub(crate) fn guest_memory_block_size_mb(&self) -> u32 {
-        warn!(
-            sl!(),
-            "QemuInner::guest_memory_block_size_mb(): NOT YET IMPLEMENTED"
-        );
-        0
+    pub(crate) fn guest_memory_block_size(&self) -> u32 {
+        if let Some(qmp) = &self.qmp {
+            qmp.guest_memory_block_size() as u32
+        } else {
+            warn!(
+                sl!(),
+                "QemuInner::guest_memory_block_size(): QMP not initialized"
+            );
+            0
+        }
     }
 
     pub(crate) fn resize_memory(&self, _new_mem_mb: u32) -> Result<(u32, MemoryConfig)> {
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -19,6 +19,7 @@ use kata_types::{
 use persist::sandbox_persist::Persist;
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::path::Path;
 use std::process::Stdio;
 use tokio::{
@@ -466,6 +467,18 @@ impl QemuInner {
                     sl!(),
                     "hotunplugging {} B of memory",
                     cur_hotplugged_memory - new_hotplugged_mem
+                );
+                let res =
+                    qmp.hotunplug_memory((cur_hotplugged_memory - new_hotplugged_mem).try_into()?);
+                if let Err(err) = res {
+                    info!(sl!(), "hotunplugging failed: {:?}", err);
+                } else {
+                    new_total_mem_mb = bytes_to_megs(coldplugged_mem + new_hotplugged_mem);
+                }
+                info!(
+                    sl!(),
+                    "hotplugged memory after hotunplugging: {}",
+                    qmp.hotplugged_memory_size()?
                 );
             }
             Ordering::Equal => info!(
--- a/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
@@ -361,14 +361,120 @@ impl QemuInner {
         }
     }
 
-    pub(crate) fn resize_memory(&self, _new_mem_mb: u32) -> Result<(u32, MemoryConfig)> {
-        warn!(sl!(), "QemuInner::resize_memory(): NOT YET IMPLEMENTED");
-        Ok((
-            _new_mem_mb,
-            MemoryConfig {
-                ..Default::default()
-            },
-        ))
+    pub(crate) fn resize_memory(
+        &mut self,
+        mut new_total_mem_mb: u32,
+    ) -> Result<(u32, MemoryConfig)> {
+        info!(
+            sl!(),
+            "QemuInner::resize_memory(): asked to resize memory to {} MB", new_total_mem_mb
+        );
+
+        // stick to the apparent de facto convention and represent megabytes
+        // as u32 and bytes as u64
+        fn bytes_to_megs(bytes: u64) -> u32 {
+            (bytes / (1 << 20)) as u32
+        }
+        fn megs_to_bytes(bytes: u32) -> u64 {
+            bytes as u64 * (1 << 20)
+        }
+
+        let qmp = match self.qmp {
+            Some(ref mut qmp) => qmp,
+            None => {
+                warn!(sl!(), "QemuInner::resize_memory(): QMP not initialized");
+                return Err(anyhow!("QMP not initialized"));
+            }
+        };
+
+        let coldplugged_mem = megs_to_bytes(self.config.memory_info.default_memory);
+        let new_total_mem = megs_to_bytes(new_total_mem_mb);
+
+        if new_total_mem < coldplugged_mem {
+            return Err(anyhow!(
+                "asked to resize to {} M but that is less than cold-plugged memory size ({})",
+                new_total_mem_mb,
+                bytes_to_megs(coldplugged_mem)
+            ));
+        }
+
+        let guest_mem_block_size = qmp.guest_memory_block_size();
+
+        let mut new_hotplugged_mem = new_total_mem - coldplugged_mem;
+
+        info!(
+            sl!(),
+            "new hotplugged mem before alignment: {} B ({} MB)",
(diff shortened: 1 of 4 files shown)


src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
```diff
@@ -128,6 +142,153 @@ impl Qmp {
...
+                    hotplugged_mem_size += dimm_info.data.size as u64;
+                }
+            }
+        }
+
+        Ok(hotplugged_mem_size)
+    }
+
+    pub fn hotplug_memory(&mut self, size: u64) -> Result<()> {
+        let memdev_idx = self
```
BbolroC, line 185: What about renaming it to `hotplugged_memdev_count `? I know it will be used to construct a new id below, but the code itself ends with `count()`. 😉

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
```diff
@@ -128,6 +142,153 @@ impl Qmp {
...
+            .into_iter()
+            .find(|memdev| {
+                if let qapi_qmp::MemoryDeviceInfo::dimm(dimm_info) = memdev {
+                    let dimm_id = match dimm_info.data.id {
+                        Some(ref id) => id,
+                        None => return false,
+                    };
+                    if dimm_info.data.hotpluggable
+                        && dimm_info.data.hotplugged
+                        && dimm_info.data.size == size
```
BbolroC, line 252: Just out of curiosity: should the size of memory requested for unplug be the same as that of memory for the plugged device?

BbolroC (approved):
Thanks for your great work. Overall it looks good to me. I've left some naive questions. 

I think the 2nd reviewer should be a QEMU expert.

src/runtime-rs/crates/hypervisor/src/qemu/inner.rs
```diff
@@ -334,29 +335,159 @@ impl QemuInner {
...
+            (bytes / (1 << 20)) as u32
+        }
+        fn megs_to_bytes(bytes: u32) -> u64 {
+            bytes as u64 * (1 << 20)
+        }
+
+        let qmp = match self.qmp {
+            Some(ref mut qmp) => qmp,
+            None => {
+                warn!(sl!(), "QemuInner::resize_memory(): no QMP connection");
```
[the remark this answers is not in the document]
pmores: Good point.  I guess I was worried that any given wording could be misleading under some circumstances and ended up with this. :-)

Actually in the hindsight both existing variants seem to me to suggest more than we know at the point where we log the message.  `QMP not yet initialized` sounds like we haven't tried yet, `no QMP connection` kind of sounds like we tried and failed... How about unifying all with a simple `QMP not initialized`?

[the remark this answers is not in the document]
BbolroC: Yeah, the same phenomenon, but two different context could end up with your log messages. I am fine with it. I just wanted to your behind-the-scene. 😉

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
```diff
@@ -128,6 +142,153 @@ impl Qmp {
...
+    pub fn guest_memory_block_size(&self) -> u64 {
+        self.guest_memory_block_size
+    }
+
+    pub fn hotplugged_memory_size(&mut self) -> Result<u64> {
+        let memory_frontends = self.qmp.execute(&qapi_qmp::query_memory_devices {})?;
+
+        let mut hotplugged_mem_size = 0_u64;
+
+        info!(sl!(), "hotplugged_memory_size(): iterating over dimms");
```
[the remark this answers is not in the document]
pmores: The idea here (and even more generally) is to log the "what" is being done rather than "how" if possible.  While it's technically true that we're iterating over all frontends at the highest level, we only do that because it's not possible (or practical) to directly iterate over dimms exclusively, and so I'd consider that an implementation detail of the iteration.

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
pmores: Yeah, well, conceptually the number starts as a count and ends as an index, so much is sure.  Now the question is where exactly the transition happens.  I put it where I did I guess precisely because the line ends in `count()` but stores in `idx` - I kind of like how explicit the transition in the number's meaning is this way. :-)

The alternative would be calling the variable `memdev_count` and then have
```
let memory_backend_id = format!("hotplugged-{}", memdev_count);
```
Is that better though?

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
pmores: I'm not overly knowledgeable about the fine points of this area, however it appears obvious from the nature of hotunplugging implementation that it's impossible to hotunplug an arbitrary amount of memory.  So the sizes need to match and it appears to be a user responsibility to ensure that that's the case.

This is fulfilled in runtime-rs because runtime-rs hotplugs memory requests of individual containers and then hotunplugs them again so match is guaranteed.  E.g. if a container in a pod requests 1G and another containers in the same pod requests 2G, runtime-rs will hotplug 1G and then 2G on launch and then hotunplug 1G and then 2G on pod termination.

The matching by size works in practice, a memory block to hotunplug is always found on the first hotunplug attempt.  The fact that hotunplugging it usually fails is more of a problem.  If that happens subsequent hotunplugging requests don't even find a matching-size block anymore.

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
BbolroC: Yeah, it was just an opinion. I will follow your decision. At least, I was confused when I firstly read this line and got it clear at the next line. Let's see what the 2nd reviewer would view this.

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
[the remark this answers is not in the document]
BbolroC: Good, and let's leave it as it is. 😉

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
BbolroC: Oh, it makes sense. The pod example definitely boosts my understanding. Thanks!

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
pmores: Thanks.  There will always be some friction I suspect since fundamentally, we're taking a count and casting it into an index.

src/runtime-rs/crates/hypervisor/src/qemu/mod.rs
```diff
@@ -173,11 +173,11 @@ impl Hypervisor for Qemu {
 
     async fn guest_memory_block_size(&self) -> u32 {
         let inner = self.inner.read().await;
-        inner.guest_memory_block_size_mb()
+        inner.guest_memory_block_size()
     }
 
     async fn resize_memory(&self, new_mem_mb: u32) -> Result<(u32, MemoryConfig)> {
-        let inner = self.inner.read().await;
+        let mut inner = self.inner.write().await;
```
gkurz, line 180: I'm getting this build failure with rust 1.75 :
```
error: variable does not need to be mutable
   --> crates/hypervisor/src/qemu/mod.rs:180:13
    |
180 |         let mut inner = self.inner.write().await;
    |             ----^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `-D unused-mut` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_mut)]`
```

Compiler is right AFAICT  :wink:
```
    pub(crate) fn resize_memory(&self, _new_mem_mb: u32) -> Result<(u32, MemoryConfig)> {
```

pmores: Right, the signature of `QemuInner::resize_memory()` is changed in the third commit where its implementation is added... this hunk can go to that commit if need be.

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
pmores: It is safe to assume that existing ids are < `count()` since the indexing is part of a naming scheme that's internal to the driver and owned by it (that's the idea anyway :-)).

As for not maintaining it on the shim side, this is actually part of a conscious decision to treat QEMU as an authoritative source of information.  Considering that QEMU keeps track of this and also gives us a convenient mechanism to query it, it doesn't seem beneficial (to me anyway) to write & maintain extra code in the runtime to track the very same information, even introducing a danger of going out of sync (say due to a bug in the runtime-side accounting).

pmores:
All review requests should be fixed now.

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
gkurz: Fair enough.

pmores:
Force-push to placate formatter - I formatted before submission but CI's formatter still managed somehow to crawl out of the woodwork and find something. <shrug>

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
```diff
@@ -235,6 +235,62 @@ impl Qmp {
...
+                    Some(id) => id.to_owned(),
+                    // This shouldn't happen as it was checked by find() above already.
+                    None => {
+                        return Err(anyhow!(
+                        "memory backend to hotunplug has id that doesn't have the expected prefix"
+                    ))
+                    }
+                };
+
+                self.qmp.execute(&qmp::device_del { id: frontend_id })?;
```
gkurz, line 281: I have the impression that this can punch a hole in the index range and cause next hotplug to try to reuse an existing id... but I could be wrong. In any case, not to be fixed now since this is experimental.

gkurz (approved):
I'll let you fix the `cargo fmt` complain or not since being able to bisect is maybe not your priority at this stage.

Let's go forward.

Thanks for your hard work and patience, Pavel !

src/runtime-rs/crates/hypervisor/src/qemu/qmp.rs
pmores: I _think_ you're right but I don't think it would be a problem.  If a hole is created by removing a memdev then the id is not "existing" anymore.  The whole idea here is to have device names that 1) allow us to tell devices that we created (by hotplugging) from any other dimms, and 2) allow to create several memdevs with names that are still unique.  The id serves the second point.

And yeah, that all might be moot since hotunplugging seems to fail fairly consistently currently and if there isn't a successful effort in the future to support hotunplugging in the agent this function might even be removed altogether.

gkurz: Yeah as discussed some times ago, memory hot-unplug requires some care from the guest kernel : it shouldn't place some long living and not movable structure in hot-plugged memory.

PR #9967, opened 2024-07-04
dragonball: kernel dragonball 6.1.x

zvonkok:
Use the latest guest kernel that we use for all other VMMs


GabyCT (approved):

lifupan (approved):
Thanks @zvonkok, LGTM.

Fupan

PR #9971, opened 2024-07-05
gha: Fix pip installation for nerdctl GHA

GabyCT:
This PR fixes the pip installation for nerdctl by removing a flag which is not longer supported and avoid the failure of no such option: --break-system-packages.


GabyCT:
We can see the error for example in the following GHA job
https://github.com/kata-containers/kata-containers/actions/runs/9801102320/job/27064266618

dborquez (approved):
LGTM, thank you @GabyCT

fidencio (approved):
lgtm, thanks @GabyCT

PR #9973, opened 2024-07-05
Add memory and vcpus info to metrics results

dborquez:


GabyCT (approved):

tests/metrics/lib/common.bash
```diff
@@ -505,6 +508,28 @@ function get_current_kata_config_file() {
...
 function check_if_root() {
 	[ "$EUID" -ne 0 ] && die "Please run as root or use sudo."
 }
+
+# This function launches a kata container using a Busybox image,
+# then collects the current number of vcpus and the free memory from the container.
+# Finalliy fullfills the global variables 'MEASURED_CONTAINER_NUM_VCPUS' and 'MEASURED_CONTAINER_FREE_MEM'
+function get_kata_memory_and_vcpus() {
+	local busybox_img="docker.io/library/busybox:latest"
+	local container_name="kata-busybox"
```
danmihai1, line 522: Would it be worth adding a random component to this container name - just in case we'll ever have two instances of this function running in parallel on the same host?

tests/metrics/lib/common.bash
```diff
@@ -505,6 +508,28 @@ function get_current_kata_config_file() {
...
+# if that is not the case, the function exits with an error message.
 function check_if_root() {
 	[ "$EUID" -ne 0 ] && die "Please run as root or use sudo."
 }
+
+# This function launches a kata container using a Busybox image,
+# then collects the current number of vcpus and the free memory from the container.
+# Finalliy fullfills the global variables 'MEASURED_CONTAINER_NUM_VCPUS' and 'MEASURED_CONTAINER_FREE_MEM'
+function get_kata_memory_and_vcpus() {
+	local busybox_img="docker.io/library/busybox:latest"
```
danmihai1, line 521: Is quay.io less restrictive than docker.io when it comes to pulling images without logging in?

danmihai1 (approved):

tests/metrics/lib/common.bash
dborquez: @danmihai1 thank you, you are correct, so I update the PR with your suggestion.

dborquez:
@stevenhorsman thank you for the finding, I opened the issue #9991, WIP to fix it.

tests/common.bash
```diff
@@ -159,12 +163,28 @@ function extract_kata_env() {
 			hypervisor_path=".Hypervisor.Path"
 			virtio_fs_daemon_path=".Hypervisor.VirtioFSDaemon"
 			initrd_path=".Initrd.Path"
+			shared_fs=".Hypervisor.SharedFS"
```
stevenhorsman, line 166: Hi @dborquez - was there are reason that `shared_fs` wasn't set on dragonball as this is also causing failures with `shared_fs: unbound variable` in https://github.com/kata-containers/kata-containers/actions/runs/9900211956/job/27350849536

PR #9868, opened 2024-06-18
runtime: implement DAN in Go kata-runtime

l8huang:
The DAN feature has already been implemented in kata-runtime-rs, and
this commit brings the same capability to the Go kata-runtime.

Fixes: https://github.com/kata-containers/kata-containers/issues/9758

PR #9977, opened 2024-07-08
runtime: add DAN support for VFIO network device in Go kata-runtime

l8huang:
When using network adapters that support SR-IOV, a VFIO device can be
plugged into a guest VM and claimed as a network interface. This can
significantly enhance network performance.

Fixes: #9758

### DIFF
--- a/src/runtime/virtcontainers/endpoint.go
+++ b/src/runtime/virtcontainers/endpoint.go
@@ -65,6 +65,13 @@ const (
 
 	// IPVlanEndpointType is ipvlan network interface.
 	IPVlanEndpointType EndpointType = "ipvlan"
+
+	// VfioEndpointType is a VFIO device that will be claimed as a network interface
+	// in the guest VM. Unlike PhysicalEndpointType, which requires a VF network interface
+	// with its network configured on the host before creating the sandbox, VfioEndpointType
+	// does not need a host network interface and instead has its network network configured
+	// through DAN.
+	VfioEndpointType EndpointType = "vfio"
 )
 
 // Set sets an endpoint type based on the input string.
@@ -94,6 +101,9 @@ func (endpointType *EndpointType) Set(value string) error {
 	case "ipvlan":
 		*endpointType = IPVlanEndpointType
 		return nil
+	case "vfio":
+		*endpointType = VfioEndpointType
+		return nil
 	default:
 		return fmt.Errorf("Unknown endpoint type %s", value)
 	}
@@ -118,6 +128,8 @@ func (endpointType *EndpointType) String() string {
 		return string(TuntapEndpointType)
 	case IPVlanEndpointType:
 		return string(IPVlanEndpointType)
+	case VfioEndpointType:
+		return string(VfioEndpointType)
 	default:
 		return ""
 	}
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -782,20 +782,10 @@ func (k *kataAgent) startSandbox(ctx context.Context, sandbox *Sandbox) error {
 
 	if sandbox.config.HypervisorType != RemoteHypervisor {
 		// Setup network interfaces and routes
-		interfaces, routes, neighs, err := generateVCNetworkStructures(ctx, sandbox.network)
+		err = k.setupNetworks(ctx, sandbox, nil)
 		if err != nil {
 			return err
 		}
-		if err = k.updateInterfaces(ctx, interfaces); err != nil {
-			return err
-		}
-		if _, err = k.updateRoutes(ctx, routes); err != nil {
-			return err
-		}
-		if err = k.addARPNeighbors(ctx, neighs); err != nil {
-			return err
-		}
-
 		kmodules = setupKernelModules(k.kmodules)
 	}
 
@@ -1249,6 +1239,67 @@ func (k *kataAgent) rollbackFailingContainerCreation(ctx context.Context, c *Con
 	}
 }
 
+func (k *kataAgent) setupNetworks(ctx context.Context, sandbox *Sandbox, c *Container) error {
+	if sandbox.network.NetworkID() == "" {
+		return nil
+	}
+
+	var err error
+	var endpoints []Endpoint
+	if c == nil || c.id == sandbox.id {
+		// TODO: VFIO network deivce has not been hotplugged when creating the Sandbox,
+		// so need to skip VFIO endpoint here.
+		// After KEP #4113(https://github.com/kubernetes/enhancements/pull/4113)
+		// is implemented, the VFIO network devices will be attached before container
+		// creation, so no need to skip them here anymore.
+		for _, ep := range sandbox.network.Endpoints() {
+			if ep.Type() != VfioEndpointType {
+				endpoints = append(endpoints, ep)
+			}
+		}
+	} else if !sandbox.hotplugNetworkConfigApplied {
+		// Apply VFIO network devices' configuration after they are hot-plugged.
+		for _, ep := range sandbox.network.Endpoints() {
+			if ep.Type() == VfioEndpointType {
+				hostBDF := ep.(*VfioEndpoint).HostBDF
+				pciPath := sandbox.GetVfioDeviceGuestPciPath(hostBDF)
+				if pciPath.IsNil() {
+					return fmt.Errorf("PCI path for VFIO interface '%s' not found", ep.Name())
+				}
+				ep.SetPciPath(pciPath)
+				endpoints = append(endpoints, ep)
+			}
+		}
+
+		defer func() {
+			if err == nil {
+				sandbox.hotplugNetworkConfigApplied = true
+			}
+		}()
+	}
+
+	if len(endpoints) == 0 {
+		return nil
+	}
+
+	interfaces, routes, neighs, err := generateVCNetworkStructures(ctx, endpoints)
+	if err != nil {
+		return err
+	}
+
+	if err = k.updateInterfaces(ctx, interfaces); err != nil {
+		return err
+	}
+	if _, err = k.updateRoutes(ctx, routes); err != nil {
+		return err
+	}
+	if err = k.addARPNeighbors(ctx, neighs); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Container) (p *Process, err error) {
 	span, ctx := katatrace.Trace(ctx, k.Logger(), "createContainer", kataAgentTracingTags)
 	defer span.End()
@@ -1393,6 +1444,11 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
 	if _, err = k.sendReq(ctx, req); err != nil {
 		return nil, err
 	}
+
+	if err = k.setupNetworks(ctx, sandbox, c); err != nil {
+		return nil, err
+	}
+
 	return buildProcessFromExecID(req.ExecId)
 }
 
--- a/src/runtime/virtcontainers/network.go
+++ b/src/runtime/virtcontainers/network.go
@@ -233,10 +233,8 @@ type Network interface {
 	GetEndpointsNum() (int, error)
 }
 
-func generateVCNetworkStructures(ctx context.Context, network Network) ([]*pbTypes.Interface, []*pbTypes.Route, []*pbTypes.ARPNeighbor, error) {
-	if network.NetworkID() == "" {
-		return nil, nil, nil, nil
-	}
+func generateVCNetworkStructures(ctx context.Context, endpoints []Endpoint) ([]*pbTypes.Interface, []*pbTypes.Route, []*pbTypes.ARPNeighbor, error) {
+
 	span, _ := networkTrace(ctx, "generateVCNetworkStructures", nil)
 	defer span.End()
 
@@ -244,7 +242,7 @@ func generateVCNetworkStructures(ctx context.Context, network Network) ([]*pbTyp
 	var ifaces []*pbTypes.Interface
 	var neighs []*pbTypes.ARPNeighbor
 
-	for _, endpoint := range network.Endpoints() {
+	for _, endpoint := range endpoints {
 		var ipAddresses []*pbTypes.IPAddress
 		for _, addr := range endpoint.Properties().Addrs {
 			// Skip localhost interface
@@ -270,6 +268,7 @@ func generateVCNetworkStructures(ctx context.Context, network Network) ([]*pbTyp
 			Device:      endpoint.Name(),
 			Name:        endpoint.Name(),
 			Mtu:         uint64(endpoint.Properties().Iface.MTU),
+			Type:        string(endpoint.Type()),
 			RawFlags:    noarp,
 			HwAddr:      endpoint.HardwareAddr(),
 			PciPath:     endpoint.PciPath().String(),
--- a/src/runtime/virtcontainers/network_linux.go
+++ b/src/runtime/virtcontainers/network_linux.go
@@ -436,6 +436,10 @@ func convertDanDeviceToNetworkInfo(device *vctypes.DanDevice) (*NetworkInfo, err
 // Load network config in DAN config
 // Create the endpoints for the interfaces in Dan.
 func (n *LinuxNetwork) addDanEndpoints() error {
+	if len(n.eps) > 0 {
+		// only load DAN config once
+		return nil
+	}
 
 	jsonData, err := os.ReadFile(n.danConfigPath)
 	if err != nil {
@@ -452,20 +456,21 @@ func (n *LinuxNetwork) addDanEndpoints() error {
 		var endpoint Endpoint
 		networkLogger().WithField("interface", device.Name).Info("DAN interface found")
 
-		_, err := convertDanDeviceToNetworkInfo(&device)
+		netInfo, err := convertDanDeviceToNetworkInfo(&device)
 		if err != nil {
 			return err
 		}
 
-		// TODO: Add endpoints that are supported via DAN
 		switch device.Device.Type {
+		case vctypes.VfioDanDeviceType:
+			endpoint, err = createVfioEndpoint(device.Device.PciDeviceID, netInfo)
+			if err != nil {
+				return err
+			}
 		default:
 			return fmt.Errorf("unknown DAN device type: '%s'", device.Device.Type)
 		}
 
-		// TODO: remove below `nolink` directive after one `case` is added for
-		// above `switch` block.
-		//nolint: govet
 		n.eps = append(n.eps, endpoint)
 	}
 
--- a/src/runtime/virtcontainers/persist/api/network.go
+++ b/src/runtime/virtcontainers/persist/api/network.go
@@ -79,6 +79,10 @@ type VhostUserEndpoint struct {
 	PCIPath   vcTypes.PciPath
 }
 
+type VfioEndpoint struct {
+	IfaceName string
+}
+
 // NetworkEndpoint contains network interface information
 type NetworkEndpoint struct {
 	// One and only one of these below are not nil according to Type.
@@ -90,6 +94,7 @@ type NetworkEndpoint struct {
 	Tap       *TapEndpoint       `json:",omitempty"`
 	IPVlan    *IPVlanEndpoint    `json:",omitempty"`
 	Tuntap    *TuntapEndpoint    `json:",omitempty"`
+	Vfio      *VfioEndpoint      `json:",omitempty"`
 
 	Type string
 }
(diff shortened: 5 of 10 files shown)


l8huang:
I'm uncertain when will https://github.com/kata-containers/kata-containers/pull/9868 be merged, so I've created this PR to initiate the review process earlier.

@justxuewei could you please take a look at the 2nd commit in this PR? Thanks

justxuewei:
> > I'm uncertain when will #9868 be merged, so I've created this PR to initiate the review process earlier.
> > @justxuewei could you please take a look at the 2nd commit in this PR? Thanks
> 
> I'd like confirm that, the first commit in this PR equals to the [PR#9868](https://github.com/kata-containers/kata-containers/pull/9868) , right ?

Yeah. Once the first PR get merged, this PR only contains one commit.

Apokleos:
> > > I'm uncertain when will #9868 be merged, so I've created this PR to initiate the review process earlier.
> > > @justxuewei could you please take a look at the 2nd commit in this PR? Thanks
> > 
> > 
> > I'd like confirm that, the first commit in this PR equals to the [PR#9868](https://github.com/kata-containers/kata-containers/pull/9868) , right ?
> 
> Yeah. Once the first PR gets merged, this PR only contains one commit.
>
Ok, Let's push the first one first

l8huang:
@justxuewei did you get a chance to review this PR?

justxuewei:
> @justxuewei @Apokleos   did you get a chance to review this PR?

I'm on vacation, so I'll review this PR next week. Is it ok to you?

l8huang:
> I'm on vacation, so I'll review this PR next week. Is it ok to you?

sure, have a good vacation, thanks.

src/runtime/virtcontainers/persist/api/network.go
```diff
@@ -90,6 +94,7 @@ type NetworkEndpoint struct {
 	Tap       *TapEndpoint       `json:",omitempty"`
 	IPVlan    *IPVlanEndpoint    `json:",omitempty"`
 	Tuntap    *TuntapEndpoint    `json:",omitempty"`
+	Vfio      *VfioEndpoint      `json:",omitempty"`
```
amshinde, line 97: @l8huang We currently have `PhysicalEndpoint` to support and detect SRIOV network interface or an entire physical network interface. 
Is that different from your use-case?

l8huang: PhysicalEndpoint is for VF device, which presents a network interface on host; this PR is for VFIO passthrough device, on host it doesn't has a network interface, it supposes to be plugged into guest VM and recognized by NIC driver in guest kernel to create its network interface.

l8huang:
@justxuewei did you get a chance to review this PR? hope you had a great vacation.

justxuewei:
> @justxuewei did you get a chance to review this PR? hope you had a great vacation.

Thanks. I'll take a look this week ;)

justxuewei:
Could you take a loot at this PR? @Apokleos

src/runtime/virtcontainers/network.go
```diff
@@ -233,18 +233,16 @@ type Network interface {
 	GetEndpointsNum() (int, error)
 }
 
-func generateVCNetworkStructures(ctx context.Context, network Network) ([]*pbTypes.Interface, []*pbTypes.Route, []*pbTypes.ARPNeighbor, error) {
-	if network.NetworkID() == "" {
-		return nil, nil, nil, nil
-	}
```
justxuewei, line 239: I think you cannot just simply remove it, because this changes the original logic. I’d like to suggest adding a new argument, `enable_netns bool`. If VFIO is enabled, or in some scenarios where you don't want to use netns, set the value to `false`.

```
if enable_netns && network.NetworkID() == "" {
	return nil, nil, nil, nil
}
```

justxuewei (commented):
Lgtm, but two nits.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1393,6 +1434,12 @@ func (k *kataAgent) createContainer(ctx context.Context, sandbox *Sandbox, c *Co
 	if _, err = k.sendReq(ctx, req); err != nil {
 		return nil, err
 	}
+
+	err = k.setupNetworks(ctx, sandbox, c)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
l8huang: updated

src/runtime/virtcontainers/network.go
l8huang: good point. Because `generateVCNetworkStructures()` is wrapped by `setupNetworks()` function now, I added the old check back to `setupNetworks()`.

l8huang (commented):
@justxuewei  thanks for reviewing this PR, I updated it according to your comments, PTAL.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1249,6 +1239,61 @@ func (k *kataAgent) rollbackFailingContainerCreation(ctx context.Context, c *Con
 	}
 }
 
+func (k *kataAgent) setupNetworks(ctx context.Context, sandbox *Sandbox, c *Container) error {
+	if sandbox.network.NetworkID() == "" {
```
l8huang, line 1243: @justxuewei the check you mentioned is added at here.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1249,6 +1239,61 @@ func (k *kataAgent) rollbackFailingContainerCreation(ctx context.Context, c *Con
...
+		}
+	} else if !sandbox.hotplugNetworkConfigApplied {
+		// Apply VFIO network devices' configuration after they are hot-plugged.
+		for _, ep := range sandbox.network.Endpoints() {
+			if ep.Type() == VfioEndpointType {
+				hostBFD := ep.(*VfioEndpoint).HostBDF
+				pciPath := sandbox.GetVfioDeviceGuestPciPath(hostBFD)
+				if pciPath.IsNil() {
+					return fmt.Errorf("PCI path for VFIO interface '%s' not found", ep.Name())
+				}
```
justxuewei, line 1267: Are any rollback measures needed here? Assuming that there are 4 VFIO endpoints.

1. Added 2 VFIO eps.
2. Some errors happen at 3 ep.
3. Rollback the above 2 eps, otherwise they might be added twice in the next call to `setupNetworks`.

justxuewei (approved):
Lgtm, thanks!

l8huang:
> Lgtm, thanks!

Cool. @Apokleos @lifupan could you please take 2nd look? thanks.

src/runtime/virtcontainers/endpoint.go
```diff
@@ -65,6 +65,9 @@ const (
 
 	// IPVlanEndpointType is ipvlan network interface.
 	IPVlanEndpointType EndpointType = "ipvlan"
+
+	// VfioEndpointType is VFIO device claimed as network interface by guest kernel
+	VfioEndpointType EndpointType = "vfio"
```
amshinde, line 70: The name and description here is quite confusing with `PhysicalEndpointType`.  `PhysicalEndpointType` was meant to represent an entire physical device or SRIOV interface that could be passed through VFIO. 
To clearly distinguish between the two, I suggest we should consider changing the naming here, maybe rename this to `DanEndPointType` or `DanVfioEndpointType`.
I also suggest to change the documentation comments to explicitly mention Direct attached networks. Without this, it will be a tall task for future developers to distinguish between the two endpoints.

amshinde: I will take a detailed look at this PR by tomorrow to provide any feedback.

l8huang: Thanks for looking at this PR.

If the name `VfioEndpointType` causing confusion, **I'm okay with `DanVfioEndpointType`** -- `DanEndPointType` looks like too generic. @justxuewei are you okay with `DanVfioEndpointType`? 

BTW: I saw `PhysicalEndpointType` unbinds VF device from its driver then rebind it to `vfio-pci` and cold-plug it to VM. When using ovn-kubernetes, the VF and VFIO devices are pre-allocated on the host, based on the network config on a workload(VM/Pod), a VF or VFIO device can be assigned to it accordingly -- a VFIO  device can be directly assigned to a VM workload, so no need to allocate and config a VF device then use PhysicalEndpoint to unbind/rebind it as VFIO, see https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/cni/helper_linux.go#L293-L301. 

One question out of curiosity: In the use case of PhysicalEndpoint, which CNI is typically utilized?

l8huang: @amshinde do you have further comments for this PR?

l8huang: @amshinde I added a doc string to clarify the difference from PhysicalEndpointType. I also renaming,  but since the endpoint type is inherently VFIO and DAN is merely the configuration method, renaming it to DanVfioEndpointType might introduce unnecessary confusion. Even if we renamed it, the value would still be 'vfio'. Therefore, I prefer to keep the current name. Does that work for you?"

justxuewei: I don't think adding DAN is a good idea, since DAN stands for Directy Attachable Network. It just a new way of reading network config, and is not relevant for endpoints.

l8huang: @amshinde do you have any further feedback on this PR?

amshinde: > One question out of curiosity: In the use case of PhysicalEndpoint, which CNI is typically utilized?

We used this with sriov-device-network-plugin : https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin
The plugin creates a VF and places it in the network namespace of the sandbox.

amshinde: got delayed a bit, I'll take another look and provide my review by today.

amshinde:
@l8huang I went over this PR, and I am not really clear about the use case here.
Can you provide an explanation of the end to end flow, what CNI is being used.

We currently have `PhysicalEndpoint` which is used in the use-case for `sriov-device-network-plugin` that places the VF in the network namespace of the container. The kata-runtime then scans the network namespace to check if is this is physical device, binds it to vfio instead and does a device passthrough.

Looks like in your case the binding to VFIO is done by CNI plugin itself? (Curious why the above flow would not work for you). My question to you then is how is the device passed to the runtime? You mentioned a KEP in the PR, but as far as I understand, the KEP is meant to pass the information of all devices beforehand to account for things like pcie address space required. 

Can you please clarify the above and include the usage in the PR/commit as well. Eventually it will be useful to have a doc including  a detailed explanation here: https://github.com/kata-containers/kata-containers/tree/main/docs/use-cases

src/runtime/virtcontainers/vfio_endpoint.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// SetPciPath sets the PCI path of the endpoint.
+func (endpoint *VfioEndpoint) SetPciPath(path vcTypes.PciPath) {
+	endpoint.PCIPath = path
+}
+
+// Attach for VFIO endpoint
+func (endpoint *VfioEndpoint) Attach(ctx context.Context, s *Sandbox) error {
+	return fmt.Errorf("attach is unsupported for VFIO endpoint")
```
amshinde, line 69: Why is Attach/Detach not implemented at all?

src/runtime/virtcontainers/sandbox.go
```diff
@@ -2235,6 +2237,29 @@ func (s *Sandbox) AddDevice(ctx context.Context, info config.DeviceInfo) (api.De
 	return add, nil
 }
 
+// GetVfioDeviceGuestPciPath return a device's guest PCI path by its host BDF
+func (s *Sandbox) GetVfioDeviceGuestPciPath(hostBDF string) types.PciPath {
```
amshinde, line 2241: Seems like an odd place to put this function here. Consider adding this to https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/device/drivers/vfio.go

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -1249,6 +1239,67 @@ func (k *kataAgent) rollbackFailingContainerCreation(ctx context.Context, c *Con
...
+		for _, ep := range sandbox.network.Endpoints() {
+			if ep.Type() != VfioEndpointType {
+				endpoints = append(endpoints, ep)
+			}
+		}
+	} else if !sandbox.hotplugNetworkConfigApplied {
+		// Apply VFIO network devices' configuration after they are hot-plugged.
+		for _, ep := range sandbox.network.Endpoints() {
+			if ep.Type() == VfioEndpointType {
+				hostBFD := ep.(*VfioEndpoint).HostBDF
```
amshinde, line 1264: /BFD/BDF/

src/runtime/virtcontainers/vfio_endpoint.go
l8huang: VFIO device is attached/detached by VFIODevice's Attach()/Detach() function.

src/runtime/virtcontainers/sandbox.go
l8huang: [src/runtime/pkg/device/drivers/vfio.go](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/pkg/device/drivers/vfio.go) is for a single VFIODevice. This function needs to iterate all `VFIODev`s **via `Sandbox.devManager`** to find the `VFIODev` which matches `hostBDF`, IMO it's not reasonable to move this function to vfio.go.

src/runtime/virtcontainers/kata_agent.go
l8huang: will do.

l8huang:
> We currently have `PhysicalEndpoint` which is used in the use-case for `sriov-device-network-plugin` that places the VF in the network namespace of the container. The kata-runtime then scans the network namespace to check if is this is physical device, binds it to vfio instead and does a device passthrough.
>
> Looks like in your case the binding to VFIO is done by CNI plugin itself? (Curious why the above flow would not work for you). My question to you then is how is the device passed to the runtime? You mentioned a KEP in the PR, but as far as I understand, the KEP is meant to pass the information of all devices beforehand to account for things like pcie address space required.

Looks like you missed the related discussions. Long story short, In our use case, we uses `sriov-device-network-plugin` to allocate VFs(for regular k8s Pod) and VFIOs(for KubeVirt VM), **the number of them are fixed based on the role of k8s node, once they are allocated we don't unbind/bind them in runtime any more**. Our CNI is ovn-k8s; for regular k8s Pod one or more VFs will be allocated and ovn-k8s configures the networks on them; for KubeVirt VM, VFIOs will be allocated, ovn-k8s only adds their representers to OVS bridge and provides network configuration to VM by annotation `k8s.v1.cni.cncf.io/network-status`(ref: [Kubernetes Network Custom Resource Definition De-facto Standard](https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit)) -- **it's not necessary to configure a VF interface with network/route/etc. and rely on VM runtime to scan the config, unbind/rebind the device**.

With Kata, DAN is a neat solution for VFIO network interface, ovn-k8s configures VFIO representers as it currently does, then generates a DAN configuration to apply network setting on the network interface claimed by guest kernel. On the host node, no bind/unbind operations are required, making it a more straightforward approach to use VFIO network device.

The KEP is for moving the `resources` from container-level to Pod lever, so the VFIO device allocation can happens before creating guest VM, so they can be cold-plugged. The VFIO network device's PCI topology is critical for GPUDirect RDMA, per my understanding, GPU/DPU will use CDI to achieve the desired topology. CC @zvonkok

l8huang:
@amshinde thanks for your comments, PR updated, PTAL

I will add a doc to `docs/use-cases` later.

l8huang:
P.S. our sriov-network-device plugin uses config like below to allocate VF and VFIO devices on network adapter:

```
 {
    "skipTopology": true,
    "resourceList": [
        {
            "resourceName": "vf",
            "resourcePrefix": "nvidia.com",
            "selectors": {
                "vendors": [ "15b3" ],
                "devices": [ "101e" ],
                "pfNames": [ "enp1s0f0#1-8" ] # 8 VFs
            }
        },
        {
            "resourceName": "vfio",
            "resourcePrefix": "nvidia.com",
            "selectors": {
                "vendors": [ "15b3" ],
                "devices": [ "101e" ],
                "drivers": [ "vfio-pci" ],
                "pfNames": [ "enp1s0f0#9-16" ] # 8 VFIOS
            }
        }
    ]
}
```
And ovn-k8s uses annotation `k8s.v1.cni.cncf.io/resourceName` to specify should VF or VFIO to be allocated for a network: 
```
- apiVersion: k8s.cni.cncf.io/v1
  kind: NetworkAttachmentDefinition
  metadata:
    annotations:
      k8s.v1.cni.cncf.io/resourceName: nvidia.com/vf # allocate a VF device, ovn-k8s config its network interface
    labels:
    name: ovn-primary
...
- apiVersion: k8s.cni.cncf.io/v1
  kind: NetworkAttachmentDefinition
  metadata:
    annotations:
      k8s.v1.cni.cncf.io/resourceName: nvidia.com/vfio # allocate a VFIo device, ovn-k8s uses DAN to config network interface inside guest VM
    labels:
    name: ovn-secondary-vfio
```

@amshinde please let me know if you have further questions.

l8huang:
@Apokleos @lifupan could you please help review this PR?

l8huang:
This PR has been awaiting a second approval for a month. Please kindly let me know how to move it forward?

Apokleos (commented):
Thx @l8huang for your work on DAN for runtime-go. I've gone over the implementation and left two comments on the parts that confused me. For the most part, looks good to me.

src/runtime/virtcontainers/network_linux.go
```diff
@@ -452,20 +456,21 @@ func (n *LinuxNetwork) addDanEndpoints() error {
...
 
-		_, err := convertDanDeviceToNetworkInfo(&device)
+		netInfo, err := convertDanDeviceToNetworkInfo(&device)
 		if err != nil {
 			return err
 		}
 
-		// TODO: Add endpoints that are supported via DAN
 		switch device.Device.Type {
+		case vctypes.VfioDanDeviceType:
```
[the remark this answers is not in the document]
l8huang: `DanDevice` defines the generic configuration of a network interface(ip/route/neigh), while `VfioDanDevice` is a specific type of device that is configured using DAN.

runtime-rs supports other types of device:

```
pub(crate) enum Device {
    #[serde(rename = "vhost-user")]
    VhostUser {
        // Vhost-user socket path
        path: String,
        #[serde(default)]
        queue_num: usize,
        #[serde(default)]
        queue_size: usize,
    },
    #[serde(rename = "host-tap")]
    HostTap {
        tap_name: String,
        #[serde(default)]
        queue_num: usize,
        #[serde(default)]
        queue_size: usize,
    },
}
```

see  https://github.com/kata-containers/kata-containers/blob/main/src/runtime-rs/crates/resource/src/network/dan.rs#L237-L255

src/runtime/virtcontainers/sandbox.go
```diff
@@ -248,6 +248,8 @@ type Sandbox struct {
 	seccompSupported  bool
 	disableVMShutdown bool
 	isVCPUsPinningOn  bool
+
+	hotplugNetworkConfigApplied bool
```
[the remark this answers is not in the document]
l8huang: it's a flag default to false, once hot-plugged network interface(VFIO in this case)'s network config applied successfully, it will be set to true. It's for preventing network config API being called multiple times when Pod has multiple containers.

PTAL comment at https://github.com/kata-containers/kata-containers/pull/9977/files#diff-6fb7f551035031ed29ad100c61f1a7e92b0fa15d2f67e4c7ea7ebf3b715583f4R1249-R1261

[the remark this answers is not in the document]
l8huang: @Apokleos thanks for reviewing this PR

src/runtime/virtcontainers/network_linux.go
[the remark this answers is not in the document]
Apokleos: Thx @l8huang do you have plan to support other types of devices listed above

src/runtime/virtcontainers/sandbox.go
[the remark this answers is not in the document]
Apokleos: Could you please add comment to explain this introduced flag ?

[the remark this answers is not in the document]
Apokleos: This PR almost looks good to me but the two comments !

src/runtime/virtcontainers/network_linux.go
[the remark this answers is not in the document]
l8huang: nope, we don't have use cases for `vhost-user` and `host-tap`.

src/runtime/virtcontainers/sandbox.go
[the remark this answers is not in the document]
l8huang: added docstring for the flag, PTAL.

Apokleos (approved):
thx @l8huang LGTM ！

Issue #9984, opened 2024-07-09
metrics: remove kata environment check which is duplicated in memory density test


dborquez:
`memory_usage.sh` is currently doing a double check of processes which adds a few seconds wasted to the test.
We need to remove the function call: `common_init` from main().

PR #9987, opened 2024-07-09
metrics: Remove duplicate check of processes from memory test.

dborquez:
This PR removes the common_init function call from the memory usage script to eliminate duplicate checking that is also done from the init_env function.

It also eliminates duplicaction of nested conditionals.

Fixes: #9984


GabyCT (approved):

Apokleos (approved):
Thx @dborquez LGTM!

PR #9992, opened 2024-07-10
ci: fix run-nydus tests

sprt:
```
GH-9973 introduced:

 * New function get_kata_memory_and_vcpus() in
   tests/metrics/lib/common.bash.
 * A call to get_kata_memory_and_vcpus() from extract_kata_env(), which
   is defined in tests/common.bash.

Because the nydus test only sources tests/common.bash, it can't find
get_kata_memory_and_vcpus() and errors out.

So we fix this by moving get_kata_memory_and_vcpus() to
tests/common.bash.
```

Related: #9940

### DIFF
--- a/tests/common.bash
+++ b/tests/common.bash
@@ -176,11 +176,6 @@ function extract_kata_env() {
 	RUNTIME_PATH="$(echo "${kata_env}" | jq -r ${runtime_path})"
 	SHARED_FS="$(echo "${kata_env}" | jq -r ${shared_fs})"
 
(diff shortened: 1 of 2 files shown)


dborquez:
@sprt Thank you, a few comments:
Because the `get_kata_memory_and_vcpus` is just of interest of metrics, you can remove the call to this function from `test/common.sh` and call it but from `metrics/lib/json.bash` inside `metrics_json_init()`. 
Thanks

stevenhorsman (approved):
LGTM. Thanks!

GabyCT (approved):

PR #9996, opened 2024-07-10
CI: Do not use cached artifacts if USE_CACHE=no

koct9i:
- CI: Do not pull builder docker images if USE_CACHE=no
  "make USE_CACHE=no kata-tarball" should build everything from scratch.
  
- CI: Build virtiofsd from source if USE_CACHE=no
  "make virtiofsd-tarball USE_CACHE=no" shouldn't use cached binaries.
  
- CI: Build cloud-hypervisor for source if USE_CACHE=no
  "make cloud-hypervisor-tarball USE_CACHE=no" shouldn't use cached binaries.
  
### DIFF
--- a/tools/packaging/scripts/lib.sh
+++ b/tools/packaging/scripts/lib.sh
@@ -115,6 +115,15 @@ get_last_modification() {
 	popd &> /dev/null
 }
 
+# $1 - The tag to be pulled from the registry
+pull_from_registry() {
+	local tag="${1}"
+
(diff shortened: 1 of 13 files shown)


fidencio:
I think you're mixing 2 things here.
USE_CACHE is related to what we push to oras.  Building artefacts from source is something we do when there are no binary officially provided by the projects themselves.

I'd suggest to add a flag to force build from source both virtiofsd and cloud-hypervisor, but please, do not mix that with USE_CACHE.

Issue #8142, opened 2023-10-06
CC: Trusted storage on guest


stevenhorsman:
When pulling and unpacking, or just decrypting an encrypted container image inside the guest we use tmpfs (in /run/kata-containers/) by default, so that when running in a TEE this data is encrypted and therefore not accessible to the untrusted host. There are limitations with this - encrypted memory is typically small and valuable, so with this feature we can use a side car to instead create a block volume on the host that we can use dm-crypt and dm-verity to protect as a place to store downloaded container images, or data storage. See https://github.com/confidential-containers/confidential-containers/issues/123 for more information

## Things to note/include with this work
I'm not an expert here, but I think the work provided is mostly cherry-picks (checking for updates) of the following commits:
- Add cryptsetup support in Guest kernel and rootfs https://github.com/kata-containers/kata-containers/pull/4762/commits/edf3cba463faebbec86aa87ac4ffd4c30558d665
- Add data-integrity agent config option: https://github.com/kata-containers/kata-containers/pull/4885/commits/b4c4ec66d067edcc42c5560a85728dd5d39ebeff
- Add trusted storage script in the rootfs and enable it in the agent, if configured https://github.com/kata-containers/kata-containers/pull/4885/commits/7e369534c8781b114c5d5e7a8e839f60f6d02c70
- Optimise trusted storage: https://github.com/kata-containers/kata-containers/commit/4906228701b0a71ef8d2e941ce7cca2b8748ad40
- I think we also might have a dependency on https://github.com/kata-containers/kata-containers/issues/8104 to ensure the security & optimisation of the pause image.
- Do we bring across (and update) some of the documentation for this in https://github.com/confidential-containers/confidential-containers/blob/main/guides/ephemeral_storage.md
- One of the things missing in `CCv0` is some test cases for this, so I think we need to add them as part of this issue

## Questions
- How do we decided whether to include the init trusted storage script in rootfs? Is there any harm in always including it as it's only used if `agent.data-integrity` is set in the kernel parameters?
- What are some good test cases for this - Is there anyway to connect to the guest without going through debug console to see that a pod is using the encrypted block storage?

## Acceptance Criteria

For each supported platform:

_**Scenario: Image pull on guest works on Kubernetes**_
[...]

PR #9999, opened 2024-07-11
Trusted image storage

ChengyuZhu6:
Mostly cherry-pick https://github.com/kata-containers/kata-containers/pull/4885 from CCv0

Fixes: #8142

### DIFF
--- a/docs/how-to/how-to-pull-images-in-guest-with-kata.md
+++ b/docs/how-to/how-to-pull-images-in-guest-with-kata.md
@@ -137,7 +137,7 @@ snapshotter = "nydus"
 $ sudo systemctl restart containerd
 ```
 
-## Verification
+## Run pod in kata containers with pulling image in guest
 
 To verify pulling images in a guest VM, please refer to the following commands:
 
@@ -148,8 +148,6 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: busybox
-  annotations:
-    io.containerd.cri.runtime-handler: kata-qemu
 spec:
   runtimeClassName: kata-qemu
   containers:
@@ -163,9 +161,6 @@ NAME         READY   STATUS    RESTARTS   AGE
 busybox      1/1     Running   0          10s
 ```
 
-> **Notes:** 
-> The `CRI Runtime Specific Snapshotter` is still an experimental feature. To pull images in the guest under the specific kata runtime (such as `kata-qemu`), we need to add the following annotation in metadata to each pod yaml: `io.containerd.cri.runtime-handler: kata-qemu`. By adding the annotation, we can ensure that the feature works as expected.
-
 2. Verify that the pod's images have been successfully downloaded in the guest.
 If images intended for deployment are deleted prior to deploying with `nydus snapshotter`, the root filesystems required for the pod's images (including the pause image and the container image) should not be present on the host.
 ```bash
@@ -173,4 +168,145 @@ $ sandbox_id=$(ps -ef| grep containerd-shim-kata-v2| grep -oP '(?<=-id\s)[a-f0-9
 $ rootfs_count=$(find /run/kata-containers/shared/sandboxes/$sandbox_id -name rootfs -type d| grep -o "rootfs" | wc -l)
 $ echo $rootfs_count
 0
+```
+
+## Run pod in kata containers with pulling large image in guest
+
+Currently, the image pulled in the guest will be downloaded and unpacked in the `/run/kata-containers/image` directory. However, by default, in rootfs-confidential image, systemd allocates 50% of the available physical RAM to the `/run` directory using a `tmpfs` filesystem. As we all know, memory is valuable, especially for confidential containers. This means that if we run a kata container with the default configuration (where the default memory assigned for a VM is 2048 MiB), `/run` would be allocated around 1024 MiB. Consequently, we can only pull images up to 1024 MiB in the guest. So we can use a block volume from the host and use `dm-crypt` and `dm-integrity` to encrypt the block volume in the guest, providing a secure place to store downloaded container images.
+
+### Create block volume with k8s
+
+There are a lot of CSI Plugins that support block volumes: AWS EBS, Azure Disk, Open-Local and so on. But as an example, we use Local Persistent Volumes to use local disks as block storage with k8s cluster.
+
+1. Create an empty disk image and attach the image to a loop device, such as `/dev/loop0`
+```bash
+$ loop_file="/tmp/trusted-image-storage.img"
+$ sudo dd if=/dev/zero of=$loop_file bs=1M count=2500
+$ sudo losetup /dev/loop0 $loop_file
+```
+
+2. Create a Storage Class
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: local-storage
+provisioner: kubernetes.io/no-provisioner
+volumeBindingMode: WaitForFirstConsumer
+```
+
+3. Create Persistent Volume
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: trusted-block-pv
+spec:
+  capacity:
+    storage: 10Gi
+  volumeMode: Block
+  accessModes:
+    - ReadWriteOnce
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: local-storage
+  local:
+    path: /dev/loop0
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - NODE_NAME
+```
+
+4. Create Persistent Volume Claim
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: trusted-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+  volumeMode: Block
+  storageClassName: local-storage
+```
+
+5. Run a pod with pulling large image in guest
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: large-image-pod
+spec:
+  runtimeClassName: kata-qemu
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: kubernetes.io/hostname
+            operator: In
+            values:
+            - NODE_NAME
+  volumes:
+    - name: trusted-storage
+      persistentVolumeClaim:
+        claimName: trusted-pvc
+  containers:
+    - name: app-container
+      image: quay.io/confidential-containers/test-images:largeimage
+      command: ["/bin/sh", "-c"]
+      args:
+        - sleep 6000
+      volumeDevices:
+        - devicePath: /dev/trusted_store
+          name: trusted-image-storage
+```
+
+5. Docker image size
+```bash
+docker image ls|grep "largeimage"
+quay.io/confidential-containers/test-images               largeimage            00bc1f6c893a   4 months ago   2.15GB
+```
+
+6. Check whether the device is encrypted and used by entering into the VM
+```bash
+$ lsblk --fs
+NAME                 FSTYPE LABEL UUID FSAVAIL FSUSE% MOUNTPOINT
+sda                                                   
+└─encrypted_disk_GsLDt
+                                          178M    87% /run/kata-containers/image
+
+$ cryptsetup status encrypted_disk_GsLDt
+/dev/mapper/encrypted_disk_GsLDt is active and is in use.
+  type:    LUKS2
+  cipher:  aes-xts-plain64
+  keysize: 512 bits
+  key location: keyring
+  device:  /dev/sda
+  sector size:  4096
+  offset:  32768 sectors
+  size:    5087232 sectors
+  mode:    read/write
+
+$ mount|grep "encrypted_disk_GsLDt"
+/dev/mapper/encrypted_disk_GsLDt on /run/kata-containers/image type ext4
+
+$ du -h --max-depth=1 /run/kata-containers/image/
+16K     /run/kata-containers/image/lost+found
+2.1G    /run/kata-containers/image/layers
+60K     /run/kata-containers/image/overlay
+2.1G    /run/kata-containers/image/
+
+$ free -m
+              total        used        free      shared  buff/cache   available
+Mem:           1989          52          43           0        1893        1904
+Swap:             0           0           0
 ```
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -31,6 +31,7 @@ const GUEST_COMPONENTS_REST_API_OPTION: &str = "agent.guest_components_rest_api"
 const GUEST_COMPONENTS_PROCS_OPTION: &str = "agent.guest_components_procs";
 #[cfg(feature = "guest-pull")]
 const IMAGE_REGISTRY_AUTH_OPTION: &str = "agent.image_registry_auth";
+const SECURE_STORAGE_INTEGRITY_OPTION: &str = "agent.secure_storage_integrity";
 
 // Configure the proxy settings for HTTPS requests in the guest,
 // to solve the problem of not being able to access the specified image in some cases.
@@ -110,6 +111,7 @@ pub struct AgentConfig {
     pub guest_components_procs: GuestComponentsProcs,
     #[cfg(feature = "guest-pull")]
     pub image_registry_auth: String,
+    pub secure_storage_integrity: bool,
 }
 
 #[derive(Debug, Deserialize)]
@@ -131,6 +133,7 @@ pub struct AgentConfigBuilder {
     pub guest_components_procs: Option<GuestComponentsProcs>,
     #[cfg(feature = "guest-pull")]
     pub image_registry_auth: Option<String>,
+    pub secure_storage_integrity: Option<bool>,
 }
 
 macro_rules! config_override {
@@ -198,6 +201,7 @@ impl Default for AgentConfig {
             guest_components_procs: GuestComponentsProcs::default(),
             #[cfg(feature = "guest-pull")]
             image_registry_auth: String::from(""),
+            secure_storage_integrity: false,
         }
     }
 }
@@ -237,7 +241,7 @@ impl FromStr for AgentConfig {
         config_override!(agent_config_builder, agent_config, guest_components_procs);
         #[cfg(feature = "guest-pull")]
         config_override!(agent_config_builder, agent_config, image_registry_auth);
-
+        config_override!(agent_config_builder, agent_config, secure_storage_integrity);
         Ok(agent_config)
     }
 }
@@ -358,6 +362,12 @@ impl AgentConfig {
                 IMAGE_REGISTRY_AUTH_OPTION,
                 config.image_registry_auth,
                 get_string_value
+            );
+            parse_cmdline_param!(
+                param,
+                SECURE_STORAGE_INTEGRITY_OPTION,
+                config.secure_storage_integrity,
+                get_bool_value
             );
         }
 
@@ -586,6 +596,7 @@ mod tests {
             guest_components_procs: GuestComponentsProcs,
             #[cfg(feature = "guest-pull")]
             image_registry_auth: &'a str,
+            secure_storage_integrity: bool,
         }
 
         impl Default for TestData<'_> {
@@ -607,6 +618,7 @@ mod tests {
                     guest_components_procs: GuestComponentsProcs::default(),
                     #[cfg(feature = "guest-pull")]
                     image_registry_auth: "",
+                    secure_storage_integrity: false,
                 }
             }
         }
@@ -1050,6 +1062,31 @@ mod tests {
                 image_registry_auth: "kbs:///default/credentials/test",
                 ..Default::default()
             },
+            TestData {
+                contents: "",
+                secure_storage_integrity: false,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.secure_storage_integrity=true",
+                secure_storage_integrity: true,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.secure_storage_integrity=false",
+                secure_storage_integrity: false,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.secure_storage_integrity=1",
+                secure_storage_integrity: true,
+                ..Default::default()
+            },
+            TestData {
+                contents: "agent.secure_storage_integrity=0",
+                secure_storage_integrity: false,
+                ..Default::default()
+            },
         ];
 
         let dir = tempdir().expect("failed to create tmpdir");
@@ -1111,6 +1148,11 @@ mod tests {
             );
             #[cfg(feature = "guest-pull")]
             assert_eq!(d.image_registry_auth, config.image_registry_auth, "{}", msg);
+            assert_eq!(
+                d.secure_storage_integrity, config.secure_storage_integrity,
+                "{}",
+                msg
+            );
 
             for v in vars_to_unset {
                 env::remove_var(v);
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -59,6 +59,7 @@ use crate::device::{
     add_devices, get_virtio_blk_pci_device_name, update_env_pci, wait_for_net_interface,
 };
 use crate::features::get_build_features;
+use crate::image::KATA_IMAGE_WORK_DIR;
 use crate::linux_abi::*;
 use crate::metrics::get_metrics;
 use crate::mount::baremount;
@@ -106,9 +107,7 @@ use kata_types::k8s;
 
 pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
-const INIT_TRUSTED_STORAGE: &str = "/usr/bin/kata-init-trusted-storage";
 const TRUSTED_IMAGE_STORAGE_DEVICE: &str = "/dev/trusted_store";
-
 /// the iptables seriers binaries could appear either in /sbin
 /// or /usr/sbin, we need to check both of them
 const USR_IPTABLES_SAVE: &str = "/usr/sbin/iptables-save";
@@ -262,10 +261,16 @@ impl AgentService {
                         secure_storage_integrity
                     );
 
-                    Command::new(INIT_TRUSTED_STORAGE)
-                        .args([dev_major_minor.as_str(), &secure_storage_integrity])
-                        .output()
-                        .expect("Failed to initialize trusted storage");
+                    if let Some(cdh) = self.cdh_client.as_ref() {
+                        let options = std::collections::HashMap::from([
+                            ("deviceId".to_string(), dev_major_minor),
+                            ("encryptType".to_string(), "LUKS".to_string()),
+                            ("dataIntegrity".to_string(), secure_storage_integrity),
+                        ]);
+                        cdh.secure_mount("BlockDevice", &options, vec![], KATA_IMAGE_WORK_DIR)
+                            .await?;
+                        break;
+                    }
                 }
             }
         }
--- a/src/agent/src/rpc.rs
+++ b/src/agent/src/rpc.rs
@@ -106,6 +106,8 @@ use kata_types::k8s;
 
 pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
+const INIT_TRUSTED_STORAGE: &str = "/usr/bin/kata-init-trusted-storage";
+const TRUSTED_IMAGE_STORAGE_DEVICE: &str = "/dev/trusted_store";
 
 /// the iptables seriers binaries could appear either in /sbin
 /// or /usr/sbin, we need to check both of them
@@ -239,6 +241,31 @@ impl AgentService {
                             warn!(sl(), "Failed to unseal secret: {}", e)
                         }
                     }
+                }
+            }
+        }
+
+        let linux = oci
+            .linux()
+            .as_ref()
+            .ok_or_else(|| anyhow!("Spec didn't contain linux field"))?;
+        if let Some(devices) = linux.devices() {
+            for specdev in devices.iter() {
+                if specdev.path().as_path().to_str() == Some(TRUSTED_IMAGE_STORAGE_DEVICE) {
+                    let dev_major_minor = format!("{}:{}", specdev.major(), specdev.minor());
+                    let secure_storage_integrity =
+                        AGENT_CONFIG.secure_storage_integrity.to_string();
+                    info!(
+                        sl(),
+                        "trusted_store device major:min {}, enable data integrity {}",
+                        dev_major_minor,
+                        secure_storage_integrity
+                    );
+
+                    Command::new(INIT_TRUSTED_STORAGE)
+                        .args([dev_major_minor.as_str(), &secure_storage_integrity])
+                        .output()
+                        .expect("Failed to initialize trusted storage");
                 }
             }
         }
--- a/tests/integration/kubernetes/confidential_common.sh
+++ b/tests/integration/kubernetes/confidential_common.sh
@@ -82,3 +82,28 @@ function is_confidential_hardware() {
 
 	return 1
 }
+
+function create_loop_device(){
+	local loop_file="${1:-/tmp/trusted-image-storage.img}"
+	cleanup_loop_device "$loop_file"
+
+	sudo dd if=/dev/zero of=$loop_file bs=1M count=2500
+	sudo losetup -fP $loop_file >/dev/null 2>&1
+	local device=$(sudo losetup -j $loop_file | awk -F'[: ]' '{print $1}')
+	echo $device
+}
+
+function cleanup_loop_device(){
+	local loop_file="${1:-/tmp/trusted-image-storage.img}"
+	# Find all loop devices associated with $loop_file
+	local existed_devices=$(sudo losetup -j $loop_file | awk -F'[: ]' '{print $1}')
+	
+	if [ -n "$existed_devices" ]; then
+		# Iterate over each found loop device and detach it
+		for d in $existed_devices; do
+			sudo losetup -d "$d" >/dev/null 2>&1
+		done
+	fi
+
+	sudo rm -f "$loop_file" >/dev/null 2>&1 || true
+}
--- a/tests/integration/kubernetes/k8s-guest-pull-image.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image.bats
@@ -9,14 +9,6 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
 
 setup() {
-    if [ "${KATA_HYPERVISOR}" = "qemu-tdx" ]; then
-	    skip "${KATA_HYPERVISOR} is already running all the tests with guest-pulling, skip this specific one"
-    fi
-
-    if is_confidential_hardware; then
-        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
-    fi
-
     if ! is_confidential_runtime_class; then
         skip "Test not supported for ${KATA_HYPERVISOR}."
     fi
@@ -24,24 +16,15 @@ setup() {
     [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
 
     setup_common
-    unencrypted_image_1="quay.io/sjenning/nginx:1.15-alpine"
-    unencrypted_image_2="quay.io/prometheus/busybox:latest"
+    unencrypted_image="quay.io/prometheus/busybox:latest"
     large_image="quay.io/confidential-containers/test-images:largeimage"
 }
 
 @test "Test we can pull an unencrypted image outside the guest with runc and then inside the guest successfully" {
-    if is_confidential_hardware; then
-        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
-    fi
-
-    if ! is_confidential_runtime_class; then
-        skip "Test not supported for ${KATA_HYPERVISOR}."
-    fi
-
-    # 1. Create one runc pod with the $unencrypted_image_1 image
+    # 1. Create one runc pod with the $unencrypted_image image
     # We want to have one runc pod, so we pass a fake runtimeclass "runc" and then delete the runtimeClassName,
     # because the runtimeclass is not optional in new_pod_config function.
-    runc_pod_config="$(new_pod_config "$unencrypted_image_1" "runc")"
+    runc_pod_config="$(new_pod_config "$unencrypted_image" "runc")"
     sed -i '/runtimeClassName:/d' $runc_pod_config
     set_node "$runc_pod_config" "$node"
     set_container_command "$runc_pod_config" "0" "sleep" "30"
@@ -56,8 +39,8 @@ setup() {
     echo "Runc pod test-e2e is running"
     kubectl delete -f "$runc_pod_config"
 
-    # 2. Create one kata pod with the $unencrypted_image_1 image and nydus annotation
-    kata_pod_with_nydus_config="$(new_pod_config "$unencrypted_image_1" "kata-${KATA_HYPERVISOR}")"
+    # 2. Create one kata pod with the $unencrypted_image image and nydus annotation
+    kata_pod_with_nydus_config="$(new_pod_config "$unencrypted_image" "kata-${KATA_HYPERVISOR}")"
     set_node "$kata_pod_with_nydus_config" "$node"
     set_container_command "$kata_pod_with_nydus_config" "0" "sleep" "30"
 
@@ -72,13 +55,6 @@ setup() {
 
     add_allow_all_policy_to_yaml "$kata_pod_with_nydus_config"
     k8s_create_pod "$kata_pod_with_nydus_config"
-    echo "Kata pod test-e2e with nydus annotation is running"
-
-    echo "Checking the image was pulled in the guest"
-    sandbox_id=$(get_node_kata_sandbox_id $node)
-    echo "sandbox_id is: $sandbox_id"
-    # With annotation for nydus, only rootfs for pause container can be found on host
-    assert_rootfs_count "$node" "$sandbox_id" "1"
 }
 
 @test "Test we can pull a large image inside the guest" {
@@ -89,7 +65,7 @@ setup() {
     set_node "$kata_pod_with_nydus_config" "$node"
     set_container_command "$kata_pod_with_nydus_config" "0" "sleep" "30"
 
-    # Set annotation to pull large image in guest
+    # Set annotation to pull image in guest
     set_metadata_annotation "$kata_pod_with_nydus_config" \
         "io.containerd.cri.runtime-handler" \
         "kata-${KATA_HYPERVISOR}"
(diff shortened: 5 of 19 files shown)


ChengyuZhu6:
Case 1: Failed to pull the large image in the guest in memory:
```
Events:
  Type     Reason   Age                From     Message
  ----     ------   ----               ----     -------
  Normal   Pulled   101s               kubelet  Successfully pulled image "ghcr.io/confidential-containers/test-container:big-size" in 1.221014717s
  Normal   Pulled   83s                kubelet  Successfully pulled image "ghcr.io/confidential-containers/test-container:big-size" in 1.25888686s
  Normal   Pulled   50s                kubelet  Successfully pulled image "ghcr.io/confidential-containers/test-container:big-size" in 1.221221808s
  Warning  Failed   35s (x3 over 85s)  kubelet  Error: failed to create containerd task: failed to create shim task: failed to handle layer: hasher sha256: channel: send failed SendError { .. }: unknown
  Warning  BackOff  21s (x4 over 65s)  kubelet  Back-off restarting failed container
  Normal   Pulling  7s (x4 over 102s)  kubelet  Pulling image "ghcr.io/confidential-containers/test-container:big-size"
  Normal   Created  6s (x4 over 101s)  kubelet  Created container test-container
  Normal   Pulled   6s                 kubelet  Successfully pulled image "ghcr.io/confidential-containers/test-container:big-size" in 1.252684668s
```

ChengyuZhu6:
Case 2: Successfully pull the large image in the guest in the trusted storage:
```
Events:
  Type    Reason                 Age   From               Message
  ----    ------                 ----  ----               -------
  Normal  Scheduled              55s   default-scheduler  Successfully assigned default/large-image-pod to zcy-ubuntu22
  Normal  SuccessfulMountVolume  56s   kubelet            MapVolume.MapPodDevice succeeded for volume "trusted-block-pv" globalMapPath "/var/lib/kubelet/plugins/kubernetes.io~local-volume/volumeDevices/trusted-block-pv"
  Normal  SuccessfulMountVolume  56s   kubelet            MapVolume.MapPodDevice succeeded for volume "trusted-block-pv" volumeMapPath "/var/lib/kubelet/pods/21574dfa-6f74-4520-bf4a-00138330c80f/volumeDevices/kubernetes.io~local-volume"
  Normal  Pulled                 52s   kubelet            Container image "ghcr.io/confidential-containers/test-container:big-size" already present on machine
  Normal  Created                52s   kubelet            Created container app-container
  Normal  Started                1s    kubelet            Started container app-container
```

ChengyuZhu6:
I think the PR is read for review.

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -77,8 +55,8 @@ setup() {
     echo "Checking the image was pulled in the guest"
     sandbox_id=$(get_node_kata_sandbox_id $node)
     echo "sandbox_id is: $sandbox_id"
-    # With annotation for nydus, only rootfs for pause container can be found on host
-    assert_rootfs_count "$node" "$sandbox_id" "1"
+    # No rootfs can be found on host with guest pull
+    assert_rootfs_count "$node" "$sandbox_id" "0"
```
stevenhorsman, line 60: We could consider removing the `assert_rootfs_count` checks entirely, as we are running with sharedfs=none now?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -9,39 +9,22 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
...
 }
 
 @test "Test we can pull an unencrypted image outside the guest with runc and then inside the guest successfully" {
-    if is_confidential_hardware; then
-        skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
-    fi
-
-    if ! is_confidential_runtime_class; then
-        skip "Test not supported for ${KATA_HYPERVISOR}."
-    fi
```
stevenhorsman, line 39: I think we still need this check so that these test don't run on non coco runtimeclasses e.g. qemu-stratovirt?

tests/integration/kubernetes/runtimeclass_workloads/pod-guest-pull-in-trusted-storage.yaml
```diff
@@ -0,0 +1,33 @@
...
+# Copyright (c) 2024 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+apiVersion: v1
+kind: Pod
+metadata:
+  name: large-image-pod
+spec:
+  runtimeClassName: RUNTIMECLASS
```
stevenhorsman, line 11: This should be `kata` and it gets automatically updated with the correct runtimeclass during testing IIRC

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -59,37 +60,59 @@ setup() {
...
     assert_logs_contain "$node" kata "$node_start_time" \
 		'context deadline exceeded'
 
-    kubectl delete -f $kata_pod_with_nydus_config
+    kubectl delete -f $pod_config
+
+    pod_config="${pod_config_dir}/pod-guest-pull-in-trusted-storage.yaml"
+    storage_config="${pod_config_dir}/confidential/trusted-storage.yaml"
+    local_device=$(create_loop_device "/tmp/trusted-storage.img")
+    sed -i "s/RUNTIMECLASS/kata-${KATA_HYPERVISOR}/g" $pod_config
```
stevenhorsman, line 90: This isn't needed

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -59,37 +60,59 @@ setup() {
...
-    echo "Pod $kata_pod_with_nydus_config file:"
-    cat $kata_pod_with_nydus_config
+    echo "Pod $pod_config file:"
+    cat $pod_config
 
     # The pod should be failed because the default timeout of CreateContainerRequest is 60s
-    assert_pod_fail "$kata_pod_with_nydus_config"
+    assert_pod_fail "$pod_config"
     assert_logs_contain "$node" kata "$node_start_time" \
 		'context deadline exceeded'
```
stevenhorsman, line 83: Don't we want to increase the create_container timeout and check it fails because the disk is too small, rather than that the pull isn't quick enough?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -16,8 +16,9 @@ setup() {
     [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
 
     setup_common
+    get_pod_config_dir
     unencrypted_image="quay.io/prometheus/busybox:latest"
-    large_image="quay.io/confidential-containers/test-images:largeimage"
+    large_image="ghcr.io/confidential-containers/test-container:big-size"
```
stevenhorsman, line 21: Out of interest, is there any particular reason for ghcr over quay.io?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -77,167 +56,66 @@ setup() {
...
-    # The expect count of rootfs in host is "2" but the found count of rootfs in host is "1"
-    # As the the first time we pull the $unencrypted_image_1 image via nydus-snapshotter in the guest
-    # for all subsequent pulls still use nydus-snapshotter in the guest
-    # More details: https://github.com/kata-containers/kata-containers/issues/8337
-    # The test case will be PASS after we use containerd 2.0 with 'image pull per runtime class' feature:
-    # https://github.com/containerd/containerd/issues/9377
-    assert_rootfs_count "$node" "$sandbox_id" "2"
-}
-
-@test "Test we can pull an other unencrypted image outside the guest and then inside the guest successfully" {
```
fitzthum, line 177: Maybe we should keep this test?

tests/integration/kubernetes/k8s-guest-pull-image.bats
fitzthum, line 1: I think we might want to rename this file. We use guest pulling for all of the coco tests. These tests are more about guest pulling edge cases. 

Also, any particular reason for dropping the tests that you delete in this PR?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -77,167 +56,66 @@ setup() {
...
-    add_allow_all_policy_to_yaml "$kata_pod_with_nydus_config"
-    k8s_create_pod "$kata_pod_with_nydus_config"
-}
-
-@test "Test we can pull an unencrypted image inside the guest twice in a row and then outside the guest successfully" {
-    # 1. Create one kata pod with the $unencrypted_image_1 image and nydus annotation twice
-    kata_pod_with_nydus_config="$(new_pod_config "$unencrypted_image_1" "kata-${KATA_HYPERVISOR}")"
-    set_node "$kata_pod_with_nydus_config" "$node"
-    set_container_command "$kata_pod_with_nydus_config" "0" "sleep" "30"
+    kubectl delete -f $pod_config
```
fitzthum, line 85: Do you think it would be better to split the positive and negative cases into two tests? We have done that in some of our other files.

src/agent/src/rpc.rs
```diff
@@ -236,6 +238,30 @@ impl AgentService {
...
+            if specdev.path == TRUSTED_STORAGE_DEVICE {
+                let data_integrity = AGENT_CONFIG.data_integrity;
+                info!(
+                    sl(),
+                    "trusted_store device major:min {}, enable data integrity {}",
+                    dev_major_minor,
+                    data_integrity.to_string()
+                );
+
+                Command::new(INIT_TRUSTED_STORAGE)
```
fitzthum, line 258: Do you think it could make sense to handle this functionality in the CDH? That might tie-in nicely with when we support persistent storage that we need a key to unlock.

fitzthum (commented):
Great work. Thanks for doing the heavy-lifting.

I have a couple questions about our plan for various types of secure storage. 

In CCv0 we had ephemeral block device support (basically this PR) and we also had secure storage driven by the CDH. I think there are actually two different parameters. 1) Whether the storage is ephemeral or persistent. 2) Whether the storage is exposed to the pods or if it is used to pull the containers transparently.

This PR is for ephemeral storage that is not exposed to the pod. The question is if we can make these cases share more code and config than they did in CCv0. For instance, how hard would it be to take an ephemeral device used here and expose it to the pod rather than using it to store the containers? What if we wanted to pull containers into persistent encrypted storage? Maybe this is totally out of scope, but I always felt like the two approaches in CCv0 shoudl have been more connected. wdyt?

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: The check will run in both `setup()` and `teardown()` for each test. From my understanding, if the runtime class is not tee’s, should it be skipped directly in the `setup()` phase? Please correct me if I have misunderstood anything.

tests/integration/kubernetes/runtimeclass_workloads/pod-guest-pull-in-trusted-storage.yaml
ChengyuZhu6: Fixed.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: What do you mean?

stevenhorsman: The RUNTIMECLASS replacement in the pod.yaml is done automatically if it's `kata`, so we don't need the sed in the test code

tests/integration/kubernetes/k8s-guest-pull-image.bats
stevenhorsman: Ah sorry, I missed that this was in the test and we also had the setup and teardown. My mistake!

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: The reason for selecting a different image is that Image `quay.io/confidential-containers/test-images:largeimage` is too large (larger than 2GB) and may exceed the default kubelet timeout of 120s in CI. Even if we set the createContainer timeout value to more than 120 seconds on the kata side, it might still fail. Therefore, I chose a smaller image (size = 956MB) that can be pulled within 120 seconds.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Yes, you're right. I’ve added two tests: one to verify trusted storage, and another to test the configuration of the createContainer timeout.

tests/integration/kubernetes/k8s-guest-pull-image.bats
zvonkok: You would need to set the remote request timeout in the kubelet. This is independent of what we set in kata since the context is created in the kubelet.
Do we have the tools in the ci to update kubelet arguments?

zvonkok: The 120s will also break the GPU use case..

ChengyuZhu6: @zvonkok I think updating the kubelet configuration might be challenging while the test is running. And I think we can focus on the feature that allows configuring the timeout specifically for pulling larger images. If users need to pull an image with a timeout exceeding 120 seconds, they can update the configuration of kubelet offline. While I'm not an expert on this, I would like any advice from the experts if possible.

docs/how-to/how-to-pull-images-in-guest-with-kata.md
```diff
@@ -163,14 +161,152 @@ NAME         READY   STATUS    RESTARTS   AGE
...
+        claimName: trusted-pvc
+  containers:
+    - name: app-container
+      image: ghcr.io/confidential-containers/test-container:big-size
+      command: ["/bin/sh", "-c"]
+      args:
+        - sleep 6000
+      volumeDevices:
+        - devicePath: /dev/trusted_store
+          name: trusted-storage
```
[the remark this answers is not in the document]
ChengyuZhu6: @fitzthum The device is exposed to the pod instead container. In this example, only one image is pulled. Therefore, if multiple containers with different images are started in a pod, we only need to define the volume device in one of them. All images would be downloaded in the trusted storage.

docs/how-to/how-to-pull-images-in-guest-with-kata.md
```diff
@@ -163,14 +161,152 @@ NAME         READY   STATUS    RESTARTS   AGE
...
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: local-storage
+provisioner: kubernetes.io/no-provisioner
+volumeBindingMode: WaitForFirstConsumer
+```
+
+3. Create PersistentVolume
```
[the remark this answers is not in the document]
ChengyuZhu6: >Also, do you think it's weird to be using persistent volumes for something that is essentially single-use? Are users expected to create a new volume for each pod or can it be reused? Maybe we should have some info about that.

Yeah, currently, we create a pv for tests. Previous works on the CCv0 branch involved creating PVs using open-local, which were subsequently managed by the CSI driver for cleanup. Without a third-party CSI driver, k8s cannot dynamically clean up the created PVs, so we need manual cleanup. To minimize additional components and reduce complexity in CI, I recommend avoiding the installation of extra CSI drivers.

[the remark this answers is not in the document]
ChengyuZhu6: >k8s does have ephemeral volumes as well. Do you think those would work here?

From my understanding, k8s ephemeral volumes can be effectively used for mounting as file systems or directories without requiring any third-party CSI drivers. However, if we need to use them as block volumes, we still require a third-party CSI driver that supports block devices. Please correct me if I have misunderstood anything.

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Intel Corporation
```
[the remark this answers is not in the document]
ChengyuZhu6: Fixed.

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -246,4 +124,6 @@ teardown() {
 
     kubectl describe pod "$pod_name"
     k8s_delete_all_pods_if_any_exists || true
+    kubectl delete -f "${pod_config_dir}/pod-guest-pull-in-trusted-storage.yaml" || true
+    kubectl delete -f "${pod_config_dir}/confidential/trusted-storage.yaml" || true
```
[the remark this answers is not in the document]
ChengyuZhu6: Yes

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Sure.

src/agent/src/rpc.rs
ChengyuZhu6: That's a good idea. Perhaps it would be better to merge this into CDH when we move forward the secure mount feature?

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Done.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Currently, we do not need annotation "io.containerd.cri.runtime-handler" to enable guest pull. In other words, we will always pull images in the guest using the runtimeclass which sets snapshotter to nydus. So these removed tests are useless.

stevenhorsman: I think we might need `"io.containerd.cri.runtime-handler"` on some of our tests. IIUC this is needed on Kuberenetes 1.28 and earlier and I think some of our self-hosted runner might have an older level of Kubernetes installed as I tried removing this in the past and guest pull wasn't working. Something to check for in the e2e results.

tests/integration/kubernetes/runtimeclass_workloads/pod-guest-pull-in-trusted-storage.yaml
stevenhorsman: It still says `runtimeClassName: $RUNTIMECLASS`?

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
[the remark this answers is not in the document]
stevenhorsman: This looks to be copy-pasted directly from the `CCv0` version, so the original copyright date would be correct

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Done.

tests/integration/kubernetes/runtimeclass_workloads/pod-guest-pull-in-trusted-storage.yaml
ChengyuZhu6: We need to replace it with the value of kata-${KATA_HYPERVISOR} using envsubst, similar to how the new_pod_config function does.

stevenhorsman: So if you look inside the other `runtimeclass_workloads` test pod specs you will see we have `runtimeClassName : kata` set. I think it's because in the CI we alias the `kata` runtimeClass to match the platform we are testing.

docs/how-to/how-to-pull-images-in-guest-with-kata.md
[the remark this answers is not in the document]
fitzthum: Yes, it is at the scope of the pod, but I am saying that in normal kubernetes the behavior of this config would be different. Normal k8s would treat this volumeDevice just like any other volumeDevice, but for us the name field triggers different behavior. This is what we should make sure to document.

src/agent/src/rpc.rs
fitzthum: Yes, although if we're going to move it to the CDH, maybe we should do it that way from the start. I'll leave that for a kata reviewer to think about.

tests/integration/kubernetes/runtimeclass_workloads/pod-guest-pull-in-trusted-storage.yaml
ChengyuZhu6: Fixed.

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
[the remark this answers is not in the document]
ChengyuZhu6: Reverted.

src/agent/src/rpc.rs
ChengyuZhu6: Yeah, I have synced with @Xynnn007. He agreed to move the logic to CDH. I’ll handle it in this PR.

Xynnn007: Yes, we can reuse the script for multiple scenarios a) encrypted storage for extra image storage b) encrypted rbd for container volume. The both could get key a) from KBS or b) generate an ephemeral key.

fitzthum: Ok, I think that's a good idea. It might take a little more work up front, but I think it will make things nicer down the road.

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -77,167 +56,66 @@ setup() {
...
-
-    # The assert_rootfs_count will be FAIL.
-    # The expect count of rootfs in host is "1" but the found count of rootfs in host is "2"
-    # As the the first time we pull the $unencrypted_image_2 image via overlayfs-snapshotter in host
-    # for all subsequent pulls still use overlayfs-snapshotter in host.
-    # More details: https://github.com/kata-containers/kata-containers/issues/8337
-    # The test case will be PASS after we use containerd 2.0 with 'image pull per runtime class' feature:
-    # https://github.com/containerd/containerd/issues/9377
-    assert_rootfs_count "$node" "$sandbox_id" "1"
+    add_allow_all_policy_to_yaml "$pod_config"
```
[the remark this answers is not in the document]
c3d: @fitzthum The status of OPA policies a good question, but I think that you'd give it more visibility outside of this specific PR (e.g. on Slack).

On this PR specifically, the question of what is the correct way to do it in a test is a really good one. At the very least, a comment would be welcome.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: I’m a bit confused. Currently, we have set `shared_fs` to none, which means we can’t share any files or directories with guest through virtio-fs/9p, including the image pulled on the host. As a result, these tests can’t run on CI. Why do we need to keep them?

docs/how-to/how-to-pull-images-in-guest-with-kata.md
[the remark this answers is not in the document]
ChengyuZhu6: Update info: I attempted to use one of k8s’ ephemeral storage types, `emptyDir`, to achieve trusted storage. However, it did not work for the reason:
The `emptyDir` type can only be used as a tmpfs mounted to the guest, and its size is influenced by the RAM allocated to the guest.

src/agent/src/rpc.rs
ChengyuZhu6: https://github.com/confidential-containers/guest-components/issues/540

ChengyuZhu6:
After discussion, we decided to have CDH handle the logic for encrypting the block device. We need to wait for the PR https://github.com/confidential-containers/guest-components/pull/617 to merge.

ChengyuZhu6:
Mark the PR as draft.

src/agent/src/config.rs
```diff
@@ -29,6 +29,7 @@ const UNIFIED_CGROUP_HIERARCHY_OPTION: &str = "systemd.unified_cgroup_hierarchy"
 const CONFIG_FILE: &str = "agent.config_file";
 const GUEST_COMPONENTS_REST_API_OPTION: &str = "agent.guest_components_rest_api";
 const GUEST_COMPONENTS_PROCS_OPTION: &str = "agent.guest_components_procs";
+const DATA_INTEGRITY_OPTION: &str = "agent.data_integrity";
```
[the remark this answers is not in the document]
ChengyuZhu6: @c3d This information is based on commit https://github.com/kata-containers/kata-containers/pull/4885/commits/b4c4ec66d067edcc42c5560a85728dd5d39ebeff. Once commit https://github.com/arronwy/kata-containers/commit/4906228701b0a71ef8d2e941ce7cca2b8748ad40 is added to optimize the integrity device creation, the actual overhead may be reduced.

ChengyuZhu6:
Currently, I attempted to update `coco-guest-components` to the commit https://github.com/confidential-containers/guest-components/commit/0d8146321c3de023f0f7f40e47fc0f860133dfc7 that supports trusted storage. However, all related coco CIs failed due to attestation agent initialization error. I have created an issue in confidential-containers to track this: https://github.com/confidential-containers/guest-components/issues/626. @Xynnn007 has quickly addressed this with a PR: https://github.com/confidential-containers/guest-components/pull/627. Thanks for the help, @Xynnn007!

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
```diff
@@ -0,0 +1,149 @@
...
+device_path="/dev/$device_name"
+
+opened_device_name=$(mktemp -u "encrypted_disk_XXXXX")
+
+if [[ -n "$device_name" && -b "$device_path" ]]; then
+
+	if [ "$is_encrypted" == "false" ]; then
+
+		if [ "$data_integrity" == "false" ]; then
+			echo "YES" | cryptsetup luksFormat --type luks2 "$device_path" --sector-size 4096 \
```
sprt, line 81: I believe we can use the `-q/--batch-mode` flag here instead of passing `YES`.

ChengyuZhu6: Nice catch. Thanks!

ChengyuZhu6: Fixed.

src/agent/src/config.rs
```diff
@@ -110,6 +111,7 @@ pub struct AgentConfig {
     pub guest_components_procs: GuestComponentsProcs,
     #[cfg(feature = "guest-pull")]
     pub image_registry_auth: String,
+    pub data_integrity: bool,
```
fitzthum, line 114: `data_integrity` is a pretty generic term. Maybe this should be something more specific to secure storage?

fitzthum (commented):
Looks great. Two small notes.

docs/how-to/how-to-pull-images-in-guest-with-kata.md
```diff
@@ -163,14 +161,152 @@ NAME         READY   STATUS    RESTARTS   AGE
...
 ```bash
 $ sandbox_id=$(ps -ef| grep containerd-shim-kata-v2| grep -oP '(?<=-id\s)[a-f0-9]+'| tail -1)
 $ rootfs_count=$(find /run/kata-containers/shared/sandboxes/$sandbox_id -name rootfs -type d| grep -o "rootfs" | wc -l)
 $ echo $rootfs_count
 0
+```
+
+## Run pod in kata containers with pulling large image in guest
+
+Currently, the image pulled in the guest will be downloaded and unpacked in the `/run/kata-containers/image` directory. However, by default, systemd allocates 10% of the available physical RAM to the `/run` directory using a `tmpfs` filesystem. As we all know, memory is valuable, especially for confidential containers. This means that if we run a kata container with the default configuration (where the default memory assigned for a VM is 2048 MiB), `/run` would be allocated around 200 MiB. Consequently, we can only pull images up to 200 MiB in the guest. So we can use a block volume from the host and use `dm-crypt` and `dm-integrity` to encrypt the block volume in the guest, providing a secure place to store downloaded container images.
```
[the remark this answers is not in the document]
ChengyuZhu6: Ah, good point. I’ve updated the description.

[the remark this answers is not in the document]
ChengyuZhu6: And I've updated the example.

src/agent/src/config.rs
ChengyuZhu6: I have renamed `data_integrity` to `secure_storage_integrity`.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: @stevenhorsman I have kept `io.containerd.cri.runtime-handler` in the tests.

ChengyuZhu6:
rebase to main to fix the issue with the deployment of kbs.

ChengyuZhu6:
@arronwy  @fitzthum @zvonkok @Xynnn007 @stevenhorsman @sprt @c3d The tests about trusted storage have passed on some TEE and non-TEE platforms. BTW, this PR is unrelated to TEE. I will track the failed tests, but I think this PR is ready for another review. I have addressed the issues you mentioned earlier and I appreciate it if you could review it again.. Thank you.

fitzthum:
Code looks good to me.

The nontee test is failing https://github.com/kata-containers/kata-containers/actions/runs/10193791423/job/28206001226?pr=9999#step:14:1714

ChengyuZhu6:
> Code looks good to me.
> 
> The nontee test is failing https://github.com/kata-containers/kata-containers/actions/runs/10193791423/job/28206001226?pr=9999#step:14:1714

Yes, it might be due to the lack of available loop devices on the node. I’ll need the CI maintainer’s help with this.

ChengyuZhu6:
The job on s390x failed because the image `quay.io/confidential-containers/test-images:largeimage` timed out while being pulled in the guest. The image size is 2.15GB.

fitzthum:
> Yes, it might be due to the lack of available loop devices on the node. I’ll need the CI maintainer’s help with this.

I think these tests run in the AKS environment, which might make things a little tricky.

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -82,3 +82,28 @@ function is_confidential_hardware() {
...
 	return 1
 }
+
+function create_loop_device(){
+	local loop_file="${1:-/tmp/trusted-storage.img}"
+	cleanup_loop_device
+
+	sudo dd if=/dev/zero of=$loop_file bs=1M count=2500
+	sudo losetup -fP $loop_file >/dev/null 2>&1
+	local device=$(sudo losetup -j $loop_file | awk -F'[: ]' '{print $1}')
```
sprt, line 92: You have to use `exec_host()` here, otherwise the loop device will be created on the runner machine, rather than the host node.

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -750,6 +750,11 @@ EOF
 		tar xvJpf ${COCO_GUEST_COMPONENTS_TARBALL} -C ${ROOTFS_DIR}
 	fi
 
+	if [ "${MEASURED_ROOTFS}" == "yes" ]; then
+		info "Install init_trusted_storage script"
+		install -o root -g root -m 0500 "${script_dir}/scripts/init_trusted_storage.sh" "${ROOTFS_DIR}/usr/local/bin/luks-encrypt-storage"
```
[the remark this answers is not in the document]
ChengyuZhu6: >This is installed for the CDH, correct?

Yes. Currently, CDH is not responsible for installing the script, so we need to install it in kata.

> As in, only the CDH is expected to call this script?
Yes, that's correct.

ChengyuZhu6:
>Agreed, I think we should clearly document this. It's also confusing to call this feature "trusted storage" as the actual scope is more constrained - I'd call it "trusted image storage".

Sure. Let's rename it to `trusted image storage`.

src/agent/src/rpc.rs
```diff
@@ -243,6 +245,37 @@ impl AgentService {
...
+                        secure_storage_integrity
+                    );
+
+                    if let Some(cdh) = self.cdh_client.as_ref() {
+                        let options = std::collections::HashMap::from([
+                            ("deviceId".to_string(), dev_major_minor),
+                            ("encryptType".to_string(), "LUKS".to_string()),
+                            ("dataIntegrity".to_string(), secure_storage_integrity),
+                        ]);
+                        cdh.secure_mount("BlockDevice", &options, vec![], KATA_IMAGE_WORK_DIR)
```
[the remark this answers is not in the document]
ChengyuZhu6: Yes, and for the rootfs-confidential image, the size of `/run` as tmpfs is 50% of the memory.

tests/integration/kubernetes/confidential_common.sh
ChengyuZhu6: Thanks for your comment @sprt. Let's change the code and see if the test passes on coco-dev CI.

ChengyuZhu6:
rebase to main only.

ChengyuZhu6:
Because the issue https://github.com/kata-containers/kata-containers/issues/10133 by using `exec_host` to manage loop device. So the tests with trusted storage will be only running in baremetal machines.

sprt (approved):
LGTM! Just 2 nits. 🙂

src/agent/src/rpc.rs
```diff
@@ -106,6 +107,7 @@ use kata_types::k8s;
 
 pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
+const TRUSTED_STORAGE_DEVICE: &str = "/dev/trusted_store";
```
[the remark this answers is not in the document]
ChengyuZhu6: Fixed.

src/agent/src/config.rs
```diff
@@ -31,6 +31,7 @@ const GUEST_COMPONENTS_REST_API_OPTION: &str = "agent.guest_components_rest_api"
 const GUEST_COMPONENTS_PROCS_OPTION: &str = "agent.guest_components_procs";
 #[cfg(feature = "guest-pull")]
 const IMAGE_REGISTRY_AUTH_OPTION: &str = "agent.image_registry_auth";
+const SECURE_STORAGE_INTEGRITY_OPTION: &str = "agent.secure_storage_integrity";
```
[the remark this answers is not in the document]
ChengyuZhu6: This option is used for secure storage，not only for image storage. I recommend to keep this name.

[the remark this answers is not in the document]
wainersm: Hi @ChengyuZhu6 !

Could you please update https://github.com/kata-containers/kata-containers/blob/main/src/agent/README.md#agent-options ?

src/agent/src/rpc.rs
```diff
@@ -106,6 +106,8 @@ use kata_types::k8s;
 
 pub const CONTAINER_BASE: &str = "/run/kata-containers";
 const MODPROBE_PATH: &str = "/sbin/modprobe";
+const INIT_TRUSTED_STORAGE: &str = "/usr/bin/kata-init-trusted-storage";
+const TRUSTED_IMAGE_STORAGE_DEVICE: &str = "/dev/trusted_store";
```
wainersm, line 110: to keep it coherent, shouldn't it be `TRUSTED_STORAGE_DEVICE` (without IMAGE)?

src/agent/src/rpc.rs
```diff
@@ -243,6 +245,32 @@ impl AgentService {
             }
         }
 
+        let linux = oci
+            .linux()
+            .as_ref()
+            .ok_or_else(|| anyhow!("Spec didn't contain linux field"))?;
+        if let Some(devices) = linux.devices() {
+            for specdev in devices.iter() {
+                let dev_major_minor = format!("{}:{}", specdev.major(), specdev.minor());
```
wainersm, line 253: micro-optimization: move this assignment to the `if specdev.path().....` block below.

src/agent/src/rpc.rs
```diff
@@ -243,6 +245,32 @@ impl AgentService {
...
+                        "trusted_store device major:min {}, enable data integrity {}",
+                        dev_major_minor,
+                        secure_storage_integrity
+                    );
+
+                    Command::new(INIT_TRUSTED_STORAGE)
+                        .args([dev_major_minor.as_str(), &secure_storage_integrity])
+                        .output()
+                        .expect("Failed to initialize confidential storage");
+                }
```
wainersm, line 274: `break` the `for` loop once trusted storage is initialized ;)

src/agent/src/rpc.rs
```diff
@@ -243,6 +245,32 @@ impl AgentService {
...
+                        sl(),
+                        "trusted_store device major:min {}, enable data integrity {}",
+                        dev_major_minor,
+                        secure_storage_integrity
+                    );
+
+                    Command::new(INIT_TRUSTED_STORAGE)
+                        .args([dev_major_minor.as_str(), &secure_storage_integrity])
+                        .output()
+                        .expect("Failed to initialize confidential storage");
```
wainersm: suggestion: s/confidential/trusted/

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
```
wainersm, line 1: Is this script a copy of https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/storage/scripts/luks-encrypt-storage ? if so then I think we should fetch the file from guest-components otherwise it can get easily out of sync and introduce regressions hard to find.

One question that I have is if any plan to replace the script by a rust app or even a function in guest-components.

ChengyuZhu6: @wainersm I think we can enhance this further in our follow-up work. WDYT?

src/agent/src/rpc.rs
ChengyuZhu6: This device is used for image storage, but the script `kata-init-trusted-storage` serves additional scenarios beyond just  trusted image storage.

src/agent/src/rpc.rs
ChengyuZhu6: Fixed.

src/agent/src/rpc.rs
ChengyuZhu6: Done.

src/agent/src/rpc.rs
ChengyuZhu6: Done.

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
ChengyuZhu6: >Is this script a copy of https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/storage/scripts/luks-encrypt-storage ? if so then I think we should fetch the file from guest-components otherwise it can get easily out of sync and introduce regressions hard to find.

Yes, I agree with you. Although the PR includes many commits, I recommend addressing the work in the next PR. WDYT?

src/agent/src/config.rs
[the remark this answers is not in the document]
ChengyuZhu6: Sure!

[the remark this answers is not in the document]
ChengyuZhu6: Done.

ChengyuZhu6:
rebase to main only

src/agent/src/config.rs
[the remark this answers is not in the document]
wainersm: thx!

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
wainersm: Hi @ChengyuZhu6 ! I'm okay to address these comments in a follow on PR

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -72,13 +55,6 @@ setup() {
 
     add_allow_all_policy_to_yaml "$kata_pod_with_nydus_config"
     k8s_create_pod "$kata_pod_with_nydus_config"
-    echo "Kata pod test-e2e with nydus annotation is running"
-
-    echo "Checking the image was pulled in the guest"
```
wainersm, line 77: Hi @ChengyuZhu6 !

I didn't understand why you removed this check. How do we ensure it was really pulled by the guest?

tests/integration/kubernetes/k8s-guest-pull-image.bats
wainersm: @ChengyuZhu6 , mind to explain why this and `"Test we can pull an unencrypted image inside the guest twice in a row and then outside the guest successfully" ` aren't needed anymore?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -57,42 +61,66 @@ setup() {
...
-          The test will be skiped until https://github.com/kata-containers/kata-containers/issues/8142 is addressed."
-    kata_pod_with_nydus_config="$(new_pod_config "$large_image" "kata-${KATA_HYPERVISOR}")"
-    set_node "$kata_pod_with_nydus_config" "$node"
-    set_container_command "$kata_pod_with_nydus_config" "0" "sleep" "30"
+@test "Test we cannot pull an image that exceeds the memory limit inside the guest" {
+    # The image pulled in the guest will be downloaded and unpacked in the `/run/kata-containers/image` directory. 
+    # However, by default, systemd allocates 50% of the available physical RAM to the `/run` directory using a `tmpfs` filesystem. 
+    # It means that if we run a kata container with the default configuration (where the default memory assigned for a VM is 2048 MiB), 
+    # `/run` would be allocated around 1024 MiB. Consequently, we can only pull images up to 1024 MiB in the guest. 
+    # However, the unpacked size of image "rust:latest" is 1.41GB.
```
wainersm, line 69: why don't keep using quay.io/confidential-containers/test-images:largeimage which is larger than 1.41GB, so will fail as well?

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -16,8 +16,12 @@ setup() {
     [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
 
     setup_common
+    get_pod_config_dir
     unencrypted_image="quay.io/prometheus/busybox:latest"
-    large_image="quay.io/confidential-containers/test-images:largeimage"
+    image_pulled_time_less_than_default_time="rust:latest" # unpacked size: 1.41GB
```
wainersm, line 21: `rust:latest` will be pulled from docker.io so changes are CI will hit pull rate limit issue. We better find a large image from quay.io, or maybe this image is not needed (see my next comment)

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -57,42 +61,66 @@ setup() {
...
 
-    kubectl delete -f $kata_pod_with_nydus_config
+@test "Test we can pull an image inside the guest using trusted storage" {
+    # The image pulled in the guest will be downloaded and unpacked in the `/run/kata-containers/image` directory. 
+    # The tests will use `cryptsetup` to encrypt a block device and mount it at `/run/kata-containers/image`.
+    storage_config=$(mktemp "${BATS_FILE_TMPDIR}/$(basename "${storage_config_template}").XXX")
+    local_device=$(create_loop_device)
+    LOCAL_DEVICE="$local_device" NODE_NAME="$node" envsubst < "$storage_config_template" > "$storage_config"
 
-    # Set CreateContainerRequest timeout in the annotation to pull large image in guest
```
wainersm, line 108: why timeout is not needed anymore? Don't we have the risk of hitting it?

wainersm: ok, next commit introduce the tests with timeout.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Hi @wainersm! In the coco CI job, we're setting `shared_fs` to `none`, as seen in the Makefile: https://github.com/kata-containers/kata-containers/blob/91084058ae620d7c3c2053cdac2de0445824fbc2/src/runtime/Makefile#L248-L252. This change implies that all kubernetes tests will now pull images in the guest. Consequently, there's no longer a need to check.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Because the default pull timeout is 60s and it would be timeout if we pull `quay.io/confidential-containers/test-images:largeimage` in the case `Test we can pull an image inside the guest using trusted storage`.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Yes, I agree. But I’ve been searching for an appropriate and dependable image from quay.io that is larger than 1GB yet smaller than 2GB, but I haven’t found one that meets these criteria yet. Do you have any advice?

ChengyuZhu6: @wainersm  I have replaced "rust:latest" with "ghcr.io/confidential-containers/test-container:rust-1.79.0". Thanks @Xynnn007's help!

tests/integration/kubernetes/k8s-guest-pull-image.bats
wainersm: My understanding of `shared_fs=none` is that even not mounting filesystems, it will be copying files over the guest (e.g. config maps). I don't know if it will attempt to copy images...

Anyway, afaik we don't have any test to test `shared_fs=none` really works. Having that checking on this test case will ensure that images are pulling from the guest as intended. Is it too much work keep this checking?

tests/integration/kubernetes/k8s-guest-pull-image.bats
wainersm: ah, ok. so it should large enough to get pulled before the deadline. Thanks for the explanation!

tools/osbuilder/rootfs-builder/scripts/init_trusted_storage.sh
wainersm: @ChengyuZhu6 created an issue to track that improvement: https://github.com/confidential-containers/guest-components/issues/670

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -82,3 +82,28 @@ function is_confidential_hardware() {
 
 	return 1
 }
+
+function create_loop_device(){
+	local loop_file="${1:-/tmp/trusted-image-storage.img}"
+	cleanup_loop_device
```
wainersm, line 88: Should pass `loop_file` to `cleanup_loop_device`:  `cleanup_loop_device "$loop_file"`

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -102,6 +130,10 @@ teardown() {
 
     [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
 
-    kubectl describe pod "$pod_name"
+    kubectl describe pods
     k8s_delete_all_pods_if_any_exists || true
+    kubectl delete pvc trusted-pvc || true
```
wainersm, line 229: Please run these `kubectl delete` with `--ignore-not-found` (and without `|| true`). For example, `kubectl delete --ignore-not-found pvc trusted-pvc`.

So avoiding nasty messages like:
```
2024-08-13T09:03:23.7388751Z # Error from server (NotFound): persistentvolumeclaims "trusted-pvc" not found
2024-08-13T09:03:23.7390598Z # Error from server (NotFound): persistentvolumes "trusted-block-pv" not found
2024-08-13T09:03:23.7392602Z # Error from server (NotFound): storageclasses.storage.k8s.io "local-storage" not found
```

ps: I know that many tests don't pass that parameter, but it's not always too late to learn new habits :)

tests/integration/kubernetes/k8s-guest-pull-image.bats
```diff
@@ -72,178 +59,175 @@ setup() {
...
     # For debug sake
-    echo "Pod $kata_pod_without_nydus_config file:"
-    cat $kata_pod_without_nydus_config
+    echo "Pod $pod_config file:"
+    cat $pod_config
+
+    # The pod should be failed because the default timeout of CreateContainerRequest is 60s
+    assert_pod_fail "$pod_config"
+    assert_logs_contain "$node" kata "$node_start_time" \
+		'context deadline exceeded'
```
wainersm, line 174: Unfortunately this error message is pretty generic and can indicate any error. As we have a test that increases the timeout and it should pass then I think we are fine.

But if we could improve the message (not for this PR)...the problem is that it cannot properly propagate errors within the guest?

wainersm:
Hi @ChengyuZhu6 !

First, thanks for yet another great work! I don't have any further comments to this PR, the improvements can be done on future PRs IMHO. I suppose you will update this PR once more, then we can check if tests still pass.

tests/integration/kubernetes/confidential_common.sh
ChengyuZhu6: Fixed.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Good point. Thanks for your suggestion @wainersm !

ChengyuZhu6: Fixed.

tests/integration/kubernetes/k8s-guest-pull-image.bats
ChengyuZhu6: Yes, I agree. I’ll address it in the upcoming PR and keep you updated.

ChengyuZhu6:
> Hi @ChengyuZhu6 !
> 
> First, thanks for yet another great work! I don't have any further comments to this PR, the improvements can be done on future PRs IMHO. I suppose you will update this PR once more, then we can check if tests still pass.

@wainersm Thanks for your comments!

wainersm:
Hi @ChengyuZhu6 !

The tests passed on SNP and SEV
TDX failed the job but on unrelated tests, k8s-guest-pull-image.bats passed.
zvsi failed even before running the tests but it leverages qemu-coco-dev, so it should skip k8s-guest-pull-image.bats tests.

wainersm (approved):
Hi @ChengyuZhu6 !

We know that there are improvements to be done, we've discussed on this PR and reached on an agreement (at least between you and me :)

This is passing tests on all TEE platforms!

So IMO it's ready to be merged. Thanks again for this amazing work!

fitzthum (approved):
Great work Chengyu. Thanks!

arronwy (approved):
Thanks @ChengyuZhu6 LGTM!

PR #10001, opened 2024-07-11
runtime: provide protection for shared pointer

### LINKED ISSUE #9989: runtime: shim maybe panic when a lot of pods are remove concurrently
# Description of problem

kata-shim maybe panic when a lot of pods are removed concurrently.
About one in every 7,000 pods cannot be deleted.
How to reproduce this:
1. create deployment with 7000 pods
2. delete the pods of  deployment
Repeat the above steps several times

# Expected result

All of the pods can be delete successfully. No kata-shim panic occurs.

# Actual result

# Further information
This issue is similar to https://github.com/kata-containers/kata-containers/issues/3440.
However,  I have cherry-picked the PR:https://github.com/kata-containers/kata-containers/pull/3441 , the kata-shim panic still occurs.
In the go programming language, reading a map variable with a value of nil will not lead to a panic. When you try access a key in a nil map, it will simply return the zero value of the map's value type.
Based on the above, I think the following modification is unnecessary. The type of `k.reqHandlers` is `map[string]reqFunc`, so reading k.reqHandlers with nil will not lead to panic.
```
+       k.Lock()
+       if k.reqHandlers == nil {
+               return nil, errors.New("Client has already disconnect")
+       }
+
        handler := k.reqHandlers[msgName]
        if msgName == "" || handler == nil {
                return nil, errors.New("Invalid request type")
        }
+       k.Unlock()
```
As shown in (#3440)[https://github.com/kata-containers/kata-containers/issues/3440]

gaohuatao-1:
The pointer type variable k.client should be used within the lock scope

Fixes #9989

### DIFF
--- a/src/runtime/virtcontainers/kata_agent.go
+++ b/src/runtime/virtcontainers/kata_agent.go
@@ -715,7 +715,7 @@ func (k *kataAgent) reuseAgent(agent agent) error {
 		return fmt.Errorf("Bug: get a wrong type of agent")
 	}
 
-	k.installReqFunc(a.client)
+	k.installReqFunc()
 	k.client = a.client
 	return nil
 }
@@ -2003,7 +2003,7 @@ func (k *kataAgent) connect(ctx context.Context) error {
 		return err
 	}
 
-	k.installReqFunc(client)
+	k.installReqFunc()
 	k.client = client
 
 	return nil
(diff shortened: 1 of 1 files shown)


gaohuatao-1:
@Apokleos @justxuewei  Could you please give a approve? Thx~

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2241,20 +2241,19 @@ func (k *kataAgent) sendReq(spanCtx context.Context, request interface{}) (inter
...
+	k.Unlock()
 
-	if k.reqHandlers == nil {
-		k.Unlock()
+	if reqHandlers == nil || client == nil {
 		return nil, errors.New("Client has already disconnected")
 	}
 
-	handler := k.reqHandlers[msgName]
+	handler := reqHandlers[msgName]
```
amshinde, line 2252: I think this operation should be done within the lock as well.
Assigning the map to `reqHandlers` with `reqHandlers := k.reqHandlers` is still a shallow copy, pointing to the original map. If you want to protect read/writes to the map, maybe you should retrieve the handler within the lock itself.

src/runtime/virtcontainers/kata_agent.go
```diff
@@ -2090,120 +2090,120 @@ func (k *kataAgent) removeStaleVirtiofsShareMounts(ctx context.Context) error {
 	return err
 }
 
-type reqFunc func(context.Context, interface{}) (interface{}, error)
+type reqFunc func(context.Context, interface{}, *kataclient.AgentClient) (interface{}, error)
 
 func (k *kataAgent) installReqFunc(c *kataclient.AgentClient) {
```
amshinde, line 2095: I was looking at this function `installReqFunc` and looks like it really does not need the parameter passed to it `c *kataclient.AgentClient` at all. While you are at it, maybe you can drop the parameter as well to clean up this function further.

src/runtime/virtcontainers/kata_agent.go
gaohuatao-1: Thx. There is no concurrent reading and writing in the map:reqHandlers, as we can see in func sendReq(). Here, I think it is necessary to protect k.reqHandlers param  itself  only. That is to say, setting k.reqHandlers value to be nil and getting k.reqHandlers param value are concurrent. However, reading/writing to the map is serial, as shown below:
image

src/runtime/virtcontainers/kata_agent.go
gaohuatao-1: @amshinde Good idea! I have fixed it. PTAL.

amshinde (approved):

gaohuatao-1:
@Apokleos  Hi， PTAL.

Apokleos (approved):
Thx @gaohuatao-1 LGTM

gaohuatao-1:
@Apokleos  Could you pls merge it? Thx~

Issue #8879, opened 2024-01-22
genpolicy: add policy support for runAsUser


danmihai1:
genpolicy doesn't generate correct policy for the K8s YAML filed runAsUser field of SecurityContext and PodSecurityContext.

Example from Kata CI: pod-security-context.yaml

PR #10008, opened 2024-07-13
genpolicy: add support for runAsUser fields

danmihai1:
1. Explain better the reason for already-existing workaround in rules.rego.
2. Add genpolicy ability to create policy based on the SecurityContext.runAsUser and PodSecurityContext.runAsUser fields.
3. Add basic tests for genpolicy + these two fields.

Fixes: #8879

### DIFF
--- a/tests/integration/kubernetes/k8s-security-context.bats
+++ b/tests/integration/kubernetes/k8s-security-context.bats
@@ -12,7 +12,14 @@ setup() {
 	get_pod_config_dir
 
 	yaml_file="${pod_config_dir}/pod-security-context.yaml"
-	add_allow_all_policy_to_yaml "${yaml_file}"
+	policy_settings_dir="$(create_tmp_policy_settings_dir "${pod_config_dir}")"
+
+	cmd="ps --user 1000 -f"
+	exec_command="sh -c ${cmd}"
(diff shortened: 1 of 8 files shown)


GabyCT (approved):

tests/integration/kubernetes/k8s-security-context.bats
```diff
@@ -25,7 +32,6 @@ setup() {
 	kubectl wait --for=condition=Ready --timeout=$timeout pod "$pod_name"
 
 	# Check user
-	cmd="ps --user 1000 -f"
 	process="tail -f /dev/null"
 	kubectl exec $pod_name -- sh -c $cmd | grep "$process"
```
wainersm, line 36: Hi @danmihai1 ,

One little adjust would be nice: so here pass `$exec_command`, i.e., `kubectl exec $pod_name -- $exec_command | grep "$process"` to ensure it's will always be exactly the command set in the policy settings.

danmihai1: That's a great point - thanks @wainersm! I missed that before, so my preference would be to clean up all instances of this bug in a separate PR. I will fix this using the reminder from https://github.com/kata-containers/kata-containers/issues/10014.

wainersm (approved):
LGTM. Thanks @danmihai1 !

PR #9904, opened 2024-06-25
Support for registry authentication in guest pull

stevenhorsman:
Add authenticated registry support for guest-pull

Fixes: #8122

PR #9988, opened 2024-07-10
initdata: add initdata annotation in hypervisor config

huoqifeng:
- Add Initdata annotation for hypervisor config, so that it can be passed when CreateVM following [initdata](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md)
    
Signed-off-by: Qi Feng Huo <huoqif@cn.ibm.com>

Issue #9360, opened 2024-03-27
CC: build and publish encrypted image(s) for testing


wainersm:
This issue is a spin off [CC: Pull image and unpack of encrypted image on guest](https://github.com/kata-containers/kata-containers/issues/8111)

In order to test pulling of encrypted image on guest, it obviously should use encrypted images. Quoting @stevenhorsman on issue #8111 :

"""
What container image do we want to use? In CCv0 we use, [docker.io/katadocker/ccv0-ssh](https://hub.docker.com/r/katadocker/ccv0-ssh) built by https://github.com/confidential-containers/confidential-containers/tree/main/demos/ssh-demo, but should we take ownership of the image and build it ourselves like we do with the unencrypted test image: https://github.com/kata-containers/kata-containers/tree/main/tests/integration/kubernetes/runtimeclass_workloads/confidential? One thing worth noting with this is that to avoid sha caching/matching, we want to ensure that the content of our encrypted image is slightly different to the unencrypted one. We also want to ensure it's a multi-arch manifest container image so that we can easily work in multiple platforms.
"""

Requirements:

 * The image(s) should be multi-arch
 * The image(s) should be built by a Github workflow
 * To avoid headache, the image(s) should be different from the unencrypted one and not made from generic images (e.g. busybox) - see more information on @fitzthum 's https://github.com/kata-containers/kata-containers/issues/8111#issuecomment-2021592428
 
Any thoughts?

portersrc:
This issue came up in our weekly CoCo CI meeting. One open question to get this rolling:
    Which repo and folder should we use?

I proposed the CoCo [infra](https://github.com/confidential-containers/infra) repo. It could have a top-level folder called `container-images` with a Makefile and some supporting folders (for dockerfiles, configs, keys, etc.). If anyone has a better suggestion or strong preference, we can do that.

One consideration is that if we put this in kata-containers, then the test images will live in [kata-containers/packages](https://github.com/orgs/kata-containers/packages), versus [confidential-containers/packages](https://github.com/orgs/confidential-containers/packages) if we use infra, for example.

wainersm:
Hi @portersrc !

> This issue came up in our weekly CoCo CI meeting. One open question to get this rolling: Which repo and folder should we use?
>

It isn't an easy question I think.

[...]

PR #10009, opened 2024-07-14
Merge to main: supporting pull cosign signed images

### LINKED ISSUE #8120: CC: Add cosign signature verification to image pull on guest
This feature builds upon the initial basic image pull work in https://github.com/kata-containers/kata-containers/issues/8103 as one of the possible next steps to extend it to support more of the current `CCv0` feature set.

The scope of this is to add the ability to pull an container image on the guest and following the rules in a provided policy (using a provided public gpg key) we have validate the signature of the image with the cosign signing scheme, if required to ensure it hasn't been modified before we run an pod from it.

Initially I/we hoped to be able to cherry-pick over commits from `CCv0`, but given how much iteration there has been we're going to need to copy over chunks of the code and tests, improving as we go. Ideally we'll add `co-authored` tags to reflect the original authors though.

- With this feature we'll build on the guest-components integration already done and add signature specifics like:
- The signatures part of [`get_security_config` in image_rpc.rs](https://github.com/kata-containers/kata-containers/blob/1ad87faaf4d3068569d99c270df640e41465c410/src/agent/src/image_rpc.rs#L154-L160), though it should probably be refactored out of that method
- The addition in agent config.rs of `agent.image_policy_file`, (via https://github.com/kata-containers/kata-containers/commit/03d83914041fc8dc2bd7815414a619d94cbaf82c)
- Add addition of `agent.enable_signature_verification` config from https://github.com/kata-containers/kata-containers/pull/5552

- New tests
  - There are five potential new test cases I'd suggest, that are detailed in the acceptance criteria (some of which could be covered by unit tests). The code for this code be based on the existing `CCv0` tests for [crictl](https://github.com/kata-containers/tests/blob/6cf430642335c010014b5d9355a861328c4213ed/integration/containerd/confidential/agent_image.bats#L90-L109) and [kubernetes](https://github.com/kata-containers/tests/blob/6cf430642335c010014b5d9355a861328c4213ed/integration/kubernetes/confidential/agent_image.bats#L96-L111) and [lib functions](https://github.com/kata-containers/tests/blob/6cf430642335c010014b5d9355a861328c4213ed/integration/confidential/lib.sh#L288-L309), but we want to work to make them easier to apply to different KBCs/TEE by extracting out the KBC/TEE logic into library functions, so the test cases are cleaner, similar to the approach done in https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/k8s-confidential.bats rather than a pure cherry-pick
  - We need to set-up the sample KBS resources: based on #8870
  - We also want to automatically create and sign (if applicable) the test images that we are using as part of a workflow like we do with the unencrypted test image: https://github.com/kata-containers/kata-containers/tree/main/tests/integration/kubernetes/runtimeclass_workloads/confidential in order to control this ourselves and not rely on manual creation in our pipelines
    > **Note:** we need to specify different test images for each of the different arches as the digest will be different for each, so we can't use a multi-arch manifest here (but we can still create it in a multi-arch way

## Questions
- In `CCv0` we decided to make `agent.enable_signature_verification` default to true for security reasons and therefore had to set `agent.enable_signature_verification=false` as a default in the configuration.toml (see https://github.com/kata-containers/kata-containers/pull/5552). When merging into main I think we want to have a discussion if this configuration should still be the default in `main` and what configuration.toml/runtime classes we'll have specifically for CC cases (which related back to a kata-deploy piece)

## Acceptance Criteria

For each supported platform (different TEEs and kubernetes/crictl):

For all scenarios the common context is:
> **Note:** this assumes that `agent.enable_signature_verification` defaults to true

**Given** I have a version of kata deployed with a guest image that has an agent with guest_pull feature enabled and the attestation-agent installed, as well as the and nydus-snapshotter installed and configured for [guest-pulling](https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/config-coco-guest-pulling.toml) and the kata configuration and doesn't have the setting `agent.enable_signature_verification=false` (unless stated)

**Scenario:** Regression: Create a pod from an unsigned image, on an `insecureAcceptAnything` registry works
**Given** A public unsigned container image *i* from registry *r* and a KBS resources configured to pass through a policy file of
```
{
    "default": [{"type": "insecureAcceptAnything"}]
}
```
e.g. for the offline_fs_kbc it would have a `/etc/aa-offline_fs_kbc-resources.json` file containing:
```
{
  "default/security-policy/test": "ewogICAgImRlZmF1bHQiOiBbeyJ0eXBlIjogImluc2VjdXJlQWNjZXB0QW55dGhpbmcifV0KfQ==",
}
```
is present in the guest image
**When** I try and create a pod from *i*
**Then** The pod is successfully created and runs

**Scenario:** Create a pod from an unsigned image, on a 'restricted registry' is rejected
**Given** A public unsigned container image *i* from registry *r* and a KBS resources, or agent config configured to read a policy file like:
```
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "<r>": [
                {
                    "type": "sigstoreSigned",
                    "keyPath": "kbs:///default/cosign-public-key/test"
                }
            ]
        }
    }
}
```

**When** I try and create a pod from *i*
**Then** The pod is not created an we get an appropriate error message in the logs e.g. `Validate image failed`

**Scenario:** Create a pod from a signed image, on a 'restricted registry' is successful
**Given** A public signed container image *s* signed with a key *k* from registry *r* and a KBS resources or agent config configured to read a policy file like:
```
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "<r>": [
                {
                    "type": "sigstoreSigned",
                     "keyPath": "kbs:///default/cosign-public-key/test"
                }
            ]
        }
    }
}
```
**And** the public gpg key is provide in the resource like:
```
"default/cosign-public-key/test": "<base64 encoding of k>"
```
**When** I try and create a pod from *i*
**Then** The pod is successfully created and runs

**Scenario:** Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected
**Given** A public signed container image *s* signed with a key *k* and a different key *w* from registry *r* and a KBS resources or agent config configured to read a policy file like:
```
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "<r>": [
                {
                    "type": "sigstoreSigned",
                    "keyPath": "kbs:///default/cosign-public-key/test"
                }
            ]
        }
    }
}
```
**And** the public gpg key is provide in the resource like:
```
"default/cosign-public-key/test": "<base64 encoding of w>"
```
**When** I try and create a pod from *i*
**Then** The pod is not created an we get an appropriate error message in the logs e.g. `Validate image failed`

**Scenario:** Create a pod from an unsigned image, on a 'restricted registry' works if `enable_signature_verfication` is false
**Given** The kata configuration and has the setting `agent.enable_signature_verification=false` and a public unsigned container image *i* from registry *r* and a KBS resources configured or agent config configured to read a policy file like:
```
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "<r>": [
                {
                    "type": "sigstoreSigned",
                    "keyPath": "kbs:///default/cosign-public-key/test"
                }
            ]
        }
    }
}
```
**When** I try and create a pod from *i*
**Then** The signature check is not enforced, so the pod is created successfully

Xynnn007:
Close #8120.

The test commit has not been testsed and I just followed what was coded in https://github.com/kata-containers/kata-containers/pull/9904.

On top of #9904.

Let's continue this after #9904 is merged.

cc @stevenhorsman

### DIFF
--- a/src/agent/Makefile
+++ b/src/agent/Makefile
@@ -159,7 +159,7 @@ vendor:
 
 #TARGET test: run cargo tests
 test: $(GENERATED_FILES)
-	@RUST_LIB_BACKTRACE=0 cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
+	@RUST_LIB_BACKTRACE=0 RUST_BACKTRACE=1 cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
 
 ##TARGET check: run test
 check: $(GENERATED_FILES) standard_rust_check
--- a/src/agent/README.md
+++ b/src/agent/README.md
@@ -131,6 +131,8 @@ The kata agent has the ability to configure agent options in guest kernel comman
 | `agent.cdh_api_timeout` | Confidential Data Hub (CDH) API timeout | Allow to configure CDH API timeout(seconds) | integer | `50` |
 | `agent.https_proxy` | HTTPS proxy | Allow to configure `https_proxy` in the guest | string | `""` |
 | `agent.image_registry_auth` | Image registry credential URI | The URI to where image-rs can find the credentials for pulling images from private registries e.g. `file:///root/.docker/config.json` to read from a file in the guest image, or `kbs:///default/credentials/test` to get the file from the KBS| string | `""` |
+| `agent.enable_signature_verification` | Image security policy flag | Whether enable image security policy enforcement. If `true`, the resource indexed by URI `agent.image_policy_file` will be got to work as image pulling policy. | string | `""` |
+| `agent.image_policy_file` | Image security policy URI | The URI to where image-rs Typical policy URIs are like `file:///etc/image.json` to read from a file in the guest image, or `kbs:///default/security-policy/test` to get the file from the KBS| string | `""` |
 | `agent.log` | Log level | Allow the agent log level to be changed (produces more or less output) | string | `"info"` |
 | `agent.log_vport` | Log port | Allow to specify the `vsock` port to read logs | integer | `0` |
 | `agent.no_proxy` | NO proxy | Allow to configure `no_proxy` in the guest | string | `""` |
--- a/src/agent/src/config.rs
+++ b/src/agent/src/config.rs
@@ -34,6 +34,12 @@ const GUEST_COMPONENTS_PROCS_OPTION: &str = "agent.guest_components_procs";
 const IMAGE_REGISTRY_AUTH_OPTION: &str = "agent.image_registry_auth";
 const SECURE_STORAGE_INTEGRITY_OPTION: &str = "agent.secure_storage_integrity";
 
+#[cfg(feature = "guest-pull")]
+const ENABLE_SIGNATURE_VERIFICATION: &str = "agent.enable_signature_verification";
+
+#[cfg(feature = "guest-pull")]
+const IMAGE_POLICY_FILE: &str = "agent.image_policy_file";
+
 // Configure the proxy settings for HTTPS requests in the guest,
 // to solve the problem of not being able to access the specified image in some cases.
 const HTTPS_PROXY: &str = "agent.https_proxy";
@@ -115,6 +121,10 @@ pub struct AgentConfig {
     #[cfg(feature = "guest-pull")]
     pub image_registry_auth: String,
     pub secure_storage_integrity: bool,
+    #[cfg(feature = "guest-pull")]
+    pub enable_signature_verification: bool,
+    #[cfg(feature = "guest-pull")]
+    pub image_policy_file: String,
 }
 
 #[derive(Debug, Deserialize)]
@@ -138,6 +148,10 @@ pub struct AgentConfigBuilder {
     #[cfg(feature = "guest-pull")]
     pub image_registry_auth: Option<String>,
     pub secure_storage_integrity: Option<bool>,
+    #[cfg(feature = "guest-pull")]
+    pub enable_signature_verification: Option<bool>,
+    #[cfg(feature = "guest-pull")]
+    pub image_policy_file: Option<String>,
 }
 
 macro_rules! config_override {
@@ -207,6 +221,10 @@ impl Default for AgentConfig {
             #[cfg(feature = "guest-pull")]
             image_registry_auth: String::from(""),
             secure_storage_integrity: false,
+            #[cfg(feature = "guest-pull")]
+            enable_signature_verification: false,
+            #[cfg(feature = "guest-pull")]
+            image_policy_file: String::from(""),
         }
     }
 }
@@ -246,8 +264,17 @@ impl FromStr for AgentConfig {
         );
         config_override!(agent_config_builder, agent_config, guest_components_procs);
         #[cfg(feature = "guest-pull")]
-        config_override!(agent_config_builder, agent_config, image_registry_auth);
+        {
+            config_override!(agent_config_builder, agent_config, image_registry_auth);
+            config_override!(
+                agent_config_builder,
+                agent_config,
+                enable_signature_verification
+            );
+            config_override!(agent_config_builder, agent_config, image_policy_file);
+        }
         config_override!(agent_config_builder, agent_config, secure_storage_integrity);
+
         Ok(agent_config)
     }
 }
@@ -372,12 +399,26 @@ impl AgentConfig {
                 get_guest_components_procs_value
             );
             #[cfg(feature = "guest-pull")]
-            parse_cmdline_param!(
-                param,
-                IMAGE_REGISTRY_AUTH_OPTION,
-                config.image_registry_auth,
-                get_string_value
-            );
+            {
+                parse_cmdline_param!(
+                    param,
+                    IMAGE_REGISTRY_AUTH_OPTION,
+                    config.image_registry_auth,
+                    get_string_value
+                );
+                parse_cmdline_param!(
+                    param,
+                    ENABLE_SIGNATURE_VERIFICATION,
+                    config.enable_signature_verification,
+                    get_bool_value
+                );
+                parse_cmdline_param!(
+                    param,
+                    IMAGE_POLICY_FILE,
+                    config.image_policy_file,
+                    get_string_value
+                );
+            }
             parse_cmdline_param!(
                 param,
                 SECURE_STORAGE_INTEGRITY_OPTION,
@@ -585,6 +626,11 @@ mod tests {
         assert!(!config.dev_mode);
         assert_eq!(config.log_level, DEFAULT_LOG_LEVEL);
         assert_eq!(config.hotplug_timeout, DEFAULT_HOTPLUG_TIMEOUT);
+        #[cfg(feature = "guest-pull")]
+        {
+            assert!(!config.enable_signature_verification);
+            assert_eq!(config.image_policy_file, "");
+        }
     }
 
     #[test]
@@ -612,6 +658,10 @@ mod tests {
             #[cfg(feature = "guest-pull")]
             image_registry_auth: &'a str,
             secure_storage_integrity: bool,
+            #[cfg(feature = "guest-pull")]
+            enable_signature_verification: bool,
+            #[cfg(feature = "guest-pull")]
+            image_policy_file: &'a str,
         }
 
         impl Default for TestData<'_> {
@@ -634,6 +684,10 @@ mod tests {
                     #[cfg(feature = "guest-pull")]
                     image_registry_auth: "",
                     secure_storage_integrity: false,
+                    #[cfg(feature = "guest-pull")]
+                    enable_signature_verification: false,
+                    #[cfg(feature = "guest-pull")]
+                    image_policy_file: "",
                 }
             }
         }
@@ -1102,6 +1156,24 @@ mod tests {
                 secure_storage_integrity: false,
                 ..Default::default()
             },
+            #[cfg(feature = "guest-pull")]
+            TestData {
+                contents: "agent.enable_signature_verification=true",
+                enable_signature_verification: true,
+                ..Default::default()
+            },
+            #[cfg(feature = "guest-pull")]
+            TestData {
+                contents: "agent.image_policy_file=kbs:///default/image-policy/test",
+                image_policy_file: "kbs:///default/image-policy/test",
+                ..Default::default()
+            },
+            #[cfg(feature = "guest-pull")]
+            TestData {
+                contents: "agent.image_policy_file=file:///etc/image-policy.json",
+                image_policy_file: "file:///etc/image-policy.json",
+                ..Default::default()
+            },
         ];
 
         let dir = tempdir().expect("failed to create tmpdir");
@@ -1162,13 +1234,20 @@ mod tests {
                 msg
             );
             #[cfg(feature = "guest-pull")]
-            assert_eq!(d.image_registry_auth, config.image_registry_auth, "{}", msg);
+            {
+                assert_eq!(d.image_registry_auth, config.image_registry_auth, "{}", msg);
+                assert_eq!(
+                    d.enable_signature_verification, config.enable_signature_verification,
+                    "{}",
+                    msg
+                );
+                assert_eq!(d.image_policy_file, config.image_policy_file, "{}", msg);
+            }
             assert_eq!(
                 d.secure_storage_integrity, config.secure_storage_integrity,
                 "{}",
                 msg
             );
-
             for v in vars_to_unset {
                 env::remove_var(v);
             }
--- a/src/agent/src/image.rs
+++ b/src/agent/src/image.rs
@@ -57,13 +57,27 @@ impl ImageService {
     pub fn new() -> Self {
         let mut image_client = ImageClient::new(PathBuf::from(KATA_IMAGE_WORK_DIR));
         #[cfg(feature = "guest-pull")]
-        if !AGENT_CONFIG.image_registry_auth.is_empty() {
-            let registry_auth = &AGENT_CONFIG.image_registry_auth;
-            debug!(sl(), "Set registry auth file {:?}", registry_auth);
-            image_client.config.file_paths.auth_file = registry_auth.clone();
-            image_client.config.auth = true;
-        }
+        {
+            if !AGENT_CONFIG.image_registry_auth.is_empty() {
+                let registry_auth = &AGENT_CONFIG.image_registry_auth;
+                debug!(sl(), "Set registry auth file {:?}", registry_auth);
+                image_client.config.file_paths.auth_file = registry_auth.clone();
+                image_client.config.auth = true;
+            }
 
+            let enable_signature_verification = &AGENT_CONFIG.enable_signature_verification;
+            debug!(
+                sl(),
+                "Enable image signature verification: {:?}", enable_signature_verification
+            );
+            image_client.config.security_validate = *enable_signature_verification;
+
+            if !AGENT_CONFIG.image_policy_file.is_empty() {
+                let image_policy_file = &AGENT_CONFIG.image_policy_file;
+                debug!(sl(), "Use imagepolicy file {:?}", image_policy_file);
+                image_client.config.file_paths.policy_path = image_policy_file.clone();
+            }
+        }
         Self { image_client }
     }
 
--- a/tests/integration/kubernetes/confidential_common.sh
+++ b/tests/integration/kubernetes/confidential_common.sh
@@ -8,6 +8,8 @@
 source "${BATS_TEST_DIRNAME}/tests_common.sh"
 source "${BATS_TEST_DIRNAME}/../../common.bash"
 
+load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
+
 SUPPORTED_TEE_HYPERVISORS=("qemu-sev" "qemu-snp" "qemu-tdx" "qemu-se")
 SUPPORTED_NON_TEE_HYPERVISORS=("qemu-coco-dev")
 
@@ -97,7 +99,7 @@ function cleanup_loop_device(){
 	local loop_file="${1:-/tmp/trusted-image-storage.img}"
 	# Find all loop devices associated with $loop_file
 	local existed_devices=$(sudo losetup -j $loop_file | awk -F'[: ]' '{print $1}')
-	
+
 	if [ -n "$existed_devices" ]; then
 		# Iterate over each found loop device and detach it
 		for d in $existed_devices; do
@@ -106,4 +108,63 @@ function cleanup_loop_device(){
 	fi
 
 	sudo rm -f "$loop_file" >/dev/null 2>&1 || true
+}
+
+# This function creates pod yaml. Parameters
+# - $1: image reference
+# - $2: image policy file. If given, `enable_signature_verification` will be set to true
+# - $3: image registry auth.
+# - $4: guest components procs parameter
+# - $5: guest components rest api parameter
+# - $6: node
+function create_coco_pod_yaml() {
+	image=$1
+	image_policy=${2:-}
+	image_registry_auth=${3:-}
+	guest_components_procs=${4:-}
+	guest_components_rest_api=${5:-}
+	node=${6:-}
+
+	local CC_KBS_ADDR
+	export CC_KBS_ADDR=$(kbs_k8s_svc_http_addr)
+
+	kernel_params_annotation="io.katacontainers.config.hypervisor.kernel_params"
+	kernel_params_value=""
+
+	if [ -n "$image_policy" ]; then
+		kernel_params_value+=" agent.image_policy_file=${image_policy}"
+		kernel_params_value+=" agent.enable_signature_verification=true"
+	fi
+
+	if [ -n "$image_registry_auth" ]; then
+		kernel_params_value+=" agent.image_registry_auth=${image_registry_auth}"
+	fi
+
+	if [ -n "$guest_components_procs" ]; then
+		kernel_params_value+=" agent.guest_components_procs=${guest_components_procs}"
+	fi
+
+	if [ -n "$guest_components_rest_api" ]; then
+		kernel_params_value+=" agent.guest_components_rest_api=${guest_components_rest_api}"
+	fi
+
+	kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"
+
+	# Note: this is not local as we use it in the caller test
+	kata_pod="$(new_pod_config "$image" "kata-${KATA_HYPERVISOR}")"
+	set_container_command "${kata_pod}" "0" "sleep" "30"
+
+	# Set annotations
+	set_metadata_annotation "${kata_pod}" \
+		"io.containerd.cri.runtime-handler" \
+		"kata-${KATA_HYPERVISOR}"
+	set_metadata_annotation "${kata_pod}" \
+		"${kernel_params_annotation}" \
+		"${kernel_params_value}"
+
+	add_allow_all_policy_to_yaml "${kata_pod}"
+
+	if [ -n "$node" ]; then
+		set_node "${kata_pod}" "$node"
+	fi
 }
--- a/tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
@@ -7,7 +7,6 @@
 
 load "${BATS_TEST_DIRNAME}/lib.sh"
 load "${BATS_TEST_DIRNAME}/confidential_common.sh"
-load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
 
 export KBS="${KBS:-false}"
 
@@ -23,6 +22,7 @@ setup() {
     UNSIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:unsigned"
     COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed"
     COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed-key2"
+    SECURITY_POLICY_KBS_URI="kbs:///default/security-policy/test"
 }
 
 function setup_kbs_image_policy() {
@@ -32,24 +32,24 @@ function setup_kbs_image_policy() {
 
     default_policy="${1:-insecureAcceptAnything}"
     policy_json=$(cat << EOF
-{                 
+{
     "default": [
-        {                                      
+        {
         "type": "${default_policy}"
         }
-    ],                 
+    ],
     "transports": {
-        "docker": {                                                           
+        "docker": {
             "ghcr.io/confidential-containers/test-container-image-rs": [
-                {                                        
+                {
                     "type": "sigstoreSigned",
                     "keyPath": "kbs:///default/cosign-public-key/test"
-                }     
-            ],                             
+                }
+            ],
             "quay.io/prometheus": [
-                {                                               
+                {
                     "type": "insecureAcceptAnything"
-                }    
+                }
             ]
         }
     }
@@ -69,46 +69,16 @@ EOF
     kbs_set_resource "default" "cosign-public-key" "test" "${public_key}"
 }
 
-function create_pod_yaml_with_signed_image() {
-    image=$1
-    image_security=${2:-true}
-
-    local CC_KBS_ADDR
-    export CC_KBS_ADDR=$(kbs_k8s_svc_http_addr)
-    kernel_params_annotation="io.katacontainers.config.hypervisor.kernel_params"
-    kernel_params_value="agent.guest_components_rest_api=resource"
-
-    if [[ $image_security == true ]]; then
-        kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"
-        kernel_params_value+=" agent.enable_signature_verification=true"
-        kernel_params_value+=" agent.image_policy_file=kbs:///default/security-policy/test"
-    fi
-
-    # Note: this is not local as we use it in the caller test
-    kata_pod_with_signed_image="$(new_pod_config "$image" "kata-${KATA_HYPERVISOR}")"
-    set_container_command "${kata_pod_with_signed_image}" "0" "sleep" "30"
-
-    # Set annotation to pull image in guest
-    set_metadata_annotation "${kata_pod_with_signed_image}" \
-        "io.containerd.cri.runtime-handler" \
-        "kata-${KATA_HYPERVISOR}"
-    set_metadata_annotation "${kata_pod_with_signed_image}" \
-        "${kernel_params_annotation}" \
-        "${kernel_params_value}"
-
-    add_allow_all_policy_to_yaml "${kata_pod_with_signed_image}"
-}
-
 @test "Create a pod from an unsigned image, on an insecureAcceptAnything registry works" {
     # We want to set the default policy to be reject to rule out false positives
     setup_kbs_image_policy "reject"
 
-    create_pod_yaml_with_signed_image "${UNSIGNED_UNPROTECTED_REGISTRY_IMAGE}"
+    create_coco_pod_yaml "${UNSIGNED_UNPROTECTED_REGISTRY_IMAGE}" "${SECURITY_POLICY_KBS_URI}" "" "" "resource" "$node"
 
     # For debug sake
-    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+    echo "Pod ${kata_pod}: $(cat ${kata_pod})"
 
-    k8s_create_pod "${kata_pod_with_signed_image}"
+    k8s_create_pod "${kata_pod}"
     echo "Kata pod test-e2e from image security policy is running"
 }
 
@@ -116,12 +86,12 @@ function create_pod_yaml_with_signed_image() {
     # We want to leave the default policy to be insecureAcceptAnything to rule out false negatives
     setup_kbs_image_policy
 
-    create_pod_yaml_with_signed_image "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}"
+    create_coco_pod_yaml "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}" "${SECURITY_POLICY_KBS_URI}" "" "" "resource" "$node"
 
     # For debug sake
-    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+    echo "Pod ${kata_pod}: $(cat ${kata_pod})"
 
-    assert_pod_fail "${kata_pod_with_signed_image}"
+    assert_pod_fail "${kata_pod}"
     assert_logs_contain "${node}" kata "${node_start_time}" "Security validate failed: Validate image failed: Cannot pull manifest"
 }
 
@@ -129,12 +99,12 @@ function create_pod_yaml_with_signed_image() {
     # We want to set the default policy to be reject to rule out false positives
     setup_kbs_image_policy "reject"
 
-    create_pod_yaml_with_signed_image "${COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE}"
+    create_coco_pod_yaml "${COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE}" "${SECURITY_POLICY_KBS_URI}" "" "" "resource" "$node"
 
     # For debug sake
-    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+    echo "Pod ${kata_pod}: $(cat ${kata_pod})"
 
-    k8s_create_pod "${kata_pod_with_signed_image}"
+    k8s_create_pod "${kata_pod}"
     echo "Kata pod test-e2e from image security policy is running"
 }
 
@@ -142,25 +112,25 @@ function create_pod_yaml_with_signed_image() {
     # We want to leave the default policy to be insecureAcceptAnything to rule out false negatives
     setup_kbs_image_policy
 
-    create_pod_yaml_with_signed_image "${COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE}"
+    create_coco_pod_yaml "${COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE}" "${SECURITY_POLICY_KBS_URI}" "" "" "resource" "$node"
 
     # For debug sake
-    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+    echo "Pod ${kata_pod}: $(cat ${kata_pod})"
 
-    assert_pod_fail "${kata_pod_with_signed_image}"
+    assert_pod_fail "${kata_pod}"
     assert_logs_contain "${node}" kata "${node_start_time}" "Security validate failed: Validate image failed: \[PublicKeyVerifier"
 }
 
-@test "Create a pod from an unsigned image, on a 'restricted registry' works if enable_signature_verfication is false" {
+@test "Create a pod from an unsigned image, on a 'restricted registry' works if policy files isn't set" {
     # We want to set the default policy to be reject to rule out false positives
     setup_kbs_image_policy "reject"
 
-    create_pod_yaml_with_signed_image "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}" "false"
+    create_coco_pod_yaml "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}" "" "" "" "resource" "$node"
 
     # For debug sake
-    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+    echo "Pod ${kata_pod}: $(cat ${kata_pod})"
 
-    k8s_create_pod "${kata_pod_with_signed_image}"
+    k8s_create_pod "${kata_pod}"
     echo "Kata pod test-e2e from image security policy is running"
 }
 
--- a/tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
@@ -17,11 +17,16 @@ setup() {
 
     [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
 
+    tag_suffix=""
+    if [ "$(uname -m)" != "x86_64" ]; then
+        tag_suffix="-$(uname -m)"
+    fi
+
     setup_common
     UNSIGNED_UNPROTECTED_REGISTRY_IMAGE="quay.io/prometheus/busybox:latest"
-    UNSIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:unsigned"
-    COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed"
-    COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed-key2"
+    UNSIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:unsigned${tag_suffix}"
+    COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed${tag_suffix}"
+    COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed-key2${tag_suffix}"
     SECURITY_POLICY_KBS_URI="kbs:///default/security-policy/test"
 }
 
--- a/tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
+++ b/tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
@@ -1,0 +1,181 @@
+#!/usr/bin/env bats
+# Copyright (c) 2024 IBM Corporation
+# Copyright (c) 2024 Alibaba Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+load "${BATS_TEST_DIRNAME}/lib.sh"
+load "${BATS_TEST_DIRNAME}/confidential_common.sh"
+load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
+
+export KBS="${KBS:-false}"
+
+setup() {
+    if ! is_confidential_runtime_class; then
+        skip "Test not supported for ${KATA_HYPERVISOR}."
+    fi
+
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
+    setup_common
+    UNSIGNED_UNPROTECTED_REGISTRY_IMAGE="quay.io/prometheus/busybox:latest"
+    UNSIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:unsigned"
+    COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed"
+    COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE="ghcr.io/confidential-containers/test-container-image-rs:cosign-signed-key2"
+}
+
+function setup_kbs_image_policy() {
+    if [ "${KBS}" = "false" ]; then
+        skip "Test skipped as KBS not setup"
+    fi
+
+    default_policy="${1:-insecureAcceptAnything}"
+    policy_json=$(cat << EOF
+{                 
+    "default": [
+        {                                      
+        "type": "${default_policy}"
+        }
+    ],                 
+    "transports": {
+        "docker": {                                                           
+            "ghcr.io/confidential-containers/test-container-image-rs": [
+                {                                        
+                    "type": "sigstoreSigned",
+                    "keyPath": "kbs:///default/cosign-public-key/test"
+                }     
+            ],                             
+            "quay.io/prometheus": [
+                {                                               
+                    "type": "insecureAcceptAnything"
+                }    
+            ]
+        }
+    }
+}
+EOF
+    )
+
+    # This public key is corresponding to a private key that was generated to test signed images in image-rs CI.
+    # TODO: Update the CI to generate a signed image together with verification. See issue #9360
+    public_key=$(curl -sSL "https://raw.githubusercontent.com/confidential-containers/guest-components/075b9a9ee77227d9d92b6f3649ef69de5e72d204/image-rs/test_data/signature/cosign/cosign1.pub")
+
+    if ! is_confidential_hardware; then
+        kbs_set_allow_all_resources
+    fi
+
+    kbs_set_resource "default" "security-policy" "test" "${policy_json}"
+    kbs_set_resource "default" "cosign-public-key" "test" "${public_key}"
+}
+
+function create_pod_yaml_with_signed_image() {
+    image=$1
+    image_security=${2:-true}
+
+    local CC_KBS_ADDR
+    export CC_KBS_ADDR=$(kbs_k8s_svc_http_addr)
+    kernel_params_annotation="io.katacontainers.config.hypervisor.kernel_params"
+    kernel_params_value="agent.guest_components_rest_api=resource"
+
+    if [[ $image_security == true ]]; then
+        kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"
+        kernel_params_value+=" agent.enable_signature_verification=true"
+        kernel_params_value+=" agent.image_policy_file=kbs:///default/security-policy/test"
+    fi
+
+    # Note: this is not local as we use it in the caller test
+    kata_pod_with_signed_image="$(new_pod_config "$image" "kata-${KATA_HYPERVISOR}")"
+    set_container_command "${kata_pod_with_signed_image}" "0" "sleep" "30"
+
+    # Set annotation to pull image in guest
+    set_metadata_annotation "${kata_pod_with_signed_image}" \
+        "io.containerd.cri.runtime-handler" \
+        "kata-${KATA_HYPERVISOR}"
+    set_metadata_annotation "${kata_pod_with_signed_image}" \
+        "${kernel_params_annotation}" \
+        "${kernel_params_value}"
+
+    add_allow_all_policy_to_yaml "${kata_pod_with_signed_image}"
+}
+
+@test "Create a pod from an unsigned image, on an insecureAcceptAnything registry works" {
+    # We want to set the default policy to be reject to rule out false positives
+    setup_kbs_image_policy "reject"
+
+    create_pod_yaml_with_signed_image "${UNSIGNED_UNPROTECTED_REGISTRY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    k8s_create_pod "${kata_pod_with_signed_image}"
+    echo "Kata pod test-e2e from image security policy is running"
+}
+
+@test "Create a pod from an unsigned image, on a 'restricted registry' is rejected" {
+    # We want to leave the default policy to be insecureAcceptAnything to rule out false negatives
+    setup_kbs_image_policy
+
+    create_pod_yaml_with_signed_image "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    assert_pod_fail "${kata_pod_with_signed_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" "Security validate failed: Validate image failed: Cannot pull manifest"
+}
+
+@test "Create a pod from a signed image, on a 'restricted registry' is successful" {
+    # We want to set the default policy to be reject to rule out false positives
+    setup_kbs_image_policy "reject"
+
+    create_pod_yaml_with_signed_image "${COSIGN_SIGNED_PROTECTED_REGISTRY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    k8s_create_pod "${kata_pod_with_signed_image}"
+    echo "Kata pod test-e2e from image security policy is running"
+}
+
+@test "Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected" {
+    # We want to leave the default policy to be insecureAcceptAnything to rule out false negatives
+    setup_kbs_image_policy
+
+    create_pod_yaml_with_signed_image "${COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    assert_pod_fail "${kata_pod_with_signed_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" "Security validate failed: Validate image failed: \[PublicKeyVerifier"
+}
+
+@test "Create a pod from an unsigned image, on a 'restricted registry' works if enable_signature_verfication is false" {
+    # We want to set the default policy to be reject to rule out false positives
+    setup_kbs_image_policy "reject"
+
+    create_pod_yaml_with_signed_image "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}" "false"
+
+    # For debug sake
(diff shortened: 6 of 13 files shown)


src/agent/src/image.rs
```diff
@@ -63,6 +63,27 @@ impl ImageService {
...
+            let image_policy_file = &AGENT_CONFIG.image_policy_file;
+            debug!(sl(), "Use imagepolicy file {:?}", image_policy_file);
+            image_client.config.file_paths.policy_path = image_policy_file.clone();
+        }
+
+        #[cfg(feature = "guest-pull")]
+        if !&AGENT_CONFIG.simple_signing_sigstore_config.is_empty() {
+            let simple_signing_sigstore_config = &AGENT_CONFIG.simple_signing_sigstore_config;
+            debug!(sl(), "Using RedHat Simple Signing image signature file {:?}", simple_signing_sigstore_config);
+            image_client.config.file_paths.sigstore_config = simple_signing_sigstore_config.clone();
```
stevenhorsman, line 84: Maybe it's an anti-pattern, but did you consider putting all of these sections inside the same `  #[cfg(feature = "guest-pull")] {` block?

src/agent/src/config.rs
```diff
@@ -198,6 +219,12 @@ impl Default for AgentConfig {
             guest_components_procs: GuestComponentsProcs::default(),
             #[cfg(feature = "guest-pull")]
             image_registry_auth: String::from(""),
+            #[cfg(feature = "guest-pull")]
+            enable_signature_verification: false,
```
stevenhorsman, line 223: Something to be aware of - in `CCv0` we decided for security sack to default to have verification on. You've changed this, which I think is probably the right decision due to the impact it will have on vanilla kata containers, but we want to ensure that we strongly document that somewhere in CoCo that it is opt-in.

stevenhorsman (commented):
Initial comments. I'll go through the tests more later

stevenhorsman (commented):
Some comments on the test cases. In addition to this there is a lot of duplicated code from the other test cases, but @wainersm and I can probably have an action to help improve this after the features have gone it.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,197 @@
...
+    # TODO - do TEEs want to run these tests?
+    # if is_confidential_hardware; then
+    #     skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    # fi
+
+    if ! is_confidential_runtime_class; then
+        skip "Test not supported for ${KATA_HYPERVISOR}."
+    fi
+
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
```
littlejawa, line 24: I think this is going to skip the test for CRI-O - we're not using a snapshotter there.
Any way to verify that we're running with containerd before we check the snapshotter?

stevenhorsman: That's a really good poin, but we have this same check in all of our guest-pull tests as we don't have any upstream testing involving cri-o at the moment, so it's effectively not supported. I think probably the best way to handle this is as part of cri-o enabling that @wainersm has been looking at?

We also need to find a way to remove the duplications and help make this sort of thing easier.

littlejawa: Answering to myself: this is already the case for generic guest-pull feature, so I guess it is unrelated to this PR.

src/agent/README.md
```diff
@@ -125,10 +125,14 @@ The kata agent has the ability to configure agent options in guest kernel comman
...
+| `agent.enable_signature_verification` | Image security policy flag | Whether enable image security policy enforcement. If `true`, the resource indexed by URI `agent.image_policy_file` will be got to work as image pulling policy. | string | `""` |
+| `agent.image_policy_file` | Image security policy URI | The URI to where image-rs Typical policy URIs are like `file:///etc/image.json` to read from a file in the guest image, or `kbs:///default/security-policy/test` to get the file from the KBS| string | `""` |
+| `agent.simple_signing_sigstore_config` | Redhat Simple Signing scheme signature URI | The URI of the signatures that RedHat Simple Signing image signing scheme uses e.g. `file:///run/image-security/simple_signing/sigstore_config/ssample.sig` to read from a file in the guest image, or `kbs:///default/sigstore-config/testtest` to get the file from the KBS| string | `""` |
```
fitzthum, line 135: Do you think these options are going to be superseded by the CDH config file? I guess maybe we could continue to support them for platforms that do not support init-data but that might be a bit messy. Is it going to be weird to add these and then later remove them?

src/agent/src/config.rs
```diff
@@ -570,6 +583,7 @@ mod tests {
             no_proxy: &'a str,
             guest_components_rest_api: GuestComponentsFeatures,
             guest_components_procs: GuestComponentsProcs,
+            image_registry_auth: &'a str,
```
c3d, line 645: Since the field is under a `#[cfg(feature = "guest-pull")]`, don't you need the same on that code as well?

src/agent/src/config.rs
```diff
@@ -589,6 +603,7 @@ mod tests {
                     no_proxy: "",
                     guest_components_rest_api: GuestComponentsFeatures::default(),
                     guest_components_procs: GuestComponentsProcs::default(),
+                    image_registry_auth: "",
```
c3d, line 665: Since the field is under a #[cfg(feature = "guest-pull")], don't you need the same on that code as well?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,197 @@
...
+    setup_kbs_image_policy
+
+    create_pod_yaml_with_signed_image "${UNSIGNED_UNPROTECTED_REGISTRY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    k8s_create_pod "${kata_pod_with_signed_image}"
+    echo "Kata pod test-e2e from image security policy is running"
+}
```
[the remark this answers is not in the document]
fitzthum: With standard image pull we don't have `agent.enable_signature_verification=true`, right?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
littlejawa: Interesting, github have let me comment after you without showing me your comment :-/ 
I actually thought we have one crio test running (on garm?) but maybe it's not running those specific tests?

stevenhorsman: Yeah, no confidential image test with cri-o yet

src/agent/README.md
stevenhorsman: I'm not sure about the CDH config and init-data. My initial reaction is that we need something for platforms that aren't using it and it's also not available currently, so I'm not sure we have another approach at the moment?

src/agent/src/config.rs
gkurz: The 4 initial commits were borrowed from #9904 actually. They will be removed at some point.

src/agent/Makefile
```diff
@@ -159,7 +159,7 @@ vendor:
 
 #TARGET test: run cargo tests
 test: $(GENERATED_FILES)
-	@cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
+	@RUST_LIB_BACKTRACE=0 cargo test --all --target $(TRIPLE) $(EXTRA_RUSTFEATURES) -- --nocapture
```
gkurz, line 162: Commit description mentions `RUST_BACKTRACE=0`, here it is `RUST_LIB_BACKTRACE=0` and https://docs.rs/anyhow/latest/anyhow/ says :

```
If you want only panics to have backtraces, set RUST_BACKTRACE=1 and RUST_LIB_BACKTRACE=0
```

I suspect that description has a typo and should mention `RUST_LIB_BACKTRACE=0`, correct ?

I don't understand why `RUST_BACKTRACE=1` isn't explicitly passed to `cargo test` though. If it is set elsewhere, please provide an explanation in the commit description.

Also this is fixing a `make test` failure introduced by the previous commit IIUC. If so, you should either move this commit before or squash into the previous commit, for the sake of bisectability.

tools/packaging/static-build/agent/Dockerfile
```diff
@@ -26,16 +26,17 @@ RUN mkdir ${RUSTUP_HOME} ${CARGO_HOME} && chmod -R a+rwX /opt
 RUN apt-get update && \
 	apt-get --no-install-recommends -y install \
 		ca-certificates \
+		clang \
 		curl \
 		g++ \
 		gcc \
+		libprotobuf-dev \
```
gkurz, line 33: Is this fixing some build break ? Can bisectability be preserved ?

stevenhorsman: FYI Greg, this code was picked up from https://github.com/kata-containers/kata-containers/pull/9382 which I merged for bisectablility based on your feedback. In this PR, only the last two commits are new

src/agent/src/image.rs
```diff
@@ -63,6 +63,27 @@ impl ImageService {
...
 
+        #[cfg(feature = "guest-pull")]
+        {
+            let enable_signature_verification = &AGENT_CONFIG.enable_signature_verification;
+            debug!(sl(), "Enable image signature verification: {:?}", enable_signature_verification);
+            image_client.config.security_validate = enable_signature_verification.clone();
+        }
+
+        #[cfg(feature = "guest-pull")]
+        if !&AGENT_CONFIG.image_policy_file.is_empty() {
```
gkurz, line 74: No need to take a reference as mentioned in https://github.com/kata-containers/kata-containers/pull/9904#discussion_r1679025221.

src/agent/src/image.rs
```diff
@@ -63,6 +63,27 @@ impl ImageService {
...
+            let image_policy_file = &AGENT_CONFIG.image_policy_file;
+            debug!(sl(), "Use imagepolicy file {:?}", image_policy_file);
+            image_client.config.file_paths.policy_path = image_policy_file.clone();
+        }
+
+        #[cfg(feature = "guest-pull")]
+        if !&AGENT_CONFIG.simple_signing_sigstore_config.is_empty() {
```
gkurz, line 81: ditto

src/agent/src/image.rs
wainersm: I thought the same.

src/agent/src/config.rs
```diff
@@ -236,6 +263,17 @@ impl FromStr for AgentConfig {
         );
         config_override!(agent_config_builder, agent_config, guest_components_procs);
         config_override!(agent_config_builder, agent_config, image_registry_auth);
+        config_override!(
```
wainersm, line 266: These `config_override!` should be conditionally compiled with `#[cfg(feature = "guest-pull")]` too, right?

src/agent/src/config.rs
```diff
@@ -358,6 +396,24 @@ impl AgentConfig {
                 config.image_registry_auth,
                 get_string_value
             );
+            parse_cmdline_param!(
```
wainersm, line 399: Same for `parse_cmdline_param!`

src/agent/src/image.rs
```diff
@@ -63,6 +63,27 @@ impl ImageService {
...
+            let simple_signing_sigstore_config = &AGENT_CONFIG.simple_signing_sigstore_config;
+            debug!(sl(), "Using RedHat Simple Signing image signature file {:?}", simple_signing_sigstore_config);
```
wainersm, line 83: Hi @Xynnn007 !

I'm trying to figure out if saying "RedHat Simple Signing" is the most adequate name. Sorry in advantage, this is a topic outside of my scope of expertise...

I don't know the full history (likely @stevenhorsman knows) so I was digging into image-rs docs. I found a reference to https://www.redhat.com/en/blog/container-image-signing who author said the simple signing mechanism would be contributed to [OCI's image library](https://github.com/containers/image) eventually. So not sure if we can claim it OCI compliance instead...

@bpradipt is this signature mechanism used in Openshift? How's it called?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
wainersm: Hi @littlejawa !

Yes, the crio test running on garm is testing only "vanilla" kata. As @stevenhorsman I plan to add a new job to test the CoCo enabled features with cri-o. The kata-deploy part is implemented already, it's really about creating a job and adapting the tests (as you realized, some made assumptions about containerd/nydus that doesn't apply for cri-o). I hope to work on that soon, I let you posted @littlejawa !

src/agent/src/image.rs
Xynnn007: Yes! Let me merge this

src/agent/src/config.rs
Xynnn007: Yes. I made an issue on CoCo side https://github.com/confidential-containers/confidential-containers/issues/221

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,197 @@
...
+@test "Create a pod from an unsigned image, on a 'restricted registry' is rejected" {
+    setup_kbs_image_policy
+
+    create_pod_yaml_with_signed_image "${UNSIGNED_PROTECTED_REGISTRY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    assert_pod_fail "${kata_pod_with_signed_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" "failed"
```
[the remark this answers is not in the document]
Xynnn007: Oh yes! When I composed the code structure I left this as a TODO but forget. Let me fill it up.

src/agent/src/config.rs
Xynnn007: Yea, I think it would be fixed inside #9904 as they are not related to this PR. I will do a rebase after 9904 is merged.

src/agent/README.md
Xynnn007: This is a compatibility issue. For platforms who support initdata we could use CDH config to support this. (see PR https://github.com/confidential-containers/guest-components/pull/608) but those who do not support initdata would still use cmdline to inject this I think. Thus I think we do not need to delete them, but add more comments to tell they are for sev/csv.

btw, in all platforms, probably only sev and csv cannot support initdata. (@huoqifeng once discussed with me about the possibility of s390x to support initdata. Sure, TDX/SNP/CCA/AzVtpm all supports)

any better suggestions?

src/agent/src/image.rs
Xynnn007: Yea, thanks!

src/agent/src/config.rs
Xynnn007: Thanks, fixed

src/agent/src/config.rs
Xynnn007: yes! done

src/agent/src/image.rs
Xynnn007: Oh, yes. Because some documents also say it with "RedHat" prefix, so I added a prefix here. In order to avoid ambiguity, let me delete the RedHat prefix - I believe everyone can understand what SimpleSigning refers to.

src/agent/Makefile
Xynnn007: Oh. This commit has been merged https://github.com/kata-containers/kata-containers/pull/9382/commits/1072658219279c7c55f520384c3cf240bf426d16. The message is actually a typo.

Let me added `RUST_BACKTRACE=1` explicitly and add a new commit for this.

Xynnn007:
@stevenhorsman We need to update the test images on ghcr.io. Actually the current tests of image-rs side succeed because of different reason from what is expected.

src/agent/README.md
huoqifeng: s390x support initdata, I got the approach described in https://github.com/confidential-containers/cloud-api-adaptor/issues/1895 worked in PR: https://github.com/confidential-containers/cloud-api-adaptor/pull/1912 and https://github.com/kata-containers/kata-containers/pull/9988, I'm preparing the PR to validate the initdata via attestation in https://github.com/confidential-containers/guest-components/pull/616 for a end to end use scenario.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,197 @@
...
+@test "Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected" {
+    setup_kbs_image_policy
+
+    create_pod_yaml_with_signed_image "${COSIGNED_SIGNED_PROTECTED_REGISTRY_WRONG_KEY_IMAGE}"
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    assert_pod_fail "${kata_pod_with_signed_image}"
+    assert_logs_contain "${node}" kata "${node_start_time}" "wrong key"
```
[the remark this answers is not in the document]
Xynnn007: I updated the logs that I got from image-rs side. Let's see if it matches what we will get in kata ci

src/agent/src/image.rs
stevenhorsman: Should simple signing be part of this change as I don't think we have the test cases and have another issue to cover this?

stevenhorsman: Saying that it's not really doing much harm, so I'm not against it being it as long as we don't document that we support simple signing without testing. I think we can also do some improvements on the image-rs side for it as well after some support got merged into the `oci-distribution` crate

Xynnn007:
@stevenhorsman @gkurz @littlejawa I finished the rebase & changed the commit message. Please take another look.

src/agent/src/config.rs
```diff
@@ -560,6 +600,8 @@ mod tests {
         assert!(!config.dev_mode);
         assert_eq!(config.log_level, DEFAULT_LOG_LEVEL);
         assert_eq!(config.hotplug_timeout, DEFAULT_HOTPLUG_TIMEOUT);
+        assert!(!config.enable_signature_verification);
+        assert_eq!(config.image_policy_file, "");
```
stevenhorsman, line 604: I think these need a `#[cfg(feature = "guest-pull")]` on them?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,179 @@
...
+    if ! is_confidential_runtime_class; then
+        skip "Test not supported for ${KATA_HYPERVISOR}."
+    fi
+
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
+    kubectl describe pods
+    k8s_delete_all_pods_if_any_exists || true
+
+    if [[ -n "${node_start_time}:-}" && -z "$BATS_TEST_COMPLETED" ]]; then
```
stevenhorsman, line 175: s/"${node_start_time}:-}"/"${node_start_time:-}"/
Sorry - this was a mistake copied from my test

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,179 @@
...
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    k8s_create_pod "${kata_pod_with_signed_image}"
+    echo "Kata pod test-e2e from image security policy is running"
+}
+
+teardown() {
+    # if is_confidential_hardware; then
+    #     skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    # fi
```
stevenhorsman, line 164: We need to decide whether or not to run this on the TEE hardware, so lets run the e2e tests and make a call to delete this, or uncomment

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,179 @@
...
+load "${BATS_TEST_DIRNAME}/confidential_common.sh"
+load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"
+
+export KBS="${KBS:-false}"
+
+setup() {
+    # TODO - do TEEs want to run these tests?
+    # if is_confidential_hardware; then
+    #     skip "Due to issues related to pull-image integration skip tests for ${KATA_HYPERVISOR}."
+    # fi
```
stevenhorsman, line 18: We need to decide whether or not to run this on the TEE hardware, so lets run the e2e tests and make a call to delete this, or uncomment

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
wainersm: The very first features were tested on non-TEE first and after a while we enabled on TEE. Then we kind of changed the approach to run in all platforms (TEE and non-TEE) and IMHO it has been a little bit frustrating because TEE CI still fails a lot. So IMO we should skip TEE testing for a while; let's have the feature merged with qemu-coco-dev.

stevenhorsman:
None of the e2e tests are working in the automation. Debugging locally....

src/agent/README.md
fitzthum: For SEV/CSV maybe we could look into the approach of sending the initdata from the KBS. SEV support in main is quite limited now, though, and I'm not totally sure what the future will be there.

stevenhorsman:
> None of the e2e tests are working in the automation. Debugging locally....

I (and Wainer) have found and I've pushed some fixes, so we'll see if that helps the CI now

stevenhorsman:
Ok, just test 2 failing now:
```
not ok 2 Create a pod from an unsigned image, on a 'restricted registry' is rejected
```
The pod is starting without issue, so either there is a code issues, test problem, or the image is actually signed? ...

gkurz (commented):
> @stevenhorsman @gkurz @littlejawa I finished the rebase & changed the commit message. Please take another look.

All my comments are addressed and the PR definitely looks better after the rebase. 

Thanks Ding !

Now let's debug the remaining issues :smile:

Xynnn007:
> The pod is starting without issue, so either there is a code issues, test problem, or the image is actually signed? ...

The image is not signed. I use my image-rs integration test locally, and the verification result is 
```
Err(Security validate failed: Validate image failed: Cannot pull manifest of ghcr.io/confidential-containers/test-container-image-rs:sha256-28e01ab32c9dbcbaae96cf0d5b472f22e231d9e603811857b295e61197e40a9b.sig: Registry error: url https://ghcr.io/v2/confidential-containers/test-container-image-rs/manifests/sha256-28e01ab32c9dbcbaae96cf0d5b472f22e231d9e603811857b295e61197e40a9b.sig, envelope: OCI API errors: [OCI API error: manifest unknown])
``

The policy
```json
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "ghcr.io/confidential-containers/test-container-image-rs": [
                {
                    "type": "sigstoreSigned",
                    "keyPath": "kbs:///default/cosign-public-key/test"
                }
            ]
        }
    }
}
```

The public key
```
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEwQEjdCiL3ILUf07NDkDVhgKCj1C6
BsCfmM/zt1kNSj0/+nAqA+25XfyClYq2lJFJ6TkgCsf57cTCkXYDz9c+Yg==
-----END PUBLIC KEY-----
```

stevenhorsman:
> > The pod is starting without issue, so either there is a code issues, test problem, or the image is actually signed? ...
> 
> The image is not signed. I use my image-rs integration test locally, and the verification result is

Yeah, I just tried using cosign to get the signature which has the same result:
```
~/go/bin/cosign download signature ghcr.io/confidential-containers/test-container-image-rs:unsigned
Error: no signatures associated with ghcr.io/confidential-containers/test-container-image-rs:unsigned
main.go:74: error during command execution: no signatures associated with ghcr.io/confidential-containers/test-container-image-rs:unsigned
```

I can't see what is failing though, so might need to debug and check the KBS like with auth tests

stevenhorsman:
So in debugging locally it looks like the KBS is called during the failing test:
```
# kubectl logs deployment/kbs -n coco-tenant
Defaulted container "kbs" out of: kbs, copy-config (init)
[2024-07-17T18:41:49Z INFO  kbs] Using config file /etc/kbs/kbs-config.toml
[2024-07-17T18:41:49Z WARN  attestation_service::rvps] No RVPS address provided and will launch a built-in rvps
[2024-07-17T18:41:49Z INFO  attestation_service::token::simple] No Token Signer key in config file, create an ephemeral key and without CA pubkey cert
[2024-07-17T18:41:49Z INFO  kbs] Starting HTTP server at [0.0.0.0:8080]
[2024-07-17T18:41:49Z INFO  actix_server::builder] starting 4 workers
[2024-07-17T18:41:49Z INFO  actix_server::server] Tokio runtime found; starting in existing Tokio runtime
[2024-07-17T18:41:57Z INFO  actix_web::middleware::logger] 10.244.1.8 "GET / HTTP/1.1" 404 0 "-" "Wget" 0.000097
[2024-07-17T18:42:04Z INFO  actix_web::middleware::logger] 10.244.0.0 "HEAD / HTTP/1.1" 404 0 "-" "curl/7.81.0" 0.000054
[2024-07-17T18:42:15Z INFO  actix_web::middleware::logger] 10.244.0.0 "POST /kbs/v0/resource-policy HTTP/1.1" 200 0 "-" "kbs-client/0.1.0" 0.002617
[2024-07-17T18:42:16Z INFO  actix_web::middleware::logger] 10.244.0.0 "POST /kbs/v0/resource/default/security-policy/test HTTP/1.1" 200 0 "-" "kbs-client/0.1.0" 0.002149
[2024-07-17T18:42:16Z INFO  actix_web::middleware::logger] 10.244.0.0 "POST /kbs/v0/resource/default/cosign-public-key/test HTTP/1.1" 200 0 "-" "kbs-client/0.1.0" 0.001907
[2024-07-17T18:42:43Z INFO  kbs::http::attest] Auth API called.
[2024-07-17T18:42:43Z INFO  actix_web::middleware::logger] 10.244.0.0 "POST /kbs/v0/auth HTTP/1.1" 200 74 "-" "attestation-agent-kbs-client/0.1.0" 0.001277
[2024-07-17T18:42:43Z INFO  kbs::http::attest] Attest API called.
[2024-07-17T18:42:43Z INFO  attestation_service] Sample Verifier/endorsement check passed.
[2024-07-17T18:42:43Z INFO  attestation_service] Policy check passed.
[2024-07-17T18:42:43Z INFO  attestation_service] Attestation Token (Simple) generated.
[2024-07-17T18:42:43Z INFO  actix_web::middleware::logger] 10.244.0.0 "POST /kbs/v0/attest HTTP/1.1" 200 2171 "-" "attestation-agent-kbs-client/0.1.0" 0.008951
[2024-07-17T18:42:43Z WARN  kbs::token::coco] No Trusted Certificate in Config, skip verification of JWK cert of Attestation Token
[2024-07-17T18:42:43Z INFO  kbs::http::resource] Get resource from kbs:///default/security-policy/test
[2024-07-17T18:42:43Z INFO  kbs::http::resource] Resource access request passes policy check.
[2024-07-17T18:42:43Z INFO  actix_web::middleware::logger] 10.244.0.0 "GET /kbs/v0/resource/default/security-policy/test HTTP/1.1" 200 1006 "-" "attestation-agent-kbs-client/0.1.0" 0.003083
```
and manually testing the get resources looks okay:
```
# policy=$(kbs-client --url http:192.168.122.26:31144 get-resource --path default/security-policy/test) && echo $policy | base64 -d
[2024-07-17T18:55:44Z WARN  attester] No TEE platform detected. Sample Attester will be used.
[2024-07-17T18:55:44Z WARN  kbs_protocol::client::rcar_client] Authenticating with KBS failed. Perform a new RCAR handshake: ErrorInformation {
        error_type: "https://github.com/confidential-containers/kbs/errors/InvalidRequest",
        detail: "The request is invalid: parse Authorization header failed: invalid Header provided",
    }
[2024-07-17T18:55:44Z WARN  attester] No TEE platform detected. Sample Attester will be used.
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker": {
            "ghcr.io/confidential-containers/test-container-image-rs": [
                {
                    "type": "sigstoreSigned",
                    "keyPath": "kbs:///default/cosign-public-key/test"
                }
            ]
        }
    }
}
```

Xynnn007:
> So in debugging locally it looks like the KBS is called during the failing test:

Is there any logs about `GET /kbs/v0/resource/default/cosign-public-key/test`? There should be one like this

src/agent/src/image.rs
Xynnn007: I deleted simple signing related things in the code. Thus resolve this conversation.

stevenhorsman:
> Is there any logs about GET /kbs/v0/resource/default/cosign-public-key/test? There should be one like this

No - nothing about that and it should request the public key based on the sigstore, so I wonder if there is an issue with the docker registry matching or something? I need to start on the peer pods release updates, but I'll try and play with this in the background to see if I can get anything interesting...

Xynnn007:
> > Is there any logs about GET /kbs/v0/resource/default/cosign-public-key/test? There should be one like this
> 
> No - nothing about that and it should request the public key based on the sigstore, so I wonder if there is an issue with the docker registry matching or something? I need to start on the peer pods release updates, but I'll try and play with this in the background to see if I can get anything interesting...

Oh.. No in hurry as I just found the cosign support is marked as `middle` proirity thus plz take your pace for this..

btw I once had the same doubt as yours thus I debugged inside the code. On image-rs side, the key will be got firstly and then try to fetch signature in the registry. When a use step-by-step debugging, the repo matching worked. The image reference `ghcr.io/confidential-containers/test-container-image-rs:unsigned` matches `ghcr.io/confidential-containers/test-container-image-rs`. But I am not sure whether the log [here](https://github.com/kata-containers/kata-containers/blob/main/src/agent/src/image.rs#L154) is `ghcr.io/confidential-containers/test-container-image-rs:unsigned`.

stevenhorsman:
Ok, I've found the problem, which is a weird one. From the agent's createContainers log:
```
# Jul 17 17:44:53 aks-nodepool1-61081638-vmss000000 kata[48121]: time="2024-07-17T17:44:53.797340622Z" level=debug msg="reading guest console" console-protocol=unix console-url=/run/vc/vm/07163221c0de2fbb140b23f1a8ae915494bbec22e6e37a85aedf886d5cc42916/console.sock name=containerd-shim-v2 pid=48121 sandbox=07163221c0de2fbb140b23f1a8ae915494bbec22e6e37a85aedf886d5cc42916 source=virtcontainers subsystem=sandbox vmconsole="{\"msg\":\"receive createcontainer, storages: [Storage { driver: \\\"image_guest_pull\\\", driver_options: [\\\"image_guest_pull={\\\\\\\"metadata\\\\\\\":{\\\\\\\"io.katacontainers.config.hypervisor.kernel_params\\\\\\\":\\\\\\\"agent.guest_components_rest_api=resource agent.aa_kbc_params=cc_kbc::http://kbs.0f89a9de776d4cf982f7.eastus.aksapp.io:80 agent.enable_signature_verification=true agent.image_policy_file=kbs:///default/security-policy/test\\\\\\\",\\\\\\\"io.katacontainers.pkg.oci.bundle_path\\\\\\\":\\\\\\\"/run/containerd/io.containerd.runtime.v2.task/k8s.io/5f72f8bb027e30be88edc876a90e7240f0b5d458dd4913fe31f040b6de1c6bb6\\\\\\\",\\\\\\\"io.katacontainers.pkg.oci.container_type\\\\\\\":\\\\\\\"pod_container\\\\\\\",\\\\\\\"io.kubernetes.cri.container-name\\\\\\\":\\\\\\\"test-container\\\\\\\",\\\\\\\"io.kubernetes.cri.container-type\\\\\\\":\\\\\\\"container\\\\\\\",\\\\\\\"io.kubernetes.cri.image-name\\\\\\\":\\\\\\\"docker.io/library/busybox:latest\\\\\\\",\\\\\\\"io.kubernetes.cri.sandbox-id\\\\\\\":\\\\\\\"07163221c0de2fbb140b23f1a8ae915494bbec22e6e37a85aedf886d5cc42916\\\\\\\",\\\\\\\"io.kubernetes.cri.sandbox-name\\\\\\\":\\\\\\\"test-e2e\\\\\\\",\\\\\\\"io.kubernetes.cri.sandbox-namespace\\\\\\\":\\\\\\\"kata-containers-k8s-tests\\\\\\\",\\\\\\\"io.kubernetes.cri.sandbox-uid\\\\\\\":\\\\\\\"b1970408-03fb-48cd-a706-4c6d0e76dba7\\\\\\\"}}\\\"], source: \\\"docker.io/library/busybox:latest\\\", fstype: \\\"overlay\\\", options: [], mount_point: \\\"/run/kata-containers/5f72f8bb027e30be88edc876a90e7240f0b5d458dd4913fe31f040b6de1c6bb6/rootfs\\\", fs_group: MessageField(None), special_fields: SpecialFields { unknown_fields: UnknownFields { fields: None }, cached_size: CachedSize { size: 0 } } }]\",\"level\":\"INFO\",\"ts\":\"2024-07-17T17:44:53.732119921Z\",\"pid\":\"124\",\"version\":\"0.1.0\",\"source\":\"agent\",\"subsystem\":\"rpc\",\"name\":\"kata-agent\"}"
```
We can see that nydus is sending the image as `docker.io/library/busybox:latest`, not `ghcr.io/confidential-containers/test-container-image-rs:unsigned`, which is why the security policy isn't used.
@Xynnn007 - I assume that the unsigned image is just the latest busybox image?

We could fix the test by pushing a unique image, but I'm not sure if this has revealed a significant security issue or not in the nydus-snapshotter:
- As a kubeadm I could try and run a malicious pod of a name of my choice and nydus would cache it, then if I can reroute the networking, such that a legit image running in CoCo now points to my malicious image, then in the kata-agent we would get the original path, which might not have signature verification enforced, so we'd just skip that check. I guess this is mitigated by us having the `default` security policy to restricted, but I'm not sure if that would prove an issue with blocking initcontainers/side cars?
@ChengyuZhu6 - do you have an idea of whether we can stop this cached image name from happening in the nydus-snapshotter?

stevenhorsman:
@Xynnn007 - as the nydus-snapshotter issue I raise here hasn't progress any further, do you think that we can change/create a new unsigned image `ghcr.io/confidential-containers/test-container-image-rs:unsigned` that is not based on `busybox` in order to get signed support merged and then work on the nydus-snapshotter problem separately rather than having that issue block this new function?

stevenhorsman:
After we discussed it I also wonder if we are hitting a false positive with the signature verification check. If `ghcr.io/confidential-containers/test-container-image-rs:cosign-signed` is also mapping to `busybox:latest`, then when we pull it we might end up actually sending the `busybox:latest` request through to image-rs, which wouldn't attempt to check the signature validation, so I think we need a unique image there are well (maybe we can just an a new text file saying unsigned/signed on top of busybox?). We can then think about if we want to make the policy more restrictive in the positive test too?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
wainersm: In meanwhile TEE CI got stable so that running on there and non-TEE is fine.

stevenhorsman: Yeah, I've updated the tests to run for all the coco runtime classes now

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,184 @@
...
+
+    # For debug sake
+    echo "Pod ${kata_pod_with_signed_image}: $(cat ${kata_pod_with_signed_image})"
+
+    k8s_create_pod "${kata_pod_with_signed_image}"
+    echo "Kata pod test-e2e from image security policy is running"
+}
+
+teardown() {
+    if [ "${KATA_HYPERVISOR}" = "qemu-tdx" ]; then
```
wainersm, line 167: Hi @stevenhorsman !

[Latest execution on TDX](https://github.com/kata-containers/kata-containers/actions/runs/10305807718/job/28563027264?pr=10009) had signature tests passing, except for the first one:  `not ok 1 Create a pod from an unsigned image, on an insecureAcceptAnything registry works`. But the exit status of k8s-guest-pull-image-signature.bats became zero because of this skip. If we want tests skipped on TDX then it should be skipped on `setup()` too. is it the case?

If you end up updating this commit, can you fix the "verificaiton" typo on the commit msg?

wainersm:
Hi @stevenhorsman !

My comment about the tests on TDX above is the last one that I've for this PR. Everything looks fine except for the bug that you guys found on nydus-snapshotter/containerd (https://github.com/kata-containers/kata-containers/issues/10139) that I don't think should block this PR.

Great debugging & clever tests! Thanks!

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
stevenhorsman: Sorry, I missed removing that. I'll tidy it up now and then can always add TDX skip as a separate commit if needed

beraldoleal (commented):
Except from a nit comment and the TDX errors, LGTM. Thanks @Xynnn007.

wainersm:
@stevenhorsman good and bad news.

The snp job failed but on unrelated test, i.e., signature tests passed. Also in qemu-coco-dev. SEV job skips all attestation related tests.

On TDX (https://github.com/kata-containers/kata-containers/actions/runs/10406882065/job/28827524811?pr=10009) we've got some signature fails:

* `not ok 1 Create a pod from an unsigned image, on an insecureAcceptAnything registry works`
   - due our old and good nydus-snapshotter problem
* `not ok 2 Create a pod from an unsigned image, on a 'restricted registry' is rejected`
   - pod is expected to fail but it ran
* `ok 3 Create a pod from a signed image, on a 'restricted registry' is successful`
* `not ok 4 Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected`
   - same as test 2)
* `ok 5 Create a pod from an unsigned image, on a 'restricted registry' works if enable_signature_verfication is false`

stevenhorsman:
> On TDX (https://github.com/kata-containers/kata-containers/actions/runs/10406882065/job/28827524811?pr=10009) we've got some signature fails:
> 
> * `not ok 1 Create a pod from an unsigned image, on an insecureAcceptAnything registry works`
>   
>   * due our old and good nydus-snapshotter problem
> * `not ok 2 Create a pod from an unsigned image, on a 'restricted registry' is rejected`
>   
>   * pod is expected to fail but it ran
> * `ok 3 Create a pod from a signed image, on a 'restricted registry' is successful`
> * `not ok 4 Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected`
>   
>   * same as test 2)
> * `ok 5 Create a pod from an unsigned image, on a 'restricted registry' works if enable_signature_verfication is false`

So there are 23 failures in that TDX job and looking at the nightly test it looks like the KBS still isn't working properly and returning data. What does concern me is the false positives - tests that pass when they are expected to fail. My guess is that the policy isn't able to be retrieved from the KBS due to whatever KBS issues there are on the node, but I'm not sure if we should fail at that point rather than ignore the policy? @Xynnn007 - I guess this logic is in image-rs/CDH, so you might have some insight here?

src/agent/README.md
```diff
@@ -130,6 +130,8 @@ The kata agent has the ability to configure agent options in guest kernel comman
 | `agent.hotplug_timeout` | Hotplug timeout | Allow to configure hotplug timeout(seconds) of block devices | integer | `3` |
 | `agent.https_proxy` | HTTPS proxy | Allow to configure `https_proxy` in the guest | string | `""` |
 | `agent.image_registry_auth` | Image registry credential URI | The URI to where image-rs can find the credentials for pulling images from private registries e.g. `file:///root/.docker/config.json` to read from a file in the guest image, or `kbs:///default/credentials/test` to get the file from the KBS| string | `""` |
+| `agent.enable_signature_verification` | Image security policy flag | Whether enable image security policy enforcement. If `true`, the resource indexed by URI `agent.image_policy_file` will be got to work as image pulling policy. | string | `""` |
+| `agent.image_policy_file` | Image security policy URI | The URI to where image-rs Typical policy URIs are like `file:///etc/image.json` to read from a file in the guest image, or `kbs:///default/security-policy/test` to get the file from the KBS| string | `""` |
```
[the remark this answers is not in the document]
stevenhorsman: I had to rebase to resolve conflicts, so I've fixed this. Thanks!

wainersm:
> > On TDX (https://github.com/kata-containers/kata-containers/actions/runs/10406882065/job/28827524811?pr=10009) we've got some signature fails:
> > 
> > * `not ok 1 Create a pod from an unsigned image, on an insecureAcceptAnything registry works`
> >   
> >   * due our old and good nydus-snapshotter problem
> > * `not ok 2 Create a pod from an unsigned image, on a 'restricted registry' is rejected`
> >   
> >   * pod is expected to fail but it ran
> > * `ok 3 Create a pod from a signed image, on a 'restricted registry' is successful`
> > * `not ok 4 Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected`
> >   
> >   * same as test 2)
> > * `ok 5 Create a pod from an unsigned image, on a 'restricted registry' works if enable_signature_verfication is false`
> 
> So there are 23 failures in that TDX job and looking at the nightly test it looks like the KBS still isn't working properly and returning data. What does concern me is the false positives - tests that pass when they are expected to fail. My guess is that the policy isn't able to be retrieved from the KBS due to whatever KBS issues there are on the node, but I think we should fail at that point rather than ignore the policy error and proceed without it? @Xynnn007 - I guess this logic is in image-rs/CDH, so you might have some insight here?

I looked at image-rs code, and my theory is:

 * `get_resource()` returns an empty string to `policy_json_string` in https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/signature/mod.rs#L34
 * `is_image_allowed()` don't fail the validation in https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/signature/mod.rs#L44
    *  `is_image_allowed()` calls `requirements_for_image()`.  This last function would return the default policy as there isn't any `transports` in the `Policy` in https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/signature/policy/mod.rs#L102

My theory would make no sense if `let mut policy = serde_json::from_slice::<Policy>(&policy_json_string)?;` returns an error in case of `policy_json_string` being empty in https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/signature/mod.rs#L35C5-L35C77 . As I am not Rust expect, does it return an error?

Xynnn007:
> My theory would make no sense if `let mut policy = serde_json::from_slice::<Policy>(&policy_json_string)?;` returns an error in case of `policy_json_string` being empty in https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/signature/mod.rs#L35C5-L35C77 . As I am not Rust expect, does it return an error?

Yes, it does return an error and I test locally.



> So there are 23 failures in that TDX job and looking at the nightly test it looks like the KBS still isn't working properly and returning data. What does concern me is the false positives - tests that pass when they are expected to fail. My guess is that the policy isn't able to be retrieved from the KBS due to whatever KBS issues there are on the node, but I think we should fail at that point rather than ignore the policy error and proceed without it? @Xynnn007 - I guess this logic is in image-rs/CDH, so you might have some insight here?

Both resource retrievement and policy deserialization would lead to raising an error. -- Could we read the logs of containerd/kata-agent? Probably we could get some more information.

BTW, coco-nontee worked fine https://github.com/kata-containers/kata-containers/actions/runs/10421377360/job/28921746903?pr=10009

src/agent/src/config.rs
```diff
@@ -240,8 +258,17 @@ impl FromStr for AgentConfig {
         );
         config_override!(agent_config_builder, agent_config, guest_components_procs);
         #[cfg(feature = "guest-pull")]
-        config_override!(agent_config_builder, agent_config, image_registry_auth);
+        {
+            config_override!(agent_config_builder, agent_config, image_registry_auth);
+            config_override!(
+                agent_config_builder,
+                agent_config,
+                enable_signature_verification
```
Xynnn007, line 266:
```suggestion
            config_override!(agent_config_builder, agent_config, enable_signature_verification);
            config_override!(
                agent_config_builder,
                agent_config,
                enable_signature_verification
```

Ohhhhhh-My-Goooooooood. TYPT is the thief?

ChengyuZhu6:
> @stevenhorsman good and bad news.
> 
> The snp job failed but on unrelated test, i.e., signature tests passed. Also in qemu-coco-dev. SEV job skips all attestation related tests.
> 
> On TDX (https://github.com/kata-containers/kata-containers/actions/runs/10406882065/job/28827524811?pr=10009) we've got some signature fails:
> 
> * `not ok 1 Create a pod from an unsigned image, on an insecureAcceptAnything registry works`
>   
>   * due our old and good nydus-snapshotter problem
> * `not ok 2 Create a pod from an unsigned image, on a 'restricted registry' is rejected`
>   
>   * pod is expected to fail but it ran
> * `ok 3 Create a pod from a signed image, on a 'restricted registry' is successful`
> * `not ok 4 Create a pod from a signed image, on a 'restricted registry', but with the wrong key is rejected`
>   
>   * same as test 2)
> * `ok 5 Create a pod from an unsigned image, on a 'restricted registry' works if enable_signature_verfication is false`

I think this may be due to some files in the directory `/opt/kata` that haven’t been cleaned up. I deleted the directory and re-triggered the tdx-ci and the tests related to cosign have passed: https://github.com/kata-containers/kata-containers/actions/runs/10421377360/job/28925966522?pr=10009

fidencio:
> I think this may be due to `kata-deploy` have no permission to copy newest files to `/opt/kata` and overwrite the existing files. I deleted the directory and re-triggered the tdx-ci and the tests related to cosign have passed:

This is not impossible, but it's very unlikely, and if that's really the case, our whole CI has a terrible flaw that must be investigated.

wainersm:
@fidencio is going to make a temporary fix on TDX machine. In meanwhile I will run the signature tests a couple of times more to see how stable the feature is.

fidencio:
> @fidencio is going to make a temporary fix on TDX machine. In meanwhile I will run the signature tests a couple of times more to see how stable the feature is.

keep in mind that this may be affecting all the baremetal machines we have.

wainersm:
> > My theory would make no sense if `let mut policy = serde_json::from_slice::<Policy>(&policy_json_string)?;` returns an error in case of `policy_json_string` being empty in https://github.com/confidential-containers/guest-components/blob/main/image-rs/src/signature/mod.rs#L35C5-L35C77 . As I am not Rust expect, does it return an error?
> 
> Yes, it does return an error and I test locally.
> 

Ok...I was so proud of that theory...

wainersm:
Blurb: yesterday I managed to run the signature tests again and they passed on all platforms. I'm running more it today.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,180 @@
...
+            ]
+        }
+    }
+}")
+
+    public_key=$(echo "-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEwQEjdCiL3ILUf07NDkDVhgKCj1C6
+BsCfmM/zt1kNSj0/+nAqA+25XfyClYq2lJFJ6TkgCsf57cTCkXYDz9c+Yg==
+-----END PUBLIC KEY-----")
+
```
gkurz, line 62: Similar cosmetic improvement :
```
public_key=$(cat<<EOF
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEwQEjdCiL3ILUf07NDkDVhgKCj1C6
BsCfmM/zt1kNSj0/+nAqA+25XfyClYq2lJFJ6TkgCsf57cTCkXYDz9c+Yg==
-----END PUBLIC KEY-----
EOF
)
```

but most importantly, where does this public key come from ? How can we ensure it will _work_ ?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,180 @@
...
+    if ! is_confidential_hardware; then
+        kbs_set_allow_all_resources
+    fi
+
+    kbs_set_resource "default" "security-policy" "test" "${policy_json}"
+    kbs_set_resource "default" "cosign-public-key" "test" "${public_key}"
+}
+
+function create_pod_yaml_with_signed_image() {
```
gkurz, line 71: The name is slightly misleading since this function will only enable the signature verification if `${2}` is `true` or empty. And the two first tests don't seem to pass signed images.

Also, this is duplicating a fair amount of code already present in `create_pod_yaml_with_encrypted_image()` and `create_pod_yaml_with_private_image()`. Please consider factoring out some common helper.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,180 @@
...
+            ],
+            \"quay.io/prometheus\": [
+                {
+                    \"type\": \"insecureAcceptAnything\"
+                }
+            ]
+        }
+    }
+}")
+
```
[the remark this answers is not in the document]
Xynnn007: Nice suggestion!

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
Xynnn007: yes, this pubkey comes from https://github.com/confidential-containers/guest-components/blob/main/image-rs/test_data/signature/cosign/cosign1.pub. 

This public key is corresponding to a private key that I generated test signed images myself and we use this for test for a long time in image-rs CI. We'd better to have a CI that generates signed image together with verification. See issue https://github.com/kata-containers/kata-containers/issues/9360

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
Xynnn007: Yup, I added a commit to do the refactoring

Xynnn007:
I added two more commits following @gkurz 's suggestions. If passed I will do a rebase.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
gkurz: > yes, this pubkey comes from https://github.com/confidential-containers/guest-components/blob/main/image-rs/test_data/signature/cosign/cosign1.pub.
> 
> This public key is corresponding to a private key that I generated test signed images myself and we use this for test for a long time in image-rs CI. We'd better to have a CI that generates signed image together with verification. See issue #9360

Ok. Until the CI fully handles the signing, I'd prefer the key to be downloaded from the guest-components repo, along with the explanation you provided here.

wainersm:
CI failed to build a component that I think it is just a noise. So I just re-triggered all jobs again.

Xynnn007:
> @Xynnn007 - do you think we can fix up the new commits you added, so we can get some more of the checks passing? Is there anything else I can help with here that's outstanding?

Yes. Let me do a rebase now

stevenhorsman:
There are lots of test failures, which I guess are as a result of the refactor. Looking into it now...

stevenhorsman:
> There are lots of test failures, which I guess are as a result of the refactor as the load of confidential_kbs.sh seems to fail. Looking into it now...

I think there is a clash of variable names and the `lib_se.sh` ones (which are pulled in via `confidential_kbs.sh`) are readonly. We have a couple of options:
1. Remove the readonly of the variables
2. Namespace them, so they are unique to that file gone with 

I've gone with 2 in https://github.com/kata-containers/kata-containers/pull/10009/commits/758327073025e4f60b785dc189ccfa9edcc37e93. I tried to "dynamically" namespace them by doing `readonly $0_script_dir`, but shellcheck doesn't like indirection like `${$0_script_dir}`, so I've just hard-coded it for now.

stevenhorsman:
> > There are lots of test failures, which I guess are as a result of the refactor as the load of confidential_kbs.sh seems to fail. Looking into it now...
> 
> I think there is a clash of variable names and the `lib_se.sh` ones (which are pulled in via `confidential_kbs.sh`) are readonly. We have a couple of options:
> 
> 1. Remove the readonly of the variables
> 2. Namespace them, so they are unique to that file gone with
> 
> I've gone with 2 in [7d7447d](https://github.com/kata-containers/kata-containers/commit/7d7447ddea77833bded792f04014cae826717084). I tried to "dynamically" namespace them by doing `readonly $0_script_dir`, but shellcheck doesn't like indirection like `${$0_script_dir}`, so I've just hard-coded it for now.

This still failed with:
```
  `load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"' failed with status 0
# /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/../../../tools/packaging/guest-image/lib_se.sh: line 8: lib_se_script_dir: readonly variable
```
So I'll try approach 1 now.

stevenhorsman:
> > > There are lots of test failures, which I guess are as a result of the refactor as the load of confidential_kbs.sh seems to fail. Looking into it now...
> > 
> > 
> > I think there is a clash of variable names and the `lib_se.sh` ones (which are pulled in via `confidential_kbs.sh`) are readonly. We have a couple of options:
> > 
> > 1. Remove the readonly of the variables
> > 2. Namespace them, so they are unique to that file gone with
> > 
> > I've gone with 2 in [7d7447d](https://github.com/kata-containers/kata-containers/commit/7d7447ddea77833bded792f04014cae826717084). I tried to "dynamically" namespace them by doing `readonly $0_script_dir`, but shellcheck doesn't like indirection like `${$0_script_dir}`, so I've just hard-coded it for now.
> 
> This still failed with:
> 
> ```
>   `load "${BATS_TEST_DIRNAME}/confidential_kbs.sh"' failed with status 0
> # /home/runner/work/kata-containers/kata-containers/tests/integration/kubernetes/../../../tools/packaging/guest-image/lib_se.sh: line 8: lib_se_script_dir: readonly variable
> ```
> 
> So I'll try approach 1 now.

Ok, we just hit an issue with another readonly variable, so I'm trying a new approach.
We have a double load of "${BATS_TEST_DIRNAME}/confidential_kbs.sh" now in the confidential-common and the original bats files, so maybe load twice of readonly would mean it tries to overwrite and errors, so I'm removing the old .bats loads

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -106,4 +108,65 @@ function cleanup_loop_device(){
...
+        "${kernel_params_value}"
+
+    add_allow_all_policy_to_yaml "${kata_pod}"
+
+	if [ -n "$node" ]; then
+	    set_node "${kata_pod}" "$node"
+	fi
+
+	return "${kata_pod}"
 }
```
ChengyuZhu6, line 172: newline problem in the end of file?

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,154 @@
...
+    kbs_set_resource "default" "security-policy" "test" "${policy_json}"
+    kbs_set_resource "default" "cosign-public-key" "test" "${public_key}"
+}
+
+@test "Create a pod from an unsigned image, on an insecureAcceptAnything registry works" {
+    # We want to set the default policy to be reject to rule out false positives
+    setup_kbs_image_policy "reject"
+
+    create_coco_pod_yaml "${UNSIGNED_UNPROTECTED_REGISTRY_IMAGE}" "kbs:///default/security-policy/test" "" "" "resource" "$node"
```
ChengyuZhu6, line 79: Would it be better to set a variable like `IMAGE_POLICY_URL` to replace the value "kbs:///default/security-policy/test" in the tests?

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -106,4 +108,65 @@ function cleanup_loop_device(){
...
+	image_registry_auth=${3:-}
+	guest_components_procs=${4:-}
+	guest_components_rest_api=${5:-}
+	node=${6:-}
+
+	local CC_KBS_ADDR
+    export CC_KBS_ADDR=$(kbs_k8s_svc_http_addr)
+
+    kernel_params_annotation="io.katacontainers.config.hypervisor.kernel_params"
+	kernel_params_value=""
```
ChengyuZhu6, line 132: minor comment: inconsistent whitespace indentation?

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -106,4 +108,65 @@ function cleanup_loop_device(){
...
+
+	if [ -n "$guest_components_rest_api" ]; then
+		kernel_params_value+=" agent.guest_components_rest_api=${guest_components_rest_api}"
+	fi
+
+	kernel_params_value+=" agent.aa_kbc_params=cc_kbc::${CC_KBS_ADDR}"
+
+    # Note: this is not local as we use it in the caller test
+    kata_pod="$(new_pod_config "$image" "kata-${KATA_HYPERVISOR}")"
+    set_container_command "${kata_pod}" "0" "sleep" "30"
```
ChengyuZhu6, line 155: Same comment as mentioned above.

tests/integration/kubernetes/confidential_common.sh
```diff
@@ -106,4 +108,65 @@ function cleanup_loop_device(){
...
+        "${kernel_params_annotation}" \
+        "${kernel_params_value}"
+
+    add_allow_all_policy_to_yaml "${kata_pod}"
+
+	if [ -n "$node" ]; then
+	    set_node "${kata_pod}" "$node"
+	fi
+
+	return "${kata_pod}"
```
ChengyuZhu6, line 171: Same comment as mentioned above.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
```diff
@@ -0,0 +1,154 @@
...
+
+    [ "${SNAPSHOTTER:-}" = "nydus" ] || skip "None snapshotter was found but this test requires one"
+
+    kubectl describe pods
+    k8s_delete_all_pods_if_any_exists || true
+
+    if [[ -n "${node_start_time:-}" && -z "$BATS_TEST_COMPLETED" ]]; then
+		echo "DEBUG: system logs of node '$node' since test start time ($node_start_time)"
+		print_node_journal "$node" "kata" --since "$node_start_time" || true
+	fi
```
ChengyuZhu6, line 153: Same comment as mentioned above.

ChengyuZhu6 (commented):
Hey @stevenhorsman, thanks for your work on it. I have a few minor comments as follows.

tests/integration/kubernetes/confidential_common.sh
stevenhorsman: Added new line

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
stevenhorsman: Done, thanks for the suggestion

tests/integration/kubernetes/confidential_common.sh
stevenhorsman: Should be fixed now

tests/integration/kubernetes/confidential_common.sh
stevenhorsman: Should be fixed now

tests/integration/kubernetes/confidential_common.sh
stevenhorsman: Should be fixed now

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
stevenhorsman: Should be fixed now

stevenhorsman:
Rebased to pick up https://github.com/kata-containers/kata-containers/pull/10204 and hopefully that will stop the kata-deploy timeout failures on the CI

stevenhorsman:
@gkurz @wainersm - the issues after the refactor commit have been fixed now and the CI is passing, so if you are able to re-review now that would be great. Thanks!

wainersm:
Hi @stevenhorsman @Xynnn007 !

> @gkurz @wainersm - the issues after the refactor commit have been fixed now and the CI is passing, so if you are able to re-review now that would be great. Thanks!

Sorry for abandoning this PR...

The refactor commit is very good! I had that aim to remove the same boilerplates, so thanks!

wainersm (approved):
Time to land this one too! Thanks @Xynnn007 @stevenhorsman !

ChengyuZhu6 (approved):
Thanks for your work @Xynnn007 @stevenhorsman!  While we may need to rebase the PR.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
littlejawa: I think this comment was not addressed ?
Not a big deal, but I'm pretty sure that in a couple months, nobody will remember where this key comes from :) Having it downloaded makes it clear what its origin is, and prevents problems if the key and test images are changed on the other side.

littlejawa:
I think this is ready to go. Just need to address one open comment.
Thanks for your hard work !

gkurz:
> I think this is ready to go. Just need to address one open comment. Thanks for your hard work !

Since rebase is needed anyway because of conflicts with main, I'd really like https://github.com/kata-containers/kata-containers/pull/10009#discussion_r1756317770 to be addressed.

tests/integration/kubernetes/k8s-guest-pull-image-signature.bats
stevenhorsman: An attempt was made - lets see if it works!

gkurz (approved):
Thanks @Xynnn007 and @stevenhorsman !

littlejawa (approved):
LGTM
Thanks!

stevenhorsman:
Just rebased to pick up the latest changes as some policy tests were failing which I think were merged overnight

PR #10016, opened 2024-07-16
workflows: setup environment to run auth registry tests on s390x

wainersm:
We missed to apply the changes of commit d8961cbd4a in the IBM SE workflow and then getting https://github.com/kata-containers/kata-containers/pull/9904 not passing CI. As this is a required change on the workflow we need to have it in a separated PR.

While here, I delinted the run-k8s-tests-on-zvsi.yaml and ran `actionlint` to ensure I'm not introducing any (obvious) bug.


beraldoleal (approved):
LGTM, thanks

fitzthum (approved):
LGTM

BbolroC (approved):
LGTM, thanks for taking care of this @wainersm !

stevenhorsman:
This is a workflow only change that isn't tested by the PR checks as it is not triggered with a on/pull_request. It also has at least two reviews with the force-skip-ci label, therefore we can skip the CI and use admin rights to merge this.

PR #10017, opened 2024-07-16
runtime: Files are not synced between host and guest VMs

squarti:
See https://github.com/kata-containers/kata-containers/issues/9986

### DIFF
--- a/src/runtime/virtcontainers/fs_share_linux.go
+++ b/src/runtime/virtcontainers/fs_share_linux.go
@@ -42,7 +42,7 @@ import (
 // /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~secret
 // /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~projected
 // /var/lib/kubelet/pods/f51ae853-557e-4ce1-b60b-a1101b555612/volumes/kubernetes.io~downward-api
-var configVolRegexString = "^/var/lib/kubelet/pods/[a-fA-F0-9\\-]{36}/volumes/kubernetes\\.io~(configmap|secret|projected|downward-api)"
+var configVolRegexString = ".*/pods/[a-fA-F0-9\\-]{36}/volumes/kubernetes\\.io~(configmap|secret|projected|downward-api)"
 var configVolRegex = regexp.MustCompile(configVolRegexString)
 
 // timestampDirRegex: Regex to match only the timestamped directory inside the above volume mount


sprt:
Cc @Sumynwa who wrote the original code.

danmihai1:
When we start sync-ing additional files from the Host to the Guest we need to also consider the possible impact to the confidentiality of the Guest for Confidential Containers.

fidencio:
@squarti, I'd love to see a better explanation of what errors you found, which kubernetes flavour you were using, and why this patch is needed, as the commit message is basically empty right now.

fidencio:
> @squarti, I'd love to see a better explanation of what errors you found, which kubernetes flavour you were using, and why this patch is needed, as the commit message is basically empty right now.

Okay, I read the description of the issue and that is very well describe and some of those bits should be also part of the commit message.

Now, the problem is basically that Kata Containers doesn't know about kubelet `--root-dir` value in order to properly set the regex, and that can really be anything the admin decided to set.

fidencio:
> > @squarti, I'd love to see a better explanation of what errors you found, which kubernetes flavour you were using, and why this patch is needed, as the commit message is basically empty right now.
> 
> Okay, I read the description of the issue and that is very well described and some of those bits should be also part of the commit message.
> 
> Now, the problem is basically that Kata Containers doesn't know about kubelet `--root-dir` value in order to properly set the regex, and that can really be anything the admin decided to set.

I am wondering here whether the runtime, somehow, could be able to check kubelet's root dir, and then adapt the regex "on the fly", rather than having this wide-open.  @kad, is this information something we could get? Or is this a KEP opportunity? Also, what would be your suggestion on how to approach this?

fidencio:
I've been going back and forth on this since Yesterday, and I think the best option we have for now is:
* Add a runtime configuration option for kubernetes_root_dir
  * Do **NOT** make it configurable via annotations
* Make it configurable at deployment time (via kata-deploy)  
  * Later on this can be expanded to the Confidential Containers operator
* Take kubernetes_root_dir into account when creating the reg_ex

@danmihai1, @sprt, @Sumynwa, @squarti, @kad, @stevenhorsman, @bpradipt, @mkulke, @burgerdev, @Redent0r, WDYT?

bpradipt:
> Add a runtime configuration option for kubernetes_root_dir
> * Do **NOT** make it configurable via annotations

@fidencio are you referring to adding `kubernetes_root_dir` as a config parameter in configuration toml  ?

fidencio:
> @fidencio are you referring to adding `kubernetes_root_dir` as a config parameter in configuration toml ?

Exactly that, as it seems we cannot infer that from anywhere at this point.

zvonkok:
I am just thinking how an application in the VM would know that `/var/lib/kubelet/pod` is not available at the default place. Rather than just bindmount the src to the same dst, shouldn't we try to map the non-default `kubernetes_root_dir` inside the VM to the default path? 
Some prom-exporters rely on `/var/lib/kubelet/pod-resources/kubelet.sock`; if I start a service in the VM, how would the exporter know that it is not there but rather in some non-obvious place?

mkulke:
> I've been going back and forth on this since Yesterday, and I think the best option we have for now is:
> 
> * Add a runtime configuration option for kubernetes_root_dir
>   
>   * Do **NOT** make it configurable via annotations
> * Make it configurable at deployment time (via kata-deploy)
>   
>   * Later on this can be expanded to the Confidential Containers operator
> * Take kubernetes_root_dir into account when creating the reg_ex
> 
> @danmihai1, @sprt, @Sumynwa, @squarti, @kad, @stevenhorsman, @bpradipt, @mkulke, @burgerdev, @Redent0r, WDYT?

a value at kata deployment time looks like the best option to me. tweaking this per pod is no valid use case afaiu.

burgerdev:
Is there a reason why the regex needs to be anchored? The directory pattern below `/var/lib/kubelet/pods` (what would be matched after @squarti's change) is pretty unique. 

With respect to the confidentiality impact, I don't think there is much we can do to secure mounts from the untrusted side, other than restricting their impact to their mount point. At least as long as dynamic updates are desired.

I fear that managing virtual paths that somehow map to host paths is a slippery slope, thinking of `/var/lib/kubelet`, `/var/run/kubernetes`, `/usr/libexec/kubernetes`, `/run/containerd` etc. At the end of the day, things that need resources from the host are node services, and need to be configured as such.

squarti:
> I am just thinking how an application in the VM would know that /var/lib/kubelet/pod is not available at the default place.

In our case, the `/var/lib/kubelet/` directory is a link to the kubelet root dir.

```
# ls -la /var/lib/kubelet
lrwxrwxrwx 1 root root 17 Jun 14 15:39 /var/lib/kubelet -> /var/data/kubelet
```

burgerdev:
> Is there a reason why the regex needs to be anchored? The directory pattern below `/var/lib/kubelet/pods` (what would be matched after @squarti's change) is pretty unique.

@fidencio: do you know why this needs to be an absolute path?

fidencio:
> @fidencio: do you know why this needs to be an absolute path?

Replying from the top of my head, and I didn't write this code, so take this with a grain of salt.

Some of the functions used (Walk / WalkDir) for sure do not work with symlinks.  But in order to work this around we could simply resolve the symlink / get the absolute path first, and do everything needed using the absolute path.

Still, adding support for kubernetes root-dir is the most future-proof thing we could do.

squarti:
@fidencio I submitted another [PR](https://github.com/kata-containers/kata-containers/pull/10151) which resolves the default kubelet root dir and uses it to initialize the RegEXs.  This is a step towards adding the the configuration property.

fidencio:
> @fidencio I submitted another [PR](https://github.com/kata-containers/kata-containers/pull/10151) which resolves the default kubelet root dir symlink (if one exists) and uses it to initialize the RegEXs. This is a step towards adding the the configuration property.

Amazing.  Do you think we can close this one for now?

squarti:
> > @fidencio I submitted another [PR](https://github.com/kata-containers/kata-containers/pull/10151) which resolves the default kubelet root dir symlink (if one exists) and uses it to initialize the RegEXs. This is a step towards adding the the configuration property.
> 
> Amazing. Do you think we can close this one for now?

Yes, configuration property can be added later.

Issue #10023, opened 2024-07-17
Dragonball support initrd


lifupan:
we need to let dragonball hypervisor support initrd.

PR #10024, opened 2024-07-17
runtime-rs: enable dragonball hypervisor support initrd

lifupan:
enable the dragonball support initrd.

Fixes: #10023

### DIFF
--- a/src/libs/kata-types/src/config/hypervisor/dragonball.rs
+++ b/src/libs/kata-types/src/config/hypervisor/dragonball.rs
@@ -122,14 +122,13 @@ impl ConfigPlugin for DragonballConfig {
                     "Guest kernel image for dragonball hypervisor is empty"
                 ));
             }
-            if db.boot_info.image.is_empty() {
+
+            if db.boot_info.image.is_empty() && db.boot_info.initrd.is_empty() {
                 return Err(eother!(
-                    "Guest boot image for dragonball hypervisor is empty"
+                    "Both of guest boot image and initrd for dragonball hypervisor is empty"
                 ));
             }
-            if !db.boot_info.initrd.is_empty() {
-                return Err(eother!("Initrd for dragonball hypervisor should be empty"));
-            }
+
             if !db.boot_info.firmware.is_empty() {
                 return Err(eother!(
                     "Firmware for dragonball hypervisor should be empty"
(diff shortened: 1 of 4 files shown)


src/libs/kata-types/src/config/hypervisor/dragonball.rs
```diff
@@ -122,14 +122,13 @@ impl ConfigPlugin for DragonballConfig {
                     "Guest kernel image for dragonball hypervisor is empty"
                 ));
             }
-            if db.boot_info.image.is_empty() {
+
+            if db.boot_info.image.is_empty() && db.boot_info.initrd.is_empty() {
                 return Err(eother!(
-                    "Guest boot image for dragonball hypervisor is empty"
+                    "Both of guest boot image and Initrd for dragonball hypervisor is empty"
```
zvonkok, line 128:
```suggestion
                    "Both of guest boot image and initrd for dragonball hypervisor is empty"
```

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
```diff
@@ -242,28 +244,23 @@ impl VirtSandbox {
...
-            let image_path = boot_info.image;
-            if !initrd_path.is_empty() {
-                Ok(initrd_path)
-            } else if !image_path.is_empty() {
-                Ok(image_path)
-            } else {
-                Err(anyhow!("failed to get image"))
-            }
+        if !boot_info.initrd.is_empty() {
+            return Ok(None);
```
pmores, line 251: This looks like hypervisor drivers will stop being called to add a block device whose `path_on_host` points in fact to an initrd, not image.  If so then that's great!

However, are there any other such consequences that could break a hypervisor support?  This is common code which affects everyone.

src/runtime-rs/crates/hypervisor/src/dragonball/inner.rs
```diff
@@ -143,8 +148,11 @@ impl DragonballInner {
 
         // set boot source
         let kernel_path = self.config.boot_info.kernel.clone();
+        let initrd_path = self.config.boot_info.initrd.clone();
```
Apokleos, line 151: Nit: maybe using `&self.config.boot_info.initrd` is better than `self.config.boot_info.initrd.clone()`, but the L150 is indeed using `.clone()`.

Apokleos (approved):
Thx @lifupan LGTM

zvonkok (approved):
LGTM, thanks @lifupan

zvonkok:
I've tested this with my GPU setup and I can start the `nvidia-gpu-initrd` on dragonball. One step closer to get better support on DB!

src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
pmores: Yeah I had to add the check and will be happy to get rid of it again since it's somewhat ugly. :-)

pmores (approved):
lgtm, thanks a lot @lifupan !

PR #10034, opened 2024-07-18
tests: Call repack_secure_image() in set_metadata_annotation()

BbolroC:
It is not good practice to call repack_secure_image() from a bats file because the test code might not consider cases where `qemu-se` is used as `KATA_HYPERVISOR`.

This PR moves the function call to set_metadata_annotation() if a key includes `kernel_params` and `KATA_HYPERVISOR` is set to `qemu-se`, allowing developers to focus on the test scenario itself.

### DIFF
--- a/tests/integration/kubernetes/lib.sh
+++ b/tests/integration/kubernetes/lib.sh
@@ -191,6 +191,15 @@ set_metadata_annotation() {
 	# yq set annotations in yaml. Quoting the key because it can have
 	# dots.
 	yq -i ".${annotation_key} = \"${value}\"" "${yaml}"
+
+	if [[ "${key}" =~ kernel_params ]] && [[ "${KATA_HYPERVISOR}" == "qemu-se" ]]; then
+		# A secure boot image for IBM SE should be rebuilt according to the KBS configuration.
+		if [ -z "${IBM_SE_CREDS_DIR:-}" ]; then
+			>&2 echo "ERROR: IBM_SE_CREDS_DIR is empty"
+			return 1
+		fi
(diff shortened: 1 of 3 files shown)


BbolroC:
For reviewers, I got all tests to pass internally:

```
[run_kubernetes_tests.sh:131] INFO: All tests SUCCEEDED
```

stevenhorsman (approved):
LGTM

BbolroC:
Please don't merge this until @gkurz is okay with this.

tests/integration/kubernetes/lib.sh
```diff
@@ -191,6 +191,15 @@ set_metadata_annotation() {
 	# yq set annotations in yaml. Quoting the key because it can have
 	# dots.
 	yq -i ".${annotation_key} = \"${value}\"" "${yaml}"
+
+	if [[ "${key}" =~ kernel_params ]] && [[ "${KATA_HYPERVISOR}" == "qemu-se" ]]; then
```
gkurz, line 195: IIUC the first check will have the test to use some pre-existing image instead of repacking one. If this is the expected behavior, it is an definitely an improvement :smiley:

gkurz (approved):
/lgtm

Thanks Choi !

@GabyCT can you help with the failing metrics checks ?

tests/integration/kubernetes/lib.sh
BbolroC: Yeah, we have an image `kata-containers-se.img` deployed in a designated place (i.e. `/opt/kata/share/kata-containers`) It is used unless a kernel parameter is changed. Thanks!

Issue #10033, opened 2024-07-18
genpolicy: should support additional pod configuration files as command line arguments


danmihai1:
The `genpolicy` command line supports config map files using "-c". Other types of config files should be supported too. For example, the `k8s-credentials-secrets.bats` CI test needs to pass `inject_secret.yaml` as a configuration file for `pod-secret-env.yaml`. 

The solution to this would probably look like: https://github.com/microsoft/kata-containers/commit/8681920da2e7fabb6faaea38fefbef4b4db717ad

PR #10035, opened 2024-07-18
tests: k8s-credentials-secrets: policy for second pod

danmihai1:
Add policy to pod-secret-env.yaml from k8s-credentials-secrets.bats.

Policy was already auto-generated for the other pod used by the same test (pod-secret.yaml). pod-secret-env.yaml was inconsistent, because it was taking advantage of the "allow all" policy built into the Guest image. Sooner or later, CI Guests for CoCo will not get the "allow all" policy built in anymore and pod-secret-env.yaml would have stopped working then.

Note that pod-secret-env.yaml continues to use an "allow all" policy after these changes. #10033 must be solved before a more restrictive policy will be generated for pod-secret-env.yaml.


GabyCT (approved):

wainersm (approved):
Thanks @danmihai1 !

Issue #9594, opened 2024-05-03
ci: reinstate Mariner host


sprt:
Related: #9593

PR #10037, opened 2024-07-18
ci: reinstate Mariner host and guest kernel

sprt:
GH-9592 addressed a bug in a previous version of the AKS Mariner host kernel that blocked the CH v39 upgrade. This bug has now been fixed so we undo that PR.

Fixes: #9594


stevenhorsman (approved):
LGTM. Thanks

danmihai1 (approved):

PR #10007, opened 2024-07-12
ci: Transition GARM tests to free runners, pt. II

wainersm:
This is a continuation of @sprt 's efforts to migrate our workflows from GARM to free runners. His https://github.com/kata-containers/kata-containers/issues/9940 tracks the migration of non-k8s workflows, which he has either disabled or already migrated. This PR, however, is focused on migrating all the k8s workflows to free runners because 1) it will save money from our azure subscription likewise; 2) GARM is broken and it might block our 3.7.0 release

You will noticed that I started just migrating the workflows but then I decided to merge two of them (merge run-k8s-tests-on-garm and run-k8s-tests-with-crio-on-garm) which will force re-triggering of CI on existing PRs I guess. Well, I had it broke already then I decided to go further and renamed the jobs :)

>Note: This will reduce the number of jobs executed in half - 1. With GARM we have two VM profiles ("small" and "normal") which run subsets of k8s tests each (see https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/run_kubernetes_tests.sh#L27 and   https://github.com/kata-containers/kata-containers/blob/main/tests/integration/kubernetes/run_kubernetes_tests.sh#L77). I removed the `instance-type` parameter from the matrix and set `K8S_TEST_HOST_TYPE: all` which will make it to run both "small" and "normal" tests on a single job (hence reducing the amount of jobs triggered).

@fidencio ^^ please let me know if I am missing something and I cannot merge "small" and "normal" jobs.

Ok, tested on my fork of kata-containers:
 * Before the merge of merge run-k8s-tests-on-garm and run-k8s-tests-with-crio-on-garm: https://github.com/wainersm/kata-containers/actions/runs/9912848706
 * After the merge and rename: https://github.com/wainersm/kata-containers/actions/runs/9913412674

PR #10038, opened 2024-07-18
ci: Transition GARM tests to free runners, pt. III

sprt:
Follow up to #9960 and #10007.

Tracked in #9940.

### DIFF
--- a/.github/workflows/PR-wip-checks.yaml
+++ b/.github/workflows/PR-wip-checks.yaml
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   pr_wip_check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: WIP Check
     steps:
     - name: WIP Check
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -13,7 +13,7 @@ concurrency:
 name: Static checks
 jobs:
   check-kernel-config-version:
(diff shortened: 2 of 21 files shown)


.github/workflows/static-checks.yaml
```diff
@@ -66,10 +64,6 @@ jobs:
           sudo -E ./ci/install_yq.sh
         env:
           INSTALL_IN_GOPATH: false
-      - name: Install rust
```
wainersm, line 69: Although rust is preinstalled in the free runner, I think we should stick with version in https://github.com/kata-containers/kata-containers/blob/main/versions.yaml#L374

.github/workflows/PR-wip-checks.yaml
```diff
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   pr_wip_check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
wainersm, line 18: Gods of versions pinned will love it. thanks @sprt !

.github/workflows/static-checks.yaml
sprt: Agreed, will do this.

sprt: Done

stevenhorsman (approved):
LGTM. Thanks @sprt

GabyCT (approved):

PR #10043, opened 2024-07-19
runtime-rs : fix the issue of stop sandbox

lifupan:
Since stop sandbox would be called in multi path,
thus it's better to set and check the sandbox's state.

### DIFF
--- a/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/container_manager/container_inner.rs
@@ -275,6 +275,10 @@ impl ContainerInner {
         signal: u32,
         all: bool,
     ) -> Result<()> {
+        if self.check_state(vec![ProcessStatus::Stopped]).await.is_ok() {
+            return Ok(());
+        }
+
         let mut process_id: agent::ContainerProcessID = process.clone().into();
         if all {
             // force signal init process
--- a/src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
+++ b/src/runtime-rs/crates/runtimes/virt_container/src/sandbox.rs
@@ -471,8 +471,15 @@ impl Sandbox for VirtSandbox {
     }
 
     async fn stop(&self) -> Result<()> {
-        info!(sl!(), "begin stop sandbox");
-        self.hypervisor.stop_vm().await.context("stop vm")?;
+        let mut sandbox_inner = self.inner.write().await;
+
+        if sandbox_inner.state != SandboxState::Stopped {
+            info!(sl!(), "begin stop sandbox");
+            self.hypervisor.stop_vm().await.context("stop vm")?;
+            sandbox_inner.state = SandboxState::Stopped;
+            info!(sl!(), "sandbox stopped");
+        }
+
         Ok(())
     }
 


pmores:
This is a good change that definitely won't hurt.  However the nagging question in my mind lately has been, are we stopping the sandbox at the right time in the first place?

The typical shutdown sequence of shimv2 events - at least according to my observation - is along the lines of

|#| Request | Response |
|---|--------------|----------------|
| 1 | StateProcess | indicates a running process (`status: Running` etc.) |
| 2 | KillProcess | |
| 3 | WaitProcess | contains `ProcessExitStatus` indicating that process exited |
| 4 | StateProcess | now indicates a terminated process (`status: Stopped`, contains exit code and time etc.) |
| 5 | KillProcess | |
| 6 | ConnectContainer | |
| 7 | WaitProcess | returns immediately, response indicates exited process |
| 8 | StateProcess | same reply as in step 4 above |
| 9 | DeleteProcess | same reply as in previous step |
| 10 | ShutdownContainer | |

Now mind you I haven't been able to obtain a specification of the protocol so I can only guess what the requests mean and what is expected of their handling.  But anyway, step 1-4 make sense to me - remote peer makes sure the process is running, kills it and verifies that the process has been killed.

I have no explanation for steps 5-8.  Why is `KillProcess` executed again immediately after it was verified that the process is gone?  Why is then the `ConnectContainer`, `WaitProcess` sequence executed which seems more typical for container _launch_?  Why is then `StateProcess` re-executed and why does its reply now seem enough to move into the final `DeleteProcess`/`ShutdownContainer` phase even though it's precisely the same reply as in step 4?

I don't know the answers to any of these questions but so much seems clear that `WaitProcess` appears flaky and prone to spurious invocations which is the cause of repeated attempts to stop sandbox.  Wouldn't it be better to move sandbox stopping to `DeleteProcess` or even `ShutdownProcess` handler then?

lifupan:
Hi @pmores 

For the twice wait call, the first one is called before starting the container, and the second wait is called in the containerd's task exit event monitor, which would receive a task exit event once the container terminated, and then it would send a second kill and wait call. That's why you would see the second kill and wait even after you terminated the container. 

Second, based on our experience, it's better to shutdown the sandbox/hypervisor/vmm as soon as possible. Since the hypervisor would take some host resources such as the vfio devices, and which would be reclaimed by some cri plugins such as cni plugin once the pause container terminated. If the hypervisor/vmm process wasn't shutdown, then it would hold the vfio resources and the cni plugin couldn't reclaim it successfully and it would even worse trigger dead lock here. Thus it's better to shutdown the sandbox as soon as possible.

That's why we call the sandbox stop/shutdown in the sandbox's wait process.

pmores:
> For the twice wait call, the first one is called before starting the container, and the second wait is called in the containerd's task exit event monitor, which would receive a task exit event once the container terminated, and then it would send a second kill and wait call. That's why you would see the second kill and wait even after you terminated the container.

Hm, do you have any idea why containerd would do that?  IIUC return of the wait process is itself an indication that container exited - why handle it by killing the container a second time?

> Second, based on our experience, it's better to shutdown the sandbox/hypervisor/vmm as soon as possible. Since the hypervisor would take some host resources such as the vfio devices, and which would be reclaimed by some cri plugins such as cni plugin once the pause container terminated. If the hypervisor/vmm process wasn't shutdown, then it would hold the vfio resources and the cni plugin couldn't reclaim it successfully and it would even worse trigger dead lock here. 

I'm not sure I understand.  So if we moved sandbox stopping from `WaitProcess` to `DeleteProcess` (or `ShutdownContainer`) we'd be risking a deadlock?  Can you tell me more about the deadlock, above all between which threads of execution it would happen?  Is the device reclaiming by CNI happening after `WaitProcess` but before `DeleteProcess`?

pmores (approved):
@lifupan Thanks for the great explanation! I can see that we don't seem to have a good case at this time for changing at which point we stop the sandbox.

Apokleos (approved):
Thx @lifupan LGTM!

Issue #10044, opened 2024-07-19
cleanup the container when create container failed


lifupan:
It should cleanup the container when create failed. Otherwise, there would be left some resources such as
devices hotplugged and mounts. All of those resouces and mounts should clean up.

PR #10045, opened 2024-07-19
runtime-rs: container: fix the issue of missing cleanup container

lifupan:
When create container failed, it should cleanup the container thus there's no device/resource left.

Fixes: #10044


pmores (approved):
Looks good to me, thanks @lifupan !

gkurz (approved):
/lgtm

Thanks Fupan !

PR #10049, opened 2024-07-19
[RFC] NUMA topology

koct9i:
- qemu: enable NUMA support
  Link qemu with libnuma and enable NUMA feature flag.
  
- kernel: enable x86-64 ACPI NUMA detection
  CONFIG_NUMA is already on,
  but without CONFIG_X86_64_ACPI_NUMA kernel cannot detect NUMA.
  
- runtime: add hypervisor options for NUMA topology
  With enable_numa=true hypervisor will expose host NUMA topology as is:
  map vm NUMA nodes to host 1:1 and bind vpus to relates CPUS.
  
  Option "numa_mapping" allows to redefine NUMA nodes mapping:
  - map each vm node to particular host node or several numa nodes
  - emulate numa on host without numa (useful for tests)
  
- runtime: enforce NUMA topology by VCPU threads affinity
  For optimal performance VCPU threads must utilize only NUMA local CPUs.
  
- govmm: setup qemu VM NUMA topology for initial CPUs and memory
  If NUMA topology is enabled:
  - groups CPUs into per NUMA node sockets
  - split memory into per NUMA node modules
  - report NUMA node for VCPU threads
  
- runtime: add annotation default_maxmemory
  It seems there is no annotation for disabling VM memory hotplug.
  This should be useful at least until hotplug become fully NUMA-aware.
  

zvonkok:
Is there a limit imposed by QEMU (per ARCH) on how many NUMA nodes we can create in the VM? 

QEMU has various limitations, we should add a check to not start a VM with an invalid NUMA count.

zvonkok:
I wonder if we need to add HMAT flags as well ... Nonetheless if we can replicate the NUMA host topology 1:1 in the guest than we should be also good to attach VFIOs with NUMA  information.

koct9i:
Yep. There're lot possible enhancements.
Passing pieces of topology as-is shouldn't be hard as far hypervisor supports it close enough.

For now, next major step is handling memory/cpu hotplug somehow NUMA-aware.
It seems better to support virtio-mem and don't touch any legacy interfaces.

zvonkok:
@pmores How hard would it be to support this in `qemu-rs` and `runtime-rs`?

zvonkok:
The NUMA command line should work 1:1 on Arm as well. Need to test this.

zvonkok:
@koct9i static checks are failing, please take a look.

pmores:
@zvonkok I'm not familiar with NUMA support in QEMU unfortunately, also stuff like pinning is not supported in qemu-rs (yet).  That said, it should always be possible to basically just port this PR to Rust, i.e. use it to guide the Rust implementation.

zvonkok:
@koct9i  Are you still working on this?

koct9i:
> @koct9i Are you still working on this?

Right now this part is on hold.
Feel free to adapt and rework any these changes.

zvonkok:
Superseeded by: https://github.com/kata-containers/kata-containers/pull/10884

Issue #10014, opened 2024-07-15
tests: genpolicy: use `exec_command` both for creating Policy and for executing the commmand


danmihai1:
@wainersm pointed out in https://github.com/kata-containers/kata-containers/pull/10008#discussion_r1678112054 that the genpolicy tests using an `exec_command` variable to allow the execution of that command should reuse the same `exec_command` variable for the actual execution of the command. We should clean-up all of these tests - e.g., 

```
k8s-attach-handlers.bats:       exec_command="sh -c ${display_message}"
k8s-attach-handlers.bats:       add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command}"
k8s-caps.bats:  exec_command="sh -c ${command}"
k8s-caps.bats:  add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command}"
k8s-configmap.bats:     exec_command="sh -c ${cmd}"
k8s-configmap.bats:     add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command}"
```

PR #10051, opened 2024-07-19
tests: k8s: reuse policy exec variable

danmihai1:
Share a single test script variable for both:
- Allowing a command to be executed using Policy settings.
- Executing that command using "kubectl exec".

Fixes: #10014

### DIFF
--- a/tests/integration/kubernetes/k8s-attach-handlers.bats
+++ b/tests/integration/kubernetes/k8s-attach-handlers.bats
@@ -25,8 +25,8 @@ setup() {
 	policy_settings_dir="$(create_tmp_policy_settings_dir "${pod_config_dir}")"
 	
 	display_message="cat /usr/share/message"
-	exec_command="sh -c ${display_message}"
-	add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command}"
+	exec_command=(sh -c "${display_message}")
+	add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command[@]}"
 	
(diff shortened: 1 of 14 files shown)


GabyCT (approved):

tests/integration/kubernetes/k8s-attach-handlers.bats
```diff
@@ -25,8 +25,8 @@ setup() {
 	policy_settings_dir="$(create_tmp_policy_settings_dir "${pod_config_dir}")"
 	
 	display_message="cat /usr/share/message"
-	exec_command="sh -c ${display_message}"
-	add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command}"
+	exec_command=(sh -c "${display_message}")
```
sprt, line 28: Just curious, why are we wrapping `sh` with `()` instead of `""` now? Is there a difference functionally?

danmihai1: If using:

```
command="/bin/ls"
exec_command="sh -c ${command}"
kubectl exec -- "${exec_command}"
```

The exec would fail because "${exec_command}" is a single argument to kubectl exec, and "sh -c /bin/ls" is not a valid binary to be executed.

However, when using an array of arguments:

```
command="/bin/ls"
exec_command=(sh -c "${command}")
kubectl exec -- "${exec_command[@]}"
```

kubectl exec gets 3 separate command line arguments and the exec works.

fidencio (approved):
lgtm, thanks @danmihai1!

danmihai1:
Rebased on the latest main code - to get a clean test run.

PR #10052, opened 2024-07-22
runtime-rs: Fix QEMU backend for runtime-rs

ananos:
It seems that PR #8070 broke QEMU for runtime-rs. This is an attempt to understand what's going on.


wainersm:
Hi @ananos ,

It's still not working: https://github.com/kata-containers/kata-containers/actions/runs/10040696370/job/27748841058?pr=10052

wainersm:
Hi @ananos !

Tests are passing, which is pretty good! But I'd like to understand the problem... IIUC it was a [change on runtime-rs/Makefile](https://github.com/kata-containers/kata-containers/pull/8070/commits/2d19f3fbd7bf41433c5d98d3bf43be660b5320c5#diff-21ae7c3a772119d42eee8b5e9037558e76af7b5e792398161985ae884ac1c461) that broke qemu-runtime-rs, but looking at your commit here I couldn't identify exactly what config value that changed. Mind to explain?

ananos:
https://github.com/kata-containers/kata-containers/pull/10052/files#diff-21ae7c3a772119d42eee8b5e9037558e76af7b5e792398161985ae884ac1c461L317 this is the part where it failed for qemu. That's the reason we moved every variable to have a custom, per-hypervisor instance, so that each one hypervisor gets it's pre-defined one.

pmores:
Thanks @ananos , this fix is great news!  However, after spending (too much) time trying to figure out what this change actually does I'm noticing it contains a lot of what looks like spurious change (mostly lines just moved around).  I suggest that you either revert those parts of the change if they are in fact spurious, or separate them into a standalone clean-up commit so as they don't obfuscate the actual functional core of this change.

Also, I'm under the impression that the idea of this change is to walk away from general settings valid across hypervisors (e.g. `DEFSTATICRESOURCEMGMT`), potentially customised by hypervisor-specific variants (e.g. `DEFSTATICRESOURCEMGMT_CLH`), and to fully rely on hypervisor-specific settings instead.  If that's the case this should be explicitly stated in the message of a commit that implements this transition (and this transition only).

This change fixes breakage caused by PR #8070 after a week's struggle due to the sheer size and impenetrability of #8070.  It would be a shame if the fix was another commit of that same kind. :-)

ananos:
agree 100%. The sheer fact that the Makefile logic assumes that if more hypervisors are available then the last one would override vars set by the previous ones seems misleading. I will update the commit message to reflect that.

The rationale was that this one commit adds per-hypervisor options, trying to address the above issue. But I agree we need to clarify it even more, so I'll split the commit into two: 
- first one will add the KERNELPARAMS_FC, decoupling the KERNELPARAMS var for FC
- second one will add the per-hypervisor vars for the resource mgt/cgroups etc.

will update later today and see how that plays out with the tests. I see a couple of required tests failing, not sure why.

src/runtime-rs/Makefile
```diff
@@ -320,7 +318,7 @@ ifneq (,$(FCCMD))
     KERNELPATH_FC = $(KERNELDIR)/$(KERNEL_NAME_FC)
     DEFSANDBOXCGROUPONLY_FC = true
     RUNTIMENAME := virt_container
-    DEFSTATICRESOURCEMGMT_FC ?= true
+    DEFSTATICRESOURCEMGMT_FC = true
```
gkurz, line 321: Any reason not to switch to simple expansion `:=` like the other ones ?

src/runtime-rs/Makefile
```diff
@@ -170,11 +170,7 @@ DEFMSIZE9P := 8192
 DEFVFIOMODE := guest-kernel
 ##VAR DEFSANDBOXCGROUPONLY=<bool> Default cgroup model
 DEFSANDBOXCGROUPONLY ?= false
-DEFSANDBOXCGROUPONLY_DB ?= true
-DEFSANDBOXCGROUPONLY_FC ?= true
 DEFSTATICRESOURCEMGMT ?= false
```
gkurz, line 173:
```
greg@bahia:[feat_fix_qemu_after_8070]kata-containers$ git grep DEFSTATICRESOURCEMGMT src/runtime-rs
src/runtime-rs/Makefile:DEFSTATICRESOURCEMGMT ?= false
src/runtime-rs/Makefile:    DEFSTATICRESOURCEMGMT_DB := false
src/runtime-rs/Makefile:    DEFSTATICRESOURCEMGMT_CLH := true
src/runtime-rs/Makefile:    DEFSTATICRESOURCEMGMT_QEMU := true
src/runtime-rs/Makefile:    DEFSTATICRESOURCEMGMT_FC = true
src/runtime-rs/Makefile:USER_VARS += DEFSTATICRESOURCEMGMT
src/runtime-rs/Makefile:USER_VARS += DEFSTATICRESOURCEMGMT_DB
src/runtime-rs/Makefile:USER_VARS += DEFSTATICRESOURCEMGMT_FC
src/runtime-rs/Makefile:USER_VARS += DEFSTATICRESOURCEMGMT_CLH
src/runtime-rs/Makefile:USER_VARS += DEFSTATICRESOURCEMGMT_QEMU
src/runtime-rs/config/configuration-cloud-hypervisor.toml.in:static_sandbox_resource_mgmt=@DEFSTATICRESOURCEMGMT_CLH@
src/runtime-rs/config/configuration-dragonball.toml.in:static_sandbox_resource_mgmt=@DEFSTATICRESOURCEMGMT_DB@
src/runtime-rs/config/configuration-qemu-runtime-rs.toml.in:static_sandbox_resource_mgmt=@DEFSTATICRESOURCEMGMT_QEMU@
src/runtime-rs/config/configuration-rs-fc.toml.in:static_sandbox_resource_mgmt=@DEFSTATICRESOURCEMGMT_FC@
```

Unless I'm missing something, it seems that `DEFSTATICRESOURCEMGMT` isn't used anymore. Maybe add a commit to fully remove it ?

src/runtime-rs/Makefile
pmores: I was wondering about this anomaly as well.

ananos: the previous ?= -> = was correcting a typo. This one (amended the previous commit) makes is homogeneous across all hypervisors. BTW, TIL the [difference of := and = in Makefiles](https://stackoverflow.com/questions/4879592/whats-the-difference-between-and-in-makefile) ;)

src/runtime-rs/Makefile
ananos: correct, pushed a new commit that removes the generic vars and keeps only the per-hypervisor ones.

lifupan (approved):

pmores (approved):
lgtm, thanks @ananos !

wainersm (approved):
@ananos thanks for promptly fixing it! I checked that the qemu-runtime-rs jobs are now passing and the jobs not passing aren't unrelated.

Issue #9921, opened 2024-06-26
release: component tagging with release isn't working


stevenhorsman:
In https://github.com/kata-containers/kata-containers/blob/fe7ef878d2101e92bb0268d44b38743e726aa0bc/tools/packaging/kata-deploy/local-build/kata-deploy-binaries.sh#L1142-L1146 I thought I added the ability for certain release components to be tagged with the release version. We couldn't test it until a release was done. With 3.6.0 I can see that this code didn't work. e.g. https://github.com/kata-containers/kata-containers/actions/runs/9552945045/job/26330643058

We can see that in the build steps the RELEASE env was set
```
 env:
    RELEASE: yes 
```

however the output
```
Pushing agent with tags: latest-main 5f11c0f144037d8d8f546c89a0392dcd84fa99e2
```
shows that the `3.6.0` wasn't added to the list of tags.

I guess there could be two things going on from what I can see:
1. The `[ "${RELEASE}" == "yes" ]` isn't getting triggered. I'm not sure why this would be the case, but bash square brackets with single vs double do often confuse me.
2. `$(cat "${version_file}")` isn't returning any output either as the path is wrong, or the content is empty. We use this content elsewhere, so I'm not sure why this is broken either.

Unfortunately I think we'd need some more debug added to understand this and it will be slow progress to only test when we release, so I'll try and get a local reproduce first and see if I can work anything out.

zvonkok:
`[ "${RELEASE}" == "yes" ]` this is not correct either `[ "${RELEASE}" = "yes" ]` or `[[ "${RELEASE}" == "yes" ]]`

zvonkok:
I think we should enable `shellcheck` as  GHA to check such "errors".

stevenhorsman:
> `[ "${RELEASE}" == "yes" ]` this is not correct either `[ "${RELEASE}" = "yes" ]` or `[[ "${RELEASE}" == "yes" ]]`

Hmm, I tested it locally and they all seem to work:
```
export RELEASE=yes
# [ "${RELEASE}" == "yes" ] && echo "success"
success
# [ "${RELEASE}" = "yes" ] && echo "success"
success
# [[ "${RELEASE}" == "yes" ]] && echo "success"
success
```

stevenhorsman:
> I think we should enable shellcheck as GHA to check such "errors".

I ran shellcheck against kata-deploy-binaries.sh and there were 122 errors/warnings, but this wasn't one of them!

[...]

PR #10053, opened 2024-07-22
ci: cache: Pass through RELEASE env

stevenhorsman:
In kata-deploy-binaries.sh we want to understand if we are running as part of a release, so we need to pass through the RELEASE env from the workflow, which I missed in
https://github.com/kata-containers/kata-containers/pull/9550

Fixes: #9921


dborquez (approved):
LGTM, thanks @stevenhorsman

GabyCT (approved):

Issue #10062, opened 2024-07-24
GHA: the action `Run tests` times out after 30 minutes in CoCo tests


ChengyuZhu6:
# Description of problem

the action `Run tests` times out after 30 minutes, causing the CI to fail.
https://github.com/kata-containers/kata-containers/actions/runs/10069229690/job/27836773696?pr=9999#:~:text=Error%3A%20The%20action%20%27Run%20tests%27%20has%20timed%20out%20after%2030%20minutes.

PR #10063, opened 2024-07-24
gha: Increase timeout to run CoCo tests

ChengyuZhu6:
This PR increases the timeout for running the CoCo tests to avoid random failures. These failures occur when the action `Run tests` times out after 30 minutes, causing the CI to fail.

Fixes: #10062


arronwy (approved):

fidencio (approved):
lgtm, thanks @ChengyuZhu6!

PR #10069, opened 2024-07-25
genpolicy: validate each exec command line arg

danmihai1:
Generate policy that validates each exec command line argument, instead of joining those args and validating the resulting string. Joining the args ignored the fact that some of the args might include space characters.

The older format from genpolicy-settings.json was similar to:

    "ExecProcessRequest": {
      "commands": [
		"sh -c cat /proc/self/status"
        ],
      "regex": []
    },

That format will not be supported anymore. genpolicy will detect if its users are trying to use the older "commands" field and will exit with a relevant error message in that case.

The new settings format is:

    "ExecProcessRequest": {
      "allowed_commands": [
        [
          "sh",
          "-c",
          "cat /proc/self/status"
        ]
      ],
      "regex": []
    },

### DIFF
--- a/tests/integration/kubernetes/k8s-policy-pod.bats
+++ b/tests/integration/kubernetes/k8s-policy-pod.bats
@@ -17,8 +17,8 @@ setup() {
 	get_pod_config_dir
 	policy_settings_dir="$(create_tmp_policy_settings_dir "${pod_config_dir}")"
 	
-	exec_command="printenv data-3"
-	add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command}"
+	exec_command=(printenv data-3)
+	add_exec_to_policy_settings "${policy_settings_dir}" "${exec_command[@]}"
 	add_requests_to_policy_settings "${policy_settings_dir}" "ReadStreamRequest"
 
 	correct_configmap_yaml="${pod_config_dir}/k8s-policy-configmap.yaml"
@@ -63,10 +63,8 @@ wait_for_pod_ready() {
 }
 
 @test "Able to read env variables sourced from configmap using envFrom" {
-	kubectl create -f "${correct_configmap_yaml}"
-	kubectl create -f "${correct_pod_yaml}"
-	kubectl wait --for=condition=Ready "--timeout=${timeout}" pod "${pod_name}"
-	expected_env_var=$(kubectl exec "${pod_name}" -- printenv data-3)
+	wait_for_pod_ready
+	expected_env_var=$(kubectl exec "${pod_name}" -- ${exec_command[@]})
 	[ "$expected_env_var" = "value-3" ] || fail "expected_env_var is not equal to value-3"
 }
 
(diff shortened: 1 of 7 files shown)


ChengyuZhu6 (commented):
Good work @danmihai1 ! Unrelated to this, do you think we should replace `sh` to `"$sh_command"` in this line? https://github.com/kata-containers/kata-containers/blob/4e003a212541851aa2c297da926dc9e221f9ed48/tests/integration/kubernetes/k8s-exec.bats#L63

tests/integration/kubernetes/tests_common.sh
```diff
@@ -237,22 +237,24 @@ add_exec_to_policy_settings() {
 
 	local -r settings_dir="$1"
 
-	# TODO: teach genpolicy to work with an array of args, instead of joining the args here.
+	# Create a JSON array of strings containing all the args of the command to be allowed.
```
burgerdev, line 240: An easier way to do that may be

```sh
printf "%s\n" "$@" | jq -R | jq -sc
```

burgerdev (approved):
Thanks, that's an important improvement.

tests/integration/kubernetes/tests_common.sh
danmihai1: Thanks! I made this change.

danmihai1:
> Good work @danmihai1 ! Unrelated to this, do you think we should replace `sh` to `"$sh_command"` in this line?
> 
> https://github.com/kata-containers/kata-containers/blob/4e003a212541851aa2c297da926dc9e221f9ed48/tests/integration/kubernetes/k8s-exec.bats#L63

@ChengyuZhu6 , good point - thanks! I pushed that change.

ChengyuZhu6 (approved):
LGTM thanks @danmihai1

tests/integration/kubernetes/k8s-policy-pod.bats
```diff
@@ -63,10 +63,8 @@ wait_for_pod_ready() {
 }
 
 @test "Able to read env variables sourced from configmap using envFrom" {
-	kubectl create -f "${correct_configmap_yaml}"
-	kubectl create -f "${correct_pod_yaml}"
-	kubectl wait --for=condition=Ready "--timeout=${timeout}" pod "${pod_name}"
-	expected_env_var=$(kubectl exec "${pod_name}" -- printenv data-3)
+	wait_for_pod_ready
+	expected_env_var=$(kubectl exec "${pod_name}" -- ${exec_command[@]})
```
burgerdev, line 67:
```suggestion
	expected_env_var=$(kubectl exec "${pod_name}" -- "${exec_command[@]}")
```
I think this needs to be quoted to prevent word splitting of the `exec_command` array entries.

burgerdev (approved):

GabyCT (approved):

tests/integration/kubernetes/k8s-policy-pod.bats
danmihai1: Good point, but I would like to clean that up in a future PR, because:
- We don't seem to have a splitting problem in this particular example, and
- Running CI tests again after they passed is a relatively expensive activity.

Issue #10067, opened 2024-07-24
Kata fails to build with Rust 1.79


emanuellima1:
# Description of problem

When trying to build kata-containers on Linux using rustc v1.79, the following error occurs:

```bash
Compiling kata-agent v0.1.0 ($PWD/kata-containers/kata-containers-3.7.0/src/agent)
error: field `supports_seccomp` is never read
   --> src/config.rs:106:9
    |
94  | pub struct AgentConfig {
    |            ----------- field in this struct
...
106 |     pub supports_seccomp: bool,
    |         ^^^^^^^^^^^^^^^^
    |
    = note: `AgentConfig` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`

error: could not compile `kata-agent` (bin "kata-agent") due to 1 previous error
make: *** [Makefile:137: target/x86_64-unknown-linux-gnu/release/kata-agent] Error 101
erro: Status de saída de /var/tmp/rpm-tmp.WB29w7 inválido (%build)
```

This was specifically when building a RPM for Fedora 40, but it fails to build with the same error when compiling just the repo. This has been observed for both Kata 3.7.0 and HEAD.

Compiling with Rust v1.78 works fine. 

# Expected result

I expected the RPM and/or the binary to be built normally, as it does with Rust v1.78.

# Actual result

Rust compilation error when compiling the kata-agent binary

emanuellima1:
FYI: this bug is blocking the update of kata-containers on Fedora 40 and Rawhide.

zvonkok:
How are you kicking off the build?
Is libseccomp installed via package manager? 
We usually build libseccomp alongside the kata-agent to make sure they're compatible.

emanuellima1:
It has been installed via package manager. I'm building just the kata-agent by doing make on it's makefile only. Is this wrong?

```bash
libseccomp.so.2 (libc6,x86-64) => /lib64/libseccomp.so.2
libseccomp.so (libc6,x86-64) => /lib64/libseccomp.so
```

```bash
-lseccomp
```

c3d:
I tried with rustc 1.80.0 from rustup, and 

```
   Compiling kata-sys-util v0.1.0 (/home/ddd/Work/kata/kata-containers/src/libs/kata-sys-util)
error: unexpected `cfg` condition value: `powerpc64le`
[...]

Issue #7892, opened 2023-09-11
RUSTFLAGS is overriden by makefiles


c3d:
# Description of problem

As documented in [this bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2167235), Fedora canot really build `kata-containers` with the standard build options they use for other components, because various makefiles override Rustflags.

For example, the agent's code has stuff like this:

```
##TARGET optimize: optimized  build
optimize: show-summary show-header
	@RUSTFLAGS="-C link-arg=-s $(EXTRA_RUSTFLAGS) --deny warnings" cargo build --target $(TRIPLE) $(if $(findstring release,$(BUILD_TYPE)),--release) $(EXTRA_RUSTFEATURES)

##TARGET install: install agent
install: install-services
	@install -D $(TARGET_PATH) $(DESTDIR)/$(BINDIR)/$(TARGET)
```

Doing things this way makes it harder for a large-scale build system like Fedora's to set build options according to their own policy by setting the `RUSTFLAGS` variable.

# Expected result

Setting `RUSTFLAGS` should define the compilation flags used for various build steps.

# Actual result

What is passed through environment is largely ignored, at least for some common targets.

# Recommendation

Replace the above scheme with per-target environment variables, something like:

```
OPT_RUSTFLAGS?=-C link-arg=-s $(EXTRA_RUSTFLAGS) --deny warnings
optimize: RUSTFLAGS=$(RUSTFLAGS) $(OPT_RUSTFLAGS)
optimize: show-summary show-header
	@cargo build --target $(TRIPLE) $(if $(findstring release,$(BUILD_TYPE)),--release) $(EXTRA_RUSTFEATURES)
```

# Further information

See [the GNU documentation](https://www.gnu.org/software/make/manual/html_node/Values.html) regarding various ways variables can be overriden. I strongly suspect the current way of doing things is incorrect, it should be something like invoking a recursive `make` with `RUSTFLAGS=` instead.

Consider this makefile:

```
FOO=foo
BAR?=bar

all:
	@echo The value of FOO is $(FOO)
	@echo The value of BAR is $(BAR)
```

Here is what you get:

```
$ make
The value of FOO is foo
The value of BAR is bar
$ make FOO=makearg
The value of FOO is makearg
The value of BAR is bar
$ FOO=envarg make
The value of FOO is foo
The value of BAR is bar
$ make BAR=makearg
The value of FOO is foo
The value of BAR is makearg
$ BAR=envarg make
[...]

PR #10073, opened 2024-07-25
Temporary fix for the rustc 1.79 compilation error

emanuellima1:
kata-containers wasn't compiling on rustc 1.79
because dead code warnings were being treated
as compilation errors. This commit instructs the
compiler to ignore these warnings for now.

This solves #10067 

In the future, we must refactor how we set RUSTFLAGS: #7892


c3d:
I'm afraid that in doing so, your Fedora build would just have seccomp  disabled. I'm in the process of finishing a comment on #10067, but for now I don't really like this fix, as I think it simply hides a serious problem.

emanuellima1:
Seccomp is not being disabled, nothing on the logic of the software is being changed. This is only a change on the linter level.

c3d:
> Seccomp is not being disabled, nothing on the logic of the software is being changed. This is only a change on the linter level.

I agree for the linter message. But then you run into the linker problem, don't you? If not, then seccomp might be disabled. That's my concern.

emanuellima1:
In the end, I was only having the linter error when I tried to build kata-agent through its makefile. When I build the whole runtime-rs. kata-agent and everything else gets built without any linking problems

emanuellima1:
I've got a better idea on how to deal with this, so I'll change my PR

c3d:
> In the end, I was only having the linter error when I tried to build kata-agent through its makefile. When I build the whole runtime-rs. kata-agent and everything else gets built without any linking problems

That's weird. @pmores @gkurz @fidencio do you expect a dependency between `runtime-rs` and the agent?

@emanuellima1 Do you mean that you ran `make` from the `runtime-rs` directory? Does that even build the agent at all?

emanuellima1:
Yes, and as far as I could tell, the agent crate got built successfully.

emanuellima1:
@c3d @gkurz @pmores now, instead of adding a bunch of linter directives, I changed the RUSTFLAGS env vars to allow compilation with warnings. I think this is cleaner.

emanuellima1:
After all, `-D warnings` is a [code smell](https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html) and shouldn't be used in Kata at all.

c3d:
> After all, `-D warnings` is a [code smell](https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html) and shouldn't be used in Kata at all.

I am not sure that the link supports the statement that `-D warnings` is a code smell. First, it talks about crates, where it is justifiable to not block your users when a warning appears. Second, it seems to suggest using `-D warnings` for crates, precisely because it lets you check the warnings for yourself without imposing that on your dependents.

That being said, I think that disabling that `RUSTFLAGS` in the Fedora build makes sense. I am not sure it is wise for upstream. It seems that the team's policy is that we want warnings to be painful enough that they get addressed.

c3d:
I think that we need the community to agree that not denying warnings is a good idea. I personally think that it catches interesting stuff. So I would rather find a way to make this a downstream-only change. What about adding a DENY_WARNINGS variable that by default has --deny warnings but that you can clear when calling it from the Fedora build?

pmores:
Sorry guys, I've only noticed now.

I'm not aware that making runtime-rs builds agent, too.  Then again, I mostly build my dev builds using a plain `cargo build` in the `runtime-rs` directory.

Now, I do think that treating warnings as errors is a terrible idea and we're seeing some its impacts on a daily basis.  It basically makes builds fragile and prone to spurious failures which creates obstacles for contributors, then during bisection if not caught during review, and in various special-case situation like this one.

That said, as much as I'd love to see this turned off, I agree with @c3d that a broader community consensus would be necessary.

emanuellima1:
Hey everybody, reopening this conversation, how can we best move forward with this?

fidencio (changes requested):
I don't think this is the correct approach to take, especially one day before the planned release.

The correct approach would be:
* Bump the rust version to the latest possible one
  * Make sure to sync with the parts involved in the project to be sure it won't affect them negatively
* Fix all the warnings in a different commit, but in the same PR where the bump is made
* Have this done as early as possible in the release cycle

I don't see the need to rush this in, mainly knowing that there was a request coming from Red Hat itself to stay on rust 1.75.  We can discuss more during the AC meeting, but I don't see this one as a release blocker.

PR #10078, opened 2024-07-26
tee: osbuilder: Set /run to use 50% of the image with systemd

fidencio:
Let's ensure at least 50% of the memory is used for /run, as systemd by default forces it to be 10%, which is way too small even for very small workloads.

This is only done for the rootfs-confidential image.

This is built atop of https://github.com/kata-containers/kata-containers/pull/7016, thus I'm keeping Alex's Signed-off-by, and also Chengyu's and Arron's as the approach taken is closer to what we have downstream.

### DIFF
--- a/tools/osbuilder/rootfs-builder/rootfs.sh
+++ b/tools/osbuilder/rootfs-builder/rootfs.sh
@@ -708,6 +708,20 @@ EOF
 		mkdir -p "${ROOTFS_DIR}/etc/systemd/system/kata-containers.target.wants"
 		ln -sf "/usr/lib/systemd/system/dbus.socket" "${ROOTFS_DIR}/etc/systemd/system/kata-containers.target.wants/dbus.socket"
 		chmod g+rx,o+x "${ROOTFS_DIR}"
+
+		if [ "${CONFIDENTIAL_GUEST}" == "yes" ]; then
+			info "Tweaking /run to use 50% of the available memory"
+			# Tweak the kata-agent service to have /run using 50% of the memory available
+			# This is needed as, by default, systemd would only allow 10%, which is way
+			# too low, even for very small test images
+			fstab_file="${ROOTFS_DIR}/etc/fstab"
+			[ -e ${fstab_file} ] && sed -i '/\/run/d' ${fstab_file}
+			echo "tmpfs /run tmpfs nodev,nosuid,size=50% 0 0" >> ${fstab_file}
+
+			kata_systemd_target="${ROOTFS_DIR}/usr/lib/systemd/system/kata-containers.target"
+			grep -qE "^Requires=.*systemd-remount-fs.service.*" ${kata_systemd_target} || \
+				echo "Requires=systemd-remount-fs.service" >> ${kata_systemd_target}
+		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -702,12 +702,27 @@ EOF
...
 	else
+		info "Setup systemd-base environment for kata-agent"
 		# Setup systemd-based environment for kata-agent
 		mkdir -p "${ROOTFS_DIR}/etc/systemd/system/basic.target.wants"
 		ln -sf "/usr/lib/systemd/system/kata-containers.target" "${ROOTFS_DIR}/etc/systemd/system/basic.target.wants/kata-containers.target"
 		mkdir -p "${ROOTFS_DIR}/etc/systemd/system/kata-containers.target.wants"
 		ln -sf "/usr/lib/systemd/system/dbus.socket" "${ROOTFS_DIR}/etc/systemd/system/kata-containers.target.wants/dbus.socket"
 		chmod g+rx,o+x "${ROOTFS_DIR}"
+
+		if [ "${CONFIDENTIAL_GUEST}" == "yes" ]; then
```
fitzthum, line 713: Any reason not to do this for non-confidential guests?

tools/osbuilder/rootfs-builder/rootfs.sh
```diff
@@ -702,12 +702,27 @@ EOF
...
+			info "Tweaking /run to use 50% of the available memory"
+			# Tweak the kata-agent service to have /run using 50% of the memory available
+			# This is needed as, by default, systemd would only allow 10%, which is way
+			# too low, even for very small test images
+			fstab_file="${ROOTFS_DIR}/etc/fstab"
+			[ -e ${fstab_file} ] && sed -i '/\/run/d' ${fstab_file}
+			echo "tmpfs /run tmpfs nodev,nosuid,size=50% 0 0" >> ${fstab_file}
```
fitzthum, line 720: It might be cool if this 50% was configurable but I'm not sure if anyone would actually use that feature.

fitzthum (approved):
LGTM. Two minor comments

ChengyuZhu6 (approved):
LGTM thanks @fidencio

GabyCT (approved):

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: non-confidential guests do not pull the images inside the podsandbox.  So, no reason to mess around with this on that case.

tools/osbuilder/rootfs-builder/rootfs.sh
fidencio: I agree, I don't think it's worth the trouble though, as this is mostly a stopgap till trusted storage is in place.

tools/osbuilder/rootfs-builder/rootfs.sh
fitzthum: Yeah it still seems a bit fishy to me that the systemd and non-systemd cases have different behavior, but I guess it's safest not to touch the non-confidential case.

Issue #10085, opened 2024-07-29
ci: Implement execution of post action script on self-hosted runners for Nightly and PR tests


AdithyaKrishnan:
Currently, the CI does not have the capability to run a post action script on the self-hosted runners. Having such a feature would greatly simplify debugging and clean-up to stabilize the nodes after a test run.

PR #10086, opened 2024-07-29
ci: Run post action script on AMD SEV/SNP nodes

AdithyaKrishnan:
ci: Fixes Issue #10085
Currently, the CI does not have the capability to run a post action script on the self-hosted runners.
This feature would greatly simplify debugging and clean-up to stabilize the nodes after a test run.

### DIFF
--- a/.github/workflows/run-kata-coco-tests.yaml
+++ b/.github/workflows/run-kata-coco-tests.yaml
@@ -154,6 +154,13 @@ jobs:
         if: always()
         run: bash tests/integration/kubernetes/gha-run.sh cleanup-snapshotter
 
+      - name: Take a post-action
+        if: always()
+        run:
+          if [ -f ${HOME}/script/post_action.sh ]; then
+            ${HOME}/script/post_action.sh sev
+          fi
+
   run-k8s-tests-sev-snp:
     strategy:
       fail-fast: false
@@ -228,6 +235,13 @@ jobs:
       - name: Delete CoCo KBS
         if: always()
         run: bash tests/integration/kubernetes/gha-run.sh delete-coco-kbs
+
+      - name: Take a post-action
+        if: always()
+        run:
+          if [ -f ${HOME}/script/post_action.sh ]; then
+            ${HOME}/script/post_action.sh sev-snp
+          fi
 
   # Generate jobs for testing CoCo on non-TEE environments
   run-k8s-tests-coco-nontee:


.github/workflows/run-kata-coco-tests.yaml
```diff
@@ -154,6 +154,13 @@ jobs:
         if: always()
         run: bash tests/integration/kubernetes/gha-run.sh cleanup-snapshotter
 
+      - name: Take a post-action
+        if: always()
+        run:
+          if [ -f ${HOME}/script/post_action.sh ]; then
+            ${HOME}/script/post_action.sh sev
+          fi
+
```
sprt, line 163: Thanks @AdithyaKrishnan! We should follow the same pattern as the previous cleanup steps, as in call `gha-run.sh` with some command. 🙂 Ditto for `run-k8s-tests-sev-snp`.

sprt (changes requested):

.github/workflows/run-kata-coco-tests.yaml
fidencio: What's this script about?

Why not following https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job ?

wainersm: @BbolroC opened https://github.com/kata-containers/kata-containers/issues/9934 aiming to create a convention on post-action scripts among the bare-metal machines. While we don't have a consensus,  @AdithyaKrishnan is implementing the same post-action logic as of some 390x jobs (e.g. https://github.com/kata-containers/kata-containers/blob/main/.github/workflows/run-k8s-tests-on-zvsi.yaml#L92) so SNP/SEV jobs aren't blocked.

sprt:
Removing `ok-to-test` label to minimize spending and since this a YAML-only change. Let's either skip CI here or add it back if we make testable changes.

.github/workflows/run-kata-coco-tests.yaml
fidencio: This is process overload, IMHO.

The script is not shared, anyways. Follow the intructions of the link above, ensure the script is executable, and be happy.

BbolroC: Please have a look at https://github.com/kata-containers/kata-containers/issues/9934#issuecomment-2258094441 if you follow Fabi's suggestion. Thanks.

zvonkok: I would also highly suggest to follow what @BbolroC and @fidencio are proposing, this way the pre/post actions are runner specific and not GHA specific. GHA should be runner agnostic.

AdithyaKrishnan: We got it working by setting the variables you shared. Closing this PR now.

PR #10090, opened 2024-07-29
ci: DEBUG=1 support in install_libseccomp.sh

danmihai1:
Add DEBUG=1 support in install_libseccomp.sh, for easier debugging of this script.


danmihai1:
install_libseccomp.sh is succeeding again today - unlike yesterday.

Issue #10094, opened 2024-07-30
runtime-rs: vfio device failed with given non-filtered /dev/vfio/vfio


Apokleos:
# Get your issue reviewed faster

The kata will fail when running vfio device passthrough with /dev/vfio/vfio which is not filtered

zvonkok:
In which `vfio_mode` are you running? `guest-kernel` or `vfio`?

Apokleos:
Hey @zvonkok I am using `vfio` mode not `guest-kernel`.

PR #10095, opened 2024-07-30
runtime-rs: bugfix for non-filtered dev-vfio-vfio

Apokleos:
The kata will fail when running vfio device passthrough with /dev/vfio/vfio which is not filtered.
We need to enhance the filtering criteria in order to make passthrough fine.

Fixes: #10094


zvonkok:
This depends on the `vfio_mode` if `guest-kernel` than filter `/dev/vfio/vfio` if `vfio_mode=vfio` than do not filter it.

Apokleos:
> This depends on the `vfio_mode` if `guest-kernel` than filter `/dev/vfio/vfio` if `vfio_mode=vfio` than do not filter it.
>
Yeah, I need more test on this and make this commit better

PR #10099, opened 2024-07-30
metrics: Update memory tests to use grep -F

GabyCT:
This PR updates the memory tests like fast footprint to use grep -F instead of fgrep as this command has been deprecated.


Apokleos (approved):
Thx @GabyCT 
Nit: IMO, the two commits do the same thing and they might be squashed into one commit.
Anyway, LGTM.

amshinde (approved):

amshinde:
LGTM. Agree with @Apokleos, better to squash both commits.

GabyCT:
changes applied :)

Issue #10011, opened 2024-07-15
tests: tdx: Get authenticated image pull tests working for TDX


stevenhorsman:
In initial testing of the authenticated image pull https://github.com/kata-containers/kata-containers/pull/9904
The positive test for TDX works, but the negative tests are both failing with:
```
assert_logs_contain "${node}" kata "${node_start_time}" "failed to pull manifest Not authorize"' failed
```
,but I can see in the log that the message is there:
```
# Jul 15 08:45:14 [984fee00bd70.jf.intel.com](http://984fee00bd70.jf.intel.com/) kata[3896558]: time="2024-07-15T08:45:14.391889998-07:00" level=error msg="createContainer failed" error="rpc error: code = Internal desc = failed to pull manifest Not authorized:
```
I'm not sure if there could be an issue related to the `node`, or `node_start_time` being missing or something else, but it's difficult to check this without having access to the system.

ChengyuZhu6:
https://github.com/kata-containers/kata-containers/pull/10031

ChengyuZhu6:
This issue not only occurred in the authenticated image pull tests but was also found in the trusted storage tests.

BbolroC:
The same failure is happening for IBM SE: https://github.com/kata-containers/kata-containers/actions/runs/10017693261/job/27692208889 The test passes when I run the test while staying in a ssh session, but no success in a scheduled cron job. I am expecting the PR by @ChengyuZhu6 will fix the issue on SE as well. 😉

BbolroC:
It seems #10031 cannot solve the issue on IBM SE. Interestingly, it failed for the first run, but second run. This kind of behaviour has been observed for the past 3 days. I was told by @ryansavino that SEV is also struggling with a similar issue. I will reopen this issue and look into the machine.

BbolroC:
Update: the nightly test passed if the journal was reset at the start of the job. I don't know why yet, but it explains that the 1st run always fails (almost 23h have passed since the last run) while the 2nd run (manually triggered) passes. I will figure out a root cause today. Thanks.

PR #10107, opened 2024-07-31
tests: k8s: Rotate & cleanup journal for every run

fidencio:
This will help to avoid huge logs, and allow us to debug issues in a better way.

I've requested the review of the ones who have baremetal machines which will be affected by this.

### DIFF
--- a/tests/integration/kubernetes/gha-run.sh
+++ b/tests/integration/kubernetes/gha-run.sh
@@ -265,6 +265,17 @@ function uninstall_kbs_client() {
 }
 
 function run_tests() {
+	if [ "${K8S_TEST_HOST_TYPE}" = "baremetal" ]; then
+		# Baremetal self-hosted runners end up accumulating way too much log
+		# and when those get displayed it's very hard to understand what's
+		# part of the current run and what's something from the past coming
+		# to haunt us.
+		#
+		# With this in mind, let's ensure we do rotate the logs on every single
+		# run of the tests, as its first step.
+		sudo journalctl --vacuum-time 1s --rotate
+	fi
+
 	ensure_yq
 	platform="${1:-}"
 


tests/integration/kubernetes/gha-run.sh
```diff
@@ -265,6 +265,10 @@ function uninstall_kbs_client() {
 }
 
 function run_tests() {
+	if [ "${K8S_TEST_HOST_TYPE}" = "baremetal" ]; then
```
danmihai1, line 268: It would be nice to explain here why this is useful for bare metal, but not for Host VMs.

danmihai1 (approved):

BbolroC (approved):
LGTM, the journal on the SE machine is cleaned up in a similar way before the nightly test. Thanks @fidencio for suggesting it.

BbolroC:
FYI: In the s390x self-hosted runner (which operates 24/7 and likely accumulates a significant amount of logs), the following error was observed:

```
not ok 2 Test we cannot pull an image that exceeds the memory limit inside the guest
# (from function `assert_logs_contain' in file lib.sh, line 80,
#  in test file k8s-guest-pull-image.bats, line 86)
#   `assert_logs_contain "$node" kata "$node_start_time" \' failed
```

But the test passed when `sudo journalctl --vacuum-time 1s --rotate` was run as a pre-hook.

I believe we could apply this approach to any machine running 24/7 (e.g., baremetal and self-hosted runners) rather than immediately resorting to provisioning.

I mentioned the similar experience at https://github.com/kata-containers/kata-containers/issues/10011#issuecomment-2247069997. Thanks.

fidencio:
I've added the comment, addressing @danmihai1's review, and rebased atop of main.
I will merge it as soon as the tests pass.

PR #10111, opened 2024-08-01
ci: Temporarily remove arm64 builds

fidencio:
It's been a reasonable time that we're not able to even build arm64 artefacts.

For now I am removing the builds as it doesn't make sense to keep running failing builds, and those can be re-enabled once we have arm64 machines plugged in that can be used for building the stuff, and maintainers for those machines.

The `arm-jetson-xavier-nx-01` is also being removed from the runners.

Looped in @ananos  and @zvonkok  to make them aware of the situation.

### DIFF
--- a/.github/workflows/payload-after-push.yaml
+++ b/.github/workflows/payload-after-push.yaml
@@ -17,13 +17,13 @@ jobs:
       target-branch: ${{ github.ref_name }}
     secrets: inherit
 
-  build-assets-arm64:
-    uses: ./.github/workflows/build-kata-static-tarball-arm64.yaml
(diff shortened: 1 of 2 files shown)


ananos (approved):
thanks @fidencio. Makes sense -- we are currently stuck setting up kcli on the ampere nodes (apparently, there's a weird error, preventing us from easily setting up VMs for the self-hosted runners). Once we have resolved that, we can revert this / re-enable these builds.

GabyCT (approved):

wainersm (approved):
Thanks @fidencio !

Issue #10112, opened 2024-08-01
scsi=off parameter for virtio-blk-pci is incompatible with QEMU 9.1


Freax13:
When the runtime sets up a `virtio-blk-pci` device, it adds `scsi=off` if SCSI is not enabled (AFAICT SCSI is never enabled):
https://github.com/kata-containers/kata-containers/blob/c784fb650862aa22b69db8533eaf2c83b38c87be/src/runtime/pkg/govmm/qemu/qemu.go#L1301-L1303
The default for `scsi` used to be `on`, but it got changed to `off` a long time ago. Additionally the `scsi` parameter was [deprecated](https://gitlab.com/qemu-project/qemu/-/commit/5847c750f3d31959022cc67a03870ba1a229d59b) a while ago and got [removed](https://gitlab.com/qemu-project/qemu/-/commit/a271b8d7b2f39275a05e49deb7c8edc20b7a8279) with QEMU 9.1.

We shouldn't emit the `scsi=off` parameter anymore.

Freax13:
I'm working on a fix.

PR #10113, opened 2024-08-01
qemu: don't emit scsi parameter

Freax13:
This parameter has been deprecated for a long time and QEMU 9.1.0 finally removes it.

Fixes: #10112

### DIFF
--- a/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
+++ b/src/runtime-rs/crates/hypervisor/src/qemu/cmdline_generator.rs
@@ -820,7 +820,6 @@ impl ToQemuParams for BlockBackend {
 struct DeviceVirtioBlk {
     bus_type: VirtioBusType,
     id: String,
-    scsi: bool,
     config_wce: bool,
     share_rw: bool,
 }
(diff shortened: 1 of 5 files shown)


Apokleos:
Thx @Freax13 
I think it'll be better to split the PR into two commits, one for runtime-rs and the other one for kata-runtime/qemu.

fidencio:
@Freax13, very nice seeing you around! :-)

So, while your approach is correct, the version of QEMU that Kata Containers ships is still v8.2.4, see: https://github.com/kata-containers/kata-containers/blob/43dca8deb4891678f8a62c112749ac0938b373c6/versions.yaml#L88-L100

My sincere recommendation is that we don't take this patch till we bump the QEMU version being used.
@gkurz, what's your take on bumping QEMU to its latest release?

Freax13:
> @Freax13, very nice seeing you around! :-)

:)

> So, while your approach is correct, the version of QEMU that Kata Containers ships is still v8.2.4, see: https://github.com/kata-containers/kata-containers/blob/43dca8deb4891678f8a62c112749ac0938b373c6/versions.yaml#L88-L100

Ah, I forgot to mention that before the parameter was deprecated, the default for `SCSI` was switched to off, so the code govmm is already outdated and serves no purpose in newer QEMU versions, including v8.2.4. We don't need to upgrade QEMU for this patch to be correct.

amshinde (approved):

fidencio:
> Ah, I forgot to mention that before the parameter was deprecated, the default for `SCSI` was switched to off, so the code govmm is already outdated and serves no purpose in newer QEMU versions, including v8.2.4. We don't need to upgrade QEMU for this patch to be correct.

That's good info.

lgtm, thanks @Freax13, and welcome to the Kata Containers community! <3

fidencio (approved):
lgtm, thanks @Freax13!

PR #10114, opened 2024-08-01
kata-manager: Ensure distro specific TDX config is set

fidencio:
We've done something quite similar for kata-deploy, but I've noticed we forgot about the kata-manager counterpart.

Here you can find results (and how to test) without the patch and with the patch:
* without: https://gist.github.com/fidencio/b95fa4ea34e8f10e18dde4eabf18327d
* with: https://gist.github.com/fidencio/368211840813afe6ae41df60264e4f07

@GabyCT, @cmaf, @dborquez, could you give it a try on your TDX machines?

### DIFF
--- a/utils/kata-manager.sh
+++ b/utils/kata-manager.sh
@@ -833,6 +833,34 @@ install_kata()
 		sudo ln -sf "$from_path" "$link_dir"
 	done
 
+	local tdx_qemu_config="/opt/kata/share/defaults/kata-containers/configuration-qemu-tdx.toml"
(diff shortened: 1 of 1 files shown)


dborquez (approved):
LGTM thank you @fidencio

cmaf (approved):
@fidencio LGTM, got the expected results per the gist

GabyCT (approved):

fidencio:
> /test

This doesn't work anymore, AFAIR.  We just need to add the `ok-to-test` label.

